Issue #75598, opened 2020-08-16
At opt-levels <= 1 the arithmetic operation methods do not get inlined, preventing other optimisations


nagisa:
Consider code like this:

```rust
#![feature(core_intrinsics)]

pub fn method(a: usize, b: usize) -> usize {
    a.wrapping_sub(b)
}

pub fn intrinsic(a: usize, b: usize) -> usize {
    std::intrinsics::wrapping_sub(a, b)
}
```

[compiler explorer](https://godbolt.org/z/5sq1a7)

at `-Copt-level=1` and lower, the generated assembly for versions with the method call will generate a function call, rather than direct operation. Once the inlining fails, other optimisations that could be done are inhibited, especially at -Copt-level=1`.

More generally speaking, I wonder if we might want to make these methods have a special annotation that would make the compiler generate the instructions directly much like it does for intrinsics right now. These seem like basic enough that `#[inline(always)]` might not be good enough (it being just a hint) and also possibly more expensive than necessary (something needs to do the inlining still).

alecmocatta:
Relatedly, https://github.com/rust-lang/rust/issues/74362 is an instance where not inlining a function that wraps an intrinsic inhibits DCE.

bugadani:
My naive assumption is that `inline(always)` would inline shorter functions more eagerly, since the relative cost of a function call is higher and also shorter functions don't contribute as much to binary bloat. Is this not the case?

Lokathor:
Using the compiler explorer link, it seems like simply tagging the `wrapping_` operations as `#[inline(always)]` will make them get inlined even at opt-level=0

AngelicosPhosphoros:
>These seem like basic enough that #[inline(always)] might not be good enough (it being just a hint) and also possibly more expensive than necessary (something needs to do the inlining still).

This is not true for `opt-level=1`, AFAIK. In this mode, LLVM runs [always-inline](https://llvm.org/docs/Passes.html#always-inline-inliner-for-always-inline-functions) which inlines functions without much thinking.
In higher optimization levels this pass doesn't run and `inline(always)` becomes hint.

AngelicosPhosphoros:
Example of prevented optimization.
[...]

PR #87150, opened 2021-07-15
Make wrapping_neg() use wrapping_sub(), #[inline(always)]

hkratz:
This is a follow-up change to the fix for #75598. It simplifies the implementation of wrapping_neg() for all integer types by just calling 0.wrapping_sub(self) and always inlines it. This leads to much less assembly code being emitted for opt-level≤1 and thus much better performance for debug-compiled code.

Background is [this discussion on the internals forum](https://internals.rust-lang.org/t/why-does-rust-generate-10x-as-much-unoptimized-assembly-as-gcc/14930).


JohnTitor (approved):
The perf results seem acceptable. The change makes sense to me, +1.

hkratz:
@yaahc Could you please have a look?

Issue #86431, opened 2021-06-18
Investigate performance impact of debuginfo typename generation


michaelwoerister:
The [code generating type names for debuginfo](https://github.com/rust-lang/rust/blob/master/compiler/rustc_codegen_ssa/src/debuginfo/type_names.rs) does not seem to do any kind of caching of repeatedly occurring string segments (e.g. path prefixes and type renderings). Given that type names can get very long when generics are involved it's worth investigating how much time the compiler spends doing redundant work here.

jyn514:
https://github.com/rust-lang/rust/issues/84873#issuecomment-832860003 is also somewhat related, the names getting passed into llvm can get very long, especially for generators

Sl1mb0:
I'd like to tackle this. I'm curious where th

Sl1mb0:
Posted this in Zulip but I'll post it here too:

- What is meant by the term "type renderings"? Does this refer to matching `TyKind` to it's string literal value e.g. `ty::Bool => output.push_str("bool")` ? or does it refer to pushing the type's identifier onto the output string? Just want to be sure.

- As for how to approach caching repeat string segments, my current idea is to use a wrapper to a hash map that uses a `TyKind` as a key to a string literal e.g. `<TyKind, &str>`. This wrapper will also implement the   `push_debuginfo_type_name()`and `compute_debuginfo_type()` functions. The caching will occur in `push` when we have matched to a given type. Just check if the type is in the cache and if not, load it. Then use it's key to push the string literal onto the output string.

- Part of this issue means analyzing redundant work that the compiler is currently doing. I am wondering if there is a "best approach" to this. As so far my only idea is to profile the compiler before I make changes, and after I make changes, and observe any differences. I don't think this will yield an accurate assessment however, and am curious if anyone would suggest something different.

michaelwoerister:
https://github.com/rust-lang/rust/pull/88898 implemented caching as an experiment but it didn't really improve performance. We concluded that we'd rather not have the additional implementation complexity.

PR #87153, opened 2021-07-15
[debuginfo] Emit associated type bindings in trait object type names.

### LINKED ISSUE #86134: Debuginfo type names for `dyn` types are incomplete
The [current implementation](https://github.com/rust-lang/rust/blob/a50d72158e08e02cfc051b863017bdbd2c45b637/compiler/rustc_codegen_ssa/src/debuginfo/type_names.rs#L133-L142) does not handle associated type bindings and auto-traits which leads to information about the type being lost. For example, `&dyn Iterator<Item=u32> + Send` will be mapped to `&dyn Iterator` and `&dyn Send + Sync` will be mapped to `& dyn '_`.

Some consumers of debuginfo (e.g. NatVis) rely on these type names to be unambiguous identifiers of the type they represent so we must not lose information about assoc type bindings and auto-traits.

michaelwoerister:
This PR updates debuginfo type name generation for trait objects to include associated type bindings and auto trait bounds -- so that, for example, the debuginfo type name of `&dyn Iterator<Item=Foo>` and `&dyn Iterator<Item=Bar>` don't both map to just `&dyn Iterator` anymore.

The following table shows examples of debuginfo type names before and after the PR:
| type | before |  after |
|------|---------|-------|
| `&dyn Iterator<Item=u32>>` | `&dyn Iterator` | `&dyn Iterator<Item=u32>` |
| `&(dyn Iterator<Item=u32>> + Sync)` | `&dyn Iterator` | `&(dyn Iterator<Item=u32> + Sync)` |
| `&(dyn SomeTrait<bool, i8, Bar=u32>> + Send)` | `&dyn SomeTrait<bool, i8>` | `&(dyn SomeTrait<bool, i8, Bar=u32>> + Send)`  |

For targets that need C++-like type names, we use `assoc$<Item,u32>` instead of `Item=u32`:
| type | before |  after |
|------|---------|-------|
| `&dyn Iterator<Item=u32>>` | `ref$<dyn$<Iterator> >` | `ref$<dyn$<Iterator<assoc$<Item,u32> > > >` |
| `&(dyn Iterator<Item=u32>> + Sync)` | `ref$<dyn$<Iterator> >` | `ref$<dyn$<Iterator<assoc$<Item,u32> >,Sync> >` |
| `&(dyn SomeTrait<bool, i8, Bar=u32>> + Send)` | `ref$<dyn$<SomeTrait<bool, i8> > >` | `ref$<dyn$<SomeTrait<bool,i8,assoc$<Bar,u32> > >,Send> >`  |

The PR also adds self-profiling measurements for debuginfo type name generation (re. https://github.com/rust-lang/rust/issues/86431). It looks like the compiler spends up to 0.5% of its time in that task, so the potential for optimizing it via caching seems limited.

However, the perf run also shows [the biggest regression](https://perf.rust-lang.org/detailed-query.html?commit=585e91c718b0b2c5319e1fffd0ff1e62aaf7ccc2&base_commit=b9197978a90be6f7570741eabe2da175fec75375&benchmark=tokio-webpush-simple-debug&run_name=incr-unchanged) in a test case that does not even invoke the code in question. This suggests that the length of the names we generate here can affect performance by influencing how much data the linker has to copy around. 

Fixes https://github.com/rust-lang/rust/issues/86134.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3734,6 +3734,7 @@ dependencies = [
  "rustc_span",
  "rustc_symbol_mangling",
  "rustc_target",
+ "smallvec",
  "tempfile",
  "tracing",
 ]
--- a/compiler/rustc_codegen_ssa/Cargo.toml
+++ b/compiler/rustc_codegen_ssa/Cargo.toml
@@ -16,6 +16,7 @@ libc = "0.2.50"
 jobserver = "0.1.22"
 tempfile = "3.2"
 pathdiff = "0.2.0"
+smallvec = { version = "1.6.1", features = ["union", "may_dangle"] }
 
 rustc_serialize = { path = "../rustc_serialize" }
 rustc_ast = { path = "../rustc_ast" }
--- a/compiler/rustc_codegen_ssa/src/debuginfo/type_names.rs
+++ b/compiler/rustc_codegen_ssa/src/debuginfo/type_names.rs
@@ -19,8 +19,9 @@ use rustc_hir::definitions::{DefPathData, DefPathDataName, DisambiguatedDefPathD
 use rustc_middle::ich::NodeIdHashingMode;
 use rustc_middle::ty::layout::IntegerExt;
 use rustc_middle::ty::subst::{GenericArgKind, SubstsRef};
-use rustc_middle::ty::{self, AdtDef, Ty, TyCtxt};
+use rustc_middle::ty::{self, AdtDef, ExistentialProjection, Ty, TyCtxt};
 use rustc_target::abi::{Integer, TagEncoding, Variants};
+use smallvec::SmallVec;
 
 use std::fmt::Write;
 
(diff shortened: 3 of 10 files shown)


michaelwoerister:
r? @wesleywiser 

@dpaoliello, do the changes here look good to you from an MSVC perspective?

michaelwoerister:
Regarding the performance regression: I noticed that other parts of the compiler don't print generic arguments if they are equal to the default. Doing this for debuginfo type names too (which is sensible, IMO) might give a nice performance boost that makes up for the regression here. Let me know if I should add that to the PR.

wesleywiser (approved):
This looks good to me! If you think omitting default type parameters is pretty easy, I think it would be worth doing in this PR to see if we can improve the performance but if not, I'd be ok landing this as-is.

michaelwoerister:
I'll look into it. For type arguments it should be straightforward.

michaelwoerister:
It looks like omitting default arguments is not as straightforward as I had hoped (I don't understand the relationship between ty::Generics and ty::InternalStubsts well enough). I'll need to leave it for another PR. Let's see if the perf regression persists.

@bors r=wesleywiser

dpaoliello (approved):

michaelwoerister:
Updated test cases. Let's see if I got all of them. My local CDB version is a bit newer than the one on CI and sometimes prints natvis visualizations where the one on CI doesn't.

@bors r=wesleywiser

rylev:
Given the [comments](https://github.com/rust-lang/rust/pull/87153#issuecomment-881428994) on how one might address this performance issue, and the fact that it doesn't seem to have caused performance regressions after merging, I'm going to add the `perf-regression-triaged` label so this is not considered an active performance regression. 

PR #87149, opened 2021-07-15
Rollup of 6 pull requests

JohnTitor:
Successful merges:

 - #86478 (Add -Zfuture-incompat-test to assist with testing future-incompat reports.)
 - #86947 (Move assert_matches to an inner module)
 - #87081 (Add tracking issue number to `wasi_ext`)
 - #87127 (Add safety comments in private core::slice::rotate::ptr_rotate function)
 - #87134 (Make SelfInTyParamDefault wording not be specific to type defaults)
 - #87147 (Update cargo)

Failed merges:

r? @ghost
@rustbot modify labels: rollup

[Create a similar rollup](https://bors.rust-lang.org/queue/rust?prs=86478,86947,87081,87127,87134,87147)

PR #87154, opened 2021-07-15
Fix misuse of rev attribute on <a> tag

GuillaumeGomez:
The `rev` attribute is supposed to talk about "ownership" as far as I could found out. This attribute seems not very well defined in the HTML spec and its usage in rustdoc is suboptimal.

It was found out in https://github.com/rust-lang/rust/pull/87149.

r? @JohnTitor

### DIFF
--- a/src/librustdoc/html/markdown.rs
+++ b/src/librustdoc/html/markdown.rs
@@ -621,8 +621,7 @@ impl<'a, I: Iterator<Item = SpannedEvent<'a>>> Iterator for Footnotes<'a, I> {
                                 is_paragraph = true;
                             }
                             html::push_html(&mut ret, content.into_iter());
-                            write!(ret, "&nbsp;<a href=\"#fnref{}\" rev=\"footnote\">↩</a>", id)
-                                .unwrap();
+                            write!(ret, "&nbsp;<a href=\"#fnref{}\">↩</a>", id).unwrap();
                             if is_paragraph {
                                 ret.push_str("</p>");
                             }


JohnTitor (approved):

GuillaumeGomez:
And of course I forgot to format as usual... :cry:

JohnTitor:
I'm not part of the rustdoc team but this fix seems quite straightforward. But I haven't understood why my rollup triggers it (i.e. previous build didn't trigger it). Could you elaborate? Should we recreate the rollup?

GuillaumeGomez:
I'm honestly not sure either. Something might have been added into the index that wasn't there before, triggering the generation of this link which then fails the HTML check (which has been added a few weeks ago). I didn't look into details but I expect something like this. Well at least it allowed us to spot this issue.

JohnTitor:
Hmm, okay, let's r+ anyway. I assume it's flaky and I'll recreate a rollup if the running one fails. Thanks!
@bors r+

the8472:
> This attribute seems not very well defined in the HTML spec and its usage in rustdoc is suboptimal.

It's [listed](https://www.w3.org/TR/html401/struct/links.html#adef-rev) in the HTML4.01 spec 
But [not in the HTML5 spec](https://html.spec.whatwg.org/multipage/text-level-semantics.html#the-a-element)
[MDN lists it as deprecated](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/a#rev)

PR #87158, opened 2021-07-15
Suggest full enum variant for local modules

In-line:


src/test/ui/suggestions/suggest-full-enum-variant-for-local-module.stderr
```diff
@@ -0,0 +1,16 @@
+error[E0308]: mismatched types
+  --> $DIR/suggest-full-enum-variant-for-local-module.rs:9:28
+   |
+LL |     let _: option::O<()> = ();
+   |            -------------   ^^
+   |            |               |
+   |            |               expected enum `O`, found `()`
+   |            |               help: try using a variant of the expected enum: `option::O::Some(())`
```
In-line, line 8: Applying suggestion without full path leads to compiler error.

estebank:
If we use the short path for suggestions we already provide a follow up suggestion to import the right thing, but I prefer this for now. Surprised that no other test is changed.

@bors r+

PR #87163, opened 2021-07-15
Optimize symbol interning for parallel rustc

bjorn3:
Avoid an exclusive lock when getting the symbol string

### DIFF
--- a/compiler/rustc_span/src/lib.rs
+++ b/compiler/rustc_span/src/lib.rs
@@ -77,7 +77,7 @@ mod tests;
 // threads within the compilation session, but is not accessible outside the
 // session.
 pub struct SessionGlobals {
-    symbol_interner: Lock<symbol::Interner>,
+    symbol_interner: symbol::Interner,
     span_interner: Lock<span_encoding::SpanInterner>,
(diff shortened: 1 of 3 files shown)


bjorn3:
Up to 0.4% regression on non-parallel rustc. Likely due to the extra lock. Let's see if parallel rustc performance improves over not making this change.

@bors try @rust-timer queue

bjorn3:
If anything this seems to be a perf regression compared to the previous parallel rustc benchmark.

this bench: https://perf.rust-lang.org/compare.html?start=26366828a40403c83d6cc29bf5614a6e0388354c&end=8e819e901bee81666d28b5275ec8594bdc75619b&stat=cpu-clock
previous bench: https://perf.rust-lang.org/compare.html?start=2218520b8adf8b8e64b817537d9eb0a84840e2f1&end=7cd4a2ec5195e5df4cb2f7292aa53fa3d870cd9f&stat=cpu-clock

(see https://github.com/rust-lang/rustc-perf/issues/915 for a workaround for a bug preventing old benchmark to be shown)

rylev:
The issue with rustc perf should be resolved

PR #87166, opened 2021-07-15
Show discriminant before overflow in diagnostic for duplicate values.

de-vri-es:
This PR adds the value before overflow for explicit discriminant values in the error for duplicate discriminant values. 
I found it rather confusing to see only the overflowed value.

It only does this for literals, since overflows in const evaluated arithmetic are already a hard error.

This is my first PR to the compiler, so please let me know if the implementation can be improved :)

Before:
image

After:
image

### DIFF
--- a/compiler/rustc_typeck/src/check/check.rs
+++ b/compiler/rustc_typeck/src/check/check.rs
@@ -1472,15 +1472,17 @@ fn check_enum<'tcx>(
                 Some(ref expr) => tcx.hir().span(expr.hir_id),
                 None => v.span,
             };
+            let display_discr = display_discriminant_value(tcx, v, discr.val);
+            let display_discr_i = display_discriminant_value(tcx, variant_i, disr_vals[i].val);
             struct_span_err!(
                 tcx.sess,
                 span,
                 E0081,
                 "discriminant value `{}` already exists",
-                disr_vals[i]
+                discr.val,
             )
-            .span_label(i_span, format!("first use of `{}`", disr_vals[i]))
-            .span_label(span, format!("enum already has `{}`", disr_vals[i]))
+            .span_label(i_span, format!("first use of {}", display_discr_i))
+            .span_label(span, format!("enum already has {}", display_discr))
             .emit();
         }
         disr_vals.push(discr);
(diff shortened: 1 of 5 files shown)


compiler/rustc_typeck/src/check/check.rs
```diff
@@ -1497,6 +1499,25 @@ fn check_enum<'tcx>(
...
+        let body = &tcx.hir().body(expr.body).value;
+        if let hir::ExprKind::Lit(lit) = &body.kind {
+            if let rustc_ast::LitKind::Int(lit_value, _int_kind) = &lit.node {
+                if evaluated != *lit_value {
+                    return format!("{} (overflowed from {})", evaluated, lit_value);
+                }
+            }
+        }
+    }
+    format!("{}", evaluated)
```
de-vri-es, line 1518: Note: I also removed the backticks around the number. They seemed unnecessary to me: the value is always a decimal number, which already stands out from normal text quite well. But I can revert this if desired.

jackh726 (changes requested):
So, I'd like to actually see a test where the *first* discriminant overflows (like in the OP). And I think that the backticks around the value should remain (backticks generally specify a value or code).

And finally (this one is a bit more opinionated): I feel like the exact number here isn't super helpful. It might be better to say some like "overflowed from `usize`"

de-vri-es:
> So, I'd like to actually see a test where the _first_ discriminant overflows (like in the OP). And I think that the backticks around the value should remain (backticks generally specify a value or code).

Makes sense. I added another test case in `src/test/ui/error-codes/E0081.rs`. Let me know if the test case should be in a new file instead, I'll move it.

I added the backticks back, and I removed the overflow value from the main error message, leaving it only in the notes. With that, it now looks like this:

image


> And finally (this one is a bit more opinionated): I feel like the exact number here isn't super helpful. It might be better to say some like "overflowed from `usize`"

Hmm, I actually like to see the value before overflow. It lets me know that neither me nor the compiler have gone mad and we're both really talking about the same number.

jackh726:
Sorry for the wait here. @bors r+ rollup

de-vri-es:
Thanks for the review! :)

Issue #87059, opened 2021-07-11
fix screen reader accessibility of rustdoc item documentation


ahicks92:
I opened this as one issue because the overarching problem is the problem, if you will.  if I should have made a bunch, let me know.

Looks like the following applies to beta.  Example page: https://doc.rust-lang.org/beta/std/collections/struct.BTreeMap.html

I'm a screen reader user (e.g. blind).  Specifically NVDA.  Whatever changes are coming down the pipeline with Rustdoc make the above-linked page difficult to navigate.  It used to be the case that you could get used to the weird heading structure (specifically that Rustdoc headings are higher levels than example/error headings) then be mostly fine.  But now:

- A lot of what used to be headings are exposed as buttons.
- All of those buttons like to start with `[-]`, which screen readers are forced to listen to as "left bracket dash right bracket", e.g. "left bracket dash right bracket pub fn".
- There's a table of contents, but it only shows up if the page is maximized, and if there's a button to expand it I can't quickly find it.
- Said table of contents provides no way to quickly navigate it, e.g. "methods" being a heading, using lists, etc.  I have to ctrl+f the categories.

I manage because I'm a screen reader power user who has previous familiarity with these pages, which meant that mostly getting around/through this was working out what actually changed rather than how to do it from scratch.  It still hit my efficiency quite a bit.  Most screen reader users aren't going to know to maximize their browser and ctrl+f "methods" and so on, or work out that actually sometimes you need next heading and sometimes you need next button.

Some concrete fixes:

- Put the `h` tags back (maybe we can do this with aria, if the buttons are important).
- move `[-]` to the end of the line, where it used to be. I think this can be faked with css if we need the visual representation to stay the same, but I'm a backend guy.
- Change the "headings" in the table of contents to be actual h tags, perhaps with nested links, or make the table of contents nested lists like mdbook.
- make sure that the table of contents is always visible, or that the button to expand/collapse it is more obvious if there is one.

I'm happy to discuss other fixes as well; my goal is to provide some sort of conversation starter.

jyn514:
cc @jsha

jyn514:
@ahicks92 when you say "coming down the pipeline" you specifically mean that these are changes in beta, that haven't made it to stable, right?

jyn514:
[...]

PR #87167, opened 2021-07-15
Fix sidebar display on small devices

GuillaumeGomez:
Part of #87059.

Instead of hiding the sidebar on small devices, we instead move it out of the viewport so that it remains "visible" to our text only users.

Could you confirm it works for you @ahicks92 and @DataTriny please? You can give it a try at [this URL](https://guillaume-gomez.fr/rustdoc-test/test_docs/index.html).

r? @notriddle

### DIFF
--- a/src/librustdoc/html/static/css/rustdoc.css
+++ b/src/librustdoc/html/static/css/rustdoc.css
@@ -1631,15 +1631,18 @@ details.undocumented[open] > summary::before {
 		display: none;
 	}
 
+	/* We do NOT hide this element so that alternative device readers still have this information
+	   available. */
 	.sidebar-elems {
 		position: fixed;
 		z-index: 1;
-		left: 0;
 		top: 45px;
 		bottom: 0;
+		width: 246px;
+		/* We move the sidebar to the left by its own width so it doesn't appear. */
+		left: -246px;
 		overflow-y: auto;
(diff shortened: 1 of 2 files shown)


notriddle:
We also want to make sure this works well for sighted people using the keyboard to navigate.

The obvious way to do that is to have JavaScript detect when focus goes inside the sidebar, and make the sidebar visible when it does.

GuillaumeGomez:
I don't expect sighted people to have a screen small enough to get the sidebar moved out of the viewport.

notriddle:
B37A8FEA-ED70-4CFC-AA65-C720E645934A

GuillaumeGomez:
😮 Ok nevermind. Please open an issue so I can do it next. 🤣

notriddle:
@GuillaumeGomez https://github.com/rust-lang/rust/issues/87172

notriddle:
Other than that concern, this seems fine to me.

ahicks92:
NVDA is claiming it's at -19% from the left of the screen, so going with yes.

notriddle (approved):

GuillaumeGomez:
@ahicks92 Awesome, thanks!

Once this PR is merged, I'll write the PR about the focus.

@bors: r=notriddle

PR #48544, opened 2018-02-26
Produce better size_hints from Flatten & FlatMap

scottmcm:
Adds `const SIZE_HINT` to IntoIterator, allowing a good estimate before the actual iterators are available.

These two adapters typically produced [largely-unhelpful](https://play.rust-lang.org/?gist=65ffd3ec820822360f3b6f29361a220a&version=nightly) `size_hint`s, as they had no idea what was going to come out of the inner iterators.  This change lets `flat_map` do as well as `filter_map` with `Option`s, and allows `flatten` to get its `size_hint` exactly right in cases like `U = &[T; N]`.

PR #87168, opened 2021-07-15
implement TrustedLen for Flatten/FlatMap if the U: IntoIterator == [T; N]

### LINKED ISSUE #87094: iter::Flatten, FlatMap should return more precise size_hint and be TrustedLen when item type is an array
If I understand correctly, presence of `TrustedLen` trait allows `.collect::<Vec<_>>()` to avoid reallocations, right?

In that case, given the following:
```rs
    #[derive(Clone, Copy)]
    pub struct Pixel(u8, u8, u8, u8);

    let pixels: Vec<Pixel> = vec![Pixel(1, 2, 3, 4); 10];
    let iter = pixels
        .iter()
        .flat_map(|p| [p.0, p.1, p.2, p.3]); // or: .map(|p| [p.0, p.1, p.2, p.3]).flatten();
    println!("{:?}", iter.size_hint());

    let v: Vec<u8> = iter.collect();
 ```
Since the type I map to has known const size (`[u8; 4]`), size_hint should be able to precisely return `(40, Some(40))`, and the iterator specialization should implement TrustedLen.
Currently, it returns `(0, None)`. This means that to implement this function efficiently, I'd need to manually do `Vec::with_capacity(pixels.len() * 4)`.

### PR DESCRIPTION
**Pull request by the8472**
This only works if arrays are passed directly instead of array iterators
because we need to be sure that they have not been advanced before
Flatten does its size calculation.

resolves #87094

### DIFF
--- a/library/core/src/iter/adapters/flatten.rs
+++ b/library/core/src/iter/adapters/flatten.rs
@@ -297,7 +297,7 @@ where
         let (blo, bhi) = self.backiter.as_ref().map_or((0, Some(0)), U::size_hint);
         let lo = flo.saturating_add(blo);
 
-        if let Some(fixed_size) = <<I as Iterator>::Item as ConstSizeIterable>::size() {
+        if let Some(fixed_size) = <<I as Iterator>::Item as ConstSizeIntoIterator>::size() {
             let (lower, upper) = self.iter.size_hint();
 
             let lower = lower.saturating_mul(fixed_size).saturating_add(lo);
@@ -474,18 +474,18 @@ where
     }
 }
 
-trait ConstSizeIterable {
+trait ConstSizeIntoIterator: IntoIterator {
     fn size() -> Option<usize>;
 }
 
-impl<T> ConstSizeIterable for T {
+impl<T> ConstSizeIntoIterator for T where T: IntoIterator {
     #[inline]
     default fn size() -> Option<usize> {
         None
     }
 }
 
-impl<T, const N: usize> ConstSizeIterable for [T; N] {
+impl<T, const N: usize> ConstSizeIntoIterator for [T; N] {
     #[inline]
     fn size() -> Option<usize> {
         Some(N)
--- a/library/core/src/iter/adapters/flatten.rs
+++ b/library/core/src/iter/adapters/flatten.rs
@@ -319,12 +319,8 @@ where
             let (lower, upper) = self.iter.size_hint();
 
             let lower = lower.saturating_mul(fixed_size).saturating_add(lo);
-            let upper = upper.and_then(|i| i.checked_mul(fixed_size));
-            let upper = fhi
-                .zip_with(bhi, usize::checked_add)
-                .flatten()
-                .zip_with(upper, usize::checked_add)
-                .flatten();
+            let upper =
+                try { fhi?.checked_add(bhi?)?.checked_add(fixed_size.checked_mul(upper?)?)? };
 
             return (lower, upper);
         }
--- a/library/core/src/iter/adapters/flatten.rs
+++ b/library/core/src/iter/adapters/flatten.rs
@@ -1,5 +1,5 @@
 use crate::fmt;
-use crate::iter::{DoubleEndedIterator, Fuse, FusedIterator, Iterator, Map};
+use crate::iter::{DoubleEndedIterator, Fuse, FusedIterator, Iterator, Map, TrustedLen};
 use crate::ops::Try;
 
 /// An iterator that maps each element to an iterator, and yields the elements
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: library/core/src/iter/adapters/flatten.rs [hunk 1]
```diff
@@ -444,3 +473,21 @@ where
         init
     }
 }
+
+trait ConstSizeIterable {
+    fn size() -> Option<usize>;
```
the8472, line 478: Ideally this would be a constant but currently specialization doesn't work for associated constants.

library/core/src/iter/adapters/flatten.rs
```diff
@@ -444,3 +473,21 @@ where
...
+}
+
+impl<T> ConstSizeIterable for T {
+    #[inline]
+    default fn size() -> Option<usize> {
+        None
+    }
+}
+
+impl<T, const N: usize> ConstSizeIterable for [T; N] {
```
SkiFire13, line 488: Maybe `IntoConstSizeIterator` would be a better name? `ConstSizeIterable` sounds like the implementer is already an `Iterator`, but `[T; N]` is not. (technically `Iterable` could be a synonymous of `IntoIterator`, at least in Java it has that meaning, however `InPlaceIterable` already uses it to mean an `Iterator`, so one of them should be changed)

the8472: I was going by the plain english meaning of the *-able* suffix meaning that you can do something with it. Iterable -> it can be iterated, iterator -> it does the iterating. So InPlaceIterable means it *can* be iterated in place, not that it always does that. Also, any Iterator is iterable, so that's a superset. But yes in rust `IntoIterator` already expresses this.

But `IntoConstSizeIterator` is misleading, since it's not the Iterator that's const-sized. I think `ConstSizeIntoIterator` expresses it more clearly instead.

SkiFire13: `ConstSizeIntoIterator` sounds good.

> So InPlaceIterable means it can be iterated in place,

The problem I have is that by going with that logic then `ConstSizeIterable` should also mean it can be somehow be iterated/advanced, however in this case it can only be turned into an iterator. Overall I think we should either properly define what we mean with `Iterable` or just stick with `Iterator`/`IntoIterator`

the8472:
The specialization now covers flattening over references to arrays in addition to owned arrays.

library/core/src/iter/adapters/flatten.rs
```diff
@@ -282,6 +314,21 @@ where
...
+        if let Some(fixed_size) = <<I as Iterator>::Item as ConstSizeIntoIterator>::size() {
+            let (lower, upper) = self.iter.size_hint();
+
+            let lower = lower.saturating_mul(fixed_size).saturating_add(lo);
+            let upper = upper.and_then(|i| i.checked_mul(fixed_size));
+            let upper = fhi
+                .zip_with(bhi, usize::checked_add)
+                .flatten()
+                .zip_with(upper, usize::checked_add)
+                .flatten();
```
scottmcm, line 327: I'm getting a bit lost in all the combinators here.  Could you rewrite it to just a `try { ... }` block?  They're already used in core for iterators in other places, so shouldn't even need a `feature`.

I think this is something like
```rust
let upper = try {
    fhi?.checked_add(bhi?)?.checked_add(i.checked_mul(fixed_size)?)?
};
```

scottmcm:
Thanks for this!  I tried to do something similar back in https://github.com/rust-lang/rust/pull/48544, but doing this with specialization avoids the complaints people had about that.

library/core/src/iter/adapters/flatten.rs
```diff
@@ -444,3 +491,52 @@ where
...
+    fn size() -> Option<usize> {
+        Some(N)
+    }
+}
+
+#[doc(hidden)]
+#[unstable(feature = "std_internals", issue = "none")]
+// FIXME(#20400): Instead of this helper trait there should be multiple impl TrustedLen for Flatten<>
+//   blocks with different bounds on Iterator::Item but the compiler erroneously considers them overlapping
+pub unsafe trait TrustedConstSize: IntoIterator {}
```
scottmcm, line 535: nit: I would have expected something like `TrustedConstSizeIntoIterator : ConstSizeIntoIterator`?  Is there some reason this mentions only `IntoIterator`?

the8472: `ConstSizeIntoIterator` is implemented for all `IntoIterator` anyway but returns a `None` for the non-const ones. So adding that bound wouldn't make a difference.

library/core/src/iter/adapters/flatten.rs
the8472: I'll give it a try.

scottmcm:
Thanks!  Let me just check perf just in case:

@bors try @rust-timer queue

PR #87171, opened 2021-07-15
Remove Option from BufWriter

### LINKED ISSUE #72925: Remove `Option<W>` in `BufWriter`
https://github.com/rust-lang/rust/blob/eeaf497b2a6bc065874e3d3367b1f3023c5bb3d3/src/libstd/io/buffered.rs#L451

`BufWriter` doesn't really need to hold its inner writer as an `Option`; it seems like the only reason it does so is to make it easier to call `into_inner`, here:

https://github.com/rust-lang/rust/blob/eeaf497b2a6bc065874e3d3367b1f3023c5bb3d3/src/libstd/io/buffered.rs#L648-L653

While it's unfortunate to add `unsafe`, there's already some in `buffered.rs`, and it seems like the better choice here would be to use `ptr::read` and `mem::forget`, especially because this method is probably infrequently used, and making this change would remove the numerous `unwrap`s that litter the `BufWriter` implementation.

Alexendoo:
Fixes #72925

### DIFF
--- a/library/std/src/io/buffered/bufwriter.rs
+++ b/library/std/src/io/buffered/bufwriter.rs
@@ -68,7 +68,7 @@ use crate::ptr;
 /// [`flush`]: BufWriter::flush
 #[stable(feature = "rust1", since = "1.0.0")]
 pub struct BufWriter<W: Write> {
-    inner: Option<W>,
+    inner: W,
     // The buffer. Avoid using this like a normal `Vec` in common code paths.
     // That is, don't use `buf.push`, `buf.extend_from_slice`, or any other
     // methods that require bounds checking or the like. This makes an enormous
@@ -112,7 +112,7 @@ impl<W: Write> BufWriter<W> {
     /// ```
     #[stable(feature = "rust1", since = "1.0.0")]
     pub fn with_capacity(capacity: usize, inner: W) -> BufWriter<W> {
-        BufWriter { inner: Some(inner), buf: Vec::with_capacity(capacity), panicked: false }
+        BufWriter { inner, buf: Vec::with_capacity(capacity), panicked: false }
     }
 
     /// Send data in our local buffer into the inner writer, looping as
@@ -161,10 +161,9 @@ impl<W: Write> BufWriter<W> {
         }
 
         let mut guard = BufGuard::new(&mut self.buf);
-        let inner = self.inner.as_mut().unwrap();
         while !guard.done() {
             self.panicked = true;
-            let r = inner.write(guard.remaining());
+            let r = self.inner.write(guard.remaining());
             self.panicked = false;
 
             match r {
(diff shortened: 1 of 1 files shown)


library/std/src/io/buffered/bufwriter.rs
```diff
@@ -327,9 +327,16 @@ impl<W: Write> BufWriter<W> {
     /// ```
     #[unstable(feature = "bufwriter_into_raw_parts", issue = "80690")]
     pub fn into_raw_parts(mut self) -> (W, Result<Vec<u8>, WriterPanicked>) {
-        let buf = mem::take(&mut self.buf);
-        let buf = if !self.panicked { Ok(buf) } else { Err(WriterPanicked { buf }) };
-        (self.inner.take().unwrap(), buf)
+        // SAFETY: forget(self) prevents double dropping inner/buf
+        unsafe {
+            let inner = ptr::read(&mut self.inner);
+            let buf = ptr::read(&mut self.buf);
```
the8472, line 333: The `mem::take` approach should still work for the buffer, no? It replaces it with an empty iterator which doesn't have any allocation so there's no issue leaking it.

the8472:
Actually, does this need any unsafe at all? [Plain destructuring should work.](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=2d5b246ab96a5e66c955081d4e0d1e67)

Mark-Simulacrum:
No, destructuring by value for types impl'ing drop doesn't work - https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=b5e4339d6b72e25199a3d02d47004108

library/std/src/io/buffered/bufwriter.rs
Alexendoo: It would yeah, I figured since the unsafe is there anyway it would be clear enough to read them both the same way, and saves zeroing out the buf in self


I can use take again if that's preferred though

Alexendoo:
Fixed the merge conflict, I was accidentally 11k commits behind master 😅

library/std/src/io/buffered/bufwriter.rs
```diff
@@ -337,9 +336,16 @@ impl<W: Write> BufWriter<W> {
...
+            let buf = if !self.panicked { Ok(buf) } else { Err(WriterPanicked { buf }) };
+
+            mem::forget(self);
```
SkiFire13, line 345: OT: this makes me wonder if we should have a way to destructure a `struct` that implements `Drop` without running its `drop` implementation (and maybe also a way with running it?). It feels bad having to use `unsafe` for this kind of stuff since it's simple enough that the compiler should be able to guarantee its safety in the general case.

library/std/src/io/buffered/bufwriter.rs
the8472: I assume that the zeroing cost is tiny compared to whatever IO is done with the writer. Using take keeps the unsafe block as small as possible.

You can shrink it further by moving the following lines out of the block, they're not unsafe after all. All that's needed is one `let inner =  unsafe { ptr::read() }`;

PR #84029, opened 2021-04-09
add `track_path::path` fn for usage in `proc_macro`s

drahnr:
Adds a way to declare a dependency on external files without including them, to either re-trigger the build of a file as well as covering the use case of including dependencies within the `rustc` invocation, such that tools like `sccache`/`cachepot` are able to handle references to external files which are not included.

Ref #73921

Issue #74690, opened 2020-07-23
Tracking Issue for tracked access to environment variables from proc macros


petrochenkov:
https://github.com/rust-lang/rust/pull/74653 introduces `proc_macro::tracked_env::var` - a new function that is similar to regular `env::var` called from a proc macro, except that it also adds the accessed variable to depinfo.

This way build system executing the compiler will know that the variable was accessed during compilation, and will be able to rerun the build when the value of that variable changes.

Besides the dependency tracking `tracked_env::var` differs from `env::var` from the standard library by requiring UTF-8 arguments.
See some discussion in https://github.com/rust-lang/rust/pull/71858 for the motivation, non UTF-8 variables were considered unimportant corner cases by the stakeholders, and tracking for them can be introduced later with new depinfo directives like `# env-dep-bytes`.

programmerjake:
This seems really useful, what would it take to stabilize it? Also, this function should be mentioned in the docs for `std`'s environment access functions, since using those from a proc-macro could be considered incorrect.

petrochenkov:
>This seems really useful, what would it take to stabilize it?

Some user experience I guess?
Things are usually stabilized if someone is using them and pushing for them.

programmerjake:
one alternative that might be more useful is to instead add tracking to `std`'s env functions that's only enabled in proc-macros, that way code wouldn't need to change and you can easily catch nearly all env usage. Idk if that's feasible though...

petrochenkov:
This function need to wired into the compiler to affect depinfo files.
In theory it can be added to `std` if `std` gets an analogue of "proc macro bridge" that `proc_macro` uses to talk with the compiler through a stable ABI, although I seriously doubt that it would be a good idea.

petrochenkov:
In https://github.com/rust-lang/rust/pull/84029#discussion_r622603746 @Xanewok has some concerns about the module structure.
It should be somehow harmonized with a similar interface for tracking files (https://github.com/rust-lang/rust/pull/84029).

m-ou-se:
Closing this in favor of this tracking issue, tracking both `path` and `env`: https://github.com/rust-lang/rust/issues/99515

PR #87173, opened 2021-07-15
unstable proc_macro tracked::* rename/restructure

drahnr:
Extracted restructure of the unstable `tracked::{fs,env}::*`

Originally part of https://github.com/rust-lang/rust/pull/84029

It's slightly clunky to to have three distinct unstable features, since one cannot use two annotations on one `mod tracked {..}` declaration.

The second thing I am unsure is `rust-analyzer` which would need some adaption for the new structure.

CC @Xanewok

### DIFF
--- a/library/proc_macro/src/lib.rs
+++ b/library/proc_macro/src/lib.rs
@@ -1489,9 +1489,34 @@ impl fmt::Debug for Literal {
     }
 }
 
-/// Tracked access to environment variables.
-#[unstable(feature = "proc_macro_tracked_env", issue = "99515")]
-pub mod tracked_env {
+#[unstable(feature = "proc_macro_tracked_env", issue = "74690")]
+/// Tracked access to env and path.
+pub mod tracked {
+    #[unstable(feature = "proc_macro_tracked_path", issue = "73921")]
+    use std::path::Path;
+
+    /// Track a file as if it was a dependency.
+    ///
+    /// The file is located relative to the current file where the proc-macro
+    /// is used (similarly to how modules are found). The provided path is
+    /// interpreted in a platform-specific way at compile time. So, for
+    /// instance, an invocation with a Windows path
+    /// containing backslashes `\` would not compile correctly on Unix.
+    ///
+    /// Errors if the provided `Path` cannot be encoded as a `str`
+    ///
+    /// Commonly used for tracking asset preprocessing.
+    #[unstable(feature = "proc_macro_tracked_path", issue = "73921")]
+    pub fn path<P: AsRef<Path>>(path: P) -> Result<(), ()> {
+        let path: &Path = path.as_ref();
+        if let Some(path) = path.to_str() {
+            crate::bridge::client::FreeFunctions::track_fs_path(path);
+            Ok(())
+        } else {
+            Err(())
+        }
+    }
+
     use std::env::{self, VarError};
     use std::ffi::OsStr;
 
@@ -1500,25 +1525,11 @@ pub mod tracked_env {
     /// compilation, and will be able to rerun the build when the value of that variable changes.
     /// Besides the dependency tracking this function should be equivalent to `env::var` from the
     /// standard library, except that the argument must be UTF-8.
-    #[unstable(feature = "proc_macro_tracked_env", issue = "99515")]
-    pub fn var<K: AsRef<OsStr> + AsRef<str>>(key: K) -> Result<String, VarError> {
+    #[unstable(feature = "proc_macro_tracked_env", issue = "74690")]
+    pub fn env_var<K: AsRef<OsStr> + AsRef<str>>(key: K) -> Result<String, VarError> {
         let key: &str = key.as_ref();
         let value = env::var(key);
         crate::bridge::client::FreeFunctions::track_env_var(key, value.as_deref().ok());
         value
-    }
-}
-
-/// Tracked access to additional files.
-#[unstable(feature = "track_path", issue = "99515")]
-pub mod tracked_path {
-
-    /// Track a file explicitly.
-    ///
-    /// Commonly used for tracking asset preprocessing.
-    #[unstable(feature = "track_path", issue = "99515")]
-    pub fn path<P: AsRef<str>>(path: P) {
-        let path: &str = path.as_ref();
-        crate::bridge::client::FreeFunctions::track_path(path);
     }
 }
--- a/library/proc_macro/src/lib.rs
+++ b/library/proc_macro/src/lib.rs
@@ -1507,18 +1507,14 @@ pub mod tracked {
     ///
     /// Commonly used for tracking asset preprocessing.
     #[unstable(feature = "proc_macro_tracked_path", issue = "73921")]
-    pub fn path<P: AsRef<Path>>(path: P) -> Result<(), ()> {
+    pub fn path<P: AsRef<Path>>(path: P) {
         let path: &Path = path.as_ref();
-        if let Some(path) = path.to_str() {
-            crate::bridge::client::FreeFunctions::track_path(path);
-            Ok(())
-        } else {
-            Err(())
-        }
+        crate::bridge::client::FreeFunctions::track_path(PathBuf::from(path));
     }
 
     use std::env::{self, VarError};
     use std::ffi::OsStr;
+    use std::path::PathBuf;
 
     /// Retrieve an environment variable and add it to build dependency info.
     /// The build system executing the compiler will know that the variable was accessed during
--- a/library/proc_macro/src/lib.rs
+++ b/library/proc_macro/src/lib.rs
@@ -1490,10 +1490,29 @@ impl fmt::Debug for Literal {
 }
 
 #[unstable(feature = "proc_macro_tracked_env", issue = "74690")]
-/// Tracked access to env and path.
-pub mod tracked {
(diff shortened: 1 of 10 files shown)


src/test/run-make/track-path-dep-info/macro_def.rs
```diff
@@ -1,11 +1,11 @@
-#![feature(track_path)]
+#![feature(proc_macro_tracked,proc_macro_tracked_path)]
```
drahnr, line 1:
```suggestion
#![feature(proc_macro_tracked,proc_macro_tracked_fs_path)]
```

petrochenkov:
I'm going to transfer this to the libs API team.
These are APIs that allow to trigger a rebuild when a certain environment variable or a file used from a proc macro change.

I'm not sure what the current library conventions are, but seeing >2 segment paths like `tracked::fs::path` in code is uncommon, so the new structure will likely result in the API being used through renamed imports like
```rust
use tracked::fs::path as track_path;
use tracked::env::var as tracked_env_var;

fn my_proc_macro() {
    track_path("my_file.txt"); // returns nothing
    let var = tracked_env_var("MY_VAR"); // returns the variable's value
}
```

drahnr:
Totally understandable, and I don't have a strong opinion, mostly going off of what was discussed in the referenced #84029 issue. If this does not meet the requirements feel free to close.

joshtriplett:
We discussed this in today's @rust-lang/libs-api meeting. We don't want to see an extra level of submodule nesting here; we should just have `tracked::path` and `tracked::env_var`. We didn't find the parallel with `std` compelling, because this will never have any significant fraction of `std`, and of the two methods we already have, one (`path`) already has no parallel in `std`.

Also, separate from this issue, several people in the meeting felt that we shouldn't restrict `path` to just `str`. We understand that the metadata format currently can't handle non-UTF-8, but we'd still like to see this accept a path, to avoid hardcoding that limitation in the API. We can always internally raise an error if it's not UTF-8 and the metadata can't handle it.

pksunkara:
> so the new structure will likely result in the API being used through renamed imports like

Shouldn't we name the functions `track_path` and `track_env_var` then?

drahnr:
> > so the new structure will likely result in the API being used through renamed imports like
> 
> Shouldn't we name the functions `track_path` and `track_env_var` then?

I'd prefer the one module level variants, which is in line with other macro impls iirc: 

> .. we should just have `tracked::path` and `tracked::env_var` ..

drahnr:
> Also, separate from this issue, several people in the meeting felt that we shouldn't restrict `path` to just `str`. We understand that the metadata format currently can't handle non-UTF-8, but we'd still like to see this accept a path, to avoid hardcoding that limitation in the API. We can always internally raise an error if it's not UTF-8 and the metadata can't handle it.

What's meant by `internally raising an error`, where should this be handled? Is `path.to_str().expect()` good enough? `Encode` does not provide means to return a result type. I'd appreciate some input.

drahnr:
Pending access to something that can be cast to a byte slice, the impl currently accepts valid UTF-8 only. I am not sure the error handling is as desired.

petrochenkov:
It's not necessary to return a `Result`, proc macro server can simply report a compiler error if the passed `Path` is non-UTF8.

I'm not sure whether proc macro bridge can support passing `Path` though.
If it cannot, *then* we could actually determine non-UTF8-ness on the client side and return a `Result`.

drahnr:
> It's not necessary to return a `Result`, proc macro server can simply report a compiler error if the passed `Path` is non-UTF8.
> 
> I'm not sure whether proc macro bridge can support passing `Path` though.
> If it cannot, _then_ we could actually determine non-UTF8-ness on the client side and return a `Result`.

Technically this would be possible, but that would require to access the currently private `fn as_u8_slice` and the `unsafe fn from_u8_slice` to be usable from the `encode`/`decode` impl.

I added an implementation adding a `Result` at the outer level, avoiding the `encode` and `decode` impls for `PathBuf`.

JohnCSimon:
Ping from triage: 
@drahnr - can you please post your status on this PR?

petrochenkov:
@drahnr 
Could you move the string -> path change to a separate PR and assign that PR to me?
It seems entirely orthogonal to the module restructuring.

I'm going to unassign myself from this PR, the module restructuring part is entirely a library team matter.
(I only want to remind about squashing commits before merging.)

drahnr:
Hey, I'd appreciate some feedback or remaining todos :)

JohnCSimon:
> @JohnCSimon JohnCSimon added S-waiting-on-review and removed S-waiting-on-review labels 28 days ago

@drahnr - removing and re-adding the tag just bumps the "Last updated on..." date. This is just part of the triage process. It's still in `S-Waiting-On-Review`  state so it's still standing by for review.

JohnCSimon:
triage: 
@m-ou-se ?

jplatte:
Ping @m-ou-se

Dylan-DPC:
@drahnr if you can resolve the conflicts we can get look at getting this reviewed after that

drahnr:
I rebased it, sorry for the delay

drahnr:
Resolved the conflicts, I'd appreciate a review :) rust-analyzer will need an update eventually to match, but I'd prefer to defer to

m-ou-se:
I've opened https://github.com/rust-lang/rust/issues/99515 to track this feature and the steps that still need to be done.

Dylan-DPC:
@drahnr any updates on this pr?

drahnr:
> @drahnr any updates on this pr?

It did appear to me that there was much interest in landing the changeset, and it still is somewhat nebulous to me who'd have to weigh in on the design of the public API on as referenced in https://github.com/rust-lang/rust/issues/99515 .

I'd be happy to pick this up again though.

Dylan-DPC:
@drahnr any updates on this?

drahnr:
Back on it.

rdrpenguin04:
What's left on this?

drahnr:
> What's left on this?

A review, and another rebase.

library/proc_macro/src/lib.rs
```diff
@@ -1489,8 +1489,8 @@ impl fmt::Debug for Literal {
     }
 }
 
-/// Tracked access to environment variables.
-#[unstable(feature = "proc_macro_tracked_env", issue = "99515")]
+#[unstable(feature = "proc_macro_tracked_env", issue = "74690")]
```
m-ou-se, line 1492: Why is the tracking issue changed? #74690 is a closed issue.

library/proc_macro/src/lib.rs
```diff
@@ -1510,15 +1510,23 @@ pub mod tracked_env {
...
+    use std::path::{Path, PathBuf};
 
-    /// Track a file explicitly.
+    /// Track a file as if it was a dependency.
+    ///
+    /// The file is located relative to the current file where the proc-macro
+    /// is used (similarly to how modules are found). The provided path is
+    /// interpreted in a platform-specific way at compile time. So, for
+    /// instance, an invocation with a Windows path
+    /// containing backslashes `\` would not compile correctly on Unix.
```
m-ou-se, line 1523: I find the note on windows/unix paths quite confusing. Who is that note for? What user error is it trying to prevent?

library/proc_macro/src/lib.rs
```diff
@@ -1510,15 +1510,23 @@ pub mod tracked_env {
...
+#[unstable(feature = "proc_macro_tracked_path", issue = "99515")]
 pub mod tracked_path {
+    use std::path::{Path, PathBuf};
 
-    /// Track a file explicitly.
+    /// Track a file as if it was a dependency.
+    ///
+    /// The file is located relative to the current file where the proc-macro
+    /// is used (similarly to how modules are found). The provided path is
+    /// interpreted in a platform-specific way at compile time. So, for
```
m-ou-se, line 1521: Is it not relative to the current working directory? So if I do `File::open("a.txt")` and `tracked_path::path("a.txt")`, can that refer to a different file?

m-ou-se (commented):
This PR does a few different things at once. One of them is changing the &str type to a PathBuf. That change was requested [here](https://github.com/rust-lang/rust/pull/87173#issuecomment-892922762) and would be fine to merge. The other changes seem unrelated and it's unclear why some of them were made. (E.g. changing the tracking issue, or changing "environment" to "env" in a comment, or reordering some attributes/doc comments, etc.)

library/proc_macro/src/lib.rs
bjorn3: If I understand the docs correct, this would behave similar to using the `include!()` family in the macro expansion, which indeed uses a different directory to be relative to than `File::open()` inside the proc macro. The behavior of `include!()` is almost certainly what you want. The current working directory that rustc is invoked with when using cargo is effectively undefined.

alex-semenyuk:
@drahnr
From wg-triage. Could you please take a look comments and make rebase?

Dylan-DPC:
Closing this as inactive. Feel free to reöpen this pr or create a new pr if you get the time to work on this. Thanks

Skgland:
Would it be alright to pick up the "uncontroversial" changes from this i.e. restructuring/renaming and changing AsRef<str> to AsRef<Path>? That appear to be the next not blocked step to me.

PR #87176, opened 2021-07-16
Improve documentation for `Vec::truncate`, `Vec::split_off`, `Vec::drain`

Legend-of-iPhoenix:
I believe that referencing `Vec::split_off` (instead of `Vec::drain`) in the `Vec::truncate` documentation is more helpful here:

- `Vec::split_off` references `Vec::truncate` through a `#[must_use]`.
- `Vec::drain` requires a `collect` to function similarly because it returns a `Drain` iterator
- The docs for `Vec::drain` do not include information on how to use it exactly like `Vec::truncate`

### DIFF
--- a/library/alloc/src/vec/mod.rs
+++ b/library/alloc/src/vec/mod.rs
@@ -1001,7 +1001,7 @@ impl<T, A: Allocator> Vec<T, A> {
     /// If `len` is greater than the vector's current length, this has no
     /// effect.
     ///
-    /// The [`split_off`] is similar to `truncate`, but causes the excess
+    /// The [`split_off`] method is similar to `truncate`, but causes the excess
     /// elements to be returned instead of dropped.
     ///
     /// Note that this method has no effect on the allocated capacity
--- a/library/alloc/src/vec/mod.rs
+++ b/library/alloc/src/vec/mod.rs
@@ -1001,11 +1001,11 @@ impl<T, A: Allocator> Vec<T, A> {
     /// If `len` is greater than the vector's current length, this has no
     /// effect.
     ///
-    /// The [`split_off`] method is similar to `truncate`, but causes the excess
-    /// elements to be returned instead of dropped.
-    ///
     /// Note that this method has no effect on the allocated capacity
     /// of the vector.
+    ///
+    /// The [`split_off`] method is similar to `truncate`, but causes the
+    /// excess elements to be returned instead of dropped.
     ///
     /// # Examples
     ///
@@ -1017,8 +1017,8 @@ impl<T, A: Allocator> Vec<T, A> {
     /// assert_eq!(vec, [1, 2]);
     /// ```
     ///
-    /// No truncation occurs when `len` is greater than the vector's current
-    /// length:
+    /// No truncation or panic occurs when `len` is greater than the vector's
+    /// current length:
     ///
     /// ```
     /// let mut vec = vec![1, 2, 3];
@@ -1835,12 +1835,15 @@ impl<T, A: Allocator> Vec<T, A> {
         self.len() == 0
     }
 
-    /// Splits the collection into two at the given index.
+    /// Splits the collection in two at the given index.
     ///
     /// Returns a newly allocated vector containing the elements in the range
     /// `[at, len)`. After the call, the original vector will be left containing
     /// the elements `[0, at)` with its previous capacity unchanged.
     ///
+    /// If you do not need to own the tail `Vec`, use [`Vec::drain`] to avoid
+    /// an extra allocation.
+    /// 
     /// # Panics
     ///
     /// Panics if `at > len`.
--- a/library/alloc/src/vec/mod.rs
+++ b/library/alloc/src/vec/mod.rs
@@ -1005,7 +1005,9 @@ impl<T, A: Allocator> Vec<T, A> {
     /// of the vector.
     ///
     /// The [`split_off`] method is similar to `truncate`, but causes the
-    /// excess elements to be returned instead of dropped.
+    /// tail elements to be returned as a newly-allocated `Vec` instead
+    /// of being dropped. Note that [`Vec::drain`] can emulate `split_off`
+    /// and avoids this extra allocation by returning an iterator.
     ///
     /// # Examples
     ///
--- a/library/alloc/src/vec/mod.rs
+++ b/library/alloc/src/vec/mod.rs
@@ -1843,7 +1843,7 @@ impl<T, A: Allocator> Vec<T, A> {
     ///
     /// If you do not need to own the tail `Vec`, use [`Vec::drain`] to avoid
     /// an extra allocation.
-    /// 
+    ///
     /// # Panics
     ///
     /// Panics if `at > len`.
--- a/library/alloc/src/vec/mod.rs
+++ b/library/alloc/src/vec/mod.rs
@@ -1001,7 +1001,7 @@ impl<T, A: Allocator> Vec<T, A> {
     /// If `len` is greater than the vector's current length, this has no
     /// effect.
     ///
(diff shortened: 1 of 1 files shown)


library/alloc/src/vec/mod.rs
```diff
@@ -1001,7 +1001,7 @@ impl<T, A: Allocator> Vec<T, A> {
     /// If `len` is greater than the vector's current length, this has no
     /// effect.
     ///
-    /// The [`drain`] method can emulate `truncate`, but causes the excess
+    /// The [`split_off`] is similar to `truncate`, but causes the excess
```
kennytm, line 1004:
```suggestion
    /// The [`split_off`] method is similar to `truncate`, but causes the excess
```

maybe further reword the sentence, calling the tail "excess elements" feel very strange.

SkiFire13:
I disagree, behaviour wise `.truncate(n)` is more similar to `.drain(n..)` than `.split_off(n)` because `.split_off(n)` causes an extra allocation.

library/alloc/src/vec/mod.rs
Legend-of-iPhoenix: Sounds good. I'll change the wording to "tail elements".

Legend-of-iPhoenix:
> I disagree, behaviour wise `.truncate(n)` is more similar to `.drain(n..)` than `.split_off(n)` because `.split_off(n)` causes an extra allocation.

This is a good point- perhaps mention on `Vec::split_off` that `Vec::drain` can be used to avoid an extra allocation?

SkiFire13:
If you want, `Vec::split_off` could mention that if you don't need the tail as an owned `Vec` then `vec.drain(at..)` could be used instead of `vec.split_off(at)` to avoid an allocation.

Going back to `Vec::truncate`, I think there's a difference between `Vec::truncate` and both `Vec::drain` and `Vec::split_off` that's not currently mentioned (both in the current docs and in your PR), and that's that `Vec::truncate` will do nothing if you pass a value bigger than the length, but the other two will panic.

Legend-of-iPhoenix:
Since these suggestions expand the scope of this PR just a little bit, I'm renaming it.

Legend-of-iPhoenix:
I took the additional liberty of fixing an extremely minor grammatical error, `split into two` is more properly `split in two`.

This is now ready for further review.

SkiFire13:
I initially somehow missed a point from the first post:

> Vec::drain requires a collect to function similarly because it returns a Drain iterator

Doesn't `split_off` also require a `.into_iter()` to work similarly (but not quite, because it still allocates) to `drain`?

Anyway is it really necessary to remove the mention of `drain` from `truncate`'s docs? Can't they mention both `drain` and `split_off`? I feel like this change just encourage people to allocate more, even if they don't need it.

Legend-of-iPhoenix:
> I initially somehow missed a point from the first post:
> 
> > Vec::drain requires a collect to function similarly because it returns a Drain iterator
> 
> Doesn't `split_off` also require a `.into_iter()` to work similarly (but not quite, because it still allocates) to `drain`?

I think you might have misinterpreted me here, in large part because I forgot to update the description when I changed the title- I meant in order for `drain` to function similar to `truncate`.

> Anyway is it really necessary to remove the mention of `drain` from `truncate`'s docs? Can't they mention both `drain` and `split_off`? I feel like this change just encourage people to allocate more, even if they don't need it.

This is kinda resolved in my mind because `split_off` now prominently mentions a way to avoid this extra allocation. I'll add it anyway because it's just a few extra words, but I was concerned that talking about too many other functions in the docs for `truncate` could be confusing.

SkiFire13:
> I meant in order for drain to function similar to truncate.

I think there's misunderstanding here: you don't need to `.collect()` for `drain` to be similar to `truncate`. `Drain` already removes the items that you didn't iterate over when it is dropped. This means that `vec.drain(n..);` is already equivalent to `vec.truncate(n);` (given `n<=vec.len()`). The `.collect()` is only needed for it to be similar to `.split_off(n)`.

> This is kinda resolved in my mind because split_off now prominently mentions a way to avoid this extra allocation.

I mean, this could also be fixed by having `drain` mention `split_off` in its documentation...

Legend-of-iPhoenix:
> > I meant in order for drain to function similar to truncate.
> 
> I think there's misunderstanding here: you don't need to `.collect()` for `drain` to be similar to `truncate`. `Drain` already removes the items that you didn't iterate over when it is dropped. This means that `vec.drain(n..);` is already equivalent to `vec.truncate(n);` (given `n<=vec.len()`). The `.collect()` is only needed for it to be similar to `.split_off(n)`.

My bad, I think I managed to confuse myself here. Any other thoughts on the current proposed changes?

library/alloc/src/vec/mod.rs
```diff
@@ -1001,12 +1001,14 @@ impl<T, A: Allocator> Vec<T, A> {
...
-    /// elements to be returned instead of dropped.
-    ///
     /// Note that this method has no effect on the allocated capacity
     /// of the vector.
     ///
+    /// The [`split_off`] method is similar to `truncate`, but causes the
+    /// tail elements to be returned as a newly-allocated `Vec` instead
+    /// of being dropped. Note that [`Vec::drain`] can emulate `split_off`
+    /// and avoids this extra allocation by returning an iterator.
+    ///
```
SkiFire13, line 1011: The dance "`split_off` emulates `truncate` and `drain` emulates `split_off`" seems a bit weird. I would just mention that `Vec::drain` also returns the tail elements, but as an iterator rather than in a new allocation.

I would mention the `panic` thing here. I saw you put it later, but I don't think that's the right place. The issue with panic is not that `truncate` doesn't panic, rather that `split_off` and `drain` can emulate `truncate` only for indexes not bigger than the length, otherwise they will panic. So I think the right place to mention it should be here, probably after mentioning `Vec::drain`.

library/alloc/src/vec/mod.rs
```diff
@@ -1017,8 +1019,8 @@ impl<T, A: Allocator> Vec<T, A> {
     /// assert_eq!(vec, [1, 2]);
     /// ```
     ///
-    /// No truncation occurs when `len` is greater than the vector's current
-    /// length:
+    /// No truncation or panic occurs when `len` is greater than the vector's
+    /// current length:
```
SkiFire13, line 1023: As said above I wouldn't mention `panic` here.

library/alloc/src/vec/mod.rs
```diff
@@ -1835,12 +1837,15 @@ impl<T, A: Allocator> Vec<T, A> {
         self.len() == 0
     }
 
-    /// Splits the collection into two at the given index.
+    /// Splits the collection in two at the given index.
```
SkiFire13, line 1840: This looked ok before.

library/alloc/src/vec/mod.rs
```diff
@@ -1835,12 +1837,15 @@ impl<T, A: Allocator> Vec<T, A> {
...
     ///
     /// Returns a newly allocated vector containing the elements in the range
     /// `[at, len)`. After the call, the original vector will be left containing
     /// the elements `[0, at)` with its previous capacity unchanged.
     ///
+    /// If you do not need to own the tail `Vec`, use [`Vec::drain`] to avoid
+    /// an extra allocation.
```
SkiFire13, line 1847: This is not technically correct, with `drain` you own the tail items, they just aren't stored in a new `Vec`. What about something like "If you don't need the tail to be stored in a separate `Vec` you can use `Vec::drain` to iterate through it, taking ownership of each element."

Legend-of-iPhoenix:
@SkiFire13 this is mostly filled with your ideas at this point, I suggest creating a PR of your own :)

PR #87180, opened 2021-07-16
feat(rustdoc): open sidebar menu when links inside it are focused

### LINKED ISSUE #87172: rustdoc sidebar is not accessible from keyboard on small screens
On a PC (or on an iPad with "Full Keyboard Access" turned on), shrink the browser so that the sidebar goes into the overflow menu, then try to get at it with a keyboard.

There's no focus ring. You can actually get keyboard focus on it, in iPadOS, but you can't see it, and I can't seem to get keyboard access to it at all on desktop Firefox. In either case, it's not good enough. The hamburger button is never highlighted, because it is not keyboard focusable (it's a DIV, and not a real button). #87167 won't make this better, since if you are using the keyboard, but are not using a screenreader, the menu items will become focused but you won't be able to see them.

To combine this with #87167, JavaScript can be used to make sidebar visibility get automatically switched based on focus (if a menu item is focused, make it visible, and if the focus goes onto something outside the menu, hide it).

notriddle:
Fixes #87172

Based on #87167 (which should be merged first)

r? @GuillaumeGomez

Preview it at https://notriddle.com/notriddle-rustdoc-test/std/index.html

### DIFF
--- a/src/librustdoc/html/static/css/rustdoc.css
+++ b/src/librustdoc/html/static/css/rustdoc.css
@@ -1729,7 +1729,8 @@ details.undocumented[open] > summary::before {
 		width: calc(100% + 30px);
 	}
 
-	.show-it {
+	.show-it, .sidebar-elems:focus-within {
+		z-index:  2;
 		left: 0;
 	}
 
--- a/src/test/rustdoc-gui/sidebar-mobile.goml
+++ b/src/test/rustdoc-gui/sidebar-mobile.goml
@@ -8,3 +8,13 @@ assert-css: (".sidebar-elems", {"display": "block", "left": "-246px"})
 // Opening the sidebar menu.
 click: ".sidebar-menu"
 assert-css: (".sidebar-elems", {"display": "block", "left": "0px"})
+// Closing the sidebar menu.
+click: ".sidebar-menu"
+assert-css: (".sidebar-elems", {"display": "block", "left": "-246px"})
+// Force the sidebar open by focusing a link inside it.
+// This makes it easier for keyboard users to get to it.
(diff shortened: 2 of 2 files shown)


src/librustdoc/html/static/js/main.js
```diff
@@ -344,6 +344,28 @@ function hideThemeButtonState() {
         document.getElementsByTagName("body")[0].style.marginTop = "";
     }
 
+    function isInSidebarElems(elem) {
+        while (elem) {
+            if (elem.classList && elem.classList.contains("sidebar-elems")) {
```
GuillaumeGomez, line 349:
```suggestion
            if (hasClass(elem, "sidebar-elems")) {
````

**Comment by GuillaumeGomez** [comment]
There is a much simpler way to do it. I tested locally with:

```
$ git diff
diff --git a/src/librustdoc/html/static/css/rustdoc.css b/src/librustdoc/html/static/css/rustdoc.css
index 9e3ad667531..8cddd85854a 100644
--- a/src/librustdoc/html/static/css/rustdoc.css
+++ b/src/librustdoc/html/static/css/rustdoc.css
@@ -1729,7 +1729,7 @@ details.undocumented[open] > summary::before {
                width: calc(100% + 30px);
        }
 
-       .show-it {
+       .show-it, .sidebar-elems:focus-within {
                left: 0;
        }
```

With this, no need for JS. :)

notriddle:
@GuillaumeGomez Okay, that's done.

GuillaumeGomez:
Awesome, thanks! Just waiting for CI then I r+.

PR #86211, opened 2021-06-11
create method overview docs for core::option and core::result

tlyu:
The `Option` and `Result` types have large lists of methods. They each could use an overview page of methods grouped by category. These proposed overviews include "truth tables" for the underappreciated boolean operators/combinators of these types. The methods are already somewhat categorized in the source, but some logical groupings are broken up by the necessities of putting related methods in different `impl` blocks, for example.

This is based on #86209, but those are small changes and unlikely to conflict.

PR #87183, opened 2021-07-16
fix typo in compile_fail doctest

RalfJung:
Fixes a typo introduced by https://github.com/rust-lang/rust/pull/86211. For some reason this typo makes Miri go all crazy when running libcore doctests (https://github.com/rust-lang/miri/issues/1852). Kudos to @hyd-dev for noticing the typo.

Cc @tlyu @joshtriplett

### DIFF
--- a/library/core/src/option.rs
+++ b/library/core/src/option.rs
@@ -356,7 +356,7 @@
 //! // must have the same concrete type.
 //! fn make_iter(do_insert: bool) -> impl Iterator<Item = i32> {
 //!     // Explicit returns to illustrate return types not matching
-//!     match x {
+//!     match do_insert {
 //!         true => return (0..4).chain(once(42)).chain(4..8),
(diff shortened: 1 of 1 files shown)


tlyu:
Thanks! I would have hoped that the ordinary doctest pass would have caught that, especially given the error code number.

RalfJung:
I assume the error code number ensures that this error code is present, but does not prevent *other* errors from also being present.

jyn514:
@bors r+ rollup

It seems unfortunate that Miri gets so confused on name errors, but you're already aware of that and this is a good fix anyway.

RalfJung:
Yes, that is tracked at https://github.com/rust-lang/miri-test-libstd/issues/4. (It's not really a Miri bug, more an unfortunate consequence of rustc's behavior when suggesting alternative names [eagerly loading all the crates it can find], and the hacky way we use to execute libcore doctests without `x.py`.)

PR #87187, opened 2021-07-16
Fix NixOS detection

oxalica:
Use `/etc/os-release` instead of `/etc/NIXOS` for detection.
The latter one does not exist on NixOS when using tmpfs as root.

### DIFF
--- a/src/bootstrap/bootstrap.py
+++ b/src/bootstrap/bootstrap.py
@@ -580,7 +580,13 @@ class RustBuild(object):
         if ostype != "Linux":
             return
 
-        if not os.path.exists("/etc/NIXOS"):
+        # Use `/etc/os-release` instead of `/etc/NIXOS`.
+        # The latter one does not exist on NixOS when using tmpfs as root.
+        try:
+            with open("/etc/os-release", "r") as f:
+                if not any(line.strip() == "ID=nixos" for line in f):
+                    return
+        except FileNotFoundError:
             return
         if os.path.exists("/lib"):
             return


nagisa:
r? @nagisa 

Under what circumstances might `/etc/NIXOS` not exist? AFAICT from grepping around `nixpkgs` this file is ubiquitous.

oxalica:
> Under what circumstances might /etc/NIXOS not exist? 

I mentioned it in the code comment. It doesn't exist if you are using tmpfs root (use auto population of /etc on every boot).
`/etc/NIXOS` is only created in installation script but not in activation script.

Generally I think `/etc/os-release` should be preferred to do "distro detection", and it's more extensible if we need special treatment for some other distro.

> AFAICT from grepping around nixpkgs this file is ubiquitous.

Not many outside installation but do have some. :thinking: Maybe I should also send a PR to nixpkgs to make it always present.

nagisa:
> I mentioned it in the code comment.

Cool, I missed that when quickly looking through the PR.

Would be great if this comment was moved to a commit message and/or the PR description. A code comment here is not particularly useful, I feel. There's no context to a person reading code (especially so the future) as to why the reference to `/etc/NIXOS` exists at all.

---

> Generally I think /etc/os-release should be preferred to do "distro detection", and it's more extensible if we need special treatment for some other distro.

Sure, though I would argue that extensibility and generality is not really a concern here given that the following fixups are extremely NixOS specific.

---

r=me once the PR description and/or commit messages are adjusted to include information about tmpfs nixos activations not having this file.

oxalica:
@nagisa Fixed.

nagisa:
@bors r+

Thanks!

PR #87191, opened 2021-07-16
Package LLVM libs for the target rather than the build host

### LINKED ISSUE #85250: x.py dist cross-compile packages the wrong LLVM build
### Reproduction, on an x86_64 host:
`./x.py dist --host aarch64-unknown-linux-gnu --target aarch64-unknown-linux-gnu`
or
`./src/ci/docker/run.sh dist-aarch64-linux`

### Expected:
All packaged artifacts are targeted at aarch64.

### Actual:
The LLVM libs in `dist/rust-dev-nightly-aarch64-unknown-linux-gnu.tar.xz` are x86_64.

### More info
The binaries in the same archive are fine.

Building rustc on an aarch64 host with `download-ci-llvm = if-available` results in linker errors.
Setting this to false and building LLVM locally works fine.

https://rust-lang.zulipchat.com/#narrow/stream/242791-t-infra/topic/llvm-project.20aarch64.20builds.20corrupt
The oldest CI artifact currently available (https://ci-artifacts.rust-lang.org/rustc-builds/6645da366eed0c61258a04265bea513e94df7ea6/rust-dev-nightly-aarch64-unknown-linux-gnu.tar.gz) has this bug.

It could conceivably affect other distribution artifacts cross-compiled from x86_64 but I haven't tested them.

adamgemmell:
Fixes https://github.com/rust-lang/rust/issues/85250

`dist.rs` uses, in the `rust-dev` stage, `llvm-config --libfiles` to get a list of the LLVM library files built but of course only for the build host. If the target differs we want to package lib files from the target's build tree instead. This is done by splitting/rejoining the paths on their build directories.

At the moment `tree` on the LLVM build directories seems to give almost identical output, but of course this might not be the case in the future. If a file is missing in the target's build tree then this stage will error in the `builder.install()` call. If the target build tree has an extra file then it silently won't be copied and we'll get a linker error when building using this artifact (via `download-ci-llvm = "if-available"`), though we would have received a linker error anyway without this change.

There was also a typo in the example config around this option.

### DIFF
--- a/config.toml.example
+++ b/config.toml.example
@@ -38,7 +38,7 @@ changelog-seen = 2
 # This is false by default so that distributions don't unexpectedly download
 # LLVM from the internet.
 #
-# All tier 1 targets are currently supported; set this to `"if-supported"` if
+# All tier 1 targets are currently supported; set this to `"if-available"` if
 # you are not sure whether you're on a tier 1 target.
 #
 # We also currently only support this when building LLVM for the build triple.
--- a/src/bootstrap/dist.rs
+++ b/src/bootstrap/dist.rs
@@ -1955,8 +1955,16 @@ fn maybe_install_llvm(builder: &Builder<'_>, target: TargetSelection, dst_libdir
         cmd.arg("--libfiles");
         builder.verbose(&format!("running {:?}", cmd));
         let files = output(&mut cmd);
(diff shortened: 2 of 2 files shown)


Mark-Simulacrum:
Seems OK. We can always iterate further, but this still seems an improvement over the previous state. Thanks!

@bors r+

PR #85530, opened 2021-05-20
Use Rust Symbol Mangling by default

tmiasko:
r? @michaelwoerister

PR #83234, opened 2021-03-17
Use ValTree in all type level constants

oli-obk:
based on https://github.com/rust-lang/rust/pull/83207

r? @lcnr 

best reviewed commit by commit, as some of the refactoring commits are noisy but don't change anything significant.

cc @rust-lang/wg-const-eval 

Type level constants are now fully backed by value trees, so it is impossible for two constants to have the same value but a different representation.

Issue #61486, opened 2019-06-03
Value mangling


varkor:
For const generics, we will eventually allow arbitrary `structural_match` values as generic arguments. Thus, we need to figure out how to mangle them. At the moment, only unsigned integers are supported.

cc @eddyb @yodaldevoid

eddyb:
cc #60705

est31:
> At the moment, only unsigned integers are supported.

bools work too, don't they? https://github.com/rust-lang/rust/issues/61383#issuecomment-497534678

varkor:
@est31: the current implementation of const generics supports more types than the mangle supports, which is something we need to address.

est31:
@varkor I see. As you are special casing bools already in #61409 do you think they should be allowed in the initial version of const generics, too? Mangling schemes for them are straightforward.

varkor:
It's unclear what "initial version of const generics" means at this point. We're not ready to talk about stabilisation yet. Hopefully we'll be able to discuss proper mangling soon. I know @eddyb already has some ideas.

yodaldevoid:
Quoting @eddyb from Discord channel #compiler for my own benefit if nothing else:

> unsigned integers is really what people want, IMO. but I have an idea of how to encode arbitrary ADTs into the symbol mangling, so it shouldn't be too bad

> (basically something like `std::fmt::Formatter::debug_{list,struct,tuple}`, and the struct/tuple forms can take a path that's the path of the variant in an enum, for enum values)

> lol I didn't finish my thought: you'd encode `Some(0i32)` as something like `TNvINt...6OptioniE4Somei0_E`
that is, `...::Option<i32>::Some(0i32)`

oli-obk:
Such printing should probably share the walking logic with the const validator (https://github.com/rust-lang/rust/blob/master/src/librustc_mir/interpret/validity.rs), because both need to do the same thing.

eddyb:
These tests fail with `-Zsymbol-mangling-version=v0`:
```
    [ui] ui/const-generics/fn-const-param-call.rs
    [ui] ui/const-generics/issues/issue-61432.rs
    [ui] ui/const-generics/issues/issue-62579-no-match.rs
    [ui] ui/const-generics/raw-ptr-const-param-deref.rs
    [ui] ui/const-generics/slice-const-param.rs
```

[...]

PR #87194, opened 2021-07-16
rustc_symbol_mangling: support structural constants and &str in v0.

eddyb:
This PR should unblock #85530 (except for float `const` generics, which AFAIK should've never worked).
(cc @tmiasko could the https://github.com/rust-lang/rust/pull/85530#issuecomment-857855379 failures be retried with a quick crater "subset" run of this PR + changing the default to `v0`? Just to make sure I didn't miss anything other than the floats)

The encoding is the one suggested before in e.g. https://github.com/rust-lang/rust/issues/61486#issuecomment-878932102, tho this PR won't by itself finish #61486, before closing that we'd likely want to move to @oli-obk's "valtrees" (i.e. #83234 and other associated work).

<hr>

**EDITs**:
1. switched unit/tuple/braced-with-named-fields `<const-fields>` prefixes from `"u"`/`"T"`/`""` to `"U"`/`"T"`/`"S"` to avoid the ambiguity reported by @tmiasko in https://github.com/rust-lang/rust/pull/87194#issuecomment-884279921.

2. `rustc-demangle` PR: https://github.com/alexcrichton/rustc-demangle/pull/55

3. RFC amendment PR: https://github.com/rust-lang/rfcs/pull/3161
    * also removed the grammar changes included in that PR, from this description

4. added tests (temporarily using my fork of `rustc-demangle`)

<hr>

r? @michaelwoerister

### DIFF
--- a/compiler/rustc_symbol_mangling/src/v0.rs
+++ b/compiler/rustc_symbol_mangling/src/v0.rs
@@ -1,8 +1,10 @@
 use rustc_data_structures::base_n;
 use rustc_data_structures::fx::{FxHashMap, FxHashSet};
 use rustc_hir as hir;
+use rustc_hir::def::CtorKind;
 use rustc_hir::def_id::{CrateNum, DefId};
 use rustc_hir::definitions::{DefPathData, DisambiguatedDefPathData};
+use rustc_middle::mir::interpret::ConstValue;
 use rustc_middle::ty::layout::IntegerExt;
 use rustc_middle::ty::print::{Print, Printer};
 use rustc_middle::ty::subst::{GenericArg, GenericArgKind, Subst};
@@ -11,6 +13,7 @@ use rustc_target::abi::Integer;
 use rustc_target::spec::abi::Abi;
 
 use std::fmt::Write;
+use std::iter;
 use std::ops::Range;
 
 pub(super) fn mangle(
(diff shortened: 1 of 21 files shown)


michaelwoerister:
Nice! Thanks for moving on that so quickly, @eddyb! For reference, here is a link to the current grammar: https://rust-lang.github.io/rfcs/2603-rust-symbol-name-mangling-v0.html#syntax-of-mangled-names

michaelwoerister:
And here is the relevant part of the grammar that this PR will supersede:
```
<const> = <type> <const-data>
        | <type> "p" // placeholder (e.g. for polymorphic constants), shown as _: T
        | <backref>

// The encoding of a constant depends on its type, currently only
// unsigned integers (mainly usize, for arrays) are supported, and they
// use their value, in base 16 (0-9a-f), not their memory representation.
<const-data> = {<hex-digit>} "_"
```

michaelwoerister:
This looks great! 

So now we just need support for this to land in `rustc-demangle` so that we can add test cases, right?

tmiasko:
```
<const-fields> = "u"                        // X
               | "T" {<const>} "E"          // X(a, b, c, ...)
               | {<identifier> <const>} "E" // X { field: value, ... }
```

The first and third production can start with `u` because of Punycode identifiers.

tmiasko:
I extended LLVM Rust demangler with support for proposed grammar. Apart from issue with `u` in `<const-fields>`, everything else seems good to me. Special casing `&str` to avoid printing `&*".."` was easy enough.

Since you are also updating the test cases here. I remember having to change the `src/test/codegen/vec-shrink-panic.rs` when enabling v0 by default. Is this no longer the case?

eddyb:
> The first and third production can start with u because of Punycode encoded identifiers.

Ugh, I forgot that's how we did that, and when I saw it recently in `rustc-demangle` I didn't connect the dots.
Thanks for pointing it out! Definitely need 
I think what I'd want to do then is maybe just use a plain `E` for that case (also I should assert in the mangler that there are no fields).

> I remember having to change the `src/test/codegen/vec-shrink-panic.rs` when enabling v0 by default. Is this no longer the case?

I run the codegen tests and didn't have to change anything, is it a stage2-only test or something?

michaelwoerister:
> The first and third production can start with `u` because of Punycode encoded identifiers.

Great catch, @tmiasko!

tmiasko:
> I think what I'd want to do then is maybe just use a plain `E` for that case (also I should assert in the mangler that there are no fields).

That was my first though as well, although we might still want to make a distinction between `struct A;` and `struct A {}`?

> > I remember having to change the `src/test/codegen/vec-shrink-panic.rs` when enabling v0 by default. Is this no longer the case?
> 
> I run the codegen tests and didn't have to change anything, is it a stage2-only test or something?

This could be the `ignore-debug` annotation used there.

eddyb:
> although we might still want to make a distinction between `struct A;` and `struct A {}`?

Ah! Thanks again, I forgot that the list of fields for the named-fields case can be empty. Seems like I should've "just" picked a prefix letter for the named-fields case, that's safer than trying to make unprefixed unambiguous.

It's just that unlike the "unit" and "tuple" case, I wasn't sure what I would want to use. But it doesn't really matter, since it just has to be different than the other two, it doesn't have to be "coherent" with the rest of the format (like how `T` is used for tuple types and const tuples and const tuple structs/variants).

michaelwoerister:
If I understand correctly the sequence for making progress here is:
- Add support for the grammar additions to https://github.com/alexcrichton/rustc-demangle
- Update rustc's version of rustc-demangle
- Add test cases for the things added here to https://github.com/rust-lang/rust/tree/master/src/test/ui/symbol-names
- Make the new mangling scheme the default (which will also give us more testing of the changes here in the wild)
- Amend the RFC with the grammar changes

Does that sound right, @eddyb & @tmiasko?

eddyb:
Apologies for not leaving a comment with the status, I've had an in-progress `rustc-demangle` branch (with necessary prerequsites - though I've landed some of them separately already - and the hardest parts like `str` demangling) for the past week or two, I've only not submitted yet because a few other things came up. I'll try to wrap it up soon.

eddyb:
Demangling PR up at https://github.com/alexcrichton/rustc-demangle/pull/55 - next step is submitting the RFC amendment, I'll try to get to that later today.

eddyb:
I've opened the RFC amendment PR (https://github.com/rust-lang/rfcs/pull/3161 - should someone FCP it?) and just pushed a few commits with `symbol-names` tests.

We've already done a crater run through #85530, so I think we're only blocked on the RFC amendment (and reviews, I guess? if any are left).

eddyb:
Ugh, right, can't use `core`/`std` paths because they will keep changing hashes.

compiler/rustc_symbol_mangling/src/v0.rs
```diff
@@ -534,39 +537,153 @@ impl Printer<'tcx> for &mut SymbolMangler<'tcx> {
...
+                                    // HACK(eddyb) this mimics `path_append`,
+                                    // instead of simply using `field_def.ident`,
+                                    // just to be able to handle disambiguators.
+                                    let disambiguated_field =
+                                        self.tcx.def_key(field_def.did).disambiguated_data;
+                                    let field_name =
+                                        disambiguated_field.data.get_opt_name().map(|s| s.as_str());
+                                    self.push_disambiguator(
+                                        disambiguated_field.disambiguator as u64,
+                                    );
```
tmiasko, line 671: Under what circumstances would we have a disambiguator here? Could you include a test case if possible?

bjorn3: `CtorKind::Fictive` is for struct variants like `Foo` for `struct Foo { a: u8 }` and `Foo::Bar` for `enum Foo { Bar { a: u8 } }`. I think the disambiguator is necessary when you have multiple fields with the same name but different hygiene due to macros.

```rust
#![feature(decl_macro)]

macro foo($a:ident) { struct Foo { $a: u8, b: u16 } }

foo!(b);
```

has two different fields called `b`.

michaelwoerister:
Great test cases, thanks @eddyb!

eddyb:
I've just rebased but this is not yet mergeable until a new version of `rustc-demangle` is released, that includes alexcrichton/rustc-demangle#55.

michaelwoerister (approved):
From my point of view this looks good to go! Maybe @oli-obk wants to have a closer look at the interactions with const-eval?

PR #87196, opened 2021-07-16
Mark `Option::insert` as must_use

oxalica:
Some people seems misled by the function name and use it in case where a simple assignment just works.
If the return value is not used, `option = Some(value);` should be preferred instead of `option.insert(value);`

### DIFF
--- a/library/core/src/option.rs
+++ b/library/core/src/option.rs
@@ -1179,6 +1179,7 @@ impl<T> Option<T> {
     /// *val = 3;
     /// assert_eq!(opt.unwrap(), 3);
     /// ```
+    #[must_use = "if you intended to set a value, consider assignment instead"]
     #[inline]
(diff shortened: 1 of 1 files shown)


the8472:
`must_use` seems to be used (hah) somewhat inconsistently. Most methods annotated with it indicate an outright programming error, e.g. ignoring results which likely indicate error states, lazy iterators or integer methods where the original isn't modified.

This on the other hand is merely a style thing where a method that expresses the intent better is available. But it isn't without precedent, e.g. `mem::replace` is annotated like that too.

But if this is added, then should `Option::take` (as `= None`) and `Option::replace` (similar to insert) be annotated too?

camsteffen:
Seems like a highly subjective style preference to me.

joshtriplett:
@the8472 Yes, both of those should be annotated as well.

PR #87200, opened 2021-07-16
TAIT: Infer all inference variables in opaque type substitutions via InferCx

oli-obk:
The previous algorithm was correct for the example given in its
documentation, but when the TAIT was declared as a free item
instead of an associated item, the generic parameters were the
wrong ones.

cc @spastorino 

r? @nikomatsakis


spastorino (approved):

spastorino:
This looks great!

PR #87203, opened 2021-07-16
Some perf optimizations and logging

jackh726:
Various bits of (potential) perf optimizations and some logging additions/changes pulled out from #85499 

The only not extremely straightforward change is adding `needs_normalization` in `trait::project`. This is just a perf optimization to avoid fold through a type with *only* opaque types in `UserFacing` mode (as they aren't replaced).

This should be a simple PR for *anyone* to review, so I'm going to let highfive assign. But I'll go ahead and cc @nikomatsakis in case he has time today.

### DIFF
--- a/compiler/rustc_data_structures/src/obligation_forest/mod.rs
+++ b/compiler/rustc_data_structures/src/obligation_forest/mod.rs
@@ -418,6 +418,7 @@ impl<O: ForestObligation> ObligationForest<O> {
     /// be called in a loop until `outcome.stalled` is false.
     ///
     /// This _cannot_ be unrolled (presently, at least).
+    #[inline(never)]
     pub fn process_obligations<P, OUT>(&mut self, processor: &mut P) -> OUT
     where
         P: ObligationProcessor<Obligation = O>,
@@ -671,6 +672,7 @@ impl<O: ForestObligation> ObligationForest<O> {
         self.reused_node_vec = node_rewrites;
     }
 
+    #[inline(never)]
     fn apply_rewrites(&mut self, node_rewrites: &[usize]) {
         let orig_nodes_len = node_rewrites.len();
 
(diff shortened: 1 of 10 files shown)


Mark-Simulacrum:
> The only not extremely straightforward change is adding needs_normalization in trait::project. This is just a perf optimization to avoid fold through a type with only opaque types in UserFacing mode (as they aren't replaced).

Is there a chance you can extract it to a separate commit? It seems like that would help a targeted revert (if necessary) and review.

jackh726:
> > The only not extremely straightforward change is adding needs_normalization in trait::project. This is just a perf optimization to avoid fold through a type with only opaque types in UserFacing mode (as they aren't replaced).
> 
> Is there a chance you can extract it to a separate commit? It seems like that would help a targeted revert (if necessary) and review.

Sure, I'll do that when the try build has finished.

nikomatsakis (approved):

nikomatsakis:
r=me once perf is complete

jackh726:
Ok, so looks like inlining changes did help a bit. But still seeing a regression. So let's continue to search.

tmiasko:
cachegrind diff from previous try build for ctfe-stress-4-check:

```
--------------------------------------------------------------------------------
Ir                    file:function
--------------------------------------------------------------------------------
444,251,381 (54.05%)  ???:rustc_mir::interpret::place::<impl rustc_mir::interpret::eval_context::InterpCx<M>>::eval_place
281,869,444 (34.29%)  ???:rustc_middle::ty::structural_impls::<impl rustc_middle::ty::fold::TypeFoldable for &rustc_middle::ty::TyS>::has_type_flags
 50,234,830 ( 6.11%)  ???:rustc_mir::interpret::place::<impl rustc_mir::interpret::eval_context::InterpCx<M>>::ref_to_mplace
 34,603,022 ( 4.21%)  ???:rustc_mir::interpret::step::<impl rustc_mir::interpret::eval_context::InterpCx<M>>::eval_rvalue_into_place
```

`has_type_flags` definitely should have been marked as inline. As far as I can see it was also partially responsible for regression in `eval_place`.

jackh726:
Okay readded the inlining change since that did seem good. Removed the `has_type_flags` for this PR, but I do that could be a nice win in a future PR, since it's a bit intersection vs a fold.

@bors r=nikomatsakis

PR #86892, opened 2021-07-05
Add "copy to clipboard" for all code blocks and "expand" buttons

GuillaumeGomez:
Fixes #86851.

A little demo of the expand/collapse part:

https://user-images.githubusercontent.com/3050060/124787285-aa231f80-df48-11eb-898e-af85111da81e.mp4

Now some technical details: to prevent the copy of an `img` tag, I moved the image in the background of the button instead, allowing me to move it inside the CSS, preventing the size of the HTML to grow too much. By doing so however, it forced me to modify `rustdoc.css` in rustdoc directly when we generate the file. Nothing super complicated but still important enough to be noted.

The generated hidden lines are generated inside a `<span class="hidden">`, the JS handles the rest, it keeps the syntax highlighting of course. It forced me to update the `write_code` but nothing fancy either.

Another important detail: the "expand/collapse" button isn't generated if there is no hidden content.

All the new buttons are hidden in case the JS is disabled (`noscript.js`).

For the "copy code" feature, unfortunately, it cannot be tested (or at least I didn't find out how). So the only thing I can recommend is to try out locally. It's supposed to work as follow: only the visible code is copied. As simple as that. :)

Last information: I originally intended to hide the code blocks buttons (even the "run" one) by default and only make them appear when hovering the code block but I'll do that in another PR because I think this one is already big enough...

EDIT: Last information bis: In the JS code, the functions `copyCode` and `copy_path` are very similar, I intend to merge them even further in a later PR (I already started by creating the `copyContentToClipboard` function).

cc @jsha @Manishearth 
r? @jyn514

PR #87207, opened 2021-07-17
Rustdoc accessibility: use an icon for the [-]/[+] controls

notriddle:
This way, we can show the plus and minus buttons on screens, while voice
control will read off actual words "Collapse" and "Expand" instead of reading
"open brace minus close brace" and "open brace plus close brace".

Part of #87059

Preview it at: https://notriddle.com/notriddle-rustdoc-test/rustdoc-brace-minus-brace/std/index.html


src/librustdoc/html/static/css/rustdoc.css
```diff
@@ -1489,11 +1489,33 @@ details.rustdoc-toggle > summary.hideme > span {
...
 details.rustdoc-toggle > summary::before {
-	content: "[+]";
+	content: "Expand";
 	font-weight: 300;
 	font-size: 0.8em;
 	letter-spacing: 1px;
 	cursor: pointer;
+	width: 17px;
+	height: max(17px, 1.1em);
+	background: data-url(plus-17x17.png) no-repeat top left;
```
GuillaumeGomez, line 1499: Why not simply using the image directly instead of going through a `data-url`?

GuillaumeGomez:
That looks quite nice overall. Just one note: the image size isn't matching the `<details>` size anymore.

And finally: is it looking better to you @ahicks92 and @DataTriny?

ahicks92:
The thing is that screen readers actually already indicate that those can be expanded and collapsed, at least so long as they're "buttons" (I know they're not in the HTML).  So how much this matters really depends on what we do about that, though for the moment it's redundant when taken in isolation.

Also moving this to the "end of the line" would still be something we should do: "collapse pub fn then" etc. is still spammy.  This is certainly an improvement, but it's important to me to flag that this doesn't really resolve that part of the issue in the event that you're thinking that it does.

src/librustdoc/html/static/css/rustdoc.css
notriddle: Right now, the CSS file never links to images. This is probably because the images have version numbers (plus.png becomes plus1.55.0.png), so it would require patching the style sheet either way.

GuillaumeGomez: Ah right, my PR doing that hasn't been merged yet. Take a look at how I did: https://github.com/rust-lang/rust/pull/86892/files#diff-39d31d43319757bdd745b6b58f1d8593af4d28454f8ad0259b802bd776898e58R178-R203

That will allow you to simply add the background url from the rust side directly.

GuillaumeGomez:
@ahicks92 This change (moving to `<details>`) allowed us to remove **a lot** of JS, improving performance and reducing maintenance cost. For us, it was really a big improvement. The goal now is to try to make it not a loss for you. I'm just not sure how exactly. 😕

notriddle:
@GuillaumeGomez 

> That looks quite nice overall. Just one note: the image size isn't matching the `<details>` font size anymore.

The funny part is that I actually did that on purpose. Take this screenshot of the old version vs the new one:

A8CC22B6-EC3E-465B-9827-1B14D01F704B

In the old version, the button on `Expand Description` is bigger than the one on `impl<T> Vec<T, Global>`. It’s sized so that it flows with the text.

In the new version, they’re exactly the same size, so they both look like they occupy a shared “gutter” space on the left. It also, hopefully, makes it look less like source code. They’re not text. They’re icons.

notriddle:
@ahicks92

I’ve pushed a new patch to put the “Expand”/“Collapse” text on the end of the line. It should now read more like “pub fn bla bla bla collapse”.

src/librustdoc/html/static/css/rustdoc.css
```diff
@@ -1489,24 +1489,29 @@ details.rustdoc-toggle > summary.hideme > span {
 }
 
 details.rustdoc-toggle > summary::before {
-	content: "Expand";
-	font-weight: 300;
-	font-size: 0.8em;
-	letter-spacing: 1px;
+	content: " ";
```
GuillaumeGomez, line 1492: Shouldn't it be `""`?

GuillaumeGomez:
@notriddle It makes sense. We can always discuss it later if someone complains about it. :)

ahicks92:
What I'm trying to say is that with the current HTML, NVDA is smart enough to realize that the entire thing is already expanded or collapsed.  Other screen readers may or may not be smart enough to do the same and the redundant info is now at the end of the line where it can be ignored, so in so much as my approval is required you have it.  I just thought it was worth offering further clarification and specificity as to what it does and doesn't solve.

In the long run it looks like the fact these are buttons is the details element itself, and as regards that specific aspect of this I'm fine if they stay buttons.  It's not ideal, but it's also using HTML which is supposed to be accessible out of the box in the way you're supposed to use it, I'm enough of a programmer to admit that makes it not your problem, and there is a next button key in the screen reader.  I'm happy to move that discussion to the main issue before anyone does anything taking a lot of man-hours.  Suffice it to say that I'm not necessarily advocating for making you give me my h tags back, just for fixing the efficiency of navigation that got broken by removing them.

GuillaumeGomez:
> Suffice it to say that I'm not necessarily advocating for making you give me my h tags back, just for fixing the efficiency of navigation that got broken by removing them.

You mean (for example) making the function headers back into h4? I thought it was only in the table of content but maybe I misunderstood...

For the rest, maybe we can simply land this improvement and check what remains to be done?

ahicks92:
As I said, I'm fine with this improvement.  I'll provide more clarity on the main issue as to what the problem is as to not have it get lost in closed prs.

GuillaumeGomez:
@notriddle: Why did you close it? It seems like you're on the right track though... If you lack time to finish it (for the css declaration as I said), I can do it if you want?

notriddle:
Exactly. I don’t really have time to finish it. It also really isn’t as important, now.

GuillaumeGomez:
No problem, I'll finish it in the next days then. Thanks a lot for paving the way. ;)

PR #87214, opened 2021-07-17
add build support for --enable-clang

ZuseZ4:
As discussed here: https://rust-lang.zulipchat.com/#narrow/stream/131828-t-compiler/topic/Add.20configure.20flag.20to.20build.20clang

This basically allows using 
```--enable-clang```
for our build configuration.

This should be (?) the minimal set of changes required to get it working. 
There is no further integration. The main goal is having a simple way of getting compatible rustc and clang versions, based on the same llvm.

### DIFF
--- a/config.toml.example
+++ b/config.toml.example
@@ -145,6 +145,9 @@ changelog-seen = 2
 # Whether to include the Polly optimizer.
 #polly = false
 
+# Whether to build the Clang compiler.
+#clang = false
+
 # =============================================================================
 # General build configuration options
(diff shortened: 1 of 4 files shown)


ZuseZ4:
I wasn't sure whether I should add clang as a Rust or as an LLVM argument. I think the later one makes more sense. 
I just ran another build and were able to compile clang successfully.

ZuseZ4:
Closed in favor of a more general ```--unstable-llvm-key-value``` flag.

PR #87224, opened 2021-07-17
miri: better ptr-out-of-bounds errors

RalfJung:
For offsets larger than `isize::MAX`, display them as negative offsets.

r? @oli-obk

### DIFF
--- a/compiler/rustc_middle/src/mir/interpret/error.rs
+++ b/compiler/rustc_middle/src/mir/interpret/error.rs
@@ -240,12 +240,13 @@ pub enum UndefinedBehaviorInfo<'tcx> {
     /// Dereferencing a dangling pointer after it got freed.
     PointerUseAfterFree(AllocId),
     /// Used a pointer outside the bounds it is valid for.
+    /// (If `ptr_size > 0`, determines the size of the memory range that was expected to be in-bounds.)
     PointerOutOfBounds {
         alloc_id: AllocId,
-        offset: Size,
-        size: Size,
+        alloc_size: Size,
+        ptr_offset: i64,
+        ptr_size: Size,
         msg: CheckInAllocMsg,
-        allocation_size: Size,
     },
     /// Using an integer as a pointer in the wrong way.
     DanglingIntPointer(u64, CheckInAllocMsg),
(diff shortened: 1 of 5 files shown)


RalfJung:
I don't think it is possible to trigger the negative case in CTFE, but it is possible in Miri: https://github.com/rust-lang/miri/pull/1853.

oli-obk:
well... we can run the same code in CTFE: https://play.rust-lang.org/?version=nightly&mode=debug&edition=2018&gist=6670eb77e14af454abf6f33d1523f4b2

... but the error message is absolutely useless:

```
warning: unused variable: `x`
 --> src/lib.rs:5:9
  |
5 |     let x = unsafe { x.offset(-1) };
  |         ^ help: if this is intentional, prefix it with an underscore: `_x`
  |

warning: constant is never used: `FOO`
 --> src/lib.rs:2:1
  |
2 | / const FOO: () = {
3 | |     let v = [0i8; 4];
4 | |     let x = &v as *const i8;
5 | |     let x = unsafe { x.offset(-1) };
6 | | };
  | |__^
  |
  = note: `#[warn(dead_code)]` on by default

error[E0080]: evaluation of constant value failed
    | 
   ::: src/lib.rs:5:22
    |
5   |     let x = unsafe { x.offset(-1) };
    |                      ------------ inside `FOO` at src/lib.rs:5:22

For more information about this error, try `rustc --explain E0080`.
error: could not compile `playground` due to previous error; 2 warnings emitted
```

I don't know how this happened, it sounds to me like the intrinsic call is treated as a separate const evaluation?

RalfJung:
The same test as CTFE doesn't work since `offset(-1)` will overflow (which is a different error message than going out-of-bounds).

But I found a test that works. :)

> I don't know how this happened, it sounds to me like the intrinsic call is treated as a separate const evaluation?

The actual error span is inside the `offset` implementation. So the ``inside `FOO` at src/lib.rs:5:22`` is just part of the backtrace. But I am not sure why the other span is not shown...

oli-obk:
> But I am not sure why the other span is not shown...

because we don't have the libcore source available. we should still print messages, even if the source of a span is missing. probably will get solved when we put the error message as the main message, instead of a label

PR #87225, opened 2021-07-17
Various diagnostics clean ups/tweaks

estebank:
* Always point at macros, including derive macros
* Point at non-local items that introduce a trait requirement
* On private associated item, point at definition

### DIFF
--- a/compiler/rustc_errors/src/emitter.rs
+++ b/compiler/rustc_errors/src/emitter.rs
@@ -365,10 +365,7 @@ pub trait Emitter {
                     continue;
                 }
 
-                if matches!(trace.kind, ExpnKind::Inlined) {
-                    new_labels
-                        .push((trace.call_site, "in the inlined copy of this code".to_string()));
-                } else if always_backtrace {
+                if always_backtrace && !matches!(trace.kind, ExpnKind::Inlined) {
                     new_labels.push((
                         trace.def_site,
                         format!(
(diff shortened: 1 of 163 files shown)


compiler/rustc_errors/src/emitter.rs
```diff
@@ -398,13 +395,24 @@ pub trait Emitter {
...
+                                ExpnKind::Macro(MacroKind::Derive, _) =>
+                                    "this derive macro expansion".to_string(),
+                                ExpnKind::Macro(MacroKind::Bang, _) =>
+                                    "this macro invocation".to_string(),
+                                ExpnKind::Inlined => "the inlined copy of this code".to_string(),
+                                ExpnKind::Root => "in the crate root".to_string(),
+                                ExpnKind::AstPass(kind) => kind.descr().to_string(),
+                                ExpnKind::Desugaring(kind) =>
+                                    format!("this {} desugaring", kind.descr()),
+                            },
```
r00ster91, line 415: How about using a cow for this? It's already imported.
```suggestion
                            {
                              let place: Cow<'static, str> = match trace.kind {
                                  ExpnKind::Macro(MacroKind::Attr, _) => "this procedural macro expansion".into(),
                                  ExpnKind::Macro(MacroKind::Derive, _) => "this derive macro expansion".into(),
                                  ExpnKind::Macro(MacroKind::Bang, _) => "this macro invocation".into(),
                                  ExpnKind::Inlined => "the inlined copy of this code".into(),
                                  ExpnKind::Root => "in the crate root".into(),
                                  ExpnKind::AstPass(kind) => kind.descr().into(),
                                  ExpnKind::Desugaring(kind) => format!("this {} desugaring", kind.descr()).into(),
                              };
                              
                              place
                            },
```

michaelwoerister:
The changes look reasonable to me but I'd prefer if someone from @rust-lang/wg-diagnostics reviewed this.

r? @rust-lang/wg-diagnostics

michaelwoerister:
(@LeSeulArtichaut & @Mark-Simulacrum, it looks like above `r? <team>` didn't work. Does the team need to be added explicitly to highfive, maybe?)

LeSeulArtichaut:
Currently highfive only knows what's declared in there: https://github.com/rust-lang/highfive/blob/master/highfive/configs/rust-lang/rust.json

If we want to support `r?` on any team we could add them all in the highfive configs, or fall back to the GitHub team members. The former is more churn, but allows people that e.g. don't have time or are on vacation to opt out.

compiler/rustc_typeck/src/check/method/suggest.rs
```diff
@@ -933,6 +933,12 @@ impl<'a, 'tcx> FnCtxt<'a, 'tcx> {
                     item_name
                 );
                 err.span_label(item_name.span, &format!("private {}", kind));
+                let sp = self
+                    .tcx
+                    .hir()
+                    .span_if_local(def_id)
+                    .unwrap_or_else(|| self.tcx.def_span(def_id));
```
oli-obk, line 940: I think this pattern could warrant its own helper function

PR #87227, opened 2021-07-17
Move asm! and global_asm! to core::arch

### LINKED ISSUE #84019: Namespace the `asm!` macro
### Click [here](https://github.com/rust-lang/rust/issues/84019#issuecomment-875101919) for the final summary report.

Original issue preserved below:

---

As of today, most macros in Rust, despite being declared by the stdlib, are not properly namespaced in the sense that every other type, function, and trait in the stdlib is namespaced. Instead, historically, macros "live" in the crate root due to the technical limitations of Rust 1.0, which did not support namespacing for macros. This is distinct from items such as `Option` or `Vec` which are merely available as though they live in the crate root (via the prelude), but are actually defined at `std::option::Option` and `std::vec::Vec`. This has two primary downsides:

1. The stdlib API index is polluted by a wall of random, unrelated macros, many of which are relatively unimportant or irrelevant: https://doc.rust-lang.org/std/index.html#macros
2. Defining an item in the crate root is tantamount to exporting that item from the prelude, which means that adding a macro there totally bypasses the discussion of whether the macro *should* be exported from the prelude.

In Rust 1.51, [`ptr::addr_of`](https://doc.rust-lang.org/std/ptr/macro.addr_of.html) became the first stable macro to not be defined in the crate root. The machinery exists to namespace macros, and there seems to be at least loose consensus that this is worth using for new macros in the stdlib.

One of the last remaining open questions for [the stabilization of the `asm` macro](https://github.com/rust-lang/rust/issues/72016) is where it should live. Given the above, and the history of prior discussion on this topic, there are two options for consideration:

1. `core::arch::asm`
2. `core::arch::foo64::asm`, where `foo64` is *every* architecture supported by the `asm` macro.

(For conciseness I will only be referring to `asm` in this document, but this decision also applies to any and all related macros, such as `global_asm`.)

First, the non-advantages of either option:

1. Compile-time rejection of unsupported platforms: while `arch::foo64::asm` makes it immediately and syntactically obvious that a given platform is unsupported by dint of a nonexistent symbol, using `arch::asm` on an unsupported platform is still a compiler error.
2. Inclusion in the prelude: while `arch::asm` is straightforwardly obvious to export from the prelude, `arch::foo64::asm` and friends could still possibly be exported from the prelude via `#cfg`.
3. Architecture-dependent behavior: even aside from the literal assembly code itself, it is possible for the `asm` macro to have slightly different semantics on different architectures, e.g. architecture-specific register classes or clobbering behavior. While `arch::foo64::asm` is more explicit about this potential difference, it is not necessary for enabling such behavior.
4. Target-specific stabilization or deprecation: while `arch::foo64::asm` is more straightforward to deprecate/stabilize on a per-target basis, `#cfg` can be used to the same effect even for `arch::asm`.

The relevant differences between the the two options:

1. Importing the symbol: even assuming that `asm` is not added to the prelude, `arch::asm` would still be trivial to `use` in architecture-dependent code. Conversely, without a prelude addition, `arch::foo64::asm` would have to fall back on a few patterns when writing architecture-dependent code, some of which are more verbose than others. Below, Pattern #4 is the one that comes closest to the ergonomics of `arch::asm`, but it may be non-obvious, and it involves using a glob-import, which some may find distasteful (or even be linting against):
```rust
// Pattern #1 ----------------------------------------
// fully-qualified names
#[cfg(foo64)]
fn qux() {
    std::arch::foo64::asm!(...);
}
#[cfg(bar64)]
fn qux() {
    std::arch::bar64::asm!(...);
}
// Pattern #2 ----------------------------------------
// doubled #cfg attributes
#[cfg(foo64)]
use std::arch::foo64::asm;
#[cfg(bar64)]
use std::arch::bar64::asm;

#[cfg(foo64)]
fn qux() {
    asm!(...);
}
#[cfg(bar64)]
fn qux() {
    asm!(...);
}
// Pattern #3 ----------------------------------------
// context-local use
#[cfg(foo64)]
fn qux() {
    use std::arch::foo64::asm;
    asm!(...);
}
#[cfg(bar64)]
fn qux() {
    use std::arch::bar64::asm;
    asm!(...);
}
// Pattern #4 ----------------------------------------
// glob import
use std::arch::*;

#[cfg(foo64)]
fn qux() {
    foo64::asm!(...);
}
#[cfg(bar64)]
fn qux() {
    bar64::asm!(...);
}
```
2. Documentation: the documentation of `arch::asm` would have to document all architecture-dependent behavior, which could make it unwieldy. `arch::foo64::asm` would isolate architecture-dependent documentation, however, every module would be forced to duplicate all *non*-architecture-dependent `asm` documentation, which seems unfortunate in its own way.
3. Error reporting in the event of unguarded `asm`: having `arch::asm`, or else having `arch::foo64::asm` but having `asm` in the prelude, increases the chances that someone will write an `asm!` invocation that is not guarded by any `#cfg` directive or any other mention of the original author's platform. Such an unguarded invocation would probably give unhelpful errors if the code is ever compiled for a platform that supports `asm` in general but was not considered by this specific `asm` invocation. Even worse, the code could compile but have unintended behavior. Conversely, `arch::foo64::asm` essentially guarantees that an author's code will have to mention their intended platform somewhere, either in a `use` or in an expression, and users on different platforms will receive obvious "cannot find value `foo64::asm` in this scope" errors when attempting to compile it. However, this benefit *requires* that `asm` never be added to the prelude (and never adding any other way of circumventing the need to mention a platform, e.g. having *both* `arch::asm` and `arch::foo64::asm`).
4. Module proliferation: currently, `asm` supports architectures that do not have a corresponding module under `std::arch`. The `arch::foo64::asm` approach would require a module for all supported architectures going forward. This includes adding modules for targets that may not ever be supported by rustc itself (e.g. SPIR-V), but are supported by the `asm!` macro for the benefit of alternative backends. However, in either scenario alternative backends would still need to provide a patch to support new targets in `asm!`, and adding a new module for any unstably-supported target doesn't seem like a particularly onerous part of that process. However, given enough time (and enough alternative backends, e.g. a GCC one) this could lead to quite a few submodules under `arch::`, although to some degree that is the point of the `arch` module.
5. Platform-agnostic assembly: it is possible for a single `asm!` invocation to properly support multiple targets. This is the flipside of point 3 stated above: where `arch::asm` is maximally permissive, `arch::foo64::asm` is maximally strict. With `arch::foo64::asm`, writing platform-agnostic assembly would require a pattern like the following:
```rust
#[cfg(foo64)]
use std::arch::foo64 as fooish;
#[cfg(foo32)]
use std::arch::foo32 as fooish;

fn qux() {
    fooish::asm!(...);
}
```

**TL;DR**: the benefits of either approach are fairly small. `arch::asm` is easier to use, but only if `asm` is not added to the prelude. `arch::foo64::asm` could lead to better error messages in some cases, but likewise only if `asm` is not added to the prelude. If the decision is made to add `asm` to the prelude, then there is essentially no advantage to either option.

bstrie:
Follow-up to https://github.com/rust-lang/stdarch/pull/1183 .

Implements the libs-api team decision from rust-lang/rust#84019 (comment) .

In order to not break nightly users, this PR also adds the newly-moved items to the prelude. However, a decision will need to be made before stabilization as to whether these items should remain in the prelude. I will file an issue for this separately.

Fixes #84019 .

r? @Amanieu

### DIFF
--- a/library/core/src/lib.rs
+++ b/library/core/src/lib.rs
@@ -316,5 +316,35 @@ pub mod primitive;
 #[unstable(feature = "stdsimd", issue = "48556")]
 mod core_arch;
 
+#[doc = include_str!("../../stdarch/crates/core_arch/src/core_arch_docs.md")]
 #[stable(feature = "simd_arch", since = "1.27.0")]
-pub use core_arch::arch;
+pub mod arch {
+    #[stable(feature = "simd_arch", since = "1.27.0")]
+    pub use crate::core_arch::arch::*;
+
+    /// Inline assembly.
+    ///
+    /// Read the [unstable book] for the usage.
+    ///
+    /// [unstable book]: ../../unstable-book/library-features/asm.html
+    #[unstable(
+        feature = "asm",
+        issue = "72016",
+        reason = "inline assembly is not stable enough for use and is subject to change"
+    )]
+    #[rustc_builtin_macro]
+    pub macro asm("assembly template", $(operands,)* $(options($(option),*))?) {
+        /* compiler built-in */
+    }
+
+    /// Module-level inline assembly.
+    #[unstable(
+        feature = "global_asm",
+        issue = "35119",
+        reason = "`global_asm!` is not stable enough for use and is subject to change"
+    )]
+    #[rustc_builtin_macro]
+    pub macro global_asm("assembly template", $(operands,)* $(options($(option),*))?) {
+        /* compiler built-in */
+    }
+}
--- a/library/core/src/macros/mod.rs
+++ b/library/core/src/macros/mod.rs
@@ -1312,27 +1312,6 @@ pub(crate) mod builtin {
         ($cond:expr, $($arg:tt)+) => {{ /* compiler built-in */ }};
     }
 
-    /// Inline assembly.
-    ///
-    /// Read the [unstable book] for the usage.
-    ///
-    /// [unstable book]: ../unstable-book/library-features/asm.html
-    #[unstable(
-        feature = "asm",
-        issue = "72016",
-        reason = "inline assembly is not stable enough for use and is subject to change"
-    )]
-    #[rustc_builtin_macro]
-    #[macro_export]
-    macro_rules! asm {
-        ("assembly template",
-            $(operands,)*
-            $(options($(option),*))?
-        ) => {
-            /* compiler built-in */
-        };
-    }
-
     /// LLVM-style inline assembly.
     ///
     /// Read the [unstable book] for the usage.
@@ -1351,23 +1330,6 @@ pub(crate) mod builtin {
                         : $("input"(operand),)*
                         : $("clobbers",)*
                         : $("options",)*) => {
-            /* compiler built-in */
-        };
-    }
-
-    /// Module-level inline assembly.
-    #[unstable(
-        feature = "global_asm",
-        issue = "35119",
-        reason = "`global_asm!` is not stable enough for use and is subject to change"
-    )]
-    #[rustc_builtin_macro]
-    #[macro_export]
-    macro_rules! global_asm {
-        ("assembly template",
-            $(operands,)*
-            $(options($(option),*))?
-        ) => {
             /* compiler built-in */
         };
     }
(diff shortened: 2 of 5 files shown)


bstrie:
The current build failure is because the documentation for these macros isn't showing up in `core::arch` for some reason. The items are there, and they can only be used when you refer to them via the correct paths, but for whatever reason the generated `arch` module docs don't list them and they can't be found anywhere in search. They have doc comments. Investigating this.

PR #87229, opened 2021-07-17
Pre-fill the query cache instead of having two copies of the information

cjgillot:
Resolver outputs are stored inside the TyCtxt for the sole purpose of being returned by a query later.

This PR proposes copy this data directly into the query cache, so as to avoid the double storage.

The DepNodeIndex are constructed from the correct dep-node and the result is hashed as it would normally be.
The function inside the Providers struct returns a fallback value and is only called when the cache is empty.

The query needs to be eval_always for this to be sound: if a compiler session initializes less cache slots than its predecessor, we need to detect that and evaluate the query on the newly empty slots.
In consequence, the method is only provided for eval_always queries.
(An alternative design would be to insert a dependency to an always-red node when inserting manually into the cache, and remove the eval_always requirement. I am not convinced this is simpler.)

This facility is restricted to crates which depend on rustc_query_impl, so as to avoid accidental misuse by normal queries.

### DIFF
--- a/compiler/rustc_interface/src/passes.rs
+++ b/compiler/rustc_interface/src/passes.rs
@@ -12,22 +12,23 @@ use rustc_data_structures::temp_dir::MaybeTempDir;
 use rustc_errors::{ErrorReported, PResult};
 use rustc_expand::base::ExtCtxt;
 use rustc_hir::def_id::{StableCrateId, LOCAL_CRATE};
+use rustc_hir::definitions::Definitions;
 use rustc_hir::Crate;
 use rustc_lint::LintStore;
 use rustc_metadata::creader::CStore;
 use rustc_middle::arena::Arena;
 use rustc_middle::dep_graph::DepGraph;
 use rustc_middle::middle;
-use rustc_middle::middle::cstore::{MetadataLoader, MetadataLoaderDyn};
+use rustc_middle::middle::cstore::{CrateStoreDyn, MetadataLoader, MetadataLoaderDyn};
 use rustc_middle::ty::query::Providers;
-use rustc_middle::ty::{self, GlobalCtxt, ResolverOutputs, TyCtxt};
+use rustc_middle::ty::{self, GlobalCtxt, TyCtxt};
 use rustc_mir as mir;
 use rustc_mir_build as mir_build;
 use rustc_parse::{parse_crate_from_file, parse_crate_from_source_str};
 use rustc_passes::{self, hir_stats, layout_test};
 use rustc_plugin_impl as plugin;
-use rustc_query_impl::{OnDiskCache, Queries as TcxQueries};
-use rustc_resolve::{Resolver, ResolverArenas};
+use rustc_query_impl::{OnDiskCache, Queries as TcxQueries, QueryCtxt};
+use rustc_resolve::{Resolver, ResolverArenas, ResolverOutputs};
 use rustc_serialize::json;
 use rustc_session::config::{CrateType, Input, OutputFilenames, OutputType, PpMode, PpSourceMode};
 use rustc_session::lint;
@@ -142,7 +143,9 @@ mod boxed_resolver {
             f((&mut *resolver).as_mut().unwrap())
         }
 
-        pub fn to_resolver_outputs(resolver: Rc<RefCell<BoxedResolver>>) -> ResolverOutputs {
+        pub fn to_resolver_outputs(
+            resolver: Rc<RefCell<BoxedResolver>>,
+        ) -> (Definitions, Box<CrateStoreDyn>, ResolverOutputs) {
             match Rc::try_unwrap(resolver) {
                 Ok(resolver) => {
                     let mut resolver = resolver.into_inner();
@@ -756,6 +759,16 @@ pub static DEFAULT_QUERY_PROVIDERS: SyncLazy<Providers> = SyncLazy::new(|| {
     rustc_lint::provide(providers);
     rustc_symbol_mangling::provide(providers);
     rustc_codegen_ssa::provide(providers);
+    // Providers for manually initialized queries. Must be kept in sync with
+    // `initialize_query_cache`.
+    //providers.visibility is provided in rustc_privacy.
+    providers.extern_mod_stmt_cnum = |_, _| None;
+    providers.maybe_unused_trait_import = |_, _| false;
+    providers.maybe_unused_extern_crates = |_, ()| panic!();
+    providers.module_exports = |_, _| None;
+    providers.names_imported_by_glob_use = |_, ()| panic!();
+    providers.extern_prelude = |_, ()| panic!();
+    providers.main_def = |_, ()| panic!();
     *providers
 });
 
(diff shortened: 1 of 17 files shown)


LeSeulArtichaut:
I think r? @petrochenkov would be the best reviewer

compiler/rustc_save_analysis/src/dump_visitor.rs
```diff
@@ -1166,8 +1166,14 @@ impl<'tcx> Visitor<'tcx> for DumpVisitor<'tcx> {
             }
             hir::ItemKind::Use(path, hir::UseKind::Glob) => {
                 // Make a comma-separated list of names of imported modules.
-                let names = self.tcx.names_imported_by_glob_use(item.def_id);
-                let names: Vec<_> = names.iter().map(|n| n.to_string()).collect();
+                let names: Vec<_> = self
```
klensy, line 1169: This can be moved closer to use site, as there possibility that it not needed.

petrochenkov:
The resolver changes here are mechanical/trivial, and I'm unfamiliar with query plumbing.
So r? @michaelwoerister or @Aaron1011

michaelwoerister:
Thanks for the PR, @cjgillot! I'll try to take a look later this week.

compiler/rustc_interface/src/passes.rs
```diff
@@ -757,6 +760,17 @@ pub static DEFAULT_QUERY_PROVIDERS: SyncLazy<Providers> = SyncLazy::new(|| {
...
+    // `initialize_query_cache`.
+    //providers.visibility is provided in rustc_privacy.
+    providers.extern_mod_stmt_cnum = |_, _| None;
+    providers.maybe_unused_trait_import = |_, _| false;
+    providers.maybe_unused_extern_crates = |_, ()| panic!();
+    providers.module_exports = |_, _| None;
+    providers.names_imported_by_glob_use = |_, ()| panic!();
+    providers.extern_prelude = |_, ()| panic!();
+    providers.main_def = |_, ()| panic!();
+    providers.output_filenames = |_, ()| panic!();
```
bjorn3, line 773: The default query provider already panics.

compiler/rustc_interface/src/passes.rs
```diff
@@ -767,6 +781,39 @@ pub static DEFAULT_EXTERN_QUERY_PROVIDERS: SyncLazy<Providers> = SyncLazy::new(|
     extern_providers
 });
 
+/// Fill the query cache with values from the resolver.
+/// Must be kept in sync with `DEFAULT_QUERY_PROVIDERS`.
+fn initialize_query_cache<'tcx>(
```
bjorn3, line 786: I think it would be nicer if this is done as part of `TyCtxt::create_global_ctxt`.

compiler/rustc_query_impl/src/plumbing.rs
```diff
@@ -245,6 +245,14 @@ macro_rules! is_anon {
     };
 }
 
+macro_rules! if_eval_always {
```
michaelwoerister, line 248: What effect does this have exactly?

compiler/rustc_interface/src/passes.rs
michaelwoerister: I agree.

compiler/rustc_query_impl/src/plumbing.rs
bjorn3: It expands to the provided item only if `eval_always` is specified for a query.

michaelwoerister:
I think the approach is generally sound: `eval_always` is how we model inputs to the query system and eagerly populating query caches should not be a problem from a correctness point of view.

I wonder if there is any performance win to be had from making use of this being a bulk operation. E.g. we know beforehand how many items we are going to insert.

compiler/rustc_resolve/src/lib.rs
```diff
@@ -1426,7 +1440,7 @@ impl<'a> Resolver<'a> {
         Default::default()
     }
 
-    pub fn into_outputs(self) -> ResolverOutputs {
+    pub fn into_outputs(self) -> (Definitions, Box<CrateStoreDyn>, ResolverOutputs) {
```
petrochenkov, line 1443: Why is it necessary to move definitions and crate store from `ResolverOutputs` though?
`ResolverOutputs` are dismantled anyway when tcx is created, so what is the point in partially dismantling it slightly earlier?

cjgillot:
> I think the approach is generally sound: `eval_always` is how we model inputs to the query system and eagerly populating query caches should not be a problem from a correctness point of view.
> 
> I wonder if there is any performance win to be had from making use of this being a bulk operation. E.g. we know beforehand how many items we are going to insert.

I think the perf problem is that we create many more dep-nodes that are actually needed in a typical incr-unchanged session.

compiler/rustc_query_impl/src/plumbing.rs
michaelwoerister: It would be great to document this.

michaelwoerister:
I'm a bit conflicted about the changes here:

- Advantage: We don't store the data twice and safe a bit of memory that way
- Disadvantage: Eagerly moving the data into the new storage might cause more work to be done than necessary. 
- Advantage: The data in question really becomes inaccessible except via queries. This reduces the risk of accidentally using the data without dependency tracking.
- Disadvantage: Pre-filling the caches requires some additional machinery with new potential for introducing bugs.

Given that this currently introduces a performance regression and future work on querifying earlier parts of the front-end would probably make the changes mostly obsolete, I'm leaning towards tabling the changes for now and revisiting at some later point.

Alternatively, we could think about adding a new kind of query backing-storage that can take ownership of the data _as is_ (e.g. if it is already in the form of convenient hash-table) and only fall back to the regular hash table storage if a new entry is allocated. But I actually don't think that's a good idea because of the additional complexity (you'd have to handle dep-tracking in a side-table and such).

What do you think, @cjgillot?

cjgillot:
> I'm a bit conflicted about the changes here:
> 
>     * Advantage: We don't store the data twice and safe a bit of memory that way
> 
>     * Disadvantage: Eagerly moving the data into the new storage might cause more work to be done than necessary.
> 
>     * Advantage: The data in question really becomes inaccessible except via queries. This reduces the risk of accidentally using the data without dependency tracking.
> 
>     * Disadvantage: Pre-filling the caches requires some additional machinery with new potential for introducing bugs.
> 
> Given that this currently introduces a performance regression and future work on querifying earlier parts of the front-end would probably make the changes mostly obsolete, I'm leaning towards tabling the changes for now and revisiting at some later point.
> 

I agree with this assessment. I'll close this PR, to revisit later.

> Alternatively, we could think about adding a new kind of query backing-storage that can take ownership of the data _as is_ (e.g. if it is already in the form of convenient hash-table) and only fall back to the regular hash table storage if a new entry is allocated. But I actually don't think that's a good idea because of the additional complexity (you'd have to handle dep-tracking in a side-table and such).

Taking ownership of the data as is does not compute the fingerprints of the values, so the query system can't know if the data has changed or not. And creating the dep-nodes on demand is essentially what we are doing right now.

PR #87230, opened 2021-07-17
Fix docblock <table> overflow

GuillaumeGomez:
Before:

Screenshot from 2021-07-18 00-05-23

After:

Screenshot from 2021-07-18 00-03-55

cc @SergioBenitez 

r? @notriddle


GuillaumeGomez:
A potential "extension" to this fix would then to add `overflow-x: auto;` on `.docblock` directly. However it has "side-effects" which need to be investigated before (padding and margin are working weirdly with it). But that's for another day.

GuillaumeGomez:
ping @notriddle :)

notriddle (approved):

PR #83723, opened 2021-03-31
Store all HIR owners in the same container

cjgillot:
This replaces the previous storage in a BTreeMap for each of Item/ImplItem/TraitItem/ForeignItem.
This should allow for a more compact storage.

Based on https://github.com/rust-lang/rust/pull/83114

PR #87234, opened 2021-07-17
Lower only one HIR owner at a time

cjgillot:
Based on https://github.com/rust-lang/rust/pull/83723
Additional diff is here: https://github.com/cjgillot/rust/compare/ownernode...lower-mono

Lowering is very tangled and has a tendency to intertwine the transformation of different items. This PR aims at simplifying the logic by:
- moving global analyses to the resolver (item_generics_num_lifetimes, proc_macros, trait_impls);
- removing a few special cases (non-exported macros and use statements);
- restricting the amount of available information at any one time;
- avoiding back-and-forth between different owners: an item must now be lowered all at once, and its parent cannot refer to its nodes.

I also removed the sorting of bodies by span.  The diagnostic ordering changes marginally, since definitions are pretty much sorted already according to the AST. This uncovered a subtlety in thir-unsafeck.

(While these items could logically be in different PRs, the dependency between commits and the amount of conflicts force a monolithic PR.)


jyn514:
cc @petrochenkov since this moves some things to rustc_resolve

petrochenkov:
Blocked on https://github.com/rust-lang/rust/pull/84373.

petrochenkov:
Wait a second, this PR is not actually based on https://github.com/rust-lang/rust/pull/84373 like the description says.
Is it ready for review?

petrochenkov:
@cjgillot 
Could you explain what's going on with the "HirId preallocation" changes in more details?
Why were the HirIds pre-allocated? Why don't they need to be pre-allocated now? When they are allocated if they are not pre-allocated? (What does it even mean to allocate a HirId?)
Why is `item_local_id_counters` keyed on `NodeId`s and not `LocalDefId`s despite the `NodeId`s having the owner semantics?

cjgillot:
> @cjgillot
> Could you explain what's going on with the "HirId preallocation" changes in more details?
> Why were the HirIds pre-allocated? Why don't they need to be pre-allocated now? When they are allocated if they are not pre-allocated? (What does it even mean to allocate a HirId?)

By "HirId preallocation", for lack of a better word, I mean calls to `allocate_hir_id_counter`.
This pre-allocation basically declared to the lowering context that such NodeId was a HIR owner, and had to be lowered as such.
For item-likes, this was not necessary: we know that the nodes are only lowered when lowering the item itself, or when lowering a reference to the item. Calling `allocate_hir_id_counter` at that moment is enough.
For `use` items, the situation is essentially the same, with the additional subtlety that each use is associated with several definitions, and only part of those definitions actually get referenced in the HIR.

> Why is `item_local_id_counters` keyed on `NodeId`s and not `LocalDefId`s despite the `NodeId`s having the owner semantics?

I had changed that before messing up the rebase.

crlf0710:
@cjgillot Ping from triage, what's next steps here?

inquisitivecrystal:
As I understand it from [cjgillot's comment here](https://github.com/rust-lang/rust/issues/87406#issuecomment-887042123), this is blocked on #88019.

petrochenkov:
#88019 has been merged, unblocking.

compiler/rustc_hir/src/hir.rs
```diff
@@ -2898,12 +2898,12 @@ impl ForeignItemId {
...
 /// means fewer edges in the incremental compilation graph).
 #[derive(Debug, HashStable_Generic)]
-pub struct ForeignItemRef<'hir> {
+pub struct ForeignItemRef {
     pub id: ForeignItemId,
     #[stable_hasher(project(name))]
     pub ident: Ident,
     pub span: Span,
-    pub vis: Visibility<'hir>,
+    pub is_pub: bool,
```
petrochenkov, line 2906: I don't think it's necessary to keep the `is_pub` flags in HIR.
It's used only in couple of places and the same information is available through `tcx.visibility(def_id)`.

compiler/rustc_ast_lowering/src/expr.rs
```diff
@@ -1142,18 +1140,9 @@ impl<'hir> LoweringContext<'_, 'hir> {
     }
 
     fn with_catch_scope<T>(&mut self, catch_id: NodeId, f: impl FnOnce(&mut Self) -> T) -> T {
-        let len = self.catch_scopes.len();
-        self.catch_scopes.push(catch_id);
-
+        let old_scope = std::mem::replace(&mut self.catch_scope, Some(catch_id));
```
petrochenkov, line 1143:
```suggestion
        let old_scope = self.catch_scope.replace(catch_id);
```

compiler/rustc_ast_lowering/src/expr.rs
```diff
@@ -1162,17 +1151,9 @@ impl<'hir> LoweringContext<'_, 'hir> {
         let was_in_loop_condition = self.is_in_loop_condition;
         self.is_in_loop_condition = false;
 
-        let len = self.loop_scopes.len();
-        self.loop_scopes.push(loop_id);
-
+        let old_scope = std::mem::replace(&mut self.loop_scope, Some(loop_id));
```
petrochenkov, line 1154:
```suggestion
        let old_scope = self.loop_scope.replace(loop_id);
```

compiler/rustc_ast_lowering/src/lib.rs
```diff
@@ -821,11 +821,11 @@ impl<'a, 'hir> LoweringContext<'a, 'hir> {
         let was_in_loop_condition = self.is_in_loop_condition;
         self.is_in_loop_condition = false;
 
-        let catch_scopes = mem::take(&mut self.catch_scopes);
-        let loop_scopes = mem::take(&mut self.loop_scopes);
+        let catch_scope = mem::take(&mut self.catch_scope);
+        let loop_scope = mem::take(&mut self.loop_scope);
```
petrochenkov, line 796:
```suggestion
        let catch_scope = self.catch_scope.take();
        let loop_scope = self.loop_scope.take();
```

compiler/rustc_resolve/src/macros.rs
```diff
@@ -1131,6 +1131,8 @@ impl<'a> Resolver<'a> {
                 ident.span,
             );
         }
+
+        self.proc_macros = krate.proc_macros.iter().map(|id| self.local_def_id(*id)).collect();
```
petrochenkov, line 1135: `compiler\rustc_builtin_macros\src\proc_macro_harness.rs` has direct access to resolver, so it's not necessary to keep the proc macro list in AST.
The harness can collect the list and just write it to resolver (requires adding a method to `ResolverExpand` though).

petrochenkov:
I don't have any more comments, but this probably still needs perf results.
I don't know what's wrong with the perf infra right now.

cjgillot:
For perf: https://perf.rust-lang.org/compare.html?start=926f069950d7211a87fbd81778b420de420daad7&end=e3ea44590ed84e6104b2317fffebf9f3faebe1b1&stat=instructions%3Au
It uses a later commit than 05cccdc, but it does the job.
Summary analysis: overall green except style-servo-check which regresses ~10%. No idea why, and the detailed view shows time reduction.

petrochenkov:
r=me after addressing https://github.com/rust-lang/rust/pull/87234#discussion_r698029538 (removing `proc_macros` from AST), squashing review commits into main commits, and performing one more perf run to make sure that the `style-servo-check` regression is spurious.

cjgillot:
I'll rebase once #88445 is merged.

petrochenkov:
Something is indeed wrong with `style-servo-check`, although I don't understand what exactly from the detailed diff.
I can only suggest landing this piecemeal to detect the regression source, there are clearly some orthogonal parts in this PR, like moving things to resolver, `*ItemRef` visibility removal, etc.

petrochenkov:
@cjgillot 
https://github.com/rust-lang/rust/pull/88627 and https://github.com/rust-lang/rust/pull/88703 has landed, could you rebase this?

compiler/rustc_ast_lowering/src/item.rs
```diff
@@ -106,12 +94,12 @@ impl<'hir> LoweringContext<'_, 'hir> {
     // only used when lowering a child item of a trait or impl.
     fn with_parent_item_lifetime_defs<T>(
         &mut self,
-        parent_hir_id: hir::ItemId,
+        parent_hir_id: LocalDefId,
```
petrochenkov, line 97:
```suggestion
        parent_def_id: LocalDefId,
```

compiler/rustc_ast_lowering/src/item.rs
```diff
@@ -550,7 +546,7 @@ impl<'hir> LoweringContext<'_, 'hir> {
 
                 // Add all the nested `PathListItem`s to the HIR.
                 for &(ref use_tree, id) in trees {
-                    let new_hir_id = self.allocate_hir_id_counter(id);
+                    let new_hir_id = self.resolver.local_def_id(id);
```
petrochenkov, line 549:
```suggestion
                    let new_def_id = self.resolver.local_def_id(id);
```

pnkfelix:
I think this commit was one that was left out of the normal perf service.

Here is what the weekly perf triage report says about it:


Lower only one HIR owner at a time [#87234](https://github.com/rust-lang/rust/issues/87234)
- Large improvement in [instruction counts](https://perf.rust-lang.org/compare.html?start=e7958d35ca2c898a223efe402481e0ecb854310a&end=49c0861ed0fa1d95186d88df0cd4310103e70957&stat=instructions:u) (up to -1.3% on `full` builds of `unused-warnings`)
- Small regression in [instruction counts](https://perf.rust-lang.org/compare.html?start=e7958d35ca2c898a223efe402481e0ecb854310a&end=49c0861ed0fa1d95186d88df0cd4310103e70957&stat=instructions:u) (up to 0.4% on `incr-unchanged` builds of `helloworld`)


Looking at the results by eye, I'd say its clearly a big win.

@rustbot label: perf-regression-triaged

PR #87235, opened 2021-07-17
Improve diagnostics for wrongly ordered keywords in function declaration

### LINKED ISSUE #87217: Confusing diagnostic for inversion of `extern "..."` and `unsafe` in function decl
Given the following code ([playground link](https://play.rust-lang.org/?version=nightly&mode=debug&edition=2018&gist=1b788bae1dfa4ef4ab19acad7284c2e1)): 

```rust
extern "C" unsafe fn test() {}
```

The current output is:

```
error: expected `{`, found keyword `unsafe`
 --> src/lib.rs:1:12
  |
1 | extern "C" unsafe fn test() {}
  |            ^^^^^^ expected `{`

```

Ideally the output should look like the one for inverting `unsafe` and a visibility like `pub`:

```
error: expected one of `extern` or `fn`, found keyword `pub`
 --> src/main.rs:1:8
  |
1 | unsafe pub fn test() {}
  | -------^^^
  | |      |
  | |      expected one of `extern` or `fn`
  | help: visibility `pub` must come before `unsafe`: `pub unsafe`

```

Meta `rustc -V`:

```
rustc 1.55.0-nightly (74ef0c3e4 2021-07-16)
```

poliorcetics:
Fix #87217

@rustbot label A-diagnostics T-compiler

### DIFF
--- a/compiler/rustc_parse/src/parser/item.rs
+++ b/compiler/rustc_parse/src/parser/item.rs
@@ -1771,8 +1771,14 @@ impl<'a> Parser<'a> {
     pub(super) fn parse_fn_front_matter(&mut self) -> PResult<'a, FnHeader> {
         let sp_start = self.token.span;
         let constness = self.parse_constness();
+
+        let async_start_sp = self.token.span;
         let asyncness = self.parse_asyncness();
+
+        let unsafe_start_sp = self.token.span;
         let unsafety = self.parse_unsafety();
+
+        let ext_start_sp = self.token.span;
         let ext = self.parse_extern();
 
         if let Async::Yes { span, .. } = asyncness {
@@ -1787,8 +1793,35 @@ impl<'a> Parser<'a> {
                 Ok(true) => {}
                 Ok(false) => unreachable!(),
                 Err(mut err) => {
+                    // Qualifier keywords ordering check
+
+                    // This will allow the machine fix to directly place the keyword in the correct place
+                    let current_qual_sp = if self.check_keyword(kw::Const) {
+                        Some(async_start_sp)
+                    } else if self.check_keyword(kw::Async) {
+                        Some(unsafe_start_sp)
+                    } else if self.check_keyword(kw::Unsafe) {
+                        Some(ext_start_sp)
+                    } else {
+                        None
+                    };
+
+                    if let Some(current_qual_sp) = current_qual_sp {
+                        let current_qual_sp = current_qual_sp.to(self.prev_token.span);
+                        if let Ok(current_qual) = self.span_to_snippet(current_qual_sp) {
+                            let invalid_qual_sp = self.token.uninterpolated_span();
+                            let invalid_qual = self.span_to_snippet(invalid_qual_sp).unwrap();
+
+                            err.span_suggestion(
+                                current_qual_sp.to(invalid_qual_sp),
+                                &format!("`{}` must come before `{}`", invalid_qual, current_qual),
+                                format!("{} {}", invalid_qual, current_qual),
+                                Applicability::MachineApplicable,
+                            ).note("keyword order for functions declaration is `default`, `pub`, `const`, `async`, `unsafe`, `extern`");
+                        }
+                    }
                     // Recover incorrect visibility order such as `async pub`.
-                    if self.check_keyword(kw::Pub) {
+                    else if self.check_keyword(kw::Pub) {
                         let sp = sp_start.to(self.prev_token.span);
                         if let Ok(snippet) = self.span_to_snippet(sp) {
                             let vis = match self.parse_visibility(FollowedByType::No) {
--- a/src/test/ui/async-await/no-async-const.stderr
+++ b/src/test/ui/async-await/no-async-const.stderr
@@ -2,7 +2,12 @@ error: expected one of `extern`, `fn`, or `unsafe`, found keyword `const`
   --> $DIR/no-async-const.rs:4:11
    |
 LL | pub async const fn x() {}
-   |           ^^^^^ expected one of `extern`, `fn`, or `unsafe`
+   |     ------^^^^^
+   |     |     |
+   |     |     expected one of `extern`, `fn`, or `unsafe`
+   |     help: `const` must come before `async`: `const async`
+   |
+   = note: keyword order for functions declaration is `default`, `pub`, `const`, `async`, `unsafe`, `extern`
 
 error: aborting due to previous error
 
--- a/src/test/ui/async-await/no-unsafe-async.stderr
+++ b/src/test/ui/async-await/no-unsafe-async.stderr
@@ -5,15 +5,25 @@ LL | impl S {
    |        - while parsing this item list starting here
 LL |     #[cfg(FALSE)]
 LL |     unsafe async fn g() {}
-   |            ^^^^^ expected one of `extern` or `fn`
+   |     -------^^^^^
+   |     |      |
+   |     |      expected one of `extern` or `fn`
+   |     help: `async` must come before `unsafe`: `async unsafe`
 LL | }
    | - the item list ends here
+   |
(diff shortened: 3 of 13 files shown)


compiler/rustc_parse/src/parser/item.rs
```diff
@@ -1731,33 +1731,59 @@ impl<'a> Parser<'a> {
...
+                if self.look_ahead(look_ahead, |t| t.is_lit() && !t.is_bool_lit()) {
+                    look_ahead += 1;
+                }
+                continue;
+            }
+
+            return false;
+        }
+
+        return false;
```
nagisa, line 1786: I don't see this handling `pub(scope)`.

compiler/rustc_parse/src/parser/item.rs
```diff
@@ -1731,33 +1731,59 @@ impl<'a> Parser<'a> {
...
+
+            // Checking for `async` here is no problem: the function will either return `false`,
+            // thus ignoring it, or it will proceed and fail on the actual parsing and consuming
+            // in Edition 2015.
+            if self.look_ahead(look_ahead, |t| quals.iter().any(|&kw| t.is_keyword(kw))) {
+                look_ahead += 1;
+                continue;
+            }
+
+            if look_ahead <= max_extern_kw
```
nagisa, line 1772: What purpose is this check serving? AFAICT it makes the code more complicated, but the correctness of this function does not depend on this check at all?

compiler/rustc_parse/src/parser/item.rs
```diff
@@ -1787,8 +1819,39 @@ impl<'a> Parser<'a> {
                 Ok(true) => {}
                 Ok(false) => unreachable!(),
                 Err(mut err) => {
+                    // Qualifier keywords ordering check
+                    // Extern is not checked since it is the last expected one
+                    let misplaced_qualifier =
+                        [kw::Const, kw::Async, kw::Unsafe].iter().any(|&k| self.check_keyword(k));
+
+                    if misplaced_qualifier {
```
nagisa, line 1827: This test seems superfluous – the if-chain below to compute `current_qual_sp` does the same thing immediately after. Removing this check will also allow you to get rid of the unnecessary `unreachable!()`.

compiler/rustc_parse/src/parser/item.rs
```diff
@@ -1787,8 +1819,39 @@ impl<'a> Parser<'a> {
...
+                        // This will allow the machine fix to directly place the keyword in the correct place
+                        let current_qual_sp = (if self.check_keyword(kw::Const) {
```
nagisa, line 1829: Consider making this `Option<_>` where `None` would be returned from the `else` branch that currently has `unreachable!()` in it.

nagisa (commented):
Reviewed the code changes as they are right now. The test failures seem to indicate there are a fair number of effects on other diagnostics so those need to be reviewed still (which I can do after the ui test outputs are blessed)

poliorcetics:
I reverted the changes to `check_fn_front_matter for now` to make it easier to check what tests where affected by the actual parsing.

I would still like to change `check_fn_front_matter` and I have left two tests that are currently failing about this. I can remove them to allow merging this PR and do it in a separate one if needed.

compiler/rustc_parse/src/parser/item.rs
poliorcetics: It didn't, I forgot so many possibilities when coding this so I reverted for now.

compiler/rustc_parse/src/parser/item.rs
```diff
@@ -1731,33 +1731,59 @@ impl<'a> Parser<'a> {
...
-                        // Rule out unsafe extern block.
-                        && !self.is_unsafe_foreign_mod())
-                })
-            // `extern ABI fn`
-            || self.check_keyword(kw::Extern)
-                && self.look_ahead(1, |t| t.can_begin_literal_maybe_minus())
-                && self.look_ahead(2, |t| t.is_keyword(kw::Fn))
+
+        // 0   1     2     3      4      5       6
+        // pub const async unsafe extern "<ABI>" fn ...
```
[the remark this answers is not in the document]
poliorcetics: Added in the parsing phase

src/test/ui/parser/issue-87217-keyword-order/wrong-unsafe copy.rs
```diff
@@ -0,0 +1,14 @@
...
+//
+// This test ensures the compiler is helpful about them being misplaced.
+// Visibilities are tested elsewhere.
+
+extern unsafe fn test() {}
+//~^ ERROR expected `fn`, found keyword `unsafe`
+//~| NOTE expected `fn`
+//~| HELP `unsafe` must come before `extern`
+//~| SUGGESTION unsafe extern
+//~| NOTE keyword order for functions declaration is `<visibility>`, `const`, `async`, `unsafe`, `extern`, `"<ABI>"`
```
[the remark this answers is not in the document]
poliorcetics: I used `default|pub` because both cannot be present at the same time but modified the rest

poliorcetics:
I tried to write a more precise `check_fn_front_matter` function but so many diagnostics are dependent on the current one I stopped trying. This means `extern "ABI" keyword fn` will still propose `{` after `"ABI"` instead of hinting at changing keyword position but apart from that, everything should work.

poliorcetics:
I don't understand how my changes could have broken those tests. I can't run them locally too, they're always ignored and I see no way to force run them.

compiler/rustc_parse/src/parser/item.rs
```diff
@@ -1787,8 +1793,35 @@ impl<'a> Parser<'a> {
...
+                        if let Ok(current_qual) = self.span_to_snippet(current_qual_sp) {
+                            let invalid_qual_sp = self.token.uninterpolated_span();
+                            let invalid_qual = self.span_to_snippet(invalid_qual_sp).unwrap();
+
+                            err.span_suggestion(
+                                current_qual_sp.to(invalid_qual_sp),
+                                &format!("`{}` must come before `{}`", invalid_qual, current_qual),
+                                format!("{} {}", invalid_qual, current_qual),
+                                Applicability::MachineApplicable,
+                            ).note("keyword order for functions declaration is `default|pub`, `const`, `async`, `unsafe`, `extern`");
```
nagisa, line 1820: Wording nit: I really don't think it is necessary in this message to describe the mutual exclusiveness between pub and default. It only makes the message seem more complicated than it needs to be.

compiler/rustc_parse/src/parser/item.rs
```diff
@@ -1787,8 +1793,35 @@ impl<'a> Parser<'a> {
                 Ok(true) => {}
                 Ok(false) => unreachable!(),
                 Err(mut err) => {
+                    // Qualifier keywords ordering check
+
+                    // This will allow the machine fix to directly place the keyword in the correct place
+                    let current_qual_sp = if self.check_keyword(kw::Const) {
```
nagisa, line 1799: If I'm reading this correctly, we will trigger the diagnostic here below for code that has duplicate keywords, such as e.g. in

```rust
const async const fn
```

and I believe that the machine applicable suggestion will suggest a `const const async fn` which is invalid. 

It may not be critical to fix this now in this PR, but could you please at least add a test for that case?

poliorcetics: You are right, I will add a test for it.

`pub pub fn test() {}` already behaves like that and has a `MachineApplicable` fix, I think it is because it is expected to sometimes misplace keywords but repeating them is not something that happens often so it should be okay ? I can try to fix it in this PR or open another issue+PR combo, I don't know what's best here

nagisa: I don't believe this must be fixed in this PR, but I would consider that a bug.

nagisa:
Please squash the commits a little so that its 1 or 2 commits.

PR #87236, opened 2021-07-18
Simplify command-line argument initialization on unix

sunfishcode:
Simplify Rust's command-line argument initialization code on unix:
 - The cleanup code isn't needed, because it was just zeroing out non-owning variables at runtime cleanup time. After 91c3eee1735ad72b579f99cbb6919c3471747d94, Rust's command-line initialization code on unix no longer allocates `CString`s and a `Vec` at startup time.
 - The `Mutex` isn't needed; if there's somehow a call to `args()` before argument initialization has happened, the code returns return an empty list, which we can do with a null check.

With these changes, a simple cdylib that doesn't use threads avoids getting `pthread_mutex_lock`/`pthread_mutex_unlock` in its symbol table.

### DIFF
--- a/library/std/src/sys/unix/args.rs
+++ b/library/std/src/sys/unix/args.rs
@@ -14,11 +14,6 @@ pub unsafe fn init(argc: isize, argv: *const *const u8) {
     imp::init(argc, argv)
 }
 
-/// One-time global cleanup.
-pub unsafe fn cleanup() {
-    imp::cleanup()
-}
-
 /// Returns the command line arguments
 pub fn args() -> Args {
     imp::args()
@@ -127,12 +122,6 @@ mod imp {
         init_wrapper
     };
 
-    pub unsafe fn cleanup() {
-        let _guard = LOCK.lock();
-        ARGC.store(0, Ordering::Relaxed);
-        ARGV.store(ptr::null_mut(), Ordering::Relaxed);
-    }
-
     pub fn args() -> Args {
         Args { iter: clone().into_iter() }
     }
@@ -158,8 +147,6 @@ mod imp {
     use crate::ffi::CStr;
 
     pub unsafe fn init(_argc: isize, _argv: *const *const u8) {}
-
-    pub fn cleanup() {}
 
     #[cfg(target_os = "macos")]
     pub fn args() -> Args {
--- a/library/std/src/sys/unix/args.rs
+++ b/library/std/src/sys/unix/args.rs
@@ -125,11 +125,7 @@ mod imp {
             // Load ARGC and ARGV without a lock. If the store to either ARGV or
             // ARGC isn't visible yet, we'll return an empty argument list.
             let argv = ARGV.load(Ordering::Relaxed);
-            let argc = if argv.is_null() {
-                0
-            } else {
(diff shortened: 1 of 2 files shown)


library/std/src/sys/unix/args.rs
```diff
@@ -127,21 +116,16 @@ mod imp {
...
     }
 
     fn clone() -> Vec<OsString> {
         unsafe {
-            let _guard = LOCK.lock();
-            let argc = ARGC.load(Ordering::Relaxed);
+            // Load ARGC and ARGV without a lock. If the store to either ARGV or
+            // ARGC isn't visible yet, we'll return an empty argument list.
             let argv = ARGV.load(Ordering::Relaxed);
+            let argc = if argv.is_null() { 0 } else { ARGC.load(Ordering::Relaxed) };
```
RalfJung, line 128: Don't you need to do an acquire load (and a release store elsewhere) to ensure that the non-atomic load later (`*argv.offset(i)`) does not cause a data race?

library/std/src/sys/unix/args.rs
```diff
@@ -82,16 +77,10 @@ mod imp {
...
     static ARGC: AtomicIsize = AtomicIsize::new(0);
     static ARGV: AtomicPtr<*const u8> = AtomicPtr::new(ptr::null_mut());
-    // We never call `ENV_LOCK.init()`, so it is UB to attempt to
-    // acquire this mutex reentrantly!
-    static LOCK: StaticMutex = StaticMutex::new();
 
     unsafe fn really_init(argc: isize, argv: *const *const u8) {
-        let _guard = LOCK.lock();
         ARGC.store(argc, Ordering::Relaxed);
         ARGV.store(argv as *mut _, Ordering::Relaxed);
```
RalfJung, line 85: This should probably be a release store to ensure the writes to the data that sits behind this `argv` pointer are visible.

library/std/src/sys/unix/args.rs
sunfishcode: I assume by this point the memory pointed to by argv is already ready to be loaded from without synchronization, since that's how C code would use argv.

library/std/src/sys/unix/args.rs
sunfishcode: Those writes are performed before `lang_start` is called, by code that must assume that we may read from that memory immediately without atomics or synchronization.

For a spec reference, I found C11 5.1.2.2.1p2 which says the argv array is given values "prior to program startup".

RalfJung: So this is directly using the system-provided argc/argv, not something constructed by Rust itself? I guess it's technically correct then. But this seems subtle enough that it warrants a comment.

I'd probably still use release/acquire accesses. They are just more obviously correct. IMO `Relaxed` should only be used when there is a good perf reason to do so, and needs to come with comments explaining why not doing release/acquire is okay. `Relaxed` is a very subtle memory order that should not be used lightly.

sunfishcode: Yes, this is the system-provided argc/argv. I submitted #87279 to address this.

PR #86857, opened 2021-07-04
Add #[default_method_body_is_const]

fee1-dead:
@rustbot label F-const_trait_impl

PR #86853, opened 2021-07-03
Constify ?-operator for Result and Option

usbalbin:
Try to make `?`-operator usable in `const fn` with `Result` and `Option`, see #74935 . Note that the try-operator itself was constified in #87237.

TODO
* [x] Add tests for const T -> T conversions
* [x] cleanup commits
* [x] Remove `#![allow(incomplete_features)]`
* [?] Await decision in #86808 - I'm not sure
* [x] Await support for parsing `~const` in bootstrapping compiler
* [x] Tracking issue(s)? - #88674

Issue #74935, opened 2020-07-30
Tracking Issue for the use of `?` in constants


oli-obk:
The `?` operator expands to invoking `Try::into_result` for the argument and `Into::into` for the error of the result. In order to support these, we need to

* stabilize `impl const Trait for Type` items https://github.com/rust-lang/rust/issues/67792
* implement `const Try` for `Result` and `Option` in libstd
* users need to implement `const Into` for their error types
    * Since we have an `impl<T> From<T> for T`, we have to make this a const impl, too, as well as the `impl<T: From<U>, U> Into<T> for U` impl

a1phyr:
#60964 also needs to be resolved to be able to implement `const Try`.
See https://play.rust-lang.org/?version=nightly&gist=bdf8f8e2774f54c084e2cb25624ce193

jhpratt:
Would it be reasonable to permit `?` in `const fn` only for `Option<T>` for now? The only thing would have to change is the desugaring, I think.

oli-obk:
You mean to make the desugaring different if we're in a const fn?

jhpratt:
I don't see any reason we couldn't desugar to this unconditionally for `Option<T>`:

```rust
match opt {
    Some(val) => val,
    None => return None,
}
```

Sure, the `Try` trait was designed with this in mind, but ultimately it is just a matter of desugaring the syntax. There's no semantic difference. `Result<T, E>` is more complicated due to the `E::from(err)` call, which can't be stable `const fn` right now anyways.

If it's possible to stabilize it on `Option<T>` without changing the desugar, that would be great. But I don't think it is possible until we stabilize both `Try` and `const_trait_impl`, which seems a ways off still.

Plus, who knows. There might be a small performance benefit to changing the desugaring to something simpler.

oli-obk:
The reason we can't do this is that during desugaring we don't know what the type is. We compute types *after* desugaring.

jhpratt:
Ah, dang. For some reason I thought the order was the other way around.

clarfonthey:
Poking around, it's *very* confusing that this tracking issue isn't just for the traits related to `?`, but also includes `AsRef`, `AsMut`, `TryFrom`, and `TryInto`. I would expect those traits and their impls to go under #143773, and for this to be reserved for just `Try`, `FromResidual`, and `Residual`.

EDIT: I've done this under #144289.

programmerjake:
experience report from my demo of how we could constify array `try_map`:
[...]

PR #87237, opened 2021-07-18
Add feature gates for `for` and `?` in consts

jonas-schievink:
These operations seems *relatively* straightforward to support, and only seem to be blocked on `impl const Trait`.

I have included a working test for `const_try`, but `const_for` is currently unusable without reimplementing *every single* defaulted `Iterator` method, so I didn't do that.

(both features still need tracking issues before this is merged)


oli-obk:
> but `const_for` is currently unusable without reimplementing _every single_ defaulted `Iterator` method, so I didn't do that.

maybe https://github.com/rust-lang/rust/pull/86857 helps here?

jonas-schievink:
In theory, yeah, this works:

```rs
#![feature(const_refs_to_cell)]
#![feature(const_mut_refs)]
#![feature(const_fn_trait_bound)]
#![feature(const_trait_impl)]

trait Iter: Copy + Sized {
    type Item: Copy;
    
    fn next(&mut self) -> Option<Self::Item>;
    
    #[default_method_body_is_const]
    fn count(mut self) -> usize {
        let mut count = 0;
        while self.next().is_some() {
            count += 1;
        }
        count
    }
}

#[derive(Copy, Clone)]
struct ConstIter {
    next: Option<u8>,
}

impl const Iter for ConstIter {
    type Item = u8;
    
    fn next(&mut self) -> Option<u8> {
        match self.next {
            Some(val) => {
                self.next = val.checked_add(1);
                Some(val)
            }
            None => None,
        }
    }
}

const _: () = {
    let c = ConstIter { next: Some(0) }.count();
    if c != 256 {
        [()][4];
    }
};
```

...but it requires a lot of feature gates and `Copy` bounds (since everything is generic) that wouldn't work on `Iterator`. Seems like there's still a bunch of work to do until iterators work in consts.

oli-obk:
Then I would say we don't add a const_for feature gate, we can start figuring out const iterators without it.

jonas-schievink:
But we'll need one eventually, right? Why not add it now?

oli-obk:
Yea. Just right now it doesn't really get us anything actionable. I guess as long as we have a canary test showing the behaviour of the feature gate it should be ok to have an unusable feature gate. Just thought it unusual.

usbalbin:
#86853 is a PR which might have some overlap with the try-operator part of this. My goal there was to try and constify the most trivial conversion impls to enable basic usage with `Option` and `Result` (however that turned out to be blocked on some more work on `const_trait_impl`)

usbalbin:
Btw is #87576 a duplicate of #74935? :)

jonas-schievink:
Yes, these tracking issues are pretty hard to discover

compiler/rustc_feature/src/active.rs
```diff
@@ -677,6 +677,12 @@ declare_features! (
     /// Infer generic args for both consts and types.
     (active, generic_arg_infer, "1.55.0", Some(85077), None),
 
+    /// Allows `for _ in _` loops in const contexts.
+    (active, const_for, "1.55.0", Some(87575), None),
```
oli-obk, line 681: we are now in 1.56 territory

src/test/ui/consts/const-for.stderr
```diff
@@ -0,0 +1,15 @@
+error[E0015]: calls in constants are limited to constant functions, tuple structs and tuple variants
+  --> $DIR/const-for.rs:5:14
+   |
+LL |     for _ in 0..5 {}
+   |              ^^^^
```
oli-obk, line 5: I guess we should start printing the function's name in the diagnostic. Very uninformative this error XD

PR #87244, opened 2021-07-18
Better diagnostics with mismatched types due to implicit static lifetime

### LINKED ISSUE #78113: GAT lifetime mismatch with type parametrised over dyn trait object with restricted lifetime
```rust

pub trait A {}
impl A for &dyn A {}
impl A for Box<dyn A> {}

pub struct P<'a, T: ?Sized>(&'a T);
impl<'a, T: ?Sized> A for P<'a, T> {}

pub struct Q<T: ?Sized> {
    _pd: std::marker::PhantomData<T>,
}
impl A for Q<dyn A> {}

pub trait B {
    type T<'a>: A;
}

impl B for () {
    // this is ok:
    // type T<'a> = &'a dyn A;
    
    // so is this:
    // type T<'a> = P<'a, dyn A + 'a>;
    
    // this is not:
    type T<'a> = Box<dyn A + 'a>;
    
    // nor is this:
    // type T<'a> = Q<dyn A + 'a>;
}

// Note that this is also fine:
fn _f<'a>(_x: &'a ()) -> Box<dyn A + 'a> { todo!() }
// and so is this:
fn _g<'a>(_x: &'a ()) -> Q<dyn A + 'a> { todo!() }
```
Error:
```
  --> src/lib.rs:27:5
   |
27 |     type T<'a> = Box<dyn A + 'a>;
   |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ lifetime mismatch
   |
   = note: expected trait `A`
              found trait `A`
note: the lifetime `'a` as defined on the associated item at 27:12...
  --> src/lib.rs:27:12
   |
27 |     type T<'a> = Box<dyn A + 'a>;
   |            ^^
   = note: ...does not necessarily outlive the static lifetime
```

Note also that a few months back, the nightly available did accept this type of code.

jackh726:
Fixes #78113

I think this is my first diagnostics PR...definitely happy to hear thoughts on the direction/implementation here.

I was originally just trying to solve the error above, where the lifetime on a GAT was causing a cryptic "mismatched types" error. But as I was writing this, I realized that this (unintentionally) also applied to a different case: `wf-in-foreign-fn-decls-issue-80468.rs`. I'm not sure if this diagnostic should get a new error code, or even reuse an existing one. And, there might be some ways to make this even more generalized. Also, the error is a bit more lengthy and verbose than probably needed. So thoughts there are welcome too.

This PR essentially ended up adding a new nice region error pass that triggers if a type doesn't match the self type of an impl which is selected because of a predicate because of an implicit static bound on that self type.

r? @estebank

### DIFF
--- a/compiler/rustc_ast_lowering/src/path.rs
+++ b/compiler/rustc_ast_lowering/src/path.rs
@@ -336,6 +336,7 @@ impl<'a, 'hir> LoweringContext<'a, 'hir> {
                             insertion_sp,
                             suggestion,
                         );
+                        err.note("assuming a `'static` lifetime...");
                         err.emit();
                     }
                     AnonymousLifetimeMode::PassThrough | AnonymousLifetimeMode::ReportError => {
--- a/compiler/rustc_infer/src/infer/error_reporting/nice_region_error/mismatched_static_lifetime.rs
+++ b/compiler/rustc_infer/src/infer/error_reporting/nice_region_error/mismatched_static_lifetime.rs
@@ -1,0 +1,84 @@
+//! Error Reporting for when the lifetime for a type doesn't match the `impl` selected for a predicate
+//! to hold.
+
+use crate::infer::error_reporting::nice_region_error::NiceRegionError;
+use crate::infer::error_reporting::note_and_explain_region;
+use crate::infer::lexical_region_resolve::RegionResolutionError;
+use crate::infer::{SubregionOrigin, TypeTrace};
+use crate::traits::ObligationCauseCode;
+use rustc_errors::{Applicability, ErrorReported};
+use rustc_hir as hir;
+use rustc_hir::intravisit::Visitor;
+use rustc_middle::ty::{self, TypeVisitor};
+use rustc_span::MultiSpan;
+
+impl<'a, 'tcx> NiceRegionError<'a, 'tcx> {
+    pub(super) fn try_report_mismatched_static_lifetime(&self) -> Option<ErrorReported> {
+        let error = self.error.as_ref()?;
+        debug!("try_report_mismatched_static_lifetime {:?}", error);
+
+        let (origin, sub, sup) = match error.clone() {
+            RegionResolutionError::ConcreteFailure(origin, sub, sup) => (origin, sub, sup),
+            _ => return None,
+        };
+        if *sub != ty::RegionKind::ReStatic {
+            return None;
+        }
+        let cause = match origin {
+            SubregionOrigin::Subtype(box TypeTrace { ref cause, .. }) => cause,
+            _ => return None,
+        };
+        let (parent, impl_def_id) = match &cause.code {
+            ObligationCauseCode::MatchImpl(parent, impl_def_id) => (parent, impl_def_id),
+            _ => return None,
+        };
+        let binding_span = match **parent {
+            ObligationCauseCode::BindingObligation(_def_id, binding_span) => binding_span,
+            _ => return None,
+        };
+        let mut err = self.tcx().sess.struct_span_err(cause.span, "incompatible lifetime on type");
+        // FIXME: we should point at the lifetime
+        let mut multi_span: MultiSpan = vec![binding_span].into();
+        multi_span
+            .push_span_label(binding_span, "introduces a `'static` lifetime requirement".into());
+        err.span_note(multi_span, "because this has an unmet lifetime requirement");
+        note_and_explain_region(self.tcx(), &mut err, "...", sup, "...");
+        if let Some(impl_node) = self.tcx().hir().get_if_local(*impl_def_id) {
+            let ty = self.tcx().type_of(*impl_def_id);
+            let mut v = super::static_impl_trait::TraitObjectVisitor(vec![]);
+            v.visit_ty(ty);
+            let matching_def_ids = v.0;
+
+            let impl_self_ty = match impl_node {
+                hir::Node::Item(hir::Item {
+                    kind: hir::ItemKind::Impl(hir::Impl { self_ty, .. }),
+                    ..
+                }) => self_ty,
+                _ => bug!("Node not an impl."),
+            };
+
+            for matching_def_id in matching_def_ids {
+                let mut hir_v =
+                    super::static_impl_trait::HirTraitObjectVisitor(vec![], matching_def_id);
+                hir_v.visit_ty(&impl_self_ty);
+
+                let mut multi_span: MultiSpan = hir_v.0.clone().into();
+                for span in &hir_v.0 {
+                    multi_span.push_span_label(
+                        *span,
+                        "this has an implicit `'static` lifetime requirement".to_string(),
+                    );
+                    err.span_suggestion_verbose(
+                        span.shrink_to_hi(),
+                        "consider relaxing the implicit `'static` requirement",
+                        " + '_".to_string(),
+                        Applicability::MaybeIncorrect,
+                    );
+                }
+                err.span_note(multi_span, "...does not necessarily outlive the static lifetime introduced by the compatible `impl`");
+            }
+        }
+        err.emit();
+        Some(ErrorReported)
+    }
+}
--- a/compiler/rustc_infer/src/infer/error_reporting/nice_region_error/mismatched_static_lifetime.rs
+++ b/compiler/rustc_infer/src/infer/error_reporting/nice_region_error/mismatched_static_lifetime.rs
@@ -6,6 +6,7 @@ use crate::infer::error_reporting::note_and_explain_region;
 use crate::infer::lexical_region_resolve::RegionResolutionError;
 use crate::infer::{SubregionOrigin, TypeTrace};
 use crate::traits::ObligationCauseCode;
+use rustc_data_structures::stable_set::FxHashSet;
 use rustc_errors::{Applicability, ErrorReported};
 use rustc_hir as hir;
 use rustc_hir::intravisit::Visitor;
@@ -42,13 +43,12 @@ impl<'a, 'tcx> NiceRegionError<'a, 'tcx> {
         multi_span
             .push_span_label(binding_span, "introduces a `'static` lifetime requirement".into());
         err.span_note(multi_span, "because this has an unmet lifetime requirement");
-        note_and_explain_region(self.tcx(), &mut err, "...", sup, "...");
+        note_and_explain_region(self.tcx(), &mut err, "", sup, "...");
         if let Some(impl_node) = self.tcx().hir().get_if_local(*impl_def_id) {
-            let ty = self.tcx().type_of(*impl_def_id);
-            let mut v = super::static_impl_trait::TraitObjectVisitor(vec![]);
-            v.visit_ty(ty);
-            let matching_def_ids = v.0;
+            // If an impl is local, then maybe this isn't what they want. Try to
+            // be as helpful as possible with implicit lifetimes.
 
+            // First, let's get the hir self type of the impl
             let impl_self_ty = match impl_node {
                 hir::Node::Item(hir::Item {
                     kind: hir::ItemKind::Impl(hir::Impl { self_ty, .. }),
@@ -57,17 +57,32 @@ impl<'a, 'tcx> NiceRegionError<'a, 'tcx> {
                 _ => bug!("Node not an impl."),
             };
 
-            for matching_def_id in matching_def_ids {
+            // Next, let's figure out the set of trait objects with implict static bounds
+            let ty = self.tcx().type_of(*impl_def_id);
+            let mut v = super::static_impl_trait::TraitObjectVisitor(FxHashSet::default());
+            v.visit_ty(ty);
+            let mut traits = vec![];
+            for matching_def_id in v.0 {
                 let mut hir_v =
-                    super::static_impl_trait::HirTraitObjectVisitor(vec![], matching_def_id);
+                    super::static_impl_trait::HirTraitObjectVisitor(&mut traits, matching_def_id);
                 hir_v.visit_ty(&impl_self_ty);
+            }
 
-                let mut multi_span: MultiSpan = hir_v.0.clone().into();
-                for span in &hir_v.0 {
-                    multi_span.push_span_label(
-                        *span,
-                        "this has an implicit `'static` lifetime requirement".to_string(),
-                    );
+            if traits.is_empty() {
+                // If there are no trait object traits to point at, either because
+                // there aren't trait objects or because none are implicit, then just
+                // write a single note on the impl itself.
+
+                let impl_span = self.tcx().def_span(*impl_def_id);
+                err.span_note(impl_span, "...does not necessarily outlive the static lifetime introduced by the compatible `impl`");
+            } else {
+                // Otherwise, point at all implicit static lifetimes
+
+                err.note("...does not necessarily outlive the static lifetime introduced by the compatible `impl`");
+                for span in &traits {
+                    err.span_note(*span, "this has an implicit `'static` lifetime requirement");
+                    // It would be nice to put this immediately under the above note, but they get
+                    // pushed to the end.
                     err.span_suggestion_verbose(
                         span.shrink_to_hi(),
                         "consider relaxing the implicit `'static` requirement",
@@ -75,8 +90,12 @@ impl<'a, 'tcx> NiceRegionError<'a, 'tcx> {
                         Applicability::MaybeIncorrect,
                     );
                 }
-                err.span_note(multi_span, "...does not necessarily outlive the static lifetime introduced by the compatible `impl`");
             }
+        } else {
+            // Otherwise just point out the impl.
+
+            let impl_span = self.tcx().def_span(*impl_def_id);
+            err.span_note(impl_span, "...does not necessarily outlive the static lifetime introduced by the compatible `impl`");
         }
         err.emit();
         Some(ErrorReported)
--- a/src/test/ui/generic-associated-types/issue-78113-lifetime-mismatch-dyn-trait-box.stderr
+++ b/src/test/ui/generic-associated-types/issue-78113-lifetime-mismatch-dyn-trait-box.stderr
@@ -1,5 +1,5 @@
 error: incompatible lifetime on type
-  --> $DIR/issue-78113-lifetime-mismatch-dyn-trait-box.rs:16:5
+  --> $DIR/issue-78113-lifetime-mismatch-dyn-trait-box.rs:17:5
    |
 LL |     type T<'a> = Box<dyn A + 'a>;
    |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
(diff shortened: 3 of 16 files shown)


src/test/ui/generic-associated-types/issue-78113-lifetime-mismatch-dyn-trait-box.stderr
```diff
@@ -0,0 +1,34 @@
...
+   = note: ...does not necessarily outlive the static lifetime
+note: the used `impl` has a `'static` requirement
+  --> $DIR/issue-78113-lifetime-mismatch-dyn-trait-box.rs:9:20
+   |
+LL | impl A for Box<dyn A> {}
+   |                    ^ this has an implicit `'static` lifetime requirement
+help: consider relaxing the implicit `'static` requirement
+   |
+LL | impl A for Box<dyn A + '_> {}
+   |                      ^^^^
```
estebank, line 31: Given that we already have the whole thing available, it'll be interesting to try and craft the error to reduce its verbosity. What do you think of

```
error: incompatible lifetime on type
  --> $DIR/issue-78113-lifetime-mismatch-dyn-trait-box.rs:16:5
   |
LL |     type T<'a> = Box<dyn A + 'a>;
   |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
   |
note: because this has an unmet lifetime requirement
  --> $DIR/issue-78113-lifetime-mismatch-dyn-trait-box.rs:12:17
   |
LL |     type T<'a>: A;
   |                 ^ introduces a `'static` lifetime requirement
note: a compatible `impl` exists, but...
  --> $DIR/issue-78113-lifetime-mismatch-dyn-trait-box.rs:9:1
   |
LL | impl A for Box<dyn A> {}
   | ^^^^^^^^^^^^^^^^^^^^^^^^
note: ...the lifetime `'a` as defined on the associated item at 16:12...
  --> $DIR/issue-78113-lifetime-mismatch-dyn-trait-box.rs:16:12
   |
LL |     type T<'a> = Box<dyn A + 'a>;
   |            ^^
note: ...does not necessarily outlive the static lifetime introduced by the `impl`
  --> $DIR/issue-78113-lifetime-mismatch-dyn-trait-box.rs:9:20
   |
LL | impl A for Box<dyn A> {}
   |                    ^ this has an implicit `'static` lifetime requirement
help: consider relaxing the implicit `'static` requirement
   |
LL | impl A for Box<dyn A + '_> {}
   |                      ^^^^
```

or

```
error: incompatible lifetime on type
  --> $DIR/issue-78113-lifetime-mismatch-dyn-trait-box.rs:16:5
   |
LL |     type T<'a> = Box<dyn A + 'a>;
   |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ incompatible lifetimes
...
LL |     type T<'a>: A;
   |                 - introduces a `'static` lifetime requirement
note: the lifetime `'a` as defined on the associated item at 16:12...
  --> $DIR/issue-78113-lifetime-mismatch-dyn-trait-box.rs:16:12
   |
LL |     type T<'a> = Box<dyn A + 'a>;
   |            ^^
note: ...does not necessarily outlive the static lifetime introduced by the compatible `impl`
  --> $DIR/issue-78113-lifetime-mismatch-dyn-trait-box.rs:9:20
   |
LL | impl A for Box<dyn A> {}
   |                    ^ this has an implicit `'static` lifetime requirement
help: consider relaxing the implicit `'static` requirement
   |
LL | impl A for Box<dyn A + '_> {}
   |                      ^^^^
```

Doing this requires fiddling with the logic in `mismatched_static_lifetime.rs` to accomplish something like these.

compiler/rustc_infer/src/infer/error_reporting/nice_region_error/mismatched_static_lifetime.rs
```diff
@@ -0,0 +1,84 @@
...
+        note_and_explain_region(self.tcx(), &mut err, "", sup, "...");
+        note_and_explain_region(self.tcx(), &mut err, "...does not necessarily outlive ", sub, "");
+        if let Some(impl_local_def_id) = impl_def_id.as_local() {
+            let ty = self.tcx().type_of(*impl_def_id);
+            let mut v = super::static_impl_trait::TraitObjectVisitor(vec![]);
+            v.visit_ty(ty);
+            let matching_def_ids = v.0;
+
+            let impl_node =
+                self.tcx().hir().get(self.tcx().hir().local_def_id_to_hir_id(impl_local_def_id));
```
estebank, line 52:
```suggestion
            let impl_node = self.tcx().hir().get_if_local(impl_local_def_id).unwrap();
```

compiler/rustc_infer/src/infer/error_reporting/nice_region_error/mismatched_static_lifetime.rs
```diff
@@ -0,0 +1,84 @@
...
+                        "this has an implicit `'static` lifetime requirement".to_string(),
+                    );
+                    err.span_note(multi_span, "the used `impl` has a `'static` requirement");
+                    err.span_suggestion_verbose(
+                        span.shrink_to_hi(),
+                        "consider relaxing the implicit `'static` requirement",
+                        " + '_".to_string(),
+                        Applicability::MaybeIncorrect,
+                    );
+                }
```
estebank, line 78: If there are multiple things adding `'static` lifetime requirements, this emits a single `note` instead of multiple.
```suggestion
                let mut multi_span: MultiSpan = vec![hir_v.0.clone()].into();
                for span in &hir_v.0 {
                    multi_span.push_span_label(
                        *span,
                        "this has an implicit `'static` lifetime requirement".to_string(),
                    );
                    err.span_suggestion_verbose(
                        span.shrink_to_hi(),
                        "consider relaxing the implicit `'static` requirement",
                        " + '_".to_string(),
                        Applicability::MaybeIncorrect,
                    );
                }
                err.span_note(multi_span, "the used `impl` has a `'static` requirement");
```

src/test/ui/wf/wf-in-foreign-fn-decls-issue-80468.stderr
```diff
@@ -4,14 +4,22 @@ error[E0726]: implicit elided lifetime not allowed here
...
-   |                     ^^^^^^^^^^^^ lifetime mismatch
+   |                     ^^^^^^^^^^^^
    |
-   = note: expected trait `Trait`
-              found trait `Trait`
+note: because of this predicate
+  --> $DIR/wf-in-foreign-fn-decls-issue-80468.rs:8:23
+   |
+LL | pub struct Wrapper<T: Trait>(T);
+   |                       ^^^^^
```
estebank, line 17: In this case the `'static` lifetime obligation is introduced by the type parameter (`T: Trait` is implicitly `'sized`). It'd be nice to proactively handle this, detect this case and mention that.

estebank (commented):
The new output is great! I have some nitpicks that I'd like to address before we land this. 
Particularly I'd like to tweak the resulting output to be as short _and_ clear as possible to people that might not know what a "requirement" or a "bound" is, or why a `'static` lifetime requirement is involved.

>I'm not sure if this diagnostic should get a new error code, or even reuse an existing one.

If we can add a new error code that'd be great, but it is perfectly fine to land things like this _without_ an error code, at least at the beginning, as long as the error itself contains as much information as an error code index entry would.

src/test/ui/generic-associated-types/issue-78113-lifetime-mismatch-dyn-trait-box.stderr
jackh726: I like the second one. The only problem with point at the `A` on the associated type bound and saying that it introduces a 'static lifetime requirement is that it's not a requirement for the whole type, only for that trait object. While this won't be a problem *here* (because we won't imply a 'static lifetime when the type itself is shorter), I could imagine in the future extending this to also include a suggestion for explicit 'static annotations:

Example of a test where we could provide an additional suggestion in the future:
```rust
pub trait A {}
impl<'a> A for &'a (dyn A + 'static) {}  // here the type itself has a lifetime of `'a`, but the trait object has a lifetime of `'static`
pub trait B {
    type T<'a>: A;
}
impl B for () {
    type T<'a> = &'a (dyn A + 'a);
}
```

src/test/ui/wf/wf-in-foreign-fn-decls-issue-80468.stderr
jackh726: Actually, this is because the lifetime on `Ref` in `impl Trait for Ref {}` falls back to `'static` in resolution.

The lifetime for `Ref` in `pub fn repro(_: Wrapper<Ref>)` on the other hand defaults to an anonymous lifetimes.

A diagnostic here still might be useful to say something like "this lifetime is being assumed as `'static'`"

src/test/ui/generic-associated-types/issue-78113-lifetime-mismatch-dyn-trait-box.stderr
jackh726: How's it look now?

src/test/ui/wf/wf-in-foreign-fn-decls-issue-80468.stderr
jackh726: I think an extra diagnostic here might make sense for another PR?

compiler/rustc_infer/src/infer/error_reporting/nice_region_error/mismatched_static_lifetime.rs
```diff
@@ -0,0 +1,84 @@
...
+                    );
+                    err.span_suggestion_verbose(
+                        span.shrink_to_hi(),
+                        "consider relaxing the implicit `'static` requirement",
+                        " + '_".to_string(),
+                        Applicability::MaybeIncorrect,
+                    );
+                }
+                err.span_note(multi_span, "...does not necessarily outlive the static lifetime introduced by the compatible `impl`");
+            }
```
estebank, line 79: Can you add an `else` branch with the `note_and_explain_region` you had before?

src/test/ui/wf/wf-in-foreign-fn-decls-issue-80468.stderr
```diff
@@ -4,21 +4,22 @@ error[E0726]: implicit elided lifetime not allowed here
...
+  --> $DIR/wf-in-foreign-fn-decls-issue-80468.rs:8:23
+   |
+LL | pub struct Wrapper<T: Trait>(T);
+   |                       ^^^^^ introduces a `'static` lifetime requirement
+note: ...the anonymous lifetime #1 defined on the method body at 16:5...
   --> $DIR/wf-in-foreign-fn-decls-issue-80468.rs:16:5
    |
 LL |     pub fn repro(_: Wrapper<Ref>);
    |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
-   = note: ...does not necessarily outlive the static lifetime
```
estebank, line 20: With the previous request, we would get back this line, and would make the text (and the ...) flow better.

src/test/ui/wf/wf-in-foreign-fn-decls-issue-80468.stderr
```diff
@@ -4,21 +4,22 @@ error[E0726]: implicit elided lifetime not allowed here
...
    |
-   = note: expected trait `Trait`
-              found trait `Trait`
-note: the anonymous lifetime #1 defined on the method body at 16:5...
+note: because this has an unmet lifetime requirement
+  --> $DIR/wf-in-foreign-fn-decls-issue-80468.rs:8:23
+   |
+LL | pub struct Wrapper<T: Trait>(T);
+   |                       ^^^^^ introduces a `'static` lifetime requirement
+note: ...the anonymous lifetime #1 defined on the method body at 16:5...
```
estebank, line 18: Now that the text reads like this...

compiler/rustc_infer/src/infer/error_reporting/nice_region_error/mismatched_static_lifetime.rs
```diff
@@ -0,0 +1,84 @@
...
+            ObligationCauseCode::BindingObligation(_def_id, binding_span) => binding_span,
+            _ => return None,
+        };
+        let mut err = self.tcx().sess.struct_span_err(cause.span, "incompatible lifetime on type");
+        // FIXME: we should point at the lifetime
+        let mut multi_span: MultiSpan = vec![binding_span].into();
+        multi_span
+            .push_span_label(binding_span, "introduces a `'static` lifetime requirement".into());
+        err.span_note(multi_span, "because this has an unmet lifetime requirement");
+        note_and_explain_region(self.tcx(), &mut err, "...", sup, "...");
```
estebank, line 45: ...I would change that back to
```suggestion
        note_and_explain_region(self.tcx(), &mut err, "", sup, "...");
```
so that the are no stray leading ... without an earlier corresponding trailing ...

estebank (commented):
> I think an extra diagnostic here might make sense for another PR?

I would prefer if it was in the same PR, but I'm ready to merge this after addressing my nitpicks if you don't have the time to try and handle that case.

src/test/ui/generic-associated-types/issue-78113-lifetime-mismatch-dyn-trait-box.stderr
estebank: Looks great!

jackh726:
Ok @estebank I modified the output just a bit, in order to make it a bit better in cases where there is an *explicit* static, multiple implicit statics, or when the impl isn't local.

I also added a small note to the assumed static in cases of missing lifetimes. This isn't in the implicit static error message, but should maybe be enough for someone to figure out what's going on. I'd rather land this and maybe work more on that another time.

compiler/rustc_ast_lowering/src/path.rs
```diff
@@ -336,6 +336,7 @@ impl<'a, 'hir> LoweringContext<'a, 'hir> {
                             insertion_sp,
                             suggestion,
                         );
+                        err.note("assuming a `'static` lifetime...");
```
estebank, line 339: Not a fan of this new note, but I see your point.

pnkfelix:
This was flagged as causing a moderate regression with respect to instruction counts in the compiler: [perf.rlo link](https://perf.rust-lang.org/compare.html?start=718d53b0cb7dde93499cb92950d60b412f5a3d05&end=da7d405357600a76f2b93b8aa41fe5ee5da7885d&stat=instructions:u)

Most of the regressions are under the 1% threshold, but there's a lot of them that exceed the 0.2% threshold we use for non-noisy benchmarks.

Was this PR, which I understand adds an extra region error pass, expected to cause that much additional overhead? Skimming, I see a place where a `Vec` was replaced with a `HashMap`, maybe it is from that change?

jackh726:
@pnkfelix nope that's unexpected; the pass should should only run after a region error has been generated. The only thing that could have possibly had an effect, I think, is https://github.com/rust-lang/rust/pull/87244/files#diff-34d5893dd95fe09f9a0fd3341efacd1c21853bb34ba29d9d79bb9af26bb8a0a0R1906-R1911. Which is...not much of a change.

Mark-Simulacrum:
Log of investigation work:

Selected the most prominent regression, stm32f4-check full.

First checked the self-profile view on perf.rlo for any changes in # of queries run. That can point to obvious causes, but not in this case. This means we need some local cachegrind diff'ing.

```
# Get our parent/master toolchains
rustup-toolchain-install-master 718d53b0cb7dde93499cb92950d60b412f5a3d05 da7d405357600a76f2b93b8aa41fe5ee5da7885d
# Benchmark our top regressed benchmark in cachegrind
cargo run --release --bin collector -- profile_local cachegrind +718d53b0cb7dde93499cb92950d60b412f5a3d05 before --include stm32f4 --builds Check --runs Full
cargo run --release --bin collector -- profile_local cachegrind +da7d405357600a76f2b93b8aa41fe5ee5da7885d after --include stm32f4 --builds Check --runs Full
```

This generates results/cgout-{before,after}-stm32f4-Check-Full

Using the output of cg_diff directly doesn't work well -- the hashes for each compiler are embedded into the paths and that makes cg_diff fail. I usually directly sed the cgout files rather than using the built-in --mod-filename support, since it's more general and works OK anyway. rustfilt is also helpful to avoid any crate hashes and such in mangled names that also interfere with diffing.

In this case:

```bash
cg_diff results/cgout-{before,after}-stm32f4-Check-Full | \
sed 's@da7d405357600a76f2b93b8aa41fe5ee5da7885d@@' | \
        sed 's@718d53b0cb7dde93499cb92950d60b412f5a3d05@@' | \
        rustfilt | \
        cg_annotate /dev/stdin | less
```

This produces:

```
140,274,957 (26.56%)  ???:rustc_trait_selection::traits::select::SelectionContext::match_impl
 55,140,172 (10.44%)  ???:core::ptr::drop_in_place<rustc_middle::traits::ObligationCauseCode>
 50,909,434 ( 9.64%)  /rustc//obj/build/x86_64-unknown-linux-gnu/stage1-rustc/x86_64-unknown-linux-gnu/release/build/tikv-jemalloc-sys-8def611f6>
 50,002,422 ( 9.47%)  /rustc//obj/build/x86_64-unknown-linux-gnu/stage1-rustc/x86_64-unknown-linux-gnu/release/build/tikv-jemalloc-sys-8def611f6>
 43,048,392 ( 8.15%)  /rustc//obj/build/x86_64-unknown-linux-gnu/stage1-rustc/x86_64-unknown-linux-gnu/release/build/tikv-jemalloc-sys-8def611f6>
 32,248,934 ( 6.11%)  /rustc//obj/build/x86_64-unknown-linux-gnu/stage1-rustc/x86_64-unknown-linux-gnu/release/build/tikv-jemalloc-sys-8def611f6>
 32,223,227 ( 6.10%)  /rustc//obj/build/x86_64-unknown-linux-gnu/stage1-rustc/x86_64-unknown-linux-gnu/release/build/tikv-jemalloc-sys-8def611f6>
 25,111,562 ( 4.75%)  /rustc///library/std/src/sys/unix/alloc.rs:__rdl_alloc
 21,427,608 ( 4.06%)  /rustc//obj/build/x86_64-unknown-linux-gnu/stage1-rustc/x86_64-unknown-linux-gnu/release/build/tikv-jemalloc-sys-8def611f6>
 14,349,464 ( 2.72%)  /rustc//obj/build/x86_64-unknown-linux-gnu/stage1-rustc/x86_64-unknown-linux-gnu/release/build/tikv-jemalloc-sys-8def611f6>
 14,349,464 ( 2.72%)  /rustc//obj/build/x86_64-unknown-linux-gnu/stage1-rustc/x86_64-unknown-linux-gnu/release/build/tikv-jemalloc-sys-8def611f6>
  7,174,732 ( 1.36%)  /rustc//obj/build/x86_64-unknown-linux-gnu/stage1-rustc/x86_64-unknown-linux-gnu/release/build/tikv-jemalloc-sys-8def611f6>
  7,174,732 ( 1.36%)  /rustc///library/std/src/alloc.rs:__rdl_alloc
  7,143,206 ( 1.35%)  /rustc//obj/build/x86_64-unknown-linux-gnu/stage1-rustc/x86_64-unknown-linux-gnu/release/build/tikv-jemalloc-sys-8def611f6>
  4,673,042 ( 0.88%)  ???:<rustc_middle::ty::sty::TraitRef as rustc_infer::infer::at::ToTrace>::to_trace
 -3,618,984 (-0.69%)  ???:hashbrown::map::RawEntryBuilderMut<K,V,S,A>::from_hash
  3,587,366 ( 0.68%)  /rustc//obj/build/x86_64-unknown-linux-gnu/stage1-rustc/x86_64-unknown-linux-gnu/release/build/tikv-jemalloc-sys-8def611f6>
  3,587,366 ( 0.68%)  ???:__rust_alloc
  3,587,366 ( 0.68%)  ???:__rust_dealloc
  3,587,366 ( 0.68%)  /rustc///library/std/src/sys/unix/alloc.rs:__rdl_dealloc
  2,586,312 ( 0.49%)  ???:rustc_infer::infer::InferCtxt::commit_if_ok
  1,605,085 ( 0.30%)  ???:rustc_resolve::Resolver::resolve_ident_in_lexical_scope
    869,968 ( 0.16%)  ???:rustc_span::hygiene::HygieneData::is_descendant_of
    748,057 ( 0.14%)  /rustc//obj/build/x86_64-unknown-linux-gnu/stage1-rustc/x86_64-unknown-linux-gnu/release/build/tikv-jemalloc-sys-8def611f6>
    625,536 ( 0.12%)  ???:rustc_infer::infer::at::At::eq_impl_headers
```

Overall cachegrind is pretty good at pointing in a vague direction -- inlining and such can obscure a little what the actual effects of a change like this are. That said, it is pretty obvious based on the above that the regression is indeed due to the [new ObligationCause being created](https://github.com/rust-lang/rust/pull/87244/files#diff-34d5893dd95fe09f9a0fd3341efacd1c21853bb34ba29d9d79bb9af26bb8a0a0R1906).

The diff above has some more minor changes in other functions - but we can mostly ignore those in this case. The majority of the delta is pretty clearly from extra allocation/deallocation and the creation of the cause code itself.

Since the investigation points at an obvious cause and a pretty 'unavoidable' one, I'm going to go ahead and mark this as triaged. I've also gone ahead and filed https://github.com/rust-lang/rust/pull/89000 with a possible mitigation, but regardless of whether that works, I don't think this needs further triage/investigation.

jackh726:
Thank you so much for looking into this @Mark-Simulacrum ♥

Issue #87184, opened 2021-07-16
const-eval: loading or overwriting parts of a pointer is not supported


RalfJung:
The Miri core engine can only represent "full" pointers in memory, no parts of a pointer. This leads to strange behavior on code like
```rust
    let mut p = &42;
    unsafe {
        let ptr: *mut _ = &mut p;
        *(ptr as *mut u8) = 123; // overwrite the first byte of the pointer
    }
    let x = *p; //~ ERROR this operation requires initialized memory
```
If overwriting a part of a pointer happens during CTFE, we halt execution (since https://github.com/rust-lang/rust/pull/87248). In Miri, instead we de-initialize the entire pointer, so a write will affect the bytes "next to it". (Halting execution is [not an option](https://github.com/rust-lang/rust/pull/87248#issuecomment-882033219) here.)

If *loading* a part of a pointer (including as the source of a mem-to-mem copy) happens during CTFE or Miri, we halt execution.

Long-term, it would be great to implement support for having just a few bytes of a pointer in a Miri core engine `Allocation`. However, this might be hard to do without a perf regression.

Cc @rust-lang/wg-const-eval

oli-obk:
>     * We throw a "not supported" error, halting execution. This at least avoids silently doing the wrong thing. However, Miri might not get away with this if there's code out there relying on this to work (which seems not unlikely).

I prefer the status quo over this, even if it would make debugging easier. We could consider emitting a lint though.

>     * Throw such an error in CTFE, but in Miri just remove the provenance of the other bytes. With CTFE/Miri engine Pointer type overhaul #87123 this is actually not terribly wrong, but it is not really right either -- losing provenance like this does not match what LLVM does, where individual bytes of a pointer can maintain their provenance separately.

That seems like we're treating this as UB or at least unspecified behaviour and doing something random that we can change in the future. I don't like it much. It will also cause differences between the runtime and compile-time behaviour of const fns

> Implement support for having just a few bytes of a pointer in a Miri core engine Allocation. However, this might be hard to do without a perf regression.

[...]

PR #87248, opened 2021-07-18
CTFE: throw unsupported error when partially overwriting a pointer

RalfJung:
Currently, during CTFE, when a write to memory would overwrite parts of a pointer, we make the remaining parts of that pointer "uninitialized". This is probably not what users expect, so if this ever happens they will be quite confused about why some of the data just vanishes for seemingly no good reason.
So I propose we change this to abort CTFE when that happens, to at last avoid silently doing the wrong thing.
Cc https://github.com/rust-lang/rust/issues/87184

Our CTFE test suite still seems to pass. However, we should probably crater this, and I want to do some tests with Miri as well.

### DIFF
--- a/compiler/rustc_middle/src/mir/interpret/error.rs
+++ b/compiler/rustc_middle/src/mir/interpret/error.rs
@@ -404,7 +404,7 @@ pub enum UnsupportedOpInfo {
     Unsupported(String),
     /// Encountered a pointer where we needed raw bytes.
     ReadPointerAsBytes,
-    /// Overwriting parts ofa  pointer; the resulting state cannot be represented in our
+    /// Overwriting parts of a pointer; the resulting state cannot be represented in our
     /// `Allocation` data structure.
     PartialPointerOverwrite(Pointer<AllocId>),
     //
--- a/compiler/rustc_middle/src/mir/interpret/error.rs
+++ b/compiler/rustc_middle/src/mir/interpret/error.rs
@@ -404,6 +404,9 @@ pub enum UnsupportedOpInfo {
     Unsupported(String),
     /// Encountered a pointer where we needed raw bytes.
     ReadPointerAsBytes,
+    /// Overwriting parts ofa  pointer; the resulting state cannot be represented in our
+    /// `Allocation` data structure.
+    PartialPointerOverwrite(Pointer<AllocId>),
     //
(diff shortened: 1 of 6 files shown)


RalfJung:
Ah turns out Miri really needs this:
```
error: unsupported operation: unable to overwrite parts of a pointer in memory at alloc1390
   --> /home/r/src/rust/rustc/library/std/src/panicking.rs:401:13
    |
401 |             data.r = ManuallyDrop::new(f());
    |             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ unable to overwrite parts of a pointer in memory at alloc1390
    |
    = help: this is likely not a bug in the program; it indicates that the program performed an operation that the interpreter does not support

    = note: inside `std::panicking::try::do_call::<&dyn std::ops::Fn() -> i32 + std::marker::Sync + std::panic::RefUnwindSafe, i32>` at /home/r/src/rust/rustc/library/std/src/panicking.rs:401:13
    = note: inside `std::panicking::try::<i32, &dyn std::ops::Fn() -> i32 + std::marker::Sync + std::panic::RefUnwindSafe>` at /home/r/src/rust/rustc/library/std/src/panicking.rs:365:19
```

RalfJung:
@craterbot check

oli-obk:
All of the regressions are spurious (some python network error and some file not found)

RalfJung:
Nice, that's what I hoped for. :)

So how do we proceed here -- what are your thoughts on this PR, @oli-obk?

compiler/rustc_middle/src/mir/interpret/error.rs
```diff
@@ -404,6 +404,9 @@ pub enum UnsupportedOpInfo {
     NoMirFor(DefId),
     /// Encountered a pointer where we needed raw bytes.
     ReadPointerAsBytes,
+    /// Overwriting parts ofa  pointer; the resulting state cannot be represented in our
```
oli-obk, line 407:
```suggestion
    /// Overwriting parts of a pointer; the resulting state cannot be represented in our
```

oli-obk (commented):
I like how this allows us to experiment on miri first!

lgtm modulo some tests

RalfJung:
Added a test, please have a look. :)

PR #87254, opened 2021-07-18
LLVM codegen: Don't emit zero-sized padding for fields

hkratz:
Currently padding is emitted before fields of a struct and at the end of the struct regardless of the ABI. Even if no padding is required zero-sized padding fields are emitted. This is not useful and - more importantly - it make it impossible to generate the exact vector types that LLVM expects for certain ARM SIMD intrinsics. This change should unblock the implementation of many ARM intrinsics using the `unadjusted` ABI, see https://github.com/rust-lang/stdarch/issues/1143#issuecomment-827404092.

This is a proof of concept only because the field lookup now takes O(number of fields) time compared to O(1) before since it recalculates the mapping at every lookup. I would like to find out how big the performance impact actually is before implementing caching or restricting this behavior to the `unadjusted` ABI.

cc @SparrowLii @bjorn3

([Discussion on internals](https://internals.rust-lang.org/t/feature-request-add-a-way-in-rustc-for-generating-struct-type-llvm-ir-without-paddings/15007))

### DIFF
--- a/compiler/rustc_codegen_llvm/src/context.rs
+++ b/compiler/rustc_codegen_llvm/src/context.rs
@@ -79,6 +79,11 @@ pub struct CodegenCx<'ll, 'tcx> {
     pub pointee_infos: RefCell<FxHashMap<(Ty<'tcx>, Size), Option<PointeeInfo>>>,
     pub isize_ty: &'ll Type,
 
+    /// Cache for the mapping from source index to llvm index for struct fields,
+    /// necessary because the mapping depends on padding and thus depens on
+    /// TyAndLayout.
+    pub field_projection_cache: RefCell<FxHashMap<TyAndLayout<'tcx>, Vec<u32>>>,
+
     pub coverage_cx: Option<coverageinfo::CrateCoverageContext<'ll, 'tcx>>,
     pub dbg_cx: Option<debuginfo::CrateDebugContext<'ll, 'tcx>>,
 
@@ -308,6 +313,7 @@ impl<'ll, 'tcx> CodegenCx<'ll, 'tcx> {
             scalar_lltypes: Default::default(),
             pointee_infos: Default::default(),
             isize_ty,
+            field_projection_cache: Default::default(),
             coverage_cx,
             dbg_cx,
             eh_personality: Cell::new(None),
--- a/compiler/rustc_codegen_llvm/src/context.rs
+++ b/compiler/rustc_codegen_llvm/src/context.rs
@@ -80,8 +80,7 @@ pub struct CodegenCx<'ll, 'tcx> {
     pub isize_ty: &'ll Type,
 
     /// Cache for the mapping from source index to llvm index for struct fields,
-    /// necessary because the mapping depends on padding and thus depens on
-    /// TyAndLayout.
+    /// only present if synthetic fields are inserted for padding.
     pub field_projection_cache: RefCell<FxHashMap<TyAndLayout<'tcx>, Vec<u32>>>,
 
     pub coverage_cx: Option<coverageinfo::CrateCoverageContext<'ll, 'tcx>>,
--- a/compiler/rustc_codegen_llvm/src/context.rs
+++ b/compiler/rustc_codegen_llvm/src/context.rs
@@ -103,7 +103,7 @@ pub struct TypeLowering<'ll> {
 
     /// If padding is used the slice maps fields from source order
     /// to llvm order.
-    pub field_remapping: Option<Box<SmallVec<[u32; 4]>>>,
+    pub field_remapping: Option<SmallVec<[u32; 4]>>,
 }
 
 fn to_llvm_tls_model(tls_model: TlsModel) -> llvm::ThreadLocalMode {
--- a/compiler/rustc_codegen_llvm/src/context.rs
+++ b/compiler/rustc_codegen_llvm/src/context.rs
@@ -24,6 +24,7 @@ use rustc_span::source_map::{Span, DUMMY_SP};
 use rustc_span::symbol::Symbol;
 use rustc_target::abi::{HasDataLayout, LayoutOf, PointeeInfo, Size, TargetDataLayout, VariantIdx};
 use rustc_target::spec::{HasTargetSpec, RelocModel, Target, TlsModel};
+use smallvec::SmallVec;
 
 use std::cell::{Cell, RefCell};
 use std::ffi::CStr;
@@ -102,7 +103,7 @@ pub struct TypeLowering<'ll> {
 
     /// If padding is used the slice maps fields from source order
     /// to llvm order.
-    pub field_remapping: Option<Box<[u32]>>,
+    pub field_remapping: Option<Box<SmallVec<[u32; 4]>>>,
 }
 
 fn to_llvm_tls_model(tls_model: TlsModel) -> llvm::ThreadLocalMode {
--- a/compiler/rustc_codegen_llvm/src/context.rs
+++ b/compiler/rustc_codegen_llvm/src/context.rs
@@ -74,14 +74,14 @@ pub struct CodegenCx<'ll, 'tcx> {
     /// See <https://llvm.org/docs/LangRef.html#the-llvm-used-global-variable> for details
     pub used_statics: RefCell<Vec<&'ll Value>>,
 
-    pub lltypes: RefCell<FxHashMap<(Ty<'tcx>, Option<VariantIdx>), &'ll Type>>,
+    /// Mapping of non-scalar types to llvm types and field remapping if needed.
+    pub type_lowering: RefCell<FxHashMap<(Ty<'tcx>, Option<VariantIdx>), TypeLowering<'ll>>>,
+
+    /// Mapping of scalar types to llvm types.
     pub scalar_lltypes: RefCell<FxHashMap<Ty<'tcx>, &'ll Type>>,
+
     pub pointee_infos: RefCell<FxHashMap<(Ty<'tcx>, Size), Option<PointeeInfo>>>,
     pub isize_ty: &'ll Type,
-
-    /// Cache for the mapping from source index to llvm index for struct fields,
-    /// only present if synthetic fields are inserted for padding.
-    pub field_projection_cache: RefCell<FxHashMap<TyAndLayout<'tcx>, Vec<u32>>>,
 
     pub coverage_cx: Option<coverageinfo::CrateCoverageContext<'ll, 'tcx>>,
     pub dbg_cx: Option<debuginfo::CrateDebugContext<'ll, 'tcx>>,
@@ -94,6 +94,15 @@ pub struct CodegenCx<'ll, 'tcx> {
 
     /// A counter that is used for generating local symbol names
     local_gen_sym_counter: Cell<usize>,
+}
+
+pub struct TypeLowering<'ll> {
+    /// Associated LLVM type
+    pub lltype: &'ll Type,
+
+    /// If padding is used the slice maps fields from source order
+    /// to llvm order.
+    pub field_remapping: Option<Box<[u32]>>,
 }
 
 fn to_llvm_tls_model(tls_model: TlsModel) -> llvm::ThreadLocalMode {
@@ -308,11 +317,10 @@ impl<'ll, 'tcx> CodegenCx<'ll, 'tcx> {
             const_globals: Default::default(),
             statics_to_rauw: RefCell::new(Vec::new()),
             used_statics: RefCell::new(Vec::new()),
-            lltypes: Default::default(),
+            type_lowering: Default::default(),
             scalar_lltypes: Default::default(),
             pointee_infos: Default::default(),
             isize_ty,
-            field_projection_cache: Default::default(),
             coverage_cx,
             dbg_cx,
             eh_personality: Cell::new(None),
--- a/compiler/rustc_codegen_llvm/src/type_of.rs
+++ b/compiler/rustc_codegen_llvm/src/type_of.rs
@@ -99,6 +99,7 @@ fn struct_llfields<'a, 'tcx>(
     let mut prev_effective_align = layout.align.abi;
     let mut result: Vec<_> = Vec::with_capacity(1 + field_count * 2);
     let mut projection = vec![0; field_count];
+    let mut padding_used = false;
     for i in layout.fields.index_by_increasing_offset() {
         let target_offset = layout.fields.offset(i as usize);
         let field = layout.field(cx, i);
@@ -118,6 +119,7 @@ fn struct_llfields<'a, 'tcx>(
         assert!(target_offset >= offset);
         let padding = target_offset - offset;
         if padding != Size::ZERO {
+            padding_used = true;
             let padding_align = prev_effective_align.min(effective_field_align);
             assert_eq!(offset.align_to(padding_align) + padding, target_offset);
             result.push(cx.type_padding_filler(padding, padding_align));
@@ -145,7 +147,9 @@ fn struct_llfields<'a, 'tcx>(
     } else {
         debug!("struct_llfields: offset: {:?} stride: {:?}", offset, layout.size);
     }
-    cx.field_projection_cache.borrow_mut().insert(layout, projection);
+    if padding_used {
+        cx.field_projection_cache.borrow_mut().insert(layout, projection);
+    }
 
     (result, packed)
 }
@@ -361,9 +365,7 @@ impl<'tcx> LayoutLlvmExt<'tcx> for TyAndLayout<'tcx> {
 
             FieldsShape::Arbitrary { .. } => match cx.field_projection_cache.borrow().get(self) {
                 Some(projection) => projection[index] as u64,
-                None => {
-                    bug!("TyAndLayout::llvm_field_index({:?}): field projection not cached", self)
-                }
+                None => self.fields.memory_index(index) as u64,
             },
         }
     }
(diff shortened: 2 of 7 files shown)


hkratz:
I will add a testcase to make sure that this actually makes it possible to fix https://github.com/rust-lang/stdarch/issues/1143 and implement caching of the projections. Once done I will request another perf run.

bjorn3:
@bors try @rust-timer queue

per the request on zulip

SparrowLii:
Thanks a lot for submitting this ^@^ I think the blocking in stdarch can finally be solved

kornelski:
Number of fields is usually small, so try using a smallvec.

hkratz:
@kornelski:

1. As far as I can see smallvec data structures are not used at all in rustc so far, maybe I am missing something. I am also not even sure a smallvec would make much of a difference here.

   (Aside: What I would like to look at is a replacement datastructure for Vec<u/isize> optimized for small values, taking up only only 1, 2, 4 bytes per value as long as all values fit (initial value range hinted at creation) reallocating as necessary and that with smallvec optimization as well (e.g. 64 bytes which could be used for 64u8, 32 u16, 16 u32, 8 u64). From first glance quite a few data structures in the compiler could benefit from that.)
 
2. Not storing the projection in case that there is no padding sounds like a good idea. Will look into that.

3. Also I am not sure yet why there is such a relatively large slowdown for this particular testcase. Will do some profiling.

klensy:
@hkratz You searched bad, even github will find usages of smallvec https://github.com/rust-lang/rust/search?q=smallvec .

hkratz:
> @hkratz You searched bad, even github will find usages of smallvec https://github.com/rust-lang/rust/search?q=smallvec .

No idea how I missed that... Not sure if it will improve performance though either. Will dig some more.

kornelski:
The idea behind using smallvec is to avoid indirection and heap allocations. I'm not entirely sure if it will help, so it'd be nice to try with and without it. 

`SmallVec` is minimum 32-bytes, and at this size stores max 5 u32 elements, so if you use `SmallVec::<[u32; 5]>` it will store indices inline in the hashmap for structs with 5 fields or fewer. If it happens that Rust uses many of such structs, then it should help. But maybe struts that do use padding tend to be larger anyway.

hkratz:
The additional time for the deep-vector-opt is not directly related to the changed code. Instead the thin-local LTO apparently takes longer for the emitted code without zero-sized padding:

With thin-local LTO (default): 4% more time taken
```
$ hyperfine -w1 -r3 -s basic "~/rusttest/master/bin/rustc -C opt-level=3 src/main.rs" "~/rusttest/new/bin/rustc -C opt-level=3 src/main.rs"
Benchmark #1: ~/rusttest/master/bin/rustc -C opt-level=3 src/main.rs
  Time (mean ± σ):      5.123 s ±  0.009 s    [User: 5.015 s, System: 0.196 s]
  Range (min … max):    5.113 s …  5.131 s    3 runs

Benchmark #2: ~/rusttest/new/bin/rustc -C opt-level=3 src/main.rs
  Time (mean ± σ):      5.319 s ±  0.007 s    [User: 5.198 s, System: 0.212 s]
  Range (min … max):    5.313 s …  5.328 s    3 runs

Summary
  '~/rusttest/master/bin/rustc -C opt-level=3 src/main.rs' ran
    1.04 ± 0.00 times faster than '~/rusttest/new/bin/rustc -C opt-level=3 src/main.rs'
```

With thin-local LTO disabled:
```
$ hyperfine -w1 -r3 -s basic "~/rusttest/master/bin/rustc -C opt-level=3 -C codegen-units=1 src/main.rs" "~/rusttest/new/bin/rustc -C opt-level=3 -C codegen-units=1 src/main.rs"
Benchmark #1: ~/rusttest/master/bin/rustc -C opt-level=3 -C codegen-units=1 src/main.rs
  Time (mean ± σ):      3.841 s ±  0.007 s    [User: 3.721 s, System: 0.128 s]
  Range (min … max):    3.834 s …  3.847 s    3 runs

Benchmark #2: ~/rusttest/new/bin/rustc -C opt-level=3 -C codegen-units=1 src/main.rs
  Time (mean ± σ):      3.847 s ±  0.019 s    [User: 3.722 s, System: 0.133 s]
  Range (min … max):    3.829 s …  3.867 s    3 runs

Summary
  '~/rusttest/master/bin/rustc -C opt-level=3 -C codegen-units=1 src/main.rs' ran
    1.00 ± 0.01 times faster than '~/rusttest/new/bin/rustc -C opt-level=3 -C codegen-units=1 src/main.rs'
```

I have no idea why that is the case but I don't think that it should block this PR.

hkratz:
@eddyb Do you want to have a look? According to git blame you worked on this part of the backend before.

SparrowLii:
I really want to help, but my understanding of the compiler is not enough to make meaningful comments. 
I am trying my best to learn the relevant code.

tmiasko:
The `emit_aapcs_va_arg` uses hard coded field indices with `struct_gep`, those probably need to be updated as well.

compiler/rustc_codegen_llvm/src/context.rs
```diff
@@ -79,6 +79,10 @@ pub struct CodegenCx<'ll, 'tcx> {
     pub pointee_infos: RefCell<FxHashMap<(Ty<'tcx>, Size), Option<PointeeInfo>>>,
     pub isize_ty: &'ll Type,
 
+    /// Cache for the mapping from source index to llvm index for struct fields,
+    /// only present if synthetic fields are inserted for padding.
+    pub field_projection_cache: RefCell<FxHashMap<TyAndLayout<'tcx>, Vec<u32>>>,
```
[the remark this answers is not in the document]
hkratz: Hmm, storing it in the same cache definitely makes sense, but for now maybe just adding an `Option<Box<Vec<u32>>>` for the field projection would be enough?

As a tuple would be less self-documenting maybe replacing `lltypes` with `type_datas` like this:
```rs
pub struct CodegenCx<'ll, 'tcx> {
    // ...
    pub type_datas: RefCell<FxHashMap<(Ty<'tcx>, Option<VariantIdx>), TypeData<'ll>>>
    // ...
}

pub struct TypeData<'ll> {
    pub lltype: &'ll Type,
    pub field_projection: Option<Box<Vec<u32>>>,
}
```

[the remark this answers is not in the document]
eddyb: You can use `Option<Box<[u32]>>`, tho I suppose if it's too performance critical, `Box<Vec<` could also make sense.

Name-wise (tho I'm not too happy with any of my suggestions), I'd use `LoweredType` / `TypeLowering` instead of `TypeData`, and something like `field_remapping` / `field_renumbering` instead of `field_projection`.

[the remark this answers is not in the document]
hkratz: I went with the boxed slice, `TypeLowering` and `field_remapping`. The code is not performance critical. The performance regression in the `deep-vector-opt` test comes from thin LTO inexplicably taking a bit longer after the change.

hkratz:
> The `emit_aapcs_va_arg` uses hard coded field indices with `struct_gep`, those probably need to be updated as well.

@tmiasko Thanks for spotting that. I have fixed that and now also replaced the hard-coded llvm indexes with lookups. The order of the fields in `VaListImpl` cannot change as the struct is `#[repr(C)]` and the memory order is mandated by FFI ABI compatibility, so the lookups with source indexes should be future-proof.

estebank:
I'm ok with the code, but the backend is not my area of expertise. @eddyb, could you take over final r+?

Running perf one more time for completeness. I would also like to see what the perf is using `SmallVec` instead of `Vec` in this file (but that's out of scope here).

@bors try @rust-timer queue

Looking at the previous runs it seems that the "slowness" is introduced in finish_ongoing_codegen and LLVM_passes, which were already the slowest part of the compilation for deep-vector-opt.

SparrowLii:
Can we modify the `memory_order`([here](https://github.com/rust-lang/rust/blob/master/compiler/rustc_target/src/abi/mod.rs#L747)) in Layout' FieldShape to express the index in llvm_type? Such as memory_index: `Vec<(u32, u32)>` , with the second number indicating the index of the item in LLVM IR.
This can avoid adding additional `Vec` `SmallVec` or `Box` and laboriously researching his impact on performance

hkratz:
> Can we modify the `memory_order`([here](https://github.com/rust-lang/rust/blob/master/compiler/rustc_target/src/abi/mod.rs#L747)) in `Layout`' `FieldsShape` to express the index in llvm_type? Such as `memory_index: Vec<(u32, u32)>` , with the second number indicating the index in the llvm_type.
> This can avoid adding additional `Vec` `SmallVec` or `Box` and laboriously researching its impact on performance IMO

I don't think that would be a good idea as this information is created in a different layer, has very clear semantics and is not expected to be changed once created. 

As for performance speed-wise the change is a slight overall win (likely due to slightly simple LLVM IR emitted) except for the synthetic `deep-vector-opt` benchmark, where it increases the thin-local LTO time for some non-obvious reason. 

Memory-wise I am a bit concerned because the latest revision of the PR uses more than previously due to entries in the ex-`lltypes` now the`type_lowering` hashmap taking 16 bytes instead of 8 for every entry even if no padding is necessary. And only about ~10% actually need a projection. The separate map was much more memory-efficient.

SparrowLii:
> I don't think that would be a good idea as this information is created in a different layer, has very clear semantics and is not expected to be changed once created.

In the process of calculating `memory_order`, `offsets` will also be calculated (which determines whether there are paddings). So we can correctly predict the actual index in the llvm_type, and no longer change it afterwards. 

Just an idea, maybe I should discuss it in a new PR. I think the current implementation is good enough, and hope it can be merged as soon as possible. So we can solve the blocking in stdarch ^_^

eddyb:
> where it increases the thin-local LTO time for some non-obvious reason

If I had to guess, maybe the old "more explicit" lowering was interfering with some ThinLTO heuristics?
cc @Mark-Simulacrum do we have a policy on "likely caused LLVM to spend more time optimizing"?

> I don't think that would be a good idea as this information is created in a different layer, has very clear semantics and is not expected to be changed once created.

I agree, and if we can remove `memory_order` from `rustc_target::abi::Layout` eventually, that would be good.

<details>
<summary>(click to open digression)</summary><br>

It's fundamental to Rust that *memory is untyped*, and LLVM making us use types to express *byte offsets* is an idiosyncrasy specific to LLVM (that better-designed IRs, such as Cranelift, don't share) - AFAIK it's not even a "C strict aliasing" thing, since TBAA is done with separate metadata from LLVM types - it's just LLVM not starting out with a good design.

The move to opaque pointers in LLVM should *eventually* alleviate the need for these silly "type" games, but we're not there yet, and many years ago when I wanted to help with opaque pointers, I received pushback on removing LLVM "types" from more positions where they're just a waste of time (such as constants - that design, coincidentally, led to the "abstract allocation" approach used in miri, but also is much closer to how linkers work).

</details>

compiler/rustc_codegen_llvm/src/type_of.rs
```diff
@@ -90,14 +93,16 @@ fn uncached_llvm_type<'a, 'tcx>(
...
-) -> (Vec<&'a Type>, bool) {
+) -> (Vec<&'a Type>, bool, Option<Box<[u32]>>) {
     debug!("struct_llfields: {:#?}", layout);
     let field_count = layout.fields.count();
 
     let mut packed = false;
     let mut offset = Size::ZERO;
     let mut prev_effective_align = layout.align.abi;
     let mut result: Vec<_> = Vec::with_capacity(1 + field_count * 2);
+    let mut projection = vec![0; field_count];
```
eddyb, line 104: Can you rename `projection` to `field_remapping`?

Also, if you want to, you can avoid allocating *until* you hit `padding_used = true` (and then you don't even need the separate `bool`).

All you have to do is to initialize it to `Vec::with_capacity(field_count)` + `.extend(0..i)` (when the `i`th field needs padding and it hasn't been initialized yet), after which you just push to it if it's `Some`.

compiler/rustc_codegen_llvm/src/type_of.rs
```diff
@@ -247,24 +256,29 @@ impl<'tcx> LayoutLlvmExt<'tcx> for TyAndLayout<'tcx> {
...
         if let Some((llty, layout)) = defer {
-            let (llfields, packed) = struct_llfields(cx, layout);
-            cx.set_struct_body(llty, &llfields, packed)
+            let (llfields, packed, new_field_remapping) = struct_llfields(cx, layout);
+            cx.set_struct_body(llty, &llfields, packed);
+            field_remapping = new_field_remapping;
         }
-
+        cx.type_lowering.borrow_mut().get_mut(&(self.ty, variant_index)).unwrap().field_remapping =
+            field_remapping;
```
eddyb, line 281: You can move this into the `if let Some(...) = defer {...}` block, if you use the original `field_remapping` in the unconditional insert.

compiler/rustc_codegen_llvm/src/type_of.rs
```diff
@@ -354,7 +368,23 @@ impl<'tcx> LayoutLlvmExt<'tcx> for TyAndLayout<'tcx> {
...
+                let variant_index = match self.variants {
+                    Variants::Single { index } => Some(index),
+                    _ => None,
+                };
+
+                // Look up llvm field if indexes do not match memory order due to padding. If
+                // `field_remapping` is `None` no padding was used and the llvm field index
+                // matches the memory index.
+                match cx.type_lowering.borrow().get(&(self.ty, variant_index)) {
+                    Some(TypeLowering { field_remapping: Some(ref prj), .. }) => prj[index] as u64,
```
eddyb, line 381: `prj` -> `remap`

eddyb (changes requested):

compiler/rustc_codegen_llvm/src/type_of.rs
hkratz: I did the rename.

Avoiding the allocation like this does not work though. We are iterating over the fields in memory order, not in source order, so `.extend(0..i)` does not generally give correct results.

hkratz:
I added a commit using a `SmallVec<[u32; 4]>` for the field projection instead of the slice. That should be a definite win as the vast majority of structs have < 5 fields. 

Number of types with remapping for the `webrender-Debug-Full` testcase, others are similar:
|#Fields|Count|
|---|---|
|1|2014|
|2|83|
|3|159|
|4|11|
|other|45|

compiler/rustc_codegen_llvm/src/context.rs
```diff
@@ -92,6 +97,15 @@ pub struct CodegenCx<'ll, 'tcx> {
     local_gen_sym_counter: Cell<usize>,
 }
 
+pub struct TypeLowering<'ll> {
+    /// Associated LLVM type
+    pub lltype: &'ll Type,
+
+    /// If padding is used the slice maps fields from source order
+    /// to llvm order.
+    pub field_remapping: Option<Box<SmallVec<[u32; 4]>>>,
```
eddyb, line 106: Another thing you can try (after this perf run) is removing the `Box` - the size of `TypeLowering` might not matter as much as not introducing additional allocations in the case where `field_remapping` is needed.

compiler/rustc_codegen_llvm/src/type_of.rs
```diff
@@ -116,33 +121,37 @@ fn struct_llfields<'a, 'tcx>(
...
+                padding, offset, layout.size
+            );
+            result.push(cx.type_padding_filler(padding, padding_align));
+        }
     } else {
         debug!("struct_llfields: offset: {:?} stride: {:?}", offset, layout.size);
     }
 
-    (result, packed)
+    (result, packed, padding_used.then_some(Box::new(field_remapping)))
```
eddyb, line 154: This always creates the `Box` - also, I don't really like the `bool -> Option` methods, they seem less clear than an `if`-`else` (unless you can remove the `bool` entirely, which in this case you kind of can, with the suggestion from one of my other comments).

hkratz: Ah you are right, will fix this right away... About removing the bool... that is not easily possible see my reply above.

eddyb:
[Difference between the last two runs](https://perf.rust-lang.org/compare.html?start=fb7c64a48d0520eb0f3ec8bd1dd3d03b3aa19bd9&end=72c47ec929f2b7caa1eca0bafc399154afb630b1) - haven't dug into it, but it seems mostly insignificant.

hkratz:
> [Difference between the last two runs](https://perf.rust-lang.org/compare.html?start=fb7c64a48d0520eb0f3ec8bd1dd3d03b3aa19bd9&end=72c47ec929f2b7caa1eca0bafc399154afb630b1) - haven't dug into it, but it seems mostly insignificant.

Yeah, seems to be a wash. 

@eddyb, I think this is ready to be merged. If you need need anything else changed please let me know.

eddyb (approved):

Mark-Simulacrum:
@bors retry spurious network

fatal: unable to access 'https://github.com/servo/servo/': GnuTLS recv error (-54): Error in the pull function.

ehuss:
@bors retry

Tests finished successfully hours ago, but bors didn't get triggered on the success.  Presumably due to GitHub incident.

pnkfelix:
Visiting for perf triage. Seems like the regression to deep-vector-opt is anticipated, and acceptable cost to get this in. Other than that, I don't see anything significant in the [perf comparison](https://perf.rust-lang.org/compare.html?start=e8e1b32a7840c07f30c04b252c379a044a73902d&end=47b41b7788a6f85c749049062f1e4eed497cd894&stat=instructions%3Au) for when this landed. Marking as perf-regression-triaged.

PR #87259, opened 2021-07-18
triagebot shortcut config

Llandy3d:
Enable the new triagebot shortcuts as per [#1381/triagebot](https://github.com/rust-lang/triagebot/pull/1381)


nikomatsakis:
@bors r+ rollup

ok, now it has

PR #87260, opened 2021-07-18
Libgccjit codegen

antoyo:
This PR introduces a subtree for a gcc-based codegen backend to the repository, per decision in https://github.com/rust-lang/compiler-team/issues/442. We do not yet expect to ship this backend on nightly or run tests in CI, but we do verify that the backend checks (i.e., `cargo check`) successfully.

Work is expected to progress primarily in https://github.com/rust-lang/rustc_codegen_gcc, with semi-regular upstreaming, like with other subtrees.


LeSeulArtichaut:
Really hyped for this! r? @Mark-Simulacrum

est31:
> The gcc-patches directory is removed in a pending PR so that the patches won't get into this repo.

Not a lawyer, but note that git works in a way that if you clone a given repository, you obtain the entire history. So from my understanding, the GPL patches are still there, and merging *this* PR might add GPL artifacts to each git clone of the rust repo.

est31:
If you *really* want to remove the directory from git, use `git filter-branch` or `git filter-repo`.

bjorn3:
* [ ] In `src/bootstrap/check.rs` and `src/bootstrap/compile.rs` there is a `CodegenBackend` struct. Currently it only compiles cg_clif, but it is designed to work with arbitrary codegen backends. The check one is probably the easiest to wire up. Just add `compiler/rustc_codegen_gcc` and `rustc_codegen_gcc` to the array in `should_run` and add `"gcc"` to the array in `make_run`. For compile you need something similar, but also wire up compilation of libgccjit somehow. Maybe that could be left for a future PR?
* [ ] Next you need to explicitly add or exclude rustc_codegen_gcc from the root cargo workspace. I explicitly excluded rustc_codegen_cranelift to ensure that the `Cargo.lock` of cg_clif is used instead of the root `Cargo.lock` to determine the right rev of the cranelift git dependency. I think you will want to do the same thing because of the `gccjit.rs` git dependency.
* [ ] Finally you will either need to run rustfmt on cg_gcc using rustc's `rustfmt.toml` and a nightly rustfmt or exclude it in `rustfmt.toml`.
* [ ] Another thing, once this PR (or any other PR updating the cg_gcc version in this repo) is merged you will want to pull the git subtree to the cg_gcc repo to prevent merge conflicts the next time you pull changes from the rust repo. See my rustup script: https://github.com/bjorn3/rustc_codegen_cranelift/blob/bd2f72f398fa6b609a005cea8ebe16e093937936/scripts/rustup.sh#L29-L53

joshtriplett:
> If you _really_ want to remove the directory from git, use `git filter-branch` or `git filter-repo`.

Or, as another option, you could squash the current git history and just add rustc_codegen_gcc in one commit. (You don't *have* to do this, but it may be simpler, and I don't think it's critical to preserve the out-of-tree history. New commits could start from the merged version.)

antoyo:
So, the CI now pass.

I still need to fix the codegen as the latest changes in rustc made the libcore's test failing.

What else needs to be done for this first PR to land?

Should I run the libcore's test in the CI?
If so, how to a get a patched libgccjit (do I use the binary from the artifacts in [my fork of gcc](https://github.com/antoyo/gcc), do I build it in the CI — if so, from my fork or do I need to apply the paches?)?

Apart from that, from the comments above, I believe only the changes to `src/bootstrap/compile.rs` that @bjorn3 mentioned is missing (should that be included in this PR or a later one?).

joshtriplett:
> Should I run the libcore's test in the CI?

Per the [MCP](https://github.com/rust-lang/compiler-team/issues/442):
> This MCP also proposes gating CI on `rustc_codegen_gcc` building, but *not* on it passing any tests.

So, you shouldn't run the testsuite, but you *should* build the backend.

Long-term, I think we'll want to have a `rust-lang/gcc` repository that contains a branch with the necessary patches, just as we currently have an LLVM branch at https://github.com/rust-lang/llvm-project/ .

I'd suggest checking with @rust-lang/infra about the best way to handle this. I'll check on Zulip.

antoyo:
> So, you shouldn't run the testsuite, but you _should_ build the backend.

@joshtriplett : By that, do you mean that doing the `check` is enough, or should I attempt to build it and link to libgccjit?

Have you got any news from asking on Zulip?

xero-lib:
Very excited to see this in action

antoyo:
@LeSeulArtichaut Can you tell me what is blocking this PR? (The GCC patches were removed from the history, btw.)

glaubitz:
> @LeSeulArtichaut Can you tell me what is blocking this PR? (The GCC patches were removed from the history, btw.)

Shouldn't the commits above be squashed into a single commit? Seems a bit unusual to keep the history of the development project.

antoyo:
> > @LeSeulArtichaut Can you tell me what is blocking this PR? (The GCC patches were removed from the history, btw.)
> 
> Shouldn't the commits above be squashed into a single commit? Seems a bit unusual to keep the history of the development project.

I tried doing that, but it stopped being a `git subtree`. Do you know if there is a way to squash a `git subtree` without creating an unrelated git history?

est31:
@glaubitz same has been done for clippy, miri, rustfmt, rustc-codegen-cranelift, so it's not without precedent. And in the case of the other repos it were way more commits than just 90.

glaubitz:
> @glaubitz same has been done for clippy, miri, rustfmt, rustc-codegen-cranelift, so it's not without precedent. And in the case of the other repos it was way more commits than just 90.

OK, I wasn't aware of that. I guess it's fine as is then. Thanks for the clarification.

est31:
> The GCC patches were removed from the history, btw.

Indeed they are gone now. According to the history, the old repo state was under a GPL 3.0 license. It depends on how you interpret the commit that changed the license to MIT OR Apache-2.0, whether it applies to old state too, or only to the state of the commit as well as the changes after that. Not a lawyer but I don't think this needs a squash or a history edit though, just a simple statement from @antoyo would be enough, e.g. with a comment on github. E.g.:

> I license my contributions in this PR under the dual MIT/Apache-2.0 license, allowing licensees to chose either at their option.

*generally* this is implied when you file a PR to the rust-lang/rust repo, however when that PR contains the GPL 3.0 license in a dedicated file, it might overrule that.

antoyo:
@LeSeulArtichaut I asked you because you added the `S-blocked` tag. So I wondered what was the reason.

LeSeulArtichaut:
Oh, that was for the MCP but it's finished now.

Mark-Simulacrum:
It looks like this was intended to be [squashed](https://github.com/antoyo/rustc_codegen_gcc/pull/51#issuecomment-883576784), but that hasn't happened yet - do you need help making that happen? I would expect the upstream repository to also need to get squashed.

I likely have some feedback on the contents of the PR but want to make sure I'm reviewing the right thing -- for example, as best as I can tell, this still contains the NOTICE file. I have a few high-level questions to start -- apologies if these have been answered elsewhere, feel free to link me to comments or so. I don't necessarily expect us to resolve all of these prior to merging this, but do want to start getting a rough idea of the plan here before we move ahead with merging this in.

* Are we intending the backend to be usable with an unpatched libgccjit in the future (i.e., are those patches being upstreamed)?
* Has there been any discussion with @bjorn3 (or perhaps other codegen backends, not in tree) about the shared bits between this and cranelift? In particular, I noticed some (presumed) duplication with the cranelift backend in terms of the sysroot and such. I wonder if it makes sense for some coordination work to be done between cranelift and gcc to unify the testing strategy (perhaps using compiletest). This will become especially relevant as we look to start running tests.
* It looks like there's a not-trivial number of commented code and FIXMEs left throughout the codebase -- I don't think this is a blocker, but I'm a little worried that it may make it harder for folks trying to fix compilation failures to navigate and understand the structure. Since we're gating on compilation succeeding out of the gate, this seems important to address.
* I suspect we'll want to move the external part of the tree to rust-lang/rustc_codegen_gcc -- we'll likely want to do the same for the cranelift backend (I missed this when adding that in, I guess).

Thanks for kicking this work off -- I've given things a quick skim so far but will definitely do a more thorough review once these questions are answered.

bjorn3:
> Are we intending the backend to be usable with an unpatched libgccjit in the future (i.e., are those patches being upstreamed)?

I believe so.

> Has there been any discussion with @bjorn3 (or perhaps other codegen backends, not in tree) about the shared bits between this and cranelift?

Not really.

> In particular, I noticed some (presumed) duplication with the cranelift backend in terms of the sysroot and such.

This code was originally adapted from cg_clif, but I somewhat recently rewrote most of it in rust at the cg_clif side to be able to support windows.

> I wonder if it makes sense for some coordination work to be done between cranelift and gcc to unify the testing strategy (perhaps using compiletest). This will become especially relevant as we look to start running tests.

At the cg_clif side I haven't ported the testing infrastructure to rust yet. Using compiletest won't cover everything as it doesn't just run self-contained tests, but also some real projects in a crater like fashion. I am also running the libcore tests, a test to bootstrap rustc itself and the rustc test suite with all tests using unsupported features disabled. I do think there is value in somewhat unifying the testing and wiring it up in `./x.py`, but it can't be completely unified due to the different levels of completeness between the various codegen backends.

> I likely have some feedback on the contents of the PR but want to make sure I'm reviewing the right thing -- for example, as best as I can tell, this still contains the NOTICE file.

I requested this file in https://github.com/antoyo/rustc_codegen_gcc/issues/1 when cg_gcc was still GPL licensed. I am fine with removing it now.

antoyo:
> It looks like this was intended to be [squashed](https://github.com/antoyo/rustc_codegen_gcc/pull/51#issuecomment-883576784), but that hasn't happened yet - do you need help making that happen? I would expect the upstream repository to also need to get squashed.

It [was said](https://github.com/rust-lang/rust/pull/87260#issuecomment-891884209) other repos that were included were not squashed, but if you prefer, I can squash upstream to make the `git subtree` work.

> * Are we intending the backend to be usable with an unpatched libgccjit in the future (i.e., are those patches being upstreamed)?
Yes, the first patch was upstream. It's gonna take time, but it will happen :) .

> * Has there been any discussion with @bjorn3 (or perhaps other codegen backends, not in tree) about the shared bits between this and cranelift? In particular, I noticed some (presumed) duplication with the cranelift backend in terms of the sysroot and such. I wonder if it makes sense for some coordination work to be done between cranelift and gcc to unify the testing strategy (perhaps using compiletest). This will become especially relevant as we look to start running tests.

That was answered in the previous comment.

> * It looks like there's a not-trivial number of commented code and FIXMEs left throughout the codebase -- I don't think this is a blocker, but I'm a little worried that it may make it harder for folks trying to fix compilation failures to navigate and understand the structure. Since we're gating on compilation succeeding out of the gate, this seems important to address.

I'll remove the commented code; I probably forgot this.
For the FIXMEs, it'll be harder (well, I'll look again at them to make sure): some of them will require a change in rustc to address and the rest will require features to be implemented.

> * I suspect we'll want to move the external part of the tree to rust-lang/rustc_codegen_gcc -- we'll likely want to do the same for the cranelift backend (I missed this when adding that in, I guess).

You mean, the `git subtree` will point to `rust-lang/rustc_codegen_gcc` instead of `antoyo/rustc_codegen_gcc`?

Mark-Simulacrum:
> It was said other repos that were included were not squashed, but if you prefer, I can squash upstream to make the git subtree work.

Yes, I think in other cases there was less worry about past artifacts/commits having anything that we're trying to delete in them. I'd prefer to see this squashed (and upstream as well, due to how subtrees work).

> You mean, the git subtree will point to rust-lang/rustc_codegen_gcc instead of antoyo/rustc_codegen_gcc?

Yeah, on a technical level. There's some other things implied by the repository being in-org (e.g., access control policies, infrastructure team access, etc.) but those are hopefully largely just details.

bjorn3:
> Yes, I think in other cases there was less worry about past artifacts/commits having anything that we're trying to delete in them. I'd prefer to see this squashed (and upstream as well, due to how subtrees work).

If you squash please keep the old commit as a tag or branch on antoyo/rustc_codegen_gcc.

antoyo:
> If you squash please keep the old commit as a tag or branch on antoyo/rustc_codegen_gcc.

Good idea. Any specific reason on your side?

bjorn3:
Mostly to preserve history. Say you ever need to run `git blame` on code that already existed before the squash.

compiler/rustc_codegen_gcc/build_sysroot/prepare_sysroot_src.sh
```diff
@@ -0,0 +1,39 @@
...
+cp -r $SRC_DIR/library $DST_DIR/
+
+pushd $DST_DIR
+echo "[GIT] init"
+git init
+echo "[GIT] add"
+git add .
+echo "[GIT] commit"
+
+# This is needed on virgin system where nothing is configured.
```
kw-fn, line 24:
```suggestion
# This is needed on systems where nothing is configured.
```

compiler/rustc_codegen_gcc/config.sh
```diff
@@ -0,0 +1,48 @@
...
+      RUN_WRAPPER='qemu-aarch64 -L /usr/aarch64-linux-gnu'
+   else
+      echo "Unknown non-native platform"
+   fi
+fi
+
+export RUSTFLAGS=$linker' -Cpanic=abort -Cdebuginfo=2 -Zpanic-abort-tests -Zcodegen-backend='$(pwd)'/target/'$CHANNEL'/librustc_codegen_gcc.'$dylib_ext' --sysroot '$(pwd)'/build_sysroot/sysroot'
+#export RUSTFLAGS=$linker' -Cpanic=abort -Cdebuginfo=2 -Zpanic-abort-tests -Zcodegen-backend='$(pwd)'/target/'$CHANNEL'/librustc_codegen_gcc.'$dylib_ext' --sysroot '$(pwd)'/build_sysroot/sysroot -Clto=fat -Cembed-bitcode=yes'
+
+# FIXME remove once the atomic shim is gone
```
kw-fn, line 36:
```suggestion
# FIXME(antoyo) remove once the atomic shim is gone
```

compiler/rustc_codegen_gcc/example/alloc_system.rs
```diff
@@ -0,0 +1,212 @@
...
+            }
+        }
+    }
+    #[cfg(any(target_os = "android",
+              target_os = "hermit",
+              target_os = "redox",
+              target_os = "solaris"))]
+    #[inline]
+    unsafe fn aligned_malloc(layout: &Layout) -> *mut u8 {
+        // On android we currently target API level 9 which unfortunately
```
kw-fn, line 103:
```suggestion
        // On Android we currently target API level 9 which unfortunately
```

compiler/rustc_codegen_gcc/example/dst-field-align.rs
```diff
@@ -0,0 +1,67 @@
...
+    // Test that the field is aligned properly
+    let f : Foo<usize> = Foo { a: 0, b: 11 };
+    assert_eq!(f.b.get(), 11);
+    let ptr1 : *const u8 = &f.b as *const _ as *const u8;
+
+    let f : &Foo<dyn Bar> = &f;
+    let ptr2 : *const u8 = &f.b as *const _ as *const u8;
+    assert_eq!(f.b.get(), 11);
+
+    // The pointers should be the same
```
kw-fn, line 41:
```suggestion
    // Test that the pointers should be the same
```

compiler/rustc_codegen_gcc/example/mini_core_hello_world.rs
```diff
@@ -0,0 +1,424 @@
...
+    }
+
+    {
+        extern {
+            #[linkage = "weak"]
+            static ABC: *const u8;
+        }
+    }
+
+    // TODO: not sure about this assert. ABC is not defined, so should it be really 0?
```
kw-fn, line 256:
```suggestion
    // TODO(antoyo): not sure about this assert. ABC is not defined, so should it be really 0?
```

compiler/rustc_codegen_gcc/example/std_example.rs
```diff
@@ -0,0 +1,279 @@
...
+
+fn main() {
+    let mutex = std::sync::Mutex::new(());
+    let _guard = mutex.lock().unwrap();
+
+    let _ = ::std::iter::repeat('a' as u8).take(10).collect::<Vec<_>>();
+    let stderr = ::std::io::stderr();
+    let mut stderr = stderr.lock();
+
+    // FIXME: this thread panics.
```
kw-fn, line 19:
```suggestion
    // FIXME(antoyo): this thread panics.
```

compiler/rustc_codegen_gcc/config.sh
bjorn3: Technically this was copied from cg_clif, which is written by me. This is no longer necessary for cg_clif, but seemingly it still is for cg_gcc. I don't think there is much value attributing this FIXME to anyone, as someone else may fix it anyway.

compiler/rustc_codegen_gcc/example/std_example.rs
```diff
@@ -0,0 +1,279 @@
...
+    println!("{}", 2.3f32.sqrt());
+    println!("{}", 2.3f32.floor());
+    println!("{}", 2.3f32.ceil());
+    println!("{}", 2.3f32.min(1.0));
+    println!("{}", 2.3f32.max(1.0));
+    println!("{}", 2.3f32.powi(2));
+    println!("{}", 2.3f32.log2());
+    assert_eq!(2.3f32.copysign(-1.0), -2.3f32);
+    println!("{}", 2.3f32.powf(2.0));
+
```
kw-fn, line 50:
```suggestion
    println!({}, 2.3f32.abs());
    println!({}, 2.3f32.ceil());
    println!({}, 2.3f32.exp());
    println!({}, 2.3f32.exp2());
    println!({}, 2.3f32.floor());
    println!({}, 2.3f32.log2());
    println!({}, 2.3f32.max(1.0));
    println!({}, 2.3f32.min(1.0));
    println!({}, 2.3f32.powf(2.0));
    println!({}, 2.3f32.powi(2));
    println!({}, 2.3f32.sqrt());

    assert_eq!(2.3f32.copysign(-1.0), -2.3f32);
    
```

compiler/rustc_codegen_gcc/example/mini_core_hello_world.rs
bjorn3: To resolve the FIXME, yes it should. Weak symbols must be 0 when they are not defined by any other object file.

compiler/rustc_codegen_gcc/example/std_example.rs
bjorn3: That is not valid rust syntax.

glaubitz:
Looking at the initial commit, I'm only seeing some architectures mentioned in the code.

But isn't the point of this backend to also support architectures that the LLVM backend doesn't support such as Alpha?

compiler/rustc_codegen_gcc/patches/0023-core-Ignore-failing-tests.patch
```diff
@@ -0,0 +1,49 @@
...
+--- a/library/core/tests/array.rs
++++ b/library/core/tests/array.rs
+@@ -242,6 +242,7 @@ fn iterator_drops() {
+     assert_eq!(i.get(), 5);
+ }
+ 
++/*
+ // This test does not work on targets without panic=unwind support.
+ // To work around this problem, test is marked is should_panic, so it will
+ // be automagically skipped on unsuitable targets, such as
```
kw-fn, line 24:
```suggestion
 // be automatically skipped on unsuitable targets, such as
```

compiler/rustc_codegen_gcc/src/abi.rs
```diff
@@ -0,0 +1,286 @@
...
+
+use crate::builder::Builder;
+use crate::context::CodegenCx;
+use crate::intrinsic::ArgAbiExt;
+use crate::type_of::LayoutGccExt;
+
+impl<'a, 'gcc, 'tcx> AbiBuilderMethods<'tcx> for Builder<'a, 'gcc, 'tcx> {
+    fn apply_attrs_callsite(&mut self, _fn_abi: &FnAbi<'tcx, Ty<'tcx>>, _callsite: Self::Value) {
+        // TODO
+        //fn_abi.apply_attrs_callsite(self, callsite)
```
kw-fn, line 15:
```suggestion
        // TODO(antoyo):  fn_abi.apply_attrs_callsite(self, callsite)
```

compiler/rustc_codegen_gcc/src/abi.rs
```diff
@@ -0,0 +1,286 @@
...
+            .iter()
+            .flat_map(|option_kind| {
+                option_kind.map(|kind| Reg { kind, size: self.prefix_chunk_size }.gcc_type(cx))
+            })
+            .chain((0..rest_count).map(|_| rest_gcc_unit))
+            .collect();
+
+        // Append final integer
+        if rem_bytes != 0 {
+            // Only integers can be really split further.
```
kw-fn, line 60:
```suggestion
            // Only integers can be split further.
```

compiler/rustc_codegen_gcc/src/abi.rs
```diff
@@ -0,0 +1,286 @@
...
+                    _ => bug!("unsupported float: {:?}", self),
+                }
+            },
+            RegKind::Vector => unimplemented!(), //cx.type_vector(cx.type_i8(), self.size.bytes()),
+        }
+    }
+}
+
+pub trait FnAbiGccExt<'gcc, 'tcx> {
+    // TODO: return a function pointer type instead?
```
kw-fn, line 90:
```suggestion
    // TODO(antoyo): return a function pointer type instead?
```

bjorn3:
> Looking at the initial commit, I'm only seeing some architectures mentioned in the code.
> 
> But isn't the point of this backend to also support architectures that the LLVM backend doesn't support such as Alpha?

Adding new architectures takes more work than simply adding a backend. There also needs to be new target triples, new abi calculation code, maybe library code needs to be adjusted, ... These things should be handled independently from this PR IMO. Especially as cg_gcc isn't even complete enough to be able to test those new parts. For example the abi calculation of rustc isn't even used on already supported targets yet, making it abi incompatible with cg_llvm and even partially C.

compiler/rustc_codegen_gcc/src/abi.rs
```diff
@@ -0,0 +1,286 @@
...
+            Conv::ArmAapcs => llvm::ArmAapcsCallConv,
+            Conv::Msp430Intr => llvm::Msp430Intr,
+            Conv::PtxKernel => llvm::PtxKernel,
+            Conv::X86Fastcall => llvm::X86FastcallCallConv,
+            Conv::X86Intr => llvm::X86_Intr,
+            Conv::X86Stdcall => llvm::X86StdcallCallConv,
+            Conv::X86ThisCall => llvm::X86_ThisCall,
+            Conv::X86VectorCall => llvm::X86_VectorCall,
+            Conv::X86_64SysV => llvm::X86_64_SysV,
+            Conv::X86_64Win64 => llvm::X86_64_Win64,
```
kw-fn, line 183:
```suggestion
            Conv::AmdGpuKernel => llvm::AmdGpuKernel,
            Conv::ArmAapcs => llvm::ArmAapcsCallConv,
            Conv::C | Conv::Rust => llvm::CCallConv,
            Conv::Msp430Intr => llvm::Msp430Intr,
            Conv::PtxKernel => llvm::PtxKernel,
            Conv::X86_64SysV => llvm::X86_64_SysV,
            Conv::X86_64Win64 => llvm::X86_64_Win64,
            Conv::X86Fastcall => llvm::X86FastcallCallConv,
            Conv::X86Intr => llvm::X86_Intr,
            Conv::X86Stdcall => llvm::X86StdcallCallConv,
            Conv::X86ThisCall => llvm::X86_ThisCall,
            Conv::X86VectorCall => llvm::X86_VectorCall,
```

compiler/rustc_codegen_gcc/src/allocator.rs
```diff
@@ -0,0 +1,115 @@
...
+            .map(|(index, typ)| context.new_parameter(None, *typ, &format!("param{}", index)))
+            .collect();
+        let func = context.new_function(None, FunctionType::Exported, output.unwrap_or(void), &args, name, false);
+
+        if tcx.sess.target.options.default_hidden_visibility {
+            //llvm::LLVMRustSetVisibility(func, llvm::Visibility::Hidden);
+        }
+        if tcx.sess.must_emit_unwind_tables() {
+            // TODO
+            //attributes::emit_uwtable(func, true);
```
kw-fn, line 57:
```suggestion
            // TODO(antoyo): attributes::emit_uwtable(func, true);
```

compiler/rustc_codegen_gcc/src/abi.rs
bjorn3: This isn't going to work. A lot of code, including this one, is commented out as it is copy pasted from the llvm backend and needs to be adapted for libgccjit. Simply uncommenting won't help as this refers to llvm, which isn't used by this backend.

compiler/rustc_codegen_gcc/src/allocator.rs
bjorn3: Going through each `TODO` doesn't add much value. Please prefer review comments like "please change all `TODO` into `TODO(antoyo)`.

compiler/rustc_codegen_gcc/src/archive.rs
```diff
@@ -0,0 +1,270 @@
...
+
+impl<'a> ArchiveBuilder<'a> for ArArchiveBuilder<'a> {
+    fn new(sess: &'a Session, output: &Path, input: Option<&Path>) -> Self {
+        use rustc_codegen_ssa::back::link::archive_search_paths;
+        let config = ArchiveConfig {
+            sess,
+            dst: output.to_path_buf(),
+            lib_search_paths: archive_search_paths(sess),
+            use_native_ar: false,
+            // FIXME test for linux and System V derivatives instead
```
kw-fn, line 44:
```suggestion
            // FIXME(antoyo): test for linux and System V derivatives instead
```

bjorn3: @kw-fn please read my reaction to your previous suggestion of this kind.

compiler/rustc_codegen_gcc/src/asm.rs
```diff
@@ -0,0 +1,632 @@
...
+use rustc_span::Span;
+use rustc_target::asm::*;
+
+use crate::builder::Builder;
+use crate::context::CodegenCx;
+use crate::type_of::LayoutGccExt;
+
+impl<'a, 'gcc, 'tcx> AsmBuilderMethods<'tcx> for Builder<'a, 'gcc, 'tcx> {
+    fn codegen_llvm_inline_asm(&mut self, _ia: &LlvmInlineAsmInner, _outputs: Vec<PlaceRef<'tcx, RValue<'gcc>>>, mut _inputs: Vec<RValue<'gcc>>, _span: Span) -> bool {
+        // TODO
```
kw-fn, line 18:
```suggestion
        // TODO(antoyo):
```

compiler/rustc_codegen_gcc/src/asm.rs
```diff
@@ -0,0 +1,632 @@
...
+        let asm_arch = self.tcx.sess.asm_arch.unwrap();
+
+        let intel_dialect =
+            match asm_arch {
+                InlineAsmArch::X86 | InlineAsmArch::X86_64 if !options.contains(InlineAsmOptions::ATT_SYNTAX) => true,
+                _ => false,
+            };
+
+        // Collect the types of output operands
+        // FIXME: we do this here instead of later because of a bug in libgccjit where creating the
```
kw-fn, line 130:
```suggestion
        // FIXME(antoyo): we do this here instead of later because of a bug in libgccjit where creating the
```

glaubitz:
> > Looking at the initial commit, I'm only seeing some architectures mentioned in the code.
> > But isn't the point of this backend to also support architectures that the LLVM backend doesn't support such as Alpha?
> 
> Adding new architectures takes more work than simply adding a backend. There also needs to be new target triples, new abi calculation code, maybe library code needs to be adjusted, ... These things should be handled independently from this PR IMO. Especially as cg_gcc isn't even complete enough to be able to test those new parts. For example the abi calculation of rustc isn't even used on already supported targets yet, making it abi incompatible with cg_llvm and even partially C.

From my own experience with adding new architectures to rustc, it's actually not that much once LLVM has support for a certain backend. But it might be different with libgccjit.

But we will have gccrs in any case which also making good progress :).

bjorn3:
> From my own experience with adding new architectures to rustc, it's actually not that much once LLVM has support for a certain backend. But it might be different with libgccjit.

Sure, it is not a lot of work, but it is also not small enough for this PR IMO, and currently not really testable.

compiler/rustc_codegen_gcc/src/abi.rs
kw-fn: Oh I see sorry

compiler/rustc_codegen_gcc/src/archive.rs
kw-fn: Sorry! I didn't see this, there were way more of these than I was expecting maybe a handful.

compiler/rustc_codegen_gcc/example/std_example.rs
antoyo: What do you mean? This file compiles on rust nightly.

compiler/rustc_codegen_gcc/config.sh
kw-fn: For this isn't it relevant to know who is originating the note for a fix or further work? I think the git blame would attribute the notice but it's possible to still be lost on copy or future file changes. It doesn't seem that hard to 'sign' these flags as it could be useful for a future reader lacking context.

compiler/rustc_codegen_gcc/src/archive.rs
kw-fn: I wish there were a better way to comb a PR and suggest small changes without totally disrupting the discussion. Sorry again.

compiler/rustc_codegen_gcc/src/abi.rs
antoyo: This comment comes from [here](https://github.com/rust-lang/rust/blob/85109e257ac97a0904106cafaf6e014c1d812326/compiler/rustc_codegen_llvm/src/abi.rs#L197).

antoyo:
Ok, I made a big cleanup to remove commented code and attributed my TODO and FIXME to myself. The other TODO were copy/pasted from elsewhere and were thus not assigned to me.

compiler/rustc_codegen_gcc/gcc_path
```diff
@@ -0,0 +1 @@
+/home/bouanto/Ordinateur/Programmation/Projets/gcc-build/build/gcc
```
est31, line 1: What happened here?

antoyo: Oops. My bad. Is there a good way to ignore changes to a commited file in git? I used `git update-index --assume-unchanged gcc_path` but I keep having issues.

antoyo:
Anything else that I need to do here?

Mark-Simulacrum:
Thanks for being patient! This looks good to me, excited to see this land. We'll want to continue iterating on the code itself -- I didn't review all that closely -- but it seems like it's in a good enough state for the initial merge (no obvious blockers to me, anyway), particularly since we're not shipping or trying to test it etc.

@bors r+ rollup=iffy

glaubitz:
> Thanks for being patient! This looks good to me, excited to see this land.

Oh wow, this is finally happening. I'm very excited as well, this will hopefully soon make rustc available on much more targets. \o/

Mark-Simulacrum:
> Thanks!
> 
> What is the process from now on?
> Should I check now that the tests still pass to know if I need to update?

I think syncing on a roughly biweekly cadence in both directions is a good idea; the longer you wait the more chance there is for rebase errors to crop up. Of course, if tests are broken or there's some reason to sync more often, there's nothing inherently stopping you from doing so.

antoyo:
@Mark-Simulacrum Ok, I made the update to today's nightly (some other changes were merged as well), so please review again to get the tests running. Thanks!

glaubitz:
@antoyo Congratulations! Could you point me to instructions on how to use the new backend? I would like to try building Rust for exotic architectures like Alpha, PA-RISC and Itanium now.

bjorn3:
I believe https://github.com/rust-lang/rustc_codegen_gcc/#how-to-build-a-cross-compiling-libgccjit has the instructions.

antoyo:
> @antoyo Congratulations! Could you point me to instructions on how to use the new backend? I would like to try building Rust for exotic architectures like Alpha, PA-RISC and Itanium now.

There are [instructions for building a cross-compiling libgccjit in the readme](https://github.com/rust-lang/rustc_codegen_gcc#how-to-build-a-cross-compiling-libgccjit).
Please note that there are some hacks needed to make this work currently.
But now that this PR has been merged, people can start [adding new targets in rustc](https://rustc-dev-guide.rust-lang.org/building/new-target.html) and the support in the different crates like `libc` and `object`.

glaubitz:
@bjorn3 @antoyo Thanks a lot for all the hints. I will have a look and start working on adding the targets to `libc`. I already have a draft for m68k here: https://github.com/rust-lang/libc/pull/2414

antoyo:
@glaubitz I misread what you wrote. If you try to build the rust compiler with this project, it's likely not going to work. It's still work in progress and not all the test suite pass, so it might either produce a non-working rustc or fail to compile it.

PR #87264, opened 2021-07-19
proc_macro: Add an expand_expr method to TokenStream

mystor:
This feature is aimed at giving proc macros access to powers similar to those used by builtin macros such as `format_args!` or `concat!`. These macros are able to accept macros in place of string literal parameters, such as the format string, as they perform recursive macro expansion while being expanded.

This can be especially useful in many cases thanks to helper macros like `concat!`, `stringify!` and `include_str!` which are often used to construct string literals at compile-time in user code.

For now, this method only allows expanding macros which produce literals, although more expressions will be supported before the method is stabilized.

In earlier versions of this PR, this method exclusively returned `Literal`, and spans on returned literals were stripped of expansion context before being returned to be as conservative as possible about permission leakage. The method's naming has been generalized to eventually support arbitrary expressions, and the context stripping has been removed (https://github.com/rust-lang/rust/pull/87264#discussion_r674863279), which should allow for more general APIs like "format_args_implicits" (https://github.com/rust-lang/rust/issues/67984) to be supported as well.

## API Surface

```rust
impl TokenStream {
    pub fn expand_expr(&self) -> Result<TokenStream, ExpandError>;
}

#[non_exhaustive]
pub struct ExpandError;

impl Debug for ExpandError { ... }
impl Display for ExpandError { ... }
impl Error for ExpandError {}
impl !Send for ExpandError {}
impl !Sync for ExpandError {}
```

### DIFF
--- a/compiler/rustc_expand/src/proc_macro_server.rs
+++ b/compiler/rustc_expand/src/proc_macro_server.rs
@@ -1,4 +1,4 @@
-use crate::base::{ExtCtxt, ResolverExpand};
+use crate::base::ExtCtxt;
 
 use rustc_ast as ast;
 use rustc_ast::token::{self, Nonterminal, NtIdent};
@@ -7,7 +7,7 @@ use rustc_ast::tokenstream::{DelimSpan, Spacing::*, TokenStream, TreeAndSpacing}
 use rustc_ast_pretty::pprust;
 use rustc_data_structures::fx::FxHashMap;
 use rustc_data_structures::sync::Lrc;
-use rustc_errors::Diagnostic;
+use rustc_errors::{Diagnostic, PResult};
 use rustc_lint_defs::builtin::PROC_MACRO_BACK_COMPAT;
 use rustc_lint_defs::BuiltinLintDiagnostics;
 use rustc_parse::lexer::nfc_normalize;
@@ -53,11 +53,11 @@ impl ToInternal<token::DelimToken> for Delimiter {
     }
 }
 
-impl FromInternal<(TreeAndSpacing, &'_ mut Vec<Self>, &mut Rustc<'_>)>
+impl FromInternal<(TreeAndSpacing, &'_ mut Vec<Self>, &mut Rustc<'_, '_>)>
     for TokenTree<Group, Punct, Ident, Literal>
 {
     fn from_internal(
-        ((tree, spacing), stack, rustc): (TreeAndSpacing, &mut Vec<Self>, &mut Rustc<'_>),
+        ((tree, spacing), stack, rustc): (TreeAndSpacing, &mut Vec<Self>, &mut Rustc<'_, '_>),
     ) -> Self {
         use rustc_ast::token::*;
 
@@ -146,10 +146,10 @@ impl FromInternal<(TreeAndSpacing, &'_ mut Vec<Self>, &mut Rustc<'_>)>
             SingleQuote => op!('\''),
 
             Ident(name, false) if name == kw::DollarCrate => tt!(Ident::dollar_crate()),
-            Ident(name, is_raw) => tt!(Ident::new(rustc.sess, name, is_raw)),
+            Ident(name, is_raw) => tt!(Ident::new(rustc.sess(), name, is_raw)),
             Lifetime(name) => {
                 let ident = symbol::Ident::new(name, span).without_first_quote();
-                stack.push(tt!(Ident::new(rustc.sess, ident.name, false)));
+                stack.push(tt!(Ident::new(rustc.sess(), ident.name, false)));
                 tt!(Punct::new('\'', true))
             }
             Literal(lit) => tt!(Literal { lit }),
@@ -181,15 +181,15 @@ impl FromInternal<(TreeAndSpacing, &'_ mut Vec<Self>, &mut Rustc<'_>)>
             Interpolated(nt)
                 if let Some((name, is_raw)) = ident_name_compatibility_hack(&nt, span, rustc) =>
             {
-                TokenTree::Ident(Ident::new(rustc.sess, name.name, is_raw, name.span))
+                TokenTree::Ident(Ident::new(rustc.sess(), name.name, is_raw, name.span))
             }
             Interpolated(nt) => {
-                let stream = nt_to_tokenstream(&nt, rustc.sess, CanSynthesizeMissingTokens::No);
+                let stream = nt_to_tokenstream(&nt, rustc.sess(), CanSynthesizeMissingTokens::No);
                 TokenTree::Group(Group {
                     delimiter: Delimiter::None,
                     stream,
                     span: DelimSpan::from_single(span),
-                    flatten: crate::base::pretty_printing_compatibility_hack(&nt, rustc.sess),
+                    flatten: crate::base::pretty_printing_compatibility_hack(&nt, rustc.sess()),
                 })
             }
 
@@ -355,30 +355,30 @@ pub struct Literal {
     span: Span,
 }
 
-pub(crate) struct Rustc<'a> {
-    resolver: &'a dyn ResolverExpand,
-    sess: &'a ParseSess,
+pub(crate) struct Rustc<'a, 'b> {
+    ecx: &'a mut ExtCtxt<'b>,
     def_site: Span,
     call_site: Span,
     mixed_site: Span,
-    span_debug: bool,
     krate: CrateNum,
     rebased_spans: FxHashMap<usize, Span>,
 }
 
-impl<'a> Rustc<'a> {
-    pub fn new(cx: &'a ExtCtxt<'_>) -> Self {
-        let expn_data = cx.current_expansion.id.expn_data();
+impl<'a, 'b> Rustc<'a, 'b> {
+    pub fn new(ecx: &'a mut ExtCtxt<'b>) -> Self {
+        let expn_data = ecx.current_expansion.id.expn_data();
         Rustc {
-            resolver: cx.resolver,
-            sess: cx.parse_sess(),
-            def_site: cx.with_def_site_ctxt(expn_data.def_site),
-            call_site: cx.with_call_site_ctxt(expn_data.call_site),
-            mixed_site: cx.with_mixed_site_ctxt(expn_data.call_site),
-            span_debug: cx.ecfg.span_debug,
+            def_site: ecx.with_def_site_ctxt(expn_data.def_site),
+            call_site: ecx.with_call_site_ctxt(expn_data.call_site),
+            mixed_site: ecx.with_mixed_site_ctxt(expn_data.call_site),
             krate: expn_data.macro_def_id.unwrap().krate,
             rebased_spans: FxHashMap::default(),
+            ecx,
         }
+    }
+
+    fn sess(&self) -> &ParseSess {
+        self.ecx.parse_sess()
     }
 
     fn lit(&mut self, kind: token::LitKind, symbol: Symbol, suffix: Option<Symbol>) -> Literal {
@@ -386,7 +386,7 @@ impl<'a> Rustc<'a> {
     }
 }
 
-impl server::Types for Rustc<'_> {
+impl server::Types for Rustc<'_, '_> {
     type FreeFunctions = FreeFunctions;
     type TokenStream = TokenStream;
     type TokenStreamBuilder = tokenstream::TokenStreamBuilder;
@@ -401,17 +401,20 @@ impl server::Types for Rustc<'_> {
     type Span = Span;
 }
 
-impl server::FreeFunctions for Rustc<'_> {
+impl server::FreeFunctions for Rustc<'_, '_> {
     fn track_env_var(&mut self, var: &str, value: Option<&str>) {
-        self.sess.env_depinfo.borrow_mut().insert((Symbol::intern(var), value.map(Symbol::intern)));
+        self.sess()
+            .env_depinfo
+            .borrow_mut()
+            .insert((Symbol::intern(var), value.map(Symbol::intern)));
     }
 
     fn track_path(&mut self, path: &str) {
-        self.sess.file_depinfo.borrow_mut().insert(Symbol::intern(path));
+        self.sess().file_depinfo.borrow_mut().insert(Symbol::intern(path));
     }
 }
 
-impl server::TokenStream for Rustc<'_> {
+impl server::TokenStream for Rustc<'_, '_> {
     fn new(&mut self) -> Self::TokenStream {
         TokenStream::default()
     }
@@ -422,12 +425,61 @@ impl server::TokenStream for Rustc<'_> {
         parse_stream_from_source_str(
             FileName::proc_macro_source_code(src),
             src.to_string(),
-            self.sess,
+            self.sess(),
             Some(self.call_site),
         )
     }
     fn to_string(&mut self, stream: &Self::TokenStream) -> String {
         pprust::tts_to_string(stream)
+    }
+    fn expand_expr(&mut self, stream: &Self::TokenStream) -> Result<Self::TokenStream, ()> {
+        // Parse the expression from our tokenstream.
+        let expr: PResult<'_, _> = try {
+            let mut p = rustc_parse::stream_to_parser(
+                self.sess(),
+                stream.clone(),
+                Some("proc_macro expand expr"),
+            );
+            let expr = p.parse_expr()?;
+            if p.token != token::Eof {
+                p.unexpected()?;
+            }
+            expr
+        };
+        let expr = expr.map_err(|mut err| err.emit())?;
+
+        // Perform eager expansion on the expression.
+        let expr = self
+            .ecx
+            .expander()
+            .fully_expand_fragment(crate::expand::AstFragment::Expr(expr))
+            .make_expr();
+
+        // NOTE: For now, limit `expand_expr` to exclusively expand to literals.
+        // This may be relaxed in the future.
+        // We don't use `nt_to_tokenstream` as the tokenstream currently cannot
+        // be recovered in the general case.
+        match &expr.kind {
+            ast::ExprKind::Lit(l) => {
+                Ok(tokenstream::TokenTree::token(token::Literal(l.token), l.span).into())
+            }
+            ast::ExprKind::Unary(ast::UnOp::Neg, e) => match &e.kind {
+                ast::ExprKind::Lit(l) => match l.token {
+                    token::Lit { kind: token::Integer | token::Float, .. } => {
+                        Ok(std::array::IntoIter::new([
+                            // FIXME: The span of the `-` token is lost when
+                            // parsing, so we cannot faithfully recover it here.
+                            tokenstream::TokenTree::token(token::BinOp(token::Minus), e.span),
+                            tokenstream::TokenTree::token(token::Literal(l.token), l.span),
+                        ])
+                        .collect())
+                    }
+                    _ => Err(()),
+                },
+                _ => Err(()),
+            },
+            _ => Err(()),
+        }
     }
     fn from_token_tree(
         &mut self,
@@ -440,7 +492,7 @@ impl server::TokenStream for Rustc<'_> {
     }
 }
 
-impl server::TokenStreamBuilder for Rustc<'_> {
+impl server::TokenStreamBuilder for Rustc<'_, '_> {
     fn new(&mut self) -> Self::TokenStreamBuilder {
         tokenstream::TokenStreamBuilder::new()
     }
@@ -452,7 +504,7 @@ impl server::TokenStreamBuilder for Rustc<'_> {
     }
 }
 
-impl server::TokenStreamIter for Rustc<'_> {
+impl server::TokenStreamIter for Rustc<'_, '_> {
     fn next(
         &mut self,
         iter: &mut Self::TokenStreamIter,
@@ -477,7 +529,7 @@ impl server::TokenStreamIter for Rustc<'_> {
     }
 }
 
-impl server::Group for Rustc<'_> {
+impl server::Group for Rustc<'_, '_> {
     fn new(&mut self, delimiter: Delimiter, stream: Self::TokenStream) -> Self::Group {
         Group {
             delimiter,
@@ -506,7 +558,7 @@ impl server::Group for Rustc<'_> {
     }
 }
 
-impl server::Punct for Rustc<'_> {
+impl server::Punct for Rustc<'_, '_> {
     fn new(&mut self, ch: char, spacing: Spacing) -> Self::Punct {
         Punct::new(ch, spacing == Spacing::Joint, server::Span::call_site(self))
     }
@@ -524,9 +576,9 @@ impl server::Punct for Rustc<'_> {
     }
 }
 
-impl server::Ident for Rustc<'_> {
+impl server::Ident for Rustc<'_, '_> {
     fn new(&mut self, string: &str, span: Self::Span, is_raw: bool) -> Self::Ident {
-        Ident::new(self.sess, Symbol::intern(string), is_raw, span)
+        Ident::new(self.sess(), Symbol::intern(string), is_raw, span)
     }
     fn span(&mut self, ident: Self::Ident) -> Self::Span {
         ident.span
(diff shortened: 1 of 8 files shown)


mystor:
I've left this as a draft PR for now as I'm unsure whether a feature like this would be desired, and/or whether an RFC would be required before landing.

petrochenkov:
>I'm unsure whether a feature like this would be desired, and/or whether an RFC would be required before landing.

There were some RFCs about eager expansion (of which `expand_literal` is a special case), all postponed - https://github.com/rust-lang/rfcs/pull/2320, https://github.com/rust-lang/rfcs/pull/1628, https://github.com/rust-lang/rfcs/pull/2628.

mystor:
> There were some RFCs about eager expansion (of which `expand_literal` is a special case), all postponed - [rust-lang/rfcs#2320](https://github.com/rust-lang/rfcs/pull/2320), [rust-lang/rfcs#1628](https://github.com/rust-lang/rfcs/pull/1628), [rust-lang/rfcs#2628](https://github.com/rust-lang/rfcs/pull/2628).

Ah, that's good to know. Thanks for the links (I should've probably looked them up earlier). I take it that that means a change like this would require a corresponding RFC :-)

compiler/rustc_expand/src/proc_macro_server.rs
```diff
@@ -354,9 +355,8 @@ pub struct Literal {
     span: Span,
 }
 
-pub(crate) struct Rustc<'a> {
-    resolver: &'a dyn ResolverExpand,
-    sess: &'a ParseSess,
+pub(crate) struct Rustc<'a, 'b> {
+    ecx: &'a mut ExtCtxt<'b>,
```
petrochenkov, line 359:
```suggestion
    ecx: &'a mut ExtCtxt<'a>,
```
A single lifetime should probably be enough here.

compiler/rustc_expand/src/proc_macro_server.rs
```diff
@@ -365,27 +365,30 @@ pub(crate) struct Rustc<'a> {
...
-            resolver: cx.resolver,
-            sess: cx.parse_sess(),
-            def_site: cx.with_def_site_ctxt(expn_data.def_site),
-            call_site: cx.with_call_site_ctxt(expn_data.call_site),
-            mixed_site: cx.with_mixed_site_ctxt(expn_data.call_site),
-            span_debug: cx.ecfg.span_debug,
+            def_site: ecx.with_def_site_ctxt(expn_data.def_site),
+            call_site: ecx.with_call_site_ctxt(expn_data.call_site),
+            mixed_site: ecx.with_mixed_site_ctxt(expn_data.call_site),
+            span_debug: ecx.ecfg.span_debug,
```
petrochenkov, line 375: `span_debug` is no longer necessary if we have `ecx`.

compiler/rustc_expand/src/proc_macro_server.rs
```diff
@@ -421,13 +427,58 @@ impl server::TokenStream for Rustc<'_> {
...
     }
+    fn expand_literal(&mut self, stream: &Self::TokenStream) -> Result<Self::Literal, ()> {
+        // Parse the expression from our tokenstream.
+        let expr = {
+            let mut p = rustc_parse::stream_to_parser(
+                self.sess(),
+                stream.clone(),
+                Some("proc_macro expand literal"),
+            );
+            let expr = p.parse_expr().map_err(|_| ())?;
```
petrochenkov, line 445:
```suggestion
            let expr = p.parse_expr().map_err(|err| { err.cancel(); })?;
```
This is going to panic on drop if the error is not reported or canceled.
Not sure what is the preferred behavior, should `expand_literal` be able to accept garbage and return `Err`, with the user ignoring `Err` and continuing compilation?

compiler/rustc_expand/src/proc_macro_server.rs
```diff
@@ -421,13 +427,58 @@ impl server::TokenStream for Rustc<'_> {
...
+        // Perform eager expansion on the expression.
+        let expr = self
+            .ecx
+            .expander()
+            .fully_expand_fragment(crate::expand::AstFragment::Expr(expr))
+            .make_expr();
+
+        // Extract the literal value from the expression.
+        let lit = match &expr.kind {
+            ast::ExprKind::Lit(l) => l.kind.to_lit_token(),
```
petrochenkov, line 461:
```suggestion
            ast::ExprKind::Lit(l) => l.token,
```
`to_lit_token` is a best-effort recovery used mostly for pretty-printing, but we need to preserve the token precisely here.

compiler/rustc_expand/src/proc_macro_server.rs
```diff
@@ -421,13 +427,58 @@ impl server::TokenStream for Rustc<'_> {
...
+            ast::ExprKind::Unary(ast::UnOp::Neg, e) => match &e.kind {
```
petrochenkov, line 462: I would prefer to avoid hacks like this, `-1` is not a single token and `token::Lit` is supposed to never keep anything with `-`.

compiler/rustc_expand/src/proc_macro_server.rs
```diff
@@ -421,13 +427,58 @@ impl server::TokenStream for Rustc<'_> {
...
+            },
+            _ => return Err(()),
+        };
+
+        // Filter to valid literal kinds to return from this method, and
+        // sanitize the returned span to ensure the calling proc_macro can't use
+        // them.
+        match lit.kind {
+            token::Bool | token::Err => Err(()),
+            _ => Ok(Literal { lit, span: expr.span.with_ctxt(self.call_site.ctxt()) }),
```
petrochenkov, line 479: This sanitizing doesn't make sense to me and I'm honestly don't understand the amount of FUD related to this in https://github.com/rust-lang/rust/issues/67984 and the relevant RFC thread.
We should preserve the hygienic context here, and specify what context is assigned to literals returned by built-in macros - call-site (most likely), def-site, or explicitly unspecified. In the same way like we specify that for identifiers.

compiler/rustc_expand/src/proc_macro_server.rs
petrochenkov: All the stuff like `// After expansion, all integer literals are decimal` should disappear from the tests after this fix.

petrochenkov:
I'm mostly ok with merging something like this in general.
The `proc_macro` crate contains a lot of experimental stuff with unclear path to stabilization.

Why stabilizing this may be simpler than eager expansion in general:
- We reserve the right to return `Err` on a whim. For example if the macro to expand is not *yet* resolvable at this point in compilation, but may become resolvable better.
     - It's not clear whether it's ok to continue compilation after receiving `Err` from `expand_literal`. `Err` not becoming `Ok` in the future must not be something that users can rely on.
     - If `expand_literal` returns `Ok`, but compiler produces errors while evaluating it, then it's still a compilation error, so we are good.
- All the issues with complex module structures existing during eager expansion are avoided, if the expansion does not result in a single literal token in the end, then it's an error (although see the previous item).
- This is a very limited and special-purpose version of some potential more general future eager-expansion API, if such API appears at some point, then this function can be easily deprecated and reimplemented in terms of that API, I think.

compiler/rustc_expand/src/proc_macro_server.rs
mystor: I believe I tried this locally and it will not build if we don't use 2 lifetimes. IIRC this is because the inner lifetime is made invariant by the `&mut`, like `&'a mut &'b ()` not being convertible to `&'c mut &'c ()` (https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=2f03be1aec5cd6811df24b957bac7fb5)

Other places which store an `&mut ExtCtxt<'_>` in a struct need to use 2 lifetimes as well (e.g. https://searchfox.org/rust/rev/7c89e389d00cfc7e86ae7e1b45880da4f5f5c9f5/compiler/rustc_expand/src/placeholders.rs#180)

compiler/rustc_expand/src/proc_macro_server.rs
mystor: I'm unsure about that. I considered making every `Err` returned also correspond with an error diagnostic being emitted, so the build will fail after that point. My reasoning was that this would communicate the most useful diagnostics to the user without exposing internal details to the calling macro. I didn't end up following through with the idea as it wasn't consistent with the behaviour of other APIs in the proc_macro interface, and I worried about emitting duplicate errors as the macro will likely panic itself. I'm open to making the change though.

compiler/rustc_expand/src/proc_macro_server.rs
mystor: Ah... I don't how how I missed that member existing, my bad.

compiler/rustc_expand/src/proc_macro_server.rs
mystor: My reasoning behind having this hack was that a `-1` `token::Lit` already exists whenever the stable `Literal::i32_unsuffixed(-1)` API is called (https://doc.rust-lang.org/nightly/proc_macro/struct.Literal.html#method.i32_unsuffixed). This change is consistent with that behaviour, even if it's quite unfortunate. I don't think there's any way we can get away at this point with not supporting negative `Literal` objects in the proc_macro API.

These weird tokens are broken apart before being exposed to the rest of rustc (https://searchfox.org/rust/rev/7c89e389d00cfc7e86ae7e1b45880da4f5f5c9f5/compiler/rustc_expand/src/proc_macro_server.rs#217,228), and with the changes in #86822 won't even be a `token::Lit` anymore.

compiler/rustc_expand/src/proc_macro_server.rs
mystor: My concern with preserving it was mostly theoretical, and focused on a concern around what might happen if someone invoked a std macro which has `#[allow_internal_unstable(...)]` on it and exposes a literal. I'm OK with dropping the sanitizing if we're confident it won't be an issue, or want to leave figuring it out for later.

compiler/rustc_expand/src/proc_macro_server.rs
```diff
@@ -421,13 +425,58 @@ impl server::TokenStream for Rustc<'_> {
...
+    fn expand_literal(&mut self, stream: &Self::TokenStream) -> Result<Self::Literal, ()> {
+        // Parse the expression from our tokenstream.
+        let expr = {
+            let mut p = rustc_parse::stream_to_parser(
+                self.sess(),
+                stream.clone(),
+                Some("proc_macro expand literal"),
+            );
+            let expr = p.parse_expr().map_err(|mut err| {
+                err.cancel();
```
petrochenkov, line 444: After some thought, perhaps it's better to report the error.
It's a strictly more conservative choice (it produces a compilation error), it makes it visible why `expand_literal` failed, and we can relax it later if convincing use cases are provided.

compiler/rustc_expand/src/proc_macro_server.rs
petrochenkov: Ok, `i32_unsuffixed` and friends are indeed packing the minus into `token::Lit`. That's unfortunate.

compiler/rustc_expand/src/proc_macro_server.rs
```diff
@@ -421,13 +425,58 @@ impl server::TokenStream for Rustc<'_> {
...
+                },
+                _ => return Err(()),
+            },
+            _ => return Err(()),
+        };
+
+        // Filter to valid literal kinds to return from this method.
+        match lit.kind {
+            token::Bool | token::Err => Err(()),
+            _ => Ok(Literal { lit, span: expr.span }),
```
petrochenkov, line 477: Using the literal span (`ast::ExprKind::Lit(l) => l.span`) instead of the expr span would be more correct here (they can potentially differ).

petrochenkov:
@mystor 
I'd like to have some second opinion before merging this.
Could you make a compiler team MCP from this PR, or a library team proposal, I'm not sure.

compiler/rustc_expand/src/proc_macro_server.rs
mystor: Should we report an error on every error return path so that callers can't rely on any failure not reporting an error & failing compilation to be even more conservative, or just report the parse error?

With just reporting an error on expansion failure or parse errors, I could see callers relying on e.g. checking if an expression is literal by calling `expand_literal`, and assuming it isn't literal & using it as a runtime expression with a different expansion if `expand_literal` fails. This could theoretically be used to support something like a `regex!` with a potentially dynamic string argument which eagerly compiles if the expression is a literal.

I don't really see much use for recovering on parse errors other than potentially avoiding duplicate errors if the macro both checks if it's a literal and uses it as a normal expression (though a macro should be able to pre-parse the expression with `syn`)

compiler/rustc_expand/src/proc_macro_server.rs
mystor: If we wanted, we could change the return type of `expand_literal` to be `TokenStream`. This would let us represent a negative numeric literal properly as two tokens, and would also allow us to potentially support boolean `true` and `false` keyword literals, as they need to be stored as an `Ident` rather than `Literal`.

I mostly used `Literal` as the return type because it seemed like the obvious choice, and it might be confusing that `TokenStream` was returned after expanding to a literal :-)

compiler/rustc_expand/src/proc_macro_server.rs
petrochenkov: >Should we report an error on every error return path so that callers can't rely on any failure not reporting an error & failing compilation to be even more conservative, or just report the parse error?

I'm going to leave this for the libs team to decide (since this PR is `T-libs-api` now).
I'd be ok with always reporting an error for now.

compiler/rustc_expand/src/proc_macro_server.rs
petrochenkov: If `i32_unsuffixed` can produce a `Literal` with minus in `Symbol`, then this is acceptable too.
Although it it would be better to use some representation not requiring keeping the minus in the string (like keeping a private boolean flag instead), perhaps it can be done as a part of #86822.

mystor: Sure, that wouldn't be a hard thing to implement.

m-ou-se:
We discussed this in the library-api meeting just now. We do like to see macro-expanding functionality on TokenStream. We do see the concerns with a general .expand() as already mentioned above, but would like to explore solutions that allow expanding to more than just literals in the near future. So we'd like to propose:

1. Rename this to not be specific to just literals and return a TokenStream instead of a Literal; and
2. Keep the implementation as restrictive as it is, failing if the result is not a single literal.

This way we sidestep the potential issues for now, but leave space for expanding the use cases of this function in the near future to include more expansions than only literals.

@mystor I'd like to hear your thoughts on this.

joshtriplett:
One other thought from that discussion (**not** something that needs to be done in this PR, just a possibility for solving the more general case in the future):

Could we provide a way to expand all macros recursively, and then if any macro expands to code that uses any unstable features internally, we expand that particular macro to an `Opaque` object whose output can be included in the output stream but not introspected or modified? That way, you can expand something like `format_args!` if you're only going to consume its output, but if you want to introspect the output of something (e.g. walk the tokens from an included file) it needs to be stable.

mystor:
> We discussed this in the library-api meeting just now. We do like to see macro-expanding functionality on TokenStream. We do see the concerns with a general .expand() as already mentioned above, but would like to explore solutions that allow expanding to more than just literals in the near future. So we'd like to propose:
> 
>   1. Rename this to not be specific to just literals and return a TokenStream instead of a Literal; and
> 
>   2. Keep the implementation as restrictive as it is, failing if the result is not a single literal.
>
> This way we sidestep the potential issues for now, but leave space for expanding the use cases of this function in the near future to include more expansions than only literals.

I'm not sure a general `expand` method would make sense, as macro expansion can be dependent on the context where it's occuring, and the `TokenStream` passed in needs to be parsed into an expression/type/whatever before expansion can start. This method could be generalized to `expand_expr()` rather than `expand_literal()`, but we'd probably also need `expand_type()`, `expand_pat()`, `expand_items()` etc. methods to cover the different types of macro expansions which are supported by rust. A general `expand` would probably need to either infer what is being expanded from how the macro being invoked is being expanded (though that feels sketchy to me, and I'd prefer not to do that), or would need to pick an expansion type like `Expr` and require wrapping other types in blocks or similar to parse items, types, or patterns.

If we do generalize to `expand_expr()`, I suppose we could also relax the requirements slightly to also allow `true` and `false` literal tokens now that they don't need to be valid `Literal`s.

> Could we provide a way to expand all macros recursively, and then if any macro expands to code that uses any unstable features internally, we expand that particular macro to an `Opaque` object whose output can be included in the output stream but not introspected or modified? That way, you can expand something like `format_args!` if you're only going to consume its output, but if you want to introspect the output of something (e.g. walk the tokens from an included file) it needs to be stable.

Having some way to represent an opaque sequence of tokens in the `TokenStream` would definitely be quite useful for this, but it would be quite a hassle to do that backwards compatibly at this point. The [`TokenTree` enum](https://doc.rust-lang.org/nightly/proc_macro/enum.TokenTree.html) is stable and only has the `Ident`, `Literal`, `Punct`, and `Group` options, so if we were to encode an opaque stream in, it would need to be done using one of those variants, as we can't add an `Opaque` variant.

The approach which seems most likely to me to not break existing macros would be to re-wrap the expanded AST as an unexpanded macro (e.g. `opaque_expr!{}`) with the specific expansion encoded into the span's context. This could theoretically be wrapped in a special `Delimiter::None`-delimited group with an `as_opaque(&self) -> Option<Opaque>` getter if we wanted to add special methods for operating on opaque expansions in the future.

Either way, this is probably something which should be handled in a separate issue or RFC.

joshtriplett:
@mystor Thanks for clarifying the issue. We discussed this in today's @rust-lang/libs-api meeting, and we understand and agree that a fully general `expand` isn't an option. It *sounds* like it'd be reasonable to have an `expand_expr` method (that then can only return a literal for now). Assuming there's no blocker to doing so, we'd like to go that route.

camelid:
triage: What's the status of this? It looks like the latest commit switched to the suggested `expand_expr` approach?

joshtriplett:
I think this is ready to merge, as soon as the conflicts are resolved. r=me with conflicts resolved and CI passing.

mystor:
Ah yeah, sorry about that. I never got around to actually rebasing the changes

library/proc_macro/src/lib.rs
```diff
@@ -109,6 +109,34 @@ impl !Send for LexError {}
...
+/// Error returned from `TokenStream::expand_expr`.
+#[unstable(feature = "proc_macro_expand", issue = "none")]
+#[non_exhaustive]
+#[derive(Debug)]
+pub struct ExpandError;
+
+impl ExpandError {
+    fn new() -> Self {
+        ExpandError
+    }
```
petrochenkov, line 121: This function is trivial and private, so it's shouldn't be necessary.

petrochenkov:
PR description still mentions `expand_literal` and needs an update.
r=me after rebasing and squashing commits.

petrochenkov:
I suspect that the restriction on the `expand_expr`'s output can be relaxed from "only literals" to "no block expressions anywhere" (not in this PR).

Blocks can contain items (including imports), which bring numerous issues to eager expansion.
If the final output doesn't contain blocks, then intermediate expansion states shouldn't contain them either (please, give a counter-example if you can).
(`#[macro_attribute] { /* block */ }` doesn't count as a block, it should be more or less equivalent to `macro_attribute! { /* block */ }` for our purpose.)

library/proc_macro/src/lib.rs
mystor: I did it for symmetry with `LexError`, but can remove if you want.

petrochenkov: Yes, `LexError::new` can be removed as well.

petrochenkov:
@jyn514 
Originally I didn't r+ this because https://github.com/rust-lang/rust/pull/87264#discussion_r702481734 was not addressed, I didn't expect @mystor to disappear for so long.
That comment is not critical though, so the PR can be landed without it.

matthiaskrgr:
@bors r-
hmm looks like this failed in a rollup pre-merge-ci:
https://github.com/rust-lang/rust/pull/90732#issuecomment-964393031
Can you rebase and check that the tests still pass? Thanks!

petrochenkov:
>Can you rebase and check that the tests still pass?

And address https://github.com/rust-lang/rust/pull/87264#discussion_r702481734 as well.

mystor:
> I didn't expect @mystor to disappear for so long. That comment is not critical though, so the PR can be landed without it.

Apologies, I unfortunately got busy with other things and then forgot about this pending PR. I've rebased it, removed the unnecessary constructor functions & re-blessed the ui test output.

jyn514:
> should the tracking issue be filed before the patch is landed so that the issue number can be included in the unstable annotations

Yes, please :)

mystor:
Filed #90765 as the tracking issue, updated the patch, and added various unresolved questions there which I think we'll need to figure out before stabilization.

petrochenkov:
I don't know why compiletest doesn't normalize that path to `$DIR`.
You can either remove the test case with `file!()`, or fix compiletest somehow (the relevant code is `fn normalize_output`).

mystor:
> I don't know why compiletest doesn't normalize that path to `$DIR`. You can either remove the test case with `file!()`, or fix compiletest somehow (the relevant code is `fn normalize_output`).

I've removed a level of escaping from the error log output which I'm hoping will make it so that `compiletest` can normalize the path strings a bit better. Is there an easy way to ask bors to run the test on windows? I unfortunately don't have a convenient windows machine set up to build rustc right now.

If this doesn't work I have some other ideas, but I'm hopeful it will.

mystor:
Pushed & reverted a patch which enabled running the `i686-msvc-*` suite for PRs to run that test at @m-ou-se's suggestion. Results can be seen here: https://github.com/rust-lang/rust/runs/4193304021?check_suite_focus=true

mystor:
Looks like it's still failing, I'm guessing the 1 level of escaping is still causing issues for the parser, as normal diagnostic output wouldn't contain that. The raw output appears to be `"D:\\a\\rust\\rust\\src/test\\ui\\proc-macro\\expand-expr.rs"`, and post-processing it's: `"D:\a\rust\rust\src/test/ui/proc-macro/expand-expr.rs"`, which has changed all of the `\\` to `/` after the `src/test` but misses the ones before that, which are just collapsed to `\` instead of `\\` for some reason.

I don't know too much about how the test suite works, but I'm guessing that the random `/` in the path is coming from https://github.com/rust-lang/rust/blob/220ed09b26177ca4b6ab525f403d251024389a41/src/bootstrap/test.rs#L1320

Either way, I think the easiest fix here is to stop relying on the compiletest normalization for testing, so I'll change the test to just compare the path extracted from `Span::call_site()` to the output of `file!()`.

mystor:
New test run: https://github.com/rust-lang/rust/runs/4194072832?check_suite_focus=true

camelid:
> The raw output appears to be `"D:\\a\\rust\\rust\\src/test\\ui\\proc-macro\\expand-expr.rs"`, and post-processing it's: `"D:\a\rust\rust\src/test/ui/proc-macro/expand-expr.rs"`, which has changed all of the `\\` to `/` after the `src/test` but misses the ones before that, which are just collapsed to `\` instead of `\\` for some reason.

Maybe open an issue for that?

mystor:
> Maybe open an issue for that?

I'm not sure if it's a serious issue in general, because normally this would work fine. The replacement of paths to `$DIR` appears to happen before any platform normalization, and be quite aware that it's flakey: https://github.com/rust-lang/rust/blob/220ed09b26177ca4b6ab525f403d251024389a41/src/tools/compiletest/src/runtest.rs#L3515-L3518

I think my particular case here of printing a string literal token and hoping that the path inside of it is normalized just isn't handled by `compiletest` and only worked on linux because it didn't contain any escaped characters.

mystor:
Huh, I'm quite surprised this caused a perf regression... Only reason I can think of off the top of my head is because accessing `sess` on the proc macro server now requires a little bit more indirection: https://github.com/rust-lang/rust/blob/1b12d01903293453dd94aa170c82caf94415629f/compiler/rustc_expand/src/proc_macro_server.rs#L380-L382

Should be fairly straightforward to reduce the interactions back down I suppose. We could add a field:

```
sess: &'b ParseSess,
```

and fetch it during the constructor before storing the `&'a mut ExtCtxt<'b>` into the type. We could also store the `span_debug` bool and the `&'b dyn ResolverExpand` if we're worried about either of those extra indirections as well.

I can't think of what else could be causing the regression of the top of my head, as the vast majority of the changes here should not be in any critical path.

PR #86867, opened 2021-07-04
Convert `debug_assert!` to `assert!` in `Binder::dummy`

JohnTitor:
This is needed for #85350 not to be passed.
r? @jackh726

PR #87271, opened 2021-07-19
Update Clippy

flip1995:
This is an out-of-cycle Clippy update, to fix 3 ICEs before the release (This should be merged before beta is branched):

rust-lang/rust-clippy#7470
rust-lang/rust-clippy#7471
rust-lang/rust-clippy#7473

cc @jackh726 @JohnTitor rust-lang/rust-clippy#7470 was caused by #86867. I saw the same ICE in the last rustup for Clippy though, so this might be a more general problem. Is there something we should check before calling `layout_of`? Should we always check for `ty.has_escaping_bound_vars()` before calling `layout_of`? Or is this overkill?

r? @Manishearth


jackh726:
Okay, regarding rust-lang/rust-clippy#7470, I would have to look into it, but it seems a bit suspicious to me. I *think* layout code should be fine to have bound vars (though, I'm not exactly sure if these should be wrapped in a `Binder`). It seems like that error is originating all the way in `type_known_to_meet_bound_modulo_regions`, where it expects `ty` to not have bound vars. I could conceivably think that we *would* allow `Binder<Ty>` there, but that's a bigger change.

Another option would be to `normalize_erasing_late_bound_regions` before calling `layout_of`, but I'm unsure if that is *correct*. This is tricky and definitely seems a bit weird.

flip1995:
@jackh726 Thanks for the explanation. The `normalize_erasing_late_bound_regions` is another problem we had with ` layout_of` before:

https://github.com/rust-lang/rust-clippy/blob/f467750680987427d7ff045bbe9e84317140d0aa/clippy_lints/src/zero_sized_map_values.rs#L58-L60

https://github.com/rust-lang/rust-clippy/blob/e9c3991d3081c79bd712d754e6f51f456e603c6c/clippy_utils/src/ty.rs#L182-L187

This problem is orthogonal to the `has_escaping_bound_vars` problem though, since removing the one of the two leads to 2 different ICEs when calling `layout_of`. It is definitely conceivable, that we misuse the `layout_of` query. But just from the documentation and looking at the code, I can't see what we're doing wrong.

PR #87273, opened 2021-07-19
Recognize bounds on impls as const bounds

fee1-dead:
r? @oli-obk

### DIFF
--- a/compiler/rustc_hir/src/hir.rs
+++ b/compiler/rustc_hir/src/hir.rs
@@ -3064,6 +3064,27 @@ impl<'hir> Node<'hir> {
             Node::Crate(_) | Node::Visibility(_) => None,
         }
     }
+
+    /// Returns `Constness::Const` when this node is a const fn/impl.
+    pub fn constness(&self) -> Constness {
+        match self {
+            Node::Item(Item {
+                kind: ItemKind::Fn(FnSig { header: FnHeader { constness, .. }, .. }, ..),
+                ..
+            })
+            | Node::TraitItem(TraitItem {
+                kind: TraitItemKind::Fn(FnSig { header: FnHeader { constness, .. }, .. }, ..),
+                ..
+            })
+            | Node::ImplItem(ImplItem {
+                kind: ImplItemKind::Fn(FnSig { header: FnHeader { constness, .. }, .. }, ..),
+                ..
+            })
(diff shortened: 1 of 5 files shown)


compiler/rustc_hir/src/hir.rs
```diff
@@ -3064,6 +3064,27 @@ impl<'hir> Node<'hir> {
...
+
+    /// Returns `Constness::Const` when this node is a const fn/impl.
+    pub fn constness(&self) -> Constness {
+        match self {
+            Node::Item(Item {
+                kind: ItemKind::Fn(FnSig { header: FnHeader { constness, .. }, .. }, ..),
+                ..
+            })
+            | Node::TraitItem(TraitItem {
+                kind: TraitItemKind::Fn(FnSig { header: FnHeader { constness, .. }, .. }, ..),
```
oli-obk, line 3076: this cannot happen (yet), maybe put it in a separate match arm and assert that the constness is `NotConst`

oli-obk:
Nice! This is a great fix.

compiler/rustc_hir/src/hir.rs
```diff
@@ -3064,6 +3064,35 @@ impl<'hir> Node<'hir> {
...
+                ..
+            })
+            | Node::Item(Item { kind: ItemKind::Impl(Impl { constness, .. }), .. }) => *constness,
+
+            Node::TraitItem(TraitItem {
+                kind: TraitItemKind::Fn(FnSig { header: FnHeader { constness, .. }, .. }, ..),
+                ..
+            })
+            | Node::ImplItem(ImplItem {
+                kind: ImplItemKind::Fn(FnSig { header: FnHeader { constness, .. }, .. }, ..),
```
oli-obk, line 3082: this on the other hand can happen for inherent impls, as we already have stable inherent const fns

fee1-dead: Okay, fixing that now.

oli-obk:
looks like I was wrong, it is reachable, but only in code that already errors. I guess go back to your original PR. Sorry about that.

compiler/rustc_hir/src/hir.rs
fee1-dead: I'll just revert this change since it is caught elsewhere.

PR #87279, opened 2021-07-19
Add comments explaining the unix command-line argument support.

sunfishcode:
Following up on #87236, add comments to the unix command-line argument
support explaining that the code doesn't mutate the system-provided
argc/argv, and that this is why the code doesn't need a lock or special
memory ordering.

r? @RalfJung

### DIFF
--- a/library/std/src/sys/unix/args.rs
+++ b/library/std/src/sys/unix/args.rs
@@ -77,10 +77,18 @@ mod imp {
     use crate::ptr;
     use crate::sync::atomic::{AtomicIsize, AtomicPtr, Ordering};
 
+    // The system-provided argc and argv, which we store in static memory
+    // here so that we can defer the work of parsing them until its actually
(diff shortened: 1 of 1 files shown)


sunfishcode:
Responding to the comment in #87236, release/acquire feels like it would obscure the intent here. This code is just deferring parsing for the system argc/argv, which is why relaxed is sufficient, but it's also the reason we don't need a mutex. If someone ever adds Rust code to mutate the argv array, they may need more than just release/acquire on argc/argv; they may need to reintroduce the mutex too. And they may need to think about potential interactions with C code which may also be able to observe the system-provided argv array (eg. if it's also registered in a .init_array function).

RalfJung:
Thanks! @bors r+ rollup

> special memory ordering

I'd say `Relaxed` is the most special memory ordering, but we clearly have different perspectives here. ;)

PR #83086, opened 2021-03-13
make default anon const substs optional

lcnr:
*this description is currently outdated*

This is the least invasive way I could think of to get a `tcx` into the `super_visit_with` impl for consts.

Needed to deal with unused substs for anonymous constants. Note that `tcx_for_anon_const_substs` isn't actually used anywhere yet. Not sure whether it makes sense to merge this PR without actually implementing the optional const substs :sweat_smile: 

More info can be found in the [meeting notes](https://github.com/rust-lang/project-const-generics/blob/master/meetings/2021.03.09-unused-substs-impl.md) and on [zulip](https://rust-lang.zulipchat.com/#narrow/stream/260443-project-const-generics/topic/default-substs)

r? @nikomatsakis cc @oli-obk

Issue #78369, opened 2020-10-25
lazy_normalization_consts breaks unsize coercion in rare cases


lcnr:
```rust
#![feature(lazy_normalization_consts)]
struct P<T: ?Sized>([u8; 1 + 4], T);

fn main() {
    let x: Box<P<[u8; 0]>> = Box::new(P(Default::default(), [0; 0]));
    let _: Box<P<[u8]>> = x;
}
```
This works on stable but fails with the following error when using `lazy_normalization_consts`
```
error[E0308]: mismatched types
 --> src/main.rs:6:27
  |
6 |     let _: Box<P<[u8]>> = x;
  |            ------------   ^ expected slice `[u8]`, found array `[u8; 0]`
  |            |
  |            expected due to this
  |
  = note: expected struct `Box<P<[u8]>>`
             found struct `Box<P<[u8; 0]>>`
```
For this to fail the struct has to contain a type referencing an unevaluated constant, `struct P<T: ?Sized>([u8; 5], T)` does work.

assigning to myself, but won't get to this right away. If you are interested in working on this before that feel free to do so.

cc @nikomatsakis @varkor

lcnr:
The issue is here
https://github.com/rust-lang/rust/blob/f5230fbf76bafd86ee4376a0e26e551df8d17fec/compiler/rustc_middle/src/ty/walk.rs#L199

We currently look at all substs while we should only look at the generic params which are actually mentioned by the constant.
The correct solution would be to instead walk the `AbstractConst` in case it exists, but for that we would have to move that to `rustc_middle` and also need to access the `TyCtxt`  everywhere which seems difficult.

I am currently thinking about moving `AbstractConst`s directly into `ConstKind` as the current `ConstKind::Unevaluated` is really unhelpful when dealing with anonymous constants. That would really simplify dealing with them as we can then just treat them pretty much exactly like other projections, :thinking: would be a big change though, so it might make sense allocate some time to discuss this before doing anything stupid here...

cc @oli-obk

oli-obk:
I don't think we can remove `Unevaluated`. There are cases where we can neither build an abstract const nor evaluate it to a concrete constant. And once we have an abstract const, we can't turn that abstract const into a concrete one, so we'd need to keep the information about the uneavluated source around anyway. Can you elaborate on the changes you were thinking about?

lcnr:
[...]

PR #87280, opened 2021-07-19
lazily "compute" anon const default substs

lcnr:
Continuing the work of #83086, this implements the discussed solution for the [unused substs problem](https://github.com/rust-lang/project-const-generics/blob/master/design-docs/anon-const-substs.md#unused-substs). As of now, anonymous constants inherit all of their parents generics, even if they do not use them, e.g. in `fn foo<T, const N: usize>() -> [T; N + 1]`, the array length has `T` as a generic parameter even though it doesn't use it. These *unused substs* cause some backwards incompatible, and imo incorrect behavior, e.g. #78369.

---
We do not actually filter any generic parameters here and the `default_anon_const_substs` query still a dummy which only checks that
- we now prevent the previously existing query cycles and are able to call `predicates_of(parent)` when computing the substs of anonymous constants
- the default anon consts substs only include the typeflags we assume it does.

Implementing that filtering will be left as future work.

---

The idea of this PR is to delay the creation of the anon const substs until after we've computed `predicates_of` for the parent of the anon const. As the predicates of the parent can however contain the anon const we still have to create a `ty::Const` for it.

We do this by changing the substs field of `ty::Unevaluated` to an option and modifying accesses to instead call the method `unevaluated.substs(tcx)` which returns the substs as before. If the substs - now `substs_` -  of `ty::Unevaluated` are `None`, it means that the anon const currently has its default substs, i.e. the substs it has when first constructed, which are the generic parameters it has available. To be able to call `unevaluated.substs(tcx)` in a `TypeVisitor`, we add the non-defaulted method `fn tcx_for_anon_const_substs(&self) -> Option<TyCtxt<'tcx>>`. In case `tcx_for_anon_const_substs` returns `None`, unknown anon const default substs are skipped entirely.

Even when `substs_` is `None` we still have to treat the constant as if it has its default substs. To do this, `TypeFlags` are modified so that it is clear whether they can still change when *exposing* any anon const default substs. A new flag, `HAS_UNKNOWN_DEFAULT_CONST_SUBSTS`, is added in case some default flags are missing.

The rest of this PR are some smaller changes to either not cause cycles by trying to access the default anon const substs too early or to be able to access the `tcx` in previously unused locations.

cc @rust-lang/project-const-generics 
r? @nikomatsakis

### DIFF
--- a/compiler/rustc_middle/src/ty/flags.rs
+++ b/compiler/rustc_middle/src/ty/flags.rs
@@ -305,7 +305,8 @@ impl FlagComputation {
     }
 
     fn add_unevaluated_const(&mut self, ct: ty::Unevaluated<'tcx>) {
-        self.add_substs(ct.substs);
+        // TODO
+        self.add_substs(ct.substs_.unwrap());
         self.add_flags(TypeFlags::HAS_CT_PROJECTION);
     }
 
--- a/compiler/rustc_middle/src/ty/flags.rs
+++ b/compiler/rustc_middle/src/ty/flags.rs
@@ -311,9 +311,20 @@ impl FlagComputation {
     }
 
     fn add_unevaluated_const<P>(&mut self, ct: ty::Unevaluated<'tcx, P>) {
+        // The generic arguments of unevaluated consts are a bit special,
+        // see the `rustc-dev-guide` for more information.
+        //
+        // FIXME(@lcnr): Actually add a link here.
         if let Some(substs) = ct.substs_ {
+            // If they are available, we treat them as ordinary generic arguments.
             self.add_substs(substs);
         } else {
+            // Otherwise, we add `HAS_UNKNOWN_DEFAULT_CONST_SUBSTS` to signify
+            // that our const may potentially refer to generic parameters.
+            //
+            // Note that depending on which generic parameters are actually
+            // used in this constant, we may not actually refer to any generic
+            // parameters at all.
             self.add_flags(TypeFlags::STILL_FURTHER_SPECIALIZABLE);
             self.add_flags(TypeFlags::HAS_UNKNOWN_DEFAULT_CONST_SUBSTS);
         }
--- a/compiler/rustc_middle/src/ty/flags.rs
+++ b/compiler/rustc_middle/src/ty/flags.rs
@@ -34,6 +34,12 @@ impl FlagComputation {
         result.flags
     }
 
+    pub fn for_unevaluated_const(uv: ty::Unevaluated<'_>) -> TypeFlags {
+        let mut result = FlagComputation::new();
+        result.add_unevaluated_const(uv);
+        result.flags
+    }
+
     fn add_flags(&mut self, flags: TypeFlags) {
         self.flags = self.flags | flags;
     }
@@ -91,7 +97,7 @@ impl FlagComputation {
             &ty::Error(_) => self.add_flags(TypeFlags::HAS_ERROR),
 
             &ty::Param(_) => {
-                self.add_flags(TypeFlags::HAS_TY_PARAM);
+                self.add_flags(TypeFlags::HAS_KNOWN_TY_PARAM);
                 self.add_flags(TypeFlags::STILL_FURTHER_SPECIALIZABLE);
             }
 
@@ -292,7 +298,7 @@ impl FlagComputation {
                 self.add_bound_var(debruijn);
             }
             ty::ConstKind::Param(_) => {
-                self.add_flags(TypeFlags::HAS_CT_PARAM);
+                self.add_flags(TypeFlags::HAS_KNOWN_CT_PARAM);
                 self.add_flags(TypeFlags::STILL_FURTHER_SPECIALIZABLE);
             }
             ty::ConstKind::Placeholder(_) => {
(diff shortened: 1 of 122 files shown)


oli-obk:
@bors try @rust-timer queue

just getting a baseline for any follow up changes

compiler/rustc_middle/src/ty/fold.rs
```diff
@@ -775,6 +788,10 @@ impl<'tcx> ValidateBoundVars<'tcx> {
 impl<'tcx> TypeVisitor<'tcx> for ValidateBoundVars<'tcx> {
     type BreakTy = ();
 
+    fn tcx_for_anon_const_substs(&self) -> TyCtxt<'tcx> {
```
nikomatsakis: Why *not* pass in a tcx everywhere? (and name this `fn tcx`?)

lcnr: Some visitors **must not** look into the default substs, as that would cause query cycles. For others a `TyCtxt` isn't
readily available.

The second issue can be fixed by adding the `TyCtxt` as an additional argument to a bunch of places, but I would like
to avoid doing that in this PR (and probably in general tbh).

The first issue could also be solved by overwriting `visit_unevaluated_const` for the affected visitors, but I prefer doing it this way for aesthetic reasons :laughing:

compiler/rustc_middle/src/ty/flags.rs
```diff
@@ -301,8 +307,12 @@ impl FlagComputation {
     }
 
     fn add_unevaluated_const(&mut self, ct: ty::Unevaluated<'tcx>) {
-        // TODO
-        self.add_substs(ct.substs_.unwrap());
+        if let Some(substs) = ct.substs_ {
+            self.add_substs(substs);
+        } else {
```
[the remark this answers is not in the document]
lcnr: >  see rustc-dev-guide

*subtly* implying that I have to write a section for the rustc-dev-guide for us to merge this :p

> When the substs are loaded, they will be the "identity substs" (is this true?) for the constant

not quite, going to write the `rustc-dev-guide` section soon, but e.g.
```rust
fn foo<const N: usize>() -> [u8; N + 1] { todo!() }

fn bar() {
    foo::<3>();
}
```
here the substs for the `N + 1` constant is `3` inside of `bar` while the default substs is the parameter `N`.

So if `substs_` is `Some`, it may either be the default substs - which are all used generic parameters of the parent and, in the future, the bound vars used by the const, e.g. `for<'a> Trait<[u8; foo::<'a>()]>`-  or an instantiation of these params.

> In other words, why do we need HAS_KNOWN_TY_PARAM vs HAS_TY_PARAM?

In other words, where does it cause problems to always assume something has generic parameters if there are unknown anon const substs? 

In wfcheck this would cause us to incorrectly not emit an error here if the default type if the default of a param contained an anon const with unknown substs:
https://github.com/rust-lang/rust/blob/effea9a2a0d501db5722d507690a1a66236933bf/compiler/rustc_typeck/src/check/wfcheck.rs#L856

Most of the other uses of `has_param_types_or_consts` would not cause any bugs as we call `var.subst(tcx, subst)` before that which exposes all default substs. Verifying that isn't trivial though and requires some subtle knowledge about how all of this works, so we should avoid relying on it imo.

For `HAS_FREE_REGIONS` the relevant method is `is_global` which uses the negated type flag, so if we always set `HAS_FREE_REGIONS` and use `var.is_global() == false` to influence user visible behavior this would also be a bug.

I guess that e.g. this would be an issue https://github.com/rust-lang/rust/blob/effea9a2a0d501db5722d507690a1a66236933bf/compiler/rustc_typeck/src/impl_wf_check/min_specialization.rs#L366

compiler/rustc_middle/src/ty/fold.rs
lcnr: I also unfortunately started with `fn tcx_for_anon_const_substs(&self) -> TyCtxt<'tcx>` and changed it to
`fn tcx_for_anon_const_substs(&self) -> Option<TyCtxt<'tcx>>` which isn't too great tbh :sweat_smile:

compiler/rustc_middle/src/ty/fold.rs
```diff
@@ -108,29 +124,44 @@ pub trait TypeFoldable<'tcx>: fmt::Debug + Clone {
...
+        !self.has_type_flags(TypeFlags::HAS_POTENTIAL_FREE_LOCAL_NAMES)
     }
 
     /// Indicates whether this value references only 'global'
     /// generic parameters that are the same regardless of what fn we are
     /// in. This is used for caching.
-    fn is_global(&self) -> bool {
-        !self.has_type_flags(TypeFlags::HAS_FREE_LOCAL_NAMES)
+    fn is_global(&self, tcx: TyCtxt<'tcx>) -> bool {
+        !self.definitely_has_type_flags(tcx, TypeFlags::HAS_KNOWN_FREE_LOCAL_NAMES)
```
lcnr, line 164: I don't really like the names of these 2 methods and we can't use `definitely_` and `potentially_` here.

We could use `has_local_names` instead but that seems a bit less clear that `is_global` :thinking:

rylev:
The performance hit was justified [here](https://github.com/rust-lang/rust/pull/87280#issuecomment-906288243) so marking it as such.

PR #87282, opened 2021-07-19
Ensure `./x.py dist` adheres to `build.tools`

### LINKED ISSUE #86436: Unable to build 1.53.0 from official tarball due to miri
Hi there, I tried to build Rust 1.53.0 from the source tarball downloaded from https://static.rust-lang.org/dist/rustc-1.53.0-src.tar.gz and it failed to compile due to `cargo-miri` could not locate the Git repository.

After a quick search, the issue supposedly should have been fixed in https://github.com/rust-lang/miri/commit/e5b340017e39dc90e3f9cf9966244faa585321b5 but for some reason, the fix didn't land in the final distribution despite in #84182 it was stated that the fix would be included in the next stable release.

Here is the backtrace:

<details><summary>Backtrace</summary>
<p>

```
  thread 'main' panicked at 'Unable to generate vergen keys!: could not find repository from '/build/i0szw9jw/rustc-1.53.0-src/src/tools/miri/cargo-miri'; class=Repository (6); code=NotFound (-3)

  Stack backtrace:
     0: anyhow::error::<impl core::convert::From<E> for anyhow::Error>::from
               at /build/i0szw9jw/rustc-1.53.0-src/vendor/anyhow-1.0.34/src/error.rs:464:25
     1: vergen::feature::git::configure_git
               at /build/i0szw9jw/rustc-1.53.0-src/vendor/vergen/src/feature/git.rs:175:55
     2: vergen::config::Instructions::config
               at /build/i0szw9jw/rustc-1.53.0-src/vendor/vergen/src/config.rs:139:9
     3: vergen::gen::config_from_instructions
               at /build/i0szw9jw/rustc-1.53.0-src/vendor/vergen/src/gen.rs:81:32
     4: vergen::gen::vergen
               at /build/i0szw9jw/rustc-1.53.0-src/vendor/vergen/src/gen.rs:69:5
     5: build_script_build::main
               at ./build.rs:10:5
     6: core::ops::function::FnOnce::call_once
               at /build/i0szw9jw/rustc-1.53.0-src/library/core/src/ops/function.rs:227:5
     7: std::sys_common::backtrace::__rust_begin_short_backtrace
               at /build/i0szw9jw/rustc-1.53.0-src/library/std/src/sys_common/backtrace.rs:125:18
     8: std::rt::lang_start::{{closure}}
               at /build/i0szw9jw/rustc-1.53.0-src/library/std/src/rt.rs:49:18
     9: core::ops::function::impls::<impl core::ops::function::FnOnce<A> for &F>::call_once
               at /build/i0szw9jw/rustc-1.53.0-src/library/core/src/ops/function.rs:259:13
        std::panicking::try::do_call
               at /build/i0szw9jw/rustc-1.53.0-src/library/std/src/panicking.rs:379:40
        std::panicking::try
               at /build/i0szw9jw/rustc-1.53.0-src/library/std/src/panicking.rs:343:19
        std::panic::catch_unwind
               at /build/i0szw9jw/rustc-1.53.0-src/library/std/src/panic.rs:431:14
        std::rt::lang_start_internal
               at /build/i0szw9jw/rustc-1.53.0-src/library/std/src/rt.rs:34:21
    10: std::rt::lang_start
               at /build/i0szw9jw/rustc-1.53.0-src/library/std/src/rt.rs:48:5
    11: main
    12: __libc_start_main
               at /build/okrh64bn/glibc/csu/../csu/libc-start.c:314:16
    13: _start
               at /build/okrh64bn/glibc/csu/../sysdeps/x86_64/start.S:120', src/tools/miri/cargo-miri/build.rs:10:24
  stack backtrace:
     0: rust_begin_unwind
               at /build/i0szw9jw/rustc-1.53.0-src/library/std/src/panicking.rs:493:5
     1: core::panicking::panic_fmt
               at /build/i0szw9jw/rustc-1.53.0-src/library/core/src/panicking.rs:92:14
     2: core::result::unwrap_failed
               at /build/i0szw9jw/rustc-1.53.0-src/library/core/src/result.rs:1355:5
     3: core::result::Result<T,E>::expect
               at /build/i0szw9jw/rustc-1.53.0-src/library/core/src/result.rs:997:23
     4: build_script_build::main
               at ./build.rs:10:5
     5: core::ops::function::FnOnce::call_once
               at /build/i0szw9jw/rustc-1.53.0-src/library/core/src/ops/function.rs:227:5
command did not execute successfully: "/build/i0szw9jw/rustc-1.53.0-src/build/x86_64-unknown-linux-gnu/stage0/bin/cargo" "build" "--target" "x86_64-unknown-linux-gnu" "-Zbinary-dep-depinfo" "-j" "32" "-v" "--release" "--frozen" "--manifest-path" "/build/i0szw9jw/rustc-1.53.0-src/src/tools/miri/cargo-miri/Cargo.toml" "--message-format" "json-render-diagnostics"
expected success, got: exit code: 101
```

</p>
</details>

emilyalbini:
According to `config.toml.example`, the way to produce dist artifacts for both the compiler and a *subset* of tools would be to enable the extended build and manually specify the list of tools to build:

```toml
[build]
extended = true
tools = ["cargo", "rustfmt"]
```

This works as expected for `./x.py build` and `./x.py install`, but *not* for `./x.py dist`. Before this PR `./x.py dist` simply ignored the contents of `build.tools`, building just rustc/rustdoc if `build.extended = false` and all of the tools otherwise. This PR does two things:

* Changes `./x.py dist extended` to only build the tools defined in `build.tools`, if `build.tools` is not empty. The rest of the extended step was refactored to simplify the code.
* Changes how dist jobs for tools are gated: instead of `assert!(builder.config.extended)` to prevent tools from being built with `build.extended = false`, tools are simply built by default depending on `build.extended` and `build.tools`. This also enables to **explicitly** dist tools even with `build.extended = false`.

This PR is best reviewed commit-by-commit.

Fixes #86436

### DIFF
--- a/src/bootstrap/builder.rs
+++ b/src/bootstrap/builder.rs
@@ -163,14 +163,8 @@ impl StepDescription {
     }
 
     fn maybe_run(&self, builder: &Builder<'_>, pathset: &PathSet) {
-        if builder.config.exclude.iter().any(|e| pathset.has(e)) {
-            eprintln!("Skipping {:?} because it is excluded", pathset);
+        if self.is_excluded(builder, pathset) {
             return;
-        } else if !builder.config.exclude.is_empty() {
-            eprintln!(
-                "{:?} not skipped for {:?} -- not in {:?}",
-                pathset, self.name, builder.config.exclude
-            );
         }
 
         // Determine the targets participating in this rule.
@@ -180,6 +174,21 @@ impl StepDescription {
             let run = RunConfig { builder, path: pathset.path(builder), target: *target };
             (self.make_run)(run);
         }
+    }
+
+    fn is_excluded(&self, builder: &Builder<'_>, pathset: &PathSet) -> bool {
+        if builder.config.exclude.iter().any(|e| pathset.has(e)) {
+            eprintln!("Skipping {:?} because it is excluded", pathset);
+            return true;
+        }
+
+        if !builder.config.exclude.is_empty() {
+            eprintln!(
+                "{:?} not skipped for {:?} -- not in {:?}",
+                pathset, self.name, builder.config.exclude
+            );
+        }
+        false
     }
 
     fn run(v: &[StepDescription], builder: &Builder<'_>, paths: &[PathBuf]) {
@@ -1578,6 +1587,27 @@ impl<'a> Builder<'a> {
         self.verbose(&format!("{}< {:?}", "  ".repeat(self.stack.borrow().len()), step));
         self.cache.put(step, out.clone());
         out
+    }
+
+    /// Ensure that a given step is built *only if it's supposed to be built by default*, returning
+    /// its output. This will cache the step, so it's safe (and good!) to call this as often as
+    /// needed to ensure that all dependencies are build.
+    pub(crate) fn ensure_if_default<T, S: Step<Output = Option<T>>>(
+        &'a self,
+        step: S,
+    ) -> S::Output {
+        let desc = StepDescription::from::<S>();
+        let should_run = (desc.should_run)(ShouldRun::new(self));
+
+        // Avoid running steps contained in --exclude
+        for pathset in &should_run.paths {
+            if desc.is_excluded(self, pathset) {
+                return None;
+            }
+        }
+
+        // Only execute if it's supposed to run as default
+        if desc.default && should_run.is_really_default() { self.ensure(step) } else { None }
     }
 }
 
--- a/src/bootstrap/dist.rs
+++ b/src/bootstrap/dist.rs
@@ -679,6 +679,7 @@ pub struct Analysis {
 
 impl Step for Analysis {
     type Output = Option<GeneratedTarball>;
+    const DEFAULT: bool = true;
 
     fn should_run(run: ShouldRun<'_>) -> ShouldRun<'_> {
         let default = should_build_extended_tool(&run.builder, "analysis");
@@ -958,6 +959,7 @@ pub struct Cargo {
 
 impl Step for Cargo {
     type Output = Option<GeneratedTarball>;
+    const DEFAULT: bool = true;
     const ONLY_HOSTS: bool = true;
 
     fn should_run(run: ShouldRun<'_>) -> ShouldRun<'_> {
@@ -1014,6 +1016,7 @@ pub struct Rls {
 impl Step for Rls {
     type Output = Option<GeneratedTarball>;
     const ONLY_HOSTS: bool = true;
+    const DEFAULT: bool = true;
 
     fn should_run(run: ShouldRun<'_>) -> ShouldRun<'_> {
         let default = should_build_extended_tool(&run.builder, "rls");
@@ -1059,6 +1062,7 @@ pub struct RustAnalyzer {
 
 impl Step for RustAnalyzer {
     type Output = Option<GeneratedTarball>;
+    const DEFAULT: bool = true;
     const ONLY_HOSTS: bool = true;
 
     fn should_run(run: ShouldRun<'_>) -> ShouldRun<'_> {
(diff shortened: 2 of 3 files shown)


paulmenzel:
Do you know if there was a regression?

According to issue https://github.com/rust-lang/rust/issues/86436 (Unable to build 1.53.0 from official tarball due to miri) this also does not work anymore with `install`.

pietroalbini:
Hmm, this PR doesn't change `./x.py install`, only `./x.py dist`. From a quick glance at the source  code `./x.py install` *should* adhere to `build.tools = ["..."]`, but I don't have much time to investigate that right now :(

src/bootstrap/dist.rs
```diff
@@ -45,6 +46,18 @@ fn missing_tool(tool_name: &str, skip: bool) {
     }
 }
 
+fn should_build_extended_tool(builder: &Builder<'_>, tool: &str) -> bool {
+    if !builder.config.extended {
+        return false;
+    }
+
+    if let Some(tools) = &builder.config.tools {
+        tools.is_empty() || tools.contains(tool)
```
Mark-Simulacrum, line 55: Can we refactor tools to be an `Option<Vec<...>>` or something to avoid `tools = []` to indicate all tools (which seems like not what the user wants)?

src/bootstrap/dist.rs
```diff
@@ -1344,6 +1357,17 @@ impl Step for Extended {
         builder.info(&format!("Dist extended stage{} ({})", compiler.stage, target));
 
         let mut tarballs = Vec::new();
+        let mut built_tools = HashSet::new();
+        macro_rules! add_tool {
+            ($name:expr => $step:expr) => {
+                if should_build_extended_tool(builder, $name) {
+                    if let Some(tarball) = builder.ensure($step) {
```
Mark-Simulacrum, line 1370: I think it's a larger refactor, but maybe not actually that hard to do -- can we instead of running builder.ensure here add a `builder.ensure_if_default` or something like it, which runs code similar to `builder.default_doc` (but constrained to just that particular step)?

That way we don't have to repeat the should_build_extended_tool here, and that seems overall nicer. It would also make `--exclude rustfmt` just work, I believe, whereas this currently wouldn't support that (which seems not nice).

Mark-Simulacrum:
Overall looks good -- I think the refactor I suggested *could* be nice, but is also fine to leave out from this PR. I think over the last few months I've started thinking about a new design for the default/should_run system anyway, which would likely address my suggestion in a slightly better way. So unless you're feeling particularly enthusiastic about the possible cleanup, let's leave it out for the time being.

I would like the first comment about tools = [] addressed though, that should be easy.

src/bootstrap/dist.rs
pietroalbini: I agree with your reading that `build.tools = []` should not build any tool instead of building all of them, I updated the code. By the way, that config option was already represented as `Option<HashSet<T>>`, I just implemented my check wrong.

src/bootstrap/dist.rs
pietroalbini: Pushed the refactoring!

pietroalbini:
Pushed the refactoring as well, so you can review it. **I haven't tested it behaves correctly in all the cases**, so let's wait until I do so before merging (reviewing is fine though).

pietroalbini:
Ok fixed a bug I found while testing, the PR should now work as expected. r? @Mark-Simulacrum

src/bootstrap/builder.rs
```diff
@@ -1579,6 +1588,27 @@ impl<'a> Builder<'a> {
...
+    pub(crate) fn ensure_if_default<T, S: Step<Output = Option<T>>>(
+        &'a self,
+        step: S,
+    ) -> S::Output {
+        let desc = StepDescription::from::<S>();
+        let should_run = (desc.should_run)(ShouldRun::new(self));
+
+        // Avoid running steps contained in --exclude
+        for pathset in &should_run.paths {
+            if desc.is_excluded(self, pathset) {
```
Mark-Simulacrum, line 1604: Hm -- this is going to end up printing a bunch of times that things are / are not excluded, which seems a little annoying (i.e., when this is invoked multiple times). Maybe we should make it a ensure-style query that gets its result cached...

Seems OK for this PR though, that printing is noisy anyway.

Issue #87169, opened 2021-07-15
Improve error messages for intra-doc links when there are backticks


GuillaumeGomez:
In #87078, we arrived at the following situation:

```
error: incompatible link kind for `Foo::bar`
  --> $DIR/field-ice.rs:5:6
   |
LL | /// [`Foo::bar()`]
   |      ^^^^^^^^^^^^ help: to link to the field, remove the disambiguator: ``Foo::bar``
```

I want to underline this part in particular:

```
``Foo::bar``
```

Having two backticks is very weird. What I suggest instead is to reduce the "^^^^^" underlining part in order to not include backticks in it so then they can be removed from the error message as well, which would give:

```
error: incompatible link kind for `Foo::bar`
  --> $DIR/field-ice.rs:5:6
   |
LL | /// [`Foo::bar()`]
   |       ^^^^^^^^^^ help: to link to the field, remove the disambiguator: `Foo::bar`
```

I'll do it in the next days if no one does until then. :)

cc @jyn514

estebank:
You can fix this by using a `span_suggestion_verbose` instead of `span_suggestion`. It'll take more vertical space, but it's ok.

GuillaumeGomez:
@estebank It's a bit different here because we create the spans ourselves because they come from inside a doc comment. So I'm sure if your suggestion will work out.

Also, in here it's `src/test/rustdoc-ui`, not `src/test/ui` (but otherwise it's the same). ;)

estebank:
@GuillaumeGomez thanks for the correction!

As long as the span was synthesized correctly (including what side of a char the boundary falls on), then just changing the method will work as expected.

GuillaumeGomez:
I had in mind to simply update the span instead in case the string was starting and ending with backticks.

jyn514:
@GuillaumeGomez look at the code that's already there: https://github.com/rust-lang/rust/blob/c49895d9049f67e07e297ee487836a587f69690e/src/librustdoc/passes/collect_intra_doc_links.rs#L2104-L2108
If you think this shouldn't use `verbose`, remove that instead. But I think it's better to show backticks if the user had them there originally.

GuillaumeGomez:
We can take advantage of the span to exclude the backticks from the underlining, so then we can avoid displaying the backticks twice in the suggestion. As for the how, I guess it's fine as long as we get the output we want.

GuillaumeGomez:
Fixed in #87285.

PR #87285, opened 2021-07-19
Improve intra doc errors display

GuillaumeGomez:
#87169

@jyn514 This is what I had in mind to avoid having duplicated backticks. I also gave a try to simply updating the span for the suggestion/help messages but I think this current one is better because less "noisy". Anyway, that allows you to see the result. ;)

### DIFF
--- a/src/librustdoc/passes/collect_intra_doc_links.rs
+++ b/src/librustdoc/passes/collect_intra_doc_links.rs
@@ -19,7 +19,7 @@ use rustc_resolve::ParentScope;
 use rustc_session::lint::Lint;
 use rustc_span::hygiene::{MacroKind, SyntaxContext};
 use rustc_span::symbol::{sym, Ident, Symbol};
-use rustc_span::DUMMY_SP;
+use rustc_span::{BytePos, DUMMY_SP};
 use smallvec::{smallvec, SmallVec};
 
 use pulldown_cmark::LinkType;
@@ -1193,7 +1193,7 @@ impl LinkCollector<'_, '_> {
         let report_mismatch = |specified: Disambiguator, resolved: Disambiguator| {
             // The resolved item did not match the disambiguator; give a better error than 'not found'
             let msg = format!("incompatible link kind for `{}`", path_str);
-            let callback = |diag: &mut DiagnosticBuilder<'_>, sp| {
+            let callback = |diag: &mut DiagnosticBuilder<'_>, sp: Option<rustc_span::Span>| {
                 let note = format!(
                     "this link resolved to {} {}, which is not {} {}",
                     resolved.article(),
@@ -1201,8 +1201,12 @@ impl LinkCollector<'_, '_> {
                     specified.article(),
                     specified.descr()
                 );
-                diag.note(&note);
-                suggest_disambiguator(resolved, diag, path_str, dox, sp, &ori_link.range);
+                if let Some(sp) = sp {
+                    diag.span_label(sp, &note);
+                } else {
+                    diag.note(&note);
+                }
+                suggest_disambiguator(resolved, diag, path_str, &ori_link.link, sp);
             };
             report_diagnostic(self.cx.tcx, BROKEN_INTRA_DOC_LINKS, &msg, &diag_info, callback);
         };
@@ -1699,6 +1703,51 @@ impl Suggestion {
             Self::RemoveDisambiguator => path_str.into(),
         }
     }
+
+    fn as_help_span(
+        &self,
+        path_str: &str,
+        ori_link: &str,
+        sp: rustc_span::Span,
+    ) -> Vec<(rustc_span::Span, String)> {
+        let inner_sp = match ori_link.find('(') {
+            Some(index) => sp.with_hi(sp.lo() + BytePos(index as _)),
+            None => sp,
+        };
+        let inner_sp = match ori_link.find('!') {
+            Some(index) => inner_sp.with_hi(inner_sp.lo() + BytePos(index as _)),
+            None => inner_sp,
+        };
+        let inner_sp = match ori_link.find('@') {
+            Some(index) => inner_sp.with_lo(inner_sp.lo() + BytePos(index as u32 + 1)),
+            None => inner_sp,
+        };
+        match self {
+            Self::Prefix(prefix) => {
+                // FIXME: if this is an implied shortcut link, it's bad style to suggest `@`
+                let mut sugg = vec![(sp.with_hi(inner_sp.lo()), format!("{}@", prefix))];
+                if sp.hi() != inner_sp.hi() {
+                    sugg.push((inner_sp.shrink_to_hi().with_hi(sp.hi()), String::new()));
+                }
+                sugg
+            }
+            Self::Function => {
+                let mut sugg = vec![(inner_sp.shrink_to_hi().with_hi(sp.hi()), "()".to_string())];
+                if sp.lo() != inner_sp.lo() {
+                    sugg.push((inner_sp.shrink_to_lo().with_lo(sp.lo()), String::new()));
+                }
+                sugg
+            }
+            Self::Macro => {
+                let mut sugg = vec![(inner_sp.shrink_to_hi(), "!".to_string())];
+                if sp.lo() != inner_sp.lo() {
+                    sugg.push((inner_sp.shrink_to_lo().with_lo(sp.lo()), String::new()));
+                }
+                sugg
+            }
+            Self::RemoveDisambiguator => return vec![(sp, path_str.into())],
+        }
+    }
 }
 
 /// Reports a diagnostic for an intra-doc link.
@@ -1732,7 +1781,16 @@ fn report_diagnostic(
     tcx.struct_span_lint_hir(lint, hir_id, sp, |lint| {
         let mut diag = lint.build(msg);
 
-        let span = super::source_span_for_markdown_range(tcx, dox, link_range, &item.attrs);
+        let span =
+            super::source_span_for_markdown_range(tcx, dox, link_range, &item.attrs).map(|sp| {
+                if dox.bytes().nth(link_range.start) == Some(b'`')
+                    && dox.bytes().nth(link_range.end - 1) == Some(b'`')
+                {
+                    sp.with_lo(sp.lo() + BytePos(1)).with_hi(sp.hi() - BytePos(1))
+                } else {
+                    sp
+                }
+            });
 
         if let Some(sp) = span {
             diag.set_span(sp);
@@ -1938,9 +1996,8 @@ fn resolution_failure(
                                 disambiguator,
                                 diag,
                                 path_str,
-                                diag_info.dox,
+                                diag_info.ori_link,
                                 sp,
-                                &diag_info.link_range,
                             )
                         }
 
@@ -2007,7 +2064,7 @@ fn anchor_failure(cx: &DocContext<'_>, diag_info: DiagnosticInfo<'_>, failure: A
             if let Some((fragment_offset, _)) =
                 diag_info.ori_link.char_indices().filter(|(_, x)| *x == '#').nth(anchor_idx)
             {
-                sp = sp.with_lo(sp.lo() + rustc_span::BytePos(fragment_offset as _));
+                sp = sp.with_lo(sp.lo() + BytePos(fragment_offset as _));
             }
             diag.span_label(sp, "invalid anchor");
         }
@@ -2075,14 +2132,7 @@ fn ambiguity_error(
 
         for res in candidates {
             let disambiguator = Disambiguator::from_res(res);
-            suggest_disambiguator(
-                disambiguator,
-                diag,
-                path_str,
-                diag_info.dox,
-                sp,
-                &diag_info.link_range,
-            );
+            suggest_disambiguator(disambiguator, diag, path_str, diag_info.ori_link, sp);
         }
     });
 }
@@ -2093,21 +2143,20 @@ fn suggest_disambiguator(
     disambiguator: Disambiguator,
     diag: &mut DiagnosticBuilder<'_>,
     path_str: &str,
-    dox: &str,
+    ori_link: &str,
     sp: Option<rustc_span::Span>,
-    link_range: &Range<usize>,
 ) {
     let suggestion = disambiguator.suggestion();
     let help = format!("to link to the {}, {}", disambiguator.descr(), suggestion.descr());
 
     if let Some(sp) = sp {
-        let msg = if dox.bytes().nth(link_range.start) == Some(b'`') {
-            format!("`{}`", suggestion.as_help(path_str))
+        let mut spans = suggestion.as_help_span(path_str, ori_link, sp);
+        if spans.len() > 1 {
+            diag.multipart_suggestion(&help, spans, Applicability::MaybeIncorrect);
         } else {
-            suggestion.as_help(path_str)
-        };
-
-        diag.span_suggestion(sp, &help, msg, Applicability::MaybeIncorrect);
+            let (sp, suggestion_text) = spans.pop().unwrap();
+            diag.span_suggestion_verbose(sp, &help, suggestion_text, Applicability::MaybeIncorrect);
+        }
     } else {
         diag.help(&format!("{}: {}", help, suggestion.as_help(path_str)));
     }
--- a/src/test/rustdoc-ui/intra-doc/ambiguity.stderr
+++ b/src/test/rustdoc-ui/intra-doc/ambiguity.stderr
@@ -12,26 +12,26 @@ LL | #![deny(rustdoc::broken_intra_doc_links)]
 help: to link to the module, prefix with `mod@`
    |
 LL | /// [mod@true]
-   |      ^^^^^^^^
+   |      ^^^^
 help: to link to the builtin type, prefix with `prim@`
    |
 LL | /// [prim@true]
-   |      ^^^^^^^^^
+   |      ^^^^^
 
 error: `ambiguous` is both a struct and a function
-  --> $DIR/ambiguity.rs:27:6
+  --> $DIR/ambiguity.rs:27:7
    |
 LL | /// [`ambiguous`] is ambiguous.
-   |      ^^^^^^^^^^^ ambiguous link
+   |       ^^^^^^^^^ ambiguous link
    |
 help: to link to the struct, prefix with `struct@`
    |
 LL | /// [`struct@ambiguous`] is ambiguous.
-   |      ^^^^^^^^^^^^^^^^^^
+   |       ^^^^^^^
 help: to link to the function, add parentheses
    |
 LL | /// [`ambiguous()`] is ambiguous.
-   |      ^^^^^^^^^^^^^
+   |                ^^
 
 error: `ambiguous` is both a struct and a function
   --> $DIR/ambiguity.rs:29:6
@@ -42,30 +42,30 @@ LL | /// [ambiguous] is ambiguous.
 help: to link to the struct, prefix with `struct@`
    |
 LL | /// [struct@ambiguous] is ambiguous.
-   |      ^^^^^^^^^^^^^^^^
+   |      ^^^^^^^
 help: to link to the function, add parentheses
    |
 LL | /// [ambiguous()] is ambiguous.
-   |      ^^^^^^^^^^^
+   |               ^^
 
 error: `multi_conflict` is a struct, a function, and a macro
-  --> $DIR/ambiguity.rs:31:6
+  --> $DIR/ambiguity.rs:31:7
    |
 LL | /// [`multi_conflict`] is a three-way conflict.
-   |      ^^^^^^^^^^^^^^^^ ambiguous link
+   |       ^^^^^^^^^^^^^^ ambiguous link
    |
 help: to link to the struct, prefix with `struct@`
    |
 LL | /// [`struct@multi_conflict`] is a three-way conflict.
-   |      ^^^^^^^^^^^^^^^^^^^^^^^
+   |       ^^^^^^^
 help: to link to the function, add parentheses
    |
 LL | /// [`multi_conflict()`] is a three-way conflict.
-   |      ^^^^^^^^^^^^^^^^^^
+   |                     ^^
 help: to link to the macro, add an exclamation mark
    |
 LL | /// [`multi_conflict!`] is a three-way conflict.
-   |      ^^^^^^^^^^^^^^^^^
+   |                     ^
 
 error: `type_and_value` is both a module and a constant
   --> $DIR/ambiguity.rs:33:16
@@ -76,26 +76,26 @@ LL | /// Ambiguous [type_and_value].
 help: to link to the module, prefix with `mod@`
    |
 LL | /// Ambiguous [mod@type_and_value].
-   |                ^^^^^^^^^^^^^^^^^^
+   |                ^^^^
 help: to link to the constant, prefix with `const@`
    |
 LL | /// Ambiguous [const@type_and_value].
-   |                ^^^^^^^^^^^^^^^^^^^^
+   |                ^^^^^^
 
 error: `foo::bar` is both an enum and a function
-  --> $DIR/ambiguity.rs:35:42
+  --> $DIR/ambiguity.rs:35:43
    |
 LL | /// Ambiguous non-implied shortcut link [`foo::bar`].
-   |                                          ^^^^^^^^^^ ambiguous link
+   |                                           ^^^^^^^^ ambiguous link
    |
 help: to link to the enum, prefix with `enum@`
    |
 LL | /// Ambiguous non-implied shortcut link [`enum@foo::bar`].
-   |                                          ^^^^^^^^^^^^^^^
+   |                                           ^^^^^
 help: to link to the function, add parentheses
    |
 LL | /// Ambiguous non-implied shortcut link [`foo::bar()`].
-   |                                          ^^^^^^^^^^^^
+   |                                                   ^^
 
 error: aborting due to 6 previous errors
 
--- a/src/test/rustdoc-ui/intra-doc/disambiguator-mismatch.stderr
+++ b/src/test/rustdoc-ui/intra-doc/disambiguator-mismatch.stderr
@@ -1,95 +1,139 @@
 error: incompatible link kind for `S`
-  --> $DIR/disambiguator-mismatch.rs:14:14
+  --> $DIR/disambiguator-mismatch.rs:16:14
    |
 LL | /// Link to [struct@S]
-   |              ^^^^^^^^ help: to link to the enum, prefix with `enum@`: `enum@S`
+   |              ^^^^^^^^ this link resolved to an enum, which is not a struct
    |
 note: the lint level is defined here
   --> $DIR/disambiguator-mismatch.rs:1:9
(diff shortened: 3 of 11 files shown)


src/librustdoc/passes/collect_intra_doc_links.rs
```diff
@@ -1732,7 +1732,14 @@ fn report_diagnostic(
     tcx.struct_span_lint_hir(lint, hir_id, sp, |lint| {
         let mut diag = lint.build(msg);
 
-        let span = super::source_span_for_markdown_range(tcx, dox, link_range, &item.attrs);
+        let span =
+            super::source_span_for_markdown_range(tcx, dox, link_range, &item.attrs).map(|sp| {
+                if dox.bytes().nth(link_range.start) == Some(b'`') {
```
jyn514, line 1737: It's not actually required that the backticks match - can you also check that the ending backtick is present?
```suggestion
                if dox.bytes().nth(link_range.start) == Some(b'`') && dox.bytes().nth(link_range.end) == Some(b'`') {
```

jyn514 (changes requested):
I still think it would be better to use @estebank's `verbose` suggestion, the backticks are misleading. Narrowing the span to remove the backticks seems useful, but I don't think it solves the original issue.

src/test/rustdoc-ui/intra-doc/ambiguity.stderr
```diff
@@ -83,19 +83,19 @@ LL | /// Ambiguous [const@type_and_value].
...
 help: to link to the enum, prefix with `enum@`
    |
 LL | /// Ambiguous non-implied shortcut link [`enum@foo::bar`].
-   |                                          ^^^^^^^^^^^^^^^
+   |                                           ^^^^^^^^^^^^^
 help: to link to the function, add parentheses
    |
 LL | /// Ambiguous non-implied shortcut link [`foo::bar()`].
-   |                                          ^^^^^^^^^^^^
+   |                                           ^^^^^^^^^^
```
estebank, line 98: Given that these are being shown as verbose already (because there's more than one suggestion), it'd be nice to make the span smaller, only covering the `enum@` and `()`.

src/librustdoc/passes/collect_intra_doc_links.rs
GuillaumeGomez: I was wondering originally why it wasn't checking both the beginning and the end. Well, I guess that answers my question. :)

src/test/rustdoc-ui/intra-doc/ambiguity.stderr
GuillaumeGomez: This is an interesting take. Although, not sure it would be an improvement in these cases because they ask to add things, not remove them.

GuillaumeGomez:
> I still think it would be better to use @estebank's `verbose` suggestion, the backticks are misleading. Narrowing the span to remove the backticks seems useful, but I don't think it solves the original issue.

@estebank confirmed that it was already using verbose because there are multiple messages, so on this part I think it's already fine?

What is unresolved with the current changes? My complain was that the error message was unclear because of the duplication of the backticks. So instead of including the whole intra-doc element, I only include the part in the backticks. If you have something else in mind, I can do it here as well if you want.

jyn514:
> @estebank confirmed that it was already using verbose because there are multiple messages, so on this part I think it's already fine?

It happened in that specific case because there happened to be multiple help messages. Almost always there will be only 0 or 1.

> What is unresolved with the current changes? My complain was that the error message was unclear because of the duplication of the backticks. So instead of including the whole intra-doc element, I only include the part in the backticks. If you have something else in mind, I can do it here as well if you want.

My concern is that showing backticks *at all* is confusing - showing 1 backtick is actually even more confusing because it's not obviously a bug. Take this diagnostic: https://github.com/rust-lang/rust/blob/7f0cb81c8e4743aacc37167a7299ea0a85f005af/src/test/rustdoc-ui/intra-doc/disambiguator-mismatch.stderr#L1-L5

That makes it look like you should *add* backticks when they weren't there before, I've seen people running into this in practice: https://github.com/rust-bitcoin/rust-lightning/pull/955#discussion_r653728459
Actually after reading that, I wonder if we should change this for *any* markdown suggestion ...

GuillaumeGomez:
> My concern is that showing backticks at all is confusing - showing 1 backtick is actually even more confusing because it's not obviously a bug. Take this diagnostic: 
> ...
> That makes it look like you should add backticks when they weren't there before, I've seen people running into this in practice: rust-bitcoin/rust-lightning#955 (comment)
> Actually after reading that, I wonder if we should change this for any markdown suggestion ...

We use backticks literally everywhere:

```
LL | fn mk_int() -> usize { let i: isize = 3; return i; }
   |                -----                            ^ expected `usize`, found `isize`
   |                |
   |                expected `usize` because of return type
```

So this is coherent with all other error outputs. But again, I might misunderstand you...

> It happened in that specific case because there happened to be multiple help messages. Almost always there will be only 0 or 1.

It's definitely worth adding a new test to ensure the output for an error with just one message then!

jyn514:
> We use backticks literally everywhere:

Yes, I understand that. In most parts of the compiler it's not ambiguous, because adding the backtick will cause the code to fail to compile. But in markdown it's valid syntax, and changes the meaning of the code. So we shouldn't suggest it, because it's not easy for the user to tell it's formatting and not part of the suggestion.

GuillaumeGomez:
I see, it's why I changed the span in order to only underline the parts without backticks so we can use the same logic. But at least I understand your point now.

src/test/rustdoc-ui/intra-doc/ambiguity.stderr
estebank: This would look like this:

```
error: `foo::bar` is both an enum and a function
  --> $DIR/ambiguity.rs:35:43
   |
LL | /// Ambiguous non-implied shortcut link [`foo::bar`].
   |                                           ^^^^^^^^ ambiguous link
   |
help: to link to the enum, prefix with `enum@`
   |
LL | /// Ambiguous non-implied shortcut link [`enum@foo::bar`].
   |                                           ^^^^^
help: to link to the function, add parentheses
   |
LL | /// Ambiguous non-implied shortcut link [`foo::bar()`].
   |                                                   ^^
```

which to my eyes looks easier to understand at a glance.

estebank:
I'm _for_ the change to make the span smaller. I would also prefer it if we could use `span_suggestion_verbose` more liberally than we currently are, except for situations where we are suggesting a complete substitution. The reason I haven't mass replaced them is because some people will complain about added verbosity, but at the same time a different subset of people does get confused on small changes (like adding`()` at the end, but pointing at the whole thing). This later group ends up frustrated thinking "But that's what I have!?". We have an opportunity to avoid that frustration.

GuillaumeGomez:
Definitely sounds like an improvement. I'm going to give it a try tomorrow!

GuillaumeGomez:
With the help of @estebank, I was able to greatly improve the error output. You can now admire! :smiley:

src/librustdoc/passes/collect_intra_doc_links.rs
```diff
@@ -1699,6 +1699,35 @@ impl Suggestion {
...
+        sp: rustc_span::Span,
+    ) -> (String, rustc_span::Span) {
+        // FIXME: if this is an implied shortcut link, it's bad style to suggest `@`
+        match self {
+            Self::Prefix(prefix) => {
+                let sp = if let Some(index) = ori_link.find('@') {
+                    sp.with_hi(sp.lo() + BytePos(index as u32 + 1))
+                } else {
+                    sp.shrink_to_lo()
+                };
```
estebank, line 1716: I usually end up writing these as a `match` (4 lines instead of 5), but it doesn't matter at all!
```suggestion
                let sp = match ori_link.find('@') {
                    Some(index) => sp.with_hi(sp.lo() + BytePos(index as u32 + 1)),
                    None => sp.shrink_to_lo(),
                };
```

GuillaumeGomez: This is actually a very good point! Updating.

src/test/rustdoc-ui/intra-doc/disambiguator-mismatch.stderr
```diff
@@ -2,94 +2,138 @@ error: incompatible link kind for `S`
...
+   |              ^^^^^
 
 error: incompatible link kind for `S`
   --> $DIR/disambiguator-mismatch.rs:24:14
    |
 LL | /// Link to [union@S]
-   |              ^^^^^^^ help: to link to the enum, prefix with `enum@`: `enum@S`
+   |              ^^^^^^^
    |
    = note: this link resolved to an enum, which is not a union
```
estebank, line 36: Can we make this `note` the primary span?

src/test/rustdoc-ui/intra-doc/disambiguator-mismatch.stderr
```diff
@@ -2,94 +2,138 @@ error: incompatible link kind for `S`
   --> $DIR/disambiguator-mismatch.rs:14:14
    |
 LL | /// Link to [struct@S]
-   |              ^^^^^^^^ help: to link to the enum, prefix with `enum@`: `enum@S`
+   |              ^^^^^^^^
```
estebank, line 5: I notice that this file has no test where _multiple_ alternatives should be suggested, could we add one?

src/test/rustdoc-ui/intra-doc/disambiguator-mismatch.stderr
GuillaumeGomez: What do you mean?

src/test/rustdoc-ui/intra-doc/disambiguator-mismatch.stderr
GuillaumeGomez: Sure!

src/test/rustdoc-ui/intra-doc/disambiguator-mismatch.stderr
```diff
@@ -2,94 +2,138 @@ error: incompatible link kind for `S`
...
    |
 LL | /// Link to [const@f]
-   |              ^^^^^^^ help: to link to the function, add parentheses: `f()`
+   |              ^^^^^^^
    |
    = note: this link resolved to a function, which is not a constant
+help: to link to the function, add parentheses
+   |
+LL | /// Link to [const@f()]
+   |                     ^^
```
jyn514, line 136: Hmm, this doesn't look right, it's suggesting that you include _both_ disambiguators. It should remove the prefix if it suggests a suffix.

src/test/rustdoc-ui/intra-doc/disambiguator-mismatch.stderr
```diff
@@ -2,94 +2,138 @@ error: incompatible link kind for `S`
...
    |
 LL | /// Link to [c()]
-   |              ^^^ help: to link to the constant, prefix with `const@`: `const@c`
+   |              ^^^
    |
    = note: this link resolved to a constant, which is not a function
+help: to link to the constant, prefix with `const@`
+   |
+LL | /// Link to [const@c()]
+   |              ^^^^^^
```
estebank, line 124: This is subtly incorrect: we're failing to remove the `()`. You have two options here: either use `multipart_suggestion`, which takes a `vec![(Span, String)]` of replacements, or for this case in particular, use the "whole span replacement" that we were already doing.

src/test/rustdoc-ui/intra-doc/disambiguator-mismatch.stderr
estebank, line 136: Same problem as before, I believe. I think we need to also remove the `const@` prefix, right?

GuillaumeGomez: Yes, @jyn514 just made the same remark. :)

src/test/rustdoc-ui/intra-doc/disambiguator-mismatch.stderr
estebank: Produce the following output by changing `err.note(&msg)` with `err.span_label(span, &msg)`:

```
error: incompatible link kind for `S`
  --> $DIR/disambiguator-mismatch.rs:24:14
   |
LL | /// Link to [union@S]
   |              ^^^^^^^ this link resolved to an enum, which is not a union
```

src/test/rustdoc-ui/intra-doc/disambiguator-mismatch.stderr
estebank: Also note that `multipart_suggestion` when given a vec with a single element behaves the same as `span_suggestion` (and also has a `_verbose` counterpart), so you don't need to special case the two modes if you go down that route.

src/test/rustdoc-ui/intra-doc/disambiguator-mismatch.stderr
GuillaumeGomez: We actually can't have two items with in the same name in a module. Did you have something else in mind maybe?

jyn514: @GuillaumeGomez there could be two items with the same name and different namespaces:
```
const C: usize = 0;
macro_rules! C { () => {}; }
fn C() {}
```

src/test/rustdoc-ui/intra-doc/disambiguator-mismatch.stderr
```diff
@@ -2,94 +2,127 @@ error: incompatible link kind for `S`
...
    |
 LL | /// Link to [derive@m]
-   |              ^^^^^^^^ help: to link to the macro, add an exclamation mark: `m!`
+   |              ^^^^^^^^ this link resolved to a macro, which is not a derive macro
+   |
+help: to link to the macro, add an exclamation mark
    |
-   = note: this link resolved to a macro, which is not a derive macro
+LL | /// Link to [derive@m!]
+   |                      ^
```
estebank, line 70: We should be removing the `derive@`, right?

src/test/rustdoc-ui/intra-doc/disambiguator-mismatch.stderr
```diff
@@ -2,94 +2,127 @@ error: incompatible link kind for `S`
...
    |
 LL | /// Link to [c()]
-   |              ^^^ help: to link to the constant, prefix with `const@`: `const@c`
+   |              ^^^ this link resolved to a constant, which is not a function
    |
-   = note: this link resolved to a constant, which is not a function
+help: to link to the constant, prefix with `const@`
+   |
+LL | /// Link to [const@c()]
+   |              ^^^^^^
```
estebank, line 114: This needs to remove the `()`, right?

src/test/rustdoc-ui/intra-doc/disambiguator-mismatch.stderr
```diff
@@ -2,94 +2,127 @@ error: incompatible link kind for `S`
...
    |
 LL | /// Link to [const@f]
-   |              ^^^^^^^ help: to link to the function, add parentheses: `f()`
+   |              ^^^^^^^ this link resolved to a function, which is not a constant
+   |
+help: to link to the function, add parentheses
    |
-   = note: this link resolved to a function, which is not a constant
+LL | /// Link to [const@f()]
+   |                     ^^
```
estebank, line 125: This needs to remove the `const@`.

src/test/rustdoc-ui/intra-doc/errors.stderr
```diff
@@ -92,37 +92,45 @@ error: unresolved link to `Vec::into_iter`
...
-   |      ^^^^^^^^^^^^^^^^^^^
-   |      |
-   |      this link resolves to the associated function `into_iter`, which is not in the type namespace
-   |      help: to link to the associated function, add parentheses: `Vec::into_iter()`
+   |      ^^^^^^^^^^^^^^^^^^^ this link resolves to the associated function `into_iter`, which is not in the type namespace
+   |
+help: to link to the associated function, add parentheses
+   |
+LL | /// [type@Vec::into_iter()]
+   |                         ^^
```
estebank, line 100: This needs to remove `type@`.

src/test/rustdoc-ui/intra-doc/errors.stderr
```diff
@@ -92,37 +92,45 @@ error: unresolved link to `Vec::into_iter`
...
-   |      ^^
-   |      |
-   |      this link resolves to the struct `S`, which is not in the macro namespace
-   |      help: to link to the struct, prefix with `struct@`: `struct@S`
+   |      ^^ this link resolves to the struct `S`, which is not in the macro namespace
+   |
+help: to link to the struct, prefix with `struct@`
+   |
+LL | /// [struct@S!]
+   |      ^^^^^^^
```
estebank, line 111: This needs to remove the trailing `!`.

src/test/rustdoc-ui/intra-doc/errors.stderr
```diff
@@ -92,37 +92,45 @@ error: unresolved link to `Vec::into_iter`
...
-   |      ^^^^^^^^^
-   |      |
-   |      this link resolves to the associated function `h`, which is not in the type namespace
-   |      help: to link to the associated function, add parentheses: `S::h()`
+   |      ^^^^^^^^^ this link resolves to the associated function `h`, which is not in the type namespace
+   |
+help: to link to the associated function, add parentheses
+   |
+LL | /// [type@S::h()]
+   |               ^^
```
estebank, line 122: This needs to remove the `type@`

src/test/rustdoc-ui/intra-doc/errors.stderr
```diff
@@ -134,10 +142,12 @@ error: unresolved link to `m`
...
-   |      ^^^
-   |      |
-   |      this link resolves to the macro `m`, which is not in the value namespace
-   |      help: to link to the macro, add an exclamation mark: `m!`
+   |      ^^^ this link resolves to the macro `m`, which is not in the value namespace
+   |
+help: to link to the macro, add an exclamation mark
+   |
+LL | /// [m!()]
+   |       ^
```
estebank, line 150: For this case the replacement span was pointing at the point immediately after `m`, but it needs to start there and end at the current `sp.hi()`.

src/test/rustdoc-ui/intra-doc/disambiguator-mismatch.stderr
GuillaumeGomez: It doesn't seem to have a `_verbose` counterpart?

src/librustdoc/passes/collect_intra_doc_links.rs
```diff
@@ -1699,6 +1703,33 @@ impl Suggestion {
...
+        ori_link: &str,
+        sp: rustc_span::Span,
+    ) -> (String, rustc_span::Span) {
+        // FIXME: if this is an implied shortcut link, it's bad style to suggest `@`
+        match self {
+            Self::Prefix(prefix) => {
+                let sp = match ori_link.find('@') {
+                    Some(index) => sp.with_hi(sp.lo() + BytePos(index as u32 + 1)),
+                    None => sp.shrink_to_lo(),
+                };
```
estebank, line 1719: I think we _always_ need to make this check.

src/librustdoc/passes/collect_intra_doc_links.rs
```diff
@@ -1699,6 +1703,33 @@ impl Suggestion {
...
+            Self::Macro => {
+                let sp = match ori_link.find('(') {
+                    Some(index) => sp.with_hi(sp.lo() + BytePos(index as _)),
+                    None => sp,
+                };
+                ("!".to_string(), sp.shrink_to_hi())
+            }
+            Self::RemoveDisambiguator => (path_str.into(), sp),
+        }
+    }
```
estebank, line 1732: I _think_ that the following code would do the trick. You will have to `use` `Span` (you already have a use list from `rustc_span`) and to change the caller to take a `Vec` instead of the tuple. You can feed that straight to `multpart_suggestion_verbose` and unless I made a mistake (it's been known to happen), it should be given correct suggestions now.

```suggestion
    fn as_help_span(
        &self,
        path_str: &str,
        ori_link: &str,
        sp: Span,
    ) -> Vec<(String, Span)> {
        // FIXME: if this is an implied shortcut link, it's bad style to suggest `@`
        let inner_sp = match ori_link.find("()") {
            Some(index) => sp.with_hi(sp.lo() + BytePos(index as _)),
            None => sp,
        };
        let inner_sp = match ori_link.find('@') {
            Some(index) => inner_sp.with_lo(inner_sp.lo() + BytePos(index as u32 + 1)),
            None => sp,
        };
        match self {
            Self::Prefix(prefix) => {
                let mut sugg = vec![(format!("{}@", prefix), sp.with_hi(inner_sp.lo()))];
                if sp.hi() != inner_sp.hi() {
                    sugg.push(("".to_string(), inner_sp.shrink_to_hi().with_hi(sp.hi()));
                }
                sugg
            }
            Self::Function => {
                let mut sugg = vec![("()".to_string(), inner_sp.shrink_to_hi().with_hi(sp.hi()))];
                if sp.lo() != inner_sp.lo() {
                    sugg.push(("".to_string(), inner_sp.shrink_to_lo().with_lo(sp.lo()));
                }
                sugg
            }
            Self::Macro => {
                let mut sugg = vec![("!".to_string(), inner_sp.shrink_to_hi().with_hi(sp.hi()))];
                if sp.lo() != inner_sp.lo() {
                    sugg.push(("".to_string(), inner_sp.shrink_to_lo().with_lo(sp.lo()));
                }
                sugg
            }
            Self::RemoveDisambiguator => {
                let mut sugg = vec![];
                if sp.lo() != inner_sp.lo() {
                    sugg.push(("".to_string(), sp.with_hi(inner_sp.lo())));
                }
                if sp.hi() != inner_sp.hi() {
                    sugg.push(("".to_string(), inner_sp.with_hi(sp.hi())));
                }
                sugg
            }
        }
```

src/test/rustdoc-ui/intra-doc/disambiguator-mismatch.stderr
GuillaumeGomez: Yes! I'm working on it. I simply pushed the another change you suggested (about the `note`).

src/test/rustdoc-ui/intra-doc/disambiguator-mismatch.stderr
GuillaumeGomez: Sadly, it's not handled yet apparently... :'(

GuillaumeGomez:
Updated!

src/librustdoc/passes/collect_intra_doc_links.rs
GuillaumeGomez: Wo thanks! A few things to modify but this completely fixed the issues I had with using `multipart_suggestion`!

src/librustdoc/passes/collect_intra_doc_links.rs
```diff
@@ -2093,21 +2143,20 @@ fn suggest_disambiguator(
...
+        if spans.len() > 1 {
+            diag.multipart_suggestion(&help, spans, Applicability::MaybeIncorrect);
         } else {
-            suggestion.as_help(path_str)
-        };
-
-        diag.span_suggestion(sp, &help, msg, Applicability::MaybeIncorrect);
+            let (sp, suggestion_text) = spans.pop().unwrap();
+            diag.span_suggestion_verbose(sp, &help, suggestion_text, Applicability::MaybeIncorrect);
+        }
```
estebank, line 2159: I _think_ that you can just always call `multipart_suggestion_verbose` to get the same effect. Otherwise, it's fine to keep this as is.

src/test/rustdoc-ui/intra-doc/errors.stderr
estebank: Somehow, even after all the other cases have been fixed, this one hasn't 🤔

src/librustdoc/passes/collect_intra_doc_links.rs
```diff
@@ -1699,6 +1703,51 @@ impl Suggestion {
...
+            }
+            Self::Function => {
+                let mut sugg = vec![(inner_sp.shrink_to_hi().with_hi(sp.hi()), "()".to_string())];
+                if sp.lo() != inner_sp.lo() {
+                    sugg.push((inner_sp.shrink_to_lo().with_lo(sp.lo()), String::new()));
+                }
+                sugg
+            }
+            Self::Macro => {
+                let mut sugg = vec![(inner_sp.shrink_to_hi(), "!".to_string())];
```
estebank, line 1742: This is needed to account for the `foo()` -> `foo!` change:

```suggestion
                let mut sugg = vec![(inner_sp.shrink_to_hi().with_hi(sp.hi()), "!".to_string())];
```

estebank (approved):
r=me after addressing the outstanding incorrect suggestion.

Love the results!

src/test/rustdoc-ui/intra-doc/disambiguator-mismatch.stderr
estebank: Ah! I added one, but in a PR that hasn't been merged yet. Disregard for now then.

src/librustdoc/passes/collect_intra_doc_links.rs
GuillaumeGomez: This function doesn't exist: https://doc.rust-lang.org/nightly/nightly-rustc/rustc_errors/diagnostic/struct.Diagnostic.html?search=multipart_suggestion_verbose#method.multipart_suggestion

src/librustdoc/passes/collect_intra_doc_links.rs
GuillaumeGomez: Good point, I'll add a specific check for that.

src/librustdoc/passes/collect_intra_doc_links.rs
estebank: Ah! I added one, but in a PR that hasn't been merged yet. Disregard for now then.

src/librustdoc/passes/collect_intra_doc_links.rs
GuillaumeGomez: Actually, it is a valid link to have `macro!()`, so I'll keep as is. :)

src/test/rustdoc-ui/intra-doc/disambiguator-mismatch.stderr
jyn514: Can you add a test so we can see for ourselves? It's ok that it's not handled, but we should at least know how it behaves.

jyn514:
I see that this has already been put in a rollup, but I'd really prefer to see a test for multiple overlaps in a namespace: https://github.com/rust-lang/rust/pull/87285#discussion_r679327006

src/test/rustdoc-ui/intra-doc/disambiguator-mismatch.stderr
GuillaumeGomez: I did: https://github.com/rust-lang/rust/pull/87285/files#diff-24d855006c44e30e19269dd00e9f07f72f50c5d0e97ceaacbc2f6fafe79eb2caR4

GuillaumeGomez:
I answered on the discussion: I added a function with the same name. ;)

PR #86157, opened 2021-06-09
Use Tera templates for rustdoc.

jsha:
Replaces a format!() call in layout::render with a template
expansion. Introduces a `templates` field in SharedContext so parts
of rustdoc can share pre-rendered templates.

This currently builds in a copy of the single template available, like
with static files. However, future work can make this live-loadable with
a perma-unstable flag, to make rustdoc developers' work easier.

Part of #84419.

Demo at https://hoffman-andrews.com/rust/tera/std/string/struct.String.html.

PR #87288, opened 2021-07-19
rustdoc: Restore --default-theme, etc, by restoring varname escaping

### LINKED ISSUE #87263: rustdoc no longer honours --default-theme
### Steps to reproduce

```
git clone https://salsa.debian.org/iwj/rust-vecdeque-stableix # any crate will do
cd rust-vecdeque-stableix
RUSTDOCFLAGS='--default-theme ayu' cargo doc
dir=$PWD
rm -rf d
mkdir d
cd d
HOME=$PWD firefox file://$dir/target/doc/vecdeque_stableix/index.html
```

### Expected results

The documentation appears in the "ayu" (dark) theme.

### Observed results

The documentation appears in the default (light) theme.

### Notes

This works on stable.  I haven't bisected it or anything.

Empirically, changing the default theme once works for any one browser profile with the default firefox settings.  AIUI the setting is intended to be recorded in web local storage.  However, my firefox configuration does not permit use of web storage or cookies, so it does not work.  This is one of the reasons for the `--default-theme` option's existence.

FTR I see this message in my JS console which seems to support this hypothesis: `Request to access cookie or storage on “file:///home/rustcargo/Rustup/Game/vecdeque-stableix/target/doc/vecdeque_stableix/struct.IntoIter.html” was blocked because we are blocking all storage access requests.`

### Meta

```
rustdoc 1.55.0-nightly (2f391da2e 2021-07-14)
binary: rustdoc
commit-hash: 2f391da2e6ac73faa3570b79de239fd8c0edf1a9
commit-date: 2021-07-14
release: 1.55.0-nightly
LLVM version: 12.0.1
```

ijackson:
In #86157 

    cd0f93193c84ddc6698f9b65909da71c084dcb74
    Use Tera templates for rustdoc.

dropped the following transformation from the keys of the default settings element's `data-` attribute names:

    .map(|(k, v)| format!(r#" data-{}="{}""#, k.replace('-', "_"), Escape(v)))

The `Escape` part is indeed no longer needed, because Tera does that for us.  But the massaging of `-` to `_` is needed, for the (bizarre) reasons explained in the new comments.

I have tested that the default theme function works again for me.  I have also verified that passing (in shell syntax)

    '--default-theme="zork&"'

escapes the value in the HTML.

Closes #87263


ijackson:
@jsha would you care to confirm my suppositions in my commit message, that your removal of the `-` to `_` transformation wasn't for some good reason that I haven't spotted?

 @jyn514 Not sure if you want to self-assign this for review...

src/librustdoc/html/static/js/storage.js
```diff
@@ -22,6 +22,8 @@ function getSettingValue(settingName) {
         return current;
     }
     if (settingsDataset !== null) {
+	// See the comment for `default_settings.into_iter()` etc. in
```
GuillaumeGomez, line 25: You failed the indent here. ;)

GuillaumeGomez:
Thanks for this fix! Could you add a test to prevent this regression to happen again please? If it cannot be checked in `src/test/rustdoc`, I'm sure it can be checked in `src/test/rustdoc-gui`.

jyn514:
@GuillaumeGomez I think it would need to be in src/test/run-make because it needs a compiletime flag. But that seems painful since I don't know how you'd test the JavaScript from the command line. Do you think it makes sense to add support for cli args to rustdoc-gui?

GuillaumeGomez:
We already support to pass some arguments in `rustdoc-gui`, so it can be extended without problem.

ijackson:
> But that seems painful since I don't know how you'd test the JavaScript from the command line.

I thought about adding a test for this.  It seems quite difficult.

In my game system [Otter](https://www.chiark.greenend.org.uk/~ianmdlvl/otter/docs/) I need to do this kind of testing and I have an arrangement involving firefox, geckodriver, bubblewrap, webdriver/thirtyfour, etc. etc. etc.  I have found this very complex (and it is also very slow, although this may matter less for `rust-lang/rust` whose tests already take quite a while).  It is a pain to work with.  Making it reliable has been really hard work (and even so I still seem t have a ~1/1000 probability of random lossage which I haven't diagnosed - some race, probably).  Having built and worked with such a thing, I think it would be a net disbenefit to add a webdriver-based in-browser test to `rust-lang/rust` for the benefit of rustdoc.

This particular regression would only have been detected by a test which assembled an HTML DOM from the rustdoc-generated HTML, and then ran the Javascript, and then looked at the resulting effective CSS to see what the style was.  I.e. it demands an HTML DOM implementation, which is basically a browser.  I'm not aware of anything significantly short of a browser for this, but perhaps someone can suggest something.

GuillaumeGomez:
Like said previously, take a look at `src/test/rustdoc-gui`. 😉

ijackson:
> Like said previously, take a look at `src/test/rustdoc-gui`. wink

I went and looked.  Good lord!  You've already got a thing.  Wow.  OK, I will see what I can do.

It might not happen right away.  Would it be OK with you to land the fix without the test?

ijackson:
```
  * branch              dd7f545a69e4b720407e458bf4ade0b207bbf9ee -> FETCH_HEAD
fatal: remote error: upload-pack: not our ref 9ad6e5b8f68ee4bcd85886e50b8b0a70cbb91a52
Errors during submodule fetch:
	src/tools/cargo
Error: Process completed with exit code 1.
```

???

GuillaumeGomez:
> > Like said previously, take a look at `src/test/rustdoc-gui`. wink
> 
> I went and looked. Good lord! You've already got a thing. Wow. OK, I will see what I can do.
> 
> It might not happen right away. Would it be OK with you to land the fix without the test?

We had a lot of regressions because we didn't add tests, so I'd really prefer to have the test included with the fix (makes it easier when going through PRs as well).

ijackson:
> > > Like said previously, take a look at `src/test/rustdoc-gui`. wink
> > 
> > 
> > I went and looked. Good lord! You've already got a thing. Wow. OK, I will see what I can do.
> > It might not happen right away. Would it be OK with you to land the fix without the test?
> 
> We had a lot of regressions because we didn't add tests, so I'd really prefer to have the test included with the fix (makes it easier when going through PRs as well).

:-(.  I think this test is rather more work than the fix.  I would also like to point out that I am fixing a thing that someone else broke (doing a thing I approve of, but, still...) and now a thing I actually use day to day is broken.

ijackson:
FTAOD I definitely see the value in the test.  It's in my interests to do the test so no-one breaks this again!

GuillaumeGomez:
I can write it for you if you prefer? But I'll need a "scenario" and what to check to ensure everything is working as expected.

jyn514:
@GuillaumeGomez I don't think it makes sense to block the fix on the test. It's enough to open an `E-needs-test` isuse.

ijackson:
> I can write it for you if you prefer?

That would be awesome!

> But I'll need a "scenario" and what to check to ensure everything is working as expected.

So I think the test would look something like this:
 * Run `cargo doc --default-theme=ayu` (or maybe `rustdoc` instead of `cargo`) on some crate (the detail of what is documented doesn't matter)
 * Load the result in a browser session (any page will do, eg the crate `index.html`).  (The browser session ought to have a fresh profile, so no existing web storage, but I guess your test isolation would ensure that anyway.)
 * Wait for page rendering to complete and the JS to run.  I am hoping your test framework has a thing already to do this.
 * Do "some check" that the display is as expected.  Either check the average luminance of a screenshot :-) or more sensibly check the computed style (or maybe, the declared style) of the `<body>` element.  Specifically, `background-color` which should be `white` with the default (light) theme and something dark for `ayu` (currently, `#0f1419`, apparently, but I suggest checking that every element is less than `0x40`, or some such).

ijackson:
I have double checked my test cases and the way to run it has to be either `rustdoc --default-theme=ayu` on a plain rust file with no dependencies (in my test, `rustdoc t.rs` generated a `doc/` directory containing a useable `doc/src/t/t.rs.html` reflecting the selected theme).

Or using `RUSTDOCFLAGS='--default-theme=ayu' cargo doc`

ijackson:
Squashed and force pushed in the hope that the retry won't hit that submodule problem.

GuillaumeGomez:
> @GuillaumeGomez I don't think it makes sense to block the fix on the test. It's enough to open an `E-needs-test` isuse.

Generally I agree but such issues tend to take very long to get tests, which is why I'm not super eager to merge without tests. The beta is "far away" so I prefer to write the test directly in the PR (even if I'm the one writing the test, doesn't matter).

ijackson:
> The job **`mingw-check`** failed! 

I could swear I had run rustfmt all this already.  Apparently I botched that.  Fixed now.

GuillaumeGomez:
It happens to me all the time as well. Welcome to the club I guess?

GuillaumeGomez:
Ok, seems like the CI is happy now. I'll add the CI test tomorrow.

GuillaumeGomez:
Ok, I found the issue: when I added the cargo config, it wasn't added because it's part of `.gitignore` and I only noticed it now... Fun times. :)

GuillaumeGomez:
Ah finally!

@bors: r+

ijackson:
Thanks, especially for writing the test!

PR #87298, opened 2021-07-20
memorialize Anna Harren in the bastion of the turbofish

boringcactus:
this seems fitting, at least to me.

### DIFF
--- a/src/test/ui/bastion-of-the-turbofish.rs
+++ b/src/test/ui/bastion-of-the-turbofish.rs
@@ -25,9 +25,12 @@
 //
 // My heart aches in sorrow, for I know I am defeated. Let this be a warning
 // to all those who come after. Here stands the bastion of the Turbofish.
+//
+// RIP Anna Harren, Guardian Angel of the Hallowed Turbofish. <3
 
 // See https://github.com/rust-lang/rust/pull/53562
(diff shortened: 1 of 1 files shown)


steveklabnik:
r+ rollup

I wish I had something better to say here than "Thank you, Anna." But I do not. Thank you, Anna.

src/test/ui/bastion-of-the-turbofish.rs
```diff
@@ -25,9 +25,12 @@
 //
 // My heart aches in sorrow, for I know I am defeated. Let this be a warning
 // to all those who come after. Here stands the bastion of the Turbofish.
+//
+// RIP Anna Harren, Guardian Angel of the Hallowed Turbofish. <3
 
 // See https://github.com/rust-lang/rust/pull/53562
 // and https://github.com/rust-lang/rfcs/pull/2527
+// and https://twitter.com/garblefart/status/1393236602856611843
```
glandium, line 33: I wonder what's going to happen to this link. The Twitter policy for accounts of deceased people is, AIUI, not settled at the moment, but they were talking about expiring accounts at some point... Would it be better to put a web archive link?

estebank: As long as the tweet itself isn't deleted, the link will survive account renaming, if it comes to that.

PR #83647, opened 2021-03-29
make `Sized` predicates coinductive

lcnr:
:shrug:

needs either an MCP/RFC and `wg-traits` signoff. While I think this is sound, I am not too knowledgeable about coinduction so I might be wrong here.

r? @nikomatsakis

PR #87304, opened 2021-07-20
Improve error messages for recursive GATs implicitly deriving `Sized`

### LINKED ISSUE #80626: Enum with GAT member fails to derive Sized
I tried this code:

```rust

trait Allocator {
    type Allocated<T>;
}

enum LinkedList<A: Allocator> {
    Head,
    Next(A::Allocated<Self>)
}

fn main() {}
```

I expected to see this happen: Enum successfully compiles and can be constructed.

Instead, this happened: 
```
error[E0275]: overflow evaluating the requirement '<A as Allocator>::Allocated<CLinkedList<A>>: Sized'
```

Interestingly, this compiles:

```rust

trait Allocator {
    type Allocated<T>;
}

enum CLinkedList<A: Allocator> {
    Head,
    Next(A::Allocated<Self>, i64)
}

fn main() {}
```

As does this:
```rust

trait Allocator {
    type Allocated<T>;
}

struct LinkedListStruct<A: Allocator> {
    value: i64,
    next: Option<A::Allocated<Self>>
}

fn main() {}
```

### Meta

`rustc --version --verbose`:
```
rustc 1.51.0-nightly (17eec1433 2021-01-01)
binary: rustc
commit-hash: 17eec1433c69972844dd228b5fe801f218e118c3
commit-date: 2021-01-01
release: 1.51.0-nightly
```

jedel1043:
Fixes #80626

- Implemented a new error message recommending the user to add a `?Sized` constraint in case a recursive GAT is found.
- Fixed a bug where a type would not be constrained to be `Sized` if it wasn't on the last position of an ADT.
- Added tests for both fixes.

### DIFF
--- a/compiler/rustc_trait_selection/src/traits/error_reporting/mod.rs
+++ b/compiler/rustc_trait_selection/src/traits/error_reporting/mod.rs
@@ -57,8 +57,9 @@ pub trait InferCtxtExt<'tcx> {
 
     fn maybe_report_recursive_gat(
         &self,
-        begin: &PredicateObligation<'tcx>,
-        end: &PredicateObligation<'tcx>,
+        err: &mut DiagnosticBuilder<'_>,
+        parent: &PredicateObligation<'tcx>,
+        child: &PredicateObligation<'tcx>,
     ) -> bool;
 
     /// The `root_obligation` parameter should be the `root_obligation` field
@@ -221,33 +222,37 @@ impl<'a, 'tcx> InferCtxtExt<'tcx> for InferCtxt<'a, 'tcx> {
 
     fn maybe_report_recursive_gat(
         &self,
-        begin: &PredicateObligation<'tcx>,
-        end: &PredicateObligation<'tcx>,
+        err: &mut DiagnosticBuilder<'_>,
+        parent: &PredicateObligation<'tcx>,
+        child: &PredicateObligation<'tcx>,
     ) -> bool {
-        let (head, tail) =
-            match (begin.predicate.kind().skip_binder(), end.predicate.kind().skip_binder()) {
+        let (parent, child) =
+            match (parent.predicate.kind().skip_binder(), child.predicate.kind().skip_binder()) {
                 (ty::PredicateKind::Trait(head, ..), ty::PredicateKind::Trait(tail, ..)) => {
                     (head, tail)
                 }
                 _ => return false,
             };
 
-        if Some(head.clone().def_id()) != self.tcx.lang_items().sized_trait() {
+        if Some(parent.clone().def_id()) != self.tcx.lang_items().sized_trait() {
             return false;
         }
 
-        let head_ty = head.self_ty();
-        let tail_ty = tail.self_ty();
-        debug!("report_overflow_error_cycle: head_ty = {:?}, tail_ty = {:?}", head_ty, tail_ty);
-
-        let head_kind = head_ty.kind();
-        let tail_kind = tail_ty.kind();
+        let parent_ty = parent.self_ty();
+        let child_ty = child.self_ty();
         debug!(
-            "report_overflow_error_cycle: head_kind = {:?}, tail_kind = {:?}",
-            head_kind, tail_kind
+            "report_overflow_error_cycle: parent_ty = {:?}, child_ty = {:?}",
+            parent_ty, child_ty
         );
 
-        let (substs, item_def_id) = match head_kind {
+        let parent_kind = parent_ty.kind();
+        let child_kind = child_ty.kind();
+        debug!(
+            "report_overflow_error_cycle: parent_kind = {:?}, child_kind = {:?}",
+            parent_kind, child_kind
+        );
+
+        let (substs, item_def_id) = match parent_kind {
             ty::Projection(ty::ProjectionTy { substs, item_def_id }) => (substs, item_def_id),
             _ => return false,
         };
@@ -278,7 +283,7 @@ impl<'a, 'tcx> InferCtxtExt<'tcx> for InferCtxt<'a, 'tcx> {
             .map(|gens| {
                 gens.iter()
                     .enumerate()
-                    .find_map(|(i, param)| if param == &tail_ty.into() { Some(i) } else { None })
+                    .find_map(|(i, param)| if param == &child_ty.into() { Some(i) } else { None })
             })
             .flatten()
         {
@@ -289,16 +294,6 @@ impl<'a, 'tcx> InferCtxtExt<'tcx> for InferCtxt<'a, 'tcx> {
 
         let param = &assoc_generics.params[index];
         debug!("report_overflow_error_cycle: param = {:?}", param);
-
-        let mut err = struct_span_err!(
-            self.tcx.sess,
-            begin.cause.span,
-            E0275,
-            "overflow evaluating the requirement `{}`",
-            head
-        );
-
-        err.note("detected recursive derive for `Sized` in a generic associated type");
 
         let (span, separator) = match param.bounds {
             [] => (param.span.shrink_to_hi(), ":"),
@@ -311,8 +306,6 @@ impl<'a, 'tcx> InferCtxtExt<'tcx> for InferCtxt<'a, 'tcx> {
             Applicability::MachineApplicable,
         );
 
-        err.emit();
-        self.tcx.sess.abort_if_errors();
         true
     }
 
@@ -328,12 +321,24 @@ impl<'a, 'tcx> InferCtxtExt<'tcx> for InferCtxt<'a, 'tcx> {
         debug!("report_overflow_error_cycle: cycle={:?}", cycle);
 
         let begin = &cycle[0];
-        let end = &cycle[cycle.len() - 1];
+        let mut err = struct_span_err!(
+            self.tcx.sess,
+            begin.cause.span,
+            E0275,
+            "overflow evaluating the requirement `{}`",
+            begin.predicate
+        );
+        err.note("detected recursive derive for `Sized` in a generic associated type");
 
-        if !self.maybe_report_recursive_gat(begin, end) {
+        if cycle.windows(2).fold(false, |valid_error, preds| {
+            valid_error || self.maybe_report_recursive_gat(&mut err, &preds[0], &preds[1])
+        }) {
+            err.emit();
+            self.tcx.sess.abort_if_errors();
+            bug!()
+        } else {
             self.report_overflow_error(begin, false);
         }
-        bug!()
     }
 
     fn report_selection_error(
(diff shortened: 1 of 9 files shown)


compiler/rustc_trait_selection/src/traits/error_reporting/mod.rs
```diff
@@ -213,6 +219,103 @@ impl<'a, 'tcx> InferCtxtExt<'tcx> for InferCtxt<'a, 'tcx> {
...
+                gens.iter()
+                    .enumerate()
+                    .find_map(|(i, param)| if param == &tail_ty.into() { Some(i) } else { None })
+            })
+            .flatten()
+        {
+            Some(index) => index,
+            None => return false,
+        };
+        debug!("report_overflow_error_cycle: index = {:?}", index);
```
jedel1043, line 288: Is there a better way to do this? I did what I could do, but I'm not 100% sure this would work on every case.

compiler/rustc_trait_selection/src/traits/error_reporting/mod.rs
```diff
@@ -213,6 +219,103 @@ impl<'a, 'tcx> InferCtxtExt<'tcx> for InferCtxt<'a, 'tcx> {
...
+
+        assert!(substs.len() >= parameter_size);
+
+        let index = match substs
+            .get(substs.len() - parameter_size..)
+            .map(|gens| {
+                gens.iter()
+                    .enumerate()
+                    .find_map(|(i, param)| if param == &tail_ty.into() { Some(i) } else { None })
+            })
```
jedel1043, line 282: Is there a better way to do this? I did what I could, but I'm still not 100% sure this would work on every case.

compiler/rustc_trait_selection/src/traits/error_reporting/mod.rs
```diff
@@ -213,6 +219,103 @@ impl<'a, 'tcx> InferCtxtExt<'tcx> for InferCtxt<'a, 'tcx> {
...
+        let mut err = struct_span_err!(
+            self.tcx.sess,
+            begin.cause.span,
+            E0275,
+            "overflow evaluating the requirement `{}`",
+            head
+        );
+
+        err.note("detected recursive derive for `Sized` in a generic associated type");
+
```
jedel1043, line 302: When we have something like `Type<T: Sized>` this suggestion will recommend to do `Type<T: Sized + ?Sized>`, which is just incorrect. Maybe we could instead suggest to replace `Sized` with `?Sized` when there's a `Sized` inside `param.bounds`.
Nonetheless, this works for the trivial case.

compiler/rustc_trait_selection/src/traits/error_reporting/mod.rs
```diff
@@ -213,6 +219,103 @@ impl<'a, 'tcx> InferCtxtExt<'tcx> for InferCtxt<'a, 'tcx> {
...
+        let (span, separator) = match param.bounds {
+            [] => (param.span.shrink_to_hi(), ":"),
+            [.., bound] => (bound.span().shrink_to_hi(), " +"),
+        };
+        err.span_suggestion_verbose(
+            span,
```
jedel1043, line 308: When we have something like Type<T: Sized> this suggestion will recommend to do Type<T: Sized + ?Sized>, which is just incorrect. Maybe we could instead suggest to replace Sized with ?Sized when there's a Sized inside param.bounds.
Nonetheless, this works for the trivial case.

compiler/rustc_trait_selection/src/traits/error_reporting/mod.rs
```diff
@@ -213,6 +219,103 @@ impl<'a, 'tcx> InferCtxtExt<'tcx> for InferCtxt<'a, 'tcx> {
...
+            head
+        );
+
+        err.note("detected recursive derive for `Sized` in a generic associated type");
+
+        let (span, separator) = match param.bounds {
+            [] => (param.span.shrink_to_hi(), ":"),
+            [.., bound] => (bound.span().shrink_to_hi(), " +"),
+        };
+        err.span_suggestion_verbose(
```
jedel1043, line 307: When we have something like Type<T: Sized> this suggestion will recommend to do Type<T: Sized + ?Sized>, which is just incorrect. Maybe we could instead suggest to replace Sized with ?Sized when there's a Sized inside param.bounds.
Nonetheless, this works for the trivial case.

compiler/rustc_trait_selection/src/traits/error_reporting/mod.rs
```diff
@@ -213,6 +219,103 @@ impl<'a, 'tcx> InferCtxtExt<'tcx> for InferCtxt<'a, 'tcx> {
...
+            span,
+            "consider relaxing the implicit `Sized` restriction",
+            format!("{} ?Sized", separator),
+            Applicability::MachineApplicable,
+        );
```
jedel1043, line 312: When we have something like Type<T: Sized> this suggestion will recommend to do Type<T: Sized + ?Sized>, which is just incorrect. Maybe we could instead suggest to replace Sized with ?Sized when there's a Sized inside param.bounds.
Nonetheless, this works for the trivial case.

compiler/rustc_trait_selection/src/traits/error_reporting/mod.rs
```diff
@@ -213,6 +219,103 @@ impl<'a, 'tcx> InferCtxtExt<'tcx> for InferCtxt<'a, 'tcx> {
...
+            .map(|gens| {
+                gens.iter()
+                    .enumerate()
+                    .find_map(|(i, param)| if param == &tail_ty.into() { Some(i) } else { None })
+            })
+            .flatten()
+        {
+            Some(index) => index,
+            None => return false,
+        };
```
jedel1043, line 287: Is there a better way to do this? I did what I could, but I'm still not 100% sure this would work on every case.

src/test/ui/generic-associated-types/issue-80626.rs
```diff
@@ -0,0 +1,60 @@
...
+enum LinkedList2<A: Allocator> {
+    Head,
+    Next(A::Allocated<Self>, u8), //~ ERROR overflow evaluating the requirement
+}
+
+impl Allocator for () {
+    type Allocated<T> = Box<T>;
+}
+
+fn main() {
```
jackh726, line 23: I don't think we need to test runtime behavior here; this can just be empty with a `// check-pass`

src/test/ui/generic-associated-types/issue-80626.stderr
```diff
@@ -0,0 +1,27 @@
...
+   |
+LL |     Next(u8, A::Allocated<Self>),
+   |              ^^^^^^^^^^^^^^^^^^
+   |
+   = note: detected recursive derive for `Sized` in a generic associated type
+help: consider relaxing the implicit `Sized` restriction
+   |
+LL |     type Allocated<T: ?Sized>;
+   |                     ^^^^^^^^
+
```
jackh726, line 12: On one hand...yes, this is a suggestion that should work. On the other, is this a suggestion that is normally helpful?

compiler/rustc_ty_utils/src/ty.rs
```diff
@@ -193,7 +193,7 @@ fn adt_sized_constraint(tcx: TyCtxt<'_>, def_id: DefId) -> ty::AdtSizedConstrain
     let result = tcx.mk_type_list(
         def.variants
             .iter()
-            .flat_map(|v| v.fields.last())
+            .flat_map(|v| &v.fields)
```
jackh726, line 196: So, this isn't needed (i.e. this isn't a bug). We know that all fields except the last for all variants must be `Sized` for the struct to be well formed. (This check is done elsewhere.)

compiler/rustc_trait_selection/src/traits/error_reporting/mod.rs
```diff
@@ -213,6 +220,95 @@ impl<'a, 'tcx> InferCtxtExt<'tcx> for InferCtxt<'a, 'tcx> {
...
+        let parent_ty = parent.self_ty();
+        let child_ty = child.self_ty();
+        debug!(
+            "report_overflow_error_cycle: parent_ty = {:?}, child_ty = {:?}",
+            parent_ty, child_ty
+        );
+
+        let parent_kind = parent_ty.kind();
+        let child_kind = child_ty.kind();
+        debug!(
```
jackh726, line 250: FYI you can do `debug!(?parent_kind, ?child_kind)`. But, I feel like there are a lot of redundant debug statements here...(for example, we should get all the same info from the debug call above)

compiler/rustc_trait_selection/src/traits/error_reporting/mod.rs
```diff
@@ -213,6 +220,95 @@ impl<'a, 'tcx> InferCtxtExt<'tcx> for InferCtxt<'a, 'tcx> {
...
+            "report_overflow_error_cycle: parent_kind = {:?}, child_kind = {:?}",
+            parent_kind, child_kind
+        );
+
+        let (substs, item_def_id) = match parent_kind {
+            ty::Projection(ty::ProjectionTy { substs, item_def_id }) => (substs, item_def_id),
+            _ => return false,
+        };
+
+        let hir_id = self.tcx.hir().local_def_id_to_hir_id(item_def_id.expect_local());
```
jackh726, line 260: So, this would fail if the parent is defined in another crate?

compiler/rustc_trait_selection/src/traits/error_reporting/mod.rs
```diff
@@ -213,6 +220,95 @@ impl<'a, 'tcx> InferCtxtExt<'tcx> for InferCtxt<'a, 'tcx> {
...
+                    .find_map(|(i, param)| if param == &child_ty.into() { Some(i) } else { None })
+            })
+            .flatten()
+        {
+            Some(index) => index,
+            None => return false,
+        };
+        debug!("report_overflow_error_cycle: index = {:?}", index);
+
+        let param = &assoc_generics.params[index];
```
jackh726, line 295: This all feels a bit weird to me. But running out of time right now to think more about this.

compiler/rustc_trait_selection/src/traits/error_reporting/mod.rs
jackh726: Yeah, there should be a flag about implicit `Sized` somewhere and we should probably honor that.

jackh726 (commented):
Ran out of time, but some comments...

Overall, I guess this is just to give the "you can add a `?Sized`" suggestion? I'm on the fence on if that's actually helpful. Niko and I discussed this but yesterday; the proper fix is to make `Sized` coinductive (#83647), but that's a bit more than just GATs.

jedel1043:
> Overall, I guess this is just to give the "you can add a `?Sized`" suggestion? I'm on the fence on if that's actually helpful. Niko and I discussed this but yesterday; the proper fix is to make `Sized` coinductive (#83647), but that's a bit more than just GATs.

So the proper fix would be to make `Sized` coinductive, and then adjust `ty_is_representable` to detect when a trait implementor causes a recursive loop?

PR #87315, opened 2021-07-20
Add docs for raw-dylib to unstable book

ricobbe:


ricobbe:
I'm not sure how to provide an example of the feature that will compile in automation: it's only supported on Windows.  Further, I don't know if automation tries to link and run the examples from the unstable book, but this particular example depends on an external library.

Is there a way to mark this particular example as exempt from the compilation requirement?

wesleywiser:
You should be able to mark the codeblock with `ignore` like this:

https://github.com/rust-lang/rust/blob/673d0db5e393e9c64897005b470bfeb6d5aec61b/src/doc/unstable-book/src/language-features/auto-traits.md#L19-L21

wesleywiser (approved):
This accurately describes the current state so 👍 but we'll want to make sure as the rest of the feature is implemented, this is revised to remove outdated limitation information.

Issue #81088, opened 2021-01-16
asm! should automatically rewrite labels in assembly code


Amanieu:
Defining labels in inline assembly is broken because inlining, function cloning and other compiler optimizations may result in an `asm!` block being emitted more than once in the final binary. This can cause duplicate symbol errors or even compiler crashes (#74262). For this reason it is recommended to only use GAS local labels when using `asm!` (#76704).

To improve user experience when using `asm!`, rustc should automatically rewrite label names into a unique name. For example:
```rust

asm!(
    "popcnt {popcnt}, {v}",
    "loop:",
    "blsi rax, {v}",
    "jz exit",
    "xor {v}, rax",
    "tzcnt rax, rax",
    "stosb",
    "jmp loop",
    "exit:",
    v = inout(reg) value => _,
    popcnt = out(reg) popcnt,
    out("rax") _, // scratch
    inout("rdi") bits.as_mut_ptr() => _,
);
```

rustc should search for lines beginning with `<ident>:` to find label names and then replace all instances of `<ident>` in the assembly code with `.Lasm_label_<ident><unique id>`. The `.L` instructs the assembler that the label is local to the current compilation unit and should never be exported, even as a debug symbol. The unique ID can be generated by LLVM using `${:uid}` and is guaranteed to be unique within the current compilation unit.

nagisa:
Any reason to bother with `.L` and `${:uid}` instead of just replacing them with numbered labels?

Amanieu:
Internally the assembler will turn numbered labels into `.L` local symbols. Doing it directly makes the implementation slightly simpler since we don't need to track whether a label is before or after a particular instruction.

Amanieu:
@asquared31415 Thanks for taking up this issue!

I would suggest doing the rewriting in the backend in `compiler/rustc_codegen_llvm/src/asm.rs`.

asquared31415:
I already have an implementation working in the simple cases, and was about to submit a pr when I realized it is going to be a lot more work to fix the edge cases and that I should claim these issues so someone else doesn't try to work on them too

joshtriplett:
Reposting a comment that I left on #81570:

Forcing all labels within an asm block to be local would certainly improve usability for the common case (so that people don't have to use local labels).

[...]

PR #87324, opened 2021-07-20
Lint against named asm labels

asquared31415:
This adds a deny-by-default lint to prevent the use of named labels in inline `asm!`.  Without a solution to #81088 about whether the compiler should rewrite named labels or a special syntax for labels, a lint against them should prevent users from writing assembly that could break for internal compiler reasons, such as inlining or anything else that could change the number of actual inline assembly blocks emitted.

This does **not** resolve the issue with rewriting labels, that still needs a decision if the compiler should do any more work to try to make them work.


asquared31415:
I decided to initially make this a deny-by-default lint because the use of named labels is almost always wrong.  The RFC even says that only GAS style labels should be used.  However I think it could be important to be able to be disabled if a user really understands what that means.  This could be made warn by default (which I don't think is a good idea, named labels are objectively wrong) or a hard error though.

asquared31415:
Whoops, had some git issues and seemingly didn't rebase that file right when trying to fix things

compiler/rustc_builtin_macros/src/asm.rs
```diff
@@ -469,6 +473,70 @@ fn expand_preparsed_asm(ecx: &mut ExtCtxt<'_>, args: AsmArgs) -> Option<ast::Inl
...
+                None
+            };
+
+            let mut found_labels = Vec::new();
+
+            // A semicolon might not actually be specified as a separator for all targets, but it seems like LLVM accepts it always
+            let statements = template_str.split(|c| matches!(c, '\n' | ';'));
+            for statement in statements {
+                let mut start_idx = 0;
+                for (idx, _) in statement.match_indices(':') {
```
bjorn3, line 498: Can this exclude comments? LLVM handles `//` as comment start for all targets.

compiler/rustc_builtin_macros/src/asm.rs
```diff
@@ -469,6 +473,70 @@ fn expand_preparsed_asm(ecx: &mut ExtCtxt<'_>, args: AsmArgs) -> Option<ast::Inl
...
+                    if let Some(c) = chars.next() {
+                        // A label starts with an alphabetic character and continues with alphanumeric characters
+                        if c.is_alphabetic() {
+                            if chars.all(char::is_alphanumeric) {
+                                found_labels.push(possible_label);
+                            }
+                        }
+                    }
+
+                    start_idx = idx + 1;
```
Amanieu, line 510: Be careful about potential false positives: there may be valid uses of `:` in assembly code outside of labels. For example, AArch64 uses `:lo12:symbol_name` to refer to the low 12 bits of a symbol address.

compiler/rustc_builtin_macros/src/asm.rs
```diff
@@ -469,6 +473,70 @@ fn expand_preparsed_asm(ecx: &mut ExtCtxt<'_>, args: AsmArgs) -> Option<ast::Inl
...
+                let spans =
+                    found_labels.into_iter().filter_map(find_label_span).collect::<Vec<Span>>();
+                if spans.len() > 0 {
+                    for span in spans.into_iter() {
+                        ecx.parse_sess().buffer_lint_with_diagnostic(
+                            lint::builtin::NAMED_ASM_LABELS,
+                            span,
+                            ecx.current_expansion.lint_node_id,
+                            "do not use named labels in inline assembly",
+                            BuiltinLintDiagnostics::NamedAsmLabel("Only GAS local labels of the form `N:` where N is a number may be used in inline asm".to_string()),
```
Amanieu, line 524: I would word this slightly differently:
```suggestion
                            "avoid using named labels in inline assembly",
                            BuiltinLintDiagnostics::NamedAsmLabel("Only GAS local labels of the form `N:` where N is a number should be used in inline asm".to_string()),
```

compiler/rustc_lint_defs/src/lib.rs
```diff
@@ -274,15 +274,15 @@ impl<HCX> ToStableHashKey<HCX> for LintId {
...
-#[derive(PartialEq, Debug)]
+#[derive(PartialEq)]
 pub enum ExternDepSpec {
     Json(Json),
     Raw(String),
 }
 
 // This could be a closure, but then implementing derive trait
 // becomes hacky (and it gets allocated).
-#[derive(PartialEq, Debug)]
```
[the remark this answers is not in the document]
asquared31415: I had some git issues and I guess I missed this when trying to fix them :sweat_smile: I definitely didn't mean to do this.

compiler/rustc_builtin_macros/src/asm.rs
asquared31415: Thanks for reminding me about comments!

asquared31415:
I think it's right that that fails, technically, but that makes that test hard to do.  What should I do, allow the lint there?

Amanieu:
Yes, just allow the lint there.

compiler/rustc_builtin_macros/src/asm.rs
Amanieu: I'm not sure if you missed my comment or if you prefer sticking to the existing wording.

asquared31415: I've incorporated this and also made the help a little more concise

Amanieu:
(sorry about the late review, I *thought* that I submitted it last week but apparently something went wrong and it just got buffered)

compiler/rustc_builtin_macros/src/asm.rs
```diff
@@ -469,6 +473,79 @@ fn expand_preparsed_asm(ecx: &mut ExtCtxt<'_>, args: AsmArgs) -> Option<ast::Inl
...
+            }
+
+            if found_labels.len() > 0 {
+                let spans =
+                    found_labels.into_iter().filter_map(find_label_span).collect::<Vec<Span>>();
+                if spans.len() > 0 {
+                    for span in spans.into_iter() {
+                        ecx.parse_sess().buffer_lint_with_diagnostic(
+                            lint::builtin::NAMED_ASM_LABELS,
+                            span,
```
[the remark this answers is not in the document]
asquared31415: Oh neat I didn't realize that was what `MultiSpan` meant, though it seems pretty obvious in hindsight and I should have checked the impls for it.  That looks a lot cleaner.

npmccallum:
@Amanieu @bjorn3 While I definitely think we want a lint similar to this, I wish it had gone through some broader review before merging. We awoke this morning to find our CI/CD broken because of this lint erroring on local symbolic labels (i.e. `.Lfoo`).

https://docs.oracle.com/cd/E19120-01/open.solaris/817-5477/esqaq/index.html
https://sourceware.org/binutils/docs-2.25/as/Symbol-Names.html

The motivation for this patch is to prevent emitting conflicting entries to the object file's symbol table. Numeric labels don't cause this problem and are therefore allowed. But this patch additionally disables the use of local symbolic labels which also don't cause this problem and these should be allowed too.

This lint breaks the compilation of existing, working and valid code. I propose that this patch be reverted until it allows local symbolic labels.

cc @haraldh @joshtriplett

Amanieu:
Actually local labels (`.Lfoo`) can still cause conflicts if there are multiple definitions within the same object file/compilation unit. This can happen if a function containing inline assembly is inlined or duplicated within the same compilation unit. So the lint is correct as it is right now.

Amanieu:
Example: https://play.rust-lang.org/?version=nightly&mode=debug&edition=2018&gist=af54e969951198a25e406c42cf37fc8a

npmccallum:
@Amanieu This still seems to me like hammering a nail with a sledge hammer.

1. Before this lint, the compiler produced a completely readable error message regarding the unsupported configuration. Therefore, this lint is not an improvement of an existing defect.

```
error: invalid symbol redefinition
  --> src/lib.rs:11:15
   |
11 |         asm!(".Lfoo: nop");
   |               ^
   |
note: instantiated into assembly here
  --> <inline asm>:2:2
   |
2  |     .Lfoo: nop
   |     ^

```

2. If we want this lint (I'm not sure we do), it should only be active in the cases where this collision can occur (i.e. during inlining).

3. It should definitely not be active on naked functions where the pain is most acute. Naked functions will never be inlined or duplicated. Naked functions also benefit the most from local named labels.

Amanieu:
1. Do you have any suggestions on how the lint message can be improved?
2. Duplication can happen *anywhere*. Even if a function is marked `#[inline(never)]`, compiler optimizations can still duplicate parts of the function body, including `asm!` blocks.
3. This is a valid point, perhaps we can disable this lint for naked functions. This seems a bit difficult with the current implementation since it emits the lint during macro expansion. Could `#[naked]` implicitly add `#[allow(named_asm_label)]`?

npmccallum:
@Amanieu 

> 1. Do you have any suggestions on how the lint message can be improved?

The lint should specify that local named labels are unsupported outside naked functions.

> 2. Duplication can happen _anywhere_. Even if a function is marked `#[inline(never)]`, compiler optimizations can still duplicate parts of the function body, including `asm!` blocks.

I concede this point.

> 3. This is a valid point, perhaps we can disable this lint for naked functions. This seems a bit difficult with the current implementation since it emits the lint during macro expansion. Could `#[naked]` implicitly add `#[allow(named_asm_label)]`?

I'd have to think through the implications of that. Would it mean that developers couldn't change this to an error for naked functions?

However, this PR still breaks existing, valid code and should be reverted until the implications of this change are properly understood and the semantics are agreed upon.

bjorn3:
> However, this PR still breaks existing, valid code

Apart from naked functions, it has never been valid to use named labels, even before this PR. A future LLVM version or other backend may cause this to break at any point. The point of this PR is exactly to prevent people from writing seemingly valid, but not actually valid inline assembly.

I agree that it makes sense to disable this lint for naked functions.

> and should be reverted until the implications of this change are properly understood and the semantics are agreed upon.

The inline asm RFC has already defined it as not allowed:

https://rust-lang.github.io/rfcs/2873-inline-asm.html

> * You cannot assume that an asm! block will appear exactly once in the output binary. The compiler is allowed to instantiate multiple copies of the asm! block, for example when the function containing it is inlined in multiple places.
>     * As a consequence, you should only use local labels inside inline assembly code. Defining symbols in assembly code may lead to assembler and/or linker errors due to duplicate symbol definitions.

Apart from naked functions I don't think there are any edge cases where it would be fine to allow named asm labels. If there are, an issue can be opened and a solution can be discussed. As `asm!()` is still nightly-only, it is allowed to break it. That doesn't mean we should randomly do so, but it is acceptable if there is an alternative that can be used. In this case that is using numbered labels. There have been much bigger breaking changes to nightly-only features in the past, including renaming `asm!()` to `llvm_asm!()` to make place for the much better `asm!()` we currently have.

npmccallum:
@bjorn3 Which still means that this PR breaks existing, valid code in the case of naked functions. That is, a change in one feature has broken valid code in another feature. The implementor of the (now broken) feature wasn't consulted and is being asked to scramble to fix the breakage or live with it. This isn't contributor friendly.

It should be the responsibility of the PR author to fix breakage in other features caused by the PR. Ideally, this should be caught before merge. But when it isn't, how should we handle this? I would argue that it is still the responsibility of the PR author to fix the breakage. And the only way, procedurally, that we can enforce this is to revert the PR and reopen it.

asquared31415:
I agree with bjorn on all 3 points.  Rust makes no stability guarantees about unstable features, and it is allowed and there is precedent to entirely change unstable features.

I was not aware that `#[naked]` functions prevent code duplication or inlining, and because of that implemented this lint unconditionally to match the the RFC.  If it is true that `#[naked]` functions are exempt from the potential issues, they should be granted an **exception** in this lint, and not be the rule.

npmccallum:
> I agree with bjorn on all 3 points. Rust makes no stability guarantees about unstable features, and it is allowed and there is precedent to entirely change unstable features. If someone must rely on a nightly feature never changing, they should pin the nightly version to a specific, known working version, and update with caution, until the feature is stabilized.

Agreed completely.

> I was not aware that `#[naked]` functions prevent code duplication or inlining, and because of that implemented this lint unconditionally to match the the RFC. If it is true that `#[naked]` functions are exempt from the potential issues, they should be granted an **exception** in this lint, and not be the rule.

I think the main question is who does this work and when. Reverting the PR gives us breathing room to figure this out and coordinate across features.

asquared31415:
`#[naked]` applies the same flag as `#[inline(never)]` for LLVM.  So unless there's additional undocumented behavior to `naked` in LLVM, it is still possible for naked functions to be inlined.  And even if there is that undocumented behavior, how much should we rely on that undocumented behavior.

asquared31415:
If it does turn out that we *can* rely on this behavior of `#[naked]` I will fix the lint to add an exception

npmccallum:
Naked functions **MUST NOT** ever be inlined. If there is some undefined behavior that causes one to be inlined, it is a bug that should be fixed.

When I disabled inlining for naked functions, I did not implement it as desugaring to `#[inline(never)]`. In fact, we lint so that mentioning inlining at all on a naked function is disallowed.

asquared31415:
Then this is a shortcoming of LLVM, it doesn't seem to have a way to specify "no never inline this ever, if you do this is wrong".  The best is `noinline` which as stated above seems to be a very strong hint, and is currently what we pass to LLVM for naked functions, and is the same attribute used for `#[inline(never)]`

bjorn3:
I think naked functions can still be inlined if and only if you ensure that you setup all registers and the stack anyway as if you would actually do a real call. This is a silly thing to do, but indistinguishable from a real call except that it would duplicate the inline asm and possibly avoid a single call instruction in favor of falling through to the naked function.

Amanieu:
At this point I'm fairly confident that we can guarantee that `asm!` in a naked function is never duplicated, so we should just go ahead and disable this lint for `asm!` in naked functions. If LLVM's behavior changes then we can always switch to lowering naked functions to `global_asm!` instead.

npmccallum:
@bjorn3 There are possibly other invariants as well (depending on platform). But while it is theoretically possible to inline a naked function, the complexity of this problem is too difficult and with too little gain. Therefore, in the Constrained Naked RFC, we commit the compiler to never inlining a naked function. This may be a promise the compiler cannot (currently) keep. But that might change in the future. And in the meantime, we clearly communicate that no one should ever depend on this behavior and can reliably depend on the opposite behavior.

asquared31415:
If we are advising users to not depend on naked functions being never inlined, shouldn't this lint still activate then, since the compiler cannot currently promise that?

npmccallum:
@asquared31415 Per the [Constrained Naked Functions RFC](https://github.com/npmccallum/rust-rfcs/blob/naked/text/2972-constrained-naked.md#naked-function-definition) we are advising users to depend on naked functions never being inlined.

asquared31415:
Oh I misinterpreted your previous statement then.  I thought `this behavior` meant that they aren't inlined.

I still think it is possible that LLVM changes in the future to actually inline things that are `noinline` if it really has a good reason to, but since it currently doesn't inline these functions, we could make this exception and handle that scenario if it happens.

npmccallum:
@asquared31415 I'm sure that if LLVM adds functionality to inline even those functions, the next thing people will ask for is a way to disable it.

haraldh:
Also the linter is triggered on comments:

```rust
#[naked]
pub unsafe extern "sysv64" fn _start() -> ! {
    asm!("
    nop          # here is a comment about syscall
    nop          // and another syscall comment
    jmp syscall
    hlt
.global syscall
syscall:
    hlt
    ",
            options(noreturn)
    );
}
```

gives:
```
error: avoid using named labels in inline assembly
  --> src/lib.rs:7:44
   |
7  |       nop          # here is a comment about syscall
   |  ____________________________________________^
8  | |     nop          // and another syscall comment
9  | |     jmp syscall
10 | |     hlt
11 | | .global syscall
12 | | syscall:
   | |_______^
   |
   = note: `#[deny(named_asm_labels)]` on by default
   = help: only local labels of the form `<number>:` should be used in inline asm
   = note: see the asm section of the unstable book <https://doc.rust-lang.org/nightly/unstable-book/library-features/asm.html#labels> for more information
```

[Playground}(https://play.rust-lang.org/?version=nightly&mode=debug&edition=2018&gist=8063d9a38abf8f8847569d895a7592aa)

bjorn3:
The lint triggers on the label definition (`syscall:`) and then tries to find uses of it using a heuristic. This heuristic skips `//` comments, but not platform dependent comments. It does not _trigger_ on comments except if the comment contains a `;`. It seems that `;` is used as statement separator before `//` comments are stripped. It should have been the opposite way around. First stripping comments and then splitting at `;`.

asquared31415:
@haraldh please file a separate issue about that, it's a lot easier to track the issues that way

I was not aware that platform specific comments existed, I tried to make sure that I wasn't missing platform specific behavior, but must have missed that anyway.

PR #65609, opened 2019-10-19
Implement PartialOrd<&B> for &mut A

Duddino:
Fixes #65589

PR #48597, opened 2018-02-28
Add Iterator::collect_into

CodeSandwich:
This is PR for adding `Iterator::extend` and `Iterator::extend_mut` as discussed in [RFC issue 2339](https://github.com/rust-lang/rfcs/issues/2339). It was advised to bypass RFC and just implement it.

PR #87328, opened 2021-07-21
Implement `PartialOrd<&B>` for `&mut A`

### LINKED ISSUE #65589: You can't compare &mut i32 with &i32, but you can compare &i32 with &mut i32
https://play.rust-lang.org/?version=nightly&mode=debug&edition=2018&gist=11d3a2fe6f9ec8544d3f64917b99f913

inquisitivecrystal:
Due to an apparent oversight, the impl `PartialOrd<&B> for &mut A` did not previously exist. This leads to some very weird asymmetry. To quote #65589, "you can't compare &mut i32 with &i32, but you can compare &i32 with &mut i32". This PR adds the missing impl.

This is essentially copied from Duddino's work in #65609.

Fixes #65589.

### DIFF
--- a/library/core/src/cmp.rs
+++ b/library/core/src/cmp.rs
@@ -1478,6 +1478,32 @@ mod impls {
             PartialOrd::ge(*self, *other)
         }
     }
+    #[stable(feature = "partial_ord_mut_ref", since = "1.55.0")]
+    impl<A: ?Sized, B: ?Sized> PartialOrd<&B> for &mut A
+    where
+        A: PartialOrd<B>,
+    {
+        #[inline]
+        fn partial_cmp(&self, other: &&B) -> Option<Ordering> {
+            PartialOrd::partial_cmp(*self, *other)
+        }
+        #[inline]
+        fn lt(&self, other: &&B) -> bool {
+            PartialOrd::lt(*self, *other)
+        }
+        #[inline]
+        fn le(&self, other: &&B) -> bool {
+            PartialOrd::le(*self, *other)
+        }
+        #[inline]
+        fn ge(&self, other: &&B) -> bool {
+            PartialOrd::ge(*self, *other)
+        }
+        #[inline]
+        fn gt(&self, other: &&B) -> bool {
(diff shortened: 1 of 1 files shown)


scottmcm:
> you can't compare &mut i32 with &i32, but you can compare &i32 with &mut i32

Hmm, but there's no `impl PartialOrd<&mut B> for &A` either.

If this were going to be added, then I'd expect both impls to be added.  But I'm not sure it's legal to add this, because references are fundamental.  (See how `impl Extend for &mut E` couldn't be added https://github.com/rust-lang/rust/pull/48597#discussion_r172677521.)

But this is an insta-stable impl, so I'll delegate this over to libs-api
r? libs

inquisitivecrystal:
> Hmm, but there's no `impl PartialOrd<&mut B> for &A` either.

I tried to add it, but it wouldn't build. I presumed, without much thinking about it, that that was broken in some way this wasn't. Silly of me, I guess. ;)

These continual trait problems are *really* frustrating from a stability without stagnation perspective. Not like I know how to fix them though.

scottmcm:
I guess one way works because of coercions but the other doesn't.

Hypothesis: if the LHS is a `&` and the RHS a `&mut`, the solver picks the `&` impl, and then the RHS reborrows.  But if the LHS is a `&mut` then it can't fix the RHS.

(AKA the classic "inference and coercions don't mix" kind of problem.)

inquisitivecrystal:
> I guess one way works because of coercions but the other doesn't.
> 
> Hypothesis: if the LHS is a `&` and the RHS a `&mut`, the solver picks the `&` impl, and then the RHS reborrows. But if the LHS is a `&mut` then it can't fix the RHS.
> 
> (AKA the classic "inference and coercions don't mix" kind of problem.)


Hmmm. If that's the problem, maybe the real solution is to tweak the coercion rules somehow. I don't understand them nearly well enough to figure out whether that's viable though.

inquisitivecrystal:
I'm going to close this. I think we need to do something about this, but this PR is probably too backwards incompatible to actually be worth merging.

Issue #87074, opened 2021-07-12
Tracking Issue for RFC 3128: I/O Safety


joshtriplett:
Feature gate: `#![feature(io_safety)]`

This is a tracking issue for [RFC 3128: I/O Safety](https://github.com/rust-lang/rfcs/pull/3128).

Raw OS handles such as `RawFd` and `RawHandle` have hazards similar to raw pointers; they may be bogus or may dangle, leading to broken encapsulation boundaries and code whose behavior is impossible to bound in general.

Introduce a concept of *I/O safety*, and introduce a new set of types and traits, led by `OwnedFd` and `BorrowedFd`, to support it.

### Public API

The public API on UNIX platforms consists of the types `OwnedFd` and `BorrowedFd`, the trait `AsFd`, and implementations of `AsFd`, `Into<OwnedFd>`, and `From<OwnedFd>` for various types (such as files and sockets).

The public API on Windows platforms consists of two sets of parallel types and traits and impls for `OwnedHandle`, `OwnedSocket`, `BorrowedHandle`, `BorrowedSocket`, etc.

### Steps / History

- [x] Implementation (based on @sunfishcode's existing work outside of std)
- [ ] Final comment period (FCP)
  - Because these include a new stable layout guarantee (#96947), this should include T-lang (https://rust-lang.zulipchat.com/#narrow/stream/213817-t-lang/topic/rustc_nonnull_optimization_guaranteed.20oversight.3F)
- [ ] Stabilization PR

### Unresolved Questions

- [ ] This RFC doesn't define a formal model for raw handle ownership and lifetimes. The rules for raw handles in this RFC are vague about their identity. What does it mean for a resource lifetime to be associated with a handle if the handle is just an integer type? Do all integer types with the same value share that association?
- [ ] The Rust [reference](https://doc.rust-lang.org/reference/behavior-considered-undefined.html) defines undefined behavior for memory in terms of [LLVM's pointer aliasing rules](http://llvm.org/docs/LangRef.html#pointer-aliasing-rules); I/O could conceivably need a similar concept of handle aliasing rules. This doesn't seem necessary for present practical needs, but it could be explored in the future.

sunfishcode:
I've now posted an initial PR implementing the proposal in the RFC: https://github.com/rust-lang/rust/pull/87329

sunfishcode:
The implementation PR is now [merged](https://github.com/rust-lang/rust/pull/87329)!

sunfishcode:
[...]

PR #81825, opened 2021-02-06
Add Linux-specific pidfd process extensions (take 2)

voidc:
Continuation of #77168.
I addressed the following concerns from the original PR:

- make `CommandExt` and `ChildExt` sealed traits
- wrap file descriptors in `PidFd` struct representing ownership over the fd
- add `take_pidfd` to take the fd out of `Child`
- close fd when dropped

Tracking Issue: #82971

PR #87329, opened 2021-07-21
I/O safety.

sunfishcode:
Introduce `OwnedFd` and `BorrowedFd`, and the `AsFd` trait, and
implementations of `AsFd`, `From<OwnedFd>` and `From<T> for OwnedFd`
for relevant types, along with Windows counterparts for handles and
sockets.

Tracking issue: <https://github.com/rust-lang/rust/issues/87074>

RFC: <https://github.com/rust-lang/rfcs/blob/master/text/3128-io-safety.md>

Highlights:
 - The doc comments at the top of library/std/src/os/unix/io/mod.rs and library/std/src/os/windows/io/mod.rs
 - The new types and traits in library/std/src/os/unix/io/fd.rs and library/std/src/os/windows/io/handle.rs
 - The removal of the `RawHandle` struct the Windows impl, which had the same name as the `RawHandle` type alias, and its functionality is now folded into `Handle`.

Managing five levels of wrapping (File wraps sys::fs::File wraps sys::fs::FileDesc wraps OwnedFd wraps RawFd, etc.) made for a fair amount of churn and verbose as/into/from sequences in some places. I've managed to simplify some of them, but I'm open to ideas here.

r? @joshtriplett


joshtriplett:
> Managing five levels of wrapping (File wraps sys::fs::File wraps sys::fs::FileDesc wraps OwnedFd wraps RawFd, etc.) made for a fair amount of churn and verbose as/into/from sequences in some places. I've managed to simplify some of them, but I'm open to ideas here.

I don't know how feasible this would be, but I wonder if we could eliminate `sys::fs::FileDesc` and have `sys::fs::File` directly use `OwnedFd`. That would give us what feels like the *right* amount of indirection: the portable `File` wraps the UNIX-specific `sys::fs::File`, which wraps an `OwnedFd`, which wraps the simple type alias `RawFd`.

library/std/src/os/unix/io/fd.rs
```diff
@@ -0,0 +1,329 @@
...
+}
+
+#[unstable(feature = "io_safety", issue = "87074")]
+impl FromRawFd for OwnedFd {
+    /// Constructs a new instance of `Self` from the given raw file descriptor.
+    ///
+    /// # Safety
+    ///
+    /// The resource pointed to by `raw` must be open and suitable for assuming
+    /// ownership.
```
joshtriplett, line 102:
```suggestion
    /// ownership. The resource must not require any cleanup other than `close`.
```

library/std/src/os/unix/io/fd.rs
```diff
@@ -0,0 +1,329 @@
...
+/// has the value `-1`.
+#[repr(transparent)]
+#[rustc_layout_scalar_valid_range_start(0)]
+// libstd/os/raw/mod.rs assures me that every libstd-supported platform has a
+// 32-bit c_int. Below is -2, in two's complement, but that only works out
+// because c_int is 32 bits.
+#[rustc_layout_scalar_valid_range_end(0xFF_FF_FF_FE)]
+#[unstable(feature = "io_safety", issue = "87074")]
+pub struct OwnedFd {
+    raw: RawFd,
```
joshtriplett, line 51: The struct uses the field name `raw`, but the debug representation uses the field name `fd`. Could we change the field name to be `fd`, for simplicity?

(We could also consider using a one-tuple struct, `OwnedFd(RawFd)`, but we unfortunately can't do that for `BorrowedFd`, and I think the asymmetry isn't worth it.)

library/std/src/os/unix/io/fd.rs
```diff
@@ -0,0 +1,329 @@
...
+impl fmt::Debug for OwnedFd {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        f.debug_struct("OwnedFd").field("fd", &self.raw).finish()
+    }
+}
+
+/// A trait to borrow the file descriptor from an underlying object.
+///
+/// This is only available on unix platforms and must be imported in order to
+/// call the method. Windows platforms have a corresponding `AsHandle` and
```
joshtriplett, line 143: Please add `AsFd`, `OwnedFd`, and `BorrowedFd` to [`std::os::unix::prelude`](https://doc.rust-lang.org/std/os/unix/prelude/index.html). (And please do the same for the corresponding Windows types and traits in the corresponding Windows prelude.)

library/std/src/os/unix/io/fd.rs
```diff
@@ -0,0 +1,329 @@
...
+    fn as_fd(&self) -> BorrowedFd<'_> {
+        *self
+    }
+}
+
+#[unstable(feature = "io_safety", issue = "87074")]
+impl AsFd for OwnedFd {
+    #[inline]
+    fn as_fd(&self) -> BorrowedFd<'_> {
+        unsafe { BorrowedFd::borrow_raw_fd(self.as_raw_fd()) }
```
joshtriplett, line 177: Shouldn't this have a `// SAFETY` comment, observing that `OwnedFd` and `BorrowedFd` have the same validity invariants?

library/std/src/os/unix/io/fd.rs
```diff
@@ -0,0 +1,329 @@
...
+impl FromRawFd for OwnedFd {
+    /// Constructs a new instance of `Self` from the given raw file descriptor.
+    ///
+    /// # Safety
+    ///
+    /// The resource pointed to by `raw` must be open and suitable for assuming
+    /// ownership.
+    #[inline]
+    unsafe fn from_raw_fd(raw: RawFd) -> Self {
+        assert_ne!(raw, -1i32);
```
joshtriplett, line 105: I deeply appreciate this assertion. This will catch any attempt to construct an OwnedFd (or a BorrowedFd given the assertion there too) from a function returning a `RawFd` without first checking for errors.

library/std/src/os/unix/io/fd.rs
```diff
@@ -0,0 +1,329 @@
...
+#[unstable(feature = "io_safety", issue = "87074")]
+impl AsFd for OwnedFd {
+    #[inline]
+    fn as_fd(&self) -> BorrowedFd<'_> {
+        unsafe { BorrowedFd::borrow_raw_fd(self.as_raw_fd()) }
+    }
+}
+
+#[unstable(feature = "io_safety", issue = "87074")]
+impl AsFd for fs::File {
```
joshtriplett, line 182: I feel like all of these impls should be near the types they're impl'd for, rather than all in this file. Putting them all in this file makes it harder to notice that they need adding when you add a new type. (For instance, if you add a new socket type or similar, you may start from the implementation of an existing socket type.)

library/std/src/os/wasi/io/fd.rs
```diff
@@ -0,0 +1,280 @@
+//! Owned and borrowed file descriptors.
```
joshtriplett, line 1: There seem to be two copies of some of the core types and traits, one for WASI and one for UNIX. I'd like to only have *one* copy of the types and traits, and only have bits in WASI and UNIX that are unique to those target families.

If need be, we could have an `io_safety_common` or something. But ideally, couldn't this all live under UNIX?

library/std/src/os/windows/io/handle.rs
```diff
@@ -0,0 +1,390 @@
...
+    }
+}
+
+impl FromRawHandle for OwnedHandle {
+    /// Constructs a new instance of `Self` from the given raw handle.
+    ///
+    /// # Safety
+    ///
+    /// The resource pointed to by `raw` must be open and suitable for assuming
+    /// ownership.
```
joshtriplett, line 178:
```suggestion
    /// ownership. The resource must not require any cleanup other than `CloseHandle`.
```

library/std/src/os/windows/io/handle.rs
```diff
@@ -0,0 +1,390 @@
...
+/// Note that it *may* have the value `INVALID_HANDLE_VALUE`. See [here] for
+/// the full story. For APIs like `CreateFileW` which report errors with
+/// `INVALID_HANDLE_VALUE` instead of null, use [`OptionFileHandle`] instead
+/// of `Option<OwnedHandle>`.
+///
+/// [here]: https://devblogs.microsoft.com/oldnewthing/20040302-00/?p=40443
+#[repr(transparent)]
+#[unstable(feature = "io_safety", issue = "87074")]
+pub struct OwnedHandle {
+    raw: NonNull<c_void>,
```
joshtriplett, line 54: The Debug impl uses `handle`, but the field name uses `raw`; could we consistently use `handle`?

library/std/src/os/unix/io/fd.rs
```diff
@@ -0,0 +1,329 @@
...
+/// This uses `repr(transparent)` and has the representation of a host file
+/// descriptor, so it can be used in FFI in places where a file descriptor is
+/// passed as an argument, it is not captured or consumed, and it never has the
+/// value `-1`.
+#[derive(Copy, Clone)]
+#[repr(transparent)]
+#[rustc_layout_scalar_valid_range_start(0)]
+// libstd/os/raw/mod.rs assures me that every libstd-supported platform has a
+// 32-bit c_int. Below is -2, in two's complement, but that only works out
+// because c_int is 32 bits.
```
joshtriplett, line 27: Could you add a test, somewhere, that does an `assert_eq!(size_of::<RawFd>(), 4);`, to catch this if someone ever ports to a platform where `RawFd` uses some other size?

library/std/src/os/windows/io/socket.rs
```diff
@@ -0,0 +1,212 @@
...
+    }
+}
+
+impl FromRawSocket for OwnedSocket {
+    /// Constructs a new instance of `Self` from the given raw socket.
+    ///
+    /// # Safety
+    ///
+    /// The resource pointed to by `raw` must be open and suitable for assuming
+    /// ownership.
```
joshtriplett, line 101:
```suggestion
    /// ownership. The resource must not require cleanup other than `closesocket`.
```

library/std/src/os/windows/io/socket.rs
```diff
@@ -0,0 +1,212 @@
...
+#[rustc_layout_scalar_valid_range_start(0)]
+// This is -2, in two's complement. -1 is `INVALID_SOCKET`.
+#[cfg_attr(target_pointer_width = "32", rustc_layout_scalar_valid_range_end(0xFF_FF_FF_FE))]
+#[cfg_attr(
+    target_pointer_width = "64",
+    rustc_layout_scalar_valid_range_end(0xFF_FF_FF_FF_FF_FF_FF_FE)
+)]
+#[unstable(feature = "io_safety", issue = "87074")]
+pub struct OwnedSocket {
+    raw: RawSocket,
```
joshtriplett, line 53: The `Debug` impls use `socket`, but the field names use `raw`; could we consistently use `socket`?

library/std/src/os/windows/io/socket.rs
```diff
@@ -0,0 +1,212 @@
...
+}
+
+impl AsSocket for OwnedSocket {
+    #[inline]
+    fn as_socket(&self) -> BorrowedSocket<'_> {
+        unsafe { BorrowedSocket::borrow_raw_socket(self.as_raw_socket()) }
+    }
+}
+
+impl AsSocket for crate::net::TcpStream {
```
joshtriplett, line 151: As with the UNIX versions, can we put these impls near the corresponding types?

library/std/src/os/unix/io/fd.rs
joshtriplett: Note: this *might* not be possible, since portable files aren't supposed to have `cfg`-based non-portable impls in them. If this turns out to not be possible, perhaps a comment near the corresponding types would be helpful.

Thomasdezeeuw (commented):
Looks really good @sunfishcode, just some small comments.

library/std/src/os/windows/io/handle.rs
```diff
@@ -0,0 +1,390 @@
...
+/// If this holds an owned handle, it closes the handle on drop.
+///
+/// This uses `repr(transparent)` and has the representation of a host handle,
+/// so it can be used in FFI in places where a non-null handle is passed as a
+/// consumed argument or returned as an owned value, or it is
+/// `INVALID_HANDLE_VALUE` indicating an error or an otherwise absent value.
+///
+#[repr(transparent)]
+#[unstable(feature = "io_safety", issue = "87074")]
+pub struct OptionFileHandle {
```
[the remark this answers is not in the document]
sunfishcode: I guess this wasn't explicitly mentioned, but it is effectively part of the Windows counterpart to `OwnedFd`, which we said we wanted. Because Windows has this surprising behavior where some APIs use `INVALID_HANDLE_VALUE` to indicate failure and other APIs use `INVALID_HANDLE_VALUE` as a valid handle value, it's useful to have two different types to be able to describe the two different kinds of APIs.

ChrisDenton:
In case it's relevant here, Windows essentially has two types of I/O `HANDLE` (it also reuses the type for other things but that's off topic):

* Kernel handles. These are "real" handles (i.e. they are an offset into a table of kernel objects).
* Pseudo handles. These are "fake" handles (i.e. managed by the Win32 subsystem). They have resources managed outside of the NT kernel so are often intentionally invalid kernel handles. The meaning of pseudo handles is very context dependent.

In both cases, zero is not a valid handle. It's also guaranteed that[ a kernel handle will not have the lowest two bits set](https://devblogs.microsoft.com/oldnewthing/20080827-00/?p=21073). A pseudo handle has no such restriction and indeed may intentionally make use of the lowest two bits to distinguish themselves from "real" handles. Therefore `INVALID_HANDLE_VALUE` is not a valid kernel handle but can be reused as a pseudo handle.

(I'm also pretty sure a kernel handle will always be less than `i32::MAX` but I'm not sure to what extent that's guaranteed).

I'm uncertain if this information is helpful to anyone but I just wanted to document some of what I know.

sunfishcode:
Thanks @ChrisDenton! Do you by chance know of an authoritative reference stating that zero is not a valid handle? The code here assumes that, and for various reasons it seems like a safe assumption, but if there's official documentation about this, I'd like to cite it.

ChrisDenton:
Unfortunately I do not. It shakes out that way, as you note, for various practical reasons but I don't know that it's explicitly documented anywhere.

ChrisDenton:
@sunfishcode From [GetStdHandle](https://docs.microsoft.com/en-us/windows/console/getstdhandle#return-value):

> If an application does not have associated standard handles, such as a service running on an interactive desktop, and has not redirected them, the return value is NULL.

So here a NULL I/O handle means "no handle". You can redirect stdin etc to any I/O handle (where "I/O handle" means "can be passed to [`WriteFile`](https://docs.microsoft.com/en-us/windows/win32/api/fileapi/nf-fileapi-writefile), etc") therefore no valid I/O handle can be NULL because it would overlap with that meaning.

I admit this falls short of explicit documentation but it's the closest I can find. You could also cite functions such as [`OpenProcess`](https://docs.microsoft.com/en-us/windows/win32/api/processthreadsapi/nf-processthreadsapi-openprocess#return-value) that use NULL to indicate errors but obviously they're specific to the type of handle.

In short there's no explicit documentation that covers all types of handles but, as the blog you linked to notes, the very fact that it was a "toss up" whether to use NULL to signal an error is a good indication that it's not a valid handle, And unlike `INVALID_HANDLE_VALUE` there's no documentation that says "for this specific type of handle NULL is treated as a special, valid, handle value".

(more broadly, it would likely break a lot of things if NULL were ever to become valid but this is supposition on my part).

library/std/src/os/unix/io/mod.rs
```diff
@@ -0,0 +1,56 @@
...
+//! any system call except for:
+//!  - `close`, because that would end the dynamic lifetime of the resource
+//!    without ending the lifetime of the file descriptor.
+//!  - `dup2`/`dup3`, in the second argument, because this argument is
+//!    closed and replaced with a new resource, which may break the assumptions
+//!    other code using that file descriptor.
+//! This list doesn't include `mmap`, since `mmap` does do a proper borrow of
+//! its file descriptor argument. That said, `mmap` is unsafe for other
+//! reasons: it operates on raw pointers, and it has undefined behavior if the
+//! underlying storage is mutated. Mutations may come from other processes, or
```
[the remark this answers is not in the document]
Thomasdezeeuw: > If I change it from "it has undefined behavior" to "it can have undefined behavior", does that capture it?

Yes, bit more nuanced.

library/std/src/os/windows/io/handle.rs
```diff
@@ -0,0 +1,390 @@
...
+/// A borrowed handle.
+///
+/// This has a lifetime parameter to tie it to the lifetime of something that
+/// owns the handle.
+///
+/// This uses `repr(transparent)` and has the representation of a host handle,
+/// so it can be used in FFI in places where a handle is passed as an argument,
+/// it is not captured or consumed, and it is never null.
+///
+/// Note that it *may* have the value `INVALID_HANDLE_VALUE`. See [here] for
```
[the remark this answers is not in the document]
Thomasdezeeuw: Maybe saying something along the lines of "The handle must always be valid. Note that in some cases `INVALID_HANDLE_VALUE` is considered a valid value, see [blog post]. The creator of `BorrowedHandle` must ensure the handle is valid".

Because if I read it correctly it just happens that `INVALID_HANDLE_VALUE` overlaps with some valid handles? So when the handle is returned by the OS it should be checked by the caller with call-specific knowledge, i.e. whether or not `INVALID_HANDLE_VALUE` is considered valid.

library/std/src/os/windows/io/handle.rs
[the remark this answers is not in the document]
Thomasdezeeuw: Also see my answer above, but I think this is mostly a result from system call thing. With that I mean the caller needs to check if the handle is valid when it's created (i.e. after a system call), after that is can just use `OwnedHandle`.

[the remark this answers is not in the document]
sunfishcode: That's essentially how `OptionFileHandle` is meant to be used too. It's just for [use it in an FFI declaration](https://github.com/sunfishcode/io-lifetimes/blob/main/src/example_ffi.rs#L43) of functions that use `INVALID_FILE_HANDLE` for errors, and the main way to use it is to [immediately convert it to an `OwnedHandle`](https://github.com/sunfishcode/io-lifetimes/blob/main/examples/hello.rs#L79), which does the check for `INVALID_HANDLE_VALUE`. So the person writing the FFI declaration has use the right type, but then all call sites are prevented from doing the wrong check. Does that make sense? I'll add some comments explaining that in more detail.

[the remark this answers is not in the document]
Thomasdezeeuw: I don't these types should be used in ffi at all to be honest. Only once the caller (of the ffi function) checked all errors, validity, etc. should these types come in to play. Similar to how allocation functions use `*mut u8` and convert it into e.g. `Box<T>` after.

sunfishcode:
I'd like to raise the visibility of a concern raised [in comments above](https://github.com/rust-lang/rust/pull/87329#discussion_r675083974):

This PR includes `OptionFileHandle` on Windows, which wasn't explicitly named in the RFC. I included it because it's conceptually part of the Windows analog for `OwnedFd`, for using I/O-safe types in FFI signatures. The RFC mentions the FFI use case [here](https://github.com/rust-lang/rfcs/blob/master/text/3128-io-safety.md#ownedfd-and-borrowedfdfd-1), but not the [Windows-specific issues](https://devblogs.microsoft.com/oldnewthing/20040302-00/?p=40443) which motivate `OptionFileHandle`.

Does anyone else have opinions as to whether `OptionFileHandle` should be included in this PR?

library/std/src/os/unix/io/mod.rs
```diff
@@ -0,0 +1,56 @@
...
+//! encouraged to migrate, either by adding `unsafe` to APIs that dereference
+//! `RawFd` values, or by using to `BorrowedFd` or `OwnedFd` instead.
+//!
+//! Like references, `BorrowedFd` values are tied to a lifetime, to ensure
+//! that they don't outlive the resource they point to. These are safe to
+//! use. `BorrowedFd` values may be used in APIs which provide safe access to
+//! any system call except for:
+//!  - `close`, because that would end the dynamic lifetime of the resource
+//!    without ending the lifetime of the file descriptor.
+//!  - `dup2`/`dup3`, in the second argument, because this argument is
```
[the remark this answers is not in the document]
sunfishcode: I've now reworded this to say "`dup2`/`dup3`, in the second argument, because this argument is closed and assigned a new resource, which may break the assumptions other code using that file descriptor." Is that clear?

library/std/src/os/windows/io/handle.rs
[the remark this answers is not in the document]
sunfishcode: I summarized this concern in a top-level comment [here](https://github.com/rust-lang/rust/pull/87329#issuecomment-887688834) to give it more visibility.

library/std/src/os/unix/io/mod.rs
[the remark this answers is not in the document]
Thomasdezeeuw: LGTM.

library/std/src/os/unix/io/fd.rs
sunfishcode: As you guessed, this would require `cfg`-based impls in portable files. I've now added comments for these types.

library/std/src/os/windows/io/socket.rs
sunfishcode: As above, this would require `cfg`s in portable files.

library/std/src/os/wasi/io/fd.rs
sunfishcode: I've now factored this out into an `fd` module that lives in library/std/src/os. A plausible future for it could include factoring out the `AsRawFd`/`IntoRawFd`/`FromRawFd` code as well.

I avoided the word "common" since it it's still platform-specific to platforms that use file descriptors, but I'm not attached to this.

library/std/src/os/windows/io/handle.rs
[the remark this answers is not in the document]
sunfishcode: I've now added comments along these lines.

sunfishcode:
Another concern is whether it's appropriate to talk about migration and encouraging new code to adopt I/O safety in the documentation, as expressed [here](https://github.com/rust-lang/rust/pull/87329#discussion_r673729341). I believe we should document this somewhere, but I'm open to ideas of where the best place to do that is.

Other than those two concerns, I believe I've addressed all the review comments up to this point.

library/std/src/os/unix/io/mod.rs
```diff
@@ -0,0 +1,56 @@
...
+//! with different ownership properties: raw, borrowed, and owned, which are
+//! analogous to types used for representing pointers:
+//!
+//! | Type               | Analogous to |
+//! | ------------------ | ------------ |
+//! | [`RawFd`]          | `*const _`   |
+//! | [`BorrowedFd<'a>`] | `&'a _`      |
+//! | [`OwnedFd`]        | `Box<_>`     |
+//!
+//! Like raw pointers, `RawFd` values are primitive values. And in new code,
```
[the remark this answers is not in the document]
sunfishcode: I surfaced this as a top-level comment [here](https://github.com/rust-lang/rust/pull/87329#issuecomment-888371073).

joshtriplett:
@sunfishcode I think it's entirely appropriate to talk about migration. One caveat to that, though: we don't want to confusingly imply that people should migrate to this *while it's still nightly-only*. Otherwise the general tone about migration seems great.

m-ou-se:
As requested by Josh, I took a quick look through all the changes (skipping some details), and it all generally looks great. Thanks for doing this, @sunfishcode!

(Putting Josh back as the sole assignee for the the detailed review and handling the open comments. ^^)

library/std/src/os/windows/io/handle.rs
```diff
@@ -0,0 +1,427 @@
...
+/// The main thing you can do with an `OptionFileHandle` is to convert it into
+/// an `OwnedHandle` using its [`TryFrom`] implementation, and this conversion
+/// takes care of the check for `INVALID_HANDLE_VALUE`.
+///
+/// If this holds an owned handle, it closes the handle on drop.
+///
+/// This uses `repr(transparent)` and has the representation of a host handle,
+/// so it can be used in FFI in places where a non-null handle is passed as a
+/// consumed argument or returned as an owned value, or it is
+/// `INVALID_HANDLE_VALUE` indicating an error or an otherwise absent value.
```
joshtriplett, line 77: /// FFI type for handles in return values or out parameters, where `INVALID_HANDLE_VALUE` is used
/// as a sentry value to indicate errors, such as in the return value of `CreateFileW`. This uses
/// `repr(transparent)` and has the representation of a host handle, so that it can be used in such
/// FFI declarations.
///
/// The only thing you can usefully do with a `HandleOrInvalid` is to convert it into an
/// `OwnedHandle` using its [`TryFrom`] implementation; this conversion takes care of the check for
/// `INVALID_HANDLE_VALUE`. This ensures that such FFI calls cannot start using the handle without
/// checking for `INVALID_HANDLE_VALUE` first.
///
/// If this holds a valid handle, it will close the handle on drop.

library/std/src/os/windows/io/handle.rs
```diff
@@ -0,0 +1,427 @@
...
+#[repr(transparent)]
+#[unstable(feature = "io_safety", issue = "87074")]
+pub struct OptionFileHandle {
```
joshtriplett, line 80:
```suggestion
pub struct HandleOrInvalid(Option<OwnedHandle>);
```

library/std/src/os/windows/io/handle.rs
```diff
@@ -0,0 +1,427 @@
...
+    handle: RawHandle,
+}
```
joshtriplett, line 82:
```suggestion
pub struct HandleOrInvalid(Option<OwnedHandle>);
```

library/std/src/os/windows/io/handle.rs
```diff
@@ -0,0 +1,427 @@
...
+
+impl OptionFileHandle {
+    /// Return an empty `OptionFileHandle` with no resource.
+    #[inline]
+    #[unstable(feature = "io_safety", issue = "87074")]
+    pub const fn none() -> Self {
+        Self { handle: c::INVALID_HANDLE_VALUE }
+    }
+}
```
joshtriplett, line 124: I think this should be dropped, in favor of people calling the TryFrom implementation.
```suggestion
```

library/std/src/os/windows/io/handle.rs
```diff
@@ -0,0 +1,427 @@
...
+
+impl TryFrom<OptionFileHandle> for OwnedHandle {
```
joshtriplett, line 126:
```suggestion
impl TryFrom<HandleOrInvalid> for OwnedHandle {
```

library/std/src/os/windows/io/handle.rs
```diff
@@ -0,0 +1,427 @@
...
+            // APIs like [`CreateFileW`] itself have `HANDLE` arguments where a
+            // null handle indicates an absent value, which wouldn't work if
+            // null were a valid handle value, so it seems very unlikely that
+            // it could ever return null. But who knows?
+            //
+            // [`CreateFileW`]: https://docs.microsoft.com/en-us/windows/win32/api/fileapi/nf-fileapi-createfilew
+            panic!("An `OptionFileHandle` was null!");
+        }
+    }
+}
```
joshtriplett, line 154:
```suggestion
    fn try_from(handle_or_invalid: HandleOrInvalid) -> Result<Self, ()> {
        // In theory, we ought to be able to assume that the pointer here is
        // never null, use `OwnedHandle` rather than `Option<OwnedHandle>`, and
        // obviate the the panic path here.  Unfortunately, Win32 documentation
        // doesn't explicitly guarantee this anywhere.
        //
        // APIs like [`CreateFileW`] itself have `HANDLE` arguments where a
        // null handle indicates an absent value, which wouldn't work if null
        // were a valid handle value, so it seems very unlikely that it could
        // ever return null. But who knows?
        //
        // [`CreateFileW`]: https://docs.microsoft.com/en-us/windows/win32/api/fileapi/nf-fileapi-createfilew
        let owned_handle = handle_or_invalid.0.expect("A `HandleOrInvalid` was null!");
        if owned_handle.0.as_ptr() == c::INVALID_HANDLE_VALUE {
            Err(())
        } else {
            owned_handle
        }
    }
}
```

(Might also be nice to have a function `OwnedHandle::is_invalid`, but I only think that's worth factoring out if we want to make it public, and ideally anyone who might otherwise want to call that should be using `HandleOrInvalid` instead.)

library/std/src/os/windows/io/handle.rs
```diff
@@ -0,0 +1,427 @@
...
+
+impl From<OwnedHandle> for OptionFileHandle {
+    #[inline]
+    fn from(owned: OwnedHandle) -> Self {
+        let handle = owned.handle;
+        forget(owned);
+        Self { handle: handle.as_ptr() }
+    }
+}
+
```
joshtriplett, line 164:
```suggestion
```

library/std/src/os/windows/io/handle.rs
```diff
@@ -0,0 +1,427 @@
...
+    /// [`RegCloseKey`]: https://docs.microsoft.com/en-us/windows/win32/api/winreg/nf-winreg-regclosekey
+    /// [here]: https://devblogs.microsoft.com/oldnewthing/20040302-00/?p=40443
+    #[inline]
+    unsafe fn from_raw_handle(handle: RawHandle) -> Self {
+        assert!(!handle.is_null());
+        Self { handle: NonNull::new_unchecked(handle) }
+    }
+}
+
+impl FromRawHandle for OptionFileHandle {
```
joshtriplett, line 215:
```suggestion
impl FromRawHandle for HandleOrInvalid {
```

library/std/src/os/windows/io/handle.rs
```diff
@@ -0,0 +1,427 @@
...
+    /// Constructs a new instance of `Self` from the given raw handle returned
```
joshtriplett, line 216:
```suggestion
    /// Constructs a new instance of `Self` from the given `RawHandle` returned
```

library/std/src/os/windows/io/handle.rs
```diff
@@ -0,0 +1,427 @@
...
+    /// The resource pointed to by `handle` must be either open and otherwise
+    /// unowned, or equal to `INVALID_HANDLE_VALUE` (-1). It must not be null.
+    /// Note that not all Windows APIs use `INVALID_HANDLE_VALUE` for errors;
+    /// see [here] for the full story.
+    ///
+    /// [here]: https://devblogs.microsoft.com/oldnewthing/20040302-00/?p=40443
+    #[inline]
+    unsafe fn from_raw_handle(handle: RawHandle) -> Self {
+        assert!(!handle.is_null());
+        Self { handle }
```
joshtriplett, line 234:
```suggestion
        // We require non-null here to catch errors earlier.
        Self(Some(OwnedHandle::from_raw_handle(handle)))
```

library/std/src/os/windows/io/handle.rs
```diff
@@ -0,0 +1,427 @@
...
+
+impl Drop for OptionFileHandle {
+    #[inline]
+    fn drop(&mut self) {
+        unsafe {
+            let _ = c::CloseHandle(self.handle);
+        }
+    }
+}
+
```
joshtriplett, line 255:
```suggestion
```

Making `HandleOrInvalid` a wrapper around `Option<OwnedHandle>` means we reuse `OwnedHandle`'s `Drop` impl.

library/std/src/os/windows/io/handle.rs
```diff
@@ -0,0 +1,427 @@
...
+        f.debug_struct("OwnedHandle").field("handle", &self.handle).finish()
+    }
+}
+
+impl fmt::Debug for OptionFileHandle {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        f.debug_struct("OptionFileHandle").field("handle", &self.handle).finish()
+    }
+}
+
```
joshtriplett, line 273:
```suggestion
```

Deleted in favor of `derive(Debug)`.

library/std/src/os/windows/io/handle.rs
```diff
@@ -0,0 +1,427 @@
...
+    handle: RawHandle,
+}
+
+// The Windows [`HANDLE`] type may be transferred across and shared between
+// thread boundaries (despite containing a `*mut void`, which in general isn't
+// `Send` or `Sync`).
+//
+// [`HANDLE`]: std::os::windows::raw::HANDLE
+unsafe impl Send for OwnedHandle {}
+unsafe impl Send for OptionFileHandle {}
```
joshtriplett, line 90:
```suggestion
unsafe impl Send for HandleOrInvalid {}
```

library/std/src/os/windows/io/handle.rs
```diff
@@ -0,0 +1,427 @@
...
+unsafe impl Send for BorrowedHandle<'_> {}
+unsafe impl Sync for OwnedHandle {}
+unsafe impl Sync for OptionFileHandle {}
```
joshtriplett, line 93:
```suggestion
unsafe impl Sync for HandleOrInvalid {}
```

library/std/src/os/windows/io/handle.rs
```diff
@@ -0,0 +1,427 @@
...
+/// This closes the handle on drop.
+///
+/// This uses `repr(transparent)` and has the representation of a host handle,
+/// so it can be used in FFI in places where a handle is passed as a consumed
+/// argument or returned as an owned value, and is never null.
+///
+/// Note that it *may* have the value `INVALID_HANDLE_VALUE` (-1), which is
+/// sometimes a valid handle value. See [here] for the full story. For APIs
+/// like `CreateFileW` which report errors with `INVALID_HANDLE_VALUE` instead
+/// of null, use [`OptionFileHandle`] instead of `Option<OwnedHandle>`.
```
joshtriplett, line 48:
```suggestion
/// of null, use [`HandleOrInvalid`] instead of `Option<OwnedHandle>`.
```

library/std/src/os/windows/io/handle.rs
```diff
@@ -0,0 +1,427 @@
...
+        let handle = self.handle.as_ptr();
+        forget(self);
+        handle
+    }
+}
+
+impl FromRawHandle for OwnedHandle {
+    /// Constructs a new instance of `Self` from the given raw handle.
+    ///
+    /// Use `OptionFileHandle` instead of `Option<OwnedHandle>` for APIs that
```
joshtriplett, line 188:
```suggestion
    /// Use `HandleOrInvalid` instead of `Option<OwnedHandle>` for APIs that
```

library/std/src/os/windows/io/handle.rs
```diff
@@ -0,0 +1,427 @@
...
+        Self { handle: NonNull::new_unchecked(handle) }
+    }
+}
+
+impl FromRawHandle for OptionFileHandle {
+    /// Constructs a new instance of `Self` from the given raw handle returned
+    /// from a Windows API that uses `INVALID_HANDLE_VALUE` to indicate
+    /// failure, such as `CreateFileW`.
+    ///
+    /// Use `Option<OwnedHandle>` instead of `OptionFileHandle` for APIs that
```
joshtriplett, line 217:
```suggestion
    /// Use `Option<OwnedHandle>` instead of `HandleOrInvalid` for APIs that
```

library/std/src/os/windows/mod.rs
```diff
@@ -32,8 +32,11 @@ pub mod prelude {
     pub use super::fs::{MetadataExt, OpenOptionsExt};
     #[doc(no_inline)]
     #[stable(feature = "rust1", since = "1.0.0")]
-    pub use super::io::{AsRawHandle, AsRawSocket, RawHandle, RawSocket};
+    pub use super::io::{
+        AsHandle, AsSocket, BorrowedHandle, BorrowedSocket, FromRawHandle, FromRawSocket,
+        IntoRawHandle, IntoRawSocket, OptionFileHandle, OwnedHandle, OwnedSocket,
```
joshtriplett, line 37:
```suggestion
        HandleOrInvalid, IntoRawHandle, IntoRawSocket, OwnedHandle, OwnedSocket,
```

library/std/src/os/windows/io/handle.rs
```diff
@@ -0,0 +1,385 @@
...
+        // were a valid handle value, so it seems very unlikely that it could
+        // ever return null. But who knows?
+        //
+        // [`CreateFileW`]: https://docs.microsoft.com/en-us/windows/win32/api/fileapi/nf-fileapi-createfilew
+        let owned_handle = handle_or_invalid.0.expect("A `HandleOrInvalid` was null!");
+        if owned_handle.0.as_ptr() == c::INVALID_HANDLE_VALUE {
+            Err(())
+        } else {
+            owned_handle
+        }
```
joshtriplett, line 134:
```suggestion
        if owned_handle.handle.as_ptr() == c::INVALID_HANDLE_VALUE {
            Err(())
        } else {
            Ok(owned_handle)
        }
```

joshtriplett:
LGTM; r=me when it passes CI. I'll r+ it later today if someone else hasn't already.

voidc:
Hi, #81825 just got merged whic introduces the [`PidFd`](https://github.com/rust-lang/rust/blob/master/library/std/src/os/linux/process.rs) type which wraps a `FileDesc`. I presume you will have to adjust [this](https://github.com/rust-lang/rust/blob/master/library/std/src/sys/unix/process/process_unix.rs#L550) code.

sunfishcode:
Thanks for the heads up! I've now rebased and updated the PR for the PidFd changes.

m-ou-se:
Failure seems unrelated?

@bors retry

library/std/src/lib.rs
```diff
@@ -282,6 +282,7 @@
 #![feature(int_log)]
 #![feature(into_future)]
 #![feature(intra_doc_pointers)]
+#![feature(io_safety)]
```
Urgau, line 285:
```suggestion
#![cfg_attr(bootstrap, feature(io_safety))]
```

m-ou-se: I don't think the problem is it not being available during bootstrap, but instead it not being available on this specific target.

Urgau: Hay, that makes much more sense. Sorry for the noise.

sunfishcode:
I've now removed that `#![feature(io_safety)]` line. It looks like those feature declarations are for enabling language features that the code needs, which isn't needed here.

sunfishcode:
I've now fixed the merge conflict.

I don't know anything about the CI failure. I submitted https://github.com/alexcrichton/cc-rs/pull/613 which may eventually help with the opaque error message.

sunfishcode:
I've now found and fixed a bug in `spawn` on Windows, which appears to be what's failing in the CI.

The code was calling `into_inner()` on a `Handle` to leak the raw handle, but that returns an `OwnedHandle`; it needs to call `into_raw_handle()` to fully leak the raw handle.

sunfishcode:
It appears the way the code is structured is incompatible with rustdoc's way of building all the unix code on windows. I'll look for ways to reorganize the code.

sunfishcode:
https://github.com/rust-lang/rust/pull/88012 changes WASI's `RawFd` to match `unix`'s `RawFd`, which is a change I've been thinking about for some time now, as it will avoid a lot of awkward casting in code that's portable to both Unix and WASI.

It will also help this PR, by making it possible to refactor the special rustdoc module to use a single shared definition of `RawFd`, since it isn't possible to have two different definitions while still sharing the code that uses them, because all the Unix and WASI code is compiled with `cfg(windows)` in this configuration.

joshtriplett:
@sunfishcode That change seems perfectly reasonable. If it ever becomes necessary, you could always document that in WASI, "negative" file descriptors are valid as long as they aren't `-1`.

sunfishcode:
I've now rebased this on top of #88012, and factored out the `RawFd`/`AsRawFd`/etc. code between Unix and WASI, into library/std/src/os/fd/*, which fixes the issue in the `doc` build, and is a nice simplification besides.

joshtriplett:
Once more unto the
@bors r+

sunfishcode:
I just found a few additional doc test failures on Windows, now fixed.

r? @joshtriplett

m-ou-se:
This will be available in nightly in about 10 hours from now! :rocket:

@sunfishcode Thanks for all your effort and making this happen!

Nashenas88:
In case this is useful, we discovered that this change adds ambiguity for code that was relying on `where T: From<F>, F: FromRawFd` (roughly). This was a simple fix on our side, but we did see this breakage:

```
[115822/211194] RUST password_authenticator_integration_test exe.unstripped/password_authenticator_integration_test
FAILED: password_authenticator_integration_test exe.unstripped/password_authenticator_integration_test
mkdir -p ./exe.unstripped &&  RUST_BACKTRACE=1 ../../../recipe_cleanup/rustupp6g4sd/bin/rustc --color=always --crate-name password_authenticator_integration_test ../../src/identity/tests/password_a...
error[E0283]: type annotations needed
   --> ../../src/identity/tests/password_authenticator_integration/src/account_manager.rs:128:17
    |
128 |                 fs::File::from(fdio::create_fd(ramdisk_handle).expect("create fd of dev root"));
    |                 ^^^^^^^^^^^^^^ --------------------------------------------------------------- this method call resolves to `T`
    |                 |
    |                 cannot infer type for type parameter `T` declared on the trait `From`
    |
    = note: cannot satisfy `std::fs::File: From<_>`
note: required by `from`

error[E0283]: type annotations needed
   --> ../../src/identity/tests/password_authenticator_integration/src/account_manager.rs:250:9
    |
250 |           fs::File::from(
    |           ^^^^^^^^^^^^^^ cannot infer type for type parameter `T` declared on the trait `From`
251 | /             fdio::create_fd(
252 | |                 dev_root_proxy
253 | |                     .into_channel()
254 | |                     .expect("Could not convert dev root DirectoryProxy into channel")
...   |
257 | |             )
258 | |             .expect("create fd of dev root"),
    | |____________________________________________- this method call resolves to `T`
    |
    = note: cannot satisfy `std::fs::File: From<_>`
note: required by `from`

For more information about this error, try `rustc --explain E0283`.
```

Where `fdio::create_fd` has a signature of `pub fn create_fd<F: FromRawFd>(handle: zx::Handle) -> Result<F, zx::Status>`

PR #87330, opened 2021-07-21
Use hashbrown's `extend_reserve()` in `HashMap`

inquisitivecrystal:
When we added `extend_reserve()` to our implementation of `Extend` for `HashMap`, hashbrown didn't have a version we could use. Now that hashbrown has added it, we should use its version instead of implementing it ourself.


JohnTitor (approved):

JohnTitor:
Looks good, thanks!
r? @JohnTitor @bors r+ rollup

PR #87331, opened 2021-07-21
Escape item search summaries

camelid:
I noticed that `Pin::new()`'s search summary looked off, and I realized
that the reason is that it has inline code containing `Pin<P>`, which is
not escaped and thus renders as a paragraph tag!

### DIFF
--- a/src/librustdoc/html/markdown.rs
+++ b/src/librustdoc/html/markdown.rs
@@ -34,14 +34,13 @@ use std::str;
 use crate::clean::RenderedLink;
 use crate::doctest;
 use crate::html::escape::Escape;
+use crate::html::format::Buffer;
 use crate::html::highlight;
 use crate::html::toc::TocBuilder;
 
 use pulldown_cmark::{
     html, BrokenLink, CodeBlockKind, CowStr, Event, LinkType, Options, Parser, Tag,
 };
-
-use super::format::Buffer;
 
 #[cfg(test)]
 mod tests;
@@ -1086,7 +1085,7 @@ fn markdown_summary_with_limit(
     let mut stopped_early = false;
 
     fn push(s: &mut String, text_length: &mut usize, text: &str) {
-        s.push_str(text);
+        write!(s, "{}", Escape(text)).unwrap();
         *text_length += text.len();
     }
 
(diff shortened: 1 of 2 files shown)


camelid:
## Test case

```rust
/// Construct a new `Pin<P>` with this function.
pub fn new() {}
```

## Output (before)

image

## Output (after)

image

src/librustdoc/html/markdown.rs
```diff
@@ -1091,7 +1087,8 @@ fn markdown_summary_with_limit(
                         break 'outer;
                     }
 
-                    push(&mut s, &mut text_length, word);
+                    s.push_str(word);
```
camelid, line 1090: Should I escape text too?

GuillaumeGomez: I guess it'd be better to do so.

src/librustdoc/html/markdown.rs
```diff
@@ -1064,11 +1065,6 @@ fn markdown_summary_with_limit(
     let mut text_length = 0;
     let mut stopped_early = false;
 
-    fn push(s: &mut String, text_length: &mut usize, text: &str) {
```
GuillaumeGomez, line 1067: Why removing this function? It allowed us to automatically update`text_length`, which seems much better than doing it "manually" every time...

camelid: Because the escaped HTML will be longer than the original text; e.g., `Rust's "type system"` would turn into something like `Rust&#39; &#34;type system&#34;`, which is 12 characters longer than how it will be displayed. I'm planning to refactor this code to separate the length-limiting and rendering parts soon, but for now I think it's best to just increase the length manually.

GuillaumeGomez: In 2 out of the 3 cases, you use `Escape` and the last one is simply a whistespace so it can be escaped without issues as well. Then why not change the `push` to this:

```rust
fn push(s: &mut String, text_length: &mut usize, text: &str) {
    write!(s, "{}", Escape(text)).unwrap();
    *text_length += text.len();
}
```

?

camelid: That's a good point; I think I didn't do that because at first I was just escaping code :)

I'll change it to do that.

GuillaumeGomez (approved):

GuillaumeGomez:
Thanks! r=me once CI pass.

camelid:
Weird... CI passed but I had accidentally removed the import of `Escape`. Bug in CI?

PR #87332, opened 2021-07-21
Don't hide fields of enum struct variants

camelid:
* The toggle adds visual clutter
* It's easy to miss that there are fields
* Tuple variant fields are always shown, so it is inconsistent to hide
  struct variant fields
* It's annoying to have to click the toggle every time

### DIFF
--- a/src/test/rustdoc/toggle-item-contents.rs
+++ b/src/test/rustdoc/toggle-item-contents.rs
@@ -62,8 +62,7 @@ pub struct PrivStruct {
 }
 
 // @has 'toggle_item_contents/enum.Enum.html'
-// @count - '//details[@class="rustdoc-toggle type-contents-toggle"]' 1
-// @has - '//details[@class="rustdoc-toggle type-contents-toggle"]' 'Show 2 fields'
+// @!has - '//details[@class="rustdoc-toggle type-contents-toggle"]'
 pub enum Enum {
     A, B, C,
     D {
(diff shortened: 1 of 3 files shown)


camelid:
## Example

```rust
pub enum Enum {
    Variant1 { x: i32, y: String },
    Variant2(Vec<u32>, std::path::PathBuf),
}
```

## Output (before)

image

## Output (after)

image

src/test/rustdoc/toggle-item-contents.rs
```diff
@@ -62,8 +62,7 @@ pub struct PrivStruct {
 }
 
 // @has 'toggle_item_contents/enum.Enum.html'
-// @count - '//details[@class="rustdoc-toggle type-contents-toggle"]' 1
-// @has - '//details[@class="rustdoc-toggle type-contents-toggle"]' 'Show fields'
+// @!has - '//details[@class="rustdoc-toggle type-contents-toggle"]'
```
GuillaumeGomez, line 65: Can you also check the "DOM path" please? `!has` doesn't cover changes, only that something doesn't exist after all.

camelid: Wouldn't this test be more general than checking a specific DOM path? I feel a bit confused about what you'd like me to change the test to. Could you give an example of what you mean?

GuillaumeGomez: I was thinking that the `<details>` tag was wrapping another tag, but this is not the case. I really don't like that but I guess we can't do better.

GuillaumeGomez:
Thanks!

@bors: r+

PR #87337, opened 2021-07-21
Don't abort compilation after giving a lint error

### LINKED ISSUE #82761: lint errors make rustc abort early
In general, the compiler tries to keep going after errors to emit as many helpful diagnostics as possible. This isn't always possible because later stages of compilation might need certain conditions to be upheld, so sometimes the compiler aborts early.

It should never be necessary to abort early due to a lint ~~from an external tool (e.g. clippy)~~ because the compiler's hard errors should be sufficient to enforce conditions needed by later stages.

At present, however, error-level lints ~~emitted by external tools~~ can cause compilation to abort early. It seems unfortunate that, after fixing all the problems reported by one run of clippy, you could run clippy again (perhaps by committing your "fixed" code and updating a PR, causing CI to run) and discover there are more problems (when CI fails).

As an example, this code:

```rust
fn main() {
    if true {} {}
    
    let x;
    if true {
        x = 1;
    } else {
        x = 1;
    }
    println!("{}", x);
}
```

Emits two lints with clippy:

```
warning: this looks like an `else {..}` but the `else` is missing
 --> src/main.rs:4:15
  |
4 |     if true {} {}
  |               ^
  |
  = note: `#[warn(clippy::suspicious_else_formatting)]` on by default
  = note: to remove this lint, add the missing `else` or add a new line before the next block
  = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#suspicious_else_formatting

error: this `if` has identical blocks
  --> src/main.rs:9:12
   |
9  |       } else {
   |  ____________^
10 | |         x = 1;
11 | |     }
   | |_____^
   |
   = note: `#[deny(clippy::if_same_then_else)]` on by default
note: same as this
  --> src/main.rs:7:13
   |
7  |       if true {
   |  _____________^
8  | |         x = 1;
9  | |     } else {
   | |_____^
   = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#if_same_then_else

error: aborting due to previous error; 1 warning emitted

```

If `#![deny(clippy::all)]` is added, only the first lint is emitted, and the compiler aborts after it:

```
error: this looks like an `else {..}` but the `else` is missing
 --> src/main.rs:4:15
  |
4 |     if true {} {}
  |               ^
  |
 --> src/main.rs:1:8
  |
1 | #[deny(clippy::all)]
  |        ^^^^^^^^^^^
  = note: `#[deny(clippy::suspicious_else_formatting)]` implied by `#[deny(clippy::all)]`
  = note: to remove this lint, add the missing `else` or add a new line before the next block
  = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#suspicious_else_formatting

```

@rustbot label A-lint

jyn514:
The only reason to use `abort_if_errors` is when the program is so broken that either:
1. later passes get confused and ICE
2. any diagnostics from later passes would be noise

This is never the case for lints, because the compiler has to be able to deal with `allow`-ed lints.
So it can continue to lint and compile even if there are lint errors.

Closes https://github.com/rust-lang/rust/issues/82761.

### DIFF
--- a/compiler/rustc_driver/src/lib.rs
+++ b/compiler/rustc_driver/src/lib.rs
@@ -410,10 +410,6 @@ fn run_compiler(
                 sess.code_stats.print_type_sizes();
             }
 
-            if sess.diagnostic().has_errors_or_lint_errors() {
-                return Err(ErrorReported);
-            }
-
             let linker = queries.linker()?;
             Ok(Some(linker))
         })?;
--- a/compiler/rustc_driver/src/lib.rs
+++ b/compiler/rustc_driver/src/lib.rs
@@ -410,6 +410,10 @@ fn run_compiler(
                 sess.code_stats.print_type_sizes();
             }
 
+            if sess.diagnostic().has_errors_or_lint_errors() {
+                return Err(ErrorReported);
+            }
+
             let linker = queries.linker()?;
             Ok(Some(linker))
         })?;
--- a/compiler/rustc_errors/src/lib.rs
+++ b/compiler/rustc_errors/src/lib.rs
@@ -411,6 +411,8 @@ pub struct Handler {
 /// as well as inconsistent state observation.
 struct HandlerInner {
     flags: HandlerFlags,
+    /// The number of lint errors that have been emitted.
+    lint_err_count: usize,
     /// The number of errors that have been emitted, including duplicates.
     ///
     /// This is not necessarily the count that's reported to the user once
@@ -550,6 +552,7 @@ impl Handler {
             flags,
             inner: Lock::new(HandlerInner {
                 flags,
+                lint_err_count: 0,
                 err_count: 0,
                 warn_count: 0,
                 deduplicated_err_count: 0,
@@ -726,7 +729,13 @@ impl Handler {
     /// Construct a builder at the `Error` level with the `msg`.
     // FIXME: This method should be removed (every error should have an associated error code).
     pub fn struct_err(&self, msg: &str) -> DiagnosticBuilder<'_> {
-        DiagnosticBuilder::new(self, Level::Error, msg)
+        DiagnosticBuilder::new(self, Level::Error { lint: false }, msg)
+    }
+
+    /// This should only be used by `rustc_middle::lint::struct_lint_level`. Do not use it for hard errors.
+    #[doc(hidden)]
+    pub fn struct_err_lint(&self, msg: &str) -> DiagnosticBuilder<'_> {
+        DiagnosticBuilder::new(self, Level::Error { lint: true }, msg)
     }
 
     /// Construct a builder at the `Error` level with the `msg` and the `code`.
@@ -790,11 +799,14 @@ impl Handler {
     }
 
     pub fn span_err(&self, span: impl Into<MultiSpan>, msg: &str) {
-        self.emit_diag_at_span(Diagnostic::new(Error, msg), span);
+        self.emit_diag_at_span(Diagnostic::new(Error { lint: false }, msg), span);
     }
 
     pub fn span_err_with_code(&self, span: impl Into<MultiSpan>, msg: &str, code: DiagnosticId) {
-        self.emit_diag_at_span(Diagnostic::new_with_code(Error, Some(code), msg), span);
+        self.emit_diag_at_span(
+            Diagnostic::new_with_code(Error { lint: false }, Some(code), msg),
+            span,
+        );
     }
 
     pub fn span_warn(&self, span: impl Into<MultiSpan>, msg: &str) {
@@ -861,6 +873,9 @@ impl Handler {
 
     pub fn has_errors(&self) -> bool {
         self.inner.borrow().has_errors()
+    }
+    pub fn has_errors_or_lint_errors(&self) -> bool {
+        self.inner.borrow().has_errors_or_lint_errors()
     }
     pub fn has_errors_or_delayed_span_bugs(&self) -> bool {
         self.inner.borrow().has_errors_or_delayed_span_bugs()
@@ -979,7 +994,11 @@ impl HandlerInner {
             }
         }
         if diagnostic.is_error() {
-            self.bump_err_count();
+            if matches!(diagnostic.level, Level::Error { lint: true }) {
+                self.bump_lint_err_count();
+            } else {
+                self.bump_err_count();
+            }
         } else {
             self.bump_warn_count();
         }
@@ -1073,11 +1092,14 @@ impl HandlerInner {
     fn has_errors(&self) -> bool {
         self.err_count() > 0
     }
+    fn has_errors_or_lint_errors(&self) -> bool {
+        self.has_errors() || self.lint_err_count > 0
+    }
     fn has_errors_or_delayed_span_bugs(&self) -> bool {
         self.has_errors() || !self.delayed_span_bugs.is_empty()
     }
     fn has_any_message(&self) -> bool {
-        self.err_count() > 0 || self.warn_count > 0
+        self.err_count() > 0 || self.lint_err_count > 0 || self.warn_count > 0
     }
 
     fn abort_if_errors(&mut self) {
@@ -1131,7 +1153,7 @@ impl HandlerInner {
     }
 
     fn err(&mut self, msg: &str) {
-        self.emit_error(Error, msg);
+        self.emit_error(Error { lint: false }, msg);
     }
 
     /// Emit an error; level should be `Error` or `Fatal`.
@@ -1165,6 +1187,11 @@ impl HandlerInner {
         if has_bugs {
             panic!("{}", explanation);
         }
+    }
+
+    fn bump_lint_err_count(&mut self) {
+        self.lint_err_count += 1;
+        self.panic_if_treat_err_as_bug();
     }
 
     fn bump_err_count(&mut self) {
@@ -1210,7 +1237,10 @@ impl DelayedDiagnostic {
 pub enum Level {
     Bug,
     Fatal,
-    Error,
+    Error {
+        /// If this error comes from a lint, don't abort compilation even when abort_if_errors() is called.
+        lint: bool,
+    },
     Warning,
     Note,
     Help,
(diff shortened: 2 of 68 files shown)


flip1995:
Copying my comment about this from Zulip:

> If you don't want the compiler to abort on a deny-by-default lint, then you can just use --cap-lints=warn. Preserving the lint levels and keep going even if there is an error makes lint levels completely nonsensical.

> IMO not aborting after an error, even with lints, is a bad idea. Here's why:

> The compiler works in stages / IRs and so do lints. So there is the AST, where source-level checks are done, the (T)HIR where type checking and similar things happen and the MIR, where borrow checking and the like happen. If you already have erroneous code at the source-level AST, then you should fix those first, before the linter tries to check type information or borrowing stuff.

> Making the compiler not stop after an error would render different lint levels useless. The only difference then would be that they have a different color in the terminal.

> So I'm against this change, especially since there already is the --cap-lints flag where you can tell the compiler to ignore all lint levels above a specific level.

---

This will definitely fail the Clippy test suite, since our UI tests make the assumption that an error in an `EarlyPass` lint will prevent `LatePassLint`s from running.  And so does rustc:

> `test result: FAILED. 11623 passed; 424 failed;`

SNCPlay42:
> Making the compiler not stop after an error would render different lint levels useless. The only difference then would be that they have a different color in the terminal.

What I imagined when filing #82761 was not that the compiler would not abort on error-level lints, but that it would not abort *early*. That preserves an important difference between error and warning lint levels: a lint being error-level would still cause the compiler to not emit compilation artifacts and exit with nonzero status (resulting in e.g. a CI run reporting failure instead of success), unlike a warning. I don't think anyone is advocating that compilation should actually *succeed* if there are error level lints (in situations where that is desirable, as you note, `--cap-lints` exists); I think it would be a bug in this PR if it causes that to happen.

> If you already have erroneous code at the source-level (AST), then you should fix those first, before the linter tries to check type information or borrowing stuff.

This is reasonable for hard errors (and sometimes completely necessary - for instance the compiler can't typecheck code that it can't parse into an AST), but I'm not convinced this is the case for lints, especially those that come from external tools like clippy - it seems less likely that a problem that would trigger a clippy lint, but otherwise would still allow the code to be compiled, would cause that kind of cascade of errors where the later diagnostics don't add value.

flip1995:
> but I'm not convinced this is the case for lints, especially those that come from external tools like clippy

There are lints in Clippy that don't check for specific things, since they will be caught by earlier (rustc) lints. I don't know if this will lead to ICEs (worst case) or just some contradicting lint output, or if it just doesn't matter.

> not that the compiler would not abort on error-level lints, but that it would not abort _early_.

Thinking about this, my concerns may be a bit unwarranted. 

Since to get this PR through, **many** `allow`s will have to be added to the UI tests, I think we'd find if lints really depend on earlier lints. So I'd be happy with this, if my concerns will be proven wrong in the test suite(s).

jyn514:
@flip1995 oh, I was just going to bless the tests, so you can see exactly what changed (I've already done that locally but I got distracted by preserving the error status). Would you prefer that I use allows instead? That seems less helpful, you wouldn't see contradictory lints.

flip1995:
The Clippy test suite is designed so that one test (mostly named after the lint) tests exactly one lint. If  that file accidentally triggers another lint, we just allow that lint instead of carefully rewriting the test. Either way, we definitely don't want to clobber the `stderr` with output from unrelated lints.

jyn514:
FWIW, another use case for this is when `deny(warnings)` is set - I've found it frustrating when working on rustc itself that I only see the first couple errors before the others show up.

flip1995:
I just ran the Clippy test suite for those changes:

```
test result: FAILED. 632 passed; 16 failed; 3 ignored; 0 measured; 0 filtered out; finished in 9.83s
```

So only 16 tests are effected. That's a surprisingly low number.

LeSeulArtichaut:
Would this change warrant a MCP?

jyn514:
@LeSeulArtichaut I'm happy to open an MCP, I'll do that this afternoon. I don't think it's necessarily *required*, but it doesn't hurt.

LeSeulArtichaut:
> So only 16 tests are effected. That's a surprisingly low number.

I'd guess that is because most tests only test one lint?

jyn514:
Hmm, that latest failure is a little concerning - the issue is that rustdoc never calls `RunCompiler::run`, so it doesn't get the check for `has_errors_or_lint_errors()`. I can of course add the check in rustdoc, but clippy and other external tools will have the same issue.

Maybe I can add the check to `create_compiler_and_run`? But that will check it *after* the tool has finished everything it's going to do with the compiler; so for example rustdoc would still create HTML before exiting with an error.

compiler/rustc_driver/src/lib.rs
```diff
@@ -410,6 +410,10 @@ fn run_compiler(
                 sess.code_stats.print_type_sizes();
             }
 
+            if sess.diagnostic().has_errors_or_lint_errors() {
+                return Err(ErrorReported);
+            }
```
bjorn3, line 415: You should rarely see an error from codegen, so maybe already stop compilation before codegen?

compiler/rustc_errors/src/lib.rs
```diff
@@ -864,7 +879,11 @@ impl HandlerInner {
...
         if diagnostic.is_error() {
-            self.bump_err_count();
+            if matches!(diagnostic.level, Level::Error { lint: true }) {
+                self.bump_lint_err_count();
+            } else {
+                self.bump_err_count();
+            }
         } else {
             self.bump_warn_count();
         }
```
oli-obk, line 889: at this point, this should probably just be an exhaustive match on `diagnostic.level`

compiler/rustc_driver/src/lib.rs
jyn514: There are still some, so the check still needs to exist after. But I can check more than once, that's not a big deal.

See also https://github.com/rust-lang/rust/pull/87337#issuecomment-884622217, I'm concerned about how this could work for tools.

jyn514: (and it's not that rare - arithmetic_overflow isn't shown until codegen)

compiler/rustc_errors/src/lib.rs
jyn514: Hmm, I was hoping to keep using `is_error` to keep it in sync with `deduplicated_err_count` above.

JohnCSimon:
triage: merge conflicts.

jyn514:
I was hoping to hear ideas about https://github.com/rust-lang/rust/pull/87337#issuecomment-884622217 before putting more time into this. I don't know the best path forward.

bjorn3:
> I am not sure what rustdoc would need in order to be able to use that API.

Rustdoc does it's own argument parsing as it has a different commandline interface. `RunCompiler` uses rustc's argument parsing.

jyn514:
> Clippy and miri already use RunCompiler, so they should be fine

If other tools are using run_compiler, it's probably fine to add the check there and also add a manual check to rustdoc, it's not anywhere near the worst hack it has. I'll try to get to that this weekend.

flip1995 (commented):
Ahhhh, 13bcbb0 just reverts some changes done in  https://github.com/rust-lang/rust/pull/87337/commits/f903ea453b53bf894293a25ea67366756dfd3727. So this is fine. There were only two test files left that looked suspicious. I pushed a fixup commit. After squashing the Clippy changes, this LGTM.

jyn514:
Ok cool, I think this is just waiting for review from @oli-obk then :)

oli-obk:
Thanks @flip1995 for the review and help on clippy.

@bors r=oli-obk,flip1995

jyn514:
Hmm, I think this is because I added `abort_if_errors()` before the call to `ongoing_codegen()`. @bjorn3 are you sure it's a good idea to stop because of lint errors? If so I can try and figure out how to update the test, but it seems better to keep the earlier behavior.

I guess another alternative is to *only* abort if there are lint errors, and not if there are hard errors.

bjorn3:
I guess it is fine to not do the abort before codege in rustc_driver. There should be an `abort_on_err` before linking somewhere though that needs to also abort for lint errors.

jyn514:
I found why the behavior changed before - there's a function called `compile_status` which only codegen calls. I changed it to use `has_errors_or_lint_errors()` and now neither of the checks in `rustc_driver` are necessary :)

jyn514:
I managed to mess up the rebase again and test src/tools/clippy --bless isn't working :/ 

<details>

```
failures:
    [ui] ui/crashes/ice-3969.rs

test result: FAILED. 661 passed; 1 failed; 3 ignored; 0 measured; 0 filtered out; finished in 9.40s

test compile_test ... FAILED

failures:

---- compile_test stdout ----
diff of stderr:

-error: trait bound str: std::marker::Sized does not depend on any type or lifetime parameters
-  --> $DIR/ice-3969.rs:20:10
-   |
-LL |     str: Sized;
-   |          ^^^^^
-   |
-   = note: `-D trivial-bounds` implied by `-D warnings`
-
-error: trait bound for<'a> Dst<(dyn A + 'a)>: std::marker::Sized does not depend on any type or lifetime parameters
-  --> $DIR/ice-3969.rs:24:30
-   |
-LL |     for<'a> Dst<dyn A + 'a>: Sized,
-   |                              ^^^^^
-
-error: trait bound str: std::marker::Sized does not depend on any type or lifetime parameters
-  --> $DIR/ice-3969.rs:31:10
-   |
-LL |     str: Sized,
-   |          ^^^^^
-
-error: trait bound std::string::String: std::ops::Neg does not depend on any type or lifetime parameters
-  --> $DIR/ice-3969.rs:38:13
-   |
-LL |     String: ::std::ops::Neg<Output = String>,
-   |             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
-
-error: trait bound i32: std::iter::Iterator does not depend on any type or lifetime parameters
-  --> $DIR/ice-3969.rs:45:10
-   |
-LL |     i32: Iterator,
-   |          ^^^^^^^^
-
-error: aborting due to 5 previous errors
-
-
```

</details>

PR #87338, opened 2021-07-21
Simplify the collecting of `? Trait` bounds in where clause

SparrowLii:
This PR fixes the FIXME about using less rightward drift and only one error reporting when collecting of `?Trait` bounds in where clause.
Checking whether the path length of `bound_ty` is 1 can be replaced by whether `unresolved_segments` in the partial_res is 0. 
Checking whether the `param.kind` is `Type{...}` can also be omitted. One Fx hash calculation will be done for Const or Lifetime param, but the impact on efficiency should be small IMO


wesleywiser (approved):

wesleywiser:
Thanks, LGTM!

Let's do a perf run just to be sure there isn't much of an impact. 

@bors try @rust-timer queue

SparrowLii:
Thanks for reviewing! I'm happy it is helpful

PR #86898, opened 2021-07-05
Add fast path for Path::cmp that skips over long shared prefixes

the8472:
```
# before
test path::tests::bench_path_cmp_fast_path_buf_sort               ... bench:      60,811 ns/iter (+/- 865)
test path::tests::bench_path_cmp_fast_path_long                   ... bench:       6,459 ns/iter (+/- 275)
test path::tests::bench_path_cmp_fast_path_short                  ... bench:       1,777 ns/iter (+/- 34)

# after
test path::tests::bench_path_cmp_fast_path_buf_sort               ... bench:      38,140 ns/iter (+/- 211)
test path::tests::bench_path_cmp_fast_path_long                   ... bench:       1,471 ns/iter (+/- 24)
test path::tests::bench_path_cmp_fast_path_short                  ... bench:       1,106 ns/iter (+/- 9)
```

PR #87339, opened 2021-07-21
Make two Paths unequal if they differ in trailing slash

dtolnay:
It seems problematic to consider two paths equal if one of them can exist when the other does not exist.

In the previous behavior, the following asserts pass:

```rust
use std::path::Path;

fn main() {
    let path1 = Path::new("/etc/passwd");
    let path2 = Path::new("/etc/passwd/");

    assert_eq!(path1, path2);

    assert!(path1.metadata().is_ok());

    // Err: Os { code: 20, kind: Other, message: "Not a directory" }
    assert!(path2.metadata().is_err());
}
```

In this PR `path1 == path2` becomes false.

### DIFF
--- a/library/std/src/path.rs
+++ b/library/std/src/path.rs
@@ -2719,11 +2719,30 @@ impl Path {
         PathBuf { inner: OsString::from(inner) }
     }
 
-    fn ends_with_separator(&self) -> bool {
+    /// Returns true if the path's syntax indicates it necessarily can only
+    /// refer to a directory.
+    ///
+    /// This is relevant to the PartialEq and PartialOrd impls of Path. For
+    /// example, compared as paths "r/s/" and "r/s/." and "r/s/./" and "r/s/./."
+    /// are all in the same equivalence class, but a different class from "r/s".
+    /// That's because if "r/s" is a file then "r/s" exists while most
+    /// filesystems would consider that "r/s/" does not exist, and thus these
+    /// must not be equal paths.
+    fn syntactically_dir(&self) -> bool {
         let components = self.components();
-        match components.path.last() {
-            None => false,
-            Some(byte) => components.is_sep_byte(*byte),
+        let bytes_after_prefix =
+            &components.path[components.prefix.as_ref().map_or(0, Prefix::len)..];
+        let mut bytes_iter = bytes_after_prefix.iter();
+        let last_byte = bytes_iter.next_back();
+        let second_last_byte = bytes_iter.next_back();
+        match (second_last_byte, last_byte) {
+            // Path ends in separator, like "path/to/"
+            (_, Some(&sep)) if components.is_sep_byte(sep) => true,
+            // Path ends in separator followed by CurDir, like "path/to/."
+            (Some(&sep), Some(b'.')) if components.is_sep_byte(sep) => true,
+            // Path is ".", because "." and "./." need to be considered equal
+            (None, Some(b'.')) => true,
+            _ => false,
         }
     }
 }
@@ -2787,7 +2806,7 @@ impl cmp::PartialEq for Path {
     #[inline]
     fn eq(&self, other: &Path) -> bool {
         self.components() == other.components()
-            && self.ends_with_separator() == other.ends_with_separator()
+            && self.syntactically_dir() == other.syntactically_dir()
     }
 }
 
@@ -2816,7 +2835,7 @@ impl cmp::Ord for Path {
     #[inline]
     fn cmp(&self, other: &Path) -> cmp::Ordering {
         compare_components(self.components(), other.components())
-            .then_with(|| self.ends_with_separator().cmp(&other.ends_with_separator()))
+            .then_with(|| self.syntactically_dir().cmp(&other.syntactically_dir()))
     }
 }
 
--- a/library/std/src/path.rs
+++ b/library/std/src/path.rs
@@ -1738,7 +1738,7 @@ impl ToOwned for Path {
 impl cmp::PartialEq for PathBuf {
     #[inline]
     fn eq(&self, other: &PathBuf) -> bool {
-        self.components() == other.components()
+        Path::eq(self, other)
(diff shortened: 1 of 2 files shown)


library/std/src/path.rs
```diff
@@ -2706,15 +2715,17 @@ impl cmp::Eq for Path {}
...
     }
 }
 
 #[stable(feature = "rust1", since = "1.0.0")]
 impl cmp::Ord for Path {
     #[inline]
     fn cmp(&self, other: &Path) -> cmp::Ordering {
-        self.components().cmp(other.components())
+        self.components()
+            .cmp(other.components())
```
the8472, line 2727: It might be better to use `Ord::cmp` here too, while implementing #86898 I ran into some ambiguities around `Ord::cmp` vs. `Iterator::cmp`.

library/std/src/path.rs
```diff
@@ -2686,7 +2694,8 @@ impl fmt::Display for Display<'_> {
 impl cmp::PartialEq for Path {
     #[inline]
     fn eq(&self, other: &Path) -> bool {
-        self.components().eq(other.components())
+        self.components() == other.components()
+            && self.ends_with_separator() == other.ends_with_separator()
```
the8472, line 2698: Shouldn't this be implemented on the `Components` level instead of `Path`? Otherwise they'll give different results.

the8472 (commented):
Couldn't this break some use-cases? E.g. shells autocomplete directories with a trailing slash, if a CLI tool then tries to deduplicate paths or exclude paths by comparing them to another set of paths that's derived in some other way that doesn't append the trailing slash they might get a false negative compared to previous behavior.

At the very least it should be documented and have a release note.

JohnCSimon:
merge conflicts

dtolnay:
- Resolved merge conflict with #86898

library/std/src/path.rs
dtolnay: Giving different results is correct. The path components are the same i.e. the `Components` iterator would produce the identical sequence of `Component` enum values, even if one path ends in separator and the other does not.

the8472: If we say that a trailing slash is semantically different then shouldn't components preserve that semantic difference? E.g. the last component could include the trailing slash. Or there could be a final zero-length component. Ideally this would be expressed through a new enum variant but it's exhaustive so we can't extend it 🙁.

Is `assert_eq!(foo, foo.components().as_path())` always true?

library/std/src/path/tests.rs
```diff
@@ -1629,10 +1654,13 @@ fn test_ord() {
...
-    ord!(Equal, "foo/./bar", "foo/bar/");
-    ord!(Equal, "foo/bar", "foo/bar/");
+    ord!(Equal, "foo/./bar", "foo/bar");
+    ord!(Less, "foo/./bar", "foo/bar/");
+    ord!(Equal, "foo/./bar/", "foo/bar/");
+    ord!(Less, "foo/bar", "foo/bar/");
     ord!(Equal, "foo/bar", "foo/bar/.");
-    ord!(Equal, "foo/bar", "foo/bar//");
+    ord!(Less, "foo/bar", "foo/bar//");
+    ord!(Equal, "foo/bar/", "foo/bar//");
```
the8472, line 1663: How does it order `bar/.` vs. `bar/`?

yaahc:
Part of me feels like it would be nice to have eq work regardless of the trailing slash if the file being named _is_ a directory, but I feel like we almost certainly don't want to be interacting with filesystem APIs to see if paths exist and are dirs when checking equality. So with all of that said this PR looks good to me. Do we need to FCP this change since it changes runtime behavior?

yaahc:
@rfcbot merge

m-ou-se:
@rfcbot cancel

m-ou-se:
@rfcbot merge

yaahc:
@the8472 are you satisfied by @dtolnay's explanations?

library/std/src/path.rs
the8472: > I don't think it follows that we need components to preserve that semantic difference.

Well... the thing is that `Components` implements `AsRef<Path>`, so if you have a generic function taking it via `AsRef` then it'll behave differently depending on whether you pass in a `Components` or something else that implements `AsRef<Path>`.

It's consistent with existing documentation, it's clearly spelled out that `path.components()` will strip trailing slashes, but previously that wasn't considered a semantic difference (since `Eq` considered them equal), now it is.

the8472:
The ordering seems fine now.

I still think the newly introduced semantic inconsistency between `Path` and `Components` is unfortunate, but at least [components()](https://doc.rust-lang.org/std/path/struct.Path.html#method.components) does mention it is normalizing.

Maybe it just needs to be documented somewhere that `Path` considers a trailing slash semantically meaningful while `Components` does not.

yaahc:
> Maybe it just needs to be documented somewhere that `Path` considers a trailing slash semantically meaningful while `Components` does not.

Sounds good, lets do this in a follow up PR?

@bors r+

dtolnay:
I think upon reflection I am going to retract this for being more disruptive than I would have wanted.

An example of code that's negatively affected is [this snippet from cargo](https://github.com/rust-lang/cargo/blob/ad50d0d266213e0cc4f6e526a39d96faae9a3842/src/cargo/util/vcs.rs#L13-L15) where it's testing whether `path` happens to be the root of a git repo:

```rust
    if let Ok(repo) = git2::Repository::discover(path) {
        if repo.workdir().map_or(false, |workdir| workdir == path) {
```

It turns out `path` comes originally from `std::env::current_dir` with no trailing slash on my machine, while git2's `Repository::workdir` always comes back with a trailing slash. Thus the `==` comparison could now never be true.

I guess code that wants the non-equal behavior should use a newtype with PartialEq and PartialOrd which mirror the syntactically_dir logic from this PR.

Issue #61137, opened 2019-05-24
Add missing long error code explanations


GuillaumeGomez:
This is a re-opening of #32777.

## Explanations

Those error codes and their explanations are declared in files named `error_codes.rs`. In those files, you can see generally two macros: `register_diagnostics` and `register_long_diagnostics`. If an error code is in `register_diagnostics`, it means it lacks long error explanation. The goal of this issue is to add those missing explanations!

This is an easy first issue but don't hesitate to ask for help if you're stuck on something!

Here is the list of **all** error codes that don't have a long error explanation yet:

 * [x] E0183 (#89885)
 * [x] E0203 (#66880)
 * [x] E0208 (#98011)
 * [x] E0212
 * [x] E0224
 * [x] E0226
 * [x] E0227 (#93809)
 * [x] E0228
 * [x] E0279 (not emitted anymore)
 * [x] E0280 (#106521)
 * [x] E0284
 * [x] E0307 (#64110)
 * [x] E0311 (https://github.com/rust-lang/rust/pull/100747)
 * [x] E0312 (#64347)
 * [x] E0313 (#106580)
 * [x] E0316
 * [x] E0320 (https://github.com/rust-lang/rust/pull/105792)
 * [x] E0377 (https://github.com/rust-lang/rust/pull/105935)
 * [x] E0457 (#105860)
 * [x] E0460 (#105860)
 * [x] E0461 (https://github.com/rust-lang/rust/pull/106028)
 * [ ] E0462
 * [x] E0464 (#89587)
 * [x] E0465 (https://github.com/rust-lang/rust/pull/106282)
 * [x] E0472 (#105791)
 * [x] E0476 (#108444)
 * [x] E0477
 * [x] E0482 (#89710)
 * [x] E0490 (https://github.com/rust-lang/rust/pull/106714)
 * [x] E0493 (#64377)
 * [x] E0495 (https://github.com/rust-lang/rust/pull/64404)
 * [x] E0498 (#87521)
 * [x] E0514 (https://github.com/rust-lang/rust/pull/106236)
 * [x] E0519 (https://github.com/rust-lang/rust/pull/106236)
 * [x] E0521
 * [ ] E0523
 * [x] E0524 (#64428)
 * [x] E0531 (#64455)
 * [x] E0533 (https://github.com/rust-lang/rust/pull/64735)
 * [x] E0539 (error than can only appear inside libstd code)
 * [x] E0542 (error than can only appear inside libstd code)
[...]

PR #87342, opened 2021-07-21
Add long explanation for E0757

midgleyc:
Helps with #61137

### DIFF
--- a/compiler/rustc_error_codes/src/error_codes/E0757.md
+++ b/compiler/rustc_error_codes/src/error_codes/E0757.md
@@ -1,0 +1,25 @@
+A function was given both the `ffi_const` and `ffi_pure` attributes.
+
+Erroneous code example:
+
+```compile_fail,E0757
+#![feature(ffi_const, ffi_pure)]
+
+extern "C" {
+    #[ffi_const]
+    #[ffi_pure] // error: `#[ffi_const]` function cannot be `#[ffi_pure]`
+    pub fn square(num: i32) -> i32;
+}
+```
+
+As `const` has a stricter set of requirements than `pure`, remove the `ffi_pure`
+attribute:
+
+```
+#![feature(ffi_const)]
+
+extern "C" {
+    #[ffi_const]
+    pub fn square(num: i32) -> i32;
+}
+```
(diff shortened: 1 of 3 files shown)


compiler/rustc_error_codes/src/error_codes/E0757.md
```diff
@@ -0,0 +1,25 @@
...
+#![feature(ffi_const, ffi_pure)]
+
+extern "C" {
+    #[ffi_const]
+    #[ffi_pure] // error: `#[ffi_const]` function cannot be `#[ffi_pure]`
+    pub fn square(num: i32) -> i32;
+}
+```
+
+As `const` has a stricter set of requirements than `pure`, remove the `ffi_pure`
```
GuillaumeGomez, line 15: `ffi_const` you mean? Also, is there a documentation we could link to by any chance?

midgleyc: Here I mean `const`. `const` and `pure` are C function attributes (e.g. see [Common Function Attributes](https://gcc.gnu.org/onlinedocs/gcc/Common-Function-Attributes.html#Common-Function-Attributes)). Annotating a function with `ffi_const` says that it's C `const`, and annotating with `ffi_pure` says that it's C `pure`. I could use `ffi_const` and `ffi_pure` instead, but I felt it was more readable to lift the covers up to what they mean in GCC or other C compilers. What do you think?

I could link to [ffi-pure](https://doc.rust-lang.org/nightly/unstable-book/language-features/ffi-pure.html) and [ffi-const](https://doc.rust-lang.org/nightly/unstable-book/language-features/ffi-const.html) from the unstable rust book. The former does mention that `#[ffi_const]` provides stronger guarantees, which implies that it's stricter.

GuillaumeGomez: Please use `ffi_*` instead then. If I asked the questions, others will wonder about the same thing as well. Please add the links as well then.

GuillaumeGomez (approved):

GuillaumeGomez:
Thanks!

@bors: r+ rollup squash

PR #87136, opened 2021-07-14
[beta] Remove unsound TrustedRandomAccess implementations

cuviper:
This re-applies #86222 for 1.54, as the larger fixes in #85874 are still in progress.

PR #86222, opened 2021-06-11
[beta] backport: Remove unsound TrustedRandomAccess implementations

steffahn:
Beta backport of #85874, but just for `vec_deque` and `array`, as discussed [in this meeting](https://rust-lang.zulipchat.com/#narrow/stream/238009-t-compiler.2Fmeetings/topic/.5Badhoc.5D.202021-06-11.20unsound.20stdlib.20specialization.20and.201.2E53/near/242356073) ([archive](https://zulip-archive.rust-lang.org/238009tcompilermeetings/97862adhoc20210611unsoundstdlibspecializationand153.html#242356073), [hackmd](https://hackmd.io/IcaLDGBlStWYkn2qzWGxUg#Consensus)).

PR #87343, opened 2021-07-21
Regression fix to avoid further beta backports: Remove unsound TrustedRandomAccess implementations

steffahn:
Removes the implementations that depend on the user-definable trait `Copy`.

Only fix regressions to ensure merge in 1.55: Does not modify `vec::IntoIter`.

<hr>

This PR applies the beta-`1.53` backport #86222 (merged as part of #86225), a reduced version of #85874 that only fixes regressions, to `master` in order to avoid the need for further backports from `1.55` onwards. Beta-`1.54` backport already happened with #87136. In case that #85874 gets merged quickly (within a week), this PR would be unnecessary.

r? @cuviper


cuviper:
Thanks! Let's go ahead and merge this first, since #85874 has more complicated changes to review.

@bors r+

PR #87346, opened 2021-07-21
Rename force-warns to force-warn

rylev:
The renames the `--force-warns` option to `--force-warn`. This mirrors other lint options like `--warn` and `--deny` which are in the singular.

r? @nikomatsakis 

cc @ehuss - this option is being used by Cargo. How do we make sure the transition to using the new name is as smooth as possible?


ehuss:
> How do we make sure the transition to using the new name is as smooth as possible?

You don't need to do anything.  Cargo's tests that use this are set up in such a way as to be disabled on rust-lang/rust's CI.  After this lands and the next nightly is published, we'll need to update Cargo to the new flag, but I can do that, it should be a trivial change.

nikomatsakis (approved):

estebank:
This might have introduced an issue reported to me on twitter:

> Hmm.... so I'm getting the "force-warns" error, even on a freshly "cargo init"'d packages: both binary and library.
>
> I can build and run fine (even when I run cargo +nightly run [not pictured]), and I have today's nightly per rustup.
>
> I'll sleep on it and then file an issue.

PTNobel:
For a little context on @estebank 's quote of my tweet: I had been attempting to run `cargo +nightly fix --edition`. I'll try and fix the issue in Cargo now.

PTNobel:
Just opened: rust-lang/cargo#9714.

PR #87348, opened 2021-07-21
Fix span when suggesting to add an associated type bound

### LINKED ISSUE #87261: Incorrect suggestion when trying to implement `Trait<Associated = ()>` with `impl Trait + 'static`
Given the following code ([Playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=bc010f95c7ce224d84a5fdd06be94dc6)): 

```rust
trait Trait {
    type Associated;
}

fn returns_opaque() -> impl Trait + 'static {
    struct Impl;

    impl Trait for Impl {
        type Associated = ();
    }

    Impl
}

fn accepts_trait<T>(_: T)
where
    T: Trait<Associated = ()>,
{
}

fn main() {
    accepts_trait(returns_opaque());
}
```

The current output is:

```
error[E0271]: type mismatch resolving `<impl Trait as Trait>::Associated == ()`
  --> src/main.rs:22:5
   |
5  | fn returns_opaque() -> impl Trait + 'static {
   |                        -------------------- the found opaque type
...
15 | fn accepts_trait<T>(_: T)
   |    ------------- required by a bound in this
16 | where
17 |     T: Trait<Associated = ()>,
   |              --------------- required by this bound in `accepts_trait`
...
22 |     accepts_trait(returns_opaque());
   |     ^^^^^^^^^^^^^ expected `()`, found associated type
   |
   = note:    expected unit type `()`
           found associated type `<impl Trait as Trait>::Associated`
help: consider constraining the associated type `<impl Trait as Trait>::Associated` to `()`
   |
5  | fn returns_opaque() -> impl Trait + 'static<Associated = ()> {
   |                                            ^^^^^^^^^^^^^^^^^

For more information about this error, try `rustc --explain E0271`.

```

The syntax `impl Trait + 'static<Associated = ()>` does not actually work.

Ideally the suggested fix should look like:

```
help: consider constraining the associated type `<impl Trait as Trait>::Associated` to `()`
   |
5  | fn returns_opaque() -> impl Trait<Associated = ()> + 'static {
   |                                  ^^^^^^^^^^^^^^^^^
```

This bug appears to be present in stable, beta, and nightly, on both Rust 2015 and 2018.

SkiFire13:
Fixes #87261 

Note that this fix is not perfect, it ~~will still give incorrect~~ won't give suggestions in some situations:
- If the associated type is defined on a supertrait of those contained in the opaque type, it will fallback to the previous behaviour, e.g. if `AssocTy` is defined on the trait `Foo`, `Bar` has `Foo` as supertrait and the opaque type is a `impl Bar + Baz`.
- If the the associated type is defined on a generic trait and the opaque type includes two versions of that generic trait, e.g. the opaque type is `impl Foo<A> + Foo<B>`

### DIFF
--- a/compiler/rustc_middle/src/ty/error.rs
+++ b/compiler/rustc_middle/src/ty/error.rs
@@ -628,6 +628,7 @@ impl<T> Trait<T> for X {
                             assoc_substs,
                             ty,
                             msg,
+                            false,
                         ) {
                             return true;
                         }
@@ -646,6 +647,7 @@ impl<T> Trait<T> for X {
                             assoc_substs,
                             ty,
                             msg,
+                            false,
                         );
                     }
                 }
@@ -771,13 +773,24 @@ fn foo(&self) -> Self::T { String::new() }
     ) -> bool {
         let assoc = self.associated_item(proj_ty.item_def_id);
         if let ty::Opaque(def_id, _) = *proj_ty.self_ty().kind() {
-            self.constrain_associated_type_structured_suggestion(
+            let opaque_local_def_id = def_id.expect_local();
+            let opaque_hir_id = self.hir().local_def_id_to_hir_id(opaque_local_def_id);
+            let opaque_hir_ty = match &self.hir().expect_item(opaque_hir_id).kind {
+                hir::ItemKind::OpaqueTy(opaque_hir_ty) => opaque_hir_ty,
+                _ => bug!("The HirId comes from a `ty::Opaque`"),
+            };
+
+            let (trait_ref, assoc_substs) = proj_ty.trait_ref_and_own_substs(self);
+
+            self.constrain_generic_bound_associated_type_structured_suggestion(
                 db,
-                self.def_span(def_id),
-                &assoc,
-                proj_ty.trait_ref_and_own_substs(self).1,
+                &trait_ref,
+                opaque_hir_ty.bounds,
+                assoc,
+                assoc_substs,
                 ty,
-                &msg,
+                msg,
+                true,
             )
         } else {
             false
@@ -899,6 +912,11 @@ fn foo(&self) -> Self::T { String::new() }
 
     /// Given a slice of `hir::GenericBound`s, if any of them corresponds to the `trait_ref`
     /// requirement, provide a structured suggestion to constrain it to a given type `ty`.
+    ///
+    /// `is_bound_surely_present` indicates whether we know the bound we're looking for is
+    /// inside `bounds`. If that's the case then we can consider `bounds` containing only one
+    /// trait bound as the one we're looking for. This can help in cases where the associated
+    /// type is defined on a supertrait of the one present in the bounds.
     fn constrain_generic_bound_associated_type_structured_suggestion(
         self,
         db: &mut DiagnosticBuilder<'_>,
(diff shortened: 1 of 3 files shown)


oli-obk:
Please add tests for the two issues you described

compiler/rustc_middle/src/ty/error.rs
```diff
@@ -771,11 +771,40 @@ fn foo(&self) -> Self::T { String::new() }
...
+                    let trait_ref_def_id = bound
+                        .trait_ref()?
+                        .trait_def_id()
+                        .expect("All `hir::TraitDef`s should already be resolved");
+                    if trait_ref_def_id != assoc_trait_ref.def_id {
+                        return None;
+                    }
+                    Some(bound.span())
+                })
+                .unwrap_or_else(|| self.def_span(def_id));
```
oli-obk, line 801: I think we should not produce any suggestion in this case. We could still emit a help message at the def_span talking about the fact that the bound needs to be added here

SkiFire13: The unfortunate thing about doing that is that we lose suggestions for cases like `impl SubTrait`. I might have had an idea to handle those cases though.

SkiFire13:
> Please add tests for the two issues you described

Should I add tests for them even if they currently produce the wrong output? Or should I make them not suggest anything and test that?

oli-obk:
> Or should I make them not suggest anything and test that?

Either is fine. We often test wrong output and add comments to the test stating that this is wrong and should be fixed.

SkiFire13:
- If the code can't find the right trait to which add the bound it no longer emits a suggestion, instead it just emits the help message on the whole opaque type span
- I've added some logic to handle cases like `impl SubTrait` or `impl SubTrait + 'lt` which weren't covered before but I think are common enough it would be unfortunate to lose the correct suggestion on them.

src/test/ui/associated-types/issue-87261.stderr
```diff
@@ -0,0 +1,138 @@
...
+   |                           --------------- required by this bound in `accepts_trait`
+...
+LL |     accepts_trait(returns_opaque_derived());
+   |     ^^^^^^^^^^^^^ expected `()`, found associated type
+   |
+   = note:    expected unit type `()`
+           found associated type `<impl DerivedTrait as Trait>::Associated`
+help: consider constraining the associated type `<impl DerivedTrait as Trait>::Associated` to `()`
+   |
+LL | fn returns_opaque_derived() -> impl DerivedTrait<Associated = ()> + 'static {
```
oli-obk, line 36: Is that legal? I thought you to add a `+ Trait<Associated = ()>`.

SkiFire13: [Yes](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=f6cf553a7001efddd3e4d99d0426f2ac)

SkiFire13:
Digging around I think I've found a similar issue with generic bounds [playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=0a10d192a407a8a492a5898d02db079e). The existing logic for them is a bit smarter than the one for opaque types, however it still fails when there are multiple generics. I'm tempted to include a fix for that issue in this PR and merge the logic.

src/test/ui/associated-types/issue-87261.stderr
oli-obk: TIL 🤯

oli-obk:
> I'm tempted to include a fix for that issue in this PR and merge the logic.

Whatever you prefer. We can also merge this right now and you can build a new PR on top of it

compiler/rustc_middle/src/ty/error.rs
```diff
@@ -908,23 +921,30 @@ fn foo(&self) -> Self::T { String::new() }
         assoc_substs: &[ty::GenericArg<'tcx>],
         ty: Ty<'tcx>,
         msg: &str,
+        is_bound_surely_present: bool,
```
oli-obk, line 924: Please document this argument's effect. It is clear in the context of this PR, but probably not for anyone just stumbling across it.

oli-obk:
@bors r+

Thanks!

PR #87349, opened 2021-07-21
Prefer suggestion paths which are not doc-hidden 

In-line:
There are cases (as clearly shown in test) when compiler will prefer doc-hidden public reexport of item which is available at alternative path.

```rust
extern crate serde;

fn main() {
    let x: Option<i32> = 1i32; // This suggests `serde::__private::Some(1i32)`, but should suggest `Some(1i32)`
}
```

### DIFF
--- a/compiler/rustc_ast/src/attr/mod.rs
+++ b/compiler/rustc_ast/src/attr/mod.rs
@@ -580,3 +580,12 @@ impl NestedMetaItem {
         MetaItem::from_tokens(tokens).map(NestedMetaItem::MetaItem)
     }
 }
+
+/// Return true if the attributes contain `#[doc(hidden)]`
+pub fn is_doc_hidden(attrs: &[ast::Attribute]) -> bool {
+    attrs
+        .iter()
+        .filter(|attr| attr.has_name(sym::doc))
+        .filter_map(ast::Attribute::meta_item_list)
+        .any(|l| list_contains_name(&l, sym::hidden))
+}
--- a/compiler/rustc_metadata/src/rmeta/decoder/cstore_impl.rs
+++ b/compiler/rustc_metadata/src/rmeta/decoder/cstore_impl.rs
@@ -133,8 +133,8 @@ provide! { <'tcx> tcx, def_id, other, cdata,
     generator_kind => { cdata.generator_kind(def_id.index) }
     opt_def_kind => { Some(cdata.def_kind(def_id.index)) }
     def_span => { cdata.get_span(def_id.index, &tcx.sess) }
-    def_ident_span => {
-        cdata.try_item_ident(def_id.index, &tcx.sess).ok().map(|ident| ident.span)
+    def_ident => {
+        cdata.try_item_ident(def_id.index, &tcx.sess).ok()
     }
     lookup_stability => {
         cdata.get_stability(def_id.index).map(|s| tcx.intern_stability(s))
@@ -290,15 +290,11 @@ pub fn provide(providers: &mut Providers) {
         // external item that is visible from at least one local module) to a
         // sufficiently visible parent (considering modules that re-export the
         // external item to be parents).
-        visible_parent_map: |tcx, ()| {
-            use std::collections::hash_map::Entry;
-            use std::collections::vec_deque::VecDeque;
+        visible_parents_map: |tcx, ()| {
+            use rustc_data_structures::fx::FxHashSet;
+            use std::collections::VecDeque;
 
-            let mut visible_parent_map: DefIdMap<DefId> = Default::default();
-            // This is a secondary visible_parent_map, storing the DefId of parents that re-export
-            // the child as `_`. Since we prefer parents that don't do this, merge this map at the
-            // end, only if we're missing any keys from the former.
-            let mut fallback_map: DefIdMap<DefId> = Default::default();
+            let mut visible_parents_map: DefIdMap<SmallVec<[DefId; 4]>> = DefIdMap::default();
 
             // Issue 46112: We want the map to prefer the shortest
             // paths when reporting the path to an item. Therefore we
@@ -310,59 +306,81 @@ pub fn provide(providers: &mut Providers) {
             // only get paths that are locally minimal with respect to
             // whatever crate we happened to encounter first in this
             // traversal, but not globally minimal across all crates.
-            let bfs_queue = &mut VecDeque::new();
+            let mut bfs_queue = VecDeque::default();
 
-            for &cnum in tcx.crates(()) {
-                // Ignore crates without a corresponding local `extern crate` item.
-                if tcx.missing_extern_crate_item(cnum) {
-                    continue;
-                }
+            bfs_queue.extend(
+                tcx.crates(())
+                    .into_iter()
+                    // Ignore crates without a corresponding local `extern crate` item.
+                    .filter(|cnum| !tcx.missing_extern_crate_item(**cnum))
+                    .map(|cnum| DefId { krate: *cnum, index: CRATE_DEF_INDEX }),
+            );
 
-                bfs_queue.push_back(DefId { krate: cnum, index: CRATE_DEF_INDEX });
-            }
-
-            let mut add_child = |bfs_queue: &mut VecDeque<_>, export: &Export, parent: DefId| {
-                if !export.vis.is_public() {
-                    return;
-                }
-
-                if let Some(child) = export.res.opt_def_id() {
-                    if export.ident.name == kw::Underscore {
-                        fallback_map.insert(child, parent);
-                        return;
-                    }
-
-                    match visible_parent_map.entry(child) {
-                        Entry::Occupied(mut entry) => {
-                            // If `child` is defined in crate `cnum`, ensure
-                            // that it is mapped to a parent in `cnum`.
-                            if child.is_local() && entry.get().is_local() {
-                                entry.insert(parent);
-                            }
-                        }
-                        Entry::Vacant(entry) => {
-                            entry.insert(parent);
+            // Iterate over graph using BFS.
+            // Filter out any non-public items.
+            while let Some(parent) = bfs_queue.pop_front() {
+                for child in tcx
+                    .item_children(parent)
+                    .iter()
+                    .filter(|child| child.vis.is_public())
+                    .filter_map(|child| child.res.opt_def_id())
+                {
+                    visible_parents_map
+                        .entry(child)
+                        .or_insert_with(|| {
+                            // If we encounter node the first time
+                            // add it to queue for next iterations
                             bfs_queue.push_back(child);
+                            Default::default()
+                        })
+                        .push(parent);
+                }
+            }
+
+            // Iterate over parents vector to remove duplicate elements
+            // while preserving order
+            let mut dedup_set = FxHashSet::default();
+            for (_, parents) in &mut visible_parents_map {
+                parents.retain(|parent| dedup_set.insert(*parent));
+
+                // Reuse hashset allocation.
+                dedup_set.clear();
+            }
+
+            visible_parents_map
+        },
+        best_visible_parent: |tcx, child| {
+            // Use `min_by_key` because it returns
+            // first match in case keys are equal
+            tcx.visible_parents_map(())
+                .get(&child)?
+                .into_iter()
+                .min_by_key(|parent| {
+                    // If this is just regular export in another module, assign it a neutral score.
+                    let mut score = 0;
+
+                    // If child and parent are local, we prefer them
+                    if child.is_local() && parent.is_local() {
+                        score += 1;
+                    }
+
+                    // Even if child and parent are local, if parent is `#[doc(hidden)]`
+                    // We reduce their score to avoid showing items not popping in documentation.
+                    if ast::attr::is_doc_hidden(tcx.item_attrs(**parent)) {
+                        score -= 2;
+                    }
+
+                    // If parent identifier is _ we prefer it only as last resort if other items are not available
+                    if let Some(ident) = tcx.def_ident(**parent) {
+                        if ident.name == kw::Underscore {
+                            score -= 3;
                         }
                     }
-                }
-            };
 
-            while let Some(def) = bfs_queue.pop_front() {
-                for child in tcx.item_children(def).iter() {
-                    add_child(bfs_queue, child, def);
-                }
-            }
-
-            // Fill in any missing entries with the (less preferable) path ending in `::_`.
-            // We still use this path in a diagnostic that suggests importing `::*`.
-            for (child, parent) in fallback_map {
-                visible_parent_map.entry(child).or_insert(parent);
-            }
-
-            visible_parent_map
+                    -score
+                })
+                .map(ToOwned::to_owned)
         },
-
         dependency_formats: |tcx, ()| Lrc::new(crate::dependency_format::calculate(tcx)),
         has_global_allocator: |tcx, cnum| {
             assert_eq!(cnum, LOCAL_CRATE);
--- a/compiler/rustc_middle/src/query/mod.rs
+++ b/compiler/rustc_middle/src/query/mod.rs
@@ -977,8 +977,8 @@ rustc_queries! {
     }
 
     /// Gets the span for the identifier of the definition.
-    query def_ident_span(def_id: DefId) -> Option<Span> {
-        desc { |tcx| "looking up span for `{}`'s identifier", tcx.def_path_str(def_id) }
+    query def_ident(def_id: DefId) -> Option<Ident> {
+        desc { |tcx| "looking up ident for `{}`'s identifier", tcx.def_path_str(def_id) }
         separate_provide_extern
     }
 
@@ -1552,9 +1552,13 @@ rustc_queries! {
         desc { "calculating the missing lang items in a crate" }
         separate_provide_extern
     }
-    query visible_parent_map(_: ()) -> DefIdMap<DefId> {
+
+    query visible_parents_map(_: ()) -> DefIdMap<smallvec::SmallVec<[DefId; 4]>> {
         storage(ArenaCacheSelector<'tcx>)
         desc { "calculating the visible parent map" }
+    }
+    query best_visible_parent(child: DefId) -> Option<DefId> {
+        desc { "calculating best visible parent" }
     }
     query trimmed_def_paths(_: ()) -> FxHashMap<DefId, Symbol> {
         storage(ArenaCacheSelector<'tcx>)
--- a/compiler/rustc_middle/src/ty/print/pretty.rs
+++ b/compiler/rustc_middle/src/ty/print/pretty.rs
@@ -387,8 +387,6 @@ pub trait PrettyPrinter<'tcx>:
             return Ok((self, false));
         }
 
-        let visible_parent_map = self.tcx().visible_parent_map(());
-
         let mut cur_def_key = self.tcx().def_key(def_id);
         debug!("try_print_visible_def_path: cur_def_key={:?}", cur_def_key);
 
@@ -404,7 +402,7 @@ pub trait PrettyPrinter<'tcx>:
             cur_def_key = self.tcx().def_key(parent);
         }
 
-        let visible_parent = match visible_parent_map.get(&def_id).cloned() {
+        let visible_parent = match self.tcx().best_visible_parent(def_id) {
             Some(parent) => parent,
             None => return Ok((self, false)),
         };
@@ -431,7 +429,7 @@ pub trait PrettyPrinter<'tcx>:
             // `std::os::unix` rexports the contents of `std::sys::unix::ext`. `std::sys` is
             // private so the "true" path to `CommandExt` isn't accessible.
             //
-            // In this case, the `visible_parent_map` will look something like this:
+            // In this case, the `visible_parents_map` will look something like this:
             //
             // (child) -> (parent)
             // `std::sys::unix::ext::process::CommandExt` -> `std::sys::unix::ext::process`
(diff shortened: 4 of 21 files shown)


In-line:
@Mark-Simulacrum can you or someone else familiar with the code provide mentorship for this issue?

I tried to remove deduplication from `rustc_typeck/src/check/demand.rs`, but I'm stack.

In-line:
If i'm understanding correctly typecheck starts here 

https://github.com/rust-lang/rust/blob/cf932aa584f2ba1a1744f3ef0b21f14387276fb6/compiler/rustc_typeck/src/check/expr.rs#L852-L861

More precisely in this part of the function body:
https://github.com/rust-lang/rust/blob/cf932aa584f2ba1a1744f3ef0b21f14387276fb6/compiler/rustc_typeck/src/check/expr.rs#L921

After that it goes into `self.demand_coerce` which forwards it to `self.demand_coerce_diag`
https://github.com/rust-lang/rust/blob/cf932aa584f2ba1a1744f3ef0b21f14387276fb6/compiler/rustc_typeck/src/check/expr.rs#L101-L110

In the body of which it emits errors
https://github.com/rust-lang/rust/blob/cf932aa584f2ba1a1744f3ef0b21f14387276fb6/compiler/rustc_typeck/src/check/demand.rs#L154

In-line:
So actual suggestion is calculated in this line
https://github.com/rust-lang/rust/blob/cf932aa584f2ba1a1744f3ef0b21f14387276fb6/compiler/rustc_typeck/src/check/demand.rs#L138-L141

Which goes here
https://github.com/rust-lang/rust/blob/cf932aa584f2ba1a1744f3ef0b21f14387276fb6/compiler/rustc_typeck/src/check/coercion.rs#L148-L150
And after that here 
https://github.com/rust-lang/rust/blob/cf932aa584f2ba1a1744f3ef0b21f14387276fb6/compiler/rustc_typeck/src/check/coercion.rs#L232-L235
And goes here
https://github.com/rust-lang/rust/blob/cf932aa584f2ba1a1744f3ef0b21f14387276fb6/compiler/rustc_typeck/src/check/coercion.rs#L132-L134

src/test/ui/suggestions/dont-suggest-doc-hidden-variant-for-enum/dont-suggest-doc-hidden-variant-for-enum.rs
```diff
@@ -0,0 +1,7 @@
+// aux-build:serde.rs
+
+extern crate serde;
```
hellow554, line 3: I don't think that serde is a good name for the external crate here. It would lead to confusion with the actual crate serde. 

Also I'm not entirely sure what the point of this test is. There is no difference in your output than to the current output.

Something like this is more what I would expect of your code:

hidden_struct.rs
```rust
#[doc(hidden)]
pub struct Foo;
```

main.rs
```rust
fn main() {
    let _ = FOo;
}
```
output is:

```
error[E0425]: cannot find value `FOo` in this scope
 --> src/main.rs:5:13
  |
2 | pub struct Foo;
  | --------------- similarly named unit struct `Foo` defined here
...
5 |     let _ = FOo;
  |             ^^^ help: a unit struct with a similar name exists (notice the capitalization): `Foo`
```

In-line: Try run test locally, after  #87158 it suggests `serde::__private::Some`

hellow554: > Try to run test locally, after #87158 it suggests serde::__private::Some.

My bad. I was testing stable.

> I don't see how your test case is related to issue.

There is a suggestion for a hidden struct/item. Isn't that what you are trying to suppress?

In-line: @fellow554 if there is doc-hidden reexport of struct enum variant, I want suggestion to prefer path which doesn't lead to doc-hidden module. Like in test case.

I want the test to pass. I will change the name of `serde` to something else, but originally this is just minimal reproduction of issue I got with serde. They have "private" reexports.

I'm not sure where is `expected` expr is selected, to make compiler prefer paths which are not doc-hidden, I need some guidance.

Thanks!

In-line: @hellow554 P.S. using local module doesn't work it should be external crate.

In-line: I already found that root cause is actually here
https://github.com/rust-lang/rust/blob/8024983ea73e1d1c16720a424d39fd1c239eee68/compiler/rustc_middle/src/ty/print/pretty.rs#L325-L358

compiler/rustc_metadata/src/rmeta/decoder/cstore_impl.rs
```diff
@@ -304,6 +304,12 @@ pub fn provide(providers: &mut Providers) {
             // traversal, but not globally minimal across all crates.
             let bfs_queue = &mut VecDeque::new();
 
+            // Some sub-trees of crate graph can be deferred to prioritize others
+            // This helps for example for decreasing priority of `#[doc(hidden)]` items,
+            // because they should be included as in final map,
+            // only when no other alternative paths are available
+            let deferred_items_queue = &mut VecDeque::new();
```
In-line, line 311: I'm not sure if this approach is most preferable algorithmically.

In-line:
Quite interesting results, but they are not suprising because we actually do more work (calling `tcx.item_attrs`) query. I tried to mitigate some of the constant factor with switching to `Vec` and reusing allocation and converting `is_doc_hidden` to query.

In-line:
@estebank can you please do another perf run?

hellow554:
@In-line you can always run tests locally by using `./x.py test  src/test/ui/proc-macro/issue-37788.rs` so you don't have to rely on bors and make things easier ;)

In-line:
I just forget sometimes :D

-------- Original Message --------

In-line:
@hellow554 will you prefer this PR to be single commit?

hellow554:
No, but things like https://github.com/rust-lang/rust/pull/87349/commits/9380d2070abb96f0d1927b9e88bf48ce296d3cb7 https://github.com/rust-lang/rust/pull/87349/commits/119967446c2495d8e97030f6f474e3b304def84b https://github.com/rust-lang/rust/pull/87349/commits/745af12621da0191bc7601ecda36b06f5a6221e9 https://github.com/rust-lang/rust/pull/87349/commits/291cc4b6e2458e09628ed270c7c5b94488924fe1 shouldn't be extra commits.
IMHO it's okay for reviewing purposes, but before merging these should be rebases.

This is my opinion, not an official one nor am I part of the official team so in the end it's up to you and the official reviewer to decide what's right and wrong ;)

In-line:
@hellow554 Okay! Waiting for perf results. I will rebase this, after regression is fixed.

In-line:
Is bors dead?

hellow554:
Not dead, but busy. A New release is beeing prepared and has higher priority

lqd:
Not for try builds though.

@bors try @rust-timer queue

lqd:
Ah I think there was a new commit pushed after the previous try build was requested, and this cancels in-progress try builds for the PR IIRC.

lqd:
BTW @In-line, to help the people who'll be reviewing this PR in the future, it would be super cool if you could update the PR description with some short context of what you're trying to achieve (or if it fixes a known github issue). 

Maybe a little example and what the compiler currently outputs and why that's bad, and what would the result be with this PR. The tests you've added would be good for this, for example.

Maybe some explanation of why the `itertools` dependency was required: it could be that some code to achieve the same thing already exists elsewhere, and maybe this could be avoided.

Thanks a bunch in advance!

In-line:
Please, schedule one more run for rust-timer.

In my local tests this version outperforms `master` branch by roughly ~10%. I expect improvement over the base now.

In-line:
Hm, I think this are pretty good results. It seems main slowdown is now because we are doing additional work, mainly calling `is_doc_hidden`. And regressions are now only in toy (small) examples.

Is this regression acceptable? Should I try to improve it further?

In-line:
await-call-tree non-incremental fresh debug full-build `visible_parent(s)_map` query time.

It seems using `SmallVec` is more efficient in this case.

| Version | Time |
| - | - |
| nightly | 917.07µs |
| stage1 (with SmallVec) |  1.170942ms |
| stage1 (without SmallVec) | 1.401424ms |

hellow554:
You really should have a pre-commit hook that tests your code locally 😛
Read https://rustc-dev-guide.rust-lang.org/building/suggested.html for what I mean

In-line:
@hellow554 I just have spurious failures in some tests (even on master branch), they make it really hard to build & test everything. I will fix them with another PR.

In-line:
Current results for await-call-tree non-incremental fresh debug full-build

| Version | Time |
| - | - |
| nightly | 928.782µs |
| stage1 |  1.065356ms |

hellow554:
Hey @In-line 

do you need any help or do you just don't have time ATM?

In-line:
If you have suggestions how to improve performance further I would be happy to hear them.

Thank you.

I kinda ran out of ideas now, preallocating with `HasMap` capacity didn't help.
-------- Original Message --------

compiler/rustc_metadata/src/rmeta/decoder/cstore_impl.rs
```diff
@@ -311,54 +312,74 @@ pub fn provide(providers: &mut Providers) {
...
-            for &cnum in crates.iter() {
-                // Ignore crates without a corresponding local `extern crate` item.
-                if tcx.missing_extern_crate_item(cnum) {
-                    continue;
+            bfs_queue.extend(
+                tcx.crates(())
+                    .into_iter()
+                    // Ignore crates without a corresponding local `extern crate` item.
+                    .filter(|cnum| !tcx.missing_extern_crate_item(**cnum))
+                    .sorted()
```
hellow554, line 320: Does this really need to be `sorted`?

compiler/rustc_middle/src/ty/print/pretty.rs
```diff
@@ -378,7 +376,7 @@ pub trait PrettyPrinter<'tcx>:
             cur_def_key = self.tcx().def_key(parent);
         }
 
-        let visible_parent = match visible_parent_map.get(&def_id).cloned() {
+        let visible_parent = match self.tcx().best_visible_parent(def_id).clone() {
```
hellow554, line 379: you don't need the clone here

hellow554:
I have no clue on what to do more. But IMHO you did very good on the performance. After all the compiler must do more work than before, so a regression is feasible.

What do you think @estebank . Is this PR in a good shape or do you think that the regression is not worth it?

compiler/rustc_metadata/src/rmeta/decoder/cstore_impl.rs
In-line: @hellow554 this was sorted in previous code for determinism.

compiler/rustc_metadata/src/rmeta/decoder/cstore_impl.rs
```diff
@@ -311,54 +312,74 @@ pub fn provide(providers: &mut Providers) {
...
+                    let mut score = 0;
+
+                    // If child and parent are local, we prefer them
+                    if child.is_local() && parent.is_local() {
+                        score += 1;
+                    }
+
+                    // Even if child and parent are local, if parent is `#[doc(hidden)]`
+                    // We should prefer it only as last resolve
+                    // If other items are not available
```
jyn514, line 374:
```suggestion
                    // we should prefer it only as last resort if other items are not available
```

compiler/rustc_metadata/src/rmeta/decoder/cstore_impl.rs
```diff
@@ -311,54 +312,74 @@ pub fn provide(providers: &mut Providers) {
...
+                    if tcx.is_doc_hidden(**parent) {
+                        score -= 2;
+                    }
 
+                    -score
```
jyn514, line 379: You can avoid negating this by using `max_by_key` instead.

compiler/rustc_middle/src/query/mod.rs
```diff
@@ -965,6 +965,11 @@ rustc_queries! {
         desc { |tcx| "collecting attributes of `{}`", tcx.def_path_str(def_id) }
     }
 
+    /// Checks if item has `#[doc(hidden)]` attribute.
+    query is_doc_hidden(def_id: DefId) -> bool {
+        desc { |tcx| "checking if item is doc-hidden `{}`", tcx.def_path_str(def_id) }
+    }
```
jyn514, line 971: Consider *not* adding a query for this, and using a normal function instead. That will reduce the overhead of caching the result, since `is_doc_hidden` is very cheap to calculate.

jyn514:
Most of the performance regression comes from calculating visible_parents_map, I would look there to see what's changed. it's hard for me to tell as a reviewer because the diff is so big - consider reverting back to a normal loop instead of rewriting it with iterators.

estebank:
> Is this PR in a good shape or do you think that the regression is not worth it?

@hellow554 I am concerned with the perf regression. 1.5% sounds small, but it can be quite significant :-(

If we can bring that down then I would have less concerns landing this.

compiler/rustc_metadata/src/rmeta/decoder/cstore_impl.rs
In-line: @jyn514 this will select last item, instead of first

In-line: https://doc.rust-lang.org/stable/core/iter/trait.Iterator.html#method.max_by_key
> If several elements are equally maximum, the **last** element is returned. If the iterator is empty, None is returned.

jyn514: Ok - that's not evident just from the code, can you leave a comment?

Also, why does min_by_key behave differently? I don't understand why you would prefer the first over the last or vice versa.

In-line:
Related #89120

compiler/rustc_metadata/src/rmeta/decoder/cstore_impl.rs
In-line: @jyn514 Preferring first was just what algorithm did before this PR.
Preferring last will produce slightly results and I think will prevent early exit from loop.

For example replacing `min_by_key` by `max_by_key` will produce following failure in one of unit tests.

```rust
---- [ui] ui/trait-impl-bound-suggestions.rs stdout ----
diff of stderr:

11         |                             ^^^^ required by this bound in `ConstrainedStruct`
12      help: consider further restricting type parameter `X`
13         |
-       LL | trait InsufficientlyConstrainedGeneric<X=()> where X: std::marker::Copy {
-          |                                              ++++++++++++++++++++++++++
+       LL | trait InsufficientlyConstrainedGeneric<X=()> where X: std::prelude::rust_2021::Copy {
+          |                                              ++++++++++++++++++++++++++++++++++++++
16
17      error: aborting due to previous error
18

```

How can I word the comment better to make it more understandable?

In-line:
Please, kindly schedule perfr un for this PR too. I removed `is_doc_hidden` query. 
Thanks!

compiler/rustc_metadata/src/rmeta/decoder/cstore_impl.rs
In-line: I found comment in `visible_parents_map` explaining the order.

https://github.com/rust-lang/rust/blob/2f9c8535bfa70159ccb431174f881b4aa47dc7d7/compiler/rustc_metadata/src/rmeta/decoder/cstore_impl.rs#L296-L299

In-line:
Regressions are only on very small examples and I think if we land #89120 before this, we will have neutral results.

In-line:
Please kindly redo a perf-run, now that #89120 is merged.

compiler/rustc_ast/src/attr/mod.rs
```diff
@@ -580,3 +580,12 @@ impl NestedMetaItem {
...
 }
+
+/// Return true if the attributes contain `#[doc(hidden)]`
+pub fn is_doc_hidden(attrs: &[ast::Attribute]) -> bool {
+    attrs
+        .iter()
+        .filter(|attr| attr.has_name(sym::doc))
+        .filter_map(ast::Attribute::meta_item_list)
+        .any(|l| list_contains_name(&l, sym::hidden))
+}
```
estebank, line 591: Could we improve performance if we modified the Ast to explicitly track this? We could still have it part of the attribute list, but have a boolean as part of the item or something so that it gets computed during parsing once, and not every time we compute an item when building the visible_parent_map. I think that might win us some of the performance we're losing back (at the cost of memory).

jyn514: @estebank it was a query before and that seemed to hurt performance more than it helped

estebank: But queries only help when you call them repeatedly, calling it just once you get the overhead of the query system _and_ still have the O(n) behavior, right? Moving that to be something that happens during parsing removes the O(n) behavior entirely for something that gets evaluated for (I'm guessing) every path, unless I'm missing something?

jyn514: That seems possible? This is only a .6% regression on some small benchmarks though.

Mark-Simulacrum: The wall-time regressions for self-bootstrap are more concerning to me. It's not clear whether those come from newly introduced code, but in combination with the [cycle regressions](https://perf.rust-lang.org/compare.html?calcNewSig=true&start=91d8da1f4ba24679e92b7939a26c681a5d2d3548&end=6a304b8fb7cdacaa12996b2b46a8f61e8a2f1ba7&stat=cycles%3Au) I would not call this "just" .6%.

Do we have any evidence that the regression here is due to doc_hidden being slow? Unfortunately cycles make it a little difficult to do good local comparisons, as cachegrind only really looks at instructions (results locally from an attempt are inconclusive). I can play with some perf-record later, perhaps.

In-line: Rustc has supressed deprecation warnings for `llvm_asm`. Currently even if warning is supressed, message is calculated.

I think this is where regression for bootstrap time comes.

https://github.com/rust-lang/rust/blob/bb744e1e9f1e193014ab3f04bf5e0640dbe4dc63/compiler/rustc_middle/src/middle/stability.rs#L307

jyn514: That should be easy enough to fix - just add a check to `lint_level_at_node`.

In-line:
Now that #89395 is merged, please kindly redo perf run

In-line:
Can't we merge this if this has neutral results?

jyn514:
@In-line I wasn't sure if @estebank wanted to review it a final time.

compiler/rustc_middle/src/query/mod.rs
```diff
@@ -1462,10 +1462,13 @@ rustc_queries! {
     query missing_lang_items(_: CrateNum) -> &'tcx [LangItem] {
         desc { "calculating the missing lang items in a crate" }
     }
-    query visible_parent_map(_: ()) -> DefIdMap<DefId> {
-        storage(ArenaCacheSelector<'tcx>)
```
In-line, line 1466: I'm not sure what `ArenaCacheSelector` does here, should I use it instead of `Lrc<_>`?

jyn514: There was a performance hit before https://github.com/rust-lang/rust/pull/89395 and https://github.com/rust-lang/rust/pull/89120 were merged. The arena change may no longer be necessary.

JohnCSimon:
Ping from triage:
@In-line What should happen with this PR?

jyn514:
Hi @In-line , could you revert the changes to the arena allocation? This looks good to go after that :)

JohnCSimon:
Ping from triage: 
@In-line can you please address the comments from jyn514

In-line:
Hi! Sorry for the delay. I reverted change with `ArenaCacheSelector`. Can you kindly schedule perf run? So we can finally merge this :shipit:

In-line:
Friendly ping @Mark-Simulacrum @JohnCSimon

compiler/rustc_middle/src/query/mod.rs
In-line: @Mark-Simulacrum  @jyn514 Now that benchmark results show small regression, should I revert this back (to remove `ArenaCacheSelector`)?

Mark-Simulacrum: That should still be done in a separate PR, so we can independently evaluate its effects.

In-line: @Mark-Simulacrum So, is this discussion resolved then?

Mark-Simulacrum: Well, I'd expect that separate PR to likely get filed before this merges (in order to attempt to prevent the regression here).

In-line: @Mark-Simulacrum I opened #92169 #92168 for that

In-line: and #92170

Mark-Simulacrum:
r+'d https://github.com/rust-lang/rust/pull/92169, once that merges we can rebase this and do another perf run.

JohnCSimon:
Ping from triage: 
@In-line:

> [Mark-Simulacrum](https://github.com/Mark-Simulacrum) commented [26 days ago](https://github.com/rust-lang/rust/pull/87349#issuecomment-1009126446)
> r+'d #92169, once that merges we can rebase this and do another perf run.



`@rustbot ready` to switch back to `S-waiting-on-review`

Dylan-DPC:
@In-line if you can rebase this, we can push this forward

JohnCSimon:
@In-line 

@rustbot label: +S-inactive

PR #87357, opened 2021-07-21
Update my name/email in .mailmap

lilyball:
I should have done this years ago.

My name is also scattered across many other rust-lang repos, but none of the other repos I've looked at have a `.mailmap`. Is there any particular reason for this, or any process for updating my name/email everywhere?


jyn514:
@bors r+ rollup

> My name is also scattered across many other rust-lang repos, but none of the other repos I've looked at have a .mailmap. Is there any particular reason for this, or any process for updating my name/email everywhere?

I don't know if there's a way to do it consistently, but FWIW I don't see anything wrong with adding a mailmap to any repo that doesn't have one already.

PR #84450, opened 2021-04-22
Give a better error when `std` or `core` are missing

jyn514:
- Suggest using `rustup target add` if `RUSTUP_HOME` is set. I don't know if there's any precedent for doing this, but it seems harmless enough and it will be a big help.
- On nightly, suggest using `cargo build -Z build-std` if `CARGO` is set
- Add a note about `#![no_std]` if `std` is missing but not core
- Add a note that std may be unsupported if `std` is missing but not core

Fixes https://github.com/rust-lang/rust/issues/84418.

r? @petrochenkov

PR #87359, opened 2021-07-22
Remove detection of rustup and cargo in 'missing extern crate' diagnostics

jyn514:
Previously, this would change the test output when RUSTUP_HOME was set:

```
---- [ui] ui/issues/issue-49851/compiler-builtins-error.rs stdout ----
diff of stderr:

1       error[E0463]: can't find crate for `core`
2          |
3          = note: the `thumbv7em-none-eabihf` target may not be installed
+          = help: consider downloading the target with `rustup target add thumbv7em-none-eabihf`
4
5       error: aborting due to previous error
6
```

Originally, I fixed it by explicitly unsetting RUSTUP_HOME in
compiletest. Then I realized that almost no one has RUSTUP_HOME set,
since rustup doesn't set it itself. It does set RUST_RECURSION_COUNT
whenever it launches a proxy, though - use that instead.

r? @estebank cc @petrochenkov @kinnison

### DIFF
--- a/compiler/rustc_metadata/src/locator.rs
+++ b/compiler/rustc_metadata/src/locator.rs
@@ -1080,7 +1080,10 @@ impl CrateError {
                                 locator.triple
                             ));
                         }
-                        if missing_core && std::env::var("RUSTUP_HOME").is_ok() {
+                        // NOTE: this suggests using rustup, even though the user may not have it installed.
+                        // That's because they could choose to install it; or this may give them a hint which
+                        // target they need to install from their distro.
+                        if missing_core {
                             err.help(&format!(
                                 "consider downloading the target with `rustup target add {}`",
                                 locator.triple
@@ -1097,7 +1100,7 @@ impl CrateError {
                                 current_crate
                             ));
                         }
-                        if sess.is_nightly_build() && std::env::var("CARGO").is_ok() {
+                        if sess.is_nightly_build() {
                             err.help("consider building the standard library from source with `cargo build -Zbuild-std`");
                         }
                     } else if Some(crate_name)
--- a/src/test/ui/crate-loading/missing-std.rs
+++ b/src/test/ui/crate-loading/missing-std.rs
@@ -1,7 +1,6 @@
 // compile-flags: --target x86_64-unknown-uefi
 // needs-llvm-components: x86
 // rustc-env:CARGO=/usr/bin/cargo
-// rustc-env:RUSTUP_HOME=/home/bors/.rustup
 #![no_core]
 extern crate core;
 //~^ ERROR can't find crate for `core`
--- a/src/test/ui/crate-loading/missing-std.stderr
+++ b/src/test/ui/crate-loading/missing-std.stderr
@@ -1,5 +1,5 @@
 error[E0463]: can't find crate for `core`
-  --> $DIR/missing-std.rs:6:1
+  --> $DIR/missing-std.rs:5:1
    |
 LL | extern crate core;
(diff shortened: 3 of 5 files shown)


crlf0710:
I have `RUSTUP_HOME` set to move all rust related things to a directory under D:\ disk. I didn't really realize the test suite difference is because of this... (and didn't investigate) Thank you for fixing it!

compiler/rustc_metadata/src/locator.rs
```diff
@@ -1080,7 +1080,7 @@ impl CrateError {
                                 locator.triple
                             ));
                         }
-                        if missing_core && std::env::var("RUSTUP_HOME").is_ok() {
+                        if missing_core && std::env::var("RUST_RECURSION_COUNT").is_ok() {
                             err.help(&format!(
                                 "consider downloading the target with `rustup target add {}`",
                                 locator.triple
```
estebank, line 1086: An option would be to use the env var only to be more specific on the `help`:

```suggestion
                        if missing_core {
                            err.help(&format!(
                                "{}consider downloading the target with `rustup target add {}`",
                                if std::env::var("RUST_RECURSION_COUNT").is_ok() {
                                    "if you have installed `rustc` with `rustup`, 
                                } else {
                                    ""
                                },
                                locator.triple
```

An even better check would be to use [`which`](https://crates.io/crates/which) (or similar) to see if `rustup` is available at all before giving the suggestion (but then we could get in the weird situation where `rustc` was installed by, let's say `apt`, but `rustup` is available and the suggestion would be wrong).

estebank:
r=me if noone objects to this approach

Mark-Simulacrum:
It looks like this detection was added in #84450 without any discussion that I can see on the appropriateness of doing this sort of runtime detection to vary error output.

I am curious about the motivation to only expose this if rustup is installed. In practice, even if the user *did* install with apt or some other tool, the pointer to rustup target install may still be quite helpful (perhaps they should switch to rustup for example). And if they're *not* running with rustup and don't want to be, the diagnostic is not really misleading IMO -- it may still be helpful to let them find the appropriate target-specific package in their distro's toolchain for example.

So I'd suggest just dropping the env variable check entirely here, rather than trying to change it to a different one. It seems annoying that rerunning tests with and without rustup wrappers will produce potentially different results; it's not something I at least would expect.

jyn514:
> I am curious about the motivation to only expose this if rustup is installed. In practice, even if the user did install with apt or some other tool, the pointer to rustup target install may still be quite helpful (perhaps they should switch to rustup for example). And if they're not running with rustup and don't want to be, the diagnostic is not really misleading IMO -- it may still be helpful to let them find the appropriate target-specific package in their distro's toolchain for example.

So I'd suggest just dropping the env variable check entirely here, rather than trying to change it to a different one.

Hmm, ok. That makes sense to me.

> It seems annoying that rerunning tests with and without rustup wrappers will produce potentially different results; it's not something I at least would expect.

Well, the idea is that you would never run tests with rustup wrappers, since x.py controls the toolchain you use and it always goes through build/stageN. But I agree that saying this unconditionally is probably a better approach.

jyn514:
By that logic, do you want me to also remove the runtime checking for cargo? https://github.com/rust-lang/rust/blob/32373c40a9781cbbd7ed292a299b28258b2910a8/compiler/rustc_metadata/src/locator.rs#L1100-L1102

Mark-Simulacrum:
I think it's a good idea, yes, for similar reasons.

PR #87372, opened 2021-07-22
Move calls to test_main into one function

### LINKED ISSUE #86254: Deduplicate doc tests logic
This comes from a [discussion](https://github.com/rust-lang/rust/pull/86230/files#r650439670) in #86230: for the moment, we call `libtest` from two places: `markdown.rs` (which is used when we provide a markdown file as input) and `doctest.rs` (when a rust file is provided as input, the default).

Both calls can be found when grepping with `testing::test_main(`.

GuillaumeGomez:
Fixes #86254.

cc @jyn514 
r? @camelid

### DIFF
--- a/src/librustdoc/doctest.rs
+++ b/src/librustdoc/doctest.rs
@@ -105,7 +105,7 @@ crate fn run(options: Options) -> Result<(), ErrorReported> {
         registry: rustc_driver::diagnostics_registry(),
     };
 
-    let mut test_args = options.test_args.clone();
+    let test_args = options.test_args.clone();
     let display_warnings = options.display_warnings;
     let nocapture = options.nocapture;
     let externs = options.externs.clone();
@@ -166,12 +166,7 @@ crate fn run(options: Options) -> Result<(), ErrorReported> {
         Err(ErrorReported) => return Err(ErrorReported),
     };
 
-    test_args.insert(0, "rustdoctest".to_string());
-    if nocapture {
-        test_args.push("--nocapture".to_string());
-    }
-
-    test::test_main(&test_args, tests, Some(test::Options::new().display_output(display_warnings)));
+    run_tests(test_args, nocapture, display_warnings, tests);
 
     // Collect and warn about unused externs, but only if we've gotten
     // reports for each doctest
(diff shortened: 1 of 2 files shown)


jyn514:
@bors r+ rollup

@GuillaumeGomez is there a reason you cc me on all these PRs?

GuillaumeGomez:
On this one it's simply because I know you like to keep an eye on rustdoc "backend" changes. But I can stop if the notifications annoy you, as you prefer. :)

jyn514:
ahh, that makes sense :) no worries, the pings are fine

camelid:
Thank you, this is definitely an improvement! It still seems surprising to me that there are two separate doctest-running functions (`markdown::test` and `doctest::run`), with significantly different contents. Why are they different, and should this PR still be marked as closing #86254?

jyn514:
@camelid they're different entry points: doctest is used by `rustdoc --test` and markdown is used as a library by error_index_generator. I want to get rid of error index's dependence on rustdoc at some point so it should go away when that happens.

camelid:
> @camelid they're different entry points: doctest is used by `rustdoc --test` and markdown is used as a library by error_index_generator. I want to get rid of error index's dependence on rustdoc at some point so it should go away when that happens.

Hmm, Guillaume mentioned previously that `markdown` is also used by `rustdoc foo.md` (i.e., calling rustdoc on a Markdown file instead of a Rust file), so I'm not sure if we'd be able to get rid of it.

Mark-Simulacrum:
Both of them are invoked from main rustdoc (rustdoc supports `rustdoc --test foo.md` on the CLI); it seems like the difference is whether the input source text is Rust code (and so needs to be parsed & expanded to get at the doc tests) or Markdown (so needs to be parsed, but in a completely different way).

I don't think either could go away, but it does seem a little suspicious that this PR is not unifying the (seemingly overlooked?) support for unused externs in markdown tests. I at least would expect that to work just as well regardless of the input type...

GuillaumeGomez:
Just like @camelid said: one is run from rust code (we iterate through hir items and then check into their documentation if they have tests) whereas the second directly check from a markdown file (which is used by the error index as @jyn514 said). So we need to keep both, the goal here is to simply to merge the parts which can be merged.

GuillaumeGomez:
> Why are they different, and should this PR still be marked as closing #86254?

And yes, this is the part of the code which I referred to when I opened this issue. ;)

PR #87375, opened 2021-07-22
Try filtering out non-const impls when we expect const impls

fee1-dead:
**TL;DR**: Associated types on const impls are now bounded; we now disallow calling a const function with bounds when the specified type param only has a non-const impl.

r? @oli-obk


compiler/rustc_mir/src/transform/check_consts/validation.rs
```diff
@@ -822,7 +822,10 @@ impl Visitor<'tcx> for Validator<'mir, 'tcx> {
                     let obligation = Obligation::new(
                         ObligationCause::dummy(),
                         param_env,
-                        Binder::dummy(TraitPredicate { trait_ref }),
+                        Binder::dummy(TraitPredicate {
+                            trait_ref,
+                            constness: hir::Constness::Const,
```
oli-obk, line 827: This should be `NotConst` if `ccx.const_context.is_none()`

compiler/rustc_trait_selection/src/traits/fulfill.rs
```diff
@@ -352,7 +352,7 @@ impl<'a, 'b, 'tcx> FulfillProcessor<'a, 'b, 'tcx> {
                 // Evaluation will discard candidates using the leak check.
                 // This means we need to pass it the bound version of our
                 // predicate.
-                ty::PredicateKind::Trait(trait_ref, _constness) => {
+                ty::PredicateKind::Trait(trait_ref) => {
                     let trait_obligation = obligation.with(binder.rebind(trait_ref));
```
oli-obk, line 356: This is the main change of this PR, did the other changes affect tests as well, or is it just this change?

compiler/rustc_mir/src/transform/check_consts/validation.rs
fee1-dead: It is not possible for the const_context to be None at this point:
```rs
fn mir_const_qualif(tcx: TyCtxt<'_>, def: ty::WithOptConstParam<LocalDefId>) -> ConstQualifs {
    let const_kind = tcx.hir().body_const_context(def.did);

    // No need to const-check a non-const `fn`.
    if const_kind.is_none() {
        return Default::default();
    }
```

compiler/rustc_trait_selection/src/traits/fulfill.rs
fee1-dead: No, this is the main change:
https://github.com/rust-lang/rust/blob/8c2ed7012383ecca11d40c621604e8d30c329873/compiler/rustc_trait_selection/src/traits/select/mod.rs#L1025-L1038

compiler/rustc_mir/src/borrow_check/type_check/mod.rs
```diff
@@ -2742,10 +2742,10 @@ impl<'a, 'tcx> TypeChecker<'a, 'tcx> {
         category: ConstraintCategory,
     ) {
         self.prove_predicates(
-            Some(ty::PredicateKind::Trait(
-                ty::TraitPredicate { trait_ref },
-                hir::Constness::NotConst,
-            )),
+            Some(ty::PredicateKind::Trait(ty::TraitPredicate {
+                trait_ref,
+                constness: hir::Constness::NotConst,
```
oli-obk, line 2747: we could probably use the const context of the item currently being type checked to decide on `Const` here.

compiler/rustc_trait_selection/src/traits/auto_trait.rs
```diff
@@ -285,6 +285,7 @@ impl AutoTraitFinder<'tcx> {
                 def_id: trait_did,
                 substs: infcx.tcx.mk_substs_trait(ty, &[]),
             },
+            constness: hir::Constness::NotConst,
```
oli-obk, line 288: similarly here, do we want `Const` if we are in a const context?

compiler/rustc_trait_selection/src/traits/auto_trait.rs
```diff
@@ -344,10 +345,7 @@ impl AutoTraitFinder<'tcx> {
                 Err(SelectionError::Unimplemented) => {
                     if self.is_param_no_infer(pred.skip_binder().trait_ref.substs) {
                         already_visited.remove(&pred);
-                        self.add_user_pred(
-                            &mut user_computed_preds,
-                            pred.without_const().to_predicate(self.tcx),
-                        );
+                        self.add_user_pred(&mut user_computed_preds, pred.to_predicate(self.tcx));
```
oli-obk, line 348: it's a bit subtle that `to_predicate` uses `NotConst` inside

compiler/rustc_trait_selection/src/traits/select/mod.rs
```diff
@@ -1016,13 +1015,31 @@ impl<'cx, 'tcx> SelectionContext<'cx, 'tcx> {
...
 
-    // Treat negative impls as unimplemented, and reservation impls as ambiguity.
-    fn filter_negative_and_reservation_impls(
+    fn filter_impls(
         &mut self,
         candidate: SelectionCandidate<'tcx>,
+        obligation: &TraitObligation<'tcx>,
     ) -> SelectionResult<'tcx, SelectionCandidate<'tcx>> {
+        let tcx = self.tcx();
+        // Respect const trait obligations
```
oli-obk, line 1024: does `self` give us enough information to check if we are in a const context?

oli-obk: If not, we could start tracking it.

compiler/rustc_traits/src/chalk/lowering.rs
```diff
@@ -702,7 +702,7 @@ impl<'tcx> LowerInto<'tcx, Option<chalk_solve::rust_ir::QuantifiedInlineBound<Ru
         let (predicate, binders, _named_regions) =
             collect_bound_vars(interner, interner.tcx, self.kind());
         match predicate {
-            ty::PredicateKind::Trait(predicate, _) => Some(chalk_ir::Binders::new(
+            ty::PredicateKind::Trait(predicate) => Some(chalk_ir::Binders::new(
                 binders,
                 chalk_solve::rust_ir::InlineBound::TraitBound(
                     predicate.trait_ref.lower_into(interner),
```
oli-obk, line 708: I think we need to start tracking this information in chalk, too

compiler/rustc_mir/src/transform/check_consts/validation.rs
oli-obk: oh heh. I was looking for that but couldn't find it.

compiler/rustc_trait_selection/src/traits/fulfill.rs
oli-obk: hmm... I would expect this change to have some effect, too.

oli-obk:
only got half way through, will look more tomorrow

compiler/rustc_mir/src/borrow_check/type_check/mod.rs
fee1-dead: This is only used to prove that a lang item is implemented e.g. `CoerceUnsized`, `Sized`, and `Copy`

compiler/rustc_typeck/src/check/compare_method.rs
```diff
@@ -1292,7 +1292,24 @@ pub fn check_type_bounds<'tcx>(
     };
 
     tcx.infer_ctxt().enter(move |infcx| {
-        let inh = Inherited::new(infcx, impl_ty.def_id.expect_local());
+        // if the item is inside a const impl, we transform the predicates to be const.
```
oli-obk, line 1295: This will also transform `?const Trait` bounds to be const, right?

fee1-dead:
Aaaah why do we have these cursed tests of `const` + `async`

compiler/rustc_trait_selection/src/traits/auto_trait.rs
fee1-dead: No it doesn't. This is converting `Binder<TraitPredicate>` to a `Predicate`. As we have a constness field in `TraitPredicate`, it does not use `NotConst` inside.

compiler/rustc_trait_selection/src/traits/select/mod.rs
fee1-dead: Done

compiler/rustc_trait_selection/src/traits/query/evaluate_obligation.rs
```diff
@@ -46,6 +49,12 @@ impl<'cx, 'tcx> InferCtxtExt<'tcx> for InferCtxt<'cx, 'tcx> {
         &self,
         obligation: &PredicateObligation<'tcx>,
     ) -> bool {
+        if let PredicateKind::Trait(pred) = obligation.predicate.kind().skip_binder() {
+            if let hir::Constness::Const = pred.constness {
+                // do not evaluate to holds when we have a const predicate.
```
oli-obk, line 54: What is going on here? Why aren't we checking const trait predicates?

fee1-dead: Well... This is an awkward fix. The problem is without this check to prevent us from using the cache and so on, the ui tests would fail. I might have time tomorrow to try to find a better way to approach this.

oli-obk: Seems fine for now, but document the behaviour + whatever you remember about details.

GuillaumeGomez:
I \*think\* this is the PR failing in https://github.com/rust-lang/rust/pull/87519#issuecomment-887792031.

fee1-dead:
See https://github.com/rust-lang/rust/pull/87519#issuecomment-887885335

fee1-dead:
@oli-obk In [this zulip thread](https://rust-lang.zulipchat.com/#narrow/stream/146212-t-compiler.2Fconst-eval/topic/Bounds.20on.20non-type-parameters.20work.2C.20is.20that.20on.20purpose.3F/near/222214631) there was an agreement to do a crater run for the fix. I am telling you that this PR actually does that fix. I was unaware of the previous discussion, so maybe `r-` this and do a crater run first?

oli-obk:
ugh I totally forgot about that. Yea, we may have to invent new syntax for const bounds, as ralf notes, it would be more consistent.

@bors r-

@bors try

oli-obk:
@craterbot chkeck

oli-obk:
@craterbot check

Mark-Simulacrum:
@craterbot retry

fee1-dead:
Two Regressions:

`enrgy v0.2.2` breaks due to BTreeMap::new now requiring a const K: Ord bound:
```
[INFO] [stdout] error[E0277]: the trait bound `TypeId: Ord` is not satisfied
[INFO] [stdout]    --> /opt/rustwide/cargo-home/git/checkouts/enrgy-6dfd4514f7a68b81/bb98d92/src/extensions.rs:14:20
[INFO] [stdout]     |
[INFO] [stdout] 14  |             inner: BTreeMap::new(),
[INFO] [stdout]     |                    ^^^^^^^^^^^^^ the trait `Ord` is not implemented for `TypeId`
[INFO] [stdout]     |
[INFO] [stdout] note: required by `BTreeMap::<K, V>::new`
[INFO] [stdout] help: consider introducing a `where` bound, but there might be an alternative better way to express this requirement
[INFO] [stdout]     |
[INFO] [stdout] 10  | impl Extensions where TypeId: Ord {
[INFO] [stdout]     |                 ^^^^^^^^^^^^^^^^^
[INFO] [stdout] 
[INFO] [stdout] 
[INFO] [stdout] error[E0277]: the trait bound `&str: Ord` is not satisfied
[INFO] [stdout]    --> /opt/rustwide/cargo-home/git/checkouts/enrgy-6dfd4514f7a68b81/bb98d92/src/http/response.rs:24:26
[INFO] [stdout]     |
[INFO] [stdout] 24  |                 headers: BTreeMap::new(),
[INFO] [stdout]     |                          ^^^^^^^^^^^^^ the trait `Ord` is not implemented for `&str`
[INFO] [stdout]     |
[INFO] [stdout]     = note: `Ord` is implemented for `&mut str`, but not for `&str`
[INFO] [stdout] note: required by `BTreeMap::<K, V>::new`
[INFO] [stdout] help: consider introducing a `where` bound, but there might be an alternative better way to express this requirement
[INFO] [stdout]     |
[INFO] [stdout] 17  | impl HttpResponse where &str: Ord {
[INFO] [stdout]     |                   ^^^^^^^^^^^^^^^
```

`enso-generics v0.2.0` breaks because const bounds on functions are now enforced:
```
[INFO] [stdout] error[E0277]: the trait bound `T: HasLength` is not satisfied
[INFO] [stdout]   --> src/hlist.rs:80:77
[INFO] [stdout]    |
[INFO] [stdout] 75 | pub const fn len<T:HasLength>() -> usize {
[INFO] [stdout]    |                    --------- required by this bound in `hlist::len`
[INFO] [stdout] ...
[INFO] [stdout] 80 | impl<H,T:HasLength> HasLength for Cons<H,T> { const LEN : usize = 1 + len::<T>(); }
[INFO] [stdout]    |                                                                             ^ the trait `HasLength` is not implemented for `T`
[INFO] [stdout]    |
[INFO] [stdout] help: consider further restricting this bound
[INFO] [stdout]    |
[INFO] [stdout] 80 | impl<H,T:HasLength + hlist::HasLength> HasLength for Cons<H,T> { const LEN : usize = 1 + len::<T>(); }
[INFO] [stdout]    |                    ^^^^^^^^^^^^^^^^^^
```

Not many crates will be broken with this change, I will contact the crate authors to fix these problems.

nbdd0121:
Why do we need `K: Ord` bound on `BTreeMap` in the first place? Can't we just remove the bound?

fee1-dead:
I don't think we should remove the bound. I will add a commit to move the bound to the impl so it will not have a const bound.

library/alloc/src/collections/btree/map.rs
```diff
@@ -490,7 +490,9 @@ impl<K: fmt::Debug, V: fmt::Debug> fmt::Debug for RangeMut<'_, K, V> {
     }
 }
 
-impl<K, V> BTreeMap<K, V> {
+// N.B. bound not on function to avoid unnecessary
+// const bound on `Ord`: see #87375
+impl<K: Ord, V> BTreeMap<K, V> {
```
oli-obk, line 495: this should not make a difference at all. All bounds on the impl must apply for functions defined inside, too.

if this truly fixes things, that is a bug. Instead we should use a `?const Ord` bound

fee1-dead: Hmmmm, are you saying bounds on impls should be treated as const bounds when inherent functions are const? That would break a lot of existing code I think.

```rust
impl<T: Clone> MyType {
    fn a() { ... }; // T: Clone
    const fn b() { ... }; // T: const Clone? 
}
```

oli-obk: Only code using feature gates, stable code already doesn't allow this: https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=428faddc9d2bce802ab7423fdfa7db33

fee1-dead: Wow, I really didn't know this. But can we use `?const` inside liballoc? That feature is marked as incomplete.

oli-obk: hmm... I guess we are just going to break ppl using btreemap in const contexts? Or try to remove the bound in a separate PR first?

fee1-dead: From the crater run I think not many people are using this feature. I don't think we should remove the bound because it might require more work when re-adding the bound back. I'm fine with breaking BTree{Map,Set}::new for now.

compiler/rustc_trait_selection/src/traits/select/mod.rs
```diff
@@ -293,6 +316,18 @@ impl<'cx, 'tcx> SelectionContext<'cx, 'tcx> {
         self.infcx.tcx
     }
 
+    /// returns `true` if the predicate is considered `const` to
```
kw-fn, line 319:
```suggestion
    /// Returns `true` if the predicate is considered `const` to
```

compiler/rustc_infer/src/traits/engine.rs
```diff
@@ -49,11 +50,28 @@ pub trait TraitEngine<'tcx>: 'tcx {
...
+    ) -> Result<(), Vec<FulfillmentError<'tcx>>> {
+        self.select_all_or_error(infcx)
+    }
+
     fn select_where_possible(
         &mut self,
         infcx: &InferCtxt<'_, 'tcx>,
     ) -> Result<(), Vec<FulfillmentError<'tcx>>>;
 
+    // FIXME this should not provide a default body for chalk as chalk should be updated
```
kw-fn, line 66:
```suggestion
    // FIXME(fee1-dead): this should not provide a default body for chalk as chalk should be updated
```

compiler/rustc_traits/src/type_op.rs
```diff
@@ -85,7 +87,16 @@ impl AscribeUserTypeCx<'me, 'tcx> {
         Ok(())
     }
 
-    fn prove_predicate(&mut self, predicate: Predicate<'tcx>) {
+    fn prove_predicate(&mut self, mut predicate: Predicate<'tcx>) {
+        if let PredicateKind::Trait(mut tr) = predicate.kind().skip_binder() {
+            if let hir::Constness::Const = tr.constness {
+                // FIXME check if we actually want to prove const predicates inside AscribeUserType
```
kw-fn, line 93:
```suggestion
                // FIXME(fee1-dead): check if we actually want to prove const predicates inside AscribeUserType
```

library/alloc/tests/const_fns.rs
```diff
@@ -0,0 +1,53 @@
...
+// 3. BTreeMap::new()
+// 4. BTreeSet::new()
+
+#[allow(dead_code)]
+pub const MY_VEC: Vec<usize> = Vec::new();
+
+#[allow(dead_code)]
+pub const MY_STRING: String = String::new();
+
+// FIXME remove this struct once we put `K: ?const Ord` on BTreeMap::new.
```
kw-fn, line 13:
```suggestion
// FIXME(fee1-dead): remove this struct once we put `K: ?const Ord` on BTreeMap::new.
```

library/alloc/tests/lib.rs
```diff
@@ -23,6 +23,10 @@
 #![feature(slice_partition_dedup)]
 #![feature(vec_spare_capacity)]
 #![feature(string_remove_matches)]
+#![feature(const_btree_new)]
+#![feature(const_trait_impl)]
+// FIXME remove this when const_trait_impl is not incomplete anymore
```
kw-fn, line 28:
```suggestion
// FIXME(fee1-dead): remove this when const_trait_impl is not incomplete anymore
```

compiler/rustc_trait_selection/src/traits/select/mod.rs
fee1-dead: I will change this after this lands or on this PR if CI fails

PR #87376, opened 2021-07-22
Change rustdoc logo to use the full container size

Thomasdezeeuw:
We have a logo in svg that scales nicely to large sizes, but by default
is only 5px large, i.e. very small. With the change the logo expands to
the full size. By only setting the height to 100% we ensure that the
width-height ratio isn't changed.


GuillaumeGomez:
What happens in case the image has a bigger width than height?

Thomasdezeeuw:
The `max-width` and `max-height` ensure it doesn't grow beyond 100px, I've also checked manually with large images.

GuillaumeGomez:
Thanks!

@bors: r+ rollup

PR #82653, opened 2021-03-01
Update all submodules that rustbuild doesn't depend on lazily

jyn514:
This only updates the submodules the first time they're needed, instead
of unconditionally the first time you run x.py.

Ideally, this would move *all* submodules to rustbuild and not exclude some tools and
backtrace. Unfortunately, cargo requires all `Cargo.toml` files in the
whole workspace to be present to build any crate.

On my machine, this takes the time for an initial submodule clone (for
`x.py --help`) from 55.70 to 15.87 seconds.

Helps with https://github.com/rust-lang/rust/issues/76653. Builds on https://github.com/rust-lang/rust/pull/86015 and should not be merged before (only the last commit is relevant).

PR #87380, opened 2021-07-22
Don't default to `submodules = true` unless the rust repo has a .git directory

jyn514:
Should hopefully fix https://github.com/rust-lang/rust/pull/82653#issuecomment-885093033 - @semarie can you confirm?

r? @Mark-Simulacrum


semarie:
@jyn514 yes, with the patch applied on top of latest tarball, it is building.

PR #77633, opened 2020-10-06
Implement advance_by, advance_back_by for slice::{Iter, IterMut}

timvermeulen:
Part of #77404.

Would be nice if we can get away with getting rid of `nth[_back]` altogether, but if not, we can keep `nth` and `advance_by` alongside each other.

Also see https://github.com/rust-lang/rust/pull/76909#issuecomment-704543329.

cc @ecstatic-morse @scottmcm

Issue #77404, opened 2020-10-01
Tracking Issue for feature(iter_advance_by)


timvermeulen:
This is a tracking issue for the methods `Iterator::advance_by` and `DoubleEndedIterator::advance_back_by`.
The feature gate for the issue is `#![feature(iter_advance_by)]`.

---

Previously the recommendation was to implement `nth` and `nth_back` on your iterators to efficiently advance them by multiple elements at once (useful for `.skip(n)` and `.step_by(n)`). After this feature is stabilized the recommendation will/should be to implement `advance_by` and `advance_back_by` instead, because they compose better and are often simpler to implement.

Iterators in libcore that wrap another iterator (possibly from elsewhere than libcore) will need to keep their `nth` and `nth_back` implementations for the foreseeable future and perhaps indefinitely, because the inner iterator may have an efficient `nth` implementation without implementing `advance_by` as well.

### About tracking issues

Tracking issues are used to record the overall progress of implementation.
They are also used as hubs connecting to other relevant issues, e.g., bugs or open design questions.
A tracking issue is however *not* meant for large scale discussion, questions, or bug reports about a feature.
Instead, open a dedicated issue for the specific matter and add the relevant feature gate label.

### Steps

- [x] Implementation (#76909)
- [ ] Stabilization PR ([see instructions on rustc-dev-guide][stabilization-guide])

[stabilization-guide]: https://rustc-dev-guide.rust-lang.org/stabilization_guide.html#stabilization-pr

timvermeulen:
I’ve already written implementations of these methods for most iterators in `core::iter`, I’ll submit PRs for them once I have enough tests.

GuillaumeGomez:
The fact that `advance_by` returns a `Result` isn't great in my opinion. In case it moves after the end of the iterator, it's not an error, just that it ran after the end. It would make more sense to return `Some(how_much_after_the_end)` instead of `Err(how_much_after_the_end)`.

m-ou-se:
> Previously the recommendation was to implement nth and nth_back on your iterators to efficiently advance them by multiple elements at once (useful for .skip(n) and .step_by(n)). After this feature is stabilized the recommendation will/should be to implement advance_by and advance_back_by instead, because they compose better and are often simpler to implement.

[...]

PR #87387, opened 2021-07-22
Implement advance_by, advance_back_by for slice::{Iter, IterMut}

the8472:
Part of #77404.

Picking up where #77633 was closed.

I have addressed https://github.com/rust-lang/rust/pull/77633#issuecomment-771842599 by restoring `nth` and `nth_back`. So according to that comment this should already be r=m-ou-se, but it has been sitting for a while.

### DIFF
--- a/library/core/src/slice/iter/macros.rs
+++ b/library/core/src/slice/iter/macros.rs
@@ -188,7 +188,7 @@ macro_rules! iterator {
             #[inline]
             fn advance_by(&mut self, n: usize) -> Result<(), usize>{
                 let advance = cmp::min(n, len!(self));
-                // SAFETY: `advance` does not exceed `self.len()` by construction
+                // SAFETY: By construction, `advance` does not exceed `self.len()`.
                 unsafe { self.post_inc_start(advance as isize) };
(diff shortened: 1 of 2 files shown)


scottmcm:
Looks good; thanks!

@bors r+ rollup=iffy

Mark-Simulacrum:
Seems to have been a consistent [regression](https://perf.rust-lang.org/compare.html?start=b289bb7fdfcb6f54d825927ab9b5722cabc2a140&end=6b0b07d41f07e1ba5808693d900903499ccf7a32&stat=instructions:u) on the regex-opt benchmark in particular, amongst a few others.

Marking as perf-regression. Seems likely that we just accept this regression, though.

the8472:
@Mark-Simulacrum it looks like I forgot an `#[inline]` there, that might perturb things a bit if anything in regex uses `iter.nth()` somewhere which delegates to `advance_by`.

I can do a perf run in another PR to see if it makes any difference.

pnkfelix:
Perf-regression was addressed by PR #87736, so removing perf-regression label.

PR #87395, opened 2021-07-23
Clear up std::env::set_var panic section.

ericonr:
The "K" parameter was being referred to as "key", which wasn't
introduced anywhere.

### DIFF
--- a/library/std/src/env.rs
+++ b/library/std/src/env.rs
@@ -294,7 +294,7 @@ impl Error for VarError {
     }
 }
 
-/// Sets the environment variable `k` to the value `v` for the currently running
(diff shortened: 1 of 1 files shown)


joshtriplett:
`key` is the name of the function parameter. I think it would make sense to change each instance of `k` to `key` rather than the other way around.

ericonr:
I was looking at https://doc.rust-lang.org/std/env/fn.set_var.html , but missed that 44a5bf1b7d8d16e73ab5631c4dd6bd5107079c5f had changed the signature. Will fix.

ericonr:
Done, I think/

joshtriplett:
Thanks!

PR #87396, opened 2021-07-23
Teach compiletest to parse arbitrary cfg options

### LINKED ISSUE #87377: Add support for `// only-cfg-target-has-atomic` to compiletest
Right now, you have to specify the platforms and architectures you want by hand, which makes it easy to leave them out or include platforms you shouldn't. It would be nice to automate this - maybe with `rustc  --print cfg`?

See also https://rust-lang.zulipchat.com/#narrow/stream/182449-t-compiler.2Fhelp/topic/How.20can.20I.20ignore.20tests.20on.20platforms.20without.20atomics.3F.

_Originally posted by @wesleywiser in https://github.com/rust-lang/rust/pull/84039#discussion_r675000124_

inquisitivecrystal:
This gives compiletest the ability to check for any cfg option that rustc sets by default for certain platforms. For instance, it is now possible to only run a test on targets where `target_has_atomic="8"` is true. 

Resolves #87377.

### DIFF
--- a/src/tools/compiletest/src/common.rs
+++ b/src/tools/compiletest/src/common.rs
@@ -1,5 +1,6 @@
 pub use self::Mode::*;
 
+use std::collections::HashMap;
 use std::ffi::OsString;
 use std::fmt;
 use std::path::{Path, PathBuf};
@@ -274,6 +275,10 @@ pub struct Config {
     /// What panic strategy the target is built with.  Unwind supports Abort, but
     /// not vice versa.
     pub target_panic: PanicStrategy,
+
+    /// A map from target cfg keys to their values. If the key is present with no value,
+    /// the vector will contain the empty string.
+    pub target_cfg: HashMap<String, Vec<String>>,
 
     /// Target system to be tested
     pub target: String,
--- a/src/tools/compiletest/src/common.rs
+++ b/src/tools/compiletest/src/common.rs
@@ -1,6 +1,5 @@
 pub use self::Mode::*;
 
-use std::collections::HashMap;
 use std::ffi::OsString;
 use std::fmt;
 use std::path::{Path, PathBuf};
@@ -275,10 +274,6 @@ pub struct Config {
     /// What panic strategy the target is built with.  Unwind supports Abort, but
     /// not vice versa.
     pub target_panic: PanicStrategy,
-
-    /// A map from target cfg keys to their values. If the key is present with no value,
-    /// the vector will contain the empty string.
-    pub target_cfg: HashMap<String, Vec<String>>,
 
     /// Target system to be tested
     pub target: String,
--- a/src/tools/compiletest/src/header.rs
+++ b/src/tools/compiletest/src/header.rs
@@ -16,9 +16,12 @@ use crate::{extract_cdb_version, extract_gdb_version};
 #[cfg(test)]
 mod tests;
 
-/// The result of parse_cfg_name_directive.
+/// Whether a given property holds true of this test instance. Usually,
+/// a property is something about the target (e.g. whether the target is Windows),
+/// but it could also be something more general (e.g. whether this test is being
+/// cross-compiled.)
 #[derive(Clone, Copy, PartialEq, Debug)]
-enum ParsedNameDirective {
+enum EvaluatedProp {
     /// No match.
     NoMatch,
     /// Match.
@@ -228,6 +231,10 @@ impl TestProps {
         if !testfile.is_dir() {
             let file = File::open(testfile).unwrap();
 
+            let mut target = config.target.clone();
+            let mut per_prop_normalize_stdout: Vec<(String, (String, String))> = Vec::new();
+            let mut per_prop_normalize_stderr: Vec<(String, (String, String))> = Vec::new();
+
             iter_header(testfile, file, &mut |revision, ln| {
                 if revision.is_some() && revision != cfg {
                     return;
@@ -340,11 +347,16 @@ impl TestProps {
                     self.ignore_pass = config.parse_ignore_pass(ln);
                 }
 
-                if let Some(rule) = config.parse_custom_normalization(ln, "normalize-stdout") {
-                    self.normalize_stdout.push(rule);
+                if let Some(target_override) = config.parse_target_override(ln) {
+                    target = target_override;
                 }
+
+                if let Some(rule) = config.parse_custom_normalization(ln, "normalize-stdout") {
+                    per_prop_normalize_stdout.push(rule);
+                }
+
                 if let Some(rule) = config.parse_custom_normalization(ln, "normalize-stderr") {
-                    self.normalize_stderr.push(rule);
+                    per_prop_normalize_stderr.push(rule);
                 }
 
                 if let Some(code) = config.parse_failure_status(ln) {
@@ -372,6 +384,17 @@ impl TestProps {
                     self.incremental = config.parse_incremental(ln);
                 }
             });
+            for (prop, normalization) in per_prop_normalize_stdout {
+                if config.evaluate_prop_for_target(&prop, &target) == EvaluatedProp::Match {
+                    self.normalize_stdout.push(normalization);
+                }
+            }
+
+            for (prop, normalization) in per_prop_normalize_stderr {
+                if config.evaluate_prop_for_target(&prop, &target) == EvaluatedProp::Match {
+                    self.normalize_stderr.push(normalization);
+                }
+            }
         }
 
         if self.failure_status == -1 {
@@ -650,14 +673,15 @@ impl Config {
         }
     }
 
-    fn parse_custom_normalization(&self, mut line: &str, prefix: &str) -> Option<(String, String)> {
-        if self.parse_cfg_name_directive(line, prefix) == ParsedNameDirective::Match {
-            let from = parse_normalization_string(&mut line)?;
-            let to = parse_normalization_string(&mut line)?;
-            Some((from, to))
-        } else {
-            None
-        }
+    fn parse_custom_normalization(
+        &self,
+        mut line: &str,
+        prefix: &str,
+    ) -> Option<(String, (String, String))> {
+        let prop = self.parse_prop_directive(line, prefix)?;
+        let from = parse_normalization_string(&mut line)?;
+        let to = parse_normalization_string(&mut line)?;
+        Some((prop, (from, to)))
     }
 
     fn parse_needs_matching_clang(&self, line: &str) -> bool {
@@ -670,7 +694,19 @@ impl Config {
 
     /// Parses a name-value directive which contains config-specific information, e.g., `ignore-x86`
     /// or `normalize-stderr-32bit`.
-    fn parse_cfg_name_directive(&self, line: &str, prefix: &str) -> ParsedNameDirective {
+    fn parse_prop_directive(&self, line: &str, prefix: &str) -> Option<String> {
+        if !line.as_bytes().starts_with(prefix.as_bytes()) {
+            return None;
+        }
+        if line.as_bytes().get(prefix.len()) != Some(&b'-') {
+            return None;
+        }
+
+        let config = &line[&prefix.len() + 1..].to_string();
+        Some(config.to_string())
+    }
+
+    fn evaluate_prop_for_target(&self, prop: &str, target: &str) -> EvaluatedProp {
         // This matches optional whitespace, followed by a group containing a series of word
         // characters (including '_' and '-'), followed optionally by a sequence consisting
         // of a colon, optional whitespace, and another group containing word characters.
@@ -697,28 +733,21 @@ impl Config {
             static ref CFG_REGEX: Regex = Regex::new(r"^\s*([\w-]+)(?::\s*([\w-]+))?").unwrap();
         }
 
-        if !line.as_bytes().starts_with(prefix.as_bytes()) {
-            return ParsedNameDirective::NoMatch;
-        }
-        if line.as_bytes().get(prefix.len()) != Some(&b'-') {
-            return ParsedNameDirective::NoMatch;
-        }
-
-        let captures = CFG_REGEX.captures(&line[&prefix.len() + 1..]).unwrap();
+        let captures = CFG_REGEX.captures(&prop).unwrap();
         let name = captures.get(1).unwrap().as_str();
         let maybe_value = captures.get(2).map(|v| v.as_str().trim());
 
         let is_match = name == "test" ||
-            self.target == name ||                              // triple
-            util::matches_os(&self.target, name) ||             // target
-            util::matches_env(&self.target, name) ||            // env
-            self.target.ends_with(name) ||                      // target and env
-            name == util::get_arch(&self.target) ||             // architecture
-            name == util::get_pointer_width(&self.target) ||    // pointer width
+            target == name ||                                   // triple
+            util::matches_os(target, name) ||                   // target
+            util::matches_env(target, name) ||                  // env
+            target.ends_with(name) ||                           // target and env
+            name == util::get_arch(target) ||                   // architecture
+            name == util::get_pointer_width(target) ||          // pointer width
             name == self.stage_id.split('-').next().unwrap() || // stage
             name == self.channel ||                             // channel
-            (self.target != self.host && name == "cross-compile") ||
-            (name == "endian-big" && util::is_big_endian(&self.target)) ||
+            (target != self.host && name == "cross-compile") ||
+            (name == "endian-big" && util::is_big_endian(target)) ||
             (self.remote_test_client.is_some() && name == "remote") ||
             match self.compare_mode {
                 Some(CompareMode::Nll) => name == "compare-mode-nll",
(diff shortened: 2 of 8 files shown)


src/tools/compiletest/src/main.rs
```diff
@@ -211,6 +211,13 @@ pub fn parse_config(args: Vec<String>) -> Config {
         // Avoid spawning an external command when we know tidy won't be used.
         false
     };
+    let rustc_cfg_data = Command::new("rustc")
```
Urgau, line 214: You should use the `rustc-path` from the config below instead of the plain `rustc` that you wrote.

inquisitivecrystal: Thank you!

src/tools/compiletest/src/main.rs
```diff
@@ -211,11 +213,24 @@ pub fn parse_config(args: Vec<String>) -> Config {
         // Avoid spawning an external command when we know tidy won't be used.
         false
     };
+
+    let target_cfg = if cfg!(test) {
```
inquisitivecrystal, line 217: I'm not sure this is great, but I don't know how else to get around the fact that a `rustc_path` seems not to be provided for test runs. There could be something obvious I'm missing though. I'm pretty tired.

src/tools/compiletest/src/main.rs
```diff
@@ -211,11 +213,24 @@ pub fn parse_config(args: Vec<String>) -> Config {
...
+        HashMap::new()
+    } else {
```
inquisitivecrystal, line 219: I'm not sure this is great, but I don't know how else to get around the fact that a rustc_path seems not to be provided for test runs. There could be something obvious I'm missing though. I'm pretty tired.

src/tools/compiletest/src/util.rs
```diff
@@ -211,6 +213,55 @@ pub fn logv(config: &Config, s: String) {
...
+
+/// Checks if the cfg HashMap has the given `name`. If the `required_value` is
+/// `Some(value)`, this will only return `true` if `name` is associated with `value`.
+pub fn cfg_has(
+    map: &HashMap<String, Vec<String>>,
+    name: &str,
+    required_value: Option<&str>,
+) -> bool {
+    let name = name.replace("-", "_");
+    let required_value = required_value.map(str::trim).map(str::to_string);
```
inquisitivecrystal, line 256: Note to self: change this to one map operation with a closure.

nagisa:
I don't see how this handles situations where the test specifies specific flags for rustc, e.g. like this:

```rust
// compile-flags: --target=wasm32-unknown-unknown -Ctarget-feature=+simd128
// cfg-target-feature: simd128
```

should probably work.

inquisitivecrystal:
> I don't see how this handles situations where the test specifies specific flags for rustc, e.g. like this:
> 
> ```rust
> // compile-flags: --target=wasm32-unknown-unknown -Ctarget-feature=+simd128
> // cfg-target-feature: simd128
> ```
> 
> should probably work.

It doesn't. I agree that it would be lovely if that were supported. However, it's more complex to implement, because that information isn't presently available to the section of the code that parses configurations. The current version of per target configuration doesn't support that either, so I don't think it makes a lot of sense to block this on adding it. Additionally, there's a relatively easy workaround for that case: just check whether the feature you want is supported on the target you have requested. It's not perfect, but it's a lot more workable than checking whether the feature is supported on every individual target, having to account for the possibility that future targets with different features might be added.

src/tools/compiletest/src/common.rs
```diff
@@ -275,6 +276,10 @@ pub struct Config {
     /// not vice versa.
     pub target_panic: PanicStrategy,
 
+    /// A map from target cfg keys to their values. If the key is present with no value,
+    /// the vector will contain the empty string.
+    pub target_cfg: HashMap<String, Vec<String>>,
```
Mark-Simulacrum, line 281: I think it'd be useful to give a short example here (e.g. the has atomic kv pairs).

src/tools/compiletest/src/header.rs
```diff
@@ -642,14 +644,42 @@ impl Config {
...
     /// or `normalize-stderr-32bit`.
     fn parse_cfg_name_directive(&self, line: &str, prefix: &str) -> ParsedNameDirective {
+        // This matches optional whitespace, followed by a group containing a series of word
+        // characters (including '_' and '-'), followed optionally by a sequence consisting
+        // of a colon, optional whitespace, and another group containing word characters.
+        //
+        // Matches in full:
+        //   cfg-target-has-atomic: 128
+        //   cfg-target-has-atomic:128
+        //   cfg-target-has-atomic
```
Mark-Simulacrum, line 654: cfg: target-has-atomic=128 rather than this syntax might be a bit better - it matches more closely the syntax used in cfg and similar, which seems positive (and more easily extendable to any or other directives in the future).

src/tools/compiletest/src/header.rs
```diff
@@ -642,14 +644,42 @@ impl Config {
...
+        //
+        // Matches up to the first space (exclusive):
+        //   ignore-test - This test really shouldn't ever be run
+        //
+        // Matches up to the second space (exclusive):
+        //   cfg-target-has-atomic: 128 - this requires fancy newfangled atomics
+        //
+        // Matches up to the second colon (exclusive)
```
Mark-Simulacrum, line 662: I'm not following the intent of this case - what is this useful for?

src/tools/compiletest/src/main.rs
Mark-Simulacrum: Hm, it seems likely that should be fixable- maybe a separate PR is best, but it seems like we can pass mostly the same args during tests via an env variable or something...

Can you leave a FIXME comment here perhaps?

src/tools/compiletest/src/util.rs
```diff
@@ -211,6 +213,55 @@ pub fn logv(config: &Config, s: String) {
     }
 }
 
+/// Parses the output of `rustc --print cfg` into a [`HashMap`] containing
+/// the cfg values for each name. If the name is present with no value,
+/// it is treated as having a value of "".
```
Mark-Simulacrum, line 218: Does this match rustc's handling of no value cfgs? If not, is there a reason for using an empty string instead of an empty vector?

src/tools/compiletest/src/util.rs
```diff
@@ -211,6 +213,55 @@ pub fn logv(config: &Config, s: String) {
...
+/// to a vector, this adds a new vector containing only `value` to the [`HashMap`].
+fn cfg_add(map: &mut HashMap<String, Vec<String>>, name: &str, value: &str) {
+    let name = name.to_string();
+    let value = value.to_string();
+
+    if let Some(values) = map.get_mut(&name) {
+        values.push(value.to_string());
+    } else {
+        map.insert(name, vec![value.to_string()]);
+    }
```
Mark-Simulacrum, line 245: Fwiw I usually write this sort of thing with the entry API, entry(key).or_default().push(...).

src/tools/compiletest/src/util.rs
```diff
@@ -211,6 +213,55 @@ pub fn logv(config: &Config, s: String) {
...
+) -> bool {
+    let name = name.replace("-", "_");
+    let required_value = required_value.map(str::trim).map(str::to_string);
+
+    match (map.get(&name), required_value) {
+        (None, _) => false,
+        (Some(_), None) => true,
+        (Some(values), Some(required_value)) => values.contains(&required_value),
+    }
+}
```
Mark-Simulacrum, line 263: I think if we go with my prior suggestion and use `cfg: foo` as the syntax, we can get rid of all this parsing and just look for a matching line in the rustc output, right?

src/tools/compiletest/src/header.rs
inquisitivecrystal: We could I guess? The reason I'm hesitating is that this syntax is standard for compiletest header directives. `=` would require you to remember that this particular directive works differently from all other directives, including other `only` and `ignore` directives, which are otherwise the same. That seems more confusing than breaking with the cfg format. This is also why I've used dashes, rather than underscores, for the examples.

src/tools/compiletest/src/header.rs
inquisitivecrystal: Comments. There are a bunch of situations where we're using comments on the same line as a directive, using `- comment` or whatever. I didn't want to break those. I could leave a comment here explaining that?

src/tools/compiletest/src/main.rs
inquisitivecrystal: Sure!

src/tools/compiletest/src/header.rs
Mark-Simulacrum: None of the other directives are trying to emulate cfg() attributes essentially; they have their own ad-hoc syntaxes. I think in this case it's closer to "this directive matches the stuff you can put inside cfg (with some limitations currently)".

src/tools/compiletest/src/util.rs
inquisitivecrystal: Well, I thought that rustc is inconsistent. In some cases it has just the value, with no `=`, and in other cases it has the empty string as a value. Looking more closely, there is a semantic difference. The empty string indicates that there could be a value but isn't one, whereas no value indicates that there couldn't be a value. For example, my Mac has `target_abi=""` but also `unix`.

I standardized on the empty string because that was easier to code into `cfg_add`. There's no way to match on the empty string specifically, and no point in adding one, so it's pretty irrelevant for our purposes.  I can change it if you want. I do like the elegance of this code though... :)

src/tools/compiletest/src/util.rs
inquisitivecrystal: I can do that. It's a bit cleaner.

src/tools/compiletest/src/header.rs
inquisitivecrystal: Oh, wait, I totally misread your comment. Yeah, that... would work fine, actually. I can do that.

src/tools/compiletest/src/util.rs
inquisitivecrystal: We'd have to do prefix based matching, if we wanted to keep the ability to do check for presence of an option without looking at its value. I'm having trouble remembering what the use case was for that... Other than that, we'd have to do some normalization on the rustc output or the header directive, to make them follow the same format while keeping it ergonomic. Do you think it's a good idea to make the change?

src/tools/compiletest/src/header.rs
Mark-Simulacrum: Those aren't present on any cfg- lines though, right?

I don't think we should use a colon to terminate lines anyway, that seems "weird". Ultimately I'm not too worried about supporting comments *on the same line* -- the only worry would be if the commit causes us to silently not run a test (which the author may not notice). Not sure how much we can detect mistakes here though... I recall there was some logic added to this effect in Cargo(?) recently, but not 100% sure if that works for us or we can use it...

src/tools/compiletest/src/util.rs
Mark-Simulacrum: What normalization are you expecting? I would expect a 1-1 match between the syntax here and in rustc output, I think...?

inquisitivecrystal: The normalization would be removing quotes and spaces. Removing quotes isn't strictly necessary, but makes things a bit easier to use and decreases the chance of errors. Removing spaces is necessary to match `cfg!`, which allows spaces. rustc's output doesn't have any spaces.

nagisa:
> Given all that, I don't think we should block this on adding compile-flags support.

I think it is fairly important that writing a test is easy to get right intuitively, and perhaps combining concepts seen in other tests. My view of this specific implementation is that the behaviour is not intuitive at all.

I don't mind this being a separate PR, but from what I can tell, a better integrated implementation would have to redo most of this PR either way. At the very least I want to see a tidy lint that disallows tricky combinations, if we land this feature as-is.

inquisitivecrystal:
> > Given all that, I don't think we should block this on adding compile-flags support.
> 
> I think it is fairly important that writing a test (possibly by combining concepts seen in other tests) is easy to get correct. My view of this specific implementation is that the behaviour is not intuitive at all.
> 
> I don't mind this being a separate PR, but from what I can tell, a better integrated implementation would have to redo most of this PR either way. At the very least I want to see a tidy lint that disallows tricky combinations, if we land this feature as-is.

That seems fair. This isn't any worse than the other `only` and `ignore` directives, but one could say that the behavior is confusing there too. When I have a chance, I'll have a go at actually supporting it, and if that seems like too much work I'll add the tidy check. I'll also implement Mark-Simulacrum's comments at the same time.

 Also, you're right that a PR that did that would have to rewrite a fair portion of this: I'd estimate about 50%.

inquisitivecrystal:
I think I'm going to temporarily close this. I am working on it again though. Hopefully I'll have progress soon, and I'll reopen then.

inquisitivecrystal:
I feel like this is worth reopening for another round of comments. I believe that I've implemented all of the feedback I previously received. I'm sincerely sorry that this took so long. It's been a busy few months, but that's no excuse.

Things that are missing from this implementation are caching of the results and testing. I'd like hear people's thoughts about whether caching is necessary and how to go about it; I'll leave a comment on the relevant area of code. I think testing would likely be advisable, but I'm still working out the best way to go about it. I'm also clarifying some of the relevant details with someone else at the moment.

One other specific thing that I wanted to draw attention to is that I changed the code to refer to the sort of thing that goes after an `ignore-` as a "prop", for property. They were previously called `cfg`s. The problem was that this file uses `cfg` and `config` and `Config` to refer to so many things that I was finding it difficult to reason about which was which. Additionally, although there are some analogies between these properties and `#[cfg(...)]` directives, they aren't all the sort of thing that would go in that kind of directive. Feedback is welcome.

Please feel welcome to point out anything that I've missed, and of course any new comments as well. I'm sure there are places where the code could be improved. CC @Mark-Simulacrum and @nagisa, who reviewed this previously.

src/tools/compiletest/src/header.rs
```diff
@@ -704,12 +736,31 @@ impl Config {
...
 
-        if is_match { ParsedNameDirective::Match } else { ParsedNameDirective::NoMatch }
+        let cfg_data = Command::new(&self.rustc_path)
+            .args(&["--target", &target])
+            .args(&["--print", "cfg"])
+            .output()
+            .unwrap()
+            .stdout;
+        let cfg_data = String::from_utf8(cfg_data).unwrap();
+        let cfg_data = cfg_data.replace('"', "");
```
inquisitivecrystal, line 756: The `cfg_data` could presumably be cached somewhere around this point. Doing so could presumably result in some small performance saving for the case where `target == self.target`, which is likely to come up with the most frequency. I'm not sure whether this is worthwhile, or if so, how to go about it.

As an amusing anecdote, I actually fiddled around with a version where all of the targets had their information calculated in advance and it was stored in `Config`. I figured that computing all of that would only take a few extra seconds. It took me a few days to figure out why doing so was causing a one third to one half regression in testing times. It turns out that `Config` gets cloned a fair number of times, and by the end was taking enough memory that it was vastly increasing the duration of forking the threads to execute the tests.

inquisitivecrystal:
I added tests!

Mark-Simulacrum:
I think I agree with @nagisa that this should respect compile-flags -- or, at least, abort if any such flags are specified so that it can't be combined with compile-flags by accident in a way that leads to misleading behavior. Is that implemented in the current version of the PR?

I'll also note that I feel a little uncertain about the exact shape of these options. It feels like they're an improvement over the manual listing of various targets to ignore, but are a little weak in terms of readability -- `ignore-cfg: ...` doesn't obviously suggest ignore *if* this cfg is true to me. Maybe `ignore-if-cfg` would be better here, not sure.

inquisitivecrystal:
> I think I agree with @nagisa that this should respect compile-flags -- or, at least, abort if any such flags are specified so that it can't be combined with compile-flags by accident in a way that leads to misleading behavior. Is that implemented in the current version of the PR?

Well, I thought it was, but... I only made it respect the `--target` compile-flag. Looking back at nagisa's example, that was a mistake. `-Ctarget-feature` is relevant as well, for instance. I'm not sure why someone would set a target feature and then check for it, but it should probably work if they do. I'll change it to just use all of the flags. The target flag will still need special-casing, so this is going to get interesting.

> I'll also note that I feel a little uncertain about the exact shape of these options. It feels like they're an improvement over the manual listing of various targets to ignore, but are a little weak in terms of readability -- `ignore-cfg: ...` doesn't obviously suggest ignore _if_ this cfg is true to me. Maybe `ignore-if-cfg` would be better here, not sure.

I feel uncertain about that too. `ignore-cfg: ...` isn't the clearest. `ignore-if-cfg` seems like a clarity improvement, but still not ideal. Perhaps I should solicit opinions on Zulip? Programmers are terrific at bikeshedding. :)

Dylan-DPC:
@inquisitivecrystal any updates on this?

inquisitivecrystal:
> @inquisitivecrystal any updates on this?

I apologize for the delay. I know this PR has been hanging around for ages at this point. I'm still planning on finishing everything up sooner or later, but it may be a month or so before I have time to work on it again, as I'm extremely busy IRL. Unfortunately, the "proper" fix is so complicated that I don't know how to implement it cleanly (and the difference wouldn't be relevant all that often, in any case). When I do implement this, I'm planning to go with a simpler implementation (using tidy to reject confusing combinations, rather than trying to support them). 

In the meantime, I don't know how to handle the PR. We can close it so WG-triage doesn't have to deal with keeping track of it or leave it open if that's more convenient. I'm even happy to let someone else take over, if there's someone who wants to work on that. In any case, I once again apologize for how long it's taken me to handle this.

Dylan-DPC:
Nothing to apologise about. I am going to close it for now, but feel free to reoopen it if nobody has got to it yet.

PR #87402, opened 2021-07-23
Direct users towards using Rust target feature names in CLI

nagisa:
This PR consists of a couple of changes on how we handle target features.

In particular there is a bug-fix wherein we avoid passing through features that aren't prefixed by `+` or `-` to LLVM. These appear to be causing LLVM to assert, which is pretty poor a behaviour (and also makes it pretty clear we expect feature names to be prefixed).

The other commit, I anticipate to be somewhat more controversial is outputting a warning when users specify a LLVM-specific, or otherwise unknown, feature name on the CLI. In those situations we request users to either replace it with a known Rust feature name (e.g. `bmi` -> `bmi1`) or file a feature request. I've a couple motivations for this: first of all, if users are specifying these features on the command line, I'm pretty confident there is also a need for these features to be usable via `#[cfg(target_feature)]` machinery.  And second, we're growing a fair number of backends recently and having ability to provide some sort of unified-ish interface in this place seems pretty useful to me.

Sponsored by: standard.ai

### DIFF
--- a/compiler/rustc_codegen_llvm/src/llvm_util.rs
+++ b/compiler/rustc_codegen_llvm/src/llvm_util.rs
@@ -221,7 +221,11 @@ pub fn target_features(sess: &Session) -> Vec<Symbol> {
         supported_target_features(sess)
             .iter()
             .filter_map(|&(feature, gate)| {
-                if sess.is_nightly_build() || gate.is_none() { Some(feature) } else { None }
+                if sess.is_nightly_build() || gate.is_none() {
+                    Some(feature)
+                } else {
+                    None
+                }
             })
             .filter(|feature| {
                 for llvm_feature in to_llvm_feature(sess, feature) {
@@ -428,20 +432,15 @@ pub fn llvm_global_features(sess: &Session) -> Vec<String> {
     }
 
     let filter = |s: &str| {
-        if s.is_empty() {
-            return vec![];
-        }
-        let feature = strip(s);
-        if feature == s {
-            return vec![s.to_string()];
-        }
-
-        // Rustc-specific feature requests like `+crt-static` or `-crt-static`
-        // are not passed down to LLVM.
-        if RUSTC_SPECIFIC_FEATURES.contains(&feature) {
-            return vec![];
-        }
-        // ... otherwise though we run through `to_llvm_feature` feature when
+        // features must start with a `+` or `-`.
+        let feature = match s.strip_prefix(&['+', '-'][..]) {
+            None => return vec![],
+            // Rustc-specific feature requests like `+crt-static` or `-crt-static`
+            // are not passed down to LLVM.
+            Some(feature) if RUSTC_SPECIFIC_FEATURES.contains(&feature) => return vec![],
+            Some(feature) => feature,
+        };
+        // ... otherwise though we run through `to_llvm_feature` when
         // passing requests down to LLVM. This means that all in-language
         // features also work on the command line instead of having two
         // different names when the LLVM name and the Rust name differ.
@@ -458,11 +457,11 @@ pub fn llvm_global_features(sess: &Session) -> Vec<String> {
         check_tied_features(sess, &feats.iter().map(|f| (strip(f), !f.starts_with("-"))).collect())
     {
         sess.err(&format!(
-            "Target features {} must all be enabled or disabled together",
+            "target features {} must all be enabled or disabled together",
             f.join(", ")
         ));
     }
-    features.extend(feats.iter().flat_map(|&f| filter(f)));
+    features.extend(feats.iter().flat_map(&filter));
     features
 }
 
--- a/compiler/rustc_codegen_llvm/src/llvm_util.rs
+++ b/compiler/rustc_codegen_llvm/src/llvm_util.rs
@@ -2,14 +2,18 @@ use crate::back::write::create_informational_target_machine;
 use crate::{llvm, llvm_util};
 use libc::c_int;
 use libloading::Library;
-use rustc_codegen_ssa::target_features::{supported_target_features, tied_target_features};
+use rustc_codegen_ssa::target_features::{
+    supported_target_features, tied_target_features, RUSTC_SPECIFIC_FEATURES,
+};
 use rustc_data_structures::fx::{FxHashMap, FxHashSet};
+use rustc_data_structures::small_c_str::SmallCStr;
 use rustc_fs_util::path_to_c_string;
 use rustc_middle::bug;
 use rustc_session::config::PrintRequest;
 use rustc_session::Session;
 use rustc_span::symbol::Symbol;
 use rustc_target::spec::{MergeFunctions, PanicStrategy};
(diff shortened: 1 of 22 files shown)


compiler/rustc_codegen_llvm/src/llvm_util.rs
```diff
@@ -370,36 +366,61 @@ pub fn llvm_global_features(sess: &Session) -> Vec<String> {
...
-    };
-
+    let supported_features = supported_target_features(sess);
     // Features implied by an implicit or explicit `--target`.
-    features.extend(sess.target.features.split(',').filter_map(&filter));
+    features.extend(
+        sess.target
+            .features
+            .split(',')
+            .filter_map(|s| backend_feature_name(s).map(|_| String::from(s))),
```
nagisa, line 375: Likely want to make it a `bug!()` if we get a `None` in this `filter_map`.

estebank (approved):

nagisa:
Hm, one remaining problem is that we call `llvm_global_features` for each function so there will be warnings repeated many times for a typical compilation. Will think about how to best address that.

nagisa:
4 digits of regression. Nice.

compiler/rustc_codegen_llvm/src/llvm_util.rs
```diff
@@ -389,36 +383,90 @@ pub fn llvm_global_features(sess: &Session) -> Vec<String> {
...
+                let enable_disable = match s.chars().next() {
+                    None => return None,
+                    Some(c @ '+' | c @ '-') => c,
+                    Some(_) => {
+                        let mut diag = sess.struct_warn(&format!(
+                            "unknown feature specified for `-Ctarget-feature`: `{}`",
+                            s
+                        ));
+                        diag.note("features must begin with a `+` to enable or `-` to disable it");
+                        diag.emit();
```
bjorn3, line 412: I think this warning should be moved to the driver, so it also works for other codegen backends and for `cargo check`.

nagisa: Given that each backend is currently in charge of parsing everything about the target features, except for splitting by comma. I don't think right now is the right time to move this diagnostic to anywhere.

I do have more changes I want to make with regards to how we handle target features, and moving some of that responsibility away from `rustc_codegen_*` is one of those things. As an example, the only thing codegen backends should probably know about “rustc” target features is how to convert a rustc target feature to its own internal interpretation.

But I think some more refactors are in order before this can happen.

compiler/rustc_codegen_llvm/src/llvm_util.rs
```diff
@@ -389,36 +383,90 @@ pub fn llvm_global_features(sess: &Session) -> Vec<String> {
...
+                        feature
+                    ));
+                    if let Some(rust_feature) = rust_feature {
+                        diag.note("it is still passed through to the codegen backend");
+                        diag.help(&format!("did you mean: `{}`", rust_feature));
+                    } else {
+                        diag.note("it is still passed through to the codegen backend");
+                        diag.note("consider filing a feature request");
+                    }
+                    diag.emit();
```
estebank, line 441:
```suggestion
                    diag.note("it is still passed through to the codegen backend");
                    if let Some(rust_feature) = rust_feature {
                        diag.help(&format!("you might have meant: `{}`", rust_feature));
                    } else {
                        diag.note("consider filing a feature request");
                    }
                    diag.emit();
```

Not too happy with the "consider filing a feature request", but I have no better idea :-/

compiler/rustc_codegen_llvm/src/llvm_util.rs
```diff
@@ -389,36 +383,90 @@ pub fn llvm_global_features(sess: &Session) -> Vec<String> {
...
+                        diag.emit();
+                        return None;
+                    }
+                };
+
+                let feature = backend_feature_name(s)?;
+                // Warn against use of LLVM specific feature names on the CLI.
+                if !supported_features.iter().any(|&(v, _)| v == feature) {
+                    let rust_feature = supported_features.iter().find_map(|&(rust_feature, _)| {
+                        let llvm_features = to_llvm_features(sess, rust_feature);
```
estebank, line 421: Should we be doing fuzzy searching in `to_llvm_features` for the suggestion given bellow?

estebank (approved):
The code changes look reasonable.

compiler/rustc_codegen_llvm/src/llvm_util.rs
nagisa: Possibly. My primary goal was to direct users from using LLVM names towards using Rust names, and much less so to handle typos better. I do agree however that a fuzzy suggestions here would be ideal. Even better if we make them work for more than just `-Ctarget-features`.

compiler/rustc_codegen_llvm/src/llvm_util.rs
nagisa: Hm, is there a reason why this wording makes you concerned? My goal is basically to drive people to describe their use-cases for specific features that they are using so that we could add them to `rustc`.

estebank: It gives the user something to do, but it is kind of... vague. I'm ok with landing it as is though.

pnkfelix:
AFAICT this has been r+'ed and just needs a rebate. Changing to S-waiting-on-author.

JohnCSimon:
> AFAICT this has been r+'ed and just needs a rebase. Changing to S-waiting-on-author.

@nagisa ☝️

nagisa:
Finally got around to rebasing this. The surrounding code changed quite a bit since since July so a re-review is probably warranted. I didn't really make any major functional changes since the last review though, and so all comments/responses above stand.

src/test/ui/target-feature/missing-plusminus.stderr
```diff
@@ -0,0 +1,6 @@
+warning: unknown feature specified for `-Ctarget-feature`: `banana`
+   |
+   = note: features must begin with a `+` to enable or `-` to disable it
```
estebank, line 3: Should we also have a test for `-Ctarget-feature=rdrand` where we tell them "you have to either enable it or disable it"?

nagisa:
AFAIK beta was cut on Friday already, but I don't anticipate coming back to take a look at your review comment before the weekend anyway, so this is definitely not landing before train rollover.

src/test/ui/target-feature/missing-plusminus.stderr
nagisa: Done!

nagisa:
@bors r=estebank

There was a platform-specific test that broke. I fixed the relevant functionality (which involved mostly just calling into the right function at a different time) and made the tests `no_core` so they are also runnable by developers running other architectures.

nagisa:
The failing test seems like it'll be flaky and it doesn't provide much value either way, so I went ahead and removed it entirely.

@bors r=estebank

PR #87404, opened 2021-07-23
Add support for artifact size profiling

rylev:
This adds support for profiling artifact file sizes (incremental compilation artifacts and query cache to begin with). 

Eventually we want to track this in perf.rlo so we can ensure that file sizes do not change dramatically on each pull request. 

This relies on support in measureme: https://github.com/rust-lang/measureme/pull/169. Once that lands we can update this PR to not point to a git dependency. 

This was worked on together with @michaelwoerister.

r? @wesleywiser

### DIFF
--- a/compiler/rustc_data_structures/src/profiling.rs
+++ b/compiler/rustc_data_structures/src/profiling.rs
@@ -110,12 +110,14 @@ bitflags::bitflags! {
         const FUNCTION_ARGS       = 1 << 6;
         const LLVM                = 1 << 7;
         const INCR_RESULT_HASHING = 1 << 8;
+        const ARTIFACT_SIZES = 1 << 9;
 
         const DEFAULT = Self::GENERIC_ACTIVITIES.bits |
                         Self::QUERY_PROVIDERS.bits |
                         Self::QUERY_BLOCKED.bits |
                         Self::INCR_CACHE_LOADS.bits |
-                        Self::INCR_RESULT_HASHING.bits;
+                        Self::INCR_RESULT_HASHING.bits |
+                        Self::ARTIFACT_SIZES.bits;
 
         const ARGS = Self::QUERY_KEYS.bits | Self::FUNCTION_ARGS.bits;
     }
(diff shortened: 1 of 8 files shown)


jyn514:
@rylev the measureme PR was merged it looks like :) do you know when you'll get a chance to follow up?

compiler/rustc_data_structures/src/profiling.rs
```diff
@@ -285,6 +288,33 @@ impl SelfProfilerRef {
         })
     }
 
+    /// Record the size of an artifact that the compiler produces
+    ///
+    /// `artifact_kind` is the class of artifact (e.g., query_cache, object_file, etc.)
+    /// `artifact_name` is an identifier to the specific artifact being stored (usually a filename)
+    #[inline(always)]
+    pub fn artifact_size<A>(&self, artifact_kind: &str, artifact_name: A, size: u64)
```
jyn514, line 296: Why inline this?

michaelwoerister: All of these event recording methods are `inline(always)` because they are thin wrappers around a function call that is never inlined. See https://github.com/rust-lang/rust/pull/87404/files#diff-98f589d77d5a9fcfd42c01231da93e4b0604804c976923f0ba3a676bc6c07d79L202-L206.

So their code results in something like `if some_condition { call_function_that_is_never_inlined() }`. 

For `artifact_size` in particular it might not be so important because it isn't called very often. But I personally preferred it if all of these similar methods followed the same pattern.

rylev:
@jyn514 I will be getting to this soon. We first wanted to make sure our changes to measureme and the related tooling don’t break perf.rlo. This should be easy to test, I just haven’t found the time yet.

rylev:
This is good to go. I'd like to see this land, and then we can start working on rustc-perf to expose some of these stats in the perf.rlo UI.

wesleywiser (approved):
Just one question. r=me when you're ready to merge

PR #87408, opened 2021-07-23
Hide allocator details from TryReserveError

kornelski:
I think there's [no need for TryReserveError to carry detailed information](https://github.com/rust-lang/rust/issues/48043#issuecomment-825139280), but I wouldn't want that issue to delay stabilization of the `try_reserve` feature.

So I'm proposing to stabilize `try_reserve` with a `TryReserveError` as an opaque structure, and if needed, expose error details later.

This PR moves the `enum` to an unstable inner `TryReserveErrorKind` that lives under a separate feature flag. `TryReserveErrorKind` could possibly be left as an implementation detail forever, and the `TryReserveError` get methods such as `allocation_size() -> Option<usize>` or `layout() -> Option<Layout>` instead, or the details could be dropped completely to make try-reserve errors just a unit struct, and thus smaller and cheaper.

### DIFF
--- a/library/alloc/src/collections/mod.rs
+++ b/library/alloc/src/collections/mod.rs
@@ -58,7 +58,31 @@ use core::fmt::Display;
 /// The error type for `try_reserve` methods.
 #[derive(Clone, PartialEq, Eq, Debug)]
 #[unstable(feature = "try_reserve", reason = "new API", issue = "48043")]
-pub enum TryReserveError {
+pub struct TryReserveError {
+    kind: TryReserveErrorKind,
+}
+
+impl TryReserveError {
+    /// Details about the allocation that caused the error
+    #[inline]
+    #[unstable(
+        feature = "try_reserve_kind",
+        reason = "Uncertain how much info should be exposed",
+        issue = "48043"
+    )]
+    pub fn kind(&self) -> TryReserveErrorKind {
+        self.kind.clone()
+    }
+}
+
+/// Details of the allocation that caused a `TryReserveError`
+#[derive(Clone, PartialEq, Eq, Debug)]
+#[unstable(
+    feature = "try_reserve_kind",
+    reason = "Uncertain how much info should be exposed",
+    issue = "48043"
+)]
+pub enum TryReserveErrorKind {
     /// Error due to the computed capacity exceeding the collection's maximum
     /// (usually `isize::MAX` bytes).
     CapacityOverflow,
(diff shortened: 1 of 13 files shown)


library/alloc/src/collections/mod.rs
```diff
@@ -58,7 +58,31 @@ use core::fmt::Display;
...
+
+impl TryReserveError {
+    /// Details about the allocation that caused the error
+    #[inline]
+    #[unstable(
+        feature = "try_reserve_detail",
+        reason = "Uncertain how much info should be exposed",
+        issue = "48043"
+    )]
+    pub fn detail(&self) -> TryReserveErrorDetail {
```
kennytm, line 73: i think this is usually called "kind" (like [`std::io::Error::kind()`](https://doc.rust-lang.org/std/io/struct.Error.html#method.kind)).

kornelski: I have considered that, but `ErrorKind` is a field-less enum. Besides, I think this type could be removed entirely in favor of getters on the error itself, or just dropped completely to make the error type small and efficient.

kennytm:
🤔 i'll defer this to T-libs-api to consider.

r? @rust-lang/libs-api

library/alloc/src/collections/mod.rs
kornelski: Ok, I see there are more variations in `kind` methods in libstd, so I've renamed it anyway

yaahc (commented):
I don't love how every match against a TryReserveError needs an explicit `map_err` to get the kind now before matching, but as far as I can tell it wouldn't be backwards incompatible to switch the opaque struct back to an enum in the future to enable direct convenient matching once we're confident we know the representation we want internally.

+1 from me but I'll wait until after today's libs-api team meeting before approving since it's already nominated.

library/alloc/src/collections/mod.rs
```diff
@@ -81,12 +105,23 @@ pub enum TryReserveError {
     },
 }
 
+#[unstable(
+    feature = "try_reserve_kind",
+    reason = "Uncertain how much info should be exposed",
+    issue = "48043"
+)]
+impl From<TryReserveErrorKind> for TryReserveError {
+    fn from(kind: TryReserveErrorKind) -> Self {
```
the8472, line 114: This is a non-generic function, it probably should be `#[inline]`

kornelski: #87843

pnkfelix:
Just a quick heads up: For some reason, this PR seems to have regressed the html5ever-opt benchmark: [4.24% more instructions:u](https://perf.rust-lang.org/compare.html?start=db3cb435c1197ef3e3919c03b7f81ca8bffbd007&end=996ff2e0a0f911f52bb1de6bdd0cfd5704de1fc9&stat=instructions%3Au), and [8.16% more max-rss](https://perf.rust-lang.org/compare.html?start=db3cb435c1197ef3e3919c03b7f81ca8bffbd007&end=996ff2e0a0f911f52bb1de6bdd0cfd5704de1fc9&stat=max-rss).

I'm not sure why this PR would cause that change, and it seems *somewhat* isolated to html5ever-opt. Skimming over the [max-rss graphs](https://perf.rust-lang.org/index.html?start=2021-08-06&end=&absolute=true&stat=max-rss), we cna see that html5-ever-opt is somewhat noisy, but it seems like a ~5% bump in max-rss can be potentially attributed to this PR.

(At the same time, it isn't a *broad* regression across many benchmarks. So it probably isn't worth investing much time looking into this. I just wanted to post a heads up.)

the8472:
#87843 might address the regression.

PR #87419, opened 2021-07-24
IEEE 754 is not an RFC

programmerjake:
If there were something between `IEEE` and `754`, it would be `STD`: [`IEEE STD 754-2019`](https://doi.org/10.1109%2FIEEESTD.2019.8766229)

### DIFF
--- a/RELEASES.md
+++ b/RELEASES.md
@@ -44,7 +44,7 @@ Libraries
 - [`leading_zeros`, and `trailing_zeros` are now available on all
   `NonZero` integer types.][84082]
 - [`{f32, f64}::from_str` now parse and print special values
-  (`NaN`, `-0`) according to IEEE RFC 754.][78618]
+  (`NaN`, `-0`) according to IEEE 754.][78618]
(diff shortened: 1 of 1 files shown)


inquisitivecrystal:
Random drive-by comment: to me, the version with `STD` looks a bit more readable, although I don't really know what the convention is here.

programmerjake:
I've always seen it written IEEE 754 or IEEE 754-2019 (or whatever date your favorite version was released on).

sthagen:
> I've always seen it written IEEE 754 or IEEE 754-2019 (or whatever date your favorite version was released on).

In my experience you cite with the edition when you want to pin on that specific document and not intend to follow later editions when they come into effect later.

So I think keeping just `IEEE 754` Is safe.

Note that one can always visit the normative pages for most standards and consider using there preferred citation. In this case:
> "IEEE Standard for Floating-Point Arithmetic," in IEEE Std 754-2019 (Revision of IEEE 754-2008) , vol., no., pp.1-84, 22 July 2019, doi: 10.1109/IEEESTD.2019.8766229.

From https://ieeexplore.ieee.org/document/8766229

Issue #87379, opened 2021-07-22
Mutate arguments at std::process::Command


cehteh:
Feature gate: `#![feature(mutate_command_args)]`

This is a tracking issue for adding an API to mutate the arguments of std::process::Command.
Discussed earlier in https://internals.rust-lang.org/t/lack-of-api-mutating-args-at-std-command/14908

### Public API

```rust
// std::process::Command

pub fn args_clear(&mut self) -> &mut Command;

pub fn arg_set<S>(&mut self, index: usize, value: S) -> &mut Command
where
    S: AsRef<OsStr>;

// maybe for convenience (same as .args_clear().args(new_args)):
pub fn args_new<I, S>(&mut self, args: I) -> &mut Command
where
    I: IntoIterator<Item = S>,
    S: AsRef<OsStr>;
```

`arg_set()` may panic when the index is out of range, only existing elements get replaced.

### Steps / History

- [x] Implementation: 
  - [x] Unix Implementation
  - [x] Other Platforms
- [ ] Final comment period (FCP)
- [ ] Stabilization PR

### Unresolved Questions

- None yet.

cehteh:
Working on this feature at:

https://github.com/cehteh/rust/tree/mutate_command_args

cehteh:
closed in #87420

PR #87420, opened 2021-07-24
Mutate command args

cehteh:
Finished implementation, the 'windows' bits are untested by me, i hope they pass the CI. Otherwise, awaiting comments.

### DIFF
--- a/library/std/src/lib.rs
+++ b/library/std/src/lib.rs
@@ -306,6 +306,7 @@
 #![feature(min_specialization)]
 #![feature(mixed_integer_ops)]
 #![feature(must_not_suspend)]
+#![feature(mutate_command_args)]
 #![feature(needs_panic_runtime)]
 #![feature(negative_impls)]
 #![feature(never_type)]
--- a/library/std/src/process.rs
+++ b/library/std/src/process.rs
@@ -714,6 +714,48 @@ impl Command {
         self
     }
 
+    /// Reset the argument list to new content
+    ///
+    /// Allows to maintain the argument list in a Vec or anything else that can provide
+    /// a suitable iterator.
+    ///
+    /// # Examples
+    ///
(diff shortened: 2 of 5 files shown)


Mark-Simulacrum:
r? @joshtriplett - I feel like I saw some T-libs discussion of these APIs recently.

FWIW it'd be ideal to integrate more context into the PR description (presuming I'm recalling correctly that this context exists).

cehteh:
> r? @joshtriplett - I feel like I saw some T-libs discussion of these APIs recently.
> 
> FWIW it'd be ideal to integrate more context into the PR description (presuming I'm recalling correctly that this context exists).

I didn't know under what circumstances github makes automatic links/references. Usually I made this explicit, but just forgotten to do on the PR request itself.

First discussion is on: https://internals.rust-lang.org/t/lack-of-api-mutating-args-at-std-command/14908
For the implementation I opened a issue: https://github.com/rust-lang/rust/issues/87379

joshtriplett:
I think I remember this being discussed elsewhere, but I don't remember the outcome:
- I think `arg_set(0, ...)` should work on UNIX, and act equivalent to calling `arg0`, rather than asserting.
- I don't think we should add the `args_new` helper; I think `.args_clear().args(...)` seems clearer.

cehteh:
Had some f*ckup with submodule hashes slipped into the commit. Fixed now, ready to merge.

m-ou-se:
I'm not convinced we should have this at all. Even the existing `Command::get_args` was quite controversial, since we don't want to provide a full Vec-like interface on `Command`. If you want to flexibly work with arguments before running the Command, it's probably better to just use a regular Vec before passing it to Command::args().

@rust-lang/libs-api - Does anyone else think we should have this?

Dylan-DPC:
closing this based on the last comment

PR #86532, opened 2021-06-22
Make deleted code in a suggestion clearer

estebank:
Show suggestions that include deletions in a way similar to `diff`'s output.

For changes that do not have deletions, use `+` as an underline for additions and `~` as an underline for replacements.

For multiline suggestions, we now use `~` in the gutter to signal replacements and `+` to signal whole line replacements/additions.

In all cases we now use color to highlight the specific spans and snippets.

PR #87421, opened 2021-07-24
Do not discard `?Sized` type params and suggest their removal

estebank:


estebank:
Split off https://github.com/rust-lang/rust/pull/85799

r? @oli-obk

compiler/rustc_middle/src/ty/diagnostics.rs
```diff
@@ -130,6 +131,120 @@ pub fn suggest_constraining_type_param(
     if def_id == tcx.lang_items().sized_trait() {
         // Type parameters are already `Sized` by default.
         err.span_label(param.span, &format!("this type parameter needs to be `{}`", constraint));
+        // See if there's a `?Sized` bound that can be removed to suggest that.
+        match param.bounds {
+            [] => {}
```
oli-obk, line 136: instead of going two nesting levels in, I think you can also just bail out via

```rust
if param.bounds.is_empty() {
    return true;
}
```

oli-obk: Also, consider pulling out this entire new code block into a separate function.

src/test/ui/trait-bounds/unsized-bound.stderr
```diff
@@ -0,0 +1,234 @@
+error[E0277]: the size for values of type `B` cannot be known at compilation time
+  --> $DIR/unsized-bound.rs:2:12
+   |
+LL | trait Trait<A> {}
+   |             - required by this bound in `Trait`
+LL | impl<A, B> Trait<(A, B)> for (A, B) where A: ?Sized, B: ?Sized, {}
+   |         -  ^^^^^^^^^^^^^ doesn't have a size known at compile-time
```
oli-obk, line 7: This span is confusing to me. The trait in an impl block is not something that can have a size. I think we would want to point at `(A, B)` inside the `Trait<(A, B)>`.

src/test/ui/trait-bounds/unsized-bound.stderr
```diff
@@ -0,0 +1,234 @@
...
+   |         |
+   |         this type parameter needs to be `std::marker::Sized`
+   |
+   = note: required because it appears within the type `(A, B)`
+help: consider removing the `?Sized` bound to make the type parameter `Sized`
+   |
+LL | impl<A, B> Trait<(A, B)> for (A, B) where A: ?Sized, {}
+   |                                                   --
```
oli-obk, line 15: not this PRs problem, but we really should move to something that makes it clear it's not the two marked chars, but the removed chars between them

oli-obk (changes requested):

src/test/ui/trait-bounds/unsized-bound.stderr
estebank: Addressed in #86532

src/test/ui/trait-bounds/unsized-bound.stderr
estebank: For what is worth that's the same span we currently use:

```
error[E0277]: the size for values of type `B` cannot be known at compilation time
 --> src/test/ui/trait-bounds/unsized-bound.rs:2:12
  |
1 | trait Trait<A> {}
  |             - required by this bound in `Trait`
2 | impl<A, B> Trait<(A, B)> for (A, B) where A: ?Sized, B: ?Sized, {}
  |         -  ^^^^^^^^^^^^^ doesn't have a size known at compile-time
  |         |
  |         this type parameter needs to be `std::marker::Sized`
  |
  = note: required because it appears within the type `(A, B)`
help: consider relaxing the implicit `Sized` restriction
  |
1 | trait Trait<A: ?Sized> {}
  |              ^^^^^^^^
```

And @Aaron1011 has been working piecemeal on HIR based checks that could give us the span you want, IIRC.

oli-obk: gotcha!

compiler/rustc_hir/src/hir.rs
```diff
@@ -442,6 +442,7 @@ pub enum GenericBound<'hir> {
     Trait(PolyTraitRef<'hir>, TraitBoundModifier),
     // FIXME(davidtwco): Introduce `PolyTraitRef::LangItem`
     LangItemTrait(LangItem, Span, HirId, &'hir GenericArgs<'hir>),
+    Unsized(Span),
```
jackh726, line 445: I didn't have time to get to this. But I'm a bit concerned about this. Would like to have potentially seen this reuse `Trait` in some way

oli-obk: I brought this up in https://github.com/rust-lang/rust/pull/85799#discussion_r675629404 but then forgot about it again entirely. Sorry, should've paid more attention. I'm fine doing a quick revert and figuring all this out in a new PR

estebank: As you all prefer, but I am concerned about the size effects of adding a `Span` to `Trait`. @oli-obk, that comment was in a separate `enum`, to carry the span in the obligation, this is to carry it in the `HIR`.

oli-obk: `PolyTraitRef` already contains a span, can we re-use that?

estebank: Looking back at this, the reason I didn't want to just use the `Trait` variant were because we want this exclusively [for this](https://github.com/rust-lang/rust/pull/87421/files#diff-0a61b538a3cec072c76fecae4635af6a12ec3256860029ac70549c2aa53ab394R1448-R1456), as an "empty signal" and wanted to make sure we didn't try to evaluate these by accident (as they are already handled differently elsewhere). I _could_ have added a field to `Trait`, but I thought that would have caused the enum's size to increase (although that might not be the case, looking at the `LangItem` variant).

PR #86344, opened 2021-06-16
Split MaybeUninit::write into new feature gate and stabilize it

est31:
This splits off the `MaybeUninit::write` function from the `maybe_uninit_extra` feature gate into a new `maybe_uninit_write` feature gate and stabilizes it.

Earlier work to improve the documentation of the write function: #86220

Tracking issue: #63567

PR #87422, opened 2021-07-24
DOC: remove unnecessary feature crate attribute from example code

mgeier:
I'm not sure whether I fully understand the stabilization process (I most likely don't), but I think this attribute isn't necessary here, right?

This was recently stabilized in #86344.

### DIFF
--- a/library/core/src/mem/maybe_uninit.rs
+++ b/library/core/src/mem/maybe_uninit.rs
@@ -461,7 +461,6 @@ impl<T> MaybeUninit<T> {
     /// With `write`, we can avoid the need to write through a raw pointer:
     ///
     /// ```rust
-    /// #![feature(maybe_uninit_extra)]
     /// use core::pin::Pin;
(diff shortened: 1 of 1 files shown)


LeSeulArtichaut (approved):
Good catch :+1:

LeSeulArtichaut:
> I'm not sure whether I fully understand the stabilization process (I most likely don't), but I think this attribute isn't necessary here, right?

This feature attribute was actually added in #86344. I can imagine that the PR author tested their example in e.g. the playground and forgot to remove the feature gate afterwards. Also, if you want to know more about the stabilization process for library feature there is a chapter in the [std-dev-guide](https://std-dev-guide.rust-lang.org/feature-lifecycle/stabilization.html) that you can read.

est31:
> I can imagine that the PR author tested their example in e.g. the playground and forgot to remove the feature gate afterwards.

Yeah that's precisely what happened :rofl: . good catch @mgeier

PR #87427, opened 2021-07-24
get rid of NoMirFor error variant

RalfJung:
The only place where we throw that error, it is very quickly caught again and turned into a different error. So raise that other error immediately.


compiler/rustc_mir/src/const_eval/machine.rs
```diff
@@ -247,20 +249,7 @@ impl<'mir, 'tcx> interpret::Machine<'mir, 'tcx> for CompileTimeInterpreter<'mir,
...
-                    return Err(ConstEvalErrKind::NeedsRfc(format!(
-                        "calling extern function `{}`",
-                        path
-                    ))
-                    .into());
-                }
-                return Err(err);
-            }
-        }))
+        Ok(Some(ecx.load_mir(instance.def, None)?))
```
jyn514, line 252:
```suggestion
        ecx.load_mir(instance.def, None).map(Some)
```

RalfJung: Honestly I think my version of the code is easier to read -- no need to invoke a higher-order function like `map` here.

Issue #85016, opened 2021-05-07
`union` and other weak keywords are not highlighted correctly in rustdoc code blocks


ThePuzzlemaker:
This issue is quite similar to #80004, which I also worked on, so I have a general idea of how it might be able to be fixed.

I tried this code:
```rust
/// ```rust ignore
/// async fn quux() {} // rust 2018+ keyword, fixed in #80226
///
/// #[repr(C)]
/// union ExampleUnion { // weak keyword, not working
///     foo: u8,
///     bar: u16
/// }
///
/// fn bar () {
///     let mut a = 0;
///     abstract blah; // reserved keyword, working
/// }
pub fn foo() {}
```

I expected to see this happen: `union` is highlighted, even though it is a weak keyword, as it is acting as a keyword in that location in the code

Instead, this happened: They were not highlighted, but regular keywords including reserved ones and edition 2018+ ones (fixed in #80266) were.
![image](https://user-images.githubusercontent.com/12666617/117382519-53478e00-aea4-11eb-88e5-76816fd8bf21.png)

The issue is here, because `Ident::is_reserved` does not check for weak keyword status: https://github.com/rust-lang/rust/blob/8db7973f1952ab07c0ceb83cfb134f9bda518552/src/librustdoc/html/highlight.rs#L139-L147

### Meta

Tested 

`rustc --version --verbose`:
stable:
```
rustc 1.52.0 (88f19c6da 2021-05-03)
binary: rustc
commit-hash: 88f19c6dab716c6281af7602e30f413e809c5974
commit-date: 2021-05-03
host: x86_64-unknown-linux-gnu
release: 1.52.0
LLVM version: 12.0.0
```
nightly:
```
rustc 1.54.0-nightly (bacf770f2 2021-05-05)
binary: rustc
commit-hash: bacf770f2983a52f31e3537db5f0fe1ef2eaa874
commit-date: 2021-05-05
host: x86_64-unknown-linux-gnu
release: 1.54.0-nightly
LLVM version: 12.0.0
```

Backtrace is not applicable.

@rustbot claim
@rustbot modify labels: +A-rustdoc +A-rustdoc-ui +C-bug

ThePuzzlemaker:
[...]

PR #87428, opened 2021-07-24
Fix union keyword highlighting in rustdoc HTML sources

GuillaumeGomez:
I followed this logic: if I find an ident "union", I check if it followed by another ident or not. If it's the case, then I consider this is a keyword because it's declaring a union type.

To do so I created a new Iterator which allows to peek the next items without moving the current iterator position.

This is part of https://github.com/rust-lang/rust/issues/85016. If the fix makes sense, I'll extend it to other weak keywords (the issue only mentions they exist but https://doc.rust-lang.org/nightly/reference/keywords.html#weak-keywords only talks about `dyn` and `'static` so not sure if there is anything else to be done?).

cc @notriddle (you're one of the last ones who worked on this part of rustdoc so here you go :wink: )
r? @jyn514

### DIFF
--- a/src/librustdoc/html/highlight.rs
+++ b/src/librustdoc/html/highlight.rs
@@ -8,8 +8,8 @@
 use crate::html::escape::Escape;
 use crate::html::render::Context;
 
+use std::collections::VecDeque;
 use std::fmt::{Display, Write};
-use std::iter::Peekable;
 
 use rustc_lexer::{LiteralKind, TokenKind};
 use rustc_span::edition::Edition;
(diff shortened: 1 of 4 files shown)


src/librustdoc/html/highlight.rs
```diff
@@ -154,10 +154,57 @@ fn get_real_ident_class(text: &str, edition: Edition) -> Class {
     }
 }
 
+/// This iterator comes from the same idea than "Peekable" except that it allows to "peek" more than
+/// just the next item by using `peek_next`. The `peek` method always returns the next item after
+/// the current one whereas `peek_next` will return the next item after the last one peeked.
+///
+/// You can use both `peek` and `peek_next` at the same time without problem.
+struct PeekIter<'a> {
```
jyn514, line 162: There's no need to rewrite this. Just use `Peekable<Peekable>`.

jyn514:
Why is rustdoc rewriting its own rust parsing? Can't we use rustc_parser somehow?

src/librustdoc/html/highlight.rs
GuillaumeGomez: Doesn't it only allow to check the two next items?

jyn514: Oh I see, this allows peeking arbitrarily many times. I guess that's ok, it seems like it could be expensive to store all those tokens in memory at once though.

GuillaumeGomez: Unless you have a lot of whitespaces, the size of stored items should never go higher than 2. But just in case... :)

GuillaumeGomez:
To be confirmed but iirc, last time we checked it was doing too much: unnecessary tokens like whitespaces for example were removed. However, we need them, which is why we have to rely on the `rustc_lexer` instead.

GuillaumeGomez:
cc @camelid

src/librustdoc/html/highlight.rs
estebank: If the memory consumption becomes concerning in any case, then you could do the equivalent of "clone current iterator, advance N times, return that value and drop the cloned iterator" to get the same functionality at the cost of recomputing the value (advancing the iterator) every time... You could even add logic to either delegate to `Peekable<Peekable<T>>` or do the above depending on the N.

estebank:
> To be confirmed but iirc, last time we checked it was doing too much: unnecessary tokens like whitespaces for example were removed. However, we need them, which is why we have to rely on the `rustc_lexer` instead.

Could we modify the parser to _not_ discard the whitespaces somehow as a new mode of operation?

GuillaumeGomez:
Again: to be checked. But that'll make the parsing slower overall if the compiler has to discard them every time. We could always add a parameter to allow to keep them around. Not sure if that's something the people in charge of the parser would want?

estebank:
The two people most involved with the parser are @petrochenkov and myself. Personally, if there's no or low perf impact I'd prefer avoiding duplicated logic, and I'm expecting @petrochenkov's concern to be of maintainability. If it can be done cleanly, I would prefer going the single parser way.

GuillaumeGomez:
Alright, I'll try to do that as soon as I can!

notriddle:
Rust-Analyzer has [a parser that preserves spaces](https://github.com/rust-analyzer/rust-analyzer/blob/master/docs/dev/syntax.md), and already uses it for syntax highlighting.

But can we use it in rust-lang/rust right now?

GuillaumeGomez:
I'd prefer use the existing compiler tools rather than add external ones.

GuillaumeGomez:
So from what I could see on my `rustc_parse` check:

 * It doesn't keep the original formatting
 * It duplicates file content to parse it (maybe I didn't use the correct way to initiate the `Parser`?)
 * It removes code comments.

So I think we're forced to keep using the lexer for now...

GuillaumeGomez:
Did I miss anything?

notriddle (approved):

ehuss:
@bors retry
Issue with Windows LLDB

Issue #87411, opened 2021-07-23
FlatMap, Flatten appear to optimize badly


adrian17:
Context: I recently saw a perf regression caused by the following change:
```rs
pub fn pixels_rgba(&self) -> Vec<u8> {
    let mut output = Vec::new();
    for p in &self.pixels { output.extend_from_slice(&[p.red(), p.green(), p.blue(), p.alpha()]) }
    output
}
->
pub fn pixels_rgba(&self) -> Vec<u8> {
    self.pixels.iter().flat_map(|p| [p.red(), p.green(), p.blue(), p.alpha()]).collect();
}
```
The change was reasonable, with assumption that this would be more idiomatic and guarantee that the output Vec is preallocated. Unfortunately, this made the function several times slower. Recently-merged https://github.com/rust-lang/rust/pull/87168 helped here slightly, but the generated code is still much slower.

The same can also be observed for other code using `flat_map` or `flatten`, like a simple iteration over the iterator, and regardless of whether the flattened type has known size (array) or not.

I made an example benchmark in repo https://github.com/adrian17/flat_map_perf , with the following results on my machine (with today's nightly rustc):
```

tests::bench_array_4x500000_collect_loop               1,269,560 ns/iter (+/- 146,537)
tests::bench_array_4x500000_collect_loop_with_prealloc 1,255,140 ns/iter (+/- 165,287)
tests::bench_array_4x500000_collect_with_flat_map      2,697,082 ns/iter (+/- 303,411)

tests::bench_array_4x500000_iteration_nested_loop        220,838 ns/iter (+/- 25,307)
tests::bench_array_4x500000_iteration_flat_map         3,029,744 ns/iter (+/- 463,749)


tests::bench_iter_4000x500_collect_loop                  243,537 ns/iter (+/- 34,574)
tests::bench_iter_4000x500_collect_loop_with_prealloc    243,246 ns/iter (+/- 34,197)
tests::bench_iter_4000x500_collect_with_flatten        3,521,586 ns/iter (+/- 597,755)

tests::bench_iter_4000x500_iteration_nested_loop         290,939 ns/iter (+/- 34,414)
tests::bench_iter_4000x500_iteration_flatten           2,099,386 ns/iter (+/- 512,732)


tests::bench_iter_4x500000_collect_loop                3,124,601 ns/iter (+/- 444,296)
tests::bench_iter_4x500000_collect_loop_with_prealloc  2,873,051 ns/iter (+/- 576,719)
[...]

PR #87431, opened 2021-07-24
implement fold() on array::IntoIter to improve flatten().collect() perf

the8472:
With #87168 flattening `array::IntoIter`s is now `TrustedLen`, the `FromIterator` implementation for `Vec` has a specialization for `TrustedLen` iterators which uses internal iteration. This implements one of the main internal iteration methods on `array::Into` to optimize the combination of those two features.

This should address the main issue in #87411

```
# old
test vec::bench_flat_map_collect                         ... bench:   2,244,024 ns/iter (+/- 18,903)

# new
test vec::bench_flat_map_collect                         ... bench:     172,863 ns/iter (+/- 2,141)
```

### DIFF
--- a/library/core/src/array/iter.rs
+++ b/library/core/src/array/iter.rs
@@ -129,6 +129,11 @@ impl<T, const N: usize> Iterator for IntoIter<T, N> {
         Fold: FnMut(Acc, Self::Item) -> Acc,
     {
         let data = &mut self.data;
+        // FIXME: This uses try_fold(&mut iter) instead of fold(iter) because the latter
+        //  would go through the blanket `impl Iterator for &mut I` implementation
+        //  which lacks inline annotations on its methods and adding those would be a larger
+        //  perturbation than using try_fold here.
+        //  Whether it would be beneficial to add those annotations should be investigated separately.
         (&mut self.alive)
             .try_fold::<_, _, Result<_, !>>(init, |acc, idx| {
                 // SAFETY: idx is obtained by folding over the `alive` range, which implies the
--- a/library/core/src/array/iter.rs
+++ b/library/core/src/array/iter.rs
@@ -123,6 +123,22 @@ impl<T, const N: usize> Iterator for IntoIter<T, N> {
         (len, Some(len))
     }
 
+    #[inline]
+    fn fold<Acc, Fold>(mut self, init: Acc, mut fold: Fold) -> Acc
+    where
+        Fold: FnMut(Acc, Self::Item) -> Acc,
+    {
+        let data = &mut self.data;
+        (&mut self.alive)
+            .try_fold::<_, _, Result<_, !>>(init, |acc, idx| {
+                // SAFETY: idx is obtained by folding over the `alive` range, which implies the
+                // value is currently considered alive but as the range is being consumed each value
+                // we read here will only be read once and then considered dead.
+                Ok(fold(acc, unsafe { data.get_unchecked(idx).assume_init_read() }))
+            })
+            .unwrap()
+    }
+
     fn count(self) -> usize {
         self.len()
     }
--- a/library/alloc/benches/vec.rs
+++ b/library/alloc/benches/vec.rs
@@ -726,3 +726,9 @@ fn bench_dedup_old_100000(b: &mut Bencher) {
 fn bench_dedup_new_100000(b: &mut Bencher) {
     bench_vec_dedup_new(b, 100000);
 }
+
+#[bench]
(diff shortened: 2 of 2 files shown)


kennytm:
while this LGTM, shouldn't the original issue be addressed by implementing `SpecExtend<T, std::array::IntoIter<T>> for Vec<T, A>`?

the8472:
The original issue involved `Flatten` which results in several adapters sitting between `SpecExtend and the `IntoIter`.

library/core/src/array/iter.rs
```diff
@@ -123,6 +123,22 @@ impl<T, const N: usize> Iterator for IntoIter<T, N> {
...
+    {
+        let data = &mut self.data;
+        (&mut self.alive)
+            .try_fold::<_, _, Result<_, !>>(init, |acc, idx| {
+                // SAFETY: idx is obtained by folding over the `alive` range, which implies the
+                // value is currently considered alive but as the range is being consumed each value
+                // we read here will only be read once and then considered dead.
+                Ok(fold(acc, unsafe { data.get_unchecked(idx).assume_init_read() }))
+            })
+            .unwrap()
```
kennytm, line 139: can we call `fold` here instead of `try_fold`? 

```suggestion
        self.alive.fold(init, |acc, idx| {
            // SAFETY: idx is obtained by folding over the `alive` range, which implies the
            // value is currently considered alive but as the range is being consumed each value
            // we read here will only be read once and then considered dead.
            fold(acc, unsafe { data.get_unchecked(idx).assume_init_read() })
        })
```

the8472: I tried that, `array::IntoIter` has a Drop impl, so `alive` can't be move out, but that would be required to call `fold(self)`, that's why I used `try_fold instead`.

kennytm: @the8472 oops right.

`(&mut self.alive).fold(init, ...)` should work though.

the8472: Yeah but that would go through `impl Iterator for &mut I` which is less optimized.

kennytm: what about `std::mem::take(&mut self.alive).fold(...)`?

the8472: Alive is needed by the drop impl, so if the fold paniced then there would be a problem. Maybe it could be put back in place before the unwind tries to drop it, but that would require a drop guard which would make things more complicated instead of simpler.

kennytm: `(&mut x).fold` handles the panic just fine, with the same behavior as `x.try_fold`.

```rust
use std::ops::Range;

struct A {
    a: Range<usize>,
}

impl Drop for A {
    fn drop(&mut self) {
        dbg!(&self.a); // should print 5..7 instead of 0..7
    }
}

fn main() {
    let mut a = A { a: 0..7 };

    (&mut a.a).fold((), |(), i| {
        dbg!(i);
        if i == 4 {
            panic!();
        }
    });
}
```

the8472: > (&mut x).fold handles the panic just fine

My previous comment was replying to the `mem::take` suggestion which was edited away after I replied.

> self.alive is a std::ops::Range<usize> and AFAIK there is no special-cased implementation of fold or try_fold for Range<usize> nor &mut Range<usize>.

Ah, it was  `RangeInclusive` that had one but not `Range`. So yeah that might work. I'll give it a try.

kennytm: even if we do the `mem::take` it will just turn `self.alive` to `0..0` and then leaks everything which is safe 🙃  (compared with `self.alive.clone().fold(...)` which will cause double-free).

the8472: I have now tried `(&mut self.alive).fold` instead of `try_fold`, it undoes all perfomance gains. I guess somehow the indirection through `&mut` inhibits optimizations.

the8472: Which is due to this not being `#[inline]`

https://github.com/rust-lang/rust/blob/71a6c7c80398143a67e27b01412f4b2ec12bde8b/library/core/src/iter/traits/iterator.rs#L3474-L3478

kennytm: heh. 

wdyt should we just add the `#[inline]` or leave a FIXME comment explaining the performance regression if we use `fold` instead of `try_fold`? either way is fine for me.

the8472: I'll add a FIXME, changing inlining on such central methods can have mixed impact on compile time even if runtime performance is better, so that should be done on a separate PR.

PR #87441, opened 2021-07-24
Emit suggestion when passing byte literal to format macro

### LINKED ISSUE #86865: Improvement to write! macro with wrong type for string literal argument
Given the following code: 
https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=4c5ff2fc026afd1a666633cba4c31865

```rust
pub fn example(f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
    write!(f, b"Foo{}", "bar")
}
```

The current output is:

```
error: format argument must be a string literal
 --> src/lib.rs:4:15
   |
4 |     write!(f, b"Foo{}", "bar")
   |               ^^^^^^^^
   |
help: you might be missing a string literal to format with
   |
4 |     write!(f, "{} {}", b"Foo{}", "bar")
```

```
 error: format argument must be a string literal, **found byte literal**
  --> src/lib.rs:4:15
   |
 4 |     write!(f, b"Foo{}", "bar")
   |               ^^^^^^^^
   |
 help: you might be missing a string literal to format with
   |
4 |     write!(f, "{} {}", b"Foo{}", "bar")
 help: you may have meant to pass in a string literal instead of a byte literal
   |
 4 |     write!(f, b"Foo{}", "bar")
  |               ^ try removing this 'b'
4 |     write!(f, "Foo{}", "bar")
```

So this is super trivial, but I just ran into this and I haven't used `write!` in a while so I sorta did a double take and it took me a second to figure out why it was suggesting what it was suggesting.

I was just refactoring some code that had used a byte literal beforehand. Two things would help:

1. Telling me that I passed in a byte literal where a string literal was expected (or whatever else - right now it only says what's expected, not what was provided). IMO this alone would have been a big help and I would have spent 0 cycles thinking about what the problem could be.

2. Since this is, to a human, clearly a case where I intended the byte literal to be the format string, in an ideal world that would be pointed out to me. I realize that this is open to interpretation - is it "clearly" the case because it's a byte literal? Or a byte literal with {}? IDK! Food for thought.

ibraheemdev:
Closes #86865

### DIFF
--- a/compiler/rustc_expand/src/base.rs
+++ b/compiler/rustc_expand/src/base.rs
@@ -9,7 +9,7 @@ use rustc_ast::{self as ast, AstLike, Attribute, Item, NodeId, PatKind};
 use rustc_attr::{self as attr, Deprecation, Stability};
 use rustc_data_structures::fx::FxHashMap;
 use rustc_data_structures::sync::{self, Lrc};
-use rustc_errors::{DiagnosticBuilder, ErrorReported};
+use rustc_errors::{Applicability, DiagnosticBuilder, ErrorReported};
 use rustc_lint_defs::builtin::PROC_MACRO_BACK_COMPAT;
 use rustc_lint_defs::BuiltinLintDiagnostics;
 use rustc_parse::{self, nt_to_tokenstream, parser, MACRO_ARGUMENTS};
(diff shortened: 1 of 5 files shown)


compiler/rustc_expand/src/base.rs
```diff
@@ -1132,37 +1132,47 @@ impl<'a> ExtCtxt<'a> {
     }
 }
 
-/// Extracts a string literal from the macro expanded version of `expr`,
-/// emitting `err_msg` if `expr` is not a string literal. This does not stop
-/// compilation on error, merely emits a non-fatal error and returns `None`.
```
cjgillot, line 1137: Can you correct the comment instead of removing it?
For instance: what is the extra bool in the return type?

ibraheemdev: I didn't remove the comment, I moved it down to `expr_to_string`, which I think it was meant for. It actually emits an error and returns an option as the comment says.. should I document `expr_to_spanned_string` as well?

cjgillot: Yes, please.

cjgillot:
r=me with commits squashed

cjgillot:
> failed to get 200 response from `https://crates.io/api/v1/crates/dissimilar/1.0.2/download`, got 503

Seems spurious.
@bors retry

ehuss:
@bors retry

Windows image update has broken things.

PR #87445, opened 2021-07-25
Fix may not to appropriate might not or must not

amalik18:
I went through and changed occurrences of `may not` to be more explicit with `might not` and `must not`.


library/alloc/src/sync.rs
```diff
@@ -1056,7 +1056,7 @@ impl<T: ?Sized> Arc<T> {
     // Non-inlined part of `drop`.
     #[inline(never)]
     unsafe fn drop_slow(&mut self) {
-        // Destroy the data at this time, even though we may not free the box
+        // Destroy the data at this time, even though we might not free the box
         // allocation itself (there may still be weak pointers lying around).
```
kennytm, line 1060: i think the "may not" in line 1059 is a "must not"
(also change the "may" in line 1060 to "might"?)

library/core/src/marker.rs
```diff
@@ -316,7 +316,7 @@ pub trait StructuralEq {
 /// attempt to derive a `Copy` implementation, we'll get an error:
 ///
 /// ```text
-/// the trait `Copy` may not be implemented for this type; field `points` does not implement `Copy`
+/// the trait `Copy` must not be implemented for this type; field `points` does not implement `Copy`
```
kennytm, line 319: keep this unchanged, unless you're going to modify the error message of E0204 too.

```
error[E0204]: the trait `Copy` may not be implemented for this type
 --> src/lib.rs:1:10
  |
1 | #[derive(Copy, Clone)]
  |          ^^^^
2 | struct X {
3 |     a: &'static mut i32,
  |     ------------------- this field does not implement `Copy`
  |
  = note: this error originates in the derive macro `Copy` (in Nightly builds, run with -Z macro-backtrace for more info)
```

library/core/src/sync/atomic.rs
```diff
@@ -41,7 +41,7 @@
 //! instructions to implement `AtomicI8`. Note that this emulation should not
 //! have an impact on correctness of code, it's just something to be aware of.
 //!
-//! The atomic types in this module may not be available on all platforms. The
+//! The atomic types in this module must not be available on all platforms. The
```
kennytm, line 44:
```suggestion
//! The atomic types in this module might not be available on all platforms. The
```

amalik18:
@kennytm How do I view why it failed? I'm not able to pull up the build logs.

Issue #83255, opened 2021-03-18
rustdoc: --show-coverage shouldn't count tuple struct fields


taylordotfish:
rustdoc's `--show-coverage` flag counts public fields of tuple structs, but this doesn't really make sense, since `doc` attributes on such fields never appear in the generated documentation.

For example, given the following in `src/lib.rs`:

```rust
//! test

/// test
pub struct Test(pub u8);
```

`cargo rustdoc -- -Z unstable-options --show-coverage` reports:

```
+-------------------------------------+------------+------------+------------+------------+
| File                                | Documented | Percentage |   Examples | Percentage |
+-------------------------------------+------------+------------+------------+------------+
| src/lib.rs                          |          2 |      66.7% |          0 |       0.0% |
+-------------------------------------+------------+------------+------------+------------+
| Total                               |          2 |      66.7% |          0 |       0.0% |
+-------------------------------------+------------+------------+------------+------------+
```

If `src/lib.rs` is changed to:

```rust
//! test

/// test
pub struct Test(
    /// test
    pub u8,
);
```

then the coverage is reported as 100%, but, as expected, the documentation for the `pub u8` field isn't included in the documentation generated by `cargo doc`.

### Meta

```
rustc 1.52.0-nightly (f5d8117c3 2021-03-16)
binary: rustc
commit-hash: f5d8117c338a788bd24abec733fd143dfceb25a0
commit-date: 2021-03-16
host: powerpc64le-unknown-linux-gnu
release: 1.52.0-nightly
LLVM version: 12.0.0
```

camelid:
Tuple struct field docs will be rendered soon, assuming that #80320 merges.

GuillaumeGomez:
I guess we can now consider this issue to be fixed since #87451 was merged?

camelid:
@GuillaumeGomez that sounds right to me. Do you want to close it?

GuillaumeGomez:
Let's close it then. :)

PR #80320, opened 2020-12-23
[Rustdoc] Render tuple fields in structs correctly

RDambrosio016:
This PR makes rustdoc render public tuple fields correctly similar to normal fields, previously no rendering would happen and the description of the field was effectively discarded.

Before:
before

After:
after

I was thinking about coloring the `0` a different color, maybe the color of primitives so it stands out more, but id like to hear opinions on that first.

PR #87451, opened 2021-07-25
Add support for tuple struct field documentation

### LINKED ISSUE #42615: Rustdoc should pick up docs for tuple struct and enum fields
Rustdoc should expose doc comments written on tuple fields:

```rust
pub struct Foo(
    /// Hello docs
    String
);
```

Tagging @QuietMisdreavus @GuillaumeGomez based on conversation on IRC.

GuillaumeGomez:
Fixes  #42615.
This is #80320 updated to new codebase and with added tests.
Part of https://github.com/rust-lang/rust/issues/83255.

cc @camelid (since you were involved on the original PR).
r? @jyn514


jyn514:
This closes https://github.com/rust-lang/rust/issues/42615, right? I'm not sure how the --show-coverage issue is related.

src/test/rustdoc/tuple-struct-fields-doc.rs
```diff
@@ -0,0 +1,19 @@
...
+// @has - '//*[@id="structfield.3"]' '3: i8'
+// @has - '//*[@id="main"]/div[@class="docblock"]' 'not hello'
+pub struct Foo(
+    /// hello
+    pub u32,
+    char,
+    pub char,
+    /// not hello
+    pub i8,
+);
```
jyn514, line 19: Can you also test tuple variant fields?

```rust
pub enum Bar {
    BarVariant(
        /// Hello docs
        String
    ),
}
```

GuillaumeGomez: Sure.

GuillaumeGomez:
> This closes #42615, right? I'm not sure how the --show-coverage issue is related.

Absolutely! As for the --show-coverage issue: it's because it complains about missing documentation on tuple structs' fields which currently can't be documented. ;)

src/test/rustdoc/tuple-struct-fields-doc.rs
GuillaumeGomez: I cna confirm it's still not working in this PR, adding it!

GuillaumeGomez:
Added support for tuple structs in enums as well. Great catch @jyn514 !

src/librustdoc/clean/mod.rs
```diff
@@ -1700,7 +1700,14 @@ impl Clean<Item> for ty::VariantDef {
         let kind = match self.ctor_kind {
             CtorKind::Const => Variant::CLike,
             CtorKind::Fn => Variant::Tuple(
-                self.fields.iter().map(|f| cx.tcx.type_of(f.did).clean(cx)).collect(),
+                self.fields
+                    .iter()
+                    .map(|f| {
+                        let kind = StructFieldItem(cx.tcx.type_of(f.did).clean(cx));
+                        let what_rustc_thinks = Item::from_def_id_and_parts(f.did, None, kind, cx);
+                        Item { visibility: Visibility::Inherited, ..what_rustc_thinks }
```
jyn514, line 1708: Can you add a comment that this is true because it's an enum variant? At first I got it confused with `struct S(usize)` and was confused why it was always public.

src/librustdoc/clean/mod.rs
```diff
@@ -1730,9 +1737,7 @@ impl Clean<Variant> for hir::VariantData<'_> {
     fn clean(&self, cx: &mut DocContext<'_>) -> Variant {
         match self {
             hir::VariantData::Struct(..) => Variant::Struct(self.clean(cx)),
-            hir::VariantData::Tuple(..) => {
-                Variant::Tuple(self.fields().iter().map(|x| x.ty.clean(cx)).collect())
-            }
+            hir::VariantData::Tuple(..) => Variant::Struct(self.clean(cx)),
```
jyn514, line 1740: Why this change? Is `Tuple` ever used now? If not, can it be removed?

src/librustdoc/clean/types.rs
```diff
@@ -1933,7 +1933,7 @@ crate struct Enum {
 #[derive(Clone, Debug)]
 crate enum Variant {
     CLike,
-    Tuple(Vec<Type>),
+    Tuple(Vec<Item>),
```
jyn514, line 1936: Why is this a full Item? You only ever use the `ItemKind`.

src/librustdoc/html/render/print_item.rs
```diff
@@ -1509,7 +1514,9 @@ fn render_struct(
             if let Some(g) = g {
                 write!(w, "{}", print_where_clause(g, cx, 0, false),)
             }
-            w.write_str(";");
+            if structhead {
+                w.write_str(";");
+            }
```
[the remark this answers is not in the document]
GuillaumeGomez: Otherwise, for the tuple struct enum variant, we get:

```rust
enum Foo {
    Bar(u32);,
}
```

Which seems not what we want. I'll add a comment though.

src/librustdoc/clean/mod.rs
GuillaumeGomez: It is:

```
failures:
    [rustdoc] rustdoc/const-generics/const-generics-docs.rs
    [rustdoc] rustdoc/intra-doc/basic.rs
    [rustdoc] rustdoc/issue-85454.rs
    [rustdoc] rustdoc/synthetic_auto/overflow.rs
    [rustdoc] rustdoc/toggle-item-contents.rs
    [rustdoc] rustdoc/tuple-struct-fields-doc.rs
    [rustdoc] rustdoc/type-layout.rs
    [rustdoc] rustdoc/viewpath-rename.rs
    [rustdoc] rustdoc/viewpath-self.rs
    [rustdoc] rustdoc/where.rs
```

Leaving it then.

src/librustdoc/clean/types.rs
GuillaumeGomez: Oups, it's completely unneeded, you're absolutely right.

src/librustdoc/clean/mod.rs
GuillaumeGomez: Done!

src/librustdoc/html/render/print_item.rs
```diff
@@ -962,11 +962,13 @@ fn item_enum(w: &mut Buffer, cx: &Context<'_>, it: &clean::Item, e: &clean::Enum
...
                             write!(w, "{}(", name);
-                            for (i, ty) in tys.iter().enumerate() {
+                            for (i, item) in tys.iter().enumerate() {
                                 if i > 0 {
                                     w.write_str(",&nbsp;")
                                 }
-                                write!(w, "{}", ty.print(cx));
+                                if let clean::ItemKind::StructFieldItem(ref ty) = *item.kind {
+                                    write!(w, "{}", ty.print(cx));
+                                }
```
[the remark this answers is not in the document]
GuillaumeGomez: I'll add it, seems like a potential trap that we can avoid ahead of time.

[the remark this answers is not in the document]
GuillaumeGomez: Actually it got removed thanks to a suggestion from @jyn514 so that solves this issue. :)

GuillaumeGomez:
Updated!

GuillaumeGomez:
And of course I forgot rustfmt...

src/librustdoc/clean/mod.rs
camelid: Hmm. Where is it used, and why did you change this code from `Tuple` to `Struct`?

GuillaumeGomez: So, `Variant::Tuple` is used on tuple structs which are not in an enum. In case we are in an enum, the kind is provided by the `Variant` wrapper directly, and since we need the fields' name (yes, even for a tuple struct variant), I simply switched to that to make things simpler.

Do you want me to use `Variant::Tuple` anyway (meaning the code will require refactorings in multiple places just like I originally did) or do you prefer to keep things as is (and then I add documentation to explain this)? The first solution requires a lot of changes but nothing impossible (I already did it and then reverted it afterwards for this solution) but might improve code clarity. So as you prefer.

jyn514: Nah, this is fine - can you add a comment?

jyn514:
r=me with the comment added :)

src/librustdoc/clean/mod.rs
GuillaumeGomez: Sure!

src/librustdoc/clean/mod.rs
```diff
@@ -1730,9 +1730,13 @@ impl Clean<Variant> for hir::VariantData<'_> {
...
-            hir::VariantData::Tuple(..) => {
-                Variant::Tuple(self.fields().iter().map(|x| x.ty.clean(cx)).collect())
-            }
+            // Important note here: `Variant::Tuple` is used on tuple structs which are not in an
+            // enum (so where converting from `ty::VariantDef`). In case we are in an enum, the kind
+            // is provided by the `Variant` wrapper directly, and since we need the fields' name
+            // (even for a tuple struct variant!), it's simpler to just store it as a
+            // `Variant::Struct` instead of a `Variant::Tuple` (otherwise it would force us to make
+            // a lot of changes when rendering them to generate the name as well).
+            hir::VariantData::Tuple(..) => Variant::Struct(self.clean(cx)),
```
dsherret, line 1739: Did this cause https://github.com/rust-lang/rust/pull/87451/files#diff-384affc1b4190940f114f3fcebbf969e7e18657a71ef9001da6b223a036687d9R1739 ?

It seems like tuple variants no longer exist, but shouldn't tuple variants be described as tuple variants?

dsherret, line 1739: Did this cause #87887?

It seems like tuple variants no longer exist, but shouldn't tuple variants be described as tuple variants?

GuillaumeGomez: It's very likely yes.

Issue #83409, opened 2021-03-23
[ER] Uplift clippy::unnecessary_mut_passed


leonardo-m:
I think it's worth moving the clippy::unnecessary_mut_passed lint into rustc because it's a very basic coding mistake:

```rust
#![allow(unused_variables)]
fn foo(x: &u32) {}
fn bar(a: &[u32]) {}
fn spam(a: &[u32; 10]) {}
fn main() {
    foo(&mut 10);
    let mut a = [0_u32; 10];
    bar(&mut a[..]);
    spam(&mut a);
}
```

See also, for further improvements:
https://github.com/rust-lang/rust-clippy/issues/5546

PR #87453, opened 2021-07-25
Suggest removing unnecessary &mut as help message

### LINKED ISSUE #68697: Suggest removing `&mut x` when `x: &mut T` and `&mut T` is expected type
An example:
```rust
struct A;

fn bar(x: &mut A) {}

fn foo(x: &mut A) {
    bar(&mut x);
}
```
results in:
```rust
error[E0596]: cannot borrow `x` as mutable, as it is not declared as mutable
 --> src/lib.rs:6:9
  |
5 | fn foo(x: &mut A) {
  |        - help: consider changing this to be mutable: `mut x`
6 |     bar(&mut x);
  |         ^^^^^^ cannot borrow as mutable
```
but we could say:
```rust
error[E0596]: cannot borrow `x` as mutable, as it is not declared as mutable
 --> src/lib.rs:6:9
  |
5 | fn foo(x: &mut A) {
6 |     bar(&mut x);
  |         ^^^^^^ cannot borrow as mutable
  |         ---- help: remove the unnecessary `&mut` here
```

ibraheemdev:
Closes #68697

### DIFF
--- a/src/test/ui/nll/issue-51191.rs
+++ b/src/test/ui/nll/issue-51191.rs
@@ -1,13 +1,15 @@
+#![allow(unconditional_recursion)]
+
 struct Struct;
 
 impl Struct {
     fn bar(self: &mut Self) {
-        //~^ WARN function cannot return without recursing
         (&mut self).bar();
         //~^ ERROR cannot borrow `self` as mutable, as it is not declared as mutable [E0596]
+        //~^^ HELP try removing `&mut` here
     }
 
-    fn imm(self) {
+    fn imm(self) { //~ HELP consider changing this to be mutable
         (&mut self).bar();
         //~^ ERROR cannot borrow `self` as mutable, as it is not declared as mutable [E0596]
     }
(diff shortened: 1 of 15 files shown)


leonardo-m:
See also:
https://github.com/rust-lang/rust-clippy/issues/5546
https://github.com/rust-lang/rust/issues/83409

src/test/ui/nll/issue-51191.rs
```diff
@@ -1,13 +1,15 @@
+#![allow(unconditional_recursion)]
+
```
wesleywiser, line 2: Any reason we're allowing this lint? Seems like it's not really related to the rest of the changes.

ibraheemdev: An error about an unexcepted help message showed up related to the unconditional recursion after I added the help message changes, I can revert it and add the error annotation if you want.

wesleywiser: That would be great, thanks!

ibraheemdev: Fixed.

wesleywiser (approved):

wesleywiser:
Thanks!

@bors r+ rollup

Issue #43575, opened 2017-07-31
rustc aborts when there is a compilation failure on FreeBSD


dumbbell:
When there is a compilation failure, the Rust compiler aborts with the following output:

```
$ rustc invalid.rs
error: requires at least a format string argument
 --> invalid.rs:2:5
  |
2 |     print!()
  |     ^^^^^^^^
  |
  = note: this error originates in a macro outside of the current crate

error: aborting due to previous error(s)

fatal runtime error: failed to initiate panic, error 5
Abort trap (core dumped)
```

I used the following invalid source file:

```rust
fn main() {
    print!()
}
```

Here is the backtrace:

```
(gdb) r
Starting program: /var/ports/obj/mnt/home/dumbbell/Projects/freebsd/ports/SVN/lang/rust/work/rustc-1.19.0-src/build/x86_64-unknown-freebsd/stage2/bin/rustc invalid.rs
[New LWP 101467 of process 50322]
error: requires at least a format string argument
 --> invalid.rs:2:5
  |
2 |     print!()
  |     ^^^^^^^^
  |
  = note: this error originates in a macro outside of the current crate

error: aborting due to previous error(s)

fatal runtime error: failed to initiate panic, error 5

Thread 2 received signal SIGABRT, Aborted.
[Switching to LWP 101467 of process 50322]
0x0000000801be1b5a in thr_kill () from /lib/libc.so.7
(gdb) bt
#0  0x0000000801be1b5a in thr_kill () from /lib/libc.so.7
#1  0x0000000801be1b24 in __raise (s=6) at /mnt/home/dumbbell/Projects/freebsd/src/SVN/head/lib/libc/gen/raise.c:52
#2  0x0000000801be1a99 in abort () at /mnt/home/dumbbell/Projects/freebsd/src/SVN/head/lib/libc/stdlib/abort.c:65
#3  0x000000080184b5a1 in std::sys_common::util::abort::h1ff3fd43e3f3affb () from /var/ports/obj/mnt/home/dumbbell/Projects/freebsd/ports/SVN/lang/rust/work/rustc-1.19.0-src/build/x86_64-unknown-freebsd/stage2/bin/../lib/libstd-8923652fdd8a524a.so
[...]

PR #87457, opened 2021-07-25
freebsd remove compiler workaround.

devnexen:
related issue #43575


bjorn3:
Have you tested if the original problem no longer happens on FreeBSD? FreeBSD is a tier 2 platform, which means that tests are not run for it. It is only checked that it keeps building.

devnexen:
Yes the issue does not occur any longer.

PR #87459, opened 2021-07-25
use rust formatting in diagnostic issue template

### LINKED ISSUE #87293: Diagnostic issue template should have rust formatting for error messages too
This is kind of a meta issue, but the diagnostic issue template could be improved slightly. Currently it only uses ```` ```rust ```` for the main code block, but ```` ``` ```` for the error messages. This is technically correct but makes them harder to read, compare:

```
error[E0277]: expected a `FnMut<()>` closure, found `Cloner<{integer}>`
  --> src/main.rs:24:9
   |
19 | fn foo(mut producer: impl Producer<String>) {
   |                           ---------------- required by this bound in `foo`
...
24 |     foo(Cloner(8));
   |         ^^^^^^^^^ expected an `FnMut<()>` closure, found `Cloner<{integer}>`
   |
   = help: the trait `FnMut<()>` is not implemented for `Cloner<{integer}>`
   = note: wrap the `Cloner<{integer}>` in a closure with no arguments: `|| { /* code */ }`
   = note: required because of the requirements on the impl of `Producer<String>` for `Cloner<{integer}>`
```

```rust
error[E0277]: expected a `FnMut<()>` closure, found `Cloner<{integer}>`
  --> src/main.rs:24:9
   |
19 | fn foo(mut producer: impl Producer<String>) {
   |                           ---------------- required by this bound in `foo`
...
24 |     foo(Cloner(8));
   |         ^^^^^^^^^ expected an `FnMut<()>` closure, found `Cloner<{integer}>`
   |
   = help: the trait `FnMut<()>` is not implemented for `Cloner<{integer}>`
   = note: wrap the `Cloner<{integer}>` in a closure with no arguments: `|| { /* code */ }`
   = note: required because of the requirements on the impl of `Producer<String>` for `Cloner<{integer}>`
```
I think this is an easy change that should be an improvement for future Diagnostic issue reports.

ibraheemdev:
Closes #87293

### DIFF
--- a/.github/ISSUE_TEMPLATE/diagnostics.md
+++ b/.github/ISSUE_TEMPLATE/diagnostics.md
@@ -20,14 +20,14 @@ Given the following code: <!-- Please provide a link to play.rust-lang.org -->
 
 The current output is:
 
-```
+```rust
 <rustc output>
 ```
 
 <!-- The following is not always necessary. -->
 Ideally the output should look like:
 
-```
+```rust
 <proposed output>
 ```
 


Mark-Simulacrum:
Hmm, this doesn't feel quite right - the code blocks aren't intended to be rust code but general compiler output, which can be quite varied. So at least personally this feels like it could confuse filers (and particularly on diagnostics issues it's important to avoid that).

Cc @rust-lang/wg-diagnostics 
r? @estebank

oli-obk:
This is a general issue with rustdoc as well as github. Ideally we'd be able to paste formatted command line output with the formatting specifiers and have that render correctly, but I don't know if/how we could achieve that.

I agree with Mark, we should not do this. Can we add custom formatting for github code blocks to just the rust-lang repo? If so, we could invent a new `stderr` format and add some logic for formatting it.

estebank:
It's also true that the results aren't _terrible_. Sadly the `.stderr` format is not really designed with highlighting in mind, so the absolute best we could do is the output that the playground has today. Still better than nothing. We could [extend `linguist` for github to have highlighting for our output](https://github.com/github/linguist/blob/master/lib/linguist/languages.yml).

JohnTitor:
If it could cause confusion, it isn't "nothing", though I don't strongly disagree with it. I think a better improvement would be having our own config as y'all said, but to me, the current isn't also _terrible_.

Issue #60302, opened 2019-04-26
Split up files with `// ignore-tidy-filelength`


varkor:
These files are over 3,000 lines, which is not ideal for navigating or comprehending. If we can, it would be good to split these up. See also https://github.com/rust-lang/rust/issues/60015, which is a specific issue of this problem.

petrochenkov:
https://github.com/rust-lang/rust/issues/61097 is going to help with some of the files.

olamiko:
Which files @varkor? The files in #60015 are done

varkor:
I count [17 different files](https://github.com/rust-lang/rust/search?q=%2F%2F+ignore-tidy-filelength&unscoped_q=%2F%2F+ignore-tidy-filelength) that still have this comment.

alex-griffiths:
Hey, I'm not sure if anyone is working on this, but, this looks like it might be a good place to get started contributing.

I am quite new to the language, but I'd like to get involved. Is this something as simple as pulling out structs and other related blocks of code into their own files and then including them in the original?

varkor:
@alex-griffiths: yes, it should be as simple as figuring out some sensible way to split up large files into related types and implementations!

alex-griffiths:
Alright, mind if I take a crack at this?

varkor:
@alex-griffiths: yes, go ahead! You can claim the issue with:
> @rustbot claim

alex-griffiths:
@varkor I'm finding that a lot of the files that include the `// ignore-tidy-filelength` comment are files that include lots of the documentation.

I'm still kind of new to the language, but for example in src/libstd/fs.rs there are 6 structs where the struct definition itself is a single line, but with varying amounts of docs. Is this the kind of thing that is able to be extracted into it's own file which is then included as a module in the fs.rs file?

varkor:
@alex-griffiths: that's an interesting observation. I think in these cases, it's not so important to split up the files, because the length doesn't pose an issue for comprehension. Anyone editing the file is going to be interested in one or two particular functions, which are well-isolated within the file, and wouldn't benefit from being in their own file. In some sense, files like these are "false positives", though I don't think we could automatically exclude them. The lint is really more for files with large functions, and interdependency that we want to make easier to navigate. Thanks for pointing this out!

alex-griffiths:
I'll start working through all the files that have the comment in them, and I'll compile a list of files that appear to mostly documentation, and then for the rest, I'll start trying to improve the navigability of them.

[...]

PR #87462, opened 2021-07-25
Ignore comments in tidy-filelength

ibraheemdev:
Ref https://github.com/rust-lang/rust/issues/60302#issuecomment-652402127

### DIFF
--- a/compiler/rustc_hir/src/hir.rs
+++ b/compiler/rustc_hir/src/hir.rs
@@ -1,4 +1,3 @@
-// ignore-tidy-filelength
 use crate::def::{CtorKind, DefKind, Res};
 use crate::def_id::{DefId, CRATE_DEF_ID};
 crate use crate::hir_id::{HirId, ItemLocalId};
--- a/compiler/rustc_lint/src/builtin.rs
+++ b/compiler/rustc_lint/src/builtin.rs
@@ -1,5 +1,3 @@
-// ignore-tidy-filelength
-
 //! Lints in the Rust compiler.
 //!
 //! This contains lints which can feasibly be implemented as their own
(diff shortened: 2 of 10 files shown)


petrochenkov:
I disagree with this change, keeping comment lines short enough have the same value as keeping anything other text or code short.
We already have some logic in tidy applying exceptions to "special" comments like `//~ ERROR text` or comments containing URLs.

ibraheemdev:
> I think in these cases, it's not so important to split up the files, because the length doesn't pose an issue for comprehension. Anyone editing the file is going to be interested in one or two particular functions, which are well-isolated within the file, and wouldn't benefit from being in their own file. In some sense, files like these are "false positives", though I don't think we could automatically exclude them. The lint is really more for files with large functions, and interdependency that we want to make easier to navigate.

It's not about being short, it's about readability. Comments don't add complexity to the file.

Mark-Simulacrum:
Hm, @petrochenkov - I don't think this PR changes behavior for *line* length checking. Line lengths are already ignored for most of the codebase: we avoid nagging people in files under rustfmt control. Whether that was the right call, I'm not sure, but this PR seems unrelated to that topic.

This PR seems to solely intend to avoid nagging on *long* files if the length is primarily due to comment lines. I think that's an interesting proposal, but as I personally don't really typically find file length an impediment or good proxy, I'm not sure I'm the right reviewer.

Regardless, for issues like this I think it's likely best to circulate them more widely; I would suggest a T-compiler MCP.

Cc @rust-lang/libs-contributors as well, since many of our longer files are such due to extensive documentation in std, core, etc.

petrochenkov:
Ah, ok, sorry for misinterpreting the changes.
I'm ok with making files longer.

Mark-Simulacrum:
Discussed during T-compiler meeting; no objections to just moving ahead were raised. Seems trivial to re-add if opposition is raised.

An additional point of discussion was the check might be dropped entirely, but that can also be proposed in the future with little trouble. It's possible being happier to ignore the lint is preferable, though, in cases where it is causing pain for no good reason.

@bors r+

NoobProgrammer31:
can i start working on mod.rs first in first file ( i am new to it )

PR #87464, opened 2021-07-25
Remove unnecessary `structhead` parameter from `render_union`

camelid:
`structhead` is used for `render_struct` so that the logic for rendering
structs can be shared between struct variants and struct items. However,
`render_union` is not used anywhere except for rendering union items, so
its `structhead` parameter is unnecessary.


jyn514:
@bors r+ rollup

@camelid is there a reason you picked me to review instead of CraftSpider? They're pretty active :)

PR #87465, opened 2021-07-25
Simplify typeck/primary_body_of, fix comment to match return signature

audunhalland:
Hi, new contributor here! I'm carefully reading through the various modules just to learn. I noticed this function, `primary_body_of`, which has gone through a couple of refactors over time, adding new `Option`s to its returned tuple. Observations:

1. the `fn`'s documentation was not all up to date with the the current return signature.
2. `FnHeader` and `FnDecl` are always both `Some` or `None`. So I figured it might just return a reference to the full `hir::FnSig`, for simplicity and more precise typing. It's a pure refactor.

I'm learning better by working with code than just reading it, so here goes! If you want to avoid pure refactor PRs that don't really fix anything, I can revert the code change to only update the comment instead.

### DIFF
--- a/compiler/rustc_typeck/src/check/mod.rs
+++ b/compiler/rustc_typeck/src/check/mod.rs
@@ -257,9 +257,7 @@ fn adt_destructor(tcx: TyCtxt<'_>, def_id: DefId) -> Option<ty::Destructor> {
 }
 
 /// If this `DefId` is a "primary tables entry", returns
-/// `Some((body_id, header, decl))` with information about
-/// its body-id, fn-header and fn-decl (if any). Otherwise,
-/// returns `None`.
+/// `Some((body_id, body_ty, fn_sig))`. Otherwise, returns `None`.
 ///
 /// If this function returns `Some`, then `typeck_results(def_id)` will
 /// succeed; if it returns `None`, then `typeck_results(def_id)` may or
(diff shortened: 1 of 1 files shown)


compiler/rustc_typeck/src/check/mod.rs
```diff
@@ -269,32 +267,28 @@ fn adt_destructor(tcx: TyCtxt<'_>, def_id: DefId) -> Option<ty::Destructor> {
...
             }
             _ => None,
         },
         Node::ImplItem(item) => match item.kind {
-            hir::ImplItemKind::Const(ref ty, body) => Some((body, Some(ty), None, None)),
-            hir::ImplItemKind::Fn(ref sig, body) => {
-                Some((body, None, Some(&sig.header), Some(&sig.decl)))
-            }
+            hir::ImplItemKind::Const(ref ty, body) => Some((body, Some(ty), None)),
+            hir::ImplItemKind::Fn(ref sig, body) => Some((body, None, Some(&sig))),
```
eddyb, line 288: This is probably why the two halves were returned separately, I seem to recall a `MethodSig` existing that was different from `FnSig`.

Aha, found it! https://github.com/rust-lang/rust/commit/30d7279628b2e4e3f9dc7efd2a9d2d7018f48a5c followed by https://github.com/rust-lang/rust/commit/27511b22dfb08c98e53a8f672b2789d4de872f55 resulted in an unified `hir::FnSig`, whereas before there was only `hir::MethodSig`, without a `hir::FnSig`.

eddyb (approved):

PR #87467, opened 2021-07-26
Implement `RefUnwindSafe` for `Rc<T>`

### LINKED ISSUE #45924: Rc<T> does not implement RefUnwindSafe
`Arc<T>` implements both `UnwindSafe` and `RefUnwindSafe` (though maybe it only gets `RefUnwindSafe` because all of it's members happen to implement `RefUnwindSafe`).

Whereas `Rc<T>` only implements `UnwindSafe`. This seems like an oversight.

https://play.rust-lang.org/?gist=0d89773218a347569615e1bea7b352f6&version=nightly

inquisitivecrystal:
This PR implements `RefUnwindSafe` for `Rc<T>`, where `T: RefUnwindSafe`. 

This impl was omitted by an apparent oversight. `Rc<T>` already implements `UnwindSafe`. `Arc<T>` implements both `UnwindSafe` and `RefUnwindSafe`. There is no reason why an `&Rc<T>` is any less unwind safe than a `Rc<T>` or an `&Arc<T>`, so this should be safe to add.

Resolves #45924.

### DIFF
--- a/library/alloc/src/rc.rs
+++ b/library/alloc/src/rc.rs
@@ -318,6 +318,8 @@ impl<T: ?Sized> !marker::Sync for Rc<T> {}
 
 #[stable(feature = "catch_unwind", since = "1.9.0")]
 impl<T: RefUnwindSafe + ?Sized> UnwindSafe for Rc<T> {}
+#[stable(feature = "rc_ref_unwind_safe", since = "1.58.0")]
+impl<T: RefUnwindSafe + ?Sized> RefUnwindSafe for Rc<T> {}
 
 #[unstable(feature = "coerce_unsized", issue = "27732")]
 impl<T: ?Sized + Unsize<U>, U: ?Sized> CoerceUnsized<Rc<U>> for Rc<T> {}


inquisitivecrystal:
Rebased!

library/alloc/src/rc.rs
```diff
@@ -317,6 +317,8 @@ impl<T: ?Sized> !marker::Sync for Rc<T> {}
 
 #[stable(feature = "catch_unwind", since = "1.9.0")]
 impl<T: RefUnwindSafe + ?Sized> UnwindSafe for Rc<T> {}
+#[stable(feature = "rc_ref_unwind_safe", since = "1.56.0")]
```
camelid, line 320: (Just noting that this will likely require a version bump because there's a release next week.)

dtolnay (approved):

dtolnay:
This LGTM. `Rc` contains a `Cell` for the strong and weak reference counts which does not implement `RefUnwindSafe`, but all the public APIs on `Rc` are atomic w.r.t. panicking. There is no possibility that a panic during some operation on the `Rc` would leave the reference counts in an unintended state.

@rust-lang/libs-api
@rfcbot fcp merge

dtolnay:
Rfcbot needed redeploy: https://rust-lang.zulipchat.com/#narrow/stream/242791-t-infra/topic/rfcbot.20needs.20kick

@rfcbot fcp merge

inquisitivecrystal:
@dtolnay: this should be ready to merge, when you have a chance. I apologize for the delay.

dtolnay (approved):
Thanks!

rylev:
The regression is extremely surprising but only impacts one test case "cargo + doc + full". This is likely a fluke of some sort, but it is is concerning as the regressed test case shows a rather large regression (19x the significance threshold) which we would not expect from a non-noisy benchmark. The `get_auto_trait_impls` query seems to have been impacted though I'm not sure if this change would exercise that query. 

I'm going to leave this as a perf regression for now, but no action is needed from the author or reviewers. The perf team will try to investigate why this regression happened. 

cc @jyn514 @Mark-Simulacrum

jyn514:
> The get_auto_trait_impls query seems to have been impacted though I'm not sure if this change would exercise that query.

I'm only slightly surprised by this; get_auto_trait_impls has *horrific* algorithmic complexity and this presumably adds a bunch more types that implement RefUnwindSafe if they use Rc internally. There's a bunch of issues floating around, I'll try to find one.

jyn514:
This is somewhat related to https://github.com/rust-lang/rust/issues/90512 in that rustdoc does a bunch of global analyses the compiler never does.

PR #87468, opened 2021-07-26
Update rustfmt

calebcartwright:
Believe this gets everything back in order as both push and pull are working fine again. May do another small sync in the near future for my own sanity, but going forward will try to get on the same recurring cadence that clippy follows


klensy:
Is there some commits lost from rustfmt's master branch? For example pr https://github.com/rust-lang/rustfmt/pull/4857 got merged into master, but now exist only is stbu(?) branch.

calebcartwright:
@klensy - the commits here are intentional and are the ones I want included in this sync. No, nothing has been "lost" but there are some others that will get pulled in as part of the next sync instead.

PsiACE:
Hi @Mark-Simulacrum , would you like to take some time to review this PR?

Mark-Simulacrum:
There's generally no reason to ping me (or other reviewers) within such a short time window after the PR has been posted; we're all quite busy. This PR seems fine to me though - so going to go ahead and approve it.

@calebcartwright fyi - generally you can self-approve subtree bumps unless there's something in particular that needs taking a look at (in which case please call it out).

@bors r+

calebcartwright:
> @calebcartwright fyi - generally you can self-approve subtree bumps unless there's something in particular that needs taking a look at (in which case please call it out).

@Mark-Simulacrum - I actually don't have bors rights so my hands are a bit tied when it comes to approvals

PR #87469, opened 2021-07-26
2229: Don't capture preicese paths on top of a union

### LINKED ISSUE #87378: 2021 edition: Unsafe block not detected when accessing a union field inside a closure
### Code

I tried this code:

```rust
union Union {
    value: u64,
}

fn main() {
    let u = Union { value: 42 };
    
    || {
       unsafe { u.value }
    };
}
```

I expected to see this happen: no error

Instead, this happened: `error[E0133]: access to union field is unsafe and requires unsafe function or block

https://play.rust-lang.org/?version=nightly&mode=debug&edition=2021&gist=091d060959b30415131533f2e6d93f71

### Version it worked on
rustc 1.55.0-nightly 2018 edition
https://play.rust-lang.org/?version=nightly&mode=debug&edition=2018&gist=091d060959b30415131533f2e6d93f71

### Version with regression
rustc 1.55.0-nightly 2021 edition
https://play.rust-lang.org/?version=nightly&mode=debug&edition=2021&gist=091d060959b30415131533f2e6d93f71

arora-aman:
- Accessing fields of a union require unsafe block
- As part of 2229 we don't allow precision where we need an unsafe block
to capture.

Fixes: #87378

r? @nikomatsakis

### DIFF
--- a/compiler/rustc_typeck/src/check/upvar.rs
+++ b/compiler/rustc_typeck/src/check/upvar.rs
@@ -1760,12 +1760,11 @@ impl<'a, 'tcx> euv::Delegate<'tcx> for InferBorrowKind<'a, 'tcx> {
         self.borrow(assignee_place, diag_expr_id, ty::BorrowKind::MutBorrow);
     }
 }
-
-/// Truncate projections so that following rules are obeyed by the captured `place`:
+/// Truncate `place` so that an `unsafe` block isn't required to capture it.
 /// - No projections are applied to raw pointers, since these require unsafe blocks. We capture
 ///   them completely.
-/// - No Index projections are captured, since arrays are captured completely.
-fn restrict_capture_precision<'tcx>(mut place: Place<'tcx>) -> Place<'tcx> {
+/// - No projections are applied on top of Union ADTs, since these require unsafe blocks.
+fn restrict_precision_for_unsafe(mut place: Place<'tcx>) -> Place<'tcx> {
     if place.projections.is_empty() {
         // Nothing to do here
         return place;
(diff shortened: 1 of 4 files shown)


nikomatsakis (approved):

compiler/rustc_typeck/src/check/upvar.rs
```diff
@@ -1760,12 +1760,11 @@ impl<'a, 'tcx> euv::Delegate<'tcx> for InferBorrowKind<'a, 'tcx> {
         self.borrow(assignee_place, diag_expr_id, ty::BorrowKind::MutBorrow);
     }
 }
-
-/// Truncate projections so that following rules are obeyed by the captured `place`:
+/// Truncate `place` so that an `unsafe` block isn't required to capture it.
```
lqd, line 1763: Note the removed newline above this doc comment. (Don't worry about it, no need to block this one from landing. A drive-by fix in one of your next PRs is fine).

compiler/rustc_typeck/src/check/upvar.rs
```diff
@@ -1776,18 +1775,45 @@ fn restrict_capture_precision<'tcx>(mut place: Place<'tcx>) -> Place<'tcx> {
...
+        }
+    }
+
+    place
+}
+
+/// Truncate projections so that following rules are obeyed by the captured `place`:
+/// - No Index projections are captured, since arrays are captured completely.
+/// - No unsafe block is required to capture `place`
+/// Truncate projections so that following rules are obeyed by the captured `place`:
```
lqd, line 1803: And here the first line of the doc comment is duplicated as the last line.

PR #22074, opened 2015-02-08
Implement equality constraints in where clauses

jroesch:
This fixes #20041. I was thinking that the necessary machinery this required must be in place for associated types. So I went ahead and implemented the parsing and translation in collect when I was bored last night. 

I currently have syntax of the form: `where A = T` since it seemed cleaner, as well as consistent with the associated types syntax of `Foo<A=T>` while not introducing ambiguity. If that is not what we want I'm happy to use the `==` syntax originally present in the RFC.

r? @nikomatsakis 

I think this should cover everything let me know otherwise.

cc @darinmorrison

PR #87471, opened 2021-07-26
parse and handle `where T = U` predicate

JulianKnodt:
Updates #20041.

I took the most low-budget approach to adding equality checks (2 subtype checks in each direction), if it makes more sense to add in one check that does just equality, please let me know.

r? @varkor

### DIFF
--- a/compiler/rustc_feature/src/active.rs
+++ b/compiler/rustc_feature/src/active.rs
@@ -680,6 +680,9 @@ declare_features! (
     /// Allows `#[derive(Default)]` and `#[default]` on enums.
     (active, derive_default_enum, "1.56.0", Some(86985), None),
 
+    /// Allows `where T = U` in where predicates.
+    (active, type_equality_constraints, "1.56.0", Some(87533), None),
+
     // -------------------------------------------------------------------------
     // feature-group-end: actual feature gates
     // -------------------------------------------------------------------------
--- a/compiler/rustc_parse/src/parser/generics.rs
+++ b/compiler/rustc_parse/src/parser/generics.rs
@@ -284,7 +284,7 @@ impl<'a> Parser<'a> {
             }))
         // FIXME: Decide what should be used here, `=` or `==`.
         // FIXME: We are just dropping the binders in lifetime_defs on the floor here.
-        } else if self.eat(&token::Eq) || self.eat(&token::EqEq) {
+        } else if self.eat(&token::Eq) {
             let rhs_ty = self.parse_ty()?;
             Ok(ast::WherePredicate::EqPredicate(ast::WhereEqPredicate {
                 span: lo.to(self.prev_token.span),
@@ -293,7 +293,15 @@ impl<'a> Parser<'a> {
                 id: ast::DUMMY_NODE_ID,
             }))
         } else {
-            self.unexpected()
+            if self.token == token::EqEq {
+                let err = self.struct_span_err(
+                    lo.to(self.prev_token.span),
+                    "type equality predicates are `=` not `==`",
+                );
+                Err(err)
+            } else {
+                self.unexpected()
+            }
         }
     }
 
--- a/compiler/rustc_trait_selection/src/traits/error_reporting/mod.rs
+++ b/compiler/rustc_trait_selection/src/traits/error_reporting/mod.rs
@@ -698,6 +698,17 @@ impl<'a, 'tcx> InferCtxtExt<'tcx> for InferCtxt<'a, 'tcx> {
                         )
                     }
 
+                    ty::PredicateKind::TypeEquate(..) => {
+                        // Errors for `TypeEquate` predicates show up as
+                        // `SelectionError::FIXME(type_equality_constraints)`,
+                        // not `Unimplemented`.
+                        span_bug!(
+                            span,
+                            "type-equate requirement gave wrong error: `{:?}`",
+                            obligation
+                        )
+                    }
+
                     ty::PredicateKind::TypeWellFormedFromEnv(..) => span_bug!(
                         span,
                         "TypeWellFormedFromEnv predicate should only exist in the environment"
--- a/compiler/rustc_trait_selection/src/traits/fulfill.rs
+++ b/compiler/rustc_trait_selection/src/traits/fulfill.rs
@@ -377,6 +377,7 @@ impl<'a, 'b, 'tcx> FulfillProcessor<'a, 'b, 'tcx> {
                 | ty::PredicateKind::ClosureKind(..)
                 | ty::PredicateKind::Subtype(_)
                 | ty::PredicateKind::ConstEvaluatable(..)
+                | ty::PredicateKind::TypeEquate(..)
                 | ty::PredicateKind::ConstEquate(..) => {
                     let pred = infcx.replace_bound_vars_with_placeholders(binder);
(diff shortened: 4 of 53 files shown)


compiler/rustc_feature/src/active.rs
```diff
@@ -684,6 +684,9 @@ declare_features! (
     /// Allows `cfg(target_abi = "...")`.
     (active, cfg_target_abi, "1.55.0", Some(80970), None),
 
+    /// Allows `cfg(target_abi = "...")`.
```
PatchMixolydic, line 687: Doc comment needs to be updated here
```suggestion
    /// Allows for type equality constraints in `where` clauses.
```

fee1-dead:
I think making it an actual PredicateKind would be better. You can use `rustc_infer::infer::equate` to do this i think

varkor:
What's the status on the [previous blockers](https://github.com/rust-lang/rust/issues/20041#issuecomment-360768487)? Have these been addressed now?

JulianKnodt:
It's not so clear to me whether the previous blockers are completed, but given the time that lapsed, I wouldn't be surprised if they have. I still need to clean up some of the branches (since I just threw them all in last night), and will explore if this is feasible.

compiler/rustc_feature/src/active.rs
```diff
@@ -680,6 +680,9 @@ declare_features! (
     /// Allows `#[derive(Default)]` and `#[default]` on enums.
     (active, derive_default_enum, "1.56.0", Some(86985), None),
 
+    /// Allows `where T = U` in where predicates.
+    (active, where_pred_equality, "1.56.0", Some(87533), None),
```
varkor, line 684:
```suggestion
    (active, type_equality_constraints, "1.56.0", Some(87533), None),
```

compiler/rustc_parse/src/parser/generics.rs
```diff
@@ -284,7 +284,7 @@ impl<'a> Parser<'a> {
             }))
         // FIXME: Decide what should be used here, `=` or `==`.
         // FIXME: We are just dropping the binders in lifetime_defs on the floor here.
-        } else if self.eat(&token::Eq) || self.eat(&token::EqEq) {
```
varkor, line 287: This would be better replaced with a diagnostic suggesting the user uses `=` instead.

varkor (commented):
This could do with more tests (both successful and error cases). Also, could you add tests for the cases @nikomatsakis mentions in https://github.com/rust-lang/rust/pull/22074#issuecomment-73678356?

varkor:
Since @nikomatsakis reviewed the original PR (https://github.com/rust-lang/rust/pull/22074) that was closed, I'm going to reassign to them.

r? @nikomatsakis

compiler/rustc_trait_selection/src/traits/object_safety.rs
```diff
@@ -302,6 +302,13 @@ fn predicate_references_self(
             // possible alternatives.
             if data.projection_ty.substs[1..].iter().any(has_self_ty) { Some(sp) } else { None }
         }
+        ty::PredicateKind::TypeEquate(lhs, rhs) => {
+            if lhs == self_ty || rhs == self_ty {
+                Some(sp)
+            } else {
+                None
```
JulianKnodt, line 309: TODO: this is wrong, need to walk the types and check for self

compiler/rustc_trait_selection/src/traits/fulfill.rs
```diff
@@ -491,6 +492,16 @@ impl<'a, 'b, 'tcx> FulfillProcessor<'a, 'b, 'tcx> {
                     }
                 }
 
+                ty::PredicateKind::TypeEquate(lhs, rhs) => {
+                    match self.selcx.infcx().can_eq(obligation.param_env, lhs, rhs) {
+                        Ok(()) => ProcessResult::Changed(vec![]),
+                        Err(e) => ProcessResult::Error(FulfillmentErrorCode::CodeSubtypeError(
+                            ExpectedFound::new(false, lhs, rhs),
```
JulianKnodt, line 499: TODO create a new fulfillment error code

compiler/rustc_trait_selection/src/traits/error_reporting/mod.rs
```diff
@@ -698,6 +698,17 @@ impl<'a, 'tcx> InferCtxtExt<'tcx> for InferCtxt<'a, 'tcx> {
...
 
+                    ty::PredicateKind::TypeEquate(..) => {
+                        // Errors for `TypeEquate` predicates show up as
+                        // `SelectionError::FIXME(where_pred_equality)`,
+                        // not `Unimplemented`.
+                        span_bug!(
+                            span,
+                            "type-equate requirement gave wrong error: `{:?}`",
+                            obligation
+                        )
```
JulianKnodt, line 709: Not entirely sure what this was referencing, I suspect it's fine to leave this?

compiler/rustc_typeck/src/collect.rs
```diff
@@ -2294,8 +2294,14 @@ fn gather_explicit_predicates_of(tcx: TyCtxt<'_>, def_id: DefId) -> ty::GenericP
                 }))
             }
 
-            hir::WherePredicate::EqPredicate(..) => {
-                // FIXME(#20041)
+            hir::WherePredicate::EqPredicate(where_eq_pred) => {
+                let hir::WhereEqPredicate { hir_id: _, span, lhs_ty, rhs_ty } = where_eq_pred;
+                let lhs = icx.to_ty(&lhs_ty);
+                let rhs = icx.to_ty(&rhs_ty);
+                // this is a cheap approximation of equality, by checking if both are subtypes
```
JulianKnodt, line 2301: TODO delete old comment of me being lazy.
Here we can also do something like check that both types aren't params, which doesn't do much for now.

src/librustdoc/clean/mod.rs
```diff
@@ -327,6 +327,7 @@ impl<'a> Clean<Option<WherePredicate>> for ty::Predicate<'a> {
             ty::PredicateKind::RegionOutlives(pred) => pred.clean(cx),
             ty::PredicateKind::TypeOutlives(pred) => pred.clean(cx),
             ty::PredicateKind::Projection(pred) => Some(pred.clean(cx)),
+            ty::PredicateKind::TypeEquate(..) => todo!(),
```
JulianKnodt, line 330: TODO not entirely sure whether to make this `None`, or have something else

nikomatsakis:
I don't think we really want this sort of where-clause. As I've come to learn more about the limits of various logics, introducing equality checks implies the ability to say things that we probably don't want to say (although I would like to support them someday, but only after we've made a transition to chalk and decided to do it). For example, being able to write `fn foo<A, B>() where A = B` would be really nifty, but we don't have the infrastructure to support that kind of equality properly, and it moves the logic from one class of logics to another (in particular we would no longer map so directly to 1st order hereditary harrop clauses).

JulianKnodt:
That's certainly fair, equality tends to imply `!=` which seems to be quite controversial.
I was hoping to add this to improve specialization, so that you can provide `impl<T: Iterator> Foo for T where T = Map/Other<...>`, but given the additional burdens it's understandable.
In the future if/when this future is desired, this PR can be revived but I'll close it for now.

Please let me know if there is additional work elsewhere I can do to help move specialization along!

PR #87472, opened 2021-07-26
Stabilize and document `--force-warn`

### LINKED ISSUE #86516: Stabilize "force warn" option
We need https://github.com/rust-lang/rust/issues/85512 to be stabilized so that `cargo fix` can use it.

inquisitivecrystal:
This PR will stabilize and document the `--force-warn` command line option. It is currently a draft, pending an FCP. 

I've taken the liberty of tidying up the lint level command line options a bit as part of this. The changes are quite minor and should only affect rustc's help output. I'm making them here because they're trivial and, in one case, necessary to unify the way `--force-warn` with the way the other options are displayed.

I also want to mention that @rylev has done a ton of work on moving this along and deserves most of the credit. I'm just the one who landed up writing this particular PR.

Resolves #86516.


src/doc/rustc/src/lints/levels.md
```diff
@@ -57,6 +58,15 @@ warning: unused variable: `x`
   = note: to avoid this warning, consider using `_x` instead
 ```
 
+## force-warn
```
steveklabnik, line 61: this is in no way an actual blocker, but the lack of symmetry in naming between "forbid" and "force-warn" is kinda sad. Maybe I should open a proposal for "force-deny", heh.

steveklabnik (commented):
I am obviously not the right person to review this, but looks great to me :)

inquisitivecrystal:
Just caught a potential doc bug. If I'm reading things correctly, [force-warn may not be affected by capping the lint level](https://github.com/rust-lang/rust/blob/32c9b7b091534f6d80e7e85da0cd425acb6c9a79/compiler/rustc_middle/src/lint.rs#L114-L120). I need to look into this and possibly update the documentation.

inquisitivecrystal:
The reason I haven't updated the documentation is that I'm discussing whether this behavior should be changed.

m-ou-se:
What's the status of this?

inquisitivecrystal:
It's basically waiting on consensus in #86516. Specifically, there are two open change proposals.

1. Mark-Simulacrum has proposed to not allow `--force-warn warnings`. 
2. I've proposed making `--cap-lints` apply even to force-warn lints. 

I can add whichever combination of those changes is desired to this PR. If I had to guess, it would be to go ahead with Mark's but not with mine. However, it would be nice to have some conformation that's what people want.

After that, the last remaining thing to do is the FCP.

src/doc/rustc/src/lints/levels.md
joshtriplett: force-deny would have a slightly different meaning: it would deny things in dependencies even if cap-lints is set.

Mark-Simulacrum:
I believe this should be ready to go, just going to wait on CI and then r+ -- I don't think we need to wait for FCP to formally end in #86516, which should happen in ~3 days. We can always back this out if it becomes necessary.

PR #87473, opened 2021-07-26
Notify the Rust 2021 edition working group in zulip of edition bugs

rylev:
Notifying the group of these issues will make it easier for us to track them. 

r? @jyn514


inquisitivecrystal:
I don't think you got the syntax right. For the first two lines, I think you want something like this:

```toml
[notify-zulip."A-edition-2021"]
required_labels = ["C-bug"]
```

PR #87475, opened 2021-07-26
Remove caching from `OpaqueTypeExpander`. Fixes #87450.

### LINKED ISSUE #87450: Exponential compile time with recursive opaque type
### Code

I tried this code:

```rust
fn bar() -> impl Fn() {
    wrap(wrap(wrap(wrap(foo()))))
}

fn foo() -> impl Fn() {
    wrap(wrap(wrap(wrap(wrap(wrap(wrap(foo())))))))
}

fn wrap(f: impl Fn()) -> impl Fn() {
    move || f()
}
```

I expected running `cargo check` to produce an error like:
```
error[E0720]: cannot resolve opaque type
 --> src/lib.rs:5:13
  |
5 | fn foo() -> impl Fn() {
  |             ^^^^^^^^^ recursive opaque type
6 |     ((wrap(wrap(wrap(wrap(wrap(foo())))))))
  |     --------------------------------------- returning here with type `impl Fn<()>`
...
9 | fn wrap(f: impl Fn()) -> impl Fn() {
  |                          --------- returning this opaque type `impl Fn<()>`
```

It does, but `cargo check` takes time exponential in the number of `wrap` calls, although doesn't appear to use any extra memory.
In `foo` the time take looks like:
| Num `wraps`s | Time taken (s) |
|-|-|
| 4 | 0.5 |
| 5 | 2.1 |
| 6 | 28 |
| 7 | 436 |

I believe it's also exponential in the number of `wrap` calls in `bar`.

The reason I had code like this was because I was using nom's parser combinators which results in lots of nested closures. I spotted it because this caused rust-analyzer to hang because cargo clippy hung (it looks like check was the issue), I had to kill clippy to fix it.

### Bisection results
I bisected this (with cargo-bisect-rustc) to find the regression in `nightly-2019-10-16`, which contained the following bors commits:
1. Auto merge of #65433 - Centril:rollup-rzvry15, r=Centril
1. Auto merge of #65454 - tmandry:rollup-0k6jiik, r=tmandry
1. Auto merge of #65450 - Manishearth:clippyup, r=Manishearth
1. Auto merge of #65172 - tanriol:explain_borrow-use-context-dominators, r=nagisa
1. Auto merge of #65445 - ehuss:update-cargo-books, r=alexcrichton

@rustbot modify labels: +regression-from-stable-to-stable -regression-untriaged

hkratz:
Draft for discussion and to run tests and timing tests.

### DIFF
--- a/compiler/rustc_middle/src/ty/util.rs
+++ b/compiler/rustc_middle/src/ty/util.rs
@@ -554,8 +554,6 @@ struct OpaqueTypeExpander<'tcx> {
     // that type, and when we finish expanding that type we remove the
     // its DefId.
     seen_opaque_tys: FxHashSet<DefId>,
-    // Cache of all expansions we've seen so far. This is a critical
-    // optimization for some large types produced by async fn trees.
     primary_def_id: Option<DefId>,
     found_recursion: bool,
     /// Whether or not to check for recursive opaque types.
--- a/compiler/rustc_middle/src/ty/util.rs
+++ b/compiler/rustc_middle/src/ty/util.rs
@@ -11,7 +11,7 @@ use crate::ty::{self, DebruijnIndex, DefIdTree, List, Ty, TyCtxt, TypeFoldable};
 use rustc_apfloat::Float as _;
 use rustc_ast as ast;
 use rustc_attr::{self as attr, SignedInt, UnsignedInt};
-use rustc_data_structures::fx::{FxHashMap, FxHashSet};
+use rustc_data_structures::fx::FxHashSet;
 use rustc_data_structures::stable_hasher::{HashStable, StableHasher};
 use rustc_errors::ErrorReported;
 use rustc_hir as hir;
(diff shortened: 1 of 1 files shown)


compiler/rustc_middle/src/ty/util.rs
```diff
@@ -557,7 +556,6 @@ struct OpaqueTypeExpander<'tcx> {
     seen_opaque_tys: FxHashSet<DefId>,
     // Cache of all expansions we've seen so far. This is a critical
     // optimization for some large types produced by async fn trees.
```
bjorn3, line 558:
```suggestion
```

jyn514:
@rust-timer queue

jyn514:
Oh, finally found it.

@rust-timer build 8c29eb63f4dc13a3a8c06c8eacc2b437574158ec

hkratz:
That would have been too easy.

PR #87478, opened 2021-07-26
Point to where clause for GATs to add bound

### LINKED ISSUE #84931: GAT: confusing diagnostic when missing where clause
Given the following code: [playground](https://play.rust-lang.org/?version=nightly&mode=debug&edition=2018&gist=a493199249734d2ce196cfc20b148245)

```rust

trait StreamingIter {
    type Item<'a> where Self: 'a;
    fn next<'a>(&'a mut self) -> Option<Self::Item::<'a>>;
}

struct StreamingSliceIter<'a, T> {
    idx: usize,
    data: &'a mut [T],
}

impl<'b, T: 'b> StreamingIter for StreamingSliceIter<'b, T> {
    type Item<'a> = &'a mut T;
    fn next(&mut self) -> Option<&mut T> {
        self.idx += 1;
        Some(&mut self.data[self.idx - 1])
    }
}
```

The current output is:

```
error[E0309]: the parameter type `T` may not live long enough
  --> src/lib.rs:14:5
   |
13 | impl<'b, T: 'b> StreamingIter for StreamingSliceIter<'b, T> {
   |          -- help: consider adding an explicit lifetime bound...: `T: 'a +`
14 |     type Item<'a> = &'a mut T;
   |     ^^^^^^^^^^^^^^^^^^^^^^^^^^ ...so that the reference type `&'a mut T` does not outlive the data it points at

error: aborting due to previous error;
```

I'm not really sure what the correct output should be but it should mention that I forgot the `where Self: 'a` bound on the trait impl like I have it written in [this compiling code](https://play.rust-lang.org/?version=nightly&mode=debug&edition=2018&gist=23e8da2ef3ca912edf78263618843a15). It should also not suggest adding adding a `T: 'a` bound because that bound is not writeable :sweat_smile:

jackh726:
Fixes #84931

r? @estebank

### DIFF
--- a/compiler/rustc_infer/src/infer/error_reporting/mod.rs
+++ b/compiler/rustc_infer/src/infer/error_reporting/mod.rs
@@ -2259,9 +2259,99 @@ impl<'a, 'tcx> InferCtxt<'a, 'tcx> {
                 }
             };
 
-        let mut err = match *sub {
-            ty::ReEarlyBound(ty::EarlyBoundRegion { name, .. })
-            | ty::ReFree(ty::FreeRegion { bound_region: ty::BrNamed(_, name), .. }) => {
+        #[derive(Debug)]
+        enum SubOrigin<'hir> {
+            GAT(&'hir hir::Generics<'hir>),
+            Impl(&'hir hir::Generics<'hir>),
+            Trait(&'hir hir::Generics<'hir>),
+            Fn(&'hir hir::Generics<'hir>),
+            Unknown,
+        }
+        let sub_origin = 'origin: {
+            match *sub {
+                ty::ReEarlyBound(ty::EarlyBoundRegion { def_id, .. }) => {
+                    let node = self.tcx.hir().get_if_local(def_id).unwrap();
+                    match node {
+                        Node::GenericParam(param) => {
+                            for h in self.tcx.hir().parent_iter(param.hir_id) {
+                                break 'origin match h.1 {
+                                    Node::ImplItem(hir::ImplItem {
+                                        kind: hir::ImplItemKind::TyAlias(..),
+                                        generics,
+                                        ..
+                                    }) => SubOrigin::GAT(generics),
+                                    Node::ImplItem(hir::ImplItem {
+                                        kind: hir::ImplItemKind::Fn(..),
+                                        generics,
+                                        ..
+                                    }) => SubOrigin::Fn(generics),
+                                    Node::TraitItem(hir::TraitItem {
+                                        kind: hir::TraitItemKind::Type(..),
+                                        generics,
+                                        ..
+                                    }) => SubOrigin::GAT(generics),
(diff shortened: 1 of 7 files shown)


compiler/rustc_driver/src/lib.rs
```diff
@@ -1293,9 +1293,9 @@ pub fn init_env_logger(env: &str) {
         .with_indent_lines(true)
         .with_ansi(color_logs)
         .with_targets(true)
-        .with_wraparound(10)
-        .with_verbose_exit(true)
-        .with_verbose_entry(true)
+        //.with_wraparound(10)
```
lqd, line 1296: did you want to comment these lines out or was it a temporary change ?

jackh726: Was temporary (though I really should just make a PR for this; I find any time I do debugging, I remove these).

compiler/rustc_infer/src/infer/error_reporting/mod.rs
```diff
@@ -2242,9 +2242,99 @@ impl<'a, 'tcx> InferCtxt<'a, 'tcx> {
...
+                    "{} may not live long enough",
+                    labeled_user_string
+                );
+                let pred = format!("{}: {}", bound_kind, sub);
+                let suggestion = format!(
+                    "{} {}",
+                    if !generics.where_clause.predicates.is_empty() { "," } else { " where" },
+                    pred,
+                );
+                err.span_suggestion(
```
estebank, line 2325: Let's make the suggestion a bit easier to read, given that it points at a span right in the middle of a larger span:

```suggestion
                err.span_suggestion_verbose(
```

estebank: I leave it to your decision what to do here.

estebank:
r=me after cleaning up the commented out code.

compiler/rustc_infer/src/infer/error_reporting/mod.rs
jackh726: So, now that the associated type points to the type itself, not the whole defintion, they don't overlap anymore.

I do think the verbose suggestion would be nice, but I'll leave that as a followup.

src/test/ui/generic-associated-types/issue-86787.rs
```diff
@@ -0,0 +1,40 @@
+#![feature(generic_associated_types)]
```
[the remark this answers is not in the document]
jackh726: So...no. The suggestion here doesn't actually result in code that compiles, because it causes the where clauses not to match the trait's where clauses. (In fact, in this case, we *can't* write this where clause in the trait, we would have to use `Self`)

Unfortunately, that's a whole separate issue with GATs. But applying the current suggestion should help, in the future, to guide the user towards matching the trait's where clauses.

Issue #82064, opened 2021-02-13
Move collection of items accessible directly or through reexports from `rustc_privacy` to `rustc_resolve`


petrochenkov:
Directly accessible item `AccessLevel::Public` is an item with a fully public path from a crate root making it nameable from other crates.
`AccessLevel::Exported` item is an item with a fully public chain of reexports or modules making it nameable from other crates.
All such items can be found either by visiting the crate repeatedly and marking new and new items as `Public`/`Exported` until no items can be marked anymore, or doing the same thing by maintaining an item queue and pushing/popping things to/from it for processing until the queue is empty.

Sets of directly accessible and reexported items are build in `rustc_privacy` (`EmbargoVisitor`).
The problem is that information about reexport chains is lost at that point, so the produced sets do not contain all the necessary items.
`fn update_visibility_of_intermediate_use_statements` hack tries to mitigate that issue a bit, but it doesn't work correctly in presence of glob imports (see PR https://github.com/rust-lang/rust/pull/57922 for more details).

What we need to do is to move the logic for determining `AccessLevel::{Public,Exported}`s to `rustc_resolve` where all the information about reexport chains is still available.
That logic should run either as a part of `fn finalize_imports` or after it.
It should build a table `NodeId` -> `AccessLevel` in `Resolver`, and that table should later move to `struct ResolverOutputs` and then to `GlobalCtxt` like all other `ResolverOutputs` fields.

Later in `rustc_privacy` the resolver-based table should be used as a seed for `EmbargoVisitor` which will then assign type-based accessibility levels like `AccessLevel::Reachable` etc.
`fn update_visibility_of_intermediate_use_statements` will need to be removed.

petrochenkov:
cc @jyn514 
More precise `AccessLevel` table will fix some things for rustdoc, IIRC we've been talking about this on Zulip or some PR.

jyn514:
Is this meant to fix https://github.com/rust-lang/rust/issues/64762? That would let me simplify parts of the `doctree` pass I think. The visibility issue we talked about was https://github.com/rust-lang/rust/pull/77820#issuecomment-727174134 (`tcx.visibility()` returns semantic visibility, even if the `pub` keyword isn't accepted), which I'm not sure this would help this.

petrochenkov:
>Is this meant to fix #64762?

Yes.

>The visibility issue we talked about was #77820 (comment)

Nah, that's not what I meant. I think we briefly talked about #64762 or related issues at some point.

[...]

Issue #64762, opened 2019-09-25
Lint `unreachable_pub` got easily confused by multiple items in a single `pub use ...` statement


imp:
This code ([playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=6f6d3006c4df35712474dc8fa6ab8442)) trips `unreachable_pub` lint (while being completely legit).

```rust
#![deny(unreachable_pub)]

pub use self::m1::{Item1, Item2};

mod m1 {
    pub use self::m2::{Item1, Item2};
    
    mod m2 {
        pub struct Item1;
        pub struct Item2;
    }
}
```

However, if we break the top-level `pub use ...` statement into two separate statements ([playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=d48e646d8acdb673edf2d2f55e91e982)) - `unreachable_pub` lint convinces itself that the code is, indeed, legit.

```rust
#![deny(unreachable_pub)]

pub use self::m1::Item1;
pub use self::m1::Item2;

mod m1 {

    pub use self::m2::{Item1, Item2};
    
    mod m2 {
        pub struct Item1;
        pub struct Item2;
    }
}
```

jhpratt:
Running into this myself. Right now I'm just `#[allow]`ing it with a link to this issue.

If someone could lead me in the right direction, I'll give fixing this a shot.

HeroicKatora:
It also gets confused if the export happens via a glob, also on a single item.

```
#![deny(unreachable_pub)]

mod m1 {
    pub mod m2 {
        pub use super::Item1;
    }
    
    pub struct Item1;
}

pub use m1::m2::*;
```

jyn514:
Looks like this is because of https://github.com/rust-lang/rust/pull/57922#discussion_r251234202

lambinoo:
Since I claimed it 8 days ago, just a follow up.
I've been able to break the compiler a few times but I'm getting somewhere I think...
Not to a proper solution tho, just more mess, but that means more knowledge!

tl;dr: I haven't given up!

petrochenkov:
How to fix this: https://github.com/rust-lang/rust/issues/82064.

lambinoo:
Closed by #87487

PR #87487, opened 2021-07-26
Fixes wrong unreachable_pub lints on nested and glob public reexport

lambinoo:
Linked issues: #64762 & #82064


lambinoo:
I've had one test weirdly failing locally in the `debuginfo-gdb` pass but I'm not too sure how it can be related to my change

petrochenkov:
What I wanted to achieve with #82064 is assigning all `Public` and `Exported` levels in resolve, to all items, not only exports.
The `EmbargoVisitor::access_levels` table will be pre-initialized from the resolver table.
(After that we'll be able to add an assert to `EmbargoVisitor::update` making sure that only `Reachable` and `ReachableFromImplTrait` levels are inserted at that point.)

lambinoo:
@petrochenkov Hey, good evening! I've been able to look at the necessary work during the week. Would that basically involve moving a big chunk of the code  from the `EmbargoVisitor` in `rustc_privacy` to a new `PrivacyVisitor` in `rustc_resolve`?

I already have some local changes towards that goal but I prefer to ask before sinking too much time in the wrong direction

petrochenkov:
Most of `EmbargoVisitor` is about `Reachable` and `ReachableFromImplTrait` so there's not much to move if you want to support `Public` and `Exported` only.
Only modules (in resolve sense, including enums and traits) need to be treated specially, all other items don't need to be visited in detail at all.

lambinoo:
@petrochenkov Mmmh, I've noticed what I think is a major blocker.
https://github.com/lambinoo/rust/blob/I-64762_unreachable_pub_lint/compiler/rustc_privacy/src/lib.rs#L589:L591

As far as I know, I cannot correctly resolve the access level of `Impl` contained inside a `Block` from the `Resolver`. I need access to a `TyCtxt` for that. Current workarounds I found are:
- being overly optimistic (and mark every `pub` items as Public) and rely on the types being private to throw errors
- keeping a  `Public/Exported` access level pass in `rustc_privacy` explicitely for `Impl`.

My last push contains those changes (I still need to see how to handle macros so it's not finished yet)

petrochenkov:
Ah, right, impls can also be public or exported.
You are right, you cannot correctly process impls early in resolve, the logic for impls will need to stay in `rustc_privacy`.

compiler/rustc_resolve/src/lib.rs
```diff
@@ -1521,6 +1533,101 @@ impl<'a> Resolver<'a> {
...
+        // as an item can be defined in a private module
+        // and exported later on
+        //
+        // mod foo {
+        //     pub enum Hungry;
+        // }
+        //
+        // pub mod bar {
+        //     pub use super::foo::*;
+        // }
```
petrochenkov, line 1554: I'm pretty sure it should be a loop and not a double pass because cases like
```rust
mod privmod1 {
    pub struct Orig;
}

mod privmod2 {
    pub use foo::Orig as Reexport1;
}

pub pubmod {
    pub use privmod2::Reexport1 as Reexport2;
}
```
(with an arbitrary number of private modules) are possible.

compiler/rustc_resolve/src/lib.rs
```diff
@@ -1521,6 +1533,101 @@ impl<'a> Resolver<'a> {
...
+        visit::walk_crate(&mut privacy_visitor, krate);
+
+        tracing::info!("resolve::access_levels: {:#?}", self.access_levels);
+    }
+
+    fn set_exports_access_level(&mut self, module_id: LocalDefId) {
+        if let Some(exports) = self.export_map.get(&module_id) {
+            let pub_exports = exports
+                .iter()
+                .filter(|ex| ex.vis == Visibility::Public)
```
petrochenkov, line 1566: `pub` exports are only `Public`/`Exported` if the `module_id` module is `Public`/`Exported` respectively, I don't see this condition here.

compiler/rustc_resolve/src/access_levels.rs
```diff
@@ -0,0 +1,128 @@
...
+
+use crate::Resolver;
+
+pub struct PrivacyVisitor<'r, 'a> {
+    r: &'r mut Resolver<'a>,
+    prev_level: Option<AccessLevel>,
+}
+
+impl<'r, 'a> PrivacyVisitor<'r, 'a> {
+    pub fn new(r: &'r mut Resolver<'a>, root_level: Option<AccessLevel>) -> PrivacyVisitor<'r, 'a> {
```
petrochenkov, line 17: Nit: the root level is a constant, no need to pass it here.

compiler/rustc_resolve/src/access_levels.rs
```diff
@@ -0,0 +1,128 @@
...
+        PrivacyVisitor { r, prev_level: root_level }
+    }
+}
+
+impl<'r, 'ast> Visitor<'ast> for PrivacyVisitor<'ast, 'r> {
+    fn visit_item(&mut self, item: &'ast ast::Item) {
+        let inherited_item_level = match item.kind {
+            // FIXME Is this the correct behavior for those macros?
+            ast::ItemKind::MacCall(..) | ast::ItemKind::MacroDef(..) => None,
```
petrochenkov, line 26: `MacCall` should be impossible here, you can panic if it appears.
`MacroDef` is a regular item so it should be in the list below.

compiler/rustc_resolve/src/access_levels.rs
```diff
@@ -0,0 +1,128 @@
...
+
+            // Resolved in privacy when hir is available
+            ast::ItemKind::Impl(..) => return,
+
+            ast::ItemKind::ForeignMod(..) => self.prev_level,
```
petrochenkov, line 31: A comment was lost here.
Could you keep all the comments from `rustc_privacy`?

compiler/rustc_resolve/src/access_levels.rs
```diff
@@ -0,0 +1,128 @@
...
+                return;
+            }
+        }
+
+        let orig_level = std::mem::replace(&mut self.prev_level, access_level);
+        visit::walk_item(self, item);
+        self.prev_level = orig_level;
+    }
+
+    fn visit_block(&mut self, block: &'ast ast::Block) {
```
petrochenkov, line 123: In this visitor we can avoid visiting blocks entirely.
To detect public and exported items we only need to recurse into `ItemKind::Mod` items, in all other cases we can return without calling `visit::walk_item`.

compiler/rustc_resolve/src/access_levels.rs
```diff
@@ -0,0 +1,128 @@
...
+                for nested in trait_kind.4.iter() {
+                    self.r.set_access_level(nested.id, access_level);
+                }
+            }
+            ast::ItemKind::Impl(ref impl_kind) => {
+                for nested in impl_kind.items.iter() {
+                    if impl_kind.of_trait.is_some() || nested.vis.kind.is_pub() {
+                        self.r.set_access_level(nested.id, access_level);
+                    }
+                }
```
petrochenkov, line 97: This is dead code, `ItemKind::Impl` is unreachable here.

compiler/rustc_resolve/src/lib.rs
```diff
@@ -1521,6 +1533,101 @@ impl<'a> Resolver<'a> {
...
+            let module = self.get_module(module_id.to_def_id());
+
+            for export in pub_exports.into_iter() {
+                if let Some(export_def_id) = export.res.opt_def_id().and_then(|id| id.as_local()) {
+                    self.set_access_level_def_id(export_def_id, Some(AccessLevel::Exported));
+                }
+
+                if let Some(ns) = export.res.ns() {
+                    let key = self.new_key(export.ident, ns);
+                    let name_res = self.resolution(module, key);
```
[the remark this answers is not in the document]
lambinoo: Now that you mention it, it is a pretty convoluted way of getting what I want.
I was mostly trying to find a way to get to what I want and never really looked back once it worked once

lambinoo:
@petrochenkov Hey, good morning! Sorry for going MIA for a bit, life kinda caught up to me.

I think I got the final version of this PR down, I won't be able to work too much on this for the next few weeks so I won't be able to rework the `new_key` + `resolution` part that could be better nor work on the macro but I do plan on following up later on.

lambinoo:
@jyn514 Oh, yeah... Resolving conflicts from the phone using the github interface was probably a bad idea

compiler/rustc_middle/src/ty/mod.rs
```diff
@@ -123,6 +124,8 @@ pub struct ResolverOutputs {
     pub definitions: rustc_hir::definitions::Definitions,
     pub cstore: Box<CrateStoreDyn>,
     pub visibilities: FxHashMap<LocalDefId, Visibility>,
+    // FIXME: We might want this table to be tracked for incremental compilation
```
petrochenkov, line 127: This comment is no longer relevant.

compiler/rustc_resolve/src/lib.rs
```diff
@@ -1506,6 +1517,9 @@ impl<'a> Resolver<'a> {
     pub fn resolve_crate(&mut self, krate: &Crate) {
         self.session.time("resolve_crate", || {
             self.session.time("finalize_imports", || ImportResolver { r: self }.finalize_imports());
+            self.session.time("resolve_access_levels", || {
+                PrivacyVisitor::compute_access_levels(self, krate)
```
petrochenkov, line 1521:
```suggestion
                AccessLevelsVisitor::compute_access_levels(self, krate)
```
(Missing part of https://github.com/rust-lang/rust/pull/87487#discussion_r696577654)
Also the timer name on the line above wasn't updated.

compiler/rustc_resolve/src/access_levels.rs
```diff
@@ -0,0 +1,234 @@
...
+            }
+        }
+    }
+
+    /// Set the given binding access level to `AccessLevel::Public` and
+    /// sets the rest of the `use` chain to `AccessLevel::Exported` until
+    /// we hit the actual exported item.
+    /// The binding has to be from an export.
```
petrochenkov, line 84: Could you add an assert making sure that it's indeed from an export?

petrochenkov:
@lambinoo 
Could you squash commits after addressing the remaining comments?

compiler/rustc_resolve/src/access_levels.rs
lambinoo: Not too sure how to have an assert that fits into one line... I moved it to a closure inside directly `set_exports_access_level` to make it clearer

lambinoo:

@petrochenkov I think I got right this time. I also removed the usage of `new_key` by just building one directory... I like overlooking simple solutions it seems.

Also, thanks a lot for your time the past month for this PR! I will be able to work on the macro part at the end of september mostly

petrochenkov:
Thanks!
@bors r+

lambinoo:
Ooh... Yeah.. need to push stability fixes for every architecture and every platform.. not just x86 Unix

JohnCSimon:
Ping from triage: 
@lambinoo Can you please address the merge conflicts and test failures? Thank you.

JohnCSimon:
Ping from triage:
@lambinoo Can you please address the merge conflicts and test failures? Thank you.

jyn514:
@matthiaskrgr it was already unapproved, bors is just buggy.

lambinoo:
Hey! I plan on coming back to this PR and finish it... Sorry for the very long delay, I had a bunch of personal issues coming up but everything is back to normal!

lambinoo:
@petrochenkov @jyn514 Took me a while but I think I finally got it :D 
I'm hopping it won't break again on other architectures like last time.

petrochenkov:
Let's make sure this doesn't regress performance too much.
@bors try @rust-timer queue

jyn514:
@lambinoo I would probably just allow the lint in rustfmt for now - cc @calebcartwright, you might want to delete some of those fields it mentions.

calebcartwright:
> @lambinoo I would probably just allow the lint in rustfmt for now - cc @calebcartwright, you might want to delete some of those fields it mentions.

I'm planning on doing a subtree sync later today so I can try to knock these out too. Realize this hasn't been proposed but just wanted to ask that we don't make any modifications to rustfmt as part of this to reduce the risk of conflicts (which are a real pain with the subtree syncs)

lambinoo:
@calebcartwright Promise I won't touch rustfmt! Just let me know when you are done so I can rebase this :D

calebcartwright:
> @calebcartwright Promise I won't touch rustfmt! Just let me know when you are done so I can rebase this :D

Depending on your time sensitivity, I'd actually echo the suggestion from @jyn514 to go ahead allow/skip the lint for rustfmt. That's because it looks like it'll probably take a few more days for the sync as there's some other work I need to get landed in both repos before proceeding. Otherwise if your willing to wait a bit then happy to let you know when a rebase should resolve things here

lambinoo:
> > @calebcartwright Promise I won't touch rustfmt! Just let me know when you are done so I can rebase this :D
> 
> Depending on your time sensitivity, I'd actually echo the suggestion from @jyn514 to go ahead and allow/skip the lint for rustfmt. That's because it looks like it'll probably take a few more days for the sync as there's some other work I need to get landed in both repos before proceeding. Otherwise if you're willing to wait a bit then happy to let you know when a rebase should resolve things here
> 
> (edit - grammar correction)

If it makes your work harder, I'd rather wait honestly. I have a lot of available time, especially with Christmas approaching very fast! ( on that, Merry Christmas to everyone that sees this! ).

petrochenkov:
I suggest pushing the version adding `#![allow(unreachable_pub)]` to rustfmt sooner, because we didn't even run benchmarks on this PR yet, who knows what they will show.
Maybe it's a regression and you spend the time until rustfmt updates addressing https://github.com/rust-lang/rust/pull/87487#discussion_r685265544 to reduce it.

petrochenkov:
Perf is fine.
r=me after squashing commits.

lambinoo:
I started to rework that weird way I get `NameBinding`s but I think it's better if we get this through first before I need to rebase with master and fix new problems that could arise. I will make a new PR afterwards

petrochenkov:
>-S-waiting-on-author +S-waiting-on-review

Nothing changed since the previous review?
This PR still needs green CI and commit squashing.

lambinoo:
@petrochenkov  Oh... I thought I squashed all those commits down.. my bad. Squashed, pushed! 

Thanks a lot for your time as always!

calebcartwright:
By the way you should be able to rebase now to pull in the rustfmt updates

petrochenkov:
I cannot r+ this until the failing test on CI is fixed.
The test is a clippy test - `src\tools\clippy\tests\ui\implicit_hasher.rs`, implementation of the corresponding lint is in `src\tools\clippy\clippy_lints\src\implicit_hasher.rs` (looks like it indeed uses `access_levels`).

petrochenkov:
Clippy tests can be run with `./x.py test --bless --stage 2 src/tools/clippy`.

petrochenkov:
@lambinoo 
Has something changed?
I see that the clippy errors on CI are gone, but I don't see any changes in clippy.

lambinoo:
I noticed a small mistake in the rustc_pricacy code. A line disappeared when I shuffled some code around.

I forgot to update the access level of the impl in visit_item

petrochenkov:
Great, hopefully it will pass now.
@bors r+

lambinoo:
Success! Gosh this one has been cooking for a while. I'll be working on improving the whole resolution thing now but I think the two linked issues can be closed now!

PR #87488, opened 2021-07-26
Track caller of Vec::remove()

kornelski:
`vec.remove(invalid)` doesn't print a helpful source position:

> thread 'main' panicked at 'removal index (is 99) should be < len (is 1)', **library/alloc/src/vec/mod.rs:1379:13**


dtolnay (approved):
Nice!

PR #87499, opened 2021-07-27
Remove ASCII fast path from `rustc_lexer::{is_id_continue, is_id_start}`

ibraheemdev:
`unicode_xid` now has a fast path built-in: https://github.com/unicode-rs/unicode-xid/commit/122b38775cf077570025c8294280f57d2004096f

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5409,9 +5409,9 @@ dependencies = [
 
 [[package]]
 name = "unicode-xid"
-version = "0.2.1"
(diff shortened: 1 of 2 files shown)


dtolnay (changes requested):
The unicode-xid dependency in Cargo.lock needs to be upgraded in order for https://github.com/unicode-rs/unicode-xid/pull/20 to be pulled in.

dtolnay (approved):

PR #87501, opened 2021-07-27
Remove min_type_alias_impl_trait in favor of type_alias_impl_trait

spastorino:
r? @oli-obk


compiler/rustc_ast_passes/src/feature_gate.rs
```diff
@@ -276,27 +276,6 @@ impl<'a> PostExpansionVisitor<'a> {
...
-    fn check_impl_trait(&self, ty: &ast::Ty) {
-        struct ImplTraitVisitor<'a> {
-            vis: &'a PostExpansionVisitor<'a>,
-        }
-        impl Visitor<'_> for ImplTraitVisitor<'_> {
-            fn visit_ty(&mut self, ty: &ast::Ty) {
-                if let ast::TyKind::ImplTrait(..) = ty.kind {
-                    gate_feature_post!(
-                        &self.vis,
-                        min_type_alias_impl_trait,
```
oli-obk, line 290: We should not remove the feature gate check, instead it should check `type_alias_impl_trait` now, otherwise this PR insta-stabilizes min_tait ^^

jhpratt:
I must've missed the conversation on this. Is this gate being removed because the soundness issues were resolved? If I'm remembering things correctly.

oli-obk:
> I must've missed the conversation on this. Is this gate being removed because the soundness issues were resolved? If I'm remembering things correctly.

we're moving to a new system in which a min_tait system makes no sense anymore, as everything "just works" and is indeed sound

jhpratt:
Awesome!

PR #87512, opened 2021-07-27
Port clippy lint `redundant_field_names` to compiler

fee1-dead:


camsteffen:
I think we should remove the lint from Clippy in the same PR.

The lint really should be named `redundant_field_initializers` or `field_init_same_name` or `field_init_repeats_name`.

Are we okay with uplifting this, warn by default, with the possibility of enhancing it if we allow `Struct { &foo }`? (I think that's fine).

Can we put this in the `nonstandard-style` group? It seems like rustc is lacking a "style" group for slightly more picky lints. Someone will say that such lints belong in Clippy, but I'm not so sure. This lint feels like it could be in rustc.

If this lint triggers in macros for `$a: $b`, that's probably a deal-breaker. Maybe we can detect cases like `$a: $a`.

fee1-dead:
> Can we put this in the nonstandard-style group?

I don't think so because AFAIK that group is only used for lints with nonstandard naming conventions.

fee1-dead:
Hey @cjgillot do you have time to review this? Or should I reassign someone else to get their eyes on this?

cjgillot:
LGTM on the rustc side. If @camsteffen is ok on the clippy side, r=me.

compiler/rustc_lint/src/passes.rs
```diff
@@ -127,6 +127,7 @@ macro_rules! declare_combined_late_lint_pass {
         }
 
         impl $name {
+            #[allow(redundant_field_initializers)]
```
camsteffen, line 130: Shouldn't we prevent this by checking if the struct expression span is from expansion?

camsteffen:
Clippy changes LGTM. I just have a concern regarding macros.

compiler/rustc_lint/src/passes.rs
fee1-dead: Yeah, already fixed.

fee1-dead:
I don't have bors permissions ATM. @cjgillot can you approve it please? Thanks.

workingjubilee:
Why are we uplifting this lint?

fee1-dead:
@joshtriplett said in [this zulip thread](https://rust-lang.zulipchat.com/#narrow/stream/122651-general/topic/clippy.20lints.20to.20CI/near/210310736) that he would like to see clippy lints uplifted.

I do see a small problem for projects that want to maintain a quite low msrv (<1.17.0). But I don't know about whether there are projects that need to maintain a msrv this ancient.

To quote the rustc-dev-guide:
> ## [Diagnostic Levels](https://rustc-dev-guide.rust-lang.org/diagnostics.html?highlight=lint#diagnostic-levels)
>
> Guidelines for different diagnostic levels:
>
> - `error`: [...]
>
> - `warning`: [...] Some examples of when it is appropriate to issue a warning:
>     - [...]
>     - **Unnecessary syntax that can be removed without affecting the semantics of the code. For example, unused code, or unnecessary `unsafe`**.
>     - [...]

workingjubilee:
This could be considered in the same category, but it is not an instance of code that seems like it was intended to be used, or reflective of a misunderstanding as with a false invocation of `unsafe`, it is simply a bit of repetition. This has usually not merited a lint in rustc, and indeed the language actually favors repetition as with e.g. trailing commas where it would make things easier to rename and refactor... as might be a sensible choice here.

Many lints have been uplifted from clippy in the past, and I would honestly be happy to see more. But what I mean is that previously many of the lints that were uplifted from clippy were correctness lints (of which there are even now many that seem like good candidates) and were both motivated and granted on such a reason. Most of the interest I have seen expressed that aim, and your excerpt comes right next to these quotes:

> Care should be taken when adding warnings to avoid warning fatigue, and avoid false-positives where there really isn't a problem with the code.

> Code that is very likely to be incorrect, dangerous, or confusing, but the language technically allows

Yes, there are projects that have an MSRV below 1.17: serde is one of them.

Mark-Simulacrum:
> @joshtriplett said in this [zulip](https://rust-lang.zulipchat.com/#narrow/stream/122651-general/topic/clippy.20lints.20to.20CI/near/210310736) thread that he would like to see clippy lints uplifted.

I am pretty sure that the interpretation here is not that *all* clippy lints should be uplifted, but rather that approval of lints for uplift via targeted PRs or T-lang MCPs is much more likely to be a successful path toward a decision, as opposed to the prior art at the time of that comment, which was an issue proposing lots of lints to be moved.

Personally, I tend to agree with @workingjubilee that it feels like this lint definitely belongs either in clippy or as an allow-by-default lint; it's primarily a stylistic lint, and doesn't necessarily mean that all people will want to write their code like this. I'd personally be frustrated when seeing a warning on this pattern because the code is equally correct either way and not really significantly more verbose or anything like that.

camsteffen:
Without a lint category, it doesn't really make sense to uplift.  I think rustc should have an allow-by-default category for lints like this:

* clearly more idiomatic
* purely stylistic, non-critical, low-risk
* lints usage of a language feature or prelude items (generally not std API usage)
* the scope of the lint is clear and simple with no perceivable future enhancements

Some candidates: `needless_return`, `write_with_newline`, `writeln_empty_string`, `redundant_pattern`, `redundant_static_lifetimes`, `single_match`, `unused_unit`

There is a very large gap between rustc lints and Clippy lints in terms of stability, false positive rate, lint pickiness, opinionatedness. Of course some of the "blame" falls squarely on Clippy to fix bugs, remove un-useful lints, etc. But Clippy will likely never have the level of stability that rustc has as it is sort of a playground for new lints. A rustc lint category can bridge that gap a bit for people who want rustc to be just a little more picky, but in a highly stable way.

workingjubilee:
I understand Clippy has a lot of highly experimental lints, but are we sure there isn't any chance that Clippy could make the first move on growing such a "low stress" mode where it just emits warnings on lints with stronger consensus behind them? It would seem an appropriate "proof of concept" for the idea, at least.

Clippy is often run in CI for Rust projects and if it denies something then CI doesn't pass. And a lot of projects run `-D warnings` in CI, so if Clippy is pretty unstable then that is probably a bit of a problem. It may be a good idea to more actively rein that in somewhat, or at least use the nursery category a bit more freely when it becomes apparent a lint is more half-baked than expected.

PR #87513, opened 2021-07-27
bootstrap.py: change `git log` option to indicate desired behavior

hudson-ayers:
When determining which LLVM artifacts to download, bootstrap.py calls: `git log --author=bors --format=%H -n1 -m --first-parent --
src/llvm-project src/bootstrap/download-ci-llvm-stamp src/version`. However, the `-m` option has no effect, per the `git log` help:

> -m
> This option makes diff output for merge commits to be shown in the
> default format. -m will produce the output only if -p is given as
> well. The default format could be changed using log.diffMerges
> configuration parameter, which default value is separate.

Accordingly, this commit removes use of the -m option in favor of ~~`--diff-merges=off`~~ `--no-patch`, since no diff information is needed, and in fact the presence of a diff breaks the command. Tested using git 2.32, this does not change the output of the command.

The motivation for this change is that some patched versions of git change the behavior of the `-m` flag to imply `-p`, rather than to do nothing unless `-p` is passed. These patched versions of git lead to this script not working. Google's corp-provided git is one such example.

### DIFF
--- a/src/bootstrap/bootstrap.py
+++ b/src/bootstrap/bootstrap.py
@@ -473,7 +473,7 @@ class RustBuild(object):
             ]).decode(sys.getdefaultencoding()).strip()
             llvm_sha = subprocess.check_output([
                 "git", "log", "--author=bors", "--format=%H", "-n1",
-                "-m", "--first-parent",
+                "--no-patch", "--first-parent",
                 "--",
                 "{}/src/llvm-project".format(top_level),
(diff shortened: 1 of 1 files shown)


hudson-ayers:
Updated to use `--no-patch` instead of `--diff-merges=off` based on https://rust-lang.zulipchat.com/#narrow/stream/122651-general/topic/.2E.2Fx.2Epy.20build.20failures/near/247342368

PR #86197, opened 2021-06-10
Remove unnecessary trailing whitespace from error messages

FabianWolff:
Some error messages currently contain unnecessary trailing whitespace. There are some legitimate reasons for having trailing whitespace in the output, such as for uniform indentation of possibly-empty input lines, but the whitespace I have addressed here occurs in a line used only for spacing, and I see no reason why that should have trailing whitespace (spacing lines inserted in other places also don't have trailing whitespace).

I have also removed a superfluous call to `buffer.putc()`, which has no effect because the same character is already placed there by `draw_col_separator()`.

Use `git diff --ignore-space-at-eol` to see my changes; otherwise the diff is quite large due to the whitespace removed from expected outputs in `src/test/ui/`.

PR #87514, opened 2021-07-27
Avoid trailing space in error message

ptrojahn:


oli-obk:
👍 

you need to bless some more files. It is also possible you may have to run with `--compare-mode=nll`

oli-obk:
Please ping me when you force push, I only get a notification for the bors message, not for your force push

ptrojahn:
This was already fixed by #86197.

PR #87522, opened 2021-07-27
Fix assert in diy_float

frogtd:
The shifting should have gone the other way, the current incarnation is always true.


yaahc:
tyvm!

@bors r+

PR #87523, opened 2021-07-27
Stop creating a reference then immediately dereferencing it.

frogtd:
Stop creating a reference then immediately dereferencing it.


dtolnay (approved):
Thanks!

XynoxTheDev (approved):
lgtm

PR #87527, opened 2021-07-27
Don't run MIR unsafeck at all when using `-Zthir-unsafeck`

LeSeulArtichaut:
I don't know how I missed this :D
r? @oli-obk

### DIFF
--- a/compiler/rustc_mir/src/transform/mod.rs
+++ b/compiler/rustc_mir/src/transform/mod.rs
@@ -259,10 +259,12 @@ fn mir_const<'tcx>(
     }
 
     // Unsafety check uses the raw mir, so make sure it is run.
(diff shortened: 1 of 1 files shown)


oli-obk:
Huh 😄 

how does it then work that we don't see MIR unsafeck messages with -Zthir-unsafeck? Do we abort compilation before we get there if there are any errors?

LeSeulArtichaut:
That's because the `unsafety_check_result` query doesn't actually report any errors, that is left to the `check_unsafety` function: https://github.com/rust-lang/rust/blob/aea2e446f09b923d513013743b37b43fca7282dc/compiler/rustc_mir/src/transform/check_unsafety.rs#L506-L570

PR #87528, opened 2021-07-27
stack overflow handler specific openbsd change.

devnexen:


library/std/src/sys/unix/stack_overflow.rs
```diff
@@ -143,11 +143,17 @@ mod imp {
     }
 
     unsafe fn get_stackp() -> *mut libc::c_void {
+        #[cfg(target_os = "openbsd")]
+        let flags = MAP_PRIVATE | MAP_ANON | libc::MAP_STACK;
+        // OpenBSD requires this flag for stack mapping
```
the8472, line 148: "this flag" is referring to `MAP_STACK`? If so the comment is on the wrong line.

devnexen: indeed :)

the8472:
The PR title and description could use some improvement. After all every commit implicitly is a *change*. And why is it needed? Is OpenBSD broken without it or is it only a correctness fix that currently doesn't make a difference but may do so in the future? If it's the latter shouldn't it also be applied on linux since the manpage there says that

> However, by employing this flag, applications can ensure that they transparently obtain support if the flag is implemented in the future.

library/std/src/sys/unix/stack_overflow.rs
```diff
@@ -143,14 +143,15 @@ mod imp {
...
-        let stackp = mmap(
-            ptr::null_mut(),
-            SIGSTKSZ + page_size(),
-            PROT_READ | PROT_WRITE,
-            MAP_PRIVATE | MAP_ANON,
-            -1,
-            0,
-        );
+        // OpenBSD requires this flag for stack mapping
+        // otherwise the said mapping will fail is a no-op in most systems
```
bjorn3, line 147: s/is/as

bjorn3: s/in/on
reads nicer I think.

library/std/src/sys/unix/stack_overflow.rs
```diff
@@ -143,14 +143,15 @@ mod imp {
...
+        // and has a different meaning on FreeBSD
+        #[cfg(target_os = "openbsd")]
+        let flags = MAP_PRIVATE | MAP_ANON | libc::MAP_STACK;
+        #[cfg(not(target_os = "openbsd"))]
+        let flags = MAP_PRIVATE | MAP_ANON;
```
bjorn3, line 152: Linux supports `MAP_STACK` too. It is a no-op on all architectures. FreeBSD supports it too. For example macOS doesn't support it though.

devnexen: yes but not necessary on these oses plus solaris/illumos does not have it and I believe haiku neither.

bjorn3: > yes but not necessary

Not **currently** necessary. A future architecture may need it. That is why it exists as no-op at all.

devnexen: ok can add for linux, netbsd but I m less sure about FreeBSD the flag is working but does not seem to work well with null start addresses, seems little usage difference.

library/std/src/sys/unix/stack_overflow.rs
devnexen: sure thing

joshtriplett:
Seems reasonable, and consistent with the OpenBSD documentation.

@bors r+

workingjubilee:
Appears to not work with all Unix libcs?
```
[RUSTC-TIMING] addr2line test:false 0.596
[RUSTC-TIMING] core test:false 27.851
[RUSTC-TIMING] gimli test:false 6.055
[RUSTC-TIMING] object test:false 6.197
error[E0425]: cannot find value `MAP_STACK` in crate `libc`
   --> library/std/src/sys/unix/stack_overflow.rs:150:52
    |
150 |         let flags = MAP_PRIVATE | MAP_ANON | libc::MAP_STACK;
    |                                                    ^^^^^^^^^ not found in `libc`
For more information about this error, try `rustc --explain E0425`.
[RUSTC-TIMING] std test:false 2.972
error: could not compile `std` due to previous error
Build completed unsuccessfully in 0:06:57
```
@bors r-

devnexen:
it might needs a libc update ... it is present since the 0.2.102 release (or updating the current PR to not enabling on netbsd).

devnexen:
having lost the original repo since, I opted for libc update, anyway it would have been needed at some point.

JohnTitor:
The `libc` dependency now has been updated, @bors r=joshtriplett

PR #87532, opened 2021-07-27
bootstrap.py: use `git rev-list` for robustness

tlyu:
Use `git rev-list` instead of `git log` to be more robust against
UI changes in git. Also, use the full email address for bors,
because `--author` uses a substring match.

Based on #87513, but is separate because it's less minimal and may require additional manual testing.

~Open questions:~
* ~Should the `merge_base` search also use `--first-parent`?~
* ~Do we exclude non-merge commits from bors? There are a few, and I'm not sure what they have in common. Some of them look like squashes, and some look like they're in rollup branches.~

r? @jyn514
@rustbot label +A-rustbuild +C-cleanup

### DIFF
--- a/src/bootstrap/bootstrap.py
+++ b/src/bootstrap/bootstrap.py
@@ -473,7 +473,7 @@ class RustBuild(object):
             ]).decode(sys.getdefaultencoding()).strip()
             llvm_sha = subprocess.check_output([
                 "git", "log", "--author=bors", "--format=%H", "-n1",
-                "-m", "--first-parent",
+                "--no-patch", "--first-parent",
                 "--",
                 "{}/src/llvm-project".format(top_level),
                 "{}/src/bootstrap/download-ci-llvm-stamp".format(top_level),
--- a/src/bootstrap/bootstrap.py
+++ b/src/bootstrap/bootstrap.py
@@ -464,7 +464,8 @@ class RustBuild(object):
             # LLVM more often than necessary.
             #
             # This git command finds that commit SHA, looking for bors-authored
-            # merges that modified src/llvm-project.
+            # commits that modified src/llvm-project or other relevant version
+            # stamp files.
             #
             # This works even in a repository that has not yet initialized
             # submodules.
@@ -472,8 +473,8 @@ class RustBuild(object):
                 "git", "rev-parse", "--show-toplevel",
             ]).decode(sys.getdefaultencoding()).strip()
             llvm_sha = subprocess.check_output([
-                "git", "log", "--author=bors", "--format=%H", "-n1",
-                "--no-patch", "--first-parent",
+                "git", "rev-list", "--author=bors@rust-lang.org", "-n1",
+                "--first-parent", "HEAD",
                 "--",
                 "{}/src/llvm-project".format(top_level),
                 "{}/src/bootstrap/download-ci-llvm-stamp".format(top_level),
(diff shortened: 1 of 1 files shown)


jyn514:
> Do we exclude non-merge commits from bors? There are a few, and I'm not sure what they have in common. Some of them look like squashes, and some look like they're in rollup branches.

I don't see why we would. The reason to use `--author=bors` is to make sure the commit has CI artifacts; as long as it does it doesn't matter if it's a merge or not.

> Should the merge_base search also use --first-parent?

Hmm, the man page says:

> Follow only the first parent commit upon seeing a merge commit. This option can give a better overview when viewing the evolution of a particular topic branch, because merges into a topic branch tend to be only about adjusting to updated upstream from time to time, and this option allows you to ignore the individual commits brought in to your history by such a merge.

I don't know why we use this in the LLVM download - I'm not confident that the first parent will always be the one with a bors author, and `--author=bors` should be enough of a filter anyway. cc @Mark-Simulacrum, you added this originally I believe.

Either way I don't expect it to make a big difference, we say pretty clearly on https://rustc-dev-guide.rust-lang.org/git.html#no-merge-policy that you shouldn't use merge commits.

jyn514 (approved):
This looks good to me.

Mark-Simulacrum:
author=bors is insufficient (in general) because clippy and potentially other repos also use bors but (obviously) don't upload the full set of rustc artifacts (or any artifacts at all today).

I believe first parent means we traverse the chain of commits appropriately, always going along rust-lang/rust bors merges.

tlyu:
> > Do we exclude non-merge commits from bors? There are a few, and I'm not sure what they have in common. Some of them look like squashes, and some look like they're in rollup branches.
> 
> I don't see why we would. The reason to use `--author=bors` is to make sure the commit has CI artifacts; as long as it does it doesn't matter if it's a merge or not.

I think bors might only run the CI on its auto-merge commits, not its non-merge commits? But I'd want to check with someone who knows bors/homu better.

> > Should the merge_base search also use --first-parent?
>
> I don't know why we use this in the LLVM download - I'm not confident that the first parent will always be the one with a bors author, and `--author=bors` should be enough of a filter anyway. cc @Mark-Simulacrum, you added this originally I believe.

I think that `--first-parent` might help exclude merge commits that are parts of rollup merges, but I'm not sure. It does seem to exclude some of the squash commits made by bors. (See below.)

> Either way I don't expect it to make a big difference, we say pretty clearly on https://rustc-dev-guide.rust-lang.org/git.html#no-merge-policy that you shouldn't use merge commits.

Most of the non-bors merge commits that I see at a casual glance seem to be parts of rollup merges.

Details on non-merge bors commits:

```
git log --no-merges --author=bors@rust-lang.org --pretty=%h upstream/master
76a3b609d0b
98b11c8f7bf
4b806878549
cb4553bdae5
733cb54d18c
204bb9b54b1
```

but

```
git log --first-parent --no-merges --author=bors@rust-lang.org --pretty=%h upstream/master
```

produces no output.

Also, digging through `git log --graph` for the commits listed by the first command shows that the next commit after each of those non-merge bors commits seems to be a merge commit from bors.

So now I'm leaning towards yes on `--first-parent` and probably yes on `--merges`.

tlyu:
Pushed an update. Now both of them use `--first-parent` and `--merges`.

tlyu:
Looks like #87513 has merged; should I rebase and/or squash?

jyn514:
@tlyu I don't think you need to, there aren't conflicts.

jyn514:
@bors r+

I'm not sure I buy that this is avoiding clippy commits, but it doesn't seem any worse than before, and we can fix it if something goes wrong.

PR #87535, opened 2021-07-28
rfc3052 followup: Remove authors field from Cargo manifests

lf-:
Since RFC 3052 soft deprecated the authors field, hiding it from
crates.io, docs.rs, and making Cargo not add it by default, and it is
not generally up to date/useful information for contributors, we may as well
remove it from crates in this repo.

### DIFF
--- a/compiler/rustc_apfloat/Cargo.toml
+++ b/compiler/rustc_apfloat/Cargo.toml
@@ -1,5 +1,4 @@
 [package]
-authors = ["The Rust Project Developers"]
 name = "rustc_apfloat"
 version = "0.0.0"
 edition = "2018"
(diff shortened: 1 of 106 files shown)


notriddle:
Clippy and Rustfmt are subtrees, and [must be updated in separate commits](https://rustc-dev-guide.rust-lang.org/contributing.html?highlight=Subtree#external-dependencies-subtree).

compiler/rustc_apfloat/Cargo.toml
```diff
@@ -1,5 +1,4 @@
 [package]
-authors = ["The Rust Project Developers"]
```
eddyb, line 2: Checking off on my ping, I guess the bot doesn't check for the `src` subdir.

Mark-Simulacrum:
@bors r+

I think this is a reasonable step to take, thanks!

PR #87537, opened 2021-07-28
Clarify undefined behaviour in binary heap, btree and hashset docs

Wilfred:
Previously, it wasn't clear whether "This could include" was referring to logic errors, or undefined behaviour. Tweak wording to clarify this sentence does not relate to UB.

### DIFF
--- a/library/alloc/src/collections/binary_heap.rs
+++ b/library/alloc/src/collections/binary_heap.rs
@@ -162,9 +162,9 @@ use super::SpecExtend;
 /// item's ordering relative to any other item, as determined by the [`Ord`]
 /// trait, changes while it is in the heap. This is normally only possible
 /// through [`Cell`], [`RefCell`], global state, I/O, or unsafe code. The
-/// behavior resulting from such a logic error is not specified, but will
-/// not result in undefined behavior. This could include panics, incorrect
-/// results, aborts, memory leaks, and non-termination.
+/// behavior resulting from such a logic error is not specified (it
+/// could include panics, incorrect results, aborts, memory leaks, or
+/// non-termination) but will not be undefined behavior.
 ///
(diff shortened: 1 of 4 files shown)


SkiFire13:
I don't think this makes it more clear, quite the opposite, at first glance I thought the "behaviour" you added was referring to the "undefined behaviour".

You should also change the docs for `HashMap`, `HashSet`, `BTreeMap` and `BTreeSet` since they all include that same sentence

kpreid:
> I don't think this makes it more clear, quite the opposite, at first glance I thought the "behaviour" you added was referring to the "undefined behaviour".

How about wording like this? “The behavior resulting from such a logic error is not specified (it could include panics, incorrect results, aborts, memory leaks, or non-termination) but will not be undefined behavior.”

(I think it would also be good to link “undefined behavior” to some page explaining what undefined behavior is, since this might be the first time someone has encountered the concept, and it might not be obvious that “undefined behavior” is a specific formal term.)

Mark-Simulacrum:
Agreed that this seems like a good clarification.

It may make sense to extract the shared language somewhere in the root of the library docs as a general statement we can link to in terms of what logic errors can do; that way we can avoid updating N places each time and potentially expand a little on the paragraph. std's top level doc comment seems like a reasonable place to add that section, I guess.

JohnCSimon:
Ping from triage: 
@Wilfred can you please address the comments from the reviewer? Thank you.

Wilfred:
OK, I've updated the docs in all four places using the wording suggested by @kpreid :). The patch isn't identical in all those files as they've used different line lengths, so I followed the convention in each file.

Mark-Simulacrum:
@bors r+ rollup

Rebased and approved, thanks!

PR #87547, opened 2021-07-28
Add missing examples for NonNull

GuillaumeGomez:


library/core/src/ptr/non_null.rs
```diff
@@ -205,6 +240,22 @@ impl<T: ?Sized> NonNull<T> {
...
 
     /// Acquires the underlying `*mut` pointer.
+    ///
+    /// # Examples
+    ///
+    /// ```
+    /// use std::ptr::NonNull;
+    ///
+    /// let mut x = 0u32;
+    /// let ptr = NonNull::<u32>::new(&mut x as *mut _).expect("ptr is null!");
```
kennytm, line 250: the turbofish and `as` cast seems unnecessary, is this intended?

```rust
let ptr = NonNull::new(&mut x).expect("ptr is null!");
```

GuillaumeGomez: You're absolutely right! Removing the turbofish.

GuillaumeGomez:
Updated!

library/core/src/ptr/non_null.rs
kennytm: @GuillaumeGomez  well change the remaining examples then 😅 r=me after that.

GuillaumeGomez: Sorry, mornings...

library/core/src/ptr/non_null.rs
```diff
@@ -155,6 +165,18 @@ impl<T: ?Sized> NonNull<T> {
...
+    /// # Examples
+    ///
+    /// ```
+    /// use std::ptr::NonNull;
+    ///
+    /// let mut x = 0u32;
+    /// let ptr = unsafe { NonNull::new_unchecked(&mut x as *mut _) };
+    ///
+    /// // NEVER DO THAT!!!
+    /// let ptr = unsafe { NonNull::<u32>::new_unchecked(std::ptr::null_mut()) };
```
RalfJung, line 178: Please mark testcases that cause UB as `no_run` (see e.g. the `MaybeUninit` docs).
Otherwise, https://github.com/rust-lang/miri-test-libstd/ will fail.

RalfJung: See https://github.com/rust-lang/rust/pull/87653

GuillaumeGomez: Noted!

PR #87548, opened 2021-07-28
WIP: Normalize associated types in structs when performing unsizing coercion 

### LINKED ISSUE #75899: Unsizing coercion does not normalize associated types in structs.
Example code I'd expect to work ([playground](https://play.rust-lang.org/?version=nightly&mode=debug&edition=2018&gist=da6411306c5266f6010ff5172838a2d3)):
```rust
// NOTE: this is only needed for coerce_tuple, you can remove both for a stable version.
#![feature(unsized_tuple_coercion)]

trait Trait {}
impl<T> Trait for T {}

trait Noop {
    type Assoc: ?Sized;
}
impl<T: ?Sized> Noop for T {
    type Assoc = T;
}

// Both of these work:
fn coerce<T: Trait>(x: &<T as Noop>::Assoc) -> &<dyn Trait as Noop>::Assoc { x }
fn coerce_tuple<T: Trait>(x: &(String, <T as Noop>::Assoc)) -> &(String, <dyn Trait as Noop>::Assoc) { x }

// But this doesn't:
struct NoopNewtype<T: ?Sized + Noop>(T::Assoc);
fn coerce_newtype<T: Trait>(x: &NoopNewtype<T>) -> &NoopNewtype<dyn Trait> { x }
```

(in case the example above seems artificial, the [realistic one](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=b3ef269c501b3e6478275335f7922dba) I've reduced this from involves using an associated trait to add `Option` around sized types, but not `dyn Trait`, since `Option<dyn Trait>` wouldn't itself work)

<hr/>

I'd expect `coerce_newtype` to work because `<T as Noop>::Assoc: Unsize<<dyn Trait as Noop>::Assoc>` should hold, after normalization to `T: Unsize<dyn Trait>` (which holds here because `T: Trait`), but nothing performs that normalization before coercion gives up entirely.

Unlike `CoerceUnsized` (where I've been wary of associated types in the past), there should be no weird interactions here, we should be able to largely treat the `struct` as a tuple (and we already ensure the type parameter being changed doesn't show up anywhere else in the `struct`, anyway).

<hr/>

The `coerce` and `coerce_tuple` examples get to normalize *before* asking the trait system about any `CoerceUnsized` or `Unsized` obligations - I wonder if we could just normalize `trait_pred` here? https://github.com/rust-lang/rust/blob/3d6a3ed15823cce765d56952d954e1bd8166dfa7/src/librustc_typeck/check/coercion.rs#L603

But we would have to be careful, because we *are* in a speculative operation, so we would have to use `normalize_associated_types_in_as_infer_ok`, like this other piece of coercion logic: https://github.com/rust-lang/rust/blob/3d6a3ed15823cce765d56952d954e1bd8166dfa7/src/librustc_typeck/check/coercion.rs#L732-L733

cc @mystor (author of the `object-provider` crate where this could be useful) @nikomatsakis

WaffleLapkin:
Closes #75899

r? @nikomatsakis

### DIFF
--- a/compiler/rustc_typeck/src/check/coercion.rs
+++ b/compiler/rustc_typeck/src/check/coercion.rs
@@ -184,7 +184,7 @@ impl<'f, 'tcx> Coerce<'f, 'tcx> {
         let unsize = self.commit_if_ok(|_| self.coerce_unsized(a, b));
         match unsize {
             Ok(_) => {
-                debug!("coerce: unsize successful");
+                debug!("coerce: unsize successful: {:?}", unsize);
                 return unsize;
             }
             Err(TypeError::ObjectUnsafeCoercion(did)) => {
@@ -603,6 +603,15 @@ impl<'f, 'tcx> Coerce<'f, 'tcx> {
                     continue;
                 }
             };
+
+            
+            let InferOk { value: trait_pred2, obligations: obls } =
+            self.normalize_associated_types_in_as_infer_ok(self.cause.span, trait_pred);
+            
+            debug!("coerce_unsized: normalized {:?} to {:?}; obls: {:?}", trait_pred, trait_pred2, obls);
+            let trait_pred = trait_pred2;
+            // coercion.obligations.extend(obls);
+            
             match selcx.select(&obligation.with(trait_pred)) {
                 // Uncertain or unimplemented.
                 Ok(None) => {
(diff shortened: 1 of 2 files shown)


jackh726 (changes requested):

jackh726:
> This is **very** WIP and actually even abandoned

I'm going to go ahead and close this as inactive...feel free to reopen if you want to continue working on it

PR #87553, opened 2021-07-28
Fix typo in rustc_driver::version

bjorn3:
This caused rustc `-Zcodegen-backend=foo.so -vV` to look for `oo.so` instead of `foo.so`

### DIFF
--- a/compiler/rustc_driver/src/lib.rs
+++ b/compiler/rustc_driver/src/lib.rs
@@ -764,13 +764,7 @@ pub fn version(binary: &str, matches: &getopts::Matches) {
         println!("release: {}", unw(util::release_str()));
 
         let debug_flags = matches.opt_strs("Z");
-        let backend_name = debug_flags.iter().find_map(|x| {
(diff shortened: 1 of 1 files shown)


wesleywiser (approved):

compiler/rustc_driver/src/lib.rs
```diff
@@ -766,7 +766,7 @@ pub fn version(binary: &str, matches: &getopts::Matches) {
         let debug_flags = matches.opt_strs("Z");
         let backend_name = debug_flags.iter().find_map(|x| {
             if x.starts_with("codegen-backend=") {
-                Some(&x["codegen-backends=".len()..])
+                Some(&x["codegen-backend=".len()..])
             } else {
                 None
             }
```
cuviper, line 772: This expression is the same as `x.strip_prefix("codegen-backend=")`.

bjorn3: Changed to `strip_prefix`.

PR #87557, opened 2021-07-28
Fix issue with autofix for ambiguous associated function from Rust 2021 prelude when struct is generic

### LINKED ISSUE #86940: Prelude collision lint: invalid suggestion for T::from_iter when T is generic.
For generic types, the prelude collision lint can produce suggestions that do not compile:

Code:
```rust
struct Generic<T>(T);

trait Hey { fn from_iter(_: i32) -> Self; }

impl Hey for Generic<i32> { fn from_iter(x: i32) -> Self { Self(x) } }

impl std::iter::FromIterator<i32> for Generic<i32> { fn from_iter<T: IntoIterator<Item = i32>>(_: T) -> Self { todo!() } }

fn main() {
    Generic::from_iter(1);
}
```
Suggestion from the prelude collision lint:
```
warning: trait-associated function `from_iter` will become ambiguous in Rust 2021
  --> src/main.rs:10:5
   |
10 |     Generic::from_iter(1);
   |     ^^^^^^^^^^^^^^^^^^ help: disambiguate the associated function: `<Generic as Hey>::from_iter`
   |
   = note: `-W rust-2021-prelude-collisions` implied by `-W rust-2021-compatibility`
   = warning: this is accepted in the current edition (Rust 2018) but is a hard error in Rust 2021!
   = note: for more information, see issue #85684 <https://github.com/rust-lang/rust/issues/85684>
```
Error after applying the suggestion:
```
error[E0107]: missing generics for struct `Generic`
  --> src/main.rs:10:6
   |
10 |     <Generic as Hey>::from_iter(1);
   |      ^^^^^^^ expected 1 generic argument
   |
```

rylev:
Fixes #86940

The test cases and associated issue should make it clear what specifically this is meant to fix. The fix is slightly hacky in that we check against the literal source code of the call site for the presence of `<` in order to determine if the user has included the generics for the struct (meaning we don't need to include them for them).

r? @nikomatsakis


compiler/rustc_typeck/src/check/method/prelude2021.rs
```diff
@@ -255,16 +255,28 @@ impl<'a, 'tcx> FnCtxt<'a, 'tcx> {
...
 
-            let self_ty = self
+            let self_ty_name = self
                 .sess()
                 .source_map()
                 .span_to_snippet(self_ty_span)
                 .unwrap_or_else(|_| self_ty.to_string());
 
+            let self_ty_generics_count = match self_ty.kind() {
+                Adt(def, _) if !self_ty_name.contains("<") => self.tcx.generics_of(def.did).count(),
```
nikomatsakis, line 265: This deserves a comment :)

nikomatsakis (changes requested):
r=me with comment

PR #87565, opened 2021-07-28
Use backticks when referring to `core::future::Ready` in panic message

ibraheemdev:


scottmcm:
Hmm, looks like we have them around None in the Option::unwrap message, so seems reasonable to me.

@bors r+

PR #87568, opened 2021-07-28
rustc: Replace `HirId`s with `LocalDefId`s in `AccessLevels` tables

petrochenkov:
and passes using those tables - primarily privacy checking, stability checking and dead code checking.

All these passes work with definitions rather than with arbitrary HIR nodes.
r? @cjgillot 
cc @lambinoo (#87487)

### DIFF
--- a/compiler/rustc_middle/src/middle/privacy.rs
+++ b/compiler/rustc_middle/src/middle/privacy.rs
@@ -3,9 +3,8 @@
 //! which are available for use externally when compiled as a library.
 
 use rustc_data_structures::fx::FxHashMap;
-use rustc_hir::HirId;
 use rustc_macros::HashStable;
-use std::fmt;
+use rustc_span::def_id::LocalDefId;
 use std::hash::Hash;
 
 /// Represents the levels of accessibility an item can have.
@@ -27,8 +26,8 @@ pub enum AccessLevel {
 }
 
 /// Holds a map of accessibility levels for reachable HIR nodes.
-#[derive(Clone)]
-pub struct AccessLevels<Id = HirId> {
+#[derive(Debug)]
+pub struct AccessLevels<Id = LocalDefId> {
     pub map: FxHashMap<Id, AccessLevel>,
 }
 
@@ -49,14 +48,8 @@ impl<Id: Hash + Eq> AccessLevels<Id> {
     }
 }
 
-impl<Id: Hash + Eq> Default for AccessLevels<Id> {
+impl<Id> Default for AccessLevels<Id> {
     fn default() -> Self {
         AccessLevels { map: Default::default() }
-    }
-}
-
-impl<Id: Hash + Eq + fmt::Debug> fmt::Debug for AccessLevels<Id> {
-    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
-        fmt::Debug::fmt(&self.map, f)
     }
 }
--- a/compiler/rustc_privacy/src/lib.rs
+++ b/compiler/rustc_privacy/src/lib.rs
@@ -11,7 +11,7 @@ use rustc_data_structures::fx::FxHashSet;
 use rustc_errors::struct_span_err;
 use rustc_hir as hir;
 use rustc_hir::def::{DefKind, Res};
-use rustc_hir::def_id::{DefId, LocalDefId, CRATE_DEF_INDEX, LOCAL_CRATE};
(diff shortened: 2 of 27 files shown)


compiler/rustc_privacy/src/lib.rs
```diff
@@ -888,32 +889,36 @@ impl Visitor<'tcx> for EmbargoVisitor<'tcx> {
...
-            let changed_reachability = self.update_macro_reachable(module_id, macro_module_def_id);
-            if changed_reachability || module_id == hir::CRATE_HIR_ID {
+            let changed_reachability =
+                self.update_macro_reachable(module_def_id, macro_module_def_id);
+            if changed_reachability || module_def_id == CRATE_DEF_ID {
                 break;
             }
-            module_id = self.tcx.hir().get_parent_node(module_id);
+            module_def_id =
+                ty::DefIdTree::parent(self.tcx, module_def_id.to_def_id()).unwrap().expect_local();
```
petrochenkov, line 921: The logic in this function was slightly changed to walk `DefId` parents instead of `HirId` parents, but the end result should be the same, I think.

jyn514 (approved):

compiler/rustc_privacy/src/lib.rs
```diff
@@ -888,32 +889,36 @@ impl Visitor<'tcx> for EmbargoVisitor<'tcx> {
...
-            _ => return,
-        };
-        let level = if md.vis.node.is_pub() { self.get(module_id) } else { None };
-        let new_level = self.update(md.hir_id(), level);
+        let macro_module_def_id =
+            ty::DefIdTree::parent(self.tcx, md.def_id.to_def_id()).unwrap().expect_local();
+        if !matches!(
+            self.tcx.hir().get(self.tcx.hir().local_def_id_to_hir_id(macro_module_def_id)),
+            Node::Item(hir::Item { kind: hir::ItemKind::Mod(_), .. }) | Node::Crate(..)
+        ) {
```
cjgillot, line 902: Checking `self.tcx.hir().opt_def_kind(macro_module_def_id) != Some(DefKind::Mod)` should be enough.

compiler/rustc_privacy/src/lib.rs
cjgillot: IIUC, this loop is only accessed when `macro_module_def_id` point to a module that can be named from the outside. As a consequence, all its parents are modules. This is relied upon by `update_macro_reachable`. All modules are HIR owners, so walking parents by def_key or through HIR is equivalent.
Do you agree with the explanation? Should there be a comment?
Can the loop be simplified using `tcx.hir().parent_owner_iter()`?

compiler/rustc_middle/src/middle/privacy.rs
```diff
@@ -27,8 +26,8 @@ pub enum AccessLevel {
 }
 
 /// Holds a map of accessibility levels for reachable HIR nodes.
-#[derive(Clone)]
-pub struct AccessLevels<Id = HirId> {
+#[derive(Debug)]
+pub struct AccessLevels<Id = LocalDefId> {
```
cjgillot, line 30: Is this used somewhere with another `Id`?

cjgillot (approved):
LGTM. I left a few questions.

compiler/rustc_middle/src/middle/privacy.rs
petrochenkov: Yes, rustdoc uses `AccessLevels<DefId>` and actually adds foreign ids to the table as public or exported, I didn't investigate why exactly.

compiler/rustc_privacy/src/lib.rs
petrochenkov: Added a comment.
`parent_owner_iter` doesn't seem to make things simpler because it requires converting to `HirId`s and back.

compiler/rustc_middle/src/middle/privacy.rs
jyn514: IIRC it's because it needs to do its own privacy pass because rustc's is incorrect (#64762). It may be possible to remove when that's fixed, but don't quote me on that.

petrochenkov:
Updated.

jyn514:
```
fatal: unable to access 'https://github.com/rust-lang-ci/rust/': Failed to connect to github.com port 443: Connection timed out
```

@bors retry

PR #87579, opened 2021-07-29
Update Clippy

flip1995:
r? @Manishearth


flip1995:
I still have to rename 2 lints. Let me do that real quick.

flip1995:
Ok, this should be ready now.

PR #87580, opened 2021-07-29
Update Windows Argument Parsing

### LINKED ISSUE #44650: env::args differs from MSVC CRT in some cases on Windows
For a command-line of the form `"a/"b.exe`, where `a/b.exe` does indeed exist, `std::env::args()` produces different results than `argc` and `argv` in a C++ program. Specifically, `CommandLineToArgvW` is returning troublesome results. It looks like the CRT and CommandLineToArgvW [disagree](http://www.windowsinspired.com/how-a-windows-programs-splits-its-command-line-into-individual-arguments/).

* Invocation in Command Prompt: `"a/"b.exe`
* Expected arguments: [`a/b.exe`]
* C++ main() arguments: [`a/b.exe`] 
* Result of GetCommandLineW: `"a/"b.exe`
* Result of CommandLineToArgvW: [`a/`, `b.exe`]
* Result of `std::env::args()`: [`a/`, `b.exe`]

Obviously, this interpretation makes no sense. `env::current_exe` does not seem to be affected.

A Rust program which demonstrates the mismatch:
```rust
fn main() {
    for (i, j) in std::env::args().enumerate() {
        println!("{}: {}", i, j);
    }
}
```

A C++ program which demonstrates the mismatch:
```cpp
#include <stdio.h>
#include <windows.h>

int main(int argc, char* argv[]) {
	printf("---- argc and argv think:\n");
	for (int i = 0; i < argc; ++i) {
		printf("%i: %s\n", i, argv[i]);
	}
	
	printf("---- GetCommandLineW is:\n");
	printf("%ws\n", GetCommandLineW());
	
	printf("---- CommandLineToArgvW thinks:\n");
	LPWSTR *szArglist;
	int nArgs;
	szArglist = CommandLineToArgvW(GetCommandLineW(), &nArgs);
	if (szArglist == NULL) {
		wprintf(L"CommandLineToArgvW failed\n");
		return 0;
	}
	for (int i = 0; i < nArgs; i++) {
		printf("%d: %ws\n", i, szArglist[i]);
	}
}
```

ChrisDenton:
Fixes #44650

The Windows command line is passed to applications [as a single string](https://docs.microsoft.com/en-us/archive/blogs/larryosterman/the-windows-command-line-is-just-a-string) which the application then parses to get a list of arguments. The standard rules (as used by C/C++) for parsing the command line have slightly changed over the years, most recently in 2008 which added new escaping rules.

This PR implements the new rules as [described on MSDN](https://docs.microsoft.com/en-us/cpp/cpp/main-function-command-line-args?view=msvc-160#parsing-c-command-line-arguments) and [further detailed here](https://daviddeley.com/autohotkey/parameters/parameters.htm#WIN). It has been tested against the behaviour of C++ by calling a C++ program that outputs its raw command line and the contents of `argv`. See [my repo](https://github.com/ChrisDenton/winarg/tree/std) if anyone wants to reproduce my work.

For an overview of how this PR changes argument parsing behavior and why we feel it is warranted see https://github.com/rust-lang/rust/pull/87580#issuecomment-893833893. 

For some examples see: https://github.com/rust-lang/rust/pull/87580#issuecomment-894299249

### DIFF
--- a/library/std/src/sys/windows/args.rs
+++ b/library/std/src/sys/windows/args.rs
@@ -1,13 +1,18 @@
-#![allow(dead_code)] // runtime init functions not used during testing
+//! The Windows command line is just a string
+//! <https://docs.microsoft.com/en-us/archive/blogs/larryosterman/the-windows-command-line-is-just-a-string>
+//!
+//! This module implements the parsing necessary to turn that string into a list of arguments.
 
 #[cfg(test)]
 mod tests;
 
 use crate::ffi::OsString;
 use crate::fmt;
+use crate::marker::PhantomData;
+use crate::num::NonZeroU16;
 use crate::os::windows::prelude::*;
 use crate::path::PathBuf;
-use crate::slice;
+use crate::ptr::NonNull;
 use crate::sys::c;
 use crate::sys::windows::os::current_exe;
 use crate::vec;
@@ -15,9 +20,11 @@ use crate::vec;
 use core::iter;
 
 pub fn args() -> Args {
+    // SAFETY: `GetCommandLineW` returns a pointer to a null terminated UTF-16
+    // string so it's safe for `WStrUnits` to use.
     unsafe {
         let lp_cmd_line = c::GetCommandLineW();
-        let parsed_args_list = parse_lp_cmd_line(lp_cmd_line as *const u16, || {
+        let parsed_args_list = parse_lp_cmd_line(WStrUnits::new(lp_cmd_line), || {
             current_exe().map(PathBuf::into_os_string).unwrap_or_else(|_| OsString::new())
         });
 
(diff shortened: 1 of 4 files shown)


library/std/src/sys/windows/args.rs
```diff
@@ -28,129 +35,121 @@ pub fn args() -> Args {
...
 ) -> Vec<OsString> {
-    const BACKSLASH: u16 = '\\' as u16;
-    const QUOTE: u16 = '"' as u16;
-    const TAB: u16 = '\t' as u16;
-    const SPACE: u16 = ' ' as u16;
+    // SAFETY: These are all non-zero ASCII characters.
+    const BACKSLASH: NonZeroU16 = unsafe { NonZeroU16::new_unchecked(b'\\' as u16) };
+    const QUOTE: NonZeroU16 = unsafe { NonZeroU16::new_unchecked(b'"' as u16) };
+    const TAB: NonZeroU16 = unsafe { NonZeroU16::new_unchecked(b'\t' as u16) };
+    const SPACE: NonZeroU16 = unsafe { NonZeroU16::new_unchecked(b' ' as u16) };
```
Urgau, line 63: Random suggestion from an outsider, using `NonZeroU16::new` with `unwrap` seems better than using the unsafe function.
```suggestion
    const BACKSLASH: NonZeroU16 = NonZeroU16::new(b'\\' as u16).unwrap();
    const QUOTE: NonZeroU16 = NonZeroU16::new(b'"' as u16).unwrap();
    const TAB: NonZeroU16 = NonZeroU16::new(b'\t' as u16).unwrap();
    const SPACE: NonZeroU16 = NonZeroU16::new(b' ' as u16).unwrap();
```

ChrisDenton: @Urgau I definitely agree! But the `const` context means I'd need to use the unstable `#![feature(const_option)]` which doesn't appear to be enabled for the standard library yet.

Urgau: Ah, sorry I only check `core`, but because the feature is enable in `core`, it can also probably be enabled in `std`. If you want to enable it, should add in the [lib.rs](https://github.com/rust-lang/rust/blob/master/library/std/src/lib.rs) of `std`.

ChrisDenton: That sounds reasonable, thanks! I've added it. If anyone has reason not to want this feature in `std` I can always revert the commit.

library/std/src/sys/windows/args.rs
```diff
@@ -28,129 +35,120 @@ pub fn args() -> Args {
...
-/// <https://docs.microsoft.com/en-us/previous-versions//17w5ykft(v=vs.85)>.
+/// <https://docs.microsoft.com/en-us/cpp/cpp/main-function-command-line-args?view=msvc-160#parsing-c-command-line-arguments>
 ///
-/// Windows includes a function to do this in shell32.dll,
-/// but linking with that DLL causes the process to be registered as a GUI application.
+/// A more in-depth explanation is here:
+/// <https://daviddeley.com/autohotkey/parameters/parameters.htm#WIN>
+///
+/// Windows includes a function to do command line parsing in shell32.dll.
+/// However, this is not use for two reasons:
```
yaahc, line 44:
```suggestion
/// However, this is not used for two reasons:
```

library/std/src/sys/windows/args.rs
```diff
@@ -28,129 +35,120 @@ pub fn args() -> Args {
...
-/// Windows 10 Pro v1803, using an exhaustive test suite available at
-/// <https://gist.github.com/notriddle/dde431930c392e428055b2dc22e638f5> or
-/// <https://paste.gg/p/anonymous/47d6ed5f5bd549168b1c69c799825223>.
-unsafe fn parse_lp_cmd_line<F: Fn() -> OsString>(
-    lp_cmd_line: *const u16,
+/// 2. It does not follow the modern C/C++ argv rules outlined in the first two links above.
+///
+/// This function was tested for equivalence to the C/C++ parsing rules using an
+/// extensive test suite available at
+/// <https://github.com/ChrisDenton/winarg/tree/std>.
```
yaahc, line 54: would it be possible to integrate some or all of this into our testsuite?

library/std/src/sys/windows/args.rs
```diff
@@ -187,3 +185,53 @@ impl ExactSizeIterator for Args {
...
+    }
+    fn peek(&self) -> Option<NonZeroU16> {
+        // SAFETY: It's always safe to read the current item because we don't
+        // ever move out of the array's bounds.
+        unsafe { NonZeroU16::new(*self.lpwstr.as_ptr()) }
+    }
+    /// Advance the iterator while `predicate` returns true.
+    /// Returns the number of items it advanced by.
+    fn advance_while<P: FnMut(NonZeroU16) -> bool>(&mut self, predicate: P) -> usize {
+        let mut predicate = predicate;
```
yaahc, line 213: TIOLI: Bit less repetitive and doesn't show up in the rustdoc signature

```suggestion
    fn advance_while<P: FnMut(NonZeroU16) -> bool>(&mut self, mut predicate: P) -> usize {
```

library/std/src/sys/windows/args.rs
```diff
@@ -28,129 +35,120 @@ pub fn args() -> Args {
...
+                // Skip whitespace.
+                code_units.advance_while(|w| w == SPACE || w == TAB);
             }
-            QUOTE if backslash_count % 2 == 0 => {
-                cur.extend(iter::repeat(b'\\' as u16).take(backslash_count / 2));
-                backslash_count = 0;
-                if was_in_quotes {
-                    cur.push('"' as u16);
-                    was_in_quotes = false;
+            // Backslashes can escape quotes or backslahes but only if consecutive backslashes are followed by a quote.
```
yaahc, line 115:
```suggestion
            // Backslashes can escape quotes or backslashes but only if consecutive backslashes are followed by a quote.
```

yaahc:
PR looks great @ChrisDenton. I caught a couple of typos you might want to fix but other than that I'm happy to merge this when you're ready.

@bors delegate+

library/std/src/sys/windows/args.rs
ChrisDenton: Hm, more extensive testing is slow because of the overhead of starting a process but perhaps some more limited examples can be tested. However, I wasn't sure about compiling and running a C++ within the testsuite. Is that ok?

yaahc:
One follow up, since this changes the behavior it would help to have an overview written up on how this changes the runtime behavior and why we feel this breakage was justified.

ChrisDenton:
Good point. I'm running local tests now and maybe I can figure out a summary in the meantime. I'm not very good a being concise but this is the gist of it:

Most of the parsing differences are minor and only really affect weird edge-cases. The big change is escaping `"` when already within a quote. This can be done by using two consecutive double quotes (i.e. `""`).

The main reason for updating the parsing rules is that it provides more consistent behaviour across modern (since 2008) utilities. This matters both for the user and for shells. E.g. powershell takes a string provided by the user, parses it according to powershell rules (expanding variables, etc), then produces a command line which it passes to the program. It expects the application to parse this command line in a standard way. So both powershell and the application should fully agree on how many arguments there are and what those arguments are.

However, the double quote escaping is mostly for users. This is of particular concern when using the command prompt (aka `cmd.exe`) which passes the command line directly to the program. Or even when creating a shortcut via the gui which allows directly entering arguments. Microsoft added the new `""` escaping rule to make this easier to follow, especially when nesting quotes and that the other escape character (`\`) is also a path separator. This [Stack Overflow answer](https://stackoverflow.com/a/15262019) has a few examples of where it can make code more legible.

yaahc:
That looks great, I've gone ahead and added a link to the overview comment in the PR description. Thank you again.

ChrisDenton:
I'd also add the small note that since Microsoft were willing to make the changes to their C/C++ argument parsing, they were presumably confident that it wouldn't be too disruptive. Of course that doesn't rule out the possibility of this change breaking something in the Rust ecosystem.

yaahc:
I just realized I should have FCPed this since it's a change to the stable API even though it's not adding anything new. In this case it's probably okay since we already discussed this a bit and the change seems pretty subtle and straightforward but I'm gonna cc @rust-lang/libs-api just to make sure everyone has a chance to see it.

ChrisDenton:
Ah, right. And I think I should have used r= there? Sorry, I'm a bit fuzzy about Bors commands.

yaahc:
> Ah, right. And I think I should have used r= there? Sorry, I'm a bit fuzzy about Bors commands.

Nonono, you didn't do anything wrong. I shouldn't have delighted bors before the FCP. You did everything right.

BurntSushi:
Are there some examples of what the breaking change here actually looks like?

One idea of a forcing function is to think about folks that maintain CLI applications. At some point, this change is going to get rolled into their release. An end user might then notice a change in how arguments are parsed and file a bug. At that point, if the maintainer isn't aware of a change like this, one might imagine that tracking down where the change stemmed from could be quite difficult. It might be helpful to characterize what the breaking changes look like, mitigations for them (if necessary) and include a mini-write-up for the release notes.

ChrisDenton:
@BurntSushi  Summary of the user-facing changes:

|Command line|old parsing|new parsing|
|--|--|--|
|`app "Hello""World" 123 456`|`["app", "Hello\"World 123 456"]`|`["app", "Hello\"World", "123", "456"]`|
|`C:\"Program Files"\app.exe`|`["C:\\\"Program", "Files\\app.exe"]`|`["C:\\Program Files\\app.exe"]`|

The big difference is that two double quotes are taken as one literal quote instead of also ending the block. This makes it more usable for escaping quotes.

The more minor differences are with parsing the zeroth argument (aka the application name). ASCII control characters are no longer considered whitespace (aside from tab). A slightly bigger differences is that a closing quote no longer ends the application name. The first change should have no impact on users because file names can't contain control characters. The second is essentially a bug fix so that both the command prompt and the application will agree on what the application name is.

Note that:
* The zeroth argument has different parsing rules to the other arguments. This has always been the case but they've been further tweaked here.
* These rules don't apply to shells such as powershell because they do their own parsing prior to passing the command line to the application. However, it does apply to `cmd.exe`.

BurntSushi:
@ChrisDenton Thanks! And yeah, that change looks quite sensible. Especially since it sounds like Windows itself has also done that change.

@yaahc Is it possible to still do an FCP for this? I don't see why it would be rejected, but this feels like a big enough visible change that it's probably worth going through the formal FCP process for this.

And I do think this should get put into the release notes somehow so that it gets visibility. My hope is that this will ultimately make things more sensible. For example, I know I've had a few issues filed against my CLIs in the past that basically amount to, "how do I quote things in a Windows shell," and I've been unable to give them a good answer. It sounds like a change like this would result in a bunch better user experience! We should advertise that. :-)

yaahc:
@rfcbot merge

Mark-Simulacrum:
It would be great to get a concise description for the release notes (just a comment on this PR is fine; linking it from the PR description would be great).

yaahc:
> It would be great to get a concise description for the release notes (just a comment on this PR is fine; linking it from the PR description would be great).

I asked and Chris already added an overview, which I think might be the same as what you mean, and it is linked in the PR description. Does that comment seem sufficient?

Mark-Simulacrum:
No, I don't think so; we usually want something closer to a 1-2 line description at most. This would likely go into the compatibility notes and/or library changes section, and look something like the language or library change descriptions in https://github.com/rust-lang/rust/blob/master/RELEASES.md#version-1530-2021-06-17 (e.g., adding unicode idents or into_iter).

That comment is likely useful if the release team selects this for inclusion in the blog post, though.

Something like this could work; presuming it's accurate:

> [Update Windows command line argument parsing](https://github.com/rust-lang/rust/pull/87580) to match the behavior of the standard libraries for C/C++. The rules have changed slightly over time, and this PR brings us to the latest set of rules (changed in 2008).

ChrisDenton:
> Something like this could work; presuming it's accurate:

The 2008 date comes [from here](https://daviddeley.com/autohotkey/parameters/parameters.htm#WINCRULESDOC) but I've not personally confirmed it (note it calls these tweaks "undocumented" but that [changed in 2020](https://docs.microsoft.com/en-us/cpp/cpp/main-function-command-line-args?view=msvc-160#parsing-c-command-line-arguments)). Whatever the date, I am confident this PR reflects the current C/C++ behavior,

BurntSushi:
@rfcbot reviewed

library/std/src/sys/windows/args.rs
```diff
@@ -1,23 +1,30 @@
-#![allow(dead_code)] // runtime init functions not used during testing
+// The Windows command line is just a string
+// https://docs.microsoft.com/en-us/archive/blogs/larryosterman/the-windows-command-line-is-just-a-string
+//
+// This module implements the parsing necessary to turn that string into a list of arguments.
```
bjorn3, line 4: Maybe
```suggestion
/// The Windows command line is just a string
/// https://docs.microsoft.com/en-us/archive/blogs/larryosterman/the-windows-command-line-is-just-a-string
///
/// This module implements the parsing necessary to turn that string into a list of arguments.
```
?

ChrisDenton: `//!` surely? But yes, that sound sensible.

bjorn3: Yeah, `//!`.

PR #87581, opened 2021-07-29
Add support for clobber_abi to asm!

### LINKED ISSUE #81092: asm! needs a better way of specifying clobbers
Clobbered registers are currently represented as [an output to `_`](https://doc.rust-lang.org/nightly/unstable-book/library-features/asm.html#clobbered-registers). However there are several limitations to this approach:
- It is impossible for registers that are not supported as inputs/outputs (e.g. x87 float registers, MMX registers, RISC-V vector registers, etc) to be marked as clobbered.
- ~~The set of registers available for clobbering is limited by the currently enabled target features. For example, it is desirable to mark SSE registers as clobbered even if SSE instructions are disabled (e.g. for a kernel). Having to wrap everything in `#[cfg(target_feature = "sse")]` is very inconvenient.~~ Fixed by #83841
- Lack of forwards compatibility with new registers that could be introduced with future ISA extensions. Consider the AVX-512 mark registers (`k0`-`k7`): if an `asm!` block calls a Rust function, that function may write to the mask registers. However the mask registers may not have been clobbered if the author of the `asm!` was not aware of AVX-512 when it was written.

Some potential solutions include:
- ~~Adding a separate `clobber("reg")` operand which is similar to the existing ones but doesn't require that the target feature for the register be enabled.~~ Fixed by #83841
- Adding some sort of wildcard clobber. For example `clobber("all")` would clobber all registers known to the compiler. This would be incompatible with register class specifiers: you would need to explicitly specify registers for all input/output operands.
- Adding ABI-based clobbers such as `clobber("C")` which clobbers all registers that would be clobbered by a call to an `extern "C" fn`.

It is not yet clear what the best solution is here. Discussion is welcome.

Amanieu:
This PR adds the `clobber_abi` feature that was proposed in #81092.

Fixes #81092

cc @rust-lang/wg-inline-asm 

r? @nagisa

### DIFF
--- a/compiler/rustc_target/src/asm/mod.rs
+++ b/compiler/rustc_target/src/asm/mod.rs
@@ -712,3 +712,185 @@ pub fn allocatable_registers(
         }
     }
 }
+
+#[derive(
+    Copy,
+    Clone,
+    Encodable,
+    Decodable,
+    Debug,
+    Eq,
+    PartialEq,
+    PartialOrd,
+    Hash,
+    HashStable_Generic
+)]
+pub enum InlineAsmClobberAbi {
+    X86,
+    X86_64Win,
+    X86_64SysV,
+    Arm,
+    AArch64,
+    RiscV,
+}
+
+impl InlineAsmClobberAbi {
+    /// Parses a clobber ABI for the given target, or returns a list of supported
+    /// clobber ABIs for the target.
+    pub fn parse(
+        arch: InlineAsmArch,
+        target: &Target,
+        name: Symbol,
+    ) -> Result<Self, &'static [&'static str]> {
+        let name = &*name.as_str();
+        match arch {
+            InlineAsmArch::X86 => match name {
+                "C" | "system" | "efiapi" | "cdecl" | "stdcall" | "fastcall" => {
+                    Ok(InlineAsmClobberAbi::X86)
+                }
+                _ => Err(&["C", "system", "efiapi", "cdecl", "stdcall", "fastcall"]),
+            },
+            InlineAsmArch::X86_64 => match name {
+                "C" | "system" if !target.is_like_windows => Ok(InlineAsmClobberAbi::X86_64SysV),
+                "C" | "system" if target.is_like_windows => Ok(InlineAsmClobberAbi::X86_64Win),
+                "win64" | "efiapi" => Ok(InlineAsmClobberAbi::X86_64Win),
+                "sysv64" => Ok(InlineAsmClobberAbi::X86_64SysV),
+                _ => Err(&["C", "system", "efiapi", "win64", "sysv64"]),
+            },
+            InlineAsmArch::Arm => match name {
+                "C" | "system" | "efiapi" | "aapcs" => Ok(InlineAsmClobberAbi::Arm),
+                _ => Err(&["C", "system", "efiapi", "aapcs"]),
+            },
+            InlineAsmArch::AArch64 => match name {
+                "C" | "system" | "efiapi" => Ok(InlineAsmClobberAbi::AArch64),
+                _ => Err(&["C", "system", "efiapi"]),
+            },
+            InlineAsmArch::RiscV32 | InlineAsmArch::RiscV64 => match name {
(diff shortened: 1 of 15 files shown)


compiler/rustc_target/src/asm/mod.rs
```diff
@@ -712,3 +712,183 @@ pub fn allocatable_registers(
...
+}
+
+impl InlineAsmClobberAbi {
+    /// Parses a clobber ABI for the given target, or returns a list of supported
+    /// clobber ABIs for the target.
+    pub fn parse(
+        arch: InlineAsmArch,
+        target: &Target,
+        name: Symbol,
+    ) -> Result<Self, &'static [&'static str]> {
```
nagisa, line 744: I think it is somewhat important that we utilize `Target::adjusted_abi` here so that e.g. ABIs such as `system` and `efiapi` work out of the box here.

compiler/rustc_target/src/asm/mod.rs
```diff
@@ -712,3 +712,183 @@ pub fn allocatable_registers(
...
+                "win64" => Ok(InlineAsmClobberAbi::X86_64Win),
+                "sysv64" => Ok(InlineAsmClobberAbi::X86_64SysV),
+                _ => Err(&["C", "win64", "sysv64"]),
+            },
+            InlineAsmArch::Arm => match name {
+                "C" | "aapcs" => Ok(InlineAsmClobberAbi::Arm),
+                _ => Err(&["C", "aapcs"]),
+            },
+            InlineAsmArch::AArch64 => match name {
+                "C" => Ok(InlineAsmClobberAbi::AArch64),
```
nagisa, line 763: Why do we handle aapcs for ARM but not aarch64?

nagisa (approved):
Seems reasonable enough, though some integration with existing abi machinery in rustc would be ideal. For example I could see adjust_cabi being changed so that is converts C to win64 or sysv64 or vice versa on the relevant systems.

compiler/rustc_target/src/asm/mod.rs
bjorn3: On AArch64

```rust
pub extern "aapcs" fn aapcs_func() {}
```

results in:

```
LLVM ERROR: Unsupported calling convention.
```

compiler/rustc_target/src/asm/mod.rs
```diff
@@ -712,3 +712,183 @@ pub fn allocatable_registers(
...
+    /// clobber ABIs for the target.
+    pub fn parse(
+        arch: InlineAsmArch,
+        target: &Target,
+        name: Symbol,
+    ) -> Result<Self, &'static [&'static str]> {
+        let name = &*name.as_str();
+        match arch {
+            InlineAsmArch::X86 => match name {
+                "C" | "cdecl" | "stdcall" | "fastcall" => Ok(InlineAsmClobberAbi::X86),
```
bjorn3, line 748: Can you handle the "system" abi on all archs too? This matches whatever abi is used on the respective platform to call into system libraries, so the same as the "C" abi on most targets, but on 32bit x86 windows it is the stdcall calling convention.

compiler/rustc_target/src/asm/mod.rs
Amanieu: `aapcs` is only a valid calling convention on ARM, not AArch64.

compiler/rustc_target/src/asm/mod.rs
Amanieu: Sure, I can add "system" and even "efiapi" to this list. It makes no difference in terms of which registers are actually clobbered since `cdecl` and `stdcall` clobber the same set of registers, and on other arches they are identical to "C".

compiler/rustc_target/src/asm/mod.rs
Amanieu: I tried using `Abi` directly but the code got a bit messy. In any case we only support a small subset of ABIs for `clobber_abi`, I'm not sure if it makes sense to use the existing ABI machinery.

PR #87583, opened 2021-07-29
Refactor compression cache in v0 symbol mangler

tmiasko:
* Remove redundant option around compression caches (they are always present).
* Flatten compression caches into symbol mangler to avoid dynamic memory allocation.
* Implement printer for `&mut SymbolMangler` instead of `SymbolMangler` to avoid passing now slightly larger symbol mangler by value.

### DIFF
--- a/compiler/rustc_symbol_mangling/src/v0.rs
+++ b/compiler/rustc_symbol_mangling/src/v0.rs
@@ -23,7 +23,7 @@ pub(super) fn mangle(
     let substs = tcx.normalize_erasing_regions(ty::ParamEnv::reveal_all(), instance.substs);
 
     let prefix = "_R";
-    let mut cx = SymbolMangler {
+    let mut cx = &mut SymbolMangler {
         tcx,
         start_offset: prefix.len(),
         paths: FxHashMap::default(),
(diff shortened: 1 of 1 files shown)


wesleywiser (approved):

compiler/rustc_symbol_mangling/src/v0.rs
```diff
@@ -81,9 +68,15 @@ struct BinderLevel {
 
 struct SymbolMangler<'tcx> {
     tcx: TyCtxt<'tcx>,
-    compress: Option<Box<CompressionCaches<'tcx>>>,
     binders: Vec<BinderLevel>,
     out: String,
+
+    /// The length of the prefix in `out` (e.g. 2 for `_R`).
+    start_offset: usize,
+    /// The values are start positions in `out`, in bytes.
```
eddyb, line 76: Comment was not doc comment because it applies to all 3 fields.
Also, the names of the fields should've probably been adjusted (or the separate `struct` kept), since now they've lost a bit of context.

eddyb:
> Remove redundant option around compression caches (they are always present).

I wonder if I should've have made `-Z symbol-mangling-version=v0-uncompressed` access the `None` case.
Anyway, this was mostly for testing, I don't think we care much about uncompressed `v0` going forward.

compiler/rustc_symbol_mangling/src/v0.rs
tmiasko: Ack. Will keep this in mind when making further changes there.

I was looking into the idea of using only shortening back-references, but differences seems to be negligible. For cargo, total symbol size reduction of -0.016% (`nm target/debug/cargo | grep -o '_R.*' | wc --bytes`). Could be more useful with const generics, I suppose.

PR #87587, opened 2021-07-29
Various refactorings of the TAIT infrastructure

oli-obk:
Before this PR we used to store the opaque type knowledge outside the `InferCtxt`, so it got recomputed on every opaque type instantiation.

I also removed a feature gate check that makes no sense in the planned lazy TAIT resolution scheme

Each commit passes all tests, so this PR is best reviewed commit by commit.

r? @spastorino

### DIFF
--- a/compiler/rustc_mir/src/borrow_check/mod.rs
+++ b/compiler/rustc_mir/src/borrow_check/mod.rs
@@ -105,7 +105,7 @@ fn mir_borrowck<'tcx>(
     let (input_body, promoted) = tcx.mir_promoted(def);
     debug!("run query mir_borrowck: {}", tcx.def_path_str(def.did.to_def_id()));
 
-    let opt_closure_req = tcx.infer_ctxt().enter(|infcx| {
+    let opt_closure_req = tcx.infer_ctxt().with_opaque_type_inference(def.did).enter(|infcx| {
         let input_body: &Body<'_> = &input_body.borrow();
         let promoted: &IndexVec<_, _> = &promoted.borrow();
         do_mir_borrowck(&infcx, input_body, promoted)
--- a/compiler/rustc_trait_selection/src/opaque_types.rs
+++ b/compiler/rustc_trait_selection/src/opaque_types.rs
@@ -2,11 +2,11 @@ use crate::infer::InferCtxtExt as _;
 use crate::traits::{self, ObligationCause, PredicateObligation};
 use rustc_data_structures::fx::FxHashMap;
 use rustc_data_structures::sync::Lrc;
-use rustc_data_structures::vec_map::VecMap;
 use rustc_hir as hir;
 use rustc_hir::def_id::{DefId, LocalDefId};
 use rustc_infer::infer::error_reporting::unexpected_hidden_region_diagnostic;
 use rustc_infer::infer::free_regions::FreeRegionRelations;
+use rustc_infer::infer::opaque_types::{OpaqueTypeDecl, OpaqueTypeMap};
 use rustc_infer::infer::type_variable::{TypeVariableOrigin, TypeVariableOriginKind};
 use rustc_infer::infer::{self, InferCtxt, InferOk};
 use rustc_middle::ty::fold::{BottomUpFolder, TypeFoldable, TypeFolder, TypeVisitor};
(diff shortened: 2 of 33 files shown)


compiler/rustc_mir/src/borrow_check/mod.rs
```diff
@@ -105,7 +105,7 @@ fn mir_borrowck<'tcx>(
     let (input_body, promoted) = tcx.mir_promoted(def);
     debug!("run query mir_borrowck: {}", tcx.def_path_str(def.did.to_def_id()));
 
-    let opt_closure_req = tcx.infer_ctxt().enter(|infcx| {
+    let opt_closure_req = tcx.infer_ctxt().with_opaque_type_inference(def.did).enter(|infcx| {
```
spastorino, line 108: Not sure I love this approach but unsure neither how to make it better. One could forget calling `with_opaqie_type_inference`, would rather have a way to force us pass the `def_id` down always.

src/test/ui/type-alias-impl-trait/issue-63279.stderr
```diff
@@ -1,12 +1,17 @@
...
-   |                ^^^^^^^^^^^^^ expected `()`, found opaque type
+   |                ------------- the found opaque type
+...
+LL |     || -> Closure { || () }
+   |     ^^^^^^^^^^^^^^^^^^^^^^^ expected closure, found a different closure
    |
-   = note: expected unit type `()`
-            found opaque type `impl FnOnce<()>`
+   = note: expected type `[closure@$DIR/issue-63279.rs:8:21: 8:26]`
+           found closure `[closure@$DIR/issue-63279.rs:8:5: 8:28]`
```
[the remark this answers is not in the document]
oli-obk: This is tracked in https://github.com/rust-lang/rust/issues/19939

closure type printing is suboptimal in general

compiler/rustc_mir/src/borrow_check/mod.rs
oli-obk: Forgetting is safe, the thing just stops being a defining use. We can always add more sites later.

We should also only ever do this in mir borrowck's "type check" pass and in the real type check pass, so both are covered. I will document this information in the function's docs.

src/test/ui/type-alias-impl-trait/issue-60371.stderr
```diff
@@ -7,15 +7,6 @@ LL |     type Item = impl Bug;
...
 
-error[E0658]: type alias impl trait is not permitted here
-  --> $DIR/issue-60371.rs:13:40
-   |
-LL |     const FUN: fn() -> Self::Item = || ();
-   |                                        ^
-   |
-   = note: see issue #63063 <https://github.com/rust-lang/rust/issues/63063> for more information
-   = help: add `#![feature(type_alias_impl_trait)]` to the crate attributes to enable
-
```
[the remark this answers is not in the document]
oli-obk: This is something that will be allowed under the new scheme

spastorino:
@oli-obk I think this should be go to go. Feel free to `r=me`

compiler/rustc_trait_selection/src/opaque_types.rs
```diff
@@ -403,19 +321,36 @@ impl<'a, 'tcx> InferCtxtExt<'tcx> for InferCtxt<'a, 'tcx> {
...
+        // the fn body). (Ultimately, writeback is responsible for this
+        // check.)
+        let bounds = tcx.explicit_item_bounds(def_id);
+        debug!("{:#?}", bounds);
+        let bounds: Vec<_> =
+            bounds.iter().map(|(bound, _)| bound.subst(tcx, opaque_type_key.substs)).collect();
+        debug!("{:#?}", bounds);
+        let opaque_type = tcx.mk_opaque(def_id, opaque_type_key.substs);
 
+        let required_region_bounds = required_region_bounds(tcx, opaque_type, bounds.into_iter());
```
klensy, line 352: Collecting bounds into Vec only to later pass it as iterator, can we didn't collect it at all?

oli-obk: yea good point. I was mainly trying to keep the changes as small as possible, but I should clean up afterwards.

ehuss:
@bors retry

macos runner exited without logs after 1h 47m 35s.

oli-obk:
The perf regression here is expected and I am keeping an eye on it. I have ideas how to resolve it once my tait refactoring is done

rylev:
@oli-obk would it be worth it to track the perf regression in an issue? Then we can mark this as being addressed by adding the `perf-regression-triaged` labeled.

oli-obk:
We could add it as a bullet point to the steps of https://github.com/rust-lang/rust/issues/63063

Mark-Simulacrum:
I'm going to actually go ahead and mark this regression as triaged. I did some investigation into where the increased instruction counts are coming from, and it looks like it's pretty much entirely down to more obligation processing. Even if we believe there's a possible workaround (e.g., the resolution @oli-obk suggests), I don't think it's useful to keep this regression marked as untriaged in the interim.

The regression fixes here are likely to come about as more general work -- this PR did not introduce some "particularly slow" code or anything like that.

Issue #64102, opened 2019-09-02
Tracking issue for `std::hint::black_box`


Centril:
This is a tracking issue for the RFC `std::hint:_black_box`.

Original RFC: [RFC 2360](https://rust-lang.github.io/rfcs/2360-bench-black-box.html)

**Public API:**

```rust
// std::hint

pub fn black_box<T>(dummy: T) -> T;
```

**Steps:**

- [x] Implementation
- [x] FCP
- [ ] Stabilization PR

**Unresolved questions:**

- [ ] `const fn`: it is unclear whether `bench_black_box` should be a `const fn`. If it
  were, that would hint that it cannot have any side-effects, or that it cannot
  do anything that `const fn`s cannot do. 

- [ ] Naming: during the RFC discussion it was unclear whether `black_box` is the
  right name for this primitive but we settled on `bench_black_box` for the time
  being. We should resolve the naming before stabilization.
  
  Also, we might want to add other benchmarking hints in the future, like
  `bench_input` and `bench_output`, so we might want to put all of this
  into a `bench` sub-module within the `core::hint` module. That might
  be a good place to explain how the benchmarking hints should be used 
  holistically.
  
  Some arguments in favor or against using "black box" are that:
     * pro: [black box] is a common term in computer programming, that conveys
       that nothing can be assumed about it except for its inputs and outputs.
       con: [black box] often hints that the function has no side-effects, but
       this is not something that can be assumed about this API.
     * con: `_box` has nothing to do with `Box` or `box`-syntax, which might be confusing
 
  Alternative names suggested: `pessimize`, `unoptimize`, `unprocessed`, `unknown`,
  `do_not_optimize` (Google Benchmark).

Centril:
I believe the main implementation work remaining is to enhance the documentation of `hint::black_box` as well as to rename it.

eddyb:
An idea I wish I had earlier (but we can still try it out):
1. use an intrinsic (required anyway for nice `wasm`/`miri` support)
2. when codegening the intrinsic in a non-`--test` crate:
   * don't generate the inline assembly (so LLVM can optimize right through it)
   * emit a warning that it should only be used in a test crate

Note that because this is a warning, and not an error, it doesn't add to the dreaded post-monomorphization errors. We could also just do 2. but not 1, if we *need* to.

`#[inline]` and/or generics could then be used to delay the codegen until the `--test` crate, so it can do its intended purpose.

[...]

PR #87590, opened 2021-07-29
Deprecate llvm_asm!

Amanieu:
We would like to remove `llvm_asm!` from the compiler once `asm!` is stabilized. This PR deprecates `llvm_asm!` to encourage any remaining users to migrate to `asm!` (or if `asm!` is not supported for their target, to add this support to rustc).

The only remaining user of `llvm_asm!` in the standard library was `black_box`, which has been rewritten to use volatile operations when `asm!` is not available on the current target.

cc @rust-lang/wg-inline-asm 

cc @RalfJung for the changes to `black_box` which might affect Miri.

r? @nagisa

### DIFF
--- a/library/core/src/hint.rs
+++ b/library/core/src/hint.rs
@@ -155,6 +155,7 @@ pub fn spin_loop() {
 #[inline]
 #[unstable(feature = "bench_black_box", issue = "64102")]
 #[cfg_attr(not(bootstrap), allow(unused_mut))]
+#[cfg_attr(bootstrap, allow(deprecated))]
 pub fn black_box<T>(mut dummy: T) -> T {
     #[cfg(bootstrap)]
     // SAFETY: the inline assembly is a no-op.
--- a/src/test/ui/abi/abi-sysv64-register-usage.rs
+++ b/src/test/ui/abi/abi-sysv64-register-usage.rs
@@ -7,6 +7,7 @@
 // ignore-aarch64
 
 #![feature(llvm_asm)]
+#![allow(deprecated)] // llvm_asm!
 
 #[cfg(target_arch = "x86_64")]
 pub extern "sysv64" fn all_the_registers(rdi: i64, rsi: i64, rdx: i64,
--- a/library/core/src/macros/mod.rs
+++ b/library/core/src/macros/mod.rs
@@ -1347,6 +1347,10 @@ pub(crate) mod builtin {
         issue = "70173",
         reason = "prefer using the new asm! syntax instead"
     )]
+    #[rustc_deprecated(
+        since = "1.56",
+        reason = "will be removed from the compiler, use asm! instead"
+    )]
     #[rustc_builtin_macro]
     #[macro_export]
     macro_rules! llvm_asm {
--- a/src/test/ui/asm/naked-functions.rs
+++ b/src/test/ui/asm/naked-functions.rs
@@ -4,6 +4,7 @@
 #![feature(naked_functions)]
 #![feature(or_patterns)]
 #![crate_type = "lib"]
+#![allow(deprecated)] // llvm_asm!
 
 #[repr(C)]
 pub struct P { x: u8, y: u16 }
--- a/src/test/ui/asm/naked-functions.stderr
+++ b/src/test/ui/asm/naked-functions.stderr
@@ -1,35 +1,35 @@
 error: asm with the `pure` option must have at least one output
-  --> $DIR/naked-functions.rs:126:14
+  --> $DIR/naked-functions.rs:127:14
    |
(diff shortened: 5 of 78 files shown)


joshtriplett:
For simplicity's sake, is there any reason not to just use volatile operations on all platforms, and drop the complex and duplicated cfg condition?

Amanieu:
Using `asm!` can be slightly faster if the value is already in memory. Also it can be further optimized in the future as suggested here: https://github.com/rust-lang/rust/issues/64102#issuecomment-708880328

library/core/src/hint.rs
```diff
@@ -152,23 +152,79 @@ pub fn spin_loop() {
...
+        dummy
     }
 
-    dummy
+    // Use volatile operations on targets that don't support inline assembly.
+    // This effectively simulates:
+    // 1. passing the value to some external process
+    // 2. allowing that process to access any memory that the value may point to
+    // 3. reading a potentially modified value back
+    #[cfg(any(
```
nagisa, line 200: My concern with this implementation is that once support for assembly is added for any of these architectures, it is trivial to miss updating this location.

Any way we can avoid this? We _could_ make up an intrinsic for this, perhaps?

nagisa:
I think the direction is reasonable. See the concern above, once its resolved and the tests are fixed, r=me.

library/core/src/hint.rs
RalfJung: Instead of using the fallback for all *other* architectures, could we have an explicit list of architectures that *do* use the fallback?

Amanieu: > Instead of using the fallback for all other architectures, could we have an explicit list of architectures that do use the fallback?

This will likely cause problems when people try to use a new architecture with a custom json target.

> Any way we can avoid this? We could make up an intrinsic for this, perhaps?

We could have a `target_has_asm` cfg flag like we do for atomics. But it seems a bit excessive to have this just for `black_box`.

joshtriplett:
On Thu, Jul 29, 2021 at 11:19:22AM -0700, Amanieu wrote:
> We could have a `target_has_asm` cfg flag like we do for atomics. But it seems a bit excessive to have this just for `black_box`.

I think it might be worthwhile regardless, and I can easily imagine
other uses of that as well.

Alternatively, is the performance improvement of using `asm!`
sufficiently large to be worth the maintenance headache? Using volatile
everywhere would leave `black_box` still entirely usable for benchmarks.

Amanieu:
The main use of `black_box` is in micro-benchmarking where you want it to have as little impact as possible on the benchmark. I would argue that performance is pretty important in this case.

There are more optimizations that can be done to `black_box`, such as passing the value directly in a register instead of just its address to avoid forcing a store to memory. However this is not the main point of this PR which is to deprecate `llvm_asm!` and remove its uses from the standard library.

RalfJung:
> cc @RalfJung for the changes to black_box which might affect Miri.

Looks good -- though Miri doesn't optimize, so skipping the function entirely like before would also work. No strong opinion either way.

joshtriplett:
On Fri, Jul 30, 2021 at 07:33:03AM -0700, Amanieu wrote:
> The main use of `black_box` is in micro-benchmarking where you want it to have as little impact as possible on the benchmark. I would argue that performance is pretty important in this case.

Then I think, for maintainability reasons, it'd be worth adding a
`cfg(target_has_asm)` or similar, even if it's internal-use-only for
now. That long architecture list seems like a painful maintenance
hazard.

library/core/src/hint.rs
nagisa: Having thought about this more, an intrinsic seems significantly more straightforward to me. But I guess I'm okay with this implementation too.

If we continue using cfgs here, consider `cfg_if!` maybe? It'll reduce the clutter here significantly I feel.

Amanieu:
Should be good to go, however this will conflict with #87916 due to the need for an `allow(deprecated)` for `black_box`.

src/test/ui/abi/abi-sysv64-register-usage.rs
```diff
@@ -7,6 +7,7 @@
 // ignore-aarch64
 
 #![feature(llvm_asm)]
+#![allow(deprecated)] // llvm_asm!
```
nbdd0121, line 10: Shouldn't this test be modified to use `asm!` instead? This test isn't testing `llvm_asm!` itself

nagisa: Perfectly fine to leave removal to a follow-up PR (potentially the same one that removes `llvm_asm` entirely.

Issue #87318, opened 2021-07-20
Un-skipped whitespace after backslash in string literal should be warned about


scottmcm:
```rust
let b = "foo\
         bar";
let c = "foo\
         bar";
dbg!((b,c));
```
https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=a8ab2d71eb82732581a9b6aedde68aed
```
[src/main.rs:6] (b, c) = (
    "foobar",
    "foo\u{a0}      bar",
)
```

Like we have confusables lints, I think this should have a warning noting that the particular kind of whitespace is not actually being skipped here, since that's completely invisible when looking at the code.

(Based on https://github.com/rust-lang/reference/pull/1042 I think this would warn for things where `char::is_whitespace` is true but that isn't ' ', '\t', '\r', or '\n' -- with https://github.com/rust-lang/rust/issues/87319 being a related lint that would warn on '\r' and '\n' after the first one.)

JohnTitor:
Fixed by #87596.

Issue #87319, opened 2021-07-20
Warn about strings that skip multiple newlines with one backslash


joshtriplett:
(This came up in today's @rust-lang/lang meeting.)

Given the following code:

```rust
fn main() {
    let s1 = "a\

b";
    let s2 = "a\

    b";
    println!("{:?} {:/}", s1, s2);
}
```

[Playground link](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=a783aa21f2ec4064d0ae0cab0f4c8777)

The compiler interprets both of the string literals as `"ab"`, omitting multiple newlines after the `\`.

This seems confusing, and different from the behavior of other languages. For instance, C and Python both produce parse errors for such string literals. Or, for a Python multi-line string literal, the backslash escapes the immediately following newline but not the subsequent newline.

Rust does allow unescaped newlines in string literals, which is *useful*. The ability to escape a newline can also be useful sometimes. But the combination of both escaped and unescaped newlines seems sufficiently confusing and surprising to merit a warning.

We'd ideally like to work towards this being an error. Depending on the presence of existing usage in the ecosystem, that might or might not require an edition.

pierwill:
Discussed in [Zulip](https://rust-lang.zulipchat.com/#narrow/stream/182449-t-compiler.2Fhelp/topic/issue.2087319):

> The escape sequence handling for string literals is handled here: https://github.com/rust-lang/rust/blob/1158367a6d60d3e2a8c68d6212a8007992b1dbf0/compiler/rustc_lexer/src/unescape.rs#L267
> 
> It looks like the issue lies in https://github.com/rust-lang/rust/blob/1158367a6d60d3e2a8c68d6212a8007992b1dbf0/compiler/rustc_lexer/src/unescape.rs#L300 - we consider \n to be 'ascii whitespace' and will therefore skip over it after we encounter an escaped newline

PR #87596, opened 2021-07-29
Add warning when whitespace is not skipped after an escaped newline

jesyspa:
Fixes issue #87318, also simplifies issue #87319.

* Add support to the lexer to emit warnings as well as errors.
* Emit a warning when a string literal contains an escaped newline, but when (some of) the whitespace on the next line is not skipped due to it being non-ASCII.

### DIFF
--- a/compiler/rustc_lexer/src/unescape.rs
+++ b/compiler/rustc_lexer/src/unescape.rs
@@ -7,7 +7,7 @@ use std::str::Chars;
 #[cfg(test)]
 mod tests;
 
-/// Errors that can occur during string unescaping.
+/// Errors and warnings that can occur during string unescaping.
 #[derive(Debug, PartialEq, Eq)]
 pub enum EscapeError {
     /// Expected 1 char, but 0 were found.
@@ -56,6 +56,20 @@ pub enum EscapeError {
     NonAsciiCharInByte,
     /// Non-ascii character in byte string literal.
     NonAsciiCharInByteString,
+
+    /// After a line ending with '\', the next line contains whitespace
+    /// characters that are not skipped.
+    UnskippedWhitespaceWarning,
+}
+
+impl EscapeError {
+    /// Returns true for actual errors, as opposed to warnings.
+    pub fn is_fatal(&self) -> bool {
+        match self {
+            EscapeError::UnskippedWhitespaceWarning => false,
+            _ => true,
+        }
+    }
 }
 
 /// Takes a contents of a literal (without quotes) and produces a
@@ -283,7 +297,7 @@ where
                         // if unescaped '\' character is followed by '\n'.
                         // For details see [Rust language reference]
                         // (https://doc.rust-lang.org/reference/tokens.html#string-literals).
-                        skip_ascii_whitespace(&mut chars);
+                        skip_ascii_whitespace(&mut chars, start, callback);
                         continue;
                     }
                     _ => scan_escape(first_char, &mut chars, mode),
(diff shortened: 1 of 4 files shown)


jesyspa:
Prior to this PR, the lexer only has support for emitting errors.  I've added warnings by changing the callback, but I'm not sure it's the best approach; please let me know if this is too messy.

One other option would be to pass an emit_char and an emit_err callback, since those require mostly orthogonal code, but that would require a bit more reworking of the `unescape.rs` internals.

jesyspa:
Sorry, I missed the clippy code when testing locally.  I'll update the PR and re-run the checks.

compiler/rustc_lexer/src/unescape.rs
```diff
@@ -294,16 +302,31 @@ where
             _ => scan_escape(first_char, &mut chars, mode),
         };
         let end = initial_len - chars.as_str().len();
-        callback(start..end, unescaped_char);
+        callback(start..end, unescaped_char.map(Ok));
     }
 
-    fn skip_ascii_whitespace(chars: &mut Chars<'_>) {
+    fn skip_ascii_whitespace<F>(chars: &mut Chars<'_>, _start: usize, _callback: &mut F)
```
estebank, line 308:
```suggestion
    fn skip_ascii_whitespace<F>(chars: &mut Chars<'_>, start: usize, callback: &mut F)
```

compiler/rustc_lexer/src/unescape.rs
```diff
@@ -294,16 +302,31 @@ where
...
             .unwrap_or(str.len());
-        *chars = str[first_non_space..].chars()
+        let tail = &str[first_non_space..];
+        if let Some(c) = tail.chars().nth(0) {
+            // For error reporting, we would like the span to contain the character that was not
+            // skipped.  The +1 is necessary to account for the leading \ that started the escape.
+            let end = _start + first_non_space + c.len_utf8() + 1;
+            if c.is_whitespace() {
+                _callback(
+                    _start..end,
```
estebank, line 324:
```suggestion
            let end = start + first_non_space + c.len_utf8() + 1;
            if c.is_whitespace() {
                callback(
                    start..end,
```

compiler/rustc_lexer/src/unescape.rs
```diff
@@ -58,19 +58,27 @@ pub enum EscapeError {
...
 }
 
+/// Wranings that can be emitted during string unescaping.
+#[derive(Debug, PartialEq, Eq)]
+pub enum EscapeWarning {
+    /// After a line ending with '\', the next line contains whitespace
+    /// characters that are not skipped.
+    UnskippedWhitespaceAfterEscapedNewline,
+}
+
```
estebank, line 68: In principle I'm ok with the approach of nested `Result`s, but in this case it feels like a better way would be to make `UnskippedWhitespaceAfterEscapedNewline` a variant of `EscapeError`. I know it muddles the waters a little bit by mixing warnings and errors, but I feel it is worth it due to the flattening of the code in this PR.

compiler/rustc_lexer/src/unescape.rs
jesyspa: Thanks, sorry about that, I thought I got a warning because of the shadowing.

compiler/rustc_lexer/src/unescape.rs
jesyspa: Makes sense, I've changed it to be a single type.  Should I also change the name to EscapeDiagnostic, or is that too vague?

estebank: I feel like `EscapeError` is slightly misleading, but it'll make more sense than `EscapeDiagnostic`.

Issue #87555, opened 2021-07-28
Tracking Issue for RFC2509: concat_bytes!()


m-ou-se:
Feature gate: `#![feature(concat_bytes)]`

This is a tracking issue for https://github.com/rust-lang/rfcs/pull/2509

### Public API

```rust
#[macro_export]
macro_rules! concat_bytes { .. }
```

### Steps / History

- [x] RFC: https://github.com/rust-lang/rfcs/pull/2509
- [ ] Implementation
- [ ] Final comment period (FCP)
- [ ] Stabilization PR

### Unresolved Questions

- Should additional literal types be supported? Byte string literals are basically the same thing as byte slice references, so it might make sense to support those as well (support `&[0, 1, 2]` in addition to `[0, 1, 2]`).
- What to do with string and character literals? They could either be supported with their underlying UTF-8 representation being concatenated, or rejected.

Smittyvb:
I'm working on an implementation at [Smittyvb:concat_bytes](https://github.com/Smittyvb/rust/tree/concat_bytes).

Smittyvb:
Implementation PR: #87599

Fishrock123:
Regarding the unresolved questions - can we reject regular string slices if we accept `&[]` since `str::as_bytes()` is const?

Smittyvb:
@Fishrock123 Macro arguments are not a constant evaluation context, so calls to `const fn`s in macros result in the tokens representing the function call getting passed to the macro, not the result of evaluating the function call.

For example,
```rust
const fn add_one(x: u8) -> u8 { x + 1 }
macro_rules! demo {
    ($x:expr) => { let x = $x; }
}
trace_macros!(true);
demo!(add_one(2));
```
gives the output
```
note: trace_macro
 --> src/main.rs:8:1
  |
8 | demo!(add_one(2));
  | ^^^^^^^^^^^^^^^^^^
  |
  = note: expanding `demo! { add_one(2) }`
  = note: to `let x = add_one(2) ;`
```
The `add_one(2)` call never gets expanded to `3`, since it's not used in a constant evaluation context. (although LLVM will optimise that)

octylFractal:
@m-ou-se should the implementation box be checked off now? It seems the PR to implement it was merged.

mkpankov:
Was FCP started?

Mythra:
[...]

PR #87599, opened 2021-07-29
Implement concat_bytes!

syvb:
This implements the unstable `concat_bytes!` macro, which has tracking issue #87555. It can be used like:
```rust
#![feature(concat_bytes)]

fn main() {
    assert_eq!(concat_bytes!(), &[]);
    assert_eq!(concat_bytes!(b'A', b"BC", [68, b'E', 70]), b"ABCDEF");
}
```
If strings or characters are used where byte strings or byte characters are required, it suggests adding a `b` prefix. If a number is used outside of an array it suggests arrayifying it. If a boolean is used it suggests replacing it with the numeric value of that number. Doubly nested arrays of bytes are disallowed.

### DIFF
--- a/compiler/rustc_builtin_macros/src/concat_bytes.rs
+++ b/compiler/rustc_builtin_macros/src/concat_bytes.rs
@@ -1,0 +1,167 @@
+use rustc_ast as ast;
+use rustc_ast::{ptr::P, tokenstream::TokenStream};
+use rustc_data_structures::sync::Lrc;
+use rustc_errors::Applicability;
+use rustc_expand::base::{self, DummyResult};
+
+/// Emits errors for literal expressions that are invalid inside and outside of an array.
+fn invalid_type_err(cx: &mut base::ExtCtxt<'_>, expr: &P<rustc_ast::Expr>, is_nested: bool) {
+    let lit = if let ast::ExprKind::Lit(lit) = &expr.kind {
+        lit
+    } else {
+        unreachable!();
+    };
+    match lit.kind {
+        ast::LitKind::Char(_) => {
+            let mut err = cx.struct_span_err(expr.span, "cannot concatenate character literals");
+            if let Ok(snippet) = cx.sess.source_map().span_to_snippet(expr.span) {
+                err.span_suggestion(
+                    expr.span,
+                    "try using a byte character",
+                    format!("b{}", snippet),
+                    Applicability::MachineApplicable,
+                )
+                .emit();
+            }
+        }
+        ast::LitKind::Str(_, _) => {
+            let mut err = cx.struct_span_err(expr.span, "cannot concatenate string literals");
+            // suggestion would be invalid if we are nested
+            if !is_nested {
+                if let Ok(snippet) = cx.sess.source_map().span_to_snippet(expr.span) {
+                    err.span_suggestion(
+                        expr.span,
+                        "try using a byte string",
+                        format!("b{}", snippet),
+                        Applicability::MachineApplicable,
+                    );
+                }
+            }
+            err.emit();
+        }
+        ast::LitKind::Float(_, _) => {
+            cx.span_err(expr.span, "cannot concatenate float literals");
+        }
+        ast::LitKind::Bool(_) => {
+            cx.span_err(expr.span, "cannot concatenate boolean literals");
+        }
+        ast::LitKind::Err(_) => {}
+        ast::LitKind::Int(_, _) if !is_nested => {
+            let mut err = cx.struct_span_err(expr.span, "cannot concatenate numeric literals");
+            if let Ok(snippet) = cx.sess.source_map().span_to_snippet(expr.span) {
+                err.span_suggestion(
+                    expr.span,
+                    "try wrapping the number in an array",
+                    format!("[{}]", snippet),
+                    Applicability::MachineApplicable,
+                );
+            }
+            err.emit();
+        }
+        ast::LitKind::Int(
+            val,
+            ast::LitIntType::Unsuffixed | ast::LitIntType::Unsigned(ast::UintTy::U8),
+        ) => {
+            assert!(val > u8::MAX.into()); // must be an error
+            cx.span_err(expr.span, "numeric literal is out of bounds");
+        }
+        ast::LitKind::Int(_, _) => {
+            cx.span_err(expr.span, "numeric literal is not a `u8`");
+        }
+        _ => unreachable!(),
+    }
+}
+
+pub fn expand_concat_bytes(
+    cx: &mut base::ExtCtxt<'_>,
+    sp: rustc_span::Span,
+    tts: TokenStream,
+) -> Box<dyn base::MacResult + 'static> {
+    let es = match base::get_exprs_from_tts(cx, sp, tts) {
+        Some(e) => e,
+        None => return DummyResult::any(sp),
+    };
+    let mut accumulator = Vec::new();
+    let mut missing_literals = vec![];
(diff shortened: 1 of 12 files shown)


Fishrock123:
Just pointing out that the RFC contains a bit of a contradiction and this PR's tests don't appear to test that case: https://github.com/rust-lang/rfcs/pull/2509#issuecomment-889346645

Smittyvb:
@Fishrock123 That appears to have been an oversight when updating the RFC. I've added a test for it.

compiler/rustc_builtin_macros/src/concat_bytes.rs
```diff
@@ -0,0 +1,210 @@
...
+    let mut missing_literal = vec![];
+    let mut has_errors = false;
+    for e in es {
+        match e.kind {
+            ast::ExprKind::Array(ref exprs) => {
+                for expr in exprs {
+                    match expr.kind {
+                        ast::ExprKind::Array(_) => {
+                            cx.span_err(expr.span, "cannot concatenate doubly nested array");
+                            has_errors = true;
```
Mark-Simulacrum, line 26: Hm, can we bundle up the spans for the wrong-type errors and give a single multi-span error or something? Otherwise it gets a bit noisy.

Alternatively it might make sense to stop complaining about further errors after the first one...

compiler/rustc_builtin_macros/src/concat_bytes.rs
```diff
@@ -0,0 +1,210 @@
...
+                                        "try using a byte character",
+                                        format!("b{}", snippet),
+                                        Applicability::MachineApplicable,
+                                    )
+                                    .emit();
+                                } else {
+                                    cx.span_err(
+                                        expr.span,
+                                        "cannot concatenate an array of character literals",
+                                    )
```
Mark-Simulacrum, line 61: Can we avoid duplicating the error creation across this if? We should be able to give the suggestion only if we have the snippet, I think.

compiler/rustc_builtin_macros/src/concat_bytes.rs
```diff
@@ -0,0 +1,210 @@
...
+                            }
+                            ast::LitKind::Int(
+                                val,
+                                ast::LitIntType::Unsuffixed
+                                | ast::LitIntType::Unsigned(ast::UintTy::U8),
+                            ) if val <= 255 => {
+                                accumulator.push(val as u8);
+                            }
+                            ast::LitKind::Int(_, _) => {
+                                cx.span_err(expr.span, "numeric literal is out of bounds");
```
Mark-Simulacrum, line 73: I think this will give an out of bounds error on e.g. `5u16`, which feels a bit unfortunate: can we provide a dedicated error for specifically the non-U8 case, regardless of the value?

compiler/rustc_builtin_macros/src/concat_bytes.rs
```diff
@@ -0,0 +1,210 @@
...
+                                        "cannot concatenate an array of character literals",
+                                    )
+                                }
+                                has_errors = true;
+                            }
+                            ast::LitKind::Int(
+                                val,
+                                ast::LitIntType::Unsuffixed
+                                | ast::LitIntType::Unsigned(ast::UintTy::U8),
+                            ) if val <= 255 => {
```
Mark-Simulacrum, line 69: nit: u8::MAX seems better than 255.

compiler/rustc_builtin_macros/src/concat_bytes.rs
```diff
@@ -0,0 +1,210 @@
...
+                            ast::LitKind::Bool(val) => {
+                                cx.struct_span_err(
+                                    expr.span,
+                                    "cannot concatenate an array of boolean literals",
+                                )
+                                .span_suggestion(
+                                    expr.span,
+                                    "try using the numeric value",
+                                    (val as u8).to_string(),
+                                    Applicability::MachineApplicable,
```
Mark-Simulacrum, line 85: I think this is *probably* not a good suggestion to give; it seems unlikely that the user wanted `concat_bytes!(true)` to give them `[1]`. I would just leave the error.

compiler/rustc_builtin_macros/src/concat_bytes.rs
```diff
@@ -0,0 +1,210 @@
...
+
+                                has_errors = true;
+                            }
+                            ast::LitKind::Byte(val) => {
+                                accumulator.push(val);
+                            }
+                            ast::LitKind::ByteStr(_) => {
+                                cx.struct_span_err(
+                                    expr.span,
+                                    "cannot concatenate doubly nested array",
```
Mark-Simulacrum, line 97: I would specifically say something about "byte strs are treated as arrays" here.

compiler/rustc_builtin_macros/src/concat_bytes.rs
```diff
@@ -0,0 +1,210 @@
...
+                ast::LitKind::Bool(val) => {
+                    cx.struct_span_err(e.span, "cannot concatenate boolean literals")
+                        .span_suggestion(
+                            e.span,
+                            "try using the numeric value",
+                            (val as u8).to_string(),
+                            Applicability::MachineApplicable,
+                        )
+                        .emit();
+                    has_errors = true;
```
Mark-Simulacrum, line 180: Ideally we'd find a way to deduplicate the invalid element types that are invalid everywhere (both inside and outside arrays). Maybe we can have a function that emits errors if needed for an element, calling itself on any nested arrays it finds?

compiler/rustc_builtin_macros/src/concat_bytes.rs
```diff
@@ -0,0 +1,210 @@
...
+            }
+        }
+    }
+    if !missing_literal.is_empty() {
+        let mut err = cx.struct_span_err(missing_literal, "expected a byte literal");
+        err.note("only byte literals (like `b\"foo\"`, `b's'`, and `[3, 4, 5]`) can be passed to `concat_bytes!()`");
+        err.emit();
+        return DummyResult::any(sp);
+    } else if has_errors {
+        return DummyResult::any(sp);
```
Mark-Simulacrum, line 206: Hm. I'm not sure if this is better, but maybe we should consider emitting `DummyResult::raw_expr(sp, true)` here? That way we're still indicating this will be an expression...

library/core/src/macros/mod.rs
```diff
@@ -949,6 +949,38 @@ pub(crate) mod builtin {
...
+    /// # fn main() {
+    /// let s: &[u8; 6] = concat_bytes!(b'A', b"BC", [68, b'E', 70]);
+    /// assert_eq!(s, b"ABCDEF");
+    /// # }
+    /// ```
+    #[cfg(not(bootstrap))]
+    #[unstable(
+        feature = "concat_bytes",
+        issue = "87555",
+        reason = "`concat_bytes` is not stable enough for use and is subject to change"
```
Mark-Simulacrum, line 976: Do we need a reason here? I think normally we just have a tracking issue which seems sufficient here.

compiler/rustc_builtin_macros/src/concat_bytes.rs
Smittyvb: I made it only emit one invalid literal error. A multispan with all of the invalid literals would still be pretty noisy due to the suggestions getting their own sections.

JohnCSimon:
Looks like this is ready for review

Mark-Simulacrum:
Squashed commits, looks good to me, thanks!

@bors r+

Issue #73494, opened 2020-06-19
Tracking issue for moving ui tests to subdirectories


pnkfelix:
EDIT(@jieyouxu): I personally find *just* moving the tests to be not *super* useful. Consider also *improving* these tests in the process -- see https://github.com/rust-lang/rust/issues/133895.

---

**NOTE: see https://github.com/rust-lang/rust/issues/73494#issuecomment-1500919846 for mentoring instructions.**

Looking at https://github.com/rust-lang/rust/tree/master/src/test/ui, github truncates the list to just the first 1000 entries (of 1564 total entries).

This is especially frustrating since github also sorts by name, not by kind, and so its truncation is cutting off whole subdirectories from the view.

TODO:
 - [ ] We should restructure the test suite so that no one directory has more than 1000 entries
 - [x] We should add code to tidy to ensure that future changes do not break the above invariant. (done by #79959)

lcnr:
The same problem exists for https://github.com/rust-lang/rust/tree/master/src/test/ui/issues with  1,839 omitted entries.

That could be solved by adding using a directory every 10000 issue ids, i.e. `ui/issues/10000-19999`, `ui/issues/20000-29999` and so on

workingjubilee:
Also see #76268: many of these are run pass tests that do not particularly need to be UI tests.

Havvy:
Is the compiler team okay with PRs that move tests into subdirectories?

Nadrieril:
> That could be solved by adding using a directory every 10000 issue ids, i.e. `ui/issues/10000-19999`, `ui/issues/20000-29999` and so on

That would technically work, but I'd much rather see some kind of thematic folders. I've been doing that for the bit of the compiler I'm hacking on, and it's been tremendously helpful to have most relevant tests be caught by `--test-args pattern`. For that same reason I like having the corresponding run-pass/check-pass tests in the same folders too. I don't find the ui/non-ui distinction all that helpful when I'm trying to check if my latest changes broke anything.

lcnr:
yeah, i agree that it would be better to group tests by what they are testing, but that's a lot of work and probably won't be done in the near future so imo it would be great to start by grouping them by issue id which is fairly easy and then continue sorting them afterwards

petrochenkov:
A simple machine learning model could assist this task well.
Putting a test into one of N subdirectories is a typical classification task.

My knowledge in this area is outdated by about 10 years, but here's one suggested solution
[...]

PR #87600, opened 2021-07-29
Move some UI tests to more suitable subdirs

JohnTitor:
The classifui result: https://gist.github.com/JohnTitor/c9e00840990b5e4a8fc562ec3571e427/e06c42226c6038da91e403c33b9947843420cf44

Some notes:
- backtrace-debuginfo.rs: previously I skipped this, I'm still not sure what the best dir is. Any ideas?
- estr-subtyping.rs: Seems a quite old test so removed, shouldn't?
- deref-suggestion.rs: moved to inference as `suggestions` is not an ideal dir.
- issue-43023.rs: a bit misclassified, moved to `derives`

cc #73494
r? @petrochenkov


petrochenkov:
Everything looks reasonable.
Maybe `deref-suggestion.rs` could be moved to `typeck` instead.
r=me after rebase.

petrochenkov:
`backtrace-debuginfo.rs` is a standard library test really, although it probably cannot be written as a unit test. I'm not sure where to put it either.

JohnTitor:
Thanks for the review! I'll leave `backtrace-debuginfo.rs` for now then.
@bors r=petrochenkov

PR #87601, opened 2021-07-29
Add functions to add unsigned and signed integers

a1phyr:
This PR adds methods to unsigned integers to add signed integers with good overflow semantics under `#![feature(mixed_integer_ops)]`.

The added API is:

```rust
// `uX` is `u8`, `u16`, `u32`, `u64`,`u128`, `usize`
impl uX {
    pub const fn checked_add_signed(self, iX) -> Option<Self>;
    pub const fn overflowing_add_signed(self, iX) -> (Self, bool);
    pub const fn saturating_add_signed(self, iX) -> Self;
    pub const fn wrapping_add_signed(self, iX) -> Self;
}


impl iX {
    pub const fn checked_add_unsigned(self, uX) -> Option<Self>;
    pub const fn overflowing_add_unsigned(self, uX) -> (Self, bool);
    pub const fn saturating_add_unsigned(self, uX) -> Self;
    pub const fn wrapping_add_unsigned(self, uX) -> Self;

    pub const fn checked_sub_unsigned(self, uX) -> Option<Self>;
    pub const fn overflowing_sub_unsigned(self, uX) -> (Self, bool);
    pub const fn saturating_sub_unsigned(self, uX) -> Self;
    pub const fn wrapping_sub_unsigned(self, uX) -> Self;
}
```

Maybe it would be interesting to also have `add_signed` that panics in debug and wraps in release ?


leonardo-m:
What are two (different from each other) use cases?

joshtriplett:
I've wanted both the checked and the wrapping versions of this many times.

library/core/src/num/uint_macros.rs
```diff
@@ -1378,6 +1449,35 @@ macro_rules! uint_impl {
...
+        #[rustc_const_unstable(feature = "uint_add_signed", issue = "none")]
+        #[must_use = "this returns the result of the operation, \
+                      without modifying the original"]
+        #[inline]
+        pub const fn overflowing_add_signed(self, rhs: $SignedT) -> (Self, bool) {
+            if rhs >= 0 {
+                self.overflowing_add(rhs as Self)
+            } else {
+                self.overflowing_sub(rhs.unsigned_abs())
+            }
```
kennytm, line 1478: this is equivalent to the following (at least it passes all test values from https://play.rust-lang.org/?version=nightly&mode=release&edition=2018&gist=641a3ff738671aef5bd6f6bec5721010):

```suggestion
            let (res, overflowed) = self.overflowing_add(rhs as Self);
            (res, overflowed ^ (rhs < 0))
```

and this implementation gets rid of much of the conditional stuff:

```asm
# this PR
playground::overflowing_add_signed:
	mov	rax, rdi
	xor	r8d, r8d
	add	rdi, rsi
	setb	r8b
	mov	rcx, rsi
	neg	rcx
	xor	edx, edx
	sub	rax, rcx
	setb	dl
	test	rsi, rsi
	cmovns	rax, rdi
	cmovns	edx, r8d
	ret

# this review comment
playground::overflowing_add_signed_2:
	mov	rax, rdi
	add	rax, rsi
	setb	cl
	test	rsi, rsi
	sets	dl
	xor	dl, cl
	ret
```

a1phyr: Thanks this is great !

On other archs it also gets rid of a branch. It also improves `checked_add_signed` which is implemented with it.

scottmcm:
Possibly-silly question: shouldn't there also be `checked_add_unsigned`?  Seems like `i32::MIN.checked_add_unsigned(u32::MAX)` ought to work too, if there's the other-way-around one...

a1phyr:
Yes, I *think* that it would be less useful but certainly more consistent. Also in this case we would need `*_sub_signed` too I think.

a1phyr:
Fixed conflicts, added tracking issue, added signed with unsigned operations and renamed feature.

a1phyr:
Rebased

library/core/src/num/uint_macros.rs
```diff
@@ -995,6 +1018,32 @@ macro_rules! uint_impl {
...
+        #[rustc_const_unstable(feature = "mixed_integer_ops", issue = "87840")]
+        #[must_use = "this returns the result of the operation, \
+                      without modifying the original"]
+        #[inline]
+        pub const fn saturating_add_signed(self, rhs: $SignedT) -> Self {
+            if rhs >= 0 {
+                self.saturating_add(rhs as Self)
+            } else {
+                self.saturating_sub(rhs.unsigned_abs())
+            }
```
kennytm, line 1044:
```suggestion
            let (res, overflowed) = self.overflowing_add(rhs as Self);
            if overflowed == (rhs < 0) {
                res
            } else if overflowed {
                0
            } else {
                Self::MAX
            }
```

<details><summary>This version generates branchless code on x86_64</summary>

```asm
playground::saturating_add_signed: # @playground::saturating_add_signed
# %bb.0:
	test	rsi, rsi
	js	.LBB0_1
# %bb.2:
	add	rdi, rsi
	mov	rax, -1
	cmovae	rax, rdi
	ret

.LBB0_1:
	neg	rsi
	xor	eax, eax
	sub	rdi, rsi
	cmovae	rax, rdi
	ret
                                        # -- End function

playground::saturating_add_signed_2: # @playground::saturating_add_signed_2
# %bb.0:
	xor	eax, eax
	add	rdi, rsi
	setb	al
	test	rsi, rsi
	setns	cl
	xor	cl, al
	neg	rax
	test	cl, cl
	cmovne	rax, rdi
	ret
                                        # -- End function
```

</details>

<details><summary>and it is also considerably faster (on x86_64; not sure about aarch64).</summary>

```rust
#![feature(test)]
extern crate test;

use test::bench::{Bencher, black_box};
use rand::{Rng, thread_rng};
use lazy_static::lazy_static;
use std::mem::transmute;

pub fn saturating_add_signed(a: u64, b: i64) -> u64 {
    if b >= 0 {
        a.saturating_add(b as _)
    } else {
        a.saturating_sub(b.unsigned_abs())
    }
}

pub fn saturating_add_signed_2(a: u64, b: i64) -> u64 {
    let (res, overflowed) = a.overflowing_add(b as _);
    if overflowed == (b < 0) {
        res
    } else if overflowed /* && b >= 0 */ {
        u64::MAX
    } else {
        0
    }
}

lazy_static! {
    static ref TEST_NUMBERS: [u128; 4096] = (|| {
        let mut v = [0; 4096];
        thread_rng().fill(&mut v[..]);
        v
    })();
}

#[bench]
fn bench_no_op(bencher: &mut Bencher) {
    bencher.iter(|| {
        for tv in &*TEST_NUMBERS {
            let (a, b): (u64, u64) = unsafe { transmute(*tv) };
            black_box(a);
            black_box(b);
        }
    });
}

#[bench]
fn bench_this_pr(bencher: &mut Bencher) {
    bencher.iter(|| {
        for tv in &*TEST_NUMBERS {
            let (a, b): (u64, i64) = unsafe { transmute(*tv) };
            black_box(saturating_add_signed(black_box(a), black_box(b)));
        }
    });
}

#[bench]
fn bench_suggestion(bencher: &mut Bencher) {
    bencher.iter(|| {
        for tv in &*TEST_NUMBERS {
            let (a, b): (u64, i64) = unsafe { transmute(*tv) };
            black_box(saturating_add_signed_2(black_box(a), black_box(b)));
        }
    });
}
```

</details>

```
test bench_no_op      ... bench:       2,504 ns/iter (+/- 68)
test bench_suggestion ... bench:       5,754 ns/iter (+/- 952)
test bench_this_pr    ... bench:      10,990 ns/iter (+/- 2,130)
```

library/core/src/num/int_macros.rs
```diff
@@ -822,6 +866,31 @@ macro_rules! int_impl {
...
+        #[rustc_const_unstable(feature = "mixed_integer_ops", issue = "87840")]
+        #[must_use = "this returns the result of the operation, \
+                      without modifying the original"]
+        #[inline]
+        pub const fn saturating_add_unsigned(self, rhs: $UnsignedT) -> Self {
+            // Overflow can only happen at the upper bound
+            match self.checked_add_unsigned(rhs) {
+                Some(x) => x,
+                None => Self::MAX,
+            }
```
kennytm, line 891:
```suggestion
            self.checked_add_unsigned(rhs).unwrap_or(Self::MAX)
```

library/core/src/num/int_macros.rs
```diff
@@ -843,6 +912,31 @@ macro_rules! int_impl {
...
+        #[rustc_const_unstable(feature = "mixed_integer_ops", issue = "87840")]
+        #[must_use = "this returns the result of the operation, \
+                      without modifying the original"]
+        #[inline]
+        pub const fn saturating_sub_unsigned(self, rhs: $UnsignedT) -> Self {
+            // Overflow can only happen at the lower bound
+            match self.checked_sub_unsigned(rhs) {
+                Some(x) => x,
+                None => Self::MIN,
+            }
```
kennytm, line 937:
```suggestion
            self.checked_sub_unsigned(rhs).unwrap_or(Self::MIN)
```

library/core/src/num/int_macros.rs
```diff
@@ -479,6 +501,28 @@ macro_rules! int_impl {
             unsafe { intrinsics::unchecked_sub(self, rhs) }
         }
 
+        /// Checked addition with an unsigned integer. Computes `self + rhs`,
```
kennytm, line 504:
```suggestion
        /// Checked subtraction with an unsigned integer. Computes `self - rhs`,
```

library/core/src/num/int_macros.rs
```diff
@@ -1018,6 +1133,27 @@ macro_rules! int_impl {
             intrinsics::wrapping_sub(self, rhs)
         }
 
+        /// Wrapping (modular) substraction with an unsigned integer. Computes
```
kennytm, line 1136:
```suggestion
        /// Wrapping (modular) subtraction with an unsigned integer. Computes
```

library/core/src/num/int_macros.rs
```diff
@@ -1419,6 +1582,33 @@ macro_rules! int_impl {
             (sum as $SelfT, borrow)
         }
 
+        /// Calculates `self` - `rhs` with an unsigned `rhs`
+        ///
+        /// Returns a tuple of the substraction along with a boolean indicating
```
kennytm, line 1587:
```suggestion
        /// Returns a tuple of the subtraction along with a boolean indicating
```

library/core/src/num/int_macros.rs
```diff
@@ -843,6 +912,31 @@ macro_rules! int_impl {
             intrinsics::saturating_sub(self, rhs)
         }
 
+        /// Saturating substraction with an unsigned integer. Computes `self - rhs`,
```
kennytm, line 915:
```suggestion
        /// Saturating subtraction with an unsigned integer. Computes `self - rhs`,
```

library/core/src/num/int_macros.rs
```diff
@@ -433,6 +433,28 @@ macro_rules! int_impl {
...
+        #[doc = concat!("assert_eq!((", stringify!($SelfT), "::MAX - 2).checked_add_unsigned(3), None);")]
+        /// ```
+        #[unstable(feature = "mixed_integer_ops", issue = "87840")]
+        #[rustc_const_unstable(feature = "mixed_integer_ops", issue = "87840")]
+        #[must_use = "this returns the result of the operation, \
+                      without modifying the original"]
+        #[inline]
+        pub const fn checked_add_unsigned(self, rhs: $UnsignedT) -> Option<Self> {
+            let (a, b) = self.overflowing_add_unsigned(rhs);
+            if unlikely!(b) {None} else {Some(a)}
```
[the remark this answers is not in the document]
a1phyr: Existing functions such as `checked_add` [already do this](https://github.com/rust-lang/rust/blob/08759c691e2e9799a3c6780ffdf910240ebd4a6b/library/core/src/num/uint_macros.rs#L417), so I copied the formatting for consistency.
Do you want me to change it through the whole file, correct only the new functions or keep everything like this ?

> strange that rustfmt doesn't catch it

The code is in a macro so I think `rustfmt` skips it.

library/core/src/num/uint_macros.rs
a1phyr: Nice optimization !

I think you swapped the last two branches, though.

library/core/src/num/int_macros.rs
a1phyr: `Option::unwrap_or` cannot be used in a `const` function yet

library/core/src/num/int_macros.rs
a1phyr: Same here

library/core/src/num/int_macros.rs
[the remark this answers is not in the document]
kennytm: @a1phyr ok just leave this alone then.

library/core/src/num/int_macros.rs
kennytm: oops.

a1phyr:
This looks like a spurious network error :/

Issue #87481, opened 2021-07-26
short backtraces do not always work on Windows without debug info


ehuss:
Ever since #75048, `RUST_BACKTRACE=1` does not show a backtrace in some situations when compiling with x86_64-pc-windows-msvc.  Some examples:

## Basic main

```rust
fn main() {
    panic!("test");
}
```

`RUST_BACKTRACE=1 cargo run --release`

On windows shows:

```
thread 'main' panicked at 'test', src\main.rs:2:5
stack backtrace:
note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace.
```

On other platforms, this shows:

```
thread 'main' panicked at 'test', src/main.rs:2:5
stack backtrace:
   0: std::panicking::begin_panic
   1: foo::main
note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace.
```

Hm, OK, maybe there's just something weird about `main` on Windows, let's panic from some function:

```rust
fn main() {
    abc();
}

#[inline(never)]
fn abc() {
    panic!("test");
}
```

Exact same results.  I wonder what `RUST_BACKTRACE=full` shows:

```
thread 'main' panicked at 'foo', src\main.rs:7:5
stack backtrace:
   0:     0x7ff732d4681e - std::backtrace_rs::backtrace::dbghelp::trace
                               at /rustc/53cb7b09b00cbea8754ffb78e7e3cb521cb8af4b\/library\std\src\..\..\backtrace\src\backtrace\dbghelp.rs:98
   1:     0x7ff732d4681e - std::backtrace_rs::backtrace::trace_unsynchronized
                               at /rustc/53cb7b09b00cbea8754ffb78e7e3cb521cb8af4b\/library\std\src\..\..\backtrace\src\backtrace\mod.rs:66
   2:     0x7ff732d4681e - std::sys_common::backtrace::_print_fmt
                               at /rustc/53cb7b09b00cbea8754ffb78e7e3cb521cb8af4b\/library\std\src\sys_common\backtrace.rs:67
   3:     0x7ff732d4681e - std::sys_common::backtrace::_print::{{impl}}::fmt
                               at /rustc/53cb7b09b00cbea8754ffb78e7e3cb521cb8af4b\/library\std\src\sys_common\backtrace.rs:46
   4:     0x7ff732d5390c - core::fmt::write
[...]

PR #75048, opened 2020-08-02
Prevent `__rust_begin_short_backtrace` frames from being tail-call optimised away

eggyal:
I've stumbled across some situations where there (unexpectedly) was no `__rust_begin_short_backtrace` frame on the stack during unwinding.

On closer examination, it appeared that the calls to that function had been tail-call optimised away.

This PR follows [@bjorn3's suggestion on Zulip](https://rust-lang.zulipchat.com/#narrow/stream/131828-t-compiler/topic/Disabling.20tail.20call.20optimisation.3F/near/205699133), by adding calls to `black_box` that hint to rustc not to perform TCO.

Fixes #47429

PR #87602, opened 2021-07-29
[backtraces]: look for the `begin` symbol only after seeing `end`

wesleywiser:
On `x86_64-pc-windows-msvc`, we often get backtraces which look like
    this:

    ```
    10:     0x7ff77e0e9be5 - std::panicking::rust_panic_with_hook
    11:     0x7ff77e0e11b4 - std::sys_common::backtrace::__rust_begin_short_backtrace::h5769736bdb11136c
    12:     0x7ff77e0e116f - std::sys_common::backtrace::__rust_end_short_backtrace::h61c7ecb1b55338ae
    13:     0x7ff77e0f89dd - std::panicking::begin_panic::h8e60ef9f82a41805
    14:     0x7ff77e0e108c - d
    15:     0x7ff77e0e1069 - c
    16:     0x7ff77e0e1059 - b
    17:     0x7ff77e0e1049 - a
    18:     0x7ff77e0e1039 - core::ptr::drop_in_place<std::rt::lang_start<()>::{{closure}}>::h1bfcd14d5e15ba81
    19:     0x7ff77e0e1186 - std::sys_common::backtrace::__rust_begin_short_backtrace::h5769736bdb11136c
    20:     0x7ff77e0e100c - std::rt::lang_start::{{closure}}::ha054184bbf9921e3
    ```

Notice that `__rust_begin_short_backtrace` appears on frame 11 before
    `__rust_end_short_backtrace` on frame 12. This is because in typical
    release binaries without debug symbols, dbghelp.dll, which we use to walk
    and symbolize the stack, does not know where CGU internal functions
    start or end and so the closure invoked by `__rust_end_short_backtrace`
    is incorrectly described as `__rust_begin_short_backtrace` because it
    happens to be near that symbol.

While that can obviously change, this has been happening quite
    consistently since #75048. Since this is a very small change to the std
    and the change makes sense by itself, I think this is worth doing.

This doesn't completely resolve the situation for release binaries on
    Windows, since without debug symbols, the stack printed can still show
    incorrect symbol names (this is why the test uses `#[no_mangle]`) but it
    does slightly improve the situation in that you see the same backtrace
    you would see with `RUST_BACKTRACE=full` or in a debugger (without the
    uninteresting bits at the top and bottom).

Fixes part of #87481

### DIFF
--- a/library/std/src/sys_common/backtrace.rs
+++ b/library/std/src/sys_common/backtrace.rs
@@ -75,7 +75,7 @@ unsafe fn _print_fmt(fmt: &mut fmt::Formatter<'_>, print_fmt: PrintFmt) -> fmt::
             hit = true;
             if print_fmt == PrintFmt::Short {
                 if let Some(sym) = symbol.name().and_then(|s| s.as_str()) {
-                    if sym.contains("__rust_begin_short_backtrace") {
+                    if start && sym.contains("__rust_begin_short_backtrace") {
                         stop = true;
                         return;
                     }
--- a/src/test/ui/panics/panic-short-backtrace-windows-x86_64.rs
+++ b/src/test/ui/panics/panic-short-backtrace-windows-x86_64.rs
@@ -10,6 +10,11 @@
 // check-run-results
 // exec-env:RUST_BACKTRACE=1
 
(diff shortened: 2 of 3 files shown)


dtolnay (approved):
Nice!

PR #87597, opened 2021-07-29
[Do Not Merge] Testing CI: Verify commits in beta are in stable.

yaymukund:
**Note**: This is not to be merged or reviewed.  

This PR exists to test the CI check I've written to address #74721

PR #87604, opened 2021-07-29
CI: Verify commits in beta & stable are in upstream branches.

### LINKED ISSUE #74721: Verify that backported commits are backports
The 1.45.1 release is primarily being made because of the accidental omission of a rustfmt backport from the 1.45 beta branch after it landed in 1.44.1.

The release team discussed this today, and we'd like to implement some automation to try and prevent this in the future at relatively low cost.

The plan is to only allow landing commits in beta / stable branches if `git cherry` (or similar, but that seems like the right thing) indicates they've already landed in the preceding branches (beta and master for stable, just master for beta). We will also want an explicit set of commits that are just allowed.

This will slightly increase the work needed during beta backports, as this list will need updating if there are any merge conflicts at least -- but since backports are rare and usually rollups, this shouldn't be much of a problem hopefully.

We would be interested in seeing someone implement this functionality!

yaymukund:
Closes #74721 

I think this does the trick. https://github.com/rust-lang/rust/pull/87597 is an example of it failing as it should.

### DIFF
--- a/src/ci/scripts/verify-backported-commits.sh
+++ b/src/ci/scripts/verify-backported-commits.sh
@@ -62,6 +62,11 @@ verify_cherries() {
     # Check each commit in <current>..<upstream>
     backport_sha=$(get_backport "$sha")
 
+    if [[ "$backport_sha" == "nothing" ]]; then
+      echo "✓ \`$sha\` backports nothing"
+      continue
+    fi
+
     if [[ -z "$backport_sha" ]]; then
       no_backports+=("$sha")
       continue
@@ -88,8 +93,8 @@ verify_cherries() {
         done
         echo
         echo "do not match any commits in \`$1\`. If this was intended, add the text"
-        echo '\`backport-of: <SHA of a commit already in master>\` somewhere in the'
-        echo 'message of each of these commits.'
+        echo '\`backport-of: <SHA of a commit already in master>\`'
+        echo 'somewhere in the message of each of these commits.'
         echo
         failure=1
   fi
@@ -112,12 +117,27 @@ verify_cherries() {
   return $failure
 }
 
-# Get the backport of a commit, or empty string if it does not exist.
+# Get the backport of a commit. It echoes one of:
+#
+# 1. A SHA of the backported commit
+# 2. The string "nothing"
+# 3. An empty string
 #
 # $1 = <sha>
 get_backport() {
+  # This regex is:
+  #
+  # ^.* - throw away any extra starting characters
+  # backport-of: - prefix
+  # \s\? - optional space
+  # \(\) - capture group
+  # [a-f0-9]\+\|nothing - a SHA or the text 'nothing'
+  # .* - throw away any extra ending characters
+  # \1 - replace it with the first match
+  # {s//\1/p;q} - print the first occurrence and quit
+  #
   git show -s --format=%B "$1" \
-    | sed -n 's/^.*backport-of:\s\?\([a-f0-9]\+\).*/\1/p'
+    | sed -n '/^.*backport-of:\s\?\([a-f0-9]\+\|nothing\).*/{s//\1/p;q}'
 }
 
 # Check if a commit is in master.
--- a/src/ci/scripts/verify-backported-commits.sh
+++ b/src/ci/scripts/verify-backported-commits.sh
@@ -1,0 +1,130 @@
+#!/bin/bash
+# Ensure commits in beta are in master & commits in stable are in beta + master.
+set -euo pipefail
+IFS=$'\n\t'
+
+source "$(cd "$(dirname "$0")" && pwd)/../shared.sh"
+
+# We don't care about commits that predate this automation check, so we pass a
+# `<limit>` argument to `git cherry`.
+BETA_LIMIT="53fd98ca776cb875bc9e5514f56b52eb74f9e7a9"
+STABLE_LIMIT="a178d0322ce20e33eac124758e837cbd80a6f633"
+
+verify_backported_commits_main() {
+  ci_base_branch=$(ciBaseBranch)
+
+  if [[ "$ci_base_branch" != "beta" && "$ci_base_branch" != "stable" ]]; then
+    echo 'Skipping. This is only run when merging to the beta or stable branches.'
+    exit 0
+  fi
+
+  echo 'git: unshallowing the repository so we can check commits'
+  git fetch \
+    --no-tags \
+    --no-recurse-submodules \
+    --progress \
+    --prune \
+    --unshallow
+
+  if [[ $ci_base_branch == "beta" ]]; then
+    verify_cherries master "$BETA_LIMIT" \
+      || exit 1
+
+  elif [[ $ci_base_branch == "stable" ]]; then
+    (verify_cherries master "$STABLE_LIMIT" \
+      & verify_cherries beta "$STABLE_LIMIT") \
+      || exit 1
+
+  fi
+}
+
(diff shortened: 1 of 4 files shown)


pietroalbini:
I'll soon go on vacation and I won't have time to review PRs then. r? @Mark-Simulacrum?

src/ci/scripts/verify-backported-commits.sh
```diff
@@ -0,0 +1,32 @@
...
+source "$(cd "$(dirname "$0")" && pwd)/../shared.sh"
+
+cherries_dir="$(dirname "$0")/cherries"
+
+verify_cherry_file() {
+  base_branch=$1
+  current_branch=$2
+  cherry_file="$cherries_dir/$current_branch-$base_branch"
+
+  git cherry "origin/$base_branch" | sort -s -t' ' -k2 > $cherry_file
```
jyn514, line 15:
```suggestion
  git cherry "origin/$base_branch" | sort -s -t' ' -k2 > "$cherry_file"
```

src/ci/scripts/verify-backported-commits.sh
```diff
@@ -0,0 +1,32 @@
...
+
+  if ! git diff --quiet $cherry_file; then
```
jyn514, line 17:
```suggestion
  if ! git diff --quiet "$cherry_file"; then
```

src/ci/scripts/verify-backported-commits.sh
```diff
@@ -0,0 +1,32 @@
...
+    echo "Found differing commits between $base_branch and $current_branch."
+    git diff $cherry_file
```
jyn514, line 19:
```suggestion
    git diff "$cherry_file"
```

src/ci/scripts/verify-backported-commits.sh
```diff
@@ -0,0 +1,32 @@
...
+    exit 1
+  fi
+}
+
+ci_base_branch=$(ciBaseBranch)
+if [[ $ci_base_branch == "beta" ]]; then
+  verify_cherry_file "master" "beta"
+elif [[ $ci_base_branch == "stable" ]]; then
+  verify_cherry_file "master" "stable"
+  verify_cherry_file "beta" "stable"
```
jyn514, line 29: All these quotes are unnecessary (if you're going to quote anything, it should be the variable).
```suggestion
if [[ $ci_base_branch == beta ]]; then
  verify_cherry_file master beta
elif [[ $ci_base_branch == stable ]]; then
  verify_cherry_file master stable
  verify_cherry_file beta stable
```

yaymukund: I've included your suggestions - thanks!

Mark-Simulacrum:
Could you add some documentation indicating what the outputs are / limitations here etc? For example, it looks like there are a number of commits different between the branches, but it's not quite clear to me what the expectations are for when those need to be updated and such.

We'll also want to be careful to avoid a situation where editing these files to include a commit is "impossible" because that would change the commit hash. At a quick glance, it seems like this could be a problem. Maybe instead of a single file tracking this information, we would prefer to ensure that each commit on the beta and stable branches have a line in them indicating the corresponding hash on the master branch, e.g., 

> backport-of: <hash>

yaymukund:
@Mark-Simulacrum That all makes sense.

I've implemented your `backport-of: <sha>` suggestion and added more details in the script & in the output [(see a sample failure)](https://github.com/rust-lang/rust/pull/87597/checks?check_run_id=3275943974#step:25:487).

src/ci/scripts/verify-backported-commits.sh
```diff
@@ -0,0 +1,130 @@
...
+
+    if [[ -z "$backport_sha" ]]; then
+      no_backports+=("$sha")
+      continue
+    fi
+
+    if ! is_in_master "$backport_sha"; then
+      bad_backports+=("$sha")
+      continue
+    fi
```
yaymukund, line 73: I wasn't sure how to validate these SHAs. Is this right?

Mark-Simulacrum: This seems like a good idea.

Mark-Simulacrum:
> This works fine if a single commit requires conflict resolution. However, if you merge an entire branch into (e.g.) beta but not master then each commit in the branch will need a backport-of: <sha> commit message. Is that acceptable?

Yeah, this is expected. I usually cherry-pick commits one by one anyway, so it shouldn't really be a big deal.

Can we add a script like src/tools/cherry-pick.sh which is essentially a wrapper around git cherry-pick but additionally edits the commit message after picking to include the backport-of line? I'd suggest using the git cherry-pick -x option, but it looks like that doesn't include the upstream commit hash if the commit has conflicts when cherry picking, so it won't quite work for us...

src/tools/cherry-pick.sh
```diff
@@ -0,0 +1,28 @@
...
+commit_message_with_backport_note() {
+  message=$(git log --format=%B -n 1 $1)
+  echo $message | awk "NR==1{print; print \"\n(backport-of: $1)\"} NR!=1"
+}
+
+sha=$(full_sha $1)
+git cherry-pick $sha
+git commit \
+  --amend \
+  --file <(commit_message_with_backport_note $sha)
```
Mark-Simulacrum, line 28: Could you refactor this to take multiple parameters (and just cherry-pick each in turn, just like the normal cherry pick)?

yaymukund:
**PLEASE NOTE: DO NOT MERGE**: This is producing strange behavior in CI and should not be merged in its current state. I shall update when it's fixed.

yaymukund:
@Mark-Simulacrum If I understand your comment correctly, I don't think this script works if your cherry-pick has a conflict. You get dumped into the conflict resolution phase and the subsequent `git commit --amend` doesn't ever happen. I can try to make this atomic if you wish - I think there's an alternate approach where I set `GIT_EDITOR=/some/script git cherry-pick -e`.

This gets worse if you have a conflict in the middle of a long series of commits, as the first n commits may be successfully applied, but n+1 breaks. I don't see how to fix this easily.

Mark-Simulacrum:
Hm. Yes, I see your point. I think it might be possible that we can make it work using `--continue` as an argument to our wrapper script, but that might be hard -- I guess the goal would be to store some state that we could use to sort of resume the script. It seems like git-cherry-pick itself does this via .git/sequencer based on the manpage:

> Continue the operation in progress using the information in .git/sequencer

But I'm not sure whether that file is documented anywhere or if editing it is in anyway feasible.

It's probably not a huge deal for us to manually deal with conflicts and amend commits after the fact; CI *will* yell at us after all...

Mark-Simulacrum:
I think if we can get it to work via GIT_EDITOR somehow, then let's try that, but if not, then no big deal.

yaymukund:
Let's stick with the present approach.

I've just tested it and `GIT_EDITOR` doesn't seem to persist through conflict resolution. We'd need to `export GIT_EDITOR`, which opens up another can of worms.

Mark-Simulacrum:
Oh, one more thought: we will have commits on beta/stable that *are* specific to those branches, such as https://github.com/rust-lang/rust/commit/56397e90e178fdc98b48031ceb79e90c61a48e3a -- those will want their own annotation like "backport-of: nothing" or something, still explicitly denoted.

Could you add that support?

I think that's the last thing.

Mark-Simulacrum:
@bors r+ rollup

Thanks! Looking forward to seeing this work in practice so we can see how effective/painful it is.

yaymukund:
Thanks for reviewing. Feel free to tag me in followup work if it would be helpful.

PR #87607, opened 2021-07-29
Add a hint that the expressions produce a value

### LINKED ISSUE #85913: Unecessary Warning message
I tried this code:

```rs
#![allow(dead_code)]

fn fun() -> i32 {
    function() && return 1;

    panic!("stop");
}

fn function() -> bool {
    true
}

fn main() {}

```

where `function` is any function that returns a `boolean` , and it gives  the warning
*unused logical operator* , but actually works as expected, so i think this warning should not come up.

JohnTitor:
Fixes #85913
The second commit is semi-_unrelated_ but it allows us to run the related tests just on `src/test/ui/lint`.


JohnTitor:
Well, well, well: https://www.githubstatus.com/incidents/8qy34dmg4kdm

compiler/rustc_lint/src/unused.rs
```diff
@@ -161,7 +161,18 @@ impl<'tcx> LateLintPass<'tcx> for UnusedResults {
...
+                lint.note(&format!("the {} produces a value", must_use_op));
+                if let Ok(snippet) = cx.sess().source_map().span_to_snippet(expr.span) {
+                    lint.span_suggestion(
+                        expr.span,
+                        "use `let _ = ...` to ignore it",
+                        format!("let _ = {}", snippet),
+                        Applicability::MachineApplicable,
+                    )
+                    .emit()
+                }
```
estebank, line 174:
```suggestion
                lint.span_label(expr.span, &format!("the {} produces a value", must_use_op));
                lint.span_suggestion_verbose(
                    expr.span,
                    "use `let _ = ...` to ignore the resulting value",
                    "let _ = ".to_string(),
                    Applicability::MachineApplicable,
                );
```

estebank (approved):
r=me after addressing the nitpick

compiler/rustc_lint/src/unused.rs
JohnTitor: Thanks! Applied it with changing `expr.span` to `expr.span.shrink_to_lo()`.

Issue #75243, opened 2020-08-07
Tracking Issue for `array_map`


JulianKnodt:
The feature gate for the issue is `#![feature(array_map)]`.

### Public API

```rust
impl<T, const N: usize> [T; N] {
    pub fn map<F, U>(self, f: F) -> [U; N] where F: FnMut(T) -> U;
}
```

### Steps / History

- [x] Implement it:  #75212
- [ ] Add in place optimization for same size & alignment (?)
- [x] FCP
- [ ] Stabilization PR #87174

### Unresolved Questions

How should order be documented/allowed for map? See https://github.com/rust-lang/rust/pull/75212#discussion_r468234354 for discussion.

the8472:
A possible followup optimization: Perform this in place when `size_of::<T>() == size_of::<U>() && align_of::<T>() == align_of<U>()`, which can be checked at compile time. Together with RVO that might save a lot of extra stack space on large arrays.

vadixidav:
This feature is currently breaking [`arraymap::ArrayMap`](https://docs.rs/arraymap/0.1.1/arraymap/trait.ArrayMap.html) without enabling the feature:

```rust
error[E0658]: use of unstable library feature 'array_map'
  --> lambda-twist/tests/consensus.rs:25:6
   |
25 |     .map(|&p| Point3::from(p));
   |      ^^^
   |
   = note: see issue #75243 <https://github.com/rust-lang/rust/issues/75243> for more information
   = help: add `#![feature(array_map)]` to the crate attributes to enable
```

This code started breaking on more recent Rust versions as it seems that it thinks I am attempting to use the map method on array created by this feature, but actually I am using [`arraymap::ArrayMap::map`](https://docs.rs/arraymap/0.1.1/arraymap/trait.ArrayMap.html#tymethod.map).

JulianKnodt:
Ah I didn't know this crate existed, thanks for the heads up. I'm not quite sure how to resolve this, but will investigate.

xfix:
That seems like a compiler bug. I think a warning like this should show, but due to `map` accepting `&self` rather than `self`, it's not being seen. The problem can be seen with the following program:

```rust
trait ArrayExt {
    fn map<F: FnMut(i32) -> i32>(&self, f: F);
}

impl ArrayExt for [i32; 1] {
    fn map<F: FnMut(i32) -> i32>(&self, f: F) {}
}

fn main() {
    [4].map(|x| x);
}
```

In my opinion, this should print the following:

[...]

PR #87609, opened 2021-07-29
Add docs about performance and `Iterator::map` to `[T; N]::map`

LukasKalbertodt:
This suboptimal code gen for some usages of array::map got a bit of
attention by multiple people throughout the community. Some cases:

- https://github.com/rust-lang/rust/issues/75243#issuecomment-866051086
- https://github.com/rust-lang/rust/issues/75243#issuecomment-874732134
- https://www.reddit.com/r/rust/comments/oeqqf7/unexpected_high_stack_usage/

My *guess* is that this gets the attention it gets because in JavaScript
(and potentially other languages), a `map` function on arrays is very
commonly used since in those languages, arrays basically take the role
of Rust's iterator. I considered explicitly naming JavaScript in the
first paragraph I added, but I couldn't find precedence of mentioning
other languages in standard library doc, so I didn't add it.

When array::map was stabilized, we still wanted to add docs, but that 
somehow did not happen in time. So here we are. Not sure if this sounds
crazy but maybe it is worth considering beta backporting this? Only if
it's not a lot of work, of course! But yeah, stabilized array::map is
already in beta and if this problem is really as big as it sometimes seems,
might be worth having the docs in place when 1.55 is released.

CC @CryZe 

r? @m-ou-se (since you were involved in that discussion and the stabilization)

### DIFF
--- a/library/core/src/array/mod.rs
+++ b/library/core/src/array/mod.rs
@@ -293,6 +293,28 @@ impl<T, const N: usize> [T; N] {
     /// Returns an array of the same size as `self`, with function `f` applied to each element
     /// in order.
     ///
+    /// If you don't necessarily need a new fixed-size array, consider using
+    /// [`Iterator::map`] instead.
+    ///
+    ///
+    /// # Note on performance and stack usage
+    ///
+    /// Unfortunately, usages of this method are currently not always optimized
+    /// as well as they could be. This mainly concerns large arrays, as mapping
+    /// over small arrays seem to be optimized just fine. Also note that in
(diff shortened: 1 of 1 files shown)


leonardo-m:
It could also be added a warning when the programmer attempts to use map on a  sufficiently large array.

LukasKalbertodt:
> It could also be added a warning when the programmer attempts to use map on a sufficiently large array.

That would require us to set some kind of threshold, which is hard. And there are cases where `map` optimizes well even for large arrays, and others where it doesn't even for medium sized ones. So I'm not sure such a warning would really help. If anything, it could be a clippy lint.

PR #87614, opened 2021-07-30
Recommend fix `count()` -> `len()` on slices

### LINKED ISSUE #87302: Suggest `len` instead of `count` on arrays, slices, or vectors
Given the following code: 

```rust
fn main() {
    let slice = [1,2,3,4];
    let vec = vec![1,2,3,4];
    
    println!("{} {} {}", slice.count(), vec.count(), vec.as_slice().count());
}
```
https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=8bd1115b12adc0eb5feb16bbb63a06cb

The current output is:

```
error[E0599]: the method `count` exists for array `[{integer}; 4]`, but its trait bounds were not satisfied
 --> src/main.rs:5:32
  |
5 |     println!("{} {} {}", slice.count(), vec.count(), vec.as_slice().count());
  |                                ^^^^^ method cannot be called on `[{integer}; 4]` due to unsatisfied trait bounds
  |
  = note: the following trait bounds were not satisfied:
          `[{integer}; 4]: Iterator`
          which is required by `&mut [{integer}; 4]: Iterator`
          `[{integer}]: Iterator`
          which is required by `&mut [{integer}]: Iterator`
```
(repeated 3 times with slight variations)

Ideally the output should:
* suggest using `slice.len()` instead of `slice.count()`

This is a common error, particularly for new users, and when switching between iterators and vectors/slices/arrays.

notriddle:
Fixes #87302

### DIFF
--- a/compiler/rustc_typeck/src/check/method/suggest.rs
+++ b/compiler/rustc_typeck/src/check/method/suggest.rs
@@ -67,6 +67,10 @@ impl<'a, 'tcx> FnCtxt<'a, 'tcx> {
         }
     }
 
+    fn is_slice_ty(&self, ty: Ty<'tcx>, span: Span) -> bool {
+        self.autoderef(span, ty).any(|(ty, _)| matches!(ty.kind(), ty::Slice(..) | ty::Array(..)))
+    }
+
     pub fn report_method_error(
         &self,
         mut span: Span,
@@ -691,7 +695,23 @@ impl<'a, 'tcx> FnCtxt<'a, 'tcx> {
 
                 let mut restrict_type_params = false;
                 let mut unsatisfied_bounds = false;
-                if !unsatisfied_predicates.is_empty() {
+                if item_name.name == sym::count && self.is_slice_ty(actual, span) {
+                    let msg = "consider using `len` instead";
+                    if let SelfSource::MethodCall(_expr) = source {
+                        err.span_suggestion_short(
+                            span,
+                            msg,
+                            String::from("len"),
+                            Applicability::MachineApplicable,
+                        );
+                    } else {
+                        err.span_label(span, msg);
+                    }
+                    if let Some(iterator_trait) = self.tcx.get_diagnostic_item(sym::Iterator) {
+                        let iterator_trait = self.tcx.def_path_str(iterator_trait);
+                        err.note(&format!("`count` is defined on `{iterator_trait}`, which `{actual}` does not implement"));
+                    }
+                } else if !unsatisfied_predicates.is_empty() {
                     let def_span = |def_id| {
                         self.tcx.sess.source_map().guess_head_span(self.tcx.def_span(def_id))
                     };
@@ -990,9 +1010,9 @@ impl<'a, 'tcx> FnCtxt<'a, 'tcx> {
                     }
                 }
 
-                let mut fallback_span = true;
-                let msg = "remove this method call";
                 if item_name.name == sym::as_str && actual.peel_refs().is_str() {
+                    let msg = "remove this method call";
+                    let mut fallback_span = true;
                     if let SelfSource::MethodCall(expr) = source {
                         let call_expr =
                             self.tcx.hir().expect_expr(self.tcx.hir().get_parent_node(expr.hir_id));
--- a/compiler/rustc_span/src/symbol.rs
+++ b/compiler/rustc_span/src/symbol.rs
@@ -496,6 +496,7 @@ symbols! {
         core_panic_macro,
         cosf32,
         cosf64,
+        count,
         cr,
         crate_id,
         crate_in_paths,
--- a/src/test/ui/suggestions/count2len.rs
+++ b/src/test/ui/suggestions/count2len.rs
@@ -1,0 +1,8 @@
+fn main() {
+    let slice = [1,2,3,4];
+    let vec = vec![1,2,3,4];
+
+    slice.count(); //~ERROR: E0599
+    vec.count(); //~ERROR: E0599
+    vec.as_slice().count(); //~ERROR: E0599
+}
--- a/src/test/ui/suggestions/count2len.stderr
+++ b/src/test/ui/suggestions/count2len.stderr
@@ -1,0 +1,36 @@
+error[E0599]: no method named `count` found for array `[{integer}; 4]` in the current scope
+  --> $DIR/count2len.rs:5:11
+   |
+LL |     slice.count();
+   |           ^^^^^
+   |           |
+   |           method cannot be called on `[{integer}; 4]` due to unsatisfied trait bounds
+   |           help: consider using `len` instead
+   |
+   = note: `count` is defined on `Iterator`, which `[{integer}; 4]` does not implement
+
+error[E0599]: no method named `count` found for struct `Vec<{integer}>` in the current scope
+  --> $DIR/count2len.rs:6:9
+   |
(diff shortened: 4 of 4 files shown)


jackh726:
My first thought is (without reading too much of the code): what other methods do we have suggestions like this for? Also, even though this is technically compiler code, it seems more like a @rust-lang/libs decision; it reminds me somewhat of the rustdoc aliases: at what point is it "okay" and when is it "too much"

notriddle:
> My first thought is (without reading too much of the code): what other methods do we have suggestions like this for?

The same file I modified also has a suggestion for `as_str(self: &str)`, which is another method that doesn't exist but developers are likely to try to call (and has exactly one, obvious, correction to apply).

cuviper:
Could this use `#[rustc_on_unimplemented(... from_method = "count" ...)]` on `Iterator`?

notriddle:
Here's what I tried, to achieve the effect using that system. It didn't work.

```diff
diff --git a/library/core/src/iter/traits/iterator.rs b/library/core/src/iter/traits/iterator.rs
index 6d83b32ae4e..bdee0e202c6 100644
--- a/library/core/src/iter/traits/iterator.rs
+++ b/library/core/src/iter/traits/iterator.rs
@@ -26,8 +26,10 @@ fn _assert_is_object_safe(_: &dyn Iterator<Item = ()>) {}
 #[stable(feature = "rust1", since = "1.0.0")]
 #[rustc_on_unimplemented(
     on(
-        any(_Self = "[]", _Self = "std::vec::Vec", _Self = "std::string::String", _Self = "str"),
-        on_method = "count",
+        all(
+            any(_Self = "[]", _Self = "std::vec::Vec", _Self = "std::string::String", _Self = "str"),
+            from_method = "count",
+        ),
         label = "consider using `.len()`"
     ),
     on(
```

Here's the message that it gives:

```
error[E0599]: the method `count` exists for array `[{integer}; 4]`, but its trait bounds were not satisfied
  --> $DIR/count2len.rs:5:11
   |
LL |     slice.count();
   |           ^^^^^ method cannot be called on `[{integer}; 4]` due to unsatisfied trait bounds
   |
   = note: the following trait bounds were not satisfied:
           `[{integer}; 4]: Iterator`
           which is required by `&mut [{integer}; 4]: Iterator`
           `[{integer}]: Iterator`
           which is required by `&mut [{integer}]: Iterator`
```

I notice that my changes have no effect on it, but also, that the `{Self} isn't an iterator` language isn't being used anywhere in here, either. Maybe because there are multiple applicable trait implementations?

cuviper:
My hunch is that the `&mut I` implementation is in the way. Maybe you need `_Self = "&[]"`?

notriddle:
No. `_Self = "&[]"` doesn't do anything different.

compiler/rustc_typeck/src/check/method/suggest.rs
```diff
@@ -673,7 +678,26 @@ impl<'a, 'tcx> FnCtxt<'a, 'tcx> {
...
                 let mut unsatisfied_bounds = false;
-                if !unsatisfied_predicates.is_empty() {
+                let mut fallback_span = true;
+                if item_name.name == sym::count && self.is_slice_ty(actual, span) {
+                    let msg = "consider using `len` instead";
+                    if let SelfSource::MethodCall(expr) = source {
+                        let call_expr =
+                            self.tcx.hir().expect_expr(self.tcx.hir().get_parent_node(expr.hir_id));
+                        if let Some(span) = call_expr.span.trim_start(expr.span) {
+                            err.span_suggestion(
```
camelid, line 688:
```suggestion
                            err.span_suggestion_short(
```

This should change the output like so:

```diff
-help: consider using `len` instead: `.len()`
+help: consider using `len` instead
```

but the suggested code will still be displayed if it's in its own diagnostic window.

jackh726:
I'm going to defer to libs (or libs-api?) here.

r? @cuviper feel free to reassign

camelid:
@notriddle Please rebase instead of merging in upstream changes because of the ["No-Merge Policy"](https://rustc-dev-guide.rust-lang.org/git.html#no-merge-policy). Thanks!

Mark-Simulacrum:
This feels like a fairly reasonable change to me (obviously, we can iterate further on the message in the future). I'm not entirely convinced at whether this is a particularly common case, but the extra work is not huge, and so it seems like an OK change to make in my opinion.

My cursory look at the compiler code suggests we don't consider on_unimplemented for this kind of error, so we might want to implement that in the future -- but it seems unnecessary to block this improvement on that larger goal.

I also agree with others on the "is this too much" angle, but ultimately, this is something relatively easy to back out and we're not yet seeing a huge influx of such targeted messages that I'm worried. :)

----

That said, I think the proposed error message in this PR is pretty suboptimal -- we're hiding the trait the user *might've* been intending to reference, not even mentioning its name:

```
error[E0599]: no method named `count` found for array `[{integer}; 4]` in the current scope
  --> $DIR/count2len.rs:5:11
   |
LL |     slice.count();
   |          -^^^^^--
   |          ||
   |          |method cannot be called on `[{integer}; 4]` due to unsatisfied trait bounds
   |          help: consider using `len` instead
```

I think it would be good to rework this to something like:

```
error[E0599]: no method named `count` found for array `[{integer}; 4]` in the current scope
  --> $DIR/count2len.rs:5:11
   |
LL |     slice.count();
   |          --------
   |          |
   |          help: consider using `len` instead
   |
   note: `count` is defined on `std::iter::Iterator`, which `[{integer}; 4]` does not implement
```

Or potentially to hide the "unsatisfied due to trait bounds" error altogether, though not sure on that front.

cc @rust-lang/wg-diagnostics -- wondering if you have opinions on this particular case or if we should instead invest in expanding rustc_on_unimplemented to cover cases like this and avoid ad-hoc code in the compiler.

notriddle:
I agree. Rebased and pushed a new error message, that's closer to what you wrote.

```text
error[E0599]: no method named `count` found for array `[{integer}; 4]` in the current scope
  --> $DIR/count2len.rs:5:11
   |
LL |     slice.count();
   |           ^^^^^
   |           |
   |           method cannot be called on `[{integer}; 4]` due to unsatisfied trait bounds
   |           help: consider using `len` instead
   |
   = note: `count` is defined on `Iterator`, which `[{integer}; 4]` does not implement
```

compiler/rustc_typeck/src/check/method/suggest.rs
```diff
@@ -67,6 +67,11 @@ impl<'a, 'tcx> FnCtxt<'a, 'tcx> {
         }
     }
 
+    fn is_slice_ty(&self, ty: Ty<'tcx>, span: Span) -> bool {
+        self.autoderef(span, ty)
+            .any(|(ty, _)| self.probe(|_| matches!(ty.kind(), ty::Slice(..) | ty::Array(..))))
```
Mark-Simulacrum, line 72: I'm pretty sure the probe(...) here is not actually meaningful, right? The body is not creating any obligations, it's just checking an enum discriminant basically?

Mark-Simulacrum (commented):
r=me with nit fixed

compiler/rustc_typeck/src/check/method/suggest.rs
notriddle: You're right. That should be fixed.

PR #87073, opened 2021-07-12
Fix rustdoc handling of primitive items

jyn514:
This is a complicated PR and does a lot of things. I'm willing to split it up a little more if it would help reviewing, but it would be tricky and I'd rather not unless it's necessary.

 ## What does this do?

- Fixes https://github.com/rust-lang/rust/issues/73423.
- Fixes https://github.com/rust-lang/rust/issues/79630. I'm not sure how to test this for the standard library explicitly, but you can see from some of the diffs from the `no_std` tests. I also tested it locally and it works correctly: image
- Fixes https://github.com/rust-lang/rust/issues/83083.

## Why are these changes interconnected?

- Allowing anchors (https://github.com/rust-lang/rust/issues/83083) without fixing the online/offline problem (https://github.com/rust-lang/rust/issues/79630) will actually just silently discard the anchors, that's not a fix. The online/offline problem is directly related to the fragment hack; links need to go through `fn href()` to be fixed.
- Technically I could fix the online/offline problem without removing the error on anchors; I am willing to separate that out if it would be helpful for reviewing. However I can't fix the anchor problem without adding docs to core, since rustdoc needs all those primitives to have docs to avoid a fallback, and currently `#![no_std]` crates don't have docs for primitives. I also can't fix the online/offline problem without removing the fragment hack, since otherwise diffs like this will be wrong for some primitives but not others:
```diff
@@ -385,7 +381,7 @@ fn resolve_primitive_associated_item(
                         ty::AssocKind::Const => "associatedconstant",
                         ty::AssocKind::Type => "associatedtype",
                     };
-                    let fragment = format!("{}#{}.{}", prim_ty.as_sym(), out, item_name);
+                    let fragment = format!("{}.{}", out, item_name);
                     (Res::Primitive(prim_ty), fragment, Some((kind.as_def_kind(), item.def_id)))
                 })
         })
```
- Adding primitive docs to core without making any other change will cause links to go to `core` instead of `std`, even for crates with `extern crate std`. See "Breaking changes to doc(primitive)" below for why this is the case. That said, I could add some special casing to rustdoc at the same time that would let me separate this change from the others (it would fix https://github.com/rust-lang/rust/issues/73423 but still special-case intra-doc links). I'm willing to separate that out if helpful for reviewing.

### Add primitive documentation to libcore

[...]

PR #87618, opened 2021-07-30
Add missing documentation for std::char types

GuillaumeGomez:


library/core/src/char/convert.rs
```diff
@@ -193,6 +193,9 @@ impl From<u8> for char {
 }
 
 /// An error which can be returned when parsing a char.
+///
+/// This `struct` is created when using the
+/// [`char::from_str`](../../std/primitive.char.html#method.from_str) method.
```
jyn514, line 198: Please use intra-doc links so this doesn't break on re-exports.
```suggestion
/// [`char::from_str`] method.
```

library/core/src/char/convert.rs
```diff
@@ -257,7 +260,11 @@ impl TryFrom<u32> for char {
     }
 }
 
-/// The error type returned when a conversion from u32 to char fails.
+/// The error type returned when a conversion from [`prim@u32`] to [`prim@char`] fails.
+///
+/// This `struct` is created by the
+/// [`char::try_from<u32>`](../../std/primitive.char.html#impl-TryFrom<u32>) method. See its
```
jyn514, line 266:
```suggestion
/// [`char::try_from<u32>`](char#impl-TryFrom<u32>) method. See its
```

library/core/src/char/convert.rs
GuillaumeGomez: I don't even know why I didn't use it directly...

GuillaumeGomez:
I forgot that we don't support for primitive types yet... Needs to wait for https://github.com/rust-lang/rust/pull/87073 to be merged.

GuillaumeGomez:
@jyn514: ^ wasn't this restriction lifted?

jyn514:
@GuillaumeGomez it needs to wait for the bootstrap compiler to be bumped, so another 6 weeks. To avoid waiting longer (e.g. if it turns out rustdoc is buggy), you could run `x.py doc --stage 1 library/std` to make sure it will work at that time.

GuillaumeGomez:
I did and it worked, which is why I was confused. Well, let's wait for 6 weeks then!

jyn514:
@GuillaumeGomez it's been 6 weeks :)

GuillaumeGomez:
Let's see! :D

library/core/src/char/convert.rs
```diff
@@ -199,6 +199,8 @@ impl const From<u8> for char {
 }
 
 /// An error which can be returned when parsing a char.
+///
+/// This `struct` is created when using the [`char::from_str`](char#tymethod.from_str) method.
```
GuillaumeGomez, line 203: I think it's still not ready:

```
warning: unresolved link to `char::from_str`
 --> foo.rs:1:7
  |
1 | /// [`char::from_str`]
  |       ^^^^^^^^^^^^^^ the builtin type `char` has no associated item named `from_str`
```

We need https://github.com/rust-lang/rust/issues/90703 to be fixed first.

GuillaumeGomez, line 203: I think it's still not ready:

```
warning: unresolved link to `char::from_str`
 --> foo.rs:1:7
  |
1 | /// [`char::from_str`]
  |       ^^^^^^^^^^^^^^ the builtin type `char` has no associated item named `from_str`
```

We need https://github.com/rust-lang/rust/issues/90703 to be fixed first.

jyn514: Wait, this doesn't match the error above though. The CI failure resolved the link correctly, it's just a dead link for some reason. What version of rustdoc did you use to test locally? Does the error change if you use beta?

jyn514: Oh, it's because you used a fragment link instead of an intra-doc link.

GuillaumeGomez: Yes, which is invalid too. I'll update the current code but even the fixed version is broken for the moment.

camelid:
> We need #90703 to be fixed first.

#90703 is fixed now. I think this PR is no longer blocked?

GuillaumeGomez:
Let's give it another try!

GuillaumeGomez:
Seems like the CI passed this time. \o/

GuillaumeGomez:
Forgot to r+ it...

@bors: r=jyn514,camelid

Issue #76923, opened 2020-09-19
Tracking Issue for feature(unix_socket_peek)


rijenkii:
This is a tracking issue for methods `peek` and `peek_from` on `UnixStream` and `UnixDatagram`  (#73761).
The feature gate for the issue is `#![feature(unix_socket_peek)]`.

### About tracking issues

Tracking issues are used to record the overall progress of implementation.
They are also used as hubs connecting to other relevant issues, e.g., bugs or open design questions.
A tracking issue is however *not* meant for large scale discussion, questions, or bug reports about a feature.
Instead, open a dedicated issue for the specific matter and add the relevant feature gate label.

### Implementation history

- #73761 - initial implementation

daniel5151:
Any idea when this might get stabilized?

LinkTed:
Any news?

rijenkii:
Sadly I have not been using Rust for the last year or so, and have not used this feature for even longer.

In more than two years that this feature has existed in unstable, no bugs have been reported.
Documentation is done, tests are done, I am not sure what else is needed for stabilization but it seems that it's all good.

@rust-lang/libs-api, asking for FCP.

spindelkryp:
@KodrAus hey there! It looks like you were the initial reviewer for this functionality :) (sorry for ping!) What is the process for getting this out of nightly? Seems like it's been there for 3 years

kornelski:
The v2 RFC (https://github.com/rust-lang/rfcs/pull/3430) proposes adding `RecvOptions` for `recv_message()` extension  trait that also requires use of `IoSliceMut` and `AncillaryData`, or wrapping in and configuring a `MessageReceiver` helper.

I don't think these two APIs are in conflict. The advanced API is quite complex and peeking with it is a multi-step process. Even with the advanced API available, users could prefer `.peek()` for its simplicity and brevity.

PR #87620, opened 2021-07-30
io: Add the Peek trait

LinkTed:
Add the trait for each struct which has a peek function.

## Motivation
This new Trait make it possible to write a generic, where a stream is required, in which the data is not removed from the queue. Unfortunately, there are already peek method for some struct for example `TcpStream`, which make it not consistent (dirty).

### DIFF
--- a/library/std/src/io/mod.rs
+++ b/library/std/src/io/mod.rs
@@ -445,6 +445,18 @@ pub(crate) fn default_read_exact<R: Read + ?Sized>(this: &mut R, mut buf: &mut [
     }
 }
 
+/// The `Peek` trait allows for reading bytes from a source, without removing that data from
+/// the queue.
+///
+/// This trait behaves like [`Read`], except the bytes are not removed from the source.
+#[unstable(feature = "peek_trait", issue = "none")]
+pub trait Peek {
+    /// This function behaves like [`Read::read`], except the bytes are not removed from the
+    /// source.
+    #[unstable(feature = "peek_trait", issue = "none")]
+    fn peek(&mut self, buf: &mut [u8]) -> Result<usize>;
+}
+
 /// The `Read` trait allows for reading bytes from a source.
 ///
 /// Implementors of the `Read` trait are called 'readers'.
--- a/library/std/src/io/prelude.rs
+++ b/library/std/src/io/prelude.rs
@@ -11,4 +11,4 @@
 #![stable(feature = "rust1", since = "1.0.0")]
 
 #[stable(feature = "rust1", since = "1.0.0")]
-pub use super::{BufRead, Read, Seek, Write};
+pub use super::{BufRead, Peek, Read, Seek, Write};
--- a/library/std/src/net/tcp.rs
+++ b/library/std/src/net/tcp.rs
@@ -605,6 +605,13 @@ impl TcpStream {
     }
 }
 
+#[unstable(feature = "peek_trait", issue = "none")]
+impl Peek for TcpStream {
+    fn peek(&mut self, buf: &mut [u8]) -> io::Result<usize> {
+        self.0.peek(buf)
+    }
+}
+
 // In addition to the `impl`s here, `TcpStream` also has `impl`s for
 // `AsFd`/`From<OwnedFd>`/`Into<OwnedFd>` and
 // `AsRawFd`/`IntoRawFd`/`FromRawFd`, on Unix and WASI, and
--- a/library/std/src/net/udp.rs
+++ b/library/std/src/net/udp.rs
@@ -2,7 +2,7 @@
 mod tests;
 
 use crate::fmt;
-use crate::io::{self, Error, ErrorKind};
+use crate::io::{self, Error, ErrorKind, Peek};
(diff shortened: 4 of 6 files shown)


JohnCSimon:
Ping from triage: 
@LinkTed Can you please address the merge conflicts? 
please set `S-waiitng-on-review` afterwards

LinkTed:
> Ping from triage:
> @LinkTed Can you please address the merge conflicts?
> please set `S-waiitng-on-review` afterwards

Okay, I will that.

LinkTed:
@JohnCSimon

I rebased it, but it seems that I cannot change the label to `S-waiitng-on-review`.

kennytm:
1. should this be implemented for [`UdpSocket`](https://doc.rust-lang.org/std/net/struct.UdpSocket.html#method.peek)?
2. can https://docs.rs/std-io-peek/0.1.0/std_io_peek/ be used instead? (This PR additionally implements for UnixStream and UnixDatagram but both are unstable under #76923; adding an unstable Peek trait isn't going to change that)

LinkTed:
>     1. should this be implemented for [`UdpSocket`](https://doc.rust-lang.org/std/net/struct.UdpSocket.html#method.peek)?
I will add implement for `UdpSocket` too.
> 
>     2. can https://docs.rs/std-io-peek/0.1.0/std_io_peek/ be used instead? (This PR additionally implements for UnixStream and UnixDatagram but both are unstable under [Tracking Issue for feature(unix_socket_peek) #76923](https://github.com/rust-lang/rust/issues/76923); adding an unstable Peek trait isn't going to change that)
Should I create a new tracking issue?

kennytm:
@LinkTed 

> Should I create a new tracking issue?

What I mean is, currently stable code can't use `Unix{Stream,Datagram}::peek` as they are unstable. But stable code can't use this `Peek` trait either since it is also unstable.

If `Unix{Stream,Datagram}::peek` is stabilized then an external crate like `std-io-peek` can also do the same job. Besides, `std` itself does not rely on the `Peek` trait anywhere.

So the actual question of this is, whether the `Peek` trait is really needed to be frozen into `std`?  This is where I have doubt.

LinkTed:
> @LinkTed
> 
> > Should I create a new tracking issue?
> 
> What I mean is, currently stable code can't use `Unix{Stream,Datagram}::peek` as they are unstable. But stable code can't use this `Peek` trait either since it is also unstable.
I understand that, so this PR has to be merged after the stabilization.
> 
> If `Unix{Stream,Datagram}::peek` is stabilized then an external crate like `std-io-peek` can also do the same job. Besides, `std` itself does not rely on the `Peek` trait anywhere.
The reason, why it does not rely on anywhere, is that it was not there yet.
> 
> So the actual question of this is, whether the `Peek` trait is really needed to be frozen into `std`? This is where I have doubt.
With this argument, then many things should not be in `std`. I don't see any harm, that it would cause.

joshtriplett:
We discussed this in today's @rust-lang/libs-api meeting. We came to consensus on two points:

- As a new trait, this should use `ReadBuf`, rather than using a `&mut [u8]` buffer that requires initialization.
- We'd like to have some details about the specific use cases for having a trait abstracting this.

LinkTed:
> We discussed this in today's @rust-lang/libs-api meeting. We came to consensus on two points:
> 
>     * As a new trait, this should use `ReadBuf`, rather than using a `&mut [u8]` buffer that requires initialization.
> 
>     * We'd like to have some details about the specific use cases for having a trait abstracting this.

My use case is:
The user can write functions, where it takes the bytes from the stream (TcpStream, UnixStream, TlsStream) until a specific character, for example a new line, without using BufReader. To do so, he has to check if the character, he is looking for, is in the current buffer of the stream, by using peek. And then, read only until the character.

mbartlett21:
This seems to be `BufRead`, except that the caller provides the buffer, not the type implementing the trait. Is that right?

LinkTed:
No, really. It is like `BufRead`, where after a read line the buffer is empty.

JohnCSimon:
Triage: 
@kennytm can you please review this?

Mark-Simulacrum:
Based on https://github.com/rust-lang/rust/pull/87620#issuecomment-989183157, it sounds like this will need an evaluation of the use case presented in https://github.com/rust-lang/rust/pull/87620#issuecomment-991946525 to determine if there's a desire to add the API.

Marking as waiting-on-team for that. 

These days, it's probably the case that the discussion should be had on an [ACP](https://github.com/rust-lang/libs-team/issues/new?assignees=&labels=api-change-proposal%2C+T-libs-api&template=api-change-proposal.md&title=%28My+API+Change+Proposal%29), but I'm not sure what our policy on pre-existing PRs is there.

yaahc:
> I'm not sure what our policy on pre-existing PRs is there.

I think we've been tending to ask existing PRs to go through the ACP process.

JohnCSimon:
Triage: 
@yaahc - What (if anything) should the author do to move forward with this PR?

cc: @LinkTed

Dylan-DPC:
Closing this as it is inactive and would need an ACP

PR #87622, opened 2021-07-30
Bump bootstrap compiler to 1.55

emilyalbini:
Changing the cfgs for stdarch is missing, but my understanding is that we don't need to do it as part of this PR?

r? @Mark-Simulacrum


Mark-Simulacrum:
r=me with CI fixed.

Technically we should prepare this PR earlier (e.g., on Monday) since in the week that passes since beta branches things get added dependent on the "new" beta. But it's probably not too important.

PR #87627, opened 2021-07-30
Bugfix for license check

gilescope:
cargo fmt would only work if you were in the root dir before if a license check template was set.
(as otherwise it would not find the template.)

- is there a better way of doing this? I don't really like to mutate global state such as the current dir.

### DIFF
--- a/src/tools/rustfmt/src/config/config_type.rs
+++ b/src/tools/rustfmt/src/config/config_type.rs
@@ -64,6 +64,8 @@ macro_rules! create_config {
             // if a license_template_path has been specified, successfully read, parsed and compiled
             // into a regex, it will be stored here
             pub license_template: Option<Regex>,
+            // The location from which to resolve relative dirs.
+            root_dir: Option<PathBuf>,
             // For each config item, we store a bool indicating whether it has
             // been accessed and the value, and a bool whether the option was
             // manually initialised, or taken from the default,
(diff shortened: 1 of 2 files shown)


gilescope:
(As mentioned as an issue in https://github.com/rust-lang/rustfmt/issues/3352 )

calebcartwright:
Thank you but rustfmt PRs need to be opened against the rustfmt repo as opposed to updating the subtree here.

gilescope:
ah ok. I thought maybe I needed to do it this way as the rustfmt repo
seemed dependent upon the rustc one (E.g. dependencies on the _ast crates).
Will test via rustc repo and commit back to the main rustfmt repo.

calebcartwright:
> I thought maybe I needed to do it this way as the rustfmt repo
seemed dependent upon the rustc one (E.g. dependencies on the _ast crates).

No worries. Tool development happens in the respective repos for all tools that are subtrees or sumbodules. Tool changes then get pulled into the rustc repo at regular intervals and/or as needed to address specific issues. The tool subtrees should only be updated directly when strictly necessary with minimal changes required as part of changes tot rustc, such as AST changes that are breaking to the tool.

This PR should be closed and submitted to rust-lang/rustfmt instead

gilescope:
Thumbsup, will do

Mark-Simulacrum:
Thanks. Closing as it seems this will be refiled on rustfmt directly.

PR #87636, opened 2021-07-30
Added the `Option::unzip()` method

Kixiron:
* Adds the `Option::unzip()` method to turn an `Option<(T, U)>` into `(Option<T>, Option<U>)` under the `unzip_option` feature
* Adds tests for both `Option::unzip()` and `Option::zip()`, I noticed that `.zip()` didn't have any
* Adds `#[inline]` to a few of `Option`'s methods that were missing it

### DIFF
--- a/library/core/src/option.rs
+++ b/library/core/src/option.rs
@@ -1408,6 +1408,8 @@ impl<T, U> Option<(T, U)> {
     /// # Examples
     ///
     /// ```
+    /// #![feature(unzip_option)]
+    ///
     /// let x = Some((1, "hi"));
     /// let y = None::<(u8, u32)>;
     ///
--- a/library/core/src/option.rs
+++ b/library/core/src/option.rs
@@ -1417,7 +1417,7 @@ impl<T, U> Option<(T, U)> {
     /// assert_eq!(y.unzip(), (None, None));
     /// ```
     #[inline]
-    #[unstable(feature = "unzip_option", issue = "none", reason = "recently added")]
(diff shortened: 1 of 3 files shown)


scottmcm:
CI is failing, so I've marked this as waiting-on-author.

You can use [rustbot](https://github.com/rust-lang/triagebot/wiki/Labeling) to mark it for review again once things are fixed.

m-ou-se:
Can you [open a tracking issue](https://github.com/rust-lang/rust/issues/new?assignees=&labels=C-tracking-issue%2C+T-libs&template=library_tracking_issue.md&title=Tracking+Issue+for+XXX)? Thanks!

Kixiron:
Opened #87800

jhpratt:
I don't think `zip` and `zip_with` need `#[inline]` because they're generic.

scottmcm:
@Kixiron Could you remove 73762bf from this commit, please?

I'm happy to take the unzip method, but as @jhpratt mentions, the unrelated inlines are a different thing, so shouldn't be part of *this* commit.  Do feel free to send them as another PR if you'd like, though.  (Since generic methods are inline-eligible without the attribute, though, historically core hasn't used them, though, so I'm not sure that libs-impl wants them added.)

And for future reference on the tests, the doctests are run as part of CI, so simple methods don't always need `#[test]`s as well.  Often they're more for tedious cases, or particular edge cases that don't read well in docs.

scottmcm:
Thanks, @Kixiron !

@bors r+ rollup=maybe

PR #87637, opened 2021-07-30
Made `Vec::len()` and `Vec::is_empty()` const

Kixiron:
`Vec::len()` and `Vec::is_empty()` have no non-const behavior (they both just fetch the `len` field of `Vec`) so there's no reason for them not to be const


dtolnay (changes requested):
What is the use case for this? In const context vec.len is always 0 and vec.is_empty is always true, so I don't see why const code would need to call them.

Kixiron:
It's more of a "why not?" thing, there's no real reason that it shouldn't be const, so why not make it const?

dtolnay:
All right -- I think I would prefer not to do this in that case.

Thanks anyway!

PR #87641, opened 2021-07-30
Allow more "unknown argument" strings from linker

HackAttack:
Some toolchains emit slightly different errors, e.g.

    ppc-vle-gcc: error: unrecognized option '-no-pie'

### DIFF
--- a/compiler/rustc_codegen_ssa/Cargo.toml
+++ b/compiler/rustc_codegen_ssa/Cargo.toml
@@ -17,6 +17,7 @@ jobserver = "0.1.22"
 tempfile = "3.2"
 pathdiff = "0.2.0"
 smallvec = { version = "1.6.1", features = ["union", "may_dangle"] }
+regex = "1.4"
 
 rustc_serialize = { path = "../rustc_serialize" }
 rustc_ast = { path = "../rustc_ast" }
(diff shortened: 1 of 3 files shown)


compiler/rustc_codegen_ssa/Cargo.toml
```diff
@@ -17,6 +17,7 @@ jobserver = "0.1.22"
 tempfile = "3.2"
 pathdiff = "0.2.0"
 smallvec = { version = "1.6.1", features = ["union", "may_dangle"] }
+regex = "1.5.4"
```
klensy, line 21: There is no point in changing regex version.

compiler/rustc_codegen_ssa/src/back/link.rs
```diff
@@ -688,16 +689,17 @@ fn link_natively<'a, B: ArchiveBuilder<'a>>(
...
-        // linking executables as pie. Different versions of gcc seem to use
-        // different quotes in the error message so don't check for them.
+        // Check to see if the link failed with an error message that indicates an
+        // unknown option. If so, reperform the link step without the -no-pie
+        // option. This is safe because if the linker doesn't support -no-pie then
+        // it should not default to linking executables as pie. Different versions
+        // of gcc seem to use different quotes in the error message so don't check
+        // for them.
+        let unknown_arg_regex =
+            Regex::new(r"(unknown|unrecognized) (command line )?(option|argument)").unwrap();
```
klensy, line 699: This will recompile regex more than once

HackAttack: Didn’t notice this was in a loop. Fixed

wesleywiser (approved):

wesleywiser:
Thanks @HackAttack!

@bors r+

PR #87644, opened 2021-07-30
Recommend `swap_remove` in `Vec::remove` docs

Flying-Toast:
I was able to increase the performance (by 20%!) of my project by changing a `Vec::remove` call to `Vec::swap_remove` in a hot function. I think we should explicitly put a note in the Vec::remove docs to guide people in the right direction so they don't make a similar oversight.


the8472:
LGTM

@bors r+ rollup

wcampbell0x2a:
Could this be a good clippy lint?

PR #87492, opened 2021-07-26
Prevent dependency cycle with `-Zthir-unsafeck`

FabianWolff:
Fixes #87414.

r? @LeSeulArtichaut

PR #87114, opened 2021-07-13
Lint missing Abi in ast validation instead of lowering.

cjgillot:

PR #87645, opened 2021-07-30
Properly find owner of closure in THIR unsafeck

### LINKED ISSUE #87414: thir-unsafeck query loops on closures in const-generics
I tried this code with `-Z thir-unsafeck=on`:

```rust
fn bad<T>() -> Box<dyn Iterator<Item = [(); { |x: u32| { x }; 4 }]>> { todo!() }
fn main() {}
```

I expected to see this happen: It compiles, as it does under the MIR checker

Instead, this happened:
```
error[E0391]: cycle detected when simplifying constant for the type system `bad::{constant#0}`
 --> y.rs:1:45
  |
1 | fn bad<T>() -> Box<dyn Iterator<Item = [(); { |x: u32| { x }; 4 }]>> { todo!() }
  |                                             ^^^^^^^^^^^^^^^^^^^^^
  |
note: ...which requires simplifying constant for the type system `bad::{constant#0}`...
 --> y.rs:1:45
  |
1 | fn bad<T>() -> Box<dyn Iterator<Item = [(); { |x: u32| { x }; 4 }]>> { todo!() }
  |                                             ^^^^^^^^^^^^^^^^^^^^^
note: ...which requires const-evaluating + checking `bad::{constant#0}`...
 --> y.rs:1:45
  |
1 | fn bad<T>() -> Box<dyn Iterator<Item = [(); { |x: u32| { x }; 4 }]>> { todo!() }
  |                                             ^^^^^^^^^^^^^^^^^^^^^
note: ...which requires const checking `bad::{constant#0}`...
 --> y.rs:1:45
  |
1 | fn bad<T>() -> Box<dyn Iterator<Item = [(); { |x: u32| { x }; 4 }]>> { todo!() }
  |                                             ^^^^^^^^^^^^^^^^^^^^^
note: ...which requires processing MIR for `bad::{constant#0}`...
 --> y.rs:1:45
  |
1 | fn bad<T>() -> Box<dyn Iterator<Item = [(); { |x: u32| { x }; 4 }]>> { todo!() }
  |                                             ^^^^^^^^^^^^^^^^^^^^^
note: ...which requires unsafety-checking `bad::{constant#0}`...
 --> y.rs:1:45
  |
1 | fn bad<T>() -> Box<dyn Iterator<Item = [(); { |x: u32| { x }; 4 }]>> { todo!() }
  |                                             ^^^^^^^^^^^^^^^^^^^^^
note: ...which requires unsafety-checking `bad::{constant#0}::{closure#0}`...
 --> y.rs:1:47
  |
1 | fn bad<T>() -> Box<dyn Iterator<Item = [(); { |x: u32| { x }; 4 }]>> { todo!() }
  |                                               ^^^^^^^^
note: ...which requires building MIR for `bad::{constant#0}::{closure#0}`...
 --> y.rs:1:47
  |
1 | fn bad<T>() -> Box<dyn Iterator<Item = [(); { |x: u32| { x }; 4 }]>> { todo!() }
  |                                               ^^^^^^^^
note: ...which requires unsafety-checking `bad::{constant#0}::{closure#0}`...
 --> y.rs:1:47
  |
1 | fn bad<T>() -> Box<dyn Iterator<Item = [(); { |x: u32| { x }; 4 }]>> { todo!() }
  |                                               ^^^^^^^^
note: ...which requires unsafety-checking `bad`...
 --> y.rs:1:1
  |
1 | fn bad<T>() -> Box<dyn Iterator<Item = [(); { |x: u32| { x }; 4 }]>> { todo!() }
  | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
note: ...which requires building THIR for `bad`...
 --> y.rs:1:1
  |
1 | fn bad<T>() -> Box<dyn Iterator<Item = [(); { |x: u32| { x }; 4 }]>> { todo!() }
  | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
note: ...which requires type-checking `bad`...
 --> y.rs:1:1
  |
1 | fn bad<T>() -> Box<dyn Iterator<Item = [(); { |x: u32| { x }; 4 }]>> { todo!() }
  | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  = note: ...which again requires simplifying constant for the type system `bad::{constant#0}`, completing the cycle
note: cycle used when checking that `bad` is well-formed
 --> y.rs:1:1
  |
1 | fn bad<T>() -> Box<dyn Iterator<Item = [(); { |x: u32| { x }; 4 }]>> { todo!() }
  | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

### Meta

`rustc --version --verbose`:
```
$ rustc +local --version --verbose
rustc 1.55.0-dev
binary: rustc
commit-hash: unknown
commit-date: unknown
release: 1.55.0-dev
LLVM version: 12.0.1
```

</p>
</details>

LeSeulArtichaut:
Previously, when encountering a closure in a constant, the THIR unsafeck gets invoked on the owner of the constant instead of the constant itself, producing cycles.
Supersedes #87492. @FabianWolff thanks for your work on that PR, I copied your test file and added you as a co-author.

Fixes #87414.
r? @oli-obk

### DIFF
--- a/compiler/rustc_mir_build/src/check_unsafety.rs
+++ b/compiler/rustc_mir_build/src/check_unsafety.rs
@@ -599,13 +599,10 @@ pub fn check_unsafety<'tcx>(tcx: TyCtxt<'tcx>, def: ty::WithOptConstParam<LocalD
 
     // Closures are handled by their owner, if it has a body
     if tcx.is_closure(def.did.to_def_id()) {
-        let owner = tcx.hir().local_def_id_to_hir_id(def.did).owner;
-        let owner_hir_id = tcx.hir().local_def_id_to_hir_id(owner);
-
-        if tcx.hir().maybe_body_owned_by(owner_hir_id).is_some() {
-            tcx.ensure().thir_check_unsafety(owner);
-            return;
-        }
+        let hir = tcx.hir();
+        let owner = hir.enclosing_body_owner(hir.local_def_id_to_hir_id(def.did));
+        tcx.ensure().thir_check_unsafety(hir.local_def_id(owner));
+        return;
     }
 
     let (thir, expr) = tcx.thir_body(def);
--- a/src/test/ui/unsafe/issue-87414-query-cycle.rs
+++ b/src/test/ui/unsafe/issue-87414-query-cycle.rs
@@ -1,0 +1,15 @@
+// Regression test for #87414.
+
+// check-pass
+// compile-flags: -Zthir-unsafeck
+
+fn bad<T>() -> Box<dyn Iterator<Item = [(); { |x: u32| { x }; 4 }]>> { todo!() }
+
+fn foo() -> [(); { |x: u32| { x }; 4 }] { todo!() }
+fn bar() { let _: [(); { |x: u32| { x }; 4 }]; }
+
+// This one should not cause any errors either:
+unsafe fn unsf() {}
+fn bad2<T>() -> Box<dyn Iterator<Item = [(); { unsafe { || { unsf() } }; 4 }]>> { todo!() }
+
+fn main() {}


compiler/rustc_mir_build/src/check_unsafety.rs
```diff
@@ -599,13 +599,10 @@ pub fn check_unsafety<'tcx>(tcx: TyCtxt<'tcx>, def: ty::WithOptConstParam<LocalD
 
     // Closures are handled by their owner, if it has a body
     if tcx.is_closure(def.did.to_def_id()) {
-        let owner = tcx.hir().local_def_id_to_hir_id(def.did).owner;
```
oli-obk, line 602: just to be clear I understand, because I understood something different from your text. `owner` here is `== def.did`, and thus invoking `thir_check_unsafety` on that will obviously cycle?

LeSeulArtichaut: No, that's not the problem. Let me illustrate with an example:
```rust
fn function() {
    let closure = || {};
}
```
We're trying to compute unsafeck for `closure`, which should invoke unsafeck for the parent body `function`. So `def.did` is the DefId of the closure, and the HIR is constructed in such a way that the corresponding HirId has `function` as its owner, so I though I could use that whenever I encounter a closure.

Enter consts:
```rust
fn function() -> [(); { || {}; 4 }] {}
```
Here, if I try the same "hack", i.e. get the HirId for the closure and take its owner, I get the DefId of `function` whereas I want the DefId of the anonymous constant `{ || {}; 4 }`.

To be fair I don't entirely understand which query invocations cycle and which don't when using `ensure`, and I've also not dived into the internals of HIR to check that `enclosing_body_owner` always does the right thing, but this PR doesn't cycle on the examples provided in #87414 and passed the rest of the test suite.

oli-obk: Your explanation makes sense, thanks! I do think we should get a more solid strategy around all this, but I think it kind of overlaps with various other things like generics in anonymous constants. So let's merge this PR now

cjgillot:
I confirm that this PR fixes the ICE I had in https://github.com/rust-lang/rust/pull/87114, where I implemented essentially the same logic.

Issue #70256, opened 2020-03-22
Equality constraints on associated constants


Centril:
We currently do not support any way to constrain an associated constant as needing to be equal to some constant value. The natural way to achieve this would be to, by analogy with e.g., `Foo<0>`, also allow `Foo<ID = 0>`. The same disambiguation rules would apply as for `Foo<X>` in terms of when a block `{ ... }` is needed, e.g., `Foo<ID = { N + 1 }>`.

In compiler internal terms, we would redefine `ast::AssocTyConstraintKind` as:
```rust
/// The kinds of an `AssocConstraint`.
#[derive(Clone, RustcEncodable, RustcDecodable, Debug)]
pub enum AssocConstraintKind {
    /// E.g., `A = Bar` in `Foo<A = Bar>`.
    EqualityTy { ty: P<Ty> },
    /// E.g., `A = 0` in `Foo<A = 0>`.
    EqualityConst { ct: AnonConst },
    /// E.g. `A: TraitA + TraitB` in `Foo<A: TraitA + TraitB>`.
    Bound { bounds: GenericBounds },
}
```

cc @varkor @eddyb

Centril:
I've assigned myself, as I will be fixing the parser / lowering aspects of this to start with.

dr-orlovsky:
Any progress on this issue? I see the required changes were made in March, so it seems to be possible to add this feature for now?

varkor:
I believe only (some of) the parsing aspects were addressed, and nothing more. There is still implementation left before this feature is even ready for a feature gate.

dtolnay:
This has been implemented in nightly, so I am closing this feature request in favor of the tracking issue at https://github.com/rust-lang/rust/issues/92827.

Per https://github.com/rust-lang/rust/issues/92827#issuecomment-1032975319, a lang team MCP is eventually going to be required, along with some medium sized doc explaining what is going on and why. @oli-obk will open a hackmd for folks to collaborate in.

PR #87648, opened 2021-07-30
allow eq constraints on associated constants

JulianKnodt:
Updates #70256 

(cc @varkor, @Centril)


compiler/rustc_ast/src/ast.rs
```diff
@@ -1799,19 +1799,21 @@ impl UintTy {
...
 }
 
-/// The kinds of an `AssocTyConstraint`.
+/// The kinds of an `AssocConstraint`.
 #[derive(Clone, Encodable, Decodable, Debug)]
-pub enum AssocTyConstraintKind {
-    /// E.g., `A = Bar` in `Foo<A = Bar>`.
+pub enum AssocConstraintKind {
+    /// E.g., `A = Bar` in `Foo<A = Bar>` where A is an associated type.
     Equality { ty: P<Ty> },
```
calebcartwright, line 1814: Seems like this variant should probably be renamed as well given the name change of the enum, perhaps this?

```suggestion
    TyEquality { ty: P<Ty> },
```

oli-obk:
I have been musing about "just" adding a `TyKind::Const` and removing all the duplication we have around const generics vs type generics. Since this would also simplify our representation of associated const vs associated type as visible in this PR, maybe that's something we should pursue?

I'll give this PR a closer review on monday, but my fingers are still itching for `TyKind::Const`

JulianKnodt:
Hmmmm, I think if it's a useful tool to simplify implementations, then it's probably useful, but at the same time I can't really picture what a `TyKind::Const` means. Is it an instance of a type which is const? Or does it represent an abstract const instantiation of a type?

compiler/rustc_ast/src/ast.rs
JulianKnodt: Will handle this at some point if I maintain the additional variant!

oli-obk:
> Hmmmm, I think if it's a useful tool to simplify implementations, then it's probably useful, but at the same time I can't really picture what a `TyKind::Const` means. Is it an instance of a type which is const? Or does it represent an abstract const instantiation of a type?

it is essentially the equivalent of a built-in `struct Foo<T, const C: T>;` without the struct/`Adt`.

think about it this way: before const generics we had `typenum`, this is essentially `typenum` but directly supported by the compiler. `typenum` encodes constants in the type system, with `TyKind::Const` we do the same thing. We don't have to support any surface syntax for this, so users can't `impl Trait for 42 {}`, but the type system would support constants as first class types.

If we want to we could even represent it as such. So not even adding a `TyKind::Const`, but having `Foo` be a lang item in libcore that all const generics desugar to, but I fear that is just as much of a problem as making `bool` an `enum bool { true, false }` lang item. Nice from a cleanliness perspective, but impractical and a performance problem.

JulianKnodt:
I've never heard of typenum, I've not been working on rustc long enough to have seen it 😂. If you're going to implement it, tag me in the changes and I'll hold off on this change until that's done?

oli-obk:
I'm going to do an MCP, and if it is declined, we go with your PR, if not, I'll implement `TyKind::Const`

Note: typenum is just a regular crate on crates.io, nothing rustc specific. It was the go-to workaround for lack of const generics and still is the go-to workaround for lack of `const_evaluatable_checked`.

JulianKnodt:
Sounds good, please cc in the MCP as well, I'm interested in those changes as well!

oli-obk:
The general consensus was that `TyKind::Const` is a bad idea and wrong. But there was also sympathy for the "often we want a const-or-type enum". Maybe we could avoid the duplication here by adding such an enum and putting it into the constraint instead of having two different equality constraint variants?

oli-obk (commented):
Only got half way through

compiler/rustc_infer/src/infer/opaque_types.rs
```diff
@@ -570,9 +570,13 @@ impl<'a, 'tcx> Instantiator<'a, 'tcx> {
             debug!(?predicate);
 
             if let ty::PredicateKind::Projection(projection) = predicate.kind().skip_binder() {
-                if projection.ty.references_error() {
-                    // No point on adding these obligations since there's a type error involved.
-                    return tcx.ty_error();
+                if let Term::Ty(ty) = projection.term {
```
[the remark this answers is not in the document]
JulianKnodt: I'm not sure exactly what you meant here, but I think I've updated it to be appropriate.

compiler/rustc_middle/src/ty/mod.rs
```diff
@@ -792,6 +792,31 @@ pub struct CoercePredicate<'tcx> {
...
+}
+
+impl<'tcx> From<&'tcx Const<'tcx>> for Term<'tcx> {
+    fn from(c: &'tcx Const<'tcx>) -> Self {
+        Term::Const(c)
+    }
+}
+
+impl<'tcx> Term<'tcx> {
+    pub fn ty(&self) -> Ty<'tcx> {
```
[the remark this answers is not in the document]
JulianKnodt: I think I mostly had it for expedience, converting every place that uses a term into a match statement is a lot of places, but will try to remove it at some point. I've also moved it into an `Option<Ty>` return type which is more convenient.

compiler/rustc_infer/src/infer/opaque_types.rs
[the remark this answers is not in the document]
oli-obk: I think you forgot to push ^^

I meant keep the previous code, only replacing `if projection.ty.foo()` with `if projection.term.foo()` at all places where `foo` is a method that exists for all `TypeFoldable`s

compiler/rustc_middle/src/ty/print/pretty.rs
```diff
@@ -904,14 +906,15 @@ pub trait PrettyPrinter<'tcx>:
                     first = false;
                 }
 
-                for (assoc_item_def_id, ty) in assoc_items {
+                for (assoc_item_def_id, term) in assoc_items {
+                    let ty = if let Term::Ty(ty) = term.skip_binder() { ty } else { continue };
```
oli-obk, line 910: I think we should print them instead of skipping them. if it's not a type argument, then you can skip the generator return check and just print the constant.

compiler/rustc_middle/src/ty/sty.rs
```diff
@@ -1580,11 +1580,16 @@ impl<'tcx> ExistentialProjection<'tcx> {
     ) -> Self {
         // Assert there is a Self.
         projection_predicate.projection_ty.substs.type_at(0);
+        let ty = if let Term::Ty(ty) = projection_predicate.term {
+            ty
+        } else {
+            panic!("Only types are permitted here");
```
oli-obk, line 1586: what's the rationale for this? why can't there be constants here?

compiler/rustc_privacy/src/lib.rs
```diff
@@ -127,8 +127,11 @@ where
                 constness: _,
                 polarity: _,
             }) => self.visit_trait(trait_ref),
-            ty::PredicateKind::Projection(ty::ProjectionPredicate { projection_ty, ty }) => {
-                ty.visit_with(self)?;
+            ty::PredicateKind::Projection(ty::ProjectionPredicate { projection_ty, term }) => {
+                match term {
+                    ty::Term::Ty(ty) => ty.visit_with(self)?,
+                    ty::Term::Const(ct) => ct.visit_with(self)?,
+                }
```
oli-obk, line 134:
```suggestion
                term.visit_with(self)?;
```

compiler/rustc_privacy/src/lib.rs
```diff
@@ -1145,10 +1148,13 @@ impl<'tcx> Visitor<'tcx> for TypePrivacyVisitor<'tcx> {
             }
 
             for (poly_predicate, _) in bounds.projection_bounds {
-                if self.visit(poly_predicate.skip_binder().ty).is_break()
```
oli-obk, line 1148: you should be able to just visit the term instead of matching here

compiler/rustc_trait_selection/src/traits/auto_trait.rs
```diff
@@ -752,7 +756,8 @@ impl<'tcx> AutoTraitFinder<'tcx> {
                             // when we started out trying to unify
                             // some inference variables. See the comment above
                             // for more infomration
-                            if p.ty().skip_binder().has_infer_types() {
+                            if p.term().skip_binder().ty().map_or(false, |ty| ty.has_infer_types())
```
oli-obk, line 759:
```suggestion
                            if p.term().skip_binder().term.has_infer_types()
```

compiler/rustc_trait_selection/src/traits/auto_trait.rs
```diff
@@ -774,7 +779,8 @@ impl<'tcx> AutoTraitFinder<'tcx> {
                             // However, we should always make progress (either by generating
                             // subobligations or getting an error) when we started off with
                             // inference variables
-                            if p.ty().skip_binder().has_infer_types() {
+                            if p.term().skip_binder().ty().map_or(false, |ty| ty.has_infer_types())
```
oli-obk, line 782:
```suggestion
                            if p.term().skip_binder().has_infer_types()
```

compiler/rustc_typeck/src/check/closure.rs
```diff
@@ -706,9 +707,10 @@ impl<'a, 'tcx> FnCtxt<'a, 'tcx> {
         // Extract the type from the projection. Note that there can
         // be no bound variables in this type because the "self type"
         // does not have any regions in it.
-        let output_ty = self.resolve_vars_if_possible(predicate.ty);
+        let output_ty = self.resolve_vars_if_possible(predicate.term);
         debug!("deduce_future_output_from_projection: output_ty={:?}", output_ty);
-        Some(output_ty)
+        // FIXME(...): How to handle consts here? Will this always be a const?
```
oli-obk, line 712: yes, this is a projection on a Fn trait, so we can safely assume that it is a type

compiler/rustc_typeck/src/check/method/suggest.rs
```diff
@@ -805,10 +805,13 @@ impl<'a, 'tcx> FnCtxt<'a, 'tcx> {
                                     item_def_id: projection_ty.item_def_id,
                                 };
 
-                                let ty = pred.skip_binder().ty;
+                                let fmt = match pred.skip_binder().term {
+                                    ty::Term::Ty(ty) => format!("{}", ty),
+                                    ty::Term::Const(c) => format!("{}", c),
+                                };
```
oli-obk, line 811: I think you can just format the term

compiler/rustc_typeck/src/check/mod.rs
```diff
@@ -695,7 +695,14 @@ fn bounds_from_generic_predicates<'tcx>(
...
         // propose this instead of the following valid resugaring:
         // `T: Trait, Trait::Assoc = K` → `T: Trait<Assoc = K>`
-        where_clauses.push(format!("{} = {}", tcx.def_path_str(p.projection_ty.item_def_id), p.ty));
+        where_clauses.push(format!(
+            "{} = {}",
+            tcx.def_path_str(p.projection_ty.item_def_id),
+            match p.term {
+                ty::Term::Ty(ty) => format!("{}", ty),
+                ty::Term::Const(c) => format!("{}", c),
+            }
```
oli-obk, line 704:
```suggestion
            p.term
```

compiler/rustc_typeck/src/impl_wf_check/min_specialization.rs
```diff
@@ -199,7 +199,7 @@ fn unconstrained_parent_impl_substs<'tcx>(
     for (predicate, _) in impl_generic_predicates.predicates.iter() {
         if let ty::PredicateKind::Projection(proj) = predicate.kind().skip_binder() {
             let projection_ty = proj.projection_ty;
-            let projected_ty = proj.ty;
+            let projected_ty = proj.term.ty();
```
oli-obk, line 202: ooof specialization 🙈  . definitely needs to handle terms

src/librustdoc/clean/types.rs
```diff
@@ -1309,7 +1309,12 @@ impl FnDecl {
             FnRetTy::Return(Type::ImplTrait(bounds)) => match &bounds[0] {
                 GenericBound::TraitBound(PolyTrait { trait_, .. }, ..) => {
                     let bindings = trait_.bindings().unwrap();
-                    FnRetTy::Return(bindings[0].ty().clone())
+                    let ret_ty = bindings[0].term();
+                    let ty = match ret_ty {
+                        Term::Type(ty) => ty,
+                        Term::Constant(_c) => unreachable!(),
+                    };
```
oli-obk, line 1316: use `.ty().expect(...)`

src/tools/clippy/clippy_lints/src/methods/mod.rs
```diff
@@ -2145,12 +2145,16 @@ impl<'tcx> LateLintPass<'tcx> for Methods {
                 // one of the associated types must be Self
                 for &(predicate, _span) in cx.tcx.explicit_item_bounds(def_id) {
                     if let ty::PredicateKind::Projection(projection_predicate) = predicate.kind().skip_binder() {
+                        let assoc_ty = match projection_predicate.term {
+                          ty::Term::Ty(ty) => ty,
+                          ty::Term::Const(c) => c.ty,
```
oli-obk, line 2150: Hmm... I wonder if we can test this... maybe

```rust
#![feature(adt_const_params)]

#[derive(PartialEq, Eq)]
struct Foo;
struct Bar<const C: Foo>;
impl Foo {
    fn new() -> Bar<{ Foo }> {
        Bar
    }
}
```

but that already doesn't lint on nightly clippy, so I'm guessing you can just skip constants with a `continue`

compiler/rustc_middle/src/ty/sty.rs
JulianKnodt: Oops, I have no idea why I wrote this. I think it might've been because the existential predicate only accepts a type, so for now I only wanted to leave it for that case.

compiler/rustc_typeck/src/constrained_generic_params.rs
```diff
@@ -210,7 +210,7 @@ pub fn setup_constraining_predicates<'tcx>(
                 if !relies_only_on_inputs {
                     continue;
                 }
-                input_parameters.extend(parameters_for(tcx, &projection.ty, false));
+                input_parameters.extend(parameters_for(tcx, &projection.term.ty(), false));
```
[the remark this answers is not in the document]
JulianKnodt: hm I might be dumb, but `parameters_for` already took an `impl TypeFoldable`

compiler/rustc_typeck/src/impl_wf_check/min_specialization.rs
JulianKnodt: oops this also took a `TypeFoldable`, I don't know what else needs to be added here 🙃

compiler/rustc_infer/src/infer/error_reporting/mod.rs
```diff
@@ -1789,7 +1789,11 @@ impl<'a, 'tcx> InferCtxt<'a, 'tcx> {
                 {
                     if projection_predicate.projection_ty.item_def_id == item_def_id {
                         // We don't account for multiple `Future::Output = Ty` contraints.
-                        return Some(projection_predicate.ty);
+                        match projection_predicate.term {
+                            Term::Ty(ty) => return Some(ty),
+                            // Can return None, but not sure if that makes sense?
+                            Term::Const(_c) => todo!(),
```
oli-obk, line 1795: this is unreachable, as we specifically check for the future trait above and that has no associated consts

compiler/rustc_middle/src/ty/print/pretty.rs
```diff
@@ -904,14 +906,21 @@ pub trait PrettyPrinter<'tcx>:
...
 
-                for (assoc_item_def_id, ty) in assoc_items {
+                for (assoc_item_def_id, term) in assoc_items {
+                    let ty = match term.skip_binder() {
+                        Term::Ty(ty) => ty,
+                        Term::Const(c) => {
+                            p!(print(c));
+                            continue;
+                        }
+                    };
```
oli-obk, line 916: you need to do this near the `match ty.kind()` below, and you can't use `continue` as that will skip the `first = false`

compiler/rustc_middle/src/ty/sty.rs
```diff
@@ -1580,11 +1580,15 @@ impl<'tcx> ExistentialProjection<'tcx> {
     ) -> Self {
         // Assert there is a Self.
         projection_predicate.projection_ty.substs.type_at(0);
+        let ty = match projection_predicate.term {
+            Term::Ty(ty) => ty,
+            Term::Const(_c) => unimplemented!(),
```
oli-obk, line 1585: I guess this is where we run into the issue that existential projections also need to be able to handle terms instead of types. Do you plan this for the current PR or as a follow up (you'd need to mark the feature gate as incomplete then)

compiler/rustc_trait_selection/src/traits/auto_trait.rs
```diff
@@ -774,7 +779,7 @@ impl<'tcx> AutoTraitFinder<'tcx> {
                             // However, we should always make progress (either by generating
                             // subobligations or getting an error) when we started off with
                             // inference variables
-                            if p.ty().skip_binder().has_infer_types() {
+                            if p.term().skip_binder().ty().has_infer_types() {
```
oli-obk, line 782:
```suggestion
                            if p.term().skip_binder().has_infer_types() {
```

compiler/rustc_parse/src/parser/path.rs
```diff
@@ -499,22 +496,22 @@ impl<'a> Parser<'a> {
...
-            Some(GenericArg::Type(ty)) => return Ok(ty),
-            Some(GenericArg::Const(expr)) => {
-                self.struct_span_err(span, "cannot constrain an associated constant to a value")
-                    .span_label(ident.span, "this associated constant...")
-                    .span_label(expr.value.span, "...cannot be constrained to this value")
-                    .emit();
-            }
+        let term = match arg {
+            Some(GenericArg::Type(ty)) => ty.into(),
+            Some(GenericArg::Const(c)) => c.into(),
```
oli-obk, line 508: you need to feature gate this here

oli-obk:
ok, that was it with the review. what are your plans for the PR? Do you have all the information you need to continue?

compiler/rustc_middle/src/ty/sty.rs
JulianKnodt: I can stick it in the current PR

JulianKnodt:
@oli-obk I would of course like to finish everything in this PR, but I think the entirety of this PR would cover too broad a scope then. I think that having incremental changes (including to chalk, since I have no idea how that repo works) would be more feasible to get done quicker than having one massive blob. I do plan on implementing the `ExistentialPredicate` term, but other than that I'll make a TrackingIssue with a list of TODOs, and try to go thru those when I get time. I'll probably get swamped again with other stuff, so it may be a while before everything is done. If this doesn't seem reasonable, please let me know what else I should do.

oli-obk:
This is perfectly reasonable. In my opinion you can leave the existential predicates to a future PR, too. The only thing is adding a feature gate and tracking issue and the ICEing test on CI that actually needs doing in this PR.

compiler/rustc_trait_selection/src/traits/auto_trait.rs
```diff
@@ -752,7 +756,7 @@ impl<'tcx> AutoTraitFinder<'tcx> {
                             // when we started out trying to unify
                             // some inference variables. See the comment above
                             // for more infomration
-                            if p.ty().skip_binder().has_infer_types() {
+                            if p.term().skip_binder().ty().has_infer_types() {
```
oli-obk, line 759:
```suggestion
                            if p.term().skip_binder().has_infer_types() {
```

compiler/rustc_trait_selection/src/traits/error_reporting/mod.rs
```diff
@@ -1801,7 +1802,7 @@ impl<'a, 'tcx> InferCtxtPrivExt<'a, 'tcx> for InferCtxt<'a, 'tcx> {
             }
             ty::PredicateKind::Projection(data) => {
                 let self_ty = data.projection_ty.self_ty();
-                let ty = data.ty;
+                let ty = data.term.ty();
```
oli-obk, line 1805:
```suggestion
                let ty = data.term;
```

compiler/rustc_trait_selection/src/traits/project.rs
```diff
@@ -1802,7 +1801,9 @@ fn confirm_param_env_candidate<'cx, 'tcx>(
         Ok(InferOk { value: _, obligations }) => {
             nested_obligations.extend(obligations);
             assoc_ty_own_obligations(selcx, obligation, &mut nested_obligations);
-            Progress { ty: cache_entry.ty, obligations: nested_obligations }
+            // FIXME(...): Handle consts here as well? Maybe this progress type should just take
```
oli-obk, line 1804: add feature gate name

src/test/ui/associated-consts/assoc-const.rs
```diff
@@ -0,0 +1,21 @@
+pub trait Foo {
```
oli-obk, line 1: you can enable the feature in this test so it actually tests that everything works

nikomatsakis:
Mixed feels. Adding equality constraints seems reasonable, but it does lead the core logic into a more complex place from a theoretical perspective; this is part of why I've been inclined to avoid it so far for where clauses. I am trying to page back in the concerns though, I can't remember, I think they were perhaps a *touch* artificial given that one can force unification in other ways. But given that it's fully gated I'm not opposed.

src/test/ui/associated-consts/assoc-const.rs
```diff
@@ -0,0 +1,21 @@
...
+
+impl Foo for Bar {
+  const N: usize = 3;
+}
+
+const TEST:usize = 3;
+
+
+fn foo<F: Foo<N=3>>() {}
+//~^ ERROR associated const equality is incomplete
```
oli-obk, line 17: Why is the test still erroring?

JulianKnodt: There was something in `astconv/mod.rs` which I did not yet fix, which adds predicates for ast type bindings currently, should I fix that in this PR?

oli-obk: Ah no, that's fine. As long as you know why it's happening, I'm happy ^^

oli-obk:
A small regression is to be expected, as crates using lots of associated types (e.g. diesel) will now always go through a match on `Term` instead of immediately accessing the type.

r=me after a rebase

klensy:
Changes to llvm and cargo looks like not intended, bad rebase?

klensy:
> Changes to llvm and cargo looks like not intended, bad rebase?

@JulianKnodt @oli-obk

oli-obk:
I don't see any submodule changes... is the mobile app hiding them from me?

klensy:
cargo https://github.com/rust-lang/rust/pull/87648/files#diff-33f03135675a053d29237bf61b13441e9f52fff29207605b5eb224c2f6e3507c
llvm https://github.com/rust-lang/rust/pull/87648/files#diff-7cdb6543de7f569620c298122578d643fe4ee829c4e2aa313b129e8932f07929

oli-obk:
Oof yea, the app just bails on me on these links... I guess no more reviews in the app :(

Please revert the submodule changes and the lockfile changes!

klensy:
r-a too, https://github.com/rust-lang/rust/pull/87648/files#diff-134ddb5af2afab7937607c91728e51356f98d5f6fd3264209d7810021d42bf9d, maybe some other changes unspotted.

JulianKnodt:
Do I have permission to revert the PR? I can push to the branch again but since it's already merged

oli-obk:
You can revert the PR locally and open a new PR with the revert.

Revert all commits, squash the reverts, and only keep the reverts of the submodules and cargo lock, the functional changes can stay

JulianKnodt:
@oli-obk I think there was another commit already for updating cargo, and the linked PR should revert the llvm change.
https://github.com/rust-lang/rust/commit/bfacc5cc73dfd0093d12118b924753265e8d1b17

Urgau:
This PR changed the crate `rustdoc-json-types` which is a **public API** (nightly, but still) without increasing the format version and without pinging once rustdoc. I've send https://github.com/rust-lang/rust/pull/93132 to fix the format version.

cc @CraftSpider (as rustdoc json maintainer)

klensy:
@Urgau simply bumping version is wrong, as this pr should be reverted.

klensy:
rustdoc hitted too, btw.

oli-obk:
The changes are expected from this PR, yes

PR #87652, opened 2021-07-31
Validate that naked functions are never inlined

npmccallum:
Reject all uses of the inline attribute on naked functions.

https://github.com/rust-lang/rfcs/pull/2774
https://github.com/rust-lang/rfcs/pull/2972

cc @joshtriplett @tmiasko @Amanieu

### DIFF
--- a/compiler/rustc_passes/src/naked_functions.rs
+++ b/compiler/rustc_passes/src/naked_functions.rs
@@ -1,6 +1,6 @@
 //! Checks validity of naked functions.
 
-use rustc_ast::InlineAsmOptions;
+use rustc_ast::{Attribute, InlineAsmOptions};
 use rustc_hir as hir;
 use rustc_hir::def_id::LocalDefId;
 use rustc_hir::intravisit::{ErasedMap, FnKind, NestedVisitorMap, Visitor};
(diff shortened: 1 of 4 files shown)


src/test/ui/asm/naked-functions.rs
```diff
@@ -132,38 +143,89 @@ unsafe extern "C" fn invalid_options_continued() {
...
 
 #[cfg(target_arch = "x86_64")]
 #[naked]
+#[inline(never)]
 pub unsafe extern "C" fn valid_att_syntax() {
     asm!("", options(noreturn, att_syntax));
 }
+
+#[naked]
+//~^ WARN naked functions require #[inline(never)]
```
josephlr, line 189: The tracking issue says that this should be an error, but it's a warning in this PR. Is this just a transitional lint to avoid breaking nightly code? If so, seems reasonable to me.

npmccallum: @josephlr There's a bunch of these lints that should be errors according to the RFC. I wanted to get them all in as warnings first and then file a follow-up which switches the relevant ones to errors. This allows us to reason about them as an entire set.

compiler/rustc_passes/src/naked_functions.rs
```diff
@@ -70,10 +70,20 @@ impl<'tcx> Visitor<'tcx> for CheckNakedFunctions<'tcx> {
...
+            check_inline(self.tcx, hir_id, attrs);
         }
     }
 }
 
+/// Check that the function isn't inlined.
+fn check_inline(tcx: TyCtxt<'_>, hir_id: HirId, attrs: &[Attribute]) {
+    for attr in attrs.iter().filter(|attr| attr.has_name(sym::inline)) {
+        tcx.struct_span_lint_hir(UNSUPPORTED_NAKED_FUNCTIONS, hir_id, attr.span, |lint| {
+            lint.build("inlining naked functions is unsupported").emit();
```
Amanieu, line 82:
```suggestion
            lint.build("naked functions cannot be inlined").emit();
```
"unsupported" implies that it could be supported in theory.

npmccallum: Fixed.

Issue #87238, opened 2021-07-18
Option and Result are missing documentation on their internal order


orlp:
`Option<T>` where `T: Ord` implements `Ord` (and similarly for `PartialOrd`), but as far as I can see the order it uses isn't documented anywhere. You have to read the source to find out that its `Ord` implementation is `derive`d, and that `None` comes before `Some` in the `enum` declaration thus `None` sorts before `Some`. I think that should be documented.

The same issue also appears to apply to `Result`.

jesyspa:
Looking into this: the instances in those cases are derived, and it looks like adding extra documentation to derived traits has been proposed previously (#44118) but is not currently supported.  So by the looks of it this is not feasible without changes to rustdoc?

orlp:
> Looking into this: the instances in those cases are derived, and it looks like adding extra documentation to derived traits has been proposed previously (#44118) but is not currently supported. So by the looks of it this is not feasible without changes to rustdoc?

You could replace the derived implementation with an equivalent explicit one, and then add a comment.

jesyspa:
@rustbot release-assignment

Sorry, I don't really feel qualified to judge whether the added code complexity would be worth it, so I'll leave this to someone with more experience.

pierwill:
Looks like this issue was addressed in https://github.com/rust-lang/rust/issues/87238.

pierwill:
Looks like this issue was addressed in https://github.com/rust-lang/rust/pull/87654.

PR #87654, opened 2021-07-31
Add documentation for the order of Option and Result

jesyspa:
This resolves issue #87238.

### DIFF
--- a/library/core/src/option.rs
+++ b/library/core/src/option.rs
@@ -285,6 +285,19 @@
 //! assert_eq!(res, ["error!", "error!", "foo", "error!", "bar"]);
 //! ```
 //!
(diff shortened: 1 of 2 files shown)


scottmcm:
It's not obvious to me that the manual implementation is worth it, here.  There's also something to be said for people seeing that it's just the derived one.

Adding documentation (and doctests) about it is definitely good, though.  Maybe, as a compromise, add another section to the [module level docs](https://doc.rust-lang.org/nightly/std/option/) about how Ord/Eq/Hash/etc work with it?  There are a bunch that were added recently, and this feels like it would fit that pattern well.

jesyspa:
Thanks, that makes sense, let me write it up there :)

PR #87659, opened 2021-07-31
Fix invalid suggestions for non-ASCII characters in byte constants

### LINKED ISSUE #87397: \xHH escaping suggestions are wrong for characters that are too big
Consider this code:

```rust
fn main() {
    b'字';
    b"字";
}
```

[playground](https://play.rust-lang.org/?version=nightly&mode=debug&edition=2018&gist=72ec9cac9d7d0323e443bb972364021f)

The current output is:

```
error: non-ASCII character in byte constant
 --> src/main.rs:2:7
  |
2 |     b'字';
  |       ^^
  |       |
  |       byte constant must be ASCII
  |       help: use a \xHH escape for a non-ASCII byte: `\x5B57`

error: non-ASCII character in byte constant
 --> src/main.rs:3:7
  |
3 |     b"字";
  |       ^^
  |       |
  |       byte constant must be ASCII
  |       help: use a \xHH escape for a non-ASCII byte: `\x5B57`

error: could not compile `playground` due to 2 previous errors
```

The suggestions are incorrect: `\x5B57` is not a valid escaped character, only the `\x5B` part is interpreted as an escaped character while the `57` part is interpreted as two normal characters. 

In the case of byte characters the suggestion leads to a byte character literal with more than one character, thus in another compile error. However in the case of byte strings this will silently compile, even though that probably wasn't what the user wanted nor expected.

This happens in both the current stable and nightly compilers.

FabianWolff:
Fixes #87397.

### DIFF
--- a/src/test/ui/attributes/key-value-non-ascii.stderr
+++ b/src/test/ui/attributes/key-value-non-ascii.stderr
@@ -2,10 +2,12 @@ error: non-ASCII character in byte constant
   --> $DIR/key-value-non-ascii.rs:3:19
    |
 LL | #[rustc_dummy = b"ﬃ.rs"]
-   |                   ^
-   |                   |
-   |                   byte constant must be ASCII
-   |                   help: use a \xHH escape for a non-ASCII byte: `\xFB03`
+   |                   ^ byte constant must be ASCII
+   |
+help: if you meant to use the UTF-8 encoding of 'ﬃ', use \xHH escapes
+   |
+LL | #[rustc_dummy = b"/xEF/xAC/x83.rs"]
+   |                   ^^^^^^^^^^^^
 
 error: aborting due to previous error
 
--- a/src/test/ui/suggestions/multibyte-escapes.rs
+++ b/src/test/ui/suggestions/multibyte-escapes.rs
@@ -1,0 +1,18 @@
+// Regression test for #87397.
+
+fn main() {
+    b'µ';
+    //~^ ERROR: non-ASCII character in byte constant
+    //~| HELP: if you meant to use the unicode code point for 'µ', use a \xHH escape
+    //~| NOTE: byte constant must be ASCII
+
+    b'字';
+    //~^ ERROR: non-ASCII character in byte constant
+    //~| NOTE: this multibyte character does not fit into a single byte
+    //~| NOTE: byte constant must be ASCII
+
+    b"字";
+    //~^ ERROR: non-ASCII character in byte constant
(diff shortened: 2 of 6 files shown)


src/test/ui/attributes/key-value-non-ascii.stderr
```diff
@@ -2,10 +2,12 @@ error: non-ASCII character in byte constant
...
-   |                   ^
-   |                   |
-   |                   byte constant must be ASCII
-   |                   help: use a \xHH escape for a non-ASCII byte: `\xFB03`
+   |                   ^ byte constant must be ASCII
+   |
+help: if you meant to use the UTF-8 encoding of 'ﬃ', use \xHH escapes
+   |
+LL | #[rustc_dummy = b"/xEF/xAC/x83.rs"]
+   |                   ^^^^^^^^^^^^
```
FabianWolff, line 10: I don't know why it does that; I am definitely using backslashes in the suggestion... This is not a regression, though, because the suggestion was also wrong before.

src/test/ui/suggestions/multibyte-escapes.rs
```diff
@@ -0,0 +1,18 @@
+// Regression test for #87397.
```
davidtwco, line 1: Could add `// run-rustfix` to this test.

src/test/ui/attributes/key-value-non-ascii.stderr
davidtwco: It looks like it's happening during diagnostic emission, does this still happen if you run the test with `--error-format human-annotate-rs`?

davidtwco (approved):
LGTM, can r=me unless you want to address any of the comments.

src/test/ui/attributes/key-value-non-ascii.stderr
FabianWolff: No, with this option, the entire suggestion disappears...

src/test/ui/suggestions/multibyte-escapes.rs
FabianWolff: Unfortunately not, because the second error is not automatically fixable.

src/test/ui/attributes/key-value-non-ascii.stderr
davidtwco: Ah, maybe that option isn’t implemented fully yet.

FabianWolff:
@davidtwco Thanks for the review!

PR #87661, opened 2021-07-31
Improve error reporting for closure return type mismatches

### LINKED ISSUE #87461: return-type mismatch in async function gives "return type inferred" pointing to use of `?`
Given the following code ([playground link](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=a781a1a9802fef4d8fe8462fe746a1f4)):

```rust
async fn func() -> Result<u16, u64> {
    let _ = async {
        Err(42u64)
    }.await?;

    Ok(())
}
```

The current output is:

```
 --> src/lib.rs:6:8
  |
6 |     Ok(())
  |        ^^ expected `u16`, found `()`
  |
note: return type inferred to be `u16` here
 --> src/lib.rs:2:13
  |
2 |       let _ = async {
  |  _____________^
3 | |         Err(42u64)
4 | |     }.await?;
  | |____________^

```

An even more minimal example:

```rust
async fn func() -> Result<u16, u64> {
    Err(42u64)?;

    Ok(())
}
```

This similarly produces a note pointing to the `Err(42u64)?;` line.

The note seems entirely unhelpful here; the function's declared type doesn't match the `Ok(())`, and the use of `?` has nothing to do with it. An equivalent non-async function omits the note, and just reports the type mismatch.

The same output occurs on both stable and today's nightly.

FabianWolff:
Fixes #87461.

### DIFF
--- a/compiler/rustc_typeck/src/check/check.rs
+++ b/compiler/rustc_typeck/src/check/check.rs
@@ -78,6 +78,7 @@ pub(super) fn check_fn<'a, 'tcx>(
     fn_id: hir::HirId,
     body: &'tcx hir::Body<'tcx>,
     can_be_generator: Option<hir::Movability>,
+    return_type_pre_known: bool,
 ) -> (FnCtxt<'a, 'tcx>, Option<GeneratorTypes<'tcx>>) {
     let mut fn_sig = fn_sig;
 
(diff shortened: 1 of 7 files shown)


wesleywiser (approved):
The code looks fine to me but I'm not sure about the interaction with existing diagnostics. Most of the changed UI tests seem like they have useful notes to me.

wesleywiser:
r? @estebank 

I'm ok with merging this if the changes in diagnostics look good to you.

FabianWolff:
> The code looks fine to me but I'm not sure about the interaction with existing diagnostics. Most of the changed UI tests seem like they have useful notes to me.

Yes, you are completely right @wesleywiser; I don't remember why I thought it would be OK for these notes to disappear. I have reworked my implementation so that it now works for all existing test cases.

estebank:
I don't know what the previous code looked like, but I'm ok with this.

@bors r+

PR #61946, opened 2019-06-19
port rust for vxWorks

BaoshanPang:
The supporting for vxWorks has been enabled in this branch. Although there are still a lots of work to do, I would like to upstream the code and fix the problems later.

Please let me know if there is anything I have to do before upstream the code.

r? @alexcrichton

Thanks,
Baoshan

PR #87666, opened 2021-07-31
STD support for the ESP-IDF framework

ivmarkov:
Dear all,

This PR is implementing libStd support for the [ESP-IDF](https://github.com/espressif/esp-idf) newlib-based framework, which is the open source SDK provided by Espressif for their MCU family (esp32, esp32s2, esp32c3 and all other forthcoming ones).

Note that this PR has a [sibling PR](https://github.com/rust-lang/libc/pull/2310) against the libc crate, which implements proper declarations for all ESP-IDF APIs which are necessary for libStd support.

# Implementation approach

The ESP-IDF framework - despite being bare metal - offers a relatively complete POSIX API based on newlib. `pthread`, BSD sockets, file descriptors, and even a small file-system VFS layer. Perhaps the only significant exception is the lack of support for processes, which is to be expected of course on bare metal.

Therefore, the libStd support is implemented as a set of (hopefully small) changes to the `sys/unix` family of modules, in the form of conditional-compilation branches based either on `target_os = "espidf"` or in a couple of cases - based on `target_env = "newlib"` (the latter was already there actually and is not part of this patch).

The PR also contains two new targets:
- `riscv32imc-esp-espidf`
- `riscv32imac-esp-espidf`

... which are essentially copies of `riscv32imc-unknown-none-elf` and `riscv32imac-unknown-none-elf`, but enriched with proper `linker`, `linker_flavor`, `families`, `os`, `env` etc. specifications so that (a) the proper conditional compilation branches in libStd are selected when compiling with these targets and (b) the correct linker is used.

Since support for atomics is a precondition for libStd, the `riscv32imc-esp-espidf` target additionally is configured in such a way, so as to emit libcalls to the `__sync*` & `__atomic*` GCC functions, which are already implemented in the ESP-IDF framework. If this modification is not acceptable, we can also live with only the `riscv32imac-esp-espidf` target as well.  While the RiscV chips of Espressif lack native atomics support, the relevant instructions are transparently emulated in the ESP-IDF framework using invalid instruction trap. This modification was implemented specifically with Rust support in mind.

# Target maintainers

In case this PR eventually gets merged, you can list myself as a Target Maintainer.

More importantly, Espressif (the chip vendor) is now actively involved and [embracing](https://github.com/espressif/rust-esp32-example/blob/main/docs/rust-on-xtensa.md) all [Rust-related efforts](https://github.com/esp-rs) which were originally a community effort. In light of that, I suppose @MabezDev - who initiated the Rust-on-Espressif efforts back in time and who now works for Espressif won't object to being listed as a maintainer as well.

**EDIT:** I was hinted (thanks, @Urgau) that answering the Tier 3 policy explicitly might be helpful. Answers below.

# Tier 3 Target Policy - answers

> A proposed target or target-specific patch that substantially changes code shared with other targets (not just target-specific code) must be reviewed and approved by the appropriate team for that shared code before acceptance.

Hopefully, the changes introduced by the ESP-IDF libStd support are rather on the small side. They are completely contained within the `sys/unix` set of modules (that is, aside from the obviously necessary one-liners in the `unwind` crate and in `build.rs`).

@ivmarkov
@MabezDev

The two introduced targets follow as much as possible the naming conventions of the other targets. I.e. taking the bare-metal `riscv32imac_unknown_none_elf` as a base:
* The name of the new target was derived by replacing `none` with `espidf` to designate the `target_os`. 
* `_elf` was removed, as the non-bare metal targets seem not to have it
* `-newlib` was deliberately NOT added at the end, as I believe the chance of having two simultaneously active separate targets for the ESP-IDF framework with different C libraries (say, newlib vs musl) is way too small
* Finally, we replaced the middle `unknown` with `esp` which is kind of the name of the whole chipset MCU family (and abbreviation from Espressif which is too long). It will stay `esp` for all RiscV32-based MCUs of the company, as they all use the riscv32imc instruction set. By necessity however (disambiguation), it will be `esp32` or `esp32s2` or `esp32s3` for the Xtensa-based MCUs as all of these have their own variation of the Xtensa architecture. (The Xtensa targets are not part of this PR, even though they would use 1:1 the same LibStd implementation provided here, as they depend on the upstreaming of the Xtensa architecture support in LLVM; this upstreaming this is currently in progress.)

There was also a preceding discussion on the topic [here](https://github.com/espressif/rust-esp32-example/issues/14).

> Target names should not introduce undue confusion or ambiguity unless absolutely necessary to maintain ecosystem compatibility. For example, if the name of the target makes people extremely likely to form incorrect beliefs about what it targets, the name should be changed or augmented to disambiguate it.

We are explicitly putting an `-espidf` suffix to designate that the target is *specifically* for Rust + ESP-IDF

> Tier 3 targets may have unusual requirements to build or use, but must not create legal issues or impose onerous legal terms for the Rust project or for Rust developers or users.

Agreed.

> The target must not introduce license incompatibilities.

To the best of our knowledge, it doesn't.
 
> Anything added to the Rust repository must be under the standard Rust license (MIT OR Apache-2.0).

MIT + Apache 2.0

> The target must not cause the Rust tools or libraries built for any other host (even when supporting cross-compilation to the target) to depend on any new dependency less permissive than the Rust licensing policy. This applies whether the dependency is a Rust crate that would require adding new license exceptions (as specified by the tidy tool in the rust-lang/rust repository), or whether the dependency is a native library or binary. In other words, the introduction of the target must not cause a user installing or running a version of Rust or the Rust tools to be subject to any new license requirements.

Requirements are not changed for any other target.

> If the target supports building host tools (such as rustc or cargo), those host tools must not depend on proprietary (non-FOSS) libraries, other than ordinary runtime libraries supplied by the platform and commonly used by other binaries built for the target. For instance, rustc built for the target may depend on a common proprietary C runtime library or console output library, but must not depend on a proprietary code generation library or code optimization library. Rust's license permits such combinations, but the Rust project has no interest in maintaining such combinations within the scope of Rust itself, even at tier 3.

The targets are for bare-metal environment which is not hosting build tools or a compiler.

> Targets should not require proprietary (non-FOSS) components to link a functional binary or library.

The linker used by the targets is the GCC linker from the GCC toolchain cross-compiled for riscv. GNU GPL.

> "onerous" here is an intentionally subjective term. At a minimum, "onerous" legal/licensing terms include but are not limited to: non-disclosure requirements, non-compete requirements, contributor license agreements (CLAs) or equivalent, "non-commercial"/"research-only"/etc terms, requirements conditional on the employer or employment of any particular Rust developers, revocable terms, any requirements that create liability for the Rust project or its developers or users, or any requirements that adversely affect the livelihood or prospects of the Rust project or its developers or users.
> This requirement does not prevent part or all of this policy from being cited in an explicit contract or work agreement (e.g. to implement or maintain support for a target). This requirement exists to ensure that a developer or team responsible for reviewing and approving a target does not face any legal threats or obligations that would prevent them from freely exercising their judgment in such approval, even if such judgment involves subjective matters or goes beyond the letter of these requirements.

Agreed.

The targets implement libStd almost in its entirety, except for the missing support for process, as this is a bare metal platform. The process `sys\unix` module is currently stubbed to return "not implemented" errors.

Target does not (yet) support running tests. We would gladly provide all documentation how to build for the target (where?). It is currently hosted in this [README.md](https://github.com/ivmarkov/rust-esp32-std-hello) file, but will likely be moved to the [esp-rs](https://github.com/esp-rs) organization. Since the build for the target is driven by cargo and [all other tooling is downloaded automatically during the build](https://github.com/esp-rs/esp-idf-sys/blob/master/build.rs), there is no need for extensive documentation.
 
Agreed.

> Backlinks such as those generated by the issue/PR tracker when linking to an issue or PR are not considered a violation of this policy, within reason. However, such messages (even on a separate repository) must not generate notifications to anyone involved with a PR who has not requested such notifications.

Agreed.

> Patches adding or updating tier 3 targets must not break any existing tier 2 or tier 1 target, and must not knowingly break another tier 3 target without approval of either the compiler team or the maintainers of the other tier 3 target.

To the best of our knowledge, we believe we are not breaking any other target (be it tier 1, 2 or 3).

> In particular, this may come up when working on closely related targets, such as variations of the same architecture with different features. Avoid introducing unconditional uses of features that another variation of the target may not have; use conditional compilation or runtime detection, as appropriate, to let each target run code supported by that target.

To the best of our knowledge, we have not introduced any unconditional use of a feature that affects any other target.

> If a tier 3 target stops meeting these requirements, or the target maintainers no longer have interest or time, or the target shows no signs of activity and has not built for some time, or removing the target would improve the quality of the Rust codebase, we may post a PR to remove it; any such PR will be CCed to the target maintainers (and potentially other people who have previously worked on the target), to check potential interest in improving the situation.

Agreed.

### DIFF
--- a/library/std/src/os/espidf/fs.rs
+++ b/library/std/src/os/espidf/fs.rs
@@ -1,0 +1,117 @@
+#![stable(feature = "metadata_ext", since = "1.1.0")]
+
+use crate::fs::Metadata;
+use crate::sys_common::AsInner;
+
+#[allow(deprecated)]
+use crate::os::espidf::raw;
+
+/// OS-specific extensions to [`fs::Metadata`].
+///
+/// [`fs::Metadata`]: crate::fs::Metadata
+#[stable(feature = "metadata_ext", since = "1.1.0")]
+pub trait MetadataExt {
+    #[stable(feature = "metadata_ext", since = "1.1.0")]
+    #[rustc_deprecated(
+        since = "1.8.0",
+        reason = "deprecated in favor of the accessor \
+                  methods of this trait"
+    )]
+    #[allow(deprecated)]
+    fn as_raw_stat(&self) -> &raw::stat;
+
+    #[stable(feature = "metadata_ext2", since = "1.8.0")]
+    fn st_dev(&self) -> u64;
+    #[stable(feature = "metadata_ext2", since = "1.8.0")]
+    fn st_ino(&self) -> u64;
+    #[stable(feature = "metadata_ext2", since = "1.8.0")]
+    fn st_mode(&self) -> u32;
+    #[stable(feature = "metadata_ext2", since = "1.8.0")]
+    fn st_nlink(&self) -> u64;
+    #[stable(feature = "metadata_ext2", since = "1.8.0")]
+    fn st_uid(&self) -> u32;
+    #[stable(feature = "metadata_ext2", since = "1.8.0")]
+    fn st_gid(&self) -> u32;
+    #[stable(feature = "metadata_ext2", since = "1.8.0")]
+    fn st_rdev(&self) -> u64;
+    #[stable(feature = "metadata_ext2", since = "1.8.0")]
+    fn st_size(&self) -> u64;
+    #[stable(feature = "metadata_ext2", since = "1.8.0")]
+    fn st_atime(&self) -> i64;
+    #[stable(feature = "metadata_ext2", since = "1.8.0")]
+    fn st_atime_nsec(&self) -> i64;
+    #[stable(feature = "metadata_ext2", since = "1.8.0")]
+    fn st_mtime(&self) -> i64;
+    #[stable(feature = "metadata_ext2", since = "1.8.0")]
+    fn st_mtime_nsec(&self) -> i64;
+    #[stable(feature = "metadata_ext2", since = "1.8.0")]
+    fn st_ctime(&self) -> i64;
+    #[stable(feature = "metadata_ext2", since = "1.8.0")]
+    fn st_ctime_nsec(&self) -> i64;
+    #[stable(feature = "metadata_ext2", since = "1.8.0")]
+    fn st_blksize(&self) -> u64;
+    #[stable(feature = "metadata_ext2", since = "1.8.0")]
+    fn st_blocks(&self) -> u64;
+    #[stable(feature = "metadata_ext2", since = "1.8.0")]
+    fn st_spare4(&self) -> [u32; 2];
+}
+
+#[stable(feature = "metadata_ext", since = "1.1.0")]
+impl MetadataExt for Metadata {
+    #[allow(deprecated)]
+    fn as_raw_stat(&self) -> &raw::stat {
+        unsafe { &*(self.as_inner().as_inner() as *const libc::stat as *const raw::stat) }
+    }
+    fn st_dev(&self) -> u64 {
+        self.as_inner().as_inner().st_dev as u64
+    }
+    fn st_ino(&self) -> u64 {
+        self.as_inner().as_inner().st_ino as u64
+    }
+    fn st_mode(&self) -> u32 {
+        self.as_inner().as_inner().st_mode as u32
+    }
+    fn st_nlink(&self) -> u64 {
+        self.as_inner().as_inner().st_nlink as u64
+    }
+    fn st_uid(&self) -> u32 {
+        self.as_inner().as_inner().st_uid as u32
+    }
+    fn st_gid(&self) -> u32 {
+        self.as_inner().as_inner().st_gid as u32
+    }
+    fn st_rdev(&self) -> u64 {
+        self.as_inner().as_inner().st_rdev as u64
+    }
+    fn st_size(&self) -> u64 {
+        self.as_inner().as_inner().st_size as u64
+    }
+    fn st_atime(&self) -> i64 {
+        self.as_inner().as_inner().st_atime as i64
+    }
+    fn st_atime_nsec(&self) -> i64 {
+        0
+    }
+    fn st_mtime(&self) -> i64 {
+        self.as_inner().as_inner().st_mtime as i64
+    }
+    fn st_mtime_nsec(&self) -> i64 {
+        0
+    }
+    fn st_ctime(&self) -> i64 {
+        self.as_inner().as_inner().st_ctime as i64
+    }
+    fn st_ctime_nsec(&self) -> i64 {
+        0
+    }
+    fn st_blksize(&self) -> u64 {
+        self.as_inner().as_inner().st_blksize as u64
+    }
+    fn st_blocks(&self) -> u64 {
+        self.as_inner().as_inner().st_blocks as u64
+    }
+    fn st_spare4(&self) -> [u32; 2] {
+        let spare4 = self.as_inner().as_inner().st_spare4;
+        [spare4[0] as u32, spare4[1] as u32]
+    }
+}
--- a/library/std/src/sys/common/alloc.rs
+++ b/library/std/src/sys/common/alloc.rs
@@ -14,7 +14,8 @@ use crate::ptr;
     target_arch = "asmjs",
     target_arch = "wasm32",
     target_arch = "hexagon",
-    target_arch = "riscv32"
+    target_arch = "riscv32",
+    target_arch = "xtensa"
 )))]
 pub const MIN_ALIGN: usize = 8;
 #[cfg(all(any(
--- a/library/std/src/sys/unix/condvar.rs
+++ b/library/std/src/sys/unix/condvar.rs
@@ -34,12 +34,23 @@ impl Condvar {
     ))]
     pub unsafe fn init(&mut self) {}
 
+    // NOTE: ESP-IDF's PTHREAD_COND_INITIALIZER support is not released yet
+    // So on that platform, init() should always be called
+    // Moreover, that platform does not have pthread_condattr_setclock support,
+    // hence that initialization should be skipped as well
+    #[cfg(target_os = "espidf")]
+    pub unsafe fn init(&mut self) {
+        let r = libc::pthread_cond_init(self.inner.get(), crate::ptr::null());
+        assert_eq!(r, 0);
+    }
+
     #[cfg(not(any(
         target_os = "macos",
         target_os = "ios",
         target_os = "l4re",
         target_os = "android",
-        target_os = "redox"
+        target_os = "redox",
+        target_os = "espidf"
     )))]
     pub unsafe fn init(&mut self) {
         use crate::mem::MaybeUninit;
@@ -76,7 +87,12 @@ impl Condvar {
     // where we configure condition variable to use monotonic clock (instead of
     // default system clock). This approach avoids all problems that result
     // from changes made to the system time.
-    #[cfg(not(any(target_os = "macos", target_os = "ios", target_os = "android")))]
+    #[cfg(not(any(
+        target_os = "macos",
+        target_os = "ios",
+        target_os = "android",
+        target_os = "espidf"
+    )))]
     pub unsafe fn wait_timeout(&self, mutex: &Mutex, dur: Duration) -> bool {
         use crate::mem;
 
@@ -103,7 +119,12 @@ impl Condvar {
     // This implementation is modeled after libcxx's condition_variable
     // https://github.com/llvm-mirror/libcxx/blob/release_35/src/condition_variable.cpp#L46
     // https://github.com/llvm-mirror/libcxx/blob/release_35/include/__mutex_base#L367
-    #[cfg(any(target_os = "macos", target_os = "ios", target_os = "android"))]
+    #[cfg(any(
+        target_os = "macos",
+        target_os = "ios",
+        target_os = "android",
+        target_os = "espidf"
+    ))]
     pub unsafe fn wait_timeout(&self, mutex: &Mutex, mut dur: Duration) -> bool {
         use crate::ptr;
         use crate::time::Instant;
--- a/library/std/src/sys/unix/fd.rs
+++ b/library/std/src/sys/unix/fd.rs
@@ -91,6 +91,7 @@ impl FileDesc {
         Ok(ret as usize)
     }
 
+    #[cfg(not(target_os = "espidf"))]
     pub fn read_vectored(&self, bufs: &mut [IoSliceMut<'_>]) -> io::Result<usize> {
         let ret = cvt(unsafe {
             libc::readv(
@@ -102,9 +103,14 @@ impl FileDesc {
         Ok(ret as usize)
     }
 
+    #[cfg(target_os = "espidf")]
+    pub fn read_vectored(&self, bufs: &mut [IoSliceMut<'_>]) -> io::Result<usize> {
+        return crate::io::default_read_vectored(|b| self.read(b), bufs);
+    }
+
     #[inline]
     pub fn is_read_vectored(&self) -> bool {
-        true
+        cfg!(not(target_os = "espidf"))
     }
 
     pub fn read_to_end(&self, buf: &mut Vec<u8>) -> io::Result<usize> {
@@ -148,6 +154,7 @@ impl FileDesc {
         Ok(ret as usize)
     }
 
+    #[cfg(not(target_os = "espidf"))]
     pub fn write_vectored(&self, bufs: &[IoSlice<'_>]) -> io::Result<usize> {
         let ret = cvt(unsafe {
             libc::writev(
@@ -159,9 +166,14 @@ impl FileDesc {
         Ok(ret as usize)
     }
 
+    #[cfg(target_os = "espidf")]
+    pub fn write_vectored(&self, bufs: &[IoSlice<'_>]) -> io::Result<usize> {
+        return crate::io::default_write_vectored(|b| self.write(b), bufs);
+    }
+
     #[inline]
     pub fn is_write_vectored(&self) -> bool {
-        true
+        cfg!(not(target_os = "espidf"))
     }
 
     pub fn write_at(&self, buf: &[u8], offset: u64) -> io::Result<usize> {
@@ -217,7 +229,7 @@ impl FileDesc {
         }
     }
     #[cfg(any(
-        target_env = "newlib",
+        all(target_env = "newlib", not(target_os = "espidf")),
         target_os = "solaris",
         target_os = "illumos",
         target_os = "emscripten",
@@ -237,6 +249,12 @@ impl FileDesc {
             }
             Ok(())
         }
+    }
+    #[cfg(target_os = "espidf")]
+    pub fn set_cloexec(&self) -> io::Result<()> {
+        // FD_CLOEXEC is not supported in ESP-IDF but there's no need to,
+        // because ESP-IDF does not support spawning processes either.
+        Ok(())
     }
 
     #[cfg(target_os = "linux")]
@@ -268,7 +286,17 @@ impl FileDesc {
         // We want to atomically duplicate this file descriptor and set the
         // CLOEXEC flag, and currently that's done via F_DUPFD_CLOEXEC. This
         // is a POSIX flag that was added to Linux in 2.6.24.
-        let fd = cvt(unsafe { libc::fcntl(self.raw(), libc::F_DUPFD_CLOEXEC, 0) })?;
+        #[cfg(not(target_os = "espidf"))]
+        let cmd = libc::F_DUPFD_CLOEXEC;
+
+        // For ESP-IDF, F_DUPFD is used instead, because the CLOEXEC semantics
+        // will never be supported, as this is a bare metal framework with
+        // no capabilities for multi-process execution.  While F_DUPFD is also
+        // not supported yet, it might be (currently it returns ENOSYS).
+        #[cfg(target_os = "espidf")]
+        let cmd = libc::F_DUPFD;
+
+        let fd = cvt(unsafe { libc::fcntl(self.raw(), cmd, 0) })?;
         Ok(FileDesc::new(fd))
     }
 }
--- a/library/std/src/sys/unix/os.rs
+++ b/library/std/src/sys/unix/os.rs
@@ -128,6 +128,12 @@ pub fn error_string(errno: i32) -> String {
     }
 }
 
+#[cfg(target_os = "espidf")]
+pub fn getcwd() -> io::Result<PathBuf> {
+    Ok(PathBuf::from("/"))
+}
+
+#[cfg(not(target_os = "espidf"))]
 pub fn getcwd() -> io::Result<PathBuf> {
     let mut buf = Vec::with_capacity(512);
     loop {
@@ -154,6 +160,12 @@ pub fn getcwd() -> io::Result<PathBuf> {
     }
 }
 
+#[cfg(target_os = "espidf")]
+pub fn chdir(p: &path::Path) -> io::Result<()> {
+    super::unsupported::unsupported()
+}
+
+#[cfg(not(target_os = "espidf"))]
 pub fn chdir(p: &path::Path) -> io::Result<()> {
     let p: &OsStr = p.as_ref();
     let p = CString::new(p.as_bytes())?;
@@ -432,6 +444,11 @@ pub fn current_exe() -> io::Result<PathBuf> {
     path.canonicalize()
 }
 
+#[cfg(target_os = "espidf")]
+pub fn current_exe() -> io::Result<PathBuf> {
+    super::unsupported::unsupported()
+}
+
 pub struct Env {
     iter: vec::IntoIter<(OsString, OsString)>,
 }
@@ -541,6 +558,7 @@ pub fn unsetenv(n: &OsStr) -> io::Result<()> {
     }
 }
 
+#[cfg(not(target_os = "espidf"))]
 pub fn page_size() -> usize {
     unsafe { libc::sysconf(libc::_SC_PAGESIZE) as usize }
 }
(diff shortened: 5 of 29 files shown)


library/std/src/sys/unix/fd.rs
```diff
@@ -66,6 +69,10 @@ const fn max_iov() -> usize {
     16 // The minimum value required by POSIX.
 }
 
+// ESP-IDF does not (yet) support duplication of file descriptors, so we are emulating it here
```
the8472, line 72: That kind of emulation seems dicy since duped file descriptors normally don't share a seek position, nonblocking status and various other properties. Here they would.

How much would break if `duplicate` always returned an error?

library/std/src/os/espidf/fs.rs
```diff
@@ -0,0 +1,126 @@
...
+    fn st_size(&self) -> u64 {
+        self.as_inner().as_inner().st_size as u64
+    }
+    fn st_atime(&self) -> i64 {
+        self.as_inner().as_inner().st_atime as i64
+    }
+    fn st_atime_nsec(&self) -> i64 {
+        //self.as_inner().as_inner().st_atime_nsec as i64
+        0
+    }
```
the8472, line 97: Since this trait is platform-specific it doesn't need to have methods it can't implement properly.

library/std/src/sys/unix/fd.rs
ivmarkov: This is here just to support off-the-shelf Rust crates (I've so far only found [Rocket](https://rocket.rs/) relying on it, but I'm unsure how many more are there). But no pressing use case, and Rocket is honestly a bit too big for this MCU anyway.

In any case and based on your explanations, this implementation is as flawed as it gets, so I'll just eradicate it completely tomorrow.

If proper file duplication semantics turns out to be necessary, I don't see any other way besides implementing it in the ESP-IDF framework itself.

library/std/src/os/espidf/fs.rs
ivmarkov: Agreed - will remove tomorrow as well. Ditto for all the others there returning 0.

ivmarkov: Btw, would you want me to continue force-pushing, or shall I start accumulating changes-after-feedback as separate commits, so that the progress is easier to track?

the8472: either is fine. in the latter case you can rewrite the history before merge.

ivmarkov: OK. While we are at the `os/espidf/mod.rs` topic. Any hits regarding what shall I put as a "version" there?
I'm referring to this [FIXME](https://github.com/rust-lang/rust/pull/87666/files#diff-5b9c55eb49b42c03102371ec9b45d87ead04588d68415c07a1166186438a5eb9).

the8472: Current nightly is 1.56, so if it were to land now and you wanted the new platform features to be instantly stable with the the first release you can put that version there.
If those traits are not critical for the platform to be usable and you have some reason to iterate on them you could mark them unstable instead.

the8472: Oh wait, that's for the raw types, not the metadata extensions? I guess those are important and should be stable with the the first release.

ivmarkov: Neither is top priority, as using filesystems on the MCU is a bit of a niche case I think, due to the limited flash size in the MCU itself and the need to jump to external SD card for these use cases.

Let me sleep over it to figure out what is most appropriate regarding stable/unstable.
 
Threads and sockets is most important to get right first.

the8472:
Note that as part of only supporting fallible allocation in the linux kernel (https://github.com/rust-lang/rfcs/pull/3140) there is ongoing discussion around making std features optional and how to support that on the cargo side. https://github.com/rust-lang/rfcs/pull/3146

I'm not sure what exactly the roadmap on those parts are but it sounds like it would be appropriate here if you only want to support a subset of the standard library in the beginning. But putting whole module behind feature gates that have yet to be introduced instead of stubbing them could require some work.

library/std/src/sys/unix/fd.rs
ivmarkov: Hey I was [refreshing my knowledge](https://www.gnu.org/software/libc/manual/html_node/Duplicating-Descriptors.html) on `dup()`/`dup2()`/`FD_DUPFD` this morning (as I wrote this code 9 months ago!), and I don't think your interpretation "since duped file descriptors normally don't share a seek position, nonblocking status and various other properties" is correct.

According to the above link, it is exactly the opposite: dup-ed file descriptors DO share seek position, nonblocking status and various other properties. Duped file descriptors are nothing else but aliases of each other, and their only (limited) use seems to be lifetime management, where you can close the original file descriptor while the duped one continues to live and vice versa.

I think more or less this is what my implementation also does, doesn't it? (we can forget about the `CLOEXEC` semantics as that's irrelevant when you don't have processes)?

As for the use case with Rocket (remembered that too) - Rocket was `dup`-ing the file descriptor (socket really) so that it spins a separate socket writer from socket reader, and manages their lifetimes separately too.

ivmarkov:
> Note that as part of only supporting fallible allocation in the linux kernel ([rust-lang/rfcs#3140](https://github.com/rust-lang/rfcs/pull/3140)) there is ongoing discussion around making std features optional and how to support that on the cargo side. [rust-lang/rfcs#3146](https://github.com/rust-lang/rfcs/pull/3146)
> 
> I'm not sure what exactly the roadmap on those parts are but it sounds like it would be appropriate here if you only want to support a subset of the standard library in the beginning. But putting whole modules behind feature gates that have yet to be introduced instead of stubbing them could require some work.

That's very useful info, thanks for sharing!

My problem with it is a little bit that this patch set has been in development since approx last November, and waiting for another Cargo nightly feature to first of all hit nightly, and then suffering through its stabilization pains, change of semantics and so on might be a bit too much. Especially considering that we are already sailing in an ocean of Cargo unstable features, that - when [changed](https://github.com/rust-lang/cargo/issues/9554#issuecomment-888095498) or regressed - [did hit us](https://github.com/rust-lang/cargo/issues/9641) in the [past](https://github.com/rust-lang/cargo/issues/8945):
- [build-std](https://github.com/rust-lang/wg-cargo-std-aware)
- [Extra link arg](https://github.com/rust-lang/cargo/issues/9426)
- [Configurable env](https://github.com/rust-lang/cargo/issues/9539)
- [Artifact dependencies](https://github.com/rust-lang/cargo/issues/9096) - lower prio; waiting on that, hasn't even hit nightly yet

So I would rather take a conservative approach here and stub what is not supported, if there is no strong, well-grounded objection from the Rust maintainers.

library/std/src/sys/unix/fd.rs
the8472: Yeah, you're right. I guess I was thinking of the mechanism that reopens a file via procfs, which isn't relevant here.

Ignore everything I have said here. 😔

library/std/src/os/espidf/fs.rs
ivmarkov: OK, so I've force-pushed a cleanup of the `os/espidf` module:
* First of all, I decided to leave as stable everything, because, well, filesystems on the ESP-IDF might be a corner case but they are there, the `stat()` function works (even if a bit slow on the SPIFFS FS), so why shall I be seeking for unstable, where it does not make sense
* I've removed all commented out code that should not be there, double-checked that the `raw::stat` module is a faithful representation of what is in there in the ESP-IDF in terms of stat structure & members
* Some stubbed out code *MUST* stay and return 0. Turns out, the "platform specific" `MetadataExt` trait is not so platform specific, as it is used by `os::unix::fs` which is re-exposing its methods to the outside world. Since the ESP-IDF claims to be in the unix family, it *must* support all methods `os::unix::fs` re-expose, even if a subset of those (the `*_nsec` ones) return 0. I'm not setting a precedent: our sibling bare-metal RTOS `vxworks` port [does the same](https://github.com/rust-lang/rust/blob/master/library/std/src/os/vxworks/fs.rs#L82) with `*_nsec`, for the very same reasons.  
* Finally, I've kept versions at 1.1.0 (for the deprecated stuff) and at 1.8.0 for the `MetadataExt` trait as everybody else does. Again, an example: the vxworks port [was merged](https://github.com/rust-lang/rust/pull/61946) ~ 2019, while 1.8.0 is at least 3 years older (2016), let alone version 1.1.0. 
* Moreover: I would actually argue that re-using these versions as they are is even the correct behavior: imagine a cross-platform crate that uses this stuff, either via `os::unix::fs` or `os::platform` and sets the minimum compiler version requirement to 1.8.0. From the point of view of the crate that's the correct behavior, as everybody else except the ESP-IDF port is having those since 1.8.0. Now, people willing to consume that fictitious crate with ESP-IDF will of course never use a 1.8.0 version of the compiler simply because ESP-IDF support is not there. So putting a higher version on those crate items does not seem to help. Quite the opposite.

Amanieu:
> Since support for atomics is a precondition for libStd, the `riscv32imc-esp-espidf` target additionally is configured in such a way, so as to emit libcalls to the `__sync*` & `__atomic*` GCC functions, which are already implemented in the ESP-IDF framework. If this modification is not acceptable, we can also live with only the `riscv32imac-esp-espidf` target as well. While the RiscV chips of Espressif lack native atomics support, the relevant instructions are transparently emulated in the ESP-IDF framework using invalid instruction trap. This modification was implemented specifically with Rust support in mind.

This is fine as long as the underlying implementation is lock-free (doesn't use a lock internally to emulate atomics). The requirement here is that a signal/interrupt handler should never deadlock if it tries to use an `Atomic*` which the interrupted thread was in the middle of accessing.

Generally this is implemented by disabling interrupt handlers or by having the interrupt handler restart the atomic sequence in the interrupted thread upon returning. We already have something similar for ARMv4 on Linux (atomic instructions were introduced in ARMv6) where support for atomic operations is provided by the kernel: https://github.com/rust-lang/compiler-builtins/blob/master/src/arm_linux.rs

ivmarkov:
> > Since support for atomics is a precondition for libStd, the `riscv32imc-esp-espidf` target additionally is configured in such a way, so as to emit libcalls to the `__sync*` & `__atomic*` GCC functions, which are already implemented in the ESP-IDF framework. If this modification is not acceptable, we can also live with only the `riscv32imac-esp-espidf` target as well. While the RiscV chips of Espressif lack native atomics support, the relevant instructions are transparently emulated in the ESP-IDF framework using invalid instruction trap. This modification was implemented specifically with Rust support in mind.
> 
> This is fine as long as the underlying implementation is lock-free (doesn't use a lock internally to emulate atomics). The requirement here is that a signal/interrupt handler should never deadlock if it tries to use an `Atomic*` which the interrupted thread was in the middle of accessing.

I think this is implemented by disabling the interrupt handlers upon entry in the `__sync` / `__atomic` functions and then restoring the previous state upon exit. In any case, @igrr can shed some light on both approaches.

> Generally this is implemented by disabling interrupt handlers or by having the interrupt handler restart the atomic sequence in the interrupted thread upon returning. We already have something similar for ARMv4 on Linux (atomic instructions were introduced in ARMv6) where support for atomic operations is provided by the kernel: https://github.com/rust-lang/compiler-builtins/blob/master/src/arm_linux.rs

Thanks for the link!

igrr:
> > > __sync* & __atomic* GCC functions, which are already implemented in the ESP-IDF framework

> > This is fine as long as the underlying implementation is lock-free (doesn't use a lock internally to emulate atomics).

> I think this is implemented by disabling the interrupt handlers upon entry in the __sync / __atomic functions and then restoring the previous state upon exit.

That's correct — `__sync_*` and `__atomic_*` built-ins are implement in esp-idf by disabling interrupts, performing the requested operation, then re-enabling interrupts. So the built-in implementation isn't expected to be preempted.

Currently this implementation is hosted in esp-idf but we could move (or copy) it into `rust-lang/compiler-builtins` at some point, if we wanted to support atomics on rv32imc without a dependency on esp-idf.

library/std/src/os/espidf/fs.rs
ivmarkov: @the8472 Just curious if you consider these changes acceptable?

the8472: Sounds reasonable considering the precedent and that `os::unix` reexports stuff.

library/std/src/sys/unix/os.rs
```diff
@@ -566,10 +583,16 @@ pub fn unsetenv(n: &OsStr) -> io::Result<()> {
...
 }
 
+#[cfg(not(target_os = "espidf"))]
 pub fn page_size() -> usize {
     unsafe { libc::sysconf(libc::_SC_PAGESIZE) as usize }
 }
 
+#[cfg(target_os = "espidf")]
+pub fn page_size() -> usize {
+    32
```
Amanieu, line 593: What is the significance of this number? AFAIK RISC-V uses 4K pages. A comment here would be useful.

Amanieu (commented):
Overall, I don't feel very confident about the file descriptor duplication hack. It seems likely that it will break horribly with conversions to/from raw FDs.

I would feel more comfortable if `duplicate` just returned an error.

library/std/src/sys/unix/os.rs
MabezDev: Using the RISC-V standard of 4K pages might not make sense when add support for the Xtensa based Espressif chips (once support is upstreamed in LLVM of course). 

I also believe setting a smaller page size reduces the stack usage of threads where the stack size is not a multiple of the page size, according to the `Thread::new` impl: https://github.com/rust-lang/rust/blob/6fe0886723c9e08b800c9951f1c6f6a57b2bf22c/library/std/src/sys/unix/thread.rs#L63-L66

So I think we want to keep this number low.

library/std/src/sys/unix/fd.rs
```diff
@@ -264,13 +288,61 @@ impl FileDesc {
...
+                        map.remove(&self.fd);
+                    }
+                    close
+                }
+                None => true,
+            }
+        }
+
+        #[cfg(not(target_os = "espidf"))]
+        true
```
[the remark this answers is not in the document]
ivmarkov: OK. Since there had been multiple concerns raised with the `duplicate` hack, I'll remove it. Thanks for the code cleanup though.

library/std/src/sys/unix/os.rs
igrr: Since none of the chips supported by esp-idf so far has an MMU, we can indeed keep this number low. Perhaps 4 bytes would work?

However it is odd that this chip-specific property ends up depending on `#[cfg(target_os = "espidf")]`.

Long term, my preference would be implementing `sysconf`  and `_SC_PAGESIZE` in ESP-IDF and deciding on the right value there (possibly depending on the chip target).

Amanieu: If this is just about the stack alignment then maybe it would be better to create a `stack_alignment` function in the thread module which can return `page_size` by default but a different value for `espidf`.

Reporting a page size of 32 is just confusing, even if this function is internal to `std`.

ivmarkov:
All,

I've pushed a small set of changes addressing the feedback accumulated so far. The changes are not force-pushed sop you can also only look at the latest (3rd) commit in the PR:

`compiler/rustc_target/src/spec/riscv32imc_esp_espidf.rs`
- I've adjusted the comment a bit as I would expect when/if this PR is merged to only keep this target, and remove the `riscv32imac-esp-espidf` one. We don't really need two targets for the same thing, and if you agree that emulating atomics with `__atomic*` /` __sync*` in ESP-IDF is the way forward, I'll remove the other target

`library/std/src/sys/unix/fd.rs`:
- I've removed the "duplicate" hack completely
- The `set_cloexec()` no-op is still necessary though, or else even basic TcpStream::connect & read test fails on the ESP-IDF
- I've changed a bit the existing `duplicate()` function, so that ESP-IDF gets `F_DUPFD` instead of `F_DUPFD_CLOEXEC`. Unlike `F_DUPFD_CLOEXEC`, `F_DUPFD` might be eventually supported (currently returns ENOSYS as expected).

`library/std/src/sys/unix/os.rs`:
- `page_size()` is now NOT defined on the ESP-IDF. That's OK, because:
(a) ESP-IDF does its own internal alignment of the stack size, so the Rust one can simply be switched off, as I did
(b) `thread::guard` and `stack_verflow.rs` where `page_size()` was also called are actually compiled as no-op on the ESP-IDF so we are in the end OK with NOT having `page_size` defined for the ESP-IDF at all. :)

`library/std/src/sys/unix/thread.rs`:
- I've implemented a code path for ESP-IDF where the stack-alignment logic is completely switched off. As per above, this logic is not necessary for the ESP-IDF, as it is anyway automatically aligning the stack size as per the requirements of the FreeRTOS.

library/std/src/sys/unix/os.rs
ivmarkov: Please see [this summary of my latest changes](https://github.com/rust-lang/rust/pull/87666#issuecomment-893460598). No more `page_size` on ESP-IDF.

library/std/src/sys/unix/fd.rs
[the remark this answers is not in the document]
ivmarkov: As per [my last set of changes](https://github.com/rust-lang/rust/pull/87666#issuecomment-893460598), the "duplicate" hack is no longer a part of the PR.

library/std/src/sys/unix/fd.rs
```diff
@@ -264,13 +288,61 @@ impl FileDesc {
...
+    fn add_ref(&self) {
+        let mut map = MAP.lock().unwrap();
+        match map.get_mut(&self.fd) {
+            Some(refcnt) => {
+                *refcnt = *refcnt + 1;
+            }
+            None => {
+                map.insert(self.fd, 2);
+            }
+        };
```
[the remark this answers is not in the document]
ivmarkov: Indeed - but no longer relevant, as all of this code is removed.

ivmarkov:
Btw, any idea why these checks are now failing? Very strange, all does compile just fine locally for me...

Amanieu:
Github actions are on fire: https://www.githubstatus.com/

ivmarkov:
Github actions are back and the build is clean again.

Anything else that we should discuss regarding this PR? Is it in a mostly acceptable state for merging now?
* E.g. shall I now remove the `riscv32imac-esp-espidf` target and keep only the `riscv32imc-esp-espidf` one? 
* Anything else that needs cleanup?

Sorry - don't mean to push - just trying to make the best use of the few remaining days of my vacation.

Amanieu:
Everything looks good and ready to merge.

Regarding `riscv32imac-esp-espidf`, it's up to you. If `riscv32imc-esp-espidf` is sufficient and you don't have any actual (or future) hardware that supports atomics then there isn't much point adding it.

ivmarkov:
> Everything looks good and ready to merge.

That's great! Thank you for helping me drive this effort.

> Regarding `riscv32imac-esp-espidf`, it's up to you. If `riscv32imc-esp-espidf` is sufficient and you don't have any actual (or future) hardware that supports atomics then there isn't much point adding it.

I've removed the `riscv32imac-esp-espidf` target in favour of the `riscv32imc-esp-espidf` one which does generate `__async*` libcalls.

Let me know if you need me to rebase the whole PR as a single commit prior to merging.

One last thing: once the new version of the [libc crate is released](https://github.com/rust-lang/libc/pull/2312#issuecomment-891561270), we'll have to raise [the libc dependency in libStd from V0.2.98, to 0.2.99](https://github.com/rust-lang/rust/blob/master/library/std/Cargo.toml#L18), as this PR of course relies on the now-merged ESP-IDF changes in the libc crate.

But I guess we can do this post-factum once the new version of libc is released.

library/std/src/sys/common/alloc.rs
```diff
@@ -14,7 +14,8 @@ use crate::ptr;
     target_arch = "asmjs",
     target_arch = "wasm32",
     target_arch = "hexagon",
-    target_arch = "riscv32"
+    target_arch = "riscv32",
+    target_arch = "xtensa"
```
Disasm, line 18: I think xtensa is not supported yet.

library/std/src/sys/unix/condvar.rs
```diff
@@ -34,12 +34,23 @@ impl Condvar {
     ))]
     pub unsafe fn init(&mut self) {}
 
+    // NOTE: ESP-IDF's PTHREAD_COND_INITIALIZER support is not released yet
```
Disasm, line 37: Maybe "is not implemented yet"?

library/std/src/sys/common/alloc.rs
ivmarkov: Support for Xtensa touches multiple crates and does not necessarily land into all of them in an atomic fashion:
- (1) `libc` crate: Here, it landed for the first time [> 1 year ago](https://github.com/rust-lang/libc/pull/1748) and was [recently extended](https://github.com/rust-lang/libc/pull/2310) with extra APIs and reworked to not really mean "Xtensa" but ESP-IDF for Risv32 *and* Xtensa and any architecture supported by ESP-IDf in fact.
- (2) `library/std`: Here, it lands now. The only place that needs "Xtensa" explicitly listed (as opposed to branching based on `target_os = "espidf"`) is this one AFAIR.
- (3) `rust/compiler/rustc_target`: Not landed yet, but we have a patchset for that. However, the Xtensa arch needs to be upstreamed in LLVM first.

Now, if you or somebody else feels uncomfortable having this conditional before the actual support for the Xtensa *targets* being upstreamed first, I can remove that line.

But if it does not do harm, why bother? We'll have to add it back anyway once (3) lands. And in the meantime, we'll know that libStd is Xtensa "ready" once the Xtensa targets get upstreamed.

Amanieu:
Let's wait for libc to publish a new release before merging this.

ivmarkov:
> Let's wait for libc to publish a new release before merging this.

No problem. Keeping fingers crossed that this would still happen before 1.56 is released.

ivmarkov:
Now that libc `0.2.99` is released, I've forced pushed the following two updates:
- Modification of `./Cargo.lock` and `./library/std/Cargo.toml` - only change is that the `libc` dependency is now raised to 0.2.99
- I've rebased all commits so far into a single one to keep the rust master branch history cleaner. Other than the rebase, no changes to the already reviewed code

Issue #79308, opened 2020-11-22
Loop without side-effect is not eliminated. Leads to O(n) instead of O(1) runtime


the8472:
Adding the follow method as part of a benchmark to library/alloc/benches/vec.rs

```rust
#[repr(transparent)]
pub struct Foo(usize);


#[inline(never)]
pub fn vec_cast(input: Vec<Foo>) -> Vec<usize> {
    input.into_iter().map(|e| unsafe { std::mem::transmute(e) }).collect()
}
```

which exercises [this specialization in Vec](https://github.com/rust-lang/rust/blob/5d5ff84130da0d74c6ece368dbe821d8f83fa526/library/alloc/src/vec.rs#L2451-L2497)

results in the following assembly (extracted with objdump):

```assembly
0000000000086130 <collectionsbenches::vec::vec_cast>:
   86130:	48 8b 0e             	mov    (%rsi),%rcx
   86133:	48 89 f8             	mov    %rdi,%rax
   86136:	48 8b 56 08          	mov    0x8(%rsi),%rdx
   8613a:	48 8b 7e 10          	mov    0x10(%rsi),%rdi
   8613e:	48 89 ce             	mov    %rcx,%rsi
   86141:	48 85 ff             	test   %rdi,%rdi
   86144:	74 10                	je     86156 <collectionsbenches::vec::vec_cast+0x26>
   86146:	48 8d 34 f9          	lea    (%rcx,%rdi,8),%rsi
   8614a:	48 c1 e7 03          	shl    $0x3,%rdi
   8614e:	66 90                	xchg   %ax,%ax
   86150:	48 83 c7 f8          	add    $0xfffffffffffffff8,%rdi ; <= RDI unused from here onwards
   86154:	75 fa                	jne    86150 <collectionsbenches::vec::vec_cast+0x20>
   86156:	48 29 ce             	sub    %rcx,%rsi
   86159:	48 89 08             	mov    %rcx,(%rax)
   8615c:	48 89 50 08          	mov    %rdx,0x8(%rax)
   86160:	48 c1 fe 03          	sar    $0x3,%rsi
   86164:	48 89 70 10          	mov    %rsi,0x10(%rax)
   86168:	c3                   	retq   
```

The ghidra decompile for the same function (comments are mine):

```C
void collectionsbenches::vec::vec_cast(long *param_1,long *param_2)

{
  long lVar1;
  long lVar2;
  long lVar3;
  long lVar4;
  
[...]

PR #87667, opened 2021-07-31
add module-level documentation for vec's in-place iteration

the8472:
As requested in the last libs team meeting and during previous reviews.

Feel free to point out any gaps you encounter, after all non-obvious things may with hindsight seem obvious to me.

r? @yaahc 

CC @steffahn


library/alloc/src/vec/source_iter_marker.rs
```diff
@@ -1,3 +1,100 @@
+///! Inplace iterate-and-collect specialization for `Vec`
+///!
+///! This applies to iterators in the shape of
+///! `source.adapter().adapter().adapter().collect::<Vec<U>>()`
+///! where `source` is a `Vec`, `Box<[]>` or `BinaryHeap`, the adapters each
```
SkiFire13, line 5: This is a bit unclear, can `.adapter()` be `.into_iter()`? If not then `source` should be the corresponding `IntoIterator::IntoIter` type of those listed.

`Box<[]>` doesn't implement `IntoIterator` (I feel like this is something that could have be done with the 2021 edition). You can still get an owned iterator over its contents but you first have to convert it to a `Vec` with `From`/`Into`, so it falls under the `Vec` case.

library/alloc/src/vec/source_iter_marker.rs
```diff
@@ -1,3 +1,100 @@
...
+///!    its read pointer at least once so that the result can be written into the allocation
+///!    without overwriting yet-to-be-iterated items. Also counts the number of items written.
+///! 3. on panic the written items must be dropped before the iterator is dropped which will
+///!    then drop any unread items. This requires the write count to be accurate when the panic occurs.
+///! 4. on success the source is accessed again, any unread items are dropped and then the allocation
+///!    is stolen so it doesn't get dropped with the iterator.
+///!    Since this changes the state of the source while still being owned by the iterator adapters
+///!    they must be able to handle this situation in their drop impl.
+///!    This is a safety requirement of `SourceIter`.
+///! 5. assemble a new vec from From the capacity, count of written items and the allocation pointer
```
SkiFire13, line 33: There's an extra "From" here

library/alloc/src/vec/source_iter_marker.rs
```diff
@@ -1,3 +1,100 @@
...
+///! to be implemented when the items don't have a destructor. Thus that implicit requirement
+///! also makes the specialization safe to use for in-place collection.
+///!
+///! The combined requirements of all these unsafe traits and the need to implement them on each
+///! part of the iterator pipeline - since there is no central struct evaluating an iterator
+///! holistically - leads to the need for non-local reasoning, i.e. the safety has to be evaluated
+///! taking the behavior of `IntoIter`, the adapters and this specialization into account.
+///! Interactions with other features such as clonable iterators, panics inside user-provided closures
+///! and partially consumed (including reverse iteration) ones also have to be considered.
+///! For example a previous implementation of `InPlaceIterable` on `Peek` turned out to be unsound
```
SkiFire13, line 51: Technically it's called `Peekable`, not `Peek`

library/alloc/src/vec/source_iter_marker.rs
```diff
@@ -1,3 +1,100 @@
...
+///! hold.
+///!
+///!
+///! Some examples that are optimized by this specialization, more can be found in the `Vec`
+///! benchmarks:
+///!
+///! ```rust
+///! /// Converts a usize vec into an isize one.
+///! fn cast(vec: Vec<usize>) -> Vec<isize> {
+///!   // Does not allocate, free or panic. On optlevel>=2 it does not loop.
```
SkiFire13, line 63: The "On optlevel>=2 it does not loop" is unfortunately true only for some cases. Sometimes LLVM produces empty loops.

the8472: Well, it's true for the usize/isize case, there even is a codegen test for that. And that it refers to opt-level itself should make clear that it depends on optimizations, no?

library/alloc/src/vec/source_iter_marker.rs
the8472: > This is a bit unclear, can .adapter() be .into_iter()? If not then source should be the corresponding IntoIterator::IntoIter type of those listed.

I'll add the `into_iter()`.

>  You can still get an owned iterator over its contents but you first have to convert it to a Vec with From/Into, so it falls under the Vec case.

Indeed, the key part is that the conversion doesn't perform any allocations either so vec's iter basically is box's iter.

the8472: I went with calling `source` the respective `IntoIter`s since that'll be what holds the allocation which the rest of the text talks about. The original collection will have ceased to exist at that point.

library/alloc/src/vec/source_iter_marker.rs
SkiFire13: My observation was meant to be something that might be worth looking into (and maybe highlight in the docs) rather than a correction.

the8472: Oh, yeah, there already is #79308 for that. I can link to it.

library/alloc/src/vec/source_iter_marker.rs
```diff
@@ -1,3 +1,106 @@
+///! Inplace iterate-and-collect specialization for `Vec`
```
scottmcm, line 1: Aren't these supposed to be `//!` comments?

the8472: 🤦‍♂️ And there I was wondering why my IDE was confused again and wouldn't render the comment. Fixed.

library/alloc/src/vec/source_iter_marker.rs
```diff
@@ -1,3 +1,108 @@
...
+//! bounds.
+//!
+//! By extension some other collections which use `collect::Vec<_>()` internally in their
+//! `FromIterator` implementation benefit from this too.
+//!
+//! Access to the underlying source goes through a further layer of indirection via the private
+//! trait `AsIntoIter` to hide the implementation detail that other collections may use
+//! `vec::IntoIter` internally.
+//!
+//! In-place iteration performs the following steps
```
steffahn, line 19: You proceed to give some semi-complete description of implementation details how the specialized `from_iter` implementation in this module operates. The description seems hard (or impossible) to follow without also looking at the actual source code. AFAIK, explaining in words exactly what some source code does is often not a very useful form of documentation.

You intermix some short mentioning of relevant safety requirements, related issues, important points. Thus this is probably more than just the “description of implementation details” that I called it in the previous paragraph. I suppose, the heading “In-place iteration performs the following steps” doesn’t hint at the presence of all this extra information either, though. Some (or all?) of the additional information provided here seems like it might be better placed alongside the implementation itself.

**Overall, I’m wondering who is intended to read this text and for what reason.**

I can think of someone who
* wants to read and understand the specialization infrastructure and concrete implementation(s) of `.collect()`
  * _but_ they’d read the source code, too, so for them there’s probably lots of redundancy, and a lack of linking the explainations to the actual source code in these docs
* wants to get a feeling of which cases of calling `.collect()` results in what kind of runtime behavior regarding optimization of code and number of allocations
  * _but_ they wouldn’t care about the exact implementation and wouldn’t necessarily read some `alloc`-internal (private) module level docs in the first place
* wants to modify the implementations in or around this module
  * _but_ they’d prefer to work with concrete invariants attatched to concrete `unsafe` functions and traits
  * I guess they do appreciate your explicit mention of the fact that the current situation requires lots of “holistic”, “non-local” reasoning because it tells them that they won’t get all the relevant details regarding memory safety of the existing specializations in the form of concrete invariants on concrete functions/traits

<hr>

Let me show in an example how there’s a problem of unclear target audience for these docs as well as some general lack of details (or alternatively references to other places that fill in the details)

Take for example the phrase “For this to be sound the source has to use pointers too to perform its reads, otherwise there would be aliasing conflicts.” in step 1. From reading this sentence, I don’t know the exact location in the source code that this part discusses, I don’t know if there’s sufficient documentation on these soundness requirements on the trait infrastructure for `SourceIter` (and/or other traits) already and if yes where to find those. I’m not told that this isn’t discussing some generic properties of a widely-used trait infrastructure; instead “the source” is referring to just two types, `IntoIter` for `Vec` and for `BinaryHeap`.

And what to take from this “A has to do B” statement!? Does it tell me that it’s important to know that these two types use raw pointers directly? Great thing to point out in the right place in the source code, although I’m not actually sure what the best place would be. Possibly somewhere on/in the `SpecFromIter` impl in this very module. Or does it say that it’s important that we never change the fact that these two types use raw pointers? Maybe add a documentation note in the implementation of those `IntoIter` types. Provided there isn’t such a note already. Otherwise someone working on these `IntoIter` iterators may very well never learn about this requirement in the first place. This seems like a good instance of the “non-local” reasoning you mentioned and explaining it here doesn’t seem to add too much value. Maybe as a FIXME to find the right place to document this interaction: We do eventually want to get rid of any need for non-local reasoning here, right?

I could probably to the same kind of analysis like on this example to other parts of the text below, too, but I’m out of time for now, any it would probably be too much to discuss at once anyways.

library/alloc/src/vec/source_iter_marker.rs
```diff
@@ -1,3 +1,108 @@
+//! Inplace iterate-and-collect specialization for `Vec`
+//!
+//! This applies to iterators in the shape of
+//! `source.adapter().adapter().adapter().collect::<Vec<U>>()`
+//! where `source` is an owning iterator obtained from `Vec`, `Box<[]>` (by conversion to `Vec`)
+//! or `BinaryHeap`, the adapters each consume one or more items per step
+//! (represented by `InPlaceIterable`), provide transitive access to `source` (via `SourceIter`)
+//! and thus the underlying allocation. And finally the layouts of `source::Item` and `U` must
```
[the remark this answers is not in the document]
the8472: Well, I assume that humans can parse pseudocode, but your proposal sounds fine.

library/alloc/src/vec/source_iter_marker.rs
```diff
@@ -1,3 +1,108 @@
...
+//! have the same size and alignment, this is currently ensured via const eval instead of trait
+//! bounds.
+//!
+//! By extension some other collections which use `collect::Vec<_>()` internally in their
+//! `FromIterator` implementation benefit from this too.
+//!
+//! Access to the underlying source goes through a further layer of indirection via the private
+//! trait `AsIntoIter` to hide the implementation detail that other collections may use
+//! `vec::IntoIter` internally.
```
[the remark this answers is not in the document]
the8472: `AsIntoIter` being private hides the fact that `binary_heap::IntoIter` uses `vec::IntoIter` internally. Originally I had a`impl SourceIter<Source = vec::IntoIter> for binary_heap::IntoIter` which somewhat exposed this detail. This came up during review of the original PR. Considering that all those traits are unstable and `doc(hidden)` I don't think this is strictly necessary but it was requested anyway and thus I'm explaining it here.

Does that answer your question? If so which part was missing?

library/alloc/src/vec/source_iter_marker.rs
```diff
@@ -1,3 +1,108 @@
+//! Inplace iterate-and-collect specialization for `Vec`
+//!
+//! This applies to iterators in the shape of
```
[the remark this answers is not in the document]
the8472: `This` here refers to the specialization from the previous sentence, which the module contains.

library/alloc/src/vec/source_iter_marker.rs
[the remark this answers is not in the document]
steffahn: Yep, that answers my question ^^

library/alloc/src/vec/source_iter_marker.rs
the8472: > Thus this is probably more than just the “description of implementation details” that I called it in the previous paragraph.

I could rewording the "performs the following steps" part then. I think the context of what is does helps to provide context for all the things that need to be considered.

Or do you think it would be better to just drop the information that this module requires the interaction of multiple unsafe traits to work correctly and then just leave the reader to figure out the details by reading the implementation and the documentation on those other traits?

> Overall, I’m wondering who is intended to read this text and for what reason.

It's not aimed at one specific group. Basically anyone involved in the standard library touching this area. This can be a reviewer, someone who explicitly wants to work on this or someone who needs to understand it in some other context such as the issues we had recently about unsoundness and needs a bird's eye view of what's happening here.

> and a lack of linking the explainations to the actual source code in these docs

The issue here is that afaik rustdoc has no mechanism to link to the source and some of the traits are `doc(hidden)` so they can't be linked to either. But I can give the latter a try anyway.

> Some (or all?) of the additional information provided here seems like it might be better placed alongside the implementation itself.

Except that "the implementation" is not a single place given all the moving parts that are involved. So to give an overview it has to be placed in one piece so that the reader doesn't have to collect all the individual pieces first to get a picture.

> I’m not told that this isn’t discussing some generic properties of a widely-used trait infrastructure; instead “the source” is referring to just two types, IntoIter for Vec and for BinaryHeap.

Well, I am considering to extend this to `VecDeque` since that should be feasible with the `make_contiguous()` function it gained a while ago. But that's speculative at the moment.

> We do eventually want to get rid of any need for non-local reasoning here, right?

We can certainly reduce it, but ultimately you'll at least have to reason about the interaction of multiple unsafe traits, a dozen implementations thereof and the safe things those implementations can do (such as the clones and panics mentioned). I think having a place that explains how those things fit together still helps even if in principle each piece already does its part in isolation.

> Or does it say that it’s important that we never change the fact that these two types use raw pointers? Maybe add a documentation note in the implementation of those IntoIter types. Provided there isn’t such a note already.

There isn't. Maybe `AsIntoIter` should be made unsafe, document that requirement and expose the methods needed for in-place iteration, that'll also help implementing this for `VecDeque`.


----

Anyway, I fail to take away many actionable items from this. I'll try adding some links and put some extra docs on `AsIntoIter`

steffahn: > The issue here is that afaik rustdoc has no mechanism to link to the source and some of the traits are `doc(hidden)` so they can't be linked to either. But I can give the latter a try anyway.

Regarding `doc(hidden)`, the linking may still work when you’re using something like
```
RUSTDOCFLAGS='--document-private-items --document-hidden-items' x.py doc library/std
```
to build the docs. Except if `doc(hidden)` items generate error messages when you intra-doc-link to them (from private items, which are usually not rendered either), I haven’t tested that much yet. Unfortunately I myself seem to have recently broken this command (with the `--document-hidden-items`) by replacing the import of `TrustedRandomAccess` with `TrustedRandomAccessNoCoerce` in a module that had a hidden doc link to `TrustedRandomAccess`… oops… maybe we should fix this and somehow add a test for this?

> Except that "the implementation" is not a single place given all the moving parts that are involved. So to give an overview it has to be placed in one piece so that the reader doesn't have to collect all the individual pieces first to get a picture.

Maybe this can be fixed… I just remembered something I’ve seen in the Haskell standard library. Take a look [here for example](https://hackage.haskell.org/package/containers-0.6.5.1/docs/src/Data.Map.Internal.html). There’s some named “[Note: …]” sections explaining implementation details that are lengthy and/or relevant at multiple places in the source code. Then, comments in the code refer to them by `See note: …`. Assuming that you’re unfamiliar with the lexical syntax of Haskell, note that these notes and the references to them are only part of comments, not any rendered documentation, and I personally don’t know if these “See notes: …” links have any sort of tooling that allows you to follow them when reading the source code or if they are just human convention. Anyways, my takeaway: We could copy this style/approach somewhat. Instead of the rather unstructured approach of just randomly addressing important soundness considerations hidden in the middle of a section describing itself as “In-place iteration performs the following steps”, you could give every notable aspect, a small headline \[e.g. **“Use of pointers for inplace-from_iter”**, etc…] and a self-contained description, and then simply refer to it from multiple relevant parts of the source code that all form “the implementatation”.



> Or do you think it would be better to just drop the information that this module requires the interaction of multiple unsafe traits to work correctly and then just leave the reader to figure out the details by reading the implementation and the documentation on those other traits?

AFAICT, the implementation is not all that long, so realistically reading the implementation is an approach that everyone trying to gaining an understanding of the implementation will attempt. Maybe all we need is to improve the inline docs of the implementation to be easier to understand without too much prior knowledge on all the traits involved, i.e. the approach could be: Make the implementation itself an even better and more readable alternative to the list you created here.

Even if we don’t do this, I think there could be: References in the implementation referring back to the “steps” laid out in the module-level docs (in order to link those more clearly, and also to keep this list of steps up to date); separating the description of what happens from discussion of safety considerations, either by completely separating the step-by-step description of what’s done from the following discussion, or at least by visibly separating the “what happens” part from the rest _within_ each of the steps listed. Also I wouldn’t mind if you added – on every aspect you’re discussing – an explanation whether that aspect is _fully_ covered/explained by documented safety-invariants on unsafe functions/traits, or if – instead – there’s non-local reasoning involved. The latter (i.e. explanations of some instance of non-local reasoning) should probably best be called out explicitly everywhere, so the reader knows to pay attention and maybe even to encourage them to think about future improvements that could make requirements more locally documented.



> We can certainly reduce it, but ultimately you'll at least have to reason about the interaction of multiple unsafe traits, a dozen implementations thereof and the safe things those implementations can do (such as the clones and panics mentioned). I think having a place that explains how those things fit together still helps even if in principle each piece already does its part in isolation.

I think you’re addressing two different things here. There’s definitely merit in explaining interactions of multiple unsafe traits, etc, even when each trait already has its own documentation with a correct complete list of safety requirements, etc. And then there’s explanations of implementation details that are only sound because we applied knowledge of lots of other implementation details of other parts in the standard library, i.e. the “non-local reasoning” we’re talking about. Making clear to the reader which is which seems valuable.

Regarding fixing the non-local stuff in the future, I think some reasonably low-effort approaches for improving the situation could be to (not mutually exclusive):
* Point out important details as you did, but put them in a place that can be referenced from multiple places, then reference them from all relevant places. (See the Haskell comparison above.)
* Make clear indications about which other parts in the standard library depend on implementation details of a function / struct / etc… especially across file boundaries. Similarly indicate when unsafe code relies on implementation details of other functions / structs / etc, especially if those other functions / types might seem unrelated. (And make sure to also add “back-references” in those implementations then.)
* For **private** (and often small) specialization traits or helper traits like `AsIntoIter`, consider documenting the trait with the exhaustive list of implementors and the exhaustive of users. (For `AsIntoIter` that’s 2 implementors and one use-case, I think.) This way we don’t necessarily need to come up with a complete list of safety invariants immediately, it could just say “This trait is implemented by `vec::IntoIter` and `binary_heap::IntoIter` and used exclusively for the `SpecFromIter` implementation in `vec/source_iter_marker.rs` where unsafe code depends on the implementation details of those types.” (Still, also marking the trait and its method `unsafe` would be useful, too.) Someone who wants to add implementations would either need to understand the use-case enough in order to soundly add an implementor (updating the documentation accordingly) or they would need to flesh out some explicit safety invariants first. Arriving at an intermediate state where _some_ invariants are listed (without claiming to be complete) and the use-cases and implementors are still listed is also reasonable.

steffahn: > Anyway, I fail to take away many actionable items from this.

Of course! I’m doing more of a writing-down-my-thoughts thing here, sorry in case that isn’t too helpful. I’ll probably look back into this tomorrow and see if I can come up with any more concrete suggestions for small, actionable improvements. I certainly don’t want you to _remove anything_ that you’ve written and considered relevant. More documentation is most often better than less, and it can also always be further improved or moved to a different place in the future.


> I'll try adding some links and put some extra docs on `AsIntoIter`

That sounds great! More “links” (and be it just in the form of “explaining where to look for more information”, not necessarily literal “links”) is probably useful, in particular to make things easier to understand in-depth. And I mentioned the suggestion to be more explicit on what things are just outlining things that are documented more in-depth elsewhere vs. what things are pointing out otherwise undocumented or barely documented non-local-reasoning kind of stuff.

steffahn: > Unfortunately I myself seem to have recently broken this command

In the meantime, `RUSTDOCFLAGS='--document-private-items --document-hidden-items -Arustdoc::broken-intra-doc-links' x.py doc library/std` helps.

the8472: Thanks, I think I can work with some of that.

Regarding the notes, headlines should work fine for that since they're already linkable, as long as the module itself is linkable.

JohnCSimon:
Triage: 
@the8472 - what is the status of this PR? It's sat idle for a few months. Still waiting on review?

the8472:
Yes, still waiting on review.

yaahc:
That's on me, I'll prioritize reviewing it this week.

yaahc:
@the8472 how have you been navigating these docs as you write them? I noticed you have a bunch of cross references to existing specialization traits like `TrustedRandomAccessNoCoerce` when I started reading through the module doc which aren't part of the PR. I wanted to be able to follow these cross references to help understand the docs you're adding so I tried pulling your pr branch and generating the docs locally but when I tried running `RUSTDOCFLAGS="--document-private-items" x doc library/alloc --open` it is failing to build because of some unresolved links.

the8472:
`--document-private-items --document-hidden-items`

yaahc:
> --document-private-items --document-hidden-items

hmm, didn't seem to fix it but I just manually fixed the one dead link and it runs fine now. Surprised you're not running into the same issue but w/e. I'll check if the same issue is on master later and fix that separately. _continues with review_

the8472:
I also have `deny-warnings = false` in my config.toml that may affect rustdoc too.

yaahc:
@the8472, can you mark the comments that have been resolved as such?

yaahc:
awesome, thank you very much @the8472!

@bors r+

PR #87673, opened 2021-07-31
Tweak opaque type mismatch error

estebank:


src/test/ui/async-await/dont-suggest-missing-await.stderr
```diff
@@ -1,13 +1,14 @@
...
-...
 LL |         take_u32(x)
    |                  ^ expected `u32`, found opaque type
    |
-   = note: while checking the return type of the `async fn`
+note: while checking the return type of the `async fn`
+  --> $DIR/dont-suggest-missing-await.rs:7:24
+   |
+LL | async fn make_u32() -> u32 {
+   |                        ^^^ checked the `Output` of this `async fn`, found opaque type
```
davidtwco, line 11: Pre-existing, but I could imagine this note being confusing without some background knowledge of the `Future` type (and the accompanying `Output` associated type) and `async fn` sugar.

Something along the lines of

```
`async fn`s return `impl Future` opaque types unless `await`ed
```

might be more approachable?

davidtwco (approved):
LGTM, r=me unless you want to address the suggestion comment.

src/test/ui/async-await/dont-suggest-missing-await.stderr
estebank: @nellshamrell is improving this label in https://github.com/rust-lang/rust/pull/86705. I have [a similar inclination of adding a bit more info about `impl Future<Output = Ty>`](https://github.com/rust-lang/rust/pull/86705#discussion_r680915935), but maybe not for _all_ errors, because I know there was some resistance to the idea (although I no longer remember what the opposing position's points were 😅).

estebank: Follow up work at https://github.com/rust-lang/rust/commit/bfcaf8be327826cf32e861e66a90d34e98b02f3d

Issue #71126, opened 2020-04-14
Tracking Issue for `destructuring_assignment`


varkor:
This is a tracking issue for the RFC "Destructuring assignment" (https://github.com/rust-lang/rfcs/pull/2909).
The feature gate for the issue is `#![feature(destructuring_assignment)]`.

### About tracking issues

Tracking issues are used to record the overall progress of implementation.
They are also uses as hubs connecting to other relevant issues, e.g., bugs or open design questions.
A tracking issue is however *not* meant for large scale discussion, questions, or bug reports about a feature.
Instead, open a dedicated issue for the specific matter and add the relevant feature gate label.

### Steps

- [x] Open a pull request: https://github.com/rust-lang/rust/pull/71156
	- [x] Tuples (https://github.com/rust-lang/rust/pull/78748)
    - [x] Structs and slices (https://github.com/rust-lang/rust/pull/78836)
    - [x] `_` expressions (https://github.com/rust-lang/rust/pull/79016)
- [x] Open an RFC: https://github.com/rust-lang/rfcs/pull/2909.
- [ ] Adjust documentation ([see instructions on rustc-dev-guide][doc-guide])
- [x] Stabilization PR (https://github.com/rust-lang/rust/pull/90521)

[stabilization-guide]: https://rustc-dev-guide.rust-lang.org/stabilization_guide.html#stabilization-pr
[doc-guide]: https://rustc-dev-guide.rust-lang.org/stabilization_guide.html#documentation-prs

### Implementation history

The initial implementation was carried out in three steps by @fanzier (split out from an initial implementation, https://github.com/rust-lang/rust/pull/71156):
- Implementing destructuring support for tuples: https://github.com/rust-lang/rust/pull/78748
- Implementing destructuring support for structs and slices: https://github.com/rust-lang/rust/pull/78836
- Implementing destructuring support for `_` expressions: https://github.com/rust-lang/rust/pull/79016

varkor:
We've opened an RFC for destructuring assignment: https://github.com/rust-lang/rfcs/pull/2909.

varkor:
Now that #78748, #78836, and #79016 have been merged, destructuring assignment is now implemented. Thanks to @fanzier for their hard work implementing this feature!

bstrie:
I got kind of a weird error message recently that probably isn't too concerning, but maybe someone more familiar with the implementation of the `_` desugaring might have more insight.

The following extremely weird program:
```rust
fn main() {
[...]

PR #87676, opened 2021-08-01
2229: Handle MutBorrow/UniqueImmBorrow better

### LINKED ISSUE #56: Code sometimes uses `UniqueImmBorrow` and not `MutBorrow`
Discussion here:

https://github.com/rust-lang/rust/pull/86995#discussion_r666814528

arora-aman:
We only want to use UniqueImmBorrow when the capture place is truncated and we
drop Deref of a MutRef.

r? @nikomatsakis

Fixes: https://github.com/rust-lang/project-rfc-2229/issues/56

### DIFF
--- a/compiler/rustc_typeck/src/check/upvar.rs
+++ b/compiler/rustc_typeck/src/check/upvar.rs
@@ -350,9 +350,13 @@ impl<'a, 'tcx> FnCtxt<'a, 'tcx> {
 
         for (place, mut capture_info) in capture_information {
             // Apply rules for safety before inferring closure kind
-            let place = restrict_capture_precision(place);
+            let (place, capture_kind) =
+                restrict_capture_precision(place, capture_info.capture_kind);
+            capture_info.capture_kind = capture_kind;
 
-            let place = truncate_capture_for_optimization(&place);
+            let (place, capture_kind) =
+                truncate_capture_for_optimization(place, capture_info.capture_kind);
+            capture_info.capture_kind = capture_kind;
 
             let usage_span = if let Some(usage_expr) = capture_info.path_expr_id {
                 self.tcx.hir().span(usage_expr)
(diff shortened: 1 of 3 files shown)


nikomatsakis (approved):

compiler/rustc_typeck/src/check/upvar.rs
```diff
@@ -1843,7 +1885,7 @@ fn adjust_for_move_closure<'tcx>(
         _ if first_deref.is_some() => {
             let place = match first_deref {
                 Some(idx) => {
-                    place.projections.truncate(idx);
+                    let (place, _) = truncate_place_to_len(place, kind, idx);
```
nikomatsakis, line 1888: why don't we care about `kind` here?

nikomatsakis: is this equivalent to just calling `truncate(idx)`?

compiler/rustc_typeck/src/check/upvar.rs
```diff
@@ -2072,6 +2116,49 @@ fn determine_capture_info(
...
 
+/// Truncates `place` to have up to `len` projections.
+/// `curr_mode` is the current required capture kind for the place.
+/// Returns the truncated `place` and the updated required capture kind.
+///
+/// Note: Capture kind changes from `MutBorrow` to `UniqueImmBorrow` if the truncated part of the `place`
+/// contained `Deref` of `&mut`.
+fn truncate_place_to_len(
+    mut place: Place<'tcx>,
+    curr_mode: ty::UpvarCapture<'tcx>,
```
nikomatsakis, line 2127: I wonder if we want this to be

```rust
truncate_place_to_len(mut place: Place<'tcx>, curr_mode: &mut ty::UpvarCapture<'tcx>, len: usize)
```

or even

```rust
truncate_place_to_len(place: &mut Place<'tcx>, curr_mode: &mut ty::UpvarCapture<'tcx>, len: usize)
```

It seemed like a lot of the callers (*almost* all) were doing

```rust
let (..., mode) = truncate_place_to_len(place, curr_mode, ...);
curr_mode = mode;
```

which is a bit silly.

arora-aman: My thoughts:
- Looking at just the function all it's not directly evident that the `mode` has changed, might need a better name.
- Some codepaths don't really have a `&mut Place` (eg:  within Delegate -- we need to handle repr packed) so that might need adding a line to create an owned `Place` variable.

Can we use unstable features? https://github.com/rust-lang/rust/issues/71126 looks is near stabilization and the syntax is what I'd prefer, i.e. directly update the variables than destructure and then assign

nikomatsakis: We are permitted to use unstable features, yes.

nikomatsakis (approved):

PR #31579, opened 2016-02-11
Enable /LARGEADDRESSAWARE & /SAFESEH for i686-pc-windows-msvc

ollie27:
/LARGEADDRESSAWARE is already enabled for i686-pc-windows-gnu so we should probably be consistent.
https://msdn.microsoft.com/en-us/library/wz223b1z.aspx

/SAFESEH is a good thing to enable by default.
https://msdn.microsoft.com/en-us/library/9a89h429.aspx

PR #87680, opened 2021-08-01
Update stacker and psm crates

mati865:
Primarily to include https://github.com/rust-lang/stacker/pull/54

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2729,9 +2729,9 @@ dependencies = [
 
 [[package]]
 name = "psm"
-version = "0.1.11"
+version = "0.1.16"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "96e0536f6528466dbbbbe6b986c34175a8d0ff25b794c4bacda22e068cd2f2c5"
+checksum = "cd136ff4382c4753fc061cb9e4712ab2af263376b95bbd5bd8cd50c020b78e69"
 dependencies = [
  "cc",
 ]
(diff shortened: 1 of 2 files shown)


mati865:
cc @nagisa

JohnTitor:
Failed in rollup: https://github.com/rust-lang/rust/pull/87714#issuecomment-891522033
@bors r-
Seems we need some tweaks to build the crate.

mati865:
Updated to include https://github.com/rust-lang/stacker/commit/948e60454a9ada5b223e8f9ea8b2c4a0ccfd6647

klensy:
Now this updates some other crates too.

mati865:
Rust-analyzer updated them while I was committing, fixed now.

Mark-Simulacrum:
@bors retry libpsm-b2b57a2c9f3f7c91.rlib(x86_windows_gnu.o) : error LNK2026: module unsafe for SAFESEH image.

Maybe spurious? Could be an actual problem. Not sure.

nagisa:
99.9% is not spurious but im still perplexed why is it coming up now.

mati865:
Definitely not spurious.
Before https://github.com/rust-lang/stacker/commit/e1dc224de6f209f531f0dd444fdf5598a72cd418#diff-af3494399696f0e51254641cc231eda98f371372b92fb4f453e43d7f8229b08c MSVC env would always use MSVC assembly files(even for Clang), now it uses GNU assembly for Clang.

nagisa:
That\s surprising if true – clang-cl does not understand MASM, so it couldn't possibly have compiled the MSVC-specific files. And yet with `CC` set to `clang-cl` we can't possibly be falling back to regular MSVC's `cl` either.

mati865:
I have no experience with MSVC but maybe Clang-cl will automatically fallback to MASM?

nagisa:
So. I was able to reproduce this. Firstly, this only affects `i686-pc-windows-msvc` as far as I can tell. Secondly, this is caused by a combination of https://github.com/rust-lang/rust/pull/31579 (which seems to be "just because") and use of `clang-cl` for assembling (which does not appear to support outputting objects that are compatible with this flag). In particular the `psm`'s assembly contains no exception handlers at all, so it has worked “by default” when using `ml` to assemble.

`clang-cl` does not support transparent fallback to `ml` when encountering masm files, so I'm still not all that sure how we managed to avoid hitting issues with this before now, though. We must have been using `ml` at all times in the past, at least as far as `msvc` target is concerned. This is confusing.

Issue #87335, opened 2021-07-21
Tracking Issue for RFC 3137: let-else statements


joshtriplett:
This is a tracking issue for [RFC 3137: let-else statements](https://github.com/rust-lang/rfcs/pull/3137).
The feature gate for the issue is `#![feature(let_else)]`.

### About tracking issues

Tracking issues are used to record the overall progress of implementation.
They are also used as hubs connecting to other relevant issues, e.g., bugs or open design questions.
A tracking issue is however *not* meant for large scale discussion, questions, or bug reports about a feature.
Instead, open a dedicated issue for the specific matter and add the relevant feature gate label.

### Steps

- [x] Implement the RFC (https://github.com/rust-lang/rust/pull/87688)
- [ ] Adjust documentation: reference PR at https://github.com/rust-lang/reference/pull/1156
- [ ] Stabilization PR: https://github.com/rust-lang/rust/pull/93628

[doc-guide]: https://rustc-dev-guide.rust-lang.org/stabilization_guide.html#documentation-prs
[stabilization-guide]: https://rustc-dev-guide.rust-lang.org/stabilization_guide.html#stabilization-pr

### Unresolved Questions

- [x] **Readability in practice**: Will `let ... else { ... };` be clear enough to humans in practical code, or will some introducer syntax be desirable?
- [x] **Conflicts with if-let-chains**: Does this conflict too much with the if-let-chains RFC or vice-versa? Neither this feature nor that feature should be stabilized without considering the other.
- [x] **Amount of special cases**: Are there too many special-case interactions with other features?
- [x] **Grammar clarity**: Does the grammar need to be clarified? This RFC has some slightly unusual grammar requirements.

### Implementation history

- https://github.com/rust-lang/rust/pull/87688 implemented let-else.
- https://github.com/rust-lang/rust/pull/89841 adds support for type annotations in let-else (`let x: Type = ... else { ... }`)

inquisitivecrystal:
Minor meta notes: the link to the RFC is broken and the F-let-else label is the wrong color. :)

This comment should probably be collapsed once they've been fixed.

LeSeulArtichaut:
Is there already someone who wants to implement the RFC? Otherwise I could maybe take a stab at it

camsteffen:
I volunteered on Zulip to implement a while ago and I have a start on it. @Fishrock123 are you still okay with me taking it since you were also interested?

Fishrock123:
[...]

PR #80357, opened 2020-12-24
Introduce `hir::ExprKind::Let` - Take 2

c410-f3r:
Builds on #68577 and depends on #79328.

cc #53667

Issue #53667, opened 2018-08-24
Tracking issue for eRFC 2497, "if- and while-let-chains, take 2"


Centril:
This is a tracking issue for the eRFC *"if- and while-let-chains, take 2"* (rust-lang/rfcs#2497).
For the tracking issue for the immediate edition changes, see #53668.

**Steps:**

- [x] Implement the RFC (cc @rust-lang/compiler -- can anyone write up mentoring instructions?)
- [ ] Adjust documentation ([see instructions on forge][doc-guide]) 
- [ ] Formatting for new syntax has been added to the [Style Guide] ([nightly-style-procedure])
- [ ] resolve expr fragment specifier issue ([#86730](https://github.com/rust-lang/rust/issues/86730))
- [x] Initial stabilization PR: #94927, reverted in #100538
- [ ] Second stabilization PR: https://github.com/rust-lang/rust/pull/132833

[stabilization-guide]: https://forge.rust-lang.org/stabilization-guide.html
[doc-guide]: https://forge.rust-lang.org/stabilization-guide.html#updating-documentation
[nightly-style-procedure]: https://github.com/rust-lang/style-team/blob/master/nightly-style-procedure.md 
[Style Guide]: https://github.com/rust-lang/rust/tree/master/src/doc/style-guide

**Unresolved questions:**

- [x] [The final syntax](https://github.com/rust-lang/rfcs/blob/master/text/2497-if-let-chains.md#the-final-syntax)
    - This would benefit from a design note summarizing previous discussions on `let` as a boolean expression, and possibilities of `is` syntax.
- [x] [Should temporary and irrefutable lets without patterns be allowed in some form?](https://github.com/rust-lang/rfcs/blob/master/text/2497-if-let-chains.md#irrefutable-let-bindings-after-the-first-refutable-binding)
- [x] [Chained `if let`s inside `match` arms](https://github.com/rust-lang/rfcs/blob/master/text/2497-if-let-chains.md#chained-if-lets-inside-match-arms)

**Collected issues:**

- [x] https://github.com/rust-lang/rust/issues/60336
- [x] https://github.com/rust-lang/rust/issues/60707
- [x] https://github.com/rust-lang/rust/issues/60254
- [x] https://github.com/rust-lang/rust/issues/60571
- [x] https://github.com/rust-lang/rust/issues/44990

**Implementation history:**

- https://github.com/rust-lang/rust/pull/59290
- https://github.com/rust-lang/rust/pull/59439
[...]

PR #87688, opened 2021-08-02
Introduce `let...else`

camsteffen:
Tracking issue: #87335

The trickiest part for me was enforcing the diverging else block with clear diagnostics. Perhaps the obvious solution is to expand to `let _: ! = ..`, but I decided against this because, when a "mismatched type" error is found in typeck, there is no way to trace where in the HIR the expected type originated, AFAICT. In order to pass down this information, I believe we should introduce `Expectation::LetElseNever(HirId)` or maybe add `HirId` to `Expectation::HasType`, but I left that as a future enhancement. For now, I simply assert that the block is `!` with a custom `ObligationCauseCode`, and I think this is clear enough, at least to start. The downside here is that the error points at the entire block rather than the specific expression with the wrong type. I left a todo to this effect.

Overall, I believe this PR is feature-complete with regard to the RFC.

### DIFF
--- a/compiler/rustc_ast/src/ast.rs
+++ b/compiler/rustc_ast/src/ast.rs
@@ -1005,11 +1005,40 @@ pub struct Local {
     pub id: NodeId,
     pub pat: P<Pat>,
     pub ty: Option<P<Ty>>,
-    /// Initializer expression to set the value, if any.
-    pub init: Option<P<Expr>>,
+    pub kind: LocalKind,
     pub span: Span,
     pub attrs: AttrVec,
     pub tokens: Option<LazyTokenStream>,
+}
+
+#[derive(Clone, Encodable, Decodable, Debug)]
+pub enum LocalKind {
+    /// Local declaration.
+    /// Example: `let x;`
+    Decl,
+    /// Local declaration with an initializer.
+    /// Example: `let x = y;`
+    Init(P<Expr>),
+    /// Local declaration with an initializer and an `else` clause.
+    /// Example: `let Some(x) = y else { return };`
+    InitElse(P<Expr>, P<Block>),
+}
+
+impl LocalKind {
+    pub fn init(&self) -> Option<&Expr> {
+        match self {
+            Self::Decl => None,
+            Self::Init(i) | Self::InitElse(i, _) => Some(i),
+        }
+    }
+
+    pub fn init_else_opt(&self) -> Option<(&Expr, Option<&Block>)> {
+        match self {
+            Self::Decl => None,
+            Self::Init(init) => Some((init, None)),
+            Self::InitElse(init, els) => Some((init, Some(els))),
+        }
+    }
 }
 
 /// An arm of a 'match'.
--- a/compiler/rustc_ast_lowering/src/block.rs
+++ b/compiler/rustc_ast_lowering/src/block.rs
@@ -1,6 +1,7 @@
 use crate::{ImplTraitContext, ImplTraitPosition, LoweringContext};
 use rustc_ast::{AttrVec, Block, BlockCheckMode, Expr, Local, LocalKind, Stmt, StmtKind};
 use rustc_hir as hir;
+use rustc_session::parse::feature_err;
 use rustc_span::symbol::Ident;
 use rustc_span::{sym, DesugaringKind};
 
@@ -170,6 +171,15 @@ impl<'a, 'hir> LoweringContext<'a, 'hir> {
             span,
             kind: hir::ExprKind::If(let_expr, then_expr, Some(else_expr)),
         });
+        if !self.sess.features_untracked().let_else {
+            feature_err(
+                &self.sess.parse_sess,
+                sym::let_else,
+                local.span,
+                "`let...else` statements are unstable",
+            )
+            .emit();
+        }
         (stmt, if_expr)
     }
 }
--- a/compiler/rustc_ast_lowering/src/block.rs
+++ b/compiler/rustc_ast_lowering/src/block.rs
@@ -2,7 +2,7 @@ use crate::{ImplTraitContext, ImplTraitPosition, LoweringContext};
 use rustc_ast::{Block, BlockCheckMode, Local, Stmt, StmtKind};
 use rustc_hir as hir;
 
-use smallvec::{smallvec, SmallVec};
+use smallvec::SmallVec;
 
 impl<'a, 'hir> LoweringContext<'a, 'hir> {
     pub(super) fn lower_block(
@@ -18,66 +18,67 @@ impl<'a, 'hir> LoweringContext<'a, 'hir> {
         b: &Block,
         targeted_by_break: bool,
     ) -> hir::Block<'hir> {
-        let (stmts, expr) = match &*b.stmts {
-            [stmts @ .., Stmt { kind: StmtKind::Expr(e), .. }] => (stmts, Some(&*e)),
-            stmts => (stmts, None),
-        };
-        let stmts = self.arena.alloc_from_iter(stmts.iter().flat_map(|stmt| self.lower_stmt(stmt)));
-        let expr = expr.map(|e| self.lower_expr(e));
+        let (stmts, expr) = self.lower_stmts(&b.stmts);
         let rules = self.lower_block_check_mode(&b.rules);
         let hir_id = self.lower_node_id(b.id);
-
         hir::Block { hir_id, stmts, expr, rules, span: self.lower_span(b.span), targeted_by_break }
     }
 
-    fn lower_stmt(&mut self, s: &Stmt) -> SmallVec<[hir::Stmt<'hir>; 1]> {
-        let (hir_id, kind) = match s.kind {
-            StmtKind::Local(ref l) => {
-                let l = self.lower_local(l);
-                let hir_id = self.lower_node_id(s.id);
-                self.alias_attrs(hir_id, l.hir_id);
-                return smallvec![hir::Stmt {
-                    hir_id,
-                    kind: hir::StmtKind::Local(self.arena.alloc(l)),
-                    span: self.lower_span(s.span),
-                }];
+    fn lower_stmts(
+        &mut self,
+        mut ast_stmts: &[Stmt],
+    ) -> (&'hir [hir::Stmt<'hir>], Option<&'hir hir::Expr<'hir>>) {
+        let mut stmts = SmallVec::<[hir::Stmt<'hir>; 8]>::new();
+        let mut expr = None;
+        while let [s, tail @ ..] = ast_stmts {
+            match s.kind {
+                StmtKind::Local(ref l) => {
+                    let l = self.lower_local(l);
+                    let hir_id = self.lower_node_id(s.id);
+                    self.alias_attrs(hir_id, l.hir_id);
+                    let kind = hir::StmtKind::Local(self.arena.alloc(l));
+                    let span = self.lower_span(s.span);
+                    stmts.push(hir::Stmt { hir_id, kind, span });
+                }
+                StmtKind::Item(ref it) => {
+                    stmts.extend(self.lower_item_id(it).into_iter().enumerate().map(
+                        |(i, item_id)| {
+                            let hir_id = match i {
+                                0 => self.lower_node_id(s.id),
+                                _ => self.next_id(),
+                            };
+                            let kind = hir::StmtKind::Item(item_id);
+                            let span = self.lower_span(s.span);
+                            hir::Stmt { hir_id, kind, span }
+                        },
+                    ));
+                }
+                StmtKind::Expr(ref e) => {
+                    let e = self.lower_expr(e);
+                    if tail.is_empty() {
+                        expr = Some(e);
+                    } else {
+                        let hir_id = self.lower_node_id(s.id);
+                        self.alias_attrs(hir_id, e.hir_id);
+                        let kind = hir::StmtKind::Expr(e);
+                        let span = self.lower_span(s.span);
+                        stmts.push(hir::Stmt { hir_id, kind, span });
+                    }
+                }
+                StmtKind::Semi(ref e) => {
+                    let e = self.lower_expr(e);
+                    let hir_id = self.lower_node_id(s.id);
+                    self.alias_attrs(hir_id, e.hir_id);
+                    let kind = hir::StmtKind::Semi(e);
+                    let span = self.lower_span(s.span);
+                    stmts.push(hir::Stmt { hir_id, kind, span });
+                }
+                StmtKind::Empty => {}
+                StmtKind::MacCall(..) => panic!("shouldn't exist here"),
             }
-            StmtKind::Item(ref it) => {
-                // Can only use the ID once.
-                let mut id = Some(s.id);
-                return self
-                    .lower_item_id(it)
-                    .into_iter()
-                    .map(|item_id| {
-                        let hir_id = id
-                            .take()
-                            .map(|id| self.lower_node_id(id))
-                            .unwrap_or_else(|| self.next_id());
-
-                        hir::Stmt {
-                            hir_id,
-                            kind: hir::StmtKind::Item(item_id),
-                            span: self.lower_span(s.span),
-                        }
-                    })
-                    .collect();
-            }
-            StmtKind::Expr(ref e) => {
-                let e = self.lower_expr(e);
-                let hir_id = self.lower_node_id(s.id);
-                self.alias_attrs(hir_id, e.hir_id);
-                (hir_id, hir::StmtKind::Expr(e))
-            }
-            StmtKind::Semi(ref e) => {
-                let e = self.lower_expr(e);
-                let hir_id = self.lower_node_id(s.id);
-                self.alias_attrs(hir_id, e.hir_id);
-                (hir_id, hir::StmtKind::Semi(e))
-            }
-            StmtKind::Empty => return smallvec![],
-            StmtKind::MacCall(..) => panic!("shouldn't exist here"),
-        };
-        smallvec![hir::Stmt { hir_id, kind, span: self.lower_span(s.span) }]
+            ast_stmts = &ast_stmts[1..];
+        }
+        (self.arena.alloc_from_iter(stmts), expr)
     }
 
     fn lower_local(&mut self, l: &Local) -> hir::Local<'hir> {
--- a/compiler/rustc_ast_lowering/src/block.rs
+++ b/compiler/rustc_ast_lowering/src/block.rs
@@ -1,6 +1,8 @@
 use crate::{ImplTraitContext, ImplTraitPosition, LoweringContext};
-use rustc_ast::{Block, BlockCheckMode, Local, Stmt, StmtKind};
+use rustc_ast::{AttrVec, Block, BlockCheckMode, Expr, Local, LocalKind, Stmt, StmtKind};
 use rustc_hir as hir;
+use rustc_span::symbol::Ident;
+use rustc_span::{sym, DesugaringKind};
 
 use smallvec::SmallVec;
 
@@ -32,13 +34,24 @@ impl<'a, 'hir> LoweringContext<'a, 'hir> {
         let mut expr = None;
         while let [s, tail @ ..] = ast_stmts {
             match s.kind {
-                StmtKind::Local(ref l) => {
-                    let l = self.lower_local(l);
+                StmtKind::Local(ref local) => {
                     let hir_id = self.lower_node_id(s.id);
-                    self.alias_attrs(hir_id, l.hir_id);
-                    let kind = hir::StmtKind::Local(self.arena.alloc(l));
-                    let span = self.lower_span(s.span);
-                    stmts.push(hir::Stmt { hir_id, kind, span });
+                    match &local.kind {
+                        LocalKind::InitElse(init, els) => {
+                            let (s, e) = self.lower_let_else(hir_id, local, init, els, tail);
+                            stmts.push(s);
+                            expr = Some(e);
+                            // remaining statements are in let-else expression
+                            break;
+                        }
+                        _ => {
+                            let local = self.lower_local(local);
+                            self.alias_attrs(hir_id, local.hir_id);
+                            let kind = hir::StmtKind::Local(local);
+                            let span = self.lower_span(s.span);
+                            stmts.push(hir::Stmt { hir_id, kind, span });
+                        }
+                    }
                 }
                 StmtKind::Item(ref it) => {
                     stmts.extend(self.lower_item_id(it).into_iter().enumerate().map(
@@ -81,22 +94,18 @@ impl<'a, 'hir> LoweringContext<'a, 'hir> {
         (self.arena.alloc_from_iter(stmts), expr)
     }
 
-    fn lower_local(&mut self, l: &Local) -> hir::Local<'hir> {
+    fn lower_local(&mut self, l: &Local) -> &'hir hir::Local<'hir> {
         let ty = l
             .ty
             .as_ref()
             .map(|t| self.lower_ty(t, ImplTraitContext::Disallowed(ImplTraitPosition::Binding)));
         let init = l.kind.init().map(|init| self.lower_expr(init));
         let hir_id = self.lower_node_id(l.id);
+        let pat = self.lower_pat(&l.pat);
+        let span = self.lower_span(l.span);
+        let source = hir::LocalSource::Normal;
         self.lower_attrs(hir_id, &l.attrs);
-        hir::Local {
-            hir_id,
-            ty,
-            pat: self.lower_pat(&l.pat),
-            init,
-            span: self.lower_span(l.span),
-            source: hir::LocalSource::Normal,
-        }
+        self.arena.alloc(hir::Local { hir_id, ty, pat, init, span, source })
     }
 
     fn lower_block_check_mode(&mut self, b: &BlockCheckMode) -> hir::BlockCheckMode {
@@ -106,5 +115,61 @@ impl<'a, 'hir> LoweringContext<'a, 'hir> {
                 hir::BlockCheckMode::UnsafeBlock(self.lower_unsafe_source(u))
             }
         }
+    }
+
+    fn lower_let_else(
+        &mut self,
+        stmt_hir_id: hir::HirId,
+        local: &Local,
+        init: &Expr,
+        els: &Block,
+        tail: &[Stmt],
+    ) -> (hir::Stmt<'hir>, &'hir hir::Expr<'hir>) {
+        let ty = local
+            .ty
+            .as_ref()
+            .map(|t| self.lower_ty(t, ImplTraitContext::Disallowed(ImplTraitPosition::Binding)));
+        let span = self.lower_span(local.span);
+        let span = self.mark_span_with_reason(DesugaringKind::LetElse, span, None);
+        let init = Some(self.lower_expr(init));
+        let val = Ident::with_dummy_span(sym::val);
+        let (pat, val_id) =
+            self.pat_ident_binding_mode(span, val, hir::BindingAnnotation::Unannotated);
+        let local_hir_id = self.lower_node_id(local.id);
+        self.lower_attrs(local_hir_id, &local.attrs);
+        // first statement which basically exists for the type annotation
+        let stmt = {
+            let local = self.arena.alloc(hir::Local {
+                hir_id: local_hir_id,
+                ty,
+                pat,
+                init,
+                span,
+                source: hir::LocalSource::Normal,
+            });
+            let kind = hir::StmtKind::Local(local);
+            hir::Stmt { hir_id: stmt_hir_id, kind, span }
+        };
+        let let_expr = {
+            let scrutinee = self.expr_ident(span, val, val_id);
+            let let_kind = hir::ExprKind::Let(self.lower_pat(&local.pat), scrutinee, span);
+            self.arena.alloc(self.expr(span, let_kind, AttrVec::new()))
+        };
+        let then_expr = {
+            let (stmts, expr) = self.lower_stmts(tail);
+            let block = self.block_all(span, stmts, expr);
+            self.arena.alloc(self.expr_block(block, AttrVec::new()))
+        };
+        let else_expr = {
+            let block = self.lower_block(els, false);
+            self.arena.alloc(self.expr_block(block, AttrVec::new()))
+        };
+        self.alias_attrs(else_expr.hir_id, local_hir_id);
+        let if_expr = self.arena.alloc(hir::Expr {
+            hir_id: self.next_id(),
+            span,
+            kind: hir::ExprKind::If(let_expr, then_expr, Some(else_expr)),
+        });
+        (stmt, if_expr)
     }
 }
--- a/compiler/rustc_ast_lowering/src/block.rs
+++ b/compiler/rustc_ast_lowering/src/block.rs
@@ -1,0 +1,109 @@
+use crate::{ImplTraitContext, ImplTraitPosition, LoweringContext};
+use rustc_ast::{Block, BlockCheckMode, Local, Stmt, StmtKind};
+use rustc_hir as hir;
+
+use smallvec::{smallvec, SmallVec};
+
+impl<'a, 'hir> LoweringContext<'a, 'hir> {
+    pub(super) fn lower_block(
+        &mut self,
+        b: &Block,
+        targeted_by_break: bool,
+    ) -> &'hir hir::Block<'hir> {
+        self.arena.alloc(self.lower_block_noalloc(b, targeted_by_break))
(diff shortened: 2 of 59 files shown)


src/tools/rustfmt/src/items.rs
```diff
@@ -106,11 +106,12 @@ impl Rewrite for ast::Local {
...
         result.push_str(&infix);
 
-        if let Some(ref ex) = self.init {
+        if let Some((init, _els)) = self.kind.init_else_opt() {
             // 1 = trailing semicolon;
             let nested_shape = shape.sub_width(1)?;
 
-            result = rewrite_assign_rhs(context, result, &**ex, nested_shape)?;
+            result = rewrite_assign_rhs(context, result, init, nested_shape)?;
+            // todo else
```
calebcartwright, line 114: This looks like rustfmt would butcher the let-else statements pretty badly. The typical first pass for new syntax with rustfmt often just involves spitting back out the unformatted/unprocessed contents of the original span, and I'd much prefer doing that here.

IIRC the span for locals doesn't include attributes so I'd suggest extending the condition above that checks for the skip attribute to also exit early if the local kind is InitElse. That should result in rustfmt emitting the statement and any attrs.

Don't feel obligated to do this, but if it helps you could add a snippet somewhere under `rustfmt/tests/target` either in a new file or an existing file that seems relevant, just to make sure that rustfmt emit invalid code that won't compile.

compiler/rustc_ast_lowering/src/local.rs
```diff
@@ -0,0 +1,68 @@
...
+use rustc_data_structures::fx::FxIndexMap;
+use rustc_data_structures::thin_vec::ThinVec;
+use rustc_hir as hir;
+use rustc_session::parse::feature_err;
+use rustc_span::symbol::Ident;
+use rustc_span::{sym, DUMMY_SP};
+
+impl<'hir> LoweringContext<'_, 'hir> {
+    /// Before: `let Some([ref x, ref y]) = z else { return }`
+    /// After: `let (ref x, ref y) = match z { Some([x, y]) => (x, y), _ => return }`
```
cjgillot, line 13: I am not convinced this is the best desugaring for this feature. I would rather see something like:
Before: `let Some([ref x, ref y]) = z else { return }; fun(x, y)`
After: `match z { Some([ref x, ref y]) => { fun(x, y) }, _ => { return }}`.
This would avoid the need for `let_else_bindings` and would avoid duplicating the bindings three times.

compiler/rustc_ast_lowering/src/local.rs
```diff
@@ -0,0 +1,68 @@
...
+    pub(super) fn lower_local_let_else(
+        &mut self,
+        l: &Local,
+        scrutinee: &Expr,
+        els: &Block,
+    ) -> (&'hir hir::Pat<'hir>, &'hir hir::Expr<'hir>) {
+        let (then_pat, bindings) = {
+            self.let_else_bindings = Some(FxIndexMap::default());
```
cjgillot, line 21: You need to save `let_else_bindinds` and to restore it later.

compiler/rustc_ast_lowering/src/lib.rs
```diff
@@ -167,6 +168,11 @@ struct LoweringContext<'a, 'hir: 'a> {
 
     allow_try_trait: Option<Lrc<[Symbol]>>,
     allow_gen_future: Option<Lrc<[Symbol]>>,
+
+    /// This is `Some` while lowering the Pat of a `let-else`. Any bindings in the pattern are given
+    /// a new `HirId`. The lowered `HirId` of the binding is added here and then used to create
+    /// bindings in the outer `let` statement of the desugared `let-else`.
+    let_else_bindings: Option<FxIndexMap<hir::HirId, (hir::BindingAnnotation, hir::HirId, Ident)>>,
```
cjgillot, line 175: I am not really fond of having an extra global variable during lowering. Is there an easy way to do it without?

compiler/rustc_ast/src/ast.rs
```diff
@@ -1005,13 +1005,42 @@ pub struct Local {
...
+    /// Local declaration.
+    /// Example: `let x;`
+    Decl,
+    /// Local declaration with an initializer.
+    /// Example: `let x = y;`
+    Init(P<Expr>),
+    /// Local declaration with an initializer and an `else` clause.
+    /// Example: `let Some(x) = y else { return };`
+    InitElse(P<Expr>, P<Block>),
+}
```
cjgillot, line 1025: Why introduce this enum instead of just having an extra `else_block: Option<P<Block>>` in `Local`?

compiler/rustc_ast_lowering/src/local.rs
```diff
@@ -0,0 +1,68 @@
+use super::LoweringContext;
```
cjgillot, line 1: A `local.rs` file seems a bit too restrictive. Should this be a `stmt.rs` file with all code lowering stmt/block/local?

cjgillot (changes requested):
Thanks @camsteffen for this implementation! I left a few comments on the lowering part: I am not convinced that you chose the simplest way to express this in HIR. I am interested in your reasonning.

compiler/rustc_ast_lowering/src/local.rs
camsteffen: The bindings need to be in scope for later statements. I don't think that is possible with your example.

camsteffen: Wait...are you suggesting to put the remainder of the block inside the `match`?

compiler/rustc_ast_lowering/src/local.rs
camsteffen: How about `block.rs` with block, stmt, local?

compiler/rustc_ast/src/ast.rs
camsteffen: This structure guarantees that locals with an else block also have an initializer.

compiler/rustc_ast_lowering/src/lib.rs
camsteffen: I don't think so. I would have to pass this structure to every `lower_pat_*` method.

compiler/rustc_ast_lowering/src/local.rs
camsteffen: Hmm I thought I was guaranteed not to have any locals within patterns, and so this would not be hit recursively? If that is not the case, maybe `let_else_bindings` needs to be a stack instead of an `Option`.

src/tools/rustfmt/src/items.rs
camsteffen: Sounds good. I will replace the todo with the following and add a quick test, pending other comments.

```rust
if let Some(els) = els {
    result.push_str(" else ");
    result.push_str(context.snippet(els.span));
}
```

compiler/rustc_ast_lowering/src/local.rs
camsteffen: Actually I can just do what you suggested and be safe in any case.

src/tools/rustfmt/src/items.rs
calebcartwright: No, I don't want to go that route. The formatting rules for the new syntax haven't been defined yet, and we don't want to be doing partial rewriting against an undefined spec.

We need to leave the original formatting in place for now. We'll update things once the corresponding formatting rules are codified and we've accounted for comments as well.

Extending the early exit check up above that starts on line 51 to include an or check for the corresponding localkind is what I'd want for the rustfmt piece, and which would be compatible with the existing changes/not require anything additional

```rust
        if contains_skip(&self.attrs) {
            return None;
        }
```

cjgillot:
Also, how does this feature articulate with the refactoring in #80357 ?

compiler/rustc_ast_lowering/src/local.rs
cjgillot: Yes :) 
However, this suggestion will require to be careful with lint attributes on the let-else statement.

camsteffen: Hmm right because the attribute should only apply to the then-pattern and the else-block. Should I duplicate the attributes? Or put some workaround in the visitor?

src/tools/rustfmt/src/items.rs
camsteffen: Sorry 🤦 I didn't read your first comment closely. I'm too excited and need to slow down.

camsteffen:
> Also, how does this feature articulate with the refactoring in #80357 ?

It don't really see any conflict, surprisingly, because that PR touches every `let`-containing syntax *except for* `let` statements.  @c410-f3r do you agree with that statement?

compiler/rustc_ast_lowering/src/local.rs
camsteffen: I think patterns do not have attributes so I would just put them on the else-block.

I found another problem - where would the type annotation go? Is `hir::ExprKind::Type` on the initializer an option, or no because type ascription is unstable? Or I suppose I could desugar into two statements? `let x: ty = ..; match x { .. }`

camsteffen: > Is `hir::ExprKind::Type` on the initializer an option?

No because it would not allow coercion.

c410-f3r:
> > Also, how does this feature articulate with the refactoring in #80357 ?
> 
> It don't really see any conflict, surprisingly, because that PR touches every `let`-containing syntax _except for_ `let` statements. @c410-f3r do you agree with that statement?

https://github.com/rust-lang/rust/pull/80357 touches more than +2k lines of code and I have doubts about no potential conflicts. My free-time is mostly limited to weekends so a confirmation, if it is worth something, will have to wait.

Its been 8 months of endless rebases and ppl are waiting years for https://github.com/rust-lang/rust/issues/53667. Personally, one more possible conflicting PR or one less possible conflicting PR doesn't make a difference any more

c410-f3r:
After cherry-picking, it looked like that most conflicts were more related to missing `MatchSource` variants than logic stuff although I am not completely sure.

But as stated before with all personal intrinsic reasons, don't mind #53667.

compiler/rustc_parse/src/parser/stmt.rs
```diff
@@ -292,8 +293,65 @@ impl<'a> Parser<'a> {
                 return Err(err);
             }
         };
+        let kind = match init {
+            None => LocalKind::Decl,
+            Some(init) => {
+                if self.eat_keyword(kw::Else) {
```
Fishrock123, line 299: This seems surprisingly simple! Nice.

src/test/ui/let-else/let-else-brace-before-else.rs
```diff
@@ -0,0 +1,22 @@
+// run-rustfix
+
+#![feature(let_else)]
+
+fn main() {
+    let Some(1) = loop { break Some(1) } else {
```
Fishrock123, line 6: I would also like to see a case with a raw block i.e.

```rust
    let Some(x) = { Some(1) } else {
        //~^ ERROR right curly brace `}` before `else` in a `let...else` statement not allowed
        return;
    };
```

compiler/rustc_infer/src/infer/error_reporting/mod.rs
```diff
@@ -792,6 +792,10 @@ impl<'a, 'tcx> InferCtxt<'a, 'tcx> {
                     );
                 }
             }
+            ObligationCauseCode::LetElse => {
+                err.help("try adding a `return` or `panic!(..)`");
```
Fishrock123, line 796:
```suggestion
                err.help("try adding a diverging expression, such as `return` or `panic!(..)`");
```

May be more clear?

compiler/rustc_infer/src/infer/error_reporting/mod.rs
```diff
@@ -792,6 +792,10 @@ impl<'a, 'tcx> InferCtxt<'a, 'tcx> {
...
+                err.help("...or use `match` instead of `let...else`");
```
Fishrock123, line 797:
```suggestion
                err.help("...or use either `match` or `if...let` instead of `let...else`");
```

I'd guess that in a single match non-diverging case, `if-let` is probably what the person wants.

camsteffen: That is possible. I think this is a "less is more" scenario. I decided on just `match` here because `if let` does not allow you to create bindings in scope for later statements. For a beginner, one suggestion that will always work is more helpful IMO.

compiler/rustc_infer/src/infer/error_reporting/mod.rs
camsteffen: Yes, I think that will be more helpful if the user doesn't know what "diverging" means. 👍

compiler/rustc_ast_lowering/src/block.rs
```diff
@@ -0,0 +1,182 @@
...
+    }
+
+    fn lower_let_else(
+        &mut self,
+        stmt_hir_id: hir::HirId,
+        l: &Local,
+        tail: &[Stmt],
+    ) -> Option<(hir::Stmt<'hir>, &'hir hir::Expr<'hir>)> {
+        // let-else would be nice right here, wouldn't you know it?
+        let (init, els) = match &l.kind {
```
cjgillot, line 124: This match should probably be in the caller function.

camsteffen:
@cjgillot Do you think this is ready to merge after addressing the comments (minor changes) or are you still reviewing?

cjgillot:
@camsteffen This will be ready to merge once the comments have been addressed.

camsteffen:
Is now based on #88187.

After rebasing over #80357, it seemed right to change the desugared form (again) to use the new HIR form of `if let`. So now the desugared form is `let value: type = init; if let pat = value { trailing_stmts } else { diverging_block }`.

compiler/rustc_parse/src/parser/stmt.rs
```diff
@@ -292,8 +293,65 @@ impl<'a> Parser<'a> {
...
+    fn check_let_else_init_trailing_brace(&self, init: &ast::Expr) {
+        if let Some(trailing) = classify::expr_trailing_brace(init) {
+            let err_span = trailing.span.with_lo(trailing.span.hi() - BytePos(1));
+            let suggs = vec![
+                (trailing.span.shrink_to_lo(), "(".to_string()),
+                (trailing.span.shrink_to_hi(), ")".to_string()),
+            ];
+            self.struct_span_err(
+                err_span,
+                "right curly brace `}` before `else` in a `let...else` statement not allowed",
```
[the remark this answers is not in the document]
Fishrock123: As Cameron said, this is disallowed as described by the RFC - see this section: https://github.com/rust-lang/rfcs/blob/master/text/3137-let-else.md#reference-level-explanations

(note: the de-sugaring may be out of date.)

compiler/rustc_ast_lowering/src/block.rs
```diff
@@ -0,0 +1,184 @@
...
+        local: &Local,
+        init: &Expr,
+        els: &Block,
+        tail: &[Stmt],
+    ) -> (hir::Stmt<'hir>, &'hir hir::Expr<'hir>) {
+        let ty = local
+            .ty
+            .as_ref()
+            .map(|t| self.lower_ty(t, ImplTraitContext::Disallowed(ImplTraitPosition::Binding)));
+        let span = self.mark_span_with_reason(DesugaringKind::LetElse, local.span, None);
```
cjgillot, line 133: You need to call lower_span here.

pnkfelix:
During weekly performance triage, this was flagged as [causing small regressions](https://perf.rust-lang.org/compare.html?start=a3956106d12cebec91be0637759e29ab6908b4cd&end=c2a408840ad18f74280805535f0b7193528ff3df&stat=instructions:u) (approximately 1%) in instruction counts on several of the *-doc tests, found via human eye.

Might be worth double-checking to see if that slowdown was expected.

compiler/rustc_ast_lowering/src/local.rs
durka: Is this change to the desugaring documented/discussed anywhere? I am not complaining, but it actually goes beyond the desugaring proposed in the RFC in that it allows a `ref` binding to a non-copy value, as exercised in the run-pass test. So do we need to note this somewhere (besides the test) to indicate that it's an intended addition?

camsteffen: @durka IMO the desugaring is an implementation detail and therefore technically not within the scope of what is specified by the RFC. Though the desugaring in the RFC works as an illustration of the expected behavior of let-else. I don't see anything that prohibits `ref` in the pattern. In user code, I would expect a `ref` to be allowed in exactly the same manner as a normal `let` or `if let`.

Mark-Simulacrum:
I've so far failed to arrive at any conclusive evidence with this change -- going to keep investigating, though.

The regression seems real -- cachegrind etc do reproduce it -- but it traces to functions like decode read_str and such, which seems rather unhelpful. I'm working now on building a debuginfo-enabled compiler for both before/after to try and get a better sense of the causes here.

Mark-Simulacrum:
This is beginning to look fairly similar to https://github.com/rust-lang/rust/pull/88881. There's around 20-30k additional instructions executed (for await-call-tree-doc), mostly in Span decoding -- but Span decoding is a very complex piece of code.

I've tried a number of things to try and figure out what the underlying cause for this is, but my current belief is that we're just seeing the effects of different optimizer decisions -- no *actual* difference in behavior -- but it is both unclear why that is or what's causing it. I've looked at the disassembly before/after and LLVM IR (manually dumped out of rustc with bootstrap adjustments locally), but none of that really yielded any clear indicators of what we can do to fix this.

I'll post further updates as I continue my investigation...

Mark-Simulacrum:
It looks like the [Span decoding function](https://github.com/rust-lang/rust/blob/6ae8912a3e7d2c4c775024f58a7ba4b1aedc4073/compiler/rustc_metadata/src/rmeta/decoder.rs#L460) has worse codegen with this PR. As best as I can tell, this isn't readily "fixable" and seems not directly connected to this PR, but adjustments to the LocalKind *Encode* impl do seem to fix it (I can't explain why changes to encoding affect decoding). I can reproduce the regression locally with and without PGO, so it doesn't seem related to that either.

AFAICT, there are some pretty weird choices made by LLVM in the assembly for this function -- I've uploaded disassembly [here](https://gist.github.com/Mark-Simulacrum/63798c060ae70c28e780e9b7a543cc9e). For example, we can see this [diff](https://gist.github.com/Mark-Simulacrum/63798c060ae70c28e780e9b7a543cc9e#file-asm-diff-L28-L34) near the top of the function:

```diff
-                       mov     rsi, qword ptr [r13 + 8]
-                       mov     rdi, qword ptr [r13 + 16]
-                       cmp     rdi, rsi
+                       mov     r15, qword ptr [rbp + 8]
+                       mov     rdi, qword ptr [rbp + 16]
+                       mov     rsi, rdi
+                       sub     rsi, r15
```

If we ignore the register renaming, the meaningful difference is the cmp is replaced with a mov and sub. However, there's no real reason for LLVM to do that. rsi is not used again before being overwritten I think, so this is just wasting a mov for no real reason. It seems like the majority of the changes to this function are similar "weird" small losses, not anything terribly material. There's also some extra stack spilling (note that the stack size increases by 32 bytes).

Replacing the LocalKind Encode impl with empty stubs or otherwise adjusting it seems to fix the codegen here. The LocalKind Encode impl is not invoked by any typical compilation (tested via inserting a panic!). I'm fairly certain now that the indirect effects on Span decoding are due to shuffling of CGU partitions for the rustc_metadata crate

I instrumented the CGU partitioning (-Zprint-mono-items=lazy) on the minimal change which replaces the Encode impl for LocalKind with an empty stub, and locally it looks like the only two crate which have changed are rustc_interface and rustc_metadata.

For rustc_metadata, if we ignore changes in which CGU gets which mono item, the only additions are the extra emit_enum/emit_enum_variant, etc. along with the closures for those:

```diff
@@ -3895,0 +3896 @@
+MONO_ITEM fn <rmeta::encoder::EncodeContext as rustc_serialize::Encoder>::emit_enum::<[closure@rustc_ast::ast::_DERIVE_rustc_serialize_Encodable_E_FOR_LocalKind::<impl rustc_serialize::Encodable<rmeta::encoder::EncodeContext> for >
@@ -4311,0 +4313,3 @@
+MONO_ITEM fn <rmeta::encoder::EncodeContext as rustc_serialize::Encoder>::emit_enum_variant::<[closure@rustc_ast::ast::_DERIVE_rustc_serialize_Encodable_E_FOR_LocalKind::<impl rustc_serialize::Encodable<rmeta::encoder::EncodeConte>
+MONO_ITEM fn <rmeta::encoder::EncodeContext as rustc_serialize::Encoder>::emit_enum_variant::<[closure@rustc_ast::ast::_DERIVE_rustc_serialize_Encodable_E_FOR_LocalKind::<impl rustc_serialize::Encodable<rmeta::encoder::EncodeConte>
+MONO_ITEM fn <rmeta::encoder::EncodeContext as rustc_serialize::Encoder>::emit_enum_variant::<[closure@rustc_ast::ast::_DERIVE_rustc_serialize_Encodable_E_FOR_LocalKind::<impl rustc_serialize::Encodable<rmeta::encoder::EncodeConte>
@@ -5977,0 +5982,3 @@
+MONO_ITEM fn <rmeta::encoder::EncodeContext as rustc_serialize::Encoder>::emit_enum_variant_arg::<[closure@rustc_ast::ast::_DERIVE_rustc_serialize_Encodable_E_FOR_LocalKind::<impl rustc_serialize::Encodable<rmeta::encoder::EncodeC>
+MONO_ITEM fn <rmeta::encoder::EncodeContext as rustc_serialize::Encoder>::emit_enum_variant_arg::<[closure@rustc_ast::ast::_DERIVE_rustc_serialize_Encodable_E_FOR_LocalKind::<impl rustc_serialize::Encodable<rmeta::encoder::EncodeC>
+MONO_ITEM fn <rmeta::encoder::EncodeContext as rustc_serialize::Encoder>::emit_enum_variant_arg::<[closure@rustc_ast::ast::_DERIVE_rustc_serialize_Encodable_E_FOR_LocalKind::<impl rustc_serialize::Encodable<rmeta::encoder::EncodeC>
@@ -7817 +7823,0 @@
-MONO_ITEM fn <rustc_ast::LocalKind as rustc_serialize::Encodable<rmeta::encoder::EncodeContext>>::encode
@@ -19645,0 +19652,8 @@
+MONO_ITEM fn rustc_ast::ast::_DERIVE_rustc_serialize_Encodable_E_FOR_LocalKind::<impl rustc_serialize::Encodable<rmeta::encoder::EncodeContext> for rustc_ast::LocalKind>::encode
+MONO_ITEM fn rustc_ast::ast::_DERIVE_rustc_serialize_Encodable_E_FOR_LocalKind::<impl rustc_serialize::Encodable<rmeta::encoder::EncodeContext> for rustc_ast::LocalKind>::encode::{closure#0}
+MONO_ITEM fn rustc_ast::ast::_DERIVE_rustc_serialize_Encodable_E_FOR_LocalKind::<impl rustc_serialize::Encodable<rmeta::encoder::EncodeContext> for rustc_ast::LocalKind>::encode::{closure#0}::{closure#0}
+MONO_ITEM fn rustc_ast::ast::_DERIVE_rustc_serialize_Encodable_E_FOR_LocalKind::<impl rustc_serialize::Encodable<rmeta::encoder::EncodeContext> for rustc_ast::LocalKind>::encode::{closure#0}::{closure#1}
+MONO_ITEM fn rustc_ast::ast::_DERIVE_rustc_serialize_Encodable_E_FOR_LocalKind::<impl rustc_serialize::Encodable<rmeta::encoder::EncodeContext> for rustc_ast::LocalKind>::encode::{closure#0}::{closure#1}::{closure#0}
+MONO_ITEM fn rustc_ast::ast::_DERIVE_rustc_serialize_Encodable_E_FOR_LocalKind::<impl rustc_serialize::Encodable<rmeta::encoder::EncodeContext> for rustc_ast::LocalKind>::encode::{closure#0}::{closure#2}
+MONO_ITEM fn rustc_ast::ast::_DERIVE_rustc_serialize_Encodable_E_FOR_LocalKind::<impl rustc_serialize::Encodable<rmeta::encoder::EncodeContext> for rustc_ast::LocalKind>::encode::{closure#0}::{closure#2}::{closure#0}
+MONO_ITEM fn rustc_ast::ast::_DERIVE_rustc_serialize_Encodable_E_FOR_LocalKind::<impl rustc_serialize::Encodable<rmeta::encoder::EncodeContext> for rustc_ast::LocalKind>::encode::{closure#0}::{closure#2}::{closure#1}
```

This is pretty expected, since all of these are getting instantiated with rmeta::encoder::EncodeContext for the first time (it's defined in rustc_metadata). This all mostly comes about from a pretty deep callstack, rooted in attribute encoding (since those contain arbitrary expressions and paths, we get essentially all AST types serialize and deserialize impl's codegened).

The partitioning has changed for CGUs 0, 3, and 4.

```
   8622 after-with-empty-encode/rustc-metadata/rustc_metadata.f82b8cc0-cgu.0
   2824 after-with-empty-encode/rustc-metadata/rustc_metadata.f82b8cc0-cgu.3
   6021 after-with-empty-encode/rustc-metadata/rustc_metadata.f82b8cc0-cgu.4
   8634 after-mono/rustc-metadata/rustc_metadata.f82b8cc0-cgu.0
   4466 after-mono/rustc-metadata/rustc_metadata.f82b8cc0-cgu.3
   4377 after-mono/rustc-metadata/rustc_metadata.f82b8cc0-cgu.4
```

Generally it seems like CGU 3 is now much larger -- presumably around 1,000 items moved from being put into CGU 4 into CGU 3.

This included a move of `fn <rmeta::decoder::DecodeContext as rustc_serialize::Decoder>::read_enum_variant_arg::<rustc_span::Span, for<'r> fn(&'r mut rmeta::decoder::DecodeContext) -> st...` from CGU 4 to 3. I believe (though have no hard evidence for it) that this likely caused transitive effects making the codegen worse. Notably other bits of that impl didn't move, so probably inlining is suffering here. (The below is a grep for `rustc_span::Span as rustc_serialize::Decodable<.*>>::decode` in rustc_metadata CGUs).

```diff
--- empty-encode-span	2021-10-11 14:42:56.626182306 -0400
+++ full-encode-span	2021-10-11 14:42:36.389748828 -0400
@@ -2,9 +2,9 @@
 rustc-metadata/rustc_metadata.f82b8cc0-cgu.15:fn <rmeta::decoder::DecodeContext as rustc_serialize::Decoder>::read_enum_variant_arg::<rustc_span::Span, for<'r> fn(&'r mut rmeta::decoder::DecodeContext) -> std::result::Result<rustc_span::Span, <rmeta::decoder::DecodeContext as rustc_serialize::Decoder>::Error> {<rustc_span::Span as rustc_serialize::Decodable<rmeta::decoder::DecodeContext>>::decode}> Internal
 rustc-metadata/rustc_metadata.f82b8cc0-cgu.15:fn <rmeta::decoder::DecodeContext as rustc_serialize::Decoder>::read_struct_field::<rustc_span::Span, for<'r> fn(&'r mut rmeta::decoder::DecodeContext) -> std::result::Result<rustc_span::Span, <rmeta::decoder::DecodeContext as rustc_serialize::Decoder>::Error> {<rustc_span::Span as rustc_serialize::Decodable<rmeta::decoder::DecodeContext>>::decode}> Internal
 rustc-metadata/rustc_metadata.f82b8cc0-cgu.3:fn <for<'r> fn(&'r mut rmeta::decoder::DecodeContext) -> std::result::Result<rustc_span::Span, <rmeta::decoder::DecodeContext as rustc_serialize::Decoder>::Error> {<rustc_span::Span as rustc_serialize::Decodable<rmeta::decoder::DecodeContext>>::decode} as std::ops::FnOnce<(&mut rmeta::decoder::DecodeContext,)>>::call_once - shim(for<'r> fn(&'r mut rmeta::decoder::DecodeContext) -> std::result::Result<rustc_span::Span, <rmeta::decoder::DecodeContext as rustc_serialize::Decoder>::Error> {<rustc_span::Span as rustc_serialize::Decodable<rmeta::decoder::DecodeContext>>::decode}) Internal
+rustc-metadata/rustc_metadata.f82b8cc0-cgu.3:fn <rmeta::decoder::DecodeContext as rustc_serialize::Decoder>::read_enum_variant_arg::<rustc_span::Span, for<'r> fn(&'r mut rmeta::decoder::DecodeContext) -> std::result::Result<rustc_span::Span, <rmeta::decoder::DecodeContext as rustc_serialize::Decoder>::Error> {<rustc_span::Span as rustc_serialize::Decodable<rmeta::decoder::DecodeContext>>::decode}> Internal
 rustc-metadata/rustc_metadata.f82b8cc0-cgu.3:fn <rmeta::decoder::DecodeContext as rustc_serialize::Decoder>::read_struct_field::<rustc_span::Span, for<'r> fn(&'r mut rmeta::decoder::DecodeContext) -> std::result::Result<rustc_span::Span, <rmeta::decoder::DecodeContext as rustc_serialize::Decoder>::Error> {<rustc_span::Span as rustc_serialize::Decodable<rmeta::decoder::DecodeContext>>::decode}> Internal
 rustc-metadata/rustc_metadata.f82b8cc0-cgu.4:fn <for<'r> fn(&'r mut rmeta::decoder::DecodeContext) -> std::result::Result<rustc_span::Span, <rmeta::decoder::DecodeContext as rustc_serialize::Decoder>::Error> {<rustc_span::Span as rustc_serialize::Decodable<rmeta::decoder::DecodeContext>>::decode} as std::ops::FnOnce<(&mut rmeta::decoder::DecodeContext,)>>::call_once - shim(for<'r> fn(&'r mut rmeta::decoder::DecodeContext) -> std::result::Result<rustc_span::Span, <rmeta::decoder::DecodeContext as rustc_serialize::Decoder>::Error> {<rustc_span::Span as rustc_serialize::Decodable<rmeta::decoder::DecodeContext>>::decode}) Internal
-rustc-metadata/rustc_metadata.f82b8cc0-cgu.4:fn <rmeta::decoder::DecodeContext as rustc_serialize::Decoder>::read_enum_variant_arg::<rustc_span::Span, for<'r> fn(&'r mut rmeta::decoder::DecodeContext) -> std::result::Result<rustc_span::Span, <rmeta::decoder::DecodeContext as rustc_serialize::Decoder>::Error> {<rustc_span::Span as rustc_serialize::Decodable<rmeta::decoder::DecodeContext>>::decode}> Internal
 rustc-metadata/rustc_metadata.f82b8cc0-cgu.4:fn <rmeta::decoder::DecodeContext as rustc_serialize::Decoder>::read_struct_field::<rustc_span::Span, for<'r> fn(&'r mut rmeta::decoder::DecodeContext) -> std::result::Result<rustc_span::Span, <rmeta::decoder::DecodeContext as rustc_serialize::Decoder>::Error> {<rustc_span::Span as rustc_serialize::Decodable<rmeta::decoder::DecodeContext>>::decode}> Internal
 rustc-metadata/rustc_metadata.f82b8cc0-cgu.7:fn <for<'r> fn(&'r mut rmeta::decoder::DecodeContext) -> std::result::Result<rustc_span::Span, <rmeta::decoder::DecodeContext as rustc_serialize::Decoder>::Error> {<rustc_span::Span as rustc_serialize::Decodable<rmeta::decoder::DecodeContext>>::decode} as std::ops::FnOnce<(&mut rmeta::decoder::DecodeContext,)>>::call_once - shim(for<'r> fn(&'r mut rmeta::decoder::DecodeContext) -> std::result::Result<rustc_span::Span, <rmeta::decoder::DecodeContext as rustc_serialize::Decoder>::Error> {<rustc_span::Span as rustc_serialize::Decodable<rmeta::decoder::DecodeContext>>::decode}) Internal
 rustc-metadata/rustc_metadata.f82b8cc0-cgu.7:fn <rmeta::decoder::DecodeContext as rustc_serialize::Decoder>::read_struct_field::<rustc_span::Span, for<'r> fn(&'r mut rmeta::decoder::DecodeContext) -> std::result::Result<rustc_span::Span, <rmeta::decoder::DecodeContext as rustc_serialize::Decoder>::Error> {<rustc_span::Span as rustc_serialize::Decodable<rmeta::decoder::DecodeContext>>::decode}> Internal

```

Overall I think this means we probably can't really do much about the hit here -- it's just more CGU partitioning sometimes leading to suboptimal results. It's possible that adding some targeted #[inline] annotations or otherwise shifting around code might help with this, but generally that seems brittle and not particularly warranted. I'm going to mark this regression as triaged, because I don't think we can "fix" it at this time. Hopefully we'll eventually have a nicer partitioning algorithm that avoids this kind of trouble.

In the meantime, it's possible we can help by avoiding so much transitive codegen being necessary in rustc_metadata for all the serialize impls for rustc_ast and such. But that's a hard project and not one with obvious solutions, so I think we shouldn't block triaging this PR on it.

PR #85782, opened 2021-05-28
Build aarch64-apple-ios-sim as part of the full macOS build

badboy:
Part of the [MCP 428](https://github.com/rust-lang/compiler-team/issues/428) to promote this target to Tier 2.

This adds the aarch64-apple-ios-sim target as a tier 2 target, currently cross-compiled from our x86_64 apple builders. The compiler team has approved the addition per the MCP noted above, and the infrastructure team has not raised concerns with this addition at this time (as the CI time impact is expected to be minimal; this is only building std).

PR #87693, opened 2021-08-02
Add `aarch64-apple-ios-sim` as a possible target to the manifest

badboy:
This should allow rustup and similar to actually make use of this new
target now.

r? @Mark-Simulacrum

Followup to #85782


Mark-Simulacrum:
Good catch. @bors r+

PR #87695, opened 2021-08-02
Prepare GEP building for opaque pointers

tmiasko:


compiler/rustc_codegen_ssa/src/mir/place.rs
```diff
@@ -103,12 +103,13 @@ impl<'a, 'tcx, V: CodegenObject> PlaceRef<'tcx, V> {
                     if offset == a.value.size(bx.cx()).align_to(b.value.align(bx.cx()).abi) =>
                 {
                     // Offset matches second field.
-                    bx.struct_gep(self.llval, 1)
+                    let ty = bx.backend_type(self.layout);
+                    bx.struct_gep(ty, self.llval, 1)
                 }
                 Abi::Scalar(_) | Abi::ScalarPair(..) | Abi::Vector { .. } if field.is_zst() => {
                     // ZST fields are not included in Scalar, ScalarPair, and Vector layouts, so manually offset the pointer.
                     let byte_ptr = bx.pointercast(self.llval, bx.cx().type_i8p());
```
nagisa, line 111: Is it necessary to pointercast this `self.llval` anymore?

compiler/rustc_codegen_ssa/src/mir/place.rs
```diff
@@ -185,7 +189,7 @@ impl<'a, 'tcx, V: CodegenObject> PlaceRef<'tcx, V> {
 
         // Cast and adjust pointer.
         let byte_ptr = bx.pointercast(self.llval, bx.cx().type_i8p());
```
nagisa, line 191: Similarly here.

tmiasko:
* Rebased to resolve conflicts with #87515.

compiler/rustc_codegen_ssa/src/mir/place.rs
tmiasko: When an explicit pointee type is provided it must be equal to the pointer element type, so this will have to wait until migration to opaque pointers is complete:

https://github.com/rust-lang/llvm-project/blob/260e0f8682098faab68af9c608534756ad378365/llvm/include/llvm/IR/Instructions.h#L956-L958

PR #87699, opened 2021-08-02
Allow specifying an deployment target version for all iOS llvm targets

### LINKED ISSUE #79408: Rustc and XCFramework won't work
Hi

We are currently trying to build a xcframework, which includes a rust static library as a binary target. When we build the xcframework with `xcodebuild  -create-xcframework -library target/aarch64-apple-ios/release/libxcframework_test.a -headers test.h -output test.xcframework` we get `The CodingKeys(stringValue: "SupportedPlatform", intValue: nil) is empty in library -arm64.`. 

To reproduce the failure just build a static lib with a function in it e.g.:
```rust
#[no_mangle]
pub unsafe extern "C" fn test() {
}
``` 

` cargo build --release --target aarch64-apple-ios`

Interestingly enough, the binary works when statically linked during the usual build process in XCode (so the code itself seems to be correct).

Further, for the darwin binary (target macos) the xcframework creation process succeeds.

I think with rustc 1.43 the xcframework also worked for the ios platform, but other than that we have no idea what is wrong.

It is though certainly somehow linked to rust, as C/C++ libraries work (e.g. libsodium) with exact the same command/folder-structure.

[EDIT] We found that LC_VERSION_MIN_* is emitted instead of LC_BUILD_VERSION in the load commands of the MachO binary. Apparently this is the newly used command to specify the platform.

> As of iOS 12 (and I’m not sure which macOS version), the loader looks for LC_BUILD_VERSION instead of LC_VERSION_MIN*.
[1](https://jetforme.org/2020/02/lc_version_min/)

As a comparison: 
XCODE:
> Load command 1
       cmd LC_BUILD_VERSION
   cmdsize 24
  platform ios
       sdk 14.2
     minos 14.2
    ntools 0

RUST:
> Load command 1
      cmd LC_VERSION_MIN_IPHONEOS
  cmdsize 16
  version 7.0
      sdk 13.7

[EDIT 2]
Getting closer found  https://github.com/rust-lang/rust/issues/29664 and made a new target, where I set  the llvm-target to the one with the correct version like this:

```json
{
  "abi-return-struct-as-int": true,
  "arch": "aarch64",
  "archive-format": "darwin",
  "bitcode-llvm-cmdline": "-triple\u0000arm64-apple-ios11.0.0\u0000-emit-obj\u0000-disable-llvm-passes\u0000-target-abi\u0000darwinpcs\u0000-Os\u0000",
  "cpu": "apple-a7",
  "data-layout": "e-m:o-i64:64-i128:128-n32:64-S128",
  "dll-suffix": ".dylib",
  "dwarf-version": 2,
  "eh-frame-header": false,
  "eliminate-frame-pointer": false,
  "emit-debug-gdb-scripts": false,
  "executables": true,
  "features": "+neon,+fp-armv8,+apple-a7",
  "forces-embed-bitcode": true,
  "function-sections": false,
  "has-rpath": true,
  "is-builtin": true,
  "is-like-osx": true,
  "link-env": [
    "ZERO_AR_DATE=1"
  ],
  "link-env-remove": [
    "MACOSX_DEPLOYMENT_TARGET"
  ],
  "llvm-target": "arm64-apple-ios9.0",
  "max-atomic-width": 128,
  "os": "ios",
  "target-family": "unix",
  "target-pointer-width": "64",
  "unsupported-abis": [
    "stdcall",
    "fastcall",
    "vectorcall",
    "thiscall",
    "win64",
    "sysv64"
  ],
  "vendor": "apple"
}
```

Now I can create an xcframework again!

ubamrein:
Closes: https://github.com/rust-lang/rust/issues/79408

This pull requests adds the same procedure to define the iOS-version for the LLVM-target as was used for the simulator target and the desktop target.

This then closes the original problem mentioned in the above issue. The problem with incompatible bitcode remains, but is probably not easy fixable.

I realised that something is still not right. Try to fix that.

r? @petrochenkov

### DIFF
--- a/compiler/rustc_target/src/spec/apple_base.rs
+++ b/compiler/rustc_target/src/spec/apple_base.rs
@@ -92,6 +92,12 @@ fn ios_deployment_target() -> (u32, u32) {
 }
 
 pub fn ios_llvm_target(arch: &str) -> String {
+    // Modern iOS tooling extracts information about deployment target
+    // from LC_BUILD_VERSION. This load command will only be emitted when
+    // we build with a version specific `llvm_target`, with the version
+    // set high enough. Luckily one LC_BUILD_VERSION is enough, for Xcode
+    // to pick it up (since std and core are still built with the fallback
+    // of version 7.0 and hence emit the old LC_IPHONE_MIN_VERSION).
     let (major, minor) = ios_deployment_target();
     format!("{}-apple-ios{}.{}.0", arch, major, minor)
 }
--- a/compiler/rustc_target/src/spec/armv7_apple_ios.rs
+++ b/compiler/rustc_target/src/spec/armv7_apple_ios.rs
@@ -3,7 +3,7 @@ use crate::spec::{Target, TargetOptions};
 
 pub fn target() -> Target {
     Target {
-        llvm_target: "armv7-apple-ios".to_string(),
+        llvm_target: super::apple_base::ios_llvm_target("armv7"),
         pointer_width: 32,
         data_layout: "e-m:o-p:32:32-Fi8-f64:32:64-v64:32:64-v128:32:128-a:0:32-n32-S32".to_string(),
         arch: "arm".to_string(),
--- a/compiler/rustc_target/src/spec/i386_apple_ios.rs
+++ b/compiler/rustc_target/src/spec/i386_apple_ios.rs
@@ -4,7 +4,7 @@ use crate::spec::{StackProbeType, Target, TargetOptions};
 pub fn target() -> Target {
     let base = opts("ios", Arch::I386);
     Target {
-        llvm_target: "i386-apple-ios".to_string(),
+        llvm_target: super::apple_base::ios_sim_llvm_target("i386"),
         pointer_width: 32,
         data_layout: "e-m:o-p:32:32-p270:32:32-p271:32:32-p272:64:64-\
             f64:32:64-f80:128-n8:16:32-S128"
--- a/compiler/rustc_target/src/spec/x86_64_apple_ios.rs
+++ b/compiler/rustc_target/src/spec/x86_64_apple_ios.rs
@@ -4,7 +4,7 @@ use crate::spec::{StackProbeType, Target, TargetOptions};
 pub fn target() -> Target {
     let base = opts("ios", Arch::X86_64);
     Target {
-        llvm_target: "x86_64-apple-ios".to_string(),
+        llvm_target: super::apple_base::ios_sim_llvm_target("x86_64"),
         pointer_width: 64,
         data_layout: "e-m:o-p270:32:32-p271:32:32-p272:64:64-i64:64-f80:128-n8:16:32:64-S128"
             .to_string(),


kit-981:
It looks like I've unknowingly put up a [pull request](https://github.com/rust-lang/rust/pull/88077) with a similar change.

When trying to run a Rust binary on a jailbroken iPhone, the operating system refused to run it because the header did not specify a minimum supported operating system version. Adding the specific version to the target is a straightforward fix because LLVM will then populate the Mach-O header with that minimum supported operating system version.

Why are the changes in e6293a4408097d437fe78b17df9a37e0677c080e required?

kit-981:
> I realised that something is still not right. Try to fix that.

Are there still known issues with these changes?

ubamrein:
> > I realised that something is still not right. Try to fix that.
> 
> Are there still known issues with these changes?

Not really issues, but when I was testing it, the compiler did not use my env-variable (and hence was not using the correct llvm-target). I am not sure if it was the way I passed the env-var or another issue. If it is working for you, then I guess I just did something wrong :D I did not have time yet to look into it that's why I put the WIP.

For the other commit, I am not how I managed to do that (it is not my commit though, so I guess I managed to fail rebasing).... Either we close this PR and use your's, or I rebase the branch to the new tip or something like that :D 

@LeSeulArtichaut any inputs?

kit-981:
> Not really issues, but when I was testing it, the compiler did not use my env-variable (and hence was not using the correct llvm-target). I am not sure if it was the way I passed the env-var or another issue. If it is working for you, then I guess I just did something wrong :D I did not have time yet to look into it that's why I put the WIP.

Makes sense. Thanks for the explanation.

> For the other commit, I am not sure how I managed to do that (it is not my commit though, so I guess I managed to fail rebasing).... Either we close this PR and use your's, or I rebase the branch to the new tip or something like that :D

I don't mind.

I've only been testing the `aarch64-apple-ios` target. I'd offer to add the `armv7`, `i386`, and `x86_64` targets to my pull request but I don't have an easy way to test the changes and make sure that things behave as expected on these platforms.

Is it reasonable to keep `aarch64-apple-ios` in my pull request, merge it, and defer supporting these other architectures to this pull request? I ask because I want to avoid holding up the fix for `aarch64-apple-ios`.

ubamrein:
> Is it reasonable to keep `aarch64-apple-ios` in my pull request, merge it, and defer supporting these other architectures to this pull request? I ask because I want to avoid holding up the fix for `aarch64-apple-ios`.

Would be fine with me! Then I can test the other targets when I find time, and we can be sure to have working target specifications!

LeSeulArtichaut:
> @LeSeulArtichaut any inputs?

This is not something I can review. I'll assign this PR and #88077 to r? @petrochenkov for review or reassignment.

compiler/rustc_target/src/spec/aarch64_apple_ios.rs
```diff
@@ -2,8 +2,9 @@ use super::apple_sdk_base::{opts, Arch};
 use crate::spec::{FramePointer, Target, TargetOptions};
 
 pub fn target() -> Target {
+    let llvm_target = super::apple_base::ios_llvm_target("arm64");
     Target {
-        llvm_target: "arm64-apple-ios".to_string(),
+        llvm_target,
```
petrochenkov, line 7:
```suggestion
        llvm_target: super::apple_base::ios_llvm_target("arm64"),
```
Nit: no need to create a separate variable (here and in other files as well).

petrochenkov:
I don't know much about Apple targets, but this looks consistent with what we are doing for other targets.

@ubamrein 
Could you rebase this PR and address nits?

kit-981:
> I don't know much about Apple targets, but this looks consistent with what we are doing for other targets.
> 
> @ubamrein
> Could you rebase this PR and address nits?

@petrochenkov  Is there a reason why you are against merging in my changes for `aarch64-apple-ios` in isolation as @ubamrein and I were discussing? #88077 is tested, known to work, and was ready to merge. In contrast, these changes are not ready to merge and it's not clear how much testing has gone into other architectures to ensure that this doesn't introduce some kind of unexpected regression.

petrochenkov:
@kit-981 
No specific reasons, just because this PR does the same things and was submitted earlier.
If you have reservations about the parts of this PR related to other architectures, then I have no issues with merging https://github.com/rust-lang/rust/pull/88077 first.

ubamrein:
I would like to test some of the targets first, but did not yet find time to do it. 

The other PR was tested and is working, so I guess for the sake of having the target upstream for the most important target (namely the only currently supported iOS target, which is not a simulator), you could first merge the other one.

ubamrein:
So I now made some tests. Note that I changed the i386 and x86_64 targets to use a simulator target, since those are architectures only available on the simulator ABI.

Further, in my tests I saw that emitting one LC_BUILD_VERSION (or at least only the ones for user-code) is enough, for XCode to pick up the deployment target. 

Note though that the rust standard library and core are still built with the default llvm-target where the version is set to 7.0. Should Apple in the future decide to block such builds completely we might have to increase the default build version (of course in the mean time one could build standard with the new `build-std` flag, which I verified to only emit LC_BUILD_VERSION).

petrochenkov:
r=me with commits squashed.

petrochenkov:
Thanks!
@bors r+

PR #87712, opened 2021-08-03
Proc macro spans: make columns 1 based

est31:
This makes proc macro spans consistent with the `column!()` macro as well as `std::panic::Location`, as both are 1-based.

https://github.com/rust-lang/rust/issues/54725#issuecomment-497246753


est31:
Note that there are still small differences between proc macro span columns and panic location columns: the panic location counts a tab as 4 characters while the proc macro spans count it as 1 character. However, the `column!()` macro does this as well so I don't think it's a problem that needs solving.

petrochenkov:
This looks reasonable and the API is (perpetually?) unstable, so I'm going to just r+ this.
@bors r+

PR #87715, opened 2021-08-03
Add long error explanation for E0625

bhgomes:
For #61137.

### DIFF
--- a/compiler/rustc_error_codes/src/error_codes/E0625.md
+++ b/compiler/rustc_error_codes/src/error_codes/E0625.md
@@ -1,5 +1,4 @@
-Static and const variables can refer to other const variables. But a const
-variable cannot refer to a thread-local static variable.
+A compile-time const variable is referring to a thread-local static variable.
 
 Erroneous code example:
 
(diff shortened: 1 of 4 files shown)


petrochenkov:
>tidy error: /checkout/compiler/rustc_error_codes/src/error_codes/E0625.md: too many trailing newlines (2)

compiler/rustc_error_codes/src/error_codes/E0625.md
```diff
@@ -0,0 +1,27 @@
+Static and const variables can refer to other const variables. But a const
+variable cannot refer to a thread-local static variable.
```
GuillaumeGomez, line 2: Generally the first sentence is just saying something very short. In this case it would be something like "A const variable is referring to a thread-local static variable".

Then you provide the full explanation below.

GuillaumeGomez (changes requested):
Please move the explanation after the failing example but otherwise looks good to me.

GuillaumeGomez (approved):

GuillaumeGomez:
Thanks! r=me once CI pass

PR #87716, opened 2021-08-03
Allow generic SIMD array element type

calebzulawski:
Fixes the following:
```rust
#[repr(simd)]
struct V<T>([T; 4]);
```

cc @workingjubilee


workingjubilee:
This code looks correct to me, including the tests, based on my previous pass through compiler/rustc_typeck/src/check/check.rs.
@bors r+

PR #87728, opened 2021-08-03
Add "doc(test(...))" attribute checks

### LINKED ISSUE #82672: check that attributes passed to doc(test) are valid
Discussed in https://github.com/rust-lang/rust/pull/82662#discussion_r584871289

GuillaumeGomez:
Fixes #82672.

r? @camelid

### DIFF
--- a/compiler/rustc_passes/src/check_attr.rs
+++ b/compiler/rustc_passes/src/check_attr.rs
@@ -717,6 +717,42 @@ impl CheckAttrVisitor<'tcx> {
         true
     }
 
+    /// Checks that `doc(test(...))` attribute contains only valid attributes. Returns `true` if
+    /// valid.
+    fn check_test_attr(&self, meta: &NestedMetaItem, hir_id: HirId) -> bool {
+        let mut is_valid = true;
+        if let Some(metas) = meta.meta_item_list() {
+            for i_meta in metas {
+                match i_meta.name_or_empty() {
+                    sym::attr | sym::no_crate_inject => {}
+                    _ => {
+                        self.tcx.struct_span_lint_hir(
+                            INVALID_DOC_ATTRIBUTES,
+                            hir_id,
+                            i_meta.span(),
+                            |lint| {
+                                lint.build(&format!(
+                                    "unknown `doc(test)` attribute `{}`",
+                                    rustc_ast_pretty::pprust::path_to_string(
+                                        &i_meta.meta_item().unwrap().path
+                                    ),
+                                ))
+                                .emit();
+                            },
+                        );
+                        is_valid = false;
+                    }
+                }
+            }
+        } else {
+            self.tcx.struct_span_lint_hir(INVALID_DOC_ATTRIBUTES, hir_id, meta.span(), |lint| {
+                lint.build("`#[doc(test(...)]` takes a list of attributes").emit();
+            });
+            is_valid = false;
+        }
+        is_valid
+    }
+
     /// Runs various checks on `#[doc]` attributes. Returns `true` if valid.
     ///
     /// `specified_inline` should be initialized to `None` and kept for the scope
@@ -793,8 +829,13 @@ impl CheckAttrVisitor<'tcx> {
                         | sym::no_inline
                         | sym::notable_trait
                         | sym::passes
-                        | sym::plugins
-                        | sym::test => {}
+                        | sym::plugins => {}
+
+                        sym::test => {
+                            if !self.check_test_attr(&meta, hir_id) {
+                                is_valid = false;
+                            }
+                        }
 
                         sym::primitive => {
                             if !self.tcx.features().doc_primitive {
--- a/src/test/rustdoc-ui/doc-test-attr.rs
+++ b/src/test/rustdoc-ui/doc-test-attr.rs
@@ -10,7 +10,5 @@
 #![doc(test(a))]
 //~^ ERROR unknown `doc(test)` attribute `a`
 //~^^ WARN this was previously accepted by the compiler
-#![doc(test(no_crate_inject))]
-#![doc(test(attr(deny(warnings))))]
 
 pub fn foo() {}
--- a/src/test/rustdoc-ui/doc-test-attr.rs
+++ b/src/test/rustdoc-ui/doc-test-attr.rs
@@ -1,0 +1,16 @@
+#![crate_type = "lib"]
+#![deny(invalid_doc_attributes)]
+
+#![doc(test)]
+//~^ ERROR `#[doc(test(...)]` takes a list of attributes
+//~^^ WARN this was previously accepted by the compiler
+#![doc(test = "hello")]
+//~^ ERROR `#[doc(test(...)]` takes a list of attributes
+//~^^ WARN this was previously accepted by the compiler
+#![doc(test(a))]
+//~^ ERROR unknown `doc(test)` attribute `a`
+//~^^ WARN this was previously accepted by the compiler
+#![doc(test(no_crate_inject))]
+#![doc(test(attr(deny(warnings))))]
+
+pub fn foo() {}
--- a/src/test/ui/rustdoc/doc-test-attr.rs
+++ b/src/test/ui/rustdoc/doc-test-attr.rs
@@ -10,7 +10,5 @@
 #![doc(test(a))]
 //~^ ERROR unknown `doc(test)` attribute `a`
 //~^^ WARN this was previously accepted by the compiler
-#![doc(test(no_crate_inject))]
-#![doc(test(attr(deny(warnings))))]
 
 pub fn foo() {}
--- a/src/test/ui/rustdoc/doc-test-attr.rs
+++ b/src/test/ui/rustdoc/doc-test-attr.rs
@@ -1,0 +1,16 @@
+#![crate_type = "lib"]
+#![deny(invalid_doc_attributes)]
+
+#![doc(test)]
+//~^ ERROR `#[doc(test(...)]` takes a list of attributes
+//~^^ WARN this was previously accepted by the compiler
+#![doc(test = "hello")]
+//~^ ERROR `#[doc(test(...)]` takes a list of attributes
+//~^^ WARN this was previously accepted by the compiler
+#![doc(test(a))]
+//~^ ERROR unknown `doc(test)` attribute `a`
+//~^^ WARN this was previously accepted by the compiler
+#![doc(test(no_crate_inject))]
+#![doc(test(attr(deny(warnings))))]
+
+pub fn foo() {}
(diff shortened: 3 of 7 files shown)


compiler/rustc_passes/src/check_attr.rs
```diff
@@ -717,6 +717,39 @@ impl CheckAttrVisitor<'tcx> {
...
+                        is_valid = false;
+                    }
+                }
+            }
+        } else {
+            self.tcx
+                .sess
+                .struct_span_err(
+                    meta.span(),
+                    "`#[doc(test(...)]` is supposed to be a list attribute",
```
camelid, line 745: Might be clearer to say something like this:

```suggestion
                    "`#[doc(test(...)]` takes a list of attributes",
```

since people may not know what a list attribute is.

compiler/rustc_passes/src/check_attr.rs
```diff
@@ -717,6 +717,39 @@ impl CheckAttrVisitor<'tcx> {
...
+        if let Some(metas) = meta.meta_item_list() {
+            for i_meta in metas {
+                match i_meta.name_or_empty() {
+                    sym::attr | sym::no_crate_inject => {}
+                    name => {
+                        self.tcx
+                            .sess
+                            .struct_span_err(
+                                i_meta.span(),
+                                &format!("unknown `doc(test)` attribute `{}`", name,),
```
camelid, line 733: Could you change this to

```suggestion
                                &format!("unknown `doc(test)` attribute `{}`", rustc_ast_pretty::pprust::path_to_string(&i_meta.path)),
```

like other places where we emit the lint so that things like `doc(test(foo::bar))` don't show up as ```unknown `doc(test)` attribute `` ```?

compiler/rustc_passes/src/check_attr.rs
GuillaumeGomez: :+1:

compiler/rustc_passes/src/check_attr.rs
GuillaumeGomez: :+1:

compiler/rustc_passes/src/check_attr.rs
```diff
@@ -717,6 +717,41 @@ impl CheckAttrVisitor<'tcx> {
...
+    fn check_test_attr(&self, meta: &NestedMetaItem) -> bool {
+        let mut is_valid = true;
+        if let Some(metas) = meta.meta_item_list() {
+            for i_meta in metas {
+                match i_meta.name_or_empty() {
+                    sym::attr | sym::no_crate_inject => {}
+                    _ => {
+                        self.tcx
+                            .sess
+                            .struct_span_err(
```
jyn514, line 731: This is wrong. It needs to be a future-incompat lint, not a hard error. This is the exact issue from https://github.com/rust-lang/rust/pull/82702, please be more careful in the future.

compiler/rustc_passes/src/check_attr.rs
```diff
@@ -717,6 +717,41 @@ impl CheckAttrVisitor<'tcx> {
...
+                            )
+                            .emit();
+                        is_valid = false;
+                    }
+                }
+            }
+        } else {
+            self.tcx
+                .sess
+                .struct_span_err(meta.span(), "`#[doc(test(...)]` takes a list of attributes")
```
jyn514, line 748: This is wrong.

jyn514 (changes requested):

compiler/rustc_passes/src/check_attr.rs
GuillaumeGomez: Some day I'll remember about the future-incompat lint... Good catch!

compiler/rustc_passes/src/check_attr.rs
```diff
@@ -717,6 +717,53 @@ impl CheckAttrVisitor<'tcx> {
...
+                        is_valid = false;
+                    }
+                }
+            }
+        } else {
+            self.tcx.struct_span_lint_hir(UNUSED_ATTRIBUTES, hir_id, meta.span(), |lint| {
+                lint.build("`#[doc(test(...)]` takes a list of attributes")
+                    .warn(
+                        "this was previously accepted by the compiler but is \
+                         being phased out; it will become a hard error in \
```
jyn514, line 757: This is not how future incompat lints work. The diagnostic machinery already prints that text,, this will repeat it. The problem is you used UNUSED_ATTRIBUTES which is not a future incompat lint - please use INVALID_DOC_ATTRS instead. There are examples above and below not 100 lines from here, you don't have to rewrite this from scratch.

src/test/rustdoc-ui/doc-test-attr.rs
```diff
@@ -0,0 +1,16 @@
+#![crate_type = "lib"]
+#![deny(unused_attributes)]
+
+#![doc(test)]
+//~^ ERROR
+//~^^ WARN
```
jyn514, line 6: Please use more specific checks.

src/test/ui/rustdoc/doc-test-attr.rs
```diff
@@ -0,0 +1,16 @@
+#![crate_type = "lib"]
+#![deny(unused_attributes)]
+
+#![doc(test)]
+//~^ ERROR
```
jyn514, line 5: Why have this in both rustdoc and rustc? Doesn't it make sense to only have the rustc test?

jyn514 (changes requested):
Please look at how this was done in the earlier PR. Rewriting things from scratch introduces bugs.

src/test/ui/rustdoc/doc-test-attr.rs
GuillaumeGomez: We need to ensure that the warning is emitted from both the compiler and rustdoc.

jyn514: Hmm, ok, I guess I buy that since rustdoc silences some lints. Can you add a comment to keep the two files in sync?

GuillaumeGomez: Normally all tests in `src/test/ui/rustdoc` have to be present in `rustdoc-ui` as well.

jyn514: That was not apparent to me. Can you add a note to the readme in ui/rustdoc?

GuillaumeGomez: I'll extend the existing README content. :)

compiler/rustc_passes/src/check_attr.rs
GuillaumeGomez: I'm really completely out of it. Sorry for the mess...

GuillaumeGomez:
Updated!

src/test/ui/rustdoc/README.md
```diff
@@ -1,3 +1,6 @@
 This directory is for tests that have to do with rustdoc, but test the behavior
 of rustc. For example, rustc should not warn that an attribute rustdoc uses is
 unknown.
+
+All these examples are supposed to have an exact equivalent in the
```
camelid, line 5: By "supposed to", do you mean "please do add an exact copy" or do you mean "tidy enforces that there is an exact copy"?

src/test/ui/rustdoc/doc-test-attr.rs
```diff
@@ -0,0 +1,16 @@
...
+//~^ ERROR `#[doc(test(...)]` takes a list of attributes
+//~^^ WARN this was previously accepted by the compiler
+#![doc(test = "hello")]
+//~^ ERROR `#[doc(test(...)]` takes a list of attributes
+//~^^ WARN this was previously accepted by the compiler
+#![doc(test(a))]
+//~^ ERROR unknown `doc(test)` attribute `a`
+//~^^ WARN this was previously accepted by the compiler
+#![doc(test(no_crate_inject))]
+#![doc(test(attr(deny(warnings))))]
```
camelid, line 14: I'm wondering if it's worth having a separate check-pass test to ensure these do definitely work. Could you add one?

src/test/ui/rustdoc/README.md
GuillaumeGomez: "please do it", there is no enforcement. Maybe I should add one...

src/test/ui/rustdoc/doc-test-attr.rs
GuillaumeGomez: I'll add one then.

GuillaumeGomez:
Updated!

src/test/ui/rustdoc/README.md
camelid: Ah, in that case, could you tweak the wording so it's clear that (as of now) the situation is "please have an exact equivalent", not "tidy will enforce there is an exact equivalent"?

GuillaumeGomez: Ok!

GuillaumeGomez:
Improved the README update wording.

jyn514:
I don't think we should change the readme until we reach a decision on https://github.com/rust-lang/rust/pull/87845. We should not document a decision that hasn't yet been made.

GuillaumeGomez:
Should I remove the changes in the README for the time being?

jyn514:
Yes, please.

GuillaumeGomez:
Done!

compiler/rustc_passes/src/check_attr.rs
```diff
@@ -717,6 +717,42 @@ impl CheckAttrVisitor<'tcx> {
         true
     }
 
+    /// Checks that `doc(test(...)` attribute contains only valid attributes. Returns `true` if
```
jyn514, line 720:
```suggestion
    /// Checks that `doc(test(...))` attribute contains only valid attributes. Returns `true` if
```

compiler/rustc_passes/src/check_attr.rs
```diff
@@ -717,6 +717,42 @@ impl CheckAttrVisitor<'tcx> {
...
+    /// valid.
+    fn check_test_attr(&self, meta: &NestedMetaItem, hir_id: HirId) -> bool {
+        let mut is_valid = true;
+        if let Some(metas) = meta.meta_item_list() {
+            for i_meta in metas {
```
jyn514, line 725: Can you test how this behaves for `doc(test())`? I'd expect it to do nothing.

jyn514 (commented):
r=me with the nit fixed and test added

compiler/rustc_passes/src/check_attr.rs
GuillaumeGomez: Sure.

PR #87640, opened 2021-07-30
Rollup of 9 pull requests

JohnTitor:
Successful merges:

 - #86072 (Cross compiling rustc_llvm on Darwin requires zlib.)
 - #87385 (Make `SEMICOLON_IN_EXPRESSIONS_FROM_MACROS` warn by default)
 - #87547 (Add missing examples for NonNull)
 - #87557 (Fix issue with autofix for ambiguous associated function from Rust 2021 prelude when struct is generic)
 - #87559 (Tweak borrowing suggestion in `for` loop)
 - #87596 (Add warning when whitespace is not skipped after an escaped newline)
 - #87606 (Add some TAIT-related regression tests)
 - #87609 (Add docs about performance and `Iterator::map` to `[T; N]::map`)
 - #87616 (Fix missing word in rustdoc book)

Failed merges:

r? @ghost
@rustbot modify labels: rollup

[Create a similar rollup](https://bors.rust-lang.org/queue/rust?prs=86072,87385,87547,87557,87559,87596,87606,87609,87616)

PR #87730, opened 2021-08-03
perf-check: Revert "Rollup merge of #87596"

JohnTitor:
Checking the perf regression in https://github.com/rust-lang/rust/pull/87640#issuecomment-891911569.
Only this PR touches `rustc_ast` in the rollup and it should affect the bootstrap time at least, let's see if this also causes the regression.
r? @ghost cc @Mark-Simulacrum


JohnTitor:
Seems this one isn't, closing.

PR #87734, opened 2021-08-03
Test dropping union fields more

syvb:
Now that #87403 is merged, a few more tests can be added for reads/writes to dropping union fields.

r? @LeSeulArtichaut


src/test/ui/unsafe/union-assignop.rs
```diff
@@ -0,0 +1,30 @@
...
+    b: Dropping, // treated as dropping
+}
+
+fn main() {
+    let mut foo = Foo { a: 42 };
+    foo.a += 5; //~ ERROR access to union field is unsafe
+    foo.b += Dropping; //~ ERROR access to union field is unsafe
+    foo.b = Dropping; //~ ERROR assignment to union field that might need dropping is unsafe
+    foo.a; //~ ERROR access to union field is unsafe
+    let foo = Foo { a: 42 };
```
LeSeulArtichaut, line 24: Why do you need to re-assign to `foo` here?

LeSeulArtichaut (approved):
:+1: Having more tests for this is definitely a good idea. Just one small thing and we can merge.

src/test/ui/unsafe/union-assignop.rs
Smittyvb: Otherwise the borrow checker complains that `foo` is used after being moved out of.

LeSeulArtichaut:
Thanks for the PR! @bors r+ rollup

PR #87736, opened 2021-08-03
#[inline] slice::Iter::advance_by

the8472:
https://github.com/rust-lang/rust/pull/87387#issuecomment-891942661 was marked as a regression. One of the methods in the PR was missing an inline annotation unlike all the other methods on slice iterators.

Let's see if that makes a difference.

### DIFF
--- a/library/core/src/slice/iter/macros.rs
+++ b/library/core/src/slice/iter/macros.rs
@@ -185,8 +185,9 @@ macro_rules! iterator {
                 }
             }
 
+            #[inline]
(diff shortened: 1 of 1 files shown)


the8472:
That recovers the [regex-opt regression](https://perf.rust-lang.org/compare.html?start=b289bb7fdfcb6f54d825927ab9b5722cabc2a140&end=6b0b07d41f07e1ba5808693d900903499ccf7a32&stat=instructions:u)

PR #87740, opened 2021-08-03
Disable unused variable lint for naked functions

npmccallum:
In most calling conventions, accessing function parameters may require
stack access. However, naked functions have no assembly prelude to set
up stack access.  This is why naked functions may only contain a single
`asm!()` block. All parameter access is done inside the `asm!()` block,
so we cannot validate the liveness of the input parameters. Therefore,
we should disable the lint for naked functions.

rust-lang/rfcs#2774
rust-lang/rfcs#2972

@joshtriplett @Amanieu @haraldh


compiler/rustc_passes/src/liveness.rs
```diff
@@ -332,6 +332,13 @@ impl<'tcx> Visitor<'tcx> for IrMaps<'tcx> {
             }
         }
 
+        // Don't run unused pass for #[naked]
+        if self.tcx.has_attr(def_id, sym::naked) {
+            if let DefKind::Fn | DefKind::AssocFn = self.tcx.def_kind(def_id.expect_local()) {
```
Amanieu, line 337: Why is this line needed? Surely we already have a check somewhere else that `#[naked]` can only be applied to functions (and not closures for example).

npmccallum: Great question. It looks like we do have a check [here](https://github.com/rust-lang/rust/blob/master/compiler/rustc_passes/src/check_attr.rs#L214-L239). I'll remove this.

PR #87756, opened 2021-08-04
Add back -Zno-profiler-runtime

Amanieu:
This was removed by #85284 in favor of `-Zprofiler-runtime=<name>`.However the suggested `-Zprofiler-runtime=None` doesn't work because`None` is treated as a crate name.


jackh726:
Wait how do the other optional flags work then? Can they ever be `None` then?

Amanieu:
All the other optional flags *default* to `None` and are changed to `Some` when you specify them on the command-line.

jackh726 (approved):
I wonder if it makes sense to change the parsing for optional args to count "None" as `None`. It'd be nice to not have two flags here.

But I think this is fine for now.

PR #87760, opened 2021-08-04
Promote `aarch64-apple-ios-sim` to Tier 2

badboy:
As per https://github.com/rust-lang/compiler-team/issues/428 this target becomes Tier 2.

---

I tested that the latest nightly has the target and I'm already building my project with that on CI and locally.

### DIFF
--- a/src/doc/rustc/src/platform-support.md
+++ b/src/doc/rustc/src/platform-support.md
@@ -165,6 +165,7 @@ target | std | notes
 `wasm32-unknown-unknown` | ✓ | WebAssembly
 `wasm32-wasi` | ✓ | WebAssembly with WASI
 `x86_64-apple-ios` | ✓ | 64-bit x86 iOS
+[`aarch64-apple-ios-sim`](platform-support/aarch64-apple-ios-sim.md) | ✓ |  | Apple iOS Simulator on ARM64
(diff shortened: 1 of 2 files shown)


badboy:
Woops, unrelated changes krept in.

jyn514:
@LeSeulArtichaut looks like https://github.com/rust-lang/highfive/pull/249 isn't working

LeSeulArtichaut:
Gah, what about r? rust-lang/compiler-team

LeSeulArtichaut:
I think it might have just been spurious, because the regex seems correct and I don't think the problem can come from somewhere else.

jyn514:
@LeSeulArtichaut highfive hardcodes the team names, it doesn't fetch them from GitHub: https://github.com/rust-lang/highfive/blob/62b1531cbab59ed601ec6087c961ed80f3d68931/highfive/configs/rust-lang/rust.json#L4
Could you use the API instead so pings and reviewers can't be different names? It seems confusing to have to look it up in the highfive source itself.

Mark-Simulacrum:
@bors r+

We should probably take highfive discussion elsewhere -- but I don't think fetching the team from GitHub or the team repository currently works well. Many people on T-compiler (and other teams, I presume) are on the team but are not reviewers in rust-lang/rust (don't have time), so we likely shouldn't roll the dice to them. Configuring the relevant teams in rust-lang/rust-highfive does not seem like that much hassle (mostly just renaming existing groups is my guess).

PR #87761, opened 2021-08-04
Fix overflow in rustc happening if the `err_count()` is reduced in a stage.

### LINKED ISSUE #84219: Internal Compiler Error: 'attempt to subtract with overflow', compiler/rustc_session/src/session.rs:478:22
Build `rustc` with overflow checking enabled:
```
RUSTFLAGS="-C overflow-checks=on" ./x.py build --stage 1
```
Then try compiling the following program (found by [fuzz-rustc](https://github.com/dwrensha/fuzz-rustc)):
```rust
const M;
```

```
$ ./build/x86_64-unknown-linux-gnu/stage1/bin/rustc bug.rs 
error: free constant item without body
 --> bug.rs:1:1
  |
1 | const M;
  | ^^^^^^^-
  |        |
  |        help: provide a definition for the constant: `= <expr>;`

error[E0601]: `main` function not found in crate `bug`
 --> bug.rs:1:1
  |
1 | const M;
  | ^^^^^^^^ consider adding a `main` function to `bug.rs`

thread 'rustc' panicked at 'attempt to subtract with overflow', /home/dwrensha/src/rust/compiler/rustc_session/src/session.rs:475:22

note: rustc 1.51.0-dev running on x86_64-unknown-linux-gnu

query stack during panic:
#0 [analysis] running analysis passes on this crate
end of query stack

For more information about this error, try `rustc --explain E0601`.

```

hkratz:
This can happen if stashed diagnostics are removed or replaced with fewer errors. The semantics stay the same if built without overflow checks. Fixes #84219.

Background: I came across this independently by running `RUSTFLAGS="-C overflow-checks=on" ./x.py test`. Fixing this will allow us to move on and find further overflow errors with testing or fuzzing.

### DIFF
--- a/compiler/rustc_session/src/session.rs
+++ b/compiler/rustc_session/src/session.rs
@@ -450,8 +450,7 @@ impl Session {
     {
         let old_count = self.err_count();
         let result = f();
-        let errors = self.err_count() - old_count;
-        if errors == 0 { Ok(result) } else { Err(ErrorReported) }
+        if self.err_count() == old_count { Ok(result) } else { Err(ErrorReported) }
     }
     pub fn span_warn<S: Into<MultiSpan>>(&self, sp: S, msg: &str) {
         self.diagnostic().span_warn(sp, msg)


jyn514:
Can you add a test for this somehow?

hkratz:
> Can you add a test for this somehow?

I don't see a way without enabling overflow checks for the compiler.

Mark-Simulacrum:
@bors r+

I'm going to go ahead and approve this - I think in terms of testing we might want to try having a debug-assert builder also enable overflow checks. I'm not confident that this is the right behavior - it's likely more interesting changes are warranted - but for now this'll get us to parity with non overflow checks enabled compilers and the stashed diagnostics problem can be examined separately if needed.

PR #87767, opened 2021-08-04
`fn windows`: do not panic for a size of 0

lcnr:
I understand `fn windows` as the following,
```rust
let mut start = 0;
while start + size <= slice.len() {
    yield &slice[start..start+size];
    start += 1;
}
```
so this method seems well defined for a size of 0 to me.

r? @m-ou-se

### DIFF
--- a/library/core/src/slice/iter.rs
+++ b/library/core/src/slice/iter.rs
@@ -11,7 +11,6 @@ use crate::intrinsics::{assume, exact_div, unchecked_sub};
 use crate::iter::{FusedIterator, TrustedLen, TrustedRandomAccess, TrustedRandomAccessNoCoerce};
 use crate::marker::{PhantomData, Send, Sized, Sync};
 use crate::mem;
-use crate::num::NonZeroUsize;
 use crate::ptr::NonNull;
 
 use super::{from_raw_parts, from_raw_parts_mut};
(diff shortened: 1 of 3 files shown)


m-ou-se:
This changes the (stable) `.windows(n)` and (unstable) `.array_windows::<n>()` methods on arrays to no longer panic for `n == 0`, but instead just return windows of length zero, simplifying the code and making the interface more consistent. I'm guessing the panicking behaviour was originally copied from `.chunks(0)`, where that makes more sense.

@rfcbot merge

BurntSushi (approved):
I can't seem to tick my box.

m-ou-se:
Same here. I can edit the comment and put an `x` in there manually, but the checkbox is unclickable. I hope that's not a perment GitHub ui change.

Mark-Simulacrum:
In the meantime, @rfcbot reviewed should work I believe (obviously not for me as I'm not on the list).

lcnr:
uh, my impls actually don't just work as I expected :sweat_smile:

---

I've introduced the following bugs:

I forgot that we can have zst slices of `usize::MAX` length (sadly), which means that `ArrayWindows` can return `usize::MAX + 1` elements for a slice with `usize::MAX` elements which we can't really deal with.
```rust
let x = &[(); usize::MAX];
x.array_windows::<N>(); // this now causes an integer overflow for any `N` because of my impl
x.array_windows::<0>(); // this has an unrepresentable size
```


My changes to `windows` cause a panic because `fn next` is currently implemented by slicing the underlying slice: `self.v[1..]`, which is oob if the size is 0.
```rust
fn test_windows_zero() {
    let v: &[i32] = &[0, 1];
    let mut iter = v.windows(0);

    assert_eq!(iter.next(), Some(&[][..]));
    assert_eq!(iter.next(), Some(&[][..]));
    assert_eq!(iter.next(), Some(&[][..])); // causes a panic here
    assert_eq!(iter.next(), None);
}
```

While the second one can be fixed by modifying the implementation, the fact that we can have an unrepresentable iterator length seems undesirable. Will go ahead and add these problems as tests and close this PR.

:sweat_smile:

yaahc:
@rfcbot cancel

Issue #87766, opened 2021-08-04
Tracking issue for organizing the standard library's #![feature]s


m-ou-se:
The standard libraries use a lot of `#![feature]`s. We should organize them a bit and make sure it's acceptable for all of them to be used in the way they are.

It'd be nice to split them in language and library features. For the language features, we should check with the relevant team(s) if it's okay that we're using the feature in the library, and document with which of them we need to be extra careful and what their pitfalls are.

There's also a few features enabled that are no longer in use. We should remove those attributes.

- [ ] Categorize and clean up features
    - [x] `core`: https://github.com/rust-lang/rust/pull/87768
    - [x] `alloc`: https://github.com/rust-lang/rust/pull/87769
    - [ ] `std`
    - [ ] other libraries
- [ ] Make an overview of used language features
- [ ] Discuss those language features and their uses with the relevant team(s)
- [ ] Document stuff for future contributors and reviewers

JmPotato:
Hi! I noticed that it had been a while since the last update of this issue, so I took some time to clean up the std library's `![feature]`s. Could you please take a look at #94546, thx!

clarfonthey:
I *think* we've done this satisfactorily enough to close this.

Issue #66494, opened 2019-11-17
-Zborrowck and #![feature(nll)] still exist


bjorn3:
The AST borrowck was removed in https://github.com/rust-lang/rust/pull/64790, so these flags don't have a purpose anymore.

bjorn3:
`#![feature(nll)]` seems to matter to `libstd` somehow. When removing `#![feature(nll)]` it gives the following errors:

```
error[E0309]: the parameter type `K` may not live long enough
    --> src/libstd/collections/hash/map.rs:2435:9
     |
2427 | impl<'a, K, V, S> Extend<(&'a K, &'a V)> for HashMap<K, V, S>
     |          - help: consider adding an explicit lifetime bound `K: 'a`...
...
2435 |         self.base.extend(iter)
     |         ^^^^^^^^^^^^^^^^^^^^^^
     |
note: ...so that the reference type `&'a K` does not outlive the data it points at
    --> src/libstd/collections/hash/map.rs:2435:9
     |
2435 |         self.base.extend(iter)
     |         ^^^^^^^^^^^^^^^^^^^^^^

error[E0309]: the parameter type `V` may not live long enough
    --> src/libstd/collections/hash/map.rs:2435:9
     |
2427 | impl<'a, K, V, S> Extend<(&'a K, &'a V)> for HashMap<K, V, S>
     |             - help: consider adding an explicit lifetime bound `V: 'a`...
...
2435 |         self.base.extend(iter)
     |         ^^^^^^^^^^^^^^^^^^^^^^
     |
note: ...so that the reference type `&'a V` does not outlive the data it points at
    --> src/libstd/collections/hash/map.rs:2435:9
     |
2435 |         self.base.extend(iter)
     |         ^^^^^^^^^^^^^^^^^^^^^^
```

Mark-Simulacrum:
There's a separate tracking issue for this (NLL at least, and I believe @Centril noted recently there was a second one to consider). Closing since this is a duplicate; cc https://github.com/rust-lang/rust/issues/58781 https://github.com/rust-lang/rust/issues/57895

PR #87769, opened 2021-08-04
Alloc features cleanup

m-ou-se:
This sorts and categorizes the `#![features]` in `alloc` and removes unused ones.

This is part of #87766

The following feature attributes were unnecessary and are removed:

```diff
// Library features:
-#![feature(cow_is_borrowed)]
-#![feature(maybe_uninit_uninit_array)]
-#![feature(slice_partition_dedup)]

// Language features:
-#![feature(arbitrary_self_types)]
-#![feature(auto_traits)]
-#![feature(box_patterns)]
-#![feature(decl_macro)]
-#![feature(nll)]
```


bjorn3:
> ```diff
> -#![feature(nll)]
> ```

These exist to test full nll mode. https://github.com/rust-lang/rust/pull/66498#issue-341885059

m-ou-se:
If we really need to keep that in `alloc` and/or `core`, we should put a clear comment why it's there. Is there any documentation of what enabling this feature will change?

bjorn3:
NLL is currently in migration mode by default, which still uses the old AST based regionck. `#![feature(nll)]` completes the transition to full NLL mode. https://github.com/rust-lang/rust/issues/57895 is the tracking issue for completing the transition. One of the last remaining blockers for completing the transition is https://github.com/rust-lang/rust/pull/86700 I think. By the way libstd used to require it at least at some point: https://github.com/rust-lang/rust/issues/66494#issuecomment-554761463

yaahc:
LGTM once the NLL bit is sorted.

dtolnay:
@bors delegate+

Looks good after a rebase and putting back `#![feature(nll)]`.

PR #87772, opened 2021-08-04
Move naked function ABI check to its own lint

npmccallum:
This check was previously categorized under the lint named
`UNSUPPORTED_NAKED_FUNCTIONS`. That lint is future incompatible and will
be turned into an error in a future release. However, as defined in the
Constrained Naked Functions RFC, this check should only be a warning.
This is because it is possible for a naked function to be implemented in
such a way that it does not break even the undefined ABI. For example, a
`jmp` to a `const`.

Therefore, this patch defines a new lint named
`UNDEFINED_NAKED_FUNCTION_ABI` which contains just this single check.
Unlike `UNSUPPORTED_NAKED_FUNCTIONS`, `UNDEFINED_NAKED_FUNCTION_ABI`
will not be converted to an error in the future.

rust-lang/rfcs#2774
rust-lang/rfcs#2972

### DIFF
--- a/compiler/rustc_lint_defs/src/builtin.rs
+++ b/compiler/rustc_lint_defs/src/builtin.rs
@@ -2733,7 +2733,7 @@ declare_lint! {
     /// #![feature(naked_functions)]
     ///
     /// #[naked]
-    /// pub extern "sysv64" fn f() -> u32 {
+    /// pub extern "C" fn f() -> u32 {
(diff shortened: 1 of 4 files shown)


npmccallum:
@Amanieu Another possible implementation of this is to move this check to a different, pre-existing lint. But I didn't see an obvious candidate.

PR #87773, opened 2021-08-04
Refactor lower_stmts

camsteffen:
I have this change in #87688 but I think it deserves its own PR and a perf run.

Changes `lower_stmt` to `lower_stmts` and avoid creating a `SmallVec` for each `ast::Stmt`.

### DIFF
--- a/compiler/rustc_ast_lowering/src/lib.rs
+++ b/compiler/rustc_ast_lowering/src/lib.rs
@@ -64,7 +64,7 @@ use rustc_span::symbol::{kw, sym, Ident, Symbol};
 use rustc_span::{Span, DUMMY_SP};
 use rustc_target::spec::abi::Abi;
 
-use smallvec::{smallvec, SmallVec};
+use smallvec::SmallVec;
 use std::collections::BTreeMap;
 use std::mem;
 use tracing::{debug, trace};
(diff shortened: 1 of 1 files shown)


compiler/rustc_ast_lowering/src/lib.rs
```diff
@@ -2414,50 +2408,56 @@ impl<'a, 'hir> LoweringContext<'a, 'hir> {
...
-            StmtKind::Semi(ref e) => {
-                let e = self.lower_expr(e);
-                let hir_id = self.lower_node_id(s.id);
-                self.alias_attrs(hir_id, e.hir_id);
-                (hir_id, hir::StmtKind::Semi(e))
+    fn lower_stmts(
+        &mut self,
+        ast_stmts: &[Stmt],
+    ) -> (&'hir [hir::Stmt<'hir>], Option<&'hir hir::Expr<'hir>>) {
+        let mut stmts = SmallVec::<[hir::Stmt<'hir>; 8]>::new();
```
camsteffen, line 2415: I don't know if 8 is optimal, but I think it is already used in the current implementation, [in `alloc_from_iter`](https://github.com/rust-lang/rust/blob/6fe0886723c9e08b800c9951f1c6f6a57b2bf22c/compiler/rustc_arena/src/lib.rs#L129).

camsteffen:
Neutral results I guess. Maybe it's easier to just close this PR.

PR #87775, opened 2021-08-04
Add hint for unresolved associated trait items if the trait has a single item

### LINKED ISSUE #87638: "cannot find associated type" could hint existing associated types
The following error gives no hints for resolving it, even though there is an obvious solution, in this case the only existing associated type of std::ops::Deref.

Given the following code: https://play.rust-lang.org/?version=nightly&mode=debug&edition=2018&gist=976e4a98c62d7a0efb2d0d6325774495

```rust
fn main() {
    let s : <String as std::ops::Deref>::Output = "hey";
}
```

The current output is:

```
error[E0576]: cannot find associated type `Output` in trait `std::ops::Deref`
 --> src/main.rs:2:42
  |
2 |     let s : <String as std::ops::Deref>::Output = "hey";
  |                                          ^^^^^^ not found in `std::ops::Deref`

For more information about this error, try `rustc --explain E0576`.
```

```
error[E0576]: cannot find associated type `Output` in trait `std::ops::Deref`
 --> src/main.rs:2:42
  |
2 |     let s : <String as std::ops::Deref>::Output = "hey";
  |                                          ^^^^^^ not found in `std::ops::Deref`

note: Did you mean `Target`?

For more information about this error, try `rustc --explain E0576`.
```

Kobzol:
This PR introduces a special-cased hint for unresolved trait items paths. It is shown if:
- the path was not resolved to any existing trait item
- and no existing trait item's name was reasonably close with regard to edit distance
- and the trait only has a single item in the corresponding namespace

I didn't know where I should put tests, therefore so far I just managed to bless two existing tests. I would be glad for hints where should tests for a hint like this be created, how should they be named (with reference to the original issue?) and what tests should I create (is it enough to test it just for types? or create separate tests also for functions and constants?).

It could also be turned into a machine applicable suggestion I suppose.

This is my first `rustc` PR, so please go easy on me :)

Fixes: https://github.com/rust-lang/rust/issues/87638

### DIFF
--- a/compiler/rustc_resolve/src/late/diagnostics.rs
+++ b/compiler/rustc_resolve/src/late/diagnostics.rs
@@ -541,6 +541,10 @@ impl<'a: 'ast, 'ast> LateResolutionVisitor<'a, '_, 'ast> {
                     }
                     _ => {}
                 }
+
+                // If the trait has a single item (which wasn't matched by Levenshtein), suggest it
+                let suggestion = self.get_single_associated_item(&path, span, &source, is_expected);
+                self.r.add_typo_suggestion(&mut err, suggestion, ident_span);
             }
             if fallback {
                 // Fallback label.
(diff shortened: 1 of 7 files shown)


oli-obk:
>I didn't know where I should put tests, therefore so far I just managed to bless two existing tests. I would be glad for hints where should tests for a hint like this be created,

not sure we have a dedicated diagnostics folder, but you could put them in the [associated items folder](https://github.com/rust-lang/rust/tree/master/src/test/ui/associated-item).

> how should they be named (with reference to the original issue?) and what tests should I create

I think a single `issue-87638.rs` file should suffice. You can then even add `//~ NOTE` comments to make sure the notes are actually emitted (and no one accidentally blesses them away).

> (is it enough to test it just for types? or create separate tests also for functions and constants?).

It would be enough, but it also doesn't hurt to test all three.

> It could also be turned into a machine applicable suggestion I suppose.

That would be cool, but not necessary. If you want to take a shot at it in this PR, feel free to do it. You can even test+bless it by adding a `// run-rustfix` comment to your test file.

compiler/rustc_resolve/src/late/diagnostics.rs
```diff
@@ -585,6 +592,37 @@ impl<'a: 'ast, 'ast> LateResolutionVisitor<'a, '_, 'ast> {
...
+                self.resolve_path(mod_path, None, false, span, CrateLint::No)
+            {
+                let resolutions = self.r.resolutions(module).borrow();
+                let targets: Vec<_> = resolutions
+                    .iter()
+                    .filter_map(|(key, resolution)| {
+                        resolution
+                            .borrow()
+                            .binding
+                            .map(|binding| binding.res())
```
oli-obk, line 614: I'm guessing the `add_typo_suggestion` function does something similar to this piece of code. Do you think these two functions can reasonably share code? Or will that just complicate things?

oli-obk: 😆 funny, but not wrong per se, it is the similarest name after all. Let's do that then!

Kobzol:
I added the distinguishing enum to `TypoSuggestion`. It could be prettier but it seems to me that changing the other functions would result in much more mess.

oli-obk:
Thanks! This looks good now. Not sure why CI is failing, I'll try restarting it.

Kobzol:
image
CI seems pretty crazy.

oli-obk:
Thanks for taking care of this so fast!

Issue #81615, opened 2021-02-01
Provide a means of turning iterators into fixed-size arrays


bstrie:
Now that min_const_generics is approaching there is a great deal of new support for working with fixed-sized arrays, such as `std::array::IntoIter`. But while converting from an array to an iterator is now well-supported, the reverse is lacking:
```rust
let mut a = std::array::IntoIter::new([1,2,3]);
let b: [i32; 3] = a.collect(); // the trait `FromIterator<{integer}>` is not implemented for `[i32; 3]`
```
It is roundaboutly possible to do the conversion by going through Vec:
```rust
let mut a = std::array::IntoIter::new([1,2,3]);
let b: [i32; 3] = a.collect::<Vec<_>>().try_into().unwrap();
```
But that isn't no_std compatible, and even with std there should be no need to allocate here.

The non-allocating way of converting the array presupposes a fair bit of familiarity with the stdlib, unsafe code, and unstable features:
```rust
#![feature(maybe_uninit_uninit_array)]
#![feature(maybe_uninit_array_assume_init)]
let mut array: [MaybeUninit<T>; N] = MaybeUninit::uninit_array();

for i in 0..N {
    array[i] = MaybeUninit::new(iter.next().unwrap());
}

let array = unsafe {
    MaybeUninit::array_assume_init(array)
};
```
...which suggests this is a prime candidate for stdlib inclusion.

The first problem is what the signature should be. There's no way to statically guarantee that an iterator is any given length, so (assuming that you don't want to panic) what is the return type when the iterator might be too short?

1. `-> [T; N]`: straightforward, but you'd have to have `T: Default`, which limits its usefulness. Furthermore this uses in-band signaling to mask what is probably an error (passing in a too-small iterator), which feels user-hostile. This seems little better than panicking.
2. `-> [Option<T>; N]`: the obvious solution to indicate potentially-missing data, but likely annoying to work with in the success case. And sadly [the existing blanket impl](https://doc.rust-lang.org/std/iter/trait.FromIterator.html#impl-FromIterator%3COption%3CA%3E%3E) that ordinarily allows you `.collect` from from a collection of options into an option of a collection can't be leveraged here because you still don't have an impl of `FromIterator<T> for [T; N]`. So if you actually want a `[T; N]` you're left with manually iterating over what was returned, which is to say, you're no better off than having the iterator you started out with!
[...]

PR #87776, opened 2021-08-04
Add Iterator::array_chunks method

johnschug:
This adds an `array_chunks` (and `array_rchunks`) method to the iterator trait similar to the method with the same name on slices. Issue #81615 seems to indicate this would be useful.


library/core/src/iter/adapters/array_chunks.rs
```diff
@@ -0,0 +1,146 @@
...
+///
+/// This `struct` is created by the [`array_chunks`] method on [`Iterator`]. See
+/// its documentation for more.
+///
+/// [`array_chunks`]: Iterator::array_chunks
+#[unstable(feature = "iter_array_chunks", issue = "none")]
+#[derive(Debug)]
+pub struct ArrayChunks<I: Iterator, const N: usize> {
+    iter: I,
+    buffer: [MaybeUninit<I::Item>; N],
```
the8472, line 16: Maybe `Option<(Box<[MaybeUninit<I::Item>; N]>, usize)>` would make sense? Then nothing has to be allocated while the iterator is being passed around and re-wrapped in other iterators. This avoids bloat when N is large. Only when the end is reached and a size-mismatch occurs an allocation will be needed.

Or perhaps even a simple `Box<[T]>` will do in that case since it only has to be allocated once.

library/core/src/iter/adapters/array_chunks.rs
```diff
@@ -0,0 +1,146 @@
...
+
+    fn next(&mut self) -> Option<Self::Item> {
+        while self.init < N {
+            self.buffer[self.init] = MaybeUninit::new(self.iter.next()?);
+            self.init += 1;
+        }
+        self.init = 0;
+        // SAFETY: This is safe: `MaybeUninit<T>` is guaranteed to have the same layout
+        // as `T` and the entire array has just been initialized.
+        unsafe { Some(mem::transmute_copy(&self.buffer)) }
```
the8472, line 68: Reusing memory (going through `self`) and then transmute-copying it to return it seems like it would optimize badly. Benchmarks exercising this adapter together with a for-in loop as baseline would help.

library/core/src/iter/adapters/array_chunks.rs
```diff
@@ -0,0 +1,146 @@
...
+    /// returned slice has at most `N-1` elements.
+    #[unstable(feature = "iter_array_chunks", issue = "none")]
+    pub fn remainder_mut(&mut self) -> &mut [I::Item] {
+        // SAFETY: We know that all elements before `init` are properly initialized.
+        unsafe { MaybeUninit::slice_assume_init_mut(&mut self.buffer[..self.init]) }
+    }
+}
+
+#[unstable(feature = "iter_array_chunks", issue = "none")]
+impl<I: Iterator, const N: usize> Iterator for ArrayChunks<I, N> {
```
the8472, line 57: The implementation should at least get `size_hint` and `advance_by`. `fold` and `try_fold` could also make a big difference for efficiency.

Once `size_hint` is correctly implemented a conditional `TrsutedLen` impl could be added.

You can look at other Iterator adapters as a reference.

library/core/src/iter/adapters/array_chunks.rs
johnschug: This is in `core`, so allocation is not available. Maybe it could be moved to `alloc`, but I don't know if `Iterator` can be extended there. I could also make a struct that just holds the buffer and implement `IntoIterator` on a mutable reference to it. This would be similar to the `IntoChunks` adapter in `itertools`, but would be the first adapter in `core` to do this.

library/core/src/iter/adapters/array_chunks.rs
johnschug: I don't know if this iterator can actually uphold the contract of `TrustedLen`. If the underlying iterator returns `(usize::MAX, None)`, in most cases it would not be possible to tell whether this iterator would return less than or equal to `usize::MAX` elements (arrays). For example, with `N=2` and actual length of the underlying iterator being `usize::MAX+1`, it would be less than `usize::MAX`, but with `N=2` and actual length of the underlying iterator being `10*usize::MAX`, it would be more than `usize::MAX`. I think the only case where it would be possible is when `N=1`.

johnschug: Ok, I have added those methods now. The `advance_by` method is potentially lossy though if the underlying iterator advances by less than the requested amount.

library/core/src/iter/adapters/array_chunks.rs
johnschug: I have added some benchmarks comparing it to looping over a mapped iterator producing arrays and the `ArrayChunks` iterator for slices.
```
test iter::bench_array_map                                      ... bench:     437,783 ns/iter (+/- 13,664)
test iter::bench_iter_array_chunks                              ... bench:   2,684,281 ns/iter (+/- 206,116)
test iter::bench_iter_array_chunks_fold                         ... bench:   2,217,095 ns/iter (+/- 127,330)
test iter::bench_slice_array_chunks                             ... bench:     335,405 ns/iter (+/- 99,203)
```

johnschug: Is there a better way to run specific benchmarks for `core`, than running all of them with `./x.py bench library/core`? It doesn't seem to accept that standard filter that cargo bench uses.

johnschug: I managed to improve the performance by quite a bit. Now it only takes twice as long as passing an array through an iterator pipeline. I also removed the `transmute_copy` (slight increase in performance).

```
test iter::bench_array_map                                      ... bench:     412,536 ns/iter (+/- 19,784)
test iter::bench_iter_array_chunks_fold                         ... bench:     856,663 ns/iter (+/- 67,238)
test iter::bench_iter_array_chunks_loop                         ... bench:   1,678,215 ns/iter (+/- 78,573)
test iter::bench_iter_array_chunks_ref_fold                     ... bench:   1,511,018 ns/iter (+/- 86,835)
test iter::bench_slice_array_chunks                             ... bench:     337,726 ns/iter (+/- 83,062)
```

library/core/src/iter/adapters/array_chunks.rs
```diff
@@ -71,76 +71,3 @@ impl<I: Iterator, const N: usize> Iterator for ArrayChunks<I, N> {
...
-/// An iterator that yields the elements of another iterator in
-/// chunks of size `N` starting from the end.
-///
-/// This `struct` is created by the [`array_rchunks`] method on [`Iterator`]. See
-/// its documentation for more.
-///
-/// [`array_rchunks`]: Iterator::array_rchunks
-#[unstable(feature = "iter_array_chunks", issue = "none")]
-#[derive(Debug)]
-pub struct ArrayRChunks<I: DoubleEndedIterator, const N: usize> {
```
jdahlstrom: Is `array_rchunks` actually redundant? Seems to me it's distinct from both `.rev().array_chunks()` (because the elements within the chunks are in the original order) and `.array_chunks().rev()` (when `.remainder().len() != 0`).

johnschug: It's mostly redundant with the exception of the remainder its equivalent to:
`.rev().array_chunks().map(|mut arr| { arr.reverse(); arr})`

If it still seems useful, I can re-include it.

library/core/src/iter/adapters/array_chunks.rs
the8472: > I think the only case where it would be possible is when N=1.

Good point. No need to implement it then, arrays of len 1 don't make much sense.

>  The advance_by method is potentially lossy though

Ah that would break the contract about the remainder. Then can only be implemented when the size is known, i.e. only via specialization. Or maybe by querying the lower size bound. Since it's only a performance optimization we can leave it out for now.

library/core/src/iter/adapters/array_chunks.rs
```diff
@@ -0,0 +1,232 @@
...
+        Self { ptr: MaybeUninit::slice_as_mut_ptr(array), init: 0 }
+    }
+
+    fn with<R, F>(buffer: &mut Buffer<T, N>, f: F) -> R
+    where
+        F: FnOnce(&mut [MaybeUninit<T>; N], &mut usize) -> R,
+    {
+        let mut array = MaybeUninit::uninit_array();
+        let mut guard = Self::new(&mut array);
+        if buffer.init > 0 {
```
the8472, line 73: Why would `buffer.init` ever be larger 0? To deal with non-fused iterators? Can't we use the presence of any item in the buffer as signal that the end was reached and thus we don't need to iterate further, semi-fusing it?

library/core/src/iter/adapters/array_chunks.rs
```diff
@@ -0,0 +1,232 @@
...
+            array = mem::replace(&mut buffer.array, MaybeUninit::uninit_array());
+            guard.init = mem::replace(&mut buffer.init, 0);
+        }
+        let res = f(&mut array, &mut guard.init);
+        if guard.init > 0 {
```
the8472, line 78: Similar here, what's the advantage of the extra compare here? Wouldn't it be better to tie moving the data into the adapter to the `next == None` condition?

the8472 (commented):
This starts to look like it's containing half of an implementation of the fabled `ArrayVec`, as the FIXME notes this also contains some redundancy with the some `array` code.

library/core/benches/iter.rs
```diff
@@ -367,3 +367,54 @@ fn bench_partial_cmp(b: &mut Bencher) {
 fn bench_lt(b: &mut Bencher) {
     b.iter(|| (0..100000).map(black_box).lt((0..100000).map(black_box)))
 }
+
+#[bench]
+fn bench_array_map(b: &mut Bencher) {
+    b.iter(|| {
+        let mut acc = 0;
+        let iter = (0i64..10000).map(black_box).map(|_| black_box([0i64; 100]));
+        for_each_fold(iter, |x| acc += x.iter().sum::<i64>());
```
the8472, line 376: `x.iter()` on an array gives a slice iter, have you tried whether an `array::IntoIter` performs better or worse?

johnschug: The closure here is the same as the closures in the other benchmarks. With the exception of the slice `array_chunks` benchmark (which I have just removed), the type of `x` in each closure is an array, so they should be comparable.

library/core/src/iter/adapters/array_chunks.rs
johnschug: Yeah, it was just to deal with non-fused iterators. I have changed it, so its now just unconditionally fused.

library/core/src/iter/adapters/array_chunks.rs
johnschug: I have now added a `DoubleEndedIterator` implementation that seems to produce the same results `array_rchunks` would. It's not quite as efficient as it could be though, just to keep complexity down and share more code with the forward implementation.

johnschug:
I have made `ArrayChunks` unconditionally fused now, and added a `DoubleEndedIterator` implementation (with benchmarks) to replace `ArrayRChunks`. I have also unified the `Guard` implementation with the one in `array`.

library/core/benches/iter.rs
the8472: Ok, I was asking in case it made them faster and the smaller baseline resulting in a wider gap.

library/core/src/iter/adapters/array_chunks.rs
the8472: But I think it can implement `ExactSizeIterator` when the inner is since that means the inner fits in usize::MAX and thus we can calculate the hint properly.

johnschug:
> This starts to look like it's containing half of an implementation of the fabled `ArrayVec`, as the FIXME notes this also contains some redundancy with the some `array` code.

Yeah, `ArrayVec` would definitely make the implementation a lot simpler. I think it would make it slower too though. For whatever reason using an array that is nested in a `struct` with a `Drop` implementation seems to optimize really poorly.  That was the major difference between the first and second set of benchmarks (removing the `transmute_copy` also shaved about 100μs off).

johnschug:
Are there any other changes this needs? Should I open a tracking issue for this?

JohnCSimon:
> Are there any other changes this needs? Should I open a tracking issue for this?
@joshtriplett

the8472:
Was this closed due to issues with the code or just lack of time? That would be relevant if someone wants to pick it up.

JohnCSimon:
> Was this closed due to issues with the code or just lack of time? That would be relevant if someone wants to pick it up.

@johnschug ?

johnschug:
@the8472 It was a bit of both. I think someone else is working on a partial implementation of `ArrayVec` for core and consolidating the existing wrappers. Although, I can't seem to find the post again. This PR overlaps quite a bit with that work and this function would be trivial to add (and without `unsafe` code) with such an implementation.

Issue #49733, opened 2018-04-06
Tracking issue for box_syntax


aidanhs:
This is a tracking issue for the box_syntax feature, the special syntax for creating boxes, e.g. `box 1usize` (note: box patterns is separate and is tracked at at https://github.com/rust-lang/rust/issues/29641)

This was previously under the same tracking issues as placement (#22181/#27779) but I didn't want to remove it (yet) due to widespread usage in the compiler (`Box::new` is implemented in terms of it) and in the wider ecosystem (unlike `<-` etc which only had a handful of crates using it).

Things to decide:
 - how should `box` syntax be phased out?
   - how easy is it to remove box syntax as a language feature? Can we just describe it as "will never be stabilised in its current form as it's an internal implementation detail, may be stabilised under a new design of placement")?
     - if it *were* possible to remove it as a language feature, is it reasonable to do so without a replacement in the form of working placement?
     - if we want to keep box syntax internally, should we rename (but keep reserved) the keyword to something else to deliberately break users and indicate it shouldn't be used for placement?
   - is there any reason for users to use box syntax (apart from it looking nice) given `Box::new` [is inline always](https://github.com/rust-lang/rust/blob/83669ec/src/liballoc/boxed.rs#L92-L95), or does that [not work](https://github.com/rust-lang/rust/pull/49646#issuecomment-379219615)?
 - if we decide to keep this, there are a number of unresolved questions in #22181 about the syntax behaviour

Note: when removing this, be careful not to remove the syntax entirely in order to not cause another case of https://github.com/rust-lang/rust/issues/50832

spacekookie:
Do I understand it correctly that the `box` syntax is dependent on placement-new (which has now been unapproved as an RFC)?

Personally I like the idea of having some syntactic sugar for creating something as common as a box (it *does* look nice). But obviously it would need to be built on something that works.

steveklabnik:
@spacekookie yes, the big deal for box syntax is placement new.

oyvindln:
>s there any reason for users to use box syntax (apart from it looking nice) given Box::new is inline always, or does that not work?

It doesn't optimize well, and if it did, you would need some other way around it to avoid blowing up the stack in debug mode.
https://github.com/rust-lang/rust/issues/41160

mark-i-m:
> is there any reason for users to use box syntax (apart from it looking nice) given Box::new is inline always, or does that not work?

[...]

PR #49646, opened 2018-04-04
Use box syntax instead of Box::new in Mutex::remutex on Windows

glandium:
The Box::new(mem::uninitialized()) pattern actually actively copies
uninitialized bytes from the stack into the box, which is a waste of
time. Using the box syntax instead avoids the useless copy.

Issue #22181, opened 2015-02-11
`box` and `in` expressions (tracking issue for RFC 809)


nikomatsakis:
Tracking issue for https://github.com/rust-lang/rfcs/pull/809. This should be filled out with TODO items. Note that in this case there are some [unsettled design questions](https://github.com/rust-lang/rfcs/pull/809#issuecomment-73910414) to be resolved as well.

To do:
- [x] land placement-`in` desugaring implementation
- [ ] land overloaded-`box` desugaring implementation
- [ ] port `libstd` APIs to support `in`
- [ ] port `libstd` APIs to support `box`
- [ ] resolve/finalize syntax for placement-`in`
- [ ] investigate improvement of inference, coercion, and cast interactions (subissues #22221, #22403, #22405)
- [ ] improve error messages when code is rejected (#22231)

Unresolved questions:
- [ ] best precedence for `box` (was https://github.com/rust-lang/rust/issues/21192)

pnkfelix:
Status update: https://internals.rust-lang.org/t/lang-team-minutes-feature-status-report-placement-in-and-box/4646

nikomatsakis:
@aidanhs has proposed to remove these features in https://github.com/rust-lang/rust/pull/48333, and the proposal is being actively discussed. =)

aidanhs:
Placement new is imminently about to be/has been removed as an unstable feature and the RFCs unaccepted. The approved/merged PR is at #48333 and the tracking issues were at #22181 (this issue) and #27779. Note that this does not affect box syntax.

You can find the summary comment at https://github.com/rust-lang/rust/issues/27779#issuecomment-378416911.

Find the internals thread where you can discuss this more at https://internals.rust-lang.org/t/removal-of-all-unstable-placement-features/7223. Please add any thoughts there.

PR #87781, opened 2021-08-05
Remove box syntax from compiler and tools

est31:
Removes box syntax from the compiler and tools. In #49733, the future of box syntax is uncertain and the use in the compiler was listed as one of the reasons to keep it. Removal of box syntax [might affect the code generated](https://github.com/rust-lang/rust/pull/49646#issuecomment-379219615) and slow down the compiler so I'd recommend doing a perf run on this.


jyn514:
https://github.com/rust-lang/rust/issues/49733 says

> how easy is it to remove box syntax as a language feature? Can we just describe it as "will never be stabilised in its current form as it's an internal implementation detail, may be stabilised under a new design of placement")? 
> is there any reason for users to use box syntax (apart from it looking nice) given Box::new is inline always, or does that not work?
> if we decide to keep this, there are a number of unresolved questions in box and in expressions (tracking issue for RFC 809) #22181 about the syntax behaviour

That sounds to me like it's not at all certain that it's going to be removed.

est31:
@jyn514 good point but note that the issue also contains:

> I didn't want to remove it (yet) due to widespread usage in the compiler (Box::new is implemented in terms of it) and in the wider ecosystem (unlike <- etc which only had a handful of crates using it).

This PR alleviates the "widespread usage in the compiler" point, if wanted.

Also note that currently the compiler has inconsistent usage of `Box::new` vs `box`. Currently if I grep for `Box::new` in the compiler/ dir, I get 278 uses. With this PR, there would be 413 uses. So it's 135 uses of `box` syntax, or 33% of all box constructions. If this PR turns out to have a significant negative performance impact, it won't just influence the decision on whether to keep the `box` language feature, one could think about introducing the box syntax to more places in the compiler, in the hope of speeding up things.

jyn514:
@rust-timer queue

est31:
(also I have edited the PR description to remove the claim that the tracking issue plans certain removal)

hellow554:
One question though: Without the `box` feature, how will [`Box::new`](https://github.com/rust-lang/rust/blob/d4ad1cfc63ba5824196bfb2370451ddb5af2e020/library/alloc/src/boxed.rs#L191-L193) be implemented in the future? Will it become a language item?

leonardo-m:
>So it's 135 uses of box syntax, or 33% of all box constructions. If this PR turns out to have a significant negative performance impact,

Another benchmark worth doing is to turn all of them into box syntax.

bjorn3:
> One question though: Without the box feature, how will Box::new be implemented in the future? Will it become a language item?

It could call the allocator without lang item indirection like it currently does (`exchange_malloc`), use `ptr::write` to store the value and then construct the `Box` directly from the pointer. All `(try_)new_*` methods already need to do this instead of `box expr`.

hellow554:
If I see it correctly, you haven't removed every occurence of the `box`, right?

I really would like to see that. I think it would be easiest to remove the features `box_syntax` and `box_patterns` and see where the compiler complains to get rid of everything.

jyn514:
Please do not remove box_patterns. It makes code vastly more easy to read and it also has a potential replacement with the Deref pattern project.

hellow554:
Can we have another timer run?

nbdd0121:
`Box::new` needs a move while `box` can construct in-place. The MIR for `Box::new` would be something like (pseudo-code)
```
_1 = construct the value
_2 = construct the box
(*_2) = move _1
```
while the box syntax gives
```
_1 = construct the box;
(*_1) = construct the value
```

And the reality this is more complicated if the `Box::new` is to be implemented by method calls to `exchange_malloc` and `write` (which is not itself a intrinsic) etc. `try_new` method is different that it cannot be assumed to be successful so for semantics you'll need to construct the value first and move it later anyway.

Can we be 100% certain with this change that values will still be constructed in place for things like `RcBox`?

est31:
@hellow554 if I check the library dir, there seem to be a lot of uses of box syntax still in the tests and stuff. Maybe make a new PR to remove it from there as well?

nbdd0121:
I digged into some of the previously poorly optimised cases (https://godbolt.org/z/o6a9b3K5M), and it seems that they are properly optimised in Rust 1.52 (while not in Rust 1.51).

Rust 1.52 updates LLVM to 12, so I guess with LLVM 12, `box` would no longer have performance advantage given most calls to `Box::new` are properly optimised. We still support LLVM 10 and 11 though. Do we have a way to do a perf run with a different LLVM version?

est31:
Thanks for digging into this @nbdd0121 ! About the LLVM versions, one could probably turn back the time and use a different LLVM version in this PR and then re-run a perf run.

I agree that older LLVM versions should be fully supported, even if this means a delay for box syntax removal by a year or two. There is no rush and the distros that use old LLVM versions will be thankful. That being said, one could also do this via a dedicated MIR optimization instead of relying on LLVM. This wouldn't just benefit `Box::new` but any other function that looks like it.

nbdd0121:
A MIR pass that turns `Box::new` into `box` does sound like a good idea (and would hopefully saves some compilation time as less work is required by LLVM).

est31:
Thats's a lot of red wow...

Screenshot_20210807_041429

I guess it doesn't like the other `Box::new` implementation that uses `Box::new_in`? That's the biggest difference I can see to the prior perf run...

hellow554:
> @hellow554 there seem to be a lot of uses of box syntax still in the tests and stuff

Yes, I haven't touched ans Tests at all. First because there are a lot of regression tests that I would like to be untouched and second I thought they wouldn't make any difference in this case, do they?

That's really a big regression 😨
Can we try to find out why? Is it really `Box::new` that causes it?

est31:
> Can we try to find out why? Is it really Box::new that causes it?

We ran a perf run before your commit and one after it. The perf run before your commit shows no performance changes, while the one after it shows significant performance changes. Outside of `Box::new`, I can't see any other changes in that commit that could have caused such large regressions.

There are two ways to test this hypothesis. First, making a fresh PR with only the `Box::new` change. However, that will cause a much smaller regression because currently, large parts of the compiler are still using box syntax, which this PR has changed, so you'd be getting different numbers. The second way to test it would be to take this PR, revert the `Box::new` change, and keep everything else unchanged and do another perf run. I think this is the best way forward.

nbdd0121:
I just created a commit that will converts `Box::new` into `box` during THIR -> MIR building (https://github.com/nbdd0121/rust/commit/4e678bb81341cf3e16c7df5a6025bbe2124bfd6e). Maybe we can try a perf run with that commit added?

est31:
@nbdd0121 thanks for that commit, that was really quick! I'll try to check if I can apply it to this PR. It's not trivial because it's changing code that uses `box` syntax itself. Rebasing my changes onto that commit seems easier I think.

nbdd0121:
It's actually an indentation change that causes conflict. I guess you can just `git checkout --theirs <conflict>` and re-remove `box`.

nbdd0121:
It seems that my commit when interacting with two-phase borrow actually causes ICE. This might be the correct way 😢

nbdd0121:
Borrow checker ICE is fixed in https://github.com/nbdd0121/rust/commit/ca0b1c74c2036e9f3747afa7541d863f2ff692ac. 
```
// Without `Box::new` to `box` transform
_1 = construct the value;
_2 = Box::new(move _1);

// With previous commit or box syntax
// but apparently `Box::new` syntax implies two-phase borrow and the borrow checker requires construct into locals
_1 = Box;
(*_1) = construct the value;

// With the fix
_1 = Box;
_2 = construct the value;
(*_1) = *2;
```

est31:
@jyn514 could you issue another perf run? Thanks.

bjorn3:
@rust-timer check 4b9300307cb984bee4668db59cd24167bd5b4ac8

bjorn3:
Wrong command.

@rust-timer build 4b9300307cb984bee4668db59cd24167bd5b4ac8

jyn514:
@rust-timer build https://github.com/rust-lang/rust/commit/4b9300307cb984bee4668db59cd24167bd5b4ac8

est31:
Huh, still red, although way less red than before.

Screenshot_20210809_004842

The test that is the most red is the [`deep-vector`](https://github.com/rust-lang/rustc-perf/blob/cfea1251520dcb69147df8965048a935fdf1eefe/collector/benchmarks/deep-vector/src/main.rs) one. It constructs a huge vector made up of lots of zeros. Among other changes, this PR alters the `vec` macro to use `Box::new` instead of box syntax. Maybe the MIR pass doesn't work and it still constructs the giant array on the stack? The big regressions of `deep-vector-opt` are in the LLVM passes. In the other regressing examples, I couldn't find a single responsible pass.

nbdd0121:
Refer to my previous comment, due to borrow checker dislike direct construction into the box, I still need to construct on a local temporary (_2) first before moving it into `*_1`. With no local copy propagation pass currently available in MIR this means that it's a large copy that LLVM has to optimise out.

est31:
@nbdd0121 I see that would explain it. I understand why two phase borrows would want parameters in locals first. I'm not familiar with the borrow checker code, but from my understanding, two phase borrows are only needed for functions taking `&mut self`, no? So maybe this requirement that ICEs if the param is not available as a local could be relaxed to only require it for functions taking `&mut self`?

leonardo-m:
If I replace some box [somearray] with Box::new([somearray]) in my code I get stack overflows. Will this be fixed before the box syntax get removed?

est31:
The PR is [blocked](https://github.com/rust-lang/rust/pull/87781#issuecomment-894872854) right now on improving the pass that replaces `Box::new` with something even more equivalent to `box` syntax.

That being said, the state before the library changes had no observable regressions, see the [first perf run](https://github.com/rust-lang/rust/pull/87781#issuecomment-893214467). If someone wants me to revert the PR to that old state and wants to approve it for merge, please tell me.

oli-obk:
> If someone wants me to revert the PR to that old state and wants to approve it for merge, please tell me.

i'd r+ that. people may reintroduce it, but doing it again later will be much easier with the big chunk already done

nbdd0121:
A friendly reminder that this'll still like to regress LLVM 10 and 11.

Also it might worth adding a few codegen tests for `Box::new` to ensure that those optimisations won't regress for previously badly optimized cases (like https://godbolt.org/z/o6a9b3K5M that I mentioned earlier).

oli-obk:
> A friendly reminder that this'll still like to regress LLVM 10 and 11.

So, compilers built with these llvm versions will be slower after this PR? It won't affect user code, as it's just compiler internal crate changes.

est31:
@nbdd0121 oh right I forgot about that. Yeah then the MIR optimization pass would be needed, or we wait. It's hard to measure how much removal of box syntax is slowing down compilers compiled with old LLVM versions, as it is hard to measure how much the MIR pass is helping with that.

nbdd0121:
Sorry, I thought that there are still library changes (probably the title should be changed).

So @oli-obk is correct, this only slows compilers built with the old LLVM version and wouldn't affect users. I suppose that wouldn't be a problem then.

oli-obk:
@bors r+ rollup=never

Ok, let's merge it then. The older llvm versions will get their compiler perf back once we have the mir opt

est31:
@oli-obk your prediction that new boxes will be added turned out to be right faster than anticipated :laughing: . This PR's state is clean, basing on aa8f27bf4d980023a8b245ceb25a490a18041eb2. But already the merge with  30a0a9b694cde95cbab863f7ef4d554f0f46b606 caused issues, namely with ad2b4f44412391ff14d96d71d7b6a212f1d94755 that added a `box` occurence. I've fixed it and re-pushed.

compiler/rustc_codegen_cranelift/example/alloc_example.rs
```diff
@@ -1,4 +1,4 @@
-#![feature(start, box_syntax, core_intrinsics, alloc_prelude, alloc_error_handler)]
+#![feature(start, core_intrinsics, alloc_prelude, alloc_error_handler)]
```
bjorn3, line 1: These actually do still use the box syntax. I guess I will need to update them at the next sync.

est31: Oh sorry about that. CI isn't ran for these so it didn't detect it but I could have double checked.

est31:
@nbdd0121 if you want to submit your commits as a PR, I've rebased them and they are available on my [remove_box](https://github.com/est31/rust/tree/remove_box) branch. cheers!

est31:
Followup PR for the testsuite: #88316

For the question of removing box syntax support from the compiler, there is not much of a difference if box syntax is used in the testsuite or the compiler itself. So I have made a followup PR to remove box syntax from there too, at least outside of the issues directory.

jyn514:
> So I have made a followup PR to remove box syntax from there too, at least outside of the issues directory.

This seems like a really bad idea. How will you possibly ensure that it's testing the same thing that it did before?

Mark-Simulacrum:
Was going to remove the perf-regression, but it looks like the final merge of this PR did actually results in a slight [regression](https://perf.rust-lang.org/compare.html?calcNewSIg=true&start=896f058f13d6c8021f7637817953a44d3a78be32&end=ba8cda2fa2c99ed6646f4dfe73bf4edad7e42a2d&stat=instructions%3Au), primarily for -doc builds. My guess is that some of those larger types in rustdoc are being copied a little more than previously.

est31:
Hmm yeah maybe the rustdoc part should be reverted? I can do a PR for it.

est31:
Or we can wait until #89030 merges.

nbdd0121:
I think regression isn't bad enough to warrant a reversion, especially since it's desirable to phase out box expression in the long term.

est31:
Hmm a glance at #89030 makes it seem that `ShallowInitBox` only affects box syntax right? So it shouldn't improve `Box::new` calls like the commit "Transform Box::new to box" commit?

nbdd0121:
No, it's not related to the transform. That PR is to allow `box` expression to unwind so that `oom=panic` support could be unblocked. (And it would in fact regress performance slightly for all box expressions)

jyn514:
> Hmm yeah maybe the rustdoc part should be reverted? I can do a PR for it.

Please do, yes - collect_blanket_impls especially is absurdly performance sensitive and I don't think this is useful enough to regress the performance there.

Mark-Simulacrum:
I don't think a straight revert is necessarily the right approach -- the regressions here are tiny, realistically, and [wall times](https://perf.rust-lang.org/compare.html?calcNewSig=true&start=896f058f13d6c8021f7637817953a44d3a78be32&end=ba8cda2fa2c99ed6646f4dfe73bf4edad7e42a2d&stat=wall-time) do not show a corresponding significant decrease.

Cachegrind [diff](https://gist.github.com/Mark-Simulacrum/cd6ea8880ec0dde9b599d0ad9684a565) for serde-doc point to a particular set of functions which likely drive the regressions -- targeted reverts likely make more sense here than a blanket revert. That said, if we revert everything in rustdoc for now that doesn't seem the end of the world either. Ultimately, this just points to Box::new being slightly less able to optimize than box, which was already well known.

Mark-Simulacrum:
I also see no particular reason to believe that collect_blanket_impls has been affected by this patch -- can you say more about that perhaps?

est31:
I've filed a PR in #89134 reverting every change of librustdoc. If anything can stay, please point it out!

safinaskar:
https://github.com/rust-lang/rust/pull/87781#issuecomment-894532456 :
> try_new method is different that it cannot be assumed to be successful so for semantics you'll need to construct the value first and move it later anyway

?????
So, `try_new` always performs move? This is very bad. Compare this with C++ code:
```c++
try {
   a = new some_constructor(some_args);
}catch(...){
  //...
}
```
Here we can check for allocation errors while directly constructing value in place!

bjorn3:
On nightly you can use `Box::new_uninit()` followed by `.assume_init()` after writing the value. This is unsafe though. On stable you can use `Box::new()` of a `Box<MaybeUninit<T>>` type and `transmute` to `Box<T>` after writing the value.

```rust
#![feature(new_uninit)]

let mut five = Box::<u32>::new_uninit();

let five: Box<u32> = unsafe {
    // Deferred initialization:
    five.as_mut_ptr().write(5);

    five.assume_init()
};
```

You can do something similar for the `try_*` variant.

PR #87784, opened 2021-08-05
Add config.toml options for enabling overflow checks in rustc and std

hkratz:
The names are `overflow-checks` and `overflow-checks-std` and they work similar to  `debug-assertions` and `debug-assertions-std`. Once added we can measure how big the performance impact actually is and maybe enable them for CI tests.

Enabling them already makes two ui tests fail:
```
failures:
    [ui] ui/parser/item-free-const-no-body-semantic-fail.rs
    [ui] ui/parser/item-free-static-no-body-semantic-fail.rs
```
(See #84219 and #87761.)


hkratz:
Performance impact for running the full stage 1 testsuite locally is less than I expected:
```
# With fix from #87761 for the failing tests.
./x.py clean; time ./x.py test --stage 1

# no checks:
real	23m19.492s
user	176m22.251s
sys	20m28.837s

# w/ overflow checks in everything but std:
real	23m26.973s
user	176m50.005s
sys	20m32.393s

# w/ overflow checks in everything:
real	23m57.156s
user	181m20.771s
sys	21m25.673s
```

Mark-Simulacrum:
Could you change them to default to true in this PR temporarily, so we can run perf on this PR? I think if we can get away with not exposing this as an option (and instead always enabling them) we should.

hkratz:
> Could you change them to default to true in this PR temporarily, so we can run perf on this PR? I think if we can get away with not exposing this as an option (and instead always enabling them) we should.

@Mark-Simulacrum Done. This should be ready for a perf test.

hkratz:
Hmm, at least now we know.

(Sorry the Close and Reopen was an accident on my mobile)

Mark-Simulacrum:
Could you drop the enablement for std and we can rerun benchmarks? I'd be interested in just rustc enablement impact.

I expect it'd be interesting to work out if there's just a few adds or whatever we can replace with explicit wrapping adds (with comments about performance or so). But that's a larger project, I just want numbers for now on the rustc-only overflow checks and we can likely merge after that.

hkratz:
@Mark-Simulacrum Ready for perf testing.

Mark-Simulacrum:
Alright, thanks! Looks like neither is an option for now as default, so if you can drop the testing commits we can probably merge (I'll give another glance over the code but don't think there's likely to be more work before merging).

src/bootstrap/CHANGELOG.md
```diff
@@ -9,6 +9,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 - `llvm-libunwind` now accepts `in-tree` (formerly true), `system` or `no` (formerly false) [#77703](https://github.com/rust-lang/rust/pull/77703)
 - The options `infodir`, `localstatedir`, and `gpg-password-file` are no longer allowed in config.toml. Previously, they were ignored without warning. Note that `infodir` and `localstatedir` are still accepted by `./configure`, with a warning. [#82451](https://github.com/rust-lang/rust/pull/82451)
+- Add options for enabling overflow checks, one for std (`overflow-checks-std`) and one for everything else (`overflow-checks-std`) defaulting to false.
```
Mark-Simulacrum, line 12:
```suggestion
- Add options for enabling overflow checks, one for std (`overflow-checks-std`) and one for everything else (`overflow-checks`) defaulting to false.
```

Please squash this in -- maybe some more adjustments are warranted, sentence feels a little run on to me.

Mark-Simulacrum:
r=me after last nit is fixed

Mark-Simulacrum:
@bors r+

Thanks!

PR #87787, opened 2021-08-05
Use `C-unwind` ABI for `__rust_start_panic` in `panic_abort`

ghost:
The function originally has `C` ABI but is called using `C-unwind` ABI in `std`:
https://github.com/rust-lang/rust/blob/d4ad1cfc63ba5824196bfb2370451ddb5af2e020/library/std/src/panicking.rs#L49-L54
Which might be [problematic](https://github.com/rust-lang/miri/pull/1745#discussion_r596096876) and triggers this [Miri error](https://github.com/rust-lang/rust/issues/87778#issuecomment-893306222):
```
error: Undefined Behavior: calling a function with ABI C using caller ABI C-unwind
   --> /home/hyd-dev/.rustup/toolchains/miri/lib/rustlib/src/rust/library/std/src/panicking.rs:672:9
    |
672 |         __rust_start_panic(obj)
    |         ^^^^^^^^^^^^^^^^^^^^^^^ calling a function with ABI C using caller ABI C-unwind
    |
```
Changing the ABI of the function to `C-unwind` fixes the error above.

### DIFF
--- a/library/panic_abort/src/lib.rs
+++ b/library/panic_abort/src/lib.rs
@@ -15,6 +15,7 @@
 #![feature(staged_api)]
 #![feature(rustc_attrs)]
 #![feature(asm)]
+#![feature(c_unwind)]
 
 #[cfg(target_os = "android")]
 mod android;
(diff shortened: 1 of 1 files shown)


RalfJung:
This sounds like Miri could accept it (though I am not sure of that), but I also don't think there is any reason to use a different ABI string on the two sides here. Thanks!

@bors r+ rollup

PR #86623, opened 2021-06-25
Add check to ensure error code explanations are not removed anymore even if not emitted

GuillaumeGomez:
The error explanations are useful in case you use older version of the compiler. Even more if they had an explanation. If they are not emitted, their explanations should be updated but not removed (as we did for a few of them, like E0001).

r? @Mark-Simulacrum

PR #87792, opened 2021-08-05
Remove git fetch from CI

GuillaumeGomez:
https://github.com/rust-lang/rust/pull/86623 added a call to `git fetch`, which is problematic for releases.

r? @pietroalbini

### DIFF
--- a/src/ci/docker/run.sh
+++ b/src/ci/docker/run.sh
@@ -220,11 +220,13 @@ else
 fi
 
 if [ "$CI" != "" ]; then
-    # Get some needed information for $BASE_COMMIT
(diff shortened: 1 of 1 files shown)


syvb (commented):
I don't think you meant to include this commit, this error shouldn't be removed?

GuillaumeGomez:
This commit is here to make the CI fail. As long as the CI isn't failing, then the PR isn't working.

GuillaumeGomez:
It's working. \o/ Removing the extra commit now...

GuillaumeGomez:
cc @Mark-Simulacrum btw

PR #87794, opened 2021-08-05
Enum should prefer discriminant zero for niche

bonega:
Given an enum with unassigned zero-discriminant, rust should prefer it for niche selection.
Zero as discriminant for `Option<Enum>` makes it possible for LLVM to optimize resulting asm.

- Eliminate branch when expected value coincides.
- Use smaller instruction `test eax, eax` instead of `cmp eax, ?`
- Possible interaction with zeroed memory?

Example:
```rust

pub enum Size {
    One = 1,
    Two = 2,
    Three = 3,
}

pub fn handle(x: Option<Size>) -> u8 {
    match x {
        None => {0}
        Some(size) => {size as u8}
    }
}
```
In this case discriminant zero is available as a niche.

Above example on nightly:
```asm
 mov     eax, edi
 cmp     al, 4
 jne     .LBB0_2
 xor     eax, eax
.LBB0_2:
 ret
```

PR:
```asm
 mov     eax, edi
 ret
```

I created this PR because I had a performance regression when I tried to use an enum to represent legal grapheme byte-length for utf8.

Using an enum instead of `NonZeroU8` [here](https://github.com/bonega/yore/blob/d683304f5dfe2e99f769e6ab8adf8d60a0d1d9b3/src/internal/decoder_incomplete.rs#L90)
resulted in a performance regression of about 5%.
I consider this to be a somewhat realistic benchmark.

Thanks to @ogoffart for pointing me in the right direction!

Edit: Updated description

### DIFF
--- a/compiler/rustc_target/src/abi/mod.rs
+++ b/compiler/rustc_target/src/abi/mod.rs
@@ -1099,19 +1099,53 @@ impl Niche {
         assert!(size.bits() <= 128);
         let max_value = size.unsigned_int_max();
 
-        if count > max_value {
+        let niche = v.end.wrapping_add(1)..v.start;
+        let available = niche.end.wrapping_sub(niche.start) & max_value;
+        if count > available {
             return None;
         }
 
-        // Compute the range of invalid values being reserved.
-        let start = v.end.wrapping_add(1) & max_value;
-        let end = v.end.wrapping_add(count) & max_value;
-
-        if v.contains(end) {
-            return None;
+        // Extend the range of valid values being reserved by moving either `v.start` or `v.end` bound.
+        // Given an eventual `Option<T>`, we try to maximize the chance for `None` to occupy the niche of zero.
+        // This is accomplished by prefering enums with 2 variants(`count==1`) and always taking the shortest path to niche zero.
+        // Having `None` in niche zero can enable some special optimizations.
+        //
+        // Bound selection criteria:
+        // 1. Select closest to zero given wrapping semantics.
+        // 2. Avoid moving past zero if possible.
+        //
+        // In practice this means that enums with `count > 1` are unlikely to claim niche zero, since they have to fit perfectly.
+        // If niche zero is already reserved, the selection of bounds are of little interest.
+        let move_start = |v: WrappingRange| {
+            let start = v.start.wrapping_sub(1) & max_value;
+            Some((start, Scalar { value, valid_range: v.with_start(start) }))
+        };
+        let move_end = |v: WrappingRange| {
+            let start = v.end.wrapping_add(1) & max_value;
+            let end = v.end.wrapping_add(count) & max_value;
+            Some((start, Scalar { value, valid_range: v.with_end(end) }))
+        };
+        let distance_end_zero = max_value - v.end;
+        if v.start > v.end {
+            // zero is unavailable because wrapping occurs
+            move_end(v)
+        } else if v.start <= distance_end_zero {
+            if count <= v.start {
+                move_start(v)
+            } else {
+                // moved past zero, use other bound
+                move_end(v)
+            }
+        } else {
+            let end = v.end.wrapping_add(count) & max_value;
+            let overshot_zero = (1..=v.end).contains(&end);
+            if overshot_zero {
+                // moved past zero, use other bound
+                move_start(v)
+            } else {
+                move_end(v)
+            }
         }
-
-        Some((start, Scalar { value, valid_range: v.with_end(end) }))
     }
 }
 
--- a/src/test/assembly/niche-prefer-zero.rs
+++ b/src/test/assembly/niche-prefer-zero.rs
@@ -1,0 +1,25 @@
+// Check that niche selection prefers zero and that jumps are optimized away.
+// See https://github.com/rust-lang/rust/pull/87794
+// assembly-output: emit-asm
+// only-x86
+// compile-flags: -Copt-level=3
+
+#![crate_type = "lib"]
+
+#[repr(u8)]
+pub enum Size {
+    One = 1,
+    Two = 2,
+    Three = 3,
+}
+
+#[no_mangle]
+pub fn handle(x: Option<Size>) -> u8 {
+    match x {
+        None => 0,
+        Some(size) => size as u8,
+    }
+}
+
+// There should be no jumps in output
+// CHECK-NOT: j
--- a/src/test/ui/enum-discriminant/niche-prefer-zero.rs
+++ b/src/test/ui/enum-discriminant/niche-prefer-zero.rs
@@ -1,0 +1,14 @@
+// Check that niche selection prefers zero.
+// See https://github.com/rust-lang/rust/pull/87794
+// run-pass
+#[repr(u8)]
+pub enum Size {
+    One = 1,
+    Two = 2,
+    Three = 3,
+}
+
+fn main() {
+    // check that `None` is zero
+    assert_eq!(0, unsafe { std::mem::transmute::<Option<Size>, u8>(None) });
+}


joshtriplett:
CI failure looks unrelated.

Once this passes CI, we should do a perf run on it.

bonega:
That will be interesting.

There is still some things that I want input on before merging.

Both problems can be seen in output from this code:
```rust

#[repr(usize)]
pub enum Size {
    One = 1,
    Two = 2,
    Three = 3,
}

pub fn handle(x: Option<Size>) -> usize {
    match x {
        None => {0}
        Some(size) => {size as usize}
    }
}
```


1. Redundant instruction:
```asm
 mov     rax, rdi
 test    rdi, rdi // this instruction is redundant - should not be part of output
 ret
```

2. Missed chance to avoid branch if `repr(usize)` attribute is removed from enum.
```asm
 test    dil, dil
 je      .LBB0_1
 movzx   eax, dil
 ret
.LBB0_1:
 xor     eax, eax
 ret
```
instead of:
```asm
 movzx   eax, dil
 ret
```
My guess is that this is not strongly connected to choice of niche zero.
I will try and look at intermediate code to confirm this.

bonega:
I think that MIR seems agnostic in regard to niche, output is identical between nightly and my PR.
That leaves LLVM-IR or LLVM itself.

1: In LLVM-IR we have:
```
  %1 = icmp eq i64 %0, 0
  br i1 %1, label %bb4, label %bb1

bb1:
  %2 = icmp ult i64 %0, 4
```
So there is a comparison for equal to zero(None) and then another comparison for 0 <= x < 4.
Apparently the branch is optimized away by LLVM, but the comparison against zero remains.

2: Difference between repr(isize/usize) isn't that big.
But we have:
```
%.0 = phi i64 [ %3, %bb1 ], [ 0, %start ]
  ret i64 %.0
```
I think this selects the value conditionally and therefore is our culprit.

For both 1 and 2 I think that all conditional checks should be optimized away at some stage.
Since the bit pattern for the enum in question is guaranteed to be correct? 
(don't know if this should be the area of rustc or LLVM).

imho this is probably not a blocker for this PR, but would be nice to fix.

compiler/rustc_target/src/abi/mod.rs
```diff
@@ -971,8 +971,14 @@ impl Niche {
         }
 
         // Compute the range of invalid values being reserved.
-        let start = v.end().wrapping_add(1) & max_value;
-        let end = v.end().wrapping_add(count) & max_value;
+        let (start, end, zero) = if *v.start() <= *v.end()
+            && *v.end() < (max_value - count)
+            && *v.start() != 0
+        {
+            (max_value - count, 0, true)
```
nbdd0121, line 978: Doesn't this reserve `count + 2` values instead of `count` values?

nbdd0121:
I am a bit worried that this change would reserve a much larger range than it current do. For example, if I have a `253..=253` range, and just need one more discriminant, currently it'll become `253..=254.`, but after this change it'll be `0..=253` (current version of this PR gives `254..=253` with the bug I mentioned above, which is the whole range!) This can potentially have an big impact for nested enums.

Another idea would be just try to creep towards zero instead of forcing zero to be used. We can extend the valid ranges just by `count` in both forward and backward directions, and see which one would take us closer (or include) to zero, and use the valid range extended toward that direction. So `254..=254` would become `254..=255` (and then another nesting takes it to `253..=0`) while `2..=4` become `1..=4` (and another nesting takes it to `0..=5`).

Or maybe we can have some heuristics about not to leave too many gaps in the representation while prefer 0 if possible.

compiler/rustc_target/src/abi/mod.rs
bonega: I think you are right about that

bonega:
You are totally right, I made some stupid mistake.
Also got turned confused by some of the terminology.

I consider it very important to allocate zero for first nesting(much more likely to have an impact).

I think of two alternatives.

1. Heuristics for selecting a bigger range when zero is unallocated(first nesting):
\+ Changes contained in `reserve` function
\- loses some nesting capacity. For anything besides `u8` I think it is very unlikely to be a problem.
Also most `u8` enums will have discriminant range with a small offset from zero.
\- doesn't work for sparse enums, where zero is unallocated but inside discriminant range
\- the optimization will sometimes be a little suprising due to heuristics.

Unanswered question: What is the cutoff point for aggressively allocating zero vs losing nesting capacity?
Is it possible to get any data from a crater(or something) run for niche allocation?
A statistical distribution for this would be great.

2. Track zero allocation. Perhaps Add `zero_reserved: bool` to `Scalar`:
\+ We lose no nesting capacity at all.
\+ Will work for sparse enums, where zero is unallocated but inside discriminant range
\+ Very simple behavior, we know that zero will be allocated if free. There is no heuristics and no consideration for the max value of enum repr
\-  Changes needed elsewhere to allocate zero outside of range(unclear of the scope or impact of this)
\- `Scalar` becomes one byte bigger(did a quick look, think this is ~3%) 40 -> 41 bytes

Because of my limited expertise with the compiler I prefer 1. one at this time.
Route 2 would involve a lot of guiding.

bonega:
@nbdd0121 Did the simple prototype without any heuristics. Just grows towards zero.

I am a bit unsure what good heuristics is because of lack of data.
Will look more at the internals of layout tomorrow.
Also trying to do a perf run on PR

bonega:
I don't understand this failure, from what I can see it shouldn't even use niches.
Also the local file on my system `build/x86_64-unknown-linux-gnu/test/codegen/issue-73338-effecient-cmp/issue-73338-effecient-cmp.ll` doesn't contain br

nbdd0121:
> I don't understand this failure, from what I can see it shouldn't even use niches.
> Also the local file on my system `build/x86_64-unknown-linux-gnu/test/codegen/issue-73338-effecient-cmp/issue-73338-effecient-cmp.ll` doesn't contain br

There is `Option<Ordering>`. In this case your current code uses `-2` for `None` while originally `2` is used.

bonega:
Oh right, derive...
Still I can't reproduce the problem, I agree that the niche has changed from `2` to `-2` but I don't understand why that would cause branching.
`./x.py test` runs without errors on my system.
I have also checked `cargo +stage2 --release --  --emit llvm-ir`, no problem there.

nbdd0121:
Could be with the LLVM version. Rust builds and uses LLVM 12 by default but LLVM 10 and 11 are still supported. The CI is using LLVM 10 according to its name.

bonega:
Thanks for the help.  
Request for review and perf run

bonega:
Will be interesting, I think that there might be a performance regression for LLVM10 fix.
Have to try and run local test for that.

ogoffart:
I wouldn't have expected any diff from the compile-time benchmark:
 - The difference in the arithmetic in Niche::reserve should have a completely negligible impact.
 - The optimization is concerning niches over enums whose discriminant doesn't start at 0. I would think this is a pretty small amount of concerned enum and wouldn't expect to see much impact of that in rustc itself.

So in order to see a difference, we would need a benchmark that specifically call a function like the `handle` function in a loop.

bonega:
Optimization passes are a bit finicky regarding where the niche is growing, sometimes there are surprising results.

I have an example of performance compared to nightly [here] (https://github.com/bonega/yore/commit/b8d6b9ad1035bedfa2f68078d2b05dca1a0d7d5c#diff-2eedbf98fcfc1f3f5c9ff87ee54dfde912667ee80ee09e6bf738806bcaf4cd8b)

group                                                                nightly                                                     enum-niche-opt
-----                                      ------------                           -----------
decode_checked/extended/yore 0.2.0/1024    1.06    753.0±9.20ns    1296.9 MB/sec     1.00    708.1±7.22ns    1379.1 MB/sec
decode_checked/extended/yore 0.2.0/2048    1.07  1451.8±11.74ns  1345.3 MB/sec     1.00  1361.7±42.92ns  1434.4 MB/sec
decode_checked/extended/yore 0.2.0/256      1.05    218.4±2.78ns    1117.7 MB/sec     1.00    207.1±2.86ns    1178.9 MB/sec
decode_checked/extended/yore 0.2.0/4096    1.07      2.8±0.03µs      1407.4 MB/sec     1.00      2.6±0.07µs      1512.5 MB/sec
decode_checked/extended/yore 0.2.0/512      1.05   430.2±22.49ns   1135.1 MB/sec     1.00   411.3±26.53ns    1187.2 MB/sec
decode_checked/extended/yore 0.2.0/64        1.00     83.8±1.64ns     728.1 MB/sec       1.00     83.7±2.49ns     729.2 MB/sec
decode_checked/extended/yore 0.2.0/8          1.00     31.9±1.61ns     239.1 MB/sec       1.00     32.0±1.56ns     238.4 MB/sec

I will do some more experiments in this area, but I consider this ready to merge

joshtriplett:
Seems reasonable to me in terms of language semantics, without commenting on the implementation. I think review of the implementation is up to compiler folks to decide about.

bonega:
This is my first time contributing to rust,
Do I need to add anything or request something to continue with the process? (just want to know if anything more is expected from me at this moment)

jackh726:
Sorry, bit busy recently. Will try to review this soon.

bonega:
Thank you

bonega:
had to rebase for #88242

bonega:
Should have another perf run, but then I consider it ready to merge.

bonega:
parent `05cccdc9b321e6565b3e62e8b52aec53d106ef2f` has no perf run.
This makes it very hard to draw any conclusions.
What to do?

jackh726:
cc @Mark-Simulacrum @jyn514 re. missing perf

Mark-Simulacrum:
Perf seems present now, I think this was just a temporary problem

jackh726:
> I am not super involved in perf, I can't answer questions about it.

whoops, pinged the wrong person...sorry!

compiler/rustc_target/src/abi/mod.rs
```diff
@@ -1031,19 +1031,22 @@ impl Niche {
...
-
-        if v.contains(end) {
+        let niche = v.end.wrapping_add(1)..v.start;
+        let available = niche.end.wrapping_sub(niche.start) & max_value;
+        if count > available {
             return None;
         }
 
-        Some((start, Scalar { value, valid_range: v.with_end(end) }))
+        let (start, valid_range) = if v.start <= max_value - v.end && v.start >= count {
```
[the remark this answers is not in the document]
eddyb: I guess what I was missing was the idea that it "preferred closeness (i.e. the smaller space)" as opposed to "preferring the larger space".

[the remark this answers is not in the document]
eddyb: I am going to unmark this as resolved only for the fact that I think the *intention* needs to be in the comment, since I don't think it's obvious from the code without already having been told the intention.

This could also be partly done by using explanative variables, e.g.:
```rust
let dist_to_0_below = v.start;
let dist_to_0_above = max_value - v.end + 1;

// Prefer the direction that reaches `0` quicker.
if dist_to_0_below < dist_to_0_above && v.start >= count {...}
```

But, what if it's a wraparound range? The "distance to 0" stuff becomes meaningless, since `0` is a valid value, and can never be reached. I guess `v.start >= count` isn't required for correctness, because of the earlier check, but it still feels weird to not have `v.start < v.end` as a condition.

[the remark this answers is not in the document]
bonega: I understand your concern about `v.start < v.end`.
It is possible to just add it in the if expression, I just feel bad about it when it is not having any influence on the outcome.
I added a more thorough explanation and introduced some helper variables. 
https://github.com/rust-lang/rust/pull/87794/commits/4e1f80cb6a75c8f81acc5292b6245ff03cd07f54

jackh726:
I'm going to r? @eddyb here, but feel free to reassign; I'm probably not the best to review here.

eddyb (approved):
LGTM, though I'll start another perf run to have some up-to-date (and categorized) results.
(**EDIT**: forgot to say, a LLVM IR / assembly test would be nice to have, to show the codegen improvement)

compiler/rustc_target/src/abi/mod.rs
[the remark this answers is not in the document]
bonega: I have commited https://github.com/rust-lang/rust/pull/87794/commits/52f4be95a8d488b60543993a526fbc3429b65ec8
Changed so it checks for `v.start > v.end`.
It avoids moving past zero for `v.end`

eddyb:
Also, I'm not sure how to tackle the decision of landing a change that affects codegen (even if positively), so for that I'm deferring to:

r? @nagisa cc @wesleywiser

bonega:
A test for assembly output, is the right place for that in `src/test/assembly/`?
This are my cherry-picked results

nightly:
```asm
 mov     eax, edi
 cmp     al, 4
 jne     .LBB0_2
 xor     eax, eax
.LBB0_2:
 ret
```

PR:
```asm
 mov     eax, edi
 ret
```

`CHECK-NOT jne`
Perhaps?

nagisa:
> A test for assembly output, is the right place for that in `src/test/assembly/`?

Yes.

> `CHECK-NOT jne` Perhaps?

That works. Looking at the differences for other architectures (e.g. AArch64?) may give more opportunities to test the expectations here too.

---

A couple of procedural nitpicks: the commits will need to be squashed together to form a meaningful change history (i.e. no "fixup"-like commits); please remove the `WIP` from the PR description as the text will end up in the merge commit.

hkratz:
Maybe an LLVM-IR check like the one in [src/test/codegen/vec-iter-collect-len.rs](https://github.com/rust-lang/rust/blob/673d0db5e393e9c64897005b470bfeb6d5aec61b/src/test/codegen/vec-iter-collect-len.rs) would be better?

nagisa:
Well, if the goal is to test the assembly quality here, LLVM-IR is a pretty poor proxy for this. LLVM-IR would basically only allow you to see if the discriminant of `Option<Enum>` is a 0 or something else, which one could also do just as well in a simple run-pass test with a transmute.

hkratz:
We can check that the comparison is optimized away, e.g. with `CHECK-NOT icmp` at the end of `handle()` or am I missing something?

The advantage is that this does not require separate tests for aarch64, etc.

bonega:
@nagisa Thanks for the tips!
I will rebase and squash once bors land #88327

@hkratz I am very bad at llvm-ir, prefer asm because of that.

Added tests

bonega:
Rebased and squashed
Hopefully ready for merge :crossed_fingers:

src/test/assembly/niche-prefer-zero.rs
```diff
@@ -0,0 +1,29 @@
+// Check that niche selection prefers zero.
+// See https://github.com/rust-lang/rust/pull/87794
+// revisions: x64 A64 ppc64le
```
nagisa, line 3: You will need to specify the `--target=$TARGET` flags for each of the architectures you're testing via `[$REVISION]compile-flags`. `revision` by itself does not have any meaning whatsoever.

In order to enable this, you may will also need to build for `#[no_core]` like some other assembly tests do. Or you could make this a `only-x86` test instead.

src/test/assembly/niche-prefer-zero.rs
```diff
@@ -0,0 +1,29 @@
...
+pub fn handle(x: Option<Size>) -> u8 {
+    match x {
+        None => 0,
+        Some(size) => size as u8,
+    }
+}
+
+// There should be no jumps in output
+// CHECK-LABEL: handle:
+// x64: movl
```
nagisa, line 24: This doesn't ensure that there won't be jumps between in between `handle`, `movl` and `retq`. You want `CHECK-NOT` (see FileCheck documentation for a guide [here](https://llvm.org/docs/CommandGuide/FileCheck.html)).

src/test/ui/enum-discriminant/niche-prefer-zero.rs
```diff
@@ -0,0 +1,13 @@
+// Check that niche selection prefers zero.
+// See https://github.com/rust-lang/rust/pull/87794
+// run-pass
+pub enum Size {
```
nagisa, line 4: Will probably need to `repr(u8)` here. If my memory serves me right there are some architectures that have a minimum "default" enum size which is not u8.

bonega: *Mind blown*
Makes sense though

src/test/assembly/niche-prefer-zero.rs
bonega: Thanks for the reference.
Reason I didn't use `CHECK-NOT` is that I didn't understand how to combine it with the other targets.
Only target x64 now.

src/test/assembly/niche-prefer-zero.rs
bonega: Going for `only-x86` instead.
I don't even now how to test the others

bonega:
So I think this failed during a non-optimizing build.
I don't know how to specify that the asm-test should only run for an optimized build

nagisa:
>  I don't know how to specify that the asm-test should only run for an optimized build

Perhaps `// compile-flags: -Copt-level=3` or `// compile-flags: -O`? Additionally, I suspect that just `// only-x86` is sufficient and `// only-x86_64-unknown-linux-gnu` is likely way too specific.

bonega:
Thanks a lot to @nagisa, @eddyb and @nbdd0121 and others for your help

PR #87804, opened 2021-08-05
Use box syntax everywhere

est31:
Alternative PR to #87781 .

As suggested by @leonardo-m :https://github.com/rust-lang/rust/pull/87781#issuecomment-893299205

### DIFF
--- a/compiler/rustc_middle/src/mir/interpret/mod.rs
+++ b/compiler/rustc_middle/src/mir/interpret/mod.rs
@@ -49,7 +49,7 @@ macro_rules! err_exhaust {
 #[macro_export]
 macro_rules! err_machine_stop {
     ($($tt:tt)*) => {
-        $crate::mir::interpret::InterpError::MachineStop(Box::new($($tt)*))
+        $crate::mir::interpret::InterpError::MachineStop(box ($($tt)*))
     };
 }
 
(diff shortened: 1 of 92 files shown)


compiler/rustc_middle/src/mir/interpret/mod.rs
```diff
@@ -49,7 +49,7 @@ macro_rules! err_exhaust {
 #[macro_export]
 macro_rules! err_machine_stop {
     ($($tt:tt)*) => {
-        $crate::mir::interpret::InterpError::MachineStop(Box::new($($tt)*))
+        $crate::mir::interpret::InterpError::MachineStop(box ($($tt)*))
```
RalfJung, line 52: This will mean other crates using the macro (such as standalone Miri) need to also enable the `box` feature, right? I think I'd prefer to avoid that.

bjorn3: There exists `#[allow_internal_unstable]` to prevent that.

bjorn3:
This is a slight regression of up to 0.3%. No improvements.

petrochenkov:
I assume this was opened only to measure performance, without intent to actually merge the changes, so closing.

compiler/rustc_middle/src/mir/interpret/mod.rs
est31: Example macro that does this for box syntax is the [vec macro](https://doc.rust-lang.org/1.54.0/src/alloc/macros.rs.html#42). As @petrochenkov correctly assumed, this PR was a test run though to explore the implications of using box syntax everywhere.

PR #87810, opened 2021-08-06
current_exe haiku code path simplification all of these part of libc

devnexen:


library/std/src/sys/unix/os.rs
```diff
@@ -388,46 +388,22 @@ pub fn current_exe() -> io::Result<PathBuf> {
...
+            &mut cookie,
+            info.as_mut_ptr(),
+            mem::size_of::<libc::image_info>(),
+        );
         if result != 0 {
             use crate::io::ErrorKind;
             Err(io::Error::new_const(ErrorKind::Uncategorized, &"Error getting executable path"))
         } else {
-            let name = CStr::from_ptr(info.name.as_ptr()).to_bytes();
+            info.assume_init();
```
bjorn3, line 405: This is a no-op. `info.assume_init()` returns a value it doesn't do anything in-place. In this particular case I don't think it is needed at all though.

devnexen: verified locally indeed not needed

PR #87818, opened 2021-08-06
Fix anchors display in rustdoc

### LINKED ISSUE #87611: rustdoc: Section icon covers part of enum variant name when hovered
**This is a regression from 1.53.0 to 1.54.0 (released today).** It's likely that the display of other parts of the docs has regressed in a similar way as well (e.g., `struct` fields), but I haven't checked those.

More broadly, the way the variant looks when hovered has changed as well. I am unsure if that part of the change was intentional; to me it seems like a regression.

### [1.54.0](https://doc.rust-lang.org/1.54.0/std/option/enum.Option.html#variant.None) (incorrect)

image

### [1.53.0](https://doc.rust-lang.org/1.53.0/std/option/enum.Option.html#variant.None) (correct)

image

GuillaumeGomez:
Fixes https://github.com/rust-lang/rust/issues/87611 (it simplifies the positioning and fix the background).

Screenshot from 2021-08-06 16-47-03
Screenshot from 2021-08-06 16-47-10
Screenshot from 2021-08-06 16-47-14

r? @camelid

### DIFF
--- a/src/librustdoc/html/static/css/rustdoc.css
+++ b/src/librustdoc/html/static/css/rustdoc.css
@@ -735,17 +735,21 @@ a {
 .anchor {
 	display: none;
 	position: absolute;
-	left: -7px;
+	left: 0;
+	background: none !important;
 }
 .anchor.field {
 	left: -5px;
 }
 .small-section-header > .anchor {
-	left: -28px;
-	padding-right: 10px; /* avoid gap that causes hover to disappear */
+	left: -15px;
+	padding-right: 8px;
 }
-.anchor:before {
-	content: '\2002\00a7\2002';
+h2.small-section-header > .anchor {
+	padding-right: 6px;
+}
+.anchor::before {
+	content: '§';
 }
 
 .docblock a:not(.srclink):not(.test-arrow):hover,


src/librustdoc/html/static/css/rustdoc.css
```diff
@@ -735,17 +735,21 @@ a {
...
+	left: -15px;
+	padding-right: 8px;
 }
-.anchor:before {
-	content: '\2002\00a7\2002';
+h2.small-section-header > .anchor {
+	padding-right: 2px;
+}
+.anchor::before {
+	content: '§';
```
camelid, line 752: What's the difference between `.anchor:before` and `.anchor::before`? Is there a reason you changed it?

Also, you changed `'\2002\00a7\2002'` to `'§'`—are they equivalent?

camelid: Ok, IIUC, `:before` is technically incorrect according to the spec and `::before` is the spec-correct way, but in practice they should behave the same. Still curious about the other change though, especially because Python says the bytes that were there before aren't valid UTF-8 or UTF-16.

GuillaumeGomez: They're not equivalent. `\2002` is a whitepace. It made the anchor positioning thing more difficult to handle.

camelid: Ah, makes sense 👍

GuillaumeGomez:
@camelid Is it good for you or is there something else to change?

camelid:
I'll try to review this by Wednesday.

jyn514:
Oh sorry, github glitched and didn't show me your last comment.

@bors r-

src/librustdoc/html/static/css/rustdoc.css
```diff
@@ -735,17 +735,21 @@ a {
...
-	left: -7px;
+	left: 0;
+	background: none !important;
 }
 .anchor.field {
 	left: -5px;
 }
 .small-section-header > .anchor {
-	left: -28px;
-	padding-right: 10px; /* avoid gap that causes hover to disappear */
```
camelid, line 745: I tried out this change locally, and it all looks good, except that the gap that causes the hover to disappear has cropped up in the new version. Is it possible to get rid of the gap?

camelid: The gap only appears with headers like `Implementations`, not with enum variants.

src/librustdoc/html/static/css/rustdoc.css
```diff
@@ -735,17 +735,21 @@ a {
...
+	left: -15px;
+	padding-right: 8px;
 }
-.anchor:before {
-	content: '\2002\00a7\2002';
+h2.small-section-header > .anchor {
+	padding-right: 2px;
```
camelid, line 749: It looks like maybe it could be fixed by increasing this padding?

camelid: It seems like just increasing this to 6px fixes the problem.

```suggestion
	padding-right: 6px;
```

camelid:
r=me after something like https://github.com/rust-lang/rust/pull/87818#discussion_r690665642

GuillaumeGomez:
Done! Interestingly enough, I couldn't reproduce the issue on firefox.

camelid:
> Done! Interestingly enough, I couldn't reproduce the issue on firefox.

Hmm, I _found_ the issue using Firefox. I discovered it by moving my mouse somewhat slowly—it doesn't happen when my mouse is moved quickly—so maybe when you were trying to reproduce it you were moving your mouse too fast?

GuillaumeGomez:
So my superhuman speed was the issue! :rofl:

PR #87819, opened 2021-08-06
Use a more accurate span on assoc types WF checks

estebank:
Before
```
error[E0275]: overflow evaluating the requirement `<FooStruct as Foo>::A == _`
  --> $DIR/issue-21946.rs:8:5
   |
LL |     type A = <FooStruct as Foo>::A;
   |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```
after
```
error[E0275]: overflow evaluating the requirement `<FooStruct as Foo>::A == _`
  --> $DIR/issue-21946.rs:8:14
   |
LL |     type A = <FooStruct as Foo>::A;
   |              ^^^^^^^^^^^^^^^^^^^^^
```


compiler/rustc_typeck/src/check/wfcheck.rs
```diff
@@ -194,12 +194,13 @@ pub fn check_trait_item(tcx: TyCtxt<'_>, def_id: LocalDefId) {
     let hir_id = tcx.hir().local_def_id_to_hir_id(def_id);
     let trait_item = tcx.hir().expect_trait_item(hir_id);
 
-    let method_sig = match trait_item.kind {
-        hir::TraitItemKind::Fn(ref sig, _) => Some(sig),
-        _ => None,
+    let (method_sig, ty) = match trait_item.kind {
```
jackh726, line 197: Why not return selectively use eitehr `trait_item.span` or `ty.span` here? Then you don't need to pass through `ty`?

compiler/rustc_typeck/src/check/wfcheck.rs
```diff
@@ -268,12 +269,13 @@ pub fn check_impl_item(tcx: TyCtxt<'_>, def_id: LocalDefId) {
...
-        hir::ImplItemKind::Fn(ref sig, _) => Some(sig),
-        _ => None,
+    let (method_sig, ty) = match impl_item.kind {
+        hir::ImplItemKind::Fn(ref sig, _) => (Some(sig), None),
+        hir::ImplItemKind::TyAlias(ty) => (None, Some(ty)),
+        _ => (None, None),
     };
 
-    check_associated_item(tcx, impl_item.hir_id(), impl_item.span, method_sig);
+    check_associated_item(tcx, impl_item.hir_id(), impl_item.span, method_sig, ty);
```
jackh726, line 278: And here...

jackh726 (changes requested):

compiler/rustc_typeck/src/check/wfcheck.rs
```diff
@@ -193,13 +193,18 @@ pub fn check_item_well_formed(tcx: TyCtxt<'_>, def_id: LocalDefId) {
...
+    let mut span = trait_item.span;
 
     let method_sig = match trait_item.kind {
         hir::TraitItemKind::Fn(ref sig, _) => Some(sig),
+        hir::TraitItemKind::Type(_bounds, Some(ty)) => {
+            span = ty.span;
+            None
+        }
         _ => None,
     };
```
jackh726, line 205: Nit but this could be
```rust
let (method_sig, span) = match trait_item.kind {
    hir::TraitItemKind::Fn(ref sig, _) => (Some(sig), trait_item.span),
    hir::TraitItemKind::Type(_bounds, Some(ty)) => (None, ty.span),
    _ => (None, trait_item.span),
};
```
and same below

jackh726 (approved):
r=me with or without nit

JohnTitor:
#87811 added one more test that needs a change to span: https://github.com/rust-lang/rust/pull/87908#issuecomment-895724624
@bors r-

PR #87823, opened 2021-08-06
Add Infallible::into_any

joshlf:
This change adds `Infallible::into_any`, which has the signature `fn into_any<T>(Infallible) -> T`. It is useful in helping unreachable code type check.

### DIFF
--- a/library/core/src/convert/mod.rs
+++ b/library/core/src/convert/mod.rs
@@ -678,6 +678,16 @@ impl AsMut<str> for str {
 #[derive(Copy)]
 pub enum Infallible {}
 
+impl Infallible {
+    /// Converts an [`Infallible`] to an instance of any type.
+    ///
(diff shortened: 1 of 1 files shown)


joshlf:
I assume there's a good chance there will be pushback on this, so I'll address the build failures if it's decided that we actually want to move forward with this.

On that subject: What's the right stability attribute to use here?

yaahc:
> On that subject: What's the right stability attribute to use here?

It should be unstable and have an associated tracking issue. Also we will probably want to check in with lang since this is so closely related to the never type.

cc @joshtriplett

m-ou-se:
The `Infallible` type will be replaced by an alias for `!` at some point in the future. So we should be careful with its api.

joshlf:
Given the replacement by `!`, I think it makes sense to just hold off on this. It's trivial to implement by hand when needed.

PR #86699, opened 2021-06-28
Allow reifying intrinsics to `fn` pointers.

eddyb:
The main change is in `ty/instance.rs`, which is to simply have `resolve_for_fn_ptr` handle `InstanceDef::Intrinsic` the same way it does the other existing cases where a shim is needed to be able to have a `fn` pointer in the first place (virtual methods i.e. `<dyn Trait as Trait>::method`, and `#[track_caller]` functions).

However, there is another important change, which is that unlike other functions, intrinsics have their type-level ABI change from the `extern "rust-intrinsic" fn` "pseudo-ABI" to the normal Rust ABI.
This is possible because pseudo-ABIs like `extern "rust-intrinsic"` do not actually change the call ABI, but rather are used as a side-channel (instead of a more principled solution, like some `#[rustc_intrinsic]` attribute).

What this all means for the user is that they can/have to do `transmute as unsafe fn(T) -> U`, without the fact that the function was an intrinsic showing up in the typesystem at all.

That type-level ABI aspect also ties in with the motivation for finally fixing this, which is #84297, where `ptr::{copy,copy_nonoverlapping}` becoming stable reexports of the intrinsics in #81238 was backwards incompatible - with this PR, that shouldn't be an issue anymore.

(I also have a revert for #86003 ready - at least the parts not reverted already by e.g. #86295, if this PR is accepted; cc @pnkfelix)

<hr/>

In terms of this capability being insta-stable, we still get to control that via which intrinsics we mark stable.
The only one I could find right now is `mem::transmute` (and I've added a test that its size check still works when going through a `fn` pointer), and if we do the above mentioned #86003 revert, we'd also have `ptr::{copy,copy_nonoverlapping}`.

PR #86003, opened 2021-06-04
Make copy/copy_nonoverlapping fn's again

pnkfelix:
Make copy/copy_nonoverlapping fn's again, rather than intrinsics.

This a short-term change to address issue #84297.

It effectively reverts PRs #81167 #81238 (and part of #82967), #83091, and parts of #79684.

PR #81238, opened 2021-01-21
directly expose copy and copy_nonoverlapping intrinsics

RalfJung:
This effectively un-does https://github.com/rust-lang/rust/pull/57997. That should help with `ptr::read` codegen in debug builds (and any other of these low-level functions that bottoms out at `copy`/`copy_nonoverlapping`), where the wrapper function will not get inlined. See the discussion in https://github.com/rust-lang/rust/pull/80290 and https://github.com/rust-lang/rust/issues/81163.

Cc @bjorn3 @therealprof

PR #87827, opened 2021-08-06
Avoid using the `copy_nonoverlapping` wrapper through `mem::replace`.

eddyb:
This is a much simpler way to achieve the pre-#86003 behavior of `mem::replace` not needing dynamically-sized `memcpy`s (at least before inlining), than re-doing #81238 (which needs #86699 or something similar).

I didn't notice it until recently, but `ptr::write` already explicitly avoided using the wrapper, while `ptr::read` just called the wrapper (and was the reason for us observing any behavior change from #86003 in Rust-GPU).

<hr/>

The codegen test I've added fails without the change to `core::ptr::read` like this (ignore the `v0` mangling, I was using a worktree with it turned on by default, for this):
```llvm
       13: ; core::intrinsics::copy_nonoverlapping::<u8>
       14: ; Function Attrs: inlinehint nonlazybind uwtable
       15: define internal void @_RINvNtCscK5tvALCJol_4core10intrinsics19copy_nonoverlappinghECsaS4X3EinRE8_25mem_replace_direct_memcpy(i8* %src, i8* %dst, i64 %count) unnamed_addr #0 {
       16: start:
       17:  %0 = mul i64 %count, 1
       18:  call void @llvm.memcpy.p0i8.p0i8.i64(i8* align 1 %dst, i8* align 1 %src, i64 %0, i1 false)
not:17      !~~~~~~~~~~~~~~~~~~~~~                                                                     error: no match expected
       19:  ret void
       20: }
```
With the `core::ptr::read` change, `core::intrinsics::copy_nonoverlapping` doesn't get instantiated and the test passes.

<hr/>

r? @m-ou-se cc @nagisa (codegen test) @oli-obk / @RalfJung (miri diagnostic changes)

### DIFF
--- a/library/core/src/ptr/mod.rs
+++ b/library/core/src/ptr/mod.rs
@@ -685,6 +685,13 @@ pub const unsafe fn replace<T>(dst: *mut T, mut src: T) -> T {
 #[stable(feature = "rust1", since = "1.0.0")]
 #[rustc_const_unstable(feature = "const_ptr_read", issue = "80377")]
 pub const unsafe fn read<T>(src: *const T) -> T {
+    // We are calling the intrinsics directly to avoid function calls in the generated code
+    // as `intrinsics::copy_nonoverlapping` is a wrapper function.
+    extern "rust-intrinsic" {
+        #[rustc_const_unstable(feature = "const_intrinsic_copy", issue = "80697")]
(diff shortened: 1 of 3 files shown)


nagisa:
Looks reasonable enough to me.

RalfJung:
Yeah this makes sense.

PR #87830, opened 2021-08-06
Suggest replacing braces for brackets on array-esque invalid block expr

### LINKED ISSUE #87672: Better Diagnostics When Using Wrong Brackets to Create an Array
I ended up spending way too much time debugging the following error, since I was trying to port some C code to rust.
```rust
const ARR: [u8; 3] = {
    1, 2, 3
}
```
Give the error:
```
error: expected one of `.`, `;`, `?`, `}`, or an operator, found `,`
 --> src/lib.rs:2:6
  |
2 |     1, 2, 3,
  |      ^ expected one of `.`, `;`, `?`, `}`, or an operator

```
The error message is very vague and hard to determine the actual error here.
Seeing as though some programming languages like Java and C use `{}` for arrays, I think it'd make sense for the compiler to suggest using `[]` if the user is trying to create an array.

hkmatsumoto:
Newcomers may write `{1, 2, 3}` for making arrays, and the current error message is not informative enough to quickly convince them what is needed to fix the error.

This PR implements a diagnostic for this case, and its output looks like this:
```text
error: this code is interpreted as a block expression, not an array
 --> src/lib.rs:1:22
  |
1 |   const FOO: [u8; 3] = {
  |  ______________________^
2 | |     1, 2, 3
3 | | };
  | |_^
  |
  = note: to define an array, one would use square brackets instead of curly braces
help: try using [] instead of {}
  |
1 | const FOO: [u8; 3] = [
2 |     1, 2, 3
3 | ];
  |
```

Fix #87672

### DIFF
--- a/compiler/rustc_parse/src/parser/expr.rs
+++ b/compiler/rustc_parse/src/parser/expr.rs
@@ -1204,7 +1204,7 @@ impl<'a> Parser<'a> {
         } else if self.check(&token::BinOp(token::Or)) || self.check(&token::OrOr) {
             self.parse_closure_expr(attrs)
         } else if self.check(&token::OpenDelim(token::Bracket)) {
-            self.parse_array_or_repeat_expr(attrs)
+            self.parse_array_or_repeat_expr(attrs, token::Bracket)
         } else if self.check_path() {
             self.parse_path_start_expr(attrs)
         } else if self.check_keyword(kw::Move) || self.check_keyword(kw::Static) {
@@ -1322,11 +1322,15 @@ impl<'a> Parser<'a> {
         self.maybe_recover_from_bad_qpath(expr, true)
     }
 
-    fn parse_array_or_repeat_expr(&mut self, attrs: AttrVec) -> PResult<'a, P<Expr>> {
+    fn parse_array_or_repeat_expr(
+        &mut self,
+        attrs: AttrVec,
+        close_delim: token::DelimToken,
+    ) -> PResult<'a, P<Expr>> {
         let lo = self.token.span;
-        self.bump(); // `[`
+        self.bump(); // `[` or other open delim
 
-        let close = &token::CloseDelim(token::Bracket);
+        let close = &token::CloseDelim(close_delim);
         let kind = if self.eat(close) {
             // Empty vector
             ExprKind::Array(Vec::new())
@@ -1752,6 +1756,46 @@ impl<'a> Parser<'a> {
         }
     }
 
+    fn is_array_like_block(&mut self) -> bool {
+        self.look_ahead(1, |t| matches!(t.kind, TokenKind::Ident(..) | TokenKind::Literal(_)))
+            && self.look_ahead(2, |t| t == &token::Comma)
+            && self.look_ahead(3, |t| t.can_begin_expr())
+    }
+
+    /// Emits a suggestion if it looks like the user meant an array but
+    /// accidentally used braces, causing the code to be interpreted as a block
+    /// expression.
+    fn maybe_suggest_brackets_instead_of_braces(
+        &mut self,
+        lo: Span,
+        attrs: AttrVec,
+    ) -> Option<P<Expr>> {
+        let mut snapshot = self.clone();
+        match snapshot.parse_array_or_repeat_expr(attrs, token::Brace) {
+            Ok(arr) => {
+                let hi = snapshot.prev_token.span;
+                self.struct_span_err(
+                    arr.span,
+                    "this code is interpreted as a block expression, not an array",
+                )
+                .multipart_suggestion(
+                    "try using [] instead of {}",
+                    vec![(lo, "[".to_owned()), (hi, "]".to_owned())],
+                    Applicability::MaybeIncorrect,
+                )
+                .note("to define an array, one would use square brackets instead of curly braces")
+                .emit();
+
+                *self = snapshot;
+                Some(self.mk_expr_err(arr.span))
+            }
+            Err(mut e) => {
+                e.cancel();
+                None
+            }
+        }
+    }
+
     /// Parses a block or unsafe block.
     pub(super) fn parse_block_expr(
         &mut self,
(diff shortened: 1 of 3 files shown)


Mark-Simulacrum:
@rust-timer build 3337a035d2b47609aaf087adc04a0cbba7e3beb0

compiler/rustc_parse/src/parser/diagnostics.rs
```diff
@@ -409,6 +409,103 @@ impl<'a> Parser<'a> {
...
+        // As a workaround, we do nothing for `token::OpenDelim(token::Brace)`
+        // in the match expression below, just waiting for
+        // `token::CloseDelim(token::Brace)` to exit the loop.
+        let mut depth = [
+            1, // {}
+            0, // <>, this appears in fn calls. e.g., {foo::<A, B>(), 2, 3}
+        ];
+
+        // We keep calling `look_ahead` until we reach the end of the block expression.
+        loop {
```
estebank, line 444: I have a [somewhat different approach](https://github.com/rust-lang/rust/pull/88729) that might be able to make some of this code easier to reason about.

Instead of using `look_ahead` for cases where we kind of need to parse things with non-deterministic lookahead, I do `let snapthot = self.clone()` on the `Parser` and *try to successfully parse* what the "fake" syntax would be. If that succeeds, we know that we can provide an accurate suggestion, if that doesn't, we cancel the error and rollback the parser to the previous state to emit the "normal" error we currently see and let the existing error recovery machinery take its course. With this approach, it should be possible to emit a suggestion on a case like `{0, if 1 < 2 {1} else {1}, 2, 3}`. Because cloning is somewhat expensive (but not terrible, from what I've seen), we want to avoid it on the "happy path" of well-formed code, but you can use limited `look_ahead` to see if *it could be* this case, like doing `self.look_ahead(2, |t| t.kind == &Comma)` or something.

We could even try to keep information on the `Parser` for "we are in a `const` assignment, of array type" that we check here in `maybe_suggest_brackets_instead_of_braces`, and if it is not the case, we exit early, like we do for things coming from a macro.

What do you think? (Sorry for the delay in reviewing)

src/test/ui/parser/bare-struct-body.stderr
```diff
@@ -14,6 +14,23 @@ LL |     val: (),
...
+LL | |     };
+   | |_____^
+   |
+   = note: to define an array, one would use square brackets instead of curly braces
+help: try using [] instead of {}
+   |
+LL |     let x = [
+LL |         val: (),
+LL |     ];
+   |
```
estebank, line 32: We probably wouldn't want this error or the one above, as we are emitting two diagnostics for the same syntax error. `maybe_suggest_brackets_instead_of_braces` doesn't signal whether this worked or not, so instead what it could do is return a `PResult` and you can call `self.maybe_suggest_brackets_instead_of_braces(lo)?` instead.

compiler/rustc_parse/src/parser/diagnostics.rs
hkmatsumoto: It's very smart of you to come up with the approach :)
What's nice is that by cloning the parser the codebase will be significantly simplified while the suggestion keeps working almost as-is except some cases like `{if 1 < 2 {1} else {1}, 2, 3}`, but that's compromisable as I assume it's unlikely to initialize an array with complex expressions.

src/test/ui/parser/bare-struct-body.stderr
hkmatsumoto: By actually parsing as if they were arrays, this is no longer emitted.

compiler/rustc_parse/src/parser/diagnostics.rs
estebank: Thank you, you're too kind.

Yes, I agree that this gets 80% there for 20% of the effort. If we were relying on this for the actual language and not just for error recovery, the story would be different.

compiler/rustc_parse/src/parser/expr.rs
```diff
@@ -1752,6 +1756,36 @@ impl<'a> Parser<'a> {
...
+    fn is_array_like_block(&mut self) -> bool {
+        self.look_ahead(1, |t| matches!(t.kind, TokenKind::Ident(..) | TokenKind::Literal(_)))
+            && self.look_ahead(2, |t| t == &token::Comma)
+            && self.look_ahead(3, |t| t.can_begin_expr())
+    }
+
+    /// Emits a suggestion if it looks like the user meant an array but
+    /// accidentally used braces, causing the code to be interpreted as a block
+    /// expression.
+    fn maybe_suggest_brackets_instead_of_braces(&mut self, lo: Span, attrs: AttrVec) {
```
estebank, line 1768: I would change this to return a `PResult<'a, P<Expr>>`...

compiler/rustc_parse/src/parser/expr.rs
```diff
@@ -1752,6 +1756,36 @@ impl<'a> Parser<'a> {
...
+                    arr.span,
+                    "this code is interpreted as a block expression, not an array",
+                )
+                .multipart_suggestion(
+                    "try using [] instead of {}",
+                    vec![(lo, "[".to_owned()), (hi, "]".to_owned())],
+                    Applicability::MaybeIncorrect,
+                )
+                .note("to define an array, one would use square brackets instead of curly braces")
+                .emit();
```
estebank, line 1783:
```suggestion
                .emit();
                *self = snapshot;
                Some(self.mk_expr_err(/*I think it takes a span here*/))
```

compiler/rustc_parse/src/parser/expr.rs
```diff
@@ -1752,6 +1756,36 @@ impl<'a> Parser<'a> {
...
+            }
+            Err(mut e) => e.emit(),
```
estebank, line 1785: We don't want to emit parse errors caused by "alternative syntax".

```suggestion
            Err(mut e) => {
                e.cancel();
                None
            }
```

compiler/rustc_parse/src/parser/expr.rs
```diff
@@ -1760,6 +1794,10 @@ impl<'a> Parser<'a> {
         blk_mode: BlockCheckMode,
         mut attrs: AttrVec,
     ) -> PResult<'a, P<Expr>> {
+        if self.is_array_like_block() {
+            self.maybe_suggest_brackets_instead_of_braces(lo, attrs.clone());
```
estebank, line 1798:
```suggestion
            if let Some(expr) = self.maybe_suggest_brackets_instead_of_braces(lo, attrs.clone()) {
                return Ok(expr);
            }
```

estebank (commented):
Let's add a test as well :)

compiler/rustc_parse/src/parser/expr.rs
hkmatsumoto: Oops, you're definitely right!

PR #87836, opened 2021-08-07
Change proc_macro::Diagnostics docs

tsoutsman:
With Rust 1.54 attributes can invoke function-like macros. This allows functions generated by macros to have non-generic documentation. This pull request changes the documentation of `proc_macro::Diagnostics` to be more specific and include a link to the level.
## Example
Before:
```markdown
Adds a new child diagnostic message to `self` with the level
identified by this method’s name with the given `message`.
```
After:
```markdown
Adds a new child diagnostic message to self with the [`Level::Error`]
level, and the given `message`.
```


petrochenkov:
Could you wrap the lines to fit into the usual 100-character limit? `rustfmt` cannot do it automatically in this case. 

r=me after doing that and squashing commits.

TypicalFork:
Hi, I'm not quite sure about the indentation so I just put it in line with `concat!`
r= @petrochenkov

petrochenkov:
Thanks!
@bors r+

petrochenkov:
@tsoutsman 
Push to a pull request after approval (`@bors r+`) cancels the approval.
Also using rebases instead of merges is recommended to update pull request branches.
@bors r+

tsoutsman:
> @tsoutsman 
> Push to a pull request after approval (`@bors r+`) cancels the approval.
> Also using rebases instead of merges is recommended to update pull request branches.

Sorry, I was just playing around with the repository. I'm not too familiar with git.

PR #87838, opened 2021-08-07
Document that fs::read_dir skips . and ..

jetomit:
Hi,

I think this is worth noting in the docs since it differs from POSIX `readdir`. I didn’t put it under platform-specific notes because it seems to be consistent across platforms, and changing this behavior in the future could cause pretty nasty bugs.

Thanks!


dtolnay (approved):
Thanks!

PR #87842, opened 2021-08-07
Fix intra doc link in hidden doc of Iterator::__iterator_get_unchecked

steffahn:
Recently, I edited the import list of the `core::iter::traits::iterator` module (in #85874). This results in a broken intra doc link in a hidden documentation with the effect that `RUSTDOCFLAGS='--document-private-items --document-hidden-items' x doc library/std` fails. (This can be worked around by adding `-Arustdoc::broken-intra-doc-links`; still, it’s a broken link so let’s fix it.)

@rustbot label C-cleanup, T-libs

### DIFF
--- a/library/core/src/iter/traits/iterator.rs
+++ b/library/core/src/iter/traits/iterator.rs
@@ -3452,7 +3452,7 @@ pub trait Iterator {
         self.map(f).is_sorted()
     }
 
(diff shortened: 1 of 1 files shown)


dtolnay (approved):
Thanks!

PR #87843, opened 2021-08-07
TryReserveErrorKind tests and inline

kornelski:
A small follow-up to #87408


the8472:
I didn't explicitly mention it, the `kind()` method probably needs an inline too, for the same reason.

kornelski:
kind already has inline

m-ou-se:
Thanks!

@bors r+

m-ou-se:
@bors rollup=never

(Because of https://github.com/rust-lang/rust/pull/87408#issuecomment-896941643)

PR #73566, opened 2020-06-20
Don't run `everybody_loops` for rustdoc; instead ignore resolution errors

jyn514:
r? @eddyb 
cc @petrochenkov, @GuillaumeGomez, @Manishearth, @ecstatic-morse, @marmeladema 

~~Blocked on https://github.com/rust-lang/rust/pull/73743~~ Merged.
~~Blocked on crater run.~~ Crater popped up some ICEs ([now fixed](https://github.com/rust-lang/rust/pull/73566#issuecomment-656934851)). See [crater run](https://crater-reports.s3.amazonaws.com/pr-73566/index.html), [ICEs](https://github.com/rust-lang/rust/pull/73566#issuecomment-653619212).
~~Blocked on #74070 so that we don't make typeck_tables_of public when it shouldn't be.~~ Merged.

Closes #71820, closes #71104, closes #65863.

## What is the motivation for this change?

As seen from a lengthy trail of PRs and issues (https://github.com/rust-lang/rust/pull/73532, https://github.com/rust-lang/rust/pull/73103, https://github.com/rust-lang/rust/issues/71820, https://github.com/rust-lang/rust/issues/71104), `everybody_loops` is causing bugs in rustdoc. The main issue is that it does not preserve the validity of the `DefId` tree, meaning that operations on DefIds may unexpectedly fail when called later. This is blocking intra-doc links (see https://github.com/rust-lang/rust/pull/73101).

This PR starts by removing `everybody_loops`, fixing #71104 and #71820. However, that brings back the bugs seen originally in https://github.com/rust-lang/rust/pull/43348: Since libstd documents items for all platforms, the function bodies sometimes do not type check. Here are the errors from documenting `libstd` with `everybody_loops` disabled and no other changes:

```rust
error[E0433]: failed to resolve: could not find `handle` in `sys`
  --> src/libstd/sys/windows/ext/process.rs:13:27
   |
13 |         let handle = sys::handle::Handle::new(handle as *mut _);
   |                           ^^^^^^ could not find `handle` in `sys`

error[E0425]: cannot find function `symlink_inner` in module `sys::fs`
   --> src/libstd/sys/windows/ext/fs.rs:544:14
    |
544 |     sys::fs::symlink_inner(src.as_ref(), dst.as_ref(), false)
    |              ^^^^^^^^^^^^^ not found in `sys::fs`

error[E0425]: cannot find function `symlink_inner` in module `sys::fs`
   --> src/libstd/sys/windows/ext/fs.rs:564:14
    |
[...]

PR #87845, opened 2021-08-07
Add tidy check to ensure rustdoc ui tests parity

GuillaumeGomez:
In https://github.com/rust-lang/rust/pull/87728, I discovered that it wasn't known that tests present in `src/test/ui/rustdoc` should also be present in `src/test/rustdoc-ui` (because we need to ensure that the rustdoc attributes and errors are thrown by both rustc and rustdoc).

cc @jyn514 @camelid 
r? @Mark-Simulacrum


Mark-Simulacrum:
Hm, this seems like an odd fix - if we expect that one of these directories is a subset of the other, I would expect us to not have two directories but instead annotate files with the compiletest revision support or a new dedicated feature to this effect. The current check just checks for same file name in both directories, which seems quite error prone - either file can get edited easily.

I think either we should move to revisions or a similar style feature. But I'm not sure the premise of this is quite right. If the code to validate attributes is in rustc's attribute validation, then rustdoc simply must run that - I don't think in that case we need to duplicate this for *all* tests. We don't rerun regular UI tests with rustdoc to make sure it's not using a different rustc parser or something; this seems like a similar case.

GuillaumeGomez:
Yes, it's not great as is. For context: for a long time, rustdoc hid most attributes checks, preventing the warnings from rustc to be deniable in rustdoc. The idea was then to ensure this behaviour doesn't happen again.

jyn514:
> Yes, it's not great as is. For context: for a long time, rustdoc hid most attributes checks, preventing the warnings from rustc to be deniable in rustdoc. The idea was then to ensure this behaviour doesn't happen again.

This can happen _anyway_ though: because of https://github.com/rust-lang/rust/pull/73566, the lints aren't ever emitted in the first place. This misleads people working on rustdoc because they could think it's a bug on their end that the lint shows up in rustc and not rustdoc, when in fact that's intentional. The tests only pass now because we happen to only test lints that run in rustdoc.

Mark-Simulacrum:
I think it makes sense for T-rustdoc to decide whether/how this check should look.

jyn514:
Like I said above, I don't think we should add this check.

GuillaumeGomez:
Closing then.

PR #87848, opened 2021-08-07
removed references to parent/child from std::thread documentation

godmar:
- also clarifies how thread.join and detaching of threads works
- the previous prose implied that there is a relationship between a
spawning thread and the thread being spawned, and that "child" threads
couldn't outlive their "parents" unless detached, which is incorrect.

### DIFF
--- a/library/std/src/thread/mod.rs
+++ b/library/std/src/thread/mod.rs
@@ -28,7 +28,7 @@
 //! When the main thread of a Rust program terminates, the entire program shuts
 //! down, even if other threads are still running. However, this module provides
 //! convenient facilities for automatically waiting for the termination of a
-//! child thread (i.e., join).
+//! thread (i.e., join).
 //!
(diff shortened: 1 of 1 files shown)


ilyagr:
I wonder, does this mean that the answer at https://stackoverflow.com/a/55228719/563359 is incorrect?

godmar:
I don't have the time to review answers on Internet Q&A sites, but the Rust documentation with my fixes is correct.

When main() returns, exit() will be called, which terminates the entire process including all threads running in it. It's this way in C, C++, and Rust also. Rust threads aren't special - they have the same semantics as that which is specified by POSIX, the standard on top of which they are implement in Linux at least.

There is no parent-child relationship between threads, either. Any thread can join any joinable thread. Detached threads cannot be joined. Reading this information misrepresented in the 2018 version of the documentation prompted this PR.

PR #87865, opened 2021-08-08
Clarify terms in rustdoc book

### LINKED ISSUE #70898: Clarification of terms
The `rustdoc` documentation mentions a number of terms which I haven't found the definitions for. Might it be that they are either undefined or defined elsewhere?

- "compiltest" from https://doc.rust-lang.org/rustdoc/unstable-features.html#--enable-per-target-ignores-allow-ignore-foo-style-filters-for-doctests
- "Attribute", "annotation", and "directive" seem to be used interchangeably here https://doc.rust-lang.org/rustdoc/documentation-tests.html#attributes

tsoutsman:
Fixes #70898

I chose to completely remove the term directive over attribute because rustdoc has a lint called `invalid_codeblock_attributes` and the term attributes is used throughout the book. I slightly changed the introductory sentence to describe the relationship between annotations and attributes.

I also moved the text explaining the example from below the blocks to above the blocks which is more in line with the rest of the book. I also changed the description for the `should_panic` attribute as I found it a little confusing. Finally, some of the blocks were `text` and some were `rust` so I changed them all to `text` which is in line with the rest of the book.

### DIFF
--- a/src/doc/rustdoc/src/documentation-tests.md
+++ b/src/doc/rustdoc/src/documentation-tests.md
@@ -297,8 +297,13 @@ we can add the `#[macro_use]` attribute. Second, we’ll need to add our own
 
 ## Attributes
 
-There are a few annotations that are useful to help `rustdoc` do the right
+Code blocks can be annotated with attributes that help `rustdoc` do the right
 thing when testing your code:
+
+The `ignore` attribute tells Rust to ignore your code. This is almost never
+what you want as it's the most generic. Instead, consider annotating it
+with `text` if it's not code or using `#`s to get a working example that
+only shows the part you care about.
 
 ```rust
 /// ```ignore
@@ -307,10 +312,8 @@ thing when testing your code:
 # fn foo() {}
 ```
 
-The `ignore` directive tells Rust to ignore your code. This is almost never
-what you want, as it's the most generic. Instead, consider annotating it
-with `text` if it's not code, or using `#`s to get a working example that
-only shows the part you care about.
+`should_panic` tells `rustdoc` that the code should compile correctly but
+panic during execution. If the code doesn't panic, the test will fail.
 
 ```rust
 /// ```should_panic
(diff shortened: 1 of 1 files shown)


src/doc/rustdoc/src/documentation-tests.md
```diff
@@ -297,32 +297,37 @@ we can add the `#[macro_use]` attribute. Second, we’ll need to add our own
...
 
-The `ignore` directive tells Rust to ignore your code. This is almost never
-what you want, as it's the most generic. Instead, consider annotating it
-with `text` if it's not code, or using `#`s to get a working example that
-only shows the part you care about.
+`should_panic` tells `rustdoc` that the code should compile correctly but
+panic during execution. If the code doesn't panic, the test will fail.
 
-```rust
+```text
```
GuillaumeGomez, line 318: It's definitely rust code. Please keep it as is. :)

src/doc/rustdoc/src/documentation-tests.md
```diff
@@ -297,32 +297,37 @@ we can add the `#[macro_use]` attribute. Second, we’ll need to add our own
...
 
-`should_panic` tells `rustdoc` that the code should compile correctly, but
-not actually pass as a test.
+The `no_run` attribute will compile your code but not run it. This is
+important for examples such as "Here's how to retrieve a web page,"
+which you would want to ensure compiles, but might be run in a test
+environment that has no network access.
 
-```rust
+```text
```
GuillaumeGomez, line 330: It's definitely rust code. Please keep it as is. :)

GuillaumeGomez (changes requested):
Looks good. Please keep the current annotations as is please (you change the other `text` ones into rust if you want but then don't forget to add an item like we did for the others (`# fn foo() {}`).

tsoutsman:
Done

GuillaumeGomez (approved):

GuillaumeGomez:
Thanks!

@bors: r+ rollup

PR #87869, opened 2021-08-08
Make io::Error use 64 bits on targets with 64 bit pointers.

thomcc:
I've wanted this for a long time, but didn't see a good way to do it without having extra allocation. When looking at it yesterday, it was more clear what to do for some reason.

This approach avoids any additional allocations, and reduces the size by half (8 bytes, down from 16). AFAICT it doesn't come additional runtime cost, and the compiler seems to do a better job with code using it.

Additionally, this `io::Error` has a niche (still), so `io::Result<()>` is *also* 64 bits (8 bytes, down from 16), and `io::Result<usize>` (used for lots of io trait functions) is 2x64 bits (16 bytes, down from 24 — this means on x86_64 it can use the nice rax/rdx 2-reg struct return). More generally, it shaves a whole 64 bit integer register off of the size of basically any `io::Result<()>`.

(For clarity: Improving `io::Result` (rather than io::Error) was most of the motivation for this)

On 32 bit (or other non-64bit) targets we still use something equivalent the old repr — I don't think think there's improving it, since one of the fields it stores is a `i32`, so we can't get below that, and it's already about as close as we can get to it.

---

### Isn't Pointer Tagging Dodgy?

The details of the layout, and why its implemented the way it is, are explained in the header comment of library/std/src/io/error/repr_bitpacked.rs. There's probably more details than there need to be, but I didn't trim it down that much, since there's a lot of stuff I did deliberately, that might have not seemed that way.

There's actually only one variant holding a pointer which gets tagged. This one is the (holder for the) user-provided error.

I believe the scheme used to tag it is not UB, and that it preserves pointer provenance (even though often pointer tagging does not) because the tagging operation is just `core::ptr::add`, and untagging is `core::ptr::sub`. The result of both operations lands inside the original allocation, so it would follow the safety contract of `core::ptr::{add,sub}`.

The other pointer this had to encode is not tagged — or rather, the tagged repr is equivalent to untagged (it's tagged with 0b00, and has >=4b alignment, so we can reuse the bottom bits). And the other variants we encode are just integers, which (which can be untagged using bitwise operations without worry — they're integers).

CC @RalfJung for the stuff in repr_bitpacked.rs, as my comments are informed by a lot of the UCG work, but it's possible I missed something or got it wrong (even if the implementation is okay, there are parts of the header comment that says things like "We can't do $x" which could be false).

---

### Why So Many Changes?

The repr change was mostly internal, but changed one widely used API: I had to switch how `io::Error::new_const` works.

This required switching `io::Error::new_const` to take the full message data (including the kind) as a `&'static`, rather than just the string. This would have been really tedious, but I made a macro that made it much simpler, but it was a wide change since `io::Error::new_const` is used everywhere.

This included changing files for a lot of targets I don't have easy access to (SGX? Haiku? Windows? Who has heard of these things), so I expect there to be spottiness in CI initially, unless luck is on my side.

Anyway this large only tangentially-related change is all in the first commit (although that commit also pulls the previous repr out into its own file), whereas the packing stuff is all in commit 2.

---

P.S. I haven't looked at all of this since writing it, and will do a pass over it again later, sorry for any obvious typos or w/e. I also definitely repeat myself in comments and such.

(It probably could use more tests too. I did some basic testing, and made it so we `debug_assert!` in cases the decode isn't what we encoded, but I don't know the degree which I can assume libstd's testing of IO would exercise this. That is: it wouldn't be surprising to me if libstds IO testing were minimal, especially around error cases, although I have no idea).

### DIFF
--- a/library/std/src/io/error/repr_bitpacked.rs
+++ b/library/std/src/io/error/repr_bitpacked.rs
@@ -6,9 +6,9 @@
 //! a more clever manner than `rustc`'s default layout algorithm would).
 //!
 //! Conceptually, it stores the same data as the "unpacked" equivalent we use on
-//! other targets. Specifically, you can imagine it as an optimized following
-//! data (which is equivalent to what's stored by `repr_unpacked::Repr`, e.g.
-//! `super::ErrorData<Box<Custom>>`):
+//! other targets. Specifically, you can imagine it as an optimized version of
+//! the following enum (which is roughly equivalent to what's stored by
+//! `repr_unpacked::Repr`, e.g. `super::ErrorData<Box<Custom>>`):
 //!
 //! ```ignore (exposition-only)
 //! enum ErrorData {
@@ -135,7 +135,16 @@ impl Repr {
         // (rather than `ptr::wrapping_add`), but it's unclear this would give
         // any benefit, so we just use `wrapping_add` instead.
         let tagged = p.wrapping_add(TAG_CUSTOM).cast::<()>();
-        // Safety: the above safety comment also means the result can't be null.
+        // Safety: `TAG_CUSTOM + p` is the same as `TAG_CUSTOM | p`,
+        // because `p`'s alignment means it isn't allowed to have any of the
+        // `TAG_BITS` set (you can verify that addition and bitwise-or are the
+        // same when the operands have no bits in common using a truth table).
+        //
+        // Then, `TAG_CUSTOM | p` is not zero, as that would require
+        // `TAG_CUSTOM` and `p` both be zero, and neither is (as `p` came from a
+        // box, and `TAG_CUSTOM` just... isn't zero -- it's `0b01`). Therefore,
+        // `TAG_CUSTOM + p` isn't zero and so `tagged` can't be, and the
+        // `new_unchecked` is safe.
         let res = Self(unsafe { NonNull::new_unchecked(tagged) });
         // quickly smoke-check we encoded the right thing (This generally will
         // only run in libstd's tests, unless the user uses -Zbuild-std)
@@ -342,12 +351,25 @@ static_assert!(@usize_eq: size_of::<NonNull<()>>(), size_of::<usize>());
 static_assert!(@usize_eq: size_of::<&'static SimpleMessage>(), 8);
 static_assert!(@usize_eq: size_of::<Box<Custom>>(), 8);
 
-// And they must have >= 4 byte alignment.
-static_assert!(align_of::<SimpleMessage>() >= 4);
-static_assert!(align_of::<Custom>() >= 4);
+static_assert!((TAG_MASK + 1).is_power_of_two());
+// And they must have sufficient alignment.
+static_assert!(align_of::<SimpleMessage>() >= TAG_MASK + 1);
+static_assert!(align_of::<Custom>() >= TAG_MASK + 1);
 
-// This is obviously true (`TAG_CUSTOM` is `0b01`), but our implementation of
-// `Repr::new_custom` and such would be wrong if it were not, so we check.
+static_assert!(@usize_eq: (TAG_MASK & TAG_SIMPLE_MESSAGE), TAG_SIMPLE_MESSAGE);
+static_assert!(@usize_eq: (TAG_MASK & TAG_CUSTOM), TAG_CUSTOM);
+static_assert!(@usize_eq: (TAG_MASK & TAG_OS), TAG_OS);
+static_assert!(@usize_eq: (TAG_MASK & TAG_SIMPLE), TAG_SIMPLE);
+
+// This is obviously true (`TAG_CUSTOM` is `0b01`), but in `Repr::new_custom` we
+// offset a pointer by this value, and expect it to both be within the same
+// object, and to not wrap around the address space. See the comment in that
+// function for further details.
+//
+// Actually, at the moment we use `ptr::wrapping_add`, not `ptr::add`, so this
+// check isn't needed for that one, although the assertion that we don't
+// actually wrap around in that wrapping_add does simplify the safety reasoning
+// elsewhere considerably.
 static_assert!(size_of::<Custom>() >= TAG_CUSTOM);
 
 // These two store a payload which is allowed to be zero, so they must be
--- a/library/std/src/io/error/repr_bitpacked.rs
+++ b/library/std/src/io/error/repr_bitpacked.rs
@@ -47,9 +47,10 @@
 //!   `Box<Custom>`. `Custom` also has alignment >= 4, so the bottom two bits
 //!   are free to use for the tag.
 //!
-//!   The only important thing to note is that `ptr::add` and `ptr::sub` are
-//!   used to tag the pointer, rather than bitwise operations. This should
-//!   preserve the pointer's provenance, which would otherwise be lost.
+//!   The only important thing to note is that `ptr::wrapping_add` and
+//!   `ptr::wrapping_sub` are used to tag the pointer, rather than bitwise
+//!   operations. This should preserve the pointer's provenance, which would
+//!   otherwise be lost.
 //!
 //! - **Tag 0b10**: Holds the data for `ErrorData::Os(i32)`. We store the `i32`
 //!   in the pointer's most significant 32 bits, and don't use the bits `2..32`
@@ -126,11 +127,14 @@ impl Repr {
         // Should only be possible if an allocator handed out a pointer with
         // wrong alignment.
         debug_assert_eq!((p as usize & TAG_MASK), 0);
-        // Safety: We know `TAG_CUSTOM <= size_of::<Custom>()` (static_assert at
+        // Note: We know `TAG_CUSTOM <= size_of::<Custom>()` (static_assert at
         // end of file), and both the start and end of the expression must be
         // valid without address space wraparound due to `Box`'s semantics.
-        // Note: `add` is used as a provenance-preserving way of pointer tagging.
-        let tagged = unsafe { p.add(TAG_CUSTOM).cast::<()>() };
+        //
+        // This means it would be correct to implement this using `ptr::add`
+        // (rather than `ptr::wrapping_add`), but it's unclear this would give
+        // any benefit, so we just use `wrapping_add` instead.
+        let tagged = p.wrapping_add(TAG_CUSTOM).cast::<()>();
         // Safety: the above safety comment also means the result can't be null.
         let res = Self(unsafe { NonNull::new_unchecked(tagged) });
         // quickly smoke-check we encoded the right thing (This generally will
@@ -238,7 +242,10 @@ where
         }
         TAG_SIMPLE_MESSAGE => ErrorData::SimpleMessage(&*ptr.cast::<SimpleMessage>().as_ptr()),
         TAG_CUSTOM => {
-            let custom = ptr.as_ptr().cast::<u8>().sub(TAG_CUSTOM).cast::<Custom>();
+            // It would be correct for us to use `ptr::sub` here (see the
+            // comment above the `wrapping_add` call in `new_custom` for why),
+            // but it isn't clear that it makes a difference, so we don't.
+            let custom = ptr.as_ptr().cast::<u8>().wrapping_sub(TAG_CUSTOM).cast::<Custom>();
             ErrorData::Custom(make_custom(custom))
         }
         _ => {
@@ -337,7 +344,7 @@ static_assert!(align_of::<SimpleMessage>() >= 4);
 static_assert!(align_of::<Custom>() >= 4);
 
 // This is obviously true (`TAG_CUSTOM` is `0b01`), but our implementation of
-// `Repr::new_custom` and such would be UB if it were not, so we check.
+// `Repr::new_custom` and such would be wrong if it were not, so we check.
 static_assert!(size_of::<Custom>() >= TAG_CUSTOM);
 // These two store a payload which is allowed to be zero, so they must be
 // non-zero to preserve the `NonNull`'s range invariant.
--- a/library/std/src/io/error/repr_bitpacked.rs
+++ b/library/std/src/io/error/repr_bitpacked.rs
@@ -28,7 +28,7 @@
 //!
 //!   This means we can skip tagging it, which is necessary as this variant can
 //!   be constructed from a `const fn`, which probably cannot tag pointers (or
-//!   at least it would be difficult.
+//!   at least it would be difficult).
 //!
 //! - **Tag 0b01**: The other pointer variant holds the data for
 //!   `ErrorData::Custom` and the remaining 62 bits are used to store a
@@ -40,7 +40,7 @@
 //!   preserve the pointer's provenance, which would otherwise be lost.
 //!
 //! - **Tag 0b10**: Holds the data for `ErrorData::Os(i32)`. We store the `i32`
-//!   in the pointers most significant 32 bits, and don't use the bits `2..32`
+//!   in the pointer's most significant 32 bits, and don't use the bits `2..32`
 //!   for anything. Using the top 32 bits is just to let us easily recover the
 //!   `i32` code with the correct sign.
 //!
--- a/library/std/src/io/error/repr_bitpacked.rs
+++ b/library/std/src/io/error/repr_bitpacked.rs
@@ -321,23 +321,26 @@ fn kind_from_prim(ek: u32) -> Option<ErrorKind> {
 // that our encoding relies on for correctness and soundness. (Some of these are
 // a bit overly thorough/cautious, admittedly)
 //
-// If any of these are hit on a platform that libstd supports, we should just
-// make sure `repr_unpacked.rs` is used instead.
+// If any of these are hit on a platform that libstd supports, we should likely
+// just use `repr_unpacked.rs` there instead (unless the fix is easy).
 macro_rules! static_assert {
     ($condition:expr) => {
-        const _: [(); 0] = [(); (!$condition) as usize];
+        const _: () = assert!($condition);
+    };
+    (@usize_eq: $lhs:expr, $rhs:expr) => {
+        const _: [(); $lhs] = [(); $rhs];
     };
 }
 
 // The bitpacking we use requires pointers be exactly 64 bits.
-static_assert!(size_of::<NonNull<()>>() == 8);
+static_assert!(@usize_eq: size_of::<NonNull<()>>(), 8);
 
 // We also require pointers and usize be the same size.
-static_assert!(size_of::<NonNull<()>>() == size_of::<usize>());
+static_assert!(@usize_eq: size_of::<NonNull<()>>(), size_of::<usize>());
 
 // `Custom` and `SimpleMessage` need to be thin pointers.
-static_assert!(size_of::<&'static SimpleMessage>() == 8);
-static_assert!(size_of::<Box<Custom>>() == 8);
+static_assert!(@usize_eq: size_of::<&'static SimpleMessage>(), 8);
+static_assert!(@usize_eq: size_of::<Box<Custom>>(), 8);
 
 // And they must have >= 4 byte alignment.
 static_assert!(align_of::<SimpleMessage>() >= 4);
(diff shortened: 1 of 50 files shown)


library/std/src/io/error/repr_bitpacked.rs
```diff
@@ -0,0 +1,338 @@
...
+    let bits = ptr.as_ptr() as usize;
+    match bits & TAG_MASK {
+        TAG_OS => {
+            let code = ((bits as i64) >> 32) as i32;
+            ErrorData::Os(code)
+        }
+        TAG_SIMPLE => {
+            let kind_bits = (bits >> 32) as u32;
+            let kind = kind_from_prim(kind_bits).unwrap_or_else(|| {
+                debug_assert!(false, "Invalid io::error::Repr bits: `Repr({:#016x})`", bits);
```
nvzqz, line 216: `:#016x` will give you 14 digits at most. You want 18 there to account for `0x` and get 16 digits.

thomcc: dangit

library/std/src/io/error/repr_bitpacked.rs
```diff
@@ -0,0 +1,338 @@
...
+            matches!(res.data(), ErrorData::Simple(k) if k == kind),
+            "repr(simple) encoding failed {:?}",
+            kind,
+        );
+        res
+    }
+
+    #[inline]
+    pub(super) const fn new_simple_message(m: &'static SimpleMessage) -> Self {
+        // Safety: We're a Repr, decode_repr is fine.
```
jfrimmel, line 162:
```suggestion
        // Safety: references are never null
```
Seems like a copy/paste-error. Despite that: is it UB-free to cast a shared reference to a `*mut`?

thomcc: > Seems like a copy/paste-error.

Nice catch.

> Despite that: is it UB-free to cast a shared reference to a *mut?

No, it's completely fine as long as it's never written to and always used as a shared reference (never a `&mut`), which is the case here.

I'd have written this function without unsafe if it weren't `const fn`: `NonNull::from(m).cast::<()>()` (https://doc.rust-lang.org/src/core/ptr/non_null.rs.html#612-620) is equivalent, but I can't use the `From` impl in `const fn`. Since that's a ways off, I need to go through `NonNull::new`, instead which takes a `*mut`.

library/std/src/io/error/repr_bitpacked.rs
```diff
@@ -0,0 +1,346 @@
...
+//!   `ErrorData::SimpleMessage`, and holds a `&'static SimpleMessage` directly.
+//!
+//!   `SimpleMessage` has an alignment >= 4 (which is requested with
+//!   `#[repr(align)]` and checked statically at the bottom of this file), which
+//!   means every `&'static SimpleMessage` should have the both tag bits as 0,
+//!   meaning its tagged and untagged representation are equivalent.
+//!
+//!   This means we can skip tagging it, which is necessary as this variant can
+//!   be constructed from a `const fn`, which probably cannot tag pointers (or
+//!   at least it would be difficult.
```
RalfJung, line 31:
```suggestion
//!   at least it would be difficult).
```

library/std/src/io/error/repr_bitpacked.rs
```diff
@@ -0,0 +1,346 @@
...
+//!
+//! - **Tag 0b01**: The other pointer variant holds the data for
+//!   `ErrorData::Custom` and the remaining 62 bits are used to store a
+//!   `Box<Custom>`. `Custom` also has alignment >= 4, so the bottom two bits
+//!   are free to use for the tag.
+//!
+//!   The only important thing to note is that `ptr::add` and `ptr::sub` are
+//!   used to tag the pointer, rather than bitwise operations. This should
+//!   preserve the pointer's provenance, which would otherwise be lost.
```
RalfJung, line 40: Any reason not to use `wrapping_add`/`wrapping_sub`? Then we don't have to worry about going out-of-bounds...

library/std/src/io/error/repr_bitpacked.rs
```diff
@@ -0,0 +1,346 @@
...
+//!
+//! - **Tag 0b10**: Holds the data for `ErrorData::Os(i32)`. We store the `i32`
+//!   in the pointers most significant 32 bits, and don't use the bits `2..32`
```
RalfJung, line 43:
```suggestion
//!   in the pointer's most significant 32 bits, and don't use the bits `2..32`
```

library/std/src/io/error/repr_bitpacked.rs
```diff
@@ -0,0 +1,346 @@
...
+//!    the motivation of this bitpacking.
+//!
+//! Storing everything in a `NonZeroUsize` (or some other integer) would be a
+//! bit more traditional for pointer tagging, but it would lose provenance
+//! information, couldn't be constructed from a `const fn`, and would probably
+//! run into other issues as well.
+//!
+//! The `NonNull<()>` seems like the only alternative, even if it's fairly odd
+//! to use a pointer type to store something that may hold an integer, some of
+//! the time.
```
RalfJung, line 90: FWIW, I think it would be even more odd to use an integer type to store something that may hold a pointer, some of the time. ;)  But I understand that is the "more traditional" thing to do.

I fully agree with your choice here.

library/std/src/io/error/repr_bitpacked.rs
```diff
@@ -0,0 +1,346 @@
...
+    })
+}
+
+// Some static checking to alert us if a change breaks any of the assumptions
+// that our encoding relies on for correctness and soundness. (Some of these are
+// a bit overly thorough/cautious, admittedly)
+//
+// If any of these are hit on a platform that libstd supports, we should just
+// make sure `repr_unpacked.rs` is used instead.
+macro_rules! static_assert {
```
RalfJung, line 307: Isn't there a better place for this macro?

library/std/src/io/error/repr_bitpacked.rs
```diff
@@ -0,0 +1,346 @@
...
+static_assert!(TAG_SIMPLE != 0);
+// We can't tag `SimpleMessage`s, the tag must be 0.
+static_assert!(TAG_SIMPLE_MESSAGE == 0);
+
+// Check that the point of all of this still holds.
+//
+// We'd check against `io::Error`, but *technically* it's allowed to vary,
+// as it's not `#[repr(transparent)]`/`#[repr(C)]`. We could add that, but
+// the `#[repr()]` would show up in rustdoc, which might be seen as a stable
+// commitment.
```
RalfJung, line 342: We should really clarify if this is a stable commitment or not... but that's a different discussion.

RalfJung (commented):
I only read the comments in `repr_bitpacked` carefully, and modulo some small comments that all sounds very sensible. :)

library/std/src/io/error/repr_bitpacked.rs
RalfJung: > No, it's completely fine as long as it's never written to and always used as a shared reference (never a &mut), which is the case here.

:+1:

library/std/src/io/error/repr_bitpacked.rs
thomcc: I think the logic is to only use a pointer if it is always a pointer, and to use an integer if it ever is instead used to store arbitrary integer data.

I'm pretty sure the reason for this is that the C and C++ standards get very cagey about promising int->ptr->int will round-trip (they don't promise it, even if the int is big enough), but they do guarantee that ptr->int->ptr will for a sufficiently sized int.

OTOH, Rust's approach actually makes more sense to me when I think about it — there are cases where the compiler really absolutely needs to know something is a pointer just to correctly compile the code (for example, if it's a constant, and it points something which may be relocated).

library/std/src/io/error/repr_bitpacked.rs
thomcc: It's intended not to be!

library/std/src/io/error/repr_bitpacked.rs
thomcc: Probably! I was surprised nothing like it already existed somewhere. I'm open for suggestions about where to put it though.

library/std/src/io/error/repr_bitpacked.rs
RalfJung: >  those methods were equivalent to casting through integers

No, that is not the case. (Where is that impression coming from? We might have to improve the docs.)
The docs currently say

> The resulting pointer “remembers” the allocated object that self points to; it must not be used to read or write other allocated objects.

which is pretty much saying that provenance is preserved. In fact, preserving the provenance is the entire reason those methods even exist!

library/std/src/io/error/repr_bitpacked.rs
RalfJung: > I'm pretty sure the reason for this is that the C and C++ standards get very cagey about promising int->ptr->int will round-trip (they don't promise it, even if the int is big enough), but they do guarantee that ptr->int->ptr will for a sufficiently sized int.

The irony is that ptr->int->ptr in C/C++ will almost certainly *not* roundtrip losslessly (so, yes it does a roundtrip, but information is lost). That is demonstrated by [this example](https://www.ralfj.de/blog/2020/12/14/provenance.html). In contrast, int->ptr->int will roundtrip without loss at least in LLVM. (But in C, int->ptr is UB if this is not a valid address...)

An alternative would be to use a `union` for this:
```rust
union PtrOrInt {
  ptr: NonNull<()>,
  int: NonZeroUsize,
}
```
You can always access the `ptr` field to check the tag (since int->ptr transmute is fine, unlike ptr->int transmute which is [problematic](https://github.com/rust-lang/unsafe-code-guidelines/issues/286)).

But that is actually fairly close to just using `NonNull` like you do, the only difference being what happens when turning an integer value into the tagged representation.

library/std/src/io/error/repr_bitpacked.rs
RalfJung: stdarch has `static_assert!`, though it is slightly different than what you do.

I'd probably add it as an unstable public macro in `library/core/src/macros/mod.rs` -- but that might be better done as a separate PR involving the libs-api team.

library/std/src/io/error/repr_bitpacked.rs
RalfJung: > there are cases where the compiler really absolutely needs to know something is a pointer just to correctly compile the code (for example, if it's a constant, and it points something which may be relocated).

This is just saying that provenance exists for real. :)  At compile-time, there's no denying this.

library/std/src/io/error/repr_bitpacked.rs
thomcc: I feel weird about exporting it from `core`, because I don't think this looks anything like what we'd want from a stdlib `static_assert` macro.

JohnCSimon:
Ping from triage: 
 
@thomcc Can you please address the merge conflict?

thomcc:
Sure, although it might not be until the start of October.

library/std/src/io/error/repr_bitpacked.rs
thomcc: The existence of both of these implies there are cases where `ptr::{sub,add}` and `ptr::wrapping_{sub,add}` have different performance characteristics. Given that the non-wrapping nature of the use here is so straightforward, shouldn't we keep it this way?

library/std/src/io/error/repr_bitpacked.rs
thomcc: Actually, maybe this should just be `const _: () = assert!(...)`? Not sure.

library/std/src/io/error/repr_bitpacked.rs
```diff
@@ -0,0 +1,346 @@
+//! This is a densely packed error representation which is used on targets with
+//! 64-bit pointers.
+//!
+//! (Note that `bitpacked` vs `unpacked` here has no relationship to
+//! `#[repr(packed)]`, it just refers to attempting to use any available
+//! bits in a more clever manner than `rustc`'s default layout algorithm would).
+//!
+//! Conceptually, it stores the same information as the "unpacked" equivalent we
+//! use on other targets: `repr_unpacked::Repr` (see repr_unpacked.rs), however
+//! it packs it into a 64bit non-zero value.
```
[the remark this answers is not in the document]
thomcc: I worry that repeating this here would be redundant and get out of date, especially when users browsing the source can pretty easily access `ErrorData`. That said, I've made this change.

library/std/src/io/error/repr_bitpacked.rs
```diff
@@ -0,0 +1,346 @@
...
+    };
+}
+
+// The bitpacking we use requires pointers be exactly 64 bits.
+static_assert!(size_of::<NonNull<()>>() == 8);
+
+// We also require pointers and usize be the same size.
+static_assert!(size_of::<NonNull<()>>() == size_of::<usize>());
+
+// `Custom` and `SimpleMessage` need to be thin pointers.
```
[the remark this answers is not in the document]
thomcc: I'm marking as resolved since I couldn't figure out how to word this without it being more confusing.

library/std/src/io/error/repr_bitpacked.rs
RalfJung: I found the in-bounds argument to be somewhat subtle, so using the unsafe arithmetic looks like a premature optimization to me. I.e., the difference in perf seems more theoretical than practical.

But, up to you, I am not at all an expert on the perf details here.

thomcc: Fair enough, I don't feel like looking at the asm for this (let alone doing benchmarks), so I've switched to `wrapping_{add,sub}`, but left the note that the `wrapping` isn't actually needed, in case someone needs to change it in the future.

kornelski:
Maybe to help with the merge conflicts, first land a PR that changes *only* the new_const function to a macro call, with no other changes?

JohnCSimon:
Ping from triage: 
@thomcc I see a lot of merge conflicts but the PR hasnt been touched in three months. Closing as inactive, feel free to reopen. 

@rustbot label: +S-inactive

thomcc:
I think this was actually waiting on the reviewer — they hadn't updated in over 5. The rebase errors is because it modified all creation sites of errors in libstd, so it's very easy to get behind.

Anyway, I'll try and rebase it in a week or so.

JohnCSimon:
> I think this was actually waiting on the reviewer 

Sorry @thomcc  it was marked as `S-waiting-on-author`
I'll switch it over.

yaahc:
Oops, sorry about that 😅, no idea why I swapped it to waiting on author, I'll take a look at this next week.

ben0x539:
If one were to want to add another variant to io::Error, one would need to start assuming align(8) instead of align(4), but that's pretty much a given anyway, right?

thomcc:
Yes, but in practice we already have that alignment anyway.

thomcc:
Also, it was brought up that maybe this includes ABI guarantees on the size. I think I've said this already in resolved comments, but no, this is intended just to be an optimization, and not include any ABI guarantees about the error type, size or otherwise.

thomcc:
(I found some time over my lunch break to update this)

library/std/src/io/error/repr_bitpacked.rs
thomcc: I went with that, but added a special case to the macro for equality comparisons (`assert_eq!` didn't seem to work, or perhaps just needed a `#![feature(...)]` added somewhere — I didn't try very hard here).

Sadly, this does upset rust-analyzer quite a bit more than the old code, but seems to be work fine.

ben0x539:
I really appreciate the extra explicitness of the updated comments!

library/std/src/io/error/tests.rs
```diff
@@ -67,3 +67,18 @@ fn test_const() {
...
+        let e = Error::from_raw_os_error(code);
+        assert_eq!(e.raw_os_error(), Some(code));
+    }
+    assert_eq!(Error::from(ErrorKind::NotFound).kind(), ErrorKind::NotFound);
+    assert_eq!(Error::from(ErrorKind::Uncategorized).kind(), ErrorKind::Uncategorized);
+    assert_eq!(Error::from(ErrorKind::NotFound).kind(), ErrorKind::NotFound);
+    assert_eq!(Error::from(ErrorKind::Uncategorized).kind(), ErrorKind::Uncategorized);
+    let dunno = const_io_error!(ErrorKind::Uncategorized, "dunno");
+    assert_eq!(dunno.kind(), ErrorKind::Uncategorized);
+    assert!(format!("{:?}", dunno).contains("dunno"))
```
yaahc, line 83: what is this checking?

library/std/src/io/error/tests.rs
```diff
@@ -67,3 +67,18 @@ fn test_const() {
     assert!(format!("{:?}", E).contains("\"hello\""));
     assert!(format!("{:?}", E).contains("NotFound"));
 }
+
+#[test]
+fn test_error_packing() {
```
yaahc, line 72: Does this get full coverage of the Repr kinds we support? I can see how it would cover OS, Simple, and SimpleMessage, but not Custom.

library/std/src/io/error/repr_bitpacked.rs
```diff
@@ -0,0 +1,369 @@
+//! This is a densely packed error representation which is used on targets with
+//! 64-bit pointers.
+//!
+//! (Note that `bitpacked` vs `unpacked` here has no relationship to
+//! `#[repr(packed)]`, it just refers to attempting to use any available bits in
+//! a more clever manner than `rustc`'s default layout algorithm would).
+//!
+//! Conceptually, it stores the same data as the "unpacked" equivalent we use on
+//! other targets. Specifically, you can imagine it as an optimized following
```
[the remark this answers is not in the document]
thomcc: Erm, ferris ate a few words out of that sentence.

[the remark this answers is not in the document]
thomcc: Ferris ate some words out of my sentence, and I definitely didn't just forget them.

library/std/src/io/error/tests.rs
thomcc: That's covered elsewhere in the file to some degree, but I've beefed the tests up and made them fairly explicit.

library/std/src/io/error/tests.rs
thomcc: I believe there isn't really anything in the public API that exposes that message right now -- it's just in the output.

I've added a comment to that effect, and also made all these tests verify check that the ErrorData variant (e.g. the portable enum that the packed stuff decodes into) is what's expected (also helps ensure there's not something giving the right answer for the wrong reason)

thomcc:
Updated to address tests, and address some comments that came outside of this issue.

Also, I noticed the `NonNull::new_unchecked` safety doc in `Repr::new_custom` had gotten stale (it said "see above", referring to the safety comment for ptr::add, which I had made more hand-wavey after switching to wrapping_add).

I also added a few more assertions to ensure that none of the tags use bits outside TAG_MASK, and replace some 4's with `TAG_MASK + 1` since it seems likely we will add more variants in the future, which would require another tag bit (which won't be a problem).

yaahc:
I think this all looks good to go to me! Thanks again for the amazing work @thomcc 

@bors r+ rollup=never

thomcc:
Hmm, interesting. The regressions seem to be in layout computation (`conservative_is_privately_uninhabited`, `layout_of`, etc, if I'm reading it right). I guess that makes sense, that was the point and stuff like `Result<T, io::Error>` is pretty common (same reason that the change seemed worthwhile).

Hmmmmm...

thomcc:
Looking more closely, the regressions are... almost all versions of [`issue-46449`](https://github.com/rust-lang/rustc-perf/blob/master/collector/benchmarks/issue-46449/src/lib.rs) with different build flags. This seems fairly synthetic (while it was derived from real code, it's pretty far from it now, and AFAICT that code had performance problems for reasons wholly unrelated to layout).

Thankfully, the benches with perf deltas that seem to more realistic crates, seem to be more consistently green -- it's not 100% green in the real usage, but its a heck of a lot more consistent, and for the most part seems like the code that uses `io::Error` more benefits more.

(Honestly... if you ignore the `issue-46449` family, it's feels like a pretty consistent (but not 100%) win for compile times, and mostly in checks of real code. But OTOH, perhaps you can always say "if you ignore the thing that my code makes worse, you'll find that my code makes everything better")

Anyway, it kinda seems fine to regress a synthetic benchmark like that in this case (where there are runtime benefits in very commonly used types), especially given how many of the non-synthetic benchmarks had a compile-time improvement. But it may be unsurprising that I'm in favor of my own patch.

(That said, I aslo I can't quite wrap my head around what I'd poke in the compiler improve this anyway (not more specifically than  ✨ layout ✨, at least), so I'd appreciate pointers if it turns out I'm responsible for un-regressing it).

Mark-Simulacrum:
I'm going to go ahead and mark this as triaged -- I tend to agree that this is likely not something to be directly addressed, and is limited to just one benchmark.

PR #87874, opened 2021-08-09
Add TcpStream type to TcpListener::incoming docs

schneems:
## Context

While going through the "The Rust Programming Language" book (Klabnik & Nichols), the TCP server example directs us to use TcpListener::incoming. I was curious how I could pass this value to a function (before reading ahead in the book), so I looked up the docs to determine the signature. 

When I opened the docs, I found https://doc.rust-lang.org/std/net/struct.TcpListener.html#method.incoming, which didn't mention TcpStream anywhere in the example.

Eventually, I clicked on https://doc.rust-lang.org/std/net/struct.TcpListener.html#method.accept in the docs (after clicking a few other locations first), and was able to surmise that the value contained TcpStream.

## Opportunity

While this type is mentioned several times in this doc, I feel that someone should be able to fully use the results of the TcpListner::incoming iterator based solely on the docs of just this method.

## Implementation

I took the code from the top-level TcpListener https://doc.rust-lang.org/std/net/struct.TcpListener.html#method.incoming and blended it with the existing docs for TcpListener::incoming https://doc.rust-lang.org/std/net/struct.TcpListener.html#method.incoming.

It does make the example a little longer, and it also introduces a little duplication. It also gives the reader the type signatures they need to move on to the next step.

## Additional considerations

I noticed that in this doc, `handle_connection` and `handle_client` are both used to accept a TcpStream in the docs on this page. I want to standardize on one function name convention, so readers don't accidentally think two different concepts are being referenced. I didn't want to cram do too much in one PR, I can update this PR to make that change, or I could send another PR (if you would like).

First attempted contribution to Rust (and I'm also still very new, hence reading through the rust book for the first time)! Would you please let me know what you think?

### DIFF
--- a/library/std/src/net/tcp.rs
+++ b/library/std/src/net/tcp.rs
@@ -767,17 +767,24 @@ impl TcpListener {
     /// # Examples
     ///
     /// ```no_run
-    /// use std::net::TcpListener;
+    /// use std::net::{TcpListener, TcpStream};
     ///
-    /// let listener = TcpListener::bind("127.0.0.1:80").unwrap();
+    /// fn handle_connection(stream: TcpStream) {
+    ///    //...
+    /// }
     ///
-    /// for stream in listener.incoming() {
-    ///     match stream {
-    ///         Ok(stream) => {
-    ///             println!("new client!");
+    /// fn main() -> std::io::Result<()> {
+    ///     let listener = TcpListener::bind("127.0.0.1:80").unwrap();
+    ///
+    ///     for stream in listener.incoming() {
+    ///         match stream {
+    ///             Ok(stream) => {
+    ///                 handle_connection(stream);
+    ///             }
+    ///             Err(e) => { /* connection failed */ }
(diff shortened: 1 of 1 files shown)


ASayre:
@matthew-piziak Hi friend!  I thought of you as our team was learning rust and going through the Rust book.

matthew-piziak:
@ASayre Hi friend! It's a pleasure to hear from you, and I hope that Rust is treating you and your team well so far.

Nice contribution, @schneems. I tossed this example into my editor and it looks good. I also see what you mean regarding `handle_connection` and `handle_client`.

Mark-Simulacrum:
@bors r+ rollup

Thank you! I agree that explicitly indicating the iterator type like this seems beneficial in this case, and helps be uniform with the top-level example too. Updating handle_connection or handle_client names to standardize on one name seems reasonable, please feel free to file a new PR with that.

schneems:
Thank you!

Issue #87573, opened 2021-07-28
ICE: panicked at 'index out of bounds: the len is 0 but the index is 0', could not prove Binder(TraitPredicate(<bool as Copy>), [])


chengniansun:
Might be a duplicate of #86535

### Code

```Rust
#![feature(no_core, lang_items)]
#![no_core]
static STATIC_BOOL: bool = true;
#[lang = "sized"]
trait Sized {}
#[lang = "copy"]
trait Copy {}
#[lang = "sync"]
trait Sync {}
impl Sync for bool {}
#[lang = "drop_in_place"]
fn main() {
    while false {}
}
```

### Meta

`rustc --version --verbose`:
```
rustc 1.56.0-nightly (2faabf579 2021-07-27)
binary: rustc
commit-hash: 2faabf579323f5252329264cc53ba9ff803429a3
commit-date: 2021-07-27
host: x86_64-unknown-linux-gnu
release: 1.56.0-nightly
LLVM version: 12.0.1
```

### Error output

```
warning: static is never used: `STATIC_BOOL`
 --> mutant.rs:3:1
  |
3 | static STATIC_BOOL: bool = true;
  | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  |
  = note: `#[warn(dead_code)]` on by default

thread 'rustc' panicked at 'index out of bounds: the len is 0 but the index is 0', /rustc/2faabf579323f5252329264cc53ba9ff803429a3/compiler/rustc_middle/src/ty/subst.rs:303:43
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace

error: internal compiler error: unexpected panic

note: the compiler unexpectedly panicked. this is a bug.

note: we would appreciate a bug report: https://github.com/rust-lang/rust/issues/new?labels=C-bug%2C+I-ICE%2C+T-compiler&template=ice.md

note: rustc 1.56.0-nightly (2faabf579 2021-07-27) running on x86_64-unknown-linux-gnu

note: compiler flags: --crate-type staticlib

query stack during panic:
#0 [resolve_instance] resolving instance `main`
#1 [collect_and_partition_mono_items] collect_and_partition_mono_items
end of query stack
warning: 1 warning emitted

error: internal compiler error: broken MIR in DefId(0:6 ~ mutant[d0c6]::main) (NoSolution): could not prove Binder(TraitPredicate(<bool as Copy>), [])
  |
[...]

Issue #9307, opened 2013-09-19
type-check lang items


thestinger:
These aren't currently well-defined by the language.  If you define anything with a somewhat matching LLVM signature, it will compile. Not only does it not care about the type signatures matching what it expects, there is no handling for these being defined on the wrong types of AST nodes.

nagisa:
Still asserts with

```
#![no_std]
#![feature(no_std,core,lang_items)]

extern crate core;

#[lang="start"]
fn main() {
}

#[lang = "stack_exhausted"] extern fn stack_exhausted() {}
#[lang = "eh_personality"] extern fn eh_personality() {}
#[lang = "panic_fmt"] fn panic_fmt() -> ! { loop{} }
```

steveklabnik:
Triage: no changes

steveklabnik:
Triage; not aware of any changes here

Mark-Simulacrum:
I'm closing this issue -- I don't think type checking language items is all that helpful. Anyone using the lang items feature is likely / should be aware of what they're doing and adding type checks in the compiler would likely be complicated and somewhat painful, and essentially just "wasted" work from a compiler performance perspective.

james-darkfox:
@Mark-Simulacrum I strongly disagree that this is waste. The compiler knows the type it is expecting, so it should meaningfully tell the user what the type it expects is, instead of accepting anything and failing if you're lucky. Worse, what happens when a type changes? However rare, I'd rather have the types than an ICE or other obscure outputs.

Mark-Simulacrum:
The point here is that unlike almost everything else in Rust, lang items are always pretty visible and when changing their signature it's almost always done inside libstd/libcore and as such the time investment into checking them (compile time and human time) seems to make this not really worth tracking. If a minimal PR was filed and/or we gradually started doing this, that's reasonable, but IMO we gain almost nothing from having an issue.

james-darkfox:
@Mark-Simulacrum But it is an issue and is far from closed. You may not see the effort in implementing it but if it remains open, someone else may do so.

Mark-Simulacrum:
The reason I don't see tracking it as useful is because it's essentially an internal implementation detail -- similar to having a dedicated issue for random FIXMEs in the code. In this case, it's not even clear that we want to do this -- at least to me. I'll reopen, I guess, because I don't feel strongly enough to argue about it more, but I don't think this is useful.

james-darkfox:
Thank you @Mark-Simulacrum.

[...]

PR #87875, opened 2021-08-09
Improve detection of generics on lang items

asquared31415:
Adds detection for the required generics for all lang items.  Many lang items require an exact or minimum amount of generic arguments and if they don't exist, the compiler will ICE.  This does not add any additional validation about bounds on generics or any other lang item restrictions.

Fixes one of the ICEs in #87573

cc @FabianWolff


asquared31415:
Somewhat related to #9307, but not entirely, and does not close it

asquared31415:
Whelp I "cleaned up" the code and didn't rerun tests by mistake, used a `!=` to compare "less than" by mistake

asquared31415:
It compiles locally with that feature enabled, weird, guess it's stabilized now though

compiler/rustc_hir/src/lang_items.rs
```diff
@@ -352,3 +352,153 @@ language_item_table! {
...
 }
+
+pub enum GenericRequirement {
+    None,
+    Minimum(usize),
+    Exact(usize),
+}
+
+impl LangItem {
+    pub fn required_generics_count(&self) -> GenericRequirement {
```
LeSeulArtichaut, line 363: Maybe we can make this a part of the `language_item_table!` macro to avoid listing all the language items again? I'm not sure though, so you might want to wait for someone else's optinion, I'd feel bad if you did the work to revert it :sweat_smile:

LeSeulArtichaut (commented):
I'm not familiar at all with this code, so I'm going to pass this to someone else.

LeSeulArtichaut:
Per the GitHub suggestions: r? @cjgillot do you want to take this?

compiler/rustc_hir/src/lang_items.rs
asquared31415: I was thinking the same thing last night right after I submitted the PR.  I think it makes sense from a maintenance standpoint as well.

compiler/rustc_passes/src/lang_items.rs
```diff
@@ -191,111 +191,59 @@ impl LanguageItemCollector<'tcx> {
...
-        // and the rest also have none except for the closure traits (one for
-        // the argument list), generators (one for the resume argument),
-        // ordering/equality relations (one for the RHS), and various conversion
-        // traits.
+        // arguments. Generally speaking, binary and indexing operations have
+        // one (for the RHS/index), unary operations have none, the closure
+        // traits have one for the argument list, generators have one for the
+        // resume argument, and ordering/equality relations have one for the RHS
+        // Some other types like Box and various functions like drop_in_place
+        // have minimum requirements.
```
cjgillot, line 199: IIUC, minimum requirements exist in cases where the other generics can be inferred, either through a function parameter or through defaulted parameters. So, why do some traits (`Unsize`, `CoerceUnsized`) have minimum requirements?

src/test/ui/lang-items/generics/wrong-number-fn.rs
```diff
@@ -0,0 +1,14 @@
+#![feature(lang_items,no_core)]
```
cjgillot, line 1: Can all those tests be put in a single file?

cjgillot (commented):
Thanks for the PR @asquared31415. I left a few nits.
Could you also squash the commits?

compiler/rustc_passes/src/lang_items.rs
```diff
@@ -191,111 +191,59 @@ impl LanguageItemCollector<'tcx> {
...
+                }
+                // If the number matches, or there is no requirement, do nothing
+                _ => return,
+            };
+            // We are issuing E0718 "incorrect target" here, because while the
+            // item kind of the target is correct, the target is still wrong
+            // because of the wrong number of generic arguments.
+            self.tcx
+                .sess
+                .struct_span_fatal_with_code(
```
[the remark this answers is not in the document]
asquared31415: When trying to figure out how to stop compilation as soon as possible, I found that method and it looked relevant.  I didn't think to check back and see if the macro worked after adding the required `abort_if_errors` though.  It seems to work with the macro, I'll change it back.

src/test/ui/lang-items/generics/wrong-number-fn.rs
asquared31415: I had to make it two, because I have two tests that should pass for a couple different lang items, and one for the ones that fail.  But they have been merged into two files in the `lang-items` directory root.

asquared31415:
Oh no CI failed enough that the bot didn't even post a comment about the failure.  How did this break this badly.

compiler/rustc_passes/src/lang_items.rs
```diff
@@ -321,6 +270,11 @@ fn get_lang_items(tcx: TyCtxt<'_>, (): ()) -> LanguageItems {
     // Find all required but not-yet-defined lang items.
     weak_lang_items::check_crate(tcx, &mut items);
 
+    if collector.has_lang_item_generic_error {
+        // Continuing compilation with a malformed lang item will not end well, so stop as soon as possible
+        tcx.sess.abort_if_errors();
```
cjgillot, line 275: Compilation should probably be aborted if we have the wrong target also.

asquared31415: That's handled by not adding the lang item to the list in the first place, so it can never be used incorrectly in the compiler.  But that actually gives me an idea to handle this better than a flag in the struct, I can probably remove these incorrect generics too, and that will allow compilation to continue a little and potentially show other errors, which is nicer.

compiler/rustc_passes/src/lang_items.rs
```diff
@@ -191,110 +191,57 @@ impl LanguageItemCollector<'tcx> {
...
-                    generics_span,
-                    format!(
-                        "this trait has {} generic argument{}, not {}",
-                        actual_num,
-                        pluralize!(actual_num),
-                        expected_num
-                    ),
-                )
-                .emit();
+            self.items.items[item_index] = None;
```
cjgillot, line 242: I may be even better to only call `collect_item` if the generics test passes.

asquared31415:
I'm not sure how to resolve that CI issue.  It *looks* like it's identical to `src/test/ui/error-codes/E0152.rs` before my changes, but I don't know where that test that is failing CI comes from.

asquared31415:
Nevermind, I found it, it was in the E0152.md file in `rustc_error_codes`

asquared31415:
I'm fairly certain that failure was spurious, is there any action I need to take to get it to retry?

PR #87879, opened 2021-08-09
Fix early exiting when `ExprUseVisitor` meets unresolved type var

### LINKED ISSUE #87814: Closure captures immutable reference when mutable reference is necessary
### Code

I tried this code:

```rust
fn main() {
    let mut schema_all = vec![];
    (0..42).for_each(|_x| match Err(()) as Result<(), _> {
        Ok(()) => schema_all.push(()),
        Err(_) => (),
    });
}

```

I expected to see this happen: *cargo build success*

Instead, this happened: *cargo build failed*
```bash
error[E0596]: cannot borrow `schema_all` as mutable, as it is behind a `&` reference
 --> src/main.rs:4:19
  |
4 |         Ok(()) => schema_all.push(()),
  |                   ^^^^^^^^^^ cannot borrow as mutable

warning: variable does not need to be mutable
 --> src/main.rs:2:9
  |
2 |     let mut schema_all = vec![];
  |         ----^^^^^^^^^^
  |         |
  |         help: remove this `mut`
  |
  = note: `#[warn(unused_mut)]` on by default

error: aborting due to previous error; 1 warning emitted

For more information about this error, try `rustc --explain E0596`.
```

### Version it worked on

It most recently worked on: `nightly-2021-03-16`

### Version with regression

`rustc --version --verbose`:
```
nightly-2021-03-17
```

ldm0:
Closes #87814

```rust
fn main() {
    let mut schema_all = vec![];
    (0..42).for_each(|_x| match Err(()) as Result<(), _> {
        Ok(()) => schema_all.push(()),
        Err(_) => (),
    });
}
```

`ExprUseVisitor::walk_expr ======> hir::ExprKind::Match => MemCategorizationContext::cat_pattern ======> self.cat_pattern_ ======> PatKind::TupleStruct => self.pat_ty_adjusted ======> self.pat_ty_unadjusted ======> self.node_ty ======> self.resolve_type_vars_or_error` 

With `as Result<(), _>` presents, the argument `ty` of `resolve_type_vars_or_error` when checking `Err(_) => ..` arm is a type var rather than a unit. And the error bubbles through all the functions listed above. Which means the `walk_arm()` call in the `ExprUseVisitor::walk_expr` is skipped.

I think `walk_arm` is a process not mangled with the `cat_pattern`. So working around this problem by swallowing the type var resolving error.

r? @roxelo @nikomatsakis

### DIFF
--- a/compiler/rustc_typeck/src/expr_use_visitor.rs
+++ b/compiler/rustc_typeck/src/expr_use_visitor.rs
@@ -243,7 +243,7 @@ impl<'a, 'tcx> ExprUseVisitor<'a, 'tcx> {
                 let ExprUseVisitor { ref mc, body_owner: _, delegate: _ } = *self;
                 let mut needs_to_be_read = false;
                 for arm in arms.iter() {
-                    return_if_err!(mc.cat_pattern(discr_place.clone(), &arm.pat, |place, pat| {
+                    match mc.cat_pattern(discr_place.clone(), &arm.pat, |place, pat| {
                         match &pat.kind {
                             PatKind::Binding(.., opt_sub_pat) => {
                                 // If the opt_sub_pat is None, than the binding does not count as
@@ -290,7 +290,13 @@ impl<'a, 'tcx> ExprUseVisitor<'a, 'tcx> {
                                 // examined
                             }
                         }
-                    }));
+                    }) {
+                        Ok(_) => (),
+                        Err(_) => {
+                            // If typeck failed, assume borrow is needed.
+                            needs_to_be_read = true;
+                        }
+                    }
                 }
 
                 if needs_to_be_read {
--- a/src/test/ui/closures/issue-87814-1.rs
+++ b/src/test/ui/closures/issue-87814-1.rs
@@ -1,0 +1,8 @@
+// check-pass
+fn main() {
+    let mut schema_all = vec![];
+    (0..42).for_each(|_x| match Err(()) as Result<(), _> {
+        Ok(()) => schema_all.push(()),
+        Err(_) => (),
+    });
+}
--- a/src/test/ui/closures/issue-87814-2.rs
+++ b/src/test/ui/closures/issue-87814-2.rs
@@ -1,0 +1,11 @@
+// check-pass
+#![feature(try_reserve)]
+
+fn main() {
+    let mut schema_all: (Vec<String>, Vec<String>) = (vec![], vec![]);
+
+    let _c = || match schema_all.0.try_reserve(1) as Result<(), _> {
+        Ok(()) => (),
+        Err(_) => (),
+    };
+}


ldm0:
No high-five bot? @rust-highfive ✋

roxelo:
I think we should add a test where the match discriminant also need to be borrowed mutably so we can confirm that this fix actually resolves the root issue. For instance, something like this also causes an ICE on nightly:
```rust
#![feature(try_reserve)]
#![feature(capture_disjoint_fields)]

fn main() {
    let mut schema_all : (Vec<String>, Vec<String>) = (vec![], vec![]);
    
    let _c = || {
        match schema_all.0.try_reserve(1) as Result<(), _> {
            Ok(()) => (),
            Err(_) => (),
        }
    };
    
}
```
```thread 'rustc' panicked at 'called `Result::unwrap()` on an `Err` value: PlaceBuilder { base: Upvar { var_hir_id: HirId { owner: DefId(0:3 ~ playground[4e8d]::main), local_id: 26 }, closure_def_id: DefId(0:4 ~ playground[4e8d]::main::{closure#0}), closure_kind: Fn }, projection: [Field(field[0], std::vec::Vec<std::string::String>)] }', compiler/rustc_mir_build/src/build/expr/as_place.rs:306:69```

ldm0:
> I think we should add a test where the match discriminant also needs to be borrowed mutably so we can confirm that this fix actually resolves the root issue. For instance, something like this also causes an ICE on nightly:
> 
> ```rust
> #![feature(try_reserve)]
> #![feature(capture_disjoint_fields)]
> 
> fn main() {
>     let mut schema_all : (Vec<String>, Vec<String>) = (vec![], vec![]);
>     
>     let _c = || {
>         match schema_all.0.try_reserve(1) as Result<(), _> {
>             Ok(()) => (),
>             Err(_) => (),
>         }
>     };
>     
> }
> ```
> 
> `` thread 'rustc' panicked at 'called `Result::unwrap()` on an `Err` value: PlaceBuilder { base: Upvar { var_hir_id: HirId { owner: DefId(0:3 ~ playground[4e8d]::main), local_id: 26 }, closure_def_id: DefId(0:4 ~ playground[4e8d]::main::{closure#0}), closure_kind: Fn }, projection: [Field(field[0], std::vec::Vec<std::string::String>)] }', compiler/rustc_mir_build/src/build/expr/as_place.rs:306:69 ``

The original use case of `cat_pattern` ignores the error it returned, I think we can do the same thing here.

ldm0:
@roxelo BTW, why not `break` when `needs_to_be_read` is set to `true`?

ldm0:
@roxelo  Thanks for reviewing!

> The good thing for us is the error seems to be returned after the op(&place_with_id, pat); is run in cat_pattern, so we should still be getting the correct information on whether the match discriminant should be read or not.

Hm.... the field checking is skipped I think. Why it's still "correct" in this situation? 🤔 

> I think there is a simpler way to do this than your current change; we should just need to replace the `return_if_err!` with a `let _ = ...`

When a typeck error is uncovered, rustc should pessimistically assume borrow is needed(correct me if I'm wrong). BTW, I changed the signature of the closure given to `cat_pattern`, so emitting an error in the closure can terminates the whole walking process. With the more complex change, rustc will ends the walking process when it finds an evidence that borrow is needed, then performance should be improved.

roxelo:
> Hm.... the field checking is skipped I think. Why it's still "correct" in this situation? 🤔

Oh right, you can ignore my earlier comment. 

> When a typeck error is uncovered, rustc should pessimistically assume borrow is needed(correct me if I'm wrong)

As a workaround for this issue I guess yes, but we really want it to be deterministic since it wasn't expected that `cat_pattern` would fail with correct code.

> BTW, I changed the signature of the closure given to cat_pattern, so emitting an error in the closure can terminates the whole walking process.

I am personally not a fan of using an error in that part of the code as it's not an error for the discriminant to need to be read/borrowed...

nikomatsakis:
Closing in favor of https://github.com/rust-lang/rust/pull/88266

PR #87880, opened 2021-08-09
Remove duplicate trait bounds in `rustc_data_structures::graph`

pierwill:
Remove duplicate trait bounds in `rustc_data_structures::graph`.


LeSeulArtichaut (approved):
Nice catch :+1:

PR #87881, opened 2021-08-09
Proper table row formatting in platform support

badboy:
Also moves the target into alphabetical order


ehuss:
Thanks! 
@bors r+ rollup

PR #87882, opened 2021-08-09
Description for `HashSet::insert` redundant and difficult to understand

porky11:
I had to read it carefully to really understand, what's going on.

There has been the same sentence twice, but once negated.
So reading the second one does not give any useful information.

And the sentence, which contains the "not" is associated with `true`.
This way, it's not clear, why `true` or `false` are returned in these cases.

Instead of having descriptions of the specific cases, there should rather be a general description like "Returns if the set has been changed by adding the value" or "Returns if the value has been added".

But for clarity, I didn't remove the explicit description of the `false` case, so if you don't understand the description of the `true` case, the description of the `false` case might still help you.

### DIFF
--- a/library/std/src/collections/hash/set.rs
+++ b/library/std/src/collections/hash/set.rs
@@ -809,11 +809,11 @@ where
         other.is_subset(self)
     }
 
-    /// Adds a value to the set.
+    /// Inserts a value to the set.
     ///
-    /// If the set did not have this value present, `true` is returned.
+    /// If the insertion changed the set, `true` is returned.
     ///
-    /// If the set did have this value present, `false` is returned.
+    /// If the value already has been present, `false` is returned.
     ///
     /// # Examples
     ///


library/std/src/collections/hash/set.rs
```diff
@@ -809,11 +809,11 @@ where
         other.is_subset(self)
     }
 
-    /// Adds a value to the set.
+    /// Inserts a value to the set.
     ///
-    /// If the set did not have this value present, `true` is returned.
+    /// If the insertion changed the set, `true` is returned.
```
kennytm, line 814: i prefer still specifying the state before the insertion. it also requires some thinking to decode what is meant by "changed".

```suggestion
    /// If the the value was absent, i.e. insertion changed the set, `true` is returned.
```

library/std/src/collections/hash/set.rs
```diff
@@ -809,11 +809,11 @@ where
         other.is_subset(self)
     }
 
-    /// Adds a value to the set.
+    /// Inserts a value to the set.
```
camsteffen, line 812: Grammatically I think it should be either "add..to" or "insert...into". I prefer the current wording.

porky11: That's the least important change. I just changed it for consistency reasons after using "insert" multiple times.

library/std/src/collections/hash/set.rs
porky11: What I care most about is the meaning of the returned value.
If there are descriptions for both cases, they should at least not be redundant.

I want a reason, why each value is returned in each case.

Maybe also "Returns if insertion was successful".
This seems like a very straightforward description. But maybe it's not correct when using the mathematical definition of a set.
You can insert the same value into a set multiple times, it wouldn't change, and would still be considered as successful, I guess.

kennytm: @porky11 The current wording in this PR maybe clear to *you* but it is not clear to *me*, and documentation should be clear to *all of us*. Documentation is not the same as coding and redundancy for clarity is not a bad thing.

By only saying "insertion changed the set", *I* need to do the calculation of "What is meant by changed? The final state is value has been inserted, and a hash-set can either contain a value or not, so 'changed' must mean the initial state is the value was absent, OK!". That's a totally unnecessary "make me think" exercise that can be avoided by just including the "value was absent" in the doc and makes everyone happy.

JohnCSimon:
@porky11 If this PR is ready for a review please adjust the label with `@rustbot ready`

JohnCSimon:
@porky11 

@rustbot label: +S-inactive

PR #87883, opened 2021-08-09
Use smallvec for FieldsShape::Arbitrary members

hkratz:
Inline sizes are chosen to match the Vec size as a first approximation.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4309,6 +4309,7 @@ dependencies = [
  "rustc_macros",
  "rustc_serialize",
  "rustc_span",
+ "smallvec",
  "tracing",
 ]
 
--- a/compiler/rustc_middle/src/ty/layout.rs
+++ b/compiler/rustc_middle/src/ty/layout.rs
@@ -19,6 +19,7 @@ use rustc_target::abi::call::{
 };
 use rustc_target::abi::*;
 use rustc_target::spec::{abi::Abi as SpecAbi, HasTargetSpec, PanicStrategy};
+use smallvec::SmallVec;
 
 use std::cmp;
 use std::fmt;
@@ -265,8 +266,8 @@ enum StructKind {
 // and `inverse_memory_index` (memory order to source field order).
 // See also `FieldsShape::Arbitrary::memory_index` for more details.
 // FIXME(eddyb) build a better abstraction for permutations, if possible.
-fn invert_mapping(map: &[u32]) -> Vec<u32> {
-    let mut inverse = vec![0; map.len()];
+fn invert_mapping(map: &[u32]) -> SmallVec<[u32; 4]> {
+    let mut inverse = smallvec![0; map.len()];
     for i in 0..map.len() {
         inverse[map[i] as usize] = i as u32;
     }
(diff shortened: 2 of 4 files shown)


hkratz:
Huh. The RSS is [higher in about every test](https://perf.rust-lang.org/compare.html?start=ae90dcf0207c57c3034f00b07048d63f8b2363c8&end=f8eccf98d11657d7baad97ad6235ab3269a4ae58&stat=max-rss), about the opposite of what I was naively expecting. Anyone any idea what's going on or what to check?

Kobzol:
If RSS went up, that (probably) means that the smallvecs take more memory than Vecs. I would expect that even a single element Vec will take more space than a smallvec with such small inline sizes. So maybe the Vecs are often empty, and therefore using a smallvec (which takes more space even if empty) is what regresses RSS? It would be intereseting to see what is the size distribution of these vectors.

hkratz:
> If RSS went up, that (probably) means that the smallvecs take more memory than Vecs. I would expect that even a single element Vec will take more space than a smallvec with such small inline sizes. So maybe the Vecs are often empty, and therefore using a smallvec (which takes more space even if empty) is what regresses RSS? It would be intereseting to see what is the size distribution of these vectors.

Thing is, with the smallvec `union` feature the smallvec 2xu64 and 4xu32 sizes are exactly the same size as empty vecs, so the `FieldsShape` is the same size with smallvecs:
```rust 
// Cargo.toml smallvec dependency:
// smallvec = { version = "*", features = ["union", "may_dangle"] }

use std::{mem, num::NonZeroUsize};

use smallvec::SmallVec;

pub struct Size {
    raw: u64,
}

pub enum FieldsShapeVec {
    Primitive,
    Union(NonZeroUsize),
    Array {
        stride: Size,
        count: u64,
    },
    Arbitrary {
        offsets: Vec<Size>,
        memory_index: Vec<u32>,
    },
}

pub enum FieldsShapeSmallVec {
    Primitive,
    Union(NonZeroUsize),
    Array {
        stride: Size,
        count: u64,
    },
    Arbitrary {
        offsets: SmallVec<[Size; 2]>,
        memory_index: SmallVec<[u32; 4]>,
    },
}

fn main() {
    assert_eq!(
        mem::size_of::<FieldsShapeVec>(),
        mem::size_of::<FieldsShapeSmallVec>()
    );
}
```

Kobzol:
Interesting. It might also be caused by different buffer growth strategies of `Vec`/`smallvec`. For example, it seems that the `vec!` macro allocates the exact capacity for the passed number of elements, while `smallvec` rounds it to a power of two:

```rust
use smallvec::SmallVec as SVec;

fn main() {
    let mut v: Vec<u32> = vec!(1, 2, 3);
    println!("{}", v.capacity());
    for i in 0..71 {
        v.push(i);
    }
    println!("{}", v.capacity());

    let mut v: SVec<[u32; 4]> = smallvec::smallvec![1, 2, 3];
    println!("{}", v.capacity());
    for i in 0..71 {
        v.push(i);
    }
    println!("{}", v.capacity());
}
```

That might cause further differences in allocated memory, because `smallvec!` might allocate more than `vec!`.

hkratz:
Huh. That also means that if len == capacity == (2^n) - 1 two subsequent `push()`es grow the SmallVec twice in row, e.g.
```
use smallvec::SmallVec as SVec;

fn main() {
    let mut v: SVec<[u32; 2]> = smallvec::smallvec![1, 2, 3, 4, 5, 6, 7];
    println!("{}", v.capacity());
    v.push(1);
    println!("{}", v.capacity());
    v.push(1);
    println!("{}", v.capacity());
}
/* 
Prints
7
8
16
*/
```

Kobzol:
FWIW, I tried to build `rustc` using a modified `smallvec` that grows capacity by multiplying by two instead of finding powers of two. Instructions remained pretty much identical across all benchmarks in `rustc-perf` and RSS was slightly regressed in most of them (about 2 %, so might be noise).

Dylan-DPC:
Closing this as inactive

PR #76524, opened 2020-09-09
typeck: don't suggest inaccessible private fields

davidtwco:
Fixes #76077.

This PR adjusts the missing field diagnostic logic in typeck so that when none of the missing fields in a struct expr are accessible then the error is less confusing.

r? @estebank

PR #87895, opened 2021-08-09
typeck: don't suggest inaccessible fields in struct literals and suggest ignoring inaccessible fields in struct patterns

### LINKED ISSUE #87872: Missing field error in initializer for struct with inaccessible fields
Given the following code:

```rust
// in file a.rs
#![crate_type = "lib"]

pub struct X {  
  pub f: u32, // remove the field f to get a different error message
  pub(crate) g: u32,
}

// this is to prevent a warning that g is not used
impl X {
  pub fn value(self) -> u32 { self.g }
}
```

```rust
// in file b.rs
#![crate_type = "bin"]

extern crate a;

fn main() {
  let x = a::X {};
  x.value();
}
```

Using ` rustc a.rs` followed by `rustc -L . b.rs`, the current output is:

```
error[E0063]: missing fields `f` and `g` in initializer of `X`
 --> b.rs:6:11
  |
6 |   let x = a::X {};
  |           ^^^^ missing `f` and `g`

```

```
error: cannot construct `X` with struct literal syntax due to inaccessible fields
 --> b.rs:6:11
  |
6 |   let x = a::X {};
  |           ^^^^

```

The rationale is the same as for https://github.com/rust-lang/rust/issues/76077. Following the suggestion provided by the compiler will lead to another error. This can be frustrating with structs that have a large number of fields or fields with complex types, since the effort of adding those fields to the struct literal is in vain.

By removing the field *f* from the struct, the compiler is reporting that *X* cannot be constructed, as expected.

Tested on both `rustc 1.56.0-nightly (574d37568 2021-08-07)` and `rustc 1.52.0-nightly (4a8b6f708 2021-03-11)`.

TheWastl:
Fixes #87872.

This PR adjusts the missing field diagnostic logic in typeck so that when any of the missing fields in a struct literal or pattern is inaccessible then the error is less confusing, even if some of the missing fields are accessible.

See also #76524.

### DIFF
--- a/compiler/rustc_typeck/src/check/expr.rs
+++ b/compiler/rustc_typeck/src/check/expr.rs
@@ -1313,15 +1313,12 @@ impl<'a, 'tcx> FnCtxt<'a, 'tcx> {
                 .emit();
             }
         } else if check_completeness && !error_happened && !remaining_fields.is_empty() {
-            let no_accessible_remaining_fields = remaining_fields
-                .iter()
-                .find(|(_, (_, field))| {
-                    field.vis.is_accessible_from(tcx.parent_module(expr_id).to_def_id(), tcx)
-                })
-                .is_none();
+            let inaccessible_remaining_fields = remaining_fields.iter().any(|(_, (_, field))| {
+                !field.vis.is_accessible_from(tcx.parent_module(expr_id).to_def_id(), tcx)
+            });
 
-            if no_accessible_remaining_fields {
-                self.report_no_accessible_fields(adt_ty, span);
+            if inaccessible_remaining_fields {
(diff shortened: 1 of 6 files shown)


wesleywiser:
r? rust-lang/compiler

wesleywiser:
r? rust-lang/compiler-team

TheWastl:
I just realized the author is wrong (so GitHub doesn't connect it to my account), should I change it with a rebase and force-push?

estebank:
If you wish to do so, go ahead.

estebank:
r=me with the author fixed or not.

While you're at it, could you squash your commits?

estebank (approved):

PR #87897, opened 2021-08-09
Add `PathBuf::with_join`

WilliamVenner:
Shorthand function allowing for pushing into an owned PathBuf in a single line.

```rust
let path1 = {
    let mut path = PathBuf::from("/etc");
    path.push("passwd");
    path
};

let path2 = PathBuf::from("/etc").with_join("passwd");

assert_eq!(path1, path2);
```

## Naming

Vote!! https://www.strawpoll.me/45563102

### DIFF
--- a/library/std/src/path.rs
+++ b/library/std/src/path.rs
@@ -1403,6 +1403,32 @@ impl PathBuf {
     pub fn shrink_to(&mut self, min_capacity: usize) {
         self.inner.shrink_to(min_capacity)
     }
+
+    /// Adjoins `path` to `self` in-place, without allocating a new `PathBuf`.
+    ///
+    /// See [`Path::join`] for more details.
+    ///
+    /// # Examples
+    ///
+    /// ```
+    /// use std::path::{Path, PathBuf};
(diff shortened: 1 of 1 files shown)


Kixiron:
`.append()`, `.join_with()` and `.concat()` were some alternative names I thought of

Plecra:
For a micro-prior-art section, this would be an alternative to constructing paths with a macro similar to https://docs.rs/path_macro

WilliamVenner:
> `.append()`, `.join_with()` and `.concat()` were some alternative names I thought of

I like `append`, but I think `join_with` and `concat` may be ambiguous.

Although, `append` feels to me like it implies that it would take a mutable reference to self, rather than an owned self.

dtolnay (commented):
I would prefer not to add this. The cost to users of learning to use this API (i.e. deciding when to join vs with_join) outweighs any benefit to non-performance-sensistive code (which is the vast majority of `PathBuf` manipulations). The minority where the distinction would be relevant can use `push`.

PR #87907, opened 2021-08-10
Switch Instant to use CLOCK_BOOTTIME

### LINKED ISSUE #87906: `Instant::duration_since` returns a shortened duration if a suspend occurred
`Instant::now` currently uses `CLOCK_MONOTONIC`, which stops ticking during suspend. This means that creating a `Duration` via `Instant::duration_since` will be notably shorter than expected if the system suspended between the `&self` instance and the `earlier` argument.

If `CLOCK_BOOTTIME` is used instead, the clock is guaranteed to continue ticking during suspend. This closer matches the documentation and real world usage of `Instant`. The documentation for `Instant::duration_since` reads "Returns the amount of time elapsed from another instant to this one." - no mention of an exception for suspend or other low power states. `tokio` uses `Instant` to govern its sleep and timeout functionality. Given that `tokio` is commonly used to implement network servers and clients, slicing out suspended time was likely not their intention.

Using `CLOCK_BOOTTIME` should have minimal impact on existing usage. Performance-wise, `CLOCK_BOOTTIME` is implemented in Linux by getting the `CLOCK_MONOTONIC` time and adding an offset to it. That offset is updated as part of exiting suspend. As a result, the total cost should be an extra two loads , two adds, and maybe two stores with bad optimization, which are all negligible compared to the cost of servicing a system call. In terms of existing applications, the only ones which will see new behavior is ones which are measuring time across suspends. The only use case I can think of for wanting the old `CLOCK_MONOTONIC` behavior is time-multiplexing system resources, but I am not aware of an actual user.

## Other projects on the same issue
* Go decided against [switching its runtime](https://github.com/golang/go/issues/24595) to use `CLOCK_BOOTTIME` due to  
* Linux [converted](https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/?id=d6ed449afdb38f89a7b38ec50e367559e1b8f71f) `CLOCK_MONOTONIC` to act like `CLOCK_BOOTTIME` since it was almost always the desired behavior, and technically spec compatible, but [reverted](https://kernel.googlesource.com/pub/scm/linux/kernel/git/stable/linux-stable/+/a3ed0e4393d6885b4af7ce84b437dc696490a530) it shortly afterwards due to compatibility issues.

maurer:
For Linux-like platforms, use CLOCK_BOOTTIME which continues ticking
during suspend. Without this change, `Duration` between two `Instant`s
can bear little relation to reality if a suspend took place in between.

Fixes rust-lang#87906

### DIFF
--- a/library/std/src/sys/unix/time.rs
+++ b/library/std/src/sys/unix/time.rs
@@ -293,7 +293,7 @@ mod inner {
 
     impl Instant {
         pub fn now() -> Instant {
-            Instant { t: now(libc::CLOCK_MONOTONIC) }
+            Instant { t: now(libc::CLOCK_BOOTTIME) }
         }
 
         pub const fn zero() -> Instant {


hellow554:
I would find it really difficult to only have this behaviour on linux.

What about windows? How does it currently do the time measurement? What about bsd, ios, redox, ... ?

I think if we touch one thing, we should update every OS.

maurer:
> I would find it really difficult to only have this behaviour on linux.
> 
> What about windows? How does it currently do the time measurement? What about bsd, ios, redox, ... ?
BSD, iOS, redox, and OSX would all be updated by this change as well - perhaps my commit message could have been worded better, but they all use `unix/time.rs` to define `Instant`, and so use these libc definitions.

Windows has difficulties because `QueryPerformanceCounter` which we use right now doesn't run through suspend, but `GetTickCount` (which runs through suspend) has lower resolution (possibly as low as 15ms, depending on system configuration). If any Windows experts know a better way to do this that'd be great, but it's harder to argue for updating Windows here since there would be a downside to the change the only way I know how to make it.

> 
> I think if we touch one thing, we should update every OS.
In an ideal world I agree, but I don't think that we want a lowest common denominator interface here. Just because Windows doesn't offer a high precision timer that ticks in suspend doesn't mean we shouldn't use it where supported.

maurer:
My previous post was based on what is evidently old information. According to [Microsoft](https://docs.microsoft.com/en-us/windows/win32/sysinfo/acquiring-high-resolution-time-stamps#low-level-hardware-clock-characteristics) they are handling any necessary adjustments for sleep. As a result, this means that we are actually bringing the Unix-like implementation in line with the Windows one, and the two do not actually currently agree on semantics.

library/std/src/sys/unix/time.rs
```diff
@@ -293,7 +293,7 @@ mod inner {
 
     impl Instant {
         pub fn now() -> Instant {
-            Instant { t: now(libc::CLOCK_MONOTONIC) }
+            Instant { t: now(libc::CLOCK_BOOTTIME) }
```
Smittyvb, line 296: `CLOCK_BOOTTIME` [isn't supported on Linux <= 2.6.39](https://linux.die.net/man/2/clock_gettime), but [Rust supports Linux >= 2.6.32](https://doc.rust-lang.org/nightly/rustc/platform-support.html). This should check if `-EINVAL` is returned and fall back to `CLOCK_MONOTONIC` if so.

Smittyvb:
@workingjubilee That would have to be done to support Tier 1 targets too, since older versions of Linux supported by Rust don't support `CLOCK_BOOTTIME` (https://github.com/rust-lang/rust/pull/87907#pullrequestreview-727662030)

maurer:
> > BSD, iOS, redox, and OSX would all be updated by this change as well
> 
> osx has a separate implementation
> 

Sorry, missed that `unix/time.rs` contained multiple implementations. It looks like we're using `mach_absolute_time`, which Apple says [does not](https://developer.apple.com/documentation/kernel/1462446-mach_absolute_time) tick during suspend. They also have a function `mach_continous_time` which [does] tick during sleep. The discussion section for both calls recommend the use of `clock_gettime_nsec_np` with an appropriate clock ID over the use of those functions directly. `clock_gettime` and friends appears to have appeared in OSX 10.12, and our [platform support](https://doc.rust-lang.org/nightly/rustc/platform-support.html) says we support back to 10.7, so when I update this patch I'll switch it to use `mach_continuous_time`.

maurer:
> This does not mean the code would remain correct. CLOCK_MONOTONIC is true on all these platforms because it is POSIX. This notably does not specify whether CLOCK_BOOTTIME is a sensical value on all Unix platforms. I expect this PR would cause other platforms to fail to build, like FreeBSD, which uses its own constants and not the "Linux-like" set in libc. As such, this PR must bifurcate the implementation of `time.rs` further in order to work and not impact tier 2 support promises.

@workingjubilee 
Since you seem to know more about FreeBSD than I do, do you have a suggestion for which clock we should use there? Reading through the [man page](https://www.freebsd.org/cgi/man.cgi?query=clock_gettime&sektion=2&apropos=0&manpath=FreeBSD+14.0-current) and a [bug](https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=236702) where someone gets confused by the (admittedly documented) behavior of `CLOCK_UPTIME`, it sounds like there's currently no monotonically increasing clock which is also inclusive of suspend on FreeBSD. I'm tempted to just set them to `CLOCK_MONOTONIC` and add a comment in the documentation that some operating systems (insert a list there) don't support including sleep, and in those cases we just use a regular monotonic clock.

thomcc:
On macOS 10.12 (iOS/tvOS 10, watchOS 3) `clock_gettime_nsec_np(CLOCK_MONOTONIC_RAW)` can be used, which continues to increase across suspends, etc.

We'd have to dynamically load it (via `weak!` or similar) and fall back to the current code though, since we support much older versions of these targets (although it seems very likely that the majority of devices in the wild are likely to have this)

---

[1]: There's also `mach_continuous_time` which is equivalent, but it was added in/supports exactly the same versions, and is the [official docs](https://developer.apple.com/documentation/kernel/1646199-mach_continuous_time) say to use `clock_gettime_nsec_np(CLOCK_MONOTONIC_RAW)` instead. That said, the API for `mach_continuous_time` is closer to `mach_absolute_time`, so it may be less of a hassle to conditionally swap out one for the other.

(That said, I would not really tempt Apple about using an API they explicitly tell you not to in the docs...)

the8472:
> and add a comment in the documentation that some operating systems (insert a list there) don't support including sleep, and in those cases we just use a regular monotonic clock.

Imo that calls into question the value of this change. If we can't promise this behavior then someone may start to rely on non-portable implementation details because it worked on their machine but not on old (but supported) or less common platforms.

thomcc:
I feel that way too, and am fairly torn on it.

But the original issue seems to imply people already rely on this, or at least failed to consider it. As a result, their their code is has some bugs in this case.

ISTM like it already behaves non-portably, but just in a weaker sense — it works fine on machines that don't get put to sleep much/ever (many servers, some desktop machines, etc), but this assumption is "not portable" to systems that do, like consumer phones, laptops, etc. (Yes, I'm aware this is an abuse of terminology)

As mentioned though, I'm torn on it, and could probably be convinced either way.

nbdd0121:
I think we should just document `Instant` to say that behaviour across suspend is not guaranteed.

CryZe:
I'd say it absolutely should be documented that it can't guarantee that across all the platforms, such that people don't rely on it. But I'd still like to see this PR be merged. There's very little reason not to fix the behavior where possible. This otherwise just causes unnecessary bugs for lots of people for no good reason.

joshtriplett:
Having a huge discontinuity over suspends isn't necessarily always the desirable behavior. If you're trying to do something like "update based on the amount of time that has passed", or "measure how long a computation took", you may want to *skip* suspended time.

I personally would like more of an ability to use other clocks. But I don't think we should change the behavior of `now`, particularly if we then get inconsistencies across platforms.

Perhaps we should have an `InstantExt` on UNIX with a method that allows passing an arbitrary `libc::clockid_t` and/or `libc::timespec`. That would let people choose whatever clock they want to use.

thomcc:
> Perhaps we should have an InstantExt on UNIX with a method that allows passing an arbitrary libc::clockid_t and/or libc::timespec. That would let people choose whatever clock they want to use.

This is problematic for Darwin, although I guess it could be used elsewhere.

joshtriplett:
@thomcc Can you elaborate on why that would be problematic for Darwin?

thomcc:
On darwin, the clock_gettime-alikes are only in Darwin 16 (macOS 10.12) and later. Before that you just have `mach_absolute_time`. This is recent enough for unchecked usage to be a bit of a problem.

Also, the versions that use timespecs are disrecommended in favor of the `_nsec_np` versions (which return direct nanoseconds), although I don't expect there's much likelihood that they'll remove the timespec-using versions, so this part probably doesn't matter.

joshtriplett:
We discussed this in today's @rust-lang/libs-api meeting. We decided that we don't want to make this change for multiple reasons:
- It isn't portable (and this PR doesn't address that)
- It doesn't work on all versions of the Linux kernel that we still support.
- This behavior is already inconsistent across platforms, and we can't make it consistently include suspend time like this.
- We're not sure that this is the desired behavior, in any case, and the inconsistency seems worse.

Thus, we're going to close this issue.

However, we'd be interested in seeing work that makes Instant consistent in the other direction: if it's possible to *ignore* suspended time on all versions of Windows, we'd love to see a patch that switches to doing that. (cc @Amanieu who had some ideas about that in the meeting)

maurer:
> Having a huge discontinuity over suspends isn't necessarily always the desirable behavior. If you're trying to do something like "update based on the amount of time that has passed", or "measure how long a computation took", you may want to _skip_ suspended time.

Part of my disagreement with this being useful is that this does not measure the amount of time passed (the system being suspended doesn't stop time), and doesn't measure how long a computation took (this clock keeps ticking even if your process gets SIGSTOP'd, or the kernel has other higher priority threads to evaluate). The only correct use I've seen for `CLOCK_MONOTONIC` that doesn't work with `CLOCK_BOOTTIME` is availability watchdogs.

> 
> I personally would like more of an ability to use other clocks. But I don't think we should change the behavior of `now`, particularly if we then get inconsistencies across platforms.
> 
> Perhaps we should have an `InstantExt` on UNIX with a method that allows passing an arbitrary `libc::clockid_t` and/or `libc::timespec`. That would let people choose whatever clock they want to use.

If we choose to go the route of "provide other clocks as things other than `Instant"`, I don't think it should be through an OS-specific interface. I think it would make more sense to provide "A clock that ticks during suspend, and advances monotonically at a constant rate" "A clock which only runs while the machine is running" etc. as generic interfaces with OS-specific implementations when possible. the `InstantExt` route would require a user who wants a clock that ticks in suspend to write different code for Linux, Windows, and any UNIX-like that uses its own name for `CLOCK_BOOTTIME` in their application.

As an aside, if we did go the `InstantExt` route, I'd at least want the source clock included in the `Instant` to have dynamic checks to prevent accidental `.duration_since()` between different clocks.

joshtriplett:
On Wed, Aug 18, 2021 at 01:12:01PM -0700, Matthew Maurer wrote:
> > Perhaps we should have an `InstantExt` on UNIX with a method that allows passing an arbitrary `libc::clockid_t` and/or `libc::timespec`. That would let people choose whatever clock they want to use.
> 
> If we choose to go the route of "provide other clocks as things other than `Instant"`, I don't think it should be through an OS-specific interface. I think it would make more sense to provide "A clock that ticks during suspend, and advances monotonically at a constant rate" "A clock which only runs while the machine is running" etc. as generic interfaces with OS-specific implementations when possible. the `InstantExt` route would require a user who wants a clock that ticks in suspend to write different code for Linux, Windows, and any UNIX-like that uses its own name for `CLOCK_BOOTTIME` in their application.
> 
> As an aside, if we did go the `InstantExt` route, I'd at least want the source clock included in the `Instant` to have dynamic checks to prevent accidental `.duration_since()` between different clocks.

I'm not even sure if InstantExt is the right interface for that; that
was just one possibility. Some of the clocks available don't actually
make sense as an Instant; for instance, some need the ability to access
the raw value to make sense of them (e.g.  CLOCK_PROCESS_CPUTIME_ID), so
Instant as an opaque type doesn't work well there.

But I'm also not sure we can provide a sufficiently portable
implementation of "a clock that ticks during suspend", either.

maurer:
> But I'm also not sure we can provide a sufficiently portable implementation of "a clock that ticks during suspend", either.

Unless I've missed something, we can do that on Linux > 2.6.39, OSX > 10.12, and Windows (not sure of versions, I think the function we're using changes behavior if you go back far enough but I'm not certain), without much difficulty. That seems fairly portable...

joshtriplett:
On Wed, Aug 18, 2021 at 01:35:30PM -0700, Matthew Maurer wrote:
> > But I'm also not sure we can provide a sufficiently portable implementation of "a clock that ticks during suspend", either.
> 
> Unless I've missed something, we can do that on Linux > 2.6.39, OSX > 10.12, and Windows (not sure of versions, I think the function we're using changes behavior if you go back far enough but I'm not certain), without much difficulty. That seems fairly portable...

What about other UNIX systems?

What should this implementation do on older Linux, older macOS, or older
Windows, all of which we still officially support?

I think it'd be appropriate for such an implementation to start in the
crate ecosystem.

workingjubilee:
> Since you seem to know more about FreeBSD than I do, do you have a suggestion for which clock we should use there? Reading through the [man page](https://www.freebsd.org/cgi/man.cgi?query=clock_gettime&sektion=2&apropos=0&manpath=FreeBSD+14.0-current) and a [bug](https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=236702) where someone gets confused by the (admittedly documented) behavior of `CLOCK_UPTIME`, it sounds like there's currently no monotonically increasing clock which is also inclusive of suspend on FreeBSD. I'm tempted to just set them to `CLOCK_MONOTONIC` and add a comment in the documentation that some operating systems (insert a list there) don't support including sleep, and in those cases we just use a regular monotonic clock.

@maurer: I am not aware of the best solution here, unfortunately. I am aware of what is POSIX and can smell a Linux-only extension at a thousand paces, and I believe that smaller platforms should be treated with appropriate respect, or at least, should successfully compile, so the introduction of Linux-only code that did not apply and would not compile, in a way that the full CI run would reject, was apparent to me in a way I could not help but note.

> > But I'm also not sure we can provide a sufficiently portable implementation of "a clock that ticks during suspend", either.
> 
> Unless I've missed something, we can do that on Linux >= 2.6.39, OSX >= 10.12, and Windows (not sure of versions, I think the function we're using changes behavior if you go back far enough but I'm not certain), without much difficulty. That seems fairly portable...

"fairly" is not "is", unfortunately.

Amanieu:
Support for multiple clocks should be implemented using a generic `Instant` type over a `Clock` trait, like [C++ does](https://en.cppreference.com/w/cpp/chrono/time_point). It should not be hacked on to the existing `Instant` type.

Perhaps this is a good starting point for a third party crate which provides support for multiple clocks with different properties.

Amanieu:
> (cc @Amanieu who had some ideas about that in the meeting)

Actually I looked into it and it's only possible with some undocumented NT interfaces which aren't even stable across Windows versions. So that's probably right out.

CryZe:
I've looked deeper into this and... this seems to literally be a Linux specific bug. On all other OSs Instant seems to continue counting the time during suspension (`CLOCK_MONOTONIC` does the right thing on other Unixes including macOS and iOS). And as this PR intended, `CLOCK_BOOTTIME` trivially resolves the bug. Also portability to old Linux kernels isn't a problem either, you just fall back to `CLOCK_MONOTONIC` if it errors (caching this so you can directly skip to `CLOCK_MONOTONIC` the next time). So with that said, I propose to reopen this PR (or I can open a fixed one).

workingjubilee:
Thank you, CryZe, for doing the work there to verify the appropriate setting is indeed `CLOCK_MONOTONIC` on OpenBSD. Given that BSDs have some consistencies on these marks, it seems plausible to me that this is a correct assessment on all BSDs. I also believe that since FreeBSD has `CLOCK_UPTIME` as a POSIX extension, we can take it as "in contrast" to `_MONOTONIC`, though verification would be ideal.

I would like to note that on OpenBSD, however, the absolute value is still considered "meaningless" in a formal sense. It only "accidentally" aligns with a correct, monotonically increasing clock. So like Linux, [they also introduced `CLOCK_BOOTTIME`](https://marc.info/?l=openbsd-tech&m=151329884315997&w=2), which is functionally identical to `_MONOTONIC` but has attached the formal promise that it preserves between suspension, and that should be preferred on OpenBSD where possible.  Consider https://man.openbsd.org/clock_gettime.2

But I think this only reinforces the point, however, that becoming consistent with "wall time" is most intuitive for the default Instant type. I believe it is also consistent with the goal of `_MONOTONIC` which is essentially to enable precise, nanosecond sleep on systems with [`clock_nanosleep`](https://pubs.opengroup.org/onlinepubs/9699919799.2008edition/functions/clock_nanosleep.html#).

joshtriplett:
If changing the Linux behavior in this regard would indeed result in consistency across all targets, then I think it'd be a reasonable change. (I'd love to have the Linux behavior on all platforms, either by default or via some other call, but it sounds like that's much harder.)

However, I don't think we should implement caching behavior; that would optimize for the performance of repeated calls on ancient Linux at the expense of a slight hit on non-ancient Linux. I'd suggest, instead, that we do one of two things:

- Always fall back from `CLOCK_BOOTTIME` to `CLOCK_MONOTONIC` on Linux if we get `EINVAL`. (Optionally omit the fallback code on Linux targets whose minimum kernel is *already* more than 2.6.39, which as far as I can tell means every non-x86 target.)
- Consider bumping the minimum Linux kernel requirement from 2.6.32+ to 2.6.39+. 2.6.32 was released 2009-12-02, and 2.6.39 was released 2011-05-18. I think it's reasonable to require at least a ten-year-old kernel that's present not only in current enterprise distributions but in *previous* versions of enterprise distributions. Does anyone actually need support for kernels 2.6.32 to 2.6.38? cc @cuviper.

joshtriplett:
@CryZe If you're interested in submitting a pull request implementing this, with an explanation in the code and commit message based on your research into the behavior of various operating systems, I'd be happy to review it.

cuviper:
> Does anyone actually need support for kernels 2.6.32 to 2.6.38?

I do, but only because our internal RHEL builders run on a kernel from the previous major release, so my RHEL 7 builds are running on RHEL 6's 2.6.32. So it's not something I need for customers, but it's important for me nonetheless.

I know it's a pain though, and I'm willing to shoulder the burden of writing/maintaining compatibility code in such cases. Just please notify me (as you did here) when there's something that needs such help.

joshtriplett:
@cuviper Alright. In that case, I think we should just always fallback on EINVAL.

the8472:
2.6.39 would also unlock `O_PATH` which makes it easier to use the `-at` version of syscalls.

joshtriplett:
@the8472 Based on further discussions with @cuviper, it sounds like upgrading the baseline version of the Linux kernel would require additional work and coordination. I'm hoping that may be possible at some point, but let's not block this proposed change on that, since it'd just take adding an unconditional fallback on `EINVAL`.

hellow554:
Wait a second, this isn't the issue, this is the PR... let's try that again there.

@rustbot release

Issue #84186, opened 2021-04-14
Tracking Issue for #82703 #![feature(nonzero_ops)]


iago-lito:
Feature gate: `#![feature(nonzero_ops)]`

This is the tracking issue for #82703. The common arithmetic operators that keep the invariants of `NonZero*` types enforced are implemented for these types.

### Public API

Example with `U8`/`I8`, but same for `U16`/`I16` *etc.*
```rust
use std::num::{NonZeroU8, NonZeroI8};

impl NonZeroU8 {
    pub fn checked_add(self, other: U8) -> Option<Self>;
    pub fn saturating_add(self, other: U8) -> Self;
    pub unsafe fn unchecked_add(self, other: U8) -> Self;
    pub fn checked_next_power_of_two(self) -> Option<Self>;
}

impl NonZeroI8 {
    pub fn abs(self) -> Self;
    pub fn checked_abs(self) -> Option<Self>;
    pub fn overflowing_abs(self) -> (Self, bool);
    pub fn saturating_abs(self) -> Self;
    pub fn wrapping_abs(self) -> Self;
    pub fn unsigned_abs(self) -> U8;
}

impl NonZeroU8 /* and NonZeroI8 */ {
    pub fn checked_mul(self, other: Self) -> Option<Self>;
    pub fn saturating_mul(self, other: Self) -> Self;
    pub unsafe fn unchecked_mul(self, other: Self) -> Self;
    pub fn checked_pow(self, other: u32) -> Option<Self>;
    pub fn saturating_pow(self, other: u32) -> Self;
}
```

### Steps / History

- [x] Implementation: #82703
- [ ] Final commenting period (FCP)
- [ ] Stabilization PR
- [ ] Add `impl Neg for NonZeroI*`, which could not be feature-gated.

### Unresolved Questions

- None yet.

iago-lito:
> -> Leave as-is (non-const) until rust-lang/rfcs#3016 lands. (#82703 (comment))

How to not forget re-checking the constness of unchecked methods when this happens? Is there a way to leave some kind of mark in the repo that'll automatically trigger this discussion again when it's time?

Nilstrieb:
I just had a use for this, what's the state of this feature?

iago-lito:
[...]

PR #87910, opened 2021-08-10
Mark unsafe methods NonZero*::unchecked_(add|mul) as const.

iago-lito:
Now that https://github.com/rust-lang/rfcs/pull/3016 has landed, these two unstable `std` function can be marked `const`, according to this detail of #84186.


iago-lito:
(oh, I think I have to write this, right?)

r? @poliorcetics
r? @m-ou-se

poliorcetics:
There can only be one reviewer I think ?

Anyway, the changes looks good

LingMan:
Might be a good idea to `s/unsafe_checked/unchecked/` in both the commit message and the PR title.

PR #87913, opened 2021-08-10
Specialize `Vec::clone_from` for `Copy` types

a1phyr:
This should improve performance and reduce code size.

This also improves `clone_from` for `String`, `OsString` and `PathBuf`.

### DIFF
--- a/library/alloc/src/vec/mod.rs
+++ b/library/alloc/src/vec/mod.rs
@@ -2378,6 +2378,32 @@ impl<T, A: Allocator> ops::DerefMut for Vec<T, A> {
     }
 }
 
+trait SpecCloneFrom {
+    fn clone_from(this: &mut Self, other: &Self);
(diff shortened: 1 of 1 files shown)


dtolnay (changes requested):
Looks good. SpecCloneFrom needs to be feature gated on `not(no_global_oom_handling)` though to fix the failed build.

```console
error[E0599]: no method named `extend_from_slice` found for mutable reference `&mut Vec<T, A>` in the current scope
```

<br>

> This should improve performance and reduce code size.

Would you be able to put together a microbenchmark measuring either of these things?

a1phyr:
Here is a godbolt link for comparison: https://rust.godbolt.org/z/6oGTYs9Ys

Specialization removes a branch and a `memcpy` call.

dtolnay (approved):

PR #87915, opened 2021-08-10
Use smaller spans for some structured suggestions

estebank:
Use more accurate suggestion spans for

* argument parse error
* fully qualified path
* missing code block type
* numeric casts


estebank:
ping

PR #87916, opened 2021-08-10
Implement `black_box` using intrinsic

nbdd0121:
Introduce `black_box` intrinsic, as suggested in https://github.com/rust-lang/rust/pull/87590#discussion_r680468700.

This is still codegenned as empty inline assembly for LLVM. For MIR interpretation and cranelift it's treated as identity.

cc @Amanieu as this is related to inline assembly
cc @bjorn3 for rustc_codegen_cranelift changes
cc @RalfJung as this affects MIRI

r? @nagisa I suppose

### DIFF
--- a/compiler/rustc_codegen_cranelift/src/intrinsics/mod.rs
+++ b/compiler/rustc_codegen_cranelift/src/intrinsics/mod.rs
@@ -1136,6 +1136,11 @@ pub(crate) fn codegen_intrinsic_call<'tcx>(
                 };
             ret.write_cvalue(fx, CValue::by_val(is_eq_value, ret.layout()));
         };
+
+        black_box, (c a) {
+            // FIXME implement black_box semantics
+            ret.write_cvalue(fx, a);
+        };
     }
 
     if let Some((_, dest)) = destination {
--- a/compiler/rustc_codegen_llvm/src/intrinsic.rs
+++ b/compiler/rustc_codegen_llvm/src/intrinsic.rs
@@ -7,6 +7,7 @@ use crate::type_of::LayoutLlvmExt;
 use crate::va_arg::emit_va_arg;
 use crate::value::Value;
 
+use rustc_ast as ast;
 use rustc_codegen_ssa::base::{compare_simd_types, wants_msvc_seh};
 use rustc_codegen_ssa::common::span_invalid_monomorphization_error;
 use rustc_codegen_ssa::common::{IntPredicate, TypeKind};
@@ -325,6 +326,31 @@ impl IntrinsicCallMethods<'tcx> for Builder<'a, 'll, 'tcx> {
                     let cmp = self.call_intrinsic("memcmp", &[a_ptr, b_ptr, n]);
                     self.icmp(IntPredicate::IntEQ, cmp, self.const_i32(0))
                 }
+            }
+
+            sym::black_box => {
+                args[0].val.store(self, result);
+
+                // We need to "use" the argument in some way LLVM can't introspect, and on
+                // targets that support it we can typically leverage inline assembly to do
+                // this. LLVM's interpretation of inline assembly is that it's, well, a black
+                // box. This isn't the greatest implementation since it probably deoptimizes
+                // more than we want, but it's so far good enough.
+                crate::asm::inline_asm_call(
+                    self,
+                    "",
+                    "r,~{memory}",
+                    &[result.llval],
+                    self.type_void(),
+                    true,
+                    false,
+                    ast::LlvmAsmDialect::Att,
+                    &[span],
+                )
+                .unwrap_or_else(|| bug!("failed to generate inline asm call for `black_box`"));
+
+                // We have copied the value to `result` already.
+                return;
             }
 
             _ if name_str.starts_with("simd_") => {
--- a/library/core/src/hint.rs
+++ b/library/core/src/hint.rs
@@ -152,23 +152,19 @@ pub fn spin_loop() {
 /// backend used. Programs cannot rely on `black_box` for *correctness* in any way.
 ///
 /// [`std::convert::identity`]: crate::convert::identity
-#[cfg_attr(not(miri), inline)]
-#[cfg_attr(miri, inline(never))]
+#[inline]
 #[unstable(feature = "bench_black_box", issue = "64102")]
-#[cfg_attr(miri, allow(unused_mut))]
+#[cfg_attr(not(bootstrap), allow(unused_mut))]
 pub fn black_box<T>(mut dummy: T) -> T {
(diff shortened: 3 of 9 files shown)


compiler/rustc_codegen_llvm/src/intrinsic.rs
```diff
@@ -327,6 +328,31 @@ impl IntrinsicCallMethods<'tcx> for Builder<'a, 'll, 'tcx> {
...
+                    "",
+                    "r,~{memory}",
+                    &[result.llval],
+                    self.type_void(),
+                    true,
+                    false,
+                    ast::LlvmAsmDialect::Att,
+                    &[span],
+                )
+                .unwrap();
```
nagisa, line 350: Consider `unwrap_or_else(|| bug!(...))`?

nagisa:
Looks pretty nice. And in the future we have freedom to implement this in a better or otherwise backend specific manner better.

nagisa:
@bors delegate=nbdd0121

(feel free to r=nagisa once the `unwrap` is `bug!`ified)

library/core/src/hint.rs
```diff
@@ -152,23 +152,18 @@ pub fn spin_loop() {
 /// backend used. Programs cannot rely on `black_box` for *correctness* in any way.
 ///
 /// [`std::convert::identity`]: crate::convert::identity
-#[cfg_attr(not(miri), inline)]
```
RalfJung, line 155: You lost the `inline` attribute.

library/core/src/hint.rs
```diff
@@ -152,23 +152,18 @@ pub fn spin_loop() {
...
     unsafe {
-        // FIXME: Cannot use `asm!` because it doesn't support MIPS and other architectures.
         llvm_asm!("" : : "r"(&mut dummy) : "memory" : "volatile");
+        dummy
     }
 
-    dummy
+    #[cfg(not(bootstrap))]
+    {
+        crate::intrinsics::black_box(dummy)
```
RalfJung, line 167: Longer-term we might even want to directly reexport the intrinsic instead of using a wrapper function... but that's blocked on the fn ptr reification PR.

nbdd0121: I guess that's also blocked by path component stability

RalfJung: Well, we already reexport `transmute` and probably will reexport `copy`/`copy_nonoverlapping` soon (we did but it was reverted due to the fn ptr reification problem).

So, while the lack of path component stability is a problem, it's not necessarily a blocker.

compiler/rustc_codegen_cranelift/src/intrinsics/mod.rs
```diff
@@ -1136,6 +1136,10 @@ pub(crate) fn codegen_intrinsic_call<'tcx>(
                 };
             ret.write_cvalue(fx, CValue::by_val(is_eq_value, ret.layout()));
         };
+
+        black_box, (c a) {
+            ret.write_cvalue(fx, a);
```
bjorn3, line 1141:
```suggestion
            // FIXME implement black_box semantics
            ret.write_cvalue(fx, a);
```

nbdd0121: https://github.com/rust-lang/rust/blob/fbc5f67c8926e18f47cbf94596f61280ee7331df/compiler/rustc_codegen_cranelift/src/inline_asm.rs#L21-L22

Oh, I thought that's the desired semantics from the above lines. Should I add a FIXME there as well?

bjorn3: That branch will become obsolete once this PR is merged. Strictly speaking the current implementation works fine right now as cg_clif doesn't inline the `std::hint::black_box` function, but in the future once mir inlining is enabled (or the `black_box` intrinsic gets directly exported) and cranelift gets better optimizations it will become important to change this to an actual black box.

nbdd0121: Should I remove that branch then?

bjorn3: I can remove it after this PR.

nbdd0121:
> ```
> failures:
> 
> ---- [ui] ui/sanitize/memory.rs stdout ----
> 
> error: error pattern ' in the stack frame of function 'random'' not found!
> command: "/checkout/obj/build/x86_64-unknown-linux-gnu/test/ui/sanitize/memory/a"
> stdout:
> ------------------------------------------
> 
> 
> ------------------------------------------
> stderr:
> ------------------------------------------
> ==10866==WARNING: MemorySanitizer: use-of-uninitialized-value
>     #0 0x5565d1c6fb44  (/checkout/obj/build/x86_64-unknown-linux-gnu/test/ui/sanitize/memory/a+0x67b44)
>     #1 0x7f93163a10b2  (/lib/x86_64-linux-gnu/libc.so.6+0x270b2)
>     #2 0x5565d1c111cd  (/checkout/obj/build/x86_64-unknown-linux-gnu/test/ui/sanitize/memory/a+0x91cd)
> 
>   Uninitialized value was created by an allocation of 'r.i' in the stack frame of function 'main'
>     #0 0x5565d1c6fa40  (/checkout/obj/build/x86_64-unknown-linux-gnu/test/ui/sanitize/memory/a+0x67a40)
> 
> SUMMARY: MemorySanitizer: use-of-uninitialized-value (/checkout/obj/build/x86_64-unknown-linux-gnu/test/ui/sanitize/memory/a+0x67b44) 
> 
> ------------------------------------------
> ```

It seems to me that the codegen is better with the intrinsic approach and a memcpy is eliminated from `random`, so the place where uninit values end up getting used becomes `main`.

Previously it's like (expressed in MIR):
```rust
fn black_box(_1: T) -> T {
    let mut _0: T; // return place
    let mut _2: &mut T;

	bb0: {
	      _2 = &mut _1;
		  llvm_asm!(""::"r"(_2):"memory":"volatile");
	      _0 = move _1; // LLVM can't optimize out
		  return;
	}
}
```

The intrinsic approach it's codegenned more like:
```rust
fn black_box(_1: T) -> T {
    let mut _0: T; // return place
    let mut _2: &mut T;

	bb0: {
		  _0 = move _1; // LLVM can optimize out
	      _2 = &mut _0;
		  llvm_asm!(""::"r"(_2):"memory":"volatile");
		  return;
	}
}
```

I think it'd be sufficient just to change the pattern so that "main" is also matched?

nbdd0121:
@nagisa A sanitize test needs to be fixed (last commit). PTAL

nagisa:
Please squash the commits. r=me once its done.

Issue #87920, opened 2021-08-10
Tracking Issue for `Saturating` type


kellerkindt:
This is a tracking issue for the RFC "928" (rust-lang/rfcs#928).
The feature gate for the issue is `#![feature(saturating_int_impl)]`.

### About tracking issues

Tracks progress for `Saturating` type analogue to #32463

### Unresolved Questions

 - [ ] What is the MVP for stabilization?
 - [ ] What is the correct impl of `Shl` and `ShalAssign` for `Saturating<signed>`
       https://github.com/rust-lang/rust/pull/87921#discussion_r695870065
     Possible solution: https://github.com/rust-lang/rust/issues/87920#issuecomment-913015985
     @avitex https://github.com/avitex/rfcs/blob/master/text/0000-saturating-bit-shifts.md

### Steps

- [x] implement (#87921)
- [x] Stabilize `saturating_div` (#89381)
- [ ] Split of saturating bit shifts (#103441)
- [ ] Stabilize `saturating_int_impl`

[stabilization-guide]: https://rustc-dev-guide.rust-lang.org/stabilization_guide.html#stabilization-pr
[doc-guide]: https://rustc-dev-guide.rust-lang.org/stabilization_guide.html#documentation-prs

### Implementation history

 - Implementation #87921
   - Adds `struct Saturating` and `feature(saturating_int_impl)`
 - Stabilization (TODO)

joshtriplett:
FCP for stabilizing `saturating_div`:
@rfcbot merge

Amanieu:
What about the remaining unresolved question regarding the correct behavior of `Shl`?

kellerkindt:
@kennytm outlined possible implementations for that (https://github.com/rust-lang/rust/pull/87921#discussion_r695870065), but I am not sure what seems to be the correct implementation and I would be happy for any input in that regards.

Amanieu:
[...]

PR #87921, opened 2021-08-10
Add Saturating type (based on Wrapping type)

kellerkindt:
Tracking #87920

### Unresolved Questions

 - [x] ~`impl Div for Saturating<T>` falls back on inner integer division - which seems alright?~
    - [x] add `saturating_div`? (to respect division by `-1`)
 - [x] There is no `::saturating_shl` and `::saturating_shr`. (How to) implement `Shl`, `ShlAssign`, `Shr` and `ShrAssign`?
   - [naively](3f7d2ce28f8cf4dec56bf65fa2e6da0cf329ec55)
 - [x] ~`saturating_neg` is only implemented on [signed integer types](https://doc.rust-lang.org/std/?search=saturating_n)~
 - [x] Is the implementation copied over from the `Wrapping`-type correct for `Saturating`?
   - [x] `Saturating::rotate_left`
   - [x] `Saturating::rotate_right`
   - [x] `Not`
   - [x] `BitXorOr` and `BitXorOrAssign`
   - [x] `BitOr` and `BitOrAssign`
   - [x] `BitAnd` and `BitAndAssign`
   - [x] `Saturating::swap_bytes`
   - [x] `Saturating::reverse_bits`

### DIFF
--- a/library/core/src/num/int_macros.rs
+++ b/library/core/src/num/int_macros.rs
@@ -946,14 +946,9 @@ macro_rules! int_impl {
                       without modifying the original"]
         #[inline]
         pub const fn saturating_div(self, rhs: Self) -> Self {
-            let (result, overflowed) = self.overflowing_div(rhs);
-
-            if !overflowed {
-                result
-            } else if (self < 0) == (rhs < 0) {
-                Self::MAX
-            } else {
-                Self::MIN
+            match self.overflowing_div(rhs) {
+                (result, false) => result,
+                (_result, true) => Self::MAX, // MIN / -1 is the only possible saturating overflow
             }
         }
 
--- a/library/core/src/num/int_macros.rs
+++ b/library/core/src/num/int_macros.rs
@@ -926,21 +926,34 @@ macro_rules! int_impl {
         /// Basic usage:
         ///
         /// ```
+        /// #![feature(saturating_div)]
+        ///
+        #[doc = concat!("assert_eq!(5", stringify!($SelfT), ".saturating_div(2), 2);")]
+        #[doc = concat!("assert_eq!(", stringify!($SelfT), "::MAX.saturating_div(-1), ", stringify!($SelfT), "::MIN + 1);")]
+        #[doc = concat!("assert_eq!(", stringify!($SelfT), "::MIN.saturating_div(-1), ", stringify!($SelfT), "::MAX);")]
         ///
         /// ```
-        #[unstable(feature = "saturating_int_impl", issue = "87920")]
-        #[rustc_const_unstable(feature = "saturating_int_impl", issue = "87920")]
+        ///
+        /// ```should_panic
+        /// #![feature(saturating_div)]
+        ///
+        #[doc = concat!("let _ = 1", stringify!($SelfT), ".saturating_div(0);")]
+        ///
+        /// ```
+        #[unstable(feature = "saturating_div", issue = "87920")]
+        #[rustc_const_unstable(feature = "saturating_div", issue = "87920")]
         #[must_use = "this returns the result of the operation, \
                       without modifying the original"]
         #[inline]
         pub const fn saturating_div(self, rhs: Self) -> Self {
-            match self.checked_div(rhs) {
-                Some(x) => x,
-                None => if (self < 0) == (rhs < 0) {
-                    Self::MAX
-                } else {
-                    Self::MIN
-                }
+            let (result, overflowed) = self.overflowing_div(rhs);
+
+            if !overflowed {
+                result
+            } else if (self < 0) == (rhs < 0) {
+                Self::MAX
+            } else {
+                Self::MIN
             }
         }
 
--- a/library/core/src/num/int_macros.rs
+++ b/library/core/src/num/int_macros.rs
@@ -918,6 +918,32 @@ macro_rules! int_impl {
             }
         }
 
+        /// Saturating integer division. Computes `self / rhs`, saturating at the
+        /// numeric bounds instead of overflowing.
+        ///
+        /// # Examples
+        ///
+        /// Basic usage:
+        ///
+        /// ```
+        ///
+        /// ```
+        #[unstable(feature = "saturating_int_impl", issue = "87920")]
+        #[rustc_const_unstable(feature = "saturating_int_impl", issue = "87920")]
+        #[must_use = "this returns the result of the operation, \
+                      without modifying the original"]
+        #[inline]
+        pub const fn saturating_div(self, rhs: Self) -> Self {
+            match self.checked_div(rhs) {
+                Some(x) => x,
+                None => if (self < 0) == (rhs < 0) {
+                    Self::MAX
+                } else {
+                    Self::MIN
+                }
+            }
+        }
+
         /// Saturating integer exponentiation. Computes `self.pow(exp)`,
         /// saturating at the numeric bounds instead of overflowing.
         ///
--- a/library/core/src/num/saturating.rs
+++ b/library/core/src/num/saturating.rs
@@ -80,39 +80,6 @@ impl<T: fmt::UpperHex> fmt::UpperHex for Saturating<T> {
     }
 }
 
-#[allow(unused_macros)]
-macro_rules! sh_impl_signed {
-    ($t:ident, $f:ident) => {
-
-    };
-}
-
-macro_rules! sh_impl_unsigned {
-    ($t:ident, $f:ident) => {
-
-    };
-}
-
-// FIXME (#23545): uncomment the remaining impls
-macro_rules! sh_impl_all {
-    ($($t:ident)*) => ($(
-        //sh_impl_unsigned! { $t, u8 }
-        //sh_impl_unsigned! { $t, u16 }
-        //sh_impl_unsigned! { $t, u32 }
-        //sh_impl_unsigned! { $t, u64 }
-        //sh_impl_unsigned! { $t, u128 }
-        sh_impl_unsigned! { $t, usize }
-
-        //sh_impl_signed! { $t, i8 }
-        //sh_impl_signed! { $t, i16 }
-        //sh_impl_signed! { $t, i32 }
-        //sh_impl_signed! { $t, i64 }
-        //sh_impl_signed! { $t, i128 }
-        //sh_impl_signed! { $t, isize }
-    )*)
-}
-
-sh_impl_all! { u8 u16 u32 u64 u128 usize i8 i16 i32 i64 i128 isize }
 
 // FIXME(30524): impl Op<T> for Saturating<T>, impl OpAssign<T> for Saturating<T>
 macro_rules! saturating_impl {
--- a/library/core/src/num/saturating.rs
+++ b/library/core/src/num/saturating.rs
@@ -141,7 +141,7 @@ macro_rules! sh_impl_unsigned {
 
             #[inline]
             fn shl(self, other: $f) -> Saturating<$t> {
-                Saturating(self.0.shl((other & self::shift_max::$t as $f) as u32))
+                Saturating(self.0.wrapping_shl(other as u32))
             }
         }
         forward_ref_binop! { impl Shl, shl for Saturating<$t>, $f,
@@ -162,7 +162,7 @@ macro_rules! sh_impl_unsigned {
 
             #[inline]
             fn shr(self, other: $f) -> Saturating<$t> {
-                Saturating(self.0.shr((other & self::shift_max::$t as $f) as u32))
+                Saturating(self.0.wrapping_shr(other as u32))
             }
         }
         forward_ref_binop! { impl Shr, shr for Saturating<$t>, $f,
--- a/library/core/src/num/saturating.rs
+++ b/library/core/src/num/saturating.rs
@@ -36,44 +36,44 @@ use crate::ops::{Sub, SubAssign};
 #[unstable(feature = "saturating_int_impl", issue = "87920")]
 #[derive(PartialEq, Eq, PartialOrd, Ord, Clone, Copy, Default, Hash)]
 #[repr(transparent)]
-pub struct Saturating<T>(#[stable(feature = "rust1", since = "1.0.0")] pub T);
+pub struct Saturating<T>(#[unstable(feature = "saturating_int_impl", issue = "87920")] pub T);
 
-#[stable(feature = "rust1", since = "1.0.0")]
+#[unstable(feature = "saturating_int_impl", issue = "87920")]
 impl<T: fmt::Debug> fmt::Debug for Saturating<T> {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
         self.0.fmt(f)
     }
 }
 
-#[stable(feature = "saturating_display", since = "1.10.0")]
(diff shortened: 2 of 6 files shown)


library/core/src/num/saturating.rs
```diff
@@ -0,0 +1,777 @@
...
+use crate::ops::{Mul, MulAssign, Neg, Not};
+use crate::ops::{Sub, SubAssign};
+
+/// Provides intentionally-wrapped arithmetic on `T`.
+///
+/// Operations like `+` on `u32` values are intended to never overflow,
+/// and in some debug configurations overflow is detected and results
+/// in a panic. While most arithmetic falls into this category, some
+/// code explicitly expects and relies upon modular arithmetic (e.g.,
+/// hashing).
```
kennytm, line 15: Missed copy-and-paste? `Saturating` is irrelevant to modular arithmetic.

library/core/src/num/saturating.rs
```diff
@@ -0,0 +1,777 @@
...
+///
+/// let max = Saturating(u32::MAX);
+/// let one = Saturating(1u32);
+///
+/// assert_eq!(u32::MAX, (max + one).0);
+/// ```
+#[unstable(feature = "saturating_int_impl", issue = "87920")]
+#[derive(PartialEq, Eq, PartialOrd, Ord, Clone, Copy, Default, Hash)]
+#[repr(transparent)]
+pub struct Saturating<T>(#[stable(feature = "rust1", since = "1.0.0")] pub T);
```
kennytm, line 39:
```suggestion
pub struct Saturating<T>(#[unstable(feature = "saturating_int_impl", issue = "87920")] pub T);
```

library/core/src/num/saturating.rs
```diff
@@ -0,0 +1,777 @@
...
+
+#[stable(feature = "rust1", since = "1.0.0")]
+impl<T: fmt::Debug> fmt::Debug for Saturating<T> {
```
kennytm, line 42: all these `impl` should be `unstable` with feature `saturating_int_impl`

library/core/src/num/saturating.rs
```diff
@@ -0,0 +1,777 @@
...
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        self.0.fmt(f)
+    }
+}
+
+#[stable(feature = "saturating_display", since = "1.10.0")]
+impl<T: fmt::Display> fmt::Display for Saturating<T> {
```
kennytm, line 49: ditto

library/core/src/num/saturating.rs
```diff
@@ -0,0 +1,777 @@
...
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        self.0.fmt(f)
+    }
+}
+
+#[stable(feature = "saturating_fmt", since = "1.11.0")]
+impl<T: fmt::Binary> fmt::Binary for Saturating<T> {
```
kennytm, line 56: ditto

library/core/src/num/saturating.rs
```diff
@@ -0,0 +1,777 @@
...
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        self.0.fmt(f)
+    }
+}
+
+#[stable(feature = "saturating_fmt", since = "1.11.0")]
+impl<T: fmt::Octal> fmt::Octal for Saturating<T> {
```
kennytm, line 63: ditto

library/core/src/num/saturating.rs
```diff
@@ -0,0 +1,777 @@
...
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        self.0.fmt(f)
+    }
+}
+
+#[stable(feature = "saturating_fmt", since = "1.11.0")]
+impl<T: fmt::LowerHex> fmt::LowerHex for Saturating<T> {
```
kennytm, line 70: ditto

library/core/src/num/saturating.rs
```diff
@@ -0,0 +1,777 @@
...
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        self.0.fmt(f)
+    }
+}
+
+#[stable(feature = "saturating_fmt", since = "1.11.0")]
+impl<T: fmt::UpperHex> fmt::UpperHex for Saturating<T> {
```
kennytm, line 77: ditto

library/core/src/num/saturating.rs
```diff
@@ -0,0 +1,777 @@
...
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        self.0.fmt(f)
+    }
+}
+
+// FIXME(30524): impl Op<T> for Saturating<T>, impl OpAssign<T> for Saturating<T>
+macro_rules! saturating_impl {
+    ($($t:ty)*) => ($(
+        #[stable(feature = "rust1", since = "1.0.0")]
+        impl Add for Saturating<$t> {
```
kennytm, line 87: ditto

library/core/src/num/saturating.rs
```diff
@@ -0,0 +1,777 @@
...
+            type Output = Saturating<$t>;
+
+            #[inline]
+            fn add(self, other: Saturating<$t>) -> Saturating<$t> {
+                Saturating(self.0.saturating_add(other.0))
+            }
+        }
+        forward_ref_binop! { impl Add, add for Saturating<$t>, Saturating<$t>,
+                #[stable(feature = "saturating_ref", since = "1.14.0")] }
```
kennytm, line 96: ditto

library/core/src/num/saturating.rs
```diff
@@ -0,0 +1,777 @@
...
+            ///
+            /// ```
+            /// #![feature(saturating_int_impl)]
+            /// use std::num::Saturating;
+            ///
+            #[doc = concat!("assert_eq!(Saturating(100", stringify!($t), ").abs(), Saturating(100));")]
+            #[doc = concat!("assert_eq!(Saturating(-100", stringify!($t), ").abs(), Saturating(100));")]
+            #[doc = concat!("assert_eq!(Saturating(", stringify!($t), "::MIN).abs(), Saturating((", stringify!($t), "::MIN + 1).abs()));")]
+            #[doc = concat!("assert_eq!(Saturating(", stringify!($t), "::MIN).abs(), Saturating(", stringify!($t), "::MIN.saturating_abs()));")]
+            /// assert_eq!(Saturating(-128i8).abs().0 as u8, i8::MAX as u8);
```
kennytm, line 654:
```suggestion
            #[doc = concat!("assert_eq!(Saturating(", stringify!($t), "::MIN).abs(), Saturating(", stringify!($t), "::MAX));")]
```

library/core/src/num/saturating.rs
```diff
@@ -0,0 +1,777 @@
...
+            ///
+            /// # Examples
+            ///
+            /// Basic usage:
+            ///
+            /// ```
+            /// #![feature(saturating_int_impl)]
+            /// use std::num::Saturating;
+            ///
+            #[doc = concat!("let n = Saturating(", stringify!($t), "::MAX) / Saturating(4", stringify!($t), ");")]
```
kennytm, line 744: why use `x / 4` instead of `x >> 2` like everywhere else.

```suggestion
            #[doc = concat!("let n = Saturating(", stringify!($t), "::MAX) >> 2;")]
```

library/core/src/num/saturating.rs
kellerkindt: Here I've just been using the test from [Wrapping](https://github.com/rust-lang/rust/blob/master/library/core/src/num/wrapping.rs#L797), I am very fine with your suggestion, though.

library/core/src/num/saturating.rs
kellerkindt: I dropped the shift impls because there are no `*::saturating_shl/r` functions like there are `*::wraping_shl/r`.

kennytm: (well there's no `saturating_bitand` either 🤷)

i guess this should be an "Unresolved Questions" then. what should be the behavior of `Saturating(1u32) << 33`?

1. keep it unimplemented (but hard to argue when `rotate_left` is implemented).
2. panic with "attempt to shift left with overflow" like `1u32 << 33`.
3. "saturate" into `Saturating(1u32 << 31)`.
4. "wrap" into `Saturating(1u32 << 1)`.
5. use the infinite-bit interpretation and return 0.

i prefer doing "2" for now which is forward-compatible with 3, 4 or 5.

kennytm: let's change this line to this for now, which doesn't rely on `impl Shr for Saturating`.

```suggestion
            #[doc = concat!("let n = Saturating(", stringify!($t), "::MAX >> 2);")]
```

kennytm:
i'm pretty sure the implementation of `BitAnd`, `rotate_{left,right}`, `swap_bytes` and `reverse_bits` are correct. the question is whether these bit-level operations are useful for `Saturating`, and if they can live independent of Shl and Shr (see discussion from https://github.com/rust-lang/rust/pull/87921#discussion_r687321524). 

while the methods can have a separate library feature, the trait impls (BitAnd, BitOr, BitXor, Shl, Shr) must be stabilized together with `Saturating`.

kellerkindt:
I added naive shift (and remainder) impls. What needs to ben done next?

kennytm:
> `saturating_eng` is only implemented on signed integer types

i don't think it makes sense to `impl Neg for Saturating<unsigned>` 😅

kellerkindt:
> > `saturating_eng` is only implemented on signed integer types
> 
> i don't think it makes sense to `impl Neg for Saturating<unsigned>` sweat_smile

Well... you are actually right about that :sweat_smile:
*fixed*

SkiFire13:
> `impl Div for Saturating<T>` falls back on inner integer division - which seems alright?

I think this is wrong in case of `iN::MIN / -1`. I would expect a saturating operation to evaluate to `iN::MAX`, instead it overflows and evaluates to `iN::MIN`. This also seems a reason to have a `saturating_div` for signed integers.

library/core/src/num/int_macros.rs
```diff
@@ -918,6 +918,32 @@ macro_rules! int_impl {
...
 
+        /// Saturating integer division. Computes `self / rhs`, saturating at the
+        /// numeric bounds instead of overflowing.
+        ///
+        /// # Examples
+        ///
+        /// Basic usage:
+        ///
+        /// ```
+        ///
```
kellerkindt, line 929: Is there any way to put some tests in here? The tests from within `Saturating` only apply to signed types...

library/core/src/num/saturating.rs
```diff
@@ -0,0 +1,978 @@
...
+        impl Div for Saturating<$t> {
+            type Output = Saturating<$t>;
+
+            #[inline]
+            fn div(self, other: Saturating<$t>) -> Saturating<$t> {
+                Saturating(self.0.saturating_div(other.0))
+            }
+        }
+        forward_ref_binop! { impl Div, div for Saturating<$t>, Saturating<$t>,
+                #[unstable(feature = "saturating_int_impl", issue = "87920")] }
```
kellerkindt, line 878: Resides here until the question about the test location from above is answered

library/core/src/num/int_macros.rs
```diff
@@ -918,6 +918,32 @@ macro_rules! int_impl {
...
+        pub const fn saturating_div(self, rhs: Self) -> Self {
+            match self.checked_div(rhs) {
+                Some(x) => x,
+                None => if (self < 0) == (rhs < 0) {
+                    Self::MAX
+                } else {
+                    Self::MIN
+                }
+            }
+        }
```
kennytm, line 945: imo we should use `overflowing_div` here. 

* https://doc.rust-lang.org/std/primitive.i32.html#method.overflowing_div
* https://doc.rust-lang.org/std/primitive.u32.html#method.overflowing_div

the current implementation makes `0 / 0 == MAX` which feels wrong to me.

library/core/src/num/int_macros.rs
```diff
@@ -918,6 +918,32 @@ macro_rules! int_impl {
...
+        /// numeric bounds instead of overflowing.
+        ///
+        /// # Examples
+        ///
+        /// Basic usage:
+        ///
+        /// ```
+        ///
+        /// ```
+        #[unstable(feature = "saturating_int_impl", issue = "87920")]
```
kennytm, line 931: use a separate feature for `iXX::saturating_div`.

library/core/src/num/int_macros.rs
kennytm: 5 / 2 = 2
MAX / -1 = MIN+1
MIN / -1 = MAX
(1 / 0 = panic)

kellerkindt: Oh, I didn't realize this is the impl for only signed types... woopsie :)

library/core/src/num/saturating.rs
```diff
@@ -266,6 +266,37 @@ macro_rules! saturating_impl {
...
+        #[doc = concat!("assert_eq!(Saturating(", stringify!($t), "::MIN), Saturating(", stringify!($t), "::MIN) / Saturating(1));")]
+        /// ```
+        ///
+        /// ```should_panic
+        /// #![feature(saturating_int_impl, saturating_div)]
+        /// use std::num::Saturating;
+        ///
+        #[doc = concat!("let _ = Saturating(0", stringify!($t), ") / Saturating(0);")]
+        /// ```
+        #[unstable(feature = "saturating_int_impl", issue = "87920")]
```
kellerkindt, line 288: Not sure whether to keep this doc(-tests)

library/core/src/num/int_macros.rs
```diff
@@ -926,21 +926,34 @@ macro_rules! int_impl {
...
-                    Self::MIN
-                }
+            let (result, overflowed) = self.overflowing_div(rhs);
+
+            if !overflowed {
+                result
+            } else if (self < 0) == (rhs < 0) {
+                Self::MAX
+            } else {
+                Self::MIN
```
kennytm, line 956:
```suggestion
            } else {
                Self::MAX
```

`MIN / -1` is the only possible saturating overflow so the branch can be simplified here.

kellerkindt:
What are  the next steps? :)

kennytm (commented):
rest LGTM

library/core/src/num/saturating.rs
```diff
@@ -0,0 +1,972 @@
...
+        impl Shl<$f> for Saturating<$t> {
+            type Output = Saturating<$t>;
+
+            #[inline]
+            fn shl(self, other: $f) -> Saturating<$t> {
+                if other < 0 {
+                    Saturating(self.0.shr((-other & self::shift_max::$t as $f) as u32))
+                } else {
+                    Saturating(self.0.shl((other & self::shift_max::$t as $f) as u32))
+                }
```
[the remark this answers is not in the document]
kellerkindt: I've gone for Ⓓ because it does not seem clear (to me) what the correct impl should look like. I'd rather have some figure it out later, than implementing it wrong now.

kennytm:
thanks!

@bors r+

Vic23M (approved):

PR #87927, opened 2021-08-10
Make concrete_opaque_types be FxHashSet<DefId>

spastorino:
r? @oli-obk 

@bors rollup=always


oli-obk:
i find it unlikely to affect perf, but let's check

@bors rollup=never

PR #87937, opened 2021-08-11
Don't mark `if_let_guard` as an incomplete feature

LeSeulArtichaut:
I don't think there is any reason for `if_let_guard` to be an incomplete feature, and I think the reason they were marked in the first place was simply because they weren't implemented at all.

r? @pnkfelix
cc tracking issue #51114


nagisa:
Having two feature gates for what seems to be a feature with no very obvious flaws seems quite excessive indeed. Just one is sufficient to signal potential issues that could remain.

r? @nagisa
@bors r+

PR #87940, opened 2021-08-11
Enable PGO for LLVM on x86_64-unknown-linux-gnu in CI

Mark-Simulacrum:
Particularly for optimized builds, rustc spends a good amount of time in LLVM
builds, so it is hopeful that this will be worthwhile.

We can likely afford to spend the extra cycles building LLVM 1-2 times per CI
build today on x86_64-unknown-linux-gnu: the builder finishes in around 50
minutes, whereas our long pole is around 2-3 hours. For try builds this is less
clear, but we will see what the impact is in this PR.

r? @ghost for now while we run perf -- and I still need to polish the code into less ad-hoc reuse of the rust pgo options for LLVM.

### DIFF
--- a/src/ci/docker/host-x86_64/dist-x86_64-linux/Dockerfile
+++ b/src/ci/docker/host-x86_64/dist-x86_64-linux/Dockerfile
@@ -82,6 +82,12 @@ COPY host-x86_64/dist-x86_64-linux/build-clang.sh /tmp/
 RUN ./build-clang.sh
 ENV CC=clang CXX=clang++
 
+ENV PERF_COMMIT 1e19fc4c6168d2f7596e512f42f358f245d8f09d
+RUN curl -LS -o perf.zip https://github.com/rust-lang/rustc-perf/archive/$PERF_COMMIT.zip && \
+    unzip perf.zip && \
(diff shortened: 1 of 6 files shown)


src/ci/docker/host-x86_64/dist-x86_64-linux/Dockerfile
```diff
@@ -82,6 +82,10 @@ COPY host-x86_64/dist-x86_64-linux/build-clang.sh /tmp/
 RUN ./build-clang.sh
 ENV CC=clang CXX=clang++
 
+ENV PERF_COMMIT 1e19fc4c6168d2f7596e512f42f358f245d8f09d
+RUN curl -LS -o perf.zip https://github.com/rust-lang/rustc-perf/archive/$PERF_COMMIT.zip
+RUN unzip perf.zip && mv rustc-perf-$PERF_COMMIT rustc-perf
```
bjorn3, line 87: You may want to do this in a single `RUN` command removing `perf.zip` afterwards to reduce the image size. Docker creates a new layer for each `RUN` command AFAIK.

Mark-Simulacrum:
Going to close this in favor of a new PR to ease discussion & review, given the large number of comments here.

PR #87944, opened 2021-08-11
add Cell::as_array_of_cells, similar to Cell::as_slice_of_cells

oconnor663:
I'd like to propose adding `Cell::as_array_of_cells`, as a natural analog to `Cell::as_slice_of_cells`. I don't have a specific use case in mind, other than that supporting slices but not arrays feels like a gap. Do other folks agree with that intuition? Would this addition be substantial enough to need an RFC?

---

Previously, converting `&mut [T; N]` to `&[Cell<T>; N]` looks like this:

```rust
let array = &mut [1, 2, 3];
let cells: &[Cell<i32>; 3] = Cell::from_mut(&mut array[..])
    .as_slice_of_cells()
    .try_into()
    .unwrap();
```

With this new helper method, it looks like this:

```rust
let array = &mut [1, 2, 3];
let cells = Cell::from_mut(array).as_array_of_cells();
```

### DIFF
--- a/library/core/src/cell.rs
+++ b/library/core/src/cell.rs
@@ -576,6 +576,26 @@ impl<T> Cell<[T]> {
     }
 }
 
+impl<T, const N: usize> Cell<[T; N]> {
+    /// Returns a `&[Cell<T>; N]` from a `&Cell<[T; N]>`
+    ///
+    /// # Examples
+    ///
+    /// ```
+    /// #![feature(as_array_of_cells)]
+    /// use std::cell::Cell;
+    ///
+    /// let mut array: [i32; 3] = [1, 2, 3];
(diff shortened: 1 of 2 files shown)


oconnor663:
I've put a standalone version of this function in this repo: https://github.com/oconnor663/cell_utils

oconnor663:
@scottmcm should I add someone else to review?

scottmcm:
Sorry for being slow.  This seems perfectly reasonable to me, given that it's already possible (much less nicely) by going through slices.  Can you [open a Library Tracking Issue](https://github.com/rust-lang/rust/issues/new/choose) for it, and add that to the `unstable` attribute?

oconnor663:
No worries. Done!

PR #87960, opened 2021-08-12
Suggest replacing an inexisting field for an unmentioned field

### LINKED ISSUE #87938: Suggest a missing field when the rest of the field is found
Given the following code: https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=2c9af7b21295b19d151a109261981cd2

```rust
enum Foo {
    Bar { alpha: u8, bravo: u8, charlie: u8 },
}

fn foo(foo: Foo) {
    match foo {
        Foo::Bar {
            alpha,
            beta, // `bravo` miswritten as `beta` here.
            charlie,
        } => todo!(),
    }
}
```

The current output is:

```
error[E0026]: variant `Foo::Bar` does not have a field named `beta`
 --> src/lib.rs:9:13
  |
9 |             beta, // `bravo` miswritten as `beta` here.
  |             ^^^^ variant `Foo::Bar` does not have this field

error[E0027]: pattern does not mention field `bravo`
  --> src/lib.rs:7:9
   |
7  | /         Foo::Bar {
8  | |             alpha,
9  | |             beta, // `bravo` miswritten as `beta` here.
10 | |             charlie,
11 | |         } => todo!(),
   | |_________^ missing field `bravo`
   |
help: include the missing field in the pattern
   |
10 |             charlie, bravo } => todo!(),
   |                    ^^^^^^^^^
help: if you don't care about this missing field, you can explicitly ignore it
   |
10 |             charlie, .. } => todo!(),
   |                    ^^^^^^

```

```
error[E0026]: variant `Foo::Bar` does not have a field named `beta`
 --> src/lib.rs:9:13
  |
9 |             beta, // `bravo` miswritten as `beta` here.
  |             ^^^^ help: maybe replacing this for a missing field: `bravo`
```

We already provide a suggestion for similar names, for example, `brav0` is suggested to replace it for `bravo`.
Analogous to that, when the rest of the field (`alpha` & `charlie`) is mentioned, suggesting to replace the miswritten field (`beta`) for the missing field (`bravo`) can be helpful for users.

hkmatsumoto:
Fix #87938

This PR adds a suggestion to replace an inexisting field for an
unmentioned field. Given the following code:
```rust
enum Foo {
    Bar { alpha: u8, bravo: u8, charlie: u8 },
}

fn foo(foo: Foo) {
    match foo {
        Foo::Bar {
            alpha,
            beta, // `bravo` miswritten as `beta` here.
            charlie,
        } => todo!(),
    }
}
```
the compiler now emits the error messages below.
```text
error[E0026]: variant `Foo::Bar` does not have a field named `beta`
 --> src/lib.rs:9:13
  |
9 |             beta, // `bravo` miswritten as `beta` here.
  |             ^^^^
  |             |
  |             variant `Foo::Bar` does not have this field
  |             help: `Foo::Bar` has a field named `bravo`: `bravo`
```

Note that this suggestion is available iff the number of inexisting
fields and unmentioned fields are both 1.

### DIFF
--- a/compiler/rustc_typeck/src/check/pat.rs
+++ b/compiler/rustc_typeck/src/check/pat.rs
@@ -1447,7 +1447,8 @@ impl<'a, 'tcx> FnCtxt<'a, 'tcx> {
                     plural
                 ),
             );
-            if plural == "" {
+
+            if unmentioned_fields.len() == 1 {
                 let input =
                     unmentioned_fields.iter().map(|(_, field)| field.name).collect::<Vec<_>>();
                 let suggested_name = find_best_match_for_name(&input, ident.name, None);
@@ -1468,6 +1469,18 @@ impl<'a, 'tcx> FnCtxt<'a, 'tcx> {
                         // We don't want to throw `E0027` in case we have thrown `E0026` for them.
                         unmentioned_fields.retain(|&(_, x)| x.name != suggested_name);
                     }
+                } else if inexistent_fields.len() == 1 {
+                    let unmentioned_field = unmentioned_fields[0].1.name;
+                    err.span_suggestion_short(
+                        ident.span,
+                        &format!(
+                            "`{}` has a field named `{}`",
+                            tcx.def_path_str(variant.def_id),
+                            unmentioned_field
+                        ),
+                        unmentioned_field.to_string(),
+                        Applicability::MaybeIncorrect,
+                    );
                 }
             }
         }
--- a/src/test/ui/issues/issue-51102.stderr
+++ b/src/test/ui/issues/issue-51102.stderr
@@ -2,7 +2,10 @@ error[E0026]: struct `SimpleStruct` does not have a field named `state`
   --> $DIR/issue-51102.rs:13:17
    |
 LL |                 state: 0,
-   |                 ^^^^^ struct `SimpleStruct` does not have this field
+   |                 ^^^^^
+   |                 |
+   |                 struct `SimpleStruct` does not have this field
+   |                 help: `SimpleStruct` has a field named `no_state_here`
 
 error[E0025]: field `no_state_here` bound multiple times in the pattern
   --> $DIR/issue-51102.rs:24:17
--- a/src/test/ui/numeric/numeric-fields.stderr
+++ b/src/test/ui/numeric/numeric-fields.stderr
@@ -16,7 +16,10 @@ error[E0026]: struct `S` does not have a field named `0x1`
   --> $DIR/numeric-fields.rs:7:17
    |
 LL |         S{0: a, 0x1: b, ..} => {}
-   |                 ^^^ struct `S` does not have this field
+   |                 ^^^
+   |                 |
+   |                 struct `S` does not have this field
+   |                 help: `S` has a field named `1`
 
 error: aborting due to 2 previous errors
 


compiler/rustc_typeck/src/check/pat.rs
```diff
@@ -1421,7 +1421,22 @@ impl<'a, 'tcx> FnCtxt<'a, 'tcx> {
...
+                let unmentioned_field = unmentioned_fields[0].1.name;
+                err.span_suggestion(
+                    ident.span,
+                    &format!(
+                        "`{}` has a field named `{}`, try replacing `{}` for it.",
+                        tcx.def_path_str(variant.def_id),
+                        unmentioned_field,
+                        inexistent_fields[0]
+                    ),
+                    format!("{}", unmentioned_field),
```
LeSeulArtichaut, line 1435: This can be simplified to
```suggestion
                    unmentioned_field.to_string(),
```

compiler/rustc_typeck/src/check/pat.rs
```diff
@@ -1421,7 +1421,22 @@ impl<'a, 'tcx> FnCtxt<'a, 'tcx> {
...
+            if inexistent_fields.len() == 1 && unmentioned_fields.len() == 1 {
+                let unmentioned_field = unmentioned_fields[0].1.name;
+                err.span_suggestion(
+                    ident.span,
+                    &format!(
+                        "`{}` has a field named `{}`, try replacing `{}` for it.",
+                        tcx.def_path_str(variant.def_id),
+                        unmentioned_field,
+                        inexistent_fields[0]
+                    ),
```
LeSeulArtichaut, line 1434: What about keeping only the start of the sentence? The error message will already have the name of the inexistant field in it.
```suggestion
                    &format!(
                        "`{}` has a field named `{}`",
                        tcx.def_path_str(variant.def_id),
                        unmentioned_field,
                    ),
```
This will read as "struct XXX does not have a field named YYY. help: XXX has a field named ZZZ", which sounds good to me (I'm definitely not an expert in diagnostics though).

Also, other nit: messages in diagnostics usually don't end with a `.`

src/test/ui/issues/issue-52717.stderr
```diff
@@ -2,10 +2,16 @@ error[E0026]: variant `A::A` does not have a field named `fob`
...
+   |            ^^^ variant `A::A` does not have this field
+   |
+help: `A::A` has a field named `foo`, try replacing `fob` for it.
+   |
+LL |     A::A { foo } => { println!("{}", fob); }
+   |            ~~~
+help: a field with a similar name exists
+   |
+LL |     A::A { foo } => { println!("{}", fob); }
+   |            ~~~
```
LeSeulArtichaut, line 14: I'm not sure we need both of these suggestions to appear in the error message. I think the second one is maybe the most accurate (?) but I could see a case for either of them to be kept.

LeSeulArtichaut (commented):
Thanks for the PR, and sorry for the delay on my part. I have a few minor comments.

src/test/ui/issues/issue-51102.stderr
```diff
@@ -3,6 +3,11 @@ error[E0026]: struct `SimpleStruct` does not have a field named `state`
    |
 LL |                 state: 0,
    |                 ^^^^^ struct `SimpleStruct` does not have this field
+   |
+help: `SimpleStruct` has a field named `no_state_here`, try replacing `state` for it.
+   |
+LL |                 no_state_here: 0,
+   |                 ~~~~~~~~~~~~~
```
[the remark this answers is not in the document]
hkmatsumoto: Do you mean to avoid emitting diagnostic when the code was something like`state: "foo"` in this case by type checking?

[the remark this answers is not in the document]
estebank: Correct. But we don't need to do that as part of this PR.

compiler/rustc_typeck/src/check/pat.rs
```diff
@@ -1468,6 +1469,18 @@ impl<'a, 'tcx> FnCtxt<'a, 'tcx> {
                         // We don't want to throw `E0027` in case we have thrown `E0026` for them.
                         unmentioned_fields.retain(|&(_, x)| x.name != suggested_name);
                     }
+                } else if inexistent_fields.len() == 1 {
+                    let unmentioned_field = unmentioned_fields[0].1.name;
+                    err.span_suggestion(
```
estebank, line 1474: One small change so that w e don't mention the field name twice in a row:
```suggestion
                    err.span_suggestion_short(
```

After that, r=me.

estebank (approved):

src/test/ui/issues/issue-51102.stderr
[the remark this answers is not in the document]
hkmatsumoto: Got it, I'll make a follow-up PR later.

compiler/rustc_typeck/src/check/pat.rs
hkmatsumoto: Applied :+1: I don't have the bors permission so please do it instead of me.

estebank:
Please squash your commits. Then we can merge this :)

src/test/ui/issues/issue-51102.stderr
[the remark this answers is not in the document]
hkmatsumoto: Opened.

PR #87971, opened 2021-08-12
Fix typo

hkmatsumoto:
I'm not 100% confident that using `code` here is wrong, but `comma` looks more natural.


hkmatsumoto:
Thanks, that cleared my confusion. Maybe I should've discussed on zulip before opening this

estebank:
No, that's absolutely fine! Glad I could clarify!

PR #87974, opened 2021-08-12
Test and fix `size_hint` for slice’s [r]split* iterators

### LINKED ISSUE #87978: Incorrect lower bound on size hint for `<[T]>::splitn` if n==0
I’m creating this issue in order to better document the bug in `std` that #87974 fixes once it’s merged.

```rs
fn this_should_hold_for_non_buggy_iterators(i: impl Iterator) {
    let (lower_hint, count) = dbg!(i.size_hint().0, i.count());
    assert!(lower_hint <= count)
}

fn main() {
    this_should_hold_for_non_buggy_iterators([()].splitn(0, |_| false));
    // [src/main.rs:3] i.size_hint().0 = 1
    // [src/main.rs:3] i.count() = 0
    // thread 'main' panicked at 'assertion failed: lower_hint <= count', src/main.rs:4:5
}
```
([playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=4e85d6a9eb58139aab5641604b52a3a6))

@rustbot label T-libs, A-slice, A-iterators

steffahn:
Adds extensive test (of `size_hint`) for all the _[r]split*_ iterators.
Fixes `size_hint` upper bound for _split_inclusive*_ iterators which was one higher than necessary for non-empty slices.
Fixes `size_hint` lower bound for _[r]splitn*_ iterators when _n == 0_, which was one too high.

**Lower bound being one too high was a logic error, violating the correctness condition of `size_hint`.**

_Edit:_ I’ve opened an issue for that bug, so this PR fixes #87978

### DIFF
--- a/library/alloc/tests/slice.rs
+++ b/library/alloc/tests/slice.rs
@@ -1001,11 +1001,7 @@ fn test_split_iterators_size_hint() {
         Lower,
         Upper,
     }
-    fn assert_precise_size_hints<I: Iterator>(
-        mut it: I,
-        which: Bounds,
-        context: impl fmt::Display,
-    ) {
+    fn assert_precise_size_hints(mut it: impl Iterator, which: Bounds, context: impl fmt::Display) {
         match which {
             Bounds::Lower => {
                 let mut lower_bounds = vec![it.size_hint().0];
--- a/library/alloc/tests/slice.rs
+++ b/library/alloc/tests/slice.rs
@@ -1001,7 +1001,7 @@ fn test_split_iterators_size_hint() {
         Lower,
         Upper,
     }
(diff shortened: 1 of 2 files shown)


steffahn:
@rust-highfive failed. I’m choosing at random from [this list](https://github.com/rust-lang/highfive/blob/c1b314239752352475aa802b5ae5fcd40f6b6fc9/highfive/configs/rust-lang/rust.json#L20-L24).

r? @dtolnay

steffahn:
Ah, the damn tidy check -.-

By the way (note for review:) I was only force-pushing to amend the new commit.

dtolnay (approved):
Thanks!

PR #87975, opened 2021-08-12
Include attributes in removal span for unused imports. 

### LINKED ISSUE #87973: `unused_imports` suggestion does not remove attributes
I tried this code:

```rust
#[deprecated]
#[allow(unsafe_code)]
#[cfg(not(foo))]
use std::fs;

fn main() {}
```

I expected to see this happen: The suggestion should suggest removing `use std::fs;` **and** its attributes.

Instead, this happened: The suggestion span does not encompass the attributes.  This causes `cargo fix` to remove the item, and leave the attributes applying to whatever happens to be next.

### Meta

`rustc 1.56.0-nightly (ccffcafd5 2021-08-11)`

See also #56328 for a similar issue (although I think that one is subtly different).

m-ou-se:
Fixes https://github.com/rust-lang/rust/issues/87973


nagisa:
@bors r+

Nice!

PR #87976, opened 2021-08-12
Account for tabs when highlighting multiline code suggestions

### LINKED ISSUE #87972: Wrong part colored in diagnostic suggestions after tabs
```rust
trait T {
	const A: &u8;
}
```
(The second line starts with a *tab*, not with spaces!)

This results in:
Screenshot 2021-08-12 at 16 48 04

It suggests to add a `'a ` after the `&`, but the green color is not applied to the `'a ` part, but instead to `A: &`, shifted a bit to the left.

cc @estebank

estebank:
Address `'\t'` case in #87972.

Before:

Screen Shot 2021-08-12 at 8 52 27 AM

After:

Screen Shot 2021-08-12 at 8 52 15 AM

### DIFF
--- a/compiler/rustc_errors/src/emitter.rs
+++ b/compiler/rustc_errors/src/emitter.rs
@@ -1679,27 +1679,19 @@ impl EmitterWriter {
                 // Colorize addition/replacements with green.
                 for &SubstitutionHighlight { start, end } in highlight_parts {
                     // Account for tabs when highlighting (#87972).
-                    let start: usize = line
-                        .chars()
-                        .take(start)
-                        .map(|ch| match ch {
-                            '\t' => 4,
-                            _ => 1,
-                        })
-                        .sum();
-
-                    let end: usize = line
-                        .chars()
-                        .take(end)
-                        .map(|ch| match ch {
-                            '\t' => 4,
-                            _ => 1,
-                        })
-                        .sum();
+                    // let tabs: usize = line
+                    //     .chars()
+                    //     .take(start)
+                    //     .map(|ch| match ch {
+                    //         '\t' => 3,
+                    //         _ => 0,
+                    //     })
+                    //     .sum();
+                    let tabs = 0;
                     buffer.set_style_range(
                         row_num,
-                        max_line_num_len + 3 + start,
-                        max_line_num_len + 3 + end,
+                        max_line_num_len + 3 + start + tabs,
+                        max_line_num_len + 3 + end + tabs,
                         Style::Addition,
                         true,
                     );
--- a/compiler/rustc_errors/src/emitter.rs
+++ b/compiler/rustc_errors/src/emitter.rs
@@ -1623,7 +1623,7 @@ impl EmitterWriter {
             let line_start = sm.lookup_char_pos(parts[0].span.lo()).line;
             draw_col_separator_no_space(&mut buffer, 1, max_line_num_len + 1);
             let mut lines = complete.lines();
-            for (line_pos, (line, parts)) in
+            for (line_pos, (line, highlight_parts)) in
                 lines.by_ref().zip(highlights).take(MAX_SUGGESTION_HIGHLIGHT_LINES).enumerate()
             {
                 // Print the span column to avoid confusion
@@ -1658,7 +1658,7 @@ impl EmitterWriter {
                     );
                     buffer.puts(row_num, max_line_num_len + 1, "+ ", Style::Addition);
                 } else if is_multiline {
-                    match &parts[..] {
+                    match &highlight_parts[..] {
                         [SubstitutionHighlight { start: 0, end }] if *end == line.len() => {
                             buffer.puts(row_num, max_line_num_len + 1, "+ ", Style::Addition);
                         }
@@ -1676,16 +1676,33 @@ impl EmitterWriter {
                 // print the suggestion
                 buffer.append(row_num, &replace_tabs(line), Style::NoStyle);
 
-                if is_multiline {
-                    for SubstitutionHighlight { start, end } in parts {
-                        buffer.set_style_range(
-                            row_num,
-                            max_line_num_len + 3 + start,
-                            max_line_num_len + 3 + end,
-                            Style::Addition,
-                            true,
-                        );
-                    }
+                // Colorize addition/replacements with green.
+                for &SubstitutionHighlight { start, end } in highlight_parts {
+                    // Account for tabs when highlighting (#87972).
+                    let start: usize = line
+                        .chars()
+                        .take(start)
+                        .map(|ch| match ch {
+                            '\t' => 4,
+                            _ => 1,
+                        })
+                        .sum();
+
+                    let end: usize = line
+                        .chars()
+                        .take(end)
+                        .map(|ch| match ch {
+                            '\t' => 4,
+                            _ => 1,
+                        })
+                        .sum();
+                    buffer.set_style_range(
+                        row_num,
+                        max_line_num_len + 3 + start,
+                        max_line_num_len + 3 + end,
+                        Style::Addition,
+                        true,
+                    );
                 }
                 row_num += 1;
             }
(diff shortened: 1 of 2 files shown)


compiler/rustc_errors/src/emitter.rs
```diff
@@ -1678,6 +1678,23 @@ impl EmitterWriter {
 
                 if is_multiline {
                     for SubstitutionHighlight { start, end } in parts {
+                        // Account for tabs when highlighting (#87972).
+                        let start: usize = line
+                            .chars()
+                            .take(*start)
+                            .map(|ch| match ch {
+                                '\t' => 4,
+                                _ => unicode_width::UnicodeWidthChar::width(ch).unwrap_or(1),
```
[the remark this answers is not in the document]
estebank: Can you verify the current version of the PR against your test cases? Sadly, we don't currently have regression tests for the coloring of the output in our tests :-/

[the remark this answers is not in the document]
m-ou-se: Yup, the highlighting looks good now!

Screenshot 2021-08-12 at 21 38 57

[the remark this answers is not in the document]
estebank: In particular, I want to confirm that [the output of the terminal without emoji support you showed here](https://github.com/rust-lang/rust/issues/87972#issuecomment-897767010) is reasonable (as I fully expect it to be, because it relies on what we were doing for the multiline case, which looks ok in that screenshot).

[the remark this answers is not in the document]
estebank: Sadly, now I'm seeing _other_ issues :(

cjgillot:
Can you take care of this PR @m-ou-se?

compiler/rustc_errors/src/lib.rs
```diff
@@ -283,20 +283,25 @@ impl CodeSuggestion {
                 let mut buf = String::new();
 
                 let mut line_highlight = vec![];
+                // We need to keep track of the difference between the existing code and the added
+                // or deleted code in order to point at the correct column *after* substitution.
+                let mut acc = 0;
```
m-ou-se, line 288: Do you have an example of a problem that's fixed by this?

estebank: It's for cases where there are multiple replacements in the same line, like in the following:

With acc:
Screen Shot 2021-08-17 at 8 09 20 AM

Without acc:
Screen Shot 2021-08-17 at 8 09 30 AM

It accounts for the shift in position of the later spans caused by the earlier replacements.

But it's _still_ buggy on the current commit in the presence of unicode:

With acc:
Screen Shot 2021-08-17 at 8 09 07 AM

Without acc:
Screen Shot 2021-08-17 at 8 08 57 AM

I'm cleaning up and pushing a new change that as far as I can see handles anything I throw at it correctly (the `%` are everything that is supposed to be highlighted and there only for debugging):

estebank: Also, @m-ou-se, it might be useful to check the PR on your sampling of terminals, but I expect things to be "fine" :)

estebank: It's something I'd like to add at some point but not as part of this PR. It should be easy and self-contained, as a `-Z` flag, but I won't be having the time to do it in the short term.

estebank:
@m-ou-se do let me know if there's anything left here. I want to avoid this not reaching master before the next beta cutoff.

compiler/rustc_errors/src/lib.rs
```diff
@@ -316,18 +321,22 @@ impl CodeSuggestion {
...
-                        end: cur_lo.col.0
-                            + part.snippet.split('\n').next().unwrap_or(&part.snippet).len(),
+                        start: (cur_lo.col.0 as isize + acc) as usize,
+                        end: (cur_lo.col.0 as isize + acc + len as isize) as usize,
                     });
+                    buf.push_str(&part.snippet);
+                    prev_hi = sm.lookup_char_pos(part.span.hi());
+                    if prev_hi.line == cur_lo.line {
+                        acc += len as isize - (prev_hi.col.0 - cur_lo.col.0) as isize;
+                    }
```
m-ou-se, line 333: (I feel like i wrote this comment before, but I don't see it anywhere. Maybe I forgot to press the button or something. Sorry if it's a duplicate.)

It seems `prev_hi` temporarily represents `cur_hi` here, making this a bit confusing. The `prev_hi.col.0 - cur_lo.col.0` looks like it is calculating the space between spans (in the wrong order), but it's actually calculating the width of the current span. Using `cur_hi.col.0 - cur_lo.col.0` (and only later doing `prev_hi = cur_hi;`) would make that a lot clearer.

compiler/rustc_errors/src/emitter.rs
```diff
@@ -1676,16 +1676,24 @@ impl EmitterWriter {
...
+                        .take(start)
+                        .map(|ch| match ch {
+                            '\t' => 3,
+                            _ => 0,
+                        })
+                        .sum();
+                    buffer.set_style_range(
+                        row_num,
+                        max_line_num_len + 3 + start + tabs,
+                        max_line_num_len + 3 + end + tabs,
```
m-ou-se, line 1693: This assumes the addition/replacement doesn't contain tabs itself. It rarely happens, but it can occur:

image

Here, the entire closure up to and including the final `}` is supposed to be highlighted, but i put some tabs in the string literal which broke it.

m-ou-se (approved):
I found one more (unlikely) edge case (which is not very important, but not hard to fix), and a have a small comment on the code.

r=me either way

compiler/rustc_errors/src/emitter.rs
estebank: Ah, yes. I assumed that we didn't have tabs in the suggestion because I didn't account for cases where snippets from the user include them :-/

m-ou-se: I suppose all cases where user snippets are included should be replaced by multispan suggestions. (This specific one is already fixed by https://github.com/rust-lang/rust/pull/87958, but I think there's quite a few places left where this happens.)

estebank: Could you give me the repro case? I think I fixed this, but the suggestions seem to be doing the right thing for me.

m-ou-se: With #87958 merged, the test case in the screenshot above doesn't trigger the problem anymore. We'll have to find another case where a bunch of user code gets put verbatim in the suggestion. Grepping for `span_to_source` now. ^^

m-ou-se: Found one:

Screenshot 2021-08-23 at 15 59 26

```rust
mod a {
    pub const A: String = String::new();
}

fn main() {
    a
    ::	A;
    //^ TAB here
}
```

estebank: The current code handles this correctly:

camsteffen:
I'm getting this when I run `./x.py test src/tools/clippy` (line 345 in this PR diff)

```text
thread 'rustc' panicked at 'attempt to subtract with overflow', compiler/rustc_errors/src/lib.rs:347:47
```

estebank:
@camsteffen it sounds like there might be a malformed `Span` involved, where the end is earlier than the end _or_ `len` is smaller than the `Span`... Although now that I see it the `if` might be incorrect (we compare `prev_hi` and `cur_lo` but then calculate using `cur_hi` and `cur_lo`. What are the steps to reproduce this, just running that command?

camsteffen:
Yes I am seeing that error on multiple Clippy UI tests just running that command on 4878034c000a0404a4c83e77eddf (and probably current master).

estebank:
I wasn't able to reproduce this :-/

camsteffen:
Well shoot. Must've been an artifact of my local setup. I'm not seeing it either and it hasn't shown up in CI.

PR #87977, opened 2021-08-12
unwind: make `build.rs` do clean rebuilds

gburgessiv:
If the build by this was interrupted or otherwise left incomplete, there
could be object files hanging around, which triggers an assertion later.
Since we unconditionally rebuild everything here anyway, start with a
clean slate each build.

r? @Mark-Simulacrum

### DIFF
--- a/library/unwind/build.rs
+++ b/library/unwind/build.rs
@@ -65,8 +65,8 @@ fn main() {
 }
 
 mod llvm_libunwind {
-    use std::env;
-    use std::path::Path;
+    use std::path::{Path, PathBuf};
+    use std::{env, fs, io};
 
     /// Compile the libunwind C/C++ source code.
     pub fn compile() {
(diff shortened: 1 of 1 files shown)


library/unwind/build.rs
```diff
@@ -76,6 +88,16 @@ mod llvm_libunwind {
         let mut cpp_cfg = cc::Build::new();
         let root = Path::new("../../src/llvm-project/libunwind");
 
+        // We depend on starting with a fresh build directory each time.
```
Mark-Simulacrum, line 91: Hm, I'm wondering if we can get away with just deleting the directory as a whole here.

gburgessiv: Works on my machine; PTAL :)

gburgessiv (commented):
Thanks for the feedback!

gburgessiv:
TIL my `cargo fmt` config doesn't mesh with this repo's. Ran `./x.py fmt` to try to fix the things -- `./x.py test` no longer complains about formatting issues :)

gburgessiv:
Looks like the changes that caused the merge conflict also added a check to tell objects generated from C++ sources apart from objects generated from C sources. Hence, this should be fixed already, so I'll close it out. Thanks again for the review! :)

PR #87983, opened 2021-08-12
Use more accurate spans when proposing adding lifetime to item

estebank:


compiler/rustc_resolve/src/late/diagnostics.rs
```diff
@@ -2076,16 +2076,40 @@ impl<'tcx> LifetimeContext<'_, 'tcx> {
                 for param in params {
                     if let Ok(snippet) = self.tcx.sess.source_map().span_to_snippet(param.span) {
                         if snippet.starts_with('&') && !snippet.starts_with("&'") {
-                            introduce_suggestion
-                                .push((param.span, format!("&'a {}", &snippet[1..])));
-                        } else if let Some(stripped) = snippet.strip_prefix("&'_ ") {
```
oli-obk, line 2081: there's a lot of repetition in this diff. maybe we should create a `(String, Span)` newtype with functions like `strip_prefix` or `slice` that operate on both at the same time?

estebank: Done

estebank:
@oli-obk just pushed a quick fix for a double space issue in the newtype code.

@bors r=oli-obk

PR #87989, opened 2021-08-12
Make `lexical_region_resolve` public for Clippy

camsteffen:
I have a locally working fix for https://github.com/rust-lang/rust-clippy/issues/4002 that involves `lexical_region_resolve::resolve`. I'm doing my best to understand region checking and how it should integrate with Clippy, so please tell me if I'm totally on the wrong path. Basically I need to construct and evaluate a `TypeOutlives` obligation.

### DIFF
--- a/compiler/rustc_infer/src/infer/mod.rs
+++ b/compiler/rustc_infer/src/infer/mod.rs
@@ -58,7 +58,7 @@ mod fudge;
 mod glb;
 mod higher_ranked;
 pub mod lattice;
-mod lexical_region_resolve;
+pub mod lexical_region_resolve;
 mod lub;
 pub mod nll_relate;
 pub mod opaque_types;


camsteffen:
I wonder if the Cilppy suggestion should compile though? Isn't any `A::b`  `'static` as long as it is a function pointer?

petrochenkov:
If something is public only for clippy, then it probably should have a comment saying that it's public only for clippy.

Regarding other questions, r? @jackh726 maybe.

camsteffen:
After looking at this again, I think this is wrong. But there may be an opportunity to improve rustc coercion. I found out that `take(G::f as fn(Arg))` works.

jackh726:
I still need to take some time to look at this. I'll try to do that soon.

jackh726:
@camsteffen do you still need this to be public? If so, do you have a branch I can look at?

jackh726:
If you still need this, I think r=me with an appropriate comment

camsteffen:
I decided not to go this route. I just opened https://github.com/rust-lang/rust-clippy/pull/7661 with a workaround for the Clippy bug - I just ignore associated functions on type parameters. I'm fairly certain this is an issue with rustc coercion so I don't think it makes sense to add this complexity to Clippy to work around a rustc bug.

camsteffen:
See https://github.com/rust-lang/rust-clippy/blob/ae2a95fadc098b326f2114fc5ee6f55ee0db81a8/clippy_lints/src/eta_reduction.rs#L113 and https://github.com/rust-lang/rust-clippy/blob/ae2a95fadc098b326f2114fc5ee6f55ee0db81a8/tests/ui/eta.rs#L248

PR #87993, opened 2021-08-12
Stabilize try_reserve

kornelski:
Stabilization PR for the [`try_reserve` feature](https://github.com/rust-lang/rust/issues/48043#issuecomment-898040475).


TimDiekmann:
cc @rust-lang/wg-allocators

joshtriplett:
FCP for stabilization:
@rfcbot merge

dtolnay:
@rfcbot concern need `From<LayoutError> for TryReserveError`? https://github.com/rust-lang/rust/issues/48043#issuecomment-903179033

joshtriplett:
@dtolnay Thanks for catching that. I don't think we need that `From` instance.

compiler/rustc_feature/src/accepted.rs
```diff
@@ -294,6 +294,8 @@ declare_features! (
     (accepted, const_fn_transmute, "1.56.0", Some(53605), None),
     /// Allows accessing fields of unions inside `const` functions.
     (accepted, const_fn_union, "1.56.0", Some(51909), None),
+    /// Allows handling memory allocation failures in std collections
+    (accepted, try_reserve, "1.56.0", Some(48043), None),
```
m-ou-se, line 298: This file is only for language features.
```suggestion
```

paolobarbolini:
I guess since 1.56 already branched the `since` version will have to be updated to 1.57.0

compiler/rustc_data_structures/src/sso/set.rs
```diff
@@ -13,7 +13,7 @@ use super::map::SsoHashMap;
 //
 // Missing HashSet API:
 //   all hasher-related
-//   try_reserve (unstable)
+//   try_reserve
```
Wodann, line 16:
```suggestion
```

Wodann: No longer missing this API

compiler/rustc_data_structures/src/sso/map.rs
```diff
@@ -31,7 +31,7 @@ const SSO_ARRAY_SIZE: usize = 8;
 //
 // Missing HashMap API:
 //   all hasher-related
-//   try_reserve (unstable)
+//   try_reserve
```
Wodann, line 34:
```suggestion
```

Wodann: No longer missing this API

Wodann (changes requested):

joshtriplett:
r=me with the conflict resolved.

kornelski:
I've also bumped it to 1.57, as I assume it's too late for 1.56 now.

LeonineKing1199:
Why wasn't the actual error kind stabilized? There's no point to this feature without the ability to introspect.

kornelski:
@LeonineKing1199 You can see in the original RFC discussion that it wasn't certain if the error should be stabilized exactly as it is implemented currently, e.g. with separate cases that just mirror an inconsistency in std (where some allocation failures panic, and some abort).

I'm interested to hear what is your use-case and what data you'd like to get out of the error type, to help guide how the error type should be designed.

LeonineKing1199:
Ha ha, looks like my case was already addressed:
> One I can think of is someone creating a custom data structure, and wanting to mimic std API. But that's not a very strong case.

It is for implementing the stdlib APIs without polluting it with polyfills that are incompatible with what's in the stdlib.

kornelski:
Great. So you're not looking to get any data out of it yourself, you just want to return a `TryReserveError` instance? If there was `TryReserveError::default()` or just a very basic `TryReserveError::new(allocation_size)`, would that work for you?

Stargateur:
> Why wasn't the actual error kind stabilized? There's no point to this feature without the ability to introspect.

In my opinion, the error is more for the end user, or very special case for something like kernel linux. For a server I will log the error and continue anyway so don't really care, even if I would consider a `CapacityOverflow` a code bug in a server or at least get a look at it.

@kornelski Do you think we could have const value `TryReserveError::CAPACITY_OVERFLOW` ? Unless if we want to be able to precise the maximum value at runtine and possibly the value used I don't expect this to change a lot.

LeonineKing1199:
> Great. So you're not looking to get any data out of it yourself, you just want to return a `TryReserveError` instance? If there was `TryReserveError::default()` or just a very basic `TryReserveError::new(allocation_size)`, would that work for you?

Actually, this would work perfectly!

PR #75438, opened 2020-08-12
Use adaptive SVG favicon for rustdoc like other rust sites

Cldfire:
Use the theme-adaptive SVG favicon that was recently introduced [for the Rust site](https://github.com/rust-lang/www.rust-lang.org/pull/1185) (and others).

(This PR is simply copied from the PR linked above, so see that for rationale.)

Closes #72165.

Before, Firefox on Linux:

image

After, Firefox on Linux (`prefers-color-scheme` set to `dark` by setting `ui.systemUsesDarkTheme` to a number value of `1` in `about:config`):

image

PR #87997, opened 2021-08-13
Updates favicon order of precedence as it matters to Chrome

sl4m:
Hi, this updates #75438 to fix an order of precedence issue for Chrome. Unfortunately, the last favicon defined wins when it comes to Chrome, hence the primary icon being placed last. I [brought it up](https://bugs.chromium.org/p/chromium/issues/detail?id=1104663) with the Chromium team last year, but so far it's a non-issue.

I've created an example website that mimics the behaviour in Chrome. https://sl4m.github.io/chrome-favicon/

This is what I'm seeing at the moment when viewing https://doc.rust-lang.org/core/index.html in Chrome. It's falling back to the PNG.

Screenshot 2021-08-12 at 21 11 58

### DIFF
--- a/src/librustdoc/html/templates/page.html
+++ b/src/librustdoc/html/templates/page.html
@@ -31,12 +31,12 @@
     {%- if layout.favicon -%}
         <link rel="shortcut icon" href="{{layout.favicon}}"> {#- -#}
     {%- else -%}
-        <link rel="icon" type="image/svg+xml" {# -#}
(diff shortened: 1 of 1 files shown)


GuillaumeGomez:
What is the behaviour in the other browsers (Im thinking about firefox mainly)?

sl4m:
SVG takes precedence in Firefox. The order in this PR is the same order in the other Rust websites

Screenshot 2021-08-13 at 06 36 36

GuillaumeGomez:
Fine by me then, thanks!

@bors: r+ rollup

PR #87998, opened 2021-08-13
Avoid spurious "previous iteration of loop" errors

### LINKED ISSUE #72649: Wrong error message about uninitialized variable in loop
Forgetting to initialize a non-Copy variable in a loop and then referring to it gives a 'borrowed after move' error, instead of a 'possibly-uninitialized variable' error:

```rust
fn bla(_: &mut String) {}

fn main() {
    for _ in 0..10 {
        let mut x: String;
        bla(&mut x);
    }
}
```

```
error[E0382]: borrow of moved value: `x`
 --> src/main.rs:6:13
  |
5 |         let mut x: String;
  |             ----- move occurs because `x` has type `std::string::String`, which does not implement the `Copy` trait
6 |         bla(&mut x);
  |             ^^^^^^ value borrowed here after move
7 |     }
  |     - value moved here, in previous iteration of loop
```

Using a `Copy` type like `i32` does result in the right error:
```
error[E0381]: borrow of possibly-uninitialized variable: `x`
 --> src/main.rs:6:13
  |
6 |         bla(&mut x);
  |             ^^^^^^ use of possibly-uninitialized `x`
```

Tested with 1.43.0 (stable) and 1.45.0-nightly.

nneonneo:
Only follow backwards edges during `get_moved_indexes` if the move path is definitely initialized at loop entry. Otherwise, the error occurred prior to the loop, so we ignore the backwards edges to avoid generating misleading "value moved here, in previous iteration of loop" errors.

This patch also slightly improves the analysis of inits, including `NonPanicPathOnly` initializations (which are ignored by `drop_flag_effects::for_location_inits`). This is required for the definite initialization analysis, but may also help find certain skipped reinits in rare cases.

Patch passes all non-ignored src/test/ui testcases.

Fixes #72649.

### DIFF
--- a/compiler/rustc_borrowck/src/diagnostics/conflict_errors.rs
+++ b/compiler/rustc_borrowck/src/diagnostics/conflict_errors.rs
@@ -10,8 +10,7 @@ use rustc_middle::mir::{
     ProjectionElem, Rvalue, Statement, StatementKind, Terminator, TerminatorKind, VarBindingForm,
 };
 use rustc_middle::ty::{self, suggest_constraining_type_param, Ty};
-use rustc_mir_dataflow::drop_flag_effects;
-use rustc_mir_dataflow::move_paths::{MoveOutIndex, MovePathIndex};
+use rustc_mir_dataflow::move_paths::{InitKind, MoveOutIndex, MovePathIndex};
 use rustc_span::source_map::DesugaringKind;
 use rustc_span::symbol::sym;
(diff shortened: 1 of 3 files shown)


nneonneo:
r? @estebank

Also, please note that this is roughly the first time I've written any serious Rust code, so I would definitely appreciate any review feedback :)

nneonneo:
@estebank (or @wesleywiser, etc.): Could I get some feedback on this PR?

ehuss:
@bors retry
Issue with Windows LLDB

PR #88004, opened 2021-08-13
ancilliary data tests only for relevant oses

devnexen:
As it stops tests otherwise, triggering only for linux/android.

### DIFF
--- a/library/std/src/os/unix/net/tests.rs
+++ b/library/std/src/os/unix/net/tests.rs
@@ -1,15 +1,7 @@
 use super::*;
 use crate::io::prelude::*;
 use crate::io::{self, ErrorKind, IoSlice, IoSliceMut};
-#[cfg(any(
-    target_os = "android",
-    target_os = "dragonfly",
-    target_os = "emscripten",
-    target_os = "freebsd",
-    target_os = "linux",
-    target_os = "netbsd",
-    target_os = "openbsd",
-))]
+#[cfg(any(target_os = "android", target_os = "linux",))]
 use crate::iter::FromIterator;
 #[cfg(any(
     target_os = "android",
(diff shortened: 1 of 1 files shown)


the8472:
> As it stops tests otherwise, triggering only for linux/android.

I have difficulty understanding the PR description. Do you mean the tests aren't working on other platforms? I think they *should* work, e.g. freebsd does have a notion of SCM_RIGHTS. https://www.freebsd.org/cgi/man.cgi?query=unix&sektion=4

devnexen:
it does not work as it is not implemented here (yet).

the8472:
Which platform is "here"?

devnexen:
I was referring to rust, not implemented for BSD in general, for instance.

the8472:
That doesn't seem right? The ancilliary module itself is exported on various BSDs.

https://github.com/rust-lang/rust/blob/5de331ba40ac722dd5513c79194c6c411766a23f/library/std/src/os/unix/net/mod.rs#L15-L25

And inside the module `ScmRights` is exported unconditionally.

https://github.com/rust-lang/rust/blob/5de331ba40ac722dd5513c79194c6c411766a23f/library/std/src/os/unix/net/ancillary.rs#L236-L237

Since that is public *should* work on all the listed platforms, at least on nightly since it's still unstable. If it doesn't then that's a bug that needs to be fixed one way or another, rather than disabling tests.

devnexen:
it is mainly for this reason 
https://github.com/rust-lang/rust/blob/5de331ba40ac722dd5513c79194c6c411766a23f/library/std/src/os/unix/net/ancillary.rs#L181

the8472:
But that's for `SocketCred` while the test is about `ScmRights`

devnexen:
true sorry. In reality, the issue seems to be the if let pattern, making the unreachable code path useless.

PR #88012, opened 2021-08-13
Change WASI's `RawFd` from `u32` to `c_int` (`i32`).

sunfishcode:
WASI previously used `u32` as its `RawFd` type, since its "file descriptors"
are unsigned table indices, and there's no fundamental reason why WASI can't
have more than 2^31 handles.

However, this creates myriad little incompability problems with code
that also supports Unix platforms, where `RawFd` is `c_int`. While WASI
isn't a Unix, it often shares code with Unix, and this difference made
such shared code inconvenient. #87329 is the most recent example of such
code.

So, switch WASI to use `c_int`, which is `i32`. This will mean that code
intending to support WASI should ideally avoid assuming that negative file
descriptors are invalid, even though POSIX itself says that file descriptors
are never negative.

This is a breaking change, but `RawFd` is considerd an experimental
feature in [the documentation].

[the documentation]: https://doc.rust-lang.org/stable/std/os/wasi/io/type.RawFd.html

r? @alexcrichton

### DIFF
--- a/library/std/src/os/wasi/io.rs
+++ b/library/std/src/os/wasi/io.rs
@@ -6,11 +6,18 @@
 use crate::fs;
 use crate::io;
 use crate::net;
+use crate::os::raw;
 use crate::sys;
 use crate::sys_common::{AsInner, FromInner, IntoInner};
 
 /// Raw file descriptors.
-pub type RawFd = u32;
+///
+/// This has type `c_int` to ease compatibility with code that also compiles on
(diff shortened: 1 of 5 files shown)


alexcrichton:
@bors: r+

Seems reasonable to me! And yeah the intention is that `std::os::wasi` is unstable still so I'm hopeful that this shouldn't break too many users in the wild.

PR #88019, opened 2021-08-14
Treat macros as HIR items

### LINKED ISSUE #59306: `allow(missing_docs)` does not work for macro definitions
It appears `deny(missing_docs)` will take macros into account since 1.33.0. So I tried to add `#[allow(missing_docs)]` for some of the macros I don't want to write docs, and it doesn't seem to be working. Minimal reproducible example:

```rust
//! Hello
#![deny(missing_docs)]

#[allow(missing_docs)]
#[macro_export]
macro_rules! hello {
    () => ()
}
```

It will fail with:

```
error: missing documentation for macro
 --> src/lib.rs:6:1
  |
6 | macro_rules! hello {
  | ^^^^^^^^^^^^^^^^^^
  |
note: lint level defined here
 --> src/lib.rs:2:9
  |
2 | #![deny(missing_docs)]
  |         ^^^^^^^^^^^^
```

Rust playground: https://play.rust-lang.org/?version=nightly&mode=debug&edition=2018&gist=cf5ab453ab982bea1b435676cf46f119

Tested on nightly.

### LINKED ISSUE #73754: [rustdoc] Private macros are not documented with `--document-private-items`
Rustdoc has `--document-private-items` and `--document-hidden-items`. It would be nice to have an option to document non-exported macros, which is currently not possible as far as I can tell.

### LINKED ISSUE #87257: Rustdoc: `pub use`ing and/or globbing a submodule containing a `pub macro` (Macros 2.0) does not document the macro
Rustdoc will properly generate the documentation for a Macros-2.0 macro when that macro is defined in a directly-exported submodule, e.g. `pub mod foo { pub macro bar()`. But if the module is not directly exported (`pub mod foo {`) but indirectly exported (`pub use foo;`) then the macro no longer exists in the documentation: the module documentation for `mod foo` will not mention it, and no page for the macro will be generated at all. The same problem occurs if a macro is exported via `pub use foo::*`.

Note that these are not problems with name resolution itself: Rust code will resolve the exported macro properly, the docs merely don't show it.

Example:

```rust
#![feature(decl_macro)]

mod outer {
    /// I am module documentation
    pub mod inner {
        /// I am macro documentation
        pub macro some_macro() {}

        /// I am struct documentation
        pub struct SomeStruct;
    }
}

// GOOD: this shows up properly in the top-level crate docs
pub use outer::inner::SomeStruct;

// GOOD: this shows up properly in the top-level crate docs
pub use outer::inner::some_macro;

// BAD: the `inner` module-level documentation shows the struct, but not the macro.
// If `outer` did not exist and `inner` were instead directly exported,
// then both items show up as expected.
pub use outer::inner;
```

(In the above example even though `some_macro` is not listed in the docs for `inner`, the docs page for `some_macro` will still be generated because of the explicit `pub use outer::inner::some_macro`; if this line is removed then that page will also vanish.)

Presumably this is a descendant of https://github.com/rust-lang/rust/issues/74355 , where Macros-2.0 were not being documented at the correct paths. The fix for that issue landed in https://github.com/rust-lang/rust/pull/77862 , although the fix itself appears to be regarded as a deplorable hack, and presumably the fix for this issue should not attempt to further hack the hack and instead make Macros 2.0 Just Work with Rustdoc.

### LINKED ISSUE #87406: Macro definitions should be regular HIR items
Macro definitions are currently handled separately from other item kinds in HIR, either as `MacroDef` nodes inside `hir::Crate::exported_macros` or as using the `hir::Crate::non_exported_macro_attrs`.

They should be transformed into additional variants in the `hir::ItemKind` enum, and iterated over like any other item.

Instructions:
- introduce two additional variants `hir::ItemKind::ExportedMacro { ast: ast::MacroDef }` and `hir::ItemKind::NonExportedMacro`;
- move the macro case in `rustc_ast_lowering::item::LoweringContext::lower_item` into `lower_item_kind`;
- remove `hir::OwnerNode::{MacroDef, NonExportedMacro}` variants.

I recomment waiting on https://github.com/rust-lang/rust/pull/83723 and https://github.com/rust-lang/rust/pull/87234 to be merged.

Original idea by @petrochenkov in https://github.com/rust-lang/rust/pull/83723#discussion_r674077466
Should help with https://github.com/rust-lang/rust/issues/73754

inquisitivecrystal:
Macros have historically been treated differently from other items at the HIR level. This PR makes them just like normal items. There are a few special cases left over, which I've attempted to lay out below. By normalizing the treatment of macro items, this PR simplifies a fair bit of code and fixes some bugs at the same time. For more information, see #87406.

r? @cjgillot

## Backwards incompatibility

This is backwards incompatible in one small way. Due to a mistake, it was previously possible to apply stability attributes to an exported macro, even without enabling the `staged_api` feature. This never should have worked. After this PR, it will error, as it should. We could make it a warning instead, but that would require a special case for a feature that shouldn't ever have worked and is likely used by no or very few crates, so I'm not thrilled about the idea.

### Commit seperation

I'd recommend reviewing this PR commit by commit. The commit chunking wasn't perfect, but it's better than looking at the combined diff, which is quite overwhelming. The compiler and standard library build after each commit, although tests do not necessarily pass and tools do not necessarily build till the end of the series.

### Special cases
There are a few special cases that remain after this change. Here are the notable ones I remember:

1. Visibility works a bit differently for `macro_rules!` macros than other items, since they aren't generally marked with `pub` but instead with `#[macro_export]`.
2. Since `#[macro_export]` macros always have paths at the top level of the crate, some additional handling needs to be done on the reexport to top level.
### Performance impact

I don't know for sure, but theses changes may slightly hurt performance. They create more work for the compiler in a few places. For instance, some operations that were previously run only for exported macros are now run for all macros. A perf run is probably advisable. For all I know we'll see performance improvements instead. :)

## Issues resolved

This resolves #87406 (the tracking issue for this change). It also fixes several bugs:

Fixes #59306.
Fixes #73754.
Fixes #87257.

### DIFF
--- a/compiler/rustc_ast_lowering/src/item.rs
+++ b/compiler/rustc_ast_lowering/src/item.rs
@@ -170,7 +170,6 @@ impl<'hir> LoweringContext<'_, 'hir> {
                 self.lower_item_id_use_tree(use_tree, i.id, &mut vec);
                 vec
             }
-            ItemKind::MacroDef(..) => SmallVec::new(),
             ItemKind::Fn(..) | ItemKind::Impl(box ImplKind { of_trait: None, .. }) => {
                 smallvec![i.id]
             }
@@ -212,28 +211,6 @@ impl<'hir> LoweringContext<'_, 'hir> {
     pub fn lower_item(&mut self, i: &Item) -> Option<hir::Item<'hir>> {
         let mut ident = i.ident;
         let mut vis = self.lower_visibility(&i.vis, None);
-
-        if let ItemKind::MacroDef(MacroDef { ref body, macro_rules }) = i.kind {
-            if !macro_rules || self.sess.contains_name(&i.attrs, sym::macro_export) {
-                let hir_id = self.lower_node_id(i.id);
-                self.lower_attrs(hir_id, &i.attrs);
-                let body = P(self.lower_mac_args(body));
-                self.insert_macro_def(hir::MacroDef {
-                    ident,
-                    vis,
-                    def_id: hir_id.expect_owner(),
-                    span: i.span,
-                    ast: MacroDef { body, macro_rules },
-                });
-            } else {
-                for a in i.attrs.iter() {
-                    let a = self.lower_attr(a);
-                    self.non_exported_macro_attrs.push(a);
-                }
-            }
-            return None;
-        }
-
         let hir_id = self.lower_node_id(i.id);
         let attrs = self.lower_attrs(hir_id, &i.attrs);
         let kind = self.lower_item_kind(i.span, i.id, hir_id, &mut ident, attrs, &mut vis, &i.kind);
@@ -465,7 +442,12 @@ impl<'hir> LoweringContext<'_, 'hir> {
                 self.lower_generics(generics, ImplTraitContext::disallowed()),
                 self.lower_param_bounds(bounds, ImplTraitContext::disallowed()),
             ),
-            ItemKind::MacroDef(..) | ItemKind::MacCall(..) => {
+            ItemKind::MacroDef(MacroDef { ref body, macro_rules }) => {
+                let body = P(self.lower_mac_args(body));
+
+                hir::ItemKind::Macro(ast::MacroDef { body, macro_rules })
+            }
+            ItemKind::MacCall(..) => {
                 panic!("`TyMac` should have been expanded by now")
             }
         }
--- a/compiler/rustc_ast_lowering/src/lib.rs
+++ b/compiler/rustc_ast_lowering/src/lib.rs
@@ -103,7 +103,6 @@ struct LoweringContext<'a, 'hir: 'a> {
     /// The items being lowered are collected here.
     owners: IndexVec<LocalDefId, Option<hir::OwnerNode<'hir>>>,
     bodies: BTreeMap<hir::BodyId, hir::Body<'hir>>,
-    non_exported_macro_attrs: Vec<ast::Attribute>,
 
     trait_impls: BTreeMap<DefId, Vec<LocalDefId>>,
 
@@ -330,7 +329,6 @@ pub fn lower_crate<'a, 'hir>(
         trait_impls: BTreeMap::new(),
         modules: BTreeMap::new(),
         attrs: BTreeMap::default(),
-        non_exported_macro_attrs: Vec::new(),
         catch_scopes: Vec::new(),
         loop_scopes: Vec::new(),
         is_in_loop_condition: false,
@@ -551,7 +549,6 @@ impl<'a, 'hir> LoweringContext<'a, 'hir> {
         }
 
         let krate = hir::Crate {
-            non_exported_macro_attrs: self.arena.alloc_from_iter(self.non_exported_macro_attrs),
             owners: self.owners,
             bodies: self.bodies,
             body_ids,
@@ -598,13 +595,6 @@ impl<'a, 'hir> LoweringContext<'a, 'hir> {
         self.owners[id.def_id] = Some(hir::OwnerNode::TraitItem(item));
         self.modules.entry(self.current_module).or_default().trait_items.insert(id);
         id
-    }
-
-    fn insert_macro_def(&mut self, item: hir::MacroDef<'hir>) {
-        let def_id = item.def_id;
-        let item = self.arena.alloc(item);
-        self.owners.ensure_contains_elem(def_id, || None);
-        self.owners[def_id] = Some(hir::OwnerNode::MacroDef(item));
     }
 
     fn allocate_hir_id_counter(&mut self, owner: NodeId) -> hir::HirId {
--- a/compiler/rustc_hir/src/stable_hash_impls.rs
+++ b/compiler/rustc_hir/src/stable_hash_impls.rs
@@ -1,8 +1,8 @@
 use rustc_data_structures::stable_hasher::{HashStable, StableHasher, ToStableHashKey};
 
 use crate::hir::{
-    BodyId, Expr, ForeignItem, ForeignItemId, ImplItem, ImplItemId, Item, ItemId, MacroDef, Mod,
-    TraitItem, TraitItemId, Ty, VisibilityKind,
+    BodyId, Expr, ForeignItem, ForeignItemId, ImplItem, ImplItemId, Item, ItemId, Mod, TraitItem,
+    TraitItemId, Ty, VisibilityKind,
 };
 use crate::hir_id::{HirId, ItemLocalId};
 use rustc_span::def_id::DefPathHash;
@@ -185,19 +185,6 @@ impl<HirCtx: crate::HashStableContext> HashStable<HirCtx> for Item<'_> {
         hcx.hash_hir_item_like(|hcx| {
             ident.name.hash_stable(hcx, hasher);
             kind.hash_stable(hcx, hasher);
-            vis.hash_stable(hcx, hasher);
-            span.hash_stable(hcx, hasher);
-        });
-    }
-}
-
-impl<HirCtx: crate::HashStableContext> HashStable<HirCtx> for MacroDef<'_> {
-    fn hash_stable(&self, hcx: &mut HirCtx, hasher: &mut StableHasher) {
-        let MacroDef { ident, def_id: _, ref ast, ref vis, span } = *self;
-
-        hcx.hash_hir_item_like(|hcx| {
-            ident.name.hash_stable(hcx, hasher);
-            ast.hash_stable(hcx, hasher);
             vis.hash_stable(hcx, hasher);
             span.hash_stable(hcx, hasher);
         });
--- a/compiler/rustc_hir_pretty/src/lib.rs
+++ b/compiler/rustc_hir_pretty/src/lib.rs
@@ -642,6 +642,11 @@ impl<'a> State<'a> {
                 self.end(); // need to close a box
                 self.ann.nested(self, Nested::Body(body));
             }
+            hir::ItemKind::Macro(ref macro_def) => {
+                self.print_mac_def(macro_def, &item.ident, &item.span, |state| {
+                    state.print_visibility(&item.vis)
+                });
+            }
             hir::ItemKind::Mod(ref _mod) => {
                 self.head(visibility_qualified(&item.vis, "mod"));
                 self.print_ident(item.ident);
--- a/compiler/rustc_hir_pretty/src/lib.rs
+++ b/compiler/rustc_hir_pretty/src/lib.rs
@@ -120,7 +120,6 @@ impl<'a> State<'a> {
             // printing.
             Node::Ctor(..) => panic!("cannot print isolated Ctor"),
             Node::Local(a) => self.print_local_decl(&a),
-            Node::MacroDef(_) => panic!("cannot print MacroDef"),
             Node::Crate(..) => panic!("cannot print Crate"),
         }
     }
--- a/compiler/rustc_lint/src/builtin.rs
+++ b/compiler/rustc_lint/src/builtin.rs
@@ -585,24 +585,6 @@ impl<'tcx> LateLintPass<'tcx> for MissingDoc {
 
     fn check_crate(&mut self, cx: &LateContext<'_>, krate: &hir::Crate<'_>) {
         self.check_missing_docs_attrs(cx, CRATE_DEF_ID, krate.module().inner, "the", "crate");
-
-        for macro_def in krate.exported_macros() {
-            // Non exported macros should be skipped, since `missing_docs` only
-            // applies to externally visible items.
-            if !cx.access_levels.is_exported(macro_def.def_id) {
-                continue;
-            }
-
-            let attrs = cx.tcx.hir().attrs(macro_def.hir_id());
-            let has_doc = attrs.iter().any(has_doc);
-            if !has_doc {
-                cx.struct_span_lint(
-                    MISSING_DOCS,
-                    cx.tcx.sess.source_map().guess_head_span(macro_def.span),
-                    |lint| lint.build("missing documentation for macro").emit(),
-                );
-            }
-        }
     }
 
     fn check_item(&mut self, cx: &LateContext<'_>, it: &hir::Item<'_>) {
@@ -636,6 +618,7 @@ impl<'tcx> LateLintPass<'tcx> for MissingDoc {
 
             hir::ItemKind::TyAlias(..)
             | hir::ItemKind::Fn(..)
+            | hir::ItemKind::Macro(..)
             | hir::ItemKind::Mod(..)
             | hir::ItemKind::Enum(..)
             | hir::ItemKind::Struct(..)
--- a/compiler/rustc_metadata/src/rmeta/decoder.rs
+++ b/compiler/rustc_metadata/src/rmeta/decoder.rs
@@ -1100,7 +1100,13 @@ impl<'a, 'tcx> CrateMetadataRef<'a> {
                     let vis = self.get_visibility(child_index);
                     let def_id = self.local_def_id(child_index);
                     let res = Res::Def(kind, def_id);
-                    callback(Export { res, ident, vis, span });
+
+                    // FIXME: Macros are currently encoded twice, once as items and once as
+                    // reexports. We ignore the items here and only use the reexports.
+                    if !matches!(kind, DefKind::Macro(..)) {
+                        callback(Export { res, ident, vis, span });
+                    }
+
                     // For non-re-export structs and variants add their constructors to children.
                     // Re-export lists automatically contain constructors when necessary.
                     match kind {
--- a/compiler/rustc_passes/src/stability.rs
+++ b/compiler/rustc_passes/src/stability.rs
@@ -538,19 +538,6 @@ impl<'a, 'tcx> Visitor<'tcx> for Annotator<'a, 'tcx> {
         );
     }
 
-    fn visit_macro_def(&mut self, md: &'tcx hir::MacroDef<'tcx>) {
-        self.annotate(
-            md.def_id,
-            md.span,
-            None,
-            AnnotationKind::Required,
-            InheritDeprecation::Yes,
-            InheritConstStability::No,
-            InheritStability::No,
-            |_| {},
-        );
-    }
-
     fn visit_generic_param(&mut self, p: &'tcx hir::GenericParam<'tcx>) {
         let kind = match &p.kind {
             // Allow stability attributes on default generic arguments.
@@ -662,11 +649,6 @@ impl<'tcx> Visitor<'tcx> for MissingStabilityAnnotations<'tcx> {
         self.check_missing_stability(i.def_id, i.span);
         intravisit::walk_foreign_item(self, i);
     }
-
-    fn visit_macro_def(&mut self, md: &'tcx hir::MacroDef<'tcx>) {
-        self.check_missing_stability(md.def_id, md.span);
-    }
-
     // Note that we don't need to `check_missing_stability` for default generic parameters,
     // as we assume that any default generic parameters without attributes are automatically
     // stable (assuming they have not inherited instability from their parent).
--- a/compiler/rustc_privacy/src/lib.rs
+++ b/compiler/rustc_privacy/src/lib.rs
@@ -6,6 +6,7 @@
 #![feature(associated_type_defaults)]
 #![recursion_limit = "256"]
 
+use rustc_ast::MacroDef;
 use rustc_attr as attr;
 use rustc_data_structures::fx::FxHashSet;
 use rustc_errors::struct_span_err;
@@ -26,7 +27,7 @@ use rustc_middle::ty::subst::{InternalSubsts, Subst};
 use rustc_middle::ty::{self, Const, GenericParamDefKind, TraitRef, Ty, TyCtxt, TypeFoldable};
 use rustc_session::lint;
 use rustc_span::hygiene::Transparency;
-use rustc_span::symbol::{kw, Ident};
+use rustc_span::symbol::{kw, sym, Ident};
 use rustc_span::Span;
 use rustc_trait_selection::traits::const_evaluatable::{self, AbstractConst};
 
@@ -462,6 +463,43 @@ impl EmbargoVisitor<'tcx> {
         }
     }
 
+    // We have to make sure that the items that macros might reference
+    // are reachable, since they might be exported transitively.
+    fn update_reachability_from_macro(&mut self, local_def_id: LocalDefId, md: &MacroDef) {
+        // Non-opaque macros cannot make other items more accessible than they already are.
+
+        let hir_id = self.tcx.hir().local_def_id_to_hir_id(local_def_id);
+        let attrs = self.tcx.hir().attrs(hir_id);
+        if attr::find_transparency(&attrs, md.macro_rules).0 != Transparency::Opaque {
+            return;
+        }
+
+        let item_def_id = local_def_id.to_def_id();
+        let macro_module_def_id =
+            ty::DefIdTree::parent(self.tcx, item_def_id).unwrap().expect_local();
+        if self.tcx.hir().opt_def_kind(macro_module_def_id) != Some(DefKind::Mod) {
+            // The macro's parent doesn't correspond to a `mod`, return early (#63164, #65252).
+            return;
+        }
+
+        if self.get(local_def_id).is_none() {
+            return;
+        }
+
+        // Since we are starting from an externally visible module,
+        // all the parents in the loop below are also guaranteed to be modules.
+        let mut module_def_id = macro_module_def_id;
+        loop {
+            let changed_reachability =
+                self.update_macro_reachable(module_def_id, macro_module_def_id);
+            if changed_reachability || module_def_id == CRATE_DEF_ID {
+                break;
+            }
+            module_def_id =
+                ty::DefIdTree::parent(self.tcx, module_def_id.to_def_id()).unwrap().expect_local();
+        }
+    }
+
     /// Updates the item as being reachable through a macro defined in the given
     /// module. Returns `true` if the level has changed.
     fn update_macro_reachable(
@@ -511,13 +549,23 @@ impl EmbargoVisitor<'tcx> {
         }
         match def_kind {
             // No type privacy, so can be directly marked as reachable.
-            DefKind::Const
-            | DefKind::Macro(_)
-            | DefKind::Static
-            | DefKind::TraitAlias
-            | DefKind::TyAlias => {
+            DefKind::Const | DefKind::Static | DefKind::TraitAlias | DefKind::TyAlias => {
                 if vis.is_accessible_from(module.to_def_id(), self.tcx) {
                     self.update(def_id, level);
+                }
+            }
+
+            // Hygine isn't really implemented for `macro_rules!` macros at the
+            // moment. Accordingly, marking them as reachable is unwise. `macro` macros
+            // have normal  hygine, so we can treat them like other items without type
+            // privacy and mark them reachable.
+            DefKind::Macro(_) => {
+                let hir_id = self.tcx.hir().local_def_id_to_hir_id(def_id);
+                let item = self.tcx.hir().expect_item(hir_id);
+                if let hir::ItemKind::Macro(MacroDef { macro_rules: false, .. }) = item.kind {
+                    if vis.is_accessible_from(module.to_def_id(), self.tcx) {
+                        self.update(def_id, level);
+                    }
                 }
             }
 
@@ -644,6 +692,12 @@ impl Visitor<'tcx> for EmbargoVisitor<'tcx> {
             hir::ItemKind::Impl { .. } => {
                 Option::<AccessLevel>::of_impl(item.def_id, self.tcx, &self.access_levels)
             }
+            // Only exported `macro_rules!` items are public, but they always are.
+            hir::ItemKind::Macro(MacroDef { macro_rules: true, .. }) => {
+                let def_id = item.def_id.to_def_id();
+                let is_macro_export = self.tcx.has_attr(def_id, sym::macro_export);
+                if is_macro_export { Some(AccessLevel::Public) } else { None }
+            }
             // Foreign modules inherit level from parents.
             hir::ItemKind::ForeignMod { .. } => self.prev_level,
             // Other `pub` items inherit levels from parents.
@@ -652,6 +706,7 @@ impl Visitor<'tcx> for EmbargoVisitor<'tcx> {
             | hir::ItemKind::ExternCrate(..)
             | hir::ItemKind::GlobalAsm(..)
             | hir::ItemKind::Fn(..)
+            | hir::ItemKind::Macro(..)
             | hir::ItemKind::Mod(..)
             | hir::ItemKind::Static(..)
             | hir::ItemKind::Struct(..)
@@ -708,6 +763,9 @@ impl Visitor<'tcx> for EmbargoVisitor<'tcx> {
                     }
                 }
             }
+            hir::ItemKind::Macro(ref macro_def) => {
+                self.update_reachability_from_macro(item.def_id, macro_def);
+            }
             hir::ItemKind::ForeignMod { items, .. } => {
                 for foreign_item in items {
                     if foreign_item.vis.node.is_pub() {
@@ -715,6 +773,7 @@ impl Visitor<'tcx> for EmbargoVisitor<'tcx> {
                     }
                 }
             }
+
             hir::ItemKind::OpaqueTy(..)
             | hir::ItemKind::Use(..)
             | hir::ItemKind::Static(..)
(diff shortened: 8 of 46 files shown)


inquisitivecrystal:
I'm consulting with the rustdoc team about the best way to solve this problem. [Link to the Zulip thread.](https://rust-lang.zulipchat.com/#narrow/stream/266220-rustdoc/topic/Doctests.20on.20private.20macros)

cjgillot:
Thanks @inquisitivecrystal. I will take a look when I get back. 
@bors try @rust-timer queue

LeSeulArtichaut:
Looks like rust-timer didn't queue this PR: @rust-timer queue

petrochenkov:
Two general comments:
- I'm not sure literally moving `#[macro_export]` macros to the root in def collector and AST lowering is the best alternative.
    For all purposes a `#[macro_export] macro_rules!` item exists at the point at which it's written in source code (for `#[allow]` attributes, for example) and behaves like non-exported `macro_rules!` items, *except* that it inserts some kind of "reexport" of itself to the root module during metadata encoding.
- The `is_exported` property as it is defined now doesn't really makes sense and shouldn't be kept as a flag in HIR (unless perhaps temporary, for preserving bug-compatibility). Visibility/reachability properties are already correctly provided for all macros with `visibility` and `privacy_access_levels` queries and otherwise exported `macro_rules` should be treated identically to non-exported `macro_rules` (modulo the metadata encoding bit mentioned above).

petrochenkov:
>...and behaves like non-exported `macro_rules!` items, *except* that it inserts some kind of "reexport" of itself to the root module during metadata encoding.

This would be similar to AST-level treatment of `#[macro_export] macro_rules!` where it's treated as a regular `macro_rules!` item, except that it adds an additional `NameBinding` to the root (see `fn define_macro` in resolve).

cjgillot:
@rust-timer build 2a825dbd5f82b9c9b0213aa4be81e09172d8122c

inquisitivecrystal:
> Two general comments:
> 
> * I'm not sure literally moving `#[macro_export]` macros to the root in def collector and AST lowering is the best alternative.
>   For all purposes a `#[macro_export] macro_rules!` item exists at the point at which it's written in source code (for `#[allow]` attributes, for example) and behaves like non-exported `macro_rules!` items, _except_ that it inserts some kind of "reexport" of itself to the root module during metadata encoding.

It was your idea in the first place! :) Seriously though, I'd be thrilled if I could remove the code for hoisting `#[macro_export] macro_rules!` items to the top level. It's both the messiest and least maintainable part of my code.

That being said, I don't really know what to replace it with. The main reason this was necessary was that rustdoc needed to see exported macros at the top level rather than where they're defined. Rustdoc works solely from the HIR: it doesn't see anything we do to the metadata. How do I make rustdoc put `#[macro_export]` macros in the right place?

Also, I'm not sure that it's necessary to do anything during metadata encoding. As I recall, everything was building before I moved exported macros to the top level; the only reason I needed to move the items was to get rustdoc to work right. Am I missing something?

> * The `is_exported` property as it is defined now doesn't really makes sense and shouldn't be kept as a flag in HIR (unless perhaps temporary, for preserving bug-compatibility). Visibility/reachability properties are already correctly provided for all macros with `visibility` and `privacy_access_levels` queries and otherwise exported `macro_rules` should be treated identically to non-exported `macro_rules` (modulo the metadata encoding bit mentioned above).

HIR normally contains the visibility of an item, through `item.vis.node.is_pub()`. This doesn't work correctly for macros, because `#[macro_export]` macros are always public regardless of their `vis`. It feels a bit weird to have less information available in the HIR for macros than other items. In practical terms though, I guess I could have a go of replacing the places I use `is_exported` with the queries and let you know if I bump into problems?

petrochenkov:
@inquisitivecrystal 
>As I recall, everything was building before I moved exported macros to the top level; the only reason I needed to move the items was to get rustdoc to work right. Am I missing something?

Macros are implemented through hacks in one more place - they are actually encoded to metadata as reexports and not as regular items.
You can see in `fn encode_info_for_mod` how `module_exports` are encoded, and those `module_exports` table is filled in `fn finalize_resolutions_in`, where it's called `export_map` and where macro bindings (`macro` items and `#[macro_export] macro_rules` root bindings) are added to it in addition to regular `use` imports.

If rustdoc can document reexports (and it can, probably through the `item_children` query), then it can document exported macros as well.
I suspect that rustdoc simply needs to skip `hir::Item` macros unless `--document-private-items` is enabled, and continue documenting them through reexports.

>In practical terms though, I guess I could have a go of replacing the places I use is_exported with the queries and let you know if I bump into problems?

Yeah, and if that ends up being non-trivial, it can be deferred to a separate PR.

petrochenkov:
>except that it inserts some kind of "reexport" of itself to the root module during metadata encoding

I suspect that this is the only place where we want to *continue* encoding macros with reexports, otherwise it would be good to remove the aforementioned hack and stop adding macros to the `module_exports` list in resolve.

petrochenkov:
>HIR normally contains the visibility of an item, through `item.vis.node.is_pub()`. This doesn't work correctly for macros, because `#[macro_export]` macros are always public regardless of their `vis`.

If the "syntactic" visibility is used for something after lowering to HIR instead of `tcx.visibility()`, then it's likely an unnoticed bug.

inquisitivecrystal:
> > HIR normally contains the visibility of an item, through `item.vis.node.is_pub()`. This doesn't work correctly for macros, because `#[macro_export]` macros are always public regardless of their `vis`.
> 
> If the "syntactic" visibility is used for something after lowering to HIR instead of `tcx.visibility()`, then it's likely an unnoticed bug.

That's [interesting to hear](https://github.com/rust-lang/rust/pull/88019/files#diff-4b76d999d0f2e74f6df3724a36a6b96302b16968a2f7af48d2b4c08048f8f871L260-R217). Should that be switched to `tcx.visibility()`?

inquisitivecrystal:
Well, that was worse than I was really expecting. I'm hoping that yanking out the hoisting to top level stuff should improve things some, though I don't know how much. I guess we'll have to see.

petrochenkov:
>That's [interesting to hear](https://github.com/rust-lang/rust/pull/88019/files#diff-4b76d999d0f2e74f6df3724a36a6b96302b16968a2f7af48d2b4c08048f8f871L260-R217). Should that be switched to `tcx.visibility()`?

I don't know, the logic in rustdoc often remind [this thing](https://en.wikipedia.org/wiki/Celestial_Emporium_of_Benevolent_Knowledge) in its consistency, so the best bet is just to preserve the old behavior.
(An then perhaps overhaul that code, and see what breaks, and talk with rustdoc maintainers to figure out what the intentions were.)

src/librustdoc/visit_ast.rs
```diff
@@ -257,7 +210,13 @@ impl<'a, 'tcx> RustdocVisitor<'a, 'tcx> {
         debug!("visiting item {:?}", item);
         let name = renamed.unwrap_or(item.ident.name);
 
-        if item.vis.node.is_pub() {
+        let is_pub = if let hir::ItemKind::Macro { is_exported: true, .. } = item.kind {
```
cjgillot, line 213: This line makes me wonder if we could replace `is_exported` by making `item.vis` public.

inquisitivecrystal:
> > That's [interesting to hear](https://github.com/rust-lang/rust/pull/88019/files#diff-4b76d999d0f2e74f6df3724a36a6b96302b16968a2f7af48d2b4c08048f8f871L260-R217). Should that be switched to `tcx.visibility()`?
> 
> I don't know, the logic in rustdoc often remind [this thing](https://en.wikipedia.org/wiki/Celestial_Emporium_of_Benevolent_Knowledge) in its consistency, so the best bet is just to preserve the old behavior.
> (An then perhaps overhaul that code in a separate PR, and see what breaks, and talk with rustdoc maintainers to figure out what the intentions were.)

I changed it in my local copy and the tests are still passing, so... *shrug*

cjgillot (commented):
Thank you for the work @inquisitivecrystal!
I agree with @petrochenkov that we should not attempt to move the exported macros to the crate root. I'll let you two discuss on the right path here.

I left a few comments on the current state of the PR. I will do a deeper pass once we figure out the top-level reexport story.

src/librustdoc/visit_ast.rs
jyn514: Isn't it already public? This used `item.vis.node.is_pub()` before - @inquisitivecrystal why did you change it?

inquisitivecrystal: The problem is that a `#[macro_export] macro_rules!` item doesn't have a `vis.node` that shows it's public, but is anyway. petrochenkov suggested that I could replace situations like this with calls to `tcx.visibility()`, which seems work fine.

compiler/rustc_ast_lowering/src/item.rs
```diff
@@ -465,7 +442,26 @@ impl<'hir> LoweringContext<'_, 'hir> {
...
-            ItemKind::MacroDef(..) | ItemKind::MacCall(..) => {
+            ItemKind::MacroDef(MacroDef { ref body, macro_rules }) => {
+                let is_exported = !macro_rules
+                    || attrs.map_or(false, |a| self.sess.contains_name(a, sym::macro_export));
+
+                let body = P(self.lower_mac_args(body));
+
+                hir::ItemKind::Macro {
+                    // `is_exported` is duplicated, to make matching more convenient.
+                    is_exported,
```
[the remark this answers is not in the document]
inquisitivecrystal: I would have argued with you on this, but petrochenkov told me it doesn't make sense for it to be there. Now it's gone from being there twice to being there zero times.

compiler/rustc_hir/src/stable_hash_impls.rs
```diff
@@ -193,7 +193,7 @@ impl<HirCtx: crate::HashStableContext> HashStable<HirCtx> for Item<'_> {
 
 impl<HirCtx: crate::HashStableContext> HashStable<HirCtx> for MacroDef<'_> {
     fn hash_stable(&self, hcx: &mut HirCtx, hasher: &mut StableHasher) {
-        let MacroDef { ident, def_id: _, ref ast, ref vis, span } = *self;
+        let MacroDef { is_exported: _, ident, def_id: _, ref ast, ref vis, span } = *self;
```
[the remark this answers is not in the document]
inquisitivecrystal: It no longer exists in my local version.

compiler/rustc_ast_lowering/src/lib.rs
```diff
@@ -597,6 +616,19 @@ impl<'a, 'hir> LoweringContext<'a, 'hir> {
         id
     }
 
+    /// This checks if an item is a `#[macro_use]` macro, and thus always
+    /// defined at the top level. As a special case, if the macro is already
+    /// at the top level, it remains at its current level.
+    fn check_force_top_level(&self, item: &Item) -> bool {
+        if self.is_at_top_level {
```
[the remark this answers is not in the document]
inquisitivecrystal: I'm pretty sure it isn't. If you go look at the cases where I set `is_at_top_level`, I don't think all of them are covered by checking `self.current_hir_id_owner.0`. I think I tried that at some point and it didn't work. It's somewhat irrelevant though, because if we don't move the exported macros to the crate root all of this code disappears.

compiler/rustc_metadata/src/rmeta/decoder.rs
```diff
@@ -1100,7 +1101,11 @@ impl<'a, 'tcx> CrateMetadataRef<'a> {
                     let vis = self.get_visibility(child_index);
                     let def_id = self.local_def_id(child_index);
                     let res = Res::Def(kind, def_id);
-                    callback(Export { res, ident, vis, span });
+
+                    if !kind.is_macro() {
```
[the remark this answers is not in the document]
inquisitivecrystal: I was going to mention this, and then I forgot. Serves me right for not leaving a comment. The truth of the matter is, I have absolutely no clue why I needed to skip macros here. If I don't do so, I get errors (which it took me five+ hours to debug :/). I think those may come from situations in which a macro is defined as a macro by example, but also defined as a builtin macro. Here's what the errors look like:

```
error[E0428]: the name `Clone` is defined multiple times
   --> /rust/library/core/src/clone.rs:139:1
    |
139 | pub macro Clone($item:item) {
    | ^^^^^^^^^^^^^^^^^^^^^^^^^^^
    | |
    | `Clone` redefined here
    | previous definition of the macro `Clone` here
    |
    = note: `Clone` must be defined only once in the macro namespace of this module

error[E0428]: the name `Copy` is defined multiple times
   --> /rust/library/core/src/marker.rs:393:1
    |
393 | pub macro Copy($item:item) {
    | ^^^^^^^^^^^^^^^^^^^^^^^^^^
    | |
    | `Copy` redefined here
    | previous definition of the macro `Copy` here
    |
    = note: `Copy` must be defined only once in the macro namespace of this module
```

I tried to fix this by not encoding the macros, rather than by skipping decoding them, but I couldn't get that to work either. Any insight into what might be going on here would be helpful. Although this solution works, I have a strong feeling that it likely isn't the right solution.

src/librustdoc/visit_ast.rs
```diff
@@ -257,7 +228,9 @@ impl<'a, 'tcx> RustdocVisitor<'a, 'tcx> {
         debug!("visiting item {:?}", item);
         let name = renamed.unwrap_or(item.ident.name);
 
-        if item.vis.node.is_pub() {
+        let is_pub = self.cx.tcx.visibility(item.def_id) == Visibility::Public;
```
jyn514, line 231: Hmm, I think this behaves differently for Inherited visibility: In `pub enum E { A, B }`, A and B will have `ty::Visibility::Public` if you call `visibility()`, but `ast::VisibilityKind::Inherited` if you call `node.is_pub()`. So this is a change in behavior, and I'm a little nervous about it.

That said, this does seem more correct to me - this should be using the computed visibility, not the syntactic visibility. Can you add a test that this now inlines enum variants? Right now it doesn't:
```rust
pub mod x {
    pub enum E { A, B }
}

#[doc(inline)]
pub use x::E::A;

#[doc(inline)]
pub use x::E;
```
image

jyn514: Actually, I'm not sure what it means to inline variants :/ they don't get their own documentation page (and the same for trait methods, which also have inherited visibility). So please do add the test, I'm curious to see what it changes. We may need to keep the original behavior depending on what it does.

jyn514:
> > That's [interesting to hear](https://github.com/rust-lang/rust/pull/88019/files#diff-4b76d999d0f2e74f6df3724a36a6b96302b16968a2f7af48d2b4c08048f8f871L260-R217). Should that be switched to `tcx.visibility()`?
> 
> I don't know, the logic in rustdoc often remind [this thing](https://en.wikipedia.org/wiki/Celestial_Emporium_of_Benevolent_Knowledge) in its consistency, so the best bet is just to preserve the old behavior.
> (An then perhaps overhaul that code in a separate PR, and see what breaks, and talk with rustdoc maintainers to figure out what the intentions were.)

See https://github.com/rust-lang/rust/pull/88019#discussion_r689221354.

src/librustdoc/visit_ast.rs
inquisitivecrystal: Tested locally. It doesn't seem to do anything. The inline attribute is silently ignored, the same way it is on stable. I'll add an actual test for it when I get a chance.

compiler/rustc_ast_lowering/src/item.rs
```diff
@@ -465,7 +442,18 @@ impl<'hir> LoweringContext<'_, 'hir> {
...
             ),
-            ItemKind::MacroDef(..) | ItemKind::MacCall(..) => {
+            ItemKind::MacroDef(MacroDef { ref body, macro_rules }) => {
+                let body = P(self.lower_mac_args(body));
+
+                hir::ItemKind::Macro(hir::MacroDef {
+                    ident: *ident,
+                    vis: *vis,
+                    def_id: hir_id.expect_owner(),
+                    span,
```
cjgillot, line 452: `def_id`, `ident`, `vis` and `span` are already in the `hir::Item`. `hir::ItemKind::Macro` only needs to contain the `ast::MacroDef`.

compiler/rustc_ast_lowering/src/item.rs
```diff
@@ -212,28 +211,6 @@ impl<'hir> LoweringContext<'_, 'hir> {
     pub fn lower_item(&mut self, i: &Item) -> Option<hir::Item<'hir>> {
```
cjgillot, line 211: This function always returns `Some`.

compiler/rustc_ast_lowering/src/item.rs
```diff
@@ -212,28 +211,6 @@ impl<'hir> LoweringContext<'_, 'hir> {
...
-                    let a = self.lower_attr(a);
-                    self.non_exported_macro_attrs.push(a);
-                }
-            }
-            return None;
-        }
-
         let hir_id = self.lower_node_id(i.id);
         let attrs = self.lower_attrs(hir_id, &i.attrs);
         let kind = self.lower_item_kind(i.span, i.id, hir_id, &mut ident, attrs, &mut vis, &i.kind);
```
cjgillot, line 216: Can `lower_item` and `lower_item_kind` be merged? (This can be left to a follow-up PR).

compiler/rustc_metadata/src/rmeta/decoder.rs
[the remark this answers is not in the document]
cjgillot: I think I've got at least part of the story. This error is created by resolution code when it enumerates all the children of a given module. Formerly, the macro did not appear in the HIR module list of items (your change to lower_item_ref), so this code path never included any macro. Now, macros are listed in the module, so appear when resolution tries to construct the reduced graph. Except that the resolver already lists all macros as reexports (rustc_resolve::imports::finalize_resolutions_in, line 1396).
Does removing the `|| binding.is_macro_def()` there work?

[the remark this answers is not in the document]
inquisitivecrystal: Well, it changes the errors I get. Here's a sample of the new ones:

```
error[E0432]: unresolved import `core::format_args`
   --> library/alloc/src/lib.rs:197:13
    |
197 |     pub use core::format_args;
    |             ^^^^^^^^^^^^^^^^^ no `format_args` in the root

error: cannot find macro `panic` in this scope
   --> library/alloc/src/alloc.rs:387:9
    |
387 |         panic!("memory allocation of {} bytes failed", size)
    |         ^^^^^

error: cannot find macro `unreachable` in this scope
   --> library/alloc/src/borrow.rs:281:37
    |
281 |                     Borrowed(..) => unreachable!(),
    |                                     ^^^^^^^^^^^

error: cannot find macro `write` in this scope
   --> library/alloc/src/collections/btree/map/entry.rs:100:9
    |
100 |         write!(
    |         ^^^^^
```

compiler/rustc_ast_lowering/src/item.rs
inquisitivecrystal: I know, but it's not my function and I'm no longer touching that part of the code.

compiler/rustc_ast_lowering/src/item.rs
inquisitivecrystal: They could, yeah. I think this layout is cleaner though.

compiler/rustc_ast_lowering/src/item.rs
inquisitivecrystal: I tried to do that, but I bumped into problems with intravisit::Visitor::visit_macro_def, which needs to receive a hir::MacroDef. I couldn't figure out how to make the lifetimes work for that if the MacroDef didn't already exist. Note that we also support, for instance, Visitor::visit_fn, so I'm not sure that Visitor::visit_macro_def should be considered redundant with Visitor::visit_item.

compiler/rustc_passes/src/stability.rs
```diff
@@ -613,10 +613,11 @@ impl<'tcx> Visitor<'tcx> for MissingStabilityAnnotations<'tcx> {
...
     fn visit_item(&mut self, i: &'tcx Item<'tcx>) {
-        // Inherent impls and foreign modules serve only as containers for other items,
-        // they don't have their own stability. They still can be annotated as unstable
-        // and propagate this unstability to children, but this annotation is completely
-        // optional. They inherit stability from their parents when unannotated.
+        // Inherent impls, foreign modules, and non exported macros serve only as
+        // containers for other items, they don't have their own stability. They still can
+        // be annotated as unstable and propagate this unstability to children, but this
+        // annotation is completely optional. They inherit stability from their parents
+        // when unannotated.
```
inquisitivecrystal, line 620: Note to self: roll back this documentation change.

compiler/rustc_lint/src/builtin.rs
```diff
@@ -635,7 +617,8 @@ impl<'tcx> LateLintPass<'tcx> for MissingDoc {
                 return;
             }
 
-            hir::ItemKind::TyAlias(..)
+            hir::ItemKind::Macro { .. }
+            | hir::ItemKind::TyAlias(..)
```
inquisitivecrystal, line 621: Note to self: the ordering is a bit weird here.

src/librustdoc/visit_ast.rs
inquisitivecrystal: I think it might be better for me to separate the change over to `tcx.visibility` into a separate PR and do something more conservative here. It seems like this might cause a regression and we might end up having to revert. Thoughts?

compiler/rustc_ast_lowering/src/item.rs
cjgillot: There are only two uses of `visit_macro_def` left: `rustc_passes::{hir_stats, stability}`. I think the function can be removed altogether.

compiler/rustc_ast_lowering/src/item.rs
cjgillot: ok

src/librustdoc/visit_ast.rs
cjgillot: A separate PR is possible, yes.

compiler/rustc_ast_lowering/src/item.rs
inquisitivecrystal: Done!

inquisitivecrystal:
I just want to state, for the record, my request that this not be merged until I have a chance to clean up the commits. No point doing that till we're closer to being done though.

cjgillot:
LGTM. I will approve once you get to cleanup commit history and PR message + add a comment in `rustc_metadata::rmeta::decoder` stating that resolve creates reexports for macros which create conflicts with the item themselves for downstream crates.

compiler/rustc_hir_pretty/src/lib.rs
```diff
@@ -660,6 +660,28 @@ impl<'a> State<'a> {
...
+                    Some(item.ident),
+                    macro_def.body.delim(),
+                    &macro_def.body.inner_tokens(),
+                    true,
+                    item.span,
+                );
+
+                if macro_def.body.need_semicolon() {
+                    self.word(";");
+                }
```
petrochenkov, line 683: Instead of copy-pasting this can be factored out into a method in `compiler\rustc_ast_pretty\src\pprust\state.rs`.

compiler/rustc_hir/src/def.rs
```diff
@@ -218,6 +218,13 @@ impl DefKind {
             | DefKind::Impl => None,
         }
     }
+
+    pub fn is_macro(self) -> bool {
+        match self {
+            DefKind::Macro(..) => true,
+            _ => false,
+        }
+    }
```
petrochenkov, line 227: I don't think this method is necessary, it's trivial `matches!(kind, DefKind::Macro(..))` and it's not commonly used.

src/test/ui/lint/lint-level-macro-def.rs
```diff
@@ -0,0 +1,17 @@
...
+
+
+#[deny(missing_docs)]
+mod module {
+    #[allow(missing_docs)]
+    #[macro_export]
+    macro_rules! hello {
+        () => ()
+    }
+}
```
petrochenkov, line 15: Could you add a test case for this case
```rust
#[allow(missing_docs)]
mod module {
    #[macro_export]
    macro_rules! hello {
        () => ()
    }
}
```
, it'll make sure that macros exported from non-root modules still have those modules as parents from linting point of view.

compiler/rustc_privacy/src/lib.rs
```diff
@@ -879,47 +931,6 @@ impl Visitor<'tcx> for EmbargoVisitor<'tcx> {
...
-        loop {
-            let changed_reachability =
-                self.update_macro_reachable(module_def_id, macro_module_def_id);
-            if changed_reachability || module_def_id == CRATE_DEF_ID {
-                break;
-            }
-            module_def_id =
-                ty::DefIdTree::parent(self.tcx, module_def_id.to_def_id()).unwrap().expect_local();
-        }
-    }
```
petrochenkov, line 922: Could you keep this logic in a separate function instead of inlining it to `EmbargoVisitor::visit_item`?

compiler/rustc_privacy/src/lib.rs
```diff
@@ -642,6 +639,21 @@ impl Visitor<'tcx> for EmbargoVisitor<'tcx> {
             }
             // Foreign modules inherit level from parents.
             hir::ItemKind::ForeignMod { .. } => self.prev_level,
+            hir::ItemKind::Macro(ref macro_def) => {
```
petrochenkov, line 642:
```suggestion
            hir::ItemKind::Macro(MacroDef { macro_rules: true, .. }) => {
```
The non-`macro_rules` case is identical to other items below.

compiler/rustc_privacy/src/lib.rs
```diff
@@ -507,11 +507,7 @@ impl EmbargoVisitor<'tcx> {
         }
         match def_kind {
             // No type privacy, so can be directly marked as reachable.
-            DefKind::Const
-            | DefKind::Macro(_)
-            | DefKind::Static
-            | DefKind::TraitAlias
-            | DefKind::TyAlias => {
+            DefKind::Const | DefKind::Static | DefKind::TraitAlias | DefKind::TyAlias => {
```
petrochenkov, line 510: Any specific reason for this change?
The old logic seemed correct to me.

compiler/rustc_metadata/src/rmeta/decoder.rs
```diff
@@ -1036,6 +1036,7 @@ impl<'a, 'tcx> CrateMetadataRef<'a> {
                         self.local_def_id(def_index),
                     );
                     let ident = self.item_ident(def_index, sess);
+
```
petrochenkov, line 1039: Nit: random whitespace change.

compiler/rustc_metadata/src/rmeta/decoder.rs
```diff
@@ -1100,7 +1101,11 @@ impl<'a, 'tcx> CrateMetadataRef<'a> {
                     let vis = self.get_visibility(child_index);
                     let def_id = self.local_def_id(child_index);
                     let res = Res::Def(kind, def_id);
-                    callback(Export { res, ident, vis, span });
+
+                    if !kind.is_macro() {
+                        callback(Export { res, ident, vis, span });
+                    }
```
petrochenkov, line 1107: Could you add a comment/FIXME explaining that macros are currently encoded twice as items and as reexports, so we are ignoring the items when decoding module children and only use the reexports.

src/librustdoc/visit_ast.rs
```diff
@@ -79,50 +79,25 @@ impl<'a, 'tcx> RustdocVisitor<'a, 'tcx> {
...
-                match cur_mod.mods.iter_mut().find(|child| child.name == path_segment_ty_ns) {
-                    Some(child_mod) => cur_mod = &mut *child_mod,
-                    None => continue 'exported_macros,
+
+        // `#[macro_export] macro_rules!` items are reexported at the top level of the
+        // crate, regardless of where they're defined. We want to document the
+        // top level rexport of the macro, not its original definition, since
+        // the rexport defines the path that a user will actually see. Accordingly,
+        // we add the rexport as an item here, and then skip over the original
+        // definition in `visit_item()` below.
```
petrochenkov, line 88: So, if non-exported `macro_rules` are documented at their in-source definition point when `--document-private-items` is enabled, then why exported `macro_rules` are not? :)

It would be pretty reasonable/consistent to document exported macros twice in the doc-private mode - once as an item and once as the root "reexport" (especially if that would make rustdoc logic simpler).

jyn514:

This fixes several long-standing bugs relating to macros, and the performance regression is quite minor.

compiler/rustc_hir/src/def.rs
inquisitivecrystal: I personally think `!matches!()` looks atrocious. That said, fair enough.

compiler/rustc_metadata/src/rmeta/decoder.rs
inquisitivecrystal: Is there a reasonably simple way to fix it? This workaround feels very hacky to me.

src/librustdoc/visit_ast.rs
inquisitivecrystal: Documenting them twice feels weird to me, especially given that the only path they can be addressed at is the root path. Would it be worth asking the rustdoc team?

compiler/rustc_metadata/src/rmeta/decoder.rs
petrochenkov: Yes, something like https://github.com/cjgillot/rust/commit/c147d8fdf8c4428ac94216f9629816b357c05435.
I think it's ok to land this as is in this PR.

src/librustdoc/visit_ast.rs
jyn514: @petrochenkov I don't understand your suggestion. Why would rustdoc document macros at the definition if they're not accessible using that path? That seems actively misleading.

petrochenkov: >given that the only path they can be addressed at is the root path

Non-exported macros cannot be addressed by any path, but they are still documented when `--document-private-items` is enabled.
This comment is certainly not a blocking one though.

jyn514: > Non-exported macros cannot be addressed by any path, but they are still documented when --document-private-items is enabled.

Can't they be addressed by that path by the module that defined them?

petrochenkov: No, `macro_rules` macros are not "modularized" and cannot be addressed by a module-relative paths.
Moreover, there can be multiple `macro_rules` with the same name in the same module.

jyn514: > Moreover, there can be multiple macro_rules with the same name in the same module.

Oh fun ... Rustdoc almost certainly can't handle that. @inquisitivecrystal can you add a test for that case? I expect all the macro.name.html files will be overwritten by each other.

compiler/rustc_privacy/src/lib.rs
inquisitivecrystal: Yeah, there was, although looking back this may not be the right fix. The problem is that otherwise you start getting "macro has missing stability attribute" errors for non-exported macros. It would be possible to work around that, but I thought it would require reconstructing macro expectedness in atom `compiler/rustc_passes/src/stability.rs`. A potential alternative would be to only skip marking the item reachable for nonexported macros (in other words, handle macro expectedness here). I'm not sure which one of these solutions is actually correct.

compiler/rustc_metadata/src/rmeta/decoder.rs
inquisitivecrystal: Alright, I'll leave it this way for now and do the comment.

compiler/rustc_hir_pretty/src/lib.rs
inquisitivecrystal: It can, but it's a bit less trivial than it may at first appear. In `rustc_ast_pretty`, the visibility is an `ast::Visibility`. By contrast, in `rustc_hir_pretty`, the visibility is an `hir::Visibility`. As far as I can tell, there's no way to interconvert between the two, though I'd be overjoyed if someone told me I'm wrong about that.

That means that we'd end up needing to make the function generic with respect to the the type of visibility. The first pass I did has a function signature that looks like this:

```rust
fn print_mac_def<V>(
        &mut self,
        macro_def: &ast::MacroDef,
        ident: &Ident,
        vis: &V,
        sp: &Span,
        print_visibility: impl FnOnce(&mut Self, &V),
    )
```

It would probably be cleaner to use traits somehow, though I'm not sure what the best way to go about that is. In any case, this approach does add some amount of complexity. The DRY principle may be enough to justify that though.

compiler/rustc_metadata/src/rmeta/decoder.rs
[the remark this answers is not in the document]
inquisitivecrystal: This got resolved [down here](https://github.com/rust-lang/rust/pull/88019#discussion_r691626636). Not sure why that got started as a seperate thread.

inquisitivecrystal:
This should be ready for further review and possible approval! 🎉

petrochenkov (approved):

petrochenkov:
The `macro-document-private-duplicate.rs` test introduced in this PR causes issues - https://github.com/rust-lang/rust/issues/89228.

PR #88025, opened 2021-08-14
ScmCredentials netbsd implementation.

devnexen:


bjorn3:
Does the same implementation work on other bsd's like openbsd and freebsd? If so can you add it to those platforms too?

devnexen:
I wish that was the case :-) but for example freebsd is a different beast in that regard.

joshtriplett:
I'm not the right person to review non-trivial BSD changes, in general. I think someone else should take a look at this.

Who are the BSD target maintainers?

devnexen:
do not know if there is a dedicated one for netbsd, being low tier platform.

joshtriplett:
@devnexen All new targets, even tier 3 targets, are supposed to have at least one maintainer. That isn't necessarily the case for older targets, but we should work towards identifying maintainers for existing targets, or considering the possibility of deprecation if there are zero people available to maintain them.

devnexen:
I see often contributions by unique individuals, no recurring patterns. I know who are the freebsd team and the most visible openbsd expert but netbsd, hard to know...

Dylan-DPC:
@Mark-Simulacrum any clue who should review BSD changes?

Amanieu:
@devnexen Have you tried running the `test_send_vectored_with_ancillary_to_unix_datagram` test from the standard library on netbsd? You can run it with `./x.py test library/std --test-args test_send_vectored_with_ancillary_to_unix_datagram`.

Other than that the code looks fine to me.

devnexen:
@Amanieu after rebasing relevant tests pass.

library/std/src/os/unix/net/ancillary.rs
```diff
@@ -252,11 +313,15 @@ impl<'a> Iterator for ScmRights<'a> {
 /// This control message contains unix credentials.
 ///
 /// The level is equal to `SOL_SOCKET` and the type is equal to `SCM_CREDENTIALS` or `SCM_CREDS`.
-#[cfg(any(doc, target_os = "android", target_os = "linux",))]
```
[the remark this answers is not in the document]
Amanieu: The best way to achieve that is probably to have a separate definition when building docs on an unsupported target:
```rust
#[cfg(all(doc, not(target_os = "linux", target_os = "android", target_os = "netbsd"))]
#[unstable(feature = "unix_socket_ancillary_data", issue = "76915")]
#[derive(Clone)]
pub struct SocketCred(());
```

library/std/src/os/unix/net/ancillary.rs
```diff
@@ -15,6 +15,7 @@ use crate::sys::net::Socket;
 mod libc {
     pub use libc::c_int;
     pub struct ucred;
+    pub struct sockcred;
```
Amanieu, line 18: This is no longer needed. Also in the cfg above you should exclude netbsd.

library/std/src/os/unix/net/ancillary.rs
```diff
@@ -177,13 +178,23 @@ impl<'a, T> Iterator for AncillaryDataIter<'a, T> {
...
 #[derive(Clone)]
 pub struct SocketCred(libc::ucred);
 
-#[cfg(any(doc, target_os = "android", target_os = "linux",))]
+#[cfg(target_os = "netbsd")]
+#[unstable(feature = "unix_socket_ancillary_data", issue = "76915")]
+#[derive(Clone)]
+pub struct SocketCred(libc::sockcred);
+
+#[cfg(any(target_os = "android", target_os = "linux",))]
```
Amanieu, line 197: Add `doc` here so the methods appear in the docs.

library/std/src/os/unix/net/ancillary.rs
```diff
@@ -234,6 +245,62 @@ impl SocketCred {
     }
 }
 
+#[doc(cfg(target_os = "netbsd"))]
```
Amanieu, line 248: This line is not needed.

PR #88028, opened 2021-08-14
Disable unused_must_use for statically known bools

### LINKED ISSUE #69466: False warning: "unused logical operation that must be used"
In a loop, in some cases, rustc seems to not understand that a boolean expression was used for sideeffects, and warns about unused logical operation.  The following code can quickly replicate the problem.

```rust
use std::sync::{atomic::{AtomicBool, Ordering}, Arc};


fn main() {
    let running = Arc::new(AtomicBool::from(true));
    
    {
        let running = Arc::clone(&running);
        std::thread::spawn(move || {
            running.swap(false, Ordering::Relaxed);
        });
    }
    
    loop {
        // ...
        println!("Running....");
        
        !running.load(Ordering::Relaxed) && break;
    }
    
    println!("Exhausted >.>");
}

```

The expected result is that rustc doesn't output a warning in this case.

### Meta

* rustc version: 1.41.0

### LINKED ISSUE #88017: Disable unused_must_use for statically known bool values
In this code ([playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=13b6c01bca742a083b0f48fee07b21d3)) compiler with warn `is_prime(x) || continue;` with `unused logical operation that must be used`. However, if we add a `let temp = is_prime(x) || continue;` the `temp` would be always true, if `is_prime(x)` is true it will be short-circuited with true, and otherwise it would be diverge and we will never see anything other than true. It is like warning unused_must_use on `Result<(), !>` which it is always `Ok(())` and using it is pointless. So is it good to turn off this warning? Or current behavior is intended?

I see that this isn't a good code, but unused_must_use isn't a good warning for preventing that. Another warning (probably in clippy) can say that `please don't abuse short circuit operators`.

hkalbasi:
I'm still not sure if this PR contradicts unused_must_use spirit or not, but I don't think anyone is interested in storing what is statically known and warning in this case has no benefit.

Fix #88017
Fix #69466
CC #85913 #87607
r? @estebank

### DIFF
--- a/compiler/rustc_lint/src/unused.rs
+++ b/compiler/rustc_lint/src/unused.rs
@@ -89,6 +89,59 @@ declare_lint! {
 }
 
 declare_lint_pass!(UnusedResults => [UNUSED_MUST_USE, UNUSED_RESULTS]);
+#[derive(Debug)]
+enum LogicPredicts {
+    Unknown,
+    True,
+    False,
+    Diverging,
+}
+
+fn is_never_expr(cx: &LateContext<'_>, e: &hir::Expr<'_>) -> bool {
+    if cx.typeck_results().expr_ty(e).is_never() {
+        return true;
+    }
+    match e.kind {
+        rustc_hir::ExprKind::Block(ref inner, _) => {
+            if let Some(ein) = inner.expr {
+                return is_never_expr(cx, ein);
+            }
+            return false;
+        }
+        _ => false,
+    }
+}
+
+fn predicte_logic_operation(cx: &LateContext<'_>, e: &hir::Expr<'_>) -> LogicPredicts {
+    if is_never_expr(cx, e) {
+        return LogicPredicts::Diverging;
+    }
+    let result = match e.kind {
+        hir::ExprKind::Binary(bin_op, e1, e2) => match bin_op.node {
+            hir::BinOpKind::And => match predicte_logic_operation(cx, e1) {
+                LogicPredicts::False => LogicPredicts::False,
+                LogicPredicts::True => predicte_logic_operation(cx, e2),
+                LogicPredicts::Diverging => LogicPredicts::Diverging,
+                LogicPredicts::Unknown => match predicte_logic_operation(cx, e2) {
+                    LogicPredicts::Diverging | LogicPredicts::False => LogicPredicts::False,
+                    _ => LogicPredicts::Unknown,
+                },
+            },
+            hir::BinOpKind::Or => match predicte_logic_operation(cx, e1) {
+                LogicPredicts::True => LogicPredicts::True,
+                LogicPredicts::False => predicte_logic_operation(cx, e2),
+                LogicPredicts::Diverging => LogicPredicts::Diverging,
+                LogicPredicts::Unknown => match predicte_logic_operation(cx, e2) {
+                    LogicPredicts::Diverging | LogicPredicts::True => LogicPredicts::True,
+                    _ => LogicPredicts::Unknown,
+                },
+            },
+            _ => LogicPredicts::Unknown,
+        },
+        _ => LogicPredicts::Unknown,
+    };
+    result
+}
 
 impl<'tcx> LateLintPass<'tcx> for UnusedResults {
     fn check_stmt(&mut self, cx: &LateContext<'_>, s: &hir::Stmt<'_>) {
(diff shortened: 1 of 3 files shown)


ecstatic-morse:
The title is a bit misleading to me, since this is really about disabling `unused_must_use` when the right-hand side has side-effects (in this case diverges). Accordingly, the `LogicPredicts` stuff seems unnecessary: we only care about the right-hand side. There are mentoring instructions in [this comment](https://github.com/rust-lang/rust/issues/69466#issuecomment-593576911), although the one  right after suggests removing `unused_must_use` entirely for short-circuiting logic, since divergence is just one possible side-effect. FWIW, I don't agree with the latter.

HKalbasi:
There are slightly different mindsets here, but with the same goal. I don't think it is about side effect, for example:
```rust
is_missiles_ready() && fire_missiles();
```
In this example, `fire_missiles()` can have side effect but result of this operation is in interest, so I see unused_must_use here valid. What is special with diverging case is in this case, output can be predicted statically. That is:
```rust
let result = is_missiles_ready() && fire_missiles() || panic!("mission failed");
```
Here, `result` is always true regardless of what the code it is. So, there isn't any point in unused_must_use here.

With this mindset, there are examples that rhs leaf is divergent but result is non trivial, and in this cases, this PR still keeps warning:
```rust
let result = !is_missiles_ready() || fire_missiles() && panic!("booom");
```
Here, if `!is_missiles_ready()` is true then result is true, else if `fire_missiles()` is false then result is false. So result is interesting and unused_must_use isn't meaningless. This in this PR tests and warned.

With side effect mindset, we should disable unused_must_use for every bool expression, and this is a more radical and debate able move. I think this PR is a good moderate solution that covers common cases.

ecstatic-morse:
Interesting. I think I understand your point-of-view a bit better, but I still think it's misguided. Specifically:

> Here, if !is_missiles_ready() is true then result is true, else if fire_missiles() is false then result is false. So result is interesting and unused_must_use isn't meaningless. This in this PR tests and warned.

Consider checking the RHS for divergent expressions, which would suppress warnings in a strict superset of the cases compared to this PR. Why are we adding logic to continue emitting a warning in such a niche case? The lesson from #69466 is that people want to use short-circuiting operators in lieu of control-flow, does that not apply here?

HKalbasi:
I wanted to use this as a justification, because I felt it was controversial to suppress warning. If it is clear that side-effect-full expression should not be catch by unused_must_use (which is clear to me and I'm personally in favor of this), we can simply turn it off for `&&` and `||` in all cases. `is_missiles_ready() && fire_missiles();` is valid for control flow proposes.

Expressions with diverging rhs are strictly bigger set, but how you describe this set? Clearly side-effect-full expressions? (simple side-effect-full expressions doesn't cover `fire_missiles` case) In practice I think people use chains with single operator, or at most something like `do_sth1() && do_sth2() || fallback()` and in these cases both behave the same. Benefit of mine over rhs is it has a clear explanation and propose. If it isn't a concern (which seems it isn't) and/or rhs has benefits, I can replace it with rhs and I see no problem.

ecstatic-morse:
Let's wait for @estebank to weigh in here.

To summarize, I want the rule to be:

> `must_use` applies to the result of short-circuiting logic operators **unless** the expression on the right-hand side diverges.

While @HKalbasi wants to add an additional condition to the exception (let me know if this is wrong or if there is a different way to phrase this):

> `must_use` applies to the result of short-circuiting logic operators **unless** the expression on the right-hand side diverges **and** the non-divergent path can only result in a single result (`true` or `false`).

OP's modification is relevant in cases like the following. It will warn while the first formulation does not:

```rust
!is_missiles_ready() || (fire_missiles() && return);
```

Neither of these handles all possible side-effects, just unconditional divergence (e.g. `panic`, `return`, `break`). There's no way to do this in the general case. If we determine that false positives are unacceptable (neither me nor OP thinks so, but [eddyb disagrees](https://github.com/rust-lang/rust/issues/69466#issuecomment-593608839)), we would never lint.

joshtriplett:
We discussed this in today's @rust-lang/lang meeting.

Some people did agree that this seems like a kind of false positive, insofar as you are "using" the value through the combination of a short-circuiting operator and a diverging RHS.

However, all of the examples seem like code we're not inclined to encourage, and all of us felt like the compiler ought to produce *some* sort of warning here. In particular, all of this code seems more clearly written using `if`. For instance, instead of `expr || panic!("...")`, we'd rather see `if !expr { panic!("..."); }`.

We'd be happy to see a structured suggestion in that regard, steering people towards `if` and `if !`.

@rfcbot close

HKalbasi:
> However, all of the examples seem like code we're not inclined to encourage, and all of us felt like the compiler ought to produce some sort of warning here.

@joshtriplett I think unused warning isn't point to what is wrong with this, and suggest `let _ = expr || panic!("...")` which is worse if being used for silencing the warning. There is a [clippy lint](https://rust-lang.github.io/rust-clippy/master/#short_circuit_statement) for this, which suggest the correct `if !expr { panic!("..."); }`. I think the correct step in this direction is upgrading that lint to a compiler lint, and removing short circuit case from unused lint because it isn't really about using the result of an always `true` expression. It has better warning, better suggestion and clearly shows why compiler isn't happy with this anti-pattern.

scottmcm:
I definitely like the idea of separating it into two lints.  I could easily imagine people wanting to allow `x == 0 || continue;` but still to not allow `foo() || bar()`.  (And if that new lint for `!` RHSs gets a nice structured suggestion, that's even better.)

I wonder if rustfmt is also partially to blame here.  If the `if` ends up always being three lines (as is pretty common in a bunch of coding standards under which I've worked) then I can see extra motivation for just wanting to write things in the simpler operator form to get one line...

nikomatsakis:
Closing per the FCP! Thanks all.

PR #88031, opened 2021-08-14
Make `BuildHasher` object safe

### LINKED ISSUE #87991: BuildHasher is no longer object safe in 1.55
From 1.55 crater run #87749, [log](https://crater-reports.s3.amazonaws.com/beta-1.55-1/beta-2021-08-01/reg/probabilistic_data_structures-0.1.0/log.txt):

```
[INFO] [stdout] error[E0038]: the trait `BuildHasher` cannot be made into an object
[INFO] [stdout]    --> src/pcsa/pcsa.rs:19:27
[INFO] [stdout]     |
[INFO] [stdout] 19  |     pub hasher:       Box<dyn BuildHasher<Hasher = HasherType>>,
[INFO] [stdout]     |                           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ `BuildHasher` cannot be made into an object
[INFO] [stdout]     |
[INFO] [stdout] note: for a trait to be "object safe" it needs to allow building a vtable to allow the call to be resolvable dynamically; for more information visit <https://doc.rust-lang.org/reference/items/traits.html#object-safety>
```

#86151 added a generic method to the trait, making it no longer object safe.

This only caused one crate in crater to fail, and the source for that crate seems to have been removed from github. Still, I don't think this is technically allowed breakage.

@rustbot modify labels: +regression-from-stable-to-beta -regression-untriaged +T-libs +I-nominated

cc @scottmcm

ibraheemdev:
Resolves #87991


src/test/ui/build-hasher-object-safe.rs
```diff
@@ -0,0 +1,10 @@
+// run-pass
+
+use std::hash::BuildHasher;
+use std::collections::hash_map::{DefaultHasher, RandomState};
+
+fn ensure_object_safe(_: &dyn BuildHasher<Hasher = DefaultHasher>) {}
+
+fn main() {
+    ensure_object_safe(&RandomState::new());
+}
```
RalfJung, line 10: No need to make this a ui test I think; a test in `library/core/tests` should suffice.

cuviper:
I'm nominating this for backport since it's a beta regression and a simple fix.

PR #84176, opened 2021-04-13
Generate links to definition in rustdoc source code pages

GuillaumeGomez:
## Description

This PR adds an option (disabled by default) to add links in the source code page on ident. So for for example:

```rust
mod other_module;
struct Foo;
fn bar() {}

fn x<T: other_module::Trait>(f: Foo, g: other_module::Whatever, t: &T) {
    let f: Foo = Foo;
    bar();
    f.some_method();
}
```

In the example (mostly in the `x` function), `other_module::Trait`, `Foo`, `other_module::Whatever`, `bar` and `some_method` are now links (and `other_module` at the top too).

In case there is a type coming from another crate, it'll link to its documentation page and not its definition (but you can then click on `[src]` so I guess it's fine).

Another important detail: I voluntarily didn't add links for primitive types. I think we can discuss about adding links on them or not in a later PR (adding the support for them would require only a few lines).

Here is a video summing up everything I wrote above:

https://user-images.githubusercontent.com/3050060/114622354-21307b00-9cae-11eb-834d-f6d8178a37bd.mp4

## Performance impact

So, on my computer, the performance remains more or less the same (which is quite surprising but that's a nice surprise). Here are the numbers:

Without the option:
 * core:  1m 21s
 * alloc: 26.78s
 * std: 27.30s
 * proc_macro: 4.50s

With source to definition links generation (I enabled by default the option):
 * core: 1m 25s
 * alloc: 25.76s
 * std: 27.07s
 * proc_macro: 4.66s

So no real change here (again, I'm very surprised by this fact).

For the size of the generated source files (only taking into account the `src` folder here since it's the only one impacted) by running `du -shc .` (when I am in the source folder).

Without the option: 11.939 MB
With the option: 12.611 MB

So not a big change here either. In all those docs, I ran `grep -nR '<a class=' . | wc -l` and got 43917. So there are quite a lot of links added. :)

cc @rust-lang/rustdoc 
r? @jyn514

PR #88033, opened 2021-08-14
Add links for primitives in "jump to definition" feature

GuillaumeGomez:
Follow-up of #84176.

I created a function `primitive_from_str` which is code that was originally in `collect_intra_doc_links::resolve_primitive` to prevent code duplication.

I also created the `primitive_link_url` function which is somewhat similar to `primitive_link` but too much different to merge both of them.

r? @jyn514

### DIFF
--- a/src/librustdoc/html/highlight.rs
+++ b/src/librustdoc/html/highlight.rs
@@ -5,6 +5,7 @@
 //!
 //! Use the `render_with_highlighting` to highlight some rust code.
 
+use crate::clean::PrimitiveType;
 use crate::html::escape::Escape;
 use crate::html::render::Context;
 
@@ -584,6 +585,13 @@ fn string<T: Display>(
                             .ok()
                             .map(|(url, _, _)| url)
                     }
+                    LinkFromSrc::Primitive(prim) => format::href_with_root_path(
+                        PrimitiveType::primitive_locations(context.tcx())[&prim],
+                        context,
+                        Some(context_info.root_path),
+                    )
+                    .ok()
+                    .map(|(url, _, _)| url),
                 }
             })
         {
--- a/src/librustdoc/html/render/span_map.rs
+++ b/src/librustdoc/html/render/span_map.rs
@@ -1,4 +1,4 @@
-use crate::clean;
+use crate::clean::{self, PrimitiveType};
 use crate::html::sources;
 
 use rustc_data_structures::fx::FxHashMap;
@@ -22,6 +22,7 @@ use std::path::{Path, PathBuf};
 crate enum LinkFromSrc {
     Local(clean::Span),
     External(DefId),
+    Primitive(PrimitiveType),
 }
 
 /// This function will do at most two things:
@@ -73,17 +74,20 @@ impl<'tcx> SpanMapVisitor<'tcx> {
                 Some(def_id)
             }
             Res::Local(_) => None,
+            Res::PrimTy(p) => {
+                // FIXME: Doesn't handle "path-like" primitives like arrays or tuples.
+                let span = path_span.unwrap_or(path.span);
+                self.matches.insert(span, LinkFromSrc::Primitive(PrimitiveType::from(p)));
+                return;
+            }
             Res::Err => return,
             _ => return,
         };
         if let Some(span) = self.tcx.hir().res_span(path.res) {
-            self.matches.insert(
-                path_span.unwrap_or_else(|| path.span),
-                LinkFromSrc::Local(clean::Span::new(span)),
-            );
-        } else if let Some(def_id) = info {
             self.matches
-                .insert(path_span.unwrap_or_else(|| path.span), LinkFromSrc::External(def_id));
+                .insert(path_span.unwrap_or(path.span), LinkFromSrc::Local(clean::Span::new(span)));
+        } else if let Some(def_id) = info {
+            self.matches.insert(path_span.unwrap_or(path.span), LinkFromSrc::External(def_id));
         }
     }
 }
--- a/src/test/rustdoc/check-source-code-urls-to-def-std.rs
+++ b/src/test/rustdoc/check-source-code-urls-to-def-std.rs
@@ -1,0 +1,17 @@
+// compile-flags: -Zunstable-options --generate-link-to-definition
+
+#![crate_name = "foo"]
+
+// @has 'src/foo/check-source-code-urls-to-def-std.rs.html'
+
+fn babar() {}
+
+// @has - '//a[@href="{{channel}}/std/primitive.u32.html"]' 'u32'
+// @has - '//a[@href="{{channel}}/std/primitive.str.html"]' 'str'
+// @has - '//a[@href="{{channel}}/std/primitive.bool.html"]' 'bool'
+// @has - '//a[@href="../../src/foo/check-source-code-urls-to-def-std.rs.html#7"]' 'babar'
+pub fn foo(a: u32, b: &str, c: String) {
+    let x = 12;
+    let y: bool = true;
+    babar();
+}
--- a/src/test/rustdoc/check-source-code-urls-to-def.rs
+++ b/src/test/rustdoc/check-source-code-urls-to-def.rs
@@ -27,6 +27,8 @@ impl Foo {
 fn babar() {}
 
 // @has - '//a/@href' '/struct.String.html'
+// @has - '//a/@href' '/primitive.u32.html'
+// @has - '//a/@href' '/primitive.str.html'
 // @count - '//a[@href="../../src/foo/check-source-code-urls-to-def.rs.html#21"]' 5
 // @has - '//a[@href="../../source_code/struct.SourceCode.html"]' 'source_code::SourceCode'
 pub fn foo(a: u32, b: &str, c: String, d: Foo, e: bar::Bar, f: source_code::SourceCode) {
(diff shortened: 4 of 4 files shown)


src/librustdoc/html/format.rs
```diff
@@ -676,6 +676,36 @@ fn primitive_link(
...
 
+crate fn primitive_link_url(
+    prim: clean::PrimitiveType,
+    cx: &Context<'_>,
+    root_path: &str,
+) -> Option<String> {
+    let m = &cx.cache();
+    match m.primitive_locations.get(&prim) {
+        Some(&def_id) if def_id.is_local() => {
+            Some(format!("{}primitive.{}.html", root_path, prim.as_sym(),))
```
jyn514, line 687: Shouldn't this include the crate name? Are there any tests that the links this generates aren't giving 404s?

src/librustdoc/html/format.rs
```diff
@@ -676,6 +676,36 @@ fn primitive_link(
...
+        }
+        Some(&def_id) => {
+            let path = match m.extern_locations[&def_id.krate] {
+                ExternalLocation::Remote(ref s) => {
+                    format!("{}/std", s.trim_end_matches('/'))
```
jyn514, line 692: Hard-coding `/std` definitely seems wrong. This should use `tcx.crate_name(def_id.krate)` instead.

src/librustdoc/html/format.rs
```diff
@@ -676,6 +676,36 @@ fn primitive_link(
     Ok(())
 }
 
+crate fn primitive_link_url(
+    prim: clean::PrimitiveType,
+    cx: &Context<'_>,
+    root_path: &str,
+) -> Option<String> {
+    let m = &cx.cache();
+    match m.primitive_locations.get(&prim) {
```
jyn514, line 685: You can make this a little smaller by using `m.primitive_locations.get()?` instead of matching.

src/librustdoc/html/format.rs
GuillaumeGomez: Yes it should actually... I'll add one.

src/librustdoc/html/format.rs
GuillaumeGomez: Good point!

src/librustdoc/html/format.rs
```diff
@@ -676,6 +676,25 @@ fn primitive_link(
...
+    if def_id.is_local() {
+        return Some(format!("{}{}/primitive.{}.html", root_path, crate_name, prim.as_sym()));
+    }
+    let path = match m.extern_locations[&def_id.krate] {
+        ExternalLocation::Remote(ref s) => format!("{}/{}", s.trim_end_matches('/'), crate_name),
+        ExternalLocation::Local => format!("{}{}", root_path, crate_name),
+        ExternalLocation::Unknown => format!("{}/{}", crate::DOC_RUST_LANG_ORG_CHANNEL, crate_name),
+    };
+    Some(format!("{}/primitive.{}.html", path, prim.as_sym(),))
+}
```
jyn514, line 696: Hmm, I kind of want to wait for https://github.com/rust-lang/rust/pull/87073 so we can just delete this function altogether. Would you be ok with that? I hope to get it merged within the week.

jyn514 (commented):
I think this may be fixed without having to add all this extra work once #87073 is merged - do you mind waiting for that? You could also try rebasing over that in the meantime to see if the tests pass without your changes.

src/librustdoc/html/format.rs
jyn514: Is there any way to test the link is now correct? I'm worried this will get out of sync with other parts of rustdoc.

GuillaumeGomez:
> I think this may be fixed without having to add all this extra work once #87073 is merged - do you mind waiting for that? You could also try rebasing over that in the meantime to see if the tests pass without your changes.

I'm totally fine with waiting for it!

src/librustdoc/html/format.rs
GuillaumeGomez: I added a test with a "local" primitive to check the full URL.

JohnCSimon:
triage: fyi  #87073 is still open

GuillaumeGomez:
#87073 has been merged, the PR has been rebased and updated to take the changes into account making it much smaller. Ready for the next review round! :)

src/librustdoc/html/render/span_map.rs
```diff
@@ -73,6 +74,22 @@ impl<'tcx> SpanMapVisitor<'tcx> {
                 Some(def_id)
             }
             Res::Local(_) => None,
+            Res::PrimTy(p) => {
+                let span = path_span.unwrap_or_else(|| path.span);
+                if let Some(prim) = self
+                    .tcx
+                    .sess
+                    .source_map()
+                    .span_to_snippet(span)
```
jyn514, line 83: Please do not use span_to_snippet, it's not reliable (e.g. it will break for libstd if people use --remap-path-prefix). Why not use PrimitiveType::from unconditionally?

src/librustdoc/html/render/span_map.rs
```diff
@@ -73,6 +74,22 @@ impl<'tcx> SpanMapVisitor<'tcx> {
                 Some(def_id)
             }
             Res::Local(_) => None,
+            Res::PrimTy(p) => {
```
jyn514, line 77: Note that not all primitives show up in PrimTy, it can only represent "path-like" primitives (so not e.g. arrays and tuples). Is that intentional?
https://doc.rust-lang.org/nightly/nightly-rustc/rustc_hir/hir/enum.PrimTy.html

src/test/rustdoc/check-source-code-urls-to-def.rs
```diff
@@ -40,5 +42,9 @@ pub fn foo(a: u32, b: &str, c: String, d: Foo, e: bar::Bar, f: source_code::Sour
 
 // @has - '//a[@href="../../src/foo/auxiliary/source-code-bar.rs.html#14-16"]' 'bar::sub::Trait'
 // @has - '//a[@href="../../src/foo/auxiliary/source-code-bar.rs.html#14-16"]' 'Trait'
-pub fn foo2<T: bar::sub::Trait, V: Trait>(t: &T, v: &V) {
+pub fn foo2<T: bar::sub::Trait, V: Trait>(t: &T, v: &V, b: bool) {
 }
+
+// @has - '//a[@href="../../foo/primitive.bool.html"]' 'bool'
+#[doc(primitive = "bool")]
+mod whatever {}
```
jyn514, line 50: This seems unnecessarily complicated - can you check it links to doc.rust-lang.org instead, so we're testing how it behaves when there's no doc(primitive) available? That should be the case for literally every crate other than libstd and core.

Feel free to add a separate test for that if you want to keep this one, but I do think it's important to have.

jyn514 (changes requested):

jyn514:
You're right, this is a lot simpler now :)

src/test/rustdoc/check-source-code-urls-to-def.rs
GuillaumeGomez: I think I'll add a second test so both cases are covered.

src/librustdoc/html/render/span_map.rs
GuillaumeGomez: I don't remember the reason so let's just use it unconditionnally.

src/librustdoc/html/render/span_map.rs
GuillaumeGomez: Yes it is. However a FIXME would be nice, I'll add one.

src/librustdoc/clean/utils.rs
```diff
@@ -600,3 +600,36 @@ pub(super) fn display_macro_source(
         }
     }
 }
+
+crate fn primitive_from_str(path_str: &str) -> Option<PrimitiveType> {
```
jyn514, line 604: This is no longer used anywhere except intra-doc links, right? Can you revert this change? I don't think most places should be using strings to parse primitives.

src/librustdoc/html/highlight.rs
```diff
@@ -584,6 +585,13 @@ fn string<T: Display>(
                             .ok()
                             .map(|(url, _, _)| url)
                     }
+                    LinkFromSrc::Primitive(prim) => format::href_with_root_path(
+                        *PrimitiveType::primitive_locations(context.tcx()).get(&prim).unwrap(),
```
jyn514, line 589:
```suggestion
                        PrimitiveType::primitive_locations(context.tcx())[&prim],
```

src/librustdoc/html/render/span_map.rs
```diff
@@ -73,6 +74,12 @@ impl<'tcx> SpanMapVisitor<'tcx> {
...
+                // FIXME: Doesn't handle "path-like" primitives like arrays or tuples.
+                let span = path_span.unwrap_or_else(|| path.span);
```
jyn514, line 79:
```suggestion
                let span = path_span.unwrap_or(path.span);
```

src/test/rustdoc/check-source-code-urls-to-def-std.rs
```diff
@@ -0,0 +1,17 @@
...
+
+#![crate_name = "foo"]
+
+// @has 'src/foo/check-source-code-urls-to-def-std.rs.html'
+
+fn babar() {}
+
+// @has - '//a[@href="https://doc.rust-lang.org/nightly/std/primitive.u32.html"]' 'u32'
+// @has - '//a[@href="https://doc.rust-lang.org/nightly/std/primitive.str.html"]' 'str'
+// @has - '//a[@href="https://doc.rust-lang.org/nightly/std/primitive.bool.html"]' 'bool'
```
jyn514, line 11: These will break on beta.
```suggestion
// @has - '//a[@href="{{channel}}/std/primitive.u32.html"]' 'u32'
// @has - '//a[@href="{{channel}}/std/primitive.str.html"]' 'str'
// @has - '//a[@href="{{channel}}/nightly/std/primitive.bool.html"]' 'bool'
```

GuillaumeGomez: I didn't even know about `{{channel}}`, thanks!

src/librustdoc/clean/utils.rs
GuillaumeGomez: Ah right, moving it back then!

src/test/rustdoc/check-source-code-urls-to-def-std.rs
jyn514: I just added it a few months ago: https://github.com/rust-lang/rust/pull/85990

GuillaumeGomez:
Updated!

jyn514:
r=me once CI is passing

PR #88034, opened 2021-08-14
rustc_privacy: Replace `HirId`s and `DefId`s with `LocalDefId`s where possible

petrochenkov:
Follow up to https://github.com/rust-lang/rust/pull/87568

### DIFF
--- a/compiler/rustc_privacy/src/lib.rs
+++ b/compiler/rustc_privacy/src/lib.rs
@@ -11,7 +11,8 @@ use rustc_data_structures::fx::FxHashSet;
 use rustc_errors::struct_span_err;
 use rustc_hir as hir;
 use rustc_hir::def::{DefKind, Res};
-use rustc_hir::def_id::{DefId, LocalDefId, CRATE_DEF_ID, CRATE_DEF_INDEX, LOCAL_CRATE};
+use rustc_hir::def_id::{DefId, LocalDefId, LocalDefIdSet};
+use rustc_hir::def_id::{CRATE_DEF_ID, CRATE_DEF_INDEX, LOCAL_CRATE};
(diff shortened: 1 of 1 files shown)


compiler/rustc_privacy/src/lib.rs
```diff
@@ -2188,7 +2193,13 @@ fn check_private_in_public(tcx: TyCtxt<'_>, (): ()) {
...
 
     // Check for private types and traits in public interfaces.
-    let mut visitor =
-        PrivateItemsInPublicInterfacesVisitor { tcx, has_pub_restricted, old_error_set_ancestry };
+    let mut visitor = PrivateItemsInPublicInterfacesVisitor {
+        tcx,
+        has_pub_restricted,
+        old_error_set_ancestry: old_error_set_ancestry
+            .into_iter()
+            .filter_map(|hir_id| tcx.hir().opt_local_def_id(hir_id))
```
wesleywiser, line 2201: Are there cases where we expect `tcx.hir().opt_local_def_id(hir_id)` to return `None` and it's ok to skip that `hir_id`? Or should we do `.map(|hir_id| tcx.hir().local_def_id(hir_id))` here instead?

petrochenkov: When the source `old_error_set_ancestry` is constructed it can indeed contain non-definition IDs, e.g. expressions.
However, only definition IDs are ever searched in the destination `old_error_set_ancestry`, so it's ok to skip non-definition IDs when converting the map.
(I should probably add a comment about this.)

wesleywiser: That would be great! Feel free to r=me with that comment.

PR #88036, opened 2021-08-14
Fix dead code warning when inline const is used in pattern

### LINKED ISSUE #78171: dead_code lint triggers even though function is used in const pattern
This should compile without warnings ([playground](https://play.rust-lang.org/?version=nightly&mode=debug&edition=2018&gist=0fc7c5eb8cb3a350c4bce0c1ab8b5676)):

```rust
#![allow(incomplete_features)]
#![feature(inline_const)]

const fn one() -> i32 {
    1
}

fn main() {
    match 1 {
        const { one() } => {}
        _ => {}
    }
}
```

Yet it warns:

```
warning: function is never used: `one`
 --> src/main.rs:4:10
  |
4 | const fn one() -> i32 {
  |          ^^^
  |
  = note: `#[warn(dead_code)]` on by default

warning: 1 warning emitted

    Finished dev [unoptimized + debuginfo] target(s) in 0.80s
     Running `target/debug/playground`
```

nbdd0121:
Fixes #78171

### DIFF
--- a/compiler/rustc_passes/src/dead.rs
+++ b/compiler/rustc_passes/src/dead.rs
@@ -15,6 +15,7 @@ use rustc_middle::middle::privacy;
 use rustc_middle::ty::{self, DefIdTree, TyCtxt};
 use rustc_session::lint;
 use rustc_span::symbol::{sym, Symbol};
+use std::mem;
 
 // Any local node that may call something in its body block should be
 // explored. For example, if it's a live Node::Item that is a
@@ -395,8 +396,14 @@ impl<'tcx> Visitor<'tcx> for MarkSymbolVisitor<'tcx> {
     }
 
     fn visit_anon_const(&mut self, c: &'tcx hir::AnonConst) {
+        // When inline const blocks are used in pattern position, paths
+        // referenced by it should be considered as used.
+        let in_pat = mem::replace(&mut self.in_pat, false);
+
         self.live_symbols.insert(self.tcx.hir().local_def_id(c.hir_id));
         intravisit::walk_anon_const(self, c);
+
+        self.in_pat = in_pat;
     }
 }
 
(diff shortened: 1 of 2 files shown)


petrochenkov:
I'd never guess that `const { ... }` patterns use `PatKind::Lit` internally.
@bors r+

nbdd0121:
Actually this approach doesn't catch `const { foo() }..2` case. Might be better to save `in_pat` and set it to false when visiting a `AnonConst` and restore it after visiting.

nbdd0121:
Added a test case for using inline const in range pattern

nbdd0121:
> I'd never guess that `const { ... }` patterns use `PatKind::Lit` internally.

It's surprising to me as well, but I guess that's probably the easiest way to implement it since it'll need to have a `P<Expr>` anyway.

compiler/rustc_passes/src/dead.rs
```diff
@@ -395,8 +396,14 @@ impl<'tcx> Visitor<'tcx> for MarkSymbolVisitor<'tcx> {
     }
 
     fn visit_anon_const(&mut self, c: &'tcx hir::AnonConst) {
+        // When inline const blocks are used in pattern position, paths
+        // refereneced by it should be considered as used.
```
klensy, line 400:
```suggestion
        // referenced by it should be considered as used.
```

petrochenkov:
>Might be better to save in_pat and set it to false when visiting a AnonConst and restore it after visiting.

Yeah, it's a better approach, it should also fix cases like
```rust
const fn one() -> i32 {
    1
}

struct S<const C: i32>;

fn main() {
    match S::<1> {
        S::<{one()}> => {}
    }
}
```

Could you actually add this case too?
And merge all the cases into a single test file, no need to run rustc multiple times to test all of this.

@bors r-

nbdd0121:
Added and merged

petrochenkov:
Thanks!
@bors r+

PR #79245, opened 2020-11-20
BTree: remove Ord bound where it is absent elsewhere

ssomers:
Some btree methods don't really need an Ord bound and don't have one, while some methods that more obviously don't need it, do have one.

An example of the former is `iter`, even though it explicitly exposes the work of the Ord implementation (["sorted by key"](https://doc.rust-lang.org/std/collections/struct.BTreeMap.html#method.iter) - but I'm not suggesting it should have the Ord bound). An example of the latter is `new`, which doesn't involve any keys whatsoever.

PR #88040, opened 2021-08-15
BTree: remove Ord bound from new

nbdd0121:
`K: Ord` bound is unnecessary on `BTree{Map,Set}::new` and their `Default` impl. No elements exist so there are nothing to compare anyway, so I don't think "future proof" would be a blocker here. This is analogous to `HashMap::new` not having a `K: Eq + Hash` bound.

#79245 originally does this and for some reason drops the change to `new` and `Default`. I can see why changes to other methods like `entry` or `symmetric_difference` need to be careful but I couldn't find out any reason not to do it on `new`.

Removing the bound also makes the stabilisation of `const fn new` not depending on const trait bounds.

cc @steffahn who suggests me to make this PR.

r? @dtolnay

### DIFF
--- a/library/alloc/src/collections/btree/map.rs
+++ b/library/alloc/src/collections/btree/map.rs
@@ -233,9 +233,7 @@ impl<K: Clone, V: Clone> Clone for BTreeMap<K, V> {
         }
 
         if self.is_empty() {
-            // Ideally we'd call `BTreeMap::new` here, but that has the `K:
-            // Ord` constraint, which this method lacks.
-            BTreeMap { root: None, length: 0 }
+            BTreeMap::new()
         } else {
             clone_subtree(self.root.as_ref().unwrap().reborrow()) // unwrap succeeds because not empty
         }
@@ -499,10 +497,7 @@ impl<K, V> BTreeMap<K, V> {
     /// ```
     #[stable(feature = "rust1", since = "1.0.0")]
     #[rustc_const_unstable(feature = "const_btree_new", issue = "71835")]
-    pub const fn new() -> BTreeMap<K, V>
-    where
-        K: Ord,
-    {
+    pub const fn new() -> BTreeMap<K, V> {
         BTreeMap { root: None, length: 0 }
     }
 
@@ -522,7 +517,7 @@ impl<K, V> BTreeMap<K, V> {
     /// ```
     #[stable(feature = "rust1", since = "1.0.0")]
     pub fn clear(&mut self) {
-        *self = BTreeMap { root: None, length: 0 };
+        *self = BTreeMap::new();
     }
 
     /// Returns a reference to the value corresponding to the key.
@@ -1957,7 +1952,7 @@ impl<K: Hash, V: Hash> Hash for BTreeMap<K, V> {
 }
 
 #[stable(feature = "rust1", since = "1.0.0")]
-impl<K: Ord, V> Default for BTreeMap<K, V> {
+impl<K, V> Default for BTreeMap<K, V> {
     /// Creates an empty `BTreeMap`.
     fn default() -> BTreeMap<K, V> {
         BTreeMap::new()
--- a/library/alloc/src/collections/btree/map/tests.rs
+++ b/library/alloc/src/collections/btree/map/tests.rs
@@ -1745,7 +1745,7 @@ fn test_send() {
     }
 }
 
-#[allow(dead_code)]
+#[test]
 fn test_ord_absence() {
     fn map<K>(mut map: BTreeMap<K, ()>) {
         map.is_empty();
@@ -1784,6 +1784,12 @@ fn test_ord_absence() {
     fn map_clone<K: Clone>(mut map: BTreeMap<K, ()>) {
         map.clone_from(&map.clone());
     }
+
+    #[derive(Debug, Clone)]
+    struct NonOrd;
+    map(BTreeMap::<NonOrd, _>::new());
+    map_debug(BTreeMap::<NonOrd, _>::new());
+    map_clone(BTreeMap::<NonOrd, _>::default());
 }
 
 #[test]
--- a/library/alloc/src/collections/btree/set.rs
+++ b/library/alloc/src/collections/btree/set.rs
@@ -246,10 +246,7 @@ impl<T> BTreeSet<T> {
     /// ```
     #[stable(feature = "rust1", since = "1.0.0")]
     #[rustc_const_unstable(feature = "const_btree_new", issue = "71835")]
-    pub const fn new() -> BTreeSet<T>
-    where
-        T: Ord,
-    {
+    pub const fn new() -> BTreeSet<T> {
         BTreeSet { map: BTreeMap::new() }
     }
 
@@ -1192,7 +1189,7 @@ impl<'a, T: 'a + Ord + Copy> Extend<&'a T> for BTreeSet<T> {
 }
 
 #[stable(feature = "rust1", since = "1.0.0")]
-impl<T: Ord> Default for BTreeSet<T> {
+impl<T> Default for BTreeSet<T> {
     /// Creates an empty `BTreeSet`.
     fn default() -> BTreeSet<T> {
         BTreeSet::new()
--- a/library/alloc/src/collections/btree/set/tests.rs
+++ b/library/alloc/src/collections/btree/set/tests.rs
@@ -607,7 +607,7 @@ fn test_send() {
     }
 }
 
-#[allow(dead_code)]
+#[test]
 fn test_ord_absence() {
     fn set<K>(mut set: BTreeSet<K>) {
         set.is_empty();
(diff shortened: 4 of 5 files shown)


dtolnay (commented):
> No elements exist so there are nothing to compare anyway

> I don't think "future proof" would be a blocker here

> I couldn't find out any reason not to do it

This is not as clear cut as the PR description implies.

I am hesitant about this because this API change makes it impossible to selectively tackle some of the code bloat caused by monomorphizations inside the internals of these collections. According to https://www.stroustrup.com/SCARY.pdf such optimizations deliver 1.2x to 2.1x faster performance and 1x to 25x smaller code size in real applications.

I pasted below a sketch of the kind of restructure that this change hinders.

Clearly we still have bounds available via the `insert` and `contains` methods etc, so a comparator could be passed down from there as a function argument to every places that needs it, but the extra argument passing could negate a chunk of the performance benefit relative to stashing a comparator at suitable places inside of the data structure up front.

@rust-lang/libs-api @rust-lang/libs

```rust
use self::private::{Element, TypeErasedSet};
use std::cmp::Ordering;
use std::marker::PhantomData;
use std::mem;

pub struct Set<T> {
    imp: TypeErasedSet,
    marker: PhantomData<T>,
}

impl<T> Set<T> {
    pub fn new() -> Self
    where
        T: Ord,
    {
        unsafe fn erased_cmp<T: Ord>(a: &dyn Element, b: &dyn Element) -> Ordering {
            T::cmp(&*(a as *const _ as *const _), &*(b as *const _ as *const _))
        }
        Set {
            imp: unsafe { TypeErasedSet::new(erased_cmp::<T>) },
            marker: PhantomData,
        }
    }

    pub fn insert(&mut self, value: T)
    where
        T: Ord,
    {
        let boxed = Box::new(value);
        unsafe {
            let erased = mem::transmute::<Box<dyn Element>, Box<dyn Element + 'static>>(boxed);
            self.imp.insert(erased)
        }
    }

    pub fn contains(&self, value: &T) -> bool
    where
        T: Ord,
    {
        unsafe { self.imp.contains(value) }
    }
}

mod private {
    use std::cmp::Ordering;

    pub(super) trait Element {}
    impl<Sized> Element for Sized {}

    type Comparator = unsafe fn(&dyn Element, &dyn Element) -> Ordering;

    pub(super) struct TypeErasedSet {
        // Obviously you could build a more efficient representation than this.
        // Just illustrating that the concept works without a monomorphized T.
        repr: Vec<Box<dyn Element>>,
        cmp: Comparator,
    }

    impl TypeErasedSet {
        pub(super) unsafe fn new(cmp: Comparator) -> Self {
            TypeErasedSet {
                repr: Vec::new(),
                cmp,
            }
        }

        pub(super) unsafe fn insert(&mut self, value: Box<dyn Element>) {
            match self.find(&*value) {
                Ok(present) => self.repr[present] = value,
                Err(absent) => self.repr.insert(absent, value),
            }
        }

        pub(super) unsafe fn contains(&self, value: &dyn Element) -> bool {
            self.find(value).is_ok()
        }

        unsafe fn find(&self, value: &dyn Element) -> Result<usize, usize> {
            self.repr
                .binary_search_by(|elem| unsafe { (self.cmp)(&**elem, value) })
        }
    }
}

fn main() {
    let mut set = Set::new();
    set.insert(5);
    set.insert(2);
    set.insert(9);
    set.insert(4);
    set.insert(9);
    println!("{}", set.contains(&4));
    println!("{}", set.contains(&3));
}
```

nbdd0121:
> Clearly we still have bounds available via the insert and contains methods etc, so a comparator could be passed down from there as a function argument to every places that needs it, but the extra argument passing could negate a chunk of the performance benefit relative to stashing a comparator at suitable places inside of the data structure up front.

I suppose this wouldn't be a big deal for `BTreeMap`, since when searching the tree we currently only take `NodeRef` without having the references to the `BTreeMap` itself. So:
* If the comparator is to be stored in `BTreeMap` but not within the nodes, then extra argument is needed anyway. This would mean that it doesn't make sense to store comparator in the `BTreeMap`.
* If the comparator is to be stored within nodes, then the comparator is not needed to create an empty `BTreeMap` because it starts with no nodes (`root = None`).

steffahn:
Storing a comparator in the data structure might kill covariance and ignoring that issue might be unsound. I will investigate more on that later.

Passing a comparator with every call (to insert, etc) seems more reasonable to me. Especially since you'd need to "pass down" the function anyway if it's stored at top level. Whether it originates from a read operation or from a compile-time constant passed down from one layer above in the call hierarchy doesn't seem like the most significant difference. But I don't even know where the proclaimed performance benefit in that "SCARY" paper comes from (it's a bit lengthy, havent read through much of it yet) so I can't be sure to have the full picture. (The code size reduction is obvious on the other hand.)

Passing the comparator every call also allows for more flexible API, where you could offer a version of `BTreeMap` (using the same underlying erased data structure) that supports a comparator dependent on short-term borrows of another data structure. E. g. a BTreeSer of indices into another Vec would be possible. (Sorted by the `Ord`er of the Vec elements).

steffahn:
@dtolnay 

> […] ignoring that issue might be unsound. I will investigate more on that later.

Here’s a demonstration using your code example.

```rs

use std::any::Any;
use once_cell::sync::OnceCell;

fn main() {
    let s = Set::<Wrapper<for<'a> fn(&'a str)>>::new();
    let mut s: Set<Wrapper<fn(&'static str)>> = s;
    static S: OnceCell<&'static str> = OnceCell::new();
    let f = |s| {
        let _ = S.try_insert(s);
    };
    s.insert(Wrapper(f));
    s.contains(&Wrapper(f));
    println!("{}", S.get().unwrap());
}

// for the record, BTreeSet is indeed covariant, too
use std::collections::BTreeSet;
fn _b_tree_set_is_covariant<'a: 'b, 'b, T>(x: BTreeSet<&'a T>) -> BTreeSet<&'b T> {
    x
}

struct Wrapper<T>(T);

impl<T: 'static> Ord for Wrapper<T> {
    fn cmp(&self, _other: &Self) -> Ordering {
        if let Some(f) = (self as &dyn Any).downcast_ref::<Wrapper<for<'a> fn(&'a str)>>() {
            f.0(&String::from("Hello world!"));
        }
        Ordering::Equal
    }
}
impl<T: 'static> PartialOrd<Self> for Wrapper<T> {
    fn partial_cmp(&self, other: &Self) -> Option<Ordering> {
        Some(self.cmp(other))
    }
}
impl<T: 'static> Eq for Wrapper<T> {}
impl<T: 'static> PartialEq<Self> for Wrapper<T> {
    fn eq(&self, other: &Self) -> bool {
        self.cmp(other).is_eq()
    }
}
```
([in the playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=84a6d7f50a1ec8d088ffca025ee63bca))
Example output:
```
��)
```

Amanieu:
Incidentally hashbrown already does something similar internally ([passing the comparator as an argument](https://docs.rs/hashbrown/0.11.2/hashbrown/raw/struct.RawTable.html#method.find)) but for different reasons:
- It makes `RawTable` independent of the details of how a value is hashed or compared.
- It makes the `RawTable` API more flexible since the comparator closure can contain borrowed data that isn't tied to the lifetime of the table.

steffahn:
> Incidentally hashbrown already does something similar internally ([passing the comparator as an argument](https://docs.rs/hashbrown/0.11.2/hashbrown/raw/struct.RawTable.html#method.find))

It’s a very slight similarity since the hasher arguments are generic (`impl Fn(&T) -> u64`), so at run-time there’s no “passing” but monomorphization if the calls come from a `HashMap` (the passed function items are zero-sized types, not function pointers).

steffahn:
@nbdd0121 To make this PR more complete, you’d probably also want to use `BTreeMap::new` here

https://github.com/rust-lang/rust/blob/cd5a90fb14bb8cb2276d9740925c9858ea507429/library/alloc/src/collections/btree/map.rs#L238-L240

and here

https://github.com/rust-lang/rust/blob/cd5a90fb14bb8cb2276d9740925c9858ea507429/library/alloc/src/collections/btree/map.rs#L535

and I suppose, we’d want to have a test for the new capabilities of `BTreeMap::new` as well.

For example, you could change the contents of this file to use a `!Ord` item type instead of `MyType`, or something like that? (E.g. keep `MyType` but remove the `const Ord` implementation and the `deriving(…)` attribute.)

https://github.com/rust-lang/rust/blob/136eaa1b25d13635b773a481ecab61a3162cb627/library/alloc/tests/const_fns.rs#L13-L53

_(For context, that `FIXME` comment seems to assume that the `Ord` bound stays but gets replaces by `?const Ord` which would allow something like `usize` as well which implements `Ord` but not `const Ord`. Dropping the bound alltogether is a different thing than what’s envisioned in that `FIXME`, but that’s okay.)_

dtolnay:
@rfcbot fcp merge

https://github.com/rust-lang/rust/pull/88040#issuecomment-899038071 is persuasive as far as the Ord impl which exists during `new` being totally useless at any point after `new` has returned, so I am on board with this.

steffahn:
@ anyone reviewing this, FYI:

this suggestion

> cc @steffahn who suggests me to make this PR.

happened in a (short) topic on Zulip which can be found [here](https://rust-lang.zulipchat.com/#narrow/stream/219381-t-libs/topic/Why.20does.20.60BTreeMap.3A.3Anew.60.20require.20.60K.3A.20Ord.60) (or [in the archive](https://zulip-archive.rust-lang.org/stream/219381-t-libs/topic/Why.20does.20.60BTreeMap.3A.3Anew.60.20require.20.60K.3A.20Ord.60.html)).

nbdd0121:
Surprised to see that covariance are not asserted, filed #88058

PR #88044, opened 2021-08-15
Docs: add corner cases of Path::file_name()

marcospb19:
Documenting corner cases where `Path::file_name()` returns `None`, and adding examples.

### DIFF
--- a/library/std/src/path.rs
+++ b/library/std/src/path.rs
@@ -2038,7 +2038,11 @@ impl Path {
     /// If the path is a normal file, this is the file name. If it's the path of a directory, this
     /// is the directory name.
     ///
-    /// Returns [`None`] if the path terminates in `..`.
+    /// Returns [`None`] if path:
+    /// - Is empty.
+    /// - Terminates in `..` (double dots).
+    /// - Is root (e.g. `/`, `/.`, `/././.`).
+    /// - Only contains current directory (e.g. `.`, `././.`).
     ///
     /// # Examples
     ///
@@ -2051,7 +2055,11 @@ impl Path {
     /// assert_eq!(Some(OsStr::new("foo.txt")), Path::new("foo.txt/.").file_name());
     /// assert_eq!(Some(OsStr::new("foo.txt")), Path::new("foo.txt/.//").file_name());
     /// assert_eq!(None, Path::new("foo.txt/..").file_name());
+    /// assert_eq!(None, Path::new("..").file_name());
     /// assert_eq!(None, Path::new("/").file_name());
+    /// assert_eq!(None, Path::new("/.").file_name());
+    /// assert_eq!(None, Path::new("/././.").file_name());
+    /// assert_eq!(None, Path::new("././.").file_name());
     /// ```
     #[stable(feature = "rust1", since = "1.0.0")]
     pub fn file_name(&self) -> Option<&OsStr> {


marcospb19:
Here's the page rendered:

image

This markdown list is looking kinda squished and confusing, maybe I should just throw these corner cases in the doc test assertions? And remove the "(e.g. ...)" examples.

library/std/src/path.rs
```diff
@@ -2038,7 +2038,11 @@ impl Path {
     /// If the path is a normal file, this is the file name. If it's the path of a directory, this
     /// is the directory name.
     ///
-    /// Returns [`None`] if the path terminates in `..`.
+    /// Returns [`None`] if path:
+    /// - Is empty.
+    /// - Terminates in "`..`".
+    /// - Is root (e.g. "`/`", "`/.`", "`/././.`").
```
jfrimmel, line 2044:
```suggestion
    /// - Is root (e.g. `/`, `/.`, `/././.`).
```

library/std/src/path.rs
```diff
@@ -2038,7 +2038,11 @@ impl Path {
...
+    /// - Only contains current directory (e.g. "`.`", "`././.`").
```
jfrimmel, line 2045:
```suggestion
    /// - Only contains current directory (e.g. `.` or `././.`).
```

jfrimmel (commented):
Maybe this isbbetter (remove quotes in favor of only markdown code blocks)?

marcospb19:
@jfrimmel I do think it looks weird:
image

"Terminates in ..." looks like 3 dots, instead of 2.

I will also add "(double dots)" to the side of it, to see if it helps.

marcospb19:
Here's the current status:
image

library/std/src/path.rs
```diff
@@ -2038,7 +2038,11 @@ impl Path {
     /// If the path is a normal file, this is the file name. If it's the path of a directory, this
     /// is the directory name.
     ///
-    /// Returns [`None`] if the path terminates in `..`.
+    /// Returns [`None`] if path:
+    /// - Is empty.
+    /// - Terminates in `..` (double dots).
+    /// - Is root (e.g. `/`, `/.`, `/././.`).
+    /// - Only contains current directory (e.g. `.`, `././.`).
```
Mark-Simulacrum, line 2045: I think this is missing the Prefix variant of Component -- maybe it is more useful to say "returns None if the last component of the path is not Normal" and link to https://doc.rust-lang.org/nightly/std/path/enum.Component.html?

marcospb19: That's a great idea, however, here is the documentation of Component:

image

It does not explain that running components on "a/." gives `Component::Normal("a")` instead of `Normal("a")` AND `CurrrentDirectory`, plus I think people are expecting "asd/.".file_name() to return ".", so, maybe this should be explained somewhere?

Maybe I could add examples to the docs of `Component` itself to show what iterating on `Path::new("/a/b/c")` retrieves, then we could also cover this corner cases.

However, there is already something like this at https://doc.rust-lang.org/nightly/std/path/struct.Path.html#method.components.

```rust
use std::path::{Path, Component};
use std::ffi::OsStr;

let mut components = Path::new("/tmp/foo.txt").components();

assert_eq!(components.next(), Some(Component::RootDir));
assert_eq!(components.next(), Some(Component::Normal(OsStr::new("tmp"))));
assert_eq!(components.next(), Some(Component::Normal(OsStr::new("foo.txt"))));
assert_eq!(components.next(), None)
```

So, maybe I should add more examples to `path::components` instead? Or add some explanation in the `Component` page itself.

Waiting for some feedback on these thoughts.

Mark-Simulacrum: I think adding more docs to Component makes sense, since it's a dedicated page for this. We can point there from file_name() and potentially components() as well.

marcospb19: Nice, I agree! I've been getting very busy this days, so I might need to delay this PR just a couple weeks, but thanks for the review.

marcospb19:
@Mark-Simulacrum @Dylan-DPC sorry for the huge delay on this, I'm struggling with managing my personal time.

Was not able to finish this _yet_.

If someone want to open a PR with it finished, I'd be glad, you might want to close mine.

(Or commit on my branch, for some repo _member_).

Dylan-DPC:
@marcospb19 that's totally fine and understandable. I'll close this for now, thanks for taking the time to contribute to rust.

Issue #42847, opened 2017-06-23
*const T has quite a few methods that seem like they don't need to require T: Sized


Mark-Simulacrum:
Not sure which exactly, but we should evaluate and relax bounds where possible.

See also https://github.com/rust-lang/rust/issues/24794, https://github.com/rust-lang/rust/issues/36952.

smmalis37:
Adding to this, std::ptr::null & std::ptr::null_mut. Technically not methods on *const T but has these bounds too.

shepmaster:
Should I open a new issue to argue about `std::ptr::null` / `std::ptr::null_mut`? Seems wrong when used with extern types:

```rust
extern {
    type foo_t;
}

fn fake_foo_new() -> *const foo_t {
    std::ptr::null()
}
```

```
error[E0277]: the trait bound `foo_t: std::marker::Sized` is not satisfied
  --> src/main.rs:15:5
   |
15 |     std::ptr::null()
   |     ^^^^^^^^^^^^^^ `foo_t` does not have a constant size known at compile-time
   |
   = help: the trait `std::marker::Sized` is not implemented for `foo_t`
   = note: required by `std::ptr::null`
```

shepmaster:
I'll be *that* person and ping in @dtolnay, @eddyb, and @mikeyhew for some more discussion around [my last comment](https://github.com/rust-lang/rust/issues/42847#issuecomment-385528055)

shepmaster:
I can certainly see that a null vtable would be a bad thing, but I also feel that the complete inability to create a `NULL` or uninitialized extern type is going to be a *real* blocker to a lot of usage of extern types:

```
#![feature(extern_types)]

extern {
    type foo_t;
    
    fn foo_init(f: *mut foo_t);
}

fn main() {
    let mut a = std::mem::uninitialized();
    unsafe { foo_init(&mut a) };
}
```

```
error[E0277]: the trait bound `foo_t: std::marker::Sized` is not satisfied
  --> src/main.rs:10:9
   |
10 |     let mut a = std::mem::uninitialized();
   |         ^^^^^ `foo_t` does not have a constant size known at compile-time
   |
   = help: the trait `std::marker::Sized` is not implemented for `foo_t`
   = note: all local variables must have a statically known size
```

dtolnay:
As a workaround for now you can get a null pointer to extern type through `as`:

```rust
#![feature(extern_types)]

extern "C" {
    type foo_t;

[...]

PR #88049, opened 2021-08-15
Allow casting unsized pointer types

dylni:
I don't think it was intentional to give these types `Sized` bounds.


dylni:
The vtables may not match. That's unfortunate.

dylni:
Related: #42847

PR #88058, opened 2021-08-15
Assert covariance of `BTree{Map,Set}` and associated iterators

nbdd0121:
`Difference` etc are not yet covariant, see #30642.

### DIFF
--- a/library/alloc/src/collections/btree/map.rs
+++ b/library/alloc/src/collections/btree/map.rs
@@ -1,3 +1,6 @@
+#[cfg(test)]
+mod tests;
+
 use core::borrow::Borrow;
 use core::cmp::Ordering;
 use core::fmt::{self, Debug};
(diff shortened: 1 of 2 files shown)


steffahn:
There’s also the variance of the lifetime parameter that could be asserted for types like `Iter<'a, K, V>`. You also don’t really need separate tests for the key and value, e.g. a single
```rs
fn iter<'a>(x: Iter<'static, &'static (), &'static ()>) -> Iter<'a, &'a (), &'a ()> {
    x
}
```
for `btree_map::Iter` should be enough.

nbdd0121:
I followed HashMap, which has separate key and value assertions.

steffahn:
Alright, still the lifetime argument of `Iter`, `Keys`, `Values`, `Range`, and even `IterMut`, `RangeMut` and `ValuesMut` could be included in the tests. Whether they’re tested separately or not.

steffahn:
If `HashMap` doesn’t check those either, then that may be a reasonable addition there as well, I guess.

nbdd0121:
In that case I'll just merge the checks into one... otherwise it'll too verbose.

As for `HashMap` and `LinkedList` I think it'll be better to file it under a separate PR. I'd like to keep this PR limited to just adding new checks for `BTree{Map,Set}`.

nbdd0121:
It turns out there are already tests for variances, just that they are named `test_variance` instead of `test_covariance` so my `grep` doesn't detect them. Sigh 🤦

nbdd0121:
Of course these tests don't cover lifetime parameters, so it might worth doing that.

PR #88060, opened 2021-08-15
Optimize unnecessary check in Vec::retain

TennyZhuang:
The function `vec::Vec::retain` only have two stages:

1. Nothing was deleted.
2. Some elements were deleted.

Here is an unnecessary check `if g.deleted_cnt > 0` in the loop, and it's difficult for compiler to optimize it. I split the loop into two stages manully and keep the code clean using const generics.

I write a special but common bench case for this optimization. I call retain on vec but keep all elements.

Before and after this optimization:

```
test vec::bench_retain_whole_100000                      ... bench:      84,803 ns/iter (+/- 17,314)
```

```
test vec::bench_retain_whole_100000                      ... bench:      42,638 ns/iter (+/- 16,910)
```

The result is expected, there are two `if`s before the optimization and one `if` after.

### DIFF
--- a/library/alloc/benches/vec.rs
+++ b/library/alloc/benches/vec.rs
@@ -732,3 +732,18 @@ fn bench_flat_map_collect(b: &mut Bencher) {
     let v = vec![777u32; 500000];
     b.iter(|| v.iter().flat_map(|color| color.rotate_left(8).to_be_bytes()).collect::<Vec<_>>());
 }
+
+#[bench]
+fn bench_retain_100000(b: &mut Bencher) {
+    let v = (1..=100000).collect::<Vec<u32>>();
+    b.iter(|| {
+        let mut v = v.clone();
+        v.retain(|x| x & 1 == 0)
+    });
+}
+
+#[bench]
+fn bench_retain_whole_100000(b: &mut Bencher) {
+    let mut v = black_box(vec![826u32; 100000]);
+    b.iter(|| v.retain(|x| *x == 826u32));
+}
--- a/library/alloc/src/vec/mod.rs
+++ b/library/alloc/src/vec/mod.rs
@@ -1485,7 +1485,15 @@ impl<T, A: Allocator> Vec<T, A> {
 
         let mut g = BackshiftOnDrop { v: self, processed_len: 0, deleted_cnt: 0, original_len };
 
-        while g.processed_len < original_len {
+        // process_one return a bool indicates whether the processing element should be retained.
+        #[inline(always)]
+        fn process_one<F, T, A: Allocator, const DELETED: bool>(
+            f: &mut F,
+            g: &mut BackshiftOnDrop<'_, T, A>,
+        ) -> bool
+        where
+            F: FnMut(&T) -> bool,
+        {
             // SAFETY: Unchecked element must be valid.
             let cur = unsafe { &mut *g.v.as_mut_ptr().add(g.processed_len) };
             if !f(cur) {
@@ -1495,9 +1503,9 @@ impl<T, A: Allocator> Vec<T, A> {
                 // SAFETY: We never touch this element again after dropped.
                 unsafe { ptr::drop_in_place(cur) };
                 // We already advanced the counter.
-                continue;
+                return false;
             }
-            if g.deleted_cnt > 0 {
+            if DELETED {
                 // SAFETY: `deleted_cnt` > 0, so the hole slot must not overlap with current element.
                 // We use copy for move, and never touch this element again.
                 unsafe {
@@ -1506,6 +1514,19 @@ impl<T, A: Allocator> Vec<T, A> {
                 }
             }
             g.processed_len += 1;
+            return true;
+        }
+
+        // Stage 1: Nothing was deleted.
+        while g.processed_len != original_len {
+            if !process_one::<F, T, A, false>(&mut f, &mut g) {
+                break;
+            }
+        }
+
+        // Stage 2: Some elements were deleted.
+        while g.processed_len != original_len {
+            process_one::<F, T, A, true>(&mut f, &mut g);
         }
 
         // All item are processed. This can be optimized to `set_len` by LLVM.


library/alloc/benches/vec.rs
```diff
@@ -732,3 +732,9 @@ fn bench_flat_map_collect(b: &mut Bencher) {
     let v = vec![777u32; 500000];
     b.iter(|| v.iter().flat_map(|color| color.rotate_left(8).to_be_bytes()).collect::<Vec<_>>());
 }
+
+#[bench]
+fn bench_retain_whole_100000(b: &mut Bencher) {
```
oxalica, line 737: Could you more bench(s) for case of many moves? Like retaining the second half of a Vec, or only even positions.

library/alloc/src/vec/mod.rs
```diff
@@ -1485,7 +1485,15 @@ impl<T, A: Allocator> Vec<T, A> {
 
         let mut g = BackshiftOnDrop { v: self, processed_len: 0, deleted_cnt: 0, original_len };
 
-        while g.processed_len < original_len {
+        // process_one return a bool indicates whether the processing element should be deleted.
+        #[inline]
```
oxalica, line 1489:
```suggestion
        #[inline(always)]
```

oxalica (commented):
The [generated code](https://godbolt.org/z/YGWThjz65) seems kind of weird. The second-stage loop is generated into many basic blocks and there are many jumps around.

Anyway, the benchmark is quite convincing.

BTW: Actually find this PR from your Twitter post. :smile:

TennyZhuang:
I add a normal benchmark:



before

```
test vec::bench_retain_100000                            ... bench:     460,816 ns/iter (+/- 24,389)
test vec::bench_retain_whole_100000                      ... bench:      88,300 ns/iter (+/- 4,635)
```

after

```
test vec::bench_retain_100000                            ... bench:      62,627 ns/iter (+/- 6,692)
test vec::bench_retain_whole_100000                      ... bench:      46,326 ns/iter (+/- 4,133)
```

It's extremely fasater, even that I can't explain it

darksv:
@TennyZhuang I believe this change would break existing code relying on the fact that the passed closure is currently called once per item. 

From docs:
`This method operates in place, visiting each element exactly once in the original order, and preserves the order of the retained elements.`

TennyZhuang:
> @TennyZhuang I believe this change would break existing code relying on the fact that the passed closure is currently called once per item.
> 
> From docs:
> `This method operates in place, visiting each element exactly once in the original order, and preserves the order of the retained elements.`

This change will not break it, each element is still called only once, we just split the loop to two stages, the processed_cnt will never go back.

darksv:
> This change will not break it, each element is still operated exactly once in the original order, we just split the loop to two stages, the `processed_len` will never go back.

You're right. Sorry for interfering.

oxalica:
I ran the benchmark on my R5 2400G PC and it is quite different, and is quite stable between multiple runs. The with-many-move case actually becomes slower for me.

Before
```
test vec::bench_retain_100000                            ... bench:      58,436 ns/iter (+/- 2,482)
test vec::bench_retain_whole_100000                      ... bench:      54,747 ns/iter (+/- 1,646)
```

After
```
test vec::bench_retain_100000                            ... bench:      53,703 ns/iter (+/- 2,905)
test vec::bench_retain_whole_100000                      ... bench:      61,588 ns/iter (+/- 689)
```

It might indicate that the generated code heavily relies on the *behavior* of branch predictor. Are you on a Intel CPU? Intel's branch predictor may work different with AMD's.

TennyZhuang:
> I ran the benchmark on my R5 2400G PC and it is quite different, and is quite stable between multiple runs. The with-many-move case actually becomes slower for me.
> 
> Before
> 
> ```
> test vec::bench_retain_100000                            ... bench:      58,436 ns/iter (+/- 2,482)
> test vec::bench_retain_whole_100000                      ... bench:      54,747 ns/iter (+/- 1,646)
> ```
> 
> After
> 
> ```
> test vec::bench_retain_100000                            ... bench:      53,703 ns/iter (+/- 2,905)
> test vec::bench_retain_whole_100000                      ... bench:      61,588 ns/iter (+/- 689)
> ```
> 
> It might indicate that the generated code heavily relies on the _behavior_ of branch predictor. Are you on a Intel CPU? Intel's branch predictor may work different with AMD's.

I test it on my MacBook Pro 13 with `2.0GHz quad-core 10th-generation Intel Core i5 processor, Turbo Boost up to 3.8GHz`.

I can run the bench on another device.

@Xuanwo could you also run a benchmark for my optimization? 

@oxalica I will also try manually expand two loop without const generics.

Xuanwo:
I ran the same benchmark on my laptop:

```shell
                   -`                    xuanwo@thinkpad-x1-carbon
                  .o+`                   -------------------------
                 `ooo/                   OS: Arch Linux x86_64
                `+oooo:                  Host: 20KHCTO1WW ThinkPad X1 Carbon 6th
               `+oooooo:                 Kernel: 5.13.10-zen1-1-zen
               -+oooooo+:                Uptime: 1 day, 5 hours, 8 mins
             `/:-:++oooo+:               Packages: 1144 (pacman)
            `/++++/+++++++:              Shell: zsh 5.8
           `/++++++++++++++:             Resolution: 1920x1080
          `/+++ooooooooooooo/`           DE: KDE5
         ./ooosssso++osssssso+`          WM: KWin
        .oossssso-````/ossssss+`         WM Theme: Breeze
       -osssssso.      :ssssssso.        Theme: Breeze Light [KDE5], Canta-light [GTK2/3]
      :osssssss/        osssso+++.       Icons: breeze [KDE5], breeze [GTK2/3]
     /ossssssss/        +ssssooo/-       Terminal: tmux
   `/ossssso+/:-        -:/+osssso+-     CPU: Intel i7-8650U (8) @ 4.200GHz
  `+sso+:-`                 `.-/+oso:    GPU: Intel UHD Graphics 620
 `++:.                           `-/+/   Memory: 10319MiB / 15754MiB
 .`                                 `/

```

Before

```
test vec::bench_retain_100000                            ... bench:      74,452 ns/iter (+/- 2,484)
test vec::bench_retain_whole_100000                      ... bench:      59,383 ns/iter (+/- 2,675)
```

After

```
test vec::bench_retain_100000                            ... bench:      69,770 ns/iter (+/- 1,764)
test vec::bench_retain_whole_100000                      ... bench:      53,829 ns/iter (+/- 1,374)
```

TennyZhuang:
I ran it on another Mac:

```
sysctl -n machdep.cpu.brand_string
Intel(R) Core(TM) i7-9750H CPU @ 2.60GHz
```

The command:

```
./x.py bench -i library/alloc --test-args bench_retain
```

result is almost the same:

Before

```
test vec::bench_retain_100000                            ... bench:     401,699 ns/iter (+/- 11,941)
test vec::bench_retain_whole_100000                      ... bench:      75,848 ns/iter (+/- 2,291)
```

After

```
test vec::bench_retain_100000                            ... bench:      52,742 ns/iter (+/- 1,948)
test vec::bench_retain_whole_100000                      ... bench:      38,554 ns/iter (+/- 3,745)
```

TennyZhuang:
It seems that it works very poor on MacOS with unknown reason

TennyZhuang:
@Xuanwo @oxalica Can you run a version without const generics?

For your convinience:

```
git remote add zty git@github.com:TennyZhuang/rust.git
git fetch zty
git checkout optimize-vec-retain-expand-manully
```

TennyZhuang:
On my mac, the performance is same between const generics and manully expand:

```
test vec::bench_retain_100000                            ... bench:      52,471 ns/iter (+/- 3,921)
test vec::bench_retain_whole_100000                      ... bench:      38,560 ns/iter (+/- 2,666)
```

Xuanwo:
Two versions work similar on my laptop:

const generics

```rust
test vec::bench_retain_100000                            ... bench:      76,786 ns/iter (+/- 7,531)
test vec::bench_retain_whole_100000                      ... bench:      62,311 ns/iter (+/- 19,721)
```

manully expand

```rust
test vec::bench_retain_100000                            ... bench:      72,178 ns/iter (+/- 3,247)
test vec::bench_retain_whole_100000                      ... bench:      56,534 ns/iter (+/- 1,479)
```

TennyZhuang:
OK, the conclusion is that:

1. const generics is OK.
2. It optimize a little (10%) in most cases or architectures.
3. Performance is downgraded a little (10%) in some case and some specific architecture.
4. It optimize much (8x) on some specific architecture with unknown reason.
5. In theory, it actually optimize a unnecessary check.
6. It introduce a little code complexity.

I tend to think this PR deserves to be merged.

@oxalica what's your opinion?

oxalica:
I figured out why the code generation is weird after your change.
LLVM failed to reasoning `g.processed_len == original_len` after the second loop, thus the `drop(g)` cannot be optimized. That's why I see a never-called `memmove` in the generated code. In you manually expanded code without const generic, this issue still exists. Not sure if it's a LLVM bug.

I'd suggest to change the `while` condition to `g.procesed_len != original_len` and it's now successfully optimized.
[Generated code after](https://godbolt.org/z/bo3oq3481)

Well, this should not affect the efficiency of main loop.

TennyZhuang:
> I figured out why the code generation is weird after your change.
> LLVM failed to reasoning `g.processed_len == original_len` after the second loop, thus the `drop(g)` cannot be optimized. That's why I see a never-called `memmove` in the generated code. In you manually expanded code without const generic, this issue still exists. Not sure if it's a LLVM bug.
> 
> I'd suggest to change the `while` condition to `g.procesed_len != original_len` and it's now successfully optimized.
> [Generated code after](https://godbolt.org/z/bo3oq3481)
> 
> Well, this should not affect the efficiency of main loop.

Good catch, I have updated my code.

Currently performance on `Intel(R) Core(TM) i5-8259U CPU @ 2.30GHz`

```
test vec::bench_retain_100000                            ... bench:      56,911 ns/iter (+/- 5,288)
test vec::bench_retain_whole_100000                      ... bench:      56,371 ns/iter (+/- 5,982)
```

TennyZhuang:
@joshtriplett PTAL, Thanks

TennyZhuang:
r? @dtolnay 

It seems that @joshtriplett is busy. Could you also review my PR? it's similar to #88075

dtolnay (changes requested):
It looks like a bunch of unrelated commits are being rebased in this PR. Could you please recreate with just the changes that you intended to submit?

TennyZhuang:
Ok, I will re cherry-pick my commit later

TennyZhuang:
@dtolnay I've fixed it

dtolnay (approved):
Thanks!

library/alloc/benches/vec.rs
```diff
@@ -732,3 +732,18 @@ fn bench_flat_map_collect(b: &mut Bencher) {
     let v = vec![777u32; 500000];
     b.iter(|| v.iter().flat_map(|color| color.rotate_left(8).to_be_bytes()).collect::<Vec<_>>());
 }
+
+#[bench]
+fn bench_retain_100000(b: &mut Bencher) {
+    let v = (1..=100000).collect::<Vec<u32>>();
+    b.iter(|| {
+        let mut v = v.clone();
```
the8472, line 740: calling `.clone()` inside the benchmark loop should be avoided since it can introduce allocator noise. It's better to just refill the vec.

TennyZhuang: Thanks but the PR was merged, should I submit another PR to fix it?

the8472: I'll take care of it. I just wanted to leave a note.

PR #88061, opened 2021-08-15
Remove `hir::GenericBound::Unsized`

jackh726:
Rather than "moving" the `?Sized` bounds to the param bounds, just also check where clauses in `astconv`. I also did some related cleanup here, but that's not strictly neccesary. Also going to do a perf run here.

r? @estebank

### DIFF
--- a/compiler/rustc_typeck/src/astconv/mod.rs
+++ b/compiler/rustc_typeck/src/astconv/mod.rs
@@ -111,11 +111,6 @@ pub trait AstConv<'tcx> {
     fn record_ty(&self, hir_id: hir::HirId, ty: Ty<'tcx>, span: Span);
 }
 
-pub enum SizedByDefault {
-    Yes,
-    No,
-}
-
 #[derive(Debug)]
 struct ConvertedBinding<'a, 'tcx> {
     hir_id: hir::HirId,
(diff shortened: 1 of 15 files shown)


compiler/rustc_typeck/src/astconv/mod.rs
```diff
@@ -856,45 +884,75 @@ impl<'o, 'tcx> dyn AstConv<'tcx> + 'o {
...
 
     // Returns `true` if a bounds list includes `?Sized`.
-    pub fn is_unsized(&self, ast_bounds: &[hir::GenericBound<'_>], span: Span) -> bool {
+    pub(crate) fn add_implicitly_sized<'hir>(
+        &self,
+        bounds: &mut Bounds<'hir>,
+        ast_bounds: &'hir [hir::GenericBound<'hir>],
+        self_ty_where_predicates: Option<(hir::HirId, &'hir [hir::WherePredicate<'hir>])>,
+        span: Span,
+    ) {
```
robojumper, line 893: This comment is now wrong.

estebank (commented):
Cursory look at the code changes make sense. When I introduced `Unsized` I think there were some comments that I replaced that we might want to bring back.

jackh726:
@estebank can you be more clear? The only comment (I think) that you replaced was in `item.rs` referring to discarding the `?Sized` bounds from where clauses, but that's no longer special cased.

estebank:
r=me after rebase

PR #88066, opened 2021-08-15
Use if-let guards in the codebase and various other pattern cleanups

LeSeulArtichaut:
Dogfooding if-let guards as experimentation for the feature.

Tracking issue #51114. Conflicts with #87937.

### DIFF
--- a/compiler/rustc_expand/src/config.rs
+++ b/compiler/rustc_expand/src/config.rs
@@ -305,15 +305,14 @@ impl<'a> StripUnconfigured<'a> {
                     Some((AttrAnnotatedTokenTree::Delimited(sp, delim, inner), *spacing))
                         .into_iter()
                 }
+                AttrAnnotatedTokenTree::Token(ref token) if let TokenKind::Interpolated(ref nt) = token.kind => {
+                    panic!(
+                        "Nonterminal should have been flattened at {:?}: {:?}",
+                        token.span, nt
+                    );
+                }
                 AttrAnnotatedTokenTree::Token(token) => {
-                    if let TokenKind::Interpolated(nt) = token.kind {
-                        panic!(
-                            "Nonterminal should have been flattened at {:?}: {:?}",
-                            token.span, nt
-                        );
-                    } else {
-                        Some((AttrAnnotatedTokenTree::Token(token), *spacing)).into_iter()
-                    }
+                    Some((AttrAnnotatedTokenTree::Token(token), *spacing)).into_iter()
                 }
             })
             .collect();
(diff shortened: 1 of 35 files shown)


nagisa:
Please split up the different kinds of cleanups (i.e. those that use `if-let-guards` and those that don't) into different commits so that the changes are better isolated and easier to review.

I think the scope of this PR is otherwise reasonable enough for dogfooding purposes.

LeSeulArtichaut:
@nagisa I split the changes in two commits as you requested. Since the bulk of the changes is actually reindenting the code (caused by removing an if-let block), the diff will look less noisy if viewed with whitespace changes hidden.

I also forgot to mention that rustfmt doesn't yet support formatting if-let guards (rust-lang/rustfmt#4955) so I tried my best to format them by hand.

compiler/rustc_expand/src/config.rs
```diff
@@ -310,15 +310,13 @@ impl<'a> StripUnconfigured<'a> {
                     Some((AttrAnnotatedTokenTree::Delimited(sp, delim, inner), *spacing))
                         .into_iter()
                 }
+                AttrAnnotatedTokenTree::Token(
+                    ref token @ Token { kind: TokenKind::Interpolated(ref nt), .. },
```
nagisa, line 314: Nit: it is exactly to avoid having to write out a pattern such as this is why people tend to write code in the way as it was before.

I wonder though if

```rust
AttrAnnotatedTokenTree::Token(token) if let TokenKind::Interpolated(nt) = token.kind => { ... }
```

would work?

compiler/rustc_parse/src/lib.rs
```diff
@@ -264,20 +264,14 @@ pub fn nt_to_tokenstream(
...
-                        TokenKind::Semi,
-                        stmt.span,
-                    )),
-                    Spacing::Alone,
-                )]);
-                prepend_attrs(&stmt.attrs(), Some(&LazyTokenStream::new(tokens)))
-            } else {
-                prepend_attrs(&stmt.attrs(), stmt.tokens())
-            }
+        Nonterminal::NtStmt(ref stmt @ ast::Stmt { kind: ast::StmtKind::Empty, .. }) => {
```
nagisa, line 267: Seems like this would benefit more from an if-let guard too?

nagisa:
Seems nice. Up to you if you want to handle the comments above.

@bors delegate=LeSeulArtichaut

LeSeulArtichaut:
@nagisa Replaced the problematic patterns with if-let guards.

PR #88069, opened 2021-08-16
PGO for LLVM builds on x86_64-unknown-linux-gnu in CI

Mark-Simulacrum:
This shows up to 6% less instruction counts with larger - up to 18% - wins on cycles
on multiple benchmarks, and up to 19% wins on the -j1 wall times for rustc self-compilation.

We can afford to spend the extra cycles building LLVM essentially once more for
the x86_64-unknown-linux-gnu CI build today. The builder finishes in around 50
minutes on average, and this adds just 10 more minutes. Given the sizeable
improvements in compiler performance, this is definitely worth it.

### DIFF
--- a/src/ci/docker/host-x86_64/dist-x86_64-linux/Dockerfile
+++ b/src/ci/docker/host-x86_64/dist-x86_64-linux/Dockerfile
@@ -82,6 +82,12 @@ COPY host-x86_64/dist-x86_64-linux/build-clang.sh /tmp/
 RUN ./build-clang.sh
 ENV CC=clang CXX=clang++
 
+ENV PERF_COMMIT 1e19fc4c6168d2f7596e512f42f358f245d8f09d
+RUN curl -LS -o perf.zip https://github.com/rust-lang/rustc-perf/archive/$PERF_COMMIT.zip && \
+    unzip perf.zip && \
+    mv rustc-perf-$PERF_COMMIT rustc-perf && \
+    rm perf.zip
+
 COPY scripts/sccache.sh /scripts/
 RUN sh /scripts/sccache.sh
 
--- a/src/ci/docker/host-x86_64/dist-x86_64-linux/build-clang.sh
+++ b/src/ci/docker/host-x86_64/dist-x86_64-linux/build-clang.sh
@@ -20,14 +20,18 @@ cd clang-build
 # include path, /rustroot/include, to clang's default include path.
 INC="/rustroot/include:/usr/include"
 
+# We need compiler-rt for the profile runtime (used later to PGO the LLVM build)
+# but sanitizers aren't currently building. Since we don't need those, just
+# disable them.
 hide_output \
     cmake ../llvm \
       -DCMAKE_C_COMPILER=/rustroot/bin/gcc \
       -DCMAKE_CXX_COMPILER=/rustroot/bin/g++ \
       -DCMAKE_BUILD_TYPE=Release \
       -DCMAKE_INSTALL_PREFIX=/rustroot \
+      -DCOMPILER_RT_BUILD_SANITIZERS=OFF \
       -DLLVM_TARGETS_TO_BUILD=X86 \
-      -DLLVM_ENABLE_PROJECTS="clang;lld" \
+      -DLLVM_ENABLE_PROJECTS="clang;lld;compiler-rt" \
       -DC_INCLUDE_DIRS="$INC"
 
 hide_output make -j$(nproc)
--- a/src/ci/pgo.sh
+++ b/src/ci/pgo.sh
@@ -5,61 +5,59 @@ set -euxo pipefail
 rm -rf /tmp/rustc-pgo
 
 python3 ../x.py build --target=$PGO_HOST --host=$PGO_HOST \
-    --stage 2 library/std --rust-profile-generate=/tmp/rustc-pgo
+    --stage 2 library/std \
+    --rust-profile-generate=/tmp/rustc-pgo \
+    --llvm-profile-generate
 
+# Profile libcore compilation in opt-level=0 and opt-level=3
 RUSTC_BOOTSTRAP=1 ./build/$PGO_HOST/stage2/bin/rustc --edition=2018 \
     --crate-type=lib ../library/core/src/lib.rs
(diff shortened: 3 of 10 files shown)


Mark-Simulacrum:
@bors try @rust-timer queue

Should produce ~equivalent results to https://github.com/rust-lang/rust/pull/87940#issuecomment-899103445.

src/ci/docker/host-x86_64/dist-x86_64-linux/Dockerfile
```diff
@@ -82,6 +82,12 @@ COPY host-x86_64/dist-x86_64-linux/build-clang.sh /tmp/
 RUN ./build-clang.sh
 ENV CC=clang CXX=clang++
 
+ENV PERF_COMMIT 1e19fc4c6168d2f7596e512f42f358f245d8f09d
```
jyn514, line 85: Hmm, what's the strategy for updating this? Presumably `collector` could change in the future.

Oh, I just saw this was hard-coded before, so this isn't changing anything.

Mark-Simulacrum: Yeah, this isn't really critical to update on any regular cadence. I don't expect us to update the hash here unless the benchmarks we're using break in rustc updates (some of them use unstable features), but that is rare and should be relatively easy to work around if necessary.

Mark-Simulacrum:
Adding perf-regression-triaged; per the PR description, this is a clear win that far exceeds the small losses.

michaelwoerister:
Also note that as per my measurements in https://blog.rust-lang.org/inside-rust/2020/11/11/exploring-pgo-for-the-rust-compiler.html the actual wall-time wins are a lot bigger than one would expect from just looking at the instruction counts.

hkratz:
The [CPU cycles comparison](https://perf.rust-lang.org/compare.html?start=2bd17c1d43bba43412cc2f051323a279d6751e43&end=d031afb4badc709d74475230f7013bfe01d08a80&stat=cycles%3Au) is probably a more accurate approximation of the real improvements.

src/ci/docker/host-x86_64/dist-x86_64-linux/build-clang.sh
```diff
@@ -26,8 +26,9 @@ hide_output \
       -DCMAKE_CXX_COMPILER=/rustroot/bin/g++ \
       -DCMAKE_BUILD_TYPE=Release \
       -DCMAKE_INSTALL_PREFIX=/rustroot \
+      -DCOMPILER_RT_BUILD_SANITIZERS=OFF \
       -DLLVM_TARGETS_TO_BUILD=X86 \
-      -DLLVM_ENABLE_PROJECTS="clang;lld" \
+      -DLLVM_ENABLE_PROJECTS="clang;lld;compiler-rt" \
```
pietroalbini, line 31: What's the rationale for these changes? They don't seem to be related to the rest of the PR, but I'm not familiar with clang build flags :)

src/ci/pgo.sh
```diff
@@ -5,61 +5,58 @@ set -euxo pipefail
...
+    --crate-type=lib -Copt-level=3 ../library/core/src/lib.rs
+
+cp -r /tmp/rustc-perf ./
+chown -R $(whoami): ./rustc-perf
+cd rustc-perf
+
+RUST_LOG=collector=debug \
+RUSTC=/checkout/obj/build/$PGO_HOST/stage0/bin/rustc \
+RUSTC_BOOTSTRAP=1 \
+/checkout/obj/build/$PGO_HOST/stage0/bin/cargo build -p collector
```
pietroalbini, line 25: Why is this in a separate step?

emilyalbini (changes requested):
The new LLVM profile data also needs to be added to `bootstrap::dist::ReproducibleArtifacts`.

src/ci/docker/host-x86_64/dist-x86_64-linux/build-clang.sh
Mark-Simulacrum: We need the profile runtime (from compiler-rt) to be able to use -fprofile-gen/use flags, so we need to build it here. Sanitizers are disabled because I couldn't get one of them to build (maybe due to our old gcc version or something, not sure, but doesn't seem worth debugging).

src/ci/pgo.sh
Mark-Simulacrum: We need to build the rustc-fake binary as part of the collector package, and cargo doesn't (AFAIK) have a way to declare that dependency. I can add a comment.

jyn514: Yeah, my "binary artifacts" RFC hasn't been implemented yet.

Mark-Simulacrum:
Pushed updates per comments.

emilyalbini (approved):

klensy:
Btw, rustc's llvm version was recently updated to 13, is PGO was tested on that version?

Mark-Simulacrum:
@bors retry

Hopefully a spurious error building llvm, but not sure. Might be due to the llvm 13 upgrade...

Mark-Simulacrum:
Cc @nikic @rust-lang/wg-llvm - it seems like the crash while building llvm should be reported upstream, maybe someone could help with that.

I haven't investigated too much but it seems likely to be a regression in llvm 13 (which we recently bumped to, right?) since previous versions were building just fine across ~6 try builds I think, just earlier this week. I'll probably try to reproduce and maybe bisect but haven't done that before so if it's something known already that would be helpful.

Mark-Simulacrum:
I *think* the crash is https://bugs.llvm.org/show_bug.cgi?id=45636 (should be an error after https://reviews.llvm.org/D79140 merges, but that hasn't happened yet). That bug seems to hint that ccache used to be broken with the -fprofile flags -- our versions of sccache are pretty old (2018/2019) so I'm going to try updating them in this PR to see if that helps. The latest versions do seem to have rudimentary support for these flags I think.

Mark-Simulacrum:
@bors try @rust-timer queue

Re-trying with two goals:

* Check that the new version of sccache is still good with another run on the same version
* Start generating docker caches again

Mark-Simulacrum:
@pietroalbini -- fyi, added a second commit bumping sccache, but it seems an obvious patch so I'll likely not wait for you to re-review, just wanted to give a heads up. It'll be several hours before the try build and perf run finishes though.

PR #88072, opened 2021-08-16
Allow the iOS toolchain to be built on Linux

kit-981:
The iOS toolchain can be built on Linux with minor changes. The compilation will invoke `xcrun` to find the path to the iPhone SDK but a fake `xcrun` executable can be used.

```
#!/bin/sh
echo "/path/to/sdk"
```

The iOS toolchain can then be built and linked with rustup.

```
$ ./x.py build --stage 2 --host x86_64-unknown-linux-gnu \
  	 --target aarch64-apple-ios
$ rustup toolchain link stage1 build/x86_64-unknown-linux-gnu/stage1
```

It's possible to take this toolchain and compile an iOS executable with it. This requires the ld64 linker and an iOS SDK. The ld64 linker can be taken from [cctools](https://github.com/tpoechtrager/cctools-port). A project's .cargo/config can then be edited to use the linker for this target.

```
[target.aarch64-apple-ios]
linker = "/path/to/cctools/bin/arm-apple-darwin-ld"
rustflags = [
    "-C",
    """
link-args=
    -F/path/to/sdk/System/Library/Frameworks
    -L/path/to/sdk/usr/lib
    -L/path/to/sdk/usr/lib/system/
    -adhoc_codesign
    """,
]
```

### DIFF
--- a/src/bootstrap/sanity.rs
+++ b/src/bootstrap/sanity.rs
@@ -166,11 +166,6 @@ pub fn check(build: &mut Build) {
     }
 
     for target in &build.targets {
-        // Can't compile for iOS unless we're on macOS
-        if target.contains("apple-ios") && !build.build.contains("apple-darwin") {
-            panic!("the iOS target is only supported on macOS");
-        }
-
         build
             .config
             .target_config


jyn514:
Hmm, could you change this to check if `xcrun` is installed instead? That way it still fails quickly if you don't have the MacOS SDK installed.

kit-981:
> Hmm, could you change this to check if `xcrun` is installed instead? That way it still fails quickly if you don't have the MacOS SDK installed.

This case is already handled:

```
patches/rust % ./x.py build --stage 2 --host x86_64-unknown-linux-gnu --target aarch64-apple-ios
Submodules updated in 0.01 seconds
    Finished dev [unoptimized + debuginfo] target(s) in 0.12s
running: "xcrun" "--show-sdk-path" "--sdk" "iphoneos"


error occurred: Failed to find tool. Is `xcrun` installed?


Build completed unsuccessfully in 0:00:00
```

PR #88077, opened 2021-08-16
Generate an iOS LLVM target with a specific version

kit-981:
This commit adds the `LC_VERSION_MIN_IPHONEOS` load command to the Mach-O header generated for `aarch64-apple-ios` binaries. The operating system will look for this load command to determine the minimum supported operating system version and will not allow the binary to run if it's absent. This logic already exists for the simulator toolchain.

I've been using `otool` from a [cctools](https://github.com/tpoechtrager/cctools-port) toolchain to parse the header and validate that this change adds the required load command.

This change appears to be enough to build Rust binaries that can run on a jailbroken iPhone.

### DIFF
--- a/compiler/rustc_target/src/spec/aarch64_apple_ios.rs
+++ b/compiler/rustc_target/src/spec/aarch64_apple_ios.rs
@@ -2,8 +2,15 @@ use super::apple_sdk_base::{opts, Arch};
 use crate::spec::{FramePointer, Target, TargetOptions};
 
 pub fn target() -> Target {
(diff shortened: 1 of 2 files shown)


LeSeulArtichaut:
This seems similar to #87699.

petrochenkov:
Closing in favor of the earlier PR https://github.com/rust-lang/rust/pull/87699.

petrochenkov:
https://github.com/rust-lang/rust/pull/87699#issuecomment-903240914
@bors r+ rollup

PR #88080, opened 2021-08-16
Skip assert ICE with default_method_body_is_const

fee1-dead:
functions marked with #[default_method_body_is_const] would
ICE when being const checked due to it not being a const function:
`tcx.is_const_fn_raw(did)` returns false. We should skip this assert
when it is marked with that attribute.

r? @oli-obk


oli-obk:
Please add a regression test. We're likely going to touch this more in the future.

oli-obk:
@bors delegate+

r=me once the test passes

PR #88082, opened 2021-08-16
Take into account jobs number for rustdoc GUI tests

### LINKED ISSUE #88054: Use the "-j" value in the rustdoc-gui tool
#86692 greatly improved the tool by allowing to run multiple tests at the same time. However, it doesn't take into account the "-j" option but it should.

GuillaumeGomez:
Fixes #88054.

r? @Mark-Simulacrum

### DIFF
--- a/src/tools/rustdoc-gui/tester.js
+++ b/src/tools/rustdoc-gui/tester.js
@@ -17,6 +17,11 @@ function showHelp() {
     console.log("  --no-headless              : disable headless mode");
     console.log("  --help                     : show this message then quit");
     console.log("  --tests-folder [PATH]      : location of the .GOML tests folder");
+    console.log("  --jobs [NUMBER]            : number of threads to run tests on");
+}
+
+function isNumeric(s) {
+    return /^\d+$/.test(s);
 }
 
 function parseOptions(args) {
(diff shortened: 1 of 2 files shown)


src/tools/rustdoc-gui/tester.js
```diff
@@ -153,7 +167,11 @@ async function main(argv) {
     files.sort();
 
     console.log(`Running ${files.length} rustdoc-gui tests...`);
-    process.setMaxListeners(files.length + 1);
+    if (opts["jobs"] < 1) {
+        process.setMaxListeners(files.length + 1);
+    } else {
+        process.setMaxListeners(opts["jobs"]);
```
dns2utf8, line 173: Will this give a warning to the users? If I remember correctly that warning was the reason why I increased the number to `files.length + 1`

GuillaumeGomez: I didn't have any warning, so no idea. But we shouldn't go above the limit provided by the tester in any case.

dns2utf8: Cool :relaxed: It is a very clean solution :+1:

GuillaumeGomez: \o/

dns2utf8 (approved):

PR #88083, opened 2021-08-16
Improve non_fmt_panics suggestion based on trait impls. 

### LINKED ISSUE #87313: non_fmt_panics cannot migrate to 2021 for assert with non-string payload
I tried this code:

```rust
fn f() {
    assert!(true, 123);
}
```

I expected to see this happen: This includes a machine-applicable fix to migrate to 2021.

Instead, this happened: The suggestion is `MaybeIncorrect` preventing automatic migration.

Part of the problem is that there isn't an easy way to convert this to `panic_any`. For example, it could convert to:

```rust
fn f() {
    if !(expr) { ::std::panic::panic_any(123); }
}
```

which is pretty awful and probably has other downsides.  Trying to format with `{:?}` is also probably not a reliable option, since it would require the payload to impl Debug, and the user's code may be relying on catching the panic and inspecting the payload.

I don't know what is really practical here.

### Meta
`rustc 1.55.0-nightly (014026d1a 2021-07-19)`

### LINKED ISSUE #87999: non_panic_fmt lint suggestion should check what traits are implemented
As suggested by @nrc in https://github.com/rust-lang/rust/issues/87313#issuecomment-898000387, the `panic_fmt_lint` should suggest `"{:?}"` if the type implements Debug and not Display.

m-ou-se:
This improves the non_fmt_panics lint suggestions by checking first which trait (Display or Debug) are actually implemented on the type.

Fixes https://github.com/rust-lang/rust/issues/87313

Fixes https://github.com/rust-lang/rust/issues/87999

Before:

```
help: add a "{}" format string to Display the message
  |
2 |     panic!("{}", Some(1));
  |            +++++
help: or use std::panic::panic_any instead
  |
2 |     std::panic::panic_any(Some(1));
  |     ~~~~~~~~~~~~~~~~~~~~~
```

After:

```
help: add a "{:?}" format string to use the Debug implementation of `Option<i32>`
  |
2 |     panic!("{:?}", Some(1));
  |            +++++++
help: or use std::panic::panic_any instead
  |
2 |     std::panic::panic_any(Some(1));
  |     ~~~~~~~~~~~~~~~~~~~~~
```

r? @estebank

### DIFF
--- a/compiler/rustc_lint/src/non_fmt_panic.rs
+++ b/compiler/rustc_lint/src/non_fmt_panic.rs
@@ -102,7 +102,7 @@ fn check_panic<'tcx>(cx: &LateContext<'tcx>, f: &'tcx hir::Expr<'tcx>, arg: &'tc
 
     cx.struct_span_lint(NON_FMT_PANICS, arg_span, |lint| {
         let mut l = lint.build("panic message is not a string literal");
-        l.note("this usage of panic!() is deprecated; it will be a hard error in Rust 2021");
+        l.note(&format!("this usage of {}!() is deprecated; it will be a hard error in Rust 2021", symbol_str));
         l.note("for more information, see <https://doc.rust-lang.org/nightly/edition-guide/rust-2021/panic-macro-consistency.html>");
         if !span.contains(arg_span) {
             // No clue where this argument is coming from.
--- a/compiler/rustc_lint/src/non_fmt_panic.rs
+++ b/compiler/rustc_lint/src/non_fmt_panic.rs
@@ -2,12 +2,15 @@ use crate::{LateContext, LateLintPass, LintContext};
 use rustc_ast as ast;
 use rustc_errors::{pluralize, Applicability};
 use rustc_hir as hir;
+use rustc_infer::infer::TyCtxtInferExt;
 use rustc_middle::lint::in_external_macro;
 use rustc_middle::ty;
+use rustc_middle::ty::subst::InternalSubsts;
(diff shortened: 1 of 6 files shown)


compiler/rustc_lint/src/non_fmt_panic.rs
```diff
@@ -129,20 +132,57 @@ fn check_panic<'tcx>(cx: &LateContext<'tcx>, f: &'tcx hir::Expr<'tcx>, arg: &'tc
...
+
+            let suggest_panic_any = !is_str && panic == sym::std_panic_macro;
+
+            let fmt_applicability = if suggest_panic_any {
+                // If we can use panic_any, use that as the MachineApplicable suggestion.
+                Applicability::MaybeIncorrect
+            } else {
+                // If we don't suggest panic_any, using a format string is our best bet.
+                Applicability::MachineApplicable
+            };
```
estebank, line 154: Isn't this logic reversed?

estebank (approved):
LGTM, except for the concern around applicability.

compiler/rustc_lint/src/non_fmt_panic.rs
m-ou-se: No. It marks the format-string suggestion (adding `"{}"` or `"{:?}"`) as machine applicable only when the `panic_any` suggestion isn't, as there shouldn't be more than one alternative marked as machine applicable.

m-ou-se:
Raising priority so it can be part of the edition migration crater run.

@bors p=1

PR #88086, opened 2021-08-16
BTree: toughen panicky test of clone()

ssomers:
Test did not cover the second half of `clone_subtree` and why this clones key & value first.


dtolnay (approved):
Thanks!

PR #88087, opened 2021-08-16
Check that a box expression's type is Sized

### LINKED ISSUE #87935: ICE with box syntax and unsized types
Originally reported by @chengniansun [here](https://github.com/rust-lang/rust/issues/86535#issuecomment-871614379)

### Code

[playground](https://play.rust-lang.org/?version=nightly&mode=debug&edition=2018&gist=34f53f7f617a2baadda6f5cdbb87232c)

```Rust
#![feature(box_syntax)]
fn main() {
    let x: Box<[isize]> = box { loop {} };
}
```

### Meta

`rustc --version --verbose`:
```
rustc 1.55.0-nightly (6d820866a 2021-06-29)
binary: rustc
commit-hash: 6d820866a27b1949e237be79b9c8c0145fe728b7
commit-date: 2021-06-29
release: 1.55.0-nightly
LLVM version: 12.0.1
```

### Error output

```
warning: unreachable expression
 --> mutant.rs:3:27
  |
3 |     let x: Box<[isize]> = box { loop {} };
  |                           ^^^^^^-------^^
  |                           |     |
  |                           |     any code following this expression is unreachable
  |                           unreachable expression
  |
  = note: `#[warn(unreachable_code)]` on by default

warning: unused variable: `x`
 --> mutant.rs:3:9
  |
3 |     let x: Box<[isize]> = box { loop {} };
  |         ^ help: if this is intentional, prefix it with an underscore: `_x`
  |

warning: 2 warnings emitted

error: internal compiler error: broken MIR in DefId(0:3 ~ mutant[317d]::main) (NoSolution): could not prove Binder(TraitPredicate(<[isize] as std::marker::Sized>), [])
 --> mutant.rs:3:27
  |
3 |     let x: Box<[isize]> = box { loop {} };
  |                           ^^^^^^^^^^^^^^^
  |
  = note: delayed at compiler/rustc_mir/src/borrow_check/type_check/mod.rs:253:27

thread 'rustc' panicked at 'no errors encountered even though `delay_span_bug` issued', compiler/rustc_errors/src/lib.rs:1022:13

note: rustc 1.55.0-nightly (6d820866a 2021-06-29) running on x86_64-unknown-linux-gnu

query stack during panic:
end of query stack
```

<p>

```
warning: unreachable expression
 --> mutant.rs:3:27
  |
3 |     let x: Box<[isize]> = box { loop {} };
  |                           ^^^^^^-------^^
  |                           |     |
  |                           |     any code following this expression is unreachable
  |                           unreachable expression
  |
  = note: `#[warn(unreachable_code)]` on by default

warning: unused variable: `x`
 --> mutant.rs:3:9
  |
3 |     let x: Box<[isize]> = box { loop {} };
  |         ^ help: if this is intentional, prefix it with an underscore: `_x`
  |

warning: 2 warnings emitted

error: internal compiler error: broken MIR in DefId(0:3 ~ mutant[317d]::main) (NoSolution): could not prove Binder(TraitPredicate(<[isize] as std::marker::Sized>), [])
 --> mutant.rs:3:27
  |
3 |     let x: Box<[isize]> = box { loop {} };
  |                           ^^^^^^^^^^^^^^^
  |
  = note: delayed at compiler/rustc_mir/src/borrow_check/type_check/mod.rs:253:27

thread 'rustc' panicked at 'no errors encountered even though `delay_span_bug` issued', compiler/rustc_errors/src/lib.rs:1022:13
stack backtrace:
   0: rust_begin_unwind
             at /rustc/6d820866a27b1949e237be79b9c8c0145fe728b7/library/std/src/panicking.rs:515:5
   1: std::panicking::begin_panic_fmt
             at /rustc/6d820866a27b1949e237be79b9c8c0145fe728b7/library/std/src/panicking.rs:457:5
   2: rustc_errors::HandlerInner::flush_delayed
   3: <rustc_errors::HandlerInner as core::ops::drop::Drop>::drop
   4: core::ptr::drop_in_place<rustc_session::parse::ParseSess>
   5: <alloc::rc::Rc<T> as core::ops::drop::Drop>::drop
   6: core::ptr::drop_in_place<rustc_interface::interface::Compiler>
   7: rustc_span::with_source_map
   8: rustc_interface::interface::create_compiler_and_run
   9: scoped_tls::ScopedKey<T>::set

note: rustc 1.55.0-nightly (6d820866a 2021-06-29) running on x86_64-unknown-linux-gnu

query stack during panic:
end of query stack
```

</p>
</details>

jesyspa:
This resolves [issue 87935](https://github.com/rust-lang/rust/issues/87935).

This makes E0161 (move from an unsized rvalue) much less common.  I've replaced the test to use [this case](https://github.com/rust-lang/rust/blob/master/src/test/ui/object-safety/object-safety-by-value-self-use.rs), when a boxed `dyn` trait is passed by value, but that isn't an error when `unsized_locals` is enabled.  I think it may be possible to get rid of E0161 entirely by checking that case earlier, but I'm not sure if that's desirable?

### DIFF
--- a/compiler/rustc_error_codes/src/error_codes/E0161.md
+++ b/compiler/rustc_error_codes/src/error_codes/E0161.md
@@ -4,11 +4,18 @@ Erroneous code example:
 
 ```compile_fail,E0161
 #![feature(box_syntax)]
+trait Bar {
+    fn f(self);
+}
+
+impl Bar for i32 {
+    fn f(self) {}
+}
 
 fn main() {
-    let array: &[isize] = &[1, 2, 3];
-    let _x: Box<[isize]> = box *array;
-    // error: cannot move a value of type [isize]: the size of [isize] cannot
+    let b: Box<dyn Bar> = box (0 as i32);
+    b.f();
+    // error: cannot move a value of type dyn Bar: the size of dyn Bar cannot
     //        be statically determined
 }
 ```
@@ -22,8 +29,17 @@ it around as usual. Example:
 ```
 #![feature(box_syntax)]
 
+trait Bar {
+    fn f(&self);
+}
+
+impl Bar for i32 {
+    fn f(&self) {}
+}
+
 fn main() {
-    let array: &[isize] = &[1, 2, 3];
-    let _x: Box<&[isize]> = box array; // ok!
+    let b: Box<dyn Bar> = box (0 as i32);
+    b.f();
+    // ok!
 }
 ```
--- a/compiler/rustc_middle/src/traits/mod.rs
+++ b/compiler/rustc_middle/src/traits/mod.rs
@@ -225,6 +225,8 @@ pub enum ObligationCauseCode<'tcx> {
     SizedReturnType,
     /// Yield type must be `Sized`.
     SizedYieldType,
+    /// Box expression result type must be `Sized`.
+    SizedBoxType,
     /// Inline asm operand type must be `Sized`.
     InlineAsmSized,
     /// `[T, ..n]` implies that `T` must be `Copy`.
--- a/compiler/rustc_mir/src/borrow_check/type_check/mod.rs
+++ b/compiler/rustc_mir/src/borrow_check/type_check/mod.rs
@@ -1893,9 +1893,7 @@ impl<'a, 'tcx> TypeChecker<'a, 'tcx> {
 
                 // While this is located in `nll::typeck` this error is not
                 // an NLL error, it's a required check to prevent creation
-                // of unsized rvalues in certain cases:
-                // * operand of a box expression
-                // * callee in a call expression
+                // of unsized rvalues in a call expression.
                 diag.emit();
             }
         }
--- a/compiler/rustc_trait_selection/src/traits/error_reporting/suggestions.rs
+++ b/compiler/rustc_trait_selection/src/traits/error_reporting/suggestions.rs
@@ -2072,6 +2072,9 @@ impl<'a, 'tcx> InferCtxtExt<'tcx> for InferCtxt<'a, 'tcx> {
             ObligationCauseCode::SizedYieldType => {
                 err.note("the yield type of a generator must have a statically known size");
             }
+            ObligationCauseCode::SizedBoxType => {
+                err.note("the type of a box expression must have a statically known size");
+            }
             ObligationCauseCode::AssignmentLhsSized => {
                 err.note("the left-hand-side of an assignment must have a statically known size");
             }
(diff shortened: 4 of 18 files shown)


jesyspa:
There's [another possible implementation](https://github.com/jesyspa/rust/commit/fe5b760967e52a1c78eebd9f247067a5b8350cbc), where we issue E0161 for unsized box expressions even when unsized_locals are disabled.  That approach keeps the checking in `rustc_mir`, but it currently emits duplicate errors for cases like ```rust
fn foo(x: Box<[i32]>) {
    box *x;
}```

Given that the implementation in this PR gives a clearer error message, I think it's probably the better option.

compiler/rustc_error_codes/src/error_codes/E0161.md
```diff
@@ -4,11 +4,18 @@ Erroneous code example:
```
jackh726, line 4: Why change this file?

jackh726 (commented):
Why change the E0161 output?

compiler/rustc_error_codes/src/error_codes/E0161.md
jesyspa: The previous example no longer works, since the "box expression type must be Sized" error is emitted before E0161.  This causes the following CI break: https://github.com/rust-lang/rust/runs/3342569972

jackh726: Can you move that to a separate PR? Along with any test changes. That way it's a bit more clear the direct impact of the change in diagnostics here.

jackh726: Or, if you want to switch the commit orders to have the test and example changes be a commit before and everything else after, that works too.

jesyspa: Sure, I've split it up as follows:
1. All changes to E0161; this commit should compile by itself.
2. Code updates.
3. Update tests where the error message has changed, plus add a regression test for this issue specifically.

jackh726:
Thanks! Going to @bors rollup=never in case there are any subtle changes here that need to be bisected.

@bors r+

nbdd0121:
I wonder if it's worth a new obligation cause code if we want to remove box syntax eventually anyway 🤔

jackh726:
Those are cheap :)

jesyspa:
Thanks for the review!

> I wonder if it's worth a new obligation cause code if we want to remove box syntax eventually anyway thinking

I don't think this will complicate it, since `check_box` can be removed in its entirety when the syntax goes away.  The cost should only be paid by code that uses box expressions.

PR #88090, opened 2021-08-16
Perform type inference in range pattern

### LINKED ISSUE #88074: Range pattern does not perform type inference
```rust
trait Zero {
    const ZERO: Self;
}

impl Zero for i32 {
    const ZERO: Self = 0;
}

fn foo() {
    match 1 {
        Zero::ZERO ..= 1 => {},
        _ => {},
    }
}
```

currently throws E0029:

```
error[E0029]: only `char` and numeric types are allowed in range patterns
  --> src/lib.rs:11:9
   |
11 |         Zero::ZERO ..= 1 => {},
   |         ^^^^^^^^^^     - this is of type `{integer}`
   |         |
   |         this is of type `_` but it should be `char` or numeric

For more information about this error, try `rustc --explain E0029`.
```

while it should infer the constant to be `<i32 as Zero>::ZERO`.

@rustbot modify labels: +T-compiler +A-inference 
@rustbot claim

nbdd0121:
Fix #88074

### DIFF
--- a/compiler/rustc_typeck/src/check/pat.rs
+++ b/compiler/rustc_typeck/src/check/pat.rs
@@ -448,16 +448,22 @@ impl<'a, 'tcx> FnCtxt<'a, 'tcx> {
         ti: TopInfo<'tcx>,
     ) -> Ty<'tcx> {
         let calc_side = |opt_expr: Option<&'tcx hir::Expr<'tcx>>| match opt_expr {
-            None => (None, None),
+            None => None,
             Some(expr) => {
                 let ty = self.check_expr(expr);
-                // Check that the end-point is of numeric or char type.
-                let fail = !(ty.is_numeric() || ty.is_char() || ty.references_error());
-                (Some(ty), Some((fail, ty, expr.span)))
+                // Check that the end-point is possibly of numeric or char type.
+                // The early check here is not for correctness, but rather better
+                // diagnostics (e.g. when `&str` is being matched, `expected` will
+                // be peeled to `str` while ty here is still `&str`, if we don't
+                // err ealy here, a rather confusing unification error will be
+                // emitted instead).
+                let fail =
+                    !(ty.is_numeric() || ty.is_char() || ty.is_ty_var() || ty.references_error());
+                Some((fail, ty, expr.span))
             }
         };
-        let (lhs_ty, lhs) = calc_side(lhs);
-        let (rhs_ty, rhs) = calc_side(rhs);
+        let mut lhs = calc_side(lhs);
+        let mut rhs = calc_side(rhs);
 
         if let (Some((true, ..)), _) | (_, Some((true, ..))) = (lhs, rhs) {
             // There exists a side that didn't meet our criteria that the end-point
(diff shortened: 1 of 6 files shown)


jackh726:
This is *technically* a behavior change (even though I agree it's more of a bug fix), so I'm just going to cc @rust-lang/compiler 

(If anyone else knows this code better and wants to take the review, feel free)

src/test/ui/pattern/issue-88074-pat-range-type-inference-err.rs
```diff
@@ -0,0 +1,15 @@
+trait Zero {
+    const ZERO: Self;
+}
+
+impl Zero for String {
+    const ZERO: Self = String::new();
```
estebank, line 6: When associated consts are involved in this error, we should likely point at them.

src/test/ui/pattern/issue-88074-pat-range-type-inference-err.stderr
```diff
@@ -0,0 +1,12 @@
+error[E0029]: only `char` and numeric types are allowed in range patterns
+  --> $DIR/issue-88074-pat-range-type-inference-err.rs:11:9
+   |
+LL |         Zero::ZERO ..= Zero::ZERO => {},
+   |         ----------^^^^^----------
+   |         |              |
+   |         |              this is of type `_` but it should be `char` or numeric
+   |         this is of type `_` but it should be `char` or numeric
```
estebank, line 8: Somewhere we are not actually using `resolve_ty_if_possible`. We should be mentioning that these are of type `String`.

compiler/rustc_typeck/src/check/pat.rs
```diff
@@ -466,25 +467,39 @@ impl<'a, 'tcx> FnCtxt<'a, 'tcx> {
...
+        // This check is needed if both sides are inference variables.
+        let ty = self.resolve_vars_if_possible(expected);
+        if !(ty.is_numeric() || ty.is_char() || ty.references_error()) {
+            if let Some((ref mut fail, _, _)) = lhs {
+                *fail = true;
+            }
+            if let Some((ref mut fail, _, _)) = rhs {
+                *fail = true;
+            }
+            self.emit_err_pat_range(span, lhs, rhs);
```
estebank, line 499: I think this is why. We pass in `lhs` and `rhs`. Inside of `emit_err_pat_range` we don't `resolve_vars_if_possible` on `lhs.1` or `rhs.1`. I *think* that doing that *should* make the labels talk about `String` instead of `_`.

estebank (commented):
The changes look reasonable after a cursory review. @rust-lang/lang should chime in and approve the behavioral change (unless this was intentional). With that I can take a closer look at the code to approve or as for more changes.

jackh726:
I'm going to go ahead and nominate this for the lang team:

Do we want to allow inference in range patterns (as in https://github.com/rust-lang/rust/blob/353c03955dfe0831e4b58f9e05e99447c06adff2/src/test/ui/pattern/issue-88074-pat-range-type-inference.rs)? The `Self` type here could be inferred to be `i32`, as would be in a non-range pattern.

nikomatsakis (approved):

nikomatsakis:
To summarize the effects of this PR, I believe it implements the following logic:

* When type checking a range pattern...
    * In a range pattern, the LHS and RHS must be of equal type
    * Further, those types must be known to be char or integral

whereas before the logic was


* When type checking a range pattern...
    * In a range pattern, the LHS and RHS must be known to be char or integral
    * And they must be the same type

This change in ordering makes all the difference for the inferencer, of course, since it is able to propagate types from one side to the other. Does this sound correct?

jackh726:
> To summarize the effects of this PR, I believe it implements the following logic:
> 
> * When type checking a range pattern...
>   
>   * In a range pattern, the LHS and RHS must be of equal type
>   * Further, those types must be known to be char or integral
> 
> whereas before the logic was
> 
> * When type checking a range pattern...
>   
>   * In a range pattern, the LHS and RHS must be known to be char or integral
>   * And they must be the same type
> 
> This change in ordering makes all the difference for the inferencer, of course, since it is able to propagate types from one side to the other. Does this sound correct?

That looks correct to me.

@nbdd0121 could you actually add that explanation as a comment? That makes it abundantly clear what's going on.

nbdd0121:
A more accurate description:

* When type checking a range pattern...
  * In a range pattern, if types of LHS and RHS are known, they must be char or integral
    * This is just for better diagnostics (see note 1 below).
  * LHS, RHS and the expected type must all be the same type
  * Further, those types must be known to be char or integral

whereas before the logic was

* When type checking a range pattern...
  * In a range pattern, the LHS and RHS must be known to be char or integral
  * And LHS, RHS and the expected type must all be the same type

Note 1: the early check if LHS or RHS types are known is needed because range pattern will strip away references. So if the early check is omitted and we just start unifying types, then
```rust
match "A" {
  "A".."B" => (),
  _ => (),
}
```
will have LHS and RHS be `&str` while expected type is `str`, and the unification will fail and produce very terrible diagnostics. So in this case I just keep the current early check and let rustc complain that `&str` isn't char or integral rather than `&str` and `str` mismatches.

nbdd0121:
BTW one more thing: with this PR,
```rust
match Zero::ZERO {
	Zero::ZERO ..= Zero::ZERO => {},
	_ => {},
}
```
will still complain that `_` is not char or integral like it does now. I wonder if I should change it to complain that "type must be known at this point" instead?

jackh726:
> Note 1: the early check if LHS or RHS types are known is needed because range pattern will strip away references. So if the early check is omitted and we just start unifying types, then
> 
> ```rust
> match "A" {
>   "A".."B" => (),
>   _ => (),
> }
> ```
> 
> will have LHS and RHS be `&str` while expected type is `str`, and the unification will fail and produce very terrible diagnostics. So in this case I just keep the current early check and let rustc complain that `&str` isn't char or integral rather than `&str` and `str` mismatches.

Interesting...can you make sure there is a test that covers this case?


> BTW one more thing: with this PR,
> 
> ```rust
> match Zero::ZERO {
> 	Zero::ZERO ..= Zero::ZERO => {},
> 	_ => {},
> }
> ```
> 
> will still complain that `_` is not char or integral like it does now. I wonder if I should change it to complain that "type must be known at this point" instead?

What happens with non-range patterns?

jackh726:
I would say if the change to emit "type annotations needed" is small, that's preferable, so behavior lines up. If not, then I think it can be left as a followup FIXME

nbdd0121:
v1 -> v2:
* add extra explanatory comments
* add `resolve_vars_if_possible` to `emit_err_pat_range`
* emit E0282 type annotations needed if type is `_`

nikomatsakis:
@rfcbot fcp cancel

nikomatsakis:
I don't think we needed both teams on this

nikomatsakis:
@rfcbot fcp merge

We discussed this in the @rust-lang/lang meeting [today](https://github.com/rust-lang/lang-team/blob/master/minutes/2021-09-21.md#perform-type-inference-in-range-pattern-rust88090). General consensus was that this change was quite reasonable; in order to change the assumption that L/R hand sides of `..` are the same type, we'd have to alter the `Range` type to have multiple type parameters, which would be a very large change (and of course the assumption is already baked in, just enforced later).

PR #88093, opened 2021-08-16
[rustdoc] Wrap code blocks in <code> tag

### LINKED ISSUE #88020: rustdoc,  when I use google translate, that translate the doc as well as the code.
rustdoc, when I use google translate. 

the  rust  code and  crates name. translated like this:

image

the rust-book work well.
I found that using `<code></code> `protect code do not translate.

Could you update the rustdoc  using `<code>` tag to fix it?

Kobzol:
This PR modifies Rustdoc output so that fenced code snippets, items and whole file source codes are wrapped in `<pre><code>` instead of just `<pre>`. This should improve the semantic meaning of the generated content.

I'm not sure what to do about `render_attributes_in_pre` and `render_attributes_in_code`. These functions were clearly expected to be used for things inside `<pre>` or `<code>`, and since I added `<code>` in this PR, some of them will be used in a different context than before. However, it seems to me that even before they were not consistent. For example, `item_constant` used `render_attributes_in_code` for its attributes, however there was no `<code>` used for constants before this PR...

Should I create some `rustdoc-gui` tests? For example to check that all `<pre>` tags have a `<code>` child?

Fixes: https://github.com/rust-lang/rust/issues/88020

### DIFF
--- a/src/librustdoc/html/highlight.rs
+++ b/src/librustdoc/html/highlight.rs
@@ -65,10 +65,11 @@ fn write_header(out: &mut Buffer, class: Option<&str>, extra_content: Option<Buf
         out.push_buffer(extra);
     }
     if let Some(class) = class {
-        writeln!(out, "<pre class=\"rust {}\">", class);
+        write!(out, "<pre class=\"rust {}\">", class);
     } else {
-        writeln!(out, "<pre class=\"rust\">");
+        write!(out, "<pre class=\"rust\">");
     }
+    write!(out, "<code>");
 }
 
 /// Convert the given `src` source code into HTML by adding classes for highlighting.
@@ -101,7 +102,7 @@ fn write_code(
 }
 
 fn write_footer(out: &mut Buffer, playground_button: Option<&str>) {
-    writeln!(out, "</pre>{}</div>", playground_button.unwrap_or_default());
+    writeln!(out, "</code></pre>{}</div>", playground_button.unwrap_or_default());
 }
 
 /// How a span of text is classified. Mostly corresponds to token kinds.
--- a/src/librustdoc/html/render/print_item.rs
+++ b/src/librustdoc/html/render/print_item.rs
@@ -455,24 +455,25 @@ fn item_function(w: &mut Buffer, cx: &Context<'_>, it: &clean::Item, f: &clean::
         + name.as_str().len()
         + generics_len;
 
-    w.write_str("<pre class=\"rust fn\">");
-    render_attributes_in_pre(w, it, "");
-    w.reserve(header_len);
-    write!(
-        w,
-        "{vis}{constness}{asyncness}{unsafety}{abi}fn \
-         {name}{generics}{decl}{notable_traits}{where_clause}</pre>",
-        vis = vis,
-        constness = constness,
-        asyncness = asyncness,
-        unsafety = unsafety,
-        abi = abi,
-        name = name,
-        generics = f.generics.print(cx),
-        where_clause = print_where_clause(&f.generics, cx, 0, true),
-        decl = f.decl.full_print(header_len, 0, f.header.asyncness, cx),
-        notable_traits = notable_traits_decl(&f.decl, cx),
-    );
+    wrap_item(w, "fn", |w| {
+        render_attributes_in_pre(w, it, "");
+        w.reserve(header_len);
+        write!(
+            w,
+            "{vis}{constness}{asyncness}{unsafety}{abi}fn \
+             {name}{generics}{decl}{notable_traits}{where_clause}",
+            vis = vis,
+            constness = constness,
+            asyncness = asyncness,
+            unsafety = unsafety,
+            abi = abi,
+            name = name,
+            generics = f.generics.print(cx),
+            where_clause = print_where_clause(&f.generics, cx, 0, true),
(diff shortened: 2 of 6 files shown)


src/librustdoc/html/highlight.rs
```diff
@@ -69,6 +69,7 @@ fn write_header(out: &mut Buffer, class: Option<&str>, extra_content: Option<Buf
     } else {
         writeln!(out, "<pre class=\"rust\">");
     }
+    write!(out, "<code>");
```
GuillaumeGomez, line 72: I think this is not ideal because before we have `writeln!`, what about:

```rust
    if let Some(class) = class {
        writeln!(out, "<pre class=\"rust {}\"><code>", class);
    } else {
        writeln!(out, "<pre class=\"rust\"><code>");
    }
```

?

src/librustdoc/html/render/print_item.rs
```diff
@@ -1322,6 +1341,15 @@ where
     w.write_str("</div>")
 }
 
+fn wrap_item<F>(w: &mut Buffer, item_name: &str, f: F)
```
GuillaumeGomez, line 1344: Great idea! :+1:

src/test/rustdoc-gui/type-declation-overflow.goml
```diff
@@ -5,4 +5,4 @@ size: (1100, 800)
 // Logically, the <body> scroll width should be the width of the window.
 assert-property: ("body", {"scrollWidth": "1100"})
 // However, since there is overflow in the type declaration, its scroll width is bigger.
-assert-property: (".type-decl pre", {"scrollWidth": "1324"})
+assert-property: (".type-decl pre", {"scrollWidth": "1332"})
```
GuillaumeGomez, line 8: Normally, this shouldn't have changed...

GuillaumeGomez:
> I'm not sure what to do about `render_attributes_in_pre` and `render_attributes_in_code`. These functions were clearly expected to be used for things inside `<pre>` or `<code>`, and since I added `<code>` in this PR, some of them will be used in a different context than before. However, it seems to me that even before they were not consistent. For example, `item_constant` used `render_attributes_in_code` for its attributes, however there was no `<code>` used for constants before this PR...

Another shortcoming of rustdoc... Welcome to a huge years old codebase! :)

For consistency it should be applied the same on equivalent items but I think it's a bit out of scope here. For things other than codeblocks, just having a `<code>` (without the `<pre>` wrapping) is fine.

> Should I create some `rustdoc-gui` tests? For example to check that all `<pre>` tags have a `<code>` child?

Well, not all `<code>` are supposed to be wrapped. For example, an inline code in documentation like `this one` isn't supposed to be wrapped into a `<pre>`.

src/librustdoc/html/highlight.rs
Kobzol: I did it in this way because if I also used `writeln!` here, it would generate `<pre>\n<code>\n`, which added a new empty line at the beginning of the code block. I originally wanted to use the code that you have posted, but it seemed like a needless duplication of `<code>`.

Is there something that you don't like with the current code here? I'm not sure what is the problem.

src/librustdoc/html/render/print_item.rs
Kobzol: I simply borrowed it from `wrap_into_docblock` above :)

src/librustdoc/html/highlight.rs
GuillaumeGomez: I want the backline after the tags. If the backline isn't needed, then better to replace `writeln!` with `write!`.

Kobzol: I see. I think that the newline isn't actually even needed, so I'll try to remove it. If it's needed, I'll put it after the tags.

src/test/rustdoc-gui/type-declation-overflow.goml
Kobzol: I reverted the change to the test and instead changed CSS to remove padding from `<pre><code>` blocks.

Kobzol:
I changed the newlines after `<pre><code>`, now no newline is generated. I also added a simple test that checks that a few item types and doc code blocks contain `<pre><code>`.

src/librustdoc/html/render/print_item.rs
```diff
@@ -455,24 +455,25 @@ fn item_function(w: &mut Buffer, cx: &Context<'_>, it: &clean::Item, f: &clean::
...
-        decl = f.decl.full_print(header_len, 0, f.header.asyncness, cx),
-        notable_traits = notable_traits_decl(&f.decl, cx),
-    );
+    wrap_item(w, "fn", |w| {
+        render_attributes_in_pre(w, it, "");
+        w.reserve(header_len);
+        write!(
+            w,
+            "{vis}{constness}{asyncness}{unsafety}{abi}fn \
+         {name}{generics}{decl}{notable_traits}{where_clause}",
```
GuillaumeGomez, line 464: Please keep the alignment in the string as well (nitpicking hello).

Kobzol: I reverted the alignment as it was before. Thanks :)

GuillaumeGomez:
I checked locally and it looks good to me.

Thanks!

@bors: r+

Issue #87959, opened 2021-08-12
Stabilize Rust 2021


m-ou-se:
- [x] FCP on closure capture changes: https://github.com/rust-lang/rust/issues/88126
- [x] FCP on reserved prefixes: https://github.com/rust-lang/rust/issues/88140
- [x] FCP on force-warn: https://github.com/rust-lang/rust/issues/86516#issuecomment-900382193
- [x] Decide that Rust 1.56 will ship Rust 2021 as stable: [Zulip thread](https://rust-lang.zulipchat.com/#narrow/stream/268952-edition-2021/topic/Release.20Decision/near/251224903)
- [x] Make `--edition=2021` stable in rustc: https://github.com/rust-lang/rust/pull/88100
- [x] Remove/stabilize experimental 2021 cargo feature: https://github.com/rust-lang/cargo/pull/9800
    - [x] Update `cargo` submodule: https://github.com/rust-lang/rust/pull/88563
- [x] Update edition guide and other documentation to reflect stability.
    - [x] Edition guide: https://github.com/rust-lang/edition-guide/pull/266
        - [x] Update edition-guide submodule: https://github.com/rust-lang/rust/pull/88563
    - [x] Check if there's any other documentation that needs updating

crlf0710:
cc #85811

HTG-YT:
May I help with the third item when it is needed to be implemented?

m-ou-se:
@HTG-YT Can you open that as a (draft) PR? Then we can review it so it's ready to be merged when we get to that point. Thanks!

HTG-YT:
Alright!

m-ou-se:
![](https://github.githubassets.com/images/icons/emoji/unicode/1f389.png)

PR #88100, opened 2021-08-17
Make Edition 2021 Stable

HTGAzureX1212:
An item of #87959.

This is an "on-demand" pull request, which means it will be merged when it is the right time to.

### DIFF
--- a/compiler/rustc_span/src/edition.rs
+++ b/compiler/rustc_span/src/edition.rs
@@ -32,7 +32,7 @@ pub const EDITION_NAME_LIST: &str = "2015|2018|2021";
 
 pub const DEFAULT_EDITION: Edition = Edition::Edition2015;
 
-pub const LATEST_STABLE_EDITION: Edition = Edition::Edition2018;
+pub const LATEST_STABLE_EDITION: Edition = Edition::Edition2021;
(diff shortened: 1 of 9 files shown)


ehuss:
I think https://github.com/rust-lang/rust/blob/master/src/doc/rustc/src/command-line-arguments.md#--edition-specify-the-edition-to-use also needs updating.

HTG-YT:
I think so.

Urgau:
Don't the constant [`LATEST_STABLE_EDITION`](https://github.com/rust-lang/rust/blob/9dd9ee55954d65e466811e23ebb56c799c05264b/compiler/rustc_span/src/edition.rs#L35) in `rust/compiler/rustc_span/src/edition.rs:35` also need to be changed to `Edition2021` ?

m-ou-se:
Yes. I've rebased the branch updated LATEST_STABLE_EDITION, and fixed another 2021-stabilization related fixme.

m-ou-se:
@bors p=10

This is blocking stabilization of Rust 2021 in cargo, which needs to happen asap.

ehuss:
@bors retry
apple failed, no logs

PR #88101, opened 2021-08-17
[WIP] parameterize `-C prefer-dynamic`

pnkfelix:
see also MCP https://github.com/rust-lang/compiler-team/issues/455

This PR generalizes `-C prefer-dynamic` so that one can specify a specific subset of the crates that should have the preferred-dynamic linkage.

Without this PR, one generally has to decide globally whether dynamic linkage will be preferred (via `-C prefer-dynamic=on`, or just `-C prefer-dynamic`) or if static linkage will be preferred (via `-C prefer-dynamic=off`, or just omitting the option entirely). The fact that this is a global choice means that one can get into tricky scenarios where, because of choices made by your crate dependencies for what output types they generate, *your* crate hits strange linkage errors. (For one example of this, see rust-lang/rust#82151.)

There are multiple ways to address the above problem, such as changing the crate dependencies to reduce the kinds of output crate types they generate (see e.g. https://github.com/alexkornitzer/hyper/commit/119904864b3dc91d77cf443af11f697821759336), or choosing `-C prefer-dynamic=no` to stop `rustc` from injecting linkage directives, and then adding on additional `RUSTFLAGS=-l<crate_name>` flags to explicitly link to the dynamic libraries you want.

This PR suggests a different way to address it: By allowing the user to specify a list of crates, `-C prefer-dynamic=crate1,crate2,...` they believe should be given preferred dynamic linkage.

Since `std` comes up as a common example here, `-C prefer-dynamic=std` is given special treatment. You can opt into arbitrary lists of crates via `-Z prefer-dynamic-subset`. Or you can enable the use of the singleton command line option `-C prefer-dynamic=std` via `-Z prefer-dynamic-std`.
 * My intent, in making separate `-Z` flags for these cases, is to fast track the stabilization of `-C prefer-dynamic=std`, because I think that will provide the biggest impact, and has relatively low risk (in terms of it not having potential for exposing sharp corners of rustc's internal heuristics for determining dynamic linkage). We can let the more general `-C prefer-dynamic=crate1,crate2,...` simmer; for example, we may want to make more changes to how chains of rlibs and dylibs are handled, and that may interact with the `prefer-dynamic-subset` feature.

### DIFF
--- a/compiler/rustc_metadata/src/dependency_format.rs
+++ b/compiler/rustc_metadata/src/dependency_format.rs
@@ -55,14 +55,16 @@ use crate::creader::CStore;
 
 use rustc_data_structures::fx::FxHashMap;
 use rustc_hir::def_id::{CrateNum, LOCAL_CRATE};
-use rustc_middle::middle::cstore::CrateDepKind;
 use rustc_middle::middle::cstore::LinkagePreference::{self, RequireDynamic, RequireStatic};
+use rustc_middle::middle::cstore::{CrateDepKind, CrateSource};
 use rustc_middle::middle::dependency_format::{Dependencies, DependencyList, Linkage};
 use rustc_middle::ty::TyCtxt;
 use rustc_session::config::CrateType;
 use rustc_target::spec::PanicStrategy;
 
 type RevDeps = FxHashMap<(CrateNum, LinkagePreference), Vec<CrateNum>>;
+
+use rustc_data_structures::sync::Lrc;
 
 crate fn calculate(tcx: TyCtxt<'_>) -> Dependencies {
     tcx.sess
@@ -135,7 +137,7 @@ fn calculate_type(tcx: TyCtxt<'_>, ty: CrateType) -> DependencyList {
                 if tcx.dep_kind(cnum).macros_only() {
                     continue;
                 }
-                let src = tcx.used_crate_source(cnum);
+                let src: Lrc<CrateSource> = tcx.used_crate_source(cnum);
                 if src.rlib.is_some() {
                     continue;
                 }
@@ -160,7 +162,7 @@ fn calculate_type(tcx: TyCtxt<'_>, ty: CrateType) -> DependencyList {
             continue;
         }
         let name = tcx.crate_name(cnum);
-        let src = tcx.used_crate_source(cnum);
+        let src: Lrc<CrateSource> = tcx.used_crate_source(cnum);
         if src.dylib.is_some() {
             tracing::info!("calculate_type {} adding dylib: {}", tcx.crate_name(LOCAL_CRATE), name);
             add_library(tcx, cnum, RequireDynamic, &mut formats, &mut rev_deps, LOCAL_CRATE);
@@ -194,7 +196,7 @@ fn calculate_type(tcx: TyCtxt<'_>, ty: CrateType) -> DependencyList {
     // If the crate hasn't been included yet and it's not actually required
     // (e.g., it's an allocator) then we skip it here as well.
     for &cnum in tcx.crates(()).iter() {
-        let src = tcx.used_crate_source(cnum);
+        let src: Lrc<CrateSource> = tcx.used_crate_source(cnum);
         if src.dylib.is_none()
             && !formats.contains_key(&cnum)
             && tcx.dep_kind(cnum) == CrateDepKind::Explicit
@@ -224,7 +226,7 @@ fn calculate_type(tcx: TyCtxt<'_>, ty: CrateType) -> DependencyList {
     // making sure that everything is available in the requested format.
     for (cnum, kind) in ret.iter().enumerate() {
         let cnum = CrateNum::new(cnum + 1);
-        let src = tcx.used_crate_source(cnum);
+        let src: Lrc<CrateSource> = tcx.used_crate_source(cnum);
         match *kind {
             Linkage::NotLinked | Linkage::IncludedFromDylib => {}
             Linkage::Static if src.rlib.is_some() => continue,
--- a/compiler/rustc_metadata/src/dependency_format.rs
+++ b/compiler/rustc_metadata/src/dependency_format.rs
@@ -54,13 +54,15 @@
 use crate::creader::CStore;
 
 use rustc_data_structures::fx::FxHashMap;
-use rustc_hir::def_id::CrateNum;
+use rustc_hir::def_id::{CrateNum, LOCAL_CRATE};
 use rustc_middle::middle::cstore::CrateDepKind;
 use rustc_middle::middle::cstore::LinkagePreference::{self, RequireDynamic, RequireStatic};
 use rustc_middle::middle::dependency_format::{Dependencies, DependencyList, Linkage};
 use rustc_middle::ty::TyCtxt;
 use rustc_session::config::CrateType;
 use rustc_target::spec::PanicStrategy;
+
+type RevDeps = FxHashMap<(CrateNum, LinkagePreference), Vec<CrateNum>>;
 
 crate fn calculate(tcx: TyCtxt<'_>) -> Dependencies {
     tcx.sess
@@ -148,6 +150,7 @@ fn calculate_type(tcx: TyCtxt<'_>, ty: CrateType) -> DependencyList {
     }
 
     let mut formats = FxHashMap::default();
+    let mut rev_deps: RevDeps = FxHashMap::default();
 
     // Sweep all crates for found dylibs. Add all dylibs, as well as their
     // dependencies, ensuring there are no conflicts. The only valid case for a
@@ -159,12 +162,18 @@ fn calculate_type(tcx: TyCtxt<'_>, ty: CrateType) -> DependencyList {
         let name = tcx.crate_name(cnum);
         let src = tcx.used_crate_source(cnum);
         if src.dylib.is_some() {
-            tracing::info!("adding dylib: {}", name);
-            add_library(tcx, cnum, RequireDynamic, &mut formats);
+            tracing::info!("calculate_type {} adding dylib: {}", tcx.crate_name(LOCAL_CRATE), name);
+            add_library(tcx, cnum, RequireDynamic, &mut formats, &mut rev_deps, LOCAL_CRATE);
             let deps = tcx.dylib_dependency_formats(cnum);
             for &(depnum, style) in deps.iter() {
-                tracing::info!("adding {:?}: {}", style, tcx.crate_name(depnum));
-                add_library(tcx, depnum, style, &mut formats);
+                tracing::info!(
+                    "calculate_type {} adding dep of {}, {:?}: {}",
+                    tcx.crate_name(LOCAL_CRATE),
+                    name,
+                    style,
+                    tcx.crate_name(depnum)
+                );
+                add_library(tcx, depnum, style, &mut formats, &mut rev_deps, cnum);
             }
         }
     }
@@ -192,7 +201,7 @@ fn calculate_type(tcx: TyCtxt<'_>, ty: CrateType) -> DependencyList {
         {
             assert!(src.rlib.is_some() || src.rmeta.is_some());
             tracing::info!("adding staticlib: {}", tcx.crate_name(cnum));
-            add_library(tcx, cnum, RequireStatic, &mut formats);
+            add_library(tcx, cnum, RequireStatic, &mut formats, &mut rev_deps, LOCAL_CRATE);
             ret[cnum.as_usize() - 1] = Linkage::Static;
         }
     }
@@ -243,7 +252,10 @@ fn add_library(
     cnum: CrateNum,
     link: LinkagePreference,
     m: &mut FxHashMap<CrateNum, LinkagePreference>,
+    rev_deps: &mut RevDeps,
+    parent_cnum: CrateNum,
 ) {
+    rev_deps.entry((cnum, link)).or_insert(Vec::new()).push(parent_cnum);
     match m.get(&cnum) {
         Some(&link2) => {
             // If the linkages differ, then we'd have two copies of the library
@@ -254,11 +266,44 @@ fn add_library(
             // This error is probably a little obscure, but I imagine that it
             // can be refined over time.
             if link2 != link || link == RequireStatic {
+                let parent_names = |link| {
+                    let mut names = String::new();
+                    let mut saw_one = false;
+                    for name in rev_deps[&(cnum, link)].iter().map(|pcnum| tcx.crate_name(*pcnum)) {
+                        if saw_one {
+                            names.push_str(", ");
+                        }
+                        names.push_str(&format!("`{}`", name));
+                        saw_one = true;
+                    }
+                    names
+                };
+                let link_parent_names = parent_names(link);
+                let link2_parent_names = parent_names(link2);
+
+                let details = match (link2, link) {
+                    (RequireDynamic, RequireStatic) => format!(
+                        "previously required dynamic, via [{}], \
+                         and now also requires static, via [{}]",
+                        link2_parent_names, link_parent_names
+                    ),
+                    (RequireStatic, RequireDynamic) => format!(
+                        "previously required static, via [{}], \
+                         and now also requires dynamic, via [{}]",
+                        link2_parent_names, link_parent_names
+                    ),
+                    (RequireStatic, RequireStatic) => format!(
+                        "two static copies from multiple different locations, via [{}]",
+                        link_parent_names
+                    ),
+                    (RequireDynamic, RequireDynamic) => unreachable!("not a problem"),
+                };
                 tcx.sess
                     .struct_err(&format!(
                         "cannot satisfy dependencies so `{}` only \
-                                              shows up once",
-                        tcx.crate_name(cnum)
+                                              shows up once ({})",
+                        tcx.crate_name(cnum),
+                        details,
                     ))
                     .help(
                         "having upstream crates all available in one format \
--- a/compiler/rustc_metadata/src/dependency_format.rs
+++ b/compiler/rustc_metadata/src/dependency_format.rs
@@ -85,23 +85,43 @@ fn calculate_type(tcx: TyCtxt<'_>, ty: CrateType) -> DependencyList {
         return Vec::new();
     }
 
-    let preferred_linkage = match ty {
-        // Generating a dylib without `-C prefer-dynamic` means that we're going
-        // to try to eagerly statically link all dependencies. This is normally
-        // done for end-product dylibs, not intermediate products.
-        //
-        // Treat cdylibs similarly. If `-C prefer-dynamic` is set, the caller may
-        // be code-size conscious, but without it, it makes sense to statically
-        // link a cdylib.
-        CrateType::Dylib | CrateType::Cdylib if !sess.opts.cg.prefer_dynamic => Linkage::Static,
-        CrateType::Dylib | CrateType::Cdylib => Linkage::Dynamic,
+    let prefer_dynamic = &sess.opts.cg.prefer_dynamic;
 
-        // If the global prefer_dynamic switch is turned off, or the final
-        // executable will be statically linked, prefer static crate linkage.
-        CrateType::Executable if !sess.opts.cg.prefer_dynamic || sess.crt_static(Some(ty)) => {
-            Linkage::Static
+    // traverses all crates to see if any are in the prefer-dynamic set.
+    let prefer_dynamic_for_any_crate = || {
+        if prefer_dynamic.is_empty() {
+            // skip traversal if we know it cannot ever bear fruit.
+            false
+        } else {
+            tcx.crates(()).iter().any(|cnum| prefer_dynamic.contains_crate(tcx.crate_name(*cnum)))
         }
-        CrateType::Executable => Linkage::Dynamic,
+    };
+
+    let preferred_linkage = match ty {
+        // If `-C prefer-dynamic` is not set for any crate, then means that we're
+        // going to try to eagerly statically link all dependencies into the dylib.
+        // This is normally done for end-product dylibs, not intermediate products.
+        //
+        // Treat cdylibs similarly. If `-C prefer-dynamic` is set for any given
+        // crate, the caller may be code-size conscious, but without it, it
+        // makes sense to statically link a cdylib.
+        CrateType::Dylib | CrateType::Cdylib => {
+            if prefer_dynamic_for_any_crate() {
+                Linkage::Dynamic
+            } else {
+                Linkage::Static
+            }
+        }
+
+        // If `prefer_dynamic` is not set for any crate, or the final
+        // executable will be statically linked, prefer static crate linkage.
+        CrateType::Executable => {
+            if !prefer_dynamic_for_any_crate() || sess.crt_static(Some(ty)) {
+                Linkage::Static
+            } else {
+                Linkage::Dynamic
+            }
+        }
 
         // proc-macro crates are mostly cdylibs, but we also need metadata.
         CrateType::ProcMacro => Linkage::Static,
@@ -148,6 +168,11 @@ fn calculate_type(tcx: TyCtxt<'_>, ty: CrateType) -> DependencyList {
                 ));
             }
             return Vec::new();
+        } else {
+            tracing::info!(
+                "calculate_type {} attempt_static failed; falling through to dynamic linkage",
+                tcx.crate_name(LOCAL_CRATE)
+            );
         }
     }
 
@@ -163,7 +188,14 @@ fn calculate_type(tcx: TyCtxt<'_>, ty: CrateType) -> DependencyList {
         }
         let name = tcx.crate_name(cnum);
         let src: Lrc<CrateSource> = tcx.used_crate_source(cnum);
-        if src.dylib.is_some() {
+        // We should prefer the dylib, when available, if either:
+        // 1. the user has requested that dylib (or all dylibs) via `-C prefer-dynamic`, or
+        // 2. for backwards compatibility: if the prefer-dynamic subset is unset, then we *still*
+        //    favor dylibs here. This way, if static linking fails above, we might still hope to
+        //    succeed at linking here.
+        let prefer_dylib =
+            |name| -> bool { prefer_dynamic.is_unset() || prefer_dynamic.contains_crate(name) };
+        if src.dylib.is_some() && prefer_dylib(name) {
             tracing::info!("calculate_type {} adding dylib: {}", tcx.crate_name(LOCAL_CRATE), name);
             add_library(tcx, cnum, RequireDynamic, &mut formats, &mut rev_deps, LOCAL_CRATE);
             let deps = tcx.dylib_dependency_formats(cnum);
@@ -196,13 +228,16 @@ fn calculate_type(tcx: TyCtxt<'_>, ty: CrateType) -> DependencyList {
     // If the crate hasn't been included yet and it's not actually required
     // (e.g., it's an allocator) then we skip it here as well.
     for &cnum in tcx.crates(()).iter() {
+        let name = tcx.crate_name(cnum);
         let src: Lrc<CrateSource> = tcx.used_crate_source(cnum);
-        if src.dylib.is_none()
-            && !formats.contains_key(&cnum)
-            && tcx.dep_kind(cnum) == CrateDepKind::Explicit
-        {
+        let static_available = src.rlib.is_some() || src.rmeta.is_some();
+        let prefer_static =
+            src.dylib.is_none() || (static_available && !prefer_dynamic.contains_crate(name));
+        let missing = !formats.contains_key(&cnum);
+        let actually_required = tcx.dep_kind(cnum) == CrateDepKind::Explicit;
+        if prefer_static && missing && actually_required {
             assert!(src.rlib.is_some() || src.rmeta.is_some());
-            tracing::info!("adding staticlib: {}", tcx.crate_name(cnum));
+            tracing::info!("adding staticlib: {}", name);
             add_library(tcx, cnum, RequireStatic, &mut formats, &mut rev_deps, LOCAL_CRATE);
             ret[cnum.as_usize() - 1] = Linkage::Static;
         }
--- a/compiler/rustc_metadata/src/dependency_format.rs
+++ b/compiler/rustc_metadata/src/dependency_format.rs
@@ -46,10 +46,10 @@
 //! all dynamic. This isn't currently very well battle tested, so it will likely
 //! fall short in some use cases.
 //!
-//! Currently, there is no way to specify the preference of linkage with a
-//! particular library (other than a global dynamic/static switch).
-//! Additionally, the algorithm is geared towards finding *any* solution rather
-//! than finding a number of solutions (there are normally quite a few).
+//! The algorithm is geared towards finding *any* solution rather than finding a
+//! number of solutions (there are normally quite a few). One can specify the
+//! preference of linkage for a particular list of crates by specifying that
+//! list via `-C prefer-dynamic=crate1,crate2,...`.
 
 use crate::creader::CStore;
 
--- a/compiler/rustc_session/src/config.rs
+++ b/compiler/rustc_session/src/config.rs
@@ -63,6 +63,68 @@ pub enum CFGuard {
     Checks,
 }
 
+/// Handle `-C prefer-dynamic` flag: either its all, or its some explicit
+/// subset (potentially the empty set).
+#[derive(Clone, PartialEq, Hash, Debug)]
+pub enum PreferDynamicSet {
+    /// Specified by the absence of `-C prefer-dynamic`, or via an explicit `-C prefer-dynamic=...`
+    /// with value `n`, `no`, or `off`.
+    ///
+    /// Under this (default) behavior, the compiler first optimistically attempts to statically
+    /// link everything, but if that fails, then attempts to dynamically link *everything* with a
+    /// dylib path available.
+    Unset,
+
+    /// Specified via `-C prefer-dynamic` with no value. For backwards-compatibility, also
+    /// specified via `-C prefer-dynamic=...` with value `y`, `yes`, or `on`.
+    All,
+
+    /// Specified via `-C prefer-dynamic=crate,...`.
+    Set(BTreeSet<String>),
+}
+
+impl PreferDynamicSet {
+    pub fn unset() -> Self {
+        PreferDynamicSet::Unset
+    }
+
+    pub fn every_crate() -> Self {
+        PreferDynamicSet::All
+    }
+
+    pub fn subset(crates: impl Iterator<Item = impl ToString>) -> Self {
+        PreferDynamicSet::Set(crates.map(|x| x.to_string()).collect())
+    }
+
+    pub fn is_unset(&self) -> bool {
+        if let PreferDynamicSet::Unset = *self { true } else { false }
+    }
+
+    pub fn is_empty(&self) -> bool {
+        match self {
+            PreferDynamicSet::Unset => true,
+            PreferDynamicSet::All => false,
+            PreferDynamicSet::Set(s) => s.len() == 0,
+        }
+    }
+
+    pub fn is_non_empty(&self) -> bool {
+        match self {
+            PreferDynamicSet::Unset => false,
+            PreferDynamicSet::All => true,
+            PreferDynamicSet::Set(s) => s.len() > 0,
+        }
+    }
+
+    pub fn contains_crate(&self, crate_name: Symbol) -> bool {
+        match self {
+            PreferDynamicSet::Unset => false,
+            PreferDynamicSet::All => true,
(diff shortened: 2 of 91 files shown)


pnkfelix:
> The job **`mingw-check`** failed! Check out the build log: [(web)](https://github.com/rust-lang/rust/runs/3346181971) [(plain)](https://github.com/rust-lang/rust/commit/83f34999a07f179bb854d34445faeb8f07052457/checks/3346181971/logs)
> Click to see the possible cause of the failure (guessed by this bot)

No rest for the wicked.

(I got punished for adding type annotations of the form `Rc<CrateSource>` when the actual type was of `Lrc`, not `Rc`, which are sometimes, but not always, type aliases.)

pnkfelix:
Hmm, yeah I was worried about something like this:

```
1 error: linking with `cc` failed: exit status: 1
2    |
-    = note: "cc" "-m64" "$TEST_BUILD_DIR/dylibs/issue-82151-serverctl-prefdyn/issue-82151-serverctl-prefdyn.issue_82151_serverctl_prefdyn.72d5a704-cgu.0.rcgu.o" "-Wl,--as-needed" "-L" "$BUILD_DIR/x86_64-unknown-linux-gnu/native/rust-test-helpers" "-L" "$TEST_BUILD_DIR/dylibs/issue-82151-serverctl-prefdyn/auxiliary" "-L" "$BUILD_DIR/x86_64-unknown-linux-gnu/stage1/lib/rustlib/x86_64-unknown-linux-gnu/lib" "-L" "$TEST_BUILD_DIR/dylibs/issue-82151-serverctl-prefdyn/auxiliary" "-lshared" "-L" "$TEST_BUILD_DIR/dylibs/issue-82151-serverctl-prefdyn/auxiliary" "-lbar" "-Wl,--start-group" "-L" "$BUILD_DIR/x86_64-unknown-linux-gnu/stage1/lib/rustlib/x86_64-unknown-linux-gnu/lib" "-lstd-f2409c1bc0340a9e" "-Wl,--end-group" "-Wl,-Bstatic" "$BUILD_DIR/x86_64-unknown-linux-gnu/stage1/lib/rustlib/x86_64-unknown-linux-gnu/lib/libcompiler_builtins-bc417570076b2daa.rlib" "-Wl,-Bdynamic" "-lgcc_s" "-lutil" "-lrt" "-lpthread" "-lm" "-ldl" "-lc" "-Wl,--eh-frame-hdr" "-Wl,-znoexecstack" "-L" "$BUILD_DIR/x86_64-unknown-linux-gnu/stage1/lib/rustlib/x86_64-unknown-linux-gnu/lib" "-o" "$TEST_BUILD_DIR/dylibs/issue-82151-serverctl-prefdyn/issue-82151-serverctl-prefdyn" "-Wl,--gc-sections" "-pie" "-Wl,-zrelro" "-Wl,-znow" "-Wl,-O1" "-nodefaultlibs" "-Wl,-rpath,$ORIGIN/auxiliary" "-Wl,-rpath,$ORIGIN/../../../../stage1/lib/rustlib/x86_64-unknown-linux-gnu/lib" "-Wl,--enable-new-dtags"
-    = note: /usr/bin/ld: $TEST_BUILD_DIR/dylibs/issue-82151-serverctl-prefdyn/auxiliary/libshared.so: undefined reference to `bar::bar'
+    = note: "cc" "-m64" "$TEST_BUILD_DIR/dylibs/issue-82151-serverctl-prefdyn/issue-82151-serverctl-prefdyn.issue_82151_serverctl_prefdyn.72d5a704-cgu.0.rcgu.o" "-Wl,--as-needed" "-L" "$BUILD_DIR/x86_64-unknown-linux-gnu/native/rust-test-helpers" "-L" "$TEST_BUILD_DIR/dylibs/issue-82151-serverctl-prefdyn/auxiliary" "-L" "$BUILD_DIR/x86_64-unknown-linux-gnu/stage2/lib/rustlib/x86_64-unknown-linux-gnu/lib" "-L" "$TEST_BUILD_DIR/dylibs/issue-82151-serverctl-prefdyn/auxiliary" "-lshared" "-L" "$TEST_BUILD_DIR/dylibs/issue-82151-serverctl-prefdyn/auxiliary" "-lbar" "-Wl,--start-group" "-L" "$BUILD_DIR/x86_64-unknown-linux-gnu/stage2/lib/rustlib/x86_64-unknown-linux-gnu/lib" "-lstd-979615b1d869d792" "-Wl,--end-group" "-Wl,-Bstatic" "$BUILD_DIR/x86_64-unknown-linux-gnu/stage2/lib/rustlib/x86_64-unknown-linux-gnu/lib/libcompiler_builtins-52fe0581016e3544.rlib" "-Wl,-Bdynamic" "-lgcc_s" "-lutil" "-lrt" "-lpthread" "-lm" "-ldl" "-lc" "-Wl,--eh-frame-hdr" "-Wl,-znoexecstack" "-L" "$BUILD_DIR/x86_64-unknown-linux-gnu/stage2/lib/rustlib/x86_64-unknown-linux-gnu/lib" "-o" "$TEST_BUILD_DIR/dylibs/issue-82151-serverctl-prefdyn/issue-82151-serverctl-prefdyn" "-Wl,--gc-sections" "-pie" "-Wl,-zrelro" "-Wl,-znow" "-Wl,-O1" "-nodefaultlibs" "-Wl,-rpath,$ORIGIN/auxiliary" "-Wl,-rpath,$ORIGIN/../../../../stage2/lib/rustlib/x86_64-unknown-linux-gnu/lib" "-Wl,--enable-new-dtags"
+    = note: $TEST_BUILD_DIR/dylibs/issue-82151-serverctl-prefdyn/auxiliary/libshared.so: undefined reference to `bar::bar'
5            collect2: error: ld returned 1 exit status
Some tests failed in compiletest suite=ui mode=ui host=x86_64-unknown-linux-gnu target=x86_64-unknown-linux-gnu
6            
7    = help: some `extern` functions couldn't be found; some native libraries may need to be installed or have their path specified
```

i.e. the error output is tightly coupled to the linker's error formatting.

I'll go see if I can use one of the old coarser grain error-pattern checking systems. Hopefully those haven't been retired yet. :)

pnkfelix:
One thing still missing from this PR is a run-make test illustrating that the dynamically linked libraries *are* the set that we expect. The usual pattern for testing that, which you can see in existing run-make tests, is to run the binary successfully, then delete the dylib and attempt to re-run the binary, now *expecting* a runtime failure. I do plan to add some run-make tests in that vein to this PR; I just wanted to get this up and get some review feedback on it before I invested time on that.

pnkfelix:
> -    = note: "cc" "-m64" "$TEST_BUILD_DIR/dylibs/issue-82151-serverctl-prefdyn/issue-82151-serverctl-prefdyn.issue_82151_serverctl_prefdyn.72d5a704-cgu.0.rcgu.o" "-Wl,--as-needed" "-L" "$BUILD_DIR/x86_64-unknown-linux-gnu/native/rust-test-helpers" "-L" "$TEST_BUILD_DIR/dylibs/issue-82151-serverctl-prefdyn/auxiliary" "-L" "$BUILD_DIR/x86_64-unknown-linux-gnu/stage1/lib/rustlib/x86_64-unknown-linux-gnu/lib" "-L" "$TEST_BUILD_DIR/dylibs/issue-82151-serverctl-prefdyn/auxiliary" "-lshared" "-L" "$TEST_BUILD_DIR/dylibs/issue-82151-serverctl-prefdyn/auxiliary" "-lbar" "-Wl,--start-group" "-L" "$BUILD_DIR/x86_64-unknown-linux-gnu/stage1/lib/rustlib/x86_64-unknown-linux-gnu/lib" "-lstd-*" "-Wl,--end-group" "-Wl,-Bstatic" "$BUILD_DIR/x86_64-unknown-linux-gnu/stage1/lib/rustlib/x86_64-unknown-linux-gnu/lib/libcompiler_builtins-*.rlib" "-Wl,-Bdynamic" "-lgcc_s" "-lutil" "-lrt" "-lpthread" "-lm" "-ldl" "-lc" "-Wl,--eh-frame-hdr" "-Wl,-znoexecstack" "-L" "$BUILD_DIR/x86_64-unknown-linux-gnu/stage1/lib/rustlib/x86_64-unknown-linux-gnu/lib" "-o" "$TEST_BUILD_DIR/dylibs/issue-82151-serverctl-prefdyn/issue-82151-serverctl-prefdyn" "-Wl,--gc-sections" "-pie" "-Wl,-zrelro" "-Wl,-znow" "-Wl,-O1" "-nodefaultlibs" "-Wl,-rpath,$ORIGIN/auxiliary" "-Wl,-rpath,$ORIGIN/../../../../stage1/lib/rustlib/x86_64-unknown-linux-gnu/lib" "-Wl,--enable-new-dtags"
> +    = note: "cc" "-m64" "$TEST_BUILD_DIR/dylibs/issue-82151-serverctl-prefdyn/issue-82151-serverctl-prefdyn.issue_82151_serverctl_prefdyn.72d5a704-cgu.0.rcgu.o" "-Wl,--as-needed" "-L" "$BUILD_DIR/x86_64-unknown-linux-gnu/native/rust-test-helpers" "-L" "$TEST_BUILD_DIR/dylibs/issue-82151-serverctl-prefdyn/auxiliary" "-L" "$BUILD_DIR/x86_64-unknown-linux-gnu/stage2/lib/rustlib/x86_64-unknown-linux-gnu/lib" "-L" "$TEST_BUILD_DIR/dylibs/issue-82151-serverctl-prefdyn/auxiliary" "-lshared" "-L" "$TEST_BUILD_DIR/dylibs/issue-82151-serverctl-prefdyn/auxiliary" "-lbar" "-Wl,--start-group" "-L" "$BUILD_DIR/x86_64-unknown-linux-gnu/stage2/lib/rustlib/x86_64-unknown-linux-gnu/lib" "-lstd-*" "-Wl,--end-group" "-Wl,-Bstatic" "$BUILD_DIR/x86_64-unknown-linux-gnu/stage2/lib/rustlib/x86_64-unknown-linux-gnu/lib/libcompiler_builtins-*.rlib" "-Wl,-Bdynamic" "-lgcc_s" "-lutil" "-lrt" "-lpthread" "-lm" "-ldl" "-lc" "-Wl,--eh-frame-hdr" "-Wl,-znoexecstack" "-L" "$BUILD_DIR/x86_64-unknown-linux-gnu/stage2/lib/rustlib/x86_64-unknown-linux-gnu/lib" "-o" "$TEST_BUILD_DIR/dylibs/issue-82151-serverctl-prefdyn/issue-82151-serverctl-prefdyn" "-Wl,--gc-sections" "-pie" "-Wl,-zrelro" "-Wl,-znow" "-Wl,-O1" "-nodefaultlibs" "-Wl,-rpath,$ORIGIN/auxiliary" "-Wl,-rpath,$ORIGIN/../../../../stage2/lib/rustlib/x86_64-unknown-linux-gnu/lib" "-Wl,--enable-new-dtags"

Okay I'll need to normalize the `stage{1,2}` in the stderr output too.

compiler/rustc_session/src/config.rs
```diff
@@ -63,6 +63,68 @@ pub enum CFGuard {
...
+
+    pub fn is_non_empty(&self) -> bool {
+        match self {
+            PreferDynamicSet::Unset => false,
+            PreferDynamicSet::All => true,
+            PreferDynamicSet::Set(s) => s.len() > 0,
+        }
+    }
+
+    pub fn contains_crate(&self, crate_name: Symbol) -> bool {
```
bjorn3, line 119: I think using just the crate name is not a good idea. It doesn't allow distinguishing between two different versions of the same library. Ideally it would accept the StableCrateId, but that is not accessible outside of rustc without depending on nightly (`-Zls`?) or implementation details. Maybe also allow accepting a dylib path or combination of crate name, if it has bin as crate type in addition to dylib and the list of `-Cmetadata` arguments. The `StableCrateId` is reproducible using just this information. I don't think we can't change what it depends on as that would change which crates can be linked into the same crate graph and which can't: https://github.com/rust-lang/rust/blob/master/compiler/rustc_span/src/def_id.rs#L148-L175

compiler/rustc_session/src/options.rs
```diff
@@ -835,6 +837,38 @@ mod parse {
...
+            // `-C prefer-dynamic=...` with `y`/`yes`/`on` is a synonym, for backwards
+            // compatibility.
+            Some("y") | Some("yes") | Some("on") | None => PreferDynamicSet::every_crate(),
+
+            // `-C prefer-dynamic=crate1,crate2,...` gives *just* crate1, crate2, ... preferred
+            // dynamic linkage.
+            Some(s) => {
+                let v = s.split(',').map(|s| s.to_string()).collect();
+                PreferDynamicSet::Set(v)
+            }
```
bjorn3, line 866: Maybe it should be a different argument as the new option is not a preference, but mandatory?

src/doc/rustc/src/codegen-options/index.md
```diff
@@ -402,6 +402,13 @@ linkage. This flag takes one of the following values:
 
 * `y`, `yes`, `on`, or no value: use dynamic linking.
 * `n`, `no`, or `off`: use static linking (the default).
+* A comma-separated list of crate names `crate1,crate2,...`: prefer dynamic
+  linking, but solely for the indicated crates. For example, to statically link
+  everything except `std`, use `-C prefer-dynamic=std`.
+
+Note that the explicit list of crate names variant of this flag is gated behind
+`-Zprefer-dynamic-subset`; as a special case, one can enable the handling of
+the single crate `std` as shown in the example above via `-Zprefer-dynamic-std`.
```
bjorn3, line 411: Those options don't exist at least as of this commit.

Edit: these are introduced in the next commit.

compiler/rustc_metadata/src/dependency_format.rs
```diff
@@ -85,23 +85,43 @@ fn calculate_type(tcx: TyCtxt<'_>, ty: CrateType) -> DependencyList {
...
 
+    let prefer_dynamic = &sess.opts.cg.prefer_dynamic;
+
+    // traverses all crates to see if any are in the prefer-dynamic set.
+    let prefer_dynamic_for_any_crate = || {
+        if prefer_dynamic.is_empty() {
+            // skip traversal if we know it cannot ever bear fruit.
+            false
+        } else {
+            tcx.crates(()).iter().any(|cnum| prefer_dynamic.contains_crate(tcx.crate_name(*cnum)))
```
bjorn3, line 96: This should skip non-linkable crates like proc-macros and their dependencies I think.

src/test/ui/dylibs/diamonds_core.rs
```diff
@@ -0,0 +1,21 @@
...
+    // Sanity-check: getting back non-trival values from addr functions
+    assert_ne!(z::s_addr(), z::t_addr());
+    assert_ne!(z::s_m_addr(), z::s_m_i_addr());
+    assert_ne!(z::s_m_i_addr(), z::s_m_j_addr());
+    assert_ne!(z::s_m_i_addr(), z::s_m_i_a_addr());
+}
+
+pub fn check_linked_function_equivalence() {
+    // Check that the linked functions are the same code by comparing their
+    // underlying addresses.
```
bjorn3, line 13: I wouldn't be surprised if the dynamic linker would resolve each symbol with the same name to the same address at runtime even if it is duplicated between dylibs.

bjorn3 (changes requested):

jackh726:
🤦‍♀️

src/test/ui/dylibs/diamonds_core.rs
pnkfelix: Possibly... I have certainly been thinking that the testing methodology here might be flawed. I'm still musing about the best way to write these tests (i.e. to detect unwanted duplication of static libraries across dynamic libraries being loaded).

I'm still planning to invest some effort into a `run-make` test, to check the most basic scenario of deleting a dynamic library and [observing that this injects](https://github.com/rust-lang/rust/pull/88101#issuecomment-900483063) a failure to run the binary. But that isn't quite the same check that I'm trying to put in here.

compiler/rustc_session/src/config.rs
pnkfelix: Would some encoding of crate name and an optionally-supplied version be sufficient to satisfy your criteria here? I'm thinking in the common cases, a single crate name is exactly what an end user will want to write down.

compiler/rustc_session/src/options.rs
pnkfelix: Hmm. My intention was that it would still be a preference, in the sense that if you do `-C prefer-dynamic=crate1`, but only a static library for `crate1` is available, then linkage will still proceed using the static library.

I probably should add a test illustrating that.

compiler/rustc_metadata/src/dependency_format.rs
pnkfelix: Okay I'll look into that.

pnkfelix:
> The job **`x86_64-gnu-llvm-10`** failed! Check out the build log: [(web)](https://github.com/rust-lang/rust/runs/3353922146) [(plain)](https://github.com/rust-lang/rust/commit/3c0b75db085060f7827b3e56683a4ee7c5dc4c64/checks/3353922146/logs)
> Click to see the possible cause of the failure (guessed by this bot)
> 
> ```
> ..............i...............................................ii.................................... 12100/12173
> .........................................................................
> failures:
> 
> ---- [ui] ui/dylibs/issue-82151-serverctl-prefdyn.rs stdout ----
> 
> 
> 1 error: linking with `cc` failed: exit status: 1
> 2    |
> -    = note: "cc" "-m64" "$TEST_BUILD_DIR/dylibs/issue-82151-serverctl-prefdyn/issue-82151-serverctl-prefdyn.issue_82151_serverctl_prefdyn.72d5a704-cgu.0.rcgu.o" "-Wl,--as-needed" "-L" "$BUILD_DIR/x86_64-unknown-linux-gnu/native/rust-test-helpers" "-L" "$TEST_BUILD_DIR/dylibs/issue-82151-serverctl-prefdyn/auxiliary" "-L" "$BUILD_DIR/x86_64-unknown-linux-gnu/stageN/lib/rustlib/x86_64-unknown-linux-gnu/lib" "-L" "$TEST_BUILD_DIR/dylibs/issue-82151-serverctl-prefdyn/auxiliary" "-lshared" "-L" "$TEST_BUILD_DIR/dylibs/issue-82151-serverctl-prefdyn/auxiliary" "-lbar" "-Wl,--start-group" "-L" "$BUILD_DIR/x86_64-unknown-linux-gnu/stageN/lib/rustlib/x86_64-unknown-linux-gnu/lib" "-lstd-*" "-Wl,--end-group" "-Wl,-Bstatic" "$BUILD_DIR/x86_64-unknown-linux-gnu/stageN/lib/rustlib/x86_64-unknown-linux-gnu/lib/libcompiler_builtins-*.rlib" "-Wl,-Bdynamic" "-lgcc_s" "-lutil" "-lrt" "-lpthread" "-lm" "-ldl" "-lc" "-Wl,--eh-frame-hdr" "-Wl,-znoexecstack" "-L" "$BUILD_DIR/x86_64-unknown-linux-gnu/stageN/lib/rustlib/x86_64-unknown-linux-gnu/lib" "-o" "$TEST_BUILD_DIR/dylibs/issue-82151-serverctl-prefdyn/issue-82151-serverctl-prefdyn" "-Wl,--gc-sections" "-pie" "-Wl,-zrelro" "-Wl,-znow" "-Wl,-O1" "-nodefaultlibs" "-Wl,-rpath,$ORIGIN/auxiliary" "-Wl,-rpath,$ORIGIN/../../../../stageN/lib/rustlib/x86_64-unknown-linux-gnu/lib" "-Wl,--enable-new-dtags"
> +    = note: "cc" "-m32" "$TEST_BUILD_DIR/dylibs/issue-82151-serverctl-prefdyn/issue-82151-serverctl-prefdyn.issue_82151_serverctl_prefdyn.72d5a704-cgu.0.rcgu.o" "-Wl,--as-needed" "-L" "$BUILD_DIR/i686-unknown-linux-gnu/native/rust-test-helpers" "-L" "$TEST_BUILD_DIR/dylibs/issue-82151-serverctl-prefdyn/auxiliary" "-L" "$BUILD_DIR/x86_64-unknown-linux-gnu/stageN/lib/rustlib/i686-unknown-linux-gnu/lib" "-L" "$TEST_BUILD_DIR/dylibs/issue-82151-serverctl-prefdyn/auxiliary" "-lshared" "-L" "$TEST_BUILD_DIR/dylibs/issue-82151-serverctl-prefdyn/auxiliary" "-lbar" "-Wl,--start-group" "-L" "$BUILD_DIR/x86_64-unknown-linux-gnu/stageN/lib/rustlib/i686-unknown-linux-gnu/lib" "-lstd-*" "-Wl,--end-group" "-Wl,-Bstatic" "$BUILD_DIR/x86_64-unknown-linux-gnu/stageN/lib/rustlib/i686-unknown-linux-gnu/lib/libcompiler_builtins-*.rlib" "-Wl,-Bdynamic" "-lgcc_s" "-lutil" "-lrt" "-lpthread" "-lm" "-ldl" "-lc" "-Wl,--eh-frame-hdr" "-Wl,-znoexecstack" "-L" "$BUILD_DIR/x86_64-unknown-linux-gnu/stageN/lib/rustlib/i686-unknown-linux-gnu/lib" "-o" "$TEST_BUILD_DIR/dylibs/issue-82151-serverctl-prefdyn/issue-82151-serverctl-prefdyn" "-Wl,--gc-sections" "-pie" "-Wl,-zrelro" "-Wl,-znow" "-Wl,-O1" "-nodefaultlibs" "-Wl,-rpath,$ORIGIN/auxiliary" "-Wl,-rpath,$ORIGIN/../../../../stageN/lib/rustlib/i686-unknown-linux-gnu/lib" "-Wl,--enable-new-dtags"
> Some tests failed in compiletest suite=ui mode=ui host=x86_64-unknown-linux-gnu target=i686-unknown-linux-gnu
> 4    = note: undefined reference to `bar::bar'
> 5            collect2: error: ld returned 1 exit status
> 
> 
> The actual stderr differed from the expected stderr.
> The actual stderr differed from the expected stderr.
> Actual stderr saved to /checkout/obj/build/x86_64-unknown-linux-gnu/test/ui/dylibs/issue-82151-serverctl-prefdyn/issue-82151-serverctl-prefdyn.stderr
> To update references, rerun the tests and pass the `--bless` flag
> To only update this specific test, also pass `--test-args dylibs/issue-82151-serverctl-prefdyn.rs`
> error: 1 errors occurred comparing output.
> status: exit status: 1
> status: exit status: 1
> command: "/checkout/obj/build/x86_64-unknown-linux-gnu/stage2/bin/rustc" "/checkout/src/test/ui/dylibs/issue-82151-serverctl-prefdyn.rs" "-Zthreads=1" "--target=i686-unknown-linux-gnu" "--error-format" "json" "-Ccodegen-units=1" "-Zui-testing" "-Zdeduplicate-diagnostics=no" "-Zemit-future-incompat-report" "--out-dir" "/checkout/obj/build/x86_64-unknown-linux-gnu/test/ui/dylibs/issue-82151-serverctl-prefdyn" "-A" "unused" "-Crpath" "-O" "-Cdebuginfo=0" "-Lnative=/checkout/obj/build/i686-unknown-linux-gnu/native/rust-test-helpers" "-Clinker=cc" "-C" "prefer-dynamic" "-L" "/checkout/obj/build/x86_64-unknown-linux-gnu/test/ui/dylibs/issue-82151-serverctl-prefdyn/auxiliary"
> ------------------------------------------
> 
> ------------------------------------------
> stderr:
> stderr:
> ------------------------------------------
> error: linking with `cc` failed: exit status: 1
>    |
>    = note: "cc" "-m32" "/checkout/obj/build/x86_64-unknown-linux-gnu/test/ui/dylibs/issue-82151-serverctl-prefdyn/issue-82151-serverctl-prefdyn.issue_82151_serverctl_prefdyn.72d5a704-cgu.0.rcgu.o" "-Wl,--as-needed" "-L" "/checkout/obj/build/i686-unknown-linux-gnu/native/rust-test-helpers" "-L" "/checkout/obj/build/x86_64-unknown-linux-gnu/test/ui/dylibs/issue-82151-serverctl-prefdyn/auxiliary" "-L" "/checkout/obj/build/x86_64-unknown-linux-gnu/stage2/lib/rustlib/i686-unknown-linux-gnu/lib" "-L" "/checkout/obj/build/x86_64-unknown-linux-gnu/test/ui/dylibs/issue-82151-serverctl-prefdyn/auxiliary" "-lshared" "-L" "/checkout/obj/build/x86_64-unknown-linux-gnu/test/ui/dylibs/issue-82151-serverctl-prefdyn/auxiliary" "-lbar" "-Wl,--start-group" "-L" "/checkout/obj/build/x86_64-unknown-linux-gnu/stage2/lib/rustlib/i686-unknown-linux-gnu/lib" "-lstd-16743a84aa11e6b8" "-Wl,--end-group" "-Wl,-Bstatic" "/checkout/obj/build/x86_64-unknown-linux-gnu/stage2/lib/rustlib/i686-unknown-linux-gnu/lib/libcompiler_builtins-0507ce3ad78674c6.rlib" "-Wl,-Bdynamic" "-lgcc_s" "-lutil" "-lrt" "-lpthread" "-lm" "-ldl" "-lc" "-Wl,--eh-frame-hdr" "-Wl,-znoexecstack" "-L" "/checkout/obj/build/x86_64-unknown-linux-gnu/stage2/lib/rustlib/i686-unknown-linux-gnu/lib" "-o" "/checkout/obj/build/x86_64-unknown-linux-gnu/test/ui/dylibs/issue-82151-serverctl-prefdyn/issue-82151-serverctl-prefdyn" "-Wl,--gc-sections" "-pie" "-Wl,-zrelro" "-Wl,-znow" "-Wl,-O1" "-nodefaultlibs" "-Wl,-rpath,$ORIGIN/auxiliary" "-Wl,-rpath,$ORIGIN/../../../../stage2/lib/rustlib/i686-unknown-linux-gnu/lib" "-Wl,--enable-new-dtags"
>    = note: /checkout/obj/build/x86_64-unknown-linux-gnu/test/ui/dylibs/issue-82151-serverctl-prefdyn/auxiliary/libshared.so: undefined reference to `bar::bar'
>            collect2: error: ld returned 1 exit status
>            
>    = help: some `extern` functions couldn't be found; some native libraries may need to be installed or have their path specified
>    = note: use the `-l` flag to specify native libraries to link
>    = note: use the `cargo:rustc-link-lib` directive to specify the native libraries to link with Cargo (see https://doc.rust-lang.org/cargo/reference/build-scripts.html#cargorustc-link-libkindname)
> error: aborting due to previous error
> 
> 
> ------------------------------------------
> ------------------------------------------
> 
> 
> 
> failures:
>     [ui] ui/dylibs/issue-82151-serverctl-prefdyn.rs
> test result: FAILED. 12012 passed; 1 failed; 160 ignored; 0 measured; 0 filtered out; finished in 75.49s
> 
> 
> 
> 
> command did not execute successfully: "/checkout/obj/build/x86_64-unknown-linux-gnu/stage0-tools-bin/compiletest" "--compile-lib-path" "/checkout/obj/build/x86_64-unknown-linux-gnu/stage2/lib" "--run-lib-path" "/checkout/obj/build/x86_64-unknown-linux-gnu/stage2/lib/rustlib/i686-unknown-linux-gnu/lib" "--rustc-path" "/checkout/obj/build/x86_64-unknown-linux-gnu/stage2/bin/rustc" "--src-base" "/checkout/src/test/ui" "--build-base" "/checkout/obj/build/x86_64-unknown-linux-gnu/test/ui" "--stage-id" "stage2-i686-unknown-linux-gnu" "--suite" "ui" "--mode" "ui" "--target" "i686-unknown-linux-gnu" "--host" "x86_64-unknown-linux-gnu" "--llvm-filecheck" "/usr/lib/llvm-10/bin/FileCheck" "--pass" "check" "--nodejs" "/usr/bin/node" "--linker" "cc" "--host-rustcflags" "-Crpath -O -Cdebuginfo=0  -Lnative=/checkout/obj/build/x86_64-unknown-linux-gnu/native/rust-test-helpers" "--target-rustcflags" "-Crpath -O -Cdebuginfo=0  -Lnative=/checkout/obj/build/i686-unknown-linux-gnu/native/rust-test-helpers" "--docck-python" "/usr/bin/python3" "--lldb-python" "/usr/bin/python3" "--gdb" "/usr/bin/gdb" "--quiet" "--llvm-version" "10.0.0" "--llvm-components" "aarch64 aarch64asmparser aarch64codegen aarch64desc aarch64disassembler aarch64info aarch64utils aggressiveinstcombine all all-targets amdgpu amdgpuasmparser amdgpucodegen amdgpudesc amdgpudisassembler amdgpuinfo amdgpuutils analysis arm armasmparser armcodegen armdesc armdisassembler arminfo armutils asmparser asmprinter avr avrasmparser avrcodegen avrdesc avrdisassembler avrinfo binaryformat bitreader bitstreamreader bitwriter bpf bpfasmparser bpfcodegen bpfdesc bpfdisassembler bpfinfo cfguard codegen core coroutines coverage debuginfocodeview debuginfodwarf debuginfogsym debuginfomsf debuginfopdb demangle dlltooldriver dwarflinker engine executionengine frontendopenmp fuzzmutate globalisel hexagon hexagonasmparser hexagoncodegen hexagondesc hexagondisassembler hexagoninfo instcombine instrumentation interpreter ipo irreader jitlink lanai lanaiasmparser lanaicodegen lanaidesc lanaidisassembler lanaiinfo libdriver lineeditor linker lto mc mca mcdisassembler mcjit mcparser mips mipsasmparser mipscodegen mipsdesc mipsdisassembler mipsinfo mirparser msp430 msp430asmparser msp430codegen msp430desc msp430disassembler msp430info native nativecodegen nvptx nvptxcodegen nvptxdesc nvptxinfo objcarcopts object objectyaml option orcerror orcjit passes perfjitevents powerpc powerpcasmparser powerpccodegen powerpcdesc powerpcdisassembler powerpcinfo profiledata remarks riscv riscvasmparser riscvcodegen riscvdesc riscvdisassembler riscvinfo riscvutils runtimedyld scalaropts selectiondag sparc sparcasmparser sparccodegen sparcdesc sparcdisassembler sparcinfo support symbolize systemz systemzasmparser systemzcodegen systemzdesc systemzdisassembler systemzinfo tablegen target textapi transformutils vectorize webassembly webassemblyasmparser webassemblycodegen webassemblydesc webassemblydisassembler webassemblyinfo windowsmanifest x86 x86asmparser x86codegen x86desc x86disassembler x86info x86utils xcore xcorecodegen xcoredesc xcoredisassembler xcoreinfo xray" "--cc" "" "--cxx" "" "--cflags" "" "--adb-path" "adb" "--adb-test-dir" "/data/tmp/work" "--android-cross-path" "" "--channel" "nightly" "--color" "always"
> 
> 
> Build completed unsuccessfully in 0:01:17
> ```

This test is a nightmare. I'm going to change the normalization header to just trivialize the command invocation entirely.

compiler/rustc_session/src/config.rs
bjorn3: Yeah, that would work I guess. Cargo could specify everything and a user just the crate name.

src/test/ui/dylibs/diamonds_core.rs
bjorn3: You could do an objdump of dylibs to check which symbols are defined and imported. Imported symbols use 0 as specified address.

compiler/rustc_session/src/options.rs
pnkfelix: On reflection, I do think I must have messed up something in my logic that ends up making *something* a requirement rather than a preference. Specifically, I'm working on a run-make test analogous to https://github.com/rust-lang/rust/blob/master/src/test/run-make-fulldeps/dylib-chain/Makefile
but where I choose variations of whether dylibs on the chain have both rlib and dylib, or just rlib, or just dylib.

The problem I'm seeing is that in some scenarios using `-C prefer-dynamic=crate,...`, a crate that is solely provided as a dylib is not getting linked in unless I include it in the above list. That wasn't my intention here: My intention was that the preference is meant to give guidance when the compiler must make a choice, not act as an assertion about what it expects to be present.

pnkfelix: Update 2: Hmm maybe I was wrong about the above: In general `.so` files don't have enough info to be statically linked. But then I'm really confused about some of the behavior I'm seeing. Still looking.

pnkfelix: Okay, I have now made a run-make test that I think illustrates the current behavior of the system, including cases where rlib's end up depending on dylibs.

I don't think pre-existing behavior is necessarily *intuitive*, but my main goal right now is to address missing use cases, **not** to try to figure out what the ideal behavior is for all cases and potentially inject breaking changes to support that idealized behavior.

pnkfelix:
It is worth explicitly noting: The `run-make` test I added in 21778cba8e73fa28849ad31d0839df3b413947fe has six variant tests (v1 through v6), but only one of them, v5, actually exercises `-Z prefer-dynamic-subset` and `-Z prefer-dynamic-std`.

The other five tests, v1-4 and 6, all work without debugflags, and represent pre-existing behavior.

I have manually confirmed that those five tests all work on the compiler before the changes in this PR. In other words, the test is largely codifying the pre-existing behavior of `-C prefer-dynamic` (for better or for worse; I am still working out what the mental model of a user for the pre-existing behavior is meant to be here).

src/test/ui/dylibs/diamonds_core.rs
pnkfelix: Okay I added a run-make test in 21778cba8e73fa28849ad31d0839df3b413947fe; it does two things:

 1. it uses the pre-existing pattern of building, running to success, then deleting dylibs, and checking that a re-run attempt fails. (I think this testing methodology is fine, as far as it goes.)
 2. it introduces a new pattern of having methods *in* the compiled code that differ depending on whether one is linked to the rlib or to the dylib. This is controlled via `cfg` flags. This way, we can directly determine which object code has been linked and loaded when we finally run the program.

src/test/run-make-fulldeps/dylib-subset/Makefile
```diff
@@ -0,0 +1,176 @@
...
+BIN_TYPE=--crate-type bin --cfg bin_type
+PREFER_DYNAMIC_INPUT=-Zprefer-dynamic-subset -Cprefer-dynamic=std,m2,m3,$(1)
+
+# How to read this command sequence, for people unversed in shell scripting:
+#
+# I have tried to write useful comments as `echo` statements. Their text is usually
+# delimited by single quotes ('') rathern than double quotes ("") so that I can use
+# backticks (``) in the notes without it being interpreted as a shell invocation.
+#
+# `COMMAND && exit 1 && exit 0` is how we run a command and say that we're
```
pnkfelix, line 32:
```suggestion
# `COMMAND && exit 1 || exit 0` is how we run a command and say that we're
```

petrochenkov:
Another idea is to use the fact that `rustc` "unifies" crates pointing to the same file and assigns a single `CrateNum` to them, regardless of whether they are linked directly or indirectly.

So to refer to an indirect dependency we can add a new direct dependency referring to our crate of interested (by path `--extern name=PATH`, or by name `--extern name`), and then use an `--extern` modifier `--extern prefer-dynamic:name=PATH` (similar to existing `--extern priv:name=PATH` and `--extern noprelude:name=PATH`) to attach the "prefer-dynamic" property to the corresponding `CrateNum`.

pnkfelix:
> Another idea is to use the fact that `rustc` "unifies" crates pointing to the same file and assigns a single `CrateNum` to them, regardless of whether they are linked directly or indirectly. (That's why `std::vec::Vec` used directly from `std` and reexported through some other crate refer to the same type and `DefId`.)
> 
> So to refer to an indirect dependency we can add a new direct dependency referring to our crate of interested (by path `--extern name=PATH`, or by name `--extern name`), and then use an `--extern` modifier `--extern prefer-dynamic:name=PATH` (similar to existing `--extern priv:name=PATH` and `--extern noprelude:name=PATH`) to attach the "prefer-dynamic" property to the corresponding `CrateNum`.
 
Sorry, @petrochenkov , I am not fully understanding this.

When you say "to refer to an indirect dependency we can add a new direct dependency referring to our crate of interested", do you mean "refer" as in what internal data structures that `rustc` uses to denote an indirect dependency?

Or are you somehow suggesting a different way of invoking the existing `rustc` compiler to work around this problem? (An interpetation that doesn't seem plausible...)

petrochenkov:
@pnkfelix 
>When you say "to refer to an indirect dependency we can add a new direct dependency referring to our crate of interested", do you mean "refer" as in what internal data structures that rustc uses to denote an indirect dependency?

Maybe, but I don't understand what you mean here.

>Or are you somehow suggesting a different way of invoking the existing rustc compiler to work around this problem? (An interpretation that doesn't seem plausible...)

Certainly not this.

By "refer to a crate X" I mean "somehow name that crate from command line".
Once we can name it we can attach various properties like `prefer-dynamic` to it.

pnkfelix:
I'm going to revisit my approach here, maybe try to work out something that better leverages things like PR #93901, (assuming that lands).

PR #88111, opened 2021-08-17
Add background-color on clickable definitions in source code

GuillaumeGomez:
Someone suggested to add a decoration on clickable elements in the source code pages:

Screenshot from 2021-08-17 14-49-39
Screenshot from 2021-08-17 14-49-47
Screenshot from 2021-08-17 14-49-52

The idea is to not disturb the reading while telling the reader "you can click on this one", which is why it's not a text decoration.

r? @Nemo157

### DIFF
--- a/src/librustdoc/html/static/css/themes/ayu.css
+++ b/src/librustdoc/html/static/css/themes/ayu.css
@@ -216,6 +216,9 @@ nav.main .separator {
 a {
 	color: #c5c5c5;
 }
+body.source .example-wrap pre.rust a {
+	background: #c5c5c5;
+}
 
 .docblock:not(.type-decl) a:not(.srclink):not(.test-arrow),
 .docblock-short a:not(.srclink):not(.test-arrow), .item-info a,
--- a/src/librustdoc/html/static/css/themes/dark.css
+++ b/src/librustdoc/html/static/css/themes/dark.css
@@ -177,6 +177,9 @@ nav.main .separator {
 a {
 	color: #ddd;
 }
+body.source .example-wrap pre.rust a {
+	background: #333;
+}
 
 .docblock:not(.type-decl) a:not(.srclink):not(.test-arrow),
 .docblock-short a:not(.srclink):not(.test-arrow), .item-info a,
--- a/src/librustdoc/html/static/css/themes/light.css
+++ b/src/librustdoc/html/static/css/themes/light.css
@@ -172,6 +172,9 @@ nav.main .separator {
 a {
 	color: #000;
 }
+body.source .example-wrap pre.rust a {
+	background: #eee;
+}
 
 .docblock:not(.type-decl) a:not(.srclink):not(.test-arrow),
 .docblock-short a:not(.srclink):not(.test-arrow), .item-info a,


src/librustdoc/html/static/css/themes/ayu.css
```diff
@@ -216,6 +216,9 @@ nav.main .separator {
 a {
 	color: #c5c5c5;
 }
+body.source .example-wrap pre.rust a {
+	background: ##c5c5c5;
```
jhpratt, line 220:
```suggestion
	background: #c5c5c5;
```

jhpratt (commented):
LGTM with this fix.

Manishearth:
Not sure of the exact choice of colors but overall I think this is a good idea. @jsha ?

CraftSpider:
I like the idea, I think the middle theme looks okay, top and bottom themes should probably make the colors stand out a bit more, it's hard to tell which parts are the links at a glance.

GuillaumeGomez:
I actually found the middle standing out too much. :laughing:

CraftSpider:
That's fair I guess, I just worry that if they don't stand out enough they won't really help indicate which ones are links. My primary skills aren't in visual design though, I'll leave that decision to those who know more :P

jhpratt:
I've already discussed this with Guillaume on Zulip…be happy that there's _something_ happening, because what I proposed was along the lines of the middle and he didn't care for it 😛

GuillaumeGomez:
@Manishearth @jsha won't be back until October, since this is an improvement, even though maybe not perfect, I think it's fine to merge now. I'll add this to my "list of things to discuss when Jacob is back" haha.

Thanks everyone for the comments!

@bors: r=jhpratt

camelid:
> I actually found the middle standing out too much. 😆

I like the idea overall, though I also feel that the Ayu stands out too much. Would that one look different with https://github.com/rust-lang/rust/pull/88111#discussion_r690414594 applied?

We could also consider styling it like Haskell's equivalent to Rustdoc, Haddock. They use a faint underline, and then add background color when the link is hovered:

image

image

I like how it's very subtle in Haddock, so it doesn't affect the reading experience, but it's strong enough to think, "Hey, I bet I can click that link!"

IIUC, nearly all of the identifiers should be clickable in the generated source view, so we just want to give a hint that the links are clickable without distracting from the reading view.

GuillaumeGomez:
> I like the idea overall, though I also feel that the Ayu stands out too much. Would that one look different with #88111 (comment) applied?

No, I badly copied into the code, I modified using the web tools in the browser.

As for your suggestion, I also like it. Hummm... Damn, don't know what to do...

Manishearth:
@GuillaumeGomez seems fair to me!

jhpratt:
Yeah, I'm not saying it couldn't be better, but imo something is better than nothing as a stopgap. Having it be non-intrusive is the issue.

GuillaumeGomez:
@camelid Do you mind opening an issue with your suggestion please?

camelid:
> @camelid Do you mind opening an issue with your suggestion please?

@GuillaumeGomez done: #88120.

PR #88121, opened 2021-08-17
Improve errors for recursive type aliases

### LINKED ISSUE #17539: "error: illegal recursive type; insert an enum or struct in the cycle, if this is desired" not very helpful
I've tried to write code that would work like this:

``` rust
let mut f = some_fn;  // many other some_fns also exist

loop {
    f = f()
}
```

To make coding some_fn and its friends easier, I wanted to create a type:

``` rust
type SomeFn = fn() -> SomeFn;
```

Unfortunately, this results in some type-theoretic error message that isn't really understandable by mere mortals:

```
<anon>:1:15: 1:29 error: illegal recursive type; insert an enum or struct in the cycle, if this is desired
<anon>:1 type SomeFn = fn() -> SomeFn;
                       ^~~~~~~~~~~~~~
```

It's not at all clear what 'the cycle' is and where to insert an enum or struct. Turns out that if you know what the compiler is talking about, the fix is simple:

``` rust
struct SomeFn(fn() -> SomeFn);

// ergonomy suffers
fn fn1() -> SomeFn { println!("1"); SomeFn(fn2) }
fn fn2() -> SomeFn { println!("2"); SomeFn(fn1) }

fn main() {
    let mut f = SomeFn(fn1);
    f = { let SomeFn(ff) = f; ff() };
    f = { let SomeFn(ff) = f; ff() };
}
```

Output:

```
1
2
```

But I was very confused by the message and tried a contraption like this instead:

``` rust
type SomeFn = fn() -> Option<SomeFn>;
```

A hint in the error message that points the programmer to the right direction would be very helpful. Ergonomy of the solution is also questionable, but at least it compiles.

camelid:
Fixes #17539.

### DIFF
--- a/compiler/rustc_middle/src/query/mod.rs
+++ b/compiler/rustc_middle/src/query/mod.rs
@@ -116,7 +116,18 @@ rustc_queries! {
 
     /// Records the type of every item.
     query type_of(key: DefId) -> Ty<'tcx> {
-        desc { |tcx| "computing type of `{}`", tcx.def_path_str(key) }
+        desc { |tcx|
+            "{action} `{path}`",
+            action = {
+                use rustc_hir::def::DefKind;
+                match tcx.def_kind(key) {
+                    DefKind::TyAlias => "expanding type alias",
+                    DefKind::TraitAlias => "expanding trait alias",
+                    _ => "computing type of",
+                }
+            },
+            path = tcx.def_path_str(key),
+        }
         cache_on_disk_if { key.is_local() }
     }
 
--- a/compiler/rustc_query_impl/src/keys.rs
+++ b/compiler/rustc_query_impl/src/keys.rs
@@ -20,6 +20,12 @@ pub trait Key {
     /// In the event that a cycle occurs, if no explicit span has been
     /// given for a query with key `self`, what span should we use?
     fn default_span(&self, tcx: TyCtxt<'_>) -> Span;
+
+    /// If the key is a [`DefId`] or `DefId`--equivalent, return that `DefId`.
+    /// Otherwise, return `None`.
+    fn key_as_def_id(&self) -> Option<DefId> {
+        None
+    }
 }
 
 impl Key for () {
@@ -95,6 +101,9 @@ impl Key for LocalDefId {
     fn default_span(&self, tcx: TyCtxt<'_>) -> Span {
         self.to_def_id().default_span(tcx)
     }
+    fn key_as_def_id(&self) -> Option<DefId> {
+        Some(self.to_def_id())
+    }
 }
 
 impl Key for DefId {
@@ -104,6 +113,10 @@ impl Key for DefId {
     }
     fn default_span(&self, tcx: TyCtxt<'_>) -> Span {
         tcx.def_span(*self)
+    }
+    #[inline(always)]
+    fn key_as_def_id(&self) -> Option<DefId> {
+        Some(*self)
     }
 }
 
--- a/compiler/rustc_query_impl/src/keys.rs
+++ b/compiler/rustc_query_impl/src/keys.rs
@@ -178,6 +178,10 @@ impl Key for (DefId, Option<Ident>) {
     fn default_span(&self, tcx: TyCtxt<'_>) -> Span {
         tcx.def_span(self.0)
     }
+    #[inline(always)]
+    fn key_as_def_id(&self) -> Option<DefId> {
+        Some(self.0)
+    }
 }
 
 impl Key for (DefId, LocalDefId, Ident) {
--- a/compiler/rustc_query_impl/src/plumbing.rs
+++ b/compiler/rustc_query_impl/src/plumbing.rs
@@ -338,10 +338,12 @@ macro_rules! define_queries {
                     Some(key.default_span(*tcx))
                 };
                 let def_id = key.key_as_def_id();
-                let def_kind = def_id.map(|def_id| {
-                    let def_kind = tcx.def_kind(def_id);
-                    $crate::util::def_kind_to_simple_def_kind(def_kind)
-                });
+                let def_kind = def_id
+                    .and_then(|def_id| def_id.as_local())
+                    // Use `tcx.hir().opt_def_kind()` to reduce the chance of
+                    // accidentally triggering an infinite query loop.
+                    .and_then(|def_id| tcx.hir().opt_def_kind(def_id))
+                    .map(|def_kind| $crate::util::def_kind_to_simple_def_kind(def_kind));
                 let hash = || {
                     let mut hcx = tcx.create_stable_hashing_context();
                     let mut hasher = StableHasher::new();
--- a/compiler/rustc_query_impl/src/plumbing.rs
+++ b/compiler/rustc_query_impl/src/plumbing.rs
@@ -337,6 +337,11 @@ macro_rules! define_queries {
                 } else {
                     Some(key.default_span(*tcx))
                 };
+                let def_id = key.key_as_def_id();
+                let def_kind = def_id.map(|def_id| {
+                    let def_kind = tcx.def_kind(def_id);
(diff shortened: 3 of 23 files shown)


src/test/ui/infinite/infinite-vec-type-recursion.stderr
```diff
@@ -1,10 +1,10 @@
-error[E0391]: cycle detected when computing type of `X`
+error[E0391]: cycle detected when expanding type alias `X`
   --> $DIR/infinite-vec-type-recursion.rs:1:14
    |
 LL | type X = Vec<X>;
    |              ^
    |
-   = note: ...which again requires computing type of `X`, completing the cycle
+   = note: ...which immediately again requires expanding type alias `X`
```
camelid, line 7: It'd be really nice to add a note that says something like "type aliases cannot be recursive" or "type aliases cannot reference themselves", but I'm not sure how to do that in a non-hacky way.

camelid: And to also add a hint to use a struct or enum as #17539 suggests.

estebank: I think that the least hacky way of accomplishing that would be to make a `DefId` part of the `QueryStackFrame`. I don't know what the perf implications might be of doing so, but that will be the most foolproof way of approaching this. Alternatively I can think of making a substring search for `type alias`, but I don't condone such things except as last resort, even though I've been guilty of them in the past 😅.

compiler/rustc_query_system/src/query/job.rs
```diff
@@ -591,10 +591,14 @@ pub(crate) fn report_cycle<'a>(
         err.span_note(span, &format!("...which requires {}...", query.description));
     }
```
estebank, line 592: I'm intrigued if the iteration above doesn't need to be reverted to be correct 🤔 (no need to address now)

compiler/rustc_query_system/src/query/job.rs
```diff
@@ -591,10 +591,14 @@ pub(crate) fn report_cycle<'a>(
...
 
-    err.note(&format!(
-        "...which again requires {}, completing the cycle",
-        stack[0].query.description
-    ));
+    if stack.len() == 1 {
+        err.note(&format!("...which immediately again requires {}", stack[0].query.description));
```
estebank, line 595: I'm not sure I'm a fan of the new wording. Something about "immediately again" sounds weird to me, too... "lyrical", but I'm ESL so it might be fine.

estebank (approved):
I'm ok with the approach. Do you want to try out to expand the query with the `DefId` to see what the impact of that is code and perf wise?

compiler/rustc_query_system/src/query/job.rs
camelid: It sounds a bit off to me as well. I thought of doing `...which immediately requires {} again` or `...which immediately requires again {}`, but those sounded weirder to me. Do you have an idea for a better phrasing?

src/test/ui/infinite/infinite-vec-type-recursion.stderr
camelid: Would it be better to just store the `DefKind`? I think that could work too.

compiler/rustc_query_system/src/query/job.rs
estebank: I think the first of the options you mention is the most grammatically correct.

src/test/ui/infinite/infinite-vec-type-recursion.stderr
estebank: I think it should work. Can you try it out?

camelid:
> Do you want to try out to expand the query with the `DefId` to see what the impact of that is code and perf wise?

I tried doing that but then realized that I'd need to execute a query (to get the `DefKind`) in the midst of reporting a query cycle error, which doesn't look to be possible with the current code architecture. I also tried recording `DefKind`, but that's defined in `rustc_hir`, which I think means that we'd get a cyclic dependency graph (it's kind of funny—I'd get a dependency cycle while working on improving type alias cycle errors...). I could move `DefKind` to `rustc_span::def_id` (or maybe a new module `rustc_span::def`), which I *think* would solve the problem. What do you think; should I move it?

camelid:
> Yet another approach could be to have a mini-`DefKind` that tracks less information but can be implemented without any other dependencies. That seems like maybe the best approach. What do you think?

Ok, that worked! I'll probably finish up and then push up my changes tomorrow or this weekend.

camelid:
I'm working on getting it working for trait aliases too, and then I think this should be ready for review.

compiler/rustc_middle/src/query/mod.rs
```diff
@@ -116,7 +116,18 @@ rustc_queries! {
...
     /// Records the type of every item.
     query type_of(key: DefId) -> Ty<'tcx> {
-        desc { |tcx| "computing type of `{}`", tcx.def_path_str(key) }
+        desc { |tcx|
+            "{action} `{path}`",
+            action = {
+                use rustc_hir::def::DefKind;
+                match tcx.def_kind(key) {
+                    DefKind::TyAlias => "expanding type alias",
+                    DefKind::TraitAlias => "expanding trait alias",
```
camelid, line 125: It doesn't look like this code path ever occurs. Should I remove this case?

compiler/rustc_query_impl/src/keys.rs
```diff
@@ -20,6 +20,12 @@ pub trait Key {
     /// In the event that a cycle occurs, if no explicit span has been
     /// given for a query with key `self`, what span should we use?
     fn default_span(&self, tcx: TyCtxt<'_>) -> Span;
+
+    /// If the key is a [`DefId`] or `DefId`--equivalent, return that `DefId`.
+    /// Otherwise, return `None`.
+    fn key_as_def_id(&self) -> Option<DefId> {
```
camelid, line 26: I named this `key_as_def_id` to avoid conflict with other method names. (E.g., `CrateNum::as_def_id()`.) Alternatively, I could use fully-qualified syntax at the call site, which might be a good idea anyway. What do you think?

camelid:
Once we think the logic is finished, we should do a perf run as you suggested earlier.

camelid:
Rebased (just to keep this up to date).

compiler/rustc_middle/src/query/mod.rs
estebank: Let's leave it for completion.

compiler/rustc_query_impl/src/keys.rs
estebank: Sounds reasonable to me

estebank (approved):
r=me. The only thing I'd want us to have is the docstring on the new enum. Feel free to ignore the other drive by comments.

compiler/rustc_query_impl/src/plumbing.rs
```diff
@@ -337,6 +337,11 @@ macro_rules! define_queries {
                 } else {
                     Some(key.default_span(*tcx))
                 };
+                let def_id = key.key_as_def_id();
+                let def_kind = def_id.map(|def_id| {
+                    let def_kind = tcx.def_kind(def_id);
```
camelid, line 342: I just realized something: since this code is executed as part of running a query, should I change it to

```suggestion
                    let def_kind = tcx.hir().opt_def_kind(def_id).unwrap();
```

in order to ensure that `def_kind` doesn't run a query that could send the compiler into a loop?

camelid:
Perf looks good, I just have one question, and then this should be ready to be merged!

compiler/rustc_query_impl/src/plumbing.rs
estebank: Yeah, that makes sense. Good catch.

camelid:
Ok, updated it to use the HIR map directly!

PR #85835, opened 2021-05-30
Implement Extend<(A, B)> for (Extend<A>, Extend<B>)

Seppel3210:
I oriented myself at the implementation of `Iterator::unzip` and also rewrote the impl in terms of `(A, B)::extend` after that.

Since (A, B) now also implements Extend we could also mention in the documentation of unzip that it can do "nested unzipping" (you could unzip `Iterator<Item=(A, (B, C))>` into `(Vec<A>, (Vec<B>, Vec<C>))` for example) but I'm not sure of that so I'm asking here 🙂 

(P.S. I saw a couple of people asking if there is an unzip3 but there isn't. So this could be a way to get equivalent functionality)

PR #88122, opened 2021-08-17
Fix example in `Extend<(A, B)>` impl

Seppel3210:
After looking over the examples in my last PR (#85835) on doc.rust-lang.org/nightly I realized that the example didn't actually show what I wanted it to show 😅 
So here's the better example


dtolnay (approved):
Great! Thank you.

PR #88123, opened 2021-08-17
Make spans for tuple patterns in E0023 more precise

### LINKED ISSUE #86307: `E0023` should have smaller, more targeted spans
Currently, when a tuple pattern has a different number of fields than a tuple variant, the output is like the following:

```
error[E0023]: this pattern has 2 fields, but the corresponding tuple variant has 1 field
   --> compiler/rustc_middle/src/ty/print/pretty.rs:634:27
    |
634 |                         | ty::PredicateKind::ImplicitSizedTrait(pred, _) =
    |                           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ expected 1 field, found 2
    | 
   ::: compiler/rustc_middle/src/ty/mod.rs:440:5
    |
440 |     ImplicitSizedTrait(TraitPredicate<'tcx>),
    |     ---------------------------------------- tuple variant defined here
```

Ideally, the primary span would be for the individual fields, with a secondary span pointing at the pattern path. As a stretch goal, when pointing at the definition we'd point at the fields in the same way, instead of the whole variant (but this isn't absolutely necessary).

camelid:
As suggested in #86307. Closes #86307.

r? @estebank

### DIFF
--- a/compiler/rustc_hir/src/hir.rs
+++ b/compiler/rustc_hir/src/hir.rs
@@ -3182,6 +3182,20 @@ pub enum Node<'hir> {
 }
 
 impl<'hir> Node<'hir> {
+    /// Get the identifier of this `Node`, if applicable.
+    ///
+    /// # Edge cases
+    ///
+    /// Calling `.ident()` on a [`Node::Ctor`] will return `None`
+    /// because `Ctor`s do not have identifiers themselves.
+    /// Instead, call `.ident()` on the parent struct/variant, like so:
+    ///
+    /// ```ignore (illustrative)
+    /// ctor
+    ///     .ctor_hir_id()
+    ///     .and_then(|ctor_id| tcx.hir().find(tcx.hir().get_parent_node(ctor_id)))
+    ///     .and_then(|parent| parent.ident())
+    /// ```
     pub fn ident(&self) -> Option<Ident> {
         match self {
             Node::TraitItem(TraitItem { ident, .. })
@@ -3190,8 +3204,25 @@ impl<'hir> Node<'hir> {
             | Node::Field(FieldDef { ident, .. })
             | Node::Variant(Variant { ident, .. })
             | Node::MacroDef(MacroDef { ident, .. })
-            | Node::Item(Item { ident, .. }) => Some(*ident),
-            _ => None,
+            | Node::Item(Item { ident, .. })
+            | Node::PathSegment(PathSegment { ident, .. }) => Some(*ident),
+            Node::Lifetime(lt) => Some(lt.name.ident()),
+            Node::GenericParam(p) => Some(p.name.ident()),
+            Node::Param(..)
+            | Node::AnonConst(..)
+            | Node::Expr(..)
+            | Node::Stmt(..)
+            | Node::Block(..)
+            | Node::Ctor(..)
+            | Node::Pat(..)
+            | Node::Binding(..)
+            | Node::Arm(..)
+            | Node::Local(..)
+            | Node::Visibility(..)
+            | Node::Crate(..)
+            | Node::Ty(..)
+            | Node::TraitRef(..)
+            | Node::Infer(..) => None,
         }
     }
 
--- a/src/test/ui/destructuring-assignment/tuple_struct_destructure_fail.stderr
+++ b/src/test/ui/destructuring-assignment/tuple_struct_destructure_fail.stderr
@@ -21,7 +21,7 @@ LL | struct TupleStruct<S, T>(S, T);
    |                          -  - tuple struct has 2 fields
 ...
 LL |     TupleStruct(a, a, b) = TupleStruct(1, 2);
-   |     ----------- ^  ^  ^ expected 2 fields, found 3
+   |                 ^  ^  ^ expected 2 fields, found 3
 
 error[E0023]: this pattern has 1 field, but the corresponding tuple struct has 2 fields
   --> $DIR/tuple_struct_destructure_fail.rs:32:17
@@ -30,7 +30,7 @@ LL | struct TupleStruct<S, T>(S, T);
    |                          -  - tuple struct has 2 fields
 ...
 LL |     TupleStruct(_) = TupleStruct(1, 2);
-   |     ----------- ^ expected 2 fields, found 1
+   |                 ^ expected 2 fields, found 1
    |
 help: use `_` to explicitly ignore each field
    |
@@ -48,7 +48,7 @@ LL |     SingleVariant(S, T)
    |                   -  - tuple variant has 2 fields
 ...
 LL |     Enum::SingleVariant(a, a, b) = Enum::SingleVariant(1, 2);
-   |     ------------------- ^  ^  ^ expected 2 fields, found 3
+   |                         ^  ^  ^ expected 2 fields, found 3
 
 error[E0023]: this pattern has 1 field, but the corresponding tuple variant has 2 fields
   --> $DIR/tuple_struct_destructure_fail.rs:36:25
@@ -57,7 +57,7 @@ LL |     SingleVariant(S, T)
    |                   -  - tuple variant has 2 fields
 ...
 LL |     Enum::SingleVariant(_) = Enum::SingleVariant(1, 2);
-   |     ------------------- ^ expected 2 fields, found 1
+   |                         ^ expected 2 fields, found 1
    |
 help: use `_` to explicitly ignore each field
    |
--- a/src/test/ui/destructuring-assignment/tuple_struct_destructure_fail.stderr
+++ b/src/test/ui/destructuring-assignment/tuple_struct_destructure_fail.stderr
@@ -15,22 +15,26 @@ LL |     Enum::SingleVariant(a, .., b, ..) = Enum::SingleVariant(0, 1);
    |                            previously used here
 
 error[E0023]: this pattern has 3 fields, but the corresponding tuple struct has 2 fields
-  --> $DIR/tuple_struct_destructure_fail.rs:30:5
+  --> $DIR/tuple_struct_destructure_fail.rs:30:16
    |
 LL | struct TupleStruct<S, T>(S, T);
    | ------------------------------- tuple struct defined here
 ...
 LL |     TupleStruct(a, a, b) = TupleStruct(1, 2);
-   |     ^^^^^^^^^^^^^^^^^^^^ expected 2 fields, found 3
+   |     -----------^^^^^^^^^ expected 2 fields, found 3
+   |     |
+   |     this tuple struct
 
 error[E0023]: this pattern has 1 field, but the corresponding tuple struct has 2 fields
-  --> $DIR/tuple_struct_destructure_fail.rs:32:5
+  --> $DIR/tuple_struct_destructure_fail.rs:32:16
    |
 LL | struct TupleStruct<S, T>(S, T);
    | ------------------------------- tuple struct defined here
 ...
 LL |     TupleStruct(_) = TupleStruct(1, 2);
-   |     ^^^^^^^^^^^^^^ expected 2 fields, found 1
+   |     -----------^^^ expected 2 fields, found 1
+   |     |
+   |     this tuple struct
    |
 help: use `_` to explicitly ignore each field
    |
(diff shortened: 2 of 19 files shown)


src/test/ui/destructuring-assignment/tuple_struct_destructure_fail.stderr
```diff
@@ -15,22 +15,26 @@ LL |     Enum::SingleVariant(a, .., b, ..) = Enum::SingleVariant(0, 1);
    |                            previously used here
 
 error[E0023]: this pattern has 3 fields, but the corresponding tuple struct has 2 fields
-  --> $DIR/tuple_struct_destructure_fail.rs:30:5
+  --> $DIR/tuple_struct_destructure_fail.rs:30:16
```
camelid, line 18: What's causing this change?

src/test/ui/destructuring-assignment/tuple_struct_destructure_fail.stderr
```diff
@@ -15,22 +15,26 @@ LL |     Enum::SingleVariant(a, .., b, ..) = Enum::SingleVariant(0, 1);
...
    |
 LL | struct TupleStruct<S, T>(S, T);
    | ------------------------------- tuple struct defined here
 ...
 LL |     TupleStruct(a, a, b) = TupleStruct(1, 2);
-   |     ^^^^^^^^^^^^^^^^^^^^ expected 2 fields, found 3
+   |     -----------^^^^^^^^^ expected 2 fields, found 3
+   |     |
+   |     this tuple struct
```
camelid, line 26: This is what I understood from

> with a secondary span pointing at the pattern path.

Is this what you had in mind? I'm wondering if could be confusing because it makes it sound a bit like this is where the tuple struct/variant is defined. Is there a way to add a secondary span *without* a label?

src/test/ui/destructuring-assignment/tuple_struct_destructure_fail.stderr
camelid: Huh, actually the old span seems incorrect and the new one is right. I'm very confused...

camelid: Oops, I just realized I was misinterpreting the column number as a line number—that explains it.

src/test/ui/destructuring-assignment/tuple_struct_destructure_fail.stderr
estebank: > Is this what you had in mind?

Yes, but with one further tweak:

```
LL |     TupleStruct(a, a, b) = TupleStruct(1, 2);
   |     ----------- ^  ^  ^ expected 2 fields, found 3
```

To do that, instead of taking the whole span and removing the beginning, you need to do something like

```rust
let field_spans: MultiSpan = fields.iter().map(|f| f.span).collect::<Vec<_>>().into();
```

> Is there a way to add a secondary span without a label?

`err.span_label(span, "");` 😅

estebank: It'd be also great if we could also do

```
LL | struct TupleStruct<S, T>(S, T);
   |        -----------       -  - tuple struct has 2 fields
   |        |
   |        tuple struct defined here
```

compiler/rustc_hir/src/hir.rs
```diff
@@ -3190,8 +3204,25 @@ impl<'hir> Node<'hir> {
             | Node::Field(FieldDef { ident, .. })
             | Node::Variant(Variant { ident, .. })
             | Node::MacroDef(MacroDef { ident, .. })
-            | Node::Item(Item { ident, .. }) => Some(*ident),
-            _ => None,
+            | Node::Item(Item { ident, .. })
+            | Node::PathSegment(PathSegment { ident, .. }) => Some(*ident),
+            Node::Lifetime(lt) => Some(lt.name.ident()),
+            Node::GenericParam(p) => Some(p.name.ident()),
```
camelid, line 3210: These should be the only changes to behavior in `Node::ident()`.

src/test/ui/pattern/pat-tuple-overfield.stderr
```diff
@@ -22,24 +171,79 @@ LL |         (1, 2, .., 3, 4) => {}
...
+   |     -- tuple struct has 5 fields
+...
+LL |         M(1, 2, 3, 4, 5, 6) => {}
+   |         - ^  ^  ^  ^  ^  ^ expected 5 fields, found 6
+
+error[E0023]: this pattern has 1 field, but the corresponding tuple struct has 0 fields
+  --> $DIR/pat-tuple-overfield.rs:45:12
+   |
+LL | struct Z1();
+   |          --- tuple struct has 0 fields
```
camelid, line 214: How should I fix the spans in cases like this?

src/test/ui/pattern/pat-tuple-underfield.stderr
```diff
@@ -40,13 +40,13 @@ LL |         S(..) => {}
...
-  --> $DIR/pat-tuple-underfield.rs:20:9
+  --> $DIR/pat-tuple-underfield.rs:20:10
    |
 LL | struct S(i32, f32);
-   | ------------------- tuple struct defined here
+   |          ---  --- tuple struct has 2 fields
 ...
 LL |         S() => {}
-   |         ^^^ expected 2 fields, found 0
+   |         -^^ expected 2 fields, found 0
```
camelid, line 49: The spans might look a little fun in these cases if the user is using non-standard formatting. I think it would like this:

```
LL |         S () => {}
   |         -^^^ expected 2 fields, found 0
```

In case of no fields, should we just use `^` for the whole pattern? (Perhaps that might be a solution to https://github.com/rust-lang/rust/pull/88123#r693390008 as well.)

src/test/ui/pattern/pat-tuple-underfield.stderr
```diff
@@ -58,27 +58,27 @@ LL |         S(..) => {}
...
-  --> $DIR/pat-tuple-underfield.rs:27:9
+  --> $DIR/pat-tuple-underfield.rs:27:14
    |
 LL |     S(i32, f32),
-   |     ----------- tuple variant defined here
+   |       ---  --- tuple variant has 2 fields
 ...
 LL |         E::S(x) => {}
-   |         ^^^^^^^ expected 2 fields, found 1
+   |         ---- ^ expected 2 fields, found 1
```
camelid, line 67: I'm wondering if we should remove the secondary span for `E::S`. I feel like it is kind of distracting.

src/test/ui/pattern/pat-tuple-field-count-cross.rs
```diff
@@ -0,0 +1,57 @@
...
+
+fn main() {
+    match Z0 {
+        Z0() => {} //~ ERROR expected tuple struct or tuple variant, found unit struct `Z0`
+        Z0(x) => {} //~ ERROR expected tuple struct or tuple variant, found unit struct `Z0`
+    }
+    match Z1() {
+        Z1 => {} //~ ERROR match bindings cannot shadow tuple structs
+        Z1(x) => {} //~ ERROR this pattern has 1 field, but the corresponding tuple struct has 0 fields
+    }
```
camelid, line 15: I can remove these and similar test cases if you like since there is an existing _non_-cross test for them. They clutter up the stderr somewhat.

src/test/ui/pattern/pat-tuple-field-count-cross.stderr
```diff
@@ -0,0 +1,536 @@
...
+   |
+LL | / pub struct M(
+LL | |     pub u8,
+   | |     ------
+LL | |     pub u8,
+   | |     ------
+LL | |     pub u8,
+   | |     ------ tuple struct has 3 fields
+LL | | );
+   | |__- tuple struct defined here
```
camelid, line 214: As you can see, in the cross-crate case, it falls back to highlighting the whole definition and not just the ident.

src/test/ui/typeck/struct-enum-wrong-args.stderr
```diff
@@ -29,6 +29,12 @@ LL |     let _ = Wrapper();
    |             ^^^^^^^-- supplied 0 arguments
    |             |
    |             expected 1 argument
+   |
+note: tuple struct defined here
+  --> $DIR/struct-enum-wrong-args.rs:2:8
+   |
+LL | struct Wrapper(i32);
+   |        ^^^^^^^
```
camelid, line 37: I think this changed because I fixed `def_ident_span`...? I guess it's a nice unexpected improvement :)

src/test/ui/pattern/pat-tuple-underfield.stderr
estebank: Yeah, that'd be reasonable. We do similar things elsewhere (special case the 0-case to point at the whole thing to make it easier to read).

src/test/ui/pattern/pat-tuple-underfield.stderr
estebank: I'm ok with removing it cases like this. We can check whether the spans are all in the same line to handle cases like

```rust
E::S(
    x,
    y,
) => {}
```
So that in that case we actually point at `E::S` to provide context that otherwise would be missing.

src/test/ui/pattern/pat-tuple-field-count-cross.stderr
estebank: That's fine. Not great, but not terrible either.

compiler/rustc_hir/src/hir.rs
```diff
@@ -3182,6 +3182,20 @@ pub enum Node<'hir> {
...
+    /// Calling `.ident()` on a [`Node::Ctor`] will return `None`
+    /// because `Ctor`s do not have identifiers themselves.
+    /// Instead, call `.ident()` on the parent struct/variant, like so:
+    ///
+    /// ```ignore (illustrative)
+    /// ctor
+    ///     .ctor_hir_id()
+    ///     .and_then(|ctor_id| tcx.hir().find(tcx.hir().get_parent_node(ctor_id)))
+    ///     .and_then(|parent| parent.ident())
+    /// ```
```
estebank, line 3198: Should we make this part of the call to `ident`, requiring people pass in `tcx` so they don't have to worry?

src/test/ui/pattern/pat-tuple-field-count-cross.rs
estebank: Nah, it's fine. Let's be thorough.

src/test/ui/pattern/pat-tuple-overfield.stderr
```diff
@@ -1,5 +1,154 @@
...
+   |         ^^^^^^^^
+   |
+help: use this syntax instead
+   |
+LL |         Z0 => {}
+   |         ~~
+help: a tuple struct with a similar name exists
+   |
+LL |         Z1(_, _) => {}
+   |         ~~
```
estebank, line 68: It seems to me that we should suggest this only if the number of fields matches _or_ if nothing else is being suggested (but that seems out of scope for this PR).

src/test/ui/pattern/pat-tuple-overfield.stderr
estebank: Either point at the `()`, if that's easy to do, otherwise point at the whole item with the same label, it should be clear enough.

src/test/ui/typeck/struct-enum-wrong-args.stderr
estebank: Drive by fixes are the best fixes.

estebank (approved):
Have to say, love how this is shaping up 😍

compiler/rustc_hir/src/hir.rs
camelid: I don't think we have access to `rustc_middle` from `rustc_hir` because `rustc_middle` depends on `rustc_hir`. If you have a suggestion for how to do it, I'm happy to, but it might not be worth it for now.

estebank: Ok, fair enough. Don't worry about it :)

src/test/ui/pattern/pat-tuple-underfield.stderr
estebank: r=me after removing the secondary span label with no text in the pattern for cases where there's more than 0 fields, and changing the 0 field case to point at the whole pattern (like it used to).

PR #88128, opened 2021-08-18
Add needs-asm-support to more tests

cuviper:
These were found as test failures on s390x for RHEL and Fedora.


Mark-Simulacrum:
r=me when CI is passing / you're happy with it being likely to pass

PR #88129, opened 2021-08-18
Fix dataflow graphviz bug, make dataflow graphviz modules public

willcrichton:
I'm working on a rustc plugin that uses the dataflow framework for MIR analysis. I've found the graphviz utilities extremely helpful for debugging. However, I had to fork the compiler to expose them since they're currently private. I would appreciate if they could be made public so I can build against a nightly instead of a custom fork. Specifically, this PR:

* Makes public the `rustc_mir::dataflow::framework::graphviz` module.
* Makes public the `rustc_mir::util::pretty::write_mir_fn` function.

Here's a concrete example of how I'm using the graphviz module: https://github.com/willcrichton/flowistry/blob/97b843b8b06b4004fbb79b5fcfca3e33c7143bc0/src/slicing/mod.rs#L186-L203

Additionally, this PR fixes a small bug in the diff code that incorrectly shows the updated object as the old object.

r? @ecstatic-morse

### DIFF
--- a/compiler/rustc_mir/src/dataflow/framework/fmt.rs
+++ b/compiler/rustc_mir/src/dataflow/framework/fmt.rs
@@ -33,7 +33,7 @@ pub trait DebugWithContext<C>: Eq + fmt::Debug {
         }
 
         write!(f, "\u{001f}-")?;
-        self.fmt_with(ctxt, f)
+        old.fmt_with(ctxt, f)
     }
 }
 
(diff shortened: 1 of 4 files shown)


ecstatic-morse:
@bors r+ rollup

Compiler APIs have no stability guarantees, so I don't see a problem making these public. Thanks for the bug fix (I must have only been using the `BitSet` impl) and happy hacking!

PR #77716, opened 2020-10-08
Revert "Allow dynamic linking for iOS/tvOS targets."

francesca64:
This reverts PR #73516.

On macOS I compile static libs for iOS, automated using [cargo-mobile](https://github.com/BrainiumLLC/cargo-mobile), which has worked smoothly for the past 2 years. However, upon updating to Rust 1.46.0, I was no longer able to use Rust on iOS. I've bisected this to the PR referenced above.

For most projects tested, apps now immediately crash with a message like this:
```
dyld: Library not loaded: /Users/francesca/Projects/example/target/aarch64-apple-ios/debug/deps/libexample.dylib
  Referenced from: /private/var/containers/Bundle/Application/745912AF-A928-465C-B340-872BD1C9F368/example.app/example
  Reason: image not found
dyld: launch, loading dependent libraries
DYLD_LIBRARY_PATH=/usr/lib/system/introspection
DYLD_INSERT_LIBRARIES=/Developer/usr/lib/libBacktraceRecording.dylib:/Developer/usr/lib/libMainThreadChecker.dylib:/Developer/Library/PrivateFrameworks/DTDDISupport.framework/libViewDebuggerSupport.dylib:/Developer/Library/PrivateFrameworks/GPUTools.framework/libglInterpose.dylib:/usr/lib/libMTLCapture.dylib
```

This can be reproduced by using cargo-mobile to generate a winit example project, and then attempting to run it on an iOS device (`cargo mobile init && cargo apple open`).

In our projects that depend on DisplayLink, the build instead fails with a linker error:
```
= note: Undefined symbols for architecture arm64:
            "_CACurrentMediaTime", referenced from:
                display_link::ios::run_callback_ios10::hda81197ff46aedbd in libapp-4f0abc1d7684103f.rlib(app-4f0abc1d7684103f.40d4iro0yz1iy487.rcgu.o)
                display_link::ios::run_callback_pre_ios10::h91f085da19374320 in libapp-4f0abc1d7684103f.rlib(app-4f0abc1d7684103f.40d4iro0yz1iy487.rcgu.o)
          ld: symbol(s) not found for architecture arm64
```

After reverting the change to enable dynamic linking on iOS, everything works the same as it did on Rust 1.45.2 for me.

In the future, would it be possible for me to be pinged when iOS-related PRs are made? I work for a company that intends on using Rust on iOS in production, so I'd gladly provide testing.

cc @aspenluxxxy

PR #88130, opened 2021-08-18
Allow iOS dynamic linking

Torrekie:
This reverts PR #77716.
Since Apple LLVM actually supported compiling .dylibs on/to iOS targets, there is no reason to prevent dynamic linking on a platform that supports it.
aarch64-apple-ios as a Tier 2 target that was 'promised to be compiled', should be allowed to access such a basic feature. Once dynamic library are prohibited to be built, some common libraries such as [Cryptography](https://github.com/pyca/cryptography) would no longer able to run on iOS targets, it is theoretically feasible though.
I was trying to get [mitmproxy](https://www.mitmproxy.org) working on my iPhone 12, as dynamic libraries written in rust cannot be compiled to this target, causing a lot of problems when attempting to run open source projects on an iOS device which was quite capable.

### DIFF
--- a/compiler/rustc_target/src/spec/apple_sdk_base.rs
+++ b/compiler/rustc_target/src/spec/apple_sdk_base.rs
@@ -50,7 +50,7 @@ pub fn opts(os: &str, arch: Arch) -> TargetOptions {
     TargetOptions {
         abi: target_abi(arch),
         cpu: target_cpu(arch),
-        dynamic_linking: false,
+        dynamic_linking: true,
         executables: true,
         link_env_remove: link_env_remove(arch),
         has_elf_tls: false,


wesleywiser:
Based on [this comment](https://github.com/rust-lang/rust/pull/77716#issuecomment-709325990), it seems like there are still several unresolved issues and merging this would re-introduce the issues that caused #77716 to be opened in the first place. 

Is my understanding correct or have those issues been mitigated in another way and there is no longer a concern of breaking users? 

Thanks!

JohnCSimon:
Ping from triage: 
@Torrekie Can you please address the question from the reviewer?

Torrekie:
> Based on [this comment](https://github.com/rust-lang/rust/pull/77716#issuecomment-709325990), it seems like there are still several unresolved issues and merging this would re-introduce the issues that caused #77716 to be opened in the first place.
> 
> Is my understanding correct or have those issues been mitigated in another way and there is no longer a concern of breaking users?
> 
> Thanks!

'breaking users' was not caused by 'dynamic linking' but 'crate-types' are not able to override by target specific configurations. rust-lang/cargo/#4881

Users could correctly build static or dynamic libraries for iOS by setting crate type in Cargo.toml before #77716, for iOS Apps that needs to be uploaded to App Store, they can just set crate type as static. Allowing dynamic linking is not stopping people from developing iOS Apps for production use, but loosing an option to build some dylibs to iOS for tests or other use. As [this comment](https://github.com/rust-lang/rust/pull/77716#issuecomment-739543335) said, he's workflow would not able to cross compile rust libraries after compiler blocked iOS dylibs (and his distro still having large ammount users in iOS jailbreak community). Aren't users who were attempting to build iOS dynamic libraries 'breaking users'?

sorry for my terrible English tho

Torrekie:
Currently I have a custom compiler with some modifications on my device that could let me create dylibs for Darwin targets. But bootstrapping full rust compiler to jailbroken iOS environment needs to compile rust for twice, which was spending much time: once for enabling rustc to build dylibs when iOS targets has been set, and use the compiler from first result to cross compile rust and rust toolchains to iOS (rustlibs cannot be built to iOS if rustc doesn't allow it).

But I still wonder if there is any future plans that regarding to those issues mentioned above. Allowing separate crate types for different targets would be pretty.

bjorn3:
You can specify a custom json target spec with dynanic linking enabled when compiling rustc. Just pass the path to the json file instead of the target name. This saves the first bootstrap.

JohnCSimon:
Ping from triage:
@Torrekie what is the status of this PR? Is it ready for review?

JohnCSimon:
Ping from triage: 
@Torrekie This PR hasn't seen any movement in the last two Months - I'm closing this as inactive. Please feel free to reopen when you're ready to continue. Thanks. 

@rustbot label: +S-inactive

dvc94ch:
@francesca64 `cargo rustc` now supports the `--crate-type` flag. Does that fix your build so dynamic library building can be reenabled for ios?

francesca64:
@dvc94ch I've changed jobs, so I'm no longer involved; @ArthurKValladares is the new point of contact.

ArthurKValladares:
I would hate to see improvements to the language lever landing because of work that has fallen on my shoulders. I'll dedicate some time in the next few days to bringing myself up to date on this issue and figuring if the `--crate-type` fix solves the problem we were running into.

ArthurKValladares:
I've specified `--crate-type=staticlib` in our iOS build step,  and I'm getting errors related to our deps not being available as `rlib`s:
```
error: crate `itertools` required to be available in rlib format, but was not found in this form

error: crate `rand` required to be available in rlib format, but was not found in this form

error: crate `serde` required to be available in rlib format, but was not found in this form

error: crate `puzzle_generator` required to be available in rlib format, but was not found in this form

error: crate `strum` required to be available in rlib format, but was not found in this form

error: crate `brainstorm` required to be available in rlib format, but was not found in this form

error: crate `runtime` required to be available in rlib format, but was not found in this form

error: crate `action` required to be available in rlib format, but was not found in this form

error: crate `num_format` required to be available in rlib format, but was not found in this form

error: crate `bincode` required to be available in rlib format, but was not found in this form

error: crate `reqwest` required to be available in rlib format, but was not found in this form
```

Manually setting:
```
[lib]
crate-type = ["staticlib"]
```

in our `Cargo.toml` works as expected.

The full `rustc` command we use to try to build on iOS is:
`cargo rustc --package jumbline --manifest-path /Users/arthurvalladares/Documents/Brainium/jumbline-reboot/Cargo.toml --target aarch64-apple-ios --no-default-features --features "firebase helpshift metal" --color auto --crate-type=staticlib --lib -Z unstable-options `

dvc94ch:
@ArthurKValladares I believe you're running into https://github.com/rust-lang/cargo/issues/10356 can you try on nightly?

ArthurKValladares:
Yeah, this is the problem. I was under the impression that I was on nightly, but I was mistaken and everything worked after a `rustup update`. Sorry for deleting the old comment, I realized it was incorrect and was about to make a new comment with the updated information, but you beat me to the punch.

dvc94ch:
@ArthurKValladares thanks for looking into this

@wesleywiser is that enough to reopen and merge this PR?

dvc94ch:
Given the new evidence, I eopened the PR in #95847

Issue #55124, opened 2018-10-16
Concrete type only used in `impl Trait` is unused


cramertj:
The following code gives "warning: enum is never used: `Empty`":

```rust
enum Empty { }
trait Bar<T> {}
impl Bar<Empty> for () {}

fn boo() -> impl Bar<Empty> {}

fn main() {
    boo();
}
```

This regressed pretty recently. cc @oli-obk

jonas-schievink:
1.27.0 does not warn, 1.28.0 does.

pnkfelix:
triage: P-medium; This seems like a papercut that should be easy to work-around (at least if one isn't using `forbid`...).

pnkfelix:
switched to P-high after prompting from @nikomatsakis

pnkfelix:
whoops forgot to tag with T-compiler in time for the meeting...

pnkfelix:
assigning to @nikomatsakis

oli-obk:
I'm grabbing this, I think I broke it and it should be easy to fix

PR #88136, opened 2021-08-18
Move private_unused.rs test to impl-trait

spastorino:
This test was added to fix this issue #55124 which is about impl traits but not related with type alias impl traits.

r? @oli-obk 

@bors rollup=always


spastorino:
I've just realized that `@bors rollup=always` doesn't work in the description ...

@bors rollup=always

PR #88137, opened 2021-08-18
On macOS, make strip="symbols" not pass any options to strip

joshtriplett:
This makes the output with `strip="symbols"` match the result of just
calling `strip` on the output binary, minimizing the size of the binary.

### DIFF
--- a/compiler/rustc_codegen_ssa/src/back/link.rs
+++ b/compiler/rustc_codegen_ssa/src/back/link.rs
@@ -977,14 +977,20 @@ fn link_natively<'a, B: ArchiveBuilder<'a>>(
     }
 
     if sess.target.is_like_osx {
-        if let Some(option) = osx_strip_opt(sess.opts.debugging_opts.strip) {
-            strip_symbols_in_osx(sess, &out_filename, option);
+        match sess.opts.debugging_opts.strip {
+            Strip::Debuginfo => strip_symbols_in_osx(sess, &out_filename, Some("-S")),
+            Strip::Symbols => strip_symbols_in_osx(sess, &out_filename, None),
+            Strip::None => {}
         }
     }
 }
 
-fn strip_symbols_in_osx<'a>(sess: &'a Session, out_filename: &Path, option: &str) {
-    let prog = Command::new("strip").arg(option).arg(out_filename).output();
+fn strip_symbols_in_osx<'a>(sess: &'a Session, out_filename: &Path, option: Option<&str>) {
+    let mut cmd = Command::new("strip");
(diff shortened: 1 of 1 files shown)


compiler/rustc_codegen_ssa/src/back/link.rs
```diff
@@ -1002,14 +1008,6 @@ fn strip_symbols_in_osx<'a>(sess: &'a Session, out_filename: &Path, option: &str
     }
 }
 
-fn osx_strip_opt<'a>(strip: Strip) -> Option<&'a str> {
-    match strip {
-        Strip::Debuginfo => Some("-S"),
-        Strip::Symbols => Some("-x"),
```
bjorn3, line 1008: `-x` strips local symbols, but keeps global symbols. I think global symbols are necessary for linking to dylibs and thus must not be stripped.

hkratz: According to the manpage strip without any options is smart. If the executable uses dylibs strip without any options and the executable does use dylibs it is equivalent to `strip -u -r` _"which saves all undefined symbols (usually defined in the dynamic shared libraries) and all global  symbols  defined  in  the  executable  referenced  by  the  dynamic libraries  (as  marked  by  the  static link editor when the executable was built). This is the maximum level of striping for an executable that will still  allow  the program to run correctly with its libraries."_

estebank (approved):
I'm not an expert on macOS, but the changes look reasonable. CC @rust-lang/compiler 
Marking as needing attention of the team next thursday.

joshtriplett:
Anyone available to review and r+ this? I'd love to get this merged, to unblock support for strip in cargo.

michaelwoerister:
During the compiler team triage meeting it was [mentioned](https://rust-lang.zulipchat.com/#narrow/stream/238009-t-compiler.2Fmeetings/topic/.5Bweekly.5D.202021-09-30.20.2354818) that we'd like @alexcrichton to take a look (because Alex knows about this kind of stuff on OSX). If there is no progress in a couple of days, I'll take a look.

michaelwoerister:
Thanks, @alexcrichton! I'll take a look then.

@joshtriplett, what is the motivation behind this change? Why exactly is it needed to unblock strip in cargo?

joshtriplett:
Stable support for strip in cargo needs stable support for strip in rustc, as well as reasonable support for common targets. Once we stabilize strip in rustc, it'll be much harder to change its behavior without risking compatibility issues. So I'd like to get this change in, then stabilize strip in rustc, then stabilize it in cargo.

PR #88144, opened 2021-08-18
codegen: accept new noundef annotation from LLVM

durin42:
This allows the array-equality.rs test case to pass on both LLVM 12 and
LLVM 14. This is similar to work done in #87570, but is a technique
recommended to me by one of my LLVM gurus.


durin42:
FYI @nikic in case this is of interest to you.

nikic:
The LLVM upgrade ended up landing first, so I'm going to close this ... we don't really care about attributes LLVM infers here.

PR #88145, opened 2021-08-18
codegen: disable this SIMD test on LLVMs 13 and 14

durin42:
Per the upgrade to LLVM 13 PR (#87570) it looks like we'll have to eat
this regression for a while. Rather than drop the test entirely, disable
this on LLVMs 13 and 14 so that we at least remember to look at the
failure again in LLVM 15's development cycle.


durin42:
FYI @nikic in case this is of interest to you.

PR #88147, opened 2021-08-18
Fix non-capturing closure return type coercion

### LINKED ISSUE #88097: "Expected type `[closure@...]`, found fn pointer" when unnecessarily `return`ing a closure
Given [the following code](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=f194bbd3fb74c407becacd9e6927e0df):

```rust
fn peculiar() -> impl Fn(u8) -> u8 {
    return |x| x + 1
}
```

The current output is:

```
 --> src/lib.rs:1:36
  |
1 |   fn peculiar() -> impl Fn(u8) -> u8 {
  |  ____________________________________^
2 | |     return |x| x + 1
3 | | }
  | |_^ expected closure, found fn pointer
  |
  = note:    expected type `[closure@src/lib.rs:2:12: 2:21]`
          found fn pointer `fn(u8) -> u8`
```

Ideally the code would compile successfully, but if that is infeasible, it would be nice if the diagnostic contained a hint to remove the `return`. (It's rather head-scratching that the closure is being converted to a fn pointer _even though the return type has also been resolved to be the closure type._)

The `return` is entirely unnecessary here, but I imagine a beginner might include it unthinkingly (as I did) and not be able to even guess that the `return` has anything to do with the type error.

----

The results are the same on stable 1.54.0, beta, and nightly.

If I modify the closure to not be coercible to a function pointer by capturing a variable, then the code compiles without error or relevant warning:

```rust
fn peculiar(y: u8) -> impl Fn(u8) -> u8 {
    return move |x| x + y
}
```

FabianWolff:
Fixes #88097. For the example given there:
```rust
fn peculiar() -> impl Fn(u8) -> u8 {
    return |x| x + 1
}
```
which incorrectly reports an error, I noticed something weird in the debug log:
```
DEBUG rustc_typeck::check::coercion coercion::try_find_coercion_lub([closure@test.rs:2:12: 2:21], [closure@test.rs:2:12: 2:21], exprs=1 exprs)
```
Apparently, `try_find_coercion_lub()` thinks that the LUB for two closure types always has to be a function pointer (which explains the `expected closure, found fn pointer` error in #88097). There is one corner case where that isn't true, though — namely, when the two closure types are equal, in which case the trivial LUB is the type itself. This PR fixes this by inserting an explicit check for type equality in `try_find_coercion_lub()`.

### DIFF
--- a/compiler/rustc_typeck/src/check/coercion.rs
+++ b/compiler/rustc_typeck/src/check/coercion.rs
@@ -941,6 +941,8 @@ impl<'a, 'tcx> FnCtxt<'a, 'tcx> {
             exprs.len()
         );
 
+        // The following check fixes #88097, where the compiler erroneously
+        // attempted to coerce a closure type to itself via a function pointer.
         if prev_ty == new_ty {
             return Ok(prev_ty);
         }
--- a/compiler/rustc_typeck/src/check/coercion.rs
+++ b/compiler/rustc_typeck/src/check/coercion.rs
@@ -941,6 +941,10 @@ impl<'a, 'tcx> FnCtxt<'a, 'tcx> {
             exprs.len()
         );
 
+        if prev_ty == new_ty {
+            return Ok(prev_ty);
+        }
+
         // Special-case that coercion alone cannot handle:
         // Function items or non-capturing closures of differing IDs or InternalSubsts.
         let (a_sig, b_sig) = {
--- a/src/test/ui/coercion/issue-88097.rs
+++ b/src/test/ui/coercion/issue-88097.rs
@@ -1,0 +1,31 @@
+// In #88097, the compiler attempted to coerce a closure type to itself via
+// a function pointer, which caused an unnecessary error. Check that this
+// behavior has been fixed.
+
+// check-pass
+
+fn peculiar() -> impl Fn(u8) -> u8 {
+    return |x| x + 1
+}
+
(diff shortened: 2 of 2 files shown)


jackh726:
So, this is certainly a fix. But I have to think a little harder on the subtlety of this error. Specifically, I would expect that the check if the two types are equal to be an optimization and this could fail in other cases still.

Regardless, in the meantime, can you add a comment?

FabianWolff:
> Regardless, in the meantime, can you add a comment?

Sure, where and about what exactly? A reference to #88097 and/or this PR?

compiler/rustc_typeck/src/check/coercion.rs
```diff
@@ -941,6 +941,10 @@ impl<'a, 'tcx> FnCtxt<'a, 'tcx> {
             exprs.len()
         );
 
+        if prev_ty == new_ty {
```
jackh726, line 944: Comment here referencing #88097 and coercion of a closure to itself

FabianWolff: Got it; fixed now.

jackh726:
Thanks, I'll try to get back to this soon and think more about this in depth.

jackh726:
I'm not 100% sure this is the only way to do this, but it's a very reasonable change, so I'm going to go ahead and @bors r+ rollup

PR #88151, opened 2021-08-19
Update the backtrace crate in libstd

### LINKED ISSUE #432: libbacktrace cannot be updated in Rust's libstd because of memchr dependency
I suppose that this is more a tracking issue for a downstream problem, but it seems to make sense to raise it here.

In libbacktrace 0.2.60, the dependency on [object ](https://github.com/gimli-rs/object/) has been raised to 0.25.0. This release contains an [unconditional dependency](https://github.com/gimli-rs/object/commit/de7ecdd13776b687ae435f499fe500a35afc8b73#diff-2e9d962a08321605940b5a657135052fbcef87b5e360662bb527c96d9a615542) on [memchr](https://github.com/BurntSushi/memchr). This crate does not support the `rustc-std-workspace-core` workaround.

If the backtrace version shipped with `rustc` needs to be set higher than 0.3.59, this problem needs to be solved.

CC: @BurntSushi CC: @philipc @fitzgen

alexcrichton:
This commit updates the backtrace crate in libstd now that dependencies
have been updated to use `memchr` from the standard library as well.
This is mostly just making sure deps are up-to-date and have all the
latest-and-greatest fixes and such.

Closes rust-lang/backtrace-rs#432

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4,9 +4,9 @@ version = 3
 
 [[package]]
 name = "addr2line"
-version = "0.14.0"
+version = "0.16.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "7c0929d69e78dd9bf5408269919fcbcaeb2e35e5d43e5815517cdc6a8e11a423"
+checksum = "3e61f2b7f93d2c7d2b08263acaa4a363b3e276806c68af6134c44f523bf1aacd"
 dependencies = [
  "compiler_builtins",
  "gimli",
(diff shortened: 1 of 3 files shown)


hkratz:
It seems that the update makes it impossible to build std with the `backtrace` feature disabled, e.g. with `backtrace=false` in config.toml  or with `cargo +nightly build -Z build-std -Zbuild-std-features=panic-unwind,default`:
```
error[E0433]: failed to resolve: use of undeclared crate or module `object`
 --> library/std/src/../../backtrace/src/symbolize/gimli/macho.rs:4:5
  |
4 | use object::read::macho::{MachHeader, Nlist, Section, Segment as _};
  |     ^^^^^^ use of undeclared crate or module `object`
```

 I rather liked this because it shaved of a bit of executable size while keeping RAII guarantees on panics.

alexcrichton:
Oh that's an unintended consequence of a historical refactoring in the crate, https://github.com/rust-lang/backtrace-rs/pull/433 should restore support.

alexcrichton:
Ok, I've updated the submodule here to include that fix.

Mark-Simulacrum:
This was a perf [regression](https://perf.rust-lang.org/compare.html?start=2451f42c1deb9379d5e8e5fa86b0bf857ae048ec&end=7960030d6915a771f5ab72c3897a7ed50c3ed4bd&stat=instructions:u), on a number of crates, but I believe it to be entirely or largely due to more metadata decoding and related processing of std, which this PR increased the size of.

@rustbot label perf-regression perf-regression-triaged

PR #88154, opened 2021-08-19
Add `core::ops::GeneratorState::{is_yielded,is_complete}` methods

yoshuawuyts:
Similar to [`Option::is_*`](https://doc.rust-lang.org/std/option/enum.Option.html#method.is_some), [`Result::is_*`](https://doc.rust-lang.org/std/result/enum.Result.html#method.is_err) and [`ControlFlow::is_*`](https://doc.rust-lang.org/std/ops/enum.ControlFlow.html#method.is_continue) methods, this adds `is_*` methods for `GeneratorState`. Thanks!

### DIFF
--- a/library/core/src/ops/generator.rs
+++ b/library/core/src/ops/generator.rs
@@ -26,6 +26,46 @@ pub enum GeneratorState<Y, R> {
     Complete(R),
 }
 
+impl<Y, R> GeneratorState<Y, R> {
+    /// Returns `true` if the generator state is [`Yielded`].
+    ///
+    /// [`Yielded`]: GeneratorState::Yielded
+    ///
+    /// # Examples
+    ///
+    /// ```
+    /// #![feature(generator_trait)]
+    ///
+    /// use std::ops::GeneratorState;
+    ///
+    /// assert!(GeneratorState::Yielded::<i32, &str>(1).is_yielded());
+    /// assert!(!GeneratorState::Complete::<i32, &str>("foo").is_yielded());
+    /// ```
+    #[unstable(feature = "generator_trait", issue = "43122")]
+    pub fn is_yielded(&self) -> bool {
+        matches!(self, Self::Yielded(..))
+    }
+
+    /// Returns `true` if the generator state is [`Complete`].
+    ///
+    /// [`Complete`]: GeneratorState::Complete
+    ///
+    /// # Examples
+    ///
+    /// ```
+    /// #![feature(generator_trait)]
+    ///
+    /// use std::ops::GeneratorState;
+    ///
+    /// assert!(!GeneratorState::Yielded::<i32, &str>(1).is_complete());
+    /// assert!(GeneratorState::Complete::<i32, &str>("foo").is_complete());
+    /// ```
+    #[unstable(feature = "generator_trait", issue = "43122")]
+    pub fn is_complete(&self) -> bool {
+        matches!(self, Self::Complete(..))
+    }
+}
+
 /// The trait implemented by builtin generator types.
 ///
 /// Generators, also commonly referred to as coroutines, are currently an


m-ou-se:
Thanks! Can you give an example of a concrete use case?

yoshuawuyts:
> Thanks! Can you give an example of a concrete use case?

Sure thing! I only recently started looking at generators more closely, so I don't have loads to back it up in terms of experience. But one thing where these methods are particularly useful for is when validating generator state. For example, we could rewrite [the main generator example](https://doc.rust-lang.org/std/ops/trait.Generator.html) using `is_` + `assert!`:
```diff
#![feature(generators, generator_trait)]

use std::ops::{Generator, GeneratorState};
use std::pin::Pin;

fn main() {
    let mut generator = || {
        yield 1;
        return "foo"
    };

-    match Pin::new(&mut generator).resume(()) {
-        GeneratorState::Yielded(1) => {}
-        _ => panic!("unexpected return from resume"),
-    }
+    assert!(Pin::new(&mut generator).resume(()).is_yielded());
-    match Pin::new(&mut generator).resume(()) {
-        GeneratorState::Complete("foo") => {}
-        _ => panic!("unexpected return from resume"),
-    }
+    assert!(Pin::new(&mut generator).resume(()).is_complete());
}
```

I don't know how much manual generator impls, generator state introspection, and `is_` methods specifically will be used in practice. But that's part of the stabilization process to figure out. Given many other enums in the stdlib which serve similar functions (`Result`, `Option`, `Poll`, `ControlFlow`, etc) implement `is_` accessors, it seemed to me like adding these methods would be fairly uncontroversial.

yoshuawuyts:
Hmm, yeah that seems like it would be useful indeed. Thinking about this a bit: if we consider `ControlFlow` and `GeneratorState` as aliases for `Result`, so too could we consider `Poll` to be an alias for `Option`. I think it could make sense to ensure that methods implemented on the base type are also guaranteed to exist on their aliases. For example on `ControlFlow` that would mean:

- keep `map_break` and add `map_continue` 
- rename `break_value` -> `ok`
- keep `is_break` and `is_continue`
- add remainder of Result methods

This could equally be applied to `GeneratorState`. The only struct where I don't think we can do this (and in hindsight this is likely a mistake) is `Poll`. Methods such as [`map_err`](https://doc.rust-lang.org/std/task/enum.Poll.html#method.map_err) and [`map_ok`](https://doc.rust-lang.org/std/task/enum.Poll.html#method.map_ok) don't have counterparts in [`Option`](https://doc.rust-lang.org/std/option/enum.Option.html). And the way [`Try for Poll`](https://doc.rust-lang.org/std/task/enum.Poll.html#impl-Try) has been implemented has lead to the need to add [`task::ready!`](https://doc.rust-lang.org/nightly/std/task/macro.ready.html).

It definitely seems like it would've made sense to keep these types consistent from the start. But because late is better than never, we might as well do it now. If this reasoning all makes sense; what would the right way be to proceed with this? I believe libs has both RFCs and policies; but I'm not sure which one this would fall under?

SkiFire13:
> For example, we could rewrite the main generator example using is_ + assert!:

That will lose informations about the values that were yielded though.

yoshuawuyts:
@SkiFire13 I'm not proposing we change the example; keeping as much information as possible makes sense here. It merely served to illustrate that when folks author their own generators by hand, being able to assert intermediate states can sometimes be simplified by providing `is_` methods.

m-ou-se:
> It definitely seems like it would've made sense to keep these types consistent from the start. But because late is better than never, we might as well do it now. If this reasoning all makes sense; what would the right way be to proceed with this?

An RFC would work well for this kind of thing.

Dylan-DPC:
@yoshuawuyts any updates on the rfc for this change?

Dylan-DPC:
Closing this as its inactive and waiting on a rfc that will take time to merge.

PR #88156, opened 2021-08-19
Adjust / fix documentation of `Arc::make_mut`

steffahn:
Related discussion in the users forum:
[Whatʼs this alleged difference between Arc::make_mut and Rc::make_mut? – The Rust Programming Language Forum](https://users.rust-lang.org/t/what-s-this-alleged-difference-between-arc-make-mut-and-rc-make-mut/63747?u=steffahn)

Also includes a small formatting improvement in the documentation of `Rc::make_mut`.

This PR makes the two documentations in question complete analogs. The previously claimed point in which one “differs from the behavior of” the other turns out to be incorrect, AFAIK.

One remaining inaccuracy: `Weak` pointers aren’t disassociated from the allocation but only from the contained value, i.e. in case of outstanding `Weak` pointers there still is a new allocation created, just the call to `.clone()` is avoided, instead the value is moved from one allocation to the other.

@rustbot label T-libs-api, A-docs

### DIFF
--- a/library/alloc/src/sync.rs
+++ b/library/alloc/src/sync.rs
@@ -1350,8 +1350,9 @@ impl<T: Clone> Arc<T> {
     /// [`clone`] the inner value to a new allocation to ensure unique ownership.  This is also
     /// referred to as clone-on-write.
     ///
-    /// If there are no other `Arc` pointers to this allocation, then [`Weak`]
-    /// pointers to this allocation will be disassociated.
+    /// However, if there are no other `Arc` pointers to this allocation, but some [`Weak`]
+    /// pointers, then the [`Weak`] pointers will be disassociated and the inner value will not
+    /// be cloned.
     ///
     /// See also [`get_mut`], which will fail rather than cloning.
     ///
--- a/library/alloc/src/sync.rs
+++ b/library/alloc/src/sync.rs
@@ -1346,18 +1346,17 @@ impl<T: ?Sized> Receiver for Arc<T> {}
 impl<T: Clone> Arc<T> {
     /// Makes a mutable reference into the given `Arc`.
     ///
-    /// If there are other `Arc` or [`Weak`] pointers to the same allocation,
-    /// then `make_mut` will create a new allocation and invoke [`clone`][clone] on the inner value
-    /// to ensure unique ownership. This is also referred to as clone-on-write.
+    /// If there are other `Arc` pointers to the same allocation, then `make_mut` will
+    /// [`clone`] the inner value to a new allocation to ensure unique ownership.  This is also
+    /// referred to as clone-on-write.
     ///
-    /// Note that this differs from the behavior of [`Rc::make_mut`] which disassociates
-    /// any remaining `Weak` pointers.
+    /// If there are no other `Arc` pointers to this allocation, then [`Weak`]
+    /// pointers to this allocation will be disassociated.
     ///
-    /// See also [`get_mut`][get_mut], which will fail rather than cloning.
+    /// See also [`get_mut`], which will fail rather than cloning.
     ///
-    /// [clone]: Clone::clone
-    /// [get_mut]: Arc::get_mut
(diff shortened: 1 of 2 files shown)


steffahn:
Oops, I’m having the sentence about `clone-on-write` twice now… let me fit that.

library/alloc/src/sync.rs
```diff
@@ -1346,18 +1346,17 @@ impl<T: ?Sized> Receiver for Arc<T> {}
...
-    /// If there are other `Arc` or [`Weak`] pointers to the same allocation,
-    /// then `make_mut` will create a new allocation and invoke [`clone`][clone] on the inner value
-    /// to ensure unique ownership. This is also referred to as clone-on-write.
+    /// If there are other `Arc` pointers to the same allocation, then `make_mut` will
+    /// [`clone`] the inner value to a new allocation to ensure unique ownership.  This is also
+    /// referred to as clone-on-write.
     ///
-    /// Note that this differs from the behavior of [`Rc::make_mut`] which disassociates
-    /// any remaining `Weak` pointers.
+    /// If there are no other `Arc` pointers to this allocation, then [`Weak`]
```
Mark-Simulacrum, line 1353:
```suggestion
    /// However, if there are no other `Arc` pointers to this allocation, but some [`Weak`] pointers, then the [`Weak`] pointers will be disassociated and the inner value will not be cloned.
```

Could use wrapping and integration into the structure overall, but maybe something like this? I think the current format seems a little disconnected from the lack of cloning and such noted in the previous paragraph.

Looks like the Rc docs could use a similar update.

steffahn: Yeah, the main point of this PR is to make the doc of `Rc::make_mut` and `Arc::make_mut` be the same; any further change should apply to both as-well.

library/alloc/src/sync.rs
```diff
@@ -1346,18 +1346,18 @@ impl<T: ?Sized> Receiver for Arc<T> {}
...
+    /// referred to as clone-on-write.
     ///
-    /// Note that this differs from the behavior of [`Rc::make_mut`] which disassociates
-    /// any remaining `Weak` pointers.
+    /// However, if there are no other `Arc` pointers to this allocation, but some [`Weak`]
+    /// pointers, then the [`Weak`] pointers will be disassociated and the inner value will not
+    /// be cloned.
     ///
-    /// See also [`get_mut`][get_mut], which will fail rather than cloning.
+    /// See also [`get_mut`], which will fail rather than cloning.
```
Mark-Simulacrum, line 1357: Hm - I think this may now be a little misleading. It looks like get_mut will not "drop" Weaks, so it might be worth clarifying here (e.g., "which will fail with any existing references, both weak and strong")

steffahn: Maybe “which will fail rather than cloning the inner value or diassociating [`Weak`] pointers”?

steffahn: `get_mut` has a similarly inaccurate sentence
> See also [`make_mut`], which will [`clone`] the inner value when there are other pointers.

steffahn: Perhaps change that to ``See also [`make_mut`], which will [`clone`] the inner value when there are other [`Arc`] pointers.`` (or ``[`Rc`]``, respectively)

Mark-Simulacrum: Your suggested phrasing seems good.

I think maybe a good alternative is to make sure each of get_mut/make_mut is clear about the conditions, and then just say "See also $other_one" and not try to describe what it does in a sentence.

steffahn: Made a new commit.

steffahn: I don’t think it hurts necessarily to have the quick comparing explanation; especially since it’s pointing out a relevant difference between the two.

steffahn: And each one by itself should be clear about the conditions right now.

PR #88157, opened 2021-08-19
bootstrap.py: recognize riscv64 when auto-detect

Icenowy:
The architecture auto-detect table has no entry for riscv64 (which rustc
uses riscv64gc for the first part of triplet, assuming it's a generic
Linux distro).

Add it to the table to allow riscv64 systems to bootstrap Rust.

Signed-off-by: Icenowy Zheng <icenowy@aosc.io>


Mark-Simulacrum:
@bors r+ rollup

Seems pretty harmless.

Issue #86727, opened 2021-06-29
Reconcile TAIT tests


nikomatsakis:
The draft [stabilization report](https://hackmd.io/o9KO-D8aSb6bJgQ1froVTA) for "min TAIT" includes a series of test scenarios:

We need to go through the existing tests and either give examples of each scenario or add new ones. You can just edit the draft stabilization report hackmd directly.

spastorino:
I think this issue can be closed now. I've built a table here https://hackmd.io/y1HwFSBUQS2ZKrSFuQ0Dnw?view with existing tests, some desired ones and ended creating new tests for the desired ones that were missing. The table reflects that showing PRs for cases that were uncovered.

Please whoever that is reading this and have time and energy :), please check it out and let me know if there's something missing or feel free to add more tests :).

spastorino:
All the open PRs about this are now closed.

PR #88159, opened 2021-08-19
Use a trait instead of the now disallowed missing trait there

spastorino:
To properly test the unused generic parameter and just that we need to use a trait in the tait. Missing the trait there is already tested and is not what we meant to test here.

r? @oli-obk


spastorino:
Related to #86727

PR #88161, opened 2021-08-19
Fix handling of +whole-archive native link modifier.

### LINKED ISSUE #88085: Whole archive modifier for native libraries is not propagated to final output.
With the additional of RFC 2951 (tracking: #81490, implementation: #83507), linking modifiers such as `+whole-archive` are now possible to use for native libraries. However, `+whole-archive` is not propagated to the final output of the build. I've minimally reproduced this issue as best I could [here](https://gitlab.com/gcoakes/minimally-reproduce). It can probably be minimized more with raw calls to rustc, but I'm not familiar enough to do so right now.

Essentially, I have a static library which provides constructor and destructor functions which must be included in the final output.

I tried this code:

foo.c
```rust
#include <stdio.h>
void __attribute__((constructor)) construct();
void __attribute__((destructor)) destruct();
void construct()
{
    printf("Constructor.\n");
}
void destruct()
{
    printf("Destructor.\n");
}
```

build.rs
```rust
fn main() {
    cc::Build::new()
        .file("src/foo.c")
        .cargo_metadata(false)
        .static_flag(true)
        .compile("foo");
    println!(
        "cargo:rustc-link-search=native={}",
        std::env::var("OUT_DIR").unwrap()
    );
    println!("cargo:rustc-link-lib=static:+whole-archive=foo");
}
```

lib.rs
```rust
pub fn it_linked() {}
```

main.rs
```rust
fn main() {
    indirect::it_linked();
    println!("Hello, indirect!");
}
```

I expected to see this happen:
```
$ RUSTFLAGS="-Z unstable-options" cargo run --bin indirect
   Compiling cc v1.0.69
   Compiling direct v0.1.0 (/home/gcoakes/Documents/Source/minimal-whole-archive/indirect)
    Finished dev [unoptimized + debuginfo] target(s) in 1.73s
     Running `target/debug/indirect`
Constructor.
Hello, indirect!
Destructor.
$ nm target/debug/indirect | grep 'construct\|destruct'
0000000000007ab5 T construct
0000000000007acc T destruct
```

Instead, this happened:
```
$ RUSTFLAGS="-Z unstable-options" cargo run --bin indirect
   Compiling indirect v0.1.0 (/home/gcoakes/Documents/Source/minimal-whole-archive/indirect)
    Finished dev [unoptimized + debuginfo] target(s) in 0.68s
     Running `target/debug/indirect`
Hello, indirect!
$ nm target/debug/indirect | grep 'construct\|destruct'
```

### Resolution

I've done a bit of code inspection to determine where the linking issue actually is. I think it comes down to the [`link_upstream`](https://github.com/rust-lang/rust/blob/73d96b090bb68065cd3a469b27cbd568e39bf0e7/compiler/rustc_codegen_ssa/src/back/link.rs#L2193) closure in `rustc_codegen_ssa::src::back::link::add_upstream_rust_crates::add_static_crate`. It should be adjusted to inspect the crates native libraries and determine if any of them have `+whole-archive`. If one or more do, then it should run `link_whole_rlib` on the crate's rlib instead of `link_rlib`.

### Meta

`rustc --version --verbose`:
```
rustc 1.56.0-nightly (8007b506a 2021-08-14)
binary: rustc
commit-hash: 8007b506ac5da629f223b755f5a5391edd5f6d01
commit-date: 2021-08-14
release: 1.56.0-nightly
LLVM version: 12.0.1
```

michaelwoerister:
This PR fixes a bug in `add_upstream_native_libraries` that led to the `+whole-archive` modifier being ignored when linking in native libs.

~~Note that the PR does not address the situation when `+whole-archive` is combined with `+bundle`.~~
@wesleywiser's commit adds validation code that turns combining `+whole-archive` with `+bundle` into an error.

Fixes https://github.com/rust-lang/rust/issues/88085.

r? @petrochenkov
cc @wesleywiser @gcoakes

### DIFF
--- a/compiler/rustc_codegen_ssa/src/back/link.rs
+++ b/compiler/rustc_codegen_ssa/src/back/link.rs
@@ -255,6 +255,19 @@ fn link_rlib<'a, B: ArchiveBuilder<'a>>(
     // metadata of the rlib we're generating somehow.
     for lib in codegen_results.crate_info.used_libraries.iter() {
         match lib.kind {
+            NativeLibKind::Static { bundle: None | Some(true), whole_archive: Some(true) }
+                if flavor == RlibFlavor::Normal =>
+            {
+                // Don't allow mixing +bundle with +whole_archive since an rlib may contain
+                // multiple native libs, some of which are +whole-archive and some of which are
+                // -whole-archive and it isn't clear how we can currently handle such a
+                // situation correctly.
+                // See https://github.com/rust-lang/rust/issues/88085#issuecomment-901050897
+                sess.err(
+                    "the linking modifiers `+bundle` and `+whole-archive` are not compatible \
+                        with each other when generating rlibs",
+                );
+            }
             NativeLibKind::Static { bundle: None | Some(true), .. } => {}
             NativeLibKind::Static { bundle: Some(false), .. }
             | NativeLibKind::Dylib { .. }
@@ -1223,6 +1236,7 @@ pub fn archive_search_paths(sess: &Session) -> Vec<PathBuf> {
     sess.target_filesearch(PathKind::Native).search_path_dirs()
 }
 
+#[derive(PartialEq)]
 enum RlibFlavor {
     Normal,
     StaticlibBase,
--- a/compiler/rustc_codegen_ssa/src/back/link.rs
+++ b/compiler/rustc_codegen_ssa/src/back/link.rs
@@ -36,6 +36,7 @@ use regex::Regex;
 use tempfile::Builder as TempFileBuilder;
 
 use std::ffi::OsString;
+use std::lazy::OnceCell;
 use std::path::{Path, PathBuf};
 use std::process::{ExitStatus, Output, Stdio};
 use std::{ascii, char, env, fmt, fs, io, mem, str};
@@ -2001,7 +2002,7 @@ fn add_local_native_libraries(
     let relevant_libs =
         codegen_results.crate_info.used_libraries.iter().filter(|l| relevant_lib(sess, l));
 
-    let search_path = archive_search_paths(sess);
+    let search_path = OnceCell::new();
     let mut last = (NativeLibKind::Unspecified, None);
     for lib in relevant_libs {
         let name = match lib.name {
@@ -2023,7 +2024,11 @@ fn add_local_native_libraries(
             }
             NativeLibKind::Static { bundle: None | Some(true), .. }
             | NativeLibKind::Static { whole_archive: Some(true), .. } => {
-                cmd.link_whole_staticlib(name, verbatim, &search_path);
+                cmd.link_whole_staticlib(
+                    name,
+                    verbatim,
+                    &search_path.get_or_init(|| archive_search_paths(sess)),
+                );
             }
             NativeLibKind::Static { .. } => cmd.link_staticlib(name, verbatim),
             NativeLibKind::RawDylib => {
@@ -2116,6 +2121,7 @@ fn add_upstream_rust_crates<'a, B: ArchiveBuilder<'a>>(
     }
 
     let mut compiler_builtins = None;
+    let search_path = OnceCell::new();
 
     for &cnum in deps.iter() {
         if group_start == Some(cnum) {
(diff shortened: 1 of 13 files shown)


src/test/run-make/native-link-modifier-whole-archive/c_static_lib_with_constructor.c
```diff
@@ -0,0 +1,13 @@
+#include <stdio.h>
+void __attribute__((constructor)) construct();
+void __attribute__((destructor)) destruct();
```
wesleywiser, line 3: `__attribute__((constructor))` and `__attribute__((destructor))` aren't supported by VC++. I'm not sure which compiler `cc` will use on the `-pc-windows-msvc` targets but we might need to skip this test if those targets use VC++ instead of clang.

michaelwoerister: Good catch! I switched to a C++-based implementation that should have the same effect.

compiler/rustc_codegen_ssa/src/back/link.rs
```diff
@@ -2328,6 +2328,7 @@ fn add_upstream_native_libraries(
         .find(|(ty, _)| *ty == crate_type)
         .expect("failed to find crate type in dependency format list");
 
+    let search_path = archive_search_paths(sess);
```
petrochenkov, line 2331: Looks like a good candidate for `std::lazy::Lazy`, it should be evaluated at most once in the loop below, but it would be better to not evaluate it if it's not actually used.

src/test/run-make/native-link-modifier-whole-archive/Makefile
```diff
@@ -0,0 +1,37 @@
...
+# Native lib linked into rlib with via commandline
+$(TMPDIR)/librlib_with_cmdline_native_lib.rlib: $(call NATIVE_STATICLIB,c_static_lib_with_constructor)
+	$(RUSTC) rlib_with_cmdline_native_lib.rs -Z unstable-options --crate-type=rlib -l static:-bundle,+whole-archive=c_static_lib_with_constructor
+
+# Native lib linked into rlib via `#[link()]` attribute on extern block.
+$(TMPDIR)/libnative_lib_in_src.rlib: $(call NATIVE_STATICLIB,c_static_lib_with_constructor)
+	$(RUSTC) native_lib_in_src.rs --crate-type=rlib
+
+$(TMPDIR)/libc_static_lib_with_constructor.o: c_static_lib_with_constructor.cpp
+	$(call COMPILE_OBJ_CXX,$@,$<)
```
petrochenkov, line 37: Hmm, I thought missing trailing newlines were supposed to trigger tidy.

compiler/rustc_codegen_ssa/src/back/link.rs
```diff
@@ -2352,19 +2353,30 @@ fn add_upstream_native_libraries(
...
                 }
                 // ignore statically included native libraries here as we've
                 // already included them when we included the rust library
                 // previously
-                NativeLibKind::Static { bundle: None | Some(true), .. } => {}
+                NativeLibKind::Static { bundle: None | Some(true), whole_archive } => {
+                    if whole_archive == Some(true) {
+                        bug!(
+                            "Combining `+bundle` with `+whole-archive` is not allowed. \
+                              This should have been caught by an earlier validation step already."
```
petrochenkov, line 2376: I don't think there is such a validation step right now.
Is there a test for this case?

michaelwoerister: I think @wesleywiser will be looking into this separately. Right, @wesleywiser?

wesleywiser: Yes, I'm working on this.

compiler/rustc_codegen_ssa/src/back/link.rs
michaelwoerister: Good idea!

src/test/run-make/native-link-modifier-whole-archive/Makefile
michaelwoerister: Makefiles are probably special. I added the newline.

compiler/rustc_codegen_ssa/src/back/link.rs
michaelwoerister: @petrochenkov, does that resolve the issue for you or should we block this PR on the validation code landing?

petrochenkov: It seems questionable to knowingly introduce an ICE.

michaelwoerister: OK, let's wait for @wesleywiser's changes to add the validation.

wesleywiser: I took the liberty of pushing into your PR @michaelwoerister.

compiler/rustc_session/src/config.rs
```diff
@@ -1734,6 +1734,17 @@ fn parse_native_lib_modifiers(
...
+    // Don't allow mixing +bundle with +whole_archive since an rlib may contain multiple
+    // native libs, some of which are +whole_archive and some of which are -whole-archive
+    // and it isn't clear how we can currently handle such a situation correctly.
+    // See https://github.com/rust-lang/rust/issues/88085#issuecomment-901050897
+    if let NativeLibKind::Static { bundle: Some(true), whole_archive: Some(true) } = kind {
+        early_error(
+            error_format,
+            "the linking modifiers +bundle and +whole-archive are not compatible with each other",
+        );
+    }
```
petrochenkov, line 1746: - This only covers command line flags, but not attributes.
- `+bundle` is currently the default, so `bundle: None, whole_archive: Some(true)` is also a potential error.
- `+bundle,+whole-archive` is only an issue if we are building a rlib, otherwise it's entirely ok, the `directly_linked.rs` test case uses this configuration, for example.

wesleywiser: Ah, yeah you're right. All three points should be covered by the latest commit I pushed.

michaelwoerister:
I think this is ready for review now.

michaelwoerister:
@petrochenkov, the merge conflict is resolved now.

GuillaumeGomez:
It failed CI in #88570.

@bors: r-

michaelwoerister:
I added `ignore-cross-compile` to the test case. Let's see if that is enough to disable the platforms that have trouble compiling C++ code.

PR #88169, opened 2021-08-19
Exempt `#[naked]` fns from `named_asm_labels` lint

asquared31415:
`#[naked]` fns are guaranteed to never be inlined or duplicated so the `named_asm_labels` lint was over-eagerly and incorrectly warning about named labels that would never be a problem within them.

Fixes an issue introduced with the lint PR #87324 

r? @Amanieu 
cc @bjorn3 @npmccallum

### DIFF
--- a/compiler/rustc_ast/src/ast.rs
+++ b/compiler/rustc_ast/src/ast.rs
@@ -2028,6 +2028,7 @@ pub enum InlineAsmOperand {
 #[derive(Clone, Encodable, Decodable, Debug)]
 pub struct InlineAsm {
     pub template: Vec<InlineAsmTemplatePiece>,
+    pub template_strs: Box<[(Symbol, Option<Symbol>, Span)]>,
     pub operands: Vec<(InlineAsmOperand, Span)>,
     pub clobber_abi: Option<(Symbol, Span)>,
     pub options: InlineAsmOptions,
--- a/compiler/rustc_ast/src/visit.rs
+++ b/compiler/rustc_ast/src/visit.rs
@@ -19,20 +19,20 @@ use crate::token;
 use rustc_span::symbol::{Ident, Symbol};
 use rustc_span::Span;
 
-#[derive(Copy, Clone, PartialEq)]
+#[derive(Copy, Clone, Debug, PartialEq)]
 pub enum AssocCtxt {
     Trait,
     Impl,
 }
 
-#[derive(Copy, Clone, PartialEq)]
+#[derive(Copy, Clone, Debug, PartialEq)]
 pub enum FnCtxt {
     Free,
     Foreign,
     Assoc(AssocCtxt),
 }
 
-#[derive(Copy, Clone)]
+#[derive(Copy, Clone, Debug)]
 pub enum FnKind<'a> {
     /// E.g., `fn foo()`, `fn foo(&self)`, or `extern "Abi" fn foo()`.
     Fn(FnCtxt, Ident, &'a FnSig, &'a Visibility, Option<&'a Block>),
--- a/compiler/rustc_builtin_macros/src/asm.rs
+++ b/compiler/rustc_builtin_macros/src/asm.rs
@@ -7,10 +7,10 @@ use rustc_errors::{Applicability, DiagnosticBuilder};
 use rustc_expand::base::{self, *};
 use rustc_parse::parser::Parser;
 use rustc_parse_format as parse;
-use rustc_session::lint::{self, BuiltinLintDiagnostics};
+use rustc_session::lint;
 use rustc_span::symbol::Ident;
 use rustc_span::symbol::{kw, sym, Symbol};
-use rustc_span::{InnerSpan, MultiSpan, Span};
+use rustc_span::{InnerSpan, Span};
 use rustc_target::asm::InlineAsmArch;
 use smallvec::smallvec;
 
@@ -484,11 +484,7 @@ fn parse_reg<'a>(
     Ok(result)
 }
 
-fn expand_preparsed_asm(
-    ecx: &mut ExtCtxt<'_>,
-    args: AsmArgs,
-    is_local_asm: bool,
-) -> Option<ast::InlineAsm> {
+fn expand_preparsed_asm(ecx: &mut ExtCtxt<'_>, args: AsmArgs) -> Option<ast::InlineAsm> {
     let mut template = vec![];
     // Register operands are implicitly used since they are not allowed to be
     // referenced in the template string.
@@ -500,6 +496,8 @@ fn expand_preparsed_asm(
         args.named_args.iter().map(|(&sym, &idx)| (idx, sym)).collect();
     let mut line_spans = Vec::with_capacity(args.templates.len());
     let mut curarg = 0;
+
+    let mut template_strs = Vec::with_capacity(args.templates.len());
 
     for template_expr in args.templates.into_iter() {
         if !template.is_empty() {
@@ -524,8 +522,13 @@ fn expand_preparsed_asm(
             ast::StrStyle::Raw(raw) => Some(raw as usize),
         };
 
-        let template_str = &template_str.as_str();
         let template_snippet = ecx.source_map().span_to_snippet(template_sp).ok();
+        template_strs.push((
+            template_str,
+            template_snippet.as_ref().map(|s| Symbol::intern(s)),
+            template_sp,
+        ));
+        let template_str = &template_str.as_str();
 
         if let Some(InlineAsmArch::X86 | InlineAsmArch::X86_64) = ecx.sess.asm_arch {
             let find_span = |needle: &str| -> Span {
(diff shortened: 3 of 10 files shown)


npmccallum:
@asquared31415 Thank you! This is helpful to me. (FYI @haraldh)

nbdd0121 (commented):
Did you account for this case?

```rust
#![feature(asm)]
#![feature(naked_functions)]

#[naked]
extern "C" fn naked() {
    fn non_naked() {
        
    }
    unsafe { asm!("/* named labels? */", options(noreturn)) };
}
```

compiler/rustc_builtin_macros/src/asm.rs
```diff
@@ -561,70 +564,6 @@ fn expand_preparsed_asm(
         }
 
         // Lint against the use of named labels in inline `asm!` but not `global_asm!`
```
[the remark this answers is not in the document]
asquared31415: Yep

asquared31415:
No I forgot to account for nested fns, despite even thinking about it and almost adding a test, then forgetting to for some reason.  Is there a better way to fix this than a stack of flags that is pushed to in `check_item` for fns and popped from in `check_item_post` for fns?

compiler/rustc_ast/src/ast.rs
```diff
@@ -2028,10 +2028,12 @@ pub enum InlineAsmOperand {
 #[derive(Clone, Encodable, Decodable, Debug)]
 pub struct InlineAsm {
     pub template: Vec<InlineAsmTemplatePiece>,
+    pub template_strs: Box<[(Symbol, Option<String>, Span)]>,
     pub operands: Vec<(InlineAsmOperand, Span)>,
     pub clobber_abi: Option<(Symbol, Span)>,
     pub options: InlineAsmOptions,
     pub line_spans: Vec<Span>,
+    pub is_local: bool,
```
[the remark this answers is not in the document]
asquared31415: Good catch, thanks!

nbdd0121:
> No I forgot to account for nested fns, despite even thinking about it and almost adding a test, then forgetting to for some reason. Is there a better way to fix this than a stack of flags that is pushed to in `check_item` for fns and popped from in `check_item_post` for fns?

I am unsure about what's the best way to do it in early lint pass. If this is done in a late lint pass, then you should be able to use `tcx.get_attrs`.

compiler/rustc_lint/src/builtin.rs
```diff
@@ -3140,3 +3140,148 @@ impl<'tcx> LateLintPass<'tcx> for DerefNullPtr {
...
+pub struct NamedAsmLabels {
+    is_naked_fn: bool,
+}
+
+impl_lint_pass!(NamedAsmLabels => [NAMED_ASM_LABELS]);
+
+impl EarlyLintPass for NamedAsmLabels {
+    fn check_expr(&mut self, cx: &EarlyContext<'_>, expr: &ast::Expr) {
+        // Skip checks for naked fns
+        if !self.is_naked_fn {
```
Amanieu, line 3186: Save a level of indentation by just doing an early return here.

compiler/rustc_lint/src/builtin.rs
```diff
@@ -3140,3 +3140,148 @@ impl<'tcx> LateLintPass<'tcx> for DerefNullPtr {
...
+            if let ExprKind::InlineAsm(asm) = &expr.kind {
+                //debug!("NamedAsmLabels::check_expr(): asm: {:#?}", asm);
```
Amanieu, line 3188: Leftover debug code?

compiler/rustc_lint/src/builtin.rs
```diff
@@ -3140,3 +3140,148 @@ impl<'tcx> LateLintPass<'tcx> for DerefNullPtr {
...
+                for (template_sym, template_snippet, template_span) in asm.template_strs.iter() {
+                    let template_str = &template_sym.as_str();
+                    debug!(
+                        "sym: {:?}, span: {:?}, snippet: {:?}",
+                        template_sym, template_span, template_snippet
+                    );
```
Amanieu, line 3194: Same here, you should remove these before merging.

compiler/rustc_ast/src/ast.rs
```diff
@@ -2028,10 +2028,12 @@ pub enum InlineAsmOperand {
 #[derive(Clone, Encodable, Decodable, Debug)]
 pub struct InlineAsm {
     pub template: Vec<InlineAsmTemplatePiece>,
+    pub template_strs: Box<[(Symbol, Option<String>, Span)]>,
```
Amanieu, line 2031: This can just be a `Vec`, it's not like any of the other `Vec`s in here are ever modified either.

compiler/rustc_ast/src/visit.rs
```diff
@@ -19,20 +19,20 @@ use crate::token;
 use rustc_span::symbol::{Ident, Symbol};
 use rustc_span::Span;
 
-#[derive(Copy, Clone, PartialEq)]
```
Amanieu, line 22: Leftover debug code?

compiler/rustc_builtin_macros/src/asm.rs
```diff
@@ -7,10 +7,10 @@ use rustc_errors::{Applicability, DiagnosticBuilder};
 use rustc_expand::base::{self, *};
 use rustc_parse::parser::Parser;
 use rustc_parse_format as parse;
-use rustc_session::lint::{self, BuiltinLintDiagnostics};
+use rustc_session::lint::{self};
```
Amanieu, line 10: I'm surprised rustfmt isn't complaining about this.
```suggestion
use rustc_session::lint;
```

compiler/rustc_ast/src/ast.rs
asquared31415: With the above `is_local` bool, a `Vec` was increasing the total size of the `ItemKind` enum and causing an assert to fail.  While I could have bumped the size of that assert, it seemed like a bad idea to increase the size of that enum for a very infrequent use case for performance concerns.  The `Box<[]>` was meant to reduce that size since I didn't need to be able to mutate it once it was created.

However with the removal of `is_local`, I can probably make that a `Vec` if you think that's clearer/better.

compiler/rustc_builtin_macros/src/asm.rs
asquared31415: That's quite odd.  I definitely have the stage0 rustfmt and the tidy commit hook, and would have expected CI to yell as well.  Probably a rustfmt bug/oversight?

asquared31415:
Moved to a late lint pass over the HIR.  It simplifies the logic for handling whether the containing fn is `#[naked]` greatly from the stack-based solution that I had.

compiler/rustc_builtin_macros/src/asm.rs
asquared31415: Update: rust-analyzer has a hint and automatic fix for this, but it's one of the small and hard to notice suggestions, and not applied by rustfmt

compiler/rustc_lint/src/builtin.rs
```diff
@@ -3140,3 +3140,137 @@ impl<'tcx> LateLintPass<'tcx> for DerefNullPtr {
...
+        } = expr
+        {
+            let hir = cx.tcx.hir();
+            if let Some(enclosing_body) = cx.enclosing_body {
+                let fn_id = hir.get_parent_item(enclosing_body.hir_id);
+                let attrs = hir.attrs(fn_id);
+                // Process lint only if the containing fn does not have a `#[naked]` attribute
+                if attrs
+                    .iter()
+                    .flat_map(Attribute::meta) // remove cases where meta doesn't exist
```
nbdd0121, line 3192: Use `Attribute::has_name`?

nbdd0121:
It seems that there are quite a few lints in `builtin.rs`. Does anyone know what's the rationale for that?

compiler/rustc_lint/src/builtin.rs
asquared31415: Oh I didn't see that method before, that makes this a lot simpler

tmiasko:
In the current implementation, local symbol names are likely to remain problematic even for naked functions. For example in LTO builds ensuring that names are unique within each crate is not sufficient to avoid duplicates.

npmccallum:
@tmiasko That appears to be true for both gcc and clang who fail to compile this due to symbol redefinition. NASM however, compiles this successfully. This is unfortunate...

```
loop1:
.Lloop:
    jmp .Lloop

loop2:
.Lloop:
    jmp .Lloop

main:
    jmp loop1
    jmp loop2
```

Amanieu:
> In the current implementation, local symbol names are likely to remain problematic even for naked functions. For example in LTO builds ensuring that names are unique within each crate is not sufficient to avoid duplicates.

That's a good point: if you use `.Lfoo` in one place, you cannot use that label name anywhere else in the whole program, **including in other crates**.

For this reason I prefer only using numbered local labels except when explicitly exporting a symbol.

compiler/rustc_lint/src/builtin.rs
```diff
@@ -3140,3 +3140,134 @@ impl<'tcx> LateLintPass<'tcx> for DerefNullPtr {
...
+                            } else {
+                                // Empty string means a leading ':' in this section, which is not a label
+                                break;
+                            }
+
+                            start_idx = idx + 1;
+                        }
+                    }
+
+                    debug!("NamedAsmLabels::check_expr(): found_labels: {:#?}", &found_labels);
```
Amanieu, line 3244: More leftover debugging code.

npmccallum:
> > In the current implementation, local symbol names are likely to remain problematic even for naked functions. For example in LTO builds ensuring that names are unique within each crate is not sufficient to avoid duplicates.
> 
> That's a good point: if you use `.Lfoo` in one place, you cannot use that label name anywhere else in the whole program, **including in other crates**.
> 
> For this reason I prefer only using numbered local labels except when explicitly exporting a symbol.

I accept that this unfortunate restriction is unlikely to be fixed soon. So I think we have no choice but to restrict this even on naked functions.

asquared31415:
It's trivial to cause the error even without LTO as well, simply use the same label name twice in two different `#[naked]` functions.  I think that because of this, we need to leave the lint as is, and users that are *certain* that their usage is correct and willing to accept the risks of using named labels should just set the lint to allow or warn as needed.

nbdd0121:
Regardless I think it's reasonable to move the check to a lint pass. Especially since it adds span info to `InlineAsm` so it allows people to write custom lints for their assemblies if they want to.

asquared31415:
> allows people to write custom lints for their assemblies if they want to.

Could you expand on this?  Do you mean in for example a proc macro using the AST?  I agree that it seems nicer to handle things as a lint using the AST or HIR (I personally like the HIR better), but I'm curious of the use case you're imagining, and would like to make it as good as possible for those cases.

nbdd0121:
> Could you expand on this? Do you mean in for example a proc macro using the AST? I agree that it seems nicer to handle things as a lint using the AST or HIR (I personally like the HIR better), but I'm curious of the use case you're imagining, and would like to make it as good as possible for those cases.

No, proc macros can't see ASTs, they can only see token trees. I meant custom lints using [plugins](https://doc.rust-lang.org/beta/unstable-book/language-features/plugin.html) or clippy lints.

asquared31415:
ah right

Amanieu:
If you are interested, https://github.com/rust-lang/rust/issues/81088#issuecomment-814589056 discusses some option for allowing named labels in `asm!`.

As for this PR, I agree that we should keep the lint enabled for naked functions, but it may still be beneficial to move this lint to HIR instead of macro expansion.

asquared31415:
Alright, I'll close this and reopen without the exception as a refactor

PR #88172, opened 2021-08-19
Test that type alias impl trait happens in a submodule

spastorino:
r? @oli-obk 

Related to #86727

### DIFF
--- a/src/test/ui/type-alias-impl-trait/defining-use-submodule.rs
+++ b/src/test/ui/type-alias-impl-trait/defining-use-submodule.rs
@@ -1,0 +1,23 @@
+// check-pass
+
+#![feature(type_alias_impl_trait)]
+#![allow(dead_code)]
+
+// test that the type alias impl trait defining use is in a submodule
+
+fn main() {}
+
+type Foo = impl std::fmt::Display;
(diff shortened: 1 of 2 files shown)


src/test/ui/type-alias-impl-trait/defining-use-submodule.rs
```diff
@@ -0,0 +1,27 @@
+// run-pass
+
+#![feature(type_alias_impl_trait)]
+
+// test that the type alias impl trait defining use is in a submodule
+
+fn main() {
```
jackh726, line 7: I don't think this actually needed. Could just be a check-pass test?

spastorino: Yes, was just following the intention of the original test https://github.com/rust-lang/rust/pull/88172/files#diff-f0138a92e0c5bb0a679a72ff03ec1c30cf35ab5cc36f1143efdda3c98ad91bd3L14
But agreed with you that we could just make a check-pass test instead

spastorino: Done!

PR #79749, opened 2020-12-05
Close all tags when rendering rustdoc summaries

camelid:
Otherwise we may generate HTML like `<em>hello<b>` if the text after the
`<b>` tag is too long. Now it should correctly generate
`<em>hello<b></b></em>`. It's perhaps not "ideal" that we generate an
empty `<b>` tag, but it's at least valid HTML.

It isn't a *huge* deal that we didn't close all the tags because the
browser is probably smart enough to figure out where the tags should
end, but it's still good to fix :)

r? @GuillaumeGomez
cc @jyn514

PR #88173, opened 2021-08-19
Refactor Markdown length-limited summary implementation

camelid:
This PR is a new approach to #79749.

This PR refactors the implementation of `markdown_summary_with_limit()`,
separating the logic of determining when the limit has been reached from
the actual rendering process.

The main advantage of the new approach is that it guarantees that all
HTML tags are closed, whereas the previous implementation could generate
tags that were never closed. It also ensures that no empty tags are
generated (e.g., `<em></em>`).

The new implementation consists of a general-purpose struct
`HtmlWithLimit` that manages the length-limiting logic and a function
`markdown_summary_with_limit()` that renders Markdown to HTML using the
struct.

r? @GuillaumeGomez

### DIFF
--- a/src/librustdoc/html/length_limit.rs
+++ b/src/librustdoc/html/length_limit.rs
@@ -72,7 +72,15 @@ impl HtmlWithLimit {
     }
 
     /// Open an HTML tag.
+    ///
+    /// **Note:** HTML attributes have not yet been implemented.
+    /// This function will panic if called with a non-alphabetic `tag_name`.
     pub(super) fn open_tag(&mut self, tag_name: &'static str) {
+        assert!(
+            tag_name.chars().all(|c| ('a'..='z').contains(&c)),
+            "tag_name contained non-alphabetic chars: {:?}",
+            tag_name
+        );
         self.queued_tags.push(tag_name);
     }
 
--- a/src/librustdoc/html/length_limit.rs
+++ b/src/librustdoc/html/length_limit.rs
@@ -98,3 +98,6 @@ impl HtmlWithLimit {
         }
     }
 }
+
+#[cfg(test)]
+mod tests;
--- a/src/librustdoc/html/length_limit.rs
+++ b/src/librustdoc/html/length_limit.rs
@@ -29,8 +29,18 @@ pub(super) struct HtmlWithLimit {
 impl HtmlWithLimit {
     /// Create a new buffer, with a limit of `length_limit`.
     pub(super) fn new(length_limit: usize) -> Self {
+        let buf = if length_limit > 1000 {
+            // If the length limit is really large, don't preallocate tons of memory.
+            String::new()
+        } else {
+            // The length limit is actually a good heuristic for initial allocation size.
+            // Measurements showed that using it as the initial capacity ended up using less memory
+            // than `String::new`.
+            // See https://github.com/rust-lang/rust/pull/88173#discussion_r692531631 for more.
+            String::with_capacity(length_limit)
+        };
         Self {
-            buf: String::new(),
+            buf,
             len: 0,
             limit: length_limit,
             unclosed_tags: Vec::new(),
--- a/src/librustdoc/html/length_limit.rs
+++ b/src/librustdoc/html/length_limit.rs
@@ -1,0 +1,94 @@
+//! See [`HtmlWithLimit`].
+
+use std::fmt::Write;
+use std::ops::ControlFlow;
+
+use crate::html::escape::Escape;
+
+/// A buffer that allows generating HTML with a length limit.
+///
+/// This buffer ensures that:
+///
+/// * all tags are closed,
+/// * only the most recently opened tag is closed,
+/// * no tags are left empty (e.g., `<em></em>`) due to the length limit being reached,
+/// * all text is escaped.
+#[derive(Debug)]
+pub(super) struct HtmlWithLimit {
+    buf: String,
+    len: usize,
+    limit: usize,
+    /// A list of tags that have been requested to be opened via [`Self::open_tag()`]
+    /// but have not actually been pushed to `buf` yet. This ensures that tags are not
+    /// left empty (e.g., `<em></em>`) due to the length limit being reached.
+    queued_tags: Vec<&'static str>,
+    /// A list of all tags that have been opened but not yet closed.
+    unclosed_tags: Vec<&'static str>,
+}
+
+impl HtmlWithLimit {
+    /// Create a new buffer, with a limit of `length_limit`.
+    pub(super) fn new(length_limit: usize) -> Self {
+        Self {
+            buf: String::new(),
+            len: 0,
+            limit: length_limit,
+            unclosed_tags: Vec::new(),
+            queued_tags: Vec::new(),
+        }
+    }
+
+    /// Finish using the buffer and get the written output.
+    /// This function will close all unclosed tags for you.
+    pub(super) fn finish(mut self) -> String {
+        self.close_all_tags();
+        self.buf
+    }
+
+    /// Write some plain text to the buffer, escaping as needed.
+    ///
+    /// This function skips writing the text if the length limit was reached
+    /// and returns [`ControlFlow::Break`].
+    pub(super) fn push(&mut self, text: &str) -> ControlFlow<(), ()> {
+        if self.len + text.len() > self.limit {
+            return ControlFlow::BREAK;
+        }
+
+        self.flush_queue();
+        write!(self.buf, "{}", Escape(text)).unwrap();
+        self.len += text.len();
+
+        ControlFlow::CONTINUE
+    }
+
+    /// Open an HTML tag.
+    pub(super) fn open_tag(&mut self, tag_name: &'static str) {
+        self.queued_tags.push(tag_name);
+    }
+
+    /// Close the most recently opened HTML tag.
+    pub(super) fn close_tag(&mut self) {
+        let tag_name = self.unclosed_tags.pop().unwrap();
+        self.buf.push_str("</");
+        self.buf.push_str(tag_name);
+        self.buf.push('>');
+    }
+
+    /// Write all queued tags and add them to the `unclosed_tags` list.
+    fn flush_queue(&mut self) {
+        for tag_name in self.queued_tags.drain(..) {
+            self.buf.push('<');
+            self.buf.push_str(tag_name);
+            self.buf.push('>');
+
+            self.unclosed_tags.push(tag_name);
+        }
+    }
+
+    /// Close all unclosed tags.
+    fn close_all_tags(&mut self) {
+        while !self.unclosed_tags.is_empty() {
+            self.close_tag();
+        }
+    }
+}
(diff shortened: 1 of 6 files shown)


src/librustdoc/html/markdown.rs
```diff
@@ -1101,56 +1094,49 @@ fn markdown_summary_with_limit(
     };
 
     let p = Parser::new_with_broken_link_callback(md, opts(), Some(&mut replacer));
-    let p = LinkReplacer::new(p, link_names);
+    let mut p = LinkReplacer::new(p, link_names);
 
-    'outer: for event in p {
+    // FIXME: capacity
```
camelid, line 1099: This FIXME is here because the old implementation used `String::with_capacity`. I could add a `HtmlWithLimit::with_capacity` function, but then it could be confusing which parameter is the length limit and which one is the capacity.

Also, are we sure that the capacity estimate is actually useful? We could be wasting memory by allocating too much...

src/librustdoc/html/length_limit.rs
```diff
@@ -0,0 +1,88 @@
...
+/// A buffer that allows generating HTML with a length limit.
+///
+/// This buffer ensures that:
+///
+/// * all tags are closed,
+/// * only the most recently opened tag is closed,
+/// * no tags are left empty (e.g., `<em></em>`) due to the length limit being reached,
+/// * all text is escaped.
+#[derive(Debug)]
+pub(super) struct HtmlWithLimit {
```
camelid, line 17: Would you like me to add some unit tests for `HtmlWithLimit`?

src/librustdoc/html/markdown.rs
camelid: I guess one way of getting around the potential confusion of `with_capacity` is to add a `with_buffer` function that allows the user to initialize the buffer themselves.

src/librustdoc/html/length_limit.rs
```diff
@@ -0,0 +1,94 @@
...
+use std::ops::ControlFlow;
+
+use crate::html::escape::Escape;
+
+/// A buffer that allows generating HTML with a length limit.
+///
+/// This buffer ensures that:
+///
+/// * all tags are closed,
+/// * only the most recently opened tag is closed,
```
GuillaumeGomez, line 13: Doesn't this sentence contradicts the previous one?

src/librustdoc/html/length_limit.rs
GuillaumeGomez: Yes!

src/librustdoc/html/length_limit.rs
```diff
@@ -0,0 +1,94 @@
...
+    pub(super) fn open_tag(&mut self, tag_name: &'static str) {
+        self.queued_tags.push(tag_name);
+    }
+
+    /// Close the most recently opened HTML tag.
+    pub(super) fn close_tag(&mut self) {
+        let tag_name = self.unclosed_tags.pop().unwrap();
+        self.buf.push_str("</");
+        self.buf.push_str(tag_name);
+        self.buf.push('>');
```
GuillaumeGomez, line 74: Wouldn't this be better:

```suggestion
        write!(self.buf, "</{}>", tag_name);
```

?

GuillaumeGomez:
Overall, looks pretty nice. Please add tests to ensure it works as expected. Also, not sure to see the added value of `ControlFlow` in here...

Last point: we used to create the string with a capacity whereas you don't anymore.

camelid:
> Please add tests to ensure it works as expected.

Will do. 👍 

> Also, not sure to see the added value of `ControlFlow` in here...

The value is that it allows automatically stopping the iteration, and it communicates the intent better than `Result` (because it's not an error if it stops early) and better than `bool` (because `bool` doesn't work with `try_for_each`, and because it can be unclear whether values mean "stop" or "continue").

Or are you referring to the feature gate? Note that `ControlFlow` itself has recently been stabilized, and that feature gate is just for some convenience API like `ControlFlow::BREAK = ControlFlow::Break(())`.

> Last point: we used to create the string with a capacity whereas you don't anymore.

See https://github.com/rust-lang/rust/pull/88173#discussion_r692531631.

src/librustdoc/html/length_limit.rs
camelid: It might be better in terms of readability, but it could be a bit less efficient because it requires constructing formatting arguments and `unwrap`ing the result of `write!`. I can change it if you would prefer though.

src/librustdoc/html/markdown.rs
GuillaumeGomez: We could simply allocate it with the limit as size.

GuillaumeGomez:
> The value is that it allows automatically stopping the iteration, and it communicates the intent better than `Result` (because it's not an error if it stops early) and better than `bool` (because `bool` doesn't work with `try_for_each`, and because it can be unclear whether values mean "stop" or "continue").

>Or are you referring to the feature gate? Note that `ControlFlow` itself has recently been stabilized, and that feature gate is just for some convenience API like `ControlFlow::BREAK = ControlFlow::Break(())`.

No, I meant `ControlFlow` in general. But it's mostly a nit at this point.

src/librustdoc/html/length_limit.rs
GuillaumeGomez: Depends on what we want to advantage I guess... Either way I'm fine with it.

src/librustdoc/html/length_limit.rs
camelid: Sorry, the wording might be a bit unclear. The first point means that all tags are closed *eventually*, once writing to the buffer is `finish`ed. To illustrate the second point, imagine we have this buffer:

```html
<p><em>Foo
```

If you call `close_tag`, it will close the most recently opened tag, `em`; it's not possible to close the `p` tag before closing the `em` tag, which would result in invalid HTML.

Perhaps it would be clearer if I rephrased it as follows?

```suggestion
/// * tags are closed in the reverse order of when they were opened (i.e., the correct HTML order),
```

GuillaumeGomez: Yes it'd be better. :)

Also, this code reminds me of a code I wrote in rustdoc as well... Funny that we both write the same in different places. :)

src/librustdoc/html/markdown.rs
camelid: That's a good idea! However, what if the docs are quite short, like `See [`Foo`].`? Then it would waste memory... Maybe it would be best to just add `with_buffer` for now.

GuillaumeGomez: I think it's even more of a waste to have to re-allocate a few times rather than allocate once but too much. Memory works in pages and realloc only move the value if there isn't enough space after the end of your current pointer. So all in all, I think it'd be a wiser move to allocate "too much" by default.

GuillaumeGomez: Maybe @jyn514 has an opinion on this point?

camelid: > I think it's even more of a waste to have to re-allocate a few times rather than allocate once but too much. Memory works in pages and realloc only move the value if there isn't enough space after the end of your current pointer. So all in all, I think it'd be a wiser move to allocate "too much" by default.

I tested your suggestion of using the limit as the initial capacity to see how much memory it would waste. Unfortunately, the results don't look great. When documenting `std` (and `core`, etc.), we allocate about half a megabyte (~410 kB to be more precise) more than is necessary. This is assuming that the rendered summary string is not modified after `HtmlWithLimit::finish()`; the summary string *is* sometimes modified later, when the `…` is sometimes added at the end. However, I think those extra few bytes that are occasionally added are unlikely to have a huge effect on the results.

As a percentage, rustdoc ends up allocating 34% more memory than is necessary.

Here are the raw numbers:

|||
|-|-|
| **Total number of summary string allocations** | 23538 |
| **Total bytes needed (i.e., sum of the lengths)** | 1227676 |
| **Total bytes allocated (i.e., sum of the capacities)** | 1646218 |

GuillaumeGomez: Thanks for the numbers! Do you have the equivalent with `String::new` by any chance?

camelid: No, but I can measure that.

GuillaumeGomez: Yes please. :)

camelid: Wow, good catch! The numbers are almost identical. So much so that I'm suspicious I'm measuring incorrectly, but there are slight differences:

Total # of allocs: 23538
Total length: 1227676 bytes
Total capacity: 1646218 bytes
Wasted memory: 418542 bytes (34% more than necessary)

camelid: Actually, there aren't any differences. I must have measured incorrectly...

camelid: So using `with_capacity` with the length limit is actually *more* memory efficient 😆

jyn514: I don't think this is worth worrying about, I would just use `String::new` unless this comes up in a profiler. `collect_blanket_impls` will still be 100x as slow either way :P

src/librustdoc/html/length_limit.rs
jyn514: I would prioritize readability over efficiency. This is not the slowest part of rustdoc by a long shot.

src/librustdoc/html/markdown.rs
GuillaumeGomez: Maybe, but it's not a reason to not pay attention on the rest. ;)

camelid: Yeah, it's probably not a big deal, but since we were preallocating before, I figure it's a good idea to try to keep the previous allocation behavior :)

src/librustdoc/html/length_limit.rs
camelid: Ok, I replaced this and the other similar spot with `write!`.

src/librustdoc/html/length_limit.rs
```diff
@@ -0,0 +1,221 @@
...
+
+        self.flush_queue();
+        write!(self.buf, "{}", Escape(text)).unwrap();
+        self.len += text.len();
+
+        ControlFlow::CONTINUE
+    }
+
+    /// Open an HTML tag.
+    pub(super) fn open_tag(&mut self, tag_name: &'static str) {
```
camelid, line 75: Note that `HtmlWithLimit` can't handle tags with attributes. Should I add an `assert!` to make sure the `tag_name` is only `[a-z]`?

GuillaumeGomez: This is actually a good point... Please add an `assert!` with a `FIXME`.

camelid: Is a FIXME really necessary? If you like, I can implement support for attributes in this PR. I don't think it should be too complex.

GuillaumeGomez: If you want to do it in this PR, no need for a FIXME indeed! I thought you wanted to do it in another PR.

camelid:
Ok, CI passed, this should be ready for a new review!

GuillaumeGomez (approved):

GuillaumeGomez:
Thanks!

@bors: r+

PR #88175, opened 2021-08-20
Add expansion to while desugar spans

camsteffen:
In the same vein as #88163, this reverts a change in Clippy behavior as a result of #80357 (and reverts some `#[allow]`s): This changes `clippy::blocks_in_if_conditions` to not fire on `while` loops. Though we might actually want Clippy to lint those cases, we should introduce the change purposefully, with tests, and possibly under a different lint name.

The actual change here is to add a desugaring expansion to the spans when lowering a `while` loop.

r? @Manishearth


camsteffen:
Fix formatting @bors=Manishearth

camsteffen:
Tweaked some error handling to maintain test output.

camsteffen:
Hey @Manishearth, I'm awaiting your review since I added some changes after your first review.

PR #88124, opened 2021-08-17
Split critical edge targeting the start block

tmiasko:
Fixes #88043.

PR #83417, opened 2021-03-23
Run RemoveZsts pass at mir-opt-level=1

erikdesjardins:
per https://github.com/rust-lang/rust/pull/83177#issuecomment-803942217

This pass removes assignments to ZST places.

Perf (from https://github.com/rust-lang/rust/pull/83177#issuecomment-803442557): https://perf.rust-lang.org/compare.html?start=41b315a470d583f6446599984ff9ad3bd61012b2&end=bd5d1b96f0c64c9938feea831789e1b5bb2cd4a2

r? @oli-obk

PR #88176, opened 2021-08-20
Reenable RemoveZsts

erikdesjardins:
Now that the underlying issue has been fixed by #88124, we can reland #83417.

r? @oli-obk

### DIFF
--- a/compiler/rustc_mir/src/transform/remove_zsts.rs
+++ b/compiler/rustc_mir/src/transform/remove_zsts.rs
@@ -9,9 +9,6 @@ pub struct RemoveZsts;
 
 impl<'tcx> MirPass<'tcx> for RemoveZsts {
     fn run_pass(&self, tcx: TyCtxt<'tcx>, body: &mut Body<'tcx>) {
-        if tcx.sess.mir_opt_level() < 3 {
-            return;
-        }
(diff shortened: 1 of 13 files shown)


oli-obk:
Please add the example from https://github.com/rust-lang/rust/issues/88043 as a regression test. I know we have a regression test for the unoptimized reproduction, but think we should cover both here. Note that the ui test needs to be `// build-pass` for it to run llvm

erikdesjardins:
We already have a test for both cases: I [added the RemoveZsts-dependent one](https://github.com/rust-lang/rust/commit/b865a761cbc2cfdffbcb1f60607da73172bc9772#diff-87c70697500285682e8f8270439f47072e394c7f65de858b8f3e6318e8aebdbc) when I reverted it, and `tmiasko` [added the other test to the same file](https://github.com/rust-lang/rust/commit/9a0ee05c3d7f355d00c365b9589c7924e40b19ab#diff-87c70697500285682e8f8270439f47072e394c7f65de858b8f3e6318e8aebdbc) when they fixed the underlying issue

oli-obk:
oh oops, forgot it was added in the revert, thanks!

@bors r+ rollup=never

Issue #84715, opened 2021-04-29
Tracking Issue for feature(unix_chroot)


joshtriplett:
Feature gate: `#![feature(unix_chroot)]`

This is a tracking issue for the `unix_chroot` feature, which adds `std::os::unix::fs::chroot` to change the root directory of the current process.

### Public API

```rust
// std::os::unix::fs
pub fn chroot<P: AsRef<Path>>(dir: P) -> io::Result<()>
```

### Steps / History

- [x] Implementation: #84716
- [x] Final commenting period (FCP): https://github.com/rust-lang/rust/pull/88177#issuecomment-902430723
- [x] Stabilization PR: https://github.com/rust-lang/rust/pull/88177

jplatte:
Boxes can be ticked and tracking issue closed :)

PR #88177, opened 2021-08-20
Stabilize std::os::unix::fs::chroot

joshtriplett:
I've verified that this works as documented, and I've tested it in (a nightly
build of) production software as a replacement for an unsafe call to
`libc::chroot`. It's been available in nightly for a few releases. I think it's
ready to stabilize.

---

Tracking issue: https://github.com/rust-lang/rust/issues/84715


joshtriplett:
FCP for stabilization proposal:
@rfcbot merge

joshtriplett:
FCP for stabilization proposal:
@rfcbot merge

PR #88178, opened 2021-08-20
netbsd add backtrace lib since it is a system one.

devnexen:


petrochenkov:
Does libstd use functionality from this library? Probably not if it wasn't linked previously?
We usually only link libraries that are actually required and used by libstd, not just "system" libraries.

devnexen:
netbsd being a low tier platform, problems are not arise through CI, however it s needed, like FreeBSD, for backtrace purpose.

bjorn3:
Shouldn't this be added to the backtrace crate? https://github.com/rust-lang/backtrace-rs/

yaahc:
> @yaahc any updates on this?

Sorry about the delay on this one. On this specific issue I would like to reiterate the questions @petrochenkov and @bjorn3 asked. It's not clear to me what this PR accomplishes so I don't feel comfortable approving it without getting a better understanding.

In general though since this PR is touching our tier2/tier3 platforms, in accordance with the new [target tier policy](https://doc.rust-lang.org/nightly/rustc/platform-support.html) we'd like to identify designated maintainers for netbsd who can in the future handle reviews of this nature. I'm still figuring out what the process here should look like but I'm gonna start with searching through the history to see who has historically been involved in maintaining this target (these targets?) and open an issue / PR to invite and document them as the designated maintainers.

devnexen:
apologies for the late reply .. since this PR I have a dedicated netbsd hardware and sees this change is unneeded.

PR #88187, opened 2021-08-20
Add `LetSource` to `hir::ExprKind::Let`

camsteffen:
...instead of determining `LetSource` using `hir::Map`. This is used in #87688. And it may also be useful for factoring out `Guard::IfLet`.

### DIFF
--- a/compiler/rustc_ast_lowering/src/expr.rs
+++ b/compiler/rustc_ast_lowering/src/expr.rs
@@ -1,4 +1,6 @@
-use super::{ImplTraitContext, LoweringContext, ParamMode, ParenthesizedGenericArgs};
+use super::{
+    ConditionScope, ImplTraitContext, LoweringContext, ParamMode, ParenthesizedGenericArgs,
+};
 
 use rustc_ast::attr;
 use rustc_ast::ptr::P as AstP;
@@ -14,6 +16,8 @@ use rustc_span::hygiene::ExpnId;
 use rustc_span::source_map::{respan, DesugaringKind, Span, Spanned};
 use rustc_span::symbol::{sym, Ident, Symbol};
 use rustc_span::{hygiene::ForLoopLoc, DUMMY_SP};
+
+use std::mem;
 
 impl<'hir> LoweringContext<'_, 'hir> {
     fn lower_exprs(&mut self, exprs: &[AstP<Expr>]) -> &'hir [hir::Expr<'hir>] {
@@ -87,7 +91,18 @@ impl<'hir> LoweringContext<'_, 'hir> {
                     hir::ExprKind::AddrOf(k, m, ohs)
                 }
                 ExprKind::Let(ref pat, ref scrutinee, span) => {
-                    hir::ExprKind::Let(self.lower_pat(pat), self.lower_expr(scrutinee), span)
+                    let source = match self.condition_scope {
+                        Some(ConditionScope::Guard) => hir::LetSource::Guard,
+                        Some(ConditionScope::If) => hir::LetSource::If,
+                        Some(ConditionScope::While) => hir::LetSource::While,
+                        _ => hir::LetSource::Local,
+                    };
+                    hir::ExprKind::Let(
+                        self.lower_pat(pat),
+                        self.lower_expr(scrutinee),
+                        span,
+                        source,
+                    )
                 }
                 ExprKind::If(ref cond, ref then, ref else_opt) => {
                     self.lower_expr_if(cond, then, else_opt.as_deref())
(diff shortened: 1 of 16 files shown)


michaelwoerister:
I'm not really familiar with the context here. @cjgillot seems to already be looking into the topic.
r? @cjgillot

compiler/rustc_mir_build/src/thir/pattern/check_match.rs
```diff
@@ -412,27 +394,33 @@ fn irrefutable_let_pattern(id: HirId, ls: LetSource, span: Span, tcx: TyCtxt<'_>
...
+            tcx.hir().parent_iter(id).nth(1).unwrap().1,
+            hir::Node::Expr(hir::Expr { kind: hir::ExprKind::If { .. }, .. })
+        )
+    };
     tcx.struct_span_lint_hir(IRREFUTABLE_LET_PATTERNS, id, span, |lint| match ls {
-        LetSource::GenericLet => {
-            emit_diag!(lint, "`let`", "`let` is useless", "removing `let`");
-        }
-        LetSource::IfLet => {
+        hir::LetSource::If if is_if_let() => {
```
cjgillot, line 404: In which cases can we have `LetSource::If` without `is_if_let` begin true? Should LetSource handle this case?

camsteffen: Basically if-let-chains. The relevant test is [here](https://github.com/rust-lang/rust/blob/47ab5f7ce27397310bd8359b8db1504fbf8a9b59/src/test/ui/pattern/issue-82290.stderr#L17).

compiler/rustc_ast_lowering/src/expr.rs
```diff
@@ -87,7 +91,18 @@ impl<'hir> LoweringContext<'_, 'hir> {
                     hir::ExprKind::AddrOf(k, m, ohs)
                 }
                 ExprKind::Let(ref pat, ref scrutinee, span) => {
-                    hir::ExprKind::Let(self.lower_pat(pat), self.lower_expr(scrutinee), span)
+                    let source = match self.condition_scope {
+                        Some(ConditionScope::Guard) => hir::LetSource::Guard,
+                        Some(ConditionScope::If) => hir::LetSource::If,
+                        Some(ConditionScope::While) => hir::LetSource::While,
+                        _ => hir::LetSource::Local,
```
cjgillot, line 98: Can this match be exhaustive?

compiler/rustc_ast_lowering/src/lib.rs
```diff
@@ -419,6 +419,13 @@ enum AnonymousLifetimeMode {
     PassThrough,
 }
 
+#[derive(Copy, Clone, Debug, PartialEq)]
+enum ConditionScope {
```
cjgillot, line 423: Why have separate LetSource and ConditionScope types?

camsteffen: Yeah I guess they are interchangeable. It's just that `LetSource` makes more sense in `ExprKind::Let` and `Option<ConditionScope>` makes more sense in `LoweringContext`. Should I just use `LetSource`?

compiler/rustc_mir_build/src/thir/pattern/check_match.rs
```diff
@@ -412,27 +394,33 @@ fn irrefutable_let_pattern(id: HirId, ls: LetSource, span: Span, tcx: TyCtxt<'_>
...
             emit_diag!(
                 lint,
                 "`if let`",
                 "`if let` is useless",
                 "replacing the `if let` with a `let`"
             );
         }
-        LetSource::IfLetGuard => {
+        hir::LetSource::If | hir::LetSource::Local => {
```
cjgillot, line 412: How can removing the `let` make sense for LetSource::Local?

compiler/rustc_mir_build/src/thir/pattern/check_match.rs
cjgillot: In that case, it would be better to have an if/else block inside the branch. `is_if_let` can be renamed to something like `single_let` with a comment about if-let-chains.

compiler/rustc_mir_build/src/thir/pattern/check_match.rs
camsteffen: I think that is unreachable. I'll add `hir::LetSource::Local => {}`.

camsteffen:
It occurred to me that this may be unnecessary. LetSource sort of conflates desugaring kind with HIR ancestor context. And I think the LetSource calculation can easily be moved off of the happy path.

PR #88188, opened 2021-08-20
Greatly improve limitation handling on parallel rustdoc GUI test run

GuillaumeGomez:
Follow-up of https://github.com/rust-lang/rust/pull/88082.

r? @dns2utf8

### DIFF
--- a/src/tools/rustdoc-gui/tester.js
+++ b/src/tools/rustdoc-gui/tester.js
@@ -3,6 +3,7 @@
 // ```
 // npm install browser-ui-test
 // ```
+
 const fs = require("fs");
 const path = require("path");
 const os = require('os');
@@ -172,12 +173,14 @@ async function main(argv) {
     files.sort();
 
     console.log(`Running ${files.length} rustdoc-gui tests...`);
+
     if (opts["jobs"] < 1) {
         process.setMaxListeners(files.length + 1);
     } else {
-        process.setMaxListeners(opts["jobs"]);
+        process.setMaxListeners(opts["jobs"] + 1);
     }
-    let tests = [];
+
+    const tests_queue = [];
     let results = {
         successful: [],
         failed: [],
@@ -187,8 +190,7 @@ async function main(argv) {
     for (let i = 0; i < files.length; ++i) {
         const file_name = files[i];
         const testPath = path.join(opts["tests_folder"], file_name);
-        tests.push(
-            runTest(testPath, options)
+        const callback = runTest(testPath, options)
             .then(out => {
                 const [output, nb_failures] = out;
                 results[nb_failures === 0 ? "successful" : "failed"].push({
@@ -196,10 +198,10 @@ async function main(argv) {
                     output: output,
                 });
                 if (nb_failures > 0) {
-                    status_bar.erroneous()
+                    status_bar.erroneous();
                     failed = true;
                 } else {
-                    status_bar.successful()
+                    status_bar.successful();
                 }
             })
             .catch(err => {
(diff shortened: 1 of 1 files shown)


src/tools/rustdoc-gui/tester.js
```diff
@@ -213,9 +208,12 @@ async function main(argv) {
         );
         if (no_headless) {
             await tests[i];
+        } else if (opts["jobs"] > 0 && tests.length >= opts["jobs"]) {
+            await Promise.all(tests);
```
the8472, line 212:
```suggestion
            await Promise.race(tests);
```

This should start a new test as soon as one finishes instead of waiting for all tests in a batch to finish, which may not happen at the same time.

GuillaumeGomez: It would be a nice improvement indeed!

GuillaumeGomez:
@the8472 I implemented your suggestion, thanks a lot! I discovered `Promise.race`. :)

src/tools/rustdoc-gui/tester.js
```diff
@@ -210,12 +204,17 @@ async function main(argv) {
                 status_bar.erroneous();
                 failed = true;
             })
-        );
+            .finally(() => {
+                tests.splice(tests.indexOf(callback), 1);
```
dns2utf8, line 208: I would add a comment explaining that this removes the current test from the queue, I had to lookup what this does

src/tools/rustdoc-gui/tester.js
```diff
@@ -210,12 +204,17 @@ async function main(argv) {
...
+            });
+        tests.push(callback);
         if (no_headless) {
             await tests[i];
+        } else if (opts["jobs"] > 0 && tests.length >= opts["jobs"]) {
+            await Promise.race(tests);
```
dns2utf8, line 214: Nice!

src/tools/rustdoc-gui/tester.js
```diff
@@ -172,12 +172,7 @@ async function main(argv) {
     files.sort();
 
     console.log(`Running ${files.length} rustdoc-gui tests...`);
-    if (opts["jobs"] < 1) {
-        process.setMaxListeners(files.length + 1);
```
dns2utf8, line 176: We still need this because the default is too low:
```
 (node:17758) MaxListenersExceededWarning: Possible EventEmitter memory leak detected. 11 exit listeners added to [process]. Use emitter.setMaxListeners() to increase limit
```

src/tools/rustdoc-gui/tester.js
```diff
@@ -172,12 +172,7 @@ async function main(argv) {
...
-    } else {
-        process.setMaxListeners(opts["jobs"]);
```
dns2utf8, line 178: This too, I would go with 
```javascript
        process.setMaxListeners(opts["jobs"] + 1);
```

src/tools/rustdoc-gui/tester.js
```diff
@@ -172,12 +172,7 @@ async function main(argv) {
...
-    }
-    let tests = [];
+    const tests = [];
```
dns2utf8, line 175: Maybe we should rename this symbol `running_tests` or `tests_queue` something alike

src/tools/rustdoc-gui/tester.js
dns2utf8: You can see it under the section "run the build" on line 9506

dns2utf8 (changes requested):
Looking good, I have some minor change requests

src/tools/rustdoc-gui/tester.js
GuillaumeGomez: Sure!

src/tools/rustdoc-gui/tester.js
GuillaumeGomez: True, I'll keep them then.

src/tools/rustdoc-gui/tester.js
GuillaumeGomez: Good idea.

PR #88193, opened 2021-08-20
Add ability to spawn Windows process with Proc Thread Attributes

TyPR124:
This adds a Windows-specific extension method on `Command` that allows users to include proc thread attributes. The test uses the PROC_THREAD_ATTRIBUTE_PARENT_PROCESS attribute to spawn a process with a custom parent process, and then verifies the child's parent matches the expected parent. That seemed to me the easiest way to test.

Without this, it is not possible to implement anything requiring these attributes without re-implementing `Command` (and friends) from scratch as a separate crate.

In particular, I would like to use this to make use of the Windows [ConPTY](https://devblogs.microsoft.com/commandline/windows-command-line-introducing-the-windows-pseudo-console-conpty/#using-the-conpty-api) API. I've already succesfully implemented this in my own library, but don't like the thought of having to maintain my own copy of what is already in the standard library, potentially missing any future bug fixes or security updates.

This addition requires a minimum of Windows Vista or Server 2008, which as far as I understand should now be acceptable to add to the standard library since XP support is not required any longer.

The exposed API is pretty much the bare minimum to make this work, but it may be possible to improve the API in the future.

I can open a tracking issue and update the PR with that in a bit.

### DIFF
--- a/library/std/src/process/tests.rs
+++ b/library/std/src/process/tests.rs
@@ -405,6 +405,42 @@ fn test_creation_flags() {
     }
     assert!(events > 0);
 }
+/// Tests proc thread attributes by spawning a process with a custom parent process,
+/// then comparing the parent process ID with the expected parent process ID.
+#[test]
+#[cfg(windows)]
+fn test_proc_thread_attributes() {
+    use crate::os::windows::io::AsRawHandle;
+    use crate::os::windows::process::CommandExt;
+    use crate::sys::c::{DWORD_PTR, HANDLE};
+    const PROC_THREAD_ATTRIBUTE_PARENT_PROCESS: DWORD_PTR = 0x00020000;
+    let mut parent = Command::new("cmd.exe").spawn().unwrap();
+    let mut parent_handle: HANDLE = parent.as_raw_handle();
+
+    let mut child_cmd = Command::new("cmd.exe");
+    unsafe {
+        child_cmd.process_thread_attribute(
+            PROC_THREAD_ATTRIBUTE_PARENT_PROCESS,
+            &mut parent_handle as *mut _ as *mut _,
+            crate::mem::size_of::<HANDLE>(),
+        );
+    }
+    let mut child = child_cmd.spawn().unwrap();
+
+    let wmic = format!("wmic process where processid={} get parentprocessid", child.id());
+    assert_eq!(
+        parent.id(),
+        String::from_utf8(Command::new("cmd.exe").args(["/c", &wmic]).output().unwrap().stdout)
+            .unwrap()
+            .lines()
+            .skip(1)
+            .map(|line| line.trim().parse().unwrap())
+            .next()
+            .unwrap()
+    );
+    child.kill().unwrap();
+    parent.kill().unwrap();
+}
 
 #[test]
 fn test_command_implements_send_sync() {
--- a/library/std/src/process/tests.rs
+++ b/library/std/src/process/tests.rs
@@ -412,17 +412,15 @@ fn test_creation_flags() {
 fn test_proc_thread_attributes() {
     use crate::os::windows::io::AsRawHandle;
     use crate::os::windows::process::CommandExt;
-    use crate::sys::c::{DWORD_PTR, HANDLE};
+    use crate::sys::c::DWORD_PTR;
     const PROC_THREAD_ATTRIBUTE_PARENT_PROCESS: DWORD_PTR = 0x00020000;
-    let mut parent = Command::new("cmd.exe").spawn().unwrap();
-    let mut parent_handle: HANDLE = parent.as_raw_handle();
 
+    let mut parent = Command::new("cmd.exe").spawn().unwrap();
     let mut child_cmd = Command::new("cmd.exe");
     unsafe {
         child_cmd.process_thread_attribute(
             PROC_THREAD_ATTRIBUTE_PARENT_PROCESS,
-            &mut parent_handle as *mut _ as *mut _,
-            crate::mem::size_of::<HANDLE>(),
+            parent.as_raw_handle() as isize,
         );
     }
     let mut child = child_cmd.spawn().unwrap();
--- a/library/std/src/sys/windows/process.rs
+++ b/library/std/src/sys/windows/process.rs
@@ -16,6 +16,7 @@ use crate::num::NonZeroI32;
 use crate::os::windows::ffi::{OsStrExt, OsStringExt};
 use crate::os::windows::io::{AsRawHandle, FromRawHandle, IntoRawHandle};
 use crate::path::{Path, PathBuf};
+use crate::pin::Pin;
 use crate::ptr;
 use crate::sys::c;
 use crate::sys::c::NonZeroDWORD;
@@ -166,6 +167,7 @@ pub struct Command {
     stdout: Option<Stdio>,
     stderr: Option<Stdio>,
     force_quotes_enabled: bool,
+    proc_thread_attributes: BTreeMap<usize, ProcThreadAttributeValue>,
 }
 
 pub enum Stdio {
@@ -202,6 +204,7 @@ impl Command {
             stdout: None,
             stderr: None,
             force_quotes_enabled: false,
+            proc_thread_attributes: Default::default(),
         }
     }
 
@@ -251,6 +254,9 @@ impl Command {
     pub fn get_current_dir(&self) -> Option<&Path> {
         self.cwd.as_ref().map(|cwd| Path::new(cwd))
     }
+    pub fn process_thread_attribute(&mut self, attribute: usize, ptr: *mut c_void, size: usize) {
+        self.proc_thread_attributes.insert(attribute, ProcThreadAttributeValue { ptr, size });
+    }
 
     pub fn spawn(
         &mut self,
@@ -259,9 +265,9 @@ impl Command {
     ) -> io::Result<(Process, StdioPipes)> {
         let maybe_env = self.env.capture_if_changed();
 
-        let mut si = zeroed_startupinfo();
-        si.cb = mem::size_of::<c::STARTUPINFO>() as c::DWORD;
-        si.dwFlags = c::STARTF_USESTDHANDLES;
+        let mut si = zeroed_startupinfo_ex();
+        si.StartupInfo.cb = mem::size_of::<c::STARTUPINFO>() as c::DWORD;
+        si.StartupInfo.dwFlags = c::STARTF_USESTDHANDLES;
 
         let child_paths = if let Some(env) = maybe_env.as_ref() {
             env.get(&EnvKey::new("PATH")).map(|s| s.as_os_str())
@@ -305,9 +311,24 @@ impl Command {
         let stdin = stdin.to_handle(c::STD_INPUT_HANDLE, &mut pipes.stdin)?;
         let stdout = stdout.to_handle(c::STD_OUTPUT_HANDLE, &mut pipes.stdout)?;
         let stderr = stderr.to_handle(c::STD_ERROR_HANDLE, &mut pipes.stderr)?;
-        si.hStdInput = stdin.as_raw_handle();
-        si.hStdOutput = stdout.as_raw_handle();
-        si.hStdError = stderr.as_raw_handle();
+        si.StartupInfo.hStdInput = stdin.as_raw_handle();
+        si.StartupInfo.hStdOutput = stdout.as_raw_handle();
+        si.StartupInfo.hStdError = stderr.as_raw_handle();
+
+        let mut attributes = if !self.proc_thread_attributes.is_empty() {
+            Some(make_proc_thread_attributes(&self.proc_thread_attributes)?)
+        } else {
+            None
+        };
+        si.lpAttributeList = attributes.as_mut().map_or(ptr::null_mut(), |buf| {
+            // Indicate that lpAttributeList is present by setting
+            // EXTENDED_STARTUPINFO_PRESENT and adjusting the size
+            // value of the struct.
+            flags |= c::EXTENDED_STARTUPINFO_PRESENT;
+            si.StartupInfo.cb = mem::size_of::<c::STARTUPINFOEX>() as u32;
+
+            buf.0.as_mut_ptr().cast()
+        });
 
         let program = to_u16s(&program)?;
         unsafe {
@@ -320,7 +341,7 @@ impl Command {
                 flags,
                 envp,
                 dirp,
-                &mut si,
+                &mut si as *mut _ as *mut _,
                 &mut pi,
             ))
         }?;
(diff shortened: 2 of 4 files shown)


ChrisDenton:
I would definitely love to see more of `CreateProcessW` exposed to users.

> The exposed API is pretty much the bare minimum to make this work, but it may be possible to improve the API in the future.

Hm, it does feel like `Command` should ideally take ownership of the value somehow.

Other than that it would be good to have safe (or at least safer) APIs for common operations. E.g. setting the parent process, setting exploit mitigations and restricting inherited handles.

library/std/src/process/tests.rs
```diff
@@ -393,6 +393,42 @@ fn test_creation_flags() {
...
+/// Tests proc thread attributes by spawning a process with a custom parent process,
+/// then comparing the parent process ID with the expected parent process ID.
+#[test]
+#[cfg(windows)]
+fn test_proc_thread_attributes() {
+    use crate::os::windows::io::AsRawHandle;
+    use crate::os::windows::process::CommandExt;
+    use crate::sys::c::{DWORD_PTR, HANDLE};
+    const PROC_THREAD_ATTRIBUTE_PARENT_PROCESS: DWORD_PTR = 0x00020000;
+    let parent = Command::new("cmd.exe").spawn().unwrap();
```
bjorn3, line 405: Do these processes need to be killed before the end of the test?

library/std/src/sys/windows/process.rs
```diff
@@ -682,6 +706,57 @@ fn make_dirp(d: Option<&OsString>) -> io::Result<(*const u16, Vec<u16>)> {
...
+unsafe impl Sync for ProcThreadAttributeValue {}
+
+fn make_proc_thread_attributes(
+    attributes: &BTreeMap<usize, ProcThreadAttributeValue>,
+) -> io::Result<ProcThreadAttributeList> {
+    let mut buffer_size = 0;
+    let count = attributes.len() as u32;
+    // First call to get required size. Error is expected.
+    unsafe { c::InitializeProcThreadAttributeList(ptr::null_mut(), count, 0, &mut buffer_size) };
+    let mut buf = Pin::new(crate::vec::from_elem(0u8, buffer_size as usize));
```
ChrisDenton, line 735: Does this need to be `Pin`? Surely the `Vec` itself can be moved around freely?

library/std/src/process/tests.rs
TyPR124: They probably should be. When I ran the tests locally I did not have any lingering cmd.exe's afterwards. But, I'll add that explicitly anyway.

library/std/src/sys/windows/process.rs
TyPR124: The `Pin` prevents us from calling (for example) vec.push() or doing anything that may re-allocate (and move) the underlying data. I'm not 100% sure it is necessary to not move the data, as I don't know the internal workings of these Initialize/Update/Delete functions, but it seemed the right choice to me since there is at least no good reason to move or grow the data after that point.

The `Pin` doesn't prevent us from moving the Vec itself (after all, we must move it out of this function to return it), just the data it points to.

TyPR124: > Hm, it does feel like `Command` should ideally take ownership of the value somehow.

I think I agree. I am considering making the method generic over a `T: Sized + Any`, in which case we could grab the size ourselves and store the `T` in some `Box<dyn Any>`, making it slightly easier on the user. A couple concerns I have about that are:

- Do we really need to allocate for every attribute?
- Do we really need ownership of the generic type? Are there any situations where the user might need to retain ownership?
    - I'm *guessing* that any value passed in here would always be `Copy` so this shouldn't be an issue, but I'm not certain of that.

That said, the more I think about it, the more I tend to think taking ownership would be the right choice.

> Other than that it would be good to have safe (or at least safer) APIs

I'm skeptical this could ever be made fully safe, however it will certainly be possible (and I'd say, encouraged) to build safe APIs on top of this.

The reason I'm skeptical basically comes down to the variety of different attributes a user could pass in and the fact that MS could add new attributes at any time. For instance, if MS added something which allowed the user to have some of the parent process's memory mapped into the child's address space, that would immediately preclude this from ever being truly safe.

My hope is that this extension method, it will enable authors of 3rd party libraries to make safe APIs via their own extension traits, even if those safe APIs never make it into the standard library.

TyPR124:
> Hm, it does feel like `Command` should ideally take ownership of the value somehow.

I think I agree. I am considering making the method generic over a `T: Sized + Any`, in which case we could grab the size ourselves and store the `T` in some `Box<dyn Any>`, making it slightly easier on the user. A couple concerns I have about that are:

- Do we really need to allocate for every attribute?
- Do we really need ownership of the generic type? Are there any situations where the user might need to retain ownership?
    - I'm *guessing* that any value passed in here would always be `Copy` so this shouldn't be an issue, but I'm not certain of that.

That said, the more I think about it, the more I tend to think taking ownership would be the right choice.

> Other than that it would be good to have safe (or at least safer) APIs

I'm skeptical this could ever be made fully safe, however it will certainly be possible (and I'd say, encouraged) to build safe APIs on top of this.

The reason I'm skeptical basically comes down to the variety of different attributes a user could pass in and the fact that MS could add new attributes at any time. For instance, if MS added something which allowed the user to have some of the parent process's memory mapped into the child's address space, that would immediately preclude this from ever being truly safe.

My hope is that this extension method, it will enable authors of 3rd party libraries to make safe APIs via their own extension traits, even if those safe APIs never make it into the standard library.

library/std/src/sys/windows/process.rs
bjorn3: `u8` is `Unpin`, so `Pin<Vec<u8>>` so both `Pin::into_inner` and `DerefMut` are implemented allowing you to move the inner bytes anyway.

https://doc.rust-lang.org/stable/std/pin/struct.Pin.html#method.into_inner

ChrisDenton:
Yes sorry, that was meant to be a separate thought. I was meaning it would be possible to have a safe API for common operations such as a `parent_handle` or `exploit_mitigation` method. I agree that arbitrary ones would be impossible because we can't know what future attributes will be added and a lower-level method will always be useful for that.

library/std/src/sys/windows/process.rs
ChrisDenton: I think perhaps a boxed slice would be enough to prevent accidentally reallocating? Or it might be a good idea to fully contain the owned data. So have an API like:

```rust
struct ProcThreadAttributeList(Vec<u8>); // or Box<[T]>
impl ProcThreadAttributeList {
    fn new(attribute_count: u32) -> io::Result<Self> { ... }
    unsafe fn update(&mut self, attribute: usize, value_ptr: *mut c_void, value_size: u32) -> io::Result<()> { ... }
}
impl Drop for ProcThreadAttributeList { ... }
```

That way there is no reason to touch the inner data.

TyPR124: @ChrisDenton I could do that, just seemed like overkill to me when I was writing it.

@bjorn3 You could copy the bytes, or overwrite them, but you can't move them. `let mut v = Pin::new(vec![]); v.push(());` does not compile. Unless I'm just being really dumb right now. You are right about `Pin::into_inner` but calling that explicitly indicates it is now okay to move the underlying data (which it might be anyway, but, again, I'm not certain).

bjorn3:
Can you please use rebase instead of merge?

Mark-Simulacrum:
r? @yaahc for T-libs-api review

I'm not familiar enough with Windows to review the impl either.

TyPR124:
I will try to fix the git history sometime this week (or this weekend) and also I still need to open a tracking issue.

TyPR124:
While considering what it might look like to make the method generic over the value type, I noticed that there is a difference in object safety between unix and windows `CommandExt`. Is this something to be concerned about? Windows is object safe, unix is not.

```rust
#[cfg(test)]
fn test_object_safety() {
    #[cfg(windows)]
    use std::os::windows::process::CommandExt;
    #[cfg(unix)]
    use std::os::unix::process::CommandExt;
    fn takes_dyn_command_ext(_: &dyn CommandExt) {}
    takes_dyn_command_ext(&std::process::Command::new(""));
}
```

Currently, if I were to add a generic method for Windows (without `Self: Sized`), it silently breaks object safety.

Looking at methods on the unix side, it appears to me the trait was originally object-safe in 1.0, but that was broken in 1.15 with `before_exec`, and then later `pre_exec` and `arg0`.

TyPR124:
> Can you please use rebase instead of merge?

I think this should be good now. Let me know if there is anything else I should do.

TyPR124:
I've gone ahead and updated this to take ownership of a generic value type on each use. The `T: Copy` bound could be removed, but it seemed reasonable since Windows API will presumably need to make a copy of the value.

This also breaks object safety of the windows trait - I decided not to care about this since it is already broken for Unix and I don't think there is any practical use for `dyn CommandExt` over `Command`

Dylan-DPC:
@TyPR124 you have conflicts to be resolved

yaahc:
> I've gone ahead and updated this to take ownership of a generic value type on each use. The `T: Copy` bound could be removed, but it seemed reasonable since Windows API will presumably need to make a copy of the value.
> 
> This also breaks object safety of the windows trait - I decided not to care about this since it is already broken for Unix and I don't think there is any practical use for `dyn CommandExt` over `Command`

This is the correct call. These traits are sealed so there should in practice only ever be one implementor of each, so using them as trait object doesn't make sense. We can make this explicit on the trait docs so people don't end up writing `&dyn CommandExt` but I don't expect this to become a problem in practice. 

Also, I'm sorry this has taken so long but I'm ready to start reviewing this PR now. I'm admittedly in a similar situation to @Mark-Simulacrum where I'm not an expert on windows APIs but I'll do my best. Though we should absolutely also get the windows ping group involved as well to help with reviewing the implementation itself.

@rustbot ping windows

---

My first question is around the safety of the interface.

> The reason I'm skeptical basically comes down to the variety of different attributes a user could pass in and the fact that MS could add new attributes at any time. For instance, if MS added something which allowed the user to have some of the parent process's memory mapped into the child's address space, that would immediately preclude this from ever being truly safe.

My first reaction upon looking up the windows docs for ProcThreadAttributes[^1] and the signature of `process_thread_attribute` was to wonder if instead of having the input be a `usize` and generic value `T` was that it looked like the interface should be an enum rather than a usize to better preserve the semantic meaning of the value. Do you think we could instead provide the input with something like this? 

```rust
#[non_exhaustive]
enum ProcThreadAttribute {
    ParentProcess(&Command),
    // other attributes we could use safely ...
}
```

[^1]: https://docs.microsoft.com/en-us/windows/win32/api/processthreadsapi/nf-processthreadsapi-updateprocthreadattribute

TyPR124:
Thank you for looking at this! I will try to get around to fixing conflicts this weekend.

> Do you think we could instead provide the input with something like this?

Honestly, I don't think this is feasible for a few reasons.

1. This would require std to keep up to date with all attributes, else it becomes useless. Anyone wanting a new attribute will have to implement it here, and if anything prevents a quick merge you have to re-implement Command entirely to get around it.
2. There are attributes which are not documented on that page, such as [`PROC_THREAD_ATTRIBUTE_PSEUDOCONSOLE`](https://devblogs.microsoft.com/commandline/windows-command-line-introducing-the-windows-pseudo-console-conpty/) (which is actually the original motivator behind this)
3. Some attributes only support certain Windows versions (I'm not sure what happens if you try those attributes in such a case, if it is even considered defined behavior)
4. There is a pretty wide variety of attributes. Should stdlib really have APIs around spectre mitigations? Pseudoconsoles? If there is commitment to support these features in Windows, is there commitment to support these features in other OSs?

Another thing to consider (which this PR does not attempt to do) is that these attributes can be applied to thread handles as well. I imagine whatever interface is adopted here would be adopted for threads eventually as well.

yaahc:
I agree with the idea of renaming to call it the raw attribute interface to specifically leave room for a safe interface in the future, even if we don't have to design that interface as part of this PR

TyPR124:
I am having a very hard time with pulling most recent master... so just hoping it works to fix the conflicts through github. Not sure if the new commit is an issue.

Kobzol:
Make sure that you don't have any pending changes in your workspace and then run this (assuming that you are on the branch of this PR):
```
$ git fetch origin master
$ git rebase origin/master
```
Then you need to fix conflicts, and run
```
$ git add -u
$ git rebase --continue
$ git push --force-with-lease
```

TyPR124:
If my remotes look like this:

```
❯ git remote -v
origin  git@github.com:TyPR124/rust.git (fetch)
origin  git@github.com:TyPR124/rust.git (push)
upstream        https://github.com/rust-lang/rust (fetch)
upstream        https://github.com/rust-lang/rust (push)
```

Then should I replaced 'origin' with 'upstream' in those commands?

I was previously trying to follow https://rustc-dev-guide.rust-lang.org/git.html which to fix merge conflicts says "First, get a local copy of the conflicting changes: Checkout your local master branch with `git checkout master`, then `git pull upstream master`" but when I do that I just get pain. Does that info need updated?

Kobzol:
Yeah if you named the rustc repo upstream you should use upstream in the commands above.

Your master branch probably got out of sync with the upstream master. In that case just switch to master (`git checkout master`) and forcefully set it to the upstream master branch (`git reset --hard upstream/master`).

Dylan-DPC:
@TyPR124 git pull does a merge (git pull is effectively fetch + merge) 

next time just add the `-r` flag to it which will do a rebase instead of a merge (i.e. `git pull -r <remote>/<branch>`) 

[Docs on --rebase](https://git-scm.com/docs/git-pull#Documentation/git-pull.txt---rebasefalsetruemergesinteractive)

TyPR124:
Sorry, it's been a while since I've been able to put much time into this.

I've found that the test I wrote is racy. I'm not sure exactly why yet. If I apply this PR to nightly-2021-08-18, the test passes every time I run it (though I'm not 100% convinced that it *never* fails on this version). At some point between 2021-11-18 and 2021-12-18, it seems like the issue was either introduced or made more likely to cause test failure.

A workaround that seems to be 100% effective in my testing is spawn the second (child) process with `sleep 1`. Without this, it seems that the child process dies before the wmic command runs, so wmic fails to find the process.

Re: renaming the method. I don't think `process_raw_thread_attribute` makes sense here. The name `PROCESS_THREAD_ATTRIBUTE` is, as far as I know, is becuase these are attributes that can be applied to either process' or threads. Since the context it is being used here is specifically for process spawning, I think just `raw_attribute` will be better - and I can't think of any reason why we'd need a `[raw-]attribute()` method other than this purpose.

ChrisDenton:
> it seems that the child process dies before the wmic command runs, so wmic fails to find the process

Could you add the wmic command to the child process itself?

TyPR124:
> `$(Get-Process -id $PID).Parent.Id`

Good idea, but for me it only works in pwsh; not native powershell. I imagine that would be a problem.

`wmic process where processid=$PID get parentprocessid` seems to work in both, so I might give that a go.

lzybkr:
> > `$(Get-Process -id $PID).Parent.Id`
> 
> Good idea, but for me it only works in pwsh; not native powershell. I imagine that would be a problem.
> 
> `wmic process where processid=$PID get parentprocessid` seems to work in both, so I might give that a go.

If you'd like a PowerShell option that works in powershell and pwsh - you can use:

```powershell
(Get-CimInstance Win32_Process -Filter "ParentProcessId=$PID").ProcessId
```

This is still using WMI under the covers, just not invoking the wmic command line tool.

EddieIvan01:
So, is there an expected plan for merging this PR?

yaahc:
> So, is there an expected plan for merging this PR?

I believe it's just waiting on the author to update the PR to rename the method and create a tracking issue.

Dylan-DPC:
@TyPR124 any updates?

JohnCSimon:
@TyPR124 
Thanks for your contribution.

@rustbot label: +S-inactive

PR #88196, opened 2021-08-20
Refactor `named_asm_labels` to a HIR lint

asquared31415:
As discussed on #88169, the `named_asm_labels` lint could be moved to a HIR lint.  That allows future lints or custom plugins or clippy lints to more easily access the `asm!` macro's data and create better error messages with the lints.

### DIFF
--- a/compiler/rustc_ast/src/visit.rs
+++ b/compiler/rustc_ast/src/visit.rs
@@ -19,20 +19,20 @@ use crate::token;
 use rustc_span::symbol::{Ident, Symbol};
 use rustc_span::Span;
 
-#[derive(Copy, Clone, PartialEq)]
+#[derive(Copy, Clone, Debug, PartialEq)]
 pub enum AssocCtxt {
     Trait,
     Impl,
 }
 
-#[derive(Copy, Clone, PartialEq)]
+#[derive(Copy, Clone, Debug, PartialEq)]
 pub enum FnCtxt {
     Free,
     Foreign,
     Assoc(AssocCtxt),
 }
 
-#[derive(Copy, Clone)]
+#[derive(Copy, Clone, Debug)]
 pub enum FnKind<'a> {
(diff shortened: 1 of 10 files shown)


asquared31415:
cc @bjorn3 @nbdd0121 @npmccallum

compiler/rustc_ast/src/visit.rs
```diff
@@ -19,20 +19,20 @@ use crate::token;
 use rustc_span::symbol::{Ident, Symbol};
 use rustc_span::Span;
 
-#[derive(Copy, Clone, PartialEq)]
+#[derive(Copy, Clone, Debug, PartialEq)]
```
nbdd0121, line 22: Are these `Debug`s necessary?

asquared31415: No they are not, but I was under the impression that things that *can* be `#[derive(Debug)]` should be, for ease of use for other people.  I had to add the derive to see some of the stuff that was passed to the ast early lint when I had an early pass version.

compiler/rustc_builtin_macros/src/asm.rs
```diff
@@ -524,8 +522,16 @@ fn expand_preparsed_asm(
             ast::StrStyle::Raw(raw) => Some(raw as usize),
         };
 
-        let template_str = &template_str.as_str();
         let template_snippet = ecx.source_map().span_to_snippet(template_sp).ok();
+        template_strs.push((
+            template_str,
+            // This should never fail, every template that exists has a span and has a souce snippet that is part of that span
```
Amanieu, line 528: I think this can fail if the span comes from another crate for which we only have a rlib and no source code. This would be something like a `macro_rules!` in that crate which calls `asm!` with a string literal, and that macro is invoked from the current crate.

asquared31415: ah :/

PR #88197, opened 2021-08-20
Test tait use in a fn type

spastorino:
r? @oli-obk 

I thought this was going to work but doesn't, quickly checked with Niko and he told me that we ruled this out for now. I'm not exactly sure why and how but here we have a test with a FIXME :)

Related to #86727


oli-obk:
Probably because of all the problems around late bound lifetimes on function pointers

@bors r+ rollup

PR #81642, opened 2021-02-01
Automatically implement AsRepr and allow deriving FromRepr for fieldless enums

illicitonion:
This auto-impl and derive are only allowed when all of the following are true:
1. The derive is on an enum.
2. The enum has an explicit repr (in an aim to show the deriver has
   considered what type this enum should be convertable into).
3. The repr is a fixed-size integer type (i.e. not `C`).
4. None of the enum's variants may hold data.
5. None of the enum's variants are declared as holding an empty tuple as
   data (because for `enum E { Empty() }`, `E::Empty as isize` converts
   the address of the function pointer, rather than the discriminant).
6. None of the enum's variants are declared as holding empty named
   values (i.e. `enum E { Empty{} }`) for consistency.

Currently this tends to be done using `as`, which has two flaws:
1. It will silently truncate. e.g. if you have a `repr(u64)` enum, and
   you write `value as u8` (perhaps because the repr changed), you get
   no warning or error that truncation may occur.
2. You cannot use enums in code which is generic over `From` or `Into`.

Other types, such as primitive numeric types, support both `as` casts
and `From`/`Into` conversions, so making this trivial for primitive
enums too seems reasonable.

There are a number of crates which provide macros to provide this
implementation, as well as `TryFrom` implementations. The advantages of
using infallible conversions, rather than possibly-truncating ones feels
like it warrants a place in `std`.

PR #88203, opened 2021-08-21
Make specifying repr optional for fieldless enums

fee1-dead:
This makes it consistent for another behavior, which is casting an enum to its discriminant.

We can currently cast "fieldless" enums to its discriminants, but only "c-like" (only unit variants) enums are allowed to leave the repr unspecified. See https://github.com/rust-lang/reference/pull/1055#discussion_r691748964 for previous discussion.

### DIFF
--- a/compiler/rustc_error_codes/src/error_codes/E0732.md
+++ b/compiler/rustc_error_codes/src/error_codes/E0732.md
@@ -4,16 +4,19 @@ Erroneous code example:
 
 ```compile_fail,E0732
 enum Enum { // error!
-    Unit = 1,
-    Tuple() = 2,
-    Struct{} = 3,
+    Unit = 3,
+    Tuple(u16) = 2,
+    Struct {
+        a: u8,
+        b: u16,
+    } = 1,
 }
 # fn main() {}
 ```
 
-A `#[repr(inttype)]` must be provided on an `enum` if it has a non-unit
-variant with a discriminant, or where there are both unit variants with
-discriminants and non-unit variants. This restriction ensures that there
+A `#[repr(inttype)]` must be provided on an `enum` if it not fieldless and
+has a discriminant, or where there are both fieldless variants with
+discriminants and variants that have fields. This restriction ensures that there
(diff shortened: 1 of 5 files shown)


joshtriplett:
I don't believe we should allow enums with struct or tuple variants (even if they don't have fields) to omit the repr.

(Ideally, I'd also prefer if such enums couldn't be `as`-cast to their discriminant, but that seems like it'd break backwards compatibility.)

Personally, I'm hoping that https://github.com/rust-lang/rust/pull/81642 helps improve this substantially, so that using `as` becomes less necessary.

ehuss:
@fee1-dead Do you think you'd be able to prepare a PR that rejected casting enums with 0-field tuple/struct variants?  (Per the discussion at https://rust-lang.zulipchat.com/#narrow/stream/213817-t-lang/topic/arbitrary.20enum.20discriminants/near/251899236)  I can maybe give it a try myself if you'd like.

fee1-dead:
I'm busy with other stuff. You can do it if you'd like.

JohnCSimon:
merge conflicts

pnkfelix:
discussed at T-compiler meeting; we think this is waiting on lang team decision regarding e.g.  #88621 and PR #89234

Lokathor:
Hello. This was discussed at the lang team meeting today.

* The main open question was if this change affects Stable or not. Currently, having a discriminant on a non-unit enum variant requires the `arbitrary_enum_discriminant` feature. This change should likely be "no more stable" than that feature, and it's unclear.
* Also, would `as` casting work for enums declared like this? In the meeting it was assumed "no".

fee1-dead:
> would as casting work for enums declared like this? In the meeting it was assumed "no".

Actually, the answer is "yes". https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=f38e9d525cd1ca67ea8836aa65d6f7b9

```rust
#[repr(u8)]
pub enum Foo {
    A,
    B(),
    C{},
}

pub fn a(foo: Foo) -> u8 {
    foo as u8
}
```

PR #88209, opened 2021-08-21
Improve error message when _ is used for in/inout asm operands

Amanieu:
As suggested by @Commeownist in https://github.com/rust-lang/rust/issues/72016#issuecomment-903102415.

### DIFF
--- a/compiler/rustc_builtin_macros/src/asm.rs
+++ b/compiler/rustc_builtin_macros/src/asm.rs
@@ -117,6 +117,10 @@ fn parse_args<'a>(
         let mut explicit_reg = false;
         let op = if !is_global_asm && p.eat_keyword(kw::In) {
             let reg = parse_reg(&mut p, &mut explicit_reg)?;
+            if p.eat_keyword(kw::Underscore) {
+                let err = ecx.struct_span_err(p.token.span, "_ cannot be used for input operands");
+                return Err(err);
+            }
             let expr = p.parse_expr()?;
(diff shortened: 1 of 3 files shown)


nagisa:
This seems like a more general problem plaguing `_` uses in an expression position. Like

```
error[E0658]: destructuring assignments are unstable
 --> src/main.rs:3:5
  |
3 |     _;
  |     ^
  |
  = note: see issue #71126 <https://github.com/rust-lang/rust/issues/71126> for more information

error: in expressions, `_` can only be used on the left-hand side of an assignment
 --> src/main.rs:3:5
  |
3 |     _;
  |     ^ `_` not allowed here
```

cc @estebank might be of interest to you.

---

Overall this seems like a good place to provide `asm!` specific diagnostic. I would perhaps have chosen to word this as "_ is not a valid input operand" or somesuch, but this seems good regardless.

@bors r+

nagisa:
Ah well, another thing that would be nice is some deduplication. I would perhaps extract 

```
            if p.eat_keyword(kw::Underscore) {
                let err = ecx.struct_span_err(p.token.span, "_ cannot be used for input operands");
                return Err(err);
            }
            let expr = p.parse_expr()?;
```

into a `parse_asm_input_operand` method or somesuch.

estebank:
@nagisa weird

https://github.com/rust-lang/rust/blob/77ac329a08b50090c9ac387f3d24ddc46bced92e/compiler/rustc_parse/src/parser/expr.rs#L1182-L1184

PR #88214, opened 2021-08-21
rustc: use more correct span data in for loop desugaring

### LINKED ISSUE #82462: Diagnostic for iterator referencing temporary with Drop may suggest an invalid semicolon
Given the following code on rust nightly (current nightly on playground is 2021-02-22 a15f484b918a4533ad63):

https://play.rust-lang.org/?version=nightly&mode=debug&edition=2018&gist=286e1d81ecdd9df68dc94e9b1b8ead5f

```rust
struct DroppingSlice<'a>(&'a [i32]);

impl Drop for DroppingSlice<'_> {
    fn drop(&mut self) {
        println!("hi from slice");
    }
}

impl DroppingSlice<'_> {
    fn iter(&self) -> std::slice::Iter<'_, i32> {
        self.0.iter()
    }
}

fn main() {
    let mut v = vec![1, 2, 3, 4];
    for x in DroppingSlice(&*v).iter() {
        v.push(*x);
        break;
    }
}
```

The current output is:

```
error[E0502]: cannot borrow `v` as mutable because it is also borrowed as immutable
  --> src/main.rs:18:9
   |
17 |     for x in DroppingSlice(&*v).iter() {
   |              ------------------      - ... and the immutable borrow might be used here, when that temporary is dropped and runs the `Drop` code for type `DroppingSlice`
   |              |               |
   |              |               immutable borrow occurs here
   |              a temporary with access to the immutable borrow is created here ...
18 |         v.push(*x);
   |         ^^^^^^^^^^ mutable borrow occurs here
   |
help: consider adding semicolon after the expression so its temporaries are dropped sooner, before the local variables declared by the block are dropped
   |
17 |     for x in DroppingSlice(&*v).iter(); {
   |                                       ^

For more information about this error, try `rustc --explain E0502`.

```

The suggestion regarding the semicolon is incorrect.

Additionally, it seems a little odd to me that the first half of the sentence beginning "a temporary with access to the immutable borrow is created here ..." is displayed below the second half, "... and the immutable borrow might be used here, when that temporary is dropped and runs the `Drop` code for type `DroppingSlice`".

notriddle:
Fixes #82462

Before:

      help: consider adding semicolon after the expression so its temporaries are dropped sooner, before the local variables declared by the block are dropped
         |
      LL |     for x in DroppingSlice(&*v).iter(); {
         |                                       +

After:

      help: consider adding semicolon after the expression so its temporaries are dropped sooner, before the local variables declared by the block are dropped
         |
      LL |     };
         |      +

This seems like a reasonable fix: since the desugared "expr_drop_temps_mut" contains the entire desugared loop construct, its span should contain the entire loop construct as well.

### DIFF
--- a/compiler/rustc_ast_lowering/src/expr.rs
+++ b/compiler/rustc_ast_lowering/src/expr.rs
@@ -1401,8 +1401,13 @@ impl<'hir> LoweringContext<'_, 'hir> {
             )
         };
 
+        // #82462: to correctly diagnose borrow errors, the block that contains
+        // the iter expr needs to have a span that covers the loop body.
+        let desugared_full_span =
+            self.mark_span_with_reason(DesugaringKind::ForLoop(ForLoopLoc::Head), e.span, None);
+
         let match_expr = self.arena.alloc(self.expr_match(
-            desugared_span,
+            desugared_full_span,
             into_iter_expr,
             arena_vec![self; iter_arm],
             hir::MatchSource::ForLoopDesugar,
@@ -1416,7 +1421,7 @@ impl<'hir> LoweringContext<'_, 'hir> {
         // surrounding scope of the `match` since the `match` is not a terminating scope.
         //
         // Also, add the attributes to the outer returned expr node.
-        self.expr_drop_temps_mut(desugared_span, match_expr, attrs.into())
+        self.expr_drop_temps_mut(desugared_full_span, match_expr, attrs.into())
     }
 
     /// Desugar `ExprKind::Try` from: `<expr>?` into:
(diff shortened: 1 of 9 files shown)


nagisa:
On one hand this looks like a nice improvement in that we no longer suggest non-sense, but on the other adding a semicolon there won't help either way (and the message sounds quite confident it would help…)

Please squash. r=me.

notriddle:
@nagisa okay, it's squashed

notriddle:
I guess I'll rebase onto the master branch? I'm not actually sure what's wrong here.

nagisa:
Did you re-generate the `.stderr` file after the rebase? Bors will always merge in master when testing a diff and this is the diff from the test output, while the PRd `stderr` file still contains the larger span.

```
---- [ui (nll)] ui/borrowck/issue-82462.rs stdout ----
diff of stderr:

7	   |              |               immutable borrow occurs here
8	   |              a temporary with access to the immutable borrow is created here ...
9	LL |         v.push(*x);
-	   |         ^^^^^^^^^^ mutable borrow occurs here
+	   |         ^ mutable borrow occurs here
11	LL |         break;
12	LL |     }
13	   |     - ... and the immutable borrow might be used here, when that temporary is dropped and runs the `Drop` code for type `DroppingSlice`
```

notriddle:
Yes, I tried to regenerate it and nothing happened.

nagisa:
Hm, since this error is generated by borrowck, it could be that we inadvertently run some MIR pass before borrowck that we don't run in non-optimized builds?

notriddle:
Found it. It's actually caused by `test-compare-mode`, and not the optimization stack.

I've added a second stderr file for this.

notriddle:
@nagisa New commit to fix the broken Clippy lints.

Issue #84738, opened 2021-04-30
Index out of bounds when running `cargo doc` in `rustc_metadata/src/creader.rs`


XAMPPRocky:
EDIT(camelid): This issue is now being used to track a different, but related, bug. The discussion starting [here](https://github.com/rust-lang/rust/issues/84738#issuecomment-950370740) is current.

### Steps

- Clone [`rust-gpu`](https://github.com/EmbarkStudios/rust-gpu/tree/2f451ee9939c021b54214d2aeb986ecb7a6d7382).
- `cargo doc -p spirv-std --no-deps`

### Meta

`rustc --version --verbose`:
```
rustc 1.53.0-nightly (42816d61e 2021-04-24)
binary: rustc
commit-hash: 42816d61ead7e46d462df997958ccfd514f8c21c
commit-date: 2021-04-24
host: x86_64-apple-darwin
release: 1.53.0-nightly
LLVM version: 12.0.0
```

### Error output

```
thread 'rustc' panicked at 'index out of bounds: the len is 8 but the index is 8', compiler/rustc_metadata/src/creader.rs:137:21
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace

error: internal compiler error: unexpected panic

error: Unrecognized option: 'crate-version'

error: could not document `spirv-std`

Caused by:
  process didn't exit successfully: `rustdoc --edition=2018 --crate-type lib --crate-name spirv_std crates/spirv-std/src/lib.rs -o /Users/erin.power/src/rust-gpu2/target/doc --cfg 'feature="default"' --error-format=json --json=diagnostic-rendered-ansi -L dependency=/Users/erin.power/src/rust-gpu2/target/debug/deps --extern bitflags=/Users/erin.power/src/rust-gpu2/target/debug/deps/libbitflags-7c13a8e06c86cae3.rmeta --extern num_traits=/Users/erin.power/src/rust-gpu2/target/debug/deps/libnum_traits-226168b7b64579d8.rmeta --extern spirv_std_macros=/Users/erin.power/src/rust-gpu2/target/debug/deps/libspirv_std_macros-63f1f4de237cb2bc.dylib --extern spirv_types=/Users/erin.power/src/rust-gpu2/target/debug/deps/libspirv_types-7559b6c8fd9bbf59.rmeta --crate-version 0.4.0-alpha.7` (exit status: 1)
```

<details><summary><strong>Backtrace</strong></summary>
<p>

```
   0: _rust_begin_unwind
[...]

PR #83738, opened 2021-04-01
rustdoc: Don't load all extern crates unconditionally

jyn514:
Instead, only load the crates that are linked to with intra-doc links.

This doesn't help very much with any of rustdoc's fundamental issues
with freezing the resolver, but it at least fixes a stable-to-stable
regression, and makes the crate loading model somewhat more consistent
with rustc's. I tested and it unfortunately does not help at all with https://github.com/rust-lang/rust/pull/82496.

Closes https://github.com/rust-lang/rust/issues/68427. Opened https://github.com/rust-lang/rust/issues/83761 to track the more fundamental issues.
r? @petrochenkov cc @eddyb @ollie27

PR #88215, opened 2021-08-21
Reland #83738: "rustdoc: Don't load all extern crates unconditionally"

### LINKED ISSUE #68427: rustdoc only: duplicate lang item `panic_halt`
This is a regression from `rustdoc 1.40.0 (73528e339 2019-12-16)` to `rustdoc 1.41.0-beta.3 (fb3056216 2020-01-19)`.

If rustdoc attempts to document a crate that has at least two dependencies which implement a `panic_handler`, it fails with an error similar to

```
error: duplicate lang item in crate `panic_halt` (which `panic2` depends on): `panic_impl`.
  |
  = note: first defined in crate `std` (which `panic2` depends on).
```

Note that the crate *builds* fine, it's just documenting that is broken.

---

**Steps to reproduce**

1. Create a new project, e.g. `cargo new panic2`
2. Add a dependency on `panic-halt = "0.2.0"`, e.g. `cargo add panic-halt`
3. Attempt to document the crate, e.g. `cargo doc`

<details><summary>An example of the different behavior between `stable`/`beta` and `build`/`doc`.</summary>

```shell
$ cargo +stable clean

$ cargo +beta clean

$ cargo +stable build
   Compiling panic-halt v0.2.0
   Compiling panic2 v0.1.0 (/Users/logic/Projects/tmp/panic2)
    Finished dev [unoptimized + debuginfo] target(s) in 0.35s

$ cargo +stable doc
 Documenting panic-halt v0.2.0
    Checking panic-halt v0.2.0
 Documenting panic2 v0.1.0 (/Users/logic/Projects/tmp/panic2)
    Finished dev [unoptimized + debuginfo] target(s) in 1.40s

$ cargo +beta build
   Compiling panic-halt v0.2.0
   Compiling panic2 v0.1.0 (/Users/logic/Projects/tmp/panic2)
    Finished dev [unoptimized + debuginfo] target(s) in 0.27s

$ cargo +beta doc
 Documenting panic-halt v0.2.0
    Checking panic-halt v0.2.0
 Documenting panic2 v0.1.0 (/Users/logic/Projects/tmp/panic2)
error: duplicate lang item in crate `panic_halt` (which `panic2` depends on): `panic_impl`.
  |
  = note: first defined in crate `std` (which `panic2` depends on).

error: Could not document `panic2`.

Caused by:
  process didn't exit successfully: `rustdoc --edition=2018 --crate-type bin --crate-name panic2 src/main.rs -o /Users/logic/Projects/tmp/panic2/target/doc --error-format=json --json=diagnostic-rendered-ansi --document-private-items -L dependency=/Users/logic/Projects/tmp/panic2/target/debug/deps --extern panic_halt=/Users/logic/Projects/tmp/panic2/target/debug/deps/libpanic_halt-6d0ca720c3c355fb.rmeta` (exit code: 1)
```

</details>

jyn514:
I hopefully found all the bugs :crossed_fingers: time for a take two. See the last commit for details on what went wrong before.

r? @petrochenkov (but feel free to reassign to Guillaume if you don't have time.)

Closes https://github.com/rust-lang/rust/issues/68427. Includes a fix for https://github.com/rust-lang/rust/issues/84738.

### DIFF
--- a/src/librustdoc/core.rs
+++ b/src/librustdoc/core.rs
@@ -1,12 +1,12 @@
+use rustc_ast as ast;
 use rustc_data_structures::fx::{FxHashMap, FxHashSet};
 use rustc_data_structures::sync::{self, Lrc};
 use rustc_driver::abort_on_err;
 use rustc_errors::emitter::{Emitter, EmitterWriter};
 use rustc_errors::json::JsonEmitter;
 use rustc_feature::UnstableFeatures;
-use rustc_hir::def::Namespace::TypeNS;
 use rustc_hir::def::Res;
-use rustc_hir::def_id::{DefId, LocalDefId, CRATE_DEF_INDEX};
+use rustc_hir::def_id::{DefId, LocalDefId};
 use rustc_hir::HirId;
 use rustc_hir::{
     intravisit::{self, NestedVisitorMap, Visitor},
@@ -23,7 +23,7 @@ use rustc_session::DiagnosticOutput;
 use rustc_session::Session;
 use rustc_span::source_map;
 use rustc_span::symbol::sym;
-use rustc_span::{Span, DUMMY_SP};
+use rustc_span::Span;
 
 use std::cell::RefCell;
 use std::mem;
@@ -301,41 +301,16 @@ crate fn create_config(
 }
 
 crate fn create_resolver<'a>(
-    externs: config::Externs,
     queries: &Queries<'a>,
     sess: &Session,
 ) -> Rc<RefCell<interface::BoxedResolver>> {
-    let extern_names: Vec<String> = externs
-        .iter()
-        .filter(|(_, entry)| entry.add_prelude)
-        .map(|(name, _)| name)
-        .cloned()
-        .collect();
+    let (krate, resolver, _) = &*abort_on_err(queries.expansion(), sess).peek();
+    let resolver = resolver.clone();
 
-    let (_, resolver, _) = &*abort_on_err(queries.expansion(), sess).peek();
+    let mut loader = crate::passes::collect_intra_doc_links::IntraLinkCrateLoader::new(resolver);
+    ast::visit::walk_crate(&mut loader, krate);
 
-    // Before we actually clone it, let's force all the extern'd crates to
-    // actually be loaded, just in case they're only referred to inside
-    // intra-doc links
-    resolver.borrow_mut().access(|resolver| {
-        sess.time("load_extern_crates", || {
-            for extern_name in &extern_names {
-                debug!("loading extern crate {}", extern_name);
-                if let Err(()) = resolver
-                    .resolve_str_path_error(
-                        DUMMY_SP,
-                        extern_name,
-                        TypeNS,
-                        LocalDefId { local_def_index: CRATE_DEF_INDEX }.to_def_id(),
-                  ) {
-                    warn!("unable to resolve external crate {} (do you have an unused `--extern` crate?)", extern_name)
-                  }
-            }
-        });
-    });
-
-    // Now we're good to clone the resolver because everything should be loaded
-    resolver.clone()
+    loader.resolver
 }
 
 crate fn run_global_ctxt(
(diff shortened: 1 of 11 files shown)


src/librustdoc/passes/collect_intra_doc_links/early.rs
```diff
@@ -0,0 +1,77 @@
...
+    crate fn visit_crate(&mut self, krate: &ast::Crate) {
+        assert_eq!(self.current_mod, CRATE_DEF_ID);
+        self.visit_attributes(&krate.attrs, krate.span);
+    }
+
+    fn visit_attributes(&mut self, attrs: &[ast::Attribute], span: Span) {
+        use crate::html::markdown::markdown_links;
+        use crate::passes::collect_intra_doc_links::preprocess_link;
+
+        // FIXME: this probably needs to consider inlining
```
jyn514, line 35: Ugh, apparently github dropped my comment because I wrote it on a commit instead of the PR itself. Do you happen to have it in your email? If not I can try to rewrite it from memory, it was discussing exactly how this bug could happen and how I don't know how to test it with a UI test.

petrochenkov: No, I don't see any traces of the extended version of this comment in my mail.

src/test/rustdoc/issue-66159.rs
```diff
@@ -1,10 +0,0 @@
-// aux-crate:priv:issue_66159_1=issue-66159-1.rs
-// compile-flags:-Z unstable-options
-
-// The issue was an ICE which meant that we never actually generated the docs
-// so if we have generated the docs, we're okay.
-// Since we don't generate the docs for the auxiliary files, we can't actually
-// verify that the struct is linked correctly.
-
-// @has issue_66159/index.html
-//! [issue_66159_1::Something]
```
petrochenkov, line 10: Perhaps it makes sense to keep this test even if the rest of the PR is reverted?

jyn514: Good catch, thanks. I'll add it back.

src/librustdoc/passes/collect_intra_doc_links/early.rs
jyn514: Ok, it went something like this: "inlining" is refering to `doc(inline)`, not `#[inline]`. The test case for it would look something like:

```rust
// inner_crate
pub fn foo() {}

// middle_crate
/// Link to [inner_crate::foo]
pub fn bar() {}

// outer_crate
pub use middle_crate::bar;
```
In particular, the first time `inner_crate` will be loaded is when resolving the links on `bar`, which I don't think this pass will catch because it doesn't look at the attributes of the original definition, only of the re-export.

I'm not sure how to test this with a UI test because it requires three nested crates, and AFAIK `aux-build` only supports two.

jyn514: Seems to work fine when testing locally though :shrug: so this is probably ok.

src/librustdoc/core.rs
```diff
@@ -301,41 +301,17 @@ crate fn create_config(
...
-                  }
-            }
-        });
-    });
+    let (krate, resolver, _) = &*abort_on_err(queries.expansion(), sess).peek();
+    let resolver = resolver.clone();
+
+    let mut loader = crate::passes::collect_intra_doc_links::IntraLinkCrateLoader::new(resolver);
+    loader.visit_crate(krate);
+    ast::visit::walk_crate(&mut loader, krate);
```
petrochenkov, line 312: This code looks like it's visiting the crate twice.
It's probably better to rename `visit_crate`/`visit_attributes` to something that doesn't look like default `Visitor` methods.

Also, there's no need to pull the `loader` to outside of `src/librustdoc/passes/collect_intra_doc_links/early.rs`, the visiting logic can live in that module entirely.
Then making `visit_crate` a separate function won't be necessary as well.

petrochenkov:
r=me with https://github.com/rust-lang/rust/pull/88215#discussion_r693559128 addressed.

PR #88220, opened 2021-08-22
Implement `AsFd` etc. for `UnixListener`.

sunfishcode:
Implement `AsFd`, `From<OwnedFd>`, and `Into<OwnedFd>` for
`UnixListener`. This is a follow-up to #87329.

r? @joshtriplett

### DIFF
--- a/library/std/src/os/unix/net/listener.rs
+++ b/library/std/src/os/unix/net/listener.rs
@@ -274,7 +274,7 @@ impl AsFd for UnixListener {
 impl From<OwnedFd> for UnixListener {
     #[inline]
     fn from(fd: OwnedFd) -> UnixListener {
-        UnixListener(Socket::from_inner(FromInner::from_inner(OwnedFd::from(fd))))
+        UnixListener(Socket::from_inner(FromInner::from_inner(fd)))
     }
 }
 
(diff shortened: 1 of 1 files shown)


library/std/src/os/unix/net/listener.rs
```diff
@@ -262,6 +262,30 @@ impl IntoRawFd for UnixListener {
...
+    fn as_fd(&self) -> BorrowedFd<'_> {
+        self.0.as_inner().as_fd()
+    }
+}
+
+#[unstable(feature = "io_safety", issue = "87074")]
+impl From<OwnedFd> for UnixListener {
+    #[inline]
+    fn from(fd: OwnedFd) -> UnixListener {
+        UnixListener(Socket::from_inner(FromInner::from_inner(OwnedFd::from(fd))))
```
joshtriplett, line 277: Unless I'm missing something, you shouldn't need `OwnedFd::from` here; I think that's just an identity, since `fd` is already an `OwnedFd`.

library/std/src/os/unix/net/listener.rs
```diff
@@ -262,6 +262,30 @@ impl IntoRawFd for UnixListener {
...
+    }
+}
+
+#[unstable(feature = "io_safety", issue = "87074")]
+impl From<UnixListener> for OwnedFd {
+    #[inline]
+    fn from(listener: UnixListener) -> OwnedFd {
+        listener.0.into_inner().into_inner().into()
```
joshtriplett, line 285: Is the last `.into()` here redundant? Is the type already `OwnedFd` at that point?

PR #88229, opened 2021-08-22
Suggest importing the right kind of macro.

### LINKED ISSUE #88228: Missing derive macro suggests importing regular macro
Found while working on https://github.com/rust-lang/rust/issues/88206

```
error: cannot find derive macro `println` in this scope
  --> src/main.rs:16:10
   |
16 | #[derive(println)]
   |          ^^^^^^^
   |
   = note: consider importing this macro:
           std::println
```

m-ou-se:
Fixes #88228.

r? @estebank

### DIFF
--- a/compiler/rustc_resolve/src/diagnostics.rs
+++ b/compiler/rustc_resolve/src/diagnostics.rs
@@ -950,9 +950,7 @@ impl<'a> Resolver<'a> {
         self.add_typo_suggestion(err, suggestion, ident.span);
 
         let import_suggestions =
-            self.lookup_import_candidates(ident, Namespace::MacroNS, parent_scope, |res| {
-                matches!(res, Res::Def(DefKind::Macro(MacroKind::Bang), _))
-            });
+            self.lookup_import_candidates(ident, Namespace::MacroNS, parent_scope, is_expected);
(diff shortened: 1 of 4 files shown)


compiler/rustc_resolve/src/diagnostics.rs
```diff
@@ -950,9 +950,7 @@ impl<'a> Resolver<'a> {
         self.add_typo_suggestion(err, suggestion, ident.span);
 
         let import_suggestions =
-            self.lookup_import_candidates(ident, Namespace::MacroNS, parent_scope, |res| {
-                matches!(res, Res::Def(DefKind::Macro(MacroKind::Bang), _))
-            });
+            self.lookup_import_candidates(ident, Namespace::MacroNS, parent_scope, is_expected);
```
estebank, line 953: This is gonna have a merge conflict with the other PR, right?

estebank (approved):

compiler/rustc_resolve/src/diagnostics.rs
m-ou-se: That other PR includes this one as the first commit (with identical hash even), so should be fine. ^^

PR #88232, opened 2021-08-22
Add notes to macro-not-found diagnostics to point out how things with the same name were not a match.

### LINKED ISSUE #88206: Confusing diagnostics on #[derive(X)] when X is imported only as trait
Reported by @Hoverbear here: https://twitter.com/a_hoverbear/status/1428763621572501507

```rust
use serde::{Deserialize, Serialize};

#[derive(Serialize)]
struct A;
```

Output:
```
error: cannot find derive macro `Serialize` in this scope
 --> src/main.rs:3:10
  |
3 | #[derive(Serialize)]
  |          ^^^^^^^^^

warning: unused import: `serde::Serialize`
 --> src/main.rs:1:5
  |
1 | use serde::Serialize;
  |     ^^^^^^^^^^^^^^^^
  |
```

The diagnostics shows that `Serialize` is not found, and then complains that `Serialize` was imported and unused. Confusing.

(Only the trait `Serialize` is imported, not the `Serialize` derive macro, as the `derive` feature wasn't enabled on `serde`.)

m-ou-se:
This adds notes like:
```
error: cannot find derive macro `Serialize` in this scope
  --> $DIR/issue-88206.rs:22:10
   |
LL | #[derive(Serialize)]
   |          ^^^^^^^^^
   |
note: `Serialize` is imported here, but it is not a derive macro
  --> $DIR/issue-88206.rs:17:11
   |
LL | use hey::{Serialize, Deserialize};
   |           ^^^^^^^^^
```

Fixes https://github.com/rust-lang/rust/issues/88206

Includes https://github.com/rust-lang/rust/pull/88229

r? @estebank


compiler/rustc_resolve/src/diagnostics.rs
```diff
@@ -950,17 +950,48 @@ impl<'a> Resolver<'a> {
...
+                ident.span,
+            ) {
+                let it_is = match binding.macro_kind() {
+                    Some(MacroKind::Bang) => "it is a function-like macro".to_string(),
+                    Some(kind) => format!("it is {} {}", kind.article(), kind.descr_expected()),
+                    None => format!(
+                        "it is not {} {}",
+                        macro_kind.article(),
+                        macro_kind.descr_expected()
+                    ),
```
estebank, line 981: Instead of telling people what this _is not_, could we use `resolve_path` (or similar) to get what _it actually is_?

src/test/ui/issues/issue-11692-2.stderr
```diff
@@ -3,6 +3,8 @@ error: cannot find macro `test` in this scope
    |
 LL |     concat!(test!());
    |             ^^^^
+   |
+   = note: `test` is in scope, but it is an attribute
```
estebank, line 7: Seeing this case makes me wonder, would some users get confused with what an attribute is? Things like functions and structs should be fine, but I'm wondering if we should show "how to use them". I also wonder whether we should be showing their `def_span` (which we can easily do if we also look beyond the macros to get the item's def id).

src/test/ui/macros/issue-88206.stderr
```diff
@@ -0,0 +1,120 @@
...
+   |
+LL | #![warn(unused_imports)]
+   |         ^^^^^^^^^^^^^^
+
+warning: unused imports: `Deserialize`, `Serialize`
+  --> $DIR/issue-88206.rs:17:11
+   |
+LL | use hey::{Serialize, Deserialize};
+   |           ^^^^^^^^^  ^^^^^^^^^^^
+
```
estebank: Ideally we should collect a HashSet of macros not found that were referenced and are available as a different item, so that we filter those out of the unused import lint.

estebank (commented):
I love how these are looking 😃

I'm leaving some nitpicks, but we don't _need_ to address them all to land this.

src/test/ui/macros/issue-88206.stderr
estebank: You cheeky gal, addressing my comments before I publish my review :P

compiler/rustc_resolve/src/diagnostics.rs
m-ou-se: Yes! I wanted to do that, but couldn't find a way to get that information. I figured that at this stage that info isn't available yet, but if that information exists at this point, we should do that!

src/test/ui/issues/issue-11692-2.stderr
m-ou-se: We could also just not show the note in some cases if it wasn't explicitly imported. Mixing up derive and attribute macros should probably always show the note, but if a regular macro is involved, it might be good to only show the note if there's a `use std::test;` or something to point at.

src/test/ui/macros/issue-88206.stderr
m-ou-se: I now used `record_use` for this. Are there any advantages to using a separate HashSet?

compiler/rustc_resolve/src/diagnostics.rs
```diff
@@ -950,17 +950,51 @@ impl<'a> Resolver<'a> {
...
+                };
+                if let crate::NameBindingKind::Import { import, .. } = binding.kind {
+                    if !import.span.is_dummy() {
+                        err.span_note(
+                            import.span,
+                            &format!("`{}` is imported here, but {}", ident, it_is),
+                        );
+                        // Silence the 'unused import' warning we might get,
+                        // since this diagnostic already covers that import.
+                        self.record_use(ident, ns, binding, false);
```
m-ou-se, line 991: I was wondering what the namespace is used for here and whether to pass `ns` or `MacroNS`. Turns out that information isn't used at all: https://github.com/rust-lang/rust/pull/88238

compiler/rustc_resolve/src/diagnostics.rs
m-ou-se: Looks like that information is already available in `binding.res()`. Apparently I didn't look hard enough. ^^'

src/test/ui/issues/issue-11692-2.stderr
m-ou-se: It now shows:

```
   = note: `test` is in scope, but it is an attribute: `#[test]`
```
and
```
   = note: `MyTrait` is in scope, but it is a derive macro: `#[derive(MyTrait)]`
```

src/test/ui/tool-attributes/tool-attributes-misplaced-1.stderr
```diff
@@ -3,24 +3,32 @@ error: cannot find derive macro `rustfmt` in this scope
    |
 LL | #[derive(rustfmt)]
    |          ^^^^^^^
+   |
+   = note: `rustfmt` is in scope, but it is a tool module
```
m-ou-se, line 7: Should we just leave out the notes for obscure DefKinds like tool modules? Which other DefKinds should be skipped?

m-ou-se: Ignoring tool modules now. All other Res/DefKinds seem fine (or impossible to reach through an import).

src/test/ui/macros/issue-88206.stderr
```diff
@@ -32,15 +32,15 @@ error: cannot find macro `Box` in this scope
 LL |     Box!();
    |     ^^^
    |
-   = note: `Box` is in scope, but it is a struct
+   = note: `Box` is in scope, but it is a struct, not a macro
```
estebank, line 35: For the cases where we don't have a span, wouldn't it make sense to turn this into the primary span? Something like:

```
error: cannot find macro `Copy` in this scope
  --> $DIR/issue-88206.rs:56:5
   |
LL |     Copy!();
   |     ^^^^ this is not a macro, it is a derive macro: `#[derive(Copy)]`
```

m-ou-se: I'd avoid pointing the note at the identifier the user wrote, because the error already says `` cannot find `Thing` ``, and the note is about *another* `Thing`, not the `Thing` they wrote. (Even though they have the same name.)

So in the case we can point at an import, we do that to highlight `Thing` here is not found, but `Thing` there exists but is something else. Pointing both at the same identifier would make it less clear that we're talking about two different `Thing`s.

PR #88238, opened 2021-08-22
Stop tracking namespace in used_imports.

m-ou-se:
This changes `used_imports` from a `FxHashSet<(NodeId, Namespace)>` to a `FxHashSet<NodeId>`, as the Namespace information isn't used.

The only point that uses it did three lookups, `|=`'ing them together.

r? @estebank


estebank:
Probably left-overs from previous refactors.

@bors r+ rollup

PR #82124, opened 2021-02-15
Pass large interpreter types by reference, not value

tmiasko:
r? @ghost

PR #88242, opened 2021-08-22
Use custom wrap-around type instead of RangeInclusive

bonega:
Two reasons:

1. More memory is allocated than necessary for `valid_range` in `Scalar`. The range is not used as an iterator and `exhausted` is never used.
2. `contains`, `count` etc. methods in `RangeInclusive` are doing very unhelpful(and dangerous!) things when used as a wrap-around range. - In general this PR wants to limit potentially confusing methods, that have a low probability of working.

Doing a local perf run, every metric shows improvement except for instructions.
Max-rss seem to have a very consistent improvement.

Sorry - newbie here, probably doing something wrong.


compiler/rustc_target/src/abi/mod.rs
```diff
@@ -988,24 +1008,17 @@ impl Niche {
...
 
         // Compute the range of invalid values being reserved.
-        let start = v.end().wrapping_add(1) & max_value;
-        let end = v.end().wrapping_add(count) & max_value;
+        let start = v.end.wrapping_add(1) & max_value;
+        let end = v.end.wrapping_add(count) & max_value;
 
         // If the `end` of our range is inside the valid range,
         // then we ran out of invalid values.
         // FIXME(eddyb) abstract this with a wraparound range type.
```
oli-obk, line 1016: this fixme is resolved with this PR

src/test/ui/consts/const-eval/ub-nonnull.32bit.stderr
```diff
@@ -52,7 +52,7 @@ error[E0080]: it is undefined behavior to use this value
   --> $DIR/ub-nonnull.rs:41:1
    |
 LL | const BAD_RANGE1: RestrictedRange1 = unsafe { RestrictedRange1(42) };
-   | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ type validation failed: encountered 42, but expected something in the range 10..=30
+   | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ type validation failed: encountered 42, but expected something in the range AllocationRange { start: 10, end: 30 }
```
oli-obk, line 55: should probably manually implement Debug so that we get the previous output back

bonega:
Thanks for your quick review!
Unfortunately I must be doing something wrong with my local perf run, which showed clear improvements.
Will analyze it and get back.

compiler/rustc_mir/src/interpret/validity.rs
```diff
@@ -634,8 +623,8 @@ impl<'rt, 'mir, 'tcx: 'mir, M: Machine<'mir, 'tcx>> ValidityVisitor<'rt, 'mir, '
         scalar_layout: &ScalarAbi,
     ) -> InterpResult<'tcx> {
         let value = self.read_scalar(op)?;
-        let valid_range = &scalar_layout.valid_range;
-        let (lo, hi) = valid_range.clone().into_inner();
+        let valid_range = scalar_layout.valid_range;
+        let AllocationRange { start: lo, end: hi } = valid_range;
         // Determine the allowed range
         // `max_hi` is as big as the size fits
         let max_hi = u128::MAX >> (128 - op.layout.size.bits());
```
RalfJung, line 630: The check being done a few lines down here (testing whether the range is equivalent to the full range containing all values) also sounds like something that could be provided as a method on the new range type.

RalfJung:
Big :+1: on making this a dedicated type!

However, the name `AllocationRange` does not make sense to me. This range has nothing to do with [our `Allocation` type](https://doc.rust-lang.org/nightly/nightly-rustc/rustc_middle/mir/interpret/struct.Allocation.html), nor with any user use of the word "allocation" that I can think of. What about `WrappingRange`?

compiler/rustc_target/src/abi/mod.rs
```diff
@@ -677,32 +677,52 @@ impl Primitive {
...
+/// start > end, it represents `start..=MAX`,
+/// followed by `0..=end`.
+///
+/// That is, for an i8 primitive, a range of `254..=2` means following
+/// sequence:
+///
+///    254 (-2), 255 (-1), 0, 1, 2
+///
+/// This is intended specifically to mirror LLVM’s `!range` metadata,
+/// semantics.
```
RalfJung, line 690: What does the ", semantics" here do? I can't quite parse this. Did you mean "to mirror the semantics of LLVM’s `!range` metadata"?

bonega:
> Big  on making this a dedicated type! Even without perf improvements, this is a good change.
> 
> However, the name `AllocationRange` does not make sense to me. This range has nothing to do with [our `Allocation` type](https://doc.rust-lang.org/nightly/nightly-rustc/rustc_middle/mir/interpret/struct.Allocation.html), nor with any user use of the word "allocation" that I can think of. What about `WrappingRange`?

Originally I thought of `AllocationRange` to be part of reserve functionality for niches, which would have meant specific considerations.
I will change it to `WrappingRange` which is much better.

compiler/rustc_target/src/abi/mod.rs
bonega: That whole comment was copied from: https://github.com/rust-lang/rust/pull/88242/files#diff-087cde07a8dbab7f1fb846c8b9367f3fa41381cac27fe06461b68c2e9ed2b623L686-L696
That is the comment for the `valid_range` field in `Scalar`
Idk if it should be kept or not

compiler/rustc_middle/src/ty/layout.rs
```diff
@@ -526,7 +529,7 @@ impl<'tcx> LayoutCx<'tcx, TyCtxt<'tcx>> {
             }),
             ty::FnPtr(_) => {
                 let mut ptr = scalar_unit(Pointer);
-                ptr.valid_range = 1..=*ptr.valid_range.end();
+                ptr.valid_range = AllocationRange { start: 1, end: ptr.valid_range.end };
```
oli-obk, line 532: This pattern probably deserves a `with_start` method that takes a range and replaces the start.

compiler/rustc_middle/src/ty/layout.rs
```diff
@@ -933,14 +938,18 @@ impl<'tcx> LayoutCx<'tcx, TyCtxt<'tcx>> {
                             if let Bound::Included(start) = start {
                                 // FIXME(eddyb) this might be incorrect - it doesn't
                                 // account for wrap-around (end < start) ranges.
-                                assert!(*scalar.valid_range.start() <= start);
-                                scalar.valid_range = start..=*scalar.valid_range.end();
+                                assert!(scalar.valid_range.start <= start);
+                                // scalar.valid_range =
+                                //     AllocationRange { start, end: scalar.valid_range.end };
```
oli-obk, line 943: The change you made seems good to me, so you can remove the commented out exact translation of the previous code

compiler/rustc_middle/src/ty/layout.rs
```diff
@@ -2846,8 +2861,8 @@ where
                 return;
             }
 
-            if scalar.valid_range.start() < scalar.valid_range.end() {
-                if *scalar.valid_range.start() > 0 {
+            if scalar.valid_range.start < scalar.valid_range.end {
+                if scalar.valid_range.start > 0 {
```
oli-obk, line 2865: This could be `!scalar.valid_range.contains(0)`

compiler/rustc_codegen_llvm/src/builder.rs
```diff
@@ -462,15 +462,15 @@ impl BuilderMethods<'a, 'tcx> for Builder<'a, 'll, 'tcx> {
...
+            let vr = scalar.valid_range;
             match scalar.value {
                 abi::Int(..) => {
                     let range = scalar.valid_range_exclusive(bx);
                     if range.start != range.end {
                         bx.range_metadata(load, range);
                     }
                 }
-                abi::Pointer if vr.start() < vr.end() && !vr.contains(&0) => {
+                abi::Pointer if vr.start < vr.end && !vr.contains(0) => {
```
oli-obk, line 473: is the `start < end` check necessary? The contains method does the same check essentially.

oli-obk:
> Even without perf improvements, this is a good change.

agreed.

compiler/rustc_codegen_llvm/src/builder.rs
```diff
@@ -462,15 +462,14 @@ impl BuilderMethods<'a, 'tcx> for Builder<'a, 'll, 'tcx> {
...
         ) {
-            let vr = scalar.valid_range;
             match scalar.value {
                 abi::Int(..) => {
                     let range = scalar.valid_range_exclusive(bx);
                     if range.start != range.end {
                         bx.range_metadata(load, range);
                     }
                 }
-                abi::Pointer if vr.start < vr.end && !vr.contains(0) => {
```
bonega: I have implemented this as a generic `contains_zero` method.
Where I check for `vr.start <= vr.end`, is there any reason that `start` and `end` couldn't be equal when checking for what I think is NonNull?
`WrappingRange{start:1, end: 1}.contains_zero()` should be true is my assumption.

compiler/rustc_middle/src/ty/layout.rs
bonega: Will implement it and `with_end`

compiler/rustc_mir/src/interpret/validity.rs
bonega: I have thought about it, only problem I see is that we have to compute `max_hi` which is dependent on not just the range, but the allowed number of bits.

compiler/rustc_codegen_llvm/src/builder.rs
RalfJung: > I have implemented this as a generic contains_zero method.

What is the difference to `.contains(0)`?

compiler/rustc_target/src/abi/mod.rs
```diff
@@ -677,32 +677,59 @@ impl Primitive {
...
+        } else {
+            self.start <= v || v <= self.end
+        }
+    }
+
+    /// Returns `true` if zero is contained in the range.
+    /// Equal to `range.contains(0)` but should be faster.
+    #[inline]
+    pub fn contains_zero(&self) -> bool {
+        !(self.start <= self.end && self.start != 0)
```
RalfJung, line 713: IMO `self.start > self.end || self.start == 0` would be a lot more readable.

bonega: Yeah, wrote it originally as is_non_null, changing it

bonega: Is `start < end` right?, shouldn't it be `start <= end`

compiler/rustc_mir/src/interpret/validity.rs
RalfJung: Yes, that helper function will have to take the `Size` of the scalar.

compiler/rustc_target/src/abi/mod.rs
RalfJung: `start > end` means it wraps around which means it does contain 0.

Might be good to have a comment explaining that. :)

bonega: Just checking if you think that the code before my PR was incorrect?

oli-obk:
@bors try @rust-timer queue

just checking again, layout computation changes can quickly cause perf regressions due to changed inlining or so

compiler/rustc_target/src/abi/mod.rs
RalfJung: The code I proposed is equivalent to your `contains_zero`.

I assume you meant to add a negation? So that would be `!(self.start < self.end && self.start != 0)` for the old code which is equivalent to `self.start >= self.end || self.start == 0`. Yeah that seems incorrect to me; the range `5..=5` does not contain 0.

bonega: great thanks!

bonega:
think bors broke...

bonega:
Here is the comparison against the previous perf run: https://perf.rust-lang.org/compare.html?start=b36ec6286bd82a4ef6fbe3af0c69fc7e9a8b316b&end=fc1c134db6044555b4ac23f5896952feaafc7f11&stat=instructions%3Au

bonega:
Request another perf run for last commit, sometimes the compiler is too restrictive with inlining.

bonega:
Thanks, unfortunately I think the picture is still a bit unclear.
Perf-run 2: `fc1c134` had no parent `5998c2e`, therefore it is somewhat problematic to judge the performance.(I posted a link, but I don't think we can use it to draw conclusions)

perf-run 3 > 1
perf-run 2 ? 3

oli-obk:
The perf page is just broken for me since two days... It looks like an unsubstituted template page to me. Not sure what's going on

oli-obk:
perf is clean, diff lgtm. Thanks for doing this!

@bors r+

bonega:
Thanks to both of you for the excellent newbie support!

compiler/rustc_target/src/abi/mod.rs
```diff
@@ -971,14 +1019,14 @@ impl Niche {
...
-        let niche = v.end().wrapping_add(1)..*v.start();
+        let niche = v.end.wrapping_add(1)..v.start;
         niche.end.wrapping_sub(niche.start) & max_value
     }
 
     pub fn reserve<C: HasDataLayout>(&self, cx: &C, count: u128) -> Option<(u128, Scalar)> {
         assert!(count > 0);
 
-        let Scalar { value, valid_range: ref v } = self.scalar;
+        let Scalar { value, valid_range: v } = self.scalar.clone();
```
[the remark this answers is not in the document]
bonega: I was a bit unsure if it was bad practice, don't really know where breakpoint for "cheap" is considered.
Doing a commit in other PR for it.

compiler/rustc_target/src/abi/mod.rs
```diff
@@ -677,32 +677,80 @@ impl Primitive {
...
+            self.start <= v || v <= self.end
+        }
+    }
+
+    /// Returns `true` if zero is contained in the range.
+    /// Equal to `range.contains(0)` but should be faster.
+    #[inline(always)]
+    pub fn contains_zero(&self) -> bool {
+        self.start > self.end || self.start == 0
+    }
```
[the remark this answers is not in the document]
bonega: I should respect the compiler, thank you!

compiler/rustc_target/src/abi/mod.rs
nagisa: ya drop the comma and it should become comprehensible (and accurate)

bonega: fixed in #88327

compiler/rustc_target/src/abi/mod.rs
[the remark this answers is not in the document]
eddyb: Pretty much anything other than "dozens of fields/elements" `struct`/arrays, that can be `Copy`, should be `Copy` in the compiler, IMO. It's both cleaner, and gives LLVM an easier time to not have to deal with references everywhere.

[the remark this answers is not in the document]
RalfJung: Well, a bunch of CTFE engine types were changed to be passed by reference in most places... https://github.com/rust-lang/rust/pull/82124 :shrug: (they are still copy though(

Issue #83116, opened 2021-03-14
"C-unwind" ABI is unsound with "-Cpanic=abort"


RalfJung:
In the following code, `test` should explicitly be allowed to unwind:
```rust
#![feature(c_unwind)]

extern "C-unwind" {
    fn test();
}

fn main() {
    unsafe { test(); }
}
```
However, when building this with `-Cpanic=abort`, we generated LLVM IR as follows:
```
; Function Attrs: nounwind nonlazybind
declare void @test() unnamed_addr #1
```
This means unwinding of `test` is UB, i.e., this is a soundness problem.

apiraino:
Assigning `P-medium` since still a nightly-only feature. Discussed as part of the [Prioritization Working Group procedure](https://rust-lang.zulipchat.com/#narrow/stream/245100-t-compiler.2Fwg-prioritization.2Falerts/topic/.2383116.20.22C-unwind.22.20ABI.20is.20unsound.20with.20.22-Cpanic.3Dabort.22) and removing `I-prioritize`.

@rustbot label -I-prioritize +P-medium

cratelyn:
I'll look into a fix for this, as the author of #76570, which introduced this issue. Thank you for flagging this @RalfJung.

nagisa:
This is probably pretty hard to fix in a way that affects least amount of code negatively.

Any function that may at any point, directly or indirectly, end up calling a `C-unwind` function must _not_ be annotated with a `nounwind`. It can't be known in general case that this is true, this knowledge could only be available at binary load time.

So AFAICT the only viable fixes here are: 

* just don't annotate anything with `nounwind`, ever, `-Cpanic=abort` or not (we can still replace `invoke`s with `call`s though);
* annotate leaf functions (functions that don't call anything else) as `nounwind` and then any functions that only call other `nounwind` functions as `nounwind`.

The latter is something LLVM is already capable of inferring by itself, so there's probably little value in doing it in rustc. Maybe later, once or if MIR optimisations start caring about this, I guess?

RalfJung:
There's another possible fix: wrap *import* `extern "C-unwind"` functions in a wrapper that turns unwinding into aborts. Rust code calling the function all goes through that wrapper. Then the `nounwind` is actually correct.

nagisa:
[...]

PR #88243, opened 2021-08-22
Enable new pass manager with LLVM 13

nikic:
The new pass manager is enabled by default in clang since Clang/LLVM 13. Per the recent discussion on llvm-dev (https://lists.llvm.org/pipermail/llvm-dev/2021-August/152305.html) the legacy pass manager will be unmaintained in LLVM 14 and removed entirely in LLVM 15.

This switches us to use the new pass manager if LLVM >= 13 is used. It's possible to still use the old pass manager using `-Z new-llvm-pass-manager=no`.

### DIFF
--- a/compiler/rustc_codegen_llvm/src/back/write.rs
+++ b/compiler/rustc_codegen_llvm/src/back/write.rs
@@ -370,8 +370,9 @@ fn get_pgo_use_path(config: &ModuleConfig) -> Option<CString> {
 }
 
 pub(crate) fn should_use_new_llvm_pass_manager(config: &ModuleConfig) -> bool {
-    // The new pass manager is disabled by default.
-    config.new_llvm_pass_manager.unwrap_or(false)
+    // The new pass manager is enabled by default for LLVM >= 13.
+    // This matches Clang, which also enables it since Clang 13.
+    config.new_llvm_pass_manager.unwrap_or_else(|| llvm_util::get_version() >= (13, 0, 0))
 }
 
 pub(crate) unsafe fn optimize_with_new_llvm_pass_manager(
--- a/compiler/rustc_llvm/llvm-wrapper/PassWrapper.cpp
+++ b/compiler/rustc_llvm/llvm-wrapper/PassWrapper.cpp
@@ -1004,7 +1004,10 @@ LLVMRustOptimizeWithNewPassManager(
 #endif
   bool NeedThinLTOBufferPasses = UseThinLTOBuffers;
   if (!NoPrepopulatePasses) {
-    if (OptLevel == OptimizationLevel::O0) {
+    // The pre-link pipelines don't support O0 and require using budilO0DefaultPipeline() instead.
+    // At the same time, the LTO pipelines do support O0 and using them is required.
+    bool IsLTO = OptStage == LLVMRustOptStage::ThinLTO || OptStage == LLVMRustOptStage::FatLTO;
+    if (OptLevel == OptimizationLevel::O0 && !IsLTO) {
 #if LLVM_VERSION_GE(12, 0)
       for (const auto &C : PipelineStartEPCallbacks)
         PB.registerPipelineStartEPCallback(C);
--- a/src/test/codegen/panic-in-drop-abort.rs
+++ b/src/test/codegen/panic-in-drop-abort.rs
@@ -1,7 +1,14 @@
-// compile-flags: -Z panic-in-drop=abort -O
+// compile-flags: -Z panic-in-drop=abort -O -Z new-llvm-pass-manager=no
 
 // Ensure that unwinding code paths are eliminated from the output after
 // optimization.
+
+// This test uses -Z new-llvm-pass-manager=no, because the expected optimization does not happen
+// on targets using SEH exceptions (i.e. MSVC) anymore. The core issue is that Rust promises that
+// the drop_in_place() function can't unwind, but implements it in a way that *can*, because we
+// currently go out of our way to allow longjmps, which also use the unwinding mechanism on MSVC
+// targets. We should either forbid longjmps, or not assume nounwind, making this optimization
+// incompatible with the current behavior of running cleanuppads on longjmp unwinding.
 
 // CHECK-NOT: {{(call|invoke).*}}should_not_appear_in_output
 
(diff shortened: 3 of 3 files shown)


klensy:
But for max-rss it's not so green, sadly https://perf.rust-lang.org/compare.html?start=91f9806208834de3fb5f62712356b0d84ec388fd&end=273c8e11ecc63776c5fe5d98ad6a74ef3a37a7f3&stat=max-rss

nikic:
Fix for `wasm-panic-small` regression: https://reviews.llvm.org/D108866

nikic:
@klensy Yeah, that's a known tradeoff in the new pass manager design. It keeps analysis results around unless invalidated. That means less recalculations, but also more memory usage.

cuviper:
Are there ways to mitigate that memory usage if needed? We already see rustc pushing address limits on 32-bit hosts...

nikic:
r? @cuviper or @nagisa 

> Are there ways to mitigate that memory usage if needed? We already see rustc pushing address limits on 32-bit hosts...

I'm at least not aware of any options that can changed to reduce memory usage :/

r00ster91:
Maybe a flag could be added that enables the old PM to trade longer compile times for less memory? `--old-pm`. If it's added, it should explicitly be a temporary flag though that gets removed in coming versions because at some point the old PM will probably be removed from newer LLVM versions. This could temporarily help with the memory usage for people that really need it until more memory optimizations arrive in rustc.

nikic:
@r00ster91 This flag already exists in the form of `-Z new-llvm-pass-manager=no` :)

nikic:
The submodule update in #88603 also includes the change relevant for this PR, so this is now down to flipping the default only.

nikic:
@cuviper @nagisa Anything in particular we're waiting on here? The beta bump happened a few days ago, so we're at the start of the release cycle now.

nikic:
We were using the wrong pipeline for LTO at O0. Let's see if that fixes it...

@bors try

nikic:
Looks like that was indeed it.

@bors r=nagisa

nikic:
Okay, looks like there's some kind of genuine optimization problem here. The input IR for MSCV targets is: https://pastebin.com/BpBkfUvi

nikic:
I believe the problem here is that the legacy PM used to run a PruneEH pass prior to inlining, that would propagate nounwind and simplify invokes to calls prior to inlining. In the new PM, that pass is no longer present, on the premise that inlining doesn't really care whether something is an invoke or call, and the functionality of PruneEH is covered by SimplifyCFG (invoke -> call) and PostOrderFunctionAttrs (nounwind inference). The thing in this case is that we're calling a non-nounwind function (drop) that calls a nounwind function (drop_in_place). If we inline drop_in_place first, we lose the information that it is nounwind. On non-MSVC targets we can still infer that the implementation is actually nounwind, but for MSVC we have a `catchswitch ... unwind to caller` that may unwind (I'm not familiar with that exception model, so no idea what that really means / if it can be avoided).

I'm not sure it makes sense to try and address this on the LLVM side, possibly the improvement suggestions that have been made on #88759 would cover this by generating better IR in the first place.

cuviper:
Re-assigning...
r? @nagisa

Also, replying to myself in https://github.com/rust-lang/rust/pull/88243#issuecomment-908805302 :
> Are there ways to mitigate that memory usage if needed? We already see rustc pushing address limits on 32-bit hosts...

One particularly bad regression I found with an `i686-unknown-linux-gnu` build host is `rustc_query_impl`. With old PM, I got max rss 1298624 kb, and new PM went up to 1686772 kb. I don't know what the virt usage was, but this is getting dangerous.

I did find a possible workaround though -- cranking codegen-units up to 256 (like we do for incremental) reduced it to max rss 1340744 kb. That should be fine.

(All this is without compiler debuginfo, because that has OOMed 32-bit hosts for a long time.)

nikic:
Returning to the unwinding issue with MSVC: Is our claim of `nounwind` even correct there? Looking at https://github.com/rust-lang/rust/blob/197fc8591e6b02ac1f359ee0a72616eb4a6d4f4c/compiler/rustc_codegen_ssa/src/mir/block.rs#L1259 we seem to specifically *not* use the obvious, actually nounwind implementation, because we do want to permit longjmps, which happen to be implementing using unwinding on MSVC targets. We can either not support longjump, properly abort and mark the function nounwind, or we support longjump but then shouldn't use nounwind with MSVC.

cc @Amanieu

nagisa:
Aha, right. I suspect on MSVC we should just disable this section of code:

https://github.com/rust-lang/rust/blob/60fe8b3a65be709fe2163b8ab438ef14209055cc/compiler/rustc_typeck/src/collect.rs#L2687-L2691

and let LLVM infer `nounwind` attribute by itself, if it does that anymore? See also somewhat relevant #83116

nikic:
I feel like that kindof defeats the purpose of any "nounwind" functionality in Rust though. It's not like this only affects panic-in-drop=abort in particular, it also affects extern function unwinding and panic=abort. All of those currently mark functions as nounwind, which is incorrect with SEH exceptions because we still permit longjmp unwinds (though as you point out, marking functions as nounwind with panic=abort is already incorrect for other reasons). I'm not really sure how we ended up here -- I would have expected the use of longjmp on MSVC to require the `"C-unwind"` ABI just like all other unwinds.

nagisa:
I think that's the plan ultimately but it's not currently past the point
where we made the jump of disallowing unwinding fro extern "not-unwind"

Amanieu:
It is my understanding that SEH longjmp does not count as an "unwind" for the purposes of LLVM's `nounwind`. Clang always emits `nounwind` for C code even on MSVC targets and yet `longjmp` still works correctly. Since it is UB for `longjmp` to skip over the execution of any destructors (source: C++ standard), LLVM can just assume that `longjmp` won't execute any landing pads and optimize them away (which is main purpose of `nounwind`).

nikic:
Oooh, I wasn't aware of that! LangRef for `nounwind` does indeed say:

> This function attribute indicates that the function never raises an exception. If the function does raise an exception, its runtime behavior is undefined. However, functions marked nounwind may still trap or generate asynchronous exceptions. Exception handling schemes that are recognized by LLVM to handle asynchronous exceptions, such as SEH, will still provide their implementation defined semantics.

So `nounwind` is really `noexpect`, and this use of `nounwind` is fine.

nikic:
I submitted https://bugs.llvm.org/show_bug.cgi?id=51969 for the optimization problem under NewPM.

rylev:
This is such a large performance gain that the few, small regressions are clearly worth the price. As noted this does regression rss, but that's an explicit tradeoff made by enabling the pass manager. 

PR #88257, opened 2021-08-23
Provide more context on incorrect inner attribute

estebank:
Suggest changing an inner attribute into an outer attribute if followed by an item.

### DIFF
--- a/compiler/rustc_parse/src/parser/attr.rs
+++ b/compiler/rustc_parse/src/parser/attr.rs
@@ -1,10 +1,10 @@
-use super::{AttrWrapper, Capturing, Parser, PathStyle};
+use super::{AttrWrapper, Capturing, ForceCollect, Parser, PathStyle};
 use rustc_ast as ast;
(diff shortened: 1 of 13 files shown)


compiler/rustc_parse/src/parser/attr.rs
```diff
@@ -125,6 +134,63 @@ impl<'a> Parser<'a> {
...
+                    &format!("the inner {} doesn't annotate this {}", attr_name, item.kind.descr()),
+                );
+                err.span_suggestion_verbose(
+                    replacement_span,
+                    &format!(
+                        "to annotate the {}, change the {} from inner to outer style",
+                        item.kind.descr(),
+                        attr_name
+                    ),
+                    (if is_doc { "/" } else { "" }).to_string(),
```
estebank, line 181: This is not correct for `/*!` docstrings.

jackh726:
Not really familiar with the parsing code, so going to reroll r? rust-lang/compiler-team

PR #88262, opened 2021-08-23
Cow'ify some pprust methods

klensy:
Reduce number of potential needless de/allocations by using `Cow<'static, str>` instead of explicit `String` type.

### DIFF
--- a/compiler/rustc_ast_pretty/src/pprust/mod.rs
+++ b/compiler/rustc_ast_pretty/src/pprust/mod.rs
@@ -8,17 +8,19 @@ use rustc_ast as ast;
 use rustc_ast::token::{Nonterminal, Token, TokenKind};
 use rustc_ast::tokenstream::{TokenStream, TokenTree};
 
+use std::borrow::Cow;
+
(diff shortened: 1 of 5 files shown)


nagisa:
Could you please adjust the PR description to include some motivation for this change (i.e. the why?)

r00ster91:
> Could you please adjust the PR description to include some motivation for this change (i.e. the why?)

I'm confident that the goal of this PR is performance. It might be worth running a benchmark on this.

klensy:
@r00ster91 I've edited PR description, previously it was empty.

nagisa:
Right, the goal of my request was to make sure there's less guessing involved as to what the motivation really is.

---

I don't think pretty printer is at all performance critical and therefore value of a benchmark here seems fairly low (and also pretty high-effort). The changes also don't seem all that intrusive either, so I'm happy to approve this based on the intuition that things are better this way.

Can you please also squash the commits together? r=me afterwards.

klensy:
Squashed.

PR #88267, opened 2021-08-23
2229: Update signature for truncate function

arora-aman:
r? @nikomatsakis

### DIFF
--- a/compiler/rustc_typeck/src/check/upvar.rs
+++ b/compiler/rustc_typeck/src/check/upvar.rs
@@ -489,7 +489,7 @@ impl<'a, 'tcx> FnCtxt<'a, 'tcx> {
         let mut root_var_min_capture_list =
             typeck_results.closure_min_captures.remove(&closure_def_id).unwrap_or_default();
 
-        for (place, capture_info) in capture_information.into_iter() {
+        for (mut place, capture_info) in capture_information.into_iter() {
             let var_hir_id = match place.base {
                 PlaceBase::Upvar(upvar_id) => upvar_id.var_path.hir_id,
                 base => bug!("Expected upvar, found={:?}", base),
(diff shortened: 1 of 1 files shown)


compiler/rustc_typeck/src/check/upvar.rs
```diff
@@ -1841,31 +1837,28 @@ impl<'a, 'tcx> euv::Delegate<'tcx> for InferBorrowKind<'a, 'tcx> {
...
+    mut curr_mode: ty::UpvarCapture<'tcx>,
 ) -> (Place<'tcx>, ty::UpvarCapture<'tcx>) {
-    if place.projections.is_empty() {
-        // Nothing to do here
-        return (place, curr_mode);
-    }
-
     if place.base_ty.is_unsafe_ptr() {
-        return truncate_place_to_len(place, curr_mode, 0);
+        truncate_place_to_len_and_update_capture_kind(&mut place, &mut curr_mode, 0);
```
nikomatsakis, line 1844: we used to return here, but now we don't? maybe this function also wants to take `&mut` arguments?

compiler/rustc_typeck/src/check/upvar.rs
```diff
@@ -1880,7 +1873,7 @@ fn restrict_capture_precision<'tcx>(
     place: Place<'tcx>,
     curr_mode: ty::UpvarCapture<'tcx>,
 ) -> (Place<'tcx>, ty::UpvarCapture<'tcx>) {
-    let (place, curr_mode) = restrict_precision_for_unsafe(place, curr_mode);
+    let (mut place, mut curr_mode) = restrict_precision_for_unsafe(place, curr_mode);
```
nikomatsakis, line 1876: as above, I suspect this could take `&mut` place arguments

nikomatsakis (changes requested):
I think we should push the `&mut` arguments through more thoroughly

compiler/rustc_typeck/src/check/upvar.rs
arora-aman: we just return once at the bottom. Otherwise each of these has a `return (place, curr_mode)`

nikomatsakis:
What do you think about having the general pattern be to take `&mut` and not to return a tuple, @arora-aman ?

arora-aman:
> What do you think about having the general pattern be to take `&mut` and not to return a tuple, @arora-aman ?

I personally feel returning is more functional  and gives the reader of the code an exact idea of what's going on / being upadted (though so does the name). 

Also returning the tuple makes it closer to how the algorithm is written out in the reference. But If you prefer `&mut` everywhere I don't mind making the change :)

nikomatsakis:
@arora-aman I see your point. Let's keep it as is.

@bors  r+

PR #88268, opened 2021-08-23
rustdoc: Fix generics generation in search index

GuillaumeGomez:
The generics were not added to the search index as they should, instead they were added as arguments. I used this opportunity to allow generics to have generics themselves (will come in very handy for my current rewrite of the search engine!).

r? @jyn514

### DIFF
--- a/src/librustdoc/formats/item_type.rs
+++ b/src/librustdoc/formats/item_type.rs
@@ -48,6 +48,7 @@ crate enum ItemType {
     ProcAttribute = 23,
     ProcDerive = 24,
     TraitAlias = 25,
+    Generic = 26,
 }
 
 impl Serialize for ItemType {
@@ -173,6 +174,7 @@ impl ItemType {
             ItemType::ProcAttribute => "attr",
             ItemType::ProcDerive => "derive",
             ItemType::TraitAlias => "traitalias",
+            ItemType::Generic => "generic",
         }
     }
 }
--- a/src/librustdoc/html/render/cache.rs
+++ b/src/librustdoc/html/render/cache.rs
@@ -1,7 +1,7 @@
 use std::collections::hash_map::Entry;
 use std::collections::BTreeMap;
 
-use rustc_data_structures::fx::{FxHashMap, FxHashSet};
+use rustc_data_structures::fx::FxHashMap;
 use rustc_middle::ty::TyCtxt;
 use rustc_span::symbol::Symbol;
 use serde::ser::{Serialize, SerializeStruct, Serializer};
@@ -192,32 +192,24 @@ crate fn get_index_search_type<'tcx>(
     item: &clean::Item,
     tcx: TyCtxt<'tcx>,
 ) -> Option<IndexItemFunctionType> {
-    let (all_types, ret_types) = match *item.kind {
+    let (mut inputs, mut output) = match *item.kind {
         clean::FunctionItem(ref f) => get_all_types(&f.generics, &f.decl, tcx),
         clean::MethodItem(ref m, _) => get_all_types(&m.generics, &m.decl, tcx),
         clean::TyMethodItem(ref m) => get_all_types(&m.generics, &m.decl, tcx),
         _ => return None,
     };
 
-    let inputs = all_types
-        .iter()
-        .map(|(ty, kind)| TypeWithKind::from((get_index_type(ty), *kind)))
-        .filter(|a| a.ty.name.is_some())
-        .collect();
-    let output = ret_types
-        .iter()
-        .map(|(ty, kind)| TypeWithKind::from((get_index_type(ty), *kind)))
-        .filter(|a| a.ty.name.is_some())
-        .collect::<Vec<_>>();
+    inputs.retain(|a| a.ty.name.is_some());
+    output.retain(|a| a.ty.name.is_some());
     let output = if output.is_empty() { None } else { Some(output) };
 
     Some(IndexItemFunctionType { inputs, output })
 }
 
-fn get_index_type(clean_type: &clean::Type) -> RenderType {
+fn get_index_type(clean_type: &clean::Type, generics: Vec<TypeWithKind>) -> RenderType {
     RenderType {
         name: get_index_type_name(clean_type, true).map(|s| s.as_str().to_ascii_lowercase()),
-        generics: get_generics(clean_type),
+        generics: if generics.is_empty() { None } else { Some(generics) },
     }
 }
 
@@ -246,23 +238,6 @@ fn get_index_type_name(clean_type: &clean::Type, accept_generic: bool) -> Option
     }
 }
 
-/// Return a list of generic parameters for use in the search index.
-///
-/// This function replaces bounds with types, so that `T where T: Debug` just becomes `Debug`.
-/// It does return duplicates, and that's intentional, since search queries like `Result<usize, usize>`
-/// are supposed to match only results where both parameters are `usize`.
-fn get_generics(clean_type: &clean::Type) -> Option<Vec<String>> {
-    clean_type.generics().and_then(|types| {
-        let r = types
-            .iter()
-            .filter_map(|t| {
-                get_index_type_name(t, false).map(|name| name.as_str().to_ascii_lowercase())
-            })
-            .collect::<Vec<_>>();
-        if r.is_empty() { None } else { Some(r) }
-    })
-}
-
 /// The point of this function is to replace bounds with types.
 ///
 /// i.e. `[T, U]` when you have the following bounds: `T: Display, U: Option<T>` will return
@@ -272,27 +247,77 @@ crate fn get_real_types<'tcx>(
     generics: &Generics,
     arg: &Type,
     tcx: TyCtxt<'tcx>,
-    recurse: i32,
-    res: &mut FxHashSet<(Type, ItemType)>,
-) -> usize {
-    fn insert(res: &mut FxHashSet<(Type, ItemType)>, tcx: TyCtxt<'_>, ty: Type) -> usize {
-        if let Some(kind) = ty.def_id_no_primitives().map(|did| tcx.def_kind(did).into()) {
-            res.insert((ty, kind));
-            1
+    recurse: usize,
+    res: &mut Vec<TypeWithKind>,
+) {
+    fn insert_ty(
+        res: &mut Vec<TypeWithKind>,
+        tcx: TyCtxt<'_>,
+        ty: Type,
+        mut generics: Vec<TypeWithKind>,
+    ) {
+        let is_full_generic = ty.is_full_generic();
+
+        if is_full_generic && generics.len() == 1 {
+            // In this case, no need to go through an intermediate state if the generics
+            // contains only one element.
+            //
+            // For example:
+            //
+            // fn foo<T: Display>(r: Option<T>) {}
+            //
+            // In this case, it would contain:
+            //
+            // ```
+            // [{
+            //     name: "option",
+            //     generics: [{
+            //         name: "",
+            //         generics: [
+            //             name: "Display",
+            //             generics: []
+            //         }]
+            //     }]
+            // }]
+            // ```
+            //
+            // After removing the intermediate (unnecessary) full generic, it'll become:
+            //
+            // ```
+            // [{
+            //     name: "option",
+            //     generics: [{
+            //         name: "Display",
+            //         generics: []
+            //     }]
+            // }]
+            // ```
+            //
+            // To be noted that it can work if there is ONLY ONE generic, otherwise we still
+            // need to keep it as is!
+            res.push(generics.pop().unwrap());
+            return;
+        }
+        let mut index_ty = get_index_type(&ty, generics);
+        if index_ty.name.as_ref().map(|s| s.is_empty()).unwrap_or(true) {
+            return;
+        }
+        if is_full_generic {
+            // We remove the name of the full generic because we have no use for it.
+            index_ty.name = Some(String::new());
+            res.push(TypeWithKind::from((index_ty, ItemType::Generic)));
+        } else if let Some(kind) = ty.def_id_no_primitives().map(|did| tcx.def_kind(did).into()) {
+            res.push(TypeWithKind::from((index_ty, kind)));
         } else if ty.is_primitive() {
             // This is a primitive, let's store it as such.
-            res.insert((ty, ItemType::Primitive));
-            1
-        } else {
-            0
+            res.push(TypeWithKind::from((index_ty, ItemType::Primitive)));
         }
     }
 
     if recurse >= 10 {
         // FIXME: remove this whole recurse thing when the recursion bug is fixed
-        return 0;
+        return;
     }
-    let mut nb_added = 0;
 
     if let Type::Generic(arg_s) = *arg {
         if let Some(where_pred) = generics.where_predicates.iter().find(|g| match g {
@@ -301,6 +326,7 @@ crate fn get_real_types<'tcx>(
             }
             _ => false,
         }) {
+            let mut ty_generics = Vec::new();
             let bounds = where_pred.get_bounds().unwrap_or_else(|| &[]);
             for bound in bounds.iter() {
                 if let GenericBound::TraitBound(poly_trait, _) = bound {
@@ -309,41 +335,32 @@ crate fn get_real_types<'tcx>(
                             continue;
                         }
                         if let Some(ty) = x.get_type() {
-                            let adds = get_real_types(generics, &ty, tcx, recurse + 1, res);
-                            nb_added += adds;
-                            if adds == 0 && !ty.is_full_generic() {
-                                nb_added += insert(res, tcx, ty);
-                            }
+                            get_real_types(generics, &ty, tcx, recurse + 1, &mut ty_generics);
                         }
                     }
                 }
             }
+            insert_ty(res, tcx, arg.clone(), ty_generics);
         }
         if let Some(bound) = generics.params.iter().find(|g| g.is_type() && g.name == arg_s) {
+            let mut ty_generics = Vec::new();
             for bound in bound.get_bounds().unwrap_or(&[]) {
                 if let Some(path) = bound.get_trait_path() {
                     let ty = Type::ResolvedPath { did: path.def_id(), path };
-                    let adds = get_real_types(generics, &ty, tcx, recurse + 1, res);
-                    nb_added += adds;
-                    if adds == 0 && !ty.is_full_generic() {
-                        nb_added += insert(res, tcx, ty);
-                    }
+                    get_real_types(generics, &ty, tcx, recurse + 1, &mut ty_generics);
                 }
             }
+            insert_ty(res, tcx, arg.clone(), ty_generics);
         }
     } else {
-        nb_added += insert(res, tcx, arg.clone());
-        if let Some(gens) = arg.generics() {
-            for gen in gens.iter() {
-                if gen.is_full_generic() {
-                    nb_added += get_real_types(generics, gen, tcx, recurse + 1, res);
-                } else {
-                    nb_added += insert(res, tcx, (*gen).clone());
-                }
+        let mut ty_generics = Vec::new();
+        if let Some(arg_generics) = arg.generics() {
+            for gen in arg_generics.iter() {
+                get_real_types(generics, gen, tcx, recurse + 1, &mut ty_generics);
             }
         }
+        insert_ty(res, tcx, arg.clone(), ty_generics);
     }
-    nb_added
 }
 
 /// Return the full list of types when bounds have been resolved.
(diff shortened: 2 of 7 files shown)


jyn514 (commented):
> I used this opportunity to allow generics to have generics themselves

What does this mean? What code does it allow you to represent? I'm not really clear what any of this code is doing ...

src/librustdoc/formats/item_type.rs
```diff
@@ -147,32 +148,33 @@ impl From<DefKind> for ItemType {
...
-            ItemType::Constant => "constant",
-            ItemType::AssocConst => "associatedconstant",
-            ItemType::ForeignType => "foreigntype",
-            ItemType::Keyword => "keyword",
-            ItemType::OpaqueTy => "opaque",
-            ItemType::ProcAttribute => "attr",
-            ItemType::ProcDerive => "derive",
-            ItemType::TraitAlias => "traitalias",
+            Self::Module => "mod",
+            Self::ExternCrate => "externcrate",
```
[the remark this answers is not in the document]
GuillaumeGomez: Arf, I didn't even pay attention... I'll revert this change.

src/librustdoc/html/render/cache.rs
```diff
@@ -194,32 +194,24 @@ crate fn get_index_search_type<'tcx>(
...
 
     Some(IndexItemFunctionType { inputs, output })
 }
 
-fn get_index_type(clean_type: &clean::Type) -> RenderType {
+fn get_index_type(clean_type: &clean::Type, generics: Vec<TypeWithKind>) -> RenderType {
     RenderType {
         name: get_index_type_name(clean_type, true).map(|s| s.as_str().to_ascii_lowercase()),
-        generics: get_generics(clean_type),
+        generics: if generics.is_empty() { None } else { Some(generics) },
```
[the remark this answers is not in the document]
GuillaumeGomez: Because if it's not an option, it'll generate `[]`, which is unneeded. Before to not generate anything then.

GuillaumeGomez:
> > I used this opportunity to allow generics to have generics themselves
>
> What does this mean? What code does it allow you to represent? I'm not really clear what any of this code is doing ...

Let's take an example then!

```rust
struct Foo<A, B>(A, B);

fn foo<A: Display, B: Foo<u32, u64>, C: Foo<A, B>>(arg: A) {}
```

Like this, `arg` will be a generic `A` which itself contains the generics `B` and `C` which themselves have generics. Before that, it was all inlined as arguments of the functions, which is at best incorrect.

src/librustdoc/html/render/cache.rs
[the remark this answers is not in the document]
GuillaumeGomez: I'll add a comment to make it clear.

src/librustdoc/html/render/cache.rs
```diff
@@ -278,27 +253,33 @@ crate fn get_real_types<'tcx>(
...
+            res.push(TypeWithKind::from((index_ty, ItemType::Generic)));
+        } else if let Some(kind) = ty.def_id().map(|did| tcx.def_kind(did).into()) {
+            res.push(TypeWithKind::from((index_ty, kind)));
         } else if ty.is_primitive() {
             // This is a primitive, let's store it as such.
-            res.insert((ty, ItemType::Primitive));
-            1
-        } else {
-            0
+            res.push(TypeWithKind::from((index_ty, ItemType::Primitive)));
```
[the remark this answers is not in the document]
GuillaumeGomez: There is a `ty.is_primitive()` check just like before. Did I miss something?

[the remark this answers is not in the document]
jyn514: Oh, I misread the code. Why did you remove the else branch? What did 0 represent?

src/librustdoc/html/render/cache.rs
```diff
@@ -278,33 +253,84 @@ crate fn get_real_types<'tcx>(
...
+            //
+            // After removing the intermediate (unnecessary) full generic, it'll become:
+            //
+            // ```
+            // [{
+            //     name: "option",
+            //     generics: [{
+            //         name: "Display",
+            //         generics: []
+            //     ]
```
jyn514, line 298:
```suggestion
            //     }]
```

src/librustdoc/html/render/cache.rs
```diff
@@ -278,33 +253,84 @@ crate fn get_real_types<'tcx>(
...
+            // }]
+            // ```
+            //
+            // To be noted that it can work if there is ONLY ONE generic, otherwise we still
```
jyn514, line 302: Why does it only work if there's a single generic argument? It seems weird to have an outer object that just has `name: ""`

src/librustdoc/json/conversions.rs
```diff
@@ -677,6 +677,7 @@ impl FromWithTcx<ItemType> for ItemKind {
             TraitAlias => ItemKind::TraitAlias,
             ProcAttribute => ItemKind::ProcAttribute,
             ProcDerive => ItemKind::ProcDerive,
+            _ => unreachable!(),
```
jyn514, line 680: Please use the name instead of a wildcard.
```suggestion
            Generic => unreachable!(),
```

src/librustdoc/html/render/cache.rs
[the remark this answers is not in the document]
jyn514: I'm confused though - can't you just do this same check wherever it would generate an empty Vec? Why does it have to be specifically here?

jyn514 (commented):
Is there a reference for how the search index works somewhere? Your examples helped but I have no idea how they'll actually affect search, I don't really feel comfortable signing off on this. Maybe @notriddle or @ollie27 are familiar with search?

src/test/rustdoc-js/generics.js
```diff
@@ -3,7 +3,6 @@
 const QUERY = [
   '"R<P>"',
   '"P"',
-  'P',
```
notriddle, line 6: It seems odd that you’re removing the non-quoted search, but not removing the quoted search.

src/librustdoc/html/render/cache.rs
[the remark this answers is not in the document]
GuillaumeGomez: Before it returned the number of items inserted to know if something needed to be added into the HashSet. Now we add all arguments (as long as they have a name).

src/librustdoc/html/render/cache.rs
GuillaumeGomez: It's what I tried to explain but I failed I guess haha.

So basically, arguments can only have a type and generics. What happens in case the argument is a full generic which is `Display + Debug + Write`? Then we can't say it's one type (from a search perspective) but a generic type (with empty name) having three generics.

src/librustdoc/json/conversions.rs
GuillaumeGomez: :+1:

src/librustdoc/html/render/cache.rs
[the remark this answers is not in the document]
GuillaumeGomez: I'm not sure about the other places... In this case, it's contained inside a vec, so if the vec length is 1 or 2, it's not really a problem. For the other places, it might be different.

src/test/rustdoc-js/generics.js
GuillaumeGomez: I updated the test to take the update into account.

GuillaumeGomez:
ping @jyn514

GuillaumeGomez:
I think @jyn514 and I need this PR to be merged to end the rewrite of the rustdoc search engine so I'll set @camelid as reviewer.

r? @camelid

camelid:
> > > I used this opportunity to allow generics to have generics themselves
> > 
> > 
> > What does this mean? What code does it allow you to represent? I'm not really clear what any of this code is doing ...
> 
> Let's take an example then!
> 
> ```rust
> struct Foo<A, B>(A, B);
> 
> fn foo<A: Display, B: Foo<u32, u64>, C: Foo<A, B>>(arg: A) {}
> ```
> 
> Like this, `arg` will be a generic `A` which itself contains the generics `B` and `C` which themselves have generics. Before that, it was all inlined as arguments of the functions, which is at best incorrect.

I don't really understand this example. Why will `A` contain `B` and `C`? Is it some kind of linked-list/tree structure? And if so, why?

Also, `Foo` is a struct, so it can't be a trait bound. I assume you meant something like `trait Foo<A, B> {}`?

How did the search index look previously? Based on your description, it sounds like the type of this function used to be treated as something like `fn(Display, Foo<u32, u64>, Foo<A, B>, A)`? Is that correct? And what will it look like after this change?

Sorry for all the questions; I'm not very familiar with the search-index code ;)

GuillaumeGomez:
Woups, I meant that `C` contains `A` and `B`. It should have been:

```rust
struct Foo<A, B>(A, B);

fn foo<A: Display, B: Foo<u32, u64>, C: Foo<A, B>>(arg: C) {}
```

And indeed, it looks like a tree:

```
arg: [
    {
        type: "Foo",
        generics: [
            {
                type: "Display",
                generics: []
            },
            {
                type: "Foo"
                generics: [
                    {
                        type: "u32",
                        generics: []
                    },
                    {
                        type: "u64",
                        generics: []
                    }
                ]
            }
        ]
    }
]
```

src/librustdoc/formats/item_type.rs
```diff
@@ -48,6 +48,7 @@ crate enum ItemType {
     ProcAttribute = 23,
     ProcDerive = 24,
     TraitAlias = 25,
+    Generic = 26,
```
camelid, line 51: Why are generic parameters treated as items?

GuillaumeGomez: The pure generic ones, like `fn foo<T: Display + Debug>(a: T)`, in here, `a` cannot be both "Display" and "Debug", it needs to go through a level:

```
arg: [
  {
    name: "",
    generics: [
      {
        name: "Display"
        generics: []
      },
      {
        name: "Debug"
        generics: []
      }
    ]
  }
]
```

So in here, the argument with name `""` is a "pure" generic.

camelid: But value parameters like `a: T` don't have item types (at least from looking at the `ItemType` enum definition), so how are they represented?

GuillaumeGomez: Just with an empty name. If they are still present, might be interesting to remove them since they're useless for the search index.

src/librustdoc/html/render/cache.rs
camelid: I just discussed this with Guillaume, and I think I can explain it now.

For purposes of search-indexing, `fn foo<T: Display>(x: T)` is treated as the type `fn(x: Display)`. This makes a search like `trait:Display ->` show `foo`. (It turns out you can filter by item type in rustdoc searches, which is part of what `ItemType` is for.)

However, `fn foo<T: Display + Clone>(x: T)`'s argument *cannot* be reduced into an `ItemType::Trait` because there are two traits involved, the `Display` trait and the `Clone` trait. So instead it's represented as a `ItemType::Generic`, which means there was more than one bound involved.

GuillaumeGomez: Exactly. :)

jyn514:
Oh, sorry, I just saw this ping. Do you still need me to review?

GuillaumeGomez:
If you have time, it'd very appreciated, otherwise I'll try to help @camelid finish it.

jyn514:
It sounds like Camelid understands what's going on better than me, I'll leave them assigned.

src/librustdoc/html/render/cache.rs
```diff
@@ -357,35 +374,35 @@ crate fn get_all_types<'tcx>(
...
     for arg in decl.inputs.values.iter() {
         if arg.type_.is_self_type() {
             continue;
         }
-        let mut args = FxHashSet::default();
+        args.clear();
         get_real_types(generics, &arg.type_, tcx, 0, &mut args);
         if !args.is_empty() {
-            all_types.extend(args);
+            all_types.extend(args.drain(..));
```
camelid, line 387: Why did you move `let mut args` outside of the loop and add `.clear()` and `.drain(..)`? Is there a behavioral difference between the old and new code?

GuillaumeGomez: There are two things happening here:

 * Before you couldn't have the same type multiple times (because of the hashset), which is problematic when you're looking for `u32, u32` for example (function taking two `u32` for example).
 * As for the `clear` and `drain`, it's simply to be sure that the `args` vector only contains what we're looking for and improve memory usage.

camelid: Could you remove the second change and instead have the `let mut args = vec![];` inside the loop? I know it might not be as performant, but having the state persist across iterations makes this change harder to review. I'd rather not mix performance and behavior changes too.

GuillaumeGomez: Is it ok if I split it into two commits?

camelid: I'd rather it be a separate PR; I see some potential edge cases with it, and the slight performance improvement doesn't seem worth it to me (at least for now) for the additional code complexity.

GuillaumeGomez: I'll add a FIXME then.

GuillaumeGomez: Done!

src/librustdoc/html/render/cache.rs
```diff
@@ -357,35 +374,36 @@ crate fn get_all_types<'tcx>(
...
             continue;
         }
-        let mut args = FxHashSet::default();
+        // FIXME: performance wise, it'd be much better to move `args` declaration outside of the
+        // loop and replace this line with `args.clear()`.
+        let mut args = Vec::new();
         get_real_types(generics, &arg.type_, tcx, 0, &mut args);
         if !args.is_empty() {
-            all_types.extend(args);
+            all_types.extend(args.drain(..));
```
camelid, line 388: I think this `drain` can be removed now too, right?

```suggestion
            all_types.extend(args);
```

GuillaumeGomez: Yes... T_T

Saying my good bye to perf improvements. :'(

GuillaumeGomez:
Updated!

GuillaumeGomez:
ping @camelid

src/librustdoc/html/render/cache.rs
```diff
@@ -194,32 +194,24 @@ crate fn get_index_search_type<'tcx>(
...
-        .collect::<Vec<_>>();
+    inputs.retain(|a| a.ty.name.is_some());
+    output.retain(|a| a.ty.name.is_some());
     let output = if output.is_empty() { None } else { Some(output) };
 
     Some(IndexItemFunctionType { inputs, output })
 }
 
-fn get_index_type(clean_type: &clean::Type) -> RenderType {
+fn get_index_type(clean_type: &clean::Type, generics: Vec<TypeWithKind>) -> RenderType {
```
camelid, line 211: Why did you remove `get_generics` and change this function to take the `generics` as an argument? I don't see any changes that would make that necessary.

src/librustdoc/html/render/cache.rs
```diff
@@ -357,35 +374,38 @@ crate fn get_all_types<'tcx>(
...
+                    ret_types.push(TypeWithKind::from((get_index_type(return_type, vec![]), kind)));
                 }
             }
-            ret.into_iter().collect()
         }
-        _ => Vec::new(),
+        _ => {}
     };
-    (all_types.into_iter().collect(), ret_types)
+    (all_types, ret_types)
```
camelid, line 410: Is this the reason you changed the `FxHashSet`s to `Vec`s?

src/librustdoc/html/static/js/search.js
```diff
@@ -345,6 +345,7 @@ window.initSearch = function(rawSearchIndex) {
         // Check for type name and type generics (if any).
         function checkType(obj, val, literalSearch) {
             var lev_distance = MAX_LEV_DISTANCE + 1;
+            var tmp_lev = MAX_LEV_DISTANCE + 1;
```
camelid, line 348: Why this `tmp_lev` change? Is it really necessary for the generics fix? IIUC, the `[NAME]` additions below are necessary, but I don't see why the `tmp_lev` addition is.

src/librustdoc/html/static/js/search.js
```diff
@@ -404,17 +405,17 @@ window.initSearch = function(rawSearchIndex) {
                 // a levenshtein distance value that isn't *this* good so it goes
                 // into the search results but not too high.
                 lev_distance = Math.ceil((checkGenerics(obj, val) + lev_distance) / 2);
-            } else if (obj.length > GENERICS_DATA && obj[GENERICS_DATA].length > 0) {
+            }
+            if (obj.length > GENERICS_DATA && obj[GENERICS_DATA].length > 0) {
```
camelid, line 409: Same here. Why is this necessary? I'm not comfortable reviewing changes to the JS search code other than adding `[NAME]` as necessary.

camelid:
Please also squash your commits together. IIUC, the test update is logically part of the other changes because the testsuite wouldn't pass without the test being updated.

src/librustdoc/html/render/cache.rs
GuillaumeGomez: Because otherwise you need to set it into the returned type. With this, it cannot be forgotten.

src/librustdoc/html/render/cache.rs
GuillaumeGomez: If by "this" you mean the fact that a same type can be present multiple times into generics, then yes. ;)

src/librustdoc/html/static/js/search.js
GuillaumeGomez: True, I'll revert this one.

GuillaumeGomez: Ah no, nevermind! So originally it was declared just below ([here](https://github.com/rust-lang/rust/pull/88268/files#diff-766c50b21af39c8c0abd3fdba5f3c49a73c38b7b66ded4c7939e83b7244dfd3dL385)), but since it's now being used [here](https://github.com/rust-lang/rust/pull/88268/files#diff-766c50b21af39c8c0abd3fdba5f3c49a73c38b7b66ded4c7939e83b7244dfd3dR413) as well, I needed to have the declaration at the top of the function.

src/librustdoc/html/static/js/search.js
GuillaumeGomez: Because before, if we found that the item had a "close enough" name to what we're looking for, it then ran checks on generics as well, otherwise it was going to look into the generics.

Now, in both cases we go look into generics and then we pick the best match out of the two.

And yes, sorry for the JS changes... If that can make you feel better, you can always confirm it's working as expected by running the JS search tests.

GuillaumeGomez:
Squashed the commits and answered your comments!

src/librustdoc/html/static/js/search.js
camelid: The way I interpret your description is that it's not necessary to bundle the JS change with the `html/formats/cache.rs` change; it's a separate bugfix. If so, can you please move this change to a separate PR and assign someone who's more knowledgeable about `search.js` on it? Or ask them to review the JS part of this PR?

> And yes, sorry for the JS changes... If that can make you feel better, you can always confirm it's working as expected by running the JS search tests.

Rustdoc doesn't currently have very exhaustive test coverage, so I'd rather have someone more knowledgeable look over the JS change.

src/librustdoc/html/static/js/search.js
camelid: I don't really understand why the other changes were made. Like with https://github.com/rust-lang/rust/pull/88268#discussion_r711662034, can you please have someone who's more knowledgeable about rustdoc's JS review this part?

src/librustdoc/html/render/cache.rs
camelid: I'm not sure what you mean. I'd like this change to be as small and simple as possible since I haven't interacted with this part of rustdoc before. I'd really appreciate it if you could revert all changes in this PR other than the part that inlines bounds better, because it's really hard for me to understand what the changes are doing when they are all mixed together.

GuillaumeGomez: All the changes are about that, including this one...

src/librustdoc/html/static/js/search.js
GuillaumeGomez: I'll ask @Nemo157.

src/librustdoc/html/static/js/search.js
```diff
@@ -404,17 +405,17 @@ window.initSearch = function(rawSearchIndex) {
...
                 // We can check if the type we're looking for is inside the generics!
                 var olength = obj[GENERICS_DATA].length;
                 for (x = 0; x < olength; ++x) {
-                    lev_distance = Math.min(levenshtein(obj[GENERICS_DATA][x], val.name),
-                                            lev_distance);
+                    tmp_lev = Math.min(levenshtein(obj[GENERICS_DATA][x][NAME], val.name), tmp_lev);
```
Nemo157, line 413: Wouldn't we need to recurse at this point to check whether the generics of the generic match the searched generics?

GuillaumeGomez: Absolutely, it's just that I wanted to keep the current (broken) behaviour for the moment. I plan to add it into the rework of the search engine. But if you prefer that I make this change now, I can as well.

GuillaumeGomez: Done!

notriddle (changes requested):
Other than the naming thing, I really have only one thing to ask. The code changes, in substance, look reasonable to me.

Can you write some test cases for the JavaScript changes?

GuillaumeGomez:
I added the test you suggested @notriddle (in `src/test/rustdoc-js/generics.*`) and improved the naming as suggested. :)

notriddle (approved):

PR #88270, opened 2021-08-23
Handle type ascription type ops in NLL HRTB diagnostics

lqd:
Currently, there are still a few cases of the "higher-ranked subtype error" of yore, 4 of which are related to type ascription. 

This PR is a follow-up to #86700, adding support for type ascription type ops, and makes 3 of these tests output the same diagnostics in NLL mode as the migrate mode (and 1 is now much closer, especially if you ignore that it already outputs an additional error in NLL mode -- which could be a duplicate caused by a lack of normalization like [these comments point out](https://github.com/rust-lang/rust/blob/9583fd1bdd0127328e25e5b8c24dff575ec2c86b/compiler/rustc_traits/src/type_op.rs#L122-L157), or an imprecision in some parts of normalization as [described here](https://github.com/rust-lang/rust/pull/86700#discussion_r689086688)).

Since we discussed these recently:
- [here](https://github.com/rust-lang/rust/pull/86700#discussion_r689158868), cc @matthewjasper, 
- and [here](https://github.com/rust-lang/rust/issues/57374#issuecomment-901500856), cc @Aaron1011.

It should only leave [this TAIT test](https://github.com/rust-lang/rust/blob/9583fd1bdd0127328e25e5b8c24dff575ec2c86b/src/test/ui/type-alias-impl-trait/issue-57611-trait-alias.rs) as still emitting [the terse error](https://github.com/rust-lang/rust/blob/9583fd1bdd0127328e25e5b8c24dff575ec2c86b/src/test/ui/type-alias-impl-trait/issue-57611-trait-alias.nll.stderr).

r? @estebank (so that they shake their fist at NLL's general direction less often) or @nikomatsakis or matthew or aaron, the more the merrier.

### DIFF
--- a/compiler/rustc_mir/src/borrow_check/diagnostics/bound_region_errors.rs
+++ b/compiler/rustc_mir/src/borrow_check/diagnostics/bound_region_errors.rs
@@ -9,7 +9,7 @@ use rustc_middle::ty::{self, Ty, TyCtxt, TypeFoldable};
 use rustc_span::Span;
 use rustc_trait_selection::traits::query::type_op;
 use rustc_trait_selection::traits::{SelectionContext, TraitEngineExt as _};
-use rustc_traits::type_op_prove_predicate_with_span;
+use rustc_traits::{type_op_ascribe_user_type_with_span, type_op_prove_predicate_with_span};
 
 use std::fmt;
 use std::rc::Rc;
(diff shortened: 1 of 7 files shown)


nikomatsakis (approved):

lqd:
An interesting thing @matthewjasper is that last [TAIT test](https://github.com/rust-lang/rust/blob/9583fd1bdd0127328e25e5b8c24dff575ec2c86b/src/test/ui/type-alias-impl-trait/issue-57611-trait-alias.rs) does trigger a higher-ranked subtype error via the new #86700 diagnostics code. It seemed unexpected because of [this comment](https://github.com/rust-lang/rust/blob/9583fd1bdd0127328e25e5b8c24dff575ec2c86b/compiler/rustc_mir/src/borrow_check/diagnostics/bound_region_errors.rs#L65-L72) but at least this case will be now easier to investigate!

PR #88272, opened 2021-08-23
Add bit removal methods to SparseBitMatrix and factor *BitSet relational methods into more extensible trait

willcrichton:
I need the ability to clear the bits out of a row from `SparseBitMatrix`. Currently, all the mutating methods only allow insertion of bits, and there is no way to get access to the underlying data. 

One approach is simply to make `ensure_row` public, since it grants `&mut` access to the underlying `HybridBitSet`. This PR adds the `pub` modifier. However, presumably this method was private for a reason, so I'm open to other designs. I would prefer general mutable access to the rows, because that way I can add many mutating operations (`clear`, `intersect`, etc.) without filing a PR each time :-)

r? @ecstatic-morse

### DIFF
--- a/compiler/rustc_index/src/bit_set.rs
+++ b/compiler/rustc_index/src/bit_set.rs
@@ -247,6 +247,8 @@ impl<T: Idx> BitRelations<BitSet<T>> for BitSet<T> {
     }
 }
 
+// Applies a function to mutate a bitset, and returns true if any
+// of the applications return true
 fn sequential_update<T: Idx>(
     mut self_update: impl FnMut(T) -> bool,
     it: impl Iterator<Item = T>,
@@ -258,6 +260,8 @@ fn sequential_update<T: Idx>(
     changed
 }
 
+// Optimization of intersection for SparseBitSet that's generic
+// over the RHS
 fn sparse_intersect<T: Idx>(
     set: &mut SparseBitSet<T>,
     other_contains: impl Fn(&T) -> bool,
@@ -267,6 +271,10 @@ fn sparse_intersect<T: Idx>(
     set.elems.len() != size
 }
 
+// Optimization of dense/sparse intersection. The resulting set is
+// guaranteed to be at most the size of the sparse set, and hence can be
+// represented as a sparse set. Therefore the sparse set is copied and filtered,
+// then returned as the new set. 
 fn dense_sparse_intersect<T: Idx>(
     dense: &BitSet<T>,
     sparse: &SparseBitSet<T>,
@@ -303,6 +311,10 @@ impl<T: Idx> BitRelations<HybridBitSet<T>> for BitSet<T> {
         match other {
             HybridBitSet::Sparse(sparse) => {
                 let (updated, changed) = dense_sparse_intersect(self, sparse);
+
+                // We can't directly assign the BitSet to the SparseBitSet, and 
+                // doing `*self = updated.to_dense()` would cause a drop / reallocation. Instead,
+                // the BitSet is cleared and `updated` is copied into `self`.
                 self.clear();
                 for elem in updated.iter() {
                     self.insert(*elem);
--- a/compiler/rustc_index/src/bit_set.rs
+++ b/compiler/rustc_index/src/bit_set.rs
@@ -956,7 +956,7 @@ impl<R: Idx, C: Idx> SparseBitMatrix<R, C> {
         Self { num_columns, rows: IndexVec::new() }
     }
 
-    fn ensure_row(&mut self, row: R) -> &mut HybridBitSet<C> {
+    pub fn ensure_row(&mut self, row: R) -> &mut HybridBitSet<C> {
         // Instantiate any missing rows up to and including row `row` with an
         // empty HybridBitSet.
         self.rows.ensure_contains_elem(row, || None);
--- a/compiler/rustc_index/src/bit_set.rs
+++ b/compiler/rustc_index/src/bit_set.rs
@@ -262,79 +262,42 @@ fn sparse_intersect<T: Idx>(
     set: &mut SparseBitSet<T>,
     other_contains: impl Fn(&T) -> bool,
 ) -> bool {
-    let mut changed = false;
-    for i in (0..set.len()).rev() {
-        if !other_contains(&set.elems[i]) {
-            set.elems.remove(i);
-            changed = true;
-        }
-    }
-    changed
+    let size = set.elems.len();
+    set.elems.retain(|elem| other_contains(elem));
+    set.elems.len() != size
 }
 
-impl<T: Idx> BitRelations<SparseBitSet<T>> for BitSet<T> {
-    fn union(&mut self, other: &SparseBitSet<T>) -> bool {
-        sequential_update(|elem| self.insert(elem), other.iter().cloned())
-    }
-
-    fn subtract(&mut self, other: &SparseBitSet<T>) -> bool {
-        sequential_update(|elem| self.remove(elem), other.iter().cloned())
-    }
-
-    fn intersect(&mut self, other: &SparseBitSet<T>) -> bool {
-        self.intersect(&other.to_dense())
-    }
-}
-
-impl<T: Idx> BitRelations<BitSet<T>> for SparseBitSet<T> {
-    fn union(&mut self, other: &BitSet<T>) -> bool {
-        sequential_update(|elem| self.insert(elem), other.iter())
-    }
-
-    fn subtract(&mut self, other: &BitSet<T>) -> bool {
-        sequential_update(|elem| self.remove(elem), other.iter())
-    }
-
-    fn intersect(&mut self, other: &BitSet<T>) -> bool {
-        sparse_intersect(self, |el| other.contains(*el))
-    }
-}
-
-impl<T: Idx> BitRelations<SparseBitSet<T>> for SparseBitSet<T> {
-    fn union(&mut self, other: &SparseBitSet<T>) -> bool {
-        sequential_update(|elem| self.insert(elem), other.iter().cloned())
-    }
-
-    fn subtract(&mut self, other: &SparseBitSet<T>) -> bool {
-        sequential_update(|elem| self.insert(elem), other.iter().cloned())
-    }
-
-    fn intersect(&mut self, other: &SparseBitSet<T>) -> bool {
-        sparse_intersect(self, |el| other.contains(*el))
-    }
-}
-
-impl<T: Idx, S> BitRelations<HybridBitSet<T>> for S
-where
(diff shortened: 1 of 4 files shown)


ecstatic-morse:
`row` is public, so the fact that each row is stored as a `HybridBitSet` is already exposed to consumers, and this change doesn't break encapsulation. However, it still feels a bit wrong to me. Calling `ensure_row` may cause the `rows` vector to grow, which isn't even necessary if you're just clearing bits.

Is there anything you could need beyond `clear`, `intersect` and `remove_bit`? That won't be too much effort, and might benefit other `rustc` users someday. I'd prefer you go that route if possible, but this is pretty low stakes.

willcrichton:
I could implement those methods, although they would still end up calling `ensure_row` internally. An alternative would be to add the following to parallel `SparseBitMatrix::row`:

```rust
fn row_mut(&mut self, row: R) -> Option<&mut HybridBitSet<C>>;
```

This eliminates the weird side effect, and also matches the existing API. But if you have any preference towards lifting the set methods to matrix, I can do that instead.

ecstatic-morse:
My whole point is that they would *not* call `ensure_mut` internally. There's no need to ensure that a row has backing storage just to intersect with or clear it.

willcrichton:
Ah I understand, I can do that.

One other question. For operations like `intersect` and `subtract`, would you like me to implement those methods on `SparseBitSet` / `HybridBitSet` first, and then extend them to `SparseBitMatrix`? I've already had to implement these methods in my own codebase, but they seem useful to have in rustc proper: https://github.com/willcrichton/flowistry/blob/032a5982984291b3d1e791706d0607b7be09410c/src/core/indexed.rs#L204-L234

ecstatic-morse:
> One other question. For operations like `intersect` and `subtract`, would you like me to implement those methods on `SparseBitSet` / `HybridBitSet` first, and then extend them to `SparseBitMatrix`? I've already had to implement these methods in my own codebase, but they seem useful to have in rustc proper: https://github.com/willcrichton/flowistry/blob/032a5982984291b3d1e791706d0607b7be09410c/src/core/indexed.rs#L204-L234

Yes, I think that's the right path, although it means handling all combinations in `{SparseBitSet, BitSet}^2`. I thought we did this already, but I guess I was wrong. This is useful, but a bit more work than I was expecting. I wonder if some parts of `rustc` convert to a dense `BitSet` just to do subtraction? 

You probably know this, but an in-tree impl should take advantage of the fact that `SparseBitSet` is in sorted order, which means you can do both `intersect` and `subtract` in a single pass with `ArrayVec::retain`.

ecstatic-morse:
If you don't want to make the changes to `HybridBitSet` now, I'm willing to sign off on the visibility change for `ensure_row` as long as you add some documentation and a FIXME noting that it should be removed in favor of `subtract_row`, `clear_row`, etc. when the corresponding methods are available on `HybridBitSet`.

Ideally we would make the `HybridBitSet` changes now, though. I'm happy to write up mentoring instructions, since there's a few helper traits (`SubtractFromBitSet`) that should be rewritten into more general forms.

willcrichton:
I'm happy to do the broader `HybridBitSet` changes. Let me know what additional changes you'd like (eg helper traits) and I can do it today / tomorrow.

ecstatic-morse:
Alright, so we have three different bitset types. `SparseBitSet`, `BitSet`, and `HybridBitSet` (the union of the other two). `clear`, `insert` and `remove` can be trivially implemented for all of them (and mostly already have been). We also have three operations we want to implement---intersection, union and subtraction. The current approach uses helper traits to make these operations generic over the different bitsets (intersection is missing at the moment):

https://github.com/rust-lang/rust/blob/b5fe3bc0650279e4283a0511670c0668fc0f2217/compiler/rustc_index/src/bit_set.rs#L213-L225

We want to change these to make **both** sides of the operation generic, just like (for example) [`ops::AddAssign`](https://doc.rust-lang.org/std/ops/trait.AddAssign.html). Note that the existing impls put the mutable argument on the right, unlike `AddAssign`. You can leave this as is, or change it, up to you. This will require 3^2 = 9 impls, but only 4 of those are interesting: the other 5 simply dispatch to the appropriate variant of `HybridBitSet`. The majority of the interesting ones already exist in the code; you just have to collect them in the appropriate place. Don't worry about going from a dense bitset back to a sparse one.

Finally, you should expose the operations as inherent methods on each of the bitset types, just like the current code does.

Once all that is done, it should be easy to add generic versions of `intersect_row`, `subtract_row`, and `union_row` to `SparseBitMatrix`, as well as `remove` and `clear`. You should avoid calling `ensure_row` unless its actually necessary. You'll have a small problem with the lazy, generic version of `union_row`, but you'll find a solution :smile: .

Let me know if any of that is unclear, or if you have a better approach. And if you want to just do the `ensure_row` thing so you can continue working on your project, just say the word. This was a bit more involved than I had first thought. :laughing:

willcrichton:
I'm trying to understand the reason for having separate a `UnionIntoBitSet::union_into` trait method and `BitSet::union` inherent method. Is it for convenience so downstream users don't have to import the `UnionIntoBitSet` trait? As opposed to just having users call `union_into` directly.

ecstatic-morse:
That's correct.

willcrichton:
Still need to add comments and tests, but @ecstatic-morse can you let me know if this general design looks right to you? In short:
* I consolidated the union/subtract (+ intersect) traits into a single `BitRelations` trait (otherwise there was an ungodly number of impl blocks).
* I added a `bit_relations_inherent_impls` macro that adds inherent implementations of each trait method to all three BitSet types.
* I factored out all the existing union/subtract/intersect logic into impls for these traits. The only impls I'm not sure about are the subtract/intersect impls for `HybridBitSet` -- I'm not sure if you want to do something especially efficient in these cases like for union.
* I added all the new methods to `SparseBitMatrix`.

compiler/rustc_index/src/bit_set.rs
```diff
@@ -1025,6 +1107,33 @@ impl<R: Idx, C: Idx> SparseBitMatrix<R, C> {
...
+            Some(Some(row)) => row.subtract(set),
+            _ => false,
+        }
+    }
+
+    pub fn union_row<Set>(&mut self, row: R, set: &Set) -> bool
+    where
+        HybridBitSet<C>: BitRelations<Set>,
+    {
+        self.ensure_row(row).union(set)
```
ecstatic-morse, line 1148: :+1:

The other choice was to add an `IsEmpty` trait and use it to check `set` before reserving the row, but that's overkill.

compiler/rustc_index/src/bit_set.rs
```diff
@@ -208,33 +226,167 @@ impl<T: Idx> BitSet<T> {
...
+            HybridBitSet::Dense(self_dense) => self_dense.subtract(other),
+        }
+    }
+
+    fn intersect(&mut self, other: &HybridBitSet<T>) -> bool {
+        // FIXME(willcrichton): should there be an optimized sparse / dense version?
+        match self {
+            HybridBitSet::Sparse(self_sparse) => self_sparse.intersect(other),
+            HybridBitSet::Dense(self_dense) => {
+                <BitSet<T> as BitRelations<HybridBitSet<T>>>::intersect(self_dense, other)
```
ecstatic-morse, line 400: Is the UFCS significant here? Might be good to leave a comment if so. Collisions are why the old helper traits had different names than the inherent methods.

Regarding the FIXMEs, the only obvious potential optimization is when we're intersecting a dense bitset with a sparse one. Since we know the end result will be no bigger than the sparse bitset, we could use it to replace `self` before intersecting it with the dense bitset. You'll be fighting the borrow checker with this one, so no need to do it now if you don't want.

ecstatic-morse (commented):
Thanks @willcrichton!

I left a few notes. There's an issue I didn't foresee with the `SparseBitSet` impls, which is going to cause problems, but for the most part this looks good.

compiler/rustc_index/src/bit_set.rs
```diff
@@ -208,33 +226,180 @@ impl<T: Idx> BitSet<T> {
...
+    fn union(&mut self, other: &SparseBitSet<T>) -> bool {
+        sequential_update(|elem| self.insert(elem), other.iter().cloned())
+    }
+
+    fn subtract(&mut self, other: &SparseBitSet<T>) -> bool {
+        sequential_update(|elem| self.remove(elem), other.iter().cloned())
+    }
+
+    fn intersect(&mut self, other: &SparseBitSet<T>) -> bool {
+        self.intersect(&other.to_dense())
```
ecstatic-morse, line 285: This one is suboptimal. It would be good to leave a `FIXME` here as well.

ecstatic-morse (changes requested):
Almost there. All of my old review comments have been resolved satisfactorily. There's a bug and an unnecessary dense `BitSet` allocation in the new code, however.

Can I ask for some unit tests for the new impls? I'm worried that I might have missed something just from staring at it.

ecstatic-morse (commented):
Alright! This is basically there. I left one nit and one question. Thanks for sticking with it @willcrichton. Really we should be using `proptest` or `quickcheck` for better test coverage, but I don't think we can do that in-tree. We have smoke tests at least, and I've looked things over as carefully as I can. Next time we'll just make the private method public :smile:.

Regarding the dense -> sparse transformation, I don't think we should try hard to make it happen. It's not always more efficient, since it can lead to thrashing when we're adding/subtracting elements while close to the threshold. I think the dense/sparse intersect case is worth it, since the result is guaranteed to be below the sparse threshold and it saves us from traversing the dense bitset to clear it, but that's pretty much the only case.

ecstatic-morse (approved):

compiler/rustc_index/src/bit_set.rs
```diff
@@ -956,7 +1071,7 @@ impl<R: Idx, C: Idx> SparseBitMatrix<R, C> {
         Self { num_columns, rows: IndexVec::new() }
     }
 
-    fn ensure_row(&mut self, row: R) -> &mut HybridBitSet<C> {
+    pub fn ensure_row(&mut self, row: R) -> &mut HybridBitSet<C> {
```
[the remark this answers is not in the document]
willcrichton: Ah no, that's leftover from the initial PR

ecstatic-morse:
I gave this one last look today. Given the number of bugs found during the review process, it seemed wise. Everything looks correct.

For posterity, in addition to the `SparseBitMatrix` changes, this PR extends `HybridBitSet` with in-place versions of the fundamental set operations. They aren't quite optimal, since they fail to take into account the sorted-ness of the elements in a sparse bitset, but should be no slower than the existing impls. As a result, we have the option to use `HybridBitSet` in more places (e.g. dataflow).

@bors r+

PR #88273, opened 2021-08-23
Fix references to `ControlFlow` in docs

jhpratt:
The `Iterator::for_each` method previously stated that it was not possible to use `break` and `continue` in it — this has been updated to acknowledge the stabilization of `ControlFlow`. Additionally, `ControlFlow` was referred to as `crate::ops::ControlFlow` which is not the correct path for an end user.

r? @jyn514

### DIFF
--- a/library/core/src/iter/traits/iterator.rs
+++ b/library/core/src/iter/traits/iterator.rs
@@ -1957,8 +1957,8 @@ pub trait Iterator {
     /// assert_eq!(it.next(), Some(&40));
     /// ```
     ///
-    /// While you cannot `break` from a closure, the [`crate::ops::ControlFlow`]
-    /// type allows a similar idea:
+    /// While you cannot `break` from a closure, the [`ControlFlow`] type allows
+    /// a similar idea:
     ///
     /// ```
     /// use std::ops::ControlFlow;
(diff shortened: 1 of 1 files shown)


the8472:
I don't see how that relates to `for_each`, barring panics it unconditionally runs to completion, `ControlFlow` doesn't change that.

jhpratt:
Returning `ControlFlow::Break(())` wouldn't terminate early? That seems unexpected.

(No I didn't try it. I made an assumption)

the8472:
It requires a `()` return type for the closure. You need `try_for_each` or `try_fold` if you want to return early.

jhpratt:
Ah, right. Sometimes I miss the obvious. I'll revert that bit of the commit, leaving the fix to the path elsewhere.

jyn514:
@jhpratt is there a reason you picked me particularly to review? I'm not on the libs team.

jyn514:
That said, this looks good to me :)

@bors r+ rollup

jhpratt:
I thought that you've done stuff with docs in the past. Not that the reviewer really matters here admittedly.

jyn514:
I mostly work on rustdoc; different sort of docs :P

jhpratt:
I had the right idea 😂 Good to know for the future.

PR #88280, opened 2021-08-24
Handle match statements with non exhaustive variants in closures

### LINKED ISSUE #59: Handle nonexhaustive in match pattern

roxelo:
This PR ensures that the behavior for match statements with non exhaustive variants is the same inside and outside closures.

If we have a non-exhaustive SingleVariant which is defined in a different crate, then we should handle the case the same way we would handle a MultiVariant: borrow the match discriminant.

Closes https://github.com/rust-lang/project-rfc-2229/issues/59
r? @nikomatsakis


nikomatsakis (approved):
Great! 🎉

PR #88281, opened 2021-08-24
Update unstable docs for `asm!` macro

asquared31415:
This adds documentation that SPIR-V is supported, expands on the restrictions for labels, and has some minor cleanups or clarifications.

r? @joshtriplett

### DIFF
--- a/src/doc/unstable-book/src/library-features/asm.md
+++ b/src/doc/unstable-book/src/library-features/asm.md
@@ -367,11 +367,11 @@ unsafe {
 
 ## Labels
 
-Any reuse of a named label, local or otherwise, can result in a assembler or linker error in the best case, ICEs or segfaults in the compiler in other cases, and generally unexpected behavior. Reuse of a named label can happen in a variety of ways including:
+Any reuse of a named label, local or otherwise, can result in a assembler or linker error or may cause other strange behavior. Reuse of a named label can happen in a variety of ways including:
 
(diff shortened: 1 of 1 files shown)


asquared31415:
cc @Amanieu @bjorn3 @nbdd0121

src/doc/unstable-book/src/library-features/asm.md
```diff
@@ -373,9 +367,15 @@ unsafe {
 
 ## Labels
 
-The compiler is allowed to instantiate multiple copies an `asm!` block, for example when the function containing it is inlined in multiple places. As a consequence, you should only use GNU assembler [local labels] inside inline assembly code. Defining symbols in assembly code may lead to assembler and/or linker errors due to duplicate symbol definitions.
+Any reuse of a named label, local or otherwise, can result in a assembler or linker error in the best case, ICEs or segfaults in the compiler in other cases, and generally unexpected behavior. Reuse of a named label can happen in a variety of ways including:
```
Amanieu, line 370: ICEs and segfaults caused by assembly code should be reported as bugs in the compiler and fixed, they are definitely not an expected behavior. I think just mentioning an assembler/linker error is enough.

Amanieu (commented):
@bors r+

asquared31415:
I don't think that registered, doesn't bors normally immediately respond with a confirmation?

PR #88282, opened 2021-08-24
Optimize BinaryHeap::extend from Vec

Neutron3529:
This improves the performance of extending `BinaryHeap`s from vectors directly. Future work may involve extending this optimization to other, similar, cases where the length of the added elements is well-known, but this is not yet done in this PR.

### DIFF
--- a/library/alloc/src/collections/binary_heap.rs
+++ b/library/alloc/src/collections/binary_heap.rs
@@ -1584,6 +1584,14 @@ impl<T: Ord, I: IntoIterator<Item = T>> SpecExtend<I> for BinaryHeap<T> {
     }
 }
 
+impl<T: Ord> SpecExtend<Vec<T>> for BinaryHeap<T> {
+    fn spec_extend(&mut self, ref mut other: Vec<T>) {
+        let start = self.data.len();
+        self.data.append(other);
+        self.rebuild_tail(start);
+    }
+}
+
 impl<T: Ord> SpecExtend<BinaryHeap<T>> for BinaryHeap<T> {
     fn spec_extend(&mut self, ref mut other: BinaryHeap<T>) {
         self.append(other);


library/alloc/src/collections/binary_heap.rs
```diff
@@ -1568,20 +1568,36 @@ impl<T: Ord, I: IntoIterator<Item = T>> SpecExtend<I> for BinaryHeap<T> {
...
-        let iterator = iter.into_iter();
-        let (lower, _) = iterator.size_hint();
-
-        self.reserve(lower);
-
-        iterator.for_each(move |elem| self.push(elem));
+        let start = self.data.len();
+        let extend = RebuildTail(self, start);
+        extend.0.data.extend(iter);
+        // `self.rebuild_tail(start)` is done by `drop(extend)`
```
SkiFire13, line 1600: I'm a bit worried about the panic behaviour. I can see that you want to maintain the heap property even when the iterator panics, like the old implementation did, however contrary to the old implementation this approach could result in an abort if both the iterator and the comparison function of `T` panic.

Neutron3529: comparison function of T panic could be fine since we could not guarantee the heap order without comparison function.

Since we have the bound <T:Ord>
(Ord: Trait for types that form a total order. exactly one of a < b, a == b or a > b is true)
which means, comparison function of T should not panic.

If it panic, T could not satisfy the Ord trait, thus we could not guarantee a correct result.

SkiFire13: > comparison function of T panic could be fine since we could not guarantee the heap order without comparison function.

My problem is not with the incorrect order, rather that if both the iterator and the comparison function panics then the process aborts, without giving the user the ability to catch it. 

Hence the question: do we value more the heap property in case of a single panic or avoiding an abort in case of a double panic?
In my opinion not preserving the heap ordering in case of panic is not a big deal because any other method on `BinaryHeap` does that, including your method. Avoiding an abort instead might be more important because there are legitimate cases where you don't want to abort if one task panics, for example web frameworks.

library/alloc/src/collections/binary_heap.rs
```diff
@@ -1568,20 +1568,36 @@ impl<T: Ord, I: IntoIterator<Item = T>> SpecExtend<I> for BinaryHeap<T> {
     }
 }
 
+impl<T: Ord> SpecExtend<Vec<T>> for BinaryHeap<T> {
```
the8472, line 1571: This shouldn't be necessary if `extend_desugared` now delegates to `vec.extend` which already has specializations to cover this case.

library/alloc/src/collections/binary_heap.rs
the8472: This has been marked as resolved, what's the conclusion of the discussion?

the8472:
BinaryHeap doesn't have many benchmarks, maybe some could be added to show an improvement.

library/alloc/src/collections/binary_heap.rs
SkiFire13: Arguably it could be optimized to append from `self` to `other` if `other` has more elements since this would copy less data while still needing a full rebuild anyway. This is not the case of the current implementation though.

SkiFire13:
I'm not sure if this is something for this PR, but I just noticed that the current specialization is on `I` rather than `I::IntoIter`. This means that only `BinaryHeap<T>` and `Vec<T>` benefit from these specializations, while `std::vec::IntoIter` and `std::collections::binary_heap::IntoIter` do not.

https://github.com/rust-lang/rust/blob/47ab5f7ce27397310bd8359b8db1504fbf8a9b59/library/alloc/src/collections/binary_heap.rs#L1550-L1552

For comparison this is how it's implemented for `Vec`:

https://github.com/rust-lang/rust/blob/47ab5f7ce27397310bd8359b8db1504fbf8a9b59/library/alloc/src/vec/mod.rs#L2555-L2557

I believe we should copy `Vec`'s code and specialize `SpecExtend` for `std::vec::IntoIter` and `std::collections::binary_heap::IntoIter` instead.

Mark-Simulacrum:
Can you please inline the relevant pieces of discussion / why this is a win in general, etc. from the internals thread? This both avoids having to re-read that for context and ensures it's preserved for any future visitor (into git commit history).

Neutron3529:
> Can you please inline the relevant pieces of discussion / why this is a win in general, etc. from the internals thread? This both avoids having to re-read that for context and ensures it's preserved for any future visitor (into git commit history).

I edited the first post:

as scottmcm mentioned:
> `BinaryHeap::append` already has heuristics for how to extend `BinaryHeap<T>` with `Vec<T>`; updating extend to do the same seems perfectly reasonable.

is it enough?

library/alloc/src/collections/binary_heap.rs
Neutron3529: > use `AssertUnwindSafe` or implement `UnwindSafe`, which are both kind of asserting that you don't care about potential inconsistencies.

As [the book](https://doc.rust-lang.org/book/ch09-03-to-panic-or-not-to-panic.html) said:
> If someone calls your code and passes in values that don’t make sense, the best choice might be to call panic! and alert the person using your library to the bug in their code so they can fix it during development. 

What's more, keep potential logic error in the program is not a good choice, and a simply change could fix that issue:
```
impl Ord for NotNaN {
    fn cmp(&self, other: &Self) -> Ordering {
        self.0.partial_cmp(&other.0).unwrap_or(Ordering::Equal)
    }
}
```
if one does not care about that NaN would lead to an invaild state, that modification is totally OK.
Otherwise, panic could be a better choice that it would notify developer that they made a mistake

SkiFire13: > As the book said:
> > If someone calls your code and passes in values that don’t make sense, the best choice might be to call panic! and alert the person using your library to the bug in their code so they can fix it during development.

That's completly unrelated to `AssertUnwindSafe` though. I'm not saying you shouldn't panic, do it if you think it's the right choice, but that alone won't make you observe broken logic invariants of e.g. `BinaryHeap`. To observe them you need to use `AssertUnwindSafe` or implement `UnwindSafe`, and by doing that you're asserting that you don't care if some logic invariants are broken.

> What's more, keep potential logic error in the program is not a good choice, and a simply change could fix that issue:
> ```
> impl Ord for NotNaN {
>     fn cmp(&self, other: &Self) -> Ordering {
>         self.0.partial_cmp(&other.0).unwrap_or(Ordering::Equal)
>     }
> }
> ```

That's the opposite of "not keeping a logic error". That way if a logic error happens it will silently succeed, even though it shouldn't! A panic there is the right choice because if a `NaN` reached that point then the program reached a state that shouldn't have been possible and some other piece of code had a bug.

JohnCSimon:
Triage: setting this to waiting on review

Mark-Simulacrum:
It looks like this PR is doing two separate things:

* Adding a specialization for extending with a `Vec<T>`
* Modifying the non-specialized extend algorithm

This first of these makes sense, seems fine, but the second I don't see a description of the goal in this PR. It also seems like these two changes are orthogonal and should belong in separate PRs?

FWIW, saying "discussion is here" and pointing into the middle of an internals thread is not ideal; the PR description should ideally be self-contained or at least only point to documents or summary comments, not just discussion threads. This will speed up review and make the flow easier for everyone.

@rustbot author (Please adjust the label with `@rustbot ready` when you're ready for another review).

JohnCSimon:
Ping from triage:

@Neutron3529 - can you please address the merge conflicts and 
adjust the label with `@rustbot ready` when you're ready for another review

Neutron3529:
@rustbot ready
I deleted the improvement that could speed up the default extend method which may end up with a double panic.
then, there is no conflict.

It might be meanful that mark `fn rebuild_tail` as `pub unsafe`, or add some new function like `extend_batch` to gain benefit from the fast `rebuild_tail` algorithm

Mark-Simulacrum:
Please update the PR description to reflect the current state of this PR. Otherwise, this PR looks good to me, r=me with the description updated and commits squashed.

Neutron3529:
@rustbot ready
done~

Issue #88260, opened 2021-08-23
`unused_unsafe` false negative in closures in unsafe functions


LeSeulArtichaut:
### Code

I tried this code:

```rust
pub unsafe fn unsf() {}

pub unsafe fn foo() {
    let _ = || unsafe { unsf() };
}
```

I expected to see this happen: compiler lints about the unnecessary unsafe block

Instead, this happened: code compiles without warnings

### Version it worked on

It most recently worked on: Rust 1.51.

```rs
warning: unnecessary `unsafe` block
 --> <source>:4:16
  |
3 | pub unsafe fn foo() {
  | ------------------- because it's nested under this `unsafe` fn
4 |     let _ = || unsafe { unsf() };
  |                ^^^^^^ unnecessary `unsafe` block
  |
  = note: `#[warn(unused_unsafe)]` on by default
```

### Version with regression

`rustc --version --verbose`:
```
rustc 1.56.0-nightly (af140757b 2021-08-22)
binary: rustc
commit-hash: af140757b4cb1a60d107c690720311ba8e06e7de
commit-date: 2021-08-22
host: x86_64-unknown-linux-gnu
release: 1.56.0-nightly
LLVM version: 13.0.0
```

@rustbot modify labels: +regression-from-stable-to-stable -regression-untriaged

PR #88286, opened 2021-08-24
Remove unnecessary unsafe block in `process_unix`

LeSeulArtichaut:
Because it's nested under this unsafe fn!

This block isn't detected as unnecessary because of a bug in the compiler: #88260.


dtolnay (approved):
Thanks!

PR #88288, opened 2021-08-24
Experimental new MIR optimization pass: Replace wildcard match with individual matches 

### LINKED ISSUE #88051: Semantically equal `Ordering` matches generate different assembly
https://rust.godbolt.org/z/orxr8sffq

I tried this code:

```rust
use std::cmp::Ordering;

pub fn a(x: i32, y: i32) -> i32 {
    match x.cmp(&y) {
        Ordering::Greater => 3,
        Ordering::Equal => 2,
        Ordering::Less => 1,
    }
}

pub fn b(x: i32, y: i32) -> i32 {
    match x.cmp(&y) {
        Ordering::Greater => 3,
        Ordering::Equal => 2,
        _ => 1,
    }
}
```

I expected to see this happen: They both run at the exact same speed and both emit the exact same assembly.

Instead, this happened:
```
example::a:
        xor     ecx, ecx
        cmp     edi, esi
        sete    cl
        xor     ecx, 3
        cmp     edi, esi
        mov     eax, 1
        cmovge  eax, ecx
        ret

example::b:
        xor     eax, eax
        cmp     edi, esi
        setne   al
        mov     rcx, -1
        cmovge  rcx, rax
        xor     eax, eax
        cmp     rcx, 1
        sete    al
        lea     edx, [rax + rax]
        add     edx, 1
        test    rcx, rcx
        mov     eax, 2
        cmovne  eax, edx
        ret
```

It appears `b` has far more assembly generated for it which doesn't make sense because semantically `a` and `b` are the same.
`b` is probably slower. Either way, it shouldn't generate this much instructions.
I believe this doesn't just apply to `Ordering` but possibly to many more similar cases.

@rustbot label: +I-slow

hkratz:
This new pass replaces a wildcard match which only matches few (currently one or two) variants with a match on the individual discriminants and the fallback basic block with an unreachable one. This allows further optimizations in the backend. 

TODO:
* benchmark
* determine best location to insert this pass
* check with one/three matches as well.
* tests

Fixes #88051

### DIFF
--- a/compiler/rustc_mir_transform/src/lib.rs
+++ b/compiler/rustc_mir_transform/src/lib.rs
@@ -63,6 +63,7 @@ mod remove_noop_landing_pads;
 mod remove_storage_markers;
 mod remove_unneeded_drops;
 mod remove_zsts;
+mod replace_few_wildcard_matches;
 mod required_consts;
 mod separate_const_switch;
 mod shim;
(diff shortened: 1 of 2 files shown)


hkratz:
Seems to be a net loss in the general case. I will think about it some more.

@rustbot label +S-waiting-on-author -S-waiting-on-review

JohnCSimon:
@hkratz 
Thanks for your contribution.

@rustbot label: +S-inactive

PR #88289, opened 2021-08-24
Fixes for LLVM change 0f45c16f2caa7c035e5c3edd40af9e0d51ad6ba7

durin42:
More details in the individual commit messages, but the summary is: LLVM deleted an unused-to-them method that we used, we worked around it to avoid annoying cleanup/restructuring in the Rust-side code.

### DIFF
--- a/compiler/rustc_llvm/llvm-wrapper/RustWrapper.cpp
+++ b/compiler/rustc_llvm/llvm-wrapper/RustWrapper.cpp
@@ -270,34 +270,30 @@ extern "C" void LLVMRustAddFunctionAttribute(LLVMValueRef Fn, unsigned Index,
                                              LLVMRustAttribute RustAttr) {
   Function *A = unwrap<Function>(Fn);
   Attribute Attr = Attribute::get(A->getContext(), fromRust(RustAttr));
-  AttrBuilder B(Attr);
-  A->addAttributes(Index, B);
+  A->addAttribute(Index, Attr);
 }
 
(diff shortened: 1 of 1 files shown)


compiler/rustc_llvm/llvm-wrapper/RustWrapper.cpp
```diff
@@ -270,8 +270,18 @@ extern "C" void LLVMRustAddFunctionAttribute(LLVMValueRef Fn, unsigned Index,
...
   A->addAttributes(Index, B);
+#else
+  A->addAttribute(Index, Attr);
+#endif
+}
+
+static void AddAttributes(Function* F, unsigned Index, AttrBuilder B) {
+  for (Attribute A : AttributeSet::get(F->getContext(), B)) {
+    F->addAttribute(Index, A);
+  }
```
nikic, line 284: This is going to construct a lot of unnecessary intermediate attribute lists.

nikic:
I think the right way now would be use lower-level APIs, something like this:
```
AttributeList AL = F->getAttributes();
AL = AL.addAttributes(F->getContext(), Index, B);
F->setAttributes(AL);
```
@aeubanks Does that sound right?

aeubanks:
These all look like single attributes, is an `AttrBuilder` necessary? Can you just directly create and add an `Attribute`?

nikic:
@aeubanks Good point, I missed that. Yes, it should be possible to use APIs like `Attribute::getWithDereferenceableBytes()` to get the Attribute and add that.

nikic (approved):
Looks good! Could you please squash the commits?

PR #88292, opened 2021-08-24
Enable --generate-link-to-definition for rustc's docs

SkiFire13:
cc @jyn514

### DIFF
--- a/src/bootstrap/doc.rs
+++ b/src/bootstrap/doc.rs
@@ -589,6 +589,7 @@ impl Step for Rustc {
         cargo.rustdocflag("-Zunstable-options");
         cargo.rustdocflag("-Znormalize-docs");
         cargo.rustdocflag("--show-type-layout");
+        cargo.rustdocflag("--generate-link-to-definition");
         compile::rustc_cargo(builder, &mut cargo, target);
         cargo.arg("-Zunstable-options");
         cargo.arg("-Zskip-rustdoc-fingerprint");


src/bootstrap/builder.rs
```diff
@@ -1543,6 +1543,12 @@ impl<'a> Builder<'a> {
             }
         }
 
+        // For now --generate-link-to-definition is only available on nightly
+        // TODO(SkiFire13): Remove the stage > 0 check when beta reaches 1.56
```
jyn514: This should be done by whoever promotes the beta compiler :)
```suggestion
        // cfg(not(bootstrap)), remove the stage > 0 check when beta reaches 1.56
```

src/bootstrap/builder.rs
```diff
@@ -1543,6 +1543,12 @@ impl<'a> Builder<'a> {
...
+        if (cmd == "doc" || cmd == "rustdoc") && stage > 0 {
```
jyn514, line 1548: I think this is actually unnecessary, sorry to mislead you before. The builder just won't pass the flags unless it's running rustdoc.
```suggestion
        if stage > 0 {
```

jyn514:
> src/test/helpers/exit_code.rs.html:32: broken link - `std/os/imp/unix/process/trait.ExitStatusExt.html`

Hmm, that looks like a bug in rustdoc - it shouldn't be generate links to private items. @GuillaumeGomez you may want to take a look

GuillaumeGomez:
Funny one. The whole `os` module is cursed though. I'll open an issue to not forget.

GuillaumeGomez:
I can't replace the issue locally... Maybe rebase?

GuillaumeGomez:
#88619 was merged so this PR is now unblocked. Please rebase and then we'll make a final review.

src/bootstrap/builder.rs
```diff
@@ -1543,6 +1543,12 @@ impl<'a> Builder<'a> {
             }
         }
 
+        // cfg(not(bootstrap)), remove the stage > 0 check when beta reaches 1.56
```
Mark-Simulacrum, line 1546: beta is now 1.56, so this can just be removed now, I think

src/bootstrap/builder.rs
```diff
@@ -1543,6 +1543,9 @@ impl<'a> Builder<'a> {
             }
         }
 
+        rustdocflags.arg("-Zunstable-options");
+        rustdocflags.arg("--generate-link-to-definition");
```
camelid, line 1547: Hmm, is this the right place to put the check? I would expect it to go in `src/bootstrap/doc.rs` like for `--show-type-layout`. At least, this seems like it should be wrapped in an `if mode == Mode::Rustdoc`.

jyn514: @camelid --show-type-layout is only enabled for the compiler and tools, so it can't be enabled in builder unconditionally. I think it makes sense to use builder here so it doesn't have to be readded in each `Step` in doc.rs.

Good catch on `mode == Rustdoc` though :)

Mark-Simulacrum: We probably want this for all docs, right? In that case I don't believe there's a need to limit to just Mode::Rustdoc.

camelid: > I think it makes sense to use builder here so it doesn't have to be readded in each Step in doc.rs.

Ok, I was wondering if that was the reason 👍

jyn514: @Mark-Simulacrum if we don't check the mode, it will be passed unconditionally for all steps, even compiles, right? I guess it doesn't really matter since most of the time rustdoc isn't invoked though.

Mark-Simulacrum: Right, RUSTDOCFLAGS should be fine to set unconditionally.

camelid:
I would rather enable `--generate-link-to-definition` on just the rustc API docs to start with. There has not been an RFC or an FCP for `--generate-link-to-definition`, and enabling it for the `std` docs feels like we're committing to this feature somewhat.

GuillaumeGomez:
I agree with @camelid. There are still a lot of things to do on this feature (apart from being on the road to stabilize it) so let's take one step at a time.

camelid:
@SkiFire13 Could you change this to only enable it for the rustc docs? If you grep for `--show-type-layout` in `src/bootstrap`, just add another argument for `--generate-link-to-definition` at whatever places show up in the grep output.

PR #88298, opened 2021-08-24
Errorkind reorder

ijackson:
I was doing a bit more work in this area and the untidiness of these two orderings bothered me.

The commit messages have the detailed rationale.  For your convenience, I c&p them here:

```
    io::ErrorKind: rationalise ordering in main enum
    
    It is useful to keep some coherent structure to this ordering.  In
    particular, Other and Uncategorized should be next to each other, at
    the end.
    
    Also it seems to make sense to treat UnexpectedEof and OutOfMemory
    specially, since they are not like the other errors (despite
    OutOfMemory also being generatable by some OS errors).
    
    So:
     * Move Other to the end, just before Uncategorized
     * Move Unsupported to between Interrupted and UnexpectedEof
     * Add some comments documenting where to add things
```

```
    io::Error: alphabeticise the match in as_str()
    
    There was no rationale for the previous ordering.
```

r? kennytm   since that's who rust-highfive picked before, in #88294 which I accidentally closed.


library/std/src/io/error.rs
```diff
@@ -307,23 +314,24 @@ pub enum ErrorKind {
 impl ErrorKind {
     pub(crate) fn as_str(&self) -> &'static str {
         use ErrorKind::*;
+        // Strictly alphabetical, please.  (Sadly rustfmt cannot do this yet.)
```
dtolnay, line 317: If you think it's warranted to check this automatically, https://crates.io/crates/remain

dtolnay (approved):

dtolnay:
Thanks!
@bors r+ rollup

library/std/src/io/error.rs
ijackson: Hah thanks :-).  That would be nice, indeed.

However, I don't think this is used in rust-lang/rust already.  I'm not at all sure what the rules/culture/whatever is for dependencies.  If it's just for this one instance, perhaps it's not worth it?

bjorn3: I would strongly prefer if no proc-macros are used by libstd or it's dependencies.

Issue #80695, opened 2021-01-04
Tracking Issue for extra unix ExitStatusExt methods


ijackson:
Feature gate: `#![feature(unix_process_wait_more)]`

This is a tracking issue for a collection of logically necessary but previously missing methods on `ExitStatusExt`, the Unix-specific extension trait for `ExitStatus`.

### Public API

```rust
// std::src::sys::unix::ext::process

pub trait ExitStatusExt : private::Sealed {
    // this feature gate adds these methods:
    fn core_dumped(&self) -> bool; // WCOREDUMP
    fn stopped_signal(&self) -> Option<i32>; // WIFSTOPPED? WSTOPSIG
    fn continued(&self) -> bool; // WIFCONTINUED
    fn into_raw(self) -> i32;
}
```

### Steps / History

- [x] Implementation: #79982
- [x] Check that nothing broke due to us sealing the `ExitStatusExt` trait on both Unix and Windows
- [x] What to do about the Fuchsia implementation? Needs eyes from a Fuchsia dev
- [x] Final commenting period (FCP)
- [x] Stabilization PR: https://github.com/rust-lang/rust/pull/88300

### Unresolved Questions

- None; <strike>other than the breakage risk due to sealing this trait.</strike> the trait has been sealed for some time now.

ijackson:
The situation with Fuchsia has been discussed in #58590, where it seems to have been agreed that Fuchsia ought not to be treated as Unix.  The current situation, with the stub impls, is [considered fine for now](https://github.com/rust-lang/rust/issues/58590#issuecomment-853408264), apparently.  So I have ticked that box in the checklist.

I will propose an MR to stabilise this.

PR #79982, opened 2020-12-12
Add missing methods to unix ExitStatusExt

ijackson:
These are the methods corresponding to the remaining exit status examination macros from `wait.h`.  `WCOREDUMP` isn't in SuS but is it is very standard.  I have not done portability testing to see if this builds everywhere, so I may need to Do Something if it doesn't.

There is also a bugfix and doc improvement to `.signal()`, and an `.into_raw()` accessor.

This would fix #73128 and fix #73129.  Please let me know if you like this direction, and if so I will open the tracking issue and so on.

If this MR goes well, I may tackle #73125 next - I have an idea for how to do it.

PR #88300, opened 2021-08-24
Stabilise unix_process_wait_more, extra ExitStatusExt methods

ijackson:
This stabilises the feature `unix_process_wait_more`.  Tracking issue #80695, FCP needed.

This was implemented in #79982 and merged in January.

### DIFF
--- a/library/std/src/os/unix/process.rs
+++ b/library/std/src/os/unix/process.rs
@@ -237,27 +237,27 @@ pub trait ExitStatusExt: Sealed {
     fn signal(&self) -> Option<i32>;
 
     /// If the process was terminated by a signal, says whether it dumped core.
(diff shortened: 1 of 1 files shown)


Mark-Simulacrum:
re-assigning to @yaahc to kick off T-libs-api FCP

yaahc:
@rfcbot merge

BurntSushi:
@ijackson Thanks, that makes sense! I don't have a strong opinion either way, and I'm happy to defer to the judgment of others here. One thing that might help is to leave a link or something to your comment here near the trait definition (not in public API docs) so if someone has the same misunderstanding as me, they'll be able to see that the current verbiage is indeed intentional.

PR #88301, opened 2021-08-24
Modified document body to adjust left margin according to toggling on off of side bar in docs.

amaush:
When browsing the documentation locally, I frequently have to jump around different files and the keeping the side bar open is not possible because it overlaps with the page. Like this
old_open_menu

The minimal changes that I made allow the menu to stay open thus easing navigation within a project.
open_menu

r? @steveklabnik


GuillaumeGomez:
This is a good idea. Not sure if changing the `left` property on the `<body>` is a good idea though...

amaush:
The other option is to apply the fix on the main class but that leaves the search bar partially hidden which could be fine I suppose. 

main_class

Another minor issue is that without applying some kind of css animation, the toggling is rather abrupt. I applied the minimum change that would get the side bar to stop overlaying the text in keeping with the established style already in the file as a guideline. I don't know what the policy is re: transitions which would make it more polished.

GuillaumeGomez:
Please add a CSS transition then.

I just thought about something as well: we could make the files sidebar top not go up to the top of the page so we don't have to wonder about the search input. Reminds me of what we did recently on docs.rs haha.

@rust-lang/rustdoc: Which approach do you prefer?

amaush:
I added a transition to make it feel a bit more natural as well as fixing the logo which was obscured previously. 

transition

JohnCSimon:
ping from triage:
@amaush 
thanks.

jsha:
Hi @amaush! Thanks for fixing this. I definitely agree that the file
selector sidebar should not overlay the document content. However, I think
this is not quite the way to do it.

Instead, I'd like to see the sidebar be a sibling element to the main
content, as it is for doc pages. The sidebar would then be shown/hidden by
using the CSS display: none property. That way the main content would
always be to the right of the sidebar without having to manually set its
left offset.

jsha:
I took a look at how we handle sidebars on doc pages, and it's the same as you have it here, so please disregard my prior feedback.

GuillaumeGomez:
We could do something similar to what we did in docs.rs: https://github.com/rust-lang/docs.rs/pull/1464 (having the sidebar always visible with a button to expand it).

JohnCSimon:
ping again from triage:
@amaush
can you address the comments from the reviewer?

GuillaumeGomez:
https://github.com/rust-lang/rust/pull/89385 actually took over so I think it's best to close this PR.

Thanks a lot for starting this @amaush! Don't hesitate to give a try to another part of rustdoc if you have time. :wink:

Issue #82363, opened 2021-02-21
Tracking Issue for NonZero_c_* integers


KodrAus:
Feature gate: `#![feature(raw_os_nonzero)]`

This is a tracking issue for non-zero variants of `c_int` type aliases.

### Public API

```rust
pub type NonZero_c_char = NonZero?8;
pub type NonZero_c_schar = NonZeroI8;
pub type NonZero_c_uchar = NonZeroU8;
pub type NonZero_c_short = NonZeroI16;
pub type NonZero_c_ushort = NonZeroU16;
pub type NonZero_c_int = NonZeroI32;
pub type NonZero_c_uint = NonZeroU32;
#[cfg(any(target_pointer_width = "32", windows))]
pub type NonZero_c_long = NonZeroI32;
#[cfg(any(target_pointer_width = "32", windows))]
pub type NonZero_c_ulong = NonZeroU32;
#[cfg(all(target_pointer_width = "64", not(windows)))]
pub type NonZero_c_long = NonZeroI64;
#[cfg(all(target_pointer_width = "64", not(windows)))]
pub type NonZero_c_ulong = NonZeroU64;
pub type NonZero_c_longlong = NonZeroI64;
pub type NonZero_c_ulonglong = NonZeroU64;
```

### Steps / History

- [x] Implementation: #82228
- [ ] Final commenting period (FCP)
- [ ] Stabilization PR

### Unresolved Questions

- [ ] Naming.....

tesuji:
I think you mean *Usize or *Isize in NonZero_c_ulong type and friends.

kennytm:
IMO introducing `AtomicT`/`NonZeroT` replacing `Atomic<T>`/`NonZero<T>` are the gravest API mistake Rust has ever made so +1 to introducing `NonZero<T>` again.

The 12 `NonZeroT` types are introduced in rust-lang/rfcs#2307 was to sidestep the question what should happen to `NonZero<CustomType>`, taking the precedent of `AtomicT` (but disregarding `Wrapping<T>`).

> One problem of the current API is that it is unclear what happens or what *should* happen to `NonZero<T>` or `Option<NonZero<T>>` when `T` is some type other than a raw pointer or a primitive integer. In particular, crates outside of `std` can implement `Zeroable` for their arbitrary types since it is a public trait.
>
[...]

PR #82228, opened 2021-02-17
Provide NonZero_c_* integers

ijackson:
I'm pretty sure I am going want this for #73125 and it seems like an
omission that would be in any case good to remedy.

<strike>Because the raw C types are in `std`, not `core`, to achieve this we
must export the relevant macros from `core` so that `std` can use
them.  That's done with a new `num_internals` perma-unstable feature.

The macros need to take more parameters for the module to get the
types from and feature attributes to use.

I have eyeballed the docs output for core, to check that my changes to
these macros have made no difference to the core docs output.</strike>

PR #88303, opened 2021-08-24
Stabilise raw_os_nonzero (NonZero_c_* integers)

ijackson:
This stabilises `raw_os_nonzero`, `NonZero_c_int` etc.  Tracking issue #82363, FCP needed.

This was implemented in #82228 and merged in February.

There is one outstanding question, that of naming: should we be consistent with `c_int`, `c_long`, etc., or with `NonZeroU32` etc. ?  I chose the former and that's what's been in nightly since February.  There are arguments for the latter.  This is a bikeshed issue, but we do need to fix the colour of the paint before stabilisation.

### DIFF
--- a/library/std/src/os/raw/mod.rs
+++ b/library/std/src/os/raw/mod.rs
@@ -39,7 +39,7 @@ macro_rules! type_alias {
         type_alias_no_nz! { $Docfile, $Alias = $Real; $( $Cfg )* }
 
         #[doc = concat!("Type alias for `NonZero` version of [`", stringify!($Alias), "`]")]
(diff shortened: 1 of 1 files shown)


camelid:
triage: The FCP seems to have been cancelled.

ijackson:
> triage: The FCP seems to have been cancelled.

So it has, thanks for spotting that.  For some reason I wasn't subscribed to the FCP so I didn't notice.  I agree with the FCP discussion that generics are much better.  I have put this rework on my todo list (but as ever I may not get to it right away...)

PR #88305, opened 2021-08-24
Manual Debug for Unix ExitCode ExitStatus ExitStatusError

### LINKED ISSUE #74832: ExitStatus Debug instance is misleading
On UNIX, `c_int` is an alias for `i32` or `i16` depending on the platform. `ExitStatus` is a wrapper around `c_int`. However its Debug instance only exposes the last few bits:

test.c
```c
#include <stdio.h>
void main() {
    printf("%s\n", (void*)123);
}
```
running:
~~~~
$ gcc test.c && ./a.out
Segmentation fault
exit code 139
~~~~

test.rs
```rust
use std::process::Command;

#[derive(PartialEq, Eq, Clone, Copy, Debug)]
pub struct X(i16);
#[derive(PartialEq, Eq, Clone, Copy, Debug)]
pub struct XX(X);

pub fn main() {
  let e = Command::new("./a.out").status().unwrap();
  let i = XX(X(139));
  println!("exit with {:?}", e);
  println!("exit with {}", e);
  println!("test reproduce {:?}", i);
}
```
running:
~~~~
$ rustc test.rs && ./test
exit with ExitStatus(ExitStatus(11))
exit with signal: 11
test reproduce XX(X(139))
~~~~

The Debug output makes it look like a regular exit, not a signal exit. This is relevant because this is what rustbuild's rustc wrapper does in `src/bootstrap/bin/rustc.rs`, making rustc builds harder to debug.

As you can see I tried to recreate the weirdness with `XX` which as far as I can tell, follows the structure of how `ExitStatus` is defined. However I didn't recreate the platform-dependent logic in libstd/sys/mod.rs which also includes `missing_debug_implementations` which might be interfering with the derivation logic.

ijackson:
These structs have misleading names.  An ExitStatus[Error] is actually a Unix wait status; an ExitCode is actually an exit status.  These misleading names appear in the `Debug` output.

The `Display` impls on Unix have been improved, but the `Debug` impls are still misleading, as reported in #74832.

Fix this by pretending that these internal structs are called `unix_exit_status` and `unix_wait_status` as applicable.  (We can't actually rename the structs because of the way that the cross-platform machinery works: the names are cross-platform.)

After this change, this program
```
#![feature(exit_status_error)]
fn main(){
    let x = std::process::Command::new("false").status().unwrap();
    dbg!(x.exit_ok());
    eprintln!("x={:?}",x);
}
```
produces this output
```
[src/main.rs:4] x.exit_ok() = Err(
    ExitStatusError(
        unix_wait_status(
            256,
        ),
    ),
)
x=ExitStatus(unix_wait_status(256))
```

Closes #74832

### DIFF
--- a/library/std/src/sys/unix/process/process_common.rs
+++ b/library/std/src/sys/unix/process/process_common.rs
@@ -457,8 +457,14 @@ impl fmt::Debug for Command {
     }
 }
 
-#[derive(PartialEq, Eq, Clone, Copy, Debug)]
+#[derive(PartialEq, Eq, Clone, Copy)]
 pub struct ExitCode(u8);
+
+impl fmt::Debug for ExitCode {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        f.debug_tuple("unix_exit_status").field(&self.0).finish()
+    }
+}
 
 impl ExitCode {
     pub const SUCCESS: ExitCode = ExitCode(EXIT_SUCCESS as _);
--- a/library/std/src/sys/unix/process/process_unix.rs
+++ b/library/std/src/sys/unix/process/process_unix.rs
@@ -607,8 +607,14 @@ impl Process {
 }
 
 /// Unix exit statuses
-#[derive(PartialEq, Eq, Clone, Copy, Debug)]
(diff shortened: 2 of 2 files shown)


dtolnay (approved):
Looks good!

Issue #84908, opened 2021-05-04
Tracking Issue for ExitStatusError


ijackson:
Feature gate: `#![feature(exit_status_error)]`

This is a tracking issue for `ExitStatusError` (and `ExitStatus::exit_ok`).

This feature makes it convenient to properly check the exit status of subprocesses (such as from `Command`)

### Example

```rust
use std::process::Command;

let bad = Command::new("false").status().unwrap().exit_ok().unwrap_err();
assert_eq!(bad.code(), Some(1));
```
### Public API

(In pseudo-syntax:)

```rust
impl ExitStatus {
    fn exit_ok(self) -> Result<(), ExitStatusError> {..}
}

pub struct ExitStatusError(...); // newtype around a NonZero integer
impl Eq,Copy,Debug,Error,Display for ExitStatusError;
impl Into<ExitStatus> for ExitStatusError;
impl ExitStatusError  {
    fn code(&self) -> Option<i32> {..} } // WIFEXITED WEXITSTATUS
    fn code_nonzero(&self) -> Option<NonZeroi32> {..}
    ...
}
impl ExitStatusExt for ExitStatusError; // .is_signal() etc.
```

### Steps / History

- [ ] Implementation: #82973
- [ ] Final commenting period (FCP)
- [ ] Stabilization PR
- [ ] Add `#[must_use]` to `process::ExitStatus` (previous attempt #81452); see also #73131.

### Unresolved Questions

- None yet.

yaahc:
Can you also add a note about implementing `Try` for exit status to the Steps section?

related: https://github.com/rust-lang/rust/issues/54889

yaahc:
@scottmcm pointed out that we should be able to implement `Try` for `ExitStatus` now since we've officially landed the v2 `Try` trait on nightly, if you wanted to go ahead and implement that @ijackson.

taralx:
Late to the party, but the double unwrap seems like it should be avoidable. Is that in the plan?

ijackson:
> Late to the party, but the double unwrap seems like it should be avoidable. Is that in the plan?

I have a plan for it :-)

[...]

PR #88306, opened 2021-08-24
ErrorKind::ProcessFailed and impl From<ExitStatusError>

ijackson:
I think this impl makes process error handling a lot more ergonomic.  Look at the new example to see what I mean.

If we are going to do this, it makes more sense to do it now, while `ExitStatusError` is unstable.  We can experiment with how we like the results, without insta-stably providing the trait impl.

There is a possible concern related to the resulting error messages.  Because of the way `io::Error`'s `Display` impl simply forwards to the underlying error, the message `"subprocess failed"` won't be part of those errors.

But the `Display` impl on `ExitStatusError` prefixes the message from `ExitStatus` with `"process exited unsuccessfully"`, which is nicely unambiguous.

https://github.com/rust-lang/rust/issues/84908#issuecomment-864580297 has more discussion about this in the context of the other possible messages arising from process handling, in particular `Command::status()`.

### Other possibilities

This ability to represent a subprocess failure status as an `io::Error` will be useful in other situations:

This would make it possible to provide an `exit_ok` method on `Command`, improving the ergonomics of simple subprocess executions.

It would also make it much more possible to provide a version of `Command::output` that is less prone to the programmer accidentally losing the exit status.  It would even be possible to provide a "streaming" version of the output: a type which would encompass the `Child` and the pipe, and `impl io::Read`, and which would give a read error if the child failed.

Just the `ErrorKind` could be helpful for some other libraries that can otherwise use `io::Error` for their error handling.

**Note** this MR branch is on top of #88298 which has already been approved and I expect to land soon.

Let me see if I can get highfive to do what I want:
r? @yaahc

### DIFF
--- a/library/std/src/io/error.rs
+++ b/library/std/src/io/error.rs
@@ -289,7 +289,7 @@ pub enum ErrorKind {
     /// A subprocess (eg, run by a
     /// [`Command`](crate::process::Command)) failed.
     ///
-    /// Often wraps an
+    /// Often an [`io::Error`](`crate::io::Error`) with this kind wraps an
     /// [`ExitStatusError`](crate::process::ExitStatusError),
     /// (perhaps via `?` and `Into`), in which case
     /// [`io::Error::get_ref`](crate::io::Error::get_ref) or
--- a/library/std/src/io/error.rs
+++ b/library/std/src/io/error.rs
@@ -287,7 +287,7 @@ pub enum ErrorKind {
     /// Subprocess failed.
     ///
     /// A subprocess (eg, run by a
-    /// [`Command`](care::process::Command)) failed.
+    /// [`Command`](crate::process::Command)) failed.
     ///
     /// Often wraps an
     /// [`ExitStatusError`](crate::process::ExitStatusError),
--- a/library/std/src/io/error.rs
+++ b/library/std/src/io/error.rs
@@ -292,7 +292,7 @@ pub enum ErrorKind {
     /// Often wraps an
     /// [`ExitStatusError`](crate::process::ExitStatusError),
     /// (perhaps via `?` and `Into`), in which case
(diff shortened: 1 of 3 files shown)


library/std/src/io/error.rs
```diff
@@ -261,19 +261,15 @@ pub enum ErrorKind {
...
-    /// [`ErrorKind`].
-    ///
-    /// This [`ErrorKind`] is not used by the standard library.
+    /// This operation is unsupported on this platform.
     ///
-    /// Errors from the standard library that do not fall under any of the I/O
-    /// error kinds cannot be `match`ed on, and will only match a wildcard (`_`) pattern.
-    /// New [`ErrorKind`]s might be added in the future for some of those.
-    #[stable(feature = "rust1", since = "1.0.0")]
-    Other,
```
yaahc, line 275: Why'd you reorder these? I think this messes with the derived `Ord` impl which I'd prefer to avoid without strong justification.

yaahc (commented):
Thank you as always for the attention you're giving to improving the `Command` APIs. I see the value to providing more convenient ways to invoke `Command`s, and I think your goal with this PR is quite similar to @matklad's goal with https://github.com/rust-lang/rust/pull/89004, which I agree with. However, I really worry that making it possible to automatically shove an `ExitStatusError` into an `io::Error` will encourage bad error reporting habits.

I think @matklad probably has the right balance of convenience and reporting. If we're going to collapse the error types we should do so in a way that works as a reasonable default and include the extra context such as the command which was run.

camelid:
> `@yaahc` Thanks for pointing me at #89004. That does look like it's going in a good directiion. I will make some time to review it properly.

Should this PR be closed then?

JohnCSimon:
Ping from triage: 
@ijackson 

@rustbot label: +S-inactive

PR #83698, opened 2021-03-31
Use undef for uninitialized bytes in constants

erikdesjardins:
Fixes #83657

This generates good code when the const is fully uninit, e.g.

```rust
#[no_mangle]
pub const fn fully_uninit() -> MaybeUninit<[u8; 10]> {
    const M: MaybeUninit<[u8; 10]> = MaybeUninit::uninit();
    M
}
```
generates
```asm
fully_uninit:
	ret
```

as you would expect.

There is no improvement, however, when it's partially uninit, e.g.

```rust
pub struct PartiallyUninit {
    x: u64,
    y: MaybeUninit<[u8; 10]>
}

#[no_mangle]
pub const fn partially_uninit() -> PartiallyUninit {
    const X: PartiallyUninit = PartiallyUninit { x: 0xdeadbeefcafe, y: MaybeUninit::uninit() };
    X
}
```
generates
```asm
partially_uninit:
	mov	rax, rdi
	mov	rcx, qword ptr [rip + .L__unnamed_1+16]
	mov	qword ptr [rdi + 16], rcx
	movups	xmm0, xmmword ptr [rip + .L__unnamed_1]
	movups	xmmword ptr [rdi], xmm0
	ret

.L__unnamed_1:
	.asciz	"\376\312\357\276\255\336\000"
	.zero	16
	.size	.L__unnamed_1, 24
```
which copies a bunch of zeros in place of the undef bytes, the same as before this change.

Edit: generating partially-undef constants isn't viable at the moment anyways due to #84565, so it's disabled

PR #88308, opened 2021-08-24
Morph `layout_raw` query into `layout_of`.

eddyb:
Before this PR, `LayoutCx::layout_of` wrapped the `layout_raw` query, to:
* normalize the type, before attempting to compute the layout
* pass the layout to `record_layout_for_printing`, for `-Zprint-type-sizes`

Moving those two responsibilities into the query may reduce overhead (due to cached calls skipping those steps), but I want to do a perf run to know.

One of the changes I had to make was changing the return type of the query, to be able to both get out the type produced by normalizing inside the query *and* to match the signature of the old `TyCtxt::layout_of`. This change may be worse, perf-wise, so that's another reason I want to check.

r? @nagisa cc @oli-obk


the8472:
While the instruction counts look good this has [regressed the wall times](https://perf.rust-lang.org/compare.html?start=20997f6ad81721542e9ef97bb2f58190903a34d8&end=4b9f4b221b92193c7e95b1beb502c6eb32c3b613&stat=wall-time) of await-call-tree, deeply-nested-closures and deeply-nested-async by up to 30%

They're synthetic benchmarks so this might be acceptable.

nagisa:
Both instructions and _cycles_ are significantly down, which would mean that the perf machines were running at a low frequency for the test or that there are significantly more stalls or we're getting de-scheduled more often (due to e.g. holding a lock). Sadly we do not track context switches in perf :(

eddyb:
(**EDIT**: @nagisa posted a more helpful comment after I had written mine, below, but before posting it, keep that in mind)

~~Eugh this is probably that memory-latency-shaped gap in our performance tracking, I'm guessing the fewer instructions now have to wait for more cache misses or the like.~~ Though 30% is *a lot*.

Wait, but a lot of these are... `-check`? What is going on, that makes no sense, I'm only touching an API that is only used by codegen and miri (for CTFE).

Looking at [the query data for [`await-call-tree-check`/`incr-full`](https://perf.rust-lang.org/detailed-query.html?commit=4b9f4b221b92193c7e95b1beb502c6eb32c3b613&base_commit=20997f6ad81721542e9ef97bb2f58190903a34d8&benchmark=await-call-tree-check&run_name=incr-full), there are `70` `layout_of` executions, but they're seeming unrelated to the regression, which is in some other (arbitrary) part of the code.

Looking at the diff again, the only cascade effect I could think of would be the size of the `HashMap` holding the results of the query, and maybe that allocation affecting other allocations, but that kind of thing tends to lose itself in noise most of the time.

erikdesjardins:
I believe the wall-time regression was caused by the CPU governor on the perf machine, as mentioned [here](https://github.com/rust-lang/rust/pull/88956#issuecomment-922277439).

For more details see https://github.com/rust-lang/rust/pull/83698#issuecomment-931940573, a (big) comment on the PR that was merged (and perf-tested) immediately before this one, and thus shared in the blame for this regression.

PR #88310, opened 2021-08-25
Lock bootstrap (x.py) build directory

### LINKED ISSUE #76661: Running "x.py check" and "x.py test" in parallel leads to errors
Since I started using vscode, it is now often the case that the `./x.py check library/std` which vscode runs is still not done when I alt-tab back to my terminal to do (for example) `./x.py test --stage 0 library/core --test-args slice`. That regularly leads to errors like this:
```
error[E0464]: multiple matching crates for `core`
  |
  = note: candidates:
          crate `core`: /home/r/src/rust/rustc.2/build/x86_64-unknown-linux-gnu/stage0-sysroot/lib/rustlib/x86_64-unknown-linux-gnu/lib/libcore-deaf58ff3f911d75.rlib

error: could not compile `getrandom`.
```
In those cases, just running `x.py test` again solves the problem -- but it would still be better if having vscode run in background would not break what I do in the terminal.

Cc @Mark-Simulacrum

### LINKED ISSUE #80849: `x.py` sometimes causes spurious rebuilds of rustc
I tried this:

src/test/ui/fn/issue-80844-no-gate.rs
```diff
  fn returns_generator() -> _ {
  //~^ ERROR the type placeholder `_` is not allowed within types on item signatures [E0121]
  //~| NOTE not allowed in type signatures
      || yield 0
      //~^ ERROR yield syntax is experimental
-     //~^ NOTE see issue
+     //~| NOTE see issue
  }
```
```console
$ ./x.py test src/test/ui/fn
```

I expected to see this happen: the `ui/fn` test suite is run, possibly building the testsuite binary using the existing stage 1 `rustc`.

Instead, this happened: the stage 1 `rustc` is rebuilt before running the test suite.

This seems to happen randomly; I had rerun the test suite quite a few times using the same command with no issues. This is rather inconvenient since even with 16 logical cores, building `rustc` takes several minutes and grinds my machine to a halt.

This doesn't just seem to happen when running the test suite, but I'm less confident about saying that rebuilds after modifying rustc's codebase are spurious.

@rustbot modify labels +A-contributor-roadblock +A-rustbuild

worldeva:
Closes #76661,  closes #80849, 
`x.py` creates a lock file at `project_root/lock.db`

r? @jyn514 , because he was one that told me about this~

### DIFF
--- a/src/bootstrap/bootstrap.py
+++ b/src/bootstrap/bootstrap.py
@@ -15,6 +15,42 @@ import tempfile
 
 from time import time
 
+# Acquire a lock on the build directory to make sure that
+# we don't cause a race condition while building
+# Lock is created in `build_dir/lock.db`
+def acquire_lock(build_dir):
+    try:
+        import sqlite3
+
+        path = os.path.join(build_dir, "lock.db")
+        try:
+            con = sqlite3.Connection(path, timeout=0)
+            curs = con.cursor()
+            curs.execute("BEGIN EXCLUSIVE")
+            # The lock is released when the cursor is dropped
+            return curs
+        # If the database is busy then lock has already been acquired
+        # so we wait for the lock.
+        # We retry every quarter second so that execution is passed back to python
+        # so that it can handle signals
+        except sqlite3.OperationalError:
+            del con
+            del curs
+            print("Waiting for lock on build directory")
+            con = sqlite3.Connection(path, timeout=0.25)
+            curs = con.cursor()
+            while True:
+                try:
+                    curs.execute("BEGIN EXCLUSIVE")
+                except sqlite3.OperationalError:
+                    pass
+            return curs
+    except ImportError:
+        print("warning: sqlite3 not available in python, skipping build directory lock")
+        print("please file an issue on rust-lang/rust")
+        print("this is not a problem for non-concurrent x.py invocations")
+        return None
+
 def support_xz():
     try:
         with tempfile.NamedTemporaryFile(delete=False) as temp_file:
@@ -1225,6 +1261,12 @@ def bootstrap(help_triggered):
     build.set_dist_environment(data["dist_server"])
 
     build.build = args.build or build.build_triple()
+
+    # Acquire the lock before doing any build actions
+    # The lock is released when `lock` is dropped
+    if not os.path.exists(build.build_dir):
+        os.makedirs(build.build_dir)
+    lock = acquire_lock(build.build_dir)
     build.update_submodules()
 
     # Fetch/build the bootstrap


jyn514:
Happy to be responsible for this, but @Mark-Simulacrum will want to take a look too I'm sure.

src/bootstrap/flock.rs
```diff
@@ -0,0 +1,206 @@
+//! Handles creating a file lock to avoid race conditions[1].
```
jyn514, line 1: This is now the third flock implementation in-tree, there's also one for rustdoc ... maybe we should separate this out into a tiny crate or something? It would be hard to get cargo to use it since it's a submodule, but we could at least share it with rustdoc.

cc @rust-lang/cargo, do you have opinions on this? Would you be willing to separate the flock implementation into a different crate, and maybe put it on rust-lang/flock? I don't think there's any need to publish to crates.io but it would be nice to share at least between different tools.
cc @kinnison @rbtcollins, I'm sure rustup has its own implementation of flock too.

jyn514 (commented):
I found that this doesn't lock in bootstrap.py, but maybe that's ok - certainly I haven't seen any bugs reported about it.

```
// terminal 1
$ x check
downloading https://ci-artifacts.rust-lang.org/rustc-builds/db002a06ae9154a35d410550bc5132df883d7baa/rust-dev-nightly-x86_64-unknown-linux-gnu.tar.xz
############################################################################### 100.0%
extracting /home/joshua/src/rust/rust/build/cache/llvm-db002a06ae9154a35d410550bc5132df883d7baa-False/rust-dev-nightly-x86_64-unknown-linux-gnu.tar.xz

// terminal 2
$ x build
Submodules updated in 0.02 seconds
downloading https://ci-artifacts.rust-lang.org/rustc-builds/db002a06ae9154a35d410550bc5132df883d7baa/rust-dev-nightly-x86_64-unknown-linux-gnu.tar.xz
###                                                                               4.9%^C
```

I don't feel qualified to review the flock code, but I also think I may not have to - I left a comment below.

I'm very happy with how this fixes the parallel builds problem :) thank you for tackling it!

src/bootstrap/flock.rs
rbtcollins: https://github.com/rust-lang/rustup/issues/988#issuecomment-582314389 and the top level summary are probably relevant.

tl;dr: don't want the cargo implementation, it is buggy; but also haven't gotten around to fixing the issue in rustup, which is complected with https://github.com/rust-lang/rustup/issues/2441 making this a little gnarly to think through.

jyn514: Ok. x.py itself probably doesn't need an airtight flock implementation, I wouldn't expect people to be using NFS for builds. I totally get that rustup does have to deal with those scenarios though.

Does rustup's flock implementation have any downsides compared to cargo's? Could we start from that as the shared codebase instead?

Mark-Simulacrum:
So... this might be a little "out there", but it feels simpler than the solution this suggests and avoids duplicating code from Cargo. Both Python 2.7.x and Python 3 natively have sqlite built-in, as far as I can tell, via the sqlite3 module.

That means that we could likely implement this by "just" leveraging sqlite transactions to block. This is sort of overkill, but I'd expect it to work pretty well across any platform or filesystem, since sqlite itself is pretty battle tested.

It's also true that the python script currently manages a bit of state via pretty adhoc files (e.g., for the download stamps). I wouldn't change those today, there's not much point, but in the future it might make sense to leverage a more structured format that sqlite trivially gives us for this storage and without creating a bunch of files all over the place.

I'm curious to hear what folks think of this idea. I haven't actually tried a sample implementation, either, but I'd naively expect it to work out to at most a couple dozen lines of Python -- not too complex. Since sqlite comes pre-compiled for us through python, this seems a nice solution that avoids some of the hassle we'd otherwise run into. If and/or when we migrate away from the python wrapper, we can explore alternatives (such as flock proposed by this PR), but for now this would seem better. It also avoids any parallelism issues in the python script's downloads and such, which also seems good -- I'm not sold those are bug free today.

I do have mixed feelings about this so I could be convinced the impl in this PR is something we should just go with, but if the sqlite-based solution actually works and isn't too hard, then it seems preferable to me personally on the grounds of not having to maintain pretty low-level platform-specific code in bootstrap -- which is a place I'd really prefer to avoid that.

jyn514:
Hi @worldeva - do you know when you'll get a chance to look at this again?

src/bootstrap/flock.rs
thomcc: There's also fd-lock, which worked relatively well for me in the past (admittedly it was a small project). https://crates.io/crates/fd-lock/

That said, perhaps you need more control than it gives.

worldeva:
Sorry for the long delay for the tiny commit; life came up >,<
As per suggestion, I've removed all the stuff in bootstrap-proper and moved it over to a `sqlite3` based bit in `x.py`.

src/bootstrap/bootstrap.py
```diff
@@ -1254,6 +1270,7 @@ def bootstrap(help_triggered):
 def main():
     """Entry point for the bootstrap process"""
     start_time = time()
+    lock = acquire_lock()
```
jyn514, line 1273: I think this is slightly earlier than necessary; it would be nice to print the help message before spending time acquiring the lock. I think a good spot is just before `build.update_submodules()` in `bootstrap()`.

src/bootstrap/bootstrap.py
```diff
@@ -12,9 +12,25 @@
...
+        curs = con.cursor()
+        curs.execute("BEGIN EXCLUSIVE")
+        return curs
+    except sqlite3.OperationalError:
+        del con
+        del curs
+        print("Waiting for lock on database")
+        con = sqlite3.Connection("xlock.db", timeout=9999)
+        curs = con.cursor()
+        curs.execute("BEGIN EXCLUSIVE")
```
jyn514, line 31: Does this need to do some sort of cleanup to release the lock? Or is exiting the process enough?

worldeva: Dropping the cursor should automatically end the transaction.

src/bootstrap/bootstrap.py
worldeva: Moved!~

jyn514 (approved):
This looks good to me with the tidy lint fixed :)

src/bootstrap/bootstrap.py
```diff
@@ -12,9 +12,25 @@
 import sys
 import tarfile
 import tempfile
+import sqlite3
 
 from time import time
 
+def acquire_lock():
```
Mark-Simulacrum, line 19: It'd be great to put a comment on this function, explaining why we're using sqlite to implement this, and the goal of the lock itself.

src/bootstrap/bootstrap.py
```diff
@@ -12,9 +12,25 @@
...
+    try:
+        con = sqlite3.Connection("xlock.db", timeout=0)
+        curs = con.cursor()
+        curs.execute("BEGIN EXCLUSIVE")
+        return curs
+    except sqlite3.OperationalError:
+        del con
+        del curs
+        print("Waiting for lock on database")
```
Mark-Simulacrum, line 28: I think database here is an internal implementation detail -- I would suggest a message closer to Cargo's:

>  Blocking waiting for file lock on the registry index

Maybe: "Blocking waiting for build directory lock".

src/bootstrap/bootstrap.py
Mark-Simulacrum: Have we verified what happens when x.py is killed (kill -9, or similar)? I would expect that the lock is just indefinitely held?

src/bootstrap/bootstrap.py
```diff
@@ -12,9 +12,25 @@
...
+        con = sqlite3.Connection("xlock.db", timeout=9999)
```
Mark-Simulacrum, line 29: sqlite3 suggests the timeout is in milliseconds -- which means this would be just 10 seconds, quite low compared to typical rustc builds. Do we have some evidence for it being in seconds or similar? Otherwise, I would suggest passing a larger value here -- probably aiming at ~1 hour or so, though it's not clear to me why we *need* a timeout -- don't we want to block indefinitely? (Can we specify that?)

src/bootstrap/bootstrap.py
```diff
@@ -12,9 +12,25 @@
...
 
 from time import time
 
+def acquire_lock():
+    try:
+        con = sqlite3.Connection("xlock.db", timeout=0)
+        curs = con.cursor()
+        curs.execute("BEGIN EXCLUSIVE")
+        return curs
+    except sqlite3.OperationalError:
```
Mark-Simulacrum, line 25: Can we document in a comment what results in this error? I presume it's a failure to acquire the lock within the (nonexistent) timeout?

src/bootstrap/bootstrap.py
```diff
@@ -12,9 +12,25 @@
 import sys
 import tarfile
 import tempfile
+import sqlite3
 
 from time import time
 
+def acquire_lock():
+    try:
+        con = sqlite3.Connection("xlock.db", timeout=0)
```
Mark-Simulacrum, line 21: Is this creating the database in the working directory? I think we probably want it in build/ -- build.build_dir probably?

Mark-Simulacrum:
Mostly just some documentation, but I expect we can move ahead once that's added. It'd be good to do some spot checking locally to confirm this mostly works, and make a note in the documentation (and PR description) the location of the database file, so if there's some bug with this it's easy to delete it.

src/bootstrap/bootstrap.py
worldeva: Unfortunately, in my testing, `float('inf')` doesn't work (just instantly times out), so it has to be some smaller value.
However, the sqlite3 lib for python does timeout in seconds ("The default for the timeout parameter is 5.0 (five seconds).")

src/bootstrap/bootstrap.py
worldeva: I changed it to creating it in `build.rust_dir`: The lock locks a lot more than just the stuff in `build` (submodules, etc), and the folder may not exist at the time of lock creation.

src/bootstrap/bootstrap.py
```diff
@@ -12,9 +12,34 @@
...
 import tempfile
+import sqlite3
 
 from time import time
 
+# Acquire a lock on the build directory to make sure that
+# we don't cause a race condition while building
+# Lock is created in `project_root/lock.db`
+def acquire_lock(root_dir):
+    path = os.path.join(root_dir, "lock.db")
```
Mark-Simulacrum, line 23: I'm not sure creating lock.db in the root is OK -- it feels a little messy. I think we should be able to create the build directory earlier, if it's not created yet, and place lock.db inside it.

worldeva:
Alright!~
Testing shows that the lock is closed even when `SIGTERM`'d: so we don't have to worry about that :D
One thing I am worried about is that it won't exit from Ctrl-C while blocking (sqlite3 seems to catch `KeyboardInterrupt`?)

Mark-Simulacrum:
I think it's worth trying to investigate that more, seems like a bad thing -- maybe we can workaround that some how?

worldeva:
Alright, fixed: it seems like sqlite3 was blocking on the C side, meaning that python could never handle any signals.

Mark-Simulacrum:
@bors r+ 

Squashed the commits, let's land this; if there's problems we can always back this out.

matthiaskrgr:
@bors r-

https://github.com/rust-lang/rust/pull/91664#issuecomment-988892478
Failed in a rollup.

src/bootstrap/bootstrap.py
```diff
@@ -12,9 +12,38 @@
 import sys
 import tarfile
 import tempfile
+import sqlite3
```
matthiaskrgr, line 16: Maybe the sqlite3 python package is not available on all CI build runners...?

Mark-Simulacrum: I think this particular feature would be fine to skip running if we don't have sqlite3 available. My hope would be that it's mostly available in practice -- so we should still have good support on practical usage, where we might expect parallel invocations.

worldeva:
Should be fixed!~

src/bootstrap/bootstrap.py
```diff
@@ -15,6 +15,40 @@
...
+            con = sqlite3.Connection(path, timeout=0.25)
+            curs = con.cursor()
+            while True:
+                try:
+                    curs.execute("BEGIN EXCLUSIVE")
+                except sqlite3.OperationalError:
+                    pass
+            return curs
+    except ImportError:
+        print("warning: sqlite3 not available, skipping lock")
```
Mark-Simulacrum, line 49: I think it's worth asking users to file an issue if they see this (on rust-lang/rust) and making it clear that the lock isn't necessary for correctness of the results (something like "build directory lock", maybe)?

worldeva:
Done-
The one thing that I'm mildly worried about is `"consider filing a issue"` makes it seem more important than it is- should I perhaps make it even more explicit with a "this lock is only to prevent multiple concurrent bootstraps from running" or such?
(...Though we're straying into the bikeshedding territory now :p )

Mark-Simulacrum:
Slightly adjusted wording, but seems good to me otherwise. Thanks!

@bors r+

PR #88312, opened 2021-08-25
Treat types in unnormalized function signatures as well-formed

### LINKED ISSUE #87748: GAT: lifetime mismatch when requiring lifetime bounds despite type not using the lifetimes at all
Compiling this code:

```rust

trait MyTrait {
    type Assoc<'a, 'b> where 'b: 'a;
    fn do_sth(arg: Self::Assoc<'_, '_>);
}

struct Foo;

impl MyTrait for Foo {
    type Assoc<'a, 'b> where 'b: 'a = u32;
    
    fn do_sth(_: u32) {}
    // fn do_sth(_: Self::Assoc<'static, 'static>) {}
    // fn do_sth(_: Self::Assoc<'_, '_>) {}
}
```

I'd expect all of these three versions to compile. The associated type has two lifetimes as well as a lifetime bound `'b: 'a`, but `u32` doesn't use the lifetime. However, I'm getting a compiler error:

```
error[E0478]: lifetime bound not satisfied
  --> src/lib.rs:13:5
   |
13 |     fn do_sth(_: u32) {}
   |     ^^^^^^^^^^^^^^^^^
   |
note: lifetime parameter instantiated with the anonymous lifetime #2 defined on the method body at 13:5
  --> src/lib.rs:13:5
   |
13 |     fn do_sth(_: u32) {}
   |     ^^^^^^^^^^^^^^^^^
note: but lifetime parameter must outlive the anonymous lifetime #1 defined on the method body at 13:5
  --> src/lib.rs:13:5
   |
13 |     fn do_sth(_: u32) {}
   |     ^^^^^^^^^^^^^^^^^

For more information about this error, try `rustc --explain E0478`.
```

Without the `where 'b: 'a`, all three compile fine. 

[Playground Link](https://play.rust-lang.org/?version=nightly&mode=debug&edition=2018&gist=ec5f5d95af1164b121b62cb130d2dced)

jackh726:
Fixes #87748

r? @nikomatsakis

### DIFF
--- a/compiler/rustc_mir/src/borrow_check/type_check/free_region_relations.rs
+++ b/compiler/rustc_mir/src/borrow_check/type_check/free_region_relations.rs
@@ -256,6 +256,9 @@ impl UniversalRegionRelationsBuilder<'cx, 'tcx> {
         let constraint_sets: Vec<_> = unnormalized_input_output_tys
             .flat_map(|ty| {
                 debug!("build: input_or_output={:?}", ty);
+                // We add implied bounds from both the unnormalized and normalized ty
+                // See issue #87748
+                let constraints_implied_1 = self.add_implied_bounds(ty);
                 let TypeOpOutput { output: ty, constraints: constraints1, .. } = self
                     .param_env
                     .and(type_op::normalize::Normalize::new(ty))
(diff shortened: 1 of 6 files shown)


nikomatsakis (approved):

rylev:
@nikomatsakis @jackh726 This change lead to large regressions in [instruction counts](https://perf.rust-lang.org/compare.html?start=66acdee9f7822ff2427f2b967e537215421eeb16&end=59ce76548484806ac4970c57c0bb6ad9e53b80f6&stat=instructions:u)  in many real world crates. The query affected by this seems to be additional calls to `implied_outlives_bounds` which I believe would be impacted by this change. Any thoughts on why this might be? 

As part of the performance triage process, I'm marking this as a performance regression. If you believe this performance regression to be justifiable or once you have an issue or PR that addresses this regression, please mark this PR with the label perf-regression-triaged.

@rustbot label +perf-regression

nikomatsakis:
This change certainly added more implied bounds, so there may just be more to sort through, but there may also be some oversights, like failure to remove duplicates. Merits more investigation for sure.

jackh726:
I'll do some experiments to see if I can fix the perf regressions.

jackh726:
We won back perf in #88771. It's not 100% of what we lost, but about the best that can be done probably.

PR #88316, opened 2021-08-25
Remove most box syntax uses from the testsuite except for src/test/ui/issues 

est31:
Removes most box syntax uses from the testsuite outside of the src/test/ui/issues directory. The goal was to only change tests where box syntax is an implementation detail instead of the actual feature being tested. So some tests were left out, like the regression test for #87935, or tests where the obtained error message changed significantly.

Mostly this replaces box syntax with `Box::new`, but there are some minor drive by improvements, like formatting improvements or `assert_eq` instead of `assert!( == )`.

Prior PR that removed box syntax from the compiler and tools: #87781

### DIFF
--- a/src/test/ui/pure-sum.rs
+++ b/src/test/ui/pure-sum.rs
@@ -5,8 +5,6 @@
 
 // pretty-expanded FIXME #23616
 
-#![feature(box_syntax)]
-
 fn sums_to(v: Vec<isize> , sum: isize) -> bool {
     let mut i = 0;
     let mut sum0 = 0;
@@ -19,7 +17,7 @@ fn sums_to(v: Vec<isize> , sum: isize) -> bool {
 
 fn sums_to_using_uniq(v: Vec<isize> , sum: isize) -> bool {
     let mut i = 0;
-    let mut sum0: Box<_> = box 0;
+    let mut sum0: Box<_> = 0.into();
     while i < v.len() {
         *sum0 += v[i];
         i += 1;
@@ -41,7 +39,7 @@ struct F<T> { f: T }
 
 fn sums_to_using_uniq_rec(v: Vec<isize> , sum: isize) -> bool {
     let mut i = 0;
-    let mut sum0 = F::<Box<_>> {f: box 0};
+    let mut sum0 = F::<Box<_>> {f: 0.into() };
     while i < v.len() {
         *sum0.f += v[i];
         i += 1;
--- a/src/test/ui/unique/unique-containing-tag.rs
+++ b/src/test/ui/unique/unique-containing-tag.rs
@@ -4,12 +4,10 @@
 
 // pretty-expanded FIXME #23616
 
-#![feature(box_syntax)]
-
 pub fn main() {
     enum t { t1(isize), t2(isize), }
 
-    let _x: Box<_> = box t::t1(10);
+    let _x: Box<_> = Box::new(t::t1(10));
 
     /*alt *x {
       t1(a) {
@@ -19,9 +17,9 @@ pub fn main() {
     }*/
 
     /*alt x {
-      box t1(a) {
+      Box::new(t1(a) {
         assert_eq!(a, 10);
-      }
+      })
       _ { panic!(); }
     }*/
 }
--- a/src/test/debuginfo/borrowed-struct.rs
+++ b/src/test/debuginfo/borrowed-struct.rs
@@ -63,7 +63,6 @@
 // lldbr-check:(f64) *unique_val_interior_ref_2 = 26.5
 
 #![allow(unused_variables)]
-#![feature(box_syntax)]
 #![feature(omit_gdb_pretty_printer_section)]
 #![omit_gdb_pretty_printer_section]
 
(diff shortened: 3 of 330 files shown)


jyn514:
I have the same comment as https://github.com/rust-lang/rust/pull/87781#issuecomment-905426207 - This seems like a really bad idea. How will you possibly ensure that it's testing the same thing that it did before?

est31:
@Mark-Simulacrum so do you want this PR or do you not want it?

Mark-Simulacrum:
It's in my review queue, I just haven't had a chance yet to take a look at your latest comments and provide feedback.

est31:
@Mark-Simulacrum okay, thanks for the feedback, take your time! :+1:

Mark-Simulacrum:
OK, I've thought some more about this and looked at the first couple commits.

I don't personally have time to review this PR for each test. On the other hand, it seems pretty unlikely we'll ever be able to really be sure (even if I or someone had that review time). In that sense, maybe we should just accept this and hope that we're not losing any valuable tests. It seems pretty unlikely that we'd have too much dependence on box syntax in the majority of tests.

I think we should still try to decrease the amount of churn by squashing everything into one commit (easier to filter out if needed and doesn't really make review any harder), and doing some work on avoiding line churn. But that's mostly just decreasing the amount of interference with git blame.

I think we should also drop anything that's not just line churn into a separate PR -- that should be reviewed more closely, and hopefully only a very small number of tests change in that way.

est31:
@Mark-Simulacrum sounds like a plan. I'll squash the PR. Should I try to reduce the line number churn as offered or can it be kept?

Mark-Simulacrum:
I think the easier cases -- e.g., no need to add comments or other weirdness, just shuffling some lines or adding a blank one -- we should, since that'll make the impact of the PR on blame and such lower, but otherwise no need to invest too much time in the complicated cases.

est31:
I've changed the PR to reduce the line number spam by adding newlines in various places, and also squashed the PR to a single commit. I've also removed the changes of the `src/test/ui/reachable/expr_box.rs` test because it seemed too box syntax specific to me.

est31:
re-r? @Mark-Simulacrum

est31:
Interesting, commit 79adda930f9b607ecb4819ed7abcf1cd285e938a has caused some test file churn. So the tests aren't write only after all.

src/test/ui/pure-sum.rs
```diff
@@ -19,7 +17,7 @@ fn sums_to(v: Vec<isize> , sum: isize) -> bool {
 
 fn sums_to_using_uniq(v: Vec<isize> , sum: isize) -> bool {
     let mut i = 0;
-    let mut sum0: Box<_> = box 0;
+    let mut sum0: Box<_> = 0.into();
```
Mark-Simulacrum, line 20: Why are sum of the tests switching to `.into()` and not `Box::new`?

Mark-Simulacrum:
> Interesting, commit [79adda9](https://github.com/rust-lang/rust/commit/79adda930f9b607ecb4819ed7abcf1cd285e938a) has caused some test file churn. So the tests aren't write once after all.

Not sure I follow -- the tests *do* get updated, of course, but my experience is that generally that update is rote -- it doesn't require comprehension / reading the test for most deltas, such as the one you point at.

src/test/ui/pure-sum.rs
est31: In this instance it's clear due to the `: Box<_>`. At the start I have sometimes added into because it's shorter, but later on I went on to using Box::new for most things.

est31: @Mark-Simulacrum should I change the into invocations to Box::new?

Mark-Simulacrum: I would've preferred that but I don't think it's really that necessary, and I don't want you to redo all that work another time.

Mark-Simulacrum:
r=me with a rebase

est31:
@Mark-Simulacrum it's rebased. I don't have r= rights so you gotta do it.

Mark-Simulacrum:
Thanks!

@bors r+

est31:
I've ran `x.py test` but hadn't done `--compare-mode nll`. Now I've done it.

est31:
I don't own a mac so I have no way of verifying that this fixed the issue, and as it seems bors try is no help for this either, so asking for another r+.

est31:
Hmm seems I've been overzealous in replacing `box foo` with `Box::new(foo)` and replaced a pattern usage, which caused compilation to fail, and I didn't notice it because i couldn't test it. I manually copied the example to the playground and it compiles now but I still can't test it.

src/test/ui/unique/unique-containing-tag.rs
```diff
@@ -19,9 +17,9 @@ pub fn main() {
...
 
     /*alt x {
-      box t1(a) {
+      Box::new(t1(a) {
         assert_eq!(a, 10);
-      }
+      })
       _ { panic!(); }
     }*/
 }
```
bjorn3, line 25: The comments in this test contain ancient match syntax.

est31: Yeah the PR touched a bunch of extremely ancient tests, observable on things like different code styles or such.

PR #88327, opened 2021-08-25
`WrappingRange` (#88242) follow-up (`is_full_for`, `Scalar: Copy`, etc.)

bonega:
Some changes related to feedback during #88242
r? @RalfJung

### DIFF
--- a/compiler/rustc_target/src/abi/mod.rs
+++ b/compiler/rustc_target/src/abi/mod.rs
@@ -741,7 +741,7 @@ impl Primitive {
 ///
 /// This is intended specifically to mirror LLVM’s `!range` metadata,
 /// semantics.
-#[derive(Clone, PartialEq, Eq, Hash)]
+#[derive(Clone, Copy, PartialEq, Eq, Hash)]
 #[derive(HashStable_Generic)]
 pub struct WrappingRange {
     pub start: u128,
@@ -797,7 +797,7 @@ impl fmt::Debug for WrappingRange {
 }
 
 /// Information about one scalar component of a Rust type.
-#[derive(Clone, PartialEq, Eq, Hash, Debug)]
+#[derive(Clone, Copy, PartialEq, Eq, Hash, Debug)]
 #[derive(HashStable_Generic)]
 pub struct Scalar {
     pub value: Primitive,
@@ -1070,7 +1070,7 @@ impl Niche {
     }
 
     pub fn available<C: HasDataLayout>(&self, cx: &C) -> u128 {
-        let Scalar { value, valid_range: ref v } = self.scalar;
+        let Scalar { value, valid_range: v } = self.scalar;
         let bits = value.size(cx).bits();
         assert!(bits <= 128);
         let max_value = !0u128 >> (128 - bits);
@@ -1083,7 +1083,7 @@ impl Niche {
     pub fn reserve<C: HasDataLayout>(&self, cx: &C, count: u128) -> Option<(u128, Scalar)> {
         assert!(count > 0);
 
-        let Scalar { value, valid_range: v } = self.scalar.clone();
+        let Scalar { value, valid_range: v } = self.scalar;
         let bits = value.size(cx).bits();
         assert!(bits <= 128);
         let max_value = !0u128 >> (128 - bits);
@@ -1137,7 +1137,7 @@ pub struct Layout {
 
 impl Layout {
     pub fn scalar<C: HasDataLayout>(cx: &C, scalar: Scalar) -> Self {
-        let largest_niche = Niche::from_scalar(cx, Size::ZERO, scalar.clone());
+        let largest_niche = Niche::from_scalar(cx, Size::ZERO, scalar);
         let size = scalar.value.size(cx);
         let align = scalar.value.align(cx);
         Layout {
--- a/compiler/rustc_target/src/abi/mod.rs
+++ b/compiler/rustc_target/src/abi/mod.rs
@@ -792,7 +792,7 @@ impl WrappingRange {
     pub fn is_full_for(&self, size: Size) -> bool {
         let max_value = size.unsigned_int_max();
         debug_assert!(self.start <= max_value && self.end <= max_value);
-        (self.start == 0 && self.end == max_value) || (self.end.wrapping_add(1) == self.start)
+        self.start == (self.end.wrapping_add(1) & max_value)
     }
 }
 
--- a/compiler/rustc_target/src/abi/mod.rs
+++ b/compiler/rustc_target/src/abi/mod.rs
@@ -955,7 +955,7 @@ impl AddressSpace {
 
 /// Describes how values of the type are passed by target ABIs,
 /// in terms of categories of C types there are ABI rules for.
-#[derive(Clone, PartialEq, Eq, Hash, Debug, HashStable_Generic)]
+#[derive(Clone, Copy, PartialEq, Eq, Hash, Debug, HashStable_Generic)]
 pub enum Abi {
     Uninhabited,
     Scalar(Scalar),
@@ -983,8 +983,8 @@ impl Abi {
     /// Returns `true` if this is a single signed integer scalar
     #[inline]
     pub fn is_signed(&self) -> bool {
-        match *self {
-            Abi::Scalar(ref scal) => match scal.value {
+        match self {
+            Abi::Scalar(scal) => match scal.value {
                 Primitive::Int(_, signed) => signed,
                 _ => false,
             },
@@ -1053,7 +1053,7 @@ pub enum TagEncoding {
     },
 }
 
-#[derive(Clone, PartialEq, Eq, Hash, Debug, HashStable_Generic)]
+#[derive(Clone, Copy, PartialEq, Eq, Hash, Debug, HashStable_Generic)]
 pub struct Niche {
     pub offset: Size,
     pub scalar: Scalar,
@@ -1259,7 +1259,7 @@ impl<'a, Ty> TyAndLayout<'a, Ty> {
         Ty: TyAbiInterface<'a, C>,
         C: HasDataLayout,
     {
-        let scalar_allows_raw_init = move |s: &Scalar| -> bool {
+        let scalar_allows_raw_init = move |s: Scalar| -> bool {
             if zero {
                 // The range must contain 0.
                 s.valid_range.contains(0)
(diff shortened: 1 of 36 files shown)


bonega:
Reason for not having method in `WrappingRange` is because I think it makes it less general.
I don't feel strongly about it though.

compiler/rustc_target/src/abi/mod.rs
```diff
@@ -753,21 +768,11 @@ impl Scalar {
...
-        let start = self.valid_range.start;
-        let end = self.valid_range.end;
-        assert_eq!(start, start & mask);
-        assert_eq!(end, end & mask);
-        start..(end.wrapping_add(1) & mask)
+    /// Returns `true` if `size` completely fills the range.
+    #[inline]
+    pub fn is_filled_by(&self, size: Size) -> bool {
+        (self.valid_range.start == 0 && self.valid_range.end == size.unsigned_max())
+            || (self.valid_range.end + 1 == self.valid_range.start)
```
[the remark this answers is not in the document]
bonega: I have seen a lot of `assert!(bits <=128)` or something like it.
Are they meant to be debug asserts instead?

For example in `Niche::available`.

bonega:
Thanks for the helpful reviews

bonega:
Just did a commit that makes `abi::Abi` `Copy` and removed a lot of refs for `Scalar` and `Abi`.
I don't really understand the big picture yet, please help by yelling at me.
Performance changes are up to the compiler I think?

bonega:
Will be interesting to see if this fails because of performance or other reasons

bonega:
Could I get a performance run for this one?

compiler/rustc_target/src/abi/mod.rs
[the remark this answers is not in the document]
RalfJung: > Are they meant to be debug asserts instead?

If they already exist as assert, probably best to keep them like that.

RalfJung:
Btw, you can probably avoid most of these CI failures that you got by running these commands before pushing
```
./x.py check compiler/rustc
./x.py test tidy --bless
```
They should not take very long (a few minutes at most).

bonega:
Thanks for your tip, it will improve my life :+1: 
I don't have anything more to add at this time

compiler/rustc_target/src/abi/mod.rs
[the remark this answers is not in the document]
bonega: Probably, I just want to understand more

compiler/rustc_mir/src/interpret/validity.rs
```diff
@@ -620,40 +620,38 @@ impl<'rt, 'mir, 'tcx: 'mir, M: Machine<'mir, 'tcx>> ValidityVisitor<'rt, 'mir, '
...
-        assert!(hi <= max_hi);
-        // We could also write `(hi + 1) % (max_hi + 1) == lo` but `max_hi + 1` overflows for `u128`
-        if (lo == 0 && hi == max_hi) || (hi + 1 == lo) {
+        if scalar_layout.valid_range.is_full_for(op.layout.size) {
             // Nothing to check
             return Ok(());
         }
-        // At least one value is excluded. Get the bits.
+        // At least one value is excluded.
+        let valid_range = scalar_layout.valid_range.clone();
```
RalfJung, line 630: This seems to still unnecessarily call clone?

compiler/rustc_target/src/abi/mod.rs
```diff
@@ -389,6 +389,21 @@ impl Size {
...
+
+    #[inline]
+    pub fn signed_max(&self) -> i128 {
+        i128::MAX >> (128 - self.bits())
+    }
+
+    #[inline]
+    pub fn unsigned_max(&self) -> u128 {
+        u128::MAX >> (128 - self.bits())
+    }
```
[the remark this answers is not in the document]
RalfJung: Well, we already have `sign_extend` and `truncate` on `Size`... and I think it would be useful to have these min/max methods somewhere as this is annoying bitfiddling.

`Integer` would be annoying unless there is an easy way to go from a `Size` to an integer. We will typically have the size of the data we are working on available as a `Size`.

compiler/rustc_mir/src/interpret/validity.rs
```diff
@@ -620,40 +620,38 @@ impl<'rt, 'mir, 'tcx: 'mir, M: Machine<'mir, 'tcx>> ValidityVisitor<'rt, 'mir, '
...
+        let WrappingRange { start, end } = valid_range;
+        let max_value = u128::MAX >> (128 - op.layout.size.bits());
```
RalfJung, line 632: Hm, a bit annoying that we compute this in two places now... but let's discuss that in https://github.com/rust-lang/rust/pull/88327#discussion_r697375647.

compiler/rustc_target/src/abi/mod.rs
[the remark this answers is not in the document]
bonega: Since lack of good alternatives(at least discussed), I moved them back into `Size` again.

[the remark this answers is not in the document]
eddyb: In that case, can they say `(un)signed_int` instead of just `(un)signed`? (or `uint` and `sint`)

I still don't like them on `Size` but at least I'm less opposed to them that way, I guess.

compiler/rustc_target/src/abi/mod.rs
```diff
@@ -803,25 +822,16 @@ pub struct Scalar {
...
-        let start = self.valid_range.start;
-        let end = self.valid_range.end;
-        assert_eq!(start, start & mask);
-        assert_eq!(end, end & mask);
-        start..(end.wrapping_add(1) & mask)
+    /// Returns `true` if all possible numbers are valid, i.e `valid_range` covers the whole layout
+    #[inline]
+    pub fn is_always_valid_for<C: HasDataLayout>(&self, cx: &C) -> bool {
+        self.valid_range.is_full_for(self.value.size(cx))
     }
```
eddyb, line 835: I like this setup! Only one tiny nit, I don't think `is_always_valid_for` needs the `_for` at the end. The argument doesn't dictate validity necessarily, it's just a required context. So `is_always_valid` seems enough.

As a complete aside (and a bad idea to tackle in this PR, heh), I wonder how many context arguments on methods we could remove if we got rid of the sole `Primitive::Pointer` and actually made it track its own size.

eddyb (approved):
I only have a couple nits now, and I've changed the title to reflect this PR does more than just the one method. Thanks for following up on the late review comments!

compiler/rustc_target/src/abi/mod.rs
bonega: I actually thought about how much happier my life would be if `Primitive::Pointer` was gone.
It seems very entrenched though, is it even possible to remove?

bonega:
Thanks for giving the reviews, they are very helpful :+1:

compiler/rustc_target/src/abi/mod.rs
eddyb: We can't remove it fully (there are semantic differences), only make it so `TargetDataLayout` isn't needed to compute its size.

eddyb (approved):

compiler/rustc_target/src/abi/mod.rs
```diff
@@ -792,7 +792,7 @@ impl WrappingRange {
     pub fn is_full_for(&self, size: Size) -> bool {
         let max_value = size.unsigned_int_max();
         debug_assert!(self.start <= max_value && self.end <= max_value);
-        (self.start == 0 && self.end == max_value) || (self.end + 1 == self.start)
+        (self.start == 0 && self.end == max_value) || (self.end.wrapping_add(1) == self.start)
```
eddyb: Wait, shouldn't this use `& max_value`, after the `wrapping_add(1)`?

compiler/rustc_target/src/abi/mod.rs
```diff
@@ -779,17 +786,29 @@ impl WrappingRange {
         self.end = end;
         self
     }
+
+    /// Returns `true` if `size` completely fills the range.
+    #[inline]
+    pub fn is_full_for(&self, size: Size) -> bool {
+        let max_value = size.unsigned_int_max();
+        debug_assert!(self.start <= max_value && self.end <= max_value);
+        (self.start == 0 && self.end == max_value) || (self.end.wrapping_add(1) == self.start)
```
eddyb, line 795: Yeah, looking at the original, I see `start..(end.wrapping_add(1) & mask)`, so `& max_value` is definitely needed.

But also, that condition on its own will account for `self.start == 0 && self.end == max_value` (since `self.end.wrapping_add(1) & max_value` will be `0` i.e. `self.start`), so that part can be removed.

eddyb (changes requested):

compiler/rustc_target/src/abi/mod.rs
bonega: Good catch, originally I took the code from https://github.com/rust-lang/rust/blob/434cb437b55d61bcb54a01921de7ac752e6dee13/compiler/rustc_mir/src/interpret/validity.rs#L631
Will fix it.

RalfJung: @eddyb I think with the extra `||`, it does not need the `& mask`. So I think the old code was correct. But maybe you find the bit magic version cleaner?

eddyb: Without the masking, the operation is not equivalent to "add 1 *as the unsigned integer type of the given size*", which makes it more confusing for me as to what the intent is.

It's less "bit magic" and more "selecting the right ring to do arithmetic on", it just so happens that `mod 2**n` is implemented as `& (2**n - 1)`.

The removal of the redundant check is just a small bonus - after all, the condition we're looking for is "there is no gap between `end` and `start`, `mod 2**n`".

RalfJung: I'd much prefer an actual `mod` and letting the compiler sort this out, that would be so much more readable.^^ The current code is obfuscated, as far as I am concerned. But annoying finite arithmetic makes that tricky too...

compiler/rustc_target/src/abi/mod.rs
bonega: I'm confused, I thought the size of the pointer would change depending on context?

compiler/rustc_target/src/abi/mod.rs
eddyb: I would gladly support using `%` if `2**128` could fit in an `u128`.
One idea I was considering is having something like `% Pow2(int_size.bits())`, that does the right thing internally.

Or maybe a wrapper for `u128` that tracks the integer size dynamically and performs all operations according to that size (I suspect miri would like something like this in several places).

bonega:
rebased

compiler/rustc_target/src/abi/mod.rs
bonega: Working on something like this.
Is it reasonable from a performance perspective to store `Size` in `Primitive`?
`./x.py test ui` passes

ehuss:
@bors retry

Windows image update has broken things.

PR #88332, opened 2021-08-25
Add argument types tait tests

spastorino:
r? @oli-obk 

Related to #86727


src/test/ui/type-alias-impl-trait/argument-types.rs
```diff
@@ -0,0 +1,43 @@
...
+fn foo2(mut x: Foo) {
+    // no constraint on x
+}
+
+fn foo3(x: Foo) {
+    println!("{:?}", x);
+}
+
+// FIXME: what do we expect this to do?
+fn foo2_1(mut x: impl Debug) {
```
spastorino, line 23: I'm not really sure if we want `impl Traits` in these tests

PR #88334, opened 2021-08-25
Simplify `bootstrap.py`

UltiRequiem:
Using Python's ternary expression is more readable.


Mark-Simulacrum:
Yes, I think our general tendency for the Python code in-tree (particularly bootstrap) should be to keep it Rust-like (e.g., not using 'bar if cond else foo`) since the primary audience for it is Rust developers in most cases. I think there's probably cleanup work to be done around this code, but I don't think this PR is quite it.

If you'd like to, I'd be happy to accept a PR adding a "styleguide" or similar comment reflecting this perspective to the bootstrap.py file in particular.

Thanks!

PR #88335, opened 2021-08-25
Do not re-export macros that are already at the crate root

cjgillot:
Follow-up to #88019 

Exported macro_rules were unconditionally reexported at the crate root. In the case of macro_rules that were already at the root, both the macro itself and its reexport appear with the same name in the same scope, which confuses the resolver for downstream crates.

This PR skips creating a re-export when the macro was already at the crate root.
r? @petrochenkov


petrochenkov:
#88019 has been merged, unblocking.

JohnCSimon:
Triage: Merge conflicts

JohnCSimon:
Ping @cjgillot 
> #88019 has been merged, unblocking.

petrochenkov:
I'm going to close this in favor of https://github.com/rust-lang/rust/pull/91795.

PR #88338, opened 2021-08-25
rustc_target: route `LayoutOf` through `TyAbiInterface`.

eddyb:
This is a bit strange (potentially an ergonomic disaster, especially around errors, see below), but I decided to try and use the `trait TyAbiInterface` from #88337 to provide a single implementation of `layout_of`, tied to `Ty<'tcx>`, and usable with any context that provides:
* `TyCtxt`, via `HasTyCtxt`
* `ParamEnv`, via `HasParamEnv`
* a way to transform `LayoutError`s into the desired error type
  * an error type of `!` can be paired with having `cx.layout_of(...)` return `TyAndLayout` *without* `Result<...>` around it, such as used by codegen
  * this is done through a new `IsLayoutCx` trait (and so is specifying the type of `cx.layout_of(...)`)

Sadly, because of the generics involved, and Rust type-checking/inference limitations (mostly around autoderef and coercions), certain things that we took for granted change:
* `cx: &mut C` doesn't autoderef in `cx.layout_of(...)`, if only `C: LayoutOf`
  * I worked around this with blanket `impl`s of the relevant traits on `&mut _`
* `cx.layout_of(ty: &Ty)` doesn't coerce `&Ty` to `Ty`
  * theoretically I could've worked around this by implementing `TyAbiInterface` on `&Ty`, but I decided to fix the callsites instead

I doubt this change can be landed in its initial state (or at all), but since I did get it to work, I wanted to put it up.

### DIFF
--- a/compiler/rustc_codegen_cranelift/src/base.rs
+++ b/compiler/rustc_codegen_cranelift/src/base.rs
@@ -79,7 +79,7 @@ pub(crate) fn codegen_fn<'tcx>(
     let arg_uninhabited = fx
         .mir
         .args_iter()
-        .any(|arg| fx.layout_of(fx.monomorphize(&fx.mir.local_decls[arg].ty)).abi.is_uninhabited());
+        .any(|arg| fx.layout_of(fx.monomorphize(fx.mir.local_decls[arg].ty)).abi.is_uninhabited());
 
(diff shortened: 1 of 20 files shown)


eddyb:
Closing as superseded by #88499, which doesn't have the ergonomic downsides.

Issue #83997, opened 2021-04-08
Documentation: inline code with links split into multiple blocks


tspiteri:
**EDIT(camelid): ⚠️ Please do not open a PR for this! It is still under discussion.**

-----

Right now in the standard library documentation, when a piece of inline code has part of it linked, the code is split into multiple `<code>` blocks, which renders poorly. For example the [`Arc` thread safety section](https://doc.rust-lang.org/nightly/std/sync/struct.Arc.html#thread-safety) contains

![image](https://user-images.githubusercontent.com/18604588/114037935-b6340e00-9881-11eb-9739-0b5b62b03158.png)

Here, between `Arc<` and `RefCell`, and between `<T>` and `>`, there is extra spacing. This is because the doc comment contains
```rust
/// `Arc<`[`RefCell<T>`]`>`
```
which is rendered using three `<code>` blocks as
```html
<code>Arc&lt;</code><a href="../../std/cell/struct.RefCell.html"><code>RefCell&lt;T&gt;</code></a><code>&gt;</code>
```

I think better would be
```rust
/// <code>Arc&lt;[RefCell&lt;T&gt;][RefCell]&gt;</code>
```
which presumably is rendered in one block as
```html
<code>Arc&lt;<a href="../../std/cell/struct.RefCell.html">RefCell&lt;T&gt;</a>&gt;</code>
```
(I'm not sure if the link to `RefCell` gets a `title` attribute as well).

The downside is that now the doc comment markdown is a bit hairy with explicit `<code>`, `&lt;` and `&gt;` spread throughout.

Is this substitution considered an improvement? If so I can grep through the code, make some substitutions, and open a PR.

An alternative solution would be to change rustdoc itself to combine consecutive code blocks, some of which could be wrapped inside `<a>`/`</a>`, into a single block, though I do not have the know-how to open such a PR. I don't even know if there is some downside to this that I am missing, where `<code>` blocks are meant to be separate, maybe for some markdown compliance or something.

jyn514:
Yes, I think this is a good change - one like it has already been accepted (https://github.com/rust-lang/rust/pull/80720) and I need to merge the other, thanks for reminding me (https://github.com/rust-lang/rust/pull/80733). I need to open an issue for rustdoc to do this automatically - I don't know if it's worth spending time to convert the whole standard library by hand in the meantime.

[...]

PR #80733, opened 2021-01-05
Improve links in inline code in `core::pin`.

steffahn:
## Context

So I recently opened #80720. That PR uses HTML-based `<code>foo</code>` syntax in place of `` `foo` `` for some inline code. It looks like usage of `<code>` tags in doc comments is without precedent in the standard library, but the HTML-based syntax has an important advantage:

You can write something like
```
<code>[Box]<[Option]\<T>></code>
```
which becomes: <code>[Box]<[Option]\<T>></code>, whereas with ordinary backtick syntax, you cannot create links for a substring of an inline code block.

## Problem
I recalled (from my own experience) that a way to partially work around this limitation is to do something like
```
[`Box`]`<`[`Option`]`<T>>`
```
which looks like this: [`Box`]`<`[`Option`]`<T>>` _(admitted, it looks even worse on GitHub than in `rustdoc`’s CSS)_.

[Box]: https://doc.rust-lang.org/std/boxed/struct.Box.html "Box"
[`Box`]: https://doc.rust-lang.org/std/boxed/struct.Box.html "Box"
[Option]: https://doc.rust-lang.org/std/option/enum.Option.html "Option"
[`Option`]: https://doc.rust-lang.org/std/option/enum.Option.html "Option"
[Pin]: https://doc.rust-lang.org/std/pin/struct.Pin.html "Pin"
[&mut]: https://doc.rust-lang.org/std/primitive.reference.html "mutable reference"

So I searched the standard library and found that e.g. the [std::pin](https://doc.rust-lang.org/std/pin/index.html) module documentation uses this hack/workaround quite a bit, with types like <code>[Pin]<[Box]\<T>></code> or <code>[Pin]<[&mut] T>></code>. Although the way they look like in this sentence is what I would like them to look like, not what they currently look.

### Status Quo

Here’s a screenshot of what it currently looks like:
Screenshot_20210105_202751

With a few HTML-style code blocks, we can fix all the spacing issues in the above screenshot that are due usage of this hack/workaround of putting multiple code blocks right next to each other being used.

### after d3915c555ee016b11ce288e107e46dbab9f78c4f:
Screenshot_20210105_202932

There’s still a problem of inconsistency. Especially in a sentence such as
> A [`Pin<P>`][Pin] where `P: Deref` should be considered as a "`P`-style pointer" to _[...]_

[...]

PR #88343, opened 2021-08-25
Fix spacing of links in inline code.

steffahn:
Similar to #80733, but the focus is different. This PR eliminates all occurrences of pieced-together inline code blocks like [`Box`]`<`[`Option`]`<T>>` and replaces them with good-looking ones (using HTML-syntax), like <code>[Box]<[Option]\<T>></code>. As far as I can tell, I should’ve found all of these in the standard library (regex search with `` r"`\]`|`\[`" ``) \[except for in `core::convert` where I’ve noticed other things in the docs that I want to fix in a separate PR]. In particular, unlike #80733, I’ve added almost no new instance of inline code that’s broken up into multiple links (or some link and some link-free part). I also added tooltips (the stuff in quotes for the markdown link listings) in places that caught my eye, but that’s by no means systematic, just opportunistic.

[Box]: https://doc.rust-lang.org/std/boxed/struct.Box.html "Box"
[`Box`]: https://doc.rust-lang.org/std/boxed/struct.Box.html "Box"
[Option]: https://doc.rust-lang.org/std/option/enum.Option.html "Option"
[`Option`]: https://doc.rust-lang.org/std/option/enum.Option.html "Option"

Context: I got annoyed by repeatedly running into new misformatted inline code while reading the standard library docs. I know that once issue #83997 (and/or related ones) are resolved, these changes become somewhat obsolete, but I fail to notice much progress on that end right now.

r? @jyn514

### DIFF
--- a/library/alloc/src/fmt.rs
+++ b/library/alloc/src/fmt.rs
@@ -348,7 +348,7 @@
 //! provides some helper methods.
 //!
 //! Additionally, the return value of this function is [`fmt::Result`] which is a
-//! type alias of [`Result`]`<(), `[`std::fmt::Error`]`>`. Formatting implementations
+//! type alias of <code>[Result]<(), [std::fmt::Error]></code>. Formatting implementations
 //! should ensure that they propagate errors from the [`Formatter`] (e.g., when
 //! calling [`write!`]). However, they should never return errors spuriously. That
 //! is, a formatting implementation must and may only return an error if the
@@ -505,23 +505,19 @@
 //! it would internally pass around this structure until it has been determined
 //! where output should go to.
 //!
-//! [`fmt::Result`]: Result
-//! [`Result`]: core::result::Result
-//! [`std::fmt::Error`]: Error
-//! [`write!`]: core::write
-//! [`write`]: core::write
-//! [`format!`]: crate::format
-//! [`to_string`]: crate::string::ToString
-//! [`writeln!`]: core::writeln
+//! [`fmt::Result`]: Result "fmt::Result"
+//! [Result]: core::result::Result "std::result::Result"
+//! [std::fmt::Error]: Error "fmt::Error"
+//! [`write`]: write() "fmt::write"
+//! [`to_string`]: crate::string::ToString::to_string "ToString::to_string"
 //! [`write_fmt`]: ../../std/io/trait.Write.html#method.write_fmt
 //! [`std::io::Write`]: ../../std/io/trait.Write.html
-//! [`print!`]: ../../std/macro.print.html
-//! [`println!`]: ../../std/macro.println.html
-//! [`eprint!`]: ../../std/macro.eprint.html
-//! [`eprintln!`]: ../../std/macro.eprintln.html
-//! [`format_args!`]: core::format_args
-//! [`fmt::Arguments`]: Arguments
-//! [`format`]: crate::format
+//! [`print!`]: ../../std/macro.print.html "print!"
+//! [`println!`]: ../../std/macro.println.html "println!"
+//! [`eprint!`]: ../../std/macro.eprint.html "eprint!"
+//! [`eprintln!`]: ../../std/macro.eprintln.html "eprintln!"
+//! [`fmt::Arguments`]: Arguments "fmt::Arguments"
+//! [`format`]: format() "fmt::format"
 
 #![stable(feature = "rust1", since = "1.0.0")]
 
--- a/library/alloc/src/sync.rs
+++ b/library/alloc/src/sync.rs
@@ -99,8 +99,8 @@ macro_rules! acquire {
 /// first: after all, isn't the point of `Arc<T>` thread safety? The key is
 /// this: `Arc<T>` makes it thread safe to have multiple ownership of the same
 /// data, but it  doesn't add thread safety to its data. Consider
-/// `Arc<`[`RefCell<T>`]`>`. [`RefCell<T>`] isn't [`Sync`], and if `Arc<T>` was always
-/// [`Send`], `Arc<`[`RefCell<T>`]`>` would be as well. But then we'd have a problem:
+/// <code>Arc<[RefCell\<T>]></code>. [`RefCell<T>`] isn't [`Sync`], and if `Arc<T>` was always
+/// [`Send`], <code>Arc<[RefCell\<T>]></code> would be as well. But then we'd have a problem:
 /// [`RefCell<T>`] is not thread safe; it keeps track of the borrowing count using
 /// non-atomic operations.
 ///
@@ -176,6 +176,7 @@ macro_rules! acquire {
 /// [deref]: core::ops::Deref
 /// [downgrade]: Arc::downgrade
 /// [upgrade]: Weak::upgrade
+/// [RefCell\<T>]: core::cell::RefCell
 /// [`RefCell<T>`]: core::cell::RefCell
 /// [`std::sync`]: ../../std/sync/index.html
 /// [`Arc::clone(&from)`]: Arc::clone
@@ -206,7 +207,7 @@ macro_rules! acquire {
 ///
 /// Sharing a mutable [`AtomicUsize`]:
 ///
-/// [`AtomicUsize`]: core::sync::atomic::AtomicUsize
+/// [`AtomicUsize`]: core::sync::atomic::AtomicUsize "sync::atomic::AtomicUsize"
 ///
 /// ```no_run
 /// use std::sync::Arc;
@@ -262,7 +263,7 @@ impl<T: ?Sized> Arc<T> {
 
 /// `Weak` is a version of [`Arc`] that holds a non-owning reference to the
 /// managed allocation. The allocation is accessed by calling [`upgrade`] on the `Weak`
-/// pointer, which returns an [`Option`]`<`[`Arc`]`<T>>`.
+/// pointer, which returns an <code>[Option]<[Arc]\<T>></code>.
 ///
 /// Since a `Weak` reference does not count towards ownership, it will not
 /// prevent the value stored in the allocation from being dropped, and `Weak` itself makes no
@@ -476,7 +477,7 @@ impl<T> Arc<T> {
     /// assert_eq!(*zero, 0)
     /// ```
     ///
-    /// [zeroed]: ../../std/mem/union.MaybeUninit.html#method.zeroed
+    /// [zeroed]: mem::MaybeUninit::zeroed
     #[cfg(not(no_global_oom_handling))]
     #[unstable(feature = "new_uninit", issue = "63291")]
     pub fn new_zeroed() -> Arc<mem::MaybeUninit<T>> {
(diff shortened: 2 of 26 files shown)


library/alloc/src/fmt.rs
```diff
@@ -505,23 +505,19 @@
...
 //!
-//! [`fmt::Result`]: Result
-//! [`Result`]: core::result::Result
-//! [`std::fmt::Error`]: Error
-//! [`write!`]: core::write
-//! [`write`]: core::write
-//! [`format!`]: crate::format
-//! [`to_string`]: crate::string::ToString
-//! [`writeln!`]: core::writeln
+//! [`fmt::Result`]: Result "fmt::Result"
```
jyn514, line 508: By default this won't have a tooltip, right? Why did you add one?

jyn514 (commented):
Looked at the first couple commits and they seem fine, but there's a lot of churn adding tooltips and I'm not really sure why they're useful to add - would like to hear more about that before spending more time on review.

library/alloc/src/fmt.rs
steffahn: In general, links generated by rustdoc do have a tooltip.

E.g. something like
```rs
/// Talking about [`Result`][std::fmt::Result] here!
pub mod m {}
```
will come with a tooltip

Screenshot_20210906_223756

Whereas (unfortunately) the following alternative
```rs
/// Talking about [`Result`] here!
/// 
/// [`Result`]: std::fmt::Result
pub mod m {}
```
creates almost the same documentation, just lacking the tooltip. That’s the general reason why I’ve converted many instances of
```markdown
[foo]: path
```
into
```markdown
[foo]: path "path"
```
and in some cases, I’ve also adjusted the tooltip slightly to make it more easy to understand. This includes the case here because `Result` is part of the prelude, a simple `"Result"` tooltip seems confusing to me, the `std::` prefix is something I’ve often avoided for tooltips OTOH (not least because the paths themselves also often are relative so some imported `std::foo` module).

<hr>

Also, as mentioned in the description of this PR, the addition of tooltips isn’t systematic and only covers places that I happened to come across. (Furthermore, the amount of tooltips added declines from earlier to later commits.)

steffahn:
> I'm not really sure why they're useful to add

For me it’s mostly a question of consistency. I don’t like _not_ having a tooltip simply because of the decision to not put the link inline but separately in a links section in markdown. I don’t like reading docs where half the links give me tooltips and the other half don’t that’s also the reason why I had systematically included tooltips in #80733.

library/alloc/src/fmt.rs
jyn514: Ok. I think this would be better to fix systematically in rustdoc rather than adding it piecemeal to individual links. Could you open an issue and revert the tooltips you've added this PR?

steffahn: "revert" as in “add another commit to revert them” or as a “rebase / edit / force push”. Because the former could easily happen in a later PR, too; these tooltips aren't particularly hard to find (and remove) with a regex search.

Either way, the revert would only remove those cases where the path and tooltip look _identical_, right? E.g. the `fmt::Result` should better stay IMO, because I don't like a `Result` tooltip.

steffahn: Hmm, no answer. I’m assuming that an additional commit is okay, and that only the ones where tooltip and path are identical need to be removed. This is more easily accomplished by a systematic find-and-replace that by looking through all my commits here. This could, as noted, easily be a separate PR, done *after* the relevant change to rustdoc was made, but I’m including a commit here now to get this PR going.

jyn514: Sorry for the delay, I haven't had much time for rust lately.

> Either way, the revert would only remove those cases where the path and tooltip look identical, right? E.g. the fmt::Result should better stay IMO, because I don't like a Result tooltip.

Well, ideally I wouldn't want any changes, because it makes them inconsistent and it will hopefully be done by rustdoc automatically eventually ... but if you feel strongly it seems fine to keep, it's just more work for me to review.

> I’m assuming that an additional commit is okay, and that only the ones where tooltip and path are identical need to be removed.

Yup, that's fine - I don't have strong opinions about git commits.

steffahn: > I think showing the link used in the source is more clear in a way but I don't feel strongly.

The problem I see is that the link used in the source code depends on what the imports are, and the imports depends on what happens to be used in the source code. If e.g. I’m writing documentation mentioning, say, `mem::transmute`, then I’d use the path `men::transmute` (resulting in a `men::transmute` tooltip) if the module I’m in does `use core::mem` (or `use crate::mem`). But if `mem` is **only** used in docs, AFAIK I can’t just add an import for that because (I think) it’ll trigger an unused imports lint and fail CI. But it’s IMO better to *concistently* have it have the tooltip `mem::transmute` and not sometimes `core::mem::transmute` or and `crate::mem::transmute`, etc…, depending on the implementation details (i.e. imports) of the module your documentation comes from (in particular, I don’t think the `crate::` prefix is a good thing showing up in standard library tooltips).

library/core/src/option.rs
```diff
@@ -111,16 +111,20 @@
...
-//! * [`as_deref`] converts from `&Option<T>` to `Option<&T::Target>`
-//! * [`as_deref_mut`] converts from `&mut Option<T>` to
-//!   `Option<&mut T::Target>`
-//! * [`as_pin_ref`] converts from [`Pin`]`<&Option<T>>` to
-//!   `Option<`[`Pin`]`<&T>>`
-//! * [`as_pin_mut`] converts from [`Pin`]`<&mut Option<T>>` to
-//!   `Option<`[`Pin`]`<&mut T>>`
-//!
+//! * [`as_ref`] converts from <code>[&][][Option]\<T></code> to <code>[Option]<[&]T></code>
+//! * [`as_mut`] converts from <code>[&mut] [Option]\<T></code> to <code>[Option]<[&mut] T></code>
```
jyn514, line 115: I assume these are doing odd things to avoid linking `&` to Option - is there a reason you use `[&][]` for one and `[&mut] ` for the other?

library/core/src/option.rs
```diff
@@ -111,16 +111,20 @@
...
+//!   <code>[Option]<[&]T::[Target]></code>
+//! * [`as_deref_mut`] converts from <code>[&mut] [Option]\<T></code> to
+//!   <code>[Option]<[&mut] T::[Target]></code>
+//! * [`as_pin_ref`] converts from <code>[Pin]<[&][][Option]\<T>></code> to
+//!   <code>[Option]<[Pin]<[&]T>></code>
+//! * [`as_pin_mut`] converts from <code>[Pin]<[&mut] [Option]\<T>></code> to
+//!   <code>[Option]<[Pin]<[&mut] T>></code>
+//!
+//! [&]: ../../std/primitive.reference.html "shared reference"
+//! [&mut]: ../../std/primitive.reference.html "mutable reference"
```
jyn514, line 126:
```suggestion
//! [&]: reference "shared reference"
//! [&mut]: reference "mutable reference"
```

library/core/src/option.rs
```diff
@@ -649,7 +653,9 @@ impl<T> Option<T> {
         }
     }
 
-    /// Converts from [`Pin`]`<&Option<T>>` to `Option<`[`Pin`]`<&T>>`.
+    /// Converts from <code>[Pin]<[&]Option\<T>></code> to <code>Option<[Pin]<[&]T>></code>.
+    ///
+    /// [&]: ../../std/primitive.reference.html "shared reference"
```
jyn514, line 658:
```suggestion
    /// [&]: reference "shared reference"
```

library/core/src/option.rs
```diff
@@ -658,7 +664,9 @@ impl<T> Option<T> {
         unsafe { Pin::get_ref(self).as_ref().map(|x| Pin::new_unchecked(x)) }
     }
 
-    /// Converts from [`Pin`]`<&mut Option<T>>` to `Option<`[`Pin`]`<&mut T>>`.
+    /// Converts from <code>[Pin]<[&mut] Option\<T>></code> to <code>Option<[Pin]<[&mut] T>></code>.
+    ///
+    /// [&mut]: ../../std/primitive.reference.html "mutable reference"
```
jyn514, line 669:
```suggestion
    /// [&mut]: reference "mutable reference"
```

jyn514 (commented):
I would prefer not to review PRs this big in the future - please wait until some of the rustdoc issues are fixed before making more PRs.

library/core/src/option.rs
steffahn: There’s no space after the `&`. The `[]` seems to be able to separate the two links without changing the fact that the `[&]: ../../std/primitive.reference.html "shared reference"` link is used. The space after `[&mut]` suffices to avoid the same problem there.

library/core/src/option.rs
steffahn: oh, that works too!? great!

library/core/src/option.rs
jyn514: Right, but why not use the space in both cases?

steffahn: Because we don’t want to render `& Option<T>` but `&Option<T>`

steffahn: It’s conventionally `&T` and `&mut T`, one has a space, the other doesn’t.

jyn514: Got it, that makes sense :+1:

library/alloc/src/sync.rs
```diff
@@ -476,7 +477,7 @@ impl<T> Arc<T> {
     /// assert_eq!(*zero, 0)
     /// ```
     ///
-    /// [zeroed]: ../../std/mem/union.MaybeUninit.html#method.zeroed
+    /// [zeroed]: mem::MaybeUninit::zeroed
```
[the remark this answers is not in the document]
steffahn: I have no idea :-)

jyn514:
r=me with the linkchecker passing. Not sure if you have bors privileges or not, just in case -

@bors delegate=steffahn

Please use `r=jyn514`, not r+.

jyn514:
oh - and please squash the commits before merging too if you don't mind :)

steffahn:
> Not sure if you have bors privileges

I don’t :-)

> please squash the commits before merging

okay

steffahn:
Squashed ([no changes](https://github.com/rust-lang/rust/compare/0cd383569d45ebf86a7f98e223ff6c2cfb779dee..67065fe93366cfdd9e88440b6f876c3dfeb1f937))

@bors r=jyn514

steffahn:
This “regression” is about how fast rustc compiles some specific crate, right? (I have no idea where to find more information about those test cases.) This PR only changes (doc) comments in the standard library, so the regression is either sporadic or very weird.

If it's a sporadic thing, the benchmark of the next PR that gets merged by bors should perform better again; we'll see.

rylev:
An update: thanks to [great investigation](https://rust-lang.zulipchat.com/#narrow/stream/247081-t-compiler.2Fperformance/topic/investigation.20.2389343) by @Mark-Simulacrum, we've identified a possible cause for this which should be fixed by https://github.com/rust-lang/rust/pull/89408

PR #88344, opened 2021-08-25
Remove premature optimization in exhaustiveness checking

Nadrieril:
There are some optimizations that I never measured but add a lot of complexity, so let's see if they're worth it.

r? @ghost


Nadrieril:
Closing in favor of https://github.com/rust-lang/rust/pull/88950

PR #88353, opened 2021-08-26
Partially stabilize `array_methods`

jhpratt:
This stabilizes `<[T; N]>::as_slice` and `<[T; N]>::as_mut_slice`, which is forms part of the `array_methods` feature: #76118.

This also makes `<[T; N]>::as_slice` const due to its trivial nature.


jhpratt:
Ping @joshtriplett FCP completed a few days ago. No rush to merge as we've a number of weeks to the next beta fork, but just want to bring this to your attention.

vityafx:
Sorry for chiming in, but is there a work-around as of now, to create a slice from an array? I need to work with `[u8; 4]` as with `&[u8]`.

PR #83813, opened 2021-04-03
Fix `--remap-path-prefix` not correctly remapping `rust-src` component paths and unify handling of path mapping with virtualized paths

cbeuw:
This PR fixes #73167 ("Binaries end up containing path to the rust-src component despite `--remap-path-prefix`") by preventing real local filesystem paths from reaching compilation output if the path is supposed to be remapped.

`RealFileName::Named` introduced in #72767 is now renamed as `LocalPath`, because this variant wraps a (most likely) valid local filesystem path.

`RealFileName::Devirtualized` is renamed as `Remapped` to be used for remapped path from a real path via `--remap-path-prefix` argument, as well as real path inferred from a virtualized (during compiler bootstrapping) `/rustc/...` path. The `local_path` field is now an `Option<PathBuf>`, as it will be set to `None` before serialisation, so it never reaches any build output. Attempting to serialise a non-`None` `local_path` will cause an assertion faliure.

When a path is remapped, a `RealFileName::Remapped` variant is created. The original path is preserved in `local_path` field and the remapped path is saved in `virtual_name` field. Previously, the `local_path` is directly modified which goes against its purpose of "suitable for reading from the file system on the local host".

`rustc_span::SourceFile`'s fields `unmapped_path` (introduced by #44940) and `name_was_remapped` (introduced by #41508 when `--remap-path-prefix` feature originally added) are removed, as these two pieces of information can be inferred from the `name` field: if it's anything other than a `FileName::Real(_)`, or if it is a `FileName::Real(RealFileName::LocalPath(_))`, then clearly `name_was_remapped` would've been false and `unmapped_path` would've been `None`. If it is a `FileName::Real(RealFileName::Remapped{local_path, virtual_name})`, then `name_was_remapped` would've been true and `unmapped_path` would've been `Some(local_path)`. 

cc @eddyb who implemented `/rustc/...` path devirtualisation

PR #70642, opened 2020-04-01
Translate the virtual `/rustc/$hash` prefix back to a real directory.

eddyb:
Closes #53486 and fixes #53081, by undoing the remapping to `/rustc/$hash` on the fly, when appropriate (e.g. our testsuites, or user crates that depend on `libstd`), but not during the Rust build itself (as that could leak the absolute build directory into the artifacts, breaking deterministic builds).

Tested locally by setting `remap-debuginfo = true` in `config.toml`, which without these changes, was causing 56 tests to fail (see https://github.com/rust-lang/rust/issues/53081#issuecomment-606703215 for more details).

cc @Mark-Simulacrum @alexcrichton @ehuss

PR #88363, opened 2021-08-26
Path remapping: Make behavior of diagnostics output dependent on presence of --remap-path-prefix.

### LINKED ISSUE #87745: --remap-path-prefix no longer works for compiler messages
### Code

Create a directory and put a Rust file with bad syntax in it:
```
mkdir src
echo bad > src/test.rs
```

Then compile it with:

```
rustc --remap-path-prefix=src=MODIFIED src/test.rs
```

I expected to see output like this:

```
error: expected one of `!` or `::`, found `<eof>`
 --> MODIFIED/test.rs:1:1
[snip]
```

Instead, the path is not modified and I get:

```
error: expected one of `!` or `::`, found `<eof>`
 --> src/test.rs:1:1
[snip]
```

### Version it worked on

rustc 1.53.0 (53cb7b09b 2021-06-17)

### Version with regression

michaelwoerister:
This PR fixes a regression (#87745) with `--remap-path-prefix` where the flag stopped causing diagnostic messages to be remapped as well. The regression was introduced in https://github.com/rust-lang/rust/pull/83813 where we erroneously assumed that remapping of diagnostic messages was not desired anymore (because #70642 partially undid that functionality with nobody objecting).

The issue is fixed by making `--remap-path-prefix` remap diagnostic messages again, including for paths that have been remapped in upstream crates (e.g. the standard library). This means that "sysroot-localization" (implemented in #70642) is also disabled if `rustc` is invoked with `--remap-path-prefix`. The assumption is that once someone starts explicitly remapping paths they also don't want paths to their local Rust installation in their build output.

In the future we might want to give more fine-grained control over this behavior via compiler flags (see https://github.com/rust-lang/rfcs/pull/3127 for a related RFC). For now this PR is intended as a regression fix.

This PR is an alternative to https://github.com/rust-lang/rust/pull/88191, which makes diagnostic messages be remapped unconditionally. That approach, however, would effectively revert #70642.

Fixes https://github.com/rust-lang/rust/issues/87745.

cc @cbeuw 
r? @ghost

### DIFF
--- a/compiler/rustc_codegen_cranelift/src/debuginfo/mod.rs
+++ b/compiler/rustc_codegen_cranelift/src/debuginfo/mod.rs
@@ -66,7 +66,7 @@ impl<'tcx> DebugContext<'tcx> {
             rustc_interface::util::version_str().unwrap_or("unknown version"),
             cranelift_codegen::VERSION,
         );
-        let comp_dir = tcx.sess.opts.working_dir.to_string_lossy(false).into_owned();
+        let comp_dir = tcx.sess.opts.working_dir.to_string_lossy(FileNameDisplayPreference::Remapped).into_owned();
         let (name, file_info) = match tcx.sess.local_crate_source_file.clone() {
             Some(path) => {
                 let name = path.to_string_lossy().into_owned();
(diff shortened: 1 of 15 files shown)


GuillaumeGomez:
Is it the one failing in https://github.com/rust-lang/rust/pull/88584?

estebank:
@GuillaumeGomez I don't see how it could be, but maybe I'm missing something?

GuillaumeGomez:
Apparently not, it was another one. I looked at the PRs changing content in `rustc_ast`. :)

Issue #81890, opened 2021-02-08
spurious CI error for MSVC


ehuss:
There have been a few instances of this error happening on CI:

>   = note: LINK : fatal error LNK1201: error writing to program database 'D:\a\rust\rust\build\i686-pc-windows-msvc\test\run-make-fulldeps\extern-flag-fun\extern-flag-fun\foo.pdb'; check for insufficient disk space, invalid path, or insufficient privilege

Or some variant of that.  See:

* https://github.com/rust-lang/rust/pull/81257#issuecomment-771231808
* https://github.com/rust-lang/rust/pull/81411#issuecomment-774824226
* https://github.com/rust-lang/rust/pull/80802#issuecomment-759450136
* https://github.com/rust-lang/rust/pull/80591#issuecomment-753535696
* https://github.com/rust-lang/rust/pull/80209#issuecomment-751316136
* https://github.com/rust-lang/rust/pull/80122#issuecomment-747612121
* https://github.com/rust-lang/rust/pull/80009#issuecomment-754213646
* https://github.com/rust-lang/rust/pull/80003#issuecomment-745227156

(and probably many more)

kennytm:
Not just i686.

* https://github.com/rust-lang/rust/pull/90542#issuecomment-967741298
* https://github.com/rust-lang/rust/pull/90684#issuecomment-966792302

hkratz:
The cargo test `lto::complicated` has been triggering this a lot lately. And cargo itself prints warnings, because the same .pdb file is written multiple times:
```
021-11-28T17:13:11.5135303Z The bin target `test` in package `test v0.0.0 (D:\a\rust\rust\build\x86_64-pc-windows-msvc\stage2-tools\x86_64-pc-windows-msvc\tmp\cit\t1300\foo)` has the same output filename as the lib target `test` in package `test v0.0.0 (D:\a\rust\rust\build\x86_64-pc-windows-msvc\stage2-tools\x86_64-pc-windows-msvc\tmp\cit\t1300\foo)`.
2021-11-28T17:13:11.5137366Z Colliding filename is: D:\a\rust\rust\build\x86_64-pc-windows-msvc\stage2-tools\x86_64-pc-windows-msvc\tmp\cit\t1300\foo\target\release\deps\test.pdb
2021-11-28T17:13:11.5138476Z The targets should have unique names.
2021-11-28T17:13:11.5139250Z Consider changing their names to be unique or compiling them separately.
[...]

PR #88368, opened 2021-08-26
 Improve error when an .rlib can't be parsed

jyn514:
This usually describes either an error in the compiler itself or some
sort of IO error. Either way, we should report it to the user rather
than just saying "crate not found".

This only gives an error if the crate couldn't be loaded at all - if the
compiler finds another .rlib or .rmeta file which was valid, it will
continue to compile the crate.

Example output:
```
error[E0785]: found invalid metadata files for crate `foo`
 --> bar.rs:3:24
  |
3 |         println!("{}", foo::FOO_11_49[0]);
  |                        ^^^
  |
  = warning: failed to parse rlib '/home/joshua/test-rustdoc/libfoo.rlib': Invalid archive extended name offset
```

cc @ehuss

### DIFF
--- a/compiler/rustc_metadata/src/locator.rs
+++ b/compiler/rustc_metadata/src/locator.rs
@@ -350,6 +350,7 @@ impl<'a> CrateLocator<'a> {
         self.crate_rejections.via_kind.clear();
         self.crate_rejections.via_version.clear();
         self.crate_rejections.via_filename.clear();
+        self.crate_rejections.via_invalid.clear();
     }
 
     crate fn maybe_load_library_crate(&mut self) -> Result<Option<Library>, CrateError> {
@@ -548,7 +549,17 @@ impl<'a> CrateLocator<'a> {
                             continue;
                         }
                     }
-                    Err(err) => {
+                    Err(MetadataError::LoadFailure(err)) => {
+                        info!("no metadata found: {}", err);
+                        // The file was present and created by the same compiler version, but we
+                        // couldn't load it for some reason.  Give a hard error instead of silently
+                        // ignoring it, but only if we would have given an error anyway.
+                        self.crate_rejections
+                            .via_invalid
+                            .push(CrateMismatch { path: lib, got: err });
+                        continue;
+                    }
+                    Err(err @ MetadataError::NotPresent(_)) => {
                         info!("no metadata found: {}", err);
                         continue;
                     }
@@ -726,25 +737,28 @@ impl<'a> CrateLocator<'a> {
 fn get_metadata_section(
     target: &Target,
     flavor: CrateFlavor,
-    filename: &Path,
+    filename: &'p Path,
     loader: &dyn MetadataLoader,
-) -> Result<MetadataBlob, String> {
+) -> Result<MetadataBlob, MetadataError<'p>> {
     if !filename.exists() {
-        return Err(format!("no such file: '{}'", filename.display()));
+        return Err(MetadataError::NotPresent(filename));
     }
     let raw_bytes: MetadataRef = match flavor {
-        CrateFlavor::Rlib => loader.get_rlib_metadata(target, filename)?,
+        CrateFlavor::Rlib => {
+            loader.get_rlib_metadata(target, filename).map_err(MetadataError::LoadFailure)?
+        }
         CrateFlavor::Dylib => {
-            let buf = loader.get_dylib_metadata(target, filename)?;
+            let buf =
+                loader.get_dylib_metadata(target, filename).map_err(MetadataError::LoadFailure)?;
             // The header is uncompressed
             let header_len = METADATA_HEADER.len();
             debug!("checking {} bytes of metadata-version stamp", header_len);
             let header = &buf[..cmp::min(header_len, buf.len())];
             if header != METADATA_HEADER {
-                return Err(format!(
-                    "incompatible metadata version found: '{}'",
+                return Err(MetadataError::LoadFailure(format!(
+                    "invalid metadata version found: {}",
                     filename.display()
-                ));
+                )));
             }
 
             // Header is okay -> inflate the actual metadata
@@ -756,17 +770,28 @@ fn get_metadata_section(
             match FrameDecoder::new(compressed_bytes).read_to_end(&mut inflated) {
                 Ok(_) => rustc_erase_owner!(OwningRef::new(inflated).map_owner_box()),
                 Err(_) => {
-                    return Err(format!("failed to decompress metadata: {}", filename.display()));
+                    return Err(MetadataError::LoadFailure(format!(
+                        "failed to decompress metadata: {}",
+                        filename.display()
+                    )));
                 }
             }
         }
         CrateFlavor::Rmeta => {
             // mmap the file, because only a small fraction of it is read.
-            let file = std::fs::File::open(filename)
-                .map_err(|_| format!("failed to open rmeta metadata: '{}'", filename.display()))?;
+            let file = std::fs::File::open(filename).map_err(|_| {
+                MetadataError::LoadFailure(format!(
+                    "failed to open rmeta metadata: '{}'",
+                    filename.display()
+                ))
+            })?;
             let mmap = unsafe { Mmap::map(file) };
-            let mmap = mmap
-                .map_err(|_| format!("failed to mmap rmeta metadata: '{}'", filename.display()))?;
+            let mmap = mmap.map_err(|_| {
+                MetadataError::LoadFailure(format!(
+                    "failed to mmap rmeta metadata: '{}'",
+                    filename.display()
+                ))
+            })?;
 
             rustc_erase_owner!(OwningRef::new(mmap).map_owner_box())
         }
@@ -775,7 +800,10 @@ fn get_metadata_section(
     if blob.is_compatible() {
         Ok(blob)
     } else {
-        Err(format!("incompatible metadata version found: '{}'", filename.display()))
+        Err(MetadataError::LoadFailure(format!(
+            "invalid metadata version found: {}",
+            filename.display()
+        )))
     }
 }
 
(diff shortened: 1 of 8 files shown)


jyn514:
> I think metadata errors from this PR can use the same approach - stashed into new rejected_via_* field(s) and reported at the top level (locator.into_error()) in cases when nothing is found.

I implemented this.

jyn514:
Err, I think I accidentally reused an existing error code. I'll add a new one.

src/test/ui/crate-loading/invalid-rlib.stderr
```diff
@@ -0,0 +1,11 @@
+error[E0785]: found invalid metadata files for crate `foo`
+  --> $DIR/invalid-rlib.rs:4:7
+   |
+LL | use ::foo;
+   |       ^^^
+   |
+   = warning: failed to mmap file '$DIR/auxiliary/libfoo.rlib': memory map must have a non-zero length
```
bjorn3, line 7: This should be a note I think and not a warning.

compiler/rustc_metadata/src/locator.rs
```diff
@@ -723,18 +733,49 @@ impl<'a> CrateLocator<'a> {
...
+            Self::LoadFailure(msg) => f.write_str(msg),
+        }
+    }
+}
+
+impl From<String> for MetadataError<'static> {
+    fn from(msg: String) -> Self {
+        MetadataError::LoadFailure(msg)
+    }
+}
```
petrochenkov, line 769: This impl looks unnecessary, `MetadataError` is never used in generic context and `msg.into()` is less readable than `MetadataError::LoadFailure(msg)`.

compiler/rustc_metadata/src/locator.rs
```diff
@@ -723,18 +733,49 @@ impl<'a> CrateLocator<'a> {
...
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        match self {
+            Self::NotPresent(filename) => {
+                f.write_str(&format!("no such file: '{}'", filename.display()))
+            }
+            Self::WrongVersion(filename) => f.write_str(&format!(
+                "incompatible metadata version found: '{}'",
+                filename.display()
+            )),
+            Self::LoadFailure(msg) => f.write_str(msg),
```
petrochenkov, line 760:
```suggestion
            MetadataError::LoadFailure(msg) => f.write_str(msg),
```
`Self` in non-generic context almost always hurts code searchability and readability.

compiler/rustc_metadata/src/locator.rs
```diff
@@ -723,18 +733,49 @@ impl<'a> CrateLocator<'a> {
...
+enum MetadataError<'a> {
+    /// The file was missing.
+    NotPresent(&'a Path),
+    /// The file was present and valid, but built by a different and incompatible version of the compiler.
+    WrongVersion(&'a Path),
+    /// The file was present and invalid.
+    LoadFailure(String),
+}
+
+impl fmt::Display for MetadataError<'_> {
```
petrochenkov, line 750: Could you move `MetadataError` and this impl to the bottom of the file where all other error reporting code lives?

compiler/rustc_metadata/src/locator.rs
```diff
@@ -723,18 +733,49 @@ impl<'a> CrateLocator<'a> {
...
+            rejected_via_invalid: self.rejected_via_invalid,
         })
     }
 }
 
+enum MetadataError<'a> {
+    /// The file was missing.
+    NotPresent(&'a Path),
+    /// The file was present and valid, but built by a different and incompatible version of the compiler.
+    WrongVersion(&'a Path),
```
petrochenkov, line 745: This one also looks legitimately useful to report.
`NotPresent` seems like the only case where we can be completely silent.

compiler/rustc_metadata/src/locator.rs
```diff
@@ -549,6 +552,13 @@ impl<'a> CrateLocator<'a> {
...
                         }
                     }
+                    Err(MetadataError::LoadFailure(err)) => {
+                        // The file was present and created by the same compiler version, but we
+                        // couldn't load it for some reason.  Give a hard error instead of silently
+                        // ignoring it, but only if we would have given an error anyway.
+                        self.rejected_via_invalid.push(CrateMismatch { path: lib, got: err });
+                        continue;
+                    }
                     Err(err) => {
```
petrochenkov, line 562: It would be better to make this matching exhaustive to see which errors we are skipping silently.

compiler/rustc_metadata/src/locator.rs
jyn514: I added it to allow using `format!("some error")?`. But if you think that's less readable I'll remove it.

compiler/rustc_metadata/src/locator.rs
jyn514: @bjorn3 asked me not to report this, since it happens quite often (e.g. when you first build with stable and again with nightly): https://github.com/rust-lang/rust/pull/88368#discussion_r697604206. But now that the logic only reports an error if no rlibs at all were found, I agree it makes sense to report the wrong version too.

jyn514: (The commit I pushed just now has nonsensical output for this case; working now on fixing it.)

petrochenkov:
r=me with the `run-make-fulldeps` test updated.

GuillaumeGomez:
It failed in rollup https://github.com/rust-lang/rust/pull/88774.

@bors: r-

jyn514:
The problem was that `.so` isn't recognized as a dynamic library suffix on all platforms. I changed the test to only run on x86.

@bors r=petrochenkov

bjorn3:
You could use `touch $(rustc --crate-name foo --crate-type dylib --print file-names)` in a run-make test to make an empty dylib file with the right prefix and suffis for the target.

I think limiting to x86_64 is not correct by the way as windows uses .dll as suffix.

jyn514:
@bjorn3 I implemented your suggestion - what do you think?

src/test/run-make/invalid-so/Makefile
```diff
@@ -0,0 +1,7 @@
+include ../../run-make-fulldeps/tools.mk
+
+DYLIB_NAME := $(shell rustc --crate-name foo --crate-type dylib --print file-names /dev/null)
```
bjorn3, line 3: `/dev/null` doesn't exist on Windows. Would `echo | rustc ... -` work?

jyn514: Sure, and I fixed it for the test itself too.

compiler/rustc_metadata/src/locator.rs
```diff
@@ -538,7 +539,16 @@ impl<'a> CrateLocator<'a> {
                             continue;
                         }
                     }
-                    Err(err) => {
+                    Err(MetadataError::LoadFailure(err)) => {
+                        // The file was present and created by the same compiler version, but we
+                        // couldn't load it for some reason.  Give a hard error instead of silently
+                        // ignoring it, but only if we would have given an error anyway.
```
bjorn3, line 545: Can you keep the `warn!()` in this case? It may still be useful.

jyn514: Sure, done.

jyn514:
Looks spurious?
```
-    = note: failed to mmap file '$DIR/auxiliary/libfoo.rlib': memory map must have a non-zero length
+    = note: failed to mmap file '\?\D:/a/rust/rust/src/test/ui/crate-loading/auxiliary/libfoo.rlib': The volume for a file has been externally altered so that the opened file is no longer valid. (os error 1006)
9 error: aborting due to previous error
```

ehuss:
Those look like they are raw OS error messages. I'm not sure what the guidelines are for UI tests, but I might recommend ignoring that part of the string. Those messages can be different on every operating system.

(It is trying to mmap the rlib, and different OS's will return different errors for an empty file. Though that Windows error is not very clear.)

ehuss:
Ah, it looks like rustc is using an older version of memmap2.  Newer versions handle 0 length files differently (they don't return an error).

camelid:
@bors r-

Tests failed, but bors still has this in the queue.

jyn514:
I managed to mess up the rebase again and `test src/tools/clippy --bless` isn't working :/
```
failures:
    [ui] ui/crashes/ice-3969.rs

test result: FAILED. 661 passed; 1 failed; 3 ignored; 0 measured; 0 filtered out; finished in 9.40s

test compile_test ... FAILED

failures:

---- compile_test stdout ----
diff of stderr:

-error: trait bound str: std::marker::Sized does not depend on any type or lifetime parameters
-  --> $DIR/ice-3969.rs:20:10
-   |
-LL |     str: Sized;
-   |          ^^^^^
-   |
-   = note: `-D trivial-bounds` implied by `-D warnings`
-
-error: trait bound for<'a> Dst<(dyn A + 'a)>: std::marker::Sized does not depend on any type or lifetime parameters
-  --> $DIR/ice-3969.rs:24:30
-   |
-LL |     for<'a> Dst<dyn A + 'a>: Sized,
-   |                              ^^^^^
-
-error: trait bound str: std::marker::Sized does not depend on any type or lifetime parameters
-  --> $DIR/ice-3969.rs:31:10
-   |
-LL |     str: Sized,
-   |          ^^^^^
-
-error: trait bound std::string::String: std::ops::Neg does not depend on any type or lifetime parameters
-  --> $DIR/ice-3969.rs:38:13
-   |
-LL |     String: ::std::ops::Neg<Output = String>,
-   |             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
-
-error: trait bound i32: std::iter::Iterator does not depend on any type or lifetime parameters
-  --> $DIR/ice-3969.rs:45:10
-   |
-LL |     i32: Iterator,
-   |          ^^^^^^^^
-
-error: aborting due to 5 previous errors
-
-
```

camelid:
Spuriuous?

ehuss:
@bors retry 

#81890

jyn514:
I suspect the regression is just because the new `include_str` makes the binary slightly larger.

rylev:
@jyn514's suggestion is certainly a possibility. The regressions are so small and the improvements are much larger (significance factor wise) that I don't think it's really worth the cost of investigation. 

PR #88369, opened 2021-08-26
update const generics feature gates

lcnr:
**tl;dr: split const generics into three features: `adt_const_params`, `const_generics_defaults` and `generic_const_exprs`**

continuing the work of @BoxyUwU in #88324, this PR
- renames `feature(const_evaluatable_checked)` to `feature(generic_const_exprs)` which now doesn't need any other feature gate to work. Previously `feature(const_evaluatable_checked)` was only useful in combination with `feature(const_generics)`.
- completely removes `feature(lazy_normalization_consts)`. This feature only supplied the parents generics to anonymous constants, which is pretty useless as generic anon consts are only allowed with `feature(generic_const_exprs)` anyways.
- moves the ability to use additional const param types from `feature(const_generics)` into `feature(adt_const_params)`. As `feature(const_generics)` is now mostly useless without `feature(generic_const_exprs)` we also remove that feature flag.
- updates tests, removing duplicates and unnecessary revisions in some cases and also deletes all unused `*.stderr` files.

I not also remove the ordering restriction for const and type parameters if any of the three const generics features is active.
This ordering restriction feels like the only "real" use of the current `feature(const_generics)` right now so this change isn't a perfect solution, but as I intend to stabilize the ordering - and `feature(const_generics_defaults)` -  in the very near future, I think this is acceptable for now.

---
 
cc @rust-lang/project-const-generics about the new feature names and this change in general.

I don't think we need any external approval for this change but I do intend to publish an update to the const generics tracking issue the day this PR lands, so I don't want this merged yet.

Apologies to whoever ends up reviewing this PR :sweat_smile: :heart: 

r? rust-lang/project-const-generics


lcnr:
renamed `const_param_types` to `adt_const_params` after some discussion

compiler/rustc_ast_passes/src/feature_gate.rs
```diff
@@ -687,7 +687,6 @@ pub fn check_crate(krate: &ast::Crate, sess: &Session) {
     gate_all!(trait_alias, "trait aliases are experimental");
     gate_all!(associated_type_bounds, "associated type bounds are unstable");
     gate_all!(crate_visibility_modifier, "`crate` visibility modifier is experimental");
-    gate_all!(const_generics, "const generics are unstable");
```
estebank, line 690: Don't we have to add these to a "removed feature" list?

estebank: Just had to scroll down a bit more to answer myself.

lcnr:
this looks spurious

@bors r=oli-obk p=1

lcnr:
this looks spurious again?  not going to retry this PR right now as the bors queue seems to be quite full and I am not certain this is spurious

jyn514:
x86_64-apple failed after 2 hours without logs

@bors retry

PR #88375, opened 2021-08-26
Clarify that ManuallyDrop<T> has same layout as T

joshlf:
This PR implements the documentation change under discussion in https://github.com/rust-lang/unsafe-code-guidelines/issues/302. It should not be approved or merged until the discussion there is resolved.

### DIFF
--- a/library/core/src/mem/manually_drop.rs
+++ b/library/core/src/mem/manually_drop.rs
@@ -4,11 +4,12 @@ use crate::ptr;
 /// A wrapper to inhibit compiler from automatically calling `T`’s destructor.
 /// This wrapper is 0-cost.
 ///
-/// `ManuallyDrop<T>` is subject to the same layout optimizations as `T`.
-/// As a consequence, it has *no effect* on the assumptions that the compiler makes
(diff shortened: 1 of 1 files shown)


joshtriplett:
Removing waiting-on-review, since this is blocked on the UCG issue.

Mark-Simulacrum:
@joshtriplett it feels like a PR modifying the std docs here is *the* way to make progress on a guarantee (in the absence of some documentation for what else should be done). If it isn't, then it doesn't seem appropriate to leave this PR open -- while it's not a huge cost, if we're not expecting a decision for an indeterminate period or discussion, it doesn't seem useful to keep PRs idle for a long time.

Perhaps this should be nominated for some team?

joshtriplett:
I don't have any objections to just FCPing this here; it just seemed like the consensus thus far was that this was a matter for the UCG.

joshlf:
What's the best way to move this forward?

dtolnay:
I think this is blatantly correct, and worth documenting.

@rfcbot fcp merge

dtolnay (approved):

PR #88397, opened 2021-08-27
Tracking issue for `UNSUPPORTED_CALLING_CONVENTIONS`

nagisa:
This was previously forgotten. Nominating for 1.55 as this lint will make it into stable in that release.

r? @Mark-Simulacrum

PR #88399, opened 2021-08-27
Disallow the aapcs CC on Aarch64

nagisa:
This never really worked and makes LLVM assert.


nagisa:
@bors r=petrochenkov

Updated the UI test output for changes in https://github.com/rust-lang/rust/pull/88397

PR #88410, opened 2021-08-27
Remove bolding on associated constants

camelid:
Associated types don't get bolded, so it looks off to have one kind
bolded and one not.

### DIFF
--- a/src/test/rustdoc-gui/font-weight.goml
+++ b/src/test/rustdoc-gui/font-weight.goml
@@ -13,6 +13,21 @@ goto: file://|DOC_PATH|/test_docs/struct.Foo.html
 assert-css: (".impl-items .method", {"font-weight": "600"}, ALL)
 
 goto: file://|DOC_PATH|/lib2/trait.Trait.html
+
+// This is a complex selector, so here's how it works:
+//
+// * //*[@class='docblock type-decl'] — selects element of any tag with classes docblock and type-decl
+// * /pre[@class='rust trait'] — selects immediate child with tag pre and classes rust and trait
+// * /code — selects immediate child with tag code
+// * /a[@class='constant'] — selects immediate child with tag a and class constant
+// * //text() — selects child that is text node
+// * /parent::* — selects immediate parent of the text node (the * means it can be any tag)
+//
+// This uses '/parent::*' as a proxy for the style of the text node.
+// We can't just select the '<a>' because intermediate tags could be added.
+assert-count: ("//*[@class='docblock type-decl']/pre[@class='rust trait']/code/a[@class='constant']//text()/parent::*", 1)
+assert-css: ("//*[@class='docblock type-decl']/pre[@class='rust trait']/code/a[@class='constant']//text()/parent::*", {"font-weight": "400"})
+
 assert-count: (".methods .type", 1)
 assert-css: (".methods .type", {"font-weight": "600"})
 assert-count: (".methods .constant", 1)
--- a/src/librustdoc/html/render/mod.rs
+++ b/src/librustdoc/html/render/mod.rs
@@ -753,7 +753,7 @@ fn assoc_const(
 ) {
     write!(
         w,
-        "{}{}const <a href=\"{}\" class=\"constant\"><b>{}</b></a>: {}",
+        "{}{}const <a href=\"{}\" class=\"constant\">{}</a>: {}",
         extra,
         it.visibility.print_with_space(it.def_id, cx),
         naive_assoc_href(it, link, cx),


GuillaumeGomez:
Please add tests in `src/test/rustdoc` to check the absence of the tag (check that the parent tag with the text exists but not the `<b>` inside it).

camelid:
> Please add tests in `src/test/rustdoc` to check the absence of the tag (check that the parent tag with the text exists but not the `<b>` inside it).

@GuillaumeGomez I added a GUI test case to check the font-weight since that will also catch if the font-weight changes because of CSS.

src/test/rustdoc-gui/font-weight.goml
```diff
@@ -13,6 +13,8 @@ goto: file://|DOC_PATH|/test_docs/struct.Foo.html
 assert-css: (".impl-items .method", {"font-weight": "600"}, ALL)
 
 goto: file://|DOC_PATH|/lib2/trait.Trait.html
+assert-count: (".type-decl > pre.trait > code > a.constant", 1)
+assert-css: (".type-decl > pre.trait > code > a.constant", {"font-weight": "400"})
```
GuillaumeGomez, line 17: I'm definitely not sure that it'll it because it's the inner content which had the `font-weight` change, not the container.

camelid: Good point, I have now changed it to use XPath with `/text()`.

src/test/rustdoc-gui/font-weight.goml
```diff
@@ -13,6 +13,10 @@ goto: file://|DOC_PATH|/test_docs/struct.Foo.html
 assert-css: (".impl-items .method", {"font-weight": "600"}, ALL)
 
 goto: file://|DOC_PATH|/lib2/trait.Trait.html
+// This uses '/parent::*' as a proxy for the style of the text node.
```
GuillaumeGomez, line 16: I still it would be much simpler to add a `rustdoc` test checking that there is no `<b>` inside `.constant`.

Something like:

```
@has - 'a[@class="constant"]' 'content'
@!has - 'a[@class="constant"]/b'
```

camelid: I could do that, but it seems relatively unlikely to me that someone would accidentally add a `<b>` tag. It's much more frequent that rustdoc styling bugs are caused by inadvertent CSS changes, which wouldn't be caught by an `@has` check.

camelid: For context, the `<b>` tag that I removed had been there for 4 1/2 years (since 081336e8eb5d94071dad7f6fee9efd00305dc381)—when it was added, it worked correctly, but rustdoc's HTML and CSS has since changed such that it is now rendered incorrectly.

camelid: And also, worst comes to worse, having this regress does not significantly impact the docs. It would just look less consistent.

GuillaumeGomez: This a good argument. Then please simplify the XPath, I have literally no idea what it's doing currently. XD

camelid: Which parts would you like me to simplify?

I can also explain what each part means:

- `//*[@class='docblock type-decl']` — selects element of any tag with classes `docblock` and `type-decl`
- `/pre[@class='rust trait']` — selects immediate child with tag `pre` and classes `rust` and `trait`
- `/code` — selects immediate child with tag `code`
- `/a[@class='constant']` — selects immediate child with tag `a` and class `constant`
- `//text()` — selects child that is text node
- `/parent::*` — selects immediate parent of the text node (the `*` means it can be any tag)

GuillaumeGomez: Please add this whole explanation as a comment then. Once done, r=me.

camelid: I added a link to my explanation comment on GitHub above because I felt that it would be hard to format the explanation well in a code comment.

GuillaumeGomez: Sorry but not a big fan of adding a link here. What do you think about:

```
// This is a complex selector, so here how it works:
//
// * //*[@class='docblock type-decl'] — selects element of any tag with classes docblock and type-decl
// * /pre[@class='rust trait'] — selects immediate child with tag pre and classes rust and trait
// * /code — selects immediate child with tag code
// * /a[@class='constant'] — selects immediate child with tag a and class constant
// * //text() — selects child that is text node
// * /parent::* — selects immediate parent of the text node (the * means it can be any tag)
```


?

camelid: The problem with adding a comment is that it can fall out of date quickly without it being obvious that it's out of date. Whereas if you link to a PR comment, people can easily see that it's old. But, I'll add the comment since I agree that there is some advantage to having it inline :)

GuillaumeGomez: Thanks!

Issue #66219, opened 2019-11-08
Document all unsafety used in libcore and libstd


oli-obk:
The repository still contains many instances of 

```rust
// ignore-tidy-undocumented-unsafe
```

We should eliminate all of them by documenting the unsafe used in those files.

Documenting unsafe blocks works by adding a `// SAFETY:` comment infront of them explaining why the unsafe block is ok. If there is an explanation about the unsafety elsewhere in the file, you can also leave a `// SAFETY:` comment explaining that the overarching logic is explained elsewhere (and mention where, too!)

cc @Centril 

not sure what labels to add to this issue.

See https://github.com/rust-lang/rust/pull/63793/files for examples and the introduction of the unsafety tidy check

foeb:
Hi! I'm interested in working on this. I can probably only commit to 6-8 hours a week, so it might be more realistic for me to work on just libcore or libstd. Is that alright?

LeSeulArtichaut:
@foeb I took the liberty of opening #71063. I hope you weren't working on those 3 files 😅

foeb:
@LeSeulArtichaut 
I actually did one of those yesterday but it's fine, I'll take any help I can get for this. If it was just me alone, it'll take me years :smile: 
Let's coordinate over email first next time though, so we don't have too much duplicated effort

hbina:
Hello,

Is anyone working on https://github.com/rust-lang/rust/blob/2dc5b602eee35d70e8e6e506a7ea07b6c7e0197d/src/libcore/slice/sort.rs#L35-L55

This can be shown to be safe because `len > 2` => `len - 2 > 0`, right?

foeb:
I haven't had the time to work on this for a while, so I don't think it's right for me to hold on to this assignment.

@rustbot release-assignment

LeSeulArtichaut:
Here's a list of all the files we have left to document. We can use it to synchronize and track our progress.
- [ ] src/libcore/panicking.rs
- [ ] src/libcore/option.rs (@LeSeulArtichaut, #71063)
- [ ] src/libcore/hint.rs
- [ ] src/libcore/alloc/layout.rs
- [ ] src/libcore/hash/mod.rs (@LeSeulArtichaut, #71063)
- [ ] src/libcore/hash/sip.rs (@LeSeulArtichaut, #71063)
- [ ] src/libcore/slice/mod.rs
- [ ] src/libcore/slice/sort.rs
- [ ] src/libcore/slice/memchr.rs
[...]

PR #88412, opened 2021-08-28
Remove ignore-tidy-undocumented-unsafe from core::slice::sort

mdsn:
Write down the missing safety arguments to be able to remove `ignore-tidy-undocumented-unsafe` from `core::slice::sort`.

Helps with #66219 

@rustbot label C-cleanup T-libs


dtolnay (approved):
Thanks!

Issue #67792, opened 2020-01-02
Tracking issue for RFC 2362, `impl const Trait for Ty` syntax


ecstatic-morse:
TODO

clarfonthey:
Now that I've read up and understand the reasoning for the `~const` syntax, here's my opinion on it.

Basically, we would prefer that constness be a bonus for generic impls to offer, but not a requirement. A trait impl can take all the steps required to make something const, but if its dependencies are not const, we would rather offer a non-const method than a compiler error.

Basically, if we say `impl const Add` and the result ends up being something that's not const, it's bad.

But, that's exactly what const does right now. Const doesn't mean to always evaluate things in a constant context; it just means that it *can* be evaluated in a constant context. Whether it's const depends entirely on the constness of the inputs; `2.pow(3)` is evaluated at const but ,`x.pow(y)` is not.

The relevant analogue here would be that the constness of an input implementation of a trait is what determines the constness of the output impl.

But... that's exactly how const works for functions, and we don't annotate arguments with constness. The only case where it would be useful is if some arguments are unused, which is exactly what we have here for the type parameters.

So, really, this annotation is letting us know which type parameters are unused when determining constness. Which seems a lot like what the original `?const` syntax was trying to indicate, although I guess the main difference with `?Sized` is it doesn't actually affect whether an impl is valid, just if it's const or not.

I really do think that it's better to opt out of the constness rather than opt in. But I guess there's probably more reasons that you chose it than I'm seeing.

RalfJung:
> Whether it's const depends entirely on the constness of the inputs; 2.pow(3) is evaluated at const but ,x.pow(y) is not.

This is wrong. `const` and `static` initializers, array lengths, and enum discriminant values are evaluated at compile-time, but *all other code* is evaluated according to runtime rules, and `const fn` vs `fn` is entirely irrelevant for that.

Lokathor:
In the language that comes after Rust it might be possible to say that all functions are const by default and mark ones which aren't const. However, for Rust that's just not possible because of backwards compatibility reasons. Since functions opt in to being const, then trait impls must also opt in to being const for consistency.

clarfonthey:
So, I'm a bit confused how `#[default_method_body_is_const]` is supposed to interact with `~const`.

Say I have these hypothetical traits:

```rust
trait Parent {
    fn parenting(self) -> Self;
}
trait Child: Parent {
    fn childing(self) -> Self {
        self.parenting()
    }
}
```

[...]

PR #88418, opened 2021-08-28
Allow `~const` bounds on trait assoc functions

fee1-dead:
r? @oli-obk

### DIFF
--- a/src/test/ui/rfc-2632-const-trait-impl/trait-where-clause-run.rs
+++ b/src/test/ui/rfc-2632-const-trait-impl/trait-where-clause-run.rs
@@ -1,0 +1,41 @@
+// run-pass
+
+#![feature(const_trait_impl)]
+#![feature(const_fn_trait_bound)]
+
+trait Bar {
+    fn bar() -> u8;
+}
+
+trait Foo {
+    #[default_method_body_is_const]
+    fn foo() -> u8 where Self: ~const Bar {
+        <Self as Bar>::bar() * 6
+    }
+}
+
+struct NonConst;
+struct Const;
+
+impl Bar for NonConst {
+    fn bar() -> u8 {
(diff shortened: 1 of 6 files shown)


src/test/ui/rfc-2632-const-trait-impl/trait-where-clause.rs
```diff
@@ -0,0 +1,33 @@
+#![feature(const_fn_trait_bound)]
+#![feature(const_trait_impl)]
+
+trait Bar {}
+
+trait Foo {
+    fn a();
+    fn b() where Self: ~const Bar;
```
oli-obk, line 8: A few more tests I think we should have:
* `where Self: ~const Foo`
* a const and a nonconst impl for such a trait
* a monomorphic use that actually runs it

src/test/ui/rfc-2632-const-trait-impl/trait-where-clause-run.rs
```diff
@@ -0,0 +1,41 @@
...
+    }
+}
+
+impl const Foo for Const {}
+
+fn main() {
+    const ANS1: u8 = Const::foo();
+    let ans2 = NonConst::foo();
+
+    assert_eq!(ANS1 + ans2, 42);
```
oli-obk, line 40: beautiful

nalply:
I got here by TWIR and I never have seen `~const` in Rust. What's this? Sorry for the noise, `~const` is not googlable.

oli-obk:
it's placeholder syntax for "const if const" trait bounds: https://github.com/rust-lang/rust/issues/67792

the full updated RFC text can be found at https://internals.rust-lang.org/t/pre-rfc-revamped-const-trait-impl-aka-rfc-2632/15192

PR #88419, opened 2021-08-28
Fix code blocks color in Ayu theme

### LINKED ISSUE #88415: rustdoc: Ayu syntax highlighting regression
My guess is that the yellow inline code style is incorrectly being applied to code blocks too.

### Nightly (incorrect)

image

### Beta & stable (correct)

image

GuillaumeGomez:
Fixes #88415.

cc @camelid 
r? @notriddle


notriddle (approved):

camelid:
(Updated the PR title to make it more searchable.)

src/librustdoc/html/static/css/themes/ayu.css
```diff
@@ -40,7 +40,7 @@ h4 {
 .code-header {
 	color: #e6e1cf;
 }
-pre > code {
+.docblock pre > code, pre > code {
```
camelid, line 43: Hmm, are you sure this is the right fix? To me, it looks like it adds a redundant CSS rule.

camelid: When I tried this fix in my browser, it didn't work. This fix did work, however:

```patch
From 093d1c74e51a0bbc87aec9dd97eba690dce7499a Mon Sep 17 00:00:00 2001
From: Noah Lev <camelidcamel@gmail.com>
Date: Sat, 28 Aug 2021 12:01:59 -0700
Subject: [PATCH] Fix color of code blocks in Ayu theme

---
 src/librustdoc/html/static/css/themes/ayu.css | 2 +-
 1 file changed, 1 insertion(+), 1 deletion(-)

diff --git a/src/librustdoc/html/static/css/themes/ayu.css b/src/librustdoc/html/static/css/themes/ayu.css
index dab6d655c6a..ceca92c8ea9 100644
--- a/src/librustdoc/html/static/css/themes/ayu.css
+++ b/src/librustdoc/html/static/css/themes/ayu.css
@@ -34,7 +34,7 @@ h4 {
 	background: rgba(0, 0, 0, 0);
 }
 
-.docblock code {
+.docblock :not(pre) > code {
 	color: #ffb454;
 }
 .code-header {
```

GuillaumeGomez: The GUI tests proved it worked.

GuillaumeGomez: To extend a bit: the current code is saying "if you are a `code` and are inside a `.docblock`, then the color is somewhat yellow". This fix adds an intermediate case saying "if are a `code` directly inside a `pre` inside a `.docblock`, then the color is somewhat white", which takes precedence over the previous one.

camelid: To me, the diff looks like the old code was

> if you are a `code` directly inside a `pre`, then the color is `#e6e1cf` (close to white)

and the new code is

> if you are a `code` directly inside a `pre` which is somewhere inside a `.docblock`, or if you are a `code` directly inside a `pre`, then the color is `#e6e1cf` (close to white)

The new `.docblock pre > code` case you added looks redundant to me relative to the `pre > code` case. Won't every element that matches `.docblock pre > code` also match `pre > code`, since `pre > code` is more general?

**P.S.:** I just discovered a GitHub feature I hadn't known about before, which is that if you put a hex color in inline code, it will add a little color swatch for you! E.g., the `#e6e1cf` above.

notriddle: I’m pretty sure it’s a specificity problem, because `.docblock code` is more specific than `pre > code`, thanks to classes being more specific than tag names.

https://developer.mozilla.org/en-US/docs/Web/CSS/Specificity

camelid: Hmm, yeah, this PR *does* fix it locally for me—I'm not sure why it didn't work when I copy-pasted it into the Firefox dev tools. But wouldn't it make the CSS much clearer to instead change the rule on line 37 that adds the `#ffb454` color to `.docblock :not(pre) > code`? The reason this solution works was quite non-obvious to me.

GuillaumeGomez: I prefer to say "do if your parent is X" rather than "do if your parent isn't X". I find it easier to read.

camelid: I prefer that too, but it seems confusing to me to rely on specificity in this case. But, now that we have a GUI test, at least people shouldn't be able to break it accidentally, so I'm OK with it if you want to do it this way :)

GuillaumeGomez: 👍

PR #88422, opened 2021-08-28
add #[inline] to Path comparison fast path

the8472:
Let's see if this fixes the regression noted in https://github.com/rust-lang/rust/pull/86898#issuecomment-905524405


the8472:
Doesn't fix the `-doc` regressions.

PR #82858, opened 2021-03-07
unix: Extend UpdSocket to send and receive TTL for a specific packet

LinkTed:
Add the function `recv_vectored_with_ancillary_from`,
`recv_vectored_with_ancillary`, `send_vectored_with_ancillary_to` and
`send_vectored_with_ancillary` to `UdpSocket` for Unix platforms. Also add
`set_recvttl` and `recvttl` to `UdpSocket` to tell the kernel to receive the
TTL. Therefore, the `AncillaryData` enum had to be extended.

PR #88425, opened 2021-08-28
ancillary data size and doc improvement

beroal:
Functions calculating the size of a control message. Improvement of the doc for `SocketAncillary` and the main doc. Clarification that `add_fds` and `add_creds` add **one** control message.
r? @LinkTed

### DIFF
--- a/library/std/src/os/unix/net/ancillary.rs
+++ b/library/std/src/os/unix/net/ancillary.rs
@@ -78,6 +78,36 @@ pub(super) fn send_vectored_with_ancillary_to(
     }
 }
 
+/// The size in octets of a control message containing `n` `T`s.
+///
+/// # Panics
+///
+/// If the size is so large that it causes arithmetic overflow.
+fn ancillary_size_data<T>(n: usize) -> u32 {
+    let r = u32::try_from(n.checked_mul(size_of::<T>()).unwrap()).unwrap();
+    unsafe { libc::CMSG_SPACE(r) }
+}
+
+/// The size in octets of a control message containing `n` file descriptors.
+///
+/// # Panics
+///
+/// If the size is so large that it causes arithmetic overflow.
+#[unstable(feature = "unix_socket_ancillary_data", issue = "76915")]
+pub fn ancillary_size_fds(n: usize) -> u32 {
+    ancillary_size_data::<RawFd>(n)
+}
+
+/// The size in octets of a control message containing `n` unix credentials.
+///
+/// # Panics
+///
+/// If the size is so large that it causes arithmetic overflow.
+#[unstable(feature = "unix_socket_ancillary_data", issue = "76915")]
+pub fn ancillary_size_creds(n: usize) -> u32 {
+    ancillary_size_data::<SocketCred>(n)
+}
+
 fn add_to_ancillary_data<T>(
     buffer: &mut [u8],
     length: &mut usize,
@@ -369,7 +399,19 @@ impl<'a> Iterator for Messages<'a> {
     }
 }
 
-/// A Unix socket Ancillary data struct.
+/// A struct for accessing a Unix socket ancillary data packet.
+///
+/// An *ancillary data* (also known as *control data*) *packet* may be attached
+/// to a usual *data segment* transferred through a Unix socket.
+/// See [“2.10.11 Socket Receive Queue”](https://pubs.opengroup.org/onlinepubs/9699919799/functions/V2_chap02.html#tag_15_10_11)
+/// in POSIX for details. An ancillary data packet is a list of *ancillary data objects*
+/// (also known as *control messages*).
+/// Here an ancillary data packet is represented by a [`SocketAncillary`] and a control message
+/// (see `struct cmsghdr` in the C API) is represented by an [`AncillaryData`].
+/// The C API for ancillary data:
+///
+/// - [`sys/socket.h`](https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/sys_socket.h.html)
+/// - [`cmsg`(3)](https://man7.org/linux/man-pages/man3/cmsg.3.html)
 ///
 /// # Example
 /// ```no_run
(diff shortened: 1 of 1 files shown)


library/std/src/os/unix/net/ancillary.rs
```diff
@@ -78,6 +78,36 @@ pub(super) fn send_vectored_with_ancillary_to(
...
+
+/// The size in octets of a control message containing `n` file descriptors.
+///
+/// # Panics
+///
+/// If the size is so large that it causes arithmetic overflow.
+#[unstable(feature = "unix_socket_ancillary_data", issue = "76915")]
+pub fn ancillary_size_fds(n: usize) -> u32 {
+    ancillary_size_data::<RawFd>(n)
+}
```
the8472, line 99: The type-specific methods don't seem particularly useful, they only avoid having to type the turbofish. Imo the uses be covered by adding examples to `ancillary_size_data`, making it public and removing the other methods.

`SocketCred` and `RawFd` aren't private types, so there's no reason to hide them.

Well, the one downside I can see is that it would tie the rust struct size to the underlying libc type needed for cmsg. But they pretty much serve the purpose of being thin wrappers for the platform types anyway, so that doesn't seem like a real loss.

library/std/src/os/unix/net/ancillary.rs
```diff
@@ -369,7 +399,19 @@ impl<'a> Iterator for Messages<'a> {
...
+/// to a usual *data segment* transferred through a Unix socket.
+/// See [“2.10.11 Socket Receive Queue”](https://pubs.opengroup.org/onlinepubs/9699919799/functions/V2_chap02.html#tag_15_10_11)
+/// in POSIX for details. An ancillary data packet is a list of *ancillary data objects*
+/// (also known as *control messages*).
+/// Here an ancillary data packet is represented by a [`SocketAncillary`] and a control message
+/// (see `struct cmsghdr` in the C API) is represented by an [`AncillaryData`].
+/// The C API for ancillary data:
+///
+/// - [`sys/socket.h`](https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/sys_socket.h.html)
+/// - [`cmsg`(3)](https://man7.org/linux/man-pages/man3/cmsg.3.html)
```
the8472, line 414: Nice, documentation is one of the things that was sorely missing. But I wouldn't phrase it in terms of unix sockets. The underlying API can be used to attach metadata to any kind of socket. #82858 is going to extend it to udp sockets.

library/std/src/os/unix/net/ancillary.rs
```diff
@@ -407,7 +449,13 @@ pub struct SocketAncillary<'a> {
 }
 
 impl<'a> SocketAncillary<'a> {
-    /// Create an ancillary data with the given buffer.
+    /// Create a struct for accessing an ancillary data packet in `buffer`.
+    /// This ancillary data packet will be empty.
+    /// The result won't own the ancillary data and will read them from
+    /// and write them into `buffer`. Obviously, this buffer isn't extensible.
```
the8472, line 455: Since it is obvious it could be rephrased as

> Since the buffer is not extensible it can instead be sized in advance by calculating the aggregate size of all expected control messages, (see ...

library/std/src/os/unix/net/ancillary.rs
LinkTed: Yes, this would documentation looks nice. After the merge, I will copy it for my PR #82858.

inquisitivecrystal:
It appears that LinkTed isn't a team member, and thus highfive wouldn't assign this to them. In the interest of having it assigned to someone...

r? @the8472

library/std/src/os/unix/net/ancillary.rs
```diff
@@ -407,7 +434,13 @@ pub struct SocketAncillary<'a> {
 }
 
 impl<'a> SocketAncillary<'a> {
-    /// Create an ancillary data with the given buffer.
+    /// Create a struct for accessing an ancillary data packet in `buffer`.
```
the8472, line 437: > accessing [...] in buffer

While this isn't wrong I think it could be misunderstood that the purpose of this constructor is to pass in a buffer that already contains messages and then deserializing it.

The next sentence clarifies it but I think we can still do better. Maybe something like

> Create a struct holding ancillary data. `buffer` is used to store the serialized data without dynamic allocations.

Exposing the buffer is mostly an optimization to play nice with low-level network code that uses preallocated buffers.

library/std/src/os/unix/net/ancillary.rs
```diff
@@ -369,7 +384,19 @@ impl<'a> Iterator for Messages<'a> {
     }
 }
 
-/// A Unix socket Ancillary data struct.
+/// A struct for accessing a socket ancillary data packet.
```
the8472, line 387: Same as the comment on `new`

library/std/src/os/unix/net/ancillary.rs
```diff
@@ -418,6 +451,7 @@ impl<'a> SocketAncillary<'a> {
     /// let mut ancillary_buffer = [0; 128];
     /// let mut ancillary = SocketAncillary::new(&mut ancillary_buffer[..]);
     /// ```
+    // FIXME: add size example
```
the8472, line 454: This is meant for a future PR or is this PR incomplete?

library/std/src/os/unix/net/ancillary.rs
beroal: I'm trying to avoid saying that the struct holds ancillary data because it doesn't.

library/std/src/os/unix/net/ancillary.rs
beroal: It's incomplete.

library/std/src/os/unix/net/ancillary.rs
the8472: >  Create a struct to process ancillary data. `buffer` is used to store the serialized data without dynamic allocations.

JohnCSimon:
Ping from triage: 
@beroal can you please post your status on this PR?

beroal:
> Ping from triage:
> @beroal can you please post your status on this PR?

I can't finish this PR now. `ancillary_data_size` needs examples. The form of examples depends on [whether the `const-extern-fn` feature of `libc` is stable](https://github.com/rust-lang/rust/issues/64926). If I include examples now, I'll need to send another PR with new examples when `const-extern-fn` is stabilized. Too much hassle for such a small change.

JohnCSimon:
Triage: 
@beroal I'm closing this as inactive, feel free to reopen when you're ready to continue. Thanks.

PR #88428, opened 2021-08-28
expand: Treat more macro calls as statement macro calls

petrochenkov:
This PR implements the suggestion from https://github.com/rust-lang/rust/pull/87981#issuecomment-906641052 and treats fn-like macro calls inside `StmtKind::Item` and `StmtKind::Semi` as statement macro calls, which is consistent with treatment of attribute invocations in the same positions and with token-based macro expansion model in general.

This also allows to remove a special case in `NodeId` assignment (previously tried in #87779), and to use statement `NodeId`s for linting (`assign_id!`).

r? @Aaron1011


GuillaumeGomez:
Is it the one failing in #88584?

petrochenkov:
Yes, looks like https://github.com/rust-lang/rust/pull/87779 had the same recursion limit issue.
@bors r-

PR #88437, opened 2021-08-28
Revert path cmp optimizations

the8472:
Check if reverting #86898 undoes the regression since I can't repro it locally


the8472:
@rust-timer queue

PR #88439, opened 2021-08-28
Unwinding support for inline assembly

cynecx:
r? @Amanieu

### DIFF
--- a/compiler/rustc_ast_pretty/src/pprust/state.rs
+++ b/compiler/rustc_ast_pretty/src/pprust/state.rs
@@ -2338,6 +2338,9 @@ impl<'a> State<'a> {
                 if opts.contains(InlineAsmOptions::RAW) {
                     options.push("raw");
                 }
+                if opts.contains(InlineAsmOptions::MAY_UNWIND) {
+                    options.push("may_unwind");
+                }
                 s.commasep(Inconsistent, &options, |s, &opt| {
                     s.word(opt);
                 });
--- a/compiler/rustc_codegen_cranelift/src/base.rs
+++ b/compiler/rustc_codegen_cranelift/src/base.rs
@@ -239,7 +239,8 @@ fn codegen_fn_content(fx: &mut FunctionCx<'_, '_, '_>) {
             fx.add_comment(inst, terminator_head);
         }
 
-        fx.set_debug_loc(bb_data.terminator().source_info);
+        let source_info = bb_data.terminator().source_info;
+        fx.set_debug_loc(source_info);
 
         match &bb_data.terminator().kind {
             TerminatorKind::Goto { target } => {
@@ -295,19 +296,19 @@ fn codegen_fn_content(fx: &mut FunctionCx<'_, '_, '_>) {
                         let len = codegen_operand(fx, len).load_scalar(fx);
                         let index = codegen_operand(fx, index).load_scalar(fx);
                         let location = fx
-                            .get_caller_location(bb_data.terminator().source_info.span)
+                            .get_caller_location(source_info.span)
                             .load_scalar(fx);
 
                         codegen_panic_inner(
                             fx,
                             rustc_hir::LangItem::PanicBoundsCheck,
                             &[index, len, location],
-                            bb_data.terminator().source_info.span,
+                            source_info.span,
                         );
                     }
                     _ => {
                         let msg_str = msg.description();
-                        codegen_panic(fx, msg_str, bb_data.terminator().source_info.span);
+                        codegen_panic(fx, msg_str, source_info.span);
                     }
                 }
             }
@@ -378,10 +379,18 @@ fn codegen_fn_content(fx: &mut FunctionCx<'_, '_, '_>) {
                 options,
                 destination,
                 line_spans: _,
+                cleanup,
             } => {
+                if cleanup.is_some() {
+                    fx.tcx.sess.span_fatal(
+                        source_info.span,
+                        "cranelift doesn't support unwinding from inline assembly.",
+                    );
+                }
+
                 crate::inline_asm::codegen_inline_asm(
                     fx,
-                    bb_data.terminator().source_info.span,
+                    source_info.span,
                     template,
                     operands,
                     *options,
@@ -415,7 +424,7 @@ fn codegen_fn_content(fx: &mut FunctionCx<'_, '_, '_>) {
             }
             TerminatorKind::Drop { place, target, unwind: _ } => {
                 let drop_place = codegen_place(fx, *place);
-                crate::abi::codegen_drop(fx, bb_data.terminator().source_info.span, drop_place);
+                crate::abi::codegen_drop(fx, source_info.span, drop_place);
 
                 let target_block = fx.get_block(*target);
                 fx.bcx.ins().jump(target_block, &[]);
--- a/compiler/rustc_codegen_cranelift/src/base.rs
+++ b/compiler/rustc_codegen_cranelift/src/base.rs
@@ -1,6 +1,7 @@
 //! Codegen of a single function
 
 use cranelift_codegen::binemit::{NullStackMapSink, NullTrapSink};
+use rustc_ast::InlineAsmOptions;
 use rustc_index::vec::IndexVec;
 use rustc_middle::ty::adjustment::PointerCast;
 use rustc_middle::ty::layout::FnAbiOf;
@@ -379,9 +380,9 @@ fn codegen_fn_content(fx: &mut FunctionCx<'_, '_, '_>) {
                 options,
                 destination,
                 line_spans: _,
-                cleanup,
+                cleanup: _,
             } => {
-                if cleanup.is_some() {
+                if options.contains(InlineAsmOptions::MAY_UNWIND) {
                     fx.tcx.sess.span_fatal(
                         source_info.span,
                         "cranelift doesn't support unwinding from inline assembly.",
--- a/compiler/rustc_codegen_gcc/src/asm.rs
+++ b/compiler/rustc_codegen_gcc/src/asm.rs
@@ -118,7 +118,14 @@ impl<'a, 'gcc, 'tcx> AsmBuilderMethods<'tcx> for Builder<'a, 'gcc, 'tcx> {
         true
     }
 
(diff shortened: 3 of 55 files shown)


compiler/rustc_llvm/llvm-wrapper/RustWrapper.cpp
```diff
@@ -427,11 +427,22 @@ LLVMRustInlineAsm(LLVMTypeRef Ty, char *AsmString, size_t AsmStringLen,
...
                   LLVMRustAsmDialect Dialect, LLVMBool CanThrow) {
+#if LLVM_VERSION_GE(13, 0)
   return wrap(InlineAsm::get(unwrap<FunctionType>(Ty),
                              StringRef(AsmString, AsmStringLen),
                              StringRef(Constraints, ConstraintsLen),
                              HasSideEffects, IsAlignStack,
                              fromRust(Dialect), CanThrow));
+#else
+  if (CanThrow) {
+    report_fatal_error("Shouldn't get called on older versions");
```
bjorn3, line 438: Is this reachable when trying to use unwinding inline asm on older LLVM versions? If so this should be a proper rustc error.

compiler/rustc_ast_pretty/src/pprust/state.rs
```diff
@@ -2300,6 +2300,9 @@ impl<'a> State<'a> {
                 if opts.contains(InlineAsmOptions::RAW) {
                     options.push("raw");
                 }
+                if opts.contains(InlineAsmOptions::UNWIND) {
+                    options.push("raw");
```
camelid, line 2304:
```suggestion
                    options.push("unwind");
```

compiler/rustc_llvm/llvm-wrapper/RustWrapper.cpp
cynecx: I agree.

compiler/rustc_codegen_cranelift/src/base.rs
```diff
@@ -379,7 +380,15 @@ fn codegen_fn_content(fx: &mut FunctionCx<'_, '_, '_>) {
...
+                if cleanup.is_some() {
+                    fx.tcx.sess.span_fatal(
+                        source_info.span,
+                        "cranelift doesn't support unwinding from inline assembly.",
+                    );
+                }
+
                 crate::inline_asm::codegen_inline_asm(
                     fx,
                     bb_data.terminator().source_info.span,
```
bjorn3, line 394: Can you please change this to use `source_info` too?

compiler/rustc_codegen_llvm/src/asm.rs
```diff
@@ -472,6 +476,13 @@ pub(crate) fn inline_asm_call(
         let constraints_ok = llvm::LLVMRustInlineAsmVerify(fty, cons.as_ptr().cast(), cons.len());
         debug!("constraint verification result: {:?}", constraints_ok);
         if constraints_ok {
+            if unwind && llvm_util::get_version() < (13, 0, 0) {
+                bx.cx.sess().span_fatal(
+                    line_spans[0],
+                    "the llvm backend doesn't support unwinding from inline assembly.",
```
bjorn3, line 482: Can you please indicate that this is a limitation of the specific LLVM version against which rustc is compiled?

cynecx: `unwinding from inline assembly is only supported on llvm >= 13`

does this sound ok?

bjorn3: I think so.

Amanieu:
I actually started working on a PR myself before you posted this one, but it's still very much a WIP: https://github.com/rust-lang/rust/compare/master...Amanieu:asm_unwind

More changes to MIR are needed to support this since MIR needs to be aware of unwinding. In particular `call_return_effect` needs to be adapted to also work with `TerminatorKind::InlineAsm`. This is used to mark the return value of a call as initialized only if a call returns normally but not when it unwinds. The same logic applies for output operands of `InlineAsm`.

compiler/rustc_codegen_llvm/src/asm.rs
```diff
@@ -443,9 +450,16 @@ pub(crate) fn inline_asm_call(
     alignstack: bool,
     dia: LlvmAsmDialect,
     line_spans: &[Span],
+    unwind: bool,
```
Amanieu, line 453: This `bool` is redundant with the `Option` just below it.

cynecx: I don’t think that’s enough because it’s possible we might want to emit an unwinding ‘call asm’ instruction (in that case ‘unwind’ is true and the option is ’None’.

bjorn3: For reference when there is nothing to cleanup on unwind in the current function, `cleanup` will be `None` and a regular `call` instruction will be used. `invoke` is only necessary when you want to actually run code on unwinding.

cynecx: Yes. But the inline assembly still has to be marked as “unwind” to be actually sound because eg. the llvm inlining pass needs to know that the ‘call asm’ might unwind (inlining  a ‘call asm unwind’ into an unwinding function might turn the ‘call asm unwind’ into an ‘invoke asm unwind’. The inliner isn’t performing that transformation when the ‘call asm’ isn’t marked as ‘unwind’).

Amanieu:
I updated my branch at https://github.com/rust-lang/rust/compare/master...Amanieu:asm_unwind, it should now handle all of the MIR parts that deal with unwinding.

cynecx:
@Amanieu That's great! Do you mind if I rebase this PR onto your branch?

Amanieu:
Sure! Just keep in mind I haven't actually tested any of it apart from checking it compiles.

cynecx:
@Amanieu No worries. I am going to add some tests anyway and so then we shall see :) (But that has to wait until tomorrow or the day after tomorrow).

Amanieu:
I updated my branch again, there were a few more things I missed in MIR.

Amanieu (commented):
Can you also:
- update the documentation in the unstable book to add `may_unwind`.
- double-check the interaction between `#[naked]` and `may_unwind`: does this allow you to unwind from a naked function?

cynecx:
> * update the documentation in the unstable book to add `may_unwind`.

Will do.

> * double-check the interaction between `#[naked]` and `may_unwind`: does this allow you to unwind from a naked function?

As far as I have seen from several llvm-ir/assembly dumps [1] it should work just fine.

[1] https://github.com/cynecx/libfringe/blob/asm/src/arch/x86_64.rs (your unwinding branch for libfringe basically)

Amanieu:
@cynecx Are you still working on this PR?

cynecx:
@Amanieu Yes I am. Sorry about the delay, I was caught up in some other things... Trying a rebase rn.

Amanieu:
@cynecx are there any other changes that you plan on making (this PR is still marked as a draft)?

Otherwise this LGTM. I'd like someone more familiar with MIR to review the MIR-level changes here.

r? @oli-obk

cynecx:
@Amanieu Yes, there are missing some tests. I was working on some the last day, however I got distracted again. But I should get them in 1-2 day.

Amanieu:
ping @cynecx, I'm very eager to see this land!

cynecx:
@Amanieu Sorry about the delay. I’ll try to push a rebase/test tomorrow.

compiler/rustc_codegen_gcc/src/asm.rs
```diff
@@ -118,7 +118,13 @@ impl<'a, 'gcc, 'tcx> AsmBuilderMethods<'tcx> for Builder<'a, 'gcc, 'tcx> {
         true
     }
 
-    fn codegen_inline_asm(&mut self, template: &[InlineAsmTemplatePiece], rust_operands: &[InlineAsmOperandRef<'tcx, Self>], options: InlineAsmOptions, _span: &[Span], _instance: Instance<'_>) {
+    fn codegen_inline_asm(&mut self, template: &[InlineAsmTemplatePiece], rust_operands: &[InlineAsmOperandRef<'tcx, Self>], options: InlineAsmOptions, span: &[Span], _instance: Instance<'_>, dest_catch_funclet: Option<(Self::BasicBlock, Self::BasicBlock, Option<&Self::Funclet>)>) {
+        if dest_catch_funclet.is_some() {
+            self.sess()
+                .struct_span_err(span[0], "GCC backend does not support unwinding from inline asm")
+                .emit();
```
cynecx, line 125: Note to myself so I don’t forget. This is missing an early return.

Amanieu:
By the way, the `may_unwind` option should be placed under a `asm_unwind` feature gate. See #90348.

compiler/rustc_mir_transform/src/check_unsafety.rs
```diff
@@ -207,7 +207,7 @@ impl<'a, 'tcx> Visitor<'tcx> for UnsafetyChecker<'a, 'tcx> {
                         PlaceContext::MutatingUse(
                             MutatingUseContext::Store
                                 | MutatingUseContext::Drop
-                                | MutatingUseContext::AsmOutput
```
Amanieu, line 210: Add `AsmOutput` here as well.

cynecx:
@Amanieu I've addressed your comments :)

cynecx:
@Amanieu ci passes now.

cynecx:
I've pushed a fix for https://github.com/rust-lang/rust/pull/91117#issuecomment-974956983.

cynecx:
@Amanieu rebased + test fixed.

cynecx:
@Amanieu tests "should" pass now.

src/test/ui/feature-gates/feature-gate-asm_unwind.rs
```diff
@@ -0,0 +1,8 @@
+#![feature(asm)]
```
cynecx, line 1: @Amanieu Do you think it's alright if I put `// only-x86_64` here? (since most of the feature gate tests are `only-x86_64`)

Amanieu: That's fine.

cynecx:
@Amanieu 🙏

cynecx:
@Amanieu rebased and test fixed.

PR #88445, opened 2021-08-28
Clean up the lowering of AST items

inquisitivecrystal:
This PR simplifies and improves `rustc_ast_lowering::item` in various minor ways. The reasons for the changes should mostly be self evident, though I'm happy to specifically explain anything if needed.

These changes used to be part of #88019, but I removed them after it was pointed out that some of my other changes to `rustc_ast_lowering` were unnecessary. It felt like a bad idea to clean up code which I didn't even need to touch anymore.

r? @cjgillot


inquisitivecrystal:
Rebased!

PR #88448, opened 2021-08-29
BTreeMap/BTreeSet::from_iter: use bulk building to improve the performance

xu-cheng:
Bulk building is a common technique to increase the performance of building a fresh btree map. Instead of inserting items one-by-one, we sort all the items beforehand then create the BtreeMap in bulk.

Benchmark
```
./x.py bench library/alloc --test-args btree::map::from_iter
```

* Before
```
test btree::map::from_iter_rand_100                      ... bench:       3,694 ns/iter (+/- 840)
test btree::map::from_iter_rand_10_000                   ... bench:   1,033,446 ns/iter (+/- 192,950)
test btree::map::from_iter_seq_100                       ... bench:       5,689 ns/iter (+/- 1,259)
test btree::map::from_iter_seq_10_000                    ... bench:     861,033 ns/iter (+/- 118,815)
```

* After
```
test btree::map::from_iter_rand_100                      ... bench:       3,033 ns/iter (+/- 707)
test btree::map::from_iter_rand_10_000                   ... bench:     775,958 ns/iter (+/- 105,152)
test btree::map::from_iter_seq_100                       ... bench:       2,969 ns/iter (+/- 336)
test btree::map::from_iter_seq_10_000                    ... bench:     258,292 ns/iter (+/- 29,364)
```

### DIFF
--- a/library/alloc/src/collections/btree/map.rs
+++ b/library/alloc/src/collections/btree/map.rs
@@ -1,3 +1,4 @@
+use crate::vec::Vec;
 use core::borrow::Borrow;
 use core::cmp::Ordering;
 use core::fmt::{self, Debug};
@@ -9,6 +10,7 @@ use core::ops::{Index, RangeBounds};
 use core::ptr;
 
 use super::borrow::DormantMutRef;
+use super::dedup_sorted_iter::DedupSortedIter;
 use super::navigate::{LazyLeafRange, LeafRange};
 use super::node::{self, marker, ForceResult::*, Handle, NodeRef, Root};
 use super::search::SearchResult::*;
@@ -1290,6 +1292,18 @@ impl<K, V> BTreeMap<K, V> {
     pub fn into_values(self) -> IntoValues<K, V> {
         IntoValues { inner: self.into_iter() }
     }
+
+    /// Makes a `BTreeMap` from a sorted iterator.
+    pub(crate) fn bulk_build_from_sorted_iter<I>(iter: I) -> Self
+    where
+        K: Ord,
+        I: Iterator<Item = (K, V)>,
+    {
+        let mut root = Root::new();
+        let mut length = 0;
+        root.bulk_push(DedupSortedIter::new(iter), &mut length);
+        BTreeMap { root: Some(root), length }
+    }
 }
 
 #[stable(feature = "rust1", since = "1.0.0")]
@@ -1914,9 +1928,15 @@ impl<K, V> FusedIterator for RangeMut<'_, K, V> {}
 #[stable(feature = "rust1", since = "1.0.0")]
 impl<K: Ord, V> FromIterator<(K, V)> for BTreeMap<K, V> {
     fn from_iter<T: IntoIterator<Item = (K, V)>>(iter: T) -> BTreeMap<K, V> {
-        let mut map = BTreeMap::new();
-        map.extend(iter);
-        map
+        let mut inputs: Vec<_> = iter.into_iter().collect();
+
+        if inputs.is_empty() {
+            return BTreeMap::new();
+        }
+
+        // use stable sort to preserve the insertion order.
+        inputs.sort_by(|a, b| a.0.cmp(&b.0));
+        BTreeMap::bulk_build_from_sorted_iter(inputs.into_iter())
     }
 }
 
@@ -2025,8 +2045,14 @@ impl<K: Ord, V, const N: usize> From<[(K, V); N]> for BTreeMap<K, V> {
     /// let map2: BTreeMap<_, _> = [(1, 2), (3, 4)].into();
     /// assert_eq!(map1, map2);
     /// ```
-    fn from(arr: [(K, V); N]) -> Self {
-        core::array::IntoIter::new(arr).collect()
+    fn from(mut arr: [(K, V); N]) -> Self {
+        if N == 0 {
+            return BTreeMap::new();
+        }
+
+        // use stable sort to preserve the insertion order.
+        arr.sort_by(|a, b| a.0.cmp(&b.0));
+        BTreeMap::bulk_build_from_sorted_iter(core::array::IntoIter::new(arr))
     }
 }
 
--- a/library/alloc/benches/btree/map.rs
+++ b/library/alloc/benches/btree/map.rs
@@ -54,6 +54,50 @@ macro_rules! map_insert_seq_bench {
     };
 }
 
+macro_rules! map_from_iter_rand_bench {
+    ($name: ident, $n: expr, $map: ident) => {
+        #[bench]
+        pub fn $name(b: &mut Bencher) {
+            let n: usize = $n;
+            // setup
+            let mut rng = thread_rng();
+            let mut vec = Vec::with_capacity(n);
+
+            for _ in 0..n {
+                let i = rng.gen::<usize>() % n;
+                vec.push((i, i));
+            }
+
+            // measure
+            b.iter(|| {
+                let map: $map<_, _> = vec.iter().copied().collect();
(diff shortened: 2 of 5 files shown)


library/alloc/src/collections/btree/map.rs
```diff
@@ -1914,9 +1928,15 @@ impl<K, V> FusedIterator for RangeMut<'_, K, V> {}
 #[stable(feature = "rust1", since = "1.0.0")]
 impl<K: Ord, V> FromIterator<(K, V)> for BTreeMap<K, V> {
     fn from_iter<T: IntoIterator<Item = (K, V)>>(iter: T) -> BTreeMap<K, V> {
-        let mut map = BTreeMap::new();
-        map.extend(iter);
-        map
+        let mut inputs: Vec<_> = iter.into_iter().collect();
```
the8472, line 1931: could this allocation be avoided by using a fixed-sized buffer on the stack instead and deduping/inserting in batches?

the8472: Otherwise this allocation might trigger an OOM when streaming more duplicate tuples from disk than would fit into memory.

xu-cheng: This would a rare but hard to solve edge case. I can think two possible solutions:

* Fallback to use `extend`.
```rust
let iter = iter.into_iter();
let mut inputs: Vec<_> = (&mut iter).take(MAX_BUF_SIZE).collect();
inputs.sort_by(|a, b| a.0.cmp(&b.0));
let mut map = BTreeMap::bulk_build_from_sorted_iter(inputs.into_iter());
map.extend(iter);
map
```
This would lead to suboptimal performance. Another issue is what `MAX_BUF_SIZE` should be.

* Use Merge sort like algorithm.
```rust
let iter = iter.into_iter();
let mut buf_len: usize = INIT_BUF_SIZE;
let mut left_buf: Vec<_> = (&mut iter).take(buf_len).collect();
left_buf.sort_by(|a, b| a.0.cmp(&b.0));
let mut right_buf: Vec<_> = (&mut iter).take(buf_len).collect();
right_buf.sort_by(|a, b| a.0.cmp(&b.0));

while !right_buf.is_empty() {
    left_buf = MergeIter(MergeIterInner::new(
        DedupSortedIter::new(left_buf.into_iter()),
        DedupSortedIter::new(right_buf.into_iter()),
    ))
    .collect();

    buf_len = left_buf.len(); // in normal case this would increase exponentially
    right_buf = (&mut iter).take(buf_len).collect();
    right_buf.sort_by(|a, b| a.0.cmp(&b.0));
}

// in the end, the left_buf should have the same size as all non-dup keys in iter.
BTreeMap::bulk_build_from_sorted_iter(left_buf.into_iter())
```
This solution is more complex, but should achieve better performance in normal case. In cases there are lots of duplicates, the buffer size would not increase by much. There is a minor issue that how should we choose  `INIT_BUF_SIZE`.

Comments?

xu-cheng: I just tested my second solution. The performance is not that good. The `INIT_BUF_SIZE` is set to be 64.
```
test btree::map::from_iter_rand_100                      ... bench:       5,457 ns/iter (+/- 819)
test btree::map::from_iter_rand_10_000                   ... bench:   1,057,839 ns/iter (+/- 138,483)
test btree::map::from_iter_seq_100                       ... bench:       5,660 ns/iter (+/- 739)
test btree::map::from_iter_seq_10_000                    ... bench:     672,612 ns/iter (+/- 100,751)
```

the8472: I had the following approach in mind

1. fill vec up to `MAX_BUF_SIZE`
2. sort
3. `bulk_build_from_sorted_iter`
4. if iterator is not empty goto 1

the8472: Oh wait, we don't have an extend method that is optimized for receiving a sorted batch.

Then it might be possible to improve your second approach by reusing allocations. There are too many `collect` calls.
And `INIT_BUF_SIZE` could be chosen based on the iterator's size hint.

xu-cheng: > Oh wait, we don't have an extend method that is optimized for receiving a sorted batch.

We actually have one, i.e., `BTreeMap::append`. However, the performance would be even worse as it works by rebuilding the btree from the sorted iter.

> Then it might be possible to improve your second approach by reusing allocations

The issue is that each `collect` calls have different buf length, which increases exponentially. I am not sure how you can reuse the allocations here.

the8472: > I am not sure how you can reuse the allocations here.

If I understood the approach correctly then it's not necessary for `right_buf` to grow much when there are many duplicates.

> We actually have one, i.e., BTreeMap::append. However, the performance would be even worse as it works by rebuilding the btree from the sorted iter.

Yeah, we would need something more efficient that only allocates the extra nodes needed and then redistribute entries from the map and the iterator.

----

Anyway, looking at the benchmarks it looks like you're only covering cases with few duplicates. Measuring an iterator with many duplicates might show different performance characteristics since you may end up allocating a large vec where before it would only need to insert a few elements into the map.

Mark-Simulacrum: I don't think we necessarily need to bother with the OOM condition here, we can potentially do that in a follow up PR.

Mark-Simulacrum:
@bors try @rust-timer queue 

Generally this seems OK, though I am mildly wary of the added code complexity. I don't think it's anything too worrisome though and in practice sorting is not typically that complicated.

library/alloc/src/collections/btree/map.rs
Mark-Simulacrum: > Anyway, looking at the benchmarks it looks like you're only covering cases with few duplicates. Measuring an iterator with many duplicates might show different performance characteristics since you may end up allocating a large vec where before it would only need to insert a few elements into the map.

My sense of typical allocation performance (though not really backed up by anything) is that generally the size doesn't matter *that* much -- you're either pulling from a pool for tiny sizes or likely falling back to mmap or similar for larger allocations. So I wouldn't expect much difference here.

I'd guess that the bulk loading is going to be pretty efficient and in most cases you're not building BTreeMap and similar from iterators with lots of duplicates, I suspect. If we optimize for that case we'll probably hurt the performance of the low-duplicates/no-duplicates case. In practice, we're going to allocate `length / 2` buffer for the merge sort *anyway* and 2x difference in allocation size probably won't matter that much.

Mark-Simulacrum:
It looks like the added codegen for BTree is increasing the work LLVM needs to do by a good bit -- 4.6% is fairly sizeable. While this does increase the performance of long iterators collecting into BTree's, I'm not sure it is worth the compile time cost here. That said, ultimately, it seems like the code being added by this PR isn't doing anything too interesting or unexpected, so I am generally inclined to move forward. The runtime performance improvement is pretty nice.

So, I'm going to say that this regression is justified -- we are improving the runtime performance, and the regression appears to be entirely due to more codegen.

@bors r+

pnkfelix:
As predicted, during weekly performance triage this was tagged as causing [mixed changes to rustc instruction counts](https://perf.rust-lang.org/compare.html?start=11bbb5231349a0a144d86d5c0c21061a06d1969d&end=ffaf857045f4f4d8bb563e0a5077f9b065f42916&stat=instructions:u).

Its already been appropriately labelled; I'm just adding the link above for ease of reference in future.

tanriol:
Sounds like `from_iter`'s new memory behavior should at least be documented somewhere.

PR #88450, opened 2021-08-29
fix(rustc_parse): correct span in `maybe_whole_expr!`

### LINKED ISSUE #87812: Incorrect suggestion with labelled breaks and blocks on nightly
When using blocks with labelled break statements, a warning can appear to avoid confusion. The following code:

```rust
let x = 'l: loop {
    break 'l { 3 };
};
```

produces:

```
warning: this labeled break expression is easy to confuse with an unlabeled break with a labeled value expression
  --> src/main.rs:71:9
   |
71 |         break 'l { 3 };
   |         ^^^^^^^^^^^^^^
   |
   = note: `#[warn(break_with_label_and_loop)]` on by default
help: wrap this expression in parentheses
   |
71 |         break 'l ({ 3 });
   |                  ^     ^
```

However when introducing a block via a macro:

```rust
macro_rules! foo {
    ( $f:block ) => {
        '_l: loop {
            break '_l $f;
        }
    };
}

fn main() {
    let x = foo!({ 3 });
}
```

The compiler warns:

```
warning: this labeled break expression is easy to confuse with an unlabeled break with a labeled value expression
  --> x.rs:4:13
   |
4  |             break '_l $f;
   |             ^^^^^^^^^^^^^
...
10 |     let x = foo!({ 3 });
   |             ----------- in this macro invocation
   |
   = note: `#[warn(break_with_label_and_loop)]` on by default
   = note: this warning originates in the macro `foo` (in Nightly builds, run with -Z macro-backtrace for more info)
help: wrap this expression in parentheses
   |
4  |             break '_l $f(;)
   |                         ^ ^
```
Which suggests the incorrect code `$f(;)`, whereas it should suggest `($f);`. This is only an issue on nightly, as the warning does not exist on stable.

`rustc --version --verbose`:
```
rustc 1.56.0-nightly (2f07ae408 2021-08-05)
binary: rustc
commit-hash: 2f07ae408fce782bf1058e3de808f1b6f9ab60a4
commit-date: 2021-08-05
release: 1.56.0-nightly
LLVM version: 12.0.1
```

notriddle:
Fixes #87812

### DIFF
--- a/src/test/ui/parser/issue-87812-path.rs
+++ b/src/test/ui/parser/issue-87812-path.rs
@@ -1,0 +1,11 @@
+macro_rules! foo {
+    ( $f:path ) => {{
+        let _: usize = $f; //~ERROR
+    }};
+}
+
+struct Baz;
+
+fn main() {
+    foo!(Baz);
+}
--- a/compiler/rustc_parse/src/parser/expr.rs
+++ b/compiler/rustc_parse/src/parser/expr.rs
@@ -50,7 +50,7 @@ macro_rules! maybe_whole_expr {
                     let block = block.clone();
                     $p.bump();
                     return Ok($p.mk_expr(
-                        $p.token.span,
+                        $p.prev_token.span,
                         ExprKind::Block(block, None),
                         AttrVec::new(),
                     ));
--- a/compiler/rustc_parse/src/parser/expr.rs
+++ b/compiler/rustc_parse/src/parser/expr.rs
@@ -41,7 +41,7 @@ macro_rules! maybe_whole_expr {
                     let path = path.clone();
                     $p.bump();
                     return Ok($p.mk_expr(
-                        $p.token.span,
+                        $p.prev_token.span,
                         ExprKind::Path(None, path),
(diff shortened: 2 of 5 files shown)


src/test/ui/parser/issue-87812-path.rs
```diff
@@ -0,0 +1,13 @@
+#![deny(break_with_label_and_loop)]
+
```
camelid, line 2: Why is this lint denied in this test?

notriddle: This is the lint that gets triggered. It’s `warn` by default.

camelid: But I don't see the lint being emitted in the stderr...?

notriddle: @camelid You're right. I had this test mixed up with the other one.

It's fixed now.

notriddle:
@camelid You're right. I had this test mixed up with the other one.

It's fixed now.

PR #88452, opened 2021-08-29
VecDeque: improve performance for From<[T; N]>

xu-cheng:
Create `VecDeque` directly from the array instead of inserting items one-by-one.

Benchmark
```
./x.py bench library/alloc --test-args vec_deque::bench_from_array_1000
```

* Before
```
test vec_deque::bench_from_array_1000                    ... bench:       3,991 ns/iter (+/- 717)
```

* After
```
test vec_deque::bench_from_array_1000                    ... bench:         268 ns/iter (+/- 37)
```


leonardo-m:
See also: https://users.rust-lang.org/t/b-tree-hinting/62985

joshtriplett:
r? @m-ou-se

This will probably conflict with https://github.com/rust-lang/rust/pull/88611 .

PR #88456, opened 2021-08-29
Fix: don't document private macros by default

### LINKED ISSUE #88453: Private macros are documented
### Code

I tried this code:

```rust
macro_rules! private_macro {
    () => {};
}
```

I expected to see this happen: `private_macro` is hidden from the documentation.

Instead, this happened: `private_macro` is shown in the documentation and has its own page. Private declarative macros are also shown.

### Version it worked on

It most recently worked on: rustc 1.56.0-nightly (8007b506a 2021-08-14)

### Version with regression

`rustc --version`:
```
```

inquisitivecrystal:
As part of #88019, I made it so private macros are documented in `--document-private-items` mode. Unfortunately, it appears that I also accidentally made them be documented when *not* in `--document-private-items` mode. This PR fixes that and adds a regression test.

r? @jyn514 (I hope you don't mind that I keep sending PRs your way)

Fixes #88453.

### DIFF
--- a/src/librustdoc/passes/stripper.rs
+++ b/src/librustdoc/passes/stripper.rs
@@ -40,6 +40,7 @@ impl<'a> DocFolder for Stripper<'a> {
             | clean::UnionItem(..)
             | clean::AssocConstItem(..)
             | clean::TraitAliasItem(..)
+            | clean::MacroItem(..)
             | clean::ForeignTypeItem => {
                 if i.def_id.is_local() {
                     if !self.access_levels.is_exported(i.def_id.expect_def_id()) {
@@ -70,8 +71,8 @@ impl<'a> DocFolder for Stripper<'a> {
 
             clean::ImplItem(..) => {}
 
-            // tymethods/macros have no control over privacy
-            clean::MacroItem(..) | clean::TyMethodItem(..) => {}
+            // tymethods have no control over privacy
+            clean::TyMethodItem(..) => {}
 
             // Proc-macros are always public
             clean::ProcMacroItem(..) => {}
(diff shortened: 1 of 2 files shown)


src/librustdoc/passes/stripper.rs
```diff
@@ -70,8 +71,8 @@ impl<'a> DocFolder for Stripper<'a> {
 
             clean::ImplItem(..) => {}
 
-            // tymethods/macros have no control over privacy
-            clean::MacroItem(..) | clean::TyMethodItem(..) => {}
+            // tymethods have no control over privacy
+            clean::TyMethodItem(..) => {}
 
             // Proc-macros are always public
```
jyn514, line 77: Should this also be changed for proc macros? Can you add a test? You can add a dependency using // aux-build in a UI test.

src/test/rustdoc/macro-private-not-documented.rs
```diff
@@ -0,0 +1,19 @@
+// Checks that private macros aren't documented by default. They
+// should be still be documented in `--document-private-items` mode,
+// but that's tested in `macro-document-private.rs`.
```
jyn514, line 3: nit: Can you combine this file with the test you wrote in the other PR? Rustdoc tests are really slow to run and I haven't gotten around to investigating it yet.

inquisitivecrystal: Which one? This fits best logically with `macro-document-private.rs`, but that needs `--document-private-items` and this one needs not to have that option. The only rustdoc test I added that doesn't have that option is `macro-indirect-use.rs`, but this doesn't fit very well there.

src/librustdoc/passes/stripper.rs
inquisitivecrystal: Probably not? Proc macros are always public, as far as I can tell. An import of a proc macro could still be private, but I'm not sure if that needs to be handled separately.

src/test/rustdoc/macro-private-not-documented.rs
jyn514: Ah ok, fair enough. No need to worry about it.

PR #88464, opened 2021-08-29
Clarify the intent of the CaseInsensitiveString borrowing example

sakateka:
For a beginner, there may be doubts about exactly what type of key was assumed in the `HashMap` for the example about `CaseInsensitiveString`

### DIFF
--- a/library/core/src/borrow.rs
+++ b/library/core/src/borrow.rs
@@ -142,8 +142,9 @@
 /// }
 /// ```
 ///
-/// Can `CaseInsensitiveString` implement `Borrow<str>`? It certainly can
-/// provide a reference to a string slice via its contained owned string.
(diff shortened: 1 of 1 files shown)


sakateka:
@yaahc do you think this is a good fix for the documentation?

dtolnay (commented):
I believe `HashMap<String, _>` isn't a relevant type here. HashMap lookups have signatures that take `&Q` where `K: Borrow<Q>` as shown higher up in the same documentation block. So in the context of `CaseInsensitiveString: Borrow<str>` we'd be talking about `HashMap<CaseInsensitiveString, _>` and performing lookups via an argument of type `&str`.

I think this paragraph is fine as is. The point it's making is independent of any concrete HashMap API and is just reinforcing the fact that Borrow and AsRef are purposely different traits, despite having a very similar shape.

PR #88465, opened 2021-08-29
Adding examples to docs of `std::time` module

marcospb19:
And adding missing link to `Duration` from `Instant`.

### DIFF
--- a/library/core/src/time.rs
+++ b/library/core/src/time.rs
@@ -2,14 +2,21 @@
 
 //! Temporal quantification.
 //!
-//! Example:
+//! # Examples:
+//!
+//! There are multiple ways to create a new [`Duration`]:
 //!
 //! ```
-//! use std::time::Duration;
+//! # use std::time::Duration;
+//! let five_seconds = Duration::from_secs(5);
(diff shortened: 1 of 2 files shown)


marcospb19:
@joshtriplett I found the `time` module pretty empty, so I tried adding some examples to it, but I'm not sure if these examples should not be inside of `Duration` and `Instant`.

image

marcospb19:
I also thought about changing some in [`Duration`](https://doc.rust-lang.org/nightly/std/time/struct.Duration.html):

```rust
use std::time::Duration;

let five_seconds = Duration::new(5, 0);
let five_seconds_and_five_nanos = five_seconds + Duration::new(0, 5);

assert_eq!(five_seconds_and_five_nanos.as_secs(), 5);
assert_eq!(five_seconds_and_five_nanos.subsec_nanos(), 5);

let ten_millis = Duration::from_millis(10);
```

The reason: I don't think `new` is as good as `from_*` to create a `Duration`.

I would prefer 
```rust
let five_seconds = Duration::from_secs(5);
```
Instead of
```rust
let five_seconds = Duration::new(5, 0);
```

PR #88477, opened 2021-08-30
2229: Don't move out of drop type

### LINKED ISSUE #88476: missing rust_2021_incompatible_closure_captures with manual drop and move closure
I tried this code:

```rust
#![allow(unused)]
use std::rc::Rc;

struct Foo(Rc<i32>);

impl Drop for Foo {
    fn drop(self: &mut Foo) {}
}

pub fn foo() {
    let f = Foo(Rc::new(1));
    let x = move || {
        println!("{:?}", f.0);
    };
}
```

This does not issue a warning for `rust_2021_incompatible_closure_captures`, however it fails to compile with `capture_disjoint_fields` enabled:

```rust
error[E0509]: cannot move out of type `Foo`, which implements the `Drop` trait
  --> src/lib.rs:13:13
   |
13 |     let x = move || {
   |             ^^^^^^^
   |             |
   |             cannot move out of here
   |             move out of `f.0` occurs here
14 |         println!("{:?}", f.0);
   |                          ---
   |                          |
   |                          move occurs because `f.0` has type `Rc<i32>`, which does not implement the `Copy` trait
   |                          move occurs due to use in closure
```

I would expect `rust_2021_incompatible_closure_captures` to give a suggestion to add `let _ = &f`.

* https://tiny.cc/oaaiuz/reg/stakker_async_await-0.0.1/log.txt
* https://tiny.cc/oaaiuz/reg/mobc-0.7.3/log.txt
* https://tiny.cc/oaaiuz/reg/sodium-rust-2.1.1/log.txt
* https://tiny.cc/oaaiuz/reg/glommio-0.5.1/log.txt
* https://tiny.cc/oaaiuz/reg/vapix-0.1.0/log.txt

I'm uncertain if this is a duplicate of #88444 or #88114, they seem very similar, but this case is specific to implementing Drop.

### Meta

`rustc --version --verbose`:
```
binary: rustc
commit-hash: 5eacec9ec7e112a0de1011519a57c45586d58414
commit-date: 2021-08-28
host: x86_64-apple-darwin
release: 1.56.0-nightly
LLVM version: 13.0.0
```

cc @rust-lang/wg-rfc-2229

arora-aman:
Fixes #88476

r? @nikomatsakis

### DIFF
--- a/compiler/rustc_typeck/src/check/upvar.rs
+++ b/compiler/rustc_typeck/src/check/upvar.rs
@@ -399,7 +399,14 @@ impl<'a, 'tcx> FnCtxt<'a, 'tcx> {
                 }
             };
 
+            // This restriction needs to be applied after we have handled adjustments for `move`
+            // closures. We want to make sure any adjustment that might make us move the place into
+            // the closure gets handled.
+            let (place, capture_kind) =
+                restrict_precision_for_drop_types(self, place, capture_kind, usage_span);
+
             capture_info.capture_kind = capture_kind;
+
             let capture_info = if let Some(existing) = processed.get(&place) {
                 determine_capture_info(*existing, capture_info)
             } else {
@@ -626,7 +633,7 @@ impl<'a, 'tcx> FnCtxt<'a, 'tcx> {
             self.tcx.struct_span_lint_hir(
                 lint::builtin::RUST_2021_INCOMPATIBLE_CLOSURE_CAPTURES,
                 closure_hir_id,
-                closure_head_span,
+                 closure_head_span,
                 |lint| {
                     let mut diagnostics_builder = lint.build(
                         format!(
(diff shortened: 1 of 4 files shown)


nikomatsakis:
@rfcbot fcp merge

As discussed in our [last triage meeting], this is an interesting semantic question and a tweak to the closure rules. I've written up the [my conclusion](https://hackmd.io/AC5EJng9SDStdsjG2M9cUQ?view#Niko%E2%80%99s-conclusion) in this hackmd, which also includes a number of other details. I wanted to do an FCP to get ✅ from the @rust-lang/lang team that they've seen this change. There is some urgency here owing to the Rust 2021 edition timeline.

compiler/rustc_typeck/src/check/upvar.rs
```diff
@@ -626,7 +633,7 @@ impl<'a, 'tcx> FnCtxt<'a, 'tcx> {
             self.tcx.struct_span_lint_hir(
                 lint::builtin::RUST_2021_INCOMPATIBLE_CLOSURE_CAPTURES,
                 closure_hir_id,
-                closure_head_span,
+                 closure_head_span,
```
nbdd0121, line 636: I am a bit surprised that rustfmt isn't complaining about this extra space.

arora-aman: I'm surprised `./x.py fmt` made this change

camelid:
@arora-aman Not a big deal at all, but was it intentional that you renamed this PR from `2229: Handle drop_defined_for_type during migration` to `2229: Don't move out of drop type 3d7c237`?

arora-aman:
> @arora-aman Not a big deal at all, but was it intentional that you renamed this PR from `2229: Handle drop_defined_for_type during migration` to `2229: Don't move out of drop type 3d7c237`?

wwops, the commit didn't need to be there. But yes we changed how we handle the issue -- so I updated the title

camelid:
(Also just noting that I think this will need to be backported to beta now that the branch has happened.)

scottmcm:
I was convinced in the triage meeting that there are already enough consequences of `Drop` and `Copy` that it's not weird for things to be trait-dependent here.  And since it only comes up in `move` closures, being more aggressive about moving larger things there seems reasonable to me.

@rfcbot reviewed

nikomatsakis:
@bors r+ p=1

Rust 2021 related

apiraino:
Beta backport accepted as per compiler team [on Zulip](https://rust-lang.zulipchat.com/#narrow/stream/238009-t-compiler.2Fmeetings/topic/.5Bweekly.5D.202021-09-09.20.2354818/near/252633238)

@rustbot label +beta-accepted

PR #88481, opened 2021-08-30
Remove some feature gates

bjorn3:
The first commit removes various feature gates that are unused. The second commit replaces some `Fn` implementations with `Iterator` implementations, which is much cleaner IMO. The third commit replaces an unboxed_closures feature gate with min_specialization. For some reason the unboxed_closures feature gate suppresses the min_specialization feature gate from triggering on an `TrustedStep` impl. The last comment just turns a regular comment into a doc comment as drive by cleanup. I can move it to a separate PR if preferred.


compiler/rustc_middle/src/ty/codec.rs
```diff
@@ -1,11 +1,10 @@
...
-// The functionality in here is shared between persisting to crate metadata and
-// persisting to incr. comp. caches.
-
+/// This module contains some shared code for encoding and decoding various
+/// things from the `ty` module, and in particular implements support for
+/// "shorthands" which allow to have pointers back into the already encoded
+/// stream instead of re-encoding the same thing twice.
+///
+/// The functionality in here is shared between persisting to crate metadata and
+/// persisting to incr. comp. caches.
```
hellow554, line 7: Shouldn't this be `//!` ?

compiler/rustc_index/src/vec.rs
```diff
@@ -817,34 +815,39 @@ impl<'a, I: Idx, T> IntoIterator for &'a mut IndexVec<I, T> {
...
+pub struct Enumerated<I: Idx, It: Iterator> {
+    iter: iter::Enumerate<It>,
     _marker: PhantomData<fn(&I)>,
 }
-impl<I: Idx, T> FnOnce<((usize, T),)> for IntoIdx<I> {
-    type Output = (I, T);
 
-    extern "rust-call" fn call_once(self, ((n, t),): ((usize, T),)) -> Self::Output {
-        (I::new(n), t)
+impl<I: Idx, It: Iterator> Iterator for Enumerated<I, It> {
```
cjgillot, line 823: Does this implementation allow the same kind of optimizations by specializations as the earlier one?
Should we rather make `IntoIdx` a function and use RPIT?

bjorn3: > Does this implementation allow the same kind of optimizations by specializations as the earlier one?

Probably not.

> Should we rather make IntoIdx a function and use RPIT?

I will try this.

bjorn3: Done

cjgillot:
r=me once CI passes.

PR #88269, opened 2021-08-23
Doctest persist full binaries when persisting

prconrad:
Tested by adding an extra debug to echo the whole compiler line. Trimmed significantly:

Persisted but not running -> full compile so we get binaries (new behavior).
```
$ rustdoc -Zunstable-options --test --persist-doctests doctests --no-run --extern t=libt.rlib t.rs

DEBUG rustdoc::doctest run_test compiler "rustc" "--crate-type" "bin" "--edition" "2015" "-o" "doctests/t_rs_8_0/rust_out" "--extern" "t=libt.rlib" "-Ccodegen-units=1" "-Z" "unstable-options" "--target" "x86_64-unknown-linux-gnu" "--color" "always"
DEBUG rustdoc::doctest run_test compiler "rustc" "--crate-type" "bin" "--edition" "2015" "-o" "doctests/t_rs_2_0/rust_out" "--extern" "t=libt.rlib" "-Ccodegen-units=1" "-Z" "unstable-options" "--target" "x86_64-unknown-linux-gnu" "--color" "always"
test t.rs - foople (line 2) - compile ... ok
test t.rs - florp (line 8) - compile ... ok
```

Persisted and running -> full compile. 
```
$ rustdoc -Zunstable-options --test --persist-doctests doctests --extern t=libt.rlib t.rs

DEBUG rustdoc::doctest run_test compiler "rustc" "--crate-type" "bin" "--edition" "2015" "-o" "doctests/t_rs_8_0/rust_out" "--extern" "t=libt.rlib" "-Ccodegen-units=1" "-Z" "unstable-options" "--target" "x86_64-unknown-linux-gnu" "--color" "always"
DEBUG rustdoc::doctest run_test compiler "rustc" "--crate-type" "bin" "--edition" "2015" "-o" "doctests/t_rs_2_0/rust_out" "--extern" "t=libt.rlib" "-Ccodegen-units=1" "-Z" "unstable-options" "--target" "x86_64-unknown-linux-gnu" "--color" "always"

```

Running but not persisted -> full compile only
```
$ rustdoc --test --extern t=libt.rlib t.rs

DEBUG rustdoc::doctest run_test compiler "rustc" "--crate-type" "bin" "--edition" "2015" "-o" "/tmp/rustdoctestixWAUI/rust_out" "--extern" "t=libt.rlib" "-Ccodegen-units=1" "--target" "x86_64-unknown-linux-gnu" "--color" "always"
[...]

PR #88482, opened 2021-08-30
Add regression test for a spurious import

### LINKED ISSUE #78744: Non existing function imported when compiling for wasm
Whenever I try to compile  some specific code for `wasm32-unknown-unknown` with enabled `overflow-checks` and `lto` the compiler inserts a function import into the resulting wasm module that looks like this:

```wat
(import "env" "_ZN4core9panicking5panic17h80a3410ec4f43255E" (func $_ZN4core9panicking5panic17h80a3410ec4f43255E (type 0))
```

These are the complete contents of the `lib.rs`:

```rust
#![no_std]

#[panic_handler]
fn my_panic(_info: &core::panic::PanicInfo) -> ! {
    loop {}
}

#[no_mangle]
pub fn multer(a: i128, b: i128) -> i128 {
    // trigger usage of the __multi3 compiler intrinsic which seems to be necessary
    // in order to reproduce the bug
    a * b
}
```

What seems to trigger the bug are the following conditions:
* The `--target` must be `wasm32-unknown-unknown`
* `overflow-checks = true`
* `lto = true`
* Recompile core libs with cargo xbuild or the std aware cargo
* The call of the panic function originates in some compiler extrinsic

The [minimal reproducer repo](https://github.com/athei/mult-panic) contains the exact command line that reproduces the bug.

I expected to see this happen: 
The resulting wasm module does **not** import any function from the environment.

Instead, this happened:
The resulting wasm module imports the previously mentioned panic function from the environment.

### Meta
```
rustc 1.49.0-nightly (ffa2e7ae8 2020-10-24)
binary: rustc
commit-hash: ffa2e7ae8fbf9badc035740db949b9dae271c29f
commit-date: 2020-10-24
host: x86_64-apple-darwin
release: 1.49.0-nightly
LLVM version: 11.0
```

athei:
This PR adds a test that verifies that the bug described in the linked issue does not creep back into the code. In essence it checks that compiling some specific code (that uses 128 bit multiplication) with a specific set of compiler options does not lead to a spurious import of a panic function.

I noticed that other wasm tests use `# only-wasm32-bare` in their `Makefile`. This will skip the test for me. I did not find out how to run this test locally. Maybe someone can help.

closes #78744
r? @jyn514

### DIFF
--- a/src/test/run-make/wasm-spurious-import/Makefile
+++ b/src/test/run-make/wasm-spurious-import/Makefile
@@ -1,0 +1,7 @@
+-include ../../run-make-fulldeps/tools.mk
+
+# only-wasm32-bare
+
+all:
+	$(RUSTC) main.rs -C overflow-checks=yes -C panic=abort -C lto -C opt-level=z --target wasm32-unknown-unknown
+	$(NODE) verify.js $(TMPDIR)/main.wasm
--- a/src/test/run-make/wasm-spurious-import/main.rs
+++ b/src/test/run-make/wasm-spurious-import/main.rs
@@ -1,0 +1,14 @@
+#![crate_type = "cdylib"]
+#![no_std]
+
+#[panic_handler]
+fn my_panic(_info: &core::panic::PanicInfo) -> ! {
+    loop {}
+}
+
+#[no_mangle]
+pub fn multer(a: i128, b: i128) -> i128 {
+    // Trigger usage of the __multi3 compiler intrinsic which then leads to an imported
+    // panic function in case of a bug. We verify that no imports exist in our verifier.
+    a * b
+}
--- a/src/test/run-make/wasm-spurious-import/verify.js
+++ b/src/test/run-make/wasm-spurious-import/verify.js
@@ -1,0 +1,9 @@
+const fs = require('fs');
+const process = require('process');
+const assert = require('assert');
+const buffer = fs.readFileSync(process.argv[2]);
+
+let m = new WebAssembly.Module(buffer);
+let imports = WebAssembly.Module.imports(m);
+console.log('imports', imports);
+assert.strictEqual(imports.length, 0);


jyn514 (commented):
Hmm, this makes me nervous that it will fail the same way https://github.com/rust-lang/rust/pull/88269 did. Could you maybe try adding wasm to the targets in config.toml so you can test locally?

athei:
> Hmm, this makes me nervous that it will fail the same way #88269 did.

I tried to understand what you mean by reading the linked issue but I don't get it. Sorry I am a complete noob when it comes to the rustc code base.

> Could you maybe try adding wasm to the targets in config.toml so you can test locally?

This is the hint I was looking for. I guess I need to change this line?:
```toml
# Which triples to build libraries (core/alloc/std/test/proc_macro) for. Each of
# these triples will be bootstrapped from the build triple themselves.
#
# Defaults to `host`. If you set this explicitly, you likely want to add all
# host triples to this list as well in order for those host toolchains to be
# able to compile programs for their native target.
#target = ["x86_64-unknown-linux-gnu"] (as an example)
```

jyn514:
> I guess I need to change this line?

Yes, that looks right.

jyn514:
@the8472 suggested running `./src/ci/docker/run.sh wasm32`, that might also work.

athei:
> @the8472 suggested running `./src/ci/docker/run.sh wasm32`, that might also work.

Tried that one. It ran for hours. It ignored the wasm test :(
```
test [run-make] run-make/wasm-abi ... ignored
test [run-make] run-make/wasm-custom-section ... ignored
test [run-make] run-make/wasm-custom-sections-opt ... ignored
test [run-make] run-make/wasm-export-all-symbols ... ignored
test [run-make] run-make/wasm-import-module ... ignored
test [run-make] run-make/wasm-panic-small ... ignored
test [run-make] run-make/wasm-spurious-import ... ignored
test [run-make] run-make/wasm-stringify-ints-small ... ok
test [run-make] run-make/wasm-symbols-different-module ... ignored
test [run-make] run-make/wasm-symbols-not-exported ... ignored
test [run-make] run-make/wasm-symbols-not-imported ... ignored
```

I also tried running this
```
./x.py test src/test/run-make/wasm-spurious-import 
```

with this `config.toml`
```toml
# Includes one of the default files in src/bootstrap/defaults
profile = "compiler"
changelog-seen = 2

[rust]
lld = true

[build]
target = ["wasm32-unknown-unknown"]
```

It can't build lld:
```
Submodules updated in 0.02 seconds
    Finished dev [unoptimized + debuginfo] target(s) in 0.39s
Building stage0 std artifacts (x86_64-apple-darwin -> x86_64-apple-darwin)
    Finished release [optimized] target(s) in 0.16s
Copying stage0 std from stage0 (x86_64-apple-darwin -> x86_64-apple-darwin / x86_64-apple-darwin)
Building stage0 compiler artifacts (x86_64-apple-darwin -> x86_64-apple-darwin)
    Finished release [optimized] target(s) in 0.39s
Copying stage0 rustc from stage0 (x86_64-apple-darwin -> x86_64-apple-darwin / x86_64-apple-darwin)
Building LLD for x86_64-apple-darwin
running: "cmake" "/Users/alexander/Developer/rust/src/llvm-project/lld" "-G" "Ninja" "-DCMAKE_INSTALL_MESSAGE=LAZY" "-DCMAKE_C_COMPILER=cc" "-DCMAKE_CXX_COMPILER=c++" "-DCMAKE_ASM_COMPILER=cc" "-DCMAKE_C_FLAGS=-ffunction-sections -fdata-sections -fPIC -m64 -arch x86_64 -stdlib=libc++" "-DCMAKE_CXX_FLAGS=-ffunction-sections -fdata-sections -fPIC -m64 -arch x86_64 -stdlib=libc++" "-DLLVM_CONFIG_PATH=/Users/alexander/Developer/rust/build/bootstrap/debug/llvm-config-wrapper" "-DLLVM_INCLUDE_TESTS=OFF" "-DCMAKE_CXX_STANDARD=14" "-DCMAKE_INSTALL_PREFIX=/Users/alexander/Developer/rust/build/x86_64-apple-darwin/lld" "-DCMAKE_ASM_FLAGS= -ffunction-sections -fdata-sections -fPIC -m64 -arch x86_64" "-DCMAKE_BUILD_TYPE=Release"
CMake Error at CMakeLists.txt:41 (message):
  LLVMConfig.cmake not found


-- Configuring incomplete, errors occurred!
See also "/Users/alexander/Developer/rust/build/x86_64-apple-darwin/lld/build/CMakeFiles/CMakeOutput.log".
thread 'main' panicked at '
command did not execute successfully, got: exit status: 1

build script failed, must exit now', /Users/alexander/.cargo/registry/src/github.com-1ecc6299db9ec823/cmake-0.1.44/src/lib.rs:885:5
	finished in 0.100 seconds
Build completed unsuccessfully in 0:00:02
```

athei:
> > LLVMConfig.cmake not found
> 
> You'll have to install additional dependencies.
> 
> https://rustc-dev-guide.rust-lang.org/building/prerequisites.html#dependencies

I can't find the information how to use a OS installed LLVM. I tried this way:

```
PATH=/usr/local/opt/llvm/bin:$PATH ./x.py test src/test/run-make/wasm-spurious-import
```

But this still yields the `LLVMConfig.cmake not found` error. This comes from cmake and I think I need to maybe specify that in another way.

Another thing that puzzles me is that in the `config.toml` it is specified that llvm should be downloaded from CI. Why do I need to do this then?

the8472:
Oh you're on osx. I don't know if the dependencies are different there. Normally the build tools equivalent to what you can find in the docker file will do the job. If llvm build acts up then removing the build directory can help too.

Anyway, I checked out the commit, ran `./x.py test --stage 1 --target wasm32-unknown-unknown src/test/run-make/wasm-spurious-import` and it passed.

Since that was @jyn514's main concern

@bors r+

athei:
I think my issues mainly stemmed from using macOS. Thanks for running the test for me.

athei:
@the8472 Thanks for helping me out here. I find it concerning that the docker run did skip the test. That should be independent of macOS. Do you have an explanation for that?

the8472:
Ah sorry, that was the wrong job. That CI job tests `wasm32-unknown-emscripten`. `wasm32-unknown-unknown` is tested by the `test-various` job.

PR #88483, opened 2021-08-30
Fix LLVM libunwind build for non-musl targets

jethrogb:
Broken in #85600. AFAICT, [only musl, mingw, and wasm](https://github.com/rust-lang/rust/blob/673d0db5e393e9c64897005b470bfeb6d5aec61b/compiler/rustc_target/src/spec/crt_objects.rs#L128-L132) should use the “self-contained” logic in rustbuild.

### DIFF
--- a/src/bootstrap/compile.rs
+++ b/src/bootstrap/compile.rs
@@ -142,6 +142,14 @@ fn copy_and_stamp(
     target_deps.push((target, dependency_type));
 }
 
+fn copy_llvm_libunwind(builder: &Builder<'_>, target: TargetSelection, libdir: &Path) -> PathBuf {
+    let libunwind_path = builder.ensure(native::Libunwind { target });
+    let libunwind_source = libunwind_path.join("libunwind.a");
+    let libunwind_target = libdir.join("libunwind.a");
+    builder.copy(&libunwind_source, &libunwind_target);
+    libunwind_target
+}
+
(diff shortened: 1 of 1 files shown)


petrochenkov:
>only musl, mingw, and wasm should use the “self-contained” logic in rustbuild.

This was the initial set, but others could be added as well when reasonable.

For example, does fortanix-sgx links to some kind of libc?
Or can it link to some "system" libunwind instead of the shipped one?

jethrogb:
> does fortanix-sgx links to some kind of libc?
No

> Or can it link to some "system" libunwind instead of the shipped one?
Well I suppose you could always swap it out for something you build yourself if you provide the right interface? The only things required to build & link for this target is distributed by the Rust project, there are no other build dependencies.

petrochenkov:
I see.
So, the main alternative is to indeed put `libunwind.a` to the `self-contained` directory and turn the `-Cself-contained` flag to true by default for fortanix-sgx targets (like https://github.com/rust-lang/rust/pull/85600#issuecomment-908198777 suggests), so that the shipped libraries can be easily disabled.

The choice probably doesn't matters from practical point of view, fortanix-sgx and fuchsia are targets for which we can change details like this at any moment, so it's mostly a matter of internal consistency right now.

jethrogb:
How to move forward on this? Nightly builds for this target have been broken for the better part of a week.

PR #88484, opened 2021-08-30
Avoid ThinVec inside rustc_ast_lowering

bjorn3:
This avoids unnecessary allocations


nagisa:
Might be worth it to also explore use of a `SmallVec`?

est31:
@bjorn3 in which way does it avoid unneccessary allocations? `ThinVec::new` does not allocate just as `Vec::new` doesn't.

bjorn3:
If there are attributes `ThinVec` will allocate a `Vec` **and** a `Box` to wrap the `Vec`. Doesn't seem like this had any perf effect though.

PR #88493, opened 2021-08-30
Fix #88256 remove duplicated diagnostics

### LINKED ISSUE #88256: Unnecessary parentheses warning is printed twice
This code:

```rust
pub fn repro() {
    if let Some(_a) = foo() {
        let _b = (foo());
    }
}

fn foo() -> Option<i32> { None }
```

Gives unnecessary parentheses warning twice:

```
warning: unnecessary parentheses around assigned value
 --> interpreter\src\lib.rs:3:18
  |
3 |         let _b = (foo());
  |                  ^^^^^^^ help: remove these parentheses
  |
  = note: `#[warn(unused_parens)]` on by default

warning: unnecessary parentheses around assigned value
 --> interpreter\src\lib.rs:3:18
  |
3 |         let _b = (foo());
  |                  ^^^^^^^ help: remove these parentheses

warning: 2 warnings emitted
```

Reproduced with stable 1.54.0 and nightly 2021-08-22 af140757b4cb1a60d107 on playground.

[Playground link](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=0e09d8ccc2e03b877f205ffb16641f27)

chenyukang:
Fix #88256

### DIFF
--- a/compiler/rustc_errors/src/diagnostic.rs
+++ b/compiler/rustc_errors/src/diagnostic.rs
@@ -9,9 +9,10 @@ use rustc_lint_defs::Applicability;
 use rustc_serialize::json::Json;
 use rustc_span::{MultiSpan, Span, DUMMY_SP};
 use std::fmt;
+use std::hash::{Hash, Hasher};
 
 #[must_use]
-#[derive(Clone, Debug, PartialEq, Hash, Encodable, Decodable)]
+#[derive(Clone, Debug, Encodable, Decodable)]
 pub struct Diagnostic {
     pub level: Level,
     pub message: Vec<(String, Style)>,
@@ -24,6 +25,10 @@ pub struct Diagnostic {
     /// as a sort key to sort a buffer of diagnostics.  By default, it is the primary span of
     /// `span` if there is one.  Otherwise, it is `DUMMY_SP`.
     pub sort_span: Span,
+
+    /// If diagnostic is from Lint, custom hash function ignores notes
+    /// otherwise hash is based on the all the fields
+    pub is_lint: bool,
 }
 
 #[derive(Clone, Debug, PartialEq, Eq, Hash, Encodable, Decodable)]
@@ -91,6 +96,7 @@ impl Diagnostic {
             children: vec![],
             suggestions: vec![],
             sort_span: DUMMY_SP,
+            is_lint: false,
         }
     }
 
@@ -558,6 +564,11 @@ impl Diagnostic {
         self
     }
 
+    pub fn set_is_lint(&mut self) -> &mut Self {
+        self.is_lint = true;
+        self
+    }
+
     pub fn code(&mut self, s: DiagnosticId) -> &mut Self {
         self.code = Some(s);
         self
@@ -616,6 +627,42 @@ impl Diagnostic {
     ) {
         let sub = SubDiagnostic { level, message, span, render_span };
         self.children.push(sub);
+    }
+
+    /// Fields used for Hash, and PartialEq trait
+    fn keys(
+        &self,
+    ) -> (
+        &Level,
+        &Vec<(String, Style)>,
+        &Option<DiagnosticId>,
+        &MultiSpan,
+        &Vec<CodeSuggestion>,
+        Option<&Vec<SubDiagnostic>>,
+    ) {
+        (
+            &self.level,
+            &self.message,
+            &self.code,
+            &self.span,
+            &self.suggestions,
+            (if self.is_lint { None } else { Some(&self.children) }),
+        )
+    }
+}
+
+impl Hash for Diagnostic {
+    fn hash<H>(&self, state: &mut H)
+    where
+        H: Hasher,
+    {
+        self.keys().hash(state);
+    }
+}
+
+impl PartialEq for Diagnostic {
+    fn eq(&self, other: &Self) -> bool {
+        self.keys() == other.keys()
     }
 }
 
--- a/compiler/rustc_errors/src/diagnostic_builder.rs
+++ b/compiler/rustc_errors/src/diagnostic_builder.rs
@@ -242,6 +242,7 @@ impl<'a> DiagnosticBuilder<'a> {
         sp: S,
         msg: &str,
     ) -> &mut Self);
+    forward!(pub fn set_is_lint(&mut self,) -> &mut Self);
 
     /// See [`Diagnostic::multipart_suggestion()`].
     pub fn multipart_suggestion(
--- a/compiler/rustc_middle/src/lint.rs
+++ b/compiler/rustc_middle/src/lint.rs
@@ -192,6 +192,7 @@ impl<'a> LintDiagnosticBuilder<'a> {
     /// Return the inner DiagnosticBuilder, first setting the primary message to `msg`.
     pub fn build(mut self, msg: &str) -> DiagnosticBuilder<'a> {
         self.0.set_primary_message(msg);
+        self.0.set_is_lint();
         self.0
     }
 
--- a/src/test/ui/lint/issue-70819-dont-override-forbid-in-same-scope.stderr
+++ b/src/test/ui/lint/issue-70819-dont-override-forbid-in-same-scope.stderr
@@ -14,16 +14,5 @@ LL | #![forbid(forbidden_lint_groups)]
    = warning: this was previously accepted by the compiler but is being phased out; it will become a hard error in a future release!
    = note: for more information, see issue #81670 <https://github.com/rust-lang/rust/issues/81670>
 
-error: deny(unused) incompatible with previous forbid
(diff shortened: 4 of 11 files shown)


src/test/ui/lint/outer-forbid.stderr
```diff
@@ -28,26 +28,14 @@ LL | #[allow(unused)]
...
-  --> $DIR/outer-forbid.rs:20:9
-   |
-LL | #![forbid(unused, non_snake_case)]
-   |           ------ `forbid` level set here
-...
-LL | #[allow(unused_variables)]
-   |         ^^^^^^^^^^^^^^^^ overruled by previous forbid
-   |
-   = warning: this was previously accepted by the compiler but is being phased out; it will become a hard error in a future release!
-   = note: for more information, see issue #81670 <https://github.com/rust-lang/rust/issues/81670>
```
estebank, line 49: This seems like a regression to me, the primary span is different for this one, but matches the one in line 26.

src/test/ui/lint/issue-70819-dont-override-forbid-in-same-scope.stderr
```diff
@@ -14,16 +14,5 @@ LL | #![forbid(forbidden_lint_groups)]
...
-error: deny(unused) incompatible with previous forbid
-  --> $DIR/issue-70819-dont-override-forbid-in-same-scope.rs:21:13
-   |
-LL |     #![forbid(unused)]
-   |               ------ `forbid` level set here
-LL |     #![deny(unused)]
-   |             ^^^^^^ overruled by previous forbid
-   |
-   = warning: this was previously accepted by the compiler but is being phased out; it will become a hard error in a future release!
-   = note: for more information, see issue #81670 <https://github.com/rust-lang/rust/issues/81670>
```
estebank, line 26: Looking at this case, it seems to me that the difference is that the first lint is showing the `note` for "lint level defined here" while this one isn't, _and_ the `self.span` might not match the diagnostic's "primary span". An option to avoid issues to special case the lints `hash` calculation, and possibly include the first subdiagnostic in the calculation, that way this one wouldn't disappear, and regular errors wouldn't ever deduplicate either (we have more control over those already).

estebank (commented):
There's one partial regression and I have a theory for how we could avoid it. Could you try it out?

src/test/ui/lint/issue-70819-dont-override-forbid-in-same-scope.stderr
chenyukang: Do you mean we add the first sub-diagnostic if possible?

```rust
    /// Fields used for Hash, and PartialEq trait
    fn key(
        &self,
    ) -> (&Level, &Vec<(String, Style)>, &Option<DiagnosticId>, &MultiSpan, &Vec<CodeSuggestion>)
    {
        (&self.level, &self.message, &self.code, &self.span, &self.suggestions, &self.children.first_or_none  //pesucode
    }
}

impl Hash for Diagnostic {
    fn hash<H>(&self, state: &mut H)
    where
        H: Hasher,
    {
        self.key().hash(state);
    }
}
```

I think we can not remove the duplicated later diagnostic in the case https://github.com/rust-lang/rust/issues/88256.
Since the first diagnostic contains one sub-diagnostic and the second one doesn't have a sub-diagnostic, then we can not avoid the second one because of different hash values.

chenyukang: whether we should try to merge those sub diagnostic, since their spans are the same, show in different lints only with different notes seems redundant.

estebank: I might be wrong, but I think that in all cases there will be at least one subdiagnostic, with the info shown on the first "window" of the error, and using it as part of the deduplication should still work for the case you're trying to tackle, because they only differ on the second subdiagnostic (the note).

chenyukang: For the case I'm tracking, there second one doesn't have any subdiagnostics, 
Details are here: https://github.com/rust-lang/rust/issues/88256#issuecomment-908368823.

When the second one is added, this line of code calculated 'refresh' false, so it does not have a note: https://github.com/rust-lang/rust/blob/master/compiler/rustc_session/src/session.rs#L507

src/test/ui/lint/outer-forbid.stderr
chenyukang: The diagnostic's primary spans are same, but only with different notes.

src/test/ui/lint/issue-70819-dont-override-forbid-in-same-scope.stderr
chenyukang: For this case, currently rustc show the error like this:
image

And we change it to this:
image

As we can see, it's more cleaner, the second diagnostic does not provide any more info but only lacks one note. 
Any other reason we must emit it?

src/test/ui/lint/outer-forbid.stderr
estebank: Ah! You're right. I didn't expand the whole diff to see that there was a prior lint at line 20, confusing it with the one on line 30.

src/test/ui/lint/issue-70819-dont-override-forbid-in-same-scope.stderr
estebank: I think that the least likely course of action here would be to calculate the hash key using _every_ field and subdiagnostic, except for `note`s with the text "the lint level is defined here". That way we are special casing lints while leaving every other error the same. If we do that, we should add a comment to the test files being affected by this pointing at where that deduplication is being performed, so that if a future change modifies the `note` text, the reviewer/contributor will have a chance to figure out how to fix the deduplication. The other (cleaner) alternative is to modify the `Diagnostic` to have a field telling us whether it is from a lint, and if it is, ignore notes.

chenyukang: The first solution is not good, since we need to manually record which notes are ignored.
As for the second solution, except from lint, `Diagnostic` may have any other sources？
Seems for lint, `Diagnostics` is build with: 

```rust
let mut err = lint.build(&span_msg);
...
err.emit()
```

estebank: Yeah, for the second solution you only need to modify how `Diagnostic` is built in `build`.

chenyukang: How about this? 
https://github.com/rust-lang/rust/pull/88493/commits/e1899bbff970283899f797cb14c95049eae3f1b9

chenyukang: Make it cleaner, https://github.com/rust-lang/rust/pull/88493/commits/9b8741971c995e0666c07a4c20b4564941c32e06

chenyukang: @estebank Test cases are fixed.

compiler/rustc_errors/src/diagnostic.rs
```diff
@@ -558,6 +564,11 @@ impl Diagnostic {
         self
     }
 
+    pub fn set_is_lint(&mut self, is_lint_diagnostic: bool) -> &mut Self {
+        self.is_lint = is_lint_diagnostic;
+        self
+    }
```
estebank, line 570: This is only being called to set it to `true`.

```suggestion
    pub fn set_is_lint(&mut self) -> &mut Self {
        self.is_lint = true;
        self
    }
```

compiler/rustc_errors/src/diagnostic_builder.rs
```diff
@@ -242,6 +242,7 @@ impl<'a> DiagnosticBuilder<'a> {
         sp: S,
         msg: &str,
     ) -> &mut Self);
+    forward!(pub fn set_is_lint(&mut self, is_lint: bool) -> &mut Self);
```
estebank, line 245:
```suggestion
    forward!(pub fn set_is_lint(&mut self) -> &mut Self);
```

compiler/rustc_middle/src/lint.rs
```diff
@@ -192,6 +192,7 @@ impl<'a> LintDiagnosticBuilder<'a> {
     /// Return the inner DiagnosticBuilder, first setting the primary message to `msg`.
     pub fn build(mut self, msg: &str) -> DiagnosticBuilder<'a> {
         self.0.set_primary_message(msg);
+        self.0.set_is_lint(true);
```
estebank, line 195:
```suggestion
        self.0.set_is_lint();
```

estebank (approved):
Could you squash all the commits?

r=me after addressing the comments.

PR #88496, opened 2021-08-30
Fix prelude collision lint suggestion for generics with lifetimes

### LINKED ISSUE #88470: rust_2021_prelude_collisions bad suggestion with lifetime generics
I tried this code:

```rust
#![warn(rust_2021_prelude_collisions)]
pub struct PyAny;

pub trait PyTryFrom<'v>: Sized {
    fn try_from<V>(value: V) -> Result<&'v Self, ()>;
}

pub trait PyTryInto<T>: Sized {
    fn try_into(&self) -> Result<&T, ()>;
}

struct Foo;

impl<U> PyTryInto<U> for Foo
where
    U: for<'v> PyTryFrom<'v>,
{
    fn try_into(&self) -> Result<&U, ()> {
        U::try_from(self)
    }
}
```

This generates a suggestion to avoid the prelude collision with:

```rust
        <U as PyTryFrom<_>>::try_from(self)
```

However, that does not compile because the generics for `PyTryFrom` is a lifetime parameter which should not be specified with `_`.  I believe the correct suggestion would be `<U as PyTryFrom>::try_from(self)` without the generic args.

* https://crater-reports.s3.amazonaws.com/pr-87190-3/try%23a7a572ce3edd6d476191fbfe92c9c1986e009b34/reg/pyo3-0.14.3/log.txt

### Meta

`rustc --version --verbose`:
```
rustc 1.56.0-nightly (ac50a5335 2021-08-27)
binary: rustc
commit-hash: ac50a53359328a5d7f2f558833e63d59d372e4f7
commit-date: 2021-08-27
host: x86_64-apple-darwin
release: 1.56.0-nightly
LLVM version: 13.0.0
```

m-ou-se:
Fixes https://github.com/rust-lang/rust/issues/88470

cc @nikomatsakis

### DIFF
--- a/compiler/rustc_typeck/src/check/method/prelude2021.rs
+++ b/compiler/rustc_typeck/src/check/method/prelude2021.rs
@@ -239,16 +239,23 @@ impl<'a, 'tcx> FnCtxt<'a, 'tcx> {
             let trait_path = self.trait_path_or_bare_name(span, expr_id, pick.item.container.id());
             let trait_generics = self.tcx.generics_of(pick.item.container.id());
 
-            let parameter_count = trait_generics.count() - (trait_generics.has_self as usize);
-            let trait_name = if parameter_count == 0 {
-                trait_path
-            } else {
-                format!(
-                    "{}<{}>",
-                    trait_path,
-                    std::iter::repeat("_").take(parameter_count).collect::<Vec<_>>().join(", ")
-                )
-            };
+            let trait_name =
+                if trait_generics.params.len() <= trait_generics.has_self as usize {
+                    trait_path
+                } else {
+                    let counts = trait_generics.own_counts();
+                    format!(
+                        "{}<{}>",
+                        trait_path,
+                        std::iter::repeat("'_")
+                            .take(counts.lifetimes)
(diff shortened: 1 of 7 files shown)


compiler/rustc_typeck/src/check/method/prelude2021.rs
```diff
@@ -239,47 +239,58 @@ impl<'a, 'tcx> FnCtxt<'a, 'tcx> {
...
-                    "{}<{}>",
-                    trait_path,
-                    std::iter::repeat("_").take(parameter_count).collect::<Vec<_>>().join(", ")
-                )
-            };
+            let trait_name =
+                if trait_generics.params.len() <= trait_generics.has_self as usize {
+                    trait_path
+                } else {
+                    let counts = trait_generics.own_counts();
```
m-ou-se, line 246: This used `.count()` before, and `.own_counts()` now, to get counts for each kind of generics individually. So that does not include the parent counts. I couldn't find any case where that makes a difference though. Will this break (or fix) anything?

nikomatsakis: For a trait, no, traits never have parents.

PR #88499, opened 2021-08-30
Provide `layout_of` automatically (given tcx + param_env + error handling).

eddyb:
After #88337, there's no longer any uses of `LayoutOf` within `rustc_target` itself, so I realized I could move the trait to `rustc_middle::ty::layout` and redesign it a bit.

This is similar to #88338 (and supersedes it), but at no ergonomic loss, since there's no funky `C: LayoutOf<Ty = Ty>` -> `Ty: TyAbiInterface<C>` generic `impl` chain, and each `LayoutOf` still corresponds to one `impl` (of `LayoutOfHelpers`) for the specific context.

After this PR, this is what's needed to get `trait LayoutOf` (with the `layout_of` method) implemented on some context type:
* `TyCtxt`, via `HasTyCtxt`
* `ParamEnv`, via `HasParamEnv`
* a way to transform `LayoutError`s into the desired error type
  * an error type of `!` can be paired with having `cx.layout_of(...)` return `TyAndLayout` *without* `Result<...>` around it, such as used by codegen
  * this is done through a new `LayoutOfHelpers` trait (and so is specifying the type of `cx.layout_of(...)`)

When going through this path (and not bypassing it with a manual `impl` of `LayoutOf`), the end result is that only the error case can be customized, the query itself and the success paths are guaranteed to be uniform.

(**EDIT**: just noticed that because of the supertrait relationship, you cannot actually implement `LayoutOf` yourself, the blanket `impl` fully covers all possible context types that could ever implement it)

Part of the motivation for this shape of API is that I've been working on querifying `FnAbi::of_*`, and what I want/need to introduce for that looks a lot like the setup in this PR - in particular, it's harder to express the `FnAbi` methods in `rustc_target`, since they're much more tied to `rustc` concepts.

r? @nagisa cc @oli-obk @bjorn3


compiler/rustc_middle/src/ty/layout.rs
```diff
@@ -2062,29 +2062,113 @@ impl<'tcx, T: HasTyCtxt<'tcx>> HasTyCtxt<'tcx> for LayoutCx<'tcx, T> {
...
+impl<T, E> MaybeResult<T> for Result<T, E> {
+    type Error = E;
+
+    fn from(x: Result<T, Self::Error>) -> Self {
+        x
+    }
+    fn to_result(self) -> Result<T, Self::Error> {
+        self
+    }
+}
```
bjorn3, line 2092: How does this work without specialization and not result in conflicting impls?

compiler/rustc_middle/src/ty/layout.rs
```diff
@@ -2062,29 +2062,113 @@ impl<'tcx, T: HasTyCtxt<'tcx>> HasTyCtxt<'tcx> for LayoutCx<'tcx, T> {
...
+        self.spanned_layout_of(ty, DUMMY_SP)
+    }
+
+    /// Computes the layout of a type, at `span`. Note that this implicitly
+    /// executes in "reveal all" mode, and will normalize the input type.
+    // FIXME(eddyb) avoid passing information like this, and instead add more
+    // `TyCtxt::at`-like APIs to be able to do e.g. `cx.at(span).layout_of(ty)`.
+    #[inline]
+    fn spanned_layout_of(&self, ty: Ty<'tcx>, span: Span) -> Self::LayoutOfResult {
+        let span = if !span.is_dummy() { span } else { self.layout_tcx_at_span() };
```
bjorn3, line 2140: Why not directly use `layout_tcx_at_span()` in `layout_of`?

eddyb: Because then I have to either duplicate the rest of the body, or end up with silly codegen that calls `layout_tcx_at_span` twice.

Overall I don't like the current design (where `spanned_layout_of` exists), and the way miri handles this (where its `tcx` field is a `TyCtxtAt` that it keeps updating) feels superior.

compiler/rustc_middle/src/ty/layout.rs
eddyb: As I understand it, trying to unify `Result<$0, _>` with `$0` (to check whether one of the `impl`s applies in the context of the other one) will fail because its sole "solution" is an infinitely nested `Result<Result<Result<...` - in the literature, I believe this is called an ["occurs check"](https://en.wikipedia.org/wiki/Occurs_check).

A crucial component of this, btw, is that `MaybeResult` is generic over the `Ok`/`T` type, instead of having it associated like the `Err`/`Error` type - if it was an associated type, it would indeed overlap.

This also shows up when implementing `From`: `impl<T> From<T> for T` doesn't involve specialization to avoid conflicting with your own different `impl`s.

eddyb:
Happy with those perf results (I was worried something might accidentally get worse in the reshuffle, but that doesn't seem to be the case).

compiler/rustc_middle/src/ty/layout.rs
```diff
@@ -2062,29 +2062,113 @@ impl<'tcx, T: HasTyCtxt<'tcx>> HasTyCtxt<'tcx> for LayoutCx<'tcx, T> {
...
-    type TyAndLayout = Result<TyAndLayout<'tcx>, LayoutError<'tcx>>;
+/// Trait for contexts that want to be able to compute layouts of types.
+/// This automatically gives access to `LayoutOf`, through a blanket `impl`.
+pub trait LayoutOfHelpers<'tcx>: HasDataLayout + HasTyCtxt<'tcx> + HasParamEnv<'tcx> {
+    /// The `TyAndLayout`-wrapping type (or `TyAndLayout` itself), which will be
+    /// returned from `layout_of` (see also `handle_layout_err`).
+    type LayoutOfResult: MaybeResult<TyAndLayout<'tcx>>;
 
+    /// `Span` to use for `tcx.at(span)`, from `layout_of`.
+    // FIXME(eddyb) perhaps make this mandatory to get contexts to track it better?
```
nagisa, line 2104: I can see something like `CodegenCx` needing to get better at it, but I don't see how having `TyCtxt` track any sort of span would work? If we made this required, we would have to drop the `LayoutOfHelpers` impls for `TyCtxt` and only keep the `TyCtxtAt`.

Seems reasonable enough, but also sounds like a big undertaking.

compiler/rustc_middle/src/ty/layout.rs
```diff
@@ -2062,29 +2062,113 @@ impl<'tcx, T: HasTyCtxt<'tcx>> HasTyCtxt<'tcx> for LayoutCx<'tcx, T> {
...
+        err
+    }
+}
+
+impl LayoutOfHelpers<'tcx> for LayoutCx<'tcx, ty::query::TyCtxtAt<'tcx>> {
+    type LayoutOfResult = Result<TyAndLayout<'tcx>, LayoutError<'tcx>>;
+
+    #[inline]
+    fn layout_tcx_at_span(&self) -> Span {
+        self.tcx.span
```
nagisa, line 2166: Ha, I spent a couple minutes wondering where is there a `span` in `TyCtxt`. Took me a while to notice this is an impl on a `LayoutCx`, not directly `TyCtxtAt`.

nagisa (approved):

Issue #63818, opened 2019-08-22
async/await unsoundness


Mark-Simulacrum:
async/await is possibly unsound -- I don't consider myself an authority here, so feel free to edit this top post with relevant information.

Current discussion happened in https://github.com/rust-lang/rust/pull/63209

Mark-Simulacrum:
cc @rust-lang/lang @comex @rkruppe @RalfJung

comex:
Responding to @RalfJung's post from #63209:

> I think this is fundamental. Adding `derive(Debug)` makes the exact content of local variables at any yield point observable. Interesting optimizations based on mutable references being unique rely on the content mutable references point to to _not_ be observable by any other alias.

Those hypothetical optimizations could be disabled in async fns across suspend points, though, couldn't they?  I can't think of any case where this would force the compiler to pessimize *all* code because of the mere possibility of async being involved.  Therefore, the question is whether those optimizations being performed on async fns specifically is more valuable than `derive(Debug)` is.

(By the way, I still want async fns to be able to implement `derive(Clone)`, although that would obviously create pretty severe restrictions on the body of the function.)

nikomatsakis:
I've not weighed in on this but I wanted to leave a few thoughts. Let me know if I'm confused about something.

First off, the conflict here is between the reference to a generator type G and some `&mut` reference T contained within the generator. What we *don't* have is two `&mut A` and `&mut B` references that overlap where neither A nor B gives any indication of arising from a generator. 

This seems important: it means that, in principle, we ought to be able to understand that the `&mut G` reference is somehow special. It is certainly true that the self-referential nature of generators will require extending "stacked borrows" to be more flexible. But I think that generators are something we want to have, and it doesn't seem surprising that our aliasing rules may need to be extended to accommodate and understand them. (If though we have conflict between two random types `&mut A` and `&mut B`, that seems more worrisome, though it is likely still true that we want generators regardless.)

Also, of course, this problem isn't really specific to **generators** but is rather a pattern that emerges with any self-referential setup. Presently, safe Rust has no way to express self-referential structs (apart from generators), but I think it's something we would eventually like to support.

So let's posit that we find a version of stacked borrows that understands self-referential structs. Clearly, this doesn't mean we can add `noalias` to `&mut` references everywhere, we'll either have to extend LLVM or be selective in some other way (as e.g. @comex suggests [here](https://github.com/rust-lang/rust/issues/63818#issuecomment-524101921)).

Does this all sound correct?

[...]

PR #88500, opened 2021-08-30
Rewrite `Pin<P>` docs to clarify guarantees and uses

mcy:
(With apologies to @RalfJung.)

The documentation today does not give a complete treatment of pinning
from first principles, which appropriately describes how to design types
that use it, nor does it provide formal statements of the guarantees
users need to be aware of.

This rewrite attempts to address these in a way that makes the concept
more approachable while also making the documentation more normative.

This is primarily motivated by my experience designing the `moveit` crate.
I'd like to thank @Manishearth and @anp for reviewing early drafts of this
new documentation.

### DIFF
--- a/library/core/src/pin.rs
+++ b/library/core/src/pin.rs
@@ -1,107 +1,219 @@
-//! Types that pin data to its location in memory.
+//! Types that pin data to a location in memory.
 //!
-//! It is sometimes useful to have objects that are guaranteed not to move,
-//! in the sense that their placement in memory does not change, and can thus be relied upon.
-//! A prime example of such a scenario would be building self-referential structs,
-//! as moving an object with pointers to itself will invalidate them, which could cause undefined
-//! behavior.
+//! It is sometimes useful to have objects which can be relied upon not to move,
+//! in the sense that their address in memory does not change. This is necessary to implement
+//! things like self-referential structs, as moving an object with pointers to itself will
+//! invalidate them, likely causing undefined behavior.
 //!
-//! At a high level, a <code>[Pin]\<P></code> ensures that the pointee of any pointer type
-//! `P` has a stable location in memory, meaning it cannot be moved elsewhere
-//! and its memory cannot be deallocated until it gets dropped. We say that the
-//! pointee is "pinned". Things get more subtle when discussing types that
-//! combine pinned with non-pinned data; [see below](#projections-and-structural-pinning)
-//! for more details.
+//! "Pinning" allows us to prevent safe code from *moving* a value, so that [`unsafe`] code
+//! can rely on it remaining stationary. This documentation is intended to be the source of truth
+//! for users of [`Pin<P>`] in unsafe code; users of [`Pin<P>`] in safe code do not need to read it
+//! in detail.
 //!
-//! By default, all types in Rust are movable. Rust allows passing all types by-value,
-//! and common smart-pointer types such as <code>[Box]\<T></code> and <code>[&mut] T</code> allow
-//! replacing and moving the values they contain: you can move out of a <code>[Box]\<T></code>,
-//! or you can use [`mem::swap`]. <code>[Pin]\<P></code> wraps a pointer type `P`, so
-//! <code>[Pin]<[Box]\<T>></code> functions much like a regular <code>[Box]\<T></code>:
-//! when a <code>[Pin]<[Box]\<T>></code> gets dropped, so do its contents, and the memory gets
-//! deallocated. Similarly, <code>[Pin]<[&mut] T></code> is a lot like <code>[&mut] T</code>.
-//! However, <code>[Pin]\<P></code> does not let clients actually obtain a <code>[Box]\<T></code>
-//! or <code>[&mut] T</code> to pinned data, which implies that you cannot use operations such
-//! as [`mem::swap`]:
+//! There are several sections to this documentation:
+//!
+//! * [What is "*moving*"?][what-is-moving]
+//! * [What is "pinning"?][what-is-pinning]
+//! * [Examples of address-sensitive types][address-sensitive-examples]
+//!   * [Self-referential struct][self-ref]
+//!   * [Intrusive, doubly-linked list][linked-list]
+//! * [Subtle Details][subtle-details]
+//!
+//! # What is "*moving*"?
+//! [what-is-moving]: #what-is-moving
+//!
+//! All values in Rust are trivially *moveable*. This means that the address at which
+//! a value is located is not stable, even if no explicit semantic moves occur and its `Drop`
+//! implementation is never called. For example:
 //!
 //! ```
-//! use std::pin::Pin;
-//! fn swap_pins<T>(x: Pin<&mut T>, y: Pin<&mut T>) {
-//!     // `mem::swap` needs `&mut T`, but we cannot get it.
-//!     // We are stuck, we cannot swap the contents of these references.
-//!     // We could use `Pin::get_unchecked_mut`, but that is unsafe for a reason:
-//!     // we are not allowed to use it for moving things out of the `Pin`.
+//! #[derive(Default)]
+//! struct AddrTracker(usize);
+//!
+//! impl AddrTracker {
+//!     fn check_for_move(&mut self) {
+//!         let addr = self as *mut Self as usize;
+//!         match self.0 {
+//!             0 => self.0 = addr,
+//!             x => assert_eq!(x, addr),
+//!         }
+//!     }
 //! }
+//!
+//! let mut tracker = AddrTracker::default();
+//! tracker.check_for_move();
+//! // May panic!
+//! // tracker.check_for_move();
 //! ```
 //!
-//! It is worth reiterating that <code>[Pin]\<P></code> does *not* change the fact that a Rust
-//! compiler considers all types movable. [`mem::swap`] remains callable for any `T`. Instead,
-//! <code>[Pin]\<P></code> prevents certain *values* (pointed to by pointers wrapped in
-//! <code>[Pin]\<P></code>) from being moved by making it impossible to call methods that require
-//! <code>[&mut] T</code> on them (like [`mem::swap`]).
+//! Rust does not guarantee that `check_for_move()` will never panic, because the
+//! compiler is permitted to *move* `tracker` to enable optimizations like pass-by-value.
 //!
-//! <code>[Pin]\<P></code> can be used to wrap any pointer type `P`, and as such it interacts with
-//! [`Deref`] and [`DerefMut`]. A <code>[Pin]\<P></code> where <code>P: [Deref]</code> should be
-//! considered as a "`P`-style pointer" to a pinned <code>P::[Target]</code> – so, a
-//! <code>[Pin]<[Box]\<T>></code> is an owned pointer to a pinned `T`, and a
-//! <code>[Pin]<[Rc]\<T>></code> is a reference-counted pointer to a pinned `T`.
-//! For correctness, <code>[Pin]\<P></code> relies on the implementations of [`Deref`] and
-//! [`DerefMut`] not to move out of their `self` parameter, and only ever to
-//! return a pointer to pinned data when they are called on a pinned pointer.
+//! > When we say a value is *moved*, we mean that the compiler copies, byte-for-byte, the
+//! > value from one location to another, without running any code to notify it that its
+//! > address changed. A move is mechanically identical to a [`Copy`] where the
+//! > moved-from value becomes inaccessible. Whenever we write *move* in italics, we mean
+//! > this precise definition of moving a value.
 //!
-//! # `Unpin`
+//! Common smart-pointer types such as [`Box<T>`] and [`&mut T`] allow removing and replacing the
+//! values they contain: you can move out of a [`Box<T>`], or you can use [`mem::swap`]. Putting
+//! `tracker` behind a pointer isn't enough on its own to ensure that its address does not change.
 //!
-//! Many types are always freely movable, even when pinned, because they do not
-//! rely on having a stable address. This includes all the basic types (like
-//! [`bool`], [`i32`], and references) as well as types consisting solely of these
-//! types. Types that do not care about pinning implement the [`Unpin`]
-//! auto-trait, which cancels the effect of <code>[Pin]\<P></code>. For <code>T: [Unpin]</code>,
-//! <code>[Pin]<[Box]\<T>></code> and <code>[Box]\<T></code> function identically, as do
-//! <code>[Pin]<[&mut] T></code> and <code>[&mut] T</code>.
+//! # What is "pinning"?
+//! [what-is-pinning]: #what-is-pinning
 //!
-//! Note that pinning and [`Unpin`] only affect the pointed-to type <code>P::[Target]</code>,
-//! not the pointer type `P` itself that got wrapped in <code>[Pin]\<P></code>. For example,
-//! whether or not <code>[Box]\<T></code> is [`Unpin`] has no effect on the behavior of
-//! <code>[Pin]<[Box]\<T>></code> (here, `T` is the pointed-to type).
+//! `Pin` wraps a pointer type `P`, but removes the ability to actually obtain a `P` from it in safe
+//! code. By disallowing access to the type behind the pointer, `Pin` prevents us from using
+//! operations like [`mem::swap`] to *move* out of the pointer.
 //!
-//! # Example: self-referential struct
+//! ## Address-sensitive types
 //!
-//! Before we go into more details to explain the guarantees and choices
-//! associated with <code>[Pin]\<P></code>, we discuss some examples for how it might be used.
-//! Feel free to [skip to where the theoretical discussion continues](#drop-guarantee).
+//! One noteworthy example of types which require `Pin` for correctness are the [`Future`]` types
+//! generated by the compiler for `async fn`s.
+//!
+//! For these types, any value which lives "across" an `.await` point is stored in the state
+//! machine, including values which borrow from other variables in the `async fn`. *Moving* the
+//! [`Future`] in between calls to [`poll`] would invalidate these pointers, leaving the next call
+//! to [`poll`] with dangling references!
+//!
+//! Such types are *address-sensitive*: they incorporate the address of `self` into an
+//! operation. These types usually follow a common lifecycle:
+//!
+//! 1. A value is created which can be freely moved around.
+//!     * e.g. calling an async function
+//! 2. An operation causes the value to depend on its own address using [`unsafe`].
+//!     * e.g. calling [`poll`] for the first time on the produced [`Future`]
+//! 3. Further [`unsafe`] operations assume that its address is stable.
+//!     * e.g. subsequent calls to [`poll`]
+//! 4. The value is destroyed, undoing its address-sensitivity.
+//!     * e.g. [`drop`]ping the [`Future`]
+//!
+//! [`Pin<P>`] helps us implement steps (2) and (3) safely.
+//!
+//! ## `Pin` and pointers
+//!
+//! [`Pin<P>`] can wrap any pointer type, forming a promise that the pointee will not be *moved*.
+//! This promise must be upheld by [`unsafe`] code which interacts with the [`Pin<P>`] so that
+//! [`unsafe`] code can place the pointee in an address-sensitive state that will not be broken
+//! by a *move*. Operations on an address-sensitive type accept an argument like
+//! <code>[Pin]<[`&mut T`]></code> or <code>[Pin]<[`Box<T>`]></code> to indicate this contract to
+//! the caller.
+//!
+//! Since [`Pin<P>`] can wrap any pointer type, it interacts with
+//! [`Deref`] and [`DerefMut`]. A [`Pin<P>`] where [`P: Deref`][Deref] is a
+//! "`P`-style pointer" to a pinned [`P::Target`][Target] – so, a
+//! <code>[Pin]<[`Box<T>`]></code> is an owned pointer to a pinned `T`, and a
+//! <code>[Pin]<[`Rc<T>`]></code> is a reference-counted pointer to a pinned `T`.
+//!
+//! [`Pin<P>`] requires that implementations of [`Deref`] and [`DerefMut`] return a pointer to
+//! pinned data when they are called on a pinned pointer and do not *move* out of their `self`
+//! parameter. It is unsound for [`unsafe`] code to wrap such "evil" pointers; see
+//! [`Pin<P>::new_unchecked`] for details.
+//!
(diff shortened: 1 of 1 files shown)


library/core/src/pin.rs
```diff
@@ -1,107 +1,218 @@
...
+//!
+//! [`Pin<P>`] requires that implementations of [`Deref`] and [`DerefMut`] return a pointer to
+//! pinned data when they are called on a pinned pointer and do not *move* out of their `self`
+//! parameter. It is unsound for [`unsafe`] code to wrap such "evil" pointers; see
+//! [`Pin<P>::new_unchecked`] for details.
+//!
+//! Pinning does not require any compiler "magic", only a specific contract between the library API
+//! and its users. This differs from e.g. [`UnsafeCell`] which changes the semantics of a program's
+//! compiled output. A [`Pin<P>`] is a handle to a value which does not allow moving the value out,
+//! but Rust still considers all values themselves to be moveable with e.g. [`mem::swap`].
```
eddyb, line 116: To the best of my knowledge, this is not 100% accurate: `Pin<&mut T>` being a newtype of `&mut T` is unsound because `&mut T` implies more (wrt aliasing) than we want `Pin<&mut T>` to.

Either "compiler magic" or a new bound (which we can't add now AFAIK) would be necessary.

For more history of this issue, see #63818.

mcy: Yes, I intentionally glossed over the fussiness around `noalias`, since I get the impression that we don't actually know what we want to do here. It's not clear to me that we want `Pin<&mut T>` to alias less or if we're going to come up with some newtype such as `&mut A<T>` to capture that aliasing contract.

mcy:
Hmm, it looks like the link checker CI pass is very confused by some of the things I did in the documentation. I'd appreciate some help here; it's entirely possible the tool rejects some otherwise-correct links.

Manishearth:
Oh i wonder if it's because those are module level docs and they may be resolved relative to the _parent_ module? I was under the impression we handled that but I could be wrong

Manishearth:
cc @GuillaumeGomez @jyn514 about the linkchecker

library/core/src/pin.rs
```diff
@@ -195,189 +448,269 @@
...
+//! [`&mut T`]: &mut
+//! [`&mut self`]: &mut
+//! [`&mut Self`]: &mut
+//! [`&mut Field`]: &mut
+//!
+//! [`Unmovable`]: self-ref
+//! [`Node`]: linked-list
+//! [`List`]: linked-list
+//! [`&mut Node`]: linked-list
+//! [`List::append`]: linked-list
```
jyn514, line 708: Where are these supposed to go? These aren't valid Rust identifiers, and they're also not HTML pages. Most of the link failures are because of these.

eddyb:
I'm not exactly qualified to review this, but there's one thing I noticed, and I'm wondering how much it matters.

A lot of the examples around address sensitivity talk about what I would call "internal address sensitivity", i.e. something akin to a self-borrow, where the data type contains its own address (whether as a pointer or not - even a hash would count) - we call most of these "self-referential data types", I suppose.

*However*, it's not the only kind of "phantom borrow" (i.e. not visible to borrowck) that can take advatange of `Pin`, and the other category I would call "external address sensitivity". This usually involves some kind of global state, e.g.:
* the OS holding the address, and either accessing the data, or associating an OS resource with it
  * the Linux "Fast Userspace muTEX" (`futex`) API works a bit like this, but AFAIK only while locked, so `Pin` shouldn't be needed, the lock guard borrow should be enough
* process/thread-wide data structures, temporarily pointing to shorter-lived data
  * this could be "roots" for some garbage collector, informing it of pointers into its GC heap, that it cannot itself see (e.g. on the native stack)
  * arguably something like `glibc`'s `FILE*` could be modeled using `Pin<Box<FILE>>` - crucially, there is a global linked list through all live `FILE`s

One thing a lot of these usecases would have in common is relying on the `Drop` guarantee (assuming it's even strong enough) offered by `Pin`, to "deregister themselves" from the global state (wherever it may exist).

Like I said at the start, I'm not sure how much it matters, and even if there's agreement this kind of usecase should be documented/exemplified, I doubt it would need to be handled by this PR, but I wanted to note it nevertheless.

library/core/src/pin.rs
```diff
@@ -195,189 +448,269 @@
...
+//!
+//! [&]: https://doc.rust-lang.org//std/primitive.reference.html "shared reference"
+//! [&mut]: https://doc.rust-lang.org//std/primitive.reference.html "mutable reference"
```
jyn514, line 711: Please use `reference` instead, which accounts for the nightly channel properly. (`&` and `&mut` would also work, but shortcut links don't allow a custom title AFAIK.)
```suggestion
//! [&]: reference "shared reference"
//! [&mut]: reference "mutable reference"
```

library/core/src/pin.rs
```diff
@@ -195,189 +448,269 @@
...
+//! [`unsafe`]: https://doc.rust-lang.org//std/keyword.unsafe.html "keyword unsafe"
```
jyn514, line 712: Ideally this would include the channel, but I don't think anything else in the standard library does today and it seems silly to block on it. I might make a PR this weekend.

library/core/src/pin.rs
```diff
@@ -195,189 +448,269 @@
...
+//! [`Vec<T>`]: https://doc.rust-lang.org/std/vec/struct.Vec.html
+//! [Vec::pop]: https://doc.rust-lang.org/std/vec/struct.Vec.html#method.pop
+//! [Vec::push]: https://doc.rust-lang.org/std/vec/struct.Vec.html#method.push
+//! [`Vec::set_len`]: https://doc.rust-lang.org/std/vec/struct.Vec.html#method.set_len
+//! [`VecDeque<T>`]: https://doc.rust-lang.org/std/collections/struct.VecDeque.html
+//!
+//! [`&mut T`]: &mut
+//! [`&mut self`]: &mut
+//! [`&mut Self`]: &mut
+//! [`&mut Field`]: &mut
```
jyn514, line 702: It looks like you intend for this to be a shortcut for this line below:

```
 //! [&mut]: https://doc.rust-lang.org//std/primitive.reference.html "mutable reference"
```

That's not what it does, instead it's using Rustdoc's link resolution, which already allows `&mut` to be valid. If you want to use the line you have below, use ``[`&mut T`][&mut]` in the paragraph itself instead.

library/core/src/pin.rs
```diff
@@ -195,189 +448,269 @@
...
-//! [`unsafe`]: ../../std/keyword.unsafe.html "keyword unsafe"
+//! mutable reference even when you just have <code>[Pin]<[`&mut Self`]></code>
+//! (such as in your own [`poll`] implementation).
+//!
+//! [Target]: Deref::Target "ops::Deref::Target"
+//! [`drop`]: Drop::drop "ops::Drop::drop"
+//! [`poll`]: Future::poll "future::Future::poll"
+//!
+//! <!-- These require `alloc`, but we're in `core`. -->
+//! [`std::alloc`]: https://doc.rust-lang.org/std/alloc/index.html
```
jyn514, line 690: Please use relative links instead so it accounts for the channel.
```suggestion
//! [`std::alloc`]: ../../std/alloc/index.html
```
and the same for the rest of the links below.

jyn514 (changes requested):
I noticed a bunch of issues with the markdown syntax - given that I doubt there's a rustdoc bug here, but happy to take another look after the issues are fixed.

library/core/src/pin.rs
```diff
@@ -195,189 +448,269 @@
...
 //! is not [`Unpin`]. What that type thinks about pinning is not relevant
-//! when no <code>[Pin]<[&mut] Field></code> is ever created.
+//! when no <code>[Pin]<[`&mut Field`]></code> is ever created.
+//!
+//! For example, the `data` field of [`Node`] does *not* need
+//! to be structurally pinned, because neither [`List`] nor
+//! [`Node`] assume anything about it.
 //!
-//! ## Pinning *is* structural for `field`
+//! ### When inning *is* structural for `field`...
```
ssbr, line 570: When pinning

library/core/src/pin.rs
```diff
@@ -110,79 +221,221 @@
...
-//! In an intrusive doubly-linked list, the collection does not actually allocate
-//! the memory for the elements itself. Allocation is controlled by the clients,
-//! and elements can live on a stack frame that lives shorter than the collection does.
+//! 1.  *Address Stability.* If [`unsafe`] code witnesses any [`p: Pin<P>`][Pin] at any time then
+//!     it may assume that `p.as_ref().get_ref() as *const _` will remain valid, pointing to the
+//!     same object until further notice.
+//! 2.  *Notice of Destruction.* If `x: T` was ever reachable through any [`Pin<P>`] type, its
+//!     destructor must be run (until it either returns or panics) before `x`'s storage can be
+//!     overwritten. The "until further notice" in (1) includes this mandatory destruction. This is
+//!     often called the "[`Drop`] guarantee".
```
Darksonn, line 360: Might be worth mentioning that you don't have to drop it if you leak the memory.

library/core/src/pin.rs
```diff
@@ -110,79 +221,221 @@
...
 //!
-//! Therefore, pinning also comes with a [`drop`]-related guarantee.
+//! When a [`List`] stores a [`Node`], it needs to assume that appending a second node will mutate the
+//! first node, so that later, when the first node is removed, it knows that its predecessor is the
+//! second node.
 //!
-//! # `Drop` guarantee
+//! When writing generic code, it's not possible to know what [`unsafe`] code has recorded about the
+//! pointee's address, so it must be very careful to observe this invariant. Thankfully, most of this
+//! is already enforced by the [`Pin<P>`] API, so only [`unsafe`] code needs to worry about this.
```
Darksonn, line 377: Maybe a better way to phrase this is that what pinning _actually_ does is simply prevent safe code from performing any mutations whatsoever to the memory location without running that memory location's destructor first. The mutations that are prevented are typically thought of as types of moves, which is why we use "prevents moves" as the terminology for this, but it is not inherently incorrect for unsafe code to allow you to `swap` two pinned values, assuming that the unsafe block doesn't violate the invariants of some other unsafe blocks.

library/core/src/pin.rs
```diff
@@ -195,189 +448,269 @@
...
-//!
-//! As the author of a data structure you get to decide for each field whether pinning
+//! pinning with a [`#[repr(packed)]`][packed] type.
+//!
+//! ## "Assigning" pinned data
+//!
+//! Although in general it is not valid to swap or assign data through a [`Pin<P>`], for
+//! the same reason that a *move* is invalid, there is no particular reason to disallow
+//! doing it with specialized functions, as long as they know how to update all uses of
+//! the pinned address. For [`Unmovable`], we could write
```
Darksonn, line 464: Note that [`Pin::set`](https://doc.rust-lang.org/std/pin/struct.Pin.html#method.set) exists.

Darksonn (commented):
I didn't read all the way through, but I've got some comments.

library/core/src/pin.rs
```diff
@@ -1,107 +1,218 @@
...
+//! This promise must be upheld by [`unsafe`] code which interacts with the [`Pin<P>`] so that
+//! [`unsafe`] code can place the pointee in an address-sensitive state that will not be broken
+//! by a *move*. Operations on an address-sensitive type accept an argument like
+//! <code>[Pin]<[`&mut T`]></code> or <code>[Pin]<[`Box<T>`]></code> to indicate this contract to
+//! the caller.
+//!
+//! Since [`Pin<P>`] can wrap any pointer type, it interacts with
+//! [`Deref`] and [`DerefMut`]. A [`Pin<P>`] where [`P: Deref`][Deref] is a
+//! "`P`-style pointer" to a pinned [`P::Target`][Target] – so, a
+//! <code>[Pin]<[`Box<T>`]></code> is an owned pointer to a pinned `T`, and a
```
letitz, line 105: `T` is not defined yet here. This should be `P::Target`?

library/core/src/pin.rs
mcy: This is noted in the larger section on the drop guarantee. I think it's a bit of a footnote more than anything else.

library/core/src/pin.rs
mcy: I'm not sure I agree. The intent of this section is to be extremely pedantic about what guarantees are made to unsafe code, rather than the emergent properties of the behavior you describe. This topic is complex enough as it is, and I'd rather not have people have to deduce the "I can record the address and no one will mess with it" lemma from that property.

library/core/src/pin.rs
mcy: No. It is referring to the `T` in the `Box<T>`.

library/core/src/pin.rs
```diff
@@ -1,107 +1,218 @@
...
+//!             0 => self.0 = addr,
+//!             x => assert_eq!(x, addr),
+//!         }
+//!     }
 //! }
+//!
+//! let mut tracker = AddrTracker::default();
+//! tracker.check_for_move();
+//! // May panic!
+//! // tracker.check_for_move();
```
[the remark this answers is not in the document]
mcy: The compiler makes no such guarantee. It is permitted to perform arbitrary moves of stack variables at any time as convenient.

library/core/src/pin.rs
mcy: Well, they're supposed to refer to this:
```rust
//! ## Intrusive, doubly-linked list
//! [linked-list]: #an-intrusive-doubly-linked-list
```
Perhaps it doesn't work the way I expect and I should just inline the #anchor?

jyn514: Either inline the anchor, or use `[`List`][linked-list]` directly in the paragraph. Reference links can't use the names of other reference links as their target if they're out-of-line.

library/core/src/pin.rs
```diff
@@ -387,6 +725,13 @@ use crate::hash::{Hash, Hasher};
 use crate::marker::{Sized, Unpin};
 use crate::ops::{CoerceUnsized, Deref, DerefMut, DispatchFromDyn, Receiver};
 
+use crate::{
```
jyn514, line 728: I assume these are for intra-doc links. You'll need to `allow(unused_imports)`, rustc doesn't know that rustdoc uses these.
```suggestion
#[allow(unused_imports)]
use crate::{
```

library/core/src/pin.rs
mcy: That is unfortunate. =/

Fixed.

library/core/src/pin.rs
mcy: Ah! I had used `#[cfg(doc)]` but that made one of the CI passes really confused.

library/core/src/pin.rs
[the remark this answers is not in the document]
Darksonn: It seems that it would follow that the unsafe block in [this snippet][1] is incorrect and that it could panic.

[1]: https://play.rust-lang.org/?version=stable&mode=release&edition=2018&gist=e2dc0cf4ae67009f36ddc228c9b660fc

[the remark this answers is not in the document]
Darksonn: Or perhaps this is a better example: https://play.rust-lang.org/?version=stable&mode=release&edition=2018&gist=a4a6b4dfac8c5d0a281e7af473fff501

library/core/src/pin.rs
```diff
@@ -195,189 +451,266 @@
...
-//! [&]: ../../std/primitive.reference.html "shared reference"
-//! [&mut]: ../../std/primitive.reference.html "mutable reference"
+//! mutable reference even when you just have <code>[Pin]<[`&mut Self`]></code>
+//! (such as in your own [`poll`] implementation).
+//!
+//! [Target]: Deref::Target "ops::Deref::Target"
+//! [`drop`]: Drop::drop "ops::Drop::drop"
+//! [`poll`]: Future::poll "future::Future::poll"
+//!
+//! <!-- These require `alloc`, but we're in `core`. -->
```
jyn514, line 697: You can simplify this:
```suggestion
// These require `alloc`, but we're in `core`.
```

library/core/src/pin.rs
jyn514: Yeah, cfg(doc) breaks the links when they're exported, it's not set for dependencies.

jyn514 (approved):
The links look good to me, I can take another look if it fails again. I don't know enough about Pin to have opinions on the actual changes.

library/core/src/pin.rs
```diff
@@ -1,107 +1,219 @@
...
+//!
+//! * [What is "*moving*"?][what-is-moving]
+//! * [What is "pinning"?][what-is-pinning]
+//! * [Examples of address-sensitive types][address-sensitive-examples]
+//!   * [Self-referential struct][self-ref]
+//!   * [Intrusive, doubly-linked list][linked-list]
+//! * [Subtle Details][subtle-details]
+//!
+//! # What is "*moving*"?
+//! [what-is-moving]: #what-is-moving
```
jyn514, line 23: Try adding `self` in front so it doesn't break when re-exported.
```suggestion
//! [what-is-moving]: self#what-is-moving
```
(and the same for all the other fragment links)

library/core/src/pin.rs
letitz: Indeed! My mistake.

m-ou-se (changes requested):
Thanks for working on this!

danielhenrymantilla (commented):
Very nice initiative, `Pin` is such a subtle API that improving its docs is always welcome! I've written my own set of nits, which can be grouped within the following categories:

  - `Pin<P>` ought to be renamed to `Pin<Ptr>`: if we are striving for clarity, that name change will make it crystal clear that it is the pointer and not the pointee which is `Pin`-wrapped, all throughout each and every explanation.

  - An intrusive doubly-linked list that is not lifetime-infected is *incredibly* tricky hard to get right: while `Pin` gives us guarantees about pointers being safe to dereference right when witnessed non-null, there can easily be concurrency / aliasing issues (including parallelism / thread-safety ones which are luckily dodged by the usage of raw pointers, but which ought to be mentioned) and APIs such as `fn next(… '_ Node) -> … '_ Node` cannot be featured.

  - adding some details / clarifications to some sentences or paragraphs, feel free to _further_ tweak my own suggestions.

  - I believe that before talking of any guarantee that `Pin` makes, a short disclaimer ought to be added saying that none of the guarantees hold when the pointee is `Unpin` (linking to the full paragraph about it). Indeed, we should consider the case of a truncated read of the docs, especially when snippets are involved: there should be a clear "opt out of `Unpin`" part of the snippet whenever it makes assumptions about `Pin`-induced properties.

JohnCSimon:
Ping from triage: 
@mcy can you please post your status on this PR?

library/core/src/pin.rs
```diff
@@ -1,107 +1,219 @@
...
-//! whether or not <code>[Box]\<T></code> is [`Unpin`] has no effect on the behavior of
-//! <code>[Pin]<[Box]\<T>></code> (here, `T` is the pointed-to type).
-//!
-//! # Example: self-referential struct
-//!
-//! Before we go into more details to explain the guarantees and choices
-//! associated with <code>[Pin]\<P></code>, we discuss some examples for how it might be used.
-//! Feel free to [skip to where the theoretical discussion continues](#drop-guarantee).
+//! Rust does not guarantee that `check_for_move()` will never panic, because the
+//! compiler is permitted to *move* `tracker` to enable optimizations like pass-by-value.
```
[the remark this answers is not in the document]
RalfJung: I don't think we allow the compiler to insert moves.

The fact that all types are movable is part of the contract between a library and its environment, it is not part of what the compiler deals with.

library/core/src/pin.rs
```diff
@@ -1,107 +1,219 @@
...
+//!
+//! One noteworthy example of types which require `Pin` for correctness are the [`Future`]` types
+//! generated by the compiler for `async fn`s.
+//!
+//! For these types, any value which lives "across" an `.await` point is stored in the state
+//! machine, including values which borrow from other variables in the `async fn`. *Moving* the
+//! [`Future`] in between calls to [`poll`] would invalidate these pointers, leaving the next call
+//! to [`poll`] with dangling references!
+//!
+//! Such types are *address-sensitive*: they incorporate the address of `self` into an
```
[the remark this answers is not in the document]
RalfJung: I am not sure if it is good to speak of address-sensitive *types* here. It remains true that all types are movable, i.e., fundamentally not address-sensitive. This is true also for futures -- a freshly created future (before `poll` is called the first time) is movable. But certain *values* can be put into an "address-sensitive state", and *that* is what `Pin` tracks.

library/core/src/pin.rs
```diff
@@ -1,107 +1,219 @@
...
+//! tracker.as_mut().check_for_move();
+//! tracker.as_mut().check_for_move();
+//! ```
+//!
+//! [As discussed below][drop-guarantee], this has consequences for running
+//! destructors of pinned memory, too.
+//!
+//! ## [`Unpin`]
+//!
+//! The vast majority of Rust types are not address-sensitive; these types
```
RalfJung, line 151: Again, it's not types that are address-sensitive, it's values. So I think a more accurate way to put this would be to say that the majority of types do not have an address-sensitive state.

library/core/src/pin.rs
```diff
@@ -110,79 +222,223 @@
...
+//! 2.  *Notice of Destruction.* If `x: T` was ever reachable through any [`Pin<P>`] type, its
+//!     destructor must be run (until it either returns or panics) before `x`'s storage can be
+//!     overwritten. The "until further notice" in (1) includes this mandatory destruction. This is
+//!     often called the "[`Drop`] guarantee".
+//!
+//! ## Address Stability
+//! [address-stability]: #address-stability
+//!
+//! The precise meaning of "address stability" is subtle, because "the same object" is not well-defined.
+//! It is easiest to reason about it in terms of *visibility of mutations*. If [`unsafe`] code mutates
```
RalfJung, line 369: FWIW, I think this is best described in terms of which invariants / properties must be satisfied by the data. That's how a mathematical formal treatment of pinning would work (or at least I don't see how else it would work, and I know this way works). The key difference between the invariant of a 'regular' Rust value and the invariant of a value in its address-sensitive state is that the latter invariant *can depend on the address*. The regular invariant satisfies the property that it keeps holding when the same sequence of bytes is put at a different address in memory; the address-sensitive invariant does not have that property.

I don't know what 'visibility' is supposed to mean.

JohnCSimon:
Ping from triage: I'm closing this one due to inactivity over the last few months, please feel to reopen when you are ready to continue with this.

@rustbot label: +S-inactive

JohnCSimon:
This is for @mcy

PR #88501, opened 2021-08-30
Use right span in prelude collision suggestions with macros. 

### LINKED ISSUE #88347: 2021 prelude collision lint expands macros when used as an expression
The `rust_2021_prelude_collisions` lint suggestion seems to expand the expression for a macro, but I don't think it should.  The following:

```rust
#![warn(rust_2021_prelude_collisions)]

fn bar() {}

macro_rules! foo {
    () => {{
        $crate::bar();
        S
    }};
}

trait MyTry<T> {
    fn try_into(self) -> Result<T, ()>;
}

struct S;

impl MyTry<i32> for S {
    fn try_into(self) -> Result<i32, ()> {
        unimplemented!();
    }
}

fn main() {
    foo!().try_into().unwrap();
}
```

The suggestion recommends replacing it with:

```rust
    MyTry::try_into({
        $crate::bar();
        S
    }).unwrap();
```

which will not compile.  I would expect it to suggest:

```rust
MyTry::try_into(foo!()).unwrap();
```

Found during the crater run for:
* https://github.com/cedric-h/sorry - https://crater-reports.s3.amazonaws.com/pr-87190-3/try%23a7a572ce3edd6d476191fbfe92c9c1986e009b34/gh/cedric-h.sorry/log.txt

### Meta

`rustc --version --verbose`:
```
rustc 1.56.0-nightly (b03ccace5 2021-08-24)
binary: rustc
commit-hash: b03ccace573bb91e27625c190a0f7807045a1012
commit-date: 2021-08-24
host: x86_64-apple-darwin
release: 1.56.0-nightly
LLVM version: 13.0.0
```

m-ou-se:
Fixes https://github.com/rust-lang/rust/issues/88347

r? @estebank

### DIFF
--- a/compiler/rustc_typeck/src/check/method/prelude2021.rs
+++ b/compiler/rustc_typeck/src/check/method/prelude2021.rs
@@ -273,11 +273,10 @@ impl<'a, 'tcx> FnCtxt<'a, 'tcx> {
                 method_name.name
             ));
 
-            let mut self_ty_name = self
-                .sess()
-                .source_map()
-                .span_to_snippet(self_ty_span.find_ancestor_inside(span).unwrap_or_default())
-                .unwrap_or_else(|_| self_ty.to_string());
+            let mut self_ty_name = self_ty_span
+                .find_ancestor_inside(span)
+                .and_then(|span| self.sess().source_map().span_to_snippet(span).ok())
+                .unwrap_or_else(|| self_ty.to_string());
 
             // Get the number of generics the self type has (if an Adt) unless we can determine that
             // the user has written the self type with generics already which we (naively) do by looking
@@ -385,10 +384,10 @@ impl<'a, 'tcx> FnCtxt<'a, 'tcx> {
             Some(probe::AutorefOrPtrAdjustment::ToConstPtr) | None => "",
         };
 
-        let (expr_text, precise) = if let Ok(expr_text) = self
-            .sess()
-            .source_map()
-            .span_to_snippet(expr.span.find_ancestor_inside(outer).unwrap_or_default())
+        let (expr_text, precise) = if let Some(expr_text) = expr
+            .span
+            .find_ancestor_inside(outer)
+            .and_then(|span| self.sess().source_map().span_to_snippet(span).ok())
         {
             (expr_text, true)
         } else {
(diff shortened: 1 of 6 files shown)


compiler/rustc_typeck/src/check/method/prelude2021.rs
```diff
@@ -258,7 +259,7 @@ impl<'a, 'tcx> FnCtxt<'a, 'tcx> {
             let self_ty_name = self
                 .sess()
                 .source_map()
-                .span_to_snippet(self_ty_span)
+                .span_to_snippet(self_ty_span.find_ancestor_inside(span).unwrap_or_default())
```
estebank, line 262: Is `unwrap_or_default` what we want here?

compiler/rustc_typeck/src/check/method/prelude2021.rs
```diff
@@ -353,12 +359,15 @@ impl<'a, 'tcx> FnCtxt<'a, 'tcx> {
...
-        let (expr_text, precise) =
-            if let Ok(expr_text) = self.sess().source_map().span_to_snippet(expr.span) {
-                (expr_text, true)
-            } else {
-                ("(..)".to_string(), false)
-            };
+        let (expr_text, precise) = if let Ok(expr_text) = self
+            .sess()
+            .source_map()
+            .span_to_snippet(expr.span.find_ancestor_inside(outer).unwrap_or_default())
```
estebank, line 365: Same. It feels like it should be `expr.span.find_ancestor_inside(outer).and_then(...`

estebank (approved):
r=me with the current changes after confirming that `unwrap_default()` is the correct code we want to use there.

compiler/rustc_typeck/src/check/method/prelude2021.rs
m-ou-se: The `or_default` means that `span_to_snippet` gets a dummy span in those cases, which will result in the `self_ty.to_string()` case below.

I can rewrite it as a `.and_then` if you prefer.

m-ou-se: Updated it to use and_then.

PR #62282, opened 2019-07-01
Add `take_...` functions to slices

cramertj:
This adds the following associated functions to `[T]`:
- take_front
- take_front_mut
- take_end
- take_end_mut
- take_first
- take_first_mut
- take_last
- take_last_mut

A previous PR that would have added some of this functionality was closed due to inactivity: https://github.com/rust-lang/rust/pull/49173

I personally use these functions fairly often, and I often feel guilty re-hand-rolling them.

One thing to note is that these are just associated functions, rather than methods, because they use `&&Self` rather than `&Self`, which isn't yet a stable `Self` type. It should be backwards-compatible to change this, however, so long as there aren't major method resolution conflicts or similar. AFAIK there aren't any technical blockers to stabilizing `&&Self` method receivers-- it was just left out in order to keep the stable arbitrary_self_types surface small.

PR #77065, opened 2020-09-22
Take 2: Add `take_...` functions to slices

timvermeulen:
Revival of #62282

This PR adds the following slice methods:

- take
- take_mut
- take_first
- take_first_mut
- take_last
- take_last_mut

The first commit is just the code of the previous PR. I removed some tests that hung compilation (see #77062), I can add them back once that's fixed. I also slightly reworded the doc comments as requested in the previous PR, let me know if that can be improved still.

r? @LukasKalbertodt cc @cramertj @nox

PR #49173, opened 2018-03-19
Introduce <&[_]>::split_off and <&str>::split_off

nox:

PR #88502, opened 2021-08-30
Add slice take methods

ibraheemdev:
Revival of #62282

This PR adds the following slice methods:

- `take`
- `take_mut`
- `take_first`
- `take_first_mut`
- `take_last`
- `take_last_mut`

r? @LukasKalbertodt

### DIFF
--- a/library/core/src/slice/mod.rs
+++ b/library/core/src/slice/mod.rs
@@ -10,7 +10,7 @@ use crate::cmp::Ordering::{self, Greater, Less};
 use crate::marker::Copy;
 use crate::mem;
 use crate::num::NonZeroUsize;
-use crate::ops::{FnMut, Range, RangeBounds};
+use crate::ops::{Bound, FnMut, OneSidedRange, Range, RangeBounds};
 use crate::option::Option;
 use crate::option::Option::{None, Some};
 use crate::ptr;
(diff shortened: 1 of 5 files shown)


inquisitivecrystal:
As far as I can tell, LukasKalbertodt is no longer a member of the rust team and that confused the automation.

r? rust-lang/libs-api

JohnCSimon:
ping from triage:
@ibraheemdev  
Returning to you to address merge conflicts and after that, please switch back to `S-waiting-on-review`.
thanks.

crlf0710:
@ibraheemdev Ping from triage, CI is still red here.

WaffleLapkin:
I want to mention that it's actually a fourth take on this feature. Previous attempts (chronological order, first to last):
1. https://github.com/rust-lang/rust/pull/49173
2. https://github.com/rust-lang/rust/pull/62282
3. https://github.com/rust-lang/rust/pull/77065

dtolnay (approved):
Thank you, this looks good. It's hard for me to judge how well these `&mut &'a Self` signatures are going to work in practice, but I am excited for people to try it.

library/core/src/slice/mod.rs
```diff
@@ -82,6 +82,29 @@ pub use index::range;
...
+#[inline]
+fn split_point_of(range: impl OneSidedRange<usize>) -> Option<(Direction, usize)> {
+    use Bound::*;
+
+    Some(match (range.start_bound(), range.end_bound()) {
+        (Unbounded, Excluded(i)) => (Direction::Front, *i),
+        (Unbounded, Included(i)) => (Direction::Front, i.checked_add(1)?),
+        (Excluded(i), Unbounded) => (Direction::Back, i.checked_add(1)?),
+        (Included(i), Unbounded) => (Direction::Back, *i),
+        _ => unreachable!(),
```
AnthonyMikh, line 99: I don't really like the presence of panic call here which optimiser will have to remove. Would it make more sense to make this a method of `OneSidedRange` in order to statically eliminate impossible cases?

library/core/tests/slice.rs
```diff
@@ -2232,3 +2232,112 @@ fn slice_split_array_mut_out_of_bounds() {
...
+    () => {
+        &mut [(); usize::MAX] as _
+    };
+}
+
+take_tests! {
+    slice: &[(); usize::MAX], method: take,
+    (take_in_bounds_max_range_to, (..usize::MAX), Some(EMPTY_MAX), &[(); 0]),
+    (take_oob_max_range_to_inclusive, (..=usize::MAX), None, EMPTY_MAX),
+    (take_in_bounds_max_range_from, (usize::MAX..), Some(&[] as _), EMPTY_MAX),
```
RalfJung, line 2335: This test fails in Miri, and I am trying to find out why... help would be appreciated. [Here's the Zulip thread](https://rust-lang.zulipchat.com/#narrow/stream/269128-miri/topic/Cron.20Job.20Failure.202021-12-02).

RalfJung: Ah, turns out that even with `rustc` this test diverges in debug mode, since it starts a loop of size `usize::MAX`. Miri doesn't optimize so it has no way to run this test.

PR #88503, opened 2021-08-30
Warn when [T; N].into_iter() is ambiguous in the new edition. 

### LINKED ISSUE #88475: array_into_iter does not trigger for existing trait impl
I tried this code:

```rust
#![allow(unused)]

struct FooIter;

trait MyIntoIter {
    fn into_iter(self: Self) -> FooIter;
}

impl<T, const N: usize> MyIntoIter for [T; N] {
    fn into_iter(self: Self) -> FooIter {
        FooIter
    }
}

struct Point;

pub fn foo() {
    let points: [Point; 1] = [Point];
    let y = points.into_iter();
}
```

This code will not emit the `array_into_iter` lint.  When updating to the 2021 edition, the code fails to compile:

```
error[E0034]: multiple applicable items in scope
  --> src/lib.rs:19:20
   |
19 |     let y = points.into_iter();
   |                    ^^^^^^^^^ multiple `into_iter` found
   |
note: candidate #1 is defined in an impl of the trait `MyIntoIter` for the type `[T; N]`
  --> src/lib.rs:10:5
   |
10 |     fn into_iter(self: Self) -> FooIter {
   |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
   = note: candidate #2 is defined in an impl of the trait `IntoIterator` for the type `[T; N]`
help: disambiguate the associated function for candidate #1
   |
19 |     let y = MyIntoIter::into_iter(points);
   |             ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
help: disambiguate the associated function for candidate #2
   |
19 |     let y = IntoIterator::into_iter(points);
   |             ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

I would expect the `array_into_iter` lint to give the appropriate suggestion.

* https://crater-reports.s3.amazonaws.com/pr-87190-3/try%23a7a572ce3edd6d476191fbfe92c9c1986e009b34/gh/dranikpg.simple-rays/log.txt

### Meta

`rustc --version --verbose`:
```
binary: rustc
commit-hash: 5eacec9ec7e112a0de1011519a57c45586d58414
commit-date: 2021-08-28
host: x86_64-apple-darwin
release: 1.56.0-nightly
LLVM version: 13.0.0
```

m-ou-se:
Fixes https://github.com/rust-lang/rust/issues/88475

In https://github.com/rust-lang/rust/issues/88475, a situation was found where `[T; N].into_iter()` becomes *ambiguous* in the new edition. This is different than the case where `(&[T; N]).into_iter()` resolves differently, which was the only case handled by the `array_into_iter` lint. This is almost identical to the new-traits-in-the-prelude problem. Effectively, due to the array-into-iter hack disappearing in Rust 2021, we effectively added `IntoIterator` to the 'prelude' in Rust 2021 specifically for arrays.

This modifies the prelude collisions lint to detect that case and emit a `array_into_iter` lint in that case.

### DIFF
--- a/compiler/rustc_typeck/src/check/method/prelude2021.rs
+++ b/compiler/rustc_typeck/src/check/method/prelude2021.rs
@@ -5,7 +5,7 @@ use rustc_ast::Mutability;
 use rustc_errors::Applicability;
 use rustc_hir as hir;
 use rustc_middle::ty::subst::InternalSubsts;
-use rustc_middle::ty::{Adt, Ref, Ty};
+use rustc_middle::ty::{Adt, Array, Ref, Ty};
 use rustc_session::lint::builtin::RUST_2021_PRELUDE_COLLISIONS;
 use rustc_span::symbol::kw::Underscore;
 use rustc_span::symbol::{sym, Ident};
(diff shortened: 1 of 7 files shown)


compiler/rustc_typeck/src/check/method/prelude2021.rs
```diff
@@ -38,11 +38,28 @@ impl<'a, 'tcx> FnCtxt<'a, 'tcx> {
...
-        if !matches!(segment.ident.name, sym::try_into) {
+        // `try_into` was added to the prelude in Rust 2021.
+        // `into_iter` wasn't, but `[T; N].into_iter()` doesn't resolve to
+        // IntoIterator::into_iter before Rust 2021, which results in the same
+        // problem.
+        if !matches!(segment.ident.name, sym::try_into | sym::into_iter) {
             return;
         }
 
+        let prelude_or_array_lint = if segment.ident.name == sym::into_iter {
```
cjgillot, line 49: Can this test be merged with the previous one, for instance as a `match`?

cjgillot:
r=me with nit fixed.

PR #88505, opened 2021-08-30
Use `unwrap_unchecked` where possible

ibraheemdev:


ibraheemdev:
I wonder if there should be a clippy lint for this.

library/alloc/src/collections/linked_list.rs
```diff
@@ -300,7 +300,7 @@ impl<T> LinkedList<T> {
         let tail = self.tail.take();
         let len = mem::replace(&mut self.len, 0);
         if let Some(head) = head {
-            let tail = tail.unwrap_or_else(|| unsafe { core::hint::unreachable_unchecked() });
+            let tail = unsafe { tail.unwrap_unchecked() };
```
kennytm, line 303:
```suggestion
            // SAFETY: In a LinkedList, either both the head and tail are None because
            // the list is empty, or both head and tail are Some because the list is populated.
            // Since we have verified the head is Some, we are sure the tail is Some too.
            let tail = unsafe { tail.unwrap_unchecked() };
```

kennytm (commented):
the whole linked_list module is full of `unsafe` without any SAFETY annotation, uh-oh.

r=me after including SAFETY of the affected line.

PR #88506, opened 2021-08-30
Fix loading large rlibs

Mark-Simulacrum:
Bumps object crate to permit parsing archives with 64-bit table entries. These
are primarily encountered when there's more than 4GB of archive data.

cc https://github.com/gimli-rs/object/issues/365

Helps with https://github.com/rust-lang/rust/issues/88351, but will also need a beta backport

r? @ehuss (mostly for the test)


Mark-Simulacrum:
beta-accepting myself since if we're OK with master we're probably good with backporting, small patch and fixes a clear regression.

ehuss:
Hm, maybe I'm confused, but I don't think this test can be checked in as-is?  On my relatively fast system it takes almost 15 minutes to finish, and uses 16 GB of disk space.  Was this just to demonstrate that it is fixed?

I don't have any particular ideas on how to make a test that could be feasible.  Perhaps write some hand-crafted ELF files that circumvents `cc`, but it'll still be pretty slow and probably take an excessive amount of disk space.

Otherwise, seems like the test passes locally for me.

Mark-Simulacrum:
Yeah, I think the test as-is does likely take too long. I'll probably drop it from this PR later today- I sort of want a slow test mode but 11-15 minutes seems excessive on any system for a singles test.

Mark-Simulacrum:
@ehuss would you be up for r+'ing this now? I just pushed a commit that deletes the test; I opted for a separate commit because it seems nice to keep the test around in the commit history so we can cherry-pick just the first commit for testing purposes. Happy to squash if you disagree though.

ehuss:
👍 
@bors r+

Not sure if you want to adjust the priority.

Mark-Simulacrum:
@bors p=1

Yeah, I'd like to get this landed so we can have the backport this week without too much hassle.

PR #88511, opened 2021-08-30
x.py clippy: don't run with --all-targets by default

matthiaskrgr:
this caused a lot of noise because benchmarks and tests were also checked

before:
`./x.py clippy |& wc -l`
`74026`

with change:
`./x.py clippy |& wc -l`
`43269`

Zulip thread: https://rust-lang.zulipchat.com/#narrow/stream/122651-general/topic/x.2Epy.20check.20default.20targets

r?  @jyn514 or @Mark-Simulacrum

### DIFF
--- a/src/bootstrap/check.rs
+++ b/src/bootstrap/check.rs
@@ -107,6 +107,11 @@ impl Step for Std {
             add_to_sysroot(&builder, &libdir, &hostdir, &libstd_stamp(builder, compiler, target));
         }
 
+        // don't run on std twice with x.py clippy
+        if builder.kind == Kind::Clippy {
+            return;
+        }
+
         // Then run cargo again, once we've put the rmeta files for the library
         // crates into the sysroot. This is needed because e.g., core's tests
         // depend on `libtest` -- Cargo presumes it will exist, but it doesn't
@@ -120,6 +125,7 @@ impl Step for Std {
             target,
             cargo_subcommand(builder.kind),
         );
+
         cargo.arg("--all-targets");
         std_cargo(builder, target, compiler.stage, &mut cargo);
 
@@ -192,7 +198,12 @@ impl Step for Rustc {
             cargo_subcommand(builder.kind),
         );
         rustc_cargo(builder, &mut cargo, target);
-        cargo.arg("--all-targets");
+
+        // For ./x.py clippy, don't run with --all-targets because
+        // linting tests and benchmarks can produce very noisy results
+        if builder.kind != Kind::Clippy {
+            cargo.arg("--all-targets");
+        }
 
         // Explicitly pass -p for all compiler krates -- this will force cargo
         // to also check the tests/benches/examples for these crates, rather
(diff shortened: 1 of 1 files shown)


src/bootstrap/check.rs
```diff
@@ -120,7 +120,9 @@ impl Step for Std {
             target,
             cargo_subcommand(builder.kind),
         );
-        cargo.arg("--all-targets");
+        if builder.kind != Kind::Clippy {
+            cargo.arg("--all-targets");
+        }
```
jyn514, line 125: Hmm, this makes it so it's no longer possible to pass --all-targets to clippy - is that intentional? I agree it doesn't seem super useful, but it seems weird to remove the option altogether ... although I guess you can still add it back with `CARGOFLAGS=--all-targets`.

jyn514: Also, for this step specifically it should just be an early return above, because we've already run on the standard library; this invocation is *only* for the other targets. The other steps are correct though.

matthiaskrgr: How did that work previously? When I check ` ./x.py clippy --all-targets` pre-branch,  I just get a 
````
Unrecognized option: 'all-targets'

Usage: x.py <subcommand> [options] [<paths>...]
````
error...

Comment (Mark-Simulacrum) [reply src/bootstrap/check.rs:125]: I think at least a comment as to why this isn't being specified is a good idea. We can consider the defaults (or ability to provide it) later.

Comment (jyn514) [reply src/bootstrap/check.rs:125]: > How did that work previously?

See the changes in https://github.com/rust-lang/rust/pull/88011/files.

Comment (matthiaskrgr) [reply src/bootstrap/check.rs:128]: Sure, added some comments.

File: src/bootstrap/check.rs [hunk 2]
```diff
@@ -120,7 +120,12 @@ impl Step for Std {
             target,
             cargo_subcommand(builder.kind),
         );
-        cargo.arg("--all-targets");
+
+        // For ./x.py clippy, don't run with --all-targets because
+        // linting teste and benchmarks can produce very noisy results
```
jyn514, line 125:
```suggestion
        // linting tests and benchmarks can produce very noisy results
```

jyn514 (changes requested):
r=me with https://github.com/rust-lang/rust/pull/88511/files#r698829310 done.

jyn514:
@matthiaskrgr did you miss https://github.com/rust-lang/rust/pull/88511/files#r698829310 ?

matthiaskrgr:
Ah yes, I missed that.
I excluded  an early return to `run()` of `Std`, I hope that's what you ment`:
```
        if builder.kind != Kind::Clippy {
            return;
        }
````

matthiaskrgr:
Ok that does not seem to work... :sweat_smile:

jyn514:
@matthiaskrgr I meant here: https://github.com/rust-lang/rust/blob/e7c15b95174e5bc78b51a7887ac9e6db3468d65b/src/bootstrap/check.rs#L114

The idea is to avoid running cargo twice, not to skip it altogether.

jyn514:
(or in other words, your "handled elsewhere" comment is incorrect)

matthiaskrgr:
@jyn514 I think this is ready to be reviewed again..? :)

jyn514:
Oh sorry, didn't realize this was waiting on me. Could you switch the labels next time?

@bors r+ rollup

PR #84147, opened 2021-04-13
Cautiously add IntoIterator for arrays by value

cuviper:
Add the attribute described in #84133, `#[rustc_skip_array_during_method_dispatch]`, which effectively hides a trait from method dispatch when the receiver type is an array.

Then cherry-pick `IntoIterator for [T; N]` from #65819 and gate it with that attribute. Arrays can now be used as `IntoIterator` normally, but `array.into_iter()` has edition-dependent behavior, returning `slice::Iter` for 2015 and 2018 editions, or `array::IntoIter` for 2021 and later.

r? @nikomatsakis 
cc @LukasKalbertodt @rust-lang/libs

PR #88512, opened 2021-08-30
Upgrade array_into_iter lint to include Deref-to-array types.

### LINKED ISSUE #88099: array_into_iter does not fire for lazy_static arrays
I tried this code:

```rust
struct Bar {}

lazy_static::lazy_static! {
    static ref FOO: [Bar; 1] = [
        Bar{}
    ];
}

fn main() {
    let _ = FOO.into_iter();
}
```

I expected to see this happen: [`array_into_iter`](https://doc.rust-lang.org/nightly/rustc/lints/listing/warn-by-default.html#array-into-iter) should fire for this.  Or, maybe this shouldn't fail in 2021?

Instead, this happened: Lint does not fire.  When migrating to 2021, this fails to compile:

```
error[E0508]: cannot move out of type `[Bar; 1]`, a non-copy array
  --> src/main.rs:10:13
   |
10 |     let _ = FOO.into_iter();
   |             ^^^
   |             |
   |             cannot move out of here
   |             move occurs because value has type `[Bar; 1]`, which does not implement the `Copy` trait
```

`lazy_static` uses a Deref to access the underlying value.  The expanded code looks like this:

```rust
struct FOO {
    __private_field: (),
}
#[doc(hidden)]
static FOO: FOO = FOO {
    __private_field: (),
};
impl ::lazy_static::__Deref for FOO {
    type Target = [Bar; 1];
    fn deref(&self) -> &[Bar; 1] {
        #[inline(always)]
        fn __static_ref_initialize() -> [Bar; 1] {
            [Bar {}]
        }
        #[inline(always)]
        fn __stability() -> &'static [Bar; 1] {
            static LAZY: ::lazy_static::lazy::Lazy<[Bar; 1]> = ::lazy_static::lazy::Lazy::INIT;
            LAZY.get(__static_ref_initialize)
        }
        __stability()
    }
}
impl ::lazy_static::LazyStatic for FOO {
    fn initialize(lazy: &Self) {
        let _ = &**lazy;
    }
}
```

I don't know if it is feasible to have a fix suggestion for this, or why exactly the existing lint doesn't fire.

### Meta

`rustc --version --verbose`:
```
rustc 1.56.0-nightly (8007b506a 2021-08-14)
binary: rustc
commit-hash: 8007b506ac5da629f223b755f5a5391edd5f6d01
commit-date: 2021-08-14
host: x86_64-apple-darwin
release: 1.56.0-nightly
LLVM version: 12.0.1
```

m-ou-se:
Fixes https://github.com/rust-lang/rust/issues/88099

Fixes the issue mentioned here: https://github.com/rust-lang/rust/pull/84147#issuecomment-819000436

### DIFF
--- a/src/test/ui/iterators/into-iter-on-arrays-lint.stderr
+++ b/src/test/ui/iterators/into-iter-on-arrays-lint.stderr
@@ -71,137 +71,73 @@ warning: this method call resolves to `<&[T; N] as IntoIterator>::into_iter` (du
   --> $DIR/into-iter-on-arrays-lint.rs:23:21
    |
 LL |     Box::new(small).into_iter();
-   |                     ^^^^^^^^^
+   |                     ^^^^^^^^^ help: use `.iter()` instead of `.into_iter()` to avoid ambiguity: `iter`
    |
    = warning: this changes meaning in Rust 2021
    = note: for more information, see <https://doc.rust-lang.org/nightly/edition-guide/rust-2021/IntoIterator-for-arrays.html>
-help: use `.iter()` instead of `.into_iter()` to avoid ambiguity
-   |
-LL |     Box::new(small).iter();
-   |                     ~~~~
-help: or use `IntoIterator::into_iter(..)` instead of `.into_iter()` to explicitly iterate by value
-   |
-LL |     IntoIterator::into_iter(Box::new(small));
-   |     ++++++++++++++++++++++++               ~
 
 warning: this method call resolves to `<&[T; N] as IntoIterator>::into_iter` (due to backwards compatibility), but will resolve to <[T; N] as IntoIterator>::into_iter in Rust 2021.
   --> $DIR/into-iter-on-arrays-lint.rs:26:22
    |
 LL |     Box::new([1, 2]).into_iter();
-   |                      ^^^^^^^^^
+   |                      ^^^^^^^^^ help: use `.iter()` instead of `.into_iter()` to avoid ambiguity: `iter`
    |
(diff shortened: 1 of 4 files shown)


m-ou-se:
cc @estebank because diagnostics

src/test/ui/iterators/into-iter-on-arrays-lint.stderr
```diff
@@ -71,137 +71,73 @@ warning: this method call resolves to `<&[T; N] as IntoIterator>::into_iter` (du
...
    = warning: this changes meaning in Rust 2021
    = note: for more information, see <https://doc.rust-lang.org/nightly/edition-guide/rust-2021/IntoIterator-for-arrays.html>
-help: use `.iter()` instead of `.into_iter()` to avoid ambiguity
-   |
-LL |     Box::new(small).iter();
-   |                     ~~~~
-help: or use `IntoIterator::into_iter(..)` instead of `.into_iter()` to explicitly iterate by value
-   |
-LL |     IntoIterator::into_iter(Box::new(small));
-   |     ++++++++++++++++++++++++               ~
```
estebank, line 85: Do we still emit this suggestion in any case?

estebank (approved):
The changes look reasonable to me.

src/test/ui/iterators/into-iter-on-arrays-lint.stderr
m-ou-se: Yes, for `[1, 2, 3].into_iter()`. See line 65 in this file.

PR #88516, opened 2021-08-30
some low hanging clippy::perf fixes

matthiaskrgr:


jyn514:
CI failure looks spurious, cc @GuillaumeGomez

GuillaumeGomez:
Failure looks spurious, weird...

jyn514:
@GuillaumeGomez right - my point is that test is flaky and we should find a way to make it less flaky. 30 seconds is already a long timeout so we probably don't want to increase it - is there a way to make whatever it's doing faster?

GuillaumeGomez:
There is no explicit timeout here, we just told the test to wait for a selector and that selector didn't appear in the (default) 30 seconds timeout. I think the problem is elsewhere.

jyn514:
Hmm, ok. Maybe it's not spurious then. @matthiaskrgr can you run this locally and see whether it passes?

@bors r-

matthiaskrgr:
`./x.py test rustdoc ` succeeds locally. I rebased, let's see if that changes anything.

GuillaumeGomez:
> `./x.py test rustdoc ` succeeds locally. I rebased, let's see if that changes anything.

Is it running `rustdoc-gui`? If not we might want to fix that (I never ran it like this). Otherwise: `x.py test src/test/rustdoc-gui --stage 1`.

matthiaskrgr:
Ah no, apparently `x.py test rustdoc` does not run the `rustdoc-gui` tests. (got an error because of some missing npm package).
But I managed to reproduce the test failure at least! :)

GuillaumeGomez:
Time to fix it then! :p

matthiaskrgr:
So, I went through all the changes one by one now and applied them and I got a lot of spurious test failures:
````
code-sidebar-toggle... FAILED
[ERROR] (line 3) Error: No node found for selector: #sidebar-toggle: for command `click: "#sidebar-toggle"` 

sidebar... FAILED
[ERROR] (line 44) expected 1 elements, found 0: for command `assert-count: (".sidebar .location", 1)`
````
but after running `./x.py test src/test/rustdoc-gui --stage 1` a couple of times I always got to a test run which succeeded at some point, so to me it looks like the tests are just flaky :/

GuillaumeGomez:
This is highly surprising and a bit worrying. Well, if it isn't due to changes in this PR, let's move forward. I'll keep a close eye on such flakiness in the future.

@bors: r=jyn514,GuillaumeGomez

GuillaumeGomez:
Since it's changing performance:

@bors: rollup=iffy

jyn514:
Apple job failed without logs.

@bors retry

PR #88379, opened 2021-08-27
rustdoc: Cleanup various `clean` types

camelid:
Cleanup various `clean` types.

Issue #83718, opened 2021-03-31
[rustdoc-json] Assertion error on blanket impls


aDotInTheVoid:
```rust
pub trait Get {
    type Target;
    fn get(&self) -> &Self::Target;
}

pub trait LoadNum {
    fn load_num(&self) -> i32;
}

impl<A: LoadNum, P: Get<Target = A>> LoadNum for P {
    fn load_num(&self) -> i32 {
        self.get().load_num()
    }
}

pub struct Wrapper<T>(T);
impl<T> Get for Wrapper<T> {
    type Target = T;
    fn get(&self) -> &T {
        &self.0
    }
}
```

`rustdoc +stage1 trait.rs -w json`

```
thread 'rustc' panicked at 'assertion failed: `(left == right)`

Diff < left / right > :
 Item {
     id: Id(
         "0:11",
     ),
     crate_id: 0,
     name: Some(
         "load_num",
     ),
     span: Some(
         Span {
             filename: "trait.rs",
             begin: (
                 11,
                 4,
             ),
             end: (
                 13,
                 5,
             ),
         },
     ),
<    visibility: Default,
>    visibility: Public,
     docs: None,
     links: {},
     attrs: [],
     deprecation: None,
     inner: Method(
         Method {
             decl: FnDecl {
                 inputs: [
                     (
<                        "self",
>                        "",
                         BorrowedRef {
                             lifetime: None,
                             mutable: false,
                             type_: Generic(
                                 "Self",
                             ),
                         },
                     ),
                 ],
                 output: Some(
                     Primitive(
                         "i32",
                     ),
                 ),
                 c_variadic: false,
             },
             generics: Generics {
                 params: [],
                 where_predicates: [],
             },
             header: {},
             abi: "\"Rust\"",
             has_body: true,
         },
     ),
 }

', src/librustdoc/json/mod.rs:179:17
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

Interesingly, and shockingly If you remove `pub struct Wrapper<T>(T);` and `impl<T> Get for Wrapper<T>`, this doesn't panic.

```
rustdoc +stage1 --version --verbose
rustdoc 1.53.0-nightly (a5029ac0a 2021-03-31)
binary: rustdoc
commit-hash: a5029ac0ab372aec515db2e718da6d7787f3d122
commit-date: 2021-03-31
host: x86_64-unknown-linux-gnu
release: 1.53.0-nightly
LLVM version: 12.0.0
```

[...]

PR #88522, opened 2021-08-31
rustdoc: Box `GenericArgs::Parenthesized.output`

camelid:
Split out from #88379.

This reduces the size of `GenericArgs` from 104 bytes to 56 bytes,
essentially reducing it by half.

`GenericArgs` is one of the fields of `PathSegment`, so this should
reduce the amount of memory allocated for `PathSegment`s in the cases
where the generics are not for a `Fn`, `FnMut`, or `FnOnce` trait.

r? @jyn514

### DIFF
--- a/src/librustdoc/clean/auto_trait.rs
+++ b/src/librustdoc/clean/auto_trait.rs
@@ -350,10 +350,8 @@ impl<'a, 'tcx> AutoTraitFinder<'a, 'tcx> {
             .into_iter()
             .flat_map(|(ty, mut bounds)| {
                 if let Some(data) = ty_to_fn.get(&ty) {
-                    let (poly_trait, output) = (
-                        data.0.as_ref().expect("as_ref failed").clone(),
-                        data.1.as_ref().cloned().map(Box::new),
-                    );
+                    let (poly_trait, output) =
+                        (data.0.as_ref().unwrap().clone(), data.1.as_ref().cloned().map(Box::new));
                     let new_ty = match poly_trait.trait_ {
                         Type::ResolvedPath { ref path, ref did, ref is_generic } => {
                             let mut new_path = path.clone();
--- a/src/librustdoc/clean/auto_trait.rs
+++ b/src/librustdoc/clean/auto_trait.rs
@@ -350,8 +350,10 @@ impl<'a, 'tcx> AutoTraitFinder<'a, 'tcx> {
             .into_iter()
             .flat_map(|(ty, mut bounds)| {
                 if let Some(data) = ty_to_fn.get(&ty) {
-                    let (poly_trait, output) =
-                        (data.0.as_ref().expect("as_ref failed").clone(), data.1.as_ref().cloned());
+                    let (poly_trait, output) = (
+                        data.0.as_ref().expect("as_ref failed").clone(),
+                        data.1.as_ref().cloned().map(Box::new),
+                    );
                     let new_ty = match poly_trait.trait_ {
                         Type::ResolvedPath { ref path, ref did, ref is_generic } => {
                             let mut new_path = path.clone();
--- a/src/librustdoc/clean/mod.rs
+++ b/src/librustdoc/clean/mod.rs
@@ -1763,10 +1763,9 @@ impl Clean<GenericArgs> for hir::GenericArgs<'_> {
     fn clean(&self, cx: &mut DocContext<'_>) -> GenericArgs {
         if self.parenthesized {
             let output = self.bindings[0].ty().clean(cx);
-            GenericArgs::Parenthesized {
-                inputs: self.inputs().clean(cx),
-                output: if output != Type::Tuple(Vec::new()) { Some(output) } else { None },
-            }
+            let output =
+                if output != Type::Tuple(Vec::new()) { Some(Box::new(output)) } else { None };
+            GenericArgs::Parenthesized { inputs: self.inputs().clean(cx), output }
         } else {
             GenericArgs::AngleBracketed {
                 args: self
--- a/src/librustdoc/clean/types.rs
+++ b/src/librustdoc/clean/types.rs
@@ -2018,14 +2018,24 @@ crate enum GenericArg {
 #[derive(Clone, PartialEq, Eq, Debug, Hash)]
 crate enum GenericArgs {
     AngleBracketed { args: Vec<GenericArg>, bindings: Vec<TypeBinding> },
-    Parenthesized { inputs: Vec<Type>, output: Option<Type> },
+    Parenthesized { inputs: Vec<Type>, output: Option<Box<Type>> },
 }
+
+// `GenericArgs` is in every `PathSegment`, so its size can significantly
+// affect rustdoc's memory usage.
+#[cfg(all(target_arch = "x86_64", target_pointer_width = "64"))]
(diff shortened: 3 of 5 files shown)


src/librustdoc/clean/auto_trait.rs
```diff
@@ -350,8 +350,10 @@ impl<'a, 'tcx> AutoTraitFinder<'a, 'tcx> {
             .into_iter()
             .flat_map(|(ty, mut bounds)| {
                 if let Some(data) = ty_to_fn.get(&ty) {
-                    let (poly_trait, output) =
-                        (data.0.as_ref().expect("as_ref failed").clone(), data.1.as_ref().cloned());
+                    let (poly_trait, output) = (
+                        data.0.as_ref().expect("as_ref failed").clone(),
```
camelid, line 354: I just noticed that this is a really odd `expect` message. We should probably just use `unwrap()`.

camelid:
max-rss looks to have improved by between 0.5% and 1.6% for several benchmarks, with no significant regressions. instruction count is unchanged.

src/librustdoc/clean/mod.rs
```diff
@@ -1763,10 +1763,9 @@ impl Clean<GenericArgs> for hir::GenericArgs<'_> {
     fn clean(&self, cx: &mut DocContext<'_>) -> GenericArgs {
         if self.parenthesized {
             let output = self.bindings[0].ty().clean(cx);
-            GenericArgs::Parenthesized {
-                inputs: self.inputs().clean(cx),
-                output: if output != Type::Tuple(Vec::new()) { Some(output) } else { None },
-            }
+            let output =
+                if output != Type::Tuple(Vec::new()) { Some(Box::new(output)) } else { None };
```
jyn514, line 1767: Personally I would use `box output` here :shrug: but it won't affect performance in practice, `output` is already on the stack.

Oh, this is the bit that breaks https://github.com/rust-lang/rust/issues/83718#issuecomment-811453368 ! I've been looking for it.

src/librustdoc/json/conversions.rs
```diff
@@ -127,7 +127,7 @@ impl FromWithTcx<clean::GenericArgs> for GenericArgs {
             },
             Parenthesized { inputs, output } => GenericArgs::Parenthesized {
                 inputs: inputs.into_iter().map(|a| a.into_tcx(tcx)).collect(),
-                output: output.map(|a| a.into_tcx(tcx)),
+                output: output.map(|a| (*a).into_tcx(tcx)),
```
jyn514, line 130: nit: this could be simpler:
```suggestion
                output: output.map(|&a| a.into_tcx(tcx)),
```
In general I don't know that the IntoTcx trait pulls its weight, but it certainly shouldn't block this PR.

src/librustdoc/clean/types.rs
```diff
@@ -2018,7 +2018,7 @@ crate enum GenericArg {
 #[derive(Clone, PartialEq, Eq, Debug, Hash)]
 crate enum GenericArgs {
     AngleBracketed { args: Vec<GenericArg>, bindings: Vec<TypeBinding> },
-    Parenthesized { inputs: Vec<Type>, output: Option<Type> },
+    Parenthesized { inputs: Vec<Type>, output: Option<Box<Type>> },
 }
```
jyn514, line 2022: Can you add a `rustc_assert_size` so this doesn't grow again by accident? As we've seen, the perf bot won't catch it if it regresses.

jyn514 (approved):
The `rustc_assert_size` is the only change I feel strongly about, the rest you can choose to add or not. Good catch! I wouldn't have found this :)

src/librustdoc/json/conversions.rs
camelid: Are you sure `&` patterns work with `Box`? They don't seem to on the Playground. I don't think Deref patterns have been implemented yet.

src/librustdoc/clean/mod.rs
camelid: > Personally I would use `box output` here 🤷

Yeah, I thought about that, but someone recently removed all the uses of `box_syntax` from rustdoc and other tools; I think the goal is to get rid of `box_syntax` eventually. IIRC from the discussion there, the newest version of LLVM has an optimization that gets rid of the perf difference, but I may be misremembering.

src/librustdoc/clean/types.rs
camelid: Sure! I was actually about to do that even before I saw this comment :)

src/librustdoc/clean/mod.rs
camelid: > Oh, this is the bit that breaks [#83718 (comment)](https://github.com/rust-lang/rust/issues/83718#issuecomment-811453368) ! I've been looking for it.

Hooray! 🎉

src/librustdoc/json/conversions.rs
camelid: Yeah, it doesn't seem to work:

```
   --> src/librustdoc/json/conversions.rs:130:37
    |
130 |                 output: output.map(|&a| a.into_tcx(tcx)),
    |                                     ^--
    |                                     ||
    |                                     |expected due to this
    |                                     expected struct `std::boxed::Box`, found reference
    |                                     help: did you mean `a`: `&std::boxed::Box<types::Type>`
    |
    = note: expected struct `std::boxed::Box<types::Type>`
            found reference `&_`
```

jyn514: Oh, I forgot this was a smart pointer. I think `box a` would work, but IDK how much simpler it is at that point.

camelid: Yeah, I think it would work, but it seems more complex in some ways. I'll just leave it as `(*x)` for now.

src/librustdoc/clean/types.rs
```diff
@@ -2018,15 +2018,23 @@ crate enum GenericArg {
...
 #[derive(Clone, PartialEq, Eq, Debug, Hash)]
 crate struct PathSegment {
     crate name: Symbol,
     crate args: GenericArgs,
 }
 
+// `PathSegment` occurs multiple times in every `Path`, so its size can
+// significantly affect rustdoc's memory usage.
+rustc_data_structures::static_assert_size!(PathSegment, 64);
+
```
jyn514, line 2037: Hmm, this is a good point - I wonder whether it hurts or helps to box GenericArgs instead of `output`? Are you interested in making another perf run with that change? (no problem if not, this is still a good perf improvement as-is)

jyn514 (approved):
r=me unless you want to try boxing GenericArgs

src/librustdoc/clean/types.rs
camelid: Huh, I just saw your comment now; I wonder if there was a glitch in GitHub.

The thing is that `GenericArgs` is constructed for every `PathSegment`, so I think rustdoc would still be allocating the same amount of memory as without the `Box` (in fact, a bit more, because of the `usize` for the `Box`). Whereas with the change I made, the `Box` is only constructed if (a) the `GenericArgs` is for `Fn` sugar and (b) the `Fn` sugar has an output type.

camelid: Actually, writing up that comment just gave me an idea for another way to improve perf :D

PR #88525, opened 2021-08-31
fix(rustc_typeck): produce better errors for dyn auto trait

### LINKED ISSUE #85026: `impl dyn Trait` on an auto trait produces a confusing diagnostic
Given the following code ([playground](https://play.rust-lang.org/?version=nightly&mode=debug&edition=2018&gist=850ae3ba756280ffd028d7342926adea)):

```rust
impl dyn Unpin {}
```

The current output is:

```
error[E0118]: no nominal type found for inherent implementation
 --> src/lib.rs:1:6
  |
1 | impl dyn Unpin {}
  |      ^^^^^^^^^ impl requires a nominal type
  |
  = note: either implement a trait on it or create a newtype to wrap it instead
```

This is rather confusing. It's not immediately clear what the difference between `impl dyn Unpin` and this working code is ([playground](https://play.rust-lang.org/?version=nightly&mode=debug&edition=2018&gist=c160754c43fd35dded133da1be1352ad)):
```rust
trait A {}
impl dyn A {}
```

Since `impl dyn A` is accepted, the user might be confused why `dyn A` would be considered a nominal type when `dyn Unpin` isn't. The real reason why `impl dyn Unpin` is an error is because it is an `impl` block for ~~a type~~ [edit: a dynamically dispatched auto trait] from another crate, which should be handled by [`E0116`](https://doc.rust-lang.org/error-index.html#E0116).

```
error[E0116]: cannot define inherent `impl` for a type outside of the crate where the type is defined
 --> src/lib.rs:1:1
  |
1 | impl dyn Unpin {}
  | ^^^^^^^^^^^^^^^^^ impl for type defined outside of crate.
  |
  = note: define and implement a trait or new type instead
```

notriddle:
Fixes #85026

### DIFF
--- a/compiler/rustc_typeck/src/coherence/inherent_impls.rs
+++ b/compiler/rustc_typeck/src/coherence/inherent_impls.rs
@@ -60,6 +60,17 @@ impl ItemLikeVisitor<'v> for InherentCollect<'tcx> {
             ty::Dynamic(ref data, ..) if data.principal_def_id().is_some() => {
                 self.check_def_id(item, data.principal_def_id().unwrap());
             }
+            ty::Dynamic(..) => {
+                struct_span_err!(
+                    self.tcx.sess,
+                    ty.span,
+                    E0785,
+                    "cannot define inherent `impl` for a dyn auto trait"
+                )
+                .span_label(ty.span, "impl requires a principal trait")
+                .note("define and implement a new trait or type instead")
+                .emit();
+            }
             ty::Bool => {
                 self.check_primitive_impl(
                     item.def_id,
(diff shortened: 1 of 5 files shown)


GuillaumeGomez (approved):

compiler/rustc_typeck/src/coherence/inherent_impls.rs
```diff
@@ -60,6 +60,17 @@ impl ItemLikeVisitor<'v> for InherentCollect<'tcx> {
...
                 self.check_def_id(item, data.principal_def_id().unwrap());
             }
+            ty::Dynamic(..) => {
+                struct_span_err!(
+                    self.tcx.sess,
+                    ty.span,
+                    E0785,
+                    "cannot define inherent `impl` for a dyn auto trait"
+                )
+                .span_label(ty.span, "impl requires a principal trait")
```
petrochenkov, line 70:
```suggestion
                .span_label(ty.span, "impl requires at least one non-auto trait")
```
"Principal trait" is an internal terminology that nobody is familiar with.

petrochenkov:
r=me with the wording change.

notriddle:
@petrochenkov Okay, the wording changes are in.

PR #88529, opened 2021-08-31
ARMv6K Nintendo 3DS Tier 3 target added

Meziu:
Addition of the target specifications to build .elf files for Nintendo 3DS (ARMv6K, Horizon). Requires devkitARM 3DS toolkit for system libraries and arm-none-eabi-gcc linker.

### DIFF
--- a/compiler/rustc_target/src/spec/armv6k_nintendo_3ds.rs
+++ b/compiler/rustc_target/src/spec/armv6k_nintendo_3ds.rs
@@ -1,0 +1,45 @@
+use crate::spec::{LinkArgs, LinkerFlavor, PanicStrategy, RelocModel, Target, TargetOptions};
+
+/// A base target for Nintendo 3DS devices using the devkitARM toolchain.
+///
+/// Requires the devkitARM toolchain for 3DS targets on the host system.
+
+pub fn target() -> Target {
+    let mut pre_link_args = LinkArgs::new();
+    pre_link_args.insert(
+        LinkerFlavor::Gcc,
+        vec![
+            "-specs=3dsx.specs".to_string(),
+            "-g".to_string(),
+            "-march=armv6k".to_string(),
+            "-mtune=mpcore".to_string(),
+            "-mfloat-abi=hard".to_string(),
+            "-mtp=soft".to_string(),
+        ],
+    );
+
+    Target {
+        llvm_target: "arm-none-eabihf".to_string(),
+        pointer_width: 32,
+        data_layout: "e-m:e-p:32:32-Fi8-i64:64-v128:64:128-a:0:32-n32-S64".to_string(),
+        arch: "arm".to_string(),
+
+        options: TargetOptions {
+            os: "horizon".to_string(),
+            env: "newlib".to_string(),
+            vendor: "nintendo".to_string(),
+            abi: "eabihf".to_string(),
+            linker_flavor: LinkerFlavor::Gcc,
+            cpu: "mpcore".to_string(),
+            executables: true,
+            families: vec!["unix".to_string()],
+            linker: Some("arm-none-eabi-gcc".to_string()),
+            relocation_model: RelocModel::Static,
+            features: "+vfp2".to_string(),
+            pre_link_args,
+            exe_suffix: ".elf".to_string(),
+            panic_strategy: PanicStrategy::Abort,
+            ..Default::default()
+        },
+    }
+}
--- a/compiler/rustc_target/src/spec/armv6k_nintendo_3ds.rs
+++ b/compiler/rustc_target/src/spec/armv6k_nintendo_3ds.rs
@@ -10,8 +10,6 @@ pub fn target() -> Target {
         LinkerFlavor::Gcc,
         vec![
             "-specs=3dsx.specs".to_string(),
-            "-g".to_string(),
-            "-march=armv6k".to_string(),
             "-mtune=mpcore".to_string(),
             "-mfloat-abi=hard".to_string(),
             "-mtp=soft".to_string(),
@@ -19,7 +17,7 @@ pub fn target() -> Target {
     );
 
     Target {
-        llvm_target: "arm-none-eabihf".to_string(),
+        llvm_target: "armv6k-none-eabihf".to_string(),
         pointer_width: 32,
         data_layout: "e-m:e-p:32:32-Fi8-i64:64-v128:64:128-a:0:32-n32-S64".to_string(),
         arch: "arm".to_string(),
--- a/compiler/rustc_target/src/spec/mod.rs
+++ b/compiler/rustc_target/src/spec/mod.rs
@@ -941,6 +941,8 @@ supported_targets! {
 
     ("bpfeb-unknown-none", bpfeb_unknown_none),
     ("bpfel-unknown-none", bpfel_unknown_none),
+
+    ("armv6k-nintendo-3ds", armv6k_nintendo_3ds),
 }
 
 /// Warnings encountered when parsing the target `json`.
--- a/src/doc/rustc/src/platform-support.md
+++ b/src/doc/rustc/src/platform-support.md
@@ -217,6 +217,7 @@ target | std | host | notes
 `armv5te-unknown-linux-uclibceabi` | ? |  | ARMv5TE Linux with uClibc
 `armv6-unknown-freebsd` | ✓ | ✓ | ARMv6 FreeBSD
 `armv6-unknown-netbsd-eabihf` | ? |  |
+`armv6k-nintendo-3ds` | * |  | ARMv6K Nintendo 3DS, Horizon (Requires devkitARM toolchain)
 `armv7-apple-ios` | ✓ |  | ARMv7 iOS, Cortex-a8
 `armv7-unknown-freebsd` | ✓ | ✓ | ARMv7 FreeBSD
 `armv7-unknown-netbsd-eabihf` | ✓ | ✓ |


compiler/rustc_target/src/spec/armv6k_nintendo_3ds.rs
```diff
@@ -0,0 +1,45 @@
...
+/// A base target for Nintendo 3DS devices using the devkitARM toolchain.
+///
+/// Requires the devkitARM toolchain for 3DS targets on the host system.
+
+pub fn target() -> Target {
+    let mut pre_link_args = LinkArgs::new();
+    pre_link_args.insert(
+        LinkerFlavor::Gcc,
+        vec![
+            "-specs=3dsx.specs".to_string(),
```
jonas-schievink, line 12: Who provides this file? devkitARM?

Meziu: Yes, together with the linker and system libraries (the file itself includes a link script and the C Runtime objects).

nagisa:
Please copy the [list of T3 target requirements](https://doc.rust-lang.org/nightly/rustc/target-tier-policy.html#tier-3-target-policy) and confirm this target satisfies each point.

Meziu:
A proposed target or target-specific patch that substantially changes code shared with other targets (not just target-specific code) must be reviewed and approved by the appropriate team for that shared code before acceptance.

>A tier 3 target must have a designated developer or developers (the "target maintainers") on record to be CCed when issues arise regarding the target. (The mechanism to track and CC such developers may evolve over time.)

I will take responsibilty of updating the target and everything related in case of issues with it as long as I can.


>Targets must use naming consistent with any existing targets; for instance, a target for the same CPU or OS as an existing Rust target should use the same name for that CPU or OS. Targets should normally use the same names and naming conventions as used elsewhere in the broader ecosystem beyond Rust (such as in other toolchains), unless they have a very good reason to diverge. Changing the name of a target can be highly disruptive, especially once the target reaches a higher tier, so getting the name right is important even for a tier 3 target.

The target name ( `armv6k-nintendo-3ds`) reflects both CPU, vendor and model of device targeted. It is also coherent with the Rust targets' naming conventions (`CPU/instruction set` - `vendor` - `model`). The OS is inherent to the device model as it's built into the device ROM and only works on that exact set of hardware.


>Tier 3 targets may have unusual requirements to build or use, but must not create legal issues or impose onerous legal terms for the Rust project or for Rust developers or users.

What I included in the target specification is only a communication to open-source and license compatible tools, derivatives of GCC and Newlib. No closed source binary/library or other sort of project is linked in any way from this target specification.


>Neither this policy nor any decisions made regarding targets shall create any binding agreement or estoppel by any party. If any member of an approving Rust team serves as one of the maintainers of a target, or has any legal or employment requirement (explicit or implicit) that might affect their decisions regarding a target, they must recuse themselves from any approval decisions regarding the target's tier status, though they may otherwise participate in discussions.

I understand.


>Tier 3 targets should attempt to implement as much of the standard libraries as possible and appropriate (core for most targets, alloc for targets that can support dynamic memory allocation, std for targets with an operating system or equivalent layer of system-provided functionality), but may leave some code unimplemented (either unavailable or stubbed out as appropriate), whether because the target makes it impossible to implement or challenging to implement. The authors of pull requests are not obligated to avoid calling any portions of the standard library on the basis of a tier 3 target not implementing those portions.

Though `std` support isn't yet available (in any capacity), both core and alloc are supported already for this target and there isn't any "offlimit" code for other contributors to use.


>The target must provide documentation for the Rust community explaining how to build for the target, using cross-compilation if possible. If the target supports running tests (even if they do not pass), the documentation must explain how to run tests for the target, using emulation if possible or dedicated hardware if necessary.

Tests aren't possible in any capacity, but (with the cited devkitARM toolchain installed properly on the host system), cross-compilation is the same as any other `no-std` target, requiring the unstable feature `build-std` with the `core` and `alloc` crates activated.


>Tier 3 targets must not impose burden on the authors of pull requests, or other developers in the community, to maintain the target. In particular, do not post comments (automated or manual) on a PR that derail or suggest a block on the PR based on a tier 3 target. Do not send automated messages or notifications (via any medium, including via @) to a PR author or others involved with a PR regarding a tier 3 target, unless they have opted into such messages. 

I understand and, as already said, will provide support in case of issues with the target as long as I can.


>Patches adding or updating tier 3 targets must not break any existing tier 2 or tier 1 target, and must not knowingly break another tier 3 target without approval of either the compiler team or the maintainers of the other tier 3 target. 

No other code has been modified or deleted other than the specification of the new target, everything works as expected.


>If a tier 3 target stops meeting these requirements, or the target maintainers no longer have interest or time, or the target shows no signs of activity and has not built for some time, or removing the target would improve the quality of the Rust codebase, we may post a PR to remove it; any such PR will be CCed to the target maintainers (and potentially other people who have previously worked on the target), to check potential interest in improving the situation.

I understand and will try to be present in case of any decision made by the Rust maintainers regarding this target implementation.

compiler/rustc_target/src/spec/armv6k_nintendo_3ds.rs
```diff
@@ -0,0 +1,45 @@
...
+            "-g".to_string(),
+            "-march=armv6k".to_string(),
+            "-mtune=mpcore".to_string(),
```
nagisa, line 15: Do the `-g`, `-march` and `-mtune` arguments do anything here? `gcc` here does not compile any code, so intuitively unless there's some link-time codegen going on, these shouldn't be doing anything.

compiler/rustc_target/src/spec/armv6k_nintendo_3ds.rs
```diff
@@ -0,0 +1,45 @@
...
+            "-mfloat-abi=hard".to_string(),
+            "-mtp=soft".to_string(),
+        ],
+    );
+
+    Target {
+        llvm_target: "arm-none-eabihf".to_string(),
```
nagisa, line 22: `armv6k-none-eabihf`?

compiler/rustc_target/src/spec/armv6k_nintendo_3ds.rs
```diff
@@ -0,0 +1,45 @@
...
+        options: TargetOptions {
+            os: "horizon".to_string(),
+            env: "newlib".to_string(),
+            vendor: "nintendo".to_string(),
+            abi: "eabihf".to_string(),
+            linker_flavor: LinkerFlavor::Gcc,
+            cpu: "mpcore".to_string(),
+            executables: true,
+            families: vec!["unix".to_string()],
+            linker: Some("arm-none-eabi-gcc".to_string()),
```
nagisa, line 36: For cross-compilation purposes users of `rustc` are expected to provide the correct `-Clinker` flag, and this field is typically only used for specifying linkers other than `gcc` (such as `rust-lld`). That said I can't imagine this being a wrong default either, so LGTM.

nagisa:
> Though std support isn't yet available (in any capacity), both core and alloc are supported already for this target and there isn't any "offlimit" code for other contributors to use.

Is `std` planned to be supported in any way? Just curious.

---

LGTM r=me pending resolution to the comments inline.

compiler/rustc_target/src/spec/armv6k_nintendo_3ds.rs
Meziu: My objective is to make it as straightforward as possible, especially when using cargo. I think it's fine this way.

compiler/rustc_target/src/spec/armv6k_nintendo_3ds.rs
Meziu: I noticed no difference, but I changed it anyways to avoid confusion.

compiler/rustc_target/src/spec/armv6k_nintendo_3ds.rs
Meziu: I knew they were needed but didn't check them one at a time. Turns out `mtune` is the only one required of the 3 during linking. Thanks for the help.

Meziu:
> Is `std` planned to be supported in any way? Just curious.

The Nintendo 3DS has had multiple out-of-tree projects porting parts or all of the standard library. The 3DS is a system not very powerful, but with many features, and supports most of what would be the normal `std` for `unix` systems. Though the work can't be directly pushed over as `libc` requires a new dependency (a platform specific library with OS functions called `libctru`), so work on that must be finalized before any work on the standard. It's not my main objective to get the standard working (especially for the big requirement in maintenance that it would take, and I'm currently alone in this). Maybe after getting `libc` sorted out I might try to implement std support in the most thin way possible, to ensure it lasting in the future. It would be great for supporting non-platform-specific crates, but it's not a necessity.

skade:
> > Is `std` planned to be supported in any way? Just curious.
> 
> The Nintendo 3DS has had multiple out-of-tree projects porting parts or all of the standard library. The 3DS is a system not very powerful, but with many features, and supports most of what would be the normal `std` for `unix` systems. Though the work can't be directly pushed over as `libc` requires a new dependency (a platform specific library with OS functions called `libctru`), so work on that must be finalized before any work on the standard. It's not my main objective to get the standard working (especially for the big requirement in maintenance that it would take, and I'm currently alone in this). Maybe after getting `libc` sorted out I might try to implement std support in the most thin way possible, to ensure it lasting in the future. It would be great for supporting non-platform-specific crates, but it's not a necessity.

@nagisa  @Meziu Just a quick note: if anyone starts work on any bindings to library functions - some companies consider the interfaces to any of their libraries a trade secret and under NDA (even if they are standard interfaces!) and console vendors are a particular case there.

This is a grey area we definitely should not move in. Please note that even if we use one of the open SDKs, this grey area still exists, because the interpretation is up to the vendor. For SDKs, thats just an area they are deliberately in, but for the `libc` crate, there's the problem that all platforms share one space, so also the set of problems. If there's a move to add further target support to libc, as annoying as it is, please get some legal review. Just to be clear: this is not to annoy y'all, but to make sure there's no frustration and annoyance in the future.

Mark-Simulacrum:
It'd be nice to put a note to that effect into https://github.com/rust-lang/rust/tree/master/src/doc/rustc/src/platform-support for this target, I think.

I'm a little surprised to not see a doc created there with at least the listing of target maintainers - @nagisa I think it would be good to add those when we add targets, even if they start out mostly empty.

Meziu:
> > > Is `std` planned to be supported in any way? Just curious.
> > 
> > 
> > The Nintendo 3DS has had multiple out-of-tree projects porting parts or all of the standard library. The 3DS is a system not very powerful, but with many features, and supports most of what would be the normal `std` for `unix` systems. Though the work can't be directly pushed over as `libc` requires a new dependency (a platform specific library with OS functions called `libctru`), so work on that must be finalized before any work on the standard. It's not my main objective to get the standard working (especially for the big requirement in maintenance that it would take, and I'm currently alone in this). Maybe after getting `libc` sorted out I might try to implement std support in the most thin way possible, to ensure it lasting in the future. It would be great for supporting non-platform-specific crates, but it's not a necessity.
> 
> @nagisa @Meziu Just a quick note: if anyone starts work on any bindings to library functions - some companies consider the interfaces to any of their libraries a trade secret and under NDA (even if they are standard interfaces!) and console vendors are a particular case there.
> 
> This is a grey area we definitely should not move in. Please note that even if we use one of the open SDKs, this grey area still exists, because the interpretation is up to the vendor. For SDKs, thats just an area they are deliberately in, but for the `libc` crate, there's the problem that all platforms share one space, so also the set of problems. If there's a move to add further target support to libc, as annoying as it is, please get some legal review. Just to be clear: this is not to annoy y'all, but to make sure there's no frustration and annoyance in the future.

I understand exactly what you mean. I’ve never heard of any problems regarding the use of open SDKs (especially for what is now an out-of-production console) but it’s true that doesn’t exclude problems in the future, especially when dealing with the Rust project as a whole. I will think about what actions to take when implementing libc, and will consult legal advice before doing so.

Issue #56935, opened 2018-12-18
creader: Host crate loaded twice produces different `CrateNum`s if host != target


petrochenkov:
Reproduction (as an UI test):
```rust
// force-host
// no-prefer-dynamic

#![crate_type = "proc-macro"]

extern crate proc_macro;
use proc_macro::TokenStream;

#[proc_macro]
pub fn mac(input: TokenStream) -> TokenStream {
    input
}
```
```rust
// edition:2018
// compile-flags:--extern reproduction
// aux-build:reproduction.rs

reproduction::mac!();

fn main() {}
```
```sh
# Assuming x86_64-unknown-linux-gnu host
./x.py test --bless --stage 1 --target i686-unknown-linux-gnu src/test/ui --test-args reproduction
```

Name resolution will try to load the crate twice, first in initial pass, then in validation pass.
The results will be different due to different loaded `CrateNum`s, so we get the "inconsistent resolution for a macro" assertion.

This happens due to the `locate_ctxt.triple == &self.sess.opts.target_triple` condition in `CrateLoader::load` preventing reuse of `CrateNum`s for host crates during cross-compilation.

cc @alexcrichton probably

petrochenkov:
If I'm not mistaken, the issue will affect any idiomatic Rust 2018 code trying to use proc macros together with cross-compilation.

petrochenkov:
The workaround is to avoid using host crates through extern prelude and use `extern crate` items instead, like it's done in https://github.com/rust-lang/rust/pull/56647/commits/a5d4544f54dfbd682e8f49f1a58c1205a25ce654.

alexcrichton:
Sorry for the late delay in checking in on this. This is somewhat expected in the sense that we load proc-macro crates from the compiler's target and the actual target differently as they may be two different crates. That was done a really long time ago and I never really followed all of the macro developments and how they affected cross compilation.

We can probably add a function that compares two CrateNum instances though to see if they only differ by target and maybe that'd help here?

jonas-schievink:
Posting a typical message and backtrace here to make this issue easier to find:

```
thread 'rustc' panicked at 'src/librustc_resolve/imports.rs:906: inconsistent resolution for an import', /rustc/1836e3b42a5b2f37fd79104eedbe8f48a5afdee6/src/libstd/macros.rs:13:23
stack backtrace:
   0: <std::sys_common::backtrace::_print::DisplayBacktrace as core::fmt::Display>::fmt
[...]

PR #88538, opened 2021-08-31
CrateLocator refactorings

bjorn3:
This makes the `CrateLocator` a lot cleaner IMHO and much more self-contained. The last commit removes `extra_filename` from the crate metadata. This is an **insta-stable** change as it allows a crate like `libfoo-abc.rlib` to be used as dependency and then be renamed as `libfoo-bcd.rlib` while still being found as indirect dependency. This may reduce performance when there are a lot of versions of the same crate available as the extra filename won't be used to do an early rejection of crates before trying to load metadata, but it makes the logic to find the right filename a lot cleaner.

### DIFF
--- a/compiler/rustc_metadata/src/locator.rs
+++ b/compiler/rustc_metadata/src/locator.rs
@@ -240,7 +240,8 @@ use tracing::{debug, info, warn};
 #[derive(Clone)]
 crate struct CrateLocator<'a> {
     // Immutable per-session configuration.
-    sess: &'a Session,
+    only_needs_metadata: bool,
+    sysroot: &'a Path,
     metadata_loader: &'a dyn MetadataLoader,
 
     // Immutable per-search configuration.
@@ -305,8 +306,18 @@ impl<'a> CrateLocator<'a> {
         root: Option<&'a CratePaths>,
         is_proc_macro: Option<bool>,
     ) -> CrateLocator<'a> {
+        // The all loop is because `--crate-type=rlib --crate-type=rlib` is
+        // legal and produces both inside this type.
+        let is_rlib = sess.crate_types().iter().all(|c| *c == CrateType::Rlib);
+        let needs_object_code = sess.opts.output_types.should_codegen();
+        // If we're producing an rlib, then we don't need object code.
+        // Or, if we're not producing object code, then we don't need it either
+        // (e.g., if we're a cdylib but emitting just metadata).
+        let only_needs_metadata = is_rlib || !needs_object_code;
+
         CrateLocator {
-            sess,
+            only_needs_metadata,
+            sysroot: &sess.sysroot,
             metadata_loader,
             crate_name,
             exact_paths: if hash.is_none() {
@@ -484,14 +495,7 @@ impl<'a> CrateLocator<'a> {
             return true;
         }
 
-        // The all loop is because `--crate-type=rlib --crate-type=rlib` is
-        // legal and produces both inside this type.
-        let is_rlib = self.sess.crate_types().iter().all(|c| *c == CrateType::Rlib);
-        let needs_object_code = self.sess.opts.output_types.should_codegen();
-        // If we're producing an rlib, then we don't need object code.
-        // Or, if we're not producing object code, then we don't need it either
-        // (e.g., if we're a cdylib but emitting just metadata).
-        if is_rlib || !needs_object_code {
+        if self.only_needs_metadata {
             flavor == CrateFlavor::Rmeta
         } else {
             // we need all flavors (perhaps not true, but what we do for now)
@@ -591,7 +595,7 @@ impl<'a> CrateLocator<'a> {
             // candidates are all canonicalized, so we canonicalize the sysroot
             // as well.
             if let Some((prev, _)) = &ret {
-                let sysroot = &self.sess.sysroot;
+                let sysroot = self.sysroot;
                 let sysroot = sysroot.canonicalize().unwrap_or_else(|_| sysroot.to_path_buf());
                 if prev.starts_with(&sysroot) {
                     continue;
--- a/compiler/rustc_metadata/src/locator.rs
+++ b/compiler/rustc_metadata/src/locator.rs
@@ -248,7 +248,6 @@ crate struct CrateLocator<'a> {
     crate_name: Symbol,
     exact_paths: Vec<CanonicalizedPath>,
     pub hash: Option<Svh>,
-    pub host_hash: Option<Svh>,
     extra_filename: Option<&'a str>,
     pub target: &'a Target,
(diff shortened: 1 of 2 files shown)


bjorn3:
I'm guessing that this failure was caused by the extra filename changes.

petrochenkov:
I'm going to review because "continue refactoring crate locator" is an item from my work queue on which I was going to work myself (cc https://github.com/rust-lang/rust/issues/56935#issuecomment-653935780).

bjorn3:
I forgot an ! in the `only_needs_metadata` calculation.

bjorn3:
I think I will revert "Don't pass Session to CrateError::report". Unlike with `CrateLocator` I don't think there are much benefits to CrateError::report not taking a Session after all.

petrochenkov:
>This may reduce performance when there are a lot of versions of the same crate available as the extra filename won't be used to do an early rejection of crates before trying to load metadata, but it makes the logic to find the right filename a lot cleaner.

As far as I understand, the crates will now be (usually) rejected via hash kept in metadata instead of the extra filename piece (which is also some kind of hash if cargo is used).

Besides many versions of the same crate the perf regression can hit many crates with the same prefix like `librustc-abcd.rlib`/`librustc_foo-efgh.rlib`/`librustc_bar-ijkl.rlib`/etc, and the search for the library `rustc` will now need to open all the `librustc*` crate files and read hashes from them, because we no longer rely on knowing `-abcd`.

bjorn3:
> As far as I understand, the crates will now be (usually) rejected via hash kept in metadata instead of the extra filename piece (which is also some kind of hash if cargo is used).

That and the file prefix depending in the crate name.

> and the search for the library rustc will now need to open all the librustc* crate files and read hashes from them, because we no longer rely on knowing -abcd.

Didn't think about that. Would have been nice if the extra filename was required to start with `-` to clearly separate the crate name and extra filenane. It is now a breaking change to change this unfortunately even though cargo always does this.

compiler/rustc_metadata/src/locator.rs
```diff
@@ -242,6 +243,7 @@ crate struct CrateLocator<'a> {
     // Immutable per-session configuration.
     sess: &'a Session,
     metadata_loader: &'a dyn MetadataLoader,
+    current_crate: Option<&'a str>,
```
petrochenkov, line 246: `current_crate` isn't used by `CrateLocator`, only by `CrateError::report`, so I'd pass it to that function instead (together with the rest of the `Session` as you wrote in https://github.com/rust-lang/rust/pull/88538#issuecomment-909456333).

bjorn3:
I think the regression is neglectable.

> and the search for the library rustc will now need to open all the librustc* crate files and read hashes from them, because we no longer rely on knowing -abcd.

I think a perf regression for this specific case can be prevented by checking with `-` as extra prefix first before checking without extra prefix. Assuming that the extra filename starts with a `-` most of the time this would prevent most of the unnecessary lookup. It would still do the unnecessary lookup if the extra filenames don't start with a `-`. Speaking of which I actually don't think this commit changes behavior. If the file is not found with the specified extra filename as extra prefix, it will already try again without extra prefix, which matches all crates with the right crate name independent of the extra filename.

compiler/rustc_metadata/src/locator.rs
```diff
@@ -256,6 +255,7 @@ crate struct CrateLocator<'a> {
     pub triple: TargetTriple,
     pub filesearch: FileSearch<'a>,
     root: Option<&'a CratePaths>,
+    only_needs_metadata: bool,
```
petrochenkov, line 255: Looks like the previous placement was actually right, `only_needs_metadata` is only based on the session and not on the currently searched crate. All other fields in the "per-search configuration" apply to the specific crate that we are searching right now and can be different for other searches.

petrochenkov:
>Speaking of which I actually don't think this commit changes behavior. If the file is not found with the specified extra filename as extra prefix, it will already try again without extra prefix, which matches all crates with the right crate name independent of the extra filename.

It can change behavior in theory if the file *is* found with the specified extra prefix.
With the old logic we stop searching, with the new logic we continue searching and can potentially report "multiple candidates" errors. This is not a very realistic case because for the error to be reported hashes must match, but the extra search can take some time.

petrochenkov:
I'm still not sure about no longer keeping the `-C extra_filename` hints, they are cheap to keep and can potentially speed up crate search.
I need a second opinion, from cargo people maybe. cc @ehuss @alexcrichton

bjorn3:
> This is not a very realistic case because for the error to be reported hashes must match, but the extra search can take some time.

Doesn't the search stop as soon a crate with the right SVH was found if a required SVH was specified?

petrochenkov:
@bjorn3 
>Doesn't the search stop as soon a crate with the right SVH was found if a required SVH was specified?

It does not? I've just looked through the code, but haven't noticed any such early return logic.

compiler/rustc_metadata/src/locator.rs
bjorn3: Reverted this commit.

compiler/rustc_metadata/src/locator.rs
bjorn3: Reverted this commit except the `PathBuf`->`Path` change.

bjorn3:
I will squash the reverts after review.

alexcrichton:
I don't know the specifics of what this PR is about, and I can't find a necessarily clear question that's being asked, but I can try to offer some thoughts anyway. From [the comment above](https://github.com/rust-lang/rust/pull/88538#issuecomment-909665421) it sounds like previously rustc is taking `-Cextra-filename` into account when searching for dependencies, first looking at rlibs with a matching `-Cextra-filename` and succeeding quickly if found. If that's the case and it's now being removed, I would caution against anything that would cause more candidate rlibs to be searched. Cargo dumps everything into `deps` for dependencies, and with rebuilds across features and Rust versions it's not uncommon for development directories to have hundreds of copies of `libfoo-xxxxx.rlib`. If all compilations have to search through all those rlibs to reject them that means that incremental compiles are going to be that much slower. This sort of test case is unlikely to really be exercised on perf.r-l.o since it's always working with pristine directories and it's relatively rare that multiple `libfoo-xxxx.rlib` files will appear in one crate graph.

bjorn3:
I reverted the extra-filename commit. While I think it would be nice to drop extra-filename, the downsides don't weight up against it. At least without some kind of cache to map from svh to filename, but that introduces much more complexity elsewhere.

petrochenkov:
r=me with commits appropriately squashed.

bjorn3:
Squashed all reverts and the bugfix commit.

@bors r=petrochenkov

Issue #71874, opened 2020-05-04
swap_unchecked ?


leonardo-m:
A slice method similar to slice::get_unchecked_mut that could be added:

```
#[inline]
pub fn swap_unchecked(&mut self, a: usize, b: usize) {
    unsafe {
        // Can't take two mutable loans from one vector, so instead
        // just cast them to their raw pointers to do the swap.
        let pa: *mut T = this.get_unchecked_mut(a);
        let pb: *mut T = this.get_unchecked_mut(b);
        ptr::swap(pa, pb);
    }
}
```
To be added to the std lib only if there's enough usage for this method.

See also:
https://crates.io/crates/unchecked-index

bluss:
It is relatively easy to implement this outside of libcore (in your own code). The preferred way (current best practice) would be to go through `self.as_mut_ptr()` and offsetting from there, instead of calling `get_unchecked_mut` (which uses an exclusive borrow of the slice; no longer exclusive if we keep an active pointer around).

leonardo-m:
Thank you for the comments bluss, second version then:
```
#[inline]
pub unsafe fn swap_unchecked(&mut self, i: usize, j: usize) {
    let ptr1 = self.as_mut_ptr().add(i);
    let ptr2 = self.as_mut_ptr().add(j);
    ptr::swap(ptr1, ptr2);
}
```

bluss:
I'd propose this for anyone who wants to copy this implementation (this is what was meant with only using one `&mut [T]`).

```rust
// Method on [T]
pub unsafe fn swap_unchecked(&mut self, i: usize, j: usize) {
    debug_assert!(i < self.len());
    debug_assert!(j < self.len());
    let data_ptr = self.as_mut_ptr();
    std::ptr::swap(data_ptr.add(i), data_ptr.add(j));
}
```

RalfJung:
@bluss maybe add a `debug-assert!(i != j)`?

bluss:
For ptr::swap, equality is ok

RalfJung:
Ah you are right.

leonardo-m:
See also:
https://doc.rust-lang.org/nightly/src/core/slice/mod.rs.html#679

PR #88540, opened 2021-08-31
add `slice::swap_unchecked`

ibraheemdev:
An unsafe version of `slice::swap` that does not do bounds checking.

### DIFF
--- a/library/core/src/slice/mod.rs
+++ b/library/core/src/slice/mod.rs
@@ -560,8 +560,8 @@ impl<T> [T] {
     #[stable(feature = "rust1", since = "1.0.0")]
     #[inline]
     pub fn swap(&mut self, a: usize, b: usize) {
-        assert_in_bounds(self.len(), a);
-        assert_in_bounds(self.len(), b);
+        let _ = &self[a];
+        let _ = &self[b];
 
         // SAFETY: we just checked that both `a` and `b` are in bounds
         unsafe { self.swap_unchecked(a, b) }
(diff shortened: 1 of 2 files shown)


leonardo-m:
See #71874

leonardo-m:
I think it's worth using this inside slice::reverse.

library/core/src/slice/mod.rs
```diff
@@ -561,15 +561,47 @@ impl<T> [T] {
...
     pub fn swap(&mut self, a: usize, b: usize) {
-        // Can't take two mutable loans from one vector, so instead use raw pointers.
-        let pa = ptr::addr_of_mut!(self[a]);
-        let pb = ptr::addr_of_mut!(self[b]);
-        // SAFETY: `pa` and `pb` have been created from safe mutable references and refer
-        // to elements in the slice and therefore are guaranteed to be valid and aligned.
-        // Note that accessing the elements behind `a` and `b` is checked and will
-        // panic when out of bounds.
+        assert!(a < self.len());
+        assert!(b < self.len());
```
kennytm, line 565: the panic message degraded.

previously the out-of-bound panic would say like

```
index out of bounds: the len is 3 but the index is 4
```

but now it will just say

```
assertion failed: a < self.len()
```

klensy: And tests didn't catch that, need test )

kennytm:
maybe revert the implementation of the safe `swap`, keeping `swap` and `swap_unchecked` separate? at least we don't need to do bounds checking 4 times in debug build.

Issue #88440, opened 2021-08-28
2021 disjoint capture suggestion breaks macro_rules macro


ehuss:
I tried this code:

```rust
#![allow(unused)]
#![warn(rust_2021_incompatible_closure_captures)]

macro_rules! foo {
    ($v:ident) => {
        move || {
            println!("{:?}", $v.0);
        }
    };
}

fn main() {
    let x = (vec![1], vec![2]);
    foo!(x);
}
```

The rust_2021_incompatible_closure_captures lint generates a suggestion which causes this to fail to compile:

```rust
macro_rules! foo {
    ($v:ident) => {
        move || {
            let _ = &x;  // ERROR: cannot find value `x` in this scope
            println!("{:?}", $v.0);
        }
    };
}
```

I'm not exactly sure what this should do.  In this particular case, the suggestion is not required.  However, that's a fundamental problem with the `rust_2021_incompatible_closure_captures` lint, as it does not know when it is required.

In general, modifying macros can be prone to failure, so maybe an alternative is to issue a warning with a non-machine-applicable fix when inside a `macro_rules` definition?

This was found in the 2021 crater run for rcodec 1.0.1: https://crater-reports.s3.amazonaws.com/pr-87190-3/try%23a7a572ce3edd6d476191fbfe92c9c1986e009b34/reg/rcodec-1.0.1/log.txt
Note that this has even worse behavior than the example above.  The sample macro:

```rust
macro_rules! forcomp_stmts {
    { $yld:block, $v:ident, $e:expr } => {
        $e.map(move |$v| $yld)
    };
    { $yld:block, $v:ident, $e:expr, $($tv:ident, $te:expr),+ } => {
        $e.and_then(move |$v| {
            forcomp_stmts!($yld, $($tv, $te),+)
        })
    };
}
```

rewrites the map call to:

```rust
        $e.map{ let _ = &decoded; (move |$v| $yld) }
```

which is invalid syntax. 

### Meta

`rustc --version --verbose`:
```
rustc 1.56.0-nightly (ac50a5335 2021-08-27)
binary: rustc
commit-hash: ac50a53359328a5d7f2f558833e63d59d372e4f7
commit-date: 2021-08-27
host: x86_64-apple-darwin
release: 1.56.0-nightly
LLVM version: 13.0.0
```

cc @rust-lang/wg-rfc-2229

arora-aman:
Some of my thought process:


[...]

PR #88543, opened 2021-08-31
Improve closure dummy capture suggestion in macros.

m-ou-se:
Fixes some cases of https://github.com/rust-lang/rust/issues/88440

Fixes https://crater-reports.s3.amazonaws.com/pr-87190-3/try%23a7a572ce3edd6d476191fbfe92c9c1986e009b34/reg/rcodec-1.0.1/log.txt

### DIFF
--- a/compiler/rustc_typeck/src/check/upvar.rs
+++ b/compiler/rustc_typeck/src/check/upvar.rs
@@ -47,7 +47,7 @@ use rustc_middle::ty::{
 };
 use rustc_session::lint;
 use rustc_span::sym;
-use rustc_span::{BytePos, MultiSpan, Pos, Span, Symbol, DUMMY_SP};
+use rustc_span::{BytePos, MultiSpan, Pos, Span, Symbol};
 use rustc_trait_selection::infer::InferCtxtExt;
 
 use rustc_data_structures::stable_map::FxHashMap;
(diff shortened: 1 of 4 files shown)


src/test/ui/closures/2229_closure_analysis/migrations/closure-body-macro-fragment.stderr
```diff
@@ -0,0 +1,37 @@
...
+   |
+LL |           let f = || $body;
+   |  _________________^
+LL | |
+LL | |         f();
+LL | |     }};
+   | |     - in Rust 2018, `a` is dropped here, but in Rust 2021, only `a.0` will be dropped here as part of the closure
+LL | |     ($body:block) => {{
+LL | |         m!(@ $body);
+   | |__________________^
```
estebank, line 12: That's... an unfortunate span.

src/test/ui/closures/2229_closure_analysis/migrations/closure-body-macro-fragment.stderr
```diff
@@ -0,0 +1,37 @@
...
+  --> $DIR/closure-body-macro-fragment.rs:4:9
+   |
+LL | #![warn(rust_2021_compatibility)]
+   |         ^^^^^^^^^^^^^^^^^^^^^^^
+   = note: `#[warn(rust_2021_incompatible_closure_captures)]` implied by `#[warn(rust_2021_compatibility)]`
+   = note: for more information, see <https://doc.rust-lang.org/nightly/edition-guide/rust-2021/disjoint-capture-in-closures.html>
+   = note: this warning originates in the macro `m` (in Nightly builds, run with -Z macro-backtrace for more info)
+help: add a dummy let to cause `a` to be fully captured
+   |
+LL ~     m!({
```
estebank, line 32: I'm intrigued at why this line is shown to have a change. The correct fix might be changing the suggestion printing machinery to verify that before and after are actually different.

estebank (approved):

compiler/rustc_typeck/src/check/upvar.rs
```diff
@@ -680,15 +680,32 @@ impl<'a, 'tcx> FnCtxt<'a, 'tcx> {
...
+                        // invocation, i.e. the body is not contained inside the
+                        // closure span, then we walk up the expansion until we
+                        // find the span before the expansion.
+                        let s = self.tcx.hir().span(body_id.hir_id);
+                        s.find_ancestor_inside(closure_span).unwrap_or(s)
+                    };
+
+                    if let Ok(mut s) = self.tcx.sess.source_map().span_to_snippet(closure_body_span) {
+                        if s.starts_with('$') {
+                            // Looks like a macro fragment. Try to find the real block.
```
nikomatsakis, line 694: Can/should we walk up the backtrace in the expansion instead of this? I guess this works, I'm thinking to think if there is some kind of "catch".

src/test/ui/closures/2229_closure_analysis/migrations/closure-body-macro-fragment.stderr
m-ou-se: Yeah, some weird things are  going on when mixing `:block` and `:expr` fragments. And closures just use their first and last token to create their full span, which isn't always great. Something to improve another time. ^^

compiler/rustc_typeck/src/check/upvar.rs
nikomatsakis: OK, I see.

src/test/ui/closures/2229_closure_analysis/migrations/closure-body-macro-fragment.stderr
m-ou-se: It got a `\n` appended to it. I'm using the span at the end of this line to add a `"\n     let _ = .."` suggestion. I guess the `\n` is counted as part of this line.

src/test/ui/closures/2229_closure_analysis/migrations/closure-body-macro-fragment.stderr
estebank: Oh aboslutely, mixed spans involving macros have long standing issues :)

PR #88546, opened 2021-08-31
Emit proper errors when on missing closure braces

### LINKED ISSUE #88065: Attempting to debug in a iterator's closure without {} gives unhelpful error
https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=9b9dfd706291c45a1f6a567324b6c860

Given the following code: 

```rust
fn main() {
    let num = 5;
    (1..num).reduce(|a, b| 
        println!("{}", a); // <== Here
        a * b
    ).unwrap();
}
```

The current output is:

```
error: expected one of `)`, `,`, `.`, `?`, or an operator, found `;`
 --> src/main.rs:4:26
  |
4 |         println!("{}", a); // <== Here
  |                          ^
  |                          |
  |                          expected one of `)`, `,`, `.`, `?`, or an operator
  |                          help: missing `,`

error[E0425]: cannot find value `a` in this scope
 --> src/main.rs:5:9
  |
5 |         a * b
  |         ^ not found in this scope

error[E0425]: cannot find value `b` in this scope
 --> src/main.rs:5:13
  |
5 |         a * b
  |             ^ not found in this scope

error[E0061]: this function takes 1 argument but 2 arguments were supplied
    --> src/main.rs:3:14
     |
3    |       (1..num).reduce(|a, b| 
     |  ______________^^^^^^_-
     | |              |
     | |              expected 1 argument
4    | |         println!("{}", a); // <== Here
     | |_________________________-
5    |           a * b
     |           ----- supplied 2 arguments
     |
note: associated function defined here

error: aborting due to 4 previous errors

Some errors have detailed explanations: E0061, E0425.
For more information about an error, try `rustc --explain E0061`.

```

```
3    |       (1..num).reduce(|a, b| {
     |  ____________________________^
     | | Multi-line closure requires scope brackets
```

I was trying to debug an iterator and added a `println!` call. To my surprise, it caused the compiler to fail and emit MANY warnings. None of them pointed at the real problem which is that I was missing curly brackets `{}`.

scrabsha:
This commit focuses on emitting clean errors for the following syntax
error:

```
Some(42).map(|a|
    dbg!(a);
    a
);
```

Previous implementation tried to recover after parsing the closure body
(the `dbg` expression) by replacing the next `;` with a `,`, which made
the next expression belong to the next function argument. As such, the
following errors were emitted (among others):
  - the semicolon token was not expected,
  - a is not in scope,
  - Option::map is supposed to take one argument, not two.

This commit allows us to gracefully handle this situation by adding
giving the parser the ability to remember when it has just parsed a
closure body inside a function call. When this happens, we can treat the
unexpected `;` specifically and try to parse as much statements as
possible in order to eat the whole block. When we can't parse statements
anymore, we generate a clean error indicating that the braces are
missing, and return an ExprKind::Err.

Closes #88065.

r? @estebank

### DIFF
--- a/compiler/rustc_parse/src/parser/expr.rs
+++ b/compiler/rustc_parse/src/parser/expr.rs
@@ -1,9 +1,12 @@
 use super::pat::{RecoverColon, RecoverComma, PARAM_EXPECTED};
 use super::ty::{AllowPlus, RecoverQPath, RecoverReturnSign};
-use super::{AttrWrapper, BlockMode, ForceCollect, Parser, PathStyle, Restrictions, TokenType};
+use super::{
+    AttrWrapper, BlockMode, ClosureSpans, ForceCollect, Parser, PathStyle, Restrictions, TokenType,
+};
 use super::{SemiColonMode, SeqSep, TokenExpectType, TrailingToken};
 use crate::maybe_recover_from_interpolated_ty_qpath;
 
+use ast::token::DelimToken;
 use rustc_ast::ptr::P;
 use rustc_ast::token::{self, Token, TokenKind};
 use rustc_ast::tokenstream::Spacing;
@@ -91,6 +94,8 @@ impl<'a> Parser<'a> {
     /// Parses an expression.
     #[inline]
     pub fn parse_expr(&mut self) -> PResult<'a, P<Expr>> {
+        self.current_closure.take();
+
         self.parse_expr_res(Restrictions::empty(), None)
     }
 
@@ -1736,7 +1741,7 @@ impl<'a> Parser<'a> {
         let capture_clause = self.parse_capture_clause()?;
         let decl = self.parse_fn_block_decl()?;
         let decl_hi = self.prev_token.span;
-        let body = match decl.output {
+        let mut body = match decl.output {
             FnRetTy::Default(_) => {
                 let restrictions = self.restrictions - Restrictions::STMT_EXPR;
                 self.parse_expr_res(restrictions, None)?
@@ -1753,11 +1758,28 @@ impl<'a> Parser<'a> {
             self.sess.gated_spans.gate(sym::async_closure, span);
         }
 
-        Ok(self.mk_expr(
+        if self.token.kind == TokenKind::Semi && self.token_cursor.frame.delim == DelimToken::Paren
+        {
+            // It is likely that the closure body is a block but where the
+            // braces have been removed. We will recover and eat the next
+            // statements later in the parsing process.
+            body = self.mk_expr_err(body.span);
+        }
+
+        let body_span = body.span;
+
+        let closure = self.mk_expr(
             lo.to(body.span),
             ExprKind::Closure(capture_clause, asyncness, movability, decl, body, lo.to(decl_hi)),
             attrs,
-        ))
+        );
+
+        // Disable recovery for closure body
+        let spans =
+            ClosureSpans { whole_closure: closure.span, closing_pipe: decl_hi, body: body_span };
+        self.current_closure = Some(spans);
+
+        Ok(closure)
     }
 
     /// Parses an optional `move` prefix to a closure-like construct.
--- a/compiler/rustc_parse/src/parser/mod.rs
+++ b/compiler/rustc_parse/src/parser/mod.rs
@@ -142,6 +142,17 @@ pub struct Parser<'a> {
     /// If present, this `Parser` is not parsing Rust code but rather a macro call.
     subparser_name: Option<&'static str>,
     capture_state: CaptureState,
+    /// This allows us to recover when the user forget to add braces around
+    /// multiple statements in the closure body.
+    pub current_closure: Option<ClosureSpans>,
+}
+
+/// Stores span informations about a closure.
+#[derive(Clone)]
+pub struct ClosureSpans {
+    pub whole_closure: Span,
+    pub closing_pipe: Span,
+    pub body: Span,
 }
 
 /// Indicates a range of tokens that should be replaced by
@@ -440,6 +451,7 @@ impl<'a> Parser<'a> {
                 replace_ranges: Vec::new(),
                 inner_attr_ranges: Default::default(),
             },
+            current_closure: None,
         };
 
         // Make parser point to the first token.
@@ -761,8 +773,11 @@ impl<'a> Parser<'a> {
                     first = false;
                 } else {
                     match self.expect(t) {
-                        Ok(false) => {}
+                        Ok(false) => {
+                            self.current_closure.take();
+                        }
                         Ok(true) => {
+                            self.current_closure.take();
                             recovered = true;
                             break;
                         }
@@ -770,10 +785,29 @@ impl<'a> Parser<'a> {
                             let sp = self.prev_token.span.shrink_to_hi();
                             let token_str = pprust::token_kind_to_string(t);
 
-                            // Attempt to keep parsing if it was a similar separator.
-                            if let Some(ref tokens) = t.similar_tokens() {
-                                if tokens.contains(&self.token.kind) && !unclosed_delims {
-                                    self.bump();
+                            match self.current_closure.take() {
+                                Some(closure_spans) if self.token.kind == TokenKind::Semi => {
+                                    // Finding a semicolon instead of a comma
+                                    // after a closure body indicates that the
+                                    // closure body may be a block but the user
+                                    // forgot to put braces around its
+                                    // statements.
+
+                                    self.recover_missing_braces_around_closure_body(
+                                        closure_spans,
+                                        expect_err,
+                                    )?;
+
+                                    continue;
+                                }
+
+                                _ => {
+                                    // Attempt to keep parsing if it was a similar separator.
+                                    if let Some(ref tokens) = t.similar_tokens() {
+                                        if tokens.contains(&self.token.kind) && !unclosed_delims {
+                                            self.bump();
+                                        }
+                                    }
                                 }
                             }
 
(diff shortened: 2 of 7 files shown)


compiler/rustc_parse/src/parser/mod.rs
```diff
@@ -839,6 +868,48 @@ impl<'a> Parser<'a> {
...
+            initial_semicolon,
+            "This `;` turns the expression into a statement, which must be placed in a block",
+        );
+
+        while self.eat(&TokenKind::Semi) {
+            let _ = self.parse_stmt(ForceCollect::Yes)?;
+        }
+
+        expect_err.set_primary_message(
+            "closure body that contain statements must be surrounded by braces",
```
inquisitivecrystal, line 888: There's a grammar problem here: you either want "**a** closure body that contain**s** statements must be surrounded by braces" or ""closure bod**ies** that contain statements must be surrounded by braces". We may have a stylistic convention on which one is correct, but if so, I don't know what it is.

compiler/rustc_parse/src/parser/mod.rs
```diff
@@ -839,6 +868,48 @@ impl<'a> Parser<'a> {
...
+    fn recover_missing_braces_around_closure_body(
+        &mut self,
+        right_pipe_span: Span,
+        mut expect_err: DiagnosticBuilder<'_>,
+    ) -> PResult<'a, ()> {
+        let initial_semicolon = self.token.span;
+
+        expect_err.span_help(
+            initial_semicolon,
+            "This `;` turns the expression into a statement, which must be placed in a block",
```
inquisitivecrystal, line 880: Somehow reading this message in the test output below doesn't get the point across for me. I'm not entirely sure why. I think it might be that "This `;`" is clear, but then I think, "what expression? what statement?". The problem might be that those concepts seem too abstract and detached from the actual code to be understood at a glance. I'm not exactly sure how to fix it though.

compiler/rustc_parse/src/parser/mod.rs
scrabsha: That makes sense, thanks. The [suggestion style guide](https://rustc-dev-guide.rust-lang.org/diagnostics.html?#suggestion-style-guide) doesn't help unfortunately. I have found no other wording suggestion in the documentation.

I like your second suggestion and I'll keep that one for no particular reason other than being grammatically correct. Feel free to suggest anything that may be better :)

compiler/rustc_parse/src/parser/mod.rs
scrabsha: I understand your pain. It took me a lot of time to write this sentence and I'm still not satisfied with it.

To mitigate your "what expression? what statement?" problem, I see two solutions:
  - reword the message as "This `;` turns the **preceding** expression into a statement [...]" (see first screenshot),
  - span the help message on the whole "expression+semicolon" chunk, changing the message to "the trailing `;` turns the expression into a statement" (see second screenshot).

Both are not ideal, though.

Overall, I'm not even sure if it is a good idea to expect from the user to know the difference between an expression and a statement, given that users may not have knowledge in language grammar.

compiler/rustc_parse/src/parser/mod.rs
```diff
@@ -761,20 +766,44 @@ impl<'a> Parser<'a> {
...
-                                if tokens.contains(&self.token.kind) && !unclosed_delims {
-                                    self.bump();
+                            match self.last_closure_body.take() {
+                                None => {
+                                    // Attempt to keep parsing if it was a similar separator.
+                                    if let Some(ref tokens) = t.similar_tokens() {
+                                        if tokens.contains(&self.token.kind) && !unclosed_delims {
+                                            self.bump();
+                                        }
+                                    }
```
estebank, line 788: Shouldn't this be in the `_ => {` arm?

compiler/rustc_parse/src/parser/mod.rs
estebank: An option we have is to abuse the `MultiSpan` API:

```rust
let mut sp: MultiSpan = vec![initial_semicolon].into();
sp.push_label(initial_semicolon, "this `;` turns the preceding expression into a statement");
sp.push_label(expr.span, "this expression is a statement because of the trailing semicolon");
sp.push_label(closure.span, "this is the parsed closure...");
sp.push_label(end_brace_span, "...but likely you meant the closure to end here");
```

compiler/rustc_parse/src/parser/expr.rs
```diff
@@ -1733,6 +1736,18 @@ impl<'a> Parser<'a> {
...
+        // Disable recovery for closure body
+        self.last_closure_body = Some(decl_hi);
+
+        if self.token.kind == TokenKind::Semi && self.token_cursor.frame.delim == DelimToken::Paren
+        {
+            // It is likely that the closure body is a block but where the
+            // braces have been removed. We will recover and eat the next
+            // statements later in the parsing process.
+
+            return Ok(self.mk_expr_err(lo.to(body.span)));
```
estebank, line 1748: What we might want to do is make a closure here where it's body is an error expression placeholder, although that doesn't change anything in this PR, I would like us to be able to mark that closure somehow (either by changing it's variant or just using the err expression as a marker) to detect cases where `{|x| let y = x + 1; y}` is written. But we can make that change after the fact.

Could we add a test for the "ruby closure" case? I'm intrigued at what it looks like under _this_ PR.

compiler/rustc_parse/src/parser/mod.rs
scrabsha: Yes it should!

compiler/rustc_parse/src/parser/expr.rs
scrabsha: Oh that makes sense. I had not noticed the difference between "a closure whose body is an expr error" and "an expr error" in my early development. Thanks

compiler/rustc_parse/src/parser/mod.rs
scrabsha: While this definitely works, I think this adds a lot of noise in the `note` section:
A screenshot of the error message with all the spanned labels suggested by estebank. The result is very cumbersome. We don't know what to read first.

I think the best solution here is to remove the `spanned_label` whose message is "this expression is a statement because of the trailing semicolon", but this needs more experiments.

compiler/rustc_parse/src/parser/expr.rs
scrabsha: @estebank I added the ruby-style closure syntax ui tests you talked about. In order to make tests pass, I had to annotate them with the (arguably not helpful) current compiler output. Please ping me if this is not was you expected :)

compiler/rustc_parse/src/parser/mod.rs
estebank: Another alternative is to add a new note for either the statement or the closure.

It could even make sense to put the "this is the parsed closure..."/"...but likely you meant the closure to end here" in its own note, which would make the output more vertically prominent, but easier to read.

src/test/ui/expr/malformed_closure/ruby_style_closure.rs
```diff
@@ -0,0 +1,8 @@
+fn main() {
```
estebank, line 1: Let's add a comment with the ticket number and mentioning that this isn't yet handled by the compiler, but we want to at some point.

src/test/ui/expr/malformed_closure/ruby_style_closure.stderr
```diff
@@ -0,0 +1,18 @@
...
+  --> $DIR/ruby_style_closure.rs:5:14
+   |
+LL |         Some(x * 2)
+   |              ^ not found in this scope
+
+error[E0277]: expected a `FnOnce<({integer},)>` closure, found `Option<_>`
+  --> $DIR/ruby_style_closure.rs:2:22
+   |
+LL |     let p = Some(45).and_then({
+   |                      ^^^^^^^^ expected an `FnOnce<({integer},)>` closure, found `Option<_>`
```
estebank, line 11: This should be pointing at the `Some(x * 2)`. (Not something to address in this PR)

compiler/rustc_parse/src/parser/mod.rs
scrabsha: > It could even make sense to put the "this is the parsed closure..."/"...but likely you meant the closure to end here" in its own note, which would make the output more vertically prominent, but easier to read.

I implemented it and IMO this is what gives the best results! It makes the error message longer, but it is way easier to read, as each subdiagnostic relate a specific concept.

compiler/rustc_parse/src/parser/mod.rs
```diff
@@ -145,7 +146,11 @@ pub struct Parser<'a> {
 
     /// This allows us to recover when the user forget to add braces around
     /// multiple statements in the closure body.
-    pub last_closure_body: Option<Span /* The closing `|` of the closure declarator. */>,
+    pub last_closure_body: Option<(
+        Span, /* The whole body. */
+        Span, /* The closing `|` of the closure declarator. */
+        Span, /* What we parsed as closure body. */
+    )>,
```
estebank, line 152: We should create a new struct for these :)

compiler/rustc_parse/src/parser/expr.rs
```diff
@@ -1747,11 +1744,18 @@ impl<'a> Parser<'a> {
             body = self.mk_expr_err(body.span);
         }
 
-        Ok(self.mk_expr(
+        let body_span = body.span;
+
+        let clo = self.mk_expr(
```
estebank, line 1749: Can we rename this to either `closure` or `expr`? `clo` isn't used anywhere else and might lead to short lived confusion, but confusion nonetheless.

compiler/rustc_parse/src/parser/mod.rs
```diff
@@ -145,7 +146,11 @@ pub struct Parser<'a> {
 
     /// This allows us to recover when the user forget to add braces around
     /// multiple statements in the closure body.
-    pub last_closure_body: Option<Span /* The closing `|` of the closure declarator. */>,
+    pub last_closure_body: Option<(
```
estebank, line 148: Given that it's no longer just the body, we should rename this to `last_closure` or `current_closure`.

estebank (approved):
r=me after addressing the comments.

src/test/ui/expr/malformed_closure/missing_braces_around_block.stderr
```diff
@@ -0,0 +1,38 @@
...
+LL |     ).unwrap();
+   |     ^
+   |
+note: statement found outside of a block
+  --> $DIR/missing_braces_around_block.rs:16:26
+   |
+LL |         println!("{}", a);
+   |         -----------------^ this `;` turns the preceding expression into a statement
+   |         |
+   |         this expression is a statement because of the trailing semicolon
```
estebank, line 16: Rereading with fresh eyes, I think that this is slightly misleading: the `;` turns the previous _closure_ into a statement, as far as the parser is concerned, but I feel it might be a reasonable "white lie" to not confuse people _too_ much.

compiler/rustc_parse/src/parser/expr.rs
```diff
@@ -1733,11 +1738,28 @@ impl<'a> Parser<'a> {
...
         }
 
-        Ok(self.mk_expr(
+        if self.token.kind == TokenKind::Semi && self.token_cursor.frame.delim == DelimToken::Paren
+        {
+            // It is likely that the closure body is a block but where the
+            // braces have been removed. We will recover and eat the next
+            // statements later in the parsing process.
+            body = self.mk_expr_err(body.span);
+        }
```
estebank, line 1747: I'm somewhat uneasy with this check, but having both `Semi` and `Paren` be true should only happen on malformed code.

src/test/ui/expr/malformed_closure/ruby_style_closure.stderr
estebank: Done in https://github.com/rust-lang/rust/pull/88719

scrabsha:
Thanks for the approval and sorry for the noise I created. I'll do better next time :)

PR #88552, opened 2021-09-01
Stop allocating vtable entries for non-object-safe methods

nbdd0121:
Current a vtable entry is allocated for all associated fns, even if the method is not object-safe: https://godbolt.org/z/h7vx6f35T

As a result, each vtable for `Iterator`' currently consumes 74 `usize`s. This PR stops allocating vtable entries for those methods, reducing vtable size of each `Iterator` vtable to 7 `usize`s.

Note that this PR introduces will cause more invocations of `is_vtable_safe_method`. So a perf run might be needed. If result isn't favorable then we might need to query-ify `is_vtable_safe_method`.

### DIFF
--- a/compiler/rustc_trait_selection/src/traits/mod.rs
+++ b/compiler/rustc_trait_selection/src/traits/mod.rs
@@ -625,6 +625,31 @@ fn dump_vtable_entries<'tcx>(
     tcx.sess.struct_span_err(sp, &msg).emit();
 }
 
+fn own_existential_vtable_entries<'tcx>(
+    tcx: TyCtxt<'tcx>,
+    trait_ref: ty::PolyExistentialTraitRef<'tcx>,
+) -> &'tcx [DefId] {
+    let trait_methods = tcx
+        .associated_items(trait_ref.def_id())
+        .in_definition_order()
+        .filter(|item| item.kind == ty::AssocKind::Fn);
+    // Now list each method's DefId (for within its trait).
+    let own_entries = trait_methods.filter_map(move |trait_method| {
+        debug!("own_existential_vtable_entry: trait_method={:?}", trait_method);
+        let def_id = trait_method.def_id;
+
+        // Some methods cannot be called on an object; skip those.
+        if !is_vtable_safe_method(tcx, trait_ref.def_id(), &trait_method) {
+            debug!("own_existential_vtable_entry: not vtable safe");
+            return None;
+        }
+
+        Some(def_id)
+    });
+
+    tcx.arena.alloc_from_iter(own_entries.into_iter())
+}
+
 /// Given a trait `trait_ref`, iterates the vtable entries
 /// that come from `trait_ref`, including its supertraits.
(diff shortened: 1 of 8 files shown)


nbdd0121:
Introduced a query. Please do another perf run

compiler/rustc_trait_selection/src/traits/mod.rs
```diff
@@ -641,21 +666,15 @@ fn vtable_entries<'tcx>(
...
-                // Now list each method's DefId and InternalSubsts (for within its trait).
-                // If the method can never be called from this object, produce `Vacant`.
-                let own_entries = trait_methods.map(move |trait_method| {
-                    debug!("vtable_entries: trait_method={:?}", trait_method);
-                    let def_id = trait_method.def_id;
-
-                    // Some methods cannot be called on an object; skip those.
-                    if !is_vtable_safe_method(tcx, trait_ref.def_id(), &trait_method) {
-                        debug!("vtable_entries: not vtable safe");
-                        return VtblEntry::Vacant;
```
nagisa, line 657: Is there any reason why we cannot remove the vtable entries for the "impossible predicates" situation below and remove the `Vacant` variant entirely?

Doesn't need to be considered for this PR, of course, just a passing thought.

nagisa:
I would like to see additional tests for this, especially good would be a test for `Iterator` (to verify the claims made in the PR description) as well as a test for a vtable where we still have `Vacant` entries of some sort due to the `impossible_predicates` situation.

compiler/rustc_trait_selection/src/traits/mod.rs
nbdd0121:
```rust
trait X {
    fn foo(&self)
    where
        Self: Send,
    {
    }
}

impl X for () {
    fn foo(&self) {}
}

impl X for *mut () {}

pub fn main() {
    <dyn X + Send as X>::foo(&());
}
```

We need to generate `X::foo` for `()`, but we cannot generate `X::foo` for `*mut ()`.

nagisa:
@bors r+

Thanks!

camelid:
@bors r-

CI is failing, looks like it just needs a quick `x.py test --bless`.

nbdd0121:
Done (I manually edited the file so I don't have to recompile everything)

nbdd0121:
Race condition :)

@bors r=nagisa

camelid:
@bors r-

CI is failing again; please wait for CI to pass before re-approving the PR (otherwise it slows down the bors queue) :)

camelid:
Heh, another race condition :)

Mark-Simulacrum:
This had [mixed](https://perf.rust-lang.org/compare.html?calcNewSig=true&start=e2750baf53aaa60db95f10759f6cf9463dc5a6bd&end=e30b68353fe22b00f40d021e7914eeb78473b3c1) results on landing, generally, with no significant cycle count results. Marking as triaged.

Issue #88276, opened 2021-08-24
Unary plus gives suboptimal parse error


CAD97:
Given the [following code](https://play.rust-lang.org/?version=nightly&mode=debug&edition=2018&gist=9d47f8301f8c0e7bbb81a74492c230b0): 

```rust
fn main() {
    let x = +1;
}
```

The current output is:

```
error: expected expression, found `+`
 --> src/main.rs:2:13
  |
2 |     let x = +1;
  |             ^ expected expression
```

Ideally, if we want to continue rejecting unary plus on number literals, the parser should recover the unexpected unary plus, and the output should look somewhat like:

```
error: leading `+` on numeric literals is not supported
 --> src/main.rs:2:13
  |
2 |     let x = +1;
  |             ^ unexpected `+`
  = help: remove it; the unsigned literal is positive and may still be used for a signed type
```

Tested on 1.56.0-nightly (2021-08-22 af140757b4cb1a60d107)

theo-lw:
Working on this @rustbot claim.

camelid:
Closing now that #88553 has been merged.

PR #88553, opened 2021-09-01
Improve diagnostics for unary plus operators (#88276)

theo-lw:
This pull request improves the diagnostics emitted on parsing a unary plus operator. See #88276.

Before:

```
 --> src/main.rs:2:13
  |
2 |     let x = +1;
  |             ^ expected expression
```

After:

```
error: leading `+` is not supported
 --> main.rs:2:13
  |
2 |     let x = +1;
  |             ^
  |             |
  |             unexpected `+`
  |             help: try removing the `+`
```

### DIFF
--- a/compiler/rustc_parse/src/parser/expr.rs
+++ b/compiler/rustc_parse/src/parser/expr.rs
@@ -23,6 +23,8 @@ use rustc_span::symbol::{kw, sym, Ident, Symbol};
 use rustc_span::{BytePos, Pos};
 use std::mem;
 
+use tracing::debug;
+
 /// Possibly accepts an `token::Interpolated` expression (a pre-parsed expression
 /// dropped into the token stream, which happens while parsing the result of
 /// macro expansion). Placement of these is not as complex as I feared it would
@@ -355,6 +357,7 @@ impl<'a> Parser<'a> {
     /// but the next token implies this should be parsed as an expression.
     /// For example: `if let Some(x) = x { x } else { 0 } / 2`.
     fn error_found_expr_would_be_stmt(&self, lhs: &Expr) {
+        debug!("error_found_expr_would_be_stmt(lhs: {:?})", lhs);
         let mut err = self.struct_span_err(
             self.token.span,
             &format!("expected expression, found `{}`", pprust::token_to_string(&self.token),),
@@ -516,6 +519,15 @@ impl<'a> Parser<'a> {
             token::BinOp(token::And) | token::AndAnd => {
                 make_it!(this, attrs, |this, _| this.parse_borrow_expr(lo))
             }
+            token::BinOp(token::Plus) => {
+                this.struct_span_err(lo, "leading `+` is not supported")
+                    .span_label(lo, "unexpected `+`")
+                    .span_suggestion_short(lo, "remove the `+`", "".to_string(), Applicability::MachineApplicable)
+                    .emit();
+                this.bump();
+
+                this.parse_prefix_expr(None)
+            } // `+expr`
             token::Ident(..) if this.token.is_keyword(kw::Box) => {
                 make_it!(this, attrs, |this, _| this.parse_box_expr(lo))
             }
--- a/compiler/rustc_parse/src/parser/expr.rs
+++ b/compiler/rustc_parse/src/parser/expr.rs
@@ -23,8 +23,6 @@ use rustc_span::symbol::{kw, sym, Ident, Symbol};
 use rustc_span::{BytePos, Pos};
 use std::mem;
 
-use tracing::debug;
-
 /// Possibly accepts an `token::Interpolated` expression (a pre-parsed expression
 /// dropped into the token stream, which happens while parsing the result of
 /// macro expansion). Placement of these is not as complex as I feared it would
@@ -165,12 +163,9 @@ impl<'a> Parser<'a> {
             if [token::DotDot, token::DotDotDot, token::DotDotEq].contains(&self.token.kind) {
                 return self.parse_prefix_range_expr(attrs);
             } else {
-                let result = self.parse_prefix_expr(attrs);
-                debug!("parse_prefix_expr result: {:?}", &result);
-                result?
+                self.parse_prefix_expr(attrs)?
             }
         };
-        debug!("parse_assoc_expr_with(lhs = {:?})", &lhs);
         let last_type_ascription_set = self.last_type_ascription.is_some();
 
         if !self.should_continue_as_assoc_expr(&lhs) {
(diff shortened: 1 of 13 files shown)


compiler/rustc_parse/src/parser/expr.rs
```diff
@@ -516,6 +516,27 @@ impl<'a> Parser<'a> {
             token::BinOp(token::And) | token::AndAnd => {
                 make_it!(this, attrs, |this, _| this.parse_borrow_expr(lo))
             }
+            token::BinOp(token::Plus) => {
```
estebank, line 519: This should only be done if the following token is a numeric literal. As this stands now, if you write `+foo` or `+""` this error will trigger as well. The suggestion won't be *wrong per-se*, but in those cases the problems are deeper than the `+`.

You can use `Parser::look_ahead` which takes a closure giving you a token for whatever `N` tokens you are advancing, so you can do something like `token::BinOp(token::Plus) if this.look_ahead(1, |t| t.is_numeric_literal() => {` or something. We need to decide whether an identifier like `+foo` is something where we want to also give this error or not. Personally I think we might not want to.

compiler/rustc_parse/src/parser/expr.rs
```diff
@@ -516,6 +516,27 @@ impl<'a> Parser<'a> {
...
+                let sp = this.sess.source_map().start_point(lo);
+                if let Some(sp) = this.sess.ambiguous_block_expr_parse.borrow().get(&sp) {
+                    this.sess.expr_parentheses_needed(&mut err, *sp);
+                } else {
+                    err.span_suggestion(
+                        lo,
+                        "try removing the `+`",
+                        "".to_string(),
+                        Applicability::MachineApplicable,
+                    );
```
estebank, line 533: Let's use `span_suggestion_verbose`, so it appears on its own subdiagnostic (the new patch-style makes removal a bit easier to see, IMO), but this is a minor thing. Alternatively we could also use `tool_only_span_suggestion` here, because the removal of the `+` is kind of obvious, and all we might want is the tooltip in an editor, the label in the output doesn't _add_ that much.

compiler/rustc_parse/src/parser/expr.rs
```diff
@@ -516,6 +516,27 @@ impl<'a> Parser<'a> {
             token::BinOp(token::And) | token::AndAnd => {
                 make_it!(this, attrs, |this, _| this.parse_borrow_expr(lo))
             }
+            token::BinOp(token::Plus) => {
+                let mut err = this.struct_span_err(lo, "leading `+` is not supported");
+                err.span_label(lo, "unexpected `+`");
```
estebank, line 521: We might want to replace this with

```suggestion
                let mut err = this.unexpected().unwrap_err();
```

compiler/rustc_parse/src/parser/expr.rs
```diff
@@ -516,6 +516,27 @@ impl<'a> Parser<'a> {
...
+
+                // a block on the LHS might have been intended to be an expression instead
+                let sp = this.sess.source_map().start_point(lo);
```
estebank, line 524: Isn't `sp == lo` always here, if the token is `+`?

compiler/rustc_parse/src/parser/expr.rs
estebank: If we change the creation of `err` to be a call to `unexpected()`, then we can check here for the `look_ahead` and see if it is a literal or a binding identifier, and if so emit the suggestion and `this.bump()`.

I don't know if we can also check if the next element would be a parenthesized expression (checking for `OpenDelim(Paren)`?) and have some confidence that there won't be other parse errors.

src/test/ui/parser/issue-88276-unary-plus.rs
```diff
@@ -0,0 +1,13 @@
+// run-rustfix
+#[allow(unused_parens)]
+fn main() {
+    let _ = +1; //~ ERROR leading `+` is not supported
+    let _ = -+(1+2)*3; //~ ERROR leading `+` is not supported
+    let _ = -+-+(1+2)*3; //~ ERROR leading `+` is not supported
+                         //~| ERROR leading `+` is not supported
```
estebank, line 7: This is an interesting case, because with the suggestions applied, this looks like C's prefix decrement `--foo`, which we don't have. We should have a lint against it 😅

src/test/ui/parser/issue-88276-unary-plus.rs
```diff
@@ -0,0 +1,13 @@
...
+    let _ = (1 + +2) * +3; //~ ERROR leading `+` is not supported
+                           //~| ERROR leading `+` is not supported
+    let _ = (+&"hello"); //~ ERROR leading `+` is not supported
```
estebank, line 10: So this is one case where removing the `+` might be obvious when written like this, but I don't know about the general case.

estebank (commented):
I love these kind of parser fixes!

Would you mind addressing the comments?

compiler/rustc_parse/src/parser/expr.rs
theo-lw: I'm getting some unexpected results after changing this to unexpected():

```
error: expected `#`, found `+`
 --> src/test/ui/parser/issue-88276-unary-plus.rs:4:13
  |
4 |     let _ = +1; //~ ERROR leading `+` is not supported
  |             ^ expected `#`
  |
```

compiler/rustc_parse/src/parser/expr.rs
theo-lw: Good point, I think you are right.

compiler/rustc_parse/src/parser/expr.rs
theo-lw: I think it makes sense to limit the check to just numeric literals, made changes to reflect that.

src/test/ui/parser/issue-88276-unary-plus.rs
theo-lw: That's a good idea for a lint!

compiler/rustc_parse/src/parser/expr.rs
theo-lw: > Let's use `span_suggestion_verbose`, so it appears on its own subdiagnostic (the new patch-style makes removal a bit easier to see, IMO), but this is a minor thing. Alternatively we could also use `tool_only_span_suggestion` here, because the removal of the `+` is kind of obvious, and all we might want is the tooltip in an editor, the label in the output doesn't _add_ that much.

`span_suggestion_verbose` makes the fix more obvious, changed.

> I don't know if we can also check if the next element would be a parenthesized expression (checking for `OpenDelim(Paren)`?) and have some confidence that there won't be other parse errors.

I think there could be other parse errors if the next token is a parenthesis. For instance, the following expression has a parse error between the parentheses: `+(+x)`.

theo-lw:
Addressed your comments @estebank, thanks for reviewing this!

src/test/ui/parser/issue-88276-unary-plus.rs
estebank: Filed https://github.com/rust-lang/rust/issues/88669 so we don't forget.

PR #88558, opened 2021-09-01
Const drop

### LINKED ISSUE #88424: New problem with const_trait_impl
This code:

```rust
#![feature(const_trait_impl, const_fn_trait_bound, const_panic)]

pub trait Foo {
    fn bar(self) -> usize;
    const MAX: Self;
}
impl const Foo for usize {
    fn bar(self) -> usize { self }
    const MAX: Self = Self::MAX;
}

const fn spam<T: Foo>(n: usize) {
    assert!(n <= T::MAX.bar());
}

fn main() {}
```

Using:

```
rustc 1.56.0-nightly (ac50a5335 2021-08-27)
binary: rustc
commit-hash: ac50a53359328a5d7f2f558833e63d59d372e4f7
commit-date: 2021-08-27
host: x86_64-pc-windows-gnu
release: 1.56.0-nightly
LLVM version: 13.0.0
```

Gives:

```rust
error[E0015]: calls in constant functions are limited to constant functions, tuple structs and tuple variants
  --> ...\test.rs:13:18
   |
13 |     assert!(n <= T::MAX.bar());
   |                  ^^^^^^^^^^^^
```

I think this used to compile up to the precedent Nightly.

fee1-dead:
The changes are pretty primitive at this point. But at least it works. ^-^

Problems with the current change that I can think of now:
 - [x] `~const Drop` shouldn't change anything in the non-const world.
 - [x] types that do not have drop glues shouldn't fail to satisfy `~const Drop` in const contexts. `struct S { a: u8, b: u16 }` This might not fail for `needs_non_const_drop`, but it will fail in `rustc_trait_selection`.
 - [x] The current change accepts types that have `const Drop` impls but have non-const `Drop` glue.

Fixes #88424.

Significant Changes:

- `~const Drop` is no longer treated as a normal trait bound. In non-const contexts, this bound has no effect, but in const contexts, this restricts the input type and all of its transitive fields to either a) have a `const Drop` impl or b) can be trivially dropped (i.e. no drop glue)
- `T: ~const Drop` will not be linted like `T: Drop`.
- Instead of recursing and iterating through the type in `rustc_mir::transform::check_consts`, we use the trait system to special case `~const Drop`. See [`rustc_trait_selection::...::candidate_assembly#assemble_const_drop_candidates`](https://github.com/fee1-dead/rust/blob/const-drop/compiler/rustc_trait_selection/src/traits/select/candidate_assembly.rs#L817) and others.

Changes not related to `const Drop`ping and/or changes that are insignificant:

 - `Node.constness_for_typeck` no longer returns `hir::Constness::Const` for type aliases in traits. This was previously used to hack how we determine default bound constness for items. But because we now use an explicit opt-in, it is no longer needed.
 - Removed `is_const_impl_raw` query. We have `impl_constness`, and the only existing use of that query uses `HirId`, which means we can just operate it with hir.
 - `ty::Destructor` now has a field `constness`, which represents the constness of the destructor.

r? @oli-obk


oli-obk:
can we desugar ~const Drop to an auto-trait bound for ConstDrop which we then properly define for everything that needs it, but never expose it to users?

src/test/ui/rfc-2632-const-trait-impl/const-drop.rs
```diff
@@ -0,0 +1,23 @@
+#![feature(const_trait_impl)]
+#![feature(const_fn_trait_bound)]
+#![feature(const_mut_refs)]
+#![feature(const_panic)]
+
+struct S;
+
+impl const Drop for S {
+    fn drop(&mut self) {
+        // NB: There is no way to tell that a const destructor is ran,
```
oli-obk, line 10: you can give S a field that is a mutable borrow to something and then mutate that on drop

fee1-dead: Yeah that is true. I originally thought about having mutable references to constant items which won't work. But for local mutable variables in const fn works fine.

oli-obk:
> In non-const contexts, this bound has no effect

This could be problematic.

```rust
struct NotDrop(String) // has drop glue, but no Drop impl

trait MyTrait {}
impl<T: Drop> MyTrait for T {}
impl MyTrait for NotDrop {}
```

works on stable and can be very useful for some pin related code (source: https://github.com/rust-lang/rfcs/pull/2632#issuecomment-528461395).

fee1-dead:
nonono. I did not make `T: Drop` unusable. I only special cased `T: ~const Drop` in non-const contexts. See [this](https://github.com/rust-lang/rust/blob/28ef0d6e9eccb97f3dd9469df0211c6eeb61949a/compiler/rustc_trait_selection/src/traits/select/candidate_assembly.rs#L286).

fee1-dead:
In fact I see this as a justification for something that is opt-in. We have `T: Drop` which means "has explicit drop impl; might have drop glue". We then define `T: (const-if-const) Drop` to be "has const drop glue or can be trivially dropped". If we go the other way (i.e. the implicit and opt-out syntax), then we might need to introduce something like `ConstDrop` trait and that would have to be another new lang item as well.

compiler/rustc_const_eval/src/transform/check_consts/qualifs.rs
```diff
@@ -97,22 +101,48 @@ impl Qualif for HasMutInterior {
...
     }
 
     fn in_any_value_of_ty(cx: &ConstCx<'_, 'tcx>, ty: Ty<'tcx>) -> bool {
-        ty.needs_drop(cx.tcx, cx.param_env)
+        let drop_trait = if let Some(did) = cx.tcx.lang_items().drop_trait() {
+            did
+        } else {
+            // there is no way to define a type that needs non-const drop
+            // without having the lang item present.
+            return false;
```
oli-obk, line 120: Beautiful

src/test/ui/rfc-2632-const-trait-impl/const-drop-fail.rs
```diff
@@ -0,0 +1,54 @@
...
+
+impl Drop for NonTrivialDrop {
+    fn drop(&mut self) {
+        println!("Non trivial drop");
+    }
+}
+
+struct ConstImplWithDropGlue(NonTrivialDrop);
+
+impl const Drop for ConstImplWithDropGlue {
```
oli-obk, line 19: Ideally we'd reject this impl, but this may be something we can leave to clippy, not sure, it's a footgun to be sure.

compiler/rustc_trait_selection/src/traits/select/mod.rs
```diff
@@ -1038,19 +1038,15 @@ impl<'cx, 'tcx> SelectionContext<'cx, 'tcx> {
         it.for_each(|o| o.recursion_depth = cmp::max(min_depth, o.recursion_depth) + 1);
     }
 
-    /// Checks that the recursion limit has not been exceeded.
-    ///
-    /// The weird return type of this function allows it to be used with the `try` (`?`)
-    /// operator within certain functions.
-    fn check_recursion_limit<T: Display + TypeFoldable<'tcx>, V: Display + TypeFoldable<'tcx>>(
+    fn check_recursion_depth<T: Display + TypeFoldable<'tcx>>(
```
oli-obk, line 1041: infcx is very sensitive to changes in hot code paths, maybe this change caused it? Can you do a PR without this change (duplicating for your use case) and see how it goes? Or maybe just inline(always) check_recursion_limit?

fee1-dead: Let me open a pr adding inline(always) and see how that goes. I don't think anything other than this code caused the regression

fee1-dead: Opened #88962.

fee1-dead:
Okay, I looked through the perf results, and there seems to be a regression for `mir_promoted` and/or `mir_const_qualif`. Perhaps the way to fix this is to add a fast path for `NeedsNonConstDrop::in_any_value_of_ty`.

compiler/rustc_const_eval/src/transform/check_consts/qualifs.rs
```diff
@@ -97,22 +101,48 @@ impl Qualif for HasMutInterior {
...
+impl Qualif for NeedsNonConstDrop {
+    const ANALYSIS_NAME: &'static str = "flow_needs_nonconst_drop";
     const IS_CLEARED_ON_MOVE: bool = true;
 
     fn in_qualifs(qualifs: &ConstQualifs) -> bool {
         qualifs.needs_drop
     }
 
     fn in_any_value_of_ty(cx: &ConstCx<'_, 'tcx>, ty: Ty<'tcx>) -> bool {
-        ty.needs_drop(cx.tcx, cx.param_env)
```
oli-obk, line 111: maybe keep this for a fast reject?

oli-obk:
regression fixed by #88965 and #88963

PR #88560, opened 2021-09-01
`fmt::Formatter::pad`: don't call chars().count() more than one time

klensy:
First commit merges two branches of match to call chars().count() only once: that should be faster if this method hits place of 3rd (previous) branch, plus quarter shorter.
Second commit fixes some clippy lints while i'm here (should it be separate PR?).


m-ou-se:
Thanks!

@bors r+

PR #30490, opened 2015-12-20
libstd: unix process spawning: fix bug with setting stdio

ipetkov:
- If the requested descriptors to inherit are stdio descriptors there
  are situations where they will not be set correctly
- Example: parent's stdout --> child's stderr
         parent's stderr --> child's stdout
- Solution: if the requested descriptors for the child are stdio
  descriptors, `dup` them before overwriting the child's stdio

Example of a program which exhibits the bug:

``` rust
// stdio.rs
use std::io::Write;
use std::io::{stdout, stderr};
use std::process::{Command, Stdio};
use std::os::unix::io::FromRawFd;

fn main() {
    stdout().write_all("parent stdout\n".as_bytes()).unwrap();
    stderr().write_all("parent stderr\n".as_bytes()).unwrap();

    Command::new("sh")
        .arg("-c")
        .arg("echo 'child stdout'; echo 'child stderr' 1>&2")
        .stdin(Stdio::inherit())
        .stdout(unsafe { FromRawFd::from_raw_fd(2) })
        .stderr(unsafe { FromRawFd::from_raw_fd(1) })
        .status()
        .unwrap_or_else(|e| { panic!("failed to execute process: {}", e) });
}
```

Before:

```
$ rustc --version
rustc 1.7.0-nightly (8ad12c3e2 2015-12-19)
$ rustc stdio.rs && ./stdio >out 2>err
$ cat out
parent stdout
$ cat err
parent stderr
child stdout
child stderr
```

After (expected): 

```
$ rustc --version
rustc 1.7.0-dev (712eccee2 2015-12-19)
$ rustc stdio.rs && ./stdio >out 2>err
$ cat out
parent stdout
child stderr
$ cat err
parent stderr
child stdout
```

PR #88561, opened 2021-09-01
Allow redirecting subprocess stdout to our stderr etc.

### LINKED ISSUE #79731: Command cannot replumb stdout to stderr
When running a `Command`, it is often desirable to plumb stdout to the same place as stderr, as if `2>&1` in a shell rune.

Unfortunately, there is no way to do this with `Command` right now without writing `unsafe`, non-portable code.

Really we want to fix this so that you can pass to `Command` anything that implements `AsRawFd`, so probably the right thing would be `std::Process::Stdio::reuse<F:AsRawFd>(fd: &f)`.  The implementation is not entirely trivial - we need an fd-permuting algorithm that avoids making unnecessary calls to `dup2`.

On Windows that would be `std::Process::Stdio::reuse<F:AsRawHandle>(fd: &f)` I guess.

It's a shame that we have totally separate names for `*RawFd` and `*RawHandle` rather than a portable-opaque but nonportably-transparent `OsOpenFileReference` type and `AsOsOpenFileReference`, `IntoOsOpenFileReference`, `FromOsOpenFileReference` but that seems like rather a major change.

ijackson:
Currently there is no safe or portable way to redirect a `Command`s stdout to the invoking program's stderr, nor vice versa.

There a comment in the source that suggests that this was intended, and trying to handle the fd shuffling problem properly, but it can never be reached other than by crazy unsafe stunts.

Instead, provide appropriate `From` impls.

<strike>I have only tested the implementation for Unix because I know how to do that.</strike>  There is now a Windows implementation which seems to pass the CI.  This branch currently contains a DO NOT MERGE commit to enable the Windows CI build on this MR branch.  That should be removed after code review and before merge.

Being trait impls, these are insta-stable.  The question is just whether we want to support platforms that cannot do this, and if so, whether we want that to be a compile error or are happy for it to be detected by `Command` methods returning errors.

Closes #79731

### DIFF
--- a/library/std/src/process.rs
+++ b/library/std/src/process.rs
@@ -1492,32 +1492,6 @@ impl From<fs::File> for Stdio {
 }
 
 #[stable(feature = "stdio_from_stdio", since = "1.58.0")]
-impl From<io::Stdin> for Stdio {
-    /// Specify to inherit our stdin.
-    ///
-    /// # Examples
-    ///
-    /// ```rust,no_run
-    /// #![feature(exit_status_error)]
-    /// use std::io;
-    /// use std::process::Command;
-    ///
-    /// # fn test() -> Result<(), Box<dyn std::error::Error>> {
-    /// let output = Command::new("wc").arg("-l")
-    ///     .stdin(io::stdin())
-    ///     .output()?;
-    /// output.status.exit_ok()?;
-    /// let input_lines: u64 = std::str::from_utf8(&output.stdout)?.trim().parse()?;
-    /// eprintln!("our standard input had {} lines", input_lines);
-    /// # Ok(())
-    /// # }
-    /// ```
-    fn from(inherit: io::Stdin) -> Stdio {
-        Stdio::from_inner(inherit.into())
-    }
-}
-
-#[stable(feature = "stdio_from_stdio", since = "1.58.0")]
 impl From<io::Stdout> for Stdio {
     /// Redirect command stdout/stderr to our stdout
     ///
--- a/library/std/src/process.rs
+++ b/library/std/src/process.rs
@@ -1491,7 +1491,7 @@ impl From<fs::File> for Stdio {
     }
 }
 
-#[stable(feature = "stdio_from_stdio", since = "1.58.0")]
+#[stable(feature = "stdio_from_stdio", since = "CURRENT_RUSTC_VERSION")]
 impl From<io::Stdout> for Stdio {
     /// Redirect command stdout/stderr to our stdout
     ///
--- a/library/std/src/process.rs
+++ b/library/std/src/process.rs
@@ -1491,6 +1491,92 @@ impl From<fs::File> for Stdio {
     }
 }
 
+#[stable(feature = "stdio_from_stdio", since = "1.58.0")]
+impl From<io::Stdin> for Stdio {
+    /// Specify to inherit our stdin.
+    ///
+    /// # Examples
+    ///
+    /// ```rust,no_run
+    /// #![feature(exit_status_error)]
+    /// use std::io;
+    /// use std::process::Command;
+    ///
+    /// # fn test() -> Result<(), Box<dyn std::error::Error>> {
+    /// let output = Command::new("wc").arg("-l")
+    ///     .stdin(io::stdin())
+    ///     .output()?;
+    /// output.status.exit_ok()?;
+    /// let input_lines: u64 = std::str::from_utf8(&output.stdout)?.trim().parse()?;
+    /// eprintln!("our standard input had {} lines", input_lines);
+    /// # Ok(())
+    /// # }
+    /// ```
+    fn from(inherit: io::Stdin) -> Stdio {
+        Stdio::from_inner(inherit.into())
+    }
+}
+
+#[stable(feature = "stdio_from_stdio", since = "1.58.0")]
+impl From<io::Stdout> for Stdio {
+    /// Redirect command stdout/stderr to our stdout
+    ///
+    /// # Examples
+    ///
+    /// ```rust
+    /// #![feature(exit_status_error)]
+    /// use std::io;
+    /// use std::process::Command;
+    ///
+    /// # fn test() -> Result<(), Box<dyn std::error::Error>> {
+    /// let output = Command::new("whoami")
+    // "whoami" is a command which exists on both Unix and Winodows,
+    // and which succeeds, producing some stdout output but no stderr.
+    ///     .stdout(io::stdout())
+    ///     .output()?;
+    /// output.status.exit_ok()?;
+    /// assert!(output.stdout.is_empty());
+    /// # Ok(())
+    /// # }
+    /// #
+    /// # if cfg!(unix) {
+    /// #     test().unwrap();
+    /// # }
+    /// ```
+    fn from(inherit: io::Stdout) -> Stdio {
+        Stdio::from_inner(inherit.into())
+    }
+}
+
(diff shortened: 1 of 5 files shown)


library/std/src/process.rs
```diff
@@ -1400,6 +1400,90 @@ impl From<fs::File> for Stdio {
     }
 }
 
+#[stable(feature = "stdio_from", since = "1.20.0")]
```
camelid, line 1403: nit: I think this should get its own feature gate and version number, even though it's insta-stable, to avoid confusing users.

ijackson:
I have rebased onto current head, given it its own feature tag (and updated the version) and provided what I thiink is an implementation on Windows.  I don't have any Windows here so I may have to iterate through the CI, in which case sorry for the noise.

ijackson:
The failing test is that for #30490.  The test case, and the code there, involves passing a static `0`, `1` or `2` to a function which says it _**consumes ownership of the specified file descriptor_, which is clearly UB.

I think I am going to have to break this MR up or work around this somehow.  Nnnng.

camelid:
Just FYI: `fixup!` commits are ***not*** automatically squashed upon merge.

ijackson:
> Just FYI: `fixup!` commits are _**not**_ automatically squashed upon merge.

Indeed.  This MR needs more work and an FCP, so there's no danger of them ending up in tree before I've had a chance to tell my git to autosquash them.  But thanks for the note.

ijackson:
Well, here's a version with Windows and which sidesteps whether `FileDesc` is allowed to contain 0/1/2.  Let's see if the CI likes it.  I have **not tested it on Windows at all**.

ijackson:
Would someone be able to help with the Windows tests?  Ideally, by adapting my new doctests so they work on Windows.  (Breaking the Unix version while doing that would be fine; I can fix it up.)

Alternatively, if someone could suggest a command to run that will succeed and produce some nonempty stdlout output, and no stderr output, that will probably suffice and I can iterate it through the CI.

I am going to try setting the E-easy tag because I think this should be straightforward for someone familiar with Windows.

@rustbot modify labels +O-windows +E-help-wanted +E-easy

Mark-Simulacrum:
@ijackson Would you still be interested in driving this forward? My recommendation for Windows testing is to enable the relevant builder on PRs in src/ci/github-actions/ci.yml, and then run  ./x.py run src/tools/expand-yaml-anchors.

You can find the pr builder section by looking for `pr:` (should have the mingw-check, llvm-12, and tools builders), and adding the x86_64-msvc-1 builder to that. Hopefully that's enough that you can test this change successfully.

I'm going to assign myself as the reviewer for this PR, if you need further help please run `@rustbot ready` to get the labels updated so I have it in my queue to take a look.

@rustbot author

ijackson:
This has got to the top of my todo list again.  Sorry for the delay...

ijackson:
Gosh!  I wasn't expecting that to pass.  But I do see
```
2023-01-03T16:42:52.9693329Z [0m[0m[1m[32m   Doc-tests[0m std
2023-01-03T16:42:54.3561006Z 
2023-01-03T16:42:54.3561378Z running 1101 tests
...
2023-01-03T16:43:26.7132116Z test result: ok. 1085 passed; 0 failed; 16 ignored; 0 measured; 0 filtered out; finished in 32.36s
```
I am going to mark this ready now and request review even though I'm leaving the "DO NOT MERGE" Windows CI commit at the top.

Mark-Simulacrum:
r? libs-api for T-libs-api FCP

pitaj:
Ping from triage. Appears the merge conflicts are due to the DO NOT MERGE commit, so I'm ignoring that. I'm going to reroll for the FCP.

r? libs-api

library/std/src/process.rs
```diff
@@ -1431,6 +1431,92 @@ impl From<fs::File> for Stdio {
     }
 }
 
+#[stable(feature = "stdio_from_stdio", since = "1.58.0")]
+impl From<io::Stdin> for Stdio {
```
dtolnay, line 1435: Questions about this impl &mdash; could you clarify whether this has any use that is not identical to `.stdin(Stdio::inherit())`?

I tried the following, but it didn't work as I'd naively expect:

```rust
use std::io::{self, Read};
use std::process::Command;

fn main() {
    let mut child = Command::new("whoami")
        .stdout(io::stdin())
        .spawn()
        .unwrap();

    let mut stdin = String::new();
    io::stdin().read_to_string(&mut stdin).unwrap();

    let status = child.wait().unwrap();
    assert!(status.success());

    println!("{:?}", stdin);
}
```

(I am on Linux.)

Instead of the child process's stdout feeding into the current process's stdin, the child's stdout goes to stdout and the current process stdin is unaffected.

The example code you gave in the documentation for this impl uses `.stdin(io::stdin())`. We would prefer that this always be written as `.stdin(Stdio::inherit())` though, correct?

dtolnay:
The other 2 impls (to support `.stdout(io::stderr())` and `.stderr(io::stdout())`) LGTM and I would be happy to propose FCP for the libs-api team after we've discussed whether stdin should be included here.

ijackson:
> Instead of the child process's stdout feeding into the current process's stdin, the child's stdout goes to stdout and the current process stdin is unaffected.

I think this is a bug in my implementation.  It also demonstrates that my tests are not sufficient.  I'm investigating.

ijackson:
@rustbot labels modify -A-testsuite -T-infra

This is tripping because of the DO NOT MERGE commit I'm using to have the CI run the new doctests on Windows.

ijackson:
FTAOD, the actual MR branch here is not in a fit state to merge - it needs some squashing etc.

So, assuming the FCP completes with approval, don't merge the branch right away.  I will tidy it and force push, first.

library/std/src/process.rs
```diff
@@ -1491,6 +1491,66 @@ impl From<fs::File> for Stdio {
...
+    /// # Examples
+    ///
+    /// ```rust
+    /// #![feature(exit_status_error)]
+    /// use std::io;
+    /// use std::process::Command;
+    ///
+    /// # fn test() -> Result<(), Box<dyn std::error::Error>> {
+    /// let output = Command::new("whoami")
+    // "whoami" is a command which exists on both Unix and Winodows,
```
Lonami, line 1507:
```suggestion
    // "whoami" is a command which exists on both Unix and Windows,
```

library/std/src/process.rs
```diff
@@ -1491,6 +1491,66 @@ impl From<fs::File> for Stdio {
     }
 }
 
+#[stable(feature = "stdio_from_stdio", since = "1.58.0")]
```
teor2345, line 1494: I think this magic constant is new since the PR was opened:
https://rustc-dev-guide.rust-lang.org/stabilization_guide.html#updating-the-feature-gate-listing

```suggestion
#[stable(feature = "stdio_from_stdio", since = "CURRENT_RUSTC_VERSION")]
```

teor2345: (This applies to all instances of the compiler version.)

ijackson:
Thanks everyone for the approval, and the code review.  I will update this branch soon, so that we can merge it.

ijackson:
For convenience, I have made the squashed and tidied branch into a new MR: #114590

pitaj:
Opinions will obviously differ, but personally, what you've done feels less convenient.

dtolnay (approved):
Thank you! I have approved your tidied PR in https://github.com/rust-lang/rust/pull/114590#pullrequestreview-1616337899.

PR #88567, opened 2021-09-01
Remove redundant `Span` in `QueryJobInfo`

camelid:
Previously, `QueryJobInfo` was composed of two parts: a `QueryInfo` and
a `QueryJob`. However, both `QueryInfo` and `QueryJob` have a `span`
field, which seem to be the same. So, the `span` was recorded twice.

Now, `QueryJobInfo` is composed of a `QueryStackFrame` (the other field
of `QueryInfo`) and a `QueryJob`. So, now, the `span` is only recorded
once.

### DIFF
--- a/compiler/rustc_query_system/src/query/job.rs
+++ b/compiler/rustc_query_system/src/query/job.rs
@@ -61,7 +61,7 @@ where
     }
 
     fn query(self, map: &QueryMap<D>) -> QueryStackFrame {
-        map.get(&self).unwrap().info.query.clone()
+        map.get(&self).unwrap().query.clone()
     }
 
     #[cfg(parallel_compiler)]
@@ -81,7 +81,7 @@ where
 }
 
 pub struct QueryJobInfo<D> {
-    pub info: QueryInfo,
+    pub query: QueryStackFrame,
     pub job: QueryJob<D>,
 }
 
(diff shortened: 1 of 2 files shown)


camelid:
r? @cjgillot 

This may have a small improvement of compile-time memory usage, but it's likely to be quite small. Should I still mark this as rollup=never?

cjgillot:
The query infra is absurdly sensitive.
@bors try @rust-timer queue

camelid:
I'm not sure why it said there were no significant changes; instruction count is down by ~0.2% on several benchmarks; the only regression is 0.4% for `deep-vector-debug`, but IIRC that benchmark is somewhat noisy.

jyn514:
This greatly increased max-rss for tuple-stress-doc for some reason. Not sure if it's spurious or not.

cjgillot:
> I'm not sure why it said there were no significant changes; instruction count is down by ~0.2% on several benchmarks; the only significant regression (according to the perf.rlo webpage) is 0.4% for `deep-vector-debug`, but IIRC that benchmark is somewhat noisy.

<1% changes are not what I call significant :)

> This greatly increased max-rss for tuple-stress-doc for some reason. Not sure if it's spurious or not.

It's spurious, those code path are only used in incremental runs, and the regressed run is not incremental.

@bors r+ rollup

camelid:
> > I'm not sure why it said there were no significant changes; instruction count is down by ~0.2% on several benchmarks; the only significant regression (according to the perf.rlo webpage) is 0.4% for `deep-vector-debug`, but IIRC that benchmark is somewhat noisy.
> 
> <1% changes are not what I call significant :)

Well, of course it's not a *large* improvement, but the web interface shows those changes as "relevant" which I would think would be similar to "significant" :)

PR #88574, opened 2021-09-01
rustdoc: Box `GenericArg::Const` to reduce enum size

camelid:
This should reduce the amount of memory allocated in the common cases
where the `GenericArg` is a lifetime or type.

### DIFF
--- a/src/librustdoc/clean/mod.rs
+++ b/src/librustdoc/clean/mod.rs
@@ -1787,7 +1787,7 @@ impl Clean<GenericArgs> for hir::GenericArgs<'_> {
                         }
                         hir::GenericArg::Lifetime(_) => GenericArg::Lifetime(Lifetime::elided()),
                         hir::GenericArg::Type(ty) => GenericArg::Type(ty.clean(cx)),
-                        hir::GenericArg::Const(ct) => GenericArg::Const(ct.clean(cx)),
+                        hir::GenericArg::Const(ct) => GenericArg::Const(Box::new(ct.clean(cx))),
                         hir::GenericArg::Infer(_inf) => GenericArg::Infer,
                     })
                     .collect(),
--- a/src/librustdoc/clean/types.rs
+++ b/src/librustdoc/clean/types.rs
@@ -2012,9 +2012,14 @@ impl Path {
 crate enum GenericArg {
     Lifetime(Lifetime),
     Type(Type),
-    Const(Constant),
+    Const(Box<Constant>),
     Infer,
 }
+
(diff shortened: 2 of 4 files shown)


camelid:
(It's blocked because they change similar code and I think that could cause the results to be spurious.)

hellow554:
#88522 has been merged successfully and no rebase seems to be needed. So let's go?

camelid:
> #88522 has been merged successfully and no rebase seems to be needed. So let's go?

Yeah, I forgot that CI rebases before testing, but I'll still rebase just in case :)

camelid:
I'm pretty sure those regressions that rust-timer is reporting are spurious, because they are for non-doc builds, which shouldn't be affected at all by this change.

No significant changes in `*-doc` instruction count, but max-rss looks to be improved on virtually all `*-doc` benchmarks; the biggest improvements are for `deeply-nested-doc` at 1.2%, `clap-rs-doc` at about 1%, and `serde-doc` at about 1%. The only significant (according to the webpage) regression is 0.21% for `tokio-webpush-simple-doc`, which could be spurious.

Mark-Simulacrum:
@bors try @rust-timer queue

Let's see if they reproduce, perhaps? But seem likely to not be related to this PR.

GuillaumeGomez:
Should we add a `static_assert_size!` for `GenericArg` then?

camelid:
Seems pretty likely that these rustc regressions and rustc improvements are spurious, because they shouldn't be affected by this change and because in the first run, it reported

> Small regression in instruction counts (up to 0.5% on full builds of deep-vector)

but on this run, it reported:

> Small improvement in instruction counts (up to -0.6% on full builds of deep-vector)

Removing the perf-regression label then.

camelid:
`tuple-stress-doc` showed a 10% max-rss regression in the second perf run, but I'm pretty sure it's spurious because there were no significant changes in the first run.

camelid:
Aside from that, significant improvements ranged from 0.3% to 1%. serde-doc improved by the exact same amount both runs too (1.03%).

camelid:
> Should we add a `static_assert_size!` for `GenericArg` then?

Yeah, that's probably a good idea.

camelid:
> Should we add a `static_assert_size!` for `GenericArg` then?

Ok, done!

GuillaumeGomez (approved):

GuillaumeGomez:
Thanks!

@bors: r+ rollup

camelid:
@bors rollup-

This affects performance.

PR #88575, opened 2021-09-01
Querify `FnAbi::of_{fn_ptr,instance}` as `fn_abi_of_{fn_ptr,instance}`.

eddyb:
*Note: opening this PR as draft because it's based on #88499*

This more or less replicates the `LayoutOf::layout_of` setup from #88499, to replace `FnAbi::of_{fn_ptr,instance}` with `FnAbiOf::fn_abi_of_{fn_ptr,instance}`, and also route them through queries (which `layout_of` has used for a while).

The two changes at the use sites (other than the names) are:
* return type is now wrapped in `&'tcx`
  * the value *is* interned, which may affect performance
* the `extra_args` list is now an interned `&'tcx ty::List<Ty<'tcx>>`
  * should be cheap (it's empty for anything other than C variadics)
 
Theoretically, a `FnAbiOfHelpers` implementer could choose to keep the `Result<...>` instead of eagerly erroring, but the only existing users of these APIs are codegen backends, so they don't (want to) take advantage of this.
At least miri could make use of this, since it prefers propagating errors (it "just" doesn't use `FnAbi` yet - cc @RalfJung).

The way this is done is probably less efficient than what is possible, because the queries handle the correctness-oriented API (i.e. the split into `fn` pointers vs instances), whereas a lower-level query could end up with more reuse between different instances with identical signatures.

r? @nagisa cc @oli-obk @bjorn3

### DIFF
--- a/compiler/rustc_middle/src/query/mod.rs
+++ b/compiler/rustc_middle/src/query/mod.rs
@@ -1128,6 +1128,27 @@ rustc_queries! {
         desc { "computing layout of `{}`", key.value }
     }
 
+    /// Compute a `FnAbi` suitable for indirect calls, i.e. to `fn` pointers.
+    ///
+    /// NB: this doesn't handle virtual calls - those should use `fn_abi_of_instance`
+    /// instead, where the instance is an `InstanceDef::Virtual`.
+    query fn_abi_of_fn_ptr(
+        key: ty::ParamEnvAnd<'tcx, (ty::PolyFnSig<'tcx>, &'tcx ty::List<Ty<'tcx>>)>
+    ) -> Result<&'tcx abi::call::FnAbi<'tcx, Ty<'tcx>>, ty::layout::FnAbiError<'tcx>> {
+        desc { "computing call ABI of `{}` function pointers", key.value.0 }
+    }
+
+    /// Compute a `FnAbi` suitable for declaring/defining an `fn` instance, and for
+    /// direct calls to an `fn`.
+    ///
+    /// NB: that includes virtual calls, which are represented by "direct calls"
+    /// to an `InstanceDef::Virtual` instance (of `<dyn Trait as Trait>::fn`).
+    query fn_abi_of_instance(
+        key: ty::ParamEnvAnd<'tcx, (ty::Instance<'tcx>, &'tcx ty::List<Ty<'tcx>>)>
+    ) -> Result<&'tcx abi::call::FnAbi<'tcx, Ty<'tcx>>, ty::layout::FnAbiError<'tcx>> {
+        desc { "computing call ABI of `{}`", key.value.0 }
+    }
+
     query dylib_dependency_formats(_: CrateNum)
                                     -> &'tcx [(CrateNum, LinkagePreference)] {
         desc { "dylib dependency formats of crate" }
--- a/compiler/rustc_middle/src/ty/context.rs
+++ b/compiler/rustc_middle/src/ty/context.rs
@@ -118,8 +118,8 @@ pub struct CtxtInterners<'tcx> {
     /// The arena that types, regions, etc. are allocated from.
     arena: &'tcx WorkerLocal<Arena<'tcx>>,
 
-    /// Specifically use a speedy hash algorithm for these hash sets, since
-    /// they're accessed quite often.
+    // Specifically use a speedy hash algorithm for these hash sets, since
+    // they're accessed quite often.
     type_: InternedSet<'tcx, TyS<'tcx>>,
     type_list: InternedSet<'tcx, List<Ty<'tcx>>>,
     substs: InternedSet<'tcx, InternalSubsts<'tcx>>,
@@ -132,9 +132,9 @@ pub struct CtxtInterners<'tcx> {
     projs: InternedSet<'tcx, List<ProjectionKind>>,
     place_elems: InternedSet<'tcx, List<PlaceElem<'tcx>>>,
     const_: InternedSet<'tcx, Const<'tcx>>,
-    /// Const allocations.
-    allocation: InternedSet<'tcx, Allocation>,
+    const_allocation: InternedSet<'tcx, Allocation>,
     bound_variable_kinds: InternedSet<'tcx, List<ty::BoundVariableKind>>,
+    layout: InternedSet<'tcx, Layout>,
 }
 
 impl<'tcx> CtxtInterners<'tcx> {
@@ -152,8 +152,9 @@ impl<'tcx> CtxtInterners<'tcx> {
             projs: Default::default(),
             place_elems: Default::default(),
             const_: Default::default(),
-            allocation: Default::default(),
+            const_allocation: Default::default(),
             bound_variable_kinds: Default::default(),
+            layout: Default::default(),
         }
     }
 
@@ -1062,10 +1063,9 @@ pub struct GlobalCtxt<'tcx> {
     /// Stores memory for globals (statics/consts).
     pub(crate) alloc_map: Lock<interpret::AllocMap<'tcx>>,
 
-    layout_interner: ShardedHashMap<&'tcx Layout, ()>,
-
     output_filenames: Arc<OutputFilenames>,
 
+    // FIXME(eddyb) this doesn't belong here and should be using a query.
     pub(super) vtables_cache:
         Lock<FxHashMap<(Ty<'tcx>, Option<ty::PolyExistentialTraitRef<'tcx>>), AllocId>>,
 }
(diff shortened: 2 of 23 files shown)


eddyb:
Thanks to https://github.com/rust-lang/rust/pull/88499#issuecomment-908688904 (i.e. #88499 being perf-neutral), we should be able to get useful data despite this being based on #88499:

@bors try @rust-timer queue

RalfJung:
> At least miri could make use of this, since it prefers propagating errors (it "just" doesn't use FnAbi yet - cc @RalfJung).

Yeah, it's on my list... 
For LayoutOf, I think we have to properly propagate errors or some CTFE or pattern matching code will end up ICEing. Not sure if the same concern applies here.

bjorn3:
Unfortunately this is a clear perf regression.

eddyb:
Trying without interning, maybe it's a waste of time (tho now there would be cloning, so that's not reassuring):
@bors try @rust-timer queue

eddyb:
Okay, so it's not the interning, at least not to a significant extent.
Next up, `eval_always` (which I'm not sure what it will do, but if it turns off caching, that'd be useful to know the effects of):

@bors try @rust-timer queue

bjorn3:
`eval_always` doesn't turn off caching within a compilation session. It only disables caching across incremental compilation sessions. There is also `anon` which causes no new dep node to be registered, but I believe that still caches.

bjorn3:
This effectively disabled most of incremental compilation it seems by causing all codegen units to always be dirty.

eddyb:
Looking again at the first results (https://github.com/rust-lang/rust/pull/88575#issuecomment-911207024), something is weird: there's a lot of `-check` regressions. I wonder if they're affected specifically by the number of queries, or their key types (size or something else).

eddyb:
Opened #88705 to test a hunch on its own (that the queries existing, even without being used, or even having any providers, can slow things down).

eddyb:
#88705 was either a dead end, or I'm missing some aspect.

Going to try now without the last commit (or rather, with a revert of it), which actually querifies:
@bors try @rust-timer queue

eddyb:
Further combining a lack of queries with a lack of interning, to try and eliminate the full perf difference this PR could produce:
@bors try @rust-timer queue

eddyb:
Now that I got more `-check` regressions with an even more reduced version of the PR, I've opened #88740 for the `intern_ref` -> `intern` change.

If that turns out to be the culprit, I may be able to get away with only arena-allocating `FnAbi`s, instead of interning them.

eddyb:
Maybe the problem with the last run is that it didn't fully remove the `FnAbi` support from the `tcx` arena and interners (it was just unused), so trying again with it removed:

@bors try @rust-timer queue

eddyb:
Alright, that's at least part of the regression tracked down - I need to look at the arena infrastructure to see why unused arenas could cause perf changes - hopefully this isn't something about the size of the `GlobalCtxt` `struct` (backing `TyCtxt`), though that could have workarounds too.

eddyb:
Looks like only interners change the size of `GlobalCtxt`, not arenas (and I can't find any interactions *between* arenas - though if I mark the `FnAbi` as `few` in the macro, I could also remove its space overhead there, since it would use the more dynamic "drop arena").

I've reinstated the queries, but only with arena-allocating, no interning - hopefully this won't impact `-check` benchmarks anymore.

@bors try @rust-timer queue

eddyb:
While this isn't perfect yet, this is much closer to what I wanted to see: a lot of improvements (albeit small), and some regressions where the reuse benefits don't apply, but e.g. incremental tracking gets a bit more overhead.

In fact, I just looked at the query view for the dozen worst regressions, and they *all* have a large increase in `incr_comp_encode_dep_graph`. So maybe it's time to try that `anon` option @bjorn3 mentioned (in https://github.com/rust-lang/rust/pull/88575#issuecomment-911638052).

eddyb:
Now with `anon` queries:

@bors try @rust-timer queue

eddyb:
`anon` doesn't have the impact I was hoping for, that looks more or less identical.

Overall I'm pretty baffled that reserializing the *unmodified* (for `incr-unchanged` at least) dep-graph is the apparent source of the regression, and not the deserializing side.

Are small (or no) changes to the dep-graph inefficiently handled for now, and every query we add comes at a price? (cc @mw @wesleywiser)

bjorn3:
> Are small (or no) changes to the dep-graph inefficiently handled for now, and every query we add comes at a price?

The dep graph is always reserialized from scratch. Small changes will require this in any case as serialization turns the dep node list from a list that may have holes (in case dep nodes were no longer used in the current incr comp session) to a dense list without holes, thus requiring remapping of dep node indexes.

eddyb:
That sounds like a compacting GC, I believe those have thresholds after which they engage, so that small holes don't cause these kinds of latency issues.

nagisa:
@eddyb you want to move this out of draft state?

eddyb:
> @eddyb you want to move this out of draft state?

Done, I'm just worried we can't land it because of the incr-unchanged perf regressions.

compiler/rustc_middle/src/ty/layout.rs
```diff
@@ -2796,9 +2808,39 @@ pub fn conv_from_spec_abi(tcx: TyCtxt<'_>, abi: SpecAbi) -> Conv {
...
+}
+
+impl<'tcx> fmt::Display for FnAbiError<'tcx> {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        match self {
+            Self::Layout(err) => err.fmt(f),
+            Self::AdjustForForeignAbi(err) => err.fmt(f),
+        }
+    }
+}
```
nagisa, line 2832: Typically I would flag this with a request to implement `std::error::Error` and its `source` method, as well as to add some additional information at each layer. In my experience with `thiserror`, its `#[transparent]` errors (which this seems a lot like it) are pretty likely to produce hard to understand errors.

Seems fine in this particular instance though.

compiler/rustc_middle/src/ty/context.rs
```diff
@@ -129,9 +129,9 @@ pub struct CtxtInterners<'tcx> {
     projs: InternedSet<'tcx, List<ProjectionKind>>,
     place_elems: InternedSet<'tcx, List<PlaceElem<'tcx>>>,
     const_: InternedSet<'tcx, Const<'tcx>>,
-    /// Const allocations.
-    allocation: InternedSet<'tcx, Allocation>,
+    const_allocation: InternedSet<'tcx, Allocation>,
```
nagisa, line 132: Nit: This asks for `const_allocations` (it looks like there may be more than 1 of them?)

nagisa (commented):
note to self: 4 commits remain to look at.

compiler/rustc_middle/src/ty/context.rs
eddyb: The interner fields are the names of the interned type, so plurals only show up for interned lists.

compiler/rustc_middle/src/query/mod.rs
```diff
@@ -1136,6 +1136,7 @@ rustc_queries! {
         key: ty::ParamEnvAnd<'tcx, (ty::PolyFnSig<'tcx>, &'tcx ty::List<Ty<'tcx>>)>
     ) -> Result<&'tcx abi::call::FnAbi<'tcx, Ty<'tcx>>, ty::layout::FnAbiError<'tcx>> {
         desc { "computing call ABI of `{}` function pointers", key.value.0 }
+        anon
```
nagisa, line 1139: Is this entirely for performance purposes? It sounds to me like it should be undesirable to enable this.

nagisa:
I don't see any problems from the implementation standpoint here. From the discussion above it also seems like the regressions are largely caused by the implementation of the query system itself?

If so, I could see ourselves arguing that the more queryfied the compiler is, the easier it will be for us to see what the parts of the query system are a lucurative optimization target and spend time on those accordingly.

Basically the local minima vs global minima thing.

compiler/rustc_middle/src/query/mod.rs
eddyb: Hmm, did I keep a HACK comment around? This was for testing perf and can be removed.

nagisa: There's a HACK in the commit message.

oli-obk:
Yea, imo we can merge this if we have an issue for the incremental-unchanged regression that somewhat collects the information from this discussion

wesleywiser:
+1, seems ok to merge to me.

pnkfelix:
Visited for weekly performance triage. Results do indeed seem moderately mixed.

eddyb:
@pnkfelix It's a bit more subtle than that (only `incr-{unchanged,patched}` regressed), and it got discussed before landing (in fact I personally wasn't sure this could land any time soon), see https://github.com/rust-lang/rust/pull/88575#issuecomment-921912025.

PR #88578, opened 2021-09-02
fix(rustc): suggest `items` be borrowed in `for i in items[x..]`

### LINKED ISSUE #87994: Suggest `items` be borrowed in `for i in items[x..]`
Given the following code: 
https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=5bb0aaa179979e54df877b3aafe9f507
```rust
use std::fmt::Display;

fn pretty_list(v: &Vec<impl Display>){
    print!("{}", v[0]);
    for item in v[1..] {
        print!(", {}", item);
    }
}

fn main(){
    let data = vec![0, 1, 2, 3, 4, 5, 6, 7, 8, 9];
    pretty_list(&data);
}
```

The current output is:

```
error[E0277]: the size for values of type `[impl Display]` cannot be known at compilation time
 --> src/main.rs:5:17
  |
5 |     for item in v[1..] {
  |                 ^^^^^^ doesn't have a size known at compile-time
  |
  = help: the trait `Sized` is not implemented for `[impl Display]`
  = note: required because of the requirements on the impl of `IntoIterator` for `[impl Display]`
  = note: required by `into_iter`

error[E0277]: `[impl Display]` is not an iterator
 --> src/main.rs:5:17
  |
5 |     for item in v[1..] {
  |                 ^^^^^^ `[impl Display]` is not an iterator
  |
  = help: the trait `Iterator` is not implemented for `[impl Display]`
  = note: required because of the requirements on the impl of `IntoIterator` for `[impl Display]`
  = note: required by `into_iter`

```

Ideally, the output should note that `&v[1..]` is a possible solution to these errors.

notriddle:
Fixes #87994

### DIFF
--- a/compiler/rustc_trait_selection/src/traits/error_reporting/suggestions.rs
+++ b/compiler/rustc_trait_selection/src/traits/error_reporting/suggestions.rs
@@ -23,7 +23,7 @@ use rustc_middle::ty::{
 use rustc_middle::ty::{TypeAndMut, TypeckResults};
 use rustc_span::def_id::LOCAL_CRATE;
 use rustc_span::symbol::{kw, sym, Ident, Symbol};
-use rustc_span::{BytePos, MultiSpan, Span, DUMMY_SP};
+use rustc_span::{BytePos, DesugaringKind, ExpnKind, ForLoopLoc, MultiSpan, Span, DUMMY_SP};
 use rustc_target::spec::abi;
 use std::fmt;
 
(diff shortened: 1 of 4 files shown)


nagisa:
Can you also add a test for the negative case if it doesn't already exist? i.e. a case where adding a borrow wouldn't actually help. It is still fine if the compiler does suggest (incorrectly) borrowing there, but at the very least we'll know what exactly is happening.

r=me otherwise.

notriddle:
@nagisa new test case for mutating through an iterator.

nagisa:
@bors r+

Thanks!

PR #88579, opened 2021-09-02
remove redundant / misplaced sentence from docs

ast-ral:
Removes sentence that seems to have drifted down into the examples section. Luckily, someone already added an explanation of what happens with packed structs back into the initial section of the doc entry and this wayward sentence can likely just be deleted.


m-ou-se:
Thanks!

@bors r+ rollup

PR #88587, opened 2021-09-02
Fix WinUWP std compilation errors due to I/O safety

bdbai:
I/O safety for Windows has landed in #87329. However, it does not cover UWP specific parts and prevents all UWP targets from building. See https://github.com/YtFlow/Maple/issues/18. This PR fixes these compile errors when building std for UWP targets.


joshtriplett:
Seems reasonable. `get_handle` is handling the invalid and null cases, so the unsafe block seems fine. I'd like to have a `// SAFETY` comment above the unsafe blocks. With that comment added, r=me.

PR #88588, opened 2021-09-02
Make feature key optional for rustc_stable, rustc_const_stable attributes

jplatte:
… so partial stabilizations no longer have to make up a feature name that never existed as unstable for the stabilized subset of the API.

This seems to break at runtime, presumably because the on-disk representation of the affected `rustc_attr` types changed. Not sure how to proceed here. If sbd. could give me a hint about how this could be made to work that would be appreciated.

Brief Zulip thread about this: https://rust-lang.zulipchat.com/#narrow/stream/219381-t-libs/topic/.60feature.60.20key.20in.20rustc_stable.20and.20rustc_const_stable

### DIFF
--- a/src/librustdoc/lib.rs
+++ b/src/librustdoc/lib.rs
@@ -12,6 +12,7 @@
 #![feature(crate_visibility_modifier)]
 #![feature(never_type)]
 #![feature(once_cell)]
+#![feature(option_result_contains)]
 #![feature(type_ascription)]
 #![recursion_limit = "256"]
 #![warn(rustc::internal)]
--- a/compiler/rustc_attr/src/builtin.rs
+++ b/compiler/rustc_attr/src/builtin.rs
@@ -95,7 +95,6 @@ pub enum OptimizeAttr {
 #[derive(HashStable_Generic)]
 pub struct Stability {
     pub level: StabilityLevel,
-    pub feature: Symbol,
 }
 
 /// Represents the `#[rustc_const_unstable]` and `#[rustc_const_stable]` attributes.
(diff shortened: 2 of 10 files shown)


jplatte:
Updated rustdoc & clippy, maybe now the failure I was seeing will pop up here. Although I'm not seeing it anymore myself after commenting out `profile = "compiler"` in `config.toml` 🤔

src/librustdoc/lib.rs
```diff
@@ -12,6 +12,7 @@
 #![feature(crate_visibility_modifier)]
 #![feature(never_type)]
 #![feature(once_cell)]
+#![feature(option_result_contains)]
```
jplatte, line 15: Just noticed I left this in but removed the use of `Option::contains`. Will remove with the next push.

jplatte:
One more thing that would be simplified: New trait impls (which are always insta-stable) for which authors would no longer have to make up a meaningless feature name.

ping @wesleywiser @m-ou-se what do I do to move this forwards (towards a decision on whether this changes makes sense or not)? Open an MCP?

wesleywiser:
> what do I do to move this forwards (towards a decision on whether this changes makes sense or not)?

This feels to me like a question for the libs team. I'm not sure what processes they use but bumping that Zulip thread would probably be a good idea. I'm happy to review the implementation once they've signed off on the feature design.

jplatte:
I posted to the Zulip thread again on Oct 6, but got no responses. Is this going to be discussed?

m-ou-se:
This was discussed in the library team meeting last week. Our conclusion was that we don't think this change really solves much, while it does make it easier to make the mistake of *not* having a feature name in cases where we do want it. We also do not want to give the impression that unstable feature names are something 'stable' that can be tracked and doesn't change meaning over time.

jplatte:
What does it mean that this is now `S-waiting-on-author`? Shouldn't it just be closed?

m-ou-se:
I didn't close it yet because I wanted to make there's space for you to respond. But unless you have new arguments to change opinions here, this should probably be closed, yes.

PR #88594, opened 2021-09-02
More symbolic doc aliases

steffahn:
A bunch of small changes, mostly adding `#[doc(alias = "…")]` entries for symbolic `"…"`.

Also a small change in documentation of `const` keywords.

### DIFF
--- a/library/core/src/ops/bit.rs
+++ b/library/core/src/ops/bit.rs
@@ -30,6 +30,7 @@
 /// ```
 #[lang = "not"]
 #[stable(feature = "rust1", since = "1.0.0")]
+#[doc(alias = "!")]
 pub trait Not {
     /// The resulting type after applying the `!` operator.
     #[stable(feature = "rust1", since = "1.0.0")]
--- a/library/std/src/primitive_docs.rs
+++ b/library/std/src/primitive_docs.rs
@@ -387,6 +387,9 @@ mod prim_unit {}
 
 #[doc(primitive = "pointer")]
 #[doc(alias = "ptr")]
+#[doc(alias = "*")]
+#[doc(alias = "*const")]
+#[doc(alias = "*mut")]
 //
 /// Raw, unsafe pointers, `*const T`, and `*mut T`.
 ///
--- a/library/std/src/primitive_docs.rs
+++ b/library/std/src/primitive_docs.rs
@@ -385,8 +385,8 @@ mod prim_char {}
 #[stable(feature = "rust1", since = "1.0.0")]
 mod prim_unit {}
 
-#[doc(alias = "ptr")]
 #[doc(primitive = "pointer")]
+#[doc(alias = "ptr")]
 //
 /// Raw, unsafe pointers, `*const T`, and `*mut T`.
 ///
@@ -499,10 +499,10 @@ mod prim_unit {}
 #[stable(feature = "rust1", since = "1.0.0")]
 mod prim_pointer {}
 
+#[doc(primitive = "array")]
 #[doc(alias = "[]")]
 #[doc(alias = "[T;N]")] // unfortunately, rustdoc doesn't have fuzzy search for aliases
 #[doc(alias = "[T; N]")]
-#[doc(primitive = "array")]
 /// A fixed-size array, denoted `[T; N]`, for the element type, `T`, and the
 /// non-negative compile-time constant size, `N`.
 ///
--- a/library/std/src/keyword_docs.rs
+++ b/library/std/src/keyword_docs.rs
@@ -119,7 +119,7 @@ mod break_keyword {}
 
 #[doc(keyword = "const")]
 //
-/// Compile-time constants and compile-time evaluable functions.
+/// Compile-time constants, compile-time evaluable functions, and raw pointers.
 ///
 /// ## Compile-time constants
 ///


library/std/src/primitive_docs.rs
```diff
@@ -499,10 +517,38 @@ mod prim_unit {}
...
+#[doc(alias = "[T; 1]")]
+#[doc(alias = "[T;1")]
+#[doc(alias = "[T; 1")]
+#[doc(alias = ";")]
+#[doc(alias = ";0")]
+#[doc(alias = "; 0")]
+#[doc(alias = ";1")]
+#[doc(alias = "; 1")]
+#[doc(alias = ";N")]
+#[doc(alias = "; N")]
```
jyn514, line 551: Rather than adding a whole bunch of these, would you be interested in adding fuzzy search support for doc aliases? That seems more useful, and less maintenance for the standard library.

steffahn: Eh… I’d have no idea how to do something like that. That search is probably JavaScript? I never really used JavaScript before… And also I’m not sure if “full” fuzzy search for non-symbolic doc aliases is even wanted in the general case. E.g. if I type “`in`”, I’m not sure if I’d want to see the “inject” → “Iterator::fold” alias in my search results.

steffahn: I do admit, that this list may have become a bit lengthy. Feel free to point out anything that goes “too far” :-D

(also, for this reason the PR is split up into many individual commits, so the “overkill” changes – if any – can be named and called out easily)

steffahn: I also figured out (after the fact) that `#[doc(alias = "…", alias = "…", …)]` lists are a thing, so if the number of lines should be reduced, I could group them.

library/core/src/ops/bit.rs
```diff
@@ -668,6 +669,7 @@ shr_impl_all! { u8 u16 u32 u64 u128 usize i8 i16 i32 i64 i128 isize }
 /// assert_eq!(bv, expected);
 /// ```
 #[lang = "bitand_assign"]
+#[doc(alias = "&")]
```
the8472, line 672: Putting an alias for the non-assignment operator on the assigning flavor would violate this part of the std doc policy

> There must be an obvious single target for the alias that is an exact analogue of the aliased name. We will not add the same alias to multiple declarations. (const and non-const versions of the same function are fine.) We will also not add an alias for a function that's only somewhat similar or related.

https://std-dev-guide.rust-lang.org/documentation/doc-alias-policy.html

library/std/src/primitive_docs.rs
the8472: Iirc the search already does fuzzy searching (based on levenshtein distance), maybe that just needs to be extended to aliases or something if it doesn't already cover these cases.

library/core/src/ops/bit.rs
steffahn: I doubt that those guidelines do fully apply too *symbolic* aliases. I.e. aliases that aren't an identifier. I was also following a convention that seems to already exist for several operators like `+=`, `-=`, `*=`, and `/=`, e. g.https://doc.rust-lang.org/std/?search=%2B

library/std/src/primitive_docs.rs
jyn514: Search lives around here IIRC: https://github.com/rust-lang/rust/blob/c5799b2a73bd03d1d7b4a432fa0a0cb93e7288db/src/librustdoc/html/static/js/search.js#L79

library/core/src/ops/bit.rs
the8472: It may be the case that those aliases should be removed rather than be taken as precedent. Especially if they only exist to work around a search limitation (see other comment thread).

library/core/src/ops/range.rs
```diff
@@ -335,6 +335,7 @@ impl<Idx: PartialOrd<Idx>> RangeTo<Idx> {
 /// assert_eq!(arr[1..=3], [   1, 2, 3   ]); // This is a `RangeInclusive`
 /// ```
 #[lang = "RangeInclusive"]
+#[doc(alias = "..")]
```
joshtriplett, line 338: This shouldn't be on RangeInclusive.

library/core/src/ops/function.rs
```diff
@@ -216,6 +216,10 @@ pub trait FnMut<Args>: FnOnce<Args> {
 )]
 #[fundamental] // so that regex can rely that `&str: !FnMut`
 #[must_use = "closures are lazy and do nothing unless called"]
+#[doc(alias = "(")]
+#[doc(alias = ")")]
+#[doc(alias = "()")]
+#[doc(alias = "( )")]
```
joshtriplett, line 222: These sholdn't be on closures; they're already on unit.

steffahn: I was thinking of function-call syntax, i.e. `f()` where `f: FnOnce`, similar to the `[]` aliases on the `Index` trait for supporting `v[]` for `v: Index`.

library/core/src/ops/range.rs
steffahn: The idea is that it’s allowing searching for the prefix `..` that’s in common for both `..` and `..=`. Similar to how [`+` finds `AddAssign`](https://doc.rust-lang.org/std/?search=%2B) even though that’s `+=` as an operator.

library/core/src/ops/index.rs
```diff
@@ -163,6 +163,12 @@ see chapter in The Book <https://doc.rust-lang.org/book/ch08-02-strings.html#ind
 #[doc(alias = "[")]
 #[doc(alias = "]")]
 #[doc(alias = "[]")]
+#[doc(alias = "[ ]")]
+#[doc(alias = "[0]")]
+#[doc(alias = "[1]")]
+#[doc(alias = "[i]")]
+#[doc(alias = "[j]")]
+#[doc(alias = "[n]")]
```
Mark-Simulacrum, line 171: I am a little surprised to see `[` and `]` as aliases here, and the 'random' indices (0, 1) and characters here also seem a little bit going too far. In practice, if you know that what you want is indexing, you'll probably look for "index", right?

And if not, then I imagine it won't occur to most people to use the standard library docs with `[1]` or similar -- and I don't know that we should encourage that. If you don't really know what `[1]` does, this is probably not the ideal place to land -- something like https://doc.rust-lang.org/nightly/reference/expressions/array-expr.html might be preferable, or even https://doc.rust-lang.org/nightly/book/ch03-02-data-types.html#accessing-array-elements. In any case, it seems a rare case, and feels like the wrong use for aliases to me.

Mark-Simulacrum:
Going to reassign to @joshtriplett for T-libs(-api?) discussion -- seems like some interpretation of the doc alias policy for symbols is needed.

library/core/src/ops/index.rs
steffahn: I agree, these probably go to far. I was doing them because `array` had `[T; N]` and `[T;N]` synonyms with seem equally questionable IMO. Who says the _type_ is called `T` and the _number_ called `N`? So I’ve added `[T; 0]` and `[T; 1]` in an effort to get more realistic small numbers, and for indexing there wasn’t really a canonical variable name, so `i`, `j` and `n` seemed reasonable.

How much shall be removed then? Are `[T; N]` and `[T;N]` for `array`s really wanted? Should all the symbolic aliases be removed from `Index` and `IndexMut` be removed (and consequently also the `(` and `)` -based aliases for `FnOnce` that I’ve added)?

library/core/src/ops/bit.rs
steffahn: Alright… what’s the proper order to do that in then? Remove the other aliases now? Or wait until search is improved?

jyn514: IMO, this whole PR should be closed, rustdoc should be changed to handle fuzzy search, and then we can remove most of the remaining symbolic aliases.

library/core/src/ops/index.rs
steffahn: OTOH, if we get a fuzzy search for aliases, terms like ` [` or `[;]` could find an alias `[T; N]` which would in and by itself then be a better preview of the syntax that this is actually about. One could throw in a `[a, b, c]` alias for array expressions, too. Similarly, indexing could be something like `a[i]` and function-call expressions could be `f(x)` or `f(x, y, z)`.

library/core/src/ops/bit.rs
steffahn: > IMO, this whole PR should be closed

I’d want to keep 49c680ada001a4c296d1c00f58be8d8b535d50f6, ab89c88faadbc5a4063c8473c9ce28af16324d79, 56634a1ef35a591aa735e883e42c976cf25cffdc, and part of 0a565985d178b3fa70af26f4fb3c8719b6e10f7e though.

For 0a565985d178b3fa70af26f4fb3c8719b6e10f7e I’d want to keep `*`, `*const`, and `*mut` for now with the intention of `*` being removed at a later time. Though the inclusion of `*` can be debated.

steffahn:
([in review above](https://github.com/rust-lang/rust/pull/88594#discussion_r702283867))
> > IMO, this whole PR should be closed
> 
> I’d want to at least keep [49c680a](https://github.com/rust-lang/rust/commit/49c680ada001a4c296d1c00f58be8d8b535d50f6), [ab89c88](https://github.com/rust-lang/rust/commit/ab89c88faadbc5a4063c8473c9ce28af16324d79), [56634a1](https://github.com/rust-lang/rust/commit/56634a1ef35a591aa735e883e42c976cf25cffdc), and part of [0a56598](https://github.com/rust-lang/rust/commit/0a565985d178b3fa70af26f4fb3c8719b6e10f7e) though.
> 
> For [0a56598](https://github.com/rust-lang/rust/commit/0a565985d178b3fa70af26f4fb3c8719b6e10f7e) I’d want to keep `*`, `*const`, and `*mut` for now with the intention of `*` being removed at a later time. Though the inclusion of `*` can be debated.

I’ll do that in a rebase -i + force-push, I guess… Review hasn’t really addressed any of these changes yet anyway. After rebase I’ll resolve the reviews on the removed changes.

jyn514:
so it doesn't get lost in the resolved thread - the link to rustdoc's fuzzy search is https://github.com/rust-lang/rust/blob/c5799b2a73bd03d1d7b4a432fa0a0cb93e7288db/src/librustdoc/html/static/js/search.js#L79

library/core/src/ops/function.rs
joshtriplett: I realize that, but that would make these ambiguous.

joshtriplett:
The remaining items look reasonable to me.

@bors r+ rollup

library/core/src/ops/function.rs
steffahn: I’m not sure what you mean by “ambiguous”. We already have `[]` on `slice`, `array`, `Index`, and `IndexMut`. We already have `()` on `unit` and `tuple`. I don’t see any additional ambiguity. Nonetheless, following the discussion in [this thread](https://github.com/rust-lang/rust/pull/88594#discussion_r702310581), a lot of changes, including the new alias for closures, are already removed from this PR, anyway.

Issue #34255, opened 2016-06-13
Misleading error message on typo "Type ascription"


danielpclark:
This had me googling for answers.  But I had entered a colon instead of a semi-colon at the end of a line.  Nothing in my search results pointed that out.

```
src/chop_basename.rs:26:3: 29:29 error: type ascription is experimental (see issue #23416)
src/chop_basename.rs:26   parts.push(CString::new(basename).unwrap().into_raw()):
src/chop_basename.rs:27   // END BASENAME
src/chop_basename.rs:28   
src/chop_basename.rs:29   RubyArray::from_vec(parts)
src/lib.rs:23:1: 23:30 note: in this expansion of include!
src/chop_basename.rs:26:3: 29:29 help: add #![feature(type_ascription)] to the crate attributes to enable
error: aborting due to previous error
```

shepmaster:
Another from Stack Overflow (missing struct name in constructor):

```rust
impl Reactor {
    pub fn new() -> Self {
        input_cells: Vec::new()
    }
}
```

gsingh93:
Same issue here (accidentally typed `Foo:foo()` instead of `Foo::foo()`):

```
struct Foo {}
impl Foo {
    fn foo() {}
}

fn main() {
    Foo:foo();
}
```

Mark-Simulacrum:
From the two issues I just closed:

```rust
fn foo() {
    let two = 2:
    let two_squared = two * two;
}

fn bar() {
x : i32 = 0;
}
```

tokideveloper:
I was also confused by that strange error message. Here is my example: Compiling

```rust
fn main() {
    let s = String:from("Just a string.");
    println!("{}", s);
}
```

using rustc 1.29.1 leads to

```
error: expected type, found `"Just a string."`
 --> src/main.rs:2:25
  |
2 |     let s = String:from("Just a string.");
  |                         ^^^^^^^^^^^^^^^^ expecting a type here because of type ascription
```

A little line could really help:

```
error: expected type, found `"Just a string."`
 --> src/main.rs:2:25
  |
2 |     let s = String:from("Just a string.");
  |                         ^^^^^^^^^^^^^^^^ expecting a type here because of type ascription
                      - help: Did you mean `::`?
```

Would that be feasible to implement?
Have I overseen any obstacles towards it?

estebank:
`Foo:foo()`/`String:from("Just a string.");` is handled in #59150.

`let two = 2:` case is already handled.

[...]

PR #88598, opened 2021-09-02
Detect bare blocks with type ascription that were meant to be a `struct` literal

estebank:
Address part of #34255.

Potential improvement: silence the other knock down errors in `issue-34255-1.rs`.


wesleywiser (approved):
Nice!

PR #88601, opened 2021-09-02
Implement `Termination` for `Result<Infallible, E>`

ibraheemdev:
As noted in #43301, `Result<!, E>` is not usable on stable.

### DIFF
--- a/library/std/src/process.rs
+++ b/library/std/src/process.rs
@@ -106,6 +106,7 @@ mod tests;
 
 use crate::io::prelude::*;
 
+use crate::convert::Infallible;
 use crate::ffi::OsStr;
 use crate::fmt;
 use crate::fs;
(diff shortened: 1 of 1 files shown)


library/std/src/process.rs
```diff
@@ -2051,6 +2052,15 @@ impl<E: fmt::Debug> Termination for Result<!, E> {
     }
 }
 
+#[unstable(feature = "termination_trait_lib", issue = "43301")]
+impl<E: fmt::Debug> Termination for Result<Infallible, E> {
+    fn report(self) -> i32 {
+        let Err(err) = self;
+        eprintln!("Error: {:?}", err);
+        ExitCode::FAILURE.report()
```
xfix, line 2060: This could forward to `Termination` implementation for `Result<!, E>` instead of duplicating code.

```suggestion
        Err::<!, _>(err).report()
```

sugar700 (changes requested):

sugar700 (approved):

kennytm:
r? @rust-lang/libs-api

While the `Termination` trait itself is unstable, merging this PR means the following construct becomes insta-stable, so I think an FCP is needed:

```rust
fn main() -> Result<std::convert::Infallible, &'static str> {
    Err("error")
}
```

BTW it is strange that `impl Termination for Infallible` does not exist.

xfix:
> BTW it is strange that `impl Termination for Infallible` does not exist.

There is no real need for that, `fn main() -> ! { ... }` works fine even on stable.

yaahc:
@rfcbot merge

dtolnay:
@rfcbot fcp merge

PR #88604, opened 2021-09-02
rustdoc: Clean up handling of lifetime bounds

camelid:
Previously, rustdoc recorded lifetime bounds by rendering them into the
name of the lifetime parameter. Now, it leaves the name as the actual
name and instead records lifetime bounds in an `outlives` list, similar
to how type parameter bounds are recorded.

Also, higher-ranked lifetimes cannot currently have bounds, so I simplified
the code to reflect that.

r? @GuillaumeGomez

### DIFF
--- a/src/librustdoc/clean/auto_trait.rs
+++ b/src/librustdoc/clean/auto_trait.rs
@@ -331,9 +331,10 @@ impl<'a, 'tcx> AutoTraitFinder<'a, 'tcx> {
                 match br {
                     // We only care about named late bound regions, as we need to add them
                     // to the 'for<>' section
-                    ty::BrNamed(_, name) => {
-                        Some(GenericParamDef { name, kind: GenericParamDefKind::Lifetime })
-                    }
+                    ty::BrNamed(_, name) => Some(GenericParamDef {
+                        name,
+                        kind: GenericParamDefKind::Lifetime { outlives: vec![] },
+                    }),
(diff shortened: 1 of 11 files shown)


GuillaumeGomez (approved):

GuillaumeGomez:
You did it faster than I expected. Looks great, thanks a lot!

@bors: r+

jyn514:
@bors r-

This doesn't update format-version in rustdoc-json-types. Please read the messages from highfive, I set them up for a reason.

jyn514:
Oh sorry, I missed https://github.com/rust-lang/rust/pull/88604/files#diff-1a2ef1ea43fd65123fdf8a158a34408aa7c31c4bdb34ce1458b7f6c22b3ab459R237. Seems weird that the format-version is in rustdoc and not in rustdoc-json-types - maybe we should change that at some point?

@bors r=GuillaumeGomez

GuillaumeGomez:
> Oh sorry, I missed https://github.com/rust-lang/rust/pull/88604/files#diff-1a2ef1ea43fd65123fdf8a158a34408aa7c31c4bdb34ce1458b7f6c22b3ab459R237. Seems weird that the format-version is in rustdoc and not in rustdoc-json-types - maybe we should change that at some point?
> 
> @bors r=GuillaumeGomez

Please open an issue so we don't forget. ;)

Issue #88606, opened 2021-09-03
[lib] Prefer `core::hint::unreachable_unchecked` over `core::unreachable` macro


Some-Dood:
At the moment, many of the core assertions in the standard library make use of the `unreachable` macro, which is simply an indirection for `panic`. This is totally fine, but I would like to begin an initiative on upgrading some of these invocations as `core::hint::unreachable_unchecked` so that we may (at least) give the compiler a chance to optimize the code better.

Below is a _non-exhaustive_ list of possible candidates. I will be editing the issue as more candidates are introduced and as certain examples are marked 100% safe or not.

# Merged
None so far.

# Definitely Safe
This is the perfect candidate for `core::hint::unreachable_unchecked` since it is apparent that `self` is now the `Cow::Owned` variant.
https://github.com/rust-lang/rust/blob/97f2698484c2d668b2f356a6c2252e5f08472234/library/alloc/src/borrow.rs#L275-L287

# Possible Candidates
https://github.com/rust-lang/rust/blob/673d0db5e393e9c64897005b470bfeb6d5aec61b/library/core/src/lazy.rs#L52-L64

https://github.com/rust-lang/rust/blob/adf168844791b213383b18d10515eaee88695f7d/library/std/src/lazy.rs#L121-L130

https://github.com/rust-lang/rust/blob/adf168844791b213383b18d10515eaee88695f7d/library/std/src/lazy.rs#L152-L158

https://github.com/rust-lang/rust/blob/17f30e5451f581d753899d2f628e5be354df33cd/library/std/src/sys/sgx/abi/mod.rs#L51-L52

https://github.com/rust-lang/rust/blob/673d0db5e393e9c64897005b470bfeb6d5aec61b/library/std/src/sync/mpsc/oneshot.rs#L105-L106

https://github.com/rust-lang/rust/blob/673d0db5e393e9c64897005b470bfeb6d5aec61b/library/std/src/sync/mpsc/oneshot.rs#L170-L173

https://github.com/rust-lang/rust/blob/673d0db5e393e9c64897005b470bfeb6d5aec61b/library/std/src/sync/mpsc/oneshot.rs#L188-L190

https://github.com/rust-lang/rust/blob/673d0db5e393e9c64897005b470bfeb6d5aec61b/library/std/src/sync/mpsc/oneshot.rs#L251-L252

[...]

PR #88608, opened 2021-09-03
Chore: prefer `unreachable_unchecked` in `Cow::to_mut` method

BastiDood:
As a small first step in #88606, this PR uses the `core::hint::unreachable_unchecked` function in place of the `unreachable` macro in the `Cow::to_mut` implementation. It should be fairly apparent that this is a safe assertion to make since `self` has just been assigned to the `Owned` variant. Thanks!

### DIFF
--- a/library/alloc/src/borrow.rs
+++ b/library/alloc/src/borrow.rs
@@ -4,6 +4,7 @@
 
 use core::cmp::Ordering;
 use core::hash::{Hash, Hasher};
+use core::hint::unreachable_unchecked;
 use core::ops::Deref;
 #[cfg(not(no_global_oom_handling))]
 use core::ops::{Add, AddAssign};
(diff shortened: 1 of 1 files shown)


hellow554:
As already said in the other thread, this actually doesn't change anything, because the compiler is smart enough to detect, that the branch will be triggered. See https://rust.godbolt.org/z/c75xv4df4 for a comparision

Some-Dood:
> ...this actually doesn't change anything...

Thanks for taking the time to set up the Godbolt example! It's incredible how the compiler even removed the entire panic branch regardless. For the meantime, I'll be looking forward to the discussion in the original thread before resolving this PR. 👍

dtolnay (commented):
Compilers are way better at this than humans are. Making the human do it, when the compiler is doing the same already, is not an improvement. That puts the maintainers now perpetually on the hook for absence of UB going forward, in the presence of changes to the surrounding code.

PR #88612, opened 2021-09-03
Add a better error message for #39364

lovasoa:
There is a known bug in the implementation of mpsc channels in rust.
This adds a clearer error message when the bug occurs, so that developers don't lose too much time looking for the origin of the bug.
See https://github.com/rust-lang/rust/issues/39364

### DIFF
--- a/library/std/src/sync/mpsc/shared.rs
+++ b/library/std/src/sync/mpsc/shared.rs
@@ -248,7 +248,11 @@ impl<T> Packet<T> {
     // Returns true if blocking should proceed.
     fn decrement(&self, token: SignalToken) -> StartResult {
         unsafe {
-            assert_eq!(self.to_wake.load(Ordering::SeqCst), 0, "This is a known bug in rust. See https://github.com/rust-lang/rust/issues/39364");
+            assert_eq!(
+                self.to_wake.load(Ordering::SeqCst),
+                0,
+                "This is a known bug in rust. See https://github.com/rust-lang/rust/issues/39364"
+            );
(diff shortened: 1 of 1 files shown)


joshtriplett:
Seems reasonable to me. Needs a rustfmt, but otherwise OK.

I'm going to nominate this for a libs team meeting, so that we can discuss the broader issues around mpsc, and confirm that we want to make this change in the interim.

m-ou-se:
I'm not sure I like the idea of refering to Rust and Rust issues from panic/error messages in the standard library. Messages from the compiler will be shown to the Rust developer, but messages like these will be shown to the user of the program, for who this might be quite meaningless or confusing.

On the other hand, the alternative of just 'assertion failed' isn't much better of course.

library/std/src/sync/mpsc/shared.rs
```diff
@@ -248,7 +248,11 @@ impl<T> Packet<T> {
     // Returns true if blocking should proceed.
     fn decrement(&self, token: SignalToken) -> StartResult {
         unsafe {
-            assert_eq!(self.to_wake.load(Ordering::SeqCst), 0);
+            assert_eq!(
+                self.to_wake.load(Ordering::SeqCst),
+                0,
+                "This is a known bug in rust. See https://github.com/rust-lang/rust/issues/39364"
```
m-ou-se, line 254:
```suggestion
                "This is a known bug in the Rust standard library. See https://github.com/rust-lang/rust/issues/39364"
```

PR #88613, opened 2021-09-03
Update primitive docs for rust 2021.

### LINKED ISSUE #87701: Update array and iterator docs in std / core for new IntoIterator impl and edition.

m-ou-se:
Fixes https://github.com/rust-lang/rust/issues/87701


m-ou-se:
Thanks for the quick reviews!

PR #88614, opened 2021-09-03
Improve `Cow::clone_from`

a1phyr:
Previously this code would cause an allocation:

```rust
let mut dst = Cow::Owned("Hello".to_string());
let src = Cow::Borrowed("Hi");

dst.clone_from(src);
dst.to_mut()
```

After this PR this is no longer the case.

### DIFF
--- a/library/alloc/src/borrow.rs
+++ b/library/alloc/src/borrow.rs
@@ -204,9 +204,12 @@ impl<B: ?Sized + ToOwned> Clone for Cow<'_, B> {
     }
 
     fn clone_from(&mut self, source: &Self) {
-        match (self, source) {
-            (&mut Owned(ref mut dest), &Owned(ref o)) => o.borrow().clone_into(dest),
-            (t, s) => *t = s.clone(),
+        match self {
+            Owned(dest) => {
+                let src: &B = source.borrow();
+                src.clone_into(dest);
+            }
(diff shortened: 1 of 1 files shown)


m-ou-se:
Your PR seems to do the exact opposite:

```rust
use std::borrow::Cow;

struct X;

impl Clone for X {
    fn clone(&self) -> Self {
        panic!("clone!");
    }
}

fn main() {
    let mut dst = Cow::<X>::Owned(X);
    let src = Cow::Borrowed(&X);

    dst.clone_from(&src);
}
```

Before your PR:
```
$ cargo r
   Compiling scratchpad v0.1.0
    Finished dev [unoptimized + debuginfo] target(s) in 0.48s
     Running `target/debug/scratchpad`
$
```

After your PR:
```
$ cargo +stage1 r
   Compiling scratchpad v0.1.0
    Finished dev [unoptimized + debuginfo] target(s) in 0.21s
     Running `target/debug/scratchpad`
thread 'main' panicked at 'clone!', src/main.rs:7:9
$
```

dtolnay (commented):
I agree with @m-ou-se's comment; I believe the behavior of this impl was correct prior to this PR. `Owned`/`Owned` is the only case where `clone_into` is applicable as an optimization.

PR #88619, opened 2021-09-03
Remove `cfg(doc)` from std::os module reexports to fix rustdoc linking issues

### LINKED ISSUE #88304: "jump to definition" feature is generating links to private item documentation pages
More information on https://github.com/rust-lang/rust/pull/88292

GuillaumeGomez:
Fixes https://github.com/rust-lang/rust/issues/88304.

I tested it based on https://github.com/rust-lang/rust/pull/88292.

Not sure if it's the best approach, but at least it makes thing a bit simpler.

cc @jyn514


jyn514:
I don't think this is the right approach. We should find why the link-to-definition pass is broken and fix it, not change the standard library to avoid rustdoc bugs.

GuillaumeGomez:
Well, it's not specific to the jump to definition feature. The link itself is correct after all... The problem is that we have a different rendering with "cfg(doc)". We literally can't fix it.

jyn514:
Got it, that makes sense. This seems like a reasonable change then, to make the information available to rustdoc when documenting it as a dependency.

GuillaumeGomez:
I'll check what's wrong with the changes I made.

GuillaumeGomez:
So: in the first commit I used `cfg-if` to have a clean state to work on. But because of https://github.com/rust-analyzer/rust-analyzer/issues/6038, I removed it in the second commit and reverted back to code without macro.

GuillaumeGomez:
Seems like it's ready for review! :)

jyn514 (commented):
@GuillaumeGomez I was planning to let @joshtriplett review, I don't think we should be changing fundamental things about libstd without asking T-libs.

GuillaumeGomez:
> @GuillaumeGomez I was planning to let @joshtriplett review, I don't think we should be changing fundamental things about libstd without asking T-libs.

Which is why I didn't set you as reviewer. ;)

But reviews from you remain very welcome!

library/std/src/os/mod.rs
```diff
@@ -10,117 +10,121 @@ pub mod raw;
...
+))]
+#[unstable(issue = "none", feature = "std_internals")]
+pub mod windows {}
+
+// unix
+#[cfg(not(any(
+    all(target_arch = "wasm32", not(target_os = "wasi")),
+    all(target_vendor = "fortanix", target_env = "sgx")
+)))]
+#[cfg(target_os = "hermit")]
```
Amanieu, line 60:
```suggestion
#[cfg(all(not(doc), target_os = "hermit"))]
```
Otherwise `hermit` doc builds will have 2 unix modules.

GuillaumeGomez:
Updated!

GuillaumeGomez:
Seems like `wasm32` is picking `unix` for whatever reason...

Amanieu:
`wasm32-emscripten` is unix, `wasm32-wasi` and `wasm32-unknown` aren't.

GuillaumeGomez:
Interesting, didn't know that. I'll update then.

GuillaumeGomez:
@matthiaskrgr: Any idea what's going on with the clippy lint by any chance?

camsteffen:
That lint uses the absolute path of `Permissions` [here](https://github.com/rust-lang/rust-clippy/blob/92ca25b3ec8cef291f1d3b37bfd01d3cce8609e2/clippy_lints/src/non_octal_unix_permissions.rs#L52) and I'm guessing the path is changed. You can change the path in Clippy. The proper fix is to use a diagnostic item.

GuillaumeGomez:
Oh I see, thanks!

PR #88627, opened 2021-09-03
Do not preallocate HirIds

cjgillot:
Part of https://github.com/rust-lang/rust/pull/87234

r? @petrochenkov

### DIFF
--- a/compiler/rustc_ast_lowering/src/item.rs
+++ b/compiler/rustc_ast_lowering/src/item.rs
@@ -40,6 +40,7 @@ impl ItemLowerer<'_, '_, '_> {
 
 impl<'a> Visitor<'a> for ItemLowerer<'a, '_, '_> {
     fn visit_item(&mut self, item: &'a Item) {
+        self.lctx.allocate_hir_id_counter(item.id);
         let hir_id = self.lctx.with_hir_id_owner(item.id, |lctx| {
             lctx.without_in_scope_lifetime_defs(|lctx| {
                 let hir_item = lctx.lower_item(item);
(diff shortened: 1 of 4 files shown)


Mark-Simulacrum:
(Perf is now ready)

petrochenkov:
Aha, here's the `style-servo-check` regression.

cjgillot:
This PR also creates a `Found unstable fingerprints for opt_def_kind` ICE in the incr-patched cherry-pick case. I need to investigate it before this is merged.

petrochenkov:
@cjgillot 
Could you set the `S-waiting-on-review` label once this is ready?
I see that you are updating the lowering related PRs from time to time, but I'm not sure whether they are finished or not, so I didn't take any action.

cjgillot:
With the latest perfect results, I believe this one is ready.

compiler/rustc_ast_lowering/src/item.rs
```diff
@@ -928,7 +920,7 @@ impl<'hir> LoweringContext<'_, 'hir> {
         let has_value = true;
         let (defaultness, _) = self.lower_defaultness(i.kind.defaultness(), has_value);
         hir::ImplItemRef {
-            id: hir::ImplItemId { def_id: self.lower_node_id(i.id).expect_owner() },
+            id: hir::ImplItemId { def_id: self.allocate_hir_id_counter(i.id) },
```
petrochenkov, line 923: I'm not sure why this difference between `lower_trait_item_ref` and `lower_impl_item_ref` exists.
For both of them `allocate_hir_id_counter` should have already been called by `visit_assoc_item`?

cjgillot: `lower_impl_item_ref` is called as part of `lower_item` for the impl block. At that time, `allocate_hir_id_counter` has not been called yet for the impl_item. However, this `allocate_hir_id_counter` is required for the call to `lower_visibility`. This difference is removed by https://github.com/rust-lang/rust/pull/87234 as it removes the call to ``lower_visibility`.

PR #85796, opened 2021-05-28
MIR opt: Expand aggregates into multiple locals

cjgillot:
The MIR deaggregator expands aggregate assignments into field accesses, downcasts and discriminant assignments. However, this creates needlessly complicated place projections that may limit later optimization passes.

This PR attempts to replace those accesses by new locals that can be optimized more easily, for instance by DestProp.

PR #88628, opened 2021-09-03
Scalar replacement of enums

cjgillot:
Extension to https://github.com/rust-lang/rust/pull/85796

r? @tmiasko

### DIFF
--- a/compiler/rustc_mir/src/transform/flatten_locals.rs
+++ b/compiler/rustc_mir/src/transform/flatten_locals.rs
@@ -1,9 +1,10 @@
 use crate::transform::MirPass;
 use rustc_data_structures::fx::{FxHashMap, FxHashSet};
 use rustc_index::vec::{Idx, IndexVec};
+use rustc_middle::mir::interpret::Scalar;
 use rustc_middle::mir::visit::*;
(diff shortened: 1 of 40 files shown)


tmiasko:
Since this transformation expands fields which originally were stored in
overlapping memory location into disjoint memory locations, is it correct only
under the assumption that enums are used in type-safe manner.

Unsafe code might violate this assumption. While the transformation validates
that the address of a local escapes, I don't think this can be consider to be a
reliable method to detect unsafe code inside optimization pipeline (since those
constructs might be optimized to a direct access).

It also inconsistent with other hypothetical transformations.  For example,
consider set-discriminant of an enum using niche encoding that switches to
dataful variant. Such a set discriminant statement is a no-op (discriminant is
actually changed as a result of writing the variant data).  A transformation
could remove such set-discriminants, which would be inconsistent with
transformation proposed here.

For those reasons, I think this can't run in optimization phase, but would
rather have to run in an earlier lowering phase.

It also seems to me that we should first introduce support for expansion of
variable debuginfo. Without such support this transformation is rarely applicable.
Though, I am might be missing use cases that motivated this.

PR #88631, opened 2021-09-04
Improve structured tuple struct suggestion

camelid:
Previously, the span was just for the constructor name, which meant it
would result in syntactically-invalid code when applied. Now, the span
is for the entire expression.

I also changed it to use `span_suggestion_verbose`, for two reasons:

1. Now that the suggestion span has been corrected, the output is a bit
   cluttered and hard to read. Putting the suggestion its own window
   creates more space.

2. It's easier to see what's being suggested, since now the version
   after the suggestion is applied is shown.

r? @davidtwco

### DIFF
--- a/compiler/rustc_typeck/src/check/expr.rs
+++ b/compiler/rustc_typeck/src/check/expr.rs
@@ -1514,7 +1514,7 @@ impl<'a, 'tcx> FnCtxt<'a, 'tcx> {
                         ),
                     );
                     err.span_label(field.ident.span, "field does not exist");
-                    err.span_suggestion(
+                    err.span_suggestion_verbose(
                         expr_span,
                         &format!(
                             "`{adt}::{variant}` is a tuple {kind_name}, use the appropriate syntax",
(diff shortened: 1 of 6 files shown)


davidtwco (approved):
LGTM

rylev:
The waiting on review label was added accidentally, since we're now posting perf results after [merges happen](https://github.com/rust-lang/rustc-perf/pull/992). I'll remove the label. 

@rustbot label: -S-waiting-on-review 

The summary that this PR led to a mixed performance change seems correct though the change is quite small. We might want to tweak the perf bot to not post in such cases.

camelid:
The perf results still seem much worse than I would have thought. The only non-error path change is that an extra `Span` is passed as a function argument. I wonder if these results are somewhat spurious. For example, `clap-rs-opt` regressed by ~0.3%, while `clap-rs-debug` was unchanged—but this change isn't in the optimizer, so it shouldn't have an effect on one but not the other.

@rylev Should this be marked as `perf-regression-triaged` then?

rylev:
Yep - it looks spurious - all the perf changes except 2 are considered "very small" by our standards with the other two barely registering as "small" changes. Given that there doesn't seem to be a plausible explanation, and the perf change is mostly a wash, we should marked this as triaged.

PR #88632, opened 2021-09-04
Fix issues with Markdown summary options

camelid:
- Use `summary_opts()` for Markdown summaries
- Enable all main body Markdown options for summaries

### DIFF
--- a/src/librustdoc/html/markdown.rs
+++ b/src/librustdoc/html/markdown.rs
@@ -48,7 +48,7 @@ use pulldown_cmark::{
 mod tests;
 
 /// Options for rendering Markdown in the main body of documentation.
-pub(crate) fn opts() -> Options {
+pub(crate) fn main_body_opts() -> Options {
     Options::ENABLE_TABLES
         | Options::ENABLE_FOOTNOTES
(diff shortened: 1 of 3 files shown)


camelid:
(There are more details about the changes in the commit messages.)

CraftSpider:
Does a tasklist actually look okay in a markdown summary? I guess it doesn't necessarily matter, not rendering it probably looks worse, but I'm curious what kind of affects this has.

camelid:
> Does a tasklist actually look okay in a markdown summary? I guess it doesn't necessarily matter, not rendering it probably looks worse, but I'm curious what kind of affects this has.

It doesn't actually get rendered, because `markdown_summary_with_limit` stops as soon as it encounters things like task lists. I've tested it locally, though I don't have a screenshot available since it was a few days ago, but the difference is like this:

#### input

```md
- [ ] foo
- [ ] bar
```

#### output before

(note that this is shown literally as the characters `- [ ] foo`)

```
- [ ] foo
```

#### output after

(nothing)

CraftSpider:
Alright, thanks for the clarification. Definitely sounds like better behavior now.
@bors r+

PR #88639, opened 2021-09-04
rustdoc: Fix ICE with `doc(hidden)` on tuple variant fields

### LINKED ISSUE #88600: rustdoc ICE: entered unreachable code on hidden tuple item
### Code

```Rust
pub enum Foo {
    Test(#[doc(hidden)] Option<()>),
}
```

panics when running `cargo doc`. This seems to come from https://github.com/rust-lang/rust/blob/ed30993456224998db3a8f347a286f8234cd75c8/src/librustdoc/html/render/print_item.rs#L943

cc #88391

### Meta

`rustc --version --verbose`:
```
binary: rustc
commit-date: 2021-09-01
host: x86_64-pc-windows-msvc
release: 1.56.0-nightly
LLVM version: 13.0.0
```

### Error output

```
cargo doc --no-deps      
 Documenting hidden v0.1.0-rc.1 (G:\workspace\small dev space\test_crate-version)
thread 'rustc' panicked at 'internal error: entered unreachable code', src\librustdoc\html\render\print_item.rs:943:80

error: could not document `hidden`

Caused by:
  process didn't exit successfully: `rustdoc --edition=2018 --crate-type lib --crate-name hidden src\lib.rs -o "G:\workspace\small dev space\test_crate-version\target\doc" --error-format=json --json=diagnostic-rendered-ansi -L "dependency=G:\workspace\small dev space\test_crate-version\target\debug\deps" --crate-version 0.1.0-rc.1` (exit code: 1)
```

<p>

```
thread 'rustc' panicked at 'internal error: entered unreachable code', src\librustdoc\html\render\print_item.rs:943:80
stack backtrace:
   0:     0x7ff80a64801e - <std::sys_common::backtrace::_print::DisplayBacktrace as core::fmt::Display>::fmt::h97844c02df23ebf1
   1:     0x7ff80a6728da - core::fmt::write::h7e004fee938dc3bd
   2:     0x7ff80a63b388 - <std::io::IoSlice as core::fmt::Debug>::fmt::h34e482c524fdd2ee
   3:     0x7ff80a64bab6 - std::panicking::take_hook::hd051283b0b015437
   4:     0x7ff80a64b5c8 - std::panicking::take_hook::hd051283b0b015437
   5:     0x7ff80390ad95 - <md5::Md5 as std::io::Write>::flush::h7054a9d890538508
   6:     0x7ff80a64c3b0 - std::panicking::rust_panic_with_hook::h1dfae7c828339eef
   7:     0x7ff80a64be3f - rust_begin_unwind
   8:     0x7ff80a648947 - <std::sys_common::backtrace::_print::DisplayBacktrace as core::fmt::Display>::fmt::h97844c02df23ebf1
   9:     0x7ff80a64bdc9 - rust_begin_unwind
  10:     0x7ff80a6a8980 - core::panicking::panic_fmt::hec18c52ab5ec8c2e
  11:     0x7ff80a6a88cc - core::panicking::panic::h5cccc34aab9ab48d
  12:     0x7ff707957d6c - <unknown>
  13:     0x7ff70794a0b7 - <unknown>
  14:     0x7ff70793c43f - <unknown>
  15:     0x7ff707947276 - <unknown>
  16:     0x7ff707a41406 - <unknown>
  17:     0x7ff70785ede5 - <unknown>
  18:     0x7ff70797661a - <unknown>
  19:     0x7ff70789aab3 - <unknown>
  20:     0x7ff70796e8f4 - <unknown>
  21:     0x7ff70789cf3a - <unknown>
  22:     0x7ff70789363a - <unknown>
  23:     0x7ff7079e6296 - <unknown>
  24:     0x7ff707978de8 - <unknown>
  25:     0x7ff70791d68d - <unknown>
  26:     0x7ff80a65a64c - std::sys::windows::thread::Thread::new::h4e02b30388f9fde7
  27:     0x7ff8dec054e0 - BaseThreadInitThunk
  28:     0x7ff8dfa2485b - RtlUserThreadStart
```

</p>
</details>

Emilgardis:
Fixes #88600.

```rust
pub struct H;
pub struct S;

pub enum FooEnum {
    HiddenTupleItem(#[doc(hidden)] H),
    MultipleHidden(#[doc(hidden)] H, #[doc(hidden)] H),
    MixedHiddenFirst(#[doc(hidden)] H, S),
    MixedHiddenLast(S, #[doc(hidden)] H),
    HiddenStruct {
        #[doc(hidden)]
        h: H,
        s: S,
    },
}
```

Generates
image

### DIFF
--- a/src/librustdoc/html/render/print_item.rs
+++ b/src/librustdoc/html/render/print_item.rs
@@ -942,15 +942,15 @@ fn item_union(w: &mut Buffer, cx: &Context<'_>, it: &clean::Item, s: &clean::Uni
 }
 
 fn print_tuple_struct_fields(w: &mut Buffer, cx: &Context<'_>, s: &[clean::Item]) {
-    for (i, ty) in s
-        .iter()
-        .map(|f| if let clean::StructFieldItem(ref ty) = *f.kind { ty } else { unreachable!() })
-        .enumerate()
-    {
+    for (i, ty) in s.iter().enumerate() {
         if i > 0 {
             w.write_str(",&nbsp;");
         }
-        write!(w, "{}", ty.print(cx));
+        match *ty.kind {
+            clean::StrippedItem(box clean::StructFieldItem(_)) => w.write_str("_"),
+            clean::StructFieldItem(ref ty) => write!(w, "{}", ty.print(cx)),
+            _ => unreachable!(),
+        }
     }
 }
 
@@ -1066,24 +1066,27 @@ fn item_enum(w: &mut Buffer, cx: &Context<'_>, it: &clean::Item, e: &clean::Enum
                     name = variant.name.as_ref().unwrap(),
                 );
                 for field in fields {
-                    use crate::clean::StructFieldItem;
-                    if let StructFieldItem(ref ty) = *field.kind {
-                        let id = cx.derive_id(format!(
-                            "variant.{}.field.{}",
-                            variant.name.as_ref().unwrap(),
-                            field.name.as_ref().unwrap()
-                        ));
-                        write!(
-                            w,
-                            "<span id=\"{id}\" class=\"variant small-section-header\">\
-                                 <a href=\"#{id}\" class=\"anchor field\"></a>\
-                                 <code>{f}:&nbsp;{t}</code>\
-                             </span>",
-                            id = id,
-                            f = field.name.as_ref().unwrap(),
-                            t = ty.print(cx)
-                        );
-                        document(w, cx, field, Some(variant));
+                    match *field.kind {
+                        clean::StrippedItem(box clean::StructFieldItem(_)) => {}
+                        clean::StructFieldItem(ref ty) => {
+                            let id = cx.derive_id(format!(
+                                "variant.{}.field.{}",
+                                variant.name.as_ref().unwrap(),
+                                field.name.as_ref().unwrap()
+                            ));
+                            write!(
+                                w,
+                                "<span id=\"{id}\" class=\"variant small-section-header\">\
+                                    <a href=\"#{id}\" class=\"anchor field\"></a>\
+                                    <code>{f}:&nbsp;{t}</code>\
+                                </span>",
+                                id = id,
+                                f = field.name.as_ref().unwrap(),
+                                t = ty.print(cx)
+                            );
+                            document(w, cx, field, Some(variant));
+                        }
+                        _ => unreachable!(),
                     }
                 }
                 w.write_str("</div></div>");
--- a/src/test/rustdoc/issue-88600.rs
+++ b/src/test/rustdoc/issue-88600.rs
@@ -1,0 +1,34 @@
+// This test ensure that #[doc(hidden)] is applied correctly in enum variant fields.
+
+// Denotes a field which should be hidden.
+pub struct H;
+
+// Denotes a field which should not be hidden (shown).
+pub struct S;
+
+// @has issue_88600/enum.FooEnum.html
+pub enum FooEnum {
+    // @has - '//*[@id="variant.HiddenTupleItem"]//code' 'HiddenTupleItem(_)'
+    // @count - '//*[@id="variant.HiddenTupleItem.field.0"]' 0
+    HiddenTupleItem(#[doc(hidden)] H),
+    // @has - '//*[@id="variant.MultipleHidden"]//code' 'MultipleHidden(_, _)'
+    // @count - '//*[@id="variant.MultipleHidden.field.0"]' 0
+    // @count - '//*[@id="variant.MultipleHidden.field.1"]' 0
+    MultipleHidden(#[doc(hidden)] H, #[doc(hidden)] H),
+    // @has - '//*[@id="variant.MixedHiddenFirst"]//code' 'MixedHiddenFirst(_, S)'
+    // @count - '//*[@id="variant.MixedHiddenFirst.field.0"]' 0
+    // @has - '//*[@id="variant.MixedHiddenFirst.field.1"]' '1: S'
+    MixedHiddenFirst(#[doc(hidden)] H, S),
+    // @has - '//*[@id="variant.MixedHiddenLast"]//code' 'MixedHiddenLast(S, _)'
+    // @has - '//*[@id="variant.MixedHiddenLast.field.0"]' '0: S'
+    // @count - '//*[@id="variant.MixedHiddenLast.field.1"]' 0
+    MixedHiddenLast(S, #[doc(hidden)] H),
+    // @has - '//*[@id="variant.HiddenStruct"]//code' 'HiddenStruct'
+    // @count - '//*[@id="variant.HiddenStruct.field.h"]' 0
+    // @has - '//*[@id="variant.HiddenStruct.field.s"]' 's: S'
+    HiddenStruct {
+        #[doc(hidden)]
+        h: H,
+        s: S,
+    },
+}


Emilgardis:
Generates from

```rust
pub enum FooEnum {
    HiddenTupleItem(#[doc(hidden)] ()),
    MultipleHidden(#[doc(hidden)] (), #[doc(hidden)] ()),
    MixedHiddenTupleItemLast((), #[doc(hidden)] ()),
    MixedHiddenTupleItemFirst(#[doc(hidden)] (), ()),
    NotHiddenTupleItem1(()),
    NotHiddenTupleItem2((), ()),
}
```

I don't like how the Variants are shown, preferably I'd have it as `0: /* omitted */` or something.


Generates
image

src/librustdoc/html/render/print_item.rs
```diff
@@ -944,13 +944,17 @@ fn item_union(w: &mut Buffer, cx: &Context<'_>, it: &clean::Item, s: &clean::Uni
 fn print_tuple_struct_fields(w: &mut Buffer, cx: &Context<'_>, s: &[clean::Item]) {
     for (i, ty) in s
         .iter()
-        .map(|f| if let clean::StructFieldItem(ref ty) = *f.kind { ty } else { unreachable!() })
+        .map(|f| if let clean::StructFieldItem(ref ty) = *f.kind { Some(ty) } else { None })
```
GuillaumeGomez, line 947: It should be possible at this stage, it means that the `clean::Item` was badly created... Can you find out where a non-`StructFieldItem` was added please?

GuillaumeGomez:
Also, please add a regression test to ensure that this doesn't panic anymore in the future.

src/librustdoc/html/render/print_item.rs
Emilgardis: Do you mean it shouldn't be possible to create a `clean::Item` with `StructFieldItem` kind?

GuillaumeGomez: I meant that a tuple item (when creating a new one) should only have `StructFieldItem` and discard everything else. (In `clean/mod.rs` iirc, sorry, not on computer at the moment)

Emilgardis: The kind we end up with is

```
StrippedItem(StructFieldItem(Tuple([])))
```

my brief look into how the render works suggests to me that this is correct, and that it just needs to be handled.

Emilgardis:
I'm trying to figure out how I'd do HTML checking with `@!has` or `@!matches`. What absence should be checked?

would

```rust
// @!has issue_88600/enum.FooEnum.html '//*[@id="variant.HiddenTupleItem.fields.0"]'
```

etc. be enough?

GuillaumeGomez:
Also check that the other fields are present, but otherwise seems good.

src/librustdoc/html/render/print_item.rs
GuillaumeGomez: Isn't there the same issue with:

```rust
pub enum Foo {
    Bar {
        x: u32,
        #[doc(hidden)]
        y: u8,
    }
}
```

Emilgardis:
Fixed :D

Seems like

```rust
// @!has issue_88600/enum.FooEnum.html '//*[@id="variant.HiddenTupleItem.fields.0"]'
```

doesn't actually work, need to specify a pattern to not match.

src/librustdoc/html/render/print_item.rs
Emilgardis: That does not panic

image

```rust
pub struct Foo(#[doc(hidden)] ());
```
does not panic either, and outputs 

image

Which gives me the idea to do the same `_` for stripped/hidden tuple field

Emilgardis:
looks like this now:

image

jyn514:
I don't think this is a good idea. Enum fields are *always* public, they shouldn't be hidden from documentation. This should only fix the ICE, but not change the behavior (and actually I would suggest warning that doc(hidden) does nothing in that position as well).

jyn514:
Also, @GuillaumeGomez please remember to check with the team before making changes in behavior like this, this is a feature we'd have to support into the future.

Emilgardis:
My previous CI run from where I encountered generated this output:

```sh
$ cargo +nightly-2021-08-21 -V
cargo 1.56.0-nightly (e96bdb0c3 2021-08-17)
```

image

so this particular change/pr is not really a change stable->nightly.

jyn514:
Well that seems like a misfeature ... But I guess if it already existed that's fine.

Have you tested this respects --document-hidden-items?

Emilgardis:
> 
> 
> Well that seems like a misfeature ... But I guess if it already existed that's fine.
> 
> Have you tested this respects --document-hidden-items?

That works just fine, I think that's because `strip-hidden` is not ran so no `StrippedItem`s are made with that option.

GuillaumeGomez:
> Also, @GuillaumeGomez please remember to check with the team before making changes in behavior like this, this is a feature we'd have to support into the future.

My bad, I thought it was something already supported. What should we do then? Opening an FCP or just discard it and do what I suggested [here](https://github.com/rust-lang/rust/pull/88639#discussion_r702274927)?

jyn514:
@GuillaumeGomez oh no, I was just confused - I thought this was a no-op before.

GuillaumeGomez:
Perfect, back to reviewing then. ;)

src/test/rustdoc/issue-88600.rs
```diff
@@ -0,0 +1,18 @@
...
+// @!has - '//*[@id="variant.MultipleHidden.field.0"]' '0: ()'
+// @!has - '//*[@id="variant.MultipleHidden.field.1"]' '1: ()'
+// @has - '//*[@id="variant.MixedHiddenFirst"]//code' 'MixedHiddenFirst(_, ())'
+// @!has - '//*[@id="variant.MixedHiddenFirst.field.0"]' '0: ()'
+// @has - '//*[@id="variant.MixedHiddenFirst.field.1"]' '1: ()'
+// @has - '//*[@id="variant.MixedHiddenLast"]//code' 'MixedHiddenLast((), _)'
+// @has - '//*[@id="variant.MixedHiddenLast.field.0"]' '0: ()'
+// @!has - '//*[@id="variant.MixedHiddenLast.field.1"]' '1: ()'
+pub enum FooEnum {
+    HiddenTupleItem(#[doc(hidden)] ()),
```
GuillaumeGomez, line 14: Sorry for the nitpicking here but: could you use different types (between hidden and non-hidden fields) so that it's much easier to see what's going on please?

GuillaumeGomez (approved):

GuillaumeGomez:
Looks good to me, thanks! r=me once CI pass.

camelid:
It actually looks like rustdoc's behavior for this changed between stable and 2021-08-21. I don't think that change was intentional, so I'm wondering if we should revert the behavior change in this PR before it hits stable?

camelid:
It looks likely that the behavior changed in #87451.

camelid:
searched nightlies: from nightly-2021-07-01 to nightly-2021-08-21
regressed nightly: nightly-2021-07-30
searched commits: from https://github.com/rust-lang/rust/commit/b70888601af92f6cdc0364abab3446e418b91d36 to https://github.com/rust-lang/rust/commit/492723897e9b4db6701b3a75b72618d08a7d5319
regressed commit: https://github.com/rust-lang/rust/commit/4533be947b011c9bde4871ae93e6553b24cc5ec6

<details>
<summary>bisected with <a href='https://github.com/rust-lang/cargo-bisect-rustc'>cargo-bisect-rustc</a> v0.6.0</summary>


Host triple: x86_64-apple-darwin
Reproduce with:
```bash
cargo bisect-rustc --start=2021-07-01 --end=2021-08-21 --preserve --prompt -- doc --open 
```
</details>

camelid:
### stable

image

### nightly-2021-07-30 (the date that the tuple field documentation PR landed)

image

### current nightly

rustdoc panics

jyn514:
:+1: to reverting back to the original behavior, but I also think it would be fine to land this PR as is since it fixes the ICE. If we do revert, we should add a warning that `doc(hidden)` does nothing for enum fields.

Emilgardis:
```diff
diff --git a/src/librustdoc/html/render/print_item.rs b/src/librustdoc/html/render/print_item.rs
index f6fd0f7a467..7227d816b9d 100644
--- a/src/librustdoc/html/render/print_item.rs
+++ b/src/librustdoc/html/render/print_item.rs
@@ -947,8 +947,8 @@ fn print_tuple_struct_fields(w: &mut Buffer, cx: &Context<'_>, s: &[clean::Item]
             w.write_str(",&nbsp;");
         }
         match *ty.kind {
-            clean::StrippedItem(box clean::StructFieldItem(..)) => w.write_str("_"),
-            clean::StructFieldItem(ref ty) => write!(w, "{}", ty.print(cx)),
+            clean::StrippedItem(box clean::StructFieldItem(ref ty))
+            | clean::StructFieldItem(ref ty) => write!(w, "{}", ty.print(cx)),
             _ => unreachable!(),
         }
     }
@@ -1066,24 +1066,27 @@ fn item_enum(w: &mut Buffer, cx: &Context<'_>, it: &clean::Item, e: &clean::Enum
                     name = variant.name.as_ref().unwrap(),
                 );
                 for field in fields {
-                    use crate::clean::StructFieldItem;
-                    if let StructFieldItem(ref ty) = *field.kind {
-                        let id = cx.derive_id(format!(
-                            "variant.{}.field.{}",
-                            variant.name.as_ref().unwrap(),
-                            field.name.as_ref().unwrap()
-                        ));
-                        write!(
-                            w,
-                            "<span id=\"{id}\" class=\"variant small-section-header\">\
-                                 <a href=\"#{id}\" class=\"anchor field\"></a>\
-                                 <code>{f}:&nbsp;{t}</code>\
-                             </span>",
-                            id = id,
-                            f = field.name.as_ref().unwrap(),
-                            t = ty.print(cx)
-                        );
-                        document(w, cx, field, Some(variant));
+                    match *field.kind {
+                        clean::StrippedItem(box clean::StructFieldItem(ref ty))
+                        | clean::StructFieldItem(ref ty) => {
+                            let id = cx.derive_id(format!(
+                                "variant.{}.field.{}",
+                                variant.name.as_ref().unwrap(),
+                                field.name.as_ref().unwrap()
+                            ));
+                            write!(
+                                w,
+                                "<span id=\"{id}\" class=\"variant small-section-header\">\
+                                <a href=\"#{id}\" class=\"anchor field\"></a>\
+                                <code>{f}:&nbsp;{t}</code>\
+                                </span>",
+                                id = id,
+                                f = field.name.as_ref().unwrap(),
+                                t = ty.print(cx)
+                            );
+                            document(w, cx, field, Some(variant));
+                        }
+                        _ => unreachable!(),
                     }
                 }
                 w.write_str("</div></div>");
```

applied on this will revert the behaviour and adapt it to the new fields output.

Should I add a warning in this pr? How/where would I do that?

jyn514:
> Should I add a warning in this pr? How/where would I do that?

Yes, please. It would go somewhere in rustc_attr I think - there should be an existing pass using `INVALID_DOC_ATTRIBUTES`, look for that.

camelid:
Beta-nominating because this fixes an ICE and soon will revert an accidental behavior change.

cc @rust-lang/rustdoc

Emilgardis:
shouldn't the lint be under the `rustdoc` tool?

GuillaumeGomez:
> shouldn't the lint be under the `rustdoc` tool?

It definitely should.

camelid:
Most of our `#[doc]` attribute lints are in `compiler/rustc_passes/src/check_attr.rs`, so as Joshua said, I think this one should go there too.

GuillaumeGomez:
Yes sorry, didn't check the previous messages... I meant where the other rustdoc attr checks are...

Emilgardis:
I can't figure out a way to do the check in rustc_passes. I'll push what I have right now.

What is missing is an ui test and to determine if it should be rustdoc or not, can imagine the pass would be quite different if it's supposed to live in rustc_passes. Feel free to edit the branch, I don't feel confident enough in the code base.

camelid:
> I can't figure out a way to do the check in rustc_passes. I'll push what I have right now.

What Joshua and I were suggesting is to add a `match` arm right around here: https://github.com/rust-lang/rust/blob/e30b68353fe22b00f40d021e7914eeb78473b3c1/compiler/rustc_passes/src/check_attr.rs#L904

Create a new function near the function I just linked to, and name it something like `check_doc_hidden`. Then, in the `match` arm, call that function in the same style as the nearby `match` arms.

Give the function a self parameter and a `hir_id: HirId` parameter, and within the function, add some code like this:

```rust
let def_id = self.tcx().hir().local_def_id(hir_id);
let def_kind = self.tcx().def_kind(def_id);
let parent_def_id = self.tcx().hir().get_parent_did(hir_id);
let parent_def_kind = self.tcx().def_kind(parent_def_id);
match (parent_def_kind, def_kind) {
    (DefKind::Ctor(CtorOf::Variant, _), DefKind::Field) => {
        /* emit INVALID_DOC_ATTRIBUTES lint (see surrounding code for examples
           of how to do that), and return `false` (meaning that the code is invalid) */
    _ => true,
}
```

You may need to tweak it slightly, but that should do the trick. Post a comment here or in `#t-compiler/help` on Zulip if you have questions.

Emilgardis:
Thank you, I did not understand how to get the parents. Ill fix this later.

Manishearth:
Yeah I think this should be backported

Emilgardis:
Question, should this also apply to the following?

```rust
pub enum Foo {
	Bar {
		#[doc(hidden)]
		a: u32,
	}
}
```

Currently that renders as

image

I see no real difference between the two. Why can't enum tuple fields be hidden, but enum struct fields can be?

jyn514:
> Beta-nominating for backport to 1.56 because this fixes an ICE and soon will revert an accidental behavior change.

I think it's premature to backport this before it lands in nightly. It sounds like *what* it does is still up in the air, let alone how.

> Currently that renders as

Currently, as in on nightly? Or as in on stable? I agree the behavior should match whatever tuple variants do.

Emilgardis:
My bad!

This is stable (1.54), beta (rustdoc 1.55.0-beta.10 (e72143457 2021-09-04)) and this branch

Emilgardis:
Example of usage: https://doc.rust-lang.org/nightly/alloc/collections/enum.TryReserveErrorKind.html

camelid:
> I think it's premature to backport this before it lands in nightly.

We wouldn't be backporting this until it lands in nightly; at least, that's what I was thinking.

jyn514:
> They all hide the field

Ok, then I think we should be consistent between tuple fields and struct variant fields. (I still think it's weird rustdoc allowed this at all, but since it already exists we should support it and have it be consistent.) That means the warning won't be necessary, sorry to waste your time.

Emilgardis:
> 
> 
> > They all hide the field
> 
> Ok, then I think we should be consistent between tuple fields and struct variant fields and hide both if they have `doc(hidden)`. (I still think it's weird rustdoc allowed this at all, but since it already exists we should support it and have it be consistent.) That means the warning won't be necessary, sorry to waste your time.

No worries! Ok, I'll fix the behaviour. I spent a good time trying to find out how to get the variant, but it's not wasted time, I now know more about the internals. Thanks for the help :)

So in summary.

* `#[doc(hidden)]` should apply to both fields in tuple variants and struct variants.
  * fields in tuple variants are hidden with `_`.
  * fields in struct variants are hidden like in normal structs. `/* (some) fields omitted */`
* These hidden fields should be hidden from the summary.
* No warning should be emitted.

Ill add a doc-test to make sure the behaviour is consistent for both variants, I'll do this in a test called `issue-88600.rs`

Emilgardis:
image

This is how it looks now

Emilgardis:
I'd want to make sure the <code> block is checked as well, but I'm not sure how I would do that

src/librustdoc/html/render/print_item.rs
```diff
@@ -1066,24 +1066,27 @@ fn item_enum(w: &mut Buffer, cx: &Context<'_>, it: &clean::Item, e: &clean::Enum
...
+                            let id = cx.derive_id(format!(
+                                "variant.{}.field.{}",
+                                variant.name.as_ref().unwrap(),
+                                field.name.as_ref().unwrap()
+                            ));
+                            write!(
+                                w,
+                                "<span id=\"{id}\" class=\"variant small-section-header\">\
+                                <a href=\"#{id}\" class=\"anchor field\"></a>\
+                                <code>{f}:&nbsp;{t}</code>\
```
Emilgardis, line 1081: Is this ok? Rustfmt didn't preserve the indentation

src/test/rustdoc/issue-88600.rs
```diff
@@ -0,0 +1,36 @@
+// This test ensure that #[doc(hidden)] is applied correctly to enums.
+// Fields in tuple variants can not be hidden (and also emit a warning).
+// Fields in struct variants can be hidden.
```
Emilgardis, line 3: Leftover

GuillaumeGomez (approved):

GuillaumeGomez:
Thanks!

@bors: r+

camelid:
Accepted for backport to 1.56 beta.

@rustbot label: +beta-accepted

Issue #88015, opened 2021-08-13
Locals aren't introduced properly in if-let guards on or-patterns


LeSeulArtichaut:
I tried this code:

```rust
#![feature(if_let_guard)]

fn main() {
    match () {
        () | () if let x = 0 => {
            x;
        },
        _ => {}
    };
}
```

I expected to see this happen: code compiles successfully

Instead, this happened: the compiler complains about `x` being possibly uninitialized
```text
error[E0381]: use of possibly-uninitialized variable: `x`
 --> src/main.rs:5:34
  |
5 |         () | () if let x = () => x,
  |                                  ^ use of possibly-uninitialized `x`
```

This is because each candidate subpattern introduces a different local for `x`.

cc tracking issue #51114.

### Meta

`rustc --version --verbose`:
```
rustc 1.56.0-nightly (ccffcafd5 2021-08-11)
binary: rustc
commit-hash: ccffcafd55e58f769d4b0efc0064bf65e76998e4
commit-date: 2021-08-11
host: x86_64-unknown-linux-gnu
release: 1.56.0-nightly
LLVM version: 12.0.1
```

osa1:
@rustbot claim

I started debugging this.

Error is generated by MIR borrow checker.

Repro I'm using:

```rust
#![feature(if_let_guard)]

pub enum Test {
    A,
    B,
}

pub fn test() -> bool {
    match Test::A {
        Test::A | Test::B if let x = false =>
            x,
        _ => true
    }
}
```

MIR after promote-consts: (I think this is the MIR used for borrow checking?)

<details>

<summary>MIR</summary>

```
// MIR for `test` after SimplifyCfg-promote-consts

fn test() -> bool {
    let mut _0: bool;                    // return place in scope 0 at test.rs:8:18: 8:22
    let mut _1: Test;                    // in scope 0 at test.rs:9:11: 9:18
    let mut _2: isize;                   // in scope 0 at test.rs:10:9: 10:16
    let mut _3: &Test;                   // in scope 0 at test.rs:9:11: 9:18
    let mut _4: bool;                    // in scope 0 at test.rs:10:38: 10:43
    let _5: bool;                        // in scope 0 at test.rs:10:34: 10:35
    let mut _6: bool;                    // in scope 0 at test.rs:10:38: 10:43
    let _7: bool;                        // in scope 0 at test.rs:10:34: 10:35
    scope 1 {
        debug x => _5;                   // in scope 1 at test.rs:10:34: 10:35
[...]

PR #88642, opened 2021-09-04
Formally implement let chains

c410-f3r:
## Let chains

Thanks to @Centril for creating the RFC and special thanks to @matthewjasper for helping me since the beginning of this journey. In fact, @matthewjasper did much of the complicated MIR stuff so it's true to say that this feature wouldn't be possible without him. Thanks again @matthewjasper!

With the changes proposed in this PR, it will be possible to chain let expressions along side local variable declarations or ordinary conditional expressions. In other words, do much of what the `if_chain` crate already does.

```rust
fn u2_beautiful_day(lyrics: &str) {
    let mut lines = lyrics.trim().lines();

    if let Some(see_the_world) = lines.next().map(|el| el.trim())
        && let Some(see_china) = lines.next().map(|el| el.trim())
        && let Some(see_the_canyons) = lines.next().map(|el| el.trim())
        && let Some(see_the_tuna_fleets) = lines.next().map(|el| el.trim())
        && let Some(see_the_bedouin_fires) = lines.next().map(|el| el.trim())
        && let Some(see_the_oil_fields) = lines.next().map(|el| el.trim())
        && let Some(and_see_the_bird) = lines.next().map(|el| el.trim())
        && let Some(after_the_flood) = lines.next().map(|el| el.trim())

        && see_the_world == "See the world in green and blue"
        && see_china == "See China right in front of you"
        && see_the_canyons == "See the canyons broken by cloud"
        && see_the_tuna_fleets == "See the tuna fleets clearing the sea out"
        && see_the_bedouin_fires == "See the bedouin fires at night"
        && see_the_oil_fields == "See the oil fields at first light"
        && and_see_the_bird == "And see the bird with a leaf in her mouth"
        && after_the_flood == "After the flood all the colors came out"
    {
        println!("It was a beautiful day, don't let it get away");
    }
}
```

## Other considerations

* `if let guard` and `let ... else` features need special care and should be handled in a following PR.

* Irrefutable patterns are allowed within a let chain context

* ~~Three Clippy lints were already converted to start dogfooding and help detect possible corner cases~~

cc #53667


src/tools/clippy/clippy_lints/src/collapsible_if.rs
```diff
@@ -118,53 +117,49 @@ fn block_starts_with_comment(cx: &EarlyContext<'_>, expr: &ast::Block) -> bool {
...
-                cx,
-                COLLAPSIBLE_ELSE_IF,
-                block.span,
-                "this `else { if .. }` block can be collapsed",
-                "collapse nested if block",
-                snippet_block_with_applicability(cx, else_.span, "..", Some(block.span), &mut applicability).into_owned(),
-                applicability,
-            );
-        }
+    if let ast::ExprKind::Block(ref block, _) = else_.kind
```
flip1995, line 120: Clippy is a huge pile of those `if_chain`s. Do you have a reason why you chose to swap out those two `if_chain` macros? Just PoC? If it isn't necessary for some technical reason, I'd like to avoid touching Clippy in this PR.

c410-f3r: > Three Clippy lints were already converted to start dogfooding and help detect possible corner cases

There was more converted lints but they disappeared for some reason.
Well, if the Clippy team is not comfortable yet with this feature, then I won't touch Clippy code.

flip1995: It's not really, that I'm not comfortable with the feature (I didn't look at your code), I just want to prevent that we mix `if_chain!`s and `let_chains`. I'd be happy to completely move Clippy to `let_chains`. But I definitely wouldn't do this in this PR, since it would explode it. Let's focus this PR on the feature implementation and not trying to fix Clippy, as long as it isn't broken :+1:

c410-f3r: > I just want to prevent that we mix if_chain!s and let_chains. 

Looks like a problem, indeed

>  I'd be happy to completely move Clippy to let_chains

Looking forward to seeing this day come true :)

JohnCSimon:
ping from triage:
@c410-f3r 
thanks.

BubbaSheen:
Coding looks extraordinary to say the least keep up the executive chain
working towards fundamental exposure.

Aaron1011:
What's the status of this PR?

c410-f3r:
Chains with the `&&` operator are working as expected but chains with the `||` operator still need to be figured out.

Believe it or not, I look into this PR quite regularly trying to solve stuff but I guess I am not good enough :)

Aaron1011:
@c410-f3r: I'd be happy to work with you on Zulip if you'd like some help

Patrick-Poitras:
@c410-f3r This PR actually closes #90722 and closes #92145, and fixes one of the examples in issue #88015

c410-f3r:
@Patrick-Poitras 

Thanks! Just added  #90722 and #92145 use-cases.
#88015 is more related to let guards though

@matthewjasper 

Can you take a look at this PR based on what we previously discussed?

TennyZhuang:
The feature is extremely useful and the PR looks very well, is there any update to push it merged?

c410-f3r:
This MVP should be good to merge after review (I guess)

src/test/mir-opt/let-chains.rs
```diff
@@ -0,0 +1,94 @@
...
+
+impl<'a, T> Drop for DropLogger<'a, T> {
+    fn drop(&mut self) {
+        self.log.0.borrow_mut().push(self.id);
+    }
+}
+
+struct InjectedFailure;
+
+// EMIT_MIR let_chains.main.ElaborateDrops.before.mir
```
matthewjasper, line 25: Why is this a MIR opt test? It has far too much code to have people able to check the output MIR.

compiler/rustc_mir_build/src/thir/pattern/check_match.rs
```diff
@@ -798,12 +806,30 @@ fn let_source(tcx: TyCtxt<'_>, pat_id: HirId) -> LetSource {
...
+
+    if let hir::Node::Expr(hir::Expr { kind: hir::ExprKind::If(..), .. }) = parent_parent_node {
+        return LetSource::IfLet;
+    }
+
+    LetSource::GenericLet
+}
+
+fn is_let_chain(tcx: TyCtxt<'_>, pat_id: HirId) -> bool {
+    let hir = tcx.hir();
```
matthewjasper, line 820: I'm not sure that this works for let chains with more than 1 `&&`. It might be simpler to check for the case this is not a let chain.

compiler/rustc_ast_lowering/src/expr.rs
```diff
@@ -393,13 +393,23 @@ impl<'hir> LoweringContext<'_, 'hir> {
...
+                self.mark_span_with_reason(DesugaringKind::CondTemporary, cond.span, None);
+            self.expr_drop_temps(span_block, cond, AttrVec::new())
+        };
         match cond.kind {
-            hir::ExprKind::Let(..) => cond,
-            _ => {
-                let span_block =
-                    self.mark_span_with_reason(DesugaringKind::CondTemporary, cond.span, None);
-                self.expr_drop_temps(span_block, cond, AttrVec::new())
+            hir::ExprKind::Binary(_, lhs, rhs) => {
```
matthewjasper, line 402: This check should be recursive: it should check if either side is an let chain rather than simply a let expression. I think that something like `if let x= .. && let y = .. && cond` won't work currently.

matthewjasper (commented):
Can you add a test for this:

```rs
fn or_chain() {
    let z;
    if false || { z = 3; false} || z == 3 {}
}

fn and_chain() {
    let z;
    if true && { z = 3; true} && z == 3 {}
}


fn and_chain_2() {
    let z;
    true && { z = 3; true} && z == 3;
}
```

both and chains should either both compile now or both still fail to compile

src/test/ui/rfc-2497-if-let-chains/irrefutable-lets.rs
```diff
@@ -0,0 +1,23 @@
+// check-pass
+
+#![feature(let_chains)]
+
+use std::ops::Range;
```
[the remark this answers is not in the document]
c410-f3r: Added these use-cases in `chains-without-let.rs`

src/test/ui/rfc-2497-if-let-chains/irrefutable-lets.rs
```diff
@@ -0,0 +1,23 @@
...
+
+fn main() {
+    let opt = Some(1..1);
+
+    if let another = &opt && let Some(_) = another {
```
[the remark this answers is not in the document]
c410-f3r: Added

src/test/ui/rfc-2497-if-let-chains/u2-beautiful-day.rs
```diff
@@ -0,0 +1,31 @@
...
+    if let Some(see_the_world) = lines.next().map(|el| el.trim())
+        && let Some(see_china) = lines.next().map(|el| el.trim())
+        && let Some(see_the_canyons) = lines.next().map(|el| el.trim())
+        && let Some(see_the_tuna_fleets) = lines.next().map(|el| el.trim())
+        && let Some(see_the_bedouin_fires) = lines.next().map(|el| el.trim())
+        && let Some(see_the_oil_fields) = lines.next().map(|el| el.trim())
+        && let Some(and_see_the_bird) = lines.next().map(|el| el.trim())
+        && let Some(after_the_flood) = lines.next().map(|el| el.trim())
+
+        && see_the_world == "See the world in green and blue"
```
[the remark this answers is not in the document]
c410-f3r: Well, at least I tried :)

Removed

src/test/mir-opt/let-chains.rs
c410-f3r: Not sure why but I removed anyway. The similar UI test is still present

cc @Aaron1011

compiler/rustc_mir_build/src/thir/pattern/check_match.rs
c410-f3r: It is working now for chains with more than one `&&` -> https://github.com/rust-lang/rust/pull/88642/files#diff-af0d549a94f69a8f6068e01f3df0cfad0bdcc2868bda71c416df6294abb241e6R820-R822

compiler/rustc_ast_lowering/src/expr.rs
c410-f3r: I will add recursion but `if let x= .. && let y = .. && cond` is working with the current logic -> https://github.com/rust-lang/rust/pull/88642/files#diff-c613ee6b091cf39afd840d0f4350a6a0c819b193ecde798c1ab8c5312810c644

Can you provide another snippet?

c410-f3r:
> Can you add a test for this:
> 
> ```rust
> fn or_chain() {
>     let z;
>     if false || { z = 3; false} || z == 3 {}
> }
> 
> fn and_chain() {
>     let z;
>     if true && { z = 3; true} && z == 3 {}
> }
> 
> 
> fn and_chain_2() {
>     let z;
>     true && { z = 3; true} && z == 3;
> }
> ```
> 
> both and chains should either both compile now or both still fail to compile

Yes, the `&&` chains are working but `||` is not as expected -> https://github.com/rust-lang/rust/pull/88642/files#diff-3b32a2ebb306c90cd5084db1b851ec6fa4f35f2825e29a5716550af380d8b8fc

compiler/rustc_ast_lowering/src/expr.rs
matthewjasper: The body needs to use `y`

c410-f3r: Oh, I see.

Just fixed the issue, thanks

src/test/ui/rfc-2497-if-let-chains/chains-without-let.rs
```diff
@@ -0,0 +1,18 @@
+fn and_chain() {
+    let z;
+    if true && { z = 3; true} && z == 3 {}
```
matthewjasper, line 3: This is the relevant test

matthewjasper (commented):
cc @rust-lang/lang (does this RFC have an owner on the lang team?)

This PR removes some branches from the MIR for chains of `&&`, allowing some more code to compile (see the test I've highlighted). Previously, the CFG included the possibility of evaluating only the first and third condition here, but this wouldn't work for if let chains, and so the behaviour is changed here even though the chain has no let expressions. This was chosen to minimise the change in behaviour that adding a let expression to a condition makes, but could be reverted. Making `||` consistent here will be done in a follow up PR.

src/test/ui/rfc-2497-if-let-chains/chains-without-let.rs
scottmcm: For others, [here](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=ef19b5299c8a9c078b063081ee7531ce)'s what this does on stable today:
```
error[E0381]: use of possibly-uninitialized variable: `z`
 --> src/lib.rs:3:34
  |
3 |     if true && { z = 3; true} && z == 3 {}
  |                                  ^ use of possibly-uninitialized `z`
```

I do like that this desugars `&&` to separate `if`s -- it makes sense to me that way -- but I also feel like that should be feature-gated initially.  If not it needs a stabilization report and FCP, I think.

Mark-Simulacrum:
The lang team discussed this in its triage meeting today -- we would like to see the MIR change you point out, @matthewjasper, factored out into a separate PR that we can FCP, though folks in the meeting were generally on board with merging such a change.

As @scottmcm mentions, there may be a desire to feature-gate such a change initially. Still, it is best to do that independently of let-chains, as they aren't necessarily coupled. It *is* true that a common use of let chains is such that it requires the sequential evaluation order, but that's not the only use.

src/test/ui/rfc-2497-if-let-chains/chains-without-let.rs
scottmcm: Said otherwise, @c410-f3r, I'm wary of merging something that's [self-described](https://github.com/rust-lang/rust/pull/88642#issuecomment-1009861864) as an MVP when reverting it would be a breaking change for ungated code.

However you want to resolve that worry is fine, though -- maybe that's adding a gate, maybe that's restricting this change to things involving `let`s temporarily, maybe that's something else.  As I said, I do *want* this CFG change; it's just a process question.

compiler/rustc_mir_build/src/build/expr/into.rs
```diff
@@ -163,27 +153,30 @@ impl<'a, 'tcx> Builder<'a, 'tcx> {
...
                 //        | (true)
-                //  [shortcurcuit_block: dest = true]
+                // [shortcircuit_block: dest = true]
 
-                let (shortcircuit_block, mut else_block, join_block) = (
-                    this.cfg.start_new_block(),
-                    this.cfg.start_new_block(),
-                    this.cfg.start_new_block(),
-                );
+                let local_scope = this.local_scope();
```
matthewjasper, line 164: The changes from here to the end of the file can be reverted to address the lang team's comment. r=me with that done.

c410-f3r: Done

src/test/ui/rfc-2497-if-let-chains/chains-without-let.rs
c410-f3r: Thanks for the comments, @scottmcm

PR #88644, opened 2021-09-04
`AbstractConst` private fields

eopb:
Calls `subst` in `AbstractConst::root` when `Node` is `Leaf`.

r? @lcnr

### DIFF
--- a/compiler/rustc_trait_selection/src/traits/const_evaluatable.rs
+++ b/compiler/rustc_trait_selection/src/traits/const_evaluatable.rs
@@ -8,6 +8,7 @@
 //! In this case we try to build an abstract representation of this constant using
 //! `mir_abstract_const` which can then be checked for structural equality with other
 //! generic constants mentioned in the `caller_bounds` of the current environment.
+use crate::traits::ty::subst::GenericArg;
 use rustc_errors::ErrorReported;
 use rustc_hir::def::DefKind;
 use rustc_index::bit_set::BitSet;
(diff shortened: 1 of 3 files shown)


compiler/rustc_trait_selection/src/traits/const_evaluatable.rs
```diff
@@ -92,7 +91,6 @@ pub fn is_const_evaluatable<'cx, 'tcx>(
                         ControlFlow::CONTINUE
                     }
                     Node::Cast(_, _, ty) => {
-                        let ty = ty.subst(tcx, ct.substs);
```
BoxyUwU, line 95: I don't think this is correct, you aren't substing this in `AbstractConst::root` like you do for leafs

BoxyUwU (changes requested):

compiler/rustc_trait_selection/src/traits/const_evaluatable.rs
```diff
@@ -218,8 +217,12 @@ impl<'tcx> AbstractConst<'tcx> {
     }
 
     #[inline]
-    pub fn root(self) -> Node<'tcx> {
-        self.inner.last().copied().unwrap()
+    pub fn root(self, tcx: TyCtxt<'tcx>) -> Node<'tcx> {
+        let node = self.inner.last().copied().unwrap();
+        if let Node::Leaf(leaf) = node {
```
lcnr, line 222: I only realized this by looking at the code again, but things like `expr as *const T` have a parameter in the `Ty` of `Node::Cast`. So we should also call `subst` for these nodes :thinking: Wasn't able to trigger a bug with this as the I couldn't find a way to get `as SomethingWith<T>` into a generic const expr, but I still think we should "do the right thing" here.

Maybe make this an exhaustive match on `Node`, adding a comment like `// Can't directly contain generic params` to the branch for the kinds we can ignore

lcnr (commented):
found a potential future bug for `Node::Cast`, apart from that, r=me

thanks :heart:

compiler/rustc_trait_selection/src/traits/const_evaluatable.rs
gilescope: nit: Just wondering if we could do something like `let node = self.inner.last().as_ref()` to avoid the copy ?

lcnr: we could, but considering that `Node` is fairly small I don't expect this to matter wrt performance

lcnr (changes requested):
after this r=me

PR #88647, opened 2021-09-04
Document when to use Windows' `symlink_dir` vs. `symlink_file`

### LINKED ISSUE #88635: Better docs for std::os::windows::fs::{symlink_dir, symlink_file};
At the moment (rustc 1.56.0-nightly (b7404c898 2021-09-03)) the docs for this two functions are identical, and the only difference is file/dir. As a developer with primarily unix background, I don't really understand why there are two functions, and when I should use one versus the other. It would be helpful if the docs explained windows/unix difference, or linked upstream docs.

ChrisDenton:
It was previously unclear why there are two functions and when they should be used.

Fixes: #88635

### DIFF
--- a/library/std/src/os/windows/fs.rs
+++ b/library/std/src/os/windows/fs.rs
@@ -517,10 +517,19 @@ impl FileTypeExt for fs::FileType {
     }
 }
 
-/// Creates a new file symbolic link on the filesystem.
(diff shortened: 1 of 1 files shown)


ChrisDenton:
I wasn't sure about the phrase "non-directory file" but it's the best I could think of considering how overloaded the word "file" is.

PR #88648, opened 2021-09-04
Correct “copies” to “moves” in `<Option<T> as From<T>>::from` doc, and other copyediting

kpreid:
The `impl<T> From<T> for Option<T>` has no `Copy` or `Clone` bound, so its operation is guaranteed to be a move. The call site might copy, but the function itself cannot.

Since that would have been a rather small PR, I also reviewed the other documentation in the file and made other improvements (in separate commits): adding periods and commas, linking `Deref::Target`, and clarifying what "a container" is in `FromIterator`.


Mark-Simulacrum:
> Since that would have been a rather small PR, I also reviewed the other documentation in the file and made other improvements (in separate commits): adding periods and commas, linking Deref::Target, and clarifying what "a container" is in FromIterator.

FWIW, small PRs are great -- don't worry about adding more in the future.

@bors r+ rollup

PR #78004, opened 2020-10-16
[WIP] Skip single use lifetime lint for generated opaque types

sapessi:
Fix: #77175

The opaque type generated by the desugaring process of an async function uses the lifetimes defined by the originating function. The `DefId` for the lifetimes in the opaque type are different from the ones in the originating async function - as they should be, as far as I understand, and could therefore be considered a single use lifetimes, this causes the `single_use_lifetimes` lint to fail compilation if explicitly denied. This fix skips the lint for lifetimes used only once in generated opaque types for an async function that are declared in the parent async function definition.

More info in the comments on the original issue: [1](https://github.com/rust-lang/rust/issues/77175#issuecomment-707939141) and [2](https://github.com/rust-lang/rust/issues/77175#issuecomment-709569322)

PR #88650, opened 2021-09-04
Skip single use lifetime lint for generated opaque types 

### LINKED ISSUE #77175: False positive: single_use_lifetimes
I tried this code:

```rust
#[deny(single_use_lifetimes)]

// error: lifetime parameter `'a` only used once
async fn bar<'a>(s1: String, s2: &'_ str, s3: &'a str) -> &'a str {
    s3
}

fn foo<'a>(s1: String, s2: &'_ str, s3: &'a str) -> &'a str {
    s3
}

fn main() {}
```

I expected to see this happen: the code is correct.

Instead, this happened: 

```
error: lifetime parameter `'a` only used once
 --> src/main.rs:4:14
  |
4 | async fn bar<'a>(s1: String, s2: &'_ str, s3: &'a str) -> &'a str {
  |              ^^ this lifetime...                           -- ...is used only here
  |
 --> src/main.rs:1:8
  |
1 | #[deny(single_use_lifetimes)]
  |        ^^^^^^^^^^^^^^^^^^^^
```

Just swap the functions, then the code compiles:

```rust
#[deny(single_use_lifetimes)]

fn foo<'a>(s1: String, s2: &'_ str, s3: &'a str) -> &'a str {
    s3
}

async fn bar<'a>(s1: String, s2: &'_ str, s3: &'a str) -> &'a str {
    s3
}

fn main() {}
```

Update 20201014:

The examples may be misleading.
I forgot that `#[deny(single_use_lifetimes)]` takes effect on a single item.
It was a typo.

Here is a new example. It should be correct. But the compiler does not accept it.

```rust
#![deny(single_use_lifetimes)]

// error: lifetime parameter `'a` only used once
async fn bar<'a>(s1: String, s2: &'_ str, s3: &'a str) -> &'a str {
    s3
}
```

### Meta

`rustc --version --verbose`:
```
rustc 1.46.0 (04488afe3 2020-08-24)
binary: rustc
commit-hash: 04488afe34512aa4c33566eb16d8c912a3ae04f9
commit-date: 2020-08-24
release: 1.46.0
LLVM version: 10.0
```

```
rustc 1.48.0-nightly (8b4085359 2020-09-23)
binary: rustc
commit-hash: 8b4085359ae798dedb05c95ad42520557bd25320
commit-date: 2020-09-23
release: 1.48.0-nightly
LLVM version: 11.0
```

sapessi:
Fix: #77175

The opaque type generated by the desugaring process of an async function uses the lifetimes defined by the originating function. The DefId for the lifetimes in the opaque type are different from the ones in the originating async function - as they should be, as far as I understand, and could therefore be considered a single use lifetimes, this causes the single_use_lifetimes lint to fail compilation if explicitly denied. This fix skips the lint for lifetimes used only once in generated opaque types for an async function that are declared in the parent async function definition.

More info in the comments on the original issue: 1 and 2

### DIFF
--- a/compiler/rustc_resolve/src/late/lifetimes.rs
+++ b/compiler/rustc_resolve/src/late/lifetimes.rs
@@ -2024,7 +2024,7 @@ impl<'a, 'tcx> LifetimeContext<'a, 'tcx> {
         // ensure that we issue lints in a repeatable order
         def_ids.sort_by_cached_key(|&def_id| self.tcx.def_path_hash(def_id));
 
-        for def_id in def_ids {
+        'lifetimes: for def_id in def_ids {
             debug!("check_uses_for_lifetimes_defined_by_scope: def_id = {:?}", def_id);
 
             let lifetimeuseset = self.lifetime_uses.remove(&def_id);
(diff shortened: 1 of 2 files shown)


sapessi:
This is a copy of the original #78004 PR. The CR comments on using an `FxHashSet` for the generic properties is not yet addressed

estebank:
@bors try @rust-timer queue

Noticed that we're fetching the item unconditionally on every DefId to see if it might be an async fn. Running a perf check to see if it has performance implications.

sapessi:
Anything else we need to do here @estebank?

Issue #56940, opened 2018-12-18
ICE 'specified instant was less than self' and BorrowMutError on aarch64-linux-unknown-gnu


IntrepidPig:
When compiling code on an aarch64 system (a [pine a64](https://www.pine64.org/?product=pine-a64-lts)) I get a compiler panic 'specified instant was less than self', which I'm guessing is because the accuracy of the timer given on this system isn't very reliable and the `std::time::Instant` guarantee that a new one won't be earlier than an old one isn't upheld. This problem occurs somewhat reliably. Usually several crates will compile successfully in a long compile job before a panic happens. This problem has been happening to me since around 1.26 nightly.

Here's a backtrace I get
```
thread 'main' panicked at 'specified instant was later than self', libstd/sys/unix/time.rs:292:17
stack backtrace:                                                                                                
   0: std::sys::unix::backtrace::tracing::imp::unwind_backtrace                                                 
             at libstd/sys/unix/backtrace/tracing/gcc_s.rs:49                                                   
   1: std::sys_common::backtrace::print                                                                         
             at libstd/sys_common/backtrace.rs:71                                                               
             at libstd/sys_common/backtrace.rs:59                                                               
   2: std::panicking::default_hook::{{closure}}                                                                 
             at libstd/panicking.rs:211                                                                         
   3: std::panicking::default_hook                                                                              
             at libstd/panicking.rs:227                                                                         
   4: rustc::util::common::panic_hook                                                                           
   5: std::panicking::rust_panic_with_hook                                                                      
             at libstd/panicking.rs:480                                                                         
   6: std::panicking::begin_panic                                                                               
             at libstd/panicking.rs:410                                                                         
   7: std::sys::unix::time::inner::Instant::sub_instant::{{closure}}                                            
             at libstd/sys/unix/time.rs:292                                                                     
   8: std::time::Instant::elapsed                                                                               
             at libcore/result.rs:774                                                                           
             at libstd/sys/unix/time.rs:291                                                                     
             at libstd/time.rs:185                                                                              
             at libstd/time.rs:250                                                                              
             at libstd/time.rs:209                                                                              
   9: rustc::util::profiling::SelfProfiler::start_activity                                                      
  10: rustc::ty::query::plumbing::<impl rustc::ty::context::TyCtxt<'a, 'gcx, 'tcx>>::ensure_query               
[...]

Issue #49281, opened 2018-03-22
unexpected panic while compiling firefox on arm64


khumarahn:
Hi, I see this strange error while trying to compile firefox-59.0.1 with rust-1.23 on arm64 laptop running gentoo linux. It says, my bug report would be appreciated. I'd also appreciate any feedback.

I would be happy to provide any additional information.

```
   Compiling quote v0.3.15
   Compiling peeking_take_while v0.1.2
   Compiling glob v0.2.11
error: internal compiler error: unexpected panic

note: the compiler unexpectedly panicked. this is a bug.

note: we would appreciate a bug report: https://github.com/rust-lang/rust/blob/master/CONTRIBUTING.md#bug-reports

note: rustc 1.23.0-dev running on aarch64-unknown-linux-gnu

note: run with `RUST_BACKTRACE=1` for a backtrace

thread 'rustc' panicked at 'other was less than the current instant', src/libstd/sys/unix/time.rs:284:16
stack backtrace:
   0:       0x7f9e7de973 - rust_metadata_std_5958eddae63d6947e0733521fc5435ed
   1:       0x7f9e7da9d7 - rust_metadata_std_5958eddae63d6947e0733521fc5435ed
   2:       0x7f9e7e9f3f - rust_metadata_std_5958eddae63d6947e0733521fc5435ed
   3:       0x7f9e7e9c4f - rust_metadata_std_5958eddae63d6947e0733521fc5435ed
   4:       0x7f9e7ea443 - std::panicking::rust_panic_with_hook::h22b280fc4fc0de87
   5:       0x7f9e7ea303 - rust_metadata_std_5958eddae63d6947e0733521fc5435ed
   6:       0x7f9e7d9eb7 - std::time::Instant::elapsed::hf802fe319eded9b8
   7:       0x7f9da24283 - <unknown>
   8:       0x7f9da4965b - <unknown>
   9:       0x7f9cd3199b - rust_metadata_rustc_3ea30897d9dddf7b5fa3a6c2523637e7
  10:       0x7f9cf028f7 - rust_metadata_rustc_3ea30897d9dddf7b5fa3a6c2523637e7
  11:       0x7f9cc1cd2b - rust_metadata_rustc_3ea30897d9dddf7b5fa3a6c2523637e7
  12:       0x7f9d09b153 - <unknown>
  13:       0x7f9cd31a33 - rust_metadata_rustc_3ea30897d9dddf7b5fa3a6c2523637e7
[...]

PR #88652, opened 2021-09-04
linux/aarch64 Now() should be actually_monotonic()

AGSaidi:
While issues have been seen on arm64 platforms the Arm architecture requires
that the counter monotonically increases and that it must provide a uniform
view of system time (e.g. it must not be possible for a core to receive a
message from another core with a time stamp and observe time going backwards
(ARM DDI 0487G.b D11.1.2). While there have been a few 64bit SoCs that have
bugs (#49281, #56940) which cause time to not monotonically increase, these have
been fixed in the Linux kernel and we shouldn't penalize all Arm SoCs for those
who refuse to update their kernels:
SUN50I_ERRATUM_UNKNOWN1 - Allwinner A64 / Pine A64 - fixed in 5.1
FSL_ERRATUM_A008585 - Freescale LS2080A/LS1043A - fixed in 4.10
HISILICON_ERRATUM_161010101 - Hisilicon 1610 - fixed in 4.11
ARM64_ERRATUM_858921 - Cortex A73 - fixed in 4.12

255a3f3e183 std: Force `Instant::now()` to be monotonic added a Mutex to work around
this problem and a small test program using glommio shows the majority of time spent
acquiring and releasing this Mutex. 3914a7b0da8 tries to improve this, but actually
makes it worse on big systems as for 128b atomics a ldxp/stxp pair (and successful loop)
for v8.4 systems that don't support FEAT_LSE2 is required which is expensive as a lock 
and because of how the load/store-exclusives scale on large Arm systems is both unfair 
to threads and tends to go backwards in performance.

A small sample program using glommio improves by 70x on a 32 core Graviton2
system with this change.

### DIFF
--- a/library/std/src/sys/unix/time.rs
+++ b/library/std/src/sys/unix/time.rs
@@ -303,6 +303,7 @@ mod inner {
         pub fn actually_monotonic() -> bool {
             (cfg!(target_os = "linux") && cfg!(target_arch = "x86_64"))
                 || (cfg!(target_os = "linux") && cfg!(target_arch = "x86"))
+                || (cfg!(target_os = "linux") && cfg!(target_arch = "aarch64"))
                 || cfg!(target_os = "fuchsia")
         }
 
--- a/library/std/src/time.rs
+++ b/library/std/src/time.rs
@@ -263,6 +263,20 @@ impl Instant {
         //
         // To hopefully mitigate the impact of this, a few platforms are
         // excluded as "these at least haven't gone backwards yet".
+        //
+        // While issues have been seen on arm64 platforms the Arm architecture
+        // requires that the counter monotonically increases and that it must
+        // provide a uniform view of system time (e.g. it must not be possible
+        // for a core to recieve a message from another core with a time stamp
+        // and observe time going backwards (ARM DDI 0487G.b D11.1.2). While
+        // there have been a few 64bit SoCs that have bugs which cause time to
+        // not monoticially increase, these have been fixed in the Linux kernel
+        // and we shouldn't penalize all Arm SoCs for those who refuse to
+        // update their kernels:
+        // SUN50I_ERRATUM_UNKNOWN1 - Allwinner A64 / Pine A64 - fixed in 5.1
+        // FSL_ERRATUM_A008585 - Freescale LS2080A/LS1043A - fixed in 4.10
+        // HISILICON_ERRATUM_161010101 - Hisilicon 1610 - fixed in 4.11
+        // ARM64_ERRATUM_858921 - Cortex A73 - fixed in 4.12
         if time::Instant::actually_monotonic() {
             return Instant(os_now);
         }


the8472:
> but actually makes it worse on big systems as for 128b atomics a ldxp/stxp pair (and successful loop)
for v8.4 systems that don't support FEAT_LSE2 is required which is expensive as a lock
and because of how the load/store-exclusives scale on large Arm systems is both unfair
to threads and tends to go backwards in performance.

This is unexpected since in the contended case an optimal implementation of a relaxed fetch_max should make better forward progress than a mutex because it can skip the conditional store part when the load returns an equal or larger value than local value. And the uncontended case should be fast either way.
But I don't know much about arm instructions, so idk if better instruction sequences would be available than what llvm is emitting.

AGSaidi:
> This is unexpected since in the contended case an optimal implementation of a relaxed fetch_max should make better forward progress than a mutex because it can skip the conditional store part when the load returns an equal or larger value than local value. And the uncontended case should be fast either way.
> But I don't know much about arm instructions, so idk if better instruction sequences would be available than what llvm is emitting.

There isn't a 128b un-torn read instruction until arm v8.4, so unless someone (a) has an armv8.4 system (almost none exist in production today);  (b) has configured RUSTFLAGS to build for v8.4+ only; (c) LLVM understands FEAT_LSE2 (i'm not sure it does) an atomic read+write is required to architectually guarantee an untorn read (e.g. both the upper and lower 8 bytes of the 128b value were read simultaneously). So even in the unconteded case every core is doing a read+write instruction. For older systems this looks like a ldxp/stxp (and the store must complete successfully for the ldxp value to be an un-torn 128b read) or on newer systems a casp. For some reason LLVM doesn't emit the casp even when configured with `RUSTFLAGS="-Ctarget-feature=+lse"` but even if it did, that would still be substantially more expensive than the 64b implementation. 

I created PR #88651 which uses 64b reads and improves performance by about ~13x in my tests, however without this change, there is still ~5x being left on the table.

AGSaidi:
A little program that demonstrates the performance impact:
[instant-now-repro.tar.gz](https://github.com/rust-lang/rust/files/7122130/instant-now-repro.tar.gz)

AGSaidi:
Any feedback?

yaahc:
It looks good to me, I'm checking to see if we have any other ARM / time module experts that should take a look at the PR but assuming nobody else has feedback I'm happy to approve.

library/std/src/sys/unix/time.rs
```diff
@@ -303,6 +303,7 @@ mod inner {
         pub fn actually_monotonic() -> bool {
             (cfg!(target_os = "linux") && cfg!(target_arch = "x86_64"))
                 || (cfg!(target_os = "linux") && cfg!(target_arch = "x86"))
+                || (cfg!(target_os = "linux") && cfg!(target_arch = "aarch64"))
```
m-ou-se, line 306: Are there any Linux targets left where the issue is not already fixed by the kernel?

library/std/src/time.rs
```diff
@@ -263,6 +263,20 @@ impl Instant {
...
+        // for a core to recieve a message from another core with a time stamp
+        // and observe time going backwards (ARM DDI 0487G.b D11.1.2). While
+        // there have been a few 64bit SoCs that have bugs which cause time to
+        // not monoticially increase, these have been fixed in the Linux kernel
+        // and we shouldn't penalize all Arm SoCs for those who refuse to
+        // update their kernels:
+        // SUN50I_ERRATUM_UNKNOWN1 - Allwinner A64 / Pine A64 - fixed in 5.1
+        // FSL_ERRATUM_A008585 - Freescale LS2080A/LS1043A - fixed in 4.10
+        // HISILICON_ERRATUM_161010101 - Hisilicon 1610 - fixed in 4.11
+        // ARM64_ERRATUM_858921 - Cortex A73 - fixed in 4.12
```
m-ou-se, line 279: On our [platform support page](https://doc.rust-lang.org/nightly/rustc/platform-support.html) we specifically name Linux 4.2 as the supported kernel. These fixes were all in later versions.

@rust-lang/libs-api How do we feel about merging this? It'd mean that running the code on an older (but supported) kernel version might give an unexpected panic where we'd have to tell the user to update their kernel (or downgrade their Rust).

AGSaidi: While these issues were fixed in the kernel versions I mention above, the patches were also adopted into earlier stable kernels. Assuming they're on a maintained branch of a stable kernel they should have the fixes.

carllerche: For what it is worth, in practice, portable code needs to defend against non-monotonic cases already. For example, [here](https://github.com/tokio-rs/tokio/pull/4044) is where we fixed it in Tokio and the associated discussion is [here](https://github.com/tokio-rs/tokio/issues/3619).

It may be worth to consider documenting `Instant` as "best-effort monotonic".

library/std/src/sys/unix/time.rs
AGSaidi: Linux does include a self-test `linux/tools/testing/selftests/timers/inconsistency-check` that checks for monotonicity, and the manpage for `clock_gettime()` says :
>CLOCK_MONOTONIC
              Clock that cannot be set and represents monotonic time since some unspecified starting point.  This clock is  not  affected  by discontinuous  jumps  in the system time (e.g., if the system administrator manually changes the clock), but is affected by the incremental adjustments performed by adjtime(3) and NTP.

So it seems like Linux is making a user-api guarantee here, but i can't seem to find anything explicitly saying that.

library/std/src/time.rs
the8472: Documenting the backports too might help, at least we can then point users encountering the error on old distros that this is considered a kernel bug fixed on old versions too and there is something they can update to.

AGSaidi: I can dig them up, but fundamentally the kernel should guarantee this is fixed. Taking it to an extreme, if a kernel+driver occasionally corrupted packets, I don't think anyone would suggest that rust's stdlib should wrap all data a user sends in an checksummed container to work around it. The answer there and the answer here should be please switch to a kernel that includes the fix.

the8472: I totally agree, but the workaround exists and if we take it back and some strange system out there starts breaking it would be helpful to be able to point out that this is an already fixed kernel bug or at least that the linux devs generally see this as their turf.

The strategic problem here is that that the fix is "easy" (if done badly...), the downsides of a fix are only noticed by a few (what's a few nanoseconds here and there?), opening a github issue is lower-friction than joining a mailing list and the libs team is excessively nice about these things because the panics may be encountered by users rather than developers in a position to fix them.

Anything that'll make future bug reports easier to handle will be helpful.

AGSaidi: Understood... I've gone digging for backports and here is what I've found. Another approach which I'm a little hesitant to suggest would be to make `is_actually_monotonic()` dependent on if the system booted via UEFI (common on servers, and this issue is almost certainly fixed because time not working right has a lot of interesting implications) vs device tree (typical on dev boards and similar).

In terms of backports i can find:
HISILICON_ERRATUM_161010101 -  i can't find any backports before 4.11, but this HiSilicon chip is for a server and RHEL7/CentOS 7 are the earliest OSes that supported Arm64 and use a 4.11 kernel, so I think were fine there.
FSL_ERRATUM_A008585 - Same for the same logic above.
ARM64_ERRATUM_858921 - This was fixed in 4.12 and the first hardware that used this CPU was released after the 4.12 release, so it's unlikely anyone is using an older kernel on it.
SUN50I_ERRATUM_UNKNOWN1 - The original was back-ported to the stable v4.19 (v4.19.31+) kernel, a follow on fix was back ported to v4.19.198. I've found Debian Buster and Ubuntu Bionic have the workaround.

the8472: > system booted via UEFI (common on servers,

On x86 the server situation isn't great due to some hypervisors introducing non-monotonoicity in the timestamp counters which should provide monotonic results. At least KVM even has an explicit flag where the host would promise that TSC is reliable and even in those cases there still seem to be a few systems out there that make and then break the promise.
I'm not sure what the exact mechanisms are (emulating the instruction? migrating virtual cores between physical ones?).
So a chunk of monotonicity violations come from virtualized systems.

Is the ARM situation different, i.e. is the time source immune to hypervisor interference?

> and this issue is almost certainly fixed because time not working right has a lot of interesting implications

I wouldn't bet on it. Some reporters say actually observing panics due to time going backwards low occurrence rate on their systems. A few times a month or something like that. Applications rarely consist of hot loops doing nothing but calling `Instant::now().duration_since(Instant::now())`.

AGSaidi: > Is the ARM situation different, i.e. is the time source immune to hypervisor interference?

It's possible to for the hypervisor to trap it, but at least KVM doesn't. The value is read from the hardware and then offset by a value originally programmed by the hyperivsor for that VM. If the hardware is functional the VM would have to work to break it. 

> I wouldn't bet on it. 
I'm not specifically talking about Rust, but other applications where the time going backwards of forwards would result in certificate errors, ensconces algorithms going awry, etc.

cuviper:
I'm tempted to say we should just make `elapsed` use saturating subtraction, and leave the monotonicity bugs with the system. Maybe even all `Sub for Instant`, just saturate all the time.

yaahc:
I thought I did this earlier but apparently I did not.

@rustbot ping arm

yaahc:
Absent any concerns I want to treat this like any change to our stable API since it changes observable behavior.

@rfcbot merge

BurntSushi:
Could someone from the ARM group sign off on this? I'd feel more comfortable if we had a domain expert chime in with their support. Pinging them again: @adamgemmell @hug-dev @jacobbramley @JamieCunliffe @joaopaulocarreiro @raw-bin @Stammark

jacobbramley:
Hi! Sorry, I'd got my threads mixed up.

I'm not a kernel/hypervisor expert by any stretch, but I've been assured (by one of the [timer driver maintainers](https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/MAINTAINERS?h=v5.15-rc3#n1387)) that the kernel is the proper place to fix any issues here. He pointed out that the kernel will trap and emulate if necessary (to work around errata) so if we do the same in userspace, we're just doing it twice.

Overall, I think @AGSaidi's reasoning is sound, and I'm happy to approve this from the Arm perspective.

AGSaidi:
@BurntSushi does that address your concern?

BurntSushi:
SGTM!

GuillaumeGomez:
It failed in #89959.

PR #88657, opened 2021-09-05
Fix 2021 `dyn` suggestion that used code as label

camelid:
The arguments to `span_suggestion` were in the wrong order, so the error
looked like this:

    error[E0783]: trait objects without an explicit `dyn` are deprecated
      --> src/test/ui/editions/dyn-trait-sugg-2021.rs:10:5
       |
    10 |     Foo::hi(123);
       |     ^^^ help: <dyn Foo>: `use `dyn``

Now the error looks like this, as expected:

    error[E0783]: trait objects without an explicit `dyn` are deprecated
      --> src/test/ui/editions/dyn-trait-sugg-2021.rs:10:5
       |
    10 |     Foo::hi(123);
       |     ^^^ help: use `dyn`: `<dyn Foo>`

This issue was only present in the 2021 error; the 2018 lint was
correct.

r? @m-ou-se

### DIFF
--- a/compiler/rustc_typeck/src/check/fn_ctxt/_impl.rs
+++ b/compiler/rustc_typeck/src/check/fn_ctxt/_impl.rs
@@ -945,7 +945,7 @@ impl<'a, 'tcx> FnCtxt<'a, 'tcx> {
                     Ok(s) if s.starts_with('<') => sugg,
                     _ => format!("<{}>", sugg),
                 };
-                let replace = String::from("use `dyn`");
+                let sugg_label = "use `dyn`";
                 if self.sess().edition() >= Edition::Edition2021 {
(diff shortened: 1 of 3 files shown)


camelid:
If this lands after 1.56 gets promoted to beta, it may make sense to backport it. It looks like the latest beta is still for Rust 1.55.

camelid:
Looks like 1.56 branched before this landed, so I'll beta-nominate this and then T-compiler can decide whether it's worth backporting.

apiraino:
Beta backport accepted as per compiler team [on Zulip](https://rust-lang.zulipchat.com/#narrow/stream/238009-t-compiler.2Fmeetings/topic/.5Bweekly.5D.202021-09-09.20.2354818/near/252633994)

@rustbot label +beta-accepted

Issue #70887, opened 2020-04-07
Tracking Issue for Integer::{ilog,ilog2,ilog10}


yoshuawuyts:
Feature gate: `#![feature(int_log)]`
Stabilization report: https://github.com/rust-lang/rust/issues/70887#issuecomment-1210602692

---

This is a tracking issue for adding `{checked_,}{ilog,ilog2,ilog10}` methods to `{u8,u16,u32,u64,u128}`. A full rationale can be found in https://github.com/rust-lang/rust/pull/80918#issue-552884252. But in short: integer logarithms [are fairly common](http://graphics.stanford.edu/~seander/bithacks.html#IntegerLog10), but require care to implement correctly, and effort to implement efficiently.

Because we expose `log` methods for floats it can be tempting for users to use that instead, which can lead to rounding errors. In the following example we're trying to calculate `base10` for an integer. We might try and calculate the `base2` for the values, and attempt [a base swap](https://www.rapidtables.com/math/algebra/Logarithm.html#log-rules) to arrive at `base10`. However because we're performing intermediate rounding we arrive at the wrong result:

```rust
// log10(900) = ~2.95 = 2
dbg!(900f32.log10() as u64);
    
// log base change rule: logb(x) = logc(x) / logc(b)
// log2(900) / log2(10) = 9/3 = 3, which is incorrect!
dbg!((900f32.log2() as u64) / (10f32.log2() as u64));
```

### Public API

```rust
// integer log
assert_eq!(5_u16.ilog(5), 1);
assert_eq!(2_u32.ilog2(), 1);
assert_eq!(10_u32.ilog10(), 1);

// checked integer log
assert_eq!(5_u16.checked_ilog(5), Some(1));
assert_eq!(2_u32.checked_ilog2(), Some(1));
assert_eq!(10_u32.checked_ilog10(), Some(1));
assert_eq!(0_u64.checked_ilog(5), None);
```

### Steps / History

- [x] Implementation: https://github.com/rust-lang/rust/pull/80918
- [x] Stabilization report https://github.com/rust-lang/rust/issues/70887#issuecomment-1210602692
- [x] Final comment period (FCP)

### Unresolved Questions

- [x] Can we further optimize the `ilog10` implementation? https://github.com/rust-lang/rust/pull/70835#issuecomment-609976593
- [x] Can we make these methods `const fn`?
[...]

PR #88665, opened 2021-09-05
Change return type for T::{log,log2,log10} to u32.

falk-hueffner:
The value is at most 128, and this is consistent with using u32 for small values elsewhere (e.g. BITS, count_ones, leading_zeros).

### DIFF
--- a/library/core/src/num/int_log10.rs
+++ b/library/core/src/num/int_log10.rs
@@ -116,8 +116,8 @@ mod unchecked {
 
 macro_rules! impl_checked {
     ($T:ident) => {
-        pub const fn $T(val: $T) -> Option<$T> {
-            if val > 0 { Some(unchecked::$T(val) as $T) } else { None }
+        pub const fn $T(val: $T) -> Option<u32> {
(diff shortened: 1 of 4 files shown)


falk-hueffner:
See also the comment here https://github.com/rust-lang/rust/issues/70887#issuecomment-894645141

scottmcm:
I think this makes good sense to me.  `u32::log2` for me is just the more useful version of `leading_zeros`, so having the same return type there seems logical.  And I saw that @joshtriplett :+1:'d the comment so I'll count that as a libs "seems reasonable" for the change, since this is just unstable anyway.

r? @scottmcm 
@bors r+

agausmann:
Super late, but I think it's worth specifically pointing out: this change also agrees with the signature of integer power functions (e.g. [`u64::pow`](https://doc.rust-lang.org/stable/std/primitive.u64.html#method.pow)). All of them use `u32` for the exponent argument.

So with this change, this (albeit contrived) expression passes typechecking:
`2u64.pow(x).log2() == x`

PR #88666, opened 2021-09-05
Improve build command for compiler docs

GuillaumeGomez:
It was rather complicated to document rustc crates. With this, you can directly run:

```console
x.py doc compiler
x.py doc compiler/rustc_hir_pretty
```

The second commit adds the handling of the `--open` flag.

r? @Mark-Simulacrum

### DIFF
--- a/src/bootstrap/doc.rs
+++ b/src/bootstrap/doc.rs
@@ -648,15 +648,24 @@ impl Step for Rustc {
             }
         }
 
+        let mut to_open = None;
         for krate in &compiler_crates {
             // Create all crate output directories first to make sure rustdoc uses
             // relative links.
             // FIXME: Cargo should probably do this itself.
             t!(fs::create_dir_all(out_dir.join(krate)));
             cargo.arg("-p").arg(krate);
+            if to_open.is_none() {
+                to_open = Some(krate);
+            }
         }
 
         builder.run(&mut cargo.into());
+        // Let's open the first crate documentation page:
+        if let Some(krate) = to_open {
+            let index = out.join(krate).join("index.html");
+            open(builder, &index);
+        }
     }
 }
 
--- a/src/bootstrap/doc.rs
+++ b/src/bootstrap/doc.rs
@@ -537,7 +537,7 @@ impl Step for Rustc {
 
     fn should_run(run: ShouldRun<'_>) -> ShouldRun<'_> {
         let builder = run.builder;
-        run.krate("rustc-main").default_condition(builder.config.docs)
+        run.krate("rustc-main").path("compiler").default_condition(builder.config.docs)
     }
 
     fn make_run(run: RunConfig<'_>) {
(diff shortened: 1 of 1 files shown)


src/bootstrap/doc.rs
```diff
@@ -537,7 +537,18 @@ impl Step for Rustc {
 
     fn should_run(run: ShouldRun<'_>) -> ShouldRun<'_> {
         let builder = run.builder;
-        run.krate("rustc-main").default_condition(builder.config.docs)
+
+        let mut paths = Vec::new();
+        for entry in fs::read_dir(builder.src.join("compiler")).expect("read_dir failed") {
```
Mark-Simulacrum, line 542: `krate("rustc-main")` should already do pretty much just this I believe, but without needing to read the directory (it relies on cargo metadata).

GuillaumeGomez: It doesn't allow to do `x.py compiler` or `x.py compiler/rustc_pretty_ast` unfortunately... Which is why I added this way too.

src/bootstrap/doc.rs
```diff
@@ -595,26 +621,52 @@ impl Step for Rustc {
...
             // Create all crate output directories first to make sure rustdoc uses
             // relative links.
             // FIXME: Cargo should probably do this itself.
             t!(fs::create_dir_all(out_dir.join(krate)));
             cargo.arg("-p").arg(krate);
+            to_open = Some(krate);
         }
 
         builder.run(&mut cargo.into());
+        // Let's open the first crate documentation page:
```
Mark-Simulacrum, line 665: This will open the last crate, not the first, right? Since to_open gets assigned in every iteration above?

GuillaumeGomez: Oups indeed! Forgot to check if the variable was already set... Good catch!

GuillaumeGomez:
Updated!

src/bootstrap/doc.rs
Mark-Simulacrum: The first of those I get, but seems like that's solved with a .path("compiler"), not this -- can you clarify the latter though? I would expect that to work (and it seems to work elsewhere -- e.g., the Std doc step has `all_krates("test")`)

GuillaumeGomez: You are absolutely right. I simplified it as you suggested!

src/bootstrap/doc.rs
Mark-Simulacrum: Actually I was also wrong above -- this is going to open a random crate pretty much, because the loop above is iterating a hashset. So just checking that it's none isn't quite so helpful.

I'm personally fine with just leaving it, I doubt people will care much -- my guess is almost no one uses --open more than once at the CLI (just refresh the browser instead).

GuillaumeGomez: Well, let's keep as is then. In the end, the crate they want will be there in any case... :D

PR #88667, opened 2021-09-05
Tweak `write_fmt` doc.

kraktus:
Found this weird sentence while reading the docs.


dtolnay (approved):
Thanks!

The original wording makes sense to me &mdash; it's saying this method is widely used indirectly through format_args, but rarely directly.

The new wording is fine too though.

Issue #65473, opened 2019-10-16
Support index size != pointer width


nw0:
## Preliminaries
`usize` is the pointer-sized unsigned integer type [1].
It is also Rust's index type for slices and loops; this definition works well when pointer size corresponds to the space of indexable objects (most targets today). Informally, `uintptr_t == size_t`.

Note that the target pointer width is indisputably set by the LLVM data layout string.
It would be correct to say that it is currently impossible to have `usize` different to `target_pointer_width` without breaking numerous assumptions in rustc [2, 3].

Unfortunately, `uintptr_t == size_t` doesn't hold for all architectures. For context, I've worked toward (not active) compiling Rust for MIPS/CHERI (CHERI128) [4]. This target has 128-bit capability pointers (as in layout string), and a 64-bit processor and address space.

I also assume that we don't want programmers messing with pointers in Safe Rust, and that they shouldn't have to care how a pointer (or reference) is represented/manipulated by an architecture.

## Problem
I think that more than one type is necessary here, to distinguish between the "index" or "size" component of a pointer (a la `size_t`), and the space required to contain a pointer (`uintptr_t`).

To me, the ideal solution is to change `usize` to be in line with `size_t` and not `uintptr_t`. As @briansmith [notes](https://github.com/rust-lang/rfcs/issues/1748#issuecomment-475368575), this would be a breaking semantic change. I claim that this is only problematic on architectures where `uintptr_t != size_t`. As such, code breakage from changing this assumption is constrained to targets where the code was _already_ broken.

Why not have a 128-bit `usize`? This _is_ technically feasible, and it's the basis of my compilation of Rust for CHERI. But:
* Bounds checks explode from 2 instructions to 7. Yes, this occurs with optimisation on, but no, I haven't profiled it on real-world applications.
* rustc tries to index into LLVM intrinsics such as `memcpy` with 128-bit integers. This isn't defined in the backend, and _arguably shouldn't be defined_. I will not be the first person to wonder why `memcpy` doesn't generate any instructions.
* The address space is 64 bits. `ptr as int` gives an LLVM `i64`, which can't be cast/isn't comparable to an `i128`; again there is no good reason to manipulate 128-bit integers here. Likewise when calling `inttoptr`, which is a valid instruction even if the result can't be dereferenced [5].

As for a `uintptr_t` type. It's optionally defined in C; I'm not sure programmers want to use such a type, and it could be relegated to the compiler. I haven't thought about this seriously, though.

[...]

PR #88668, opened 2021-09-05
Change more x64 size checks to not apply to x32.

hvdijk:
Commit 95e096d6 changed a bunch of size checks already, but more have
been added, so this fixes the new ones the same way: the various size
checks that are conditional on target_arch = "x86_64" were not intended
to apply to x86_64-unknown-linux-gnux32, so add
target_pointer_width = "64" to the conditions.

### DIFF
--- a/compiler/rustc_middle/src/mir/mod.rs
+++ b/compiler/rustc_middle/src/mir/mod.rs
@@ -1708,7 +1708,7 @@ pub struct Place<'tcx> {
     pub projection: &'tcx List<PlaceElem<'tcx>>,
 }
 
-#[cfg(target_arch = "x86_64")]
+#[cfg(all(target_arch = "x86_64", target_pointer_width = "64"))]
 static_assert_size!(Place<'_>, 16);
 
 #[derive(Copy, Clone, Debug, PartialEq, Eq, PartialOrd, Ord, Hash)]
@@ -2034,7 +2034,7 @@ pub enum Operand<'tcx> {
     Constant(Box<Constant<'tcx>>),
 }
 
-#[cfg(target_arch = "x86_64")]
+#[cfg(all(target_arch = "x86_64", target_pointer_width = "64"))]
 static_assert_size!(Operand<'_>, 24);
 
 impl<'tcx> Debug for Operand<'tcx> {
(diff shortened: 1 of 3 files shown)


CryZe:
Do these even need the target_arch at all and not just the pointer width?

hvdijk:
> Do these even need the target_arch at all and not just the pointer width?

Good question. I think some of the checks could potentially fail if some target has 64-bit pointers that do not require 64-bit alignment, but I do not know whether Rust supports any such target. I wouldn't want to have that change as part of this one in case it needs to be reverted (the revert would break x32 again), but I can submit it as a followup PR if you like.

jrtc27:
Does `target_pointer_width` mean `sizeof(void *)` or `log2(__UINTPTR_MAX__ + 1)` (i.e. is it defined to be the size of a pointer or the range you can address)? If the latter, using a bare `target_pointer_width == 64` would make things worse for CHERI, where pointers carry additional metadata beyond the address by using "capabilities", so 64-bit architectures have 128-bit pointers but still only a 64-bit address space (even the current patch would cause issues for a CHERI x86_64, but that is so far just a sketch in our ISA spec). The former is fine for CHERI but is often not what people use these kinds of things to mean.

(See also #65473 that tracks `usize` as defined conflating the ideas of `size_t` and `uintptr_t`)

hvdijk:
> Does `target_pointer_width` mean `sizeof(void *)` or `log2(__UINTPTR_MAX__ + 1)` (i.e. is it defined to be the size of a pointer or the range you can address)?

That's not clear to me from the specification. https://doc.rust-lang.org/reference/conditional-compilation.html#target_pointer_width just says "Key-value option set once with the target's pointer width in bits."

> If the latter, using a bare `target_pointer_width == 64` would make things worse for CHERI, where pointers carry additional metadata beyond the address by using "capabilities",

Let's play it safe and forget about that for now, then.

> (even the current patch would cause issues for a CHERI x86_64, but that is so far just a sketch in our ISA spec)

The current patch wouldn't cause issues for that, as it doesn't enable any static asserts that were previously disabled: any assert that would fail with this patch already fails the same way without this patch.

jrtc27:
> > Does `target_pointer_width` mean `sizeof(void *)` or `log2(__UINTPTR_MAX__ + 1)` (i.e. is it defined to be the size of a pointer or the range you can address)?
> 
> That's not clear to me from the specification. https://doc.rust-lang.org/reference/conditional-compilation.html#target_pointer_width just says "Key-value option set once with the target's pointer width in bits."
> 
> > If the latter, using a bare `target_pointer_width == 64` would make things worse for CHERI, where pointers carry additional metadata beyond the address by using "capabilities",
> 
> Let's play it safe and forget about that for now, then.
> 
> > (even the current patch would cause issues for a CHERI x86_64, but that is so far just a sketch in our ISA spec)
> 
> The current patch wouldn't cause issues for that, as it doesn't enable any static asserts that were previously disabled: any assert that would fail with this patch already fails the same way without this patch.

Agreed :) and there are already asserts of that form elsewhere

joshtriplett:
This seems reasonable to me.

@bors r+

PR #83536, opened 2021-03-26
Suggest `i += 1` when we see `i++` or `++i`

camelid:
Closes #83502 (for `i++` and `++i`; `--i` should be covered by #82987 and `i--`
is tricky to handle).

PR #88672, opened 2021-09-05
Suggest `i += 1` when we see `i++` or `++i`

### LINKED ISSUE #83502: Add diagnostic for ++ and -- usages
Given the following code [in the playground:](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=14fc6b865440a61afb2d57546fe99d35)

```rust
fn main() {
    let i = 0;
    i++;
}
```

The current output is:

```
 --> src/main.rs:3:7
  |
3 |     i++;
  |       ^ expected expression

```

I think the message for errors involving `++` or `--` should understand that the user is trying to use an operator which doesn't exist in Rust and suggest an idiomatic way to do the increment/decrement operation.

# Explanation

Languages like C, Java or Javascript use `++` and `--` to increment or decrement integer variables, respectively. As Rust does not support them, it may be good to add an specific diagnostic for those cases, suggesting the user to use something like `i += 1` instead.

camelid:
Closes #83502 (for `i++` and `++i`; `--i` should be covered by #82987, and `i--`
is tricky to handle).

This is a continuation of #83536.

r? @estebank


src/test/ui/parser/increment.stderr
```diff
@@ -0,0 +1,52 @@
...
+error: expected expression, found `+`
+  --> $DIR/increment.rs:8:13
+   |
+LL |     while i++ < 5 {
+   |             ^ expected expression
+   |
+   = note: Rust has no dedicated increment operator
+help: try using `+= 1` instead
+   |
+LL |     while i += 1 < 5 {
```
camelid, line 22: I'm not sure how to fix this suggestion so it works like the pre-increment suggestion. The issue is that by the time we get to the second `+`, we've forgotten about the `i` token, so I don't see a good way of getting the starting span.

compiler/rustc_parse/src/parser/diagnostics.rs
```diff
@@ -1788,6 +1788,49 @@ impl<'a> Parser<'a> {
...
+            && self.look_ahead(2, |t| !t.is_lit())
+        {
+            let target_span = self.look_ahead(2, |t| t.span);
+            let left_brace_span = target_span.shrink_to_lo();
+            let pre_span = self.token.span.to(self.look_ahead(1, |t| t.span));
+            let post_span = target_span.shrink_to_hi();
+
+            err.note("Rust has no dedicated increment operator");
+
+            if self.prev_token.kind == TokenKind::Semi {
```
camelid, line 1814: This seems to be a good heuristic so that we don't suggest `{ i += 1; i};`. However, I wonder if there are places where this heuristic yields a false positive, and we emit code that is different from what the user meant. It seems unlikely, but should we change the applicability for this version of the suggestion to `MaybeIncorrect` just in case?

src/test/ui/parser/increment.stderr
camelid: Note to self: If we fix this, I should make sure to add `// run-rustfix` to the test.

compiler/rustc_parse/src/parser/diagnostics.rs
```diff
@@ -1788,6 +1788,49 @@ impl<'a> Parser<'a> {
             self.sess.expr_parentheses_needed(&mut err, *sp);
         }
         err.span_label(span, "expected expression");
+        if self.prev_token.kind == TokenKind::BinOp(token::Plus)
```
b-naber, line 1791: Can you maybe add a short comment here explaining what is done is this block, so that one can more easily skip reading this block?

estebank: Or even better move it to its own method.

compiler/rustc_parse/src/parser/diagnostics.rs
estebank: Looking at all these changes, I feel like a better place to put this is in `parse_lit_expr`. That way we can

* give a specific error instead of the generic "expected expression":

```
error: Rust has no postfix increment operator
  --> $DIR/increment.rs:3:7
   |
LL |     i++;
   |      ^^ not a valid unary operator
   |
help: use in-place increment instead
   |
LL |     i += 1;
   |       ~~~~
```

* we can recover the parser and return that as an expression (which would also give you mutability and type errors if appropriate)
* gives us a good place where we can look for the prefix version: you add a `None if matches!(self.token.kind, Plus | Minus)` (and a lookahead 1 and 2 to check for another `Plus | Minus` and for a numeric literal or ident) case in the match in that function
* you can still check for whether the next thing will be a `Semi` to not give the `{}` surrounded version

`parse_lit_expr` is already the fall-through of `parse_bottom_expr`, which is effectively where expressions are parsed.

camelid: > you can still check for whether the next thing will be a `Semi` to not give the `{}` surrounded version

Actually I need to check for whether the *previous* thing was a `Semi`, because otherwise on code like this:

```rust
let y = x++;

we will suggest this:

```rust
let y = x += 1;
```

How should I get around this? Will the `prev_token` in `parse_lit_expr` work?

camelid: It looks like by the time `parse_lit_expr` is invoked for the postfix version, we've lost the context of the previous `;` because the `prev_token` is the variable being incremented. What should I do?

camelid: Actually, it looks like `prev_token` is the first `+`, not even the variable. This seems like a problem...

camelid: I moved the code to `parse_assoc_expr_with()` (and extracted a bunch of helper functions), which has more information.

compiler/rustc_parse/src/parser/diagnostics.rs
```diff
@@ -1788,6 +1788,49 @@ impl<'a> Parser<'a> {
...
+            && self.token.kind == TokenKind::BinOp(token::Plus)
+            && self.look_ahead(1, |t| !t.is_lit())
+        {
+            let span = self.prev_token.span.to(self.token.span);
+            err.note("Rust has no dedicated increment operator");
+            err.span_suggestion_verbose(
+                span,
+                "try using `+= 1` instead",
```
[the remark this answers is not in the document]
camelid: The version you're suggesting sounds confusing to me – `i++` is also "in-place increment".

src/test/ui/parser/increment.stderr
camelid: Fixed in newer version (still working on it, so haven't pushed).

compiler/rustc_parse/src/parser/diagnostics.rs
camelid: It's in its own method (actually, multiple methods) now.

compiler/rustc_parse/src/parser/diagnostics.rs
```diff
@@ -1076,6 +1120,145 @@ impl<'a> Parser<'a> {
...
+        err.multipart_suggestion(
+            &format!("use `{}= 1` instead", kind.op.chr()),
+            vec![
+                (pre_span, "{ ".to_string()),
+                (post_span, format!(" {}= 1; {} }}", kind.op.chr(), ident)),
+            ],
+            Applicability::MachineApplicable,
+        );
+        err.emit();
+        // TODO: recover
```
camelid, line 1219: It seems like recovering from this might be fairly involved. Is it worth the added code complexity to return an expression like `{ i += 1; i }` instead of having the user just fix their code and rebuild to see further errors?

compiler/rustc_parse/src/parser/expr.rs
```diff
@@ -230,6 +230,18 @@ impl<'a> Parser<'a> {
...
             }
 
+            if self.prev_token == token::BinOp(token::Plus)
+                && self.token == token::BinOp(token::Plus)
+            {
+                let op_span = self.prev_token.span.to(self.token.span);
+                // Eat the second `+`
+                self.bump();
+                // TODO: implement
+                let start_is_semi = false;
```
camelid, line 240: I'm not sure how to figure out if the previous token is a semicolon. It seems like even at the start of this function, `prev_token` refers to the `i`.

src/test/ui/parser/increment.stderr
```diff
@@ -0,0 +1,47 @@
+error: Rust has no postfix increment operator
+  --> $DIR/increment.rs:5:6
+   |
+LL |     i++;
+   |      ^^ not a valid postfix operator
+   |
+help: use `+= 1` instead
+   |
+LL |     { let tmp = i; i += 1; tmp };
+   |     +++++++++++  ~~~~~~~~~~~~~~~
```
camelid, line 10: This is because of the code I commented on above where I don't know how to tell if this previous token is a semicolon. I'm not even certain if that's a good heuristic, because if we're at the beginning of a function, the previous token will be a `{`. What I really want is to know if are at the beginning of a new statement. Is there any way to determine that?

compiler/rustc_parse/src/parser/diagnostics.rs
```diff
@@ -1076,6 +1120,145 @@ impl<'a> Parser<'a> {
...
+        mut err: DiagnosticBuilder<'_>,
+        kind: IncDecRecovery,
+        ident: Ident,
+        (pre_span, post_span): (Span, Span),
+    ) -> PResult<'a, P<Expr>> {
+        err.multipart_suggestion(
+            &format!("use `{}= 1` instead", kind.op.chr()),
+            vec![
+                (pre_span, "{ let tmp = ".to_string()),
+                (post_span, format!("; {} {}= 1; tmp }}", ident, kind.op.chr())),
```
camelid, line 1235: One problem here is if the user's code was `tmp++`. Then the generated code will be incorrect. Is there a better way to do this (especially because it's currently marked as machine-applicable)?

compiler/rustc_parse/src/parser/diagnostics.rs
```diff
@@ -1076,6 +1120,145 @@ impl<'a> Parser<'a> {
...
+    ) -> PResult<'a, P<Expr>> {
+        let mut err = self.struct_span_err(
+            op_span,
+            &format!("Rust has no {} {} operator", kind.fixity, kind.op.name()),
+        );
+        err.span_label(op_span, &format!("not a valid {} operator", kind.fixity));
+
+        if let ExprKind::Path(_, path) = &base.kind {
+            if let [segment] = path.segments.as_slice() {
+                let ident = segment.ident;
```
camelid, line 1167: This doesn't work for code like `foo.bar++`. Instead, we fall back to the base case:

```
error: Rust has no postfix increment operator
 --> bar.rs:2:12
  |
2 |     foo.bar++;
  |            ^^ not a valid postfix operator
  |
  = help: use `+= 1` instead
```

Is there a way to fix this simply?

camelid:
Is there any way this could accidentally break macros that have `i++`–style syntax? I tested it a bit, and I didn't find any issues, but I thought it'd be worth bringing up.

camelid:
(CI is just failing because I used `// TODO`, which was intentional so they would be resolved before this is merged.)

pnkfelix:
Hi @camelid ; we visited this for triage. 

It looks like you had some questions you wanted answered, but also, @estebank had a suggestion for a [change](https://github.com/rust-lang/rust/pull/88672#discussion_r702776614) they wanted to see.

What is the status on those fronts? Do you think you are stuck waiting for more feedback from @estebank, or can you make forward progress on the suggestion they provided in the meantime?

camelid:
It's been a while, but my recollection is that I implemented most of the user-facing side of what estebank suggested, but in a different way in the parser. The way the change is implemented currently works pretty well, though sometimes the suggestions aren't as good as they could be. Perhaps it would be good to accept the current state of output as good enough for an "MVP PR" since it can be improved further in the future. That's of course assuming there aren't any major issues in the output.

So, in summary: I think I need estebank to see if the impl looks okay and the error output looks good enough for a first implementation.

camelid:
@estebank could you take another look at this when you get a chance?

camelid:
I'll wait to rebase until this is reviewed in case it needs more rebases.

compiler/rustc_parse/src/parser/diagnostics.rs
davidtwco: Just guessing from looking at the AST we generate on Compiler Explorer, I assume you don't just need to peel back N layers of `ExprKind::Field` to get to the innermost `ExprKind::Path` and then continue as normal?

compiler/rustc_parse/src/parser/diagnostics.rs
davidtwco: At least initially, if recovery is challenging to implement, we can skip that so that this can land.

compiler/rustc_parse/src/parser/diagnostics.rs
```diff
@@ -1076,6 +1122,145 @@ impl<'a> Parser<'a> {
...
+                // (pre, post)
+                let spans = match kind.fixity {
+                    UnaryFixity::Pre => (op_span, ident.span.shrink_to_hi()),
+                    UnaryFixity::Post => (ident.span.shrink_to_lo(), op_span),
+                };
+
+                if !ident.is_reserved() {
```
davidtwco, line 1176: Is there any way we could cut down on how nested this expression is getting?

compiler/rustc_parse/src/parser/expr.rs
davidtwco: I can't see a way to do this in this function outside of threading some context through (which we probably don't want?) or finding somewhere else to hook this diagnostic in at.

davidtwco (commented):
Apologies for how long it has taken for this to be reviewed again, this looks  basically good to me (at least as a starting point for this suggestion), I've left some comments on your open questions.

compiler/rustc_parse/src/parser/diagnostics.rs
camelid: I added early returns. How is it now?

compiler/rustc_parse/src/parser/diagnostics.rs
camelid: Sounds good.

compiler/rustc_parse/src/parser/diagnostics.rs
camelid: Looking at this now, I realize this code isn't the issue (we can just use `span_to_snippet` to avoid the AST manipulation). The issue is that when parsing `++foo.bar`, we need to determine how far ahead to parse. Currently it just parses a single path segment, but that doesn't work for `foo.bar`. Thoughts?

compiler/rustc_parse/src/parser/expr.rs
camelid: Any ideas what I should do?

davidtwco: I think it would be best if just assume there isn't a semicolon here and downgrade the applicability of the suggestion accordingly (perhaps adjusting the message to emphasise that the user _might_ want the suggestion). Then we can land this as an improvement and revisit that case in future.

src/test/ui/parser/increment.stderr
davidtwco: I don't think there's a great way of knowing that, outside of keeping track of more context - it can be left for a follow-up and we can consider the suggestion possibly correct.

davidtwco:
I think we should prioritize tidying up the suggestions this adds which do work so we can land those, and we can revisit the difficult cases in another PR.

src/test/ui/parser/increment.stderr
camelid: Well, it is actually "correct", just unnecessarily complicated.

compiler/rustc_parse/src/parser/expr.rs
camelid: Ok, I think I found a good workaround. Basically, emit both suggestions. However, rustfix is crashing:

```

---- [ui] ui/parser/increment-autofix.rs stdout ----
thread '[ui] ui/parser/increment-autofix.rs' panicked at 'failed to apply suggestions for "/Users/noahlev/code/rust/r1/src/test/ui/parser/increment-autofix.rs" with rustfix: Cannot replace slice of data that was already replaced', src/tools/compiletest/src/runtest.rs:3131:17
```

Any ideas?

compiler/rustc_parse/src/parser/diagnostics.rs
camelid: Added a workaround.

src/test/ui/parser/increment.stderr
camelid: See https://github.com/rust-lang/rust/pull/88672#discussion_r809538489.

compiler/rustc_parse/src/parser/diagnostics.rs
```diff
@@ -1112,6 +1165,150 @@ impl<'a> Parser<'a> {
...
+    fn postfix_inc_dec_suggest(
+        &mut self,
+        base_src: String,
+        err: &mut DiagnosticBuilder<'a>,
+        kind: IncDecRecovery,
+        (pre_span, post_span): (Span, Span),
+    ) {
+        let mut msg = format!("use `{}= 1` instead", kind.op.chr());
+        if base_src.trim() == "tmp" {
+            msg.push_str(" (rename `tmp` so it doesn't conflict with your variable)");
```
camelid, line 1275: Wording could probably be improved here.

src/test/ui/parser/increment-autofix.stderr
```diff
@@ -0,0 +1,89 @@
...
+  --> $DIR/increment-autofix.rs:5:6
+   |
+LL |     i++;
+   |      ^^ not a valid postfix operator
+   |
+help: use `+= 1` instead
+   |
+LL |     { let tmp = i; i += 1; tmp };
+   |     +++++++++++  ~~~~~~~~~~~~~~~
+help: or, if you don't need to use it as an expression, change it to this
```
davidtwco, line 11:
```suggestion
help: or, if you don't need to use the incremented value
```
I think this might be a little bit clearer, avoids terminology like "expression" which might not be obvious to a beginner.

compiler/rustc_parse/src/parser/diagnostics.rs
davidtwco: If we can detect this, why can't we just change the variable name in our suggestion?

compiler/rustc_parse/src/parser/diagnostics.rs
```diff
@@ -150,6 +150,59 @@ impl AttemptLocalParseRecovery {
...
+    /// Is this an increment or decrement?
+    op: IncOrDec,
+    /// Is this pre- or postfix?
+    fixity: UnaryFixity,
+}
+
+#[derive(Debug, Copy, Clone, PartialEq, Eq)]
+enum IncOrDec {
+    Inc,
+    // FIXME: `i--` recovery isn't implemented yet
```
davidtwco, line 170: Are you intending to do this in the current PR?

compiler/rustc_parse/src/parser/diagnostics.rs
```diff
@@ -150,6 +150,59 @@ impl AttemptLocalParseRecovery {
...
 }
 
+/// Information for emitting suggestions and recovering from
+/// C-style `i++`, `--i`, etc.
+#[derive(Debug, Copy, Clone)]
+struct IncDecRecovery {
+    /// Is this increment/decrement its own statement?
+    ///
+    /// This is `None` when we are unsure.
+    standalone: Option<bool>,
```
davidtwco, line 160: nit: it might be more self-documenting if this was an enum with three variants, I skipped over the comment at first and it took a second to understand what situation `None` corresponded to

davidtwco (changes requested):
This is great, left a few more comments, should be able to land this soon.

compiler/rustc_parse/src/parser/expr.rs
davidtwco: Can you use [`span_suggestions`](https://doc.rust-lang.org/stable/nightly-rustc/rustc_errors/diagnostic/struct.Diagnostic.html#method.span_suggestions)?

camelid: Ah, is that new? The only problem with it is I can't have different messages, which might be confusing.

compiler/rustc_parse/src/parser/diagnostics.rs
camelid: No, `i--` and `--i` are tricky to handle because in many cases they're valid code. I'd rather save them for later so this PR can land.

src/test/ui/parser/increment-autofix.stderr
camelid: Yeah, I was a bit concerned about saying "expression". However, "the incremented value" is actually not quite right because the suggestion yields the pre-increment value. What about this?

```suggestion
help: or, if you don't need to use the pre-increment value
```

or this?

```suggestion
help: or, if you don't need to use the old value
```

compiler/rustc_parse/src/parser/diagnostics.rs
camelid: I thought about doing that but worried it might be overkill. My idea was to have a loop where we add underscores until the name doesn't match. Alternatively, it could append a number. Personally I prefer the underscore approach because I doubt it'll require more than one or two underscores, whereas `tmp4` could be a real variable.

camelid: I just realized that there's no need for a loop because we're only comparing against one variable 🤦

compiler/rustc_parse/src/parser/diagnostics.rs
camelid: Done, good idea.

camelid:
Thanks for the helpful reviews by the way! :)

camelid:
`let_else` was added a couple days ago so I had to rebase to remove my copy of it.

src/test/ui/parser/increment-autofix.stderr
davidtwco: Either of these are good, I don't mind which you pick, I think the first is clearer but it also has more terminology

compiler/rustc_parse/src/parser/expr.rs
davidtwco: I think it's the only option we've got, so we'll just need to find a message that works.

camelid:
Hmm, rustfix is still panicking, even with now using `multipart_suggestions` (notice that it's plural).

davidtwco:
> Hmm, rustfix is still panicking, even with now using `multipart_suggestions` (notice that it's plural). How do I fix this?

Apologies for the delay in responding - I'm not entirely sure, could you try commenting some of the suggestions locally and see if removing any of those fixes the problem? Perhaps the `IsStandalone::Maybe` case? Failing that, I'd see if you can't work out what is going on within `rustfix` to cause it to fail, as much as that's a bunch more work.

camelid:
I opened an issue for rustfix: https://github.com/rust-lang/rustfix/issues/209. I'm pretty sure the issue is on its end, so I'm not sure we can fix it in rustc.

How about if we mark the suggestions as non-MachineApplicable in the meantime until the bug is resolved?

davidtwco:
> I opened an issue for rustfix: [rust-lang/rustfix#209](https://github.com/rust-lang/rustfix/issues/209). I'm pretty sure the issue is on its end, so I'm not sure we can fix it in rustc.
> 
> How about if we mark the suggestions as non-MachineApplicable in the meantime until the bug is resolved?

I think that makes sense.

camelid:
Turns out the panic is still happening without MachineApplicable. I asked ehuss about it on the rustfix issue.

camelid:
Ok, I got the tests working now! (You can read the rustfix issue for the background on what the problem and solution were.) Thanks to ehuss for their help with solving this.

I'm going to rebase now and then this should finally be ready for merge!

camelid:
Ok, hopefully CI should pass now.

davidtwco (approved):

davidtwco:
This is great, thanks for seeing this through. Could you squash some of the commits? r=me after that.

camelid:
@davidtwco It's hard to squash without squashing all of them into one commit, and I'd rather keep the history to make it easier to understand the design choices for the code (e.g., when `git blame`ing). Is it okay to leave it unsquashed?

camelid:
Thank you so much for helping me with this PR and being patient even though it took me forever to finish it! ❤️

Issue #83761, opened 2021-04-01
rustdoc: Stop cloning the resolver


jyn514:
Rustdoc currently creates a copy of the resolver to use for intra-doc links: https://github.com/rust-lang/rust/blob/d474075a8f28ae9a410e95d849d009006db4b176/src/librustdoc/core.rs#L350-L354
This is a Terrible, Horrible, No Good, Very Bad Idea. In particular, it causes rustdoc's copy of the resolver and the TyCtxt to disagree about what crates exist:
- https://github.com/rust-lang/rust/issues/66159#issuecomment-550528338
- https://github.com/rust-lang/rust/pull/75176
- https://github.com/rust-lang/rust/pull/82496#issuecomment-785399205

It's also distorting rustc_resolve, since not all of the outputs make sense to clone in the first place: https://github.com/rust-lang/rust/pull/65625#discussion_r336780545

We should refactor rustdoc somehow to allow getting rid of `Resolver::clone_outputs`. @petrochenkov suggests moving anything that needs to touch the resolver before HIR lowering: https://github.com/rust-lang/rust/issues/68427#issuecomment-578493425.

@petrochenkov what do you think about @eddyb's comment in https://github.com/rust-lang/rust/pull/65625#discussion_r336781320 ?
> Why would cloning be needed? Is this that support for resolving things after rustc_resolve finishes?
In that case I'm not sure why we need to clone the outputs - is it to be able to keep the original resolver around?

Would it be possible for `lower_to_hir` to stop stealing the resolver? https://github.com/rust-lang/rust/blob/d474075a8f28ae9a410e95d849d009006db4b176/compiler/rustc_interface/src/queries.rs#L227

 Then rustdoc wouldn't need to clone it in the first place, it could just call `queries.expansion().peek().1` whenever it needs access to the resolver.

See https://github.com/rust-lang/rust/issues/68427 for previous discussion.

---

## Implementation History

- [x] Initial "early crate loader": https://github.com/rust-lang/rust/pull/83738
- [x] Don't require `clean::Item`s before resolving intra-doc links: https://github.com/rust-lang/rust/pull/83833
- [x] Separate type-relative resolution from path resolution (at least somewhat): https://github.com/rust-lang/rust/pull/83849
- [x] Move early loader into `collect_intra_doc_links`: https://github.com/rust-lang/rust/pull/84101
- [ ] Move all path resolution from late to early module and add new field on DocContext for partially resolved links
[...]

PR #78797, opened 2020-11-06
Use real Spans when resolving intra-doc links

pitaj:
Fixes #78696

Tested this works locally by modifying `bufreader.rs` like so:
```diff
 /// data loss.
 ///
-/// [`TcpStream::read`]: Read::read
+/// [`TcpStream::read`]: crate::net::TcpStream::read
 /// [`TcpStream`]: crate::net::TcpStream
```

@rustbot modify labels: T-doc, A-intra-doc-links

r? @jyn514

Issue #78696, opened 2020-11-03
intra-doc links: `crate::` considers the traits in the crate scope, not the current scope


pitaj:
I tried this code:

```rust
#![allow(unused_imports, dead_code)]

// Uncomment this and the warning will go away
// use crate::io::Read;

pub mod io {
    pub trait Read {
        fn read(&mut self, buf: &mut [u8]) -> usize;
    }
}

pub mod bufreader {
    use crate::io::Read;

    /// It can be excessively inefficient to work directly with a [`Read`] instance.
    /// For example, every [`Read::read`] call to [`TcpStream::read`] or [`TcpStream::yes`] on [`TcpStream`]
    ///
    /// [`TcpStream::yes`]: crate::net::TcpStream::yes
    /// [`TcpStream::read`]: crate::net::TcpStream::read
    /// [`Read::read`]: Read::read
    /// [`TcpStream`]: crate::net::TcpStream
    pub struct BufReader;
}

pub mod net {
    pub struct TcpStream;

    impl TcpStream {
        fn yes() -> bool {
            true
        }
    }

    impl crate::io::Read for TcpStream {
        fn read(&mut self, buf: &mut [u8]) -> usize {
            buf.len()
        }
    }
}
```

I expected to see this happen: Rustdoc should complete with no warnings

Instead, this happened:

Only the following warning is output by rustdoc
```sh
warning: unresolved link to `self::net::TcpStream::read`
  --> src/main.rs:13:30
   |
13 |     /// [`TcpStream::read`]: crate::net::TcpStream::read
   |                              ^^^^^^^^^^^^^^^^^^^^^^^^^^^ the struct `TcpStream` has no field or associated item named `read`
   |
   = note: `#[warn(broken_intra_doc_links)]` on by default
```

If you uncomment the line as it says, the warning goes away.

### Meta

`rustc +stage2 --version --verbose`:
```
rustc 1.49.0-dev
binary: rustc
commit-hash: unknown
commit-date: unknown
host: x86_64-unknown-linux-gnu
release: 1.49.0-dev
```

Commit hash: b2025326088b54fb3f083bebeba14e0a15bf00d3

Here's a crate that repros this issue: https://github.com/pitaj/rustdoc-78006-repro

@rustbot modify labels: T-doc, A-intra-doc-links

@jyn514 Here's the repro I promised.

First came across this in #78006

jyn514:
Slightly smaller:

```rust
// Uncomment this and the warning will go away
[...]

PR #88679, opened 2021-09-05
rustdoc: Pre-calculate traits that are in scope for doc links

petrochenkov:
This eliminates one more late use of resolver (part of #83761).
At early doc link resolution time we go through parent modules of items from the current crate, reexports of items from other crates, trait items, and impl items collected by `collect-intra-doc-links` pass, determine traits that are in scope in each such module, and put those traits into a map used by later rustdoc passes.
r? @jyn514

### DIFF
--- a/compiler/rustc_resolve/src/build_reduced_graph.rs
+++ b/compiler/rustc_resolve/src/build_reduced_graph.rs
@@ -108,7 +108,7 @@ impl<'a> Resolver<'a> {
     /// Reachable macros with block module parents exist due to `#[macro_export] macro_rules!`,
     /// but they cannot use def-site hygiene, so the assumption holds
     /// (<https://github.com/rust-lang/rust/pull/77984#issuecomment-712445508>).
-    crate fn get_nearest_non_block_module(&mut self, mut def_id: DefId) -> Module<'a> {
+    pub fn get_nearest_non_block_module(&mut self, mut def_id: DefId) -> Module<'a> {
         loop {
             match self.get_module(def_id) {
                 Some(module) => return module,
--- a/compiler/rustc_resolve/src/lib.rs
+++ b/compiler/rustc_resolve/src/lib.rs
@@ -614,7 +614,8 @@ impl<'a> ModuleData<'a> {
         }
     }
 
-    fn def_id(&self) -> DefId {
+    // Public for rustdoc.
+    pub fn def_id(&self) -> DefId {
         self.opt_def_id().expect("`ModuleData::def_id` is called on a block module")
     }
 
@@ -3405,6 +3406,16 @@ impl<'a> Resolver<'a> {
     // For rustdoc.
     pub fn all_macros(&self) -> &FxHashMap<Symbol, Res> {
         &self.all_macros
+    }
+
+    /// For rustdoc.
+    /// For local modules returns only reexports, for external modules returns all children.
+    pub fn module_children_or_reexports(&self, def_id: DefId) -> Vec<ModChild> {
+        if let Some(def_id) = def_id.as_local() {
+            self.reexport_map.get(&def_id).cloned().unwrap_or_default()
+        } else {
+            self.cstore().module_children_untracked(def_id, self.session)
+        }
     }
 
     /// Retrieves the span of the given `DefId` if `DefId` is in the local crate.
--- a/src/librustdoc/passes/collect_intra_doc_links/early.rs
+++ b/src/librustdoc/passes/collect_intra_doc_links/early.rs
@@ -7,11 +7,15 @@ use rustc_ast::visit::{self, AssocCtxt, Visitor};
 use rustc_ast::{self as ast, ItemKind};
 use rustc_ast_lowering::ResolverAstLowering;
 use rustc_hir::def::Namespace::TypeNS;
-use rustc_hir::def_id::{DefId, LocalDefId, CRATE_DEF_ID};
-use rustc_resolve::Resolver;
+use rustc_hir::def::{DefKind, Res};
+use rustc_hir::def_id::{DefId, DefIdMap, DefIdSet, LocalDefId, CRATE_DEF_ID};
+use rustc_hir::TraitCandidate;
+use rustc_middle::ty::{DefIdTree, Visibility};
+use rustc_resolve::{ParentScope, Resolver};
 use rustc_session::config::Externs;
-use rustc_span::{Span, DUMMY_SP};
+use rustc_span::{Span, SyntaxContext, DUMMY_SP};
 
+use std::collections::hash_map::Entry;
 use std::mem;
 
 crate fn early_resolve_intra_doc_links(
@@ -22,15 +26,18 @@ crate fn early_resolve_intra_doc_links(
     let mut loader = IntraLinkCrateLoader {
         resolver,
         current_mod: CRATE_DEF_ID,
+        visited_mods: Default::default(),
+        traits_in_scope: Default::default(),
         all_traits: Default::default(),
         all_trait_impls: Default::default(),
     };
 
     // Overridden `visit_item` below doesn't apply to the crate root,
-    // so we have to visit its attributes and exports separately.
+    // so we have to visit its attributes and reexports separately.
     loader.load_links_in_attrs(&krate.attrs, krate.span);
+    loader.process_module_children_or_reexports(CRATE_DEF_ID.to_def_id());
     visit::walk_crate(&mut loader, krate);
-    loader.fill_resolver_caches();
+    loader.add_foreign_traits_in_scope();
 
     // FIXME: somehow rustdoc is still missing crates even though we loaded all
     // the known necessary crates. Load them all unconditionally until we find a way to fix this.
@@ -46,6 +53,7 @@ crate fn early_resolve_intra_doc_links(
     }
 
     ResolverCaches {
+        traits_in_scope: loader.traits_in_scope,
         all_traits: Some(loader.all_traits),
         all_trait_impls: Some(loader.all_trait_impls),
     }
@@ -54,26 +62,86 @@ crate fn early_resolve_intra_doc_links(
 struct IntraLinkCrateLoader<'r, 'ra> {
     resolver: &'r mut Resolver<'ra>,
     current_mod: LocalDefId,
+    visited_mods: DefIdSet,
+    traits_in_scope: DefIdMap<Vec<TraitCandidate>>,
     all_traits: Vec<DefId>,
     all_trait_impls: Vec<DefId>,
 }
 
 impl IntraLinkCrateLoader<'_, '_> {
-    fn fill_resolver_caches(&mut self) {
-        for cnum in self.resolver.cstore().crates_untracked() {
-            let all_traits = self.resolver.cstore().traits_in_crate_untracked(cnum);
-            let all_trait_impls = self.resolver.cstore().trait_impls_in_crate_untracked(cnum);
+    fn add_traits_in_scope(&mut self, def_id: DefId) {
+        // Calls to `traits_in_scope` are expensive, so try to avoid them if only possible.
+        // Keys in the `traits_in_scope` cache are always module IDs.
+        if let Entry::Vacant(entry) = self.traits_in_scope.entry(def_id) {
+            let module = self.resolver.get_nearest_non_block_module(def_id);
+            let module_id = module.def_id();
+            let entry = if module_id == def_id {
+                Some(entry)
+            } else if let Entry::Vacant(entry) = self.traits_in_scope.entry(module_id) {
+                Some(entry)
+            } else {
+                None
+            };
+            if let Some(entry) = entry {
+                entry.insert(self.resolver.traits_in_scope(
+                    None,
+                    &ParentScope::module(module, self.resolver),
+                    SyntaxContext::root(),
+                    None,
+                ));
+            }
+        }
+    }
+
+    fn add_traits_in_parent_scope(&mut self, def_id: DefId) {
+        if let Some(module_id) = self.resolver.parent(def_id) {
+            self.add_traits_in_scope(module_id);
+        }
+    }
+
+    /// Add traits in scope for links in impls collected by the `collect-intra-doc-links` pass.
+    /// That pass filters impls using type-based information, but we don't yet have such
+    /// information here, so we just conservatively calculate traits in scope for *all* modules
+    /// having impls in them.
+    fn add_foreign_traits_in_scope(&mut self) {
+        for cnum in Vec::from_iter(self.resolver.cstore().crates_untracked()) {
+            // FIXME: Due to #78696 rustdoc can query traits in scope for any crate root.
+            self.add_traits_in_scope(cnum.as_def_id());
+
+            let all_traits = Vec::from_iter(self.resolver.cstore().traits_in_crate_untracked(cnum));
+            let all_trait_impls =
+                Vec::from_iter(self.resolver.cstore().trait_impls_in_crate_untracked(cnum));
+
+            // Querying traits in scope is expensive so we try to prune the impl and traits lists
+            // using privacy, private traits and impls from other crates are never documented in
+            // the current crate, and links in their doc comments are not resolved.
+            for &def_id in &all_traits {
+                if self.resolver.cstore().visibility_untracked(def_id) == Visibility::Public {
+                    self.add_traits_in_parent_scope(def_id);
+                }
+            }
+            for &(trait_def_id, impl_def_id, simplified_self_ty) in &all_trait_impls {
+                if self.resolver.cstore().visibility_untracked(trait_def_id) == Visibility::Public
+                    && simplified_self_ty.and_then(|ty| ty.def()).map_or(true, |ty_def_id| {
+                        self.resolver.cstore().visibility_untracked(ty_def_id) == Visibility::Public
+                    })
+                {
+                    self.add_traits_in_parent_scope(impl_def_id);
+                }
+            }
 
             self.all_traits.extend(all_traits);
-            self.all_trait_impls.extend(all_trait_impls.into_iter().map(|(def_id, _)| def_id));
+            self.all_trait_impls.extend(all_trait_impls.into_iter().map(|(_, def_id, _)| def_id));
         }
     }
 
     fn load_links_in_attrs(&mut self, attrs: &[ast::Attribute], span: Span) {
-        // FIXME: this needs to consider export inlining.
+        // FIXME: this needs to consider reexport inlining.
         let attrs = clean::Attributes::from_ast(attrs, None);
         for (parent_module, doc) in attrs.collapsed_doc_value_by_module_level() {
             let module_id = parent_module.unwrap_or(self.current_mod.to_def_id());
+
+            self.add_traits_in_scope(module_id);
 
             for link in markdown_links(&doc.as_str()) {
                 let path_str = if let Some(Ok(x)) = preprocess_link(&link) {
@@ -85,6 +153,26 @@ impl IntraLinkCrateLoader<'_, '_> {
             }
         }
     }
+
+    /// When reexports are inlined, they are replaced with item which they refer to, those items
+    /// may have links in their doc comments, those links are resolved at the item definition site,
+    /// so we need to know traits in scope at that definition site.
+    fn process_module_children_or_reexports(&mut self, module_id: DefId) {
+        if !self.visited_mods.insert(module_id) {
+            return; // avoid infinite recursion
+        }
+
+        for child in self.resolver.module_children_or_reexports(module_id) {
+            if child.vis == Visibility::Public {
+                if let Some(def_id) = child.res.opt_def_id() {
+                    self.add_traits_in_parent_scope(def_id);
+                }
+                if let Res::Def(DefKind::Mod, module_id) = child.res {
+                    self.process_module_children_or_reexports(module_id);
+                }
+            }
+        }
+    }
 }
 
 impl Visitor<'_> for IntraLinkCrateLoader<'_, '_> {
@@ -93,6 +181,7 @@ impl Visitor<'_> for IntraLinkCrateLoader<'_, '_> {
             let old_mod = mem::replace(&mut self.current_mod, self.resolver.local_def_id(item.id));
 
             self.load_links_in_attrs(&item.attrs, item.span);
+            self.process_module_children_or_reexports(self.current_mod.to_def_id());
             visit::walk_item(self, item);
 
             self.current_mod = old_mod;
--- a/compiler/rustc_metadata/src/rmeta/decoder.rs
+++ b/compiler/rustc_metadata/src/rmeta/decoder.rs
@@ -1373,11 +1373,15 @@ impl<'a, 'tcx> CrateMetadataRef<'a> {
         self.root.traits.decode(self).map(move |index| self.local_def_id(index))
     }
 
-    fn get_trait_impls(self) -> impl Iterator<Item = (DefId, Option<SimplifiedType>)> + 'a {
-        self.cdata.trait_impls.values().flat_map(move |impls| {
-            impls
-                .decode(self)
-                .map(move |(idx, simplified_self_ty)| (self.local_def_id(idx), simplified_self_ty))
+    fn get_trait_impls(self) -> impl Iterator<Item = (DefId, DefId, Option<SimplifiedType>)> + 'a {
+        self.cdata.trait_impls.iter().flat_map(move |((trait_cnum_raw, trait_index), impls)| {
+            let trait_def_id = DefId {
+                krate: self.cnum_map[CrateNum::from_u32(*trait_cnum_raw)],
+                index: *trait_index,
+            };
+            impls.decode(self).map(move |(impl_index, simplified_self_ty)| {
+                (trait_def_id, self.local_def_id(impl_index), simplified_self_ty)
+            })
         })
     }
 
--- a/compiler/rustc_metadata/src/rmeta/decoder/cstore_impl.rs
+++ b/compiler/rustc_metadata/src/rmeta/decoder/cstore_impl.rs
@@ -493,7 +493,7 @@ impl CStore {
     pub fn trait_impls_in_crate_untracked(
         &self,
         cnum: CrateNum,
-    ) -> impl Iterator<Item = (DefId, Option<SimplifiedType>)> + '_ {
+    ) -> impl Iterator<Item = (DefId, DefId, Option<SimplifiedType>)> + '_ {
         self.get_crate_data(cnum).get_trait_impls()
     }
 }
--- a/src/librustdoc/core.rs
+++ b/src/librustdoc/core.rs
@@ -4,9 +4,9 @@ use rustc_errors::emitter::{Emitter, EmitterWriter};
 use rustc_errors::json::JsonEmitter;
 use rustc_feature::UnstableFeatures;
 use rustc_hir::def::Res;
-use rustc_hir::def_id::{DefId, LocalDefId};
+use rustc_hir::def_id::{DefId, DefIdMap, LocalDefId};
 use rustc_hir::intravisit::{self, Visitor};
-use rustc_hir::{HirId, Path};
+use rustc_hir::{HirId, Path, TraitCandidate};
 use rustc_interface::interface;
 use rustc_middle::hir::nested_filter;
 use rustc_middle::middle::privacy::AccessLevels;
@@ -33,6 +33,9 @@ use crate::passes::{self, Condition::*};
 crate use rustc_session::config::{DebuggingOptions, Input, Options};
 
 crate struct ResolverCaches {
+    /// Traits in scope for a given module.
+    /// See `collect_intra_doc_links::traits_implemented_by` for more details.
+    crate traits_in_scope: DefIdMap<Vec<TraitCandidate>>,
     crate all_traits: Option<Vec<DefId>>,
     crate all_trait_impls: Option<Vec<DefId>>,
 }
@@ -67,11 +70,6 @@ crate struct DocContext<'tcx> {
     crate auto_traits: Vec<DefId>,
     /// The options given to rustdoc that could be relevant to a pass.
     crate render_options: RenderOptions,
-    /// The traits in scope for a given module.
-    ///
-    /// See `collect_intra_doc_links::traits_implemented_by` for more details.
-    /// `map<module, set<trait>>`
-    crate module_trait_cache: FxHashMap<DefId, FxHashSet<DefId>>,
     /// This same cache is used throughout rustdoc, including in [`crate::html::render`].
     crate cache: Cache,
     /// Used by [`clean::inline`] to tell if an item has already been inlined.
@@ -350,7 +348,6 @@ crate fn run_global_ctxt(
         impl_trait_bounds: Default::default(),
         generated_synthetics: Default::default(),
         auto_traits,
-        module_trait_cache: FxHashMap::default(),
         cache: Cache::new(access_levels, render_options.document_private),
         inlined: FxHashSet::default(),
         output_format,
--- a/src/librustdoc/passes/collect_intra_doc_links.rs
+++ b/src/librustdoc/passes/collect_intra_doc_links.rs
@@ -16,7 +16,7 @@ use rustc_middle::ty::{DefIdTree, Ty, TyCtxt};
 use rustc_middle::{bug, span_bug, ty};
 use rustc_resolve::ParentScope;
 use rustc_session::lint::Lint;
-use rustc_span::hygiene::{MacroKind, SyntaxContext};
+use rustc_span::hygiene::MacroKind;
 use rustc_span::symbol::{sym, Ident, Symbol};
 use rustc_span::{BytePos, DUMMY_SP};
 use smallvec::{smallvec, SmallVec};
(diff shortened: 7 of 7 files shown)


jyn514:
@bors try @rust-timer queue

Since encoding and decoding are performance sensitive.

compiler/rustc_data_structures/src/stable_hasher.rs
```diff
@@ -209,6 +209,12 @@ impl_stable_hash_via_hash!(i128);
 impl_stable_hash_via_hash!(char);
 impl_stable_hash_via_hash!(());
 
+impl<CTX> HashStable<CTX> for ! {
+    fn hash_stable(&self, _ctx: &mut CTX, _hasher: &mut StableHasher) {
+        unreachable!()
```
bjorn3, line 214: `match *self {}`?

petrochenkov: Looks like standard library impls for `!` use just `*self`.

src/librustdoc/passes/collect_intra_doc_links/early.rs
```diff
@@ -1,75 +1,148 @@
...
+                );
+            }
+        }
+    }
+
+    /// Add traits in scope for links in impls collected by the `collect-intra-doc-links` pass.
+    /// That pass filters impls using type-based information, but we don't yet have such
+    /// information here, so we just conservatively calculate traits in scope for *all* modules
+    /// having impls in them.
+    fn add_foreign_traits_in_scope(&mut self) {
```
jyn514, line 66: nit: since it's wrong for this function to ever be called more than once, do you think it would make sense to inline it into `load_intra_link_crates`?

src/librustdoc/passes/collect_intra_doc_links/early.rs
```diff
@@ -1,75 +1,148 @@
...
+        for cnum in self.resolver.cstore().crates_untracked() {
+            // For some reason rustdoc also wants to know traits in scope for every crate root.
+            self.add_traits_in_scope(cnum.as_def_id());
```
jyn514, line 69: Hmm, that's surprising to me - the only way the module should be from an external crate is if the item being documented was inlined from that crate. So it shouldn't need *each* crate, only the ones where an item is inlined.

In practice it doesn't really matter, because `IntraLinkCrateLoader` doesn't know which items are inlined. But it would be nice for the comment to be correct.

```suggestion
            // Add the traits in the crate root, since an item defined in the root might be inlined to the current crate.
            self.add_traits_in_scope(cnum.as_def_id());
```

src/librustdoc/passes/collect_intra_doc_links/early.rs
```diff
@@ -1,75 +1,148 @@
...
+                    Namespace::TypeNS,
+                    module_id,
+                );
+            }
+        }
+    }
+
+    /// Exports are replaced with item which they refer to, those items may have links in their
+    /// doc comments, those links are resolved at the item definition site, so we need to know
+    /// traits in scope at that definition site.
```
jyn514, line 107: This comment is not quite right: `use` statements are only replaced with the item to which they refer if they're inlined. The behavior looks right though, since this pass doesn't know which items are inlined and which aren't.
```suggestion
    /// When items are inlined, their links are resolved at the item definition site, so we need to know
    /// traits in scope at that definition site.
```

src/librustdoc/passes/collect_intra_doc_links/early.rs
```diff
@@ -1,75 +1,148 @@
...
+    fn process_exports(&mut self, module_id: DefId) {
+        if !self.visited_mods.insert(module_id) {
+            return; // avoid infinite recursion
+        }
+
+        for export in self.resolver.module_exports(module_id) {
+            if export.vis == Visibility::Public {
```
jyn514, line 114: Hmm, this uses syntactic visibility, right? Would it be better to check if it's reachable from other crates?

Also, I'm not sure this is correct when `--document-private-items` is passed - can you add a test case for that? It would be something like:
```rust
mod inner {
    trait Trait {}
    /// Link to [Trait]
    fn foo() {}
}

File: src/librustdoc/passes/collect_intra_doc_links/early.rs [hunk 6]
```diff
@@ -1,75 +1,148 @@
...
             self.load_links_in_attrs(&item.attrs, item.span);
-            ast::visit::walk_item(self, item);
+            visit::walk_item(self, item);
         }
     }
+
+    fn visit_assoc_item(&mut self, item: &ast::AssocItem, ctxt: AssocCtxt) {
+        self.load_links_in_attrs(&item.attrs, item.span);
+        visit::walk_assoc_item(self, item, ctxt)
+    }
```
jyn514, line 147: Ooh good catch, this would be good to land even if we weren't precalculating traits.

compiler/rustc_resolve/src/build_reduced_graph.rs
```diff
@@ -115,6 +115,15 @@ impl<'a> Resolver<'a> {
         self.get_module(parent_id)
     }
 
+    /// Tells whether `get_module` below can be used on this `DefId`, used by rustdoc.
+    pub fn can_get_module(&self, def_id: DefId) -> bool {
```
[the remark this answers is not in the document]
petrochenkov: This is a separate compiler change that I didn't want to do in a rustdoc PR, but I'll probably make it separately before this PR lands, similarly to #88677.

src/librustdoc/passes/collect_intra_doc_links/early.rs
petrochenkov: Items cannot be inlined without exports (`use`), right?
After adding the export processing code that should cover inlining I tried to remove this line, and majority of tests in the `rustdoc` test suite failed.
I didn't figure out why this happens, but it seems to be not inlining.
Maybe it's `crate::foo` paths in impls collected by `collect-intra-doc-links`, I remember seeing some logic about that, need to investigate.

src/librustdoc/passes/collect_intra_doc_links/early.rs
petrochenkov: This is probably the source of CI failures, I didn't consider the `--document-private-items` mode at all.

src/librustdoc/passes/collect_intra_doc_links/early.rs
jyn514: >  Maybe it's crate::foo paths in impls collected by collect-intra-doc-links, I remember seeing some logic about that, need to investigate.

Oh, this is probably https://github.com/rust-lang/rust/issues/78696. Fixing it is waiting on https://github.com/rust-lang/rust/pull/78797#issuecomment-731538913 IIRC.

src/librustdoc/passes/collect_intra_doc_links/early.rs
jyn514: I think it would be enough to just remove the `if vis == Public` check.

compiler/rustc_resolve/src/build_reduced_graph.rs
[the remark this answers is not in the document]
petrochenkov: Done in https://github.com/rust-lang/rust/pull/88872.

petrochenkov:
Blocked on https://github.com/rust-lang/rust/pull/88677 and https://github.com/rust-lang/rust/pull/88679.

mati865:
Typo? #88679 is this PR.

src/librustdoc/passes/collect_intra_doc_links/early.rs
petrochenkov: Added a comment referencing https://github.com/rust-lang/rust/issues/78696.

src/librustdoc/passes/collect_intra_doc_links/early.rs
petrochenkov: I've left it as is for now, even if `fn add_foreign_traits_in_scope` is called once, it runs its own isolated piece of logic, and `fn load_intra_link_crates` therefore contains only "top level" calls.

src/librustdoc/passes/collect_intra_doc_links/early.rs
petrochenkov: The CI failure was due to a removed import, not rustdoc logic.
I've run the whole `x.py test --stage 2` with `compiler-docs = true`, which documents private items, and didn't encountered any panics.
So given that `fn build_module` doesn't inline reexports without `ty::Visibility::Public`, the current logic seems to be correct.

(The test snippet in https://github.com/rust-lang/rust/pull/88679#discussion_r703055115 is not relevant to this code because it doesn't contain associated items in links or reexports.)

compiler/rustc_resolve/src/build_reduced_graph.rs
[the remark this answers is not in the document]
petrochenkov: Ok, I think it's more important to land this PR sooner than to wait for #88872.
I'll remove `fn can_get_module` in that PR if this PR lands sooner.

jyn514 (approved):

jyn514:
@bors r-

Ugh, this is neglecting rustdoc times again. There's about a 7% regression on all crates. Let me investigate.

jyn514:
Here's the cachegrind diff (using the instructions from https://github.com/rust-lang/rust/pull/87244#issuecomment-920538189):

<details>

```
 cg_diff cgout-{before,after}-regression-31157-Doc-Full | sed 's@e4828d5b7f745a9e867a9b0cc7f080f287bcf55d@@' | sed 's@eec63217b0c2bd2bd82596fe0f8f6ccc5b9cc1a0@@' | rustfilt | cg_annotate /dev/stdin
--------------------------------------------------------------------------------
Files compared:   cgout-before-regression-31157-Doc-Full; cgout-after-regression-31157-Doc-Full
Command:          /home/joshua/.local/lib/rustup/toolchains//bin/rustdoc --crate-type lib --crate-name parser src/lib.rs -o /home/joshua/.local/lib/cargo/target/doc --error-format=json --json=diagnostic-rendered-ansi -L dependency=/home/joshua/.local/lib/cargo/target/debug/deps --extern peruse=/home/joshua/.local/lib/cargo/target/debug/deps/libperuse-d114ff6eeca80a88.rmeta --crate-version 0.1.0 -Adeprecated -Aunknown-lints; /home/joshua/.local/lib/rustup/toolchains//bin/rustdoc --crate-type lib --crate-name parser src/lib.rs -o /home/joshua/.local/lib/cargo/target/doc --error-format=json --json=diagnostic-rendered-ansi -L dependency=/home/joshua/.local/lib/cargo/target/debug/deps --extern peruse=/home/joshua/.local/lib/cargo/target/debug/deps/libperuse-d114ff6eeca80a88.rmeta --crate-version 0.1.0 -Adeprecated -Aunknown-lints
Data file:        /dev/stdin
Events recorded:  Ir
Events shown:     Ir
Event sort order: Ir
Thresholds:       0.1
Include dirs:     
User annotated:   
Auto-annotation:  off

--------------------------------------------------------------------------------
Ir          
--------------------------------------------------------------------------------
166,969,656  PROGRAM TOTALS

--------------------------------------------------------------------------------
Ir          file:function
--------------------------------------------------------------------------------
31,502,496  ???:rustc_metadata::rmeta::decoder::<impl rustc_serialize::serialize::Decodable<rustc_metadata::rmeta::decoder::DecodeContext> for rustc_span::span_encoding::Span>::decode
 7,653,789  ???:rustc_metadata::rmeta::decoder::cstore_impl::<impl rustc_metadata::creader::CStore>::item_children_untracked
 6,913,556  ???:rustc_metadata::rmeta::decoder::<impl rustc_serialize::serialize::Decodable<rustc_metadata::rmeta::decoder::DecodeContext> for rustc_span::hygiene::SyntaxContext>::decode
 6,608,013  ???:rustc_resolve::imports::<impl rustc_resolve::Resolver>::try_define
 5,902,636  ???:indexmap::map::core::Entry<K,V>::or_insert_with
 5,805,920  ???:rustc_metadata::rmeta::decoder::<impl rustc_metadata::creader::CrateMetadataRef>::imported_source_files
 5,553,403  ???:rustc_resolve::build_reduced_graph::BuildReducedGraphVisitor::build_reduced_graph_for_external_crate_res
 5,047,864  ???:hashbrown::map::HashMap<K,V,S,A>::insert
 4,886,330  ???:hashbrown::raw::RawTable<T,A>::reserve_rehash
 4,742,352  ???:indexmap::map::IndexMap<K,V,S>::entry
 4,470,310  ???:<rustc_span::def_id::DefId as rustc_serialize::serialize::Decodable<D>>::decode
 4,223,606  ???:scoped_tls::ScopedKey<T>::with
 4,015,378  ???:rustc_resolve::build_reduced_graph::<impl rustc_resolve::Resolver>::define
 3,988,003  /rustc///library/core/src/str/validations.rs:core::str::converts::from_utf8
 3,711,738  ???:rustc_metadata::rmeta::decoder::<impl rustc_metadata::creader::CrateMetadataRef>::def_key
 3,604,660  ???:rustc_resolve::Resolver::resolutions
 3,230,864  ???:rustc_span::symbol::Ident::normalize_to_macros_2_0
 3,001,575  ???:hashbrown::raw::inner::RawTable<T,A>::reserve_rehash
 2,992,579  ???:rustc_middle::hir::exports::_DERIVE_rustc_serialize_Decodable_D_FOR_Export::<impl rustc_serialize::serialize::Decodable<__D> for rustc_middle::hir::exports::Export>::decode
 2,842,145  ???:rustc_metadata::rmeta::table::<impl rustc_metadata::rmeta::Lazy<rustc_metadata::rmeta::table::Table<I,T>>>::get
 2,467,516  ???:core::lazy::OnceCell<T>::get_or_init
 2,388,161  ???:rustc_metadata::rmeta::_DERIVE_rustc_serialize_Decodable_DecodeContext_a_tcx_FOR_EntryKind::<impl rustc_serialize::serialize::Decodable<rustc_metadata::rmeta::decoder::DecodeContext> for rustc_metadata::rmeta::EntryKind>::decode
 2,352,269  ???:rustc_hir::def::_DERIVE_rustc_serialize_Decodable_D_FOR_DefKind::<impl rustc_serialize::serialize::Decodable<__D> for rustc_hir::def::DefKind>::decode
 2,233,897  ???:rustc_hir::def::_DERIVE_rustc_serialize_Decodable_D_FOR_Res::<impl rustc_serialize::serialize::Decodable<__D> for rustc_hir::def::Res<Id>>::decode
 2,144,697  ???:rustc_metadata::rmeta::decoder::<impl rustc_metadata::creader::CrateMetadataRef>::try_item_ident
 2,050,836  ???:rustc_resolve::Resolver::traits_in_module
 1,993,505  ???:rustc_metadata::rmeta::decoder::__ty_decoder_impl::<impl rustc_serialize::serialize::Decoder for rustc_metadata::rmeta::decoder::DecodeContext>::read_str
 1,976,679  ???:rustc_serialize::serialize::Decoder::read_option
 1,847,980  ???:rustc_middle::ty::_DERIVE_rustc_serialize_Decodable_D_FOR_Visibility::<impl rustc_serialize::serialize::Decodable<__D> for rustc_middle::ty::Visibility>::decode
 1,845,837  ???:<core::iter::adapters::map::Map<I,F> as core::iter::traits::iterator::Iterator>::fold
 1,660,508  ???:<(T10,T11) as rustc_serialize::serialize::Decodable<D>>::decode
 1,557,738  ???:rustc_span::hygiene::HygieneData::with
 1,401,608  ???:rustc_metadata::rmeta::decoder::DecodeContext::read_lazy_with_meta
 1,311,606  /build/glibc-eX1tMB/glibc-2.31/string/../sysdeps/x86_64/multiarch/memmove-vec-unaligned-erms.S:__memcpy_sse2_unaligned_erms
 1,096,224  ???:rustc_resolve::macros::<impl rustc_resolve::Resolver>::check_reserved_macro_name
 1,055,352  /build/glibc-eX1tMB/glibc-2.31/string/../sysdeps/x86_64/multiarch/memcmp-avx2-movbe.S:__memcmp_avx2_movbe
   836,472  ???:<core::iter::adapters::flatten::FlatMap<I,U,F> as core::iter::traits::iterator::Iterator>::next
   796,983  ???:rustdoc::passes::collect_intra_doc_links::early::IntraLinkCrateLoader::add_traits_in_scope
   714,315  ???:rustc_span::SESSION_GLOBALS::FOO::__getit
   710,028  ???:hashbrown::raw::RawTable<T,A>::insert
   678,640  /rustc///library/core/src/str/converts.rs:core::str::converts::from_utf8
   637,346  ???:rustc_metadata::rmeta::decoder::<impl rustc_metadata::creader::CrateMetadataRef>::get_span
   611,723  ???:<alloc::vec::Vec<T> as alloc::vec::spec_from_iter::SpecFromIter<T,I>>::from_iter
   600,778  ???:rustc_middle::ty::fast_reject::_DERIVE_rustc_serialize_Decodable_D_FOR_SimplifiedTypeGen::<impl rustc_serialize::serialize::Decodable<__D> for rustc_middle::ty::fast_reject::SimplifiedTypeGen<D>>::decode
   541,236  ???:rustc_middle::mir::query::_DERIVE_rustc_serialize_Decodable_D_FOR_ConstQualifs::<impl rustc_serialize::serialize::Decodable<__D> for rustc_middle::mir::query::ConstQualifs>::decode
   529,200  ???:rustc_metadata::rmeta::decoder::<impl rustc_metadata::rmeta::Lazy<T>>::decode
   439,146  /build/glibc-eX1tMB/glibc-2.31/string/../sysdeps/x86_64/multiarch/memset-vec-unaligned-erms.S:__memset_avx2_erms
   419,874  ???:rustc_metadata::rmeta::decoder::cstore_impl::<impl rustc_metadata::creader::CStore>::ctor_def_id_and_kind_untracked
   408,310  ???:rustc_hir::definitions::_DERIVE_rustc_serialize_Decodable_D_FOR_DefKey::<impl rustc_serialize::serialize::Decodable<__D> for rustc_hir::definitions::DefKey>::decode
   401,632  ???:<&rustc_resolve::Resolver as rustc_middle::ty::DefIdTree>::parent
   373,248  ???:<rustc_arena::TypedArena<T> as core::ops::drop::Drop>::drop
   369,502  /rustc///library/core/src/num/uint_macros.rs:core::str::converts::from_utf8
   339,320  ???:rustc_span::symbol::Symbol::intern
   331,632  ???:rustc_middle::ty::_DERIVE_rustc_serialize_Decodable_D_FOR_VariantDiscr::<impl rustc_serialize::serialize::Decodable<__D> for rustc_middle::ty::VariantDiscr>::decode
   323,634  ???:core::ops::function::impls::<impl core::ops::function::FnOnce<A> for &mut F>::call_once
   295,152  /rustc//obj/build/x86_64-unknown-linux-gnu/stage1-rustc/x86_64-unknown-linux-gnu/release/build/tikv-jemalloc-sys-fde45a9a697cc078/out/build/src/arena.c:_rjem_je_arena_ralloc
   274,104  ???:rustc_metadata::rmeta::decoder::cstore_impl::<impl rustc_middle::middle::cstore::CrateStore for rustc_metadata::creader::CStore>::def_key
   267,978  ???:rustc_hir::definitions::DefPathData::get_opt_name
  -262,298  ???:alloc::str::<impl str>::replace
   253,800  ???:rustc_metadata::rmeta::decoder::cstore_impl::<impl rustc_metadata::creader::CStore>::struct_field_names_untracked
   241,850  ???:core::str::pattern::TwoWaySearcher::next
   230,640  ???:rustdoc::passes::collect_intra_doc_links::early::load_intra_link_crates
   189,924  ???:rustc_metadata::rmeta::decoder::<impl rustc_serialize::serialize::Decodable<rustc_metadata::rmeta::decoder::DecodeContext> for rustc_span::def_id::CrateNum>::decode
   172,159  ???:core::ptr::drop_in_place<[rustdoc::clean::types::Item]>
```

</details>

So it looks like most of the time is being spent decoding Spans. Not sure why loading the DefIds is also loading the spans - maybe they're being decoded here in this parameter you ignore? Does it make sense to add a variant of this that only decodes the DefId and not the type?
https://github.com/rust-lang/rust/pull/88679/files#diff-03c3ed2f6e27e58df63382e00323eea6fd1ce3a4f9546941b2b717fead25e520R1398

petrochenkov:
Blocked on https://github.com/rust-lang/rust/pull/89059.

petrochenkov:
https://github.com/rust-lang/rust/pull/89059 has issues that will require time to resolve, so I'll add some caching to rustdoc specifically to move this forward.
I think staying under ~2% of regressions for most benchmarks would be acceptable, we need to address the architectural issue with resolver sooner.

petrochenkov:
I should be able to return to it this Saturday.

In October I got issues with making a local perf measuring setup on Windows 8.1, then I got busy for a few weeks and grew a huge review queue.
Now I've emptied the queue and accidentally got a new Windows 10 laptop, so perhaps it's time for continuing this work.

petrochenkov:
Almost all time of the `load_intra_link_crates` pass is spent in `add_foreign_traits_in_scope`, not many opportunities to optimize that.

I'll look in more detail when my Linux build is ready.
Starting benchmarking with the windows-gnu toolchain was the worst choice ever, no Windows perf tools really work without PDB-style debuginfo and `-Z self-profile` is not suitable for this kind of work.

petrochenkov:
Blocked on https://github.com/rust-lang/rust/pull/92244.

petrochenkov:
>I think staying under ~2% of regressions for most benchmarks would be acceptable

This target is much harder to meet now after the improvements in https://github.com/rust-lang/rust/pull/92244.
I suggest to just land this PR and continue the work on removing late uses of the resolver.

The main source of the slowness here is calls to `resolver.traits_in_scope(...)` which will decode the item's (trait or impl) parent modules and build module structures for them.

The solution is to prune the list of all traits and impls somehow to leave only those that will end up in documentation and with links resolved in their comments.
I tried to check for links in comments (see the logic commented out in https://github.com/rust-lang/rust/pull/88679/commits/e41e64f2d57a3cdc9dd5ebb3be88a7f701c09f3b), but doing that is even more expensive than calling `resolver.traits_in_scope(...)`.
Maybe some filter based on `pub`licity of the involved traits and impl types will work better.

I'm not actually sure under what exactly circumstances impls from other crates end up being documented in the docs for the current crate. I'll read the relevant code eventually, but if someone explains it before that it would be helpful.

petrochenkov:
(I'm going to split the rustc changes and preparatory refactorings in rustdoc to a separate PR(s).)

jyn514:
> I'm not actually sure under what exactly circumstances impls from other crates end up being documented in the docs for the current crate. I'll read the relevant code eventually, but if someone explains it before that it would be helpful.

I don't know without reading the code. @GuillaumeGomez, @jsha, or @Aaron1011 might have some idea but I suspect everyone who knows for sure has left the project.

camelid:
I don't know for sure, but some of them are:

* Impl is of a trait in the current crate
* Impl is for a type in the current crate
* Impl is of a trait re-exported in the current crate
* Impl is for a type re-exported in the current crate

Impls are also mentioned (though not documented) in the current crate if
* Return type of a function in the current crate has an impl of a trait that is notable (`doc(notable_trait)`)

Some of these may not even occur though; they're partially guesses.

petrochenkov:
>- Impl is of a trait in the current crate
>- Impl is for a type in the current crate

These two cannot come from other crates.

Also unmentioned:
- Blanket impls
- Are trait impls available through `Deref` documented somehow, or the "Methods from Deref" section is only for inherent methods?

petrochenkov:
Ok, privacy-based pruning seems slightly helpful.
@bors try @rust-timer queue

camelid:
> These two cannot come from other crates.

Why? I'm pretty sure they do show up sometimes.

petrochenkov:
@petrochenkov 
If a trait or a type is defined in the current crate, how can they be used in its dependencies?
It can only happen through blanket impls (impls for generic parameters), but not for the type itself.

camelid:
Traits or types defined in the current crate can be used in its _dependents_ (downstream crates). Some of those usages show up in the current crate's docs.

petrochenkov:
I'm not sure what you are talking about, when rustdoc is running the current crate is the root of the tree, there's no dependents.

camelid:
This is what I mean:

> Rustdoc uses JS to add items to a crate's docs if the crate's documentation has already been generated.

Impls can be added to a crate's documentation **even after rustdoc is run on that crate**.

petrochenkov:
Ah, I see, yes, this is not relevant to this PR.

petrochenkov:
Blocked on https://github.com/rust-lang/rust/pull/92607 and https://github.com/rust-lang/rust/pull/92608.

jyn514:
I don't want to be responsible for this PR.

r? @camelid (but feel free to reassign if you have too many PRs to review already).

petrochenkov:
This should be ready now.
r? @CraftSpider

GuillaumeGomez (approved):

GuillaumeGomez:
Looks good to me, thanks! Let's check the performance impact too.

@bors try @rust-timer queue

petrochenkov:
It's going to be more or less the same as the previous run (https://github.com/rust-lang/rust/pull/88679#issuecomment-1005820703) - +7-8% on helloworld-like programs, maybe slightly more because https://github.com/rust-lang/rust/pull/92608 contained at least one optimization.

GuillaumeGomez:
Doesn't hurt to check. ;)

GuillaumeGomez:
Wow, this is quite bad for both instructions and max RSS. Is it a bug? O.o

petrochenkov:
No, this is literally what I wrote in my previous comment.

The architectural issue with resolver will never be fixed if we cannot tolerate some performance hit, at least temporarily.

GuillaumeGomez:
It's quite a big one which is why I'm very surprised about this. It makes rustdoc slower and use more memory (this second part is actually very surprising). I'm not the most up-to-date to this part of the code. Maybe @camelid knows more about it?

camelid:
I understand that to solve the structural issues in the code, we need to take a (temporary?) performance hit; but so I understand better, what exactly is causing the slowdown? I can't imagine it's just crawling the crate because rustdoc already does that a ton.

compiler/rustc_resolve/src/lib.rs
```diff
@@ -3384,6 +3385,16 @@ impl<'a> Resolver<'a> {
         &self.all_macros
     }
 
+    // For rustdoc.
+    // For local modules returns only reexports, for external modules returns all children.
```
camelid, line 3389: Can you make this a doc comment?

GuillaumeGomez:
I'm more in favour of merging it if it's fine for you too @camelid ?

camelid:
> `resolver.traits_in_scope(...)` is expensive because it builds all the relevant module structures in resolver for the given module chain.

Aren't those structures already built while rustc's code (e.g., the type-checker) is running? Is the issue here that because rustdoc clones the resolver, the structures are being built twice? (So if rustdoc stops cloning the resolver, we'll get the performance back?)

> I'm more in favour of merging it if it's fine for you too camelid?

I haven't reviewed the code, but I'm good with the idea of this PR since it's a necessary refactoring step.

petrochenkov:
@camelid 
>Aren't those structures already built while rustc's code (e.g., the type-checker) is running?

No, those structures are built lazily when needed, and for majority modules in dependency crates they are not needed and not build, unless `traits_in_scope` requests them.

>Is the issue here that because rustdoc clones the resolver, the structures are being built twice?

No, see the previous answer.

 >So if rustdoc stops cloning the resolver, we'll get the performance back?

Unfortunately not.

petrochenkov:
The fundamental issue of rustdoc is that unlike rustc it has to "compile dependency crates again" to retrieve information, while rustc just takes the pre-compiled results from rlibs.

So the fundamental solution for the doc link resolution slowness (and correctness) is to resolve them in rustc and store the results in rlibs, that would probably speed up rustdoc more than https://github.com/rust-lang/rust/pull/92608.
Unfortunately that would slow down rustc (but maybe only slightly, then that may be ok), and made rustc dependent on some markdown parsing library too, for extracting links.

camelid:
> No, those structures are built lazily when needed, and for majority modules in dependency crates they are not needed and not build, unless traits_in_scope requests them.

> The fundamental issue of rustdoc is that unlike rustc it has to "compile dependency crates again" to retrieve information, while rustc just takes the pre-compiled results from rlibs.

Thanks for the explanation. I understand a bit better, but I'm still a bit confused: don't uses of paths like `MyStruct::my_trait_fn` in source code have to invoke `traits_in_scope` as well? So then the `traits_in_scope` code will have already been run for many parts of the code?

I'm not quite sure what you mean about compiling dependency crates again; could you elaborate on what parts of rustdoc do this?

Perhaps it's best to continue this discussion on Zulip, though. @GuillaumeGomez if you're ready to approve it, go ahead.

GuillaumeGomez:
Let's get the full explanation first. I don't think we're in a hurry here. :)

petrochenkov:
@camelid 
>don't uses of paths like `MyStruct::my_trait_fn` in source code have to invoke `traits_in_scope` as well?

They do.

>So then the `traits_in_scope` code will have already been run for many parts of the code?

Yes, but only for code in the current crate.
The slowness come from calling `traits_in_scope` for code from other crates (dependency tree of the current crate).
Rustc doesn't call `traits_in_scope` for any of that code, because unlike rustdoc it compiles only the current crate, not the whole dependency tree.

>I'm not quite sure what you mean about compiling dependency crates again; could you elaborate on what parts of rustdoc do this?

Rustdoc does doc inlining from dependency crates (and doc generation for things like blanket impls from dependency crates). therefore it has to "compile" (in some sense) everything it inlines and documents.
In particular it has to resolve paths in doc links from other crates, and resolve them in the context of their original external location.
Rustc would never do that, it would resolve the paths when compiling the crate they are used in, and then write the result into metadata to be readily used when compiling dependent crates.

camelid:
Ah, I was afraid this would have to do with inlining. 😆 

Would it be possible to save the results of `traits_in_scope` in metadata or would it be too big? Another idea (unless this is already the case): could rustdoc only run `traits_in_scope` for inlined code? Perhaps that'd be a bit tricky to figure out ahead-of-time though.

petrochenkov:
>Would it be possible to save the results of `traits_in_scope` in metadata or would it be too big?

This https://github.com/rust-lang/rust/pull/88679#issuecomment-1021085756 is the result of implementing that suggestion.
So it's possible to significantly reduce the overhead this way.

I'll save the optimization on a branch, but I don't want to land it now, because this PR is an intermediate step and I want to finish the whole resolver cloning removal before starting to encode more data for rustdoc, maybe we'll have different bottlenecks and need a different set of data in the end.

camelid:
I still feel a bit uncomfortable with landing such a large regression, but I won't block this since I understand that it's an important refactoring. I'll let Guillaume take it from here.

GuillaumeGomez (approved):

GuillaumeGomez:
@petrochenkov Let's go with this then. Can you open an issue with the next steps please (with explanations too, it'd be very much appreciated!). Once done, r=me. Thanks a lot for working on this!

petrochenkov:
@GuillaumeGomez 
https://github.com/rust-lang/rust/issues/83761 is the issue, https://github.com/rust-lang/rust/issues/83761#issuecomment-1021809921 is the next step.

@bors r=GuillaumeGomez

trevyn:
@petrochenkov FYI, this seems to have broken one of my nightly CI runs, not sure if that's expected:

https://github.com/trevyn/turbosql/runs/4960657119?check_suite_focus=true

```
 Documenting turbosql-impl v0.4.0 (/home/runner/work/turbosql/turbosql/turbosql-impl)
Error: thread 'rustc' panicked at 'no entry found for key', src/librustdoc/passes/collect_intra_doc_links.rs:929:16

error: could not document `turbosql-impl`
```

petrochenkov:
@trevyn 
Could you make a minimized self-contained reproduction?
I'll look at this tomorrow, and having such a reproduction would make it much faster.

trevyn:
Ok, apparently this is enough to trigger it on my side, not sure I feel like minimizing `rusqlite` right now:

Cargo.toml:
```toml
[package]
edition = "2018"
name = "test"
version = "0.1.0"

[dependencies]
rusqlite = "0.26"
```

lib.rs:
```rust
use rusqlite;
```

jyn514:
@petrochenkov https://github.com/rust-lang/rust/issues/93428

petrochenkov:
The panic in https://github.com/rust-lang/rust/pull/88679#issuecomment-1023929480 happens because an *inherent* impl from `rusqlite` (`impl Connection` from `mod busy`) gets inlined into the current crate.

How is that even possible?
If I remove the doc link to make the documentation process to succeed, then that impl is not even in the generated documentation.

camelid:
I think it would happen if `Connection` was inlined. Is `Connection` inlined?

petrochenkov:
Maybe it's inlined internally, but it doesn't end up in the generated doc.
The documented crate in https://github.com/rust-lang/rust/pull/88679#issuecomment-1023929480 is just a single private import (it doesn't reexport `Connection` in particular), nothing should be inlined in that case.

camelid:
Weird. I have no idea why that impl is being inlined then.

petrochenkov:
I didn't figure out why that specific impl was inlined, but there are legitimate cases where inherent impls are inlined correctly, but traits in scope are not collected for them:
```rust
// Dependency crate

#[derive(Clone)]
pub struct PublicStruct;

mod inner {
    use super::PublicStruct;

    impl PublicStruct {
        /// [PublicStruct::clone]
        pub fn method() {}
    }
}
```
```rust
// Main crate

pub use dependency::PublicStruct;
```

petrochenkov:
Both cases (https://github.com/rust-lang/rust/pull/88679#issuecomment-1023929480 and https://github.com/rust-lang/rust/pull/88679#issuecomment-1026520300) are fixed in https://github.com/rust-lang/rust/pull/93539.

Issue #87681, opened 2021-08-01
Module path attribute override with nested macros doesnt work


Veetaha:
I tried this code:

[Repository with the reproduction](https://github.com/Veetaha/path-nested-macro-bug-repro)

I expected to see this happen: *explanation*

The code should compile without errors, because this feature was advertised in Rust 1.54 announcement and in https://github.com/rust-lang/rust/pull/83366.

Instead, this happened: *explanation*

Compile error `file not found for module generated` was issued

More info in README.md of the reproduction repo

### Meta

`rustc --version --verbose`:
```
rustc 1.54.0 (a178d0322 2021-07-26)
binary: rustc
commit-hash: a178d0322ce20e33eac124758e837cbd80a6f633
commit-date: 2021-07-26
host: x86_64-unknown-linux-gnu
release: 1.54.0
LLVM version: 12.0.1
```

cc @jyn514

ehuss:
I can confirm this doesn't work, and is a bit unfortunate it slipped through (I admit I didn't test it).

From what I can tell, the extended `attr=expr` expansion happens during lowering to HIR (maybe? I'm not sure), but `#[path]` is needed during expansion before that happens. The check [here](https://github.com/rust-lang/rust/blob/8d57c0ab2b2e1aae07c1b8638358fb7a909940bc/compiler/rustc_expand/src/module.rs#L172) silently ignores the attribute because the value is not a string (it is a macro expression).

I don't think there is any true support for eager expansion, so I don't think there is any existing support for something like this.  At a bare minimum, the `#[path]` attribute should generate an error if the value is not a string literal. 

I think we should maybe update the release post, since it is misleading.

danobi:
@ehuss do you think adding support for this eager expansion is feasible / welcome? I'd be interested in hacking on it if it's not terribly difficult.

ehuss:
Unfortunately I think it might be difficult to support.  `#[path]` would likely need special handling because it is needed during expansion.  I am not on the compiler or language teams, so I'm not the one to say definitively.

Mark-Simulacrum:
cc @petrochenkov @jyn514 -- I would have expected this to work based on discussions when landing this feature, so I'm a little surprised that it doesn't.

petrochenkov:
@Mark-Simulacrum 
>I would have expected this to work based on discussions when landing this feature, so I'm a little surprised that it doesn't.

[...]

PR #83744, opened 2021-04-01
Deprecate crate_type and crate_name nested inside #![cfg_attr]

bjorn3:
This implements the proposal in https://github.com/rust-lang/rust/pull/83676#issuecomment-811213956, with a future compatibility lint imposed on usage of crate_type/crate_name inside cfg's.

This is a compromise between removing `#![crate_type]` and `#![crate_name]` completely and keeping them as a whole, which requires somewhat of a hack in rustc and is impossible to support by gcc-rust. By only removing `#![crate_type]` and `#![crate_name]` nested inside `#![cfg_attr]` it becomes possible to parse them before a big chunk of the compiler has started.

Replaces https://github.com/rust-lang/rust/pull/83676

```rust
#![crate_type = "lib"] // remains working
#![cfg_attr(foo, crate_type = "bin")] // will stop working
```

# Rationale

As it currently is it is possible to try to access the stable crate id before it is actually set, which will panic. The fact that the Session contains mutable state beyond debugging things also doesn't completely sit well with me. Especially once parallel rustc becomes the default.

I think there is currently also a cyclic dependency where you need to set the stable crate id to be able to load crates, but you need to load crates to expand proc macro attributes that may define #![crate_name] or #![crate_type]. Currently crate level proc macro attributes are unstable or completely unsupported (can't remember which), so this is not a problem, but it may become an issue in the future.

Finally if we want to add incremental compilation to macro expansion or even parsing, we need the StableCrateId to be created together with the Session or even earlier as incremental compilation determines the incremental compilation session dir based on the StableCrateId.

PR #65342, opened 2019-10-12
Revert "Auto merge of #64823 - cuviper:min-std, r=Mark-Simulacrum"

Mark-Simulacrum:
This reverts commit 000d90b11f7be70ffb7812680f7abc6deb52ec88, reversing
changes made to 898f36c83cc28d7921a1d7b3605323dc5cfcf533.

This turned out to break quite a few targets, cc https://github.com/rust-lang/rust/issues/65335

PR #88680, opened 2021-09-06
Validate builtin attributes for macro args.

ehuss:
This adds some validation for `path`, `crate_type`, and `recursion_limit` attributes so that they will now return an error if you attempt to pass a macro into them (such as `#[path = foo!()]`). Previously, the attribute would be completely ignored. These attributes are special because their values need to be known before/during expansion.

cc #87681

### DIFF
--- a/compiler/rustc_interface/src/util.rs
+++ b/compiler/rustc_interface/src/util.rs
@@ -10,6 +10,7 @@ use rustc_errors::registry::Registry;
 use rustc_metadata::dynamic_lib::DynamicLibrary;
 #[cfg(parallel_compiler)]
 use rustc_middle::ty::tls;
+use rustc_parse::validate_attr;
 #[cfg(parallel_compiler)]
 use rustc_query_impl::QueryCtxt;
 use rustc_resolve::{self, Resolver};
@@ -475,7 +476,7 @@ pub fn get_codegen_sysroot(
 }
 
 pub(crate) fn check_attr_crate_type(
-    _sess: &Session,
+    sess: &Session,
     attrs: &[ast::Attribute],
     lint_buffer: &mut LintBuffer,
 ) {
@@ -515,6 +516,19 @@ pub(crate) fn check_attr_crate_type(
                         );
                     }
                 }
+            } else {
+                // This is here mainly to check for using a macro, such as
+                // #![crate_type = foo!()]. That is not supported since the
+                // crate type needs to be known very early in compilation long
+                // before expansion. Otherwise, validation would normally be
+                // caught in AstValidator (via `check_builtin_attribute`), but
+                // by the time that runs the macro is expanded, and it doesn't
+                // give an error.
+                validate_attr::emit_fatal_malformed_builtin_attribute(
+                    &sess.parse_sess,
+                    a,
+                    sym::crate_type,
+                );
             }
         }
     }
--- a/compiler/rustc_expand/src/module.rs
+++ b/compiler/rustc_expand/src/module.rs
@@ -3,6 +3,7 @@ use rustc_ast::ptr::P;
 use rustc_ast::{token, Attribute, Inline, Item};
 use rustc_errors::{struct_span_err, DiagnosticBuilder};
 use rustc_parse::new_parser_from_file;
+use rustc_parse::validate_attr;
 use rustc_session::parse::ParseSess;
 use rustc_session::Session;
 use rustc_span::symbol::{sym, Ident};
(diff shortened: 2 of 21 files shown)


ehuss:
I'm not too happy with this change, as it seems to add some specific validation that would probably better be part of something more general. However, attribute validation happens long after expansion, and I'm not aware of any pre-expansion validation where this could live.

petrochenkov:
We actually validate some built-in attributes during expansion, those that are applied to macro calls and disappear during expansion.
```rust
macro_rules! m { () => () }

#[path] // error: malformed `path` attribute input
m!();

fn main() {}
```
This happens in `expand.rs / fn check_attributes`.

You can also reuse `validate_attr::check_meta` if you want to get the same wording as during regular attribute validation.

compiler/rustc_interface/src/util.rs
```diff
@@ -528,6 +528,18 @@ pub(crate) fn check_attr_crate_type(
                         );
                     }
                 }
+            } else {
+                // This is here mainly to check for using a macro, such as
+                // #![crate_type = foo!()]. That is not supported since the
+                // crate type needs to be known very early in compilation long
+                // before expansion. Otherwise, validation would normally be
```
petrochenkov, line 535: >crate type needs to be known very early in compilation long before expansion

Why tho?
This information seems irrelevant to parser or macro expander.

petrochenkov:
This seems fine.
The point is to run validation before the attribute is used (that's why most uses simply ignore ill-formed attributes).
Since most attributes are used late, the validation is run soon after expansion, but for a few attributes that are used early the validation can be done immediately before use, don't see problem in that.

compiler/rustc_interface/src/util.rs
petrochenkov: Ah, I see `StableCrateId::new` uses the crate types, although the use seems incorrect.

Otherwise it's only used before proc macro harness generation (to determine whether generate it or not), which happens only after all macros are expanded.

In other words, this is something that needs to be investigated.

petrochenkov: cc @bjorn3 
This is related to https://github.com/rust-lang/rust/pull/83744 (for which I don't see much motivation).

bjorn3: > Ah, I see StableCrateId::new uses the crate types, although the use seems incorrect.

This is necessary for backwards compatibility. Without it the following would give an error about a duplicate `StableCrateId`:

```bash
$ echo "" | rustc - --crate-type lib --crate-name foo
$ echo "extern crate foo; fn main() {}" | rustc - --crate-type bin --crate-name foo -L .
```

the error would be (forced it by using lib as crate type in both cases):

```
error[E0519]: the current crate is indistinguishable from one of its dependencies: it has the same crate-name `foo` and was compiled with the same `-C metadata` arguments. This will result in symbol conflicts between the two.
 --> <anon>:1:1
  |
1 | extern crate foo; fn main() {}
  | ^^^^^^^^^^^^^^^^^

```

I would personally be happy if it were to be turned into an error as it complicates things, but it is a breaking change.

> Otherwise it's only used before proc macro harness generation (to determine whether generate it or not), which happens only after all macros are expanded.

Currently expansion has to happen before the `StableCrateId` can be set in the `Session` due to the dependency on expansion to determine all `#![crate_type = "..."]`. Once the future compat lint in #83744 becomes a hard error, it becomes possible to determine the final crate name and crate type for certain without any expansion by only parsing the crate level attributes, thus allowing the `StableCrateId` to be set immediately while creating the `Session`.

petrochenkov: >allowing the `StableCrateId` to be set immediately while creating the `Session`

Are there specific reasons to want `StableCrateId` be created together with the `Session`?

ehuss:
I'm not sure if you meant that it should check in `InvocationCollector`, but I liked the idea so I ran with it.  WDYT?

I also fixed a minor issue with the suggested fix not checking if it was an inner attribute.

compiler/rustc_interface/src/util.rs
petrochenkov: @bjorn3 
Could you copy this motivation into the description of https://github.com/rust-lang/rust/pull/83744?

petrochenkov:
Nah, I only used `InvocationCollector` as an example, showing that we are already checking attributes early in some cases.

I liked the initial version of the PR better, TBH.
I'd just suggest to return to it, only use `validate_attr::check_meta` instead of hand-rolled checking.

petrochenkov:
The inner attribute fix and moving the `emit_malformed_attribute` logic to a separate function are good though.

compiler/rustc_interface/src/util.rs
bjorn3: Done

ehuss:
>  I'd just suggest to return to it, only use `validate_attr::check_meta` instead of hand-rolled checking.

Sure, no problem.  Unfortunately, that means adding a dependency from rustc_middle to rustc_parse, which does not seem good to me?

bjorn3:
Can `validate_attr::check_meta` be moved to rustc_attr?

petrochenkov:
Or `validate_attr::check_meta` can be called in `fn configure_and_expand` at call site of `get_recursion_limit`.

ehuss:
> Or `validate_attr::check_meta` can be called in `fn configure_and_expand` at call site of `get_recursion_limit`.

Sure, I updated it to do the check in `configure_and_expand`.

ehuss:
@bors retry

Error in `test_cat::prop_cat_rows` is xsv, which seems unlikely to be related here. A similar issue was noted a few years ago at https://github.com/rust-lang/rust/pull/65342#issuecomment-541349793

alex-berger:
The following code used to compile up-to-and-including `nightly-2021-09-26`.

```rust
#[path = concat!(env!("OUT_DIR"), "/v1_22/mod.rs")]
mod v1_22;
```

And starting with `nightly-2021-09-27` it fails with:

```
error: malformed `path` attribute input
 --> ...
  |
2 | #[path = concat!(env!("OUT_DIR"), "/v1_22/mod.rs")]
  | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ help: must be of the form: `#[path = "file"]`
```

Is this intended?

ehuss:
The effect before was that the `#[path]` attribute was ignored.

Hm, so my expectation was that nobody would be using that, since the attribute is ignored and you would need a `v1_22.rs` file in the same directory to avoid a compilation error (and I would suspect most people would notice that it was using the wrong file).  Did you maybe have a `v1_22.rs` or `v1_22/mod.rs` file that was being used instead?

We could maybe downgrade it to a future-incompatibility warning if people feel this was too aggressive.

alex-berger:
@ehuss indeed, there is a `$OUT_SIR/v1_22/model.rs` in place and I guess this is why it used to work. 

Personally, I am fine with this change, just wanted to understand why it used to work.

PR #88681, opened 2021-09-06
Check for duplicate attributes.

### LINKED ISSUE #55112: emit unused_attributes lint for silly attributes
We already have an `UnusedAttributes` late lint pass for attributes that don't get [marked as having been used somewhere](https://github.com/rust-lang/rust/blob/df0d6ad9c1510630b234f6bb96a69a081de0b307/src/libsyntax/attr/mod.rs#L42-L47), but recent discussion has touched on the idea of also directly (during normal complication, not as a lint pass) issuing the lint for attributes that don't make much sense (but which we don't want to go to the length of declaring errors, _e.g._, for backwards-compatibility reasons). These include:
 - [x] [`#[derive()]`](https://github.com/rust-lang/rust/issues/54651#issuecomment-425565263)
 - [x] ~[`#[must_use]` on functions with unit or uninhabited return type](https://github.com/rust-lang/rust/issues/54828#issuecomment-427246850)~
 - [ ] [`#[allow()]`, `#[warn()]`, _&c._](https://github.com/rust-lang/rust/pull/54683#discussion_r225381179)

ehuss:
This adds some checks for duplicate attributes. In many cases, the duplicates were being ignored without error or warning. This adds several kinds of checks (see `AttributeDuplicates` enum).

The motivation here is to issue unused warnings with similar reasoning for any unused lint, and to error for cases where there are conflicts. 

This also adds a check for empty attribute lists in a few attributes where this causes the attribute to be ignored.

Closes #55112.

### DIFF
--- a/compiler/rustc_feature/src/builtin_attrs.rs
+++ b/compiler/rustc_feature/src/builtin_attrs.rs
@@ -1,5 +1,6 @@
 //! Built-in attributes and `cfg` flag gating.
 
+use AttributeDuplicates::*;
 use AttributeGate::*;
 use AttributeType::*;
 
@@ -88,9 +89,64 @@ impl AttributeGate {
 /// Only top-level shape (`#[attr]` vs `#[attr(...)]` vs `#[attr = ...]`) is considered now.
 #[derive(Clone, Copy, Default)]
 pub struct AttributeTemplate {
+    /// If `true`, the attribute is allowed to be a bare word like `#[test]`.
     pub word: bool,
+    /// If `Some`, the attribute is allowed to take a list of items like `#[allow(..)]`.
     pub list: Option<&'static str>,
+    /// If `Some`, the attribute is allowed to be a name/value pair where the
+    /// value is a string, like `#[must_use = "reason"]`.
     pub name_value_str: Option<&'static str>,
+}
+
+/// How to handle multiple duplicate attributes on the same item.
+#[derive(Clone, Copy, Default)]
+pub enum AttributeDuplicates {
+    /// Duplicates of this attribute are allowed.
+    ///
+    /// This should only be used with attributes where duplicates have semantic
+    /// meaning, or some kind of "additive" behavior. For example, `#[warn(..)]`
+    /// can be specified multiple times, and it combines all the entries. Or use
+    /// this if there is validation done elsewhere.
+    #[default]
+    DuplicatesOk,
+    /// Duplicates after the first attribute will be an unused_attribute warning.
+    ///
+    /// This is usually used for "word" attributes, where they are used as a
+    /// boolean marker, like `#[used]`. It is not necessarily wrong that there
+    /// are duplicates, but the others should probably be removed.
+    WarnFollowing,
+    /// Same as `WarnFollowing`, but only issues warnings for word-style attributes.
+    ///
+    /// This is only for special cases, for example multiple `#[macro_use]` can
+    /// be warned, but multiple `#[macro_use(...)]` should not because the list
+    /// form has different meaning from the word form.
+    WarnFollowingWordOnly,
+    /// Duplicates after the first attribute will be an error.
+    ///
+    /// This should be used where duplicates would be ignored, but carry extra
+    /// meaning that could cause confusion. For example, `#[stable(since="1.0")]
+    /// #[stable(since="2.0")]`, which version should be used for `stable`?
+    ErrorFollowing,
+    /// Duplicates preceding the last instance of the attribute will be an error.
+    ///
+    /// This is the same as `ErrorFollowing`, except the last attribute is the
+    /// one that is "used". This is typically used in cases like codegen
+    /// attributes which usually only honor the last attribute.
+    ErrorPreceding,
+    /// Duplicates after the first attribute will be an unused_attribute warning
+    /// with a note that this will be an error in the future.
+    ///
+    /// This should be used for attributes that should be `ErrorFollowing`, but
+    /// because older versions of rustc silently accepted (and ignored) the
+    /// attributes, this is used to transition.
+    FutureWarnFollowing,
+    /// Duplicates preceding the last instance of the attribute will be a
+    /// warning, with a note that this will be an error in the future.
+    ///
+    /// This is the same as `FutureWarnFollowing`, except the last attribute is
+    /// the one that is "used". Ideally these can eventually migrate to
+    /// `ErrorPreceding`.
+    FutureWarnPreceding,
 }
 
 /// A convenience macro for constructing attribute templates.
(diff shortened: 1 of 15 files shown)


ehuss:
cc @Aaron1011, since you recently worked on this in #87739. I'm not sure I fully understood the `macro_use` check, as it didn't seem to be working as I would expect. `macro_use(..)` duplicates were already handled by `unused_imports`. This changes it so that bare `#[macro_use]` duplicates are warned, and `#[macro_use()]` is warned since it has no effect.

cc @petrochenkov who has written a lot of attribute validation.

ehuss:
@wesleywiser I was wondering if you'd have a chance to take a look at this.

compiler/rustc_feature/src/builtin_attrs.rs
```diff
@@ -300,46 +393,58 @@ pub const BUILTIN_ATTRIBUTES: &[BuiltinAttribute] = &[
...
+    ),
+    gated!(
+        thread_local, Normal, template!(Word), WarnFollowing,
         "`#[thread_local]` is an experimental feature, and does not currently handle destructors",
     ),
-    gated!(no_core, CrateLevel, template!(Word), experimental!(no_core)),
+    gated!(no_core, CrateLevel, template!(Word), WarnFollowing, experimental!(no_core)),
     // RFC 2412
+    // FIXME: optimize are DuplicatesOk, but it ignores all previous values,
+    // so those should probably be warnings.
```
petrochenkov, line 416: So why don't you make it a warning?
(Or rather an error, because it's a key-value attribute, and it's unstable.)

compiler/rustc_feature/src/builtin_attrs.rs
```diff
@@ -214,84 +297,94 @@ pub const BUILTIN_ATTRIBUTES: &[BuiltinAttribute] = &[
...
     gated!(
         no_sanitize, Normal,
-        template!(List: "address, memory, thread"),
+        template!(List: "address, memory, thread"), DuplicatesOk,
         experimental!(no_sanitize)
     ),
-    gated!(no_coverage, Normal, template!(Word), experimental!(no_coverage)),
+    gated!(no_coverage, Normal, template!(Word), WarnFollowing, experimental!(no_coverage)),
 
     // FIXME: #14408 assume docs are used since rustdoc looks at them.
```
petrochenkov, line 368: I'm pretty sure this FIXME is no longer relevant.

compiler/rustc_feature/src/builtin_attrs.rs
```diff
@@ -171,40 +246,48 @@ pub const BUILTIN_ATTRIBUTES: &[BuiltinAttribute] = &[
...
+        template!(Word, List: r#"expected = "reason"#, NameValueStr: "reason"), FutureWarnFollowing,
     ),
     // FIXME(Centril): This can be used on stable but shouldn't.
-    ungated!(reexport_test_harness_main, CrateLevel, template!(NameValueStr: "name")),
+    ungated!(reexport_test_harness_main, CrateLevel, template!(NameValueStr: "name"), ErrorFollowing),
 
     // Macros:
-    ungated!(automatically_derived, Normal, template!(Word)),
+    ungated!(automatically_derived, Normal, template!(Word), WarnFollowing),
     // FIXME(#14407)
```
petrochenkov, line 263: All FIXMEs with this issue number can be removed, and the issue can be closed.
The mechanism of reporting `unused_attributes` changed entirely, so this issue is no longer relevant.

compiler/rustc_feature/src/builtin_attrs.rs
```diff
@@ -88,11 +89,70 @@ impl AttributeGate {
...
+    /// This should be used where duplicates would be ignored, but carry extra
+    /// meaning that could cause confusion. For example, `#[stable(since="1.0")]
+    /// #[stable(since="2.0")]`, which version should be used for `stable`?
+    ErrorFollowing,
+    /// Duplicates preceding the last instance of the attribute will be an error.
+    ///
+    /// This is the same as `ErrorFollowing`, except the last attribute is the
+    /// one that is "used". This is typically used in cases like codegen
+    /// attributes which usually only honor the last attribute.
+    ErrorPreceding,
```
petrochenkov, line 133: We shouldn't make this distinction, IMO, and should merge `ErrorFollowing` into this variant, and always use the last specified value. (And error on the previously specified values.)

compiler/rustc_feature/src/builtin_attrs.rs
ehuss: I don't remember why I wrote that. I switched it to `ErrorPreceding`.

compiler/rustc_feature/src/builtin_attrs.rs
ehuss: I removed all the 14408 comments and closed it.

compiler/rustc_feature/src/builtin_attrs.rs
ehuss: Yea, I was a little uneasy with having different following/preceding behaviors.

I can easily remove `ErrorFollowing`, but I feel like that emphasizes the inconsistencies:

* Normally when there are duplicate definitions (like items), the compiler generates an error about the following entries. Why would attributes have the opposite behavior?
* It seems inconsistent to have `WarnFollowing` and `ErrorPreceding`. Why have one behavior for warning and the opposite for errors?
    * Comparing to `unused_variables`, if you have `let x = 1; let x = 2; ...`, it will warn about the first `x`.

So, perhaps it would be better to remove `ErrorPreceding` instead? At least, I feel like that would be more consistent with the rest of the language. That would also match the behavior of things like the [proc-macro check](https://github.com/rust-lang/rust/blob/f31622a50b06e4e5775cf71d1dfe3e59ffc21aa5/compiler/rustc_builtin_macros/src/proc_macro_harness.rs#L194-L225) which generates errors for following duplicates.

petrochenkov: I mean if
```rust
#[key = "value1"]
#[key = "value2"]
NODE
```
was intentionally allowed, then `value2` would probably override `value1`, similarly to how later `let` or `macro_rules` "override" (shadow) earlier `let` or `macro_rules`, or how later command line options override earlier command line options.
So `value1` would be unused.

But this requires changing other parts of the compiler to harmonize the errors and the actual behavior after the error is reported, so it's probably not a material for this PR.

petrochenkov:
Blocked on https://github.com/rust-lang/rust/pull/90881.

ehuss:
This should be unblocked now via #90382.

@bors r=petrochenkov

matthiaskrgr:
Looks like this failed in a rollup: https://github.com/rust-lang/rust/pull/91041#issuecomment-974290323

@bors r-

ehuss:
Posted #91052, I mistakenly thought another PR had a complete update.

petrochenkov:
The `link-ordinal-multiple.rs` test needs an update.

ehuss:
@bors retry 

#81890

PR #88690, opened 2021-09-06
Accept `m!{ .. }.method()` and `m!{ .. }?` statements. 

m-ou-se:
This PR fixes something that I keep running into when using `quote!{}.into()` in a proc macro to convert the `proc_macro2::TokenStream` to a `proc_macro::TokenStream`:

Before:

```
error: expected expression, found `.`
 --> src/lib.rs:6:6
  |
4 |     quote! {
5 |         ...
6 |     }.into()
  |      ^ expected expression
```

After:
```
```
(No output, compiles fine.)

---

Context:

For expressions like `{ 1 }` and `if true { 1 } else { 2 }`, we accept them as full statements without a trailing `;`, which means the following is not accepted:

```rust
{ 1 } - 1 // error
```

since that is parsed as two statements: `{ 1 }` and `-1`. Syntactically correct, but the type of `{ 1 }` should be `()` as there is no `;`.

However, for specifically `.` and `?` after the `}`, we do [continue parsing it as an expression](https://github.com/rust-lang/rust/blob/13db8440bbbe42870bc828d4ec3e965b38670277/compiler/rustc_parse/src/parser/expr.rs#L864-L876):

```rust
{ "abc" }.len(); // ok
```

For braced macro invocations, we do not do this:

```rust
vec![1, 2, 3].len(); // ok
vec!{1, 2, 3}.len(); // error
```

(It parses `vec!{1, 2, 3}` as a full statement, and then complains about `.len()` not being a valid expression.)

This PR changes this to also look for a `.` and `?` after a braced macro invocation. We can be sure the macro is an expression and not a full statement in those cases, since no statement can start with a `.` or `?`.

### DIFF
--- a/compiler/rustc_parse/src/parser/stmt.rs
+++ b/compiler/rustc_parse/src/parser/stmt.rs
@@ -155,17 +155,20 @@ impl<'a> Parser<'a> {
 
         let mac = MacCall { path, args, prior_type_ascription: self.last_type_ascription };
 
-        let kind = if delim == token::Brace || self.token == token::Semi || self.token == token::Eof
-        {
-            StmtKind::MacCall(P(MacCallStmt { mac, style, attrs, tokens: None }))
-        } else {
-            // Since none of the above applied, this is an expression statement macro.
-            let e = self.mk_expr(lo.to(hi), ExprKind::MacCall(mac), AttrVec::new());
-            let e = self.maybe_recover_from_bad_qpath(e, true)?;
-            let e = self.parse_dot_or_call_expr_with(e, lo, attrs.into())?;
-            let e = self.parse_assoc_expr_with(0, LhsExpr::AlreadyParsed(e))?;
-            StmtKind::Expr(e)
-        };
+        let kind =
+            if (delim == token::Brace && self.token != token::Dot && self.token != token::Question)
+                || self.token == token::Semi
+                || self.token == token::Eof
+            {
+                StmtKind::MacCall(P(MacCallStmt { mac, style, attrs, tokens: None }))
+            } else {
+                // Since none of the above applied, this is an expression statement macro.
+                let e = self.mk_expr(lo.to(hi), ExprKind::MacCall(mac), AttrVec::new());
+                let e = self.maybe_recover_from_bad_qpath(e, true)?;
+                let e = self.parse_dot_or_call_expr_with(e, lo, attrs.into())?;
+                let e = self.parse_assoc_expr_with(0, LhsExpr::AlreadyParsed(e))?;
+                StmtKind::Expr(e)
+            };
         Ok(self.mk_stmt(lo.to(hi), kind))
     }
 
--- a/src/test/ui/parser/macro-braces-dot-question.rs
+++ b/src/test/ui/parser/macro-braces-dot-question.rs
@@ -1,0 +1,9 @@
+// check-pass
+
+use std::io::Write;
+
+fn main() -> Result<(), std::io::Error> {
+    vec! { 1, 2, 3 }.len();
+    write! { vec![], "" }?;
+    Ok(())
+}
--- a/src/test/ui/parser/macro-braces-dot-question.rs
+++ b/src/test/ui/parser/macro-braces-dot-question.rs
@@ -5,5 +5,7 @@ use std::io::Write;
 fn main() -> Result<(), std::io::Error> {
     vec! { 1, 2, 3 }.len();
     write! { vec![], "" }?;
+    println!{""}
+    [0]; // separate statement, not indexing into the result of println.
     Ok(())
 }


joshtriplett:
I think, before approving something like this, we'd want to know if this restriction is important for any particular reason (e.g. interactions with parsing and follow sets). I don't *think* it would be, given the examples you gave, but it'd be helpful to have some guidance from someone with clear knowledge of how macros interact with the parser.

For my part, assuming it wouldn't break anything, I think this seems reasonable.

petrochenkov:
I think this is the right thing to do and don't see any reasons for braced macros `m! { ... }` to be different from blocks `{ ... }` in this position.
This shouldn't break anything since we are only making previously illegal code legal, and I don't see issues with expansion either.

cc @Aaron1011 just in case

nikomatsakis:
+1 from me, I agree that this seems more consistent.

compiler/rustc_parse/src/parser/stmt.rs
```diff
@@ -155,17 +155,20 @@ impl<'a> Parser<'a> {
...
-        } else {
-            // Since none of the above applied, this is an expression statement macro.
-            let e = self.mk_expr(lo.to(hi), ExprKind::MacCall(mac), AttrVec::new());
-            let e = self.maybe_recover_from_bad_qpath(e, true)?;
-            let e = self.parse_dot_or_call_expr_with(e, lo, attrs.into())?;
-            let e = self.parse_assoc_expr_with(0, LhsExpr::AlreadyParsed(e))?;
-            StmtKind::Expr(e)
-        };
+        let kind =
+            if (delim == token::Brace && self.token != token::Dot && self.token != token::Question)
```
nagisa, line 159: Could this condition be split out into a variable?

src/test/ui/parser/macro-braces-dot-question.rs
```diff
@@ -0,0 +1,9 @@
+// check-pass
+
+use std::io::Write;
+
+fn main() -> Result<(), std::io::Error> {
```
nagisa, line 5: Some negative tests such as `vec![1, 2, 3].len()` still failing to build would be great.

compiler/rustc_parse/src/parser/stmt.rs
nagisa: Side observation: the reference's specification of the grammar is fairly inaccurate concerning this specific behaviour. I was going to check if the reference needs adjusting for this change, but it seems like the grammar already allows treating `m!{ ... }` exactly like `m![...]` and `m!(...)` in any `ExpressionWithoutBlock` position.

If we go by reference the `delim == token::Brace` check should not be present here at all.

In practice I suspect that we want to make `m! { ... }` be considered a `ExpressionWithBlock` and allow use of `m! { ... }` anywhere where other `ExpressionWithBlock`s are usable. One example that comes to mind immediately is `vec! { ... }[0]` or some such.

nagisa (approved):
I'm okay with this change on the implementation side of things.

On the language design side of things I feel like with this adjustment we end up neither here nor there for this syntax element appears to be treated in a fairly inconsistent manner compared to the rest of Rust's syntactic constructs. That is it now acts kind of like an expression but only in very specific contexts. I think it would be a good idea to go all the way to making it act like all the other `ExpressionWithBlock`s.

compiler/rustc_parse/src/parser/stmt.rs
m-ou-se: That would be a breaking change.

This is already accepted as two statements:

```rust
bla! {}
[0];
```

nagisa: Aha, right, my bad. Upon further checking [this also holds true for other `ExpressionWithBlock`s](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=18966072e2da8e8a0cd239483fb4e2a1) (in that `ExpressionWithBlock [idx]` parses as two statements)  though, apparently, so the point still stands I feel.

m-ou-se: There are only two cases where an ExpressionWithBlock continues after the `}`: When followed by either a `.` or a `?`. So this PR makes that consistent with `m!{}`.

nikomatsakis: My understanding of the description was exactly as @m-ou-se described, that this PR made `Foo { }` and `m!{ }` consistent. I agree that they should be consistent.

scottmcm:
+1 from me as well for making it consistent.  Given that things like
```rust
let x = loop { break 7_u32 }.count_ones();
```
work, then those same things working after macro_rules with braces seems like the right choice.

nagisa:
r=me with a negative test added.

PR #88678, opened 2021-09-05
Change scope of temporaries in match guards

matthewjasper:
Each pattern in a match arm has its own copy of the match guard in MIR, with its own temporary, so it has to be dropped before the the guards are joined to the single copy of the arm. This PR changes `then_else_break` to allow it to put the temporary in the innermost scope possible. This change isn't done for `if` expressions because that affects a large number of mir-opt tests and could more significantly affect performance.

closes #88649

r? @oli-obk

Issue #88649, opened 2021-09-04
StorageLive is added on a local that was already live


camelid:
Miri's CI started failing recently, with an error similar to:

```
error: Undefined Behavior: StorageLive on a local that was already live
 --> foo.rs:9:52
  |
9 |             Foo::Variant1(x) | Foo::Variant2(x) if x => {}
  |                                                    ^ StorageLive on a local that was already live
  |
  = help: this indicates a bug in the program: it performed an invalid operation, and caused Undefined Behavior
  = help: see https://doc.rust-lang.org/nightly/reference/behavior-considered-undefined.html for further information
```

It seems likely that this is a regression from #88572. cc @matthewjasper 

Thanks to @hyd-dev for doing the original minimization.

### MCVE

```rust
enum Foo {
    Variant1(bool),
    Variant2(bool),
}

fn main() {
    loop {
        match Foo::Variant1(true) {
            Foo::Variant1(x) | Foo::Variant2(x) if x => {}
            _ => {}
        }
    }
}
```

### Generated MIR (from right after building)

```rust
// MIR for `main` 0 mir_map

fn main() -> () {
    let mut _0: ();                      // return place in scope 0 at foo.rs:6:11: 6:11
    let mut _1: !;                       // in scope 0 at foo.rs:7:5: 12:6
    let mut _2: ();                      // in scope 0 at foo.rs:6:1: 13:2
    let mut _3: Foo;                     // in scope 0 at foo.rs:8:15: 8:34
    let mut _4: isize;                   // in scope 0 at foo.rs:9:13: 9:29
    let mut _5: &Foo;                    // in scope 0 at foo.rs:8:15: 8:34
    let _6: bool;                        // in scope 0 at foo.rs:9:27: 9:28
    let _7: &bool;                       // in scope 0 at foo.rs:9:27: 9:28
    let mut _8: bool;                    // in scope 0 at foo.rs:9:52: 9:53
    let mut _9: bool;                    // in scope 0 at foo.rs:9:52: 9:53
    scope 1 {
        debug x => _6;                   // in scope 1 at foo.rs:9:27: 9:28
        debug x => _7;                   // in scope 1 at foo.rs:9:27: 9:28
    }

    bb0: {
        StorageLive(_1);                 // scope 0 at foo.rs:7:5: 12:6
        goto -> bb1;                     // scope 0 at foo.rs:7:5: 12:6
    }

    bb1: {
[...]

PR #88691, opened 2021-09-06
Add a regression test for #88649

ghost:
I noticed that #88649 does not have a regression test, so I add one in this PR.

The test fails with this without #88678:
```
error[E0080]: evaluation of constant value failed
  --> /checkout/src/test/ui/consts/issue-88649.rs:13:52
   |
LL |             Foo::Variant1(x) | Foo::Variant2(x) if x => {}
   |                                                    ^ StorageLive on a local that was already live

For more information about this error, try `rustc --explain E0080`.
```

### DIFF
--- a/src/test/ui/consts/issue-88649.rs
+++ b/src/test/ui/consts/issue-88649.rs
@@ -1,0 +1,18 @@
+// check-pass
+#![crate_type = "lib"]
+
+enum Foo {
+    Variant1(bool),
+    Variant2(bool),
+}
+
+const _: () = {
+    let mut n = 0;
+    while n < 2 {
+        match Foo::Variant1(true) {
+            Foo::Variant1(x) | Foo::Variant2(x) if x => {}
+            _ => {}
+        }
+        n += 1;
+    }
+};


src/test/ui/consts/issue-88649.rs
```diff
@@ -0,0 +1,18 @@
...
+const _: () = {
+    let mut n = 0;
+    while n < 2 {
+        match Foo::Variant1(true) {
+            Foo::Variant1(x) | Foo::Variant2(x) if x => {}
+            _ => {}
+        }
+        n += 1;
+    }
+};
```
camelid, line 18: Why did you put it in a `const`? There's a simpler version of this that failed before #88678 and succeeds after:

```rust
fn main() {
    loop {
        match Foo::Variant1(true) {
            Foo::Variant1(x) | Foo::Variant2(x) if x => {}
            _ => {}
        }
    }
}
```

(From #88649.)

hyd-dev: I think that requires Miri to run? It does not seem to fail in `rustc`, so I don't think it works in a UI test.

camelid: Ah, I see; you put it in a `const` so that the rustc version of Miri for const-eval would run 👍 

I think you could still simplify the `while` into an unqualified `loop`, but your call.

hyd-dev: > I think you could still simplify the `while` into an unqualified `loop`, but your call.

Using an unqualified `loop` means that it will enter an infinite loop and fails with this:
```
9  | / const _: () = {
10 | |     loop {
11 | |         match Foo::Variant1(true) {
12 | |             Foo::Variant1(x) | Foo::Variant2(x) if x => {}
   | |                                                          ^ exceeded interpreter step limit (see `#[const_eval_limit]`)
...  |
15 | |     }
16 | | };
   | |__-
```
I think I can test ``//~ ERROR exceeded interpreter step limit (see `#[const_eval_limit]`)`` instead, but I think it may be clearer to make this a `check-pass` test without an infinite loop.
(However, if `while n < 2` is actually more confusing, I'm fine with changing this to `loop`.)

camelid: That makes a lot of sense. Sorry for the noise :)

Mark-Simulacrum:
@bors r+ rollup

Also beta-nom alongside #88678 to make sure we have a test if we're backporting the original patch.

pnkfelix:
beta-backport approved, *coupled* to PR #88678. (I.e. their backport should be treated as one unit.)

apiraino:
Discussion of the compiler team [on Zulip](https://rust-lang.zulipchat.com/#narrow/stream/238009-t-compiler.2Fmeetings/topic/.5Bweekly.5D.202021-09-16.20.2354818/near/253589067)

@rustbot label +beta-accepted

PR #88692, opened 2021-09-06
Update Miri

### LINKED ISSUE #88671: `miri` no longer builds after rust-lang/rust#88499
Hello, this is your friendly neighborhood mergebot.
After merging PR rust-lang/rust#88499, I observed that the tool miri no longer builds.
A follow-up PR to the repository https://github.com/rust-lang/miri is needed to fix the fallout.

cc @eddyb, do you think you would have time to do the follow-up work?
If so, that would be great!

ghost:
Fixes #88671.

r? @RalfJung

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -546,7 +546,7 @@ version = "0.1.56"
 dependencies = [
  "cargo_metadata 0.12.0",
  "clippy_lints",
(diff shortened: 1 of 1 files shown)


RalfJung:
@bors r+ p=1

Thanks. :-)

PR #88702, opened 2021-09-06
Give sidebar a z-index

nbdd0121:
This ensures that sidebar does not overlap with content when horizontal scrollbar is present.

The overlap is visible in https://docs.rs/ndarray/0.15.3/ndarray/struct.ArrayBase.html.

Discovered by @m13253.

### DIFF
--- a/src/librustdoc/html/static/css/rustdoc.css
+++ b/src/librustdoc/html/static/css/rustdoc.css
@@ -290,6 +290,12 @@ nav.sub {
 	overflow: auto;
 }
 
+.rustdoc:not(.source) .sidebar {
+	/* Ensure that sidebar does not overlap with content when horizontal
+	 * scrollbar is present */
+	z-index: 1;
+}
+
 /* Improve the scrollbar display on firefox */
 * {
 	scrollbar-width: initial;


nbdd0121:
The error doesn't seem relevant...

GuillaumeGomez:
I don't think this is the right fix. The content shouldn't be larger that the container. The scrollbar should be on the `<table>` directly.

nbdd0121:
> The scrollbar should be on the `<table>` directly.

Many elements could cause the overflow, not just a table. E.g. a very long inline code block could cause the content to overflow too, but you cannot put a scrollbar on it because it isn't and shouldn't be `display: block`.

For this particular case a scrollbar on the table might be desirable, but in general style still shouldn't break when content overflows.

GuillaumeGomez:
Agreed. I'm planning to wrap the `<table>` items into `<div>` and to add this CSS rule:

```css
.docblock * {
    max-width: 100%;
    overflow-x: scroll;
}
```

That should prevent most issues.

nbdd0121:
Should be `overflow-x: auto;`

GuillaumeGomez:
I opened #88742 so I think this PR can be closed. To give a few more information about why I don't think this is a good idea: hiding the issue (but making the scrolled content go under the sidebar) doesn't fix the original problem. Instead of making the whole doc block scroll, I prefer to limit it to the element having a too big width directly.

Thanks in any case for bringing this issue to our attention! :)

nbdd0121:
No I think both PRs should be merged. They tackle different problems:
* #88742 reduces the chance of overflow happening;
* This PR addresses the behaviour when overflow happens: even if overflow happens, the result should still be sane.

As I said earlier, it not feasible in general to prevent overflow from happening at all. Have
```markdown
`<insert a very long symbol name here>` Note that's an inline code block so it doesn't make sense to put a scroll bar on it.
```
in the doc still shouldn't break the style and cause the overlap to happen.

nbdd0121:
Oh I miss the `.docblock > * { max-width: 100%; }` part. I suppose that'll work too.

PR #88703, opened 2021-09-06
Gather module items after lowering.

cjgillot:
This avoids having a non-local analysis inside lowering.

By implementing `hir_module_items` using a visitor, we make sure that iterations and visitors are consistent.


compiler/rustc_middle/src/hir/mod.rs
```diff
@@ -121,6 +121,17 @@ impl<'tcx> AttributeMap<'tcx> {
...
+#[derive(Default, Encodable, Debug, HashStable)]
+pub struct ModuleItems {
+    // Use BTreeSets here so items are in the same order as in the
+    // list of all items in Crate
+    submodules: BTreeSet<LocalDefId>,
+    items: BTreeSet<ItemId>,
+    trait_items: BTreeSet<TraitItemId>,
+    impl_items: BTreeSet<ImplItemId>,
+    foreign_items: BTreeSet<ForeignItemId>,
+}
```
petrochenkov, line 133: This structure is certainly missing a comment describing what these sets mean.
It's far from being obvious that the sets include *all* items in the module, including items inside function bodies etc, not only direct children, and such sets only make sense as some partitioning for parallel processing.

compiler/rustc_middle/src/query/mod.rs
```diff
@@ -44,8 +44,8 @@ rustc_queries! {
     ///
     /// This can be conveniently accessed by `tcx.hir().visit_item_likes_in_module`.
     /// Avoid calling this query directly.
-    query hir_module_items(key: LocalDefId) -> &'tcx hir::ModuleItems {
-        eval_always
+    query hir_module_items(key: LocalDefId) -> rustc_middle::hir::ModuleItems {
+        storage(ArenaCacheSelector<'tcx>)
```
petrochenkov, line 48: What does this line mean? Is changing the cache storage mandatory here, or is it an optimization?

petrochenkov:
>r? @petrochenkov

Nooo, I only left a comment...

petrochenkov:
Looks like this PR does two different things:
- Moves the global analysis pass from lowering to somewhere else.
- Changes how iteration through modules is done - walking a single static list (no allocations or other overhead, maximum parallelism) vs. walking many lists of submodules (allocations in the queue, parallelism is apparently reduces as well).

The second part may be responsible for the perf differences.

compiler/rustc_middle/src/query/mod.rs
cjgillot: If we don't change the cache storage, the query system will clone the return value for each invocation.

jackh726:
> > r? @petrochenkov
> 
> Nooo, I only left a comment...

oops, tag? (but for real, I can take back the review if you've got too much on your plate, just figured this was more along what you'd normally review ♥)

pnkfelix:
Visited for weekly performance triage. Several benchmarks improved, by around -1% (+/-0.20%).

Some others regressed, by around +0.5% (+/- 0.2%)

To me, this seems like a wash; I'm not sure its worth spending time investigating the source of the regressions.

But since this was only labelled as a perf-regression two days ago, I'll hold off on officially triaging it; maybe someone will notice something I missed.

PR #88705, opened 2021-09-06
[perf experiment] Add unused `fn_abi_of_{fn_ptr,instance}` queries.

eddyb:
*Opening as draft as it's not intended to be merged, it's only for perf testing*

This is based on the observation that the two queries #88575 attempts to introduce, cause perf regressions in `-check` benchmarks, which shouldn't be invoking the queries.

Since my suspicion is on the query infrastructure itself becoming less efficient with more queries, and to minimize the diff, I've replaced the return types with `()`.


eddyb:
Looks like this was a dead end, closing.

PR #88706, opened 2021-09-06
Normalize associated type projections when checking return type of main

### LINKED ISSUE #88609: Checking that `main`'s output type implements `Termination` doesn't normalize associated types
I tried this code:

```rust
#![feature(termination_trait_lib)]

trait Same {
    type Output;
}
impl<T> Same for T {
    type Output = T;
}
type Unit = <() as Same>::Output;

fn test() {
    fn foo() -> Result<Unit, std::io::Error> { todo!() }

    fn assert_impl_termination<T: std::process::Termination>(_: fn() -> T) {}
    // Compiles
    assert_impl_termination(foo);
}

// Doesn't compile
fn main() -> Result<Unit, std::io::Error> {
    todo!()
}
```

[Playground link](https://play.rust-lang.org/?version=nightly&mode=debug&edition=2018&gist=51c0b2d9e9500deb7b0d9a1a709dd443)

I expected to see both `test` and `main` compile, instead only `test` compiles, while trying to compile `main` gives this error:

```
error[E0277]: `main` has invalid return type `Result<<() as Same>::Output, std::io::Error>`
  --> src/main.rs:20:14
   |
20 | fn main() -> Result<Unit, std::io::Error> {
   |              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^ `main` can only return types that implement `Termination`
   |
   = help: consider using `()`, or a `Result`
help: consider introducing a `where` bound, but there might be an alternative better way to express this requirement
   |
20 | fn main() -> Result<Unit, std::io::Error> where Result<<() as Same>::Output, std::io::Error>: Termination {
   |                                           +++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++

For more information about this error, try `rustc --explain E0277`.
```

Note that the `feature` attribute is only needed for the bound in `assert_impl_termination`. The issue with `main` happens both in stable and nightly, with or without the feature.

### Meta

`rustc --version --verbose`:
```
binary: rustc
commit-hash: a178d0322ce20e33eac124758e837cbd80a6f633
commit-date: 2021-07-26
host: x86_64-pc-windows-msvc
release: 1.54.0
LLVM version: 12.0.1
```
```
rustc 1.56.0-nightly (29ef6cf16 2021-08-31)
binary: rustc
commit-hash: 29ef6cf1637aa8317f8911f93f14e18d404c1b0e
commit-date: 2021-08-31
host: x86_64-pc-windows-msvc
release: 1.56.0-nightly
LLVM version: 13.0.0
```

ThePuzzlemaker:
This fixes #88609.

Previously, the return type of `fn main()` would not have any associated type projections within normalized before checking if it implements the standard library trait `std::process::Termination`. This commit appears to fix it.

This feels vaguely symptomatic of a problem in the underlying trait solving engine, but I am not sure how I would solve that. I am unsure why the example in #88609 with `assert_impl_termination` and `fn foo()` work, but simply `fn main()` doesn't. The way that I solved this is also probably not the best way to do this, so please let me know if there is a better way to do this.

I have added a build-pass regression test for this issue.

### DIFF
--- a/compiler/rustc_typeck/src/lib.rs
+++ b/compiler/rustc_typeck/src/lib.rs
@@ -335,16 +335,20 @@ fn check_main_fn_ty(tcx: TyCtxt<'_>, main_def_id: DefId) {
             // checking trait fulfillment, not this here. I'm not sure why it
             // works in the example in `fn test()` given in #88609? This also
             // probably isn't the best way to do this.
-            let normalized = infcx.partially_normalize_associated_types_in(
-                cause.clone(),
+            let InferOk { value: norm_return_ty, obligations } = infcx
+                .partially_normalize_associated_types_in(
+                    cause.clone(),
+                    ty::ParamEnv::empty(),
+                    return_ty,
+                );
+            fulfillment_cx.register_predicate_obligations(&infcx, obligations);
+            fulfillment_cx.register_bound(
+                &infcx,
                 ty::ParamEnv::empty(),
-                return_ty,
+                norm_return_ty,
+                term_id,
+                cause,
             );
-            let new_ty = normalized.value;
-            for obligation in normalized.obligations {
-                fulfillment_cx.register_predicate_obligation(&infcx, obligation);
-            }
-            fulfillment_cx.register_bound(&infcx, ty::ParamEnv::empty(), new_ty, term_id, cause);
             if let Err(err) = fulfillment_cx.select_all_or_error(&infcx) {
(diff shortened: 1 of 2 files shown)


ThePuzzlemaker:
This is my first time actually touching anything within typechecking, so please let me know if I accidentally silently broke something or did something wrong. :sweat_smile:

jackh726 (approved):
I wanted to think more about this, if there's a better way to do this. But I think this is fine for now.

PR #88707, opened 2021-09-06
String.split_terminator: Add an example when using a slice of chars

sylvestre:


yaahc:
Thank you very much ^^

@bors r+ rollup

sylvestre:
thanks!

PR #88710, opened 2021-09-07
Use index newtyping for TyVid

Mark-Simulacrum:
This is useful for using TyVid in types like VecGraph, and just otherwise seems like a small win.

### DIFF
--- a/compiler/rustc_infer/src/infer/type_variable.rs
+++ b/compiler/rustc_infer/src/infer/type_variable.rs
@@ -196,7 +196,7 @@ impl<'tcx> TypeVariableTable<'_, 'tcx> {
     /// Note that this function does not return care whether
     /// `vid` has been unified with something else or not.
     pub fn var_diverges(&self, vid: ty::TyVid) -> Diverging {
-        self.storage.values.get(vid.index as usize).diverging
+        self.storage.values.get(vid.index()).diverging
     }
 
     /// Returns the origin that was given when `vid` was created.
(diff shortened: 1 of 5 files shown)


compiler/rustc_infer/src/infer/type_variable.rs
```diff
@@ -357,11 +357,11 @@ impl<'tcx> TypeVariableTable<'_, 'tcx> {
         &mut self,
         value_count: usize,
     ) -> (Range<TyVid>, Vec<TypeVariableOrigin>) {
-        let range = TyVid { index: value_count as u32 }..TyVid { index: self.num_vars() as u32 };
+        let range = TyVid::from_u32(value_count as u32)..TyVid::from_usize(self.num_vars());
```
jackh726, line 360: This could be `TyVid::from_usize(value_count)` right?

jackh726 (approved):
One small nit, but r=me otherwise

compiler/rustc_infer/src/infer/type_variable.rs
Mark-Simulacrum: Yeah, good catch.

jackh726:
@bors rollup=never

It's *possible* this is perf sensitive, given `TyVid`s are used practically everywhere.

Mark-Simulacrum:
Majority of the builders finished, with distcheck being the odd one out - I'm guessing this is spurious.

@bors retry

Mark-Simulacrum:
Hm, interesting. My guess is the extra checks for from_u32/from_usize being in-range (now that we have some niche bits) are causing the worse performance -- that said, cycle times seem more improved than not (https://perf.rust-lang.org/compare.html?start=73641cd23ba470c6b4dcd72b8d5f62d27c735254&end=69c4aa2901ffadf69deaf91b2f90604bcbc2eb36&stat=cycles%3Au) and generally keccak is more of a stress test than real world code.

I'll try to spend some time with cachegrind investigating this.

Mark-Simulacrum:
Cachegrind seems to confirm that the index manipulation is the likely cause of the added overhead. I don't think that's worth fixing with unsafe code (and would be difficult to justify).

PR #88713, opened 2021-09-07
Improve docs for int_log

falk-hueffner:
* Clarify rounding.
* Avoid "wrapping" wording.
* Omit wrong claim on 0 only being returned in error cases.
* Typo fix for one_less_than_next_power_of_two.


scottmcm:
Thanks for the PR, @falk-hueffner!  These updates all look good to me.

@bors r+ rollup

PR #88714, opened 2021-09-07
use CLOCK_BOOTTIME in Instant::now

### LINKED ISSUE #87906: `Instant::duration_since` returns a shortened duration if a suspend occurred

hellow554:
There was a discussion on rust-lang/rust#87907 that all platforms expect
linux advance its monotonic timer if the system goes into a sleep state
(e.g. suspend or hibernate).
It was decided that CLOCK_BOOTTIME should be used on inux targets, but
because that constant would break older systems (e.g. linux kernels
<2.6.39) there should be a fallback to CLOCK_MONOTONIC.

Related to #87907
Closes #87906

cc @joshtriplett @cuviper

### DIFF
--- a/library/std/src/sys/unix/time.rs
+++ b/library/std/src/sys/unix/time.rs
@@ -291,8 +291,13 @@ mod inner {
         pub fn now() -> Instant {
             #[cfg(target_os = "macos")]
             const clock_id: libc::clockid_t = libc::CLOCK_UPTIME_RAW;
-            #[cfg(not(target_os = "macos"))]
+
+            #[cfg(target_os = "linux")]
+            const clock_id: libc::clockid_t = libc::CLOCK_BOOTTIME;
+
+            #[cfg(not(any(target_os = "macos", target_os = "linux")))]
             const clock_id: libc::clockid_t = libc::CLOCK_MONOTONIC;
+
             Instant { t: Timespec::now(clock_id) }
         }
 
--- a/src/tools/miri/src/shims/time.rs
+++ b/src/tools/miri/src/shims/time.rs
@@ -45,6 +45,9 @@ pub trait EvalContextExt<'mir, 'tcx: 'mir>: crate::MiriInterpCxExt<'mir, 'tcx> {
                 relative_clocks = vec![
                     this.eval_libc_i32("CLOCK_MONOTONIC")?,
                     this.eval_libc_i32("CLOCK_MONOTONIC_COARSE")?,
+                    // This is the equivalent to `CLOCK_UPTIME_RAW` in the macos section
+                    // We support it, because std relies on it
+                    this.eval_libc_i32("CLOCK_BOOTTIME")?,
                 ];
             }
             "macos" => {


library/std/src/sys/unix/time.rs
```diff
@@ -293,7 +294,13 @@ mod inner {
 
     impl Instant {
         pub fn now() -> Instant {
-            Instant { t: now(libc::CLOCK_MONOTONIC) }
+            let t = match now(libc::CLOCK_BOOTTIME) {
+                Err(e) if e.kind() == io::ErrorKind::InvalidInput => now(libc::CLOCK_MONOTONIC),
+                v => v,
+            }
+            .unwrap();
```
hellow554, line 301: IMHO this match is easier to read than a `or_else`, but I'm not sure if that's faster/better/stronger.

Here's a godbolt link, maybe someone can decide what to use: https://rust.godbolt.org/z/6Eoxc7PEz

library/std/src/sys/unix/time.rs
```diff
@@ -366,9 +373,8 @@ mod inner {
     #[cfg(any(target_os = "dragonfly", target_os = "espidf"))]
     pub type clock_t = libc::c_ulong;
 
-    fn now(clock: clock_t) -> Timespec {
+    fn now(clock: clock_t) -> io::Result<Timespec> {
         let mut t = Timespec { t: libc::timespec { tv_sec: 0, tv_nsec: 0 } };
-        cvt(unsafe { libc::clock_gettime(clock, &mut t.t) }).unwrap();
-        t
+        cvt(unsafe { libc::clock_gettime(clock, &mut t.t) }).map(|_| t)
```
hellow554, line 378: Should I write:

```rust
cvt(...)?;
t
```

instead?

CryZe:
Alright, so the situation is the following:
 - All the resources I found seem to indicate that except for Linux `CLOCK_MONOTONIC` does actually include suspend on basically all the operating systems.
 - I have tried to verify this since then, but had a lot of trouble setting up FreeBSD for example (it refused to ever wake up again after suspending)
 - Therefore I am not super confident that the claim is actually true.
 - However if anything macOS would also need to switch over to `CLOCK_MONOTONIC` as it currently uses it's own implementation in std. This function that is called is: deprecated, less precise (u32 + u32 as opposed to u64 + u32) and doesn't include suspends.
 
 So yeah, that's why I didn't open this PR yet.

hellow554:
macos currently uses [`mach_absolute_time`](https://developer.apple.com/documentation/kernel/1462446-mach_absolute_time) ([here](https://github.com/rust-lang/rust/blob/385f8e2078b00282a7a7ffcd58aad17b405f57bf/library/std/src/sys/unix/time.rs#L150-L155)) which states:

> [...] this clock does not increment while the system is asleep.

and

> Prefer to use the equivalent clock_gettime_nsec_np(CLOCK_UPTIME_RAW) in nanoseconds.

Following that: http://manpagez.com/man/3/clock_gettime_nsec_np/

> CLOCK_MONOTONIC    clock that increments monotonically, tracking the time since an arbitrary point, and will continue to increment while the system is asleep.

I'm more than willing to also add a change for that. Only problem I don't have a mac to test that.

thomcc:
Essentially all the clock APIs on Darwin other than `mach_absolute_time` were added in Darwin 16 (macOS 10.12, iOS 10.0, ...). This includes `clock_gettime`, `clock_gettime_nsec_np`, `mach_continuous_time`, `CLOCK_UPTIME_RAW` and all the other `CLOCK_*` constants...

We support back to macOS 10.6 or 10.7, which is considerably older. Darwin 16 is only about 4 or 5 years old, and I think it would be a problem for libstd to require this.

library/std/src/sys/unix/time.rs
workingjubilee: There is not what **I** would consider a significant enough difference in the assembly to merit one approach over another, which is as it should be for such similar expressions.

workingjubilee:
It would be a good idea if further discussion took place in the original issue, since it seems to be that the discussion is ongoing.

joshtriplett:
@thomcc I think it's reasonable to handle macOS in a separate PR. And as with this PR, it'd be acceptable if the primary approach required new macOS and the fallback didn't account for suspend time.

library/std/src/sys/unix/time.rs
```diff
@@ -293,7 +294,13 @@ mod inner {
 
     impl Instant {
         pub fn now() -> Instant {
-            Instant { t: now(libc::CLOCK_MONOTONIC) }
+            let t = match now(libc::CLOCK_BOOTTIME) {
```
joshtriplett, line 297: I think this should be Linux-specific, and other platforms should just use `CLOCK_MONOTONIC` the first time.

joshtriplett: Actually, this should be Linux-specific *and* only on x86 or x86-64, nowhere else. Every other target requires a sufficiently recent kernel to guarantee support for `CLOCK_BOOTTIME`.

workingjubilee: We can only use a constant that our libc actually defines anyways, or it will break compilation for that platform.

maurer: Please please don't limit it to `x86 and `x86_64`, that defeats the much of the purpose of using `CLOCK_BOOTTIME` - since the difference is for cases where the device goes into suspend, which are mostly `arm` and `arm64` devices. Limiting it to Linux seems expected.

joshtriplett: @maurer Clarification: the *fallback* should be limited to x86 and x86_64. We should use `CLOCK_BOOTTIME` on every Linux target. Only `x86` and `x86_64` still support sufficiently old kernels that `CLOCK_BOOTTIME` may not exist, so we only need the fallback to `CLOCK_MONOTONIC` on those targets.

maurer: OK, I am in agreement there, thanks for the clarification.

cuviper: @joshtriplett powerpc, powerpc64, and s390x also document 2.6.32 support.

joshtriplett: @cuviper You're entirely right, and I don't know how I missed those.

jonhoo:
Should this also be tagged with `relnotes`?

library/std/src/sys/unix/time.rs
joshtriplett: Yes, please.

joshtriplett:
This needs an update to make the use of `CLOCK_BOOTTIME` conditional (only on Linux), and to make the fallback conditional (only on x86, x86_64, powerpc, powerpc64, and s390x).

This also needs an update (as noticed by the author) to use `?` in error handling.

JohnCSimon:
Ping from triage:
@hellow554 Can you please post your status on this PR?

library/std/src/sys/unix/time.rs
```diff
@@ -293,7 +294,23 @@ mod inner {
...
-            Instant { t: now(libc::CLOCK_MONOTONIC) }
+            let t = if cfg!(target_os = "linux") {
+                match now(libc::CLOCK_BOOTTIME) {
+                    #[cfg(any(
+                        target_arch = "x86",
+                        target_arch = "x86_64",
+                        target_arch = "powerpc",
+                        target_arch = "powerpc64",
+                        target_arch = "s390x",
+                    ))]
```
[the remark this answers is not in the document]
hellow554: So, should I add a comment or remove the condition at all?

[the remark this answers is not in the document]
cuviper: Let's go ahead with a comment. Even without the conditional, I think we would want something explaining why we need a fallback at all, so we also know that this is something we could remove in the future when our support levels change.

jstarks:
I think this might be a mistake. Go has gone the opposite direction here. In fact, they plan to expose separate times for monotonic vs. boot time:
https://github.com/golang/go/issues/24595
https://github.com/golang/go/issues/36141

A primary reason is that it's generally desirable that timers not all fire at once when a machine resumes from a lengthy sleep, since this leads to spurious timeouts (communication with a local service timed out) as well as performance issues (every process wakes up to do something immediately). Therefore, timeouts in modern versions of Windows (`WaitForSingleObject`, `Sleep`) and Linux (`poll`, `select`, `nanosleep`) follow the behavior of `CLOCK_MONOTONIC` and ignore time the machine is asleep. Linux tried to change this behavior a few years ago to use `CLOCK_BOOTTIME` and they had to revert it due to regressions in user applications: https://lkml.org/lkml/2018/4/26/929.

Of course, there are cases where you _do_ want timers to expire based on real elapsed time, e.g. for network timeouts. Go decided that this was not the appropriate default, so they are adding explicit APIs to support this scenario.

Now, it's true that `Instant` can still use `CLOCK_BOOTTIME` while `thread::sleep` continues to use `CLOCK_MONOTONIC` under the covers. But I think it's desirable to have consistency, so that if you measure an `Instant` before and after a `sleep`, you get a duration that roughly corresponds to the requested sleep time (and not, say, days larger).

In #87907, there was discussion of switching Windows to use something else to resolve the inconsistency instead. That API would be `QueryUnbiasedInterruptTimePrecise` on Windows 10+. Unfortunately, the precise variant did not exist in Windows 7 or Windows 8--`QueryUnbiasedInterruptTime` has only 15ms precision (by default).

maurer:
> I think this might be a mistake. Go has gone the opposite direction here. In fact, they plan to expose separate times for monotonic vs. boot time: [golang/go#24595](https://github.com/golang/go/issues/24595) [golang/go#36141](https://github.com/golang/go/issues/36141)

This was already [mentioned](https://github.com/rust-lang/rust/issues/87906). Exposing both in the long term seems reasonable to me. The discussion/PR is about picking a default for existing code which is:
1. Most likely to match what users expect.
2. As coherent as possible for the OS versions we support.

Go is also mostly used in server environments, and occasionally in desktop/laptop environments, where the fallout of a timer expiring late after suspend is more minimal. In embedded or handheld environments, sleeping is a regular occurrence, not something special, and *most* of your timers will be incorrect if `CLOCK_MONOTONIC` is used. Coming out of suspend with a timer still running which should have expired is also more important, because usually this means the user is beginning to use the device immediately, so waiting 30 seconds while a connection times out is not acceptable.

> 
> A primary reason is that it's generally desirable that timers not all fire at once when a machine resumes from a lengthy sleep, since this leads to spurious timeouts (communication with a local service timed out) as well as performance issues (every process wakes up to do something immediately). Therefore, timeouts in modern versions of Windows (`WaitForSingleObject`, `Sleep`) and Linux (`poll`, `select`, `nanosleep`) follow the behavior of `CLOCK_MONOTONIC` and ignore time the machine is asleep. Linux tried to change this behavior a few years ago to use `CLOCK_BOOTTIME` and they had to revert it due to regressions in user applications: https://lkml.org/lkml/2018/4/26/929.

That was also  mentioned in the previous issue, and I actually consider it evidence as to why we should change it now, before we get user applications that make that assumption. Those assumptions appear in a very specific class of user application (ones which directly manage hardware or watchdogs from userspace), and we have few or none of these right now, and also wouldn't have the same version mismatch problem Linux would have (library updates cannot be shipped separately from program updates with current Rust). If we wait, we may have these types of programs in the future.

> 
> Of course, there are cases where you _do_ want timers to expire based on real elapsed time, e.g. for network timeouts. Go decided that this was not the appropriate default, so they are adding explicit APIs to support this scenario.

Again, I would argue that it is an appropriate default, and Go made this decision due to being almost exclusively used in environments which rarely sleep.

> 
> Now, it's true that `Instant` can still use `CLOCK_BOOTTIME` while `thread::sleep` continues to use `CLOCK_MONOTONIC` under the covers. But I think it's desirable to have consistency, so that if you measure an `Instant` before and after a `sleep`, you get a duration that roughly corresponds to the requested sleep time (and not, say, days larger).
> 
> In #87907, there was discussion of switching Windows to use something else to resolve the inconsistency instead. That API would be `QueryUnbiasedInterruptTimePrecise` on Windows 10+. Unfortunately, the precise variant did not exist in Windows 7 or Windows 8--`QueryUnbiasedInterruptTime` has only 15ms precision (by default).

Additionally, that still wouldn't bring [OSX](https://github.com/rust-lang/rust/pull/87907#issuecomment-910221189) to be coherent with the rest of the world if we did that.

To avoid cluttering the PR itself, I think further discussion about *whether* we make this change should probably occur back on the [issue](https://github.com/rust-lang/rust/issues/87906).

jstarks:
> To avoid cluttering the PR itself, I think further discussion about whether we make this change should probably occur back on the issue.

This PR is in final review period, though. I'm concerned if the discussion happens elsewhere then it will be missed. Apologies if this is not the convention.

> Again, I would argue that it is an appropriate default, and Go made this decision due to being almost exclusively used in environments which rarely sleep.

I can understand that argument, and I might agree in a vacuum. But consistency of timer behavior _within_ a platform is important, arguably more important than consistency across platforms. Is it OK that `sleep` is inconsistent with `Instant`? Or would you propose changing that to use `CLOCK_BOOTTIME`, too?

maurer:
> I can understand that argument, and I might agree in a vacuum. But consistency of timer behavior _within_ a platform is important, arguably more important than consistency across platforms. Is it OK that `sleep` is inconsistent with `Instant`? Or would you propose changing that to use `CLOCK_BOOTTIME`, too?

`sleep` has never guaranteed a prompt wakeup absent other special configuration. The only guarantee `sleep` gives is that you will advance *at least* the specified amount. If `sleep` operates on the `CLOCK_MONOTONIC` semantics, strictly more time will pass on `CLOCK_BOOTTIME` as measured, so the call is still within spec. The only invalid configuration with regards to sleep is to use `CLOCK_BOOTTIME` for sleep but `CLOCK_MONOTONIC` for keeping time.

m-ou-se:
We discussed this in the libs meeting just now. Given the comments that were posted here during the final comment period, we've decided to hold off on merging it, since it doesn't seem obvious anymore that this change is the right one. (This isn't a decision that we *don't* want this change. We could start a new FCP once these comments are discussed and resolved.)

g2p:
Attempting to summarize things.

The clock Instant is based on only has to be monotonic. In practice it is something like boottime or uptime (both can be implemented with rdtsc and offsets that are updated after resume). Observed behavior will therefore depend on the platform as well as the device (servers rarely suspend, phones do). Portability issues would single out platforms that implement suspend but provide no way to measure boottime.

Documentation of the syscalls [exists](https://doc.rust-lang.org/std/time/struct.Instant.html#underlying-system-calls) (it needs to be combined with platform documentation) but is explicitly not binding; the syscalls may change.

Rust primitives for sleeping (`std::thread::sleep` and `std::sync::CondVar`) are based on `Duration`, not `Instant`. While Instants are anchored to a specific clock (one that can be observed by substracting Instants to get a Duration), sleep primitives can be offered for any clock (eg exposing Linux [`clock_nanosleep`](https://man7.org/linux/man-pages/man2/clock_nanosleep.2.html)/`timerfd_create`/`timer_create`) without requiring the standard library to expose new types.

There is past experience from Linux and Go which was brought up here.

Specifically, @jstarks works for Microsoft and [argued](https://github.com/golang/go/issues/35447#issuecomment-552115687) in the Go [issue tracker](https://github.com/golang/go/issues/35482) for Go on Windows not to change its default sleep primitives to switch from uptime to boottime, because some programs written before the change would assume they could make forward progress after calling sleep and then break after a resume.

This PR changes the definition of Instant::now() without affecting any sleep primitive, the objection to the Go change doesn't apply.

The Linux change to CLOCK_MONOTONIC, from an uptime clock to a boottime clock, was [reverted](https://lwn.net/Articles/752757/)  because it affected programs that were sleeping before suspend and woke up simultaneously at resume.

The change here, changing `Instant::now` without changing the various sleep primitives, would not trigger this kind of breakage.  The new behaviour presumably has been tested because it matches Rust on Windows: sleeping goes past suspends but changes to wall time are still observable in `Instant::now()`.

I'm in favour of this PR because it makes the Instant type more useful, in particular for programs that need to observe the passage of time (starting with any kind of distributed system). It also takes advantage of the way the type system only uses Duration for sleeps.

joshtriplett:
I would prefer if Instant remains consistent with the time used for sleeps and timeouts.

g2p:
> I would prefer if Instant remains consistent with the time used for sleeps and timeouts.

Sleep only guarantees that at least a certain duration has elapsed.  Sleeping close to the specified duration can't be guaranteed unless either you require the machine to never suspend or you try to make suspend unobservable.

Also, Instant doesn't currently match uptime on macOS or Windows.

cuviper:
FYI, #95026 is increasing the minimum to kernel 3.2, and then we could use `CLOCK_BOOTTIME` unconditionally.

hellow554:
thanks for the update @cuviper 
I changed the PR :)

cuviper:
@hellow554 I meant that it could be unconditional *on Linux*, but it still needs to be `cfg`'ed for other targets. You won't need the fallback logic though, just directly use the "best" choice for each target.

hellow554:
Thanks for the ping.

I `cfg`ed the `CLOCK_BOOTTIME` call to `linux` only. Any other OS I should include? Android, fuchsia, vxworks, bdf, redox?

ernestask:
Android for sure, for things like https://github.com/WireGuard/wireguard-android/blob/c762bc3120c54467ab25665214a96f539b9f9139/tunnel/tools/libwg-go/goruntime-boottime-over-monotonic.diff.

thomcc:
We could remember whether or not it's failed before, and use monotonic otherwise. We do something like this for GRND_INSECURE: https://github.com/rust-lang/rust/blob/master/library/std/src/sys/unix/rand.rs#L49-L57.

JohnCSimon:
Ping from triage: 
@hellow554 Can you please post the status of this PR?

hellow554:
I'm honestly not sure what the status is.
I applied all the concerns.

IIRC we had at some point a fallback and it was mentioned, that it should be removed, because it's not needed. 

Can I please get some feedback for that? I'm willing to implement a fallback, but I don't want to implement it and revert it a day later :)

ChrisDenton:
I'll try changing the labels to make it clear that you're waiting on a decision.

@rustbot label -S-waiting-on-author +S-waiting-on-team

maurer:
Please add `"android"` to the list of CFG'd OSes for using `CLOCK_BOOTTIME` - we currently carry a [patch](https://android.googlesource.com/toolchain/android_rust/+/refs/heads/master/patches/rustc-0018-Switch-Instant-to-use-CLOCK_BOOTTIME.patch) to `std` in Android to use this clock for `Instant`.

I'm not on the libs team, but I see no reason we'd need a fallback path now that #95026 is merged - we no longer support any of the Linux systems where `CLOCK_BOOTTIME` does not exist.

cuviper:
We don't need a fallback anymore -- we just need to decide **if** we want to change this after all.

faern:
Can we please make this happen now? My impression is that we had agreed on switching to `CLOCK_BOOTTIME`, we just needed to fix the compatibility issues. And those are long gone now.

JohnCSimon:
@hellow554 - PR has merge conflicts

mati865:
> We don't need a fallback anymore -- we just need to decide **if** we want to change this after all.

My impression is this PR is still waiting on the decision.

hellow554:
That's an error I can't explain. Maybe somebody can?

```diff
-    Instant {
-        t: Timespec::now(if cfg!(target_os = "linux") {
-            libc::CLOCK_BOOTTIME
-        } else {
-            libc::CLOCK_MONOTONIC
-        }),
-    }
+    #[cfg(target_os = "macos")]
+    const clock_id: libc::clockid_t = libc::CLOCK_UPTIME_RAW;
+
+    #[cfg(target_os = "linux")]
+    const clock_id: libc::clockid_t = libc::CLOCK_BOOTTIME;
+
+    #[cfg(not(any(target_os = "macos", target_os = "linux")))]
+    const clock_id: libc::clockid_t = libc::CLOCK_MONOTONIC;
+
+    Instant { t: Timespec::now(clock_id) }
```
that's not a lot of changes and certainly no changes that should do have an impact. Maybe something else changed in the meantime?

maniwani:
It looks like Miri tests are failing when emulating `x86_64-unknown-linux-gnu`. I believe the error is that Miri doesn't recognize `libc::CLOCK_BOOTTIME`.

I think it needs to be added to the shims: 
https://github.com/rust-lang/miri/blob/master/src/shims/time.rs

RalfJung:
> There was a discussion on https://github.com/rust-lang/rust/pull/87907 that all platforms expect
linux advance its monotonic timer if the system goes into a sleep state
(e.g. suspend or hibernate).

This is not true. macOS on aarch64 uses [CLOCK_UPTIME_RAW](https://www.manpagez.com/man/3/clock_gettime/) which does *not* increment when the system is asleep.

src/tools/miri/src/shims/time.rs
```diff
@@ -45,6 +45,9 @@ pub trait EvalContextExt<'mir, 'tcx: 'mir>: crate::MiriInterpCxExt<'mir, 'tcx> {
                 relative_clocks = vec![
                     this.eval_libc_i32("CLOCK_MONOTONIC")?,
                     this.eval_libc_i32("CLOCK_MONOTONIC_COARSE")?,
+                    // This is the equivalent to `CLOCK_UPTIME_RAW` in the macos section
+                    // We support it, because std relies on it
```
RalfJung, line 49: This comment is backwards -- BOOTTIME you say *does* increment when the machine is asleep, but CLOCK_UPTIME_RAW does not.

If anything, it seems like CLOCK_BOOTTIME on linux is like CLOCK_MONOTONIC on macOS, and CLOCK_MONOTONIC on linux is like CLOCK_UPTIME_RAW on macos?

RalfJung: Yeah judging from a quick search, looks like MONOTONIC on Linux does not increment when sleeping, but on macos it does. It also looks before this PR we are actually consistent (both Linux and macos use clocks that do not count when the system is asleep), and this PR makes us inconsistent by changing only Linux.

Or maybe I am looking at the wrong docs? [This](https://opensource.apple.com/source/Libc/Libc-1158.1.2/gen/clock_gettime.3.auto.html) seems to look reasonably official though for macos.

library/std/src/sys/unix/time.rs
```diff
@@ -291,8 +291,13 @@ mod inner {
         pub fn now() -> Instant {
             #[cfg(target_os = "macos")]
             const clock_id: libc::clockid_t = libc::CLOCK_UPTIME_RAW;
-            #[cfg(not(target_os = "macos"))]
+
+            #[cfg(target_os = "linux")]
+            const clock_id: libc::clockid_t = libc::CLOCK_BOOTTIME;
```
RalfJung, line 296: Should we have documentation saying that `Instant`, when possible, will use a clock that keeps going when the system is asleep -- and list on which targets this is definitely the case?

src/tools/miri/src/shims/time.rs
RalfJung, line 49:
```suggestion
                    // Unlike `CLOCK_MONOTONIC`, this *is* supposed to increment during system sleep...
                    // but that's not really something a program running inside Miri can tell, anyway.
                    // We support it, because std relies on it
```

library/std/src/sys/unix/time.rs
CryZe: When I fixed it for my crate I looked into most platforms: https://github.com/LiveSplit/livesplit-core/blob/092e336e24abd26e97e5a2e55abb0bbd97ff0b44/src/platform/normal/mod.rs#L6

Formatted as Markdown: https://gist.github.com/CryZe/afb052b82839d36d5bdbe4e88551f4dc

RalfJung: Okay so it seems like after this PR it's pretty much "everything except macOS"?

CryZe: Fuchsia is also problematic as they specifically defined their monotonic to behave like Linux's... but they didn't provide an alternative BOOTTIME one yet. Also the PR only checks for "linux" specific and not Linux derivatives such as Android, which it probably should (just not emscripten as it errors out on BOOTTIME atm).

RalfJung: Okay, doesn't look like this actually closes https://github.com/rust-lang/rust/issues/87906 quite yet then.

m-ou-se:
This issue was [discussed](https://rust-lang.zulipchat.com/#narrow/stream/259402-t-libs.2Fmeetings/topic/Meeting.202022-12-07/near/314496929) again in the libs meeting last week. There is currently no consensus for switching to CLOCK_BOOTTIME.

One of the points that was brought up is that many Linux APIs like futex(), futex_waitv(), and pthread_cond_t only support CLOCK_REALTIME and CLOCK_MONOTONIC. In general, it seems that CLOCK_MONOTONIC is the 'default' monotonic clock on Linux in many situations, unlike CLOCK_BOOTTIME. It seems best for std's Instant to just use whatever is the 'standard' monotonic clock on each platform.

What was also mentioned is that without providing additional guarantees that hold across all platforms, switching just one platform doesn't seem that useful; users still wouldn't be able to rely on the exact behaviour.

hellow554:
So I'm gonna close this.
Feel free to return to this PR whenever we decide elsewise, but for now, it's okay to allow this PR to rest :)

PR #53571, opened 2018-08-21
Fix unsoundness for VecDeque

MaloJaffre:
See individual commit for more details.

r? @RalfJung.

Fixes https://github.com/rust-lang/rust/issues/53566, fixes https://github.com/rust-lang/rust/issues/53529

PR #53564, opened 2018-08-21
Reoptimize VecDeque::append

MaloJaffre:
~Unfortunately, I don't know if these changes fix the unsoundness mentioned in #53529, so it is stil a WIP.
This is also completely untested.
The VecDeque code contains other unsound code: one example : [reading unitialized memory](https://play.rust-lang.org/?gist=6ff47551769af61fd8adc45c44010887&version=nightly&mode=release&edition=2015) (detected by MIRI), so I think this code will need a bigger refactor to make it clearer and safer.~

Note: this is based on #53571.
r? @SimonSapin 
Cc: #53529 #52553 @YorickPeterse @jonas-schievink @Pazzaz @shepmaster.

PR #52553, opened 2018-07-19
Non-naive implementation of `VecDeque.append`

Pazzaz:
Replaces the old, simple implementation with a more manual (and **unsafe** 😱) one. I've added 1 more test and verified that it covers all 6 code paths in the function.

This new implementation was about 60% faster than the old naive one when I tried benchmarking it.

PR #88717, opened 2021-09-07
Optimize VecDeque::append

tabokie:
Optimize `VecDeque::append` to do unsafe copy rather than iterating through each element.

On my `Intel(R) Xeon(R) CPU E5-2630 v4 @ 2.20GHz`, the benchmark shows 37% improvements:
```
Master:
custom-bench vec_deque_append 583164 ns/iter
custom-bench vec_deque_append 550040 ns/iter

Patched:
custom-bench vec_deque_append 349204 ns/iter
custom-bench vec_deque_append 368164 ns/iter
```

Additional notes on the context: this is the third attempt to implement a non-trivial version of `VecDeque::append`, the last two are reverted due to unsoundness or regression, see:
- https://github.com/rust-lang/rust/pull/52553, reverted in https://github.com/rust-lang/rust/pull/53571
- https://github.com/rust-lang/rust/pull/53564, reverted in https://github.com/rust-lang/rust/pull/54851

Both cases are covered by existing tests.

Signed-off-by: tabokie <xy.tao@outlook.com>

### DIFF
--- a/library/alloc/src/collections/vec_deque/mod.rs
+++ b/library/alloc/src/collections/vec_deque/mod.rs
@@ -417,24 +417,23 @@ impl<T, A: Allocator> VecDeque<T, A> {
         }
     }
 
-    /// Append all values from `src` to `self`, wrapping around if needed.
+    /// Append all values from `src` to `dst`, wrapping around if needed.
     /// Assumes capacity is sufficient.
     #[inline]
-    unsafe fn append_slice(&mut self, src: &[T]) {
-        debug_assert!(self.len() + src.len() + 1 <= self.cap());
-        let head_room = self.cap() - self.head;
-        if self.head < self.tail || src.len() <= head_room {
+    unsafe fn copy_slice(&mut self, dst: usize, src: &[T]) {
+        debug_assert!(src.len() <= self.cap());
+        let head_room = self.cap() - dst;
+        if src.len() <= head_room {
             unsafe {
-                ptr::copy_nonoverlapping(src.as_ptr(), self.ptr().add(self.head), src.len());
+                ptr::copy_nonoverlapping(src.as_ptr(), self.ptr().add(dst), src.len());
             }
         } else {
             let (left, right) = src.split_at(head_room);
             unsafe {
-                ptr::copy_nonoverlapping(left.as_ptr(), self.ptr().add(self.head), left.len());
+                ptr::copy_nonoverlapping(left.as_ptr(), self.ptr().add(dst), left.len());
                 ptr::copy_nonoverlapping(right.as_ptr(), self.ptr(), right.len());
             }
         }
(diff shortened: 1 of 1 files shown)


leonardo-m:
Is Miri silent?

library/alloc/src/collections/vec_deque/mod.rs
```diff
@@ -2088,8 +2108,14 @@ impl<T, A: Allocator> VecDeque<T, A> {
...
-        // naive impl
-        self.extend(other.drain(..));
+        self.reserve(other.len());
+        unsafe {
+            let (left, right) = other.as_slices();
+            self.append_slice(left);
+            self.append_slice(right);
+        }
+        // Silently drop values in `other`.
+        other.tail = other.head;
```
CryZe, line 2118: This doesn't seem "panic safe", as in, if append_slice were to panic, then the elements would exist in both VecDeques afterwards, causing double frees and other corruptions. And indeed you even added an explicit panic (assert) in append_slice, so it can indeed panic.

tabokie: You're right, I haven't thought of that.
But I wonder if it's always possible to write panic-safe code. Here I could `other.wrap_add(other.tail, some_len)` after each `append_slice` to make sure values are dropped from `other` only after they are copied and `self.head` is updated. But that relies on `wrap_add` no panicking, which sadly not the case.

tabokie:
> Is Miri silent?

Sorry I'm new here, are you referring to running `./x.py test src/tools/miri`? If so the results look good.

library/alloc/src/collections/vec_deque/mod.rs
the8472: you could append both slices in a single function, then you can do all necessary calculations that might panic up-front before doing any assignments. That way the structs will either be in their old or in their new state.

tabokie: Comment addressed~ Now pointers are updated after copying.

the8472:
I think that only runs the dedicated tests in the miri test directory. You can also run all `alloc` tests under miri. https://github.com/rust-lang/miri-test-libstd

tabokie:
> I think that only runs the dedicated tests in the miri test directory. You can also run all `alloc` tests under miri. https://github.com/rust-lang/miri-test-libstd

Passed.

m-ou-se:
Thanks!

@bors r+

tabokie:
Kindly ping @m-ou-se, do you know why bors hasn't attempted to merge this PR yet? 5 days of inactivity seems unusual. Is it because I appended one commit after your review?

Mark-Simulacrum:
@bors r=m-ou-se

Yes, pushing code will invalidate the approved state.

PR #86986, opened 2021-07-08
extend `simplify_type`

lcnr:
might cause a slight perf inprovement and imo more accurately represents what types there are.

considering that I was going to use this in #85048 it seems like we might need this in the future anyways :shrug:

PR #88718, opened 2021-09-07
fix non_blanket_impls iteration order 

lcnr:
We sometimes iterate over all non_blanket_impls, not sure if this is observable outside
of error messages (i.e. as incremental bugs), but it hopefully fixes the underlying issue of #86986.

It does not fix the issue of #86986, as the iteration order still isn't stable between different targets

r? @nikomatsakis @Aaron1011


lcnr:
that doesn't seem like an acceptable regression here, will have to do this differently

nikomatsakis:
OK. Closing this PR, then.

PR #88720, opened 2021-09-07
Rustdoc coverage fields count

GuillaumeGomez:
Follow-up of #88688.

Instead of requiring enum tuple variant fields and tuple struct fields to be documented, we count them if they are documented, otherwise we don't include them in the count.

r? @Manishearth


Manishearth (commented):
I don't understand why we need an additional hashmap: is there a reason we cannot look at the parent of the node and check what kind of field it is?

GuillaumeGomez:
I feel very stupid not doing that. Updating.

GuillaumeGomez:
Rewrote using the direct parent through hir map this time.

Manishearth (approved):
r=me

PR #88726, opened 2021-09-07
Fix typo in `const_generics` replaced with `adt_const_params` note

MingweiSamuel:


wesleywiser (approved):
Nice catch! Thanks for fixing it.

Issue #73328, opened 2020-06-13
Document guarantees (or lack thereof) regarding sign, quietness, and payload of `NaN`s


ecstatic-morse:
NaNs can behave in surprising ways. On top of that, a [very common target is inherently buggy](https://github.com/rust-lang/rust/issues/114479) in [more than one way](https://github.com/rust-lang/rust/issues/115567). But on all other targets we actually follow fairly clear, if improperly documented, rules. See [here](https://github.com/rust-lang/rust/issues/73328#issuecomment-1666048562) for the current status.

### Original issue

Several issues have been filed about surprising behavior of NaNs.

- #55131, in which the sign of the result of `0.0 / 0.0` changed depending on whether the right-hand side came from a function argument or a literal.
- #73288/#46948, in which the result of `f32::from_bits(x).to_bits()` was not always equal to `x`.

The root cause of these issues is that LLVM [does not guarantee that NaN payload bits are preserved](https://reviews.llvm.org/D54353). Empirically, this applies to the signaling/quiet bit as well as (surprisingly) the sign bit. At least one LLVM developer  seems [open to changing this](http://lists.llvm.org/pipermail/llvm-dev/2018-November/127670.html), although doing so may not be easy.

Unless we are prepared to guarantee more, we should do a better job of documenting that, besides having all 1s in the exponent and a non-zero significand, the bitwise value of a NaN is unspecified and may change at any point during program execution. In particular, the [`from_bits`](https://doc.rust-lang.org/std/primitive.f32.html#method.from_bits) method on `f32` and `f64` types currently states:

> This is currently identical to transmute::<u32, f32>(v) on all platforms.

and 

> this implementation favors preserving the exact bits. This means that any payloads encoded in NaNs will be preserved

These statements are misleading and should be changed.

We may also want to add documentation to `{f32,f64}::NAN` to this effect, see https://github.com/rust-lang/rust/issues/52897#issuecomment-496672336.

cc #10186?

ecstatic-morse:
This also affects the documentation for the methods in #72568.

RalfJung:
@ecstatic-morse wrote elsewhere
> Indeed. The underlying cause is clear. I wonder what we should do here, though? Does Rust currently guarantee that extended precision is not used for operations on f64? If so, this is technically a miscompilation. However, I don't know whether it's worth fixing. Maybe we should just document the status quo and move on?

[...]

Issue #73288, opened 2020-06-12
i686 floating point behavior does not agree with unit tests in debug mode


ecstatic-morse:
Compiling and running the following code with `-C opt-level=0` for 32-bit Linux on a recent nightly (2020-06-07) results in a failing assertion.

```rust
fn main() {
    let masked_nan2 = f32::NAN.to_bits() ^ 0x0055_5555;
    assert_eq!(f32::from_bits(masked_nan2).to_bits(), masked_nan2);
}
```

```sh
rustc +nightly --target=i686-unknown-linux-gnu test.rs && ./test
```

Output:

```
thread 'main' panicked at 'assertion failed: `(left == right)`
  left: `2144687445`,
 right: `2140493141`', test.rs:3:5
```

However, this code is taken directly from a unit test for `libstd` added in #46012, so presumably it should succeed.

https://github.com/rust-lang/rust/blob/59493917be3e87e1dfb44a9ccb66a9f9b17228e6/src/libstd/f32.rs#L1508-L1514

RalfJung:
This is probably caused by x87 instructions... I think we have an issue about that somewhere already, not sure where though.
Cc @hanna-kruppe

ecstatic-morse:
#46012 added the test.

ecstatic-morse:
Indeed, after the mask, the value in `masked_nan2` is:

```
0 11111111 00101010101010101010101
    exp           significand
```
The call to `f32::from_bits` puts the same value onto the FPU stack (`ST[0]`) with an `FLD` instruction before returning.

When the return value is popped off the FPU stack with an `FSTP` instruction in `main`, the signaling/quiet bit (the most-significant bit of the significand) has been set:

```
0 11111111 10101010101010101010101
    exp           significand
```

hanna-kruppe:
Independently of x87, the intent behind the test is on shaky ground. LLVM optimizations do not generally preserve NaN bit strings, see e.g. #55131.

ecstatic-morse:
Seems like we should update the docs for `{f32,f64}::from_bits` then. Currently, they claim

> This is currently identical to transmute::<u32, f32>(v) on all platforms

and

> this implementation favors preserving the exact bits

hanna-kruppe:
The first part isn't wrong, as far as I can tell. It literally is implemented as transmute, it's just that handling any float values *in general* does not always "preserves the exact bits".

ecstatic-morse:
```rust
fn main() {
[...]

PR #88728, opened 2021-09-07
Added next_up and next_down for f32/f64.

orlp:
This is a pull request implementing the features described at https://github.com/rust-lang/rfcs/pull/3173.

### DIFF
--- a/library/core/src/num/f32.rs
+++ b/library/core/src/num/f32.rs
@@ -646,7 +646,8 @@ impl f32 {
     /// [`INFINITY`]: Self::INFINITY
     /// [`MIN`]: Self::MIN
     /// [`MAX`]: Self::MAX
-    #[unstable(feature = "float_next_up_down", issue = "none")]
+    #[unstable(feature = "float_next_up_down", issue = "91399")]
+    #[rustc_const_unstable(feature = "float_next_up_down", issue = "91399")]
     pub const fn next_up(self) -> Self {
         // We must use strictly integer arithmetic to prevent denormals from
         // flushing to zero after an arithmetic operation on some platforms.
@@ -695,7 +696,8 @@ impl f32 {
     /// [`INFINITY`]: Self::INFINITY
     /// [`MIN`]: Self::MIN
     /// [`MAX`]: Self::MAX
-    #[unstable(feature = "float_next_up_down", issue = "none")]
+    #[unstable(feature = "float_next_up_down", issue = "91399")]
+    #[rustc_const_unstable(feature = "float_next_up_down", issue = "91399")]
     pub const fn next_down(self) -> Self {
         // We must use strictly integer arithmetic to prevent denormals from
         // flushing to zero after an arithmetic operation on some platforms.
--- a/library/core/src/num/f32.rs
+++ b/library/core/src/num/f32.rs
@@ -620,6 +620,104 @@ impl f32 {
         self.to_bits() & 0x8000_0000 != 0
     }
 
+    /// Returns the least number greater than `self`.
+    ///
+    /// Let `TINY` be the smallest representable positive `f32`. Then,
+    ///  - if `self.is_nan()`, this returns `self`;
+    ///  - if `self` is [`NEG_INFINITY`], this returns [`MIN`];
+    ///  - if `self` is `-TINY`, this returns -0.0;
+    ///  - if `self` is -0.0 or +0.0, this returns `TINY`;
+    ///  - if `self` is [`MAX`] or [`INFINITY`], this returns [`INFINITY`];
+    ///  - otherwise the unique least value greater than `self` is returned.
+    ///
+    /// The identity `x.next_up() == -(-x).next_down()` holds for all `x`. When `x`
+    /// is finite `x == x.next_up().next_down()` also holds.
+    ///
+    /// ```rust
+    /// #![feature(float_next_up_down)]
+    /// // f32::EPSILON is the difference between 1.0 and the next number up.
+    /// assert_eq!(1.0f32.next_up(), 1.0 + f32::EPSILON);
+    /// // But not for most numbers.
+    /// assert!(0.1f32.next_up() < 0.1 + f32::EPSILON);
+    /// assert_eq!(16777216f32.next_up(), 16777218.0);
+    /// ```
+    ///
+    /// [`NEG_INFINITY`]: Self::NEG_INFINITY
+    /// [`INFINITY`]: Self::INFINITY
+    /// [`MIN`]: Self::MIN
+    /// [`MAX`]: Self::MAX
+    #[unstable(feature = "float_next_up_down", issue = "none")]
+    pub const fn next_up(self) -> Self {
+        // We must use strictly integer arithmetic to prevent denormals from
+        // flushing to zero after an arithmetic operation on some platforms.
+        const TINY_BITS: u32 = 0x1; // Smallest positive f32.
+        const CLEAR_SIGN_MASK: u32 = 0x7fff_ffff;
+
+        let bits = self.to_bits();
+        if self.is_nan() || bits == Self::INFINITY.to_bits() {
+            return self;
+        }
+
+        let abs = bits & CLEAR_SIGN_MASK;
+        let next_bits = if abs == 0 {
+            TINY_BITS
+        } else if bits == abs {
+            bits + 1
+        } else {
+            bits - 1
+        };
+        Self::from_bits(next_bits)
+    }
+
+    /// Returns the greatest number less than `self`.
+    ///
+    /// Let `TINY` be the smallest representable positive `f32`. Then,
+    ///  - if `self.is_nan()`, this returns `self`;
+    ///  - if `self` is [`INFINITY`], this returns [`MAX`];
+    ///  - if `self` is `TINY`, this returns 0.0;
+    ///  - if `self` is -0.0 or +0.0, this returns `-TINY`;
+    ///  - if `self` is [`MIN`] or [`NEG_INFINITY`], this returns [`NEG_INFINITY`];
+    ///  - otherwise the unique greatest value less than `self` is returned.
+    ///
+    /// The identity `x.next_down() == -(-x).next_up()` holds for all `x`. When `x`
+    /// is finite `x == x.next_down().next_up()` also holds.
+    ///
+    /// ```rust
+    /// #![feature(float_next_up_down)]
+    /// let x = 1.0f32;
+    /// // Clamp value into range [0, 1).
+    /// let clamped = x.clamp(0.0, 1.0f32.next_down());
+    /// assert!(clamped < 1.0);
+    /// assert_eq!(clamped.next_up(), 1.0);
+    /// ```
+    ///
+    /// [`NEG_INFINITY`]: Self::NEG_INFINITY
+    /// [`INFINITY`]: Self::INFINITY
+    /// [`MIN`]: Self::MIN
+    /// [`MAX`]: Self::MAX
+    #[unstable(feature = "float_next_up_down", issue = "none")]
+    pub const fn next_down(self) -> Self {
+        // We must use strictly integer arithmetic to prevent denormals from
+        // flushing to zero after an arithmetic operation on some platforms.
+        const NEG_TINY_BITS: u32 = 0x8000_0001; // Smallest (in magnitude) negative f32.
+        const CLEAR_SIGN_MASK: u32 = 0x7fff_ffff;
+
+        let bits = self.to_bits();
+        if self.is_nan() || bits == Self::NEG_INFINITY.to_bits() {
+            return self;
+        }
+
+        let abs = bits & CLEAR_SIGN_MASK;
+        let next_bits = if abs == 0 {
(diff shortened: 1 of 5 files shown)


kpreid:
In the RFC you mention “Repeatedly calling next_up on a negative number will iterate over all values above it, with the exception of +0.0, only -0.0 will be visited.” I think this point would be good to make explicitly in the documentation of these functions — it is a place where the two goals “exhaust the space of float-representable numbers” and “move along the number line in the smallest nonzero step” differ, and it's worth mentioning this to, if nothing else, prompt the programmer to think about which thing they're doing and whether the function's inherent behavior is correct for their use case.

scottmcm:
:+1: to having these.  I think we should explicitly mention that these are expected to match IEEE's `nextUp` and `nextDown` operations, and keep the `next_up` and `next_down` names.

(I don't think we'll find any name that's substantially better, so sticking with precedent seems best.)

orlp:
So investigating why my pull request failed, I'm not sure that it's my fault. This fails on i586:

```rust
f32::from_bits(0x7F955555).to_bits() == 0x7F955555
```

This violates the standard library guarantees, which say that `from_bits` and `to_bits` are transmutes.

scottmcm:
NANs are always a bit weird, and pre-i686 floats are particularly annoying.  You might be interested in the conversations in https://github.com/rust-lang/rust/issues/73328

But honestly for now I'd just say `cfg` out that test on i586.

orlp:
@scottmcm I managed to rewrite my test case to simply avoid the issue altogether I believe by converting to `f32` earlier:

```rust
// Check that NaNs roundtrip.
let nan0 = f32::NAN;
let nan1 = f32::from_bits(f32::NAN.to_bits() ^ 0x002a_aaaa);
let nan2 = f32::from_bits(f32::NAN.to_bits() ^ 0x0055_5555);
assert_eq!(nan0.next_down().to_bits(), nan0.to_bits());
assert_eq!(nan1.next_down().to_bits(), nan1.to_bits());
assert_eq!(nan2.next_down().to_bits(), nan2.to_bits());
```

Now the correctness of the testcase does not rely on `f32::from_bits(u).to_bits() == u`.

I also rebased to resolve the merge conflicts so this should be ready to merge once again.

orlp:
I am *really* puzzled how this could fail. Note the test case that fails:

```rust
let nan0 = f32::NAN;
let nan1 = f32::from_bits(f32::NAN.to_bits() ^ 0x002a_aaaa);
let nan2 = f32::from_bits(f32::NAN.to_bits() ^ 0x0055_5555);
assert_eq!(nan0.next_down().to_bits(), nan0.to_bits());
assert_eq!(nan1.next_down().to_bits(), nan1.to_bits());
assert_eq!(nan2.next_down().to_bits(), nan2.to_bits()); // Fails.
```

And the implementation:

```rust
pub const fn next_down(self) -> Self {
    // Ommitted for brevity...
    if self.is_nan() || bits == Self::INFINITY.to_bits() {
        return self;
    }
    // Ommitted for brevity...
```

Note that we actually just return `self` when `self` is a nan. So how can `nan2.next_down().to_bits() == nan2.to_bits()` fail?

I can only imagine four scenarios:

1. `f32::from_bits(f32::NAN.to_bits() ^ 0x0055_5555)` somehow does not form a nan value even though it's only modifying the odd mantissa bits of what used to be a nan value.
2. `f32::is_nan` is bugged.
3. `f32::to_bits` is non-deterministic.
4. A compiler bug.

I have no idea how to get access to a `dist-i586-gnu-i586-i686-musl` platform to reproduce for some investigation to find which of these scenarios occurs. The testcases pass on the `i586-pc-windows-msvc` and `i686-pc-windows-msvc` targets.

For those who want a stand-alone testcase, see this playground: https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=737192093518dcc3a3d3e107dd5f2a31

orlp:
There's a fifth scenario @SkiFire13 suggests as a possibility: the floating point value `nan2` canonicalizes when passed to `next_down` because it has to be passed on the stack, but does not do so when passed directly to `to_bits`.

SkiFire13:
I think the issue is the same as #73288, except now debug mode is not required, probably due to LLVM not inlining `nan2.next_down()`.

> because it has to be passed on the stack

It has to be passed on the FPU stack, which means it has to become extended-precision and then truncated back to 32 bits. Since the latter is an arithmetic operation, it has to be canonicalized. See https://github.com/rust-lang/rust/issues/73328#issuecomment-692186924

scottmcm:
@bors r-

Again, the easiest option here might just be to skip these on i586, since there's lots of float stuff that doesn't work the same there.

orlp:
@scottmcm Do you have a suggested way of checking for that platform?

scottmcm:
Hmm, I was expecting something like `cfg!(target_cpu = "i586")`, but it looks like maybe that doesn't exist? <https://doc.rust-lang.org/reference/attributes/codegen.html#the-target_feature-attribute>

Maybe you can fake it with `x86`-but-not-`sse` or something.

library/std/src/f32/tests.rs
```diff
@@ -287,6 +287,69 @@ fn test_is_sign_negative() {
     assert!((-f32::NAN).is_sign_negative());
 }
 
+#[test]
```
Urgau, line 290: Maybe this ? This seems enough to [only exclude](https://godbolt.org/z/KarxqzEna) `i586-unknown-linux-gnu`.
```suggestion
#[test]
#[cfg_attr(not(all(target_arch = "x86", not(target_feature = "fxsr"))), ignore)]
```

orlp: What is the `fxsr` target feature? Either way I would only exclude the NaN block from the test, not the whole test.

Urgau: From what I can tell this feature tell if these float `FXSAVE, FXRESTOR` instructions are available.
This CPU feature is not be available for the `pentium` cpu which is the base cpu of `i586-unknown-linux-gnu` but is available after that. This was the only difference I could find to only exclude the `i586-unknown-linux-gnu` target.

orlp:
I'm going to try one more thing before going into conditional compilation to see if it works: forcibly pass the reference NaN value through a function boundary. Sadly I can't test locally as I don't have access to a i586 platform, so I'll have to rely on the CI.

If this doesn't work I'll give @Urgau 's idea a spin.

scottmcm:
Let's give it a try
@bors r=yaahc,scottmcm

library/core/src/num/f32.rs
```diff
@@ -620,6 +620,104 @@ impl f32 {
...
+    /// // But not for most numbers.
+    /// assert!(0.1f32.next_up() < 0.1 + f32::EPSILON);
+    /// assert_eq!(16777216f32.next_up(), 16777218.0);
+    /// ```
+    ///
+    /// [`NEG_INFINITY`]: Self::NEG_INFINITY
+    /// [`INFINITY`]: Self::INFINITY
+    /// [`MIN`]: Self::MIN
+    /// [`MAX`]: Self::MAX
+    #[unstable(feature = "float_next_up_down", issue = "none")]
```
Urgau, line 649: All of the `[unstable]` attributes are missing a tracking issue. It can be created by following the template [here](https://github.com/rust-lang/rust/issues/new?assignees=&labels=C-tracking-issue%2C+T-libs-api&template=library_tracking_issue.md&title=Tracking+Issue+for+XXX).

library/core/src/num/f32.rs
```diff
@@ -620,6 +620,104 @@ impl f32 {
...
+    pub const fn next_up(self) -> Self {
```
Urgau, line 650: They are also all missing the `[rustc_const_unstable]` attribute (which is not currently enforced due to a bug in rustc).
```suggestion
    #[rustc_const_unstable(feature = "float_next_up_down", issue = "?")]
    pub const fn next_up(self) -> Self {
```

yaahc:
I've gone ahead and merged the RFC and setup a tracking issue, https://github.com/rust-lang/rust/issues/91399. Could you update the unstable attributes to refer to this issue?

orlp:
I can. And it seems my attempted fix was not successful... What a nasty platform that `i586` is. Oh well, conditional compilation it is.

library/core/src/num/f32.rs
orlp: Had no idea this was a thing, will add them.

library/core/src/num/f32.rs
orlp: yaahc created one, will refer to them.

library/std/src/f32/tests.rs
orlp: I don't want to skip the entire test case on this platform, only the NaN round-trip, so I'm just going to conditionally compile out the NaN payload block only.

orlp:
@scottmcm I've given up trying to cleverly construct the test around i586, so I've just added conditional compilation now. If @Urgau 's suggested config works it should pass all tests.

scottmcm:
🤞

@bors r=yaahc,scottmcm

Urgau:
> ... I don't understand `dist-i586-gnu-i586-i686-musl`. I guess it also _sometimes_ flushes denormals, but not on the reference values. Probably something to do with normalization only on function boundaries again.

Btw, my `cfg` block only covers `i586-unknown-linux-musl` but the tests failed on `i686-unknown-linux-musl`. If you want to also ignore that one: `#[cfg(not(target_arch = "x86"))]` should do the trick.

> At this point I am thinking of just skipping the whole platform in the tests entirely.

If you do that, don't forget to add this to the tracking issue as an "Unresolved Questions".

> I'd test a bunch of things but since I don't have access to the platform the feedback loop through CI is... not ideal.

You could maybe try run the tests (from you own binary or directly from std) with `qemu-i386-static` from `qemu-user-static` (in the repo of all major distributions).

matthiaskrgr:
Still failing in ci I guess: https://github.com/rust-lang/rust/pull/91983#issuecomment-995379026

@bors rollup=never r-

orlp:
I don't know how it was marked to be ready to merged since we knew it was failing two weeks ago, sorry.

The problem is that x87 floating point is very problematic for this set of functions. At first I thought it was just NaN bits not being preserved, but it appears that it's also flushing denormals to zero, not just when doing arithmetic operations, but also *when passing values through function boundaries*.

I think I'll disable testing on x87 floating point platforms for now and indeed add an unresolved question to the tracking issue. I honestly don't know what to do with this platform.

matthiaskrgr:
Oh strange, you're right I didn't see that.
Sometimes bors acts funny and the queue shows prs that were already unapproved as approved, guess that was one of those times.
Sorry for the noise... :smile:

library/core/src/num/f32.rs
```diff
@@ -620,6 +620,106 @@ impl f32 {
...
+    ///
+    /// Let `TINY` be the smallest representable positive `f32`. Then,
+    ///  - if `self.is_nan()`, this returns `self`;
+    ///  - if `self` is [`NEG_INFINITY`], this returns [`MIN`];
+    ///  - if `self` is `-TINY`, this returns -0.0;
+    ///  - if `self` is -0.0 or +0.0, this returns `TINY`;
+    ///  - if `self` is [`MAX`] or [`INFINITY`], this returns [`INFINITY`];
+    ///  - otherwise the unique least value greater than `self` is returned.
+    ///
+    /// The identity `x.next_up() == -(-x).next_down()` holds for all `x`. When `x`
```
goffrie, line 633: This is technically not true for `x = NAN`.

orlp: Why not?

orlp: Oh I think I see what you're getting at, `==` here is to be understood as a bitwise / definition equality, not floating point comparison where `NaN != NaN`.

JohnCSimon:
ping from triage:
@orlp Can you address the review comments and resolve the conversation? we have some build failures, too. 

`@rustbot ready` to switch to `S-waiting-on-review`

orlp:
@[Dylan-DPC](https://github.com/Dylan-DPC) I think I've been procrastinating on this because I don't see a good solution. Today I'll update the tracking issue with the unresolved question on what to do with x87, disable the testing for that platform for now and push my changes.

orlp:
I've posted the unresolved question to the tracking issue: https://github.com/rust-lang/rust/issues/91399#issuecomment-1045989643

Should I wait with asking a review for this until the unresolved question is resolved or do it now @JohnCSimon ?

library/core/src/num/f32.rs
orlp: Should I update this doc comment to say "for all non-NaN `x`"? Or maybe reword it to "`x.next_up()` is equivalent to `-(-x).next_down()` for all `x`" to prevent confusion with `==` failing for NaNs?

JohnCSimon:
@orlp - if you think it's ready for review, go ahead and mark it ready for review. I'm just doing PR triage.

JohnCSimon:
@orlp 
Thanks for your contribution.

@rustbot label: +S-inactive

PR #88729, opened 2021-09-07
Recover from `Foo(a: 1, b: 2)`

### LINKED ISSUE #61326: Handle `Foo { a: usize, b: usize }` used as `Foo(1, 2)` gracefully
Given the code

https://github.com/rust-lang/rust/blob/3592079765d4d76f19b3a9501de5f8cc47f11a04/src/test/ui/parser/recover-from-bad-variant.rs#L1-L14

we emit

https://github.com/rust-lang/rust/blob/3592079765d4d76f19b3a9501de5f8cc47f11a04/src/test/ui/parser/recover-from-bad-variant.stderr#L1-L21

We should emit something closer to

```
error: expected type, found `3`
  --> $DIR/recover-from-bad-variant.rs:7:26
   |
LL |     let x = Enum::Foo(a: 3, b: 4);
   |                          ^ expecting a type here because of type ascription

error[E0532]: expected tuple struct/variant, found struct variant `Enum::Foo`
  --> $DIR/recover-from-bad-variant.rs:10:9
   |
LL |         Enum::Foo(a, b) => {}
   |         ^^^^^^^^^ did you mean `Enum::Foo { a, b }`?

```

estebank:
Detect likely `struct` literal using parentheses as delimiters and emit
targeted suggestion instead of type ascription parse error.

Fix #61326.

### DIFF
--- a/compiler/rustc_parse/src/parser/expr.rs
+++ b/compiler/rustc_parse/src/parser/expr.rs
@@ -1049,6 +1049,23 @@ impl<'a> Parser<'a> {
         let mut seq = self.parse_paren_expr_seq().map(|args| {
             self.mk_expr(lo.to(self.prev_token.span), self.mk_call(fun, args), AttrVec::new())
         });
+        if let Some(expr) =
+            self.maybe_recover_struct_lit_bad_delims(lo, open_paren, &mut seq, snapshot)
+        {
+            return expr;
+        }
+        self.recover_seq_parse_error(token::Paren, lo, seq)
+    }
+
+    /// If we encounter a parser state that looks like the user has written a `struct` literal with
+    /// parentheses instead of braces, recover the parser state and provide suggestions.
+    fn maybe_recover_struct_lit_bad_delims(
+        &mut self,
+        lo: Span,
+        open_paren: Span,
+        seq: &mut PResult<'a, P<Expr>>,
+        snapshot: Option<(Self, ExprKind)>,
+    ) -> Option<P<Expr>> {
         match (seq.as_mut(), snapshot) {
             (Err(ref mut err), Some((mut snapshot, ExprKind::Path(None, path)))) => {
                 let name = pprust::path_to_string(&path);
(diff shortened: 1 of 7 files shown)


src/test/ui/parser/issue-44406.stderr
```diff
@@ -9,17 +9,25 @@ help: you can escape reserved keywords to use them as identifiers
...
+   |         ^^^^^^^^^^^^^^^
 ...
 LL |     foo!(true);
-   |          ^^^^ expected type
+   |     ----------- in this macro invocation
+   |
+   = note: this error originates in the macro `foo` (in Nightly builds, run with -Z macro-backtrace for more info)
+help: if `bar` is a struct, use braces as delimiters
+   |
+LL |         bar {  }
```
notriddle, line 24: This seems messed up. Shouldn't it be `bar { baz: $rest }` ?

estebank: This is indeed "not correct", caused by the macro context of the spans involved. I could _never_ show the suggestions if a macro is involved, but that would hide the suggestion even if the macro _isn't_ mixing fragments coming from the caller with internal spans. Then again a middle point would be to give just a `help` without a suggestion for things in macros now that I think about it 🤔

compiler/rustc_parse/src/parser/expr.rs
```diff
@@ -1031,9 +1037,56 @@ impl<'a> Parser<'a> {
...
+            Some((self.clone(), fun.kind.clone()))
+        } else {
+            None
+        };
+        let open_paren = self.token.span;
+
+        let mut seq = self.parse_paren_expr_seq().map(|args| {
             self.mk_expr(lo.to(self.prev_token.span), self.mk_call(fun, args), AttrVec::new())
         });
+        match (seq.as_mut(), snapshot) {
```
oli-obk, line 1052: Please pull this out of the happy path.

r=me with that done

estebank: very rude of you to make me fix this while Mara is speaking :)

PR #88731, opened 2021-09-07
use openat when encountering ENAMETOOLONG

### LINKED ISSUE #53339: std::fs::remove_dir_all doesn't handle paths that are too long
On Linux `unlink()` might fail with `ENAMETOOLONG` but `std::fs::remove_dir_all` doesn't handle that and ignores the error, leaving the directory as it was.

When the path is too long a parent directory should be `open()`ed instead and `unlinkat()` should be called with the relative path.

the8472:
This adds a fallback to various fs methods that handles ENAMETOOLONG by splitting the path into relative segments and then opening each segment with openat(dirfd, segment, O_PATH).

Limitations:

* not all fs methods are covered. the primary motivation is to get  remove_dir_all working to delete a deep directory structure created by accident
* currently linux-only. this could be extended to platforms which have either O_PATH or O_EXEC but there's no CI coverage for the BSDs, so I don't want to foist it on them.

fixes #53339

### DIFF
--- a/library/std/src/sys/unix/fs.rs
+++ b/library/std/src/sys/unix/fs.rs
@@ -3,7 +3,7 @@
 
 use crate::os::unix::prelude::*;
 
-use crate::ffi::{CStr, CString, OsStr, OsString};
+use crate::ffi::{CStr, OsStr, OsString};
 use crate::fmt;
 use crate::io::{self, BorrowedCursor, Error, IoSlice, IoSliceMut, SeekFrom};
 use crate::mem;
@@ -86,7 +86,11 @@ use libc::{
     lstat as lstat64, off_t as off64_t, open as open64, stat as stat64,
 };
 #[cfg(any(target_os = "linux", target_os = "emscripten", target_os = "l4re"))]
-use libc::{dirent64, fstat64, ftruncate64, lseek64, lstat64, off64_t, open64, stat64};
+use libc::{dirent64, fstat64, ftruncate64, lseek64, lstat64, off64_t, stat64};
+
+// FIXME: port this to other unices that support *at syscalls
+#[cfg(any(target_os = "linux", target_os = "android"))]
+mod dir_fd;
 
 pub use crate::sys_common::fs::try_exists;
 
@@ -269,8 +273,24 @@ pub struct ReadDir {
 }
 
 impl ReadDir {
+    #[cfg(not(any(target_os = "android", target_os = "linux")))]
     fn new(inner: InnerReadDir) -> Self {
         Self { inner: Arc::new(inner), end_of_stream: false }
+    }
+
+    #[cfg(any(target_os = "android", target_os = "linux"))]
+    fn from_dirp(ptr: *mut libc::DIR, root: PathBuf) -> ReadDir {
+        let inner = InnerReadDir { dirp: Dir(ptr), root };
+        ReadDir {
+            inner: Arc::new(inner),
+            #[cfg(not(any(
+                target_os = "solaris",
+                target_os = "illumos",
+                target_os = "fuchsia",
+                target_os = "redox",
+            )))]
+            end_of_stream: false,
+        }
     }
 }
 
@@ -1070,7 +1090,9 @@ impl File {
             use crate::io::ErrorKind;
             match result {
                 Ok(file) => Ok(file),
-                Err(e) if e.kind() == ErrorKind::InvalidFilename => open_deep(path, opts),
+                Err(e) if e.kind() == ErrorKind::InvalidFilename => {
+                    dir_fd::open_deep(None, path, opts)
+                }
                 Err(e) => Err(e),
             }
         };
@@ -1078,6 +1100,7 @@ impl File {
         result
     }
 
+    #[cfg(not(any(target_os = "linux", target_os = "android")))]
     pub fn open_c(
         dirfd: Option<BorrowedFd<'_>>,
         path: &CStr,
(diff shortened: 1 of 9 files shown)


library/std/src/sys/unix/fs.rs
```diff
@@ -751,20 +759,56 @@ impl OpenOptions {
...
+                // some platforms (like macOS, where `open64` is actually `open`), `mode_t` is `u16`.
+                // However, since this is a variadic function, C integer promotion rules mean that on
+                // the ABI level, this still gets passed as `c_int` (aka `u32` on Unix platforms).
+                cvt_r(|| unsafe { open64(path.as_ptr(), flags, opts.mode as c_int) })?
+            }
+            Some(dirfd) => {
+                cfg_if::cfg_if! {
+                    if #[cfg(any(target_os = "linux", target_os = "android"))] {
+                        use libc::openat64;
+                        cvt_r(|| unsafe {
```
fbstj, line 799: does this need to also indicate that the int promotion stuff applies to it as above?

the8472: I'm not even passing mode `mode`, will fix 😔
And yes, for some 32bit androids that may be needed.

lukaslueg:
IMHO this is very close to unexpected magic behaviour. Essentially `ENAMETOLONG` is filtered, which may provide comfort, but is unexpected. There is also no way Ask-For-Forgiveness-Not-Permission, expecting and handling `ENAMETOLONG` in the caller. Maybe this should get a separate entry point.

the8472:
File *names* which are too long (>255 bytes on many filesystems) would still result in `ENAMETOOLONG`, i.e. the error is not filtered. This fallback only handles *paths* exceeding `PATH_MAX` (4096 bytes on linux).

Having a separate deep version API for all the `std::fs` functions would require them to be platform-specific extension because some platforms lack the `-at` syscalls or `O_PATH/O_EXEC`. That would be quite unergonomic, especially for the convenience methods like `fs::write`.

m-ou-se:
@joshtriplett, @cuviper, what's your opinion on this change?

joshtriplett:
Why does this need `O_PATH`, rather than `O_DIRECTORY`? The latter is supported by every kernel version we support, and I *think* it'll work here.

the8472:
Already answered on zulip, but repeating it here for the record:

O_PATH is necessary on linux because open requires permission flags to be passed (this uses O_RDONLY) but that would fail when traversing exec/search-only directories. O_PATH circumvents that because it ignores the permission flags. Some other platforms have O_EXEC that could be used instead.

@cuviper josh suggested I check with you whether RHEL happened to backport O_PATH support since the minimum kernel version the only thing keeping us from using it by default for `remove_dir_all`.

cuviper:
> @cuviper josh suggested I check with you whether RHEL happened to backport O_PATH support since the minimum kernel version the only thing keeping us from using it by default for `remove_dir_all`.

No, that wasn't backported.

However, Linux doesn't check for invalid/unknown flags in `open`/`openat` -- this is how we used to get away with passing `O_CLOEXEC` even before we updated to a sufficient minimum kernel. So if we pass `O_RDONLY | O_PATH`, it should behave the same as `O_RDONLY`.

> but that would fail when traversing exec/search-only directories.

So now we're talking about the failure of a new fallback, which means this is a corner case that didn't work before due to long paths, and will still fail (on old kernels) if it can't read. This seems acceptable to me, but should we let that `EACCES` through or return the original `ENAMETOOLONG` error?

the8472:
> So if we pass O_RDONLY | O_PATH, it should behave the same as O_RDONLY.

That may be good enough to always use openat in  `remove_dir_all` instead of the current approach + fallback for each file visited.

> This seems acceptable to me, but should we let that EACCES through or return the original ENAMETOOLONG error?

I lean towards passing them through since on newer kernels that do support O_PATH it'll be a genuine traversal error. And a complete lack of permissions is probably more common than -r+x.

the8472:
let's see if this also builds on the other unixes

@bors try

the8472:
There's some overlap with #93160 I'll wait until that lands and then trim the redundant parts.
Marking as blocked until then.

hkratz:
Should be unblocked once #95925 is merged.

JohnCSimon:
triage: merge conflicts

cuviper:
BTW, the discussion about `O_PATH` support is moot now that we require at least Linux 3.2.

workingjubilee:
https://github.com/rust-lang/rust/pull/95925 was closed so this is de facto unblocked, I guess.

the8472:
Ah yes. Though this PR was written prior to the fix for CVE-2022-21658 (which inconveniently has no associated issue...) so I'll have to go over it again to make sure it wouldn't reintroduce that.

workingjubilee:
Oh. :grimacing:

library/std/src/sys/unix/fs.rs
```diff
@@ -54,6 +54,8 @@ use libc::c_char;
 use libc::dirfd;
 #[cfg(any(target_os = "linux", target_os = "emscripten"))]
 use libc::fstatat64;
+#[cfg(all(miri, any(target_os = "linux")))]
+use libc::open64;
```
RalfJung, line 58: I'm a bit surprised that this import is only present for Miri -- that seems strange?

RalfJung:
FWIW the reason Miri needs special cases here as that the standard library doesn't expose `openat`-like functionality itself itself -- something like a way to open a file on a `ReadDir`. Without that we can't implement an `openat` shim. (Miri can run Linux targets on Windows hosts, which is quite valuable and means falling back to the host `openat` isn't really an option.)

library/std/src/sys/unix/fs.rs
the8472: There are two code-paths, one that uses openat and one that uses open. Miri only supports the latter, so the entire machinery around -at syscalls is configured-out under miri and the alternative is configured-in.

RalfJung: Ah, and there are some other import for `open64` below for other targets. The entire set of `cfg`s is just tricky to have an overview of. Makes sense.

the8472:
> FWIW the reason Miri needs special cases here as that the standard library doesn't expose `openat`-like functionality itself itself -- something like a way to open a file on a `ReadDir`

You could use [cap-std](https://docs.rs/cap-std/2.0.0/cap_std/fs/struct.Dir.html#method.open) for that which offers that functionality. Maybe it's not race-free on all platforms if they don't support the necessary primitives but I assume at least on the major unix-likes it would be using openat too.

But yeah, we should get the basics into std. Though that'll pose a compatibility question around windows since the necessary API there isn't officially stable.

JohnCSimon:
@the8472 
ping from triage - can you post your status on this PR? This PR has not received an update in a few months, and the PR itself is ... 2021 . Thank you!

the8472:
Closing for now. This will be more easily implemented on top of [ACP 259](https://github.com/rust-lang/libs-team/issues/259)

PR #88732, opened 2021-09-07
RustWrapper: avoid deleted unclear attribute methods

durin42:
These were deleted in https://reviews.llvm.org/D108614, and in C++ I
definitely see the argument for their removal. I didn't try and
propagate the changes up into higher layers of rustc in this change
because my initial goal was to get rustc working against LLVM HEAD
promptly, but I'm happy to follow up with some refactoring to make the
API on the Rust side match the LLVM API more directly (though the way
the enum works in Rust makes the API less scary IMO).

r? @nagisa cc @nikic

### DIFF
--- a/compiler/rustc_llvm/llvm-wrapper/RustWrapper.cpp
+++ b/compiler/rustc_llvm/llvm-wrapper/RustWrapper.cpp
@@ -207,18 +207,7 @@ template<typename T> static inline void AddAttribute(T *t, unsigned Index, Attri
 #if LLVM_VERSION_LT(14, 0)
   t->addAttribute(Index, Attr);
 #else
-  // TODO(durin42): we should probably surface the explicit functions to Rust
-  // instead of this switch statement?
-  switch (Index) {
-    case AttributeList::ReturnIndex:
-      t->addRetAttr(Attr);
-      break;
-    case AttributeList::FunctionIndex:
-      t->addFnAttr(Attr);
-      break;
-    default:
-      t->addParamAttr(Index-AttributeList::FirstArgIndex, Attr);
-  }
+  t->addAttributeAtIndex(Index, Attr);
 #endif
 }
 
@@ -241,18 +230,7 @@ static inline void AddCallAttributes(CallBase *Call, unsigned Index, const AttrB
 #if LLVM_VERSION_LT(14, 0)
   Attrs = Attrs.addAttributes(Call->getContext(), Index, B);
 #else
-  // TODO(durin42): we should probably surface the explicit functions to Rust
-  // instead of this switch statement?
-  switch (Index) {
-    case AttributeList::ReturnIndex:
-      Attrs = Attrs.addRetAttributes(Call->getContext(), B);
-      break;
-    case AttributeList::FunctionIndex:
-      Attrs = Attrs.addFnAttributes(Call->getContext(), B);
-      break;
-    default:
-      Attrs = Attrs.addParamAttributes(Call->getContext(), Index-AttributeList::FirstArgIndex, B);
-  }
+  Attrs = Attrs.addAttributesAtIndex(Call->getContext(), Index, B);
 #endif
   Call->setAttributes(Attrs);
 }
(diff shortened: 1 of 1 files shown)


compiler/rustc_llvm/llvm-wrapper/RustWrapper.cpp
```diff
@@ -203,29 +203,67 @@ static Attribute::AttrKind fromRust(LLVMRustAttribute Kind) {
...
+  switch (Index) {
+    case AttributeList::ReturnIndex:
+      t->addRetAttr(Attr);
+      break;
+    case AttributeList::FunctionIndex:
+      t->addFnAttr(Attr);
+      break;
+    default:
+      t->addParamAttr(Index-AttributeList::FirstArgIndex, Attr);
+  }
```
nikic, line 221: Can't we still use addAttributeAtIndex() here? I believe these methods were just (effectively) renamed, not really removed altogether.

durin42: We totally can, but it felt nicer to use the more-obviously-named versions of them since we already have to have LLVM14+ and LLVM13- codepaths. Also possibly more future-proof, though aeubanks indicated the *AtIndex() methods aren't on the chopping block...

durin42:
Oh, a linter is sad about my use of TODO: do you want me to upgrade it to a FIXME or just drop the TODO? I don't feel strongly about the API shape, but figured we should record the intent if we want the Rust side cleaned up.

compiler/rustc_llvm/llvm-wrapper/RustWrapper.cpp
nikic: I don't think it's useful to unpack the calls in this way here -- calling more specific functions is great if you know what kind of attribute you're dealing with and can forgo attribute indexes entirely, but this is a generic API. Effectively you're reimplementing the functionality of addAttributeAtIndex().

durin42: Sure, done.

compiler/rustc_llvm/llvm-wrapper/RustWrapper.cpp
```diff
@@ -203,29 +203,45 @@ static Attribute::AttrKind fromRust(LLVMRustAttribute Kind) {
...
 
 extern "C" void LLVMRustAddAlignmentCallSiteAttr(LLVMValueRef Instr,
                                                  unsigned Index,
                                                  uint32_t Bytes) {
   CallBase *Call = unwrap<CallBase>(Instr);
   AttrBuilder B;
   B.addAlignmentAttr(Bytes);
-  Call->setAttributes(Call->getAttributes().addAttributes(
-      Call->getContext(), Index, B));
+  AddCallAttributes(Call, Index, B);
```
nikic, line 244: I realize this is kind of a pre-existing problem, but why does this need a separate AddCallAttributes? Couldn't it use exactly the same code as LLVMRustAddAlignmentAttr? That is `AddAttribute(Call, Index, Attribute::getWithAlignment(Call->getContext(), llvm::Align(Bytes)));` or so? Same for the other uses of AddCallAttributes() below.

durin42: It's because these functions are getting a uint{32,64}_t in instead of an LLVMRustAttribute or similar, so we have to bounce through the AttrBuilder to go from the uint* to the Attribute. I've felt like there was probably room for cleanup both times I've been in here, but haven't had time to go wading around.

nikic: Right, but this applies the same way to the corresponding setters on `Function` (LLVMRustAddAlignmentAttr etc). Those were changed to use `Attribute::getXXX()` methods in #88289, but the same wasn't done for the call-site functions.

durin42: Done.

PR #88740, opened 2021-09-08
[perf experiment] Use `intern` instead of `intern_ref` in `direct_interners!`.

eddyb:
*Opening as draft as it's not intended to be merged, it's only for perf testing*

This is me testing the theory that the regressions measured in https://github.com/rust-lang/rust/pull/88575#issuecomment-914537788 cannot come from the main changes of that PR, but rather from the switch of `ty::Region` and `ty::Const` interning, since `-check` benchmarks are affected, while the PR mostly changes `FnAbi` (used only by codegen).

r? @ghost


eddyb:
That's another deadend, I guess. Also it looks like I keep forgetting to press the draft button.

PR #88742, opened 2021-09-08
Fix table in docblocks

GuillaumeGomez:
"Overwrite" of #88702.

Instead of adding a z-index to the sidebar (which only hides the issue, doesn't fix it), I wrap `<table>` elements inside a `<div>` and limit all chidren of `.docblock` elements' width to prevent having the scrollbar on the whole doc block.

Screenshot from 2021-09-08 15-11-24

Thanks @nbdd0121 for `overflow-x: auto;`. ;)

r? @notriddle

### DIFF
--- a/src/librustdoc/html/static/css/rustdoc.css
+++ b/src/librustdoc/html/static/css/rustdoc.css
@@ -522,6 +522,11 @@ nav.sub {
 	position: relative;
 }
 
+.docblock > * {
+	max-width: 100%;
+	overflow-x: auto;
+}
+
 .content .out-of-band {
 	flex-grow: 0;
 	text-align: right;
--- a/src/librustdoc/html/markdown.rs
+++ b/src/librustdoc/html/markdown.rs
@@ -441,6 +441,42 @@ impl<'a, I: Iterator<Item = Event<'a>>> Iterator for LinkReplacer<'a, I> {
     }
 }
 
+/// Wrap HTML tables into `<div>` to prevent having the doc blocks width being too big.
+struct TableWrapper<'a, I: Iterator<Item = Event<'a>>> {
+    inner: I,
+    stored_events: VecDeque<Event<'a>>,
+}
+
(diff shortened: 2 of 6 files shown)


src/librustdoc/html/static/css/rustdoc.css
```diff
@@ -522,6 +522,11 @@ nav.sub {
 	position: relative;
 }
 
+.docblock > * {
+	max-width: 100%;
```
nbdd0121, line 526: I think this is actually unnecessary. According to the MDN, [`overflow: auto` should introduce new block formatting context by itself](https://developer.mozilla.org/en-US/docs/Web/CSS/overflow). So if the element this applies to is a block element, then `max-width: 100%` is not needed. If the element this applies to is an inline element, then `max-width` is not useful because [it does not apply to the non-replaced inline elements](https://developer.mozilla.org/en-US/docs/Web/CSS/max-width).

So I think we just need
```css
.docblock > * {
	overflow-x: auto;
}
```

Or maybe
```css
.docblock > * {
	display: block;
	overflow-x: auto;
}
```
if we want to be absolutely sure, but this is likely not necessary since inline elements should be wrapped with `<p></p>` already.

GuillaumeGomez: I did it for images actually. :)

nbdd0121: Wouldn't images be wrapped within `<p></p>`?

GuillaumeGomez: Not always, for example:

```rust
//! <img src="whatever.png">
//!
//! ![alt text](https://github.com/adam-p/markdown-here/raw/master/src/common/images/icon48.png "Logo Title Text 1")
```

First one isn't but the second one is. The beauty of markdown. :D

GuillaumeGomez:
Let's move on then. Thanks for the @nbdd0121 ! (if there is anything else, please tell me so I r-).

@bors: r=nbdd0121 rollup

PR #88745, opened 2021-09-08
Allow missing code examples in trait impls.

### LINKED ISSUE #88741: Rustdoc shouldn't lint about missing code examples in trait impls
My suggestion is to exclude Trait implementations from the items that need to have doc code examples when using the `rustdoc::missing_doc_code_examples` lint.

## Examples

Consider the following example:
```rust
//! docs
//! ```
//! let s = SomeStruct;
//! ```

#![deny(rustdoc::missing_doc_code_examples)]

/// docs
/// ```
/// let s = SomeStruct;
/// ```
pub struct SomeStruct;

impl Clone for SomeStruct {
    fn clone(&self) -> Self { Self }
}
```
The `missing_doc_code_exmaples` lint causes the following errors:
```
error: missing code example in this documentation
  --> src/lib.rs:14:1
   |
14 | / impl Clone for SomeStruct {
15 | |     fn clone(&self) -> Self { Self }
16 | | }
   | |_^
   |
  --> src/lib.rs:6:9
   |
6  | #![deny(rustdoc::missing_doc_code_examples)]
   |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

error: missing code example in this documentation
  --> src/lib.rs:15:5
   |
15 |     fn clone(&self) -> Self { Self }
   |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

### Pros

 1. This is inconsistent with the `missing_docs` lint, which does not require documentation for Trait implementations.
 2. The documentation errors/warnings should be issued when compiling the Traits documentation (as is the case with e.g. the missing documentation lint).
 3. Once "#81775" is fixed, the missing code examples lint is inconsistent with its own behavior when using derive macro.

### Workarounds

There are two options available to avoid these documentation errors:

 1. One can replace the traits documentations:
    ```
    /// docs
    /// ```
    /// let c = SomeStruct.clone();
    /// ```
    impl Clone for SomeStruct {
        /// docs
        /// ```
        /// let c = SomeStruct.clone();
        /// ```
        fn clone(&self) -> Self { Self }
    }
    ```
    This is a bad workaround because:
     1. It cause a lot of copy-paste code duplication.
     3. In most cases the traits documentation is preferred to an implementors documentation.
     4. The Trait and it's associated methods, types, etc. can already have a documentation with code examples.
        Those are not considered by the lint, which prompts the implementor rewrite the documentation or ignore the lint.
     5. The Trait's provided methods might not have code examples (as is the case with `Clone::clone_from`).
        Those is not considered by the lint.
     6. The lint is inconsistent when it comes to required and provided methods of a trait (compare the previous two points).
        Even though both types of methods are displayed equally in the documentation the missing doc code examples lint treats them differently.
 2. One can allow missing doc code examples for each of these cases:
    ```rust
    #[allow(rustdoc::missing_doc_code_examples)]
    impl Clone for SomeStruct {
        fn clone(&self) -> Self { Self }
    }
    ```
    This workaround is also not ideal due to the amount of copy paste code duplication necessary, its inconsistency with derive macros and its inconsistency with other lints like the missing documentations lint.

### Cons

The only good case (that I can think of) for the current behavior of the lint is:
If an implementation has some special behavior (like not implementing a method of a Trait) then the author of the implementation is reminded to document this behavior by the lint.

hnj2:
Excludes Trait implementations from the items that need to have doc code examples when using the `rustdoc::missing_doc_code_examples` lint.

For details see #88741

fixes #88741

r? @jyn514

### DIFF
--- a/src/librustdoc/passes/doc_test_lints.rs
+++ b/src/librustdoc/passes/doc_test_lints.rs
@@ -10,6 +10,7 @@ use crate::core::DocContext;
 use crate::fold::DocFolder;
 use crate::html::markdown::{find_testable_code, ErrorCodes, Ignore, LangString};
 use crate::visit_ast::inherits_doc_hidden;
+use rustc_hir as hir;
 use rustc_middle::lint::LintLevelSource;
 use rustc_session::lint;
 use rustc_span::symbol::sym;
@@ -67,13 +68,32 @@ crate fn should_have_doc_example(cx: &DocContext<'_>, item: &clean::Item) -> boo
                 | clean::ImportItem(_)
                 | clean::PrimitiveItem(_)
                 | clean::KeywordItem(_)
+                // check for trait impl
+                | clean::ImplItem(clean::Impl { trait_: Some(_), .. })
         )
     {
         return false;
     }
+
     // The `expect_def_id()` should be okay because `local_def_id_to_hir_id`
     // would presumably panic if a fake `DefIndex` were passed.
     let hir_id = cx.tcx.hir().local_def_id_to_hir_id(item.def_id.expect_def_id().expect_local());
+
+    // check if parent is trait impl
+    if let Some(parent_hir_id) = cx.tcx.hir().find_parent_node(hir_id) {
+        if let Some(parent_node) = cx.tcx.hir().find(parent_hir_id) {
+            if matches!(
+                parent_node,
+                hir::Node::Item(hir::Item {
+                    kind: hir::ItemKind::Impl(hir::Impl { of_trait: Some(_), .. }),
+                    ..
+                })
+            ) {
+                return false;
+            }
+        }
+    }
+
     if cx.tcx.hir().attrs(hir_id).lists(sym::doc).has_word(sym::hidden)
         || inherits_doc_hidden(cx.tcx, hir_id)
     {
--- a/src/test/rustdoc-ui/lint-missing-doc-code-example.rs
+++ b/src/test/rustdoc-ui/lint-missing-doc-code-example.rs
@@ -70,6 +70,13 @@ pub union Union {
     b: f32,
 }
 
+// no code example and it's fine!
+impl Clone for Struct {
+    fn clone(&self) -> Self {
+        Self { field: self.field }
+    }
+}
+
 
 #[doc(hidden)]
 pub mod foo {


src/librustdoc/passes/doc_test_lints.rs
```diff
@@ -67,13 +67,28 @@ crate fn should_have_doc_example(cx: &DocContext<'_>, item: &clean::Item) -> boo
...
     {
         return false;
     }
+
     // The `expect_def_id()` should be okay because `local_def_id_to_hir_id`
     // would presumably panic if a fake `DefIndex` were passed.
     let hir_id = cx.tcx.hir().local_def_id_to_hir_id(item.def_id.expect_def_id().expect_local());
+
+    // check if parent is trait impl
+    if let Some((_, parent_node)) = cx.tcx.hir().parent_iter(hir_id).next() {
```
hnj2, line 82: I think only checking for the parent of a node is enough, but I am unsure if there is any edge case where an item more than immediately below a trait impl could run through here.

src/librustdoc/passes/doc_test_lints.rs
```diff
@@ -67,13 +67,28 @@ crate fn should_have_doc_example(cx: &DocContext<'_>, item: &clean::Item) -> boo
...
+        use rustc_hir::*;
+        if matches!(
+            parent_node,
+            Node::Item(Item { kind: ItemKind::Impl(Impl { of_trait: Some(_), .. }), .. })
```
hnj2, line 86: This might be too hard to read?

jyn514:
I don't have time to review this.

r? @GuillaumeGomez

src/librustdoc/passes/doc_test_lints.rs
GuillaumeGomez: I don't think so, unless you're documenting inside an item in a trait impl.

GuillaumeGomez: [get_parent_node](https://doc.rust-lang.org/nightly/nightly-rustc/rustc_middle/hir/map/struct.Map.html#method.get_parent_node) might be better in this case:

```suggestion
    if let Some((_, parent_node)) = cx.tcx.hir().get_parent_node(hir_id) {
```

src/librustdoc/passes/doc_test_lints.rs
```diff
@@ -67,13 +67,28 @@ crate fn should_have_doc_example(cx: &DocContext<'_>, item: &clean::Item) -> boo
...
         return false;
     }
+
     // The `expect_def_id()` should be okay because `local_def_id_to_hir_id`
     // would presumably panic if a fake `DefIndex` were passed.
     let hir_id = cx.tcx.hir().local_def_id_to_hir_id(item.def_id.expect_def_id().expect_local());
+
+    // check if parent is trait impl
+    if let Some((_, parent_node)) = cx.tcx.hir().parent_iter(hir_id).next() {
+        use rustc_hir::*;
```
GuillaumeGomez, line 83: I'm not a big fan of imports in the middle of the code, can you instead import at the top (of the function?) `use rustc_hir as hir` and then prepend the `hir` types with it please?

src/librustdoc/passes/doc_test_lints.rs
GuillaumeGomez: It's fine.

GuillaumeGomez:
Looks good overall, just a few nits and it's good to go, thanks!

src/librustdoc/passes/doc_test_lints.rs
hnj2: So [`get_parent_node` just calls `find_parent_node` and returns the crate id if nothing is found](https://doc.rust-lang.org/nightly/nightly-rustc/src/rustc_middle/hir/map/mod.rs.html#308-310) and returns a `HirId` and thus makes another `cx.tcx.hir().find` call necessary, which is basically what [`ParentHirIterator::next` does](https://doc.rust-lang.org/nightly/nightly-rustc/rustc_middle/hir/map/struct.ParentHirIterator.html#method.next).

Overall this might be more efficient but harder to read:
```rust
    if let Some(parent_node) =
        cx.tcx.hir().find_parent_node(hir_id).and_then(|id| cx.tcx.hir().find(id))
    {
        if matches!(
            parent_node,
            hir::Node::Item(hir::Item {
                kind: hir::ItemKind::Impl(hir::Impl { of_trait: Some(_), .. }),
                ..
            })
        ) {
            return false;
        }
    }
```

I am guessing the more efficient version is generally preferred?

GuillaumeGomez: I actually prefer the `find_parent_node` version, I find it more explicit. And if it's the most performant, it's even better! :)

GuillaumeGomez (approved):

GuillaumeGomez:
Looks good to me, thanks! Just waiting for CI then I r+ it.

PR #88757, opened 2021-09-08
Suggest wapping expr in parentheses on invalid unary negation

### LINKED ISSUE #88701: diagnostics when parsing `{ {10} - 1 }` is confusing
Given the following [code](https://play.rust-lang.org/?version=nightly&mode=debug&edition=2018&gist=291d09b9657643097c2214cc6c5348fe):

```rust
#![allow(unused_braces)]

fn main() {
    let _: usize = { {10} - 1 };
}
```

The current output is:

```
 --> src/main.rs:4:23
  |
4 |     let _: usize = { {10} - 1 };
  |                       ^^ expected `()`, found integer

error[E0600]: cannot apply unary operator `-` to type `usize`
 --> src/main.rs:4:27
  |
4 |     let _: usize = { {10} - 1 };
  |                           ^^^
  |                           |
  |                           cannot apply unary operator `-`
  |                           help: you may have meant the maximum value of `usize`: `usize::MAX`
  |
  = note: unsigned values cannot be negated
```

Ideally this would give the same diagnostics as `{10} + 1` gives:
```
 --> src/main.rs:4:27
  |
4 |     let _: usize = { {10} + 1 };
  |                           ^ expected expression
  |
help: parentheses are required to parse this as an expression
  |
4 |     let _: usize = { ({10}) + 1 };
  |                      +    +

 --> src/main.rs:4:23
  |
4 |     let _: usize = { {10} + 1 };
  |                       ^^ expected `()`, found integer
```

andrewhickman:
Fixes #88701

### DIFF
--- a/compiler/rustc_typeck/src/check/op.rs
+++ b/compiler/rustc_typeck/src/check/op.rs
@@ -685,6 +685,9 @@ impl<'a, 'tcx> FnCtxt<'a, 'tcx> {
                     if let Some(sp) =
                         self.tcx.sess.parse_sess.ambiguous_block_expr_parse.borrow().get(&sp)
                     {
+                        // If the previous expression was a block expression, suggest parentheses
+                        // (turning this into a binary subtraction operation instead.)
+                        // for example, `{2} - 2` -> `({2}) - 2` (see src\test\ui\parser\expr-as-stmt.rs)
                         self.tcx.sess.parse_sess.expr_parentheses_needed(&mut err, *sp);
                     } else {
                         match actual.kind() {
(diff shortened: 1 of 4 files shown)


compiler/rustc_typeck/src/check/op.rs
```diff
@@ -680,42 +680,50 @@ impl<'a, 'tcx> FnCtxt<'a, 'tcx> {
...
-                                    &format!(
-                                        "you may have meant the maximum value of `{}`",
-                                        actual
-                                    ),
-                                    format!("{}::MAX", actual),
-                                    Applicability::MaybeIncorrect,
-                                );
+
+                    let sp = self.tcx.sess.source_map().start_point(ex.span);
+                    if let Some(sp) =
```
jackh726, line 685: Can you add a comment here? Just a brief explanation of what we're doing and a reference to the test?

jackh726 (approved):
Just would like a comment and this is LGTM. It would be nice to not have another layer of nesting, but it's fine.

PR #88759, opened 2021-09-08
Add -Z panic-in-drop={unwind,abort} command-line option

Amanieu:
This PR changes `Drop` to abort if an unwinding panic attempts to escape it, making the process abort instead. This has several benefits:
- The current behavior when unwinding out of `Drop` is very unintuitive and easy to miss: unwinding continues, but the remaining drops in scope are simply leaked.
- A lot of unsafe code doesn't expect drops to unwind, which can lead to unsoundness:
  - https://github.com/servo/rust-smallvec/issues/14 
  - https://github.com/bluss/arrayvec/issues/3
- There is a code size and compilation time cost to this: LLVM needs to generate extra landing pads out of all calls in a drop implementation. This can compound when functions are inlined since unwinding will then continue on to process drops in the callee, which can itself unwind, etc.
  - Initial measurements show a 3% size reduction and up to 10% compilation time reduction on some crates (`syn`).

One thing to note about `-Z panic-in-drop=abort` is that *all* crates must be built with this option for it to be sound since it makes the compiler assume that dropping `Box<dyn Any>` will never unwind.

cc https://github.com/rust-lang/lang-team/issues/97

### DIFF
--- a/compiler/rustc_codegen_llvm/src/abi.rs
+++ b/compiler/rustc_codegen_llvm/src/abi.rs
@@ -511,7 +511,12 @@ impl<'tcx> FnAbiLlvmExt<'tcx> for FnAbi<'tcx, Ty<'tcx>> {
     }
 
     fn apply_attrs_callsite(&self, bx: &mut Builder<'a, 'll, 'tcx>, callsite: &'ll Value) {
-        // FIXME(wesleywiser, eddyb): We should apply `nounwind` and `noreturn` as appropriate to this callsite.
+        if self.ret.layout.abi.is_uninhabited() {
+            llvm::Attribute::NoReturn.apply_callsite(llvm::AttributePlace::Function, callsite);
+        }
+        if !self.can_unwind {
+            llvm::Attribute::NoUnwind.apply_callsite(llvm::AttributePlace::Function, callsite);
+        }
 
         let mut i = 0;
         let mut apply = |cx: &CodegenCx<'_, '_>, attrs: &ArgAttributes| {
--- a/compiler/rustc_metadata/src/dependency_format.rs
+++ b/compiler/rustc_metadata/src/dependency_format.rs
@@ -400,21 +400,35 @@ fn verify_ok(tcx: TyCtxt<'_>, list: &[Linkage]) {
                 continue;
             }
             let cnum = CrateNum::new(i + 1);
-            let found_strategy = tcx.panic_strategy(cnum);
-            let is_compiler_builtins = tcx.is_compiler_builtins(cnum);
-            if is_compiler_builtins || desired_strategy == found_strategy {
+            if tcx.is_compiler_builtins(cnum) {
                 continue;
             }
 
-            sess.err(&format!(
-                "the crate `{}` is compiled with the \
+            let found_strategy = tcx.panic_strategy(cnum);
+            if desired_strategy != found_strategy {
+                sess.err(&format!(
+                    "the crate `{}` is compiled with the \
                                panic strategy `{}` which is \
                                incompatible with this crate's \
                                strategy of `{}`",
-                tcx.crate_name(cnum),
-                found_strategy.desc(),
-                desired_strategy.desc()
-            ));
+                    tcx.crate_name(cnum),
+                    found_strategy.desc(),
+                    desired_strategy.desc()
+                ));
+            }
+
+            let found_drop_strategy = tcx.panic_in_drop_strategy(cnum);
+            if tcx.sess.opts.debugging_opts.panic_in_drop != found_drop_strategy {
+                sess.err(&format!(
+                    "the crate `{}` is compiled with the \
+                               panic-in-drop strategy `{}` which is \
+                               incompatible with this crate's \
+                               strategy of `{}`",
+                    tcx.crate_name(cnum),
+                    found_drop_strategy.desc(),
+                    tcx.sess.opts.debugging_opts.panic_in_drop.desc()
+                ));
+            }
         }
     }
 }
--- a/compiler/rustc_session/src/options.rs
+++ b/compiler/rustc_session/src/options.rs
@@ -349,6 +349,7 @@ mod desc {
     pub const parse_threads: &str = parse_number;
     pub const parse_passes: &str = "a space-separated list of passes, or `all`";
     pub const parse_panic_strategy: &str = "either `unwind` or `abort`";
+    pub const parse_opt_panic_strategy: &str = parse_panic_strategy;
     pub const parse_relro_level: &str = "one of: `full`, `partial`, or `off`";
     pub const parse_sanitizers: &str =
         "comma separated list of sanitizers: `address`, `hwaddress`, `leak`, `memory` or `thread`";
@@ -549,10 +550,19 @@ mod parse {
         }
     }
 
-    crate fn parse_panic_strategy(slot: &mut Option<PanicStrategy>, v: Option<&str>) -> bool {
+    crate fn parse_opt_panic_strategy(slot: &mut Option<PanicStrategy>, v: Option<&str>) -> bool {
         match v {
             Some("unwind") => *slot = Some(PanicStrategy::Unwind),
             Some("abort") => *slot = Some(PanicStrategy::Abort),
+            _ => return false,
+        }
+        true
+    }
+
+    crate fn parse_panic_strategy(slot: &mut PanicStrategy, v: Option<&str>) -> bool {
+        match v {
+            Some("unwind") => *slot = PanicStrategy::Unwind,
+            Some("abort") => *slot = PanicStrategy::Abort,
             _ => return false,
         }
         true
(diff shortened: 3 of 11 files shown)


compiler/rustc_typeck/src/collect.rs
```diff
@@ -2683,6 +2683,14 @@ fn codegen_fn_attrs(tcx: TyCtxt<'_>, id: DefId) -> CodegenFnAttrs {
         codegen_fn_attrs.flags |= CodegenFnAttrFlags::TRACK_CALLER;
     }
 
+    // With -Z panic-in-drop=abort, drop_in_place never unwinds.
+    if tcx.sess.opts.debugging_opts.panic_in_drop == PanicStrategy::Abort {
+        let drop_in_place = tcx.require_lang_item(LangItem::DropInPlace, None);
+        if id == drop_in_place {
+            codegen_fn_attrs.flags |= CodegenFnAttrFlags::NEVER_UNWIND;
```
nbdd0121, line 2690: It's undefined behaviour if a `nounwind` function actually unwinds.

nbdd0121: Oh actually this attribute is already covered by `abort_unwinding_calls` transform. So it won't be UB.

jyn514:
> One thing to note about -Z panic-in-drop=abort is that all crates must be built with this option for it to be sound since it makes the compiler assume that dropping Box<dyn Any> will never unwind.

Could we make it a hard error if there's a mismatch in this between crates?

Amanieu:
> Could we make it a hard error if there's a mismatch in this between crates?

Where would be the best place to add this check? Do we have any precedent for validating crate compatibility?

hyd-dev:
I think there is a similar check here:
https://github.com/rust-lang/rust/blob/c5cbf7852a7692c7c51df64c09a59e7838b55202/compiler/rustc_metadata/src/dependency_format.rs#L391-L418

Amanieu:
This shaves 5MB off `librustc_driver.so`, a 3% reduction (197MB -> 192MB).

joshtriplett:
This looks really promising!

I think this would be a good first step. Before we can even *consider* changing the default behavior, we'd need to have an option for the new behavior, and experiment with it extensively.

estebank:
Can somebody else in @rust-lang/compiler take this review on? I am not as confident in codegen as the rest of the compiler.

compiler/rustc_codegen_llvm/src/abi.rs
```diff
@@ -510,8 +516,19 @@ impl<'tcx> FnAbiLlvmExt<'tcx> for FnAbi<'tcx, Ty<'tcx>> {
         }
     }
 
-    fn apply_attrs_callsite(&self, bx: &mut Builder<'a, 'll, 'tcx>, callsite: &'ll Value) {
-        // FIXME(wesleywiser, eddyb): We should apply `nounwind` and `noreturn` as appropriate to this callsite.
+    fn apply_attrs_callsite(
+        &self,
+        bx: &mut Builder<'a, 'll, 'tcx>,
+        callsite: &'ll Value,
+        returns: bool,
```
eddyb, line 523: Could these flags be tracked in the `FnAbi` and applied from there?

compiler/rustc_metadata/src/dependency_format.rs
```diff
@@ -402,19 +402,34 @@ fn verify_ok(tcx: TyCtxt<'_>, list: &[Linkage]) {
             let cnum = CrateNum::new(i + 1);
             let found_strategy = tcx.panic_strategy(cnum);
             let is_compiler_builtins = tcx.is_compiler_builtins(cnum);
-            if is_compiler_builtins || desired_strategy == found_strategy {
+            if is_compiler_builtins {
```
eddyb, line 405: The variable could be inlined, and the `let found_strategy` moved lower down (to match the new code more).

compiler/rustc_typeck/src/collect.rs
```diff
@@ -2683,6 +2683,14 @@ fn codegen_fn_attrs(tcx: TyCtxt<'_>, id: DefId) -> CodegenFnAttrs {
         codegen_fn_attrs.flags |= CodegenFnAttrFlags::TRACK_CALLER;
     }
 
+    // With -Z panic-in-drop=abort, drop_in_place never unwinds.
+    if tcx.sess.opts.debugging_opts.panic_in_drop == PanicStrategy::Abort {
+        let drop_in_place = tcx.require_lang_item(LangItem::DropInPlace, None);
```
eddyb, line 2688: You can check `if Some(id) == tcx.lang_items.drop_in_place()` instead of requiring the lang item.

src/test/ui/asm/inline-syntax.rs
```diff
@@ -25,6 +25,8 @@ macro_rules! global_asm {
 
 #[lang = "sized"]
 trait Sized {}
+#[lang = "drop_in_place"]
+fn drop_in_place<T>(_: *mut T) {}
```
eddyb, line 29: Seems like removing the requirement for the lang item would remove these unnecessary test changes.

eddyb (changes requested):
LGTM so far, but I'd like to review test changes once the `drop_in_place` lang item isn't spuriously required, which seems to account for a lot of them.

nagisa:
Another thing to add on a quick skim: in MIR when building cleanup blocks, we spend a lot of effort to build unwind ladders just in case a cleanup/drop panics. All of that can also be disabled when this flag is enabled, potentially giving you another significant perf boost in build times.

nikomatsakis:
I'm in favor of landing a PR like this to permit experimentation! Data would be good. We'd have to think about the best way to phase the change in.

Depending on what we find in the wild, it occurs to me that this change could be done over an Edition, although it doesn't fit perfectly because interop between editions would be limited. The idea would be like this: code from the new edition would abort when a drop panics. When such code calls a Drop from an older edition, it would wrap that particular drop in a "abort-if-this-panics" block. Interop is limited here because something which is *designed* to panic becomes an abort.

Amanieu:
I've addressed all the review feedback.

I'm having a bit of trouble wrapping my head around the drop tree building code, so I haven't made any changes there yet.

compiler/rustc_mir_build/src/build/expr/into.rs
```diff
@@ -268,6 +268,8 @@ impl<'a, 'tcx> Builder<'a, 'tcx> {
                         // FIXME(varkor): replace this with an uninhabitedness-based check.
                         // This requires getting access to the current module to call
                         // `tcx.is_ty_uninhabited_from`, which is currently tricky to do.
+                        // FIXME(Amanieu): Make sure to also update FnAbi::noreturn when
```
eddyb, line 271: We don't have to wait - `FnAbi` can rely on details that we don't expose to borrowck (the reason MIR needs this weird limitation is because borrowck could tell there is dead code after the call, I believe - basically a back-compat hazard if the type is *privately* uninhabited, but there is no compat limitation on codegen).

eddyb (changes requested):

Amanieu:
Addressed feedback.

eddyb (approved):

compiler/rustc_session/src/options.rs
```diff
@@ -1184,6 +1194,8 @@ options! {
         "pass `-install_name @rpath/...` to the macOS linker (default: no)"),
     panic_abort_tests: bool = (false, parse_bool, [TRACKED],
         "support compiling tests with panic=abort (default: no)"),
+    panic_in_drop: PanicStrategy = (PanicStrategy::Unwind, parse_panic_strategy, [TRACKED],
+        "panic strategy for panics in drops"),
```
klensy, line 1198: Add accepted values list and default (like in some other options), plus maybe unstable book entry?

nagisa:
@bors r=nagisa,eddyb

I think this is reasonable starting point. I suspect there are more places we can reduce the amount of work we do, but there's no reason for all that to be covered in this specific PR.

Wyvern:
Now, I use "-Zpanic-in-drop=abort" and `build-std` to build rust project hit lots of errors, like below:
```
error: the crate `std` is compiled with the panic-in-drop strategy `unwind` which is incompatible with this crate's strategy of `abort`

error: the crate `core` is compiled with the panic-in-drop strategy `unwind` which is incompatible with this crate's strategy of `abort`

error: the crate `rustc_std_workspace_core` is compiled with the panic-in-drop strategy `unwind` which is incompatible with this crate's strategy of `abort`

error: the crate `alloc` is compiled with the panic-in-drop strategy `unwind` which is incompatible with this crate's strategy of `abort`

error: the crate `libc` is compiled with the panic-in-drop strategy `unwind` which is incompatible with this crate's strategy of `abort`
```
Anybody knows how to resolve it?

Amanieu:
> Now, I use "-Zpanic-in-drop=abort" and `build-std` to build rust project hit lots of errors, like below:

It seems to work fine for me:
```
RUSTFLAGS=-Zpanic-in-drop=abort cargo build -Zbuild-std --target x86_64-unknown-linux-gnu
```

Issue #72631, opened 2020-05-26
Tracking Issue for Extend::{extend_one,extend_reserve}


cuviper:
This is a tracking issue for the trait methods `Extend::extend_one` and `extend_reserve`.
The feature gate for the issue is `#![feature(extend_one)]`.

### About tracking issues

Tracking issues are used to record the overall progress of implementation.
They are also uses as hubs connecting to other relevant issues, e.g., bugs or open design questions.
A tracking issue is however *not* meant for large scale discussion, questions, or bug reports about a feature.
Instead, open a dedicated issue for the specific matter and add the relevant feature gate label.

### Steps

- [x] Implement the API
- [x] Adjust documentation ([see instructions on rustc-dev-guide][doc-guide])
- [ ] Stabilization PR ([see instructions on rustc-dev-guide][stabilization-guide])

[stabilization-guide]: https://rustc-dev-guide.rust-lang.org/stabilization_guide.html#stabilization-pr
[doc-guide]: https://rustc-dev-guide.rust-lang.org/stabilization_guide.html#documentation-prs

### Unresolved Questions

- Are these method names appropriate?
  - Note that `Extend` is in the prelude, so the explicit `extend_` prefix avoids ambiguity.
- Is `Extend` the best place for just methods?
  - One goal is to improve `Iterator::unzip`, which only has `Default + Extend`, so there's not really much choice.

### Implementation history

- Initial implementation in #72162.

dtolnay:
Preserving a relevant concern from **@​SimonSapin** in https://github.com/rust-lang/rust/pull/72162#issuecomment-634279732:

> These methods look very out of place in the `Extend` trait, and the `extend_` prefix in their name is apparently unrelated to the meaning of the name of the `extend` method.
>
> From an API design perspective, this seems to be asking for new `Push` and `Reserve` traits. This might deserve to be part of a larger overall design for traits for generic collections. The standard library has mostly avoided doing this so far, if fact `Extend` is the odd one out and arguably should not have been a trait.

est31:
With arrays implementing the IntoIterator trait, I'm not sure `extend_one` is still pulling its weight? It seems the reasoning was that `extend_one(v)` is nicer than `extend(Some(v))`? To me at least, `extend([v])` seems even nicer.

cuviper:
[...]

PR #88761, opened 2021-09-08
Smarter default implementation of Extend::extend_reserve

dtolnay:
Following up on https://github.com/rust-lang/rust/issues/72631#issuecomment-915534130: there is a way for `extend_reserve` to perform a sensible reserve by default, without always needing to be overridden by every `Extend` impl, as long as `extend` checks its argument's size_hint as many already do. For example the following code in `String`'s Extend impl:

https://github.com/rust-lang/rust/blob/c9db3e0fbc84d8409285698486375f080d361ef3/library/alloc/src/string.rs#L1927-L1933

In a future PR we can clean up handwritten implementations of `extend_reserve` which are no longer providing value over this default implementation.

### DIFF
--- a/library/core/src/iter/traits/collect.rs
+++ b/library/core/src/iter/traits/collect.rs
@@ -1,3 +1,5 @@
+use crate::marker::PhantomData;
+
 /// Conversion from an [`Iterator`].
 ///
 /// By implementing `FromIterator` for a type, you define how it will be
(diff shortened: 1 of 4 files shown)


dtolnay (commented):
I have used this in the past to work around lack of stable `extend_reserve` successfully, but it may be possible there are `Extend` implementations in which this would trigger a panic.

cuviper:
`Iterator::size_hint` docs say:
> It is not enforced that an iterator implementation yields the declared number of elements. A buggy iterator may yield less than the lower bound or more than the upper bound of elements.

How comfortable are we with intentionally using a "buggy" iterator?

> but it may be possible there are `Extend` implementations in which this would trigger a panic.

Funny, @scottmcm  just demonstrated that in the users forum:
https://users.rust-lang.org/t/is-string-splitn-missing-a-huge-easy-optimisation/64555/3?u=cuviper

the8472:
I have seen suggestions similar to the one in the urlo thread before. Since specialization isn't stable it might be used as a poor man's `TrustedLen` when the length is fixed.

dtolnay:
I think I am going to withdraw this. While it works well in practice for a wide range of `Extend` impls, there is nothing that makes it incorrect for an `Extend` impl to panic in response to this situation, which would become a bad/surprising `extend_reserve` default behavior.

PR #88763, opened 2021-09-08
Improve `size_hint` bounds for `DecodeUtf16` iterator

### LINKED ISSUE #88762: Incorrect upper bound for `size_hint` of `char::DecodeUtf16`
```rs
use std::char::decode_utf16;

fn main() {
    let x = [0xD834, 0x006d];
    let mut d = decode_utf16(x);
    d.next();
    let (_, u) = d.size_hint();
    let c = d.count();
    println!("upper bound: {:?}, actual length {}", u, c);
    assert!(u.unwrap() >= c);
}
```
```
    Finished dev [unoptimized + debuginfo] target(s) in 1.04s
     Running `target/debug/playground`
thread 'main' panicked at 'assertion failed: u.unwrap() >= c', src/main.rs:10:5
Standard Output
upper bound: Some(0), actual length 1
```
([playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=3a5d9936c6fc80c3e466c4b0124c4c85))

steffahn:
If at most 2 `u16`s make up one item in the iterator, then there has to be a minimum of `len + 1 / 2` elements for an odd-numbered iterator. In other words, the best lower bound would be determined by rounding up.

Edit: I noticed #88762 while creating a PR including just bea5cd1d213d25fc812793d09a5fcab7dcf56353. So now, I’ve added a commit that fixes #88762 as-well.

### DIFF
--- a/library/core/src/char/decode.rs
+++ b/library/core/src/char/decode.rs
@@ -122,7 +122,12 @@ impl<I: Iterator<Item = u16>> Iterator for DecodeUtf16<I> {
         let (low, high) = self.iter.size_hint();
         // we could be entirely valid surrogates (2 elements per
         // char), or entirely non-surrogates (1 element per char)
-        (low.div_ceil(2), high)
+        //
+        // In addition to `self.iter`, there's potentially one
+        // additional number in `self.buf`.
+        // On odd lower bound, at least one element must stay unpaired
+        // (with other elements from `self.iter`).
+        (low.div_ceil(2), try { high?.checked_add(1)? })
     }
 }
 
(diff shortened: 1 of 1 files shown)


steffahn:
There’s the option of also inspecting `buf` to create a more accurate size_hint.

steffahn:
Also missing tests. I’m going to create an alternative version that _does_ inspect `buf`, for comparison.

@rustbot S-waiting-on-author -S-waiting-on-review

steffahn:
> I’m going to create an alternative version that _does_ inspect `buf`, for comparison.

986894702408d60b761a778044ea8995e3c8310b might work. I won’t continue this before tomorrow.

crlf0710:
@steffahn Ping from triage, any updates on this?

PR #88765, opened 2021-09-08
Update LLVM submodule

mk12:
This merges upstream `release/13.x` changes to our fork. In particular,
this includes the bugfix https://reviews.llvm.org/D108608 (see also
https://bugs.llvm.org/show_bug.cgi?id=51637).

The motivation for this is fixing Rust coverage in Fuchsia: https://fxbug.dev/77427.


nikic:
I've updated the submodule to a slightly newer merge, as some fixes relevant for us have landed upstream in the meantime, in particular the one for #88769.

@bors r=cuviper rollup=never

Mark-Simulacrum:

This is necessary for us to keep up to date with upstream LLVM, and the changes here do not seem to merit deep investigation given their magnitude and relative scattering across positive/negative changes.

PR #88771, opened 2021-09-09
Use FxHashSet instead of Vec for well formed tys

jackh726:
Trying to recover perf from #88312

r? @ghost

### DIFF
--- a/compiler/rustc_trait_selection/src/infer.rs
+++ b/compiler/rustc_trait_selection/src/infer.rs
@@ -2,6 +2,7 @@ use crate::traits::query::evaluate_obligation::InferCtxtExt as _;
 use crate::traits::query::outlives_bounds::InferCtxtExt as _;
 use crate::traits::{self, TraitEngine, TraitEngineExt};
 
+use rustc_data_structures::stable_set::FxHashSet;
 use rustc_hir as hir;
 use rustc_hir::def_id::DefId;
 use rustc_hir::lang_items::LangItem;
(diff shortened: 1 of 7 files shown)


jackh726:
Doesn't win back *all* the perf lost from #88312, but is better. I'm not sure that there's a way to win more perf without changing the approach completely. So, I think this is probably worth merging and accepting the difference.

cc @rylev regarding perf
cc @nikomatsakis who might want to steal this review

r? rust-lang/compiler

compiler/rustc_trait_selection/src/infer.rs
```diff
@@ -210,13 +211,13 @@ impl<'tcx> OutlivesEnvironmentExt<'tcx> for OutlivesEnvironment<'tcx> {
...
         infcx: &InferCtxt<'a, 'tcx>,
-        fn_sig_tys: &[Ty<'tcx>],
+        fn_sig_tys: FxHashSet<Ty<'tcx>>,
         body_id: hir::HirId,
         span: Span,
     ) {
         debug!("add_implied_bounds()");
 
-        for &ty in fn_sig_tys {
+        for ty in fn_sig_tys {
```
eddyb, line 220: This is iterating on a `FxHashSet<Ty<'tcx>>`, which will depend on ASLR (random memory allocation order) and the results could be observable (e.g. potentially in diagnostics order).

I wonder if we should ban iterating `FxHash{Set,Map}` and force the use of `FxIndex{Set,Map}` instead (since presumably the *insertion* order is deterministic).

eddyb: See also https://github.com/rust-lang/rust/issues/63713#issuecomment-685109097 - we should probably use internal lints to force this harder.

lcnr: > we should ban iterating `FxHash{Set,Map}`

YES! I think recommending `FxIndex{Set,Map}` is probably the right choice, though instead of banning it we could use an internal lint whenever using a `FxIndexSet` in a potentially nondet way. Now if the order doesn't matter, e.g. checking for existence without any sideeffects, this lint could be explicitly allowed.

Not sure what's the best way to write such a lint though :sweat_smile:

PR #88772, opened 2021-09-09
Fixed confusing wording on Result::map_or_else.

### LINKED ISSUE #88195: Documentation on `Result::map_or_else` uses very unfortunate wording
The type signature of `Result::map_or_else`:

```rust
pub fn map_or_else<U, D, F>(self, default: D, f: F) -> U where
    F: FnOnce(T) -> U,
    D: FnOnce(E) -> U, 
```
Note how `default` is the function that gets applied to the `Err` variant. Then, the docstring (emphasis mine):

> Maps a `Result<T, E>` to `U` by applying a fallback function to a contained `Err` value, or **a default function** to a contained `Ok` value.

I think this is a very unfortunate choice of words and should be changed.

orlp:
Fixes https://github.com/rust-lang/rust/issues/88195.

### DIFF
--- a/library/core/src/result.rs
+++ b/library/core/src/result.rs
@@ -795,9 +795,8 @@ impl<T, E> Result<T, E> {
         }
     }
 
-    /// Maps a `Result<T, E>` to `U` by applying a provided default fallback
-    /// function to a contained [`Err`] value, or a provided function to a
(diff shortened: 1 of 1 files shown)


yaahc:
r? yaahc

PR #85515, opened 2021-05-20
Fix ast pretty printing for anonymous types

jedel1043:
Fixes #85480.

PR #84571, opened 2021-04-26
Parse unnamed fields of struct and union type

jedel1043:
Added the `unnamed_fields` feature gate.

This is a prototype of [RFC 2102](https://github.com/rust-lang/rust/issues/49804), so any suggestions are greatly appreciated.

r? @petrochenkov

PR #88775, opened 2021-09-09
Revert anon union parsing

### LINKED ISSUE #88583: Type called union wreaks havoc since 1.54
When C-compatible untagged unions were introduced by RFC 1444, it was careful not to break existing valid uses of `union` identifier, such as https://doc.rust-lang.org/std/collections/struct.BTreeSet.html#method.union.

However, the unions from RFC 2102 (https://github.com/rust-lang/rust/issues/49804) appear to be more disruptive, at least as currently implemented. The following stable code works with every stable rustc from 1.0.0 through 1.53.0, but can no longer be parsed by rustc 1.54.0 or current nightly.

```rust
struct union;

impl union {
    pub fn new() -> Self {
        unimplemented!()
    }
}

fn main() {}
```

```console
error: functions are not allowed in union definitions
 --> src/main.rs:4:5
  |
4 | /     pub fn new() -> Self {
5 | |         unimplemented!()
6 | |     }
  | |_____^
  |
  = help: unlike in C++, Java, and C#, functions are declared in `impl` blocks
  = help: see https://doc.rust-lang.org/book/ch05-03-method-syntax.html for more information

error: expected `(`, found `main`
 --> src/main.rs:9:4
  |
9 | fn main() {}
  |    ^^^^ expected `(`
```

@jedel1043 @joshtriplett 

Labeling <kbd>T-compiler</kbd> for what to do about the regression and <kbd>T-lang</kbd> because it's not clear to me what the grammar of anonymous union is intended to be from the parser's perspective.

pnkfelix:
Revert PR #84571 and #85515, which implemented anonymous union parsing in a manner that broke the context-sensitivity for the `union` keyword and thus broke stable Rust code.

Fix #88583.


davidtwco (approved):
LGTM, only checked the manual reverts/commits looked reasonable, r=me when you're ready and CI is passing.

estebank:
I opened https://github.com/rust-lang/rust/pull/88815 as an alternative to reverting. If we still revert, it can be used once we reintroduce the feature.

pnkfelix:
beta nominating: We want this fix for #88583 to end up in the 1.56 release, if possible, in order to coincide with the 2021 edition.

estebank:
Closing #88815. It's changes need to be included when relanding this on master.

pnkfelix:
beta backport approved

apiraino:
Relevant discussion [on Zulip](https://rust-lang.zulipchat.com/#narrow/stream/238009-t-compiler.2Fmeetings/topic/.5Bweekly.5D.202021-09-16.20.2354818/near/253589761)

Issue #88545, opened 2021-08-31
Generated docs on long pages render text on top of itself


mqudsi:
_I hope this is the right repo to report this, I wasn't sure if it should go in the main rust repo instead._

The generated docs for pages with a crazy number of items seem to give up on positioning and then just dump all subsequent entries on the same line past a certain point. It looks like it might be related to hardware acceleration, but I'm not sure.

For example, viewing https://doc.rust-lang.org/beta/core/arch/arm/index.html in Chrome 92 on Windows and hitting ctrl+pgdwn to go to the very bottom shows this:

![image](https://user-images.githubusercontent.com/606923/131541457-818982b5-447f-4c12-a0e2-dbbc99a26bb1.png)

jyn514:
@mqudsi no worries - rustdoc is used for both docs.rs and the main doc.rust-lang.org website, but docs.rs doesn't have control over the main site.

jyn514:
I can't reproduce this on mobile Chrome though - maybe it's related to the screen size?
![Screenshot_20210831-161236](https://user-images.githubusercontent.com/23638587/131576640-18919152-b28a-44a2-b7e3-6b8cb2b3b03d.png)

Seadragon91:
Can confirm that problem on my PC in Edge.
Also on my smartphone in Edge and Chrome, I need to change the view of the website to "Desktop website" then I see it there too.

camelid:
I can't reproduce this in Firefox on macOS, but I *can* in Chrome on macOS. So I'm guessing it's a bug specific to Blink because I think Chrome and Edge both use Blink, whereas Firefox does not.

mqudsi:
> It seems to be a bug with Blink's grid implementation. If I turn off the display: grid style it works (but obviously changes the layout).

I’m not saying it’s *not* grid, but just pointing out that changing the display completely changes the layout and so may avoid the problem even if it’s something else. In theory, anyway.

ChrisDenton:
Fair point. A minimal test case appears to show the same issue but I'll admit that does not necessarily determine the root cause:

```html
<!DOCTYPE html>
<html lang="en">
<head>
    <meta http-equiv="Content-Type" content="text/html; charset=UTF-8">
    <title>Test display grid with lots of children</title>
</head>

<body style="display: grid;">
<script><!--//<![CDATA[
for (let i = 0; i < 600; i++) {
    document.write(`<div>Testing</div>`);
    document.write(`<div>Overtype</div>`);
}
[...]

PR #85651, opened 2021-05-24
rustdoc: staggered layout for module contents on mobile

dns2utf8:
This PR adds the container `<item-table>` with its two children `<item-left>` and `<item-right>`.
It uses grid-layout on desktop and flexbox on mobile to make better use of the available space.

Additionally it allows to share parts of the CSS with the search function.

* Demo: https://data.estada.ch/rustdoc-nightly_126561cb3_2021-05-25/generic_array/index.html
* Related: #85540

## Desktop
grafik

## Mobile
grafik

r? @GuillaumeGomez @jsha

PR #88776, opened 2021-09-09
Workaround blink/chromium grid layout limitation of 1000 rows

dns2utf8:
I made this in case we don't come up with a better solution in time.

See https://github.com/rust-lang/rust/issues/88545 for more details.

A rendered version of the standard library is hosted here:
https://data.estada.ch/rustdoc-nightly_497ee321af_2021-09-09/core/arch/arm/index.html

r? @GuillaumeGomez @jsha

### DIFF
--- a/src/librustdoc/html/render/print_item.rs
+++ b/src/librustdoc/html/render/print_item.rs
@@ -256,6 +256,10 @@ fn item_module(w: &mut Buffer, cx: &Context<'_>, item: &clean::Item, items: &[cl
 
     debug!("{:?}", indices);
     let mut curty = None;
+    // See: https://github.com/rust-lang/rust/issues/88545
+    let item_table_block_size = 900usize;
+    let mut item_table_nth_element = 0usize;
+
     for &idx in &indices {
         let myitem = &items[idx];
         if myitem.is_stripped() {
@@ -280,6 +284,7 @@ fn item_module(w: &mut Buffer, cx: &Context<'_>, item: &clean::Item, items: &[cl
                 id = cx.derive_id(short.to_owned()),
                 name = name
             );
+            item_table_nth_element = 0;
         }
 
         match *myitem.kind {
@@ -385,6 +390,13 @@ fn item_module(w: &mut Buffer, cx: &Context<'_>, item: &clean::Item, items: &[cl
                         .join(" "),
                 );
             }
+        }
+
+        item_table_nth_element += 1;
+        if item_table_nth_element > item_table_block_size {
+            w.write_str(ITEM_TABLE_CLOSE);
+            w.write_str(ITEM_TABLE_OPEN);
+            item_table_nth_element = 0;
         }
     }
 


GuillaumeGomez (approved):

GuillaumeGomez:
Waiting for then CI then I r+ it.

mqudsi:
*cough* tables *cough*. 

But seriously, if you use a css reset to remove all browser styling (are we already doing that?) you can use the table layout in the renderer to create code like our current grid (because we are barely using any of the actual grid features). A little bit of css to the cell classes makes it responsive.

dns2utf8:
You are welcome! 
Btw. I just realised, if there is a 1.55.1 release: This PR should probably be included in that

dns2utf8:
I tested the workaround with IE Edge 92 and Chrome 93 and it works okay.
Safari on Mac and Firefox worked before already and behave the same as the blink based browsers now ☺️

GuillaumeGomez:
Good point, let's nominate it for beta backport.

Do you accept this to be backported to beta @rust-lang/rustdoc?

Nemo157:
Was the issue a beta regression?

GuillaumeGomez:
I think it reached the 1.56 beta version, so yes, it's a regression.

Nemo157:
I was more wondering whether it was not already in stable, looking back at the history I assume https://github.com/rust-lang/rust/pull/85651 introduced the issue? That is on 1.55, but maybe it's worth a beta backport to only have a single version with the problem.

GuillaumeGomez:
Yes it's the one.

dns2utf8:
Unfortunately it has reached stable too: https://doc.rust-lang.org/beta/core/arch/arm/index.html

GuillaumeGomez:
Ah damn... Oh well, at least it'll only be one version with this issue...

dns2utf8:
If there is a 1.55.1 release we could backport the workaround there too

dns2utf8:
Should I open a PR with this commit on the beta branch too?

GuillaumeGomez:
Yes please. :)

dns2utf8:
Done :) Could you move the beta-nominated label to the new PR?

jyn514:
This is not how beta nominations work. Please use the existing process so the release team can keep track of which PRs have been backported and which haven't.

https://forge.rust-lang.org/release/backporting.html?highlight=Beta%20n#beta-backporting-in-rust-langrust

dns2utf8:
Very sorry. I renamed the other PR, but I am not sure If I should be reopened ir or if someone else will cherrypick one of the PRs or me doing nothing?

jyn514:
@dns2utf8 it's ok, that's why there's docs :) if you want to make a backport PR that's fine, but please include all the other beta accepted PRs at the same time, and ask a member of T-release to review, not Guillaume.

camelid:
It's probably best to just wait for T-release to backport this PR with the other beta backports. Someone from T-release goes through and backports beta-accepted PRs periodically.

PR #88795, opened 2021-09-09
Print a note if a character literal contains a variation selector

### LINKED ISSUE #88684: Improve diagnostics for character literals containing multiple codepoints
I tried to compile the following code ([playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=17b195911187376f740b8e4e1c6f6fba)):

```rust
const SPADE: char = '♠️';
```

For the avoidance of doubt, `♠️` is [U+2660](https://util.unicode.org/UnicodeJsps/character.jsp?a=2660).  I expected the code to compile—but instead, the literal is rejected:

```
error: character literal may only contain one codepoint
 --> src/lib.rs:1:21
  |
1 | const SPADE: char = '♠️';
  |                     ^^^
  |
help: if you meant to write a `str` literal, use double quotes
  |
1 | const SPADE: char = "♠️";
  |                     ^^^
```

Notably, it is accepted if unicode-escaped:

```rust
const SPADE: char = '\u{2660}';
```

### Meta

`rustc --version --verbose`:
```
rustc 1.57.0-nightly (e30b68353 2021-09-05)
binary: rustc
commit-hash: e30b68353fe22b00f40d021e7914eeb78473b3c1
commit-date: 2021-09-05
host: x86_64-apple-darwin
release: 1.57.0-nightly
LLVM version: 13.0.0
```

FabianWolff:
Fixes #88684.

### DIFF
--- a/compiler/rustc_parse/src/lexer/unescape_error_reporting.rs
+++ b/compiler/rustc_parse/src/lexer/unescape_error_reporting.rs
@@ -3,7 +3,7 @@
 use std::iter::once;
 use std::ops::Range;
 
-use rustc_errors::{Applicability, Handler};
+use rustc_errors::{pluralize, Applicability, Handler};
 use rustc_lexer::unescape::{EscapeError, Mode};
 use rustc_span::{BytePos, Span};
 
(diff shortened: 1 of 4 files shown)


compiler/rustc_parse/src/lexer/unescape_error_reporting.rs
```diff
@@ -55,11 +55,23 @@ pub(crate) fn emit_unescape_error(
...
             };
 
-            handler
-                .struct_span_err(
+            let mut handler = handler.struct_span_err(
+                span_with_quotes,
+                "character literal may only contain one codepoint",
+            );
+            if lit.chars().count() == 2
+                && ('\u{FE00}'..='\u{FE0F}').contains(&lit.chars().last().unwrap())
```
eggyal, line 63: Don't forget [Variation Selectors Supplement](https://en.wikipedia.org/wiki/Variation_Selectors_Supplement) and the [Mongolian Variation Selectors](https://en.wikipedia.org/wiki/Mongolian_(Unicode_block)#Variations_and_vowel_separation)... indeed, perhaps [`unicode_normalization::char::is_combining_mark`](https://docs.rs/unicode-normalization/0.1.19/unicode_normalization/char/fn.is_combining_mark.html) could offer a more general approach?

FabianWolff: I have expanded my implementation now to print a note for all combining marks, as you suggested. I've also added a suggestion to use the normalized form if the character can be represented as one codepoint (as in [this](https://www.unicode.org/reports/tr15/images/UAX15-NormFig4.jpg) example).

eggyal: Very nice!

wesleywiser (approved):
This is fantastic, thank you!

Issue #88796, opened 2021-09-09
Windows CI is borked


ehuss:
Windows CI has stopped working due to an image update on GitHub's side.  The new image is [win19/20210907.4](https://github.com/actions/virtual-environments/releases/tag/win19%2F20210907.4) ([included software](https://github.com/actions/virtual-environments/blob/win19/20210907.4/images/win/Windows2019-Readme.md)). The previous image was [win19/20210903.7](https://github.com/actions/virtual-environments/releases/tag/win19%2F20210903.7) ([included software](https://github.com/actions/virtual-environments/blob/win19/20210903.7/images/win/Windows2019-Readme.md)). The image diff is [here](https://github.com/actions/virtual-environments/commit/b5e5d9f68d1dc94da0abe5857f89efe0783c5805).

There are two errors showing up described below.

I think the most likely culprit is the addition of Windows 10 SDK 20348.  Previously, the most recent SDK installed was 19041.

There are some interesting release notes for 20348 [here](https://developer.microsoft.com/en-us/windows/downloads/windows-10-sdk/#serverdev):

> Clang/LLVM for Windows v11 targeting ARM64 is not compatible with the latest winnt.h
> 
> As a workaround, use the previous version of the Windows 10 SDK (build 19041), or clang/LLVM for Windows v10 when targeting ARM64 platforms

I don't know what "LLVM for Windows v11" means in this context, but it seems fishy.

## dist-aarch64-msvc fails with undeclared identifier '__umulh'

While building `profiler_builtins` there is an error:

```
cargo:warning=In file included from ../../src/llvm-project/compiler-rt\lib\profile\InstrProfilingRuntime.cpp:11:
cargo:warning=In file included from ../../src/llvm-project/compiler-rt\lib\profile/InstrProfiling.h:12:
cargo:warning=In file included from ../../src/llvm-project/compiler-rt\lib\profile/InstrProfilingPort.h:65:
cargo:warning=In file included from C:\Program Files (x86)\Windows Kits\10\include\10.0.20348.0\um\windows.h:171:
cargo:warning=In file included from C:\Program Files (x86)\Windows Kits\10\include\10.0.20348.0\shared\windef.h:24:
cargo:warning=In file included from C:\Program Files (x86)\Windows Kits\10\include\10.0.20348.0\shared\minwindef.h:182:
[...]

PR #88797, opened 2021-09-09
Work around CI issue with windows sdk 10.0.20348.0.

m-ou-se:
See https://github.com/rust-lang/rust/issues/88796#issuecomment-916844522


m-ou-se:
(This was tested on `dist-aarch64-msvc` here: https://github.com/rust-lang/rust/pull/88797/checks?check_run_id=3566033204. Everything passed, except the last part where it would otherwise upload things to aws.)

Issue #88576, opened 2021-09-01
New panic with env_logger and with #[windows_subsystem = "windows"]


parasyte:
This was initially reported in https://github.com/env-logger-rs/env_logger/issues/214, but I'm adding it here for additional eyes and because this issue doesn't occur on stable Rust. I could be wrong but this looks like this is caused by #87329 ... cc @sunfishcode ?

----

I spent an hour trying to reduce a minimal test case for this error I started experiencing with my app. So far I have produced the following:

```toml
[package]
name = "windows_explodey"
version = "0.1.0"
edition = "2018"
resolver = "2"

[dependencies]
env_logger = "0.9"

[profile.release]
panic = "abort"
```

```rust
#![windows_subsystem = "windows"]

fn main() {
    env_logger::init();

    println!("Hello, world!");
}
```

When running this on any nightly compiler since `nightly-2021-08-21`, I get an error like this:

```
$ cargo +nightly-2021-08-21 run
    Finished dev [unoptimized + debuginfo] target(s) in 0.03s
     Running `target\debug\windows_explodey.exe`
error: process didn't exit successfully: `target\debug\windows_explodey.exe` (exit code: 101)
```

Adjusting the profile to `panic = "abort"` causes an even more ominous error:

```
$ cargo +nightly-2021-08-21 run --release
    Finished release [optimized] target(s) in 0.03s
     Running `target\release\windows_explodey.exe`
error: process didn't exit successfully: `target\release\windows_explodey.exe` (exit code: 0xc0000409, STATUS_STACK_BUFFER_OVERRUN)
```

Here's a stack trace from windbg:

```
[0x0]   windows_explodey!panic_abort::__rust_start_panic::abort + 0x5   
[0x1]   windows_explodey!panic_abort::__rust_start_panic + 0x6   
[0x2]   windows_explodey!std::panicking::rust_panic + 0x1b   
[0x3]   windows_explodey!std::panicking::rust_panic_with_hook + 0x297   
[0x4]   windows_explodey!std::panicking::begin_panic_handler::closure$0 + 0x6f   
[0x5]   windows_explodey!std::sys_common::backtrace::__rust_end_short_backtrace<std::panicking::begin_panic_handler::closure$0,never$> + 0x17   
[0x6]   windows_explodey!std::panicking::begin_panic_handler + 0x49   
[0x7]   windows_explodey!core::panicking::panic_fmt + 0x30   
[0x8]   windows_explodey!core::panicking::panic + 0x4c   
[...]

PR #88798, opened 2021-09-09
Fix assertion failures in `OwnedHandle` with `windows_subsystem`.

sunfishcode:
As discussed in #88576, raw handle values in Windows can be null, such
as in `windows_subsystem` mode, or when consoles are detached from a
process. So, don't use `NonNull` to hold them, don't assert that they're
not null, and remove `OwnedHandle`'s `repr(transparent)`. Introduce a
new `HandleOrNull` type, similar to `HandleOrInvalid`, to cover the FFI
use case.

r? @joshtriplett

### DIFF
--- a/library/std/src/os/windows/io/handle.rs
+++ b/library/std/src/os/windows/io/handle.rs
@@ -4,7 +4,6 @@
 
 use super::raw::{AsRawHandle, FromRawHandle, IntoRawHandle, RawHandle};
 use crate::convert::TryFrom;
-use crate::ffi::c_void;
 use crate::fmt;
 use crate::fs;
 use crate::marker::PhantomData;
@@ -133,7 +132,7 @@ impl TryFrom<HandleOrNull> for OwnedHandle {
     #[inline]
     fn try_from(handle_or_null: HandleOrNull) -> Result<Self, ()> {
         let owned_handle = handle_or_null.0;
-        if owned_handle.handle.as_ptr().is_null() { Err(()) } else { Ok(owned_handle) }
+        if owned_handle.handle.is_null() { Err(()) } else { Ok(owned_handle) }
     }
 }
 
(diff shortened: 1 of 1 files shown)


joshtriplett:
Can you please clarify in the comments whether `HandleOrNull` can be invalid, and whether `HandleOrInvalid` can be NULL?

I don't know which way around makes sense, but I would *expect* that there's some type that doesn't allow either invalid or NULL, and whose `try_into()` implementation rejects both; I'd expect that to be common in FFI. I don't think we need the entire 2x2 matrix to handle disallowing NULL, invalid, or both; I think it'd make sense to have a type for the "both" case and a type for whichever one can appear without the other.

sunfishcode:
> Can you please clarify in the comments whether `HandleOrNull` can be invalid, and whether `HandleOrInvalid` can be NULL?
>
> I don't know which way around makes sense, but I would _expect_ that there's some type that doesn't allow either invalid or NULL, and whose `try_into()` implementation rejects both; I'd expect that to be common in FFI. I don't think we need the entire 2x2 matrix to handle disallowing NULL, invalid, or both; I think it'd make sense to have a type for the "both" case and a type for whichever one can appear without the other.

I've now added a patch documenting this.

My current understanding of `windows_subsystem = "windows"` mode and detached consoles is that we basically have to treat NULL as a valid handle value except in APIs where it's the documented error value. And my understanding of `GetCurrentProcess()` is that we basically have to treat `INVALID_HANDLE_VALUE` as a valid handle value except in APIs where it's the documented error value. Under that assumption, neither of `HandleOrNull` or `HandleOrInvalid` make sense as a "both" type.

sunfishcode:
@joshtriplett I've now added comments clarifying the valid values for `HandleIfNull` and `HandleIfInvalid`.

sunfishcode:
I've now addressed the review comments.

r? @joshtriplett

sunfishcode:
I've addressed the comments, so this is ready for review.

r? @joshtriplett

Subsentient:
@joshtriplett

joshtriplett:
Will review this today; sorry for the delay. I feel like this might not be exactly the right design for handling NULL, but that's a secondary concern. It's a nightly API, and we can always evolve it further, but we need to fix the issue first.

joshtriplett:
Nominating for both stable and beta backport. T-libs thinks this is *definitely* worth a beta backport, but it's only worth a stable backport if we're already going to do a stable release for some other reason.

Subsentient:
> Nominating for both stable and beta backport. T-libs thinks this is _definitely_ worth a beta backport, but it's only worth a stable backport if we're already going to do a stable release for some other reason.

Personally I'd backport to stable as well -- I feel like having the GUI target broken on stable Rust does not look good or inspire confidence in Rust for anyone who hits this bug. It's core functionality for Windows targets.

joshtriplett:
Reviewing this carefully, I'm *not* sure this is the right API, but I *am* sure that we should prioritize a fix for the observed assertion failures that occur with *stable* APIs backed by this over further refinements to the *nightly-only* API.

@bors r+ rollup=never p=10

joshtriplett:
> Personally I'd backport to stable as well -- I feel like having the GUI target broken on stable Rust does not look good or inspire confidence in Rust for anyone who hits this bug. It's core functionality for Windows targets.

I personally agree.

Subsentient:
@joshtriplett Restart the CI? ^^

PR #88801, opened 2021-09-09
Fix for raw-dylib when linking with MinGW BFD

mati865:
Basically a proof of concept right now, needs more work/thoughts.

### DIFF
--- a/compiler/rustc_codegen_llvm/src/back/archive.rs
+++ b/compiler/rustc_codegen_llvm/src/back/archive.rs
@@ -8,7 +8,7 @@ use std::ptr;
 use std::str;
 
 use crate::llvm::archive_ro::{ArchiveRO, Child};
-use crate::llvm::{self, ArchiveKind, LLVMMachineType, LLVMRustCOFFShortExport};
+use crate::llvm::{self, ArchiveKind, LLVMMachineType};
 use rustc_codegen_ssa::back::archive::ArchiveBuilder;
 use rustc_data_structures::temp_dir::MaybeTempDir;
 use rustc_middle::middle::cstore::{DllCallingConvention, DllImport};
@@ -54,6 +54,7 @@ fn archive_config<'a>(sess: &'a Session, output: &Path, input: Option<&Path>) ->
     ArchiveConfig { sess, dst: output.to_path_buf(), src: input.map(|p| p.to_path_buf()) }
 }
 
+#[allow(dead_code)]
 /// Map machine type strings to values of LLVM's MachineTypes enum.
 fn llvm_machine_type(cpu: &str) -> LLVMMachineType {
     match cpu {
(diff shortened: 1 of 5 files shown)


ricobbe:
This approach is admirable in its simplicity, but unfortunately, I don't think it supports all of the cases specified in the RFC.  To the best of my knowledge, there is no way to specify stdcall, fastcall, or vectorcall functions in a .DEF file, and stdcall in particular is required in order to link against the Windows API on 32-bit Intel.

mati865:
Right, I tend to forget i686 targets still do exist.

I think there is no problem with most of calling conventions as long as we properly save them into `.def`.
Consider this snippet:
```C
// $ cat calling_conventions.c
__cdecl __declspec(dllexport) int cdeclSum(int a, int b)
{
  return a + b;
}

__stdcall __declspec(dllexport) int stdcallSum(int a, int b)
{
  return a + b;
}

__fastcall __declspec(dllexport) int fastcallSum(int a, int b)
{
  return a + b;
}
```
Compile, link and check the symbols:
```bash
$ gcc -shared calling_conventions.c -o calling_conventions.dll

$ nm calling_conventions.dll | rg Sum
64ec14fc T @fastcallSum@8
64ec14e0 T _cdeclSum
64ec14ed T _stdcallSum@8
```
So far so good, now `.def` file:
```bash
$ gendef calling_conventions.dll
 * [calling_conventions.dll] Found PE image

$ cat calling_conventions.def
;
; Definition file of calling_conventions.dll
; Automatic generated by gendef
; written by Kai Tietz 2008
;
LIBRARY "calling_conventions.dll"
EXPORTS
@fastcallSum@8
cdeclSum
stdcallSum@8@8
```
Now the import library:
```bash
$ dlltool -d calling_conventions.def -l calling_conventions.dll.a

mateusz@DESKTOP-MATI MINGW32 /d/tmp
$ nm calling_conventions.dll.a | rg Sum
00000000 I __imp__stdcallSum@8@8
00000000 T _stdcallSum@8@8
00000000 I __imp__cdeclSum
00000000 T _cdeclSum
00000000 T @fastcallSum@8
00000000 I __imp_@fastcallSum@8
```
I'm not an Windows expert but seems fine for me.

---

Now the hard part for GNU toolchain, `__vectorcall`. Let's add this snippet to the previous file:
```C
__vectorcall __declspec(dllexport) int vectorcallSum(int a, int b)
{
  return a + b;
}
```
Let's build, link and see the symbols:
```bash
$ gcc -shared calling_conventions.c -o calling_conventions.dll
calling_conventions.c:16:1: warning: data definition has no type or storage class
   16 | __vectorcall __declspec(dllexport) int vectorcallSum(int a, int b)
      | ^~~~~~~~~~~~
calling_conventions.c:16:1: warning: type defaults to 'int' in declaration of '__vectorcall' [-Wimplicit-int]
calling_conventions.c:16:36: error: expected ',' or ';' before 'int'
   16 | __vectorcall __declspec(dllexport) int vectorcallSum(int a, int b)
      |                                    ^~~

# GCC didn't work, trying Clang
$ clang -shared calling_conventions.c -o calling_conventions.dll
D:/msys64/mingw32/bin/ld: cannot export vectorcallSum@@8: symbol not found
clang: error: linker command failed with exit code 1 (use -v to see invocation)

# Which failed because of BFD..., now LLD
$ clang -fuse-ld=lld -shared calling_conventions.c -o calling_conventions.dll

$ nm calling_conventions.dll | rg Sum
10001520 T @fastcallSum@8
100014e0 T _cdeclSum
10001500 T _stdcallSum@8
10001540 T vectorcallSum@@8
```
So Clang + LLD worked, let's see about GNU dlltool:
```bash
$ gendef calling_conventions.dll
 * [calling_conventions.dll] Found PE image

$ cat calling_conventions.def
;
; Definition file of calling_conventions.dll
; Automatic generated by gendef
; written by Kai Tietz 2008
;
LIBRARY "calling_conventions.dll"
EXPORTS
@fastcallSum@8
cdeclSum
stdcallSum@8@8
vectorcallSum@@8

$ dlltool -d calling_conventions.def -l calling_conventions.dll.a

$ nm calling_conventions.dll.a | rg Sum
00000000 I __imp__vectorcallSum@@8
00000000 T _vectorcallSum@@8
00000000 I __imp__stdcallSum@8@8
00000000 T _stdcallSum@8@8
00000000 I __imp__cdeclSum
00000000 T _cdeclSum
00000000 T @fastcallSum@8
00000000 I __imp_@fastcallSum@8
```
It worked I guess.
Does this sound right?

mati865:
So this has almost worked, the main raw-dylib test passes on i686, alt calling convections not so much:
```
--- output.txt  2021-08-01 14:42:35.185553200 +0200
+++ /d/Projekty/rust/build/i686-pc-windows-gnu/test/run-make/raw-dylib-alt-calling-convention/raw-dylib-alt-calling-convention/output.txt       2021-09-11 01:51:42.126128500 +0200
@@ -11,8 +11,8 @@
 fastcall_fn_2(16, 3.5)
 fastcall_fn_3(3.5)
 fastcall_fn_4(1, 2, 3.0)
-fastcall_fn_5(S { x: 1, y: 2 }, 16)
+fastcall_fn_5(S { x: 1, y: 2 }, 1074266112)
 fastcall_fn_6(S { x: 10, y: 12 })
-fastcall_fn_7(S2 { x: 15, y: 16 }, 3)
+fastcall_fn_7(S2 { x: 15, y: 16 }, 10)
 fastcall_fn_8(S3 { x: [1, 2, 3, 4, 5] }, S3 { x: [6, 7, 8, 9, 10] })
 fastcall_fn_9(1, 3.0)
```
So far I had no success getting debuggers on Windows to help me with that but I wouldn't be surprised if it's alignment issue.

mati865:
A bit cleaned up the implementation and added 2 MinGW builders to the CI. That way I can test Linux -> Windows and {i686,x86_64} Windows -> {i686,x86_64} Windows cross builds.

Can I have `@bors try`?

mati865:
It works in all cases except self-contained mode because it's not added to the PATH (even with `-C link-self-contained=true`).
Bits that handle it for linking are located in `compiler\rustc_codegen_ssa\src\back\link.rs` so I'll look if I can extract it somewhere.

compiler/rustc_codegen_llvm/src/back/archive.rs
```diff
@@ -351,3 +403,26 @@ impl<'a> LlvmArchiveBuilder<'a> {
 fn string_to_io_error(s: String) -> io::Error {
     io::Error::new(io::ErrorKind::Other, format!("bad archive: {}", s))
 }
+
+fn find_dlltool(sess: &Session) -> OsString {
+    // When cross-compiling first try binary prefixed with target triple
+    if sess.host.llvm_target != sess.target.llvm_target {
```
nagisa, line 409: `llvm_target` is probably not the best way to check for this. There are many llvm target triples that are actually equivalent in the end. (e.g. `i686-pc-windows-gnu` and `i686-unknown-windows-gnu` and e.g. json targets could specify either).

compiler/rustc_codegen_llvm/src/back/archive.rs
```diff
@@ -351,3 +403,26 @@ impl<'a> LlvmArchiveBuilder<'a> {
...
+            "i686-w64-mingw32-dlltool"
+        } else {
+            "x86_64-w64-mingw32-dlltool"
+        };
+        let prefixed_dlltool = if cfg!(windows) {
+            [prefixed_dlltool, "exe"].concat()
+        } else {
+            prefixed_dlltool.to_string()
+        };
+        for dir in env::split_paths(&env::var_os("PATH").unwrap_or_default()) {
```
nagisa, line 420: This should also either verify that the binary is executable at all in some way or provide a way to override this detection. Right now if the system is borked enough to have a non-working `i686-w64-mingw32-dlltool` but a working `dlltool`, then there is fairly little the user user can do to work around the problem.

src/test/run-make/raw-dylib-alt-calling-convention/Makefile
```diff
@@ -1,12 +1,12 @@
 # Test the behavior of #[link(.., kind = "raw-dylib")] with alternative calling conventions.
 
-# only-i686-pc-windows-msvc
+# only-i686-pc-windows-gnu
```
nagisa, line 3: These tests are supposed to eventually test both architectures, right?

mati865: Yeah, I just didn't want to make spend time on proper fix yet.

compiler/rustc_codegen_llvm/src/back/archive.rs
mati865: TBH the more I think about this approach the more hacky it seems.

mati865:
With https://github.com/rust-lang/rust/pull/89025 it's clear how this approach is lacking and won't help much to stabilise `raw-dylib`.
I'll reach out `mstorsjo` from LLVM if he can find his patch for long import lib support in LLVM. I think there is no point to keep it open in meantime.

iainnicol:
> With #89025 it's clear how this approach is lacking and won't help much to stabilise `raw-dylib`.

Can you not use just the syntax

`func_name @ordinal`?

As a separate point, we should probably double-quote the func_name, in case func_name is a def-file keyword.

PR #84573, opened 2021-04-26
[WIP] Never type experiments

Mark-Simulacrum:
This currently just rebases https://github.com/rust-lang/rust/pull/79470 (hopefully all correctly), but will "soon" follow up with an implementation as per https://github.com/rust-lang/lang-team/issues/60#issuecomment-814509681.

r? @ghost

PR #88804, opened 2021-09-10
Revise never type fallback algorithm

Mark-Simulacrum:
This is a rebase of https://github.com/rust-lang/rust/pull/84573, but dropping the stabilization of never type (and the accompanying large test diff).

Each commit builds & has tests updated alongside it, and could be reviewed in a more or less standalone fashion. But it may make more sense to review the PR as a whole, I'm not sure. It should be noted that tests being updated isn't really a good indicator of final behavior -- never_type_fallback is not enabled by default in this PR, so we can't really see the full effects of the commits here.

This combines the work by Niko, which is [documented in this gist](https://gist.github.com/nikomatsakis/7a07b265dc12f5c3b3bd0422018fa660), with some additional rules largely derived to target specific known patterns that regress with the algorithm solely derived by Niko. We build these from an intuition that:

* In general, fallback to `()` is *sound* in all cases
* But, in general, we *prefer* fallback to `!` as it accepts more code, particularly that written to intentionally use `!` (e.g., Result's with a Infallible/! variant).

When evaluating Niko's proposed algorithm, we find that there are certain cases where fallback to `!` leads to compilation failures in real-world code, and fallback to `()` fixes those errors. In order to allow for stabilization, we need to fix a good portion of these patterns.

The final rule set this PR proposes is that, by default, we fallback from `?T` to `!`, with the following exceptions:

1. `?T: Foo` and `Bar::Baz = ?T` and `(): Foo`, then fallback to `()`
2. Per [Niko's algorithm](https://gist.github.com/nikomatsakis/7a07b265dc12f5c3b3bd0422018fa660#proposal-fallback-chooses-between--and--based-on-the-coercion-graph), the "live" `?T` also fallback to `()`.

The first rule is necessary to address a fairly common pattern which boils down to something like the snippet below. Without rule 1, we do not see the closure's return type as needing a () fallback, which leads to compilation failure.

```rust
#![feature(never_type_fallback)]

trait Bar { }
impl Bar for () {  }
impl Bar for u32 {  }

fn foo<R: Bar>(_: impl Fn() -> R) {}

fn main() {
    foo(|| panic!());
}
```

r? @jackh726

### DIFF
--- a/compiler/rustc_middle/src/ty/mod.rs
+++ b/compiler/rustc_middle/src/ty/mod.rs
@@ -2090,3 +2090,16 @@ impl<'tcx> fmt::Debug for SymbolName<'tcx> {
         fmt::Display::fmt(&self.name, fmt)
     }
 }
+
+#[derive(Debug, Default, Copy, Clone)]
+pub struct FoundRelationships {
+    /// This is true if we identified that this Ty (`?T`) is found in a `?T: Foo`
+    /// obligation, where:
+    ///
+    ///  * `Foo` is not `Sized`
+    ///  * `(): Foo` may be satisfied
+    pub self_in_trait: bool,
+    /// This is true if we identified that this Ty (`?T`) is found in a `<_ as
+    /// _>::AssocType = ?T`
+    pub output: bool,
+}
--- a/compiler/rustc_trait_selection/src/traits/relationships.rs
+++ b/compiler/rustc_trait_selection/src/traits/relationships.rs
@@ -1,0 +1,69 @@
+use crate::infer::InferCtxt;
+use crate::traits::query::evaluate_obligation::InferCtxtExt;
+use crate::traits::{ObligationCause, PredicateObligation};
+use rustc_infer::traits::TraitEngine;
+use rustc_middle::ty::{self, ToPredicate};
+
+pub(crate) fn update<'tcx, T>(
+    engine: &mut T,
+    infcx: &InferCtxt<'_, 'tcx>,
+    obligation: &PredicateObligation<'tcx>,
+) where
+    T: TraitEngine<'tcx>,
+{
+    // (*) binder skipped
+    if let ty::PredicateKind::Trait(predicate) = obligation.predicate.kind().skip_binder() {
+        if let Some(ty) =
+            infcx.shallow_resolve(predicate.self_ty()).ty_vid().map(|t| infcx.root_var(t))
+        {
+            if infcx
+                .tcx
+                .lang_items()
+                .sized_trait()
+                .map_or(false, |st| st != predicate.trait_ref.def_id)
+            {
+                let new_self_ty = infcx.tcx.types.unit;
+
+                let trait_ref = ty::TraitRef {
+                    substs: infcx
+                        .tcx
+                        .mk_substs_trait(new_self_ty, &predicate.trait_ref.substs[1..]),
+                    ..predicate.trait_ref
+                };
+
+                // Then contstruct a new obligation with Self = () added
+                // to the ParamEnv, and see if it holds.
+                let o = rustc_infer::traits::Obligation::new(
+                    ObligationCause::dummy(),
+                    obligation.param_env,
+                    obligation
+                        .predicate
+                        .kind()
+                        .map_bound(|_| {
+                            // (*) binder moved here
+                            ty::PredicateKind::Trait(ty::TraitPredicate {
+                                trait_ref,
+                                constness: predicate.constness,
+                            })
+                        })
+                        .to_predicate(infcx.tcx),
+                );
+                // Don't report overflow errors. Otherwise equivalent to may_hold.
+                if let Ok(result) = infcx.probe(|_| infcx.evaluate_obligation(&o)) {
+                    if result.may_apply() {
+                        engine.relationships().entry(ty).or_default().self_in_trait = true;
+                    }
+                }
+            }
+        }
+    }
+
+    if let ty::PredicateKind::Projection(predicate) = obligation.predicate.kind().skip_binder() {
+        // If the projection predicate (Foo::Bar == X) has X as a non-TyVid,
+        // we need to make it into one.
+        if let Some(vid) = predicate.ty.ty_vid() {
+            debug!("relationship: {:?}.output = true", vid);
+            engine.relationships().entry(vid).or_default().output = true;
+        }
+    }
+}
--- a/compiler/rustc_typeck/src/check/fallback.rs
+++ b/compiler/rustc_typeck/src/check/fallback.rs
@@ -1,6 +1,9 @@
 use crate::check::FnCtxt;
 use rustc_data_structures::{
-    fx::FxHashMap, graph::vec_graph::VecGraph, graph::WithSuccessors, stable_set::FxHashSet,
+    fx::FxHashMap,
+    graph::WithSuccessors,
+    graph::{iterate::DepthFirstSearch, vec_graph::VecGraph},
+    stable_set::FxHashSet,
 };
 use rustc_middle::ty::{self, Ty};
 
@@ -280,7 +283,7 @@ impl<'tcx> FnCtxt<'_, 'tcx> {
         // type variable. These will typically default to `!`, unless
         // we find later that they are *also* reachable from some
         // other type variable outside this set.
-        let mut roots_reachable_from_diverging = FxHashSet::default();
+        let mut roots_reachable_from_diverging = DepthFirstSearch::new(&coercion_graph);
         let mut diverging_vids = vec![];
         let mut non_diverging_vids = vec![];
         for unsolved_vid in unsolved_vids {
@@ -293,16 +296,21 @@ impl<'tcx> FnCtxt<'_, 'tcx> {
             );
             if diverging_roots.contains(&root_vid) {
                 diverging_vids.push(unsolved_vid);
+                roots_reachable_from_diverging.push_start_node(root_vid);
+
                 debug!(
                     "calculate_diverging_fallback: root_vid={:?} reaches {:?}",
                     root_vid,
                     coercion_graph.depth_first_search(root_vid).collect::<Vec<_>>()
                 );
-                roots_reachable_from_diverging.extend(coercion_graph.depth_first_search(root_vid));
+
+                // drain the iterator to visit all nodes reachable from this node
+                roots_reachable_from_diverging.complete_search();
             } else {
                 non_diverging_vids.push(unsolved_vid);
             }
         }
+
         debug!(
             "calculate_diverging_fallback: roots_reachable_from_diverging={:?}",
             roots_reachable_from_diverging,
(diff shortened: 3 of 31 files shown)


Mark-Simulacrum:
@jackh726 The opening comment is hopefully helpful, but in terms of review I feel like the best bet is for you to take a look at the PR as a whole at a high level, calling out notable deficiencies or questions before we dive in to each commit in detail. But really up to you.

At a high level, this PR has the first several commits which add Niko's coercion graph and detect when we need to fallback to (), and then the one commit from me atop that which includes the special case and code necessary to handle it. The coercion graph is fairly self explanatory in some sense in terms of the code, though is likely not very intuitive in terms of the mapping to effect on actual Rust code. I'm not sure I fully grasp it at a low level - mostly because coercion and subtyping predicates aren't (to me) obviously visible in Rust source code.

My special casing instruments predicate registration to capture the appropriate projection and trait predicates and stores away the inference variables. I'm not confident in this code being placed appropriately -- happy to move it around. My main goal was to avoid a dependency on `select_obligations_where_possible` not being called "before" us and eliminating the obligations we need to successfully determine which direction we should fall back. That said, it's possible I got some things wrong there.

There's also a few commented sections and FIXMEs, some of which likely could use a more experienced hand to help with -- but I don't think they represent major roadblocks, more like polish, so we can iterate on that after the high level direction I suspect.

----

FWIW, I believe this PR shouldn't affect end-user visible behavior modulo new ICEs or something (not that I'm aware of any). So it can be merged once we deem it ready, and then we'd follow up with a stabilization PR and testing. My impression based on discussions with @nikomatsakis is that never type is sufficiently unstable that iterating like this on nightly is OK -- we will want to poll T-lang for opinions on the algorithm as part of a stabilization report, of course, but there is no need to surface before then. Niko's version of the algorithm was already discussed in T-lang as well, and my addition is a pretty small delta anyway.

Mark-Simulacrum:
(Rebased to drop first commit, now that the PR has been merged).

compiler/rustc_typeck/src/check/fallback.rs
```diff
@@ -373,6 +407,15 @@ impl<'tcx> FnCtxt<'_, 'tcx> {
...
                     Some((a_vid, b_vid))
+                } else if let ty::PredicateKind::Subtype(ty::SubtypePredicate {
+                    a_is_expected: _,
+                    a,
+                    b,
+                }) = atom
+                {
+                    let a_vid = self.root_vid(a)?;
+                    let b_vid = self.root_vid(b)?;
+                    Some((a_vid, b_vid))
```
jackh726, line 418: I think this needs a comment...

Is the idea that even though it's not a "coercion", it still counts because we effectively unify them?

compiler/rustc_typeck/src/check/fallback.rs
```diff
@@ -384,7 +427,7 @@ impl<'tcx> FnCtxt<'_, 'tcx> {
     }
 
     /// If `ty` is an unresolved type variable, returns its root vid.
-    fn root_vid(&self, ty: Ty<'tcx>) -> Option<ty::TyVid> {
+    pub fn root_vid(&self, ty: Ty<'tcx>) -> Option<ty::TyVid> {
```
jackh726, line 430: Does this need to be `pub`? I'm now seeing where it gets used.

src/test/ui/never_type/fallback-closure-wrap.rs
```diff
@@ -0,0 +1,17 @@
+// check-pass
```
jackh726, line 1: Can you add a comment for this test?

And add the example in the OP (for projection outputs) as a test?

compiler/rustc_typeck/src/check/fallback.rs
```diff
@@ -334,21 +349,40 @@ impl<'tcx> FnCtxt<'_, 'tcx> {
...
+
+            for (vid, rel) in relationships.iter() {
+                //if self.infcx.shallow_resolve(*ty).ty_vid().map(|t| self.infcx.root_var(t))
+                if self.infcx.root_var(*vid) == root_vid {
+                    relationship.self_in_trait |= rel.self_in_trait;
+                    relationship.output |= rel.output;
+                }
+            }
+
+            if relationship.self_in_trait && relationship.output {
```
jackh726, line 378: Please add a comment here about why this is important, relevant test, etc.

jackh726 (changes requested):
A few comments. Most of the commits look good. I have some reservations about the last commit, as discussed on [zulip](https://rust-lang.zulipchat.com/#narrow/stream/259160-t-lang.2Fproject-never-type/topic/PR.20.2388804). But, I've reviewed based on "the logic here is what we want, at least for now" and "there might be another way to do this, but this is straightforward enough"

jackh726:
Going to queue a perf run. That last commit looks suspicious.

@bors try @rust-timer queue

jackh726:
Wow! Surprising...did I miss something.

compiler/rustc_trait_selection/src/traits/relationships.rs
```diff
@@ -0,0 +1,89 @@
...
+    infcx: &InferCtxt<'_, 'tcx>,
+    obligation: &PredicateObligation<'tcx>,
+) where
+    T: TraitEngine<'tcx>,
+{
+    if let ty::PredicateKind::Trait(predicate) = obligation.predicate.kind().skip_binder() {
+        if predicate.trait_ref.def_id
+            != infcx.tcx.require_lang_item(rustc_hir::LangItem::Sized, None)
+        {
+            // fixme: copy of mk_trait_obligation_with_new_self_ty
```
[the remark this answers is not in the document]
Mark-Simulacrum: Ah, dropped this comment now. That function doesn't take the right parameters, so it can't preserve constness for example -- there might be a refactoring of that function in order, not sure. It's also private to the error_reporting module...

compiler/rustc_trait_selection/src/traits/relationships.rs
```diff
@@ -0,0 +1,89 @@
...
+                // (if this ends up being important) we just don't register a relationship and then end up falling back to !,
+                // which is not terrible.
+
+                //engine.normalize_projection_type(
+                //    infcx,
+                //    obligation.param_env,
+                //    predicate.projection_ty,
+                //    obligation.cause.clone(),
+                //);
+            }
```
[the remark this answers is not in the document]
Mark-Simulacrum: Yeah, this is a bit complicated. I think it's probably fine to just delete the else branch here. Unfortunately, that does mean that some cases won't be considered, I think -- for example, if you register `<ConcreteType as Trait>::Proj = Foo`, that will not end up registering as 'live' for any of the involved types, even if we eventually unify `<ConcreteType as Trait>::Proj` with `?T`, I think, so we may miss that we wanted to fall back to () in this case.

But it should be sound, still, so I don't think this is actually a problem, just 'less nice', in some sesne.

compiler/rustc_trait_selection/src/traits/relationships.rs
```diff
@@ -0,0 +1,89 @@
...
+                    .map_bound(|_| {
+                        ty::PredicateKind::Trait(ty::TraitPredicate {
+                            trait_ref,
+                            constness: predicate.constness,
+                        })
+                    })
+                    .to_predicate(infcx.tcx),
+            );
+            // Don't report overflow errors. Otherwise equivalent to may_hold.
+            if let Ok(result) = infcx.probe(|_| infcx.evaluate_obligation(&o)) {
```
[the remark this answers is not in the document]
nikomatsakis: I had the same thought. It seems like we should only do this if we observe that some potentially diverging variable is involved. Is there a reason that doesn't work? (Maybe this was covered on zulip?)

compiler/rustc_middle/src/ty/mod.rs
```diff
@@ -2088,3 +2088,16 @@ impl<'tcx> fmt::Debug for SymbolName<'tcx> {
...
+#[derive(Debug, Default, Copy, Clone)]
+pub struct FoundRelationships {
+    /// This is true if we identified that this Ty (`?T`) is found in a `?T: Foo`
+    /// obligation, where:
+    ///
+    ///  * `Foo` is not `Sized`
+    ///  * `(): Foo` may be satisfied
+    pub self_in_trait: bool,
+    /// This is true if we identified that this Ty (`?T`) is found in a `<_ as
+    /// _>::AssocType = ?T`
```
nikomatsakis, line 2101: We should document *why* these conditions are being used -- but it would be fine (imo) for this to be on a never-type-initiative repo (and linked from here).

Mark-Simulacrum:
Ok, pushed up a rebase that actually ended up only touching the last commit (last two, now). Going to follow up to several comments here:

> We should document why these conditions are being used -- but it would be fine (imo) for this to be on a never-type-initiative repo (and linked from here).

I am not sure that we will end up with "great" reasoning beyond the 'it fixes regressions'. In discussions we've had about the fallback rules we *could* propose, I believe we largely ended up concluding that there's not really a rule or set of rules we can come up with that *is* principled that covers a good portion of the "remaining regressions" found in the wild.

> [requests for additional tests]

The last commit now adds fallback/nofallback revisions to several of the tests in tree to showcase the need for fallback. However, it also shows several tests that are not fixed by the current set of fallback rules (some pre-existing, some new in this PR):

* src/test/ui/never_type/fallback-closure-wrap.rs: This test requires us to consider an `as` cast for fallback, which is explicitly not part of fallback today - https://github.com/rust-lang/rust/blob/207d9558d00dd5cc438a6418ba96912d396e2155/compiler/rustc_typeck/src/check/mod.rs#L459-L460
* src/test/ui/never_type/defaulted-never-note.rs
   * This is a test that *could* be tackled with a simpler rule to the one we've added here, limited to just `?T: Foo` and `(): Foo`. IIRC, we need the more complicated rule because stabilization otherwise triggers errors in other tests. I think [binop-3](https://github.com/rust-lang/rust/blob/207d9558d00dd5cc438a6418ba96912d396e2155/src/test/ui/binop/binop-fail-3.rs) is a good example here. If we just have `?T: Foo` and `(): Foo`, then one side there will fallback to `()` (since it is `?T: PartialEq<_>`, but the other side has no such condition so results in `!` and `! == ()` is not implemented.
* src/test/ui/never_type/diverging-fallback-no-leak.rs
   * This seems like basically the same test as never-note, and I think is similarly explained as somewhat unavoidable.


@bors try @rust-timer queue

Rechecking performance after fixing our code to avoid testing `(): Foo` when we have a concrete type in the original predicate, but rather only for type vars. I don't think we can limit this to just *diverging* type vars, because those are only tracked when we detect a never to any adjustment -- and that may not have happened when we register this bound. This is basically the same reason we have to check these as they get registered rather than waiting until fallback and looking at the pending obligations.

(That said, the previous diff did not look terrible for what this code is doing, and I think we could accept the hit. I cannot readily explain the *improvements* -- I would expect the code added by this PR to mostly only make things worse).


---

I think the next steps here might be that we:

* Do several more rounds of review to get this in a "good enough" state to land, though likely not to stabilize never type
   * Would rather get this in-tree sooner rather than later to avoid rebasing it too much; it's gated behind feature gates anyway, so we don't need to decide on the specific rule set and get T-lang approval yet. (IMO, anyway).
* Open a PR stabilizing never type and collect a fresh round of crater data. My projection is ~32 regressions, with around 70 fixed by the last commit in this PR (based on comparison with the crater run with just Niko's commits).

jackh726:
r=me here. I would feel better if @nikomatsakis can take a look and give the final signoff, but given that this is gated behind a feature gate, it might be better to merge and iterate in-tree.

nikomatsakis:
> I am not sure that we will end up with "great" reasoning beyond the 'it fixes regressions'.

I think that is fine, I'd just like to comment on what kind of regression we are looking to fix.

Mark-Simulacrum:
> > I am not sure that we will end up with "great" reasoning beyond the 'it fixes regressions'.
> 
> I think that is fine, I'd just like to comment on what kind of regression we are looking to fix.

There's now a comment in the fallback code pointing to the regression test the relationship code is fixing. Happy to edit it, or expand it - I think creating that repository for the alternatives documentation is a good idea - once I rebase the stabilization commits it'll be easier to off hand indicate which patterns break with particular algorithms without needing to refer to various hackmds :)

jackh726:
@bors r=nikomatsakis,jackh726

even though it might be procedural, I'd like to note that I did review this (particularly because Niko did originally author many of these commits)

PR #88805, opened 2021-09-10
Clarification of default socket flags

krhancoc:
This PR outlines the decision to disable inheritance of socket objects when possible to child processes in the documentation.

### DIFF
--- a/library/std/src/net/mod.rs
+++ b/library/std/src/net/mod.rs
@@ -15,7 +15,7 @@
 //!   with networking objects like [`TcpListener`], [`TcpStream`] or [`UdpSocket`]
 //! * Other types are return or parameter types for various methods in this module
 //!
-//! Rust disables inheritance of socket objects to child processes by default when possible.  For 
+//! Rust disables inheritance of socket objects to child processes by default when possible.  For
 //! example, through the use of the `CLOEXEC` flag in UNIX systems or the `HANDLE_FLAG_INHERIT`
 //! flag on Windows.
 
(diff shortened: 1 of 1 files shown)


krhancoc:
As an additional comment, I was unsure where this form of documentation belongs as its a design decision that is done throughout the Rust networking stack, but also is an implementation detail.  I do think default options/flags should be explicitly stated somewhere though.

yaahc:
> As an additional comment, I was unsure where this form of documentation belongs as its a design decision that is done throughout the Rust networking stack, but also is an implementation detail.

This seems like as good a place as any. The other option I can think of would be to copy-paste this same explanation into the docs of every function where this info applies.

> I do think default options/flags should be explicitly stated somewhere though.

That's fair, though I want to note that we treat comments like these in our documentations as part of our stable API so this will require approval from the entire libs-api team.

@rust-lang/libs-api do we want to document a commitment to disable socket inheritance where possible?

@rfcbot merge

joshtriplett:
This is generally true for any kind of handle or file descriptor: we always set cloexec or equivalent. This doesn't seem like a change in our policy, just documentation of the existing policy.

@rfcbot reviewed

joshtriplett:
@krhancoc As a separate PR, please. And thank you for working on this!

camelid:
triage: It looks like the FCP finished. @yaahc I believe this is waiting on your review.

dtolnay (approved):
:heart:

PR #88810, opened 2021-09-10
rustdoc: Cleanup `clean` part 1

camelid:
Split out from #88379.

These commits are completely independent of each other, and each is a fairly
small change (the last few are new commits; they are not from #88379):

- Remove unnecessary `Cache.*_did` fields
- rustdoc: Get symbol for `TyParam` directly
- Create a valid `Res` in `external_path()`
- Remove unused `hir_id` parameter from `resolve_type`
- Fix redundant arguments in `external_path()`
- Remove unnecessary `is_trait` argument
- rustdoc: Cleanup a pattern match in `external_generic_args()`

r? @jyn514

### DIFF
--- a/src/librustdoc/clean/utils.rs
+++ b/src/librustdoc/clean/utils.rs
@@ -414,7 +414,7 @@ crate fn resolve_type(cx: &mut DocContext<'_>, path: Path, id: hir::HirId) -> Ty
             return Generic(kw::SelfUpper);
         }
         Res::Def(DefKind::TyParam, _) if path.segments.len() == 1 => {
-            return Generic(Symbol::intern(&path.whole_name()));
+            return Generic(path.segments[0].name);
         }
         Res::SelfTy(..) | Res::Def(DefKind::TyParam | DefKind::AssocTy, _) => true,
         _ => false,
--- a/src/librustdoc/clean/utils.rs
+++ b/src/librustdoc/clean/utils.rs
@@ -141,19 +141,21 @@ fn external_generic_args(
     }
 }
 
-// trait_did should be set to a trait's DefId if called on a TraitRef, in order to sugar
-// from Fn<(A, B,), C> to Fn(A, B) -> C
+/// trait_did should be set to a trait's DefId if called on a TraitRef, in order to sugar
+/// from `Fn<(A, B,), C>` to `Fn(A, B) -> C`
 pub(super) fn external_path(
     cx: &mut DocContext<'_>,
-    name: Symbol,
+    did: DefId,
     trait_did: Option<DefId>,
     has_self: bool,
     bindings: Vec<TypeBinding>,
     substs: SubstsRef<'_>,
 ) -> Path {
+    let def_kind = cx.tcx.def_kind(did);
+    let name = cx.tcx.item_name(did);
     Path {
         global: false,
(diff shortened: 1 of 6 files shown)


camelid:
It looks like 6b2ee44 is causing CI to fail because some part of rustdoc relies on the `Path` having a `Res::Err`. But it works fine in my larger "Cleanup `clean`" PR, so one of my other cleanups must have fixed that other part of rustdoc in some way. I'll drop that commit from this one and leave it for part 2.

camelid:
Actually, the test failure was just because the ordering of the `where` bounds changed (this is why it changed in the "Cleanup `clean`" PR). So, I re-added the commit and updated its commit message:

> The order of the `where` bounds on auto trait impls changed because
    rustdoc currently sorts auto trait `where` bounds based on the `Debug`
    output for the bound. Now that the bounds have an actual `Res`, they are
    being unintentionally sorted by their `DefId` rather than their path.
    So, I had to update a test for the change in ordering of the rendered
    bounds.

src/librustdoc/html/render/mod.rs
```diff
@@ -1165,7 +1165,7 @@ fn render_deref_methods(
     }
 }
 
-fn should_render_item(item: &clean::Item, deref_mut_: bool, cache: &Cache) -> bool {
+fn should_render_item(item: &clean::Item, deref_mut_: bool, cx: &Context<'_>) -> bool {
```
jyn514, line 1168: Same nit as on the other PR, this could take a tcx instead.

src/librustdoc/clean/utils.rs
```diff
@@ -141,19 +141,21 @@ fn external_generic_args(
...
 
-// trait_did should be set to a trait's DefId if called on a TraitRef, in order to sugar
-// from Fn<(A, B,), C> to Fn(A, B) -> C
+/// trait_did should be set to a trait's DefId if called on a TraitRef, in order to sugar
+/// from `Fn<(A, B,), C>` to `Fn(A, B) -> C`
 pub(super) fn external_path(
     cx: &mut DocContext<'_>,
-    name: Symbol,
+    did: DefId,
     trait_did: Option<DefId>,
```
jyn514, line 149: `did` and `trait_did` will never be different if `trait_did.is_some()`, right? Could you change `trait_did: Option<DefId>` to `is_trait: bool` instead to avoid them getting out of sync?

jyn514 (approved):
LGTM with nits fixed

src/librustdoc/html/render/mod.rs
camelid: Ah, thanks for catching that. I forgot to cherry-pick the commit from the other PR that fixed it. Done.

src/librustdoc/clean/utils.rs
camelid: I did that, and then I did one better: It actually looks like `is_trait` is not even necessary. It is only used for sugaring `Fn` trait bounds, and rustdoc already checks that the `did` is a `Fn*` lang item.

camelid: I made them as two separate commits, but I can squash them together if you would prefer.

jyn514 (approved):

PR #88814, opened 2021-09-10
No --as-needed for -fuse-ld=lld

sledgehammervampire:
I have a crate that uses https://github.com/Andlon/mkl-sys as a dependency, which does not work when the rustflag `-C link-arg=-fuse-ld=lld` is used, but works when the flag is not used. It seems like this is because using --as-needed with lld makes lld not pick up some libraries which are actually needed.

This is the binary compiled using  `-C link-arg=-fuse-ld=lld` without my PR:
```
$ readelf -d target/debug/deps/spam_csr-f895edd13d6929fd | rg NEEDED
 0x0000000000000001 (NEEDED)             Shared library: [libmkl_intel_ilp64.so.1]
 0x0000000000000001 (NEEDED)             Shared library: [libpthread.so.0]
 0x0000000000000001 (NEEDED)             Shared library: [libm.so.6]
 0x0000000000000001 (NEEDED)             Shared library: [libdl.so.2]
 0x0000000000000001 (NEEDED)             Shared library: [libgcc_s.so.1]
 0x0000000000000001 (NEEDED)             Shared library: [libc.so.6]
 0x0000000000000001 (NEEDED)             Shared library: [ld-linux-x86-64.so.2]
```

This is the binary compiled without:
```
$ readelf -d target/debug/deps/spam_csr-f895edd13d6929fd | rg NEEDED
 0x0000000000000001 (NEEDED)             Shared library: [libmkl_intel_ilp64.so.1]
 0x0000000000000001 (NEEDED)             Shared library: [libmkl_intel_thread.so.1]
 0x0000000000000001 (NEEDED)             Shared library: [libmkl_core.so.1]
 0x0000000000000001 (NEEDED)             Shared library: [libiomp5.so]
 0x0000000000000001 (NEEDED)             Shared library: [libpthread.so.0]
 0x0000000000000001 (NEEDED)             Shared library: [libm.so.6]
 0x0000000000000001 (NEEDED)             Shared library: [libdl.so.2]
 0x0000000000000001 (NEEDED)             Shared library: [libgcc_s.so.1]
 0x0000000000000001 (NEEDED)             Shared library: [libc.so.6]
 0x0000000000000001 (NEEDED)             Shared library: [ld-linux-x86-64.so.2]
```

This is the binary compiled using  `-C link-arg=-fuse-ld=lld` with my PR:
```
$ readelf -d target/debug/deps/spam_csr-740b6e93024ccac3 | rg NEEDED
 0x0000000000000001 (NEEDED)             Shared library: [libmkl_intel_ilp64.so.1]
 0x0000000000000001 (NEEDED)             Shared library: [libmkl_intel_thread.so.1]
 0x0000000000000001 (NEEDED)             Shared library: [libmkl_core.so.1]
 0x0000000000000001 (NEEDED)             Shared library: [libiomp5.so]
 0x0000000000000001 (NEEDED)             Shared library: [libpthread.so.0]
 0x0000000000000001 (NEEDED)             Shared library: [libm.so.6]
 0x0000000000000001 (NEEDED)             Shared library: [libdl.so.2]
 0x0000000000000001 (NEEDED)             Shared library: [libgcc_s.so.1]
 0x0000000000000001 (NEEDED)             Shared library: [libutil.so.1]
 0x0000000000000001 (NEEDED)             Shared library: [librt.so.1]
 0x0000000000000001 (NEEDED)             Shared library: [libc.so.6]
 0x0000000000000001 (NEEDED)             Shared library: [ld-linux-x86-64.so.2]
```

### DIFF
--- a/compiler/rustc_codegen_ssa/src/back/link.rs
+++ b/compiler/rustc_codegen_ssa/src/back/link.rs
@@ -2426,7 +2426,14 @@ fn add_upstream_native_libraries(
                 NativeLibKind::Dylib { as_needed } => {
                     cmd.link_dylib(name, verbatim, as_needed.unwrap_or(true))
                 }
-                NativeLibKind::Unspecified => cmd.link_dylib(name, verbatim, true),
+                NativeLibKind::Unspecified => {
+                    cmd.link_dylib(
+                        name,
+                        verbatim,
+                        // FIXME: lld doesn't include some needed shared libraries without this, but includes some unneeded ones with this
+                        sess.opts.cg.link_args.iter().all(|arg| arg != "-fuse-ld=lld"),
+                    )
+                }
                 NativeLibKind::Framework { as_needed } => {
                     cmd.link_framework(name, as_needed.unwrap_or(true))
                 }


davidtwco (approved):
This looks good to me, but I'm not incredibly familiar with our behavior and the implications of things here, so going to get a second opinion.

davidtwco:
Oops, r? @petrochenkov

compiler/rustc_codegen_ssa/src/back/link.rs
```diff
@@ -2426,7 +2426,14 @@ fn add_upstream_native_libraries(
                 NativeLibKind::Dylib { as_needed } => {
                     cmd.link_dylib(name, verbatim, as_needed.unwrap_or(true))
                 }
-                NativeLibKind::Unspecified => cmd.link_dylib(name, verbatim, true),
+                NativeLibKind::Unspecified => {
+                    cmd.link_dylib(
+                        name,
+                        verbatim,
+                        // FIXME: lld doesn't include some needed shared libraries without this, but includes some unneeded ones with this
+                        sess.opts.cg.link_args.iter().all(|arg| arg != "-fuse-ld=lld"),
```
bjorn3, line 2434: If no symbols from a dylib are used by the current dso, like I suspect is the case for you, `--as-needed` will not link the dylib. According to https://rust-lang.github.io/rfcs/2951-native-link-modifiers.html we may actually want to extend this behavior to all platforms by default. It provides the `-as-needed` linkage modifier to explicitly use `--no-as-needed`. It is not yet stabilized though.

1000teslas: Why would the symbols not be used? What behavior should be extended to all platforms by default? Using `--as-needed`? I didn't know my change was platform specific somehow. That RFC definitely went over my head lol.

nagisa:
> using --as-needed with lld makes lld not pick up some libraries which are actually needed.

This smells to me like potentially an incorrect ordering of linker arguments (specifically the `-l` arguments) or some similar problem instead. `--as-needed` is most likely a red herring here.

1000teslas:
But why would it work with ld from gcc and not lld then? Isn't lld supposed to be compatible with ld from gcc?

compiler/rustc_codegen_ssa/src/back/link.rs
bjorn3: I think only symbols from `mkl_core` are actually used by the binary. I think `mkl_core` itself depends on symbols provided by `mkl_intel_ilp64` or `mkl_intel_lp64` and `mkl_intel_thread` or `mkl_sequential` without declaring a dependency on any of them. `--as-needed` doesn't add a dependency on the latter libraries as they aren't used by the binary. It just so happens that during dynamic linking at program startup symbols provided by the latter libraries are actually used by the `mkl_core` version used at runtime. The `mkl_core` version used at runtime may be different from the version used while linking and as such have different symbol dependencies. They may even be provided by completely different dynamic libraries. As for why it works at all with ld.bfd (the default linker on most linux systems) I have no clue.

> What behavior should be extended to all platforms by default? Using --as-needed?

Yes

> That RFC definitely went over my head lol.

The RFC allows you to specify how to link against libraries. In you case you would need something like:

```rust
#[link(name = "mkl_intel_ilp64", modifiers = "-as-needed")]
extern "C" {}
#[link(name = "mkl_intel_thread", modifiers = "-as-needed")]
extern "C" {}
```

or equivalent in the build script:

```rust
println!("cargo:rustc-link-lib=dylib:-as-needed=mkl_intel_ilp64");
println!("cargo:rustc-link-lib=dylib:-as-needed=mkl_intel_thread");
```

Unfortunately this hasn't been stabilized yet.

petrochenkov:
@1000teslas 
Could you show the `NEEDED` entries and symbols for all of the `mkl` libraries?
I also suspect that this is an issue with either linking order, or some kind of missing dependencies between `mkl` libraries.

Even if `mkl` libraries hit some LLD bug, then using the `-as-needed` modifier to work it around seems more appropriate than disabling `--as-needed` on LLD for all libraries.

1000teslas:
@petrochenkov Here are the various NEEDED sections:

```
$ readelf -d /nix/store/kp0akwshg31bc5mj9qbfzjswr1sa3xvi-mkl-2021.1.1.52/lib/libmkl_intel_ilp64.so | rg NEEDED
 0x0000000000000001 (NEEDED)             Shared library: [libdl.so.2]
```

```
$ readelf -d /nix/store/kp0akwshg31bc5mj9qbfzjswr1sa3xvi-mkl-2021.1.1.52/lib/libmkl_intel_thread.so | rg NEEDED
 0x0000000000000001 (NEEDED)             Shared library: [libdl.so.2]
 0x0000000000000001 (NEEDED)             Shared library: [libpthread.so.0]
```

```
$ readelf -d /nix/store/kp0akwshg31bc5mj9qbfzjswr1sa3xvi-mkl-2021.1.1.52/lib/libmkl_core.so | rg NEEDED
 0x0000000000000001 (NEEDED)             Shared library: [libdl.so.2]
 0x0000000000000001 (NEEDED)             Shared library: [libpthread.so.0]
```

```
$ readelf -d /nix/store/kp0akwshg31bc5mj9qbfzjswr1sa3xvi-mkl-2021.1.1.52/lib/libiomp5.so | rg NEEDED
 0x0000000000000001 (NEEDED)             Shared library: [libgcc_s.so.1]
 0x0000000000000001 (NEEDED)             Shared library: [libpthread.so.0]
 0x0000000000000001 (NEEDED)             Shared library: [libc.so.6]
 0x0000000000000001 (NEEDED)             Shared library: [ld-linux-x86-64.so.2]
 0x0000000000000001 (NEEDED)             Shared library: [libdl.so.2]
```

I'm not sure of a way to get the symbols. Something like `readelf --syms library.so`? That prints out quite a lot of stuff. Do you want the output as a file?

petrochenkov:
Posting the results of `readelf -a` would be most useful, it's indeed a lot of text, so uploading it as a file would be reasonable.

1000teslas:
@petrochenkov Here is the output of `readelf -a` for each library.
[libiomp5.txt](https://github.com/rust-lang/rust/files/7143608/libiomp5.txt)
[libmkl_core.txt](https://github.com/rust-lang/rust/files/7143609/libmkl_core.txt)
[libmkl_intel_ilp64.txt](https://github.com/rust-lang/rust/files/7143610/libmkl_intel_ilp64.txt)
[libmkl_intel_thread.txt](https://github.com/rust-lang/rust/files/7143613/libmkl_intel_thread.txt)

petrochenkov:
@1000teslas 
When running the final binary you are observing run-time errors from dynamic linker, right? Could you show what they say?

petrochenkov:
Also, could you build with `-C link-dead-code` and check whether the `NEEDED` entries for `mkl` libraries appear in that case?

bjorn3:
I am pretty sure the following is the case (https://github.com/rust-lang/rust/pull/88814#discussion_r706091854): 

> I think mkl_core itself depends on symbols provided by mkl_intel_ilp64 or mkl_intel_lp64 and mkl_intel_thread or mkl_sequential without declaring a dependency on any of them. --as-needed doesn't add a dependency on the latter libraries as they aren't used by the binary. It just so happens that during dynamic linking at program startup symbols provided by the latter libraries are actually used by the mkl_core version used at runtime.

So `--no-as-needed` is necessary for those specific libraries. I don't know why it works with ld.bfd without `--no-as-needed` though.

1000teslas:
> @1000teslas
> When running the final binary you are observing run-time errors from dynamic linker, right? Could you show what they say?

I get this error
```
/home/rdp/src/spam/target/debug/deps/spam_csr-f895edd13d6929fd: symbol lookup error: /nix/store/kp0akwshg31bc5mj9qbfzjswr1sa3xvi-mkl-2021.1.1.52/lib/libmkl_intel_ilp64.so.1: undefined symbol: mkl_lapack_zhesvxx
```

1000teslas:
> Also, could you build with `-C link-dead-code` and check whether the `NEEDED` entries for `mkl` libraries appear in that case?

No, I get the same error:
```
/home/rdp/src/spam/target/debug/deps/spam_csr-f895edd13d6929fd: symbol lookup error: /nix/store/kp0akwshg31bc5mj9qbfzjswr1sa3xvi-mkl-2021.1.1.52/lib/libmkl_intel_ilp64.so.1: undefined symbol: mkl_lapack_zhesvxx
```
and the mkl libraries do not appear in the NEEDED section
```
$ readelf -d target/debug/deps/spam_csr-f895edd13d6929fd | rg NEEDED
 0x0000000000000001 (NEEDED)             Shared library: [libmkl_intel_ilp64.so.1]
 0x0000000000000001 (NEEDED)             Shared library: [libpthread.so.0]
 0x0000000000000001 (NEEDED)             Shared library: [libm.so.6]
 0x0000000000000001 (NEEDED)             Shared library: [libdl.so.2]
 0x0000000000000001 (NEEDED)             Shared library: [libgcc_s.so.1]
 0x0000000000000001 (NEEDED)             Shared library: [libc.so.6]
 0x0000000000000001 (NEEDED)             Shared library: [ld-linux-x86-64.so.2]
 ```

**Comment by 1000teslas** [comment]
> I am pretty sure the following is the case ([#88814 (comment)](https://github.com/rust-lang/rust/pull/88814#discussion_r706091854)):
> 
> > I think mkl_core itself depends on symbols provided by mkl_intel_ilp64 or mkl_intel_lp64 and mkl_intel_thread or mkl_sequential without declaring a dependency on any of them. --as-needed doesn't add a dependency on the latter libraries as they aren't used by the binary. It just so happens that during dynamic linking at program startup symbols provided by the latter libraries are actually used by the mkl_core version used at runtime.
> 
> So `--no-as-needed` is necessary for those specific libraries. I don't know why it works with ld.bfd without `--no-as-needed` though.

I think NixOS injects libraries into the linker's search path, since the linker is actually a wrapper script around `ld.bfd`/`ld.lld`. I don't have an installation of a more conventional linux distro to test whether the library actually works on them.

**Comment by petrochenkov** [comment]
@1000teslas 
Could you check whether this works with `-fuse-ld=gold`?

**Comment by 1000teslas** [comment]
> @1000teslas
> Could you check whether this works with `-fuse-ld=gold`?

Nope, same error.
```
/home/rdp/src/spam/target/debug/deps/spam_csr-f895edd13d6929fd: symbol lookup error: /nix/store/kp0akwshg31bc5mj9qbfzjswr1sa3xvi-mkl-2021.1.1.52/lib/libmkl_intel_ilp64.so.1: undefined symbol: mkl_lapack_zhesvxx
```

petrochenkov:
I see.
Looks like more modern linkers move into the direction of more explicit specification of dependencies, and also abandoning old BDF stuff like `--copy-dt-needed-entries`/`--add-needed` as well.
In this case using the modifiers (https://github.com/rust-lang/rust/pull/88814#discussion_r706091854) is the optimal solution, IMO.

In the meantime I think you can use raw linker arguments for passing `--no-as-needed` when building your final executable, like `-Clink-arg=-Wl,--no-as-neeeded -Clink-arg=-Wl,-lmkl_intel_thread -Clink-arg=-Wl,-lmkl_core -Clink-arg=-Wl,--as-needed`.

1000teslas:
> I see.
> Looks like more modern linkers move into the direction of more explicit specification of dependencies, and also abandoning old BDF stuff like `--copy-dt-needed-entries`/`--add-needed` as well.
> In this case using the modifiers ([#88814 (comment)](https://github.com/rust-lang/rust/pull/88814#discussion_r706091854)) is the optimal solution, IMO.
> 
> In the meantime I think you can use raw linker arguments for passing `--no-as-needed` when building your final executable, like `-Clink-arg=-Wl,--no-as-neeeded -Clink-arg=-Wl,-lmkl_intel_thread -Clink-arg=-Wl,-lmkl_core -Clink-arg=-Wl,--as-needed`.

Seems like a good enough workaround for me. Thanks.

PR #88815, opened 2021-09-10
Restrict parsing of bare union/struct to field types

### LINKED ISSUE #88583: Type called union wreaks havoc since 1.54

estebank:
Do not unconditionally parse bare types everywhere, only parse them if
they are in a field type, and try them when recovering a misparse
everywhere else, using them only if they are successfuly fully parsed.

Fix #88583.

r? @pnkfelix

### DIFF
--- a/compiler/rustc_parse/src/parser/item.rs
+++ b/compiler/rustc_parse/src/parser/item.rs
@@ -765,10 +765,58 @@ impl<'a> Parser<'a> {
             if self.eat(&token::Colon) { self.parse_generic_bounds(None)? } else { Vec::new() };
         generics.where_clause = self.parse_where_clause()?;
 
-        let default = if self.eat(&token::Eq) { Some(self.parse_ty()?) } else { None };
-        self.expect_semi()?;
+        let default = if self.eat(&token::Eq) {
+            Some(self.parse_ty_recover_anon_adt(|this| this.expect_semi())?.0)
+        } else {
+            self.expect_semi()?;
+            None
+        };
 
         Ok((ident, ItemKind::TyAlias(Box::new(TyAliasKind(def, generics, bounds, default)))))
+    }
+
+    /// Parses a type. If a misparse happens, we attempt to parse the type again, allowing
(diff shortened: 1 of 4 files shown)


estebank:
This PR can be simplified if we do not care to have good diagnostics in _all_ incorrect places where an anonymous ADT might end up in.

estebank:
@petrochenkov could you provide some feedback on what you find objectionable in this PR? The parser recovery included is to keep the same level of diagnostics. Most of it can be safely removed with no loss of functionality, just confusion when trying to use bare ADTs anywhere else than type fields.

estebank:
Closing in favor of #88775, but the changes in this PR need to be included when relanding anon ADTs.

PR #88821, opened 2021-09-10
Use the `matches` macro in `Option::contains`

vrmiguel:
This is a simple code-style PR that makes `Option::contains` use the `matches` macro, as per Clippy's [match_like_matches_macro](https://rust-lang.github.io/rust-clippy/master/#match_like_matches_macro)

### DIFF
--- a/library/core/src/option.rs
+++ b/library/core/src/option.rs
@@ -589,10 +589,7 @@ impl<T> Option<T> {
     where
         U: PartialEq<T>,
     {
-        match self {
-            Some(y) => x == y,
-            None => false,
-        }
+        matches!(self, Some(y) if x == y)
     }
 
     /////////////////////////////////////////////////////////////////////////


jyn514:
@bors try @rust-timer queue

The option/result combinators are used so often that tiny changes often have an outsided impact.

jyn514:
Note that in general, the standard library does not use clippy, and a PR mass converting it will probably not be accepted.

matthiaskrgr:
Seems to cause some `max-rss` changes, I wonder why that is https://perf.rust-lang.org/compare.html?start=497ee321af3b8496eaccd7af7b437f18bab81abf&end=a1e9d1b89790ebfdf34f9687a18bd8646466294b&stat=max-rss

JohnCSimon:
Ping from triage: 
@vrmiguel - Can you address the comments from the reviewer? When you have please set this back to S-waiting-on-reviewer.

jyn514:
@JohnCSimon it's an improvement, not a regression, I'm not sure what there is to address.

kennytm:
I can't get Clippy the show the warning. Surely if the input was `match self { Some(x) if x == y => true, _ => false }` the lint would fire, but the existing code is not. I guess this is unlinted by rust-lang/rust-clippy#6216.

The max-rss decrease in `clap-rs check / full` doesn't quite make sense, since the MIR of the `matches!()` macro version is longer. It is also marked as `-5.37%?`, a `?` meaning uncertain. We could also see the `instructions:u` becomes slightly worse in `ripgrep opt / incr-unchanged` at `+0.31%`.

I lean towards closing this PR as there is no significant positive or negative impact.

Issue #28728, opened 2015-09-29
LLVM loop optimization can make safe programs crash


RalfJung:
The following snippet crashes when compiled in release mode on current stable, beta and nightly:

``` rust
enum Null {}

fn foo() -> Null { loop { } }

fn create_null() -> Null {
    let n = foo();

    let mut i = 0;
    while i < 100 { i += 1; }
    return n;
}

fn use_null(n: Null) -> ! {
    match n { }
}


fn main() {
    use_null(create_null());
}
```

https://play.rust-lang.org/?gist=1f99432e4f2dccdf7d7e&version=stable

This is based on the following example of LLVM removing a loop that I was made aware of: https://github.com/simnalamburt/snippets/blob/12e73f45f3/rust/infinite.rs.
What seems to happen is that since C allows LLVM to remove endless loops that have no side-effect, we end up executing a `match` that has to arms.

ranma42:
The LLVM IR of the optimised code is

```llvm
; Function Attrs: noreturn nounwind readnone uwtable
define internal void @_ZN4main20h5ec738167109b800UaaE() unnamed_addr #0 {
entry-block:
  unreachable
}
```

This kind of optimisation breaks the main assumption that should normally hold on uninhabited types: it should be impossible to have a value of that type.
rust-lang/rfcs#1216 proposes to explicitly handle such types in Rust. It might be effective in ensuring that LLVM never has to handle them and in injecting the appropriate code to ensure divergence when needed (IIUIC this could be achieved with appropriate attributes or intrinsic calls).
This topic has also been recently discussed in the LLVM mailing list: http://lists.llvm.org/pipermail/llvm-dev/2015-July/088095.html

alexcrichton:
triage: I-nominated

Seems bad! If LLVM doesn't have a way to say "yes, this loop really is infinite" though then we may just have to sit-and-wait for the upstream discussion to settle.

ranma42:
A way to prevent infinite loops from being optimised away is to add `unsafe {asm!("" :::: "volatile")}` inside of them. This is similar to the `llvm.noop.sideeffect` intrinsic that has been proposed in the LLVM mailing list, but it might prevent some optimisations.
In order to avoid the performance loss and to still guarantee that diverging functions/loops are not optimised away, I believe that it should be sufficient to insert an empty non-optimisable loop (i.e. `loop { unsafe {asm!("" :::: "volatile")} `) if uninhabited values are in scope.
[...]

PR #88826, opened 2021-09-10
Turn `TrapUnreachable` off by default

jonas-schievink:
Now that https://github.com/rust-lang/rust/issues/28728 is fixed, there's no longer a risk of programs executing past the end of a function, so we don't have to put a trap instruction there anymore.

This should ever so slightly shrink the average Rust program.

(the second commit is a drive-by-fix that reverts an incorrect part of https://github.com/rust-lang/rust/commit/62e7331bd2bd592e30acf63467f140084684bc8f, which made the `alloc-optimisation.rs` test fail with my system's LLVM)

### DIFF
--- a/compiler/rustc_target/src/spec/mod.rs
+++ b/compiler/rustc_target/src/spec/mod.rs
@@ -1421,7 +1421,7 @@ impl Default for TargetOptions {
             stack_probes: StackProbeType::None,
             min_global_align: None,
             default_codegen_units: None,
-            trap_unreachable: true,
+            trap_unreachable: false,
             requires_lto: false,
             singlethread: false,
             no_builtins: false,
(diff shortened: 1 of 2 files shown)


nikic:
FWIW upstream is considering turning on TrapUnreachable or some variant thereof on by default, to at least prevent zero-size functions.

I think a more conservative step we can take is to enable the NoTrapAfterNoreturn option, which probably is the form in which the vast majority of unreachable appears in rust code (basically after every panic), and also the one where it has little practical benefit.

oli-obk:
I don't know enough about the effects of this to review this properly.

r? @nikic

JohnCSimon:
Ping from triage: 
@jonas-schievink  Can you please post the status of this PR?

PR #88830, opened 2021-09-10
Add help message for E0463

### LINKED ISSUE #87871: Better diagnostic for missing components
When using the `rustc_private` feature without the components being installed and you use a compiler crate, you'll get an error like this:
```
can't find crate for `rustc_metadata`
```
which is not helpful.

Given the following code:

```rust
#![feature(rustc_private)]

extern crate rustc_metadata;

fn main() {
}
```

The current output is:

```
error[E0463]: can't find crate for `rustc_metadata`
 --> src/main.rs:4:1
  |
4 | extern crate rustc_metadata;
  | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^ can't find crate

For more information about this error, try `rustc --explain E0463`.
error: could not compile `metadata-reader` due to previous error
```

```
error[E0463]: can't find crate for `rustc_metadata`
 --> src/main.rs:4:1
  |
4 | extern crate rustc_metadata;
  | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^ can't find crate

For more information about this error, try `rustc --explain E0463`.
error: could not compile `metadata-reader` due to previous error
help: maybe you need to install the following components: rust-src rustc-dev llvm-tools-preview
```

(the last line was added, specifying what could be done to fix the issue)

cc @GuillaumeGomez

GuillaumeGomez:
Fixes #87871.

r? @estebank

### DIFF
--- a/compiler/rustc_metadata/src/locator.rs
+++ b/compiler/rustc_metadata/src/locator.rs
@@ -1095,6 +1095,11 @@ impl CrateError {
                         == Symbol::intern(&sess.opts.debugging_opts.profiler_runtime)
                     {
                         err.note(&"the compiler may have been built without the profiler runtime");
+                    } else if crate_name.as_str().starts_with("rustc_") {
+                        err.help(
+                            "maybe you need to install the missing components with: \
+                             `rustup component add rust-src rustc-dev llvm-tools-preview`",
+                        );
                     }
                     err.span_label(span, "can't find crate");
                     err


compiler/rustc_metadata/src/locator.rs
```diff
@@ -1095,6 +1095,11 @@ impl CrateError {
                         == Symbol::intern(&sess.opts.debugging_opts.profiler_runtime)
                     {
                         err.note(&"the compiler may have been built without the profiler runtime");
+                    } else if crate_name.as_str().starts_with("rustc_") {
+                        err.help(
+                            "maybe you need to install the following components: \
+                             `rust-src rustc-dev llvm-tools-preview`",
```
estebank, line 1101: Can we include the `rustup` command needed for these?

Either way, r=me.

Issue #88756, opened 2021-09-08
rustdoc: `rustdoc -W help` prints error unless a file argument is given


lilyball:
Compiler lints can be printed with `rustc -W help`, with no filename arguments given. Unfortunately this doesn't work the same way for `rustdoc`:

```console
❯ rustdoc -W help
error: missing file operand
```

If I pass it an arbitrary file, it will print any errors produced by parsing the file as rust. This includes errors from using `async` without passing `--edition 2018`. This part is actually the same as `rustc -W help`, in that passing arbitrary files to that produces the same behavior.

If I pass it a file that parses cleanly (either 2015 rust, or 2018 rust with `--edition 2018` specified) then and only then will `rustdoc -W help` actually print the lint groups. In this way it behaves similarly to `rustc -W help $filename` in that `rustc` will also print the lint groups when passed a filename that parses cleanly. The difference is that `rustc -W help` doesn't require a filename at all.

The simplest workaround here is to type `rustdoc -W help /dev/null` as the empty `/dev/null` file does not trigger any parse errors.

## Potential root cause

It's possible that the root cause here is that calling `rustdoc` with no args produces the same error:

```console
❯ rustdoc
error: missing file operand
```

Compare this with `rustc`, which prints the help:

```console
❯ rustc
Usage: rustc [OPTIONS] INPUT

Options:
    -h, --help          Display this message
        --cfg SPEC      Configure the compilation environment
    -L [KIND=]PATH      Add a directory to the library search path. The
︙
```

This suggests that `rustdoc` is checking for filename arguments and erroring out early on, whereas `rustc` skips that check and then later interprets "no filename args" as "fallback to help".

### Meta
```console
❯ rustdoc --version --verbose
rustdoc 1.54.0 (a178d0322 2021-07-26)
binary: rustdoc
commit-hash: a178d0322ce20e33eac124758e837cbd80a6f633
commit-date: 2021-07-26
host: x86_64-apple-darwin
release: 1.54.0
LLVM version: 12.0.1
❯ rustdoc +nightly --version --verbose
rustdoc 1.57.0-nightly (fdf65053e 2021-09-07)
binary: rustdoc
commit-hash: fdf65053e99e8966f9bd83b5a8491326cb33d638
commit-date: 2021-09-07
host: x86_64-apple-darwin
release: 1.57.0-nightly
LLVM version: 13.0.0
```

jyn514:
[...]

PR #88831, opened 2021-09-10
Fix `rustdoc` argument error

inashivb:
This is the fix is corresponding to https://github.com/rust-lang/rust/issues/88756 mentored by @jyn514

### DIFF
--- a/src/librustdoc/config.rs
+++ b/src/librustdoc/config.rs
@@ -6,6 +6,7 @@ use std::path::PathBuf;
 use std::str::FromStr;
 
 use rustc_data_structures::fx::FxHashMap;
+use rustc_driver::print_flag_list;
 use rustc_session::config::{
     self, parse_crate_types_from_list, parse_externs, parse_target_triple, CrateType,
 };
@@ -310,11 +311,12 @@ impl RenderOptions {
 impl Options {
     /// Parses the given command-line for options. If an error message or other early-return has
     /// been printed, returns `Err` with the exit code.
-    crate fn from_matches(matches: &getopts::Matches) -> Result<Options, i32> {
+    crate fn from_matches(matches: &getopts::Matches, args: Vec<String>) -> Result<Options, i32> {
+        let args = &args[1..];
         // Check for unstable options.
         nightly_options::check_nightly_options(matches, &opts());
 
-        if matches.opt_present("h") || matches.opt_present("help") {
+        if args.is_empty() || matches.opt_present("h") || matches.opt_present("help") {
             crate::usage("rustdoc");
             return Err(0);
         } else if matches.opt_present("version") {
@@ -335,6 +337,21 @@ impl Options {
         // check for deprecated options
         check_deprecated_options(matches, &diag);
 
+        let z_flags = matches.opt_strs("Z");
+        if z_flags.iter().any(|x| *x == "help") {
+            print_flag_list("-Z", config::DB_OPTIONS);
+            return Err(0);
+        }
+        let c_flags = matches.opt_strs("C");
+        if c_flags.iter().any(|x| *x == "help") {
+            print_flag_list("-C", config::CG_OPTIONS);
+            return Err(0);
+        }
+        let w_flags = matches.opt_strs("W");
+        if w_flags.iter().any(|x| *x == "help") {
+            print_flag_list("-W", config::DB_OPTIONS);
+            return Err(0);
+        }
         if matches.opt_strs("passes") == ["list"] {
             println!("Available passes for running rustdoc:");
             for pass in passes::PASSES {
@@ -415,6 +432,7 @@ impl Options {
             }
             return Err(0);
         }
+        let (_lint_opts, _describe_lints, _lint_cap) = get_cmd_lint_options(matches, error_format);
 
         if matches.free.is_empty() {
             diag.struct_err("missing file operand").emit();
--- a/src/test/run-make/issue-88756-default-output/Makefile
+++ b/src/test/run-make/issue-88756-default-output/Makefile
@@ -1,0 +1,4 @@
+-include ../../run-make-fulldeps/tools.mk
+
+all:
+	$(BARE_RUSTDOC) 2>&1 | diff - output-default.stdout
--- a/compiler/rustc_driver/src/lib.rs
+++ b/compiler/rustc_driver/src/lib.rs
@@ -934,7 +934,7 @@ fn describe_codegen_flags() {
     print_flag_list("-C", config::CG_OPTIONS);
 }
 
-fn print_flag_list<T>(
+pub fn print_flag_list<T>(
     cmdline_opt: &str,
     flag_list: &[(&'static str, T, &'static str, &'static str)],
 ) {
--- a/src/librustdoc/lib.rs
+++ b/src/librustdoc/lib.rs
@@ -691,7 +691,7 @@ fn main_args(at_args: &[String]) -> MainResult {
 
     // Note that we discard any distinction between different non-zero exit
     // codes from `from_matches` here.
-    let options = match config::Options::from_matches(&matches) {
+    let options = match config::Options::from_matches(&matches, args) {
         Ok(opts) => opts,
         Err(code) => {
             return if code == 0 {
--- a/src/test/run-make/issue-88756-default-output/README.md
+++ b/src/test/run-make/issue-88756-default-output/README.md
@@ -1,0 +1,1 @@
+This is a test to verify that the default behavior of `rustdoc` is printing out help output instead of erroring out (#88756).
--- a/src/test/run-make/issue-88756-default-output/output-default.stdout
+++ b/src/test/run-make/issue-88756-default-output/output-default.stdout
@@ -1,0 +1,193 @@
+rustdoc [options] <input>
+
+Options:
+    -h, --help          show this help message
+    -V, --version       print rustdoc's version
+    -v, --verbose       use verbose output
+    -w, --output-format [html]
+                        the output type to write
+        --output PATH   Which directory to place the output. This option is
+                        deprecated, use --out-dir instead.
+    -o, --out-dir PATH  which directory to place the output
+        --crate-name NAME
+                        specify the name of this crate
+        --crate-type [bin|lib|rlib|dylib|cdylib|staticlib|proc-macro]
+                        Comma separated list of types of crates
+                        for the compiler to emit
+    -L, --library-path DIR
+                        directory to add to crate search path
+        --cfg           pass a --cfg to rustc
+        --check-cfg     pass a --check-cfg to rustc
+        --extern NAME[=PATH]
+                        pass an --extern to rustc
+        --extern-html-root-url NAME=URL
+                        base URL to use for dependencies; for example,
+                        "std=/doc" links std::vec::Vec to
+                        /doc/std/vec/struct.Vec.html
+        --extern-html-root-takes-precedence 
+                        give precedence to `--extern-html-root-url`, not
+                        `html_root_url`
+    -C, --codegen OPT[=VALUE]
+                        pass a codegen option to rustc
+        --document-private-items 
+                        document private items
+        --document-hidden-items 
+                        document items that have doc(hidden)
+        --test          run code examples as tests
+        --test-args ARGS
+                        arguments to pass to the test runner
+        --test-run-directory PATH
+                        The working directory in which to run tests
+        --target TRIPLE target triple to document
+        --markdown-css FILES
+                        CSS files to include via <link> in a rendered Markdown
+                        file
+        --html-in-header FILES
+                        files to include inline in the <head> section of a
+                        rendered Markdown file or generated documentation
+        --html-before-content FILES
+                        files to include inline between <body> and the content
+                        of a rendered Markdown file or generated documentation
+        --html-after-content FILES
+                        files to include inline between the content and
+                        </body> of a rendered Markdown file or generated
+                        documentation
+        --markdown-before-content FILES
+                        files to include inline between <body> and the content
+                        of a rendered Markdown file or generated documentation
+        --markdown-after-content FILES
+                        files to include inline between the content and
+                        </body> of a rendered Markdown file or generated
+                        documentation
+        --markdown-playground-url URL
+                        URL to send code snippets to
+        --markdown-no-toc 
+                        don't include table of contents
+    -e, --extend-css PATH
+                        To add some CSS rules with a given file to generate
+                        doc with your own theme. However, your theme might
+                        break if the rustdoc's generated HTML changes, so be
+                        careful!
+    -Z FLAG             internal and debugging options (only on nightly build)
+        --sysroot PATH  Override the system root
+        --playground-url URL
+                        URL to send code snippets to, may be reset by
+                        --markdown-playground-url or
+                        `#![doc(html_playground_url=...)]`
+        --display-doctest-warnings 
+                        show warnings that originate in doctests
+        --crate-version VERSION
+                        crate version to print into documentation
+        --sort-modules-by-appearance 
+                        sort modules by where they appear in the program,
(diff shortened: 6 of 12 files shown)


src/librustdoc/config.rs
```diff
@@ -417,7 +417,12 @@ impl Options {
             }
             return Err(0);
         }
+        let (lint_opts, describe_lints, lint_cap) = get_cmd_lint_options(matches, error_format);
 
+        if lint_opts.is_empty() {
+            crate::usage("rustdoc");
+            return Err(0);
+        }
```
jyn514, line 425: What does this do? Can you add a test showing the behavior?

jyn514 (commented):
Please add a test for this - it would go somewhere in `run-make`, there should be existing rustdoc tests there you can use as examples.

GuillaumeGomez:
Can you add a test in `src/test/rustdoc-ui` where you give no arguments to rustdoc please?

src/librustdoc/config.rs
inashivb: Hi @jyn514 !
I used this because `rustdoc` without any arguments and `rustdoc` with ambiguous arguments to an argument e.g. `rustdoc -W isrgnlk` had `matches.free` as an empty list. I looked for other params to differentiate between these two cases so that we print the default help in case no args were passed and error out in the latter case. I found that `lint_opts` was empty in case of no args and it was not empty when there was garbled output.

Do you think this is the right behavior for the two cases?

Note: Thanks a lot for the prompt review! :) I'm looking at how to add tests and keeping a check on any other feedback that you or GuillaumeGomez might have.

inashivb:
> Can you add a test in `src/test/rustdoc-ui` where you give no arguments to rustdoc please?

Thank you for the prompt review! :)
I'm observing many existing tests failing on `x86_64-gnu-llvm-10, ubuntu-latest-xl` which were not really failing on my system, could you please tell me how should I investigate these?

GuillaumeGomez:
Just in case you don't know how to run tests:

```console
$ x.py test src/test/rustdoc --stage 1
$ x.py test src/test/rustdoc-ui --stage 1
```

If one of the two doesn't work, it's very likely your change which broke them.

inashivb:
> Just in case you don't know how to run tests:
> 
> ```
> $ x.py test src/test/rustdoc --stage 1
> $ x.py test src/test/rustdoc-ui --stage 1
> ```
> 
> If one of the two doesn't work, it's very likely your change which broke them.

Thank you. I indeed did not test the right way. I just ran `x.py test` and that seemed to work. I seem to have broken many things. I'll work on resolving them.

src/test/rustdoc-ui/issue-88756-default-output.rs
```diff
@@ -0,0 +1,2 @@
+// compile-flags: -h
```
inashivb, line 1: Hi @GuillaumeGomez !
I'm unable to figure out how to make the test for no f;gas. I'm not sure even if this `-h` is correct here as the key says `compile-flags` but it did give the same results as `rustdoc -h`. Could you please point me to any existing similar tests?

GuillaumeGomez: That is actually a good question since the current file name is passed as argument... I can't find documentation about this either.

@jyn514: Do you know if there is a way to prevent any argument to be passed?

jyn514: You may need a run-make test instead of a UI test.

inashivb: Hi @jyn514 and @GuillaumeGomez !
I added the run-make test, could you please check if I did things the right way?

GuillaumeGomez (approved):

GuillaumeGomez:
Looks good to me, thanks! Wanna take a last look @jyn514 ?

src/test/run-make/issue-88756-default-output/Makefile
```diff
@@ -0,0 +1,4 @@
+-include ../../run-make-fulldeps/tools.mk
+
+all:
+	$(RUSTDOC) 2>&1 | diff - output-default.stdout
```
jyn514, line 4: This is not testing the right thing. The original issue was that -W help didn't show the lint help.

src/librustdoc/config.rs
```diff
@@ -417,7 +417,12 @@ impl Options {
             }
             return Err(0);
         }
+        let (lint_opts, describe_lints, lint_cap) = get_cmd_lint_options(matches, error_format);
 
+        if matches.free.is_empty() && lint_opts.is_empty() {
+            crate::usage("rustdoc");
+            return Err(0);
```
jyn514, line 424: Why was this change necessary? I'm pretty sure it had this behavior even before.

jyn514 (changes requested):

src/librustdoc/config.rs
inashivb: Hi @jyn514 !
Apologies for late response, my dev env was broken.
If I just move the `get_cmd_lint_options`, I still get:
```
└─ $ ▶ ./build/x86_64-unknown-linux-gnu/stage1/bin/rustdoc 
error: missing file operand
```
So, I checked for the `matches` attributes as the condition leading to this message was checking whether `matches.free` was empty. I found out that `matches.free` was indeed empty even after this move but what was also different with this situation compared to other places where `matches.free` is expected to error out was that `lint_opts` was also empty.

Before this change,
```
└─ $ ▶ ./build/x86_64-unknown-linux-gnu/stage1/bin/rustdoc -W help
error: missing file operand
```
and 
```
└─ $ ▶ ./build/x86_64-unknown-linux-gnu/stage1/bin/rustdoc -W
error: Argument to option 'W' missing
```

After adding these condition checks,
```
└─ $ ▶ ./build/x86_64-unknown-linux-gnu/stage1/bin/rustdoc -W help
rustdoc [options] <input>

Options:
    -h, --help          show this help message
    -V, --version       print rustdoc's version
    -v, --verbose       use verbose output
    -r, --input-format [rust]
                        the input type of the specified file
    -w, --output-format [html]
                        the output type to write
    -o, --output PATH   where to place the output
        --crate-name NAME
                        specify the name of this crate
        --crate-type [bin|lib|rlib|dylib|cdylib|staticlib|proc-macro]
                        Comma separated list of types of crates
                        for the compiler to emit
    -L, --library-path DIR
                        directory to add to crate search path
        --cfg           pass a --cfg to rustc
        --extern NAME[=PATH]
```
and
```
└─ $ ▶ ./build/x86_64-unknown-linux-gnu/stage1/bin/rustdoc -W
error: Argument to option 'W' missing
```
Is it still incorrect behavior? I'm not sure what changed from before if it was working with just the move.

src/test/run-make/issue-88756-default-output/Makefile
inashivb: OK. Shall I add test for default behavior and `-W help` as both seem to have been mentioned?

src/librustdoc/config.rs
jyn514: > Is it still incorrect behavior?

You can test by checking against rustc :) rustdoc and rustc should behave the same for invalid arguments.

(I haven't checked myself, I can do so if you have trouble.)

inashivb: hmm I am a little confused now. It seems as if rustdoc already (i.e. without this patch) works like rustc.
For example:
```
└─ $ ▶ ./build/x86_64-unknown-linux-gnu/stage1/bin/rustc -L help
error: no input filename given
```
It does not however behave this way particularly for 4 options (out of which one is `-W`) which are mentioned in the "additional help section"
```
Additional help:
    -C help             Print codegen options
    -W help             Print 'lint' options and default settings
    -Z help             Print unstable compiler options
    --help -v           Print the full set of options rustc accepts
```
So, if I were to pass `help` as mentioned in the cmd output above with either of these 4 options, I'll get some help corresponding to that option, e.g.
```
└─ $ ▶ ./build/x86_64-unknown-linux-gnu/stage1/bin/rustc -W help

Available lint options:
    -W <foo>           Warn about <foo>
    -A <foo>           Allow <foo>
    -D <foo>           Deny <foo>
    -F <foo>           Forbid <foo> (deny <foo> and all attempts to override)


Lint checks provided by rustc:

                                                       name  default  meaning
                                                       ----  -------  -------
                     absolute-paths-not-starting-with-crate  allow    fully qualified paths that start with a module name instead of `crate`, `self`, or an extern crate name
                                               box-pointers  allow    use of owned (Box type) heap memory
                                       deprecated-in-future  allow    detects use of items that will be deprecated in a future version
                                  elided-lifetimes-in-paths  allow    hidden lifetime parameters in types are deprecated
```
But, it seems that rustdoc does not promise any additional help or has a separate document for any of its options. Also, `-W` seems to mean different things for rustc and rustdoc.

The only difference that I am able to see is that rustdoc by default also errors out
```
─ $ ▶ ./build/x86_64-unknown-linux-gnu/stage1/bin/rustdoc 
error: missing file operand
```
while rustc prints help
```
└─ $ ▶ ./build/x86_64-unknown-linux-gnu/stage1/bin/rustc
Usage: rustc [OPTIONS] INPUT

Options:
    -h, --help          Display this message
        --cfg SPEC      Configure the compilation environment
    -L [KIND=]PATH      Add a directory to the library search path. The
                        optional KIND can be one of dependency, crate, native,
                        framework, or all (the default).
    -l [KIND[:MODIFIERS]=]NAME[:RENAME]
                        Link the generated crate(s) to the specified native
                        library NAME. The optional KIND can be one of
                        static, framework, or dylib (the default).
                        Optional comma separated MODIFIERS
                        (bundle|verbatim|whole-archive|as-needed)
                        may be specified each with a prefix of either '+' to
                        enable or '-' to disable.
        --crate-type [bin|lib|rlib|dylib|cdylib|staticlib|proc-macro]
                        Comma separated list of types of crates
                        for the compiler to emit
        --crate-name NAME
```

Did I do what you asked me to check, @jyn514 ? If yes, where does it make us stand on this current rustdoc issue (problem with `-W help`)?

jyn514: > Is it still incorrect behavior? I'm not sure what changed from before if it was working with just the move.

Yes, it's still correct - it should be showing the same lint help as rustc for `-W help`, not the generic usage.

> But, it seems that rustdoc does not promise any additional help or has a separate document for any of its options. Also, -W seems to mean different things for rustc and rustdoc.

That should not be the case. If rustdoc doesn't give help for `-W help`, that's also a bug that should be fixed.

> The only difference that I am able to see is that rustdoc by default also errors out while rustc prints help

We should fix that too.

> hmm I am a little confused now. It seems as if rustdoc already (i.e. without this patch) works like rustc (for the -W help like scenarios).

I'm not sure why you think `-L` and `-W` are similar. `rustc --help` says
```
    -L [KIND=]PATH      Add a directory to the library search path. The
                        optional KIND can be one of dependency, crate, native,
                        framework, or all (the default).
...

Additional help:
    -C help             Print codegen options
    -W help             Print 'lint' options and default settings
    -Z help             Print unstable compiler options
    --help -v           Print the full set of options rustc accepts
```
`-L help` is a valid option that means "add the directory named help/ to the search path".

jyn514:
Hi @inashivb do you know how to make progress here? Do you need help?

inashivb:
> Hi @inashivb do you know how to make progress here? Do you need help?

Hi @jyn514 ! Yes. Could you please tell where in rustc code can I find the cmdline arguments setting and rendering? I could not find any usage of  `crate::usage` or the help strings for any option.

jyn514:
@inashivb I think you're looking for https://doc.rust-lang.org/nightly/nightly-rustc/src/rustc_driver/lib.rs.html#985

inashivb:
@jyn514 apologies for the massive delay. Looking at it this weekend.

JohnCSimon:
Ping from triage: 
@inashivb What is the status of this PR? 
.Please adjust the label with `@rustbot ready` when this is ready for review.

inashivb:
@JohnCSimon it's not ready yet. Sorry about this. Looking at it today. Trying to match rustc's behavior here.

jyn514:
FYI I've been chatting with @inashivb in DM's, they're still working on this :) no need to triage for now.

jyn514:
I am no longer on the Rustdoc team. You can assign a new reviewer with

r? rust-lang/rustdoc

notriddle:
It looks like some new arguments were added, and the new output needs blessed.

Dylan-DPC:
@inashivb any updates on this?

inashivb:
> @inashivb any updates on this?

Hi @Dylan-DPC !

I'm afraid I'm not sure how to deal with the issue that is there now. I'll try to explain:
In order to show the default help output in case no argument was given to `rustdoc` on cmdline, I check for the length of the arguments given on the cmdline (following how `rustc` does it), if it was 0, we output default help else let the existing code take care of whatever arg was passed.
However, when I run `x.py test --stage 1 tests/../mytest` I observe that `rustdoc` is supplied with some args like `-L` by default. I pass no arguments to the test (https://github.com/rust-lang/rust/pull/88831/commits/69e6d4a7e936744e31b14c0c0ca47f28cab22f5a#diff-0fa1bfe4e7dc2832e4d253c4bd571dc4a223b0d5f874e2bf4e3a5ed6889b5fc8R4) and it still has at least the library in the args so the check I have for `args.is_empty()` fails.

In conclusion,
If I check the generated `rustdoc` binary in `stage1` after building this code, it works just as expected. I give it no args and it prints the default help output. But, I cannot seem to convert the same behavior to test because of mandatory extra args passed during test.

Could you please tell what's the way to go about this?

notriddle:
How about changing `$(RUSTDOC)` to `$(BARE_RUSTDOC)`?

inashivb:
> How about changing `$(RUSTDOC)` to `$(BARE_RUSTDOC)`?

@notriddle thank you very much! That works locally at least. Here I see a bunch of more options that come with `rustc -W help` but not with `rustdoc -W help` when I run locally. Do you know what I might be doing wrng now?

notriddle (approved):
It looks fine to me, but I admit I’m not an expert on this part of the code.

JohnCSimon:
triage:
@notriddle anyone else who can review this?

GuillaumeGomez (approved):

GuillaumeGomez:
Looks good to me as well. Thanks @inashivb !

@bors: r=GuillaumeGomez,notriddle

matthiaskrgr:
@bors r- 
looks like this ended up in a rollup  https://github.com/rust-lang/rust/pull/95288  while it failed already in its own ci

Dylan-DPC:
@inashivb if you can fix the CI failures, we can get this merged again

inashivb:
Hi @Dylan-DPC ! thanks for tagging me. I must have missed the notifications for change in labels. Could you please take a look at the branch now?

inashivb:
I see the cause of failure as
```
< More information available at https://doc.rust-lang.org/1.62.0/rustdoc/what-is-rustdoc.html
---
> More information available at https://doc.rust-lang.org/nightly/rustdoc/what-is-rustdoc.html
```

Is there a VERSION or something env var? I tried looking for version in small and caps but can't see rust specific version setting. Could you please assist here?

compiler-errors:
I did a [bors resync](https://github.com/rust-lang/rust/pull/94206#issuecomment-1120251623) and this seems to have gotten reapproved, sorry for the noise

@bors r-

inashivb:
> Yeah, it's called `$CHANNEL`

@notriddle could you please assist in how to use it? I cannot seem to make it work by setting it on Makefile or the `main.rs` file (that I saw some UI tests do) [not sure if I'm reading it right though 😓 ]

GuillaumeGomez:
I finished this PR in https://github.com/rust-lang/rust/pull/98331. 
Thanks @inashivb for working on this!

Issue #88793, opened 2021-09-09
Missing MIR optimization: Replace matches with loads if possible


pcwalton:
Many times, people write matches on enums that are really loads from a table. It'd be nice if we could codegen them as such.

Here's an example, at https://godbolt.org/z/4P9v61an7:

```rust
pub enum Char {
    A,
    B,
    C,
    D,
    E,
    F,
    G,
    H,
    I,
    J,
    K,
    L,
}

pub fn to_str(val: Char) -> &'static str {
    match val {
        Char::A => "A",
        Char::B => "B",
        Char::C => "C",
        Char::D => "D",
        Char::E => "E",
        Char::F => "F",
        Char::G => "G",
        Char::H => "H",
        Char::I => "I",
        Char::J => "J",
        Char::K => "K",
        Char::L => "L",
    }
}
```

The codegen here has a lot to be desired:

```asm
example::to_str:
        lea     rax, [rip + .L__unnamed_1]
        movzx   ecx, dil
        lea     rdx, [rip + .LJTI0_0]
        movsxd  rcx, dword ptr [rdx + 4*rcx]
        add     rcx, rdx
        jmp     rcx
.LBB0_1:
        lea     rax, [rip + .L__unnamed_2]
        mov     edx, 1
        ret
.LBB0_2:
        lea     rax, [rip + .L__unnamed_3]
        mov     edx, 1
        ret
.LBB0_3:
        lea     rax, [rip + .L__unnamed_4]
        mov     edx, 1
        ret
.LBB0_4:
        lea     rax, [rip + .L__unnamed_5]
        mov     edx, 1
        ret
.LBB0_5:
        lea     rax, [rip + .L__unnamed_6]
        mov     edx, 1
        ret
.LBB0_6:
        lea     rax, [rip + .L__unnamed_7]
        mov     edx, 1
        ret
.LBB0_7:
        lea     rax, [rip + .L__unnamed_8]
        mov     edx, 1
        ret
.LBB0_8:
        lea     rax, [rip + .L__unnamed_9]
        mov     edx, 1
        ret
.LBB0_9:
        lea     rax, [rip + .L__unnamed_10]
        mov     edx, 1
        ret
.LBB0_10:
        lea     rax, [rip + .L__unnamed_11]
        mov     edx, 1
        ret
.LBB0_11:
        lea     rax, [rip + .L__unnamed_12]
.LBB0_12:
        mov     edx, 1
        ret
.LJTI0_0:
[...]

PR #88832, opened 2021-09-10
Introduce a fast path that avoids the `debug_tuple` abstraction when deriving Debug for unit-like enum variants.

pcwalton:
The intent here is to allow LLVM to remove the switch entirely in favor of an
indexed load from a table of constant strings, which is likely what the
programmer would write in C. Unfortunately, LLVM currently doesn't perform this
optimization due to a bug, but there is [a
patch](https://reviews.llvm.org/D109565) that fixes this issue. I've verified
that, with that patch applied on top of this commit, Debug for unit-like tuple
variants becomes a load, reducing the O(n) code bloat to O(1).

Note that inlining `DebugTuple::finish()` wasn't enough to allow LLVM to
optimize the code properly; I had to avoid the abstraction entirely. Not using
the abstraction is likely better for compile time anyway.

Part of #88793.

r? @oli-obk

### DIFF
--- a/compiler/rustc_builtin_macros/src/deriving/debug.rs
+++ b/compiler/rustc_builtin_macros/src/deriving/debug.rs
@@ -65,15 +65,29 @@ fn show_substructure(cx: &mut ExtCtxt<'_>, span: Span, substr: &Substructure<'_>
     // We want to make sure we have the ctxt set so that we can use unstable methods
     let span = cx.with_def_site_ctxt(span);
     let name = cx.expr_lit(span, ast::LitKind::Str(ident.name, ast::StrStyle::Cooked));
+    let fmt = substr.nonself_args[0].clone();
+
+    // Special fast path for unit variants. In the common case of an enum that is entirely unit
+    // variants (i.e. a C-like enum), this fast path allows LLVM to eliminate the entire switch in
+    // favor of a lookup table.
+    if let ast::VariantData::Unit(..) = vdata {
+        let fn_path_write_str = cx.std_path(&[sym::fmt, sym::Formatter, sym::write_str]);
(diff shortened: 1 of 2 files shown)


pcwalton:
If anyone else wants to review, feel free.

oli-obk:
@bors try @rust-timer queue

Since we're trying to make LLVM behave here, I think we should have a codegen test that checks that the generated code is indeed match-free

oli-obk:
I can't tell where the keccak regression is coming from, it doesn't have or use any debug impls. So I'm assuming it's changes in rustc possibly due to changed inlining.

pcwalton:
> @bors try @rust-timer queue
> 
> Since we're trying to make LLVM behave here, I think we should have a codegen test that checks that the generated code is indeed match-free

Unfortunately that's not possible right now because https://reviews.llvm.org/D109565 hasn't landed yet.

oli-obk:
@bors r+

we're keeping the issue open, I'm going to mark it as needs-test and blocked

ehuss:
@bors retry

#88924

Mark-Simulacrum:
> I can't tell where the keccak regression is coming from, it doesn't have or use any debug impls. So I'm assuming it's changes in rustc possibly due to changed inlining.

@oli-obk I think we probably want a little more investigation on perf regressions before approving PRs. In particular, in this case, I would say that the instruction count regressions are justified because the same benchmarks do not regress in cycles, so the instruction count regressions are not representative of a real performance loss. (They're just a proxy for potential performance losses). It'd be good to justify regressions and mark them as "triaged" before approving a PR if the performance loss is noted in a pre-merge perf run.

PR #88834, opened 2021-09-10
optimize str::from_utf8() validation when slice contains multibyte chars and str.chars().count() in all cases

the8472:
The change shows small but consistent improvements across several x86 target feature levels. I also tried to optimize counting with `slice.as_chunks` but that yielded more inconsistent results, bigger improvements for some optimization levels, lesser ones in others.

```
old, -O2, x86-64
test str::str_char_count_emoji                                  ... bench:       1,924 ns/iter (+/- 26)
test str::str_char_count_lorem                                  ... bench:         879 ns/iter (+/- 12)
test str::str_char_count_lorem_short                            ... bench:           5 ns/iter (+/- 0)

new, -O2, x86-64
test str::str_char_count_emoji                                  ... bench:       1,878 ns/iter (+/- 21)
test str::str_char_count_lorem                                  ... bench:         851 ns/iter (+/- 11)
test str::str_char_count_lorem_short                            ... bench:           4 ns/iter (+/- 0)

old, -O2, x86-64-v2
test str::str_char_count_emoji                                  ... bench:       1,477 ns/iter (+/- 46)
test str::str_char_count_lorem                                  ... bench:         675 ns/iter (+/- 15)
test str::str_char_count_lorem_short                            ... bench:           5 ns/iter (+/- 0)

new, -O2, x86-64-v2
test str::str_char_count_emoji                                  ... bench:       1,323 ns/iter (+/- 39)
test str::str_char_count_lorem                                  ... bench:         593 ns/iter (+/- 18)
test str::str_char_count_lorem_short                            ... bench:           4 ns/iter (+/- 0)

old, -O2, x86-64-v3
test str::str_char_count_emoji                                  ... bench:         748 ns/iter (+/- 7)
test str::str_char_count_lorem                                  ... bench:         348 ns/iter (+/- 2)
test str::str_char_count_lorem_short                            ... bench:           5 ns/iter (+/- 0)

new, -O2, x86-64-v3
test str::str_char_count_emoji                                  ... bench:         650 ns/iter (+/- 4)
test str::str_char_count_lorem                                  ... bench:         301 ns/iter (+/- 1)
test str::str_char_count_lorem_short                            ... bench:           5 ns/iter (+/- 0)
```

and for the multibyte-char string validation:

```
old, -O2, x86-64
test str::str_validate_emoji                                    ... bench:       4,606 ns/iter (+/- 64)

new, -O2, x86-64
test str::str_validate_emoji                                    ... bench:       3,837 ns/iter (+/- 60)
```


the8472:
perf run since since `utf8_is_cont_byte` is also used in other places and I only benched `str.chars().count()`

@bors try @rust-timer queue

library/core/src/str/validations.rs
```diff
@@ -22,7 +22,7 @@ fn utf8_acc_cont_byte(ch: u32, byte: u8) -> u32 {
 /// bits `10`).
 #[inline]
 pub(super) fn utf8_is_cont_byte(byte: u8) -> bool {
-    (byte & !CONT_MASK) == TAG_CONT_U8
+    byte.wrapping_sub(128) < 64
```
falk-hueffner, line 25: Even shorter: `(byte as i8) < -64` (although it seems llvm does this transformation by itself).

the8472: might as well make it more obvious how this needs fewer arithmetic ops

the8472:
let's see if it was due to the extra function calls even though they should be inlined.

@bors try @rust-timer queue

the8472:
Several of the improved and regressed benchmarks spend less/more time in `LLVM_passes`, `finish_ongoing_codegen` and `run_linker` so my guess is that this is mostly llvm noise rather than rust code being affected.

joshtriplett:
I think the microbenchmark results seem clear here, and the rest seems likely to be noise. On balance, this seems likely to be a win.

@bors r+

PR #88835, opened 2021-09-10
Fix error recovery in format macro parsing

### LINKED ISSUE #88770: 'called `Result::unwrap_err()` on an `Ok` value: false', compiler/rustc_builtin_macros/src/format.rs:173:51
I'm seeing an internal compiler error on the following input, found by [fuzz-rustc](https://github.com/dwrensha/fuzz-rustc):

```rust
fn m(){print!("",(c for&g
u
e
e
```

```
$ rustc bug.rs
error: this file contains an unclosed delimiter
 --> bug.rs:4:2
  |
1 | fn m(){print!("",(c for&g
  |       -      -   - unclosed delimiter
  |       |      |
  |       |      unclosed delimiter
  |       unclosed delimiter
...
4 | e
  |  ^

error: missing `in` in `for` loop
 --> bug.rs:1:26
  |
1 |   fn m(){print!("",(c for&g
  |  __________________________^
2 | | u
  | |_ help: try adding `in` here

error: expected one of `!`, `)`, `,`, `.`, `::`, `?`, `{`, or an operator, found keyword `for`
 --> bug.rs:1:21
  |
1 | fn m(){print!("",(c for&g
  |                     ^^^ expected one of 8 possible tokens

error: expected `;`, found `e`
 --> bug.rs:3:2
  |
3 | e
  |  ^ help: add `;` here
4 | e
  | - unexpected token

thread 'rustc' panicked at 'called `Result::unwrap_err()` on an `Ok` value: false', compiler/rustc_builtin_macros/src/format.rs:173:51

note: rustc 1.57.0-nightly (fdf65053e 2021-09-07) running on x86_64-unknown-linux-gnu

query stack during panic:
end of query stack
error: aborting due to 4 previous errors
```

FabianWolff:
Fixes #88770. Basically, the assumption in the following comment is incorrect:
https://github.com/rust-lang/rust/blob/b69fe57261086e70aea9d5b58819a1794bf7c121/compiler/rustc_builtin_macros/src/format.rs#L167-L172

This is only true in the first iteration of the loop, when [`p.clear_expected_tokens()`](https://github.com/rust-lang/rust/blob/b69fe57261086e70aea9d5b58819a1794bf7c121/compiler/rustc_builtin_macros/src/format.rs#L164) is called. In subsequent iterations, `p.expected_tokens` won't be empty, so `p.expect()` won't actually call `unexpected_try_recover()`:
https://github.com/rust-lang/rust/blob/b69fe57261086e70aea9d5b58819a1794bf7c121/compiler/rustc_parse/src/parser/mod.rs#L487-L498

Instead, it will call `expect_one_of()`, which _can_ recover and return `Ok()`. This PR handles this case to fix the ICE in #88770.

### DIFF
--- a/compiler/rustc_builtin_macros/src/format.rs
+++ b/compiler/rustc_builtin_macros/src/format.rs
@@ -164,23 +164,22 @@ fn parse_args<'a>(
                 p.clear_expected_tokens();
             }
 
-            // `Parser::expect` tries to recover using the
-            // `Parser::unexpected_try_recover` function. This function is able
-            // to recover if the expected token is a closing delimiter.
-            //
-            // As `,` is not a closing delimiter, it will always return an `Err`
-            // variant.
-            let mut err = p.expect(&token::Comma).unwrap_err();
-
-            match token::TokenKind::Comma.similar_tokens() {
-                Some(tks) if tks.contains(&p.token.kind) => {
-                    // If a similar token is found, then it may be a typo. We
-                    // consider it as a comma, and continue parsing.
-                    err.emit();
-                    p.bump();
+            match p.expect(&token::Comma) {
+                Err(mut err) => {
+                    match token::TokenKind::Comma.similar_tokens() {
+                        Some(tks) if tks.contains(&p.token.kind) => {
+                            // If a similar token is found, then it may be a typo. We
+                            // consider it as a comma, and continue parsing.
+                            err.emit();
+                            p.bump();
+                        }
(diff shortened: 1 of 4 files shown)


FabianWolff:
It seems that `@rustbot` somehow passed over this PR and did not assign anybody. Therefore (feel free to reassign):

r? @petrochenkov

petrochenkov:
Thanks!
@bors r+

petrochenkov:
Looks like some infra issue.
@bors retry

PR #88837, opened 2021-09-10
Add index/iter methods for `MaybeUninit<[T; N]>`

kpp:
While it's possible to create a `MaybeUninit::<[T; N]>` there is no way to index elements to initialize them other than to call `.as_mut_ptr()` and index over a raw pointer. I added several methods to make life a little bit easier. The idea is to get `&[mut] MaybeUninit<T>` from `MaybeUninit<[T; N]>` either through index or iter methods.

## Use cases

```
#[test]
fn uninit_array_iter_mut() {
    let mut v = MaybeUninit::<[i32; 4]>::uninit();

    v.iter_mut().zip(1..=4).for_each(|(v, val)| {
        v.write(val);
    });

    let v = unsafe { v.assume_init() };
    assert_eq!(v, [1, 2, 4, 8]);
}
```

### DIFF
--- a/library/core/src/mem/maybe_uninit.rs
+++ b/library/core/src/mem/maybe_uninit.rs
@@ -2,7 +2,9 @@ use crate::any::type_name;
 use crate::fmt;
 use crate::intrinsics;
 use crate::mem::ManuallyDrop;
+use crate::ops::{Index, IndexMut};
 use crate::ptr;
+use crate::slice::IterMut;
 
 /// A wrapper type to construct uninitialized instances of `T`.
 ///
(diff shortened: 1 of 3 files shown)


the8472:
That seems redundant with [uninit_array](https://doc.rust-lang.org/std/mem/union.MaybeUninit.html#method.uninit_array)  and [array_assume_init](https://doc.rust-lang.org/std/mem/union.MaybeUninit.html#method.array_assume_init).

Maybe conversions between `[MaybeUninit<T>; N]` and `MaybeUninit<[T; N]>` should be added instead?

nbdd0121:
I agree with @the8472 that this is redundant, and `uninit_array` and `array_assume_init` should be used instead.

Plus, implementation of stable trait of stable type is insta-stable. The unstable annotation has no effect. Therefore this would need libs team signoff.

kpp:
Well, you are right. `::transpose()`?

nbdd0121:
What code would benefit from the transpose method?

kpp:
Same as assume_init_ref/mut. This can be useful when we want to access it but don’t have ownership of the MaybeUninit.

the8472:
No new method is needed, two `From` impls should do the job.

crlf0710:
@kpp Ping from triage, seems CI is still red here.

kpp:
I can't fight with `implementation has missing stability attribute` vs `this stability annotation is useless`. Would you please help me?

the8472:
The impl block needs a stability attribute, not the methods. But it must be `stable` since trait impls are insta-stable.

JohnCSimon:
Ping from triage: 
@kpp 
I'm closing this due to inactivity, please feel to reopen when you are ready to continue with this. Thank you.

Issue #88636, opened 2021-09-04
Incorrect hint for `decl_macro` import.


crlf0710:
In this program:
```rust
#![feature(decl_macro)]

mod foo1 {
    pub macro foo() {}
}

mod foo2 {
    pub macro bar() {
        foo!();
    }
   
}

mod foo3 {
    fn bar() {
        crate::foo2::bar!();
    }
    pub use crate::foo1::foo;
}
```
The compilation output is:
```text
error: cannot find macro `foo` in this scope
  --> src/lib.rs:9:9
   |
9  |         foo!();
   |         ^^^
...
16 |         crate::foo2::bar!();
   |         -------------------- in this macro invocation
   |
   = note: consider importing this macro:
           crate::foo3::foo
   = note: this error originates in the macro `crate::foo2::bar` (in Nightly builds, run with -Z macro-backtrace for more info)
```
However the hint is wrong - `crate::foo3::foo` is inaccessible to `foo2`. The correct hint should be suggesting `crate::foo1::foo`.

FabianWolff:
@rustbot claim

This is not quite fixed by #88838 yet, but I have a patch that should solve this. Since the patch is based on #88838, I'll wait until that PR is merged before opening a PR for this issue.

PR #88838, opened 2021-09-10
Do not suggest importing inaccessible items

### LINKED ISSUE #88472: Bad diagnostic when attemting to import private item via glob
Given the following code: 

```rust
mod a {
    struct Foo;
}

mod b {
    use crate::a::*;
    type Bar = Foo;
}
```

The current output is:

```
error[E0412]: cannot find type `Foo` in this scope
 --> src/lib.rs:7:16
  |
7 |     type Bar = Foo;
  |                ^^^ not found in this scope
  |
help: consider importing this struct
  |
6 |     use crate::a::Foo;
  |

warning: unused import: `crate::a::*`
 --> src/lib.rs:6:9
  |
6 |     use crate::a::*;
  |         ^^^^^^^^^^^
  |
```

```
error[E0412]: cannot find type `Foo` in this scope
 --> src/lib.rs:7:16
  |
7 |     type Bar = Foo;
  |                ^^^ not found in this scope
  |
Note: Foo imported [here], but cannot be imported as it is defined as private [here]
```

FabianWolff:
Fixes #88472. For this example:
```rust
mod a {
    struct Foo;
}

mod b {
    type Bar = Foo;
}
```
rustc currently emits:
```
error[E0412]: cannot find type `Foo` in this scope
 --> test.rs:6:16
  |
6 |     type Bar = Foo;
  |                ^^^ not found in this scope
  |
help: consider importing this struct
  |
6 |     use a::Foo;
  |
```
this is incorrect, as applying this suggestion leads to
```
error[E0603]: struct `Foo` is private
 --> test.rs:6:12
  |
6 |     use a::Foo;
  |            ^^^ private struct
  |
note: the struct `Foo` is defined here
 --> test.rs:2:5
  |
2 |     struct Foo;
  |     ^^^^^^^^^^^
```
With my changes, I get:
```
error[E0412]: cannot find type `Foo` in this scope
 --> test.rs:6:16
  |
6 |     type Bar = Foo;
  |                ^^^ not found in this scope
  |
  = note: this struct exists but is inaccessible:
          a::Foo
```
As for the wildcard mentioned in #88472, I would argue that the warning is actually correct, since the import _is_ unused. I think the real issue is the wrong suggestion, which I have fixed here.

### DIFF
--- a/src/test/ui/imports/glob-resolve1.stderr
+++ b/src/test/ui/imports/glob-resolve1.stderr
@@ -4,10 +4,8 @@ error[E0425]: cannot find function `fpriv` in this scope
 LL |     fpriv();
    |     ^^^^^ not found in this scope
    |
-help: consider importing this function
-   |
-LL | use bar::fpriv;
-   |
+   = note: this function exists but is inaccessible:
+           bar::fpriv
 
 error[E0425]: cannot find function `epriv` in this scope
   --> $DIR/glob-resolve1.rs:27:5
@@ -15,10 +13,8 @@ error[E0425]: cannot find function `epriv` in this scope
 LL |     epriv();
    |     ^^^^^ not found in this scope
    |
-help: consider importing this function
-   |
-LL | use bar::epriv;
-   |
+   = note: this function exists but is inaccessible:
+           bar::epriv
 
 error[E0423]: expected value, found enum `B`
   --> $DIR/glob-resolve1.rs:28:5
@@ -44,10 +40,8 @@ error[E0425]: cannot find value `C` in this scope
 LL |     C;
    |     ^ not found in this scope
    |
-help: consider importing this unit struct
-   |
-LL | use bar::C;
-   |
+   = note: this unit struct exists but is inaccessible:
+           bar::C
 
 error[E0425]: cannot find function `import` in this scope
   --> $DIR/glob-resolve1.rs:30:5
@@ -67,16 +61,10 @@ LL |     pub enum B {
    |     ---------- similarly named enum `B` defined here
 ...
 LL |     foo::<A>();
-   |           ^
+   |           ^ help: an enum with a similar name exists: `B`
    |
-help: an enum with a similar name exists
-   |
-LL |     foo::<B>();
-   |           ~
-help: consider importing this enum
-   |
-LL | use bar::A;
-   |
+   = note: this enum exists but is inaccessible:
+           bar::A
 
 error[E0412]: cannot find type `C` in this scope
   --> $DIR/glob-resolve1.rs:33:11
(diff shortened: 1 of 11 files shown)


JohnTitor:
Seems good but I'm not part of t-compiler, r? @estebank could you take a look?

src/test/ui/imports/glob-resolve1.stderr
```diff
@@ -4,21 +4,17 @@ error[E0425]: cannot find function `fpriv` in this scope
 LL |     fpriv();
    |     ^^^^^ not found in this scope
    |
-help: consider importing this function
-   |
-LL | use bar::fpriv;
-   |
+   = note: this function exists but is inaccessible:
+           bar::fpriv
```
estebank, line 8: For the case where there's a single function being mentioned, we should not present the "list" output, instead show "function `bar::fpriv` exists but is inaccessible".

src/test/ui/resolve/issue-88472.stderr
```diff
@@ -0,0 +1,34 @@
+error[E0412]: cannot find type `Foo` in this scope
+  --> $DIR/issue-88472.rs:14:16
+   |
+LL |     type Bar = Foo;
+   |                ^^^ not found in this scope
+   |
+   = note: this struct exists but is inaccessible:
+           a::Foo
```
estebank, line 8: We should use the `ImportSuggestion.did` along with `def_span` to show more verbose output:

```
note: struct `a::Foo` exists but is inacessible
  --> $DIR/issue-88472.rs:8:16
   |
LL |     struct Foo;
   |            ^^^ not accessible
```

src/test/ui/resolve/issue-88472.stderr
```diff
@@ -0,0 +1,34 @@
...
+
+error[E0412]: cannot find type `Eee` in this scope
+  --> $DIR/issue-88472.rs:29:16
+   |
+LL |     type Baz = Eee;
+   |                ^^^ not found in this scope
+   |
+   = note: these items exist but are inaccessible:
+           c::Eee
+           c::d::Eee
```
estebank, line 18: When there are multiple items, we can use `MultiSpan` and `push_span_label` to point at all the items' definition span.

estebank (approved):
I think the changes are an improvement already, but I have a few ideas on how we could improve them further. I'm r=me for the code changes as they are, but I would like to continue tweaking them so the output is nicer. Would you have time to make them *perfect*, @FabianWolff?

FabianWolff:
> I think the changes are an improvement already, but I have a few ideas on how we could improve them further. I'm r=me for the code changes as they are, but I would like to continue tweaking them so the output is nicer. Would you have time to make them _perfect_, @FabianWolff?

Thanks for the review! Yes, I'll try to implement your suggestions (though I won't get around to it today).

estebank:
One last thing: if the `DefId` belongs to the local crate, we should maybe suggest making the item (and all its parents, if applicable) `pub`? Anyone experienced will not need it and figure things out just with even the current message in this PR, but I could envision anyone just starting asking themselves "ok, so what do I do? how do I make it accessible?". Of course, that is not necessarily needed for this PR, but we don't address it, could you file a follow up ticket to do that?

FabianWolff:
@estebank I have now implemented your suggestions! I did not get around to your proposal in https://github.com/rust-lang/rust/pull/88838#issuecomment-921720642; instead, I have opened #89057 as a follow-up, as you have suggested (I might come back to it if I have time in the next few days).

estebank (approved):
r=me after rebasing

crlf0710:
I'm curious whether this will fix #88636 too

estebank:
@crlf0710 it's very likely it will, good catch!

FabianWolff:
@estebank Rebased & ready.

PR #88700, opened 2021-09-06
[experimental] Rvalue::InitBox prototype

nbdd0121:
r? @ghost

PR #88839, opened 2021-09-10
Introduce NullOp::AlignOf

nbdd0121:
This PR introduces `Rvalue::NullaryOp(NullOp::AlignOf, ty)`, which will be lowered from `align_of`, similar to `size_of` lowering to `Rvalue::NullaryOp(NullOp::SizeOf, ty)`.

The changes are originally part of #88700 but since it's not dependent on other changes and could have performance impact on its own, it's separated into its own PR.

### DIFF
--- a/compiler/rustc_codegen_cranelift/src/base.rs
+++ b/compiler/rustc_codegen_cranelift/src/base.rs
@@ -726,15 +726,20 @@ fn codegen_stmt<'tcx>(
                     let ptr = fx.bcx.inst_results(call)[0];
                     lval.write_cvalue(fx, CValue::by_val(ptr, box_layout));
                 }
-                Rvalue::NullaryOp(NullOp::SizeOf, ty) => {
+                Rvalue::NullaryOp(null_op, ty) => {
                     assert!(
                         lval.layout()
                             .ty
                             .is_sized(fx.tcx.at(stmt.source_info.span), ParamEnv::reveal_all())
                     );
-                    let ty_size = fx.layout_of(fx.monomorphize(ty)).size.bytes();
+                    let layout = fx.layout_of(fx.monomorphize(ty));
+                    let val = match null_op {
+                        NullOp::SizeOf => layout.size.bytes(),
+                        NullOp::AlignOf => layout.align.abi.bytes(),
+                        NullOp::Box => unreachable!(),
+                    };
                     let val =
-                        CValue::const_val(fx, fx.layout_of(fx.tcx.types.usize), ty_size.into());
+                        CValue::const_val(fx, fx.layout_of(fx.tcx.types.usize), val.into());
                     lval.write_cvalue(fx, val);
                 }
                 Rvalue::Aggregate(ref kind, ref operands) => match kind.as_ref() {
--- a/compiler/rustc_codegen_ssa/src/mir/rvalue.rs
+++ b/compiler/rustc_codegen_ssa/src/mir/rvalue.rs
@@ -487,20 +487,6 @@ impl<'a, 'tcx, Bx: BuilderMethods<'a, 'tcx>> FunctionCx<'a, 'tcx, Bx> {
                 )
             }
 
-            mir::Rvalue::NullaryOp(mir::NullOp::SizeOf, ty) => {
-                let ty = self.monomorphize(ty);
-                assert!(bx.cx().type_is_sized(ty));
-                let val = bx.cx().const_usize(bx.cx().layout_of(ty).size.bytes());
-                let tcx = self.cx.tcx();
-                (
-                    bx,
-                    OperandRef {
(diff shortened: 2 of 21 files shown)


compiler/rustc_codegen_cranelift/src/base.rs
```diff
@@ -737,6 +737,17 @@ fn codegen_stmt<'tcx>(
                         CValue::const_val(fx, fx.layout_of(fx.tcx.types.usize), ty_size.into());
                     lval.write_cvalue(fx, val);
                 }
+                Rvalue::NullaryOp(NullOp::AlignOf, ty) => {
```
oli-obk, line 740: Please merge all these arms with the `SizeOf` one by only matching on the `AlignOf`/`SizeOf` part for the actual value computation.

compiler/rustc_middle/src/mir/mod.rs
```diff
@@ -2278,6 +2278,8 @@ impl BinOp {
 pub enum NullOp {
     /// Returns the size of a value of that type
     SizeOf,
+    /// Returns the minimum alignment of a type
+    AlignOf,
```
bjorn3, line 2282: Maybe call this `MinAlignOf` to distinguish it from the preferred alignment?

nbdd0121: Given that preferred alignment is never used in the library and hardly used in the compiler (in fact the only usage of preferred alignment is in cg_clif), and the name we expose `min_align_of` to the user is `align_of`, I think the name `align_of/AlignOf` is not ambiguous. I doubt if we'll ever add a `PrefAlignOf` nullary operator.

bjorn3: Huh, indeed. Should we remove preferred alignment then if it isn't used anyway?

compiler/rustc_const_eval/src/interpret/step.rs
```diff
@@ -270,18 +270,23 @@ impl<'mir, 'tcx: 'mir, M: Machine<'mir, 'tcx>> InterpCx<'mir, 'tcx, M> {
                 M::box_alloc(self, &dest)?;
             }
 
-            NullaryOp(mir::NullOp::SizeOf, ty) => {
+            NullaryOp(null_op, ty) => {
```
nagisa, line 273: Please match `mir::NullOp::SizeOf | mir::NullOp::AlignOf` here.  so that the match remains properly exhaustive and will catch additional `NullOp` additions in the future.

compiler/rustc_codegen_ssa/src/mir/rvalue.rs
```diff
@@ -525,6 +511,27 @@ impl<'a, 'tcx, Bx: BuilderMethods<'a, 'tcx>> FunctionCx<'a, 'tcx, Bx> {
                 let operand = OperandRef { val: OperandValue::Immediate(val), layout: box_layout };
                 (bx, operand)
             }
+
+            mir::Rvalue::NullaryOp(null_op, ty) => {
```
nagisa, line 515:
```suggestion
            mir::Rvalue::NullaryOp(null_op@mir::NullOp::SizeOf | null_op@mir::NullOp::AlignOf, ty) => {
```

compiler/rustc_codegen_cranelift/src/base.rs
```diff
@@ -726,15 +726,20 @@ fn codegen_stmt<'tcx>(
                     let ptr = fx.bcx.inst_results(call)[0];
                     lval.write_cvalue(fx, CValue::by_val(ptr, box_layout));
                 }
-                Rvalue::NullaryOp(NullOp::SizeOf, ty) => {
+                Rvalue::NullaryOp(null_op, ty) => {
```
nagisa, line 729:
```suggestion
                Rvalue::NullaryOp(null_op@NullOp::SizeOf | null_op@NullOp::AlignOf, ty) => {
```

nagisa (approved):
Broadly LGTM, r=me with exhaustive matches back.

compiler/rustc_const_eval/src/interpret/step.rs
nbdd0121: How about change the _ arm below matching null_op to a NullOp::Box arm instead? This way we still get exhaustive match and is IMO cleaner.

nagisa: I don't have a strong opinion either way.

compiler/rustc_middle/src/mir/mod.rs
RalfJung: > Should we remove preferred alignment then if it isn't used anyway?

Yeah it might be time to remove that intrinsic... @eddyb or do you think that will ever be useful again in the future?

PR #88841, opened 2021-09-11
feat(rustc_typeck): suggest removing bad parens in `(recv.method)()`

### LINKED ISSUE #88803: With over-parenthesized method call, E0615 suggests adding parentheses in the wrong place
Given the following [code](https://play.rust-lang.org/?version=nightly&mode=debug&edition=2018&gist=702c89fffada54a39f90d2ad99dae4bb): 

```rust
fn main() {
    let a = Some(42);
    println!(
        "The value is {}.",
        (a.unwrap)()
    );
}
```

The current output is:

```
error[E0615]: attempted to take value of method `unwrap` on type `Option<{integer}>`
 --> src/main.rs:5:12
  |
5 |         (a.unwrap)()
  |            ^^^^^^ method, not a field
  |
help: use parentheses to call the method
  |
5 |         (a.unwrap)()()
  |                   ++
```

```
help: remove these parentheses

(a.unwrap)()
^^^^^^^^^^ 
```

But at least adding parentheses right next to the method name would be an improvement:

```
help: use parentheses to call the method
  |
5 |         (a.unwrap())()
  |                  ++
```

notriddle:
Fixes #88803

### DIFF
--- a/compiler/rustc_typeck/src/check/expr.rs
+++ b/compiler/rustc_typeck/src/check/expr.rs
@@ -1842,13 +1842,14 @@ impl<'a, 'tcx> FnCtxt<'a, 'tcx> {
             expr_t
         );
         err.span_label(field.span, "method, not a field");
-        let expr_is_call = if let hir::Node::Expr(parent_expr) =
-            self.tcx.hir().get(self.tcx.hir().get_parent_node(expr.hir_id))
-        {
-            matches!(parent_expr.kind, ExprKind::Call(..))
-        } else {
-            false
-        };
+        let expr_is_call =
+            if let hir::Node::Expr(hir::Expr { kind: ExprKind::Call(callee, _args), .. }) =
+                self.tcx.hir().get(self.tcx.hir().get_parent_node(expr.hir_id))
+            {
+                expr.hir_id == callee.hir_id
+            } else {
+                false
+            };
         let expr_snippet =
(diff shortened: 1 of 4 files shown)


compiler/rustc_typeck/src/check/expr.rs
```diff
@@ -1842,7 +1842,24 @@ impl<'a, 'tcx> FnCtxt<'a, 'tcx> {
...
+            };
+        let expr_snippet =
+            self.tcx.sess.source_map().span_to_snippet(expr.span).unwrap_or(String::new());
+        if expr_is_call && expr_snippet.starts_with("(") && expr_snippet.ends_with(")") {
+            err.span_suggestion_short(
+                expr.span,
+                "remove wrapping parentheses to call the method",
+                expr_snippet[1..expr_snippet.len() - 1].to_owned(),
+                Applicability::MachineApplicable,
+            );
```
estebank, line 1861: (The AST holds an `ast::ExprKind::Paren`, but the HIR doesn't).

Ideally we would give a `multipart_suggestion` where the suggestion is something like

```
let suggestion = vec![
    (expr.span.with_hi(callee.span.lo()), String::new()),
    (expr.span.with_lo(callee.span.hi()), String::new()),
];
```

estebank (approved):

notriddle:
@estebank it's been converted to a multipart suggestion.

PR #88845, opened 2021-09-11
Make crate_dependencies_in_postorder public

RDambrosio016:
Small PR, i need these functions for a custom codegen i am working on, for lazily loading modules (only loading symbols that previous modules need)

### DIFF
--- a/compiler/rustc_metadata/src/creader.rs
+++ b/compiler/rustc_metadata/src/creader.rs
@@ -175,7 +175,7 @@ impl CStore {
         }
     }
 
-    crate fn crate_dependencies_in_postorder(&self, cnum: CrateNum) -> Vec<CrateNum> {
+    pub fn crate_dependencies_in_postorder(&self, cnum: CrateNum) -> Vec<CrateNum> {
         let mut deps = Vec::new();
         if cnum == LOCAL_CRATE {
             self.iter_crate_data(|cnum, _| self.push_dependencies_in_postorder(&mut deps, cnum));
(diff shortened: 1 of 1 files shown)


jyn514:
cc @petrochenkov

davidtwco (commented):
Despite being a very minor change, I'm not sure that we normally accept internal compiler changes to support personal projects using rustc internals. This could be something we'd accept once a project is more mature and released, but it might make sense for you to maintain a small fork of rustc for now. I've [left a comment to double check our policy here](https://github.com/rust-lang/rust-forge/pull/553#issuecomment-917938159).

davidtwco:
Seems like we'd be fine with privacy changes like these so long as the current privacy isn't an important part of an invariant. Git blame suggests @petrochenkov might be more familiar with this code.

r? @petrochenkov

petrochenkov:
@RDambrosio016 
Is `tcx` available in the place from which you want to call these functions?
In that case using `postorder_cnums` is preferable.

RDambrosio016:
@petrochenkov it kind of is but it also isn't, I need this during `link()` in the codegen, but rustc drops tyctxt before linking. So i hack around it by storing a mutex'd `Vec<String>` of dependencies which i override on `codegen_crate`. I did not notice that `postorder_cnums()` existed, i asked about this but was only pointed to the method on cstore, i will try this instead and see if it works. To provide more context, i am writing a backend which targets the `libnvvm` CUDA library for writing fast GPU kernels in rust. However, nvvm does not perform DCE over all functions, which results in about an 11mb ptx file on debug. However it has a [lazy loading method](https://docs.nvidia.com/cuda/libnvvm-api/group__compilation.html#group__compilation_1g5356ce5063db232cd4330b666c62219b). Using this reduces code size to a couple hundred lines for a simple kernel.

RDambrosio016:
reversed postorder_cnums seems to be working from what i can see, so i will just use that 👍

PR #88851, opened 2021-09-11
Fix duplicate bounds for const_trait_impl

### LINKED ISSUE #88383: Duplicate bounds behave differently depending on their order for const_trait_impl
I tried this code:

```rust
const fn equals_self<T: PartialEq + ~const PartialEq>(t: &T) -> bool {
    *t == *t
}
```

I expected to see this happen: It should be the same as
```rust
const fn equals_self<T: ~const PartialEq + PartialEq>(t: &T) -> bool {
    *t == *t
}
```

Instead, this happened: It errors for the former but does not error for the second example.

@rustbot label F-const_trait_impl

See https://github.com/rust-lang/rust/pull/88328/files#r697229693

fee1-dead:
Fixes #88383.

Compare the constness of the candidates before winnowing and removing a `~const` `BoundCandidate`.

### DIFF
--- a/compiler/rustc_trait_selection/src/traits/select/mod.rs
+++ b/compiler/rustc_trait_selection/src/traits/select/mod.rs
@@ -1487,10 +1487,11 @@ impl<'cx, 'tcx> SelectionContext<'cx, 'tcx> {
             ) => false,
 
             (ParamCandidate(other), ParamCandidate(victim)) => {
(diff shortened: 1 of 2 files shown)


oli-obk:
Would it make sense to drop the `NotConst` bound if the constness is unequal? It does not add anything new after all

PR #88855, opened 2021-09-11
Allow simd_shuffle to accept vectors of any length

calebzulawski:
cc @rust-lang/project-portable-simd @workingjubilee

### DIFF
--- a/src/test/ui/simd-intrinsic/simd-intrinsic-generic-shuffle.rs
+++ b/src/test/ui/simd-intrinsic/simd-intrinsic-generic-shuffle.rs
@@ -1,0 +1,33 @@
+// build-fail
+
+// Test that the simd_shuffle intrinsic produces ok-ish error
+// messages when misused.
+
+#![feature(repr_simd, platform_intrinsics)]
+
+#[repr(simd)]
+#[derive(Copy, Clone)]
+pub struct Simd<T, const N: usize>([T; N]);
+
+extern "platform-intrinsic" {
+    fn simd_shuffle<T, I, U>(a: T, b: T, i: I) -> U;
+}
+
+fn main() {
+    const I: [u32; 2] = [0; 2];
+    const I2: [f32; 2] = [0.; 2];
+    let v = Simd::<u32, 4>([0; 4]);
+
+    unsafe {
+        let _: Simd<u32, 2> = simd_shuffle(v, v, I);
+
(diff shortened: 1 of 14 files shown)


src/test/ui/simd-intrinsic/simd-intrinsic-generic-shuffle.rs
```diff
@@ -0,0 +1,33 @@
+// build-fail
+
+// Test that the simd_shuffle intrinsic produces ok-ish error
+// messages when misused.
```
nagisa, line 4: Are these error messages going to be the same for whatever wraps these intrinsics? If so, the diagnostics should definitely be better than ok-ish, I feel.

calebzulawski: Honestly, I just copied this file from another in this directory 🙂 This error should be completely prevented by the type system in `std::simd`, so this error should never make it to a user.  Actually, it was decided a while back that `std::simd` should never produce monomorphisation errors.

nagisa: I guess its fine to not add a test like this at all, in that case, I feel, as its only an implementation detail, but I guess the cost of such a test is pretty low regardless.

calebzulawski: For now, it's useful to test the intrinsics because `std::simd` isn't in tree yet.  After it is, you are probably right.

workingjubilee: Yeah, I would rather test that we generate a distinct error at all if something slips through, but it should "never occur" for external users who use the intended API, so from our perspective anyone deliberately using this is opting into the Bad Error.

workingjubilee:
While experimenting with more complex actual uses of this than it was initially tested for, we found a snag in the current version that causes an ICE if it is hit. We are drafting a fix but figured we might as well push a new revision here, since it seems like, due to recent queue issues, following up in another PR would take even longer.

calebzulawski:
The simple tests I wrote did not catch that if the constant arg needed any evaluation, it was not monomorphized and crashed.  I added a more complex test case that tests this scenario.  Should be good to go now.

PR #88779, opened 2021-09-09
Use more accurate spans for "unused delimiter" lint

estebank:

PR #88857, opened 2021-09-11
Rollup of 10 pull requests

workingjubilee:
Successful merges:

 - #87904 (Reword description of automatic impls of `Unsize`.)
 - #88147 (Fix non-capturing closure return type coercion)
 - #88209 (Improve error message when _ is used for in/inout asm operands)
 - #88668 (Change more x64 size checks to not apply to x32.)
 - #88733 (Fix ICE for functions with more than 65535 arguments)
 - #88757 (Suggest wapping expr in parentheses on invalid unary negation)
 - #88779 (Use more accurate spans for "unused delimiter" lint)
 - #88830 (Add help for E0463)
 - #88849 (don't clone types that are Copy (clippy::clone_on_copy))
 - #88850 (don't convert types into identical types)

Failed merges:

r? @ghost
@rustbot modify labels: rollup

[Create a similar rollup](https://bors.rust-lang.org/queue/rust?prs=87904,88147,88209,88668,88733,88757,88779,88830,88849,88850)

### DIFF
--- a/compiler/rustc_ast_passes/src/ast_validation.rs
+++ b/compiler/rustc_ast_passes/src/ast_validation.rs
@@ -422,9 +422,23 @@ impl<'a> AstValidator<'a> {
     }
 
     fn check_fn_decl(&self, fn_decl: &FnDecl, self_semantic: SelfSemantic) {
+        self.check_decl_num_args(fn_decl);
         self.check_decl_cvaradic_pos(fn_decl);
         self.check_decl_attrs(fn_decl);
         self.check_decl_self_param(fn_decl, self_semantic);
+    }
+
+    /// Emits fatal error if function declaration has more than `u16::MAX` arguments
+    /// Error is fatal to prevent errors during typechecking
+    fn check_decl_num_args(&self, fn_decl: &FnDecl) {
+        let max_num_args: usize = u16::MAX.into();
+        if fn_decl.inputs.len() > max_num_args {
+            let Param { span, .. } = fn_decl.inputs[0];
+            self.err_handler().span_fatal(
+                span,
+                &format!("function can not have more than {} arguments", max_num_args),
+            );
+        }
     }
 
     fn check_decl_cvaradic_pos(&self, fn_decl: &FnDecl) {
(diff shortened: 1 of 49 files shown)


workingjubilee:
Omitted PRs that seemed like they might affect exports from `std::os` somehow.
No obvious reason for why that would be the case, but that's the only clue to go on for the last rollup's failure.

Mark-Simulacrum:
This is very surprising to me, given the list of benchmarks in this PR -- do we have a suspect for this?

@jackh726 do you think #88147 may be responsible here? If so, it seems a little surprising -- I'd expect that the rather small change there wouldn't be responsible (or is perhaps subtly wrong or something?). Maybe this is just undiscovered low hanging fruit though!

jackh726:
Woah. Uh, maybe...

I would be surprised if #88147 is wrong, given that the LUB of two equal types *is* itself. I guess it's not surprising that we can be much faster in the easy case; I think before we would have done a bunch of work that we didn't really need to do. But I think maybe I might poke @nikomatsakis here.

jackh726:
The perf certainly are nothing to scoff at...

workingjubilee:
We could post the reversed PR and bench-compare that, see if we observe a similar difference.

workingjubilee:
I ran that experiment in #88926 and we can see decisively it is indeed that change: https://perf.rust-lang.org/compare.html?start=9f85cd6f2ab2769c16e89dcdddb3e11d9736b351&end=cc08b15420582202a8bffdb12043459d014ffb75

I am not surprised at the result having a massive improvement, though. The fastest code is the code that doesn't run. An occasional branch, costing a few bytes and cycles, especially a predictable one (so there's no pipeline stalls from a mispredict) will in fact yield a massive win on a test for something in that niche.

Mark-Simulacrum:
Thanks @workingjubilee!

Yeah, the big win is not *that* surprising here, it's more so that I was worried we hadn't previously done it for some good reason (e.g., side effects in the code below that new if). But it seems like that's not the case.

PR #88858, opened 2021-09-11
Allow reverse iteration of lowercase'd/uppercase'd chars

mispecto:
The PR implements `DoubleEndedIterator` trait for `ToLowercase` and `ToUppercase`.

This enables reverse iteration of lowercase/uppercase variants of character sequences.
One of use cases:  determining whether a char sequence is a suffix of another one.

Example:

```rust
fn endswith_ignore_case(s1: &str, s2: &str) -> bool {
    for eob in s1
        .chars()
        .flat_map(|c| c.to_lowercase())
        .rev()
        .zip_longest(s2.chars().flat_map(|c| c.to_lowercase()).rev())
    {
        match eob {
            EitherOrBoth::Both(c1, c2) => {
                if c1 != c2 {
                    return false;
                }
            }
            EitherOrBoth::Left(_) => return true,
            EitherOrBoth::Right(_) => return false,
        }
    }
    true
}
```

### DIFF
--- a/library/alloc/tests/str.rs
+++ b/library/alloc/tests/str.rs
@@ -1126,6 +1126,37 @@ fn test_rev_iterator() {
 }
 
 #[test]
+fn test_to_lowercase_rev_iterator() {
+    let s = "AÖßÜ💩ΣΤΙΓΜΑΣǅﬁİ";
+    let v = ['\u{307}', 'i', 'ﬁ', 'ǆ', 'σ', 'α', 'μ', 'γ', 'ι', 'τ', 'σ', '💩', 'ü', 'ß', 'ö', 'a'];
+
+    let mut pos = 0;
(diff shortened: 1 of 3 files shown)


scottmcm:
Thanks for the PR, @spektom!  This seems reasonable to me.

As trait impls are insta-stable, though, this needs an FCP from the libs team, so...

r? @yaahc 

(It also needs the versions in the `#[stable]` attributes changed, but that can wait until we know when it could go in.)

dtolnay:
@rfcbot fcp merge

Reverse iteration for https://doc.rust-lang.org/core/char/struct.ToLowercase.html and https://doc.rust-lang.org/core/char/struct.ToUppercase.html.

dtolnay (approved):

PR #88860, opened 2021-09-11
Deduplicate panic_fmt

nbdd0121:
std's begin_panic_fmt and core's panic_fmt are duplicates. Merge them to declutter code and remove a lang item.

### DIFF
--- a/library/std/src/panic.rs
+++ b/library/std/src/panic.rs
@@ -25,7 +25,7 @@ pub macro panic_2015 {
         $crate::rt::panic_display(&$arg)
     }),
     ($fmt:expr, $($arg:tt)+) => ({
-        $crate::rt::begin_panic_fmt(&$crate::const_format_args!($fmt, $($arg)+))
+        $crate::rt::panic_fmt($crate::const_format_args!($fmt, $($arg)+))
     }),
 }
 
--- a/compiler/rustc_const_eval/src/const_eval/machine.rs
+++ b/compiler/rustc_const_eval/src/const_eval/machine.rs
@@ -51,9 +51,7 @@ impl<'mir, 'tcx> InterpCx<'mir, 'tcx, CompileTimeInterpreter<'mir, 'tcx>> {
             let span = self.find_closest_untracked_caller_location();
             let (file, line, col) = self.location_triple_for_span(span);
             return Err(ConstEvalErrKind::Panic { msg, file, line, col }.into());
-        } else if Some(def_id) == self.tcx.lang_items().panic_fmt()
-            || Some(def_id) == self.tcx.lang_items().begin_panic_fmt()
-        {
+        } else if Some(def_id) == self.tcx.lang_items().panic_fmt() {
             // For panic_fmt, call const_panic_fmt instead.
             if let Some(const_panic_fmt) = self.tcx.lang_items().const_panic_fmt() {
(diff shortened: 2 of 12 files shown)


library/std/src/panic.rs
```diff
@@ -21,7 +21,7 @@ pub macro panic_2015 {
         $crate::rt::begin_panic($msg)
     }),
     ($fmt:expr, $($arg:tt)+) => ({
-        $crate::rt::begin_panic_fmt(&$crate::const_format_args!($fmt, $($arg)+))
+        $crate::rt::begin_panic_fmt($crate::const_format_args!($fmt, $($arg)+))
```
klensy, line 24: Why?

nbdd0121: It's an existing inconsistency between libcore and libstd macro API. The inconsistency actually makes the `compiler/rustc_const_eval/src/const_eval/machine.rs`'s `begin_panic_fmt` -> `const_panic_fmt` replacement unsound (but since const panic doesn't allow arguments today this unsound path is never hit).

nbdd0121: > Ahh, `begin_panic_fmt` get params by reference, when `panic_fmt` takes by value. It's actually interesting to change panic_fmt to get it by reference too, as reference should be shorter that `Arguments` struct.

I actually don't think there'll be any meaningful difference. Size of `Arguments` would make it be passed by reference anyway.

Plus, many existing methods like `write_fmt` by value and `Arguments` is Copy so I think it's actually good to take it by value for consistency.

klensy:
```
print-type-size type: `std::fmt::Arguments`: 48 bytes, alignment: 8 bytes
print-type-size     field `.pieces`: 16 bytes
print-type-size     field `.fmt`: 16 bytes
print-type-size     field `.args`: 16 bytes
```

nbdd0121: What are you suggesting by listing the numbers? As I've said, the size of `Arguments` (is larger than 2 usize) would make it be passed by reference. So codegen will behave exactly the same with or without the `&` (https://godbolt.org/z/h69MKjdsn), and for semantics and consistency it should be `Arguments` rather than `&Arguments`.

klensy: Sorry, i didn't knew that there some optimization that pass big Copy types by ref even if i explicitly pass it by value.

Value passed into `PanicInfo::internal_constructor` by `Option`'ing reference anyway, so there is no sense to copying it before that.

klensy: Tried to check that in small example, looks like parameter really isn't copied in `panic_fmt` (didn't checked this in stdlib). I remember that i played somewhere with fmt::Arguments and saw needless copying in stdlib.

m-ou-se:
I'm guessing that it trips on the `&` that got removed, which makes it unable to recognize the panic message and arguments anymore.

nbdd0121:
https://github.com/rust-lang/rust-clippy/blob/389a74b31aabca4aac3289763eeb2eccedd1b988/clippy_utils/src/higher.rs#L717 -- That's a very detailed assumption about what it expands to 🙃.

So `AddrOf` is indeed matched. And I guess the `let ExprKind::Block` is the reason it won't recognize core's macro. Another data point on why the brace should be removed.

PR #88865, opened 2021-09-11
Implement `#[must_not_suspend]`

guswynn:
implements #83310

Some notes on the impl:

1. The code that searches for the attribute on the ADT is basically copied from the `must_use` lint. It's not shared, as the logic did diverge
2. The RFC does specify that the attribute can be placed on fn's (and fn-like objects), like `must_use`. I think this is a direct copy from the `must_use` reference definition. This implementation does NOT support this, as I felt that ADT's (+ `impl Trait` + `dyn Trait`) cover the usecase's people actually want on the RFC, and adding an imp for the fn call case would be significantly harder. The `must_use` impl can do a single check at fn call stmt time, but `must_not_suspend` would need to answer the question: "for some value X with type T, find any fn call that COULD have produced this value". That would require significant changes to `generator_interior.rs`, and I would need mentorship on that. @eholk and I are discussing it.
3. @estebank do you know a way I can make the user-provided `reason` note pop out? right now it seems quite hidden

Also, I am not sure if we should run perf on this

r? @nikomatsakis


compiler/rustc_lint_defs/src/builtin.rs
```diff
@@ -314,6 +314,13 @@ declare_lint! {
     "imports that are never used"
 }
 
+declare_lint! {
+    /// [the reference]: https://doc.rust-lang.org/reference/attributes/diagnostics.html#the-must_use-attribute
```
jyn514, line 318: You probably want to expand this before landing. There's examples above and below.

compiler/rustc_passes/src/check_attr.rs
```diff
@@ -1014,6 +1015,21 @@ impl CheckAttrVisitor<'tcx> {
...
     }
 
+    /// Checks if `#[must_not_suspend]` is applied to a function. Returns `true` if valid.
+    fn check_must_not_suspend(&self, attr: &Attribute, span: &Span, target: Target) -> bool {
+        match target {
+            Target::Fn | Target::Method(..) | Target::ForeignFn | Target::Closure => {
+                self.tcx
+                    .sess
+                    .struct_span_err(attr.span, "`must_not_suspend` attribute should be applied to a struct, enum, `impl Trait`, or `dyn Trait`")
+                        .span_label(*span, "is a function")
```
jyn514, line 1025: Hmm, shouldn't this check other items besides functions? It seems wrong to apply this to modules/re-exports/constants too.

compiler/rustc_typeck/src/check/generator_interior.rs
```diff
@@ -409,3 +448,195 @@ impl<'a, 'tcx> Visitor<'tcx> for ArmPatCollector<'a> {
...
+        ty::Tuple(ref tys) => {
+            let mut has_emitted = false;
+            /*
+            let spans = if let hir::ExprKind::Tup(comps) = &expr.kind {
+                debug_assert_eq!(comps.len(), tys.len());
+                comps.iter().map(|e| e.span).collect()
+            } else {
+                vec![]
+            };
+            */
```
jyn514, line 557:
```suggestion
```

compiler/rustc_typeck/src/check/generator_interior.rs
```diff
@@ -409,3 +448,195 @@ impl<'a, 'tcx> Visitor<'tcx> for ArmPatCollector<'a> {
...
+                    plural_len,
+                ) {
+                    has_emitted = true;
+                }
+            }
+            has_emitted
+        }
+        ty::Array(ty, len) => match len.try_eval_usize(fcx.tcx, param_env) {
+            // If the array is empty we don't lint, to avoid false positives
+            Some(0) | None => false,
```
jyn514, line 573: Hmm, when would this be `None`? Are you sure it shouldn't lint in that case?

src/test/ui/lint/must_not_suspend/boxed.rs
```diff
@@ -0,0 +1,25 @@
...
+fn bar() -> Box<Umm> {
+    Box::new(Umm {
+        i: 1
+    })
+}
+
+async fn other() {}
+
+pub async fn uhoh() {
+    let _guard = bar(); //~ boxed `Umm` held across
```
jyn514, line 20:
```suggestion
    let _guard = bar(); //~ ERROR boxed `Umm` held across
```

src/test/ui/lint/must_not_suspend/return.rs
```diff
@@ -0,0 +1,9 @@
+// edition:2018
+#![feature(must_not_suspend)]
+#![deny(must_not_suspend)]
+
+#[must_not_suspend] //~ attribute should be
```
jyn514, line 5:
```suggestion
#[must_not_suspend] //~ ERROR attribute should be
```

src/test/ui/lint/must_not_suspend/return.rs
```diff
@@ -0,0 +1,9 @@
+// edition:2018
+#![feature(must_not_suspend)]
+#![deny(must_not_suspend)]
```
jyn514, line 3:
```suggestion
#![deny(must_not_suspend)]
#![must_not_suspend] //~ ERROR attribute should be
```

src/test/ui/lint/must_not_suspend/unit.stderr
```diff
@@ -0,0 +1,23 @@
...
+   |         ^^^^^^
+LL |     other().await;
+   |     ------------- The value is held across this yield point
+   |
+note: the lint level is defined here
+  --> $DIR/unit.rs:3:9
+   |
+LL | #![deny(must_not_suspend)]
+   |         ^^^^^^^^^^^^^^^^
+   = note: You gotta use Umm's, ya know?
```
jyn514, line 14: How were you expecting this would look? It seems fine to me :) but you mentioned in the PR you wanted to improve it.

compiler/rustc_passes/src/check_attr.rs
guswynn: My understanding is that the attribute is already defined in `builtin_attrs.rs` as `Normal`, which already narrows the scope of what it can be applied to? I can add a test and check

compiler/rustc_typeck/src/check/generator_interior.rs
guswynn: I think expr is an Option, but yah I can fix this

compiler/rustc_typeck/src/check/generator_interior.rs
guswynn: I am not sure, this is copied from `must_use`'s impl

I think you are right, we should still lint if we have an empty array that holds `must_not_suspend` types

src/test/ui/lint/must_not_suspend/unit.stderr
guswynn: I think it looks hidden under the lint level diagnostic

guswynn:
@jyn514 the panic is:
```
thread 'rustc' panicked at 'Box<dyn Any>', compiler/rustc_errors/src/lib.rs:1147:9
stack backtrace:
   0: std::panicking::begin_panic
   1: std::panic::panic_any
   2: rustc_errors::HandlerInner::bug
   3: rustc_errors::Handler::bug
   4: rustc_middle::util::bug::opt_span_bug_fmt::{{closure}}
   5: rustc_middle::ty::context::tls::with_opt::{{closure}}
   6: rustc_middle::ty::context::tls::with_opt
   7: rustc_middle::util::bug::opt_span_bug_fmt
   8: rustc_middle::util::bug::bug_fmt
   9: rustc_middle::ich::impls_ty::<impl rustc_data_structures::stable_hasher::HashStable<rustc_middle::ich::hcx::StableHashingContext> for rustc_middle::ty::sty::RegionKind>::hash_stable
  10: rustc_middle::ty::sty::_DERIVE_rustc_data_structures_stable_hasher_HashStable_rustc_middle_ich_StableHashingContext_ctx_FOR_TyKind::<impl rustc_data_structures::stable_hasher::HashStable<rustc_middle::ich::hcx::StableHashingContext> for rustc_middle::ty::sty::TyKind>::hash_stable
  11: <T as rustc_query_system::dep_graph::dep_node::DepNodeParams<Ctxt>>::to_fingerprint
  12: rustc_query_system::dep_graph::dep_node::DepNode<K>::construct
  13: rustc_query_system::query::plumbing::get_query_impl
  14: rustc_query_system::query::plumbing::get_query
  15: <rustc_query_impl::Queries as rustc_middle::ty::query::QueryEngine>::type_uninhabited_from
  16: rustc_middle::ty::inhabitedness::<impl rustc_middle::ty::TyS>::uninhabited_from
  17: rustc_middle::ty::inhabitedness::<impl rustc_middle::ty::context::TyCtxt>::is_ty_uninhabited_from
  18: rustc_typeck::check::generator_interior::check_must_not_suspend_ty
  19: rustc_typeck::check::generator_interior::InteriorVisitor::record
  20: <rustc_typeck::check::generator_interior::InteriorVisitor as rustc_hir::intravisit::Visitor>::visit_pat
  21: rustc_hir::intravisit::walk_block
  22: <rustc_typeck::check::generator_interior::InteriorVisitor as rustc_hir::intravisit::Visitor>::visit_expr
  23: rustc_typeck::check::generator_interior::resolve_interior
  24: rustc_typeck::check::fn_ctxt::_impl::<impl rustc_typeck::check::fn_ctxt::FnCtxt>::resolve_generator_interiors
  25: rustc_infer::infer::InferCtxtBuilder::enter
  ...
  ```

Do you know what that could be?

**Comment by guswynn** [comment]
Seems like the `is_ty_uninhabitated_from` check, which was copied from the `#[must_use]` check, is a query only really available AFTER typeck, as it required that `ReVar` and `RePlaceholder` are purged. 

I am not sure what this does, and if its needed for this lint, the last person to touch it seems to be @oli-obk in https://github.com/rust-lang/rust/commit/7894509b000157185b10cfae64ac1e88acd88f4a ?

**Comment by guswynn** [comment]
this pr should be reviewable, except for I am unsure about if the is_ty_uninhabitated check is needed, currently its off

File: compiler/rustc_typeck/src/check/generator_interior.rs [hunk 9]
```diff
@@ -409,3 +447,184 @@ impl<'a, 'tcx> Visitor<'tcx> for ArmPatCollector<'a> {
...
+                        err.note(&note.as_str());
+                    }
+
+                    // add span pointing to the offending yield/await)
+                    err.span_label(yield_span, "The value is held across this yield point");
+
+                    // Add some quick suggestions on what to do
+                    err.span_help(
+                        source_span,
+                        "`drop` this value before the yield point, or use a block (`{ ... }`) \"
```
camelid, line 618: You have an extra `"` here, which shows up in the error:

```suggestion
                        "`drop` this value before the yield point, or use a block (`{ ... }`) \
```

compiler/rustc_typeck/src/check/generator_interior.rs
```diff
@@ -409,3 +447,184 @@ impl<'a, 'tcx> Visitor<'tcx> for ArmPatCollector<'a> {
...
+                        descr_post_path
+                    );
+                    let mut err = lint.build(&msg);
+
+                    // Add optional reason note
+                    if let Some(note) = attr.value_str() {
+                        err.note(&note.as_str());
+                    }
+
+                    // add span pointing to the offending yield/await)
```
camelid, line 612: nit: You have an extra `)` here:

```suggestion
                    // add span pointing to the offending yield/await
```

compiler/rustc_typeck/src/check/generator_interior.rs
```diff
@@ -409,3 +447,184 @@ impl<'a, 'tcx> Visitor<'tcx> for ArmPatCollector<'a> {
...
+                    err.span_label(yield_span, "The value is held across this yield point");
```
camelid, line 613: Labels usually start with a lowercase letter:

```suggestion
                    err.span_label(yield_span, "the value is held across this yield point");
```

compiler/rustc_feature/src/active.rs
```diff
@@ -679,6 +679,10 @@ declare_features! (
     /// Allows `let...else` statements.
     (active, let_else, "1.56.0", Some(87335), None),
 
+    /// Allows `#[must_not_suspend]`.
+    (active, must_not_suspend, "1.56.0", Some(83310), None),
+
```
camelid, line 684: Note: 1.56.0 has already branched off as beta, so this will land in 1.57.0 or later.

compiler/rustc_passes/src/check_attr.rs
jyn514: A test would be great, thanks :)

guswynn: I think this is now covered by the other_items.rs test

compiler/rustc_lint_defs/src/builtin.rs
```diff
@@ -314,6 +314,36 @@ declare_lint! {
...
+    /// async fn yield_now() {}
+    ///
+    /// pub async fn uhoh() {
+    ///     let guard = SyncThing {};
+    ///     yield_now().await;
+    /// }
+    /// ```
+    pub MUST_NOT_SUSPEND,
+    Warn,
+    "Use of a `#[must_not_suspend]` value across a yield point",
```
estebank, line 344:
```suggestion
    "use of a `#[must_not_suspend]` value across a yield point",
```

compiler/rustc_typeck/src/check/generator_interior.rs
```diff
@@ -409,3 +447,184 @@ impl<'a, 'tcx> Visitor<'tcx> for ArmPatCollector<'a> {
...
+    hir_id: HirId,
+    expr: Option<&'tcx Expr<'tcx>>,
+    source_span: Span,
+    yield_span: Span,
+    descr_pre: &str,
+    descr_post: &str,
+    plural_len: usize,
+) -> bool {
+    if ty.is_unit()
+    // || fcx.tcx.is_ty_uninhabited_from(fcx.tcx.parent_module(hir_id).to_def_id(), ty, fcx.param_env)
```
estebank, line 466: Remove commented out code or explain in a comment why it is being left in but commented out.

estebank:
> do you know a way I can make the user-provided reason note pop out? right now it seems quite hidden

There are a few things I do to make important info appear more obvious:

 - make them part of the primary label, instead of using a `note`, use a `span_label`
   - it might make sense to make `must_not_suspend` have a way to be used as `#[must_not_suspend(label = "fooberize this", note = "foos must bar")`
 - use a `span_note` to make subdiagnostic more obvious, this also gives the header (`note`/`help`) a color, which we don't show when they are shown as a single line
 - move the `note` earlier or later, they are harder to read in the middle

Sadly this is about as much as we can accomplish with the APIs we have today.

src/test/ui/lint/must_not_suspend/unit.stderr
estebank: We could maybe change the logic for lints to always include the `the lint level is defined here` last, but that would go beyond the scope of this PR, IMO.

compiler/rustc_typeck/src/check/generator_interior.rs
```diff
@@ -290,7 +307,7 @@ impl<'a, 'tcx> Visitor<'tcx> for InteriorVisitor<'a, 'tcx> {
         if let PatKind::Binding(..) = pat.kind {
             let scope = self.region_scope_tree.var_scope(pat.hir_id.local_id);
             let ty = self.fcx.typeck_results.borrow().pat_ty(pat);
-            self.record(ty, Some(scope), None, pat.span, false);
+            self.record(ty, pat.hir_id, Some(scope), None, pat.span, false);
```
oli-obk, line 310: Not your fault, but this list of arguments is getting out of hand... Maybe we should make it a builder pattern or just pass a struct with appropriate defaults, constructor functions or sth.

compiler/rustc_typeck/src/check/generator_interior.rs
```diff
@@ -409,3 +447,184 @@ impl<'a, 'tcx> Visitor<'tcx> for ArmPatCollector<'a> {
...
+        ty::Adt(def, _) => check_must_not_suspend_def(
+            fcx.tcx,
+            def.did,
+            hir_id,
+            source_span,
+            yield_span,
+            descr_pre,
+            descr_post,
+        ),
+        ty::Opaque(def, _) => {
```
oli-obk, line 499: This only checks for the traits on the opaque type, but it may make sense to additionally reveal the opaque type similar to how `Send` and `Sync` are leaked through opaque types.

Alternatively, with type alias impl trait, it may be useful to allow adding the attribute to the declaring type alias.

compiler/rustc_typeck/src/check/generator_interior.rs
```diff
@@ -409,3 +447,184 @@ impl<'a, 'tcx> Visitor<'tcx> for ArmPatCollector<'a> {
...
+            check_must_not_suspend_ty(
+                fcx,
+                boxed_ty,
+                hir_id,
+                expr,
+                source_span,
+                yield_span,
+                descr_pre,
+                descr_post,
+                plural_len,
```
oli-obk, line 487: There's a lot of arguments that are always the same or only some get changed. Maybe put all of them in a struct, so you can just use the `..defaults` syntax to only overwrite a few fields.

src/test/ui/lint/must_not_suspend/unit.stderr
guswynn: Filed #88908 as a followup

compiler/rustc_typeck/src/check/generator_interior.rs
guswynn: This lint only triggers on an `impl Trait` if the trait itself is marked as `must_not_suspend`. If the value made opaque by that `impl Trait` (or `dyn Trait`) is `must_not_suspend`, that is not caught

I can definitely add a comment to add TAIT support though!

guswynn:
@estebank I think out of the options, `span_note` is the best!

estebank:
@guswynn let's make a note (pun intended) to fix it later.

nikomatsakis:
I think I'd prefer to have @estebank or @oli-obk review this! I'm going to put

r? @oli-obk 

since it feels pretty close to clippy lint territory. =)

compiler/rustc_typeck/src/check/generator_interior.rs
```diff
@@ -409,3 +447,187 @@ impl<'a, 'tcx> Visitor<'tcx> for ArmPatCollector<'a> {
...
+        return true;
+    }
+
+    let plural_suffix = pluralize!(plural_len);
+
+    match *ty.kind() {
+        ty::Adt(..) if ty.is_box() => {
+            let boxed_ty = ty.boxed_ty();
+            let descr_pre = &format!("{}boxed ", descr_pre);
+            check_must_not_suspend_ty(
```
eholk, line 480: This is basically unpeeling the type to check the type underlying the box, right?

eholk (commented):
Do you have any test cases that make sure dropping a `must_not_suspend` value before a suspend point works as expected? I suspect we may have some trouble with this until I finish the liveness-based analysis for generators.

compiler/rustc_typeck/src/check/generator_interior.rs
guswynn: yep! We may find in the future we want to handle things like Arc (maybe "things that Deref to `#[must_not_suspend]`) but time will tell

guswynn:
@eholk good point, it looks like `drop` doesn't work, but the scope does, so I'll narrow the suggestion for now

compiler/rustc_typeck/src/check/generator_interior.rs
```diff
@@ -409,3 +448,170 @@ impl<'a, 'tcx> Visitor<'tcx> for ArmPatCollector<'a> {
...
+    yield_span: Span,
+    descr_pre: &'a str,
+    descr_post: &'a str,
+    plural_len: usize,
+}
+
+// Returns whether it emitted a diagnostic or not
+// Note that this fn and the proceding one are based on the code
+// for creating must_use diagnostics
+pub fn check_must_not_suspend_ty<'tcx>(
```
oli-obk, line 465: The complexity of this function seems ~okay to me, but I'm still wondering if the (discussed and dismissed in the RFC) marker trait (`Suspend`?) approach would work better. The attribute could impl `!Suspend` for structs, add it as a supertrait for traits and then the one caller to `check_must_not_suspend_ty` could make the InferCtxt add an obligation for `Suspend` to exist.

This would require some extra logic to make that obligation failure a lint instead of an error, so if we don't already have something like this, we should keep doing what we're doing here. Maybe just add the above paragraph as a comment for why we are implementing it this way?

oli-obk:
One additional test that I could think of would be a generic function that holds a generic argument over an await point and a caller to that generic function that uses a must_not_suspend type for the generic argument. Obviously this can't lint within the generic function, but we could probably inspect the generator created by the generic async function to see whether it stores said argument or whether it just uses it before its first resume.

basically

```rust
#[must_not_suspend]
struct No;

async fn shushspend() {}

async fn wheeee<T>(t: T) {
    shushspend().await;
    drop(t);
}

async fn yes() {
    wheeee(No).await;
}
```

compiler/rustc_typeck/src/check/generator_interior.rs
guswynn: Is the choice basically "we decided against this in the RFC"?

oli-obk: it was never discussed in the RFC except for an honorable mention [here](https://github.com/rust-lang/rfcs/blob/master/text/3014-must-not-suspend-lint.md#auto-trait-vs-attribute)

guswynn:
@oli-obk interestingly your example actually triggers the lint fine (although, its not perfectly formatted, it doubles up just like the reference case)

guswynn:
actually, no its triggering on the `whee` yield point, ill update the test to show whats going on

guswynn:
Added deduplication logic, currently at the hirid level, let me know what you think about what the ref test looks like now (It triggers now only on the value being borrowed, not the reference variable

oli-obk:
> 
> 
> actually, no its triggering on the `whee` yield point, ill update the test to show whats going on

yea, I expected it to trigger in `yes`, while pointing into wheeee with secondary spans can be useful, that function may be in a different crate, so we should not put too much effort into things

src/test/ui/lint/must_not_suspend/dedup.stderr
```diff
@@ -0,0 +1,19 @@
...
+note: the lint level is defined here
+  --> $DIR/dedup.rs:3:9
+   |
+LL | #![deny(must_not_suspend)]
+   |         ^^^^^^^^^^^^^^^^
+help: consider using a block (`{ ... }`) to shrink the value's scope, ending before the suspend point
+  --> $DIR/dedup.rs:16:12
+   |
+LL |     wheeee(No {}).await;
+   |            ^^^^^
```
oli-obk, line 16: this only makes sense where the value is held locally. If it was found in anther generator, then we should probably not emit this help

guswynn: this is triggering on the `No { }` value being held across the `whee` yield point, I think? I believe that @eholk is improving this analysis

nayato:
> @eholk good point, it looks like `drop` doesn't work, but the scope does, so I'll narrow the suggestion for now

Does that mean that warning should not fire when `drop` is used?
I am seeing on current nightly that if Ref dropped warning for `must_not_suspend` is issued.

PR #88868, opened 2021-09-11
Allow simd_bitmask to return byte arrays

calebzulawski:
cc @rust-lang/project-portable-simd @workingjubilee

### DIFF
--- a/compiler/rustc_codegen_llvm/src/intrinsic.rs
+++ b/compiler/rustc_codegen_llvm/src/intrinsic.rs
@@ -1072,6 +1072,9 @@ fn generic_simd_intrinsic(
         // * an unsigned integer
         // * an array of `u8`
         // If the vector has less than 8 lanes, a u8 is returned with zeroed trailing bits.
+        //
+        // The bit order of the result depends on the byte endianness, LSB-first for little
+        // endian and MSB-first for big endian.
         let expected_int_bits = in_len.max(8);
         let expected_bytes = expected_int_bits / 8 + ((expected_int_bits % 8 > 0) as u64);
 
--- a/compiler/rustc_codegen_llvm/src/intrinsic.rs
+++ b/compiler/rustc_codegen_llvm/src/intrinsic.rs
@@ -857,28 +857,39 @@ fn generic_simd_intrinsic(
     let arg_tys = sig.inputs();
 
     if name == sym::simd_select_bitmask {
-        let in_ty = arg_tys[0];
-        let m_len = match in_ty.kind() {
-            // Note that this `.unwrap()` crashes for isize/usize, that's sort
-            // of intentional as there's not currently a use case for that.
-            ty::Int(i) => i.bit_width().unwrap(),
-            ty::Uint(i) => i.bit_width().unwrap(),
-            _ => return_error!("`{}` is not an integral type", in_ty),
-        };
         require_simd!(arg_tys[1], "argument");
-        let (v_len, _) = arg_tys[1].simd_size_and_type(bx.tcx());
-        require!(
-            // Allow masks for vectors with fewer than 8 elements to be
-            // represented with a u8 or i8.
-            m_len == v_len || (m_len == 8 && v_len < 8),
-            "mismatched lengths: mask length `{}` != other vector length `{}`",
-            m_len,
-            v_len
-        );
+        let (len, _) = arg_tys[1].simd_size_and_type(bx.tcx());
+
+        let expected_int_bits = (len.max(8) - 1).next_power_of_two();
+        let expected_bytes = len / 8 + ((len % 8 > 1) as u64);
+
+        let mask_ty = arg_tys[0];
+        let mask = match mask_ty.kind() {
+            ty::Int(i) if i.bit_width() == Some(expected_int_bits) => args[0].immediate(),
+            ty::Uint(i) if i.bit_width() == Some(expected_int_bits) => args[0].immediate(),
+            ty::Array(elem, len)
+                if matches!(elem.kind(), ty::Uint(ty::UintTy::U8))
+                    && len.try_eval_usize(bx.tcx, ty::ParamEnv::reveal_all())
+                        == Some(expected_bytes) =>
+            {
+                let place = PlaceRef::alloca(bx, args[0].layout);
+                args[0].val.store(bx, place);
+                let int_ty = bx.type_ix(expected_bytes * 8);
+                let ptr = bx.pointercast(place.llval, bx.cx.type_ptr_to(int_ty));
+                bx.load(int_ty, ptr, Align::ONE)
+            }
+            _ => return_error!(
+                "invalid bitmask `{}`, expected `u{}` or `[u8; {}]`",
+                mask_ty,
+                expected_int_bits,
+                expected_bytes
+            ),
+        };
+
         let i1 = bx.type_i1();
-        let im = bx.type_ix(v_len);
-        let i1xn = bx.type_vector(i1, v_len);
-        let m_im = bx.trunc(args[0].immediate(), im);
+        let im = bx.type_ix(len);
+        let i1xn = bx.type_vector(i1, len);
+        let m_im = bx.trunc(mask, im);
         let m_i1s = bx.bitcast(m_im, i1xn);
         return Ok(bx.select(m_i1s, args[1].immediate(), args[2].immediate()));
     }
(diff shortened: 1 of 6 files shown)


compiler/rustc_codegen_llvm/src/intrinsic.rs
```diff
@@ -1067,8 +1064,34 @@ fn generic_simd_intrinsic(
...
+                if matches!(elem.kind(), ty::Uint(ty::UintTy::U8))
+                    && len.try_eval_usize(bx.tcx, ty::ParamEnv::reveal_all())
+                        == Some(expected_bytes) =>
+            {
+                // Zero-extend iN to the array lengh:
+                let ze = bx.zext(i_, bx.type_ix(expected_bytes * 8));
+
+                // Convert the integer to a byte array
+                let ptr = bx.alloca(bx.type_ix(expected_bytes * 8), Align::ONE);
+                bx.store(ze, ptr, Align::ONE);
```
calebzulawski, line 1083: For anyone more familiar with this than me: is there any reason this alignment or the one below must be greater than a byte?

calebzulawski:
Forgot about `simd_select_bitmask`--just updated that too.

src/test/ui/simd/simd-bitmask.rs
```diff
@@ -0,0 +1,51 @@
...
+        assert_eq!(r.0, e);
+
+        let a = Simd::<i32, 16>([0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15]);
+        let b = Simd::<i32, 16>([16, 17, 18, 19, 20, 21, 22, 23, 24, 25, 26, 27, 28, 29, 30, 31]);
+        let e = [16, 17, 2, 3, 20, 21, 22, 23, 24, 25, 26, 27, 12, 29, 14, 31];
+
+        let r = simd_select_bitmask(0b0101000000001100u16, a, b);
+        assert_eq!(r.0, e);
+
+        let r = simd_select_bitmask([0b1100u8, 0b01010000u8], a, b);
```
programmerjake, line 48: won't this break on big-endian?

calebzulawski: I'm not exactly sure how LLVM handles the bitcast.  We've actually seen the opposite issue (bit order, not byte order) on BE: https://github.com/rust-lang/portable-simd/blob/master/crates/core_simd/src/masks/full_masks.rs#L118-L125

programmerjake: bitcasts in llvm are supposed to be semantically equivalent to a memcpy iirc

programmerjake: try running tests on power64 (not power64le), it'll likely break

programmerjake: Additional conversation: https://rust-lang.zulipchat.com/#narrow/stream/257879-project-portable-simd/topic/Meeting.202021-09-13/near/253161847

compiler/rustc_codegen_llvm/src/intrinsic.rs
workingjubilee: Well, when these are handled by SSE or AVX512 instructions, they are normal integers of 16 to 64 bits. Even though we are saying we want to **interpret** this as a byte array, we should probably use the natural word-alignment of the processor: that way, `kmovq mem, reg` (reg to mem) naturally hits the alignment to pick it up in a GPR correctly, and if we for some reason are executing AVX512 features on a 32-bit processor, we still are correctly aligned for such (since then it would go into two GPRs).

src/test/ui/simd/simd-bitmask.rs
workingjubilee: Inlined from conversation, relevant snippets from @programmerjake:
> iirc the issue is that architectures have both a bit-level endianness and a byte-level endianness, both of which can independently vary
> install qemu-user and use that to run tests† 

†  for ppc64

I am not as sure that bit-endianness should be a problem, however, per the [LLVM LangRef: Vector Type]:
> In general vector elements are laid out in memory in the same way as array types. Such an analogy works fine as long as the vector elements are byte sized. However, when the elements of the vector aren’t byte sized it gets a bit more complicated. One way to describe the layout is by describing what happens when a vector such as <N x iM> is bitcasted to an integer type with N*M bits, and then following the rules for storing such an integer to memory.

I think the greater problem might be that [it is not clear the i1 to integer bitcast was ever well-formed][SO#61647221].

That said, the [LLVM LangRef on bitcasts] also expounds on vectors:
> There is a caveat for bitcasts involving vector types in relation to endianess. For example bitcast <2 x i8> <value> to i16 puts element zero of the vector in the least significant bits of the i16 for little-endian while element zero ends up in the most significant bits for big-endian.

[LLVM LangRef: Vector Type]: https://llvm.org/docs/LangRef.html#vector-type
[SO#61647221]: https://stackoverflow.com/questions/61647221/is-llvm-bitcast-from-vector-of-bool-i1-to-i8-i16-etc-well-defined
[LLVM LangRef on bitcasts]: https://llvm.org/docs/LangRef.html#bitcast-to-instruction

workingjubilee (commented):
r? @workingjubilee

workingjubilee:
odd.
r? @workingjubilee

src/test/ui/simd/simd-bitmask.rs
workingjubilee: Hmm, I found a comment in a test that completely confirms @programmerjake's remarks, actually:
> // ignore-endian-big behavior of simd_select_bitmask is endian-specific

🙃

compiler/rustc_codegen_llvm/src/intrinsic.rs
calebzulawski: I don't think anything is preventing LLVM from using a greater alignment if it's more efficient, but the actual alignment is just 1, right?  Regular `u8` arrays are definitely used with larger alignments in most situations, but their alignment is still 1.

src/test/ui/simd/simd-bitmask.rs
calebzulawski: I just copied this to the test.  We can work out the endianness (and add tests if necessary) in another issue.

calebzulawski:
Not sure how to change it to waiting for review 🤷‍♂️

compiler/rustc_codegen_llvm/src/intrinsic.rs
eddyb: As long as you're coherent about this (I believe `PlaceRef` tracks it for you, so it may be worth trying to use it more), you can increase the alignment of the `alloca` (and then pass that alignment to the `store`/`load` - whether manually, or relying on the `PlaceRef` API, or letting LLVM infer it).

I'm not sure how useful it is here, but it may make LLVM more likely to generate the aligned versions of instructions that have both aligned and unaligned variants (assuming LLVM doesn't rewrite this cast away).

Also, if you're using `alloca`s, please consider using `place.storage_{live,dead}()` before/after all of the memory operations on the `alloca` - again I'm unsure whether this is significant in the context (since e.g. a small wrapper that gets inlined will do the right thing itself during LLVM inlining), but the default without those (i.e. without `llvm.lifetime.{start,end}`) is every `alloca` adds to the stack frame size (for the entire function).

workingjubilee:
Hm.
I have thought about this and I am fine proceeding without an explicit test for big-endian, since we may as well be testing that issue in our high level API instead, but I would appreciate it if you documented that for `simd_bitmask`
- currently, all tier 1 (and thus "little endian") platforms emit a predictable representation
- but the exact representation for other platforms, especially big endian ones, is less clear, thus
- in any case, the repr is platform-dependent as a result, and
- should only be considered consistent with `simd_select_bitmask`

somewhere in either the associated tests or the codegen infrastructure itself.

compiler/rustc_codegen_llvm/src/intrinsic.rs
workingjubilee: Hmmm. At least hypothetically this should in fact be "a small wrapper that gets inlined". I don't think either of us is familiar with the PlaceRef API tho'. But that shouldn't be a blocking concern.

workingjubilee:
I gave it a spin on `std::arch`'s codegen tests and it seems fine. Let's see how it plays against CI.
@bors r+

PR #88875, opened 2021-09-12
cleanup(rustc_trait_selection): remove vestigial code from rustc_on_unimplemented

notriddle:
This isn't allowed by the validator, and seems to be unused.
When it was added in ed10a3faae1fd1176b2edf4a61438e0542c103b9,
it was used on `Sized`, and that usage is gone.


Mark-Simulacrum:
@bors r+ rollup

Thanks!

Issue #85247, opened 2021-05-13
rustc is too strict about reserved registers on ARM


Skirmisher:
Recently, [this commit](https://github.com/rust-lang/rust/commit/e6a731eb90fe3d47d89416e199832af4248399f6#diff-8818a2f974f5c898a567797bb58287139c8a4256ba8344e1e63eb556ac06bc82) came to my attention, as it added r9 to the list of LLVM-reserved registers on ARM according to rustc, breaking existing inline asm. I decided to look into this further, and found [this page](https://developer.arm.com/documentation/101754/0615/armclang-Reference/Compiler-specific-Keywords-and-Operators/Global-named-register-variables) (section "Restrictions") which concisely states under what conditions certain registers are reserved in the ABI (with regards to armclang as described in the page, but it applies more generally). The relevant points are:

- If RWPI (read-write position independent) code is enabled, R9 is reserved.
- If needed, R7 is reserved for the frame pointer in T32 mode, and R11 is reserved in A32 mode.

However, rustc reserves these registers unconditionally. This seems incorrect, as both RWPI and frame pointers can be disabled via codegen options or a custom target. In our case, the custom target file (`armv4t-none-eabi.json`) has both `"eliminate-frame-pointer": true` and `"relocation-model": "static"` set, but rustc will reject uses of r9 and r11 with their respective error messages ("r9 is used internally by llvm" and "frame pointer (r11) cannot be used as an operand").

Lokathor:
The `gba` crate (`armv4t-none-eabi` target) also got hit with this.

Skirmisher:
Okay, so the codegen backend situation has changed a lot since this target file was first generated, and `eliminate-frame-pointer` was [replaced](https://github.com/rust-lang/rust/commit/9b67cba4f6294e0db900397cdf5c59c1dc901ade#diff-aa810a3be0834da891b171a8b04b09d0d3bbb76ac27c757cd232235099e062d2) a while back with `frame-pointer` which defaults to `MayOmit`—no longer implying a guarantee, mainly because a different codegen backend can make arbitrary decisions about the frame pointer (and it's unclear how or whether this setting eliminated the frame pointer in LLVM in the first place). I was not aware of the codegen backend abstraction work or anything else when filing this issue.

[...]

PR #88879, opened 2021-09-12
asm/arm: allow r9 when usable, make diagnostics more specific

Skirmisher:
As described in #85247, there are a number of cases where inline ARM assembly in Rust programs is prevented from using certain registers. Sometimes these registers are only reserved in certain situations; however, they are all unconditionally restricted by the compiler. Many of rustc's diagnostics describe the registers as being "used internally by LLVM", which obfuscates the meaning of the registers and is ultimately unhelpful to low-level programmers.

At the moment, it does not seem possible to ensure that the frame pointer or base pointer will be elided, or to understand the conditions under which this may occur. Some of this is due to the addition of a codegen abstraction layer, where these behaviors may vary between backends and may not be configurable. As such, this information is not available to Rust's assembly parser.

However, register r9 has a condition that is rather easily checked: the relocation model. r9 is used as the "static base" (SB) for "read-write position independence" (RWPI) as a base from which to address writable data, as defined in the [Procedure Call Standard for the Arm Architecture][1]. The register is [always available for use if RWPI is not used][2].

Rust allows RWPI mode to be selected on a per-target basis, as well as making it available as a codegen option in rustc. Thus, we can predicate usage of r9 on this setting with confidence.

Allow r9 to be used in `asm!` blocks when the target does not use RWPI, and define what r6 and r9 are reserved for in their respective diagnostics. The frame pointer (r7/r11) and base pointer (r6) remain unconditionally restricted for now.

[1]: https://developer.arm.com/documentation/ihi0042/j/
[2]: https://developer.arm.com/documentation/dui0056/d/using-the-procedure-call-standard/read-write-position-independence/register-usage-with-rwpi

### DIFF
--- a/compiler/rustc_target/src/asm/arm.rs
+++ b/compiler/rustc_target/src/asm/arm.rs
@@ -1,5 +1,5 @@
 use super::{InlineAsmArch, InlineAsmType};
-use crate::spec::Target;
+use crate::spec::{RelocModel, Target};
 use rustc_macros::HashStable_Generic;
 use std::fmt;
 
@@ -88,6 +88,18 @@ fn frame_pointer_r7(
     }
 }
 
+fn rwpi_enabled(
+    _arch: InlineAsmArch,
+    _has_feature: impl FnMut(&str) -> bool,
+    target: &Target,
+) -> Result<(), &'static str> {
+    if let RelocModel::Rwpi | RelocModel::RopiRwpi = target.relocation_model {
+        Err("the RWPI static base register (r9) cannot be used as an operand for inline asm")
+    } else {
+        Ok(())
+    }
+}
+
 def_regs! {
     Arm ArmInlineAsmReg ArmInlineAsmRegClass {
         r0: reg, reg_thumb = ["r0", "a1"],
@@ -98,6 +110,7 @@ def_regs! {
         r5: reg, reg_thumb = ["r5", "v2"],
         r7: reg, reg_thumb = ["r7", "v4"] % frame_pointer_r7,
         r8: reg = ["r8", "v5"],
+        r9: reg = ["r9", "v6", "rfp"] % rwpi_enabled,
         r10: reg = ["r10", "sl"],
         r11: reg = ["r11", "fp"] % frame_pointer_r11,
         r12: reg = ["r12", "ip"],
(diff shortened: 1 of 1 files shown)


Amanieu:
Technically we should be checking for the `reserve-r9` LLVM feature here. But it's not currently exposed by rustc.

Skirmisher:
That is more precise, but that will still leave out other codegen backends, no?

compiler/rustc_target/src/asm/arm.rs
```diff
@@ -88,6 +88,18 @@ fn frame_pointer_r7(
     }
 }
 
+fn rwpi_enabled(
+    _arch: InlineAsmArch,
+    _has_feature: impl FnMut(&str) -> bool,
+    target: &Target,
+) -> Result<(), &'static str> {
+    if let RelocModel::Rwpi | RelocModel::RopiRwpi = target.relocation_model {
```
bjorn3, line 96: Should we only allow r9 when the relocation model was explicitly given to allow for more freedom in choosing which relocation model will be used by default without risking breaking this?

Skirmisher: That might be a good idea, though it would make the diagnostic less clear if the user hasn't specifically opted in to that mode. I wanted to add a "note" to it, but this is my first time contributing to rustc and I don't know how to do that, much less how to make it work in the context of `def_regs!`.

How does one check if a target attribute is explicitly set?

bjorn3: You can check the value of `sess.opts.cg.relocation_model`. By the way I think `target.relocation_model` is only the default relocation model in all cases and not the one that will actually be used when compiling if `-Crelocation-model` is used.

Skirmisher: Good catch, I didn't think of that since in my use case I'm using a custom target. I can fix that up. I'd still like to improve the diagnostics though.

Skirmisher: Oh, `Session` already has a `relocation_model` getter that does the logic I was going to write anyway.

Skirmisher: Ah right, I can't actually do that because the asm parser isn't passed the `Session`, only the `Target`... Looks like I would need to change the definition of `parse` in the entire `rustc_target::asm` module at least, including arch-specific filter functions, then alter `rustc_ast_lowering::LoweringContext` to pass `Session` as well. That feels like a fairly invasive change just to make a small thing more correct, but the overall effect should be negligible. If the maintainers are okay with it, I can go ahead and do it.

Amanieu: This won't work since `rustc_target` can't depend on `rustc_session` since that would create a circular dependency.

pnkfelix:
Hi @Amanieu ; do you think you'll be able to review this PR, or delegate it to some other expert on ARM `asm!` ?

Amanieu:
I still think this should be changed to look for the `reserve-r9` target feature since that will automatically pick up the case where the relocation model is changed on the command-line. Also, if we're going to do this then we should probably do the same with x18 on AArch64 which is reserved on some targets, but it's fine to leave this to a later PR.

Finally, the documentation in the unstable book needs to be updated to reflect the new behavior of r9.

Amanieu:
Closing in favor of #91643.

PR #88883, opened 2021-09-12
Move some tests to more reasonable directories - 7

c410-f3r:
cc #73494
r? @petrochenkov


petrochenkov:
issue-5791.rs -> can be removed, already tested in clashing-extern-fn.rs
expr-empty-ret.rs -> never_type
guards.rs -> match
issue-33498.rs -> match
enable-unstable-lib-feature.rs -> stability-attribute
backtrace-debuginfo.rs -> certainly not macros, not sure about the proper subdirectory, let's keep it in its old place
issue-41255.rs -> lint or match
issue-79744.rs -> lint
out-of-order-shadowing.rs -> macros
issue-78720.rs -> closures
use-self-in-inner-fn.rs -> resolve

(`binding` and `use` are confirmed to be garbage subdirectories)

Could you also update the test number limits in tidy?

c410-f3r:
Done

c410-f3r:
`enable-unstable-lib-feature`, `lint-stability` and `list-stability-deprecated` are all placed inside `lint` because of shared `aux-build:stability-cfg2.rs`.

PR #88885, opened 2021-09-12
Fix jump def background

### LINKED ISSUE #88870: rustdoc (unstable): `--generate-link-to-definition` makes sources unreadable with Ayu theme
Interestingly, the dark theme works fine.

Screenshot of Ayu:

image

cc @GuillaumeGomez

GuillaumeGomez:
Fixes #88870.

I somehow badly wrote the color in #88111.

r? @camelid


src/test/rustdoc-gui/jump-to-def-background.goml
```diff
@@ -0,0 +1,23 @@
...
+
+// Set the theme to dark.
+local-storage: {"rustdoc-theme": "dark", "rustdoc-preferred-dark-theme": "dark", "rustdoc-use-system-theme": "false"}
+// We reload the page so the local storage settings are being used.
+reload:
+
+assert-css: ("body.source .example-wrap pre.rust a", {"background-color": "rgb(51, 51, 51)"}, ALL)
+
+// Set the theme to ayu.
+local-storage: {"rustdoc-theme": "ayu", "rustdoc-preferred-dark-theme": "ayu", "rustdoc-use-system-theme": "false"}
```
camelid, line 12: Not specific to this PR, but I'm curious: is it necessary to set the preferred dark theme given that it says to ignore the system theme and manually specifies the current theme?

src/bootstrap/test.rs
```diff
@@ -924,6 +924,7 @@ impl Step for RustdocGUI {
                     .arg(&out_dir)
                     .env("RUSTDOC", builder.rustdoc(self.compiler))
                     .env("RUSTC", builder.rustc(self.compiler))
+                    .env("RUSTDOCFLAGS", "-Zunstable-options --generate-link-to-definition")
```
camelid, line 927: Seems somewhat unfortunate to enable `--generate-link-to-definition` for *all* GUI tests. I guess it could have a `compile-flags` comment like compiletest does, but that would be a larger change.

src/test/rustdoc-gui/code-sidebar-toggle.goml
```diff
@@ -1,5 +1,6 @@
 goto: file://|DOC_PATH|/test_docs/index.html
 click: ".srclink"
+wait-for: "#sidebar-toggle"
```
camelid, line 3: Why is this change being made?

camelid (commented):
By the way, one of the GUI tests seems to be failing.

src/test/rustdoc-gui/jump-to-def-background.goml
GuillaumeGomez: Not sure it is, but just in case... :)

src/bootstrap/test.rs
GuillaumeGomez: Actually I think I'll enable it only for one test to simplify things.

src/test/rustdoc-gui/code-sidebar-toggle.goml
GuillaumeGomez: In some very rare cases, the `click` command just below was run before the content was there I think, making the `click` fail because of the missing element.

src/bootstrap/test.rs
```diff
@@ -925,6 +925,9 @@ impl Step for RustdocGUI {
                     .env("RUSTDOC", builder.rustdoc(self.compiler))
                     .env("RUSTC", builder.rustc(self.compiler))
                     .current_dir(path);
+                if entry.file_name() == "test_docs" {
+                    cargo.env("RUSTDOCFLAGS", "-Zunstable-options --generate-link-to-definition");
+                }
```
camelid, line 930: Could you instead add a new test called something like `link_to_definition` and enable `--generate-link-to-definition` only for that? Enabling `--generate-link-to-definition` for `test_docs` in particular is surprising because there's nothing in the name that would make you think that.

src/test/rustdoc-gui/code-sidebar-toggle.goml
camelid: But why did you make the change in this PR? Could you open a separate PR for this change since it seems to be unrelated?

src/test/rustdoc-gui/jump-to-def-background.goml
camelid: Ok, that sounds fine 👍

src/bootstrap/test.rs
GuillaumeGomez: Ok!

GuillaumeGomez:
> But why did you make the change in this PR? Could you open a separate PR for this change since it seems to be unrelated?

I have put it in a commit on its own but let's make another PR for it.

GuillaumeGomez:
Updated!

src/test/rustdoc-gui/src/link_to_definition/Cargo.toml
```diff
@@ -0,0 +1,7 @@
+[package]
+name = "link_to_definition"
+version = "0.1.0"
+edition = "2018"
+
+[lib]
+path = "lib.rs"
```
camelid, line 7: There's no dependency on `implementors` in this `Cargo.toml`. I think you can just remove the `implementors` crate since it's not necessary for what we're testing?

src/test/rustdoc-gui/jump-to-def-background.goml
```diff
@@ -0,0 +1,23 @@
+// We check the background color on the jump to definition links in the source code page.
+goto: file://|DOC_PATH|/src/link_to_definition/lib.rs.html
+
+// Set the theme to dark.
+local-storage: {"rustdoc-theme": "dark", "rustdoc-preferred-dark-theme": "dark", "rustdoc-use-system-theme": "false"}
+// We reload the page so the local storage settings are being used.
+reload:
+
+assert-css: ("body.source .example-wrap pre.rust a", {"background-color": "rgb(51, 51, 51)"}, ALL)
```
camelid, line 9: It'd probably be a good idea to add a check that there are any links at all, just in case the bootstrap check isn't working.

src/test/rustdoc-gui/src/link_to_definition/Cargo.toml
GuillaumeGomez: Oh damn, I added too much without paying attention, removed it!

src/test/rustdoc-gui/jump-to-def-background.goml
GuillaumeGomez: It fails if no element matches the given selector. :)

GuillaumeGomez:
Updated!

src/test/rustdoc-gui/jump-to-def-background.goml
camelid: Ah, that makes sense 👍

src/bootstrap/test.rs
```diff
@@ -925,6 +925,9 @@ impl Step for RustdocGUI {
                     .env("RUSTDOC", builder.rustdoc(self.compiler))
                     .env("RUSTC", builder.rustc(self.compiler))
                     .current_dir(path);
+                if entry.file_name() == "link_to_definition" {
```
camelid, line 928: Could you add a FIXME here that says this?

```rust
// FIXME: implement a `// compile-flags` command or similar
//        instead of hard-coding this test
```

src/test/rustdoc-gui/src/link_to_definition/lib.rs
```diff
@@ -0,0 +1,6 @@
+pub struct Bar {
+    a: String,
+    b: u32,
```
camelid, line 3: Could you make these fields public to add more test coverage?

camelid (commented):
r=me after addressing these comments and squashing the second and third commits together (it's confusing to have them be separate)

GuillaumeGomez:
Updated!

camelid:
Could you squash the second and third commits together? It's confusing to have them be separate since neither stands alone.

r=me after that

GuillaumeGomez:
I always keep the tests in a different commit but I can merge the second and third together.

camelid:
> I always keep the tests in a different commit but I can merge the second and third together.

The reason I think it's confusing is because in the first commit you were enabling `--generate-link-to-definition` on the test before the test even existed. Even if the commits were reversed, the commit with the test would fail because `--generate-link-to-definition` wouldn't be passed.

GuillaumeGomez:
Good point. :)

PR #88886, opened 2021-09-12
Encode span parent in the inlined representation.

cjgillot:
Follow-up to #84373 attempting to store `SpanData::parent` in the inline `Span` implementation.


michaelwoerister:
Sorry, wrong PR `:)`

r? @michaelwoerister

michaelwoerister:
Is this PR still active? It does not seem to give the performance improvements we hoped for.

JohnCSimon:
@cjgillot - ping from triage:
> Is this PR still active? It does not seem to give the performance improvements we hoped for.

PR #88887, opened 2021-09-12
Const Deref

fee1-dead:
Implements `const Deref`/`const DerefMut` for `&mut T`, `&T`, `Cow<'_, B>` and `ManuallyDrop<T>`

### DIFF
--- a/library/alloc/src/borrow.rs
+++ b/library/alloc/src/borrow.rs
@@ -330,7 +330,11 @@ impl<B: ?Sized + ToOwned> Cow<'_, B> {
 }
 
 #[stable(feature = "rust1", since = "1.0.0")]
-impl<B: ?Sized + ToOwned> Deref for Cow<'_, B> {
+#[rustc_const_unstable(feature = "const_deref", issue = "88955")]
+impl<B: ?Sized + ToOwned> const Deref for Cow<'_, B>
+where
+    B::Owned: ~const Borrow<B>,
+{
     type Target = B;
 
     fn deref(&self) -> &B {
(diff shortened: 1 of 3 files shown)


library/alloc/src/string.rs
```diff
@@ -2298,7 +2298,8 @@ impl ops::IndexMut<ops::RangeToInclusive<usize>> for String {
 }
 
 #[stable(feature = "rust1", since = "1.0.0")]
-impl ops::Deref for String {
+#[rustc_const_unstable(feature = "const_deref", issue = "none")]
+impl const ops::Deref for String {
```
oli-obk, line 2302: We should not make this (and `Vec`'s impl) `const`, out of the same reason we didn't make `len` `const fn`: there is no way to construct any interesting (non-empty) `String`s, so all these functions are useless. It was decided that, until we have some sort of const heap, we should not add any new const functions beyond constructors.

library/core/src/panic/unwind_safe.rs
```diff
@@ -248,6 +248,7 @@ impl RefUnwindSafe for crate::sync::atomic::AtomicBool {}
 impl<T> RefUnwindSafe for crate::sync::atomic::AtomicPtr<T> {}
 
 #[stable(feature = "catch_unwind", since = "1.9.0")]
+#[rustc_const_unstable(feature = "const_deref", issue = "none")]
 impl<T> Deref for AssertUnwindSafe<T> {
```
oli-obk, line 252: These are never constructed in const contexts, so with the same reasoning as for String/Vec, we should not add const methods or impls on them.

oli-obk (changes requested):
Please create a tracking issue or point to the const trait impls tracking issue.

Also include a short message in the main post explaining what this PR does.

I think this PR would also benefit from a test or two

jhpratt:
Does `#[rustc_const_unstable]` actually work on trait impls? Only asking because I know regular unstable attributes don't.

oli-obk:
> 
> 
> Does `#[rustc_const_unstable]` actually work on trait impls? Only asking because I know regular unstable attributes don't.

yes, and there are explicit tests for this

PR #88548, opened 2021-08-31
Stabilize `Iterator::intersperse()`

inquisitivecrystal:
This PR stabilizes the methods `Iterator::intersperse()` and `Iterator::intersperse_with()`. The FCP has [already completed](https://github.com/rust-lang/rust/issues/79524#issuecomment-909663616).

Closes #79524.

PR #88250, opened 2021-08-23
Make `-Z gcc-ld=lld` work for Apple targets

hkratz:
`-Z gcc-ld=lld` was introduced in #85961. It does not work on Macos because lld needs be either named `ld64` or passed `-flavor darwin` as the first two arguments in order to select the Mach-O flavor. Rust invokes cc (=clang) on Macos for linking which calls `ld` as linker binary and not `ld64`, so just creating an `ld64` binary and modifying the search path with `-B` does not work.

In order to solve this patch does:
* Set the `lld_flavor` for all Apple-derived targets to `LldFlavor::Ld64`. As far as I can see this actually works towards fixing `-Xlinker=rust-lld` as all those targets use the Mach-O object format.
* Copy/hardlink rust-lld to the gcc-ld subdirectory as ld64 next to ld.
* If `-Z gcc-ld=lld` is used and the target lld flavor is Ld64 add `-fuse-ld=/path/to/ld64` to the linker invocation.

Fixes #86945.

PR #88893, opened 2021-09-12
Add 1.56.0 release notes

Mark-Simulacrum:
Notable things:

* Rustdoc section currently absent, for lack of things to call out that I could find. It seems like https://github.com/rust-lang/rust/pull/87451 is the only potential candidate, but that seems like more of a bugfix and doesn't seem to warrant inclusion to me. But we can add it if desired.

As with the 1.55.0 release notes, my intent is to leave this open for approximately 1-2 weeks at the early part of the cycle, and then merge it in (after fixing any feedback). Further iteration can happen in subsequent issues (and PRs).

[Rendered](https://github.com/Mark-Simulacrum/rust/blob/relnotes/RELEASES.md)

### DIFF
--- a/RELEASES.md
+++ b/RELEASES.md
@@ -1,3 +1,188 @@
+Rust 1.56.0 (2021-10-21)
+========================
+
+Language
+--------
+
+- [The 2021 Edition is now stable.][rust#88100]
+  See [the edition guide][rust-2021-edition-guide] for more details.
+- [You can now specify explicit discriminant values on any Rust enum.][rust#86860]
+- [The pattern in `binding @ pattern` can now also introduce new bindings.][rust#85305]
+- [Union field access is permitted in `const fn`.][rust#85769]
+
+[rust-2021-edition-guide]: https://doc.rust-lang.org/nightly/edition-guide/rust-2021/index.html
+
+Compiler
+--------
+
+- [Upgrade to LLVM 13.][rust#87570]
+- [Support memory, address, and thread sanitizers on aarch64-unknown-freebsd.][rust#88023]
+- [Allow specifying an deployment target version for all iOS targets][rust#87699]
+- [Warnings can be forced on with `--force-warn`.][rust#87472]
+  This feature is primarily intended for usage by `cargo fix`, rather than end users.
+- [Promote `aarch64-apple-ios-sim` to Tier 2\*.][rust#87760]
+- [Add `powerpc-unknown-freebsd` at Tier 3\*.][rust#87370]
+
+\* Refer to Rust's [platform support page][platform-support-doc] for more
+information on Rust's tiered platform support.
+
+Libraries
+---------
+
+- [Allow writing of incomplete UTF-8 sequences via stdout/stderr on Windows.][rust#83342]
+  The Windows console still requires valid Unicode, but this change allows
+  splitting a UTF-8 character across multiple write calls. This allows, for
+  instance, programs that just read and write data buffers (e.g. copying a file
+  to stdout) without regard for Unicode or character boundaries.
+- [Prefer `AtomicU{64,128}` over Mutex for Instant backsliding protection.][rust#83093]
+  For this use case, atomics scale much better under contention.
+- [Implement `Extend<(A, B)>` for `(Extend<A>, Extend<B>)`][rust#85835]
+- [impl Default, Copy, Clone for std::io::Sink and std::io::Empty][rust#86744]
+- [`impl From<[(K, V); N]>` for all collections.][rust#84111]
+- [Remove `P: Unpin` bound on impl Future for Pin.][rust#81363]
+- [Treat invalid environment variable names as non-existent.][rust#86183]
+  Previously, the environment functions would panic if given a variable name
+  with an internal null character or equal sign (`=`). Now, these functions will
+  just treat such names as non-existent variables, since the OS cannot represent
+  the existence of a variable with such a name.
+
+Stabilised APIs
+---------------
+
+- [`std::os::unix::fs::chroot`]
+- [`Iterator::intersperse`]
+- [`Iterator::intersperse_with`]
+- [`UnsafeCell::raw_get`]
+- [`BufWriter::into_parts`]
+- [`core::panic::{UnwindSafe, RefUnwindSafe, AssertUnwindSafe}`]
+  These APIs were previously stable in `std`, but are now also available in `core`.
+- [`Vec::shrink_to`]
+- [`String::shrink_to`]
+- [`OsString::shrink_to`]
+- [`PathBuf::shrink_to`]
+- [`BinaryHeap::shrink_to`]
+- [`VecDeque::shrink_to`]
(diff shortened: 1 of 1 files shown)


RELEASES.md
```diff
@@ -1,3 +1,167 @@
...
+- [impl Default, Copy, Clone for std::io::Sink and std::io::Empty][rust#86744]
+- [`impl From<[(K, V); N]>` for all collections.][rust#84111]
+- [Remove `P: Unpin` bound on impl Future for Pin.][rust#81363]
+- [Treat invalid environment variable names as non-existent.][rust#86183]
+
+Stabilised APIs
+---------------
+
+- [`std::os::unix::fs::chroot`]
+- [`Iterator::intersperse`]
```
lukaslueg, line 45: #88548 also stabilized `Iterator::intersperse_with`

jyn514:
If we're including bigfixes, https://github.com/rust-lang/rust/pull/88019 is another good ones, it fixes a metric ton of issues with rustdoc and macros.

https://github.com/rust-lang/rust/pull/87728 and https://github.com/rust-lang/rust/pull/87050 might be good to mention in compatibility notes (not sure if they're major enough since it's only a warning).

Other than that I agree there's not a lot.

RELEASES.md
```diff
@@ -1,3 +1,167 @@
...
+- [`UnsafeCell::raw_get`]
+- [`BufWriter::into_parts`]
+- [`core::panic::{UnwindSafe, RefUnwindSafe, AssertUnwindSafe}`]
+  These APIs were previously stable in `std`, but are now also available in `core`.
```
jyn514, line 49: This can't be right, though - Vec is only defined `alloc`, isn't it? Do you mean to say "now available in alloc"?

tschuett:
I would love a `rendered` link. I would also enjoy https://github.com/rust-lang/rust/pull/88250. PGO builds for other targets would also be nice.

RELEASES.md
CryZe: Pretty sure it refers to the UnwindSafe stuff.

Mark-Simulacrum:
https://github.com/rust-lang/rust/pull/88250 is nightly-only, right? We don't mention nightly-only things in release notes.

RELEASES.md
```diff
@@ -1,3 +1,167 @@
...
+- [You can now specify explicit discriminant values on any Rust enum.][rust#86860]
+- [The pattern in `binding @ pattern` can now also introduce new bindings.][rust#85305]
+- [Union field access is permitted in `const fn`.][rust#85769]
+
+[rust-2021-edition-guide]: https://doc.rust-lang.org/nightly/edition-guide/rust-2021/index.html
+
+Compiler
+--------
+
+- [Upgrade to LLVM 13.][rust#87570]
```
kellerkindt, line 18: This links to the wrong ticket?

RELEASES.md
```diff
@@ -1,3 +1,167 @@
...
+- [Promote `aarch64-apple-ios-sim` to Tier 2\*.][rust#87760]
+- [Add `powerpc-unknown-freebsd` at Tier 3\*.][rust#87370]
+
+\* Refer to Rust's [platform support page][platform-support-doc] for more
+information on Rust's tiered platform support.
+
+Libraries
+---------
+
+- [Allow writing of incomplete UTF-8 sequences via stdout/stderr on Windows.][rust#83342]
```
joshtriplett, line 32:
```suggestion
- [Allow writing of incomplete UTF-8 sequences via stdout/stderr on Windows.][rust#83342]
  The Windows console still requires valid Unicode, but this change allows splitting a UTF-8 character across multiple write calls. This allows, for instance, programs that just read and write data buffers (e.g. copying a file to stdout) without regard for Unicode or character boundaries.
```

RELEASES.md
```diff
@@ -1,3 +1,167 @@
...
+- [Prefer `AtomicU{64,128}` over Mutex for Instant backsliding protection.][rust#83093]
+  For this use case, atomics scale much better under contention.
+- [Implement `Extend<(A, B)>` for `(Extend<A>, Extend<B>)`][rust#85835]
+- [impl Default, Copy, Clone for std::io::Sink and std::io::Empty][rust#86744]
+- [`impl From<[(K, V); N]>` for all collections.][rust#84111]
+- [Remove `P: Unpin` bound on impl Future for Pin.][rust#81363]
+- [Treat invalid environment variable names as non-existent.][rust#86183]
```
joshtriplett, line 39:
```suggestion
- [Treat invalid environment variable names as non-existent.][rust#86183]
  Previously, the environment functions would panic if given a variable name with an internal null character or equal sign (`=`). Now, these functions will just treat those names as non-existent variables, since the OS cannot represent the existence of a variable with such a name.
```

RELEASES.md
```diff
@@ -1,3 +1,167 @@
...
+- [`[T]::first`][`slice::first`]
+- [`[T]::split_first`][`slice::split_first`]
+- [`[T]::last`][`slice::last`]
+- [`[T]::split_last`][`slice::split_last`]
+
+Cargo
+-----
+
+- [Cargo supports specifying a minimum supported Rust version in Cargo.toml.][`rust-version`]
+  This has no effect at present on dependency version selection.
```
joshtriplett, line 72:
```suggestion
  This has no effect at present on dependency version selection.
  We encourage crates to specify their minimum supported Rust version, and we encourage CI systems that support Rust code to include a crate's specified minimum version in the text matrix for that crate by default.
```

RELEASES.md
```diff
@@ -1,3 +1,167 @@
...
+  This adjusts Rust's standard library to match the behavior of the standard
+  libraries for C/C++. The rules have changed slightly over time, and this PR
+  brings us to the latest set of rules (changed in 2008).
+- [Disallow the aapcs calling convention on aarch64][rust#88399]
+  This was already not supported by LLVM; this change surfaces this lack of
+  support with a better error message.
+- [Make `SEMICOLON_IN_EXPRESSIONS_FROM_MACROS` warn by default][rust#87385]
+- [Warn when an escaped newline skips multiple lines.][rust#87671]
+- [Calls to `libc::getpid` / `std::process::id` from `Command::pre_exec`
+   may return different values on glibc <= 2.24.][rust#81825]
```
joshtriplett, line 87:
```suggestion
   may return different values on glibc <= 2.24.][rust#81825]
   Rust now invokes the `clone3` system call directly, when available, to use new functionality available via that system call. Older versions of glibc cache the result of `getpid`, and only update that cache when calling glibc's clone/fork functions, so a direct system call bypasses that cache update. glibc 2.25 and newer no longer cache `getpid` for exactly this reason.
```

Mark-Simulacrum:
> #87728 and #87050 might be good to mention in compatibility notes (not sure if they're major enough since it's only a warning).

I intentionally didn't include things that are "obviously" made incompatible in this release with just warnings.

Mark-Simulacrum:
Thanks everyone for their review!

PR #88897, opened 2021-09-12
rustdoc: Make `Impl.trait_` a `Path`, not a `Type`

camelid:
Split out from #88379. So, this PR is also known as "Cleanup `clean` part 3".

It should only ever be a `ResolvedPath`, so this (a) enforces that, and
(b) reduces the size of `Impl`.

I had to update a test because the order of the rendered auto trait impl
bounds changed. I think the order changed because rustdoc sorts auto
trait bounds using their `Debug` output.

r? @jyn514


camelid:
Oops, actually this should probably be in a PR with the `PolyTrait` refactor commit and some other ones.

Issue #63291, opened 2019-08-05
Tracking issue for uninitialized constructors for Box, Rc, Arc


SimonSapin:
Assigning `MaybeUninit::<Foo>::uninit()` to a local variable is usually free, even when `size_of::<Foo>()` is large. However, passing it for example to `Arc::new` [causes at least one copy](https://youtu.be/F1AquroPfcI?t=4116) (from the stack to the newly allocated heap memory) even though there is no meaningful data. It is theoretically possible that a Sufficiently Advanced Compiler could optimize this copy away, but this is [reportedly unlikely to happen soon in LLVM](https://youtu.be/F1AquroPfcI?t=5431).

This issue tracks constructors for containers (`Box`, `Rc`, `Arc`) of `MaybeUninit<T>` or `[MaybeUninit<T>]` that do not initialized the data, and unsafe conversions to the known-initialized types (without `MaybeUninit`). The constructors are guaranteed not to make unnecessary copies.

PR https://github.com/rust-lang/rust/pull/62451 adds:

```rust
impl<T> Box<T> { pub fn new_uninit() -> Box<MaybeUninit<T>> {…} }
impl<T> Box<MaybeUninit<T>> { pub unsafe fn assume_init(self) -> Box<T> {…} }
impl<T> Box<[T]> { pub fn new_uninit_slice(len: usize) -> Box<[MaybeUninit<T>]> {…} }
impl<T> Box<[MaybeUninit<T>]> { pub unsafe fn assume_init(self) -> Box<[T]> {…} }

impl<T> Rc<T> { pub fn new_uninit() -> Rc<MaybeUninit<T>> {…} }
impl<T> Rc<MaybeUninit<T>> { pub unsafe fn assume_init(self) -> Rc<T> {…} }
impl<T> Rc<[T]> { pub fn new_uninit_slice(len: usize) -> Rc<[MaybeUninit<T>]> {…} }
impl<T> Rc<[MaybeUninit<T>]> { pub unsafe fn assume_init(self) -> Rc<[T]> {…} }

impl<T> Arc<T> { pub fn new_uninit() -> Arc<MaybeUninit<T>> {…} }
impl<T> Arc<MaybeUninit<T>> { pub unsafe fn assume_init(self) -> Arc<T> {…} }
impl<T> Arc<[T]> { pub fn new_uninit_slice(len: usize) -> Arc<[MaybeUninit<T>]> {…} }
[...]

PR #88906, opened 2021-09-13
Implement write() method for Box<MaybeUninit<T>>

Kixunil:
This adds method similar to `MaybeUninit::write` main difference being
it returns owned `Box`. This can be used to elide copy from stack
safely, however it's not currently tested that the optimization actually
occurs.

Analogous methods are not provided for `Rc` and `Arc` as those need to
handle the possibility of sharing. Some version of them may be added in
the future.

This was discussed in #63291 which this change extends.

### DIFF
--- a/compiler/rustc_data_structures/src/functor.rs
+++ b/compiler/rustc_data_structures/src/functor.rs
@@ -31,11 +31,9 @@ impl<T> IdFunctor for Box<T> {
             let value = raw.read();
             // SAFETY: Converts `Box<T>` to `Box<MaybeUninit<T>>` which is the
             // inverse of `Box::assume_init()` and should be safe.
-            let mut raw: Box<mem::MaybeUninit<T>> = Box::from_raw(raw.cast());
+            let raw: Box<mem::MaybeUninit<T>> = Box::from_raw(raw.cast());
             // SAFETY: Write the mapped value back into the `Box`.
-            raw.write(f(value)?);
-            // SAFETY: We just initialized `raw`.
-            raw.assume_init()
+            Box::write(raw, f(value)?)
         })
     }
 }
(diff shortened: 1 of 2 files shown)


the8472:
There's a pending RFC that proposes another approach which would probably make this one redundant.

https://github.com/rust-lang/rfcs/pull/2884

Kixunil:
@the8472 thanks for heads up! It could be argued it also makes #6321 redundant (although that one can be useful in FFI but then why not add this one for completeness?)

dtolnay (commented):
> I believe this change is simple and obvious.

To point out why this change isn't simple and obvious: this is a breaking change. It can make previously working code fail to compile, or keep compiling but do the wrong thing, or silently introduce UB into existing correct code.

Here is an example of an ordinary idiomatic use of MaybeUninit which would no longer work:

```rust
struct Thing {
    buf: Box<MaybeUninit<[u8; N]>>,
}

impl Thing {
    fn method(&mut self) {
        let _arr = self.buf.write([0; N]);
    }
}
```

Kixunil:
Ah, I see, good point. I think changing it to `fn write(this: Self)` would fix it, right? Will try to do it when I have time (shouldn't be more than couple of days).

Kixunil:
Fixed and rebased. Also noticed that wording in the example looked like a promise of optimization which is not actually guaranteed, so I reworded it accordingly.

dtolnay (approved):
Thanks!

Kixunil:
I'm confused by this - it conflicted with an unmerged PR? Is this some fluke or am I actually supposed to do something (what?) about it?

pietroalbini:
This has been merged into master by bors (in https://github.com/rust-lang/rust/pull/91486), but GitHub didn't detect it for whatever reason. Closing.

PR #88907, opened 2021-09-13
Highlight the `const fn` if error happened because of a bound on the impl block

WaffleLapkin:
Currently, for the following code, the compiler produces the errors like the
following:
```rust
struct Type<T>(T);

impl<T: Clone> Type<T> {
    const fn f() {}
}
```
```text
error[E0658]: trait bounds other than `Sized` on const fn parameters are unstable
 --> ./test.rs:3:6
  |
3 | impl<T: Clone> Type<T> {
  |      ^
  |
  = note: see issue #57563 <https://github.com/rust-lang/rust/issues/57563> for more information
  = help: add `#![feature(const_fn_trait_bound)]` to the crate attributes to enable
```

This can be confusing (especially to newcomers) since the error mentions "const fn parameters", but highlights only the impl.

This PR adds function highlighting, changing the error to the following:

```text
error[E0658]: trait bounds other than `Sized` on const fn parameters are unstable
 --> ./test.rs:3:6
  |
3 | impl<T: Clone> Type<T> {
  |      ^
4 |     pub const fn f() {}
  |     ---------------- function declared as const here
  |
  = note: see issue #57563 <https://github.com/rust-lang/rust/issues/57563> for more information
  = help: add `#![feature(const_fn_trait_bound)]` to the crate attributes to enable
```

---

I've originally wanted to point directly to `const` token, but couldn't find a way to get it's span. It seems like this span is lost during the AST -> HIR lowering.

Also, since the errors for object casts in `const fn`s (`&T` -> `&dyn Trait`) seem to trigger the same error, this PR accidentally changes these errors too. Not sure if it's desired or how to fix this.

P.S. it's my first time contributing to diagnostics, so feedback is very appreciated! 

---

r? @estebank 

@rustbot label: +A-diagnostics

### DIFF
--- a/compiler/rustc_const_eval/src/transform/check_consts/ops.rs
+++ b/compiler/rustc_const_eval/src/transform/check_consts/ops.rs
@@ -599,12 +599,21 @@ pub mod ty {
         }
 
         fn build_error(&self, ccx: &ConstCx<'_, 'tcx>, span: Span) -> DiagnosticBuilder<'tcx> {
-            feature_err(
+            let mut builder = feature_err(
                 &ccx.tcx.sess.parse_sess,
                 sym::const_fn_trait_bound,
                 span,
                 "trait bounds other than `Sized` on const fn parameters are unstable",
-            )
+            );
+
+            match ccx.fn_sig() {
+                Some(fn_sig) if !fn_sig.span.contains(span) => {
+                    builder.span_label(fn_sig.span, "function declared as const here");
+                }
+                _ => {}
+            }
+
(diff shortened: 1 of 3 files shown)


estebank (approved):
> Also, since the errors for object casts in const fns (&T -> &dyn Trait) seem to trigger the same error, this PR accidentally changes these errors too. Not sure if it's desired or how to fix this.

Those should probably get their own errors instead of these, as they are today they are a bit confusing, but I think your changes are worth it.

PR #88909, opened 2021-09-13
enable `PathBuf` usage in proc_macro bridge, `tracked::path`

drahnr:
Extracted from #87173 

Adds limited `PathBuf` support for convenience to the `proc_macro`-bridge code. Currently does _not_ handle invalid utf-t cases gracefully.

@rustbot assign @petrochenkov

### DIFF
--- a/compiler/rustc_expand/src/proc_macro_server.rs
+++ b/compiler/rustc_expand/src/proc_macro_server.rs
@@ -21,6 +21,7 @@ use rustc_span::{BytePos, FileName, MultiSpan, Pos, RealFileName, SourceFile, Sp
 use pm::bridge::{server, TokenTree};
 use pm::{Delimiter, Level, LineColumn, Spacing};
 use std::ops::Bound;
+use std::path;
 use std::{ascii, panic};
 
 trait FromInternal<T> {
@@ -406,7 +407,7 @@ impl server::FreeFunctions for Rustc<'_> {
     }
 
     fn track_path(&mut self, path: &str) {
-        self.sess.file_depinfo.borrow_mut().insert(Symbol::intern(path));
+        self.sess.file_depinfo.borrow_mut().insert(path::PathBuf::from(path));
     }
 }
 
--- a/src/test/run-make/track-path-dep-info/macro_def.rs
+++ b/src/test/run-make/track-path-dep-info/macro_def.rs
@@ -1,11 +1,21 @@
-#![feature(track_path)]
+#![feature(proc_macro_tracked_env,proc_macro_tracked_path)]
 #![crate_type = "proc-macro"]
 
 extern crate proc_macro;
 use proc_macro::*;
 
+use std::str;
+
 #[proc_macro]
 pub fn access_tracked_paths(_: TokenStream) -> TokenStream {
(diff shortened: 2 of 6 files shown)


compiler/rustc_expand/src/proc_macro_server.rs
```diff
@@ -406,7 +407,7 @@ impl server::FreeFunctions for Rustc<'_> {
     }
 
     fn track_path(&mut self, path: &str) {
-        self.sess.file_depinfo.borrow_mut().insert(Symbol::intern(path));
+        self.sess.file_depinfo.borrow_mut().insert(path::PathBuf::from(path));
```
petrochenkov, line 410:
```suggestion
        self.sess.file_depinfo.borrow_mut().insert(PathBuf::from(path));
```
Nit (here and in the other files as well): types are usually imported directly, without parent module.

src/test/run-make/track-path-dep-info/macro_def.rs
```diff
@@ -1,11 +1,21 @@
...
+use std::str;
+
 #[proc_macro]
 pub fn access_tracked_paths(_: TokenStream) -> TokenStream {
     tracked_path::path("emojis.txt");
+
+    // currently only valid utf-8 paths are supported
+    let invalid = [1_u8, 2,123, 254, 0, 0, 1, 1];
+    let invalid: &str = unsafe {
+        str::from_utf8_unchecked(&invalid[..])
```
petrochenkov, line 16: This is undefined behavior though.
I don't think UB is something that we should be testing.

petrochenkov:
What is the purpose of changing this internally, if the user-facing interface in `proc_macro/src/lib.rs` still only accepts strings?

drahnr:
Technically the limitation does not need to be there, since the internal representation of a `Pathbuf` is a `Vec<u8>`, this could hence be resolved getting access to that internal representation.

src/test/run-make/track-path-dep-info/macro_def.rs
bjorn3: In https://github.com/rust-lang/rust/pull/87173 it would be possible to safely construct a `PathBuf` that is not a valid UTF-8 path and then pass it to libproc_macro.

petrochenkov:
@drahnr 
I'm not sure what the previous commit was answering.
In https://github.com/rust-lang/rust/pull/88909#issuecomment-919072802 I meant that this PR doesn't make sense without including the `AsRef<Path>` part from https://github.com/rust-lang/rust/pull/87173.

(Also, https://github.com/rust-lang/rust/pull/88909#discussion_r708180513 wasn't addressed.)

JohnCSimon:
Ping from triage: 
Can you please address the build failures and the comments?
@drahnr

JohnCSimon:
Thanks, please adjust the label with `@rustbot ready` when you're ready for another review

camelid:
triage: @drahnr any updates?

JohnCSimon:
Ping from triage: 
@drahnr 
I'm closing this due to inactivity, Please to reopen when you are ready to continue with this. Thank you.

@rustbot label: +S-inactive

PR #88916, opened 2021-09-13
Mark `String::clone` as inline-able

### LINKED ISSUE #88905: String cloning is not optimized the same way as String construction
I don't know if it is really worth looking into, but I think I found some weird behavior in the rust compiler:
Let's consider the two following functions ([godbolt link](https://godbolt.org/z/T5M1a1e8a)):
```rust
pub fn nop(a: &String) {
    a.as_str().to_owned();
}

pub fn complex(a: &String) {
    a.to_owned();
}
```
They both allocate a String struct but do not use or return it. Rustc is able to optimize away the first one, but not the second one.
If we dig deeper, we can observe that the compiler is not able to optimize away the `to_owned` and `clone` method when called on a string. Let's look at the following functions ([godbolt link](https://godbolt.org/z/qM5vEx69x)):
```rust
pub fn a() {
    "ok".to_owned();
}

pub fn b() {
    "ok".to_owned().clone();
}

pub fn c() {
    "ok".to_owned().as_str().to_owned();
}

pub fn d() {
    "ok".to_owned().to_owned();
}

pub fn e() {
    String::from("ok");
}
```
The fonctions `a`, `c` and `e` are optimized away even on opt-level 2. Whereas the functions `b` and `d` are not and still produce calls to `alloc`, `dealloc`, `drop_in_place`, etc...

Looking at the stdlib code, [`<&str>::to_owned`](https://github.com/rust-lang/rust/blob/master/library/alloc/src/str.rs#L219) internally calls [`<&[u8]>::to_owned`](https://github.com/rust-lang/rust/blob/master/library/alloc/src/slice.rs#L833), which calls [`<&[u8]>::to_vec`](https://github.com/rust-lang/rust/blob/master/library/alloc/src/slice.rs#L466), which calls [`<&[u8]>::to_vec_in`](https://github.com/rust-lang/rust/blob/master/library/alloc/src/slice.rs#L489) which ends up calling [`hack::to_vec`](https://github.com/rust-lang/rust/blob/master/library/alloc/src/slice.rs#L185).

[`String::to_string`](https://github.com/rust-lang/rust/blob/master/library/alloc/src/string.rs#L2462) calls [`String::to_owned`](https://github.com/rust-lang/rust/blob/master/library/alloc/src/borrow.rs#L89) which calls [`String::clone`](https://github.com/rust-lang/rust/blob/master/library/alloc/src/string.rs#L1837) which calls [`Vec<u8>::clone`](https://github.com/rust-lang/rust/blob/master/library/alloc/src/vec/mod.rs#L2414) which calls  [`<&[u8]>::to_vec_in`](https://github.com/rust-lang/rust/blob/master/library/alloc/src/slice.rs#L489) with the only difference from the `<&str>::to_owned` path being that the possibly custom allocator of the `Vec` internal buffer is cloned and passed to `<&[u8]>::to_vec_in` in `Vec<u8>::clone`.

Is it possible that rustc is not able to correctly inline the allocator functions when the default allocator is not directly assumed ?

This was tested on `Rustc 1.55` and `Rustc 1.57.0-nightly (b69fe5726 2021-09-10)

notriddle:
It calls `Vec::clone` and performs a newtype-wrap around it, so it should be trivial enough that it doesn't cause major size regressions. It also fixes #88905

### DIFF
--- a/library/alloc/src/string.rs
+++ b/library/alloc/src/string.rs
@@ -1834,10 +1834,14 @@ impl fmt::Display for FromUtf16Error {
 #[cfg(not(no_global_oom_handling))]
 #[stable(feature = "rust1", since = "1.0.0")]
 impl Clone for String {
+    // Not marking it inline causes worse assembly to be generated.
+    // https://github.com/rust-lang/rust/issues/88905
+    #[inline]
     fn clone(&self) -> Self {
         String { vec: self.vec.clone() }
     }
 
+    #[inline]
     fn clone_from(&mut self, source: &Self) {
         self.vec.clone_from(&source.vec);
     }
--- a/src/test/codegen/issue-88905.rs
+++ b/src/test/codegen/issue-88905.rs
@@ -1,0 +1,13 @@
+//
+// min-llvm-version: 10.0.1
+// compile-flags: -C opt-level=3
+#![crate_type="lib"]
+
+#[no_mangle]
+pub fn alloc_test(a: &String) {
+    // CHECK-LABEL: @alloc_test
+    // CHECK-NEXT: start:
+    // CHECK-NEXT: ret void
+    let x = String::clone(a);
+    drop(x);
+}


camelid:
Hmm, seems like the results are quite mixed. It could be worth doing another perf run to see if the results are spurious, but I'll let T-libs decide.

the8472:
Note that LTO does inline this, it'll only benefit non-LTO builds.

dtolnay (commented):
I am not convinced that it is worth making this change. Is there a common idiom that leads to strings getting cloned and never used? If this is minimized from real world code then we can evaluate vs the costs of inline, but I am having trouble seeing how this would come up in a hot path in practice.

PR #88926, opened 2021-09-14
[BENCH] Revert "Fix non-capturing closure return type coercion"

workingjubilee:
Using to attempt to answer profiling questions raised in discussion in https://github.com/rust-lang/rust/pull/88857

r? @ghost


jackh726:
Closing since we don't want to land; was just for perf.

Thanks @workingjubilee

PR #88936, opened 2021-09-14
Suggest using `Arc` on `!Send`/`!Sync` types

estebank:


src/test/ui/associated-type-bounds/bad-bounds-on-assoc-in-trait.stderr
```diff
@@ -47,17 +41,7 @@ LL |     type C: Clone + Iterator<Item: Send + Iterator<Item: for<'a> Lam<&'a u8
...
-   |
-LL | / pub unsafe auto trait Sync {
-LL | |     // FIXME(estebank): once support to add notes in `rustc_on_unimplemented`
-LL | |     // lands in beta, and it has been extended to check whether a closure is
-LL | |     // anywhere in the requirement chain, extend it as such (#48534):
-...  |
-LL | |     // Empty
-LL | | }
-   | |_^ required by this bound in `Sync`
+   = note: consider using `std::sync::Arc<<<Self as Case1>::C as Iterator>::Item>`; for more information visit <https://doc.rust-lang.org/book/ch16-03-shared-state.html>
```
m-ou-se, line 44: This now seems to suggest to use `Arc<..::Item>` instead of `..::Item`. It might be useful to say `consider using Arc instead of Rc: ...`.

estebank: But That's exactly what this is suggesting: we _didn't_ have an `Rc` here. Ideally we should be able to tell what context this error is being emitted on, but sadly `rustc_on_unimplemented` doesn't have that filtering today.

yaahc:
r? rust-lang/compiler

compiler-errors:
A bors sync made this re-approved, sorry hehe

@bors r-

Dylan-DPC:
Closing this as it is inactive

estebank:
Squashed, rebased on top of current main and re-blessed the tests. No other changes since the prior approval.

@bors r=nagisa

compiler/rustc_trait_selection/src/traits/error_reporting/suggestions.rs
```diff
@@ -2743,6 +2743,12 @@ impl<'tcx> TypeErrCtxtExt<'tcx> for TypeErrCtxt<'_, 'tcx> {
             }
             ObligationCauseCode::BindingObligation(item_def_id, span)
             | ObligationCauseCode::ExprBindingObligation(item_def_id, span, ..) => {
+                if self.tcx.is_diagnostic_item(sym::Send, item_def_id)
```
compiler-errors, line 2746: why isn't this commented? seems like something that could use a justification

estebank: I'll add a comment.

PR #88946, opened 2021-09-14
[beta] backports

cuviper:
- rustdoc: Fix ICE with `doc(hidden)` on tuple variant fields #88639 
- Fix 2021 `dyn` suggestion that used code as label #88657
- Workaround blink/chromium grid layout limitation of 1000 rows #88776
- Change scope of temporaries in match guards #88678 
- Add a regression test for #88649 #88691 
- Revert anon union parsing #88775 
- Disable validate_maintainers. #88977 

Also drop stage0 rustfmt, because that's only supposed to be used on master.

r? @Mark-Simulacrum


Mark-Simulacrum:
I think this'll need a `backport-of: none` on the first commit, but CI will complain if necessary about that, I think.

cuviper:
Ah, sorry, I didn't use the local `cherry-pick.sh` at all, just `git cherry-pick -x ...` which has it's own format.

However:
```
Run src/ci/scripts/verify-backported-commits.sh
  src/ci/scripts/verify-backported-commits.sh
  shell: /bin/bash --noprofile --norc -e -o pipefail {0}
  env:
    CI_JOB_NAME: x86_64-gnu-tools
    SCCACHE_BUCKET: rust-lang-ci-sccache2
    TOOLSTATE_REPO: https://github.com/rust-lang-nursery/rust-toolstate
    CACHE_DOMAIN: ci-caches.rust-lang.org
    CI_ONLY_WHEN_SUBMODULES_CHANGED: 1
    IMAGE: x86_64-gnu-tools
fatal: unknown commit 53fd98ca776cb875bc9e5514f56b52eb74f9e7a9
All commits in `HEAD` are present in `master`
```

:thinking:

cuviper:
I added `backport-of: nothing` on the new commit, and left the others alone since they were clean cherry-picks and shouldn't need a note. I also filed #88947 for the apparently-broken CI script.

@bors r=Mark-Simulacrum

cuviper:
I guess that failure is #88924.

ehuss:
@bors retry

#88924

ehuss:
@bors r-

I'm pretty sure this will fail without #88977.

cuviper:
I added additional backports, including that CI fix.

PR #80632, opened 2021-01-02
Identify unreachable subpatterns more reliably

Nadrieril:
In https://github.com/rust-lang/rust/pull/80104 I used `Span`s to identify unreachable sub-patterns in the presence of or-patterns during exhaustiveness checking. In https://github.com/rust-lang/rust/issues/80501 it was revealed that `Span`s are complicated and that this was not a good idea.
Instead, this PR identifies subpatterns logically: as a path in the tree of subpatterns of a given pattern. I made a struct that captures a set of such subpatterns. This is a bit complex, but thankfully self-contained; the rest of the code does not need to know anything about it.
Fixes https://github.com/rust-lang/rust/issues/80501. I think I managed to keep the perf neutral.

r? @varkor

PR #88950, opened 2021-09-15
Add an intermediate representation to exhaustiveness checking

Nadrieril:
The exhaustiveness checking algorithm keeps deconstructing patterns into a `Constructor` and some `Fields`, but does so a bit all over the place. This PR introduces a new representation for patterns that already has that information, so we only compute it once at the start.
I find this makes code easier to follow. In particular `DeconstructedPat::specialize` is a lot simpler than what happened before, and more closely matches the description of the algorithm. I'm also hoping this could help for the project of librarifying exhaustiveness for rust_analyzer since it decouples the algorithm from `rustc_middle::Pat`.

### DIFF
--- a/compiler/rustc_mir_build/src/thir/pattern/check_match.rs
+++ b/compiler/rustc_mir_build/src/thir/pattern/check_match.rs
@@ -190,7 +190,7 @@ impl<'tcx> MatchVisitor<'_, 'tcx> {
         let scrut_ty = self.typeck_results.expr_ty_adjusted(scrut);
         let report = compute_match_usefulness(&cx, &arms, scrut.hir_id, scrut_ty);
 
-        report_arm_reachability(&cx, &report, |_, arm_span, arm_hir_id, catchall| {
+        report_arm_reachability(&cx, &report, |arm_span, arm_hir_id, catchall| {
             match source {
                 hir::MatchSource::ForLoopDesugar | hir::MatchSource::Normal => {
                     unreachable_pattern(cx.tcx, arm_span, arm_hir_id, catchall);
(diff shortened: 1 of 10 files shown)


camelid:
Interesting, `issue-58319` doesn't seem to have a pattern match. I guess it regressed because it has a large `struct` definition that is being lowered?

Nadrieril:
It's weird though, because patterns used to be lowered already via a `PatternFolder`, which I imagine traverses everythings and reallocates. So I didn't expect that my new lowering would be much slower

Nadrieril:
Ok, some progress. My local measurements differ quite a bit from perf here, so I want to see what it says now.

@bors try @rust-timer queue

Nadrieril:
Huh ok I found the main issue; commit [`180204f` (#88950)](https://github.com/rust-lang/rust/pull/88950/commits/180204f0e8f27ceb2c34ca62c82cf131c43416c7) fixes it. Apparently changing from an explicit for loop to the equivalent iterator expression killed perf specifically for match-stress-enum. The rest of the regression was because cloning my new struct is slower than copying a ref, so I arena-allocated my struct too.
I don't know how to stop the timer, let's hope rescheduling does it.

@bors try @rust-timer queue

Nadrieril:
Oops, I confused rustc-timer, that's the benchmark for the previous branch head. Let's try again, I'm measuring only improvements locally.

@bors try @rust-timer queue

Nadrieril:
Cleaned up, let's test again

@bors try @rust-timer queue

Nadrieril:
Last perf run before this is reviewable :crossed_fingers: 

@bors try @rust-timer queue

Nadrieril:
The small perf regression on one benchmark is made up for by the improvements on quite a few benchmarks.

Cool, this is ready for review! I would usually ask @varkor but he seems unavailable at the moment.
r? @oli-obk if you have the time

Nadrieril:
Ohh I can trivialize the terrible mess that was `SubPatSet` now! Biig simplification, this was the most arcane part of the algorithm. Help I can't stop simplifying things :upside_down_face: . I'll have to stop when a reviewer comes by

@bors try @rust-timer queue

Nadrieril:
@varkor remember the mess I introduced to solve https://github.com/rust-lang/rust/pull/80632? It's all in this cute function now https://github.com/Nadrieril/rust/blob/b7e358ee17a5794603b2324858de078c4586acfc/compiler/rustc_mir_build/src/thir/pattern/deconstruct_pat.rs#L1577-L1586 ! 7 lines of this do the job of about 250 lines of convoluted `SubPatSet` logic x)

oli-obk:
😆 I'm getting to it right now

oli-obk:
r=me unless you want to keep investigating the stress test perf regression

Nadrieril:
Yay thanks!

Most (1%) of the regression comes from the last commit, presumably the part where we retraverse all the patterns. Since that didn't affect the more realistic benchmarks, and no real match has 8000 arms, I believe that retraversing the patterns takes negligible time in real cases. So I'll ignore that 1%.
Imo it's plenty worth it in code clarity anyways.

@bors r=oli-obk

DevinR528:
Just want to second how much more readable the code is without `SubPatSet` (after https://github.com/rust-lang/rust/pull/88950/commits/b7e358ee17a5794603b2324858de078c4586acfc) everything looks better look forward to having to rebase :stuck_out_tongue_winking_eye:!

BubbaSheen:
Hypothetical information in check. Move pawns forward to await checkmate
from King W VV.

PR #88952, opened 2021-09-15
Add new tier-3 target: armv7-unknown-linux-uclibceabihf

skrap:
This change adds a new tier-3 target: armv7-unknown-linux-uclibceabihf

This target is primarily used in embedded linux devices where system resources are slim and glibc is deemed too heavyweight.  Cross compilation C toolchains are available [here](https://toolchains.bootlin.com/) or via [buildroot](https://buildroot.org).

The change is based largely on a previous PR #79380 with a few minor modifications.  The author of that PR was unable to push the PR forward, and graciously allowed me to take it over.

Per the [target tier 3 policy](https://github.com/rust-lang/rfcs/blob/master/text/2803-target-tier-policy.md), I volunteer to be the "target maintainer".

This is my first PR to Rust itself, so I apologize if I've missed things!

### DIFF
--- a/compiler/rustc_target/src/spec/armv7_unknown_linux_uclibceabihf.rs
+++ b/compiler/rustc_target/src/spec/armv7_unknown_linux_uclibceabihf.rs
@@ -17,6 +17,7 @@ pub fn target() -> Target {
             cpu: "generic".to_string(),
             max_atomic_width: Some(64),
             mcount: "_mcount".to_string(),
+            abi: "eabihf".to_string(),
             ..base
         },
     }
--- a/compiler/rustc_target/src/spec/armv7_unknown_linux_uclibceabihf.rs
+++ b/compiler/rustc_target/src/spec/armv7_unknown_linux_uclibceabihf.rs
@@ -1,0 +1,23 @@
+use crate::spec::{Target, TargetOptions};
+
+// This target is for uclibc Linux on ARMv7 without NEON or
+// thumb-mode. See the thumbv7neon variant for enabling both.
+
+pub fn target() -> Target {
+    let base = super::linux_uclibc_base::opts();
+    Target {
+        llvm_target: "armv7-unknown-linux-gnueabihf".to_string(),
+        pointer_width: 32,
+        data_layout: "e-m:e-p:32:32-Fi8-i64:64-v128:64:128-a:0:32-n32-S64".to_string(),
+        arch: "arm".to_string(),
+
+        options: TargetOptions {
+            // Info about features at https://wiki.debian.org/ArmHardFloatPort
+            features: "+v7,+vfp3,-d32,+thumb2,-neon".to_string(),
+            cpu: "generic".to_string(),
+            max_atomic_width: Some(64),
+            mcount: "_mcount".to_string(),
+            ..base
+        },
+    }
+}
--- a/library/std/src/sys/unix/process/process_unix.rs
+++ b/library/std/src/sys/unix/process/process_unix.rs
@@ -419,7 +419,7 @@ impl Command {
         }
 
         // Only glibc 2.24+ posix_spawn() supports returning ENOENT directly.
-        #[cfg(all(target_os = "linux", any(target_env = "gnu", target_env = "uclibc")))]
+        #[cfg(all(target_os = "linux", target_env = "gnu"))]
         {
             if let Some(version) = sys::os::glibc_version() {
                 if version < (2, 24) {
--- a/library/std/src/sys/unix/process/process_unix.rs
+++ b/library/std/src/sys/unix/process/process_unix.rs
@@ -419,7 +419,7 @@ impl Command {
         }
 
         // Only glibc 2.24+ posix_spawn() supports returning ENOENT directly.
-        #[cfg(all(target_os = "linux", target_env = "gnu"))]
+        #[cfg(all(target_os = "linux", any(target_env = "gnu", target_env = "uclibc")))]
         {
             if let Some(version) = sys::os::glibc_version() {
                 if version < (2, 24) {
--- a/library/std/src/sys/unix/thread.rs
+++ b/library/std/src/sys/unix/thread.rs
@@ -594,7 +594,8 @@ pub mod guard {
                 Some(stackaddr - guardsize..stackaddr)
             } else if cfg!(all(target_os = "linux", target_env = "musl")) {
                 Some(stackaddr - guardsize..stackaddr)
-            } else if cfg!(all(target_os = "linux", target_env = "gnu")) {
+            } else if cfg!(all(target_os = "linux", any(target_env = "gnu", target_env = "uclibc")))
+            {
                 // glibc used to include the guard area within the stack, as noted in the BUGS
                 // section of `man pthread_attr_getguardsize`.  This has been corrected starting
                 // with glibc 2.27, and in some distro backports, so the guard is now placed at the
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1879,9 +1879,9 @@ checksum = "830d08ce1d1d941e6b30645f1a0eb5643013d835ce3779a5fc208261dbe10f55"
 
 [[package]]
 name = "libc"
-version = "0.2.99"
+version = "0.2.103"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "a7f823d141fe0a24df1e23b4af4e3c7ba9e5966ec514ea068c93024aa7deb765"
+checksum = "dd8f7255a17a627354f321ef0055d63b898c6fb27eff628af4d1b66b7331edf6"
 dependencies = [
  "rustc-std-workspace-core",
 ]
--- a/compiler/rustc_codegen_llvm/src/callee.rs
+++ b/compiler/rustc_codegen_llvm/src/callee.rs
@@ -176,6 +176,7 @@ pub fn get_fn(cx: &CodegenCx<'ll, 'tcx>, instance: Instance<'tcx>) -> &'ll Value
             if cx.use_dll_storage_attrs
                 && tcx.is_dllimport_foreign_item(instance_def_id)
                 && tcx.sess.target.env != "gnu"
+                && tcx.sess.target.env != "uclibc"
             {
                 llvm::LLVMSetDLLStorageClass(llfn, llvm::DLLStorageClass::DllImport);
             }
--- a/compiler/rustc_codegen_llvm/src/callee.rs
+++ b/compiler/rustc_codegen_llvm/src/callee.rs
@@ -175,8 +175,7 @@ pub fn get_fn(cx: &CodegenCx<'ll, 'tcx>, instance: Instance<'tcx>) -> &'ll Value
             // should use dllimport for functions.
             if cx.use_dll_storage_attrs
                 && tcx.is_dllimport_foreign_item(instance_def_id)
-                && tcx.sess.target.env != "gnu"
-                && tcx.sess.target.env != "uclibc"
+                && !matches!(tcx.sess.target.env.as_ref(), "gnu" | "uclibc")
             {
                 llvm::LLVMSetDLLStorageClass(llfn, llvm::DLLStorageClass::DllImport);
             }
--- a/compiler/rustc_middle/src/ty/layout.rs
+++ b/compiler/rustc_middle/src/ty/layout.rs
@@ -3009,7 +3009,7 @@ impl<'tcx> LayoutCx<'tcx, TyCtxt<'tcx>> {
         };
 
         let target = &self.tcx.sess.target;
-        let target_env_gnu_like = matches!(&target.env[..], "gnu" | "musl");
+        let target_env_gnu_like = matches!(&target.env[..], "gnu" | "musl" | "uclibc");
         let win_x64_gnu = target.os == "windows" && target.arch == "x86_64" && target.env == "gnu";
         let linux_s390x_gnu_like =
             target.os == "linux" && target.arch == "s390x" && target_env_gnu_like;
@@ -3107,7 +3107,7 @@ impl<'tcx> LayoutCx<'tcx, TyCtxt<'tcx>> {
             if arg.layout.is_zst() {
                 // For some forsaken reason, x86_64-pc-windows-gnu
                 // doesn't ignore zero-sized struct arguments.
-                // The same is true for {s390x,sparc64,powerpc}-unknown-linux-{gnu,musl}.
+                // The same is true for {s390x,sparc64,powerpc}-unknown-linux-{gnu,musl,uclibc}.
                 if is_return
                     || rust_abi
                     || (!win_x64_gnu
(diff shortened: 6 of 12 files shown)


compiler/rustc_target/src/spec/armv7_unknown_linux_uclibceabihf.rs
```diff
@@ -0,0 +1,23 @@
+use crate::spec::{Target, TargetOptions};
+
+// This target is for glibc Linux on ARMv7 without NEON or
```
oli-obk, line 3:
```suggestion
// This target is for uclibc Linux on ARMv7 without NEON or
```

library/std/src/sys/unix/args.rs
```diff
@@ -98,13 +98,16 @@ mod imp {
         // On Linux-GNU, we rely on `ARGV_INIT_ARRAY` below to initialize
         // `ARGC` and `ARGV`. But in Miri that does not actually happen so we
         // still initialize here.
-        #[cfg(any(miri, not(all(target_os = "linux", target_env = "gnu"))))]
+        #[cfg(any(
+            miri,
+            not(all(target_os = "linux", any(target_env = "gnu")))
```
oli-obk, line 103: This `any` doesn't do anything, right?

library/std/src/sys/unix/args.rs
```diff
@@ -98,13 +98,16 @@ mod imp {
...
+        ))]
         really_init(_argc, _argv);
     }
 
     /// glibc passes argc, argv, and envp to functions in .init_array, as a non-standard extension.
     /// This allows `std::env::args` to work even in a `cdylib`, as it does on macOS and Windows.
-    #[cfg(all(target_os = "linux", target_env = "gnu"))]
+    #[cfg(all(target_os = "linux", any(target_env = "gnu")))]
```
oli-obk, line 110: same here.

compiler/rustc_target/src/spec/armv7_unknown_linux_uclibceabihf.rs
```diff
@@ -0,0 +1,23 @@
+use crate::spec::{Target, TargetOptions};
```
oli-obk, line 1: Please leave a comment in this file with your preferred way of being contacted when someone touches this target.

oli-obk (changes requested):
Some nits and rustfmt is unhappy, other than than 👍 from me

compiler/rustc_target/src/spec/armv7_unknown_linux_uclibceabihf.rs
Mark-Simulacrum: I think this documentation should go into https://github.com/rust-lang/rust/tree/master/src/doc/rustc/src/platform-support, not target definition files -- I think we should also work on a template with some other things that new targets should fill out (similar to the existing file in there).

Notably, I think this target doesn't yet have the following requirement fulfilled, though it may be something that we document for *all* uclibc targets in a common place and then link to (still needs to happen though).

skrap:
HI All - Thanks for the feedback.  I actually wasn't intending this to be reviewed just yet, as I'd left it in "draft" status, but I will certainly address the feedback.  Sorry for engaging you all too soon!  TIL: rust-highfive will add S-waiting-on-review to draft PRs!

Would it be appropriate to file a issue against rust-highfive?

compiler/rustc_target/src/spec/armv7_unknown_linux_uclibceabihf.rs
skrap: Thanks for the feedback!  I'm glad to do so.  Is there a high-quality example of such instructions for another platfrm which you could point me towards?  That would give me something to imitate.

Mark-Simulacrum:
We don't generally use draft status to indicate review state, so I don't think a PR/issue is warranted. Different authors have different preferences for early review :)

You can adjust the waiting-on-{review,author} labels with rustbot -- for example:

@rustbot author

(See [shortcuts](https://github.com/rust-lang/triagebot/wiki/Shortcuts) and [more freeform labeling docs](https://github.com/rust-lang/triagebot/wiki/Labeling)).

Mark-Simulacrum:
Please quote the corresponding requirements verbatim from the tier 3 [target tier policy](https://doc.rust-lang.org/nightly/rustc/target-tier-policy.html#tier-3-target-policy) to explain how the target meets those requirements (just a comment here is fine).

Mark-Simulacrum:
Most of the modified code is outside my domain, so I will r? rust-lang/compiler-team

skrap:
@Mark-Simulacrum Here's what I've come up with for conformance to the tier 3 requirements.  Please LMK if you see any issues!

I volunteer to be the target maintainer.  Happy to have help as well (anyone reading this, feel free to reach out!)

>        Target names should not introduce undue confusion or ambiguity unless absolutely necessary to maintain ecosystem compatibility. For example, if the name of the target makes people extremely likely to form incorrect beliefs about what it targets, the name should be changed or augmented to disambiguate it.

The name is patterned after the current tier-2 target `armv7-unknown-linux-gnueabihf`.  It targets the same architecture, and the only difference is that it uses the `uclibc` standard library rather than the `gnu` standard library.  The "eabi" text designates the ABI (the [ARM EABI](https://github.com/ARM-software/abi-aa/blob/main/bsabi32/bsabi32.rst) ) and the "hf" designates the "hard float" floating point passing style, where floating point values are passed in registers rather than on the stack.

There are additional tier-3 `uclibc` targets which adopt similar naming conventions: `armv5te-unknown-linux-uclibceabi`, `mipsel-unknown-linux-uclibc`, `mips-unknown-linux-uclibc`.

>    Tier 3 targets may have unusual requirements to build or use, but must not create legal issues or impose onerous legal terms for the Rust project or for Rust developers or users.
>        The target must not introduce license incompatibilities.

This target does not seem to include any further licensing burden on Rust.  The primary components of the toolchain are:
<html><head></head><body>

component | version
-- | --
binutils | 2.34
gcc | 10.2.0
linux-headers | 5.4.61
uclibc | 1.0.34

</body></html>

uclibc (https://uclibc-ng.org/) is [licensed as LGPL](https://gogs.waldemar-brodkorb.de/oss/uclibc-ng).  All of the other toolchain components are presumably already used in any Rust on Linux build.

>  Anything added to the Rust repository must be under the standard Rust license (MIT OR Apache-2.0).

The code in this PR is offered under the standard Rust license.

>  The target must not cause the Rust tools or libraries built for any other host (even when supporting cross-compilation to the target) to depend on any new dependency less permissive than the Rust licensing policy. This applies whether the dependency is a Rust crate that would require adding new license exceptions (as specified by the tidy tool in the rust-lang/rust repository), or whether the dependency is a native library or binary. In other words, the introduction of the target must not cause a user installing or running a version of Rust or the Rust tools to be subject to any new license requirements.

This PR conforms to this requirement.  I don't expect any change in requirements for any other target.  

>  If the target supports building host tools (such as rustc or cargo), those host tools must not depend on proprietary (non-FOSS) libraries, other than ordinary runtime libraries supplied by the platform and commonly used by other binaries built for the target. For instance, rustc built for the target may depend on a common proprietary C runtime library or console output library, but must not depend on a proprietary code generation library or code optimization library. Rust's license permits such combinations, but the Rust project has no interest in maintaining such combinations within the scope of Rust itself, even at tier 3.

This PR doesn't reference or depend on any non-FOSS code.

>  Targets should not require proprietary (non-FOSS) components to link a functional binary or library.

It doesn't! 👍 

>  "onerous" here is an intentionally subjective term. At a minimum, "onerous" legal/licensing terms include but are not limited to: non-disclosure requirements, non-compete requirements, contributor license agreements (CLAs) or equivalent, "non-commercial"/"research-only"/etc terms, requirements conditional on the employer or employment of any particular Rust developers, revocable terms, any requirements that create liability for the Rust project or its developers or users, or any requirements that adversely affect the livelihood or prospects of the Rust project or its developers or users.

Acknowledged.  I don't expect the requirements here will be considered onerous, as they should not differ from any standard Rust build on Linux.

👍 I believe this requirement is satisfied.

>  This requirement does not prevent part or all of this policy from being cited in an explicit contract or work agreement (e.g. to implement or maintain support for a target). This requirement exists to ensure that a developer or team responsible for reviewing and approving a target does not face any legal threats or obligations that would prevent them from freely exercising their judgment in such approval, even if such judgment involves subjective matters or goes beyond the letter of these requirements.

👍 I believe this requirement is satisfied.

This target (once some PRs are accepted elsewhere, e.g. `libc`) will build the full Rust `x.py dist` distribution, including the full Rust standard library.  (I'm already doing this in my dev builds.)

Done, see diff.  (It would be possible to expand the `x.py test` emulation targets to include this target, but given that this is only tier 3 it seemed premature.) 

>        Backlinks such as those generated by the issue/PR tracker when linking to an issue or PR are not considered a violation of this policy, within reason. However, such messages (even on a separate repository) must not generate notifications to anyone involved with a PR who has not requested such notifications.

👍 Acknowledged.

>  Patches adding or updating tier 3 targets must not break any existing tier 2 or tier 1 target, and must not knowingly break another tier 3 target without approval of either the compiler team or the maintainers of the other tier 3 target.
>   In particular, this may come up when working on closely related targets, such as variations of the same architecture with different features. Avoid introducing unconditional uses of features that another variation of the target may not have; use conditional compilation or runtime detection, as appropriate, to let each target run code supported by that target.

👍 Acknowledged.

library/std/src/sys/unix/process/process_unix.rs
```diff
@@ -409,7 +409,7 @@ impl Command {
         }
 
         // Only glibc 2.24+ posix_spawn() supports returning ENOENT directly.
-        #[cfg(all(target_os = "linux", target_env = "gnu"))]
+        #[cfg(all(target_os = "linux", any(target_env = "gnu", target_env = "uclibc")))]
```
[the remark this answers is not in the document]
nagisa: Ah, and to add, from what I can tell uClibc does implement `posix_spawn` so ideally it does get used, somewhow^^

WooDzu:
Has anybody tried compiling library/std for armv7-unknown-linux-uclibceabihf architecture already?

I built rust from this branch and then try to cross-compile library/std:

> ./x.py build --stage 1 --host x86_64-unknown-linux-gnu --target armv7-unknown-linux-uclibceabihf library/std 

but I'm running into this error:

> error[E0425]: cannot find value `SYS_clone3` in this scope</summary>

scr1

skrap:
> Has anybody tried compiling library/std for armv7-unknown-linux-uclibceabihf architecture already?

Yes.  The issue is that you need a recent change to libc: https://github.com/rust-lang/libc/pull/2403

It's in the latest `libc` (only released yesterday) so I guess I should update this PR to include a bump in the `libc` version in Config.toml.

skrap:
@nagisa Thanks for the review!  I will get to your comments above.  @ykoehler may have some comments as well, as he is the author of the code in question.

WooDzu:
> > Has anybody tried compiling library/std for armv7-unknown-linux-uclibceabihf architecture already?
> 
> Yes. The issue is that you need a recent change to libc: [rust-lang/libc#2403](https://github.com/rust-lang/libc/pull/2403)
> 
> It's in the latest `libc` (only released yesterday) so I guess I should update this PR to include a bump in the `libc` version in Config.toml.

Thanks! I can confirm that [updating this](https://github.com/skrap/rust/blob/add-armv7-uclibc/library/std/Cargo.toml#L18) from `0.2.99` to `0.2.103` works and library/std builds fine.

compiler/rustc_target/src/spec/armv7_unknown_linux_uclibceabihf.rs
```diff
@@ -0,0 +1,23 @@
...
+
+// This target is for uclibc Linux on ARMv7 without NEON or
+// thumb-mode. See the thumbv7neon variant for enabling both.
+
+pub fn target() -> Target {
+    let base = super::linux_uclibc_base::opts();
+    Target {
+        llvm_target: "armv7-unknown-linux-gnueabihf".to_string(),
```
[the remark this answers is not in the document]
skrap: I actually don't know what the implications of the LLVM target triple are here.  The code in `armv5te_unknown_linux_uclibceabi.rs` uses the same `gnueabihf` suffix, for what that's worth.  I made this requested change, but compilation now fails with:
```          
/home/vagrant/armv7-eabihf--uclibc--bleeding-edge-2020.08-1/bin/../lib/gcc/arm-buildroot-linux-uclibcgnueabihf/10.2.0/../../../../arm-buildroot-linux-uclibcgnueabihf/bin/ld: failed to merge target specific data of file /home/vagrant/rust/build/x86_64-unknown-linux-gnu/stage1-std/armv7-unknown-linux-uclibceabihf/release/deps/liballoc-b07e17534c57ab21.rlib(alloc-b07e17534c57ab21.alloc.674e4019-cgu.14.rcgu.o)
```

I note that uclibceabihf isn't a known `EnvironmentType` listed here: https://llvm.org/doxygen/classllvm_1_1Triple.html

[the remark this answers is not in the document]
nagisa: I see, gnueabihf is a okay fallback to use in that case. Otherwise LLVM will use an `unknown` and that's why you'll see mismatches like these come up.

library/std/src/sys/unix/thread.rs
```diff
@@ -581,7 +581,8 @@ pub mod guard {
                 Some(stackaddr - guardsize..stackaddr)
             } else if cfg!(all(target_os = "linux", target_env = "musl")) {
                 Some(stackaddr - guardsize..stackaddr)
-            } else if cfg!(all(target_os = "linux", target_env = "gnu")) {
+            } else if cfg!(all(target_os = "linux", any(target_env = "gnu", target_env = "uclibc")))
```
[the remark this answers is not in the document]
skrap: uClibc uses nptl directly from glibc, and does not have the fix from newer glibc.

The [glibc fix](https://sourceware.org/git/?p=glibc.git;a=blobdiff;f=nptl/allocatestack.c;h=b374f4794d0614ebc3af89edd857b798b7d9c6fd;hp=05b8ed331bef09bc3ff6c2f9f3c2a1b69c4e2abb;hb=630f4cc3aa019ede55976ea561f1a7af2f068639;hpb=c70e4e9c9efff9df4c847dd7cfd81bae674219ab).
The [code from uclibc](https://cgit.uclibc-ng.org/cgi/cgit/uclibc-ng.git/tree/libpthread/nptl/allocatestack.c#n462).

library/std/src/sys/unix/process/process_unix.rs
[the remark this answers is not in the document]
skrap: Ok, I did some experiments, and `uclibc` does implement `posix_spawn` but it doesn't directly return `ENOENT` directly as required by this code.  I will remove this condition within the function — the condition guarding the whole function is sufficient.

PR #88953, opened 2021-09-15
Add chown functions to std::os::unix::fs to change the owner and group of files

joshtriplett:
This is a straightforward wrapper that uses the existing helpers for C
string handling and errno handling.

Having this available is convenient for UNIX utility programs written in
Rust, and avoids having to call unsafe functions like `libc::chown`
directly and handle errors manually, in a program that may otherwise be
entirely safe code.

In addition, these functions provide a more Rustic interface by
accepting appropriate traits and using `None` rather than `-1`.

### DIFF
--- a/library/std/src/os/unix/fs.rs
+++ b/library/std/src/os/unix/fs.rs
@@ -5,6 +5,7 @@
 use super::platform::fs::MetadataExt as _;
 use crate::fs::{self, OpenOptions, Permissions};
 use crate::io;
+use crate::os::unix::io::{AsFd, AsRawFd};
 use crate::path::Path;
 use crate::sys;
 use crate::sys_common::{AsInner, AsInnerMut, FromInner};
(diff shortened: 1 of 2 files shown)


dtolnay (approved):
This seems reasonable to me.

The `nix` crate implements this using newtypes to prevent permuting the uid/gid arguments: https://docs.rs/nix/0.22.1/nix/unistd/fn.chown.html. But u32 seems all right to me.

the8472:
>  and avoids having to call unsafe functions like libc::chown

That's a poor justification considering that [nix offers a safe wrapper](https://docs.rs/nix/0.22.1/nix/unistd/fn.chown.html) exists. I'm not opposed to inclusion in the standard library but the reasoning for inclusion should be defensible.

dtolnay (approved):

joshtriplett:
@the8472 I'm currently using that exact wrapper in my own code, along with other bits of `nix`.

I think https://github.com/rust-lang/rust/issues/88989 would be the right place to continue the conversation about this (and about more general policies). I've already posted a comment there about where I think the boundary should be between `std` functionality and `nix` functionality.

Issue #51999, opened 2018-07-02
Tracking issue for RFC 2345, "Allow panicking in constants" (const_panic)


Centril:
This is a tracking issue for the RFC *"Allow panicking in constants"* (rust-lang/rfcs#2345).

**Steps:**

- [x] Implement the RFC (cc @rust-lang/compiler -- can anyone write up mentoring instructions?)
- [ ] Adjust documentation ([see instructions on forge][doc-guide])
- [ ] Stabilization PR ([see instructions on forge][stabilization-guide])

[stabilization-guide]: https://forge.rust-lang.org/stabilization-guide.html
[doc-guide]: https://forge.rust-lang.org/stabilization-guide.html#updating-documentation

**Unresolved questions:**

- [ ] Should there be some additional message in the error about this being a panic turned error?
   Or do we just produce the exact message the panic would produce?

- [ ] This change becomes really useful if `Result::unwrap` and `Option::unwrap` become `const fn`, doing both in one go might be a good idea.

**Blockers:**

* ICE: https://github.com/rust-lang/rust/issues/66693

clarcharr:
Don't forget `expect` and all the other unwrap-like methods. :P

vi:
I suppose `unimplemented!()` and `unreachable!()` should also be included?

SimonSapin:
I think they would be "automatically", since macros don’t have a separation similar to `fn` v.s. `const fn`.

oli-obk:
I'll add some tests to #52011, but yes, it just works.

oli-obk:
Yes

IsaacWoods:
Now this is implemented, is there anything stopping `Option::unwrap` and friends from becoming `const fn`s (I'm assuming with a `rustc_const_unstable(feature = "??")`)?

Centril:
@IsaacWoods to my knowledge there shouldn't be anything in the way of that.

IsaacWoods:
@Centril great! I'll start work on a PR in that case

clarcharr:
The thing in the way of that is running `fmt::Debug::fmt` in `const` context.

oli-obk:
Yea, we need conditions in constants first

Centril:
Right; bummer.

@oli-obk Any thoughts about the first unresolved question?

oli-obk:
I'm for basically reporting it verbatim, but having the same prefix (`error[0080]:`) as all other errors.

Centril:
[...]

PR #88954, opened 2021-09-15
Allow `panic!("{}", computed_str)` in const fn.

nbdd0121:
Special-case `panic!("{}", arg)` and translate it to `panic_display(&arg)`. `panic_display` will behave like `panic_any` in cosnt eval and behave like `panic!(format_args!("{}", arg))` in runtime.

This should bring Rust 2015 and 2021 to feature parity in terms of `const_panic`; and hopefully would unblock the stabilisation of #51999.

@rustbot modify labels: +T-compiler +T-libs +A-const-eval +A-const-fn

r? @oli-obk

### DIFF
--- a/compiler/rustc_const_eval/src/transform/check_consts/check.rs
+++ b/compiler/rustc_const_eval/src/transform/check_consts/check.rs
@@ -888,10 +888,23 @@ impl Visitor<'tcx> for Checker<'mir, 'tcx> {
                 if is_lang_panic_fn(tcx, callee) {
                     self.check_op(ops::Panic);
 
+                    // `begin_panic` and `panic_display` are generic functions that accept
+                    // types other than str. Check to enforce that only str can be used in
+                    // const-eval.
+
                     // const-eval of the `begin_panic` fn assumes the argument is `&str`
                     if Some(callee) == tcx.lang_items().begin_panic_fn() {
                         match args[0].ty(&self.ccx.body.local_decls, tcx).kind() {
                             ty::Ref(_, ty, _) if ty.is_str() => (),
+                            _ => self.check_op(ops::PanicNonStr),
+                        }
+                    }
+
+                    // const-eval of the `panic_display` fn assumes the argument is `&&str`
+                    if Some(callee) == tcx.lang_items().panic_display() {
+                        match args[0].ty(&self.ccx.body.local_decls, tcx).kind() {
+                            ty::Ref(_, ty, _) if matches!(ty.kind(), ty::Ref(_, ty, _) if ty.is_str()) =>
+                                {}
                             _ => self.check_op(ops::PanicNonStr),
                         }
                     }
--- a/library/core/src/panicking.rs
+++ b/library/core/src/panicking.rs
@@ -60,6 +60,13 @@ pub fn panic_str(expr: &str) -> ! {
     panic_fmt(format_args!("{}", expr));
 }
 
+#[inline]
+#[track_caller]
+#[cfg_attr(not(bootstrap), lang = "panic_display")] // needed for const-evaluated panics
+pub fn panic_display<T: fmt::Display>(x: &T) -> ! {
+    panic_fmt(format_args!("{}", *x));
+}
+
 #[cold]
 #[cfg_attr(not(feature = "panic_immediate_abort"), inline(never))]
 #[track_caller]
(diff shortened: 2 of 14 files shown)


oli-obk:
looks good from the @rust-lang/wg-const-eval perspective

r? @m-ou-se

compiler/rustc_const_eval/src/transform/check_consts/check.rs
```diff
@@ -896,6 +896,20 @@ impl Visitor<'tcx> for Checker<'mir, 'tcx> {
                         }
                     }
 
+                    // const-eval of the `panic_display` fn assumes the argument is `&&str`
```
RalfJung, line 899: I know this is pre-existing, but it seems odd to specially check 2 out of the 4 panic functions here.

library/core/src/panicking.rs
```diff
@@ -60,6 +60,13 @@ pub fn panic_str(expr: &str) -> ! {
     panic_fmt(format_args!("{}", expr));
 }
 
+#[inline]
+#[track_caller]
+#[cfg_attr(not(bootstrap), lang = "panic_display")] // needed for const-evaluated panics
+pub fn panic_display<T: fmt::Display>(x: &T) -> ! {
```
RalfJung, line 66: Looks like every call to `panic_str` could be replaced by a call to `panic_display`... why do we need both functions?

library/std/src/rt.rs
```diff
@@ -16,6 +16,7 @@
 
 // Re-export some of our utilities which are expected by other crates.
 pub use crate::panicking::{begin_panic, begin_panic_fmt, panic_count};
+pub use core::panicking::panic_display as begin_panic_display;
```
RalfJung, line 19: Why import with a rename? That is confusing and makes the code harder to follow.

library/core/src/panicking.rs
oli-obk: We should bench that first, it may cause compile-time regressions, but if not, yea, we could just replace panic_str

library/std/src/rt.rs
nbdd0121: Haha it's really just because all these other functions are named this way. I can remove the rename.

library/core/src/panicking.rs
nbdd0121: That'll allow no_std code to do `panic!(1)`. I don't think that's desirable...

compiler/rustc_const_eval/src/transform/check_consts/check.rs
nbdd0121: Added a comment above

compiler/rustc_const_eval/src/transform/check_consts/check.rs
```diff
@@ -896,6 +900,20 @@ impl Visitor<'tcx> for Checker<'mir, 'tcx> {
...
+                        let is_str = match args[0].ty(&self.ccx.body.local_decls, tcx).kind() {
+                            ty::Ref(_, ty, _) => match ty.kind() {
+                                ty::Ref(_, ty, _) => ty.is_str(),
+                                _ => false,
+                            },
+                            _ => false,
+                        };
+                        if !is_str {
+                            self.check_op(ops::PanicNonStr);
+                        }
```
m-ou-se, line 914: Note that you could use `matches!()` here to turn this into a single `match` if you want:
```suggestion
                        match args[0].ty(&self.ccx.body.local_decls, tcx).kind() {
                            ty::Ref(_, ty, _) if matches!(ty.kind(), ty::Ref(_, ty, _) if ty.is_str()) => {}
                            _ => self.check_op(ops::PanicNonStr),
                        }
```

library/core/src/panicking.rs
nbdd0121: Oh of course, explicit args, how did I forget that :D

nbdd0121: I had a quick look and it turns out `panic_str` is needed because `non_fmt_panic` lint requires that to tell the difference between `panic!(STR)` and `panic!("{}", STR)` and generate a future compat lint for the former.

RalfJung: In that case pleas add a comment all the panic_str call site explaining this,.

nbdd0121: Added

m-ou-se:
@nbdd0121 Thanks for making this happen! :tada:

oli-obk:
new attempt at stabilization: https://github.com/rust-lang/rust/issues/89006

PR #88956, opened 2021-09-15
Update cargo

ehuss:
13 commits in e515c3277bf0681bfc79a9e763861bfe26bb05db..33ee5f82edb50af87b952c5b28de0f5fb41ebf18
2021-09-08 14:32:15 +0000 to 2021-09-17 13:51:54 +0000
- Update curl-sys (rust-lang/cargo#9917)
- Bump Cargo's curl requirement to 7.79.0 (rust-lang/cargo#9914)
- Revert "When a dependency does not have a version, git or path, fails directly" (rust-lang/cargo#9911)
- Add some contributor docs for debugging testsuite tests. (rust-lang/cargo#9904)
- Fix warnings when documenting with `--document-private-items` (rust-lang/cargo#9903)
- Improve "wrong output" error. (rust-lang/cargo#9905)
- Fix warnings from better precision of `dead_code` lint (rust-lang/cargo#9906)
- Bump to 0.58.0, update changelog (rust-lang/cargo#9900)
- Fix rustc --profile=dev unstable check. (rust-lang/cargo#9898)
- config.md: fix typo (rust-lang/cargo#9896)
- Enable some tests on windows. (rust-lang/cargo#9893)
- Enable strip test on macos. (rust-lang/cargo#9889)
- Fix `cargo fix --edition` on stable. (rust-lang/cargo#9890)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -267,7 +267,7 @@ checksum = "81a18687293a1546b67c246452202bbbf143d239cb43494cc163da14979082da"
 
 [[package]]
 name = "cargo"
-version = "0.57.0"
+version = "0.58.0"
 dependencies = [
  "anyhow",
  "atty",
@@ -882,9 +882,9 @@ dependencies = [
 
 [[package]]
 name = "curl-sys"
-version = "0.4.45+curl-7.78.0"
+version = "0.4.47+curl-7.79.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "de9e5a72b1c744eb5dd20b2be4d7eb84625070bb5c4ab9b347b70464ab1e62eb"
+checksum = "1ab94a47d9b61f2d905beb7a3d46aba7704c9f1dfcf84e7d178998d9e95f7989"
 dependencies = [
  "cc",
  "libc",


alexcrichton:
Continuing [this](https://github.com/alexcrichton/curl-rust/pull/405#issuecomment-920588639) here since this may be the best place for this.

Looking into this I'm confused what's happening. The undefined symbols are in `Curl_auth_create_digest_md5_message` and it includes symbols from `bufref.c` and also `vauth.c`. The `vauth.c` is built (via `build.rs`) on Windows and for the `spnego` feature (neither of which I think is applicable here). The `bufref.c` is only built on Windows. This means that indeed the build script isn't building the symbols required by `Curl_auth_create_digest_md5_message`. That being said the md5 function is only called by one other function in curl, `Curl_sasl_continue`, which is part of `curl_sasl.c` which is never built by `build.rs`. This means that there shouldn't actually be any dependencies on `Curl_auth_create_digest_md5_message` which is probably why this works on other targets (the `--gc-sections` linker argument coupled with the `-ffunction-sections` argument should ensure that `Curl_auth_create_digest_md5_message` is gc'd away and we ignore its unresolved symbols).

This means that the bug here could be in the illumos-specific toolchain. I see that gcc 8.4.0 is being used with binutils 2.25.1. Those aren't really all that old and I just built and ran tests for `curl` in a ubuntu:14.04 container which uses gcc 4.8.4 and binutils 2.24 and everything worked out ok. My best guess at this point is that maybe one of the illumos-specific libraries references `curl` symbols and expects to get the system curl but ends up getting our more-incomplete bundled version. I... don't know why that would happen though.

I'll try to build the illumos container locally with this PR and dig further.

alexcrichton:
Well silly me should have checked this first. The `--gc-sections` flag isn't passed on illumos due to its configuration of `linker_is_gnu: false`. I'll go try to add CI for this in the curl crate.

Mark-Simulacrum:
For any perf triagers noticing the bootstrap and wall time improvement from this PR, it is almost certainly due to enabling performance rather than on demand cpu frequency governor on the perf collector. I think this got disabled by accident at some point (likely after a reboot).

PR #88962, opened 2021-09-15
inline(always) on check_recursion_limit

fee1-dead:
r? @oli-obk 

#88558 caused a regression, this PR adds `#[inline(always)]` to `check_recursion_limit`, a possible suspect of that regression.


oli-obk:
it does get us some of the perf back (these were all small regressions in #88558)

so let's keep looking, but also

@bors r+

PR #88963, opened 2021-09-15
Coerce const FnDefs to implement const Fn traits 

fee1-dead:
You can now pass a FnDef to a function expecting `F` where `F: ~const FnTrait`.

r? @oli-obk

@rustbot label T-compiler F-const_trait_impl

### DIFF
--- a/compiler/rustc_const_eval/src/transform/promote_consts.rs
+++ b/compiler/rustc_const_eval/src/transform/promote_consts.rs
@@ -26,7 +26,6 @@ use rustc_index::vec::{Idx, IndexVec};
 use std::cell::Cell;
 use std::{cmp, iter, mem};
 
-use crate::const_eval::{is_const_fn, is_unstable_const_fn};
 use crate::transform::check_consts::{is_lang_panic_fn, qualifs, ConstCx};
 use crate::transform::MirPass;
 
@@ -656,8 +655,7 @@ impl<'tcx> Validator<'_, 'tcx> {
 
         let is_const_fn = match *fn_ty.kind() {
             ty::FnDef(def_id, _) => {
-                is_const_fn(self.tcx, def_id)
-                    || is_unstable_const_fn(self.tcx, def_id).is_some()
+                self.tcx.is_const_fn_raw(def_id)
                     || is_lang_panic_fn(self.tcx, def_id)
             }
             _ => false,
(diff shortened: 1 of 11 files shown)


fee1-dead:
@oli-obk Can we nominate this and #88558 for beta backport? It is necessary for me to go on and actually make it possible for `Iterator` to have const impls.

oli-obk:
> Can we nominate this and #88558 for beta backport?

that's not really something we do beta backports for.

Why do you need a bootstrap compiler with this? There is no syntactical component here, afaict you can use cfg(bootstrap)?

fee1-dead:
That is true, but does cfg work on bounds? Some trait functions require things like:
`where Self::Item: ~const Drop`

compiler/rustc_const_eval/src/transform/promote_consts.rs
```diff
@@ -656,8 +655,7 @@ impl<'tcx> Validator<'_, 'tcx> {
 
         let is_const_fn = match *fn_ty.kind() {
             ty::FnDef(def_id, _) => {
-                is_const_fn(self.tcx, def_id)
-                    || is_unstable_const_fn(self.tcx, def_id).is_some()
+                self.tcx.is_const_fn_raw(def_id)
```
oli-obk: why is this using raw? Won't that avoid feature gates?

fee1-dead: The original code is either stably-const or unstably-const, which can just be simplified.

oli-obk: oh I see... this is a weird piece of code

oli-obk:
I don't think it works on bounds, but you also don't need to duplicate the docs. The bootstrap version is a copy of the original function with dummy docs, while the original function gets the new attribute and a not(bootstrap) cfg

fee1-dead:
I still want the "const Drop" mega PR to be backported. It allows ~const bounds on assoc fns on inherent impls. That is an AST change which should be backported. Maybe we could cherry pick the commit? https://github.com/rust-lang/rust/pull/88558/commits/f749e05f6b95b66485c0417996d9aeb369def8da

oli-obk:
oh, I didn't know the ast validator ran on cfged out code.

Still, even if I nominate this (which I'm not sure I want to), it's likely going to get rejected by the compiler team as "not critical for beta". We only backport bugfixes, nothing else. This would be a new kind of backport, so it would probably need an MCP and all. Better to just let it ride the trains and work without libcore/libstd right now.

fee1-dead:
I can write a MCP. I don't have much to do now, anyways.

bjorn3:
You can use a macro to hide code from the ast validator.

PR #88965, opened 2021-09-15
Fast reject for NeedsNonConstDrop

fee1-dead:
Hopefully fixes the regression in #88558.

I've always wanted to help with the performance of rustc, but it doesn't feel the same when you are fixing a regression caused by your own PR...

r? @oli-obk


oli-obk:
That did the trick, thanks

@bors r+

oli-obk:

fixes the larger part of the perf regression from https://github.com/rust-lang/rust/pull/88558

PR #88968, opened 2021-09-15
Start block is not allowed to have basic block predecessors

tmiasko:
* The MIR validator is extended to detect potential violations. 
* The start block has no predecessors after building MIR, so no changes are required there.
* The SimplifyCfg could previously violate this requirement when collapsing goto chains, so transformation is disabled for the start block, which also substantially simplifies the implementation.
* The LLVM function entry block also must not have basic block predecessors. Previously, to ensure that code generation had to perform necessary adjustments. This now became unnecessary.

The motivation behind the change is to align with analogous requirement in LLVM, and to avoid potential latent bugs like the one reported in #88043.


davidtwco (approved):
LGTM!

tmiasko:
Rebased and reblessed mir opt coverage tests.

PR #88973, opened 2021-09-15
Expose the std_detect env_override feature

lu-zero:


Amanieu:
You need to add it to the `test` crate as well for `-Z build-std-features` to work.

Also keep in mind that features on standard library crates are considered unstable and may change until `-Z build-std-features` is stabilized.

lu-zero:
Thank you for spotting it! It is good to consider it unstable.

PR #87867, opened 2021-08-08
Use a separate interner type for UniqueTypeId

bjorn3:
Using symbol::Interner makes it very easy to mixup UniqueTypeId symbols
with the global interner. In fact the Debug implementation of
UniqueTypeId did exactly this.

Using a separate interner type also avoids prefilling the interner with
unused symbols and allow for optimizing the symbol interner for parallel
access without negatively affecting the single threaded module codegen.

PR #88978, opened 2021-09-15
Move the Lock into symbol::Interner

bjorn3:
This makes it easier to make the symbol interner (near) lock free in case of concurrent accesses in the future.

With https://github.com/rust-lang/rust/pull/87867 landed this shouldn't affect performance anymore.


Mark-Simulacrum:
r=me presuming perf is OK.

I'm not sure the 2nd and 3rd commits should be in this PR, but they seem okay anyway.

bjorn3:
A slight improvement for 2 benchmarks. No significant changes for the rest.

@bors r=Mark-Simulacrum

PR #88980, opened 2021-09-15
Use explicit log level in tracing instrument macro

tmiasko:
Specify a log level in tracing instrument macro explicitly.

Additionally reduce the used log level from a default info level to a
debug level (all of those appear to be developer oriented logs, so there
should be no need to include them in release builds).


tmiasko:
@cjgillot, @oli-obk  let me know if you would like to retain the existing info log level for some of those.

oli-obk:
@bors try @rust-timer queue

No, this is great. I wanted to figure out a way to set the default to debug, but until then, all of them should have an explicit level

jyn514:
@hawkw do you happen to know if there's a way to change the default log level for `#[instrument]`?

oli-obk:
@bors r+

> 
> 
> hawkw do you happen to know if there's a way to change the default log level for `#[instrument]`?

this is tracked in https://github.com/tokio-rs/tracing/issues/1070 (removing default log levels entirely and always requiring them)

PR #88981, opened 2021-09-15
rustc_codegen_llvm: make sse4.2 imply crc32 for LLVM 14

durin42:
This fixes compiling things like the `snap` crate after
https://reviews.llvm.org/D105462. I added a test that verifies the
additional attribute gets specified, and confirmed that I can build
cargo with both LLVM 13 and 14 with this change applied.

r? @nagisa cc @nikic

### DIFF
--- a/compiler/rustc_codegen_llvm/src/attributes.rs
+++ b/compiler/rustc_codegen_llvm/src/attributes.rs
@@ -305,9 +305,12 @@ pub fn from_fn_attrs(cx: &CodegenCx<'ll, 'tcx>, llfn: &'ll Value, instance: ty::
     let mut function_features = codegen_fn_attrs
         .target_features
         .iter()
-        .map(|f| {
+        .flat_map(|f| {
             let feature = &f.as_str();
-            format!("+{}", llvm_util::to_llvm_feature(cx.tcx.sess, feature))
+            llvm_util::to_llvm_feature(cx.tcx.sess, feature)
+                .into_iter()
+                .map(|f| format!("+{}", f))
+                .collect::<Vec<String>>()
         })
         .chain(codegen_fn_attrs.instruction_set.iter().map(|x| match x {
             InstructionSetAttr::ArmA32 => "-thumb-mode".to_string(),
--- a/compiler/rustc_codegen_llvm/src/llvm_util.rs
+++ b/compiler/rustc_codegen_llvm/src/llvm_util.rs
@@ -166,25 +166,32 @@ pub fn time_trace_profiler_finish(file_name: &str) {
 // Though note that Rust can also be build with an external precompiled version of LLVM
 // which might lead to failures if the oldest tested / supported LLVM version
 // doesn't yet support the relevant intrinsics
-pub fn to_llvm_feature<'a>(sess: &Session, s: &'a str) -> &'a str {
+pub fn to_llvm_feature<'a>(sess: &Session, s: &'a str) -> Vec<&'a str> {
     let arch = if sess.target.arch == "x86_64" { "x86" } else { &*sess.target.arch };
     match (arch, s) {
-        ("x86", "pclmulqdq") => "pclmul",
-        ("x86", "rdrand") => "rdrnd",
-        ("x86", "bmi1") => "bmi",
-        ("x86", "cmpxchg16b") => "cx16",
-        ("x86", "avx512vaes") => "vaes",
-        ("x86", "avx512gfni") => "gfni",
-        ("x86", "avx512vpclmulqdq") => "vpclmulqdq",
-        ("aarch64", "fp") => "fp-armv8",
-        ("aarch64", "fp16") => "fullfp16",
-        ("aarch64", "fhm") => "fp16fml",
-        ("aarch64", "rcpc2") => "rcpc-immo",
-        ("aarch64", "dpb") => "ccpp",
-        ("aarch64", "dpb2") => "ccdp",
-        ("aarch64", "frintts") => "fptoint",
-        ("aarch64", "fcma") => "complxnum",
-        (_, s) => s,
+        ("x86", "sse4.2") => {
+            if get_version() >= (14, 0, 0) {
+                vec!["sse4.2", "crc32"]
+            } else {
+                vec!["sse4.2"]
+            }
+        }
+        ("x86", "pclmulqdq") => vec!["pclmul"],
+        ("x86", "rdrand") => vec!["rdrnd"],
+        ("x86", "bmi1") => vec!["bmi"],
+        ("x86", "cmpxchg16b") => vec!["cx16"],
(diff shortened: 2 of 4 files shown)


compiler/rustc_codegen_llvm/src/attributes.rs
```diff
@@ -305,9 +305,12 @@ pub fn from_fn_attrs(cx: &CodegenCx<'ll, 'tcx>, llfn: &'ll Value, instance: ty::
...
         .target_features
         .iter()
-        .map(|f| {
+        .flat_map(|f| {
             let feature = &f.as_str();
-            format!("+{}", llvm_util::to_llvm_feature(cx.tcx.sess, feature))
+            llvm_util::to_llvm_feature(cx.tcx.sess, feature)
+                .iter()
+                .map(|f| format!("+{}", f))
+                .collect::<Vec<String>>()
```
nagisa, line 313: Probably don't need the `collect` here if you replace the `iter` with `into_iter`.

compiler/rustc_codegen_llvm/src/llvm_util.rs
```diff
@@ -166,25 +166,32 @@ pub fn time_trace_profiler_finish(file_name: &str) {
 // Though note that Rust can also be build with an external precompiled version of LLVM
 // which might lead to failures if the oldest tested / supported LLVM version
 // doesn't yet support the relevant intrinsics
-pub fn to_llvm_feature<'a>(sess: &Session, s: &'a str) -> &'a str {
+pub fn to_llvm_feature(sess: &Session, s: &str) -> Vec<String> {
```
nagisa, line 169: Returning a `&'static [&'static str]` would simplify the code in this function below significantly. Like you can then write `&["pclmul"]` instead, which is significantly cleaner (and probably more efficient)

durin42: How do I transform the input `&'a str` into a `&'static str`? I didn't see an obvious way to do the "no change" case.

compiler/rustc_codegen_llvm/src/attributes.rs
durin42: Done.

compiler/rustc_codegen_llvm/src/llvm_util.rs
nagisa: Ah, I totally forgot/missed that. You can still make `pub fn to_llvm_feature(sess: &Session, s: &str) -> Vec<&str> {` work though, I believe.

compiler/rustc_codegen_llvm/src/llvm_util.rs
```diff
@@ -253,12 +264,20 @@ fn print_target_features(sess: &Session, tm: &llvm::TargetMachine) {
...
-            let llvm_feature = to_llvm_feature(sess, *feature);
-            // LLVM asserts that these are sorted. LLVM and Rust both use byte comparison for these strings.
-            target_features.binary_search_by_key(&llvm_feature, |(f, _d)| *f).ok().map(|index| {
-                let (_f, desc) = target_features.remove(index);
-                (*feature, desc)
-            })
+            for llvm_feature in to_llvm_feature(sess, *feature) {
+                // LLVM asserts that these are sorted. LLVM and Rust both use byte comparison for these strings.
+                match target_features
+                    .binary_search_by_key(&llvm_feature, |(f, _d)| (*f).to_string())
```
nagisa, line 270:
```suggestion
                    .binary_search_by_key(&llvm_feature, |(f, _d)| f)
```

should work.

durin42: This dissolved away when I switched back to &str above.

compiler/rustc_codegen_llvm/src/llvm_util.rs
durin42: Done.

durin42:
Added an assembly test that covers the `-Ctarget-feature=+sse4.2` case.

src/test/codegen/sse42-implies-crc32.rs
```diff
@@ -0,0 +1,17 @@
...
+
+#![crate_type = "lib"]
+
+#[cfg(target_arch = "x86_64")]
+#[target_feature(enable = "sse4.2")]
+#[no_mangle]
+pub unsafe fn crc32sse(v: u8) -> u32 {
+    use std::arch::x86_64::*;
+    let mut out = !0u32;
+    _mm_crc32_u8(out, v);
```
nagisa, line 13: I think you want to do the same with `crc32` return value here as with the one in the assembly test. Otherwise the intrinsic will get optimized out as we've seen and doesn't actually test much at all.

(I see that in this instance we test the attribute, but even if the attribute is present, its plausible that something else might break down the line much the same way as the breakage this PR is fixing)

durin42: Good catch! Fixed.

nagisa:
@bors try @rust-timer queue

r=me once the commits are squashed and the perf run does not indicate any significant negative trend.

durin42:
Now that the perf run is done, rebased and squashed.

src/test/assembly/x86_64-sse_crc.rs
```diff
@@ -0,0 +1,12 @@
+// assembly-output: emit-asm
+// compile-flags: --target=x86_64-unknown-linux-gnu --crate-type staticlib -Ctarget-feature=+sse4.2
+// needs-llvm-components: x86
```
nagisa, line 3: If you want this to work, you will need to either:

1. Make this a `#[no_core]` test (allowing this test to run on any architecture);
2. Make this a `// only-x86` test instead (removing the `--target` flag and the `needs-llvm-components` annotation).

durin42:
Do you have a preference between those options?

nagisa:
The `only-x86` approach is definitely more straightforward, so its probably going to be better for you to adjust it that way. 

That said, I personally prefer the `no_core`. It allows for better cross-architecture coverage when developing on machines that execute a different instruction set.

durin42:
I got a little lost trying to figure out the `no_core` version, so I did the `only-x86_64` flavor with an amend+rebase.

src/test/codegen/sse42-implies-crc32.rs
```diff
@@ -0,0 +1,16 @@
+// min-llvm-version: 14.0
+// needs-llvm-components: x86
```
nagisa, line 2: You'll need to do exactly the same thing here as for the assembly test.

durin42: Done.

src/test/assembly/x86_64-sse_crc.rs
durin42: Done.

PR #88984, opened 2021-09-15
test - simulate default discriminant start at 1

bonega:
*WIP/TEST*
Trying to gauge the general performance impact of enums not starting with zero.
This is only for `repr(rust)`

Please sate my curiosity with a perf-run

Edit: sorry - didn't mean to set of all the sirens...

### DIFF
--- a/compiler/rustc_codegen_ssa/src/common.rs
+++ b/compiler/rustc_codegen_ssa/src/common.rs
@@ -27,7 +27,7 @@ pub enum IntPredicate {
 }
 
 pub enum RealPredicate {
-    RealPredicateFalse,
+    RealPredicateFalse = 0, // FIXME(bonega)
     RealOEQ,
     RealOGT,
     RealOGE,
--- a/compiler/rustc_hir/src/hir.rs
+++ b/compiler/rustc_hir/src/hir.rs
@@ -2631,7 +2631,7 @@ impl Item<'_> {
 #[derive(Copy, Clone, PartialEq, Eq, PartialOrd, Ord, Hash, Debug)]
 #[derive(Encodable, Decodable, HashStable_Generic)]
 pub enum Unsafety {
-    Unsafe,
+    Unsafe = 0, //FIXME(bonega)
     Normal,
 }
 
--- a/compiler/rustc_hir/src/lang_items.rs
+++ b/compiler/rustc_hir/src/lang_items.rs
@@ -20,7 +20,7 @@ use rustc_span::Span;
 use std::lazy::SyncLazy;
 
 pub enum LangItemGroup {
-    Op,
+    Op = 0, //FIXME(bonega)
 }
 
 const NUM_GROUPS: usize = 1;
@@ -44,6 +44,7 @@ macro_rules! language_item_table {
         enum_from_u32! {
             /// A representation of all the valid language items in Rust.
             #[derive(Debug, Copy, Clone, PartialEq, Eq, Hash, Encodable, Decodable)]
+            #[repr(u32)]
             pub enum LangItem {
                 $(
                     #[doc = concat!("The `", stringify!($name), "` lang item.")]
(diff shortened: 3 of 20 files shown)


the8472:
Wouldn't it be simpler to adjust the compiler logic that determines the implicit discriminants when no explicit ones are given? That'll also catch those in external crates.

bonega:
There is a couple of enums in the compiler that expects to have discriminant zero, without it being explicit.
So had to manually go sort them out, that was fun.

My first try to change the default discriminant didn't work either. Can't remember if it was because of before mentioned reason or not

bonega:
Thanks for the perf-run.

I tried to change `initial_discriminant` function to `val: 1` and disable overflow check in `wrap_incr`. 
I hoped that would mean that only overflow-tests would break.

Get a *lot* of coredumps instead.
Any tips for this?

oli-obk:
> disable overflow check in `wrap_incr`.

What does this mean?

I would have assumed we'd need to remove the -1 in https://github.com/rust-lang/rust/blob/cdeba02ff71416e014f7130f75166890688be986/compiler/rustc_middle/src/ty/util.rs#L71 and https://github.com/rust-lang/rust/blob/cdeba02ff71416e014f7130f75166890688be986/compiler/rustc_middle/src/ty/util.rs#L80 so that we essentially do +1 on overflow. There may be more sites to touch, but I would hope this would be the only place to touch

bonega:
Thanks!

But I don't want to forbid zero as discriminant.
Sorry for the confusing title, my bad - Changing it.
I want to start non explicit discriminant selection at `1`.
If it wraps around to `0` then that is fine.

I think your change would make it so that we can only fit 255 variants in an `repr(u8)`?

Anyhow the change still seem to result in same problem: Lot of coredumps

oli-obk:
right, I misunderstood the intention. Just wrapping around normally is good then. Can you push your changes somewhere so we can be sure to be talking about the same things?

bonega:
Pushed my best guess.

Intention: For `repr(rust)` enums with no explicit discriminants, use `1` instead of `0` for initial discriminant.

I get a lot of coredumps for this.
Unsure how to best debug/structure it.
Thankful for any tips

oli-obk:
my recommendation for debugging it is putting it behind a feature gate and then creating a very simple test.

compile the test once with the feature and once without, but with explicit discriminants that should be equal and compare various tracing dumps between the two runs to see how things differ.

tmiasko:
What is motivation behind an additional optimize argument to `AdtDef::discriminants`? In commit above, the MIR building uses false, while uninhabited enum branches through `discriminant_for_variant` will potentially use true, so at least one arm of every match one enum using the optimization will be completely removed out at MIR level `:)`.

I would also look at codegen / interpreter of niche encoding, if I recall correctly it assumes that discriminants start at 0.

bonega:
Thanks to both of you.

I have pushed a new version that only has one test failing for `/x.py test ui`.
`rust/src/test/ui/mir/issue-76803-branches-not-same.rs`.
Will think about it some more

tmiasko:
```console
$ cat a.rs 
#[derive(Debug, Eq, PartialEq)]
pub enum Type {
    A = 1,
    B = 2,
}
pub fn encode(v: Type) -> Type {
    match v {
        Type::A => Type::B,
        _ => v,
    }
}
fn main() {
  assert_eq!(Type::B, encode(Type::A));
}
$ rustc -Zmir-opt-level=0 a.rs && ./a
$ rustc a.rs && ./a
  left: `B`,
 right: `A`', a.rs:13:3
```

...

bonega:
Good catch.
I was thinking it was my code...

JohnCSimon:
triage: 
@bonega Can you please address the build failures  and adjust the label with `@rustbot ready` when you're ready for review.

bonega:
Been stuck on this for a while, will try and look at it again

JohnCSimon:
Ping from triage: 
@bonega 

@rustbot label: +S-inactive

PR #88985, opened 2021-09-15
Update clobber_abi list to include k[1-7] regs

ghost:


nagisa:
Please adjust the PR description to describe why this change is being made even if it is just a very brief

> These registers have been accepted since such and such #abcdef PR. This brings the documentation to match the implementation.

Or somesuch.

PR #88986, opened 2021-09-15
Update the backtrace crate

hargoniX:
https://github.com/rust-lang/backtrace-rs/pull/437 fixed backtraces in
OpenBSD -> update it here as well so OpenBSD Rust code can produce
proper backtraces.


hargoniX:
@semarie This might be an interesting diff for the OpenBSD Rust port for 7.0 before the freeze happens? It's very small but adds proper backtraces to OpenBSD Rust again.

PR #86034, opened 2021-06-05
Change entry point to 🛡️ against 💥 💥-payloads

nagisa:
Guard against panic payloads panicking within entrypoints, where it is
UB to do so.

Note that there are a number of tradeoffs to consider. For instance, I
considered guarding against accidental panics inside the `rt::init` and
`rt::cleanup` code as well, as it is not all that obvious these may not
panic, but doing so would mean that we initialize certain thread-local
slots unconditionally, which has its own problems.

Fixes #86030 
r? @m-ou-se

PR #88988, opened 2021-09-15
Avoid codegen for Result::into_ok in lang_start

Mark-Simulacrum:
This extra codegen seems to be the cause for the regressions in max-rss on #86034. While LLVM will certainly optimize the dead code away, avoiding it's generation in the first place seems good, particularly when it is so simple.

#86034 produced this [diff](https://gist.github.com/Mark-Simulacrum/95c7599883093af3b960c35ffadf4dab#file-86034-diff) for a simple `fn main() {}`. With this PR, that diff [becomes limited to just a few extra IR instructions](https://gist.github.com/Mark-Simulacrum/95c7599883093af3b960c35ffadf4dab#file-88988-from-pre-diff) -- no extra functions.

Note that these are pre-optimization; LLVM surely will eliminate this during optimization. However, that optimization can end up generating more work and bump memory usage, and this eliminates that.

### DIFF
--- a/library/std/src/rt.rs
+++ b/library/std/src/rt.rs
@@ -59,10 +59,10 @@ fn lang_start<T: crate::process::Termination + 'static>(
     argc: isize,
     argv: *const *const u8,
 ) -> isize {
-    lang_start_internal(
+    let Ok(v) = lang_start_internal(
         &move || crate::sys_common::backtrace::__rust_begin_short_backtrace(main).report(),
         argc,
         argv,
-    )
-    .into_ok()
+    );
+    v
 }


library/std/src/rt.rs
```diff
@@ -59,10 +59,11 @@ fn lang_start<T: crate::process::Termination + 'static>(
     argc: isize,
     argv: *const *const u8,
 ) -> isize {
-    lang_start_internal(
+    match lang_start_internal(
```
nagisa, line 62: Looks like just `let Ok(v) = ...` would work?

nagisa (approved):
I feel like we may want to open-code `map_err` in the function above as well which would likely help too.

Mark-Simulacrum:
> I feel like we may want to open-code map_err in the function above as well which would likely help too.

lang_start_internal is already codegen'd as part of std, so unless we're doing LTO or similar, it's body doesn't really matter. If  LTO is running, then based on some local testing it looks like we pull in tons of code from std anyway (and backtrace etc) so it really doesn't matter whether we're using map_err or not.

I also want to avoid doing too much in this particular PR - makes it harder to compare things in perf, for example - so I am going to go ahead and @bors r=nagisa

Issue #27730, opened 2015-08-12
Tracking issue for `NonZero`/`Unique`/`Shared` stabilization


aturon:
We currently have three internal types, `NonZero`, `Unique` and `Shared`, that are very useful when working with unsafe Rust code. Ultimately, the hope is to have a clear vision for semi-high-level, but unsafe programming using types like these.

None of the current APIs have gone through the RFC process, and a comprehensive RFC should be written before any stabilization is done.

cc @Gankro

Gankro:
I'm not a fan of NonZero. I would like a more robust system that can hook into the things described in https://github.com/rust-lang/rfcs/issues/1230. 

Unique is great, but I would like to rename it to Owned to better reflect what its actual semantics are (e.g. it's fine to have other pointers into it just like it's fine for Box). 

I would like to add a Shared equivalent for Rc/Arc to use. This is blocked by https://github.com/rust-lang/rust/issues/26905

huonw:
I expanded this issue to cover the new `Shared` type (https://github.com/rust-lang/rust/pull/29110).

Gankro:
CC @SimonSapin 

Does Shared/Unique cover all your NonZero usecases?

eddyb:
I commented about this in #29110, reposting here for visibility:

The `Deref` impls on these types are bad interfaces because they remove the optimizations NonZero can trigger, specifically the control-flow ones, not the layout optimizations, which are unaffected.

They also allow writing `*p` where `p` is an `ptr::Unique<T>` or a `ptr::Shared<T>` to obtain a `*mut T`, which might make unsafe code more confusing (you need `**p` to get at the `T` lvalue).

`NonZero` and wrappers such as `Unique` and `Shared`, should use a `Cell`-like API (`.get()` and `.set()` - or just `.get()`, really), allowing LLVM to reason about their non-nullability.

Gankro:
@eddyb alternatively, we could just promote these types to compiler primitives, right? Have them coerce to raw pointers or summat? This may also improve debug perf, as there's less abstraction in the way if it's really "just" a raw pointer, and not a Foo(Bar(*const T)) that is accessed through two levels of methods.

alexreg:
I agree with @eddyb's comment about using a Cell-like API for `NonZero` (and other types). It makes semantic sense.

alexcrichton:
[...]

PR #88990, opened 2021-09-15
core::num::NonZero<N> type alias

dtolnay:
This PR implements `NonZero<N>` as I proposed in https://github.com/rust-lang/rust/issues/82363#issuecomment-906658138, and deletes unstable `NonZero_c_*`.

`NonZero<N>` is a generic type alias for the non-zero type that corresponds to each integer primitive.

This type alias simply provides an alternative spelling of the various `NonZero` structs found in [`core::num`](https://doc.rust-lang.org/nightly/core/num/index.html). For example `NonZero<u16>` refers to the struct [`core::num::NonZeroU16`](https://doc.rust-lang.org/nightly/core/num/struct.NonZeroU16.html).

**Note:** this is *not* intended for writing code that is generic over multiple types of non-zero integers. The relevant trait bound you would need for that is not exposed. The only thing this is, is an alternative spelling.

## Example

Where this comes in handy is if the primitive types themselves are being used via a type alias, such as [`std::os::raw::c_char`](https://doc.rust-lang.org/nightly/std/os/raw/type.c_char.html). The `c_char` type refers to either `i8` or `u8` depending on whether the platform's C character type is signed or unsigned, and without `NonZero<N>` there is not a straightforward way to name either `NonZeroI8` or `NonZeroU8` depending on which one a non-zero `c_char` corresponds to.

```rust
#![feature(nonzero_generic_type_alias)]

use std::num::NonZero;
use std::os::raw::c_char;

// A separator that we're planning to use with nul-terminated C strings,
// where \0 would be nonsensical.
pub struct Separator {
    ch: NonZero<c_char>,
}

impl Separator {
    pub fn new(ch: c_char) -> Option<Self> {
        NonZero::new(ch).map(|ch| Separator { ch })
    }
}
```

### DIFF
--- a/library/core/src/num/mod.rs
+++ b/library/core/src/num/mod.rs
@@ -59,6 +59,9 @@ pub use dec2flt::ParseFloatError;
 #[stable(feature = "rust1", since = "1.0.0")]
 pub use error::ParseIntError;
 
+#[unstable(feature = "nonzero_generic_type_alias", issue = "82363")]
+pub use nonzero::NonZero;
+
 #[stable(feature = "nonzero", since = "1.28.0")]
 pub use nonzero::{NonZeroU128, NonZeroU16, NonZeroU32, NonZeroU64, NonZeroU8, NonZeroUsize};
 
--- a/library/core/src/num/nonzero.rs
+++ b/library/core/src/num/nonzero.rs
@@ -8,6 +8,55 @@ use super::from_str_radix;
 use super::{IntErrorKind, ParseIntError};
 use crate::intrinsics;
 
+/// Generic type alias for the non-zero type that corresponds to each integer
+/// primitive.
+///
+/// This type alias simply provides an alternative spelling of the various
+/// `NonZero` structs found in [`core::num`]. For example `NonZero<u16>` refers
+/// to the struct [`core::num::NonZeroU16`].
+///
+/// **Note:** this is *not* intended for writing code that is generic over
+/// multiple types of non-zero integers. The relevant trait bound you would need
+/// for that is not exposed. The only thing this is, is an alternative spelling.
+///
+/// # Example
+///
+/// Where this comes in handy is if the primitive types themselves are being
+/// used via a type alias, such as [`std::os::raw::c_char`][c_char]. The
+/// `c_char` type refers to either `i8` or `u8` depending on whether the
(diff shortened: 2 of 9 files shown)


kennytm:
heh. so we have come to a full circle of introducing `NonZero<T>` again. why was rust-lang/rfcs#2307 accepted in the first place 🤷

dtolnay:
Shrug, this is not at all the `NonZero` from #27730. Every single benefit from https://github.com/rust-lang/rfcs/pull/2307 still applies. I can't tell if you are just joking but this is forward progress, not full circle.

kennytm:
Well both? 

Surely the primary purpose of of rust-lang/rust#2307 is to remove the *public* `Zeroable` trait. But now the role of that trait comes back as the *private* `IntegerPrimitive` trait, which API-wise the trait's privacy is the only difference compared with #27730. The private trait negates the entire Motivation of introducing the 12 distinct types since users can't `impl NonZero<CustomType>` at all:

> One problem of the current API is that it is unclear what happens or what should happen to `NonZero<T>` or `Option<NonZero<T>>` when `T` is some type other than a raw pointer or a primitive integer. **In particular, crates outside of `std` can implement `Zeroable` for their arbitrary types since it is a *public* trait.**
>
> To avoid this question entirely, this RFC proposes replacing the generic type and trait with twelve concrete types in `std::num`, one for each primitive integer type. This is similar to the existing atomic integer types like `std::sync::atomic::AtomicU32`.

So I think it is only natural to take one step further (backward?) and make it

```rust
#[repr(transparent)]
#[rustc_layout_scalar_valid_range_start(1)]
#[rustc_nonnull_optimization_guaranteed]
pub struct NonZero<T: IntegerPrimitive>(T);

// ... then most impl can be freed from the macros ...

// stability compatibility ↓

pub type NonZeroU8 = NonZero<u8>;
...
pub type NonZeroI128 = NonZero<i128>;
```

kennytm:
i'm going to wait till #82363's current FCP is **canceled** since this PR removes the `NonZero_c_*` types.

dtolnay:
Kenny's writeup in https://github.com/rust-lang/rust/issues/82363#issuecomment-921513910 is compelling. Gonna close this in favor of `struct NonZero` :slightly_smiling_face:

Issue #73131, opened 2020-06-08
Tracking Issue for std::process error handling


ijackson:
Hi.  This is a tracking issue for various error handling awkwardnesses in `std::process`.  It seemed useful to gather them together here.  I hope you consider this helpful.  I don't think this needs an RFC but I can make an RFC if people feel that would be best.

### About tracking issues

Tracking issues are used to record the overall progress of implementation.
They are also uses as hubs connecting to other relevant issues, e.g., bugs or open design questions.
A tracking issue is however *not* meant for large scale discussion, questions, or bug reports about a feature.
Instead, open a dedicated issue for the specific matter and add the relevant feature gate label.

### Platform-independent issues

- [ ] #70186 request for `#[must_use]` on `std::process::Child`
- [ ] #73127 request for `#[must_use]` on `std::process::ExitStatus`
- [ ] #73126 request for less hazardous version of `output()`
- [ ] #73125 request for affordance to make `Result` from `ExitStatus`

### Unix issues relating to `ExitStatusExt`

- [ ] #73128 request for `std::os::unix::process::ExitStatusExt` `into_raw`
- [ ] #73129 request for `std::os::unix::process::ExitStatusExt` `.coredumped()` method

KodrAus:
@ijackson Are you happy if we collapse those child issues into this one and consider everything we want to do to improve our `std::process` error handling together?

KodrAus:
I was thinking our path could look something like:

1. Design our API for converting `ExitStatus` into a `Result`.
2. Update our docs (in particular our docs around `ExitStatus` are pretty sparse).
3. Get the `ExitStatus` into `Result` API stabilized and update our examples to use it instead of ignoring them.
4. Consider `#[must_use]` on `ExitStatus`.

What do you think?

KodrAus:
> On unix, failure to use a Child leaks the process. It continues running in parallel, even after we have exited (or crashed), possibly interfering with the user's terminal or driveling nonsense into a logfile long after the original event.

In #81452, am I reading that right and you're saying it's possible to spawn a child that outlives its parent process entirely?

KodrAus:
As for what to make the API for `ExitStatus` into `Result` look like, I arrived at `success_or_else` after thinking:

- We should try give our methods that convert a non `Option`/`Result`/`Poll`-like type into one a more semantic name, so that you can identify it in a chain of combinators.
[...]

Issue #73126, opened 2020-06-08
std::process::Command output() method error handling hazards


ijackson:
Hi. Normally, Rust makes it difficult to accidentally write buggy code. This is one of its great strengths. However, the API of `output()` on `std::process::Command `API has two serious error handling hazards:

1. It requires the user to explicitly fish out the program's exit status and check it.
2. It requires the user to explicitly deal, somehow, with the program's stderr output (if any).

See the example below.

I think this is very difficult to fix with the current return value from `output()`.  It seems to me that there should be a new function whose behaviour is as follows:
 * Unless the user explictly called `.stderr(...)` when building the `Command`, any nonempty stderr is treated as an error, giving an `Err` return value (whose `Debug` impl prints the stderr output).
 * Nonzero exit status is treated as an error, giving an `Err` return value,
 * In case of nonzero exit status together with nonempty stderr output (the usual case) the error object contains both (and its `Debug` impl displays both).
 * The returned value is purely the actual stdout.

I don't know what this should be called.  Unfortunately the name `output` has been taken for the more hazardous, raw, function.  Which we must retain because if you want to run a command that sometimes succeeds returning nonzero (eg, `diff`) you need something like it.

(See also #70186 which is about the return value from `spawn()`.  I am about to file another issue about the return value from `wait()`)

```rust
use std::process::*;
use std::io::stdout;
use std::io::Write;

fn main() {
    let command = ["ls","--no-such-option"];
    
    let output = Command::new(command[0])
                     .args(&command[1..])
                     .output()
                     .expect("output() failed");
    println!("Output from {:?}:", &command);
    stdout().write_all(output.stdout.as_slice()).expect("write failed");
    
    println!("All went well!")
}
```

([Playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=2dd5c16d2d5bf4306c70b8fd92cf5f7f))

Actual output:

```
   Compiling playground v0.0.1 (/playground)
    Finished dev [unoptimized + debuginfo] target(s) in 0.72s
     Running `target/debug/playground`

Standard Output

Output from ["ls", "--no-such-option"]:
All went well!
```

Expected output:

Some kind of compiler warning.  Or something in the docs to say not to use `.output()` (and, therefore, something convenient to use instead).

joshtriplett:
[...]

PR #89004, opened 2021-09-16
Add convenience API to std::process::Command

matklad:
This is similar in spirit to `fs::read_to_string` -- not strictly
necessary, but very convenient for many use-cases.

Often you need to write essentially a "bash script" in Rust which spawns
other processes. This usually happens in `build.rs`, tests, or other
auxiliary code. Using std's Command API for this is inconvenient,
primarily because checking for exit code (and utf8-validity, if you need
output) turns one-liner into a paragraph of boilerplate.

While there are crates.io crates to help with this, using them often is
not convenient. In fact, I maintain one such crate (`xshell`) and, while
I do use it when I am writing something substantial, for smaller things
I tend to copy-paste this *std-only* snippet:

https://github.com/rust-analyzer/rust-analyzer/blob/ae36af2bd43906ddb1eeff96d76754f012c0a2c7/crates/rust-analyzer/build.rs#L61-L73

So, this PR adds two convenience functions to cover two most common
use-cases:

* `run` is like `status`, except that it check that `status` is zero
* `read_stdout` is like `output`, except that it checks status and
  decodes to `String`. It also chomps the last newline, which is modeled
  after shell substitution behavior (`echo -n "$(git rev-parse HEAD)"`)
  and Julia's [`readchomp`](https://docs.julialang.org/en/v1/manual/running-external-programs/)

Note that this is not the ideal API. In particular, error messages do
not include the command line or stderr. So, for user-facing commands,
you'd have to write you own code or use a process-spawning library like
`duct`.

### DIFF
--- a/library/std/src/process.rs
+++ b/library/std/src/process.rs
@@ -1029,6 +1029,84 @@ impl Command {
     pub fn get_current_dir(&self) -> Option<&Path> {
         self.inner.get_current_dir()
     }
+
+    /// Convenience wrapper around [`Command::status`].
+    ///
+    /// Returns an error if the command exited with non-zero status.
+    ///
+    /// # Examples
+    ///
+    /// ```no_run
+    /// # #![feature(command_easy_api)]
+    /// use std::process::Command;
+    ///
+    /// let res = Command::new("cat")
+    ///                   .arg("no-such-file.txt")
+    ///                   .run();
+    /// assert!(res.is_err());
+    /// ```
+    #[unstable(feature = "command_easy_api", issue = "none")]
+    pub fn run(&mut self) -> io::Result<()> {
+        let status = self.status()?;
+        self.check_status(status)
+    }
+
+    /// Convenience wrapper around [`Command::output`] to get the contents of
+    /// standard output as [`String`].
+    ///
+    /// The final newline (`\n`) is stripped from the output. Unlike
+    /// [`Command::output`], `stderr` is inherited by default.
+    ///
+    /// Returns an error if the command exited with non-zero status or if the
+    /// output was not valid UTF-8.
+    ///
+    /// # Examples
+    ///
+    /// ```no_run
+    /// # #![feature(command_easy_api)]
+    /// use std::process::Command;
+    /// let output = Command::new("git")
+    ///                      .args(["rev-parse", "--short", "1.0.0"])
+    ///                      .read_stdout()?;
+    ///
+    /// assert_eq!(output, "55bd4f8ff2b");
+    /// # Ok::<(), std::io::Error>(())
+    /// ```
+    #[unstable(feature = "command_easy_api", issue = "none")]
+    pub fn read_stdout(&mut self) -> io::Result<String> {
+        // FIXME: This shouldn't override the stderr to inherit, and merely use
+        // a default.
+        self.stderr(Stdio::inherit());
+
+        let output = self.output()?;
+        self.check_status(output.status)?;
+        let mut stdout = output.stdout;
+        if stdout.last() == Some(&b'\n') {
+            stdout.pop();
+        }
+        String::from_utf8(stdout).map_err(|_| {
+            io::Error::new_const(
+                io::ErrorKind::InvalidData,
+                format!("command {:?} produced non-UTF-8 output"),
+            )
+        })
+    }
+
+    fn check_status(&self, status: ExitStatus) -> io::Result<()> {
+        if status.success() {
+            Ok(())
+        } else {
+            Err(io::Error::new_const(
+                io::ErrorKind::Uncategorized,
+                match status.code() {
+                    Some(code) => {
+                        format!("command {:?} exited with non zero status ({})", self, code)
+                    }
+                    None => format!("command {:?} was terminated", self),
+                },
+            ))
+        }
+    }
 }
 
 #[stable(feature = "rust1", since = "1.0.0")]


matklad:
Will file a tracking issue if this is something we'd want to have at all!

est31:
cc #84908

est31:
What about renaming `read_stdout` to `read_stdout_to_string` or something? Because in `std::fs`, there is a `read` function, which outputs a binary `Vec<u8>`, and a `read_to_string` function which outputs an UTF-8 encoded `String`. On the other hand, I think it's way easier to retrieve binary outputs of a process than it is to read a file into a `Vec<u8>` without using `fs::read`, so maybe there wouldn't be much reason to add `read_stdout` ever.

matklad:
Yeah, the name/API here is definitely not "orthogonal". For full generality, we'd need a matrix with binary/text stdour/stderr/both. But, as this is a convenience API, I figured it's better to make the common case short. We have the precedent in missing "_to_string": we have `read_line` rather than `read_line_to_string`. 

Another thing which is reasonable to have is  `fn checked_output(&mut) -> io::Result<(Vec<u8>, Vec<u8>)>`, but then perhaps we need to add `checked_`  prefix to `read_stdout`. 

Overall, I feel that there's certainly an unfortunately uncomfortably large design space here, and it's unclear how to best explore it. So, on the meta level, I propose reaching consensus in the PR on that:

* this is a problem worth solving (I became convinced of that after realising that the status quo hurts versatility quite a bit)
* the API is of the right general shape (maybe it is not, #84908 is a plausible alternative)

After that, we can merge and iterate on the tracking issue.

the8472:
I think these methods would be better on `Output`. That way you don't throw away information you might want to check when encountering an non-zero exit code.

matklad:
I don't think `run` will work with `.output`, as it, like `.status` inherits streams. We can add this to `ExitStatus`, and that'll give us #84908. 

I think I still prefer the API to be available on the command directly. While #84908 is a bit more general, we already have a fully general API for arbitrary custom error management. What we don't have is a sane-defaults one-liner.

the8472:
> I don't think run will work with .output, as it, like .status inherits streams.

The *it* is quite ambiguous here, I had to look at the methods to get what you actually mean 😅

Ok, `run()` makes sense as-is since it inherits the streams since you don't lose any information. But `read_stdout` would still make sense on `Output` so you can split it off to a new line if you also want to get the stderr.

yaahc:
> Overall, I feel that there's certainly an unfortunately uncomfortably large design space here, and it's unclear how to best explore it. So, on the meta level, I propose reaching consensus in the PR on that:
> 
> * this is a problem worth solving
> * the API is of the right general shape

I agree that this is a good place to start building consensus. My opinion is that this is a problem worth solving. I've built similar abstractions on top of the `Command` APIs quite a few times, including most recently for this [test suite](https://github.com/ZcashFoundation/zebra/blob/main/zebra-test/src/command.rs). However, in these I'm generally quite careful to preserve the [full context of the command that was run](https://github.com/ZcashFoundation/zebra/blob/main/zebra-test/src/command.rs#L691-L742), why it failed, and all of its output. I see the importance of having some of these APIs be more convenient but I want to make sure we don't oversimplify.

Regarding the currently suggested API's shape:

My first issue is that `run` seems to be almost identical to `.status()?.exit_ok()?;` except that it uses a custom error message with less information than `ExitStatusError` and combines that into the `io::Error`. I don't think `run` is appreciably more convenient to use than this alternative, and between the two I'd prefer to keep the error types distinct than merge them.

Second, I agree with @the8472, `read_line` should be a method on `Output` not on `Command`, similar to how I prefer the `exit_ok` method on `ExitStatus` to `run` on `Command`.

Overall `read_line` seems more strongly motivated than `run`, though I really don't love the idea of an API that encourages users to throw away so much context. I worry that throwing away `stderr` as a convenient default would lead to a bunch of rust CLI tools that have subpar error reporting whenever they interact with subprocesses. My feeling is that if we're going to provide a convenience API like this is should be proactive and opinionated in a way that preserves as much context about errors as possible.

On that note, I'm curious how other language's convenience APIs for capturing the output of a subprocess handle stderr. Does Julia's `readchomp` throw away stderr? Does this change if the exit status indicates an error? Do they discard the `stderr` or forward it or something else entirely?

matklad:
>I don't think run is appreciably more convenient to use than this alternative

Yeah, as a designer of the API, I also see how two calls isn't really a big deal, and how this is cleaner and more orthogonal. 

However, from an API user point of view, the difference seem rather meaningful. As a user of the API, I want to "run the command", or "get the command's output". It's a single atomic operation. As a metaphor, when running commands in a terminal, I don't think whether I misstyped the command name or the argument name, both are just "command failed to run":

```console
$ cargo tst
error: no such subcommand: `tst`

	Did you mean `test`?

$ carg test
carg: command not found
```

For use-case I have in mind, one would always just write `.status()?.exit_ok()?;`, and that would be just boilerplate and noise. 

It's not *that* big deal (although the extra possibility of forgetting to check the exit status is a bit concerning), but, still, it would be odd if shortcut API included more than strictly necessary. 

Regarding the context, one interesting bit here is that, by letting the API hang off `Command` itself, we actually leave ourselves room to add more context later. In particular, we can add the most crucial bit of info, the command line itself, into the error message. In the `Output` or `ExitStatus` we don't have access to the `Command`. 

This is similar to `fs::read_to_string` -- that also conflates three separate error cases (can't open path, can't read from fd, can't decode to str) into just one `io::Error`, but that is *exactly* what you need when you operate at a high level of abstraction, where you want to "just read the file". 

I sort-of feel that we are conflating three use-cases here:

* Alice writes a CLI utility to orchestrate processes and wants top-notch error reporting from subprocesses.
* Bob migrates ad-hoc automation scripts from bash to Rust.
* Carol writes a generic application and just wants all errors to be maximally useful by default.

I feel that Alice is rather happy -- the current API already allows to do fine-grained error reporting. She only needs [`read2`](https://github.com/rust-lang/cargo/blob/58a961314437258065e23cb6316dfc121d96fb71/crates/cargo-util/src/read2.rs) stabilized in some form, as she is tired from copy-pasting this bit of tribal knowledge. 

Bob has become considerably happier since the introduction of `fs::read_to_string`, but he cringes every time he needs to shell out to git and do the usual output, check exit status, decode stdout dance. 

Carol feels that the language is a joke: seriously, upon failing to read a file you get *neither* the filename *nor* backtrace by default?

The proposed API helps Bob a lot. If at some point we'll learn how to add more context to `io::Error`s out of the box, Carol will also benefit from it. Orthogonality and context factoring apply to Alice use-case, but it is already served by exiting APIs.

>On that note, I'm curious how other language's convenience APIs for capturing the output of a subprocess handle stderr.

Good call! It's interesting that Python's `check_output` and `check_call` both just inherit stderr. ~~Julia is still compiling~~ Julia behaves the same. That's an argument for actually changing the semantics of `read_stdout` to *not* just forward to `.output` but to use a different streams setup:

* `output` defaults to capturing both stout&stderr
* `read_stdout` should probably default to capturing stdout and inhereting stderr

matklad:
>It's not that big deal

I've realized that that's actually a bigger deal:

```rust
use anyhow

// V1
let stdout = command.read_stdout()
    .with_context(|| format!("failed to run {:?}", comand))?;

// V2
let stdout = command.output()
    .with_context(|| format!("failed to run {:?}", comand))?
    .read_stdout()
    .with_context(|| format!("failed to run {:?}", comand))?;
```

yaahc:
The V2 version isn't as bad if you use try blocks but yea I see what you're saying, particularly the bit about wanting errors to be maximally useful by default. I worry about being able to pick a format that actually works for everyone. My fear being that it will be too long or will be missing env information that is relevant, but I think it's worth a try as long as it doesn't affect reporting of existing usages of `Command`.

If you're going to go with this argument though I don't think this PR should be merged as a half measure. I'd want to see this additional reporting of the command itself added to the new `run` method before landing it, along with it having `read_stdout` inherit stderr as you suggested.

library/std/src/process.rs
```diff
@@ -1029,6 +1029,84 @@ impl Command {
...
+        }
+        String::from_utf8(stdout).map_err(|_| {
+            io::Error::new_const(
+                io::ErrorKind::InvalidData,
+                format!("command {:?} produced non-UTF-8 output"),
+            )
+        })
+    }
+
+    fn check_status(&self, status: ExitStatus) -> io::Result<()> {
```
ijackson, line 1095: I wonder if this would be usefully `pub`.

library/std/src/process.rs
```diff
@@ -1029,6 +1029,84 @@ impl Command {
...
+    /// use std::process::Command;
+    /// let output = Command::new("git")
+    ///                      .args(["rev-parse", "--short", "1.0.0"])
+    ///                      .read_stdout()?;
+    ///
+    /// assert_eq!(output, "55bd4f8ff2b");
+    /// # Ok::<(), std::io::Error>(())
+    /// ```
+    #[unstable(feature = "command_easy_api", issue = "none")]
+    pub fn read_stdout(&mut self) -> io::Result<String> {
```
ijackson, line 1076: Bikeshed: I suggest `get_stdout` for the name?  To me, `read_stdout` fails to suggest that it also waits for termination.

ijackson:
> I think these methods would be better on `Output`. That way you don't throw away information you might want to check when encountering an non-zero exit code.

`Output` has a missing-error-check a beartrap and is, I think, unfixable.  So IMO `.output()` needs to be deprecated, or at least have a big health warning.  So it needs to be replaced.  I don't think it's worth trying toi improve it much.  The `.check_status()` method can't be on `Output` anyway because `Output` doesn't have the command and arguments any more.

ijackson:
@matklad Thank you so much for this MR.  IMO we absolutely definitely need both of the things you are doing here.  See my tracking issue #73131.

I wonder if it would be easier to do your two halves in two separate MRs?  I'm hoping `run` is less controversial.  The output handling one is complex because of the need to make choices about UTF-8 or bytes, and naming. 

I've made a few comments on details.  See also #73126

ijackson:
> My first issue is that `run` seems to be almost identical to `.status()?.exit_ok()?;` except that it uses a custom error message with less information than `ExitStatusError` and combines that into an `io::Error`. I don't think `run` is appreciably more convenient to use than this alternative, and between the two I'd prefer to keep the error types distinct than merge them.

I think `run()`s implementation here isn't quite right.  I think it should have its own error type, not just use `String`.  But the snippet `.status()?.exit_ok()?` is actually wrong.  As you note in https://github.com/rust-lang/rust/pull/88306#pullrequestreview-771911506 it doesn't print the command that failed.  What you have to do instead is some kind of `.map_err()` and `.context()` and so on - assuming you are using a portmanteau error type like `anyhow` or `eyre`.  If you're not you don't have many good options.


I found @matklad's use case personae in https://github.com/rust-lang/rust/pull/89004#issuecomment-923803209 very convincing.  I have been (roughly speaking) all of those people.

Alice is well-served apart from the fact that `output()` has a lost error beartrap (#73131) and she, by her nature, will probably notice that and DTRT.

Bob and Carol are very ill-served.  We should provide facilities for them that DTRT conveniently.

I think being able to conveniently bundle a subprocess failure as an `io::Error` is quite important f0or that.  And this MR demonstrates that it can be done without compromising on error message quality.  Of course someone like Alice won't want to use a cooked API like `run` but Alice will realise that and can use `status()` and so on.

> Second, I agree with @the8472, `read_line` should be a method on `Output` not on `Command`, similar to how I prefer the `exit_ok` method on `ExitStatus` to `run` on `Command`.

See my complaints in #73131 about `Output`.  I don't think it can be fixed without replacing it.
 
> On that note, I'm curious how other language's convenience APIs for capturing the output of a subprocess handle stderr. Does Julia's `readchomp` throw away stderr? Does this change if the exit status indicates an error? Do they discard the `stderr` or forward it or something else entirely?

I can't speak to Julia.  I have done a lot of this kind of thing in Perl and quite a bit in Tcl, and also some in Python.  Frankly, this area is rather subtle and most languages have made a mess of it.  Rust's current `Command` is not as bad as some, at least.

Perl's backquote operator has the command inherit the script's stderr.  If you want to do something else you need to do a lot of hand-coding.  It reports errors in a funky and not particularly convenient way (but, frankly, that is typical for Perl).  It doesn't trim a final newline but Perl has a `chomp` operator that does that very conveniently.

Tcl's `exec` captures stderr by default (there are facilities for redirecting or inheriting it), calling any stderr an error (throwing a Tcl exception).  It always calls nonzero exit status an error (and there is no easy way to get the stdout separately from the error in that situation).  It unconditionally chomps a final newline.

Python3's `subprocess.run` fails to call nonzero exit status an exception.  If you don't write explicit error handling code (highly unusual in Python) you have a lost output bug.

### What I would like

 * `run()` roughly like this, only with a custom error type
 * `get_output_bytes()` gives you a `Vec<u8>`
 * `get_output()` converts to UTF-8, does not strip newline
 * `get_output_line()` calls it an error if the result is not precisely one newline-terminated line.
 * `get_output_read()` gives you `impl Read` which will wait on EOF, and then call any nonzero status an error.  That lets you do error-checked streaming reading without risk of deadlock.
 * Strong health warning in the docs for `output()`, maybe even deprecation

matklad:
Yeah, evidently I've run out of steam on this one, so closing :)

ijackson:
@matklad thanks for your hard work, and sorry if I contributed to bogging it down.

Please comment on my https://github.com/rust-lang/rfcs/pull/3362

PR #89011, opened 2021-09-16
Restructure std::rt

bjorn3:
These changes should reduce binary size slightly while at the same slightly improving performance of startup, thread spawning and `std::thread::current()`. I haven't verified if the compiler is able to optimize some of these cases already, but at least for some others the compiler is unable to do these optimizations as they slightly change behavior in cases where program startup would crash anyway by omitting a backtrace and panic location.

I can remove 6f6bb16 if preferred.


bjorn3:
Binary sizes for `fn main() {}`:

config|before this PR|after this PR|diff (bytes)
---|---|---|---
-|3_639_256|3_627_968|-11_288
\- (debuginfo stripped)|360_976|361_024|+48
\- (fully stripped)|297_216|297_288|+72
`-Copt-level=3`|3_638_112|3_626_824|-11_288
`-Copt-level=3` (debuginfo stripped)|360_128|360_176|+48
`-Copt-level=3` (fully stripped)|297_216|297_288|+72
`-Copt-level=3 -Clto`|1_581_104|1_568_992|-12_112
`-Copt-level=3 -Clto` (debuginfo stripped)|321_648|317_600|-4048
`-Copt-level=3 -Clto` (fully stripped)|276_736|272_712|-4024

When stripping binaries, but not doing LTO, this increases binary size by up to 72 bytes, but in all other cases this reduces binary size by up to 12kb without stripping or 4kb with stripping.

klensy:
Diff with `-Cprefer-dynamic` should be bigger perhaps (in % of size change).

bjorn3:
I think in that case only the size of libstd.so would change and not the actual executable.

dtolnay (commented):
This PR is labelled <kbd>T-libs-api</kbd> but it's not immediately clear to me why. Is there an externally observable API tradeoff that we'll need to reach a consensus on? Otherwise this seems like <kbd>T-libs</kbd>'s business.

(Haven't gotten a chance to review yet.)

bjorn3:
I meant libs-impl, not libs-api. My bad.

dtolnay (approved):
Looks good. :) Thank you for nicely splitting the commits.

Issue #88802, opened 2021-09-09
Missing help when trying to enable stabilized library feature on stable


jruderman:
Given the following [code](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=ecc8d91eb04ae1c75051db406627d0a1): 

```rust
#![feature(discriminant_value)]
fn main() {
    let y = Some(1);
    println!("{:?}", std::mem::discriminant(&y));
}
```

The current output is:

```
error[E0554]: `#![feature]` may not be used on the stable release channel
 --> src/main.rs:1:1
  |
1 | #![feature(discriminant_value)]
  | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

Ideally the output should look like:

```
error: ...
help: the feature `discriminant_value` has been stable since 1.21.0 and no longer requires an attribute to enable
```

(or it could be downgraded from error to warning)

#83722 fixed this for `#![feature(min_const_generics)]` but not for `#![feature(discriminant_value)]`, somehow

jyn514:
Hmm, this might be because it's a library feature and not a language feature?

jyn514:
Mentoring instructions: include library features in this check:
https://github.com/rust-lang/rust/blob/497ee321af3b8496eaccd7af7b437f18bab81abf/compiler/rustc_ast_passes/src/feature_gate.rs#L723

vishadGoyal:
Hi! I started working on this but didn't know I had to claim it first. My first time contributing to open source.
Anyway, so far I realized we need to combine these two Vectors instead of just using the declared_lang_features vector
[https://github.com/rust-lang/rust/blob/4e880f8cbc191374ce7db335962489f41d6d4e3e/compiler/rustc_feature/src/active.rs#L45](url)

[https://github.com/rust-lang/rust/blob/4e880f8cbc191374ce7db335962489f41d6d4e3e/compiler/rustc_ast_passes/src/feature_gate.rs#L709](url)

The remaining challenge is to combine two vectors with different data types.
Hope this helps @chansuke Best of luck!

chansuke:
@vishadGoyal I appreciate your great work. You should work on this :)

vishadGoyal:
Thanks a lot! I hope I didn't guilt you into doing this 😅 @chansuke

vishadGoyal:
Hi I'm having trouble with debugging my changes. 

[...]

PR #89012, opened 2021-09-16
Suggest removing `#![feature]` for library features that have been stabilized

vishadGoyal:
Issue: https://github.com/rust-lang/rust/issues/88802
Delayed the check if #![feature] has been used to enable lib features in a non-nightly build to occur after TyCtxt has been constructed.

### DIFF
--- a/compiler/rustc_ast_passes/src/feature_gate.rs
+++ b/compiler/rustc_ast_passes/src/feature_gate.rs
@@ -702,10 +702,16 @@ pub fn check_crate(krate: &ast::Crate, sess: &Session) {
 }
 
 fn maybe_stage_features(sess: &Session, krate: &ast::Crate) {
+    // checks if `#![feature]` has been used to enable any lang feature
+    // does not check the same for lib features unless there's at least one
+    // declared lang feature
     use rustc_errors::Applicability;
 
     if !sess.opts.unstable_features.is_nightly_build() {
         let lang_features = &sess.features_untracked().declared_lang_features;
+        if lang_features.len() == 0 {
+            return;
+        }
         for attr in krate.attrs.iter().filter(|attr| attr.has_name(sym::feature)) {
             let mut err = struct_span_err!(
                 sess.parse_sess.span_diagnostic,
--- a/compiler/rustc_passes/src/stability.rs
+++ b/compiler/rustc_passes/src/stability.rs
@@ -929,6 +929,16 @@ pub fn check_unused_or_stable_features(tcx: TyCtxt<'_>) {
     let declared_lib_features = &tcx.features().declared_lib_features;
     let mut remaining_lib_features = FxHashMap::default();
     for (feature, span) in declared_lib_features {
+        if !tcx.sess.opts.unstable_features.is_nightly_build() {
+            struct_span_err!(
+                tcx.sess,
+                *span,
+                E0554,
+                "`#![feature]` may not be used on the {} release channel",
+                env!("CFG_RELEASE_CHANNEL")
+            )
+            .emit();
+        }
         if remaining_lib_features.contains_key(&feature) {
             // Warn if the user enables a lib feature multiple times.
             duplicate_feature_err(tcx.sess, *span, *feature);


compiler/rustc_ast_passes/src/feature_gate.rs
```diff
@@ -701,11 +701,16 @@ pub fn check_crate(krate: &ast::Crate, sess: &Session) {
...
-fn maybe_stage_features(sess: &Session, krate: &ast::Crate) {
+fn maybe_stage_lang_features(sess: &Session, krate: &ast::Crate) {
+    // checks if `#![feature]` has been used to enable any lang feature
+    // does not check the same for lib features
     use rustc_errors::Applicability;
 
     if !sess.opts.unstable_features.is_nightly_build() {
         let lang_features = &sess.features_untracked().declared_lang_features;
+        if lang_features.len() == 0 {
+            return;
```
jyn514, line 712: What is this check doing? It should never be hit since there will always be some features built-in, but it also doesn't seem to have anything to do with the rest of the code.

compiler/rustc_passes/src/stability.rs
```diff
@@ -981,6 +984,12 @@ pub fn check_unused_or_stable_features(tcx: TyCtxt<'_>) {
     // don't lint about unused features. We should re-enable this one day!
 }
 
+fn enabling_feature_not_allowed(tcx: TyCtxt<'_>, span: Span) {
```
jyn514, line 987: I wouldn't add a function for this, you can just inline it manually.

jyn514 (changes requested):
Isn't the point to suggest removing the feature? This looks like you've just duplicated the check later into the compiler.

compiler/rustc_ast_passes/src/feature_gate.rs
vishadGoyal: The `lang_features` vector contains the `declared_lang_features` in the session. So only the ones which have been enabled show up here.
I've added this check to skip emitting an error if only lib_features have been enabled and no lang_features

jyn514: Ok. That still seems like the wrong check, though - if there's one language and one library feature, it should still suggest removing the stable library feature.

vishadGoyal:
> Isn't the point to suggest removing the feature? This looks like you've just duplicated the check later into the compiler.

The warnings for unnecessary use of `#![feature]` are already emitted in this check. After my changes, it also emits an error if appropriate.
Behaviour:
Screenshot from 2021-09-16 19-56-33

compiler/rustc_passes/src/stability.rs
```diff
@@ -981,6 +984,12 @@ pub fn check_unused_or_stable_features(tcx: TyCtxt<'_>) {
...
+    struct_span_err!(
+        tcx.sess, span, E0554, "`#![feature]` may not be used on the {} release channel", option_env!("CFG_RELEASE_CHANNEL").unwrap_or("(unknown)")
```
jyn514, line 989: Just use env! here, it's better than behaving strangely at runtime.

compiler/rustc_ast_passes/src/feature_gate.rs
vishadGoyal: How will that be possible since the suggestion can be generated only after TyCtxt is generated? And we will be throwing an error at the parsing stage if a lang feature has been enabled, the compiler will abort and won't go to the next stage

jyn514: > And we will be throwing an error at the parsing stage if a lang feature has been enabled, the compiler will abort and won't go to the next stage

Oh hmm, good point. Ok, this seems fine then - if the language feature wasn't stable removing the library feature wouldn't have helped anyway.

compiler/rustc_ast_passes/src/feature_gate.rs
```diff
@@ -701,11 +701,16 @@ pub fn check_crate(krate: &ast::Crate, sess: &Session) {
     visit::walk_crate(&mut visitor, krate);
 }
 
-fn maybe_stage_features(sess: &Session, krate: &ast::Crate) {
+fn maybe_stage_lang_features(sess: &Session, krate: &ast::Crate) {
+    // checks if `#![feature]` has been used to enable any lang feature
+    // does not check the same for lib features
```
jyn514, line 706: Hmm, this isn't always true though, right? It will still error on lib features if there's at least one lang feature.

vishadGoyal: but the check is only on lang_features. It will error only because of that one lang feature

jyn514: I don't think that's true. The check is `for attr in krate.attrs.iter().filter(|attr| attr.has_name(sym::feature)) {`, which doesn't distinguish the two.

Please write a test, it will be much easier than trying to figure it out from the code.

vishadGoyal:
> Ah! That's perfect then :) can you add that as a test case? It would go somewhere in src/test/ui - anywhere other than issues/ is probably fine.

@jyn514 Is there a way to specify which channel the test should be run on? I'm adding a sample code to reproduce this situation and adding the expected stderr output. However, this will be different for different channels (beta, nightly, etc.)

jyn514:
> @jyn514 Is there a way to specify which channel the test should be run on? I'm adding a sample code to reproduce this situation and adding the expected stderr output. However, this will be different for different channels (beta, nightly, etc.)

No, unfortunately, I forgot about that. If you do a manual test and post the output that's good enough.

vishadGoyal:
Manual test:
Screenshot from 2021-09-16 21-56-02

Is this good enough? @jyn514

jyn514:
@vishadGoyal how does it behave when there's at least one language feature (e.g. `#![feature(never_type_fallback)]`)?

vishadGoyal:
> @vishadGoyal how does it behave when there's both a library feature and at least one language feature (e.g. `#![feature(never_type_fallback)]`)?
Screenshot from 2021-09-16 22-20-42

I guess that's functional but not exactly what I had in mind.

compiler/rustc_ast_passes/src/feature_gate.rs
jyn514: Ok yup, https://github.com/rust-lang/rust/pull/89012#issuecomment-921068622 is what I expected. Please update the comment and then I think this is good to merge :)

jyn514: and I think it makes sense to revert changing the name of the function, too - it still checks lib features when there's only one library feature.

jyn514:
@bors r+ squash

Thanks!

Mark-Simulacrum:
@bors r=jyn514 squash-

Squashed the commits locally -- I don't think we've fixed that to behave well in bors just yet.

PR #89015, opened 2021-09-16
core::ascii::escape_default: reduce struct size

klensy:


Mark-Simulacrum:
Is there some concrete motivation for this? Generally speaking I don't think the trade off of the unsafe code is worth it here; the escape_default iterator is unlikely to be in a particularly hot path. Further, it's not inlined today (and can't be without -Zbuild-std or equivalent), since all of this code is not generic and so is codegen'd into std. (This is not to say it *should* be inlined; I think this is fine).

klensy:
Motivation was that i looked around std and checked if there some needless bound checks or similar things (plus looking for function usage in compiler or deps).
Display implementation already uses unsafe, but i can drop second commit, anyway.

Btw, can i replace things like this collect-via-iterator with to_string in this commit?
https://github.com/rust-lang/rust/blob/e0c38af27cb5f6f961809601b717d6afc3b190ee/compiler/rustc_ast/src/util/literal.rs#L170-L172

Mark-Simulacrum:
> Motivation was that i looked around std and checked if there some needless bound checks or similar things (plus looking for function usage in compiler or deps).

I don't feel like this is sufficient justification for introducing unsafe code here. While it is not too hard to justify soundness here, it doesn't seem warranted to me, since it makes the code harder to read and edit and for little benefit. Bounds checking is not that expensive.

Making the struct smaller also seems unlikely to be of much help, though I'm not really opposed. Practically speaking it'll almost always be put on the stack, and there it doesn't really matter that much whether we are using 24 or 6 bytes -- it's a bump allocation anyway. But I'm OK r+'ing the first commit here.

> Btw, can i replace things like this collect-via-iterator with to_string in this PR?

This PR does not seem like the right place for changes to the compiler, no. Not sure I follow your question entirely, but that feels like a change best left for separate PR(s).

library/core/src/ascii.rs
```diff
@@ -138,7 +138,7 @@ impl FusedIterator for EscapeDefault {}
 impl fmt::Display for EscapeDefault {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
         // SAFETY: ok because `escape_default` created only valid utf-8 data
-        f.write_str(unsafe { from_utf8_unchecked(&self.data[self.range.clone()]) })
+        f.write_str(unsafe { from_utf8_unchecked(&self.data[..self.range.end as usize]) })
```
Mark-Simulacrum, line 141: This is a change in behavior if the iterator has already been advanced from the front, right?

klensy: Nope, when dropped second commit noticed, that i changed this in first.

PR #89016, opened 2021-09-16
fix non_blanket_impls iteration order

lcnr:
We sometimes iterate over all `non_blanket_impls`, not sure if this is observable outside
of error messages (i.e. as incremental bugs). This should fix the underlying issue of #86986.

second attempt of #88718

r? @nikomatsakis

### DIFF
--- a/compiler/rustc_metadata/src/dependency_format.rs
+++ b/compiler/rustc_metadata/src/dependency_format.rs
@@ -277,7 +277,7 @@ fn attempt_static(tcx: TyCtxt<'_>) -> Option<DependencyList> {
     let all_crates_available_as_rlib = tcx
         .crates(())
         .iter()
-        .cloned()
+        .copied()
         .filter_map(|cnum| {
             if tcx.dep_kind(cnum).macros_only() {
                 return None;
@@ -291,10 +291,11 @@ fn attempt_static(tcx: TyCtxt<'_>) -> Option<DependencyList> {
 
     // All crates are available in an rlib format, so we're just going to link
     // everything in explicitly so long as it's actually required.
-    let last_crate = tcx.crates(()).len();
-    let mut ret = (1..last_crate + 1)
-        .map(|cnum| {
-            if tcx.dep_kind(CrateNum::new(cnum)) == CrateDepKind::Explicit {
+    let mut ret = tcx
+        .crates(())
+        .iter()
+        .map(|&cnum| {
+            if tcx.dep_kind(cnum) == CrateDepKind::Explicit {
                 Linkage::Static
             } else {
                 Linkage::NotLinked
--- a/compiler/rustc_metadata/src/rmeta/decoder/cstore_impl.rs
+++ b/compiler/rustc_metadata/src/rmeta/decoder/cstore_impl.rs
@@ -304,17 +304,7 @@ pub fn provide(providers: &mut Providers) {
             // traversal, but not globally minimal across all crates.
             let bfs_queue = &mut VecDeque::new();
 
-            // Preferring shortest paths alone does not guarantee a
-            // deterministic result; so sort by crate num to avoid
-            // hashtable iteration non-determinism. This only makes
-            // things as deterministic as crate-nums assignment is,
-            // which is to say, its not deterministic in general. But
-            // we believe that libstd is consistently assigned crate
-            // num 1, so it should be enough to resolve #46112.
-            let mut crates: Vec<CrateNum> = (*tcx.crates(())).to_owned();
-            crates.sort();
-
-            for &cnum in crates.iter() {
+            for &cnum in tcx.crates(()) {
                 // Ignore crates without a corresponding local `extern crate` item.
                 if tcx.missing_extern_crate_item(cnum) {
                     continue;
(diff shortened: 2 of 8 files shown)


compiler/rustc_middle/src/ty/trait_def.rs
```diff
@@ -66,11 +64,11 @@ pub enum TraitSpecializationKind {
     AlwaysApplicable,
 }
 
-#[derive(Default, Debug)]
+#[derive(Default, Debug, HashStable)]
 pub struct TraitImpls {
     blanket_impls: Vec<DefId>,
     /// Impls indexed by their simplified self type, for fast lookup.
-    non_blanket_impls: FxHashMap<fast_reject::SimplifiedType, Vec<DefId>>,
+    non_blanket_impls: FxIndexMap<fast_reject::SimplifiedType, Vec<DefId>>,
```
Mark-Simulacrum, line 71: Hm, is this enough?

It looks like insertion into non_blanket_impls happens in a loop which is a result of `tcx.crates(())` which isn't stable across compilation sessions I'd expect (though I could be wrong).

lcnr: The way `crates(())` is computed, it looks like it is always `(1..=num_crates).map(CrateNum)` (0 is reserved for the current crate and does not have any data) and it I would assume the crates get added to the `CStore` in a deterministic order.

We do have `stable_crate_id` which orders crates independently of their `cnum`. I could be that `CrateNum`s are not deterministically assigned to their actual crate data, though it seems quite buggy to me if this doesn't invalidate query caches in that case. e.g. `trait_impls_of_provider` just uses the order of `crates(())` which influences the way winnowing works for example.

A bit out of my depth here, maybe @rust-lang/wg-incr-comp can help and clarify here :sweat_smile: :heart:

compiler/rustc_metadata/src/rmeta/decoder/cstore_impl.rs
```diff
@@ -304,17 +304,7 @@ pub fn provide(providers: &mut Providers) {
...
-            // hashtable iteration non-determinism. This only makes
-            // things as deterministic as crate-nums assignment is,
-            // which is to say, its not deterministic in general. But
-            // we believe that libstd is consistently assigned crate
-            // num 1, so it should be enough to resolve #46112.
-            let mut crates: Vec<CrateNum> = (*tcx.crates(())).to_owned();
-            crates.sort();
-
-            for &cnum in crates.iter() {
+            for &cnum in tcx.crates(()) {
```
nikomatsakis, line 307: are these changes just "cleanup" or is there some substantive change in the order in which things are done here?

lcnr: just cleanup, the only change of behavior is the second commit here.

`crates(())` already returns the `cnums` in order, so the code i deleted in the other commits should just be noops

compiler/rustc_middle/src/ty/trait_def.rs
Mark-Simulacrum: Yeah, we probably do end up with an ordering that is mostly stable from run to run (presuming no shifts in code or other sources), and I think incremental will track the query fine. My guess is that we could avoid marking the trait query as red if the crate load order changes (e.g., because the first code using that crate is moved?) if this used a stable crate ID, but it may not be worth it. Ultimately this PR does seem like it makes things strictly better here :)

nikomatsakis: I *think* that crate ordering is indeed deterministic, since it isn't something that we discover dynamically? But indeed I bet @michaelwoerister or @petrochenkov would be certain.

nikomatsakis:
r=me, but I'm going to make this 

r? @michaelwoerister 

since the primary question seems to have to do with incremental.

compiler/rustc_middle/src/ty/trait_def.rs
cjgillot: Crate are assigned their CrateNum in the order in which they are required by the resolver. As the resolver is not incremental, this order is deterministic but not stable. On the other hand, it is not obvious at all.
Incremental computes a hash for `crates(())` based on their order in the vector and their `stable_crate_id`. We could make it independent on the order of crates by having the `crates` query sort its result by increasing `stable_crate_id`.

compiler/rustc_metadata/src/dependency_format.rs
```diff
@@ -291,10 +291,9 @@ fn attempt_static(tcx: TyCtxt<'_>) -> Option<DependencyList> {
 
     // All crates are available in an rlib format, so we're just going to link
     // everything in explicitly so long as it's actually required.
-    let last_crate = tcx.crates(()).len();
-    let mut ret = (1..last_crate + 1)
-        .map(|cnum| {
-            if tcx.dep_kind(CrateNum::new(cnum)) == CrateDepKind::Explicit {
+    let mut ret = tcx.crates(()).iter()
+        .map(|&cnum| {
```
cjgillot, line 295: Do we need to skip the leading `CrateNum(0)`?

lcnr: :shrug: `CrateNum(0)` is the local crate so I don't think we need to do any linking for that.

I think that explicitly dealing with the local crate and only iterating over all external ones is probably better than specialcasing the first item of the `crates` iter. But that's a fairly uninformed opinion

michaelwoerister:
I'll take a look tomorrow.

compiler/rustc_metadata/src/dependency_format.rs
michaelwoerister: I don't understand how that code works in the first place? `last_crate` is `tcx.crates(()).len()` and then it creates a `CrateNum(last_crate)`? That should not be a valid `CrateNum`, right?

michaelwoerister: Oh, `tcx.crates(())` only returns _upstream_ `CrateNum`s. That is far from obvious. I think the query should be renamed to `upstream_crate_nums` or something. The same goes for `CStore::crates_untracked`. Would you be up for doing that, @lcnr? (but maybe in a separate PR)

compiler/rustc_middle/src/ty/trait_def.rs
michaelwoerister: I think that parts of the code might rely on `tcx.crates(())` being sorted by `CrateNum`, for better or worse. I would not recommend trying to change that as part of this PR.

compiler/rustc_middle/src/ich/hcx.rs
```diff
@@ -241,39 +238,3 @@ impl<'a> rustc_span::HashStableContext for StableHashingContext<'a> {
...
-    {
-        let mut blanket_impls: SmallVec<[_; 8]> =
-            blanket_impls.iter().map(|&def_id| hcx.def_path_hash(def_id)).collect();
-
-        if blanket_impls.len() > 1 {
-            blanket_impls.sort_unstable();
-        }
-
-        blanket_impls.hash_stable(hcx, hasher);
-    }
```
michaelwoerister, line 260: This actually looks like it is unsound, right @Aaron1011 & @cjgillot? If the order is observable, we must not "stabilize" it via sorting.

michaelwoerister:
This looks good to me. If anything it makes the code more correct (see https://github.com/rust-lang/rust/pull/89016#discussion_r714701327).

I would also recommend renaming the `crates` query to `upstream_crate_nums` (since it does not include `LOCAL_CRATE`) and document that its result is expected to be sorted by the numerical `CrateNum` value.

compiler/rustc_metadata/src/dependency_format.rs
lcnr: `extern_crates` maybe? :thinking: though i prefer doing that in a separate PR, want to merge this as it is blocking some further work

michaelwoerister: I've personally always used the term "upstream" (e.g. see https://github.com/rust-lang/rust/blob/15d9ba0133ce0b35348e1c8367afe00aec841ffa/compiler/rustc_middle/src/query/mod.rs#L1264). "Extern" often refers to something non-Rust. Although `extern crate foo` suggests otherwise `:)`. 

Anyway, I'm fine with doing this in another PR. You can r=me on this one.

lcnr: `upstream` is also fine :laughing: was thinking of `extern crate something` ^^

rylev:
The improvements outweigh the regressions with the only regressions being in historically noisy benchmarks. As such I'm going to mark this as triaged.

PR #83093, opened 2021-03-13
where available use AtomicU{64,128} instead of mutex for Instant backsliding protection

the8472:
This decreases the overhead of backsliding protection on x86 systems with unreliable TSC, e.g. windows. And on aarch64 systems where 128bit atomics are available.

The following benchmarks were taken on x86_64 linux though by overriding `actually_monotonic()`, the numbers may look different on other platforms

```
# actually_monotonic() == true
test time::tests::instant_contention_01_threads                   ... bench:          44 ns/iter (+/- 0)
test time::tests::instant_contention_02_threads                   ... bench:          44 ns/iter (+/- 0)
test time::tests::instant_contention_04_threads                   ... bench:          44 ns/iter (+/- 0)
test time::tests::instant_contention_08_threads                   ... bench:          44 ns/iter (+/- 0)
test time::tests::instant_contention_16_threads                   ... bench:          44 ns/iter (+/- 0)

# 1x AtomicU64
test time::tests::instant_contention_01_threads                   ... bench:          65 ns/iter (+/- 0)
test time::tests::instant_contention_02_threads                   ... bench:         157 ns/iter (+/- 20)
test time::tests::instant_contention_04_threads                   ... bench:         281 ns/iter (+/- 53)
test time::tests::instant_contention_08_threads                   ... bench:         555 ns/iter (+/- 77)
test time::tests::instant_contention_16_threads                   ... bench:         883 ns/iter (+/- 107)

# mutex
test time::tests::instant_contention_01_threads                   ... bench:          60 ns/iter (+/- 2)
test time::tests::instant_contention_02_threads                   ... bench:         770 ns/iter (+/- 231)
test time::tests::instant_contention_04_threads                   ... bench:       1,347 ns/iter (+/- 45)
test time::tests::instant_contention_08_threads                   ... bench:       2,693 ns/iter (+/- 114)
test time::tests::instant_contention_16_threads                   ... bench:       5,244 ns/iter (+/- 487)
```

Since I don't have an arm machine with 128bit atomics I wasn't able to benchmark the AtomicU128 implementation.

PR #89017, opened 2021-09-16
fix potential race in AtomicU64 time monotonizer

the8472:
The AtomicU64-based monotonizer introduced in #83093 is incorrect because several threads could try to update the value concurrently and a thread which doesn't have the newest value among all the updates could win.

That bug probably has little real world impact since it doesn't make observed time worse than hardware clocks. The worst case would probably be a thread which has a clock that is behind by several cycles observing several inconsistent fixups, which should be similar to observing the unfiltered backslide in the first place.

New benchmarks, they don't look as good as the original PR but still an improvement compared to the mutex.
I don't know why the contended mutex case is faster now than in the previous benchmarks.

```
actually_monotonic() == true:
test time::tests::instant_contention_01_threads                   ... bench:          44 ns/iter (+/- 0)
test time::tests::instant_contention_02_threads                   ... bench:          45 ns/iter (+/- 0)
test time::tests::instant_contention_04_threads                   ... bench:          45 ns/iter (+/- 0)
test time::tests::instant_contention_08_threads                   ... bench:          45 ns/iter (+/- 0)
test time::tests::instant_contention_16_threads                   ... bench:          46 ns/iter (+/- 0)


atomic u64:
test time::tests::instant_contention_01_threads                   ... bench:          66 ns/iter (+/- 0)
test time::tests::instant_contention_02_threads                   ... bench:         287 ns/iter (+/- 14)
test time::tests::instant_contention_04_threads                   ... bench:         296 ns/iter (+/- 43)
test time::tests::instant_contention_08_threads                   ... bench:         604 ns/iter (+/- 163)
test time::tests::instant_contention_16_threads                   ... bench:       1,147 ns/iter (+/- 29)

mutex:
test time::tests::instant_contention_01_threads                   ... bench:          78 ns/iter (+/- 0)
test time::tests::instant_contention_02_threads                   ... bench:         652 ns/iter (+/- 275)
test time::tests::instant_contention_04_threads                   ... bench:         900 ns/iter (+/- 32)
test time::tests::instant_contention_08_threads                   ... bench:       1,927 ns/iter (+/- 62)
test time::tests::instant_contention_16_threads                   ... bench:       3,748 ns/iter (+/- 146)
```

### DIFF
--- a/library/std/src/time/monotonic.rs
+++ b/library/std/src/time/monotonic.rs
@@ -37,20 +37,15 @@ pub mod inner {
         // This could be a problem for programs that call instants at intervals greater
         // than 68 years. Interstellar probes may want to ensure that actually_monotonic() is true.
         let packed = (secs << 32) | nanos;
-        let mut old = mono.load(Relaxed);
-        loop {
-            if old == UNINITIALIZED || packed.wrapping_sub(old) < u64::MAX / 2 {
-                match mono.compare_exchange_weak(old, packed, Relaxed, Relaxed) {
-                    Ok(_) => return raw,
-                    Err(x) => {
-                        old = x;
-                        continue;
-                    }
-                }
-            } else {
+        let updated = mono.fetch_update(Relaxed, Relaxed, |old| {
+            (old == UNINITIALIZED || packed.wrapping_sub(old) < u64::MAX / 2).then_some(packed)
+        });
+        match updated {
+            Ok(_) => raw,
+            Err(newer) => {
                 // Backslide occurred. We reconstruct monotonized time from the upper 32 bit of the
                 // passed in value and the 64bits loaded from the atomic
-                let seconds_lower = old >> 32;
+                let seconds_lower = newer >> 32;
                 let mut seconds_upper = secs & 0xffff_ffff_0000_0000;
                 if secs & 0xffff_ffff > seconds_lower {
                     // Backslide caused the lower 32bit of the seconds part to wrap.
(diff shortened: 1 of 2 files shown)


library/std/src/time/monotonic.rs
```diff
@@ -37,35 +37,41 @@ pub mod inner {
...
+            if old == UNINITIALIZED || packed.wrapping_sub(old) < u64::MAX / 2 {
+                match mono.compare_exchange_weak(old, packed, Relaxed, Relaxed) {
+                    Ok(_) => return raw,
+                    Err(x) => {
+                        old = x;
+                        continue;
+                    }
+                }
+            } else {
+                // Backslide occurred. We reconstruct monotonized time from the upper 32 bit of the
```
AGSaidi, line 51: i feel like i'm missing something here. Shouldn't mono be updated in the else block?

the8472: In the else block the thread-local time is older than what is stored in the atomic. so the atomic is the most recent time and needs no updating.

library/std/src/time/monotonic.rs
```diff
@@ -37,35 +37,41 @@ pub mod inner {
...
-                // most recent observed time: 0xA1_0000_0000_0000_0000u128
-                // bits stored in AtomicU64:     0x0000_0000_0000_0000u64
-                // backslide by 1s
-                // caller time is             0xA0_ffff_ffff_0000_0000u128
-                // -> we can fix up the upper half time by adding 1 << 32
-                seconds_upper = seconds_upper.wrapping_add(0x1_0000_0000);
+        let mut old = mono.load(Relaxed);
+        loop {
+            if old == UNINITIALIZED || packed.wrapping_sub(old) < u64::MAX / 2 {
+                match mono.compare_exchange_weak(old, packed, Relaxed, Relaxed) {
```
kennytm, line 43: can this loop be written using `fetch_update` instead?

```rust
let updated = mono.fetch_update(Relaxed, Relaxed, |old| {
    (old == UNINITIALIZED || packed.wrapping_sub(old) < u64::MAX / 2).then_some(packed)
});
if updated.is_ok() {
    raw
} else {
    // ...<the entire else branch>...
    ZERO.checked_add_duration(&Duration::new(secs, nanos)).unwrap()
}
```

the8472: Yeah that works. Applied.

PR #89021, opened 2021-09-16
Add a separate error for `dyn Trait` in `const fn`

WaffleLapkin:
Previously "trait bounds other than `Sized` on const fn parameters are unstable" error was used for both trait bounds (`<T: Trait>`) and trait objects (`dyn Trait`). This was pretty confusing.

This PR adds a separate error for trait objects: "trait objects in const fn are unstable". The error for trait bounds is otherwise intact.

This is follow up to #88907

r? @estebank 

@rustbot label: +A-diagnostics

### DIFF
--- a/compiler/rustc_const_eval/src/transform/check_consts/check.rs
+++ b/compiler/rustc_const_eval/src/transform/check_consts/check.rs
@@ -384,11 +384,11 @@ impl Checker<'mir, 'tcx> {
                         match pred.skip_binder() {
                             ty::ExistentialPredicate::AutoTrait(_)
                             | ty::ExistentialPredicate::Projection(_) => {
-                                self.check_op(ops::ty::TraitBound(kind))
+                                self.check_op(ops::ty::DynTrait(kind))
                             }
                             ty::ExistentialPredicate::Trait(trait_ref) => {
                                 if Some(trait_ref.def_id) != self.tcx.lang_items().sized_trait() {
-                                    self.check_op(ops::ty::TraitBound(kind))
+                                    self.check_op(ops::ty::DynTrait(kind))
                                 }
                             }
                         }
--- a/src/test/ui/consts/min_const_fn/min_const_fn_dyn.rs
+++ b/src/test/ui/consts/min_const_fn/min_const_fn_dyn.rs
@@ -7,9 +7,9 @@ struct Hide(HasDyn);
 const fn no_inner_dyn_trait(_x: Hide) {}
 const fn no_inner_dyn_trait2(x: Hide) {
     x.0.field;
-//~^ ERROR trait bounds other than `Sized`
+//~^ ERROR trait objects in const fn are unstable
 }
 const fn no_inner_dyn_trait_ret() -> Hide { Hide(HasDyn { field: &0 }) }
-//~^ ERROR trait bounds other than `Sized`
+//~^ ERROR trait objects in const fn are unstable
 
 fn main() {}
--- a/compiler/rustc_const_eval/src/transform/check_consts/ops.rs
+++ b/compiler/rustc_const_eval/src/transform/check_consts/ops.rs
@@ -599,7 +599,7 @@ pub mod ty {
         }
 
         fn build_error(&self, ccx: &ConstCx<'_, 'tcx>, span: Span) -> DiagnosticBuilder<'tcx> {
-            let mut builder = feature_err(
+            let mut err = feature_err(
                 &ccx.tcx.sess.parse_sess,
                 sym::const_fn_trait_bound,
                 span,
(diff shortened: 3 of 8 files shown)


compiler/rustc_const_eval/src/transform/check_consts/check.rs
```diff
@@ -384,11 +384,11 @@ impl Checker<'mir, 'tcx> {
                         match pred.skip_binder() {
                             ty::ExistentialPredicate::AutoTrait(_)
                             | ty::ExistentialPredicate::Projection(_) => {
-                                self.check_op(ops::ty::TraitBound(kind))
+                                self.check_op(ops::ty::DynTrait(kind))
                             }
                             ty::ExistentialPredicate::Trait(trait_ref) => {
                                 if Some(trait_ref.def_id) != self.tcx.lang_items().sized_trait() {
```
WaffleLapkin, line 390: I just noticed: isn't this check redundant? `Sized` trait can't be made into an object anyway.

src/test/ui/consts/min_const_fn/min_const_fn_dyn.rs
```diff
@@ -7,9 +7,9 @@ struct Hide(HasDyn);
 const fn no_inner_dyn_trait(_x: Hide) {}
 const fn no_inner_dyn_trait2(x: Hide) {
     x.0.field;
-//~^ ERROR trait bounds other than `Sized`
+//~^ ERROR trait objects in const fn are unstable
```
estebank, line 10: For the cases of fields like this one it would be nice to have a `span_note` pointing at the type and field's type to show how it is a trait object.

compiler/rustc_const_eval/src/transform/check_consts/check.rs
estebank: They can't be made into an object (yet, at least), but user's can write all kinds of broken code. Checks like these are "defense in depth" by silencing errors that would otherwise be redundant. Now, in those cases it is good style to leave a comment stating this, and I can't know _for sure_ that that's what happened here and @ecstatic-morse['s code is a year old](https://github.com/rust-lang/rust/commit/3569bb63231979c89b71736cd8b9399aad249784#diff-430f3a60a526044dff4ee953a9f6f25691db98573ecd05c67642488f24a0c153R312) so it's unreasonable to expect them to remember 😊

Let's keep it as is.

estebank (approved):

compiler/rustc_const_eval/src/transform/check_consts/check.rs
WaffleLapkin: Ah, so this protects users from seeing both the "`Sized` cannot be made into an object" _and_  "trait objects in const fn are unstable" at the same time? That's nice! I thought that was just copy-pasted from the [bounds check](https://github.com/rust-lang/rust/commit/3569bb63231979c89b71736cd8b9399aad249784#diff-430f3a60a526044dff4ee953a9f6f25691db98573ecd05c67642488f24a0c153R349) 😄

estebank: That might *also* be the case. Regardless of what the _intent_ was, that's the effect, and I want us to keep it 😅

src/test/ui/consts/min_const_fn/min_const_fn_dyn.rs
WaffleLapkin: It would be really nice! Can we implement this though? It seems like we only check locals and not expressions: https://github.com/rust-lang/rust/blob/f84000d9bc93f7c09243144d1a729448da3d714d/compiler/rustc_const_eval/src/transform/check_consts/check.rs#L245-L259

estebank: I'm not sure if we can't get that info from `local.local_info` or `local.source_info` (I'm not _that_ familiar with const eval and have to relearn it every time I look at it).

WaffleLapkin: Hm, after reading docs I can't see how we can get the expression, but maybe I'm missing something

estebank: Can you check what the contents of `local.local_info` are? I assume it is going to be a `VarBindingForm`. If it is, what are the contents of `opt_match_place`?

WaffleLapkin: So in the case of our interest (field access) `local.local_info` is simply `None` :(

estebank: In that case, we'll have to make more dramatic changes so that _we can_ track that data :-/

It's probably not high priority, but if you desire to take that on, I'd be more than happy to see it through!

WaffleLapkin: I don't think I'm interested enough in that. Thanks for offering help though!

PR #89024, opened 2021-09-16
Lazy TAIT preparation cleanups

oli-obk:
Check that TAIT generics are fully generic in mir typeck instead of wf-check, as wf-check can by definition only check TAIT in return position and not account for TAITs defined in the body of the function

r? @spastorino @nikomatsakis

### DIFF
--- a/compiler/rustc_typeck/src/collect/type_of.rs
+++ b/compiler/rustc_typeck/src/collect/type_of.rs
@@ -1,4 +1,3 @@
-use rustc_data_structures::fx::FxHashSet;
 use rustc_errors::{Applicability, ErrorReported, StashKey};
 use rustc_hir as hir;
 use rustc_hir::def::{DefKind, Res};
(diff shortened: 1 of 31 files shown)


compiler/rustc_typeck/src/collect/type_of.rs
```diff
@@ -584,50 +602,8 @@ fn find_opaque_ty_constraints(tcx: TyCtxt<'_>, def_id: LocalDefId) -> Ty<'_> {
                 // FIXME(oli-obk): trace the actual span from inference to improve errors.
                 let span = self.tcx.def_span(def_id);
 
-                // HACK(eddyb) this check shouldn't be needed, as `wfcheck`
```
oli-obk, line 587: Since the `mir_borrowck` call above will invoke the checking logic, duplicating it here with delay_span_bug adds no additional safety net

nikomatsakis (approved):

nikomatsakis:
(lgtm)

Issue #58713, opened 2019-02-24
Tracking issue for RFC 2627: #[link(kind="raw-dylib")]


joshtriplett:
This is the tracking issue for [RFC 2627](rust-lang/rfcs#2627), `#[link(kind="raw-dylib")]`.

I will be spearheading the implementation work of this RFC, but would appreciate mentorship as I have relatively little experience working in rustc.

Opens:

- [ ] Implementation for Windows (top priority)
  - [ ] Emitting idata sections
  - [ ] New `link_ordinal` attribute
  - [ ] Testing in winapi with nightly Rust before stabilization
- [ ] Tests, including all the corner cases mentioned in the reference-level explanation
- [ ] Documentation
- [ ] Implementation for Linux and other platforms
- [ ] Implementation of a pure Rust target for Windows (no libc, no msvc, no mingw). This may require another RFC

lzutao:
The RFC link leads to 404 page now.

crlf0710:
cc https://github.com/rust-lang/rfcs/issues/1061

ZerothLaw:
I don't think I have the expertise to implement this feature, but I could write some (currently failing) tests for use by the implementor? 

(Background, Ran into this exact problem linking mscorlib.dll into executables. Had to write a C++ level wrapper to get it to work.)

crlf0710:
A [very first PR](https://github.com/rust-lang/rust/pull/63948) has landed, now the `link_ordinal` attribute is added. Want to say thank you to @Centril for helping me on this~

retep998:
There's now a bounty for this issue to incentivize someone to work on implementing this feature.

https://www.bountysource.com/issues/70533351-tracking-issue-for-rfc-2627-link-kind-raw-dylib

est31:
Some information I've gathered:

* The [spec](https://docs.microsoft.com/en-us/windows/win32/debug/pe-format#the-idata-section) can be found here. It's a bit hard to understand in isolation as it's incomplete and some numbers are given in bits and others in bytes. You need to figure this out from other resources (see below).
[...]

PR #84171, opened 2021-04-13
Partial support for raw-dylib linkage

ricobbe:
First cut of functionality for issue #58713: add support for `#[link(kind = "raw-dylib")]` on `extern` blocks in lib crates compiled to .rlib files.  Does not yet support `#[link_name]` attributes on functions, or the `#[link_ordinal]` attribute, or `#[link(kind = "raw-dylib")]` on `extern` blocks in bin crates; I intend to publish subsequent PRs to fill those gaps.  It's also not yet clear whether this works for functions in `extern "stdcall"` blocks; I also intend to investigate that shortly and make any necessary changes as a follow-on PR.

This implementation calls out to an LLVM function to construct the actual `.idata` sections as temporary `.lib` files on disk and then links those into the generated .rlib.

PR #89025, opened 2021-09-16
Implement `#[link_ordinal(n)]`

ricobbe:
Allows the use of `#[link_ordinal(n)]` with `#[link(kind = "raw-dylib")]`, allowing Rust to link against DLLs that export symbols by ordinal rather than by name.  As long as the ordinal matches, the name of the function in Rust is not required to match the name of the corresponding function in the exporting DLL.

Part of #58713.

### DIFF
--- a/compiler/rustc_codegen_llvm/src/llvm/ffi.rs
+++ b/compiler/rustc_codegen_llvm/src/llvm/ffi.rs
@@ -34,11 +34,18 @@ pub enum LLVMRustResult {
 #[repr(C)]
 pub struct LLVMRustCOFFShortExport {
     pub name: *const c_char,
+    pub ordinal_present: bool,
+    // value of `ordinal` only important when `ordinal_present` is true
+    pub ordinal: u16,
 }
 
 impl LLVMRustCOFFShortExport {
-    pub fn from_name(name: *const c_char) -> LLVMRustCOFFShortExport {
-        LLVMRustCOFFShortExport { name }
+    pub fn new(name: *const c_char, ordinal: Option<u16>) -> LLVMRustCOFFShortExport {
+        LLVMRustCOFFShortExport {
+            name,
+            ordinal_present: ordinal.is_some(),
+            ordinal: ordinal.unwrap_or(0),
+        }
     }
 }
 
--- a/compiler/rustc_metadata/src/native_libs.rs
+++ b/compiler/rustc_metadata/src/native_libs.rs
@@ -433,6 +433,12 @@ impl Collector<'tcx> {
                 }
             }
         };
-        DllImport { name: item.ident.name, ordinal: None, calling_convention, span: item.span }
+
+        DllImport {
+            name: item.ident.name,
+            ordinal: self.tcx.codegen_fn_attrs(item.id.def_id).link_ordinal,
+            calling_convention,
+            span: item.span,
+        }
     }
 }
--- a/compiler/rustc_codegen_llvm/src/back/archive.rs
+++ b/compiler/rustc_codegen_llvm/src/back/archive.rs
@@ -163,13 +163,13 @@ impl<'a> ArchiveBuilder<'a> for LlvmArchiveBuilder<'a> {
         // All import names are Rust identifiers and therefore cannot contain \0 characters.
         // FIXME: when support for #[link_name] implemented, ensure that import.name values don't
         // have any \0 characters
-        let import_name_vector: Vec<CString> = dll_imports
+        let import_name_and_ordinal_vector: Vec<(CString, Option<u16>)> = dll_imports
             .iter()
             .map(|import: &DllImport| {
                 if self.config.sess.target.arch == "x86" {
-                    LlvmArchiveBuilder::i686_decorated_name(import)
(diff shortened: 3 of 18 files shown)


compiler/rustc_metadata/src/native_libs.rs
```diff
@@ -433,6 +434,52 @@ impl Collector<'tcx> {
...
+        let mut result: Option<u16> = None;
+
+        for attr in self
+            .tcx
+            .hir()
+            .attrs(foreign_function.id.hir_id())
+            .iter()
+            .filter(|a| a.has_name(sym::link_ordinal))
+        {
+            let items = match attr.meta_item_list() {
```
est31, line 456: This can be added to the top in the form of a filter_map invocation. `filter_map(|attr| attr.meta_item_list())`. Doesn't have to though, if attr is needed in the loop's body e.g. to emit a misformatting error on its span.

compiler/rustc_metadata/src/native_libs.rs
```diff
@@ -433,6 +434,52 @@ impl Collector<'tcx> {
...
+                Some(items) => items,
+                None => {
+                    continue;
+                }
+            };
+
+            if result.is_some() {
+                // FIXME: warn or error on multiple #[link_ordinal] attributes on a single function?
```
est31, line 464: Personally I think an error is best. It can always be changed to a warning in the future if a use case is discovered but the opposite is not possible.

compiler/rustc_metadata/src/native_libs.rs
```diff
@@ -433,6 +434,52 @@ impl Collector<'tcx> {
...
+                self.tcx.sess.span_warn(
+                    foreign_function.span,
+                    "multiple `#[link_ordinal]` attributes on a single function",
+                );
+                return result;
+            }
+
+            // FIXME: do we need to check that there's only one item here?
```
est31, line 472: I'd recommend that, yes. You can directly match on `items.as_slice()` I think, no need for a loop.

ricobbe:
Question for reviewers: in attempting to clean up the error checking for `link_ordinal` in rustc_typeck/src/collect.rs beginning at about line 3156, I discovered that it's recording the ordinal, if present, to the function's `CodegenFnAttrs` struct.  Should I be getting this information from there, rather than re-analyzing the attributes, in native_libs.rs?  If so, how do I get a reference to the relevant `CodegenFnAttrs` object?  Or should we drop this field from `CodegenFnAttrs` instead?  As far as I can determine, it's only ever read by `check_link_name_xor_ordinal` in rustc_typeck/src/collect.rs.

(I refer to cleaning up the error checking because the messages rustc issues in certain circumstances are, at best, misleading: `#[link_ordinal(3, 4)]` is indeed invalid input, but complaining about an illegal argument format (unsuffixed integer value expected) in response is confusing.)

est31:
@ricobbe I'm just giving my personal views here, am no reviewer in any official capacity, but I think that duplication should be avoided and ideally it should be re-used. I think it should be possible to obtain the attrs with something like `self.tcx.codegen_fn_attrs(item.id.def_id)`, not entirely sure though.

Also, IMO the error message doesn't have to be perfect, but of course improvements are always nice.

compiler/rustc_metadata/src/native_libs.rs
ricobbe: obsolete due to refactoring

compiler/rustc_metadata/src/native_libs.rs
ricobbe: done, but added to rustc_typeck/src/collect.rs.

ricobbe:
@est31 Ok, I've pushed changes that use the info from `CodegenFnAttrs`.  As another benefit, this also streamlines the error-detection code as well!

est31 (approved):

compiler/rustc_codegen_llvm/src/llvm/ffi.rs
```diff
@@ -34,11 +34,18 @@ pub enum LLVMRustResult {
 #[repr(C)]
 pub struct LLVMRustCOFFShortExport {
     pub name: *const c_char,
+    pub ordinal_present: bool,
+    // value of `ordinal` only important when `ordinal_present` is true
+    pub ordinal: u16,
```
joshtriplett, line 39: This seems like it should be an `Option`.

joshtriplett:
LGTM from a purely language perspective. Deferring to compiler folks on the implementation strategy.

ricobbe:
> LGTM from a purely language perspective. Deferring to compiler folks on the implementation strategy.

@joshtriplett Thanks for the review!  I'm happy for the compiler folks to take a look at it, though I will say that this PR is a straightforward extension of the basic import-by-name raw-dylib work from PR #84171.

joshtriplett:
r? @michaelwoerister
for compiler team review of the implementation. No issues from a language semantics perspective.

michaelwoerister:
@rust-lang/infra, any idea what [that](https://github.com/rust-lang/rust/pull/89025#issuecomment-937740646) is about?

PR #89026, opened 2021-09-16
outline heavyweight closure that caused regressions in #86898

the8472:
Testing the suggestion in https://github.com/rust-lang/rust/pull/86898#issuecomment-920138051


nagisa:
The nesting is getting quite nasty there, but also waaaaat o.O

@bors r+

the8472:
@bors r-

#89031 shows similar results, is more general and has been approved too.

Issue #43596, opened 2017-08-01
oom=panic knob


Gankro:
Several users who are invested in the "thread is a task" model that Rust was originally designed around have expressed a desire to unwind in the case of OOM. Specifically each task has large and unpredictable memory usage, and tearing down the first task that encounters OOM is considered a reasonable strategy.

Aborting on OOM is considered unacceptable because it would be expensive to discard all the work that the other tasks have managed to do.

We already (accidentally?) specified that all allocators can panic on OOM, including the global one, with the introduction of generic allocators on collections. So in theory no one "should" be relying on the default `oom=abort` semantics. This knob would only affect the default global allocator. Presumably there would just be a function somewhere in the stdlib which is cfg'd to be a panic or abort based on this flag?

This is not a replacement for proper fallible allocation handling routines requested in #29802 because most of the potential users of **that** API build with `panic=abort`. Also the requesters of **this** API are unwilling to go through the effort to ensure all their dependents are using fallible allocations everywhere.

I am dubious on this API but I'm filling an issue so that it's written down and everyone who wants it has a place to go and assert that it is in fact good and desirable.

Gankro:
Note that in previous topics I stated that libunwind doesn't handle OOM well, but this appears to be outdated or incorrect information. They preallocate their memory before an unwind is even requested. In the limit they might run out and give up but it's not like we've lost anything for trying.

alexcrichton:
I figured I'd comment on this with respect to https://github.com/rust-lang/rust/issues/42808 and the outcome there along with the @rust-lang/libs team discussion. It was discovered in https://github.com/rust-lang/rust/issues/42808 that the integration of the `Alloc` trait caused a number of code size related regressions which [when diagnosed](https://github.com/rust-lang/rust/issues/42808#issuecomment-323839473) were almost exclusively related to unwind tables and landing pads. This investigation then caused https://github.com/rust-lang/rust/pull/44049 as the only two stable implementations of allocators don't panic.

Note, though, that this is not necessarily directly related to OOM but just allocator implementations themselves. Should we forbid, in general, panics from allocators? Should we only allow the oom method to panic?

Unsure!

retep998:
[...]

PR #53575, opened 2018-08-21
Don't reduce E0161 to a warning in NLL migrate mode

matthewjasper:
This error has been on stable for a while, and allowing such code cause the compile to later ICE (since we can't codegen it). Errors `box UNSIZED EXPR` with unsized locals because it's not compatible with the current evaluation order (create the box before evaluating the expressions).

cc #53469 (fixes the ICE in this case)
cc @qnighy

PR #89030, opened 2021-09-16
Introduce `Rvalue::ShallowInitBox`

nbdd0121:
Polished version of #88700.

Implements MCP rust-lang/compiler-team#460, and should allow #43596 to go forward.

In short, creating an empty box is split from a nullary-op `NullOp::Box` into two steps, first a call to `exchange_malloc`, then a `Rvalue::ShallowInitBox` which transmutes `*mut u8` to a shallow-initialized `Box<T>`. This allows the `exchange_malloc` call to unwind. Details can be found in the MCP.

`NullOp::Box` is not yet removed, purely to make reverting easier in case anything goes wrong as the result of this PR. If revert is needed a reversion of "Use Rvalue::ShallowInitBox for box expression" commit followed by a test bless should be sufficient.

Experiments in #88700 showed a very slight compile-time perf regression due to (supposedly) slightly more time spent in LLVM. We could omit unwind edge generation (in non-`oom=panic` case) in box expression MIR construction to restore perf; but I don't think it's necessary since runtime perf isn't affected and perf difference is rather small.

### DIFF
--- a/compiler/rustc_borrowck/src/type_check/mod.rs
+++ b/compiler/rustc_borrowck/src/type_check/mod.rs
@@ -2018,13 +2018,7 @@ impl<'a, 'tcx> TypeChecker<'a, 'tcx> {
                 }
             }
 
-            Rvalue::NullaryOp(_, ty) => {
-                // Even with unsized locals cannot box an unsized value.
-                if self.unsized_feature_enabled() {
-                    let span = body.source_info(location).span;
-                    self.ensure_place_sized(ty, span);
-                }
-
+            Rvalue::NullaryOp(_, ty) | Rvalue::ShallowInitBox(_, ty) => {
                 let trait_ref = ty::TraitRef {
                     def_id: tcx.require_lang_item(LangItem::Sized, Some(self.last_span)),
                     substs: tcx.mk_substs_trait(ty, &[]),
@@ -2357,6 +2351,7 @@ impl<'a, 'tcx> TypeChecker<'a, 'tcx> {
             | Rvalue::AddressOf(..)
             | Rvalue::Len(..)
             | Rvalue::Cast(..)
+            | Rvalue::ShallowInitBox(..)
             | Rvalue::BinaryOp(..)
             | Rvalue::CheckedBinaryOp(..)
             | Rvalue::NullaryOp(..)
--- a/compiler/rustc_middle/src/mir/mod.rs
+++ b/compiler/rustc_middle/src/mir/mod.rs
@@ -2200,6 +2200,12 @@ pub enum Rvalue<'tcx> {
     /// that `Foo` has a destructor. These rvalues can be optimized
     /// away after type-checking and before lowering.
     Aggregate(Box<AggregateKind<'tcx>>, Vec<Operand<'tcx>>),
+
+    /// Transmutes a `*mut u8` into shallow-initialized `Box<T>`.
+    ///
+    /// This is different a normal transmute because dataflow analysis will treat the box
+    /// as initialized but its content as uninitialized.
+    ShallowInitBox(Operand<'tcx>, Ty<'tcx>),
 }
 
 #[cfg(all(target_arch = "x86_64", target_pointer_width = "64"))]
@@ -2449,6 +2455,10 @@ impl<'tcx> Debug for Rvalue<'tcx> {
                         }
                     }),
                 }
+            }
+
+            ShallowInitBox(ref place, ref ty) => {
+                write!(fmt, "ShallowInitBox({:?}, {:?})", place, ty)
             }
         }
     }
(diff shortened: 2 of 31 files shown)


nbdd0121:
The error is due to neglected `no_global_oom_handling` cfg-gating. I filed #89036 to fix it but that means this PR will be blocked until #89036 is merged :(

compiler/rustc_borrowck/src/type_check/mod.rs
```diff
@@ -2334,6 +2334,25 @@ impl<'a, 'tcx> TypeChecker<'a, 'tcx> {
...
+                    let span = body.source_info(location).span;
+                    self.ensure_place_sized(ty, span);
+                }
+
+                let trait_ref = ty::TraitRef {
+                    def_id: tcx.require_lang_item(LangItem::Sized, Some(self.last_span)),
+                    substs: tcx.mk_substs_trait(ty, &[]),
+                };
+
+                self.prove_trait_ref(
```
jonas-schievink, line 2349: I guess this is copied from `NullaryOp` (can the branches be merged?), but I don't get why this is both doing `ensure_place_sized` *and* proving that `ty: Sized`

compiler/rustc_middle/src/mir/mod.rs
```diff
@@ -2450,6 +2456,10 @@ impl<'tcx> Debug for Rvalue<'tcx> {
                     }),
                 }
             }
+
+            ShallowInitBox(ref place, ref ty) => {
+                write!(fmt, "ShallowInitBox({:?}, {:?})", place, ty)
```
jonas-schievink, line 2461: This looks a bit odd in the MIR dumps (with a value and a type being passed the same way), but it does match how `NullaryOp` is printed, so I guess this can be improved later on.

compiler/rustc_middle/src/mir/tcx.rs
```diff
@@ -215,6 +216,7 @@ impl<'tcx> Rvalue<'tcx> {
     pub fn initialization_state(&self) -> RvalueInitializationState {
         match *self {
             Rvalue::NullaryOp(NullOp::Box, _) => RvalueInitializationState::Shallow,
+            Rvalue::ShallowInitBox(_, _) => RvalueInitializationState::Shallow,
```
jonas-schievink, line 219:
```suggestion
            Rvalue::NullaryOp(NullOp::Box, _) |
            Rvalue::ShallowInitBox(_, _) => RvalueInitializationState::Shallow,
```

jonas-schievink (commented):
Changes look generally good to me! A few notes:

- [This comment](https://github.com/rust-lang/rust/blob/a49e38e672c60da788360e088f00ad12353e3913/library/alloc/src/alloc.rs#L310) can now be removed
- Is `NullOp::Box` still used anywhere or can/should it be removed?

compiler/rustc_borrowck/src/type_check/mod.rs
nbdd0121: Indeed. With a bit of archaeology I found that this check was introduced in #53575, and with #87935 the Sized constraint is verified in rustc_typeck so the error path will never be hit. The unsized path definitely could be removed.

nbdd0121:
> Is `NullOp::Box` still used anywhere or can/should it be removed?

No, it won't be used anymore. My thought is that if we keep it, it'll be easier to revert if anything goes wrong. Reverting "Use Rvalue::ShallowInitBox for box expression" would be enough to bring back the old code path.

If you feel that it should be removed in this PR, I can do that as well.

jonas-schievink:
> No, it won't be used anymore. My thought is that if we keep it, it'll be easier to revert if anything goes wrong. Reverting "Use Rvalue::ShallowInitBox for box expression" would be enough to bring back the old code path.

Seems reasonable!

@bors r+ rollup=never

(for the perf triage team: there's a slight perf impact expected, which was previously measured in https://github.com/rust-lang/rust/pull/88700#issuecomment-914600970)

nbdd0121:
> GitHub Actions has encountered an internal error when running your job.

Looks like a bogus failure.

rylev:
@nbdd0121 @jonas-schievink I see there was a bit of a justification for the perf regression given when opening this PR:
> Experiments in #88700 showed a very slight compile-time perf regression due to (supposedly) slightly more time spent in LLVM. We could omit unwind edge generation (in non-oom=panic case) in box expression MIR construction to restore perf; but I don't think it's necessary since runtime perf isn't affected and perf difference is rather small.

The perf regressions aren't huge but I wouldn't classify themselves as small either. We have some non-noisy, real world crate test cases that are showing regressions that are 9 times larger than the calculated significance threshold. This isn't nearly big enough to merit a revert, but if there's a clear path forward for trying to gain the performance back it's probably worth trying to do that sooner rather than later (considering this is for a feature flagged feature and doesn't really benefit any users at this point). Thoughts?

nbdd0121:
The wall-clock time of some of these "regressions" are improved though; instruction count isn't the full picture.

rylev:
@nbdd0121 wall-clock time is an extremely noisy metric which we don't usually look at unless the gains are consistent and large which is not the case here. Of course, instruction count is only part of the picture, but it's generally the best we have for determining runtime performance changes, so I still think discussing the targeted fixed is warranted.

PR #89031, opened 2021-09-16
Don't inline OnceCell initialization closures

the8472:
The more general variant of #89026, originally suggested in https://github.com/rust-lang/rust/pull/86898#issuecomment-920138051

### DIFF
--- a/library/core/src/lazy.rs
+++ b/library/core/src/lazy.rs
@@ -214,7 +214,16 @@ impl<T> OnceCell<T> {
         if let Some(val) = self.get() {
             return Ok(val);
         }
-        let val = f()?;
+        /// Avoid inlining the initialization closure into the common path that fetches
+        /// the already initialized value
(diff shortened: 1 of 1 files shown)


Mark-Simulacrum:
I guess it's possible this will end up hurting performance if the initialization is actually hot (e.g., the OnceCell is guarding something that must be done only once but *usually* only happens once). But that seems like a relatively unlikely edge case, and the wins here are nice -- though notably primarily limited to -doc builds, which suggests that this is mostly an improvement for those compiling without PGO (like rustdoc is today). That's most Rust users though.

@bors r+

pnkfelix:
Weekly perf triage. This was categorized as a regression under our old performance evaluation scheme.

However, we have revised our evaluation metrics, especially how we determine significance, and with that in place, the regression to deep-vector incr-patched: println builds are no longer considered significant in this case.

Issue #66118, opened 2019-11-05
rustc generates invalid DWARF when LTO is enabled


hannobraun:
When combining `lto = true` with `debug = true` in Cargo.toml, the compiler generates invalid DWARF, according to [Gimli](https://github.com/gimli-rs/gimli.git).

I've created a minimal example here: https://github.com/hannobraun/dwarf-test

If you follow the instructions in the example's README, you should see something like this:
```
DWARF error in ../dwarf-test/target/debug/dwarf-test: Invalid intra-unit reference in unit 0x102 from DIE 0x13d1b to 0x492
DWARF error in ../dwarf-test/target/debug/dwarf-test: Invalid intra-unit reference in unit 0x102 from DIE 0x13d30 to 0x491e
DWARF error in ../dwarf-test/target/debug/dwarf-test: Invalid intra-unit reference in unit 0x102 from DIE 0x13d4a to 0x522f
DWARF error in ../dwarf-test/target/debug/dwarf-test: Invalid intra-unit reference in unit 0x102 from DIE 0x13d59 to 0x5e2
DWARF error in ../dwarf-test/target/debug/dwarf-test: Invalid intra-unit reference in unit 0x102 from DIE 0x13d91 to 0x4857
DWARF error in ../dwarf-test/target/debug/dwarf-test: Invalid intra-unit reference in unit 0x102 from DIE 0x13da6 to 0x536c
DWARF error in ../dwarf-test/target/debug/dwarf-test: Invalid intra-unit reference in unit 0x102 from DIE 0x13dce to 0xf5
DWARF error in ../dwarf-test/target/debug/dwarf-test: Invalid intra-unit reference in unit 0x102 from DIE 0x13de2 to 0xe8
DWARF error in ../dwarf-test/target/debug/dwarf-test: Invalid intra-unit reference in unit 0x102 from DIE 0x13dfb to 0x2b31
DWARF error in ../dwarf-test/target/debug/dwarf-test: Invalid intra-unit reference in unit 0x102 from DIE 0x13e10 to 0x4857
DWARF error in ../dwarf-test/target/debug/dwarf-test: Invalid intra-unit reference in unit 0x102 from DIE 0x13e25 to 0x536c
DWARF error in ../dwarf-test/target/debug/dwarf-test: Invalid intra-unit reference in unit 0x102 from DIE 0x13e4d to 0xf5
[...]

PR #46772, opened 2017-12-16
rustc: Work around `DICompileUnit` bugs in LLVM

alexcrichton:
This commit implements a workaround for #46346 which basically just
avoids triggering the situation that LLVM's bug
https://bugs.llvm.org/show_bug.cgi?id=35562 arises. More details can be
found in the code itself but this commit is also intended to ...

Closes #46346

PR #89041, opened 2021-09-17
Work around invalid DWARF bugs for fat LTO

sticnarf:
This PR applies the same workaround in #46772 to fat LTO.

It seems to fix the bug reported in https://github.com/rust-lang/rust/issues/66118#issuecomment-917434036.

### DIFF
--- a/compiler/rustc_codegen_llvm/src/back/lto.rs
+++ b/compiler/rustc_codegen_llvm/src/back/lto.rs
@@ -325,6 +325,20 @@ fn fat_lto(
         drop(linker);
         save_temp_bitcode(&cgcx, &module, "lto.input");
 
+        // Fat LTO also suffers from the invalid DWARF issue similar to Thin LTO.
+        // Here we rewrite all `DICompileUnit` pointers if there is only one `DICompileUnit`.
+        // This only works around the problem when codegen-units = 1.
+        // Refer to the comments in the `optimize_thin_module` function for more details.
+        let mut cu1 = ptr::null_mut();
+        let mut cu2 = ptr::null_mut();
+        unsafe { llvm::LLVMRustLTOGetDICompileUnit(llmod, &mut cu1, &mut cu2) };
+        if !cu2.is_null() {
+            let _timer =
+                cgcx.prof.generic_activity_with_arg("LLVM_fat_lto_patch_debuginfo", &*module.name);
+            unsafe { llvm::LLVMRustLTOPatchDICompileUnit(llmod, cu1) };
+            save_temp_bitcode(cgcx, &module, "fat-lto-after-patch");
+        }
+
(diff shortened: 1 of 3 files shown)


oli-obk:
r? rust-lang/wg-llvm

nagisa:
Can you please add a regression test?

sticnarf:
> 
> 
> Can you please add a regression test? If its too hard then its okay to r=me this.

Sorry but now I don't know how to construct a minimized case for reproducing the issue from Rust code (as you see the code I posted in the issue page has multiple dependencies, and the issue disappears after removing any of them). So I'm afraid it's really too hard for me to add a regression test for it.

compiler/rustc_codegen_llvm/src/back/lto.rs
```diff
@@ -325,6 +325,20 @@ fn fat_lto(
...
         save_temp_bitcode(&cgcx, &module, "lto.input");
 
+        // Fat LTO also suffers from the invalid DWARF issue similar to Thin LTO.
+        // Here we rewrite all `DICompileUnit` pointers if there is only one `DICompileUnit`.
+        // This only works around the problem when codegen-units = 1.
+        // Refer to the comments in the `optimize_thin_module` function for more details.
+        let mut cu1 = ptr::null_mut();
+        let mut cu2 = ptr::null_mut();
+        unsafe { llvm::LLVMRustLTOGetDICompileUnit(llmod, &mut cu1, &mut cu2) };
+        if !cu2.is_null() {
```
philipc, line 335: The Thin LTO use of this seems to have the opposite check? It only proceeds if cu2 is nul.

sticnarf: Oops... I didn't mean to do the opposite, actually.

However, after the patch, rustc seems to generate valid dwarf for binaries (such as `ripgrep`) while using full LTO. So I truly misunderstood something.

I am not sure how much missing debug info it will increase. But at least Linux perf does not show unknown userspace addresses in my practice...

philipc: I tried it out. It seems to be doing what I expect (puts everything in one huge CU), and this check makes sense to me (it's only worth trying to combine CUs if there is more than one CU), so it's probably just the comment that is wrong.

sticnarf:
It seems quite some regressions are reported due to this PR. Better to revert this immature workaround.

cbiffle:
> It seems quite some regressions are reported due to this PR. Better to revert this immature workaround.

Can't really tell from the github output here, did this happen? This change without regression tests has been suggested as the culprit for broken DWARF output for static vars in LTO'd binaries over in #90357.

If this is still in-tree we can try reverting it on our side, I suppose.

PR #89045, opened 2021-09-17
Register normalization obligations instead of immediately normalizing in opaque type instantiation

oli-obk:
For lazy TAIT we will need to instantiate opaque types from within `rustc_infer`, which cannot invoke normalization methods (they are in `rustc_trait_resolution`). So before we move the logic over to `rustc_infer`, we need make sure no normalization happens anymore. This PR resolves that by just registering normalization obligations and continuing.

This PR is best reviewed commit by commit

I also included f7ad36e which is just an independent cleanup that touches the same code and reduces diagnostics noise a bit

r? @nikomatsakis cc @spastorino

### DIFF
--- a/compiler/rustc_infer/src/infer/projection.rs
+++ b/compiler/rustc_infer/src/infer/projection.rs
@@ -1,0 +1,33 @@
+use rustc_middle::traits::ObligationCause;
+use rustc_middle::ty::{self, ToPredicate, Ty};
+
+use crate::traits::{Obligation, PredicateObligation};
+
+use super::type_variable::{TypeVariableOrigin, TypeVariableOriginKind};
+use super::InferCtxt;
+
+impl<'a, 'tcx> InferCtxt<'a, 'tcx> {
+    pub fn infer_projection(
+        &self,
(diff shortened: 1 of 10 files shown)


compiler/rustc_infer/src/infer/projection.rs
```diff
@@ -0,0 +1,33 @@
+use rustc_middle::traits::ObligationCause;
+use rustc_middle::ty::{self, ToPredicate, Ty};
+
+use crate::traits::{Obligation, PredicateObligation};
+
+use super::type_variable::{TypeVariableOrigin, TypeVariableOriginKind};
+use super::InferCtxt;
+
+impl<'a, 'tcx> InferCtxt<'a, 'tcx> {
+    pub fn infer_projection(
```
nikomatsakis, line 10: Comment please =) What is the purpose of this function?

nikomatsakis (approved):

nikomatsakis:
r=me once comment is added

oli-obk:
looks like codegen noise to me, and this PR should not affect codegen at all, it's purely in the type system

@bors r=nikomatsakis

PR #89046, opened 2021-09-17
"Fix" an overflow in byte position math

oli-obk:
r? @estebank 

help! I fixed the ICE only to brick the diagnostic.

I mean, it was wrong previously (using an already expanded macro span), but it is really bad now XD

### DIFF
--- a/compiler/rustc_errors/src/lib.rs
+++ b/compiler/rustc_errors/src/lib.rs
@@ -13,9 +13,11 @@
 #[macro_use]
 extern crate rustc_macros;
 
+#[macro_use]
+extern crate tracing;
+
 pub use emitter::ColorConfig;
 
-use tracing::debug;
 use Level::*;
 
 use emitter::{is_case_difference, Emitter, EmitterWriter};
--- a/compiler/rustc_span/src/lib.rs
+++ b/compiler/rustc_span/src/lib.rs
@@ -785,13 +785,30 @@ impl Span {
     ///     ^^^^^^^^^^^^^^^^^
     /// ```
     pub fn until(self, end: Span) -> Span {
-        let span = self.data();
-        let end = end.data();
+        // Most of this function's body is copied from `to`.
+        // We can't just do `self.to(end.shrink_to_lo())`,
+        // because to also does some magic where it uses min/max so
+        // it can handle overlapping spans. Some advanced mis-use of
+        // `until` with different ctxts makes this visible.
+        let span_data = self.data();
+        let end_data = end.data();
+        // FIXME(jseyfried): `self.ctxt` should always equal `end.ctxt` here (cf. issue #23480).
+        // Return the macro span on its own to avoid weird diagnostic output. It is preferable to
+        // have an incomplete span than a completely nonsensical one.
+        if span_data.ctxt != end_data.ctxt {
+            if span_data.ctxt == SyntaxContext::root() {
+                return end;
+            } else if end_data.ctxt == SyntaxContext::root() {
+                return self;
+            }
+            // Both spans fall within a macro.
+            // FIXME(estebank): check if it is the *same* macro.
+        }
         Span::new(
-            span.lo,
-            end.lo,
-            if end.ctxt == SyntaxContext::root() { end.ctxt } else { span.ctxt },
-            if span.parent == end.parent { span.parent } else { None },
+            span_data.lo,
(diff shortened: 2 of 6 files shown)


compiler/rustc_span/src/lib.rs
```diff
@@ -782,13 +785,25 @@ impl Span {
     ///     ^^^^^^^^^^^^^^^^^
     /// ```
     pub fn until(self, end: Span) -> Span {
-        let span = self.data();
-        let end = end.data();
+        let span_data = self.data();
```
oli-obk, line 788: Most of this function's body is copied from `to`. We can't just do `self.to(end.shrink_to_lo())`, because `to` also does some magic where it uses min/max so it can handle overlapping spans (I think, the logic is a bit wonky).

compiler/rustc_errors/src/lib.rs
```diff
@@ -339,10 +345,12 @@ impl CodeSuggestion {
                     });
                     buf.push_str(&part.snippet);
                     let cur_hi = sm.lookup_char_pos(part.span.hi());
+                    trace!(?cur_hi);
                     if prev_hi.line == cur_lo.line {
                         // Account for the difference between the width of the current code and the
                         // snippet being suggested, so that the *later* suggestions are correctly
                         // aligned on the screen.
+                        trace!(?len);
                         acc += len as isize - (cur_hi.col.0 - cur_lo.col.0) as isize;
```
oli-obk, line 354: The overflow happens here, because the lines of cur_hi and cur_lo are never checked against each other, so if lo starts in an earlier line but with a higher column, we get an overflow.

Now, we can properly fix this, and probably should, but the span that we have here is nonsensical to begin with. It goes from `bar(` inside the macro to the start of `true` at the *call site* of the macro.

estebank: Ah! Yes, that's why we are getting these malformed spans. Yeah, macros screw everything up. Maybe we should have an `assert` (not an assert, a check) for well formed spans before we even try to emit or add a suggestion to the `DiagnosticBuilder`? That would make it so we never show suggestions like this one, while also not breaking the emitter logic. We should probably log that though.

estebank: I _hate_ Spans sourced from macros 🥺

estebank: We could add an actual `assert!` here. At least if anyone hits it we'll get a traceback.

compiler/rustc_span/src/lib.rs
estebank: Add a comment saying so _or_ refactor the common logic to a new method.

estebank (approved):
r=me without the new tracing logic (or should we keep it?)

code changes are good, leaving it to you whether we address my comments on this PR or afterwards.

compiler/rustc_errors/src/lib.rs
oli-obk: By the time we get here, the span doesn't look malformed anymore. The fault is either at the until method or at the caller to it. Maybe we should return an option instead of a span? Or a dummy span in case there are macros involved?

estebank: We could tell that `lo > hi`, right? And in the Span wrangling methods we should check that both spans belong to the same Ctxt, and if not die in nightly and return only one of the original spans in beta/stable. Changing this to `Option` would be a royal pain in the neck.

oli-obk: yea, debug asserting this would be best. We'll have to fix a bunch of diagnostics code that currently relies on it.

camsteffen: I'm bumping into this overflow (again) in Clippy. Still not sure why it hasn't shown up in CI. But I narrowed it down to a suggestion that replaces an else block, so the block starts at a higher column than where it ends.

estebank: @camsteffen yeah, didn't ping you on this PR when oli noticed this, but it is likely the same issue and this PR should help (but likely give... incorrect suggestions).

camsteffen: I don't think the problem is only with invalid spans. Here is a minimal reproduction (triggers anonymous_parameters).

```rust
trait X {
    fn test(x: u32, (
    )) {}
}
```

PR #89050, opened 2021-09-17
refactor: VecDeques Drain fields to private

DeveloperC286:
Made the fields of VecDeque's Drain private by creating a Drain::new(...) function to create a new instance of Drain and migrating usage to use Drain::new(...).


joshtriplett:
This seems like a reasonable refactor to me.

Shouldn't the `new` function be `unsafe`, though, since it effectively works with raw pointers?

DeveloperC286:
> This seems like a reasonable refactor to me.
> 
> Shouldn't the `new` function be `unsafe`, though, since it effectively works with raw pointers?

Updated the `new` function to be unsafe.

Issue #73445, opened 2020-06-17
intra-rustdoc-link: Cannot use core in core crate


lzutao:
I tried to document this code in libcore (`./x.py doc src/libcore`):

```rust
impl<T: Deref, E> Result<T, E> {
    /// Converts from `Result<T, E>` (or `&Result<T, E>`) to `Result<&<T as Deref>::Target, &E>`.
    ///
    /// [`core::ops::Deref`]
    pub fn as_deref(&self) -> Result<&T::Target, &E>;
}
```

I expected to see the above code build pass.

Instead, this happened: Build failed the following errors:

```rust
 Documenting core v0.0.0 (/home/lzutao/fork/rust/compiler/src/libcore)
error: `[core::ops::Deref]` cannot be resolved, ignoring it.
    --> src/libcore/result.rs:1151:68
     |
1151 |     /// Coerces the [`Ok`] variant of the original [`Result`] via [`core::ops::Deref`]
     |                                                                    ^^^^^^^^^^^^^^^^^^ cannot be resolved, ignoring
     |
note: the lint level is defined here
    --> src/libcore/lib.rs:64:9
     |
64   | #![deny(intra_doc_link_resolution_failure)] // rustdoc is run without -D warnings
     |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
     = help: to escape `[` and `]` characters, just add '\' before them like `\[` or `\]`

error: Could not document `core`.
```

cc #43466
cc @Manishearth @GuillaumeGomez @jyn514

lzutao:
Note: Document build pass in user crate:
```rust
/// Intra link testings
///
/// [`Deref::Target`](core::ops::Deref)
pub struct Foo;
```

jyn514:
What happens if you use `crate::ops::Deref` instead?

jyn514:
Actually this might be a markdown issue, does `[Deref::Target](core::ops::Deref)` work in libcore?

Manishearth:
@lzutao `core::ops::Deref` will never work in the core crate, crates do not know their own names typically

Manishearth:
@jyn514 no, it's not a markdown issue

jplatte:
> crates do not know their own names typically

Which can be worked around with an `extern crate self as foo;` in the crate root, in case anybody needs that (I've previously required it to get a proc macro working both inside its "parent crate" and outside of it).

lzutao:
[...]

PR #89051, opened 2021-09-17
Add intra-doc links and small changes to `std::os` to be more consistent

schctl:
I believe that a few items in `std::os` should be linked. I've also added a basic example in `std::os::windows`.

### DIFF
--- a/library/std/src/os/linux/fs.rs
+++ b/library/std/src/os/linux/fs.rs
@@ -1,4 +1,6 @@
-//! Linux-specific extensions to primitives in the `std::fs` module.
+//! Linux-specific extensions to primitives in the [`std::fs`] module.
+//!
+//! [`std::fs`]: crate::fs
 
 #![stable(feature = "metadata_ext", since = "1.1.0")]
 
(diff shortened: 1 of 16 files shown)


library/std/src/os/linux/fs.rs
```diff
@@ -1,4 +1,6 @@
-//! Linux-specific extensions to primitives in the `std::fs` module.
+//! Linux-specific extensions to primitives in the [`std::fs`] module.
+//!
+//! [`std::fs`]: crate::fs
```
jyn514, line 3: cc https://github.com/rust-lang/rust/issues/73445#issuecomment-830141476

library/std/src/os/windows/mod.rs
```diff
@@ -5,6 +5,22 @@
 //! the core `std` library. These extensions allow developers to use
 //! `std` types and idioms with Windows in a way that the normal
 //! platform-agnostic idioms would not normally support.
+//!
+//! # Examples
+//!
+//! ```no_run
+//! use std::fs::File;
+//! use std::os::windows::prelude::*;
```
jyn514, line 13: This will fail on any OS other than windows.
```suggestion
#![cfg_attr(windows, doc = "```no_run")]
#![cfg_attr(not(windows), doc = "```ignore")]
//! use std::fs::File;
//! use std::os::windows::prelude::*;
```

jyn514: Err, I just realized the whole `windows` module has `cfg(windows)`. So this should be fine.

PR #89053, opened 2021-09-17
refactor: VecDeques IntoIter fields to private

DeveloperC286:
Made the fields of VecDeque's IntoIter private by creating a IntoIter::from(...) function to create a new instance of IntoIter and migrating usage to use IntoIter::from(...).

### DIFF
--- a/library/alloc/src/collections/vec_deque/into_iter.rs
+++ b/library/alloc/src/collections/vec_deque/into_iter.rs
@@ -17,7 +17,13 @@ pub struct IntoIter<
     T,
     #[unstable(feature = "allocator_api", issue = "32838")] A: Allocator = Global,
 > {
-    pub(crate) inner: VecDeque<T, A>,
+    inner: VecDeque<T, A>,
+}
+
+impl<T, A: Allocator> IntoIter<T, A> {
+    pub(super) fn new(inner: VecDeque<T, A>) -> Self {
+        IntoIter { inner }
+    }
 }
 
 #[stable(feature = "collection_debug", since = "1.17.0")]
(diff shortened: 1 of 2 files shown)


library/alloc/src/collections/vec_deque/into_iter.rs
```diff
@@ -17,7 +17,13 @@ pub struct IntoIter<
     T,
     #[unstable(feature = "allocator_api", issue = "32838")] A: Allocator = Global,
 > {
-    pub(crate) inner: VecDeque<T, A>,
+    inner: VecDeque<T, A>,
+}
+
+impl<T, A: Allocator> IntoIter<T, A> {
+    pub(super) fn from(inner: VecDeque<T, A>) -> Self {
```
Mark-Simulacrum, line 24: nit: I feel like I would name this function new, not from. It's not in  a From impl...

Mark-Simulacrum (commented):
Is there a motivation for this change? It looks like it's safe currently to construct the iterator from any VecDeque and there's no internal state otherwise, so the added constructor doesn't seem necessary?

DeveloperC286:
> Is there a motivation for this change? It looks like it's safe currently to construct the iterator from any VecDeque and there's no internal state otherwise, so the added constructor doesn't seem necessary?

Hey, thank you for your quick feedback. My motivation behind this change and a few others I will be submitting, is I was told having fields of a struct accessible directly was a code smell. While no code is currently directly accessing the field this removes the possibility.

library/alloc/src/collections/vec_deque/into_iter.rs
DeveloperC286: Renamed the function.

Mark-Simulacrum:
@bors r+ rollup

Alright, seems OK. I don't think this was necessarily a code smell or otherwise bad previously, since it was in practice just a crate-local field -- that's usually pretty harmless, especially when there's no invariants attached to the field (like here). But it is a little nicer if you can be confident that no other module is touching the field, so I'll approve this PR :)

PR #89055, opened 2021-09-17
Suggest better place to add call parentheses for method expressions wrapped in parentheses

### LINKED ISSUE #89044: With over-parenthesized method that isn't called, E0615 suggests adding parentheses in the wrong place
Given the following [code](https://play.rust-lang.org/?version=nightly&mode=debug&edition=2018&gist=b36d6150a12ac621b7d0b38568538f22): 

```rust
fn main() {
    let a = Some(42);
    println!(
        "The value is {}.",
        (a.unwrap)
    );
}
```

The current output is:

```
error[E0615]: attempted to take value of method `unwrap` on type `Option<{integer}>`
 --> src/main.rs:5:12
  |
5 |         (a.unwrap)
  |            ^^^^^^ method, not a field
  |
help: use parentheses to call the method
  |
5 |         (a.unwrap)()
  |                   ++
```

The fix suggested by the compiler still has the problem, because the parentheses-to-call are added outside of the parentheses in the code.

```
help: use parentheses to call the method
  |
5 |         (a.unwrap())
  |                  ++
```

This is similar to the recently fixed #88803, but differs in that the input code cannot be fixed solely by removing parentheses.

Kobzol:
I wanted to improve the suggestion a bit to both remove the wrapping parentheses **and** add call parentheses by both calling `suggest_method_call` and using `multipart_suggestion`. But I very quickly ran into a problem where multiple overlapping machine applicable suggestions cannot be properly applied together. So I applied the suggestion from the issue and only added the call parentheses directly after the expression.

Fixes: https://github.com/rust-lang/rust/issues/89044

### DIFF
--- a/compiler/rustc_typeck/src/check/expr.rs
+++ b/compiler/rustc_typeck/src/check/expr.rs
@@ -1844,6 +1844,7 @@ impl<'a, 'tcx> FnCtxt<'a, 'tcx> {
                 field,
                 expr_t,
                 expr,
+                None,
             );
         }
         err.emit();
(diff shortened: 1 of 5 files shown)


wesleywiser (approved):
Awesome! 👍

Issue #88827, opened 2021-09-10
Crater runs for 1.56


tmandry:
Note: Please do not conduct triage on these runs without discussing how to do so with a release team member first. Thanks!

tmandry:
@craterbot run name=beta-1.56-1 start=1.55.0 end=beta-2021-09-07 mode=build-and-test cap-lints=warn p=10

tmandry:
@craterbot name=beta-1.56-1 end=beta-2021-09-08

tmandry:
@craterbot abort name=beta-1.56-1

tmandry:
@craterbot run name=beta-1.56-1 start=1.55.0 end=beta-2021-09-08 mode=build-and-test cap-lints=warn p=10

tmandry:
@craterbot run name=beta-1.56-rustdoc-1 start=1.55.0 end=beta-2021-09-08 mode=rustdoc cap-lints=warn p=5

Mark-Simulacrum:
@craterbot retry name=beta-1.56-rustdoc-1

Mark-Simulacrum:
@craterbot retry-report name=beta-1.56-1

Mark-Simulacrum:
@craterbot run name=beta-1.56-2 start=1.55.0 end=beta-2021-09-12 mode=build-and-test cap-lints=warn p=10 crates=https://crater-reports.s3.amazonaws.com/beta-1.56-1/retry-regressed-list.txt

Mark-Simulacrum:
rustc run:

* https://github.com/rust-lang/rust/issues/88967
* https://github.com/rust-lang/rust/issues/87757
* https://github.com/rust-lang/rust/issues/88969
* https://github.com/rust-lang/rust/issues/88970
* https://github.com/rust-lang/rust/issues/88971

Mark-Simulacrum:
@craterbot retry

PR #89056, opened 2021-09-17
make member constraints pick static if no upper bounds

### LINKED ISSUE #63033: Async fn does not compile if lifetime does not appear in bounds (sometimes)
```rust
#![feature(async_await)]

use std::marker::PhantomData;

trait Trait {}

async fn f<'a>(reference: &(), marker: PhantomData<dyn Trait>) {}
```

This fails with:

```console
error[E0700]: hidden type for `impl Trait` captures lifetime that does not appear in bounds
 --> src/main.rs:7:64
  |
7 | async fn f<'a>(reference: &(), marker: PhantomData<dyn Trait>) {}
  |                                                                ^
  |
note: hidden type `impl std::future::Future` captures the scope of call-site for function at 7:64
 --> src/main.rs:7:64
  |
7 | async fn f<'a>(reference: &(), marker: PhantomData<dyn Trait>) {}
  |                                                                ^^
```

I believe this should compile because:

1. The non-async function with the same signature does compile;

    ```rust
    fn f<'a>(reference: &(), marker: PhantomData<dyn Trait>) {}
    ```

2. Seemingly insignificant tweaks, like adding a wrapper struct, make it compile.

    ```rust
    struct Wrapper(PhantomData<dyn Trait>);
    async fn f<'a>(reference: &(), marker: Wrapper) {}
    ```

Mentioning @cramertj who fixed a similar error message in #59001.
Mentioning @nikomatsakis who touched this error message in #49041 and #61775.

*rustc 1.38.0-nightly (c43753f91 2019-07-26)*

nikomatsakis:
The current member constraint algorithm has a failure mode when it encounters a
variable `'0` and the only constraint is that `':static: '0`. In that case,
there are no upper bounds, or at least no non-trivial upper bounds.  As a
result, we are not able to rule out any of the choices, so if you have a
constraint like `'0 member ['a, 'b, 'static]`, where `'a` and `'b` are
unrelated, then the algorithm gets stuck as there is no 'least choice' from
that set.

The tweak in this commit changes the algorithm so that *if* there are no upper
bounds (and hence `'0` can get as large as desired without creating a region
check error), it will just pick `'static`. This should only occur in cases
where the data is flowing out from a `'static` value.

This change is probably *not* right for impl Trait in let bindings, but those
are challenging with member constraints anyway, and not currently supported.
Furthermore, this change is not needed in a polonius-like formulation, which
effectively permits "ad-hoc intersections" of lifetimes as the value for a
region, and hence could give a value like `'a ^ 'b` as the resulting lifetime.
Therefore I think there isn't forwards compat danger here. (famous last words?)

r? @lqd 
cc @tmandry @eholk @jackh726 @pnkfelix 

I was thinking it might be nice to schedule a recorded zoom session to talk over this PR and how this bit of the code works, as I imagine it has a ... low bus factor. =)

Fixes #63033

### DIFF
--- a/compiler/rustc_borrowck/src/region_infer/mod.rs
+++ b/compiler/rustc_borrowck/src/region_infer/mod.rs
@@ -733,44 +733,58 @@ impl<'tcx> RegionInferenceContext<'tcx> {
             return false;
         }
 
-        // If there WERE no upper bounds (apart from static), and static is one of the options,
-        // then we can just pick that (c.f. #63033).
-        let choice = if !any_upper_bounds && choice_regions.contains(&fr_static) {
-            fr_static
-        } else {
-            // Otherwise, we need to find the minimum remaining choice, if
-            // any, and take that.
-            debug!("apply_member_constraint: choice_regions remaining are {:#?}", choice_regions);
-            let min = |r1: ty::RegionVid, r2: ty::RegionVid| -> Option<ty::RegionVid> {
-                let r1_outlives_r2 = self.universal_region_relations.outlives(r1, r2);
-                let r2_outlives_r1 = self.universal_region_relations.outlives(r2, r1);
-                match (r1_outlives_r2, r2_outlives_r1) {
-                    (true, true) => Some(r1.min(r2)),
-                    (true, false) => Some(r2),
-                    (false, true) => Some(r1),
-                    (false, false) => None,
-                }
-            };
-            let mut min_choice = choice_regions[0];
-            for &other_option in &choice_regions[1..] {
-                debug!(
-                    "apply_member_constraint: min_choice={:?} other_option={:?}",
-                    min_choice, other_option,
-                );
-                match min(min_choice, other_option) {
-                    Some(m) => min_choice = m,
-                    None => {
-                        debug!(
-                            "apply_member_constraint: {:?} and {:?} are incomparable; no min choice",
-                            min_choice, other_option,
+        // Otherwise, we need to find the minimum remaining choice, if
+        // any, and take that.
+        debug!("apply_member_constraint: choice_regions remaining are {:#?}", choice_regions);
+        let min = |r1: ty::RegionVid, r2: ty::RegionVid| -> Option<ty::RegionVid> {
+            let r1_outlives_r2 = self.universal_region_relations.outlives(r1, r2);
+            let r2_outlives_r1 = self.universal_region_relations.outlives(r2, r1);
+            match (r1_outlives_r2, r2_outlives_r1) {
+                (true, true) => Some(r1.min(r2)),
+                (true, false) => Some(r2),
+                (false, true) => Some(r1),
+                (false, false) => None,
+            }
+        };
+        let mut min_choice = choice_regions[0];
+        for &other_option in &choice_regions[1..] {
+            debug!(
+                "apply_member_constraint: min_choice={:?} other_option={:?}",
+                min_choice, other_option,
+            );
+            match min(min_choice, other_option) {
+                Some(m) => min_choice = m,
+                None => {
+                    debug!(
+                        "apply_member_constraint: {:?} and {:?} are incomparable; no min choice",
+                        min_choice, other_option,
+                    );
+
+                    // If there is no minimum choice, then we *may* wish to use `'static`
+                    // instead, but only if there are no upper bounds, and `'static` is a valid
+                    // choice.
+                    if !any_upper_bounds && choice_regions.contains(&fr_static) {
+                        return self.apply_member_constraint_choice(
+                            scc,
+                            member_constraint_index,
+                            fr_static,
                         );
-                        return false;
                     }
+
+                    // Otherwise, we give up.
+                    return false;
                 }
             }
-            min_choice
-        };
+        }
+        self.apply_member_constraint_choice(scc, member_constraint_index, min_choice)
+    }
 
+    fn apply_member_constraint_choice(
+        &mut self,
+        scc: ConstraintSccIndex,
+        member_constraint_index: NllMemberConstraintIndex,
+        choice: ty::RegionVid,
+    ) -> bool {
         let choice_scc = self.constraint_sccs.scc(choice);
         debug!("apply_member_constraint: min_choice={:?} best_choice_scc={:?}", choice, choice_scc,);
         if self.scc_values.add_region(scc, choice_scc) {
--- a/compiler/rustc_borrowck/src/region_infer/mod.rs
+++ b/compiler/rustc_borrowck/src/region_infer/mod.rs
@@ -717,11 +717,14 @@ impl<'tcx> RegionInferenceContext<'tcx> {
         // free region that must outlive the member region `R0` (`UB:
         // R0`). Therefore, we need only keep an option `O` if `UB: O`
         // for all UB.
+        let fr_static = self.universal_regions.fr_static;
         let rev_scc_graph = self.reverse_scc_graph();
         let universal_region_relations = &self.universal_region_relations;
-        for ub in rev_scc_graph.upper_bounds(scc) {
+        let mut any_upper_bounds = false;
+        for ub in rev_scc_graph.upper_bounds(scc).filter(|ub| *ub != fr_static) {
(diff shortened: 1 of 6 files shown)


compiler/rustc_infer/src/infer/lexical_region_resolve/mod.rs
```diff
@@ -287,16 +291,31 @@ impl<'cx, 'tcx> LexicalResolver<'cx, 'tcx> {
...
 
         // Find all the "upper bounds" -- that is, each region `b` such that
         // `r0 <= b` must hold.
         let (member_upper_bounds, ..) =
             self.collect_bounding_regions(graph, member_vid, OUTGOING, None);
+        debug!("enforce_member_constraint: upper_bounds={:#?}", member_upper_bounds);
+
+        // If there are no upper bounds, and static is a choice (in practice, it always is),
+        // then we should just pick static.
+        if member_upper_bounds.is_empty()
```
danielhenrymantilla, line 308: I notice the check here is simpler than the one with `any_upper_bounds` done above (_e.g._, non-empty case with only `'static`s inside): does `collect_bounding_regions` never yield `'static`s?

lqd:
> I was thinking it might be nice to schedule a recorded zoom session to talk over this PR and how this bit of the code works, as I imagine it has a ... low bus factor. =)

Sounds good. Even though the docs on member constraints are excellent, diving into the code and match it to the prose will still be enlightening, I expect.

nikomatsakis:
@bors try 

Seems like a good idea to do a check crater run

lqd:
> Seems like a good idea to do a check crater run

Let's do that, the queue is empty ! 

@craterbot check

compiler/rustc_infer/src/infer/lexical_region_resolve/mod.rs
lqd: Conjecture on my part: I wouldn't expect it to ? It wouldn't constrain the variable to add it as an upper bound, it's already the `⊤` element. 

We'll see when we do the Zoom session ^^

lqd:
I've been looking at the results, and as mentioned on Zulip, here's the `bb8-0.3.x` regression minimized:

```rust
trait Future {}

fn sink_error1<'a, F>(f: F) -> impl Future + 'a
where
    F: Future + 'a,
{
    sink_error2(f) // error: F may not live long enough
}

fn sink_error2<'a, F>(f: F) -> impl Future + 'a
where
    F: Future + 'a,
{
    f
}
```

compiler/rustc_borrowck/src/region_infer/mod.rs
```diff
@@ -730,46 +733,50 @@ impl<'tcx> RegionInferenceContext<'tcx> {
...
+        } else {
+            // Otherwise, we need to find the minimum remaining choice, if
+            // any, and take that.
+            debug!("apply_member_constraint: choice_regions remaining are {:#?}", choice_regions);
+            let min = |r1: ty::RegionVid, r2: ty::RegionVid| -> Option<ty::RegionVid> {
+                let r1_outlives_r2 = self.universal_region_relations.outlives(r1, r2);
+                let r2_outlives_r1 = self.universal_region_relations.outlives(r2, r1);
+                match (r1_outlives_r2, r2_outlives_r1) {
+                    (true, true) => Some(r1.min(r2)),
+                    (true, false) => Some(r2),
```
danielhenrymantilla, line 749: Btw, unrelated to this PR, just out of curiosity / to make sure I understand this part correctly: wouldn't this second branch already cover the first? (at the cost of breaking the "symmetry" in the code)

compiler/rustc_infer/src/infer/lexical_region_resolve/mod.rs
danielhenrymantilla: Was the zoom session recorded somewhere btw?

lqd: It'll [likely happen tomorrow](https://rust-lang.zulipchat.com/#narrow/stream/187312-wg-async-foundations/topic/member.20constraints.20PR.20.2389056/near/254337073)

nikomatsakis:
@lqd that's an interesting regression. I guess the problem is that our region inference here is kind of "missing" some of the *actual* bounds. Let me think about that.

nikomatsakis:
@bors try

Pushed a fix for that crater case. We can re-run crater, perhaps.

jackh726:
We should be able to only rerun crater on the regressed crates. I don't know if we really need to do a full crater run. (In fact, if we do the subset, it might be done by the meeting tomorrow since the queue is empty)

@Mark-Simulacrum can you help here? I'm not entirely sure how to run with a subset.

Mark-Simulacrum:
https://github.com/rust-lang/rust/issues/88827#issuecomment-919368935 is an example of queueing the regressions from a prior crater run, you can probably duplicate that once the try is done here.

jackh726:
@craterbot run name=pr-89056-2 start=db1fb85cff63ad5fffe435e17128f99f9e1d970c end=8de4b308b608cded0ebccf658a0fc719bce87d4e mode=build-and-test crates=https://crater-reports.s3.amazonaws.com/pr-89056/retry-regressed-list.txt

Mark-Simulacrum:
@craterbot retry

Probably just startup trouble, so let's try again. If not I'll take a look tomorrow.

jackh726:
@craterbot name=pr-89056-2 start=master#db1fb85cff63ad5fffe435e17128f99f9e1d970c end=try#8de4b308b608cded0ebccf658a0fc719bce87d4e

let's try this..

jackh726:
@craterbot abort name=pr-89056-2

jackh726:
@craterbot run name=pr-89056-2 start=master#db1fb85cff63ad5fffe435e17128f99f9e1d970c end=try#8de4b308b608cded0ebccf658a0fc719bce87d4e mode=build-and-test crates=https://crater-reports.s3.amazonaws.com/pr-89056/retry-regressed-list.txt

jackh726:
@craterbot abort name=pr-89056-2

🤦‍♀️ I used the wrong try build...

jackh726:
@craterbot run name=pr-89056-2 start=master#db1fb85cff63ad5fffe435e17128f99f9e1d970c end=try#11084977fccce286b348dbb5473938f22b430250 mode=build-and-test crates=https://crater-reports.s3.amazonaws.com/pr-89056/retry-regressed-list.txt

lqd:
@craterbot run name=pr-89056-3 start=try#8de4b308b608cded0ebccf658a0fc719bce87d4e end=try#11084977fccce286b348dbb5473938f22b430250 mode=check-only crates=https://crater-reports.s3.amazonaws.com/pr-89056/retry-regressed-list.txt

Mark-Simulacrum:
@craterbot abort name=pr-89056-2

oli-obk:
I haven't gotten beyond code reading on the intersection regions. Let's unblock this, as afaict this is strictly forwards compatible with intersection regions

lqd:
This PR has been brought up during today's t-compiler meeting as part of the "oldest PRs waiting for review".

I don't think we've published the walkthrough on youtube we did back then, right ? I would have loved to rewatch it, and bring back some of that context in cache.

If we're going with this PR rather than intersection regions, then
- it's going to need a rebase. That will also make sure this doesn't change diagnostics, as we expect.
- a new crater run (a bunch of async code must have been published to crates.io in the meantime, so it will be good to check) but it will likely come back clean.
- I believe @jackh726's concern at the time was resolved by https://github.com/rust-lang/rust/pull/89327
- I remember the walkthrough making sense to me at the time within the context of member constraints, and this PR looking good, especially with the concern above having been resolved.

Here's a @rust-lang/types ping to notify interested parties who weren't involved when this PR was initially made, and a @rust-lang/wg-nll one for good measure.

With that, @rustbot author.

aliemjay:
cc #105300

lcnr:
unnominating in favor of an fcp in https://github.com/rust-lang/rust/pull/105300

jackh726:
Closing since #105300 has landed and is a more general version of this.

PR #89059, opened 2021-09-18
[WIP] cstore: Add a new caching layer for data decoded from resolver

petrochenkov:
Implement something to address https://rust-lang.zulipchat.com/#narrow/stream/131828-t-compiler/topic/.22Querifying.22.20early.20cstore.20accesses/near/253737343

One obvious further optimization would be to use some kind of arena to avoid cloning heavy decoded data like vectors.

### DIFF
--- a/compiler/rustc_metadata/src/rmeta/decoder.rs
+++ b/compiler/rustc_metadata/src/rmeta/decoder.rs
@@ -28,7 +28,7 @@ use rustc_middle::mir::interpret::{AllocDecodingSession, AllocDecodingState};
 use rustc_middle::mir::{self, Body, Promoted};
 use rustc_middle::thir;
 use rustc_middle::ty::codec::TyDecoder;
-use rustc_middle::ty::{self, Ty, TyCtxt, Visibility};
+use rustc_middle::ty::{self, TcxArena, Ty, TyCtxt, Visibility};
 use rustc_serialize::{opaque, Decodable, Decoder};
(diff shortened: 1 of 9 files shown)


compiler/rustc_metadata/src/rmeta/decoder.rs
```diff
@@ -714,23 +728,29 @@ impl<'a, 'tcx> CrateMetadataRef<'a> {
...
+        *self.def_kinds.borrow_mut().entry(item_id).or_insert_with(|| {
+            self.root.tables.def_kind.get(self, item_id).map(|k| k.decode(self)).unwrap_or_else(
+                || {
+                    bug!(
+                        "CrateMetadata::def_kind({:?}): id not found, in crate {:?} with number {}",
+                        item_id,
+                        self.root.name,
+                        self.cnum,
+                    )
+                },
```
bjorn3, line 740: `DefKind` is very cheap to decode I think. https://github.com/rust-lang/rust/blob/c7dbe7a830100c70d59994fd940bf75bb6e39b39/compiler/rustc_hir/src/def.rs#L55-L119

petrochenkov:
I'm not going to pursue this general mechanism anymore, given the results.
All the useful parts were extracted to https://github.com/rust-lang/rust/pull/92156, https://github.com/rust-lang/rust/pull/92159 and https://github.com/rust-lang/rust/pull/92161.

PR #53693, opened 2018-08-25
Support an explicit annotation for marker traits

scottmcm:
From the tracking issue for rust-lang/rfcs#1268:
> It seems obvious that we should make a `#[marker]` annotation. ~ https://github.com/rust-lang/rust/issues/29864#issuecomment-368959441

This PR allows you to put `#[marker]` on a trait, at which point:
- [x] The trait must not have any items ~~All of the trait's items must have defaults~~
- [x] Any impl of the trait must be empty (not override any items)
- [x] But impls of the trait are allowed to overlap

r? @nikomatsakis

PR #89061, opened 2021-09-18
Allow marker traits as additional traits on trait objects.

beepster4096:
Currently, only auto traits are allowed to be extra traits on trait objects because they are guaranteed to have no functions. However, there is another type of trait that has this property: marker traits. It seems like an oversight that this is not allowed.

This PR fixes this.

Work left to do:
- [x]  See if this change is enough to make it work
- [x]  Change error message

### DIFF
--- a/compiler/rustc_typeck/src/astconv/mod.rs
+++ b/compiler/rustc_typeck/src/astconv/mod.rs
@@ -1297,12 +1297,15 @@ impl<'o, 'tcx> dyn AstConv<'tcx> + 'o {
             }
         }
 
-        // Expand trait aliases recursively and check that only one regular (non-auto) trait
+        // Expand trait aliases recursively and check that only one regular (non-auto or marker) trait
         // is used and no 'maybe' bounds are used.
         let expanded_traits =
             traits::expand_trait_aliases(tcx, bounds.trait_bounds.iter().map(|&(a, b, _)| (a, b)));
-        let (mut auto_traits, regular_traits): (Vec<_>, Vec<_>) =
-            expanded_traits.partition(|i| tcx.trait_is_auto(i.trait_ref().def_id()));
+        let (mut auto_traits, regular_traits): (Vec<_>, Vec<_>) = expanded_traits.partition(|i| {
+            let trait_def = tcx.trait_def(i.trait_ref().def_id());
+
+            trait_def.has_auto_impl || trait_def.is_marker
+        });
         if regular_traits.len() > 1 {
             let first_trait = &regular_traits[0];
             let additional_trait = &regular_traits[1];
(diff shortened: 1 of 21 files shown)


DrMeepster:
Tests are passing, so this is ready for review

oli-obk:
@bors r+

Makes sense to me and impl looks good.

jackh726:
Is this something we want to feature gate? Maybe not because marker traits are unstable anyways...

DrMeepster:
Technically, a crate could use this without feature flags by depending on a crate with a marker trait in it. I doubt this will cause problems however.

Mark-Simulacrum:
I think at least we should cc @rust-lang/lang for awareness. I was going to suggest noting this on the tracking issue for marker traits, but that doesn't seem to exist -- I guess the other thing would be the tracking issue for "multi trait trait objects"? I think @nikomatsakis you were working with someone on that, maybe in wg-traits context?

Mark-Simulacrum:
@bors r-

Can we squash the commits here, please? I was going to do it for you, but it looks like that may be disabled on your fork/this PR?

nikomatsakis:
I agree with cc'ing lang on this =)

cramertj:
What does `is_marker` currently mean here? Is it only specifically tagged traits, or is it all traits that have no methods? I remember discussing similar proposals before, and at the time I was concerned that "marker trait" wasn't a clear pre-existing user-visible term. I also am worried about the idea of making it a breaking change to add a single defaulted method to a trait that previously had no methods.

scottmcm:
Assuming this is based on https://doc.rust-lang.org/nightly/unstable-book/language-features/marker-trait-attr.html (not just the trait being empty) then it seems plausible to me.  The big question I'd have is around marker traits with supertraits -- it might need to be "`#[marker]` traits where all their supertraits are also `#[marker]`" or something.

DrMeepster:
That's a good point I didn't consider. This code doesn't check super traits because auto traits cannot have any.

nikomatsakis:
> What does is_marker currently mean here? 

I believe @cramertj that marker traits currently require a `#[marker]` annotation (@scottmcm made this change in https://github.com/rust-lang/rust/pull/53693).

nikomatsakis:
I'm comfortable overall with this change, but I'm a bit nervous about exposing the word "marker trait" to end users in diagnostics.

JohnCSimon:
> I'm comfortable overall with this change, but I'm a bit nervous about exposing the word "marker trait" to end users in diagnostics.

ping from triage: 
@DrMeepster can you address this comment?
Please adjust the label with `@rustbot ready` when you're ready for another review

JohnCSimon:
ping again from triage:
@DrMeepster 
Please adjust the label with @rustbot ready when you're ready for another review

JohnCSimon:
ping again from triage:
@DrMeepster can you address this comment?

DrMeepster:
Can't seem to figure out how to do this right, and not interested in spending any more time trying.

PR #89068, opened 2021-09-18
Restructure std::rt (part 2)

bjorn3:
A couple more cleanups on top of https://github.com/rust-lang/rust/pull/89011

Blocked on #89011


JohnCSimon:
Triage: 
@bjorn3 returning to author and removing blocked tag because #89011 is merged

joshtriplett:
Would putting `#[inline(always)]` on `Handler::null` generate the same code?

PR #89069, opened 2021-09-18
Use <[T; N]>::map in Sharded instead of SmallVec and unsafe code

bjorn3:
This results in a lot less assembly

### DIFF
--- a/compiler/rustc_data_structures/src/sharded.rs
+++ b/compiler/rustc_data_structures/src/sharded.rs
@@ -1,6 +1,5 @@
 use crate::fx::{FxHashMap, FxHasher};
 use crate::sync::{Lock, LockGuard};
-use smallvec::SmallVec;
 use std::borrow::Borrow;
 use std::collections::hash_map::RawEntryMut;
 use std::hash::{Hash, Hasher};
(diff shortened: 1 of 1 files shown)


Mark-Simulacrum:
Probably doesn't matter too much for perf - I don't think we frequently initialize Sharded - but r=me, seems like a good improvement if perf is good.

bjorn3:
A couple of regressions, mostly improvements.

Mark-Simulacrum:
Looking at a cachegrind diff locally, the regression seems to be caused by noise rather than something genuine -- I'm seeing a bump in the number of instructions used for PartialEq ConstKind not equals, which seems pretty unrelated to this change, and some LLVM functions. Ultimately doesn't seem like this is something we can address and the change does bring a win on both how simple the code is and bootstrap compilation time.

@bors r+

pnkfelix:

The report from when this PR was [first timed](https://perf.rust-lang.org/compare.html?start=6cdd42f9f8dd4e5e5ba0aa816bc4c99ab8b102f9&end=d7e5e4ab57e3fa2aca41f679c8baab3581926670) looks pretty different from the report from [when it landed](https://perf.rust-lang.org/compare.html?start=db1fb85cff63ad5fffe435e17128f99f9e1d970c&end=3bb9eecf07630be796c587a4bba70c49ae6a1bae).

But either way, I'm inclined to apply @Mark-Simulacrum 's [analysis](https://github.com/rust-lang/rust/pull/89069#issuecomment-922955188) above.

PR #89072, opened 2021-09-18
Avoid a couple of Symbol::as_str calls in cg_llvm 

bjorn3:
This should improve performance a tiny bit. Also remove `Symbol::len` and make `SymbolIndex` private.


michaelwoerister (approved):
Thanks, @bjorn3! r=me once rebased.

Issue #89065, opened 2021-09-18
Add `NonZero{I,U}{8,16,32,64,128,size}::{MIN,MAX}`


EFanZh:
Currently, there is no way to obtain a `NonZero*` value without introducing unsafe codes or `unwrap()` calls, providing these constants can make it easier to use these types. For example:

```rust
use std::num::NonZeroU64;

fn foo(x: u64) -> NonZeroU64 {
    x.checked_add(1).and_then(NonZeroU64::new).unwrap_or(NonZeroU64::MAX)
}
```

mjclements:
I think this sounds like a good feature, and I would like to claim it for a first contribution.

LingMan:
Not that I mind the addition of these constants in general, but your motivating example is a bit unfortunate in that it's a slightly obfuscated `saturating_add`. It would be better served by:
```rust
#![feature(nonzero_ops)]
use std::num::NonZeroU64;

pub fn baz(x: u64) -> NonZeroU64 {
    NonZeroU64::new(1).unwrap().saturating_add(x)

    // or `NonZeroU64::new(x.saturating_add(1)).unwrap()` if you need a solution on stable
}
```
Yes, that does contain an `unwrap`, but it's provably unreachable and really the right tool for the job here. (Slightly better codegen too https://rust.godbolt.org/z/Tansn3Y61)

EFanZh:
There are projects that forbid the usage of `unwrap` with Clippy. Using `unwrap` will require explicit `#[allow(...)]` attribute and maybe proper comment, which seems a little too much.

mjclements:
This might just be my head being in a more Ada space, but not having immediate access to the MIN and MAX for a given scalar type feels wrong to me. Obviously with a little forethought you can come up with a totally safe unwrap, but when you're doing something simple like initializing values for a comparison algorithm, it seems like it might be a bit better if you can use a constant which took care of that thinking about safety for you.

LingMan:
@mjclements: Thank you. This:
> but when you're doing something simple like initializing values for a comparison algorithm

is a good motivating example if you ask me.

@EFanZh: Sounds like these projects blanket forbid `unwrap` as a very coarse heuristic for the behavior they actually want to disallow. If this unwanted behavior is "there is an input that causes the program to panic (ignoring OOM)", you could teach Clippy to allow `unwrap` in const contexts. That way you get a compilation error if the panic path ever fires. The example you provided would then look like:

```rust
#![feature(nonzero_ops)]
[...]

PR #89077, opened 2021-09-18
Implemented MIN and MAX constants for the non-zero integer types.#89065

mjclements:
This implements the feature requested in #89065 . I wrote some tests, but it appears that non-stable features aren't supposed to be tested in the normal nonzero.rs test file. Is there a place to add tests for non-stable features?

### DIFF
--- a/library/core/src/num/nonzero.rs
+++ b/library/core/src/num/nonzero.rs
@@ -897,7 +897,7 @@ macro_rules! nonzero_unsigned_is_power_of_two {
 nonzero_unsigned_is_power_of_two! { NonZeroU8 NonZeroU16 NonZeroU32 NonZeroU64 NonZeroU128 NonZeroUsize }
 
 macro_rules! nonzero_min_max {
-    ( $( $min:expr , $Ty: ident($Int: ty); )+ ) => {
+    ( $( $MinVal:expr , $Ty: ident($Int: ty); )+ ) => {
         $(
             impl $Ty {
                 #[unstable(feature = "nonzero_max", issue = "89065")]
--- a/library/core/src/num/nonzero.rs
+++ b/library/core/src/num/nonzero.rs
@@ -896,35 +896,63 @@ macro_rules! nonzero_unsigned_is_power_of_two {
 
 nonzero_unsigned_is_power_of_two! { NonZeroU8 NonZeroU16 NonZeroU32 NonZeroU64 NonZeroU128 NonZeroUsize }
 
-macro_rules! nonzero_min_max {
-    ( $( $MinVal:expr , $Ty: ident($Int: ty); )+ ) => {
+macro_rules! nonzero_constants_signed {
+    ( $( $Ty: ident($Int: ty); )+ ) => {
         $(
             impl $Ty {
                 #[unstable(feature = "nonzero_min_max", issue = "89065")]
                 #[doc = concat!("The maximum value for a`", stringify!($Ty), "` is the same as `", stringify!($Int), "`")]
+                /// # Examples
                 #[doc = concat!("assert_eq!(", stringify!($Ty), "::MAX, ", stringify!($Int), "::MAX);")]
                 pub const MAX : $Ty = $Ty::new(<$Int>::MAX).unwrap() ;
                 #[unstable(feature = "nonzero_min_max", issue = "89065")]
                 #[doc = concat!("The minimum value for a`", stringify!($Ty), "`.")]
                 /// # Examples
-                #[doc = concat!("assert_eq!(", stringify!($Ty), "::MIN, ", stringify!($MinVal), ";")]
-                pub const MIN : $Ty = $Ty::new($MinVal).unwrap();
+                #[doc = concat!("assert_eq!(", stringify!($Ty), "::MIN, ", stringify!($Int), "::MIN;")]
+                pub const MIN : $Ty = $Ty::new(<$Int>::MIN).unwrap();
                 }
         )+
     }
 }
 
-nonzero_min_max! {
-    1 , NonZeroU8(u8);
-    1 , NonZeroU16(u16);
-    1 , NonZeroU32(u32);
-    1 , NonZeroU64(u64);
-    1 , NonZeroU128(u128);
-    1 , NonZeroUsize(usize);
-    i8::MIN , NonZeroI8(i8);
-    i16::MIN , NonZeroI16(i16);
-    i32::MIN  , NonZeroI32(i32);
-    i64::MIN , NonZeroI64(i64);
-    i128::MIN  , NonZeroI128(i128);
-    isize::MIN  , NonZeroIsize(isize);
+nonzero_constants_signed! {
+    NonZeroI8(i8);
+    NonZeroI16(i16);
+    NonZeroI32(i32);
+    NonZeroI64(i64);
+    NonZeroI128(i128);
+    NonZeroIsize(isize);
+}
+
+macro_rules! nonzero_constants_unsigned{
+    ( $( $Ty: ident($Int: ty); )+ ) => {
+        $(
+            impl $Ty {
+                #[unstable(feature = "nonzero_min_max", issue = "89065")]
+                #[doc = concat!("The maximum value for a`", stringify!($Ty), "` is the same as `", stringify!($Int), "`")]
+                #[doc = concat!("assert_eq!(", stringify!($Ty), "::MAX, ", stringify!($Int), "::MAX);")]
+                /// Note, while most integer types are defined for every whole number between MIN and
+                /// MAX, signed non-zero integers are a special case. They have a 'gap' at 0.
+                /// # Examples  
+                pub const MAX : $Ty = $Ty::new(<$Int>::MAX).unwrap() ;
+                #[unstable(feature = "nonzero_min_max", issue = "89065")]
+                #[doc = concat!("The minimum value for a`", stringify!($Ty), "`.")]
+                /// Note, while most integer types are defined for every whole number between MIN and
+                /// MAX, signed non-zero integers are a special case. They have a 'gap' at 0.
+                /// # Examples
+                #[doc = concat!("assert_eq!(", stringify!($Ty), "::MIN, ", stringify!($Int), "::MIN;")]
+                pub const MIN : $Ty = $Ty::new(1).unwrap();
+                }
+        )+
+    }
+}
+
+
+nonzero_constants_unsigned! {
+   NonZeroU8(u8);
+    NonZeroU16(u16);
+    NonZeroU32(u32);
+    NonZeroU64(u64);
+    NonZeroU128(u128);
+    NonZeroUsize(usize);
 }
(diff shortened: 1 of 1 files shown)


library/core/src/num/nonzero.rs
```diff
@@ -895,3 +895,39 @@ macro_rules! nonzero_unsigned_is_power_of_two {
 }
 
 nonzero_unsigned_is_power_of_two! { NonZeroU8 NonZeroU16 NonZeroU32 NonZeroU64 NonZeroU128 NonZeroUsize }
+
+macro_rules! nonzero_min_max {
+    ( $( $MinVal:expr , $Ty: ident($Int: ty); )+ ) => {
+        $(
+            impl $Ty {
+                #[unstable(feature = "nonzero_max", issue = "89065")]
```
LingMan, line 903: Don't see a reason to use two separate feature gates for min and max.

library/core/src/num/nonzero.rs
```diff
@@ -895,3 +895,39 @@ macro_rules! nonzero_unsigned_is_power_of_two {
...
+                #[doc = concat!("The maximum value for a`", stringify!($Ty), "` is the same as `", stringify!($Int), "`")]
+                #[doc = concat!("assert_eq!(", stringify!($Ty), "::MAX, ", stringify!($Int), "::MAX);")]
+                // SAFETY: Since the MAX value, for any supported integer type, is greater than 0, the MAX will always be non-zero.
+                pub const MAX : $Ty = unsafe { $Ty::new_unchecked(<$Int>::MAX) };
```
LingMan, line 907:
```suggestion
                pub const MAX : $Ty = $Ty::new(<$Int>::MAX).unwrap();
```
Since we're in a const context this is guaranteed to get evaluated (and potentially fail if you pass in a zero after all) at compile time. No need for unsafe.

Same for the min case below. Unfortunately `unwrap_or` isn't const yet or we could get rid of the `$MinVal` input entirely.

library/core/src/num/nonzero.rs
```diff
@@ -895,3 +895,39 @@ macro_rules! nonzero_unsigned_is_power_of_two {
...
+                #[unstable(feature = "nonzero_min", issue = "89065")]
+                #[doc = concat!("The minimum value for a`", stringify!($Ty), "`.")]
+                /// # Examples
+                #[doc = concat!("assert_eq!(", stringify!($Ty), "::MIN, ", stringify!($MinVal), ";")]
+                // SAFETY: In the signed case, the minimum integer is negative, and therefore non-zero.
+                // SAFETY:  In the unsignedd case, we use one, which is non-zero.
+                pub const MIN : $Ty = unsafe { $Ty::new_unchecked($MinVal)};
+                }
```
LingMan, line 915: Indentation looks off here. Did the formatter really demand it like that?

ibraheemdev: It's a macro, it should be formatted manually.

library/core/src/num/nonzero.rs
mjclements: Agreed. 591b89dbd58d890f2e13fadbcfc78f590ec5da4b

library/core/src/num/nonzero.rs
mjclements: Do you have specific advice on how to format it? I noticed that macros needed manual formatting. Wasn't sure if I had it right though @ibraheemdev .

library/core/src/num/nonzero.rs
mjclements: I agree! I really wanted to get rid of that $MinVal, but couldn't find a clean way. Fixed the unsafe.
591b89dbd58d890f2e13fadbcfc78f590ec5da4b

library/core/src/num/nonzero.rs
```diff
@@ -895,3 +895,36 @@ macro_rules! nonzero_unsigned_is_power_of_two {
...
+    1 , NonZeroU64(u64);
+    1 , NonZeroU128(u128);
+    1 , NonZeroUsize(usize);
+    i8::MIN , NonZeroI8(i8);
+    i16::MIN , NonZeroI16(i16);
+    i32::MIN  , NonZeroI32(i32);
+    i64::MIN , NonZeroI64(i64);
+    i128::MIN  , NonZeroI128(i128);
+    isize::MIN  , NonZeroIsize(isize);
+}
```
ibraheemdev, line 930:
```suggestion
macro_rules! nonzero_min_max {
    ( $( $Ty:ident($Int:ty) => $min_val:expr )+ ) => {
        $(
            impl $Ty {
                #[unstable(feature = "nonzero_min_max", issue = "89065")]
                #[doc = concat!("The maximum value for a`", stringify!($Ty), "` is the same as `", stringify!($Int), "`")]
                #[doc = concat!("assert_eq!(", stringify!($Ty), "::MAX, ", stringify!($Int), "::MAX);")]
                pub const MAX : $Ty = $Ty::new(<$Int>::MAX).unwrap() ;
                #[unstable(feature = "nonzero_min_max", issue = "89065")]
                #[doc = concat!("The minimum value for a`", stringify!($Ty), "`.")]
                /// # Examples
                #[doc = concat!("assert_eq!(", stringify!($Ty), "::MIN, ", stringify!($MinVal), ";")]
                pub const MIN: $Ty = $Ty::new($min_val).unwrap();
            }
        )+
    }
}

nonzero_min_max! {
    NonZeroU8(u8) => 1;
    NonZeroU16(u16) => 1;
    NonZeroU32(u32) => 1;
    NonZeroU64(u64) => 1;
    NonZeroU128(u128) => 1;
    NonZeroUsize(usize) => 1;
    NonZeroI8(i8) => i8::MIN;
    NonZeroI16(i16) => i16::MIN;
    NonZeroI32(i32) => i32::MIN;
    NonZeroI64(i64) => i64::MIN;
    NonZeroI128(i128) => i128::MIN;
    NonZeroIsize(isize) => isize::MIN;
}
```

ibraheemdev: Formatting is up to you, but I'd probably do something like this.

LingMan:
Mirroring part of a [comment](https://github.com/rust-lang/rust/issues/89065#issuecomment-923238190) from @kpreid so it doesn't get forgotten about:
>    * `NonZeroI*::{MIN,MAX}` are a little weird, because in all other integer cases, all integers between MIN and MAX exist, but in the case of nonzero signed integers, there's also a hole in the middle. This seems an oddity worth thinking about and possibly **documenting.**

I agree that this hole should be called out explicitly in the documentation.

mjclements:
> Mirroring part of a [comment](https://github.com/rust-lang/rust/issues/89065#issuecomment-923238190) from @kpreid so it doesn't get forgotten about:
> 
> > * `NonZeroI*::{MIN,MAX}` are a little weird, because in all other integer cases, all integers between MIN and MAX exist, but in the case of nonzero signed integers, there's also a hole in the middle. This seems an oddity worth thinking about and possibly **documenting.**
> 
> I agree that this hole should be called out explicitly in the documentation.

I think this is resolved.

library/core/src/num/nonzero.rs
```diff
@@ -895,3 +895,63 @@ macro_rules! nonzero_unsigned_is_power_of_two {
 }
 
 nonzero_unsigned_is_power_of_two! { NonZeroU8 NonZeroU16 NonZeroU32 NonZeroU64 NonZeroU128 NonZeroUsize }
+
+macro_rules! nonzero_constants_signed {
+    ( $( $Ty: ident($Int: ty); )+ ) => {
+        $(
+            impl $Ty {
+                #[unstable(feature = "nonzero_min_max", issue = "89065")]
+                #[doc = concat!("The maximum value for a`", stringify!($Ty), "` is the same as `", stringify!($Int), "`")]
```
kpreid, line 904: In <code>for a\`",</code> there should be a space between `a` and <code>`</code>, and the sentence needs a period at the end.

library/core/src/num/nonzero.rs
kpreid: This closing brace, and the analogous one in `nonzero_constants_signed` should be unindented by one level = four spaces. The rest looks fine.

library/core/src/num/nonzero.rs
```diff
@@ -895,3 +895,63 @@ macro_rules! nonzero_unsigned_is_power_of_two {
...
+}
+
+macro_rules! nonzero_constants_unsigned{
+    ( $( $Ty: ident($Int: ty); )+ ) => {
+        $(
+            impl $Ty {
+                #[unstable(feature = "nonzero_min_max", issue = "89065")]
+                #[doc = concat!("The maximum value for a`", stringify!($Ty), "` is the same as `", stringify!($Int), "`")]
+                #[doc = concat!("assert_eq!(", stringify!($Ty), "::MAX, ", stringify!($Int), "::MAX);")]
+                /// Note, while most integer types are defined for every whole number between MIN and
```
kpreid, line 934: I think this should be either `Note that, while` or `Note: while` rather than `Note, while`, but I'm not sure whether others would agree with that grammar nitpick.

kpreid (changes requested):
I agree that the zero case has been mentioned adequately, but since I looked, here's some minor suggestions (all to do with the documentation or formatting).

library/core/src/num/nonzero.rs
```diff
@@ -895,3 +895,63 @@ macro_rules! nonzero_unsigned_is_power_of_two {
...
+
+macro_rules! nonzero_constants_signed {
+    ( $( $Ty: ident($Int: ty); )+ ) => {
+        $(
+            impl $Ty {
+                #[unstable(feature = "nonzero_min_max", issue = "89065")]
+                #[doc = concat!("The maximum value for a`", stringify!($Ty), "` is the same as `", stringify!($Int), "`")]
+                /// # Examples
+                #[doc = concat!("assert_eq!(", stringify!($Ty), "::MAX, ", stringify!($Int), "::MAX);")]
+                pub const MAX : $Ty = $Ty::new(<$Int>::MAX).unwrap() ;
```
LingMan, line 907:
```suggestion
                pub const MAX : $Ty = $Ty::new(<$Int>::MAX).unwrap();
```
Same in the other macro.

library/core/src/num/nonzero.rs
LingMan: Yeah, I'd vote for `Note: While` with a capital `W`.

library/core/src/num/nonzero.rs
```diff
@@ -895,3 +895,75 @@ macro_rules! nonzero_unsigned_is_power_of_two {
...
+                #[doc = concat!("assert_eq!(", stringify!($Ty), "::MIN, ", stringify!($Int), "::MIN);")]
+                /// ```
+                pub const MIN : $Ty = $Ty::new(1).unwrap();
+            }
+        )+
+    }
+}
+
+nonzero_constants_unsigned! {
+   NonZeroU8(u8);
```
joshtriplett, line 963: Indentation mismatch here.

joshtriplett:
It looks like this is failing because it needs to reference these types with a path.

JohnCSimon:
Ping from triage: 
@mjclements Can you please address the build failures?

JohnCSimon:
Ping from triage: 
@mjclements 

@rustbot label: +S-inactive

Issue #85440, opened 2021-05-18
Random test ordering for cargo test


stchris:
**Describe the problem you are trying to solve**
Sometimes tests depend on each other and one of the possible counter measures is to shuffle the order in which tests are run.

**Describe the solution you'd like**
I would like `cargo test` to take a `--random` flag, which would shuffle the test order. Ideally it would generate and print the seed used (a u64 integer). This could be used to debug situations, maybe with a corresponding `--random-seed <seed>` parameter (which would imply `--random`).

**Notes**
I had a look at the codebase and I think I identified the place where the sorting currently happens (https://github.com/rust-lang/cargo/blob/master/src/cargo/ops/cargo_test.rs#L67). I'd be happy to try to contribute a PR.

ehuss:
Ah. I have transferred this to the rust-lang/rust repository where the libtest harness lives.  The code for the test harness lives in the [`library/test`](https://github.com/rust-lang/rust/blob/master/library/test/src/lib.rs) directory.

stchris:
I see. Thank you @ehuss !

smoelius:
FWIW, Go recently acquired such a feature (maybe that inspired this issue?):
https://go-review.googlesource.com/c/go/+/310033/
https://www.reddit.com/r/golang/comments/nktggq/as_of_go_117_youll_be_able_to_shuffle_the/

smoelius:
@stchris Have you tried implementing this?

hanbollar:
hiya ~ @ehuss / @smoelius / @stchris  is this still planned? would be great to use

smoelius:
Hi, @hanbollar. This [is implemented](https://github.com/rust-lang/rust/pull/89082), but only for nightly.

I [run it regularly](https://github.com/trailofbits/test-fuzz/blob/b6e0b1a21ff8b4dd790ad65c46602330ead4b312/.github/workflows/ci.yml#L137-L145), actually.

hanbollar:
> Hi, @hanbollar. This [is implemented](https://github.com/rust-lang/rust/pull/89082), but only for nightly.
> 
> I [run it regularly](https://github.com/trailofbits/test-fuzz/blob/b6e0b1a21ff8b4dd790ad65c46602330ead4b312/.github/workflows/ci.yml#L137-L145), actually.

o perf! thx for the info @smoelius

epage:
[...]

PR #89082, opened 2021-09-18
Implement #85440 (Random test ordering)

smoelius:
This PR adds `--shuffle` and `--shuffle-seed` options to `libtest`. The options are similar to the [`-shuffle` option](https://github.com/golang/go/blob/c894b442d1e5e150ad33fa3ce13dbfab1c037b3a/src/testing/testing.go#L1482-L1499) that was recently added to Go.

Here are the relevant parts of the help message:
```
        --shuffle       Run tests in random order
        --shuffle-seed SEED
                        Run tests in random order; seed the random number
                        generator with SEED
...
By default, the tests are run in alphabetical order. Use --shuffle or set
RUST_TEST_SHUFFLE to run the tests in random order. Pass the generated
"shuffle seed" to --shuffle-seed (or set RUST_TEST_SHUFFLE_SEED) to run the
tests in the same order again. Note that --shuffle and --shuffle-seed do not
affect whether the tests are run in parallel.
```
Is an RFC needed for this?

### DIFF
--- a/library/test/src/cli.rs
+++ b/library/test/src/cli.rs
@@ -339,7 +339,7 @@ fn get_time_options(
 
 fn get_shuffle(matches: &getopts::Matches, allow_unstable: bool) -> OptPartRes<bool> {
     let mut shuffle = unstable_optflag!(matches, allow_unstable, "shuffle");
-    if !shuffle {
+    if !shuffle && allow_unstable {
         shuffle = match env::var("RUST_TEST_SHUFFLE") {
             Ok(val) => &val != "0",
             Err(_) => false,
@@ -364,7 +364,7 @@ fn get_shuffle_seed(matches: &getopts::Matches, allow_unstable: bool) -> OptPart
         None => None,
     };
 
-    if shuffle_seed.is_none() {
+    if shuffle_seed.is_none() && allow_unstable {
         shuffle_seed = match env::var("RUST_TEST_SHUFFLE_SEED") {
             Ok(val) => match val.parse::<u64>() {
                 Ok(n) => Some(n),
--- a/library/test/src/cli.rs
+++ b/library/test/src/cli.rs
@@ -21,6 +21,8 @@ pub struct TestOpts {
     pub nocapture: bool,
     pub color: ColorConfig,
     pub format: OutputFormat,
+    pub shuffle: bool,
+    pub shuffle_seed: Option<u64>,
     pub test_threads: Option<usize>,
     pub skip: Vec<String>,
     pub time_options: Option<TestTimeOptions>,
@@ -138,6 +140,13 @@ fn optgroups() -> getopts::Options {
 
             `CRITICAL_TIME` here means the limit that should not be exceeded by test.
             ",
+        )
+        .optflag("", "shuffle", "Run tests in random order")
+        .optopt(
+            "",
+            "shuffle-seed",
+            "Run tests in random order; seed the random number generator with SEED",
+            "SEED",
         );
     opts
 }
@@ -154,6 +163,12 @@ be passed, which will run all tests matching any of the filters.
 By default, all tests are run in parallel. This can be altered with the
 --test-threads flag or the RUST_TEST_THREADS environment variable when running
 tests (set it to 1).
+
+By default, the tests are run in alphabetical order. Use --shuffle or set
+RUST_TEST_SHUFFLE to run the tests in random order. Pass the generated
+"shuffle seed" to --shuffle-seed (or set RUST_TEST_SHUFFLE_SEED) to run the
+tests in the same order again. Note that --shuffle and --shuffle-seed do not
+affect whether the tests are run in parallel.
 
 All tests have their standard output and standard error captured by default.
 This can be overridden with the --nocapture flag or setting RUST_TEST_NOCAPTURE
@@ -218,6 +233,21 @@ macro_rules! unstable_optflag {
     }};
 }
 
+// Gets the option value and checks if unstable features are enabled.
+macro_rules! unstable_optopt {
+    ($matches:ident, $allow_unstable:ident, $option_name:literal) => {{
+        let opt = $matches.opt_str($option_name);
+        if !$allow_unstable && opt.is_some() {
+            return Err(format!(
+                "The \"{}\" option is only accepted on the nightly compiler with -Z unstable-options",
+                $option_name
+            ));
+        }
+
+        opt
+    }};
+}
+
 // Implementation of `parse_opts` that doesn't care about help message
 // and returns a `Result`.
 fn parse_opts_impl(matches: getopts::Matches) -> OptRes {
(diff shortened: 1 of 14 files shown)


smoelius:
Converting this to a WIP while I figure out this randomness situation.

library/test/src/cli.rs
```diff
@@ -303,6 +322,46 @@ fn get_time_options(
...
+            }
+        },
+        None => None,
+    };
+
+    if shuffle_seed.is_none() {
+        shuffle_seed = match env::var("RUST_TEST_SHUFFLE_SEED") {
+            Ok(val) => match val.parse::<u64>() {
+                Ok(n) => Some(n),
+                Err(_) => panic!("RUST_TEST_SHUFFLE_SEED is `{}`, should be a number.", val),
```
lukaslueg, line 356: Why the `panic!` here instead of returning an Error as above?

smoelius: I was trying to be consistent with `get_concurrency`: https://github.com/rust-lang/rust/blob/0a0688e6f51229d2860af6317adaef378f4b45a5/library/test/src/helpers/concurrency.rs#L9

But I don't have a strong opinion, and I would be happy to change this.

smoelius:
@kennytm I think you can start reviewing this whenever you are ready.

library/test/src/helpers/shuffle.rs
```diff
@@ -0,0 +1,67 @@
...
+    fn new(seed: u64, extra: u64) -> Self {
+        Self { state: seed, extra }
+    }
+
+    fn rand_range(&mut self, range: core::ops::Range<u64>) -> u64 {
+        self.rand_u64() % (range.end - range.start) + range.start
+    }
+
+    fn rand_u64(&mut self) -> u64 {
+        self.state = calculate_hash(&(self.state, self.extra));
```
kennytm, line 57: can we use a proper RNG library like `fastrand` or `oorandom` or `rand`

there is no proof that `state = sip13_hash((state, extra))` is a good RNG.

(and additionally `rand_u64() % n` produces  biased result)

library/test/src/cli.rs
```diff
@@ -303,6 +322,46 @@ fn get_time_options(
     Ok(options)
 }
 
+fn get_shuffle(matches: &getopts::Matches) -> OptPartRes<bool> {
+    let mut shuffle = matches.opt_present("shuffle");
```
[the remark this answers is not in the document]
smoelius: Sorry, I should have noticed that. Fixed.

library/test/src/helpers/shuffle.rs
smoelius: > can we use a proper RNG library like `fastrand` or `oorandom` or `rand`

I tried `rand` and `oorandom` actually. In both cases, I got `can't find crate for core` errors.

My impression (based on [this](https://github.com/rust-lang/rust/blob/0a0688e6f51229d2860af6317adaef378f4b45a5/library/test/Cargo.toml#L10-L11), [this](https://github.com/alexcrichton/cfg-if/blob/59f5b77e6255ce8baf304f74f141a67d42f339c5/Cargo.toml#L21), and [this](https://github.com/rust-lang/getopts/blob/c11eb65097d0dc9fe9ceb351d63c3ac8b5f05e0b/Cargo.toml#L19)) is that a crate has to go out of its way to be a dependency of a builtin library. And none of those crates seems to do that.

Note that `oorandom` barely uses `core`, but it does use it for `Range` ([here](https://hg.sr.ht/~icefox/oorandom/browse/src/lib.rs?rev=tip#L23)).

> there is no proof that `state = sip13_hash((state, extra))` is a good RNG.
> 
> (and additionally `rand_u64() % n` produces biased result)

Your points are very well taken. However, I don't think this random number generator needs to be "strong." Put another way, I think the risk of an attacker trying to manipulate the test ordering is low. I am open to counterarguments though.

(Thanks for doing this, BTW.)

smoelius:
@kennytm Since we seem to have reached a pause, I took the opportunity to do some squashing.

Please let me know if there is anything I can do to make reviewing this PR easier.

library/test/src/cli.rs
```diff
@@ -303,6 +337,46 @@ fn get_time_options(
     Ok(options)
 }
 
+fn get_shuffle(matches: &getopts::Matches, allow_unstable: bool) -> OptPartRes<bool> {
+    let mut shuffle = unstable_optflag!(matches, allow_unstable, "shuffle");
+    if !shuffle {
+        shuffle = match env::var("RUST_TEST_SHUFFLE") {
```
kennytm, line 343: should the env var be also protected behind `allow_unstable`?

smoelius: Yes, I agree. Fixed.

kennytm:
please also update [`/src/doc/rustc/src/tests/index.md`](https://github.com/rust-lang/rust/blob/master/src/doc/rustc/src/tests/index.md) to document the new unstable option.

r=me after that.

smoelius:
@kennytm I added some documentation. Please let me know if it seems inadequate.

Thank you again for all of your trouble.

src/doc/rustc/src/tests/index.md
```diff
@@ -181,6 +181,38 @@ unstable-options` flag. See [tracking issue
...
+
+The random number generator seed that is output can be passed to
+[`--shuffle-seed`](#--shuffle-seed-seed) to run the tests in the same order
+again.
+
+Note that `--shuffle` does not affect whether the tests are run in parallel. To
+run the tests in random order sequentially, use `--shuffle --test-threads 1`.
+
+⚠️ 🚧 This option is [unstable](#unstable-options), and requires the `-Z
+unstable-options` flag.
```
kennytm, line 199:
```suggestion
unstable-options` flag. See [tracking issue
#85440](https://github.com/rust-lang/rust/issues/85440) for more 
information.
```

src/doc/rustc/src/tests/index.md
```diff
@@ -181,6 +181,38 @@ unstable-options` flag. See [tracking issue
...
+the tests in the same order both times.
+
+_SEED_ is any 64-bit unsigned integer, for example, one produced by
+[`--shuffle`](#--shuffle).
+
+This can also be specified with the `RUST_TEST_SHUFFLE_SEED` environment
+variable.
+
+⚠️ 🚧 This option is [unstable](#unstable-options), and requires the `-Z
+unstable-options` flag.
```
kennytm, line 214:
```suggestion
unstable-options` flag. See [tracking issue
#85440](https://github.com/rust-lang/rust/issues/85440) for more 
information.
```

kennytm (approved):
basically I just want a link to a tracking issue. 

or you could file a different issue for tracking if you like 🙃

joshtriplett:
Might be worth looking at lld's   --shuffle-sections for comparison.

smoelius:
> basically I just want a link to a tracking issue.
> 
> or you could file a different issue for tracking if you like upside_down_face

Sorry, I misunderstood how these things work.

I opened a new issue so that I can edit it if I have to.

Thanks again, @kennytm.

Issue #89058, opened 2021-09-17
Drop support for FreeBSD 10 from std


asomers:
Rust's standard library currently uses a FreeBSD 10 compatible ABI.  But FreeBSD 10 went EoL on October 3, 2017.  And now FreeBSD 11 will itself be EoL on September 30, 2021.  Isn't it finally time to drop FreeBSD 10 support and update the standard library's ABI to FreeBSD 12, or at least 11?

nagisa:
Ultimately users of the platform are best positioned to answer this question. There are probably quite few casual users of FreeBSD around here, so somebody who is actively involved in the FreeBSD community would likely be best positioned to provide some guidance here.

asomers:
Well, I'm a FreeBSD developer.  And I can tell you that we're very proud of our backwards-compatibility, but it doesn't require maintaining new tools for old OS versions.  Instead, if a person must use FreeBSD 10, the best way to build new software for it would be to use a snapshot or fork of an old version of the Ports tree.  That would include an older version of rustc, one that was known to work with FreeBSD 10.  In fact, the ports tree itself quickly drops support for old OS versions (it's in the process of dropping support for 11 right now).

cemeyer:
I’m also a FreeBSD and Rust developer and would support dropping 10 support immediately, and 11 support in a few days, when it is officially EOL.

delphij:
I'm also a FreeBSD developer and I would second @cemeyer . FreeBSD 10 support should be gone as it's EoL'ed almost 3 years ago, and FreeBSD 11 support would be ended [really soon](https://www.freebsd.org/security/) too.

At FreeBSD we are committed to provide backward compatibility for applications compiled for older version of FreeBSD releases by maintaining stable system call interfaces; for C runtime we are providing versioned symbols, so legacy binaries would call the old interfaces, which are usually wrappers of new interfaces.  While we are proud that we have provided compatibility shims which worked well for the old binaries, we generally encourage application developers to adopt newer API to take full advantage of the performance improvements and flexibility on newer FreeBSD releases.

laffer1:
Dropping FreeBSD 10 support is likely fine at this point.  I think it's premature with 11, despite the upcoming EOL date.

bsdimp:
[...]

PR #89083, opened 2021-09-18
Upgrade the FreeBSD toolchain to version 12.2

asomers:
FreeBSD 11.4 will be EoL on 30-Sept-2021

### DIFF
--- a/src/ci/docker/host-x86_64/dist-various-2/Dockerfile
+++ b/src/ci/docker/host-x86_64/dist-various-2/Dockerfile
@@ -52,9 +52,9 @@ ENV \
     CFLAGS_x86_64_fortanix_unknown_sgx="-D__ELF__ -isystem/usr/include/x86_64-linux-gnu -mlvi-hardening -mllvm -x86-experimental-lvi-inline-asm-hardening" \
     CXX_x86_64_fortanix_unknown_sgx=clang++-11 \
     CXXFLAGS_x86_64_fortanix_unknown_sgx="-D__ELF__ -isystem/usr/include/x86_64-linux-gnu -mlvi-hardening -mllvm -x86-experimental-lvi-inline-asm-hardening" \
-    AR_i686_unknown_freebsd=i686-unknown-freebsd11-ar \
-    CC_i686_unknown_freebsd=i686-unknown-freebsd11-clang \
-    CXX_i686_unknown_freebsd=i686-unknown-freebsd11-clang++ \
+    AR_i686_unknown_freebsd=i686-unknown-freebsd12-ar \
+    CC_i686_unknown_freebsd=i686-unknown-freebsd12-clang \
+    CXX_i686_unknown_freebsd=i686-unknown-freebsd12-clang++ \
     CC=gcc-8 \
     CXX=g++-8
 
--- a/src/ci/docker/host-x86_64/dist-x86_64-freebsd/Dockerfile
+++ b/src/ci/docker/host-x86_64/dist-x86_64-freebsd/Dockerfile
@@ -27,9 +27,9 @@ COPY scripts/cmake.sh /scripts/
 RUN /scripts/cmake.sh
 
 ENV \
-    AR_x86_64_unknown_freebsd=x86_64-unknown-freebsd11-ar \
-    CC_x86_64_unknown_freebsd=x86_64-unknown-freebsd11-clang \
-    CXX_x86_64_unknown_freebsd=x86_64-unknown-freebsd11-clang++
+    AR_x86_64_unknown_freebsd=x86_64-unknown-freebsd12-ar \
+    CC_x86_64_unknown_freebsd=x86_64-unknown-freebsd12-clang \
+    CXX_x86_64_unknown_freebsd=x86_64-unknown-freebsd12-clang++
 
 ENV HOSTS=x86_64-unknown-freebsd
 
--- a/src/ci/docker/scripts/freebsd-toolchain.sh
+++ b/src/ci/docker/scripts/freebsd-toolchain.sh
@@ -5,8 +5,8 @@ set -eux
 
 arch=$1
 binutils_version=2.25.1
-freebsd_version=11.4
-triple=$arch-unknown-freebsd11
+freebsd_version=12.2
+triple=$arch-unknown-freebsd12
 sysroot=/usr/local/$triple
 
 hide_output() {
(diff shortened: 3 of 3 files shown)


asomers:
Somebody with more privilege than I must update the mirror at https://ci-mirrors.rust-lang.org/rustc/ to include https://download.freebsd.org/ftp/releases/amd64/12.2-RELEASE/base.txz and https://download.freebsd.org/ftp/releases/i386/12.2-RELEASE/base.txz .

Mark-Simulacrum:
Hey! So I tried to read the discussion in https://github.com/rust-lang/rust/issues/89058, but I admit I am not entirely following the picture here or what this change implies for Rust. Currently, it seems like our [platform support docs](https://doc.rust-lang.org/nightly/rustc/platform-support.html) leave out the FreeBSD version that we're targeting, which seems like an oversight -- presumably, we should provide something similar to the Linux builders, though I am not familiar with how FreeBSD's versioning works (e.g., do we want to specify the kernel version? libc version? Is that the same?)

We don't currently have a set of target maintainers on file for FreeBSD -- perhaps @asomers you and @bdrewery might be interested in signing up for that? There's some more details in what this entails in the target tier policy; I think FreeBSD on x86_64 is at [Tier 2 with host tools right now](https://doc.rust-lang.org/nightly/rustc/target-tier-policy.html#tier-2-with-host-tools).

If I'm following https://github.com/rust-lang/rust/issues/89058#issuecomment-926089544 correctly, then it sounds like accepting the bump to the compiler environment used on our FreeBSD builders will not affect the compatibility of rustc or std with older versions of FreeBSD (i.e., the binaries we ship will still run out of the box on FreeBSD 11 or 10). Is that right? This would be different from how our compatibility story works on e.g. linux targets, where we need to build in an environment with an old glibc toolchain set in order to maintain compatibility.

asomers:
> Hey! So I tried to read the discussion in #89058, but I admit I am not entirely following the picture here or what this change implies for Rust. Currently, it seems like our [platform support docs](https://doc.rust-lang.org/nightly/rustc/platform-support.html) leave out the FreeBSD version that we're targeting, which seems like an oversight -- presumably, we should provide something similar to the Linux builders, though I am not familiar with how FreeBSD's versioning works (e.g., do we want to specify the kernel version? libc version? Is that the same?)

Kernel and libc version are the same, and there's little point to supporting a specific minor version.  We should just specify the minimum supported major version.

> 
> We don't currently have a set of target maintainers on file for FreeBSD -- perhaps @asomers you and @bdrewery might be interested in signing up for that? There's some more details in what this entails in the target tier policy; I think FreeBSD on x86_64 is at [Tier 2 with host tools right now](https://doc.rust-lang.org/nightly/rustc/target-tier-policy.html#tier-2-with-host-tools).

I can ask around.  What are the responsibilities of a target maintainer?

> 
> If I'm following [#89058 (comment)](https://github.com/rust-lang/rust/issues/89058#issuecomment-926089544) correctly, then it sounds like accepting the bump to the compiler environment used on our FreeBSD builders will not affect the compatibility of rustc or std with older versions of FreeBSD (i.e., the binaries we ship will still run out of the box on FreeBSD 11 or 10). Is that right? This would be different from how our compatibility story works on e.g. linux targets, where we need to build in an environment with an old glibc toolchain set in order to maintain compatibility.

This PR alone definitely won't affect the usability of libc on FreeBSD 11.  I'm not sure about the toolchain itself; I don't know enough about how rustc is built.

Mark-Simulacrum:
> I can ask around. What are the responsibilities of a target maintainer?

There's no concrete list, but generally it's doing the educational work with rustc/std maintainers for PRs like this one or updates to the standard library, as well as keeping an eye on Rust's story on that target (e.g., if there is some new tooling or whatever, letting us know or doing the work to add support in rustc/std). It's also totally fine for these folks to step back, but current policy is that if they're needed and not available we may drop the target.

Ideally these folks are also regularly running the test suite (e.g., once a month or so) on tier-2-with-host-tools or lower targets, primarily to ensure that the target still works OK and to file issues for and fix any broken tests.

> This PR alone definitely won't affect the usability of libc on FreeBSD 11. I'm not sure about the toolchain itself; I don't know enough about how rustc is built.

Can you say more about what you would need to know? I'm happy to clarify things or help investigate, but without more information I am hesitant to land this patch since it seems plausibly likely to break users. (Of course, some indications are that *not* doing this also will...)

asomers:
I need to know whether the toolchain's build process links any symbols according to the build host's installed libc, or if it just uses the symbol versions defined in rust's libc.  I'm not sure what that's called in compiler-speak.  I suppose I could just try it and find out.  If I build the toolchain locally, is there an easy way to run its test suite on a host other than the one that built it?
Oh, and where are target maintainers recorded?  I can't find it in the repo anywhere.

sanxiyn:
> Oh, and where are target maintainers recorded?

Here is an example: https://github.com/rust-lang/rust/blob/master/src/doc/rustc/src/platform-support/x86_64-unknown-none.md.

JohnCSimon:
Ping from triage:
@asomers can you please post the of this PR?

asomers:
@JohnCSimon still waiting for somebody to upload the FreeBSD release files.  They should be named something like
https://ci-mirrors.rust-lang.org/rustc/2020-08-09-freebsd-{amd64,i386}-12.2-RELEASE-base.txz .  I don't have the permission to do it myself.  However, this PR has lingered so long that release 12.3 is now expected in a matter of days.

Mark-Simulacrum:
I'm happy to take care of uploading artifacts, but I don't think we've yet resolved the questions around what the implications for this change are for users, so we're not quite at the stage where that would unblock us from what I can tell.

> I need to know whether the toolchain's build process links any symbols according to the build host's installed libc, or if it just uses the symbol versions defined in rust's libc. I'm not sure what that's called in compiler-speak. I suppose I could just try it and find out. If I build the toolchain locally, is there an easy way to run its test suite on a host other than the one that built it?

There's not a great way to run tests on a different host, though I suppose it should mostly work to copy the directory (so long as all paths are the same) as a whole to a different host. That would need the target/host triples to match though.

nikic:
As FreeBSD 11 is EOL for about half a year now, and FreeBSD is a tier 2 target, I think we can ahead here even if it does break rustc on FreeBSD completely.

JohnCSimon:
> As FreeBSD 11 is EOL for about half a year now, and FreeBSD is a tier 2 target, I think we can go ahead here even if it does break rustc on FreeBSD 11 completely.

triage: assigning to reviewer.

Mark-Simulacrum:
Yeah, it's probably OK to move ahead with this. Can someone provide links to the right files to upload, and then I can do that and approve this PR?

asomers:
@Mark-Simulacrum use https://download.freebsd.org/ftp/releases/amd64/12.3-RELEASE/base.txz and https://download.freebsd.org/ftp/releases/i386/12.3-RELEASE/base.txz .

asomers:
ping @Mark-Simulacrum could you please upload those files?

Mark-Simulacrum:
https://ci-mirrors.rust-lang.org/rustc/2022-05-06-freebsd-12.3-amd64-base.txz and https://ci-mirrors.rust-lang.org/rustc/2022-05-06-freebsd-12.3-i386-base.txz

For the future, @rustbot ready will update labels so that PRs get in reviewers' queues and folks can take a look at them more quickly.

Dylan-DPC:
Closing this in favour of #97944

PR #89090, opened 2021-09-19
Lint bare traits in AstConv.

cjgillot:
Removing the lint from lowering allows to:
- make lowering querification easier;
- have the lint implementation in only one place.

r? @estebank

### DIFF
--- a/compiler/rustc_typeck/src/astconv/mod.rs
+++ b/compiler/rustc_typeck/src/astconv/mod.rs
@@ -13,6 +13,7 @@ use crate::errors::{
 };
 use crate::middle::resolve_lifetime as rl;
 use crate::require_c_abi_if_c_variadic;
+use rustc_ast::TraitObjectSyntax;
 use rustc_data_structures::fx::{FxHashMap, FxHashSet};
 use rustc_errors::{struct_span_err, Applicability, ErrorReported, FatalError};
 use rustc_hir as hir;
@@ -24,7 +25,8 @@ use rustc_hir::{GenericArg, GenericArgs};
 use rustc_middle::ty::subst::{self, GenericArgKind, InternalSubsts, Subst, SubstsRef};
 use rustc_middle::ty::GenericParamDefKind;
 use rustc_middle::ty::{self, Const, DefIdTree, Ty, TyCtxt, TypeFoldable};
-use rustc_session::lint::builtin::AMBIGUOUS_ASSOCIATED_ITEMS;
+use rustc_session::lint::builtin::{AMBIGUOUS_ASSOCIATED_ITEMS, BARE_TRAIT_OBJECTS};
+use rustc_span::edition::Edition;
 use rustc_span::lev_distance::find_best_match_for_name;
 use rustc_span::symbol::{Ident, Symbol};
 use rustc_span::{Span, DUMMY_SP};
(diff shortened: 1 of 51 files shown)


src/test/ui/did_you_mean/bad-assoc-ty.stderr
```diff
@@ -99,14 +99,24 @@ error[E0223]: ambiguous associated type
 LL | type G = dyn 'static + (Send)::AssocTy;
    |          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ help: use fully-qualified syntax: `<(dyn Send + 'static) as Trait>::AssocTy`
 
+warning: trait objects without an explicit `dyn` are deprecated
+  --> $DIR/bad-assoc-ty.rs:33:10
+   |
+LL | type H = Fn(u8) -> (u8)::Output;
+   |          ^^^^^^ help: use `dyn`: `<dyn Fn(u8)>`
```
cjgillot, line 106: This one makes me uncomfortable, but I could not find how to fix it. Suggestions?

compiler/rustc_typeck/src/astconv/mod.rs
```diff
@@ -2609,4 +2618,49 @@ impl<'o, 'tcx> dyn AstConv<'tcx> + 'o {
...
+                err.span_suggestion_verbose(self_ty.span, label, sugg, app).emit();
+            } else {
+                let msg = "trait objects without an explicit `dyn` are deprecated";
+                tcx.struct_span_lint_hir(
+                    BARE_TRAIT_OBJECTS,
+                    self_ty.hir_id,
+                    self_ty.span,
+                    |lint| {
+                        let mut db = lint.build(msg);
+                        db.span_suggestion(self_ty.span, "use `dyn`", sugg, app);
```
estebank, line 2659: Can we make this a `multipart_suggestion`? We previously only highlighted the change to add `dyn` in most cases, now we underline the whole type, which is fine, but the more targeted spans aid in understanding what the change is supposed to be. Making that change requires instead of dealing with a string for the suggestion to deal with a vec of changes.

src/test/ui/did_you_mean/bad-assoc-ty.stderr
estebank: This is outright incorrect, right? The whole type is `<dyn Fn(u8) -> (u8)::Output>`, if I'm reading this correctly. I think we can special case `Fn` traits in this format by looking at the `poly_trait_ref`.

It's also interesting that the error below is already making some suggestions, so it might make sense to only emit this one if E0233 isn't (but this might not be needed in this PR).

src/test/ui/dyn-keyword/dyn-2018-edition-lint.rs
```diff
@@ -7,8 +7,6 @@ fn function(x: &SomeTrait, y: Box<SomeTrait>) {
     //~| ERROR trait objects without an explicit `dyn` are deprecated
     //~| WARN this is accepted in the current edition
     let _x: &SomeTrait = todo!();
-    //~^ ERROR trait objects without an explicit `dyn` are deprecated
-    //~| WARN this is accepted in the current edition
```
estebank, line 11: These shouldn't go away, right?

estebank (commented):
The reasoning behind the change makes sense, but I am concerned at some of the places we are silencing these errors, particularly in binding types (`let x: &Trait = todo!()` isn't being linted anymore).

petrochenkov:
>I am concerned at some of the places we are silencing these errors, particularly in binding types 

On 2021 edition bare traits are a hard error, so we must always report them regardless of their positions.
The 2021 edition error should ideally be moved to resolve (https://github.com/rust-lang/rust/pull/83213#issuecomment-800603507), it's simpler and more reliable to report it that way.

cjgillot:
@estebank, I introduced the multipart suggestions. Do you still have unresolved remarks on this PR?

PR #89098, opened 2021-09-19
Fix generics where bounds order

### LINKED ISSUE #88809: PartialOrd docs for generic tuples show the where clauses in a seemingly random order
On https://doc.rust-lang.org/std/cmp/trait.PartialOrd.html:

The docs for PartialOrd on tuples show the `where` clauses in a weird order, obscuring the fact that the _last_ element of the tuple is allowed to be unsized.

```rust
impl<A, B, C, D, E> PartialOrd<(A, B, C, D, E)> for (A, B, C, D, E) where
    C: PartialOrd<C> + PartialEq<C>,
    E: PartialOrd<E> + PartialEq<E> + ?Sized,
    B: PartialOrd<B> + PartialEq<B>,
    A: PartialOrd<A> + PartialEq<A>,
    D: PartialOrd<D> + PartialEq<D>, 
```

GuillaumeGomez:
Fixes #88809.

Like said on the above issue, the issue is that we were expecting `Symbol` comparisons to be string-based but they are integer-based (because `Symbol` is an integer), messing up the bounds order. To fix it, I simply stored into a `FxIndexMap` instead.

r? @jyn514

### DIFF
--- a/src/librustdoc/clean/simplify.rs
+++ b/src/librustdoc/clean/simplify.rs
@@ -11,8 +11,7 @@
 //! This module attempts to reconstruct the original where and/or parameter
 //! bounds by special casing scenarios such as these. Fun!
 
-use std::collections::BTreeMap;
-
+use rustc_data_structures::fx::FxIndexMap;
 use rustc_hir::def_id::DefId;
 use rustc_middle::ty;
 use rustc_span::Symbol;
@@ -23,8 +22,11 @@ use crate::clean::WherePredicate as WP;
 use crate::core::DocContext;
 
 crate fn where_clauses(cx: &DocContext<'_>, clauses: Vec<WP>) -> Vec<WP> {
-    // First, partition the where clause into its separate components
-    let mut params: BTreeMap<_, (Vec<_>, Vec<_>)> = BTreeMap::new();
+    // First, partition the where clause into its separate components.
+    //
+    // We use `FxIndexMap` so that the insertion order is preserved to prevent messing up to
+    // the order of the generated bounds.
+    let mut params: FxIndexMap<Symbol, (Vec<_>, Vec<_>)> = FxIndexMap::default();
     let mut lifetimes = Vec::new();
     let mut equalities = Vec::new();
     let mut tybounds = Vec::new();


src/librustdoc/clean/simplify.rs
```diff
@@ -24,7 +24,10 @@ use crate::core::DocContext;
 
 crate fn where_clauses(cx: &DocContext<'_>, clauses: Vec<WP>) -> Vec<WP> {
     // First, partition the where clause into its separate components
-    let mut params: BTreeMap<_, (Vec<_>, Vec<_>)> = BTreeMap::new();
+    let mut params: BTreeMap<Symbol, (Vec<_>, Vec<_>)> = BTreeMap::new();
+    // The BTreeMap cannot keep items sorted simply because the `PartialOrd`implementation on
+    // `Symbol` isn't string based but integer based!
```
Mark-Simulacrum, line 29: If the goal is to keep these insert-ordered, it seems like the right thing to do is switch to a FxIndexMap here.

I'm not sure why *sorting* the symbols was ever intended; that doesn't seem like it would have kept them in the original order even if they sorted by string (and not integer).

Mark-Simulacrum: Seems like this was done here - https://github.com/rust-lang/rust/commit/421e6fc24d2e5bfc59746d2662531fb22f7e0720 - presumably with the goal of deterministic output. I'm not actually sure that was achieved anyway (since Symbol isn't *really* ordered), but regardless, it seems like an FxIndexMap should do the trick.

GuillaumeGomez: Nice, thanks!

GuillaumeGomez:
I implemented @Mark-Simulacrum's suggestion.

camelid:
@GuillaumeGomez This bug seems to be cross-crate-export related, so you'll need to re-export a type implementing a trait to reproduce this. I came up with a repro that should work as a `src/test/rustdoc/` test (with an auxiliary file) if you add the appropriate `@has` checks:

```rust
// In lib1.rs

pub struct Struct<A, B, C>(A, B, C);

// NOTE: would it make sense to define `struct B`, `struct A`, `struct C`, in that order,
// to make sure the symbol indices are in a random order?

pub trait Trait<Other> {}

impl<A, B, C> Trait<Struct<A, B, C>> for Struct<A, B, C>
where
    A: Trait<A>,
    B: Trait<B>,
    C: Trait<C>,
{
}
```

```rust
// In lib2.rs

// aux-build:lib1.rs

pub use lib1::*;

// Add `@has` checks here
```

GuillaumeGomez:
@camelid: Just realized that we can't check that in `test/rustdoc` because the `implementors` content is generated through HTML.

camelid:
> @camelid: Just realized that we can't check that in `test/rustdoc` because the `implementors` content is generated through HTML.

Ah, that's unfortunate. I think it might not have been possible anyway because I don't know if there's a way to check ordering of elements with `@has`.

camelid:
> @camelid: Just realized that we can't check that in `test/rustdoc` because the `implementors` content is generated through HTML.

Actually, I don't think it is. You can check the implementors section of `lib2::Trait` or the implementations of `lib2::Struct` – they both work with JS disabled AFAICT. Both the trait and the struct are re-exported, so if you use the re-exported versions, it should work.

However, there's still the challenge of checking the order of the HTML elements in the `where` clause. Is there a way to do that?

GuillaumeGomez:
We could add this check with a better XPath support:

```
// @has	- '//*[@id="implementors-list"]//span[@class="where fmt-newline"]/text()[1]' 'B:'
```

Unfortunately, it's not supported:

```console
Traceback (most recent call last):
  File "/usr/lib/python3.9/xml/etree/ElementPath.py", line 354, in iterfind
    selector = _cache[cache_key]
KeyError: ('.//*[@id="implementors-list"]//span[@class="where fmt-newline"]/text()[1]',)

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/home/imperio/rust/rust/src/etc/htmldocck.py", line 490, in <module>
    check(sys.argv[1], get_commands(sys.argv[2]))
  File "/home/imperio/rust/rust/src/etc/htmldocck.py", line 482, in check
    check_command(c, cache)
  File "/home/imperio/rust/rust/src/etc/htmldocck.py", line 439, in check_command
    ret = check_tree_text(cache.get_tree(c.args[0]), pat, c.args[2], regexp)
  File "/home/imperio/rust/rust/src/etc/htmldocck.py", line 370, in check_tree_text
    for e in tree.findall(path):
  File "/usr/lib/python3.9/xml/etree/ElementPath.py", line 395, in findall
    return list(iterfind(elem, path, namespaces))
  File "/usr/lib/python3.9/xml/etree/ElementPath.py", line 368, in iterfind
    selector.append(ops[token[0]](next, token))
KeyError: '()'
```

GuillaumeGomez:
The python module we use has been deprecated a while ago. We should switch to another parser like `BeautifulSoup` which supports XPath entirely, that would make such checks possible.

What do you think @jyn514 ? If we agree on that, let's approve this one, open an issue to add a test whenever it's possible and I'll send a PR to switch the HTML parser we use.

PR #89100, opened 2021-09-19
resolve: Use `NameBinding` for local variables and generic parameters

petrochenkov:
`NameBinding` is a structure used for representing any name introduction (an item, or import, or even a built-in).
Except that local variables and generic parameters weren't represented as `NameBinding`s, for this reason they requires separate paths in name resolution code in several places.

This PR introduces `NameBinding`s for local variables as well and simplifies all the code working with them leaving only the `NameBinding` paths.

### DIFF
--- a/compiler/rustc_resolve/src/late.rs
+++ b/compiler/rustc_resolve/src/late.rs
@@ -7,7 +7,7 @@
 
 use RibKind::*;
 
-use crate::{path_names_to_string, BindingError, CrateLint, LexicalScopeBinding};
+use crate::{path_names_to_string, BindingError, CrateLint, NameBinding, ToNameBinding};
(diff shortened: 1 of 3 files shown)


ehuss:
@bors retry

test-various hung after downloading the docker image, no clear reason.

Mark-Simulacrum:
This'll need some manual checking locally to verify the regression given the potential environment changes to the collector, I'll try to do that.

petrochenkov:
Wow, not something that I'd expect.
I suggest reverting, I'll try to investigate this but not soon.

Mark-Simulacrum:
Posted a revert https://github.com/rust-lang/rust/pull/90130.

PR #89101, opened 2021-09-19
Support incremental in compiletest for non-incremental modes.

ehuss:
This adds first-class support for using incremental builds in non-incremental-mode tests.  These tests previously manually passed `-C incremental=tmp/foo` which resulted in reusing the same tmp folder between runs.  This means that these tests could fail whenever the on-disk incremental format changed (such as when updating one's local source tree).  This changes it so that these tests can pass a `// incremental-build` header which instructs compiletest to create a set aside a dedicated incremental directory which will be cleared before the test starts to ensure it has a clean slate.

### DIFF
--- a/src/tools/compiletest/src/header.rs
+++ b/src/tools/compiletest/src/header.rs
@@ -113,6 +113,21 @@ pub struct TestProps {
     // testing harness and used when generating compilation
     // arguments. (In particular, it propagates to the aux-builds.)
     pub incremental_dir: Option<PathBuf>,
+    // If `true`, this test will use incremental compilation.
+    //
+    // This can be set manually with the `incremental` header, or implicitly
+    // by being a part of an incremental mode test. Using the `incremental`
+    // header should be avoided if possible; using an incremental mode test is
+    // preferred. Incremental mode tests support multiple passes, which can
+    // verify that the incremental cache can be loaded properly after being
+    // created. Just setting the header will only verify the behavior with
+    // creating an incremental cache, but doesn't check that it is created
+    // correctly.
+    //
+    // Compiletest will create the incremental directory, and ensure it is
+    // empty before the test starts. Incremental mode tests will reuse the
+    // incremental directory between passes in the same test.
+    pub incremental: bool,
     // How far should the test proceed while still passing.
     pass_mode: Option<PassMode>,
     // Ignore `--pass` overrides from the command line for this test.
@@ -163,6 +178,7 @@ impl TestProps {
             pretty_compare_only: false,
             forbid_output: vec![],
             incremental_dir: None,
+            incremental: false,
             pass_mode: None,
             fail_mode: None,
             ignore_pass: false,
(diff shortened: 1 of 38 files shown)


ehuss:
There was prior discussion about this at https://rust-lang.zulipchat.com/#narrow/stream/131828-t-compiler/topic/Incremental.20ICE.20running.20rustc.20tests.20on.20origin.2Fmaster/near/251211217

It was suggested that these tests could be moved to the incremental directory, but I don't think that is possible because these are codegen tests which are tested differently.  I'm not familiar with these tests, so maybe they can be moved, so let me know if an alternate solution would be preferred.

cc @michaelwoerister

src/tools/compiletest/src/header.rs
```diff
@@ -113,6 +113,8 @@ pub struct TestProps {
     // testing harness and used when generating compilation
     // arguments. (In particular, it propagates to the aux-builds.)
     pub incremental_dir: Option<PathBuf>,
+    // If `true`, this test will use incremental compilation.
```
Mark-Simulacrum, line 116: Can we add some docs here (since this is basically where this is documented) about what 'scenario' of incr comp the test will be run under? In particular, I think we will ensure that the test runs each time starting with a clean incr. comp. dir and then any revisions will continue in the same directory (i.e., become "incr-patched" scenarios effectively).

Mark-Simulacrum:
I think we should also consider renaming the flag to "incremental" or "incremental-mode" -- I want to avoid accidentally reading it as similar to 'build-pass' or otherwise tying it specifically to *build* in that sense.

r=me with the docs and a possible rename, happy to leave it up to you which to choose -- or if you feel strongly about -build, that's also OK.

src/tools/compiletest/src/header.rs
ehuss: Of course!  I was intending to update the [rustc dev guide docs](https://rustc-dev-guide.rust-lang.org/tests/adding.html#other-header-commands) after this merged, but I understand that some people will only read one source or the other.

ehuss:
Yea, I can see how that can be confusing with `-build`.  I originally used just `incremental`, but then I noticed that a single English word can be accidentally picked up by a comment like [this](https://github.com/rust-lang/rust/blob/bf642323d621dcefeef1d8ab4711aae36e357615/src/test/codegen-units/partitioning/extern-generic.rs#L2-L3).

I went ahead and renamed it back to just `incremental`, since I think `incremental-mode` would be confusing with true incremental mode tests that support passes. I think the risk of accidentally making a test incremental is low (the existing examples all use `-C incremental`), and the consequences are low (making a test incremental usually shouldn't interfere with it).

I kinda wish the header parsing was a little more strict so that risk of accidental matches wasn't there.

Feel free to r- if that doesn't sound good, or if the comment I added isn't correct or clear.

Mark-Simulacrum:
Sounds great, thanks! Yeah, our header parsing story is not great :)

PR #89102, opened 2021-09-19
[WIP] Use `DefPath` instead of `Cache.stack` for `Cache.paths`

camelid:
All empty or anonymous segments are skipped.

After this, I'll be able to remove `Cache.stack`.

r? @ghost
cc @jyn514

### DIFF
--- a/src/librustdoc/formats/cache.rs
+++ b/src/librustdoc/formats/cache.rs
@@ -2,6 +2,7 @@ use std::mem;
 
 use rustc_data_structures::fx::{FxHashMap, FxHashSet};
 use rustc_hir::def_id::{CrateNum, DefId, CRATE_DEF_INDEX};
+use rustc_hir::definitions::DefPathDataName;
 use rustc_middle::middle::privacy::AccessLevels;
 use rustc_middle::ty::TyCtxt;
 use rustc_span::symbol::sym;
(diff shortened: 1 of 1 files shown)


camelid:
I was having weird linker errors with some of (or all of?) the rustdoc `auxiliary` tests. I'm not sure what's going on, so let's see if they happen on CI.

camelid:
CI did indeed fail, but without linker errors...

jyn514:
Hi @camelid, are you planning to follow-up on this?

camelid:
> Hi @camelid, are you planning to follow-up on this?

I tried to recently but got stuck trying to get the user-visible name for proc macro derives.

jyn514:
Ah right, I remember seeing that. I glanced around rustc_expand briefly - I found `parse_macro_name_and_helper_attrs` which looks vaguely right, maybe you could use that by getting the attributes for the item if you know it's a proc-macro?

camelid:
> Ah right, I remember seeing that. I glanced around rustc_expand briefly - I found `parse_macro_name_and_helper_attrs` which looks vaguely right, maybe you could use that by getting the attributes for the item if you know it's a proc-macro?

Hmm, okay, that looks potentially promising.

camelid:
I'd like to get back to this at some point, but I'm going to close it for now so it doesn't just sit getting stale.

PR #89103, opened 2021-09-19
Migrate in-tree crates to 2021

### LINKED ISSUE #88637: Switch compiler to Rust 2021
This is blocked on the beta/bootstrap compiler being updated to 1.56.

### LINKED ISSUE #89074: Switch rustdoc to Rust 2021

Mark-Simulacrum:
This replaces #89075 (cherry picking some of the commits from there), and closes #88637 and fixes #89074.

It excludes a migration of the library crates for now (see tidy diff) because we have some pending bugs around macro spans to fix there.

I instrumented bootstrap during the migration to make sure all crates moved from 2018 to 2021 had the compatibility warnings applied first.

Originally, the intent was to support cargo fix --edition within bootstrap, but this proved fairly difficult to pull off. We'd need to architect the check functionality to support running cargo check and cargo fix within the same x.py invocation, and only resetting sysroots on check. Further, it was found that cargo fix doesn't behave too well with "not quite workspaces", such as Clippy which has several crates. Bootstrap runs with --manifest-path ... for all the tools, and this makes cargo fix only attempt migration for that crate. We can't use e.g. --workspace due to needing to maintain sysroots for different phases of compilation appropriately.

It is recommended to skip the mass migration of Cargo.toml's to 2021 for review purposes; you can also use `git diff d6cd2c6c877110748296760aefddc21a0ea1d316 -I'^edition = .20...$'` to ignore the edition = 2018/21 lines in the diff.

### DIFF
--- a/compiler/rustc_codegen_cranelift/src/bin/cg_clif_build_sysroot.rs
+++ b/compiler/rustc_codegen_cranelift/src/bin/cg_clif_build_sysroot.rs
@@ -8,7 +8,6 @@
 
 #![feature(rustc_private)]
 
-extern crate rustc_data_structures;
 extern crate rustc_driver;
 extern crate rustc_interface;
 extern crate rustc_session;
--- a/compiler/rustc_interface/src/util.rs
+++ b/compiler/rustc_interface/src/util.rs
@@ -125,6 +125,7 @@ pub fn scoped_thread<F: FnOnce() -> R + Send, R: Send>(cfg: thread::Builder, f: 
     let result_ptr = Ptr(&mut result as *mut _ as *mut ());
 
     let thread = cfg.spawn(move || {
+        let _ = (&run, &result_ptr);
         let run = unsafe { (*(run.0 as *mut Option<F>)).take().unwrap() };
         let result = unsafe { &mut *(result_ptr.0 as *mut Option<R>) };
         *result = Some(run());
--- a/compiler/rustc_typeck/src/astconv/mod.rs
+++ b/compiler/rustc_typeck/src/astconv/mod.rs
@@ -2417,13 +2417,14 @@ impl<'o, 'tcx> dyn AstConv<'tcx> + 'o {
         let substs = InternalSubsts::for_item(tcx, def_id, |param, _| {
             if let Some(i) = (param.index as usize).checked_sub(generics.parent_count) {
                 // Our own parameters are the resolved lifetimes.
-                match param.kind {
-                    GenericParamDefKind::Lifetime
-                        if let hir::GenericArg::Lifetime(lifetime) = &lifetimes[i] =>
-                    {
+                if let GenericParamDefKind::Lifetime = param.kind {
+                    if let hir::GenericArg::Lifetime(lifetime) = &lifetimes[i] {
                         self.ast_region_to_region(lifetime, None).into()
+                    } else {
+                        bug!()
                     }
-                    _ => bug!(),
+                } else {
+                    bug!()
                 }
             } else {
                 match param.kind {
--- a/src/tools/tidy/src/edition.rs
+++ b/src/tools/tidy/src/edition.rs
@@ -1,10 +1,15 @@
-//! Tidy check to ensure that crate `edition` is '2018'
+//! Tidy check to ensure that crate `edition` is '2018' or '2021'.
 
 use std::path::Path;
 
 fn is_edition_2018(mut line: &str) -> bool {
     line = line.trim();
-    line == "edition = \"2018\"" || line == "edition = \'2018\'"
+    line == "edition = \"2018\""
+}
+
+fn is_edition_2021(mut line: &str) -> bool {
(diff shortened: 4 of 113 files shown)


Mark-Simulacrum:
The drop time changes which we're accepting as part of this migration *could* be perf sensitive, but I believe it to be fairly unlikely -- they're mostly not in hot code, and moving the drop slightly shouldn't change too much. Closure sizes are also unlikely to be that important. That said, we should check: @bors try @rust-timer queue

compiler/rustc_interface/src/util.rs
```diff
@@ -125,6 +125,7 @@ pub fn scoped_thread<F: FnOnce() -> R + Send, R: Send>(cfg: thread::Builder, f:
     let result_ptr = Ptr(&mut result as *mut _ as *mut ());
 
     let thread = cfg.spawn(move || {
+        let _ = (&run, &result_ptr);
```
Mark-Simulacrum, line 128: These are necessary since raw pointers aren't send/sync, but this whole block is likely to go away shortly anyway - https://github.com/rust-lang/rust/pull/89104

src/tools/tidy/src/edition.rs
```diff
@@ -1,10 +1,15 @@
-//! Tidy check to ensure that crate `edition` is '2018'
+//! Tidy check to ensure that crate `edition` is '2018' or '2021'.
 
 use std::path::Path;
 
 fn is_edition_2018(mut line: &str) -> bool {
     line = line.trim();
-    line == "edition = \"2018\"" || line == "edition = \'2018\'"
+    line == "edition = \"2018\""
```
camelid, line 7: Thanks for cleaning this up to only allow `"`; I was thinking of doing that but didn't want to churn other crates in my PR.

src/tools/tidy/src/edition.rs
```diff
@@ -13,17 +18,38 @@ pub fn check(path: &Path, bad: &mut bool) {
...
-                );
+
+            // Library crates are not yet ready to migrate to 2021.
+            //
+            // The reference and rustc-dev-guide are submodules, so are left at
+            // 2018 for now. They should be removed from this exception list
+            // when bumped.
+            if path.components().any(|c| c.as_os_str() == "library")
+                || filestr.contains("src/doc/reference/style-check/Cargo.toml")
+                || filestr.contains("src/doc/rustc-dev-guide/ci/date-check/Cargo.toml")
```
camelid, line 34: I opened a (blocked) issue for this: rust-lang/rustc-dev-guide#1212.

camelid:
It looks like some examples in the lint docs need ` ```edition2018 ` added to them.

compiler/rustc_typeck/src/astconv/mod.rs
```diff
@@ -2417,13 +2417,14 @@ impl<'o, 'tcx> dyn AstConv<'tcx> + 'o {
...
-                    {
+                if let GenericParamDefKind::Lifetime = param.kind {
+                    if let hir::GenericArg::Lifetime(lifetime) = &lifetimes[i] {
                         self.ast_region_to_region(lifetime, None).into()
+                    } else {
+                        bug!()
                     }
-                    _ => bug!(),
+                } else {
+                    bug!()
```
estebank, line 2427: This could have been

```rust
match (param.kind, lifetimes.get(i)) {
    (GenericParamDefKind::Lifetime, Some(hir::GenericArg::Lifetime(lifetime))) => {
        self.ast_region_to_region(lifetime, None).into()
    }
    _ => bug!(),
}
```

but there's no need to change it in this PR :)

estebank (approved):

estebank:
@bors try @rust-timer queue

r=me with clean bill of perf

compiler/rustc_codegen_cranelift/src/bin/cg_clif_build_sysroot.rs
```diff
@@ -8,7 +8,6 @@
 
 #![feature(rustc_private)]
 
-extern crate rustc_data_structures;
```
bjorn3, line 11: Looks like you didn't update cg_clif to the 2021 edition in it's `Cargo.toml`.

Mark-Simulacrum: This change specifically comes from the move to SourceType::InTree in compile.rs for codegen backends. But yes, I didn't end up migrating cranelift -- I guess tidy ignores it -- I'll do that.

estebank:
@bors r+ rollup=never

cc @rust-lang/compiler

Mark-Simulacrum:
We're pretty clearly going to do this migration, and the regression is limited to a small percentage on a single benchmark. I'm going to approve this for now, but not mark it as triaged -- I'd like to make sure we identify the source of the regression, but don't want to block this on that.

@bors r=estebank

PR #86809, opened 2021-07-02
Add non_exhaustive_omitted_patterns lint related to rfc-2008-non_exhaustive

DevinR528:
Fixes: #84332 

This PR adds `non_exhaustive_omitted_patterns`, an allow by default lint that is triggered when a `non_exhaustive` type is missing explicit patterns. The warning or deny attribute can be put above the wildcard `_` pattern on enums or on the expression for enums or structs. The lint is capable of warning about multiple types within the same pattern. This lint will not be triggered for `if let ..` patterns.

```rust
// crate A
#[non_exhaustive]
pub struct Foo {
    a: u8,
    b: usize,
}
#[non_exhaustive]
pub enum Bar {
    A(Foo),
    B,
}

// crate B
#[deny(non_exhaustive_omitted_patterns)]
match Bar::B {
    Bar::B => {}
    _ => {}
}

#[warn(non_exhaustive_omitted_patterns)]
let Foo { a, .. } = Foo::default();

#[deny(non_exhaustive_omitted_patterns)]
match Bar::B {
    // triggers for Bar::B, and Foo.b
    Bar::A(Foo { a, .. }) => {}
    _ => {}
}
```

PR #89105, opened 2021-09-19
Fix: non_exhaustive_omitted_patterns by filtering unstable and doc hidden variants

### LINKED ISSUE #89042: New non_exhaustive_omitted_patterns lint exposes unstable and hidden enum variants.
See https://github.com/rust-lang/rust/pull/86809#issuecomment-921737544

The `non_exhaustive_omitted_patterns` lint exposes `#[unstable]` and `#[doc(hidden)]` variants.

E.g. for `io::Error`:

```
error: some variants are not matched explicitly
  --> src/main.rs:47:9
   |
47 |         _ => {}
   |         ^ pattern `Uncategorized` not covered
   |
  --> src/main.rs:4:12
   |
4  |     #[deny(non_exhaustive_omitted_patterns)]
   |            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
   = help: ensure that all variants are matched explicitly by adding the suggested match arms
   = note: the matched value is of type `ErrorKind` and the `non_exhaustive_omitted_patterns` attribute was found
```

But `Uncategorized` is both hidden and unstable, and should not be considered part of the public api.

It'd be good to fix this before it hits stable in 1.57.

DevinR528:
Fixes: #89042

Now that #86809 has been merged there are cases (std::io::ErrorKind) where unstable feature gated variants were included in warning/error messages when the feature was not turned on. This filters those variants out of the return of `SplitWildcard::new`.

Variants marked `doc(hidden)` are filtered out of the witnesses list in `Usefulness::apply_constructor`.

Probably worth a perf run :shrug: since this area can be sensitive.

### DIFF
--- a/compiler/rustc_mir_build/src/thir/pattern/deconstruct_pat.rs
+++ b/compiler/rustc_mir_build/src/thir/pattern/deconstruct_pat.rs
@@ -52,11 +52,11 @@ use rustc_data_structures::captures::Captures;
 use rustc_index::vec::Idx;
 
 use rustc_hir::{HirId, RangeEnd};
-use rustc_middle::mir::interpret::ConstValue;
 use rustc_middle::mir::Field;
 use rustc_middle::thir::{FieldPat, Pat, PatKind, PatRange};
 use rustc_middle::ty::layout::IntegerExt;
 use rustc_middle::ty::{self, Const, Ty, TyCtxt, VariantDef};
+use rustc_middle::{middle::stability::EvalResult, mir::interpret::ConstValue};
 use rustc_session::lint;
 use rustc_span::{Span, DUMMY_SP};
 use rustc_target::abi::{Integer, Size, VariantIdx};
@@ -675,6 +675,36 @@ impl<'tcx> Constructor<'tcx> {
         }
     }
 
+    /// Checks if the `Constructor` is a variant and `TyCtxt::eval_stability` returns
+    /// `EvalResult::Deny { .. }`.
+    ///
+    /// This means that the variant has a stdlib unstable feature marking it.
+    pub(super) fn is_unstable_variant(&self, pcx: PatCtxt<'_, '_, 'tcx>) -> bool {
+        if let Constructor::Variant(idx) = self {
+            if let ty::Adt(adt, _) = pcx.ty.kind() {
+                let variant_def_id = adt.variants[*idx].def_id;
+                // Filter variants that depend on a disabled unstable feature.
+                return matches!(
+                    pcx.cx.tcx.eval_stability(variant_def_id, None, DUMMY_SP, None),
+                    EvalResult::Deny { .. }
+                );
+            }
+        }
+        false
+    }
+
+    /// Checks if the `Constructor` is a `Constructor::Variant` with a `#[doc(hidden)]`
+    /// attribute.
+    pub(super) fn is_doc_hidden_variant(&self, pcx: PatCtxt<'_, '_, 'tcx>) -> bool {
+        if let Constructor::Variant(idx) = self {
+            if let ty::Adt(adt, _) = pcx.ty.kind() {
+                let variant_def_id = adt.variants[*idx].def_id;
+                return pcx.cx.tcx.is_doc_hidden(variant_def_id);
+            }
+        }
+        false
+    }
+
     fn variant_index_for_adt(&self, adt: &'tcx ty::AdtDef) -> VariantIdx {
         match *self {
             Variant(idx) => idx,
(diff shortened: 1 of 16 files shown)


src/test/ui/rfc-2008-non-exhaustive/reachable-unstable.rs
```diff
@@ -0,0 +1,49 @@
...
+        Foo::A => {}
+        Foo::B => {}
+    }
+    //~^^^^ non-exhaustive patterns: `_` not covered
+
+    match Foo::A {
+        Foo::A => {}
+        Foo::C => {}
+    }
+    //~^^^^ non-exhaustive patterns: `B` not covered
```
Nadrieril, line 48: Could you also add
```rust
    match Foo::A {
        Foo::A => {}
    }
```

compiler/rustc_mir_build/src/thir/pattern/usefulness.rs
```diff
@@ -969,12 +969,38 @@ impl<'p, 'tcx> Usefulness<'p, 'tcx> {
...
+                                        })
+                                        .map(|items| {
+                                            items
+                                                .iter()
+                                                .find(|item| item.has_name(sym::hidden))
+                                                .is_some()
+                                        })
+                                        .unwrap_or_default()
+                                    {
+                                        new = vec![Pat::wildcard_from_ty(pcx.ty)];
```
Nadrieril, line 996: This logic won't do what we want for
```rust
match Foo::A {
	Foo::A => {}
}
```
We probably want to just skip the hidden variants here, and add a wildcard at the end if we skipped anything

Nadrieril:
Nice, that was quick! I'd just ask for separate tests for the unstable part and the `doc(hidden)` part. It's cleaner to define our enum than to rely on `ErrorKind` since it will change over time.

`unstable` is tricky because it can normally only be defined within `std`. This seems to work:
```rust
#![feature(staged_api)]
#![stable(feature = "stable_test_feature", since = "1.0.0")]

#[unstable(feature = "unstable_test_feature", issue = "none")]
pub enum Foo {
	...
}
```
Look at e.g. `src/test/ui/pattern/usefulness/empty-match.rs` and `src/test/ui/pattern/usefulness/auxiliary/empty.rs` for how to use an external file (which will be compiled as a different crate) in a test.

Nadrieril:
Btw I'm only revieweing the logic, I don't feel competent to review the actual diagnostic change.
Actually, could you add the tests first, so we see the diagnostic changes in the following commits?

DevinR528:
Hopefully, these commits are clear enough and I didn't go the other way (splitting them too much so they are hard to interpret as a whole).

Nadrieril:
Woo, great! LGTM.

I expect a perf degradation because of the extra stability test. The `doc(hidden)` test is on a diagnostics-only path so won't matter. I expect that match-stress-enum will particularly suffer because it has a ton of variants.
Given that this is a correctness fix we might have to accept the perf regression anyway.

@bors try @rust-timer queue

Nadrieril:
@DevinR528 if you're curious, my general rules for splitting commits: each commit should compile and make sense on its own (I also commit the test outputs and that stage), and if two changes make sense individually I commit them separately. So I often end up with a ton of small commits ^^'. I haven't completely figured out how to limit that

DevinR528:
@Nadrieril cool thanks for the commt feedback/ideas! I will defiantly keep that in mind/try to use that as a guide.

Of course, no perf problems when we both expect them :roll_eyes:, you just never know.

Nadrieril:
Merge conflict incoming because of https://github.com/rust-lang/rust/pull/88950 . It's a rather massive change, but the main difference for this PR is that I've replaced `Fields::wildcards(pcx, ctor).apply(pcx, ctor)` with `DeconstructedPat::wild_from_ctor(pcx, ctor)` and `Pat::wildcard_from_ty(ty)` with `DeconstructedPat::wildcard(ty)`. That PR replaces `Pat` with my own version that knows about constructors and fields, which should make things simpler.

DevinR528:
I started rebasing will hopefully have time to finish it today and I'll check out https://github.com/rust-lang/rust/issues/89382#issuecomment-930622378 (but open separate PR to fix it)

src/test/ui/rfc-2008-non-exhaustive/auxiliary/unstable.rs
```diff
@@ -0,0 +1,13 @@
+#![feature(staged_api)]
+#![stable(feature = "stable_test_feature", since = "1.0.0")]
+
+#[stable(feature = "stable_test_feature", since = "1.0.0")]
+#[non_exhaustive]
```
Nadrieril, line 5: could you remove that so we only test variant filtering in general?

compiler/rustc_mir_build/src/thir/pattern/deconstruct_pat.rs
```diff
@@ -941,8 +942,24 @@ impl<'tcx> SplitWildcard<'tcx> {
...
+                                // The unstable feature is not activated so ignore this variant
+                                if !cx.tcx.stability().active_features.contains(&feature) {
+                                    None
+                                } else {
+                                    Some(Variant(idx))
+                                }
+                            } else {
+                                Some(Variant(idx))
+                            }
+                        })
```
Nadrieril, line 962: Ah I had missed something: right now you're filtering variants only in the `non_exhaustive` case, but we want to do it for all enums

Nadrieril: And I discovered https://doc.rust-lang.org/nightly/nightly-rustc/rustc_middle/ty/struct.TyCtxt.html#method.eval_stability which seems to be more appropriate since it checks a few more things

Nadrieril: I would do:
```rust
let mut ctors: SmallVec<[_; 1]> = def
    .variants
    .iter_enumerated()
    .filter(|(_, v)| {
        // Filter variants that depend on a disabled unstalbe feature.
        let is_enabled = !matches!(
            cx.tcx.eval_stability(v.def_id, None, DUMMY_SP, None),
            EvalResult::Deny { .. }
        );
        // If `exhaustive_patterns` is enabled, we exclude variants known to be
        // uninhabited.
        let is_uninhabited = cx.tcx.features().exhaustive_patterns
            && v.uninhabited_from(cx.tcx, substs, def.adt_kind(), cx.param_env)
                .contains(cx.tcx, cx.module);
        is_enabled && !is_uninhabited
    })
    .map(|(idx, _)| Variant(idx))
    .collect();
```
and then push a `NonExhaustive` into `ctors` in the `secretly_empty` and `non_exhaustive` cases.

Nadrieril: Perfect :+1:

Nadrieril:
LGTM in terms of implementation logic. Excluding unstable variants is obviously something we want to do.

The remaining question is whether it makes sense to hide `doc(hidden)` variants when listing missing cases in a match. How would something like that be decided? @oli-obk can I let you decide on that?

Nadrieril:
Oh and now that the logic also applies to normal enums, I expect a perf regression on match-stress-enum.

@bors try @rust-timer queue

compiler/rustc_mir_build/src/thir/pattern/usefulness.rs
```diff
@@ -589,11 +589,38 @@ impl<'p, 'tcx> Usefulness<'p, 'tcx> {
...
+                                                None
+                                            }
+                                        })
+                                        .map(|items| {
+                                            items
+                                                .iter()
+                                                .find(|item| item.has_name(sym::hidden))
+                                                .is_some()
+                                        })
+                                        .unwrap_or_default()
```
oli-obk: Please pull this or part of it into a method. There may be a chance for deduplication with other users of `sym::hidden`

src/test/ui/rfc-2008-non-exhaustive/reachable-patterns.rs
```diff
@@ -92,35 +92,6 @@ fn main() {
...
-        std::io::ErrorKind::Interrupted => {}
-        std::io::ErrorKind::Other => {}
-        std::io::ErrorKind::UnexpectedEof => {}
-        std::io::ErrorKind::Unsupported => {}
-        std::io::ErrorKind::OutOfMemory => {}
-        // All stable variants are above and unstable in `_`
-        _ => {}
-    }
-    //~^^ some variants are not matched explicitly
-
```
oli-obk, line 123: we could keep this and only remove the `//~^^ some variants are not matched explicitly`

src/test/ui/pattern/usefulness/stable-gated-patterns.rs
```diff
@@ -5,18 +5,14 @@ extern crate unstable;
...
-    //~^^ some variants are not matched explicitly
+    //~^^^ non-exhaustive patterns: `Stable2` not covered
 
     // Ok: all variants are explicitly matched
-    #[deny(non_exhaustive_omitted_patterns)]
     match Foo::Stable {
         Foo::Stable => {}
         Foo::Stable2 => {}
-        _ => {}
     }
```
oli-obk: shouldn't we still need the wildcard?

Nadrieril:
I like how this looks. Also this doesn't touch the core logic anymore so we don't risk unsoundness; good catch @oli-obk ><. Let's perf again since there used to be a small regression.

@bors try @rust-timer queue

PR #87815, opened 2021-08-06
encode `generics_of` for fields and ty params

BoxyUwU:
Fixes #87674
Fixes #87603

ICE was caused by calling `generics_of` on a `DefId` without any `generics_of` results. This was happening when we call `generics_of` on parent `DefId`s of an unevaluated const when we evaluate it.

r? @lcnr

PR #89111, opened 2021-09-19
Don't serialize empty predicates to metadata

Mark-Simulacrum:
This keeps empty predicates out of the metadata. In practice, it seems like it
is quite common to have a default predicate set, so this should be a win.
Diesel, for example, has 40% of decoded predicates being default/empty.

I believe this is an attempt to implement [this suggestion](https://github.com/rust-lang/rust/pull/87815#issuecomment-901974003), and hopefully will help with that regression.


Mark-Simulacrum:
Doesn't seem like this really has an effect. Closing.

Issue #89022, opened 2021-09-16
ICE While Testing generic_const_exprs, adt_const_params, & const_fn_trait_bound features


philliard3:
I was testing the limits of Rust's experimental const generics capabilities, and I came across this bug. The bug seems to occur only when remove_key is used.

I've had trouble minimizing this bug, so for context, I'll explain the program's purpose. The goal is to create a HashMap that has compile-time restrictions on what keys can be inserted. The first part of this program establishes a compile-time-checkable list of keys that can be any value that can be a const generic. In this case I've replaced my "Key" type alias with `&'static str` because that's where the error came up, as well as for readability. The second part of the program makes up the definition of the hashmap itself, which relies on the invariant that its "schema" contains a specific const-evaluable key in order to insert or remove a value. This also requires being able to build out a schema by adding and removing keys, producing a new type in the process because the types are what track the schema. The third part of the program is a small test for adding and removing keys, and it's when these tests are run that the error occurs.

This code compiles and runs if we do not run the test involving removing a key from a map that has already been created. Something about resolving the requirements in that test produces an error in the compiler.

### Code ([playground link](https://play.rust-lang.org/?version=nightly&mode=debug&edition=2018&gist=e6686b97436bea6b0c71d59cee03631d))

```Rust
#![feature(adt_const_params)]
#![feature(generic_const_exprs)]
#![feature(const_fn_trait_bound)]

use core::marker::PhantomData;
use std::collections::HashMap;

// const-evaluable equality for string slices
pub const fn str_eq(lhs: &str, rhs: &str) -> bool {
    let lhs_bytes = lhs.as_bytes();
    let rhs_bytes = rhs.as_bytes();
    let mut i = 0;
    let bytes = if lhs_bytes.len() == rhs_bytes.len() {
        lhs_bytes.len()
    } else {
        return false;
    };

    while i < bytes {
        if lhs_bytes[i] != rhs_bytes[i] {
            return false;
        }
        i += 1;
    }
    return true;
}

pub trait ContainsKey<const K: &'static str> {}

// trait used to compare two types that have type-encoded lists of keys (in this cast static strings)
pub trait KeySchema {
    const KEYS: &'static [&'static str];
    const SIZE: usize;
}

pub struct KeyNil;
impl KeySchema for KeyNil {
    const KEYS: &'static [&'static str] = &[];
    const SIZE: usize = 0;
}

pub struct KeyCons<Tail, const KEY_ID: &'static str>
where
[...]

PR #89113, opened 2021-09-20
dont `.ensure()` the `thir_abstract_const` query call in `mir_build`

BoxyUwU:
might fix an ICE seen in #89022 (note: this PR does not close that issue) about attempting to read stolen thir. I couldn't repro the ICE but this `.ensure` seems sus anyway.

r? @lcnr


compiler/rustc_mir_build/src/build/mod.rs
```diff
@@ -44,15 +44,15 @@ fn mir_build(tcx: TyCtxt<'_>, def: ty::WithOptConstParam<LocalDefId>) -> Body<'_
     let body_owner_kind = tcx.hir().body_owner_kind(id);
     let typeck_results = tcx.typeck_opt_const_arg(def);
 
-    // Ensure unsafeck is ran before we steal the THIR.
+    // Ensure unsafeck and abstract const building is ran before we steal the THIR.
```
lcnr, line 47: can you add a short comment about why we can't use `ensure` here?

after that r=me

PR #89120, opened 2021-09-20
Disable visible path calculation for PrettyPrinter in Ok path of compiler

In-line:


In-line:
Please, kindly schedule a perf run

estebank (approved):

In-line:
I forgot to commit main check, sorry! Please cancel scheduled perf run and redo it 😄

In-line:
There are multiple benchmarks in rustc-perf, which generate a lot of warnings. Even a single warning which will print the visible path of type will invoke `visible_parent_map`.

This explains a lot of neutral results.

I think we have a few questions now.
1) Is regression with uninhabited type panic acceptable? 
2) Is regression with query description acceptable?
    1) If no, will converting description() to be calculated lazily help?

For more information on impact to user, you can check out changed tests output.

jyn514:
Wow, this is a really amazing perf improvement! Definitely don't worry about the one or two regressions, this is really good :)

My only concern is the diagnostic regressions - I'm worried it will be confusing to get errors about items that aren't publicly visible. Maybe we could change those error paths to call try_print_visible_def_path explicitly?

jyn514:
> Is regression with query description acceptable?

This seems fine to me - it should only be shown in ICE messages and logging anyway.

Mark-Simulacrum:
The query description is shown in cycle errors (as can be observed with the test result changes). I think it is unfortunate to regress there. Could we check performance impact without that change? I'd guess that it is the primary contributor here, but would be happy to be surprised.

In-line:
Performance improvement comes from removal of `visible_parent_map`.

Even one `def_path_str` call will execute this query.

I think the solution should be to calculate description lazily.
-------- Original Message --------

estebank:
> My only concern is the diagnostic regressions - I'm worried it will be confusing to get errors about items that aren't publicly visible. Maybe we could change those error paths to call try_print_visible_def_path explicitly?

I think those regressions are unfortunate, but not _terrible_. We've had much worse diagnostic regressions in the past. As long as we have a plan, or at least a _ticket_ to address these, I'm ok with taking that hit. (My gut tells me that cycle errors aren't as common, and if a newcomer hits them they type's paths will not be in the top 3 things that confuse them about these.)

PR #89123, opened 2021-09-20
add Vec::push_within_capacity - fallible, does not allocate

### LINKED ISSUE #84649: Vec::push_in_capacity ?
This enhancement suggestion looks a bit silly to me, but I think it's worth asking for an opinion in this bug tracker.

I think this is a reasonable usage pattern of `Vec::spare_capacity_mut` and `Vec::set_len`. I've used it with a larger SomeData to avoid both an useless initialization and the use of `Vec::push` that in those cases is slightly slower, the disadvantage is that this code contains `unsafe{}` and it's a bit fiddly:

```rust
#![feature(vec_spare_capacity, maybe_uninit_extra)]
fn cond(i: usize) -> bool { i % 2 == 0 }
struct SomeData(usize); // Something larger.
fn main() {
    const N: usize = 1_000_000;
    let mut data = Vec::with_capacity(N);
    let mut data_len = 0;
    let data_uninit = data.spare_capacity_mut();

    for i in 0 .. N {
        if cond(i) {
            data_uninit[data_len].write(SomeData(i));
            data_len += 1;
        }
    }
    unsafe {
        data.set_len(data_len);
    }
}
```

An API that removes the unsafety could be:

```rust
#![feature(vec_push_in_capacity)]
fn cond(i: usize) -> bool { i % 2 == 0 }
struct SomeData(usize); // Something larger.
fn main() {
    const N: usize = 1_000_000;
    let mut data = Vec::with_capacity(N);

    for i in 0 .. N {
        if cond(i) {
            data.push_in_capacity(SomeData(i));
        }
    }
}
```

`Vec::push_in_capacity` is similar to `Vec::push`, but it doesn't include any growing strategy (and doesn't include such growing/realloc code in the binary), so it just uses the capacity. If the capacity isn't enough it panics (just like the first code panics if the capacity is finished at `data_uninit[data_len]`).

the8472:
This method can serve several purposes. It

* is fallible
* guarantees that items in Vec aren't moved
* allows loops that do `reserve` and `push` separately to avoid pulling in the allocation machinery a second time in the `push` part which should make things easier on the optimizer
* eases the path towards `ArrayVec` a bit since - compared to `push()` - there are fewer questions around how it should be implemented

I haven't named it `try_push` because that should probably occupy a middle ground that will still try to reserve and only return an error in the unlikely OOM case.

resolves #84649

### DIFF
--- a/library/alloc/src/vec/mod.rs
+++ b/library/alloc/src/vec/mod.rs
@@ -1773,6 +1773,50 @@ impl<T, A: Allocator> Vec<T, A> {
         }
     }
 
+    /// Appends an element if there is sufficient spare capacity, otherwise the element is returned.
+    ///
+    /// Unlike [`push`] method will not reallocate when there's insufficient capacity.
+    /// The caller should use [`reserve`] or [`try_reserve`] to ensure that there is enough capacity.
+    ///
+    /// [`push`]: Vec::push
+    /// [`reserve`]: Vec::reserve
+    /// [`try_reserve`]: Vec::try_reserve
+    ///
+    /// # Examples
+    ///
+    /// A manual, panic-free alternative to FromIterator
+    ///
+    /// ```
+    /// #![feature(vec_push_within_capacity, try_reserve)]
+    ///
+    /// use std::collections::TryReserveError;
+    /// fn from_iter<T>(iter: impl Iterator<Item=T>) -> Result<Vec<T>, TryReserveError> {
+    ///     let mut vec = Vec::new();
+    ///     for value in iter {
+    ///         if let Err(value) = vec.push_within_capacity(value) {
+    ///             vec.try_reserve(1)?;
+    ///             // this cannot fail, the previous line either returned or added at least 1 free slot
+    ///             let _ = vec.push_within_capacity(value);
+    ///         }
+    ///     }
+    ///     Ok(vec)
+    /// }
+    /// # from_iter(0..100).expect("please insert more memory");
+    /// ```
+    #[inline]
+    #[unstable(feature = "vec_push_within_capacity", issue = "none")]
+    pub fn push_within_capacity(&mut self, value: T) -> Result<(), T> {
+        if self.len == self.buf.capacity() {
+            return Err(value);
+        }
+        unsafe {
+            let end = self.as_mut_ptr().add(self.len);
+            ptr::write(end, value);
+            self.len += 1;
+        }
+        Ok(())
+    }
+
     /// Removes the last element from a vector and returns it, or [`None`] if it
     /// is empty.
     ///
--- a/library/alloc/src/vec/mod.rs
+++ b/library/alloc/src/vec/mod.rs
@@ -1773,9 +1773,10 @@ impl<T, A: Allocator> Vec<T, A> {
         }
     }
 
-    /// Appends an element if there is sufficient spare capacity, otherwise the element is returned.
+    /// Appends an element if there is sufficient spare capacity, otherwise an error is returned
+    /// with the element.
     ///
-    /// Unlike [`push`] method will not reallocate when there's insufficient capacity.
+    /// Unlike [`push`] this method will not reallocate when there's insufficient capacity.
     /// The caller should use [`reserve`] or [`try_reserve`] to ensure that there is enough capacity.
     ///
     /// [`push`]: Vec::push
@@ -1784,13 +1785,13 @@ impl<T, A: Allocator> Vec<T, A> {
     ///
     /// # Examples
     ///
-    /// A manual, panic-free alternative to FromIterator
+    /// A manual, panic-free alternative to [`FromIterator`]:
     ///
     /// ```
-    /// #![feature(vec_push_within_capacity, try_reserve)]
+    /// #![feature(vec_push_within_capacity)]
     ///
     /// use std::collections::TryReserveError;
-    /// fn from_iter<T>(iter: impl Iterator<Item=T>) -> Result<Vec<T>, TryReserveError> {
+    /// fn from_iter_fallible<T>(iter: impl Iterator<Item=T>) -> Result<Vec<T>, TryReserveError> {
     ///     let mut vec = Vec::new();
     ///     for value in iter {
     ///         if let Err(value) = vec.push_within_capacity(value) {
@@ -1801,10 +1802,10 @@ impl<T, A: Allocator> Vec<T, A> {
     ///     }
     ///     Ok(vec)
     /// }
-    /// # from_iter(0..100).expect("please insert more memory");
+    /// assert_eq!(from_iter_fallible(0..100), Ok(Vec::from_iter(0..100)));
     /// ```
     #[inline]
-    #[unstable(feature = "vec_push_within_capacity", issue = "none")]
+    #[unstable(feature = "vec_push_within_capacity", issue = "100486")]
     pub fn push_within_capacity(&mut self, value: T) -> Result<(), T> {
         if self.len == self.buf.capacity() {
             return Err(value);


the8472:
Since `push` and `extend_desugared` are very hot code:

@bors try @rust-timer queue

library/alloc/src/vec/mod.rs
```diff
@@ -1672,11 +1672,47 @@ impl<T, A: Allocator> Vec<T, A> {
         if self.len == self.buf.capacity() {
             self.reserve(1);
         }
+        unsafe { self.push_within_capacity(value).unwrap_unchecked() };
```
cynecx, line 1675: Since this is probably hot code, I'd just leave this as it is since the unsafe code is trivial enough to duplicate (debug builds may have additional call/eq/unwrap overhead, which is undesirable imo)

the8472: Yeah, let's see what perf says.

the8472: It already breaks a UI test checking stack usage in debug mode 🤷‍♂️

library/alloc/src/vec/mod.rs
```diff
@@ -2586,11 +2630,10 @@ impl<T, A: Allocator> Vec<T, A> {
...
             }
             unsafe {
-                ptr::write(self.as_mut_ptr().add(len), element);
                 // Since next() executes user code which can panic we have to bump the length
-                // after each step.
-                // NB can't overflow since we would have had to alloc the address space
-                self.set_len(len + 1);
+                // after each step, i.e. we can't move that after the loop.
+                // NB can't fail since we would have had to alloc the address space
+                self.push_within_capacity(element).unwrap_unchecked()
```
leonardo-m, line 2636: Is this .unwrap_unchecked() done to avoid a unused result warn-allow annotation?

the8472: It's to tell the compiler that that the length check (the `Err` case) isn't needed, otherwise I would have used `let _ =` like in the example.

lukaslueg:
Quick thought: The machinery provided and the example given could be expressed more concise as

```rust
let mut vec = Vec::new();
for value in iter {
    // push() will not reallocate
    vec.try_reserve(1).and_then(|v| v.push(value))?;
}
Ok(vec)
```

if `try_reserve(&mut self) -> Result<&mut Self, TryReserveError>` instead of `Result<(), TryReserveError>`.

Am I missing something?

the8472:
> Am I missing something?

`push` is `#[cfg(not(no_global_oom_handling))]` which means it's not available in code that doesn't want OOM panics. And it contains a call to `reserve()` which makes things more difficult for the optimizer when you call it in a loop.

Amanieu:
I would prefer if we wrapped the value in a proper error type instead of just returning a raw `T`. It's a common pattern in the standard library (see `BTreeMap::try_insert` which returns an `OccupiedError`).

library/alloc/src/vec/mod.rs
MSxDOS: In this case you're basically adding an unnecessary check coming from `push_within_capacity` and then tell the compiler to skip it using the `unreachable_unchecked` hint inside `unwrap_unchecked`. It's kinda ok, but it may degrade both compile and sometimes runtime performance as well. 
I think it would be better to have a separate, probably even public `push_unchecked` method that only does write + inc length and then use it where appropriate. 
Also, the length increment could benefit from the `unchecked_add` intrinsic in some niche cases since we're guaranteed that length can't be greater than capacity and capacity can't overflow. Optimization example: https://rust.godbolt.org/z/9fv8q7T9r

cynecx: > It's kinda ok, but it may degrade both compile and sometimes runtime performance as well.

Especially debug builds are more affected by this.

the8472: Reverted this part since the perf impact tended to be negative.

the8472: @MSxDOS tried that in #89152, didn't have any impact on the compiler and microbenchmarks didn't show any difference either.

jhpratt:
Couldn't this just be called `try_push`? Also, wouldn't a `try_extend` method make sense as well as an all-or-nothing approach?

hkratz:
> Couldn't this just be called `try_push`?

Personally I like `push_within_capacity()` better. With `try_push()` I would expect it to fail if an allocation fails.

leonardo-m:
Is "push_in_capacity" an acceptable name?

library/alloc/src/vec/mod.rs
```diff
@@ -1679,6 +1679,50 @@ impl<T, A: Allocator> Vec<T, A> {
...
+    ///     for value in iter {
+    ///         if let Err(value) = vec.push_within_capacity(value) {
+    ///             vec.try_reserve(1)?;
+    ///             // this cannot fail, the previous line either returned or added at least 1 free slot
+    ///             let _ = vec.push_within_capacity(value);
+    ///         }
+    ///     }
+    ///     Ok(vec)
+    /// }
+    /// # from_iter(0..100).expect("please insert more memory");
```
camelid, line 1710: nit: Seems like an `assert_eq!` would be good here.

leonardo-m:
In my code I've seen that a similar useful function could be extend_from_slice_within_capacity.

JohnCSimon:
triage: @the8472 what is the status of this PR? Is it waiting for the author or are you going to review it?

the8472:
@JohnCSimon I'm the author, not the reviewer. And yes it's waiting for review.

JohnTitor:
This hasn't received a review from @kennytm for 10 months so let's re-assign, maybe r? @Amanieu as you left a comment in the past

library/alloc/src/vec/mod.rs
```diff
@@ -1679,6 +1679,50 @@ impl<T, A: Allocator> Vec<T, A> {
         }
     }
 
+    /// Appends an element if there is sufficient spare capacity, otherwise the element is returned.
```
Amanieu, line 1682:
```suggestion
    /// Appends an element if there is sufficient spare capacity, otherwise an error is returned with the element.
```

library/alloc/src/vec/mod.rs
```diff
@@ -1679,6 +1679,50 @@ impl<T, A: Allocator> Vec<T, A> {
...
+    ///             // this cannot fail, the previous line either returned or added at least 1 free slot
+    ///             let _ = vec.push_within_capacity(value);
+    ///         }
+    ///     }
+    ///     Ok(vec)
+    /// }
+    /// # from_iter(0..100).expect("please insert more memory");
+    /// ```
+    #[inline]
+    #[unstable(feature = "vec_push_within_capacity", issue = "none")]
```
Amanieu, line 1713: This needs a tracking issue.

library/alloc/src/vec/mod.rs
```diff
@@ -1679,6 +1679,50 @@ impl<T, A: Allocator> Vec<T, A> {
...
+    /// ```
+    /// #![feature(vec_push_within_capacity, try_reserve)]
+    ///
+    /// use std::collections::TryReserveError;
+    /// fn from_iter<T>(iter: impl Iterator<Item=T>) -> Result<Vec<T>, TryReserveError> {
+    ///     let mut vec = Vec::new();
+    ///     for value in iter {
+    ///         if let Err(value) = vec.push_within_capacity(value) {
+    ///             vec.try_reserve(1)?;
+    ///             // this cannot fail, the previous line either returned or added at least 1 free slot
```
Amanieu, line 1704: Use `.unwrap()` to show that this cannot fail instead of ignoring the error.

Amanieu (approved):
LGTM!

Dylan-DPC:
@the8472 any updates on this? thanks

library/alloc/src/vec/mod.rs
the8472: I wrote it this way to avoid panic codegen which can be one of the benefits of this method.

the8472:
I addressed review comments except the one on the doc comment which I disagree with and haven't received further feedback on it. Since the reviewer gave his LGTM I'll take it as approval.

@bors r=amanieu

PR #88186, opened 2021-08-20
Make AST->HIR lowering incremental

cjgillot:
Implementation of MCP https://github.com/rust-lang/compiler-team/issues/452

AST->HIR lowering is currently done ahead of the construction of the incremental compilation engine. This issue proposes to make lowering incremental by introducing a query `lower_to_hir(def: LocalDefId) -> Option<&hir::OwnerInfo<'_>>` which lowers the AST item corresponding to definition def into the corresponding HIR and some associated information.

The objective is to eventually skip lowering some parts of the AST, and to allow progress towards making resolution and expansion incremental.

The proposed design is as follows:

* remove global analyses from the current lowering pass, and pull them back to resolution stage;
* attach lowering information to definitions instead of having multiple maps inside `hir::Crate`, effectively making `hir::Crate` an `IndexVec<LocalDefId, Option<OwnerInfo<'_>>>`;
* extract the required resolution outputs from the resolver and keep them immutable;
* allow to create new definitions from inside the query system (*);
* index the AST by `LocalDefId`, using an `IndexVec<LocalDefId, AstOwner>` where `AstOwner` is an enum holding `OwningRef<Lrc<Crate>, ItemLike>` for `ItemLike` different AST item types;
* perform HIR indexing per-owner;
* make lowering incremental.

One of the most subtle changes is (*), which allows to create new definitions from inside queries. It is made safe using an eval-always query which forces re-execution of its caller, and effectively enforces the side-effects creating the definition. The first time definitions are iterated over (for instance for metadata output), a read-lock to the definitions is leaked, which seals the list of definitions.

This PR is blocked on:
- [x]  rust-lang/rust#87234;
- [x]  rust-lang/rust#88019;
- [x]  rust-lang/rust#88703;
- [x]  rust-lang/rust#88880;
- [x]  rust-lang/rust#89124;
- [x]  rust-lang/rust#90146;
- [x]  rust-lang/rust#91112;
- [x]  rust-lang/rust#89278;
- [x]  rust-lang/rust#90451;
- [x]  rust-lang/rust#90204;
- [x] rust-lang/rust#95573.

The query scheme should basically look like this:
- the untracked ast is used to compute lower_to_hir;
- lower_to_hir yields hir_attrs, hir_owner_nodes and hir_owner_parent;
- hir_owner_nodes yields hir_owner;
- hir_module_items uses hir_owner_nodes to gather all the items inside a module;
- hir_crate uses hir_owner_nodes to gather all the items inside the crate.

[...]

PR #89124, opened 2021-09-20
Index and hash HIR as part of lowering

cjgillot:
Part of https://github.com/rust-lang/rust/pull/88186
~Based on https://github.com/rust-lang/rust/pull/88880 (see merge commit).~

Once HIR is lowered, it is later indexed by the `index_hir` query and hashed for `crate_hash`. This PR moves those post-processing steps to lowering itself. As a side objective, the HIR crate data structure is refactored as an `IndexVec<LocalDefId, Option<OwnerInfo<'hir>>>` where `OwnerInfo` stores all the relevant information for an HIR owner.

r? @michaelwoerister 
cc @petrochenkov


michaelwoerister:
Thanks, @cjgillot, I've blocked off some time to review this tomorrow.

michaelwoerister:
This looks pretty promising `:)`

> The first commits are other pending PRs (see merge commits).

Is this still based on pending PRs? If yes, would you mind listing them? If no, please rebase.

compiler/rustc_middle/src/hir/mod.rs
```diff
@@ -53,19 +41,19 @@ impl<'a, 'tcx> HashStable<StableHashingContext<'a>> for Owner<'tcx> {
...
 #[derive(Clone, Debug)]
 pub struct ParentedNode<'tcx> {
-    parent: ItemLocalId,
-    node: Node<'tcx>,
+    pub parent: ItemLocalId,
+    pub node: Node<'tcx>,
 }
 
 #[derive(Debug)]
 pub struct OwnerNodes<'tcx> {
```
michaelwoerister, line 49: I find the name of this struct a bit confusing. I read it as "a set of `OwnerNode` items" but it is really "all the node related information for a single Owner", right? Could this whole struct be merged into `OwnerInfo`?

compiler/rustc_middle/src/hir/mod.rs
```diff
@@ -53,19 +41,19 @@ impl<'a, 'tcx> HashStable<StableHashingContext<'a>> for Owner<'tcx> {
...
     /// Pre-computed hash of the full HIR.
-    hash: Fingerprint,
+    pub hash: Fingerprint,
     /// Full HIR for the current owner.
     // The zeroth node's parent is trash, but is never accessed.
```
michaelwoerister, line 53: What does "trash" mean? Is it `None`? Should it be `None`?

compiler/rustc_middle/src/hir/mod.rs
```diff
@@ -132,6 +111,46 @@ pub struct ModuleItems {
     foreign_items: Box<[ForeignItemId]>,
 }
 
+#[derive(Debug, HashStable)]
```
michaelwoerister, line 114: Can you add a doc string `/// Definition of a single HIR owner.`?

michaelwoerister (commented):
This is not a complete review. Just leaving a few comments.

michaelwoerister:
Do we have a high-level description of what the lowering process and query setup will look like after https://github.com/rust-lang/compiler-team/issues/452?

compiler/rustc_middle/src/hir/mod.rs
cjgillot: Trash in the sense it is set to `ItemLocalId::new(0)`, which is wrong since it's not even inside the same owner, but it's not a big deal since client code does not access `nodes[0].parent`.

cjgillot:
I added a summary description in #88186 description.

michaelwoerister (commented):
This looks great, thanks for the PR, @cjgillot! 

I left some comments below, mostly about adding a bit more documentation. One thing I'd like to clarify before merging though is the `bodies` field in `OwnerNodes`. How many bodies can an item have? Can it be more than one? And even if so, I expect that the number of bodies per item is small compared to the number of `ItemLocalId`s in it, so `FxHashMap<ItemLocalId, _>` or `SortedMap<ItemLocalId, _>` might be a better fit than `IndexMap<ItemLocalId, Option<_>>` for the `bodies` field.

the8472:
This PR also [regressed max-rss](https://perf.rust-lang.org/compare.html?start=5dab47dcd8267b8769421b46532414ec36d625e3&end=bd41e09da334697c0f993b36685cb599061d9faa&stat=max-rss) across several benchmarks. It's not noise, as can be seen in the [stm32f4 timelines](https://perf.rust-lang.org/?start=2021-10-13&end=2021-10-30&kind=percentfromfirst&stat=max-rss) which are low-noise for max-rss benchmarks.

PR #89127, opened 2021-09-20
Re-enable the `src/test/debuginfo/mutex.rs` test on Windows

wesleywiser:
This test required a newer version of cdb than was previously enabled in
CI thus leading to some bitrot in the test since the time it was
originally created. With the update to the `windows-latest` image last
week, we're now running this test in CI and thus uncovered the
regression.

I've updated the test and it now passes.

r? @ehuss


ehuss:
👍  I'm no expert here, but that output looks roughly like what I saw.

@bors r+

PR #86963, opened 2021-07-07
Update rand deps

klensy:
rand 0.7 -> 0.8
rand_xorshift 0.2 -> 0.3

This gives -6 crates for rustc compiler build

PR #89131, opened 2021-09-20
Update rand to 0.8 in library/ and compiler/

est31:
Removes some duplication. According to cargo tree, rand 0.7 is still used in various tools like miri, rls, or mdbook, but it's gone from library or compiler dependencies.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -39,7 +39,7 @@ version = "0.0.0"
 dependencies = [
  "compiler_builtins",
  "core",
- "rand 0.7.3",
+ "rand 0.8.3",
  "rand_xorshift",
 ]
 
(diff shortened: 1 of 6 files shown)


klensy:
There was attempt in #86963, perhaps you will find similar target build issue.

est31:
Good catch! Closing in favour of that PR.

klensy:
Btw, if you knew how to fix PR for wasm target - i'm open for suggestions.

est31:
@klensy first, apparently the error message [has been improved](https://github.com/rust-random/getrandom/commit/0eb0be130ea8120bb565c9abd3fa203c80f1b5f0) in the 0.2.3 release. It suggests enabling the js feature. For this to work with cargo test however one needs to have a test runner that has a js environment around. I'm not sure which environment rustc uses.

Another alternative might be refactoring tests in core to disable getrandom usage from the rand crate, and then just use fixed seeds for testing.

PR #89134, opened 2021-09-20
Revert the rustdoc box syntax removal

est31:
Reverts the rustdoc box syntax removal from #87781.

It turned out to cause (minor) perf regressions.

Requested in https://github.com/rust-lang/rust/pull/87781#issuecomment-922589168

### DIFF
--- a/src/librustdoc/clean/auto_trait.rs
+++ b/src/librustdoc/clean/auto_trait.rs
@@ -114,7 +114,7 @@ impl<'a, 'tcx> AutoTraitFinder<'a, 'tcx> {
             attrs: Default::default(),
             visibility: Inherited,
             def_id: ItemId::Auto { trait_: trait_def_id, for_: item_def_id },
-            kind: Box::new(ImplItem(Impl {
+            kind: box ImplItem(Impl {
                 span: Span::dummy(),
                 unsafety: hir::Unsafety::Normal,
                 generics: new_generics,
@@ -124,7 +124,7 @@ impl<'a, 'tcx> AutoTraitFinder<'a, 'tcx> {
                 negative_polarity,
                 synthetic: true,
                 blanket_impl: None,
-            })),
+            }),
             cfg: None,
         })
     }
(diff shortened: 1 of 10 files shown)


the8472:
since it's perf-relevant

@bors rollup=never

bjorn3:
This barely had any effect for any benchmark except two doc benchmarks which got up to 0.45% faster.

GuillaumeGomez:
For rustdoc we only look at doc benchmarks, so I guess it's good. :)

est31:
Great, the number in the improvement matches the number in the regression (also up to 0.45%, but negative): https://github.com/rust-lang/rust/pull/87781#issuecomment-922547631

I had the fear that this PR wasn't enough and more of #87781 had to be reverted as e.g. rustdoc would specially use some regressing component in compiler/ that the compiler itself made no heavy use of... but thankfully that wasn't the case.

Yeah the perf regression is minimal, and maybe a revert wouldn't have been needed, but I don't want to regress performance even by a little bit by removing box syntax usage from the compiler. The question is always what you gain from it, and box syntax removal is not worth "much" in my eyes, at least at this point when it's not clear if box syntax will go or maybe even get stabilized one day. Anyways, 90% of #87781 hasn't been reverted (check the diff numbers, this PR is roughly a tenth of #87781), and it's pretty easy to remove the remainder from rustdoc one day should the need arise.

GuillaumeGomez:
Thanks for that! Even though it was a small regression, we've been trying to go the other way for quite some time now so every improvement is very welcome (even when it's reverting a regression ;) ).

bjorn3:
This only partially fixes the original regression of the doc benchmarks: https://perf.rust-lang.org/compare.html?start=896f058f13d6c8021f7637817953a44d3a78be32&end=ba8cda2fa2c99ed6646f4dfe73bf4edad7e42a2d

This fixes serde-doc and stm32f4-doc, but doesn't fix many-assoc-items-doc, match-stress-enum-doc, unused-warnings-doc and the less significant regressions derive-doc, match-stress-exhaustive_patterns-doc, style-servo-doc and syn-doc.

est31:
@bjorn3 where is the unused-warnings-doc regression? I can't see it.

est31:
@bjorn3 oh right, there they are. If you check that checkbox in this PR's perf run though, you'll see that all these tests had improvements from this PR, so I think this PR fixes them all.

PR #83302, opened 2021-03-19
Get piece unchecked in `write`

camsteffen:
We already use specialized `zip`, but it seems like we can do a little better by not checking `pieces` length at all.

`Arguments` constructors are now unsafe. So the `format_args!` expansion now includes an `unsafe` block.

<details>
<summary>Local Bench Diff</summary>

```text
 name                        before ns/iter  after ns/iter  diff ns/iter   diff %  speedup 
 fmt::write_str_macro1       22,967          19,718               -3,249  -14.15%   x 1.16 
 fmt::write_str_macro2       35,527          32,654               -2,873   -8.09%   x 1.09 
 fmt::write_str_macro_debug  571,953         575,973               4,020    0.70%   x 0.99 
 fmt::write_str_ref          9,579           9,459                  -120   -1.25%   x 1.01 
 fmt::write_str_value        9,573           9,572                    -1   -0.01%   x 1.00 
 fmt::write_u128_max         176             173                      -3   -1.70%   x 1.02 
 fmt::write_u128_min         138             134                      -4   -2.90%   x 1.03 
 fmt::write_u64_max          139             136                      -3   -2.16%   x 1.02 
 fmt::write_u64_min          129             135                       6    4.65%   x 0.96 
 fmt::write_vec_macro1       24,401          22,273               -2,128   -8.72%   x 1.10 
 fmt::write_vec_macro2       37,096          35,602               -1,494   -4.03%   x 1.04 
 fmt::write_vec_macro_debug  588,291         589,575               1,284    0.22%   x 1.00 
 fmt::write_vec_ref          9,568           9,732                   164    1.71%   x 0.98 
 fmt::write_vec_value        9,516           9,625                   109    1.15%   x 0.99 
```
</details>

PR #89139, opened 2021-09-21
Use ZST for fmt unsafety

camsteffen:
as suggested here - https://github.com/rust-lang/rust/pull/83302#issuecomment-923529151.

### DIFF
--- a/library/core/src/fmt/mod.rs
+++ b/library/core/src/fmt/mod.rs
@@ -265,6 +265,26 @@ pub struct ArgumentV1<'a> {
     formatter: fn(&Opaque, &mut Formatter<'_>) -> Result,
 }
 
+/// This struct represents the unsafety of constructing an `Arguments`.
+/// It exists, rather than an unsafe function, in order to simplify the expansion
+/// of `format_args!(..)` and reduce the scope of the `unsafe` block.
+#[allow(missing_debug_implementations)]
+#[doc(hidden)]
+#[non_exhaustive]
+#[unstable(feature = "fmt_internals", reason = "internal to format_args!", issue = "none")]
+pub struct UnsafeArg;
+
+impl UnsafeArg {
+    /// See documentation where `UnsafeArg` is required to know when it is safe to
+    /// create and use `UnsafeArg`.
+    #[doc(hidden)]
+    #[unstable(feature = "fmt_internals", reason = "internal to format_args!", issue = "none")]
+    #[inline(always)]
+    pub unsafe fn new() -> Self {
+        Self
+    }
+}
+
 // This guarantees a single stable value for the function pointer associated with
 // indices/counts in the formatting infrastructure.
 //
@@ -337,10 +357,7 @@ impl<'a> Arguments<'a> {
     #[inline]
     #[unstable(feature = "fmt_internals", reason = "internal to format_args!", issue = "none")]
     #[rustc_const_unstable(feature = "const_fmt_arguments_new", issue = "none")]
-    pub const unsafe fn new_v1(
-        pieces: &'a [&'static str],
-        args: &'a [ArgumentV1<'a>],
-    ) -> Arguments<'a> {
+    pub const fn new_v1(pieces: &'a [&'static str], args: &'a [ArgumentV1<'a>]) -> Arguments<'a> {
         if pieces.len() < args.len() || pieces.len() > args.len() + 1 {
             panic!("invalid args");
         }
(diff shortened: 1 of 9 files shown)


library/core/src/fmt/mod.rs
```diff
@@ -337,10 +356,7 @@ impl<'a> Arguments<'a> {
     #[inline]
     #[unstable(feature = "fmt_internals", reason = "internal to format_args!", issue = "none")]
     #[rustc_const_unstable(feature = "const_fmt_arguments_new", issue = "none")]
-    pub const unsafe fn new_v1(
-        pieces: &'a [&'static str],
-        args: &'a [ArgumentV1<'a>],
-    ) -> Arguments<'a> {
+    pub const fn new_v1(pieces: &'a [&'static str], args: &'a [ArgumentV1<'a>]) -> Arguments<'a> {
```
Mark-Simulacrum, line 359: Was this previously already safe? I'm not sure I follow why we don't need the token as an arg here, too.

camsteffen: I think all invariants are held with the panic case below. This was discussed and [decided against](https://github.com/rust-lang/rust/pull/83302#discussion_r606724772) in the previous PR, but I figured we can reconsider with the perf issue at hand.

library/core/src/fmt/mod.rs
```diff
@@ -265,6 +265,25 @@ pub struct ArgumentV1<'a> {
...
+#[doc(hidden)]
+#[non_exhaustive]
+#[unstable(feature = "fmt_internals", reason = "internal to format_args!", issue = "none")]
+pub struct UnsafeArg;
+
+impl UnsafeArg {
+    /// This constructor in unsafe by design
+    #[doc(hidden)]
+    #[unstable(feature = "fmt_internals", reason = "internal to format_args!", issue = "none")]
+    #[inline(always)]
```
camsteffen, line 281: Would it be worth testing without `always`?

library/core/src/fmt/mod.rs
camsteffen: Just realized it should be okay to wrap the entire call with an unsafe block for this function. Want to try that and remove the panic?

Mark-Simulacrum:
r=me with an expanded comment on the UnsafeArg constructor. I think we should basically say something like "must be used for at most one call to new_v1_formatted, and created only if the other arguments meet the preconditions", with ideally those preconditions documented somewhere but for now we can just put in a FIXME for that.

Mark-Simulacrum:
Regressions in deeply-nested-async appear to be due to extra instructions in the type privacy visitor, presumably caused by the println! expanding to include an additional type after this PR. That then creates extra work for the visitor.

Regressions in ctfe-stress have no such obvious cause, and the test doesn't actually contain any println!s. The self-profile view shows this as increasing the number of specializes queries, which is somewhat surprising; it looks like all specializes queries on this benchmark are coming from typeck > evaluation_obligation > specializes based on self-profile results.

My best guess is that the code [here](https://github.com/rust-lang/rust/blob/9620f3a84b079decfdc2e557be007580b097fe43/compiler/rustc_trait_selection/src/traits/select/mod.rs#L1632) is the cause of the regression: presumably, we're trying to eliminate a newly added impl when checking some ImplCandidate. I compiled locally with before/after and -Zself-profile-events=all to get the inputs to both queries which produces some pretty weird results.

For the record, this was done with https://github.com/rust-lang/rustc-perf/pull/1031 and then using:

```bash
./target/release/collector diff_local self-profile +30278d3cf92b581550933546370443a5d5700002 +67365d64bcdfeae1334bf2ff49587c27d1c973f0 --include ctfe-stress-4 --builds Check --runs Full
mmview ./results/Zsp-30278d3cf92b581550933546370443a5d5700002-ctfe-stress-4-Check-Full/Zsp.mm_profdata | rg 'label: specializes' -A1 | rg additional_data > before
mmview ./results/Zsp-67365d64bcdfeae1334bf2ff49587c27d1c973f0-ctfe-stress-4-Check-Full/Zsp.mm_profdata | rg 'label: specializes' -A1 | rg additional_data > after
diff -u <(sort ./base) <(sort ./after) > query.diff
```

I put up the [diff](https://gist.github.com/Mark-Simulacrum/721265dd6a5b88877004f98dd1d9e854) here, but it shows some pretty weird inconsistencies that I wouldn't have expected from this PR -- my expectation would be that the specializes() query here is invoked deterministically, but that doesn't seem to be the case, or these indices are off. That said, *all* the changes in invocation are from core::ops::bit, which is *not changed* by this PR (right?) -- not even impls should be getting added in there.

There's some discussion on this particular topic on Zulip - https://rust-lang.zulipchat.com/#narrow/stream/182449-t-compiler.2Fhelp/topic/defpath.20printing

Mark-Simulacrum:
Zulip discussion indicates that the shifts in indices are not unexpected. I still don't have a good explanation for why there's extra queries though -- that seems like it shouldn't happen for the changes in this PR, perhaps suggesting that trait resolution is somehow being affected here :/

library/core/src/fmt/mod.rs
```diff
@@ -265,6 +265,26 @@ pub struct ArgumentV1<'a> {
...
 }
 
+/// This struct represents the unsafety of constructing an `Arguments`.
+/// It exists, rather than an unsafe function, in order to simplify the expansion
+/// of `format_args!(..)` and reduce the scope of the `unsafe` block.
+#[allow(missing_debug_implementations)]
+#[doc(hidden)]
+#[non_exhaustive]
+#[unstable(feature = "fmt_internals", reason = "internal to format_args!", issue = "none")]
+pub struct UnsafeArg;
```
cjgillot, line 275: Since the struct is public, it can be constructed directly without an unsafe block.

Mark-Simulacrum: It's non_exhaustive so I don't think that's true.

Mark-Simulacrum:
Okay, upon rebuilding before/after locally with debug logging enabled and some investigation, I'm pretty sure that the delta in query counts here was due to this landing before https://github.com/rust-lang/rust/pull/89016 which fixed some "noise" in the iteration order here. That PR also reduces the amount of specializes queries (though it's not expected that this has any direct relationship to this PR); we are just seeing that there is now a hopefully more stable order.

Regardless, this source of noise is now understood and we can mark this regression as triaged; we don't expect any fixes to the ctfe-stress to be possible. The previous comment already triaged the await-call-tree regression as something we're also not expecting to fix and consider acceptable.

(cc @lcnr in case you wanted some additional evidence for why that was important...)

library/core/src/fmt/mod.rs
RalfJung: `non_exhaustive` is insufficient, it only applies across crate boundaries. This compiles:
```rust
mod test {
#[non_exhaustive]
pub struct UnsafeArg;
}

fn main() {
    let x = test::UnsafeArg;
}
```
So I think this really should have a private field.

Mark-Simulacrum: Ah, good catch. Yeah, would be good to add a private field then.

camsteffen: Like this? `struct UnsafeArg(())`?

Mark-Simulacrum: That should work. We can probably drop non_exhaustive at that point as well.

camsteffen: > `non_exhaustive` is insufficient, it only applies across crate boundaries.

Isn't that enough since `UnsafeArg` is in the core crate?

Mark-Simulacrum: We shouldn't rely on that - it doesn't make this unsound for external users, but libcore itself can still get it wrong.

camsteffen: I see.

PR #89144, opened 2021-09-21
2229: Mark insignificant dtor in stdlib

arora-aman:
I looked at all public [stdlib Drop implementations](https://doc.rust-lang.org/stable/std/ops/trait.Drop.html#implementors) and categorized them into Insigificant/Maybe/Significant Drop. 

Reasons are noted here: https://docs.google.com/spreadsheets/d/19edb9r5lo2UqMrCOVjV0fwcSdS-R7qvKNL76q7tO8VA/edit#gid=1838773501

One thing missing from this PR is tagging HashMap as insigificant destructor as that needs some discussion.

r? @Mark-Simulacrum 

cc @nikomatsakis


library/std/src/lazy.rs
```diff
@@ -492,6 +492,7 @@ impl<T> SyncOnceCell<T> {
 }
 
 unsafe impl<#[may_dangle] T> Drop for SyncOnceCell<T> {
```
nikomatsakis, line 494: I'm not sure about *this* one -- I would leave this off

src/test/ui/closures/2229_closure_analysis/migrations/insignificant_drop.rs
```diff
@@ -1,162 +1,42 @@
...
+test_insig_dtor_for_type!(Vec<String>, vec_string);
+//test_insig_dtor_for_type!(HashMap<String, String>, hash_map);
+test_insig_dtor_for_type!(BTreeMap<String, i32>, btree_map);
+test_insig_dtor_for_type!(LinkedList<String>, linked_list);
+test_insig_dtor_for_type!(Rc<i32>, rc_i32);
+test_insig_dtor_for_type!(Rc<String>, rc_string);
+test_insig_dtor_for_type!(SyncOnceCell<String>, onecell);
+test_insig_dtor_for_type!(std::vec::IntoIter<String>, vec_into_iter);
+test_insig_dtor_for_type!(btree_map::IntoIter<String, String>, btree_map_into_iter);
+test_insig_dtor_for_type!(std::array::IntoIter<String, 5>, array_into_iter);
```
nikomatsakis, line 40: I think we need a test for e.g. `Vec<std::sync::MutexGuard>`, to show that this IS significant.

nikomatsakis (changes requested):

src/test/ui/closures/2229_closure_analysis/migrations/insignificant_drop.rs
arora-aman: The test for this is in significant_drop.rs, forgot to check that in last time 😅

petrochenkov:
It's pretty worrying when something that was rejected as a general purpose language feature is widely employed by a standard library and makes observable difference that cannot (?) be emulated on stable Rust.

Or is `rustc_insignificant_dtor` going to be stabilized in some form?

Mark-Simulacrum:
This feature only affects migrations for RFC 2229, so is both largely a quality of life improvement and something we could drop with no impact to stability guarantees (just affects lint's, and ones usually not used).

For these reasons I don't find it too concerning that stable Rust can't use this attribute. In practice I believe our expectation is that most crates will find std annotations sufficient to avoid spurious migrations.

petrochenkov:
I see, thanks.
I assumed that it actually affects the logic of the fine-grained captures.

compiler/rustc_ty_utils/src/needs_drop.rs
```diff
@@ -171,22 +173,44 @@ where
...
+                    debug!("adt_drop_tys: `{:?}` implements `Drop`", adt_def);
+                    return Err(AlwaysRequiresDrop);
+                }
+                DtorType::Insignificant => {
+                    debug!("adt_drop_tys: `{:?}` drop is insignificant", adt_def);
+
+                    // Since the destructor is insignificant, we just want to make sure all of
+                    // the passed in type parameters are also insignificant.
+                    // Eg: Vec<T> dtor is insignificant when T=i32 but significant when T=Mutex.
+                    return Ok(substs.types().collect::<Vec<Ty<'_>>>().into_iter());
```
Mark-Simulacrum, line 211: @nikomatsakis is probably better to say here, but I think this is actually the wrong thing to check? We probably need to look not at the subst types but at the fields with substs applied -- in particular, for something like the following I suspect this will do the wrong thing right now?

```rust
struct Foo<T: Trait> {
     var: T::Bar, // T::Bar is significant drop, T is not
}

Mark-Simulacrum: That said I suspect this won't affect any of the std types... so it might be OK

nikomatsakis: Exactly-- I agree that it's not necessarily what you would want in the "general case" but in practice I think it's adequate.

nikomatsakis (approved):

Mark-Simulacrum:
r? @nikomatsakis -- I could approve this I think, but ultimately you seem like the better candidate :)

Also beta-nominating since I think it's important we have this for the RFC 2229 migrations out of the box, it gets less useful the more people migrate their code.

ehuss:
@bors retry

macro-document-private-duplicate.rs has been disabled.

apiraino:
Beta backport accepted as per compiler team [on Zulip](https://rust-lang.zulipchat.com/#narrow/stream/238009-t-compiler.2Fmeetings/topic/.5Bweekly.5D.202021-09-30.20.2354818/near/255582058)

@rustbot label +beta-accepted

PR #89148, opened 2021-09-21
Suggest `_` in turbofish if param will be inferred from fn argument

estebank:


compiler/rustc_typeck/src/structured_errors/wrong_number_of_generic_args.rs
```diff
@@ -292,12 +292,37 @@ impl<'a, 'tcx> WrongNumberOfGenericArgs<'a, 'tcx> {
...
+        let is_used_in_input = |def_id| match node {
+            Some(hir::Node::Item(hir::Item { kind: hir::ItemKind::Fn(fn_sig, _, _), .. }))
+            | Some(hir::Node::TraitItem(hir::TraitItem {
+                kind: hir::TraitItemKind::Fn(fn_sig, _),
+                ..
+            }))
+            | Some(hir::Node::ImplItem(hir::ImplItem {
+                kind: hir::ImplItemKind::Fn(fn_sig, _),
+                ..
+            })) => fn_sig.decl.inputs.iter().any(|ty| match ty.kind {
```
oli-obk, line 305: Someone was so kind and wrote a function that does this signature extraction: https://doc.rust-lang.org/nightly/nightly-rustc/rustc_middle/hir/map/fn.fn_sig.html

estebank: Changed.

oli-obk:
r=me with CI passing

PR #89152, opened 2021-09-21
try unchecked add/sub for vec length calculation

the8472:
suggested in https://github.com/rust-lang/rust/pull/89123#discussion_r712968533

### DIFF
--- a/library/alloc/src/lib.rs
+++ b/library/alloc/src/lib.rs
@@ -133,6 +133,7 @@
 #![feature(str_internals)]
 #![feature(trusted_len)]
 #![feature(unboxed_closures)]
+#![feature(unchecked_math)]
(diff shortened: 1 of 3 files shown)


MSxDOS:
@the8472
https://github.com/rust-lang/rust/pull/89123#discussion_r713676220
> tried that in #89152, didn't have any impact on the compiler and microbenchmarks didn't show any difference either.

The `unchecked_add` in `push` alone removes unnecessary `if vec.is_empty()` checks and branches in some cases. Example: https://rust.godbolt.org/z/4raG4x5nK

the8472:
Yeah, but I'm saying it doesn't seem to have a measurable impact on the set of benchmarks we have. Although that is somewhat limited.

PR #89155, opened 2021-09-21
Use UnsafeArg in Arguments::new_v1 

camsteffen:
...instead of having a panic case.

Continues #89139

r? @Mark-Simulacrum

### DIFF
--- a/compiler/rustc_builtin_macros/src/format.rs
+++ b/compiler/rustc_builtin_macros/src/format.rs
@@ -846,26 +846,28 @@ impl<'a, 'b> Context<'a, 'b> {
 
         let args_slice = self.ecx.expr_addr_of(self.macsp, args_match);
 
-        // Now create the fmt::Arguments struct with all our locals we created.
-        let (fn_name, fn_args) = if self.all_pieces_simple {
-            ("new_v1", vec![pieces, args_slice])
-        } else {
-            // Build up the static array which will store our precompiled
-            // nonstandard placeholders, if there are any.
-            let fmt = self.ecx.expr_vec_slice(self.macsp, self.pieces);
-
+        let unsafe_arg = {
             let path = self.ecx.std_path(&[sym::fmt, sym::UnsafeArg, sym::new]);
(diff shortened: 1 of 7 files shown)


camsteffen:
@rust-timer queue

camsteffen:
@rust-timer queue

camsteffen:
@rust-timer queue

camsteffen:
@bors try @rust-timer queue

(probably should have waited for #89139 to be merged in the first place)

Mark-Simulacrum:
Looking at the regressions in the perf run above, they're mostly down to some slight codegen time regressions. Looking at the IR generated by format_args!, though, it seems like it's pretty much the same or maybe slightly better at most -- I think the primary difference is that the UnsafeArg unsafe constructor creates an extra basic block in the MIR and LLVM IR (since `call` is a terminator), even though it's not actually codegen'd into LLVM IR even without optimizations.

My suspicion is that these are related to UnsafeArg::new being const now -- I've adjusted to *just* apply that change for another evaluation.

@bors try @rust-timer queue

Mark-Simulacrum:
I believe the regressions, both in incremental and non-incremental benchmarks, are largely due to added queries checking that the UnsafeArg arguments being added are well-formed and such. That seems like a cost of this change, for sure, but one that seems not too unexpected.

I don't really have an explanation for the improvements, though. It seems like they are due to a large amount of instructions in process_obligations no longer running -- but that doesn't seem like it should be tied into this patch. It might be shuffling of instructions or something into different basic blocks which causes changes to optimizers, but I have no real proof of this. They don't seem as important to investigate regardless.

I am a little hesitant to proceed with accepting the regressions, particularly when it seems like they don't actually really benefit us (and indeed the API is "needlessly" made unsafe...). I think as of now, I'd prefer that we close this PR, but I am open to being convinced otherwise. I don't think there's an approach to this that avoids paying the cost for the extra work getting expanded into a bunch of user code (even if the runtime cost is ~0).

Issue #89095, opened 2021-09-19
rustdoc: Tracking issue for generating links to definitions on source code pages


GuillaumeGomez:
This is a tracking issue for generating links to definitions on source code pages.
The feature can be enabled by passing `--generate-link-to-definition -Zunstable-options` to a nightly `rustdoc`.

> [!NOTE]
>
> Tracking issues are used to record the overall progress of implementation.
> They are also used as hubs connecting to other relevant issues, e.g., bugs or open design questions.
>
> A tracking issue is however *not* meant for large scale discussion, questions, or bug reports.
> Instead, open a dedicated issue for the specific matter or post in the [#t-rustdoc](https://rust-lang.zulipchat.com/#narrow/channel/266220-t-rustdoc) Zulip channel.
>
> Discussion comments will get marked as off-topic or deleted.
> Repeated discussions on the tracking issue may lead to the tracking issue getting locked.

### Outstanding work

 * [ ] Improve "jump links" UI: for now it's highlighted background but it was suggested we could try an "underline dotted": https://github.com/rust-lang/rust/issues/88120
 * [ ] Add support to go to both source definition and documentation page (if available!). We could do it by adding a popup when hovering the item. (#89157)
 * [ ] Make all links go to source location, not documentation (currently, external types link to documentation)

### Steps

To go through stabilization, we will gradually enable it in different places:

 * [x] Enable it by default on the compiler docs (https://github.com/rust-lang/rust/pull/88292 and #89217)
 * [ ] Enable it by default on the std docs
 * [ ] Enable the feature by default (on nightly only) and remove the flag.
 * [ ] Send the stabilization PR.

### Implementation history

FIXME: Likely missing some PRs.

* #84176
* #88033 
* #91264
* #134216
* #135771

jyn514:
> Enable the feature by default (on nightly only) and reverse the flag (so that it disables instead of enabling).

I think we should *not* do this, this is exactly how we got in trouble with intra-doc links (see https://github.com/rust-lang/rust/issues/63305). Instead we should just enable it by default at the same time we remove the flag and stabilize the feature.

GuillaumeGomez:
This is a very good point. I had in mind that maybe people would want to disable it but I guess we can always add it later on if the needs appear.

PR #89157, opened 2021-09-21
Provide doc links at item definitions on source pages

GuillaumeGomez:
Part of #89095.

You can test it [here](https://rustdoc.crud.net/imperio/jump-to-def-extension/foo/index.html).

We now generate links to an item documentation on its definition.

cc @rust-lang/rustdoc

### DIFF
--- a/src/librustdoc/html/format.rs
+++ b/src/librustdoc/html/format.rs
@@ -517,6 +517,7 @@ impl clean::GenericArgs {
 }
 
 // Possible errors when computing href link source for a `DefId`
+#[derive(Debug)]
 pub(crate) enum HrefError {
     /// This item is known to rustdoc, but from a crate that does not have documentation generated.
     ///
--- a/src/librustdoc/html/highlight.rs
+++ b/src/librustdoc/html/highlight.rs
@@ -740,6 +740,11 @@ fn string<T: Display>(
                     )
                     .ok()
                     .map(|(url, _, _)| url),
+                    LinkFromSrc::Doc(def_id) => {
+                        format::href_with_root_path(*def_id, context, Some(&context_info.root_path))
+                            .ok()
+                            .map(|(doc_link, _, _)| doc_link)
+                    }
                 }
             })
         {
--- a/src/librustdoc/html/highlight.rs
+++ b/src/librustdoc/html/highlight.rs
@@ -728,11 +728,13 @@ fn string<T: Display>(
                     LinkFromSrc::Local(span) => context
                         .href_from_span(*span, true)
                         .map(|s| format!("{}{}", context_info.root_path, s)),
-                    LinkFromSrc::External(def_id) => {
-                        format::href_with_root_path(*def_id, context, Some(context_info.root_path))
-                            .ok()
-                            .map(|(url, _, _)| url)
-                    }
+                    LinkFromSrc::External(span) => context.href_from_span(*span, true).map(|s| {
+                        if s.starts_with("http://") || s.starts_with("https://") {
+                            s
+                        } else {
+                            format!("{}{}", context_info.root_path, s)
+                        }
+                    }),
                     LinkFromSrc::Primitive(prim) => format::href_with_root_path(
                         PrimitiveType::primitive_locations(context.tcx())[prim],
                         context,
--- a/src/librustdoc/html/render/context.rs
+++ b/src/librustdoc/html/render/context.rs
@@ -116,7 +116,6 @@ pub(crate) struct SharedContext<'tcx> {
     /// to `Some(...)`, it'll store redirections and then generate a JSON file at the top level of
     /// the crate.
     redirections: Option<RefCell<FxHashMap<String, String>>>,
-
     /// Correspondance map used to link types used in the source code pages to allow to click on
     /// links to jump to the type's definition.
     pub(crate) span_correspondance_map: FxHashMap<rustc_span::Span, LinkFromSrc>,
--- a/src/librustdoc/html/render/context.rs
+++ b/src/librustdoc/html/render/context.rs
@@ -335,7 +335,12 @@ impl<'tcx> Context<'tcx> {
                     let e = ExternalCrate { crate_num: cnum };
                     (e.name(self.tcx()), e.src_root(self.tcx()))
                 }
-                ExternalLocation::Unknown => return None,
+                ExternalLocation::Unknown => {
+                    let e = ExternalCrate { crate_num: cnum };
+                    let name = e.name(self.tcx());
+                    root = name.to_string();
+                    (name, e.src_root(self.tcx()))
+                }
             };
 
             sources::clean_path(&src_root, file, false, |component| {
--- a/src/librustdoc/html/render/span_map.rs
+++ b/src/librustdoc/html/render/span_map.rs
@@ -6,9 +6,8 @@ use rustc_data_structures::fx::FxHashMap;
 use rustc_hir::def::{DefKind, Res};
 use rustc_hir::def_id::{DefId, LOCAL_CRATE};
 use rustc_hir::intravisit::{self, Visitor};
-use rustc_hir::{ExprKind, HirId, Mod, Node};
+use rustc_hir::{ExprKind, HirId, Item, ItemKind, Mod, Node};
 use rustc_middle::hir::nested_filter;
-use rustc_hir::{ExprKind, GenericParam, GenericParamKind, HirId, Item, ItemKind, Mod, Node};
 use rustc_middle::ty::TyCtxt;
 use rustc_span::Span;
 
@@ -24,7 +23,7 @@ use std::path::{Path, PathBuf};
 #[derive(Debug)]
 pub(crate) enum LinkFromSrc {
     Local(clean::Span),
-    External(DefId),
+    External(clean::Span),
     Primitive(PrimitiveType),
     Doc(DefId),
 }
@@ -67,7 +66,7 @@ struct SpanMapVisitor<'tcx> {
 impl<'tcx> SpanMapVisitor<'tcx> {
     /// This function is where we handle `hir::Path` elements and add them into the "span map".
     fn handle_path(&mut self, path: &rustc_hir::Path<'_>, path_span: Option<Span>) {
-        let info = match path.res {
+        match path.res {
             // FIXME: For now, we only handle `DefKind` if it's not `DefKind::TyParam` or
             // `DefKind::Macro`. Would be nice to support them too alongside the other `DefKind`
             // (such as primitive types!).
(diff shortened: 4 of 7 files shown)


jsha:
I'd like to propose an alternative solution that is more complete and also
helps with other use cases. Instead of a popup, how about we put a link in
the margin next to all source view items, which links to their
corresponding documentation. That would be useful for general navigation of
a code base, and would provide a complement to the current [src] link that
takes you to source view.

If you click on a link to a definition, and want the doc view, it would be
two clicks, same as this popup solution. If you only wanted the source
view, it would be just one click.

This would also maintain the property that clicking links in rustdoc
usually navigates you somewhere rather than creating a popup.

camelid:
I believe jsha is suggesting that the source view would look something like this:

```
[docs]  1  fn my_fn() {
        2      println!("Hello, world!");
        3  }
```

It's an interesting idea, although I'm not sure how we'd fit it into the margin. Could you elaborate @jsha?

> The popup appears on hover too, so one click. But I remain very curious about what you're suggesting!

IIUC, you have to click once to get the popup, and then click again to actually go to the destination, so it's two clicks.

jsha:
Noah Lev accurately summarized my thinking. As for how to fit it, a couple
of ideas:

We could use an icon in the left margin, probably a "..." similar to what
GitHub shows when you click on a line number. This would be handy if we
later wanted to put additional things there, like "copy permalink".

Or, we could put [docs] in the right margin, by analogy with [src] in the
doc view. We may also want to consider using an icon for both src and doc
links.

GuillaumeGomez:
What happens in case you multiple items on one line like:

```rust
fn foo(a: Foo, b: Bar, c: String) {
```

?

In this case we have three different items, so that might make it a bit weird/dense, no?

> Or, we could put [docs] in the right margin, by analogy with [src] in the
> doc view. We may also want to consider using an icon for both src and doc
> links.

I don't think using the right margin in the source code viewer is a good idea because the width depends on the content currently.

camelid:
> What happens in case you multiple items on one line like:
> 
> ```rust
> fn foo(a: Foo, b: Bar, c: String) {
> ```
> 
> ?
> 
> In this case we have three different items, so that might make it a bit weird/dense, no?

There's only one item, `foo`, so there would only be one link. If you clicked on `Bar`, you would be taken to `Bar`'s definition in the source code, which would have its own `[docs]` link.

camelid:
> > Or, we could put [docs] in the right margin, by analogy with [src] in the
> > doc view. We may also want to consider using an icon for both src and doc
> > links.
> 
> I don't think using the right margin in the source code viewer is a good idea because the width depends on the content currently.

I agree; it seems like it could make the UI confusing. The `...` icon idea sounds interesting though.

jsha:
Yep, exactly. There is a possibility case of multiple items on one line,
like:

```
struct Foo; struct Bar;
```

But that's unusual, and rustfmt would break it up anyhow. If we encounter
such a case we should link to docs for the first one.

camelid:
> Yep, exactly. There is a possibility case of multiple items on one line, like:
>
> ```
> struct Foo; struct Bar;
> ```
>
> But that's unusual, and rustfmt would break it up anyhow. If we encounter such a case we should link to docs for the first one.

Hmm, that does seem likely uncommon, but maybe we could change the UI to avoid it altogether? E.g., we could have clicking on the `Foo` in `struct Foo;` give the same popup that the `...` would. Although that might not be as clean and simple as it could be...

camelid:
> I think I'd remove the code link highlighting.
> 
> Could you stylize the popup like this for consistency with the existing `[src]` feature?
> 
> > **[docs] [src]**

Yeah, I was just using `[docs]` as a mockup :)

Manishearth:
Stepping back a bit I think it's important to settle on the _purpose_ of jump-to-def, because what I see here are conflicting purposes. Adding a context menu for every JTD click is _also_ two actions (hover, then click) for an operation many here would want to be immediate. We need to figure out what problems we are solving, how we expect people to use this feature, and then use that to figure out which operations should and shouldn't be convenient.

In general we've had a lack of clarity on this part of this feature for a while and I suspect this will keep cropping up unless we do that work first.

GuillaumeGomez:
This is a good point. I like the idea to have access to both the documentation and the source from every item, but it slows down the browsing so I understand how that can be an issue. The other solution however (to put the documentation link on the definition) doesn't convince me in its current form. Oh well, we're not in a hurry so we can debate over this and try to reach the best solution for everyone.

camsteffen:
I suppose that, while browsing the docs, you're most likely going to see more docs next, and when browsing source, you're most likely going to navigate to more source next. So then, maybe a single click in source should navigate to the source, and the popup can just have docs (or both?). And probably add a small delay to the hover if not there already.

jyn514:
Please do not make the target of the link timing sensitive. That a) is super frustrating when you mean to go to one place and instead go to another; b) makes it impossible to use the longer version on mobile; and c) makes it much more difficult for people with slow reaction times to use the faster link.

In fact I'm again *any* version of this that relies on hover for navigation because it will be completely broken on mobile.

GuillaumeGomez:
> Please do not make the target of the link timing sensitive.

@camsteffen: I agree with @jyn514: that sounds like a very bad idea.

> In fact I'm again *any* version of this that relies on hover for navigation because it will be completely broken on mobile.

No, currently when you click or hover the link, it makes the popup appear (I thought about mobile devices as well ;) ). It's in the "What it does" explanation in the first comment.

camsteffen:
> Please do not make the target of the link timing sensitive.

I don't see how that applies. The popup shows *above* the link. So a direct click is always to source regardless of whether the popup shows. There should be a very small delay (like .3 secs) so that when you drag your mouse across the screen there is not a firework show of popups.

jsha:
> while browsing the docs, you're most likely going to see more docs next,
and when browsing source, you're most likely going to navigate to more
source next

Yes, this is also my intuition and underlying assumption. Another way of
putting it:

1. From a doc page, clicking an identifier takes you to docs.
2. From a source page, clicking an identifier takes you to source.
3. From a doc page, you can easily navigate to the source page for the item
you are looking at.
4. From a source page, you can easily navigate to the doc page for the item
you are looking at.

We already have 1, 2, and 3. We're only missing 4 to have a nice,
symmetric, easy to learn system. Also, 4 is useful not only in the "go to
definition" use case. It's also useful if you are manually scrolling
through a source page and decide you want to see the rendered docs for
something.

Talking about the popup implementation in this PR: We should strive to
always use existing, familiar UI elements. The closest UI element here
would be a context menu. However, the current implementation doesn't act
like a real context menu: it displays on hover (like a tooltip), but it
contains links (like some menus). It appears above the mouse cursor rather
than below and to the right. And it appears on left click rather than right
click.

So one direction we could go would be to make this more like a real context
menu. That has the advantage of being similar to how code navigation works
in an IDE. But it also has some disadvantages. For instance, overriding
right click on a web page can be quite annoying for the user, and should be
avoided without strong need. And opening a context menu for an item at the
bottom or right of the page can wind up with the menu awkwardly outside the
viewport.

I could see us wanting a context menu in the long run, if we do turn this
into a full code navigation tool. We would want multiple options, like go
to type definition, go to references, etc. However since we're considering
that explicitly out of scope, a context menu is probably overkill.

@GuillameGomez, can you say more about why you think my 1/2/3/4 approach
would wind up with lots of clicking back and forth between source and docs?
Do you disagree with the assumption that people most often want to go
docs->docs and source->source?

GuillaumeGomez:
No, I completely agree. I'm just not sure that it'll be obvious to people that to go to the item's docs, you need to go to its definition (even though I'm not a big fan, but I can live with it). My biggest issue is the UI suggested with `[doc]` in the right/left margin. The biggest issue with that is in case we have multiple definitions on a same line ; even though it's rare and unlikely, if there is a chance then it'll happen (declarations in macro for example?).

GuillaumeGomez:
I'm trying to understand why we have one extra pixel on the source code. CSS dark magic here we go!

GuillaumeGomez:
So what I did in this PR is the following: links to documentation are generated in the left border with the line numbers. It is the solution suggested above by @jsha. It matches the requirements for both mobile and desktop targets and works even with JS disabled.

We also discussed about making all idents links target source. For now, links on external types target the documentation page, I'll update it in a following PR. In the meantime, I updated #89095 to add this requirement.

So only thing remaining to discuss on this PR is the UI I think (mostly the colors of the "doc" links).

camelid:
How did you handle the "two items on one line" problem? From the demo, it looks like those just use the last item for the doc link.

I do feel concerned about how much clutter these links add to the source code pages. They at least should not have such a solid bright background IMO.

GuillaumeGomez:
> How did you handle the "two items on one line" problem? From the demo, it looks like those just use the last item for the doc link.

Sorry, forgot to precise it: only the first one has its link generated. The test ensures it.

> I do feel concerned about how much clutter these links add to the source code pages

It's very hard to have a result working on both mobile (meaning no hover event) and desktop. And in this case, it even works when JS is disabled. I think the result is a good compromise but I'm very open to alternatives as well. :)

> They at least should not have such a solid bright background IMO.

I'll quote myself:

> So only thing remaining to discuss on this PR is the UI I think (mostly the colors of the "doc" links).

Any suggestion for the appearance/colors is very welcome at this point! (I'm very bad at picking colors...)

camelid:
> It's very hard to have a result working on both mobile (meaning no hover event) and desktop. And in this case, it even works when JS is disabled. I think the result is a good compromise but I'm very open to alternatives as well. :)

I meant more that I'm not sure the doc links feature is worth it given the clutter it adds.

GuillaumeGomez:
Oh I see. For me it's quite important to have a way to go back to the documentation without using the search. It's the missing "connection" between the two in my point of view.

jsha:
One thing we could do if clutter is a problem: we could make it so that a "targeted"[^1] line number has the doc link next to it but others don't.

That would mean for the "jump to definition" use case, the doc link is quickly available. For the "browsing source generally" use case, it would take two clicks to go to the doc view - click the line number, then click the doc link that appears next to it.

[^1]: By this I mean, a line number that is the target of a URL fragment, like `#101`

camelid:
> Oh I see. For me it's quite important to have a way to go back to the documentation without using the search. It's the missing "connection" between the two in my point of view.

Well, I see the value of the feature, but I also see the value of rustdoc having a simple UI. It's hard to balance the two.

jsha:
We could additionally show the doc links when you hover the space to the left of the numbers.

camelid:
> We could additionally show the doc links when you hover the space to the left of the numbers.

Yeah, I thought about that too, but then it won't work on mobile. I guess on mobile we could require a tap?

GuillaumeGomez:
Just like we have the sidebar hidden by default for files, we could try having something similar for doc links I guess? An idea on what it would look like?

jsha:
> Yeah, I thought about that too, but then it won't work on mobile. I guess on mobile we could require a tap?

My thinking was that two things would trigger visibility of the doc link:
 - being the target of a URL fragment (which can be accomplished by clicking the line number, among other things).
 - hovering the left-side gutter.
 
 So on mobile you would always have to use the first of those two.

GuillaumeGomez:
I only find it not obvious that you need to go to an item definition to have its documentation link, but if you need to click on the line in addition to see it, it seems even worse. :-/

camelid:
(I think the title of this PR needs to be updated.)

GuillaumeGomez:
With the last update it looks like this:
Screenshot from 2021-12-08 21-54-04
Screenshot from 2021-12-08 21-54-24
Screenshot from 2021-12-08 21-54-30

GuillaumeGomez:
> (I think the title of this PR needs to be updated.)

(I think you're absolutely right and so I updated the PR)

camelid:
I updated it again since it's no longer strictly an extension to jump-to-def.

GuillaumeGomez:
I cleaned things up until we have an approved RFC and now generate the link on the item definition directly. All other links target the sources.

I also updated the generated docs if you want to give it a try.

JohnCSimon:
Triage:
@GuillaumeGomez This PR has merge conflicts

GuillaumeGomez:
Updated (and fixed the GUI tests).

GuillaumeGomez:
I'll fix the conflict once https://github.com/rust-lang/rust/pull/96636 is merged.

ojeda:
Cross referencers like Elixir (see https://github.com/GuillaumeGomez/rfcs/pull/1#issuecomment-1150068788) support references (i.e. uses) of an identifier too (similar to "Search All References" in IDEs). Since we are discussing a link back to the doc view, I thought it would be interesting to mention that other links may be useful too.

For instance, if `rustdoc` gained at some point the ability to list references for an item like Elixir (e.g. https://elixir.bootlin.com/linux/latest/A/ident/list_head), then it would be nice to have a 2nd option in the doc view (as a `[refs]` link) and maybe a way to go to that directly from the source view too. They could also appear in the search results in a 4th tab too.

GuillaumeGomez:
Finally took the time to fixed the merge conflicts.

JohnCSimon:
Ping from triage: 
@GuillaumeGomez  what is the status of this PR? Is it ready for review after the merge conflicts are resolved?

GuillaumeGomez:
We first need to finish the RFC so closing.

PR #89162, opened 2021-09-22
rustc_index: Add some map-like APIs to `IndexVec`

petrochenkov:
`IndexVec` is often used as a map, but its map APIs are lacking.
This PR adds a couple of useful methods.


davidtwco (approved):
LGTM!

Issue #85267, opened 2021-05-13
gdb pretty-printers not working for struct fields with gdb 10.x


mullr:
Hi! I'm trying to get pretty printing to work smoothly for rust in gdb. For a long time I was pretty sure nothing was working at all, but after liberal amounts of print(...) debugging in the associated python scripts, it appears that the pretty printing only kicks in when directly printing out one of the special structs, and not when they appear as members of something else. 

For example, if I have a Vec<u8> and print that directly from gdb, I see the result of the pretty printers. But if I have a struct TrivialNewtype(Vec<u8>) and try to print that from gdb, I see all the guts of the Vec (the pretty printer is not active). 

I made a small repro: https://github.com/mullr/gdb-pretty-test

For convenience, the code is: 
```rust
struct TrivialNewtype(Vec<u8>);

fn test() {
    let bare_vec = vec![1u8, 2, 3];
    let newtype_vec = TrivialNewtype(vec![1u8, 2, 3]);
    let placeholder = 12;
}

fn main() {
    test()
}
```
I am building and debugging like this:
```
cargo build
rust-gdb target/debug/gdb-pretty-test -ex 'b test' -ex 'r' -ex 'n' -ex 'n' -ex 'p bare_vec' -ex 'p newtype_vec'
```

I expected that when I print `newtype_vec`, the pretty printer for Vec would be used in a nested way. 

Instead, this was printed: 
```
Breakpoint 1, gdb_pretty_test::test () at src/main.rs:5
5           let bare_vec = vec![1u8, 2, 3];
6           let newtype_vec = TrivialNewtype(vec![1u8, 2, 3]);
7           let placeholder = 12;
$1 = Vec(size=3) = {1, 2, 3}
$2 = gdb_pretty_test::TrivialNewtype (alloc::vec::Vec<u8, alloc::alloc::Global> {buf: alloc::raw_vec::RawVec<u8, alloc::alloc::Global> {ptr: core::ptr::unique::Unique<u8> {pointer: 0x55555559abc0, _marker: core::marker::PhantomData<u8>}, cap: 3, alloc: alloc::alloc::Global}, len: 3})
```

### Meta
`rustc --version --verbose`:
```
rustc 1.52.1 (9bc8c42bb 2021-05-09)
binary: rustc
commit-hash: 9bc8c42bb2f19e745a63f3445f1ac248fb015e53
commit-date: 2021-05-09
host: x86_64-unknown-linux-gnu
release: 1.52.1
[...]

Issue #81813, opened 2021-02-06
lldb tests only run on macos


ehuss:
The lldb tests only run on macos on CI.  I think it might be worthwhile to run them on one CI builder for Linux, just to give a little more coverage.  However, it's not really clear if anyone is actually using rust-lldb.  

I discovered this because the debuginfo tests always fail on my local development system because I have lldb installed.  This was a little frustrating, so I investigated and discovered that the [`pretty-std-collections.rs`](https://github.com/rust-lang/rust/blob/master/src/test/debuginfo/pretty-std-collections.rs) was not being tested on CI because it is ignored on macos, and it was broken a while ago.

It might be as easy as just adding `apt install lldb` to a builder.  I don't think it is important to try to build it from source. Note that due to some oddity with how lldb apt package is installed, there is a missing symlink.  Something like `ln -s /usr/lib/llvm-10/bin/lldb-server /usr/bin/lldb-server-10.0.0` may be needed to get it to work.

ortem:
A related and possibly helpful comment: https://github.com/rust-lang/rust/issues/83891#issuecomment-813994524

pnkfelix:
I want to discuss this at the T-compiler meeting.

The on-going breakage of the debuginfo tests (and subsequent failure to catch such breakage in CI) is pretty worrisome, and frustrating as a contributor when all you want to do is run `x.py test` and let it go, without having to think too hard about "well which test suites are going to run that may be broken and that I have to skip myself by hand."

apiraino:
Issue has been discussed in T-compiler meeting [on Zulip](https://zulip-archive.rust-lang.org/238009tcompilermeetings/19557weekly2021042954818.html#236685766). This will probably need a follow up with `T-infra`.

Thus removing nomination:
@rustbot label -I-nominated

VladimirMakaev:
I think something needs to be done here to make CI more robust. This isn't just about who uses "rust-lldb" (btw we do use the synth provider shipped with rust-lang in our tooling) this is also about changes in compiler potentially breaking LLDB. For instance while working on https://github.com/rust-lang/rust/pull/124781#issuecomment-2095016283 i've added and fixed a bunch of tests and made sure they work on LLDB 18.1.

[...]

PR #89163, opened 2021-09-22
debuginfo: Add script for Rust support in lldb-mi

YakoYakoYokuYoku:
As of the merge of PR lldb-tools/lldb-mi#80 made possible for `lldb-mi` to run a command before execution this PR extends Rust debugging with the tool.

Tested via launching a debugging session with the C/C++ VSCode extension and setting the `miDebuggerPath` field to `/usr/bin/rust-lldb-mi`.

### DIFF
--- a/src/etc/rust-lldb-mi
+++ b/src/etc/rust-lldb-mi
@@ -1,0 +1,40 @@
+#!/bin/sh
+
+# Exit if anything fails
+set -e
+
+# Find the host triple so we can find lldb in rustlib.
+host=$(rustc -vV | sed -n -e 's/^host: //p')
+
+# Find out where to look for the pretty printer Python module
+RUSTC_SYSROOT=$(rustc --print sysroot)
+RUST_LLDB_MI="$RUSTC_SYSROOT/lib/rustlib/$host/bin/lldb-mi"
+RUST_LLDB="$RUSTC_SYSROOT/lib/rustlib/$host/bin/lldb"
+
+lldb=lldb
+lldb_mi=lldb-mi
+if [[ -f "$RUST_LLDB_MI" && -f "$RUST_LLDB" ]]; then
+    lldb_mi="$RUST_LLDB_MI"
+    lldb="$RUST_LLDB"
+else
+    if ! command -v "$lldb_mi" > /dev/null; then
+        echo "$lldb_mi not found! Please install it." >&2
+        exit 1
+    else
+        LLDB_VERSION=$("$lldb" --version | cut -d ' ' -f3)
+
+        if [ "$LLDB_VERSION" = "3.5.0" ]; then
+            cat << EOF >&2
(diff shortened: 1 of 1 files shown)


Mark-Simulacrum:
At a glance, I'd guess that this is not enough for the script to get shipped with Rust -- it probably needs to get added somewhere around [here](https://github.com/rust-lang/rust/blob/master/src/bootstrap/dist.rs#L528).

Additionally, I haven't heard of lldb-mi before -- can you share more details about whether it's an official component of the upstream lldb project? I'm not sure we should be shipping a wrapper for it otherwise...

cc @rust-lang/wg-llvm in case you have useful context

nagisa:
It seems like it used to be a part of the lldb project, but [not anymore](https://lists.llvm.org/pipermail/lldb-dev/2019-August/015357.html).

Mark-Simulacrum:
I'll r? @cuviper on this (as a member of wg-llvm as well as someone with distro experience), since I don't feel prepared to just merge this. I don't think we have any policy on which debuggers we ship "shims" for -- historically, it's just been lldb and gdb, which are the most prominent ones.

Ideally I think we'd want to avoid an ever increasing set of such scripts, particularly with no tests in tree, since it's very easy for them to unknowingly go stale (and that seems like a subpar experience). It looks like lldb-mi is shipped in the standard Ubuntu packages, so that's a potential "minimum" for inclusion, and may ease us testing it in CI to at least some level.

Curious to hear what others think though.

cuviper:
As for _what_, I think it's an implementation of [GDB/MI](https://sourceware.org/gdb/onlinedocs/gdb/GDB_002fMI.html), which you could use for embedding lldb in other tools that expect gdb.

YakoYakoYokuYoku:
Much of the purpose of this PR was to provide a GDB/MI LLDB interface for Rust debugging with tools that use MI (e.g. the [C/C++ tools](https://marketplace.visualstudio.com/items?itemName=ms-vscode.cpptools) for VSCode) for situations that may require an alternative like issue #85267. As for merging this I have no problems with doing a PR to the lldb-mi repo, although I think integration can be done here via tests in-tree nonetheless.

pnkfelix:
Thanks for putting up this Pull Request, @YakoYakoYokuYoku .

As others have already noted, we are not comfortable with adding more scripts to src/etc; at least, not scripts that don't have a stake-holder on the team.

We already are not doing a great job at supporting our "existing" debugger support, and I would prefer to first invest effort in fixing that before we even think about how to expand the scope of what debuggers we support in tree.

It sounds like you are willing to look into pushing this support into the lldb-mi repository instead. Which would be great. Maybe once *that* lands, we will be in a better place to look into adding the tests you suggest to our own repository. (Though I will just note that one reason that our debugger support is subpar is because of issues like #81813.)

PR #89164, opened 2021-09-22
Document `--show-type-layout` in the rustdoc book

camelid:
I also made a few small, related changes as separate commits.

r? @jyn514

### DIFF
--- a/src/doc/rustdoc/src/unstable-features.md
+++ b/src/doc/rustdoc/src/unstable-features.md
@@ -209,6 +209,22 @@ some consideration for their stability, and names that end in a number). Giving 
 `rustdoc` will disable this sorting and instead make it print the items in the order they appear in
 the source.
 
+### `--show-type-layout`: add a section to each type's docs describing its memory layout
+
+Using this flag looks like this:
+
+```bash
+$ rustdoc src/lib.rs -Z unstable-options --show-type-layout
+```
+
+When this flag is passed, rustdoc will add a "Layout" section at the bottom of
+each type's docs page that includes a summary of the type's memory layout as
+computed by rustc. For example, rustdoc will show the size in bytes that a value
+of that type will take in memory.
+
+Note that most layout information is **completely unstable** and may even differ
+between compilations.
+
 ### `--resource-suffix`: modifying the name of CSS/JavaScript in crate docs
 
 Using this flag looks like this:
--- a/src/doc/rustdoc/src/unstable-features.md
+++ b/src/doc/rustdoc/src/unstable-features.md
@@ -349,7 +349,7 @@ Some methodology notes about what rustdoc counts in this metric:
 Public items that are not documented can be seen with the built-in `missing_docs` lint. Private
 items that are not documented can be seen with Clippy's `missing_docs_in_private_items` lint.
 
-## `-w`/`--output-format`: output format
+### `-w`/`--output-format`: output format
 
 When using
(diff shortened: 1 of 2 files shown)


camelid:
*See also the discussion [on Discord](https://discord.com/channels/442252698964721669/459149231702278154/889463303997362206).*

src/doc/rustdoc/src/unstable-features.md
```diff
@@ -209,6 +209,22 @@ some consideration for their stability, and names that end in a number). Giving
 `rustdoc` will disable this sorting and instead make it print the items in the order they appear in
 the source.
 
+### `--show-type-layout`: add a section to each type's docs describing its memory layout
+
+Using this flag looks like this:
+
+```bash
+$ rustdoc src/lib.rs -Z unstable-options --show-type-layout
```
jyn514, line 217: Let's not show the prompt, it's different on Windows.
```suggestion
rustdoc src/lib.rs -Z unstable-options --show-type-layout
```
(And maybe change `bash` to `console` above.)

src/doc/rustdoc/src/unstable-features.md
```diff
@@ -209,6 +209,22 @@ some consideration for their stability, and names that end in a number). Giving
...
+```
+
+When this flag is passed, rustdoc will add a "Layout" section at the bottom of
+each type's docs page that includes a summary of the type's memory layout as
+computed by rustc. For example, rustdoc will show the size in bytes that a value
+of that type will take in memory.
```
jyn514, line 223: Why do you say "for example"? Isn't that the only part of the layout it shows?

src/librustdoc/html/render/print_item.rs
```diff
@@ -1663,7 +1663,7 @@ fn document_type_layout(w: &mut Buffer, cx: &Context<'_>, ty_def_id: DefId) {
             writeln!(
                 w,
                 "<div class=\"warning\"><p><strong>Note:</strong> Most layout information is \
-                 completely unstable and may be different between compiler versions and platforms. \
+                 completely unstable and may even differ between compilations. \
```
jyn514: How can this be true? The compiler has to be able to depend on the ABI to create dynamic libraries.

jyn514 (changes requested):

src/doc/rustdoc/src/unstable-features.md
camelid: No, it also shows the size of each variant. The things it displays could change later, so I'd rather not have an exhaustive list that needs to be updated.

src/librustdoc/html/render/print_item.rs
camelid: From [the Reference](https://doc.rust-lang.org/nightly/reference/type-layout.html#type-layout) (emphasis mine):

> Type layout can be changed with *each compilation*. Instead of trying to document exactly what is done, we only document what is guaranteed today.

I assume that means that the compiler is free to change layout if your crate is a static library or a binary. Also, there are (future) flags like `-Z randomize-layout` which would change layout between compilations.

src/doc/rustdoc/src/unstable-features.md
camelid: I agree, but I wanted to be consistent with the other sections, which use `bash` and `$ `.

jyn514: Ah ok, that makes sense. I didn't realize they were using this format already.

jyn514:
r=me with CI passing

Thanks for documenting this!

PR #89165, opened 2021-09-22
Fix read_to_end to not grow an exact size buffer

jkugelman:
If you know how much data to expect and use `Vec::with_capacity` to pre-allocate a buffer of that capacity, `Read::read_to_end` will still double its capacity. It needs some space to perform a read, even though that read ends up returning `0`.

It's a bummer to carefully pre-allocate 1GB to read a 1GB file into memory and end up using 2GB.

This fixes that behavior by special casing a full buffer and reading into a small "probe" buffer instead. If that read returns `0` then it's confirmed that the buffer was the perfect size. If it doesn't, the probe buffer is appended to the normal buffer and the read loop continues.

Fixing this allows several workarounds in the standard library to be removed:

- `Take` no longer needs to override `Read::read_to_end`.
- The `reservation_size` callback that allowed `Take` to inhibit the previous over-allocation behavior isn't needed.
- `fs::read` doesn't need to reserve an extra byte in `initial_buffer_size`.

Curiously, there was a unit test that specifically checked that `Read::read_to_end` *does* over-allocate. I removed that test, too.

### DIFF
--- a/library/std/src/fs.rs
+++ b/library/std/src/fs.rs
@@ -200,10 +200,9 @@ pub struct DirBuilder {
 
 /// Indicates how large a buffer to pre-allocate before reading the entire file.
 fn initial_buffer_size(file: &File) -> usize {
-    // Allocate one extra byte so the buffer doesn't need to grow before the
-    // final `read` call at the end of the file.  Don't worry about `usize`
-    // overflow because reading will fail regardless in that case.
-    file.metadata().map(|m| m.len() as usize + 1).unwrap_or(0)
+    // Don't worry about `usize` overflow because reading will fail regardless
+    // in that case.
+    file.metadata().map(|m| m.len() as usize).unwrap_or(0)
 }
 
 /// Read the entire contents of a file into a bytes vector.
--- a/library/std/src/io/mod.rs
+++ b/library/std/src/io/mod.rs
@@ -362,22 +362,18 @@ where
 // Because we're extending the buffer with uninitialized data for trusted
 // readers, we need to make sure to truncate that if any of this panics.
 fn read_to_end<R: Read + ?Sized>(r: &mut R, buf: &mut Vec<u8>) -> Result<usize> {
-    read_to_end_with_reservation(r, buf, |_| 32)
-}
-
-fn read_to_end_with_reservation<R, F>(
-    r: &mut R,
-    buf: &mut Vec<u8>,
-    mut reservation_size: F,
-) -> Result<usize>
-where
-    R: Read + ?Sized,
-    F: FnMut(&R) -> usize,
-{
     let start_len = buf.len();
+    let start_cap = buf.capacity();
     let mut g = Guard { len: buf.len(), buf };
     loop {
-        if g.len == g.buf.len() {
+        // If we've read all the way up to the capacity, reserve more space.
+        if g.len == g.buf.capacity() {
+            g.buf.reserve(32);
+        }
+
+        // Initialize any excess capacity and adjust the length so we can write
+        // to it.
+        if g.buf.len() < g.buf.capacity() {
             unsafe {
                 // FIXME(danielhenrymantilla): #42788
                 //
@@ -387,7 +383,6 @@ where
                 //   - Only the standard library gets to soundly "ignore" this,
                 //     based on its privileged knowledge of unstable rustc
                 //     internals;
-                g.buf.reserve(reservation_size(r));
                 let capacity = g.buf.capacity();
                 g.buf.set_len(capacity);
                 r.initializer().initialize(&mut g.buf[g.len..]);
@@ -404,8 +399,29 @@ where
                 assert!(n <= buf.len());
                 g.len += n;
             }
-            Err(ref e) if e.kind() == ErrorKind::Interrupted => {}
+            Err(ref e) if e.kind() == ErrorKind::Interrupted => continue,
             Err(e) => return Err(e),
+        }
+
+        if g.len == g.buf.capacity() && g.buf.capacity() == start_cap {
+            // The buffer might be an exact fit. Let's read into a probe buffer
+            // and see if it returns `Ok(0)`. If so, we've avoided an
+            // unnecessary doubling of the capacity. But if not, append the
+            // probe buffer to the primary buffer and let its capacity grow.
+            let mut probe = [0u8; 32];
+
+            loop {
+                match r.read(&mut probe) {
+                    Ok(0) => return Ok(g.len - start_len),
+                    Ok(n) => {
+                        g.buf.extend_from_slice(&probe[..n]);
+                        g.len += n;
+                        break;
+                    }
+                    Err(ref e) if e.kind() == ErrorKind::Interrupted => continue,
+                    Err(e) => return Err(e),
+                }
+            }
         }
     }
 }
@@ -2582,13 +2598,6 @@ impl<T: Read> Read for Take<T> {
 
     unsafe fn initializer(&self) -> Initializer {
         self.inner.initializer()
-    }
-
-    fn read_to_end(&mut self, buf: &mut Vec<u8>) -> Result<usize> {
-        // Pass in a reservation_size closure that respects the current value
-        // of limit for each read. If we hit the read limit, this prevents the
-        // final zero-byte read from allocating again.
-        read_to_end_with_reservation(self, buf, |self_| cmp::min(self_.limit, 32) as usize)
     }
 }
 
--- a/library/std/src/io/tests.rs
+++ b/library/std/src/io/tests.rs
@@ -362,24 +362,12 @@ impl<'a> Read for ExampleSliceReader<'a> {
 fn test_read_to_end_capacity() -> io::Result<()> {
     let input = &b"foo"[..];
 
-    // read_to_end() generally needs to over-allocate, both for efficiency
-    // and so that it can distinguish EOF. Assert that this is the case
-    // with this simple ExampleSliceReader struct, which uses the default
-    // implementation of read_to_end. Even though vec1 is allocated with
-    // exactly enough capacity for the read, read_to_end will allocate more
-    // space here.
+    // read_to_end() takes care not to over-allocate when a buffer is the
+    // exact size needed.
     let mut vec1 = Vec::with_capacity(input.len());
     ExampleSliceReader { slice: input }.read_to_end(&mut vec1)?;
     assert_eq!(vec1.len(), input.len());
-    assert!(vec1.capacity() > input.len(), "allocated more");
-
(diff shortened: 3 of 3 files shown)


joshtriplett:
@bors try @rust-timer queue

I don't expect this to turn up any surprises, but that's why they're called "surprises".

the8472:
This will cause the free functions `fs::read` and `fs::read_to_string` to do more syscalls since they now have to probe for EOF even though they already know the file size.

An alternative would be to update the documentation to recommend passing some extra capacity.

jkugelman:
> This will cause the free functions `fs::read` and `fs::read_to_string` to do more syscalls since they now have to probe for EOF even though they already know the file size.

It shouldn't cause them an extra syscall. With or without this change, there will be *n* syscalls that return data plus a final one that returns `Ok(0)`.

When the buffer is *not* an exact fit then there might be one extra `read`. I made sure to only do the probe read a single time when the buffer is still at its starting capacity. And it only happens if the buffer started with extra capacity -- an empty `Vec::new()` with 0 capacity won't require an extra read.

joshtriplett:
I would normally think this was noise, but it looks like a *lot* of small regressions about the same size. Is that a coincidence, or is this causing an issue?

Might be worth running the two toolchains with strace and seeing if the syscalls look noticeably different.

jkugelman:
> Might be worth running the two toolchains with strace and seeing if the syscalls look noticeably different.

Will do!

What's a good way to do that? I know strace well but this is my first contribution to this project and I'm not very familiar with testing it. Are you suggesting stracing on my computer or is there a CI tool I should be using? To test this PR I've been running `./x.py test library/std` on my computer. I could strace that command -- or is there a smarter way to do it?

Also, is there some way I can re-test the regressions in the report? I don't really understand what the benchmark did or how to read the results.

joshtriplett:
@jkugelman I'd suggest using https://rustc-dev-guide.rust-lang.org/compiler-debugging.html#downloading-artifacts-from-rusts-ci to install the exact before-and-after builds from the try above, the same ones used in the comparison. Then use both builds to do both a large and a small rustc compile, under `strace`, and try to compare the output side-by-side to see if there's an interesting difference in syscalls or their parameters. If there is a difference, you could also look at how long those syscalls take, and compare it to the regressions.

jkugelman:
@joshtriplett I wrote a script to count number of `read` syscalls and number of bytes read. I ran it on the hello world `main.rs` spit out by `cargo new`, 10 times with the old toolchain and 10 with the new. The new version is consistently *slightly* better on both measures.

<details>
<pre>
old (ce45663e14dac3f0f58be698cc530bc2e6e21682): reads=4405, bytes_read=30448537 (0)
new (4b920a40932f74a7159435b06d96cb50212514ff): reads=4395, bytes_read=30435031 (0)
old (ce45663e14dac3f0f58be698cc530bc2e6e21682): reads=4404, bytes_read=30448537 (1)
new (4b920a40932f74a7159435b06d96cb50212514ff): reads=4395, bytes_read=30435031 (1)
old (ce45663e14dac3f0f58be698cc530bc2e6e21682): reads=4404, bytes_read=30448537 (2)
new (4b920a40932f74a7159435b06d96cb50212514ff): reads=4394, bytes_read=30435031 (2)
old (ce45663e14dac3f0f58be698cc530bc2e6e21682): reads=4404, bytes_read=30448537 (3)
new (4b920a40932f74a7159435b06d96cb50212514ff): reads=4394, bytes_read=30435031 (3)
old (ce45663e14dac3f0f58be698cc530bc2e6e21682): reads=4404, bytes_read=30448537 (4)
new (4b920a40932f74a7159435b06d96cb50212514ff): reads=4394, bytes_read=30435031 (4)
old (ce45663e14dac3f0f58be698cc530bc2e6e21682): reads=4404, bytes_read=30448537 (5)
new (4b920a40932f74a7159435b06d96cb50212514ff): reads=4395, bytes_read=30435031 (5)
old (ce45663e14dac3f0f58be698cc530bc2e6e21682): reads=4404, bytes_read=30448537 (6)
new (4b920a40932f74a7159435b06d96cb50212514ff): reads=4394, bytes_read=30435031 (6)
old (ce45663e14dac3f0f58be698cc530bc2e6e21682): reads=4404, bytes_read=30448537 (7)
new (4b920a40932f74a7159435b06d96cb50212514ff): reads=4394, bytes_read=30435031 (7)
old (ce45663e14dac3f0f58be698cc530bc2e6e21682): reads=4406, bytes_read=30448537 (8)
new (4b920a40932f74a7159435b06d96cb50212514ff): reads=4394, bytes_read=30435031 (8)
old (ce45663e14dac3f0f58be698cc530bc2e6e21682): reads=4405, bytes_read=30448537 (9)
new (4b920a40932f74a7159435b06d96cb50212514ff): reads=4394, bytes_read=30435031 (9)
</pre>
</details>

```none
TOTAL:
old (ce45663e14dac3f0f58be698cc530bc2e6e21682): reads=44044, bytes_read=304485370 (total)
new (4b920a40932f74a7159435b06d96cb50212514ff): reads=43943, bytes_read=304350310 (total)
```

Is this what you had in mind? If so I can go throw a big hunk of Rust code at it and see what happens.

joshtriplett:
I'm genuinely surprised that it's non-deterministic.

I'm primarily interested in seeing the actual differences in the straces. Can you upload an strace from each toolchain for comparison?

(Also, you may want to invoke the toolchain directly rather than via the rustup wrapper, if possible, to minimize noise.)

jkugelman:
@joshtriplett 

```
$ cat main.rs
fn main() {
    println!("Hello, world!");
}
$ strace -o old-ce45663e14dac3f0f58be698cc530bc2e6e21682.full.strace -f ~/.rustup/toolchains/ce45663e14dac3f0f58be698cc530bc2e6e21682/bin/rustc main.rs
$ strace -o new-4b920a40932f74a7159435b06d96cb50212514ff.full.strace -f ~/.rustup/toolchains/4b920a40932f74a7159435b06d96cb50212514ff/bin/rustc main.rs
```

[old-ce45663e14dac3f0f58be698cc530bc2e6e21682.full.strace.gz](https://github.com/rust-lang/rust/files/7214407/old-ce45663e14dac3f0f58be698cc530bc2e6e21682.full.strace.gz)
[new-4b920a40932f74a7159435b06d96cb50212514ff.full.strace.gz](https://github.com/rust-lang/rust/files/7214408/new-4b920a40932f74a7159435b06d96cb50212514ff.full.strace.gz)

jkugelman:
I found it really difficult to visually diff them. I spent about half an hour trying to clean them up and hide all the spurious differences so I could find the "real" diffs. It was a real slog so I gave up and went with counting overall reads.

I don't know if you know an easy trick to compare the files. If not, I can go do more thorough analysis. I'm sure you don't want to spend hours on my PR. (And I really appreciate your help so far!)

Mark-Simulacrum:
Hm, so locally I seem to get entirely reproducible results:

```bash
$ sudo perf stat -r 30 -e syscalls:sys_enter_read ~/.rustup/toolchains/ce45663e14dac3f0f58be698cc530bc2e6e21682/bin/rustc t.rs
 Performance counter stats for '/home/mark/.rustup/toolchains/ce45663e14dac3f0f58be698cc530bc2e6e21682/bin/rustc t.rs' (30 runs):
              4370      syscalls:sys_enter_read
          0.109497 +- 0.000566 seconds time elapsed  ( +-  0.52% )
$ sudo perf stat -r 30 -e syscalls:sys_enter_read ~/.rustup/toolchains/4b920a40932f74a7159435b06d96cb50212514ff/bin/rustc t.rs
 Performance counter stats for '/home/mark/.rustup/toolchains/4b920a40932f74a7159435b06d96cb50212514ff/bin/rustc t.rs' (30 runs):
              4360      syscalls:sys_enter_read
          0.108526 +- 0.000544 seconds time elapsed  ( +-  0.50% )
```

This seems like it's both deterministic (I ran the above command several times; same result) and gives us the (expected) slight decrease in syscalls.

A good number of the regressions seem to be in -doc builds, and in some of those it looks like query counts went down (presumably since this PR deletes a function from std?). That creates some amount of noise in e.g. cachegrind diffs, as we're just doing different/less work. My inclination is to not consider these as blockers for this PR. However, I would like to get some sense why you're seeing nondeterminism with the syscalls -- what system are you running on? Can you try with the above perf commands and maybe see if those are OK?

jkugelman:
The `perf` runs on my system are consistent like yours and I get the exact same results: 4,370 vs. 4,360.

I fixed my `strace` script -- it was counting "exited with 0" and "SIGCHLD" lines. Now it's only counting reads and gives 100% consistent results, same as `perf`:

* 4,370 vs. 4,360 reads
* 30,434,827 vs. 30,421,321 bytes read

Script attached for transparency: [count-reads.gz](https://github.com/rust-lang/rust/files/7259456/count-reads.gz)

Mark-Simulacrum:
Ah, great, OK. Sounds like we're on the same page then.

I'm still not sure exactly where the regressions in the perf run are coming from. [Cycle count regressions](https://perf.rust-lang.org/compare.html?start=ce45663e14dac3f0f58be698cc530bc2e6e21682&end=4b920a40932f74a7159435b06d96cb50212514ff&stat=cycles%3Au&calcNewSig=true) are limited to -doc benchmarks entirely, which is a little weird, but may point to the regressions actually being caused by the change to the standard library (e.g., driven by different hashmap layout or something).

I am inclined to move ahead -- @joshtriplett, do you have any remaining concerns with this PR? I think the regressions are fairly minimal and not readily diagnosable (cachegrind doesn't point to anything obviously responsible, for example).

joshtriplett:
I reviewed both straces carefully, filtering out spurious differences, and I don't see any issues either. This seems good to go ahead with.

(I do think there are some interesting things to learn from the straces, for the purposes of future performance improvements. Thanks for collecting those!)

@bors r+

joshtriplett:
It looks like almost all of the changes in read calls are in the linker process, and caused by changes in ELF binaries.

The only meaningful behavior change is this:

Old:
```
read(5, "fn main() {\n    println!(\"Hello,"..., 46) = 45
read(5, "", 1)                    = 0
```

New:
```
read(5, "fn main() {\n    println!(\"Hello,"..., 45) = 45
read(5, "", 32)                   = 0
```

So the first call no longer reads an extra byte, and then the second call tries to read 32 bytes rather than 1. This matches the description of the "probe buffer" approach.

I think there'd be some value in trying to eliminate even the "probe buffer" call, but in the meantime, this seems like an improvement.

jkugelman:
> I think there'd be some value in trying to eliminate even the "probe buffer" call, but in the meantime, this seems like an improvement.

I created a follow-up issue: #89516.

Issue #86656, opened 2021-06-27
Tracking Issue for RFC 2948: Portable SIMD


calebzulawski:
Feature gate: `#![feature(portable_simd)]`

This is a tracking issue for RFC 2948 (rust-lang/rfcs#2948), the portable SIMD library (`std::simd`).

Portable SIMD project group: https://github.com/rust-lang/project-portable-simd
Implementation: https://github.com/rust-lang/stdsimd

More discussion can be found in the [#project-portable-simd](https://rust-lang.zulipchat.com/#narrow/stream/257879-project-portable-simd) zulip stream.

### Steps

- [ ] Implementation
- [ ] Update RFC
- [ ] Stabilization PR

### Unresolved Questions

- None yet.

### Implementation History

- None yet.

HannesGitH:
> Feature gate: `#![feature(portable_simd)]`

i'm sorry if this is the wrong place to ask but im rather new to rust and stumbled upon this issues as my compiler told me to

if i want to use this feature as soon as my compiler supports it can i gate it like:
```rust
#[cfg(feature = "portable_simd")]
use std::simd::Simd;
```

or is that only for feautures regarding my package (set in toml or passed to cargo?) if so what would be the appropriate way to use simd as soon as this issue is resolved?

Lokathor:
The `#![feature(portable_simd)]` part goes at the top of a binary or library.

It's a *language feature* not a *cargo feature* so it works a little differently.

It's unfortunate that they're both just "feature". Rust is often too terse when it counts.

agausmann:
@HannesGitH You might be able to do this with the [`rustversion`](https://github.com/dtolnay/rustversion) macro, which works similar to `#[cfg]`. However, I haven't tested any of these myself.

For example:

- To enable the experimental feature flag on nightly,

```rust
#![rustversion::attr(nightly, feature(portable_simd))]
```

- To conditionally import something on nightly only,

```rust
#[rustversion::nightly]
use std::simd::Simd;
```

I would recommend encapsulating all of your nightly-specific code in a module, so you only have to conditionally define the module itself and any external `use`s;

```rust
#[rustversion::nightly]
mod portable_simd;
```

HannesGitH:
thats pretty close to what i intented, thanks for letting me know, thats what i'll use 👍 

only problem is that once this feature becomes stable and one would compile my code using the stable version (which would then include actual simd) the resulting binary would still use my simd workarounds

[...]

Issue #88769, opened 2021-09-09
SIMD [iu]8xN tests SIGILL in --release


workingjubilee:
Hello, we have encountered an issue in the portable-simd CI unique to tests on our `i8xN` and `u8xN` types on x86_64. It emits a SIGTRAP, or sometimes a SIGILL, or on Windows, a STATUS_ACCESS_VIOLATION! How exciting! I can reproduce this locally. Our other tests do not do this, and it only happens in `--release`, so I suspect this is a miscompilation by LLVM 13 inappropriately optimizing something. Full command to repro: `cargo test --release --test [iu]8_ops`

I am going to try to minimize this from "literally the entire test suite" but I might take a bit. The issue became apparent about 7 days ago. Tests were passing about 20~30 days ago (August 8~18) judging by the last green in https://github.com/rust-lang/portable-simd/pull/155. I believe the time frame is correct for this to have been due to https://github.com/rust-lang/rust/pull/87570

### More minutiae
`RUSTFLAGS="-Copt-level=1"` passes with the above tests.
`RUSTFLAGS="-Copt-level=2"` crashes with them.

### Version it worked on

It most recently worked on: Rust 1.56 nightly (2021-07-17)

### Version with regression
`rustc --version --verbose`:
```
rustc 1.57.0-nightly (fdf65053e 2021-09-07)
binary: rustc
commit-hash: fdf65053e99e8966f9bd83b5a8491326cb33d638
commit-date: 2021-09-07
host: x86_64-pc-windows-msvc
release: 1.57.0-nightly
LLVM version: 13.0.0
```
but it also happened on rustc 1.57.0-nightly 2021-09-01

### Sample Crash

>Caused by:
[...]

PR #89167, opened 2021-09-22
pub use core::simd;

workingjubilee:
A portable abstraction over SIMD has been a major pursuit in recent years for several programming languages. In Rust, `std::arch` offers explicit SIMD acceleration via compiler intrinsics, but it does so at the cost of having to individually maintain each and every single such API, and is almost completely `unsafe` to use.  `core::simd` offers safe abstractions that are resolved to the appropriate SIMD instructions by LLVM during compilation, including scalar instructions if that is all that is available.

`core::simd` is enabled by the `#![portable_simd]` nightly feature tracked in https://github.com/rust-lang/rust/issues/86656 and is introduced here by pulling in the https://github.com/rust-lang/portable-simd repository as a subtree. We built the repository out-of-tree to allow faster compilation and a stochastic test suite backed by the proptest crate to verify that different targets, features, and optimizations produce the same result, so that using this library does not introduce any surprises. As these tests are technically non-deterministic, and thus can introduce overly interesting Heisenbugs if included in the rustc CI, they are visible in the commit history of the subtree but do nothing here. Some tests **are** introduced via the documentation, but these use deterministic asserts.

There are multiple unsolved problems with the library at the current moment, including a want for better documentation, technical issues with LLVM scalarizing and lowering to libm, room for improvement for the APIs, and so far I have not added the necessary plumbing for allowing the more experimental or libm-dependent APIs to be used. However, I thought it would be prudent to open this for review in its current condition, as it is both usable and it is likely I am going to learn something else needs to be fixed when bors tries this out.

The major types are
- `core::simd::Simd<T, N>`
- `core::simd::Mask<T, N>`

There is also the `LaneCount` struct, which, together with the SimdElement and SupportedLaneCount traits, limit the implementation's maximum support to vectors we know will actually compile and provide supporting logic for bitmasks. I'm hoping to simplify at least some of these out of the way as the compiler and library evolve.


workingjubilee:
cc @rust-lang/project-portable-simd

library/core/src/lib.rs
```diff
@@ -160,6 +160,10 @@
 #![feature(unboxed_closures)]
 #![feature(unsized_fn_params)]
 //
+// Extra-experimental language feature only used by core::simd
```
Mark-Simulacrum, line 163: I think we would want to either drop this from core::simd or inquire with the const development efforts as to the nature of the incompleteness. (By the time this PR lands, it may be a moot point, though, as the feature may well become complete...).

In particular, there's no current way to "silo" the incomplete_features allow to just core::simd, I believe, so this would effectively make all of core able to use this feature. We try to avoid doing that in libcore, AFAIK; incomplete features generally point at unsoundness or similar "bad bugs".

Mark-Simulacrum (commented):
Generally speaking this is probably "pretty alright" to include, but I am wondering if you can help answer several questions:

* Are the current APIs sound (in the safety sense) and any unsafe functions documented with preconditions?
* What is the expected evolution of the inclusion here? Are we expecting to maintain the out-of-tree repository and sync the subtree regularly?
* What is the reason for inclusion as a subtree, rather than a submodule? (Generally, IMO, if the subtree is not expected to be changed in rust-lang/rust due to e.g. compiler changes, a submodule is better. It seems to me that stdsimd should not really have changes forced on it by rust-lang/rust development, unless I'm missing something).

Let me know if I should be reading some standard docs for these questions. And just to be clear there are no "right" answers here, I just want to know what the thinking was - this might well be mergeable nearly as-is, modulo the in-code comments perhaps.

library/core/src/lib.rs
```diff
@@ -357,4 +361,23 @@ pub mod arch {
...
+// `core_simd` depends on libcore, but the contents of this module are
+// set up in such a way that directly pulling it here works such that the
+// crate uses this crate as its libcore.
+#[path = "../../portable-simd/crates/core_simd/src/mod.rs"]
+#[allow(missing_debug_implementations, dead_code, unsafe_op_in_unsafe_fn)]
+#[allow(rustdoc::bare_urls)]
+#[unstable(feature = "portable_simd", issue = "86656")]
+mod core_simd;
+
+#[doc = include_str!("../../portable-simd/crates/core_simd/src/core_simd_docs.md")]
```
Mark-Simulacrum, line 376: This suggests to me that a better approach -- if we go down the subtree route -- may be to have the rust-lang/rust subtree *not* be the root of the portable-simd repository, but rather the core_simd/src directory (or something along those lines).

But I'm not sure subtrees can support that kind of subdirectory workflow...

calebzulawski:
Regarding the `unsafe` question, yes all of the APIs are sound, and `unsafe` functions are documented.

library/core/src/lib.rs
calebzulawski: @workingjubilee I missed this previously--why is this feature enabled in `core`?

This feature isn't required for the main `core::simd` functionality, it adds a handful of functions that we have feature-gated (and I assumed would be disabled in `core::simd`).

library/core/src/lib.rs
bjorn3: I don't think git subtree supports that.

library/core/src/lib.rs
workingjubilee: They should be disabled, yes, but for some reason the crate requires the feature on when included as a module. I did not understand why.

workingjubilee:
Regarding "why subtree": in practice, portable-simd has directly informed (as in, we have committed) changes in, so far:
- compiler/rustc_typeck
- compiler/rustc_codegen_cranelift
- compiler/rustc_codegen_llvm

and sometimes changes hitting those may also need to touch this library to guarantee it compiles. Similarly, this has happened with stdarch at least once. They are both unfortunately "tightly coupled" with a lot of deep backend details, in other words. I intend to work on decoupling those and making them more loosely associated, but I do expect simultaneous changes to occur every now and then.

workingjubilee:
This library exposes very few unsafe functions: 2, and a currently-open PR will bring it up to all of 4. They all have their preconditions documented.

We have pretty carefully gone over it and have carefully scoped back to a limited initial attempt, and our test suite effectively gently fuzzes the library, so we are pretty sure it is sound.

The regression in `core` is expected, but the regression in the other marks is somewhat surprising, if those start with a built compiler? But it's mostly in docs, apparently, so maybe it's something to do with the different `core` affecting the generated documentation?

library/core/src/lib.rs
calebzulawski: I realized I was mixing this up with a different feature.  I just entirely removed this feature in https://github.com/rust-lang/portable-simd/pull/161 (since it was only needed for the old shuffles, which used arrays as const generic params).  Perhaps we should merge that PR and update this PR with those changes.

workingjubilee: Oh, great! That simplifies things a lot.

library/core/src/lib.rs
workingjubilee: Well... that depends on how things go over time, I think. The main other crate in our workspace is just designed to assist our test suite, and while it is stochastic it is rarely super-unpredictable and not that heavyweight. I think we expected early on that we might have a number of crates, like stdarch does, but it just kinda never panned out. Maybe that will change anyways sometime soon? But it seems equally likely that we might simplify down to a single crate in the root anyways.

And when I mentioned the tests to @jyn514, he suggested infra might actually be interested in moving them out into the main repo as well, as it would have forced issues like #88769 to get resolved faster, that's for sure. I don't know if that makes sense at the moment, though, and I didn't want to make a big fuss, but it's something I am thinking about in terms of future directions.

workingjubilee:
Regarding evolution: We have been developing it out of tree a lot because it minimizes our friction for the more lightweight, surface-level changes, it helps us mentally compartmentalize what we can, and it lets us experiment with unique test suites and etc., without impacting how main's CI goes. Yes, we intend to periodically sync it into the repo, like how other things have gone.

However, because we are designing an abstraction, it is far more likely that, unlike stdarch which will continue to differ as long as vendors keep releasing stuff, that sometime after our RFC and hashing that out, we eventually will hit a virtually "frozen" API where the code doesn't change much, and mostly with minor expansions, akin to a lot of libcore. This is another reason to prefer the subtree model: at that point we may well archive the other repo and simply have this live as part of `library/core` and `library/std`. In that way, having `portable-simd` as a subtree and thus "just a pile of commits" rather than an independent sub-repo-thingy may make more sense.

Mark-Simulacrum:
> The regression in core is expected, but the regression in the other marks is somewhat surprising, if those start with a built compiler? But it's mostly in docs, apparently, so maybe it's something to do with the different core affecting the generated documentation?

I looked into this a little. Rustdoc feels additions to core the most, because it has several "let's look at the whole world" places which often dominate its runtime. rustc does this too, but to a lesser extent in comparison to all the other logic, so it's a lighter weight compilation step. I don't think the regressions here are avoidable, unfortunately, with the current architecture in rustc and introduction into a mandatory-loaded crate (if we could have these intrinsics as a separate crate, then crates not loading it wouldn't suffer the perf hit).

So I think we'll have to accept this hit, unless we don't try to inline portable SIMD intrinsics directly into libcore (we could ship them in a separate crate and not make use of them from libcore, in theory). I don't think we want to do that right now though, and in general this is something that I believe we're hoping to help resolve in the future with chalk -- which in theory enables better caching and avoids some of this pain, IIRC.

> evolution / subtree vs. submodule

I think it's probably OK to move ahead as-is here. It seems likely that for the time being this is probably going to be fine and gives us a path to more extensive feedback and testing, I imagine, and if we restructure in the future it wouldn't be a big deal -- the files can just be moved. It looks like portable simd is pretty small, too, so we're not moving that much if we do that.

On the other hand, we could take an alternative approach here, I believe, and treat rust-lang/portable-simd as the "downstream" repository, if we believe (and I think this is reasonable) that it is likely we will eventually just archive rust-lang/portable-simd and fully develop here as you note in https://github.com/rust-lang/rust/pull/89167#issuecomment-933031740. This would require a bit more work on syncs, I think, but it should produce a slightly cleaner and more streamlined look (avoiding any of the path = ../../ dance, for example). It's probably not worth it unless we think just merging development of portable-simd into rust-lang/rust *now* makes sense (which it doesn't seem like a terrible idea to me, but you probably know better) -- the PR volume looks pretty low and you could still keep the out of tree repository for coordination work (issues etc).

workingjubilee:
- re: moving it in entirely right now:
Hmm. At the same time that I think it is inevitable, at the moment I feel like I am the main one working on the library who doesn't feel like she takes a huge "hit" from committing to rust-lang/rust, mentally: it's a barrier but it's more manageable, to me, because I have worked through it repeatedly, and I... well, I bought a computer specially for compiling rustc 😅. I feel moving immediately to this repo would add too much friction to the smaller "driveby" interactions we have gotten, as those are pretty direct community feedback, and while the "meaty internals" of the library have deep roots in the compiler, those are usually more about surface-level API. So, I don't want to go immediately into "we're just on rust-lang/rust master and PRing there now" before we RFC this. I think if it does arrive as our likely natural resting point it will just make sense to do some rearrangement to remove the odd positioning we have currently adopted.

- re: perf and compilation:
Yeah, I suspected it might be something like that, re: adding stuff makes trait or type resolution flail a bit more, it's just surprising to me that it hit rustdoc specially. I guess it makes sense that rustdoc by comparison does less overall so it's more visible for rustdoc.

- re: libcore or separate:
I have had a few people express interest to me in trying to use portable SIMD for Rust's internal algorithms to speed things up in various places, though whether it will or not I cannot say. Regardless, I expect it would be more of a disappointment if we were to push them into a separate crate, since that would mean libcore couldn't use them. ^^;

Mark-Simulacrum:
> The most material change is the not(bootstrap) setting arrived because rustc's extern "platform-intrinsics" fn simd_shuffle was adjusted to allow more interpretations on how we handle that in our exposed functions. This removes the need for adt_const_params.

Hm, sorry for the long wait -- checking back in here now. I am a little surprised that we're removing *the whole module* in the bootstrap compiler, rather than just cfg'ing a particular extern definition somewhere inside the portable-simd tree. It might also be that I've been slow enough the bootstrap compiler is now new enough this is not necessary?

That should be pretty simple to address either way (it's not critical we immediately ungate it, but if and when e.g. std or the compiler start using portable SIMD, we will need to not gate out the whole functionality).

All that said, I think this looks almost ready to land with this last question addressed, and then we can proceed here.

workingjubilee:
I expected it to not be required by now but for whatever reason it seems bootstrap is still needed.

Unfortunately also I looked through and a lot of random bits of code will depend on an adjustment to how `simd_bitmask` can be interpreted that we were going to try to land so that we can loosen up our mask definition, drop certain type boundaries, etc., simplify things. Since that establishes a notion of "equality", it's not really easy to simply cut out of all the higher-level functions, so it's easier to simply go ahead with things as-is and remove bootstrap when the relevant trains have left the station again. This shouldn't be a common thing, it just is today.

Mark-Simulacrum:
@bors r+ rollup=never

Thanks! We discussed a little in Zulip as well, but to note my impression here -- we may wish to leave cfg(not(bootstrap)) on semi-permanently until things stabilize (not in the rust #[stable] sense) so that we're not running into trouble when dropping this module temporarily due to codegen interface changes.

https://github.com/rust-lang/rust/pull/89167#issuecomment-933063065 contains a summary from me on the perf regression triaged label.

workingjubilee:
...Lovely!

workingjubilee:
One more "learning more about GitHub Actions and its workflow syntax than I really wanted to know" later and rust-lang/portable-simd should now have sufficient linting in its own CI to catch that kind of mistake so it never happens again.

@bors r=MarkSimulacrum rollup=never

workingjubilee:
Lovely! Now it's an EXCITING error which I will likely have to spend all day trying to learn how to debug!
@programmerjake thinks it might be spurious, and he's our Resident PowerPC Expert in effect, so we could simply retry it.
It is also an error on a "PPC64LE host and target", yet the CPU the log dumps data on is an Intel Xeon... we may be looking at an emulation bug? We certainly ran into one in https://github.com/rust-lang/portable-simd/pull/175, though I would expect that to be unrelated.

programmerjake:
> Lovely! Now it's an EXCITING error which I will likely have to spend all day trying to learn how to debug! @programmerjake thinks it might be spurious, and he's our Resident PowerPC Expert in effect, so we could simply retry it.

Oh, so it is on powerpc...i totally missed reading that part of the error log...

> It is also an error on a "PPC64LE host and target", yet the CPU the log dumps data on is an Intel Xeon... we may be looking at an emulation bug?

Definitely possible

jyn514:
This is a segfault bootstrapping libstd itself, so the only explanations I can think of are a miscompilation (seems unlikely since AFAIK none of the new functions here are used in the compiler) or a bug in the PowerPC CI (which seems very possible given how often the MacOS runners have bugs).

@bors retry

jyn514:
@workingjubilee ^ this one doesn't look spurious

workingjubilee:
So we apparently are tacitly using QEMU on x86-64 to reproduce a ppc64le environment with seemingly unknown settings.

I have an x86-64 machine myself, so if I put QEMU on this, I could try to repro this and recompile LLVM with debug assertions to find out what happens (since it is probably an error in LLVM). I could also ask Jake to try this out on his POWER9 machine, or finally ask the UoOregon myself for access.

workingjubilee:
Oh, new error. I guess I should have reloaded the page. But I thought I did...?

Okay, this looks like it is erring from trying to make various bytes of vendor floats available on Arm platforms that don't actually have those available, so we need to narrow down to those which actually do. We currently mostly designed around the assumption of Neon availability, which means we are looking at any of `target_arch = "aarch64"` or `all(target_arch = "arm", target_feature = "+neon")`.

There's actually other features that would enable some subsets of the vendor types also, like the DSP stuff, but that requires a more careful audit and refactor.

workingjubilee:
This was somewhat surprising since it's not immediately obvious the typedefs should not be valid on older Arm architectures.
Also I think I misspoke / misread earlier, the host is x86-64 I think but the build and target are for powerpc64le...?
Canadian crosses make my heard hurt.

workingjubilee:
It looks like we pass thumbv6m-none-eabi and then... fail on thumbv7m-none-eabi?
...what.
And it's specifically on the `int16x2_t` ones... there is no way out of this except adding a check job for every single Arm target, huh.

workingjubilee:
This time the conversions available should match exactly with the types in std::arch. 🙈 
@bors r=MarkSimulacrum

therealprof:
> It looks like we pass thumbv6m-none-eabi and then... fail on thumbv7m-none-eabi? ...what. And it's specifically on the `int16x2_t` ones... there is no way out of this except adding a check job for every single Arm target, huh.

Probably because v6m only has scalar operations while v7m has DSP instructions which provide SIMD support for `(u)int8x2_t` and `(u)int16x2_t`.

workingjubilee:
> Probably because v6m only has scalar operations 

That is not correct. Arm v6-M processors can potentially include DSP instructions.

> while v7m has DSP instructions which provide SIMD support for (u)int8x2_t and (u)int16x2_t.

That is also not the reason why the build failed.

While I mused aloud at first, there is no reason to speculate at this stage in the game. The target definition rustc uses for `thumbv7m-none-eabi` does not include the `"dsp"` feature by default, which is the problem, and the `#[cfg]` used in `core::arch` only exposes that feature for "M-class" Arm processors which **also** enable the DSP flag.

My first attempt was to simply restrict all the `core::arch` type conversions to `v7` processors or later, which, at my first reading (and testing the build for a few Arm targets) seemed correct and sufficient, and involved the minimum possible change. There was still some debate as to what we should do. As I did not exhaust the possibility space, I missed the Arm v7-M case, where the architecture version was sufficient to pass the `cfg` I used and attempt to build the conversions but failed on not actually having them present (which is why it suggested the hypothetically-available Neon typedefs instead).

The second try which resolved the problem required more rigorously examining the `core::arch` code in question, and we have developed an inclination to move the relevant type conversion code out of `core::simd` in the future, since it is clear the code is at risk of becoming incorrect if anything in `core::arch` changes anyways.

therealprof:
> That is not correct. Arm v6-M processors can potentially include DSP instructions. They were introduced as early as Arm v5TE (the Nintendo DS being a somewhat famous example).

Not sure how ARM v6-**M** and ARM v5TE are related. v6-m definitely can not have the DSP extension. I was wrong about v7m being able to use DSP instructions, too: For some reason I thought it would be optional in that architecture but actually it's only in Cortex-M4 and up (which is v7-**E**M) and there it is a standard feature.

LLVM can resolve SIMD types internally quite nicely, if the target doesn't support any operations on them it will simply turn them into scalar code and there're also some cases where it will use DSP (or other SIMD) instructions (if available) on scalar types if benefitial, e.g. on targets with the DSP instruction it can turn saturating operations on i8/u8/i16/u16 types into SIMD types ignoring remainder of the register.

This is all a bit shaky though and I totally expect that using `core::simd` will potentially expose some fancy bugs there.

Glad you hear you managed to get it sorted out and thanks for working on this important topic. ❤️

PR #56548, opened 2018-12-05
Optimized string FromIterator + Extend impls

Lucretiel:
I noticed that there was a lost opportunity to reuse string buffers in `FromIterator<String>` and `FromIterator<Cow<str>>`; updated the implementations to use these. In practice this translates to at least one fewer allocation when using these APIs.

Additionally, rewrote `Extend` implementations to use `iter.for_each`, which (supposedly) helps the compiler optimize those loops (because iterator adapters are encouraged to provide optimized implementations of `fold` and `try_fold`.

PR #89169, opened 2021-09-22
Try to reuse buffers in String::extend

Lucretiel:
This PR extends the logic around buffer reuse in `FromIterator<String> for String` into `Extend` and a few other `FromIterator` implementations. Basically, in cases where we're extending from allocated types (`String`, `Cow<str>`, `Box<str>`), we check if `self` is empty and, if so, just take the incoming value by move rather than copying the data.

Follow up to https://github.com/rust-lang/rust/pull/56548.

### DIFF
--- a/compiler/rustc_builtin_macros/src/deriving/generic/mod.rs
+++ b/compiler/rustc_builtin_macros/src/deriving/generic/mod.rs
@@ -567,8 +567,11 @@ impl<'a> TraitDef<'a> {
             })
         });
 
-        let Generics { mut params, mut where_clause, span } =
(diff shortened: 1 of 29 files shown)


joshtriplett:
Looks reasonable to me. r=me once CI passes and once a perf run confirms that this is either neutral or an improvement.

Mark-Simulacrum:
It's not clear to me why it's failing either. It may help to rebase on latest master so we can rerun CI, at least?

It does seem likely to be related to cfg(test) somehow, but I'm not sure how. It may be that we just need to cfg out this impl in cfg(test) or something like that...

Issue #88132, opened 2021-08-18
Leak sanitizer UI test hangs on Macos ARM64


hkratz:
Try on a Apple Silicon Mac:
```
./configure --build=aarch64-apple-darwin --enable-sanitizers --enable-profiler --set rust.jemalloc --set llvm.ninja=false
./x.py test --stage 1  src/test/ui/sanitize/leak.rs
```
Expected result: Test passes.

Actual result: Test hangs:
```
running 1 test
test [ui] ui/sanitize/leak.rs has been running for over 60 seconds
```

See also: #88131

@rustbot label +A-sanitizer +A-llvm +O-macos +O-arm

hkratz:
This problem is still reproducible with LLVM 13.

Coder-256:
The test seems to pass now (tested with `./x.py test --stage 1 tests/ui/sanitizer/leak.rs`), would it be reasonable to consider re-enabling leak sanitizer on aarch64-apple-darwin now?

madsmtm:
I think this issue can be closed (using the sanitizer no longer hangs), but https://github.com/rust-lang/rust/issues/121624 is still a blocker before we can enable it as supported.

PR #79883, opened 2020-12-10
Enable ASan, TSan, UBSan for aarch64-apple-darwin.

frewsxcv:
I confirmed ASan, TSan, UBSan all work for me locally with `clang` on my new Macbook Air.

~This requires https://github.com/rust-lang/llvm-project/pull/86~

PR #89170, opened 2021-09-22
Disable the leak sanitizer on Macos aarch64 for now

hkratz:
It is currently broken, see #88132.


petrochenkov:
The leak sanitizer for this target was enabled in https://github.com/rust-lang/rust/pull/79883, cc @frewsxcv.

@bors r+

Issue #79449, opened 2020-11-26
std::fs::canonicalize uses GetFinalPathNameByHandleW, which can fail


Dwedit:
I was just testing out the Win32 function `GetFinalPathNameByHandleW`, and saw a case where it failed to work.

Steps to reproduce:
* Install OSFMount (Passmark)
* Mount a Disk Image (I used a NTFS image file, mounted as drive X)
* Use `CreateFileW` on a file on the disk image  (my path: "X:/flash Movies/11-29-04-thanksgiving.swf")
* Call `GetFinalPathNameByHandleW` on that handle, using `FILE_NAME_NORMALIZED`

Function returns 0 instead of returning a path name, and `GetLastError` returns 1 (`ERROR_INVALID_FUNCTION`).

However, if I call `GetFinalPathNameByHandleW` using `VOLUME_NAME_NT`, I get something:  "\Device\OSFMDisk0\Flash Movies\11-29-04-thanksgiving.swf", but this path isn't usable with `CreateFileW`.

There are alternative ways to canonicalize a filename on Win32:
`PathCanonicalizeW` is the traditional way to do it, works on Windows 2000 and later, but it is limited to 260 character filenames, and is marked as deprecated by Microsoft.
`PathCchCanonicalizeEx` is the replacement function since Windows 8, and is unavailable in Windows 7 and earlier, but does support very long path names.

Dwedit:
According to http://mail.openjdk.java.net/pipermail/core-libs-dev/2020-July/067594.html,

> With ImDisk virtual disks, applications that rely on Windows Volume Mount Manager APIs, like GetFinalPathNameByHandleW, will fail because ImDisk does not interact with Windows Volume Mount Manager at all, it effectively bypasses it. Essentially, GetFinalPathNameByHandle and few other APIs will never work properly with ImDisk virtual drives.

This references another virtual disk tool called ImDisk which creates a Ram disk.

ehuss:
This is a fairly well known problem, #59117 discusses some of the problems along with links to many other issues.

Another option I have considered is using [`GetFullPathNameW`](https://docs.microsoft.com/en-us/windows/win32/api/fileapi/nf-fileapi-getfullpathnamew).  I haven't tested that in depth to know what issues there would be.  I think long-path handling might be tricky.  I'm not sure if it would make sense for `canonicalize` to conditionally return device-style paths with `//?/`.  I suspect that would lead to more unexpected bugs, since many programs can't handle them properly.

[...]

PR #89174, opened 2021-09-22
Automatically convert paths to verbatim for filesystem operations that support it

### LINKED ISSUE #32689: Automatically insert `\\?\` where needed on Windows
`libstd` should automatically insert `\\?\` when necessary.  The path must be canonicalized first; this can be done using the Windows API function `GetFullPathNameW`.

A necessary condition for needing to call `GetFullPathNameW` is that the path does not begin with `\\?\`.  Afterwards, if the path begins with `\\` but not `\\?\`, the `\\` must be replaced with `\\?\UNC\`.

ChrisDenton:
This allows using longer paths without the user needing to `canonicalize` or manually prefix paths. If the path is already verbatim then this has no effect.

Fixes: #32689

### DIFF
--- a/library/std/src/sys/windows/path.rs
+++ b/library/std/src/sys/windows/path.rs
@@ -150,7 +150,7 @@ fn parse_next_component(path: &OsStr, verbatim: bool) -> (&OsStr, &OsStr) {
 ///
 /// This path may or may not have a verbatim prefix.
 pub(crate) fn maybe_verbatim(path: &Path) -> io::Result<Vec<u16>> {
-    // Normally the MAX path is 260 UTF-16 code units (including the NULL).
+    // Normally the MAX_PATH is 260 UTF-16 code units (including the NULL).
     // However, for APIs such as CreateDirectory[1], the limit is 248.
     //
     // [1]: https://docs.microsoft.com/en-us/windows/win32/api/fileapi/nf-fileapi-createdirectorya#parameters
--- a/library/std/src/sys/windows/path.rs
+++ b/library/std/src/sys/windows/path.rs
@@ -150,7 +150,11 @@ fn parse_next_component(path: &OsStr, verbatim: bool) -> (&OsStr, &OsStr) {
 ///
 /// This path may or may not have a verbatim prefix.
 pub(crate) fn maybe_verbatim(path: &Path) -> io::Result<Vec<u16>> {
-    const LEGACY_MAX_PATH: usize = 260;
+    // Normally the MAX path is 260 UTF-16 code units (including the NULL).
+    // However, for APIs such as CreateDirectory[1], the limit is 248.
+    //
+    // [1]: https://docs.microsoft.com/en-us/windows/win32/api/fileapi/nf-fileapi-createdirectorya#parameters
+    const LEGACY_MAX_PATH: usize = 248;
     // UTF-16 encoded code points, used in parsing and building UTF-16 paths.
     // All of these are in the ASCII range so they can be cast directly to `u16`.
     const SEP: u16 = b'\\' as _;
--- a/library/std/src/sys/windows/path.rs
+++ b/library/std/src/sys/windows/path.rs
@@ -1,6 +1,10 @@
+use super::{c, fill_utf16_buf, to_u16s};
 use crate::ffi::OsStr;
+use crate::io;
 use crate::mem;
+use crate::path::Path;
 use crate::path::Prefix;
+use crate::ptr;
 
 #[cfg(test)]
 mod tests;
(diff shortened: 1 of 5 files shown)


ChrisDenton:
I'm not sure which labels I can set but I'll try...

@rustbot label +O-windows +T-libs

library/std/src/sys/windows/path.rs
```diff
@@ -1,13 +1,34 @@
...
 
 #[cfg(test)]
 mod tests;
 
 pub const MAIN_SEP_STR: &str = "\\";
 pub const MAIN_SEP: char = '\\';
 
+// UTF-16 encoded code points, used in parsing and building UTF-16 paths.
+// All of these are in the ASCII range so they can be cast directly to `u16`.
+const SLASH: u16 = b'\\' as _;
```
hkratz, line 17: Nit: Maybe rename `SLASH` to `BACKSLASH`and move those constants into `to_verbatim()` as they are not used anywhere else?

library/std/src/sys/windows/path.rs
```diff
@@ -141,3 +162,55 @@ fn parse_next_component(path: &OsStr, verbatim: bool) -> (&OsStr, &OsStr) {
...
 }
+
+/// Gets an absolute path with a verbatim prefix.
+///
+/// This allows bypassing the legacy `MAX_PATH` limits.
+pub(crate) fn to_verbatim(path: &Path) -> io::Result<Vec<u16>> {
+    let mut path = to_u16s(path)?;
+
+    // Early return for paths that are already verbatim.
+    if path.starts_with(VERBATIM_PREFIX) || path.starts_with(NT_PREFIX) {
```
hkratz, line 173: Device paths starting with `\\.\` should be treated specially as well here and below

klensy:
This will give additional `GetFullPathNameW` for some calls, but we can't check perf, until there will be windows machine for perf :-( Hope some day Rust Foundation will find few $ (and time) to build it.

hkratz:
> This will give additional `GetFullPathNameW` for some calls, but we can't check perf, until there will be windows machine for perf :-( Hope some day Rust Foundation will find few $ (and time) to build it.

That could be easily remedied by using a shortcut for short paths where the whole machinery is not needed.

library/std/src/sys/windows/path.rs
```diff
@@ -141,3 +162,55 @@ fn parse_next_component(path: &OsStr, verbatim: bool) -> (&OsStr, &OsStr) {
...
+    }
+
+    // Firstly, get the absolute path using `GetFullPathNameW`.
+    // https://docs.microsoft.com/en-us/windows/win32/api/fileapi/nf-fileapi-getfullpathnamew
+    let lpfilename = path.as_ptr();
+    fill_utf16_buf(
+        // SAFETY: `fill_utf16_buf` ensures the `buffer` and `size` are valid.
+        // `lpfilename` is a pointer to a null terminated string that is not
+        // invalidated until after `GetFullPathNameW` returns successfully.
+        |buffer, size| unsafe { c::GetFullPathNameW(lpfilename, size, buffer, ptr::null_mut()) },
```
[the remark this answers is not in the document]
hkratz: Hmm, according to [this stackoverflow answer](https://stackoverflow.com/questions/38036943/getfullpathnamew-and-long-windows-file-paths) and the [this Google Blog](https://googleprojectzero.blogspot.com/2016/02/the-definitive-guide-on-win32-to-nt.html) it should work though.

library/std/src/sys/windows/path.rs
ChrisDenton: Hm, how about something like `SEP`? I'm keen to keep it short because I think it's easier to read when used repeatedly.

library/std/src/sys/windows/path.rs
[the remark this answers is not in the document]
ChrisDenton: I think that's automatic boiler plate. I've tested on Windows 7 to Windows 11 and none enforce `MAX_PATH`. I'm pretty sure it would be breaking code to change it now.

If `GetFullPathNameW()` cannot be used without `\\?\` then it would defeat the purpose of using it here. So it will be necessary to reimplement it in Rust.

library/std/src/sys/windows/path.rs
ChrisDenton: But device paths are also limited to `MAX_PATH`, so surely they need to be handled too?

ChrisDenton:
> This will give additional `GetFullPathNameW` for some calls, but we can't check perf, until there will be windows machine for perf :-( Hope some day Rust Foundation will find few $ (and time) to build it.

Indeed. However, Windows internally has to do the equivalent to get an NT path from a Win32 path. Admittedly it's code is probably better optimized.

> That could be easily remedied by using a shortcut for short paths where the whole machinery is not needed.

Yes, this could be done for absolute paths. Though relative paths would still need to made absolute first.

library/std/src/sys/windows/path.rs
hkratz: `SEP` would be fine.

library/std/src/sys/windows/path.rs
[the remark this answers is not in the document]
ChrisDenton: > Maybe add a comment that this is not according to doc but should work (TM).

Oh sorry, missed this as I was typing my last comment. Yes, this seems like a good idea. And I do have a test case that explicitly checks for long paths.

library/std/src/sys/windows/path.rs
hkratz: With the current code `\\.\COM1` is converted to `\\?\UNC\.\COM1`, similarly `\\.\BootPartition\` to `\\?\UNC\.\BootPartition`. I don't have a Windows machine nearby right now to test. Do paths like this work?

ChrisDenton: I thought I had a test to make sure that worked correctly? E.g.  `check(r"\\.\CON", r"\\?\CON");`

ChrisDenton:
I've added a shortcut for absolute paths so that they can bypass the `GetFullPathNameW` call if they are short enough. I also changed the function name because this now means it won't always return a verbatim path.

library/std/src/sys/windows/path.rs
hkratz: Ah, you are right of course, sorry I missed that. Looks good to me.

hkratz:
> I've added a shortcut for absolute paths so that they can bypass the `GetFullPathNameW` call if they are short enough. I also changed the function name because this now means it won't always return a verbatim path.

I may be missing something obvious here but aren't short relative paths fine as well?

ChrisDenton:
No, the length of the current directory matters too. Example program:

```rust
use std::fs;
use std::env;
use std::io::Result;

fn main() -> Result<()> {
    // Set the current directory to some long path that exists and is just short of the 259 limit.
    let dir = r"Z:\aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa\abcdefghi";
    env::set_current_dir(dir)?;

    // This is successful.
    create_file("a")?;

    // This fails.
    create_file("abc")?;

    Ok(())
}

fn create_file(name: &str) -> Result<()> {
    println!("Creating `{}`... ", name);
    fs::File::create(name)?;
    println!("Success");
    Ok(())
}
```

Output:

```plain
Creating `a`
Creating `abc`
Error: Os { code: 3, kind: NotFound, message: "The system cannot find the path specified." }
```

hkratz:
> No, the length of the current directory matters too. Example program: [...]

Ugh, OK. That is worse than I remembered. Thanks for hashing all this out!

library/std/src/sys/windows/path.rs
```diff
@@ -141,3 +145,87 @@ fn parse_next_component(path: &OsStr, verbatim: bool) -> (&OsStr, &OsStr) {
...
+    if path.starts_with(VERBATIM_PREFIX) || path.starts_with(NT_PREFIX) {
+        // Early return for paths that are already verbatim.
+        return Ok(path);
+    } else if path.len() < LEGACY_MAX_PATH {
+        // Early return if an absolute path is less < 260 UTF-16 code units.
+        // This is an optimization to avoid calling `GetFullPathNameW` unnecessarily.
+        // Matches D:, D:\, \\
+        if let [_, COLON, 0] | [_, COLON, SEP, ..] | [SEP, SEP, ..] = path.as_slice() {
+            return Ok(path);
+        }
```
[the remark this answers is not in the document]
ChrisDenton: Yeah, I should probably match both.

There's also another edge case: `\:\`. For path parsing purposes, any UTF-16 code unit (aka `u16`) is considered a valid drive letter, except for the two path separators.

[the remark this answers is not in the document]
ChrisDenton: Ok I squished my commits but the only code that's changed is the absolute test, which is now a little bit more complicated but hopefully still readable.

dmitry-zakablukov:
Looking forward for this pr to be merged! It will help us to avoid changing all of our code to support long paths on Windows.
@m-ou-se , @hkratz , is there anything we can do to speed up merging process?

dtolnay (approved):

ChrisDenton:
I think this should be marked as relnotes? It's a pretty big deal for Windows applications because it allows supporting long paths by default and avoids some of the reasons for using `canonicalize` ([which can fail](https://github.com/rust-lang/rust/issues/79449)). The latter reason in particular is why I think this could do with a brief mention in the release notes.

@rustbot label +relnotes

PR #89176, opened 2021-09-22
Change singular to plural

TaKO8Ki:


compiler/rustc_infer/src/infer/error_reporting/mod.rs
```diff
@@ -7,7 +7,7 @@
 //! inference graph arose so that we can explain to the user what gave
 //! rise to a particular error.
 //!
-//! The basis of the system are the "origin" types. An "origin" is the
+//! The bases of the system are the "origin" types. An "origin" is the
```
mbartlett21, line 10: Perhaps it would be clearer not using the plural of basis, instead rewording the sentence:

```suggestion
//! The system is based around using "origin" types. An "origin" is the
```

TaKO8Ki: That's sounds good!

PR #89182, opened 2021-09-22
Simplify explicit request check and allow to run "doc src/librustdoc" even without config set

GuillaumeGomez:
Originally I wanted to allow the command `doc src/librustdoc` to work when passed explicitly but then @Mark-Simulacrum recommended me to generalize it, so here we are!

r? @Mark-Simulacrum

### DIFF
--- a/src/bootstrap/builder.rs
+++ b/src/bootstrap/builder.rs
@@ -1617,6 +1617,22 @@ impl<'a> Builder<'a> {
         // Only execute if it's supposed to run as default
         if desc.default && should_run.is_really_default() { self.ensure(step) } else { None }
     }
+
+    /// Checks if any of the "should_run" paths is in the `Builder` paths.
+    pub(crate) fn was_invoked_explicitly<S: Step>(&'a self) -> bool {
+        let desc = StepDescription::from::<S>();
+        let should_run = (desc.should_run)(ShouldRun::new(self));
+
+        for path in &self.paths {
+            if should_run.paths.iter().any(|s| s.has(path))
+                && !desc.is_excluded(self, &PathSet::Suite(path.clone()))
+            {
+                return true;
+            }
+        }
+
+        false
+    }
 }
 
 #[cfg(test)]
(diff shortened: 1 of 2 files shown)


src/bootstrap/builder.rs
```diff
@@ -1617,6 +1617,22 @@ impl<'a> Builder<'a> {
...
+
+    /// Checks if any of the "should_run" paths is in the `Builder` paths.
+    pub(crate) fn was_invoked_explicitly<S: Step>(
+        &'a self,
+    ) -> bool {
+        let desc = StepDescription::from::<S>();
+        let should_run = (desc.should_run)(ShouldRun::new(self));
+
+        for path in self.paths.iter() {
+            if should_run.paths.iter().any(|s| s.has(path)) {
```
Mark-Simulacrum, line 1629: This should have the check for desc.excluded too, right? Even if that's a pretty "weird" thing to do (`--exclude foo foo`)

GuillaumeGomez: "A bit" yes. Adding it anyway haha.

GuillaumeGomez:
Updated!

src/bootstrap/builder.rs
```diff
@@ -1617,6 +1617,22 @@ impl<'a> Builder<'a> {
...
     }
+
+    /// Checks if any of the "should_run" paths is in the `Builder` paths.
+    pub(crate) fn was_invoked_explicitly<S: Step>(&'a self) -> bool {
+        let desc = StepDescription::from::<S>();
+        let should_run = (desc.should_run)(ShouldRun::new(self));
+
+        for path in &self.paths {
+            if should_run.paths.iter().any(|s| s.has(path))
+                || desc.is_excluded(self, &PathSet::Suite(path.clone()))
```
Mark-Simulacrum, line 1628:
```suggestion
              && !desc.is_excluded(self, &PathSet::Suite(path.clone()))
```

And squash this in? Unless I'm missing something...

Mark-Simulacrum:
r=me with last nit fixed (or telling me I got the comment wrong) and squashed in

src/bootstrap/builder.rs
GuillaumeGomez: Good catch! I was too tired I guess...

PR #89183, opened 2021-09-22
Move encode_metadata out of CrateStore.

cjgillot:
`rustc_metadata` is already accessible by all client crates. It does not need to be called trough a trait object.


bjorn3:
Can you move EncodedMetadata to rustc_metadata?

compiler/rustc_codegen_cranelift/scripts/filter_profile.rs
```diff
@@ -96,7 +96,7 @@ fn main() -> Result<(), Box<dyn std::error::Error>> {
             stack = &stack[..index + REPORT_SYMBOL_NAMES.len()];
         }
 
-        const ENCODE_METADATA: &str = "rustc_middle::ty::context::TyCtxt::encode_metadata";
+        const ENCODE_METADATA: &str = "rustc_metadata::encode_metadata";
```
bjorn3, line 99:
```suggestion
        const ENCODE_METADATA: &str = "rustc_metadata::rmeta::encoder::encode_metadata";
```

Mark-Simulacrum:
@bors try @rust-timer queue

Calls through a trait object *might* be saving us some codegen time and such (if things are not getting inlined, etc.). Seems unlikely, but worth checking, I think.

oli-obk:
the change lgtm, but considering it's just a redesign, but regresses perf, should we really do this? Do you need this for some other PRs?

PR #89198, opened 2021-09-23
rustdoc: Don't show hidden trait methods

### LINKED ISSUE #89186: Rustdoc does not hide implementations of `doc(hidden)` trait methods
I tried this code:

```rust
pub trait Trait {
    #[doc(hidden)]
    fn f();
}
pub struct S;
impl Trait for S {
    fn f() {}
}
```

I expected to see this happen: Rustdoc does not show `f` in the documentation for `S`.

Instead, this happened: 
image

Since `f` is properly hidden in the documentation for `Trait`, the link for `f` is broken.

### Meta

`rustc --version --verbose`:
```
rustc 1.57.0-nightly (ac2d9fc50 2021-09-21)
```

_Originally posted by @rust-log-analyzer in https://github.com/rust-lang/rust/issues/89167#issuecomment-924613564_

hkmatsumoto:
Fix #89186.

By skipping trait items whose attributes include `hidden`, we avoid showing such trait methods.

### DIFF
--- a/src/librustdoc/clean/inline.rs
+++ b/src/librustdoc/clean/inline.rs
@@ -400,7 +400,7 @@ crate fn build_impl(
                 .filter(|item| {
                     // Filter out impl items whose corresponding trait item has `doc(hidden)`
                     // not to document such impl items.
-                    // For inherent impls, we don't do any filtering.
+                    // For inherent impls, we don't do any filtering, because that's already done in strip_hidden.rs.
 
                     // When `--document-hidden-items` is passed, we don't
                     // do any filtering, too.
--- a/src/librustdoc/clean/inline.rs
+++ b/src/librustdoc/clean/inline.rs
@@ -389,13 +389,45 @@ crate fn build_impl(
         }
     }
 
+    let document_hidden = cx.render_options.document_hidden;
     let predicates = tcx.explicit_predicates_of(did);
     let (trait_items, generics) = match impl_item {
         Some(impl_) => (
             impl_
                 .items
                 .iter()
-                .map(|item| tcx.hir().impl_item(item.id).clean(cx))
+                .map(|item| tcx.hir().impl_item(item.id))
+                .filter(|item| {
+                    // Filter out impl items whose corresponding trait item has `doc(hidden)`
+                    // not to document such impl items.
+                    // For inherent impls, we don't do any filtering.
+
+                    // When `--document-hidden-items` is passed, we don't
+                    // do any filtering, too.
+                    if document_hidden {
+                        return true;
+                    }
+                    if let Some(associated_trait) = associated_trait {
+                        let assoc_kind = match item.kind {
+                            hir::ImplItemKind::Const(..) => ty::AssocKind::Const,
+                            hir::ImplItemKind::Fn(..) => ty::AssocKind::Fn,
+                            hir::ImplItemKind::TyAlias(..) => ty::AssocKind::Type,
+                        };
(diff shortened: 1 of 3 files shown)


src/librustdoc/clean/inline.rs
```diff
@@ -395,7 +395,14 @@ crate fn build_impl(
             impl_
                 .items
                 .iter()
-                .map(|item| tcx.hir().impl_item(item.id).clean(cx))
+                .filter_map(|item_ref| {
+                    let item = tcx.hir().impl_item(item_ref.id);
+                    if tcx.has_attr(item.def_id.to_def_id(), sym::hidden) {
+                        Some(item.clean(cx))
```
Kobzol, line 401: Isn't this the other way around? :thinking:

hkmatsumoto: You're right, lol
What's funny is that the change was working well and what's weird is that the change after making it inverted keeps working well...

Kobzol: Maybe it's a compilation issue? I recently had a problem where my `rustdoc` changes weren't being reflected, because I was using a wrong build command I think.

jyn514 (changes requested):
Can you add a test for this? It would go in src/test/rustdoc, there are instructions at https://rustc-dev-guide.rust-lang.org/rustdoc-internals.html#dotting-is-and-crossing-ts :)

The code looks great though!

hkmatsumoto:
2021-09-24-003435_7680x2160_scrot
Documenting with `--document-hidden-items` the indent is broken, and I'm little clueless what's causing this; do you have any idea?

src/librustdoc/clean/inline.rs
```diff
@@ -395,7 +395,31 @@ crate fn build_impl(
             impl_
                 .items
                 .iter()
-                .map(|item| tcx.hir().impl_item(item.id).clean(cx))
+                .map(|item| tcx.hir().impl_item(item.id))
+                .filter(|item| {
+                    // Filter out impl items whose corresponding trait item has `doc(hidden)`.
```
jyn514, line 400: I think you need to add `if cx.document_hidden { return true; }`.

jyn514: Can you also add a test for // document-hidden-items at the same time?

hkmatsumoto: 2021-09-25-160757_7680x2160_scrot
Checking if `--document-hidden-items` passed did the trick :100:

src/librustdoc/clean/inline.rs
```diff
@@ -389,13 +389,45 @@ crate fn build_impl(
...
         Some(impl_) => (
             impl_
                 .items
                 .iter()
-                .map(|item| tcx.hir().impl_item(item.id).clean(cx))
+                .map(|item| tcx.hir().impl_item(item.id))
+                .filter(|item| {
+                    // Filter out impl items whose corresponding trait item has `doc(hidden)`
+                    // not to document such impl items.
+                    // For inherent impls, we don't do any filtering.
```
jyn514, line 403:
```suggestion
                    // For inherent impls, we don't do any filtering, because that's already done in strip_hidden.rs.
```

jyn514 (approved):
This is great, thank you! r=me with or without the nit (please use `r=jyn514`, not r+).

jyn514:
@hkmatsumoto thank you for fixing this, it's awesome work!

Issue #51430, opened 2018-06-08
Many `From` implementations are undocumented


skade:
While `std::convert::From` has [very detailed high level documentation](https://doc.rust-lang.org/std/convert/trait.From.html), the implementations often lack docs on how they work and use the default "performs the conversion" boilerplate.

This at least includes all `String` methods, also, all integer conversions. Both of these can have interesting behaviour, especially in relation to each other. For example, the different allocation behaviour between `From<Box<str>>` and `From<&'a str>` is notable and hard to figure out from the source, as it goes through three layers of indirection.

skade:
## Tutorial

To fix this issue, we have to go through the whole rustc codebase
and document all `impl From` implementations.

[`std::convert::From`](https://doc.rust-lang.org/std/convert/trait.From.html) represents the
possible conversion from one type to another. This conversion is _usually_ cheap,
but the exact nature of it is not specified. Currently, most implementations
in the std lib are using the standard documentation "performs the conversion".

The goal is to improve this situation.

## Picking a task

This assumes some basic git knowledge. If you need help with git, [this short tutorial](https://guides.github.com/activities/hello-world/)
covers all you need.

1. Pick a file from the rustc source that has undocumented From implementations.
2. Write a comment here, mentioning @skade that you want to work on this file.
  - This is to ensure that you don't do double work
3. Download rust-lang/rust and create a branch
4. Document the `from` function like this:

```rust
impl From<Foo> for Bar {
    /// Converts a `Foo` into a `Bar`.
    /// This conversion does not allocate.
    fn from(f: Foo) -> Bar {
        //....
    }
}
```

5. Post a [pull request](https://help.github.com/articles/about-pull-requests/) and add a link to it to this issue.

## What to document

Have a close look at the implementation and try to find out what it does.
Write that down. The implementation might be spread and refer to other functions,
follow as deep as you need. Write down everything you learn in the documentation.

Keep an eye out for:
* Heap allocations ("The result is allocated on the heap" [or not])
* Give a rough estimate of the cost. ("The conversion copies the data" or "The conversion happens in place")
* Is the conversion _free_? (yes, this may happen and should be documented)

Don't feel bad if the resulting text feels formulaic or repetetive, that's very
usual in documentation.

If you have some brain cycles free, try to find a clear, understandable wording.
But don't overthink it, we will improve during review together.

If you are confused or stuck, please get back here as well. This is what
the mentor can help with.

skade:
[...]

PR #89199, opened 2021-09-23
Add docs to Box conversions

timClicks:
Part of #51430.


library/alloc/src/boxed.rs
```diff
@@ -1330,6 +1331,10 @@ impl<T: Copy> From<&[T]> for Box<[T]> {
 #[cfg(not(no_global_oom_handling))]
 #[stable(feature = "box_from_cow", since = "1.45.0")]
 impl<T: Copy> From<Cow<'_, [T]>> for Box<[T]> {
+    /// Converts a `Cow<'_, [T]>` into a `Box<[T]>`
+    ///
+    /// This conversion allocates on the heap
+    /// and performs a copy of `s`.
```
m-ou-se, line 1337: > This conversion allocates on the heap and performs a copy of `s`.

- It only copies into a new allocation when the slice was borrowed. If it was owned, it uses the `From<Vec<T>> for Box<[T]>` implementation, which doesn't always reallocate.

- What does `s` refer to?

library/alloc/src/boxed.rs
```diff
@@ -1361,6 +1367,27 @@ impl From<&str> for Box<str> {
 #[cfg(not(no_global_oom_handling))]
 #[stable(feature = "box_from_cow", since = "1.45.0")]
 impl From<Cow<'_, str>> for Box<str> {
+    /// Converts a `Cow<'_, str>` into a `Box<str>`
+    ///
+    /// This conversion allocates on the heap
+    /// and performs a copy of `cow`.
```
m-ou-se, line 1373: Same here. It doesn't necesarily allocate or copy anything for a `Cow::Owned`.

library/alloc/src/boxed.rs
timClicks: > It only copies into a new allocation when the slice was borrowed. If it was owned, it uses the From<Vec<T>> for Box<[T]> implementation, which doesn't always reallocate.

I was spent too much time thinking about this. Ended up convincing myself that an extra allocation was required. Will fix up.

timClicks: > What does s refer to?

Copied the message from a pre-existing example too quickly.

library/alloc/src/boxed.rs
timClicks: Fixed in e41d5ae65c0f51ca459064808961e6d59de6f6c4

library/alloc/src/boxed.rs
timClicks: Fixed in e41d5ae65c0f51ca459064808961e6d59de6f6c4

library/alloc/src/boxed.rs
```diff
@@ -1330,6 +1331,12 @@ impl<T: Copy> From<&[T]> for Box<[T]> {
 #[cfg(not(no_global_oom_handling))]
 #[stable(feature = "box_from_cow", since = "1.45.0")]
 impl<T: Copy> From<Cow<'_, [T]>> for Box<[T]> {
+    /// Converts a `Cow<'_, [T]>` into a `Box<[T]>`
+    ///
+    /// When `cow` is the `Cow::Borrowed` variant, this
+    /// conversion allocates on the heap and performs a the
+    /// underlying `str`. Otherwise, it re-used the owned
+    /// string.
```
SkiFire13, line 1339: There's a missing "copy" and a typo in "used". "string" is also wrong here (probably copied from the `Box<str>` docs?).

Also, technically it's not guaranteed to reuse the `Vec`'s allocation since it will end up calling `Vec::into_boxed_slice`, which will reallocate if the vec has spare capacity. It might, but it's not guaranteed.

```suggestion
    /// When `cow` is the `Cow::Borrowed` variant, this
    /// conversion allocates on the heap and performs a copy the
    /// underlying `str`. Otherwise, it will try to re-use the owned
    /// vec's allocation.
```

library/alloc/src/boxed.rs
```diff
@@ -1361,6 +1369,29 @@ impl From<&str> for Box<str> {
 #[cfg(not(no_global_oom_handling))]
 #[stable(feature = "box_from_cow", since = "1.45.0")]
 impl From<Cow<'_, str>> for Box<str> {
+    /// Converts a `Cow<'_, str>` into a `Box<str>`
+    ///
+    /// When `cow` is the `Cow::Borrowed` variant, this
+    /// conversion allocates on the heap and performs a the
+    /// underlying `str`. Otherwise, it re-used the owned
+    /// string.
```
SkiFire13, line 1377: Same as above.

```suggestion
    /// When `cow` is the `Cow::Borrowed` variant, this
    /// conversion allocates on the heap and performs a copy the
    /// underlying `str`. Otherwise, it try to re-use the owned
    /// string's allocation.
```

library/alloc/src/boxed.rs
```diff
@@ -1416,6 +1448,14 @@ impl<T, const N: usize> From<[T; N]> for Box<[T]> {
...
     type Error = Box<[T]>;
 
+    /// Attempts to convert a `Box<[T]>` into a `Box<[T; N]>`.
+    ///
+    /// The conversion occurs in-place and does not require a
+    /// new memory allocation.
+    ///
+    /// # Errors
+    ///
+    /// Returns an error if `boxed_slice.len()`  does not equal `N`.
```
SkiFire13, line 1458: It doesn't return an error, but the old `Box<[T]>`. There's also a double space after "`boxed_slice.len()`".

```suggestion
    /// Returns the old `Box<[T]>` in the `Err` variant if `boxed_slice.len()` does not equal `N`.
```

SkiFire13 (changes requested):

library/alloc/src/boxed.rs
timClicks: Thank you for your close read. Should we capitalize vec in "owned vec's allocation"?

timClicks: I should also change that to "a copy _of_ the underlying.."

library/alloc/src/boxed.rs
timClicks:
```suggestion
    /// When `cow` is the `Cow::Borrowed` variant, this
    /// conversion allocates on the heap and performs a copy of the
    /// underlying `str`. Otherwise, it try to re-use the owned
    /// string's allocation.
```

library/alloc/src/boxed.rs
timClicks:
```suggestion
    /// When `cow` is the `Cow::Borrowed` variant, this
    /// conversion allocates on the heap and performs a copy of the
    /// underlying `str`. Otherwise, it will try to re-use the owned
    /// vec's allocation.
```

library/alloc/src/boxed.rs
```diff
@@ -1330,6 +1331,12 @@ impl<T: Copy> From<&[T]> for Box<[T]> {
 #[cfg(not(no_global_oom_handling))]
 #[stable(feature = "box_from_cow", since = "1.45.0")]
 impl<T: Copy> From<Cow<'_, [T]>> for Box<[T]> {
+    /// Converts a `Cow<'_, [T]>` into a `Box<[T]>`
+    ///
+    /// When `cow` is the `Cow::Borrowed` variant, this
+    /// conversion allocates on the heap and performs a copy of the
+    /// underlying `str`. Otherwise, it will try to re-use the owned
```
jyn514, line 1338:
```suggestion
    /// underlying slice. Otherwise, it will try to re-use the owned
```

library/alloc/src/boxed.rs
```diff
@@ -1361,6 +1369,29 @@ impl From<&str> for Box<str> {
 #[cfg(not(no_global_oom_handling))]
 #[stable(feature = "box_from_cow", since = "1.45.0")]
 impl From<Cow<'_, str>> for Box<str> {
+    /// Converts a `Cow<'_, str>` into a `Box<str>`
+    ///
+    /// When `cow` is the `Cow::Borrowed` variant, this
+    /// conversion allocates on the heap and performs a copy of the
+    /// underlying `str`. Otherwise, it will try to re-use the owned
+    /// string's allocation.
```
jyn514, line 1377:
```suggestion
    /// `String`'s allocation.
```

jyn514 (changes requested):

timClicks:
Thanks for the review @jyn514. I've incorporated those suggestions.

m-ou-se:
@timClicks can you squash the commits?

PR #89202, opened 2021-09-23
Resolve infered types when complaining about unexpected call type 

estebank:
```
error[E0618]: expected function, found `{integer}`
  --> $DIR/call-block.rs:2:13
   |
LL |     let _ = {42}();
   |             ^^^^--
   |             |
   |             call expression requires function
```
instead of
```
error[E0618]: expected function, found `_`
  --> $DIR/call-block.rs:2:13
   |
LL |     let _ = {42}();
   |             ^^^^--
   |             |
   |             call expression requires function
```

### DIFF
--- a/compiler/rustc_typeck/src/check/callee.rs
+++ b/compiler/rustc_typeck/src/check/callee.rs
@@ -350,6 +350,7 @@ impl<'a, 'tcx> FnCtxt<'a, 'tcx> {
                     }
                 }
 
+                let callee_ty = self.resolve_vars_if_possible(callee_ty);
                 let mut err = type_error_struct!(
                     self.tcx.sess,
                     callee_expr.span,
--- a/src/test/ui/typeck/call-block.rs
+++ b/src/test/ui/typeck/call-block.rs
@@ -1,0 +1,3 @@
+fn main() {
+    let _ = {42}(); //~ ERROR expected function, found `_`
+}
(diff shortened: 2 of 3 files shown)


compiler/rustc_typeck/src/check/callee.rs
```diff
@@ -350,6 +350,7 @@ impl<'a, 'tcx> FnCtxt<'a, 'tcx> {
                     }
                 }
 
+                let callee_ty = self.resolve_vars_if_possible(callee_ty);
```
oli-obk, line 353: I wonder why callee_ty is an TyVar and not an IntVar, I think we should dig into this a bit instead of just patching up the diagnostic

estebank: I'm guessing the `TyVar` is coming from the `{}` block's tail expression not having unified yet?

oli-obk: If that were the case, wouldn't we also see the tyvar, when using a function type?

I think we're doing something odd when type checking the literal expression.

I'm currently somewhere deep in that code anyway, I'll check and post my findings

oli-obk: Yea, you're right. The tail block expression will fall back to a type variable in case we don't know anything about the type. This happens in https://github.com/rust-lang/rust/blob/3982ac22497efda2bddf261021e88aa37b55e47b/compiler/rustc_typeck/src/check/expectation.rs#L119

So... I guess we can't avoid this at creation time. oh well...

PR #89203, opened 2021-09-23
Clean up clean/types.rs file by making the implementations follow the type declaration

GuillaumeGomez:
This PR doesn't change anything, it simply moves things around: when reading the code, I realized a few times that a type declaration and implementations on it might be separated by some other type declarations, which makes the reading much more complicated. I put back impl and declaration together.

r? @camelid

### DIFF
--- a/src/librustdoc/clean/types.rs
+++ b/src/librustdoc/clean/types.rs
@@ -908,18 +908,10 @@ impl<'a> FromIterator<&'a DocFragment> for String {
     }
 }
 
-/// The attributes on an [`Item`], including attributes like `#[derive(...)]` and `#[inline]`,
-/// as well as doc comments.
-#[derive(Clone, Debug, Default)]
-crate struct Attributes {
-    crate doc_strings: Vec<DocFragment>,
-    crate other_attrs: Vec<ast::Attribute>,
-}
-
-#[derive(Clone, Debug, PartialEq, Eq, Hash)]
 /// A link that has not yet been rendered.
 ///
 /// This link will be turned into a rendered link by [`Item::links`].
+#[derive(Clone, Debug, PartialEq, Eq, Hash)]
 crate struct ItemLink {
     /// The original link written in the markdown
     pub(crate) link: String,
@@ -942,6 +934,14 @@ pub struct RenderedLink {
     pub(crate) new_text: String,
     /// The URL to put in the `href`
     pub(crate) href: String,
+}
+
+/// The attributes on an [`Item`], including attributes like `#[derive(...)]` and `#[inline]`,
+/// as well as doc comments.
+#[derive(Clone, Debug, Default)]
+crate struct Attributes {
+    crate doc_strings: Vec<DocFragment>,
+    crate other_attrs: Vec<ast::Attribute>,
 }
 
 impl Attributes {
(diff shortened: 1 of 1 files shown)


src/librustdoc/clean/types.rs
```diff
@@ -1301,19 +1301,19 @@ crate struct Arguments {
...
     SelfBorrowed(Option<Lifetime>, Mutability),
     SelfExplicit(Type),
 }
 
+#[derive(Clone, PartialEq, Eq, Debug, Hash)]
+crate struct Argument {
+    crate type_: Type,
+    crate name: Symbol,
+}
+
```
camelid, line 1316: Can you leave this part as it was before? Both `Argument` and `SelfTy` are logically connected to the following impl, so this feels like unnecessary churn. I'm likely going to be removing `SelfTy` soon anyway.

src/librustdoc/clean/types.rs
```diff
@@ -1375,55 +1375,6 @@ crate struct PolyTrait {
     crate generic_params: Vec<GenericParamDef>,
 }
 
-/// A representation of a type suitable for hyperlinking purposes. Ideally, one can get the original
-/// type out of the AST/`TyCtxt` given one of these, if more information is needed. Most
-/// importantly, it does not preserve mutability or boxes.
-#[derive(Clone, PartialEq, Eq, Debug, Hash)]
-crate enum Type {
```
camelid, line 1382: Moving `Type` and its impls will cause a **lot** of conflicts with my "Cleanup `clean`" PR and probably other PRs as well. Can you please revert this part of the change? I'm okay with keeping the merging of the two inherent impls on `Type` (the general one and the `inner_def_id` one), and maybe also moving `trait GetDefId { ... }` so it's not in the middle of the `Type` code, but most of the other changes feel like unnecessary churn to me.

camelid (changes requested):
There's a cost-benefit ratio to moving code: the cost of the Git churn (primarily merge conflicts, but also making Git blame more annoying to use), and the benefit of the new version of the code.

Some of the changes here feel to me like the benefit outweighs the cost (e.g., for small sections of code, like `Attributes`, that also aren't changed that often), while others feel like the cost outweighs the benefit (e.g., moving `Type` and its impls, which are large amounts of code that change somewhat frequently, and which I'm currently in the process of refactoring).

I left comments on the main parts where I feel the cost outweighs the benefit and asked to revert them.

src/librustdoc/clean/types.rs
GuillaumeGomez: Oh nice!

src/librustdoc/clean/types.rs
GuillaumeGomez: Ok!

GuillaumeGomez:
Done!

src/librustdoc/clean/types.rs
```diff
@@ -1608,6 +1575,37 @@ impl GetDefId for Type {
     }
 }
 
+#[derive(Clone, PartialEq, Eq, Hash, Copy, Debug)]
+/// N.B. this has to be different from `hir::PrimTy` because it also includes types that aren't
+/// paths, like `Unit`.
```
camelid, line 1580: While you're at it, and since you did this for `ItemLink`, can you move the `derive` to after the docs?

camelid: After that, I'll take another look at the other changes, and then I think this should be good to go :)

GuillaumeGomez: Yes. :)

camelid:
Thanks!

@bors r+ rollup=iffy

(iffy because it has some potential for merge conflicts)

jyn514:
@camelid bors will automatically keep the PR out of the rollup if it conflicts with other PRs, so there's no need to mark it as iffy.

@bors rollup=maybe

PR #89208, opened 2021-09-23
[rfc 2229] Drop fully captured upvars in the same order as the regular drop code

### LINKED ISSUE #42: Drop order affected because of use order
Consider the following example

```rust
let x : DropType;
let y : DropType;

let p = P { x, y };
let c = || {
   move(p.y);
   move(p.x);
};
```

Even though the closure captures all paths starting from `p`, without the feature the fields are dropped in order, so `p.x` and then `p.y`. However, with the feature enabled, we would drop `p.y`, and then `p.x` because that's the order (use within the closure) in which we would place them in the desugared closure structure.

wesleywiser:
Currently, with the new 2021 edition, if a closure captures all of the
fields of an upvar, we'll drop those fields in the order they are used
within the closure instead of the normal drop order (the definition
order of the fields in the type).

This changes that so we sort the captured fields by the definition order
which causes them to drop in that same order as well.

Fixes rust-lang/project-rfc-2229#42

r? @nikomatsakis

### DIFF
--- a/compiler/rustc_typeck/src/check/upvar.rs
+++ b/compiler/rustc_typeck/src/check/upvar.rs
@@ -602,7 +602,47 @@ impl<'a, 'tcx> FnCtxt<'a, 'tcx> {
             }
         }
 
-        debug!("For closure={:?}, min_captures={:#?}", closure_def_id, root_var_min_capture_list);
+        debug!(
+            "For closure={:?}, min_captures before sorting={:?}",
+            closure_def_id, root_var_min_capture_list
+        );
+
+        // Now that we have the minimized list of captures, sort the captures by field id.
+        // This causes the closure to capture the upvars in the same order as the fields are
+        // declared which is also the drop order. Thus, in situations where we capture all the
+        // fields of some type, the obserable drop order will remain the same as it previously
+        // was even though we're dropping each capture individually.
+        // See https://github.com/rust-lang/project-rfc-2229/issues/42 and
+        // `src/test/ui/closures/2229_closure_analysis/preserve_field_drop_order.rs`.
+        for (_, captures) in &mut root_var_min_capture_list {
+            captures.sort_by(|capture1, capture2| {
+                for (p1, p2) in capture1.place.projections.iter().zip(&capture2.place.projections) {
+                    match (p1.kind, p2.kind) {
+                        // Paths are the same, continue to next loop.
+                        (ProjectionKind::Deref, ProjectionKind::Deref) => {}
+                        (ProjectionKind::Field(i1, _), ProjectionKind::Field(i2, _))
+                            if i1 == i2 => {}
+
+                        // Fields are different, compare them.
+                        (ProjectionKind::Field(i1, _), ProjectionKind::Field(i2, _)) => {
+                            return i1.cmp(&i2);
+                        }
+
+                        (l, r) => bug!("ProjectionKinds were different: ({:?}, {:?})", l, r),
+                    }
+                }
+
+                unreachable!(
+                    "we captured two identical projections: capture1 = {:?}, capture2 = {:?}",
+                    capture1, capture2
+                );
+            });
+        }
+
+        debug!(
+            "For closure={:?}, min_captures after sorting={:#?}",
+            closure_def_id, root_var_min_capture_list
+        );
         typeck_results.closure_min_captures.insert(closure_def_id, root_var_min_capture_list);
     }
 
--- a/compiler/rustc_typeck/src/check/upvar.rs
+++ b/compiler/rustc_typeck/src/check/upvar.rs
@@ -617,6 +617,10 @@ impl<'a, 'tcx> FnCtxt<'a, 'tcx> {
         for (_, captures) in &mut root_var_min_capture_list {
             captures.sort_by(|capture1, capture2| {
                 for (p1, p2) in capture1.place.projections.iter().zip(&capture2.place.projections) {
+                    // We do not need to look at the `Projection.ty` fields here because at each
+                    // step of the iteration, the projections will either be the same and therefore
+                    // the types must be as well or the current projection will be different and
+                    // we will return the result of comparing the field indexes.
                     match (p1.kind, p2.kind) {
                         // Paths are the same, continue to next loop.
                         (ProjectionKind::Deref, ProjectionKind::Deref) => {}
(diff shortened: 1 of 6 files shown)


compiler/rustc_typeck/src/check/upvar.rs
```diff
@@ -602,7 +602,47 @@ impl<'a, 'tcx> FnCtxt<'a, 'tcx> {
...
+                        (ProjectionKind::Deref, ProjectionKind::Deref) => {}
+                        (ProjectionKind::Field(i1, _), ProjectionKind::Field(i2, _))
+                            if i1 == i2 => {}
+
+                        // Fields are different, compare them.
+                        (ProjectionKind::Field(i1, _), ProjectionKind::Field(i2, _)) => {
+                            return i1.cmp(&i2);
+                        }
+
+                        (l, r) => bug!("ProjectionKinds were different: ({:?}, {:?})", l, r),
```
arora-aman, line 631: a) Might be a good idea to mention why this is a bug
b) I'd prefer if we matched on `(Projection::Index | Subslice, _) | (_,  Projection::Index | Subslice) => bug!("We don't capture index/subslice)`

src/test/ui/closures/2229_closure_analysis/preserve_field_drop_order2.rs
```diff
@@ -0,0 +1,28 @@
+// run-pass
+// check-run-results
```
arora-aman, line 2: I think we should add some test cases for paths that are multilevel  eg: a.b.c and a.b.d.

src/test/ui/closures/2229_closure_analysis/preserve_field_drop_order.rs
```diff
@@ -0,0 +1,98 @@
...
+    fn drop(&mut self) {
+        println!("dropped");
+    }
+}
+
+fn test_one() {
+    let a = (HasDrop, HasDrop);
+    let b = (HasDrop, HasDrop);
+
+    let c = #[rustc_capture_analysis]
```
arora-aman, line 20: My concern with this test is that we need to reply on checking the stderr, and I feel like if the fields order does get changed, we won't need to update the annoatations in this file and someone might ignore the stderr changes because the ouput needed is all there.

I think we should go with testing with stdout here because changes to stdout are observable behaviour to the user and might be more cause of concern in case something breaks down the line.

compiler/rustc_typeck/src/check/upvar.rs
```diff
@@ -602,7 +602,47 @@ impl<'a, 'tcx> FnCtxt<'a, 'tcx> {
...
+        // This causes the closure to capture the upvars in the same order as the fields are
+        // declared which is also the drop order. Thus, in situations where we capture all the
+        // fields of some type, the obserable drop order will remain the same as it previously
+        // was even though we're dropping each capture individually.
+        // See https://github.com/rust-lang/project-rfc-2229/issues/42 and
+        // `src/test/ui/closures/2229_closure_analysis/preserve_field_drop_order.rs`.
+        for (_, captures) in &mut root_var_min_capture_list {
+            captures.sort_by(|capture1, capture2| {
+                for (p1, p2) in capture1.place.projections.iter().zip(&capture2.place.projections) {
+                    match (p1.kind, p2.kind) {
```
arora-aman, line 620: Might be a good idea to document why we are only comparing the same type here because it might not be apparent to a future reader.

compiler/rustc_typeck/src/check/upvar.rs
```diff
@@ -602,7 +602,47 @@ impl<'a, 'tcx> FnCtxt<'a, 'tcx> {
...
+
+        // Now that we have the minimized list of captures, sort the captures by field id.
+        // This causes the closure to capture the upvars in the same order as the fields are
+        // declared which is also the drop order. Thus, in situations where we capture all the
+        // fields of some type, the obserable drop order will remain the same as it previously
+        // was even though we're dropping each capture individually.
+        // See https://github.com/rust-lang/project-rfc-2229/issues/42 and
+        // `src/test/ui/closures/2229_closure_analysis/preserve_field_drop_order.rs`.
+        for (_, captures) in &mut root_var_min_capture_list {
+            captures.sort_by(|capture1, capture2| {
```
nikomatsakis, line 618: We *could* use `sort_by_key(|c| &c.place.projections)`, right? It wouldn't detect the buggy cases, but it'd be a heck of a lot shorter!

wesleywiser: `hir::place::Projection` doesn't impl `Ord` so that won't currently work. The MIR version of `Projection` does impl Ord so that doesn't seem entirely unreasonable to do. Would you rather I do that and use `sort_by_key` or leave the code as-is? 🙂

compiler/rustc_typeck/src/check/upvar.rs
wesleywiser: I implemented both but in order to make the `match` exhaustive, it's pretty ugly 🙁

arora-aman: Little bit of duplication but this works

```
(l @ ProjectionKind::Slice | ProjectionKind::Index, r)  => bug!(Didn't expect index/slice , l, r)
(l, r  @ ProjectionKind::Slice | ProjectionKind::Index)  => bug!(Didn't expect index/slice , l, r)
(l, r) => bug!("Projections must be of same kind, found", l, r);
```

compiler/rustc_typeck/src/check/upvar.rs
```diff
@@ -602,7 +602,78 @@ impl<'a, 'tcx> FnCtxt<'a, 'tcx> {
...
+                            l @ (ProjectionKind::Deref | ProjectionKind::Field(..)),
+                            r @ (ProjectionKind::Deref | ProjectionKind::Field(..)),
+                        ) => bug!(
+                            "ProjectionKinds Deref and Field were mismatched: ({:?}, {:?})",
+                            l,
+                            r
+                        ),
+                        (
+                            l
+                            @
```
arora-aman, line 647: Not sure why we need to list all of them out explictly

compiler/rustc_typeck/src/check/upvar.rs
nikomatsakis: I'm of two minds. Thinking about it more, I thought it seems a bit *fragile* to have the runtime semantics be derived by Ord in this way. At minimum, I would want a comment. =)

compiler/rustc_typeck/src/check/upvar.rs
nikomatsakis: I am a fan of listing things explicitly, because it helps detect when new cases are added.

compiler/rustc_typeck/src/check/upvar.rs
wesleywiser: I don't think this really gets us very much benefit since we have the wildcard pattern match at the end, if new variants were to be added to `ProjectionKind`, this code would continue to compile. At the same time, we have more code than just falling back to the wildcard match which would show the `ProjectionKind` anyway. I think Niko basically raised the same point [here](https://github.com/rust-lang/rust/pull/89208#discussion_r715630997).

src/test/ui/closures/2229_closure_analysis/preserve_field_drop_order.rs
wesleywiser: I think it's useful to check both the internal results via `#[rustc_capture_analysis]` as well as the runtime results which is what these two tests do in combination. However, I totally agree with your point that as it is, it would be pretty easy for someone to change this test without realizing what parts are important and which parts aren't. I believe I can remove the checking for the notes we do not care about which should make it much more clear (this is the convention used in other ui tests as well).

nbdd0121:
Should this be backported to beta?

wesleywiser:
Added some comments to existing tests, clarified which notes are important in the `preserve_field_drop_order` test and added some new test cases in `preserve_field_drop_order2`. 

@bors r=nikomatsakis

wesleywiser:
> Should this be backported to beta?

Yes, I believe we want this to ship with the 2021 edition.

apiraino:
Beta backport accepted as per compiler team [on Zulip](https://rust-lang.zulipchat.com/#narrow/stream/238009-t-compiler.2Fmeetings/topic/.5Bweekly.5D.202021-09-30.20.2354818/near/255583604)

@rustbot label +beta-accepted

PR #89210, opened 2021-09-23
Add missing time complexities to linked_list.rs

Takashiidobe:
Most functions in LinkedList have time complexities in their description:
Like push front:

```
Adds an element first in the list.

This operation should compute in O(1) time.
```

Time complexities were missing for the following, so I've added them in this PR:

contains: O(n)
front: O(1)
front_mut: O(1)
back: O(1)
back_mut: O(1)


library/alloc/src/collections/linked_list.rs
```diff
@@ -656,6 +658,8 @@ impl<T> LinkedList<T> {
     /// Provides a reference to the front element, or `None` if the list is
     /// empty.
     ///
+    /// This operation should compute in *O*(*1*) time.
```
kennytm, line 661: the "1" doesn't need to be italic.

Takashiidobe: Unitalicized. Ready for another look when you have the time, @kennytm.

kennytm:
@Takashiidobe don't modify the PR after it has been approved. The original changes has already been rolled up into #89230. please file a new PR for fa1a87888c49b72b7963055fafe8cd22b32a7a46.

Issue #89206, opened 2021-09-23
Link a rustup toolchain automatically in `x.py setup`


Sl1mb0:
Issue tracker for linking a rustup toolchain automatically in `x.py setup`.

Mark-Simulacrum:
Can you say more about the motivation? Initially, I would say that it is pretty likely that we do not want to support this kind of "aggregation" in the x.py interface; it complicates the information and I'm not sure it improves the user interface.

Sl1mb0:
I agree that I don't think it improves the user interface much, but I think that the interface for this specific feature could be made better. My only argument for why I think this capability should be added is that it's natural to create a toolchain for the compiler immediately after having built it, so why not give `./.xpy` the capability to do so? 

However, as I understand it you only need to run `rustup` once to create a toolchain for the build, and subsequent builds will always be used by the toolchain. So I can see why this is kind of pointless.

Mark-Simulacrum:
I think streamlining the toolchain creation process is worthwhile to invest into; e.g., we could give a suggestion at the end of the first x.py build in a new build directory (with a stamp file just saying 'did that') to run rustup link .... with the appropriate path.

Sl1mb0:
I like that idea. I'll wait for more feedback before I start looking into it though.

jyn514:
Hmm, I would rather set up the toolchain automatically when running `x.py setup`, and give a warning if a toolchain named `stage1` already exists, along with the command to set it up yourself.

If we gave a warning the first time you run `x.py build`, then a) people who aren't developing locally get confused (e.g. distros building from source) because they don't actually care about this, and b) x.py has no way of knowing whether you actually add the toolchain or not; I can definitely see people missing it the first time and then never getting the warning again just because x.py is using a stamp file.

Mark-Simulacrum:
Yeah, x.py setup seems better. I think rustup won't let you do it without the sysroot somewhat existing, though:

```
$ rustup toolchain link foo foo
error: not a directory: 'foo/lib'
```

But I think we have enough in setup that we can probably create the directory temporarily...

jyn514:
Mentoring instructions: just after https://github.com/rust-lang/rust/blob/bf642323d621dcefeef1d8ab4711aae36e357615/src/bootstrap/setup.rs#L108, call `rustup toolchain link stage1 build/TARGET/stage1`, where `TARGET` is `builder.config.build`. If that errors, give a warning and suggest that the user do it themselves.

[...]

Issue #76444, opened 2020-09-07
Give a better error message when running `cargo test` on libstd itself


jyn514:
Right now, you get really terrible errors like this:
```
$ cargo test
error[E0277]: the size for values of type `()` cannot be known at compilation time
   --> /home/joshua/.local/lib/cargo/registry/src/github.com-1ecc6299db9ec823/getrandom-0.1.14/src/use_file.rs:30:44
    |
30  | pub fn getrandom_inner(dest: &mut [u8]) -> Result<(), Error> {
    |                                            ^^^^^^^^^^^^^^^^^ doesn't have a size known at compile-time
    |
    = help: the trait `Sized` is not implemented for `()`
```
162 times. It would be more helpful to instead say 'use `x.py test` instead'.

dawnFromDark:
i agree ,rust check is shit~!

jyn514:
Fixed by https://github.com/rust-lang/rust/pull/94796.

PR #89212, opened 2021-09-24
x.py: run `rustup toolchain link` in setup

Sl1mb0:
Addresses #89206  

r? @jyn514

### DIFF
--- a/src/bootstrap/setup.rs
+++ b/src/bootstrap/setup.rs
@@ -1,3 +1,4 @@
+use crate::TargetSelection;
 use crate::{t, VERSION};
 use std::fmt::Write as _;
 use std::path::{Path, PathBuf};
@@ -107,6 +108,17 @@ pub fn setup(src_path: &Path, profile: Profile) {
     let include_path = profile.include_path(src_path);
     println!("`x.py` will now use the configuration at {}", include_path.display());
 
+    let build = TargetSelection::from_user(&env!("BUILD_TRIPLE"));
+    let stage_path = ["build", build.rustc_target_arg(), "stage1"].join("/");
+
+    println!();
+
+    if !rustup_installed() && profile != Profile::User {
+        println!("`rustup` is not installed; cannot link `stage1` toolchain");
+    } else if stage_dir_exists(&stage_path[..]) {
+        attempt_toolchain_link(&stage_path[..]);
+    }
+
     let suggestions = match profile {
         Profile::Codegen | Profile::Compiler => &["check", "build", "test"][..],
         Profile::Tools => &[
@@ -137,6 +149,74 @@ pub fn setup(src_path: &Path, profile: Profile) {
             "For more suggestions, see https://rustc-dev-guide.rust-lang.org/building/suggested.html"
         );
     }
+}
+
+fn rustup_installed() -> bool {
+    Command::new("rustup")
+        .arg("--version")
+        .stdout(std::process::Stdio::null())
+        .output()
+        .map_or(false, |output| output.status.success())
+}
+
+fn stage_dir_exists(stage_path: &str) -> bool {
+    match fs::create_dir(&stage_path[..]) {
+        Ok(_) => true,
+        Err(_) => Path::new(&stage_path[..]).exists(),
+    }
+}
+
+fn attempt_toolchain_link(stage_path: &str) {
+    if toolchain_is_linked() {
+        return;
+    }
+
+    if try_link_toolchain(&stage_path[..]) {
+        println!(
+            "Added `stage1` rustup toolchain; try `cargo +stage1 build` on a separate rust project to run a newly-built toolchain"
+        );
+    } else {
+        println!("`rustup` failed to link stage 1 build to `stage1` toolchain");
+        println!(
+            "To manually link stage 1 build to `stage1` toolchain, run:\n
+            `rustup toolchain link stage1 {}`",
+            &stage_path[..]
+        );
+    }
+}
+
+fn toolchain_is_linked() -> bool {
+    match Command::new("rustup")
+        .args(&["toolchain", "list"])
+        .stdout(std::process::Stdio::piped())
+        .output()
+    {
+        Ok(toolchain_list) => {
+            if !String::from_utf8_lossy(&toolchain_list.stdout).contains("stage1") {
+                return false;
+            }
+            // The toolchain has already been linked.
+            println!(
+                "`stage1` toolchain already linked; not attempting to link `stage1` toolchain"
+            );
+        }
+        Err(_) => {
+            // In this case, we don't know if the `stage1` toolchain has been linked;
+            // but `rustup` failed, so let's not go any further.
+            println!(
+                "`rustup` failed to list current toolchains; not attempting to link `stage1` toolchain"
+            );
+        }
+    }
+    true
+}
+
+fn try_link_toolchain(stage_path: &str) -> bool {
+    Command::new("rustup")
+        .stdout(std::process::Stdio::null())
+        .args(&["toolchain", "link", "stage1", &stage_path[..]])
+        .output()
+        .map_or(false, |output| output.status.success())
 }
 
 // Used to get the path for `Subcommand::Setup`


src/bootstrap/setup.rs
```diff
@@ -107,6 +108,19 @@ pub fn setup(src_path: &Path, profile: Profile) {
     let include_path = profile.include_path(src_path);
     println!("`x.py` will now use the configuration at {}", include_path.display());
 
+    let build = TargetSelection::from_user(&env!("BUILD_TRIPLE"));
+    let stage_path = ["build", build.rustc_target_arg(), "stage1"].join("/");
+    let toolchain_link =
+        Command::new("rustup").args(&["toolchain", "link", "stage1", &stage_path[..]]).output();
+    if toolchain_link.is_err() {
```
jyn514, line 115: This is checking the wrong thing - `output()` only fails if the syscall itself fails, not if the status is non-zero.
```suggestion
    if toolchain_link.is_err() || !toolchain_link.status().success() {
```

src/bootstrap/setup.rs
```diff
@@ -107,6 +108,19 @@ pub fn setup(src_path: &Path, profile: Profile) {
     let include_path = profile.include_path(src_path);
     println!("`x.py` will now use the configuration at {}", include_path.display());
 
+    let build = TargetSelection::from_user(&env!("BUILD_TRIPLE"));
+    let stage_path = ["build", build.rustc_target_arg(), "stage1"].join("/");
```
jyn514, line 112: You need to create this directory first, otherwise it will give an error:
```
$ rustup toolchain link stage1 build/x86_64-unknown-linux-gnu/stage1
error: not a directory: 'build/x86_64-unknown-linux-gnu/stage1/lib'
```

src/bootstrap/setup.rs
```diff
@@ -107,6 +108,19 @@ pub fn setup(src_path: &Path, profile: Profile) {
...
+    let toolchain_link =
+        Command::new("rustup").args(&["toolchain", "link", "stage1", &stage_path[..]]).output();
```
jyn514, line 114: I just realized this will overwrite any existing toolchain - that seems kind of not great, it means we'll delete any existing toolchain with this name. @kinnison @rbtcollins is that intentional that it will silently overwrite?

I guess to avoid that we could explicitly run `rustup toolchain list` and see if it already has `stage1`.

src/bootstrap/setup.rs
```diff
@@ -107,6 +108,19 @@ pub fn setup(src_path: &Path, profile: Profile) {
...
+    if toolchain_link.is_err() {
+        println!("WARNING: `rustup`failed to link stage 1 build to `stage1` toolchain");
+        println!(
+            "To manually link stage 1 build to `stage1` tool chain, run:\n
+            `rustup toolchain link stage1 build/{}/stage1`",
+            build.rustc_target_arg()
+        );
+    }
```
jyn514, line 122: It would be nice to let people know x.py did this, so they know to use it in the future.
```suggestion
    } else {
        println!("Added `stage1` rustup toolchain; use `cargo +stage1 build` to run a newly-built toolchain");
    }
```

jyn514 (changes requested):

src/bootstrap/setup.rs
```diff
@@ -107,6 +108,19 @@ pub fn setup(src_path: &Path, profile: Profile) {
...
+    let build = TargetSelection::from_user(&env!("BUILD_TRIPLE"));
+    let stage_path = ["build", build.rustc_target_arg(), "stage1"].join("/");
+    let toolchain_link =
+        Command::new("rustup").args(&["toolchain", "link", "stage1", &stage_path[..]]).output();
+    if toolchain_link.is_err() {
+        println!("WARNING: `rustup`failed to link stage 1 build to `stage1` toolchain");
+        println!(
+            "To manually link stage 1 build to `stage1` tool chain, run:\n
+            `rustup toolchain link stage1 build/{}/stage1`",
+            build.rustc_target_arg()
```
workingjubilee, line 120: So... I could have missed something about the previous code, but at this point in the program, what proof do we have that the user **has** Rustup on their system that justifies emitting this message all "WARNING!" style, telling them about a binary that possibly isn't installed on their system? It is entirely possible that the user is bootstrapping Rust on their system for the first time, Rustup is not available for this target, and we should be careful to neither alarm nor distract them. Something gentler and softer-voiced seems appropriate.

src/bootstrap/setup.rs
Sl1mb0: Do you think we should also check if `rustup` has been installed and inform the user if it hasn't?

src/bootstrap/setup.rs
workingjubilee: Perhaps a command that would be less disruptive if it was executed immediately on the repository they are in, which could cause amusing errors? Or perhaps additional context to when and where they should run it. There are actually some people who first try Rust by building from the source, so we shouldn't assume they are an old hand.

src/bootstrap/setup.rs
workingjubilee: Usually this command is executed by the user, so like most Unix-y commands, dispatches as silently and swiftly as possible, and alerting someone is a sign of failure.

workingjubilee (commented):
Have we considered the possible failure mode of a distro-installed Rustup that is later overshadowed by Rustup's normal `/home/user/.{cargo,rustup}` install process? Possibly that doesn't matter: maybe such a case would make one expect possible overshadowing.

src/bootstrap/setup.rs
jyn514: Hmm - but `ln -s` (which you said in the original issue is basically what `toolchain link` does) gives an error if the file already exists:
```
$ ln -s x config.toml
ln: failed to create symbolic link 'config.toml': File exists
```

src/bootstrap/setup.rs
Sl1mb0: If rustup is installed:
```
Linking stage1 toolchain
```

If not:
```
rustup is not installed
Unable to link stage1 toolchain
```

jyn514: I think it would be fine to only warn if the profile isn't `"user"` (which is what distros should be using); for everyone else they would find rustup really useful for rust-lang/rust even if they wouldn't use it normally.

src/bootstrap/setup.rs
jyn514: > Perhaps a command that would be less disruptive if it was executed immediately on the repository they are in, which could cause amusing errors?

Ideally we would fix this with https://github.com/rust-lang/rust/issues/76444 ... but I agree in the meantime we should suggest something else. Maybe add "on a separate rust project" after the `cargo build` snippet?

jyn514:
> Have we considered the possible failure mode of a distro-installed Rustup that is later overshadowed by Rustup's normal /home/user/.{cargo,rustup} install process?

I'm not sure how this could go wrong. If there's a distro-installed rustup, it could run `toolchain link` just like a binary-installed rustup. `toolchain link` doesn't change the default, just add additional toolchains.

src/bootstrap/setup.rs
workingjubilee: Reasonable~
But the other reason I mentioned links is due to problems due to OS control over when files may be edited/clobbered/removed, which has less to do with the particular link.

src/bootstrap/setup.rs
Sl1mb0: Could change `use` to `try`

src/bootstrap/setup.rs
jyn514: Sure, doing that if the profile isn't "user" seems helpful :)

workingjubilee:
I more meant user surprisal due to the overshadowing and the linked toolchain "disappearing", if that mattered.

jyn514:
I clarified offline - the way this can go wrong is:
1. Someone has a version of rustup installed from their distro (in /usr/bin/rustup or wherever).
2. They build rust from source for *reasons* and x.py setup runs `rustup toolchain link stage1`.
3. They install a new version of rustup so they can download new toolchains.
4. The linked toolchain disappears.

Given that in this (somewhat convoluted scenario) they'll be able to easily rerun `toolchain link` or just run `rustup install`, I'm ok with the toolchain disappearing in that case; I don't see a way to avoid it and the benefit here massively outweighs the risk IMO.

src/bootstrap/setup.rs
workingjubilee: It's mostly about framing: gently observing that they do so is fine, as long as it doesn't appear on every invocation after, if they decide they don't want to use it.

jyn514: @workingjubilee yup - this is only running once when you first run `setup`, it's never repeated.

src/bootstrap/setup.rs
Sl1mb0: Since we have to create the directory in the first place, couldn't we just not link if it already exists?

jyn514: > Since we have to create the directory in the first place, couldn't we just not link if it already exists?

That doesn't seem great - that won't warn or setup the toolchain if the user has already run `x.py build` but not created a rustup toolchain.

Sl1mb0: I think this is why @Mark-Simulacrum suggested using a stamp file. We could also extend this by adding configuration for it in the `config.toml`; e.g. `toolchain-overwrite` or something where the user can specify if they want to overwrite the toolchain every time they run `x.py setup` or not, and by default `./x.py setup` will only do it once.

jyn514: I really don't think this needs additional configuration. We can just check if `rustup toolchain list` contains `stage1`.

src/bootstrap/setup.rs
Sl1mb0: Is there a best practice for getting this information?

jyn514: The profile is passed in as an argument to `setup`.

src/bootstrap/setup.rs
```diff
@@ -110,15 +110,50 @@ pub fn setup(src_path: &Path, profile: Profile) {
...
-            "To manually link stage 1 build to `stage1` tool chain, run:\n
-            `rustup toolchain link stage1 build/{}/stage1`",
-            build.rustc_target_arg()
-        );
+
+    let created = fs::create_dir(&stage_path[..]).is_ok();
+
+    // Link `stage1` toolchain if `rustup` is installed
+    // and we have not already done so.
+    if Path::new("~/.cargo/bin/rustup").exists() && created {
```
jyn514, line 118: This is not right. It needs to check if rustup is anywhere in PATH; and ideally if it's executable. Try seeing if `rustup --version` succeeds or something like that.

src/bootstrap/setup.rs
```diff
@@ -110,15 +110,50 @@ pub fn setup(src_path: &Path, profile: Profile) {
...
+        // Let's see if we have already linked the stage1 toolchain.
+        let toolchain_list = Command::new("rustup")
+            .args(&["toolchain", "list"])
+            .stdout(std::process::Stdio::piped())
+            .output();
+
+        if toolchain_list.is_ok() {
+            let toolchain_list = toolchain_list.unwrap();
```
jyn514, line 126:
```suggestion
        if let Ok(toolchain_list) = toolchain_list {
```

src/bootstrap/setup.rs
```diff
@@ -110,15 +110,50 @@ pub fn setup(src_path: &Path, profile: Profile) {
...
+                    );
+                }
+            } else {
+                println!(
+                    "`stage1` toolchain already linked; `rustup` will not attempt to link `stage1` toolchain",
+                );
+            }
+        } else {
+            println!(
+                "`rustup` failed to list the current toolchains; `rustup` will not attempt to link `stage1` toolchain"
```
jyn514, line 152: I don't think saying "rustup" is correct here - it's x.py that's making the decision. But the output comes from x.py already so I don't think we need to mention that explicitly.
```suggestion
                "`rustup` failed to list the current toolchains; not attempting to link `stage1` toolchain"
```

src/bootstrap/setup.rs
```diff
@@ -110,15 +110,50 @@ pub fn setup(src_path: &Path, profile: Profile) {
...
+                        &stage_path[..]
+                    );
+                } else {
+                    println!(
+                        "Added `stage1` rustup toolchain; try `cargo +stage1 build` on a separate rust project to run a newly-built toolchain"
+                    );
+                }
+            } else {
+                println!(
+                    "`stage1` toolchain already linked; `rustup` will not attempt to link `stage1` toolchain",
```
jyn514, line 147: ditto here, saying "rustup will not attempt" is not right

jyn514 (approved):

src/bootstrap/setup.rs
```diff
@@ -111,21 +111,28 @@ pub fn setup(src_path: &Path, profile: Profile) {
     let build = TargetSelection::from_user(&env!("BUILD_TRIPLE"));
     let stage_path = ["build", build.rustc_target_arg(), "stage1"].join("/");
 
-    let created = fs::create_dir(&stage_path[..]).is_ok();
+    let rustup_installed = Command::new("rustup")
+        .arg("--version")
+        .stdout(std::process::Stdio::null())
+        .output()
+        .is_ok();
```
jyn514, line 118: This needs to check `status()`, remember.

Sl1mb0: 🤦

src/bootstrap/setup.rs
```diff
@@ -111,11 +111,10 @@ pub fn setup(src_path: &Path, profile: Profile) {
...
 
-    let rustup_installed = Command::new("rustup")
-        .arg("--version")
-        .stdout(std::process::Stdio::null())
-        .output()
-        .is_ok();
+    let rustup_installed =
+        Command::new("rustup").arg("--version").stdout(std::process::Stdio::null()).output();
+
+    let rustup_installed = rustup_installed.is_ok() && rustup_installed.unwrap().status.success();
```
jyn514, line 117:
```suggestion
    let rustup_installed = rustup_installed.map_or(false, |status| status.success());
```

src/bootstrap/setup.rs
```diff
@@ -107,6 +108,60 @@ pub fn setup(src_path: &Path, profile: Profile) {
...
+                );
+            }
+        } else {
+            println!(
+                "`rustup` failed to list current toolchains; not attempting to link `stage1` toolchain"
+            );
+        }
+    } else if profile != Profile::User {
+        println!("`rustup` is not installed; cannot link `stage1` toolchain");
+    }
```
jyn514, line 163: This long list of `else` blocks makes it hard to tell which error corresponds with which check IMO - could you instead use a separate functions and early returns so it's more clear? Something like this:
```rust
    maybe_create_rustup_toolchain();
// ...
}

fn maybe_create_rustup_toolchain() {
     if !rustup_installed || !stage_dir_exists {
        // ...
        return
     }
     // ...
}
```

src/bootstrap/setup.rs
```diff
@@ -107,6 +108,60 @@ pub fn setup(src_path: &Path, profile: Profile) {
...
+        .arg("--version")
+        .stdout(std::process::Stdio::null())
+        .output()
+        .map_or(false, |output| output.status.success());
+
+    let stage_dir_exists = match fs::create_dir(&stage_path[..]) {
+        Ok(_) => true,
+        Err(_) => Path::new(&stage_path[..]).exists(),
+    };
+
```
jyn514, line 124: IMO the current output is a little hard to read - maybe add a newline before any possible output?
```
Please choose one (a/b/c/d/e): a
`x.py` will now use the configuration at /home/joshua/src/rust/rust/src/bootstrap/defaults/config.library.toml
Added `stage1` rustup toolchain; try `cargo +stage1 build` on a separate rust project to run a newly-built toolchain

Rust's CI will automatically fail if it doesn't pass `tidy`, the internal tool for ensuring code quality.
```
```suggestion
println!();
```

jyn514 (commented):
Looks pretty good to me :) I just have a couple last nits.

@Mark-Simulacrum do you want to take a look before I approve? No problem if not, this seems pretty low risk.

Mark-Simulacrum:
No, happy to let you own the review here. Can always adjust later in future PRs if it becomes necessary :)

jyn514:
@bors r+

This is awesome, thank you! :heart:

Mark-Simulacrum:
@bors r-

Please squash the commits here to keep history a little cleaner.

PR #69838, opened 2020-03-08
Expansion-driven outline module parsing

Centril:
After this PR, the parser will not do any conditional compilation or loading of external module files when `mod foo;` is encountered. Instead, the parser only leaves `mod foo;` in place in the AST, with no items filled in. Expansion later kicks in and will load the actual files and do the parsing. This entails that the following is now valid:

```rust
#[cfg(FALSE)]
mod foo {
    mod bar {
        mod baz; // `foo/bar/baz.rs` doesn't exist, but no error!
    }
}
```

Fixes https://github.com/rust-lang/rust/issues/64197.

r? @petrochenkov

PR #89214, opened 2021-09-24
Pass real crate-level attributes to `pre_expansion_lint`

smoelius:
The PR concerns the unstable feature `register_tool` (#66079).

The feature's implementation requires the attributes of the crate being compiled, so that when attributes like `allow(foo::bar)` are encountered, it can be verified that `register_tool(foo)` appears in the crate root.

However, the crate's attributes are not readily available during early lint passes. Specifically, on this line, `krate.attrs` appears to be the attributes of the current source file, not the attributes of the whole crate: https://github.com/rust-lang/rust/blob/bf642323d621dcefeef1d8ab4711aae36e357615/compiler/rustc_lint/src/context.rs#L815

Consequently, "unknown tool" errors were being produced when `allow(foo::bar)` appeared in a submodule, even though `register_tool(foo)` appeared in the crate root.

EDITED: The proposed fix is to obtain the real crate-level attributes in `configure_and_expand` and pass them to `pre_expansion_lint`. (See @petrochenkov's [comment](https://github.com/rust-lang/rust/pull/89214#issuecomment-926927072) below.)

The original "prosed fix" text follows.

---

The proposed fix is to add an `error_on_unknown_tool` flag to `LintLevelsBuilder`. The flag controls whether "unknown tool" errors are emitted. The flag is set during late passes, but not earlier.

More specifically, this PR contains two commits:

* The first adds a `known-tool-in-submodule` UI test that does not currently pass.
* The second adds the `error_on_unknown_tool` flag. The new test passes with the addition of this flag.

This change has the added benefit of eliminating some errors that were duplicated in existing tests.

To the reviewer: please check that I implemented the UI test correctly.

### DIFF
--- a/compiler/rustc_interface/src/passes.rs
+++ b/compiler/rustc_interface/src/passes.rs
@@ -240,6 +240,7 @@ fn pre_expansion_lint(
     sess: &Session,
     lint_store: &LintStore,
     krate: &ast::Crate,
+    crate_attrs: &[ast::Attribute],
     crate_name: &str,
 ) {
     sess.prof.generic_activity_with_arg("pre_AST_expansion_lint_checks", crate_name).run(|| {
@@ -247,6 +248,7 @@ fn pre_expansion_lint(
             sess,
             lint_store,
             &krate,
+            crate_attrs,
             true,
             None,
             rustc_lint::BuiltinCombinedPreExpansionLintPass::new(),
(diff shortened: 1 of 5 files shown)


petrochenkov:
>`krate.attrs` appears to be the attributes of the current source file, not the attributes of the whole crate

This is a bug and a regression from https://github.com/rust-lang/rust/pull/69838.
`pre_expansion_lint` is called after loading any new module, and that module is passed to it after being artificially turned into `ast::Crate`, just because `pre_expansion_lint` works with `ast::Crate`.
Real crate-level attributes are available at that point and just need to be additionally passed to `pre_expansion_lint`.

smoelius:
Thanks, @petrochenkov. Please leave the PR open for now while I try to find a proper fix.

smoelius:
@petrochenkov Is this about what you had in mind?

petrochenkov:
Yes, this is exactly what I meant.
@bors r+

@smoelius Could you also update the PR description before the merge? (It's outdated, but I'll be included into the merge commit.)

smoelius:
> @smoelius Could you also update the PR description before the merge? (It's outdated, but I'll be included into the merge commit.)

Absolutely. Please let me know if what I wrote is insufficient.

PR #89215, opened 2021-09-24
Add Clone-less converter from Arc to Vec

197g:
Adds a new converter to `Arc<T: ?Sized>`: `try_unwrap_as_box(self) -> Result<Box<T>, Self>`
and its counterpart `Rc<T: ?Sized>`: `try_unwrap_as_box(self) -> Result<Box<T>, Self>`

This method is a safe encapsulation when one wants to extract the (unsized) items behind an `Arc` into an owned allocation. For example, after a computation phase where a slice of items had been shared with workers. None of the existing methods allow _taking_ ownership of an unsized pointee. The only current alternatives are downcasting to a concrete sized types, or in that specific case of a slice a more expensive _cloning_ of the entire slice which requires an additional bound and might invoke arbitrary code.

Alternatives:
* The implementation detail `leak_as_owning_weak` is the 'necessary' operation for performing this in 'userspace'. Concretely, it contains the minimal code around the `CAS(1, 0)` that moves ownership of the strong count to the caller. It's been factored out as a shared pattern between `try_unwrap` and the new interface. It _could_ in theory be used by non-std code to perform the same operations or even different ones if it were made `pub`. 
  For example user code might provide `Deref<T>` and `DerefMut<T>` in-place by keeping the `Weak`. This requires a proper newtype that also _could_ be added to std but is another non-trivial addition, I suppose. That pattern alone does not solve moving the pointee to a properly owned `Box` though.

### DIFF
--- a/library/alloc/src/boxed.rs
+++ b/library/alloc/src/boxed.rs
@@ -1077,6 +1077,36 @@ impl<T: ?Sized, A: Allocator> Box<T, A> {
         // additional requirements.
         unsafe { Pin::new_unchecked(boxed) }
     }
+
+    /// Allocates a box with the requested layout, which may be for a possibly-unsized value where
+    /// the layout has been determined from a value.
+    ///
+    /// The function `mem_to_imbued` is the called to turn the raw memory pointer into a pointer to
+    /// the type `T`. It's expected that this will add additional fat-pointer metadata.
+    #[cfg(not(no_global_oom_handling))]
+    pub(crate) unsafe fn allocate_for_layout(
+        allocator: &A,
+        layout: Layout,
+        mem_to_imbued: impl FnOnce(*mut u8) -> *mut T,
+    ) -> *mut T {
+        mem_to_imbued(
+            allocator.allocate(layout).unwrap_or_else(|_| handle_alloc_error(layout)).as_ptr()
+                as *mut u8,
+        )
+    }
+
+    /// Allocates a box with sufficient space for the pointee and copy the metadata.
+    #[unstable(feature = "unwrap_rc_as_box", issue = "none")]
+    #[cfg(not(no_global_oom_handling))]
+    #[doc(hidden)]
+    pub unsafe fn allocate_for_ptr(allocator: &A, ptr: *const T) -> *mut T {
+        // Allocate for the `ArcInner<T>` using the given value.
+        unsafe {
+            Self::allocate_for_layout(allocator, Layout::for_value(&*ptr), |mem| {
+                ptr.set_ptr_value(mem) as *mut T
+            })
+        }
+    }
 }
 
 #[stable(feature = "rust1", since = "1.0.0")]
--- a/library/alloc/src/lib.rs
+++ b/library/alloc/src/lib.rs
@@ -136,6 +136,7 @@
 #![feature(unicode_internals)]
 #![feature(unsize)]
 #![feature(unsized_fn_params)]
+#![cfg_attr(not(no_global_oom_handling), feature(unwrap_rc_as_box))]
 #![feature(allocator_internals)]
 #![feature(slice_partition_dedup)]
 #![feature(maybe_uninit_extra, maybe_uninit_slice, maybe_uninit_uninit_array)]
--- a/library/alloc/src/rc.rs
+++ b/library/alloc/src/rc.rs
@@ -612,21 +612,10 @@ impl<T> Rc<T> {
     #[inline]
     #[stable(feature = "rc_unique", since = "1.4.0")]
     pub fn try_unwrap(this: Self) -> Result<T, Self> {
-        if Rc::strong_count(&this) == 1 {
-            unsafe {
-                let val = ptr::read(&*this); // copy the contained object
-
-                // Indicate to Weaks that they can't be promoted by decrementing
-                // the strong count, and then remove the implicit "strong weak"
-                // pointer while also handling drop logic by just crafting a
-                // fake Weak.
-                this.inner().dec_strong();
-                let _weak = Weak { ptr: this.ptr };
-                forget(this);
-                Ok(val)
-            }
-        } else {
-            Err(this)
+        let weak = Self::leak_as_owning_weak(this)?;
+        unsafe {
+            let val = ptr::read(weak.as_ptr()); // copy the contained object
+            Ok(val)
         }
     }
 }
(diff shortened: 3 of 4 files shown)


SkiFire13:
Isn't this a bit limited if it works only for `Arc<[T]> -> Vec<T>`? I would prefer having a `try_unwrap_boxed(self) -> Result<Box<T>, Arc<T>>` that works similar to `try_unwrap`, except it works for `T: ?Sized` since it produces a `Box<T>`. Then this can work for `T=[U]`, `T=str`, `T=Path`, `T=OsStr` ecc ecc thanks to the various implementations of `From<Box<_>>` 

Also, shouldn't `Rc` also get a similar method?

HeroicKatora:
> Also, shouldn't Rc also get a similar method?

Very likely. It's simply that I had the implementation for `Arc` already around (and that's the more complicated one anyways).

> Isn't this a bit limited if it works only for `Arc<[T]> -> Vec<T>`?

You're probably right. I did paint myself into a bit of a corner with the idea of possibly having a zero-allocation conversion by moving the elements to overwrite the `ArcInner`—leaving the remaining space for extra capacity. This idea can only work for `Vec<_>` and not `Box<_>` but it is very much over-engineered for the implementation of this specifically.

Another idea—although with significant additional API surface—would be to make a concrete new type `BoxArc<T: ?Sized>` to wrap around the `Weak<_>` from `leak_as_owning_weak`, with the semantics of owning the allocation without having modified or moved the data in any way. Maybe this PR should be dropped in favor of such an idea (that definitely warrants a major library change proposal though).

HeroicKatora:
Maybe in the first step I will reduce this PR to `leak_as_owning_weak`. This encapsulates the key soundness issue of taking ownership which then allows building the missing abstractions in library space much, much more easily. (Well, except of course the zero-allocation approach because that one specifically would require relying on `std`'s layout of `*Inner<T>`).

SkiFire13:
> possibly having a zero-allocation conversion by moving the elements to overwrite the `ArcInner`— leaving the remaining space for extra capacity

This feels very niche. For this to work I think you would need:
- no outstanding `Weak` reference;
- the align of `T` to be at least 8;
- the size of `T` to be a divisor of 16 or be the same as its alignment.

HeroicKatora:
So which direction seems most promising to you:
* Provide safe `leak_as_owning_weak` and punt the rest to more PRs.
* Directly propose a wrapper `{Arc,Rc}Box<_>`.
* Change the method to returning `Box<T: ?Sized>`.

HeroicKatora:
> Change the method to returning Box<T: ?Sized>.

I just realize that this isn't really feasible. Doing so would require allocating a place with the correct layout _and then duplicating the pointer meta data_, in the process of which we are asserting that it is valid to do so and that this produces a pointer that is valid to dereference. That's true for the existing fat pointers (slice, dyn) but I don't think it's been fully discussed. I don't feel comfortable at all with adding a new allocation method to `Box` for the sake of _this_ PR. So I would go with the first option?

SkiFire13:
> Provide safe `leak_as_owning_weak` and punt the rest to more PRs.

In my opinion this leaks more implementation details than necessary. For example currently there's no public concept of an "owning `Weak`", nor it's considered safe to access the value pointed by a `Weak` whose strong count is 0.

> Directly propose a wrapper `{Arc,Rc}Box<_>`.

Adding a new type for the sake of this PR seems overkill to me.

> Doing so would require allocating a place with the correct layout and then duplicating the pointer meta data, in the process of which we are asserting that it is valid to do so and that this produces a pointer that is valid to dereference. That's true for the existing fat pointers (slice, dyn) but I don't think it's been fully discussed. 

This is already done in the opposite direction by the implementation of `From<Box<T>> for `Rc<T>` and `Arc<T>` (with `T: ?Sized`), so I don't think it would be that problematic to do the opposite. 

> I don't feel comfortable at all with adding a new allocation method to Box for the sake of this PR.

You can use `Box::from_raw` with pointers obtained by allocating with the `Global` allocator. This is publicly documented in the "Memory layout" section of the `std::boxed` module.

HeroicKatora:
I've adjusted everything according to our conversation.

> Adding a new type for the sake of this PR seems overkill to me.

Maybe, however I don't fancy contributing to the proliferation of `_in` methods and duplicate methods to handle allocation failure either. That type would have a significant advantage of separating the point of failure to acquire ownership from the callsite where the allocation can fail—two different errors and two different error sites.

> This is already done in the opposite direction by the implementation of From<Box<T>> for Rc<T> and Arc<T> (with T: ?Sized), so I don't think it would be that problematic to do the opposite.

To learn about those pointer operations being allowed through this channel makes me uncomfortable :grimacing: Anyways, then that concern of course disappears.

HeroicKatora:
Okay, I had messed up the combination between `no_global_oom_handler` and the new feature gate a bunch of times but it should be ready now.

Mark-Simulacrum:
Can you update the PR description/title with a summary of the current state of this PR? As far as I can tell, the proposal is no longer to add the Arc -> Vec conversion, but rather something different.

It looks like this is now adding an API for converting `Arc<T>` to `Box<T>` -- but this movement will always require a copy of the contained value, right? Can you say more about what the use cases (ideally real, not theoretical) for such an API would be?

In particular, a "userspace" library could be written that verifies the passed Arc is unique (via get_mut) when taking ownership of it and then provides &T and &mut T access with no overhead, via the Deref and get_mut_unchecked APIs on the internal Arc. That would require a little unsafe code, but nothing all that interesting.

SkiFire13:
I'm not OP, but I followed the evolution of this PR, and:

> It looks like this is now adding an API for converting Arc<T> to Box<T> -- but this movement will always require a copy of the contained value, right?

It kinda depends with what you mean with "copy". There's a memcpy, but it's not a `clone`, so I would call it a move. Note that this is equivalent to what happens when converting `Box<T>` to `Arc<T>`.

> Can you say more about what the use cases (ideally real, not theoretical) for such an API would be? In particular, a "userspace" library could be written that verifies the passed Arc is unique (via get_mut) when taking ownership of it and then provides &T and &mut T access with no overhead, via the Deref and get_mut_unchecked APIs on the internal Arc. That would require a little unsafe code, but nothing all that interesting.

There are two differences between the PR proposal and yours:
- the PR allows to get common types (`Box`, which can then be turned in a `Vec` ecc ecc), while yours requires an additional type that doesn't play well with the rest of the ecosystem
- your approach will fail if there are outstanding `Weak`, while this PR will just disassociate them (this might need more documentation though). AFAIK this is not something that currently be done in "userspace" code.

HeroicKatora:
> Can you say more about what the use cases (ideally real, not theoretical) for such an API would be

It's an inverse of the existing `Box<_> -> Arc<_>` conversion which wasn't possible for unsized types before. The use case is still the same as in the Vec case (converting the box to vec is almost zero-cost as observed).

> In particular, a "userspace" library could be written that verifies the passed Arc is unique […]

No. There is no way for a userspace library to emulate atomically performing the last CAS (1, 0) on the strong count other than `try_unwrap` which does not work for unsized types (nor for very large types as it also reads them into the stack but this is secondary). The only other comparable function is `make_mut` which performs another allocation of its own and requires the `T: Clone` bound. It might be possible for `Rc` but that still involves a highly non-trivial amount of reasoning for keeping the correct count alive while forgetting the value at the right time etc.

HeroicKatora:
I've updated the description to reflect the changes to the code and some parts of the discussionl

JohnCSimon:
Ping from triage:
@HeroicKatora If this PR is ready to review, adjust the label with `@rustbot ready` . Thanks.

JohnCSimon:

@rustbot label: +S-inactive

PR #89217, opened 2021-09-24
Enable "generate-link-to-definition" option on rust tools docs as well

GuillaumeGomez:
Just realized that we enable the option for the compiler crates, but we don't have it for rustdoc and the other tools documentation...

cc @rust-lang/rustdoc 
r? @Mark-Simulacrum


jyn514:
@GuillaumeGomez why did you cc the whole team?

@bors r+ rollup

GuillaumeGomez:
To keep everyone up-to-date. I think it's an important change after all.

Issue #88869, opened 2021-09-11
Doublicated ld\ld64 executables added into rustc package, increasing it's size


klensy:
Checked that on `x86_64-pc-windows-msvc` target,
found duplicates in versions:
```
rustc 1.56.0-beta.1 (e784c962e 2021-09-07)
rustc 1.57.0-nightly (b69fe5726 2021-09-10)
```
and no duplicates in stable `rustc 1.55.0 (c8dfcfe04 2021-09-06)`, `rustc 1.56.0-nightly (4e282795d 2021-07-31)`
(bisecting is kinda slow on my connection)

@rustbot label +regression-from-stable-to-beta

klensy:
Can be found in:
```
.rustup\toolchains\nightly-x86_64-pc-windows-msvc\lib\rustlib\x86_64-pc-windows-msvc\bin\gcc-ld\ld.exe
.rustup\toolchains\nightly-x86_64-pc-windows-msvc\lib\rustlib\x86_64-pc-windows-msvc\bin\gcc-ld\ld64.exe
```

hyd-dev:
I think they are added in #88250 and #85961.

> and no duplicates (only `ld.exe`)

I believe `ld` is also a copy of `../rust-lld`.

apiraino:
Assigning priority as discussed in the [Zulip thread](https://rust-lang.zulipchat.com/#narrow/stream/245100-t-compiler.2Fwg-prioritization.2Falerts/topic/.2388869.20Doublicated.20ld.5Cld64.20executables.20added.20into.20rustc.20pac.E2.80.A6) of the Prioritization Working Group.

@rustbot label -I-prioritize +P-medium

hkratz:
Sorry, I only came across this issue just now. The copies are there for a reason, see linked issues. 

Multiple things to consider:
* The files in the `gcc-ld` directory are only used with the `-Z gcc-ld=lld` unstable issue. So they could be removed for the beta and stable.
* They can be replaced with small Rust executables which just exec `../rust-lld -flavor ld <supplied args>` and `../rust-lld -flavor ld64 <supplied args>` respectively.

PR #89218, opened 2021-09-24
Include copies of rust-lld for `-Z gcc-ld=lld` only in nightly builds.

hkratz:
Takes care of #88869 for beta/future stable.


klensy:
We can ignore that this will break things with `RUSTC_BOOTSTRAP` on beta\stable?

hkratz:
Closing in favor of #89288.

PR #89219, opened 2021-09-24
Use get_unchecked in str::[r]split_once

nickkuk:
This PR removes indices checking in `str::split_once` and `str::rsplit_once` methods.

### DIFF
--- a/library/core/src/str/mod.rs
+++ b/library/core/src/str/mod.rs
@@ -1524,7 +1524,8 @@ impl str {
     #[inline]
     pub fn split_once<'a, P: Pattern<'a>>(&'a self, delimiter: P) -> Option<(&'a str, &'a str)> {
         let (start, end) = delimiter.into_searcher(self).next_match()?;
-        Some((&self[..start], &self[end..]))
+        // SAFETY: `Searcher` is known to return valid indices.
+        unsafe { Some((self.get_unchecked(..start), self.get_unchecked(end..))) }
     }
 
     /// Splits the string on the last occurrence of the specified delimiter and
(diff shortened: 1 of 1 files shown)


nagisa:
This seems reasonable to me. `Searcher` trait is `unsafe` to implement with this exact invariant for implementers.

Do you have any examples of the improvements achieved by this change?

nickkuk:
Yes, I noticed that current implementation adds panic handling, then I found the reason of this.

Godbolt comparison is [here](https://godbolt.org/z/W6Kd36fnj).

library/core/src/str/mod.rs
```diff
@@ -1524,7 +1524,8 @@ impl str {
     #[inline]
     pub fn split_once<'a, P: Pattern<'a>>(&'a self, delimiter: P) -> Option<(&'a str, &'a str)> {
         let (start, end) = delimiter.into_searcher(self).next_match()?;
-        Some((&self[..start], &self[end..]))
+        // SAFETY: `Searcher` is known to return valid indices.
+        unsafe { Some((self.get_unchecked(0..start), self.get_unchecked(end..self.len()))) }
```
Mark-Simulacrum, line 1528: Why the change to passing ranges bounded on both ends, instead of just `..start` and `end..`?

nickkuk: I looked through the `str/mod.rs` file and used this method with the same style: [1](https://github.com/rust-lang/rust/blob/0ba416580a547853189e82b4940b63afddda84ff/library/core/src/str/mod.rs#L573), [2](https://github.com/rust-lang/rust/blob/0ba416580a547853189e82b4940b63afddda84ff/library/core/src/str/mod.rs#L1983), [3](https://github.com/rust-lang/rust/blob/0ba416580a547853189e82b4940b63afddda84ff/library/core/src/str/mod.rs#L2090). So I'm not sure about this; possible solutions:
1. use ranges bounded on both ends everywhere;
2. use "minimal" ranges where possible on the whole file;
3. as you say, use not bounded ranges only in `split_once` and `rsplit_once`.

nickkuk:
I've removed redundant bounds in `str::split_once` and `str::rsplit_once` methods only.

PR #89221, opened 2021-09-24
Give better error for `macro_rules! name!`

aDotInTheVoid:
r? @estebank 

@rustbot modify labels: +A-diagnostics +A-parser

### DIFF
--- a/compiler/rustc_parse/src/parser/item.rs
+++ b/compiler/rustc_parse/src/parser/item.rs
@@ -1547,6 +1547,20 @@ impl<'a> Parser<'a> {
         self.expect(&token::Not)?; // `!`
 
         let ident = self.parse_ident()?;
+
+        if self.eat(&token::Not) {
+            // Handle macro_rules! foo!
+            let span = self.prev_token.span;
+            self.struct_span_err(span, "macro names aren't followed by a `!`")
+                .span_suggestion(
+                    span,
+                    "remove the `!`",
+                    "".to_owned(),
+                    Applicability::MachineApplicable,
+                )
+                .emit();
+        }
+
         let body = self.parse_mac_args()?;
         self.eat_semi_for_macro_if_needed(&body);
         self.complain_if_pub_macro(vis, true);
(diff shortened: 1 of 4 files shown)


compiler/rustc_parse/src/parser/item.rs
```diff
@@ -1547,6 +1547,20 @@ impl<'a> Parser<'a> {
         self.expect(&token::Not)?; // `!`
 
         let ident = self.parse_ident()?;
+
+        if self.eat(&token::Not) {
+            // Handle macro_rules! foo!
+            let span = self.prev_token.span;
+            self.struct_span_err(self.prev_token.span, "macro names arn't followed by a `!`")
```
estebank, line 1554:
```suggestion
            self.struct_span_err(span, "macro names aren't followed by a `!`")
```

compiler/rustc_parse/src/parser/item.rs
```diff
@@ -1547,6 +1547,20 @@ impl<'a> Parser<'a> {
...
+                .span_suggestion_verbose(
+                    span,
+                    "try removing this `!`",
```
estebank, line 1557:
```suggestion
                    "removing the `!`",
```

compiler/rustc_parse/src/parser/item.rs
```diff
@@ -1547,6 +1547,20 @@ impl<'a> Parser<'a> {
         self.expect(&token::Not)?; // `!`
 
         let ident = self.parse_ident()?;
+
+        if self.eat(&token::Not) {
+            // Handle macro_rules! foo!
+            let span = self.prev_token.span;
+            self.struct_span_err(self.prev_token.span, "macro names arn't followed by a `!`")
+                .span_suggestion_verbose(
```
estebank, line 1555: I think this one is ok to be inline: we're asking them to remove the whole thing in that is underlined, it should be fine.

```suggestion
                .span_suggestion(
```

estebank (approved):
You'll have to rebless after addressing the review comments, but after that r=me

PR #89229, opened 2021-09-24
Remove redundant member-constraint check

oli-obk:
impl trait will, for each lifetime in the hidden type, register a "member constraint" that says the lifetime must be equal or outlive one of the lifetimes of the impl trait. These member constraints will be solved by borrowck

But, as you can see in the big red block of removed code, there was an ad-hoc check for member constraints happening at the site where they get registered. This check had some minor effects on diagnostics, but will fall down on its feet with my big type alias impl trait refactor. So we removed it and I pulled the removal out into a (hopefully) reviewable PR that works on master directly.


src/test/ui/impl-trait/multiple-lifetimes/ordinary-bounds-unrelated.stderr
```diff
@@ -4,11 +4,7 @@ error[E0700]: hidden type for `impl Trait` captures lifetime that does not appea
 LL | fn upper_bounds<'a, 'b, 'c, 'd, 'e>(a: Ordinary<'a>, b: Ordinary<'b>) -> impl Trait<'d, 'e>
    |                                                                          ^^^^^^^^^^^^^^^^^^
    |
-note: hidden type `Ordinary<'_>` captures lifetime smaller than the function body
-  --> $DIR/ordinary-bounds-unrelated.rs:16:74
-   |
-LL | fn upper_bounds<'a, 'b, 'c, 'd, 'e>(a: Ordinary<'a>, b: Ordinary<'b>) -> impl Trait<'d, 'e>
-   |                                                                          ^^^^^^^^^^^^^^^^^^
+   = note: hidden type `Ordinary<'_>` captures lifetime '_#9r
```
oli-obk, line 7: I'll make a separate PR that fixes this diagnostic just for nll. We should probably not merge this PR before that not-yet-written nll diagnostics PR

oli-obk: This PR is up now: https://github.com/rust-lang/rust/pull/89327

src/test/ui/self/arbitrary_self_types_pin_lifetime_impl_trait.stderr
```diff
@@ -1,21 +1,15 @@
-error[E0759]: `self` has an anonymous lifetime `'_` but it needs to satisfy a `'static` lifetime requirement
-  --> $DIR/arbitrary_self_types_pin_lifetime_impl_trait.rs:6:44
+error[E0700]: hidden type for `impl Trait` captures lifetime that does not appear in bounds
```
nagisa, line 1: This seems like a significant regression to diagnostics, which I'm not sure we can afford. Do we have a plan for recovering usability of the diagnostic here?

(I'm not saying that the previous diagnostic was perfect, but it did in point out the relevant pieces of code with reasonably useful explanations, and this… doesn't)

nagisa (commented):
Nit: the "Remove dead code" commit message would imply that its a non-functional change, but from the changes to the tests there are clearly some behavioural changes, and so I don't believe that the commit message accurately reflects the change.

The code changes in commits in isolation all seem reasonable to me, otherwise. But see the inline comments for other concerns.

nagisa:
Ah, to add, I'm not super familiar with this code, so having somebody else to review this as well would be good too.

src/test/ui/self/arbitrary_self_types_pin_lifetime_impl_trait.stderr
oli-obk: yea, this is a problem. I can't make the span point to `self` until I do lazy TAITs, and I can't do lazy TAITs in reasonably reviewable PRs without making PRs such as this one first.

We are later actually pointing to the `&Self` here, so that's actually an improvement.

I will do a separate PR to merge before this one that will introduce the structured suggestion in nll mode and add the label with "this data with an anonymous lifetime".

Since this PR essentially removes old regionchck, I can do all improvements in nll mode and we'll see it in the .nll.stderr files.

oli-obk:
Diagnostics quality has been restored by https://github.com/rust-lang/rust/pull/89555, and in some cases even improved, in others I expect the spans to improve with my lazy TAIT branch. There are a few that are completely different in their content, but that is the expected change from this PR. We don't eagerly figure out lifetime constraints, but instead leave it to member constraints.

In some cases this causes diagnostics that said "`'a` must outlive `'b`" to change to "hidden type captures `'b`" without referencing `'a`. This may seem like a regression, but the previous diagnostic wasn't correct in all cases. It is just one way to resolve the error, and now we propose a slightly different way. Ideally we'd report all ways, and member constraints can actually report more information, but that would change too much within this PR than I'm comfortable doing. I'll open an issue so we remember to revisit this.

jackh726:
I'm going to mark as blocked on #89555

oli-obk:
> I'm going to mark as blocked on #89555

that PR has been merged, and this one rebased

I'll open an issue for src/test/ui/self/arbitrary_self_types_pin_lifetime_impl_trait-async.rs so we get the better diagnostic back. I think we should merge it without fixing it immediately, as it was a pre-existing problem in other situations

jackh726:
@oli-obk I'll try to review this weekend!

jackh726 (commented):
I think this is fine...

I'm seeing some duplicate errors being emitted; it would be nice to not do that, but I'd be fine landing this regardless.

I'm going to go ahead and say r=me here with or without fixing duplicate errors being emitted.

Issue #88621, opened 2021-09-03
Determine how to handle casting of arbitrary discriminants with different enum styles


ehuss:
This is an issue to track making a decision on how to handle certain styles of enums with respect to the new `arbitrary_enum_discriminant` which has been stabilized in 1.56 (#86860) which will be released October 21st 2021.  I think it would be good to make a decision before it reaches stable.

I have a concern that this introduces three styles of enums, and I think it would be best to only have two styles.  In summary, as of 1.56, they will be:

Kind 1: Unit-only variants:

```rust
// Does not require repr
enum Kind1 {
    Foo = 1,
    Bar = 2,
    Baz = 3,
}

// Can be cast to its discriminant:
let x = Kind1::Foo as u8;
```

Kind 2: Tuple or struct variants without fields:

```rust
#[repr(u8)]   // Required repr
enum Kind2 {
    Foo() = 1,
    Bar{} = 2,
    Baz = 3,
}

// Can be cast to its discriminant:
let x = Kind2::Foo() as u8;
```

Kind 3: Tuple or struct variants with fields:

```rust
#[repr(u8)]   // Required repr
enum Kind3 {
    Foo(i32) = 1,
    Bar{f: i32} = 2,
    Bar = 3,
}

// ERROR: Does not allow casting.
let x = Kind3::Foo(1) as u8;
```

The question mainly centers on how `Kind2` should be treated.  #88203 is a proposal to make it treated the same as `Kind1` (that is, does not require `repr`).  Another proposal is to make it treated the same as `Kind3` (does not allow casting), PR up at #89234.  Another option it to leave it as-is, and keep three styles.

Only `Kind1` is allowed in 1.55 and older versions.

(As an aside, this also highlights that our current terminology of "Fieldless enums" is confusing, since it has historically meant only `Kind1` style, but it can be confused with `Kind2` since those do not have fields, either.)

More discussion at https://rust-lang.zulipchat.com/#narrow/stream/213817-t-lang/topic/arbitrary.20enum.20discriminants/near/251899236

ehuss:
@rust-lang/lang  Nominating to verify what we should do.  It was observed that the following has been allowed forever:

```rust
// Works with or without repr.
enum Fieldless {
    Foo(),
    Bar{},
    Baz,
}

fn main() {
    assert_eq!(Fieldless::Foo() as u8, 0);
    assert_eq!(Fieldless::Bar{} as u8, 1);
[...]

Issue #60553, opened 2019-05-05
Tracking issue for RFC 2363, "Allow arbitrary enums to have explicit discriminants"


Centril:
This is a tracking issue for the RFC "Allow arbitrary enums to have explicit discriminants" (rust-lang/rfcs#2363).

**Steps:**

- [x] Implement the RFC – https://github.com/rust-lang/rust/pull/60732
- [ ] Adjust documentation ([see instructions on rustc-guide][doc-guide])
- [x] Stabilization PR – #86860

[stabilization-guide]: https://rust-lang.github.io/rustc-guide/stabilization_guide.html#stabilization-pr
[doc-guide]: https://rust-lang.github.io/rustc-guide/stabilization_guide.html#documentation-prs

**Unresolved questions:**

1. [ ] Should discriminants of enumerations with fields be specified as variant attributes?
2. [ ] Should this apply only to enumerations with an explicit representation?

eddyb:
Mentoring-wise: I'd suggest trying out an example and then tracking down where the error comes from. I don't remember, it's either during parsing or in `ast_validation`.

Most of the compiler should already support this, where doing the uniform thing is easier.
The "niche" optimization might assume variants with any data have consecutive discriminants, but it should be easy to fix, and we can figure it out after an initial PR is open.

jswrenn:
Has anyone stepped up to implement this? If not, I'd like to take a stab.

The current error arises in parsing. I haven't fiddled with it yet, but I suspect a good first step is to remove the `else` before this `if`: https://github.com/rust-lang/rust/blob/33cde4aac2ec1b0a493d0acaa4c4fb45de0c6e94/src/libsyntax/parse/parser.rs#L7781-L7789

oli-obk:
> Has anyone stepped up to implement this?

I don't know of anyone

>  If not, I'd like to take a stab.

Go for it!

> I haven't fiddled with it yet, but I suspect a good first step is to remove the else before this if: 

That looks like the right start, yes.

jswrenn:
I've started work on this [here](https://github.com/jswrenn/rust/tree/arbitrary_enum_discriminant) by [removing the parsing restriction preventing non-unit variants from having explicit discriminants](https://github.com/jswrenn/rust/commit/ff93aed9022145beb1a37dc618d0db4735e0df18). As you predicted @eddyb, things seem to basically just work; e.g.:
```rust
#![feature(core_intrinsics)]

enum Enum {
  A(u16) = 7,
[...]

PR #89234, opened 2021-09-24
Disallow non-c-like but "fieldless" ADTs from being casted to integer if they use arbitrary enum discriminant

nbdd0121:
Code like

```rust
#[repr(u8)]
enum Enum {
    Foo /* = 0 */,
    Bar(),
    Baz{}
}

let x = Enum::Bar() as u8;
```

seems to be unintentionally allowed so we couldn't disallow them now ~~, but we could disallow them if arbitrary enum discriminant is used before 1.56 hits stable~~ (stabilization was reverted).

Related: #88621

@rustbot label +T-lang

### DIFF
--- a/compiler/rustc_middle/src/ty/adt.rs
+++ b/compiler/rustc_middle/src/ty/adt.rs
@@ -7,7 +7,7 @@ use rustc_data_structures::fx::FxHashMap;
 use rustc_data_structures::stable_hasher::{HashStable, StableHasher};
 use rustc_errors::ErrorReported;
 use rustc_hir as hir;
-use rustc_hir::def::{DefKind, Res};
+use rustc_hir::def::{CtorKind, DefKind, Res};
 use rustc_hir::def_id::DefId;
 use rustc_index::vec::{Idx, IndexVec};
 use rustc_query_system::ich::StableHashingContext;
@@ -314,6 +314,22 @@ impl<'tcx> AdtDef {
     /// Whether the ADT lacks fields. Note that this includes uninhabited enums,
     /// e.g., `enum Void {}` is considered payload free as well.
     pub fn is_payloadfree(&self) -> bool {
+        // Treat the ADT as not payload-free if arbitrary_enum_discriminant is used (#88621).
+        // This would disallow the following kind of enum from being casted into integer.
+        // ```
+        // enum Enum {
+        //    Foo() = 1,
+        //    Bar{} = 2,
+        //    Baz = 3,
+        // }
+        // ```
+        if self
+            .variants
+            .iter()
+            .any(|v| matches!(v.discr, VariantDiscr::Explicit(_)) && v.ctor_kind != CtorKind::Const)
+        {
+            return false;
+        }
         self.variants.iter().all(|v| v.fields.is_empty())
     }
 
--- a/src/test/ui/cast/issue-88621.rs
+++ b/src/test/ui/cast/issue-88621.rs
@@ -1,0 +1,13 @@
+#![feature(arbitrary_enum_discriminant)]
+
+#[repr(u8)]
+enum Kind2 {
+    Foo() = 1,
+    Bar{} = 2,
+    Baz = 3,
+}
+
+fn main() {
+    let _ = Kind2::Foo() as u8;
+    //~^ ERROR non-primitive cast
+}
--- a/src/test/ui/cast/issue-88621.stderr
+++ b/src/test/ui/cast/issue-88621.stderr
@@ -1,0 +1,9 @@
+error[E0605]: non-primitive cast: `Kind2` as `u8`
+  --> $DIR/issue-88621.rs:11:13
+   |
+LL |     let _ = Kind2::Foo() as u8;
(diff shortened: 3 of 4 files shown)


nikomatsakis:
We did discuss this general topic in a recent meeting but we haven't had much quorum to figure out how to go. I believe the conclusion in *that* meeting was that, until we have time to discuss this properly, we should roll back the stabilization.

jackh726:
Uh...this seems to have been forgotten. And we just passed a release, so it's probably too late to roll back stabilization.

So, nominating for T-lang to make a decision on what to do going forward.

jackh726:
Oh, okay. Then I will continue to keep this nominated (and update my previous comment) for lang team to discuss and figure out what to do.

nikomatsakis:
@nbdd0121 is this issue specifically about an accidental stabilization? If so, I cannot compile any snippet like the above, butI can do this on nightly (both feature + repr are required):

```rust
#![feature(arbitrary_enum_discriminant)]

#[repr(u8)]
enum Enum {
    Foo = 1,
    Bar(),
    Baz{}
}

fn main() { }
```

If this issue is just about stabilization, then I think it can be closed. Is it also about deciding the right behavior on nightly?

jackh726:
I kept the nomination is to decide the right behavior on nightly. I don't personally have the background here, but I see there's some discussion in #88621.

nbdd0121:
> @nbdd0121 is this issue specifically about an accidental stabilization? If so, I cannot compile any snippet like the above, butI can do this on nightly (both feature + repr are required):

I think I made a mistake in the description, it's now fixed.

> ```rust
> #![feature(arbitrary_enum_discriminant)]
> 
> #[repr(u8)]
> enum Enum {
>     Foo = 1,
>     Bar(),
>     Baz{}
> }
> 
> fn main() { }
> ```
> 
> If this issue is just about stabilization, then I think it can be closed. Is it also about deciding the right behavior on nightly?

This PR was about preventing the accidental stabilization of the example you mentioned. The stabilization is now reverted, but I think the changes made by this PR is still the desirable behaviour for nightly.

nikomatsakis:
@nbdd0121 I edited the OP to indicate that there is no "time deadline" to fixing this, now that stabilization was reverted.

joshtriplett:
We discussed this in today's @rust-lang/lang meeting, and we're comfortable going forward with this. This is forward-compatible; we could always change this and allow such casts in the future. But we felt in the meeting that it was reasonable to disallow this.

jswrenn:
My understanding of #88621 is to reduce some of the complexity surrounding enums (i.e., go from "three styles" to "two styles), but I worry this PR actually adds documentation complexity and introduces a minor (but subtle) stability hazard to library authors.

## Documentation Complexity 
**Without this PR**:
> The unit and struct-like variants of fieldless enums can be `as`-casted.

**With this PR:**
> The unit and struct-like variants of fieldless enums without explicit discriminants can be `as`-casted. All values of C-like enums (i.e., enums where all variants are unit-like) with explicit discriminants can be `as`-casted.

AFAIK, this might be the *only* place in the language reference actually needing to invoke the concept of "C-like" enums. I'd prefer only having to explain two kinds of enums (fieldless and non-fieldless) rather than three (fieldless, C-like, and non-fieldless). I'd also like to ditch "C-like" for another reason: the term isn't self-explanatory; rather it leans on Rust programmers knowing a completely different language. (We wouldn't call non-fieldless enums "OCaml-like enums", would we!?)

## Stability Hazard
Implicit discriminants mean that seemingly-innocuous acts like adding a variant to a `non_exhaustive` enum or reordering variants is potentially a breaking change. We can't easily walk back this design decision, but library authors can mitigate it by adding explicit discriminants wherever possible. Unfortunately PR would make adding explicit discriminants to *some* `as`-castable enums a *breaking* change:
```rust
// uncommenting the explicit discriminants...
#[repr(u8)]
enum Enum {
    Foo /* = 0 */,
    Bar() /* = 1 */,
    Baz{} /* = 2 */,
}
// ...will break downstream casts like this:
let x = Enum::Bar() as u8;
```

This is so niche that I (hope) breakage would be uncommon, but it's nonetheless a subtlety that would need to be documented.

nbdd0121:
@jswrenn I think what we need in addition to this PR is a lint that warns against casting non-all-unit enum.

joshtriplett:
Follow-up: we would very much appreciate if one of the folks working on either this or https://github.com/rust-lang/rust/pull/88203 would be up for adding documentation on enum discriminant handling to the Rust reference.

jackh726:
Given that lang-team is good to move forward with this and that appropriate tracking (in the tracking issue) is done for a decision on this prior to next stabilization, this LGTM.

@bors r+

ehuss:
@jackh726 Since the lang team is asking questions at https://github.com/rust-lang/rust/issues/60553#issuecomment-982920247, it is not clear to me that a decision was made.

jackh726:
@ehuss I'm going off of this comment: https://github.com/rust-lang/rust/pull/89234#issuecomment-977022018

I think the questions posted on the tracking issue are for PRs going forward, as well as for future stabilization.

matthiaskrgr:
Failed in a rollup: https://github.com/rust-lang/rust/pull/91414#issuecomment-983322570
@bors r-

PR #89237, opened 2021-09-24
make `#[track_caller]` actually do stuff in `Steal::borrow`

BoxyUwU:
makes this ICE message useful:
``thread 'rustc' panicked at 'attempted to read from stolen value', /rustc/ac2d9fc509e36d1b32513744adf58c34bcc4f43c\compiler\rustc_data_structures\src\steal.rs:37:21``


eddyb:
This might be sensitive for incremental:

@bors try @rust-timer queue

eddyb (approved):
r=me if perf comes out neutral

klensy:
Better add test to track error message :-)

PR #89239, opened 2021-09-25
resolve: Cache module loading for all foreign modules

petrochenkov:
It was previously cached for modules loaded from `fn get_module`, but not for modules loaded from `fn build_reduced_graph_for_external_crate_res`.
This also makes all foreign modules use their real parent, span and expansion instead of possibly a parent/span/expansion of their reexport.
Modules are also often compared using referential equality (`ptr::eq`), this change makes such comparisons correct in all cases.

An ICE happening on attempt to decode expansions for foreign enums and traits is avoided.

Also local enums and traits are now added to the module map.

Follow up to https://github.com/rust-lang/rust/pull/88872.
r? @cjgillot

### DIFF
--- a/compiler/rustc_span/src/hygiene.rs
+++ b/compiler/rustc_span/src/hygiene.rs
@@ -601,7 +601,10 @@ pub fn debug_hygiene_data(verbose: bool) -> String {
                 let expn_data = expn_data.as_ref().expect("no expansion data for an expansion ID");
                 debug_expn_data((&id.to_expn_id(), expn_data))
             });
-            data.foreign_expn_data.iter().for_each(debug_expn_data);
+            // Sort the hash map for more reproducible output.
+            let mut foreign_expn_data: Vec<_> = data.foreign_expn_data.iter().collect();
+            foreign_expn_data.sort_by_key(|(id, _)| (id.krate, id.local_id));
+            foreign_expn_data.into_iter().for_each(debug_expn_data);
             s.push_str("\n\nSyntaxContexts:");
             data.syntax_context_data.iter().enumerate().for_each(|(id, ctxt)| {
                 s.push_str(&format!(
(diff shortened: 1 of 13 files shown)


petrochenkov:
>This also makes all foreign modules use their real parent, span and expansion instead of possibly a parent/span/expansion of their reexport.

This is a correctness fix (for macros 2.0, I'll try to add a test case, e.g. with a macro 2.0 using `super`), but it does require decoding more in some cases.

The largest regression (`webrender-wrench-check`) is due to a larger number of the `type_of` query invocations, so it's either spurious, or is a result of some larger number of incremental invalidations due to the correctness fix above.
Other regressions are under 0.7%.

petrochenkov:
Updated.

compiler/rustc_span/src/hygiene.rs
```diff
@@ -67,7 +67,7 @@ rustc_index::newtype_index! {
 }
 
 /// A unique ID associated with a macro invocation and expansion.
-#[derive(Clone, Copy, PartialEq, Eq, Hash)]
+#[derive(Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Hash)]
```
cjgillot, line 70: Can this cause future issues like the PartialOrd/Ord impls for `DefId` do?

petrochenkov: Maybe? I didn't know `DefId` have such issues.
I can remove this and rely on `sort_by` instead.

cjgillot: The issue with DefId is: since the API allows to sort by DefId, some queries used to sort their result by DefId. Since a later compilation may assign different DefIds to corresponding definitions, we would get in another order.
Making this kind of mistake difficult by restricting the API and using `sort_by` would probably be better.

petrochenkov: Fixed.

cjgillot:
This corrects behaviour for macros 2.0 and avoids an ICE, so a .7% regression seems acceptable.
@bors r+

Mark-Simulacrum:
> This corrects behaviour for macros 2.0 and avoids an ICE, so a .7% regression seems acceptable.

Issue #40713, opened 2017-03-21
Switch `run-make` tests from Makefiles to rust


TimNN:
Rust currently has a suite of [`run-make`](https://github.com/rust-lang/rust/tree/master/src/test/run-make) tests, which generally test specific `rustc` invocations or behaviour,  or require external tools (eg. `grep` or `nm`).

The goal of this issue is to rewrite these `run-make` tests, which are currently written as Makefiles, in rust to a) get rid of the dependency on external tools such as `make` and b) make them more accessible to rust contributors by not requiring arcane knowledge of the `make` tool.

The transition will require at least the following steps:

* [ ] Survey the existing `run-make` tests with regard to what they actually do / which programs they use (and how). I assume the programs will likely fall into one of three categories:
    * Rust tools (`rustc`, `rustdoc`)
    * Utilities easily replaceable by rust code, eg. `grep`
    * Complex utilities, eg. `nm`, which cannot be easily rewritten in rust
* [ ] Design and (partially) implement a support library, which makes the actions identified above easily possible. For example I imagine a test should look something like this:
    ```rust
    extern crate support;
    
    fn main() {
        let s = support::init();
        
        let lib = s.rustc().compile("file1.rs").output_rlib();
        
        assert!(s.nm(lib).filter_lines("some_symbol").count() == 2);
    }
    ```
* [ ] Add a new mode to [`compile-test`](https://github.com/rust-lang/rust/tree/master/src/tools/compiletest), maybe `run-rmake`, which runs the new rust-based `run-make` tests. This will either include compiling the support library, or receiving the support library from a previous build stage.
* [ ] Start porting the actual `run-make` tests, which may involve adding additional functionality to the support library. At that time this issue, or another, will track the state of all the existing tests and include some detailed instruction to allow people to easily contributing by porting one of the existing tests.

---

There are some open questions:

* Should the support library allow execution of arbitrary commands? Limiting commands to only those explicitly added to the support library would mean that there is a single place which lists the external tools we depend on.
* Is the proposed integration with `compiletest` the correct choice? For example the support library itself may have external dependencies  (maybe `regex` or `gcc-rs`) which means we should probably use cargo to compile the actual tests. At which point it may be worth considering if `compile-test` is needed at all or if cargo is enough (Have the support library in `src/lib.rs`,  the new `run-make` tests in `tests/*.rs` and auxiliary files in subdirectories of `tests/` named after the main test file).

---

[...]

PR #89240, opened 2021-09-25
Trim runtest

Nicholas-Baron:
Work towards #60302.

`src/tools/compiletest/src/runtest.rs` was over 4k lines. To break it down into smaller chunks,
1. The `TestCx` struct was extracted, as it constituted the bulk of the file.
2. The functions following the pattern `run_foo_test` were moved into a separate file, as they were ~1k lines by themselves.


jyn514:
> Along the way, runtest.rs was replaced with runtest/mod.rs.

Why this change? Rustc will happily accept `runtest.rs` with `runtest/test_cx.rs` as a submodule.

Nicholas-Baron:
@jyn514 More a stylistic preference than anything else.

I will be redoing this PR as some smaller changes were unnecessary.

jyn514:
@Nicholas-Baron it seems better to keep the original filename to avoid more churn than necessary.

camsteffen:
Rather than `test_cx`, I would like to see more semantically meaningful modules. For instance, there could be a module for constructing rustc args and a module for parsing output.

Nicholas-Baron:
@camsteffen While I agree with the sentiment, the `TestCx` struct seems to cover most of the semantics in the runtest module. I am unfamiliar with this part of the codebase, so taking the time to divide up the `TestCx` impl block semantically would be a bit longer than what I initially intended.

Mark-Simulacrum:
I think this is not quite the change I would expect for this goal. Generally speaking, just moving code to a separate file (particularly when the move is based on a somewhat arbitrary line) isn't really making it easier to understand the file I think. Instead, I would suggest some of these changes as a start, if you'd like to cleanup compiletest a little. I think these will lead to slightly better structure for the code, while keeping the bits that have a shared structure to them in the same place for easier reading.

* RunPassValgrind is practically useless (valgrind is not installed on CI, and the tests are just run in run-pass mode). The right approach here is likely to refactor it as a modifier atop the regular run-pass tests -- it looks like the code generally just mirrors the RunPass mode other than setting the runtool to valgrind. That mode can be enabled in the configuration of the test with `// use-valgrind` or something like that.
* Move diff computation out into a separate module (DiffLine, Mismatch, make_diff, ...)
* Move read2_abbreviated to the read2.rs module
* Move DebuggerCommands and their parsing out to a separate file, alongside check_debugger_output (which is similarly parsing code).

I suspect this set of changes is likely harder, and might not get to <3000 lines; there's likely more cleanup work possible though. This module has grown pretty organically over the last ~10 years -- there's a lot of small bits that can be adjusted to make things better. There's probably also subtle inconsistencies between various functions that would likely be nice to cleanup and unify, and filing PRs for those would be helpful -- likely some care is needed, perhaps even some small test updates, but overall a good goal.

You might also be interested in https://github.com/rust-lang/rust/issues/40713, as a somewhat larger project.

Nicholas-Baron:
@Mark-Simulacrum Thank you for the suggestions. I have spun them off as a "subtask", that will probably take multiple PRs to finish.

PR #89245, opened 2021-09-25
refactor: make VecDeque's IterMut fields module-private, not just crate-private

DeveloperC286:
Made the fields of VecDeque's IterMut private by creating a IterMut::new(...) function to create a new instance of IterMut and migrating usage to use IterMut::new(...).


joshtriplett:
Seems reasonable, but I think the `new` function should be `unsafe` since it works with a raw pointer and the passed-in values must meet invariants that aren't checked.

DeveloperC286:
> Seems reasonable, but I think the `new` function should be `unsafe` since it works with a raw pointer and the passed-in values must meet invariants that aren't checked.

Changed the `new` function to be unsafe.

PR #89247, opened 2021-09-25
Add `const_eval_select` intrinsic 

fee1-dead:
Adds an intrinsic that calls a given function when evaluated at compiler time, but generates a call to another function when called at runtime.

See https://github.com/rust-lang/const-eval/issues/7 for previous discussion.

r? @oli-obk.


oli-obk:
cc @rust-lang/wg-const-eval

compiler/rustc_codegen_ssa/src/mir/block.rs
```diff
@@ -569,6 +569,23 @@ impl<'a, 'tcx, Bx: BuilderMethods<'a, 'tcx>> FunctionCx<'a, 'tcx, Bx> {
             _ => None,
         };
 
+        if intrinsic == Some(sym::const_eval_select) {
```
oli-obk, line 572: The cranelift backend probably needs something similar (cc @bjorn3)

library/core/src/intrinsics.rs
```diff
@@ -784,6 +784,42 @@ extern "rust-intrinsic" {
...
+    /// fn` with the result produced by evaluating it at compile-time. If
+    /// evaluating the function at run-time were to produce a different result,
+    /// or have any other observable side-effects, the behavior is undefined.
+    ///
+    /// [referential transparency]: https://en.wikipedia.org/wiki/Referential_transparency
+    #[cfg(not(bootstrap))]
+    // Although it is const stable, it should never be exposed outside core::intrinsics,
+    // and never stabilized.
+    #[rustc_const_stable(feature = "const_eval_select", since = "1.57")]
+    pub fn const_eval_select<ARG, F, G, RET>(arg: ARG, called_in_const: F, called_at_rt: G) -> RET;
```
oli-obk, line 821: Do you think we can avoid some of the typeck complexity by adding Fn trait bounds to F and G?

oli-obk (changes requested):
I would like to see some tests that hit the invalid use cases that typeck is catching and thus report errors. While it is not important for intrinsics, it's still good to check that they aren't completely bonkers

compiler/rustc_codegen_ssa/src/mir/block.rs
bjorn3: Indeed. That did be in rustc_codegen_cranelift/src/intrinsics/mod.rs. You can call `abi::codegen_terminator_call` for codegening the actual call. Make sure that you return from the intrinsic handling function directly after this call to avoid two jump instructions after each other in the same basic block.

compiler/rustc_monomorphize/src/collector.rs
```diff
@@ -775,10 +777,23 @@ impl<'a, 'tcx> MirVisitor<'tcx> for MirNeighborCollector<'a, 'tcx> {
...
                 let callee_ty = func.ty(self.body, tcx);
                 let callee_ty = self.monomorphize(callee_ty);
-                visit_fn_use(self.tcx, callee_ty, true, source, &mut self.output);
+                match callee_ty.kind() {
+                    ty::FnDef(def_id, _)
+                        if self.tcx.fn_sig(*def_id).skip_binder().abi == RustIntrinsic
+                            && self.tcx.item_name(*def_id) == sym::const_eval_select =>
+                    {
+                        // mark the second arg and the third arg as this function's neighbor.
+                        for func in &args[1..=2] {
```
bjorn3, line 789: Only the third arg should be necessary, right? The ctfe version shouldn't need to be codegened, does it?

compiler/rustc_typeck/src/check/fn_ctxt/checks.rs
```diff
@@ -450,6 +456,78 @@ impl<'a, 'tcx> FnCtxt<'a, 'tcx> {
...
+            let result: Result<_, DiagnosticBuilder<'_>> = try {
+                for (i, ty, _) in final_arg_types {
+                    match (i, ty.kind()) {
+                        (0, ty::Tuple(_)) => {
+                            args_ty = Some(ty);
+                        }
+                        (0, _) => Err(self
+                            .tcx
+                            .sess
+                            .struct_span_err(sp, "expected ARG to be unit or tuple"))?,
```
bjorn3, line 472: Why are only tuples allowed?

nbdd0121:
Alternative design idea #89279: have a `call_if_rt`
```rust
pub fn call_if_rt<R, F: FnOnce() -> R + Copy>(f: F) -> Option<R>;
```
that returns `Some(f())` in runtime and `None` in compile time.

Pro:
* No changes required to typeck, codegen
* Allows everything to be written in the same function: 
```rust
const fn some_const_fn() {
    if let Some(ret) = call_if_rt(|| {
		// runtime part
	}) {
		return ret;
	}
	// const part
}
```

Con:
* Can't handle move.

nbdd0121:
Came up with a new idea while writing the above comment:

How about have two functions
```rust
pub fn is_const_eval() -> bool;
```
and
```rust
pub fn call_at_rt<R, F: FnOnce() -> R>(f: F) -> R;
```
which will just call `f()` in runtime, and abort if called in compile time.

This should have the same pro as `call_if_rt` that it allows no typeck, codegen changes and allows everything to be written in the same function: 
```rust
const fn some_const_fn() {
    if !is_const_eval() {
		return call_at_rt(|| {
			// runtime part
		});
	}
	// const part
}
```
while it could still handle move.

RalfJung:
Seems a bit annoying to have two intrinsics. Also, const-check special handling will still be required to not subject the closure called by `call_at_rt` to const-check rules.

Why does this PR require such large typeck changes in the first place?

nbdd0121:
> Seems a bit annoying to have two intrinsics. Also, const-check special handling will still be required to not subject the closure called by `call_at_rt` to const-check rules.

Closure isn't considered const, so no changes are needed for that regard: https://play.rust-lang.org/?version=nightly&mode=debug&edition=2018&gist=1fc25330c53df1da0cfffec169fa61e3

fee1-dead:
Both cannot handle move, which can be a limitation, and I think that we shouldn't allow closures to be used in const contexts like this. There could be hidden problems. Therefore this might be the best and the most straightforward solution we have.

I wouldn't consider being able to write them in one function a pro, IMO I think we should make this intrinsic a little bit intentionally complex and hard to use.

oli-obk:
> Both cannot handle move, which can be a limitation, and I think that we shouldn't allow closures to be used in const contexts like this. There could be hidden problems. Therefore this might be the best and the most straightforward solution we have.

Could we pick the API of this PR, but still use the lang item solution? So

```rust
#[lang = "const_eval_select"]
pub fn const_eval_select<R, F: FnOnce() -> R, G: const FnOnce() -> R>(f: F) -> R {
    f()
}
```

Then the typeck change could be just checking that F and G are function items, while leaving all the rest (signature, constness, ..) to the regular typeck logic?

This would avoid touching codegen backends entirely.

nbdd0121:
> Could we pick the API of this PR, but still use the lang item solution? So

#88963 would be needed first.

> ```rust
> #[lang = "const_eval_select"]
> pub fn const_eval_select<R, F: FnOnce() -> R, G: const FnOnce() -> R>(f: F) -> R {
>     f()
> }
> ```
> 
> Then the typeck change could be just checking that F and G are function items, while leaving all the rest (signature, constness, ..) to the regular typeck logic?

I don't think function item check would be necessary then.

oli-obk:
🙈 I thought I r+ed that already...

Only allowing function items, in order to make it very explicit what is going on still stands, but it could "simply" be circumvented via `const_eval_select(call_first, const_fn_call, (|x| x.stuff(), real_arg))` where `fn call_first<T>((f, arg): (impl FnOnce(T), T)) {f(arg)}`, so I guess you're right, we can just allow closures.

oli-obk:
the impl looks good to me now.

Since this is a bit different from intrinsics' const stability, please add a test checking that stable const fn cannot depend on `const_eval_select` for now

oli-obk:
r=me with CI passing

PR #89251, opened 2021-09-25
Detect when negative literal indices are used and suggest appropriate code

estebank:


NieDzejkob:
Suggested test case (no error should trigger):

```rust
use core::ops::Index;
struct X;
impl Index<i32> for X {
    type Output = ();
    
    fn index(&self, _: i32) -> &() {
        &()
    }
}

fn main() {
    X[-1];
}
```

estebank:
@NieDzejkob added your case to the test.

compiler/rustc_typeck/src/check/place_op.rs
```diff
@@ -73,6 +76,7 @@ impl<'a, 'tcx> FnCtxt<'a, 'tcx> {
         base_expr: &hir::Expr<'_>,
         autoderef: &Autoderef<'a, 'tcx>,
         index_ty: Ty<'tcx>,
+        idx_expr: &hir::Expr<'_>,
```
davidtwco, line 79: nit: looks like in other functions, the type argument follows the corresponding expression argument, and this might be more consistently named `index_expr` in this function

compiler/rustc_typeck/src/check/place_op.rs
```diff
@@ -82,6 +86,53 @@ impl<'a, 'tcx> FnCtxt<'a, 'tcx> {
...
+                &format!("cannot use a negative integer for indexing on `{}`", ty),
+            );
+            if let (hir::ExprKind::Path(..), Ok(snippet)) =
+                (&base_expr.kind, self.tcx.sess.source_map().span_to_snippet(base_expr.span))
+            {
+                // `foo[-1]` to `foo[foo.len() - 1]`
+                err.span_suggestion_verbose(
+                    idx_expr.span.shrink_to_lo(),
+                    &format!(
+                        "if you wanted to access an element starting from the end of the `{}`, you \
```
davidtwco, line 106:
```suggestion
                        "to access an element starting from the end of the `{}`, compute the index"
```
nit: this wording is more concise

compiler/rustc_typeck/src/check/place_op.rs
```diff
@@ -82,6 +86,53 @@ impl<'a, 'tcx> FnCtxt<'a, 'tcx> {
             expr, base_expr, adjusted_ty, index_ty
         );
 
+        let negative_index = || {
```
davidtwco, line 89: Could you add a sibling function for this logic rather than a closure? It's nice if we can keep the diagnostic logic as off-to-the-side as possible and out of the main functions so as not to impede understanding of the function's purpose.

davidtwco (changes requested):
This is looking really good, some minor changes. r=me after fixing.

PR #32586, opened 2016-03-29
specialize ToString for str

seanmonstar:
If there was some conditional compiling we could do, such that this impl only exists in nightly, and is turned off in beta/stable, I think that'd be an improvement here, as we could test specialization out without affecting stable builds.

PR #89253, opened 2021-09-25
Hide `ToString` specializations behind a private trait

SkiFire13:
Normally we use private traits for specializing `impl`s but `ToString`'s were public and the original PR #32586 doesn't seem to have a reason for this. This PR moves those impls to a hidden perma-unstable `SpecToString` trait (I would have made it private if `proc_macro` didn't have to specialize it too) and delegates to it, leaving only the `impl<T: fmt::Display + ?Sized> ToStringSpec for T` as public.

### DIFF
--- a/library/alloc/src/string.rs
+++ b/library/alloc/src/string.rs
@@ -2375,7 +2375,22 @@ impl<T: fmt::Display + ?Sized> ToString for T {
     // See <https://github.com/rust-lang/rust/pull/74852>, the last attempt
     // to try to remove it.
     #[inline]
-    default fn to_string(&self) -> String {
+    fn to_string(&self) -> String {
+        SpecToString::spec_to_string(self)
+    }
+}
+
+#[unstable(issue = "none", feature = "spec_to_string")]
+#[doc(hidden)]
+pub trait SpecToString {
+    fn spec_to_string(&self) -> String;
+}
+
+#[cfg(not(no_global_oom_handling))]
+#[unstable(issue = "none", feature = "spec_to_string")]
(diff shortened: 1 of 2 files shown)


kennytm:
but isn't `proc_macro` specializing `ToString` an example why the `default fn` should be kept public 🤔

SkiFire13:
`proc_macro` will still use the specialized version, just not publicly

kennytm:
actually https://github.com/rust-lang/rust/pull/32586#issuecomment-204292183 questioned the rationale of making `ToString` publicly specializable, the response was

> Very good question. We talked about this some in a recent libs meeting -- ultimately, we're going to want a conventions RFC here, I think. But it's hard to get there without some experimentation first.

what i interpret is that there's no preference in wanting either way. afaik no conventions RFC exists yet. has any experimentation been done?

&nbsp;

IMO creating a private-but-actually-public `SpecToString` trait to hide away the specialization possibility does not provide much benefits over keeping `ToString` specializable, as long as #31844 `min_specialization` is still unstable.

SkiFire13:
> what i interpret is that there's no preference in wanting either way. afaik no conventions RFC exists yet

While not technically being an RFC, the std-dev-guide specifies that specialization should use private traits https://std-dev-guide.rust-lang.org/code-considerations/using-unstable-lang/specialization.html

> IMO creating a private-but-actually-public SpecToString trait to hide away the specialization possibility does not provide much benefits over keeping ToString specializable, as long as #31844 min_specialization is still unstable.

I agree with that, however at least the change makes it consistent with how specialization is handled in the rest of the stdlib.

PR #89257, opened 2021-09-25
Give better error for `macro_rules name`

aDotInTheVoid:
follow up to #89221

r? @estebank

@rustbot modify labels: +A-diagnostics +A-parser

### DIFF
--- a/compiler/rustc_parse/src/parser/item.rs
+++ b/compiler/rustc_parse/src/parser/item.rs
@@ -279,9 +279,9 @@ impl<'a> Parser<'a> {
         } else if self.eat_keyword(kw::Macro) {
             // MACROS 2.0 ITEM
             self.parse_item_decl_macro(lo)?
-        } else if self.is_macro_rules_item() {
+        } else if let IsMacroRulesItem::Yes { has_bang } = self.is_macro_rules_item() {
             // MACRO_RULES ITEM
-            self.parse_item_macro_rules(vis)?
+            self.parse_item_macro_rules(vis, has_bang)?
         } else if vis.kind.is_pub() && self.isnt_macro_invocation() {
             self.recover_missing_kw_before_item()?;
             return Ok(None);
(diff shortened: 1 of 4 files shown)


compiler/rustc_parse/src/parser/item.rs
```diff
@@ -1534,18 +1534,45 @@ impl<'a> Parser<'a> {
...
-    /// Is this unambiguously the start of a `macro_rules! foo` item definition?
-    fn is_macro_rules_item(&mut self) -> bool {
-        self.check_keyword(kw::MacroRules)
-            && self.look_ahead(1, |t| *t == token::Not)
-            && self.look_ahead(2, |t| t.is_ident())
+    /// Is this a possibly malformed start of a `macro_rules! foo` item definition?
+    ///
+    /// The first bool is whether it's a macro_rules item.
+    /// The second is whether the `!` after `macro_rules` is present
+    fn is_macro_rules_item(&mut self) -> (bool, bool) {
```
camelid, line 1541: nit: Seems like this code might be clearer if the return type were a struct with two boolean fields? E.g.:

```rust
struct IsMacroRulesItem {
    is_macro_rules: bool,
    has_bang_after_macro_rules: bool,
}
```

compiler/rustc_parse/src/parser/item.rs
```diff
@@ -1534,18 +1534,45 @@ impl<'a> Parser<'a> {
...
+        if self.check_keyword(kw::MacroRules) {
+            let macro_rules_span = self.look_ahead(0, |t| t.span);
```
estebank, line 1543: Isn't this equivalent to `self.token.span`?

estebank (approved):
After addressing the nitpicks, r=me

compiler/rustc_parse/src/parser/item.rs
camsteffen: Or return `Option<IsMacroRulesItem>` with `struct IsMacroRulesItem { has_bang: bool }`

compiler/rustc_parse/src/parser/item.rs
aDotInTheVoid: Yep, didn't realise you could do that

compiler/rustc_parse/src/parser/item.rs
aDotInTheVoid: Went with 

```rust
enum IsMacroRulesItem {
    Yes { has_bang: bool },
    No,
}
```

we could bikeshed this forever, but I think this is pritty clear

PR #89258, opened 2021-09-25
Make char conversion functions unstably const

est31:
The char conversion functions like `char::from_u32` do trivial computations and can easily be converted into const fns. Only smaller tricks are needed to avoid non-const standard library functions like `Result::ok` or `bool::then_some`.

Tracking issue: https://github.com/rust-lang/rust/issues/89259

### DIFF
--- a/library/core/src/char/convert.rs
+++ b/library/core/src/char/convert.rs
@@ -51,8 +51,13 @@ use super::MAX;
 #[must_use]
 #[inline]
 #[stable(feature = "rust1", since = "1.0.0")]
-pub fn from_u32(i: u32) -> Option<char> {
-    char::try_from(i).ok()
+#[rustc_const_unstable(feature = "const_char_convert", issue = "89259")]
+pub const fn from_u32(i: u32) -> Option<char> {
+    // FIXME: once Result::ok is const fn, use it here
+    match char_try_from_u32(i) {
+        Ok(c) => Some(c),
+        Err(_) => None,
+    }
 }
 
 /// Converts a `u32` to a `char`, ignoring validity.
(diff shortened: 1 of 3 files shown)


library/core/src/char/convert.rs
```diff
@@ -50,8 +50,12 @@ use super::MAX;
 #[doc(alias = "chr")]
 #[inline]
 #[stable(feature = "rust1", since = "1.0.0")]
-pub fn from_u32(i: u32) -> Option<char> {
-    char::try_from(i).ok()
+#[rustc_const_unstable(feature = "const_char_convert", issue = "89259")]
+pub const fn from_u32(i: u32) -> Option<char> {
+    match char_try_from_u32(i) {
```
r00ster91, line 55: I wish we wouldn't have to sacrifice readability for constability, especially because we can eventually have both when `Result::ok` is `const` too. How about adding a `FIXME` or similar to change this back later?

library/core/src/char/methods.rs
```diff
@@ -340,7 +344,7 @@ impl char {
             // Force the 6th bit to be set to ensure ascii is lower case.
             digit = (self as u32 | 0b10_0000).wrapping_sub('a' as u32).saturating_add(10);
         }
-        (digit < radix).then_some(digit)
+        if digit < radix { Some(digit) } else { None }
```
r00ster91, line 347: Same here.

library/core/src/char/convert.rs
est31: I've tried making `Result::ok` const, but it is blocked by const drop.

r00ster91: I suggested to add a `FIXME`, not to fix it immediately.

Mark-Simulacrum:
I admit that I feel a little "meh" about small const-ifications that also need to change the bodies of functions to make them amenable to const-ifying (or even extracting them into helper functions for sharing). It seems like there is no particularly pressing need to do so, and const-fn evaluation is progressing gradually to the point where that becomes less necessary to do.

That said, this PR seems OK to me.

r? @yaahc for libs-api approval

est31:
@mark-simulacrum does this need libs-api approval? @oli-obk has recently r+'d a very similar PR #90607 without libs-api approval.

Mark-Simulacrum:
I think current policy is const eval OR libs-api, but am not confident on that.

oli-obk:
wg-const-eval may unstably add const to functions without querying libs-api

PR #89263, opened 2021-09-26
Suggest both of immutable and mutable trait implementations

### LINKED ISSUE #85865: Suggest borrowing if a trait implementation is found for &/&mut <type> regression
**When both `&S` and `&mut S` is implemented**,

Given the following code: 

```rust
trait Tr {}
struct S {}

impl Tr for &S {}
impl Tr for &mut S {}

fn foo<T: Tr>(t: T) {}

fn main() {
    let s = S {};
    foo(s);
}
```

The previous output was:
```
error[E0277]: the trait bound `S: Tr` is not satisfied
  --> src/main.rs:11:9
   |
7  | fn foo<T: Tr>(t: T) {}
   |           -- required by this bound in `foo`
...
11 |     foo(s);
   |         ^ the trait `Tr` is not implemented for `S`
   |
   = help: the following implementations were found:
             <&S as Tr>
             <&mut S as Tr>
```

The current output is:

```
error[E0277]: the trait bound `S: Tr` is not satisfied
  --> src/main.rs:11:9
   |
7  | fn foo<T: Tr>(t: T) {}
   |           -- required by this bound in `foo`
...
11 |     foo(s);
   |         ^
   |         |
   |         expected an implementor of trait `Tr`
   |         help: consider borrowing here: `&s`
```

Not sure but it should show both `&s` and `&mut s`. The previous output seemed to do a better job there. Maybe we could keep the last part of the previous output.
```
   = help: the following implementations were found:
             <&S as Tr>
             <&mut S as Tr>
```

Caused by https://github.com/rust-lang/rust/pull/85369, although it is better in the sense it is more specific but I think it omitted quite some useful details.

TaKO8Ki:
closes #85865


compiler/rustc_trait_selection/src/traits/error_reporting/suggestions.rs
```diff
@@ -773,15 +779,24 @@ impl<'a, 'tcx> InferCtxtExt<'tcx> for InferCtxt<'a, 'tcx> {
...
-                            format!("&{}{}", if mtbl { "mut " } else { "" }, snippet),
-                            Applicability::MaybeIncorrect,
-                        );
+                        if imm_result && mut_result {
+                            err.span_suggestions(
+                                span,
+                                "consider borrowing here",
+                                [format!("&{}", snippet), format!("&mut {}", snippet)].into_iter(),
+                                Applicability::MaybeIncorrect,
+                            );
```
estebank, line 788:
```suggestion
                            err.span_suggestions(
                                span.shrink_to_lo(),
                                "consider borrowing here",
                                ["&".to_string(), "&mut ".to_string()].into_iter(),
                                Applicability::MaybeIncorrect,
                            );
```

compiler/rustc_trait_selection/src/traits/error_reporting/suggestions.rs
```diff
@@ -773,15 +779,24 @@ impl<'a, 'tcx> InferCtxtExt<'tcx> for InferCtxt<'a, 'tcx> {
...
+                        } else {
+                            err.span_suggestion(
+                                span,
+                                &format!(
+                                    "consider{} borrowing here",
+                                    if mut_result { " mutably" } else { "" }
+                                ),
+                                format!("&{}{}", if mut_result { "mut " } else { "" }, snippet),
```
estebank, line 796: This change isn't necessary, but might look better:

```suggestion
                            err.span_suggestion_verbose(
                                span.shrink_to_lo(),
                                &format!(
                                    "consider{} borrowing here",
                                    if mut_result { " mutably" } else { "" }
                                ),
                                format!("&{}", if mut_result { "mut " } else { "" }),
```

I _would_ like us to add a test exercising the `consider mutably borrowing here` case, for which you might have to write a new trait impl that's only valid if it is a mutable borrow (and likely one for the non mutable case as well).

estebank (approved):
The changes look good, just left a couple of nitpicks.

rylev:
@TaKO8Ki @estebank The regression occurs in `evaluate_obligation` which seems like it would be effected by this change, but the only impacted benchmark is diesel doc which doesn't trigger this diagnostic, so I'm not sure I see where this regression could be coming from. The change is large enough though that it being noise would be quite surprising. cc @jyn514

jyn514:
I don't know when `suggest_add_reference_to_arg` is called, but rustdoc does make a lot more calls to `evaluate_obligation` than rustc because it's trying to collect an exhaustive list of blanket impls.

jyn514:
Maybe LLVM is making different inlining decisions or something?

estebank:
@rylev @jyn514 this is as surprising to me as it is to you :-/

PR #89266, opened 2021-09-26
Move ICH to rustc_query_system

cjgillot:
Based on https://github.com/rust-lang/rust/pull/89183

The StableHashingContext does not need to be in rustc_middle.

This PR moves it to rustc_query_system. This will avoid a dependency between rustc_ast_lowering and rustc_middle in https://github.com/rust-lang/rust/pull/89124.


davidtwco (commented):
This all seems sensible to me, but I'll reassign to get another opinion.

cjgillot:
Actually, putting it in rustc_query_system would be more consistent.

michaelwoerister (approved):

michaelwoerister:
Looks good to me. Thanks, @cjgillot!

Let's do a perf run for good measure:
@bors try @rust-timer queue

If that doesn't yield any surprises, r=davidtwco,michaelwoerister after rebasing and updating the PR message.

michaelwoerister:
Performance looks good now 👍 

@bors r+

Issue #59117, opened 2019-03-11
Add function to make paths absolute, which is different from canonicalization


retep998:
Many users want to turn relative paths into absolute paths, but the only tool that libstd currently offers is `canonicalize` which is bad because it has to hit the filesystem which has several downsides:

1. It takes time to hit the filesystem.
2. The path has to actually exist.
3. It fails on certain types of drives such as RAM drives on Windows.
4. The path it creates is a `\\?\` path which not all software can handle correctly and imposes requirements on further path manipulation that few users are even aware of.

Needing symbolic links actually resolved is an extremely rare use case, and is often misused to compare paths for equality when in reality you should be comparing file IDs due to things like hard links existing.

A new function (`normalize` or `make_absolute` or something, bikeshed away) should be added that will turn a relative path into an absolute path without touching the filesystem. On Windows this should either call `GetFullPathNameW` or do the pure Rust equivalent while on unixy platforms... *something* should happen, I have no idea.

If such a function did exist, rustc could start using that instead of canonicalize which would fix a whole host of issues including:
https://github.com/rust-lang/rust/issues/74327
https://github.com/rust-lang/rust/pull/74146
https://github.com/rust-lang/rust/issues/59107
https://github.com/rust-lang/rust/issues/58613
https://github.com/rust-lang/rust/issues/55812
https://github.com/rust-lang/rust/issues/52440
https://github.com/rust-lang/rust/issues/48249
https://github.com/rust-lang/rust/issues/45067
https://github.com/rust-lang/rust/issues/42869

Mark-Simulacrum:
Semi-related issues/PRs: https://github.com/rust-lang/rust/issues/47402, https://github.com/rust-lang/rust/pull/47363, and https://github.com/rust-lang/rfcs/issues/2208.

retep998:
One big question for the libs team to decide on is how to handle `..` and `.` on linux. On Windows `C:\foo\..\bar` is equivalent to `C:\bar` and both paths will always point to the same thing. On linux however it is possible for `/foo/../bar` to not point to the same thing as `/bar`. Given a current directory of `/foo` and a relative path of `../bar`, should `normalize` on linux return `/foo/../bar` or `/bar`?

nayeemrmn:
A function called `normalize()` will not be expected to convert to an absolute path, that name should be used for https://github.com/rust-lang/rfcs/issues/2208.

nayeemrmn:
[...]

PR #89270, opened 2021-09-26
path.push() should work as expected on windows verbatim paths

seanyoung:
On Windows, std::fs::canonicalize() returns an so-called UNC path.  UNC paths differ with regular paths because:

- This type of path can much longer than a non-UNC path (32k vs 260 characters). 
- The prefix for a UNC path is ``Component::Prefix(Prefix::DiskVerbatim(..)))``
- No `/` is allowed
- No `.` is allowed
- No `..` is allowed

Rust has poor handling of such paths. If you join a UNC path with a path with any of the above, then this will not work.

I've implemented a new method `fn join_fold()` which joins paths and also removes any `.` and `..` from it, and replaces `/` with `\` on Windows. Using this function it is possible to use UNC paths without issue. In addition, this function is useful on Linux too; paths can be appended without having to call `canonicalize()` to remove the `.` and `..`.

This PR needs test cases, which can I add. I hope this will a start of a discussion.

### DIFF
--- a/library/std/src/path.rs
+++ b/library/std/src/path.rs
@@ -1231,19 +1231,58 @@ impl PathBuf {
         let mut need_sep = self.as_mut_vec().last().map(|c| !is_sep_byte(*c)).unwrap_or(false);
 
         // in the special case of `C:` on Windows, do *not* add a separator
+        let comps = self.components();
+
+        if comps.prefix_len() > 0
+            && comps.prefix_len() == comps.path.len()
+            && comps.prefix.unwrap().is_drive()
         {
-            let comps = self.components();
-            if comps.prefix_len() > 0
-                && comps.prefix_len() == comps.path.len()
-                && comps.prefix.unwrap().is_drive()
-            {
-                need_sep = false
-            }
+            need_sep = false
         }
 
         // absolute `path` replaces `self`
         if path.is_absolute() || path.prefix().is_some() {
             self.as_mut_vec().truncate(0);
+
+        // verbatim paths need . and .. removed
+        } else if comps.prefix_verbatim() {
+            let mut buf: Vec<_> = comps.collect();
+            for c in path.components() {
+                match c {
+                    Component::RootDir => {
+                        buf.truncate(1);
+                        buf.push(c);
+                    }
+                    Component::CurDir => (),
+                    Component::ParentDir => {
+                        if let Some(Component::Normal(_)) = buf.last() {
+                            buf.pop();
+                        }
+                    }
+                    _ => buf.push(c),
+                }
+            }
+
+            let mut res = OsString::new();
+            let mut need_sep = false;
+
+            for c in buf {
+                if need_sep && c != Component::RootDir {
+                    res.push(MAIN_SEP_STR);
+                }
+                res.push(c.as_os_str());
+
+                need_sep = match c {
+                    Component::RootDir => false,
+                    Component::Prefix(prefix) => {
+                        !prefix.parsed.is_drive() && prefix.parsed.len() > 0
+                    }
+                    _ => true,
+                }
+            }
+
+            self.inner = res;
+            return;
 
         // `path` has a root but no prefix, e.g., `\windows` (Windows only)
         } else if path.has_root() {
--- a/library/std/src/path/tests.rs
+++ b/library/std/src/path/tests.rs
@@ -1262,6 +1262,15 @@ pub fn test_push() {
         tp!("\\\\.\\foo", "..\\bar", "\\\\.\\foo\\..\\bar");
 
         tp!("\\\\?\\C:", "foo", "\\\\?\\C:\\foo"); // this is a weird one
+
+        tp!(r"\\?\C:\bar", "../foo", r"\\?\C:\foo");
+        tp!(r"\\?\C:\bar", "../../foo", r"\\?\C:\foo");
+        tp!(r"\\?\C:\", "../foo", r"\\?\C:\foo");
+        tp!(r"\\?\C:", r"D:\foo/./", r"D:\foo/./");
+        tp!(r"\\?\C:", r"\\?\D:\foo\.\", r"\\?\D:\foo\.\");
+        tp!(r"\\?\A:\x\y", "/foo", r"\\?\A:\foo");
+        tp!(r"\\?\A:", r"..\foo\.", r"\\?\A:\foo");
+        tp!(r"\\?\A:\x\y", r".\foo\.", r"\\?\A:\x\y\foo");
     }
 }
 


the8472:
This mixes two orthogonal functions `join` and an [often-proposed](https://github.com/rust-lang/rust/issues/59117) but currently not implemented `normalize`. What's the benefit of putting them into a single method?

> * No `.` is allowed
> * No `..` is allowed

The [microsoft docs](https://docs.microsoft.com/en-us/windows/win32/fileio/naming-a-file#win32-file-namespaces) say that those actually are allowed in paths with the verbatim prefix and will access files named `.` and `..` instead of traversing directories. So interpreting those differently can have the opposite of the intended effect of using the verbatim prefix.

ChrisDenton:
I think it would be great to have something that addresses this but there I there should perhaps be a holistic look at Windows path handling in Rust before adding adhoc methods as there are currently numerous open issues with paths. Discussions like this are probably best had on the [internals](https://internals.rust-lang.org/) forum or [Zulip](https://rust-lang.zulipchat.com/).

Personally speaking, my thinking at the moment is that `push` should "just work" for verbatim paths, with the path being pushed normalized according to platform conventions as this is what's wanted 99.999999% of the time (approx). There might have to be a `push_verbatim` for pushing paths that you want to include `.`, `..` or `/` components. Though it would be very rarely used because Windows filesystems do not allow these as file names (and indeed they would screw up non-verbatim paths if they were allowed).

I guess `path.push(normalize(subpath))` would also work but I'd prefer the common case to be simpler to do than the exceedingly rare case. Otherwise it's too easy to do the wrong thing unless you have knowledge of the arcane corners of Windows.

seanyoung:
> I think it would be great to have something that addresses this but there I there should perhaps be a holistic look at Windows path handling in Rust before adding adhoc methods as there are currently numerous open issues with paths. Discussions like this are probably best had on the [internals](https://internals.rust-lang.org/) forum or [Zulip](https://rust-lang.zulipchat.com/).

That is probably a good idea. In order to facilitate that discussion, I'm updating this PR.

> Personally speaking, my thinking at the moment is that `push` should "just work" for verbatim paths, with the path being pushed normalized according to platform conventions as this is what's wanted 99.999999% of the time (approx). There might have to be a `push_verbatim` for pushing paths that you want to include `.`, `..` or `/` components. Though it would be very rarely used because Windows filesystems do not allow these as file names (and indeed they would screw up non-verbatim paths if they were allowed).

This is a good point and I totally agree this. I've updated this PR to do exactly that.

> I guess `path.push(normalize(subpath))` would also work but I'd prefer the common case to be simpler to do than the exceedingly rare case. Otherwise it's too easy to do the wrong thing unless you have knowledge of the arcane corners of Windows.

`path.push(normalize(subpath))` would not work. If subpath is `../foo` then `path.push()` still needs to have special handling for verbatim paths.

seanyoung:
> This mixes two orthogonal functions `join` and an [often-proposed](https://github.com/rust-lang/rust/issues/59117) but currently not implemented `normalize`. What's the benefit of putting them into a single method?

`normalize()` would not solve the issue. I have changed the PR so now `join()` works with verbatim paths.

> > * No `.` is allowed
> > * No `..` is allowed
> 
> The [microsoft docs](https://docs.microsoft.com/en-us/windows/win32/fileio/naming-a-file#win32-file-namespaces) say that those actually are allowed in paths with the verbatim prefix and will access files named `.` and `..` instead of traversing directories. So interpreting those differently can have the opposite of the intended effect of using the verbatim prefix.

I don't think this is true. I've tried creating `.` and `..` files using unc paths and it does not work. Besides, even the folks at microsoft are not insane. This would be a terrible idea.

the8472:
> normalize() would not solve the issue.

Can you explain how it wouldn't solve it? 

> I don't think this is true. I've tried creating . and .. files using unc paths and it does not work. 

That might depend on the filesystem. I think it's similar on linux. On most filesystems `.` point will do the expected thing but with FUSE you can create funky directories where it behaves differently or doesn't exist at all.

seanyoung:
> > normalize() would not solve the issue.
> 
> Can you explain how it wouldn't solve it?

Consider`path.join(normalize("../foo"))`:
- `normalize()` would return `../foo` or `..\foo` on Windows
- `path.join()` would still have to process `..` and remove the last directory from self before appending `foo`.

> > I don't think this is true. I've tried creating . and .. files using unc paths and it does not work.
> 
> That might depend on the filesystem. I think it's similar on linux. On most filesystems `.` point will do the expected thing but with FUSE you can create funky directories where it behaves differently or doesn't exist at all.

Your original point was about Windows, so I tried this on Windows:
```
use std::fs::File;
use std::io::Write;

fn main() {
    let mut file =
        File::create(r"\\?\..").expect("create dot file");
    file.write_all(b"Hello, world!").expect("write dot file");
}
```
(I tried `.`, `..` with relative and absolute paths, does not work).

On Linux `.` and `..` are special names which are handled in kernel space. You cannot create `.` or `..` files, it does not depend on the filesystem, fuse or otherwise. See https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/fs/namei.c#n2215

(BTW I'm a kernel maintainer)

the8472:
> > > normalize() would not solve the issue.
> > 
> > 
> > Can you explain how it wouldn't solve it?
> 
> Consider`path.join(normalize("../foo"))`:
> 
>  * `normalize()` would return `../foo` or `..\foo` on Windows
>  * `path.join()` would still have to process `..` and remove the last directory from self before appending `foo`.

Then why not `normalize(path.join("../foo"))`?


> On Linux . and .. are special names which are handled in kernel space. You cannot create . or .. files, it does not depend on the filesystem, fuse or otherwise. 

Maybe it's limited to dir entry enumeration, but I do recall getting surprising results when using FUSE



> ```
> use std::fs::File;
> use std::io::Write;
> 
> fn main() {
>     let mut file =
>         File::create(r"\\?\..").expect("create dot file");
>     file.write_all(b"Hello, world!").expect("write dot file");
> }
> ```
> 
> (I tried `.`, `..` with relative and absolute paths, does not work).

But did you try with different filesystems? E.g. FAT? I don't have a windows machine here, otherwise I'd poke around myself.

I assume there is a reason why the documentation says:

> Because it turns off automatic expansion of the path string, the "\\?\" prefix also allows the use of ".." and "." in the path names, which can be useful if you are attempting to perform operations on a file with these otherwise reserved relative path specifiers as part of the fully qualified path.

seanyoung:
> > Consider`path.join(normalize("../foo"))`:
> > 
> > * `normalize()` would return `../foo` or `..\foo` on Windows
> > * `path.join()` would still have to process `..` and remove the last directory from self before appending `foo`.
> 
> Then why not `normalize(path.join("../foo"))`?

Because I assumed that `path.join("../foo"))` returns garbage, and you're using normalize to clean up the mess. I'm not so sure now, see below.

> > On Linux . and .. are special names which are handled in kernel space. You cannot create . or .. files, it does not depend on the filesystem, fuse or otherwise.
> 
> Maybe it's limited to dir entry enumeration, but I do recall getting surprising results when using FUSE

path walks are done for file creation too. It would be interesting to hear more about "surprising results".

> > ```
> > use std::fs::File;
> > use std::io::Write;
> > 
> > fn main() {
> >     let mut file =
> >         File::create(r"\\?\..").expect("create dot file");
> >     file.write_all(b"Hello, world!").expect("write dot file");
> > }
> > ```
> > 
> > (I tried `.`, `..` with relative and absolute paths, does not work).
> 
> But did you try with different filesystems? E.g. FAT? I don't have a windows machine here, otherwise I'd poke around myself.
> 
> I assume there is a reason why the documentation says:
> 
> > Because it turns off automatic expansion of the path string, the "\?" prefix also allows the use of ".." and "." in the path names, which can be useful if you are attempting to perform operations on a file with these otherwise reserved relative path specifiers as part of the fully qualified path.

Now this is interesting - and I didn't expect this. With FAT and exFAT you can create files called `.` and `..`. They're only accessible using UNC paths, but they show up with dir. Normally dir does not show `.` or `..`.

If you then mount the file system on Linux, it cannot deal with the `.` or `..` files. This is must be a bug in the linux exfat driver. It actually enumerates them as directories.

So now the question is, if verbatim paths permit `..` and `.` on Windows and (ex)fat file systems, then what do we do? What does `PathBuf::from("\\?\bar").join("../foo")` mean?
- `\\?\bar\../foo`
- `\\?\foo`

I can see the appeal of the `fn normalize()` as you proposed.

ChrisDenton:
Btw, I wrote [a thing about NT and Win32 paths](https://gist.github.com/ChrisDenton/f432ca0003cf25520b448972e11f9097), in case it helps.

The short version is NT paths can include anything except `\` (including `NUL`). What I didn't mention there is that the NT kernel does only the minimum to resolve the relevant device, the rest of the path it simply passes to that device to resolve. The device can do anything with that sub-path.

So these NT paths are essentially impossible to handle sensibly with Rust's `Path`/`PathBuf` without special knowledge of the device being used. Maybe the device uses `$` as the path separator? In that case `push` is already broken. Maybe `.` is a filename? In that case `components` is broken. Etc, etc.

I think `Path` methods should assume a "normal" filesystem path as the default and work as expected without the programmer needing special knowledge of the arcane workings of the NT kernel. Special methods could be added to address the special cases but understanding them should not be necessary for writing cross-platform software.

seanyoung:
https://lore.kernel.org/linux-fsdevel/20210927111948.GA16257@gofer.mess.org/T/

seanyoung:
> Btw, I wrote [a thing about NT and Win32 paths](https://gist.github.com/ChrisDenton/f432ca0003cf25520b448972e11f9097), in case it helps.
> 
> The short version is NT paths can include anything except `\` (including `NUL`). What I didn't mention there is that the NT kernel does only the minimum to resolve the relevant device, the rest of the path it simply passes to that device to resolve. The device can do anything with that sub-path.

Very interesting.
 
> So these NT paths are essentially impossible to handle sensibly with Rust's `Path`/`PathBuf` without special knowledge of the device being used. Maybe the device uses `$` as the path separator? In that case `push` is already broken. Maybe `.` is a filename? In that case `components` is broken. Etc, etc.
> 
> I think `Path` methods should assume a "normal" filesystem path as the default and work as expected without the programmer needing special knowledge of the arcane workings of the NT kernel. Special methods could be added to address the special cases but understanding them should not be necessary for writing cross-platform software.

I concur completely. I would much rather see path join handling of `.` and `..` do what you expect, and do what the rest of the path handling does. The fact that _some_ windows filesystem drivers can do files called `.` and `..` is an oddity which rust `Path`/`PathBuf` can do without.

If anyone wants to create a path like `\\?\D:\foo\..`, then that is still possible. I don't think it would surprise anyone that a `Path`/`PathBuf` join thinks this is a walk to the parent of `foo`.

the8472:
Started a thread on zulip. https://rust-lang.zulipchat.com/#narrow/stream/219381-t-libs/topic/windows.20verbatim.20paths.2C.20normalization.20vs.2E.20canonicalize.20etc.2E

the8472:
r? rust-lang/libs

seanyoung:
@m-ou-se I've made some last changes after a lot more testing. Ready for review

m-ou-se:
Thanks!

@bors r+

PR #81500, opened 2021-01-29
Remove struct_type from union output

CraftSpider:
Also bumps the format number and adds a test

Rationale: It's illegal to have unions of the form `union Union(i32, f32);`, or `union Union;`. The struct_type field was recently removed from the rustdoc Union AST, at which time this field was changed to always just read "union". It makes sense to completely remove it, as it provides no information.

PR #89276, opened 2021-09-26
Fix the population of the `union.impls` field

Urgau:
This pull-request fix the population of the `union.impls` field that was forgot when the `Union` type was introduce as a split from the `Struct` type https://github.com/rust-lang/rust/pull/81500.

@rustbot label +T-rustdoc +A-rustdoc-json


GuillaumeGomez (approved):

GuillaumeGomez:
Thanks!

@bors: r+ rollup

Issue #89135, opened 2021-09-21
2021 rustdoc INVALID_RUST_CODEBLOCKS "breakage"


Mark-Simulacrum:
rustdoc currently has a warn-by-default lint, `rustdoc::invalid_rust_codeblocks`, which is intended to tell users when rustdoc sees a Rust language code block which is not valid Rust. However, this lint seems to have always been pretty broken: it only attempted to tokenize the input, not actually parse it into a Rust program. So in most cases before almost any text passed in would've not triggered the lint (e.g., the code blocks on this page -- which are clearly not valid Rust -- do not: https://doc.rust-lang.org/1.55.0/nightly-rustc/rustc_metadata/dependency_format/index.html).

However, with the 2021 edition, we have reserved ~any prefix in strings (`foo'bar`, for example), which causes this lint to fire on quite a lot of this previously accepted Rust code. This is not *technically* a breaking change -- the lint is warn by default, and is just a lint -- but in practice I worry that the perception will be quite a lot of breakage. The compiler hit quite a few [cases](https://github.com/rust-lang/rust/commit/49ed5cefcedba18f4d2d2757e44c72dec69effc0) in the migration, which all had to be fixed manually -- no rust fix exists here, and can't easily be added I suspect.

camelid:
cc @rust-lang/rustdoc

jyn514:
Yet more discussion: https://rust-lang.zulipchat.com/#narrow/stream/268952-edition-2021/topic/Edition.20Blockers.28.3F.29

GuillaumeGomez:
Can we close this issue now that #89277? Otherwise, what remains to be done?

Mark-Simulacrum:
Yes, looks like the beta backport is also in.

PR #89277, opened 2021-09-26
Use the correct edition for syntax highlighting doctests

jyn514:
Previously it would unconditionally use edition 2015, which was incorrect.

Helps with https://github.com/rust-lang/rust/issues/89135 in that you can now override the doctest to be 2018 edition instead of being forced to fix the error. This doesn't resolve any of the deeper problems that rustdoc disagrees with most rust users on what a code block is.

cc @Mark-Simulacrum


Mark-Simulacrum:
beta-nominating, needed for 2021 edition

GuillaumeGomez (approved):

GuillaumeGomez:
Thanks!

@bors: r+

PR #89278, opened 2021-09-26
Remove LocalDefId <-> HirId global maps

cjgillot:
~Based on https://github.com/rust-lang/rust/pull/89124.~

The LocalDefId <-> HirId maps were globally accessed through `tcx.definitions`. They are now encoded using the HIR queries.
- For HirId -> LocalDefId, a side map is added inside `hir_owner_nodes(hir_id.owner)`, indexed by `hir_id.local_id`.
- For LocalDefId -> HirId, the mapping is obvious if the `local_def_id` is an HIR owner, ie. if `hir_owner` is `Some`. Therefore, we  use the `None` case to store the non-owning `HirId`.

There is a special case with `use` trees, which can have a LocalDefId, which should be an HIR owner, but with no associated HIR data structure. In that case, we store the information like non-owning definitions.

### DIFF
--- a/compiler/rustc_ast_lowering/src/item.rs
+++ b/compiler/rustc_ast_lowering/src/item.rs
@@ -105,12 +105,11 @@ impl<'hir> LoweringContext<'_, 'hir> {
     ) -> T {
         let old_len = self.in_scope_lifetimes.len();
 
-        let parent_generics =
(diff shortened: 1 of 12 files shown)


oli-obk:
Should we rerun perf or is this just a rebase?

cjgillot:
This is just a rebase. I haven't figured how to reduce the regression yet.

cjgillot:
Replaced by #93373.

PR #89282, opened 2021-09-26
2229: Consume IfLet expr

### LINKED ISSUE #88118: `PlaceBuilder` ICE, Part 2
#88039 fixed #87987 but this issue still fails:

```rust
#![feature(capture_disjoint_fields)]

fn foo<MsU>(handler: impl FnOnce() -> MsU + Clone + 'static) {
    Box::new(move |value| {
        (|_| handler.clone()())(value);
        None
    }) as Box<dyn Fn(i32) -> Option<i32>>;
}

fn main() {}
```

Found trying to migrate `seed 0.8.0`.

Let me know if you need any more information.

_Originally posted by @ehuss in https://github.com/rust-lang/rust/issues/87987#issuecomment-899910552_

roxelo:
When using the IfLet guard feature, we can ICE when attempting to resolve PlaceBuilders.
For pattern matching, we currently don't consume the IfLet expression when "visiting" the arms leading us to not "read" all variables and hence not being able to resolve them.

r? @nikomatsakis 
Closes https://github.com/rust-lang/rust/issues/88118

### DIFF
--- a/compiler/rustc_typeck/src/expr_use_visitor.rs
+++ b/compiler/rustc_typeck/src/expr_use_visitor.rs
@@ -619,6 +619,8 @@ impl<'a, 'tcx> ExprUseVisitor<'a, 'tcx> {
 
         if let Some(hir::Guard::If(ref e)) = arm.guard {
             self.consume_expr(e)
+        } else if let Some(hir::Guard::IfLet(_, ref e)) = arm.guard {
+            self.consume_expr(e)
         }
 
         self.consume_expr(&arm.body);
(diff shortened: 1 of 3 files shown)


compiler/rustc_typeck/src/expr_use_visitor.rs
```diff
@@ -619,6 +619,8 @@ impl<'a, 'tcx> ExprUseVisitor<'a, 'tcx> {
 
         if let Some(hir::Guard::If(ref e)) = arm.guard {
             self.consume_expr(e)
+        } else if let Some(hir::Guard::IfLet(_, ref e)) = arm.guard {
+            self.consume_expr(e)
         }
```
camsteffen, line 624: This can be an exhaustive match

nikomatsakis:
Nominating for beta backport, I believe this affects the Rust 2021 edition that is in beta.

PR #89283, opened 2021-09-26
Add regression test for issue #83564

camelid:
cc #83564

r? @davidtwco

### DIFF
--- a/src/test/ui/suggestions/core-std-import-order-issue-83564.stderr
+++ b/src/test/ui/suggestions/core-std-import-order-issue-83564.stderr
@@ -1,0 +1,16 @@
+error[E0433]: failed to resolve: use of undeclared type `NonZeroU32`
+  --> $DIR/core-std-import-order-issue-83564.rs:8:14
+   |
+LL |     let _x = NonZeroU32::new(5).unwrap();
+   |              ^^^^^^^^^^ not found in this scope
+   |
+help: consider importing one of these items
+   |
+LL | use std::num::NonZeroU32;
+   |
+LL | use core::num::NonZeroU32;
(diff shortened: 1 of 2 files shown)


src/test/ui/suggestions/core-std-import-order-issue-83564.stderr
```diff
@@ -0,0 +1,14 @@
...
+   |              ^^^^^^^^^^ not found in this scope
+   |
+help: consider importing this struct
+   |
+LL | use std::num::NonZeroU32;
+   |
+
+error: aborting due to previous error
+
+For more information about this error, try `rustc --explain E0433`.
```
camelid, line 14: Hmm, actually, this is the same error I get on stable, so this may not be testing what it should.

camelid: For some reason, Rust 2018 or higher is required to reproduce the original bug, so I added `// edition:2018`.

camelid:
Diff from nightly to master:

```diff
diff --git a/old.txt b/new.txt
index e22f6b1433f..b9fd6b98937 100644
--- a/old.txt
+++ b/new.txt
@@ -6,10 +6,10 @@ error[E0433]: failed to resolve: use of undeclared type `NonZeroU32`
   |
 help: consider importing one of these items
   |
-6 | use core::num::NonZeroU32;
-  |
 6 | use std::num::NonZeroU32;
   |
+6 | use core::num::NonZeroU32;
+  |
 
 error: aborting due to previous error
```

davidtwco (approved):
LGTM, thanks.

PR #89284, opened 2021-09-26
Make `Impl.trait_` a `Path`, not a `Type`

camelid:
Split out from #88379.

It should only ever be a `ResolvedPath`, so this (a) enforces that, and
(b) reduces the size of `Impl`.

I had to update a test because the order of the rendered auto trait impl
bounds changed. I think the order changed because rustdoc sorts auto
trait bounds using their `Debug` output.

r? @jyn514

### DIFF
--- a/src/librustdoc/json/conversions.rs
+++ b/src/librustdoc/json/conversions.rs
@@ -507,6 +507,11 @@ impl FromWithTcx<clean::Impl> for Impl {
             blanket_impl,
             span: _span,
         } = impl_;
+        // FIXME: should `trait_` be a Path in JSON?
+        let trait_ = trait_.map(|path| {
+            let did = path.res.def_id();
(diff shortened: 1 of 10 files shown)


src/librustdoc/json/conversions.rs
```diff
@@ -507,14 +507,19 @@ impl FromWithTcx<clean::Impl> for Impl {
             blanket_impl,
             span: _span,
         } = impl_;
+        // FIXME: should `trait_` be a Path in JSON?
+        let trait_ = trait_.map(|path| {
+            let did = path.res.def_id();
+            clean::ResolvedPath { path, did }.into_tcx(tcx)
+        });
```
camelid, line 514: I did this to avoid bumping the format-version, but it would be better to cleanup the JSON type too. Should I do that and bump the format-version?

camelid:
Closing in favor of #88379.

PR #89297, opened 2021-09-27
Remove Never variant from clean::Type enum

### LINKED ISSUE #89287: rustdoc: Remove `clean::Type::Never` in favor of always using `clean::PrimitiveType::Never`
It's trickier than it sounds; there's some weird rustdoc behavior with `Type::Never` that will need to be sorted out.

GuillaumeGomez:
Fixes #89287.

r? @camelid

### DIFF
--- a/src/librustdoc/clean/types.rs
+++ b/src/librustdoc/clean/types.rs
@@ -1396,7 +1396,6 @@ crate enum Type {
     Slice(Box<Type>),
     /// The `String` field is about the size or the constant representing the array's length.
     Array(Box<Type>, String),
-    Never,
     RawPointer(Mutability, Box<Type>),
     BorrowedRef {
         lifetime: Option<Lifetime>,
@@ -1462,7 +1461,6 @@ impl Type {
             }
             RawPointer(..) => Some(PrimitiveType::RawPointer),
             BareFunction(..) => Some(PrimitiveType::Fn),
-            Never => Some(PrimitiveType::Never),
             _ => None,
         }
     }
@@ -1550,7 +1548,6 @@ impl Type {
                 }
             }
             BareFunction(..) => PrimitiveType::Fn,
-            Never => PrimitiveType::Never,
             Slice(..) => PrimitiveType::Slice,
             Array(..) => PrimitiveType::Array,
             RawPointer(..) => PrimitiveType::RawPointer,
--- a/src/librustdoc/html/format.rs
+++ b/src/librustdoc/html/format.rs
@@ -761,6 +761,9 @@ fn fmt_type<'cx>(
             fmt::Display::fmt(&tybounds(bounds, lt, cx), f)
         }
         clean::Infer => write!(f, "_"),
+        clean::Primitive(clean::PrimitiveType::Never) => {
+            primitive_link(f, PrimitiveType::Never, "!", cx)
+        }
         clean::Primitive(prim) => primitive_link(f, prim, &*prim.as_sym().as_str(), cx),
         clean::BareFunction(ref decl) => {
             if f.alternate() {
@@ -819,7 +822,6 @@ fn fmt_type<'cx>(
                 primitive_link(f, PrimitiveType::Array, &format!("; {}]", Escape(n)), cx)
             }
         }
-        clean::Never => primitive_link(f, PrimitiveType::Never, "!", cx),
         clean::RawPointer(m, ref t) => {
             let m = match m {
                 hir::Mutability::Mut => "mut",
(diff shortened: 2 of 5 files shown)


src/librustdoc/clean/types.rs
```diff
@@ -1652,7 +1649,7 @@ impl PrimitiveType {
             sym::pointer => Some(PrimitiveType::RawPointer),
             sym::reference => Some(PrimitiveType::Reference),
             kw::Fn => Some(PrimitiveType::Fn),
-            sym::never => Some(PrimitiveType::Never),
+            sym::never | sym::never_type => Some(PrimitiveType::Never),
```
camelid, line 1652: `sym::never_type == "never_type"`, so this isn't right. I think you may need to add a `sym::bang == "!"` symbol.

src/librustdoc/html/format.rs
```diff
@@ -761,6 +761,9 @@ fn fmt_type<'cx>(
             fmt::Display::fmt(&tybounds(bounds, lt, cx), f)
         }
         clean::Infer => write!(f, "_"),
+        clean::Primitive(clean::PrimitiveType::Never) => {
+            primitive_link(f, PrimitiveType::Never, "!", cx)
+        }
```
camelid, line 766: Are there other places where there needs to be a special-case to replace `never` with `!`?

src/librustdoc/json/conversions.rs
```diff
@@ -417,13 +417,13 @@ impl FromWithTcx<clean::Type> for Type {
                 }
             }
             Generic(s) => Type::Generic(s.to_string()),
+            Primitive(clean::PrimitiveType::Never) => Type::Never,
```
camelid, line 420: Seems like the JSON format should be cleaned up to use `Type::Primitive("never")` instead of special-casing. Can you open an issue?

src/librustdoc/html/render/cache.rs
```diff
@@ -237,14 +237,14 @@ fn get_index_type_name(clean_type: &clean::Type, accept_generic: bool) -> Option
         }
         clean::DynTrait(ref bounds, _) => get_index_type_name(&bounds[0].trait_, accept_generic),
         clean::Generic(s) if accept_generic => Some(s),
+        clean::Primitive(clean::PrimitiveType::Never) => None,
```
camelid, line 240: I don't think this is right. `clean::Primitive(clean::PrimitiveType::Never)` _was_ used before this change; I think it was used for things like the `primitive.never.html` page. This change seems like it'll break searching for `never`. I think you can just remove this line and let the existing `clean::Primitive(ref p)` match arm handle it.

camelid (changes requested):

camelid:
(Why did you label this PR `A-rustdoc-json`? I removed it because this PR doesn't currently affect the JSON output.)

GuillaumeGomez:
> (Why did you label this PR `A-rustdoc-json`? I removed it because this PR doesn't currently affect the JSON output.)

I don't remember doing it, weird...

src/librustdoc/clean/types.rs
GuillaumeGomez: Good point!

src/librustdoc/html/format.rs
GuillaumeGomez: Normally they're all here.

src/librustdoc/json/conversions.rs
GuillaumeGomez: Opened https://github.com/rust-lang/rust/issues/89349.

src/librustdoc/html/render/cache.rs
GuillaumeGomez: I can confirm that it works perfectly fine as is.

src/librustdoc/clean/types.rs
GuillaumeGomez: Fixed.

src/librustdoc/clean/types.rs
```diff
@@ -1652,7 +1649,7 @@ impl PrimitiveType {
             sym::pointer => Some(PrimitiveType::RawPointer),
             sym::reference => Some(PrimitiveType::Reference),
             kw::Fn => Some(PrimitiveType::Fn),
-            sym::never => Some(PrimitiveType::Never),
+            sym::never | sym::bang => Some(PrimitiveType::Never),
```
camelid, line 1652: Unfortunately, I'm pretty sure this is still incorrect. `sym::bang == "bang"`; notice that it's in the `Symbols {` section of `rustc_span::symbol`.

GuillaumeGomez: I'll just remove this change as it's unneeded in this PR.

src/librustdoc/html/render/cache.rs
GuillaumeGomez: It was unneeded as well, removed this case too.

GuillaumeGomez:
Updated!

PR #89301, opened 2021-09-27
Add `is_absent` and `is_unicode_error` to VarError

### LINKED ISSUE #89302: Tracking Issue for `varerror_methods`
Feature gate: `#![feature(varerror_methods)]`

This is a tracking issue for the `varerror_methods` feature that adds two methods to `std::env::VarError`:
- `is_absent`: Which simply checks if the error resulted from a missing variable
- `is_unicode_error`: Which checks if the error resulted from unicode parsing failure

Generally, these methods make it more convenient to check if an environment variable was present, absent or failed to parse.

### Public API
```rust
use std::env;

fn main() {
    let var = env::var("bad_unicode");
    if var.is_unicode_error() {
        println!("Uh oh, your variable had incorrect unicode data");
    }
}
```

### Steps / History

- [x] Implementation (see #89301)
- [ ] Tests
- [ ] Stabilization PR

### Unresolved Questions

- None yet.

ohsayan:
Added the following functions to VarError:
- `is_absent`: Check if the error was caused by a missing env var
- `is_unicode_error`: Check if the error was caused by unicode parsing
failure

The addition of these methods generally simplifies some cases
where you'd just want to check if the variable is missing or if
an error has occurred while parsing it.

Signed-off-by: Sayan Nandan <nandansayan@outlook.com>

---
This closes #89302

### DIFF
--- a/library/std/src/env.rs
+++ b/library/std/src/env.rs
@@ -274,11 +274,13 @@ pub enum VarError {
 
 impl VarError {
     /// Checks if the error was caused by the environment variable being absent
+    #[unstable(feature = "varerror_methods", issue = "89302")]
     pub const fn is_absent(&self) -> bool {
         matches!(self, VarError::NotPresent)
     }
     /// Checks if the error was caused by the environment variable failing to be valid
     /// Unicode
+    #[unstable(feature = "varerror_methods", issue = "89302")]
     pub const fn is_unicode_error(&self) -> bool {
         matches!(self, VarError::NotUnicode(_))
     }
--- a/library/std/src/env.rs
+++ b/library/std/src/env.rs
@@ -275,13 +275,13 @@ pub enum VarError {
 impl VarError {
     /// Checks if the error was caused by the environment variable being absent
     #[unstable(feature = "varerror_methods", issue = "89302")]
-    pub const fn is_absent(&self) -> bool {
+    pub const fn is_not_present(&self) -> bool {
         matches!(self, VarError::NotPresent)
     }
     /// Checks if the error was caused by the environment variable failing to be valid
     /// Unicode
     #[unstable(feature = "varerror_methods", issue = "89302")]
-    pub const fn is_unicode_error(&self) -> bool {
+    pub const fn is_not_unicode(&self) -> bool {
         matches!(self, VarError::NotUnicode(_))
     }
 }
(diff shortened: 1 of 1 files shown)


ohsayan:
Opening the tracking issue now

library/std/src/env.rs
```diff
@@ -272,6 +272,20 @@ pub enum VarError {
     NotUnicode(#[stable(feature = "env", since = "1.0.0")] OsString),
 }
 
+impl VarError {
+    /// Checks if the error was caused by the environment variable being absent
+    #[unstable(feature = "varerror_methods", issue = "89302")]
+    pub const fn is_absent(&self) -> bool {
```
joshtriplett, line 278:
```suggestion
    pub const fn is_not_present(&self) -> bool {
```

library/std/src/env.rs
```diff
@@ -272,6 +272,20 @@ pub enum VarError {
...
+        matches!(self, VarError::NotPresent)
+    }
+    /// Checks if the error was caused by the environment variable failing to be valid
+    /// Unicode
+    #[unstable(feature = "varerror_methods", issue = "89302")]
+    pub const fn is_unicode_error(&self) -> bool {
```
joshtriplett, line 284:
```suggestion
    pub const fn is_not_unicode(&self) -> bool {
```

joshtriplett (changes requested):
I made a couple of naming suggestions for consistency with the names of the enum variants.

Separate from that, I'm going to nominate this issue for discussion, to address the general question of whether we want to offer APIs like this for enums that just trivially wrap `matches!`. I'm not sure if we need such methods or if we just want to suggest `matches!` or an `if let`.

yaahc:
> The addition of these methods generally simplifies some cases
> where you'd just want to check if the variable is missing or if
> an error has occurred while parsing it.

Can you give an example of this? My expectation is that the difference would be rather slight. The brevity of the method impls themselves seems to support this.

library/std/src/env.rs
ohsayan: Done

library/std/src/env.rs
ohsayan: Done

ohsayan:
> > The addition of these methods generally simplifies some cases
> > where you'd just want to check if the variable is missing or if
> > an error has occurred while parsing it.
> 
> Can you give an example of this? My expectation is that the difference would be rather slight. The brevity of the method impls themselves seems to support this.

I agree. It was just that I needed it in a place and I think it may be a far too niche addition.

yaahc:
> I agree. It was just that I needed it in a place and I think it may be a far too niche addition.

Sounds good, I'm going to go ahead and close this PR. Thank you again.

PR #89306, opened 2021-09-27
thread: implements available_concurrency on haiku

devnexen:


nagisa:
Seems reasonable to me.

@bors r+

nielx:
This change does not compile and is incorrect.

- mem::zeroed() is unsafe, see compiler error message below.
- `system_info()` returns a `status_t` indicating the error code or B_OK. It does not set `errno`, thus calling `io::Error::last_os_error()` will give a previous unrelated error code.

@devnexen are you okay to fix this?

{{{
error[E0133]: call to unsafe function is unsafe and requires unsafe function or block
   --> library/std/src/sys/unix/thread.rs:342:48
    |
342 |             let mut sinfo: libc::system_info = crate::mem::zeroed();
    |                                                ^^^^^^^^^^^^^^^^^^^^ call to unsafe function
    |
    = note: consult the function's documentation for information on how to avoid undefined behavior

error[E0133]: call to unsafe function is unsafe and requires unsafe function or block
   --> library/std/src/sys/unix/thread.rs:343:23
    |
343 |             let res = libc::get_system_info(&mut sinfo);
    |                       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ call to unsafe function
    |
    = note: consult the function's documentation for information on how to avoid undefined behavior
}}}

nagisa:
Oh well. I'm open to having this reverted until a fixed version is submitted, too.

devnexen:
I pushed a PR #89462

PR #89310, opened 2021-09-27
Make `std::thread::available_concurrency` support process-limited number of CPUs

joshtriplett:
Use `libc::sched_getaffinity` and count the number of CPUs in the returned mask. This handles cases where the process doesn't have access to all CPUs, such as when limited via `taskset` or similar.

This also covers cgroup cpusets.

### DIFF
--- a/library/std/src/sys/unix/thread.rs
+++ b/library/std/src/sys/unix/thread.rs
@@ -275,6 +275,14 @@ pub fn available_parallelism() -> io::Result<NonZeroUsize> {
             target_os = "solaris",
             target_os = "illumos",
         ))] {
+            #[cfg(any(target_os = "android", target_os = "linux"))]
+            {
+                let mut set: libc::cpu_set_t = unsafe { mem::zeroed() };
+                if unsafe { libc::sched_getaffinity(0, mem::size_of::<libc::cpu_set_t>(), &mut set) } == 0 {
+                    let count = unsafe { libc::CPU_COUNT(&set) };
+                    return Ok(unsafe { NonZeroUsize::new_unchecked(count as usize) });
+                }
+            }
             match unsafe { libc::sysconf(libc::_SC_NPROCESSORS_ONLN) } {
                 -1 => Err(io::Error::last_os_error()),
                 0 => Err(io::Error::new_const(io::ErrorKind::NotFound, &"The number of hardware threads is not known for the target platform")),
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1900,9 +1900,9 @@ checksum = "830d08ce1d1d941e6b30645f1a0eb5643013d835ce3779a5fc208261dbe10f55"
 
 [[package]]
 name = "libc"
-version = "0.2.103"
+version = "0.2.106"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "dd8f7255a17a627354f321ef0055d63b898c6fb27eff628af4d1b66b7331edf6"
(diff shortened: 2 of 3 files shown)


yoshuawuyts (approved):
Compared this against [`num_cpus/linux.rs`](https://github.com/seanmonstar/num_cpus/blob/fa15b13b72b4079401bfa44e849e478d147cd230/src/linux.rs#L39-L57), and this looks good!

library/std/src/sys/unix/thread.rs
```diff
@@ -265,12 +265,21 @@ impl Drop for Thread {
...
-            target_os = "android",
+        if #[cfg(any(target_os = "android", target_os = "linux"))] {
+            let mut set: libc::cpu_set_t = unsafe { mem::zeroed() };
+            if unsafe { libc::sched_getaffinity(0, mem::size_of::<libc::cpu_set_t>(), &mut set) } == 0 {
+                let count = unsafe { libc::CPU_COUNT(&set) };
+                return Ok(unsafe { NonZeroUsize::new_unchecked(count as usize) });
+            }
+            match unsafe { libc::sysconf(libc::_SC_NPROCESSORS_ONLN) } {
+                -1 => Err(io::Error::last_os_error()),
+                0 => Err(io::Error::new_const(io::ErrorKind::NotFound, &"The number of hardware threads is not known for the target platform")),
```
yoshuawuyts, line 276: I like this error message!

library/std/src/sys/unix/thread.rs
```diff
@@ -265,12 +265,21 @@ impl Drop for Thread {
 
 pub fn available_concurrency() -> io::Result<NonZeroUsize> {
     cfg_if::cfg_if! {
-        if #[cfg(any(
```
m-ou-se, line 268: Instead of duplicating the `match unsafe { libc::sysconf(libc::_SC_NPROCESSORS_ONLN) }` part, it might be clearer to conditionally add the new code to the existing branch, keeping a single sysconf(_SC_NPROCESSORS_ONLN) call:

```rust
    cfg_if::cfg_if! {
        if #[cfg(any(
            target_os = "android",
            target_os = "emscripten",
            target_os = "fuchsia",
            target_os = "ios",
            target_os = "linux",
            target_os = "macos",
            target_os = "solaris",
            target_os = "illumos",
        ))] {
            #[cfg(any(target_os = "android", target_os = "linux"))]
            {
                let mut set: libc::cpu_set_t = unsafe { mem::zeroed() };
                if unsafe { libc::sched_getaffinity(0, mem::size_of::<libc::cpu_set_t>(), &mut set) } == 0 {
                    let count = unsafe { libc::CPU_COUNT(&set) };
                    return Ok(unsafe { NonZeroUsize::new_unchecked(count as usize) });
                }
            }
            match unsafe { libc::sysconf(libc::_SC_NPROCESSORS_ONLN) } {
                -1 => Err(io::Error::last_os_error()),
                0 => Err(io::Error::new_const(io::ErrorKind::NotFound, &"The number of hardware threads is not known for the target platform")),
                cpus => Ok(unsafe { NonZeroUsize::new_unchecked(cpus as usize) }),
            }
    } ...
```

joshtriplett: Done. I think the use of `cfg_if` had made me momentarily forget that std (unlike `libc`) doesn't *require* the use of `cfg_if`.

joshtriplett:
Updated  to libc 0.2.106 which includes the fix for the issue that broke CI on macOS.

@bors r=m-ou-se

joshtriplett:
Let's try again now that https://github.com/rust-lang/rust/pull/90495 is fixed:

@bors r=m-ou-se

PR #89313, opened 2021-09-27
PassWrapper: handle function rename from upstream D36850

durin42:
thinLTOResolvePrevailingInModule became thinLTOFinalizeInModule and
gained the ability to propagate noRecurse and noUnwind function
attributes. I ran codegen tests with it both on and off, as the upstream
patch uses it in both modes, and the tests pass both ways. Given that,
it seemed reasonable to go ahead and let the propagation be enabled in
rustc, and see what happens. See https://reviews.llvm.org/D36850 for
more examples of how the new version of the function gets used.

r? @nikic cc @nagisa


nikic:
@bors r+ rollup

We should remember this when doing the LLVM 14 upgrade, in case there are unexpected compile-time regressions.

PR #89314, opened 2021-09-28
fix(lint): don't suggest refutable patterns to "fix" irrefutable bind

### LINKED ISSUE #88730: bindings_with_variant_name is unhelpful and incorrect for function arguments
Given the following code:

```rust
#![allow(non_camel_case_types, unused_variables, dead_code)]

enum Foo {
    c
}

fn foo(c: Foo) {}
```

https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=e6829290d29151853fdd98844ba6391c

The current output is:

```
warning[E0170]: pattern binding `c` is named the same as one of the variants of the type `Foo`
 --> src/lib.rs:7:8
  |
7 | fn foo(c: Foo) {}
  |        ^ help: to match on the variant, qualify the path: `Foo::c`
  |
  = note: `#[warn(bindings_with_variant_name)]` on by default
```

Ideally the lint should not trigger on such a pattern, it does not really make sense and the suggestion is wrong altogether,

notriddle:
In function arguments and let bindings, do not suggest changing `C` to `Foo::C` unless `C` is the only variant of `Foo`, because it won't work.

The general warning is still kept, because code like this is confusing.

Fixes #88730

p.s. `src/test/ui/lint/lint-uppercase-variables.rs` already tests the one-variant case.

### DIFF
--- a/compiler/rustc_mir_build/src/thir/pattern/check_match.rs
+++ b/compiler/rustc_mir_build/src/thir/pattern/check_match.rs
@@ -39,6 +39,13 @@ fn create_e0004(sess: &Session, sp: Span, error_message: String) -> DiagnosticBu
     struct_span_err!(sess, sp, E0004, "{}", &error_message)
 }
 
+#[derive(PartialEq)]
+enum RefutableFlag {
+    Irrefutable,
+    Refutable,
+}
+use RefutableFlag::*;
+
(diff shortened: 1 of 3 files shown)


davidtwco (approved):
LGTM!

notriddle:
@davidtwco I had to rebase this, because some pattern matching machinery changed.

PR #89316, opened 2021-09-28
Add support for specifying multiple clobber_abi in `asm!`

asquared31415:
r? @Amanieu 
cc #72016 
@rustbot label: +A-inline-assembly +F-asm

### DIFF
--- a/compiler/rustc_ast/src/ast.rs
+++ b/compiler/rustc_ast/src/ast.rs
@@ -2058,7 +2058,7 @@ pub struct InlineAsm {
     pub template: Vec<InlineAsmTemplatePiece>,
     pub template_strs: Box<[(Symbol, Option<Symbol>, Span)]>,
     pub operands: Vec<(InlineAsmOperand, Span)>,
-    pub clobber_abi: Option<(Symbol, Span)>,
+    pub clobber_abis: Vec<(Symbol, Span)>,
     pub options: InlineAsmOptions,
     pub line_spans: Vec<Span>,
 }
@@ -2715,7 +2715,7 @@ pub enum ItemKind {
     /// E.g., `extern {}` or `extern "C" {}`.
     ForeignMod(ForeignMod),
     /// Module-level inline assembly (from `global_asm!()`).
-    GlobalAsm(InlineAsm),
+    GlobalAsm(Box<InlineAsm>),
     /// A type alias (`type`).
     ///
     /// E.g., `type Foo = Bar<u8>;`.
--- a/compiler/rustc_ast_lowering/src/asm.rs
+++ b/compiler/rustc_ast_lowering/src/asm.rs
@@ -2,6 +2,7 @@ use super::LoweringContext;
 
 use rustc_ast::*;
 use rustc_data_structures::fx::FxHashMap;
+use rustc_data_structures::stable_set::FxHashSet;
 use rustc_errors::struct_span_err;
 use rustc_hir as hir;
 use rustc_session::parse::feature_err;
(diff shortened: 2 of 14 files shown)


asquared31415:
One concern I have is the increase of the size of ast::ItemKind, if that is a problem I think that it can be fixed with some boxed slices?

compiler/rustc_ast/src/ast.rs
```diff
@@ -2744,7 +2744,7 @@ pub enum ItemKind {
 }
 
 #[cfg(all(target_arch = "x86_64", target_pointer_width = "64"))]
-rustc_data_structures::static_assert_size!(ItemKind, 112);
```
Amanieu, line 2747: Consider boxing the `InlineAsm` to avoid increasing the size of `ItemKind`.

compiler/rustc_ast_lowering/src/asm.rs
```diff
@@ -27,11 +28,13 @@ impl<'a, 'hir> LoweringContext<'a, 'hir> {
                 .emit();
         }
 
-        let mut clobber_abi = None;
+        let mut clobber_abis = FxHashSet::default();
```
Amanieu, line 31: Use a `HashMap` keyed on the ABI, otherwise 2 identical ABIs with different spans will both be included.

compiler/rustc_ast_lowering/src/asm.rs
```diff
@@ -368,7 +371,7 @@ impl<'a, 'hir> LoweringContext<'a, 'hir> {
 
         // If a clobber_abi is specified, add the necessary clobbers to the
         // operands list.
-        if let Some((abi, abi_span)) = clobber_abi {
+        for (abi, abi_span) in clobber_abis {
```
Amanieu, line 374: Make sure you don't emit duplicate `Out` operands for the same register from multiple clobbers.

compiler/rustc_builtin_macros/src/asm.rs
```diff
@@ -210,7 +210,7 @@ fn parse_args<'a>(
                 .span_labels(args.options_spans.clone(), "previous options")
                 .span_label(span, "argument")
                 .emit();
-        } else if let Some((_, abi_span)) = args.clobber_abi {
+        } else if let Some(&(_, abi_span)) = args.clobber_abis.last() {
             ecx.struct_span_err(span, "arguments are not allowed after clobber_abi")
                 .span_label(abi_span, "clobber_abi")
```
Amanieu, line 215: I'd like this to point to all the `clobber_abi` spans, not just the last one.

compiler/rustc_builtin_macros/src/asm.rs
```diff
@@ -322,7 +322,7 @@ fn parse_args<'a>(
         // Bail out now since this is likely to confuse MIR
         return Err(err);
     }
-    if let Some((_, abi_span)) = args.clobber_abi {
+    if let Some(&(_, abi_span)) = args.clobber_abis.last() {
         if is_global_asm {
             let err =
                 ecx.struct_span_err(abi_span, "`clobber_abi` cannot be used with `global_asm!`");
```
Amanieu, line 328: Same here.

Amanieu:
Can you add this to grammar in the asm documentation in the unstable book? Also in the reference-level explanation add a note saying that `clobber_abi` can be specified multiple times.

fee1-dead:
@bors r-

Failed in https://github.com/rust-lang/rust/pull/89420#issuecomment-932171775

nbdd0121 (commented):
I wonder if the syntax should be
```rust
clobber_abi("A"), clobber_abi("B")
```
or
```rust
clobber_abi("A", "B")
```

matthiaskrgr:
the queue showed this as still approved while Ci failed already, strange.
Just in case: bors r-

matthiaskrgr:
@bors r- 
failed in a rollup: https://github.com/rust-lang/rust/pull/89959#issuecomment-945011434

matthiaskrgr:
Mmh, this had some test failure in an experimental rollup  https://github.com/rust-lang/rust/pull/90740#issuecomment-964616507
but in theory it should have triggered here as well.. :thinking:

Amanieu:
`sym` was moved to a different feature since this PR was last updated.

PR #89317, opened 2021-09-28
Move generic error message to separate branches

JulianKnodt:
This decomposes an error message in generic constants into more specific branches, for better
readability.

r? @lcnr

### DIFF
--- a/compiler/rustc_trait_selection/src/traits/const_evaluatable.rs
+++ b/compiler/rustc_trait_selection/src/traits/const_evaluatable.rs
@@ -236,12 +236,23 @@ impl<'a, 'tcx> AbstractConstBuilder<'a, 'tcx> {
         self.body.exprs[self.body_id].span
     }
 
-    fn error(&mut self, span: Option<Span>, msg: &str) -> Result<!, ErrorReported> {
+    fn error(&mut self, span: Span, msg: &str) -> Result<!, ErrorReported> {
         self.tcx
             .sess
             .struct_span_err(self.root_span(), "overly complex generic constant")
-            .span_label(span.unwrap_or(self.root_span()), msg)
+            .span_label(span, msg)
             .help("consider moving this anonymous constant into a `const` function")
+            .emit();
+
+        Err(ErrorReported)
+    }
+    fn maybe_supported_error(&mut self, span: Span, msg: &str) -> Result<!, ErrorReported> {
+        self.tcx
+            .sess
+            .struct_span_err(self.root_span(), "overly complex generic constant")
+            .span_label(span, msg)
+            .help("consider moving this anonymous constant into a `const` function")
+            .note("this operation may be supported in the future")
             .emit();
 
         Err(ErrorReported)
@@ -337,14 +348,14 @@ impl<'a, 'tcx> AbstractConstBuilder<'a, 'tcx> {
         Ok(match &node.kind {
             // I dont know if handling of these 3 is correct
             &ExprKind::Scope { value, .. } => self.recurse_build(value)?,
-            &ExprKind::PlaceTypeAscription { source, .. } |
-            &ExprKind::ValueTypeAscription { source, .. } => self.recurse_build(source)?,
+            &ExprKind::PlaceTypeAscription { source, .. }
+            | &ExprKind::ValueTypeAscription { source, .. } => self.recurse_build(source)?,
 
             // subtle: associated consts are literals this arm handles
             // `<T as Trait>::ASSOC` as well as `12`
             &ExprKind::Literal { literal, .. } => self.nodes.push(Node::Leaf(literal)),
 
-            ExprKind::Call { fun, args,  .. } => {
+            ExprKind::Call { fun, args, .. } => {
                 let fun = self.recurse_build(*fun)?;
 
                 let mut new_args = Vec::<NodeId>::with_capacity(args.len());
@@ -353,7 +364,7 @@ impl<'a, 'tcx> AbstractConstBuilder<'a, 'tcx> {
                 }
                 let new_args = self.tcx.arena.alloc_slice(&new_args);
                 self.nodes.push(Node::FunctionCall(fun, new_args))
-            },
+            }
             &ExprKind::Binary { op, lhs, rhs } if Self::check_binop(op) => {
                 let lhs = self.recurse_build(lhs)?;
                 let rhs = self.recurse_build(rhs)?;
@@ -362,7 +373,7 @@ impl<'a, 'tcx> AbstractConstBuilder<'a, 'tcx> {
             &ExprKind::Unary { op, arg } if Self::check_unop(op) => {
                 let arg = self.recurse_build(arg)?;
                 self.nodes.push(Node::UnaryOp(op, arg))
-            },
+            }
             // This is necessary so that the following compiles:
             //
             // ```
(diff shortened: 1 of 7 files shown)


compiler/rustc_trait_selection/src/traits/const_evaluatable.rs
```diff
@@ -396,34 +398,53 @@ impl<'a, 'tcx> AbstractConstBuilder<'a, 'tcx> {
...
+                self.error(Some(node.span), "unsupported operation in generic constants")?
+            }
+            ExprKind::Assign { .. } | ExprKind::AssignOp { .. } => {
+                self.error(Some(node.span), "assignment is not supported in generic constants")?
+            }
+            ExprKind::Closure { .. } | ExprKind::Return { .. } => self.error(
+                Some(node.span),
+                "closures and function keywords are not supported in generic constants",
+            )?,
+            ExprKind::Match { .. } | ExprKind::If { .. } | ExprKind::Let { .. } =>
```
BoxyUwU, line 431: I think we should handle `Yield` `Return` `Continue` `Break` `Loop` in this arm as the reason they are all disallowed is because we dont allow controlflow

compiler/rustc_trait_selection/src/traits/const_evaluatable.rs
```diff
@@ -396,34 +398,53 @@ impl<'a, 'tcx> AbstractConstBuilder<'a, 'tcx> {
...
+                self.error(Some(node.span), "pointers are not allowed in generic constants")?
+            }
+            ExprKind::Continue { .. } | ExprKind::Break { .. } | ExprKind::Loop { .. } => self
+                .error(
                     Some(node.span),
-                    "unsupported operation in generic constant, this may be supported in the future",
+                    "loops and loop control flow are not supported in generic constants",
                 )?,
+            ExprKind::Box { .. } =>
+            // allocations not allowed in constants
```
BoxyUwU, line 415: we can probably remove this comment now that there's an error message right next to it saying the same thing

compiler/rustc_trait_selection/src/traits/const_evaluatable.rs
```diff
@@ -396,34 +398,53 @@ impl<'a, 'tcx> AbstractConstBuilder<'a, 'tcx> {
             | ExprKind::Field { .. }
             | ExprKind::ConstBlock { .. }
             | ExprKind::Adt(_) => self.error(
+                Some(node.span),
+                "unsupported operation in generic constant, this may be supported in the future",
+            )?,
+            ExprKind::Pointer { .. } =>
+            // dont know if this is correct
+            {
+                self.error(Some(node.span), "pointers are not allowed in generic constants")?
```
BoxyUwU, line 407: I think "pointer casts are not allowed in generic constants" is more what this is actually forbidding

compiler/rustc_trait_selection/src/traits/const_evaluatable.rs
```diff
@@ -396,34 +398,53 @@ impl<'a, 'tcx> AbstractConstBuilder<'a, 'tcx> {
...
                 )?,
+            ExprKind::Box { .. } =>
+            // allocations not allowed in constants
+            {
+                self.error(Some(node.span), "allocations are not allowed in generic constants")?
+            }
+
+            // we handle valid unary/binary ops above
+            ExprKind::Unary { .. } | ExprKind::Binary { .. } | ExprKind::LogicalOp { .. } => {
+                self.error(Some(node.span), "unsupported operation in generic constants")?
```
BoxyUwU, line 422: The reason we disallow `ExprKind::LogicalOp` is because `&&` and `||` are short circuiting and so imply controlflow. I think we should mention this. I'm not sure why we disallow `Binop::Offset` ``/// The `ptr.offset` operator`` (or what it is)

compiler/rustc_trait_selection/src/traits/const_evaluatable.rs
```diff
@@ -396,34 +398,53 @@ impl<'a, 'tcx> AbstractConstBuilder<'a, 'tcx> {
             | ExprKind::Field { .. }
             | ExprKind::ConstBlock { .. }
             | ExprKind::Adt(_) => self.error(
+                Some(node.span),
+                "unsupported operation in generic constant, this may be supported in the future",
```
BoxyUwU, line 402: I think we should split this arm up a bit to give more specific error messages. I can imagine someone might be confused as to what the unsupported thing is when doing `foo.bar` or building a tuple or making an adt etc

BoxyUwU (changes requested):

compiler/rustc_trait_selection/src/traits/const_evaluatable.rs
JulianKnodt: I've added a msg for LogicalOps but nothing else

compiler/rustc_trait_selection/src/traits/const_evaluatable.rs
```diff
@@ -362,68 +373,110 @@ impl<'a, 'tcx> AbstractConstBuilder<'a, 'tcx> {
...
+                "adts are not supported in generic constant",
+            )?,
+            // dont know if this is correct
+            ExprKind::Pointer { .. } =>
+                self.error(Some(node.span), "pointer casts are not allowed in generic constants")?,
+            ExprKind::Yield { .. } =>
+                self.error(Some(node.span), "generator control flow is not allowed in generic constants")?,
+            ExprKind::Continue { .. } | ExprKind::Break { .. } | ExprKind::Loop { .. } => self
+                .error(
                     Some(node.span),
```
lcnr, line 445: as we now have `thir`, it is probably good to change error to always require a span.

lcnr (approved):
am pretty happy with this

if you want you it would be great to have tests for all the new added arms, but I don't think that's too important.

r=me, but as @BoxyUwU was already reviewing this pr

r? @BoxyUwU

compiler/rustc_trait_selection/src/traits/const_evaluatable.rs
JulianKnodt: will update in a bit

compiler/rustc_trait_selection/src/traits/const_evaluatable.rs
```diff
@@ -362,68 +373,108 @@ impl<'a, 'tcx> AbstractConstBuilder<'a, 'tcx> {
...
+                node.span,
+                "field access is not supported in generic constant",
+            )?,
+            ExprKind::ConstBlock { .. } => self.maybe_supported_error(
+                node.span,
+                "const blocks are not supported in generic constant",
+            )?,
+            ExprKind::Adt(_) => self.maybe_supported_error(
+                node.span,
+                "adts are not supported in generic constant",
```
BoxyUwU, line 436:
```suggestion
                "struct/enum construction is not supported in generic constants",
```

compiler/rustc_trait_selection/src/traits/const_evaluatable.rs
```diff
@@ -362,68 +373,108 @@ impl<'a, 'tcx> AbstractConstBuilder<'a, 'tcx> {
...
+                node.span,
+                "blocks are not supported in generic constant",
+            )?,
+            ExprKind::NeverToAny { .. } => self.maybe_supported_error(
+                node.span,
+                "converting nevers to any is not supported in generic constant",
+            )?,
+            ExprKind::Tuple { .. } => self.maybe_supported_error(
+                node.span,
+                "tuples are not supported in generic constant",
```
BoxyUwU, line 420:
```suggestion
                "tuple construction is not supported in generic constants",
```

compiler/rustc_trait_selection/src/traits/const_evaluatable.rs
```diff
@@ -362,68 +373,108 @@ impl<'a, 'tcx> AbstractConstBuilder<'a, 'tcx> {
...
-            | ExprKind::Adt(_) => self.error(
-                    Some(node.span),
-                    "unsupported operation in generic constant, this may be supported in the future",
+            | ExprKind::Deref { .. } => self.maybe_supported_error(
+                node.span,
+                "references are not supported in generic constant",
+            )?,
+            ExprKind::Repeat { .. } | ExprKind::Array { .. } =>  self.maybe_supported_error(
+                node.span,
+                "arrays are not supported in generic constant",
```
BoxyUwU, line 408:
```suggestion
                "array construction is not supported in generic constants",
```

compiler/rustc_trait_selection/src/traits/const_evaluatable.rs
```diff
@@ -362,68 +373,108 @@ impl<'a, 'tcx> AbstractConstBuilder<'a, 'tcx> {
...
-            | ExprKind::Tuple { .. }
-            | ExprKind::Index { .. }
-            | ExprKind::Field { .. }
-            | ExprKind::ConstBlock { .. }
-            | ExprKind::Adt(_) => self.error(
-                    Some(node.span),
-                    "unsupported operation in generic constant, this may be supported in the future",
+            | ExprKind::Deref { .. } => self.maybe_supported_error(
+                node.span,
+                "references are not supported in generic constant",
```
BoxyUwU, line 404:
```suggestion
                "dereferencing is not supported in generic constants",
```

BoxyUwU (changes requested):

PR #88748, opened 2021-09-08
Revert "Remove optimization_fuel_crate from Session"

bjorn3:
This reverts commit 5464b2e713d5366b3aec5c6eebbe1b84a782c51e.

This hopefully fixes the perf regression in https://github.com/rust-lang/rust/pull/88530#issuecomment-915314117.

PR #89322, opened 2021-09-28
Reapply "Remove optimization_fuel_crate from Session"

tmiasko:


tmiasko:
This change was [identified as performance regression](https://github.com/rust-lang/rust/pull/88530#issuecomment-915314117) and [reverted](https://github.com/rust-lang/rust/pull/88748). It turns out that the regression was in an unrelated area of the compiler and fixed after splitting the mir crate, before revert even landed.

cc @bjorn3

michaelwoerister (approved):

michaelwoerister:
Thanks, @tmiasko! 
@bors r+ rollup

PR #89323, opened 2021-09-28
Consider unfulfilled obligations in binop errors

### LINKED ISSUE #84515: Suggest implementing `PartialEq` for `Foo` when comparing `Option<Foo>` with `Option<Foo>`
Given the following code: [Link to Playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=45fe113003a2e744c8b2bde3edbbde8a)

```rust
struct Foo;

fn main() {
    let _ = Some(Foo) == Some(Foo);
}
```

The current output is:

```
error[E0369]: binary operation `==` cannot be applied to type `Option<Foo>`
 --> src/main.rs:4:23
  |
4 |     let _ = Some(Foo) == Some(Foo);
  |             --------- ^^ --------- Option<Foo>
  |             |
  |             Option<Foo>
```

```
error[E0369]: binary operation `==` cannot be applied to type `Option<Foo>`
 --> src/main.rs:4:23
  |
4 |     let _ = Some(Foo) == Some(Foo);
  |             --------- ^^ --------- Option<Foo>
  |             |
  |             Option<Foo>
  |
  = note: an implementation of `std::cmp::PartialEq<Option<Foo>>` is found for `Option<Foo>`
          but the trait bounds were not satisfied:
          `Foo: PartialEq<Foo>`
  = note: an implementation of `std::cmp::PartialEq` might be missing for `Foo`
```

estebank:
When encountering a binop where the types would have been accepted, if
all the predicates had been fulfilled, include information about the
predicates and suggest appropriate `#[derive]`s if possible.

Fix #84515.

### DIFF
--- a/compiler/rustc_typeck/src/check/method/mod.rs
+++ b/compiler/rustc_typeck/src/check/method/mod.rs
@@ -290,29 +290,14 @@ impl<'a, 'tcx> FnCtxt<'a, 'tcx> {
         )
     }
 
-    /// `lookup_method_in_trait` is used for overloaded operators.
-    /// It does a very narrow slice of what the normal probe/confirm path does.
-    /// In particular, it doesn't really do any probing: it simply constructs
-    /// an obligation for a particular trait with the given self type and checks
-    /// whether that trait is implemented.
-    //
-    // FIXME(#18741): it seems likely that we can consolidate some of this
-    // code with the other method-lookup code. In particular, the second half
-    // of this method is basically the same as confirmation.
-    #[instrument(level = "debug", skip(self, span, opt_input_types))]
-    pub fn lookup_method_in_trait(
+    pub(super) fn obligation_for_method(
         &self,
         span: Span,
-        m_name: Ident,
         trait_def_id: DefId,
         self_ty: Ty<'tcx>,
         opt_input_types: Option<&[Ty<'tcx>]>,
-    ) -> Option<InferOk<'tcx, MethodCallee<'tcx>>> {
-        debug!(
-            "lookup_in_trait_adjusted(self_ty={:?}, m_name={}, trait_def_id={:?}, opt_input_types={:?})",
-            self_ty, m_name, trait_def_id, opt_input_types
-        );
-
+    ) -> (traits::Obligation<'tcx, ty::Predicate<'tcx>>, &'tcx ty::List<ty::subst::GenericArg<'tcx>>)
+    {
         // Construct a trait-reference `self_ty : Trait<input_tys>`
         let substs = InternalSubsts::for_item(self.tcx, trait_def_id, |param, _| {
             match param.kind {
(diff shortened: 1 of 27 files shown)


compiler/rustc_typeck/src/check/method/mod.rs
```diff
@@ -342,7 +344,11 @@ impl<'a, 'tcx> FnCtxt<'a, 'tcx> {
         // Now we want to know if this can be matched
         if !self.predicate_may_hold(&obligation) {
             debug!("--> Cannot match obligation");
-            return None; // Cannot be matched, no such method resolution is possible.
+            let mut fulfill = <dyn traits::TraitEngine<'_>>::new(self.tcx);
+            fulfill.register_predicate_obligation(self, obligation);
+            // Cannot be matched, no such method resolution is possible, but we collect the
+            // unfulfilled obligations for better error reporting.
+            return Err(fulfill.select_where_possible(&self.infcx).unwrap_err());
```
petrochenkov, line 351: `lookup_method_in_trait` can be tried on many traits, so this code is run on a good path and may affect performance even if no errors are reported.

petrochenkov: Is it possible to return `Err(obligation)` instead, and perform these actions only if the error is actually emitted?

estebank:
@petrochenkov, sorry I hadn't realized we changed the `waiting-on` labels back and forth. Why should this be waiting on author?

petrochenkov:
I set the label after writing https://github.com/rust-lang/rust/pull/89323#discussion_r720278389 and before the last commit appeared, now it's waiting on perf, I guess.

compiler/rustc_typeck/src/check/method/mod.rs
```diff
@@ -328,21 +313,52 @@ impl<'a, 'tcx> FnCtxt<'a, 'tcx> {
             self.var_for_def(span, param)
         });
 
-        let trait_ref = ty::TraitRef::new(trait_def_id, substs);
+        let trait_ref = ty::TraitRef::new(trait_def_id, substs.clone());
```
petrochenkov, line 316:
```suggestion
        let trait_ref = ty::TraitRef::new(trait_def_id, substs);
```
`substs` is a reference, no?

compiler/rustc_typeck/src/check/method/mod.rs
```diff
@@ -290,29 +290,14 @@ impl<'a, 'tcx> FnCtxt<'a, 'tcx> {
...
-    /// In particular, it doesn't really do any probing: it simply constructs
-    /// an obligation for a particular trait with the given self type and checks
-    /// whether that trait is implemented.
-    //
-    // FIXME(#18741): it seems likely that we can consolidate some of this
-    // code with the other method-lookup code. In particular, the second half
-    // of this method is basically the same as confirmation.
-    #[instrument(level = "debug", skip(self, span, opt_input_types))]
-    pub fn lookup_method_in_trait(
+    pub fn obligation_for_method(
```
petrochenkov, line 293:
```suggestion
    pub(super) fn obligation_for_method(
```
(And `fn lookup_method_in_trait` too.)

src/test/ui/issues/issue-31076.stderr
```diff
@@ -13,6 +15,8 @@ LL |     let y = 5i32 + 6i32;
    |             ---- ^ ---- i32
    |             |
    |             i32
+   |
+   = note: the following types would have to `impl` their required traits for this operation to be valid
```
petrochenkov, line 19: Which following types?

PR #74480, opened 2020-07-18
Add `std::thread::available_concurrency`

yoshuawuyts:
This PR adds a counterpart to [C++'s `std::thread::hardware_concurrency`](https://en.cppreference.com/w/cpp/thread/thread/hardware_concurrency) to Rust, tracking issue https://github.com/rust-lang/rust/issues/74479.

cc/ @rust-lang/libs 

## Motivation

Being able to know how many hardware threads a platform supports is a core part of building multi-threaded code. In C++ 11 this has become available through the [`std::thread::hardware_concurrency`](https://en.cppreference.com/w/cpp/thread/thread/hardware_concurrency) API. Currently in Rust most of the ecosystem depends on the [`num_cpus` crate](https://docs.rs/num_cpus/1.13.0/num_cpus/) ([no.35 in top 500 crates](https://docs.google.com/spreadsheets/d/1wwahRMHG3buvnfHjmPQFU4Kyfq15oTwbfsuZpwHUKc4/edit#gid=1253069234)) to provide this functionality. This PR proposes an API to provide access to the number of hardware threads available on a given platform.

__edit (2020-07-24):__ The purpose of this PR is to provide a hint for how many threads to spawn to saturate the processor. There's value in introducing APIs for NUMA and Windows processor groups, but those are intentionally out of scope for this PR. See: https://github.com/rust-lang/rust/pull/74480#issuecomment-662116186.

## Naming

Discussing the naming of the API on Zulip surfaced two options:

- `std::thread::hardware_concurrency`
- `std::thread::hardware_threads`

Both options seemed acceptable, but overall people seem to gravitate the most towards `hardware_threads`. Additionally @jonas-schievink pointed out that the "hardware threads" terminology is well-established and is used in among other the [RISC-V specification](https://riscv.org/specifications/isa-spec-pdf/) (page 20):

> A component is termed a core if it contains an independent instruction fetch unit. A RISC-V-compatible core might support multiple RISC-V-compatible __hardware threads__, or harts, through multithreading.

It's also worth noting that [the original paper introducing C++'s `std::thread` submodule](http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2007/n2320.html) unfortunately doesn't feature any discussion on the naming of `hardware_concurrency`, so we can't use that to help inform our decision here.

## Return type

[...]

Issue #74479, opened 2020-07-18
Tracking Issue for `#![feature(available_parallelism)]`


yoshuawuyts:
The feature gate for the issue is `#![feature(available_parallelism)]`.

This is a tracking issue for `std::thread::available_parallelism`; a portable API to determine how many threads to spawn in order to ensure a program can make use of all available parallelism available on a machine.

[stabilization-guide]: https://rustc-dev-guide.rust-lang.org/stabilization_guide.html#stabilization-pr
[doc-guide]: https://rustc-dev-guide.rust-lang.org/stabilization_guide.html#documentation-prs

### Public API

```rust
#![feature(available_parallelism)]
use std::thread;

fn main() -> std::io::Result<()> {
    let count = thread::available_parallelism()?.get();
    assert!(count >= 1);
    Ok(())
}
```

### Tasks

- [x] Resolve discussion on naming.
- [x] Resolve discussion on function signature.
- [x] Resolve discussion on terminology.
- [x] Add support for externally-set limits on Linux ([ref](https://github.com/rust-lang/rust/pull/89310)).
- [ ] Add support for externally-set limits on Windows ([ref](https://github.com/rust-lang/rust/pull/74480#issuecomment-673311001)).
- [x] Update documentation to remove mentions of "hardware" ([ref](https://github.com/rust-lang/rust/pull/89324#issuecomment-934343254)).
- [x] Smooth out the docs example.

### Notes on Platform-specific behavior

`available_parallelism` is (for now) only considered a hint. The following platform limitations exist:

- On Windows the available concurrency may be undercounted when exceeding 64 threads https://github.com/rust-lang/rust/pull/74480#issuecomment-660557465
- On Windows the available concurrency may be overcounted when limited by a process wide affinity mask or job object limitations https://github.com/rust-lang/rust/pull/74480#issuecomment-673311001
- On Linux the available concurrency may be overcounted in the face of cgroup limits and task-wise affinity masks https://github.com/rust-lang/rust/pull/74480#issuecomment-693390069

### Implementation history

- https://github.com/rust-lang/rust/pull/74480
- https://github.com/rust-lang/rust/pull/85182
- https://github.com/rust-lang/rust/pull/89310
- https://github.com/rust-lang/rust/pull/89324

yoshuawuyts:
[...]

PR #89324, opened 2021-09-28
Rename `std::thread::available_conccurrency` to `std::thread::available_parallelism`

yoshuawuyts:
_Tracking issue: https://github.com/rust-lang/rust/issues/74479_

This PR renames  `std::thread::available_conccurrency` to `std::thread::available_parallelism`.

## Rationale

The API was initially named `std::thread::hardware_concurrency`, mirroring the [C++ API of the same name](https://en.cppreference.com/w/cpp/thread/thread/hardware_concurrency). We eventually decided to omit any reference to the word "hardware" after [this comment](https://github.com/rust-lang/rust/pull/74480#issuecomment-662045841). And so we ended up with `available_concurrency` instead.

---

For a talk I was preparing this week I was reading through ["Understanding and expressing scalable concurrency" (A. Turon, 2013)](http://aturon.github.io/academic/turon-thesis.pdf), and the following passage stood out to me (emphasis mine):

> __Concurrency is a system-structuring mechanism.__ An interactive system that deals with disparate asynchronous events is naturally structured by division into concurrent threads with disparate responsibilities. Doing so creates a better fit between problem and solution, and can also decrease the average latency of the system by preventing long-running computations from obstructing quicker ones.

> __Parallelism is a resource.__ A given machine provides a certain capacity for parallelism, i.e., a bound on the number of computations it can perform simultaneously. The goal is to maximize throughput by intelligently using this resource. For interactive systems, parallelism can decrease latency as well.

_Chapter 2.1: Concurrency is not Parallelism. Page 30._

---

_"Concurrency is a system-structuring mechanism. Parallelism is a resource."_ — It feels like this accurately captures the way we should be thinking about these APIs. What this API returns is not "the amount of concurrency available to the program" which is a property of the program, and thus even with just a single thread is effectively unbounded. But instead it returns "the amount of _parallelism_ available to the program", which is a resource hard-constrained by the machine's capacity (and can be further restricted by e.g. operating systems).

That's why I'd like to propose we rename this API from `available_concurrency` to `available_parallelism`. This still meets the criteria we previously established of not attempting to define what exactly we mean by "hardware", "threads", and other such words. Instead we only talk about "concurrency" as an abstract resource available to our program.

r? @joshtriplett

### DIFF
--- a/library/std/src/sys/hermit/thread.rs
+++ b/library/std/src/sys/hermit/thread.rs
@@ -97,7 +97,7 @@ impl Thread {
     }
 }
 
-pub fn available_concurrency() -> io::Result<NonZeroUsize> {
+pub fn available_parallelism() -> io::Result<NonZeroUsize> {
     unsupported()
 }
 
--- a/library/std/src/sys/sgx/thread.rs
+++ b/library/std/src/sys/sgx/thread.rs
@@ -137,7 +137,7 @@ impl Thread {
     }
 }
 
-pub fn available_concurrency() -> io::Result<NonZeroUsize> {
+pub fn available_parallelism() -> io::Result<NonZeroUsize> {
     unsupported()
 }
 
--- a/library/std/src/sys/unix/thread.rs
+++ b/library/std/src/sys/unix/thread.rs
@@ -263,7 +263,7 @@ impl Drop for Thread {
     }
 }
 
-pub fn available_concurrency() -> io::Result<NonZeroUsize> {
+pub fn available_parallelism() -> io::Result<NonZeroUsize> {
     cfg_if::cfg_if! {
         if #[cfg(any(
             target_os = "android",
--- a/library/std/src/sys/unsupported/thread.rs
+++ b/library/std/src/sys/unsupported/thread.rs
@@ -31,7 +31,7 @@ impl Thread {
     }
 }
 
-pub fn available_concurrency() -> io::Result<NonZeroUsize> {
+pub fn available_parallelism() -> io::Result<NonZeroUsize> {
     unsupported()
 }
 
--- a/library/std/src/sys/wasi/thread.rs
+++ b/library/std/src/sys/wasi/thread.rs
@@ -64,7 +64,7 @@ impl Thread {
     }
 }
 
-pub fn available_concurrency() -> io::Result<NonZeroUsize> {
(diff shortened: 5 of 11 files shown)


joshtriplett:
I don't have any objections to such a rename. Worth thinking about the cost of promoting the terminology distinction, given the divergence from the C++ name. I think people will *understand* what this does under either name. So the question is what will convey the concept most clearly taking into account both existing names and established terminology.

camsteffen:
In the case of hyper-threading, the word "concurrency" seems more accurate - the CPU is not offering X amount of parallelism, but it is recommending X amount of concurrency.

joshtriplett:
Thinking about it further, I think this change seems reasonable. In practice, I think "parallel" is actually the more likely name.

Would you consider also adding doc aliases for both `hardware_concurrency` (for C++ folks) and `available_concurrency` (for folks who search and find references to the previous name)?

joshtriplett:
@rfcbot merge

yoshuawuyts:
> Would you consider also adding doc aliases for both `hardware_concurrency` (for C++ folks) and `available_concurrency` (for folks who search and find references to the previous name)?

Ah yeah, good call. Pushed!

BurntSushi:
The new name SGTM.

The docs still contain a few mentions of the word "hardware." At the very least, we might want to update the opening line to be more consistent with the name of the routine.

I do think it would be nice to perhaps mention the word hardware or "CPU" in the docs somewhere. Something like, "this often corresponds to the number of logical CPUs available to the current program, but not always. For example, in scenarios Foo and Bar, ..." Basically, to give a bit more detail on what folks can realistically expect _in terms of some common environments_.

yoshuawuyts:
> I do think it would be nice to perhaps mention the word hardware or "CPU" in the docs somewhere.

I like that suggestion quite a bit! - If it's okay with y'all, once we merge this I'll submit a follow-up to this PR with docs changes.

Thomasdezeeuw:
A little side note for the overall process: it would really be nice if these kinds of changes, such as renaming, could be announced in the tracking issue. I know this pr links to the tracking issue, but GitHub doesn't send email for that. So for people like myself who mostly follow tracking issue using email I was in for a bit of a surprise that `available_conccurrency` was gone, not knowing it was renamed.

PR #89327, opened 2021-09-28
Pick one possible lifetime in case there are multiple choices

oli-obk:
In case a lifetime variable is created, but doesn't have an obvious lifetime in the list of named lifetimes that it should be inferred to, just pick the first one for the diagnostic.

This happens e.g. in

```rust
fn foo<'a, 'b>(a: Struct<'a>, b: Struct<'b>) -> impl Trait<'a, 'b> {
    if bar() { a } else { b }
}
```

where we get a lifetime variable that combines the lifetimes of `a` and `b` creating a lifetime that is the intersection of both. Right now the type system cannot express this and thus we get an error, but that error also can't express this.

I can also create an entirely new diagnostic that mentions all involved lifetimes, so it would actually mention `'a` and `'b` instead of just `'b`.

### DIFF
--- a/src/test/ui/impl-trait/multiple-lifetimes/ordinary-bounds-unrelated.nll.stderr
+++ b/src/test/ui/impl-trait/multiple-lifetimes/ordinary-bounds-unrelated.nll.stderr
@@ -4,7 +4,11 @@ error[E0700]: hidden type for `impl Trait` captures lifetime that does not appea
 LL | fn upper_bounds<'a, 'b, 'c, 'd, 'e>(a: Ordinary<'a>, b: Ordinary<'b>) -> impl Trait<'d, 'e>
    |                                                                          ^^^^^^^^^^^^^^^^^^
    |
-   = note: hidden type `Ordinary<'_>` captures lifetime '_#9r
+note: hidden type `Ordinary<'b>` captures the lifetime `'b` as defined on the function body at 16:21
+  --> $DIR/ordinary-bounds-unrelated.rs:16:21
+   |
+LL | fn upper_bounds<'a, 'b, 'c, 'd, 'e>(a: Ordinary<'a>, b: Ordinary<'b>) -> impl Trait<'d, 'e>
+   |                     ^^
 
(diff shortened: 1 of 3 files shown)


jackh726:
@nikomatsakis I think this was precisely the kind of example I was thinking of during the discussion of #89056

wesleywiser:
> I can also create an entirely new diagnostic that mentions all involved lifetimes, so it would actually mention `'a` and `'b` instead of just `'b`.

I think that would be ideal but I'm also happy to merge this and put an issue in to improve it later.

wesleywiser (approved):

oli-obk:
@bors r=wesleywiser

I'll open an issue

src/test/ui/impl-trait/multiple-lifetimes/ordinary-bounds-unrelated.nll.stderr
```diff
@@ -4,7 +4,11 @@ error[E0700]: hidden type for `impl Trait` captures lifetime that does not appea
 LL | fn upper_bounds<'a, 'b, 'c, 'd, 'e>(a: Ordinary<'a>, b: Ordinary<'b>) -> impl Trait<'d, 'e>
    |                                                                          ^^^^^^^^^^^^^^^^^^
    |
-   = note: hidden type `Ordinary<'_>` captures lifetime '_#9r
+note: hidden type `Ordinary<'b>` captures the lifetime `'b` as defined on the function body at 16:21
```
lqd, line 7: (This is pre-existing) 

This message mentions `Ordinary<'b>` here as a "hidden type", shouldn't this be "hidden type `impl Trait`" instead, or am I misunderstanding this diagnostic ?

oli-obk: You are completely right.

We should completely rewrite the diagnostics around TAIT in general... they are all not too helpful. I'll open an issue to track the various ways in which it is not great

PR #89329, opened 2021-09-28
print-type-sizes: skip field printing for primitives

### LINKED ISSUE #86528: -Z print-type-sizes: internal error: entered unreachable code: FieldsShape::offset: `Primitive`s have no fields'
### Code

https://github.com/rust-lang/docs.rs/commit/69b1f3d5cdedf0e2feba2d87c64c16a670cdd57a

MCVE:
```rust
pub fn test() {
    Vec::<()>::new()
        .into_iter()
        .filter(|_| true)
        .collect::<Vec<_>>();
}
```

### Meta

`rustc --version`: `rustc 1.54.0-nightly (cc77ba46f 2021-06-03)`

### Error output

```
thread 'rustc' panicked at 'internal error: entered unreachable code: FieldsShape::offset: `Primitive`s have no fields', compiler\rustc_target\src\abi\mod.rs:768:17
stack backtrace:

note: rustc 1.54.0-nightly (cc77ba46f 2021-06-03) running on x86_64-pc-windows-msvc

note: compiler flags: -Z print-type-sizes -C embed-bitcode=no -C debuginfo=2 -C incremental --crate-type lib

query stack during panic:
end of query stack
```

tmiasko:
Fixes #86528.

### DIFF
--- a/compiler/rustc_middle/src/ty/layout.rs
+++ b/compiler/rustc_middle/src/ty/layout.rs
@@ -1805,8 +1805,11 @@ impl<'tcx> LayoutCx<'tcx, TyCtxt<'tcx>> {
 
         match layout.variants {
             Variants::Single { index } => {
-                debug!("print-type-size `{:#?}` variant {}", layout, adt_def.variants[index].ident);
-                if !adt_def.variants.is_empty() {
+                if !adt_def.variants.is_empty() && layout.fields != FieldsShape::Primitive {
+                    debug!(
+                        "print-type-size `{:#?}` variant {}",
+                        layout, adt_def.variants[index].ident
+                    );
                     let variant_def = &adt_def.variants[index];
                     let fields: Vec<_> = variant_def.fields.iter().map(|f| f.ident.name).collect();
                     record(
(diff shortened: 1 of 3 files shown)


compiler/rustc_middle/src/ty/layout.rs
```diff
@@ -1805,8 +1805,11 @@ impl<'tcx> LayoutCx<'tcx, TyCtxt<'tcx>> {
 
         match layout.variants {
             Variants::Single { index } => {
-                debug!("print-type-size `{:#?}` variant {}", layout, adt_def.variants[index].ident);
-                if !adt_def.variants.is_empty() {
+                if !adt_def.variants.is_empty() && layout.fields != FieldsShape::Primitive {
```
jyn514, line 1808: Why is variants non-empty here? Primitives aren't enums.

tmiasko: FieldsShape describes how fields are located in memory, with Primitive used for things that have no fields, including the never type.

Variants that are uninhabited and have ZST fields only are consider to be absent. Layout computation omits absents variants entirely. If all variants of an enum are absent, the layout of a never type is used (Variants::Single, FieldsShape::Primitive).

PR #89332, opened 2021-09-28
[experiment] remove diverge_from in box expr building

nbdd0121:
See if some instruction count regressions can be regained from #89030

r? @ghost

### DIFF
--- a/compiler/rustc_mir_build/src/build/expr/as_rvalue.rs
+++ b/compiler/rustc_mir_build/src/build/expr/as_rvalue.rs
@@ -136,7 +136,8 @@ impl<'a, 'tcx> Builder<'a, 'tcx> {
                         fn_span: expr_span,
                     },
                 );
-                this.diverge_from(block);
+                // FIXME(nbdd0121): add this back when oom=panic is implemented.
+                // this.diverge_from(block);
                 block = success;
 
                 // The `Box<T>` temporary created here is not a part of the HIR,


nbdd0121:
@rylev could you do a perf run please?

nbdd0121:
Trying another approach. This should have similar result, but the change is made in library not in compiler. Please have another perf run.

nbdd0121:
As I expected the performance is largely similar to removing the diverging path. I have no idea why adding rustc_allocator_nounwind makes expand_crate slower though.

JohnCSimon:
This needs to assign a reviewer.

camelid:
r? rust-lang/libs

the8472:
> This needs to assign a reviewer.

does it? it's still labeled as experiment

nbdd0121:
I am unsure whether this should proceed. I can turn this into a PR that removes the unwinding path and gain a bit perf now, but it'll need to reverted anyway to support oom=panic.

nbdd0121:
Also this is more T-compiler than T-libs.

@rustbot label: -T-libs +T-compiler

camelid:
r? rust-lang/compiler

pnkfelix:
> 

Discussed in T-compiler meeting.

Given that our  trajectory is to support oom=panic (PR #88098, Tracking Issue #43596), we do not think we will be landing this.

It was good to gather the performance data, though!

Issue #62195, opened 2019-06-27
Tracking issue for `Vec::leak`


cramertj:
Similar to `Box::leak`, this function would allow leaking a `Vec` and getting back an `&'a mut [T]`.

jonas-schievink:
Implementation in https://github.com/rust-lang/rust/pull/62196

SimonSapin:
`Box::leak` is an associated function instead of a method to avoid shadowing a possible `Foo::leak` method in `Box<Foo>`, normally accessible through auto-deref.

This is less relevant to `Vec::leak` since `Vec<T>` dereferences to `[T]`. There is no `[T]::leak` method in the standard library, so making `Vec::leak` a method would only shadow such a method if it’s defined in a user trait.

@cramertj Do you think it should still be an associated function?

Lucretiel:
Is there any reason to not have this method return an `&'static [T]`?

rodrigorc:
I find weird that there is a `Vec::leak`, but not a `String::leak`, `PathBuf::leak`, `OsString::leak` or `CString::leak`.

IMHO, if `Vec::leak` exists, then any standard type that owns data and has a `into_boxed_whatever` member function should be able to leak too.

cramertj:
> IMHO, if Vec::leak exists, then any standard type that owns data and has a into_boxed_whatever member function should be able to leak too.

Seems fine to me. Sadly, I don't think there's a trait that covers these, so they'll all have to be added independently.

rodrigorc:
> Sadly, I don't think there's a trait that covers these, so they'll all have to be added independently.

I don't think the lack of trait is a problem at all. For example I would expect any standard type that has a `len()` function to have an `is_empty()` too, no trait involved. It's like an expected pattern.

lambda-fairy:
@Lucretiel the lifetime parameter is necessary in case `T` has non-`'static` lifetimes, e.g. if it's a reference ([playground](https://play.rust-lang.org/?version=nightly&mode=debug&edition=2018&gist=a4cd4a4927f4d87260ff2976dfa25cf3)).

`Box::leak` has a lifetime parameter for the same reason.

shepmaster:
What do people think about changing the argument type to be more generic:

```rust
pub fn leak<'a>(vec: impl Into<Vec<T>>) -> &'a mut [T]
where
    T: 'a, 
```

I'm thinking about adding a similar `String::leak` and it'd be nice to do `String::leak("hello")`.

[...]

PR #77709, opened 2020-10-08
Link Vec leak doc to Box

pickfire:

PR #89337, opened 2021-09-28
Avoid allocations and copying in Vec::leak

mbrubeck:
The [`Vec::leak`] method (#62195) is currently implemented by calling `Vec::into_boxed_slice` and `Box::leak`.  This shrinks the vector before leaking it, which potentially causes a reallocation and copies the vector's contents.

By avoiding the conversion to `Box`, we can instead leak the vector without any expensive operations, just by returning a slice reference and forgetting the `Vec`.  Users who *want* to shrink the vector first can still do so by calling `shrink_to_fit` explicitly.

**Note:**  This could break code that uses `Box::from_raw` to “un-leak” the slice returned by `Vec::leak`.  However, the `Vec::leak` docs explicitly forbid this, so such code is already incorrect.

[`Vec::leak`]: https://doc.rust-lang.org/stable/std/vec/struct.Vec.html#method.leak

### DIFF
--- a/library/alloc/src/vec/mod.rs
+++ b/library/alloc/src/vec/mod.rs
@@ -1952,10 +1952,11 @@ impl<T, A: Allocator> Vec<T, A> {
     /// `'a`. If the type has only static references, or none at all, then this
     /// may be chosen to be `'static`.
     ///
-    /// This method does not reallocate or shrink the `Vec`, so the leaked
-    /// allocation may include unused capacity that is not part of the returned
-    /// slice.  Unsafe code that later reconstructs or deallocates the `Vec`
-    /// (for example, by calling [`Vec::from_raw_parts`]) must keep track of the
+    /// As of Rust 1.57, this method does not reallocate or shrink the `Vec`,
+    /// so the leaked allocation may include unused capacity that is not part
+    /// of the returned slice.
+    /// Unsafe code that later reconstructs or deallocates the `Vec` (for
+    /// example, by calling [`Vec::from_raw_parts`]) must keep track of the
     /// original capacity.
     ///
     /// This function is mainly useful for data that lives for the remainder of
--- a/library/alloc/src/vec/mod.rs
+++ b/library/alloc/src/vec/mod.rs
@@ -1952,8 +1952,11 @@ impl<T, A: Allocator> Vec<T, A> {
     /// `'a`. If the type has only static references, or none at all, then this
     /// may be chosen to be `'static`.
     ///
-    /// This function is similar to the [`leak`][Box::leak] function on [`Box`]
-    /// except that there is no way to recover the leaked memory.
+    /// This method does not reallocate or shrink the `Vec`, so the leaked
+    /// allocation may include unused capacity that is not part of the returned
+    /// slice.  Unsafe code that later reconstructs or deallocates the `Vec`
+    /// (for example, by calling [`Vec::from_raw_parts`]) must keep track of the
+    /// original capacity.
     ///
     /// This function is mainly useful for data that lives for the remainder of
     /// the program's life. Dropping the returned reference will cause a memory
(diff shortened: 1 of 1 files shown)


mbrubeck:
This was discussed previously at: https://github.com/rust-lang/rfcs/pull/2969#pullrequestreview-485089769

cuviper:
> **Note:** This could break code that uses `Box::from_raw` to “un-leak” the slice returned by `Vec::leak`. However, the `Vec::leak` docs explicitly forbid this, so such code is already incorrect.

That clause was added in #77709, but saying "there is no way" is a pretty casual way to forbid this IMO.

I think we should make an explicit guarantee about the capacity here, so there _would_ be a way to use `Vec::from_raw_parts`. Either we do shrink and you can use `len == capacity`, or we don't shrink and you can call `capacity()` beforehand. If we don't shrink, the point about data movement is also a useful performance guarantee to document.

mbrubeck:
I pushed a new commit updating the docs to say:

> This method does not reallocate or shrink the `Vec`, so the leaked allocation may include unused capacity that is not part of the returned slice.  Unsafe code that later reconstructs or deallocates the `Vec` (for example, by calling [`Vec::from_raw_parts`]) must keep track of the original capacity.

(This replaces the paragraph that said “there is no way to recover the leaked memory.“)


[`Vec::from_raw_parts`]: https://doc.rust-lang.org/std/vec/struct.Vec.html#method.from_raw_parts

cuviper:
I'm happy with that, but @rust-lang/libs-api should discuss that as a new API guarantee.

dtolnay (approved):

dtolnay:
I was not involved in the discussion over at https://github.com/rust-lang/rfcs/pull/2969, but I think this is pretty clearly the right behavior, for the reasons described by @m-ou-se in https://github.com/rust-lang/rfcs/pull/2969#pullrequestreview-485089769.

- Using "leak" to mean "leak A COPY OF, but destroy the argument" rather than "leak the argument" is uncharacteristic.

- If leak does not shrink, getting shrink behavior is as simple as `vec.shrink_to_fit(); vec.leak()`, or `Box::leak(vec.into_boxed_slice())`. If leak does shrink, getting the non-shrinking behavior requires the user to write unsafe code.

I'll add one more:

- If the user has built a data structure containing pointers into the vector, it's reasonable to want to be able to leak without all those pointers getting invalidated.

@rfcbot fcp merge

jplatte:
> I pushed a new commit updating the docs to say:
> 
> > This method does not reallocate or shrink the `Vec`, so the leaked allocation may include unused capacity that is not part of the returned slice.  Unsafe code that later reconstructs or deallocates the `Vec` (for example, by calling [`Vec::from_raw_parts`](https://doc.rust-lang.org/std/vec/struct.Vec.html#method.from_raw_parts)) must keep track of the original capacity.
> 
> (This replaces the paragraph that said “there is no way to recover the leaked memory.“)

Maybe that should be prefaced by "As of <version>"? I think it's not that uncommon to look at the "regular" stable docs but use an older compiler.

mbrubeck:
> IME it's not that uncommon to look at the "regular" stable docs but use an older compiler / std.

Hmm, it's even worse than that.  Even if the author of the code is using the most recent toolchain, there's no easy way to prevent users of the code from compiling it with an older toolchain.  For example, if a published library contains unsafe code that relies on this guarantee, it would have **undefined behavior** when any project using the library is built with an older toolchain, where the guarantee does not hold.

This seems bad enough that we might want to avoid making any guarantee at all that doesn't apply across toolchain versions.  But if we do document this guarantee, I agree that we need to document when it changed.

jplatte:
Well, once 1.56 lands, people will be able to prevent compilation using an older rustc, at least in Cargo projects, through [rust-version](https://doc.rust-lang.org/beta/cargo/reference/manifest.html#the-rust-version-field) in `Cargo.toml` :)

RustyYato:
Note: if the `len != cap` then the leaked slice doesn't have provenance over the unused capacity. So reconstructing via `from_raw_parts` is unsound. This should be explicitly documented.

cc @RalfJung

PR #89341, opened 2021-09-28
Deriving: Include bound generic params in type parameters for where clause

### LINKED ISSUE #89188: derive(Clone) with HRTB and GAT
I tried this code:

```rust

trait CallWithShim: Sized {
    type Shim<'s>
    where
        Self: 's;
        
    fn call(&mut self, method: ShimMethod<Self>);
}

#[derive(Clone)]
struct ShimMethod<T: CallWithShim + 'static>(
    pub &'static dyn for<'s> Fn(&'s mut T::Shim<'s>),
);

fn main() {}
```

I expected to see this happen: *it compiles*

Instead, this happened: *it fails to compile*, with
```
  --> src/main.rs:13:49
   |
11 | #[derive(Clone)]
   |          - help: consider introducing lifetime `'s` here: `'s,`
12 | struct ShimMethod<T: CallWithShim + 'static>(
13 |     pub &'static dyn for<'s> Fn(&'s mut T::Shim<'s>),
   |                                                 ^^ undeclared lifetime
```

Removing `#[derive(Clone)]` makes it compile.

A bit longer playground demonstrating what this code snippet can be used for:
https://play.rust-lang.org/?version=nightly&mode=debug&edition=2018&gist=a20f6a74ae1c1e8b2838aa428beee1fe

I'm not an experienced user of HRTBs (nor GATs, obviously). What I'm trying to express is that the data borrowed by `Shim` does not outlive the `ShimMethod` call. I _think_ the HRTB is the right way to express that, and it all compiles and works without the `#[derive(Clone)]`, but I'm a bit unsure if I'm really doing something wrong and stupid and the compiler fails to catch it because potential GAT bugs 🤷 

Anyway, it seems there is _something_ fishy about the `derive` code here..

### Meta
`rustc --version --verbose`:
```
rustc 1.57.0-nightly (ac2d9fc50 2021-09-21)
binary: rustc
commit-hash: ac2d9fc509e36d1b32513744adf58c34bcc4f43c
commit-date: 2021-09-21
host: x86_64-apple-darwin
release: 1.57.0-nightly
LLVM version: 13.0.0
```

audunhalland:
Fixes #89188.

The `derive` macro ignored the `for<'s>` needed with the `Fn` trait in that code example.

edit: I'm unsure if this might cause regressions. I'm not an experienced compiler developer so I'm not used to thinking about unwanted side effects code changes like this might have.

### DIFF
--- a/compiler/rustc_builtin_macros/src/deriving/generic/mod.rs
+++ b/compiler/rustc_builtin_macros/src/deriving/generic/mod.rs
@@ -332,20 +332,27 @@ pub fn combine_substructure(
     RefCell::new(f)
 }
 
+struct TypeParameter {
+    bound_generic_params: Vec<ast::GenericParam>,
+    ty: P<ast::Ty>,
+}
+
 /// This method helps to extract all the type parameters referenced from a
 /// type. For a type parameter `<T>`, it looks for either a `TyPath` that
 /// is not global and starts with `T`, or a `TyQPath`.
+/// Also include bound generic params from the input type.
 fn find_type_parameters(
     ty: &ast::Ty,
     ty_param_names: &[Symbol],
     cx: &ExtCtxt<'_>,
-) -> Vec<P<ast::Ty>> {
+) -> Vec<TypeParameter> {
     use rustc_ast::visit;
 
     struct Visitor<'a, 'b> {
         cx: &'a ExtCtxt<'b>,
         ty_param_names: &'a [Symbol],
-        types: Vec<P<ast::Ty>>,
+        bound_generic_params_stack: Vec<ast::GenericParam>,
+        type_params: Vec<TypeParameter>,
     }
 
     impl<'a, 'b> visit::Visitor<'a> for Visitor<'a, 'b> {
@@ -353,7 +360,10 @@ fn find_type_parameters(
             if let ast::TyKind::Path(_, ref path) = ty.kind {
                 if let Some(segment) = path.segments.first() {
                     if self.ty_param_names.contains(&segment.ident.name) {
-                        self.types.push(P(ty.clone()));
+                        self.type_params.push(TypeParameter {
+                            bound_generic_params: self.bound_generic_params_stack.clone(),
+                            ty: P(ty.clone()),
+                        });
                     }
                 }
             }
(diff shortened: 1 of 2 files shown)


compiler/rustc_builtin_macros/src/deriving/generic/mod.rs
```diff
@@ -332,44 +332,74 @@ pub fn combine_substructure(
...
             visit::walk_ty(self, ty)
         }
 
+        // Place bound generic params on a stack, to extract them when a type is encountered.
+        fn visit_poly_trait_ref(
+            &mut self,
+            trait_ref: &'a ast::PolyTraitRef,
+            modifier: &'a ast::TraitBoundModifier,
+        ) {
+            let stack_len = trait_ref.bound_generic_params.len();
```
audunhalland, line 380: This is a bug. It was supposed to take the length of `self.bound_generic_params_stack`.

jackh726:
I think this looks correct. I was trying to come up with an example without GATs where we could run into this case, but I haven't been able to yet. We basically have to name a late-bound lifetime within the type but without *also* being required to name it outside. (So, `<T as Trait<'a>::As` wouldn't work, because we would need a `T: Trait<'a>` bound).

I think I'd like to see the following added to the test:
```rust
trait CallWithShim2: Sized {
    type Shim<T>;
}

struct S<'s>(&'s ());

#[derive(Clone)]
struct ShimMethod2<T: CallWithShim2 + 'static>(
    pub &'static dyn for<'s> Fn(&'s mut T::Shim<S<'s>>),
);

trait Trait<'s, 't> {}

#[derive(Clone)]
struct ShimMethod3<T: CallWithShim2 + 'static>(
    pub &'static dyn for<'s> Fn(&'s mut T::Shim<dyn for<'t> Trait<'s, 't>>),
);

trait Trait2 {
    type As;
}

#[derive(Clone)]
struct ShimMethod4<T: Trait2 + 'static>(
    pub &'static dyn for<'s> Fn(&'s mut T::As),
);
```

compiler/rustc_builtin_macros/src/deriving/generic/mod.rs
```diff
@@ -332,44 +332,74 @@ pub fn combine_substructure(
...
 
+        // Place bound generic params on a stack, to extract them when a type is encountered.
+        fn visit_poly_trait_ref(
+            &mut self,
+            trait_ref: &'a ast::PolyTraitRef,
+            modifier: &'a ast::TraitBoundModifier,
+        ) {
+            let stack_len = self.bound_generic_params_stack.len();
+            self.bound_generic_params_stack
+                .extend(trait_ref.bound_generic_params.clone().into_iter());
```
audunhalland, line 382: It would probably break by HRTB variable shadowing, e.g.

`for<'a> (...   for<'a>  ( ... )) `

by outputting `for<'a, 'a>`? Unsure if it's important enough to address?

jackh726: I was trying to come up with a test case where this might happen. But I'm unsure; I couldn't find one.

jackh726:
Okay, can you instead actually change the test case to:
```rust
trait Trait<'s, 't, 'u> {}

#[derive(Clone)]
struct ShimMethod3<T: CallWithShim2 + 'static>(
    pub &'static dyn for<'s> Fn(&'s mut T::Shim<dyn for<'t> Fn(&'s mut T::Shim<dyn for<'u> Trait<'s, 't, 'u>>)>),
);
```
This better tests multiple levels of binding. Made me remember that we don't allow lifetime shadowing, so we should be good.

jackh726:
After that change, this PR LGTM to merge.

jackh726:
@bors r+ rollup=never

Given that this effects built-in derives, I want to make sure this is easy to bisect if there are any regressions.

PR #85830, opened 2021-05-30
 Avoid a branch on key being local for queries that use the same local and extern providers

bjorn3:
Currently based on https://github.com/rust-lang/rust/pull/85810 as it slightly conflicts with it. Only the last two commits are new.

PR #89343, opened 2021-09-28
Refactor fingerprint reconstruction

Mark-Simulacrum:
This PR replaces can_reconstruct_query_key with fingerprint_style, which returns the style of the fingerprint for that query. This allows us to avoid trying to extract a DefId (or equivalent) from keys which *are* reconstructible because they're () but not as DefIds.

This is done with the goal of fixing -Zdump-dep-graph, which seems to have broken a while ago (I didn't try to bisect). Currently even on a `fn main() {}` file it'll ICE (you need to also pass -Zquery-dep-graph for it to work at all), and this patch indirectly fixes the cause of that ICE. This also adds a test for it continuing to work.


Mark-Simulacrum:
cc @rust-lang/wg-incr-comp -- I'm not sure if folks recall the intent of has_params, which might be helpful to determine if this is the "right" thing to do.

compiler/rustc_middle/src/ty/query.rs
```diff
@@ -114,8 +114,8 @@ macro_rules! query_storage {
     ([storage($ty:ty) $($rest:tt)*][$K:ty, $V:ty]) => {
         <$ty as CacheSelector<$K, $V>>::Cache
     };
-    ([$other:ident $(($($other_args:tt)*))* $(, $($modifiers:tt)*)*][$($args:tt)*]) => {
-        query_storage!([$($($modifiers)*)*][$($args)*])
+    ([$modifier:tt $($rest:tt)*][$($args:tt)*]) => {
```
Mark-Simulacrum, line 117: FWIW as noted in the commit message, I don't really fully understand what is going on here - I was getting pretty weird errors without the first commit though. You can see a pretty minimal example of this here - https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=de08d72c9a23062bf3d224f0d873d27d.

I think the changes to the macro are correct, but I'm not an expert with macro_rules so I definitely could have gotten this wrong. Happy to adjust!

cjgillot: This is a known bug in the macro. 0358edeb546dbdc9d2f894c19b54c07884d9b9c8 should fix it.

Mark-Simulacrum: Hm, the first commit in this PR looks simpler than that one, but happy to rebase over that (not sure where that's from -- GitHub isn't showing a PR link).

cjgillot: It's from #85830.

cjgillot:
IIUC, the crash happens when trying to debug-print a DepNode whch parameter () by mistakenly trying to extract a DefId from a Fingerprint which is not a proper DefPathHash. The bug is therefore in `rustc_middle::dep_graph::dep_node::DepNodeExt::extract_def_id` which tests `can_reconstruct_query_key()` to know whether the query key is a `CrateNum`/`LocalDefId`/`DefId`. I think this PR hides the problem rather than solving it. The correct fix would be to create a separate `has_def_path_hash` field in `DepKindStruct` which tests whether the query key is any of the three, and equivalently if a DefId can be extracted from the fingerprint.

Mark-Simulacrum:
Hm, that seems like a good thing to tackle, I agree. I think this PR is still desirable (or we should just remove has_params), since it seems like the meaning of that is not too clear right now.

I'll have to dig into the 2 manually declared queries with has_params: false, and see if that's significant for them...

Mark-Simulacrum:
Alright, replaced pretty much the whole PR with different patch -- and extracted https://github.com/rust-lang/rust/pull/89466 for the base macro changes which are not actually needed for this PR now.

I looked into fully removing has_params but didn't end up making that change. The only uses left are the debug assert in DepKind::new_no_params and the debug_node method. The latter could be removed with no loss, I believe -- just a difference in how nodes look (printed as `foo` today, `foo()` with that condition dropped). Since has_params is true currently for all but Null and TraitSelect DepKinds (neither of which is a proper query), it's probably the case that this could be refactored. But seems like future work.

compiler/rustc_middle/src/dep_graph/dep_node.rs
```diff
@@ -350,14 +352,16 @@ impl DepNodeExt for DepNode {
...
-        }
-
-        if kind.has_params {
-            Ok(DepNode::from_def_path_hash(def_path_hash, kind))
-        } else {
-            Ok(DepNode::new_no_params(kind))
+        match kind.fingerprint_style() {
+            FingerprintStyle::Opaque => Err(()),
+            FingerprintStyle::Unit => {
+                if !kind.has_params {
```
cjgillot, line 358: `FingerprintStyle::Unit` should now be equivalent to `has_params == false`, isn't it? We should be able to get rid of `has_params`.

Mark-Simulacrum: Not currently. A query with `()` key currently "has_params"; the only two things to not have params are Null and TraitSelect which are not generated through the normal macro infrastructure. dropped the "fix" for that from this branch for now, though I can re-add it, it's not big (b2b53bb9a4f2f5e1f96f8f01ad7b378f5eb1b4d1).

workingjubilee:
This looks like yours? Apologies if I am wrong.
```
RUST_DEP_GRAPH=/checkout/obj/build/x86_64-unknown-linux-gnu/test/run-make/dep-graph/dep-graph/dep-graph LD_LIBRARY_PATH="/checkout/obj/build/x86_64-unknown-linux-gnu/test/run-make/dep-graph/dep-graph:/checkout/obj/build/x86_64-unknown-linux-gnu/stage2/lib:/checkout/obj/build/x86_64-unknown-linux-gnu/stage0-bootstrap-tools/x86_64-unknown-linux-gnu/release/deps:/checkout/obj/build/x86_64-unknown-linux-gnu/stage0/lib" '/checkout/obj/build/x86_64-unknown-linux-gnu/stage2/bin/rustc' --out-dir /checkout/obj/build/x86_64-unknown-linux-gnu/test/run-make/dep-graph/dep-graph -L /checkout/obj/build/x86_64-unknown-linux-gnu/test/run-make/dep-graph/dep-graph  -Clinker='arm-none-eabi-gcc' \
        -Cincremental=/checkout/obj/build/x86_64-unknown-linux-gnu/test/run-make/dep-graph/dep-graph/incr \
        -Zquery-dep-graph -Zdump-dep-graph foo.rs
stderr:
error: linking with `arm-none-eabi-gcc` failed: exit status: 1
  |
  = note: "arm-none-eabi-gcc" "-m64" "/checkout/obj/build/x86_64-unknown-linux-gnu/test/run-make/dep-graph/dep-graph/foo.18u5b2jmv51dofh3.rcgu.o" "/checkout/obj/build/x86_64-unknown-linux-gnu/test/run-make/dep-graph/dep-graph/foo.25sg5ox0q16b7auf.rcgu.o" "/checkout/obj/build/x86_64-unknown-linux-gnu/test/run-make/dep-graph/dep-graph/foo.3c6c78fzwfnp5rr9.rcgu.o" "/checkout/obj/build/x86_64-unknown-linux-gnu/test/run-make/dep-graph/dep-graph/foo.3pwxpcv7ig8jfb30.rcgu.o" "/checkout/obj/build/x86_64-unknown-linux-gnu/test/run-make/dep-graph/dep-graph/foo.3wccqld67w709v08.rcgu.o" "/checkout/obj/build/x86_64-unknown-linux-gnu/test/run-make/dep-graph/dep-graph/foo.4rlnbcoxr7ik1e3w.rcgu.o" "/checkout/obj/build/x86_64-unknown-linux-gnu/test/run-make/dep-graph/dep-graph/foo.4unz5qo9qikgp26p.rcgu.o" "/checkout/obj/build/x86_64-unknown-linux-gnu/test/run-make/dep-graph/dep-graph/foo.50rnet0thytpgybx.rcgu.o" "-Wl,--as-needed" "-L" "/checkout/obj/build/x86_64-unknown-linux-gnu/test/run-make/dep-graph/dep-graph" "-L" "/checkout/obj/build/x86_64-unknown-linux-gnu/stage2/lib/rustlib/x86_64-unknown-linux-gnu/lib" "-Wl,--start-group" "-Wl,-Bstatic" "/checkout/obj/build/x86_64-unknown-linux-gnu/stage2/lib/rustlib/x86_64-unknown-linux-gnu/lib/libstd-5548ef64e6092712.rlib" "/checkout/obj/build/x86_64-unknown-linux-gnu/stage2/lib/rustlib/x86_64-unknown-linux-gnu/lib/libpanic_unwind-a8f484438681c15e.rlib" "/checkout/obj/build/x86_64-unknown-linux-gnu/stage2/lib/rustlib/x86_64-unknown-linux-gnu/lib/libminiz_oxide-b69fd8507c8409af.rlib" "/checkout/obj/build/x86_64-unknown-linux-gnu/stage2/lib/rustlib/x86_64-unknown-linux-gnu/lib/libadler-2dff396b99681a6b.rlib" "/checkout/obj/build/x86_64-unknown-linux-gnu/stage2/lib/rustlib/x86_64-unknown-linux-gnu/lib/libobject-a78bd5aa183c7115.rlib" "/checkout/obj/build/x86_64-unknown-linux-gnu/stage2/lib/rustlib/x86_64-unknown-linux-gnu/lib/libmemchr-846c64d5a2fbc5ee.rlib" "/checkout/obj/build/x86_64-unknown-linux-gnu/stage2/lib/rustlib/x86_64-unknown-linux-gnu/lib/libaddr2line-24c09062b6dc787b.rlib" "/checkout/obj/build/x86_64-unknown-linux-gnu/stage2/lib/rustlib/x86_64-unknown-linux-gnu/lib/libgimli-ce93f351ca41a57b.rlib" "/checkout/obj/build/x86_64-unknown-linux-gnu/stage2/lib/rustlib/x86_64-unknown-linux-gnu/lib/libstd_detect-5b404d65540eadf8.rlib" "/checkout/obj/build/x86_64-unknown-linux-gnu/stage2/lib/rustlib/x86_64-unknown-linux-gnu/lib/librustc_demangle-514cb174319eb6d4.rlib" "/checkout/obj/build/x86_64-unknown-linux-gnu/stage2/lib/rustlib/x86_64-unknown-linux-gnu/lib/libhashbrown-37c1534c42bc4f8c.rlib" "/checkout/obj/build/x86_64-unknown-linux-gnu/stage2/lib/rustlib/x86_64-unknown-linux-gnu/lib/librustc_std_workspace_alloc-06d0f7780fda2fd9.rlib" "/checkout/obj/build/x86_64-unknown-linux-gnu/stage2/lib/rustlib/x86_64-unknown-linux-gnu/lib/libunwind-126d4eb258557375.rlib" "/checkout/obj/build/x86_64-unknown-linux-gnu/stage2/lib/rustlib/x86_64-unknown-linux-gnu/lib/libcfg_if-88bb1ba4dd271224.rlib" "/checkout/obj/build/x86_64-unknown-linux-gnu/stage2/lib/rustlib/x86_64-unknown-linux-gnu/lib/liblibc-b4a707e310d7480c.rlib" "/checkout/obj/build/x86_64-unknown-linux-gnu/stage2/lib/rustlib/x86_64-unknown-linux-gnu/lib/liballoc-0b5a376b429677a6.rlib" "/checkout/obj/build/x86_64-unknown-linux-gnu/stage2/lib/rustlib/x86_64-unknown-linux-gnu/lib/librustc_std_workspace_core-2a6a2797f7a73818.rlib" "/checkout/obj/build/x86_64-unknown-linux-gnu/stage2/lib/rustlib/x86_64-unknown-linux-gnu/lib/libcore-0e3656b1fda5fd7b.rlib" "-Wl,--end-group" "/checkout/obj/build/x86_64-unknown-linux-gnu/stage2/lib/rustlib/x86_64-unknown-linux-gnu/lib/libcompiler_builtins-eecefd843a0dbc02.rlib" "-Wl,-Bdynamic" "-lgcc_s" "-lutil" "-lrt" "-lpthread" "-lm" "-ldl" "-lc" "-Wl,--eh-frame-hdr" "-Wl,-znoexecstack" "-L" "/checkout/obj/build/x86_64-unknown-linux-gnu/stage2/lib/rustlib/x86_64-unknown-linux-gnu/lib" "-o" "/checkout/obj/build/x86_64-unknown-linux-gnu/test/run-make/dep-graph/dep-graph/foo" "-Wl,--gc-sections" "-pie" "-Wl,-zrelro" "-Wl,-znow" "-nodefaultlibs"
  = note: arm-none-eabi-gcc: error: unrecognized command line option '-m64'
```
@bors r-

PR #86298, opened 2021-06-14
refactor LoweringContext::lower_generics_mut

tlyu:
Reduce nesting by using early exits and factoring out a helper
function. Also add comments explaining why the copying of
`?Sized` bounds is necessary.

r? @estebank 
@rustbot label +A-hir +A-traits +C-cleanup +T-compiler

PR #89344, opened 2021-09-28
Cleanup lower_generics_mut and make span be the bound itself

jackh726:
Closes #86298 (supersedes those changes)

r? @cjgillot since you reviewed the other PR

(Used wrong branch for #89338)

### DIFF
--- a/compiler/rustc_ast_lowering/src/item.rs
+++ b/compiler/rustc_ast_lowering/src/item.rs
@@ -1356,32 +1356,45 @@ impl<'hir> LoweringContext<'_, 'hir> {
         // keep track of the Span info. Now, `add_implicitly_sized` in `AstConv` checks both param bounds and
         // where clauses for `?Sized`.
         for pred in &generics.where_clause.predicates {
-            if let WherePredicate::BoundPredicate(ref bound_pred) = *pred {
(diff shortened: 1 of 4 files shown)


compiler/rustc_ast_lowering/src/item.rs
```diff
@@ -1356,32 +1356,46 @@ impl<'hir> LoweringContext<'_, 'hir> {
...
+            };
+            let mut is_param: Option<bool> = None;
+            for bound in &bound_pred.bounds {
+                if !matches!(*bound, GenericBound::Trait(_, TraitBoundModifier::Maybe)) {
+                    continue;
+                }
+                // We only need to compute this once per `WherePredicate`, but don't
+                // need to compute this at all unless there is a Maybe bound.
+                // This closure should be able to be moved out of the loop,
+                // but `get_partial_res` takes a `&mut self`
```
cjgillot, line 1371: `get_partial_res` should take `&self`, it does not need to be mutable.

jackh726: Yeah, I wasn't sure if it was worth changing that for; I wasn't sure if there's some design reason that it's `&mut self`. I'll go ahead and do that though.

jackh726:
@cjgillot updated

PR #89351, opened 2021-09-29
for signed wrapping remainder, do not compare lhs with MIN

tspiteri:
Since the wrapped remainder is going to be 0 for all cases when the rhs is -1, there is no need to compare the lhs with MIN.

### DIFF
--- a/library/core/src/num/int_macros.rs
+++ b/library/core/src/num/int_macros.rs
@@ -1533,9 +1533,8 @@ macro_rules! int_impl {
         #[must_use = "this returns the result of the operation, \
                       without modifying the original"]
         pub const fn overflowing_rem(self, rhs: Self) -> (Self, bool) {
-            // Using `&` helps LLVM see that it is the same check made in division.
-            if unlikely!((self == Self::MIN) & (rhs == -1)) {
-                (0, true)
+            if unlikely!(rhs == -1) {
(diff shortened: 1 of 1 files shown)


tspiteri:
This results in slightly better code generation for [`wrapping_rem`](https://godbolt.org/z/qdbbP1xbP) and [`wrapping_rem_euclid`](https://godbolt.org/z/4Wre4415W).

dtolnay (changes requested):
This leaves `overflowing_rem` still generating poor code though, right? The current implementation is:

```rust
if unlikely!(self == Self::MIN && rhs == -1) {
    (0, true)
} else {
    (self % rhs, false)
}
```

If I've correctly understood the optimization in this PR, it seems like we could change that to:

```rust
if unlikely!(rhs == -1) {
    (0, self == Self::MIN)
} else {
    (self % rhs, false)
}
```

after which changing `wrapping_rem` would probably no longer be necessary (but check godbolt).

tspiteri:
Yes, after that suggestion, my original change is [no longer necessary](https://godbolt.org/z/q91adfa6d) (and [similarly for `rem_euclid`](https://godbolt.org/z/Edxxs5c4d)).

dtolnay (approved):

Issue #50186, opened 2018-04-23
Tracking issue for Cell::update


ghost:
This issue tracks the `cell_update` feature.

The feature adds `Cell::update`, which allows one to more easily modify the inner value.
For example, instead of `c.set(c.get() + 1)` now you can just do `c.update(|x| x + 1)`:

```rust
let c = Cell::new(5);
let new = c.update(|x| x + 1);

assert_eq!(new, 6);
assert_eq!(c.get(), 6);
```

oli-obk:
Not a blocking issue for merging this as an unstable feature, so I'm registering my concern here:

The `update` function returning its new value feels very `++i`-ish. Similar to `+=` returning the result of the operation (which rust luckily does not!). I don't think we should be doing this. Options I see:

* return `()` (that was the suggestion in the PR)
* make the closure argument a `&mut T`, and allow the closure to return any value.
    * if users want the "return new/old value" behaviour, they can now encode it themselves via `|x| {*x += 1; *y }`

Maybe there are other options?

Data point: the `volatile` crate's `update` method returns `()` (https://github.com/embed-rs/volatile/blob/master/src/lib.rs#L134)

stjepang:
@oli-obk 

I like the second option (taking a `&mut T` in the closure). There's an interesting variation on this option - instead of requiring `T: Copy` we could require `T: Default` to support updates like this one:

```rust
let c = Cell::new(Vec::new());
c.update(|v| v.push("foo"));
```

Perhaps we could have two methods with different names that clearly reflect how they internally work?

```rust
fn get_update<F>(&self, f: F) where T: Copy, F: FnMut(&mut T);
fn take_update<F>(&self, f: F) where T: Default, F: FnMut(&mut T);
```

stjepang:
What do you think, @SimonSapin? Would two such methods, `get_update` and `take_update`, be a more appropriate interface?

SimonSapin:
The `T: Default` flavor *feels* kinda awkward but I don’t know how to put that in more concrete terms, sorry :/

It looks like there is a number of slightly different possible APIs for this. I don’t have a strong opinion on which is (or are) preferable.

CodeSandwich:
Taking `&mut` in this method would be awesome, I would love to use it.

Unfortunately it seems that it would have to be `unsafe`, because AFAIK there is no way in Rust to restrict usage like this:
```
let c = Cell::new(vec![123]);
c.update(|vec| {
[...]

PR #68970, opened 2020-02-08
Implement a feature for a sound specialization subset

matthewjasper:
This implements a new feature (`min_specialization`) that restricts specialization to a subset that is reasonable for the standard library to use.

The plan is to then:

* Update `libcore` and `liballoc` to compile with `min_specialization`.
* Add a lint to forbid use of `feature(specialization)` (and other unsound, type system extending features) in the standard library.
* Fix the soundness issues around `specialization`.
* Remove `min_specialization`

The rest of this is an overview from a comment in this PR

## Basic approach

To enforce this requirement on specializations we take the following approach:
1. Match up the substs for `impl2` so that the implemented trait and self-type match those for `impl1`.
2. Check for any direct use of `'static` in the substs of `impl2`.
3. Check that all of the generic parameters of `impl1` occur at most once in the *unconstrained* substs for `impl2`. A parameter is constrained if its value is completely determined by an associated type projection predicate.
4. Check that all predicates on `impl1` also exist on `impl2` (after matching substs).

## Example

Suppose we have the following always applicable impl:

```rust
impl<T> SpecExtend<T> for std::vec::IntoIter<T> { /* specialized impl */ }
impl<T, I: Iterator<Item=T>> SpecExtend<T> for I { /* default impl */ }
```

We get that the subst for `impl2` are `[T, std::vec::IntoIter<T>]`. `T` is constrained to be `<I as Iterator>::Item`, so we check only `std::vec::IntoIter<T>` for repeated parameters, which it doesn't have. The predicates of `impl1` are only `T: Sized`, which is also a predicate of impl2`. So this specialization is sound.

## Extensions

Unfortunately not all specializations in the standard library are allowed by this. So there are two extensions to these rules that allow specializing on some traits.

### rustc_specialization_trait

If a trait is always applicable, then it's sound to specialize on it. We check trait is always applicable in the same way as impls, except that step 4 is now "all predicates on `impl1` are always applicable". We require that `specialization` or `min_specialization` is enabled to implement these traits.

### rustc_unsafe_specialization_marker

There are also some specialization on traits with no methods, including the `FusedIterator` trait which is advertised as allowing optimizations. We allow marking marker traits with an unstable attribute that means we ignore them in point 3 of the checks above. This is unsound but we allow it in the short term because it can't cause use after frees with purely safe code in the same way as specializing on traits methods can. 

r? @nikomatsakis
cc #31844 #67194

PR #89357, opened 2021-09-29
[experiment] Make `Cell<T>::update` work for `T: Default | Copy`.

m-ou-se:
This uses some unstable trickery to implement `Cell::update` for types that are either Copy or Default or both, and use the `.get() + .set()` behaviour if possible (if it is Copy), or use `.take() + .set()` otherwise (if it is Default and not Copy).

See https://github.com/rust-lang/rust/issues/50186#issuecomment-930127090

image

### DIFF
--- a/library/core/src/cell.rs
+++ b/library/core/src/cell.rs
@@ -435,8 +435,15 @@ impl<T: Copy> Cell<T> {
         // but `Cell` is `!Sync` so this won't happen.
         unsafe { *self.value.get() }
     }
+}
 
-    /// Updates the contained value using a function and returns the new value.
+impl<T> Cell<T> {
+    /// Updates the contained value using a function.
+    ///
+    /// If `T` implements [`Copy`], the function is called on a copy of the
+    /// contain value. Otherwise, if `T` implements [`Default`], the value
+    /// contained in the cell is temporarily replaced by [`Default::default()]`
+    /// while the function runs.
     ///
     /// # Examples
     ///
@@ -445,22 +452,78 @@ impl<T: Copy> Cell<T> {
     ///
     /// use std::cell::Cell;
     ///
-    /// let c = Cell::new(5);
-    /// let new = c.update(|x| x + 1);
+    /// let a = Cell::new(5);
+    /// a.update(|x| x + 1);
     ///
-    /// assert_eq!(new, 6);
-    /// assert_eq!(c.get(), 6);
+    /// let b = Cell::new("Hello".to_string());
+    /// b.update(|x| x + " World!");
+    ///
+    /// assert_eq!(a.get(), 6);
+    /// assert_eq!(b.take(), "Hello World!");
     /// ```
     #[inline]
     #[unstable(feature = "cell_update", issue = "50186")]
-    pub fn update<F>(&self, f: F) -> T
+    pub fn update<F>(&self, f: F)
     where
         F: FnOnce(T) -> T,
+        T: get_or_take::CopyOrDefault,
     {
-        let old = self.get();
+        let old = <T as get_or_take::CopySpec>::get_or_take(self);
         let new = f(old);
         self.set(new);
-        new
+    }
+}
+
+#[unstable(feature = "cell_update_specializations", issue = "none")]
+mod get_or_take {
+    use super::Cell;
+
+    #[rustc_on_unimplemented(
+        label = "`{Self}` implements neither Copy nor Default",
+        message = "`Cell<{Self}>::update()` requires either `{Self}: Copy` or `{Self}: Default`"
+    )]
+    #[marker]
+    pub trait CopyOrDefault: Sized {}
+
+    impl<T: Copy> CopyOrDefault for T {}
+    impl<T: Default> CopyOrDefault for T {}
+
+    #[rustc_unsafe_specialization_marker]
+    trait IsDefault: Default {}
+    impl<T: Default> IsDefault for T {}
+
+    pub(super) trait CopySpec {
+        fn get_or_take(cell: &Cell<Self>) -> Self;
+    }
+
+    trait DefaultSpec {
+        fn get_or_take(cell: &Cell<Self>) -> Self;
+    }
+
+    impl<T> CopySpec for T {
+        default fn get_or_take(cell: &Cell<Self>) -> Self {
+            <T as DefaultSpec>::get_or_take(cell)
+        }
+    }
+
+    impl<T: Copy> CopySpec for T {
+        fn get_or_take(cell: &Cell<Self>) -> Self {
+            cell.get()
+        }
+    }
+
+    impl<T> DefaultSpec for T {
+        default fn get_or_take(_: &Cell<Self>) -> Self {
+            // This is unreachable because we'd only get here for types that
+            // implement CopyOrDefault, but not Copy or Default, which isn't possible.
+            unreachable!()
+        }
+    }
+
+    impl<T: IsDefault> DefaultSpec for T {
+        fn get_or_take(cell: &Cell<Self>) -> Self {
+            cell.take()
+        }
     }
 }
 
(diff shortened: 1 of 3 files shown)


library/core/src/cell.rs
```diff
@@ -445,22 +452,78 @@ impl<T: Copy> Cell<T> {
...
+        label = "`{Self}` implements neither Copy nor Default",
+        message = "`Cell<{Self}>::update()` requires either `{Self}: Copy` or `{Self}: Default`"
+    )]
+    #[marker]
+    pub trait CopyOrDefault: Sized {}
+
+    impl<T: Copy> CopyOrDefault for T {}
+    impl<T: Default> CopyOrDefault for T {}
+
+    #[rustc_unsafe_specialization_marker]
```
m-ou-se, line 491: Who would be the right person to evaluate soundness of hacks like this?

@matthewjasper maybe?

library/core/src/cell.rs
```diff
@@ -435,8 +435,15 @@ impl<T: Copy> Cell<T> {
...
         unsafe { *self.value.get() }
     }
+}
 
-    /// Updates the contained value using a function and returns the new value.
+impl<T> Cell<T> {
+    /// Updates the contained value using a function.
+    ///
+    /// If `T` implements [`Copy`], the function is called on a copy of the
+    /// contain value. Otherwise, if `T` implements [`Default`], the value
```
fee1-dead, line 444:
```suggestion
    /// contained value. Otherwise, if `T` implements [`Default`], the value
```

library/core/src/cell.rs
```diff
@@ -435,8 +435,15 @@ impl<T: Copy> Cell<T> {
...
+    /// contained in the cell is temporarily replaced by [`Default::default()]`
```
fee1-dead, line 445:
```suggestion
    /// contained in the cell is temporarily replaced by [`Default::default()`]
```

library/core/src/cell.rs
eddyb: My notes so far: thanks to `CopyOrDefault`, the only way `Cell::update` could be unsound is if *both* traits are implemented, but with *different* imposed relationships on lifetimes.

However, that means `Copy` is used, and `Default` isn't called. So this is just as unsound as any other specialization on `Copy`.

IOW, when `Default::default` gets called, `T: CopyOrDefault` is equivalent to `T: Default` and any requirements from the `Default` impl will be imposed on the caller of `Cell::update`.

library/core/src/cell.rs
m-ou-se: Thanks!

library/core/src/cell.rs
danielhenrymantilla: Yes, this is right: `IsDefault` here is used to specialize over *nothing*.

In that regard, I find that writing the overriding impl before the default one, and slightly changing the naming of the things involved, can improve the readability of the code, as I've showcased [in this Playground](https://play.rust-lang.org/?version=nightly&mode=debug&edition=2018&gist=6c0c94312d80bfadc2569e9b3ebf3785) <sub><sup>(I've also nested some of the trait definitions and impls, which as a bonus scopes the `unsafe_specialization_marker`, since I can fathom the branches better that way, but I understand that that's debatable. I stand by the order of my impls with the `/* + !Trait */` annotations, though)</sup></sub>

comex: It may be sound in this particular case, but isn't this basically a… I guess you could say soundness hole, in `rustc_unsafe_specialization_marker`?  To quote [the PR that created the attribute](https://github.com/rust-lang/rust/pull/68970#issue-562084481):

>  This is unsound but we allow it in the short term because it can't cause use after frees with purely safe code in the same way as specializing on traits methods can.

So it's not completely sound, but it's in some sense 'less unsound' than unrestricted specialization because you can't use it to call methods.

Except you *are* using it to call methods, by taking advantage of the supertrait `Default`, from a specialized impl (`impl<T: IsDefault> DefaultSpec for T`) whose parent impl does not have a `Default` bound.

With that ability, how is this any different from unrestricted specialization?  In other words, what dangerous thing could you do by specializing on `Default` (which is not allowed) that you cannot do by creating a wrapper trait like this `IsDefault` and specializing on that?  It seems like the answer is 'nothing', in which case the idea of `rustc_unsafe_specialization_marker` being limited to marker traits that have no methods is meaningless.  But I may be missing something.

I note that in the case of some existing `rustc_unsafe_specialization_marker` traits, such as `FusedIterator` and `TrustedLen`, they do have supertraits – in both cases, `Iterator` – but impls that specialize on them always(?) have parent impls that are already bounded on `Iterator`.  On the other hand, another such trait, `MarkerEq`, as used in `RcEqIdent` and `ArcEqIdent`, is the same kind of wrapper as `IsDefault`.  It even has a comment saying:
> Hack to allow specializing on `Eq` even though `Eq` has a method.

danielhenrymantilla: @comex I find that with the Playground I linked to, which is equivalent to @m-ou-se's, semantic-wise, but clearer, bounds-wise, the "parent impl" for the `IsDefault` specialization is `CopyOrDefault`, with the specialized `Copy` impls stripped. This means that we are only dealing with `Default` types in practice (modulo `Copy` specialization issues), so the specialization is effectively specializing over the whole subset / it is as if there effectively were a `Default` supertrait. That's why I consider this `IsDefault` specialization, _when contained to this very usage_, sound. But that does assume that the `Copy` specialization is fine to begin with, which it may technically not be.

  - So, if anything, the only _potential_  issue / area I'm not that confident about is whether a "flaw" with the `Copy` specialization could be "amplified" by allowing a different "flaw" with the `IsDefault` specialization.

  - And even then, I could only see that causing "unsoundness" if we were to _guarantee_, at the API level, that for `Copy + Default` types `update()` shall necessarily `Copy` (because a hypothetical failed application of `Copy` specialization could break that guarantee). That's why, unless this situation is cleared up of all suspicion, I think we could loosen that guarantee to something like `mem::needs_drop()` and other such stuff:

      - For `Copy + Default` types, the "using `get()`" optimization is "likely" to be favored, but it is not guaranteed to do so: `unsafe` code cannot rely on this for soundness.

matthewjasper: I think this might be a soundness hole in the current `min_specialization` impl that I didn't get around to fixing.

matthewjasper: I've opened #89413 for this

dtolnay (commented):
I am :+1: on this overlapping marker trait + specialization approach if it works and is sound.

I tried it though, and saw some confusing behavior. Any idea what the deal is here? This code compiles on current nightly, and I believe you're intending for the PR to be strictly more general than what is now in nightly.

```rust
#![feature(cell_update)]

use std::cell::Cell;

#[derive(Copy, Clone, Default)]
struct Thing<'a>(&'a str);

fn main() {
    let cell: Cell<Thing> = Cell::new(Thing(""));
    cell.update(|Thing(_)| -> Thing { panic!() });
}
```

```console
error[E0283]: type annotations needed for `Thing<'_>`
  --> src/main.rs:10:18
   |
10 |     cell.update(|Thing(_)| -> Thing { panic!() });
   |                  ^^^^^^^^ consider giving this closure parameter a type
   |
   = note: cannot satisfy `Thing<'_>: cell::get_or_take::CopyOrDefault`
```

m-ou-se:
I suppose the problem is that either `Copy` or `Default` could be implemented *only* for `'static`:

```rust
impl Copy for Thing<'static> {}
```
or
```rust
impl Default for Thing<'static> {
    fn default() -> Self {
        Self("shark")
    }
}
```

Overlapping trait implementations and lifetimes don't mix perfectly, it seems.

Dylan-DPC:
Closing this as it was an experiment

PR #89359, opened 2021-09-29
Various fixes for const_trait_impl

fee1-dead:
A few problems I found while making `Iterator` easier to const-implement.

1. More generous `~const Drop` check.

We check for nested fields with caller bounds.

For example, an ADT type with fields of types `A`, `B`, `C`, check if all of them are either:
 - Bounded (`A: ~const Drop`, `B: Copy`)
 - Known to be able to destruct at compile time (`C = i32`, `struct C(i32)`, `C = some_fn`)

2. Don't treat trait functions marked with `#[default_method_body_is_const]` as stable const fns when checking `const_for` and `const_try` feature gates.

I think anyone can review this, so no r? this time.

### DIFF
--- a/src/test/ui/rfc-2632-const-trait-impl/const-drop-bound.rs
+++ b/src/test/ui/rfc-2632-const-trait-impl/const-drop-bound.rs
@@ -1,0 +1,20 @@
+// check-pass
+
+#![feature(const_trait_impl)]
+#![feature(const_fn_trait_bound)]
+#![feature(const_precise_live_drops)]
+
+const fn foo<T, E>(res: Result<T, E>) -> Option<T> where E: ~const Drop {
+    match res {
+        Ok(t) => Some(t),
+        Err(_e) => None,
+    }
+}
+
+pub struct Foo<T>(T);
+
+const fn baz<T: ~const Drop, E: ~const Drop>(res: Result<Foo<T>, Foo<E>>) -> Option<Foo<T>> {
+    foo(res)
+}
+
+fn main() {}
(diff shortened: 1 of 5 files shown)


src/test/ui/rfc-2632-const-trait-impl/const-drop-bound.rs
```diff
@@ -0,0 +1,20 @@
...
+const fn foo<T, E>(res: Result<T, E>) -> Option<T> where E: ~const Drop {
+    match res {
+        Ok(t) => Some(t),
+        Err(_e) => None,
+    }
+}
+
+pub struct Foo<T>(T);
+
+const fn baz<T: ~const Drop, E: ~const Drop>(res: Result<Foo<T>, Foo<E>>) -> Option<Foo<T>> {
```
mbartlett21, line 16: So `~const Drop` bounds already are unlike `Drop` bounds in that the type does not have to actually implement the `Drop` trait. Is this noted anywhere?

fee1-dead: See https://internals.rust-lang.org/t/pre-rfc-revamped-const-trait-impl-aka-rfc-2632/15192#drop-5

mbartlett21: Ok. That makes sense now, seeing as there isn't really any other way to easily do it.

oli-obk:
makes sense

r=me after a rebase

PR #89360, opened 2021-09-29
rustdoc-json: Always encode with UTF-8 when opening json files

hkmatsumoto:
Noticed by https://github.com/rust-lang/rust/pull/88234#issuecomment-927304228

I doubt `check_missing_items.py` malfunctions when the index contains some letters like emojis. This PR is to check that.


hkmatsumoto:
I'll close this PR and open another, referring this as a proof.

PR #89048, opened 2021-09-17
Add more tracing instrumentation

oli-obk:
I changed or added all this while working on a branch and pulled it out so we can merge it on its own.

PR #89363, opened 2021-09-29
Fix performance regression with #[instrument]

oli-obk:
linked tracing PR: https://github.com/tokio-rs/tracing/pull/1600

regression introduced by #89048

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5345,9 +5345,9 @@ checksum = "360dfd1d6d30e05fda32ace2c8c70e9c0a9da713275777f5a4dbb8a1893930c6"
 
 [[package]]
 name = "tracing"
-version = "0.1.25"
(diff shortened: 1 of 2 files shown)


oli-obk:
r? rust-lang/compiler

joshtriplett:
Reviewing the results, this looks like a substantial improvement.

Mark-Simulacrum:
I do note that tracing-subscriber now takes ~1.8x as long to compile -- some of that might be noise, but maybe worth mentioning upstream.

The minor performance regressions on a number of benchmarks do seem to be outweighed by the far larger and more consistent wins, though, so this seems good to merge to me in the meantime - we can always bump again if there's further improvements (e.g., to tracing-subscriber compile times).

@bors r+

PR #89366, opened 2021-09-29
rustdoc: Remove lazy_static dependency

GuillaumeGomez:
The macro was used in only one place and there are equivalents in the std, so it seemed weird to keep it around...

I think running a perf check would be a good idea though, just in case.

r? @jyn514


jyn514:
This part of the code isn't perf-sensitive, I think it's fine to just approve. Good catch, thanks!

@bors r+ rollup

Issue #89339, opened 2021-09-28
Windows LLVM Sanitizer Support


voteblake:
In an effort to fuzz a Windows-specific Rust program with `cargo-fuzz` I've done a bit of digging that leads me to think the compiler is the current blocker.

[`cargo-fuzz`](https://github.com/rust-fuzz/cargo-fuzz) mentions [`libFuzzer`](https://llvm.org/docs/LibFuzzer.html) "only works on x86-64 Linux and x86-64 macOS for now". The `libFuzzer` docs in turn suggest Windows should now be [support-able](https://llvm.org/docs/LibFuzzer.html#q-does-libfuzzer-support-windows), and the LLVM [AddressSanitizer](https://clang.llvm.org/docs/AddressSanitizer.html) documentation appears to agree.

Looking at the history of sanitizer discussion here I see Windows support being mentioned as possible as early as 2017 https://github.com/rust-lang/rust/issues/39699#issuecomment-297589826. Windows support was also mentioned in https://github.com/rust-lang/rust/issues/47174 as part of the discussion on stabilizing the -Z sanitize options.

I'm opening this issue to document the data I've gathered, ask for input on if I'm understanding the current state correctly, and also understand if anyone is working on this. I may be able to help, but am not an experienced contributor to the compiler. I've started to work locally to understand what this change would look like if it is in-fact possible.

voteblake:
Hi @danielframpton - thanks for tagging this in your PR yesterday, it was great to see that. Very timely! I wanted to invite you to use this issue to track any other outstanding work you're aware of to enable sanitizer support on Windows. If you have any areas you're looking for help or input on I can't guarantee I'd be useful but I'd be happy to try.

workingjubilee:
Yes, this seems to be an accurate (partial) description. I am guessing you are mostly speaking about the the MSVC target as the linked document references `link.exe` arguments?

voteblake:
All the digging I've done is MSVC specific, but the change in #89369 applies anywhere [`is_like_windows`](https://github.com/rust-lang/rust/blob/1d71ba862309d59df710078a845c8772ffb22aba/compiler/rustc_target/src/spec/mod.rs#L1155) not just [`is_like_msvc`](https://github.com/rust-lang/rust/blob/1d71ba862309d59df710078a845c8772ffb22aba/compiler/rustc_target/src/spec/mod.rs#L1165).

Issue #47174, opened 2018-01-04
Stabilize -Z sanitize


Firstyear:
Hi,

I would really like to see -Z sanitize options available in stable rust. 

Today they are available in nightly. I have used them for 6 months in the 389 Directory Server project in a production ldap environment as part of the code IO event path. I have exercise the leak and address sanitize options extensively, used them to find many FFI safety issues, and tested multiple build types (exec, dylib, more).

The sanitizer support is very important for FFI assertions during development where a code base uses Rust and C in some form. It allows strict analysis of behaviour, and finding where items are not droped for example.

An important reason for inclusion in stable is *distributions*. As a project, we often have the requirement to build debug builds with the compiler that ships in a distribution to reproduce problems. That means rustc stable that ships in products like fedora or RHEL. 

If -Z sanitize isn't in stable, we can't achieve this - and this adds a barrier to adoption.

Many people have put a lot of time into sanitizer support, and today it's locked behind nightly. :( 

I would love to see this in stable as it brings great run-time analysis features to rust especially for FFI use cases, and having paths to stabilise debugging options into stable rustc will strongly aid adoptions of rust in distributions of linux. It's an advertised feature of the language on email lists, so to have this accessible to anyone would be a huge benefit.

I'm happy to do the coding to enable this, as well as documentation and examples, but I need some help and guidance to achieve this,

Thanks, 

See also:

https://github.com/rust-lang/rfcs/issues/670
https://github.com/rust-lang/rust/pull/42711
https://users.rust-lang.org/t/howto-sanitize-your-rust-code/9378

cuviper:
Sorry this has been neglected -- let's try to flag a team:
cc @rust-lang/compiler

nikomatsakis:
@Firstyear you make a strong case for stable sanitizers. However, I don't really feel comfortable stabilizing `-Zsanitize` as it stands now, because -- to be quite honest -- I have no idea what it does! It was merged by @japaric in https://github.com/rust-lang/rust/pull/38699 on an explicitly experimental basis. At the time, [I wrote this](https://github.com/rust-lang/rust/pull/38699#issuecomment-271710464):

> I have to say that while I support the aims of this design, I would like to see some sort of process around these kinds of ad-hoc additions, even though they are unstable. At minimum some kind of decision to experiment in the area (with the goal of an eventual RFC). (Or do we just never expect this process to be stable? Which seems...plausible.)

[...]

PR #89369, opened 2021-09-29
Allow use of AddressSanitizer on Windows by linking to existing libraries

danielframpton:
This change makes it possible to use address sanitizer on Windows by determining the right library name to use based on the way that the crt is being linked.

Currently, this change does not build the address sanitizer libraries (as we do for other platforms) so it is necessary to provide the libraries either from building those libraries from llvm or using other libraries from your C++ toolchain.

The current support for AddressSanitizer outside of windows involves renaming things compiled out of the llvm fork. I don't quite follow how this will work when sanitizing hybrid C/C++/Rust programs, and the approach causes additional problems on Windows because the renamed import library still references the dll with the original name.

I expect some of these questions will need to be resolved as part of stabilization (see #47174) but I wanted to make some progress on the Windows support as mentioned there and in other issues like #89339 so that we can see how this will all fit together.

### DIFF
--- a/compiler/rustc_codegen_ssa/src/back/link.rs
+++ b/compiler/rustc_codegen_ssa/src/back/link.rs
@@ -1070,9 +1070,12 @@ fn add_sanitizer_libraries(sess: &Session, crate_type: CrateType, linker: &mut d
     // both executables and dynamic shared objects. Everywhere else the runtimes
     // are currently distributed as static liraries which should be linked to
     // executables only.
+    // On Windows there are both static and dynamic runtimes available.
     let needs_runtime = match crate_type {
         CrateType::Executable => true,
-        CrateType::Dylib | CrateType::Cdylib | CrateType::ProcMacro => sess.target.is_like_osx,
+        CrateType::Dylib | CrateType::Cdylib | CrateType::ProcMacro => {
+            sess.target.is_like_osx || sess.target.is_like_windows
+        }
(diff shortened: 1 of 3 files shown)


apiraino:
Reporting some sparse notes from the T-compiler [discussion on Zulip](https://rust-lang.zulipchat.com/#narrow/stream/238009-t-compiler.2Fmeetings/topic/.5Bweekly.5D.202021-10-28.20.2354818/near/259397383) of a while ago.

Here a Windows / ASAN knowledgeable person would help with the review, possibly opening an MCP or similar. The general consensus was that the Windows / Microsoft people have better context here, perhaps [drafting a writeup of the tradeoffs](https://rust-lang.zulipchat.com/#narrow/stream/238009-t-compiler.2Fmeetings/topic/.5Bweekly.5D.202021-10-28.20.2354818/near/259398638) in the tracking issue.

Dylan-DPC:
@danielframpton any updates on this?

danielframpton:
I don't have an update, but I may get some more cycles to look at this soon. If anyone gets to it first, then I would be happy to discuss any details of this change or what we need to have things working on Windows.

I think in addition to this change we may want to do the work to also build the runtimes for Windows (with correct dll names) and also make it possible to configure an override for the runtime libraries to use (to support hybrid Rust and C/C++ scenarios).

JohnCSimon:
@danielframpton 

@rustbot label: +S-inactive

PR #89370, opened 2021-09-29
CTFE: tweak aggregate rvalue handling

RalfJung:
I have not looked at this code in ages... I think Miri does not even hit it, since (most?) aggregate rvalues are lowered somewhere in the MIR pipeline, but CTFE does hit it.

So this adds some extra sanity assertions, and removes a ZST special case -- ZST should only be special cased fairly late (when the actual memory access happens); e.g. `!` is a ZST and we still want `copy_op` to be called for it since it will perform validation (and raise UB, since `!` is never valid).


RalfJung:
r? @oli-obk 

Also making sure the extra assertions do not regress performance... this is not the array case though so I think we are good.
@bors try @rust-timer queue

PR #89375, opened 2021-09-29
Fix #74256

### LINKED ISSUE #74256: E0623 highlights wrong parameter in async fn
```rust
struct Struct { }

impl Struct {
    async fn async_fn(self: &Struct, f: &u32) -> &u32 {
        f
    }
    
    fn sync_fn(self: &Struct, f: &u32) -> &u32 {
        f
    }
}
```

Output ([playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=c4cfd4b3a84d86ea5e43494b8c11aa0f))

```
error[E0623]: lifetime mismatch
 --> src/main.rs:5:9
  |
4 |     async fn async_fn(self: &Struct, f: &u32) -> &u32 {
  |                             -------              ----
  |                             |
  |                             this parameter and the return type are declared with different lifetimes...
5 |         f
  |         ^ ...but data from `f` is returned here

error[E0623]: lifetime mismatch
 --> src/main.rs:9:9
  |
8 |     fn sync_fn(self: &Struct, f: &u32) -> &u32 {
  |                                  ----     ----
  |                                  |
  |                                  this parameter and the return type are declared with different lifetimes...
9 |         f
  |         ^ ...but data from `f` is returned here

For more information about this error, try `rustc --explain E0623`.
error: could not compile `playground`.

```

The error in `sync_fn` is correct, but the one for `async_fn` incorrectly claims the return type lifetime is different from `self`'s lifetime, when they are in fact the same per the elision rules.

eholk:


nagisa:
I understand this is currently a draft, but please adjust the title and PR description with something more meaningful.

eholk:
Oh sorry, I didn't expect rust-highfive to assign anyone while it's a draft! I put this up to make it easy to share and get some early feedback, but I'll definitely have a better title and description before I'm ready to have it reviewed.

PR #89380, opened 2021-09-29
Fix Windows LLVM issue.

ehuss:
GitHub image 20210928.2 added LLVM 12.0.1 to the stock image.  However, the `lldb` executable doesn't work, it fails with:

> C:/Program Files/LLVM/bin/lldb.exe: error while loading shared libraries: ?: cannot open shared object file: No such file or directory

We probably don't want to start testing LLDB on windows anyways (at least not without intent).

The hacky solution for now is to just delete the system LLVM.


ehuss:
Some alternate solutions could be:
* Remove `C:\Program Files\LLVM\bin` from PATH.  I don't know how to do that in a safe way in GitHub Actions.
* Add an option to rustbuild to explicitly ignore lldb.
* Change rustbuild to behave as-if lldb isn't installed if `lldb -P` fails.

Mark-Simulacrum:
I think this is the right solution for now.

@bors r+ p=101

PR #89385, opened 2021-09-30
rustdoc: use flexbox to layout sidebar and main content

cynecx:
The way rustdoc layouts the sidebar and main content is quite hacky because their content boxes are basically overlapping which eg. causes [subpixel anti-aliasing issues on Firefox](https://bugzilla.mozilla.org/show_bug.cgi?id=1722163) (with WebRender).

This PR utilizes flexbox and fixes most of the cosmetic issues affecting it (the layout and cosmetics are basically unchanged).

### DIFF
--- a/src/librustdoc/html/static/css/rustdoc.css
+++ b/src/librustdoc/html/static/css/rustdoc.css
@@ -1740,11 +1740,11 @@ details.undocumented[open] > summary::before {
 
 	.rustdoc:not(.source) .sidebar .logo-container {
 		width: 35px;
-	    height: 35px;
-	    margin-top: 5px;
-	    margin-bottom: 5px;
-	    float: left;
-	    margin-left: 50px;
+		height: 35px;
+		margin-top: 5px;
+		margin-bottom: 5px;
+		float: left;
+		margin-left: 50px;
 	}
 
 	.sidebar .logo-container > img {
--- a/src/librustdoc/html/static/css/rustdoc.css
+++ b/src/librustdoc/html/static/css/rustdoc.css
@@ -315,7 +315,7 @@ nav.sub {
 }
 
 .source .sidebar {
-	width: unset;
+	width: 0px;
 	min-width: 0px;
 	max-width: 300px;
 	flex-grow: 0;
--- a/src/librustdoc/html/static/css/rustdoc.css
+++ b/src/librustdoc/html/static/css/rustdoc.css
@@ -317,7 +317,7 @@ nav.sub {
 }
 
 .source .sidebar {
-	width: 0px;
+	width: 50px;
 	min-width: 0px;
 	max-width: 300px;
 	flex-grow: 0;
@@ -326,6 +326,24 @@ nav.sub {
 	border-right: 1px solid;
 	transition: width .5s;
 	overflow-x: hidden;
+	/* The sidebar is by default hidden  */
+	overflow-y: hidden;
+}
+
+.source .sidebar > *:not(:first-child) {
+	transition: opacity 0.5s, visibility 0.2s;
+	opacity: 0;
+	visibility: hidden;
+}
+
+.source .sidebar.expanded {
+	overflow-y: auto;
+	width: 300px !important;
+}
+
+.source .sidebar.expanded > * {
+	opacity: 1;
+	visibility: visible;
 }
 
 /* Improve the scrollbar display on firefox */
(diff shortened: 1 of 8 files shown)


GuillaumeGomez:
The source code pages are broken, however it seems like a good idea overall. I'll cc @jsha in here in case I missed something obvious accessibility-wise.

jsha (commented):
I'm excited about this change. Thanks for working on it. One thing I noticed recently looking at source view: there's actually a regular sidebar (hidden), plus a source view sidebar at a different level of the DOM. I think it would be nice to move the source view sidebar up to the same level as the regular sidebar and use the same CSS if we can.

cynecx:
The source page fix was a tricky/hacky one because now the sidebar is basically hidden but the sidebar also has the logo but on the source page we'd probably want to keep the logo, so I've workaround this by duplicating the logo node inside content and showing it only on source pages.

cynecx:
> I think it would be nice to move the source view sidebar up to the same level as the regular sidebar and use the same CSS if we can.

I am not sure if this is a good thing because the source side bar is toggable and overlays the main content. It's basically a "separate" thing, since it is excluded from the main layout (because of `position: fixed`). It's just easier that way.

cynecx:
Future work: Rework the search form. It's really hard to work with. We could use `display: grid` here which would simplify the css/html complexity greatly.

GuillaumeGomez:
> Future work: Rework the search form. It's really hard to work with. We could use `display: grid` here which would simplify the css/html complexity greatly.

We recently encountered huge issues with `display: grid` so please don't use it.

> > the source side bar is toggable and overlays the main content
> 
> FYI in #88301 we concluded we should change that so the source side bar does not overlay the main content.

Another idea would be to display the source sidebar all the time (not on mobile) and expand it when clicking on a button. You can see what I mean on docs.rs: https://docs.rs/crate/ftml/1.12.3/source/src/lib.rs (I'm quite proud of the usage of `display: sticky` :D ).

cynecx:
> FYI in #88301 we concluded we should change that so the source side bar does not overlay the main content.

That simplifies things.

> We recently encountered huge issues with display: grid so please don't use it.

What kind of issues if I may ask?

> Another idea would be to display the source sidebar all the time (not on mobile) and expand it when clicking on a button. You can see what I mean on docs.rs: https://docs.rs/crate/ftml/1.12.3/source/src/lib.rs (I'm quite proud of the usage of display: sticky :D ).

I've merged the source level sidebar into the actual page sidebar. It's shown by default but can be collapsed. On mobile its hidden by default but can be shown by clicking the toggle.

GuillaumeGomez:
> What kind of issues if I may ask?

This one: https://github.com/rust-lang/rust/issues/88545

> I've merged the source level sidebar into the actual page sidebar. It's shown by default but can be collapsed. On mobile its hidden by default but can be shown by clicking the toggle.

Can you make it collapsed by default please?

cynecx:
> Can you make it collapsed by default please?

Oh I misread. I thought the consensus was to make it visible by default.

GuillaumeGomez:
Well, the source sidebar is always visible, but collapsed by default.

cynecx:
> This one: #88545

This issue doesn't seem to affect layouts with a low count of rows or column. Imho, it doesn't justify a strict hands-off of `display: grid`, since the benefits of having simpler/lower complexity in css outweighs the potential issues (I was specifically talking about using `display: grid` for the search form which has like 1-2 rows and 4-5 columns, which shouldn't trigger the bug you linked).

cynecx:
`src/test/rustdoc-gui/search-filter.goml` seems to fail for whatever reason...

GuillaumeGomez:
You can move a DOM element using JS (since the sidebar is only available with JS in any case...).

cynecx:
@GuillaumeGomez Sorry, I am still caught up in something else :/ I might be able to address your comments at the weekend.

GuillaumeGomez:
No problem! Waiting for the next update then. :)

GuillaumeGomez:
@cynecx Any news on this? If you don't have time, I'll use your commits and finish it if you don't mind.

cynecx:
@GuillaumeGomez Sorry, but I don't think I'll get to this within the next 2 weeks.

> If you don't have time, I'll use your commits and finish it if you don't mind.

I don't mind at all. That would be great :)

GuillaumeGomez:
Closing it in favour of https://github.com/rust-lang/rust/pull/91356. Thanks again for your work, I had very little to do in the end. :laughing:

PR #89391, opened 2021-09-30
Add more span suggestions

Milo123459:
Hey, first PR. This just adds some new span_suggestion's to some errors.


Milo123459:
I was going to attempt to add something to the `async closures are unstable` error, but I'm not sure how to add another help section.

nagisa:
r? @estebank 

From my perspective a more descriptive PR title and description would be helfpul. Something like "Implement machine applicable suggestions for static closures and loop control flow constructs" for the title and describe why these are okay as machine-applicable hints in these situations.

src/test/ui/error-codes/E0697.stderr
```diff
@@ -0,0 +1,9 @@
+error[E0697]: closures cannot be static
+  --> $DIR/E0697.rs:2:5
+   |
+LL |     static || {};
+   |     ^^^^^^^^^ help: consider removing the: `static`
```
estebank, line 5: This suggestion won't apply correctly, because the resulting file will be `static {};`, which is not what we want. Also, I would change the wording to be more prescriptive: removing `static` isn't optional here, you just can't do that.

src/test/ui/error-codes/E0268.stderr
```diff
@@ -2,7 +2,10 @@ error[E0268]: `break` outside of a loop
   --> $DIR/E0268.rs:2:5
    |
 LL |     break;
-   |     ^^^^^ cannot `break` outside of a loop
+   |     ^^^^^
+   |     |
+   |     cannot `break` outside of a loop
+   |     help: consider removing the: `break`
```
estebank, line 8: We have a similar issue here, we'd want to include the trailing `;` in the removal suggestion (although a bare `;` isn't the end of the world) as well *and* the result of applying this change is no change (we are replacing `break` with `break`).

src/test/ui/error-codes/E0025.stderr
```diff
@@ -2,8 +2,10 @@ error[E0025]: field `a` bound multiple times in the pattern
   --> $DIR/E0025.rs:8:21
    |
 LL |     let Foo { a: x, a: y, b: 0 } = x;
-   |               ----  ^^^^ multiple uses of `a` in pattern
-   |               |
+   |               ----  ^^^^
+   |               |     |
+   |               |     multiple uses of `a` in pattern
+   |               |     help: consider removing one usage of: ``a``
```
estebank, line 8: Similar issue here, the result is

```rust
let Foo { a: x, `a`, b: 0 } = x;
```

We also can't know for sure which of the two usages was correct, so at least this would have to be `Applicability::MaybeIncorrect`.

compiler/rustc_ast_lowering/src/expr.rs
```diff
@@ -809,6 +809,12 @@ impl<'hir> LoweringContext<'_, 'hir> {
             None => {
                 if movability == Movability::Static {
                     struct_span_err!(self.sess, fn_decl_span, E0697, "closures cannot be static")
+                        .span_suggestion(
+                            fn_decl_span,
+                            "consider removing the",
+                            "static".into(),
+                            rustc_errors::Applicability::MachineApplicable,
+                        )
```
estebank, line 817:
```suggestion
                        .span_suggestion(
                            fn_decl_span.with_lo(fn_decl_span.lo() + BytePos("static ".len())),
                            "remove the `static` keyword to define a regular closure",
                            "".into(),
                            rustc_errors::Applicability::MachineApplicable,
                        )
```

compiler/rustc_passes/src/loops.rs
```diff
@@ -233,6 +233,12 @@ impl<'a, 'hir> CheckLoopVisitor<'a, 'hir> {
             Normal | AnonConst => {
                 struct_span_err!(self.sess, span, E0268, "`{}` outside of a loop", name)
                     .span_label(span, format!("cannot `{}` outside of a loop", name))
+                    .span_suggestion(
+                        span,
+                        "consider removing the",
+                        name.into(),
+                        rustc_errors::Applicability::MachineApplicable,
+                    )
```
estebank, line 241:
```suggestion
                    .span_suggestion(
                        span,
                        &format!("consider removing the `{}`", name),
                        String::new(),
                        Applicability::MaybeIncorrect,
                    )
```

Please add a `use` at the top of the file for `rustc_errors::Applicability` too.

----

In an ideal world we would also look for potential places where the user wanted to iterate, but this is fine for now.

compiler/rustc_typeck/src/check/pat.rs
```diff
@@ -1404,6 +1404,7 @@ impl<'a, 'tcx> FnCtxt<'a, 'tcx> {
         )
         .span_label(span, format!("multiple uses of `{}` in pattern", ident))
         .span_label(other_field, format!("first use of `{}`", ident))
+        .span_suggestion(span, "consider removing one usage of", ident.to_string(), rustc_errors::Applicability::MachineApplicable)
```
estebank, line 1407: I would probably drop this one: we have no way of figuring out which of the uses was likely intended.

estebank (commented):
Thanks for doing this! I left some inline comments asking for changes.

estebank:
Also, you'll need to run `./x.py fmt` before submitting, otherwise CI will complain and won't let us merge.

src/test/ui/error-codes/E0697.stderr
Milo123459: Oh, thought you could. Should I just remove this change?

src/test/ui/error-codes/E0268.stderr
Milo123459: So I need to implement a way to check if there is a ; and then display that. Got it.

src/test/ui/error-codes/E0697.stderr
estebank: You can change this to be a `span_label` instead.

src/test/ui/error-codes/E0268.stderr
estebank: Not display it, but extend the span to cover it. Thinking about it a bare `;` will be follow up error, so this suggestion isn't _terrible_, you just have to change the code to actually remove `break` instead of replacing it with itself.

src/test/ui/error-codes/E0697.stderr
estebank: Wait, your code changes do what we want. Disregard.

Milo123459: Oh

Milo123459:
Kinda confused as I added the use in and it still fails?

jyn514:
@Milo123459 the error is in compiler/rustc_ast_lowering/src/expr.rs, but you only added the import to compiler/rustc_typeck/src/check/pat.rs. I recommend running `x.py check` locally, it will give much faster feedback and also the errors will be easier to read.

nagisa:
Please undo the changes to `Cargo.lock`. Updates to the dependency versions should only occur as a separate and well informed decisions.

Milo123459:
Hm.. running `git rebase master` has done that..

Milo123459:
> Please undo the changes to `Cargo.lock`. Updates to the dependency versions should only occur as a separate and well informed decisions.

Yes, I have tried that with git rebase: can't figure out what's going on here, I'm questioning why this even changed as well

Milo123459:
I think something locally has broken my branch 🤔

PR #89404, opened 2021-09-30
Slightly optimize hash map stable hashing

Kobzol:
I was profiling some of the `rustc-perf` benchmarks locally and noticed that quite some time is spent inside the stable hash of hashmaps. I tried to use a `SmallVec` instead of a `Vec` there, which helped very slightly.

Then I tried to remove the sorting, which was a bottleneck, and replaced it with insertion into a binary heap. Locally, it yielded nice improvements in instruction counts and RSS in several benchmarks for incremental builds. The implementation could probably be much nicer and possibly extended to other stable hashes, but first I wanted to test the perf impact properly.

Can I ask someone to do a perf run? Thank you!

### DIFF
--- a/compiler/rustc_data_structures/src/stable_hasher.rs
+++ b/compiler/rustc_data_structures/src/stable_hasher.rs
@@ -552,7 +552,8 @@ pub fn hash_stable_hashmap<HCX, K, V, R, SK, F>(
     SK: HashStable<HCX> + Ord,
     F: Fn(&K, &HCX) -> SK,
 {
-    let mut entries: Vec<_> = map.iter().map(|(k, v)| (to_stable_hash_key(k, hcx), v)).collect();
+    let mut entries: SmallVec<[_; 3]> =
+        map.iter().map(|(k, v)| (to_stable_hash_key(k, hcx), v)).collect();
     entries.sort_unstable_by(|&(ref sk1, _), &(ref sk2, _)| sk1.cmp(sk2));
     entries.hash_stable(hcx, hasher);
 }


Kobzol:
Well, this is embarrassing :sweat_smile: I assumed that iterating a `BinaryHeap` would return the items in a sorted order (a bit of an API footgun, although I understand why it doesn't do that). And I was wondering why it was so fast.. now it probably won't be nearly as good, but let's see.

Kobzol:
@jackh726 Ok I'm not actually sure if that last failure is caused by the PR or if it's spurious?

leonardo-m:
>I assumed that iterating a BinaryHeap would return the items in a sorted order (a bit of an API footgun, although I understand why it doesn't do that).

No problem. Take a look at the "Heap Sort" algorithm :)

Kobzol:
> No problem. Take a look at the "Heap Sort" algorithm :)

I assumed that `iter()` would behave like [`into_iter_sorted`](https://doc.rust-lang.org/std/collections/struct.BinaryHeap.html#method.into_iter_sorted) (which I didn't know existed, otherwise it would be obvious). But of course it can't do that without consuming the heap, which I didn't realize.

Kobzol:
Sigh, back to the drawing board I guess. It would really help to make this sort faster, but my usual tricks (parallelizing the sort or using radix sort) probably can't be applied here. Let's see if the SmallVec alone helps with anything, if not, I'll close the PR.

Can someone please run another perf run? :) Thanks.

Kobzol:
It seems that the sort is a bigger problem than the allocation, and `SmallVec` won't help with that (it will probably make it worse because of constant spill checks). I tried sorting only the hash stable keys and then looking up the values to avoid shuffling too much data during the sort. But the values are usually not very large (~80 bytes at the tail end), so it doesn't help.

The results are probably too mixed to merge this, but I'll leave the decision up to @matthewjasper.

JohnCSimon:
Triage:
@Kobzol If this MR is ready for review can you post `@rustbot ready` to switch to `S-waiting-on-review` thank you.

Kobzol:
I think that it's ready to be merged (https://github.com/rust-lang/rust/pull/89404#issuecomment-950384956 last sentence), I just don't have rights to `r=`.

michaelwoerister:
I just saw this now. Very interesting! Note that we don't actually need to "sort" the hash table contents before hashing. We just need to have them in a deterministic order. Maybe that can be exploited somehow?

Kobzol:
Hmm, so, any idea how to produce a deterministic order without sorting? :D And without depending on the insertion order I suppose.

If all hash map insertions have the same order and the hash function is deterministic, we could probably just iterate the items of the hashmap as they are laid out in the map. But that would depend on the insertion order, which seems to be too brittle (basically this was the reason why we decided to postpone https://perf.rust-lang.org/compare.html?start=55ccbd090d96ec3bb28dbcb383e65bbfa3c293ff&end=7b55f66571824a0ba6e2ee9303d94854dcc8f378 until we make sure that the insertion order won't break anything).

the8472:
> Hmm, so, any idea how to produce a deterministic order without sorting?

An idea how to produce a stable hash without order: Use a commutative function.

The problem with that is that it requires the hash for each item to be computed separately before it can be merged with a commutative function. The initialization and `finish()` costs of `SipHasher` might be significant enough that the overhead of hashing elements individually might outweigh the gains from not having to sort. In that case some FastStableHasher would be needed.

Kobzol:
@the8472 from my benchmarks it looked like hashing and sorting takes a similar amount of time here, based on the workload. So keeping the hashing + a commutative function might theoretically work. I'll try it, thanks.

PR #89405, opened 2021-09-30
Fix clippy lints

GuillaumeGomez:
I'm currently working on allowing clippy to run on librustdoc after a discussion I had with @Mark-Simulacrum. So in the meantime, I fixed a few lints on the compiler crates.


GuillaumeGomez:
Seems spurious...

@bors retry

GuillaumeGomez:
Seems like there is a small perf regression, no?

Mark-Simulacrum:
Indeed, our heuristics did not deem this definite, but did consider it probable. [We're working on those](https://github.com/rust-lang/rustc-perf/issues/983).

Tagging as a regression. Likely to be painful to investigate, but we'll see.

PR #83343, opened 2021-03-21
Simplify and fix byte skipping in format! string parser

osa1:
Fixes '\\' handling in format strings.

Fixes #83340

PR #89408, opened 2021-09-30
Avoid nondeterminism in trimmed_def_paths

Mark-Simulacrum:
Previously this query depended on the global interning order of Symbols, which
meant that irrelevant changes could influence the query and cause
recompilations. This commit ensures that the return set is stable and will not
be affected by the global order by deterministically (in lexicographic order)
choosing a name to use if there are multiple names for a single DefId.

This should fix the cause of the [regressions] in #83343.

[regressions]: https://perf.rust-lang.org/compare.html?start=9620f3a84b079decfdc2e557be007580b097fe43&end=addb4da686a97da46159f0123cb6cdc2ce3d7fdb

### DIFF
--- a/compiler/rustc_middle/src/ty/print/pretty.rs
+++ b/compiler/rustc_middle/src/ty/print/pretty.rs
@@ -2408,7 +2408,7 @@ fn for_each_def(tcx: TyCtxt<'_>, mut collect_fn: impl for<'b> FnMut(&'b Ident, N
 ///
 /// The implementation uses similar import discovery logic to that of 'use' suggestions.
 fn trimmed_def_paths(tcx: TyCtxt<'_>, (): ()) -> FxHashMap<DefId, Symbol> {
-    let mut map = FxHashMap::default();
+    let mut map: FxHashMap<DefId, Symbol> = FxHashMap::default();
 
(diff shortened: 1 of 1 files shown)


Mark-Simulacrum:
(Nondeterminism is maybe not the right name -- a specific compilation should always be the same. But "irrelevant" changes can cause the behavior of this query to differ, which seems not great).

tgnottingham:
Seems like a good idea.

Outside the scope of this, but it might be nice if trimmed paths in error messages used the most appropriate name when there are multiple possibilities, taking into account any `use ... as ...` as applicable. That may get tricky if the error message involves multiple locations and they use different names for the same def id though.

rylev:
This indeed [seems to fix](https://perf.rust-lang.org/compare.html?start=6dc08b909b469d58dd8fa54c57ab193b8cf95257&end=bde948fea1e2e10b64d5edf9ca9c1aba01f2da10&calcNewSig=true) the performance regressions we saw [previously](https://perf.rust-lang.org/compare.html?start=9620f3a84b079decfdc2e557be007580b097fe43&end=addb4da686a97da46159f0123cb6cdc2ce3d7fdb)! 🎉

PR #82280, opened 2021-02-19
[WIP] Test performance of running MIR inliner on inline(always) function calls

wesleywiser:
r? @ghost

PR #89415, opened 2021-10-01
Correct caller/callsite confusion in inliner message

wesleywiser:
`callee_body` is the MIR `Body` for the `callsite.callee` so this message basically says `"Inline {bar span} into bar"` when it should say `"Inline bar into foo"`. 

Extracted out of #82280


michaelwoerister (approved):

michaelwoerister:
Looks good to me, thanks @wesleywiser!

@bors r+ rollup

PR #89419, opened 2021-10-01
Add bpf(el|eb)-unknown-none targets to dist-various-2

dave-tucker:
This commit adds the bpfel-unknown-none and bpfeb-unknown-none targets
to the dist-various-2. The motivation for this is to allow for libcore
to be added to a nightly toolchain using `rustup target add`. This will
allow us to perform testing of the bpf-linker against rust nightly.

See: https://github.com/aya-rs/bpf-linker/pull/4, which currently requires building a nightly toolchain from source for these tests, which takes ~50 minutes (x3 for testing against various LLVMs)

### DIFF
--- a/src/ci/docker/host-x86_64/dist-various-2/Dockerfile
+++ b/src/ci/docker/host-x86_64/dist-various-2/Dockerfile
@@ -100,6 +100,8 @@ ENV CARGO_TARGET_AARCH64_FUCHSIA_RUSTFLAGS \
 
 ENV TARGETS=x86_64-fuchsia
 ENV TARGETS=$TARGETS,aarch64-fuchsia
+ENV TARGETS=$TARGETS,bpfel-unknown-none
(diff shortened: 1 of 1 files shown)


cuviper:
I think this amounts to a Tier 2 promotion:
https://rust-lang.github.io/rfcs/2803-target-tier-policy.html#tier-2-target-policy

dave-tucker:
> I think this amounts to a Tier 2 promotion

Oh no! 😯 I don't think we're ready for Tier 2 just yet.

kennytm:
`dist-various-2` is part of the CI, so yes this PR effectively makes these targets T2 ("At this tier, the Rust project guarantees that a target builds, and will reject patches that fail to build on a target.").

dave-tucker:
Thanks for confirming. I'll go ahead and close this for now then and we'll come trigger the Tier 2 promotion when we're more ready.

PR #89422, opened 2021-10-01
Remove whitespaces in doctests' name

### LINKED ISSUE #88263: rustdoc does not filter correctly doctests when their name contain spaces.
Consider this file `main.rs`:

```rust
struct Iter<T,P>(T,P);

impl<T,P> Iter<T,P> {
    ///```
    ///assert!(true)
    ///```
    fn len() {}
}

struct Iter2<T,P>(T,P);

impl<T,P> Iter2<T,P> {
    ///```
    ///assert!(true)
    ///```
    fn len() {}
}

fn main() {
}
```
If one run `rustdoc main.rs --test --test-args 'Iter<T, P>::len'`, he will expect that only
the doctest of `Iter<T, P>::len` will be run. But the two doctests will be run.

Below is the output of `rustdoc main.rs --test --test-args 'Iter<T, P>::len'`
```
running 2 tests
test main.rs - Iter<T, P>::len (line 4) ... ok
test main.rs - Iter2<T, P>::len (line 13) ... ok

test result: ok. 2 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.12s
```

The problem seems to be the space: it looks like, in the example above, `rustdoc` will run all tests that matches 'P>::len' (in the argument `'Iter<T, P>::len'` there is a space between the `,` and the `P`).

### Meta
This bug is present since at least one year in all releases.

`rustc --version --verbose`:
```
rustc 1.56.0-nightly (a0035916e 2021-08-20)
binary: rustc
commit-hash: a0035916e01d8e644ccd44554c57f0874cef8c8c
commit-date: 2021-08-20
release: 1.56.0-nightly
LLVM version: 12.0.1
```

GuillaumeGomez:
Fixes #88263.

Instead of handling white spaces when we filter tests (which would be quite complicated since we split on them!), I propose to instead replace them with dashes.

So for example, this:

```console
test foo.rs - Iter2<T, P>::len (line 13) ... ok
test foo.rs - Iter<T, P>::len (line 4) ... ok
```

becomes:

```console
test foo.rs - Iter<T,-P>::len (line 4) ... ok
test foo.rs - Iter2<T,-P>::len (line 13) ... ok
```

r? @jyn514

### DIFF
--- a/src/librustdoc/doctest.rs
+++ b/src/librustdoc/doctest.rs
@@ -853,6 +853,7 @@ impl Collector {
 
     fn generate_name(&self, line: usize, filename: &FileName) -> String {
         let mut item_path = self.names.join("::");
+        item_path.retain(|c| c != ' ');
         if !item_path.is_empty() {
             item_path.push(' ');
         }


GuillaumeGomez:
Failed the assignment...

r? @jyn514

Urgau:
Is it not possible to use a [no-break space](https://graphemica.com/00A0) « » (ie U+00A0) instead of « - » ?
This would have the benefits of keeping the « same » visual appearance.

GuillaumeGomez:
If we do that, it's very likely that people will simply enter the name (with normal white space) and end up in the original issue.

jyn514:
r? @CraftSpider

I don't have time to review this.

CraftSpider:
I'm not sure the look with dashes is great, but I also agree that using no-break spaces is likely to lead people to the same issue.

Possible Alternative: How often do spaces show up, and where? If we're pretty sure spaces won't show up between identifiers, would it make sense to just strip them entirely? So we instead end up with `Iter<T,P>::len`, which while imperfect I think still looks a little better, and solves the issue.

GuillaumeGomez:
Great idea! I prefer this alternative much more! I'll update it then.

GuillaumeGomez:
Done!

CraftSpider:
Awesome, looks good to me
@bors r+

Issue #89374, opened 2021-09-29
Feature gate new `non_exhaustive_omitted_patterns` lint


camelid:
It will need to be feature-gated before it lands on stable in **1.57.0**.

cc https://github.com/rust-lang/rust/issues/89042#issuecomment-929464660
cc @Nadrieril 

This issue has been assigned to @DevinR528 via [this comment](https://github.com/rust-lang/rust/issues/89374#issuecomment-931657255).

Nadrieril:
By when should the PR be merged if we want to get it in before 1.57.0 roughly?

camelid:
I'm not 100% sure, but I think 1.57 will be promoted to beta in ~2 weeks. After that, the PR would need a backport to the beta branch in order to get it in before 1.57 is released as stable. (You can see the release dates [here](https://forge.rust-lang.org/#current-release-versions). Nightly branches off as beta around 7 weeks before it is released as stable IIRC.)

camelid:
From https://github.com/rust-lang/rust/issues/89042#issuecomment-931497696:

> I can work on adding the feature gate I'll open a PR and cc both of you (as long as that's ok?).

@rustbot assign @DevinR528

PR #89423, opened 2021-10-01
Fix ICE caused by non_exaustive_omitted_patterns struct lint

### LINKED ISSUE #89382: ICE related to non_exhaustive_omitted_patterns lint
Just tried to update my toolchain from `nightly-2021-09-04` to `nightly-2021-09-29`, and I now get an ICE in a project of mine when I run `cargo check`.

### Code

I don't have a self-contained reproduction (though I can try to put one together if necessary). I see that the error is originating from the `non_exhaustive_omitted_patterns` lint introduced in 33a06b73d90bde1d2fc4902672f4cbeb233b83e2 (cc @DevinR528), and the only pattern destructuring happening in the `convert_bulk_write_failure` function mentioned in the error is:

```Rust
let mongodb::error::BulkWriteFailure {
	write_concern_error,
	write_errors,
	..
} = error;
```
In the version of the mongodb crate I am using (2.0.0-beta.3), this struct is defined as:
```rust
pub struct BulkWriteFailure {
    /// The error(s) that occurred on account of a non write concern failure.
    pub write_errors: Option<Vec<BulkWriteError>>,

    /// The error that occurred on account of write concern failure.
    pub write_concern_error: Option<WriteConcernError>,

    #[serde(skip)]
    pub(crate) inserted_ids: HashMap<usize, Bson>,
}
```
This makes me wonder if the compiler is getting confused by the `pub(crate)` visibility of the last field?

### Meta

`rustc --version --verbose`:
```
rustc --version --verbose
rustc 1.57.0-nightly (8f8092cc3 2021-09-28)
binary: rustc
commit-hash: 8f8092cc32ec171becef8ceacec7dbb06c5d7d7e
commit-date: 2021-09-28
release: 1.57.0-nightly
LLVM version: 13.0.0
```

### Error output

```
error: internal compiler error: compiler/rustc_typeck/src/check/pat.rs:1638:23: impossible case reached

thread 'rustc' panicked at 'Box<dyn Any>', compiler/rustc_errors/src/lib.rs:1146:9

note: rustc 1.57.0-nightly (8f8092cc3 2021-09-28) running on x86_64-unknown-linux-gnu

note: compiler flags: -Z share-generics=y -C embed-bitcode=no -C debuginfo=2 -C linker=/usr/bin/clang -C incremental -C link-arg=-fuse-ld=lld --crate-type lib

query stack during panic:
#0 [typeck] type-checking `error_conversion::bulk_write_error::convert_bulk_write_failure`
#1 [typeck_item_bodies] type-checking all item bodies
end of query stack
```

<p>

```
error: internal compiler error: compiler/rustc_typeck/src/check/pat.rs:1638:23: impossible case reached

thread 'rustc' panicked at 'Box<dyn Any>', compiler/rustc_errors/src/lib.rs:1146:9
stack backtrace:
   0: std::panicking::begin_panic
   1: std::panic::panic_any
   2: rustc_errors::HandlerInner::bug
   3: rustc_errors::Handler::bug
   4: rustc_middle::ty::context::tls::with_opt
   5: rustc_middle::util::bug::opt_span_bug_fmt
   6: rustc_middle::util::bug::bug_fmt
   7: rustc_typeck::check::pat::<impl rustc_typeck::check::fn_ctxt::FnCtxt>::non_exhaustive_reachable_pattern::joined_uncovered_patterns
   8: rustc_typeck::check::pat::<impl rustc_typeck::check::fn_ctxt::FnCtxt>::check_struct_pat_fields
   9: rustc_typeck::check::pat::<impl rustc_typeck::check::fn_ctxt::FnCtxt>::check_pat
  10: rustc_typeck::check::fn_ctxt::checks::<impl rustc_typeck::check::fn_ctxt::FnCtxt>::check_decl_local
  11: rustc_typeck::check::fn_ctxt::checks::<impl rustc_typeck::check::fn_ctxt::FnCtxt>::check_stmt
  12: rustc_typeck::check::fn_ctxt::checks::<impl rustc_typeck::check::fn_ctxt::FnCtxt>::check_block_with_expected
  13: rustc_typeck::check::expr::<impl rustc_typeck::check::fn_ctxt::FnCtxt>::check_expr_kind
  14: rustc_typeck::check::expr::<impl rustc_typeck::check::fn_ctxt::FnCtxt>::check_expr_with_expectation
  15: rustc_typeck::check::check::check_fn
  16: rustc_infer::infer::InferCtxtBuilder::enter
  17: rustc_typeck::check::typeck
  18: rustc_query_system::dep_graph::graph::DepGraph<K>::with_task
  19: rustc_data_structures::stack::ensure_sufficient_stack
  20: rustc_query_system::query::plumbing::try_execute_query
  21: <rustc_query_impl::Queries as rustc_middle::ty::query::QueryEngine>::typeck
  22: rustc_middle::ty::<impl rustc_middle::ty::context::TyCtxt>::par_body_owners
  23: rustc_typeck::check::typeck_item_bodies
  24: rustc_query_system::dep_graph::graph::DepGraph<K>::with_task
  25: rustc_data_structures::stack::ensure_sufficient_stack
  26: rustc_query_system::query::plumbing::try_execute_query
  27: <rustc_query_impl::Queries as rustc_middle::ty::query::QueryEngine>::typeck_item_bodies
  28: rustc_session::utils::<impl rustc_session::session::Session>::time
  29: rustc_typeck::check_crate
  30: rustc_interface::passes::analysis
  31: rustc_query_system::dep_graph::graph::DepGraph<K>::with_task
  32: rustc_data_structures::stack::ensure_sufficient_stack
  33: rustc_query_system::query::plumbing::try_execute_query
  34: <rustc_query_impl::Queries as rustc_middle::ty::query::QueryEngine>::analysis
  35: rustc_interface::queries::<impl rustc_interface::interface::Compiler>::enter
  36: rustc_span::with_source_map
  37: rustc_interface::interface::create_compiler_and_run
```

</p>
</details>

DevinR528:
fixes #89382

Add check that a list of `Pat`s is non empty to prevent ICE in `FnCtxt::lint_non_exhaustive_omitted_patterns`.

Is related to #89374 and #89105

### DIFF
--- a/compiler/rustc_typeck/src/check/pat.rs
+++ b/compiler/rustc_typeck/src/check/pat.rs
@@ -181,8 +181,8 @@ impl<'a, 'tcx> FnCtxt<'a, 'tcx> {
                 self.check_pat_tuple_struct(pat, qpath, subpats, ddpos, expected, def_bm, ti)
             }
             PatKind::Path(_) => self.check_pat_path(pat, path_res.unwrap(), expected, ti),
-            PatKind::Struct(ref qpath, fields, etc) => {
-                self.check_pat_struct(pat, qpath, fields, etc, expected, def_bm, ti)
+            PatKind::Struct(ref qpath, fields, has_rest_pat) => {
+                self.check_pat_struct(pat, qpath, fields, has_rest_pat, expected, def_bm, ti)
             }
             PatKind::Or(pats) => {
                 let parent_pat = Some(pat);
@@ -685,7 +685,7 @@ impl<'a, 'tcx> FnCtxt<'a, 'tcx> {
         pat: &'tcx Pat<'tcx>,
         qpath: &hir::QPath<'_>,
         fields: &'tcx [hir::PatField<'tcx>],
-        etc: bool,
+        has_rest_pat: bool,
         expected: Ty<'tcx>,
         def_bm: BindingMode,
         ti: TopInfo<'tcx>,
@@ -707,7 +707,7 @@ impl<'a, 'tcx> FnCtxt<'a, 'tcx> {
         self.demand_eqtype_pat(pat.span, expected, pat_ty, ti);
 
         // Type-check subpatterns.
-        if self.check_struct_pat_fields(pat_ty, pat, variant, fields, etc, def_bm, ti) {
+        if self.check_struct_pat_fields(pat_ty, &pat, variant, fields, has_rest_pat, def_bm, ti) {
             pat_ty
         } else {
             self.tcx.ty_error()
@@ -1189,7 +1189,7 @@ impl<'a, 'tcx> FnCtxt<'a, 'tcx> {
         pat: &'tcx Pat<'tcx>,
         variant: &'tcx ty::VariantDef,
         fields: &'tcx [hir::PatField<'tcx>],
-        etc: bool,
+        has_rest_pat: bool,
         def_bm: BindingMode,
         ti: TopInfo<'tcx>,
     ) -> bool {
@@ -1263,7 +1263,7 @@ impl<'a, 'tcx> FnCtxt<'a, 'tcx> {
 
         // Require `..` if struct has non_exhaustive attribute.
         let non_exhaustive = variant.is_field_list_non_exhaustive() && !adt.did.is_local();
-        if non_exhaustive && !etc {
+        if non_exhaustive && !has_rest_pat {
             self.error_foreign_non_exhaustive_spat(pat, adt.variant_descr(), fields.is_empty());
         }
 
@@ -1275,7 +1275,7 @@ impl<'a, 'tcx> FnCtxt<'a, 'tcx> {
                     .struct_span_err(pat.span, "union patterns should have exactly one field")
                     .emit();
             }
-            if etc {
+            if has_rest_pat {
                 tcx.sess.struct_span_err(pat.span, "`..` cannot be used in union patterns").emit();
             }
         } else if !unmentioned_fields.is_empty() {
@@ -1287,7 +1287,7 @@ impl<'a, 'tcx> FnCtxt<'a, 'tcx> {
                 })
                 .collect();
 
-            if !etc {
+            if !has_rest_pat {
                 if accessible_unmentioned_fields.is_empty() {
                     unmentioned_err = Some(self.error_no_accessible_fields(pat, fields));
                 } else {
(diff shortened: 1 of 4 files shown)


compiler/rustc_typeck/src/check/pat.rs
```diff
@@ -1286,8 +1286,12 @@ impl<'a, 'tcx> FnCtxt<'a, 'tcx> {
                     field.vis.is_accessible_from(tcx.parent_module(pat.hir_id).to_def_id(), tcx)
                 })
                 .collect();
-            if non_exhaustive {
-                self.non_exhaustive_reachable_pattern(pat, &accessible_unmentioned_fields, adt_ty)
+            if non_exhaustive && !accessible_unmentioned_fields.is_empty() {
```
Nadrieril, line 1289: I think that would still cause errors. If I follow correctly in the following example we'd lint for `non_exhaustive_omitted_patterns` and then fail to emit the important "some fields are missing" error because there's an `else if`.
```rust
#[non_exhaustive]
struct Foo {
	a: bool,
	b: bool
}
fn do_something(foo: Foo) {
	let Foo { a } = foo;
}
```
So we probably want `if etc && non_exhaustive && !accessible_unmentioned_fields.is_empty()`

Nadrieril: Ah wait never mind, the lack of `, ..` would be caught earlier. Still, could you swap the two conditions, so the `!etc` test is done first? That'd be clearer in case things change later.

DevinR528: Ahh yeah, good point, I should probably also add tests then since it is more complicated than originally seemed.

DevinR528: How do you feel about calling `etc` -> `struct_wild` or `struct_rest` or `rest_pat` or `rest_wild`? `etc` is almost too short to be easily spotted :shrug: I didn't do it yet, I'm still kinda iffy about any of those being a better name.

Nadrieril: oh yea please do, took me a while to figure out what it was too. The rust book doesn't give a name to that syntax sadly. Maybe `has_etc_pat` or `has_rest_pat`?

Nadrieril:
LGTM!
r? @Nadrieril
@bors r+

PR #89425, opened 2021-10-01
Fix x.py clippy command

GuillaumeGomez:
We can now run `x.py clippy src/librustdoc`. :)

Thanks a lot @Mark-Simulacrum for the help!

r? @Mark-Simulacrum


Mark-Simulacrum:
The logic in https://github.com/rust-lang/rust/blob/69eb996b2617621c7088e567089a86a6657f2a0f/src/bootstrap/builder.rs#L939, particularly around not passing --cfg=bootstrap seems a little suspect now. I'm not sure if the sysroot thing is still true, but I suspect it would be. But we should be passing --cfg=bootstrap to clippy invocations.

We'll also want to adjust config.toml.example to add a `clippy = ...` field.

GuillaumeGomez:
Do you want them to be done in this PR or in another?

Mark-Simulacrum:
Hm, so looking into it a little more, are you *sure* this actually fixed anything? I suspect the error you pinged me with last night was transient and the changes you made here just invalidated caches to cause it to disappear.

From what I can tell, bootstrap currently just pulls cargo-clippy from the PATH, so it's probably not using the downloaded clippy even with your changes. This patch doesn't yet change that -- I'm not actually sure if it's possible to adjust that easily.

GuillaumeGomez:
I have absolutely no idea to be honest... What would be required to do to ensure that this very specific clippy is used?

Mark-Simulacrum:
I don't know -- it would require an investigation into how it's invoked by Cargo, whether that's something we can substitute or not. For rustc and rustdoc, we rely on Cargo's built-in subcommands, but cargo-clippy doesn't seem to be built in, so it may need special care. I admit that I don't fully grasp *why* that is -- I would have expected it to be roughly equivalent to `RUSTC=clippy cargo check`, but that doesn't seem to be the case.

I don't currently have the bandwidth to investigate this question and suggest a path forward though. My guess is that this patch is not actually necessary to fix ./x.py clippy regardless.

src/bootstrap/bootstrap.py
```diff
@@ -508,14 +516,25 @@ def downloading_llvm(self):
...
+
     def _download_component_helper(
         self, filename, pattern, tarball_suffix, stage0=True, key=None
     ):
         if key is None:
             if stage0:
                 key = self.stage0_compiler.date
             else:
-                key = self.rustc_commit
+                key = self.get_rustc_commit()
```
jyn514, line 537: Please don't do this without caching, it will rerun git like 12 times each time you invoke x.py.

src/bootstrap/bootstrap.py
```diff
@@ -1172,6 +1187,8 @@ def bootstrap(help_triggered):
     build.stage0_compiler = Stage0Toolchain(data["compiler"])
     if data.get("rustfmt") is not None:
         build.stage0_rustfmt = Stage0Toolchain(data["rustfmt"])
+    if data.get("clippy") is not None:
+        build.stage0_clippy = Stage0Toolchain(data["clippy"])
```
jyn514, line 1191: Does this actually do anything? I don't see stage0_clippy ever used.

jyn514 (changes requested):
@Mark-Simulacrum is right - the issue is that rustbuild is using the default clippy in PATH. Downloading clippy from static.rlo is necessary but not sufficient to fix it.

Mark-Simulacrum:
I'm going to mark as waiting-on-author but it may also be best to just close this PR for the time being.

camsteffen:
In case this is helpful:
https://github.com/rust-lang/rust-clippy/blob/master/doc/basics.md#install-from-source

jyn514:
@camsteffen we shouldn't need to build clippy, it's already been compiled for beta. rustbuild just needs to make sure it downloads and uses it rather than using the clippy in PATH.

camsteffen:
Parts of the guide still may be helpful.

PR #89426, opened 2021-10-01
bootstrap: add config option for nix patching

davidtwco:
On NixOS systems, bootstrap will patch rustc used in bootstrapping after checking `/etc/os-release` (to confirm the current distribution is NixOS). However, when using Nix on a non-NixOS system, it can be desirable for bootstrap to patch rustc. In this commit, a `patch-binaries-for-nix` option is added to `config.toml`, which allows for user opt-in to bootstrap's Nix patching.

r? @Mark-Simulacrum

### DIFF
--- a/config.toml.example
+++ b/config.toml.example
@@ -313,6 +313,12 @@ changelog-seen = 2
 # this setting's very existence, are all subject to change.)
 #print-step-rusage = false
 
+# Always patch binaries for usage with Nix toolchains. If `true` then binaries
+# will be patched unconditionally. If `false` or unset, binaries will be patched
+# only if the current distribution is NixOS. This option is useful when using
+# a Nix toolchain on non-NixOS distributions.
+#patch-binaries-for-nix = false
+
 # =============================================================================
 # General install configuration options
 # =============================================================================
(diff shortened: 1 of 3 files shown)


Mark-Simulacrum:
Hm, I would prefer that we have this as a config.toml flag (e.g., "patch-binaries-for-nix = true/false" or so). Would that work for you?

The reason to prefer an env variable is that you can set it in e.g. .bashrc or so and just get it for any rustc bootstrap, I suppose, but it feels a little icky to me to have configuration via that vector. Is there a way for us to detect that you're "wanting" to use nix that's not /etc/os-release? For example, maybe if .nix-deps exists or so we can assume we're on nix?

davidtwco:
> Hm, I would prefer that we have this as a config.toml flag (e.g., "patch-binaries-for-nix = true/false" or so). Would that work for you?

This works for me, I’ve updated the PR with this approach.

Mark-Simulacrum:
@bors r+

We might eventually want to make an explicit false opt-out (with the default being conditional on platform, like before this PR), but seems OK for now. We haven't heard any asks to opt out anyway.

Issue #89254, opened 2021-09-25
Cannot infer type for Vec, but only when serde_json is used (?)


m-ou-se:
```rust
extern crate serde_json; // Code works fine if you remove this line.

fn main() {
	let data = Vec::<u8>::new();
	assert_eq!(data[..], []);
}
```

That produces:

```
error[E0282]: type annotations needed
 --> src/main.rs:5:2
  |
5 |     assert_eq!(data[..], []);
  |     ^^^^^^^^^^^^^^^^^^^^^^^^^ cannot infer type
  |
  = note: this error originates in the macro `assert_eq` (in Nightly builds, run with -Z macro-backtrace for more info)
```

Somehow, referencing `serde_json` makes rustc unable to infer some type here. This happens on both the current stable and nightly versions of Rust.

m-ou-se:
Looks like the problem is `impl PartialEq<Value> for u8` in serde_json. Without that, `impl PartialEq<u8> for u8` is the only `PartialEq` impl for `u8`.

So I suppose this is expected, although very confusing.

Here's a reproduction without a dependency:

```rust
struct Shark;

impl PartialEq<Shark> for u8 {
    fn eq(&self, other: &Shark) -> bool {
        false
    }
}

fn main() {
    Vec::<u8>::new() == [];
}
```

Labeling this as a diagnostics issue.

jhpratt:
Of course having better diagnostics here is best for the immediate future, but has it ever been considered to have a "preferred" implementation where `Self` is the generic type? I can't recall the exact situation off-hand, but I know I've run into inference issues where that would have solved it in the past.

estebank:
For the general diagnostic improvement, we should be pointing at all the `impl`s that apply (and also that info when suggesting the turbofish, while we are at it). Having a way to nudge inference towards a certain type is something I'd like, but it feels like it'd have similar problems as specialization.

estebank:
Current output:
```
error[E0282]: type annotations needed
  --> f71.rs:10:25
   |
10 |     Vec::<u8>::new() == [];
   |                         ^^ cannot infer type

error[E0283]: type annotations needed
  --> f71.rs:10:22
   |
10 |     Vec::<u8>::new() == [];
   |                      ^^ cannot infer type for type parameter `U`
   |
note: multiple `impl`s satisfying `u8: PartialEq<_>` found
  --> f71.rs:3:1
   |
3  | impl PartialEq<Shark> for u8 {
[...]

PR #89427, opened 2021-10-01
Point at overlapping impls when type annotations are needed

estebank:
Address https://github.com/rust-lang/rust/issues/89254.

### DIFF
--- a/compiler/rustc_errors/src/diagnostic.rs
+++ b/compiler/rustc_errors/src/diagnostic.rs
@@ -465,10 +465,14 @@ impl Diagnostic {
         suggestions: impl Iterator<Item = String>,
         applicability: Applicability,
     ) -> &mut Self {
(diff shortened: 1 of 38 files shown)


petrochenkov:
It didn't work, selecting a random active member.
r? @spastorino

jackh726:
> It didn't work, selecting a random active member.

Yeah, should be "rust-lang/traits", but it might be prudent to make "wg-traits" an alias for that.

nikomatsakis (commented):
this looks good to me! I didn't give it a detailed read, though, so maybe better for @spastorino or others to look a touch more closely at the code.

jackh726 (approved):
I think mostly just nits from me.

Some "regressions" in errors, imo. But I think on balance, they're fine.

r=me with or without the like one actionable comment done

PR #89428, opened 2021-10-01
Feature gate the non_exhaustive_omitted_patterns lint

### LINKED ISSUE #89374: Feature gate new `non_exhaustive_omitted_patterns` lint

DevinR528:
Fixes https://github.com/rust-lang/rust/issues/89374

Add the machinery to gate the new `non_exhaustive_omitted_patterns` lint.

relates to https://github.com/rust-lang/rust/pull/89105 and https://github.com/rust-lang/rust/pull/89423

### DIFF
--- a/compiler/rustc_feature/src/active.rs
+++ b/compiler/rustc_feature/src/active.rs
@@ -678,6 +678,9 @@ declare_features! (
     /// Allows `#[doc(cfg_hide(...))]`.
     (active, doc_cfg_hide, "1.57.0", Some(43781), None),
 
+    /// Allows using the `non_exhaustive_omitted_patterns` lint.
+    (active, non_exhaustive_omitted_patterns_lint, "1.57.0", Some(89554), None),
+
     // -------------------------------------------------------------------------
     // feature-group-end: actual feature gates
     // -------------------------------------------------------------------------
--- a/src/test/ui/rfc-2008-non-exhaustive/reachable-patterns.stderr
+++ b/src/test/ui/rfc-2008-non-exhaustive/reachable-patterns.stderr
@@ -1,11 +1,11 @@
 warning: some fields are not explicitly listed
-  --> $DIR/reachable-patterns.rs:127:9
+  --> $DIR/reachable-patterns.rs:129:9
    |
 LL |         VariantNonExhaustive::Bar { x, .. } => {}
    |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ field `y` not listed
    |
 note: the lint level is defined here
-  --> $DIR/reachable-patterns.rs:124:12
+  --> $DIR/reachable-patterns.rs:126:12
    |
 LL |     #[warn(non_exhaustive_omitted_patterns)]
    |            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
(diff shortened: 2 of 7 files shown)


DevinR528:
cc @Nadrieril @camelid

DevinR528:
An, even more, succinct error report:

`tidy error: /checkout/compiler/rustc_feature/src/active.rs:682: no tracking issue for feature non_exhaustive_omitted_patterns
some tidy checks failed`

Nadrieril:
Ok, I created the tracking issue: https://github.com/rust-lang/rust/issues/89554
I also propose we rename the feature to `non_exhaustive_omitted_patterns_lint`, otherwise it sounds like we're making changes to how non_exhaustive works.

DevinR528:
Sweet thanks!! I rebased and added the tracking issue number.

Nadrieril:
Yes please

Nadrieril:
Haha so many checks :D Seems like you also need to rename a test file

compiler/rustc_feature/src/active.rs
```diff
@@ -675,6 +675,8 @@ declare_features! (
     /// Allows `#[track_caller]` on closures and generators.
     (active, closure_track_caller, "1.57.0", Some(87417), None),
 
+    /// Allows using the `non_exhaustive_omitted_patterns` lint.
+    (active, non_exhaustive_omitted_patterns_lint, "1.57.0", Some(89549), None),
```
camelid, line 679: This is a nit, but please leave the blank line after this feature declaration.

src/test/ui/rfc-2008-non-exhaustive/reachable-patterns.stderr
```diff
@@ -1,47 +1,47 @@
...
    |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ field `y` not listed
    |
 note: the lint level is defined here
-  --> $DIR/reachable-patterns.rs:124:12
+  --> $DIR/reachable-patterns.rs:126:12
    |
 LL |     #[warn(non_exhaustive_omitted_patterns)]
    |            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    = help: ensure that all fields are mentioned explicitly by adding the suggested fields
    = note: the pattern is of type `VariantNonExhaustive` and the `non_exhaustive_omitted_patterns` attribute was found
```
camelid, line 13: Is this lint supposed to say this instead? I'm not sure what it means by "`non_exhaustive_omitted_patterns` attribute" – that's the name of the lint, not an attribute IIUC.

```suggestion
   = note: the pattern is of type `VariantNonExhaustive` and the `non_exhaustive` attribute was found
```

(Note: Don't apply this suggestion; it's just for demonstration purposes.)

DevinR528: Because the lint is allowed by default it needs `#[warn(non_exhaustive_omitted_patterns)]` somewhere on the `match` or the above the wildcard pattern. Now that you point it out its completely redundant info since the snippet shows the lint level. 

Maybe something like `note: the pattern is of type 'VariantNonExhaustive' which is 'non_exhaustive'`

Nadrieril: Hm good point. I don't think it actually helps the user much to mention `non_exhaustive` here, since there's already a message that mentions the lint itself. "note: the pattern is of type `VariantNonExhaustive`" sounds enough

Nadrieril: Let's do that in a different PR though. I'll note the comment in the tracking issue.

DevinR528:
squashed

Thanks for the help everybody!!!

Nadrieril:
That's not because you added a blank line, it's because that line has some spaces on it instead of being empty

camelid (commented):
After fixing the tracking issue declaration and squashing, `r=Nadrieril,camelid`. Thanks for doing this follow-up feature-gating!

PR #89430, opened 2021-10-01
Fix clippy lints in librustdoc

GuillaumeGomez:
I ran clippy on librustdoc and simply fixed the lints. :)

r? @notriddle

### DIFF
--- a/src/librustdoc/clean/auto_trait.rs
+++ b/src/librustdoc/clean/auto_trait.rs
@@ -136,7 +136,7 @@ impl<'a, 'tcx> AutoTraitFinder<'a, 'tcx> {
         let f = auto_trait::AutoTraitFinder::new(tcx);
 
         debug!("get_auto_trait_impls({:?})", ty);
-        let auto_traits: Vec<_> = self.cx.auto_traits.iter().cloned().collect();
+        let auto_traits: Vec<_> = self.cx.auto_traits.iter().copied().collect();
         let mut auto_traits: Vec<Item> = auto_traits
             .into_iter()
             .filter_map(|trait_def_id| {
@@ -193,8 +193,8 @@ impl<'a, 'tcx> AutoTraitFinder<'a, 'tcx> {
         // to its smaller and larger regions. Note that 'larger' regions correspond
         // to sub-regions in Rust code (e.g., in 'a: 'b, 'a is the larger region).
         for constraint in regions.constraints.keys() {
-            match constraint {
-                &Constraint::VarSubVar(r1, r2) => {
+            match *constraint {
+                Constraint::VarSubVar(r1, r2) => {
                     {
                         let deps1 = vid_map.entry(RegionTarget::RegionVid(r1)).or_default();
                         deps1.larger.insert(RegionTarget::RegionVid(r2));
@@ -203,15 +203,15 @@ impl<'a, 'tcx> AutoTraitFinder<'a, 'tcx> {
                     let deps2 = vid_map.entry(RegionTarget::RegionVid(r2)).or_default();
                     deps2.smaller.insert(RegionTarget::RegionVid(r1));
                 }
-                &Constraint::RegSubVar(region, vid) => {
+                Constraint::RegSubVar(region, vid) => {
                     let deps = vid_map.entry(RegionTarget::RegionVid(vid)).or_default();
                     deps.smaller.insert(RegionTarget::Region(region));
                 }
-                &Constraint::VarSubReg(vid, region) => {
+                Constraint::VarSubReg(vid, region) => {
                     let deps = vid_map.entry(RegionTarget::RegionVid(vid)).or_default();
                     deps.larger.insert(RegionTarget::Region(region));
                 }
-                &Constraint::RegSubReg(r1, r2) => {
+                Constraint::RegSubReg(r1, r2) => {
                     // The constraint is already in the form that we want, so we're done with it
                     // Desired order is 'larger, smaller', so flip then
                     if region_name(r1) != region_name(r2) {
@@ -513,8 +513,8 @@ impl<'a, 'tcx> AutoTraitFinder<'a, 'tcx> {
                         // as we want to combine them with any 'Output' qpaths
                         // later
 
-                        let is_fn = match &mut b {
-                            &mut GenericBound::TraitBound(ref mut p, _) => {
+                        let is_fn = match b {
+                            GenericBound::TraitBound(ref mut p, _) => {
                                 // Insert regions into the for_generics hash map first, to ensure
                                 // that we don't end up with duplicate bounds (e.g., for<'b, 'b>)
                                 for_generics.extend(p.generic_params.clone());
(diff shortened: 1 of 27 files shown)


src/librustdoc/clean/auto_trait.rs
```diff
@@ -526,8 +526,8 @@ impl<'a, 'tcx> AutoTraitFinder<'a, 'tcx> {
                         // as we want to combine them with any 'Output' qpaths
                         // later
 
-                        let is_fn = match &mut b {
-                            &mut GenericBound::TraitBound(ref mut p, _) => {
+                        let is_fn = match *(&mut b) {
```
jyn514, line 529:
```suggestion
                        let is_fn = match b {
```

src/librustdoc/clean/auto_trait.rs
```diff
@@ -723,8 +723,8 @@ impl<'a, 'tcx> AutoTraitFinder<'a, 'tcx> {
 
     fn is_fn_ty(&self, ty: &Type) -> bool {
         let tcx = self.cx.tcx;
-        match ty {
-            &Type::ResolvedPath { did, .. } => {
+        match *ty {
+            Type::ResolvedPath { did, .. } => {
```
jyn514, line 727: What benefit does this change have? It just seems like churn.

src/librustdoc/clean/mod.rs
```diff
@@ -275,7 +273,7 @@ impl Clean<WherePredicate> for hir::WherePredicate<'_> {
             hir::WherePredicate::BoundPredicate(ref wbp) => {
                 let bound_params = wbp
                     .bound_generic_params
-                    .into_iter()
+                    .iter()
```
jyn514, line 276: Why this change?

src/librustdoc/core.rs
```diff
@@ -464,7 +464,7 @@ crate fn run_global_ctxt(
                 _ => continue,
             };
             for name in value.as_str().split_whitespace() {
-                let span = attr.name_value_literal_span().unwrap_or(attr.span());
+                let span = attr.name_value_literal_span().unwrap_or_else(|| attr.span());
```
jyn514, line 467: span() is really cheap, this just seems like churn.

src/librustdoc/formats/cache.rs
```diff
@@ -464,7 +464,7 @@ impl<'a, 'tcx> DocFolder for CacheBuilder<'a, 'tcx> {
             let impl_item = Impl { impl_item: item };
             if impl_item.trait_did().map_or(true, |d| self.cache.traits.contains_key(&d)) {
                 for did in dids {
-                    self.cache.impls.entry(did).or_insert(vec![]).push(impl_item.clone());
+                    self.cache.impls.entry(did).or_insert_with(Vec::new).push(impl_item.clone());
                 }
```
jyn514, line 468: This also seems like churn.

jyn514 (changes requested):

src/librustdoc/clean/auto_trait.rs
GuillaumeGomez: In case we add other pattern matching, it allows to not have `&` (which isn't very useful to change in this context, I agree). Oh well, clippy lints. :)

I found this one interesting for future developments so I decided to "fix" it.

src/librustdoc/clean/mod.rs
GuillaumeGomez: Lint. But it didn't make much sense I guess to consume the type here I guess?

notriddle (changes requested):

src/librustdoc/passes/html_tags.rs
```diff
@@ -207,7 +207,7 @@ impl<'a, 'tcx> DocFolder for InvalidHtmlTagsLinter<'a, 'tcx> {
 
             for (tag, range) in tags.iter().filter(|(t, _)| {
                 let t = t.to_lowercase();
-                ALLOWED_UNCLOSED.iter().find(|&&at| at == t).is_none()
+                !ALLOWED_UNCLOSED.iter().any(|&at| at == t)
```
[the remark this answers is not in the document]
GuillaumeGomez: Let's do that in another PR then. The PR is already big enough.

src/librustdoc/html/sources.rs
```diff
@@ -168,7 +168,11 @@ impl SourceCollector<'_, 'tcx> {
         };
 
         // Remove the utf-8 BOM if any
-        let contents = if contents.starts_with('\u{feff}') { &contents[3..] } else { &contents };
+        let contents = if let Some(stripped) = contents.strip_prefix('\u{feff}') {
+            stripped
+        } else {
+            &contents
+        };
```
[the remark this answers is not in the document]
GuillaumeGomez: Interesting that it wasn't suggested. Your version is far better!

src/librustdoc/core.rs
GuillaumeGomez: It has a small `match` (see [here](https://doc.rust-lang.org/nightly/nightly-rustc/src/rustc_ast/attr/mod.rs.html#548-553)), so I guess it counts as micro optimization?

src/librustdoc/formats/cache.rs
GuillaumeGomez: Not entirely. At least we don't allocate a `Vec` (even if it doesn't take much space). Without clippy, I wouldn't have cared but since it showed it to me, why not simply fixing it...

src/librustdoc/passes/collect_intra_doc_links.rs
```diff
@@ -1497,7 +1497,7 @@ fn range_between_backticks(ori_link: &MarkdownLink) -> Range<usize> {
         .bytes()
         .skip(after_first_backtick_group)
         .position(|b| b == b'`')
-        .unwrap_or(ori_link.link.len());
+        .unwrap_or_else(|| ori_link.link.len());
```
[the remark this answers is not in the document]
GuillaumeGomez: Micro optimization one would say: `String::len` calls `self.vec.len()` (which simply returns its `len` field). Well, lint fixed at least haha.

src/librustdoc/clean/mod.rs
jyn514: I'm asking *your* opinion, not clippy's opinion. I don't think it's useful to change code just because clippy says to, there should be a reason.

In particular, this prevents reusing the underlying allocation, which uses more memory than it did before.

src/librustdoc/formats/cache.rs
jyn514: `vec![]` does not allocate.

src/librustdoc/passes/html_tags.rs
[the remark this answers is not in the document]
jyn514: If you're going to add churn, it should be making the code better. Adding itertools doesn't make the PR any bigger than before.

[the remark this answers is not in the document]
notriddle: Opened #89444

src/librustdoc/clean/mod.rs
GuillaumeGomez: So it's a regression. Reverting it then.

The full story is actually that clippy was linting about the `collect` (which can't be removed unfortunately). After a few tries, I simply ended with this and didn't revert completely, my bad.

src/librustdoc/formats/cache.rs
GuillaumeGomez: It creates a `Vec` type which is 8 bytes? Well, it's completely anecdotic but in this case I still consider this as better code.

GuillaumeGomez:
Updated!

notriddle (approved):

PR #89433, opened 2021-10-01
Fix ctrl-c causing reads of stdin to return empty on Windows.

### LINKED ISSUE #89177: Execution continues after stdin read_line when ctrl-c is pressed on Windows
When blocking on `read_line` from `stdin`, if I press Ctrl-C, program execution continues past the blocking `read_line` call for a short period before the program is terminated on Windows.

I tried this code:

```rust
fn main() -> std::io::Result<()> {
    use std::io::BufRead;
    println!("Press Ctrl-C");
    let mut line = String::new();
    let bytes = std::io::stdin().lock().read_line(&mut line)?;
    println!("This should not print if Ctrl-C was pressed. Read {} bytes", bytes);
    Ok(())
}
```

**I expected to see this happen:** Program should **exit** after pressing Ctrl-C without printing the second message.
**Instead, this happened:** Program prints `This should not print if Ctrl-C was pressed. Read 0 bytes` after pressing Ctrl-C. 

This only occurs on Windows. *nix platforms exit without printing the second message.

Since program execution continues after ctrl-c, unexpected results occur. For example, running `cargo login`, then pressing `ctrl-c` instead of entering a token, `cargo` will overwrite the saved token with `""` on Windows.

*nix platforms continue blocking in `read_line` when ctrl-c is pressed. Including if a handler for ctrl-c is set up.

### Why it's happening
`read_line` calls the Windows `ReadConsoleW` API. That API returns *success* with an empty buffer for Ctrl-C (and Ctrl-Break). `read_line` then also returns `Ok`.  In parallel, Windows uses a separate thread that raises an exception for the ctrl-c event that terminates the process. This leads to a short window of time where program execution continues after ctrl-c is pressed.

### Potential solutions
Even though `ReadConsoleW` returns success, it also sets LastError to `ERROR_OPERATION_ABORTED` for this case, so we *can* detect this case.

The Rust standard library calls `ReadConsoleW` here
https://github.com/rust-lang/rust/blob/cfff31bc833070a00578bd6178160aeed56f28ba/library/std/src/sys/windows/stdio.rs#L295

#### Re-try the call immediately
We could re-try the call to `ReadConsoleW` immediately for this specific case by adding a check after the `ReadConsoleW` call.
```rust
    loop {
        cvt(unsafe {
            c::SetLastError(0);
            c::ReadConsoleW(
                handle,
                buf.as_mut_ptr() as c::LPVOID,
                buf.len() as u32,
                &mut amount,
                &mut input_control as c::PCONSOLE_READCONSOLE_CONTROL,
            )
        })?;

        // ReadConsoleW returns success with ERROR_OPERATION_ABORTED for Ctrl-C or Ctrl-Break.
        // Explicitly check for that case here and try again.
        if amount == 0 {
            if unsafe { c::GetLastError() } == c::ERROR_OPERATION_ABORTED {
                continue;
            }
        }
        break;
    }
```

#### Return an error
Alternately, after the call to `ReadConsoleW`, we could add a check to detect this error and propagate it:
```rust
    if amount == 0 {
        let err = crate::io::Error::last_os_error();
        if err.raw_os_error() == Some(c::ERROR_OPERATION_ABORTED as i32) {
            return Err(err);
        }
    }
```
Returning an error still differs from *nix, in that it raises an error instead of continuing to block.

#### Return an error then retry
We could return the error as described above, and change the mapping of `ERROR_OPERATION_ABORTED` from `ErrorKind::TimedOut` to `ErrorKind::Interrupted`. 
https://github.com/rust-lang/rust/blob/cfff31bc833070a00578bd6178160aeed56f28ba/library/std/src/sys/windows/mod.rs#L79

`ErrorKind::Interrupted` would then re-tried the `BufRead` infrastructure.

#### Do nothing
Maybe this is expected behavior and we want a difference between Windows and *nix here.

### Meta
Occurs in both stable and nightly.

`rustc --version --verbose`:
```
rustc 1.57.0-nightly (497ee321a 2021-09-09)
binary: rustc
commit-hash: 497ee321af3b8496eaccd7af7b437f18bab81abf
commit-date: 2021-09-09
host: x86_64-pc-windows-msvc
release: 1.57.0-nightly
LLVM version: 13.0.0
```

arlosi:
Pressing ctrl+c (or ctrl+break) on Windows caused a blocking read of stdin to unblock and return empty, unlike other platforms which continue to block.

On ctrl-c, `ReadConsoleW` will return success, but also set `LastError` to `ERROR_OPERATION_ABORTED`.

This change detects this case, and re-tries the call to `ReadConsoleW`.

Fixes #89177. See issue for further details.

Tested on Windows 7 and Windows 10 with both MSVC and GNU toolchains

### DIFF
--- a/library/std/src/sys/windows/stdio.rs
+++ b/library/std/src/sys/windows/stdio.rs
@@ -291,15 +291,25 @@ fn read_u16s(handle: c::HANDLE, buf: &mut [u16]) -> io::Result<usize> {
     };
 
     let mut amount = 0;
-    cvt(unsafe {
-        c::ReadConsoleW(
-            handle,
-            buf.as_mut_ptr() as c::LPVOID,
-            buf.len() as u32,
-            &mut amount,
-            &mut input_control as c::PCONSOLE_READCONSOLE_CONTROL,
-        )
-    })?;
+    loop {
+        cvt(unsafe {
+            c::SetLastError(0);
+            c::ReadConsoleW(
+                handle,
+                buf.as_mut_ptr() as c::LPVOID,
+                buf.len() as u32,
+                &mut amount,
+                &mut input_control as c::PCONSOLE_READCONSOLE_CONTROL,
+            )
+        })?;
+
+        // ReadConsoleW returns success with ERROR_OPERATION_ABORTED for Ctrl-C or Ctrl-Break.
+        // Explicitly check for that case here and try again.
+        if amount == 0 && unsafe { c::GetLastError() } == c::ERROR_OPERATION_ABORTED {
+            continue;
+        }
+        break;
+    }
 
     if amount > 0 && buf[amount as usize - 1] == CTRL_Z {
         amount -= 1;


fee1-dead:
Can you add an ui test for this? ([see this](https://rustc-dev-guide.rust-lang.org/tests/adding.html) for how to add one yourself)

Ideally we would want to make sure this behavior is consistent across all the platforms that we support.

It would probably look like [this file](https://github.com/rust-lang/rust/blob/master/src/test/ui/print-stdout-eprint-stderr.rs), although I am not sure how to programmatically send Ctrl-C to a process... (and there might be platforms that do not care about Ctrl-Cs)

Mark-Simulacrum:
I think a test will be difficult -- maybe a run-make test with a `kill` command sending the appropriate signal? Seems complicated and may not be warranted.

I'm going to r? @dtolnay since this seems like it'll require a T-libs-api (or so) FCP, as a breaking change.

arlosi:
Yes, adding a test is rather difficult. Windows doesn't really have a `kill` like syscall that sends `SIGINT`.

The closest thing seems to be setting up separate process group for a subprocess, then calling the Windows `GenerateConsoleCtrlEvent` API on the group. However, I wasn't able to reproduce the current behavior using this approach -- and even if we could make it work, the test would be very Windows-specific.

dtolnay:
This PR closes an inconsistency between how the standard library implements reading from stdin on Windows vs other platforms. https://github.com/rust-lang/rust/issues/89177#issuecomment-926642320 is the clearest example of the discrepancy: ctrl+C during a read would previously cause the read to return on Windows, but not other platforms, while the ctrl+C handler runs.

I don't believe we want to keep this as a platform-specific inconsistency. As for which behavior to go with, I would biasedly say the non-Windows behavior seems better to me, which this PR makes the Windows implementation consistent with.

@rust-lang/libs-api 
@rfcbot fcp merge

BurntSushi:
This seems okay to me from a consistency perspective, but I wonder how other ecosystems handle this? Has anyone done a survey?

arlosi:
I tested a few different languages on Windows 10 with the program:
* Set signal handler to catch ctrl-c and print a message
* Read a line of text

The behavior of the readline call after pressing ctrl-c are a mixture on Windows:
| Language          | readline keeps blocking | readline raises error     |
|-------------------|----------------|-----------------|
| Rust (now)        | ❌ | ❌ |
| Rust (this PR)    | ✔️| ❌ |
| Go                | ❌ | ✔️ (EOF)       |
| Node (JavaScript) | ✔️| ❌ |
| Ruby              | ✔️| ❌ |
| .NET Core         | ❌ | ❌ |
| Python            | ❌ | ✔️ (Exception) |

**All** test programs have the same behavior on Linux (keep blocking, no error). For consistency, converging on "keep blocking, no error" (as this PR does) makes the most sense to me.

The alternative of returning an error in Rust can cause problems if no signal handler is set up, since Windows kills the thread after a few milliseconds, sometimes causing a half-printed error message.

PR #89437, opened 2021-10-01
Specialize `Vec::extend` and `Vec::from_iter` for arrays

a1phyr:
This brings performance and code size improvements.

Drawback: this adds one layer one specialization for `SpecExtend`.

### DIFF
--- a/library/alloc/src/vec/mod.rs
+++ b/library/alloc/src/vec/mod.rs
@@ -110,6 +110,9 @@ mod is_zero;
 #[cfg(not(no_global_oom_handling))]
 mod source_iter_marker;
 
+#[cfg(not(no_global_oom_handling))]
+use source_iter_marker::SpecFromIterWithSource;
+
 mod partial_eq;
 
 #[cfg(not(no_global_oom_handling))]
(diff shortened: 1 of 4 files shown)


Mark-Simulacrum:
I don't feel solid enough with specialization to make sure this isn't adding a soundness hole or so. Do we have a guide yet for review on that?

r? @m-ou-se

the8472:
Are you sure this requires a new specialization helper trait? Both `array::IntoIter` and `vec::IntoIter` are struct types (as opposed to traits) which means they should be more specific than `SourceIterMarker` and can't overlap with each other. I don't see where the ambiguity comes from that would requires manual resolution.

the8472:
Taking over review for this one since due to prior experience in this area. r? @the8472 

@a1phyr do you have any benchmarks/assembly examples showing improvements? There aren't many uses of array iterators in the compiler or perf rlo benchmark suite yet so I wouldn't expect big improvements there. The additional layers of indirection might make things worse though. @bors try @rust-timer queue

And some additional questions in my previous comment.

JohnCSimon:
Ping from triage: 
@a1phyr Can you please post your status on this PR?

the8472:
> This brings performance and code size improvements.

@a1phyr as anticipated in https://github.com/rust-lang/rust/pull/89437#issuecomment-971896765 perf doesn't show a huge difference, do you have any benchmarks?

JohnCSimon:
@a1phyr 

@rustbot label: +S-inactive

PR #89438, opened 2021-10-01
docs: `std::hash::Hash` should ensure prefix-free data

### LINKED ISSUE #89429: `std::hash::Hash` documentation should suggest that the hash data should be prefix-free
As discussed at [Why does `str.hash(…)` pass an extra byte to the Hasher?](https://users.rust-lang.org/t/why-does-str-hash-pass-an-extra-byte-to-the-hasher/65404): the code of `impl Hash for &str` specifically passes an extra 0xFF byte to the Hasher, so that values like `("ab", "c")` and `("a", "bc")` hash differently (added in https://github.com/rust-lang/rust/commit/6066118b2988fee1b758d542cee18ffe7a84acab).

This is a subtle property of hashing and should probably be mentioned in the documentation for `Hash` — _particularly_ as [the documentation for `Hasher`](https://doc.rust-lang.org/std/hash/trait.Hasher.html) already says “you cannot assume, for example, that a `write_u32` call is equivalent to four calls of `write_u8`” which could be sloppily interpreted as an expectation that a _good_ `Hasher` implementation will handle “quoting” of sequential calls itself.

If an implementor of `Hash` fails to have this property, it could compromise the hash-DoS resistance that Rust tries to offer by default.

@rustbot labels: +A-docs +T-libs-api +C-enhancement

pierwill:
Attempt to synthesize the discussion in #89429 into a suggestion regarding `Hash` implementations (not a hard requirement).

Closes #89429.

### DIFF
--- a/library/core/src/hash/mod.rs
+++ b/library/core/src/hash/mod.rs
@@ -156,8 +156,10 @@ mod sip;
 /// ## Prefix collisions
 ///
 /// Implementations of `hash` should ensure that the data they
-/// pass to the `Hasher` are prefix-free. That is, different concatenations
-/// of the same data should not produce the same output.
+/// pass to the `Hasher` are prefix-free. That is,
+/// unequal values should cause two different byte sequences to be written,
+/// and neither of the two sequences should be a prefix of the other.
+///
 /// For example, the standard implementation of [`Hash` for `&str`][impl] passes an extra
 /// `0xFF` byte to the `Hasher` so that the values `("ab", "c")` and `("a",
 /// "bc")` hash differently.
--- a/library/core/src/hash/mod.rs
+++ b/library/core/src/hash/mod.rs
@@ -153,9 +153,19 @@ mod sip;
 /// Thankfully, you won't need to worry about upholding this property when
 /// deriving both [`Eq`] and `Hash` with `#[derive(PartialEq, Eq, Hash)]`.
 ///
+/// ## Prefix collisions
+///
+/// Implementations of `hash` should ensure that the data they
+/// pass to the `Hasher` are prefix-free. That is, different concatenations
+/// of the same data should not produce the same output.
+/// For example, the standard implementation of [`Hash` for `&str`][impl] passes an extra
+/// `0xFF` byte to the `Hasher` so that the values `("ab", "c")` and `("a",
+/// "bc")` hash differently.
+///
 /// [`HashMap`]: ../../std/collections/struct.HashMap.html
 /// [`HashSet`]: ../../std/collections/struct.HashSet.html
 /// [`hash`]: Hash::hash
+/// [impl]: ../../std/primitive.str.html#impl-Hash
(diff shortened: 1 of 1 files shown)


pierwill:
@cuviper

library/core/src/hash/mod.rs
```diff
@@ -153,9 +153,19 @@ mod sip;
 /// Thankfully, you won't need to worry about upholding this property when
 /// deriving both [`Eq`] and `Hash` with `#[derive(PartialEq, Eq, Hash)]`.
 ///
+/// ## Prefix collisions
+///
+/// Implementations of `hash` should ensure that the data they
+/// pass to the `Hasher` are prefix-free. That is, different concatenations
```
tczajka, line 159: I this this explanation of what "prefix-free" means is not accurate or incomplete. It should say that unequal values should cause two different byte sequences to be written, and neither of the two sequences should be a prefix of the other.

Note that it's not sufficient to say that concatenations of multiple values of the type should result in different outputs. It has to be true when concatenated with outputs for other types as well. Which is true if all these types satisfy the prefix-free property.

pierwill: Thanks @tczajka! I'm not sure I understand the idea of one sequence being a prefix of another. Does it simply mean "starts with", or is it another kind of relation? Is there a way we can rephrase this?

pierwill: Another way to ask the question: in the example of `("ab", "c")` and `("a", "bc")` where and how would the "prefix" occur, and how does the extra byte prevent it?

tczajka: A "prefix" is a beginning of a string, so it's same as "starts with". https://en.wikipedia.org/wiki/Prefix

If strings were hashed without the extra 0xff at the end, hashing `("ab", "c")` and `("a", "bc")` would write the same byte sequence `"abc"` to `Hasher`. The problem is that `"a"` is a prefix of `"ab"`. Whereas `"a\xff"` is not a prefix of `"ab\xff"`, so if `Hash` outputs these sequences instead that solves the problem. `"ab\xffc\xff" != "a\xffbc\xff"`.

cuviper: Note: `\xff` is not actually allowed in string literals, since it would be invalid UTF-8 -- which is also what makes it a useful separator here.

library/core/src/hash/mod.rs
```diff
@@ -153,9 +153,21 @@ mod sip;
 /// Thankfully, you won't need to worry about upholding this property when
 /// deriving both [`Eq`] and `Hash` with `#[derive(PartialEq, Eq, Hash)]`.
 ///
+/// ## Prefix collisions
```
pierwill, line 156: Thinking more about this... "Collision" isn't the right term, here, is it?

pierwill:
@Amanieu Does this need to be rebased?

PR #89439, opened 2021-10-01
Refactor error creation from `ResolutionError`

tomerze:
A small refactor to change the `Resolver::into_struct_error` method to `ResolutionError::into_struct_error` and split it to multiple methods in a separate file.


petrochenkov:
I don't see much point in this change, it creates some churn, but I don't see what benefits it gives instead.

PR #89440, opened 2021-10-01
Clarify a sentence in the documentation of Vec (#84488)

### LINKED ISSUE #84488: Clarify Vec's documentation about buffer reuse
The [documentation for `Vec`](https://github.com/rust-lang/rust/blob/41f0e13bc53396852f4f1338ce5d8be0d1125b08/library/alloc/src/vec/mod.rs#L347) states:

> Do not rely on removed data to be erased for security purposes. Even if you drop a `Vec`, its buffer may simply be reused by another `Vec`.

Here, "its buffer may simply be reused" is simply meant to refer to the behavior of the underlying allocator.  But I have seen people with low-level experience be misled by this into thinking that the implementation of `Vec` itself has its own global allocation cache.  They then thought this cache was responsible for excess memory usage in their programs.

I'm not sure what the best way to clarify this would be.  Part of the reason the current wording can be misleading is that "reused by another `Vec`" suggests a mechanism specific to `Vec`.  On the other hand, it would also help to just explicitly mention the allocator.  Here is a possible version that addresses both points, but it's more wordy than the current version:

> Even if you drop a `Vec`, the allocator may reuse its buffer for another `Vec` or other allocation.

chrismit3s:
I literally changed a single word, but this should clear up the confusion of #84488. Can probably be `rollup`ed :) (My first PR, hope I'm doing this right)

fixes #84488 

r? @steveklabnik

### DIFF
--- a/library/alloc/src/vec/mod.rs
+++ b/library/alloc/src/vec/mod.rs
@@ -369,7 +369,7 @@ mod spec_extend;
 /// scratch space that it may use however it wants. It will generally just do
 /// whatever is most efficient or otherwise easy to implement. Do not rely on
 /// removed data to be erased for security purposes. Even if you drop a `Vec`, its
-/// buffer may simply be reused by another `Vec`. Even if you zero a `Vec`'s memory
+/// buffer may simply be reused by another allocation. Even if you zero a `Vec`'s memory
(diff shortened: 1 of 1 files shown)


steveklabnik:
@bors: r+ rollup

yep, looks great, thank you!

PR #89442, opened 2021-10-01
Add check for duplicated doc aliases

GuillaumeGomez:
r? @estebank

### DIFF
--- a/compiler/rustc_passes/src/check_attr.rs
+++ b/compiler/rustc_passes/src/check_attr.rs
@@ -9,7 +9,7 @@ use rustc_middle::ty::query::Providers;
 use rustc_middle::ty::TyCtxt;
 
 use rustc_ast::{ast, AttrStyle, Attribute, Lit, LitKind, NestedMetaItem};
-use rustc_data_structures::stable_set::FxHashSet;
+use rustc_data_structures::fx::{FxHashMap, FxHashSet};
 use rustc_errors::{pluralize, struct_span_err, Applicability};
 use rustc_feature::{AttributeType, BUILTIN_ATTRIBUTE_MAP};
 use rustc_hir as hir;
@@ -66,6 +66,7 @@ impl CheckAttrVisitor<'tcx> {
         target: Target,
         item: Option<ItemLike<'_>>,
     ) {
+        let mut doc_aliases = FxHashMap::default();
         let mut is_valid = true;
         let mut specified_inline = None;
         let mut seen = FxHashSet::default();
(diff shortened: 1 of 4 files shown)


joshtriplett:
Does this check only for duplicate aliases on one item, or could it also check for duplicates throughout a crate? I would love to see a warning lint for the latter.

GuillaumeGomez:
The latter is perfectly fine though. You can wrap a same C function differently a few times so I definitely don't think that we should warn about it.

compiler/rustc_passes/src/check_attr.rs
```diff
@@ -582,17 +590,38 @@ impl CheckAttrVisitor<'tcx> {
         if &*item_name.as_str() == doc_alias {
             return err_fn(meta.span(), "is the same as the item's name");
         }
+        let span = meta.span();
+        if aliases.try_insert(doc_alias.to_owned(), span).is_err() {
+            self.tcx.struct_span_lint_hir(UNUSED_ATTRIBUTES, hir_id, span, |lint| {
+                lint.build("doc alias is duplicated")
+                    .span_label(*aliases.get(doc_alias).unwrap(), "first defined here")
```
Mark-Simulacrum, line 597: nit: you don't need the second lookup here, right? You can do `if let Err(entry) = ...` above and then `*entry.get()` here?

GuillaumeGomez: Absolutely! Great catch!

GuillaumeGomez: Ah actually no, I did read it well the first time: it returns the non-inserted value, which is useless in here.

Mark-Simulacrum: *entry.entry.get() should still work: the returned Err contains both the value passed in and the hashmap entry you would have overwritten with a regular insert.

GuillaumeGomez: Damn, didn't pay attention to that. This API is even better than what I originally thought!

GuillaumeGomez:
Updated!

joshtriplett:
@GuillaumeGomez Libs, at least, has a policy against doing that, so it'd be helpful to have a lint we could turn on to enforce that.

GuillaumeGomez:
No, even libs don't want it: [example](https://doc.rust-lang.org/nightly/std/collections/hash_map/struct.HashMap.html?search=!).

joshtriplett:
Fair enough, it wouldn't be universal. It might be useful as a warning to be opted out of, but in any case that's not a matter for this PR.

GuillaumeGomez:
I think a rustc-specific lint might be the way to go. But yes, out of scope here. ;)

PR #89443, opened 2021-10-01
Include the length in BTree hashes

cuviper:
This change makes it consistent with `Hash` for all other collections.


dtolnay (approved):
Thanks!

PR #89444, opened 2021-10-01
rustdoc: use slice::contains instead of open-coding it

notriddle:


GuillaumeGomez (approved):

GuillaumeGomez:
Thanks!

r=me once CI pass.

src/librustdoc/passes/html_tags.rs
```diff
@@ -207,7 +208,7 @@ impl<'a, 'tcx> DocFolder for InvalidHtmlTagsLinter<'a, 'tcx> {
 
             for (tag, range) in tags.iter().filter(|(t, _)| {
                 let t = t.to_lowercase();
-                ALLOWED_UNCLOSED.iter().find(|&&at| at == t).is_none()
+                !ALLOWED_UNCLOSED.iter().contains(&t.as_str())
```
cuviper, line 211: Better yet, just use the slice [`contains`](https://doc.rust-lang.org/std/primitive.slice.html#method.contains).

There's another instance with `ALLOWED_UNCLOSED` like this in `fn drop_tag`.

notriddle:
Sorry about that. I probably should've noticed that `slice::contains` existed.

PR #89446, opened 2021-10-01
Add paragraph to ControlFlow docs to menion it works with the ? operator (#88715)

### LINKED ISSUE #88715: ControlFlow docs should explicitly mention that it works with ?
We stabilized ControlFlow: https://doc.rust-lang.org/stable/std/ops/enum.ControlFlow.html

That’s somewhat of a big deal, as it is now the third type (in addition to Option and Result), which works with `?` operator on stable. This isn’t immediately obvious from the docs though!

chrismit3s:
fixes #88715

r? @steveklabnik

### DIFF
--- a/library/core/src/ops/control_flow.rs
+++ b/library/core/src/ops/control_flow.rs
@@ -7,6 +7,10 @@ use crate::{convert, ops};
 /// Having the enum makes it clearer -- no more wondering "wait, what did `false`
 /// mean again?" -- and allows including a value.
 ///
+/// Similar to [`Option`] and [`Result`], this enum can be used with the `?` operator
+/// to return immediatly if the [`Break`] variant is present or to continue normally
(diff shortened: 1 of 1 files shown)


library/core/src/ops/control_flow.rs
```diff
@@ -7,6 +7,10 @@ use crate::{convert, ops};
 /// Having the enum makes it clearer -- no more wondering "wait, what did `false`
 /// mean again?" -- and allows including a value.
 ///
+/// Similar to [`Option`] and [`Result`], this enum can be used with the `?` operator
+/// to return immediatly if the [`Break`] variant is present or to continue normally
```
r00ster91, line 11:
```suggestion
/// to return immediately if the [`Break`] variant is present or otherwise continue normally
```

steveklabnik (approved):

Issue #89432, opened 2021-10-01
nightly-2021-10-01: the trait `From<&str>` is not implemented for `std::string::String`


npmccallum:
I tried this code:

```rust
pub struct Datum {
    /// The name of this datum.
    pub name: String,

    /// Whether the datum indicates support for the platform or not.
    pub pass: bool,

    /// Short additional information to display to the user.
    pub info: Option<String>,

    /// Longer explanatory message on how to resolve problems.
    pub mesg: Option<String>,
}
```

```rust
fn dev_kvm() -> super::Datum {
    let dev_kvm = std::path::Path::new("/dev/kvm");

    super::Datum {
        name: "Driver".into(),
        pass: dev_kvm.exists(),
        info: Some("/dev/kvm".into()),
        mesg: None,
    }
}
```

I expected to see this happen:

Successful compilation. This has compiled and worked for years until `nightly-2021-10-01`. Reverting to `nightly-2021-09-30` produces a successful compilation.

Instead, this happened:

```
error[E0277]: the trait bound `std::string::String: From<&str>` is not satisfied
  --> src/backend/kvm/mod.rs:23:24
   |
23 |         name: "Driver".into(),
   |                        ^^^^ the trait `From<&str>` is not implemented for `std::string::String`
   |
   = note: required because of the requirements on the impl of `Into<std::string::String>` for `&str`
help: consider introducing a `where` bound, but there might be an alternative better way to express this requirement
   |
19 | fn dev_kvm() -> super::Datum where std::string::String: From<&str> {
   |                              +++++++++++++++++++++++++++++++++++++

error[E0277]: the trait bound `std::string::String: From<&str>` is not satisfied
  --> src/backend/kvm/mod.rs:25:20
   |
25 |         info: Some("/dev/kvm".into()),
   |               ---- ^^^^^^^^^^^^^^^^^ the trait `From<&str>` is not implemented for `std::string::String`
   |               |
   |               required by a bound introduced by this call
   |
   = note: required because of the requirements on the impl of `Into<std::string::String>` for `&str`
help: consider introducing a `where` bound, but there might be an alternative better way to express this requirement
   |
19 | fn dev_kvm() -> super::Datum where std::string::String: From<&str> {
   |                              +++++++++++++++++++++++++++++++++++++
```

### Meta

`rustc --version --verbose`:
```
rustc 1.57.0-nightly (aa7aca3b9 2021-09-30)
binary: rustc
[...]

PR #89450, opened 2021-10-01
Revert #86853

usbalbin:
Should fix issue found in #89432


eddyb:
For the record, while I'm not going to open my own PR, I was able to get a simpler revert by reverting the *merge commit* instead of every single commit (https://github.com/eddyb/rust/commit/5345239be2e347dae1c5bc887959b8b7813c7aa6), using:
```sh
git revert -m 1 c6007fdc7059c677a6c089e8d2915b264c0d1326
```

usbalbin:
Oh, thanks :)

usbalbin:
Should I add some sort of regression test like the one mentioned [here](https://github.com/rust-lang/rust/issues/89432#issuecomment-932539620)? If so where should I put it?

oli-obk:
Yea, that seems useful, but it could also wait for the real fix PR. I'm not sure where we have the tests for const trait bounds, but probably in src/test/consts

oli-obk:
r? @oli-obk 

@bors delegate+

Feel free to r=oli-obk in case you add a test

usbalbin:
> Yea, that seems useful, but it could also wait for the real fix PR. [...]

If the highest prio is getting the revert landed, then I'll wait for another PR/the fix PR. :)

PR #89453, opened 2021-10-02
Consistently use 'supertrait'.

waywardmonkeys:
A subset of places referred to 'super-trait', so this changes them
to all use 'supertrait'. This matches 'supertype' and some other
usages. An exception is 'auto-trait' which is consistently used
in that manner.


nagisa:
There are a ton of prior art of "super" and especially "sub" being combined into a single word (cf. subset, superset, subtext, supernormal...) so this seems… right?

There are a fair number of uses of `super type` around too, though.

waywardmonkeys:
The reference also defines it as `supertrait` and `subtrait` and uses that consistently in all but one spot, which I've also submitted a PR for.

I also looked at which usage had more steam in this repo and it was the `supertrait` form. :)

PR #89454, opened 2021-10-02
perf: only check for `rustc_trivial_field_reads` attribute on traits, not items, impls, etc.

erikdesjardins:
The checks that are removed in this PR (originally added in #85200) caused a small perf regression: https://github.com/rust-lang/rust/pull/88824#issuecomment-932664761

Since the attribute is currently only applied to traits, I don't think it's worth keeping the additional checks for now.
If/when we decide to apply the attribute somewhere else, we can (partially) revert this and reevaluate the perf impact.

r? @nikomatsakis cc @FabianWolff


erikdesjardins:
Can I get another perf run?

PR #89455, opened 2021-10-02
Introduce linter for diagnostic messages

hkmatsumoto:
Sometimes, contributors add diagnostics not practicing [the diagnostic message convention](https://rustc-dev-guide.rust-lang.org/diagnostics.html#diagnostic-output-style-guide) and overlooked by reviewers. This PR adds a linter for diagnostic messages to preemptively warn such diagnostics and lighten the burden of reviewers.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -693,6 +693,7 @@ name = "compiletest"
 version = "0.0.0"
 dependencies = [
  "colored",
+ "diaglint",
  "diff",
  "getopts",
  "glob",
@@ -945,6 +946,19 @@ dependencies = [
  "proc-macro2",
  "quote",
  "syn",
+]
+
+[[package]]
+name = "diaglint"
+version = "0.1.8"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "3730e1cddacaff508f3dfc1cc050e27e2823a55c44664e0b3a64cbc1008b37f8"
+dependencies = [
+ "annotate-snippets",
+ "nom",
+ "nom_locate",
+ "serde",
+ "serde_json",
 ]
 
 [[package]]
@@ -2253,6 +2267,12 @@ dependencies = [
 ]
 
 [[package]]
+name = "minimal-lexical"
+version = "0.1.3"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "0c835948974f68e0bd58636fc6c5b1fbff7b297e3046f11b3b3c18bbac012c6d"
+
+[[package]]
 name = "miniz_oxide"
 version = "0.4.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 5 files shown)


Mark-Simulacrum:
Please loop me in on the bootstrap details here when this is ready for review.

hkmatsumoto:
There are cases where it's not clear to say ok or not okay.
`To be or not to be, that is the question.`
`to be or not to be, that is the question.`
`to be or not to be, that is the question`

`Let there be light: and there was light.`
`let there be light: and there was light.`
`let there be light: and there was light`

Any thoughts on these?

estebank:
> Any thoughts on these?

I'd lean towards the last of each example.

Cargo.lock
```diff
@@ -924,6 +935,19 @@ dependencies = [
  "syn",
 ]
 
+[[package]]
+name = "diaglint"
+version = "0.1.3"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "6c1144248ed7ef2fc01b2e650877e085d651530dead491da0a088725138a5b5a"
+dependencies = [
+ "codespan-reporting",
```
estebank, line 944: I see that `diaglint` uses codespan-reporting, which might be fine, but we already have _two_ error reporting libs in the codebase. I'd prefer us to reuse one of them (likely annotate-snippets).

estebank:
The code changes in this repo look good, but am slightly concerned about the out of repo linter. That being said, I think that this linter will be useful.

Mark-Simulacrum:
I don't understand why it's optional to check diagnostic messages -- shouldn't we always do that in compiletest, not gated on a CLI option?

Is there a quick link to the lint rules in diaglint? It looks like we're using a default set right now, but I'm not quickly finding docs on what that set actually is. Can you show an example of a potential lint message?

hkmatsumoto:
> I don't understand why it's optional to check diagnostic messages -- shouldn't we always do that in compiletest, not gated on a CLI option?

Fair enough, we can make it default and can manually ignore some lint if needed.

> Is there a quick link to the lint rules in diaglint? It looks like we're using a default set right now, but I'm not quickly finding docs on what that set actually is. Can you show an example of a potential lint message?

Currently I haven't written any docs, but the lints are defined [here](https://github.com/hkmatsumoto/diaglint/tree/main/src/rules). These lints are inspired from [here](https://rustc-dev-guide.rust-lang.org/diagnostics.html#diagnostic-output-style-guide)

Cargo.lock
hkmatsumoto: Switched to `annotate-snippets`. ec0a6ee3127b0a005807e919676aa578aa0a93f1

Mark-Simulacrum:
I'll leave it up to @estebank to decide, but I think this probably should receive some T-compiler attention (or at least wg-diagnostics). I'm not personally a fan of out-of-tree lints against our lint definitions -- generally, I would expect us to do that in-tree. Bringing in extra dependencies and another diagnostics engine also feels a little too much, particularly since it's in the critical path for all "first" compiletest invocations rather than some separate not-run-by-users suite. (To be clear, I don't think we should move it out of compiletest; that seems like the right place. It's just that the cost feels higher than needed).

>  It looks like we're using a default set right now, but I'm not quickly finding docs on what that set actually is. Can you show an example of a potential lint message?

I asked this earlier, but I don't think I've seen how this looks in practice -- maybe I missed it? I would want to make sure if we are linting there is a clear message included on (a) how to disable the lint and (b) what to do to fix the lint, if desired.

JohnCSimon:
Ping from triage: @hkmatsumoto tests are failing.

apiraino:
mentioned in today's T-compiler meeting [on Zulip](https://rust-lang.zulipchat.com/#narrow/stream/238009-t-compiler.2Fmeetings/topic/.5Bweekly.5D.202021-12-09.20.2354818/near/264317442), perhaps this needs a bit more discussion during a steering meeting or drafting an MCP

hkmatsumoto:
I guess having a meeting is very good to reorganize the pros and cons. I'm not very sure if I can speak/write English fast enough for smooth conversations, though.
Either way, I plan to draft an MCP in the year-end and new-year holidays.

cc @estebank @Mark-Simulacrum

apiraino:
@hkmatsumoto apologies for my poor communication and vague comment :) (I've just dropped a note to let everyone know this is progressing). Don't worry ! In case your input is needed, the team will gladly help you sort out what is needed.

thanks for your patience and contributions! :)

Dylan-DPC:
Closing this as it is waiting on a MCP

PR #89459, opened 2021-10-02
Use bitand when checking for signed integer division overflow

tspiteri:
For `self == Self::MIN && rhs == -1`, LLVM does not realize that this is the same check made by `self / rhs`, so the code generated may have some unnecessary duplication. For `(self == Self::MIN) & (rhs == -1)`, LLVM realizes it is the same check.

### DIFF
--- a/library/core/src/num/int_macros.rs
+++ b/library/core/src/num/int_macros.rs
@@ -543,7 +543,8 @@ macro_rules! int_impl {
                       without modifying the original"]
         #[inline]
         pub const fn checked_div(self, rhs: Self) -> Option<Self> {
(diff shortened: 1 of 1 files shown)


tspiteri:
An example for [`checked_div_euclid`](https://godbolt.org/z/WxKPsGKvc) shows the better code generation.

Note that this PR does not affect `checked_div` and `checked_rem`, since they make use of `unchecked_div` and `unchecked_rem` which do not have any checks, so there is already no duplication. I included the changes in the two methods for internal consistency in the file.

r00ster91:
I really think this is something that should be improved at the code generation level instead of doing this manually for every case because there are likely a lot more cases like this and it'd be cleaner IMO.

kennytm (approved):

PR #89462, opened 2021-10-02
haiku thread affinity build fix

devnexen:


nagisa:
Please rebase/adjust to remove the updates to submodules.

library/std/src/sys/unix/thread.rs
```diff
@@ -339,14 +339,16 @@ pub fn available_concurrency() -> io::Result<NonZeroUsize> {
...
 
-            if res != libc::B_OK {
-                return Err(io::Error::last_os_error());
-            }
+                if res != libc::B_OK {
+                    return Err(io::Error::new_const(io::ErrorKind::NotFound, &"The number of hardware threads is not known for the target platform"));
+                }
 
-            Ok(unsafe { NonZeroUsize::new_unchecked(sinfo.cpu_count as usize) })
+                Ok(NonZeroUsize::new_unchecked(sinfo.cpu_count as usize))
```
nagisa, line 350: This escaped my attention last time, but calls of unsafe functions in new code should be annotated with comments describing why a call like this is safe. In particular I'm having trouble finding haiku documentation for `get_system_info` that would guarantee that the returned `cpu_count` will be non-zero. A reference here would be great.

See e.g. https://github.com/rust-lang/rust/blob/673d0db5e393e9c64897005b470bfeb6d5aec61b/library/core/src/ptr/unique.rs#L88 for an example of what I mean

nagisa: Or you could use a checked constructor instead, too.

devnexen: The cpu_count field is filled with the number of cpus at boot time set here https://github.com/haiku/haiku/blob/04be20a75a6e09d9c2f801fc0e6ffd4e5b486950/src/system/kernel/main.cpp#L114 and taken here afterwards https://github.com/haiku/haiku/blob/8f16317a5b6db5c672f331814273e5857555020f/src/system/kernel/smp.cpp#L1481

PR #89465, opened 2021-10-02
Give inlining bonus to functions with single caller

In-line:


compiler/rustc_mir_transform/src/lib.rs
```diff
@@ -592,6 +593,14 @@ fn inner_optimized_mir(tcx: TyCtxt<'_>, did: LocalDefId) -> Body<'_> {
         None => {}
         Some(other) => panic!("do not use `optimized_mir` for constants: {:?}", other),
     }
+
+    // Run `mir_inliner_callers_count_map` query for inliner
+    // which depends on `mir_drops_elaborated_and_const_checked`
+    // which we are going to steal below, so cache it's result.
+    if crate::inline::is_enabled(tcx) {
+        let _ = tcx.mir_inliner_callers_count_map(());
```
cjgillot, line 601: `mir_inliner_callers_count_map` is whole-crate analysis. By calling it here, the computation of `optimized_mir` depends on this whole-crate analysis, and needs to be recomputed each time *any* caller count is changed in the crate. This dependency needs to be threaded through a `mir_inliner_callers_count(def_id)` query to avoid this invalidation.

compiler/rustc_mir_transform/src/inline.rs
```diff
@@ -351,9 +365,14 @@ impl Inliner<'tcx> {
         if callee_body.basic_blocks().len() <= 3 {
             threshold += threshold / 4;
         }
+
+        // Give a huge bonus to private functions called only once in current module.
+        if self.tcx.mir_inliner_callers_count_map(()).get(&callsite.callee.def_id()) == Some(&1) {
```
cjgillot, line 370: Likewise, we should use some kind of  `mir_inliner_callers_count(callside.callee.def_id())` here.

compiler/rustc_mir_transform/src/inline.rs
```diff
@@ -964,3 +983,41 @@ impl<'a, 'tcx> MutVisitor<'tcx> for Integrator<'a, 'tcx> {
...
+crate fn mir_inliner_callers_count_map<'tcx>(
+    tcx: TyCtxt<'tcx>,
+    _: (),
+) -> Lrc<FxHashMap<DefId, usize>> {
+    let mut callers_count_map: FxHashMap<DefId, usize> = Default::default();
+
+    // Iterate over all body owners of current crate
+    for did in tcx
+        .hir()
+        .body_owners()
```
cjgillot, line 996: Should we use `mir_keys` here?

compiler/rustc_mir_transform/src/inline.rs
```diff
@@ -964,3 +983,41 @@ impl<'a, 'tcx> MutVisitor<'tcx> for Integrator<'a, 'tcx> {
...
+            .iter_enumerated()
+            .filter_map(|(bb, bb_data)| {
+                // Try to find call site in this basic block.
+                CallSite::resolve_callsite(tcx, param_env, &body, bb, bb_data)
+            })
+            // Get function def_id which are we calling.
+            .map(|callsite| callsite.callee.def_id())
+            .for_each(|did| {
+                *callers_count_map.entry(did).or_insert(0) += 1;
+            });
```
cjgillot, line 1019: Could you rewrite this combinator chain as a simple `for` loop? I find it a little hard to follow.

compiler/rustc_mir_transform/src/inline.rs
```diff
@@ -964,3 +983,41 @@ impl<'a, 'tcx> MutVisitor<'tcx> for Integrator<'a, 'tcx> {
...
 }
+
+crate fn mir_inliner_callers_count_map<'tcx>(
+    tcx: TyCtxt<'tcx>,
+    _: (),
+) -> Lrc<FxHashMap<DefId, usize>> {
+    let mut callers_count_map: FxHashMap<DefId, usize> = Default::default();
+
+    // Iterate over all body owners of current crate
+    for did in tcx
```
cjgillot, line 994: Please prefer `def_id` to `did` as an abbreviation.

compiler/rustc_middle/src/query/mod.rs
```diff
@@ -827,6 +827,11 @@ rustc_queries! {
         }
     }
 
+    query mir_inliner_callers_count_map(_: ()) -> Lrc<FxHashMap<DefId, usize>> {
+        fatal_cycle
```
cjgillot, line 831: Is this modifier useful?

compiler/rustc_middle/src/query/mod.rs
```diff
@@ -827,6 +827,11 @@ rustc_queries! {
         }
     }
 
+    query mir_inliner_callers_count_map(_: ()) -> Lrc<FxHashMap<DefId, usize>> {
```
cjgillot, line 830: This query only looks at the current crate, so is can only store `LocalDefId`s.

cjgillot (commented):
Thanks for the PR @In-line. I left a few comments on the implementation.

About the motivation: I understand that single-use functions are in general good inlining candidates. However, there may be other situations where we would prefer to inline the caller into downstream crates rather than inlining the single-use callee. Should there be some kind of heuristic for that?

compiler/rustc_mir_transform/src/lib.rs
In-line: I'm not sure, what kind of dependency should be passed here. 

I mean `optimized_mir` really starts to depend on the number of calls in the functions, but that's the whole point of this PR. 

I can try to add per `def_id` computation of the number of callers, but it will turn algorithm from O(N) to O(N^2), because it will iterate over crate for every function in the crate.

I can add this add additional indirection `mir_inliner_callers_count(def_id) -> mir_inliner_callers_count_map(()) -> ...`, but will it give any difference? 

I mean `mir_inliner_callers_count(def_id)` will be still invalidated, when any caller count is changed.

I can add one more additional layer to break the dependency and recompute only what's needed.
`mir_inliner_crate_callers_count(def_id) -> mir_keys() -> mir_inliner_callers_count(key_def_id, def_id) `

Would you prefer the last option?

cjgillot: I agree that we want to modify the optimization of a function `f` depending on the number of calls to `f`. I want to avoid having to redo the optimization because of a change to the number of calls to `g`.

I mean I'd rather have `optimized_mir(def_id) -> mir_inliner_callers_count(def_id) -> mir_inliner_callers_count_map(())`. That way, if `mir_inliner_callers_count_map` changes for an unrelated `def_id2`, the query system will recompute `mir_inliner_callers_count(def_id)`, which is cheap. The query system will see that `mir_inliner_callers_count(def_id)` has not changed. Therefore `optimized_mir` won't be recomputed.

compiler/rustc_mir_transform/src/inline.rs
In-line: `mir_keys` adds introduces for type check.

compiler/rustc_middle/src/query/mod.rs
In-line: Callees can be from other crates too.

compiler/rustc_mir_transform/src/lib.rs
```diff
@@ -165,6 +170,20 @@ fn mir_keys(tcx: TyCtxt<'_>, (): ()) -> FxHashSet<LocalDefId> {
     set
 }
 
+/// Returns array of all the function or closures `DefId`s in this crate that have MIR
+fn mir_fn_and_closures_keys<'tcx>(tcx: TyCtxt<'tcx>, _: ()) -> &'tcx [LocalDefId] {
```
cjgillot, line 174: Why do you need to introduce this other query?

compiler/rustc_mir_transform/src/inline.rs
```diff
@@ -964,3 +975,39 @@ impl<'a, 'tcx> MutVisitor<'tcx> for Integrator<'a, 'tcx> {
...
+        // Iterate over all functions and closures of current crate
+        .mir_fn_and_closures_keys(())
+        .iter()
+        // Get all calless in crate functions
+        .map(|def_id| {
+            tcx.mir_inliner_callees(ty::InstanceDef::Item(ty::WithOptConstParam::unknown(
+                def_id.to_def_id(),
+            )))
+        })
+        .flatten()
```
cjgillot, line 996: Please prefer nested loops.

compiler/rustc_mir_transform/src/inline.rs
```diff
@@ -351,9 +355,16 @@ impl Inliner<'tcx> {
         if callee_body.basic_blocks().len() <= 3 {
             threshold += threshold / 4;
         }
+
+        // Give a huge bonus to local, unexported functions called only once in current module.
+        if let Some(callee_def_id) = callsite.callee.def_id().as_local() {
+            if (self.tcx.mir_inliner_crate_fn_unexported_callees_count(callee_def_id)) == Some(1) {
```
cjgillot, line 361: Nit: extra parentheses around the function call.

compiler/rustc_mir_transform/src/inline.rs
```diff
@@ -351,9 +355,16 @@ impl Inliner<'tcx> {
...
+                threshold *= 2;
```
cjgillot, line 362: Does this factor need to be tuned somehow?
cc @wesleywiser

wesleywiser: I suspect we will probably need to tune it yeah, but it also seems ok to land as-is since this pass isn't enabled by default.

compiler/rustc_mir_transform/src/lib.rs
In-line: To cache it's results. 

Number of calls can change inside some function `f(x)`, but that doesn't mean the count of functions will change.

compiler/rustc_mir_transform/src/inline.rs
In-line: Do we have style guide for situations like this? I wasn't able to find any.

src/test/mir-opt/inline/inline_single_private.main.Inline.diff
```diff
@@ -0,0 +1,111 @@
...
+                                           // + literal: Const { ty: fn() {more_than_one_call}, val: Value(Scalar(<ZST>)) }
+      }
++ }
+  
+-     bb2: {
+-         StorageDead(_2);                 // scope 0 at $DIR/inline-single-private.rs:22:25: 22:26
+-         _0 = const ();                   // scope 0 at $DIR/inline-single-private.rs:20:11: 23:2
+-         return;                          // scope 0 at $DIR/inline-single-private.rs:23:2: 23:2
+-     }
++ alloc5 (size: 0, align: 8) {}
```
In-line, line 102: How can I exclude 32 bit targets from the test? This seems to be the reason test fails.

compiler/rustc_mir_transform/src/inline.rs
cjgillot: My personal rule is to avoid mixing functional and imperative style. This loop is imperative because it mutates `callers_count_map`, so I find an explicit loop and branches much easier to reason with.

compiler/rustc_mir_transform/src/lib.rs
cjgillot: `mir_inliner_crate_fn_callees_count_map` is a global analysis of all functions in the crate. We pretty much expect to recompute during each compilation session. `mir_inliner_crate_fn_callees_count` is already there to shield downstream code from invalidations.
If you want to avoid spurious invalidations, I'd recommend listing definitions by using `mir_keys` followed by the filter: `tcx.hir().body_owners` creates a dependency on the full HIR, so may get invalidated by span changes. `mir_keys` does not have this issue.

src/test/mir-opt/inline/inline_single_private.main.Inline.diff
cjgillot: You have two solutions:
1. add EMIT_MIR_FOR_EACH_BIT_WIDTH to the test and bless both 32bit and 64bit tests;
2. rewrite the contents of the functions to use regular instructions and no promoted constants (and adjust the threshold accordingly).

JohnCSimon:
Ping from triage: 
@In-line - what is the status of this PR? can you address the build failures?

JohnCSimon:

@rustbot label: +S-inactive

compiler/rustc_mir_transform/src/inline.rs
In-line: Factors currently are just with gut feeling. Feel free to suggest better or well known values.

compiler/rustc_mir_transform/src/lib.rs
In-line: @cjgillot so I should inline this function to where it is used? Did I understood correctly?

compiler/rustc_mir_transform/src/inline.rs
In-line: I changed it to full functional style 😄

In-line:
@JohnCSimon Please, kindly reopen the PR

In-line:
ping @JohnCSimon

r00ster91:
@In-line Seeing that you are the original author of this PR I would imagine that you are able to reopen this PR yourself with a "Reopen pull request" button?

In-line:
@r00ster91 ¯\\_(ツ)_/¯

Screen Shot 2021-12-22 at 14 58 43

In-line:
ping @JohnCSimon

bjorn3:
Looks like you force-pushed the PR branch, so github doesn't allow re-opening even for people with write permission. You will need to force-push the commit as of closing this PR to allow it to be re-opened.

image

PR #89466, opened 2021-10-02
Fix bug with query modifier parsing

Mark-Simulacrum:
The previous macro_rules! parsers failed when an additional modifier was added
with ambiguity errors. The error is pretty unclear as to what exactly the cause
here is, but this change simplifies the argument parsing code such that the
error is avoided.

Extracted from other work, and somewhat duplicates 0358edeb5 from #85830, but
this approach seems a little simpler to me. Not technically currently necessary but seems
like a good cleanup.

### DIFF
--- a/compiler/rustc_macros/src/query.rs
+++ b/compiler/rustc_macros/src/query.rs
@@ -455,28 +455,28 @@ pub fn rustc_queries(input: TokenStream) -> TokenStream {
 
         // Pass on the fatal_cycle modifier
         if let Some(fatal_cycle) = &modifiers.fatal_cycle {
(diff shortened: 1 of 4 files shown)


bjorn3:
> Extracted from other work, and somewhat duplicates 0358ede from #85830, but
this approach seems a little simpler to me. Not technically currently necessary but seems
like a good cleanup.

Maybe your diff is smaller, but I think my commit is less likely to cause issues in the long run by making every modifier a token tree instead of either an ident or an ident followed by a group.

Mark-Simulacrum:
Yeah, I guess that's true. I rebased your commit.

PR #89471, opened 2021-10-02
Use Ancestory to check default fn in const impl instead of comparing idents

nbdd0121:
Fixes https://rust-lang.zulipchat.com/#narrow/stream/146212-t-compiler.2Fconst-eval/topic/Const.20trait.20impl.20inside.20macro

### DIFF
--- a/compiler/rustc_passes/src/check_const.rs
+++ b/compiler/rustc_passes/src/check_const.rs
@@ -8,7 +8,6 @@
 //! through, but errors for structured control flow in a `const` should be emitted here.
 
 use rustc_attr as attr;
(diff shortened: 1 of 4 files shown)


cjgillot:
Using Symbols completely ignores macro hygiene. In which extent can it be ignored? How does rustc_typeck check impls are complete?

nbdd0121:
I was under the impression that method names are not hygenic, but then I realized it not be true for macro 2.0.

So a proper comparision would involve calling `tcx.hygienic_eq`. The typeck code uses `Ancestory` to find the defining node of a particular method (which internally uses `tcx.hygienic_eq`), so I updated the code to use `Ancestory` as well.

Issue #89475, opened 2021-10-02
Splitting up `compiletest::runtest`


Nicholas-Baron:
Subtask to #60302.

- [ ] RunPassValgrind is practically useless (valgrind is not installed on CI, and the tests are just run in run-pass mode). The right approach here is likely to refactor it as a modifier atop the regular run-pass tests -- it looks like the code generally just mirrors the RunPass mode, other than setting the runtool to valgrind. That mode can be enabled in the configuration of the test with `// use-valgrind` or something like that.
- [x] Move diff computation out into a separate module (DiffLine, Mismatch, make_diff, ...)
- [x] Move read2_abbreviated to the read2.rs module
- [x] Move DebuggerCommands and their parsing out to a separate file, alongside check_debugger_output (which is similarly parsing code).

I suspect this set of changes is likely harder, and might not get to <3000 lines; there's likely more cleanup work possible, though. This module has grown pretty organically over the last ~10 years -- there's a lot of small bits that can be adjusted to make things better. There's probably also subtle inconsistencies between various functions that would likely be nice to clean up and unify, and filing PRs for those would be helpful -- likely some care is needed, perhaps even some small test updates, but overall a good goal.

You might also be interested in https://github.com/rust-lang/rust/issues/40713, as a somewhat larger project.

_Originally posted by @Mark-Simulacrum in https://github.com/rust-lang/rust/issues/89240#issuecomment-932658341_

Mark-Simulacrum:
I think there's room for unifying the two -- we already have some support for running tests under a wrapper for e.g. android testing, so I imagine we might be able to reuse parts of that perhaps.

jieyouxu:
I really need to make sure we split up `runtest.rs`, each time I look at it the more cursed it becomes.

PR #89477, opened 2021-10-02
Move items related to computing diffs to a separate file

Nicholas-Baron:
Work towards #89475.

### DIFF
--- a/src/tools/compiletest/src/compute_diff.rs
+++ b/src/tools/compiletest/src/compute_diff.rs
@@ -1,4 +1,5 @@
 use std::collections::VecDeque;
+use std::fs::{File, FileType};
 use std::path::Path;
 
 #[derive(Debug, PartialEq)]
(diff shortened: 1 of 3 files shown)


Mark-Simulacrum:
There's some diff-related code here as well - https://github.com/rust-lang/rust/blob/cb501a6fe6cdb1928611b7193dec96f3d7bd2293/src/tools/compiletest/src/runtest.rs#L2404

I'm not sure how much that is "the same" or can be unified, haven't looked too closely. What do you think?

Nicholas-Baron:
@Mark-Simulacrum While this does seem diff-related, its structure is very different from `write_diff` and `make_diff`, suggesting some semantic differences. This makes unification a task I think beyond this PR.

It may be possible to extract some parts as separate functions, as `self` is not reference heavily in them. However, this may be beyond the scope of this PR as well. The following code block is an example. https://github.com/rust-lang/rust/blob/cb501a6fe6cdb1928611b7193dec96f3d7bd2293/src/tools/compiletest/src/runtest.rs#L2406-L2431

Mark-Simulacrum:
Hm. Ok. Well, I guess we can add a "why do we have two different diff implementations" (unified-diff and diff crates) to the list of open questions...

But regardless I think much of the code *can* be moved into this module, if not unified just yet -- it looks pretty separate from being *rustdoc* specific to me, beyond the minor filter on extensions (which seems easy to pass in).

PR #89478, opened 2021-10-02
Fixed numerus of error message

zvavybir:
When there are redundant trait requirements and these are hidden, a message is generated by the following code snippet:
`format!("{} redundant requirements hidden", count)`
But if there is only a single hidden requirement, it will still print this message in plural instead of singular.

### DIFF
--- a/compiler/rustc_trait_selection/src/traits/error_reporting/suggestions.rs
+++ b/compiler/rustc_trait_selection/src/traits/error_reporting/suggestions.rs
@@ -10,7 +10,9 @@ use crate::traits::normalize_projection_type;
 use rustc_data_structures::fx::FxHashSet;
 use rustc_data_structures::stack::ensure_sufficient_stack;
 use rustc_data_structures::sync::Lrc;
-use rustc_errors::{error_code, struct_span_err, Applicability, DiagnosticBuilder, Style};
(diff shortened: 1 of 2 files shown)


compiler/rustc_trait_selection/src/traits/error_reporting/suggestions.rs
```diff
@@ -2272,7 +2272,15 @@ impl<'a, 'tcx> InferCtxtExt<'tcx> for InferCtxt<'a, 'tcx> {
...
-                if count > 0 {
+                if count == 1 {
+                    err.note(&format!("1 redundant requirement hidden"));
+                    err.note(&format!(
+                        "required because of the requirements on the impl of `{}` for `{}`",
+                        parent_trait_ref.print_only_trait_path(),
+                        parent_trait_ref.skip_binder().self_ty()
+                    ));
+                }
+                if count > 1 {
```
klensy, line 2283: There exist `pluralize!` macro for it

zvavybir: Oh, I didn't knew that.

klensy:
4 commits, squash maybe?

PR #89487, opened 2021-10-03
Try to recover from a `=>` -> `=` or `->` typo in a match arm

### LINKED ISSUE #89396: `=>` to `=` typo in match arm results in lots of parse errors
A typo of `=>` to `=` in a match arm currently results in a lot of cascading errors.

```
error: expected one of `!`, `(`, `...`, `..=`, `..`, `::`, `=>`, `if`, `{`, or `|`, found `=`
   --> compiler/rustc_trait_selection/src/traits/error_reporting/mod.rs:244:39
    |
244 |             SelectionError::Ambiguous = {
    |                                       ^ expected one of 10 possible tokens
```

We should gracefully handle some cases like `>` and `=` instead of `=>` if followed by a `{` in a match arm.

FabianWolff:
Fixes #89396.

### DIFF
--- a/compiler/rustc_parse/src/parser/expr.rs
+++ b/compiler/rustc_parse/src/parser/expr.rs
@@ -2322,7 +2322,21 @@ impl<'a> Parser<'a> {
                 None
             };
             let arrow_span = this.token.span;
-            this.expect(&token::FatArrow)?;
+            if let Err(mut err) = this.expect(&token::FatArrow) {
+                // We might have a `=>` -> `=` or `->` typo (issue #89396).
+                if let token::Eq | token::RArrow = this.token.kind {
+                    err.span_suggestion(
+                        this.token.span,
+                        "try using a fat arrow here",
+                        "=>".to_string(),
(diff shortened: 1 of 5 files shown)


compiler/rustc_parse/src/parser/expr.rs
```diff
@@ -2322,7 +2322,21 @@ impl<'a> Parser<'a> {
                 None
             };
             let arrow_span = this.token.span;
-            this.expect(&token::FatArrow)?;
+            if let Err(mut err) = this.expect(&token::FatArrow) {
+                // We might have a `=>` -> `=` or `->` typo (issue #89396).
+                if let token::Eq | token::RArrow = this.token.kind {
```
petrochenkov, line 2327: There's a thing called `TokenKind::similar_tokens`, could you use it here?
(Ideally it would be integrated to `expect`, so errors like this would be reported automatically.)

FabianWolff: Fixed; thanks for the review!

PR #89491, opened 2021-10-03
Add `impl From<Ordering> for i32`

lopopolo:
This commit adds a first-class `From` conversion of `Ordering` into `i32`.

`Ordering` is defined to be compatible with the return types of the various `...cmp...` routines in the C standard library, but the only way to convert an `Ordering` to an `i32` is with a non-obvious `as` cast.

This commit adds a proper API for doing this and hides a "scary" `as` cast inside the `core` library.

### DIFF
--- a/library/core/src/cmp.rs
+++ b/library/core/src/cmp.rs
@@ -568,6 +568,18 @@ impl Ordering {
     }
 }
 
+#[stable(feature = "convert_ordering_into_i32", since = "1.57.0")]
+impl From<Ordering> for i32 {
+    /// Convert an [`Ordering`] into a signed 32-bit integer.
+    ///
+    /// This function converts an `Ordering` into an integer compatible with C
+    /// FFI. This `i32` value is compatible with the return value of C library
+    /// functions such as `strncmp`.
+    fn from(ordering: Ordering) -> Self {
+        ordering as i32
+    }
+}
+
 /// A helper struct for reverse ordering.
 ///
 /// This struct is a helper to be used with functions like [`Vec::sort_by_key`] and


joshtriplett:
Tagging libs-api.

This would be insta-stable, so it needs team review.

@lopopolo Where is `Ordering` documented to be compatible with the return value of `strcmp`/`memcmp`/etc? Converting an `Ordering` to such a value doesn't seem like a particularly common use case.

lopopolo:
> Where is `Ordering` documented to be compatible with the return value of `strcmp`/`memcmp`/etc?

😅 I didn't say this was documented. I meant to say that `Ordering` is defined in the source to have the same semantics as the `int` return value/sign (-1/0/1) of these functions.

I'm implementing several "cmp" style routines as part of an oxidation project and being able to turn a nicely typed `Ordering` into the `int` expected at the FFI boundary would be something I'd like to have.

lopopolo:
> Converting an `Ordering` to such a value doesn't seem like a particularly common use case.

I was of the impression that such a conversion being common was not a requirement for adding `From` impls to `core` and `std`. I think I have a fairly clear use case and such a `From` impl would remove a little bit of friction here for little cost.

joshtriplett:
It's not that it needs to be common, it's that the usefulness needs to outweigh the potential for an error-prone conversion. `.into()` is non-specific, and can sometimes result in an unexpected conversion.

There's a proposal in https://github.com/rust-lang/rust/pull/81642 to add `derive(FromRepr)` and automatically derive `AsRepr`, which would serve the same purpose, and I think that'd be more self-explanatory.

I would propose that we merge https://github.com/rust-lang/rust/pull/81642 (in some form) and then `Ordering` will gain an `as_repr()` method for this.

lopopolo:
> There's a proposal in #81642 to add `derive(FromRepr)` and automatically derive `AsRepr`, which would serve the same purpose, and I think that'd be more self-explanatory.

ooooooh I didn't know about this proposal. That looks like it would do the trick. One sticking point is it doesn't look like `Ordering` has a `#[repr(...)]` attribute:

https://github.com/rust-lang/rust/blob/e737694a4d66b01308b73d4559a35b43e414faf9/library/core/src/cmp.rs#L324-L336

> I would propose that we merge #81642 (in some form) and then `Ordering` will gain an `as_repr()` method for this.

This sounds good to me. Do we need to cut another ticket to add a `repr` to ordering?

joshtriplett:
@lopopolo Yes, please do submit a PR adding `repr(i8)` to `Ordering`, and I'd be happy to FCP it with libs.

Also, I submitted https://github.com/rust-lang/nomicon/pull/315 to clarify that adding such a `repr` should not cause a problem with the actual size of the type.

lopopolo:
thanks for the help @joshtriplett. The new PR is here:

- https://github.com/rust-lang/rust/pull/89507

PR #89495, opened 2021-10-03
Add two inline annotations for hot functions

Mark-Simulacrum:
These two functions are essentially no-ops (and compile to just a load and
return), but show up in process_obligations profiles with a high call count --
so worthwhile to try and inline them. This is not normally possible as they're
non-generic, so they don't get offered for inlining by our current algorithm.


Mark-Simulacrum:
r? rust-lang/compiler-team

michaelwoerister:
Nice find!

@bors r+

PR #89499, opened 2021-10-03
Split out LLVM PGO step and use clang 13 to compile LLVM

Mark-Simulacrum:
We're seeing a PGO version mismatch error in CI logs:

    LLVM Profile Error: Runtime and instrumentation version mismatch : expected 5, but get 7

which is likely due to the version bumped here differing from that used by
rustc.

This PR fixes this by splitting out the PGO step for LLVM into a separate phase of the pgo.sh script, which nets no change to performance (see [these results](https://perf.rust-lang.org/compare.html?start=c34ac8747ca96d09cb08b8f5adddead826e77c06&end=e272c2af45f40c74dab83948235903ffbe3ad57f)). Then, it follows that up with an upgrade to LLVM/clang version 13 as our bootstrap compiler, which yields the performance improvements for this PR -- around 5%. This depends on the first step here, because otherwise we end up somehow clobbering or otherwise hurting our ability to effectively collect performance data, yielding reductions in performance for a subset of benchmarks -- it is not clear what the cause here was precisely, but the split only costs ~10 minutes and seems worthwhile.

### DIFF
--- a/src/ci/docker/host-x86_64/dist-x86_64-linux/build-clang.sh
+++ b/src/ci/docker/host-x86_64/dist-x86_64-linux/build-clang.sh
@@ -4,7 +4,7 @@ set -ex
 
 source shared.sh
 
-LLVM=llvmorg-12.0.1
+LLVM=llvmorg-13.0.0
 
 mkdir llvm-project
 cd llvm-project
--- a/src/ci/pgo.sh
+++ b/src/ci/pgo.sh
@@ -4,10 +4,37 @@ set -euxo pipefail
 
 rm -rf /tmp/rustc-pgo
 
+# We collect LLVM profiling information and rustc profiling information in
+# separate phases. This increases build time -- though not by a huge amount --
+# but prevents any problems from arising due to different profiling runtimes
+# being simultaneously linked in.
+
 python3 ../x.py build --target=$PGO_HOST --host=$PGO_HOST \
     --stage 2 library/std \
-    --rust-profile-generate=/tmp/rustc-pgo \
     --llvm-profile-generate
+
+# Profile libcore compilation in opt-level=0 and opt-level=3
+RUSTC_BOOTSTRAP=1 ./build/$PGO_HOST/stage2/bin/rustc --edition=2018 \
+    --crate-type=lib ../library/core/src/lib.rs
+RUSTC_BOOTSTRAP=1 ./build/$PGO_HOST/stage2/bin/rustc --edition=2018 \
+    --crate-type=lib -Copt-level=3 ../library/core/src/lib.rs
+
+# Merge the profile data we gathered for LLVM
+# Note that this uses the profdata from the clang we used to build LLVM,
+# which likely has a different version than our in-tree clang.
+/rustroot/bin/llvm-profdata \
+    merge -o /tmp/llvm-pgo.profdata ./build/$PGO_HOST/llvm/build/profiles
+
+# Rustbuild currently doesn't support rebuilding LLVM when PGO options
+# change (or any other llvm-related options); so just clear out the relevant
+# directories ourselves.
+rm -r ./build/$PGO_HOST/llvm ./build/$PGO_HOST/lld
+
+# Okay, LLVM profiling is done, switch to rustc PGO.
+
+python3 ../x.py build --target=$PGO_HOST --host=$PGO_HOST \
+    --stage 2 library/std \
+    --rust-profile-generate=/tmp/rustc-pgo
 
 # Profile libcore compilation in opt-level=0 and opt-level=3
 RUSTC_BOOTSTRAP=1 ./build/$PGO_HOST/stage2/bin/rustc --edition=2018 \
@@ -46,18 +73,13 @@ cd /checkout/obj
 ./build/$PGO_HOST/llvm/bin/llvm-profdata \
     merge -o /tmp/rustc-pgo.profdata /tmp/rustc-pgo
 
-# Merge the profile data we gathered for LLVM
-# Note that this uses the profdata from the clang we used to build LLVM,
-# which likely has a different version than our in-tree clang.
-/rustroot/bin/llvm-profdata \
-    merge -o /tmp/llvm-pgo.profdata ./build/$PGO_HOST/llvm/build/profiles
-
 # Rustbuild currently doesn't support rebuilding LLVM when PGO options
 # change (or any other llvm-related options); so just clear out the relevant
 # directories ourselves.
 rm -r ./build/$PGO_HOST/llvm ./build/$PGO_HOST/lld
 
-# This produces the actual final set of artifacts.
+# This produces the actual final set of artifacts, using both the LLVM and rustc
+# collected profiling data.
 $@ \
     --rust-profile-use=/tmp/rustc-pgo.profdata \
     --llvm-profile-use=/tmp/llvm-pgo.profdata


Swatinem:
> We're seeing a PGO version mismatch error in CI logs

Sorry for the driveby question:

Is it possible the same can manifest as a `malformed instrumentation profile data` error when trying to use `llvm-profdata merge`?  I have seen that `-Z instrument-coverage` has been broken for some time in nightly, but didn’t invest the time yet to fully diagnose the cause of that.

Mark-Simulacrum:
Not sure. Coverage instrumentation goes through pretty different paths as far as I know, I don't think it should be related to this PR.

joshtriplett:
Since we currently cache LLVM in our CI and don't rebuild it every time, would it potentially make sense to just always rebuild our LLVM using itself, so that it always matches?

joshtriplett:
Reviewing the performance results, this looks like a massive improvement.

Mark-Simulacrum:
> Since we currently cache LLVM in our CI and don't rebuild it every time, would it potentially make sense to just always rebuild our LLVM using itself, so that it always matches?

We rebuild LLVM twice each CI build right now, because https://github.com/mozilla/sccache/pull/952 hasn't been merged/finished. If the docker cache is invalidated, that adds a third build.

We can adjust the in-docker build to use the same LLVM as in our fork, I suppose, but I'm not sure if that's necessary - we can probably just apply patches like this PR on future LLVM upgrades.

nikic:
> > Since we currently cache LLVM in our CI and don't rebuild it every time, would it potentially make sense to just always rebuild our LLVM using itself, so that it always matches?
> 
> We rebuild LLVM twice each CI build right now, because [mozilla/sccache#952](https://github.com/mozilla/sccache/pull/952) hasn't been merged/finished. If the docker cache is invalidated, that adds a third build.
> 
> We can adjust the in-docker build to use the same LLVM as in our fork, I suppose, but I'm not sure if that's necessary - we can probably just apply patches like this PR on future LLVM upgrades.

One thing I'm concerned about is that the current setup will result in spurious PGO related regressions with future LLVM updates, as we typically update our LLVM before we update the clang used to build it.

Mark-Simulacrum:
I'm not sure how to check, but my current theory is that the optimization here is pretty much entirely due to better optimized LLVM, not PGO. The PGO PR for LLVM did produce a speedup on merge despite these "version mismatch errors".

Our setup is such that the LLVM PGO data is configured to go to a separate location from the rustc data, and we use the right llvm-profdata for each to merge it. So I think the warning/error message is *probably* spurious, perhaps arising from LLVM's profiling runtime not supporting being dlopen'd while already linked in, or something like that...

In other words, at least as far as I can tell, future LLVM upgrades are likely to reintroduce the error message until we bump our c toolchain, but there's no evidence this will cause performance regressions at this time.

nikic:
Right, it's hard to distinguish whether the impact is due to Clang 13 doing a better job of optimizing, or something about the handling of profile data being fixed. What's interesting is that if you filter the perf result by check builds only, this is a universal small regression. I would have expected that check builds wouldn't be affected by optimizations applied to LLVM.

Mark-Simulacrum:
Overlap in profiles from 9dbb26efe8a886f7bba94c2b996c9535ce07f917 and its parent commit (last "noop" change I found):

```text
$ llvm-profdata overlap before-rustc-pgo.profdata after-rustc-pgo.profdata
Profile overlap infomation for base_profile: before-rustc-pgo.profdata and test_profile: after-rustc-pgo.profdata
Program level:
  # of functions overlap: 260502
  Edge profile overlap: 99.910%
  Edge profile base count sum: 230867115195
  Edge profile test count sum: 231072749498
  IndirectCall profile overlap: 99.996%
  IndirectCall profile base count sum: 722318628
  IndirectCall profile test count sum: 722335980
  MemOP profile overlap: 99.981%
  MemOP profile base count sum: 488656567
  MemOP profile test count sum: 488639611

$ llvm-profdata overlap before-llvm-pgo.profdata after-llvm-pgo.profdata
Profile overlap infomation for base_profile: before-llvm-pgo.profdata and test_profile: after-llvm-pgo.profdata
Program level:
  # of functions overlap: 96161
  Edge profile overlap: 99.742%
  Edge profile base count sum: 1054888702705
  Edge profile test count sum: 1050720657781
  IndirectCall profile overlap: 99.894%
  IndirectCall profile base count sum: 8906460397
  IndirectCall profile test count sum: 8906009877
  MemOP profile overlap: 99.921%
  MemOP profile base count sum: 5927918788
  MemOP profile test count sum: 5920677320
```

And for this PR's try commit:

```
$ llvm-profdata overlap before-rustc-pgo.profdata after-rustc-pgo.profdata
Profile overlap infomation for base_profile: before-rustc-pgo.profdata and test_profile: after-rustc-pgo.profdata
Program level:
  # of functions overlap: 260504
  # of functions only in test_profile: 89905
  Edge profile overlap: 40.433%
  Percentage of Edge profile only in test_profile: 59.566%
  Edge profile base count sum: 230860056427
  Edge profile test count sum: 571155717471
  IndirectCall profile overlap: 18.746%
  Percentage of IndirectCall profile only in test_profile: 81.254%
  IndirectCall profile base count sum: 722242054
  IndirectCall profile test count sum: 3853524645
  MemOP profile overlap: 21.263%
  Percentage of MemOP profile only in test_profile: 78.737%
  MemOP profile base count sum: 488641949
  MemOP profile test count sum: 2298031353

$ llvm-profdata overlap before-llvm-pgo.profdata after-llvm-pgo.profdata
Profile overlap infomation for base_profile: before-llvm-pgo.profdata and test_profile: after-llvm-pgo.profdata
Program level:
  # of functions overlap: 84177
  # of functions mismatch: 6928
  # of functions only in test_profile: 2696
  Edge profile overlap: 46.840%
  Mismatched count percentage (Edge): 52.323%
  Percentage of Edge profile only in test_profile: 0.685%
  Edge profile base count sum: 1047797180915
  Edge profile test count sum: 1153062758787
  IndirectCall profile overlap: 40.414%
  Mismatched count percentage (IndirectCall): 59.122%
  Percentage of IndirectCall profile only in test_profile: 0.310%
  IndirectCall profile base count sum: 8901901446
  IndirectCall profile test count sum: 8877805375
  MemOP profile overlap: 84.344%
  Mismatched count percentage (MemOP): 15.542%
  Percentage of MemOP profile only in test_profile: 0.001%
  MemOP profile base count sum: 5916707088
  MemOP profile test count sum: 5921229195
```

So it does look like the version mismatch presumably had a not insignificant effect, though the results are a bit weird. Presumably this relates to the check builds as well.

Maybe we're not successfully putting LLVM results in one file and rustc in another (e.g., race which one loads first or something)? Worth figuring it out, probably... though not sure how to investigate that.

Mark-Simulacrum:
Okay, so that try build fixes the "LLVM Profile error" message by doing LLVM PGO collection and rustc PGO collection in separate phases of the build. This just adds ~10 minutes to our build so is pretty cheap -- but it demonstrates that fixing the error doesn't seem to have any major impact. (FWIW, that also only collects information of the LLVM compilation from libcore, and based on these results that seems to be sufficient or better).

I've now readded the LLVM 13 bootstrap (i.e., building with clang v13 for all C/C++ compilation we do). Presumably this has less chance of somehow creating interference between the two profiling runtimes now, so maybe we'll see less weird noise in the rustc results.

@bors try @rust-timer queue

Issue #89493, opened 2021-10-03
Lower/UpperExp inconsistent between integers and floats


nrabulinski:
I tried this code:

```rust
println!("{:.*e}", 0, 25);
println!("{:.*e}", 0, 25.0);
```

I expected to see this happen: 
The output should be the same regardless if I'm formatting an integer or a float, in this case most likely `3e1`. 

Instead, this happened: 
```
3.e1
2e1
```

Happens in both latest stable and nightly.

EDIT: I'm suspecting this is actually two separate bugs in both the integer and float implementation - the integer result contains a `.` which should not be there, and the float gets rounded down to `2` despite `25.1` rounding to `3e1` - which is the result I would expect for `25.0` as well.

joshtriplett:
For the floating-point portion of this, I *think* the default is "round to nearest even", which would make 2e1 the correct rounding.

It seems odd that integers and floating-point values have different rounding here.

nrabulinski:
> For the floating-point portion of this, I *think* the default is "round to nearest even", which would make 2e1 the correct rounding.
> 
> 
> 
> It seems odd that integers and floating-point values have different rounding here.

https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=a056ffde1e5a237e2e26351615775809

fmease:
I can no longer reproduce this on the latest nightly (1.77, 2024-01-22 d5fd0997291ca0135401), I get:

```
2e1
2e1
```

Interestingly, stable (1.75) produces the following which still contains the rounding error:

```
3e1
2e1
```

Closing as completed. Please ping me if I should reopen this issue.

PR #89502, opened 2021-10-03
Fix Lower/UpperExp formatting for integers and precision zero

FabianWolff:
Fixes the integer part of #89493 (I daren't touch the floating-point formatting code). The issue is that the "subtracted" precision essentially behaves like extra trailing zeros, but this is not currently reflected in the code properly.

### DIFF
--- a/library/core/tests/fmt/num.rs
+++ b/library/core/tests/fmt/num.rs
@@ -146,7 +146,7 @@ fn test_format_int_exp_precision() {
     assert_eq!(format!("{:.1000e}", 1), format!("1.{}e0", "0".repeat(1000)));
     //test zero precision
     assert_eq!(format!("{:.0e}", 1), format!("1e0",));
-    assert_eq!(format!("{:.0e}", 25), format!("3e1",));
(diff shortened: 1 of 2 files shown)


joshtriplett:
While this fixes the trailing `.`, I'm not sure if this is the intended rounding. It's inconsistent with floating-point values, which (I believe) default to "round to nearest even".

library/core/tests/fmt/num.rs
```diff
@@ -146,6 +146,7 @@ fn test_format_int_exp_precision() {
     assert_eq!(format!("{:.1000e}", 1), format!("1.{}e0", "0".repeat(1000)));
     //test zero precision
     assert_eq!(format!("{:.0e}", 1), format!("1e0",));
+    assert_eq!(format!("{:.0e}", 25), format!("3e1",));
```
joshtriplett, line 149:
```suggestion
    assert_eq!(format!("{:.0e}", 35), format!("4e1",));
```

joshtriplett:
r=me with the test changed.

PR #89506, opened 2021-10-04
librustdoc: Use correct heading levels.

### LINKED ISSUE #89309: Use appropriate heading levels in docblocks
Following up on https://github.com/rust-lang/rust/issues/87059 (big thanks @ahicks92 for the detailed feedback!):

> [...] Rustdoc didn't do h tags right. It's not exactly systemic, but it's something like (level + 3) % 6, where what should be heading levels 4-6 (examples, etc) become heading level 1 and 2. This is confusing, but once you learn things like impl blocks tend to be heading level 3 and functions are usually level 4, you effectively have a "jump to next function" keystroke and so on.

Here's what's happening: in Markdown `#` generally represents an `h1`, `##` represents `h2`, and so on. That means each docblock often has `h1`s and `h2`s in them, even though they are nested under higher-up `h1`s and `h2`s. For instance, https://doc.rust-lang.org/1.55.0/std/io/trait.Read.html has an `<h1>Trait std::io::Read</h1>` and an `<h1>Examples</h1>`. The latter should be `<h2>`. Similarly, there's an `<h4>fn read...</h4>`, with an `<h1>Errors</h1>` nested under it. The latter should be `<h5>`.

Per https://developer.mozilla.org/en-US/docs/Web/HTML/Element/Heading_Elements#usage_notes:

> Avoid skipping heading levels: always start from <h1>, followed by <h2> and so on.
> Use only one `<h1>` per page or view. It should concisely describe the overall purpose of the content.

Fortunately, we already control the `<h{level}>` tag that's emitted based on Markdown. The code is here: https://github.com/rust-lang/rust/blob/2b6ed3b675475abc01ce7e68bb75b457f0c85684/src/librustdoc/html/markdown.rs#L535-L540. What we additionally need:

 - Add a way to keep track of what heading level we're generating Markdown into: are we nested under an `<h1>` or under an `<h4>`?
 - Adjust the `{level}` interpolation to add the number of levels appropriate. This should be capped at 6, since h6 is the highest defined in HTML.
 - Adjust CSS styles appropriately.

yaymukund:
Closes #89309 

This fixes the `<h#>` header tags throughout the docs to reflect a semantic hierarchy.

- I ran a script to manually check that we don't have any files with multiple `<h1>` tags.
- Also checked that we never incorrectly nest e.g. a `<h2>` under an `<h3>`.
- I also spot-checked a bunch of pages (`trait.Read`, `enum.Ordering`, `primitive.isize`, `trait.Iterator`).

### DIFF
--- a/compiler/rustc_error_codes/src/error_codes/E0001.md
+++ b/compiler/rustc_error_codes/src/error_codes/E0001.md
@@ -1,4 +1,4 @@
-### Note: this error code is no longer emitted by the compiler.
+#### Note: this error code is no longer emitted by the compiler.
 
 This error suggests that the expression arm corresponding to the noted pattern
 will never be reached as for all possible values of the expression being
--- a/compiler/rustc_error_codes/src/error_codes/E0001.md
+++ b/compiler/rustc_error_codes/src/error_codes/E0001.md
@@ -1,4 +1,4 @@
-#### Note: this error code is no longer emitted by the compiler.
+### Note: this error code is no longer emitted by the compiler.
 
 This error suggests that the expression arm corresponding to the noted pattern
 will never be reached as for all possible values of the expression being
--- a/src/librustdoc/externalfiles.rs
+++ b/src/librustdoc/externalfiles.rs
@@ -39,14 +39,32 @@ impl ExternalHtml {
         let bc = format!(
             "{}{}",
             bc,
-            Markdown(&m_bc, &[], id_map, codes, edition, playground, 0).into_string()
+            Markdown {
+                content: &m_bc,
+                links: &[],
+                ids: id_map,
+                error_codes: codes,
+                edition,
+                playground,
+                heading_level: 0
+            }
+            .into_string()
         );
         let ac = load_external_files(after_content, diag)?;
         let m_ac = load_external_files(md_after_content, diag)?;
         let ac = format!(
             "{}{}",
             ac,
-            Markdown(&m_ac, &[], id_map, codes, edition, playground, 0).into_string()
+            Markdown {
+                content: &m_ac,
+                links: &[],
+                ids: id_map,
+                error_codes: codes,
+                edition,
+                playground,
+                heading_level: 0
+            }
+            .into_string()
         );
         Some(ExternalHtml { in_header: ih, before_content: bc, after_content: ac })
     }
--- a/src/librustdoc/externalfiles.rs
+++ b/src/librustdoc/externalfiles.rs
@@ -1,4 +1,4 @@
-use crate::html::markdown::{ErrorCodes, IdMap, Markdown, Playground};
+use crate::html::markdown::{ErrorCodes, HeadingOffset, IdMap, Markdown, Playground};
 use crate::rustc_span::edition::Edition;
 use std::fs;
 use std::path::Path;
@@ -46,7 +46,7 @@ impl ExternalHtml {
                 error_codes: codes,
                 edition,
                 playground,
-                heading_level: 1
+                heading_offset: HeadingOffset::H2,
             }
             .into_string()
         );
@@ -62,7 +62,7 @@ impl ExternalHtml {
                 error_codes: codes,
                 edition,
                 playground,
-                heading_level: 1
+                heading_offset: HeadingOffset::H2,
             }
             .into_string()
         );
--- a/src/librustdoc/externalfiles.rs
+++ b/src/librustdoc/externalfiles.rs
@@ -46,7 +46,7 @@ impl ExternalHtml {
                 error_codes: codes,
                 edition,
                 playground,
-                heading_level: 0
+                heading_level: 1
             }
             .into_string()
         );
@@ -62,7 +62,7 @@ impl ExternalHtml {
                 error_codes: codes,
                 edition,
                 playground,
-                heading_level: 0
+                heading_level: 1
             }
             .into_string()
         );
--- a/src/librustdoc/externalfiles.rs
+++ b/src/librustdoc/externalfiles.rs
@@ -39,14 +39,14 @@ impl ExternalHtml {
         let bc = format!(
             "{}{}",
             bc,
-            Markdown(&m_bc, &[], id_map, codes, edition, playground).into_string()
+            Markdown(&m_bc, &[], id_map, codes, edition, playground, 0).into_string()
         );
         let ac = load_external_files(after_content, diag)?;
         let m_ac = load_external_files(md_after_content, diag)?;
         let ac = format!(
             "{}{}",
             ac,
-            Markdown(&m_ac, &[], id_map, codes, edition, playground).into_string()
+            Markdown(&m_ac, &[], id_map, codes, edition, playground, 0).into_string()
         );
         Some(ExternalHtml { in_header: ih, before_content: bc, after_content: ac })
     }
--- a/src/librustdoc/html/markdown.rs
+++ b/src/librustdoc/html/markdown.rs
@@ -19,7 +19,7 @@
 //!     error_codes: ErrorCodes::Yes,
 //!     edition: Edition::Edition2015,
 //!     playground: &None,
-//!     heading_level: 0
+//!     heading_level: 1
 //! };
 //! let html = md.into_string();
 //! // ... something using html
@@ -544,7 +544,7 @@ impl<'a, 'b, 'ids, I: Iterator<Item = SpannedEvent<'a>>> Iterator
                 self.buf.push_front((Event::Html(format!("{} ", sec).into()), 0..0));
             }
 
-            let level = std::cmp::min(level + self.level + 1, MAX_HEADER_LEVEL);
+            let level = std::cmp::min(level + self.level, MAX_HEADER_LEVEL);
             self.buf.push_back((Event::Html(format!("</a></h{}>", level).into()), 0..0));
 
             let start_tags = format!(
--- a/src/librustdoc/html/markdown.rs
+++ b/src/librustdoc/html/markdown.rs
@@ -12,7 +12,15 @@
 //!
 //! let s = "My *markdown* _text_";
 //! let mut id_map = IdMap::new();
-//! let md = Markdown(s, &[], &mut id_map, ErrorCodes::Yes, Edition::Edition2015, &None, 0);
+//! let md = Markdown {
+//!     content: s,
+//!     links: &[],
+//!     ids: &mut id_map,
+//!     error_codes: ErrorCodes::Yes,
+//!     edition: Edition::Edition2015,
+//!     playground: &None,
+//!     heading_level: 0
+//! };
 //! let html = md.into_string();
 //! // ... something using html
 //! ```
@@ -69,19 +77,21 @@ pub(crate) fn summary_opts() -> Options {
 
 /// When `to_string` is called, this struct will emit the HTML corresponding to
 /// the rendered version of the contained markdown string.
-pub struct Markdown<'a>(
-    pub &'a str,
+pub struct Markdown<'a> {
+    pub content: &'a str,
     /// A list of link replacements.
-    pub &'a [RenderedLink],
+    pub links: &'a [RenderedLink],
     /// The current list of used header IDs.
-    pub &'a mut IdMap,
+    pub ids: &'a mut IdMap,
     /// Whether to allow the use of explicit error codes in doctest lang strings.
-    pub ErrorCodes,
+    pub error_codes: ErrorCodes,
     /// Default edition to use when parsing doctests (to add a `fn main`).
-    pub Edition,
-    pub &'a Option<Playground>,
-    pub u32,
-);
+    pub edition: Edition,
+    pub playground: &'a Option<Playground>,
+    /// Offset at which we render headings.
+    /// E.g. if `heading_level: 1`, then `# something` renders an `<h2>` instead of `<h1>`
+    pub heading_level: u32,
+}
 /// A tuple struct like `Markdown` that renders the markdown with a table of contents.
 crate struct MarkdownWithToc<'a>(
     crate &'a str,
@@ -1010,7 +1020,15 @@ impl LangString {
 
 impl Markdown<'_> {
     pub fn into_string(self) -> String {
-        let Markdown(md, links, mut ids, codes, edition, playground, level) = self;
+        let Markdown {
+            content: md,
+            links,
+            mut ids,
+            error_codes: codes,
+            edition,
+            playground,
+            heading_level,
+        } = self;
 
         // This is actually common enough to special-case
         if md.is_empty() {
@@ -1031,7 +1049,7 @@ impl Markdown<'_> {
 
         let mut s = String::with_capacity(md.len() * 3 / 2);
 
-        let p = HeadingLinks::new(p, None, &mut ids, level);
+        let p = HeadingLinks::new(p, None, &mut ids, heading_level);
         let p = Footnotes::new(p);
         let p = LinkReplacer::new(p.map(|(ev, _)| ev), links);
         let p = TableWrapper::new(p);
--- a/src/librustdoc/html/markdown.rs
+++ b/src/librustdoc/html/markdown.rs
@@ -8,7 +8,7 @@
 //! extern crate rustc_span;
 //!
 //! use rustc_span::edition::Edition;
-//! use rustdoc::html::markdown::{IdMap, Markdown, ErrorCodes};
+//! use rustdoc::html::markdown::{HeadingOffset, IdMap, Markdown, ErrorCodes};
 //!
 //! let s = "My *markdown* _text_";
 //! let mut id_map = IdMap::new();
@@ -19,7 +19,7 @@
 //!     error_codes: ErrorCodes::Yes,
 //!     edition: Edition::Edition2015,
 //!     playground: &None,
-//!     heading_level: 1
+//!     heading_offset: HeadingOffset::H2,
 //! };
 //! let html = md.into_string();
 //! // ... something using html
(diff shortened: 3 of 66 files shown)


GuillaumeGomez:
I added an invalid file: `compiler/rustc_error_codes/10`. ;)

> Also checked that we never incorrectly nest e.g. a <h2> under an <h3>.

Not sure what you meant by that.

So globally, if I understand correctly, you want to prevent `<h1>` to be added inside doc comments (which is what the issue is explaining as well). Can you add a picture where you show a before/after of a std page with headings please? Please also add a page where all possible headings are display (from `<h2>` to `<h6>`).

yaymukund:
Thank you for the review! 

> I added an invalid file: `compiler/rustc_error_codes/10`. ;)

Thanks, I've removed it.

> Not sure what you meant by that.

I wrote [a python script](https://gist.github.com/yaymukund/4613ecf344f2c710509767a7022517e8) to check that there is never `<h3> ... <h2> ... </h2> ... </h3>`, which would be invalid html. 

> So globally, if I understand correctly, you want to prevent `<h1>` to be added inside doc comments (which is what the issue is explaining as well).

Yes. The hierarchy of `<h1>`, `<h2>`, `<h3>` should also make sense. This is described in the original issue (see the 'fn read' example).

> Can you add a picture where you show a before/after of a std page with headings please? Please also add a page where all possible headings are display (from `<h2>` to `<h6>`).

Yes. Please be aware these screenshots are fairly large (~1.3Mb) so I've linked to them rather than display them inline. I opened them in two tabs, scrolled to the same location, and switched between tabs.

| Page  | `master` | `docblock-headings` |
| ------------- | ------------- | ------- |
| `MaybeUninit` | [Link](https://i.imgur.com/jEzLemO.png) | [Link](https://i.imgur.com/YH78QV0.png)|
| `OpenOptions` | [Link](https://i.imgur.com/mHY8vY9.png) | [Link](https://i.imgur.com/rK9JfZm.png) |
| `String`  | [Link](https://i.imgur.com/doRt1CD.png)| [Link](https://i.imgur.com/yHeMIBW.png)|

GuillaumeGomez:
> I wrote a python script to check that there is never `<h3>` ... `<h2>` ... `</h2>` ... `</h3>`, which would be invalid html.

That shouldn't be needed, we already have checks for that (you can run them with `x.py test html-checker --stage 1`).

> Yes. Please be aware these screenshots are fairly large (~1.3Mb) so I've linked to them rather than display them inline. I opened them in two tabs, scrolled to the same location, and switched between tabs and they look identical aside from a slight (2px?) shift in the search bar spacing.

Nice, thanks! Btw, you forgot to crop the right side on your last link. ;)

But if I understand correctly, the display shouldn't be changed at all, only the heading level. If so, well done!

src/librustdoc/html/markdown.rs
```diff
@@ -78,6 +80,7 @@ pub struct Markdown<'a>(
     /// Default edition to use when parsing doctests (to add a `fn main`).
     pub Edition,
     pub &'a Option<Playground>,
+    pub u32,
```
GuillaumeGomez, line 83: Please add a doc comment explaining what this field is. Out of context, it's impossible to know what it is without going through the code.

src/librustdoc/html/markdown.rs
```diff
@@ -489,11 +492,12 @@ struct HeadingLinks<'a, 'b, 'ids, I> {
     toc: Option<&'b mut TocBuilder>,
     buf: VecDeque<SpannedEvent<'a>>,
     id_map: &'ids mut IdMap,
+    level: u32,
```
GuillaumeGomez, line 495:
```suggestion
        heading_level: u32,
```

src/librustdoc/html/markdown.rs
```diff
@@ -489,11 +492,12 @@ struct HeadingLinks<'a, 'b, 'ids, I> {
...
 }
 
 impl<'a, 'b, 'ids, I> HeadingLinks<'a, 'b, 'ids, I> {
-    fn new(iter: I, toc: Option<&'b mut TocBuilder>, ids: &'ids mut IdMap) -> Self {
-        HeadingLinks { inner: iter, toc, buf: VecDeque::new(), id_map: ids }
+    fn new(iter: I, toc: Option<&'b mut TocBuilder>, ids: &'ids mut IdMap, level: u32) -> Self {
```
GuillaumeGomez, line 499:
```suggestion
    fn new(iter: I, toc: Option<&'b mut TocBuilder>, ids: &'ids mut IdMap, heading_level: u32) -> Self {
```

src/librustdoc/html/markdown.rs
notriddle, line 83: This is getting silly. A struct with seven member variables should just be a regular struct, with named fields, not a tuple.

src/librustdoc/html/markdown.rs
```diff
@@ -1005,7 +1010,7 @@ impl LangString {
 
 impl Markdown<'_> {
     pub fn into_string(self) -> String {
-        let Markdown(md, links, mut ids, codes, edition, playground) = self;
+        let Markdown(md, links, mut ids, codes, edition, playground, level) = self;
```
GuillaumeGomez, line 1013:
```suggestion
        let Markdown(md, links, mut ids, codes, edition, playground, heading_level) = self;
```

src/librustdoc/html/markdown.rs
```diff
@@ -1026,7 +1031,7 @@ impl Markdown<'_> {
 
         let mut s = String::with_capacity(md.len() * 3 / 2);
 
-        let p = HeadingLinks::new(p, None, &mut ids);
+        let p = HeadingLinks::new(p, None, &mut ids, level);
```
GuillaumeGomez, line 1034: Same.

src/librustdoc/html/render/mod.rs
```diff
@@ -471,19 +471,35 @@ fn settings(root_path: &str, suffix: &str, themes: &[StylePath]) -> Result<Strin
...
 fn document(w: &mut Buffer, cx: &Context<'_>, item: &clean::Item, parent: Option<&clean::Item>) {
+    document_at_level(w, cx, item, parent, 0)
+}
+
+fn document_at_level(
+    w: &mut Buffer,
+    cx: &Context<'_>,
+    item: &clean::Item,
+    parent: Option<&clean::Item>,
+    level: u32,
```
GuillaumeGomez, line 482:
```suggestion
    heading_level: u32,
```

src/librustdoc/html/render/mod.rs
```diff
@@ -471,19 +471,35 @@ fn settings(root_path: &str, suffix: &str, themes: &[StylePath]) -> Result<Strin
...
 }
 
 /// Render md_text as markdown.
-fn render_markdown(w: &mut Buffer, cx: &Context<'_>, md_text: &str, links: Vec<RenderedLink>) {
+fn render_markdown(
+    w: &mut Buffer,
+    cx: &Context<'_>,
+    md_text: &str,
+    links: Vec<RenderedLink>,
+    level: u32,
```
GuillaumeGomez, line 501:
```suggestion
    heading_level: u32,
```

src/librustdoc/html/render/mod.rs
```diff
@@ -531,15 +548,21 @@ fn document_short(
     }
 }
 
-fn document_full_collapsible(w: &mut Buffer, item: &clean::Item, cx: &Context<'_>) {
-    document_full_inner(w, item, cx, true);
+fn document_full_collapsible(w: &mut Buffer, item: &clean::Item, cx: &Context<'_>, level: u32) {
```
GuillaumeGomez, line 551:
```suggestion
fn document_full_collapsible(w: &mut Buffer, item: &clean::Item, cx: &Context<'_>, heading_level: u32) {
```

src/librustdoc/html/render/mod.rs
```diff
@@ -531,15 +548,21 @@ fn document_short(
...
+    document_full_inner(w, item, cx, true, level);
 }
 
-fn document_full(w: &mut Buffer, item: &clean::Item, cx: &Context<'_>) {
-    document_full_inner(w, item, cx, false);
+fn document_full(w: &mut Buffer, item: &clean::Item, cx: &Context<'_>, level: u32) {
```
GuillaumeGomez, line 555:
```suggestion
fn document_full(w: &mut Buffer, item: &clean::Item, cx: &Context<'_>, heading_level: u32) {
```

src/librustdoc/html/render/mod.rs
```diff
@@ -531,15 +548,21 @@ fn document_short(
...
+    document_full_inner(w, item, cx, false, level);
 }
 
-fn document_full_inner(w: &mut Buffer, item: &clean::Item, cx: &Context<'_>, is_collapsible: bool) {
+fn document_full_inner(
+    w: &mut Buffer,
+    item: &clean::Item,
+    cx: &Context<'_>,
+    is_collapsible: bool,
+    level: u32,
```
GuillaumeGomez, line 564:
```suggestion
    heading_level: u32,
```

src/librustdoc/html/render/mod.rs
```diff
@@ -1339,7 +1362,7 @@ fn render_impl(
                 } else {
                     document_item_info(&mut info_buffer, cx, item, Some(parent));
                     if rendering_params.show_def_docs {
-                        document_full(&mut doc_buffer, item, cx);
+                        document_full(&mut doc_buffer, item, cx, 3);
```
GuillaumeGomez, line 1365: That is a change compared to before. Please add a rustdoc test for it.

src/librustdoc/html/render/print_item.rs
```diff
@@ -626,7 +626,7 @@ fn item_trait(w: &mut Buffer, cx: &Context<'_>, it: &clean::Item, t: &clean::Tra
         let item_type = m.type_();
         let id = cx.derive_id(format!("{}.{}", item_type, name));
         let mut content = Buffer::empty_from(w);
-        document(&mut content, cx, m, Some(t));
+        document_at_level(&mut content, cx, m, Some(t), 3);
```
GuillaumeGomez, line 629: Same, this is a change compared to before, please add a rustdoc test for it.

src/librustdoc/html/markdown.rs
```diff
@@ -530,6 +534,7 @@ impl<'a, 'b, 'ids, I: Iterator<Item = SpannedEvent<'a>>> Iterator
                 self.buf.push_front((Event::Html(format!("{} ", sec).into()), 0..0));
             }
 
+            let level = std::cmp::min(level + self.level + 1, MAX_HEADER_LEVEL);
```
notriddle, line 537: When `level = 0`, you get `# My Header` :arrow_right: `<h2>My Header</h2>`. This seems weird to me.

I would have expected that `HeadingLinks { level: 0 }` would mean that headings aren't getting modified. Instead, you've made it so that there is no way to tell HeadingLinks to not bump the levels. It's also silently changing how the standalone markdown files (stuff like `/src/doc/index.md`) are generated, and it's why the error index got changed.

notriddle (changes requested):

src/librustdoc/html/markdown.rs
GuillaumeGomez: Completely forgot about the standalone markdown files, this is an excellent point!

src/librustdoc/html/markdown.rs
GuillaumeGomez: Also, `u32` could be replaced with `u8` since it never goes above 6 in principle... (just nitpicking)

src/librustdoc/html/markdown.rs
GuillaumeGomez: > When `level = 0`, you get `# My Header` :arrow_right: `<h2>My Header</h2>`. This seems weird to me.

I wasn't feeling great about it either but since I'm not the only thinking this, we should definitely make the heading level match the level we use.

yaymukund: Just to be clear - you want doc comments in the error index to behave differently (`# this is an h1`) from doc comments in std docs (`# this is an h2`)?

notriddle: I'm not sure why it makes sense to talk about them "behaving the same" as doc comments. They're totally different.

For one thing, the biggest reason why this probably has to be fixed by changing rustdoc's behavior is that there's no way for the end-user to know what the right heading level is. The same exact doc comment might be rendered at multiple levels in multiple pages, like [slice::first on the slice page](https://doc.rust-lang.org/stable/std/primitive.slice.html#method.first) (which is a fourth-level header, so the items in the doc comment should be h5), and [slice::first on Vec](https://doc.rust-lang.org/stable/std/vec/struct.Vec.html#method.first) (which is an h3, so its doc comment headers should start at h4).

For the error index, you're probably in the right here. The error index pages each have h1's at the beginning, and when they're viewed as individual documents, this works (they're `.md` files, so you expect them to be viewed using things like GitHub's markdown preview pane, or inside an IDE). But when they're combined to form the full error index, the resulting page has multiple h1's.

**But for standalone markdown files, this seems like a breaking change made with insufficient motivation.** Those aren't being combined to form huge mega-documents, and [it's a stable feature](https://doc.rust-lang.org/stable/rustdoc/what-is-rustdoc.html#using-standalone-markdown-files), not just some internal thing used by the compiler, so changing its behavior should only be done with very compelling reasons.

notriddle: The short version: yes to changing doc comments, yes to changing the error index, no to changing standalone markdown files.

yaymukund: Great - this is exactly the context I needed, thanks for explaining. I agree with you. I'll update the code + maybe add a test for this standalone markdown feature if I can figure out how to do that.

yaymukund:
I've done the following:

- Refactor `level: u32` to `heading_offset: HeadingOffset`. I can revert this if you dislike, but I felt it was worth the slight additional verbosity to distinguish the *offset* from the level itself.
- Refactor `Markdown(...)` to `Markdown { ... }`
- Bugfix: fix headings in docs for trait implementations.
- No more default `heading_offset`. We explicitly specify `HeadingOffset::H2` (docblocks), `H5` (deeply nested portions of docblocks), or `H1` (rendering raw markdown).
- Add tests for the three `<h#>` changes.

compiler/rustc_error_codes/src/error_codes/E0001.md
```diff
@@ -1,4 +1,4 @@
-#### Note: this error code is no longer emitted by the compiler.
+### Note: this error code is no longer emitted by the compiler.
```
GuillaumeGomez, line 1: Are those changes still needed? They are markdown files, so you changed the titles from `h4` to `h3`.

src/librustdoc/html/static/css/themes/dark.css
```diff
@@ -93,7 +93,7 @@ pre, .rustdoc.source .example-wrap {
 	background-color: #0a042f !important;
 }
 
-.docblock h1, .docblock h2, .docblock h3, .docblock h4, .docblock h5 {
+.docblock h2, .docblock h3, .docblock h4, .docblock h5, .docblock h6 {
```
GuillaumeGomez, line 96: Same.

src/librustdoc/html/static/css/themes/light.css
```diff
@@ -93,7 +93,7 @@ pre, .rustdoc.source .example-wrap {
 	background-color: #f6fdb0 !important;
 }
 
-.docblock h1, .docblock h2, .docblock h3, .docblock h4, .docblock h5 {
+.docblock h2, .docblock h3, .docblock h4, .docblock h5, .docblock h6 {
```
GuillaumeGomez, line 96: Same.

GuillaumeGomez:
> Refactor `level: u32` to `heading_offset: HeadingOffset`. I can revert this if you dislike, but I felt it was worth the slight additional verbosity to distinguish the offset from the level itself.

Please keep it, it's a very good idea!

With this, it's mostly nits remaining. Nice work, thanks!

compiler/rustc_error_codes/src/error_codes/E0001.md
yaymukund: This is the 2nd point in @notriddle 's summary: "yes to changing the error index".

However, after thinking about it, I don't think it's needed. We control all the error markdown files, so we can just render them at the top level and ensure there are no `# ` in them. I've gone ahead and reverted this change.

yaymukund:
For some reason, GitHub won't let me 'request a re-review' from notriddle, but all outstanding comments have been addressed. Thanks.

GuillaumeGomez (approved):

src/librustdoc/externalfiles.rs
```diff
@@ -39,14 +39,32 @@ impl ExternalHtml {
...
             bc,
-            Markdown(&m_bc, &[], id_map, codes, edition, playground).into_string()
+            Markdown {
+                content: &m_bc,
+                links: &[],
+                ids: id_map,
+                error_codes: codes,
+                edition,
+                playground,
+                heading_offset: HeadingOffset::H2,
```
notriddle, line 49: I must have done a bad job of explaining myself, because this is exactly backwards from what I wanted.

I was hoping for H1 on external files, and H2 on the error index, because the error index is formed by combining multiple files together, while external files are formed from one single document.

GuillaumeGomez: Which reminds me that no test with only markdown and no rust code was added (completely forgot, thanks @notriddle !).

notriddle: That makes sense to me.

notriddle (approved):

PR #89507, opened 2021-10-04
Add `#[repr(i8)]` to `Ordering`

lopopolo:
Followup to #89491 to allow `Ordering` to auto-derive `AsRepr` once the proposal to add `AsRepr` (#81642) lands.

cc @joshtriplett

### DIFF
--- a/library/core/src/cmp.rs
+++ b/library/core/src/cmp.rs
@@ -323,6 +323,7 @@ pub struct AssertParamIsEq<T: Eq + ?Sized> {
 /// ```
 #[derive(Clone, Copy, PartialEq, Debug, Hash)]
 #[stable(feature = "rust1", since = "1.0.0")]
+#[repr(i8)]
 pub enum Ordering {
     /// An ordering where a compared value is less than another.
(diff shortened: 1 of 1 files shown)


joshtriplett:
This seems reasonable to me, and I think we should make this part of the ABI of Ordering.

@rfcbot merge

joshtriplett:
Note that I think this change makes sense no matter what happens with #81642; I don't think this needs to wait on #81642.

joshtriplett:
Snippet for release notes:

- The `Ordering` enum is now guaranteed to use an `i8` representation, with the values `-1`, `0`, and `1`. This makes it easier to implement other `cmp`-style APIs.

JanBeh:
> The `Ordering` enum is now guaranteed to use an `i8` representation, with the values `-1`, `0`, and `1`. This makes it easier to implement other `cmp`-style APIs.

@joshtriplett Maybe that guarantee should explicitly be added to the documentation of `std::cmp::Ordering` as well? See also [this discussion on URLO](https://users.rust-lang.org/t/std-ordering-as-integer/73450?u=jbe).

PR #89508, opened 2021-10-04
Stabilize `const_panic`

jhpratt:
Closes #51999

FCP completed in #89006

@rustbot label +A-const-eval +A-const-fn +T-lang

cc @oli-obk for review (not `r?`'ing as not on lang team)

PR #89509, opened 2021-10-04
Stabilize `unreachable_unchecked` as `const fn`

### LINKED ISSUE #53188: Tracking Issue for const `unreachable_unchecked`
Currently one cannot call `unreachable_unchecked` in constant functions, so one cannot add hints of the form:

```rust
const fn foo(x: i32) -> i32 {
    if x < 0 { unreachable_unchecked() } else { x * 2 } 
}
```

These types of hints can heavily influence code generation, which can substantially change the performance of `const fn` when they are invoked at run-time. 

It would be a shame to have to write and call `foo_compile_time` and `foo_run_time` because the code generated for `foo_compile_time` is sub-par.

This issue has been assigned to @canova via [this comment](https://github.com/rust-lang/rust/issues/53188#issuecomment-660317181).

jhpratt:
Closes #53188

This PR stabilizes `core::hint::unreachable_unchecked` as `const fn`. MIRI is able to detect when this method is called. Stabilization was delayed until `const_panic` was stabilized so as to avoid users calling this method in its place (thus resulting in runtime UB). With #89508, that is no longer an issue.

@rustbot label +A-const-eval +A-const-fn +T-lang +S-blocked

(not sure why it's T-lang, but that's what the tracking issue is)

### DIFF
--- a/library/core/src/hint.rs
+++ b/library/core/src/hint.rs
@@ -44,7 +44,7 @@ use crate::intrinsics;
 /// ```
 #[inline]
 #[stable(feature = "unreachable", since = "1.27.0")]
-#[rustc_const_unstable(feature = "const_unreachable_unchecked", issue = "53188")]
+#[rustc_const_stable(feature = "const_unreachable_unchecked", since = "1.57.0")]
 pub const unsafe fn unreachable_unchecked() -> ! {
     // SAFETY: the safety contract for `intrinsics::unreachable` must
     // be upheld by the caller.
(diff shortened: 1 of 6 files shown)


joshtriplett:
@rfcbot merge

nikomatsakis:
I'd like to see @RalfJung or @oli-obk weigh in, or perhaps a brief stabilization report. I'm not really particularly familiar with this intrinsic, I have to admit.

RalfJung:
Yeah, this intrinsic is basically the simplest kind of UB we have, so with const_panic stable I see no reason not to stabilize this as well.

nikomatsakis:
@rfcbot fcp reviewed :)

PR #89513, opened 2021-10-04
BTree: refactor clone, mostly for readability

ssomers:
Simplifies the `clone` member, in particular its `clone_subtree` function, at the expense of having more code elsewhere, which might be reusable but wasn't reused up to now. The new destructor exploits the fact that the underlying navigation code spots the end of the tree itself, if we destroy the tree in a single direction. In theory, this performs whimsically better than before because it no longer keeps track of the number of elements already cloned, just in case the clone needs to be rolled back after a panic.

### DIFF
--- a/library/alloc/src/collections/btree/bare_tree.rs
+++ b/library/alloc/src/collections/btree/bare_tree.rs
@@ -1,0 +1,54 @@
+use super::node::{marker, NodeRef};
+
+/// Simple tree around a raw `NodeRef`, that provides a destructor.
+/// Unlike smarter sister `BTreeMap`, it does not keep track of element count,
+/// does not allow replacing the root, therefore can be statically tyoed by the
+/// node type of the root, and cannot represent a tree without root node
+/// (its `Option` exists only to disable the `drop` method when needed).
+pub struct BareTree<K, V, RootType>(Option<NodeRef<marker::Owned, K, V, RootType>>);
+
+impl<K, V, RootType> Drop for BareTree<K, V, RootType> {
+    fn drop(&mut self) {
+        if let Some(root) = self.0.take() {
+            let mut cur_edge = root.forget_type().into_dying().first_leaf_edge();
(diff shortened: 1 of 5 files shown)


joshtriplett:
r? rust-lang/libs

Mark-Simulacrum (commented):
Hm. I am not convinced that this is an improvement -- it looks like we're creating a new abstraction but only using it inside the clone code. Is there an expectation that future work will be able to reuse that abstraction? The PR description seems to imply maybe, but I'm not sure I'm reading it right.

I'm worried that the abstraction doesn't seem like it's really abstracting any significant code (rather just providing one-line wrappers), so I'm not sure it pulls its weight yet at least. Maybe you can talk a little bit about the intent/future plans?

library/alloc/src/collections/btree/bare_tree.rs
```diff
@@ -0,0 +1,54 @@
...
+
+impl<K, V, RootType> Drop for BareTree<K, V, RootType> {
+    fn drop(&mut self) {
+        if let Some(root) = self.0.take() {
+            let mut cur_edge = root.forget_type().into_dying().first_leaf_edge();
+            while let Some((next_edge, kv)) = unsafe { cur_edge.deallocating_next() } {
+                unsafe { kv.drop_key_val() };
+                cur_edge = next_edge;
+            }
+        }
```
Mark-Simulacrum, line 18: It looks like this doesn't replicate the continue-dropping-if-we-see-a-panic code, presumably because the currently only user only drops BareTrees if we're already panicking. But that seems like a footgun should we start using it somewhere that *does* need to keep dropping.

ssomers: > presumably because

Actually because I didn't realize it belongs there.

ssomers:
> Is there an expectation that future work will be able to reuse that abstraction?

Not any work that I have lying around or plan to do.

Issue #89517, opened 2021-10-04
Tracking Issue for XXX


a1phyr:
Feature gate: `#![feature(unix_file_vectored_at)]`

This is a tracking issue for positioned vectored I/O on Unix platforms.

### Public API

```rust
trait std::os::unix::fs::FileExt {
    // Existing methods: `read_at`, `read_exact_at`, `write_at`, `write_all_at`

    fn read_vectored_at(&mut self, bufs: &mut [io::IoSliceMut<'_>], offset: u64) -> io::Result<usize>;
    fn write_vectored_at(&mut self, bufs: &[io::IoSlice<'_>], offset: u64) -> io::Result<usize>;
}
```

### Steps / History

- [ ] Implementation
- [ ] Final comment period (FCP)
- [ ] Stabilization PR

PR #89518, opened 2021-10-04
Add vectored positioned I/O on Unix

a1phyr:
Add methods for vectored I/O with an offset on `File` for `unix` under `#![feature(unix_file_vectored_at)]`.

The new methods are wrappers around `preadv` and `pwritev`.

Tracking issue: #89517

### DIFF
--- a/library/std/src/sys/unix/fd.rs
+++ b/library/std/src/sys/unix/fd.rs
@@ -152,9 +152,7 @@ impl FileDesc {
         target_os = "freebsd",
         target_os = "fuchsia",
         target_os = "illumos",
-        target_os = "ios",
         target_os = "linux",
-        target_os = "macos",
         target_os = "netbsd",
     ))]
     pub fn read_vectored_at(&self, bufs: &mut [IoSliceMut<'_>], offset: u64) -> io::Result<usize> {
@@ -170,6 +168,7 @@ impl FileDesc {
     }
 
     #[cfg(not(any(
+        target_os = "android",
         target_os = "emscripten",
         target_os = "freebsd",
         target_os = "fuchsia",
@@ -188,7 +187,7 @@ impl FileDesc {
     //
     // On 32-bit targets, we don't want to deal with weird ABI issues around
     // passing 64-bits parameters to syscalls, so we fallback to the default
-    // implementation.
+    // implementation if `preadv` is not available.
     #[cfg(all(target_os = "android", target_pointer_width = "64"))]
     pub fn read_vectored_at(&self, bufs: &mut [IoSliceMut<'_>], offset: u64) -> io::Result<usize> {
         super::weak::syscall! {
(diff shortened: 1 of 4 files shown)


a1phyr:
Well, I copied `read_vectored` for this.

Anyway, `readv` takes a `const struct iovec*`, so this is the good type. However, according to Rust rules we cannot modify something behind a shared reference, so I think we should get a `*mut` pointer and cast it to `*const`.

a1phyr:
Rebased

Dylan-DPC:
@joshtriplett this is ready for review

workingjubilee:
@a1phyr Sorry for this slipping through the cracks! If you can fix the latest merge conflicts I can get this approved.

r? @workingjubilee

a1phyr:
Rebased

workingjubilee:
Aa, eeto... bleh!

matthiaskrgr:
@bors r- rollup=iffy
failed in https://github.com/rust-lang/rust/pull/108559
`CANNOT LINK EXECUTABLE: could not load library "libstd-ef8e480f4327ce79.so" needed by "/data/local/tmp/work/test0/a"; caused by cannot locate symbol "preadv" referenced by "libstd-ef8e480f4327ce79.so"...`

https://github.com/rust-lang-ci/rust/actions/runs/4292167740/jobs/7478270814#step:26:16943

workingjubilee:
@a1phyr Do you have any insight into the linkage issues?

a1phyr:
Wild guess: `preadv` was added to bionic in 2015 (cf https://android.googlesource.com/platform/bionic/+/6f4594d). The tests may run with on older version.

To fix this, we could either disable `preadv` on android or use weak linkage.

workingjubilee:
Our Android minimum is increasing to Android KitKat. That's lower than Marshmallow, the Oct 2015 release. I need more coffee before I remember the details of how weak symbol linkage works, but what we do here has to work on KitKat and Lollipop, so go with whichever feels right there.

library/std/src/sys/unix/fd.rs
```diff
@@ -197,6 +281,90 @@ impl FileDesc {
...
+                cmp::min(bufs.len(), max_iov()) as libc::c_int,
+                offset as _,
+            )
+        })?;
+        Ok(ret as usize)
+    }
+
+    #[cfg(all(target_os = "android", target_pointer_width = "32"))]
+    pub fn write_vectored_at(&self, bufs: &[IoSlice<'_>], offset: u64) -> io::Result<usize> {
+        super::weak::weak!(fn pwritev64(libc::c_int, *const libc::iovec, libc::c_int, off64_t) -> isize);
```
workingjubilee, line 350: I, too, feel super weak when thinking about these sorts of things.

workingjubilee (changes requested):
Bad news, I looked it up and it seems this was only introduced relatively late in the game by macOS (and iOS). It seems even https://github.com/rust-lang/rust/pull/104385 won't raise the minimum supported versions high enough to guarantee pwritev on those platforms.

workingjubilee (approved):
Huzzah, let's see if this goes or if it tells us to do more.

@bors r+

workingjubilee:
Not again?
@bors r+

rylev:
Small regression in Diesel which has started to show some bimodality. I think it's safe to call this triaged.

PR #89521, opened 2021-10-04
Perf experiment for "Get piece unchecked in write"

camsteffen:
Measuring the true cost of #83302 by reverting it and #89139 

r? @ghost


camsteffen:
@rust-timer queue

PR #89486, opened 2021-10-03
Update Let's Encrypt ROOT CA certificate in dist-(i686|x86_64)-linux docker images

hkratz:
The DST Root CA X3 used by Let's Encrypt has expired ([Let's Encrypt announcement](https://letsencrypt.org/docs/dst-root-ca-x3-expiration-september-2021/)). This patch installs the new root certificate (ISRG Root X1) and disables the old one. Disabling the old one is necessary because otherwise curl still fails to download from servers with Let's Encrypt certs even though they are cross-signed.

Fixes #89484.

PR #89285, opened 2021-09-26
Don't normalize opaque types with escaping late-bound regions

jackh726:
Fixes #88862

Turns out, this has some really bad perf implications in large types (issue #88862). While we technically can handle them fine, it doesn't change test output either way. For now, revert with an added benchmark. Future attempts to change this back will have to consider perf.

Needs a perf run once https://github.com/rust-lang/rustc-perf/pull/1033 is merged

r? @nikomatsakis

PR #89527, opened 2021-10-04
[beta] Beta rollup

ehuss:
* Fix WinUWP std compilation errors due to I/O safety #88587 
* Disable RemoveZsts in generators to avoid query cycles #88979 
* Disable the evaluation cache when in intercrate mode #88994 
* Fix linting when trailing macro expands to a trailing semi #88996 
* Don't use projection cache or candidate cache in intercrate mode #89125 
* 2229: Mark insignificant dtor in stdlib #89144 
* Temporarily rename int_roundings functions to avoid conflicts #89184 
* [rfc 2229] Drop fully captured upvars in the same order as the regular drop code #89208 
* Use the correct edition for syntax highlighting doctests #89277 
* Don't normalize opaque types with escaping late-bound regions #89285 
* Update Let's Encrypt ROOT CA certificate in dist-(i686|x86_64)-linux docker images #89486 
* Fix Windows LLVM issue. #89380 

Cargo update:
* - [beta] 1.56 backports (rust-lang/cargo#9958)
* - [beta] Revert "When a dependency does not have a version, git or path… (rust-lang/cargo#9912)
* - [beta] Fix rustc --profile=dev unstable check. (rust-lang/cargo#9901)


ehuss:
Oh, apologies!  I didn't realize that was part of the process.  I'll send an update to the forge in a bit to clarify that.

PR #89531, opened 2021-10-04
library std, libc dependency update

devnexen:
to solve #87528 build.


joshtriplett:
You need to update the lock file and commit the resulting update.

dtolnay (approved):

PR #89532, opened 2021-10-04
Document behavior of  `MaybeLiveLocals` regarding enums and field-senstivity

ecstatic-morse:
This arose from a [discussion on Zulip](https://rust-lang.zulipchat.com/#narrow/stream/189540-t-compiler.2Fwg-mir-opt/topic/MaybeLiveLocals.20and.20Discriminants) where a new contributor attempted to implement a dead-store elimination pass using this analysis. They ran into a nasty hack around `SetDiscriminant`, which lets us handle assignments of literals to enum-typed locals (e.g. `x = Some(4)`) correctly. This took me a while to figure out.

Document this oddity, so the next person will have an easier time, and add a test to enshrine the current behavior.

r? @tmiasko


tmiasko (approved):
Thanks for documenting this!

PR #89534, opened 2021-10-04
Introduce `tcx.get_diagnostic_name`

camsteffen:
Introduces a "reverse lookup" for diagnostic items. This is mainly intended for @rust-lang/clippy which often does a long series of `is_diagnostic_item` calls for the same `DefId`. 

r? @oli-obk

### DIFF
--- a/compiler/rustc_hir/src/diagnostic_items.rs
+++ b/compiler/rustc_hir/src/diagnostic_items.rs
@@ -1,0 +1,17 @@
+use crate::def_id::DefId;
+use rustc_data_structures::fx::FxHashMap;
+use rustc_data_structures::stable_hasher::{HashStable, StableHasher};
+use rustc_span::Symbol;
+
+#[derive(Debug, Default)]
+pub struct DiagnosticItems {
+    pub id_to_name: FxHashMap<DefId, Symbol>,
+    pub name_to_id: FxHashMap<Symbol, DefId>,
+}
+
+impl<CTX: crate::HashStableContext> HashStable<CTX> for DiagnosticItems {
+    #[inline]
+    fn hash_stable(&self, ctx: &mut CTX, hasher: &mut StableHasher) {
+        self.name_to_id.hash_stable(ctx, hasher);
+    }
+}
--- a/compiler/rustc_middle/src/ty/context.rs
+++ b/compiler/rustc_middle/src/ty/context.rs
@@ -1232,12 +1232,17 @@ impl<'tcx> TyCtxt<'tcx> {
     /// Obtain the given diagnostic item's `DefId`. Use `is_diagnostic_item` if you just want to
     /// compare against another `DefId`, since `is_diagnostic_item` is cheaper.
     pub fn get_diagnostic_item(self, name: Symbol) -> Option<DefId> {
-        self.all_diagnostic_items(()).get(&name).copied()
+        self.all_diagnostic_items(()).name_to_id.get(&name).copied()
+    }
+
+    /// Obtain the diagnostic item's name
+    pub fn get_diagnostic_name(self, id: DefId) -> Option<Symbol> {
+        self.diagnostic_items(id.krate).id_to_name.get(&id).copied()
     }
 
     /// Check whether the diagnostic item with the given `name` has the given `DefId`.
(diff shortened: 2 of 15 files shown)


oli-obk:
Nice!

Two (not really important things):

* do we want to get rid of is_diagnostic_item, to nudge ppl to using one of the other two functions?
* it may be good from an API perspective to make the fields private and add a setter that inserts into both at the same time.

But these don't need to happen, let's get this rolling first and update clippy and stuff.

@bors r+

compiler/rustc_hir/src/diagnostic_items.rs
```diff
@@ -0,0 +1,9 @@
+use crate::def_id::DefId;
+use rustc_data_structures::fx::FxHashMap;
+use rustc_span::Symbol;
+
+#[derive(Debug, Default, HashStable_Generic)]
```
camsteffen, line 5: @oli-obk Without much understanding of queries, it seems like something is suboptimal if this struct is `HashStable_Generic`? Because only one of the fields should be hashed.

oli-obk: hmm... good catch. it could be a perf issue, we can perf it, and if it is, implement HashStable manually?

camsteffen: Another thing I considered is to break it into two queries. But I don't know if you can define a query that runs once and is only cached in memory?

oli-obk: idk, my initial reaction is that another query is overkill for this. I think I would rather use a manual HashStable impl

camsteffen:
> do we want to get rid of is_diagnostic_item, to nudge ppl to using one of the other two functions?

I think `is_diagnostic_item` will still be useful for a lot of cases.

> it may be good from an API perspective to make the fields private and add a setter that inserts into both at the same time

I didn't do this because we only observe the `insert` return value for one of the maps, so putting an API layer there feels a little weird. And we only insert in one place.

camsteffen:
@rust-timer queue

camsteffen:
I guess this is okay?

compiler/rustc_middle/src/ty/context.rs
```diff
@@ -1232,12 +1232,17 @@ impl<'tcx> TyCtxt<'tcx> {
     /// Obtain the given diagnostic item's `DefId`. Use `is_diagnostic_item` if you just want to
     /// compare against another `DefId`, since `is_diagnostic_item` is cheaper.
     pub fn get_diagnostic_item(self, name: Symbol) -> Option<DefId> {
-        self.all_diagnostic_items(()).get(&name).copied()
+        self.all_diagnostic_items(()).name_to_id.get(&name).copied()
+    }
+
+    /// Obtain the diagnostic item's name
+    pub fn get_diagnostic_name(self, id: DefId) -> Option<Symbol> {
+        self.all_diagnostic_items(()).id_to_name.get(&id).copied()
```
oli-obk, line 1240: I believe this is the source of the perf changes. The following should be equivalent I think

```suggestion
        self.diagnostic_items(id.krate).id_to_name.get(&id).copied()
```

camsteffen:
Had to do a little more refactoring for a recent change in `compiler/rustc_typeck/src/check/method/suggest.rs`. I added `#[inline]` to the hash impl.

@bors try @rust-timer queue

compiler/rustc_typeck/src/check/method/suggest.rs
```diff
@@ -1047,51 +1047,41 @@ impl<'a, 'tcx> FnCtxt<'a, 'tcx> {
...
+            let trait_pred = match pred.kind().skip_binder() {
+                ty::PredicateKind::Trait(trait_pred) => trait_pred,
+                _ => continue,
+            };
+            let adt = match trait_pred.self_ty().ty_adt_def() {
+                Some(adt) if adt.did.is_local() => adt,
+                _ => continue,
+            };
+            let can_derive = match self.tcx.get_diagnostic_name(trait_pred.def_id()) {
+                Some(sym::Default) => !adt.is_enum(),
```
camsteffen, line 1062: CC @estebank since you just worked on this. I think I am fixing a missing suggestion here for `Default` on enums.

PR #89535, opened 2021-10-04
fix busted JavaScript in error index generator

notriddle:
The old JavaScript didn't work. It filled the browser console with "e.previousElementSibling not defined" errors, because it didn't account for the example-wrap div that a newer version of rustdoc added.

Additionally, it had copied versions of utility functions that had been optimized in rustdoc main.js. This version updates those.


Mark-Simulacrum:
Seems okay. 

@bors r+ rollup

PR #89538, opened 2021-10-05
Make rustdoc not highlight `->` and `=>` as operators

notriddle:
It was marking them up as `<span class="op">=</span><span class="op">&gt;</span>`,
which is bloaty and wrong (at least, I think `<=` and `=>` should probably be different colors, since they're so different and yet made from the same symbols).

Before:

image

After:

image


GuillaumeGomez (approved):

GuillaumeGomez:
Thanks!

@bors: r+

PR #89541, opened 2021-10-05
Cleanup src/test/ui/{simd,simd-intrinsic}

workingjubilee:
Initial motivation was to simplify a huge macro expansion using a tuple, since we can just use an array in `#[repr(simd)]` now for the same result. But also, several tests were going unnoticed during development of SIMD intrinsics because people kept looking in the wrong directory, and many are basically run-pass vs. build-fail versions of the same tests, so let's keep them close together and simplify their names, so they're easier to sift through.


workingjubilee:
hmmm, sorry to bother, I think I should just tag...
r? @calebzulawski

workingjubilee:
I guess that didn't work, eh!

Mark-Simulacrum:
> I guess that didn't work, eh!

Is this referring to something in particular? Should I hold off on approving?

Otherwise, r=me

PR #89546, opened 2021-10-05
Make an initial guess for metadata size to reduce buffer resizes

joshtriplett:
When reading metadata, the compiler starts with a `Vec::new()`, which will need to grow repeatedly as the metadata gets decompressed into it. Reduce the number of resizes by starting out at the size of the compressed data.


petrochenkov:
@bors r+ rollup
A trivial micro-optimization that shouldn't hurt, even if the benchmarks don't show noticeable improvements.

PR #89551, opened 2021-10-05
Stabilize `const_raw_ptr_deref` for `*const T`

### LINKED ISSUE #51911: [tracking issue] dereferencing raw pointers inside constants (const_raw_ptr_deref)
The `const_raw_ptr_deref` feature allows dereferencing raw pointers inside constants.

This can already be emulated via union transmuting raw pointers to references and referencing the references. This feature gate helps us experiment with the semantics of actually dereferencing normally.

Blocked on https://github.com/rust-lang/const-eval/issues/14.

jhpratt:
This stabilizes dereferencing immutable raw pointers in const contexts.
It does not stabilize `*mut T` dereferencing. This is behind the
same feature gate as mutable references.

closes https://github.com/rust-lang/rust/issues/51911


jhpratt:
@rustbot label +A-const-eval +A-const-fn +T-lang

r? @joshtriplett 

cc @oli-obk @RalfJung

compiler/rustc_const_eval/src/transform/check_consts/check.rs
```diff
@@ -737,7 +742,7 @@ impl Visitor<'tcx> for Checker<'mir, 'tcx> {
                             return;
                         }
                     }
-                    self.check_op(ops::RawPtrDeref);
+                    self.check_op(ops::RawMutPtrDeref);
```
oli-obk, line 745: I think we can safely bundle this with MutDeref. I see no reason to stabilize raw mut ptr deref separately from reference mut deref

oli-obk (commented):
The test diffs lgtm

I'm only unsure about the new feature gate

oli-obk:
cc @rust-lang/wg-const-eval

compiler/rustc_const_eval/src/transform/check_consts/check.rs
jhpratt: The reason I split it is because mutable references are unstable. If mutable pointer dereferencing is stabilized, doesn't that effectively open a back door?

oli-obk: I meant to only use a single feature gate for mutable references and raw mutable pointers. This way, we will stabilize mutable references together with mutable pointers. What I meant to say is that I believe that once we stabilize mutable references, there is no reason not to also stabilize raw mutable pointers.

so basically use https://github.com/rust-lang/rust/blob/735c1339bd2a5e16dc0fde04f24930eda1bd049b/compiler/rustc_const_eval/src/transform/check_consts/ops.rs#L350-L369 here and adjust its message to also talk about mutable raw pointers.

nbdd0121:
I think `*mut T` should either stay in the current feature gate, or be merged to `const_mut_refs` feature gate like @oli-obk suggested.

compiler/rustc_const_eval/src/transform/check_consts/check.rs
jhpratt: Ah, gotcha. I mentioned that in the tracking issue but edited it out afterwards. Should the feature gate's name stay the same or something else bikesheddable?

jhpratt:
Just pushed up changes that move `*mut T` to the existing `const_mut_refs` feature.

src/test/ui/consts/const-eval/assign-to-static-within-other-static-2.rs
```diff
@@ -1,7 +1,7 @@
 // New test for #53818: modifying static memory at compile-time is not allowed.
 // The test should never compile successfully
 
-#![feature(const_raw_ptr_deref, const_mut_refs)]
+#![feature(const_mut_refs)]
```
oli-obk, line 4: this test is now equal to src/test/ui/consts/const-eval/mod-static-with-const-fn.rs, so you can remove one of them

oli-obk (approved):

oli-obk:
At present, we do not allow mutable references at all in const fn. If we allowed mutable raw pointers, it would mean that you could (unsafely) write the same logic as you could with mutable references. That would invite people to write unsafe code that should not be unsafe. And it would open up all issues with mutable references that we haven't solved yet (it's out of my cache, I need to look into mutable references again, but it just seems bad to have mutable pointers before mutable references)

jhpratt:
Merge conflict is just a clash of accepted language features. I'll address once FCP has finished to avoid doing it multiple times.

jhpratt:
Rebased.

jhpratt:
Pretty sure I can't do this, but let's give it a shot.

@bors retry

ehuss:
Does the [const eval](https://doc.rust-lang.org/nightly/reference/const_eval.html#constant-expressions) section need to be updated for this change?

jhpratt:
Not seeing anything notable in the perf run.

PR #89553, opened 2021-10-05
Add trait object safety information on trait documentation page

### LINKED ISSUE #85138: Show whether a trait is object-safe on its document page
So that it is easy for user to recognize object-safe traits.

GuillaumeGomez:
Fixes #85138.

It looks like this:

Screenshot from 2021-10-05 12-42-18

So now, what remains to be done is debating over the appearance of the information.

r? @jsha

### DIFF
--- a/src/librustdoc/html/render/print_item.rs
+++ b/src/librustdoc/html/render/print_item.rs
@@ -497,6 +497,13 @@ fn item_trait(w: &mut Buffer, cx: &Context<'_>, it: &clean::Item, t: &clean::Tra
 
     // Output the trait definition
     wrap_into_docblock(w, |w| {
+        if let Some(def_id) = it.def_id.as_def_id() {
+            write!(
+                w,
+                "<div class=\"obj-info\">This trait is <b>{}object safe</b>.</div>",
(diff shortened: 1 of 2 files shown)


jsha:
Thanks for working on this! Two issues:

 - This puts the "this trait is object safe" text in a very high-priority place, right at the very top of each trait page. But it's relatively low-priority information: most uses of traits are not using trait objects. Also, putting it at the top of so many trait pages will make beginners feel like they need to learn about trait objects and object safety right away, even though they are an intermediate-skill part of the language.
 
 - We put English text in the UI, which I think is fine for now - i18n would be a big project. But there are relatively few places we inject English text into the documentation itself. The main place I can think of is [`#[non_exhaustive]`](https://github.com/rust-lang/rust/issues/84309#issuecomment-822659742), and I've seen some interest in changing that to just be a link to the doc page that explains the non_exhaustive attribute. An issue with injecting English text in the documentation is that it may be jarring when the rest of the documentation is non-English.

Let's brainstorm some more UI over in the issue thread.

JohnCSimon:
@GuillaumeGomez 
Thanks for your contribution.

@rustbot label: +S-inactive

PR #89558, opened 2021-10-05
Add rustc lint, warning when iterating over hashmaps

lcnr:
r? rust-lang/wg-incr-comp

### DIFF
--- a/src/test/ui-fulldeps/internal-lints/query_stability.rs
+++ b/src/test/ui-fulldeps/internal-lints/query_stability.rs
@@ -1,0 +1,24 @@
+// compile-flags: -Z unstable-options
+
+#![feature(rustc_private)]
+#![deny(rustc::potential_query_instability)]
+
+extern crate rustc_data_structures;
+
+use rustc_data_structures::fx::{FxHashMap, FxHashSet};
+
+fn main() {
+    let mut x = FxHashMap::<u32, i32>::default();
+
+    for _ in x.drain() {}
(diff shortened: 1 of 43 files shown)


src/test/ui-fulldeps/internal-lints/query_stability.rs
```diff
@@ -0,0 +1,24 @@
...
+
+    for _ in x.iter() {}
+    //~^ ERROR using `iter`
+
+    for _ in Some(&mut x).unwrap().iter_mut() {}
+    //~^ ERROR using `iter_mut`
+
+    for _ in x {}
+    //~^ ERROR using `into_iter`
+}
```
estebank, line 24: `fmt` is gonna complain about the missing newline

estebank:
@bors r+

We don't need to mark the `Fx` versions?

lcnr:
@bors r- 

somehow CI doesn't run at all for this PR and this lint causes a lot of warnings which is why i've opened https://github.com/rust-lang/compiler-team/issues/465 for this.

> We don't need to mark the `Fx versions?

The `Fx` versions are only type aliases, which are pretty much ignored by rustc (as you probably also painfully aware)

petrochenkov:
@lcnr 
We have some lint naming conventions - https://github.com/rust-lang/rfcs/blob/master/text/0344-conventions-galore.md#lints.
As you can see `allow(query_stability)` doesn't make much sense as a sentence.

estebank (approved):

PR #89568, opened 2021-10-05
Profile CGUs

### LINKED ISSUE #89308: Uniquely identify CGU's and accurately track their size and merges
## Overview

This issue is in relation to [optimizing codegen scheduling](https://github.com/rust-lang/rust/issues/89281). To better profile codegen and see where/what we need to optimize, we need to get metrics for individual codegen units. This requires a mechanism for uniquely identifying them; I'm guessing this mechanism will be similar to how query invocations are uniquely identified. E.g.

```rust
pub struct QueryInvocationId(pub u32);

pub struct CodegenUnitId(pub u32);
```

Using this mapping, we should be able to track the various metrics we need.

## CGU Size

Tracking CGU size by itself should not be too hard; at some arbitrary point in time we determine the CGU's size and map this quantity to a CGU ID for later lookup. @Mark-Simulacrum mentioned that we should have support for knowing if a CGU is big just because, or because it absorbed other CGU's; and that this is something that should be displayed on perf.r-l.o. my guess is we can do this using a similar structure to what's already in place in `SelfProfiler`:

```rust
pub struct SelfProfiler {
    profiler: Profiler,
    event_filter_mask: EventFilter,

    string_cache: RwLock<FxHashMap<String, StringId>>,

    query_event_kind: StringId,
    generic_activity_event_kind: StringId,
    incremental_load_result_event_kind: StringId,
    incremental_result_hashing_event_kind: StringId,
    query_blocked_event_kind: StringId,
    query_cache_hit_event_kind: StringId,
    /*** NEW ***/
    codegen_unit_merge_event_kind: StringId,
}
```

However, as I understand it, this approach ignores tracking _which_ CGU's have been merged together and just informs
us _when_ and _where_ a merge has occurred; shouldn't we also know _what_ got merged? Or does using a `CodegenUnitId` solve this?

Lastly, it has been mentioned that the size metric is not currently accurate, and that the quantity of LLVM instructions
might not provide enough insight. Until I know more, I plan on using CGU instruction count as the size metric.

## To Do

- [ ] Add support to uniquely identify individual CGU's
- [ ] Change size metric to reflect inst counts of CGU's
- [ ] Add support for tracking which CGU's have been merged; and where and when it happened.

cc @Mark-Simulacrum

Sl1mb0:
Will eventually resolve #89308.

r? @Mark-Simulacrum

### DIFF
--- a/compiler/rustc_data_structures/src/profiling.rs
+++ b/compiler/rustc_data_structures/src/profiling.rs
@@ -111,6 +111,8 @@ bitflags::bitflags! {
         const LLVM                = 1 << 7;
         const INCR_RESULT_HASHING = 1 << 8;
 
+        const CGU_NAMES           = 1 << 9;
+
         const DEFAULT = Self::GENERIC_ACTIVITIES.bits |
                         Self::QUERY_PROVIDERS.bits |
                         Self::QUERY_BLOCKED.bits |
@@ -136,6 +138,7 @@ const EVENT_FILTERS_BY_NAME: &[(&str, EventFilter)] = &[
     ("args", EventFilter::ARGS),
     ("llvm", EventFilter::LLVM),
     ("incr-result-hashing", EventFilter::INCR_RESULT_HASHING),
+    ("cgu-names", EventFilter::CGU_NAMES),
 ];
 
 /// Something that uniquely identifies a query invocation.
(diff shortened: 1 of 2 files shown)


Dylan-DPC:
Thanks for contributing. I'm closing this since it's inactive and seems you might not have the time to get back to it. Feel free to reopen the PR when you have the time in the future

Issue #75397, opened 2020-08-11
Nonsensical error in rustc output.


zakarumych:
Given a code with few issues I got completely unrelated error bound to line with one of those issues.

I tried this code:

```rust
#![recursion_limit="8"] // default value makes compilation longer.

extern crate num_traits; // 0.2.12

struct Foo(u8);

#[repr(transparent)]
struct Bar(Foo);

impl Bar {
    fn cast(foo: &mut Foo) -> &mut Self {
        unsafe { &*(foo as *mut _ as *mut _) } // It should be error about coercing `&_` to `&mut _`
    }
}

impl Baz {} // If you delete this, or replace `Baz` with existing type - weired error vanishes.


```

I expected to see this happen:
```
error[E0308]: mismatched types
  --> src/lib.rs:12:18
   |
12 |         unsafe { &*(foo as *mut _ as *mut _) } // It should be error about coercing `&_` to `&mut _`
   |                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^ types differ in mutability
   |
   = note: expected mutable reference `&mut Bar`
                      found reference `&_`
```

Instead, this happened:

```
error[E0275]: overflow evaluating the requirement `&num_rational::Ratio<_>: num_traits::Pow<u128>`
  --> src/lib.rs:12:18
   |
12 |         unsafe { &*(foo as *mut _ as *mut _) } // It should be error about coercing `&_` to `&mut _`
   |                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^
   |
   = help: consider adding a `#![recursion_limit="16"]` attribute to your crate (`playground`)
   = note: required because of the requirements on the impl of `num_traits::Pow<u128>` for `&num_rational::Ratio<num_rational::Ratio<_>>`
   = note: required because of the requirements on the impl of `num_traits::Pow<u128>` for `&num_rational::Ratio<num_rational::Ratio<num_rational::Ratio<_>>>`
   = note: required because of the requirements on the impl of `num_traits::Pow<u128>` for `&num_rational::Ratio<num_rational::Ratio<num_rational::Ratio<num_rational::Ratio<_>>>>`
   = note: required because of the requirements on the impl of `num_traits::Pow<u128>` for `&num_rational::Ratio<num_rational::Ratio<num_rational::Ratio<num_rational::Ratio<num_rational::Ratio<_>>>>>`
[...]

PR #89569, opened 2021-10-05
Fix for Issue 89275

### LINKED ISSUE #89275: Overflow evaluating the requirement [stable/nightly] (buggy trait collisions?)
Given the following code snippet:

```rust
use num_rational as _;

struct Foo;

impl Foo {
    fn downcast<W>(&self) -> &W {
        todo!()
    }
}

struct Other;

fn main() {
    let other: &mut Other = Foo.downcast();
}
```

One would expect the code to error, explaining that the types differ in mutability (*we will come back to this in a moment*).

However, it instead leaves a long and confusing error:
```
error[E0275]: overflow evaluating the requirement `&Ratio<_>: num_traits::pow::Pow<u32>`
  --> src/main.rs:14:29
   |
14 |     let other: &mut Other = Foo.downcast();
   |                             ^^^^^^^^^^^^^^
   |
   = help: consider adding a `#![recursion_limit="256"]` attribute to your crate (`bug`)
   = note: required because of the requirements on the impl of `num_traits::pow::Pow<u32>` for `&Ratio<Ratio<_>>`
   = note: 127 redundant requirements hidden
   = note: required because of the requirements on the impl of `num_traits::pow::Pow<&i32>` for `&Ratio<Ratio<Ratio<Ratio<Ratio<Ratio<Ratio<Ratio<Ratio<Ratio<Ratio<Ratio<Ratio<Ratio<Ratio<Ratio<Ratio<Ratio<Ratio<Ratio<Ratio<Ratio<Ratio<Ratio<Ratio<Ratio<Ratio<Ratio<Ratio<Ratio<Ratio<Ratio<Ratio<Ratio<Ratio<Ratio<Ratio<Ratio<Ratio<Ratio<Ratio<Ratio<Ratio<Ratio<Ratio<Ratio<Ratio<Ratio<Ratio<Ratio<Ratio<Ratio<Ratio<Ratio<Ratio<Ratio<Ratio<Ratio<Ratio<Ratio<Ratio<Ratio<Ratio<Ratio<Ratio<Ratio<Ratio<Ratio<Ratio<Ratio<Ratio<Ratio<Ratio<Ratio<Ratio<Ratio<Ratio<Ratio<Ratio<Ratio<Ratio<Ratio<Ratio<Ratio<Ratio<Ratio<Ratio<Ratio<Ratio<Ratio<Ratio<Ratio<Ratio<Ratio<Ratio<Ratio<Ratio<Ratio<Ratio<Ratio<Ratio<Ratio<Ratio<Ratio<Ratio<Ratio<Ratio<Ratio<Ratio<Ratio<Ratio<Ratio<Ratio<Ratio<Ratio<Ratio<Ratio<Ratio<Ratio<Ratio<Ratio<Ratio<Ratio<Ratio<Ratio<Ratio<Ratio<Ratio<Ratio<_>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>`

For more information about this error, try `rustc --explain E0275`.
error: could not compile `bug`

```

If you follow the suggestion of increasing the recursion limit each time, it will eventually segfault once you have `#![recursion_limit="2048"]`
```
error: could not compile `bug`

Caused by:
  process didn't exit successfully: `rustc --crate-name bug --edition=2018 src/main.rs --error-format=json --json=diagnostic-rendered-ansi --crate-type bin --emit=dep-info,link -C embed-bitcode=no -C debuginfo=2 -C metadata=f6406766fb3e32c0 -C extra-filename=-f6406766fb3e32c0 --out-dir /home/zippy/bug/target/debug/deps -C incremental=/home/zippy/bug/target/debug/incremental -L dependency=/home/zippy/bug/target/debug/deps --extern num_rational=/home/zippy/bug/target/debug/deps/libnum_rational-13f69084195ac9b5.rlib` (signal: 11, SIGSEGV: invalid memory reference)
```

Furthermore, removing the import of `num-rational` leads to the program erroring *as expected*
```
  --> src/main.rs:12:29
   |
12 |     let other: &mut Other = Foo.downcast();
   |                ----------   ^^^^^^^^^^^^^^ types differ in mutability
   |                |
   |                expected due to this
   |
   = note: expected mutable reference `&mut Other`
                      found reference `&_`

For more information about this error, try `rustc --explain E0308`.
error: could not compile `bug` due to previous error
```

### Meta

`rustc --version --verbose`:
```
rustc 1.57.0-nightly (addb4da68 2021-09-25)
binary: rustc
commit-hash: addb4da686a97da46159f0123cb6cdc2ce3d7fdb
commit-date: 2021-09-25
release: 1.57.0-nightly
LLVM version: 13.0.0
```
This error exists on stable, too.

<details><summary>Backtrace</summary>
<p>

```
The code doesn't compile, and as such there's no backtrace
```

</p>
</details>

tom7980:
This should fix #89275 and potentially #77291 #75397

The problem of error reporting iterating over ambiguous traits and emitting a recursion limit error message on the initial mismatched type error. 

I updated the OverflowError type from a struct to an enum so I believe it is no longer a ZST - I don't think this should cause any issue as it should only be constructed if you actually cause a recursion error in type checking or in this case where there's an ambiguous trait in scope that it checks for a conversion method.


tom7980:
I need to sort out my commits first - seem to have ended up in an extremely complicated place

PR #89570, opened 2021-10-05
[Experiment] Split exhaustiveness logic into its own crate

Nadrieril:
`rust-analyzer` currently maintains a patched copy of the exhaustiveness algorithm for their own analyses. That's too bad, so I factored out everything rustc-specific from the relevant files. If perf isn't killed, I'll move it into its own crate.

Until the type library gets librarified, there'll still be some nontrivial code that rust-analyzer will have to duplicate. But the core functionality will be shared and we should be able to implement cool features for the rust-analyzer side.


compiler/rustc_mir_build/src/thir/pattern/check_match.rs
```diff
@@ -356,11 +361,11 @@ fn check_for_bindings_named_same_as_variants(
 }
 
 /// Checks for common cases of "catchall" patterns that may not be intended as such.
-fn pat_is_catchall(pat: &DeconstructedPat<'_, '_>) -> bool {
+fn pat_is_catchall(pat: &DeconstructedPat<'_, MatchCheckCtxt<'_>>) -> bool {
     use Constructor::*;
     match pat.ctor() {
         Wildcard => true,
-        Single => pat.iter_fields().all(|pat| pat_is_catchall(pat)),
+        Single | Tuple(_) | Ref(_) | BoxPat(_) => pat.iter_fields().all(|pat| pat_is_catchall(pat)),
```
joshtriplett, line 368: Is this a bugfix?

This seems like it might be a source of performance impact.

Nadrieril: No, this is because I split the `Single` case into more specific subcases so that I could abstract it all in a trait. I suspected this could be the source of perf degradation but my local measurements don't show much impact from this.
The problem of course is that my local measurements don't match what `rust-timer` reports here so you could be right. When I get back to this I'll measure intermediate commits with `rust-timer`.

Nadrieril:
(I still intend get this done. Will get back to it whenever my work deadlines are done and motivation strikes)

Dylan-DPC:
Closing this as it is inactive

PR #89580, opened 2021-10-05
Point at source of trait bound obligations in more places

estebank:
Be more thorough in using `ItemObligation` and `BindingObligation` when
evaluating obligations so that we can point at trait bounds that
introduced unfulfilled obligations. We no longer incorrectly point at
unrelated trait bounds (`substs-ppaux.verbose.stderr`).

In particular, we now point at trait bounds on method calls.

We no longer point at "obvious" obligation sources (we no longer have a
note pointing at `Trait` saying "required by a bound in `Trait`", like
in `associated-types-no-suitable-supertrait*`).

We no longer point at associated items (`ImplObligation`), as they didn't
add any user actionable information, they just added noise.

Address part of #89418.


compiler/rustc_infer/src/infer/error_reporting/mod.rs
```diff
@@ -2138,10 +2138,19 @@ impl<'a, 'tcx> InferCtxt<'a, 'tcx> {
...
                     },
                     self.tcx.generics_of(owner.to_def_id()),
+                    hir.span(hir_id),
                 )
             });
+
+        let span = match generics {
+            // This is to get around the trait identity obligation, that has a `DUMMY_SP` as signal
+            // for other diagnostics, so we need to recover it here.
+            Some((_, _, node)) if span.is_dummy() => node,
```
cjgillot, line 2148: Should this test be performed around the `hir.span(hir_id)` call?

cjgillot:
> @cjgillot, I see that you touched this file. Do you know if there's _any_ reliance on `predicates_of` changing when `trait` modifiers like `pub` and `unsafe` are changed? Currently they do because the change of the span, but with this PR's change, that's no longer true. Would it be reasonable to change the test to no longer look for that?

I don't know enough the trait solving algorithm to comment on the need for the value to change when pub and unsafe change. OTOH, I would consider that any incremental code that relies on span-related invalidation to work correctly would be a red flag. So IMHO, it's fine to update the test.

If you want confirmation, you'd need to ask to professional trait solvers like @jackh726.

jackh726:
> If you want confirmation, you'd need to ask to professional trait solvers like @jackh726.
😂❤you give me too much credit

> Do you know if there's _any_ reliance on `predicates_of` changing when `trait` modifiers like `pub` and `unsafe` are changed? 
I would be extremely surprised if this were the case. I imagine for everything but diagnostics, we throw away the span *anyways*.

nagisa (commented):
Looked through the test changes for now, there seem to be some unfortunate regressions at least from my perspective.

nagisa (approved):
This seems very nice. r=me with one remaining nit.

compiler/rustc_typeck/src/check/compare_method.rs
```diff
@@ -227,12 +223,22 @@ fn compare_predicate_entailment<'tcx>(
...
-        for predicate in impl_m_own_bounds.predicates {
+        for (predicate, span) in iter::zip(impl_m_own_bounds.predicates, impl_m_own_bounds.spans) {
+            let normalize_cause = traits::ObligationCause::misc(span, impl_m_hir_id);
             let traits::Normalized { value: predicate, obligations } =
-                traits::normalize(&mut selcx, param_env, normalize_cause.clone(), predicate);
+                traits::normalize(&mut selcx, param_env, normalize_cause, predicate);
 
             inh.register_predicates(obligations);
-            inh.register_predicate(traits::Obligation::new(cause.clone(), param_env, predicate));
+            let cause = ObligationCause::new(
```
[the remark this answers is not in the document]
estebank: Done.

compiler/rustc_errors/src/emitter.rs
```diff
@@ -1266,22 +1266,37 @@ impl EmitterWriter {
...
             if *level != Level::FailureNote {
                 buffer.append(0, level.to_str(), Style::Level(*level));
+                label_width += level.to_str().len();
             }
             // only render error codes, not lint codes
             if let Some(DiagnosticId::Error(ref code)) = *code {
                 buffer.append(0, "[", Style::Level(*level));
                 buffer.append(0, &code, Style::Level(*level));
                 buffer.append(0, "]", Style::Level(*level));
+                label_width += 2 + code.len();
```
Mark-Simulacrum, line 1280: Can we say something like `": ".len()` presuming I've guessed correctly as to the meaning of this 2? (Or just a comment on at least one of these).

estebank: This `2` belongs to the two `[` `]` surrounding the `code`. There's another `2` below for the `": "`. Kept these magic numbers very close to their causes to try and make them obvious, but I guess they aren't?

src/test/ui/associated-types/substs-ppaux.verbose.stderr
```diff
@@ -82,11 +82,6 @@ note: required because of the requirements on the impl of `Foo<'_#0r, '_#1r, u8>
    |
 LL | impl<'a,'b,T,S> Foo<'a, 'b, S> for T {}
    |                 ^^^^^^^^^^^^^^     ^
-note: required by a bound in `Foo::bar`
-  --> $DIR/substs-ppaux.rs:7:30
-   |
-LL |     fn bar<'a, T>() where T: 'a {}
-   |                              ^^ required by this bound in `Foo::bar`
```
estebank, line 89: This looks like a regression :(

pnkfelix:
Visiting for weekly performance triage.

Many other PR's over the past week have observed a similar delta to `incr-unchanged` builds of `inflate`, so I am dismissing *that* change as spurious. (The current hypothesis is that it is due to https://github.com/rust-lang/rustc-perf/issues/1105.)

Other than that, there were a broad set of small regressions. Putting aside the ones tagged with `?` ("noisy"), there are 19 benchmarks that regressed by 0.10% to 0.42%. This seems like an acceptable cost.

Issue #89516, opened 2021-10-04
Remove EOF "probe" read in Read::read_to_end


jkugelman:
In #89165 I updated `Read::read_to_end` to try to detect EOF when the input buffer is full by reading into a small "probe" buffer. If that read returns `Ok(0)` then it avoids unnecessarily doubling the capacity of the input buffer.

> I think there'd be some value in trying to eliminate even the "probe buffer" call, but in the meantime, this seems like an improvement.

_Originally posted by @joshtriplett in https://github.com/rust-lang/rust/issues/89165#issuecomment-933081857_

Josh, did you have something in mind?

I thought a way: use [`read_vectored`](https://doc.rust-lang.org/std/io/trait.Read.html#method.read_vectored). I could add the probe buffer to a vectorized `readv`, which would eliminate the extra `read` syscall.

1. Is that idea worth pursuing?

2. Is it okay to simply call `read_vectored`, or would it be better to check [`is_read_vectored`](https://doc.rust-lang.org/std/io/trait.Read.html#method.is_read_vectored) and have separate vectorized and non-vectorized code paths? I'm inclined to keep it simple and do the former.

joshtriplett:
My original thought was, if we have a file size and it doesn't look like a placeholder value, we should just read that many bytes and stop.

the8472:
You're probably aware of the edge-case around procfs hence "it doesn't look like a placeholder value", but that isn't the only one. For example FUSE filesystems can make up any metadata. So we can't treat the size as anything but a hint. Only a zero-length read is the true indicator of EOF.

joshtriplett:
I'm aware that filesystems can give incorrect information. However, there's a *lot* of code out there that just does stat/read or stat/mmap, all of which breaks with an incorrect size. I don't think it would be unreasonable to make the same assumption.

In any case, I think using readv seems like a good optimization. I do think you would need to make sure vectored reads are possible, since without that, it would just look like a partial read from the stub implementation rather than from the OS.

the8472:
> However, there's a lot of code out there that just does stat/read or stat/mmap, all of which breaks with an incorrect size.

At least in the mmap case they're already accepting that this won't work on unusual files. If you're writing a rust program that replaces a bash script that reads from strange sources then you're more likely to reach for standard methods.

> FUSE filesystems presenting incorrect metadata that doesn't look like a placeholder value seem like a pathological case.

In my case it was a bug caused by an incorrect size calculation that only affected the metadata but not the reads.

[...]

Issue #80218, opened 2020-12-20
Tracking Issue for `std::io::read_to_string`


camelid:
Feature gate: `#![feature(io_read_to_string)]`

This is a tracking issue for the `std::io::read_to_string` function.

The equivalent of `std::fs::read_to_string`, but generalized to all
`Read` impls.

As the documentation on `std::io::read_to_string` says, the advantage of
this function is that it means you don't have to create a variable first
and it provides more type safety since you can only get the buffer out
if there were no errors. If you use `Read::read_to_string`, you have to
remember to check whether the read succeeded because otherwise your
buffer will be empty.

It's friendlier to newcomers and better in most cases to use an explicit
return value instead of an out parameter.

### Public API

```rust
pub fn read_to_string<R: Read>(reader: R) -> Result<String>;
```

### Steps / History

- [x] Implementation: #80217
- [x] FCP
- [ ] Stabilization PR (#100337)

### Unresolved Questions

- [x] Function signature (#92863)

DesmondWillowbrook:
Is this stabilized? Implementation PR looks like it has been merged in 1.51, but I'm getting an error from `cargo check` that this hasn't been stabilized.

rustc version: `rustc 1.52.1 (9bc8c42bb 2021-05-09)`

DesmondWillowbrook:
What is the status on stabilizing this addition to the library?

fee1-dead:
@rust-lang/libs

Can we start an FCP?

m-ou-se:
Thinking a bit about the signature: Considering `Read` is already implemented for `&mut impl Read`, shouldn't this function just take an `R` instead of `&mut R`?

camelid:
> Thinking a bit about the signature: Considering `Read` is already implemented for `&mut impl Read`, shouldn't this function just take an `R` instead of `&mut R`?

I modeled it after `Read::read_to_string`, which takes an `&mut self`, but that has to use `&mut self` because it's defined as part of the trait. I don't know enough about the idiom for `std` here, so if that works it seems fine to me.

jkugelman:
I'm working on a related PR https://github.com/rust-lang/rust/pull/89582 to have `File::read_to_string` enlarge the string buffer before reading to accommodate the full file size. A nice side effect is that it will make `io::read_to_string` more performant when the reader is a file.

If it lands do you think there's anything in [this comment](https://doc.rust-lang.org/std/io/fn.read_to_string.html#performance) that should be updated?

> #### Performance
>
[...]

PR #89582, opened 2021-10-06
Optimize File::read_to_end and read_to_string

jkugelman:
Reading a file into an empty vector or string buffer can incur unnecessary `read` syscalls and memory re-allocations as the buffer "warms up" and grows to its final size. This is perhaps a necessary evil with generic readers, but files can be read in more intelligently by checking the file size and reserving that much capacity.

`std::fs::read` and `std::fs::read_to_string` already perform this optimization: they open the file, reads its metadata, and call `with_capacity` with the file size. This ensures that the buffer does not need to be resized and avoids an initial string of small `read` syscalls.

However, if a user opens the `File` themselves and calls `file.read_to_end` or `file.read_to_string` they do not get this optimization.

```rust
let mut buf = Vec::new();
file.read_to_end(&mut buf)?;
```

I searched through this project's codebase and even here there are a *lot* of examples of this. They're found all over in unit tests, which isn't a big deal, but there are also several real instances in the compiler and in Cargo. I've documented the ones I found in a comment here:

https://github.com/rust-lang/rust/issues/89516#issuecomment-934423999

Most telling, the documentation for both the `Read` trait and the `Read::read_to_end` method both show this exact pattern as examples of how to use readers. What this says to me is that this shouldn't be solved by simply fixing the instances of it in this codebase. If it's here it's certain to be prevalent in the wider Rust ecosystem.

To that end, this commit adds specializations of `read_to_end` and `read_to_string` directly on `File`. This way it's no longer a minor footgun to start with an empty buffer when reading a file in.

A nice side effect of this change is that code that accesses a `File` as `impl Read` or `dyn Read` will benefit. For example, this code from `compiler/rustc_serialize/src/json.rs`:

```rust
pub fn from_reader(rdr: &mut dyn Read) -> Result<Json, BuilderError> {
    let mut contents = Vec::new();
    match rdr.read_to_end(&mut contents) {
```

Related changes:

- I also added specializations to `BufReader` to delegate to `self.inner`'s methods. That way it can call `File`'s optimized  implementations if the inner reader is a file.

- The private `std::io::append_to_string` function is now marked `unsafe`.

- `File::read_to_string` being more efficient means that the performance note for `io::read_to_string` can be softened. I've added @camelid's suggested wording from https://github.com/rust-lang/rust/issues/80218#issuecomment-936806502.

r? @joshtriplett

### DIFF
--- a/library/std/src/io/buffered/bufreader.rs
+++ b/library/std/src/io/buffered/bufreader.rs
@@ -308,6 +308,51 @@ impl<R: Read> Read for BufReader<R> {
     unsafe fn initializer(&self) -> Initializer {
         self.inner.initializer()
     }
+
+    // The inner reader might have an optimized `read_to_end`. Drain our buffer and then
+    // delegate to the inner implementation.
+    fn read_to_end(&mut self, buf: &mut Vec<u8>) -> io::Result<usize> {
+        let nread = self.cap - self.pos;
+        buf.extend_from_slice(&self.buf[self.pos..self.cap]);
+        self.discard_buffer();
+        Ok(nread + self.inner.read_to_end(buf)?)
+    }
+
+    // The inner reader might have an optimized `read_to_end`. Drain our buffer and then
+    // delegate to the inner implementation.
+    fn read_to_string(&mut self, buf: &mut String) -> io::Result<usize> {
+        // In the general `else` case below we must read bytes into a side buffer, check
+        // that they are valid UTF-8, and then append them to `buf`. This requires a
+        // potentially large memcpy.
+        //
+        // If `buf` is empty--the most common case--we can leverage `append_to_string`
+        // to read directly into `buf`'s internal byte buffer, saving an allocation and
+        // a memcpy.
+        if buf.is_empty() {
+            // `append_to_string`'s safety relies on the buffer only being appended to since
+            // it only checks the UTF-8 validity of new data. If there were existing content in
+            // `buf` then an untrustworthy reader (i.e. `self.inner`) could not only append
+            // bytes but also modify existing bytes and render them invalid. On the other hand,
+            // if `buf` is empty then by definition any writes must be appends and
+            // `append_to_string` will validate all of the new bytes.
+            unsafe { crate::io::append_to_string(buf, |b| self.read_to_end(b)) }
+        } else {
+            // We cannot append our byte buffer directly onto the `buf` String as there could
+            // be an incomplete UTF-8 sequence that has only been partially read. We must read
+            // everything into a side buffer first and then call `from_utf8` on the complete
+            // buffer.
+            let mut bytes = Vec::new();
+            self.read_to_end(&mut bytes)?;
+            let string = crate::str::from_utf8(&bytes).map_err(|_| {
+                io::Error::new_const(
+                    io::ErrorKind::InvalidData,
+                    &"stream did not contain valid UTF-8",
+                )
+            })?;
+            *buf += string;
+            Ok(string.len())
+        }
+    }
 }
 
 #[stable(feature = "rust1", since = "1.0.0")]
--- a/library/std/src/fs.rs
+++ b/library/std/src/fs.rs
@@ -198,19 +198,10 @@ pub struct DirBuilder {
     recursive: bool,
 }
 
-/// Indicates how large a buffer to pre-allocate before reading the entire file.
-fn initial_buffer_size(file: &File) -> usize {
-    // Don't worry about `usize` overflow because reading will fail regardless
-    // in that case.
-    file.metadata().map(|m| m.len() as usize).unwrap_or(0)
-}
-
 /// Read the entire contents of a file into a bytes vector.
 ///
 /// This is a convenience function for using [`File::open`] and [`read_to_end`]
-/// with fewer imports and without an intermediate variable. It pre-allocates a
-/// buffer based on the file size when available, so it is generally faster than
-/// reading into a vector created with [`Vec::new()`].
+/// with fewer imports and without an intermediate variable.
 ///
 /// [`read_to_end`]: Read::read_to_end
 ///
@@ -237,7 +228,7 @@ fn initial_buffer_size(file: &File) -> usize {
 pub fn read<P: AsRef<Path>>(path: P) -> io::Result<Vec<u8>> {
     fn inner(path: &Path) -> io::Result<Vec<u8>> {
         let mut file = File::open(path)?;
-        let mut bytes = Vec::with_capacity(initial_buffer_size(&file));
+        let mut bytes = Vec::new();
         file.read_to_end(&mut bytes)?;
         Ok(bytes)
     }
@@ -247,9 +238,7 @@ pub fn read<P: AsRef<Path>>(path: P) -> io::Result<Vec<u8>> {
 /// Read the entire contents of a file into a string.
 ///
 /// This is a convenience function for using [`File::open`] and [`read_to_string`]
-/// with fewer imports and without an intermediate variable. It pre-allocates a
-/// buffer based on the file size when available, so it is generally faster than
-/// reading into a string created with [`String::new()`].
+/// with fewer imports and without an intermediate variable.
 ///
 /// [`read_to_string`]: Read::read_to_string
 ///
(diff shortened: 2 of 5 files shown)


jkugelman:
I tested both `read_to_end` and `read_to_string` on a 10MB input file. It took 21 `read` syscalls to read it before, and now it only takes 2.

<details>
<summary>Before and after straces of <code>read_to_end</code></summary>

```rust
use std::env;
use std::fs::File;
use std::io::Read;

pub fn main() {
    let mut file = File::open(env::args().nth(1).expect("no file name")).expect("open failed");
    let mut buf = Vec::new();
    file.read_to_end(&mut buf).expect("read_to_end failed");
}
```

Before:

```
openat(AT_FDCWD, "large.txt", O_RDONLY|O_CLOEXEC) = 3
read(3, ""..., 32)                      = 32
read(3, ""..., 32)                      = 32
read(3, ""..., 64)                      = 64
read(3, ""..., 128)                     = 128
read(3, ""..., 256)                     = 256
read(3, ""..., 512)                     = 512
read(3, ""..., 1024)                    = 1024
read(3, ""..., 2048)                    = 2048
read(3, ""..., 4096)                    = 4096
read(3, ""..., 8192)                    = 8192
read(3, ""..., 16384)                   = 16384
read(3, ""..., 32768)                   = 32768
read(3, ""..., 65536)                   = 65536
mmap(NULL, 266240, PROT_READ|PROT_WRITE, MAP_PRIVATE|MAP_ANONYMOUS, -1, 0) = 0x7fbd9ad0b000
read(3, ""..., 131072)                  = 131072
mremap(0x7fbd9ad0b000, 266240, 528384, MREMAP_MAYMOVE) = 0x7fbd9ac8a000
read(3, ""..., 262144)                  = 262144
mremap(0x7fbd9ac8a000, 528384, 1052672, MREMAP_MAYMOVE) = 0x7fbd9ab89000
read(3, ""..., 524288)                  = 524288
mremap(0x7fbd9ab89000, 1052672, 2101248, MREMAP_MAYMOVE) = 0x7fbd9a988000
read(3, ""..., 1048576)                 = 1048576
mremap(0x7fbd9a988000, 2101248, 4198400, MREMAP_MAYMOVE) = 0x7fbd9a587000
read(3, ""..., 2097152)                 = 2097152
mremap(0x7fbd9a587000, 4198400, 8392704, MREMAP_MAYMOVE) = 0x7fbd99d86000
read(3, ""..., 4194304)                 = 4194304
mremap(0x7fbd99d86000, 8392704, 16781312, MREMAP_MAYMOVE) = 0x7fbd98d85000
read(3, ""..., 8388608)                 = 1711392
read(3, "", 6677216)                    = 0
munmap(0x7fbd98d85000, 16781312)        = 0
close(3)                                = 0
```

After:

```
$ strace -s 0 ./read_to_end large.txt
openat(AT_FDCWD, "large.txt", O_RDONLY|O_CLOEXEC) = 3
futex(0x7fb36787f0c8, FUTEX_WAKE_PRIVATE, 2147483647) = 0
statx(0, NULL, AT_STATX_SYNC_AS_STAT, STATX_ALL, NULL) = -1 EFAULT (Bad address)
statx(3, "", AT_STATX_SYNC_AS_STAT|AT_EMPTY_PATH, STATX_ALL, {stx_mask=STATX_ALL|0x1000, stx_attributes=0, stx_mode=S_IFREG|0644, stx_size=10100000, ...}) = 0
mmap(NULL, 10100736, PROT_READ|PROT_WRITE, MAP_PRIVATE|MAP_ANONYMOUS, -1, 0) = 0x7fb366ce4000
read(3, ""..., 10100000)                = 10100000
read(3, "", 32)                         = 0
munmap(0x7fb366ce4000, 10100736)        = 0
close(3)                                = 0
```
</details>

<details>
<summary>Before and after straces of <code>read_to_string</code></summary>

```rust
use std::env;
use std::fs::File;
use std::io::Read;

pub fn main() {
    let mut file = File::open(env::args().nth(1).expect("no file name")).expect("open failed");
    let mut buf = String::new();
    file.read_to_string(&mut buf).expect("read_to_string failed");
}
```

Before:

```
openat(AT_FDCWD, "large.txt", O_RDONLY|O_CLOEXEC) = 3
read(3, ""..., 32)                      = 32
read(3, ""..., 32)                      = 32
read(3, ""..., 64)                      = 64
read(3, ""..., 128)                     = 128
read(3, ""..., 256)                     = 256
read(3, ""..., 512)                     = 512
read(3, ""..., 1024)                    = 1024
read(3, ""..., 2048)                    = 2048
read(3, ""..., 4096)                    = 4096
read(3, ""..., 8192)                    = 8192
read(3, ""..., 16384)                   = 16384
read(3, ""..., 32768)                   = 32768
read(3, ""..., 65536)                   = 65536
mmap(NULL, 266240, PROT_READ|PROT_WRITE, MAP_PRIVATE|MAP_ANONYMOUS, -1, 0) = 0x7f0d54602000
read(3, ""..., 131072)                  = 131072
mremap(0x7f0d54602000, 266240, 528384, MREMAP_MAYMOVE) = 0x7f0d54581000
read(3, ""..., 262144)                  = 262144
mremap(0x7f0d54581000, 528384, 1052672, MREMAP_MAYMOVE) = 0x7f0d54480000
read(3, ""..., 524288)                  = 524288
mremap(0x7f0d54480000, 1052672, 2101248, MREMAP_MAYMOVE) = 0x7f0d5427f000
read(3, ""..., 1048576)                 = 1048576
mremap(0x7f0d5427f000, 2101248, 4198400, MREMAP_MAYMOVE) = 0x7f0d53e7e000
read(3, ""..., 2097152)                 = 2097152
mremap(0x7f0d53e7e000, 4198400, 8392704, MREMAP_MAYMOVE) = 0x7f0d5367d000
read(3, ""..., 4194304)                 = 4194304
mremap(0x7f0d5367d000, 8392704, 16781312, MREMAP_MAYMOVE) = 0x7f0d5267c000
read(3, ""..., 8388608)                 = 1711392
read(3, "", 6677216)                    = 0
munmap(0x7f0d5267c000, 16781312)        = 0
close(3)                                = 0
```


After:


```
openat(AT_FDCWD, "large.txt", O_RDONLY|O_CLOEXEC) = 3
futex(0x7f65ecb160c8, FUTEX_WAKE_PRIVATE, 2147483647) = 0
statx(0, NULL, AT_STATX_SYNC_AS_STAT, STATX_ALL, NULL) = -1 EFAULT (Bad address)
statx(3, "", AT_STATX_SYNC_AS_STAT|AT_EMPTY_PATH, STATX_ALL, {stx_mask=STATX_ALL|0x1000, stx_attributes=0, stx_mode=S_IFREG|0644, stx_size=10100000, ...}) = 0
mmap(NULL, 10100736, PROT_READ|PROT_WRITE, MAP_PRIVATE|MAP_ANONYMOUS, -1, 0) = 0x7f65ebf7b000
read(3, ""..., 10100000)                = 10100000
read(3, "", 32)                         = 0
munmap(0x7f65ebf7b000, 10100736)        = 0
close(3)                                = 0
```
</details>

library/std/src/io/buffered/bufreader.rs
```diff
@@ -308,6 +308,14 @@ impl<R: Read> Read for BufReader<R> {
...
         self.inner.initializer()
     }
+
+    fn read_to_end(&mut self, buf: &mut Vec<u8>) -> io::Result<usize> {
+        self.inner.read_to_end(buf)
+    }
+
+    fn read_to_string(&mut self, buf: &mut String) -> io::Result<usize> {
+        self.inner.read_to_string(buf)
+    }
```
a1phyr, line 318: These functions should take the internal buffer into account and use its data before deferring to the inner reader.

jkugelman: Thanks, that's a good catch. I've fixed them and added corresponding unit tests to verify the fixes. Could you take a look at what I did? Getting these right was tricky. I'm wondering if I missed a simpler way.

jkugelman: I don't love how `BufReader::read_to_string` reads into a side byte buffer, checks if it's valid UTF-8, and then appends the bytes to the passed in `String` if the UTF-8 validation succeeds. I wish I could do what [`append_to_string`](https://doc.rust-lang.org/src/std/io/mod.rs.html#335) does and append bytes directly to the input `String`, but I don't think it's safe to do so. `append_to_string` uses unsafe code and is only safe because it calls a trusted version of `read_to_end` that it knows will only append data. I can't do the same thing in `BufReader` because I don't trust `self.inner.read_to_end`, so there's an extra allocation/memcpy.

a1phyr: I think you could `append_to_string` the internal buffer, where you know you'll only append to the string, and then call `self.inner.read_to_string` to read what is left of the reader.
In this case, you don't have to trust `self.inner.read_to_end`.

PR #89587, opened 2021-10-06
Include rmeta candidates in "multiple matching crates" error

camelid:
Only dylib and rlib candidates were included in the error. I think the
reason is that at the time this error was originally implemented, rmeta
crate sources were represented different from dylib and rlib sources.
I wrote up more detailed analysis in [this comment][1].

The new version of the code is also a bit easier to read and should be
more robust to future changes since it uses `CrateSources::paths()`.

I also changed the code to sort the candidates to make the output deterministic;
added full stderr tests for the error; and added a long error code explanation.

[1]: https://github.com/rust-lang/rust/pull/88675#issuecomment-935282436

cc @Mark-Simulacrum @jyn514

### DIFF
--- a/compiler/rustc_error_codes/src/error_codes/E0464.md
+++ b/compiler/rustc_error_codes/src/error_codes/E0464.md
@@ -1,0 +1,6 @@
+The compiler found multiple library files with the requested crate name.
+
+This error can occur in several different cases -- for example, when using
+`extern crate` or passing `--extern` options without crate paths. It can also be
+caused by caching issues with the build directory, in which case `cargo clean`
+may help.
--- a/compiler/rustc_metadata/src/locator.rs
+++ b/compiler/rustc_metadata/src/locator.rs
@@ -910,9 +910,15 @@ impl CrateError {
                     "multiple matching crates for `{}`",
                     crate_name
                 );
+                let mut libraries: Vec<_> = libraries.into_values().collect();
+                // Make ordering of candidates deterministic.
+                // This has to `clone()` to work around lifetime restrictions with `sort_by_key()`.
+                // `sort_by()` could be used instead, but this is in the error path,
+                // so the performance shouldn't matter.
+                libraries.sort_by_cached_key(|lib| lib.source.paths().next().unwrap().clone());
                 let candidates = libraries
                     .iter()
-                    .filter_map(|(_, lib)| {
+                    .filter_map(|lib| {
                         let crate_name = &lib.metadata.get_root().name().as_str();
                         match (&lib.source.dylib, &lib.source.rlib) {
                             (Some((pd, _)), Some((pr, _))) => Some(format!(
--- a/compiler/rustc_metadata/src/locator.rs
+++ b/compiler/rustc_metadata/src/locator.rs
@@ -232,6 +232,7 @@ use rustc_span::Span;
 use rustc_target::spec::{Target, TargetTriple};
 
 use snap::read::FrameDecoder;
+use std::fmt::Write as _;
 use std::io::{Read, Result as IoResult, Write};
 use std::path::{Path, PathBuf};
 use std::{cmp, fmt, fs};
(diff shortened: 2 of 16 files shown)


compiler/rustc_metadata/src/locator.rs
```diff
@@ -912,21 +912,26 @@ impl CrateError {
...
+                            crate_name,
+                            paths.next().unwrap().display()
+                        );
+                        let padding = 8 + crate_name.len();
+                        for path in paths {
+                            s.push_str(&format!(
+                                "\n{:>padding$}",
+                                path.display(),
+                                padding = padding
+                            ));
```
camelid, line 932: I switched to `push_str(&format!(...))` from `write!` because `std::io::Write` is already in scope, so importing `std::fmt::Write` causes a name conflict. This code is in the error path, so the performance difference shouldn't matter.

compiler/rustc_metadata/src/locator.rs
```diff
@@ -912,21 +912,26 @@ impl CrateError {
...
+                        let mut paths = lib.source.paths();
+
+                        // This `unwrap()` should be okay because there has to be at least one
+                        // source file. `CrateSource`'s docs confirm that too.
+                        let mut s = format!(
+                            "\ncrate `{}`: {}",
+                            crate_name,
+                            paths.next().unwrap().display()
+                        );
+                        let padding = 8 + crate_name.len();
```
camelid, line 926: This padding doesn't seem correct to me. It's the same as before, but `` crate ` `` is 7 chars and `` `: `` is 3 chars (backtick, colon, space). Together, that makes 10 characters, not 8. Additionally, I don't think the format specifier `{:>padding$}` is correct for paths shorter than `padding`.

Should I change it to `push_str` the padding, and then push the path, rather than using format specifiers? I think that would be more correct.

compiler/rustc_metadata/src/locator.rs
jyn514: You can do `use io::Write as _` if you like.

src/test/ui/crate-loading/crateresolve1.stderr
```diff
@@ -0,0 +1,13 @@
+error[E0464]: multiple matching crates for `crateresolve1`
+  --> $DIR/crateresolve1.rs:5:1
+   |
+LL | extern crate crateresolve1;
+   | ^^^^^^^^^^^^^^^^^^^^^^^^^^^
+   |
+   = note: candidates:
+           crate `crateresolve1`: $TEST_BUILD_DIR/crate-loading/crateresolve1/auxiliary/libcrateresolve1-1.dylib
+           crate `crateresolve1`: $TEST_BUILD_DIR/crate-loading/crateresolve1/auxiliary/libcrateresolve1-2.dylib
+           crate `crateresolve1`: $TEST_BUILD_DIR/crate-loading/crateresolve1/auxiliary/libcrateresolve1-3.dylib
```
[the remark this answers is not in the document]
petrochenkov: >Also, slashes vs. backslashes may need to be taken into account for the normalizing...

IIRC, compiletest may do that automatically, since it's a very common compatibility issue.

camelid:
Looks like tests failed for two reasons: (1) https://github.com/rust-lang/rust/pull/89587#discussion_r722861796 and (2) inconsistent ordering of candidates. For (1), I just need to add `normalize-stderr-test`. For (2), I'm thinking I could just sort the candidate paths. Would that be reasonable?

camelid:
(1) should be fixed now.

compiler/rustc_error_codes/src/error_codes/E0464.md
```diff
@@ -0,0 +1,11 @@
+The compiler was unable to determine which version of a crate dependency should
+be used, due to multiple potential versions being available.
+
+For example, this error can occur if you import a crate named `dependency`, but
+you specify multiple crates with the name `dependency` in your `Cargo.toml`. Or,
+if you're not using Cargo, this error can occur if you pass multiple `--extern`
+flags for different versions of a crate library file.
```
camelid, line 7: Is this part correct? I feel unsure about when exactly this error occurs and what the cause is. \[Which is why I'm adding an explanation ;)\]

camelid: Would the following be a more accurate description? Let me know if anything is incorrect in the current version or the suggested version :)

```suggestion
The compiler found multiple library files with the requested crate name.

For example, this error can occur if you import a crate named `dependency`, but
you specify multiple crates with the name `dependency` in your `Cargo.toml`. Or,
if you're not using Cargo, this error can occur if you pass multiple `--extern`
flags where the library files have the same crate name (not the same as the
filename).
```

camelid:
Ok, I think the latest commit should fix the bors failure.

camelid:
Does `// only-x86_64` apply to the host or the target? In the case of WASM, we need to detect it based on the target to ignore the test properly.

PR #83485, opened 2021-03-25
Mark asm tests as requiring LLVM 10.0.1

Amanieu:

PR #89591, opened 2021-10-06
fix: alloc-optimisation is only for rust llvm

infinity0:
As discussed at the bottom of #83485.

On a separate note I'll take this chance ask, is it worth pulling in that patch (to recognise `__rust_dealloc`) into Debian's system LLVM? The main factors for us to consider would be (1) is the optimisation significant and (2) is there not any significant negative impact to non-rust packages that use LLVM.

### DIFF
--- a/src/test/codegen/alloc-optimisation.rs
+++ b/src/test/codegen/alloc-optimisation.rs
@@ -1,4 +1,5 @@
 //
+// no-system-llvm
 // min-llvm-version: 10.0.1
 // compile-flags: -O
 #![crate_type="lib"]


infinity0:
CC @Amanieu

Amanieu:
There was a [thread](https://rust-lang.zulipchat.com/#narrow/stream/131828-t-compiler/topic/.E2.9C.94.20upstreaming.20LLVM.20patch.20for.20__rust_.20alloc.20functions) on Zulip about upstreaming the LLVM patch for `__rust_alloc`. A quick skim seems to indicate that it was preferred to add a new generic attribute to LLVM which marks a function as being alloc/free-like.

PR #89602, opened 2021-10-06
Test performance impact of removing untracked vtable cache from tcx

michaelwoerister:
See https://github.com/rust-lang/rust/issues/89598 for more information


oli-obk:
This is probably due to https://github.com/rust-lang/rustc-perf/blob/c7680837bb2b0a781ec2d7f1025b155ac64c0bc5/collector/benchmarks/ctfe-stress-4/src/lib.rs#L67 generating the same vtable a couple thousand times.

It's fine to regress this stress test in this way

Issue #48731, opened 2018-03-04
&str should have split_first


SoniEx2:
string slices should have a split_first, just like array slices.

returns -> (char, &str)

kennytm:
This function was called `slice_shift_char` but it was [removed](https://github.com/rust-lang/rust/issues/27754#issuecomment-195478632) because you could do this instead:

```rust
fn slice_shift_char(a: &str) -> Option<(char, &str)> {
    let mut chars = a.chars();
    chars.next().map(|c| (c, chars.as_str()))
}

fn main() {
    assert_eq!(slice_shift_char("hello"), Some(('h', "ello")));
    assert_eq!(slice_shift_char("ĺḿńóṕ"), Some(('ĺ', "ḿńóṕ")));
    assert_eq!(slice_shift_char(""), None);
}
```

steveklabnik:
We don't keep bugs open for small feature requests for libraries like this; big things need an RFC, small things can be PR'd. Thanks!

Stargateur:
> 
> 
> This function was called `slice_shift_char` but it was [removed](https://github.com/rust-lang/rust/issues/27754#issuecomment-195478632) because you could do this instead:
> 
> ```rust
> fn slice_shift_char(a: &str) -> Option<(char, &str)> {
>     let mut chars = a.chars();
>     chars.next().map(|c| (c, chars.as_str()))
> }
> 
> fn main() {
>     assert_eq!(slice_shift_char("hello"), Some(('h', "ello")));
>     assert_eq!(slice_shift_char("ĺḿńóṕ"), Some(('ĺ', "ḿńóṕ")));
>     assert_eq!(slice_shift_char(""), None);
> }
> ```

I don't feel it's a good reason. The same reasoning could be say about slice.

kennytm:
There isn't a `slice` method in `str`

Stargateur:
@kennytm One could say the same for slice `split_first()`:

```rust
fn slice_shift_slice(a: &[u8]) -> Option<(u8, &[u8])> {
    let mut iter = a.iter();
    iter.next().map(|&n| (n, iter.as_slice()))
}
```

kennytm:
`slice::split_first` has been stabilized 1.5.0. We can't retroactively destabilize a std method now even if better alternative exists.

```rust
fn split_first<T>(a: &[T]) -> Option<(&T, &[T])> {
    match a {
        [head, tail @ ..] => Some((head, tail)),
        [] => None,
    }
}
```

[...]

PR #89603, opened 2021-10-06
libcore: add `str::first`, `str::split_first`, `str::last` and `str::split_last` methods

eduardosm:
These methods are equivalent to their slice (`[T]`) counterparts.

If this PR is accepted, I will create a tracking issue and update the `issue` value accordingly.

### DIFF
--- a/library/core/src/str/mod.rs
+++ b/library/core/src/str/mod.rs
@@ -173,16 +173,16 @@ impl str {
     /// # Examples
     ///
     /// ```
-    /// #![feature(str_first_last)]
+    /// #![feature(str_first_last_char)]
     /// let s = "🗻∈🌏";
-    /// assert_eq!(s.first(), Some('🗻'));
+    /// assert_eq!(s.first_char(), Some('🗻'));
     ///
     /// let s = "";
-    /// assert_eq!(s.first(), None);
+    /// assert_eq!(s.first_char(), None);
     /// ```
-    #[unstable(feature = "str_first_last", issue = "none")]
+    #[unstable(feature = "str_first_last_char", issue = "none")]
     #[inline]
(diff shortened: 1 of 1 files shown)


camsteffen:
I would add `_char` to the names. A `str` is `[u8]`, not `[char]`. Consider `chars` is not named `iter`.

jfrimmel:
> I would add `_char` to the names. A `str` is `[u8]`, not `[char]`. 

I disagree. Mentally a `str` is a "list" of characters, that only happens to be stored as bytes. If somebody needs to get the first byte, `s.as_bytes().first()` (etc.) can be used.

> Consider `chars` is not named `iter`.

Yes, but there are other iterators (`bytes()`, `char_indices()`) as well.

eduardosm:
I like the idea of adding the `_char` suffix to the function names. I have pushed a new commit adding that

dtolnay (commented):
Could you provide some more detail on how you're planning to use these methods? Is there maybe an example of real-world open source code you could link to that would benefit from having these?

crlf0710:
@eduardosm Ping from triage, any updates on this?

eduardosm:
> Could you provide some more detail on how you're planning to use these methods? Is there maybe an example of real-world open source code you could link to that would benefit from having these?

I missed these methods while I was writing a lexer, although its code is not public yet. There is also https://github.com/rust-lang/rust/issues/48731.

JohnCSimon:
@eduardosm
Triage: I'm closing this due to inactivity. Please reopen when you are ready to continue with this.  Thanks for your contribution
@rustbot label: +S-inactive

reddevilmidzy:
@eduardosm I tried to reopen this PR, but I cannot open it because the repository has been deleted.

PR #89605, opened 2021-10-06
Fix stabilization version for `bindings_after_at`

camelid:
According to the release notes and its PR milestone, it was stabilized
in 1.56.0.


camelid:
Is it worth backporting this PR to 1.56.0? Maybe not, but I'll nominate so the compiler team can decide.

Issue #78149, opened 2020-10-20
#[must_use] on async fns works on returned `Future` instead of the awaited value


ghost:
I tried this code:

```rust
#[must_use]
async fn foo() -> i32 {
    std::future::pending().await
}

pub fn with_warnings() {
    foo();
}

pub async fn without_warning() {
    foo().await;
}
```

I expected the compiler to warn about the unused return value in `without_warning()`, but the compiler emitted two warnings in `with_warnings()`:
```
warning: unused implementer of `Future` that must be used
 --> src/lib.rs:7:5
  |
7 |     foo();
  |     ^^^^^^
  |
  = note: `#[warn(unused_must_use)]` on by default
  = note: futures do nothing unless you `.await` or poll them

warning: unused return value of `foo` that must be used
 --> src/lib.rs:7:5
  |
7 |     foo();
  |     ^^^^^^
```

[Playground](https://play.rust-lang.org/?version=nightly&mode=debug&edition=2018&code=%23%5Bmust_use%5D%0Aasync%20fn%20foo()%20-%3E%20i32%20%7B%0A%20%20%20%20std%3A%3Afuture%3A%3Apending().await%0A%7D%0A%0Apub%20fn%20with_warnings()%20%7B%0A%20%20%20%20foo()%3B%0A%7D%0A%0Apub%20async%20fn%20without_warning()%20%7B%0A%20%20%20%20foo().await%3B%0A%7D%0A) (tried using nightly 2020-10-18 b1496c6e606dd908dd651ac2cce89815e10d7fc5)

As the *real* return values of `async fn`s are actually `impl Future`s, you may consider this as the expected behavior (I don't agree, and `impl Future`s are [already `#[must_use]`](https://github.com/rust-lang/rust/blob/9832374f6e378971e1a933362cf9781b121bb845/library/core/src/future/future.rs#L28)). If that is the case, there should be a way to make the compiler warn on unused `.await`ed values (when type of the `.await`ed value is not `#[must_use]`).

@rustbot modify labels: A-async-await A-lint -A-diagnostics

tmandry:
I think this is indeed a bug, or at the very least unexpected behavior. It might be challenging to fix, since I don't know of a way to express the desired desugaring.

nikomatsakis:
[...]

PR #89610, opened 2021-10-06
warn on must_use use on async fn's

guswynn:
As referenced in #78149

This only works on `async` fn's for now, I can also look into if I can get `Box<dyn Future>` and `impl Future` working at this level (hir)

### DIFF
--- a/compiler/rustc_passes/src/check_attr.rs
+++ b/compiler/rustc_passes/src/check_attr.rs
@@ -1062,7 +1062,7 @@ impl CheckAttrVisitor<'tcx> {
                     lint.build(
                         "`must_use` attribute on `async` functions \
                               applies to the anonymous `Future` returned by the \
-                              function, not the value within.",
+                              function, not the value within",
                     )
                     .span_label(
                         *span,
--- a/compiler/rustc_passes/src/check_attr.rs
+++ b/compiler/rustc_passes/src/check_attr.rs
@@ -112,6 +112,7 @@ impl CheckAttrVisitor<'tcx> {
                     self.check_default_method_body_is_const(attr, span, target)
                 }
                 sym::must_not_suspend => self.check_must_not_suspend(&attr, span, target),
+                sym::must_use => self.check_must_use(hir_id, &attr, span, target),
                 sym::rustc_const_unstable
                 | sym::rustc_const_stable
                 | sym::unstable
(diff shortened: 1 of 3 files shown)


eholk:
Looks good to me!

Mark-Simulacrum:
I suspect this might just be an oversight in this PR, but it seems like this should be an `unused_attributes` *lint*, not an unconditional warning (and a custom message is still fine).

If so, then I suspect it need not receive T-lang approval... though that is a little murky these days :)

Mark-Simulacrum:
T-lang discussed this in today's meeting and agreed that this can move ahead. We may eventually want to make the must_use have behavior related to return of the `.await`'ing the resulting future, but that's not happening immediately (if ever), so we can move ahead here in the meantime.

We briefly discussed my concern around unused_attributes and generally felt positive about its use here, I think.

guswynn:
@wesleywiser what review is outstanding?

wesleywiser:
@guswynn I believe this is still waiting on Mark's feedback to be addressed:

> it seems like this should be an unused_attributes lint, not an unconditional warning (and a custom message is still fine).

guswynn:
@wesleywiser sorry about the delay, was on a break, this should now be ready!

wesleywiser (approved):
Looks great, thanks @guswynn!

wesleywiser:
> curl: (6) Could not resolve host: ci-mirrors.rust-lang.org

Spurious.

@bors retry

compiler/rustc_passes/src/check_attr.rs
```diff
@@ -1046,6 +1047,37 @@ impl CheckAttrVisitor<'tcx> {
...
+        _target: Target,
+    ) -> bool {
+        let node = self.tcx.hir().get(hir_id);
+        if let Some(fn_node) = node.fn_kind() {
+            if let rustc_hir::IsAsync::Async = fn_node.asyncness() {
+                self.tcx.struct_span_lint_hir(UNUSED_ATTRIBUTES, hir_id, attr.span, |lint| {
+                    lint.build(
+                        "`must_use` attribute on `async` functions \
+                              applies to the anonymous `Future` returned by the \
+                              function, not the value within.",
```
JohnTitor, line 1065: In general a message shouldn't have `.` at the end:

```suggestion
                              function, not the value within",
```

wesleywiser: Thanks @JohnTitor!

src/test/ui/lint/unused/unused-async.stderr
```diff
@@ -0,0 +1,26 @@
+warning: `must_use` attribute on `async` functions applies to the anonymous `Future` returned by the function, not the value within.
```
wesleywiser, line 1:
```suggestion
warning: `must_use` attribute on `async` functions applies to the anonymous `Future` returned by the function, not the value within
```

src/test/ui/lint/unused/unused-async.stderr
```diff
@@ -0,0 +1,26 @@
...
+   |   ^^^^^^^^^^^
+LL |
+LL | / async fn test() -> i32 {
+LL | |     1
+LL | | }
+   | |_- this attribute does nothing, the `Future`s returned by async functions are already `must_use`
+   |
+   = note: `#[warn(unused_attributes)]` on by default
+
+warning: `must_use` attribute on `async` functions applies to the anonymous `Future` returned by the function, not the value within.
```
wesleywiser, line 14:
```suggestion
warning: `must_use` attribute on `async` functions applies to the anonymous `Future` returned by the function, not the value within
```

wesleywiser:
Blessed tests. 

@bors r+

guswynn:
@wesleywiser you are a legend for blessing those tests, sorry this fell off my radar!

wesleywiser:
@guswynn Thanks for working on it!! 🎉

PR #89613, opened 2021-10-06
[EXPERIMENT] Make `CrateSource` an enum

camelid:
This depends on #89587.

In practice, it seems like the compiler doesn't need to try to fetch all of
them.

r? @ghost


camelid:
(Closed and reopened to stop the try build since CI failed.)

camelid:
Ok, this time it got through stage1 std, but it panicked during stage1 rustc build. I'll try https://github.com/rust-lang/rust/pull/89613#discussion_r723761764.

camelid:
Looks like Clippy just needs to be updated.

PR #81358, opened 2021-01-24
Add a check for ASCII characters in to_upper and to_lower

mcastorina:
This extra check has better performance. See discussion here:
https://internals.rust-lang.org/t/to-upper-speed/13896

Thanks to @gilescope for helping discover and test this.

PR #89614, opened 2021-10-07
Update to Unicode 14.0

cuviper:
The Unicode Standard [announced Version 14.0](https://home.unicode.org/announcing-the-unicode-standard-version-14-0/) on September 14, 2021, and this pull request updates the generated tables in `core` accordingly.

This did require a little prep-work in `unicode-table-generator`. First, #81358 had modified the generated file instead of the tool, so that change is now reflected in the tool as well. Next, I found that the "Alphabetic" property in version 14 was panicking when generating a bitset, "cannot pack 264 into 8 bits". We've been using the skiplist for that anyway, so I changed this to fail gracefully. Finally, I confirmed that the tool still created the exact same tables for 13 before moving to 14.


gilescope:
Ideally there's some kind of #![autogen_code] metadata and rust analyzer
warns me I am altering autogenerated code.

camelid:
> Or maybe a rust-highfive mention?

That's a good idea; a rust-highfive ping should be quick to set up and sufficient since the file is not changed often.

PR #89619, opened 2021-10-07
Turn vtable_allocation() into a query

michaelwoerister:
This PR removes the untracked vtable-const-allocation cache from the `tcx` and turns the `vtable_allocation()` method into a query.

The change is pretty straightforward and should be backportable without too much effort.

Fixes https://github.com/rust-lang/rust/issues/89598.


nagisa:
Implementation LGTM; r=me once the test is adjusted as necessary.

PR #89621, opened 2021-10-07
doc: guarantee call order for sort_by_cached_key

digama0:
`slice::sort_by_cached_key` takes a caching function `f: impl FnMut(&T) -> K`, which means that the order that calls to the caching function are made is user-visible. This adds a clause to the documentation to promise the current behavior, which is that `f` is called on all elements of the slice from left to right, unless the slice has len < 2 in which case `f` is not called.

For example, this can be used to ensure that the following code is a correct way to involve the index of the element in the sort key:
```rust
let mut index = 0;
slice.sort_by_cached_key(|x| (my_key(index, x), index += 1).0);
```

### DIFF
--- a/library/alloc/src/slice.rs
+++ b/library/alloc/src/slice.rs
@@ -374,7 +374,8 @@ impl<T> [T] {
     /// During sorting, the key function is called only once per element.
     ///
     /// This sort is stable (i.e., does not reorder equal elements) and *O*(*m* \* *n* + *n* \* log(*n*))
-    /// worst-case, where the key function is *O*(*m*).
+    /// worst-case, where the key function is *O*(*m*). If the slice requires sorting,
+    /// the key function is called on all elements of the slice in the original order.
     ///
     /// For simple key functions (e.g., functions that are property accesses or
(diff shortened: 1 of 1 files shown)


yaahc:
I cant think of any reasons we'd want to change the iteration order for how we setup the keys, so I see no reason not to commit to a stable iteration order and am happy to start the FCP.

@rfcbot merge

joshtriplett:
I share @BurntSushi's concern here about potential future restrictions of such a guarantee. I'd like to see if we can guarantee less and still be useful; notably, I'd be in favor of the "at most once" guarantee.

yaahc:
Sorry for the late update on this, but we reviewed this PR during our [libs-api team meeting](https://hackmd.io/3QZn0wUxTQKn9qjzHqvb3Q) on 2021/12/22 and concluded that we'd like to unblock this PR changing to the "at most once" guarantee mentioned by @joshtriplett.

digama0:
Wording has been updated to remove the ordering guarantee and clarify the "at most once" part.

yaahc:
@rfcbot resolve example-usecases

yaahc:
Thank you again @digama0!

@bors r+ rollup

PR #89623, opened 2021-10-07
Always register a tracing subscriber, even if no env var is set

### LINKED ISSUE #76824: Consider making RUSTC_LOG=warn the default (instead of just error).
We don't use `error!`/`warn!`/`info!` a lot, but they come in handy for "debugging `rustc` without a `debug-assertions` build" situations, e.g. we use `info!` to show the function being codegen'd (as it's low-overhead and really helpful for e.g. LLVM crashes).

While `info!` makes sense for off-by-default debugging helpers like that, `error!`/`warn!` seem closer to user-facing diagnostics, where the main difference is them being usable outside of code which has access to a `Session` (either due to running before the `Session` is created, or just in dependencies that have access to `log`/`tracing` but not any `rustc`-specific diagnostics APIs).

However, one surprising aspect is *only* `error!` is enabled by default, while `warn!` *is not*, meaning that, with the current defaults, one is forced to choose between `error!` (which would output the word "error" which sounds more dire than "warning") or `warn!` (which may never be seen by an user).

<hr/>

My usecase for `warn!` is in the upcoming `-Z self-profile` support for using hardware performance counters (which needs to be implemented in the separate `measureme` project), e.g. [this `warn!` instructs the user to open an issue](https://gist.github.com/eddyb/269491b0b59605f39f1d0a9dcf535c4a/025ee76590ad4e24a7dce9754080fb7c2b30df85#file-rdpmc-bench-rs-L565-L570) if they have an unrecognized (newer) CPU, but there's nothing preventing `-Z self-profile` from working (if the newer CPUs aren't drastically different), so showing "error" seems quite inappropriate.

Sadly, I don't think ~~`log`~~ `env_logger` itself will change this behavior, so even if we change `RUSTC_LOG`'s default, `measureme` might still have to use the less-appropriate `error!` if it wants to account for being used in other projects.

cc @Mark-Simulacrum @gdhuper @hawkw

oli-obk:
If nothing is specified, the log crate emits at least error! messages, tracing would, too, but if we skip the subscriber, it can't do that.

fixes #76824

cc @hawkw @eddyb

### DIFF
--- a/compiler/rustc_driver/src/lib.rs
+++ b/compiler/rustc_driver/src/lib.rs
@@ -1253,12 +1253,6 @@ pub fn init_rustc_env_logger() {
 /// tracing crate version. In contrast to `init_rustc_env_logger` it allows you to choose an env var
 /// other than `RUSTC_LOG`.
 pub fn init_env_logger(env: &str) {
-    // Don't register a dispatcher if there's no filter to print anything
-    match std::env::var(env) {
-        Err(_) => return,
-        Ok(s) if s.is_empty() => return,
-        Ok(_) => {}
-    }
     let color_logs = match std::env::var(String::from(env) + "_COLOR") {
         Ok(value) => match value.as_ref() {
             "always" => true,


hawkw (approved):
looks good to me!

PR #89630, opened 2021-10-07
[crater only] stabilize never type

Mark-Simulacrum:
This stabilizes never type and never type fallback, with the goal of running crater to evaluate the current implementations breakage to existing code.

r? @ghost


Mark-Simulacrum:
@craterbot check

Mark-Simulacrum:
@craterbot retry

PR #89632, opened 2021-10-07
Fix docblock code display on mobile

### LINKED ISSUE #89618: `<code>` in docblocks should use "work-wrap" CSS rule
Otherwise, if too long we can't read them:

Screenshot from 2021-10-07 12-01-37

With `word-wrap` rule:

Screenshot from 2021-10-07 12-03-21

GuillaumeGomez:
Fixes https://github.com/rust-lang/rust/issues/89618.

Before:

Screenshot from 2021-10-07 12-01-37

After:

Screenshot from 2021-10-07 20-17-21

r? @jsha

### DIFF
--- a/src/librustdoc/html/static/css/rustdoc.css
+++ b/src/librustdoc/html/static/css/rustdoc.css
@@ -1969,4 +1969,8 @@ details.undocumented[open] > summary::before {
 	.docblock {
 		margin-left: 12px;
 	}
+
+	.docblock code {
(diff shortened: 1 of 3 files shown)


jsha (approved):

dns2utf8:
An alternative would be to insert a very, very long text in the `<code>`
image

GuillaumeGomez:
I prefer to not change the layout too much considering that it's not what's being tested here. ;)

PR #89633, opened 2021-10-07
Show detailed expected/found types in error message when trait paths are the same

### LINKED ISSUE #65230: Compiler should not omit associated type in compile error message
## Environment

- Rust: Nightly 2019-10-07

## Repro

1. Open https://play.rust-lang.org/?version=nightly&mode=debug&edition=2018&gist=933123bb14852662962ab5bf5f2073c8
2. Click run

```rust
struct X<'a>(&'a mut i32);

impl<'a> Iterator for X<'a> {
    type Item = &'a i32;
    fn next(&mut self) -> Option<Self::Item> {
        Some(self.0)
    }
}

fn main() {}
```

## Expected behavior

Compiler should raise an error with error message which explains why lifetime parameters are conflicting clearly

## Actual behavior

It shows following error:

```
error[E0495]: cannot infer an appropriate lifetime due to conflicting requirements
 --> src/main.rs:6:9
  |
6 |         Some(self.0)
  |         ^^^^^^^^^^^^
  |
note: first, the lifetime cannot outlive the anonymous lifetime #1 defined on the method body at 5:5...
 --> src/main.rs:5:5
  |
5 | /     fn next(&mut self) -> Option<Self::Item> {
6 | |         Some(self.0)
7 | |     }
  | |_____^
note: ...so that reference does not outlive borrowed content
 --> src/main.rs:6:14
  |
6 |         Some(self.0)
  |              ^^^^^^
note: but, the lifetime must be valid for the lifetime 'a as defined on the impl at 3:6...
 --> src/main.rs:3:6
  |
3 | impl<'a> Iterator for X<'a> {
  |      ^^
  = note: ...so that the types are compatible:
          expected std::iter::Iterator
             found std::iter::Iterator

error: could not compile `playground`.

```

Confusing part is as follows:

```
  = note: ...so that the types are compatible:
          expected std::iter::Iterator
             found std::iter::Iterator
```

Compiler says 'expected' and 'found' are incompatible but they look the same. This is caused because associated type `Item` is omitted from error output of 'expected' and 'found'. Actually lifetime parameter in the associated type between 'found' and 'expected' are different but compiler hides them.

rhysd:
Fixes #65230.

### Issue solved by this PR

```rust
trait T {
    type U;
    fn f(&self) -> Self::U;
}

struct X<'a>(&'a mut i32);

impl<'a> T for X<'a> {
    type U = &'a i32;
    fn f(&self) -> Self::U {
        self.0
    }
}

fn main() {}
```

Compiler generates the following note:

```
note: ...so that the types are compatible
  --> test.rs:10:28
   |
10 |       fn f(&self) -> Self::U {
   |  ____________________________^
11 | |         self.0
12 | |     }
   | |_____^
   = note: expected `T`
              found `T`
```

This note is not useful since the expected type and the found type are the same.

### How this PR solve the issue

When the expected type and the found type are exactly the same in string representation, the note falls back to the detailed string representation of trait ref:

```
note: ...so that the types are compatible
  --> test.rs:10:28
   |
10 |       fn f(&self) -> Self::U {
   |  ____________________________^
11 | |         self.0
12 | |     }
   | |_____^
   = note: expected `<X<'a> as T>`
              found `<X<'_> as T>`
```

So that a user can notice what was different between the expected one and the found one.

### DIFF
--- a/compiler/rustc_infer/src/infer/error_reporting/mod.rs
+++ b/compiler/rustc_infer/src/infer/error_reporting/mod.rs
@@ -2060,14 +2060,24 @@ impl<'a, 'tcx> InferCtxt<'a, 'tcx> {
                     expected: exp_found.expected.print_only_trait_path(),
                     found: exp_found.found.print_only_trait_path(),
                 };
-                self.expected_found_str(pretty_exp_found)
+                match self.expected_found_str(pretty_exp_found) {
+                    Some((expected, found)) if expected == found => {
+                        self.expected_found_str(exp_found)
+                    }
+                    ret => ret,
+                }
             }
             infer::PolyTraitRefs(exp_found) => {
                 let pretty_exp_found = ty::error::ExpectedFound {
                     expected: exp_found.expected.print_only_trait_path(),
                     found: exp_found.found.print_only_trait_path(),
                 };
-                self.expected_found_str(pretty_exp_found)
+                match self.expected_found_str(pretty_exp_found) {
+                    Some((expected, found)) if expected == found => {
+                        self.expected_found_str(exp_found)
+                    }
+                    ret => ret,
+                }
             }
         }
     }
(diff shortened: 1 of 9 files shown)


rhysd:
@petrochenkov I updated the test to pass with `-Z borrowck=mir`. Would you review this again? I confirmed it passed with the following command locally:

```sh
./x.py test ./src/test/ui/issues/issue-65230.rs --compare-mode nll
```

PR #89636, opened 2021-10-07
Extract rustc_mir_pretty crate from rustc_middle

camsteffen:
This shrinks and removes dependencies from rustc_middle so I'm hoping this is beneficial for bootstrap times. It also just seems sensible and consistent.


oli-obk:
One way to find out where to add inline attributes is to run rustc via cachegrind before and after your change on one of the regressed benchmarks. Then let kcachegrind give you a diff view between those runs. From there it should be obvious what functions were inlined before and aren'tnow, as you'll suddenly see a huge amount of new executions of some functions. Those will then need inlining...

If you have the extra time, could you write down more detailed instructions as you go (with exakt commands that you used for your situation)? I should have done so last time and added it to the guide, but I didn't think of it

camsteffen:
Well it's still a small regression so I don't know if this is desirable.

apiraino:
Mentioned during today's compiler meeting [on Zulip](https://zulip-archive.rust-lang.org/stream/238009-t-compiler/meetings/topic/.5Bweekly.5D.202021-12-23.20.2354818.html#265927269).

In order to move this further on, [this perf. regression ](https://github.com/rust-lang/rust/pull/89636#issuecomment-938136063) should be investigated.
Another question was if adding any inline attributes suggested [in this comment](https://github.com/rust-lang/rust/pull/89636#issuecomment-939231486) were tried and, if yes, which results were observed, if there is any feedback on this.

Finally a question about any possible build time improvement with `-j2` or higher (perf.r-l.o builds with-j1 and 2 threads).

camsteffen:
> In order to move this further on, [this perf. regression ](https://github.com/rust-lang/rust/pull/89636#issuecomment-938136063) should be investigated.
> Another question was if adding any inline attributes suggested [in this comment](https://github.com/rust-lang/rust/pull/89636#issuecomment-939231486) were tried and, if yes, which results were observed, if there is any feedback on this.


Don't we only care about the [most recent](https://github.com/rust-lang/rust/pull/89636#issuecomment-964676741) perf run?
I did try this and I didn't see anything in the cachegrind output diff that looked like a candidate for inlining. I did however submit #89893 and the subsequent perf run looks neutral to me.

> Finally a question about any possible build time improvement with `-j2` or higher (perf.r-l.o builds with-j1 and 2 threads).

Hmm I'm not sure. Should I just time it locally?

Mark-Simulacrum:
The stated motivation in the PR description is that this improves bootstrap times, so before we proceed we should actually have some numbers showing that's the case - current perf.rlo data rather shows the opposite (or at least net neutral/within margin of error).

If we don't observe bootstrap time improvement, it may still make sense to merge as a cleanup/refactoring improving code quality, but we should be clear about that being the motivation rather than bootstrap perf.

Running numbers locally should be good, and looking at x.py check performance would also be good (perf.rlo reflects x.py build performance only).

bjorn3:
perf.r-l.o shows the performance of building each crate individually. It doesn't show the effect of building multiple crates in parallel.

apiraino:
@camsteffen checking if the `s-waiting-on-review` label is correct here. Do you need clarifications or further review after the latest comments? I'll tentatively set this to be waiting on author, but please ask for another check if it's the case, thanks!

@rustbot label -S-waiting-for-review +S-waiting-on-author

Dylan-DPC:
Closing this due to inactivity

Issue #89384, opened 2021-09-30
Rustdoc search should support space-separated queries (in additon to `::`-separated)


thomcc:
I often try to do searches on rustdoc as `foo bar` rather than `foo::bar`, since it's easier to type and to me it feels more natural, especially in situations where the path isn't exact (e.g. see example 2 below).

Unfortunately, this pretty much never works, and it would be nice if it did. Concrete examples:

1. I'd like it if searching [`vec truncate`](https://doc.rust-lang.org/nightly/std/?search=vec%20truncate) brought up links to [std::vec::Vec::trunctate](https://doc.rust-lang.org/nightly/std/vec/struct.Vec.html#method.truncate) and such, the way searching [`vec::truncate`](https://doc.rust-lang.org/nightly/std/?search=vec%3A%3Atruncate) does.

2. I'd like it if [`arch mm_cvt`](https://doc.rust-lang.org/nightly/std/?search=arch%20mm_cvt) brought up the list of `_mm_cvt` functions in core::arch, as searching [`arch::mm_cvt`](https://doc.rust-lang.org/nightly/std/?search=arch%3A%3Amm_cvt) does

    (Note that in this case, the `::` path isn't really actually a thing that appears anywhere in the results, which is part of why I think spaces for this can be more natural)

3. Ideally when spaces are used, the order wouldn't be important. e.g. [`mm_cvt x86_64`](https://doc.rust-lang.org/nightly/std/?search=x86_64%20mm_cvt) would work even though `mm_cvt::x86_64` doesn't — it would ideally something like [this](https://doc.rust-lang.org/nightly/std/?search=x86_64%3A%3Amm_cvt).

    That said, this might be better addressed later, I can't really make heads or tales of most of what the code is doing, so I don't know how big of a deal it is.

I think this is a matter of replacing some of the use of `"::"` in [this file](https://github.com/rust-lang/rust/blob/master/src/librustdoc/html/static/js/search.js), but I'm not completely sure all the changes that are needed.

Dylan-DPC:
fixed by https://github.com/rust-lang/rust/pull/88268

PR #89637, opened 2021-10-07
Use a regex instead of hard coded :: for rustdoc search.

accusitive:
Hi, I saw #89384 and thought it was easy enough for a first contribution.
I replaced the hard coded `::` separator with a regex that supports white space and `::`.


GuillaumeGomez:
No problem. Let's hope I'll be able to open it not too far in the future so all these parsing issues can be resolved at once.

PR #89646, opened 2021-10-07
Update Miri

### LINKED ISSUE #89612: `miri` no longer builds after rust-lang/rust#89608
Hello, this is your friendly neighborhood mergebot.
After merging PR rust-lang/rust#89608, I observed that the tool miri has failing tests.
A follow-up PR to the repository https://github.com/rust-lang/miri is needed to fix the fallout.

cc @Manishearth, do you think you would have time to do the follow-up work?
If so, that would be great!

camelid:
Fixes #89612.

r? @RalfJung


camelid:
I've never updated a tool submodule before, so let me know if I did this correctly :)

RalfJung:
It looks just right, thanks a lot for the help. :)
@bors r+ p=1

PR #89651, opened 2021-10-07
Add `Poll::ready` and revert stabilization of `task::ready!`

ibraheemdev:
This PR adds an inherent `ready` method to `Poll` that can be used with the `?` operator as an alternative to the `task::ready!` macro:
```rust
let val = ready!(fut.poll(cx));
let val = fut.poll(cx).ready()?;
```

I think this form is a nice, non-breaking middle ground between changing the `impl Try for Poll`, and adding a separate macro. It looks better than `ready!` in my opinion, and it composes well:

```rust
let elem = ready!(fut.poll(cx)).pop().unwrap();
let elem = fut.poll(cx).ready()?.pop().unwrap();
```

The planned stabilization of `ready!` in 1.56 has been reverted because I think this alternate approach is worth considering.

r? rust-lang/libs

### DIFF
--- a/RELEASES.md
+++ b/RELEASES.md
@@ -64,7 +64,6 @@ Stabilised APIs
 - [`VecDeque::shrink_to`]
 - [`HashMap::shrink_to`]
 - [`HashSet::shrink_to`]
-- [`task::ready!`]
 
 These APIs are now usable in const contexts:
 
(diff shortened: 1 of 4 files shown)


tmandry:
This is quite nice. Worth considering at least.

camelid:
(Just noting that 1.56 release notes would likely need to be updated.)

Mark-Simulacrum:
Please note that if you choose to revert things landing in 1.56, the timeline on that is quite short -- ideally PRs for that are merged into master and tagged with beta-nominated + beta-accepted by mid-next week -- we have just 2 weeks before the release must ship.

dtolnay:
@rfcbot poll libs-api Retract `ready!` stabilization from 1.56.0?

joshtriplett:
While I don't see much harm in having both, this does seem better in every way.

dtolnay (approved):

dtolnay:
@bors r+
@rustbot label +beta-nominated

Nominating only the 2nd and 3rd commit (5f7e7d2e93afd9478856d3dc026c6923a0f21641 + 7a7dfa8b67364be9a6fcd0d5876ed1c55b55843b) to be backported.

qm3ster:
### pls remov `Try` impls for `Poll<Result<T, F>>` and `Poll<Option<Result<T, E>>>` before anyone sees REEEEEEE

(someone had to make this comment, I think I expressed what many of us are thinking eloquently enough, now it can be used for upvotes)

### Q&A Section:

Q: Was this the right PR/Issue to make this comment?
A: Yes.

PR #89652, opened 2021-10-07
Add LLVM CFI support to the Rust compiler

rcvalle:
This PR adds LLVM Control Flow Integrity (CFI) support to the Rust compiler. It initially provides forward-edge control flow protection for Rust-compiled code only by aggregating function pointers in groups identified by their number of arguments.

Forward-edge control flow protection for C or C++ and Rust -compiled code "mixed binaries" (i.e., for when C or C++ and Rust -compiled code share the same virtual address space) will be provided in later work as part of this project by defining and using compatible type identifiers (see Type metadata in the design document in the tracking issue #89653).

LLVM CFI can be enabled with -Zsanitizer=cfi and requires LTO (i.e., -Clto).

Thank you, @eddyb and @pcc, for all the help!

### DIFF
--- a/compiler/rustc_codegen_gcc/src/builder.rs
+++ b/compiler/rustc_codegen_gcc/src/builder.rs
@@ -915,6 +915,16 @@ impl<'a, 'gcc, 'tcx> BuilderMethods<'a, 'tcx> for Builder<'a, 'gcc, 'tcx> {
         // TODO(antoyo)
     }
 
+    fn type_metadata(&mut self, _function: RValue<'gcc>, _typeid: String) {
+        // Unsupported.
+    }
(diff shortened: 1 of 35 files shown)


nagisa:
Huh, there seem to be more caveats with `CFI` compared to other sanitizers as far as I can tell at a glance. I would like to see this be landed behind a `-Zunstable-options` or somesuch initially to give us more leeway in adjusting the specifics of the implementation and interactions with other ecosystems.

nagisa:
This does look pretty reasonable first step in the implementation.

@bors r+

rcvalle:
Thank you for your time, @nagisa!

Issue #84175, opened 2021-04-13
Share drpp glue too when -Zshare-generics is enabled


bjorn3:
cc https://github.com/bevyengine/bevy/pull/1914

csmoe:
https://github.com/rust-lang/rust/pull/68414 seems already shared there? @bjorn3

bjorn3:
```
$ cat > lib.rs
pub struct Foo(pub Vec<u8>);
impl Drop for Foo {
    fn drop(&mut self) {}
}
$ rustc +nightly --crate-type dylib lib.rs -Zshare-generics -Cprefer-dynamic
$ cat > main.rs
fn main() {
    lib::Foo(vec![]);
}
$ rustc +nightly --crate-type bin main.rs -Zshare-generics --extern lib=liblib.so -Cprefer-dynamic
$ objdump -d ./main | grep Foo
00000000000013c0 <_ZN4core3ptr29drop_in_place$LT$lib..Foo$GT$17ha6b7547f0832ce86E>:
    13c8:       48 8b 0d f9 2b 00 00    mov    0x2bf9(%rip),%rcx        # 3fc8 <_ZN50_$LT$lib..Foo$u20$as$u20$core..ops..drop..Drop$GT$4drop17hcc626ef64eb19045E>
    13d1:       eb 00                   jmp    13d3 <_ZN4core3ptr29drop_in_place$LT$lib..Foo$GT$17ha6b7547f0832ce86E+0x13>
    13dc:       eb 27                   jmp    1405 <_ZN4core3ptr29drop_in_place$LT$lib..Foo$GT$17ha6b7547f0832ce86E+0x45>
    13e7:       eb 10                   jmp    13f9 <_ZN4core3ptr29drop_in_place$LT$lib..Foo$GT$17ha6b7547f0832ce86E+0x39>
    13f7:       eb e5                   jmp    13de <_ZN4core3ptr29drop_in_place$LT$lib..Foo$GT$17ha6b7547f0832ce86E+0x1e>
    1531:       e8 8a fe ff ff          callq  13c0 <_ZN4core3ptr29drop_in_place$LT$lib..Foo$GT$17ha6b7547f0832ce86E>
```

The `core::ptr::drop_in_place<lib::Foo>` shouldn't be codegened in `main`, but in `liblib.so` if `-Zshare-generics` is used. The `<lib::Foo as core::ops::Drop>::drop` is codegened in `liblib.so` as intended though.

Kobzol:
[...]

PR #30900, opened 2016-01-14
Implement the translation item collector

michaelwoerister:
The purpose of the translation item collector is to find all monomorphic instances of functions, methods and statics that need to be translated into LLVM IR in order to compile the current crate.

So far these instances have been discovered lazily during the trans path. For incremental compilation we want to know the set of these instances in advance, and that is what the trans::collect module provides.
In the future, incremental and regular translation will be driven by the collector implemented here.

r? @nikomatsakis 
cc @rust-lang/compiler 
# Translation Item Collection

This module is responsible for discovering all items that will contribute to
to code generation of the crate. The important part here is that it not only
needs to find syntax-level items (functions, structs, etc) but also all
their monomorphized instantiations. Every non-generic, non-const function
maps to one LLVM artifact. Every generic function can produce
from zero to N artifacts, depending on the sets of type arguments it
is instantiated with.
This also applies to generic items from other crates: A generic definition
in crate X might produce monomorphizations that are compiled into crate Y.
We also have to collect these here.

The following kinds of "translation items" are handled here:
- Functions
- Methods
- Closures
- Statics
- Drop glue

The following things also result in LLVM artifacts, but are not collected
here, since we instantiate them locally on demand when needed in a given
codegen unit:
- Constants
- Vtables
- Object Shims
## General Algorithm

Let's define some terms first:
- A "translation item" is something that results in a function or global in
  the LLVM IR of a codegen unit. Translation items do not stand on their
  own, they can reference other translation items. For example, if function
  `foo()` calls function `bar()` then the translation item for `foo()`
  references the translation item for function `bar()`. In general, the
  definition for translation item A referencing a translation item B is that
  the LLVM artifact produced for A references the LLVM artifact produced
  for B.
- Translation items and the references between them for a directed graph,
  where the translation items are the nodes and references form the edges.
  Let's call this graph the "translation item graph".
- The translation item graph for a program contains all translation items
  that are needed in order to produce the complete LLVM IR of the program.

The purpose of the algorithm implemented in this module is to build the
translation item graph for the current crate. It runs in two phases:
1. Discover the roots of the graph by traversing the HIR of the crate.
2. Starting from the roots, find neighboring nodes by inspecting the MIR
   representation of the item corresponding to a given node, until no more
   new nodes are found.
### Discovering roots

The roots of the translation item graph correspond to the non-generic
syntactic items in the source code. We find them by walking the HIR of the
crate, and whenever we hit upon a function, method, or static item, we
[...]

PR #89660, opened 2021-10-08
[Experiment] Force to generate drop glue of ADT locally

csmoe:
cc https://github.com/rust-lang/rust/issues/84175#issuecomment-927983624
r? @ghost
@michaelwoerister May I have a perf run on this? If so, how can I perf on `share-generics` vs `no-share-generics`.

### DIFF
--- a/compiler/rustc_monomorphize/src/collector.rs
+++ b/compiler/rustc_monomorphize/src/collector.rs
@@ -1167,13 +1167,16 @@ impl ItemLikeVisitor<'v> for RootCollector<'_, 'v> {
             | hir::ItemKind::Struct(_, ref generics)
             | hir::ItemKind::Union(_, ref generics) => {
                 if generics.params.is_empty() {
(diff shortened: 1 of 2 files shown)


Kobzol:
Nice! I wanted to try this, but didn't know how to do it properly. It won't help with generic drop glues, but I'm interested whether it can help at least a bit with non-generic ones.

csmoe:
@oli-obk git blame shows that you implemented the [drop glue collection](https://github.com/rust-lang/rust/blame/48291a9dda821a81957b44034572c0ec62728f35/src/librustc_mir/monomorphize/collector.rs#L1007) in #30900 
Does my diff match my purpose here?

oli-obk:
That PR is by @michaelwoerister, but it's also ages ago...

jrmuizel:
What's needed to get this merged? An investigation of the performance regression?

PR #89662, opened 2021-10-08
Disable new pass manager by default for the s390x targets.

hkratz:
With this change target specifications can opt in to provide a default setting for using the new LLVM pass manager via `TargetOptions::new_llvm_pass_manager`. In this PR it is only set to `Some(false)` for s390x, disabling the new pass manager due to some major performance problems with it (see #89609). 

If no default is specified `rustc_codegen_llvm::back::write::should_use_new_llvm_pass_manager()` continues to enable the new pass manager iff the LLVM  version is 13 or higher.

The command line option `-Z new-llvm-pass-manager=(yes|no)` continues to override default values and should_use_new_llvm_pass_manager() logic.

This is meant as a temporary measure until the problems with s390x are fixed.

### DIFF
--- a/compiler/rustc_codegen_ssa/src/back/write.rs
+++ b/compiler/rustc_codegen_ssa/src/back/write.rs
@@ -254,7 +254,11 @@ impl ModuleConfig {
             },
 
             inline_threshold: sess.opts.cg.inline_threshold,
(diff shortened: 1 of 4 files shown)


hkratz:
Closing in favor of #89666. I did not realize that adding a field to the target spec would make it public contract. As this is only meant as a stop-gap measure exposing that is not a good idea.

PR #89665, opened 2021-10-08
Ensure that pushing empty path works as before on verbatim paths

seanyoung:
Fixes: https://github.com/rust-lang/rust/issues/89658

Signed-off-by: Sean Young <sean@mess.org>

### DIFF
--- a/library/std/src/path.rs
+++ b/library/std/src/path.rs
@@ -1208,6 +1208,9 @@ impl PathBuf {
     /// * if `path` has a root but no prefix (e.g., `\windows`), it
     ///   replaces everything except for the prefix (if any) of `self`.
     /// * if `path` has a prefix but no root, it replaces `self`.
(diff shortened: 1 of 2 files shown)


dylni:
Thanks! Do you think the documentation for [`PathBuf::push`](https://doc.rust-lang.org/nightly/std/path/struct.PathBuf.html#method.push) should also be updated to give some information about [rust-lang/89270](https://github.com/rust-lang/rust/pull/89270)?

camelid:
> See: #89658

Should this PR be marked as fixing that issue?

seanyoung:
> > See: #89658
> 
> Should this PR be marked as fixing that issue? (If so, change "See:" to "Fixes".)

Good point. I've changed it. I've also tested it more, all good as far as I can see.

joshtriplett:
r? rust-lang/libs-api

PR #89666, opened 2021-10-08
Default to disabling the new pass manager for the s390x arch targets.

hkratz:
This hack disables the new LLVM pass manager by default for s390x arch targets until the performance issues are fixed (see #89609). The command line option `-Z new-llvm-pass-manager=(yes|no)` continues to take precedence over this default.


cuviper:
> auto (dist-s390x-linux, ubuntu-latest-xl)
> succeeded 2 hours ago in **41m 27s** 

:tada:

PR #89668, opened 2021-10-08
Cfg hide more conditions for core and alloc

Urgau:
Fixes https://github.com/rust-lang/rust/issues/89663

Before:
image

After: 
image

*Same for alloc*

r? @GuillaumeGomez

### DIFF
--- a/library/alloc/src/lib.rs
+++ b/library/alloc/src/lib.rs
@@ -69,7 +69,12 @@
 )]
 #![cfg_attr(
     not(bootstrap),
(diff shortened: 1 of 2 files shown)


GuillaumeGomez (approved):

GuillaumeGomez:
It does indeed: https://doc.rust-lang.org/nightly/alloc/index.html

@Urgau: Please add the fix there as well.

GuillaumeGomez (approved):

sugar700 (approved):

PR #89670, opened 2021-10-08
Improve `std::thread::available_parallelism` docs

yoshuawuyts:
_Tracking issue: https://github.com/rust-lang/rust/issues/74479_

This PR reworks the documentation of `std::thread::available_parallelism`, as requested [here](https://github.com/rust-lang/rust/pull/89324#issuecomment-934343254).

## Changes

The following changes are made:

- We've removed prior mentions of "hardware threads" and instead centers the docs around "parallelism" as a resource available to a program.
- We now provide examples of when `available_parallelism` may return numbers that differ from the number of CPU cores in the host machine.
- We now mention that the amount of available parallelism may change over time.
- We make note of which platform components we don't take into account which more advanced users may want to take note of.
- The example has been updated, which should be a bit easier to use.
- We've added a docs alias to `num-cpus` which provides similar functionality to `available_parallelism`, and is one of the most popular crates on crates.io.

---

Thanks!

r? @BurntSushi

### DIFF
--- a/library/std/src/thread/mod.rs
+++ b/library/std/src/thread/mod.rs
@@ -1427,39 +1427,76 @@ fn _assert_sync_and_send() {
     _assert_both::<Thread>();
 }
 
-/// Returns the number of hardware threads available to the program.
+/// Returns an estimate of the default amount of parallelism a program should use.
 ///
-/// This value should be considered only a hint.
+/// Parallelism is a resource. A given machine provides a certain capacity for
+/// parallelism, i.e., a bound on the number of computations it can perform
+/// simultaneously. This number often corresponds to the amount of CPUs or
+/// computer has, but it may diverge in various cases.
 ///
-/// # Platform-specific behavior
+/// Host environments such as VMs or container orchestrators may want to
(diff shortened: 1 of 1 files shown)


shoffmeister:
For the sake of maintainability (technical documentation debt), I'd recommend to simply omit the whole section on `The following platform limitations currently apply:`.

Instead, I'd suggest wording to the effect of `Platform-specific implementations will come with further individual limitations. Please review the respective target platform implementation for details.` This is clearly less expressive than the existing text, but also makes less promises.

IMHO, anybody who is interested in the "fully correct picture on parallelism" will *have* to study the implementation, anyway (or get this from highly specialized implementation)

nagisa (commented):
Really nice.

BurntSushi:
It looks like @joshtriplett is all over this! Thank you. I think my concern is totally addressed here, so I'll hand things over if you don't mind.

r? @joshtriplett

yoshuawuyts:
The last run contained a typo in the example which caused the test to fail. Can someone re-run bors again?

PR #89675, opened 2021-10-08
Re-use TypeChecker instead of passing around some of its fields 

oli-obk:
In the future (for lazy TAIT) we will need more of its fields, but even ignoring that, this change seems reasonable on its own to me.


davidtwco (approved):
LGTM!

Issue #85748, opened 2021-05-27
Tracking Issue for is_symlink


maxwase:
Feature gate: `#![feature(is_symlink)]`

### Public API

```rust
// for `Path` and `Metadata`
pub fn is_symlink(&self) -> bool
```

### Steps / History

- [x] Implementation: #85747 
- [x] Final commenting period (FCP)
- [ ] Stabilization PR

### Unresolved Questions

None

marcospb19:
My thoughts on the unresolved questions:

> Does exists() or try_exists() method work correctly? How do we implement these for symlinks?

Every function in `std::fs` is supposed to follow links, except for the ones who contain `link` name in it:
1. `fs::metadata_symlink`
2. `fs::hard_link`
3. `fs::read_link`

Assuming this pattern is consistent in all library stable functions, I'd say `exists()` and `try_exists()` are implemented correctly, they fail if there is a broken link at the path.

If we had `exists_symlink()` and `try_exists_symlink()`, I'd assume they both return true for broken links.

---

> Linked issues: fs::Metadata::is_dir() and is_file() should be deprecated #76487

I created that issue when I was confused why there wasn't a `is_symlink` available for `fs::Metadata`, as seen by the feedback given, people prefer those methods to not be deprecated, as they are useful shorthands to metadata's filetype methods, I agree, and I believe that `Path::is_symlink` and `Metadata::is_symlink` would be just as useful.

Issue #83186, opened 2021-03-16
Tracking Issue for fs_try_exists


Kixunil:
Feature gate: `#![feature(path_try_exists)]`

This is a tracking issue for `try_exists()` method on `std::path::Path`.

This method is similar to `exists()` method except it does **not** silently ignore errors that made it impossible to find out if the path actually exists (e.g. lack of permission on parent dir). Thus it naturally returns `io::Result<bool>` instead of just `bool`.

### Public API

```rust
mod fs {
    pub fn try_exists<P: AsRef<Path>>(path: P) -> Result<bool>;
}

impl Path {
    #[stable(since = "1.63.0")]
    pub fn try_exists(&self) -> io::Result<bool>;
}
```

### Steps / History

- https://github.com/rust-lang/rust/pull/81822
- https://github.com/rust-lang/rust/pull/85060
- https://github.com/rust-lang/rust/pull/97912

### Unresolved Questions

- None yet.

ChrisDenton:
Could `try_exists` be added as a free function to `std::fs` and then the `std::path::Path` method forwards to that? This would be similar to the `metadata` function.

My motivation for this is to make it easier for platforms to override the `metadata` based implementation.

Kixunil:
In the internals discussion @matklad pointed out that `is_file()` and `is_dir()` have same issue. I believe it may be even worse in their case. I wonder would it be reasonable to add those method within the scope of this feature or as a separate change?

roblabla:
Since https://github.com/rust-lang/rust/pull/85060, there's now `std::fs::try_exists` but no `std::fs::exists` (the method only exists on `Path`). This is somewhat confusing, especially since the documentation for `std::fs::try_exists` [says](https://doc.rust-lang.org/stable/std/fs/fn.try_exists.html):

> As opposed to the `exists()` method, this one doesn’t silently ignore errors unrelated to the path not existing.

Without linking the appropriate exists method.

I think if we're going to have an `std::fs::try_exists`, we'll want to make an `std::fs::exists` method as well for symmetry.

ChrisDenton:
I think there are a number of options here:

* Add `std::fs::exists` as you suggest.
* Don't stabilize `std::fs::try_exists`. I'd still want there to be some kind of `fs::try_exists` so that different platforms can specialize the implementation. But that can be purely internal to the standard library.
[...]

Issue #76487, opened 2020-09-08
fs::Metadata::is_dir() and is_file() should be deprecated


marcospb19:
## Resume
I'm personally confused about the existence of

`fs::Metadata::is_dir()` and
`fs::Metadata::is_file()`

At the same time that we DON'T have a
`fs::Metadata::is_symlink()`

It doesn't make a lot of sense

## Context before we go:
`fs::Metadata::file_type()` returns a `fs::FileType`

`fs::FileType` just contains 3 functions, 2 of them can be accessed by `fs::Metadata`, they are `is_dir()` and `is_file()` (yes, same names!).

But, keep in mind that they are all mutually exclusive, that is, a file can never be of _directory type_ and _file type_, or _file type_ and _symlink type_.

## Problems (I'm confused about this O.o)
Considering that we got the metadata from a file:
```rust
let metadata = fs::symlink_metadata("file").unwrap();
```

If we want to know if the file is a symbolic link, we have 2 ways:

1. Without acessing `.file_type()`:
```rust
let is_symlink: bool = !(metadata.is_dir() || metadata.is_file()); // This is wrong, see conversation below
```

2. Getting the `fs::FileType`:
```rust
metadata.is_symlink();
```

So, if the information is already accessible without `file_type()`, why isn't a `is_symlink()` function at `fs::Metadata` just like `is_dir()` and `is_file()`?

Is some other reason that I'm not aware of?

## Conclusion and proposals (need more discussion)
There's no point on having functions in `fs::Metadata` that are **exactly** equal to what is available getting `fs::FileType` by calling `file_type()`.

I originally wanted to propose the removal of `is_dir()` and `is_file()` from `fs::Metadata`, but maybe this would make getting the file_type a little too long, so here are my proposals to fix it sanely.

Ranked by the solutions I personally find the better for the language (please add comments into the debate):

1. Deprecate `is_dir()` and `is_file()` from `fs::Metadata`, and add `fs::file_type` that returns directly the `fs::FileType`.
2. Just deprecate `is_dir()` and `is_file()` from `fs::Metadata`.
3. Add another redundant function, `fs::Metadata::is_symlink()`, to help like `is_file()` and `is_dir()` have been doing since `1.1.0` (we're at `1.4.8` at the time of this writing).

marcospb19:
[...]

PR #89677, opened 2021-10-08
Stabilize `is_symlink()` for `Metadata` and `Path`

maxwase:
I'm not fully sure about `since` version, correct me if I'm wrong 

Needs update after stabilization: [cargo-test-support](https://github.com/rust-lang/cargo/blob/8063672238a5b6c3a901c0fc17f3164692d0be85/crates/cargo-test-support/src/paths.rs#L202)

Linked issue: #85748


joshtriplett:
@rfcbot merge

Kixunil:
Shouldn't this return `Result` in the spirit of `try_exists`? Or is it considered too inconsistent given existing methods return `bool`?

joshtriplett:
@Kixunil Good catch!

@rfcbot concern should-return-result

maxwase:
> Shouldn't this return `Result` in the spirit of `try_exists`? Or is it considered too inconsistent given existing methods return `bool`?

I don't think it should. There are `is_dir` and `if_file` methods with this signature (stable since 1.5.0). This question is also discussed in #76487 and #83186.

Personally, I think it would be inconvenient to return `Result` in this method, but it might be useful to have `try_exists_symlink` to some extend, but to what extent?

For now it is working well [playground](https://play.rust-lang.org/?version=nightly&mode=debug&edition=2018&gist=e4de590d40de85b5c1ee608a4d92d855) 

@Kixunil @joshtriplett

joshtriplett:
I agree that on Metadata we can just return bool. On Path, I think we should return a result.

maxwase:
@joshtriplett, Wouldn't that be illogical to have `is_dir() -> bool` and `is_file() -> bool`, but `is_symlink() -> Result<bool, Error>`?

I see 3 ways to solve this:
1. Deprecate `is_dir` and `is_file` -> bool signatures and replace these with `try_is_...`. This will brake A LOT of code and make these methods so inconviniet to use due to error handling. These methods alwo used in functions that retuns a lot of types of errors, so you can't just use `?`, also you can't use `.unwrap()` because it will be strange to see in production code, so you will use  `.unwrap_or(false)`, as currently implemented. A programmer who wants to handle methadata error can do it by himself.
2. Accept this PR to comlete method hole and continue thinking about it.
3. Accept this PR and open new, not deprecating old methods. New PR will add `try_is...`, so it will be convinient to use foe everyone.

yaahc (changes requested):

yaahc (approved):

WhyNotHugo:
I don't fully grasp the state of this feature; isn't it available as of 1.57?

I'm getting an error indicating that it's unstable:

```
error[E0658]: use of unstable library feature 'is_symlink'
  --> src/main.rs:82:22
   |
82 |         if home_path.is_symlink() {
   |                      ^^^^^^^^^^
   |
   = note: see issue #85748 <https://github.com/rust-lang/rust/issues/85748> for more information

error[E0658]: use of unstable library feature 'is_symlink'
  --> src/main.rs:87:26
   |
87 |     } else if !home_path.is_symlink() {
   |                          ^^^^^^^^^^
   |
   = note: see issue #85748 <https://github.com/rust-lang/rust/issues/85748> for more information
```

ChrisDenton:
If you look at the sidebar you should see a "milestone" tag. This says it'll be stable in 1.58.

WhyNotHugo:
Ah, that completely slipped past me, thanks!

ChrisDenton:
No worries. It's easy to miss if you don't know to look for it.

maxwase:
> I don't fully grasp the state of this feature; isn't it available as of 1.57?
> 
> I'm getting an error indicating that it's unstable:
> 
> ```
> error[E0658]: use of unstable library feature 'is_symlink'
>   --> src/main.rs:82:22
>    |
> 82 |         if home_path.is_symlink() {
>    |                      ^^^^^^^^^^
>    |
>    = note: see issue #85748 <https://github.com/rust-lang/rust/issues/85748> for more information
> 
> error[E0658]: use of unstable library feature 'is_symlink'
>   --> src/main.rs:87:26
>    |
> 87 |     } else if !home_path.is_symlink() {
>    |                          ^^^^^^^^^^
>    |
>    = note: see issue #85748 <https://github.com/rust-lang/rust/issues/85748> for more information
> ```

I don't understand this too. Stable attribute is 1.57 in master, but milestone is 1.58 here. Also this PR was not mentioned in release list
@joshtriplett, can you please explain?

kennytm:
it means the `#[stable]` version property is wrongly tagged 🤷

maxwase:
> it means the `#[stable]` version property is wrongly tagged 🤷 

Should I replace it in new PR?

kennytm:
yes

Issue #84419, opened 2021-04-22
rustdoc: use a templating engine to generate HTML


jsha:
Right now rustdoc generates HTML with a series of writes embedded in the Rust code. This means the Rust code needs to process the elements of a page in the same order that they will be emitted in HTML. It makes it hard to get a holistic view of the page structure. It means we need to take care to make sure tags are balanced, which can be particularly hard when a function has multiple return points. It means in order to make any changes to the HTML, you need a good understanding of the Rust code.

If we use a templating engine, we can separate the templates from the Rust code, improving those problems. It would also mean we could have a mode for rustdoc to read templates at runtime instead of using its compiled-in templates. That would make iterating on HTML improvements easier, because we wouldn't have to wait for a compile cycle for each change.

There are a variety of templating engines out there, but I think [mustache](https://mustache.github.io/) is quite good. In particular the choice to have no logic in templates is important, because otherwise it is tempting to put some logic in templates and some in the driving code, which becomes confusing. In Rust, the [handlebars](https://lib.rs/crates/handlebars) crate looks mature and actively maintained. It implements the [handlebars template language](https://handlebarsjs.com/), which is evidently different from mustache but pretty close.

netroby:
https://github.com/Keats/tera 
Tera is a good engine too. 
It's similar the Django template engine, which the best and friendly engine . 
https://docs.djangoproject.com/en/3.1/topics/templates/
I like Rust, and like Python.

GuillaumeGomez:
> If we use a templating engine, we can separate the templates from the Rust code, improving those problems. It would also mean we could have a mode for rustdoc to read templates at runtime instead of using its compiled-in templates. That would make iterating on HTML improvements easier, because we wouldn't have to wait for a compile cycle for each change.

No, we don't want anything at runtime because it requires running a server, which we definitely don't want. Closing this issue then.

Swatinem:
> we don't want anything at runtime because it requires running a server, which we definitely don't want.

I think you misinterpreted this. Its rather "at documentation generation time". So it would be more like a static site generator. No need to run a server.

I could imagine an unstable `--template-dir` switch as well to control the templates being used, for maximizing iteration time on this.

GuillaumeGomez:
Oh, my bad. I'm not a big fan of the idea of depending on external template engine

jyn514:
> In particular the choice to have no logic in templates is important, because otherwise it is tempting to put some logic in templates and some in the driving code, which becomes confusing. 

[...]

PR #89695, opened 2021-10-09
Move top part of print_item to Tera templates

jsha:
Part of #84419.

This moves the first line of each item page (E.g. `Struct foo::Bar .... 1.0.0 [-][src]` into a Tera template.

I also moved template initialization into its own module and added a small macro to reduce duplication and opportunity for errors.


GuillaumeGomez (approved):

GuillaumeGomez:
Since it can impact performance, let's not put it into the rollup.

@bors: rollup=iffy

Issue #89566, opened 2021-10-05
Diagnostics: improve error message for incorrect inner attribute


ogham:
I was assisting someone new to Rust and they were confused by an error message. This is the complete file of code they had written:

```rust
#![derive(Debug)]
struct Test {
    s: String,
}

fn main() {
}
```

Using rustc 1.55.0 (c8dfcfe04 2021-09-06) and rustc 1.57.0-nightly (f03eb6bef 2021-10-02), the compiler outputs the following error ([playground link](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=148060b0318329961ad9ec4d1e7af0b4)):

```
error: cannot determine resolution for the attribute macro `derive`
 --> src/main.rs:1:4
  |
1 | #![derive(Debug)]
  |    ^^^^^^
  |
  = note: import resolution is stuck, try simplifying macro imports
```

The fix is to use `#[derive]`, an outer attribute, rather than `#![derive]`, an inner attribute, but the compiler's error does not make that clear. And its suggestion to simplify macro imports doesn't help, as from the user's perspective, nothing is being imported in this source file!

I think a more specific error message that checks whether the inner attribute's name would fit as an outer attribute would be helpful here. A suggestion to remove the `!` would be even better. Annoyingly, having `#!` at the start of the file looked correct as it resembles a shebang.

Note that there _is_ a good error for using `#![]` instead of `#[]` in the general case. It even mentions the difference between inner and outer attributes in its message. It was just bad luck that this particular struct happened to be at the very start of the source file, so that error was not displayed.

I also spotted that in issue #67107, which has a _very_ similar piece of example code, the compiler used to provide a suggestion to "try an outer attribute". But, running that example on the latest stable and nightly, it looks like that suggestion is not being produced anymore.

estebank:
Current output:

```
error: cannot determine resolution for the attribute macro `derive`
 --> src/main.rs:1:4
  |
1 | #![derive(Debug)]
  |    ^^^^^^
  |
  = note: import resolution is stuck, try simplifying macro imports

error: `derive` attribute cannot be used at crate level
 --> src/main.rs:1:1
  |
1 | #![derive(Debug)]
  | ^^^^^^^^^^^^^^^^^
  |
help: perhaps you meant to use an outer attribute
  |
1 | #[derive(Debug)]
  |
```

All we need to do is try to silence the first error, if possible, if the second one is emitted.

tom7980:
[...]

Issue #54726, opened 2018-10-01
Tracking issue for custom inner attributes


alexcrichton:
*[View all comments](https://triagebot.infra.rust-lang.org/gh-comments/rust-lang/rust/issues/54726)*

Added in https://github.com/rust-lang/rust/pull/54093 inner attributes are not allowed to be procedural macros right now, and this issue is tracking that feature of procedural macros!

Assistance in filling in this issue would be greatly appreciated!
UPDATE: https://github.com/rust-lang/rust/issues/54726#issuecomment-431931166 below contains the detailed description.

dhardy:
Summary:
```
error[E0658]: non-builtin inner attributes are unstable (see issue #54726)
 --> src/lib.rs:3:1
  |
3 | #![foo]
  | ^^^^^^^
  |
  = help: add #![feature(custom_inner_attributes)] to the crate attributes to enable

error[E0658]: The attribute `foo` is currently unknown to the compiler and may have meaning added to it in the future (see issue #29642)
 --> src/lib.rs:3:4
  |
3 | #![foo]
  |    ^^^
  |
  = help: add #![feature(custom_attribute)] to the crate attributes to enable
```

petrochenkov:
(I'll fill in details about why this is unstable soon.)

petrochenkov:
So, the primary open question about inner attributes is what scope they are resolved in.

For example:
```rust
use xxx::attr;

mod m {
    #![attr]

    use yyy::attr;
}
```

Does `#![attr]` refer to `xxx::attr` or `yyy::attr`?

Right now, `#![attr]` is resolved from outside of the module in the same way as outer attributes (with exception of the crate attributes maybe, not sure).

However, if something is inside a module (or block, etc) it's normally reasonable to expect that it's resolved from inside of that module as well.
IIRC, @nrc argued for applying this logic to attributes too (UPDATE: this happened in https://github.com/rust-lang/rust/issues/41430).

With attributes such resolution from the inside causes problems though.

---
First of all, there's no "inside" before we expand all the macro attributes.

```
mod m {
    #![attr]
    
    /* items */
}
```
is just a token stream, even if it *looks* like module and items, it effectively doesn't exist in the module structure of the crate yet.

`#![attr]` can transform that token stream in any way, for example turn module into a block, or struct, or even into nothing, possibly invalidating our "inner" resolution for `attr` even if we somehow obtained it speculatively from the unexpanded `mod m`.

---
Possible solutions to this problem:

[...]

PR #89701, opened 2021-10-09
Updated error message for accidental uses of derive attribute as a crate attribute

tom7980:
This partially fixes the original issue #89566 by adding derive to the list of invalid crate attributes and then providing an updated error message however I'm not sure how to prevent the resolution error message from emitting without causing the compiler to just abort when it finds an invalid crate attribute (which I'd prefer not to do so we can find and emit other errors).

@petrochenkov I have been told you may have some insight on why it's emitting the resolution error though honestly I'm not sure if we need to worry about fixing it as long as we can provide the invalid crate attribute error also (which happens first anyway)

### DIFF
--- a/compiler/rustc_passes/src/check_attr.rs
+++ b/compiler/rustc_passes/src/check_attr.rs
@@ -1953,7 +1953,12 @@ fn is_c_like_enum(item: &Item<'_>) -> bool {
     }
 }
 
+// FIXME: Fix "Cannot determine resolution" error and remove built-in macros
+// from this check.
 fn check_invalid_crate_level_attr(tcx: TyCtxt<'_>, attrs: &[Attribute]) {
+    // Check for builtin attributes at the crate level
+    // which were unsuccessfully resolved due to cannot determine
+    // resolution for the attribute macro error.
     const ATTRS_TO_CHECK: &[Symbol] = &[
         sym::macro_export,
         sym::repr,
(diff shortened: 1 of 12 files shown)


petrochenkov:
Blocked on https://github.com/rust-lang/rust/pull/91313.

compiler/rustc_passes/src/check_attr.rs
```diff
@@ -1910,11 +1910,15 @@ fn check_invalid_crate_level_attr(tcx: TyCtxt<'_>, attrs: &[Attribute]) {
...
         sym::start,
         sym::rustc_main,
+        sym::derive,
     ];
 
     for attr in attrs {
         for attr_to_check in ATTRS_TO_CHECK {
             if attr.has_name(*attr_to_check) {
+                let lo = attr.span.lo() + BytePos(1);
+                let hi = lo + BytePos(1);
```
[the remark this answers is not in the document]
petrochenkov: You can try using `span_to_snippet` to get the attribute text, then try replacing the `#![` prefix in that text with `#[` for the suggestion text, and only emit the suggestion if both operations succeed.

petrochenkov:
I don't think this is the right fix for `derive`, but we can keep the added suggestion to change the inner attribute to an outer one.

compiler/rustc_passes/src/check_attr.rs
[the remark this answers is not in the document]
tom7980: Okay I'll have a look at doing that tonight - the method `span_suggestion_verbose` does the replacement on the span you give it for you so I suppose I can pass in the `#![` prefix and give it a `#[` to replace it with which should also put the label on the correct span too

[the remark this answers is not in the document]
tom7980: though thinking about it - I need to be able to get the prefix as a span to be able to pass it to `span_suggestion_verbose` system rather than as a `String` as it's signature is `span_suggestion_verbose(span_to_replace: Span, msg: &str, replacement: String, applicability: Applicability)` so if I use `span_to_snippet` I won't get the section of the span I want to replace but a String instead which I then can't pass to error reporting to do the replacement

tom7980:
No longer relying on BytePos to generate the replacement suggestion - the resolution error is still there but I'm not sure exactly what's causing that and I believe you are working on it. 

I think this is more of a stopgap personally and if it would be better to fix it another way it may not even be worth changing but take a look and let me know what you think!

compiler/rustc_passes/src/check_attr.rs
```diff
@@ -1961,20 +1961,35 @@ fn check_invalid_crate_level_attr(tcx: TyCtxt<'_>, attrs: &[Attribute]) {
         sym::automatically_derived,
         sym::start,
         sym::rustc_main,
+        sym::derive,
```
petrochenkov, line 1964: Could you add `global_allocator`, `bench`, `test` and `test_case` to this list as well?
Also add a comment telling that these attributes are built-in attribute macros, so finding them in this position means that they weren't successfully resolved due to the "cannot determine resolution" errors, so we can report the suggestion.
Also add a FIXME telling to fix the "cannot determine resolution" error instead, and then remove the built-in macros from this list.

petrochenkov: And squash commits into one after that.

tom7980: Sure will do that all later today, do you know if it will ever actually be possible to resolve these as inner attributes in that position anyway? 

I assumed that because they can't be used in that position resolving them was impossible but I may be missing some key understanding of resolving.

petrochenkov: >do you know if it will ever actually be possible to resolve these as inner attributes in that position anyway?

Yes, if we skip the current module (and some prelude layers?) during name lookup.
In which context inner attributes should be resolved is a question in general (see https://github.com/rust-lang/rust/issues/54726 for some more details).

tom7980: Cool thanks I think I remember seeing the discussion about where we resolve inner attributes but it will be good to go back and read it with some better context. 

I suppose my thoughts are that attributes like derive almost don't do anything to a module (although maybe I'm wrong there) so I would expect something to fail, though I guess maybe that would be during expansion rather than resolution?

tom7980: Should be all sorted now - have added the other attributes into the test as well so it can be updated once inner attribute resolution is sorted

petrochenkov: >I guess maybe that would be during expansion rather than resolution?

Yes, the macro itself will check its target and report an error if necessary (`fn report_bad_target` in case of `derive`).

PR #89705, opened 2021-10-09
Cfg hide no_global_oom_handling and no_fp_fmt_parse

nbdd0121:
These are unstable sysroot customisation cfg options that only projects building their own sysroot will use (e.g. Rust-for-linux). Most users shouldn't care. `no_global_oom_handling` can be especially annoying since it's applied on many commonly used alloc crate methods (e.g. `Box::new`, `Vec::push`).

r? @GuillaumeGomez

### DIFF
--- a/library/alloc/src/lib.rs
+++ b/library/alloc/src/lib.rs
@@ -73,6 +73,7 @@
         not(test),
         not(any(test, bootstrap)),
         any(not(feature = "miri-test-libstd"), test, doctest),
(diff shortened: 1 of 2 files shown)


GuillaumeGomez (approved):

GuillaumeGomez:
Thanks!

Little precision: can be seen on `Vec::with_capacity` for example.

@bors: r+ rollup

Issue #80536, opened 2020-12-30
MIR body serialization may be a bottleneck


oli-obk:
What's interesting to me is that in the incremental cases, we lose some time to `optimized_mir` as expected but this is often offset by wins in `codegen`, however the other place we regress is time spent dealing with the incremental cache. That suggests to me that our MIR is larger due to inlining which causes more time to be spent during serialization/deserialization. 

It might be worth exploring if there's any optimizations we can make to the serialized `Body` representation.

_Originally posted by @wesleywiser in https://github.com/rust-lang/rust/issues/68828#issuecomment-752782277_

tmiasko:
Decoding of foreign spans is one of hotspots quite unique to builds with inlining: https://github.com/rust-lang/rust/blob/9320b121b5b20d2201c9f5ec40fb3c241fcac6f0/compiler/rustc_metadata/src/rmeta/decoder.rs#L485-L492

PR #89708, opened 2021-10-09
Introduce MIR summary to avoid loading large bodies without inlining them

cjgillot:
MIR inlining loads the bodies of all callees in order to assess whether to inline them. This implies useless decoding of large MIR bodies, which may incur a significant cost (https://github.com/rust-lang/rust/issues/80536).

In order to avoid loading such large MIR, this PR introduces a MIR summary structure with pre-computed statistics. Those statistics are used as a fast-reject path for inlining.

r? @wesleywiser


cjgillot:
Marking as blocked on https://github.com/rust-lang/rust/pull/82280

PR #89709, opened 2021-10-09
Apply clippy suggestions for rustc and core

clemenswasser:


eddyb (changes requested):

pnkfelix:
> @bors rollup=iffy just in case this has some unforeseen perf implications

I think its fair to say this had some unforeseen [perf implications](https://perf.rust-lang.org/compare.html?start=86d6d2b7389fe1b339402c1798edae8b695fc9ef&end=6ae8912a3e7d2c4c775024f58a7ba4b1aedc4073)

We see instruction-count regressions in the 1.1% to 2.7% range in keccak and inflate

matthiaskrgr:
I reverted parts of the merge in https://github.com/rust-lang/rust/pull/89855 can we get a perf run there to see if that fixes the regression?

PR #89711, opened 2021-10-09
Add Read::is_append_only method to optimize read_to_string

jkugelman:
I've added a new method called `is_append_only` to the `Read` trait. It returns true if `read_to_end` and `read_to_string` can be trusted not to read or overwrite any existing data in their input buffer.

These readers use it to signal that they have optimized `read_to_end` methods:

* `[u8]` extends the input buffer to the full slice size all at once instead of doubling its size repeatedly until it's big enough.

* `File` pre-allocates enough space for the full file size, allowing for fewer `read` syscalls and reducing buffer reallocations.

These readers take advantage of it:

* If `BufReader` wraps a `[u8]` or `File` its `read_to_string` will call their optimized `read_to_end`s.

Overall, this fixes a [slow path] where `BufReader::read_to_string` would be forced to read into a side buffer when passed a non-empty input buffer.

[slow path]: https://github.com/rust-lang/rust/pull/89582#discussion_r725158871

### DIFF
--- a/library/std/src/fs.rs
+++ b/library/std/src/fs.rs
@@ -640,10 +640,9 @@ impl Read for File {
         io::default_read_to_end(self, buf)
     }
 
-    // Reserves space in the buffer based on the file size when available.
-    fn read_to_string(&mut self, buf: &mut String) -> io::Result<usize> {
-        buf.reserve(buffer_capacity_required(self));
-        io::default_read_to_string(self, buf)
+    unsafe fn is_append_only(&self) -> bool {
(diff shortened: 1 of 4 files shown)


jkugelman:
Looking for feedback. Is a new trait method sensible? I haven't come up with a simpler way to fix the slow path in `BufReader::read_to_string`. (Could just leave it...)

Previous discussion at https://github.com/rust-lang/rust/pull/89582#discussion_r725158871.

cc @a1phyr

PR #73227, opened 2020-06-11
Allow multiple `asm!` options groups and report an error on duplicate options

camelid:
Fixes #73193

Cc @joshtriplett @Amanieu

- [x] Allow multiple options
- [x] Update existing test
- [x] Add new tests
- [x] Check for duplicate options
- [x] Add duplicate options tests
- [x] Finalize suggestion format for duplicate options error

PR #89713, opened 2021-10-09
Fix ABNF of inline asm options

nbdd0121:
This is the case since #73227.

r? @camelid


camelid:
This LGTM, but I'll r? @Amanieu since I haven't worked on inline assembly in a while.

camelid (approved):

PR #89718, opened 2021-10-10
Add #[must_use] to is_condition tests

jkugelman:
There's nothing insightful to say about these so I didn't write any extra explanations.

Parent issue: #89692


Urgau:
Maybe every function that return a `bool` should have a implicit `must_use` attribute ?

PR #89729, opened 2021-10-10
Add #[must_use] to core and std constructors

jkugelman:
Parent issue: #89692

r? @joshtriplett


jkugelman:
Heads up, I threw `MaybeUninit` into the mix.

```rust
core::mem::MaybeUninit<T>   const fn uninit() -> MaybeUninit<T>;
core::mem::MaybeUninit<T>   const fn uninit_array<const LEN: usize>() -> [Self; LEN];
core::mem::MaybeUninit<T>   fn zeroed() -> MaybeUninit<T>;
```

I left `MaybeUninit::new(val: T)` out since it could be used to nullify `T`'s drop code. Let me know if you think I should add it. I could put a note saying to use `ManuallyDrop::new` instead.

PR #89735, opened 2021-10-10
Stabilize proc_macro::is_available

bjorn3:
Tracking issue: https://github.com/rust-lang/rust/issues/71436

The FCP for the stabilization of `proc_macro::is_available` has completed.


petrochenkov:
Ah, wait, the FCP was on the tracking issue.

PR #89745, opened 2021-10-10
enum variant support

zhamlin:
Proof of concept for enum variant types.

### DIFF
--- a/compiler/rustc_borrowck/src/type_check/mod.rs
+++ b/compiler/rustc_borrowck/src/type_check/mod.rs
@@ -744,6 +744,22 @@ impl<'a, 'b, 'tcx> TypeVerifier<'a, 'b, 'tcx> {
                         PlaceTy { ty: base_ty, variant_index: Some(index) }
                     }
                 }
+                ty::Variant(ty, _) => match ty.kind() {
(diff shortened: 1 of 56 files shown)


oli-obk (changes requested):
First of all, this is amazing. I thought an MVP would be significantly more involved, but you even got surface syntax working from the start.

I left a lot of comments, I do realize this is a draft and as a first iteration the copy paste is expected to get things running quickly. Please let me know how nit-picky you want the review to be. I'm also available on zulip for chatting if you prefer, just open a thread in the compiler team stream.

One general thing is that I think we should also change `TyLayout::for_variant` to return a `ty::Variant` in the type field instead of just returning the original type.

jackh726 (commented):
Oli beat me to the review here, but I'm just as excited :)

Just one comment for now!

oli-obk:
1 and 2 are invisible to users. They are essentially cleanups of the compiler's logic.

For 3 we'd at least need an accepted MCP, from the zulip discussion it sounded like there was nothing speaking against it, so maybe just asking for a "second" to be able to experiment is all that is needed at this stage?

zhamlin:
Would zulip be the best place to discuss this going forward?
I have some questions before I can get started on 1

oli-obk:
`@oli`

PR #89753, opened 2021-10-10
Add #[must_use] to from_value conversions

jkugelman:
I added two methods to the list myself. Clippy did not flag them because they take `mut` args, but neither modifies their argument. 

```rust
core::str           const unsafe fn from_utf8_unchecked_mut(v: &mut [u8]) -> &mut str;
std::ffi::CString   unsafe fn from_raw(ptr: *mut c_char) -> CString;
```

I put a custom note on `from_raw`:

```rust
#[must_use = "call `drop(from_raw(ptr))` if you intend to drop the `CString`"]
pub unsafe fn from_raw(ptr: *mut c_char) -> CString {
```

Parent issue: #89692

r? @joshtriplett

### DIFF
--- a/library/alloc/src/str.rs
+++ b/library/alloc/src/str.rs
@@ -595,6 +595,7 @@ impl str {
 /// assert_eq!("☺", &*smile);
 /// ```
 #[stable(feature = "str_box_extras", since = "1.20.0")]
+#[must_use]
 #[inline]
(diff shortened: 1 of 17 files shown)


BurntSushi:
Is there any particular reason why this was added to `CString::from_raw` but not, say, `Box::from_raw`?

PR #89764, opened 2021-10-11
Fix variant index / discriminant confusion in uninhabited enum branching

tmiasko:
Fix confusion between variant index and variant discriminant. The pass
incorrectly assumed that for `Variants::Single` variant index is the same as
variant discriminant.

r? @wesleywiser

### DIFF
--- a/compiler/rustc_mir_transform/src/uninhabited_enum_branching.rs
+++ b/compiler/rustc_mir_transform/src/uninhabited_enum_branching.rs
@@ -3,8 +3,7 @@
 use crate::MirPass;
 use rustc_data_structures::stable_set::FxHashSet;
 use rustc_middle::mir::{
-    BasicBlock, BasicBlockData, Body, Local, Operand, Rvalue, StatementKind, SwitchTargets,
-    TerminatorKind,
+    BasicBlockData, Body, Local, Operand, Rvalue, StatementKind, SwitchTargets, TerminatorKind,
 };
 use rustc_middle::ty::layout::TyAndLayout;
 use rustc_middle::ty::{Ty, TyCtxt};
@@ -73,10 +72,7 @@ impl<'tcx> MirPass<'tcx> for UninhabitedEnumBranching {
     fn run_pass(&self, tcx: TyCtxt<'tcx>, body: &mut Body<'tcx>) {
         trace!("UninhabitedEnumBranching starting for {:?}", body.source);
 
-        let basic_block_count = body.basic_blocks().len();
-
-        for bb in 0..basic_block_count {
-            let bb = BasicBlock::from_usize(bb);
+        for bb in body.basic_blocks().indices() {
             trace!("processing block {:?}", bb);
 
             let Some(discriminant_ty) = get_switched_on_type(&body.basic_blocks()[bb], tcx, body) else {
--- a/compiler/rustc_mir_transform/src/uninhabited_enum_branching.rs
+++ b/compiler/rustc_mir_transform/src/uninhabited_enum_branching.rs
@@ -55,7 +55,10 @@ fn variant_discriminants<'tcx>(
     match &layout.variants {
         Variants::Single { index } => {
             let mut res = FxHashSet::default();
-            res.insert(index.as_u32() as u128);
+            res.insert(
+                ty.discriminant_for_variant(tcx, *index)
+                    .map_or(index.as_u32() as u128, |discr| discr.val),
+            );
             res
         }
         Variants::Multiple { variants, .. } => variants
--- a/compiler/rustc_mir_transform/src/uninhabited_enum_branching.rs
+++ b/compiler/rustc_mir_transform/src/uninhabited_enum_branching.rs
@@ -71,10 +71,6 @@ fn variant_discriminants<'tcx>(
 
 impl<'tcx> MirPass<'tcx> for UninhabitedEnumBranching {
     fn run_pass(&self, tcx: TyCtxt<'tcx>, body: &mut Body<'tcx>) {
-        if body.source.promoted.is_some() {
-            return;
-        }
-
         trace!("UninhabitedEnumBranching starting for {:?}", body.source);
 
         let basic_block_count = body.basic_blocks().len();


RalfJung:
> @RalfJung could you check if this makes sense from MIR semantics perspective. Thanks.

I'm sorry, but it is rather hard to figure out from a diff of a MIR optimization what the semantic question is.

> The pass incorrectly assumed
that Variants::Single indicates that there is a single inhabited variant.

This (for the 2nd time today^^) reminds me of discussions with @eddyb about enum layouts... yeah, `Single` really means "no more than one", which is indeed rather confusing.

> The pass
incorrectly assumed that for Variants::Single variant index is the same as
variant discriminant.

The `read_discriminant` and `write_discriminant` functions in the CTFE interpreter might be illuminating pieces of code to read. I think they are mostly getting things right these days, though there are still corner cases (https://github.com/rust-lang/rust/issues/89765).

RalfJung:
I guess the question is if we want to consider MIR like this to be UB:

```
enum Void {}

_1: Void

bb1 {
  _2 = discriminant(_1);
  return;
}
```
Due to https://github.com/rust-lang/rust/issues/89765, Miri would currently ICE on such MIR. My preferred solution in that issue would say that asking for the discriminant of `Void` is like asking for the discriminant of `()`: both types do not really have discriminants, so we just return 0.

But I take it from your question that you would prefer to make this UB? What about the case where `_1` has type `!`, or some other non-enum uninhabited type? What about the case where it has type `bool` but is currently initialized to an invalid value? We could say the `Discriminant` MIR primitive requires the place it operates on to be fully initialized, that would trivially give you what you want -- but it might be too aggressive?

RalfJung:
> As far as I understand, the LLVM code generation already makes this assumption. If type is uninhabited, trying to read a discriminant produces undef

It does? Is that some explicit case in our codegen backend?

tmiasko:
If the current state is that there is a basis for the code generation and this MIR pass (in the valididy requirement for operand), but we have yet to decided this more formally, I think, it would make sense to land this, since it is a bug fix and does not make any new assumptions.

At the same time we can independently consider if we want to continue relying on this semantics, and whether the code generation and this MIR pass need to be changed both.

RalfJung:
Fair -- but then Miri should also be changed to actually check the validity invariant in `read_discriminant`, to match the (for now) intended semantics.

wesleywiser:
Sorry for the delay, just getting caught up on GH notifications now. 

> Fair -- but then Miri should also be changed to actually check the validity invariant in read_discriminant, to match the (for now) intended semantics.

@RalfJung do you think we should change Miri before/in conjunction with this PR or are you simply saying we should do that at some point in the future?

RalfJung:
I am doing the Miri change in https://github.com/rust-lang/rust/pull/90895.

wesleywiser:
Thanks @RalfJung!

RalfJung:
It seems like MIR building does not agree with the idea that only valid values have their discriminant read: https://github.com/rust-lang/rust/issues/91029.

RalfJung:
Furthermore, this triggers a Stacked Borrows violation [here](https://github.com/rust-lang/miri/blob/8c700f544ca3ae815b4bb2edb65f85208f78bbb6/tests/run-pass/stacked-borrows/refcell.rs#L75) because `optional.is_some()` now reads the *entire* `Option` value, not just the discriminant, which is in conflict with the mutable reference to the contents of that `Option`.

This seems like a pretty clear sign that reading a discriminant should do the minimal amount of work necessary to determine the discriminant value -- but then there is no justification for [this part of codegen](https://github.com/rust-lang/rust/blob/81117ff930fbf3792b4f9504e3c6bccc87b10823/compiler/rustc_codegen_ssa/src/mir/place.rs#L206-L215). Cc @eddyb

tmiasko:
> It seems like MIR building does not agree with the idea that only valid values have their discriminant read: #91029.

Ah, after a partial move from a variant an elaborated drop still need to read the enum discriminant to determine which variant is active.

> This seems like a pretty clear sign that reading a discriminant should do the minimal amount of work necessary to determine the discriminant value

This suggest an approach which requires only a discriminant to be valid?

RalfJung:
> Ah, after a partial move from a variant an elaborated drop still need to read the enum discriminant to determine which variant is active.

I did not realize we support partial moves out of enums. In principle we do not need to re-read the discriminant since the after the partial move we know which variant it is in, I think... but that is probably tricky to do.

> This suggest an approach which requires only a discriminant to be valid?

Well, that's what we did before. But enums without variants do not have a discriminant so it is trivially valid. Thus the optimization you want to do is unsound under this approach.

tmiasko:
> > This suggest an approach which requires only a discriminant to be valid?
> 
> Well, that's what we did before. But enums without variants do not have a discriminant so it is trivially valid.

Another way to look at this would be to say that they are never valid.

> Sorry for the delay, just getting caught up on GH notifications now.
> 
> > Fair -- but then Miri should also be changed to actually check the validity invariant in read_discriminant, to match the (for now) intended semantics.
> 
> RalfJung do you think we should change Miri before/in conjunction with this PR or are you simply saying we should do that at some point in the future?

Anyway, I am worried that this discussion distracted us from the actual purpose of this pull request, and delayed fixing the issues it was intended to fix, while being completely about preexisting state of the matter.

RalfJung:
> Another way to look at this would be to say that they are never valid.

That would be a peculiar special case. Many types do not have discriminants and asking for their discriminant is always fine. I am not a fan if special-casing zero-variant enums. The codegen backend goes even further and specializes uninhbaited types. But again it seems entirely ad-hoc that `(i32, ())` has an always-valid discriminant and `(i32, !)` has an always-invalid discriminant -- I don't like such special cases, they make everything more complicated and harder to reason about.

If we really want to treat zero-variant enums separately we need a `Variants::None` layout so we can properly reflect this in a type's layout. Currently, zero-variant enums are `Variants::Single` and we need a lot of ad-hoc special casing all over the place because of this.

> Anyway, I am worried that this discussion distracted us from the actual purpose of this pull request, and delayed fixing the issues it was intended to fix, while being completely about preexisting state of the matter.

Well, we learned that the preeexistng semantics are weird at best, and IMO we should change the codegen backend to no longer special-case uninhabited types.

tmiasko:
* Updated the pull request to use `discrminant_for_variant` unconditionally, which is now also valid for zero-variant enums thanks to #90910.

RalfJung:
I opened https://github.com/rust-lang/rust/issues/91095 for the MIR semantics issue. This already exists in master so it does not have to block this PR.

wesleywiser (approved):
It would be good to add a test that shows the effect of this change but I don't think it needs to block merging this.

PR #89774, opened 2021-10-11
inline next() on &mut Iterator impl

the8472:
In [#87431](https://github.com/rust-lang/rust/pull/87431/files#diff-79a6b417b85ecf4f1a4ef2235135fedf540199caf6e9e1d154ac6a413b40a757R132-R136)   I found that `(&mut range).fold` doesn't optimize well because the default impl for for `fold` on `&mut Iterator` doesn't inline `next`. In that particular case it was worked around by using `try_fold` which takes a `&mut self` instead of `self`.

Let's see if this can be fixed more broadly.


the8472:
This now addresses the FIXME from #87431 since `#[inline]` works just as well as the original workaround without disturbing perf.

m-ou-se:
Thanks!

@bors r+

Issue #89280, opened 2021-09-26
ICE: attempt to subtract with overflow, compiler/rustc_errors/src/lib.rs:348:47


jyn514:
### Code

Rebase bdb5a5ffe17 over ac8dd1b2f24dc62c962172b27433106b4e84dc62, then run `x test src/tools/clippy`. I think this is the `write_literal_2` test.

cc @rust-lang/clippy, but I think this is an issue with rustc_errors, not clippy itself.

### Meta

`rustc --version --verbose`:
```
note: Clippy version: clippy 0.1.57 (ac8dd1b2f24 2021-09-26)
```

### Error output

```
thread 'rustc' panicked at 'attempt to subtract with overflow', compiler/rustc_errors/src/lib.rs:348:47
```

<details><summary><strong>Backtrace</strong></summary>
<p>

```
thread 'rustc' panicked at 'attempt to subtract with overflow', compiler/rustc_errors/src/lib.rs:348:47
stack backtrace:
   0: rust_begin_unwind
             at /home/joshua/rustc3/library/std/src/panicking.rs:517:5
   1: core::panicking::panic_fmt
             at /home/joshua/rustc3/library/core/src/panicking.rs:100:14
   2: core::panicking::panic
             at /home/joshua/rustc3/library/core/src/panicking.rs:50:5
   3: rustc_errors::CodeSuggestion::splice_lines::{{closure}}
             at /home/joshua/rustc3/compiler/rustc_errors/src/lib.rs:348:47
   4: core::ops::function::impls::<impl core::ops::function::FnMut<A> for &mut F>::call_mut
             at /home/joshua/rustc3/library/core/src/ops/function.rs:269:13
   5: core::iter::traits::iterator::Iterator::find_map::check::{{closure}}
             at /home/joshua/rustc3/library/core/src/iter/traits/iterator.rs:2397:32
   6: core::iter::adapters::cloned::clone_try_fold::{{closure}}
             at /home/joshua/rustc3/library/core/src/iter/adapters/cloned.rs:28:21
   7: core::iter::adapters::filter::filter_try_fold::{{closure}}
             at /home/joshua/rustc3/library/core/src/iter/adapters/filter.rs:44:44
   8: core::iter::traits::iterator::Iterator::try_fold
             at /home/joshua/rustc3/library/core/src/iter/traits/iterator.rs:1983:21
   9: <core::iter::adapters::filter::Filter<I,P> as core::iter::traits::iterator::Iterator>::try_fold
[...]

PR #89776, opened 2021-10-11
CI: Enable overflow checks for test (non-dist) builds

hkratz:
They stay disabled for Apple builds though, which take the most time already due to running on slow hw.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -293,6 +293,7 @@ jobs:
               MACOSX_DEPLOYMENT_TARGET: 10.7
               NO_LLVM_ASSERTIONS: 1
               NO_DEBUG_ASSERTIONS: 1
+              NO_OVERFLOW_CHECKS: 1
               DIST_REQUIRE_ALL_TOOLS: 1
             os: macos-latest
           - name: dist-x86_64-apple-alt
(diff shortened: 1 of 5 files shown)


hkratz:
rustbot label -S-waiting-on-author  +S-blocked

Mark-Simulacrum:
Do we have a recent-ish measurement of how much this would slow down the compiler if enabled by default? I feel like there was a measurement a few months (years?) back...

Even if we don't enable the overflow checks within std, it seems reasonable to enable them for the compiler only, if we can afford it.

Mark-Simulacrum:
Thanks. Yeah, I recall those measurements now. Agreed we likely can't accept that hit just yet.

I'm fine with us merging this and we can evaluate the CI time impact if it becomes a problem, it seems reasonable to have our test builders run with more assertions where we can afford it.

oli-obk:
will also fail the tools builder due to #89280 which breaks clippy tests

hkratz:
#89280 has been fixed. I have tested this locally with `x.py test --stage 2` and `./src/ci/docker/run.sh x86_64-gnu-tools`.

@rustbot label -S-blocked +S-waiting-on-review

r? @Mark-Simulacrum

PR #89778, opened 2021-10-11
Add #[must_use] to as_type conversions

jkugelman:
Clippy missed these:

```rust
alloc::string::String   fn as_mut_str(&mut self) -> &mut str;
core::mem::NonNull<T>   unsafe fn as_uninit_mut<'a>(&mut self) -> &'a MaybeUninit<T>;
str                     unsafe fn as_bytes_mut(&mut self) -> &mut [u8];
str                     fn as_mut_ptr(&mut self) -> *mut u8;
```

Parent issue: #89692

r? @joshtriplett


jkugelman:
Actually jeez, I just grepped by hand and there are another 69 `as_xxx` methods Clippy missed. Oh well, that's for another time.

PR #89790, opened 2021-10-11
Avoid unnecessary check from query hot path

Mark-Simulacrum:
r? @ghost for perf check


Mark-Simulacrum:
Locally this check seems to be basically duplicated with the other non-inlined check for the event filter bitmask containing the right event, and so I suspect it's actually hurting us more than helping. But these things are sensitive, so let's see what perf says.

@bors try @rust-timer queue

PR #89793, opened 2021-10-11
Add `slice::{from_ptr_range, from_mut_ptr_range} `

ibraheemdev:
Adds `slice::{from_ptr_range, from_mut_ptr_range}` as counterparts to `slice::{as_ptr_range, as_mut_ptr_range}`.

### DIFF
--- a/library/core/src/slice/mod.rs
+++ b/library/core/src/slice/mod.rs
@@ -71,6 +71,9 @@ pub use raw::{from_raw_parts, from_raw_parts_mut};
 #[stable(feature = "from_ref", since = "1.28.0")]
 pub use raw::{from_mut, from_ref};
 
+#[unstable(feature = "slice_from_ptr_range", issue = "89792")]
+pub use raw::{from_mut_ptr_range, from_ptr_range};
+
 // This function is public only because there is no other way to unit test heapsort.
 #[unstable(feature = "sort_internals", reason = "internal to sort module", issue = "none")]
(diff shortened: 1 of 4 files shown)


ibraheemdev:
@m-ou-se can you review this when you have the time, or should it be assigned to someone else?

m-ou-se (approved):
Apologies.

This looks good to me.

There is just one issue that worries me: There have been (incomplete) plans for quite a while to replace `Range` by two types: One that is `Copy`, and one that is an iterator. Using `Range` in signatures like this makes that change a little bit harder than it already is. Of course, the existing (and stable) `.as_ptr_range()` already has this problem as it uses `Range` as its return type. Still, I'm worried about making a potential change to `Range` even harder by adding the functions in this PR.

That doesn't have to stand in the way of adding these as unstable. I have added this as an unresolved question to the tracking issue, so we can make sure to address it before stabilizing this.

If you can fix the merge conflict, this is ready to merge.

ibraheemdev:
Fixed @m-ou-se

m-ou-se:
Can you rebase/squash the commits? Because now there's a commit named "[update rust-analyzer submodule](https://github.com/rust-lang/rust/pull/89793/commits/7012b17d1326e47527e5482d4bbc969c737ed9b9)", which shouldn't be part of this PR.

ibraheemdev:
squashed @m-ou-se

PR #89794, opened 2021-10-11
Add #[must_use] to to_value conversions

jkugelman:
`NonNull<T>::cast` snuck in when I wasn't looking. What a scamp!

Parent issue: #89692

r? @joshtriplett


jkugelman:
The ones that take `self` are on `char` and `NonNull`, which are `Copy`.

PR #89809, opened 2021-10-12
Remap ssa RealPredicate to llvm RealPredicate

tmiasko:
to avoid relying on the discriminant of the former for FFI purposes


michaelwoerister:
Thanks, @tmiasko! 

@bors r+

PR #89815, opened 2021-10-12
Associated consts sidebar

### LINKED ISSUE #89354: Rustdoc does not list associated constants for types in the sidebar
Consider the following struct and trait:

```rust
pub trait Trait {
    /// Foo.
    fn foo() { }

    /// Bar.
    const BAR: usize = 0;
}

pub struct Struct { }

impl Struct {
    /// Foo.
    pub fn foo() { }

    /// Bar.
    pub const BAR: usize = 0;
}
```

The sidebar generated by rustdoc for the trait is informative, allowing you to quickly jump to the methods and associated constants:
image

Sadly, the associated constant is completely missing in the sidebar for the struct:
image

This is particularly painful for types with many members, such as [`f32`](https://doc.rust-lang.org/std/primitive.f32.html). It is not immediately obvious which associated constants it has. In my opinion rustdoc should also generate an 'associated constants' section in the sidebar for types, not just for traits.

GuillaumeGomez:
Fixes #89354.

A screenshot with `f32`:

Screenshot from 2021-10-12 15-07-57

### DIFF
--- a/src/librustdoc/html/render/mod.rs
+++ b/src/librustdoc/html/render/mod.rs
@@ -1837,6 +1837,20 @@ fn get_methods(
         .collect::<Vec<_>>()
 }
 
+fn get_associated_constants(i: &clean::Impl, used_links: &mut FxHashSet<String>) -> Vec<String> {
+    i.items
+        .iter()
+        .filter_map(|item| match item.name {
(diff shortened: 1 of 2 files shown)


notriddle (approved):
Fix that one last nit, and r=me.

PR #89816, opened 2021-10-12
Fix invalid rules in .gitignore

Canop:
`**node_modules` in a .gitignore is the same than
`*node_modules` or `*****node_modules`.

It matches every file whose name ends with `node_modules`,
including `not_node_modules`.

The intent here was obviously to have `**/node_modules`
which is the same than just `node_modules`.

Reference on git ignoring rules format: https://git-scm.com/docs/gitignore

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -69,8 +69,8 @@ __pycache__/
 *$py.class
 
 ## Node
-**node_modules
-**package-lock.json
+node_modules
+package-lock.json
 
 ## Rustdoc GUI tests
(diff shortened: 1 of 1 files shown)


Canop:
Arguably, `node_modules/` would be more specific than just `node_modules` but I'm not sure it's sane to even accept files with this name in a repository.

Mark-Simulacrum:
I think this is fine, though the slash would work as well.

Thanks!

@bors r+ rollup

PR #79570, opened 2020-11-30
rustc: Stabilize `-Zrun-dsymutil` as `-Csplit-debuginfo`

alexcrichton:
This commit adds a new stable codegen option to rustc,
`-Csplit-debuginfo`. The old `-Zrun-dsymutil` flag is deleted and now
subsumed by this stable flag. Additionally `-Zsplit-dwarf` is also
subsumed by this flag but still requires `-Zunstable-options` to
actually activate. The `-Csplit-debuginfo` flag takes one of
three values:

* `off` - This indicates that split-debuginfo from the final artifact is
  not desired. This is not supported on Windows and is the default on
  Unix platforms except macOS. On macOS this means that `dsymutil` is
  not executed.

* `packed` - This means that debuginfo is desired in one location
  separate from the main executable. This is the default on Windows
  (`*.pdb`) and macOS (`*.dSYM`). On other Unix platforms this subsumes
  `-Zsplit-dwarf=single` and produces a `*.dwp` file.

* `unpacked` - This means that debuginfo will be roughly equivalent to
  object files, meaning that it's throughout the build directory
  rather than in one location (often the fastest for local development).
  This is not the default on any platform and is not supported on Windows.

Each target can indicate its own default preference for how debuginfo is
handled. Almost all platforms default to `off` except for Windows and
macOS which default to `packed` for historical reasons.

Some equivalencies for previous unstable flags with the new flags are:

* `-Zrun-dsymutil=yes` -> `-Csplit-debuginfo=packed`
* `-Zrun-dsymutil=no` -> `-Csplit-debuginfo=unpacked`
* `-Zsplit-dwarf=single` -> `-Csplit-debuginfo=packed`
* `-Zsplit-dwarf=split` -> `-Csplit-debuginfo=unpacked`

Note that `-Csplit-debuginfo` still requires `-Zunstable-options` for
non-macOS platforms since split-dwarf support was *just* implemented in
rustc.

There's some more rationale listed on #79361, but the main gist of the
motivation for this commit is that `dsymutil` can take quite a long time
to execute in debug builds and provides little benefit. This means that
incremental compile times appear that much worse on macOS because the
compiler is constantly running `dsymutil` over every single binary it
produces during `cargo build` (even build scripts!). Ideally rustc would
switch to not running `dsymutil` by default, but that's a problem left
to get tackled another day.

Closes #79361

PR #77117, opened 2020-09-23
cg_llvm: split dwarf support

davidtwco:
cc #34651

This PR adds initial support for Split DWARF to rustc, based on the implementation in Clang.

##### Current Status
This PR currently has functioning split-dwarf, running rustc with `-Zsplit-dwarf=split` when compiling a binary will produce a `dwp` alongside the binary, which contains the linked dwarf objects.

```shell-session
$ rustc -Cdebuginfo=2 -Zsplit-dwarf=split -C save-temps ./foo.rs
$ ls foo*
foo
foo.belfx9afw9cmv8.rcgu.dwo
foo.belfx9afw9cmv8.rcgu.o
foo.foo.7rcbfp3g-cgu.0.rcgu.dwo
foo.foo.7rcbfp3g-cgu.0.rcgu.o
foo.foo.7rcbfp3g-cgu.1.rcgu.dwo
foo.foo.7rcbfp3g-cgu.1.rcgu.o
foo.foo.7rcbfp3g-cgu.2.rcgu.dwo
foo.foo.7rcbfp3g-cgu.2.rcgu.o
foo.foo.7rcbfp3g-cgu.3.rcgu.dwo
foo.foo.7rcbfp3g-cgu.3.rcgu.o
foo.foo.7rcbfp3g-cgu.4.rcgu.dwo
foo.foo.7rcbfp3g-cgu.4.rcgu.o
foo.foo.7rcbfp3g-cgu.5.rcgu.dwo
foo.foo.7rcbfp3g-cgu.5.rcgu.o
foo.foo.7rcbfp3g-cgu.6.rcgu.dwo
foo.foo.7rcbfp3g-cgu.6.rcgu.o
foo.foo.7rcbfp3g-cgu.7.rcgu.dwo
foo.foo.7rcbfp3g-cgu.7.rcgu.o
foo.dwp
foo.rs
$ readelf -wi foo.foo.7rcbfp3g-cgu.0.rcgu.o
# ...
  Compilation Unit @ offset 0x90:
   Length:        0x2c (32-bit)
   Version:       4
   Abbrev Offset: 0x5b
   Pointer Size:  8
 <0><9b>: Abbrev Number: 1 (DW_TAG_compile_unit)
    <9c>   DW_AT_stmt_list   : 0xe8
    <a0>   DW_AT_comp_dir    : (indirect string, offset: 0x13b): /home/david/Projects/rust/rust0
[...]

PR #89819, opened 2021-10-12
cg: split dwarf for crate dependencies

### LINKED ISSUE #81024: split dwarf doesn't work with crate dependencies
As reported in https://github.com/rust-lang/rust/pull/77117#issuecomment-752872709 (cc @philipc), Split DWARF currently fails when compiling a crate with a dependency. 

To reproduce this, compile  `backtrace-rs` with the following command:

```shell-session
$ RUSTFLAGS="-Z split-dwarf=split" cargo +nightly --verbose build
```

davidtwco:
Fixes #81024.

- In #79570, `-Z split-dwarf-kind={none,single,split}` was replaced by `-C split-debuginfo={off,packed,unpacked}`. `-C split-debuginfo`'s packed and unpacked aren't exact parallels to single and split, respectively.

  On Unix, `-C split-debuginfo=packed` will put debuginfo in object files and package debuginfo into a DWARF package file (`.dwp`) and `-C split-debuginfo=unpacked` will put debuginfo in dwarf object files and won't package it.

  In the initial implementation of Split DWARF, split mode wrote sections which did not require relocation into a DWARF object (`.dwo`) file which was ignored by the linker and then packaged those DWARF objects into DWARF packages (`.dwp`). In single mode, sections which did not require relocation were written into object files but ignored by the linker and were not packaged. However, both split and single modes could be packaged or not, the primary difference in behaviour was where the debuginfo sections that did not require link-time relocation were written (in a DWARF object or the object file).

  In the first commit of this PR, I re-introduce a `-Z split-dwarf-kind` flag, which can be used to pick between split and single modes when `-C split-debuginfo` is used to enable Split DWARF (either packed or unpacked).
- Split DWARF packaging requires all of the object files to exist, including those in dependencies. ~~Therefore, the second commit of this PR makes rustc keep all objects or dwarf objects for unpacked mode and if the crate is a dependency in packed mode (determined by heuristic: if no linking is taking place), then objects or dwarf objects are kept. Objects are kept if `-Z split-dwarf-kind` is `SplitDwarfKind::Single`, and dwarf objects if `SplitDwarfKind::Split`.~~

  ~~There are other approaches that could be taken to supporting packed Split DWARF with crate dependencies but this seemed like the least complicated and was contained to only rustc. Other potential approaches are described in https://github.com/rust-lang/rust/issues/81024#issuecomment-760478223, I'm happy to change the approach I've taken here if it isn't what we're looking for.~~ See https://github.com/rust-lang/rust/pull/89819#issuecomment-985671867 for the current approach.
- ~~There's still a dependency on `llvm-dwp` after this change, which [we probably want to move away from](https://rust-lang.zulipchat.com/#narrow/stream/131828-t-compiler/topic/llvm-dwp.20is.20not.20recommended) but that seems out-of-scope for this PR. Ideally, Split DWARF (in packed or unpacked modes) will be usable on nightly after this lands. If there aren't any bugs reported then it's possible we could allow Split DWARF to be used on stable after this change, it depends whether or not switching away from `llvm-dwp` later would break any guarantees, or whether we'd want to change how we handle this cross-crate case in future.~~ See https://github.com/rust-lang/rust/pull/89819#issuecomment-985671867.

r? @nagisa
cc @alexcrichton


alexcrichton:
I think I'm probably missing something, but could you reiterate why a new compiler option is needed? Dealing with the matrix of `-Csplit-debuginfo` and `-Zsplit-dwarf-kind` seems hairy and difficult to manage, and the discussion for `-Csplit-debuginfo` was in theory all about not requiring a separate option for split-dwarf on Unix.

davidtwco:
> Hm so I was under the impression that the 4 options for dwarf were:
> 
> * Nothing, no debug info
> * Debuginfo in `*.o` files themselves, then also linked into final executable
> * Debuginfo split out into `*.dwo` files and nothing else
> * Debuginfo "linked" into `*.dwp` from all the `*.dwo` files
> 
> That corresponds to `-Cdebuginfo=0`, `-Csplit-debuginfo={off,unpacked,packed}`. Does split-dwarf have more options we need to encompass though? (sorry I feel like I'm missing something obvious you've probably already explained but on rereading I keep getting confused...)

As I understand it, it would be:

- Nothing, no debuginfo.
- Debuginfo in object files, linked into the final executable.
  - `-g -Cdebuginfo=2 -Csplit-debuginfo=off` 
- Debuginfo in object files, but in such a way that the linker skips it, not in the final executable - **single mode**
  - `-g -Cdebuginfo=2 -Csplit-debuginfo=unpacked -Zsplit-dwarf-kind=single`
- Debuginfo in object files, but in such a way that the linker skips it, not in the final executable, debuginfo from object files linked into dwarf package - **single mode, with packaging**
  - `-g -Cdebuginfo=2 -Csplit-debuginfo=packed -Zsplit-dwarf-kind=single`
- Debuginfo in dwarf object files, not in the final executable - **split mode**
  - `-g -Cdebuginfo=2 -Csplit-debuginfo=unpacked` (uses the default of `-Zsplit-dwarf-kind=split`)
- Debuginfo in dwarf object files, not in the final executable, debuginfo from object files linked into dwarf package - **split mode, with packaging**
  - `-g -Cdebuginfo=2 -Csplit-debuginfo=packed` (uses the default of `-Zsplit-dwarf-kind=split`)

I've described this in an [earlier comment](https://github.com/rust-lang/rust/pull/77117#issuecomment-739836394) which phrases it differently, that might help. I don't know what the specific use-cases for single vs split mode are for split dwarf, I just noticed that these were the variants provided by C++ compilers and so implemented both in rustc.

davidtwco:
Noticed there's still a bug in this somewhere, trying to build rustc with packed split dwarf will fail with an error about a duplicate DWO ID, should work fine for smaller crates, haven't been able to produce a small reproduction but I didn't spend too much time on it.

compiler/rustc_codegen_ssa/src/back/link.rs
```diff
@@ -1252,17 +1246,44 @@ pub fn linker_and_flavor(sess: &Session) -> (PathBuf, LinkerFlavor) {
     bug!("Not enough information provided to determine how to invoke the linker");
 }
 
-/// Returns a boolean indicating whether we should preserve both object and dwarf object files on
-/// the filesystem for their debug information. This is often useful with split-dwarf like schemes.
-fn preserve_temps_for_their_debuginfo(sess: &Session) -> bool {
+/// Returns a `(bool, bool)` tuple of indicating whether we should preserve object and dwarf object
+/// files (respectively) on the filesystem for their debug information. This is often useful with
+/// split-dwarf like schemes.
+fn preserve_temps_for_their_debuginfo(sess: &Session) -> (bool, bool) {
```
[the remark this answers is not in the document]
davidtwco: Fixed with recent push using [davidtwco/thorin](https://github.com/davidtwco/thorin).

davidtwco:
I've pushed a big change to this PR, which takes an entirely different approach than before: 

Instead of keeping temporaries around longer so that dwarf objects from dependencies are available to `llvm-dwp` invocations in the binary crate, we now put dwarf objects into rlibs and use `thorin` rather than `llvm-dwp`. `thorin` is a DWARF packaging utility that I've been working on for exactly this purpose, written in Rust using gimli and object. We wanted to switch away from `llvm-dwp` eventually because the author said it wasn't recommended, so it seemed reasonable to implement our own DWARF packaging utility (binutils' `dwp` and `llvm-dwp` both support DWARF 5 now, so there's no difference in compatibility, just whether or not they can read an rlib and are written in Rust).

There's still some things to do here, but it's a good step forward:

- [ ] `thorin` doesn't have tests yet, I need to add some - it passes the important tests from `llvm-dwp`, the failing tests differ on diagnostics or the checks we perform (e.g. I don't check for duplicate `DWO_ID` like `llvm-dwp` does, I just skip the duplicate). I just haven't added tests to `thorin` itself yet.
- [ ] `thorin` will need to be published on crates.io, and presumably we'll want to move it to the rust-lang organization if it's a part of the compiler.
- [ ] I haven't checked the performance of `thorin` compared to the alternatives (performance is probably similar), but we might want to do that.

I'd appreciate any reviews or contributions to `thorin` if you can see anything that needs improved.

khuey:
>     * [ ]  `thorin` will need to be published on crates.io, and presumably we'll want to move it to the rust-lang organization if it's a part of the compiler.

gimli is "part of the compiler" and not in the rust-lang org.

davidtwco:
> > ```
> > * [ ]  `thorin` will need to be published on crates.io, and presumably we'll want to move it to the rust-lang organization if it's a part of the compiler.
> > ```
> 
> gimli is "part of the compiler" and not in the rust-lang org.

As far as I can tell, for compiler-team maintained crates, it's preferred that they live in `rust-lang`:

> #### Can existing out-of-tree crates from personal accounts or other organizations be transferred?
> Yes, this is encouraged. In order to do this, discuss this with the team and familiarize yourself with the GitHub documentation for repository transfers and then arrange to perform the transfer. Once a transfer is complete, a redirect will exist in the original account or organization and this will conflict with the names of any new forks of the transferred repository - an email to GitHub is required to resolve this.

([source](https://rust-lang.github.io/compiler-team/procedures/crates/#can-existing-out-of-tree-crates-from-personal-accounts-or-other-organizations-be-transferred))

bjorn3:
Putting it in the gimli-rs organization may make sense too. That is where object, gimli and addr2line live too.

compiler/rustc_codegen_ssa/src/back/link.rs
```diff
@@ -635,57 +653,33 @@ fn escape_stdout_stderr_string(s: &[u8]) -> String {
...
-
-            if !dwp_not_found {
-                err.note(&format!("{:?}", &cmd));
-            }
-
-            err.emit();
+            sess.struct_err("linking dwarf objects with thorin failed")
+                .note(&format!("{:?}", e))
+                .emit();
+            info!("linker err:\n{}", format!("{:?}", e));
```
bjorn3, line 682: I don't think there is a need for this given that there is already an error with the exact same information.

src/bootstrap/dist.rs
```diff
@@ -398,12 +398,12 @@ impl Step for Rustc {
...
 
-            let src_dir = builder.sysroot_libdir(compiler, host).parent().unwrap().join("bin");
-            let dst_dir = image.join("lib/rustlib").join(&*host.triple).join("bin");
-            t!(fs::create_dir_all(&dst_dir));
-
             // Copy over lld if it's there
             if builder.config.lld_enabled {
+                let src_dir = builder.sysroot_libdir(compiler, host).parent().unwrap().join("bin");
+                let dst_dir = image.join("lib/rustlib").join(&*host.triple).join("bin");
+                t!(fs::create_dir_all(&dst_dir));
```
bjorn3, line 405: Other things may end up in the bin dir too in the future, so maybe keep unconditionally creating it?

compiler/rustc_session/src/config.rs
```diff
@@ -231,6 +231,37 @@ pub enum DebugInfo {
...
+/// Some debug-information requires link-time relocation and some does not. LLVM can partition
+/// the debuginfo into sections depending on whether or not it requires link-time relocation. Split
+/// DWARF provides a mechanism which allows the linker to skip the sections which don't require
+/// link-time relocation - either by putting those sections in DWARF object files, or by keeping
+/// them in the object file in such a way that the linker will skip them.
+#[derive(Clone, Copy, Debug, PartialEq, Hash)]
+pub enum SplitDwarfKind {
+    /// Sections which do not require relocation are written into object file but ignored by the
+    /// linker.
+    Single,
```
bjorn3, line 247: What is the advantage of using this when using `-Csplit-debuginfo=unpacked/packed`?

davidtwco: > Looks like the motivation is primarily to retain compatability with tools that only understand object files ([source](https://reviews.llvm.org/D52296#1242164)):
> 
> > To summarise:
> > 
> > * It shares most of the benefits with the .dwo solution.
> > * Unlike .dwo, there is no need to split the file and it is friendlier to other tools (ccache, distcc, etc)
> > * But unlike .dwo a distributed build system has to copy larger .o files.

(https://github.com/rust-lang/rust/pull/89819#issuecomment-949515345)

I don't think there's any advantage to using this with `-Csplit-debuginfo=packed`, but it might be useful with `-Csplit-debuginfo=unpacked`.

bjorn3: For `-Csplit-debuginfo=unpacked` we don't tell the build system where to find the debuginfo, so I don't think ccache or distcc can be safely used with `-Csplit-debuginfo=unpacked` if they were to cache dylibs or executables. For rlibs the `.dwo` files would be in the `.rlib` the build system knows about anyway, so it doesn't matter. With `-Csplit-debuginfo=unpacked` the only way to distinguish between debuginfo files for different if by assuming a specific filename scheme (and are indistinguishable if multiple crates have the same name, even if they have different `-Cmetadata`). Also with `-Csplit-debuginfo=unpacked` there would still be only a single kind of file left over after compilation. The object files are deleted after linking and the debuginfo files are kept for consumption by a debugger or other tools. Another thing is that `-Csplit-debuginfo=unpacked` produces multiple files anyway. One file for each cgu.

compiler/rustc_codegen_ssa/src/back/link.rs
davidtwco: Fixed!

src/bootstrap/dist.rs
davidtwco: Fixed!

bjorn3:
If this is ready for review you can comment `@rustbot ready`. This will flip the label from S-waiting-on-author to S-waiting-on-review.

davidtwco:
> If this is ready for review you can comment `@rustbot ready`. This will flip the label from S-waiting-on-author to S-waiting-on-review.

Thanks, had forgot to check the labels.

compiler/rustc_codegen_ssa/src/back/link.rs
```diff
@@ -514,57 +525,87 @@ fn escape_stdout_stderr_string(s: &[u8]) -> String {
...
 }
 
-const LLVM_DWP_EXECUTABLE: &'static str = "rust-llvm-dwp";
-
-/// Invoke `llvm-dwp` (shipped alongside rustc) to link `dwo` files from Split DWARF into a `dwp`
-/// file.
-fn link_dwarf_object<'a>(sess: &'a Session, executable_out_filename: &Path) {
-    info!("preparing dwp to {}.dwp", executable_out_filename.to_str().unwrap());
-
+/// Invoke `thorin` (rust implementation of a dwarf packaging utility) to link DWARF objects into a
```
nagisa, line 528: Nit: “Invoke” evokes an interpretation of starting an external process. “Use” would probably be more accurate.

compiler/rustc_session/src/config.rs
nagisa: Since this flag is a `-Z`, I'm comfortable with landing the implementation as is, but we should definitely revisit the concern if we ever end up considering the stabilization of the behaviour this flag exposes.

nagisa (approved):

nagisa:
r=me once the conflicts are resolved.

compiler/rustc_codegen_ssa/src/back/link.rs
davidtwco: Oops, this is just a leftover from when we were invoking an external process. Fixed!

davidtwco:
> r=me once the conflicts are resolved.

Fixed the conflicts (this adds a tiny amount of additional logic to `link_dwarf_object`), and also consolidated some of the split dwarf testing while I was at it.

Before landing this, I think we [might need an FCP here](https://rust-lang.github.io/compiler-team/procedures/crates/) to approve the new dependency; we might want to move the repository; and make sure that the compiler team can publish new versions.

davidtwco:
> > r=me once the conflicts are resolved.
> 
> Fixed the conflicts (this adds a tiny amount of additional logic to `link_dwarf_object`), and also consolidated some of the split dwarf testing while I was at it.
> 
> Before landing this, I think we [might need an FCP here](https://rust-lang.github.io/compiler-team/procedures/crates/) to approve the new dependency; we might want to move the repository; and make sure that the compiler team can publish new versions.

`thorin` now lives at https://github.com/rust-lang/thorin and the compiler team are owners on crates.io.

lqd:
Looks like an apple runner issue ? let's @bors retry

PR #67330, opened 2019-12-15
Implement split_inclusive for slice and str

golddranks:
# Overview
* Implement `split_inclusive` for `slice` and `str` and `split_inclusive_mut` for `slice`
* `split_inclusive` is a substring/subslice splitting iterator that includes the matched part in the iterated substrings as a terminator.
* EDIT: The behaviour has now changed, as per @KodrAus 's input, to the same semantics with the `split_terminator` function. I updated the examples below.
* Two examples below:
```Rust
    let data = "\nMäry häd ä little lämb\nLittle lämb\n";
    let split: Vec<&str> = data.split_inclusive('\n').collect();
    assert_eq!(split, ["\n", "Märy häd ä little lämb\n", "Little lämb\n"]);
```

```Rust
    let uppercase_separated = "SheePSharKTurtlECaT";
    let mut first_char = true;
    let split: Vec<&str> = uppercase_separated.split_inclusive(|c: char| {
        let split = !first_char && c.is_uppercase();
        first_char = split;
        split
    }).collect();
    assert_eq!(split, ["SheeP", "SharK", "TurtlE", "CaT"]);
```

# Justification for the API
* I was surprised to find that stdlib currently only has splitting iterators that leave out the matched part. In my experience, wanting to leave a substring terminator as a part of the substring is a pretty common usecase.
* This API is strictly more expressive than the standard `split` API: it's easy to get the behaviour of `split` by mapping a subslicing operation that drops the terminator. On the other hand it's impossible to derive this behaviour from `split` without using hacky and brittle `unsafe` code. The normal way to achieve this functionality would be implementing the iterator yourself.
* Especially when dealing with mutable slices, the only way currently is to use `split_at_mut`. This API provides an ergonomic alternative that plays to the strengths of the iterating capabilities of Rust. (Using `split_at_mut` iteratively used to be a real pain before NLL, fortunately the situation is a bit better now.)

# Discussion items
* <s>Does it make sense to mimic `split_terminator` in that the final empty slice would be left off in case of the string/slice ending with a terminator? It might do, as this use case is naturally geared towards considering the matching part as a terminator instead of a separator.</s>
  * EDIT: The behaviour was changed to mimic `split_terminator`.
* Does it make sense to have `split_inclusive_mut` for `&mut str`?

PR #89825, opened 2021-10-12
Make split_inclusive() on an empty slice yield an empty output

### LINKED ISSUE #89716: split_inclusive() on empty slice yields a single empty slice (instead of no slices at all)
I tried this code:

```rust
    let empty_str = "";
    let expected : Vec<&str> = vec![];
    assert_eq!(empty_str.split_inclusive(|ch| ch == '\n').collect::<Vec<_>>(), expected);
    let empty_slice: &[u8] = &[];
    let expected : Vec<&[u8]> = vec![];
    assert_eq!(empty_slice.split_inclusive(|b| *b == b'\n').collect::<Vec<_>>(), expected);
```

I expected to see this happen: Both assertions pass, i.e. `split_inclusive()` yields no results on empty input.

Instead, this happened: The second assertion fails because `split_inclusive()` on an empty slice yields a single empty element instead of no elements.

### Meta
`rustc --version --verbose`:
```
cargo 1.57.0-nightly (0121d66aa 2021-09-22)
```

(Skipping backtrace from template as that doesn't seem relevant.)

martinvonz:
`[].split_inclusive()` currently yields a single, empty slice. That's
different from `"".split_inslusive()`, which yields no output at
all. I think that makes the slice version harder to use.

The case where I ran into this bug was when writing code for
generating a diff between two slices of bytes. I wanted to prefix
removed lines with "-" and a added lines with "+". Due to
`split_inclusive()`'s current behavior, that means that my code prints
just a "-" or "+" for empty files. I suspect most existing callers
have similar "bugs" (which would be fixed by this patch).

Closes #89716.

### DIFF
--- a/library/alloc/tests/slice.rs
+++ b/library/alloc/tests/slice.rs
@@ -863,7 +863,7 @@ fn test_splitator_inclusive() {
     assert_eq!(xs.split_inclusive(|_| true).collect::<Vec<&[i32]>>(), splits);
 
     let xs: &[i32] = &[];
-    let splits: &[&[i32]] = &[&[]];
+    let splits: &[&[i32]] = &[];
     assert_eq!(xs.split_inclusive(|x| *x == 5).collect::<Vec<&[i32]>>(), splits);
 }
 
@@ -883,7 +883,7 @@ fn test_splitator_inclusive_reverse() {
     assert_eq!(xs.split_inclusive(|_| true).rev().collect::<Vec<_>>(), splits);
 
     let xs: &[i32] = &[];
-    let splits: &[&[i32]] = &[&[]];
+    let splits: &[&[i32]] = &[];
     assert_eq!(xs.split_inclusive(|x| *x == 5).rev().collect::<Vec<_>>(), splits);
 }
 
@@ -903,7 +903,7 @@ fn test_splitator_mut_inclusive() {
     assert_eq!(xs.split_inclusive_mut(|_| true).collect::<Vec<_>>(), splits);
 
     let xs: &mut [i32] = &mut [];
-    let splits: &[&[i32]] = &[&[]];
+    let splits: &[&[i32]] = &[];
     assert_eq!(xs.split_inclusive_mut(|x| *x == 5).collect::<Vec<_>>(), splits);
 }
 
@@ -923,7 +923,7 @@ fn test_splitator_mut_inclusive_reverse() {
     assert_eq!(xs.split_inclusive_mut(|_| true).rev().collect::<Vec<_>>(), splits);
 
     let xs: &mut [i32] = &mut [];
-    let splits: &[&[i32]] = &[&[]];
+    let splits: &[&[i32]] = &[];
     assert_eq!(xs.split_inclusive_mut(|x| *x == 5).rev().collect::<Vec<_>>(), splits);
 }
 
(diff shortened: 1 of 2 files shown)


cuviper:
Hmm, slice's behavior is consistent with `split`, so it seems to me that `str` is the odd one.

That is, `str::split`, `[_]::split`, and `[_]::split_inclusive` all return an empty item, but `str::split_inclusive` does not. The difference might be related to `str::split_terminator`, which does not yield a final unterminated empty string.

martinvonz:
> The difference might be related to `str::split_terminator`, which does not yield a final unterminated empty string.

Yes, I think so. There was some discussion in #67330 about that, and they decided to match `str::split_terminator`.

I spent some time going through occurrences of `split_inclusive` on GitHub but it was hard to find examples where it would make a difference because most were already checking their inputs in unrelated ways (not just for emptiness). But see how I worked around it in my repo if you want an example: https://github.com/martinvonz/jj/commit/c0a26f76426e4aa457f5c84080893bc926f4e844

joshtriplett:
I think it makes sense to match the behavior of not emitting a final empty item when the thing to split ends with a terminator.

That's separate from the question of what splitting an empty string/slice should yield. There's a consistent interpretation either way that could justify either behavior. We should figure out which one makes more sense, and also make sure we can still make a change here without breakage.

joshtriplett:
FWIW, my first instinct would be for split on an empty string/slice to yield no items. I would prefer to go with whatever is most likely to be both unsurprising and minimize special-cases in callers. Based on the reports thus far, I think yielding no items would eliminate more special cases.

martinvonz:
> We should figure out which one makes more sense, and also make sure we can still make a change here without breakage.

Agreed. I think the worst outcome would be if the `str` and the slice versions end up behaving differently forever. I suspect you'll find more more failures from Crater if you instead change the `str` version to yield a single empty string.

joshtriplett:
I'd be in favor of making this change while we can, to not yield an empty slice.

What are the chances we can make the same change to `[]::split`?

m-ou-se:
An overview of the current behaviour:
```
"..".split_inclusive('.') = [".", "."]
 ".".split_inclusive('.') = ["."]
  "".split_inclusive('.') = []

[0, 0].split_inclusive(|_| true) = [[0], [0]]
   [0].split_inclusive(|_| true) = [[0]]
    [].split_inclusive(|_| true) = [[]]

"..".split('.') = ["", "", ""]
 ".".split('.') = ["", ""]
  "".split('.') = [""]

[0, 0].split(|_| true) = [[], [], []]
   [0].split(|_| true) = [[], []]
    [].split(|_| true) = [[]]
```

`[].split_inclusive(|_| true) = [[]]` seems to be the odd one out.

martinvonz:
> An overview of the current behaviour:

Thanks for sharing that.

> Not sure which of these would be the odd one out:
> 
> * `"".split_inclusive('.') = []`
> * `[].split_inclusive(|_| true) = [[]]`

I think it's clearly ``[].split_inclusive()` that doesn't fit the pattern in the examples you shared. In all the other cases, the number of elements in the outer list drop by 1.

m-ou-se:
@rfcbot merge

martinvonz:
@Amanieu, @BurntSushi, @yaahc: Can I get a review from either of you?

martinvonz:
@m-ou-se: Does this need anything else from a human or does it just take this long for automation to get to it?

m-ou-se:
Thanks for the ping!

@bors r+

PR #60058, opened 2019-04-17
Make const parameters enforce no variance constraints

varkor:
Fixes https://github.com/rust-lang/rust/issues/60047. Also includes some minor const refactoring for convenience.

PR #89829, opened 2021-10-12
Consider types appearing in const expressions to be invariant

### LINKED ISSUE #80977: Using associated constants for generics is not considered using a type generic
Filing in addition to #80976 because this seems to be a separate issue. Includes the same example below. Essentially, even though the associated constant `To::DIM` is being used in the definition of the struct, the compiler marks `To` as being unused since no types related to it are being used.

```rust
#![feature(const_generics, const_evaluatable_checked)]

trait Vector {
    type Scalar;
    const DIM: usize;
}

struct Transform<From, To>
where
    From: Vector,
    To: Vector<Scalar = From::Scalar>,
{
    columns: [[From::Scalar; From::DIM]; To::DIM],
}
```

([Playground](https://play.rust-lang.org/?version=nightly&mode=debug&edition=2018&gist=4dbd1bdce73b47ca4ae2abc65e6fd0d2))

voidc:
This is an approach to fix #80977.
Currently, a type parameter which is only used in a constant expression is considered bivariant and will trigger error E0392 *"parameter T is never used"*.
Here is a short example:

```rust
pub trait Foo {
    const N: usize;
}

struct Bar<T: Foo>([u8; T::N])
where [(); T::N]:;
```
([playgound](https://play.rust-lang.org/?version=nightly&mode=debug&edition=2015&gist=b51a272853f75925e72efc1597478aa5))

While it is possible to silence this error by adding a `PhantomData<T>` field, I think the better solution would be to make `T` invariant.
This would be analogous to the invariance constraints added for associated types.
However, I'm quite new to the compiler and unsure whether this is the right approach.

r? @varkor (since you authored #60058)

### DIFF
--- a/compiler/rustc_typeck/src/variance/constraints.rs
+++ b/compiler/rustc_typeck/src/variance/constraints.rs
@@ -223,8 +223,8 @@ impl<'a, 'tcx> ConstraintContext<'a, 'tcx> {
                     self.add_constraints_from_region(current, lt, variance_i)
                 }
                 GenericArgKind::Type(ty) => self.add_constraints_from_ty(current, ty, variance_i),
-                GenericArgKind::Const(_) => {
-                    // Consts impose no constraints.
+                GenericArgKind::Const(val) => {
+                    self.add_constraints_from_const(current, val, variance_i)
                 }
             }
         }
@@ -263,7 +263,8 @@ impl<'a, 'tcx> ConstraintContext<'a, 'tcx> {
                 self.add_constraints_from_mt(current, &ty::TypeAndMut { ty, mutbl }, variance);
             }
 
-            ty::Array(typ, _) => {
+            ty::Array(typ, len) => {
+                self.add_constraints_from_const(current, len, variance);
                 self.add_constraints_from_ty(current, typ, variance);
             }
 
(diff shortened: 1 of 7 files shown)


lcnr:
talked with @nikomatsakis about this and came to the following conclusions if i remember correctly:
- as long as the `coherence_leak_check` future compat lint is not an error having the substs of constants be bivariant is unsound, consider the following example
```rust
#![feature(generic_const_exprs)]
use std::marker::PhantomData;

trait SadBee {
    const ASSOC: usize;
}
// fn(&'static ())` is a supertype of `for<'a> fn(&'a ())` while
// we allow two different impls for these types, leading
// to different const eval results.
impl SadBee for for<'a> fn(&'a ()) {
    const ASSOC: usize = 0;
}
impl SadBee for fn(&'static ()) {
    const ASSOC: usize = 100;
}

struct Foo<T: SadBee>([u8; <T as SadBee>::ASSOC], PhantomData<T>) where [(); <T as SadBee>::ASSOC]:,;

fn covariant(v: &'static Foo<for<'a> fn(&'a ())>) -> &'static Foo<fn(&'static ())> {
    v
}

fn main() {
    let y = covariant(&Foo([], PhantomData));
    println!("{:?}", y.0);
}
```
- making the substs invariant is therefore probably the right choice, if a bit unfortunate
- I am going to open an issue to improve this in the future once this PR is merged and also add the above example as a test.

lcnr:
going to approve this PR even if it was previously a draft as it looks ready to me. Good job :+1: 

@bors r+

PR #89839, opened 2021-10-13
Add #[must_use] to mem/ptr functions

jkugelman:
There's a lot of low-level / unsafe stuff here. Are there legit use cases for ignoring any of these return values?

* No regressions in `./x.py test --stage 1 library/std src/tools/clippy`.
* One regression in `./x.py test --stage 1 src/test/ui`. Fixed.
* I am unable to run `./x.py doc` on my machine so I'll need to wait for the CI to verify doctests pass. I eyeballed all the adjacent tests and they all look okay.

Parent issue: #89692

r? @joshtriplett

### DIFF
--- a/src/test/ui/issues/issue-31299.rs
+++ b/src/test/ui/issues/issue-31299.rs
@@ -29,6 +29,7 @@ struct PtrBack<T: Front>(Vec<T::Back>);
 
 struct M(PtrBack<Vec<M>>);
 
+#[allow(unused_must_use)]
 fn main() {
     std::mem::size_of::<M>();
 }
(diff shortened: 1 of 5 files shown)


src/test/ui/issues/issue-31299.rs
```diff
@@ -30,5 +30,5 @@ struct PtrBack<T: Front>(Vec<T::Back>);
 struct M(PtrBack<Vec<M>>);
 
 fn main() {
-    std::mem::size_of::<M>();
+    let _ = std::mem::size_of::<M>();
```
joshtriplett, line 33: For this one, I think you should ignore the lint rather than adding `let _ =`, just in case it affects the test.

joshtriplett:
One minor nit; r=me with that fixed.

src/test/ui/issues/issue-31299.rs
jkugelman: You got it.

PR #89841, opened 2021-10-13
Implement let-else type annotations natively

### LINKED ISSUE #89688: Let-else source expression move behavior
The move behavior on non-matching of expressions by `feature(let_else)` seems currently inconsistent with behavior of, for example, `if let`.

An `if-let` construct does not move from the matched expression if the pattern doesn't match:

```rust
fn example_if_let(value: Option<String>) {
    if let Some(inner) = value {
        println!("inner: {}", inner);
    } else {
        println!("other: {:?}", value);
    }
}

fn main() {
    example_if_let(Some("foo".into()));
    example_if_let(None);
}
```
compiles and prints
```
inner: foo
other: None
```

However, `let-else` seems to unconditionally move from the matched expression:
```rust
#![feature(let_else)]

fn example_let_else(value: Option<String>) {
    let Some(inner) = value else {
        println!("other: {:?}", value);
        return;
    };
    println!("inner: {}", inner);
}

fn main() {
    example_let_else(Some("foo".into()));
    example_let_else(None);
}
```
fails to compile:
```
error[E0382]: borrow of moved value: `value`
 --> src/main.rs:5:33
  |
3 | fn example_let_else(value: Option<String>) {
  |                     ----- move occurs because `value` has type `Option<String>`, which does not implement the `Copy` trait
4 |     let Some(inner) = value else {
  |                       ----- value moved here
5 |         println!("other: {:?}", value);
  |                                 ^^^^^ value borrowed here after move
```

I'm wondering if this is intentional, as it would seem to make it impossible to use the original unmatched value in an panic message (or diverge otherwise with the value or any value based on it).

```
rustc 1.57.0-nightly (485ced56b 2021-10-07)
binary: rustc
commit-hash: 485ced56b8753ec86936903f2a8c95e9be8996a1
commit-date: 2021-10-07
release: 1.57.0-nightly
LLVM version: 13.0.0
```

Proposed Label: F-let-else

### LINKED ISSUE #89807: let-else: linting for `unused` bindings is not handled correctly
```rust
#![feature(let_else)]

fn example_let_else(value: Option<String>) {
    #[allow(unused)]
    let banana = 1;
    #[allow(unused)]
    let Some(chaenomeles) = value else { return };
}
```

here `banana` will not be linted against as an unused variable, but `chaenomeles`will be:

```
warning: unused variable: `chaenomeles`
 --> src/lib.rs:7:14
  |
7 |     let Some(chaenomeles) = value else { return };
  |              ^^^^^^^^^^^ help: if this is intentional, prefix it with an underscore: `_chaenomeles`
  |
```

[playground](https://play.rust-lang.org/?version=nightly&mode=debug&edition=2018&gist=2eeb93396c1e5c4f4968544ca10fae2a)

cc https://github.com/rust-lang/rust/issues/87335

### LINKED ISSUE #89960: Erroneous 'ref mut' in let-else binding causes ICE
### Code

```Rust
#![feature(let_else)]
fn main() {
    let Some(ref mut meow) = Some(()) else { return };
}
```

### Meta

`rustc --version --verbose`:
```
rustc 1.57.0-nightly (c1026539b 2021-10-15)
binary: rustc
commit-hash: c1026539bd22e9d070988deaa47b1360cbc76436
commit-date: 2021-10-15
release: 1.57.0-nightly
LLVM version: 13.0.0
```

### Error output

```
warning: unused variable: `meow`
 --> x.rs:3:22
  |
3 |     let Some(ref mut meow) = Some(()) else { return };
  |                      ^^^^ help: if this is intentional, prefix it with an underscore: `_meow`
  |
```

<p>

```
thread 'rustc' panicked at 'called `Option::unwrap()` on a `None` value', compiler/rustc_borrowck/src/diagnostics/mutability_errors.rs:53:62
stack backtrace:
   0: rust_begin_unwind
             at /rustc/c1026539bd22e9d070988deaa47b1360cbc76436/library/std/src/panicking.rs:517:5
   1: core::panicking::panic_fmt
             at /rustc/c1026539bd22e9d070988deaa47b1360cbc76436/library/core/src/panicking.rs:100:14
   2: core::panicking::panic
             at /rustc/c1026539bd22e9d070988deaa47b1360cbc76436/library/core/src/panicking.rs:50:5
   3: rustc_borrowck::diagnostics::mutability_errors::<impl rustc_borrowck::MirBorrowckCtxt>::report_mutability_error
   4: rustc_borrowck::MirBorrowckCtxt::access_place
   5: <rustc_borrowck::MirBorrowckCtxt as rustc_mir_dataflow::framework::visitor::ResultsVisitor>::visit_statement_before_primary_effect
   6: <rustc_mir_dataflow::framework::direction::Forward as rustc_mir_dataflow::framework::direction::Direction>::visit_results_in_block
   7: rustc_borrowck::do_mir_borrowck
   8: rustc_infer::infer::InferCtxtBuilder::enter
   9: core::ops::function::FnOnce::call_once
  10: rustc_query_system::query::plumbing::try_execute_query
  11: <rustc_query_impl::Queries as rustc_middle::ty::query::QueryEngine>::mir_borrowck
  12: rustc_data_structures::sync::par_for_each_in
  13: rustc_session::utils::<impl rustc_session::session::Session>::time
  14: rustc_interface::passes::analysis
  15: rustc_query_system::query::plumbing::try_execute_query
  16: <rustc_query_impl::Queries as rustc_middle::ty::query::QueryEngine>::analysis
  17: rustc_interface::passes::QueryContext::enter
  18: rustc_interface::queries::<impl rustc_interface::interface::Compiler>::enter
  19: rustc_span::with_source_map
  20: scoped_tls::ScopedKey<T>::set
```

</p>
</details>

cormacrelf:
Tracking issue: #87335

Fixes #89688, fixes #89807, edit: fixes  #89960 as well

As explained in https://github.com/rust-lang/rust/issues/89688#issuecomment-940405082, the previous desugaring moved the let-else scrutinee into a dummy variable, which meant if you wanted to refer to it again in the else block, it had moved.

This introduces a new hir type, ~~`hir::LetExpr`~~ `hir::Let`, which takes over all the fields of `hir::ExprKind::Let(...)` and adds an optional type annotation. The `hir::Let` is then treated like a `hir::Local` when type checking a function body, specifically:

* `GatherLocalsVisitor` overrides a new `Visitor::visit_let_expr` and does pretty much exactly what it does for `visit_local`, assigning a local type to the `hir::Let` ~~(they could be deduplicated but they are right next to each other, so at least we know they're the same)~~
* It reuses the code in `check_decl_local` to typecheck the `hir::Let`, simply returning 'bool' for the expression type after doing that.

* ~~`FnCtxt::check_expr_let` passes this local type in to `demand_scrutinee_type`, and then imitates check_decl_local's pattern checking~~
* ~~`demand_scrutinee_type` (the blindest change for me, please give this extra scrutiny) uses this local type instead of of creating a new one~~
    * ~~Just realised the `check_expr_with_needs` was passing NoExpectation further down, need to pass the type there too. And apparently this Expectation API already exists.~~

Some other misc notes:

* ~~Is the clippy code supposed to be autoformatted? I tried not to give huge diffs but maybe some rustfmt changes simply haven't hit it yet.~~
* in `rustc_ast_lowering/src/block.rs`, I noticed some existing `self.alias_attrs()` calls in `LoweringContext::lower_stmts` seem to be copying attributes from the lowered locals/etc to the statements. Is that right? I'm new at this, I don't know.


cormacrelf:
Ok, we're back in business I think. Let-else is now type-checked exactly as a let statement is checked; they use the same code path via a `Declaration` abstraction. Obviously this raises a fair question, why not just use `hir::Local` all the way? I thought of this, tried it, and it was painful.

- The `init` being optional on a Local means lots of `local.init.unwrap_or_else(|| span_bug!(local.span, "..."))`. It's nice to have a type that maps exactly to the construct, and this remains true for future use cases e.g. `Guard::IfLet`.
- The visitor implementations got knotty, wondering whether you need to `.visit_local(local)` or just use its component parts. It's less confusing this way, at least for me, it forces you to think about them differently and I think that is generally justified by the non-optional inititaliser.

I have added a bunch of tests, one adapted from @est31's rustc work (see the tracking issue) and a bunch adapted from rfc2005 match ergonomics to check that that all works, especially with type annotations attached.

cjgillot (commented):
Thanks for the PR @cormacrelf. I wonder if we would simplify the code by using directly a `hir::Local` in `ExprKind::Let` instead of introducing `hir::Let` (which has essentially the same fields).

EDIT: sorry, did not see your earlier comment.

compiler/rustc_typeck/src/check/gather_locals.rs
```diff
@@ -7,6 +7,31 @@ use rustc_middle::ty::Ty;
...
+/// A declaration is an abstraction of [hir::Local] and [hir::Let].
+///
+/// It must have a hir_id, as this is how we connect gather_locals to the check functions.
+pub(super) struct Declaration<'a> {
+    pub hir_id: hir::HirId,
+    pub pat: &'a hir::Pat<'a>,
+    pub ty: Option<&'a hir::Ty<'a>>,
+    pub span: Span,
+    pub init: Option<&'a hir::Expr<'a>>,
+}
```
nagisa, line 19: Addition of this further suggests to me that the HIR representation of `let-else` should have been modeled after a `hir::Local` and not `if-let`… Out of scope for this PR though.

compiler/rustc_typeck/src/check/fn_ctxt/checks.rs
```diff
@@ -890,19 +896,20 @@ impl<'a, 'tcx> FnCtxt<'a, 'tcx> {
         expr.span
     }
 
-    fn overwrite_local_ty_if_err(
+    pub(crate) fn overwrite_local_ty_if_err(
```
nagisa, line 899: This doesn't need to become `pub(crate)` does it?

nagisa (commented):
One more test case that would be nice is for `ref` bindings (i.e. references introduced to bindings manually rather than through the match ergonomics stuff). There is one test with `ref mut` but given how thorough the autoderef ones are, the `ref mut` one looks pretty underrepresented.

See the comments inline, but the overall feeling I have is that the compiler side of the changes are pretty good.

JohnCSimon:
@cormacrelf ping from triage
Can you please address the review comments and then set this PR to be ready for review?

src/test/ui/let-else/issue-89960.rs
```diff
@@ -1,7 +1,9 @@
+// This had an ICE, see issue #89960
+
 #![feature(let_else)]
+#![deny(unused_variables)]
```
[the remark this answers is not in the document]
cormacrelf: This is now covered by `let-else-ref-bindings-pass.rs`

src/test/ui/let-else/let-else-binding-explicit-mut-annotated.rs
```diff
@@ -0,0 +1,20 @@
...
+fn main() {
+    let Some(n): &mut Option<i32> = &&Some(5i32) else { return }; //~ ERROR mismatched types
+    *n += 1;
+    let _ = n;
+
+    let Some(n): &mut Option<i32> = &&mut Some(5i32) else { return }; //~ ERROR mismatched types
+    *n += 1;
+    let _ = n;
+
+    let Some(n): &mut Option<i32> = &mut &mut Some(5i32) else { return }; // OK
```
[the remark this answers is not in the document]
cormacrelf: I've moved the passing cases into `let-else-binding-explicit-mut-pass.rs`

src/test/ui/let-else/let-else-deref-coercion.rs
```diff
@@ -0,0 +1,90 @@
...
+    fn bar(&self) -> Option<u32> {
+        let Bar::Present(z): &Bar = self else {
+            return None;
+        };
+        return Some(*z);
+    }
+}
+impl Foo {
+    fn set_bar(&mut self, value: u32) {
+        let Bar::Present(z): &mut Bar = self else { // OK
```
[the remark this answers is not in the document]
cormacrelf: I've split this file into two similar ones, one `check-pass` = `let-else-deref-coercion-annotated.rs`

compiler/rustc_typeck/src/check/gather_locals.rs
cormacrelf: At least we could pretty easily add support for type annotations to `if-let`, i.e. this which you can't do today:

```rust
if let Some(x): &mut i32 = optional() { ... }
```

If anything, of all the name bindings in the language, `hir::Local` is the odd one out, having an optional init expression.

cormacrelf:
Alright, changes applied as requested. I've rebased this and fixed the merge conflicts with the recent clippy changes.

The only interesting thing from the rebase is that the "try `self.0`" suggestion no longer appears, and thinking about that briefly, you realise the suggestion was wrong (you would have to write `&mut self.0`, so someone's rightly disabled this.

compiler/rustc_typeck/src/check/gather_locals.rs
cormacrelf: I guess a more perfect refactoring is to extract the common parts (excluding the ty/option ty) into a `hir::Decl` or something. But similarly this is a big enough change already without doing to `hir::Local` what I've done to let-else.

Ultimately it's just not that big a deal for the codebase. The other thing that's apparent from `gather_locals.rs` is that we only needed common functionality in one place and the rest of a change like that would just be busywork.

compiler/rustc_typeck/src/check/expr.rs
```diff
@@ -119,7 +119,7 @@ impl<'a, 'tcx> FnCtxt<'a, 'tcx> {
         self.check_expr_with_expectation(expr, ExpectHasType(expected))
     }
 
-    fn check_expr_with_expectation_and_needs(
+    pub(super) fn check_expr_with_expectation_and_needs(
```
nagisa, line 122: Nit: Does this need to become a `pub(super)`?

nagisa:
Can you please also squash the history slightly so that it conveys meaningful information once it is merged into the mainline?

r=me after! Thanks for taking time to do all this!

cormacrelf:
@nagisa i think that's done, thanks for your help!

nagisa:
@bors r+ rollup=never

Thanks!

PR #89849, opened 2021-10-13
CI: Selecting the Xcode version no longer needed with the macos-11 runners.

hkratz:
It does nothing nowadays since `/Applications/Xcode_12.2.app` does not exist in the GH runner environment and
automatically using the latest version selected by the GH environment is better anyway.


Mark-Simulacrum:
If we're going to delete all uses, I think we should delete the script in the same PR. It will be easy to revert this commit if we need to, but otherwise it's pretty likely we'll forget and be confused later down the line.

ehuss:
Just a note for posterity: To be a little clearer, 12.2 still exists on the 10.15 image.  I was a little concerned when testing macOS 11 of removing this since it wasn't clear how the `-latest` transition would work since github said it would take 8 weeks to roll out fully.  However, the default on 10.15 is xcode 12.4, so I think it should be safe to remove regardless.  (12.2 was fixed since it was the minimum version needed for aarch64).

PR #89860, opened 2021-10-13
Remove trailing semicolon from macro call span

camsteffen:
Macro call site spans are now less surprising/more consistent since they no longer contain a semicolon after the macro call.

The downside is that we need to do a little guesswork to get the semicolon in diagnostics. But this should not be noticeable since it is rare for the semicolon to not immediately follow the macro call.

### DIFF
--- a/compiler/rustc_expand/src/expand.rs
+++ b/compiler/rustc_expand/src/expand.rs
@@ -1087,25 +1087,19 @@ impl<'a, 'b> InvocationCollector<'a, 'b> {
                 let MacCallStmt { mac, style, attrs, .. } = mac.into_inner();
                 Ok((style == MacStmtStyle::Semicolon, mac, attrs.into()))
             }
-            StmtKind::Item(ref item) if matches!(item.kind, ItemKind::MacCall(..)) => {
-                match stmt.kind {
-                    StmtKind::Item(item) => match item.into_inner() {
-                        ast::Item { kind: ItemKind::MacCall(mac), attrs, .. } => {
-                            Ok((mac.args.need_semicolon(), mac, attrs))
-                        }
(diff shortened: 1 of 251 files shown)


camsteffen:
> Any specific reasons to do this?

Honestly I was motivated from finding multiple instances in Clippy where we have to work around the Span surprisingly containing the semicolon. But I also think the output just seems more sensible. It also fixes some wrong "this warning originates in the macro..." messages.

> The change seems fine to me if all bang macro spans are changed, not only statements.
> This can be done by removing `span` from `InvocationKind::Bang` and taking it from `mac` when necessary.

I removed the `span` arg from `collect_span`, but I think `InvocationKind::Bang` still needs span because of [this](https://github.com/rust-lang/rust/blob/eeb16a2a892c2a29b1da3085e29f39efa3486e1c/compiler/rustc_expand/src/expand.rs#L621-L626).

Also note I added a special case for macros that expand to an expression. The expanded statement in this case will have the pre-expansion span with the semicolon.

petrochenkov:
>`InvocationKind::Bang` still needs span because of [this](https://github.com/rust-lang/rust/blob/eeb16a2a892c2a29b1da3085e29f39efa3486e1c/compiler/rustc_expand/src/expand.rs#L621-L626).

Nah, no span - no need to update it there.
r? @petrochenkov

flip1995 (commented):
Small NIT in the Clippy changes. Not sure if even worth fixing.

camsteffen:
Comments are addressed. I added a couple functions to `SourceMap` for retrieving the stmt/semicolon span.

Now that I am no longer modifying the statement span, the test output changes are more uniform, less surprising.

PR #89862, opened 2021-10-13
rewrite error handling for unresolved inference vars

lcnr:
Pretty much completely rewrites `fn emit_inference_failure_err`.

This new setup should hopefully be easier to extend and is already a lot better when looking for generic arguments.
Because this is a rewrite there are still some parts which are lacking, these are tracked in #94483 and will be fixed in later PRs.

r? @estebank @petrochenkov


JohnCSimon:
ping from triage: 
@lcnr what is the status of this PR?

lcnr:
it's blocked on me taking a full day to work on this, should happen at some point during febuary :laughing:

src/test/ui/issues/issue-16966.stderr
```diff
@@ -2,9 +2,13 @@ error[E0282]: type annotations needed
   --> $DIR/issue-16966.rs:2:5
    |
 LL |     panic!(std::default::Default::default());
-   |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ cannot infer type for type parameter `M` declared on the function `begin_panic`
+   |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ cannot infer type of the type parameter `M` declared on the function `begin_panic`
    |
    = note: this error originates in the macro `$crate::panic::panic_2015` (in Nightly builds, run with -Z macro-backtrace for more info)
+help: consider specifying the generic argument
+   |
+LL |         $crate::rt::begin_panic::<M>($msg)
```
lcnr, line 10: macros xx

src/test/ui/issues/issue-25368.stderr
```diff
@@ -1,11 +1,13 @@
...
 LL |     let (tx, rx) = channel();
-   |         -------- consider giving this pattern the explicit type `(Sender<Foo<T>>, std::sync::mpsc::Receiver<Foo<T>>)`, where the type parameter `T` is specified
-...
-LL |         tx.send(Foo{ foo: PhantomData });
-   |                 ^^^ cannot infer type for type parameter `T` declared on the struct `Foo`
+   |         ^^^^^^^^
+   |
+help: consider giving this pattern a type
+   |
+LL |     let (tx, rx): (Sender<Foo<T>>, std::sync::mpsc::Receiver<Foo<T>>) = channel();
```
lcnr, line 9: i dislike that, but can fix it in this PR.

src/test/ui/issues/issue-51116.stderr
```diff
@@ -1,9 +1,6 @@
 error[E0282]: type annotations needed
   --> $DIR/issue-51116.rs:6:13
    |
-LL |         for tile in row {
```
lcnr, line 4: still have to deal with iterators

src/test/ui/issues/issue-69455.stderr
```diff
@@ -1,22 +1,20 @@
...
-   |                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^
-   |                    |
-   |                    this method call resolves to `<Self as Test<Rhs>>::Output`
-   |                    cannot infer type for type parameter `T` declared on the associated function `new_display`
+   |                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^ cannot infer type of the type parameter `T` declared on the associated function `new_display`
    |
    = note: this error originates in the macro `$crate::format_args_nl` (in Nightly builds, run with -Z macro-backtrace for more info)
+help: consider specifying the generic argument
+   |
+LL |     println!("{}", 23u64.test(xs.iter().sum())::<T>);
```
lcnr, line 10: yikes, that span is wrong

src/test/ui/suggestions/suggest-closure-return-type-1.stderr
```diff
@@ -1,13 +1,13 @@
...
+  --> $DIR/suggest-closure-return-type-1.rs:2:9
    |
 LL |     let _v = || -> _ { [] };
-   |                        ^^ cannot infer type
+   |         ^^
    |
-help: give this closure an explicit return type without `_` placeholders
+help: consider giving `_v` a type
    |
-LL |     let _v = || -> [_; 0] { [] };
```
lcnr, line 9: here closure return type is a lot better

compiler/rustc_infer/src/infer/error_reporting/need_type_info.rs
```diff
@@ -409,72 +223,48 @@ impl<'a, 'tcx> InferCtxt<'a, 'tcx> {
...
-                        bug!("unexpect const: {:?}", ct);
+                    let _ = ct.print(printer);
+                    InferenceDiagnosticsData {
+                        name: s,
+                        span: Some(origin.span),
+                        kind: UnderspecifiedArgKind::Const { is_parameter: false },
+                        parent: None,
+                    }
+                } else {
+                    // FIXME: This code seems a bit wrong, idk.
```
estebank, line 255: Could you elaborate?

compiler/rustc_infer/src/infer/error_reporting/need_type_info.rs
```diff
@@ -363,6 +119,64 @@ impl UnderspecifiedArgKind {
...
+        // does not capture anything.
+        //
+        // We do have to hide the `extern "rust-call"` ABI in that case though,
+        // which is too much of a bother for now.
+        _ => ty.print(printer),
+    };
+    s
+}
+
+fn closure_as_fn_str<'tcx>(infcx: &InferCtxt<'_, 'tcx>, ty: Ty<'tcx>) -> String {
```
estebank, line 160: Can we try to call this from `ty_to_string` as well?

compiler/rustc_infer/src/infer/error_reporting/need_type_info.rs
```diff
@@ -363,6 +119,64 @@ impl UnderspecifiedArgKind {
     }
 }
 
+fn ty_to_string<'tcx>(infcx: &InferCtxt<'_, 'tcx>, ty: Ty<'tcx>) -> String {
+    let mut s = String::new();
+    let mut printer = ty::print::FmtPrinter::new(infcx.tcx, &mut s, Namespace::TypeNS);
+    let ty_getter = move |ty_vid| {
```
estebank, line 125: Given these, we could potentially also return the list of type params to use them as part of the message.

compiler/rustc_infer/src/infer/error_reporting/need_type_info.rs
```diff
@@ -493,410 +283,121 @@ impl<'a, 'tcx> InferCtxt<'a, 'tcx> {
...
-                //    |             this method call resolves to `std::option::Option<&T>`
-                //    |
-                //    = note: type must be known at this point
-                self.annotate_method_call(segment, e, &mut err);
+            InferSourceKind::GenericArg {
+                insert_span,
+                argument_index,
+                generics_def_id,
+                def_id: _,
+            } => {
```
estebank, line 339: This reminds me: we should collect inference errors in the same expression and emit a single error, instead of one per arg that couldn't be inferred.

compiler/rustc_infer/src/infer/error_reporting/need_type_info.rs
```diff
@@ -493,410 +283,121 @@ impl<'a, 'tcx> InferCtxt<'a, 'tcx> {
...
+                        "consider specifying the generic argument{}",
+                        pluralize!(generics.params.len()),
+                    ),
+                    format!(
+                        "::<{}>",
+                        generics
+                            .params
+                            .iter()
+                            .enumerate()
+                            .take_while(|&(i, p)| i <= argument_index || !p.has_default())
```
estebank, line 369: 👍

compiler/rustc_infer/src/infer/error_reporting/need_type_info.rs
```diff
@@ -913,121 +414,454 @@ impl<'a, 'tcx> InferCtxt<'a, 'tcx> {
...
+                _ => 1,
+            }
+        }
+
+        match source.kind {
+            InferSourceKind::LetBinding { ty, .. } => ty_cost(ty),
+            InferSourceKind::ClosureArg { ty, .. } => 5 + ty_cost(ty),
+            InferSourceKind::GenericArg { def_id, .. } => {
+                match tcx.def_kind(def_id) {
+                    DefKind::Variant | DefKind::Ctor(CtorOf::Variant, _) => 15, // `None::<u32>` and friends is ugly.
```
estebank, line 502:
```suggestion
                    DefKind::Variant | DefKind::Ctor(CtorOf::Variant, _) => 15, // `None::<u32>` and friends are ugly.
```

compiler/rustc_infer/src/infer/error_reporting/need_type_info.rs
```diff
@@ -913,121 +414,454 @@ impl<'a, 'tcx> InferCtxt<'a, 'tcx> {
...
+                        None => return Box::new(iter::empty()),
+                    };
+                    return Box::new(iter::once(InsertableGenericArgs {
+                        insert_span: segment.ident.span.shrink_to_hi(),
+                        inferred: &substs[substs.len() - generics.params.len()..],
+                        generics_def_id: def_id,
+                        def_id,
+                    }));
                 }
             }
```
estebank, line 628: I think that in these two branches we could recover the logic we previously had to point only at the last relevant method call in the chain instead of the whole expression (as well as just the struct literal name instead of the whole struct literal with its fields) if we make `FindInferSourceVisitor` also keep the span we'll use for the actual error.

src/test/ui/const-generics/defaults/doesnt_infer.stderr
```diff
@@ -1,10 +1,13 @@
...
 LL |     let foo = Foo::foo();
-   |         ---   ^^^^^^^^ cannot infer the value of const parameter `N`
-   |         |
-   |         consider giving `foo` the explicit type `Foo<N>`, where the const parameter `N` is specified
+   |         ^^^
+   |
+help: consider giving `foo` a type
+   |
+LL |     let foo: Foo<N> = Foo::foo();
+   |            ++++++++
```
estebank, line 10: I think we'll need two different pprinters here, one for the messages and one for the suggestions.

Here, we would want the message to talk about `Foo<{N: u32}>` and the suggestion about `Foo<{ value }>` (ideally `Foo<0i32>`, but whatever, not this PR).

src/test/ui/const-generics/generic_const_exprs/const_eval_resolve_canonical.stderr
```diff
@@ -2,13 +2,18 @@ error[E0282]: type annotations needed
   --> $DIR/const_eval_resolve_canonical.rs:26:9
    |
 LL |     let mut _q = Default::default();
-   |         ^^^^^^ consider giving `_q` a type
+   |         ^^^^^^
+   |
+help: consider giving `_q` an explicit type, where the placeholders `_` are specified
```
bjorn3, line 7: could you specialize the error message to use singular placeholder when there is only a single placeholder?

src/test/ui/const-generics/generic_arg_infer/issue-91614.stderr
```diff
@@ -1,17 +1,19 @@
...
    = note: cannot satisfy `_: MaskElement`
 note: required by a bound in `Mask::<T, LANES>::splat`
   --> $SRC_DIR/core/src/../../portable-simd/crates/core_simd/src/masks.rs:LL:COL
    |
 LL |     T: MaskElement,
    |        ^^^^^^^^^^^ required by this bound in `Mask::<T, LANES>::splat`
+help: consider giving `y` an explicit type, where the type for type parameter `T` is specified
+   |
+LL |     let y: Mask<_, LANES> = Mask::<_, _>::splat(false);
+   |          ++++++++++++++++
```
estebank, line 16: It's a shame that it is `_` and not `T` here, but that shouldn't block this PR from landing.

src/test/ui/const-generics/generic_const_exprs/const_eval_resolve_canonical.stderr
estebank: Yes, that'd be lovely, maybe something like "consider giving `_q` an explicit type" and the suggestion be `: Type`.

src/test/ui/const-generics/infer/cannot-infer-const-args.stderr
```diff
@@ -2,12 +2,12 @@ error[E0282]: type annotations needed
...
 LL |     foo();
-   |     ^^^ cannot infer the value of const parameter `X` declared on the function `foo`
+   |     ^^^ cannot infer the value of the const parameter `X` declared on the function `foo`
    |
-help: consider specifying the const argument
+help: consider specifying the generic argument
    |
 LL |     foo::<X>();
-   |     ~~~~~~~~
+   |        +++++
```
estebank, line 10: I'm realizing that this message should mention what type `X` is and/or point at the `foo` definition.

compiler/rustc_infer/src/infer/error_reporting/need_type_info.rs
estebank: Comment still applies :)

estebank (commented):
There are some outstanding comments from the previous review, and I added a few nitpicks as well, but I'd love to land this as quickly as possible. I'm ok if we create tickets for follow up work after rebasing and doing a pass over the things that can be quickly addressed now.

src/test/ui/const-generics/generic_const_exprs/const_eval_resolve_canonical.stderr
lcnr: for `let x: _` this now just uses "consider giving `_q` an explicit type". Added a FIXME to special-case types with only a placeholder as that isn't as trivial

compiler/rustc_infer/src/infer/error_reporting/need_type_info.rs
lcnr: added a fixme for that in the code for now

compiler/rustc_infer/src/infer/error_reporting/need_type_info.rs
lcnr: added this to #94483

compiler/rustc_infer/src/infer/error_reporting/need_type_info.rs
lcnr: do you have a test where this is an issue? in `ui/const-generics/infer/method-chain` this already looks about right to me

src/test/ui/const-generics/infer/cannot-infer-const-args.stderr
lcnr: added to #94483

compiler/rustc_infer/src/infer/error_reporting/need_type_info.rs
```diff
@@ -485,454 +293,157 @@ impl<'a, 'tcx> InferCtxt<'a, 'tcx> {
...
-                //    |             |      cannot infer type for `T`
-                //    |             this method call resolves to `std::option::Option<&T>`
-                //    |
-                //    = note: type must be known at this point
-                self.annotate_method_call(segment, e, &mut err);
+            // FIXME: Consider autoref here. We sometimes
+            // have to add `&` or `&mut` in front of `self_span` here.
+            InferSourceKind::FullyQualifiedMethodCall { self_span, successor, substs, def_id } => {
+                let printer = fmt_printer(self, Namespace::ValueNS);
+
```
estebank: We need to add this code https://github.com/rust-lang/rust/pull/89862/files#diff-885afd8bd7572095e2347dc7eabce51cde3f47e07841d705770d94e0fccf2479L735-L776 in some way here.

compiler/rustc_infer/src/infer/error_reporting/need_type_info.rs
```diff
@@ -485,454 +293,154 @@ impl<'a, 'tcx> InferCtxt<'a, 'tcx> {
...
-                // LL |     let x = buffer.last().unwrap().0.clone();
-                //    |             -------^^^^--
-                //    |             |      |
-                //    |             |      cannot infer type for `T`
-                //    |             this method call resolves to `std::option::Option<&T>`
-                //    |
-                //    = note: type must be known at this point
-                self.annotate_method_call(segment, e, &mut err);
+            // FIXME: Consider autoref here. We sometimes
+            // have to add `&` or `&mut` in front of `self_span` here.
```
estebank, line 407: You can use the same logic as https://github.com/rust-lang/rust/pull/96772/files

src/test/ui/traits/multidispatch-convert-ambig-dest.stderr
```diff
@@ -2,13 +2,18 @@ error[E0282]: type annotations needed
   --> $DIR/multidispatch-convert-ambig-dest.rs:26:5
    |
 LL |     test(22, std::default::Default::default());
-   |     ^^^^ cannot infer type for type parameter `U` declared on the function `test`
+   |     ^^^^ cannot infer type of the type parameter `U` declared on the function `test`
+   |
+help: consider specifying the generic arguments
+   |
+LL |     test::<T, U>(22, std::default::Default::default());
+   |         ++++++++
```
estebank, line 10: I'm noticing that we already know `U` in the main label, we could use that to replace the other args with `_` in the suggestion.

compiler/rustc_infer/src/infer/error_reporting/need_type_info.rs
```diff
@@ -392,14 +393,34 @@ impl<'a, 'tcx> InferCtxt<'a, 'tcx> {
                     Applicability::HasPlaceholders,
                 );
             }
-            // FIXME: Consider autoref here. We sometimes
-            // have to add `&` or `&mut` in front of `self_span` here.
-            InferSourceKind::FullyQualifiedMethodCall { self_span, successor, substs, def_id } => {
+            InferSourceKind::FullyQualifiedMethodCall { receiver, successor, substs, def_id } => {
+                let typeck_results = self.in_progress_typeck_results.unwrap();
+                let typeck_results = typeck_results.borrow();
```
estebank, line 398: I'd move all of these closer to the assignment of adjustment and make it an `if let` chain so that if `in_progress_typeck_results` happens to be `None`, we don't blow up. I don't know in what circumstance that would happen, but I've already caused ICEs that way.

lcnr: > I don't know in what circumstance that would happen, but I've already caused ICEs that way.

can't happen here, as we only add this infer source if there are typeck results. We don't have any typeck_results if the inference errors happens outside of a body, e.g. during coherence or whatever?

> I'd move all of these closer to the assignment of adjustment

do you want that regardless?

estebank: The reason I want this is because this might be entirely safe today, but might not be after a refactor.

src/test/ui/traits/issue-77982.stderr
```diff
@@ -76,14 +68,16 @@ LL | impl Foo<'static, u32> for () {}
    | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 LL | impl<'a> Foo<'a, i16> for () {}
    | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+help: consider giving this pattern a type, where the type for type parameter `T` is specified
+   |
+LL |     let _: Box<T> = ().foo();
+   |          ++++++++
```
estebank, line 74: Note to self: it seems like we could potentially produce a suggestion here that we weren't before.

src/test/ui/span/method-and-field-eager-resolution.stderr
```diff
@@ -1,22 +1,26 @@
...
-   |         ----- consider giving `x` a type
-LL |     x.0;
-   |     ^ cannot infer type
+   |         ^^^^^
    |
    = note: type must be known at this point
+help: consider giving `x` an explicit type
+   |
+LL |     let mut x: _ = Default::default();
+   |              +++
```
estebank, line 11: Note to self: rework the output so we special case `: _` with a different placeholder.

src/test/ui/span/issue-42234-unknown-receiver-type.rs
```diff
@@ -6,9 +6,8 @@
 // the fix of which this tests).
 
 fn shines_a_beacon_through_the_darkness() {
-    let x: Option<_> = None;
+    let x: Option<_> = None::<_>; //~ ERROR type annotations needed
```
estebank, line 9: Why was the `::<_>` needing to be added here? If that isn't present, does the suggestion point at the binding type?

src/test/ui/issues/issue-20261.stderr
```diff
@@ -1,8 +1,6 @@
-error[E0282]: type annotations needed for `(_,)`
+error[E0282]: type annotations needed
   --> $DIR/issue-20261.rs:4:11
    |
-LL |     for (ref i,) in [].iter() {
-   |                     --------- this method call resolves to `std::slice::Iter<'_, T>`
 LL |         i.clone();
    |           ^^^^^ cannot infer type
```
estebank, line 5: Note to self: regression, we should point at *something* to explain where `i` came from.

src/test/ui/span/issue-42234-unknown-receiver-type.rs
lcnr: that was back when we only added the `::<X>` suggestion if no explicit generic args were previously present. Getting this test to test the right thing is actually somewhat difficult now :sweat_smile:

estebank:
Outstanding work left to do tracked in https://github.com/rust-lang/rust/issues/94483

@bors r+

compiler/rustc_middle/src/hir/map/mod.rs
```diff
@@ -983,7 +983,11 @@ impl<'hir> Map<'hir> {
             Node::AnonConst(constant) => self.body(constant.body).value.span,
             Node::Expr(expr) => expr.span,
             Node::Stmt(stmt) => stmt.span,
-            Node::PathSegment(seg) => seg.ident.span,
+            Node::PathSegment(seg) => {
+                let ident_span = seg.ident.span;
+                ident_span
+                    .with_hi(seg.args.map_or_else(|| ident_span.hi(), |args| args.span_ext.hi()))
+            }
```
lcnr, line 990: I would guess that the only thing which may impact the amount of `process_obligations` would be this.

We now use a different `span` from `seg.ident` which may mean that we don't dedup obligations as readily as they now have different spans? Can try to take a deeper look at this in the near future

PR #89865, opened 2021-10-13
Allow static linking LLVM with ThinLTO

tmandry:
There's no reason not to allow this if the user wants it. It works, at least in a local build on linux host.

For our use case, we're happy to spend more time building the compiler if it creates a speedup every time we run it, and we've observed speedups like this with clang.

### DIFF
--- a/src/bootstrap/config.rs
+++ b/src/bootstrap/config.rs
@@ -824,15 +824,10 @@ impl Config {
                 };
             }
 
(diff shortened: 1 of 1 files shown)


Mark-Simulacrum:
Seems OK. I'll note that another reason we don't do this is that our default dist build includes a bunch of llvm tools, and statically linking everything creates a pretty big install size as a result.

r=me

PR #89869, opened 2021-10-14
Add documentation to more `From::from` implementations.

kpreid:
For users looking at documentation through IDE popups, this gives them relevant information rather than the generic trait documentation wording “Performs the conversion”. For users reading the documentation for a specific type for any reason, this informs them when the conversion may allocate or copy significant memory versus when it is always a move or cheap copy.

Notes on specific cases:
* The new documentation for `From<T> for T` explains that it is not a conversion at all.
* Also documented `impl<T, U> Into<U> for T where U: From<T>`, the other central blanket implementation of conversion.
* The new documentation for construction of maps and sets from arrays of keys mentions the handling of duplicates. Future work could be to do this for *all* code paths that convert an iterable to a map or set.
* I did not add documentation to conversions of a specific error type to a more general error type.
* I did not add documentation to unstable code.

This change was prepared by searching for the text "From<... for" and so may have missed some cases that for whatever reason did not match. I also looked for `Into` impls but did not find any worth documenting by the above criteria.

### DIFF
--- a/library/alloc/src/collections/btree/map.rs
+++ b/library/alloc/src/collections/btree/map.rs
@@ -2047,6 +2047,8 @@ where
 
 #[stable(feature = "std_collections_from_array", since = "1.56.0")]
 impl<K: Ord, V, const N: usize> From<[(K, V); N]> for BTreeMap<K, V> {
+    /// Converts a `[(K, V); N]` into a `BTreeMap<(K, V)>`.
+    ///
     /// ```
     /// use std::collections::BTreeMap;
     ///
--- a/library/alloc/src/collections/btree/set.rs
+++ b/library/alloc/src/collections/btree/set.rs
@@ -1092,6 +1092,8 @@ impl<T: Ord> FromIterator<T> for BTreeSet<T> {
 
 #[stable(feature = "std_collections_from_array", since = "1.56.0")]
 impl<T: Ord, const N: usize> From<[T; N]> for BTreeSet<T> {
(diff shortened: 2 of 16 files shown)


Mark-Simulacrum:
Hm. I I seem to recall a thread a while back discussing what we want out of the documentation comments on impls like this. I personally don't know that repeating the signature of the impl pretty much directly in the documentation is helpful to those reading it; if editors are able to show that documentation line it seems like they could be convinced to show the signature of the method (which I would guess is *more* useful, particularly for mostly trivial functions like this). Maybe we should have some policy work in this area so that we can better guide PRs like this -- I certainly have opinions, but it feels like individual PRs being approved/rejected based on the particular reviewer is not a great experience.

This PR also adds the new documented constraint for deduplicating keys for the Set/Map From impls which seem pretty obviously like what we'd want to do, but I'd like at least one member of T-libs-api to weigh in there.

r? @dtolnay

dtolnay (changes requested):
Could you please pull the `BTreeMap` and `BTreeSet` documentation about the behavior on duplicates out to a separate PR? I'll bring that up with T-libs-api. I agree with Mark it's likely this is the behavior we want to commit to but I don't think I can do that unilaterally.

For the rest, I'll bump this over to T-libs to work out whether the cost/benefit of maintaining this kind of documentation is the right tradeoff for them.

Thanks!

kpreid:
I have removed the wording about duplicate items.

yaahc:
Looks good, thank you again for the help!

@bors r+ rollup

PR #89876, opened 2021-10-14
Make most std::ops traits const on numeric types

AlexApps99:
This PR makes existing implementations of `std::ops` traits (`Add`, `Sub`, etc) [`impl const`](https://github.com/rust-lang/rust/issues/67792) where possible.
This affects:
- All numeric primitives (`u*`, `i*`, `f*`)
- `NonZero*`
- `Wrapping`

This is under the `rustc_const_unstable` feature `const_ops`.
I will write tests once I know what can and can't be kept for the final version of this PR.

Since this is my first PR to rustc (and hopefully one of many), please give me feedback on how to better handle the PR process wherever possible. Thanks

[Zulip discussion](https://rust-lang.zulipchat.com/#narrow/stream/219381-t-libs/topic/Const.20std.3A.3Aops.20traits.20PR)

### DIFF
--- a/library/core/src/internal_macros.rs
+++ b/library/core/src/internal_macros.rs
@@ -9,6 +9,7 @@ macro_rules! forward_ref_unop {
         forward_ref_unop!(impl const $imp, $method for $t,
                 #[stable(feature = "rust1", since = "1.0.0")]);
     };
+    // Equivalent to the non-const version, with the addition of `rustc_const_unstable`
     (impl const $imp:ident, $method:ident for $t:ty, #[$attr:meta]) => {
         #[$attr]
         #[rustc_const_unstable(feature = "const_ops", issue = "90080")]
@@ -45,6 +46,7 @@ macro_rules! forward_ref_binop {
         forward_ref_binop!(impl const $imp, $method for $t, $u,
                 #[stable(feature = "rust1", since = "1.0.0")]);
     };
+    // Equivalent to the non-const version, with the addition of `rustc_const_unstable`
     (impl const $imp:ident, $method:ident for $t:ty, $u:ty, #[$attr:meta]) => {
         #[$attr]
         #[rustc_const_unstable(feature = "const_ops", issue = "90080")]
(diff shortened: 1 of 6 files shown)


AlexApps99:
Looks like someone's PR beat me to it (#90009).
A bit disheartening, but only a small part of this PR.
Hopefully it can be reviewed soon.
In the mean time, I am writing a tracking issue.

oli-obk:
As discussed in the libs-api meeting today, @rust-lang/wg-const-eval may now approve *unstably* adding `const` to trait impls, including using `~const`.

oli-obk (changes requested):
The impl looks good! We can go ahead with all of these. Just needs those tests you talked about.

AlexApps99:
I haven't completed the tests yet (mostly for a lack of time), are you sure it's OK to be merged without them?

RalfJung:
> I haven't completed the tests yet (mostly for a lack of time), are you sure it's OK to be merged without them?

Could you open an issue to track adding those tests, maybe leaving some notes for what kinds of tests you think would be good to have?

AlexApps99:
I'll put it on the tracking issue, I'm busy with exams right now but I should be able to set aside some time to implement the tests in a week or two.

ChaiTRex:
This seems to work on `u8 >> i32`, but not on `&u8 >> i32`, even though there's an implementation for both.

fee1-dead:
> This seems to add `const` to `u8 >> i32`, but not to `&u8 >> i32`, even though there's an implementation for both.

Does [this](https://play.rust-lang.org/?version=nightly&mode=debug&edition=2021&gist=4a9743179ec6f21e4ee57429e269b451) work?

ChaiTRex:
> > This seems to add `const` to `u8 >> i32`, but not to `&u8 >> i32`, even though there's an implementation for both.
> 
> Does [this](https://play.rust-lang.org/?version=nightly&mode=debug&edition=2021&gist=4a9743179ec6f21e4ee57429e269b451) work?

Ahh, I see the problem. I was using `#![feature(const_ops)]` instead of `#![feature(const_trait_impl)]`. The latter works. Thanks.

PR #89881, opened 2021-10-14
Avoid generating empty closures for fieldless enum variants

Mark-Simulacrum:
For many enums, this avoids generating lots of tiny stubs that need to be codegen'd and then inlined and removed by LLVM. perf shows this to be a fairly small, but significant, win on rustc bootstrap time -- with minimal impact on runtime performance (which is at times even positive).

### DIFF
--- a/compiler/rustc_macros/src/serialize.rs
+++ b/compiler/rustc_macros/src/serialize.rs
@@ -247,13 +247,24 @@ fn encodable_body(
                     })
                     .collect();
 
-                let result = quote! { ::rustc_serialize::Encoder::emit_enum_variant(
-                    __encoder,
-                   #variant_name,
-                   #variant_idx,
-                   #field_idx,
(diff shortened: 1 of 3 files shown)


Mark-Simulacrum:
The regression is not readily explainable -- the new code should perform ~equivalently or better. I suspect LLVM is optimizing worse since the closure isn't forcing unique instantiations per call site (which means the emit_usize in particular is going to be with a runtime, not constant, value).

Latest commit tries to force that question with a const-generic argument, which should still be cheap(er) than the closure - particularly since it'll be shared amongst like-numbered enums, hopefully.

@bors try @rust-timer queue

Mark-Simulacrum:
r? rust-lang/compiler-contributors

Perf impact outside the self-bootstrap timing is likely to be not entirely noise, but is fairly small and IMO justified by the improvements in bootstrap time this patch gives.

davidtwco (approved):

PR #89886, opened 2021-10-14
Update the wasi-libc built with the wasm32-wasi target

alexcrichton:
This commit updates the wasi-libc that we include with the wasm32-wasi
target, which brings in various misc fixes such as musl updates and some
math tweaks.


kennytm:
(https://github.com/WebAssembly/wasi-libc/compare/58795582905e08fa7748846c1971b4ab911d1e16...ad5133410f66b93a2381db5b542aad5e0964db96)

PR #89887, opened 2021-10-14
Change `char` type in debuginfo to DW_ATE_UTF

arlosi:
Rust previously encoded the `char` type as DW_ATE_unsigned_char. The more appropriate encoding is `DW_ATE_UTF`.

Clang also uses the DW_ATE_UTF for `char32_t` in C++.

This fixes the display of the `char` type in the Windows debuggers. Without this change, the variable did not show in the locals window.
image

LLDB 13 is also able to display the char value, when before it failed with `need to add support for DW_TAG_base_type 'char' encoded with DW_ATE = 0x8, bit_size = 32`

r? @wesleywiser

### DIFF
--- a/compiler/rustc_codegen_llvm/src/debuginfo/metadata.rs
+++ b/compiler/rustc_codegen_llvm/src/debuginfo/metadata.rs
@@ -87,7 +87,7 @@ const DW_ATE_signed: c_uint = 0x05;
 #[allow(non_upper_case_globals)]
 const DW_ATE_unsigned: c_uint = 0x07;
 #[allow(non_upper_case_globals)]
-const DW_ATE_unsigned_char: c_uint = 0x08;
+const DW_ATE_UTF: c_uint = 0x10;
 
 pub const UNKNOWN_LINE_NUMBER: c_uint = 0;
 pub const UNKNOWN_COLUMN_NUMBER: c_uint = 0;
@@ -933,7 +933,7 @@ fn basic_type_metadata<'ll, 'tcx>(cx: &CodegenCx<'ll, 'tcx>, t: Ty<'tcx>) -> &'l
         ty::Never => ("!", DW_ATE_unsigned),
         ty::Tuple(elements) if elements.is_empty() => ("()", DW_ATE_unsigned),
         ty::Bool => ("bool", DW_ATE_boolean),
-        ty::Char => ("char", DW_ATE_unsigned_char),
+        ty::Char => ("char", DW_ATE_UTF),
         ty::Int(int_ty) if cpp_like_debuginfo => (int_ty.msvc_basic_name(), DW_ATE_signed),
         ty::Uint(uint_ty) if cpp_like_debuginfo => (uint_ty.msvc_basic_name(), DW_ATE_unsigned),
         ty::Float(float_ty) if cpp_like_debuginfo => (float_ty.msvc_basic_name(), DW_ATE_float),
--- a/src/test/debuginfo/basic-types.rs
+++ b/src/test/debuginfo/basic-types.rs
@@ -104,7 +104,8 @@
 // cdb-check:b                : false [Type: bool]
 // cdb-command:dx i
 // cdb-check:i                : -1 [Type: [...]]
-// The variable 'c' doesn't appear for some reason...
+// cdb-command:dx c
+// cdb-check:c                : 0x61 'a' [Type: char32_t]
 // cdb-command:dx i8
 // cdb-check:i8               : 68 [Type: char]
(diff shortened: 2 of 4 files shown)


arlosi:
@joshtriplett I agree, I think we should fix gdb and continue with this change. I'll update the tests.

**TLDR:** only GDB is regressed, and other debuggers are either the same or improved by this change.

* LLDB 10 is broken both with and without the change.
* LLDB 13 works correctly with the change, and does not work without it.
* WinDbg, VS, CDB work correctly with the change and do not work without it.
* GDB is regressed by this change such that only the numeric value is printed.


LLDB 13 (without change)
```
(lldb) p c
error: need to add support for DW_TAG_base_type 'char' encoded with DW_ATE = 0x8, bit_size = 32
error: Couldn't materialize: couldn't get the value of variable asdf: Unable to determine byte size.
```

LLDB 13 (with change):
```
(lldb) p c
(char32_t) $0 = U+0x00000061 U'a'
```

LLDB 10 (with change)
```
(lldb) p c
error: need to add support for DW_TAG_base_type 'char' encoded with DW_ATE = 0x10, bit_size = 32
error: Couldn't materialize: couldn't get the value of variable asdf: Unable to determine byte size.
```

LLDB 10 (without change)
```
(lldb) p c
error: need to add support for DW_TAG_base_type 'char' encoded with DW_ATE = 0x8, bit_size = 32
error: Couldn't materialize: couldn't get the value of variable asdf: Unable to determine byte size.
```

Tested GDB versions 9, 11 with same result
GDB (without change)
```
(gdb) p c
$1 = 97 'a'
```

GDB (with change)
```
(gdb) p c
$1 = 97
```

joshtriplett:
I'm not sure. cc @tromey perhaps?

tromey:
I sent a gdb patch: https://sourceware.org/pipermail/gdb-patches/2021-October/182960.html

arlosi:
@tromey thanks for getting this fixed in GDB!

@joshtriplett how long should we wait before merging this?

wesleywiser:
Nominating for compiler team discussion. This PR changes debug info for `char` which fixes LLDB and WinDbg and causes variables of this type to be correctly displayed in the debugger. However, this PR also causes GDB to regress slightly in that it no longer will display the character value, just the numeric value. @tromey has fixed this in gdb but we're unsure how long we should wait before merging this PR.

Mark-Simulacrum:
Has that patch been merged upstream? I think I would block on a merge at least, personally.

Can we have an interim situation where we change the type (as per this PR) only on Windows targets (which seem the primary audience for WinDbg, though not LLDB) and then enable it more broadly when the next GDB release occurs?

tromey:
> Has that patch been merged upstream? I think I would block on a merge at least, personally.

I'll be merging it soon.  It will appear in gdb 11.2 as well.  This should be coming reasonably soon; the last status report is here: https://sourceware.org/pipermail/gdb-patches/2021-November/183905.html

> Can we have an interim situation where we change the type (as per this PR) only on Windows targets (which seem the primary audience for WinDbg, though not LLDB) and then enable it more broadly when the next GDB release occurs?

I think it's better to simply ship it.  Distros that are concerned with this, and that don't want to wait for GDB 11.2, can just apply the upstream patch.

wesleywiser:
We discussed this in yesterday's compiler team meeting and the consensus was that we should wait for the gdb 11.2 release to occur before merging. It sounds like that should be relatively soon but we will re-evaluate if that proves incorrect. 

I'm going to mark this PR as blocked for now.

mati865:
Gdb 11.2 has been released. Shouldn't this wait for it to popularise a bit though?

tromey:
> Gdb 11.2 has been released. Shouldn't this wait for it to popularise a bit though?

What's the lead time on a rustc patch making it into a release?  It used to be multiple weeks, and if that's still true, then I personally think it's fine to land it now.

wesleywiser:
If we merged right now, this would go into 1.60 which ships on April 7th.

mati865:
FWIW Ubuntu 22.04 LTS with this version should be available by the end of April.

mati865:
Beta promotion has happened already which means distribuitions with working GDB version should be already available when this hits stable. So should be good to go from POV.

wesleywiser:
Thanks for the reminder @mati865! Since the compiler team has already decided the gdb 11.2 would be sufficient and we're now at the point where that will release will be available in various distros by the time this reaches stable, I think we should merge. 

Thanks to @arlosi for your patience and also to @tromey for contributing that gdb patch! It's very much appreciated. 

@bors r+

PR #89888, opened 2021-10-14
Make `llvm.download-ci-llvm="if-available"` work for tier 2 targets with host tools

hkratz:
`llvm.download-ci-llvm="if-available"` is used for most profiles configured via `x.py setup`. It allows downloading prebuilt LLVM tarballs from the CI artifacts for a configured list of platforms. Currently this list is restricted to tier 1 targets but it makes sense for all tier 2 targets with host tools.


Mark-Simulacrum:
@bors r+ rollup

Thanks!

Issue #78543, opened 2020-10-29
Bad diagnostic for Send bound on Future coming from async fn in foreign crate


estebank:
As documented in [_Debugging async generator errors_](https://meltware.com/2020/10/21/rust-async-nonsense.html), when the problematic generator is in a foreign crate, we don't give the [nice errors for things being kept through an `await` point](https://blog.rust-lang.org/inside-rust/2019/10/11/AsyncAwait-Not-Send-Error-Improvements.html). The reason for that is [this early return](https://github.com/rust-lang/rust/blob/708fc3b1a2c6f9e6c83cfdba3e5e90855b054cd8/compiler/rustc_trait_selection/src/traits/error_reporting/suggestions.rs#L1369-L1372). We need to figure out a way of performing the subsequent analysis on foreign crates.

You can find some of my analysis so far in [this thread](https://twitter.com/ekuber/status/1321933923463557120).

CC @tmandry @davidtwco

```
error[E0277]: `(dyn futures::Stream<Item = std::result::Result<std::vec::Vec<repr::Row>, comm::Error>> + std::marker::Send + std::marker::Unpin + 'static)` cannot be shared between threads safely
   --> src/materialized/src/mux.rs:138:100
    |
138 |       async fn handle_connection(&self, conn: SniffedStream<TcpStream>) -> Result<(), anyhow::Error> {
    |  ____________________________________________________________________________________________________^
139 | |         self.handle_connection(conn).await
140 | |     }
    | |_____^ `(dyn futures::Stream<Item = std::result::Result<std::vec::Vec<repr::Row>, comm::Error>> + std::marker::Send + std::marker::Unpin + 'static)` cannot be shared between threads safely
    |
    = help: the trait `std::marker::Sync` is not implemented for `(dyn futures::Stream<Item = std::result::Result<std::vec::Vec<repr::Row>, comm::Error>> + std::marker::Send + std::marker::Unpin + 'static)`
    = note: required because of the requirements on the impl of `std::marker::Sync` for `std::ptr::Unique<(dyn futures::Stream<Item = std::result::Result<std::vec::Vec<repr::Row>, comm::Error>> + std::marker::Send + std::marker::Unpin + 'static)>`
    = note: required because it appears within the type `std::boxed::Box<(dyn futures::Stream<Item = std::result::Result<std::vec::Vec<repr::Row>, comm::Error>> + std::marker::Send + std::marker::Unpin + 'static)>`
[...]

PR #89889, opened 2021-10-14
Use the "nice E0277 errors"[1] for `!Send` `impl Future` from foreign crate

estebank:
Partly address #78543 by making the error quieter.

We don't have access to the `typeck` tables from foreign crates, so we
used to completely skip the new code when checking foreign crates. Now,
we carry on and don't provide as nice output (we don't clarify *what* is
making the `Future: !Send`), but at least we no longer emit a sea of
derived obligations in the output.

[1]: https://blog.rust-lang.org/inside-rust/2019/10/11/AsyncAwait-Not-Send-Error-Improvements.html

r? @tmandry


tmandry (approved):
Curious to know if these obligation clauses were helping anyone, but I suspect not!

r=me with comments addressed

Issue #76168, opened 2020-08-31
ICE: cannot convert `RePlaceholder` to a region vid


ririsoft:
Hello,

I am getting an ICE while "playing" with [`unboxed_closure`](https://github.com/rust-lang/rust/issues/29625)

### Code

MCVE can be found here: https://github.com/rust-lang/rust/issues/76168#issuecomment-683910817

<details><summary>Original</summary>

```Rust
#![feature(unboxed_closures)]
use futures::executor::block_on;
use std::future::Future;

async fn walk<F>(mut filter: F)
where
    for<'a> F: FnMut<(&'a u32,)> + 'a,
    for<'a> <F as FnOnce<(&'a u32,)>>::Output: Future<Output = bool> + 'a,
{
    let res = filter(&5u32).await;
    println!("res: {}", res);
}

fn main() {
    block_on(async { walk(|&i| async move { true }).await });
}

```
</details>

### Meta

`rustc --version --verbose`: (from Playground)

```
rustc 1.48.0-nightly (d006f5734 2020-08-28) running on x86_64-unknown-linux-gnu
```

### Error output

```
error: internal compiler error: src/librustc_mir/borrow_check/universal_regions.rs:768:36: cannot convert `RePlaceholder(Placeholder { universe: U5, name: BrNamed(DefId(0:8 ~ playground[a09e]::walk[0]::'a[1]), 'a) })` to a region vid

thread 'rustc' panicked at 'Box<Any>', src/librustc_errors/lib.rs:918:9
```

<details><summary><strong>Backtrace</strong></summary>
<p>

```
thread 'rustc' panicked at 'Box<Any>', src/librustc_errors/lib.rs:918:9
stack backtrace:
   0: std::panicking::begin_panic
   1: rustc_errors::HandlerInner::bug
   2: rustc_errors::Handler::bug
   3: rustc_middle::util::bug::opt_span_bug_fmt::{{closure}}
   4: rustc_middle::ty::context::tls::with_opt::{{closure}}
   5: rustc_middle::ty::context::tls::with_opt
   6: rustc_middle::util::bug::opt_span_bug_fmt
   7: rustc_middle::util::bug::bug_fmt
   8: rustc_mir::borrow_check::universal_regions::UniversalRegionIndices::to_region_vid::{{closure}}
   9: rustc_mir::borrow_check::universal_regions::UniversalRegionIndices::to_region_vid
[...]

PR #89890, opened 2021-10-14
Side-step ICE on `RePlaceholder` in `eval_verify_bound`

estebank:
Address #76168.

I don't think this is the *correct* way of handling this, but it does get rid of a stable-to-stable regression ICE.

The previous output (before the regression) was:

```
error[E0310]: the parameter type `F` may not live long enough
  --> src/lib.rs:9:1
   |
9  | / {
10 | | }
   | |_^
   |
   = help: consider adding an explicit lifetime bound `F: 'static`...
```

while the output with this PR is

```
error: `<F as FnOnce<(&'a mut i32,)>>::Output` does not live long enough
  --> $DIR/issue-76168.rs:8:1
   |
LL | / {
LL | |     let mut i = 41;
LL | |     f(&mut i).await;
LL | | }
   | |_^
```

### DIFF
--- a/compiler/rustc_borrowck/src/region_infer/mod.rs
+++ b/compiler/rustc_borrowck/src/region_infer/mod.rs
@@ -1167,6 +1167,10 @@ impl<'tcx> RegionInferenceContext<'tcx> {
                 self.scc_values.elements_contained_in(lower_bound_scc).next().is_none()
             }
 
+            // FIXME: this is likely to be incorrect, but it side-steps stable ICE #76168,
+            // without any changes to what we accept (or reject for that matter).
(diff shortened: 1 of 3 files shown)


estebank:
@matthewjasper would you have the time to open a PR doing the change you propose? I'm a bit swamped at the moment. You can close this PR if you do so. I'll keep this one open for now just as a note to self, to remind me to look at this if there's no movement in the medium term.

matthewjasper:
Opened #91321

crlf0710:
Triage: Since #91321 has landed and #76168.is now closed, i'm closing this PR.

PR #86048, opened 2021-06-06
core: add unstable no_fp_fmt_parse to disable float formatting code

nbdd0121:
In some projects (e.g. kernel), floating point is forbidden. They can disable
hardware floating point support and use `+soft-float` to avoid fp instructions
from being generated, but as libcore contains the formatting code for `f32`
and `f64`, some fp intrinsics are depended. One could define stubs for these
intrinsics that just panic [1], but it means that if any formatting functions
are accidentally used, mistake can only be caught during the runtime rather
than during compile-time or link-time, and they consume a lot of space without
LTO.

This patch provides an unstable cfg `no_fp_fmt_parse` to disable these.
A panicking stub is still provided for the `Debug` implementation (unfortunately)
because there are some SIMD types that use `#[derive(Debug)]`.

[1]: https://lkml.org/lkml/2021/4/14/1028

PR #89891, opened 2021-10-14
`alloc`: add unstable cfg features `no_rc` and `no_sync`

ojeda:
In Rust for Linux we are using these to make `alloc` a bit more modular.

See https://github.com/rust-lang/rust/pull/86048 and https://github.com/rust-lang/rust/pull/84266 for similar requests.

Of course, the particular names are not important.

### DIFF
--- a/library/alloc/src/lib.rs
+++ b/library/alloc/src/lib.rs
@@ -69,6 +69,8 @@
     any(not(feature = "miri-test-libstd"), test, doctest),
     no_global_oom_handling,
     not(no_global_oom_handling),
+    not(no_rc),
+    not(no_sync),
     target_has_atomic = "ptr"
 ))]
 #![no_std]
@@ -223,16 +225,17 @@ mod boxed {
 }
 pub mod borrow;
 pub mod collections;
-#[cfg(not(no_global_oom_handling))]
+#[cfg(all(not(no_rc), not(no_sync), not(no_global_oom_handling)))]
 pub mod ffi;
 pub mod fmt;
+#[cfg(not(no_rc))]
 pub mod rc;
 pub mod slice;
 pub mod str;
 pub mod string;
-#[cfg(target_has_atomic = "ptr")]
+#[cfg(all(not(no_rc), not(no_sync), target_has_atomic = "ptr"))]
 pub mod sync;
-#[cfg(all(not(no_global_oom_handling), target_has_atomic = "ptr"))]
+#[cfg(all(not(no_global_oom_handling), not(no_rc), not(no_sync), target_has_atomic = "ptr"))]
 pub mod task;
 #[cfg(test)]
 mod tests;
--- a/src/bootstrap/lib.rs
+++ b/src/bootstrap/lib.rs
@@ -197,6 +197,8 @@ const EXTRA_CHECK_CFGS: &[(Option<Mode>, &'static str, Option<&[&'static str]>)]
     (Some(Mode::Std), "stdarch_intel_sde", None),
     (Some(Mode::Std), "no_fp_fmt_parse", None),
     (Some(Mode::Std), "no_global_oom_handling", None),
+    (Some(Mode::Std), "no_rc", None),
+    (Some(Mode::Std), "no_sync", None),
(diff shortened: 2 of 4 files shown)


kennytm:
> Do we want to keep all possible combinations working and maintained if we keep adding this sort of flags (in the extreme, for every module)?

The `no_rc` and `no_sync` cfg features as-is which any tests is like a Tier-3 promise. Reviewers can't possibly figure out if all these combinations are working, so those responsible for these features have to fix them up after the fact.

The [`no_global_oom_handling`](https://github.com/rust-lang/rust/tree/master/src/test/run-make-fulldeps/alloc-no-oom-handling) (#84266) and [`no_fp_fmt_parse`](https://github.com/rust-lang/rust/blob/master/src/test/run-make-fulldeps/core-no-fp-fmt-parse/Makefile) (#86048) cfg features have got their own run-make-fulldeps tests to ensure building with them still works, like a Tier-2 promise. You probably needs to add 2 run-make-fulldeps tests as well.

`#[modular_crate]` needs something RFC-ish and is out of scope here 🙂.

ojeda:
> The `no_rc` and `no_sync` cfg features as-is which any tests is like a Tier-3 promise. Reviewers can't possibly figure out if all these combinations are working, so those responsible for these features have to fix them up after the fact.

So my question was whether we want to promise that every single combination of `cfg` flags will work to begin with.

If yes, one option is to have flags override others automatically, but that may be surprising, especially since these are negative (e.g. someone explicitly requested to remove `rc`, but something else re-selected it).

Thus it may be reasonable to only allow a subset of combinations instead, with the constraints encoded somewhere. From what I see, the Cargo docs suggest using `compile_error!` in the source code, which is not great, but I guess we could do it.

> The [`no_global_oom_handling`](https://github.com/rust-lang/rust/tree/master/src/test/run-make-fulldeps/alloc-no-oom-handling) (#84266) and [`no_fp_fmt_parse`](https://github.com/rust-lang/rust/blob/master/src/test/run-make-fulldeps/core-no-fp-fmt-parse/Makefile) (#86048) cfg features have got their own run-make-fulldeps tests to ensure building with them still works, like a Tier-2 promise. You probably needs to add 2 run-make-fulldeps tests as well.

That is definitely a good idea, will do. However, those PRs do not have a test for every combination, unless I am missing something.

joshtriplett:
> So my question was whether we want to promise that every single combination of cfg flags will work to begin with.

Right now, I don't think we're prepared to promise that any of those flags work. :)

If we want to stabilize these, we'll need to make the specific stability promise clear and any dependencies explicit. Until then, I think it's fine if these only work if used as tested.

Dylan-DPC:
@ojeda if you can resolve the conflicts we can get this merged

JohnCSimon:
@ojeda 
Ping from triage: Can you please post the status of this PR? 
Thank you.

ojeda:
@Dylan-DPC @JohnCSimon Thanks for the ping and apologies for the delay.

Rebased, added missing gates to allow for independent compilation of each flag, added a `run-make-fulldeps` test for each and compile-tested.

bjorn3:
You will need to add no_rc and no_sync to the array at https://github.com/rust-lang/rust/blob/64137f0b15b752d0c734661dc713bcd140858320/src/bootstrap/lib.rs#L197

ojeda:
> You will need to add no_rc and no_sync to the array at

Thanks! I took the chance to look for other instances and noticed we should put them in the `doc(cfg_hide())` too, so I did that too.

ojeda:
I also noticed the other `run-make-fulldeps` test in `alloc` was a bit out of date, so I sent https://github.com/rust-lang/rust/pull/95219 to align it to the ones introduced here.

Dylan-DPC:
r? rust-lang/libs

ojeda:
Rebased, added `-Dwarnings` to the tests and changed `-include`s to `include` according to ba74a628adfa which got merged a couple days ago.

Mark-Simulacrum:
@bors r+

This looks good to me; I imagine it might end up needing rebasing if we uncover a few more places we need the cfg but for now seems like a good start.

Issue #85991, opened 2021-06-04
Suggest using `-> impl Trait` when returning type parameter `-> T where T: Trait`


estebank:
Some newcomers are confused by `impl Trait` in return position because they have first internalized `impl Trait` as an alias for type parameters in argument position. Because of this, some people try the following:

```rust
trait Trait {}
impl Trait for () {}

// this works
fn foo() -> impl Trait {
    ()
}

// this doesn't
fn bar<T: Trait>() -> T {
    ()
}
```

We don't currently point them in the right direction, like we do in other cases.

_Took inspiration from https://twitter.com/JaapRood/status/1400782159812628482_

jackh726:
Should have been closed by #89892

PR #89892, opened 2021-10-14
Suggest `impl Trait` return type when incorrectly using a generic return type

Noratrieb:
Address #85991

When there is a type mismatch error and the return type is generic, and that generic parameter is not used in the function parameters, suggest replacing that generic with the `impl Trait` syntax.

r? @estebank

### DIFF
--- a/compiler/rustc_typeck/src/check/fn_ctxt/suggestions.rs
+++ b/compiler/rustc_typeck/src/check/fn_ctxt/suggestions.rs
@@ -8,8 +8,12 @@ use rustc_errors::{Applicability, DiagnosticBuilder};
 use rustc_hir as hir;
 use rustc_hir::def::{CtorOf, DefKind};
 use rustc_hir::lang_items::LangItem;
-use rustc_hir::{Expr, ExprKind, ItemKind, Node, Path, QPath, Stmt, StmtKind, TyKind};
+use rustc_hir::{
+    Expr, ExprKind, GenericBound, ItemKind, Node, Path, QPath, Stmt, StmtKind, TyKind,
+    WherePredicate,
+};
 use rustc_infer::infer::{self, TyCtxtInferExt};
+
 use rustc_middle::lint::in_external_macro;
 use rustc_middle::ty::{self, Binder, Ty};
 use rustc_span::symbol::{kw, sym};
@@ -559,11 +563,121 @@ impl<'a, 'tcx> FnCtxt<'a, 'tcx> {
                 let ty = self.tcx.erase_late_bound_regions(ty);
                 if self.can_coerce(expected, ty) {
                     err.span_label(sp, format!("expected `{}` because of return type", expected));
+                    self.try_suggest_return_impl_trait(err, expected, ty, fn_id);
                     return true;
                 }
                 false
             }
         }
+    }
+
+    /// check whether the return type is a generic type with a trait bound
+    /// only suggest this if the generic param is not present in the arguments
+    /// if this is true, hint them towards changing the return type to `impl Trait`
+    /// ```
+    /// fn cant_name_it<T: Fn() -> u32>() -> T {
+    ///     || 3
+    /// }
+    /// ```
+    fn try_suggest_return_impl_trait(
+        &self,
+        err: &mut DiagnosticBuilder<'_>,
+        expected: Ty<'tcx>,
+        found: Ty<'tcx>,
+        fn_id: hir::HirId,
+    ) {
+        // Only apply the suggestion if:
+        //  - the return type is a generic parameter
+        //  - the generic param is not used as a fn param
+        //  - the generic param has at least one bound
+        //  - the generic param doesn't appear in any other bounds where it's not the Self type
+        // Suggest:
+        //  - Changing the return type to be `impl <all bounds>`
+
+        debug!("try_suggest_return_impl_trait, expected = {:?}, found = {:?}", expected, found);
+
+        let ty::Param(expected_ty_as_param) = expected.kind() else { return };
+
+        let fn_node = self.tcx.hir().find(fn_id);
+
+        let Some(hir::Node::Item(hir::Item {
+            kind:
+                hir::ItemKind::Fn(
+                    hir::FnSig { decl: hir::FnDecl { inputs: fn_parameters, output: fn_return, .. }, .. },
+                    hir::Generics { params, where_clause, .. },
+                    _body_id,
+                ),
+            ..
+        })) = fn_node else { return };
+
+        let Some(expected_generic_param) = params.get(expected_ty_as_param.index as usize) else { return };
+
+        // get all where BoundPredicates here, because they are used in to cases below
+        let where_predicates = where_clause
+            .predicates
+            .iter()
+            .filter_map(|p| match p {
+                WherePredicate::BoundPredicate(hir::WhereBoundPredicate {
+                    bounds,
+                    bounded_ty,
+                    ..
+                }) => {
+                    // FIXME: Maybe these calls to `ast_ty_to_ty` can be removed (and the ones below)
+                    let ty = <dyn AstConv<'_>>::ast_ty_to_ty(self, bounded_ty);
+                    Some((ty, bounds))
+                }
+                _ => None,
+            })
+            .map(|(ty, bounds)| match ty.kind() {
+                ty::Param(param_ty) if param_ty == expected_ty_as_param => Ok(Some(bounds)),
+                // check whether there is any predicate that contains our `T`, like `Option<T>: Send`
+                _ => match ty.contains(expected) {
+                    true => Err(()),
+                    false => Ok(None),
+                },
+            })
+            .collect::<Result<Vec<_>, _>>();
+
+        let Ok(where_predicates) =  where_predicates else { return };
+
+        // now get all predicates in the same types as the where bounds, so we can chain them
+        let predicates_from_where =
+            where_predicates.iter().flatten().map(|bounds| bounds.iter()).flatten();
+
+        // extract all bounds from the source code using their spans
+        let all_matching_bounds_strs = expected_generic_param
+            .bounds
+            .iter()
+            .chain(predicates_from_where)
+            .filter_map(|bound| match bound {
+                GenericBound::Trait(_, _) => {
+                    self.tcx.sess.source_map().span_to_snippet(bound.span()).ok()
+                }
+                _ => None,
+            })
+            .collect::<Vec<String>>();
+
+        if all_matching_bounds_strs.len() == 0 {
+            return;
+        }
+
+        let all_bounds_str = all_matching_bounds_strs.join(" + ");
+
+        let ty_param_used_in_fn_params = fn_parameters.iter().any(|param| {
+                let ty = <dyn AstConv<'_>>::ast_ty_to_ty(self, param);
+                matches!(ty.kind(), ty::Param(fn_param_ty_param) if expected_ty_as_param == fn_param_ty_param)
+            });
+
+        if ty_param_used_in_fn_params {
+            return;
+        }
+
+        err.span_suggestion(
+            fn_return.span(),
+            "consider using an impl return type",
+            format!("impl {}", all_bounds_str),
+            Applicability::MaybeIncorrect,
+        );
     }
 
     pub(in super::super) fn suggest_missing_break_or_return_expr(
--- a/src/test/ui/return/return-impl-trait.rs
+++ b/src/test/ui/return/return-impl-trait.rs
@@ -1,0 +1,30 @@
+// run-rustfix
+
+trait Trait {}
+impl Trait for () {}
+
+// this works
+fn foo() -> impl Trait {
+    ()
+}
+
+fn bar<T: Trait + std::marker::Sync>() -> T
+where
+    T: Send,
+{
+    () //~ ERROR mismatched types
+}
+
+fn other_bounds<T>() -> T
+where
+    T: Trait,
+    Vec<usize>: Clone,
+{
+    () //~ ERROR mismatched types
+}
+
+fn main() {
+    foo();
+    bar::<()>();
+    other_bounds::<()>();
+}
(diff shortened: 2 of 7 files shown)


Nilstrieb:
I'm done with the changes, you can review it again if you have time

compiler/rustc_typeck/src/check/fn_ctxt/suggestions.rs
```diff
@@ -534,13 +537,95 @@ impl<'a, 'tcx> FnCtxt<'a, 'tcx> {
...
+                    })
+                    .collect::<Vec<String>>();
+                let all_bounds_str = all_bounds.join(" + ");
+                if all_bounds.len() != 0 {
+                    if inputs.iter().any(|param| {
+                        let ty = <dyn AstConv<'_>>::ast_ty_to_ty(self, param);
+                        matches!(ty.kind(), ty::Param(fn_param_ty_param) if ty_param == fn_param_ty_param)
+                    }) {
+                        err.note(
+                            &format!("you could try using `impl {}` as the return type", all_bounds_str)
```
jyn514, line 614:
```suggestion
                            &format!("try using `impl {}` as the return type", all_bounds_str)
```

compiler/rustc_typeck/src/check/fn_ctxt/suggestions.rs
```diff
@@ -534,13 +537,95 @@ impl<'a, 'tcx> FnCtxt<'a, 'tcx> {
...
+        {
+            if let Some(generic_param) = params.get(ty_param.index as usize) {
+                let all_bounds = generic_param
+                    .bounds
+                    .iter()
+                    .chain(
+                        where_clause
+                            .predicates
+                            .iter()
+                            .flat_map(|p| match p {
```
jyn514, line 590: You can simplify this chain.
```suggestion
                            .filter_map(|p| match p {
```

compiler/rustc_typeck/src/check/fn_ctxt/suggestions.rs
```diff
@@ -534,13 +537,95 @@ impl<'a, 'tcx> FnCtxt<'a, 'tcx> {
...
+                            self.tcx.sess.source_map().span_to_snippet(bound.span()).ok()
+                        }
+                        _ => None,
+                    })
+                    .collect::<Vec<String>>();
+                let all_bounds_str = all_bounds.join(" + ");
+                if all_bounds.len() != 0 {
+                    if inputs.iter().any(|param| {
+                        let ty = <dyn AstConv<'_>>::ast_ty_to_ty(self, param);
+                        matches!(ty.kind(), ty::Param(fn_param_ty_param) if ty_param == fn_param_ty_param)
```
jyn514, line 611: What is this checking? Why is it different than the case below?

src/test/ui/return/return-impl-trait.rs
```diff
@@ -0,0 +1,24 @@
+trait Trait {}
+impl Trait for () {}
+
+// this works
+fn foo() -> impl Trait {
+    ()
+}
+
+fn bar<T: Trait + std::marker::Sync>() -> T
```
jyn514, line 9: Please also add the NOTEs to the test so it's easier to see what it's testing.

compiler/rustc_typeck/src/check/fn_ctxt/suggestions.rs
```diff
@@ -534,13 +537,95 @@ impl<'a, 'tcx> FnCtxt<'a, 'tcx> {
...
+                    }) {
+                        err.note(
+                            &format!("you could try using `impl {}` as the return type", all_bounds_str)
+                        );
+                    } else {
```
jyn514, line 616: Can you please add a test for this case? Right now it's not triggered because `bad_echo` has no trait bounds.

jyn514 (commented):
This is a neat idea :) it looks in pretty good shape, just a few suggestions.

compiler/rustc_typeck/src/check/fn_ctxt/suggestions.rs
Nilstrieb: I changed the naming and added a comment, I hope that will clear it up

compiler/rustc_typeck/src/check/fn_ctxt/suggestions.rs
Nilstrieb: `bad_echo_2` does, and triggers this note

src/test/ui/return/return-impl-trait.rs
Nilstrieb: The compiler outputs a lot of notes, some even have line numbers in them. Is there a way to not have to write all of them but just the most important ones?

jyn514: Not that I know of, unfortunately.

Nilstrieb:
I added the notes, also added `// run-rustfix` since it's a suggestion related change, and squashed all commits into one.

jyn514:
@Nilstrieb you need to add the `.fixed` file I think - you might be able to generate it with `--bless --run-rustfix`.

Nilstrieb:
It does have the fixed code, but the problem is that this fixed code is not compiling successfully because of `bad_echo`. How can I allow this?

jyn514:
@Nilstrieb I'd suggest making the rustfix test separate from the rest of the tests.

Nilstrieb:
I've squashed it again since this is hopefully the last change

compiler/rustc_typeck/src/check/fn_ctxt/suggestions.rs
```diff
@@ -534,13 +537,100 @@ impl<'a, 'tcx> FnCtxt<'a, 'tcx> {
...
+                if all_bounds.len() != 0 {
+                    let ty_param_used_in_fn_params = fn_parameters.iter().any(|param| {
+                        let ty = <dyn AstConv<'_>>::ast_ty_to_ty(self, param);
+                        matches!(ty.kind(), ty::Param(fn_param_ty_param) if expected_ty_as_param == fn_param_ty_param)
+                    });
+
+                    if ty_param_used_in_fn_params {
+                        // The returned type parameter is used as a function parameter, impl Trait
+                        // return type is probably not what the user wants, but leave a note
+                        err.note(&format!(
```
jackh726, line 617: I don't think this is a good suggestion. First, it's not clear if impl Trait should be used as the argument, output, or both. Second, doing so can alter the intent of the function signature.

src/test/ui/return/return-impl-trait.rs
```diff
@@ -0,0 +1,25 @@
...
+//~^ this type parameter
+//~| expected `T` because of return type
+where
```
jackh726, line 14: So...what's difficult about this, is that you can have clauses like `Option<T>: Send`. This would effectively constrain `T: Send`, but wouldn't be picked up if T in the output is converted to `impl Trait`.

I think the above should be added as a test first and foremost. But, we shouldn't suggest changing if `T` appears in any clauses where it isn't the self type.

compiler/rustc_typeck/src/check/fn_ctxt/suggestions.rs
```diff
@@ -534,13 +537,100 @@ impl<'a, 'tcx> FnCtxt<'a, 'tcx> {
...
+        {
+            if let Some(generic_param) = params.get(expected_ty_as_param.index as usize) {
+                let all_bounds = generic_param
+                    .bounds
+                    .iter()
+                    .chain(
+                        where_clause
+                            .predicates
+                            .iter()
+                            .filter_map(|p| match p {
```
jackh726, line 590: This doesn't properly handle multiple params right? Or any bounds with a different self type at all?

jackh726 (changes requested):

Nilstrieb:
Thank you for your review, I'll implement this next week!

compiler/rustc_typeck/src/check/fn_ctxt/suggestions.rs
```diff
@@ -559,13 +562,101 @@ impl<'a, 'tcx> FnCtxt<'a, 'tcx> {
...
+                                WherePredicate::BoundPredicate(hir::WhereBoundPredicate {
+                                    bounds,
+                                    bounded_ty,
+                                    ..
+                                }) => {
+                                    let ty = <dyn AstConv<'_>>::ast_ty_to_ty(self, bounded_ty);
```
jackh726, line 621: I'm wondering if we can avoid these ast_ty_to_ty calls? Since we're just checking for Param, not something more complicated.

Nilstrieb:
This was a long "next week", but I finally did it! @jackh726

compiler/rustc_typeck/src/check/fn_ctxt/suggestions.rs
```diff
@@ -559,13 +563,118 @@ impl<'a, 'tcx> FnCtxt<'a, 'tcx> {
...
+            }) => {
+                let ty = <dyn AstConv<'_>>::ast_ty_to_ty(self, bounded_ty);
+                Some((ty, bounds))
+            }
+            _ => None,
+        });
+
+        // check whether there is any predicate that contains our `T`, like `Option<T>: Send`
+        let any_predicate_contains_expected_ty =
+            where_predicates.clone().any(|(ty, _)| match ty.kind() {
```
jackh726, line 628: Does this really need to be cloned?

compiler/rustc_typeck/src/check/fn_ctxt/suggestions.rs
```diff
@@ -559,13 +563,118 @@ impl<'a, 'tcx> FnCtxt<'a, 'tcx> {
...
+                ty::Param(param_ty) if param_ty == expected_ty_as_param => false,
+                _ => ty.contains(expected, self.tcx),
+            });
+
+        if any_predicate_contains_expected_ty {
```
jackh726, line 633: This doesn't necessarily cover cases where `T` appears in the bounds, right? So like `Option<()>: Trait<T>`

compiler/rustc_typeck/src/check/fn_ctxt/suggestions.rs
```diff
@@ -559,13 +563,118 @@ impl<'a, 'tcx> FnCtxt<'a, 'tcx> {
...
+        }
+
+        // now get all predicates in the same types as the where bounds, so we can chain them
+        let predicates_from_where = where_predicates
+            .filter_map(|(ty, bounds)| match ty.kind() {
+                ty::Param(param_ty) if param_ty == expected_ty_as_param => Some(bounds),
+                _ => None,
+            })
+            .map(|bounds| bounds.iter())
+            .flatten();
```
jackh726, line 644: You could combine this with the above by collecting into a `Result`

compiler/rustc_typeck/src/check/fn_ctxt/suggestions.rs
```diff
@@ -559,13 +563,118 @@ impl<'a, 'tcx> FnCtxt<'a, 'tcx> {
...
+                }
+                _ => None,
+            })
+            .collect::<Vec<String>>();
+
+        if all_matching_bounds_strs.len() != 0 {
+            let all_bounds_str = all_matching_bounds_strs.join(" + ");
+
+            let ty_param_used_in_fn_params = fn_parameters.iter().any(|param| {
+                let ty = <dyn AstConv<'_>>::ast_ty_to_ty(self, param);
```
jackh726, line 663: Did you get a chance to look at the `ast_ty_to_ty` calls? I'd at least add a FIXME, even if you don't change this now.

compiler/rustc_typeck/src/check/fn_ctxt/suggestions.rs
```diff
@@ -559,13 +563,118 @@ impl<'a, 'tcx> FnCtxt<'a, 'tcx> {
...
+            .chain(predicates_from_where)
+            .filter_map(|bound| match bound {
+                GenericBound::Trait(_, _) => {
+                    self.tcx.sess.source_map().span_to_snippet(bound.span()).ok()
+                }
+                _ => None,
+            })
+            .collect::<Vec<String>>();
+
+        if all_matching_bounds_strs.len() != 0 {
```
jackh726, line 659: This could be an early return?

compiler/rustc_typeck/src/check/fn_ctxt/suggestions.rs
```diff
@@ -559,13 +563,118 @@ impl<'a, 'tcx> FnCtxt<'a, 'tcx> {
...
+            let all_bounds_str = all_matching_bounds_strs.join(" + ");
+
+            let ty_param_used_in_fn_params = fn_parameters.iter().any(|param| {
+                let ty = <dyn AstConv<'_>>::ast_ty_to_ty(self, param);
+                matches!(ty.kind(), ty::Param(fn_param_ty_param) if expected_ty_as_param == fn_param_ty_param)
+            });
+
+            if !ty_param_used_in_fn_params {
```
jackh726, line 667: As could this.

jackh726 (changes requested):
Sorry, more reviews.

Also, can you squash the commits?

Nilstrieb:
I'll squash it once it's done.

compiler/rustc_typeck/src/check/fn_ctxt/suggestions.rs
Nilstrieb: Hmm, I wasn't able to figure out how I'd do that. Do you have an idea?

jackh726: Maybe instead of making `contains` be on Ty, you can make it a visitor. Then just visit both the ty and bounds.

Nilstrieb: For the `contains` I was able to use the visitor on `Ty`. How should I visit the `PolyTraitRef` in the bounds? I found `hir::intravisit` but wasn't able to find a good way to use it.

jackh726: Sorry, thought I replied to this before.

There is a hir visitor: https://doc.rust-lang.org/stable/nightly-rustc/rustc_hir/intravisit/trait.Visitor.html

Alternatively, you *could* lower the trait ref.

I'm also fine it you just want to add this as a test case with a FIXME. This is just suggestion after all and doesn't have to he "perfect"

Nilstrieb: I'll just leave a fixme in a test 👍

Nilstrieb:
Finally did it, I wanted to have `TyS::contains` as a separate commit at first, but after a little git accident I decided to just squash them anyways

jackh726:
@bors delegate+

(just do r=jackh726)

PR #89895, opened 2021-10-14
Don't mark for loop iter expression as desugared

camsteffen:
We typically don't mark spans of lowered things as desugared. This helps Clippy rightly discern when code is (not) from expansion. This was discovered by @flip1995 at https://github.com/rust-lang/rust-clippy/pull/7789#issuecomment-939289501.


camsteffen:
I guess I improved incremental compilation by using more lowered spans but I'm not familiar with incremental.

davidtwco (approved):
LGTM

PR #89897, opened 2021-10-14
Add #[must_use] to remaining core functions

jkugelman:
I've run out of compelling reasons to group functions together across crates so I'm just going to go module-by-module. This is everything remaining from the `core` crate.

Ignored by clippy for reasons unknown:

```rust
core::alloc::Layout   unsafe fn for_value_raw<T: ?Sized>(t: *const T) -> Self;
core::any             const fn type_name_of_val<T: ?Sized>(_val: &T) -> &'static str;
```

Ignored by clippy because of `mut`:

```rust
str   fn split_at_mut(&mut self, mid: usize) -> (&mut str, &mut str);
```

<del>
Ignored by clippy presumably because a caller might want `f` called for side effects. That seems like a bad usage of `map` to me.

```rust
core::cell::Ref<'b, T>   fn map<U: ?Sized, F>(orig: Ref<'b, T>, f: F) -> Ref<'b, T>;
core::cell::Ref<'b, T>   fn map_split<U: ?Sized, V: ?Sized, F>(orig: Ref<'b, T>, f: F) -> (Ref<'b, U>, Ref<'b, V>);
```
</del>

Parent issue: #89692

r? @joshtriplett

### DIFF
--- a/library/core/src/cell.rs
+++ b/library/core/src/cell.rs
@@ -1333,6 +1333,7 @@ impl<'b, T: ?Sized> Ref<'b, T> {
     /// with the widespread use of `r.borrow().clone()` to clone the contents of
     /// a `RefCell`.
     #[stable(feature = "cell_extras", since = "1.15.0")]
+    #[must_use]
     #[inline]
     pub fn clone(orig: &Ref<'b, T>) -> Ref<'b, T> {
         Ref { value: orig.value, borrow: orig.borrow.clone() }
--- a/library/alloc/tests/str.rs
+++ b/library/alloc/tests/str.rs
@@ -1031,7 +1031,7 @@ fn test_split_at_mut() {
 #[should_panic]
 fn test_split_at_boundscheck() {
     let s = "ศไทย中华Việt Nam";
-    s.split_at(1);
+    let _ = s.split_at(1);
 }
 
 #[test]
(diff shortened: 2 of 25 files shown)


library/core/src/cell.rs
```diff
@@ -1357,6 +1358,7 @@ impl<'b, T: ?Sized> Ref<'b, T> {
     /// assert_eq!(*b2, 5)
     /// ```
     #[stable(feature = "cell_map", since = "1.8.0")]
+    #[must_use = "should not be called for side effects"]
```
joshtriplett, line 1361: While `Iterator::map` doesn't do anything unless you use the resulting iterator, functions like `Cell::map` do actually call their function, so it *is* legitimate (even if undesirable) to use them for their side effects. I think you should omit these and similar functions that call a closure, for which it seems plausible to use a closure with side effects.

library/core/src/cell.rs
```diff
@@ -1420,6 +1422,7 @@ impl<'b, T: ?Sized> Ref<'b, T> {
     /// assert_eq!(*end, [3, 4]);
     /// ```
     #[stable(feature = "refcell_map_split", since = "1.35.0")]
+    #[must_use = "should not be called for side effects"]
```
joshtriplett, line 1425: As above.

joshtriplett:
I left a comment on two functions that I think should not have `must_use`. r=me with those two dropped.

library/core/src/cell.rs
jkugelman: Removed.

library/core/src/cell.rs
jkugelman: Removed.

jkugelman:
@joshtriplett What does `r=me` mean, by the way?

joshtriplett:
@jkugelman If someone says `r=me with xyz fixed`, that means "anyone with permission who comes along later is welcome to `bors r=thepersonwhosaidthat` if they see that xyz has been fixed; you don't need to wait for me to come back and re-review".

PR #89900, opened 2021-10-15
Make ptr range iterable

dtolnay:
This PR adds [`std::iter::Step`](https://doc.rust-lang.org/1.55.0/std/iter/trait.Step.html) impls for `*const T` and `*mut T`, enabling iteration over a range of pointers, such as produced by [`<[T]>::as_ptr_range`](https://doc.rust-lang.org/1.55.0/std/primitive.slice.html#method.as_ptr_range) and [`<[T]>::as_mut_ptr_range`](https://doc.rust-lang.org/1.55.0/std/primitive.slice.html#method.as_mut_ptr_range).

It is common for systems interacting with C++ to need to use pointer ranges for iteration, as `begin`/`end` pointer pairs are a design pattern in the C++ standard library.

```rust
// Designed to be called from C++ or C.
#[no_mangle]
unsafe extern "C" fn demo(start: *const u16, end: *const u16) {
    for ptr in start..end {
        println!("{}", *ptr);
    }
}

fn main() {
    let slice = &[1u16, 2, 3];
    let range = slice.as_ptr_range();
    unsafe { demo(range.start, range.end); }
}
```

### DIFF
--- a/library/core/src/iter/range.rs
+++ b/library/core/src/iter/range.rs
@@ -1,4 +1,5 @@
 use crate::char;
+use crate::cmp::Ordering;
 use crate::convert::TryFrom;
 use crate::mem;
 use crate::ops::{self, Try};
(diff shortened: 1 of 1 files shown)


dtolnay (commented):
> ``` `Step` invariants not upheld```

It turns out the desired semantics for pointer ranges is incompatible with the current requirements of `Step`, so I've opened #91000 with a different approach. Closing this PR in favor of that one.

PR #89902, opened 2021-10-15
Restrict the aarch64 outline atomics test to Linux

hkratz:
The test was introduced in #83655, which enables the `outline-atomics` feature for aarch64-unknown-linux-* but not for any other aarch64 targets. The test did not check for Linux causing test failures on aarch64-apple-darwin.

r? @workingjubilee


workingjubilee:
🙃 I should have thought of that.
Thank you!
@bors r+ rollup=always

Issue #88620, opened 2021-09-03
Move format-version constant to rustdoc-json-types


jyn514:
Right now, the format-version is in rustdoc and not in rustdoc-json-types, which has two downsides:
1. It's harder to make sure they're kept in sync
2. Downstream users can't know the version ahead of time, they're forced to hard-code it in their own code as well.

We should move it to rustdoc-json-types instead.

_Originally posted by @jyn514 in https://github.com/rust-lang/rust/issues/88604#issuecomment-912613463_

jyn514:
Hi @hi-rustin, are you still planning to work on this?

jyn514:
@yuvaldolev go for it :)

PR #89906, opened 2021-10-15
Moved format-version constant to rustdoc-json-types

yuvaldolev:
Addresses #88620

Moved format-version constant from rustdoc to rustdoc-json-types.


CraftSpider:
I like this change, keeping the version inline with the definitions

@bors r+

PR #89907, opened 2021-10-15
Remove FIXME since there is nothing to be fixed

### LINKED ISSUE #88593: rustdoc reports error twice for ambiguous (inherent) associated type
```rust
#![feature(inherent_associated_types)]
#![allow(incomplete_features)]

pub struct Struct;

impl Struct {
    pub type AssocTy = usize;
    pub const AssocConst: Self::AssocTy = 42;
}
```

rustc only emits the error once, but rustdoc emits it twice. rustdoc output:

```
error[E0223]: ambiguous associated type
  --> $DIR/ambiguous-inherent-assoc-ty.rs:11:27
   |
LL |     pub const AssocConst: Self::AssocTy = 42;
   |                           ^^^^^^^^^^^^^ help: use fully-qualified syntax: `<Struct as Trait>::AssocTy`

error[E0223]: ambiguous associated type
  --> $DIR/ambiguous-inherent-assoc-ty.rs:11:27
   |
LL |     pub const AssocConst: Self::AssocTy = 42;
   |                           ^^^^^^^^^^^^^ help: use fully-qualified syntax: `<Struct as Trait>::AssocTy`

For more information about this error, try `rustc --explain E0223`.
```

GuillaumeGomez:
Resolves #88593.

The errors are deduplicated when displayed to users. They only appear
multiple times in UI tests.

cc @jyn514 
r? @camelid

### DIFF
--- a/src/test/rustdoc-ui/ambiguous-inherent-assoc-ty.rs
+++ b/src/test/rustdoc-ui/ambiguous-inherent-assoc-ty.rs
@@ -11,7 +11,6 @@ impl Struct {
     pub const AssocConst: Self::AssocTy = 42;
     //~^ ERROR ambiguous associated type
     //~| HELP use fully-qualified syntax
-    // FIXME: for some reason, the error is shown twice with rustdoc but only once with rustc
     //~| ERROR ambiguous associated type
     //~| HELP use fully-qualified syntax
 }


jyn514 (approved):
Haven't looked at the test case, but the abort_if_errors() call looks right.

Mark-Simulacrum:
rustc itself generally deduplicates errors, that's just turned off for UI tests. Does this duplication reproduce outside of UI tests? If not, it may be preferable to avoid the early abort, which we're generally trying to move away from where possible.

camelid:
r=me with the commit message amended to include something like this:

> The errors are deduplicated when displayed to users. They only appear
multiple times in UI tests.

PR #89916, opened 2021-10-15
Fix Iterator::advance_by contract inconsistency

the8472:
The `advance_by(n)` docs state that in the error case `Err(k)` that k is always less than n.
It also states that `advance_by(0)` may return `Err(0)` to indicate an exhausted iterator.
These statements are inconsistent.
Since only one implementation (Skip) actually made use of that I changed it to return Ok(()) in that case too.

While adding some tests I also found a bug in `Take::advance_back_by`.

### DIFF
--- a/library/alloc/tests/vec.rs
+++ b/library/alloc/tests/vec.rs
@@ -985,6 +985,9 @@ fn test_into_iter_advance_by() {
 
     assert_eq!(i.advance_by(usize::MAX), Err(0));
 
+    i.advance_by(0).unwrap();
+    i.advance_back_by(0).unwrap();
+
     assert_eq!(i.len(), 0);
 }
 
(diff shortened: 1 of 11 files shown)


dtolnay (approved):
Great! I also noticed in #91000 that the "k always less than n" was inconsistent with the 0 special case. Thanks for fixing this.

ehuss:
@bors r-

Failed in rollup: https://github.com/rust-lang/rust/pull/91024#issuecomment-973477446

Looks like this fails when run with overflow-checks enabled.

ehuss:
The reason CI checks are green here is because overflow checks were added after this PR was submitted (in #89776).

dtolnay (approved):

the8472:
Thanks for all the reviews @dtolnay

Issue #70579, opened 2020-03-30
Measure binary size impact of implicit caller location


anp:
The largest unanswered questions about [implicit caller location](https://github.com/rust-lang/rust/issues/47809) are about the binary size impact of the feature and how to offer users control over it.

There are three ways I see that `#[track_caller]` inflates binaries:

1. code bloat from panicking branches that were previously fused
2. encoding the `Location` structs themselves
3. encoding the `&'static str` for every filename referenced by a `Location`

We need to determine the impact to the ecosystem for each of these before deciding on mitigation strategy.

(1) was the main concern in the RFC but it was predicated on the assumption that `#[track_caller]` would affect inlining (an artifact of the implementation proposed). Since it doesn't affect inlining now, the only lost optimization in panicking branches would be for cases where LLVM previously used knowledge of a single constant panic location to make the failure path of a function smaller.

We've not yet had any reports of this and some ad-hoc measurements don't show any difference. This matches my mental model of LLVM's ability to optimize calling conventions which is "very expensive magic". I'm open to ideas for how to better assess this, including "wait and see".

**TLDR** Re (2) and (3),

`Location`s are 24 bytes when `size_of::<usize>() == size_of::<u64>()` and we don't currently measure how many we encode. We should put these in their own section to make it easier to measure.

Which filenames do we encode? In practice: probably all source files. We should put these in their own section too.

anp:
@eddyb could you provide some pointers on how to put static allocations in their own sections? I assume it's a change to `librustc_codegen_ssa`?

eddyb:
@anp Yeah, look at how `link_section` is handled and do the same.

Mark-Simulacrum:
Have we looked at a diff of a stage1 (or stage2) rustc build? rustc has a lot of expect/unwrap etc code so I suspect if there's essentially no delta then I'm not too worried. I haven't noticed nightlies getting much bigger so I suspect we in practice aren't growing by too much.

eddyb:
@Mark-Simulacrum this reminds me that we're not measuring size changes on perf.rust-lang.org.

anp:
Filed: https://github.com/rust-lang/rustc-perf/issues/643

anp:
[...]

PR #89920, opened 2021-10-15
Implement -Z location-detail flag

### LINKED ISSUE #70580: Mitigate binary size impact of implicit caller location
Once we've [measured the size impact](https://github.com/rust-lang/rust/issues/70579) of [implicit caller location](https://github.com/rust-lang/rust/issues/47809), we need to decide what (if any) knobs to offer users to mitigate that impact.

The RFC originally proposed a [`-Zlocation-detail-control`](https://github.com/rust-lang/rfcs/blob/master/text/2091-inline-semantic.md#location-detail-control) flag for rustc which would allow users to control which fields of `Location` were preserved.

On the tracking issue [I also proposed](https://github.com/rust-lang/rust/issues/47809#issuecomment-605530430) an all-or-nothing flag like `-Zredact-caller-location` which would either preserve `Location`s or rewrite them all to be the same.

Note that both of these options are limited to controlling the locations reported by user code -- unless a user recompiles `std` with whatever flag we offer, there will always be "unredacted" locations in their binary.

I'm fond of ideas that would allow us to replace `Location`s with something smaller that can be "symbolized" similarly to the way a debugger for an embedded device can identify the source location for a program counter value in a stacktrace. I *think* most options in this direction require help from the linker that isn't available, but I'd love to be wrong.

hudson-ayers:
This PR implements the `-Z location-detail` flag as described in https://github.com/rust-lang/rfcs/pull/2091 . 

`-Z location-detail=val` controls what location details are tracked when using `caller_location`. This allows users to control what location details are printed as part of panic messages, by allowing them to exclude any combination of filenames, line numbers, and column numbers. This option is intended to provide users with a way to mitigate the size impact of `#[track_caller]`.

Some measurements of the savings of this approach on an embedded binary can be found here: https://github.com/rust-lang/rust/issues/70579#issuecomment-942556822 .

Closes #70580 (unless people want to leave that open as a place for discussion of further improvements).

This is my first real PR to rust, so any help correcting mistakes / understanding side effects / improving my tests is appreciated :) 

I have one question: RFC 2091 specified this as a debugging option (I think that is what -Z implies?). Does that mean this can never be stabilized without a separate MCP? If so, do I need to submit an MCP now, or is the initial RFC specifying this option sufficient for this to be merged as is, and then an MCP would be needed for eventual stabilization?

### DIFF
--- a/compiler/rustc_const_eval/src/interpret/intrinsics/caller_location.rs
+++ b/compiler/rustc_const_eval/src/interpret/intrinsics/caller_location.rs
@@ -80,10 +80,17 @@ impl<'mir, 'tcx: 'mir, M: Machine<'mir, 'tcx>> InterpCx<'mir, 'tcx, M> {
         line: u32,
         col: u32,
     ) -> MPlaceTy<'tcx, M::PointerTag> {
-        let file =
-            self.allocate_str(&filename.as_str(), MemoryKind::CallerLocation, Mutability::Not);
-        let line = Scalar::from_u32(line);
-        let col = Scalar::from_u32(col);
+        let loc_details = &self.tcx.sess.opts.debugging_opts.location_detail;
+        let file = if loc_details.file {
+            self.allocate_str(&filename.as_str(), MemoryKind::CallerLocation, Mutability::Not)
+        } else {
+            // FIXME: This creates a new allocation each time. It might be preferable to
+            // perform this allocation only once, and re-use the `MPlaceTy`.
+            // See https://github.com/rust-lang/rust/pull/89920#discussion_r730012398
+            self.allocate_str("<redacted>", MemoryKind::CallerLocation, Mutability::Not)
+        };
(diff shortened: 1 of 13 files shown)


davidtwco (approved):
LGTM, r=me once these comments are resolved.

korran:
How do you use this flag to disable ALL the location details?

When I set `-Z location-detail=` (an empty comma-separated list), the compiler complains:

```
incorrect value `` for unstable option `location-detail` - comma separated list of location details to track: `file`, `line`, or `column` was expected
```

hudson-ayers:
I think I misunderstood how CLI option parsing works when implementing this, I thought passing an empty string to the option would pass `Some("")` to the parser, but that does not seem to be the case, so it is not possible to disable all the location details. I will followup with a fix and an additional test for this case soon

hudson-ayers:
#99620 contains the fix

PR #84448, opened 2021-04-22
add env var to override sys::Instant::actually_monotonic() for windows and unix

the8472:
The environment variable is read on first use and cached in an atomic.

Various operating systems *promise* monotonic clocks and so does hardware and various hypervisors, often with explicit flags such as  `constant_tsc`. And if that flag is absent they already force the timers to be monotonic through atomic operations (just as the standard library does). And this tends to work on most systems.

But there's the occasional broken hardware or hypervisor that makes this promise but then doesn't deliver. That's why the standard library doesn't rely on the API guarantees in some cases (e.g. windows). And in other cases (e.g. x86 linux) it does trust the OS guarantee and then this gets broken because rust trusts the os which trusts the hypervisor which trusts the hardware which is broken.

The result is that either we err on the side of caution and introduce cache contention in an operation that should be very fast and perfectly scalable even on systems that have reliable clocks or we trust too much and our guarantees get violated on some fraction of systems.

With the environment variable we can offer a way out. We can make a default decision for a particular platform depending on how common broken hardware is and then give users that encounter the opposite case a way out.

Questions:

* this, #83093 or both?

PR #89926, opened 2021-10-15
make `Instant::{duration_since, elapsed, sub}` saturating and remove workarounds

### LINKED ISSUE #86470: supplied instant is later than self
### Meta

`rustc --version --verbose`:
```
rustc 1.52.1 (9bc8c42bb 2021-05-09)
binary: rustc
commit-hash: 9bc8c42bb2f19e745a63f3445f1ac248fb015e53
commit-date: 2021-05-09
release: 1.52.1
LLVM version: 12.0.0
```

### Error output

```
thread 'rustc' panicked at 'supplied instant is later than self', library/std/src/time.rs:281:48

note: rustc 1.52.1 (9bc8c42bb 2021-05-09) running on x86_64-unknown-linux-gnu

note: compiler flags: -C opt-level=3 -C embed-bitcode=no --crate-type lib

```

<p>

```
0: rust_begin_unwind
             at /rustc/9bc8c42bb2f19e745a63f3445f1ac248fb015e53/library/std/src/panicking.rs:493:5
   1: core::panicking::panic_fmt
             at /rustc/9bc8c42bb2f19e745a63f3445f1ac248fb015e53/library/core/src/panicking.rs:92:14
   2: core::option::expect_failed
             at /rustc/9bc8c42bb2f19e745a63f3445f1ac248fb015e53/library/core/src/option.rs:1321:5
   3: core::option::Option<T>::expect
             at /rustc/9bc8c42bb2f19e745a63f3445f1ac248fb015e53/library/core/src/option.rs:349:21
   4: std::time::Instant::duration_since
             at /rustc/9bc8c42bb2f19e745a63f3445f1ac248fb015e53/library/std/src/time.rs:281:9
   5: <std::time::Instant as core::ops::arith::Sub>::sub
             at /rustc/9bc8c42bb2f19e745a63f3445f1ac248fb015e53/library/std/src/time.rs:406:9
   6: std::time::Instant::elapsed
             at /rustc/9bc8c42bb2f19e745a63f3445f1ac248fb015e53/library/std/src/time.rs:345:9
   7: rustc_codegen_llvm::base::compile_codegen_unit
   8: <rustc_codegen_llvm::LlvmCodegenBackend as rustc_codegen_ssa::traits::backend::CodegenBackend>::codegen_crate
   9: rustc_interface::passes::QueryContext::enter
  10: rustc_interface::queries::Queries::ongoing_codegen
  11: rustc_interface::queries::<impl rustc_interface::interface::Compiler>::enter
  12: rustc_span::with_source_map
  13: rustc_interface::interface::create_compiler_and_run
  14: scoped_tls::ScopedKey<T>::set
```

</p>
</details>

This might be an LLVM bug, as it never occurs on Windows and occurs in every rust program that I've compiled on Linux, not when compiling, but running it.

the8472:
This removes all mutex/atomic-based workarounds for non-monotonic clocks and makes the previously panicking methods saturating instead. Additionally `saturating_duration_since` becomes deprecated since `duration_since` now fills that role.

Effectively this moves the fixup from `Instant` construction to the comparisons.

This has some observable effects, especially on platforms without monotonic clocks:

* Incorrectly ordered Instant comparisons no longer panic in release mode. This could hide some programming errors, but since debug mode still panics tests can still catch them.
* `checked_duration_since` will now return `None` in more cases. Previously it only happened when one compared instants obtained in the wrong order or manually created ones. Now it also does on backslides.
* non-monotonic intervals will not be transitive, i.e. `b.duration_since(a) + c.duration_since(b) != c.duration_since(a)`

The upsides are reduced complexity and lower overhead of `Instant::now`.

## Motivation

Currently we must choose between two poisons. One is high worst-case latency and jitter of `Instant::now()` due to explicit synchronization; see #83093 for benchmarks, the worst-case overhead is > 100x. The other is sporadic panics on specific, rare combinations of CPU/hypervisor/operating system due to platform bugs.

Use-cases where low-overhead, fine-grained timestamps are needed - such as syscall tracing, performance profiles or sensor data acquisition (drone flight controllers were mentioned in a libs meeting) in multi-threaded programs - are negatively impacted by the synchronization.

The panics are user-visible (program crashes), hard to reproduce and can be triggered by any dependency that might be using Instants for any reason.

A solution that is fast _and_ doesn't panic is desirable.

----

closes #84448 
closes #86470


jhpratt:
If I have `instant2.duration_since(instant1) == Duration::from_secs(0)`, logic would say that `instant1 == instant2`. If the operation saturates, this invariant would no longer be upheld. Making `elapsed` saturate instead of panic has similar issues as you can modify the inner value of an `Instant` (via add-assigning a `Duration`).

While I understand the motivation, I don't think breaking what to me are logical invariants (even if not explicitly stated) is a reasonable solution.

joshtriplett:
I love the idea of making this change, and removing all the associated complexity and panics.

I think in the common case people won't get two instants backwards, because they can call `start.elapsed()` rather than calling `Instant::now()` again. I can imagine cases motivating a second call to `Instant::now()` (e.g. making sure several durations are back-to-back by using the end of one interval as the start of the next), but it still seems acceptable to not catch reversing the two. It might also be possible to implement a simple lint for common cases.

I don't think this would affect any invariants people rely on in practice. And people can always call `checked_duration_since(...).expect(...)` if they want a panic.

nagisa:
Is there any reason why we might want to keep `actually_monotonic` in places and continue using a simple subtraction instead of a saturating one on those platforms?

the8472:
Currently there is no wrapping subtraction available on the `sys::time::Instant` implementations because on many platforms it's a composite type consisting of separate seconds and nanos fields. Only checked add/sub are provided.
Only the solid/itron target uses an Instant that's a single `u64` field.

the8472:
As discussed in the libs-api meeting I have changed the methods to panic in debug mode (via overflow-checks) and saturate in release mode.

Discussion on exactly which of `{duration_since, elapsed, sub}` should saturate or panic is still open.

matklad:
I think we hit "time goes backwards in a VM" bug in NEAR: https://near.zulipchat.com/#narrow/stream/295302-dev-general/topic/general/near/248247028. So, for us, such a change would be very much welcome. 

*However*, there's a wrinkle -- we compile our code in `--release` mode, but with overflow checks on. That is:

* We **do** want to panic due to "logical errors" with arithmetics
* If the underlying hardware/hypervisor/OS breaks monotonicity guarantees, we **don't** want to panic

For `--release,overflow-checks = true`, I fear that this specific implementation would be problematic -- it'll lead to strictly more panics than we have today (as we remove `is_actually_monotonic` check).

Ideally, I think we shouldn't be piggybacking on `overflow-checks` here, but instead used `cfg!(debug_assertions)`. But of course it's not something we can do in std. 

To sum up: 

* I really like "panic in debug, saturate in release" approach, as it improves both reliability and performance 
* The specific `overflow_checks` approach seems to make `--release,overflow-checks = true` more panicky, and that feels like a major regression. 
* don't know how to what alternative impl approach would be better here though.

the8472:
> I think we hit "time goes backwards in a VM" bug in NEAR

If you're already hitting those panics that should mean that `actually_monotonic() == false` on your platform and the hardware clock is getting passed through unfiltered. So this change wouldn't make things worse for you, you just wouldn't get the intended benefits.

Also, that link would require signup to read it.

----

Depending on how much control you have over where the Instant's are compared maybe you could selectively change the overflow-checks flag for those crates via profile overrides?

matklad:
Yeah, practically for near specifically I think we currently only run on `is_actually_monotonic` platforms. But in general any code which is compiled with `release,overflow-checks=true` mode and is run on `!is_actually_monotonic` platform will see more panics. Can't point to specific projects which use this scenario, but I'd be surprised if there aren't any. 

>Also, that link would require signup to read it.

Oups, good catch! That's the usual Zulip's "login with GitHub" thing, we should set-up zulip archive. The short discussion here basically just links to https://github.com/tokio-rs/tokio/issues/3619

the8472:
Overflow checks are now enabled in CI but they aren't by default when testing locally. I haven't found anything in the std test suite that's conditional on overflow checks... that means the tests would have to be removed completely since their behavior differs depending on configuration. And that also means overflow panics aren't tested anywhere? 😕

joshtriplett:
Proposal from the @rust-lang/libs-api meeting today:

- We'd like `duration_since` and `sub` to saturate in release mode, but still assert when overflow checks are on. This will help catch cases where people get the arguments backwards (so asserts here are not just "your system is broken").
- We'd like `elapsed` to always saturate. This removes a panic that always means either "your system is broken" or "you went out of your way to hand-construct an `Instant` in the future and call `elapsed` on it".

Let's use an FCP to gauge consensus on this proposal:

@rfcbot merge

the8472:
> Proposal from the rust-lang/libs-api meeting today:
> 
>  * We'd like `duration_since` and `sub` to saturate in release mode, but still assert when overflow checks are on. This will help catch cases where people get the arguments backwards (so asserts here are not just "your system is broken").
> 
>  * We'd like `elapsed` to always saturate. This removes a panic that always means either "your system is broken" or "you went out of your way to hand-construct an `Instant` in the future and call `elapsed` on it".
> 

I have made those changes and also removed the deprecation of `saturating_duration_since` since `duration_since` now saturates conditionally while the former saturates unconditionally.

library/std/src/time.rs
```diff
@@ -359,9 +344,9 @@ impl Instant {
     ///
     /// # Panics
     ///
-    /// This function may panic if the current time is earlier than this
-    /// instant, which is something that can happen if an `Instant` is
-    /// produced synthetically.
+    /// In debug mode this function will panic if the current time is earlier than `self`,
+    /// which is something that can happen if an `Instant` is produced synthetically or due to
+    /// hardware bugs. In release mode it will saturate instead.
```
yaahc, line 349: Is this comment still accurate? Based on my read of the issue this should saturate on debug too now

yaahc:
> While I understand the motivation, I don't think breaking logical invariants (even if not explicitly stated) is a reasonable solution.

I empathize with this position and am vaguely worried about this potentially causing issues down the line, but I feel that I have to acknowledge that this is putting a vague worry against concrete problems with performance that our current workarounds have. I think I'm in favor of this change but I do think we should also update the documentation to mention this unintuitive behavior.

Additionally, I worry that if our vague fear ever does manifest as a concrete issue in it's own right we won't be able to undo it. I want to take some time to fully consider the forward compatibility of the changes in this PR, and what breaking changes we may want to explicitly reserve the right to make now in case this becomes an issue in the future. I think we may want to reserve the right to introduce panics in the future on all of these functions.

> Incorrectly ordered Instant comparisons no longer panic in release mode. This could hide some programming errors, but since debug mode still panics tests can still catch them.

I'm also a little worried about this assumption. In my experience as codebases get larger they tend to trend towards running tests in release mode. We may want to make an extra effort to make sure people are aware of the this behavior change in release mode and specifically recommend people consider enabling `overflow-checks` on tests even if compiled in release.

library/std/src/time.rs
```diff
@@ -306,11 +279,23 @@ impl Instant {
...
+    #[rustc_inherit_overflow_checks]
+    #[allow(arithmetic_overflow)]
     pub fn duration_since(&self, earlier: Instant) -> Duration {
-        self.0.checked_sub_instant(&earlier.0).expect("supplied instant is later than self")
+        match self.checked_duration_since(earlier) {
+            Some(d) => d,
+            None => {
+                // In debug builds, trigger a panic on None.
+                // This should optimize completely out in release builds.
+                let _ = usize::MAX + 1;
```
ben0x539, line 294: As an aside, to me as a casual, this is mystifying. Does this do something different than a `debug_assert!(false)`? I believe I understand that we cannot use `debug_assert!()` in `std` because debug builds of user code still link against a release build of std, but surely this approach isn't fundamentally different? Is this specifically targeted at people building their own std who don't want debug assertions in general but do want this particular panic?

the8472: That's because std is shipped prebuilt with rustc and without debug asserts enabled. But for overflow checks there's the special mechanism `#[rustc_inherit_overflow_checks]` that does enable overflow checks in std when the user crate has them enabled. So we're abusing that mechanism here to allow conditionally panic depending on user configuration.

The documentation is vague by saying "debug mode" since that covers both overflow checks and debug asserts so we can change it in the future from overflow checks to debug asserts.

the8472:
> We may want to make an extra effort to make sure people are aware of the this behavior change in release mode and specifically recommend people consider enabling overflow-checks on tests even if compiled in release.

Do people run `cargo test --profile release` or do they adjust the test profile? I would expect the latter and in that case they would have to explicitly disable overflow checks in addition to bumping the optimization level for this to be relevant.

the8472:
> Additionally, I worry that if our vague fear ever does manifest as a concrete issue in it's own right we won't be able to undo it. I want to take some time to fully consider the forward compatibility of the changes in this PR, and what breaking changes we may want to explicitly reserve the right to make now in case this becomes an issue in the future. For now I think we may want to reserve the right to introduce panics in the future on all of these functions.

In java documentation there's the distinction between {API, implementation}×{specification, note} with [explicit annotations](https://openjdk.java.net/jeps/8068562#Description) for the latter three.

Since we lack standardized means to distinguish those I can phrase this as a "may panic" and elaborate the history, current state and future possibilities on the struct docs and reference that from the method docs.

That said, if we change this in response to bug reports then inevitably a different set will of people will file bugs after the change back, as they do today. Of course there are still other options left such as declaring the whole thing not our problem or #84448

yaahc:
> Since we lack standardized means to distinguish those I can phrase this as a "may panic" and elaborate the history, current state and future possibilities on the struct docs and reference that from the method docs.

Including the "may panic" phrase is all I really want personally, I'll leave it up to you on whether or not you wish to include the additional context.

library/std/src/time.rs
eddyb: > But for overflow checks there's the special mechanism `#[rustc_inherit_overflow_checks]` that does enable overflow checks in std when the user crate has them enabled.

Keep in mind this can only work with cross-crate codegen, and I can't find anything in the compiler where the attribute implies `#[inline]` or similar (and there's no checks for whether it's misused, either) - so you likely also need `#[inline]` on this function for it to do anything.

I would suggest adding a test that explicitly sets the overflow checks flags and then making sure the test only passes with `#[inline]`, to confirm this.

eddyb: You probably also need `#[track_caller]` to avoid the panic pointing at the implementation, but that part seems more subjective.

Mark-Simulacrum:
In general impl looks OK to me, though looks like there are a couple review comments -- and FCP still needs to complete.

library/std/src/time.rs
the8472: I asked here https://github.com/rust-lang/rust/pull/89926#discussion_r733149105 and was informed that putting it on the method itself is sufficient.

library/std/src/time.rs
the8472: Updated the comments with the requested "may panic" changes.

library/std/src/time.rs
```diff
@@ -125,6 +124,26 @@ pub use core::time::FromSecsError;
...
 ///
 /// [`add`]: Instant::add
+///
+/// ## Monotonicity
+///
+/// All [tier 1] and most tier 2 platforms with std support rely on monotonicity guarantees
+/// from the operating system. Rarely these guarantees are broken by hardware, virtualization
+/// or operating system bugs. In previous versions this resulted in panics in
+/// [`duration_since`], [`elapsed`] and [`sub`]. Currently they will either saturate to zero or,
+/// to let tests detect logic bugs such as swapped earlier and later times, only panic in debug mode.
```
Mark-Simulacrum, line 134:
```suggestion
/// to let tests detect logic bugs such as swapped earlier and later times, panic when overflow checks are enabled.
```

Right? Debug mode is not really describing this perfectly.

library/std/src/time.rs
```diff
@@ -306,11 +283,23 @@ impl Instant {
...
+    #[rustc_inherit_overflow_checks]
+    #[allow(arithmetic_overflow)]
     pub fn duration_since(&self, earlier: Instant) -> Duration {
-        self.0.checked_sub_instant(&earlier.0).expect("supplied instant is later than self")
+        match self.checked_duration_since(earlier) {
+            Some(d) => d,
+            None => {
+                // In debug builds, trigger a panic on None.
+                // This should optimize completely out in release builds.
+                let _ = usize::MAX + 1;
```
Mark-Simulacrum, line 298: Presumably the error message produced here is *terrible* -- not saying anything about duration/instant. That feels pretty unfortunate to me. Maybe we should land a compiler intrinsic or similar that basically is a `is_overflow_checks_enabled()` function? That way we can do a normal panic here, with a much better message.

I guess without track_caller we'd at least have the panic originate in library/std/src/time.rs, which seems at least a little better, but maybe indicates we should have a comment on this pointing readers to Instant's monotonic section in the docs so if they look at source code they can figure out what's up fairly quickly.

Mark-Simulacrum:
FWIW, the effect of forcing users to choose between overflow checks (otherwise, pretty much always a good thing to enable, if you can afford it) and seeing this bug from the underlying hardware/OS seems pretty unfortunate. I don't think we necessarily have a better option as things stand today, unfortunately -- exposing a dedicated flag seems a bit much, though maybe is not a terrible idea. I worry that it hurts our long-term desire (AIUI) to enable overflow checks on all profiles by default.

This is basically @matklad's concern in https://github.com/rust-lang/rust/pull/89926#issuecomment-955202825, I suppose, but I don't think the libs-api meetings I was in at least discussed it too prominently.

I've added the relnotes tag so we at least indicate this, but I do wonder if it makes sense to try to do a broader survey of folks using overflow-checks = true today. It would seem pretty unfortunate if folks start *disabling* that as a result of this landing, because they can't afford their binaries faulting when deployed to random systems but do want to check arithmetic (since that's programmer bugs). I guess the message is "use saturating_duration_since/elapsed" there -- maybe a lint, allow by default, is appropriate to allow people to migrate more easily before moving forward here.

Maybe I'm over-rotating on this concern, too, not sure.

m-ou-se:
@rfcbot concern debug-vs-release / overflowchecks

library/std/src/time.rs
teor2345: At the very least, we could rename the constant, so that the error message contains a hint (and a string that's easy to search for).

```suggestion
                const DURATION_SINCE_ON_FUTURE_TIME: usize = usize::MAX;
                let _ = DURATION_SINCE_ON_FUTURE_TIME + 1;
```

joshtriplett:
We discussed this extensively in the last @rust-lang/libs-api meeting, and many people had concerns about tying this to overflow checks. In particular, people may want to have overflow checks enabled (for arithmetic operations) but still not want `Instant` operations to panic (because that will break for reasons outside their control, such as running on buggy VMs/hardware where time appears to go backwards).

We subsequently did an informal poll on Zulip. The question of which alternative people would *prefer* had a broader spectrum of preferences. However, the question of which alternatives people *object to* was quite definitive. There was only one option that nobody on the team objected to: Always saturate and return 0 if the "later" time is earlier than the "earlier" time.

Given that we've changed out minds before, I don't want to ask @the8472 to proceed down that path without confirming that consensus.

So, can we establish consensus on the approach of always saturating in `duration_since`, `sub`, and `elapsed`?

@rfcbot cancel
@rfcbot merge

the8472:
> So, can we establish consensus on the approach of always saturating in `duration_since`, `sub`, and `elapsed`?

Ok, FCP has enough boxes ticked. There are two questions I still have:

* does it still make sense to include language that we might reintroduce the panics in the future? originally raised by @yaahc 
* should `saturating_duration_since` be deprecated since it's now redundant with `duration_since`?

Mark-Simulacrum:
I am personally inclined to leave the mention of panicking for the future, but perhaps open a separate PR to remove it (which can be immediately FCP'd/nominated), so that we can merge this in rather than waiting on settling that question. In general it seems *plausible* that if we "find" a cheap oracle that indicated definitely wrong ordering on elapsed, we may wish to introduce a panic in the future -- I could imagine this being done as a sanitizer or similar; but in any case, it feels wrong to block the good improvements in this PR on that :)

I agree with not deprecating saturating_duration_since, at least yet.

joshtriplett:
I agree; we can FCP that separately to avoid blocking.

library/std/src/time.rs
```diff
@@ -125,6 +126,25 @@ pub use core::time::FromSecsError;
 /// > structure cannot represent the new point in time.
 ///
 /// [`add`]: Instant::add
+///
+/// ## Monotonicity
+///
+/// All [tier 1] and most tier 2 platforms with std support rely on monotonicity guarantees
```
Mark-Simulacrum, line 132: This sort of misleadingly (I think) implies that tier 3 platforms are *not* relying on these guarantees. I think what we're trying to say here is that "Instant depends on the underlying platform to provide a monotonic clock" [and sometimes that platform lies/gets it wrong].

Can we avoid talking about tiers at all, perhaps? It doesn't seem particularly meaningful, given we "need" the same guarantees realistically regardless of tier.

library/std/src/time.rs
```diff
@@ -125,6 +126,25 @@ pub use core::time::FromSecsError;
...
+/// from the operating system. Rarely these guarantees are broken by hardware, virtualization
+/// or operating system bugs. To work around these bugs and platforms not offering monotonic clocks
+/// [`duration_since`], [`elapsed`] and [`sub`] saturate to zero. In older rust versions this
```
Mark-Simulacrum, line 135:
```suggestion
/// [`duration_since`], [`elapsed`] and [`sub`] saturate to zero. In older Rust versions this
```

library/std/src/time.rs
```diff
@@ -125,6 +126,25 @@ pub use core::time::FromSecsError;
...
+/// lead to a panic instead. [`checked_duration_since`] can be used to detect and handle situations
+/// where monotonicity is violated.
```
Mark-Simulacrum, line 137:
```suggestion
/// where monotonicity is violated, or `Instant`s are subtracted in the wrong order.
```

Not quite sure about this wording, but I feel like the previous language made checked_duration_since sound like a magic wand a little more than we probably want it to; 99% of users likely won't want to use it.

Can we add some language in this PR to checked_duration_since that points to this paragraph, and says something to the effect of "due to the bugs discussed, even under correct logical ordering of the passed instants, this method can return None".

Mark-Simulacrum (commented):
It also looks like the PR description doesn't quite match the current impl (e.g. mentioning a deprecation, panics depending on debug).

library/std/src/time.rs
the8472: Some of the lower tiers just don't support `Instant` at all and return errors.

Mark-Simulacrum: I presume we always either return errors or ask for a monotonic clock? Not *all* tier 3 targets are not-supporting of instant, right?

the8472: This SOLID (tier3 platform) comment reads like `Instant` is using a system timer rather than a cycle counter.

https://github.com/rust-lang/rust/blob/72e74d7b9cf1a7901650227e74650f1fcc797600/library/std/src/sys/itron/time.rs#L21-L24

the8472: ping @Mark-Simulacrum

Do you still want me to update the wording? The documentation can't speak for all (especially 3rd-party) or future platforms. Some OSes might not guarantee monotonicity but still have a high resolution time source. The question is whether saturating will be enough in those cases or we should keep the mutex/atomics solution around for those cases?

Mark-Simulacrum: Hm... Not sure. I missed your comment before, but it seems to me that if we know Instant isn't going to be monotonic then saturating can yield fairly unexpected and (in some sense) wrong behavior.

I'm not sure either whether the atomic/mutex solution really helps there - if we're not using a source that even claims to be monotonic, we can't really effectively fake it atop imo, since you'd just get a long period of always zero or panics.

I think we may need some team discussion, but my initial inclination is that it may be worth cfg'ing out and/or deprecating Instant for platforms we do not know to have an API that claims monotonicity. (Even if in some hardware or virtualization environments that monotonicity is buggy).

Mark-Simulacrum: DIscussed in libs-api meeting -- I think we probably want to rephrase to something like this:

On tier 1 platforms, Instant will request a monotonic time from the underlying system. On other platforms, the standard library makes a best-effort attempt to use a monotonic API. Note that regardless of the target, the underlying platform may suffer from bugs which break the monotonic guarantee [...].

The trailer here is the rest of the paragraph as typical.

@the8472 let me know if that makes sense to you or you think we should otherwise rephrase.

Mark-Simulacrum: Yeah, I specifically wanted to avoid a promise that sounded like we *definitely* use the API (e.g., we may not on a tier 2/3 platform because no one has told us it exists yet). I'm not sure how best to communicate that; my best-effort was an attempt to do so.

uazu:
Seeing this on Rust Weekly News gave me a shock. So `Instant` is no longer guaranteed non-decreasing? I think people need a heads-up about this. I think there are cases where this could break people's assumptions about ordering of operations, for example if I set two timers at one second from now, setting them from times 1ms apart, they could execute in the opposite order. This could break some assumptions the coder is relying on, for example if both of these timers update the same state, with the assumption that the second timer will execute second, then the state could be left with an intermediate value.

I just checked my code, and I don't use `Instant::now()` directly, but I keep a local copy which already has checks so that it can't go backwards (i.e. I already don't trust `Instant::now()`), so my code is okay. I wonder whether it's possible to judge whether this will break things for anyone else. I am in favour of the change, because getting rid of that overhead is an advantage. But if it is a breaking change for anyone, they need to know.

I'm trying to think of the cases where this could break. Maybe when two ordered events come in from an external source, e.g. from hardware or over a network stream, and two timers are set from those events, and then the timers execute in the opposite order, breaking the coder's assumptions. For this to have worked in the case of the previous `Instant` implementation, the timer queue would need to maintain the order of timers submitted for the same `Instant`, but if it did then this would be a breaking change for that coder. If this sounds unlikely to happen in real life code, then perhaps everything is okay.

the8472:
> So Instant is no longer guaranteed non-decreasing? 

From the [updated docs](https://doc.rust-lang.org/nightly/std/time/struct.Instant.html)

> Instants are always guaranteed, barring [platform bugs](https://doc.rust-lang.org/nightly/std/time/struct.Instant.html#monotonicity), to be no less than any previously measured instant when created, and are often useful for tasks such as measuring benchmarks or timing how long an operation takes.

From an API perspective Rust already relied on the [OS guarantees](https://man7.org/linux/man-pages/man3/clock_gettime.3.html#DESCRIPTION) to provide monotonic time. It additionally applies some workarounds that are only relevant in situations when operating system fails to uphold its contract. What changed are the workarounds. What doesn't change is that any violation of monotonicity already is a platform bug, i.e. running rust on a broken system. This is akin to, albeit less severe, a filesystem not upholding its durability guarantees for example.

uazu:
> From an API perspective Rust already relied on the [OS guarantees](https://man7.org/linux/man-pages/man3/clock_gettime.3.html#DESCRIPTION) to provide monotonic time. It additionally applies some workarounds that are only relevant in situations when operating system fails to uphold its contract. What changed are the workarounds. What doesn't change is that any violation of monotonicity already is a platform bug, i.e. running rust on a broken system. This is akin to, albeit less severe, a filesystem not upholding its durability guarantees for example.

I understand that. However if a coder requires a strictly non-decreasing `Instant` (e.g. for their code to operate correctly), previously they could have relied on Rust's implementation to do the work-around in case of platform bugs. Now they need to do the work-around themselves. Maybe that's a reasonable thing to ask them to do, but at least they'll need a heads-up so that they can check their code (as I did), e.g. a warning in the Rust release notes or something.

cuviper:
This is labeled `relnotes` already, but please feel free to check on that when we're drafting 1.60.0.

nyetwurk:
IMO what is missing from this thread is any mention of pressuring vendors that have *known broken VM/hypervisors* in production (AWS). Hiding this is enabling their continued bad behavior.

PR #89930, opened 2021-10-15
Only use `clone3` when needed for pidfd

cuviper:
In #89522 we learned that `clone3` is interacting poorly with Gentoo's
`sandbox` tool. We only need that for the unstable pidfd extensions, so
otherwise avoid that and use a normal `fork`.

This is a re-application of beta #89924, now that we're aware that we need
more than just a temporary release fix. I also reverted 12fbabd27f700, as
that was just fallout from using `clone3` instead of `fork`.

r? @Mark-Simulacrum 
cc @joshtriplett

### DIFF
--- a/library/std/src/sys/unix/process/process_unix.rs
+++ b/library/std/src/sys/unix/process/process_unix.rs
@@ -166,6 +166,10 @@ impl Command {
             fn clone3(cl_args: *mut clone_args, len: libc::size_t) -> libc::c_long
         }
 
+        // Bypassing libc for `clone3` can make further libc calls unsafe,
+        // so we use it sparingly for now. See #89522 for details.
+        let want_clone3_pidfd = self.get_create_pidfd();
+
         // If we fail to create a pidfd for any reason, this will
         // stay as -1, which indicates an error.
         let mut pidfd: pid_t = -1;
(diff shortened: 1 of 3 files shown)


cuviper:
@joshtriplett ping?

We'll need to backport this for 1.57 too.

library/std/src/sys/unix/process/process_unix.rs
```diff
@@ -166,14 +166,20 @@ impl Command {
...
         // stay as -1, which indicates an error.
         let mut pidfd: pid_t = -1;
 
         // Attempt to use the `clone3` syscall, which supports more arguments
         // (in particular, the ability to create a pidfd). If this fails,
         // we will fall through this block to a call to `fork()`
-        if HAS_CLONE3.load(Ordering::Relaxed) {
+        if want_clone3 && HAS_CLONE3.load(Ordering::Relaxed) {
             let mut flags = 0;
             if self.get_create_pidfd() {
```
joshtriplett, line 184: This conditional is now redundant.

cuviper: It's only redundant if you assume that `want_clone3` will never encompass anything else.

cuviper: OK, I've renamed the flag to narrow its scope, then removed the inner conditional.

PR #89937, opened 2021-10-16
Properly check `target_features` not to trigger an assertion

### LINKED ISSUE #89875: ICE with asm macro
### Code

```Rust
#![feature(asm)]

#[target_feature(enable = "avx")]
pub unsafe fn foo() {
    asm!(
        "/* {} */",
        out(ymm_reg) _,
    );
}
```

### Meta

`rustc --version --verbose`:
```
rustc 1.57.0-nightly (dfc5add91 2021-10-13)
binary: rustc
commit-hash: dfc5add915e8bf4accbb7cf4de00351a7c6126a1
commit-date: 2021-10-13
host: x86_64-pc-windows-msvc
release: 1.57.0-nightly
LLVM version: 13.0.0
```

### Error output

```
thread 'rustc' panicked at 'assertion failed: matches!(reg, InlineAsmRegOrRegClass :: Reg(_))', compiler\rustc_codegen_llvm\src\asm.rs:179:25

note: rustc 1.57.0-nightly (dfc5add91 2021-10-13) running on x86_64-pc-windows-msvc

note: compiler flags: -C embed-bitcode=no -C debuginfo=2 -C incremental --crate-type lib

query stack during panic:
end of query stack
error: could not compile `foo`
```

Also, `#[target_feature(enable = "avx2")]` doesn't allow the use of `ymm_reg`:
```
error: register class `ymm_reg` requires the `avx` target feature
```

JohnTitor:
Fixes #89875
I think it should be a condition instead of an assertion to check if it's a register as it's possible that `reg` is a register class.
Also, this isn't related to the issue directly, but `is_target_supported` doesn't check `target_features` attributes. Is there any way to check it on rustc_codegen_llvm?

r? @Amanieu

### DIFF
--- a/compiler/rustc_ast_lowering/src/asm.rs
+++ b/compiler/rustc_ast_lowering/src/asm.rs
@@ -214,9 +214,7 @@ impl<'a, 'hir> LoweringContext<'a, 'hir> {
                 // means that we disallow passing a value in/out of the asm and
                 // require that the operand name an explicit register, not a
                 // register class.
-                if reg_class.is_clobber_only(asm_arch.unwrap())
-                    && !(op.is_clobber() && matches!(reg, asm::InlineAsmRegOrRegClass::Reg(_)))
-                {
+                if reg_class.is_clobber_only(asm_arch.unwrap()) && !op.is_clobber() {
                     let msg = format!(
                         "register class `{}` can only be used as a clobber, \
                              not as an input or output",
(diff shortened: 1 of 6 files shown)


JohnTitor:
Hmm, I'm not sure about the right condition then. Since #89641, the target_features check has been moved to `rustc_passes`, we should add a check there instead of `rustc_ast_lowering`, right? I'm also not sure how we skip clobbering on `rustc_passes` (or `rustc_ast_lowering`).

By the way, the test passes fine on my local and [godbolt](https://rust.godbolt.org/z/rnzTeoT5K) if I pass the target feature as a compile flag, but CI doesn't, any ideas?

PR #89945, opened 2021-10-16
Remove a mention to `copy_from_slice` from `clone_from_slice` doc

### LINKED ISSUE #84736: `clone_from_slice` doc can be updated after specialization
Current `clone_from_slice` doc says that the user might prefer to use `copy_from_slice` for `Copy` types, but since #81854 we automatically take care of this with specialization. We can consider to clarify docs.

The most careful action would be to let the current docs stand, if we don't want to say anything about the implementation details. I would prefer to remove the sentence about `copy_from_slice`, or something similar.

JohnTitor:
Fixes #84736
I think removing it would be the best but I'm happy to clarify it instead if someone would like.

### DIFF
--- a/library/core/src/slice/mod.rs
+++ b/library/core/src/slice/mod.rs
@@ -2953,9 +2953,6 @@ impl<T> [T] {
     ///
     /// The length of `src` must be the same as `self`.
     ///
-    /// If `T` implements `Copy`, it can be more performant to use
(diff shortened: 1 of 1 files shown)


the8472:
Removal seems fine to me since it doesn't commit to anything we're doing via specialization

@bors r+

PR #89951, opened 2021-10-16
Stabilize `option_result_unwrap_unchecked`

### LINKED ISSUE #81383: Tracking Issue for `option_result_unwrap_unchecked`
Feature gate: `#![feature(option_result_unwrap_unchecked)]`

This is a tracking issue for the unchecked variants of `unwrap()` and `unwrap_err()` of `Option` and `Result`.

These are unsafe functions intended to be used in hot paths of performance sensitive applications where the variant an `Option`/`Result` holds is known, allowing to skip the branch that the safe version introduces. They complement other `*_unchecked()` methods already available, such as `get_unchecked()` in slices.

### Public API

```rust
impl<T> Option<T> {
    pub unsafe fn unwrap_unchecked(self) -> T;
}

impl<T, E> Result<T, E> {
    pub unsafe fn unwrap_unchecked(self) -> T;
    pub unsafe fn unwrap_err_unchecked(self) -> E;
}
```

### Steps / History

- [x] Implementation: #80876
- [x] Stabilization report: https://github.com/rust-lang/rust/issues/81383#issuecomment-944498212
- [ ] Final commenting period (FCP): https://github.com/rust-lang/rust/issues/81383#issuecomment-945379226
- [ ] Stabilization PR: https://github.com/rust-lang/rust/pull/89951

### Unresolved Questions

- [x] Should we add `expect_unchecked`, and if yes, do it within another feature flag, to unblock `unwrap_unchecked`? https://github.com/rust-lang/rust/issues/81383#issuecomment-801492960
  + It seems best to have it under another feature flag to avoid blocking this one.

- [x] Should we panic on failure in debug (i.e. for those projects that build `core` themselves)? If yes, that would mean there is no way to actually remove the check for those projects, which could be potentially needed by someone. If we do, should it be documented? https://github.com/rust-lang/rust/issues/81383#issuecomment-824654722
  + The current implementation uses `debug_assert!`, which may be useful for projects that build `core` with debug assertions enabled (e.g. the bors test suite or some embedded projects), but this is *not* documented, and indeed most users will not see such behavior, since `core` is shipped in release mode.
  + Furthermore, since it is UB anyway, we have room for changing the implementation, e.g. if we end up distributing the standard library with debug assertions enabled for some use cases.

- [x] Should we name them `unchecked_unwrap` instead? https://github.com/rust-lang/rust/issues/81383#issuecomment-942533948.
  + The current name (`unwrap_unchecked`) follows the convention of other stable methods in the standard library such as `get_unchecked`, `map_unchecked`, `new_unchecked`, `unreachable_unchecked`... (see as well https://rust-lang.github.io/api-guidelines/naming.html#getter-names-follow-rust-convention-c-getter).
  + On the other hand, the `unchecked_{add,...}` methods seem to follow the opposite convention, but those are unstable and they likely do it because the stable `{checked,...}_{add,...}` exist.

ojeda:
Closes https://github.com/rust-lang/rust/issues/81383.

Stabilization report: https://github.com/rust-lang/rust/issues/81383#issuecomment-944498212.

@rustbot label +A-option-result +T-libs-api

### DIFF
--- a/library/alloc/src/lib.rs
+++ b/library/alloc/src/lib.rs
@@ -112,7 +112,6 @@
 #![feature(maybe_uninit_slice)]
 #![cfg_attr(test, feature(new_uninit))]
 #![feature(nonnull_slice_from_raw_parts)]
-#![feature(option_result_unwrap_unchecked)]
 #![feature(pattern)]
 #![feature(ptr_internals)]
 #![feature(receiver_trait)]
--- a/library/core/src/option.rs
+++ b/library/core/src/option.rs
@@ -800,19 +800,17 @@ impl<T> Option<T> {
     /// # Examples
     ///
     /// ```
-    /// #![feature(option_result_unwrap_unchecked)]
     /// let x = Some("air");
     /// assert_eq!(unsafe { x.unwrap_unchecked() }, "air");
     /// ```
     ///
     /// ```no_run
-    /// #![feature(option_result_unwrap_unchecked)]
     /// let x: Option<&str> = None;
     /// assert_eq!(unsafe { x.unwrap_unchecked() }, "air"); // Undefined behavior!
     /// ```
     #[inline]
     #[track_caller]
-    #[unstable(feature = "option_result_unwrap_unchecked", reason = "newly added", issue = "81383")]
+    #[stable(feature = "option_result_unwrap_unchecked", since = "1.58.0")]
     pub unsafe fn unwrap_unchecked(self) -> T {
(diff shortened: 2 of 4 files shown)


dtolnay (approved):
The PR looks good to me once FCP is completed in #81383.

dtolnay (approved):

ojeda:
Not sure if I can do it myself but:

@rustbot label +relnotes

PR #89957, opened 2021-10-16
Add aarch64-apple-darwin to list of platforms with CI LLVM

winterqt:
`aarch64-apple-darwin` is Tier 2 with Host Tools, and has CI LLVM builds.


hkratz:
Duplicate of #89888. Seems like a good idea. :-)

winterqt:
Ah, my apologies @hkratz.

PR #89980, opened 2021-10-17
Avoid copies while creating the metadata object file

bjorn3:
This is now possible thanks to https://github.com/gimli-rs/object/pull/369 and https://github.com/gimli-rs/object/pull/370.


bjorn3:
In the last commit I forgot to replace an occurrence of `append_section_data` with `set_section_data` causing this commit to only improve perf on windows and macos. Fixed it now.

@bors try @rust-timer queue

nagisa:
@bjorn3 do you think we should still land this despite no significant performance improvements? The changes seem non-intrusive enough that I'm happy to see this landing either way. (r=me, your call if you want to land this or not)

bjorn3:
I think it is worth it. It should wait for backtrace-rs to update it's object dependency first to avoid multiple versions of object being used I think.

PR #89983, opened 2021-10-17
Cleanup: Get rid of -static-libstdc++ argument unused warning

hkratz:
Building llvm with llvm.static-libstdcpp enabled leads to lots of warnings like this as this is linker-only option:
```
warning: argument unused during compilation: '-static-libstdc++' [-Wunused-command-line-argument]
```
 Just setting it for linking shared libraries and executables gets rid of the warning.

@rustbot label +A-rustbuild +C-cleanup +T-infra


Mark-Simulacrum:
Hm, it doesn't seem like we dump sccache stats on failed builds unfortunately.

Maybe sccache fails to cache or something like that without this argument? That would be unexpected, IMO, but who knows...?

PR #89989, opened 2021-10-17
rustdoc: Add static size assertion for `clean::Type`

camelid:
r? @jyn514

### DIFF
--- a/src/librustdoc/clean/types.rs
+++ b/src/librustdoc/clean/types.rs
@@ -1450,6 +1450,10 @@ crate enum Type {
     ImplTrait(Vec<GenericBound>),
 }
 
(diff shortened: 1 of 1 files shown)


jyn514:
Looks like no change in compiletime and a moderate regression for memory on tokio-webpush-simple-doc. So I think this should be closed unless you have ideas for improving it.

camelid:
Actually, there are a lot of small improvements:

image

jyn514:
Those are *very* small, though ... I don't know if the improvement is worth it.

camelid:
Well, up to you. I think we should at least keep the static size assertion for `Type`, so I can drop the boxing commit.

jyn514:
> I think we should at least keep the static size assertion for Type, so I can drop the boxing commit

👍 Seems good, r=me with that done

PR #89999, opened 2021-10-18
Update std::env::temp_dir to use GetTempPath2 on Windows when available.

talagrand:
As a security measure, Windows 11 introduces a new temporary directory API, GetTempPath2.
When the calling process is running as SYSTEM, a separate temporary directory
will be returned inaccessible to non-SYSTEM processes. For non-SYSTEM processes
the behavior will be the same as before.

This can help mitigate against attacks such as this one:
https://medium.com/csis-techblog/cve-2020-1088-yet-another-arbitrary-delete-eop-a00b97d8c3e2

Compatibility risk: Software which relies on temporary files to communicate between SYSTEM and non-SYSTEM
processes may be affected by this change. In many cases, such patterns may be vulnerable to the very
attacks the new API was introduced to harden against. 
I'm unclear on the Rust project's tolerance for such change-of-behavior in the standard library. If anything,
this PR is meant to raise awareness of the issue and hopefully start the conversation.

How tested: Taking the example code from the documentation and running it through psexec (from SysInternals) on
Win10 and Win11.  
On Win10:
C:\test>psexec -s C:\test\main.exe
<...>
Temporary directory: C:\WINDOWS\TEMP\

On Win11: 
C:\test>psexec -s C:\test\main.exe
<...>
Temporary directory: C:\Windows\SystemTemp\

### DIFF
--- a/library/std/src/env.rs
+++ b/library/std/src/env.rs
@@ -578,28 +578,25 @@ pub fn home_dir() -> Option<PathBuf> {
 /// may result in "insecure temporary file" security vulnerabilities. Consider
 /// using a crate that securely creates temporary files or directories.
 ///
-/// # Unix
+/// # Platform-specific behavior
 ///
-/// Returns the value of the `TMPDIR` environment variable if it is
+/// On Unix, returns the value of the `TMPDIR` environment variable if it is
 /// set, otherwise for non-Android it returns `/tmp`. If Android, since there
 /// is no global temporary folder (it is usually allocated per-app), it returns
 /// `/data/local/tmp`.
+/// On Windows, the behavior is equivalent to that of [`GetTempPath2`][GetTempPath2] /
+/// [`GetTempPath`][GetTempPath], which this function uses internally.
+/// Note that, this [may change in the future][changes].
 ///
-/// # Windows
-///
-/// Returns the value of, in order, the `TMP`, `TEMP`,
-/// `USERPROFILE` environment variable if any are set and not the empty
-/// string. Otherwise, `temp_dir` returns the path of the Windows directory.
-/// This behavior is identical to that of [`GetTempPath`][msdn], which this
-/// function uses internally.
-///
-/// [msdn]: https://docs.microsoft.com/en-us/windows/win32/api/fileapi/nf-fileapi-gettemppatha
+/// [changes]: io#platform-specific-behavior
+/// [GetTempPath2]: https://docs.microsoft.com/en-us/windows/win32/api/fileapi/nf-fileapi-gettemppath2a
+/// [GetTempPath]: https://docs.microsoft.com/en-us/windows/win32/api/fileapi/nf-fileapi-gettemppatha
 ///
 /// ```no_run
 /// use std::env;
 ///
 /// fn main() {
-///     let mut dir = env::temp_dir();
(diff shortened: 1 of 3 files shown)


BurntSushi:
I think this should probably require an FCP. In particular, this is changing the documented behavior of a routine. I suppose it's questionable that we listed the precise platform call we're using in the first place. But if we don't consider changing it to be breaking the contract, then we should probably either remove it or make the wording more relaxed.

One concern I have here is the difference in behavior between Windows versions. I don't really know how to measure it.

Another thought here is: why didn't Windows just change the behavior of of `GetTempPath` instead of adding a new `GetTempPath2`? A guess would be that they were worried enough about compatibility to do it. Should we be similarly worried?

Finally, what are other language ecosystems doing here, if anything?

We should also ping the Windows folks. (I forget how.)

wesleywiser:
This seems like a reasonable change to me as well. I also agree with the concerns regarding the current documentation. I think a note similar to the one @ChrisDenton referred to in the `std::io` docs would be appropriate here.

talagrand:
Updated documentation as suggested - I consolidated the platform-specific behavior notice centrally for std. The previous commit simply adds a disclaimer to the "env" module if that's deemed preferrable.
@BurntSushi unclear how to initiate an FCP, do we have a documented process surrounding that?

talagrand:
Updated documentation for temp_dir to fall more in line with the pattern established by std::io.

yaahc:
> Another thought here is: why didn't Windows just change the behavior of of `GetTempPath` instead of adding a new `GetTempPath2`? A guess would be that they were worried enough about compatibility to do it. Should we be similarly worried?

I want to repeat @BurntSushi's question here on why didn't Windows do this as a breaking change? If they did it because they're worried that makes me wonder if we should be mirroring their API choice and instead introduce a new method + deprecate the old one.

talagrand:
I'll reiterate Sivadeilra's feedback here on the importance in Windows to not change the externally visible behavior of Win32 APIs release-over-release to maximize app compatibility due to the size and age of the ecosystem.

Additionally, supposing Windows had changed the behavior of GetTempPath, there would be limited ability for apps to restore the old behavior if absolutely necessary, since it's the lowest-level API.
As for what language runtimes should do, I think this dips into the cultural expectations of what that runtime is providing - is it an abstract platform, implemented in terms of what the underlying system provides, or is it a direct binding of underlying OS primitives? Personally, I would expect a C++ std::thread to bind directly to an OS thread, but the story around a java.lang.Thread to be more nuanced.

In the case of a Rust, I'd imagine the standard library falls somewhere along this abstraction gradient. I believe that, despite std::fs::File::create existing, there is still a role to play for winapi::um::fileapi::CreateFileW existing for developers who want the precise semantics of the underlying OS. In a similar vein, developers who want the precise OS behavior here could always call into winapi::um::fileapi::GetTempPathW to get that behavior. But ideally, we wouldn't want folks to have to reach into winapi to get safe semantics.

My primary motivation for proposing this change is to expose safe defaults, and establish safety even when higher-order abstractions are built on top. For example, I'm currently using the "tempfile" crate, and this change makes that more secure.

wesleywiser:
Seems spurious?

@bors retry

PR #90001, opened 2021-10-18
Make rlib metadata strip works with MIPSr6 architecture

Fearyncess:
Because MIPSr6 has many differences with previous MIPSr2 arch, the previous rlib metadata stripping code in `rustc_codegen_ssa` is only for MIPSr2/r3/r5 (which share the same elf e_flags).

This commit fixed this problem. It makes `rustc_codegen_ssa` happy when compiling rustc for MIPSr6 target or hosts.

e_flags REF: https://github.com/llvm/llvm-project/blob/e356027016c6365b3d8924f54c33e2c63d931492/llvm/include/llvm/BinaryFormat/ELF.h#L562

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -25,6 +25,17 @@ dependencies = [
 ]
 
 [[package]]
+name = "ahash"
+version = "0.7.4"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "43bb833f0bf979d8475d38fbf09ed3b8a55e1885fe93ad3f93239fc6a4f17b98"
+dependencies = [
+ "getrandom 0.2.0",
+ "once_cell",
+ "version_check",
+]
+
+[[package]]
 name = "aho-corasick"
 version = "0.7.18"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -1549,6 +1560,7 @@ version = "0.11.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "362385356d610bd1e5a408ddf8d022041774b683f345a1d2cfcb4f60f8ae2db5"
 dependencies = [
+ "ahash",
  "compiler_builtins",
  "rustc-std-workspace-alloc",
  "rustc-std-workspace-core",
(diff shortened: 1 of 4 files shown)


chenx97:
9859b818c88ceae3c4c3620b06e27d07d84576f8 is the initial support for version 6 of MIPS. @wzssyqa Could you help review this pull request?

> It looks like the only addition here is `EF_MIPS_NAN2008`, so could the code be refactored to not have duplicate flags on each branch and instead only add that flag as necessary? Otherwise seems fine to me, but if you're up for it it'd probably be good to update the `object` crate dependency from other crates to ideally only use one version instead of introducing another.

@alexcrichton I would love to help factor out the duplication. Should I initialize a mutable `e_flags` with common flags and then add new ones based on the r6 detection?

Also, regarding having NaN2008 on MIPSr5, I think this should be out of the scope of this pull request. This pull request is meant to only deal with mipsisa{32,64}r6el-related targets. What we need to make sure here is that this pull request won't change anything for those Tier-2 MIPS targets, targeting chips from MIPSr2 to MIPSr5.

chenx97:
@alexcrichton Sorry for suddenly jumping into the conversation. I'm a colleague of the author of this pull request, so I asked the questions on behalf of our team. We are at CIP United and we will continue to improve Rust's support for the MIPSr6 family, and hopefully make it tier-2 in the future (which also means making real MIPSr6 hardware, and we look forward to it).😊

alexcrichton:
My review of this PR is basically purely stylistic, not really technical. The MIPS flags are all duplicated where the presence of the `r6` feature implies `EF_MIPS_NAN2008` is or-ed into the list of flags. I would stylistically request less duplication, along the lines of what https://github.com/rust-lang/rust/pull/90001#issuecomment-976203260 mentions.

Fearyncess:
> My review of this PR is basically purely stylistic, not really technical. The MIPS flags are all duplicated where the presence of the `r6` feature implies `EF_MIPS_NAN2008` is or-ed into the list of flags. I would stylistically request less duplication, along the lines of what [#90001 (comment)](https://github.com/rust-lang/rust/pull/90001#issuecomment-976203260) mentions.

i have make the code more simpilified and update the dep of `rustc_codegen_ssa` for it. it seems works well

Issue #52544, opened 2018-07-19
Consider borrowing here: `&0`


jjl:
```
error[E0308]: mismatched types
  --> src/schema_parser.rs:91:20
   |
91 |             if i < 0 {
   |                    ^
   |                    |
   |                    expected &i64, found integral variable
   |                    help: consider borrowing here: `&0`
   |
   = note: expected type `&i64`
              found type `{integer}`
```

What I actually wanted was `*i`.

oli-obk:
The problem is by the time this error is emittedwe don't have the information about the root cause anymore. In this case that would be the type of `i` being checked for `PartialOrd` and figuring out the the rhs needs to be of the same type.

Clippy has a lint that will tell you to clean it up afterwards, but we should probably issue a better diagnostic to begin with.

hellow554:
@jjl, can you please create a MCVE?

csmoe:
@hellow554
```rust
fn main() {
    let i: &i64 = &1;
    if x < 0 {}
}
```

hellow554:
Please excuse my naïvety, but why is this (according to csmoes MCVE) a problem? You can actually write `i < &0` and everything works just fine. Can you please explain this part (to me)?

oli-obk:
The comparison operator already takes references to both sides. so it seems simpler to dereference the lhs and save the monomorphizer and the optimizer some effort

rakshith-ravi:
Resurrecting this. 

@oli-obk what're the benefits of dereferencing the lhs instead of referencing the rhs?

Also, if I were to go about fixing this, where would I start?

oli-obk:
The benefit is that `x < y` expands to `Ord::cmp(&x, &y)`, so if you start out with `x < &y` you get `Ord::cmp(&x, &&y)`, seems better to just do `Ord::cmp(&*x, y)` instead (since `&*` is a no-op that gets easily removed).

I'm not sure how to solve this or whether we should try at all.

rakshith-ravi:
I want to get started with this, but would appreciate some mentoring from someone before I do, to point me in the right direction please

oli-obk:
There's a reason I marked this as `E-hard`. I have no idea how to do this reasonably. I'm sorry, but I have no hints for you, and I don't even know if it's possible to do in a reasonable way.

rakshith-ravi:
If someone wants to claim this up, feel free to. Until then, I'll try and fix this.

Alexendoo:
[...]

Issue #40660, opened 2017-03-20
Better error messages for &T == T


iqualfragile:
Something along the lines of "you are trying to compare a refference to <Type> to Type, try dereferencing first" would be nice, right now it just has the usual PartialEq not implemented.

estebank:
Would the following output (as proposed in #40565) be enough to make the situation a bit more clear?

```rust
error[E0277]: the trait bound `&T: std::cmp::PartialEq<T>>` is not satisfied
  --> $DIR/file.rs:14:5
   |
14 |     &t == t;
   |     ^^^^^^^ can't compare `&T` with `T`
```

iqualfragile:
that would certainly improve things, but in this case the error message could be even more specific and thereby helpful by providing the information about deref-ing, so i would keep it as a separate issue, i think

estebank:
[Current output for](https://play.rust-lang.org/?gist=257e75bb296aa347712e230b33991b95&version=nightly):

```rust
fn main() {
    let x: usize = 3;
    let y: &usize = &3;
    if x == y {
        println!("asdf");
    }
    if y == x {
        println!("asdf");
    }
    if 3 == &3 {
        println!("asdf");
    }
    if &3 == 3 {
        println!("asdf");
    }
}
```

```
error[E0308]: mismatched types
 --> src/main.rs:4:13
  |
4 |     if x == y {
  |             ^
  |             |
  |             expected usize, found &usize
  |             help: consider dereferencing the borrow: `*y`
  |
  = note: expected type `usize`
             found type `&usize`

error[E0277]: the trait bound `&usize: std::cmp::PartialEq<usize>` is not satisfied
 --> src/main.rs:7:10
  |
7 |     if y == x {
  |          ^^ can't compare `&usize` with `usize`
  |
  = help: the trait `std::cmp::PartialEq<usize>` is not implemented for `&usize`

error[E0277]: the trait bound `{integer}: std::cmp::PartialEq<&{integer}>` is not satisfied
  --> src/main.rs:10:10
   |
10 |     if 3 == &3 {
   |          ^^ can't compare `{integer}` with `&{integer}`
   |
   = help: the trait `std::cmp::PartialEq<&{integer}>` is not implemented for `{integer}`

error[E0277]: the trait bound `&{integer}: std::cmp::PartialEq<{integer}>` is not satisfied
  --> src/main.rs:13:11
   |
13 |     if &3 == 3 {
[...]

PR #90006, opened 2021-10-18
More accurate error for binop errors after identifying RHS type

### LINKED ISSUE #44695: Ergonomics: &String does not implement PartialEq<str>
Start with this:

```rust
fn cmp_prefix() -> bool {
    let owned = "foo".to_owned();
    let string_ref = &owned;
    let partial = "foobar";
    string_ref == partial[..3]
}
```

It doesn't compile: "the trait bound `&String: PartialEq<str>` is not satisfied".

Let's [Google this](https://www.google.nl/search?q=rust+compare+String+with+str), first hit is [StackOverflow](https://stackoverflow.com/questions/25383488/how-to-match-a-string-against-string-literals-in-rust), first/accepted answer talks about `.as_ref()`, try that. `string_ref.as_ref() == &partial[..3]` results in "the trait bound `&_: PartialEq<str>` is not satisfied" [oh, how I hate these underscores in errors].

At this point I try adding `*` to the LHS or `&` to the RHS. The latter results in "type annotations required: cannot resolve `String: AsRef<_>`" [huh?].

Here's actually everything I tried (from my actual code instead of simple example):

```rust
meta_subj == db_subj[..998]
&meta_subj == &db_subj[..998]
meta_subj.as_ref() == &db_subj[..998] // I think this was before Googling, even
meta_subj.as_ref() == db_subj[..998]
meta_subj.as_ref::<&str>() == &db_subj[..998] // WTF is this thing with not resolving String: AsRef?
&meta_subj.as_ref() == db_subj[..998]
*meta_subj.as_ref() == db_subj[..998]
```

At this point I'm pretty much out of ideas other than brute-forcing with some `to_string()`. I don't really consider myself a beginner at this point, having successfully written a few thousand lines of code over the past 15 months, so here's where I have some empathy for those saying Rust's learning curve is just too steep.

I went and read some more on SO and found this, which actually seems to work `(meta_subj as &str) == &db_subj[..998]`. The reason I would not have come up with this is that generally the compiler seems to be able to coerce `&T where T: String` to `&str`, so I find it hard to come up with a reason it cannot do so in this case.

estebank:
Evaluate binops a second time with the resolved RHS type so we can provide suggestions for errors caused by trait predicates that didn't fail when looking for the binop with a type variable in the RHS.

Provide more casting and borrowing suggestions in binops.

Fix #44695, partially address #40660, partially address #57621 (we provide bad suggestion spans when `assert_eq!` is involved), doesn't account for #52544.

Zulip thread: https://rust-lang.zulipchat.com/#narrow/stream/247081-t-compiler.2Fperformance/topic/Up.20to.20.255.20regression.20on.20kekkak.20with.20.2390006/near/258044068


estebank:
@bors try @rust-timer queue

This is doing more work on the happy path (evaluating the binop twice, with a ty var and with a concrete ty). Let's see if I need to rework this for perf's sake.

JohnCSimon:
Ping from triage: 
@estebank can you please post the status of this PR?

JohnCSimon:
Ping for triage: 
@estebank FYI this PR has sat idle for months.

Dylan-DPC:
I am closing this as it is bitrotted.

PR #90020, opened 2021-10-18
Enable `windows-gnu` target to link to `libgcc` dynamically.

Berrysoft:
Closes #89919.

It seems that the change is trival, and I've tested on my own computer with success, with both `crt-static` is enabled or disabled.

The `libunwind` part is not tested, but it should work in theory, according to the [patch](https://github.com/msys2/MINGW-packages/blob/master/mingw-w64-rust/0007-clang-subsystem.patch) of MSYS2.

There's one more problem needs to talk about: should we enable `crt-static` as default when using GNU toolchain on Windows?


Amanieu:
I think this breaks the case where `-C prefer-dynamic` is used since we must always use the shared libgcc_s.so in this case.

petrochenkov:
>you cannot specify `+crt-static` together with `prefer-dynamic`

Technically you can, the compiler won't report an error, it just doesn't make sense on most targets.
I'm not sure what result this combination will give in practice if used, need to check the code.

Berrysoft:
> Technically you can, the compiler won't report an error...

No. The compiler will complain about not finding `_Unwind_***` symbols when linking.

> it just doesn't make sense on most targets.

Yes, at least on this target, it doesn't make sense at all. MinGW won't statically link with `msvcrt` because of license restrictions, and those who want to use `crt-static` must be those who want to link with `libgcc` statically. Even though if the compiler doesn't report any error, they will find that their output binary will still depend on `libgcc` dynamically because of `libstd`. They will then understand that their attempt is wrong:)

Berrysoft:
> This will break exceptions across DLLs on 32-bit target which is also has a test IIRC.

Which test? Do you mean it will break because it enables `prefer-dynamic`? We should specify `-crt-static` in that case. As I commented earlier, `crt-static` and `prefer-dynamic` are incompatible features.

mati865:
> > This will break exceptions across DLLs on 32-bit target which is also has a test IIRC.
> 
> Which test? Do you mean it will break because it enables `prefer-dynamic`? We should specify `-crt-static` in that case. As I commented earlier, `crt-static` and `prefer-dynamic` are incompatible features.

I don't remember which one, you would have to run them.
It would fail for two reasons: you are not linking pthread for static case and broken exceptions across DLLs. I have no idea if it used `prefer-dynamic` or not.

Berrysoft:
> That issue itself needs another target triple anyway.

No, it needn't. `system-llvm-libunwind` is already a feature to control whether `libgcc` or `libunwind` should be used. It doesn't need another new target to distinguish gcc toolchain with llvm toolchain.

> It would fail for two reasons: you are not linking pthread for static case and broken exceptions across DLLs.

Oh, OK. I finally know why there was `-l:libpthread.a` in the static link args. I'll add them.

For the exceptions across DLLs, I'll investigate further.

Berrysoft:
@mati865 I ran the test against target `i686-pc-windows-gnu` and stage0 compiler, but didn't failed any test. Maybe I should test for stage1 compiler? Or have you recalled the test name?

Berrysoft:
I've tested on Linux that when `-C prefer-dynamic` and `-C target-feature=+crt-static` are both specified, the compiler will choose to statically link to C runtime when building executable, and will report error when building dylibs. Therefore, the priority of `crt-static` is higher than `prefer-dynamic`.

petrochenkov:
@Berrysoft 
Could you check what happens if you set `+crt-static` (explicitly or through a default) for an executable, and link that executable to a dylib (the standard library, for example) built with `-crt-static`.

In theory if we are linking to any dylib, then we should switch the executable to `-crt-static` too (that's basically what `any_dynamic_crate` does now), but if executables are `+crt-static` by default, then such a switch will need to be performed manually by the user.

petrochenkov:
I'm basically ok with switching the windows-gnu targets to `+crt-static` by default, given that it's what we are de-facto doing in typical cases (no dynamic crate deps -> libgcc and libpthread are linked statically).

The only case that worries me is the previous comment https://github.com/rust-lang/rust/pull/90020#issuecomment-968547132 - an executable with dynamic crate deps (e.g. dynamic libstd).
In that case either we should switch to `-crt-static` automatically (at least by default), or require a user to do it explicitly.

Perhaps we just need more fine-grained options related to `crt-static` in target specs instead of two `bool`s for executables and dynamic libraries.
This is some generalization task over all our supported targets rather than something windows-gnu-specific, I need to think about it a bit.

petrochenkov:
I went through major targets supporting dynamic linking and they all seem to expose relatively similar behavior with regards to linking CRT (libc or analogues) dynamically or statically.
Here's some summary.

An executable and all its dynamic dependencies must share a common CRT for things like cross-binary exceptions and cross-binary allocations/deallocations on platform like Windows (which don't do most of OS interactions through CRT), and maybe for other shared resource management as well on platforms like Linux (which do most of OS interactions through CRT).
Next I'll consider different kinds of end binaries and their relationships with dynamic and static CRT.

- Executable without other dynamic dependencies - static libc is possible and not exotic.
- Executable with other dynamic dependencies - depends on the dependencies:
	- The dependencies don't produce cross-binary exceptions or allocations (on Windows), and don't link to CRT (on Linux) - static libc is possible and not exotic.
		- Example: the executable only depends on WinAPI DLLs.
	- The dependencies may produce cross-binary exceptions or allocations (on Windows), or link to CRT (on Linux) - static libc is possible, but exotic and shouldn't be done without an explicit request.
		- Example: the executable depends on Rust standard library as a dylib.
- Dynamic library with or without other dynamic dependencies - depends on itself and dependencies:
	 - Neither the library itself nor its dependencies may produce cross-binary exceptions or allocations (on Windows), and the dependencies don't link to CRT (on Linux) - static libc is possible and not exotic.
		- Example: the library is a cdylib with C interface and only depends on WinAPI DLLs.
		- Example: the library is a cdylib which is a plugin for [DynamoRIO](https://github.com/dynamorio/dynamorio) binary instrumentation framework. It's an ELF, but it will be loaded by `libdynamorio.so` instead of a standard dynamic linker.
	 - The library itself or its dependencies may produce cross-binary exceptions or allocations (on Windows), or the dependencies link to CRT (on Linux) - static libc is possible, but exotic and shouldn't be done without an explicit request.
		- Example: the library is a Rust dylib, or depends on Rust standard library as a dylib.
- Linking CRT dynamically is always possible and not exotic in all cases above.

Whether a dynamic library may require managing shared resources is ultimately a user's choice.
We can easily take "yes, it requires managing shared resources, and thus requires shared CRT" as a default for both Rust dylibs and native dylibs, and add a linking modifier for opting out, `-needs-shareed-crt` or something.

Side note: for targets doing dynamic linking in userspace via ELF interpreter field (Linux), we can remove the interpreter field and speedup their launch if there's no dynamic dependencies at all (CRT or others).

So the question is what "with(out) an explicit request" above means.
In general, I'd say `-Ctarget-feature=-crt-static` is explicit enough because dynamic linking is never exotic, but `-Ctarget-feature=+crt-static` is not explicit enough.

Considering the above, I think my suggested solution is:
- Keep the `crt_static_respected` target spec option, target with `crt_static_respected=false` behave in some hard-coded way, targets with `crt_static_respected=true` can be controlled using command-line options.
- Keep the `-Ctarget-feature=+/-crt-static` option and the `crt_static_default` target spec option. Take `crt-static` as `value(-Ctarget-feature=+/-crt-static).unwrap_or(crt_static_default)`.
- If `crt-static` is false, then we link CRT dynamically.
- If `crt-static` is true, then we link CRT statically, if the situation is not exotic, otherwise we link CRT dynamically and maybe emit a warning telling the reason.
- The situation is exotic when
	- This is an executable and it has dynamic dependencies requiring managing shared resources.
	- This is a dynamic library and it has dynamic dependencies requiring managing shared resources.
	- This is a dynamic library and it requires managing shared resources.
	- For dependencies "requires managing shared resources" means that
		- It's a Rust dylib or proc macro crate.
		- It's a native dylib not marked with the `-needs-shareed-crt` link modifier.
	- For the current dynamic library "requires managing shared resources" means that
		- It's a Rust dylib or proc macro crate
			- Cdylibs are not considered requiring shared resources by default, if your cdylib requires them, just use `-Ctarget-feature=-crt-static`.
				- Or maybe we can have a target spec option controling whether cdylibs (the current one and native dependencies) require shared resources by default, e.g. it would be false on Windows and true on Linux.
- Add a new option `-Cforce-crt-static`.
	- It removes the "non-exotic" requirement from the conditions above and makes `crt-static` enable static CRT linking even in exotic cases.
	- Example: if `+needs-shareed-crt` is the default on Windows, then `-Cforce-crt-static` may be required to build statically while depending on WinAPI DLLs, until the winapi crate adopts `-needs-shared-crt` modifiers.
- Remove the `crt_static_allows_dylibs` target spec option, it's no longer necessary with our scheme.
- Issue: evaluating effective `crt-static` requires determining whether our situation is exotic, which requires evaluating some link-time `cfg`s, which requires evaluating effective `crt-static`.
	- Fortnately, it's boolean so there are only two values and we can likely try them both speculatively.

petrochenkov:
What this means for windows gnu - it'll get `crt_static_respected=true`, `crt_static_default=true`, and probably `cdylibs_need_shared_crt_by_default=false` as well because it's still Windows, where cross-binary exceptions and allocations/deallocations are discouraged, and other shared resources are managed through OS/WinAPI instead of CRT.

petrochenkov:
@Berrysoft 
To clarify, I don't suggest you to implement the design above, it'll likely need an MCP anyway.
Not sure what to do with this PR short term though.

Berrysoft:
@petrochenkov
Hmm, you've given such a detailed suggestion, and I'm sorry about responsing late.

Maybe you're right, and we need to rethink about all these problems.

mati865:
> * Example: if `+needs-shareed-crt` is the default on Windows, then `-Cforce-crt-static` may be required to build statically while depending on WinAPI DLLs

What about MSVC where CRT can be statically linked?

petrochenkov:
@mati865 
That line is mostly about MSVC actually.
Majority of the Windows mentions and examples above have MSVC in mind.

mati865:
Well I find it confusing because MSVC is the only Windows target right now that allows statically linking CRT into executables. So judging by the name I'd expect if I pass `-Cforce-crt-static` then CRT would be really static.

petrochenkov:
@mati865 
By CRT for windows-gnu I mean the same thing as Berrysoft in this PR - libgcc and libpthread (but not e.g. msvcrt or kernel32), https://github.com/rust-lang/rust/pull/90020#issuecomment-948455023.
I.e. the components for which the platform gives this choice at all.

The design in https://github.com/rust-lang/rust/pull/90020#issuecomment-985215220 still works in this case - shared libgcc/libpthread are indeed necessary for managing shared resources (cross-binary exceptions).

Dylan-DPC:
@Berrysoft any updates on this?

JohnCSimon:
@Berrysoft 

@rustbot label: +S-inactive

PR #90022, opened 2021-10-18
Explain why `Self` is invalid in generic parameters

### LINKED ISSUE #89985: Add note on attempts of using `Self` as a normal generic arg
Given the following code: https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=53b02445623677643befa3e7dbe0f3b4

```rust
fn foo<Self>() {}
```

The current output is:

```
error: expected identifier, found keyword `Self`
 --> src/lib.rs:1:8
  |
1 | fn foo<Self>() {}
  |        ^^^^ expected identifier, found keyword
```
The diagnostic itself is right but confusing; it lacks why `Self` is a reserved keyword (IIUC it's because `Self` is special-cased to represent traits or ADTs in its implementations).

hkmatsumoto:
Close #89985.

r? @estebank


jackh726:
@bors r+

This seems reasonable.

Issue #86618, opened 2021-06-25
 Implementation Issue for RFC 2528: type-changing struct update syntax #86555 


nikomatsakis:
This issue tracks the implementation of rust-lang/rfcs#2528, "type changing struct update syntax" (general tracking issue: https://github.com/rust-lang/rust/issues/86555).

nikomatsakis:
@rustbot assign @kolharsam 

They expressed an interest in this!

nikomatsakis:
Here are some mentoring notes.

## Step 0: Create the feature gate

We need a feature gate to control when users are accessing this feature. Instructions for adding a feature gate are here:

https://rustc-dev-guide.rust-lang.org/feature-gates.html#adding-a-feature-gate

**Adding a feature gate can and should be its own PR!**

nikomatsakis:
Oh, one thing I forgot to mention! We are going to want to make these changes *conditional* on whether the feature gate is enabled.

kolharsam:
> Here are some mentoring notes.
> 
> ## PR 0: Create the feature gate
> We need a feature gate to control when users are accessing this feature. Instructions for adding a feature gate are here:
> 
> https://rustc-dev-guide.rust-lang.org/feature-gates.html#adding-a-feature-gate
> 
> **Adding a feature gate can and should be its own PR!**

@nikomatsakis Have made a respective PR for the same here: https://github.com/rust-lang/rust/pull/86646

kolharsam:
> ## PR 1: Write up some tests
> A good next step is to open a PR creating a directory in `src/tests/ui` for this feature and adding the various tests. It's ok if they get errors today! Just document the current behavior. But I've always found it's actually much easier to write tests BEFORE you start the code then after!

@nikomatsakis I've made a PR with a failing test here: #86660. 
Currently, it just document's the error that might be seen while using the suggested updates to the syntax in within the RFC. 

Please let me know if there are any other test cases that need to be added because I wasn't really sure about what needed to be done here.

nikomatsakis:
I don't believe we've implemented this feature yet =) there are mentoring notes [here](https://github.com/rust-lang/rust/issues/86618#issuecomment-868976919). I know that someone was looking into it, maybe it was @kolharsam? (If I failed to get back to you, @kolharsam, sorry, are you still interested?)

kolharsam:
@nikomatsakis Yes, I was looking into it, and I'm still interested in continuing the work.

Kobzol:
[...]

PR #90035, opened 2021-10-19
implement rfc-2528 type_changing-struct-update

SparrowLii:
This PR implement rfc2528-type_changing-struct-update.
The main change process is as follows:
1. Move the processing part of `base_expr` into `check_expr_struct_fields` to avoid returning `remaining_fields` (a relatively complex hash table)
2. Before performing the type consistency check(`check_expr_has_type_or_error`), if the `type_changing_struct_update` feature is set, enter a different processing flow, otherwise keep the original flow
3. In the case of the same structure definition, check each field in `remaining_fields`. If the field in `base_expr` is not the suptype of the field in `adt_ty`, an error(`FeildMisMatch`) will be reported.

The MIR part does not need to be changed, because only the items contained in `remaining_fields` will be extracted from `base_expr` when MIR is generated. This means that fields with different types in `base_expr` will not be used
Updates #86618
cc @nikomatsakis

### DIFF
--- a/compiler/rustc_typeck/src/check/expr.rs
+++ b/compiler/rustc_typeck/src/check/expr.rs
@@ -23,7 +23,7 @@ use crate::type_error_struct;
 
 use crate::errors::{AddressOfTemporaryTaken, ReturnStmtOutsideOfFnBody, StructExprNonExhaustive};
 use rustc_ast as ast;
-use rustc_data_structures::fx::FxHashMap;
+use rustc_data_structures::fx::{FxHashMap, FxHashSet};
 use rustc_data_structures::stack::ensure_sufficient_stack;
 use rustc_errors::ErrorReported;
 use rustc_errors::{pluralize, struct_span_err, Applicability, DiagnosticBuilder, DiagnosticId};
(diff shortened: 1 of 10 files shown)


jackh726 (changes requested):
Overall pretty straightforward, a few nits. Can you squash the commits (other than spliting out that one commit)?

compiler/rustc_typeck/src/check/expr.rs
```diff
@@ -1374,78 +1375,124 @@ impl<'a, 'tcx> FnCtxt<'a, 'tcx> {
...
-                        return self
-                            .tcx
-                            .sess
-                            .emit_err(FunctionalRecordUpdateOnNonStruct { span: base_expr.span });
-                    }
-                }
-            };
-            self.typeck_results.borrow_mut().fru_field_types_mut().insert(expr_id, fru_tys);
+                err = self.demand_suptype_diag(base_expr.span, expected_ty, ty);
+                is_struct = false;
```
jackh726, line 1482: Why not just emit the error here?

Or, better yet, any way we can just return early and do away with the nesting?

jackh726 (changes requested):
Last commit feels a bit messy to me; not sure how best to clean it up though.

Left a couple commits.

SparrowLii:
I manually copied most of the implementation of check_expr_has_type_or_error, which would make the code messy, which is obviously unwise. So I went back to the previous submission and found a new way. It should be clear enough now and can be reviewed.

PR #90036, opened 2021-10-19
Remove border-bottom from headings in most docblocks.

### LINKED ISSUE #90033: remove horizontal rule from Markdown-generated HTML
When documentation has markdown headings (e.g. `# Foo`, `## Bar`, etc), we currently display those with a horizontal rule (`<hr>`) across the whole page. However, that makes sub-sub-headings, like `Panics` in the below screenshot of https://docs.rs/rctree/0.4.0/rctree/struct.Node.html, look more prominent than their parents in the document structure. They also make the page look grouped along lines that it's not intended to be grouped along.

image

We should remove or hide these elements to improve the scannability of pages. Most likely we want to keep them when they are part of the docblock at the very top of the page, but hide them elsewhere.

jsha:
Headings in the top-doc docblock still get a border-bottom due to a rule
that covers all h2, h3, and h4. Method docblocks are generally h5, and
so don't get a border-bottom anymore.

This fixes a problem where a sub-sub-heading within a method would have
a line that went all the way across the page, creating a division that
made that sub-sub-heading look much more important than it really is.

Fixes #90033 

Demo at https://jacob.hoffman-andrews.com/rust/less-rule/std/string/struct.String.html

r? @GuillaumeGomez

### DIFF
--- a/src/librustdoc/html/static/css/rustdoc.css
+++ b/src/librustdoc/html/static/css/rustdoc.css
@@ -502,10 +502,6 @@ nav.sub {
 	white-space: pre-wrap;
 }
 
-.docblock h1, .docblock h2, .docblock h3, .docblock h4, .docblock h5, .docblock h6 {
-	border-bottom: 1px solid;
-}
-
 .top-doc .docblock h2 { font-size: 1.3em; }
 .top-doc .docblock h3 { font-size: 1.15em; }
 .top-doc .docblock h4,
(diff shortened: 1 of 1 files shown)


jsha:
I pushed an update with the requested changes, and updated the demo. Let me know what you think!

GuillaumeGomez (approved):

camelid:
> I did: https://github.com/rust-lang/rust/pull/90036/files#diff-67c3587d759aa9578993149e68eb8d3a9de6b28e5c0b85091b0a71a07f441795R514
> 
> Previously h4, h5, and h6 were font-size: 1em. Now h4 and h5 are 1.1em, while h6 is still 1em.

Ah, maybe you didn't update your demo with this part then? Firefox devtools show that h5 is still 1em.

PR #79684, opened 2020-12-04
Make copy[_nonoverlapping] const

usbalbin:
Constifies
* `intrinsics::copy` and `intrinsics::copy_nonoverlapping`
* `ptr::read` and `ptr::read_unaligned`
  * `*const T::read` and `*const T::read_unaligned` 
  * `*mut T::read` and `*mut T::read_unaligned` 
* `MaybeUninit::assume_init_read`

PR #89488, opened 2021-10-03
Move some tests to more reasonable directories - 8

c410-f3r:
cc #73494
r? @petrochenkov

PR #90041, opened 2021-10-19
Re-enable `copy[_nonoverlapping]()` debug-checks

### LINKED ISSUE #90012: Re-enable debug checks in `copy[_nonoverlapping]`
#79684 removed `debug_assert!` from `intrinsic::copy[_nonoverlapping]` to make it `const`:
https://github.com/rust-lang/rust/blob/5dab47dcd8267b8769421b46532414ec36d625e3/library/core/src/intrinsics.rs#L2045-L2049
https://github.com/rust-lang/rust/blob/5dab47dcd8267b8769421b46532414ec36d625e3/library/core/src/intrinsics.rs#L2130-L2131
We can't do these checks at compile-time, since `is_aligned_and_not_null`, for example, involves ptr->int cast to check the alignment:
https://github.com/rust-lang/rust/blob/5dab47dcd8267b8769421b46532414ec36d625e3/library/core/src/intrinsics.rs#L1950-L1952

So, as the FIXME suggests, we should enable the checks only at runtime. Recently `const_eval_select` intrinsic [was implemented](https://github.com/rust-lang/rust/pull/89247), it allows for exactly this use case - running different code in CTFE and runtime.

cc @rust-lang/lang, @rust-lang/libs and @rust-lang/wg-const-eval (it seems like use of `const_eval_select` requires approval of all of the above teams)

@rustbot label +T-lang +T-libs +A-const-eval +A-const-fn

jfrimmel:
This commit re-enables the debug checks for valid usages of the two functions `copy()` and `copy_nonoverlapping()`. Those checks were commented out in #79684 in order to make the functions const. All that's been left was a FIXME, that could not be resolved until there is was way to only do the checks at runtime.
Since #89247 there is such a way: `const_eval_select()`. This commit uses that new intrinsic in order to either do nothing (at compile time) or to do the old checks (at runtime).

The change itself is rather small: in order to make the checks usable with `const_eval_select`, they are moved into a local function (one for `copy` and one for `copy_nonoverlapping` to keep symmetry).

The change does not break referential transparency, as there is nothing you can do at compile time, which you cannot do on runtime without getting undefined behavior. The CTFE-engine won't allow missuses. The other way round is also fine.

I've refactored the code to use `#[cfg(debug_assertions)]` on the new items. If that is not desired, the second commit can be dropped.
I haven't added any checks, as I currently don't know, how to test this properly.

Closes #90012.

cc @rust-lang/lang, @rust-lang/libs and @rust-lang/wg-const-eval (as those teams are linked in the issue above).

### DIFF
--- a/library/core/src/intrinsics.rs
+++ b/library/core/src/intrinsics.rs
@@ -1951,6 +1951,19 @@ pub(crate) fn is_aligned_and_not_null<T>(ptr: *const T) -> bool {
     !ptr.is_null() && ptr as usize % mem::align_of::<T>() == 0
 }
 
+/// Checks whether the regions of memory starting at `src` and `dst` of size
+/// `count * size_of::<T>()` do *not* overlap.
+#[cfg(debug_assertions)]
+pub(crate) fn is_nonoverlapping<T>(src: *const T, dst: *const T, count: usize) -> bool {
+    let src_usize = src as usize;
+    let dst_usize = dst as usize;
+    let size = mem::size_of::<T>().checked_mul(count).unwrap();
+    let diff = if src_usize > dst_usize { src_usize - dst_usize } else { dst_usize - src_usize };
+    // If the absolute distance between the ptrs is at least as big as the size of the buffer,
+    // they do not overlap.
+    diff >= size
+}
+
 /// Copies `count * size_of::<T>()` bytes from `src` to `dst`. The source
 /// and destination must *not* overlap.
 ///
@@ -2042,15 +2055,24 @@ pub const unsafe fn copy_nonoverlapping<T>(src: *const T, dst: *mut T, count: us
         pub fn copy_nonoverlapping<T>(src: *const T, dst: *mut T, count: usize);
     }
 
-    // FIXME: Perform these checks only at run time
-    /*if cfg!(debug_assertions)
-        && !(is_aligned_and_not_null(src)
-            && is_aligned_and_not_null(dst)
-            && is_nonoverlapping(src, dst, count))
-    {
-        // Not panicking to keep codegen impact smaller.
-        abort();
-    }*/
+    #[cfg(debug_assertions)]
+    fn runtime_check<T>(src: *const T, dst: *mut T, count: usize) {
+        if !is_aligned_and_not_null(src)
+            || !is_aligned_and_not_null(dst)
+            || !is_nonoverlapping(src, dst, count)
+        {
+            // Not panicking to keep codegen impact smaller.
+            abort();
+        }
+    }
+    #[cfg(debug_assertions)]
+    const fn compiletime_check<T>(_src: *const T, _dst: *mut T, _count: usize) {}
+    #[cfg(debug_assertions)]
+    // SAFETY: runtime debug-assertions are a best-effort basis; it's fine to
+    // not do them during compile time
+    unsafe {
+        const_eval_select((src, dst, count), compiletime_check, runtime_check);
+    }
 
     // SAFETY: the safety contract for `copy_nonoverlapping` must be
     // upheld by the caller.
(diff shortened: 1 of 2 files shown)


oli-obk:
blocked on lang team sign off in the issue

Mark-Simulacrum:
const_eval_select appears to be present on beta, so the `cfg(bootstrap)`s can be dropped now. That should help simplify the code a little.

I would also like to see the comments updated to reflect that we're *not* actually behaving the same way at runtime and within CTFE, but that we're OK to do so because the checks are best-effort (indeed, debug asserts) and not necessary for correctness.

jfrimmel:
I've incorporated the review feedback and rebased onto the current master (while dropping the `cfg(bootstrap)`).

library/core/src/intrinsics.rs
```diff
@@ -2127,11 +2152,24 @@ pub const unsafe fn copy<T>(src: *const T, dst: *mut T, count: usize) {
...
+    #[cfg(debug_assertions)]
+    const fn compiletime_check<T>(_src: *const T, _dst: *mut T, _count: usize) {
+        // nothing, some basic checks are done in the CTFE engine (alignment is
+        // currently NOT checked.
+    }
+    #[cfg(debug_assertions)]
+    // SAFETY: runtime debug-assertions are a best-effort basis; it's fine to
+    // not do them during compile time
+    unsafe {
+        const_eval_select((src, dst, count), compiletime_check, runtime_check);
```
Mark-Simulacrum, line 2171: nit: passing count here seems unnecessary, seems likely it can just be dropped. Symmetry with _nonoverlapping doesn't seem necessary.

library/core/src/intrinsics.rs
```diff
@@ -2127,11 +2152,24 @@ pub const unsafe fn copy<T>(src: *const T, dst: *mut T, count: usize) {
...
+    fn runtime_check<T>(src: *const T, dst: *mut T, _count: usize) {
+        if !is_aligned_and_not_null(src) || !is_aligned_and_not_null(dst) {
+            // Not panicking to keep codegen impact smaller.
+            abort();
+        }
+    }
+    #[cfg(debug_assertions)]
+    const fn compiletime_check<T>(_src: *const T, _dst: *mut T, _count: usize) {
+        // nothing, some basic checks are done in the CTFE engine (alignment is
+        // currently NOT checked.
```
Mark-Simulacrum, line 2165: I don't think we should say this -- the comment on const_eval_select seems sufficient. (Likewise above).

Mark-Simulacrum (commented):
Please also squash the commits. Thanks!

jfrimmel:
@Mark-Simulacrum Sorry, I've missed your comments... I've now implemented the feedback.

@rustbot modify labels to -S-waiting-on-author +S-waiting-on-review

Mark-Simulacrum:
`= note: LINK : fatal error LNK1201: error writing to program database 'D:\a\rust\rust\build\x86_64-pc-windows-msvc\stage2-tools\x86_64-pc-windows-msvc\tmp\cit\t1288\foo\target\release\deps\test.pdb'; check for insufficient disk space, invalid path, or insufficient privilege`

Seems likely to be unrelated to this PR, I hope.

@bors retry

jfrimmel:
Hm, the same error again, but I cannot imagine how this should be related to this PR. There are [other PRs](https://github.com/rust-lang/rust/pulls?q=LINK+fatal+error+LNK1201+) "broken" with this error, e.g. #89488, #80802, ...

So it seems spurious, despite it happening twice.

RalfJung:
Seems to be an instance of https://github.com/rust-lang/rust/issues/81890. Let's hope it eventually works...
@bors retry

PR #90042, opened 2021-10-19
Bump bootstrap compiler to 1.57

### LINKED ISSUE #90152: rustc_mir_dataflow tests fail with debug-assertions enabled
The `rustc_mir_dataflow` tests [`forward_cursor`](https://github.com/rust-lang/rust/blob/68a698baf6bfc61d85ce6e25122a092c60c7f21a/compiler/rustc_mir_dataflow/src/framework/tests.rs#L315) and [`backward_cursor`](https://github.com/rust-lang/rust/blob/68a698baf6bfc61d85ce6e25122a092c60c7f21a/compiler/rustc_mir_dataflow/src/framework/tests.rs#L307) are failing when built with debug-assertions.

This is related to the bootstrap bump in #90042. This brings in a change in cargo where tests now inherit profile settings from the dev/release profiles. When `debug-assertions` is enabled in `config.toml`, tests are now built with debug-assertions enabled where they previously didn't.

These tests seem to intentionally create a [mir body with disconnected basic blocks](https://github.com/rust-lang/rust/blob/68a698baf6bfc61d85ce6e25122a092c60c7f21a/compiler/rustc_mir_dataflow/src/framework/tests.rs#L13-L17). However in #74169 a debug assertion was added to prevent API consumers from visiting those basic blocks by accident. That assertion is now triggering in the test.

I don't have any particular ideas on how to resolve this, since the test seems to be fundamentally going against that assertion.

emilyalbini:
Fixes https://github.com/rust-lang/rust/issues/90152

r? @Mark-Simulacrum


Mark-Simulacrum:
r=me with CI fixed, but I'm not sure what's up with that assert failing...

Issue #89626, opened 2021-10-07
Undefined reference to `getauxval` in function `init_have_lse_atomics`  when compiling to nightly `aarch64-unknown-linux-musl`


XAMPPRocky:
### Code

I received this pull request, which caught a regression for `aarch64-unknown-linux-musl` when compiling for nightly. The same process works for stable and beta. I don't have a machine that can easily compile to this platform, so I can't reduce it at the moment: https://github.com/XAMPPRocky/rasn/pull/49

### Version it worked on

It most recently worked on: Works on stable and beta `1.56.0-beta.4 (e6e620e1c 2021-10-04)`

- https://github.com/XAMPPRocky/rasn/pull/49/checks?check_run_id=3827011175

### Version with regression

`rustc --version --verbose`: 1.57.0-nightly (0eabf25b9 2021-10-06)

- https://github.com/XAMPPRocky/rasn/pull/49/checks?check_run_id=3827012571

### Backtrace

```
= note: /rust/lib/rustlib/aarch64-unknown-linux-musl/lib/libcompiler_builtins-cbf6a032efba8222.rlib(cpu_model.o): In function `init_have_lse_atomics':
          /cargo/registry/src/github.com-1ecc6299db9ec823/compiler_builtins-0.1.49/./lib/builtins/cpu_model.c:786: undefined reference to `getauxval'
          collect2: error: ld returned 1 exit status
          
= help: some `extern` functions couldn't be found; some native libraries may need to be installed or have their path specified
= note: use the `-l` flag to specify native libraries to link
= note: use the `cargo:rustc-link-lib` directive to specify the native libraries to link with Cargo (see https://doc.rust-lang.org/cargo/reference/build-scripts.html#cargorustc-link-libkindname)
```

XAMPPRocky:
Tagging https://github.com/rust-lang/rust/issues/76992 and https://github.com/rust-lang/rust/pull/85805 since they seem relevant and maybe a potential fix.

jbg:
I tried to use this workaround: `RUSTFLAGS="-C target-feature=-outline-atomics"`

But it doesn't change anything. Is it because `+outline-atomics` is added after checking `-C target-feature` [here](https://github.com/rust-lang/rust/pull/83655/files#diff-9f3bfd2fdc228dcb1b352082a7ebc39b19c8bea15b73141f72d9f5021aa3c66cR420)?

apiraino:
[...]

PR #90044, opened 2021-10-19
Restrict aarch64 outline atomics to glibc for now.

hkratz:
The introduced dependency on `getauxval` causes linking problems with musl, making compiling any binaries for `aarch64-unknown-linux-musl` impossible without workarounds such as using lld or adding liblibc.rlib again to the linker invocation, see #89626.

This is a workaround until libc>0.2.108 is merged.

### DIFF
--- a/compiler/rustc_target/src/spec/aarch64_unknown_linux_musl.rs
+++ b/compiler/rustc_target/src/spec/aarch64_unknown_linux_musl.rs
@@ -9,10 +9,6 @@ pub fn target() -> Target {
         pointer_width: 64,
         data_layout: "e-m:e-i8:8:32-i16:16:32-i64:64-i128:128-n32:64-S128".to_string(),
         arch: "aarch64".to_string(),
-        options: TargetOptions {
-            features: "+outline-atomics".to_string(),
-            mcount: "\u{1}_mcount".to_string(),
-            ..base
-        },
+        options: TargetOptions { mcount: "\u{1}_mcount".to_string(), ..base },
     }
 }


jbg:
I may not understand how the `features` argument is interpreted, but shouldn't this be before the `-C target-feature` processing on [line 417](https://github.com/rust-lang/rust/pull/90044/files#diff-9f3bfd2fdc228dcb1b352082a7ebc39b19c8bea15b73141f72d9f5021aa3c66cR417)? It should be *defaulting* outline-atomics to on, but right now it seems to be *forcing* it to on, so `-C target-feature=-outline-atomics` doesn't do anything.

For example, right now you can't get a successful build on `aarch64-unknown-linux-musl` by setting `RUSTFLAGS="-C target-feature=-outline-atomics"`.

hkratz:
Allowing `-C target-feature=-outline-atomics` to override this setting would definitely be an improvement, but it would not work because the std library would still be compiled with it enabled.

But I can move the outline atomics block before `-C target-feature` processing code as well.

hkratz:
If an updated libc with https://github.com/rust-lang/libc/pull/2272 fixes it, then this is no longer needed. Let's wait for that.

@rustbot label -S-waiting-on-review +S-waiting-on-author

jbg:
This issue does not only occur with a libc dependency in the tree. Even a no-deps Hello World program triggers it.

hkratz:
> This issue does not only occur with a libc dependency in the tree. Even a no-deps Hello World program triggers it.

Yes, but the std library uses rust-lang/libc. Having an unbundled libc.a at the right place in link order will resolve this problem as well.

workingjubilee:
😩 Sorry for missing this. Let me see...

workingjubilee:
So it looks like we accidentally broke musl on aarch64 because we aren't compiling musl in a way that is convenient for musl (already a known factor, really, but I didn't expect it to cause an issue here), and:
- we may find out that a recent libc crate update fixes it when it lands
- but we should try to get this resolved before it hits stable
- and another PR against rust-lang/rust offers a known permanent fix Eventually

so this one should wait a little for that but not a long time, correct? Alright, I will keep this on my radar.

workingjubilee:
Speaking of radar, it looks like we bumped LLVM to min 12. Accordingly, https://github.com/rust-lang/rust/pull/90623 was opened, which remedies the FIXME and also will pose a merge conflict with this PR.

I would stop the rollup, but currently we seem to be inclined to let things ride for a few nights and hope a new libc fixes it, so might as well let it ride. It also addresses the second commit in this set.

hkratz:
If the libc update does not fix it, I will update this PR to just remove the `+outline-atomics` feature from the aarch64_unknown_linux_musl target spec.

hkratz:
Damn, this can't be backported to beta as #90623 (on which this builds now) was after the beta cut-off.

workingjubilee:
Whoops, didn't notice that this got updated, sorry. Unfortunately didn't find time to push through the libc cascade all the way, so it looks like this will be it for now. We can figure out if #90681 will fix it on the next train, when we have the luxury of time.
This looks correct.

workingjubilee:
Ah damn.

workingjubilee:
@pnkfelix It seems this PR can't land on beta, but I agree with Hans' remarks on Zulip regarding the severity of the regression. I have no idea how we should proceed in terms of mitigation.

hkratz:
I can open a minimal, functionally equivalent PR against beta (just restricting aarch64 outline atomics to glibc) and test that thoroughly. But I don't know if policy allows PRs that land only in beta.

workingjubilee:
Sometimes we do point releases, so I can't imagine it's completely banned? I am not familiar with that nuance, however.

workingjubilee:
Though, speaking of host tools guarantees:
How did the original regression land if this issue blocks even linking a simple "Hello, world!"?

workingjubilee:
I find this state of affairs baffling but I am going to approve this commit with a note here that it can't actually be backported in as-is form, so that this issue doesn't persist into the next beta cut.
@bors r+ rollup=always

hkratz:
> Though, speaking of host tools guarantees: How did the original regression land if this issue blocks even linking a simple "Hello, world!"?

The host tools are linked against shared musl in CI, which works fine. This is only an issue with +crt-static, which is the default for the aarch64-unknown-linux-musl target.

workingjubilee:
Oh, then the severity of this is somewhat reduced, as musl-based distros patch rustc's musl targets to default to -crt-static (a patch we will eventually take up, and to some degree this hammers home why we should).

Issue #79649, opened 2020-12-02
Add support for branch coverage in source-based coverage


marco-c:
Similarly to what is being done in LLVM and Clang: https://reviews.llvm.org/D84467.

mati865:
cc @richkadel you might want to include it in https://github.com/rust-lang/rust/issues/79121

richkadel:
Actually, at last I checked, this is not in any stable version of LLVM yet. I think it's feasible to stabilize Rust coverage in rustc, based on the current feature of LLVM InstrProf, without depending on proposed future LLVM features. So, IMO, I don't think we should add this to the tracking issue, as a dependency. In fact, until this capability is stabilized in LLVM, and available in a version of LLVM that's supported by rustc, I think it makes sense to keep this as a stand-alone feature request / enhancement.

Let me know if any of my assumptions are incorrect.

richkadel:
Cc @tmandry

mati865:
I have no idea how adding features like that works, I thought it will be available right away in LLVM 12 around March.

tmandry:
Neat. This would certainly improve the way we present coverage for these things, and might make it possible to be more intuitive about how we handle closing braces.

Just to throw it out there, since I'm thinking about async/await right now: This does offer an intriguing possibility of measuring drops at await points. It's possible for buggy code to assume it will never get canceled at an await point, when in fact that's part of the contract. It would be nice if we could highlight this in coverage results. Though explicitly testing cancelation at every await point sounds daunting...

It should be there in the next major version of LLVM, whenever that is.

matthargett:
Looks like this feature was released in LLVM 12.0.0:
https://releases.llvm.org/12.0.0/docs/CoverageMappingFormat.html

14.0.0 was released 2022 March. Similar to the approved suggestion in other issues, where it was decided that an LLVM 11 feature could be used once LLVM 13 was released, could the requirement for code coverage working now be LLVM 12?

I see the original umbrella tracking issue for standardized Rust coverage options still has a few issues open, but it otherwise looks stable and settled.

richkadel:
I should have posted this comment a while ago. (It was discussed when upgrading the coverage mapping format to version 6, in PR #91207).

Before anyone puts much effort into implementing LLVM "branch coverage", I recommend writing a test program that demonstrates where the existing Rust coverage region report would be improved with branch coverage. It's possible that branch coverage adds no value.

[...]

PR #90047, opened 2021-10-19
Update CoverageMappingFormat to Version6

Swatinem:
Version 5 adds Branch Regions which are a prerequisite for branch coverage (#79649) which I plan to take a look at next.
Version 6 can use the zeroth filename as prefix for other relative files.


richkadel (changes requested):

Swatinem:
@richkadel thanks for your feedback so far!

I think I will pause this at least until LLVM 11 and stabilization of the flag happens, in the meanwhile the feedback has guided me in another direction I want to explore.

Swatinem:
Updated the PR and rebased on latest `master`, should be ready for review now.

r? @tmandry

compiler/rustc_llvm/llvm-wrapper/CoverageMappingWrapper.cpp
```diff
@@ -108,5 +109,9 @@ extern "C" void LLVMRustCoverageWriteMappingVarNameToString(RustStringRef Str) {
 }
 
 extern "C" uint32_t LLVMRustCoverageMappingVersion() {
```
tmandry, line 111: This is only used to get `coverageinfo::mapping_version()`?

Swatinem:
Removed the date (and corrected the comment to say that we *output* version 5) as suggested by @wesleywiser and rebased on recent master.

I’m a bit baffled by the CI failure. I can’t reproduce that on my Windows, will try with Linux tomorrow.
From the logs, it looks like the functions (counters) inlined into the main crate are being discarded.

tmandry:
@Swatinem are you still available to investigate the CI failure?

Swatinem:
Haven’t had time yet to investigate, pretty much saturated with other things right now.

compiler/rustc_codegen_llvm/src/coverageinfo/mapgen.rs
```diff
@@ -30,12 +30,11 @@ use tracing::debug;
...
-    // If not, the LLVM Version must be less than 11.
-    let version = coverageinfo::mapping_version();
-    if version != 3 {
-        tcx.sess.fatal("rustc option `-Z instrument-coverage` requires LLVM 11 or higher.");
-    }
+    // While our bundled LLVM might support Coverage Map Version 6
+    // (encoded as a zero-based value: 5), we clamp that to Version 5,
+    // as Version 6 would require us to use the 0-th filename as a path prefix
+    // for all other relative paths, which we don't take advantage of right now.
+    let version = coverageinfo::mapping_version().max(4);
```
trinity-1686a, line 37: I have no idea if this can be related to the CI failure, but the comment seems to not match the behavior of the code : if mapping_version() is 5 (Coverage Map Version 6), `5.max(4)` will return 5.

Swatinem: Oh, you are absolutely right! I will fix this in a minute with another rebase.

richkadel:
> Haven’t had time yet to investigate, pretty much saturated with other things right now.

I could not repro that CI error, neither on my local Linux machine nor on the docker image that I believe corresponds to where that error occurred. I think it was an intermittent error that may not happen again?

Swatinem:
Closing in favor of #91207 which subsumes this commit

PR #90050, opened 2021-10-19
Erase late-bound regions before computing vtable debuginfo name.

michaelwoerister:
Fixes #90019.

The `msvc_enum_fallback()` for computing enum type names needs to access the memory layout of niche enums in order to determine the type name. `compute_debuginfo_vtable_name()` did not properly erase regions before computing type names which made memory layout computation ICE when encountering un-erased regions. 

r? @wesleywiser


wesleywiser (approved):

wesleywiser:
@bors p=1

Fixes stable-to-beta regression.

PR #90054, opened 2021-10-19
Make new symbol mangling scheme default for compiler itself.

michaelwoerister:
As suggest in https://github.com/rust-lang/rust/pull/89917#issuecomment-945888574, this PR enables the new symbol mangling scheme for the compiler itself. The standard library is still compiled using the legacy mangling scheme so that the new symbol format does not show up in user code (yet).

r? @Mark-Simulacrum


jackh726:
Given that #89917 had a perf effect. I think this shouldn't be rolled up. @bors rollup=never

PR #90057, opened 2021-10-19
[BENCH] prepare mono item collection for polymorphization

lcnr:
experimenting with a slightly different approach to polymorphization

r? @davidtwco


lcnr:
that's really good :sweat_smile: both instruction count and max rss regression is acceptable, if not completely negligible

PR #90066, opened 2021-10-19
Add new ThinBox type for 1 stack pointer wide heap allocated trait objects

yaahc:
**Zulip Thread**: https://rust-lang.zulipchat.com/#narrow/stream/219381-t-libs/topic/ThinBox

Based on https://github.com/matthieu-m/rfc2580/blob/b58d1d3cba0d4b5e859d3617ea2d0943aaa31329/examples/thin.rs

Tracking Issue: https://github.com/rust-lang/rust/issues/92791

Usage Trial: https://github.com/yaahc/pgx/pull/1/files

## TODO

- [x] make sure to test with #[repr(align(1024))] structs etc


yaahc:
> Did you intentionally move `box.rs` to `box/mod.rs`? It feels like unnecessary churn.

yea, I didn't want to put all of `ThinBox` in the same module but I can undo that if others prefer.

yaahc:
cc @matthieu-m since you wrote https://github.com/matthieu-m/rfc2580/blob/b58d1d3cba0d4b5e859d3617ea2d0943aaa31329/examples/thin.rs

yaahc:
> Do feel free to reuse as much or as little as you wish from https://github.com/matthieu-m/rfc2580/blob/b58d1d3cba0d4b5e859d3617ea2d0943aaa31329/examples/thin.rs, I open-sourced it under a permissive license so that others could build upon it.

Perfect. I figured as much but I always prefer asking before copying other people's code.

yaahc:
@matthieu-m I updated the PR to utilize your design. I swapped all of the `NonNull` usage to be working with direct raw pointers since that seemed to be the primary diff between your rfc2580 and the current unstable APIs. I think it's all working right now, or at least it is passing the alignment test I added which the previous version failed on, but I'm relatively inexperienced at writing unsafe code like this in rust so if you have time to do a quick review I'd greatly appreciate it.

library/alloc/src/boxed/thin.rs
```diff
@@ -0,0 +1,165 @@
...
+use core::ops::{Deref, DerefMut};
+use core::ptr::Pointee;
+use core::ptr::{self, NonNull};
+
+/// ThinBox.
+///
+/// A thin pointer, regardless of T.
+#[unstable(feature = "thin_box", issue = "92791")]
+pub struct ThinBox<T: ?Sized + Pointee> {
+    ptr: WithHeader<T::Metadata>,
```
compiler-errors, line 18: `T` always implements `Pointee`, so it might be clearer to remove `Pointee` bounds (here and elsewhere) and just write uses of `T::Metadata` like `<T as Pointee>::Metadata`:

```suggestion
pub struct ThinBox<T: ?Sized> {
    ptr: WithHeader<<T as Pointee>::Metadata>,
```

library/alloc/src/boxed/thin.rs
```diff
@@ -0,0 +1,198 @@
...
+        unsafe { &*pointer }
+    }
+}
+
+#[unstable(feature = "thin_box", issue = "92791")]
+impl<T: ?Sized> DerefMut for ThinBox<T> {
+    fn deref_mut(&mut self) -> &mut T {
+        let value = self.data();
+        let metadata = self.meta();
+        let pointer = ptr::from_raw_parts::<T>(value as *const (), metadata) as *mut T;
```
a1phyr, line 106:
```suggestion
        let pointer = ptr::from_raw_parts_mut::<T>(value as *mut (), metadata);
```

matthieu-m:
> @matthieu-m I updated the PR to utilize your design. I swapped all of the `NonNull` usage to be working with direct raw pointers since that seemed to be the primary diff between your rfc2580 and the current unstable APIs. I think it's all working right now, or at least it is passing the alignment test I added which the previous version failed on, but I'm relatively inexperienced at writing unsafe code like this in rust so if you have time to do a quick review I'd greatly appreciate it.

There's a very good reason to use `NonNull` though: it allows compiler optimizations for the `Option<ThinBox<T>>` case.

That is:

- With raw pointer: `size_of::<Option<ThinBox<u8>>() == 16`.
- With `NonNull`: `size_of::<Option<ThinBox<u8>>() == 8`.

When I played around with `Pointee`, `NonNull` implemented all the necessary APIs:

- [`NonNull::from_raw_parts`](https://doc.rust-lang.org/std/ptr/struct.NonNull.html#method.from_raw_parts)
- [`NonNull::to_raw_parts`](https://doc.rust-lang.org/std/ptr/struct.NonNull.html#method.to_raw_parts)

And I'd suggest adding a check that `size_of::<Option<ThinBox<T>>() == size_of::<*const ()>()` for a normal type, a slice, and a trait to ensure this doesn't regress (either as test, or debug assert).

library/alloc/src/boxed/thin.rs
```diff
@@ -0,0 +1,198 @@
...
+    ///
+    /// let thin_slice = ThinBox::<[i32]>::new_unsize([1, 2, 3, 4]);
+    /// ```
+    pub fn new_unsize<T>(value: T) -> Self
+    where
+        T: Unsize<Dyn>,
+    {
+        let dyn_ref: &Dyn = &value;
+        let (_, meta) = (dyn_ref as *const Dyn).to_raw_parts();
+        let ptr = WithHeader::new(meta, value).expect("No allocation failure");
```
a1phyr, line 77: Ibid: I think `alloc::handle_alloc_error` would be more appropriate here.

library/alloc/src/boxed/thin.rs
```diff
@@ -0,0 +1,198 @@
...
+    /// ```
+    /// #![feature(thin_box)]
+    /// use std::boxed::ThinBox;
+    ///
+    /// let five = ThinBox::new(5);
+    /// ```
+    pub fn new(value: T) -> Self {
+        let dyn_ref: &T = &value;
+        let (_, meta) = (dyn_ref as *const T).to_raw_parts();
+        let ptr = WithHeader::new(meta, value).expect("No allocation failure");
```
a1phyr, line 53: I think `alloc::handle_alloc_error` would be more appropriate here.

library/alloc/src/boxed/thin.rs
```diff
@@ -0,0 +1,198 @@
...
+        //  -   NonNull and valid.
+        unsafe { *self.ptr.header() }
+    }
+
+    fn data(&self) -> *mut u8 {
+        self.ptr.value()
+    }
+}
+
+struct WithHeader<H>(*mut u8, PhantomData<H>);
```
a1phyr, line 135: Using `NonNull` here would add a niche so `Option<ThinBox<T>>` would fit in a `usize`.

library/alloc/src/boxed/thin.rs
```diff
@@ -0,0 +1,198 @@
...
+
+impl<H> WithHeader<H> {
+    fn new<T>(header: H, value: T) -> Option<WithHeader<H>> {
+        let layout = Self::alloc_layout(mem::size_of::<T>(), mem::align_of::<T>());
+        let aligned_header_size = Self::aligned_header_size(layout.align());
+
+        unsafe {
+            let ptr = NonNull::new(alloc::alloc(layout))?;
```
a1phyr, line 143: It would be interesting to avoid calling `alloc::alloc` if both `T` and `H` are ZST.

library/alloc/src/boxed/thin.rs
```diff
@@ -0,0 +1,198 @@
...
+            ptr::write(result.header(), header);
+            ptr::write(result.value().cast(), value);
+
+            Some(result)
+        }
+    }
+
+    //  Safety:
+    //  -   Assumes that `value` can be dereferenced.
+    unsafe fn drop<T: ?Sized>(&self, value: *mut T) {
```
a1phyr, line 159: This should definitely take a `&mut`

library/alloc/src/boxed/thin.rs
```diff
@@ -0,0 +1,198 @@
...
+///
+/// use std::mem::{size_of, size_of_val};
+/// let size_of_ptr = size_of::<*const ()>();
+/// assert_eq!(size_of_ptr, size_of_val(&five));
+/// assert_eq!(size_of_ptr, size_of_val(&thin_slice));
+/// ```
+#[unstable(feature = "thin_box", issue = "92791")]
+pub struct ThinBox<T: ?Sized> {
+    ptr: WithHeader<<T as Pointee>::Metadata>,
+    _marker: PhantomData<fn(T) -> T>,
```
a1phyr, line 34: I assume this `PhantomData` is there for invariance, but I think (not sure though) that using `<T as Pointee>::Metadata` is enough here. Additionally, there should be a `PhantomData<T>` for dropck.

a1phyr (commented):
This a great idea, thanks for bringing this to `std` !

compiler-errors:
@matthieu-m, @yaahc - should we use actually [`Unique<T>`](https://docs.rs/ptr/latest/ptr/struct.Unique.html) instead of just `NonNull<T>`, since this type acts semantically like a `Box` and we get more things for free (e.g. `Send`/`Sync`)?

library/alloc/src/boxed/thin.rs
yaahc: > but I think (not sure though) that using `<T as Pointee>::Metadata` is enough here

Can you explain how for me? I believe you I just don't understand how.

library/alloc/src/boxed/thin.rs
ben0x539: It's probably fine because `&self` points at the `NonNull<u8>` while `value` points at (the neighborhood of) the pointee, so it's not actually overlapping, right?

library/alloc/src/boxed/thin.rs
a1phyr: Even if `T` can coerce to `U`, the compiler cannot guarantee that `<T as Pointee>::Metadata` can coerce to `<U as Pointee>::Metadata`, which makes the struct invariant.

[Playground example](https://play.rust-lang.org/?gist=b91fa44920e7dfcc47535d52be7c6f15)

yaahc: Okay, that makes sense, I went ahead and updated the PhantomData to T so I think this should be resolved, thank you!

library/alloc/src/boxed/thin.rs
```diff
@@ -0,0 +1,211 @@
...
+    fn new<T>(header: H, value: T) -> WithHeader<H> {
+        let layout = Self::alloc_layout(mem::size_of::<T>(), mem::align_of::<T>());
+        let aligned_header_size = Self::aligned_header_size(layout.align());
+
+        unsafe {
+            let ptr = alloc::alloc(layout);
+
+            //  Safety:
+            //  -   The size is at least `aligned_header_size`.
+            let ptr = ptr.offset(aligned_header_size as isize) as *mut _;
```
a1phyr, line 156: This offset calculation should happen after the `NonNull` check

library/alloc/src/boxed/thin.rs
yaahc: I'm not exactly sure how to do that, though that does sound fun. I think I'd like to leave that for a followup PR, though I will add some tests right now to make sure our ZST behavior isn't completely broken.

library/alloc/src/boxed/thin.rs
```diff
@@ -0,0 +1,212 @@
...
+        //  -   NonNull and valid.
+        unsafe { *self.ptr.header() }
+    }
+
+    fn data(&self) -> *mut u8 {
+        self.ptr.value()
+    }
+}
+
+struct WithHeader<H>(NonNull<u8>, PhantomData<H>);
```
yaahc, line 143: From https://github.com/rust-lang/rust/pull/90066#issuecomment-1011277371

Should this be `Unique` instead of `NonNull`?

library/alloc/src/boxed/thin.rs
yaahc: Actually though, what tests can I even run to ensure thinbox is working when the T and H are both ZSTs? My understanding is that would mean that the vtable is empty so it would have to be a trait object for something like `trait Useless {}`. what can I even do with that type?

library/alloc/src/boxed/thin.rs
compiler-errors: Regarding my comment, I don't think we gain anything by making this `Unique`. I forgot that we're not storing a `NonNull<T>` but a `NonNull<u8>`, and we are already storing `PhantomData<T>` in the `ThinBox` itself :)

library/alloc/src/boxed/thin.rs
a1phyr: Nevermind, using `alloc::System` does this for us if we add a generic allocator parameter to `ThinBox`, which will be done as a follow-up, so no need to bother about that small optimization now.

library/alloc/src/boxed/thin.rs
```diff
@@ -0,0 +1,219 @@
...
+            ptr::write(result.header(), header);
+            ptr::write(result.value().cast(), value);
+
+            result
+        }
+    }
+
+    //  Safety:
+    //  -   Assumes that `value` can be dereferenced.
+    unsafe fn drop<T: ?Sized>(&self, value: &mut T) {
```
eddyb, line 180: Isn't this risky? I remember there being issues in the past with an (argument) reference's backing storage being deallocated mid-way through a function. Something about the (argument) reference being guaranteed alive at the start vs "throughout". Might need someone else to explain it though.

Maybe running this in miri, with stacked borrows enabled, might be enough to detect an issue? Not sure though, the raw pointers might mask it partly.

Simplest change AFAICT (if this is a problem) would be to make it `value: *mut T` instead, and turn it into a reference only for `{size,align}_of_val`.

library/alloc/src/boxed/thin.rs
```diff
@@ -0,0 +1,219 @@
...
+        let alloc_size = Self::aligned_header_size(alloc_align) + value_size;
+
+        //  Safety:
+        //  -   `align` is not zero and a power of two, as guaranteed by `mem::align_of`.
+        //  -   The size can be rounded up to `align` without overflow, or `H` is way too big.
+        unsafe { Layout::from_size_align_unchecked(alloc_size, alloc_align) }
+    }
+
+    fn aligned_header_size(alloc_align: usize) -> usize {
+        cmp::max(Self::header_size(), alloc_align)
```
thomcc, line 217: This looks a bit suspicious, esp in conjunction with how alloc size is computed above.

I think this will work for currently existing types, but will fail for things where with metadata larger than the value (or header) alignment, but not aligned to value align.

For example, metadata type as `[u8; 17]`, and value type as `u64` (I know this is a bit silly, but we should probably not leave this bug until later). Then the alloc_align will be 8 (for u64), `align_of::<H>()` will be 1, and we'll end up storing the u64 at an unaligned address. (Sorry if this isn't explained well)

So, I think this should be aligning/rounding up here, rather than computing the max. I also like https://doc.rust-lang.org/nightly/std/alloc/struct.Layout.html#method.extend for this kind of thing, although perhaps it's annoying to integrate here without a performance overhead from the `Result`. You could use it to check the values in `alloc_layout` in debug builds, though.

library/alloc/src/boxed/thin.rs
yaahc: I updated it from `value: *mut T` because of [this comment](https://github.com/rust-lang/rust/pull/90066#discussion_r783283147) by @a1phyr. I'm not really sure who is right or what to do so I'm gonna also cc @matthieu-m who originally wrote the code and did use `*mut T` so they can weigh in as well since this is one of the last unresolved comments atm.

cuviper: I think #55005 is an example of what @eddyb means.

library/alloc/src/boxed/thin.rs
eddyb: > Also, using the pointer here may violate an aliasing rule (not sure, but this signature seems dubious).

cc @RalfJung but my understanding was that between `&mut T` and `*mut T`, the latter was more flexible, and the former is the one that can cause more aliasing issues

thomcc: Yes, it seems a lot less dubious to use a pointer here.

library/alloc/src/boxed/thin.rs
yaahc: Okay, this should be resolved now.

library/alloc/src/boxed/thin.rs
yaahc: I swapped it back based on the consensus, marking this as resolved now.

library/alloc/src/boxed/thin.rs
RalfJung: Yeah using a raw ptr seems better.

yaahc:
second attempt!

r? rust-lang/libs

library/alloc/src/boxed/thin.rs
```diff
@@ -0,0 +1,216 @@
...
+    }
+
+    //  Safety:
+    //  -   Assumes that `value` can be dereferenced.
+    unsafe fn drop<T: ?Sized>(&self, value: *mut T) {
+        unsafe {
+            let Ok((layout, value_offset)) = Self::alloc_layout(Layout::for_value_raw(value)) else {
+                // SAFETY: Layout must have been computable if we're in drop
+                unreachable_unchecked()
+            };
```
joshtriplett, line 185: Isn't this `unwrap_unchecked`?

```suggestion
            // SAFETY: Layout must have been computable if we're in drop
            let (layout, value_offset) = Self::alloc_layout(Layout::for_value_raw(value)).unwrap_unchecked();
```

library/alloc/src/boxed/thin.rs
```diff
@@ -0,0 +1,216 @@
...
+        //  -   NonNull and valid.
+        unsafe { *self.ptr.header() }
+    }
+
+    fn data(&self) -> *mut u8 {
+        self.ptr.value()
+    }
+}
+
+struct WithHeader<H>(NonNull<u8>, PhantomData<H>);
```
joshtriplett, line 142:
```suggestion
/// A pointer to type-erased data, guaranteed to have a header `H` before the pointed-to location.
struct WithHeader<H>(NonNull<u8>, PhantomData<H>);
```

library/alloc/src/boxed/thin.rs
```diff
@@ -0,0 +1,216 @@
...
+        //    will always result in an aligned header pointer, it just may not point to the
+        //    beginning of the allocation.
+        unsafe { self.0.as_ptr().sub(Self::header_size()) as *mut H }
+    }
+
+    fn value(&self) -> *mut u8 {
+        self.0.as_ptr()
+    }
+
+    fn header_size() -> usize {
```
joshtriplett, line 209: Could this be `const`?

```suggestion
    const fn header_size() -> usize {
```

(Not entirely sure if this is within our current const capabilities.)

library/alloc/src/lib.rs
```diff
@@ -152,6 +153,7 @@
 #![feature(fundamental)]
 #![cfg_attr(not(test), feature(generator_trait))]
 #![feature(lang_items)]
+#![feature(let_else)]
```
joshtriplett, line 156: :tada:

joshtriplett:
Posted a few comments, and un-resolved one other comment that didn't get addressed.

LGTM otherwise; r=me with those addressed.

PR #89634, opened 2021-10-07
rustc_driver: Enable the `WARN` log level by default

hawkw:
This commit changes the `tracing_subscriber` initialization in
`rustc_driver` so that the `WARN` verbosity level is enabled by default
when the `RUSTC_LOG` env variable is empty. If the `RUSTC_LOG` env
variable is set, the filter string in the environment variable is
honored, instead.

Fixes #76824
Closes #89623

cc @eddyb, @oli-obk

PR #90072, opened 2021-10-20
Don't emit a warning for empty rmeta files.

ehuss:
This avoids displaying a warning when attempting to load an empty rmeta file. Warnings were enabled via #89634 which can cause a lot of noise (for example, running `./x.py check`).  rustc generates empty rmeta files for things like binaries, which can happen when checking libraries as unittests.

Closes #89795


Mark-Simulacrum:
@bors r+ p=1

I think this makes sense. FWIW, it may be worth putting in the effort to avoid the user-visible tracing warnings, because they don't provide guidance on what to do -- we should strive to not give such warnings to users, IMO. Unconditional warnings with a help: text would already be an improvement.

cc @petrochenkov

PR #90075, opened 2021-10-20
Edit error messages for `rustc_resolve::AmbiguityKind` variants

### LINKED ISSUE #79717: Ambiguous name error message has weird wording
```rust
mod serde {}

use serde::foo;
```

([Playground](https://play.rust-lang.org/?version=nightly&mode=debug&edition=2018&gist=f65bf7e111dd7ac8f3c871058e9b1b72))

Errors:

```
error[E0432]: unresolved import `serde::foo`
 --> src/lib.rs:3:5
  |
3 | use serde::foo;
  |     ^^^^^^^^^^ no `foo` in `serde`

error[E0659]: `serde` is ambiguous (name vs any other name during import resolution)
 --> src/lib.rs:3:5
  |
3 | use serde::foo;
  |     ^^^^^ ambiguous name
  |
  = note: `serde` could refer to a crate passed with `--extern`
  = help: use `::serde` to refer to this crate unambiguously
note: `serde` could also refer to the module defined here
 --> src/lib.rs:1:1
  |
1 | mod serde {}
  | ^^^^^^^^^^^^
  = help: use `crate::serde` to refer to this module unambiguously

Some errors have detailed explanations: E0432, E0659.
For more information about an error, try `rustc --explain E0432`.

```

The error message says:

> `serde` is ambiguous (name vs any other name during import resolution)

But what does it mean by "name vs any other name during import resolution"? I think this should be reworded or even deleted – it doesn't seem to add anything.

pierwill:
Edit the language of the ambiguity descriptions for E0659. These strings now appear as notes.

Closes #79717.

### DIFF
--- a/compiler/rustc_resolve/src/diagnostics.rs
+++ b/compiler/rustc_resolve/src/diagnostics.rs
@@ -1165,14 +1165,9 @@ impl<'a> Resolver<'a> {
             (b1, b2, misc1, misc2, false)
         };
 
-        let mut err = struct_span_err!(
-            self.session,
-            ident.span,
-            E0659,
-            "`{ident}` is ambiguous ({why})",
-            why = kind.descr()
-        );
+        let mut err = struct_span_err!(self.session, ident.span, E0659, "`{ident}` is ambiguous");
         err.span_label(ident.span, "ambiguous name");
+        err.note(&format!("ambiguous because of {}", kind.descr()));
 
         let mut could_refer_to = |b: &NameBinding<'_>, misc: AmbiguityErrorMisc, also: &str| {
             let what = self.binding_description(b, ident, misc == AmbiguityErrorMisc::FromPrelude);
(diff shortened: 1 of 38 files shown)


camelid:
@pierwill you need to bless tests :)

pierwill:
I'll take another look at these, @camelid.

camelid:
Stepping back a bit, I think it might be better if these notes were displayed as actual `note`s, rather than in parentheses in the main message. I'm not sure what others think, but do you want to try changing that and see how it looks?

Basically, I'm envisioning something like this:

```
error: `foo` is ambiguous
   --> $INFO
    |
$LL | $SRC_CODE
    |
    = note: ambiguous due to conflicts between ...
```

pierwill:
Tests pass locally with `err.note`. 🎉

pierwill:
Whoops. Tagged for right before a new crop of failures. 😬

camelid (commented):
I left a few comments on some changes that I felt could be made more informative. In general, some of the existing descriptions included "during import/macro resolution" or similar, which seems unhelpful and confusing. @petrochenkov i see you've assigned yourself on this PR; what do you think of removing the "during import/macro resolution" part?

camelid (approved):
`r=petrochenkov,camelid` with the below nits addressed and petrochenkov's approval. Thanks for making this change!

Issue #17904, opened 2014-10-10
Can't use `where` clauses with tuple structs


japaric:
### STR

``` rust
// OK
struct Bar<T> where T: Copy {
    inner: T,
}

// Is this the right syntax?
struct Foo<T>(T) where T: Copy;  //~ error: expected `;`, found `where`

fn main() {}
```

### Version

``` rust
rustc 0.12.0-dev (10fc25a39 2014-10-09 16:04:25 -0500)
```

(PR 17669 on top of b08ecf7)

PR #90076, opened 2021-10-20
Change location of where clause on GATs

### LINKED ISSUE #89122: Proposal: Change syntax of where clauses on type aliases
[Source](https://rust-lang.github.io/generic-associated-types-initiative/design-discussions/where-the-where.html)

## Summary

Proposed: to alter the syntax of where clauses on type aliases so that they appear *after* the value:

```
type StringMap<K> = BTreeMap<K, String>
where
    K: PartialOrd
```

This applies both in top-level modules and in trats (associated types, generic or otherwise).

## Background

The current syntax for where to place the "where clause" of a generic associated types is awkward. Consider this example ([playground](https://play.rust-lang.org/?version=nightly&mode=debug&edition=2018&gist=bdb55a5d5cb17e20d73e22a3f2db0e57)):

```rust
trait Iterable {
    type Iter<'a> where Self: 'a;

    fn iter(&self) -> Self::Iter<'_>;
}

impl<T> Iterable for Vec<T> {
    type Iter<'a>
    where 
        Self: 'a = <&'a [T] as IntoIterator>::IntoIter;

    fn iter(&self) -> Self::Iter<'_> {
        self.iter()
    }
}
```

Note the impl. Most people expect the impl to be written as follows (indeed, the author wrote it this way in the first draft):

```rust
impl Iterable for Vec<T> {
    type Iter<'a>  = <&'a [T] as Iterator>::Iter
    where 
        Self: 'a;

    fn iter(&self) -> Self::Iter<'_> {
        self.iter()
    }
}
```

However, this placement of the where clause is in fact rather inconsistent, since the `= <&'a [T] as Iterator>::Iter` is in some sense the "body" of the item.

The same current syntax is used for where clauses on type aliases ([playground](https://play.rust-lang.org/?version=nightly&mode=debug&edition=2018&gist=74eeed1795b693f238150f825a0e8438)):

```rust
type Foo<T> where T: Eq = Vec<T>;

fn main() { }
```

## Top-level type aliases

Currently, we accept where clauses in top-level type aliases, but they are deprecated (warning) and semi-ignored:

```
type StringMap<K> where
    K: PartialOrd
= BTreeMap<K, String>
```

Under this proposal, this syntax remains, but is deprecated. The newer syntax for type aliases (with `where` coming after the type) would remain feature gated until such time as we enforce the expected semantics.

### Keep the current syntax.

In this case, we must settle the question of how we expect it to be formatted (surely not as I have shown it above).

```rust
impl<T> Iterable for Vec<T> {
    type Iter<'a> where Self: 'a 
        = <&'a [T] as IntoIterator>::IntoIter;

    fn iter(&self) -> Self::Iter<'_> {
        self.iter()
    }
}
```

### Accept either

What do we do if both are supplied?

jackh726:
Closes #89122

~Blocked on lang FCP~

r? @nikomatsakis

### DIFF
--- a/compiler/rustc_ast/src/ast.rs
+++ b/compiler/rustc_ast/src/ast.rs
@@ -2662,10 +2662,23 @@ pub struct Trait {
     pub items: Vec<P<AssocItem>>,
 }
 
+/// The location of a where clause on a `TyAlias` (`Span`) and whether there was
+/// a `where` keyword (`bool`). This is split out from `WhereClause`, since there
+/// are two locations for where clause on type aliases, but their predicates
+/// are concatenated together.
+#[derive(Copy, Clone, Encodable, Decodable, Debug, Default)]
+pub struct TyAliasWhereClause(pub bool, pub Span);
+
 #[derive(Clone, Encodable, Decodable, Debug)]
 pub struct TyAlias {
     pub defaultness: Defaultness,
     pub generics: Generics,
+    /// The span information for the two where clauses (before equals, after equals)
+    pub where_clauses: (TyAliasWhereClause, TyAliasWhereClause),
+    /// The index in `generics.where_clause.predicates` that would split into
+    /// predicates from the where clause before the equals and the predicates
+    /// from the where clause after the equals
+    pub where_predicates_split: usize,
     pub bounds: GenericBounds,
     pub ty: Option<P<Ty>>,
 }
--- a/compiler/rustc_ast/src/ast.rs
+++ b/compiler/rustc_ast/src/ast.rs
@@ -2668,7 +2668,7 @@ pub struct Trait {
 /// are concatenated together.
 ///
 /// Take this example:
-/// ```rust, ignore
+/// ```ignore (only-for-syntax-highlight)
 /// trait Foo {
 ///   type Assoc<'a, 'b> where Self: 'a, Self: 'b;
 /// }
@@ -2677,9 +2677,9 @@ pub struct Trait {
 ///   //                 ^^^^^^^^^^^^^^ first where clause
 ///   //                                     ^^^^^^^^^^^^^^ second where clause
 /// }
+/// ```
 ///
 /// If there is no where clause, then this is `false` with `DUMMY_SP`.
-/// ```
 #[derive(Copy, Clone, Encodable, Decodable, Debug, Default)]
 pub struct TyAliasWhereClause(pub bool, pub Span);
 
--- a/compiler/rustc_ast/src/ast.rs
+++ b/compiler/rustc_ast/src/ast.rs
@@ -2666,6 +2666,18 @@ pub struct Trait {
 /// a `where` keyword (`bool`). This is split out from `WhereClause`, since there
 /// are two locations for where clause on type aliases, but their predicates
 /// are concatenated together.
+///
+/// Take this example:
+/// ```rust, ignore
+/// trait Foo {
+///   type Assoc<'a, 'b> where Self: 'a, Self: 'b;
+/// }
+/// impl Foo for () {
+///   type Assoc<'a, 'b> where Self: 'a = () where Self: 'b;
+///   //                 ^^^^^^^^^^^^^^ first where clause
+///   //                                     ^^^^^^^^^^^^^^ second where clause
+/// }
+/// ```
 #[derive(Copy, Clone, Encodable, Decodable, Debug, Default)]
 pub struct TyAliasWhereClause(pub bool, pub Span);
 
--- a/compiler/rustc_ast/src/ast.rs
+++ b/compiler/rustc_ast/src/ast.rs
@@ -2677,6 +2677,8 @@ pub struct Trait {
 ///   //                 ^^^^^^^^^^^^^^ first where clause
 ///   //                                     ^^^^^^^^^^^^^^ second where clause
 /// }
+///
+/// If there is no where clause, then this is `false` with `DUMMY_SP`.
 /// ```
 #[derive(Copy, Clone, Encodable, Decodable, Debug, Default)]
 pub struct TyAliasWhereClause(pub bool, pub Span);
--- a/compiler/rustc_ast_passes/src/ast_validation.rs
+++ b/compiler/rustc_ast_passes/src/ast_validation.rs
@@ -1275,10 +1275,14 @@ impl<'a> Visitor<'a> for AstValidator<'a> {
                 }
                 self.check_type_no_bounds(bounds, "this context");
                 if where_clauses.1.0 {
-                    self.err_handler().span_err(
+                    let mut err = self.err_handler().struct_span_err(
                         where_clauses.1.1,
                         "where clauses are not allowed after the type for type aliases",
-                    )
+                    );
+                    err.note(
+                        "see issue #89122 <https://github.com/rust-lang/rust/issues/89122> for more information",
+                    );
+                    err.emit();
                 }
             }
             _ => {}
--- a/compiler/rustc_ast_passes/src/ast_validation.rs
+++ b/compiler/rustc_ast_passes/src/ast_validation.rs
@@ -11,7 +11,7 @@ use rustc_ast::ptr::P;
 use rustc_ast::visit::{self, AssocCtxt, FnCtxt, FnKind, Visitor};
 use rustc_ast::walk_list;
 use rustc_ast::*;
-use rustc_ast_pretty::pprust;
+use rustc_ast_pretty::pprust::{self, State};
 use rustc_data_structures::fx::FxHashMap;
 use rustc_errors::{error_code, pluralize, struct_span_err, Applicability};
 use rustc_parse::validate_attr;
@@ -118,6 +118,40 @@ impl<'a> AstValidator<'a> {
         } else {
             sess.struct_span_err(expr.span, "expected expression, found statement (`let`)")
                 .note("variable declaration using `let` is a statement")
+                .emit();
+        }
+    }
+
+    fn check_gat_where(
+        &self,
+        before_predicates: &[WherePredicate],
+        where_clauses: (ast::TyAliasWhereClause, ast::TyAliasWhereClause),
+    ) {
+        let sess = &self.session;
+        if !before_predicates.is_empty() {
+            let mut state = State::new();
+            if !where_clauses.1.0 {
+                state.space();
+                state.word_space("where");
+            } else {
+                state.word_space(",");
+            }
+            let mut first = true;
+            for p in before_predicates.iter() {
+                if !first {
+                    state.word_space(",");
+                }
+                first = false;
+                state.print_where_predicate(p);
+            }
+            let suggestion = state.s.eof();
+            sess.struct_span_err(where_clauses.0.1, "where clause not allowed here")
+                .span_suggestion(
+                    where_clauses.1.1.shrink_to_hi(),
+                    "move it here",
+                    suggestion,
+                    Applicability::MachineApplicable,
+                )
                 .emit();
         }
     }
@@ -454,7 +488,7 @@ impl<'a> AstValidator<'a> {
             .emit();
     }
 
-    fn check_foreign_ty_genericless(&self, generics: &Generics) {
+    fn check_foreign_ty_genericless(&self, generics: &Generics, where_span: Span) {
         let cannot_have = |span, descr, remove_descr| {
             self.err_handler()
                 .struct_span_err(
@@ -477,7 +511,7 @@ impl<'a> AstValidator<'a> {
         }
 
         if !generics.where_clause.predicates.is_empty() {
-            cannot_have(generics.where_clause.span, "`where` clauses", "`where` clause");
+            cannot_have(where_span, "`where` clauses", "`where` clause");
         }
     }
 
@@ -1223,13 +1257,25 @@ impl<'a> Visitor<'a> for AstValidator<'a> {
                 let msg = "free static item without body";
                 self.error_item_without_body(item.span, "static", msg, " = <expr>;");
             }
-            ItemKind::TyAlias(box TyAlias { defaultness, ref bounds, ref ty, .. }) => {
+            ItemKind::TyAlias(box TyAlias {
+                defaultness,
+                where_clauses,
+                ref bounds,
+                ref ty,
+                ..
+            }) => {
                 self.check_defaultness(item.span, defaultness);
                 if ty.is_none() {
                     let msg = "free type alias without body";
                     self.error_item_without_body(item.span, "type", msg, " = <type>;");
                 }
                 self.check_type_no_bounds(bounds, "this context");
+                if where_clauses.1.0 {
+                    self.err_handler().span_err(
+                        where_clauses.1.1,
+                        "where clauses are not allowed after the type for type aliases",
+                    )
+                }
             }
             _ => {}
         }
@@ -1245,11 +1291,18 @@ impl<'a> Visitor<'a> for AstValidator<'a> {
                 self.check_foreign_fn_headerless(fi.ident, fi.span, sig.header);
                 self.check_foreign_item_ascii_only(fi.ident);
             }
-            ForeignItemKind::TyAlias(box TyAlias { defaultness, generics, bounds, ty, .. }) => {
+            ForeignItemKind::TyAlias(box TyAlias {
+                defaultness,
+                generics,
+                where_clauses,
+                bounds,
+                ty,
+                ..
+            }) => {
                 self.check_defaultness(fi.span, *defaultness);
                 self.check_foreign_kind_bodyless(fi.ident, "type", ty.as_ref().map(|b| b.span));
                 self.check_type_no_bounds(bounds, "`extern` blocks");
-                self.check_foreign_ty_genericless(generics);
+                self.check_foreign_ty_genericless(generics, where_clauses.0.1);
                 self.check_foreign_item_ascii_only(fi.ident);
             }
             ForeignItemKind::Static(_, _, body) => {
@@ -1503,9 +1556,22 @@ impl<'a> Visitor<'a> for AstValidator<'a> {
                 AssocItemKind::Fn(box Fn { body, .. }) => {
                     self.check_impl_item_provided(item.span, body, "function", " { <body> }");
                 }
-                AssocItemKind::TyAlias(box TyAlias { bounds, ty, .. }) => {
+                AssocItemKind::TyAlias(box TyAlias {
+                    generics,
+                    where_clauses,
+                    where_predicates_split,
+                    bounds,
+                    ty,
+                    ..
+                }) => {
                     self.check_impl_item_provided(item.span, ty, "type", " = <type>;");
                     self.check_type_no_bounds(bounds, "`impl`s");
+                    if ty.is_some() {
+                        self.check_gat_where(
+                            generics.where_clause.predicates.split_at(*where_predicates_split).0,
+                            *where_clauses,
+                        );
+                    }
                 }
                 _ => {}
             }
--- a/compiler/rustc_ast_passes/src/ast_validation.rs
+++ b/compiler/rustc_ast_passes/src/ast_validation.rs
@@ -15,7 +15,9 @@ use rustc_ast_pretty::pprust::{self, State};
 use rustc_data_structures::fx::FxHashMap;
 use rustc_errors::{error_code, pluralize, struct_span_err, Applicability};
 use rustc_parse::validate_attr;
-use rustc_session::lint::builtin::{MISSING_ABI, PATTERNS_IN_FNS_WITHOUT_BODY};
+use rustc_session::lint::builtin::{
+    DEPRECATED_WHERE_CLAUSE_LOCATION, MISSING_ABI, PATTERNS_IN_FNS_WITHOUT_BODY,
+};
 use rustc_session::lint::{BuiltinLintDiagnostics, LintBuffer};
 use rustc_session::Session;
 use rustc_span::source_map::Spanned;
@@ -123,11 +125,11 @@ impl<'a> AstValidator<'a> {
     }
 
     fn check_gat_where(
-        &self,
+        &mut self,
+        id: NodeId,
         before_predicates: &[WherePredicate],
         where_clauses: (ast::TyAliasWhereClause, ast::TyAliasWhereClause),
     ) {
-        let sess = &self.session;
         if !before_predicates.is_empty() {
             let mut state = State::new();
             if !where_clauses.1.0 {
@@ -145,14 +147,16 @@ impl<'a> AstValidator<'a> {
                 state.print_where_predicate(p);
             }
             let suggestion = state.s.eof();
-            sess.struct_span_err(where_clauses.0.1, "where clause not allowed here")
-                .span_suggestion(
+            self.lint_buffer.buffer_lint_with_diagnostic(
+                DEPRECATED_WHERE_CLAUSE_LOCATION,
+                id,
+                where_clauses.0.1,
+                "where clause not allowed here",
+                BuiltinLintDiagnostics::DeprecatedWhereclauseLocation(
                     where_clauses.1.1.shrink_to_hi(),
-                    "move it here",
                     suggestion,
-                    Applicability::MachineApplicable,
-                )
-                .emit();
+                ),
+            );
         }
     }
 
@@ -1568,6 +1572,7 @@ impl<'a> Visitor<'a> for AstValidator<'a> {
                     self.check_type_no_bounds(bounds, "`impl`s");
                     if ty.is_some() {
                         self.check_gat_where(
+                            item.id,
                             generics.where_clause.predicates.split_at(*where_predicates_split).0,
                             *where_clauses,
                         );
--- a/src/test/ui/generic-associated-types/issue-84931.stderr
+++ b/src/test/ui/generic-associated-types/issue-84931.stderr
@@ -2,9 +2,9 @@ error[E0309]: the parameter type `T` may not live long enough
   --> $DIR/issue-84931.rs:15:21
    |
 LL |     type Item<'a> = &'a mut T;
-   |                  -  ^^^^^^^^^ ...so that the reference type `&'a mut T` does not outlive the data it points at
-   |                  |
-   |                  help: consider adding a where clause: `where T: 'a`
+   |                     ^^^^^^^^^- help: consider adding a where clause: `where T: 'a`
+   |                     |
+   |                     ...so that the reference type `&'a mut T` does not outlive the data it points at
 
 error: aborting due to previous error
 
--- a/src/test/ui/parser/type-alias-where.stderr
+++ b/src/test/ui/parser/type-alias-where.stderr
@@ -10,23 +10,25 @@ error: where clauses are not allowed after the type for type aliases
 LL | type Baz = () where;
    |               ^^^^^
 
-error: where clause not allowed here
+warning: where clause not allowed here
   --> $DIR/type-alias-where.rs:22:16
    |
 LL |     type Assoc where u32: Copy = ();
    |                ^^^^^^^^^^^^^^^     - help: move it here: `where u32: Copy`
+   |
+   = note: `#[warn(deprecated_where_clause_location)]` on by default
 
-error: where clause not allowed here
+warning: where clause not allowed here
   --> $DIR/type-alias-where.rs:25:17
    |
 LL |     type Assoc2 where u32: Copy = () where i32: Copy;
    |                 ^^^^^^^^^^^^^^^                     - help: move it here: `, u32: Copy`
 
-error: where clause not allowed here
+warning: where clause not allowed here
   --> $DIR/type-alias-where.rs:33:17
    |
 LL |     type Assoc2 where u32: Copy, i32: Copy = ();
    |                 ^^^^^^^^^^^^^^^^^^^^^^^^^^     - help: move it here: `where u32: Copy, i32: Copy`
 
-error: aborting due to 5 previous errors
+error: aborting due to 2 previous errors; 3 warnings emitted
 
--- a/src/test/ui/parser/type-alias-where.stderr
+++ b/src/test/ui/parser/type-alias-where.stderr
@@ -1,40 +1,32 @@
-error: where clause not allowed here
+error: where clauses are not allowed after the type for type aliases
   --> $DIR/type-alias-where.rs:8:15
    |
 LL | type Bar = () where u32: Copy;
-   |         -     ^^^^^^^^^^^^^^^
-   |         |
-   |         help: move it here: `where u32: Copy`
+   |               ^^^^^^^^^^^^^^^
 
-error: where clause not allowed here
+error: where clauses are not allowed after the type for type aliases
   --> $DIR/type-alias-where.rs:10:15
    |
 LL | type Baz = () where;
    |               ^^^^^
 
 error: where clause not allowed here
-  --> $DIR/type-alias-where.rs:24:38
+  --> $DIR/type-alias-where.rs:22:16
+   |
+LL |     type Assoc where u32: Copy = ();
+   |                ^^^^^^^^^^^^^^^     - help: move it here: `where u32: Copy`
+
+error: where clause not allowed here
+  --> $DIR/type-alias-where.rs:25:17
    |
 LL |     type Assoc2 where u32: Copy = () where i32: Copy;
-   |                                -     ^^^^^^^^^^^^^^^
-   |                                |
-   |                                help: move it here: `, i32: Copy`
+   |                 ^^^^^^^^^^^^^^^                     - help: move it here: `, u32: Copy`
 
 error: where clause not allowed here
-  --> $DIR/type-alias-where.rs:30:21
+  --> $DIR/type-alias-where.rs:33:17
    |
-LL |     type Assoc = () where u32: Copy;
-   |               -     ^^^^^^^^^^^^^^^
-   |               |
-   |               help: move it here: `where u32: Copy`
-
-error: where clause not allowed here
-  --> $DIR/type-alias-where.rs:33:22
-   |
-LL |     type Assoc2 = () where u32: Copy, i32: Copy;
-   |                -     ^^^^^^^^^^^^^^^^^^^^^^^^^^
-   |                |
-   |                help: move it here: `where u32: Copy, i32: Copy`
+LL |     type Assoc2 where u32: Copy, i32: Copy = ();
+   |                 ^^^^^^^^^^^^^^^^^^^^^^^^^^     - help: move it here: `where u32: Copy, i32: Copy`
 
 error: aborting due to 5 previous errors
 
--- a/src/test/ui/parser/type-alias-where.stderr
+++ b/src/test/ui/parser/type-alias-where.stderr
@@ -3,32 +3,16 @@ error: where clauses are not allowed after the type for type aliases
    |
 LL | type Bar = () where u32: Copy;
    |               ^^^^^^^^^^^^^^^
+   |
+   = note: see issue #89122 <https://github.com/rust-lang/rust/issues/89122> for more information
 
 error: where clauses are not allowed after the type for type aliases
   --> $DIR/type-alias-where.rs:10:15
    |
 LL | type Baz = () where;
    |               ^^^^^
-
-warning: where clause not allowed here
-  --> $DIR/type-alias-where.rs:22:16
    |
-LL |     type Assoc where u32: Copy = ();
-   |                ^^^^^^^^^^^^^^^     - help: move it here: `where u32: Copy`
-   |
-   = note: `#[warn(deprecated_where_clause_location)]` on by default
+   = note: see issue #89122 <https://github.com/rust-lang/rust/issues/89122> for more information
 
-warning: where clause not allowed here
-  --> $DIR/type-alias-where.rs:25:17
-   |
-LL |     type Assoc2 where u32: Copy = () where i32: Copy;
-   |                 ^^^^^^^^^^^^^^^                     - help: move it here: `, u32: Copy`
-
-warning: where clause not allowed here
-  --> $DIR/type-alias-where.rs:33:17
-   |
-LL |     type Assoc2 where u32: Copy, i32: Copy = ();
-   |                 ^^^^^^^^^^^^^^^^^^^^^^^^^^     - help: move it here: `where u32: Copy, i32: Copy`
-
-error: aborting due to 2 previous errors; 3 warnings emitted
+error: aborting due to 2 previous errors
 
--- a/compiler/rustc_ast/src/mut_visit.rs
+++ b/compiler/rustc_ast/src/mut_visit.rs
@@ -1018,9 +1018,13 @@ pub fn noop_visit_item_kind<T: MutVisitor>(kind: &mut ItemKind, vis: &mut T) {
         }
         ItemKind::ForeignMod(nm) => vis.visit_foreign_mod(nm),
         ItemKind::GlobalAsm(asm) => noop_visit_inline_asm(asm, vis),
-        ItemKind::TyAlias(box TyAlias { defaultness, generics, bounds, ty }) => {
+        ItemKind::TyAlias(box TyAlias {
+            defaultness, generics, where_clauses, bounds, ty, ..
+        }) => {
             visit_defaultness(defaultness, vis);
             vis.visit_generics(generics);
+            vis.visit_span(&mut where_clauses.0.1);
+            vis.visit_span(&mut where_clauses.1.1);
             visit_bounds(bounds, vis);
             visit_opt(ty, |ty| vis.visit_ty(ty));
         }
@@ -1087,9 +1091,18 @@ pub fn noop_flat_map_assoc_item<T: MutVisitor>(
             visit_fn_sig(sig, visitor);
             visit_opt(body, |body| visitor.visit_block(body));
         }
-        AssocItemKind::TyAlias(box TyAlias { defaultness, generics, bounds, ty }) => {
+        AssocItemKind::TyAlias(box TyAlias {
+            defaultness,
+            generics,
+            where_clauses,
+            bounds,
+            ty,
+            ..
+        }) => {
             visit_defaultness(defaultness, visitor);
             visitor.visit_generics(generics);
+            visitor.visit_span(&mut where_clauses.0.1);
+            visitor.visit_span(&mut where_clauses.1.1);
             visit_bounds(bounds, visitor);
             visit_opt(ty, |ty| visitor.visit_ty(ty));
         }
@@ -1152,9 +1165,18 @@ pub fn noop_flat_map_foreign_item<T: MutVisitor>(
             visit_fn_sig(sig, visitor);
             visit_opt(body, |body| visitor.visit_block(body));
         }
-        ForeignItemKind::TyAlias(box TyAlias { defaultness, generics, bounds, ty }) => {
+        ForeignItemKind::TyAlias(box TyAlias {
+            defaultness,
+            generics,
+            where_clauses,
+            bounds,
+            ty,
+            ..
+        }) => {
             visit_defaultness(defaultness, visitor);
             visitor.visit_generics(generics);
+            visitor.visit_span(&mut where_clauses.0.1);
+            visitor.visit_span(&mut where_clauses.1.1);
             visit_bounds(bounds, visitor);
             visit_opt(ty, |ty| visitor.visit_ty(ty));
         }
--- a/compiler/rustc_ast/src/visit.rs
+++ b/compiler/rustc_ast/src/visit.rs
@@ -303,7 +303,7 @@ pub fn walk_item<'a, V: Visitor<'a>>(visitor: &mut V, item: &'a Item) {
             walk_list!(visitor, visit_foreign_item, &foreign_module.items);
         }
         ItemKind::GlobalAsm(ref asm) => walk_inline_asm(visitor, asm),
-        ItemKind::TyAlias(box TyAlias { defaultness: _, ref generics, ref bounds, ref ty }) => {
+        ItemKind::TyAlias(box TyAlias { ref generics, ref bounds, ref ty, .. }) => {
             visitor.visit_generics(generics);
             walk_list!(visitor, visit_param_bound, bounds);
             walk_list!(visitor, visit_ty, ty);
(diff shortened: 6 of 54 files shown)


nikomatsakis (changes requested):
As discussed on our call, let's change this to store a `usize` "split point" instead.

nikomatsakis:
@rfcbot fcp merge

[This document is also published on hackmd](https://hackmd.io/@nikomatsakis/BJuyGgXvY)

# Where does the where clause go? (Conclusion)

This is write-up of the conclusion to the [where does the where clause go?] question.

## Conclusion

### Where clauses in generic associated types comes after value/binding

Example:

```rust
trait MyTrait {
    type MyType<T>: Iterator
    where
        T: Ord;
}
    
impl MyTrait for MyOtherType {
    type MyType<T> = MyIterator
    where
        T: Ord;
}
```

Effectively the `= type` in the impl replaces the `: Bound` from the declaration with the value that has to meet those bounds.

### Later phase: type aliases

Type aliases will eventually be aligned to this syntax as well:

```rust
type MyType<T> = Vec<T> where T: Ord;
```

Currently, however, where clauses on type aliases are ignored, so we will not stabilize this new syntax until they have the meaning we want.

### Suggestions for users who put the where clause in the wrong place

We will parse where clauses in both positions and suggest to users that they be moved:


```rust
impl MyTrait for MyOtherType {
    type MyType<T>
    where
        T: Ord
    = MyIterator;
}
```

Gets an error with a suggested rewrite. The compiler proceeds "as if" the where clauses had been written after the `= Type`.

### Where clause syntax for trait aliases will have to be revisited

As described in the FAQ below, we currently support trait alias syntax like

```rust
trait ReverseEq<T> = where T: PartialEq<Self>;
```

This syntax will be removed. Although its capabilities could be useful, it is also quite confusing (the placement of the `where` is a subtle distinction), and not clearly needed. If we find that we do want it, we can add in a similar syntax later, but hopefully in a way that is more broadly consistent with the language.

## Discussion and FAQ

### But isn't it inconsistent with other trait items to put the where clauses *before* the `=`?

From one perspective, yes. One can view the value of an associated type as its "body", and the where clauses typically come before the "body" of an item. Put another way, typically you can "copy and paste" the impl and then add some text to the end of each item to specify its value: but with this syntax, you have to edit the "middle" of an associated type to specify its value.

The analogy of an associated type value to a function body, however, is somewhat flawed. The value of an associated type needs to be considered part of the "signature", or public facing, part of the impl. Consider: you can change the body of a function and be certain that your callees will still compile, but you cannot do the same for the value of an associated type.

Given this perspective, when you copy the associated type from the trait to the impl, you are "completing" the signature that was left incomplete by the trait. Moreover, to do so, you replace the `: Bound1 + Bound2` list (which constraints what kinds of types the impl might use) with a specific type, thus making it more specific.

### What about a more purely syntactic point-of-view? What is more consistent?

There is precedent that the placement of the where clause has less to do with the logical role that it plays and more to do with other factors, like whether it is followed by a braced list of items:

* With `struct Foo<T> where T: Ord { t: T }`, the "body" of the struct is its fields, and the where clause comes first.
* But we write `struct Foo<T>(T) where T: Ord`, thus placing the "body" (the fields `(T)`) first and the where clause second. Moreover, we initially implemented the grammar `struct Foo<T> where T: Ord (T)` but this was deemed so obviously confusing that it was [changed with little discussion](https://github.com/rust-lang/rust/issues/17904#issuecomment-58603749).

As further evidence that this syntax is inconsistent with Rust's traditions, placing the where clauses before the `= ty` makes it objectively hard to determine how to run rustfmt in a way that feels natural. rustfmt handles `where` by putting the `where` onto its own line, with one line per where clause. This structure works for existing Rust items because where clauses are always either following by nothing (tuple structs) or by a braced (`{}`) list of items (e.g., struct fields, fn body, etc). That opening `{` can therefore go on its own line. This `where` clause formatting does not work well with `=`.

The idea of having where clauses come at the "end" of the signature is also supported by the [original RFC](https://rust-lang.github.io/rfcs/0135-where.html#readability), which motivated where clauses in part by describing how they allow you to treat the precise bounds as "secondary" to the "important stuff":

> If I may step aside from the "impersonal voice" of the RFC for a moment, I personally find that when writing generic code it is helpful to focus on the types and signatures, and come to the bounds later. Where clauses help to separate these distinctions. Naturally, your mileage may vary. - nmatsakis

In the case of an impl specifying the value for an associated type, the "important stuff" the value of the associated type.

### What about trait aliases, don't they distinguish where clause placement?

As currently implemented, trait aliases have two distinct possible placements for where clauses, which effectively distinguishes between a *where clause* (which must be proven true in order to use the alias) and an *implied bound* (which is part of what the alias expands to). One can write:

```rust
trait Foo<T: Debug> = Bar<T> + Baz<T>
```

in which case `where X: Foo<Y>` is only legal if `Y: Debug` is known from some other place. This is roughly equivalent to a trait like so:

```rust
trait Foo1<T: Debug>: Bar<T> + Baz<T> { }
```

The clause `where X: Foo1<Y>` is also only valid when `Y: Debug` is known. This is in contrast to the "supertraits" `Bar<Y>` and `Baz<Y>`, which are implied by `X: Foo1<Y>` ("supertraits" are also sometimes called "implied bounds").

Alternatively, one can include the where clause in the "value" of the trait alias like so:

```rust
trait ReverseEq<T> = where T: PartialEq<Self>;
```

In this case, `where X: ReverseEq<Y>` is equivalent to `Y: PartialEq<X>`. There is no "equivalent trait" for usage like this; the `T: PartialEq<Self>` effectively acts like a supertrait or implied bound.

Our decision was that this is a subtle distinction and that using the placement of the where clause was not a great way to make it.

### Is that trait alias syntax consistent with the rest of the language?

Not really. There are other places in the language that could benefit from a similar flexibility around implied bounds. For example, one could imagine wanting to have an associated type `T::MyType<Y>` where it is known that `Y: PartialEq<T::MyType<Y>>`, but this cannot be readily written with today's syntax:

```rust
trait MyTrait {
    type MyType<T>: PartialEq<T>;
    //              ^^^^^^^^^ not what we wanted
}
```

We decided that if we were going to offer that capability, we should find a way to offer it more generally, and hopefully with more clarity than putting the where clause before or after the `=`. As we have seen, where clauses for different kinds of items can be rather variable in their placement, so it is not clear that all users will recognize that distinction and understand it (many present in the meeting were surprised by the distinction as well).

Alternatively, the implied bounds proposal goes another way, turning most where clauses into implied bounds by default!

### Why do you even need where clauses in the impl anyway?

Given that the where clauses appear in the trait, you might wonder why they are needed in the impl anyway. After all, the impl could just assume that the trait bounds are met when checking the value of the associated type for validity, making the whole issue moot.

This would however be inconsistent with other sorts of items, which do require users to copy over the bounds from the trait. Furthermore, we have discussed the idea of allowing impls to relax the bounds from those described in the trait if they are not needed in the impl -- this came up most recently in the context of allowing impls to forego the `unsafe` keyword for `unsafe fn` declared in the trait if the fn in the impl body is completely safe. This could then even be relied upon by people invoking the method who know the precise impl they will be using.

In short, this might be a reasonable choice to make, but we should make it uniformly, and it shuts down the direction of using the lack of bounds in the impl as a kind of signal.

### Why not change type alias notation too?

Top-level type aliases currently parse with the where clause before the `=`:

```rust
type Foo<T> where T: Ord = T;
```

If you try that above example, however, you will find that you get a warning: this is because the `where T: Ord` is completely ignored! This is an implementation limitation in the way the current compiler eagerly expands type aliases. Moving the placement of where clauses actually gives us an opportunity to change this behavior without breaking any existing code, which is nice. It will however require some kind of opt-in (such as a `cargo fix` run) to migrate existing code that uses where clauses in the "ignored place" to the new format.

## Links for other discussion

* [Design document](https://rust-lang.github.io/generic-associated-types-initiative/design-discussions/where-the-where.html)
* [Design meeting minutes](https://github.com/rust-lang/lang-team/blob/master/design-meeting-minutes/2021-10-13-where-the-where.md)
* [Rust issue](https://github.com/rust-lang/rust/issues/89122)

lcnr:
One concern I have are generic constants, for example
```rust
const SIZE_OF<T>: usize = std::mem::size_of::<T>();

trait IsEqual: 'static {
    const TO_TYPE<U: 'static>: bool;
}
impl<T: 'static> IsEqual for T {
    const TO_TYPE<U: 'static>: bool = TypeId::of::<T>() == TypeId::of::<U>();
}
```
while both of these are currently not allowed in Rust, I personally quite strongly think that we should add them in the future and want to at least want to keep this option open. Can we unambiguously parse `where X` after an expression? I would assume yes but am not sure myself.
```rust
impl<T: 'static> IsEqual for T {
    const TO_TYPE<U>: bool = TypeId::of::<T>() == TypeId::of::<U>()
    where
        U: 'static;
}
```

nikomatsakis:
@rfcbot fcp cancel

nikomatsakis:
This shouldn't have been a T-compiler FCP!

nikomatsakis:
@rfcbot fcp merge

[This document is also published on the repo](https://rust-lang.github.io/generic-associated-types-initiative/design-discussions/where-the-where-1.html)

# Where does the where clause go? (Conclusion)

This is write-up of the conclusion to the [where does the where clause go?] question.

## Conclusion

### Where clauses in generic associated types comes after value/binding

Example:

```rust
trait MyTrait {
    type MyType<T>: Iterator
    where
        T: Ord;
}
    
impl MyTrait for MyOtherType {
    type MyType<T> = MyIterator
    where
        T: Ord;
}
```

Effectively the `= type` in the impl replaces the `: Bound` from the declaration with the value that has to meet those bounds.

### Later phase: type aliases

Type aliases will eventually be aligned to this syntax as well:

```rust
type MyType<T> = Vec<T> where T: Ord;
```

Currently, however, where clauses on type aliases are ignored, so we will not stabilize this new syntax until they have the meaning we want.

### Suggestions for users who put the where clause in the wrong place

We will parse where clauses in both positions and suggest to users that they be moved:


```rust
impl MyTrait for MyOtherType {
    type MyType<T>
    where
        T: Ord
    = MyIterator;
}
```

Gets an error with a suggested rewrite. The compiler proceeds "as if" the where clauses had been written after the `= Type`.

### Where clause syntax for trait aliases will have to be revisited

As described in the FAQ below, we currently support trait alias syntax like

```rust
trait ReverseEq<T> = where T: PartialEq<Self>;
```

This syntax will be removed. Although its capabilities could be useful, it is also quite confusing (the placement of the `where` is a subtle distinction), and not clearly needed. If we find that we do want it, we can add in a similar syntax later, but hopefully in a way that is more broadly consistent with the language.

## Discussion and FAQ

### But isn't it inconsistent with other trait items to put the where clauses *before* the `=`?

From one perspective, yes. One can view the value of an associated type as its "body", and the where clauses typically come before the "body" of an item. Put another way, typically you can "copy and paste" the impl and then add some text to the end of each item to specify its value: but with this syntax, you have to edit the "middle" of an associated type to specify its value.

The analogy of an associated type value to a function body, however, is somewhat flawed. The value of an associated type needs to be considered part of the "signature", or public facing, part of the impl. Consider: you can change the body of a function and be certain that your callees will still compile, but you cannot do the same for the value of an associated type.

Given this perspective, when you copy the associated type from the trait to the impl, you are "completing" the signature that was left incomplete by the trait. Moreover, to do so, you replace the `: Bound1 + Bound2` list (which constraints what kinds of types the impl might use) with a specific type, thus making it more specific.

### What about a more purely syntactic point-of-view? What is more consistent?

There is precedent that the placement of the where clause has less to do with the logical role that it plays and more to do with other factors, like whether it is followed by a braced list of items:

* With `struct Foo<T> where T: Ord { t: T }`, the "body" of the struct is its fields, and the where clause comes first.
* But we write `struct Foo<T>(T) where T: Ord`, thus placing the "body" (the fields `(T)`) first and the where clause second. Moreover, we initially implemented the grammar `struct Foo<T> where T: Ord (T)` but this was deemed so obviously confusing that it was [changed with little discussion](https://github.com/rust-lang/rust/issues/17904#issuecomment-58603749).

As further evidence that this syntax is inconsistent with Rust's traditions, placing the where clauses before the `= ty` makes it objectively hard to determine how to run rustfmt in a way that feels natural. rustfmt handles `where` by putting the `where` onto its own line, with one line per where clause. This structure works for existing Rust items because where clauses are always either following by nothing (tuple structs) or by a braced (`{}`) list of items (e.g., struct fields, fn body, etc). That opening `{` can therefore go on its own line. This `where` clause formatting does not work well with `=`.

The idea of having where clauses come at the "end" of the signature is also supported by the [original RFC](https://rust-lang.github.io/rfcs/0135-where.html#readability), which motivated where clauses in part by describing how they allow you to treat the precise bounds as "secondary" to the "important stuff":

> If I may step aside from the "impersonal voice" of the RFC for a moment, I personally find that when writing generic code it is helpful to focus on the types and signatures, and come to the bounds later. Where clauses help to separate these distinctions. Naturally, your mileage may vary. - nmatsakis

In the case of an impl specifying the value for an associated type, the "important stuff" the value of the associated type.

### What about trait aliases, don't they distinguish where clause placement?

As currently implemented, trait aliases have two distinct possible placements for where clauses, which effectively distinguishes between a *where clause* (which must be proven true in order to use the alias) and an *implied bound* (which is part of what the alias expands to). One can write:

```rust
trait Foo<T: Debug> = Bar<T> + Baz<T>
```

in which case `where X: Foo<Y>` is only legal if `Y: Debug` is known from some other place. This is roughly equivalent to a trait like so:

```rust
trait Foo1<T: Debug>: Bar<T> + Baz<T> { }
```

The clause `where X: Foo1<Y>` is also only valid when `Y: Debug` is known. This is in contrast to the "supertraits" `Bar<Y>` and `Baz<Y>`, which are implied by `X: Foo1<Y>` ("supertraits" are also sometimes called "implied bounds").

Alternatively, one can include the where clause in the "value" of the trait alias like so:

```rust
trait ReverseEq<T> = where T: PartialEq<Self>;
```

In this case, `where X: ReverseEq<Y>` is equivalent to `Y: PartialEq<X>`. There is no "equivalent trait" for usage like this; the `T: PartialEq<Self>` effectively acts like a supertrait or implied bound.

Our decision was that this is a subtle distinction and that using the placement of the where clause was not a great way to make it.

### Is that trait alias syntax consistent with the rest of the language?

Not really. There are other places in the language that could benefit from a similar flexibility around implied bounds. For example, one could imagine wanting to have an associated type `T::MyType<Y>` where it is known that `Y: PartialEq<T::MyType<Y>>`, but this cannot be readily written with today's syntax:

```rust
trait MyTrait {
    type MyType<T>: PartialEq<T>;
    //              ^^^^^^^^^ not what we wanted
}
```

We decided that if we were going to offer that capability, we should find a way to offer it more generally, and hopefully with more clarity than putting the where clause before or after the `=`. As we have seen, where clauses for different kinds of items can be rather variable in their placement, so it is not clear that all users will recognize that distinction and understand it (many present in the meeting were surprised by the distinction as well).

Alternatively, the implied bounds proposal goes another way, turning most where clauses into implied bounds by default!

### Why do you even need where clauses in the impl anyway?

Given that the where clauses appear in the trait, you might wonder why they are needed in the impl anyway. After all, the impl could just assume that the trait bounds are met when checking the value of the associated type for validity, making the whole issue moot.

This would however be inconsistent with other sorts of items, which do require users to copy over the bounds from the trait. Furthermore, we have discussed the idea of allowing impls to relax the bounds from those described in the trait if they are not needed in the impl -- this came up most recently in the context of allowing impls to forego the `unsafe` keyword for `unsafe fn` declared in the trait if the fn in the impl body is completely safe. This could then even be relied upon by people invoking the method who know the precise impl they will be using.

In short, this might be a reasonable choice to make, but we should make it uniformly, and it shuts down the direction of using the lack of bounds in the impl as a kind of signal.

### Why not change type alias notation too?

Top-level type aliases currently parse with the where clause before the `=`:

```rust
type Foo<T> where T: Ord = T;
```

If you try that above example, however, you will find that you get a warning: this is because the `where T: Ord` is completely ignored! This is an implementation limitation in the way the current compiler eagerly expands type aliases. Moving the placement of where clauses actually gives us an opportunity to change this behavior without breaking any existing code, which is nice. It will however require some kind of opt-in (such as a `cargo fix` run) to migrate existing code that uses where clauses in the "ignored place" to the new format.

## Links for other discussion

* [Design document](https://rust-lang.github.io/generic-associated-types-initiative/design-discussions/where-the-where.html)
* [Design meeting minutes](https://github.com/rust-lang/lang-team/blob/master/design-meeting-minutes/2021-10-13-where-the-where.md)
* [Rust issue](https://github.com/rust-lang/rust/issues/89122)

nikomatsakis:
> One concern I have are generic constants, for example
> 
> ```rust
> const SIZE_OF<T>: usize = std::mem::size_of::<T>();
> 
> trait IsEqual: 'static {
>     const TO<U: 'static>: bool;
> }
> impl<T: 'static> IsEqual for T {
>     const TO<U: 'static>: bool = TypeId::of::<T>() == TypeId::of::<U>();
> }
> ```
> 
> while both of these are currently not allowed in Rust, I personally quite strongly believe that we should add them in the future and want to at least want to keep this option open. Can we unambiguously parse `where X` after an expression? I would assume yes but am not sure myself.

Ah, good call! I'm glad you brought that up, we didn't discuss associated constants in the meeting. I'm not worried about parsing ambiguities, the `where` keyword is a pretty strong signal that we've reached the end of the expression (it's not presently used anywhere else in Rust that I know of). 

I'm not 100% sure where I think `where` should go relative to the `=` though. It seems to depend a bit on the size of the constant expression.

petrochenkov:
@nikomatsakis 
>I'm not 100% sure where I think `where` should go relative to the `=` though. It seems to depend a bit on the size of the constant expression.

Do you remember the Centril's proposal to abbreviate `fn foo() { result }` to `fn foo() = result;`?
That is a very reasonable thing to do if the body is short, or the body is some kind of `match` or `loop`, and I'd still want to have it accepted one day.
(I can't find the relevant RFC PR/issue, unfortunately.)
```rust
fn foo() -> u8 = match {
    A => 1,
    B => 2,
    C => 3,
    _ => 999,
};
```
What is the preferable position for `where` in this case?

petrochenkov:
(Self-assigning, because I have some comments about the implementation as well.)

camelid:
>  (I can't find the relevant RFC PR/issue, unfortunately.)

I believe this is it: https://github.com/Centril/rfcs/pull/17

petrochenkov:
Something is wrong with the FCP here, it's still active and assigned to the compiler team.
@rfcbot fcp cancel

petrochenkov:
@rfcbot fcp cancel

petrochenkov:
@rfcbot fcp cancel

petrochenkov:
@rfcbot fcp merge

Details: https://github.com/rust-lang/rust/pull/90076#issuecomment-962195563

petrochenkov:
(Also blocked on #90671.)

nikomatsakis:
Let's try this again... 3rd time's the charm...

@rfcbot fcp merge

I propose that we...

* move the `where` clause for associated types to after the `=`;
* parse `where` clauses after `=` type aliases but feature gate them until they "work properly", for some definition of properly yet to be determined;
* deprecate the `= where` syntax for trait aliases and perhaps just remove it without replacement (but we can settle that separately).

Below is a write-up of the conclusion to the [where does the where clause go?] question. [This document is also published on the repo](https://rust-lang.github.io/generic-associated-types-initiative/design-discussions/where-the-where-1.html)

<details>
<summary>Detailed write-up</summary>

## Conclusion

### Where clauses in generic associated types comes after value/binding

Example:

```rust
trait MyTrait {
    type MyType<T>: Iterator
    where
        T: Ord;
}
    
impl MyTrait for MyOtherType {
    type MyType<T> = MyIterator
    where
        T: Ord;
}
```

Effectively the `= type` in the impl replaces the `: Bound` from the declaration with the value that has to meet those bounds.

### Later phase: type aliases

Type aliases will eventually be aligned to this syntax as well:

```rust
type MyType<T> = Vec<T> where T: Ord;
```

Currently, however, where clauses on type aliases are ignored, so we will not stabilize this new syntax until they have the meaning we want.

### Suggestions for users who put the where clause in the wrong place

We will parse where clauses in both positions and suggest to users that they be moved:


```rust
impl MyTrait for MyOtherType {
    type MyType<T>
    where
        T: Ord
    = MyIterator;
}
```

Gets an error with a suggested rewrite. The compiler proceeds "as if" the where clauses had been written after the `= Type`.

### Where clause syntax for trait aliases will have to be revisited

As described in the FAQ below, we currently support trait alias syntax like

```rust
trait ReverseEq<T> = where T: PartialEq<Self>;
```

This syntax will be removed. Although its capabilities could be useful, it is also quite confusing (the placement of the `where` is a subtle distinction), and not clearly needed. If we find that we do want it, we can add in a similar syntax later, but hopefully in a way that is more broadly consistent with the language.

## Discussion and FAQ

### But isn't it inconsistent with other trait items to put the where clauses *before* the `=`?

From one perspective, yes. One can view the value of an associated type as its "body", and the where clauses typically come before the "body" of an item. Put another way, typically you can "copy and paste" the impl and then add some text to the end of each item to specify its value: but with this syntax, you have to edit the "middle" of an associated type to specify its value.

The analogy of an associated type value to a function body, however, is somewhat flawed. The value of an associated type needs to be considered part of the "signature", or public facing, part of the impl. Consider: you can change the body of a function and be certain that your callees will still compile, but you cannot do the same for the value of an associated type.

Given this perspective, when you copy the associated type from the trait to the impl, you are "completing" the signature that was left incomplete by the trait. Moreover, to do so, you replace the `: Bound1 + Bound2` list (which constraints what kinds of types the impl might use) with a specific type, thus making it more specific.

### What about a more purely syntactic point-of-view? What is more consistent?

There is precedent that the placement of the where clause has less to do with the logical role that it plays and more to do with other factors, like whether it is followed by a braced list of items:

* With `struct Foo<T> where T: Ord { t: T }`, the "body" of the struct is its fields, and the where clause comes first.
* But we write `struct Foo<T>(T) where T: Ord`, thus placing the "body" (the fields `(T)`) first and the where clause second. Moreover, we initially implemented the grammar `struct Foo<T> where T: Ord (T)` but this was deemed so obviously confusing that it was [changed with little discussion](https://github.com/rust-lang/rust/issues/17904#issuecomment-58603749).

As further evidence that this syntax is inconsistent with Rust's traditions, placing the where clauses before the `= ty` makes it objectively hard to determine how to run rustfmt in a way that feels natural. rustfmt handles `where` by putting the `where` onto its own line, with one line per where clause. This structure works for existing Rust items because where clauses are always either following by nothing (tuple structs) or by a braced (`{}`) list of items (e.g., struct fields, fn body, etc). That opening `{` can therefore go on its own line. This `where` clause formatting does not work well with `=`.

The idea of having where clauses come at the "end" of the signature is also supported by the [original RFC](https://rust-lang.github.io/rfcs/0135-where.html#readability), which motivated where clauses in part by describing how they allow you to treat the precise bounds as "secondary" to the "important stuff":

> If I may step aside from the "impersonal voice" of the RFC for a moment, I personally find that when writing generic code it is helpful to focus on the types and signatures, and come to the bounds later. Where clauses help to separate these distinctions. Naturally, your mileage may vary. - nmatsakis

In the case of an impl specifying the value for an associated type, the "important stuff" the value of the associated type.

### What about trait aliases, don't they distinguish where clause placement?

As currently implemented, trait aliases have two distinct possible placements for where clauses, which effectively distinguishes between a *where clause* (which must be proven true in order to use the alias) and an *implied bound* (which is part of what the alias expands to). One can write:

```rust
trait Foo<T: Debug> = Bar<T> + Baz<T>
```

in which case `where X: Foo<Y>` is only legal if `Y: Debug` is known from some other place. This is roughly equivalent to a trait like so:

```rust
trait Foo1<T: Debug>: Bar<T> + Baz<T> { }
```

The clause `where X: Foo1<Y>` is also only valid when `Y: Debug` is known. This is in contrast to the "supertraits" `Bar<Y>` and `Baz<Y>`, which are implied by `X: Foo1<Y>` ("supertraits" are also sometimes called "implied bounds").

Alternatively, one can include the where clause in the "value" of the trait alias like so:

```rust
trait ReverseEq<T> = where T: PartialEq<Self>;
```

In this case, `where X: ReverseEq<Y>` is equivalent to `Y: PartialEq<X>`. There is no "equivalent trait" for usage like this; the `T: PartialEq<Self>` effectively acts like a supertrait or implied bound.

Our decision was that this is a subtle distinction and that using the placement of the where clause was not a great way to make it.

### Is that trait alias syntax consistent with the rest of the language?

Not really. There are other places in the language that could benefit from a similar flexibility around implied bounds. For example, one could imagine wanting to have an associated type `T::MyType<Y>` where it is known that `Y: PartialEq<T::MyType<Y>>`, but this cannot be readily written with today's syntax:

```rust
trait MyTrait {
    type MyType<T>: PartialEq<T>;
    //              ^^^^^^^^^ not what we wanted
}
```

We decided that if we were going to offer that capability, we should find a way to offer it more generally, and hopefully with more clarity than putting the where clause before or after the `=`. As we have seen, where clauses for different kinds of items can be rather variable in their placement, so it is not clear that all users will recognize that distinction and understand it (many present in the meeting were surprised by the distinction as well).

Alternatively, the implied bounds proposal goes another way, turning most where clauses into implied bounds by default!

### Why do you even need where clauses in the impl anyway?

Given that the where clauses appear in the trait, you might wonder why they are needed in the impl anyway. After all, the impl could just assume that the trait bounds are met when checking the value of the associated type for validity, making the whole issue moot.

This would however be inconsistent with other sorts of items, which do require users to copy over the bounds from the trait. Furthermore, we have discussed the idea of allowing impls to relax the bounds from those described in the trait if they are not needed in the impl -- this came up most recently in the context of allowing impls to forego the `unsafe` keyword for `unsafe fn` declared in the trait if the fn in the impl body is completely safe. This could then even be relied upon by people invoking the method who know the precise impl they will be using.

In short, this might be a reasonable choice to make, but we should make it uniformly, and it shuts down the direction of using the lack of bounds in the impl as a kind of signal.

### Why not change type alias notation too?

Top-level type aliases currently parse with the where clause before the `=`:

```rust
type Foo<T> where T: Ord = T;
```

If you try that above example, however, you will find that you get a warning: this is because the `where T: Ord` is completely ignored! This is an implementation limitation in the way the current compiler eagerly expands type aliases. Moving the placement of where clauses actually gives us an opportunity to change this behavior without breaking any existing code, which is nice. It will however require some kind of opt-in (such as a `cargo fix` run) to migrate existing code that uses where clauses in the "ignored place" to the new format.

## Links for other discussion

* [Design document](https://rust-lang.github.io/generic-associated-types-initiative/design-discussions/where-the-where.html)
* [Design meeting minutes](https://github.com/rust-lang/lang-team/blob/master/design-meeting-minutes/2021-10-13-where-the-where.md)
* [Rust issue](https://github.com/rust-lang/rust/issues/89122)

</details>

calebcartwright:
I'm not really sure where the proposal is being actively tracked so I wanted to cross post my comment from https://github.com/rust-lang/rust/issues/89122#issuecomment-962723502 related to formatting.

It's not a concern with the syntax changes, but the formatting pieces need deeper discussion, so for now rustfmt should leave aliases with clauses in the new position alone until those discussions are finalized.

For example, formatting will have to account for non-single-line cases, such as where the rhs cannot fit on the same line, either due to the combined length or because the type itself needs to be formatted across multiple lines.

Worth noting those cases could result in the clause being shifted quite a few lines down from the lhs contents and potentially impact readability, but it's the type of thing we need to hammer out in the style guide/rustfmt before rustfmt starts doing any reformatting of these aliases.

dhardy:
Why have arguments moved from #89122 to here?

calebcartwright:
> Why have arguments moved from #89122 to here?
> 
> Personally I find most arguments to be subjective and don't care much, but @calebcartwright does make a good point.

I'm just trying to cover all bases because this appears on the verge of moving ahead, and I'm trying to play catch up in the final minute since this (currently) includes changes to rustfmt that will potentially create some formatting headaches.

I don't really have any opinions nor much to offer on the actual proposed syntax changes.

nikomatsakis:
@calebcartwright I might be missing something here but I don't think this change introduces a lot of complexity to rustfmt. This is what I imagine would happen:

### If user wrote a top-level type alias with where after `=`..

```rust
type Foo<T> = u32 where T: Debug;
```

This would be an error today. It will eventually be made to work, but not until we figure out how to give it the semantics we expect. It will parse, however, so rustfmt *could* rewrite it to the obvious formatting:

```rust
type Foo<T> = u32
where
    T: Debug;
```

### If user wrote a top-level type alias with where before `=`..

```rust
type Foo<T> where T: Debug = u32;
```

No change. Do whatever rustfmt does now. This may eventually be deprecated, but it won't be removed; it would not make sense for rustfmt to rewrite it because there would be a semantic shift involved (as you noted). Most likely, it would mean the where clauses go from not being enforced to being enforced.

### If user wrote a generic associated type with where after `=`...

```rust
impl ... {
    type Foo<T> = u32 where T: Debug;
}
```

This would be formatted like so, I expect. This seems like the obvious way to do things, but let me know if you think we should write a separate rustfmt RFC:

```rust
impl ... {
    type Foo<T> = u32
    where
        T: Debug;
}
```

### If user wrote a generic associated type with where before `=`...

This would be a compilation error. The compiler would suggest a rewrite. rustfmt would not necessarily be expected to do anything here, although I think it would be **useful** if rustfmt would do the rewrite on the user's behalf. I expect that the parser will parse this result, but it would fail ast validation, so that seems within scope for rustfmt. Note that, unlike in the case of free-standing type aliases, this is not a "semantics changing operation". We are going from "compilation error" to "successful compile":

```rust
impl ... {
    type Foo<T> where T: Debug = u32;
}
```

becomes

```rust
impl ... {
    type Foo<T> = u32
    where
        T: Debug;
}
```

nikomatsakis:
@jackh726 and I spent a while talking today about both the const generics syntax and the possible `fn()=` syntax. One of our conclusions was that *semantically*, the two cases are quite different:

* The value of an associated constant, like the value of a type alias or associated type, is conceptually part of the "signature". By "part of the signature" I mean that if you change the value of a constant, you can cause other code to stop compiling (even though the constant compiles just fine). The general rule seems to be that the "where" clause goes at the end of the signature, as the "least important part".
* The body of a function, however, is not. 

There does seem to be a general rule in Rust that "where clauses come at the end of the signature". This holds also for tuple structs, since the field types of a tuple struct are part of its signature, albeit a part that is private to the current module. This would lead one to expect the following:

```rust
type Foo<T> = Value where T: Ord;
const FOO<T>: Type = value where T: Ord;
fn foo<T>() -> Type where T: Ord = value;
```

There is an interesting asymmetry between associated types and constants: for associated types, we don't repeat the bounds from the trait, but we *do* repeat the type for an associated constant. And yet both the bounds and types play a very similar role of constraining the possible values for that associated item. My hunch is that we eventually want to accept *optional* bounds and an *optional* type -- if either is omitted, then they are taken from the trait definition. (This would permit, for example, one to have an associated type that has *tighter* bounds than the trait definition: this could be important if one had a `default` associated type with specialization.)

It is true that having the `=` sign come at the *end* of a function but the *middle* of a constant might be hard for people to remember. On the other hand, there seems to be a clear grouping:

* Associated types
* Tuple structs

nikomatsakis:
@jackh726 and I spent a while talking today about both the const generics syntax and the possible `fn()=` syntax. The TL;DR is that we still feel good about this change, but that we think I should update the write-up to include discussion of associated constants and a possible `fn foo() = bar` syntax.

### Where clauses come at the end of the signature

One of our conclusions was that *semantically*, the two cases are quite different:

* The value of an associated constant, like the value of a type alias or associated type, is conceptually part of the "signature". By "part of the signature" I mean that if you change the value of a constant, you can cause other code to stop compiling (even though the constant compiles just fine). The general rule seems to be that the "where" clause goes at the end of the signature, as the "least important part".
* The body of a function, however, is not. 

There does seem to be a general rule in Rust that "where clauses come at the end of the signature". This holds also for tuple structs, since the field types of a tuple struct are part of its signature, albeit a part that is private to the current module. This would lead one to expect the following:

```rust
type Foo<T> = Value where T: Ord;
const FOO<T>: Type = value where T: Ord;
fn foo<T>() -> Type where T: Ord = value;
```

There is an interesting asymmetry between associated types and constants: for associated types, we don't repeat the bounds from the trait, but we *do* repeat the type for an associated constant. And yet both the bounds and types play a very similar role of constraining the possible values for that associated item. My hunch is that we eventually want to accept *optional* bounds and an *optional* type -- if either is omitted, then they are taken from the trait definition. (This would permit, for example, one to have an associated type that has *tighter* bounds than the trait definition: this could be important if one had a `default` associated type with specialization.)

### What we have to decide *today*: associated types, constants

So, putting it all together, there seems to be a clear grouping of things where you want the where clauses at the end, all of which we have in the language now in at least some form:

* Associated types
* Associated constants
* Tuple structs

At the end of the day, I think that *today* we should make the right decision for these things, and I think that this decision is to put the where clauses *after* the `=`, for the various reasons enumerated.

### On muscle memory and surprise

One thing I have noticed is that, for myself and I think many other users, having the where clauses come *before* the `=` in associated type values is a constant source of surprise. I have "muscle memory" for specifying the value of an associated type that consists of:

1. Copy the `type Foo: Bar;` from the trait definition.
2. Replace `: Bar` with `= Baz`.

With the current `=` placement, that muscle memory leads me astray. I have noticed that I am constantly trying to replace `: Bounds where WC` with `= Type where WC` but getting syntax errors.

### For the future: on functions

If and when we decide to add `fn()=` syntax, we'll have to decide how to manage the placement of where clauses. I think there is a solid argument for having the where clauses in a `fn` come **before** the `=`, as I said above, even though that is different from associated types/constants. In short, it is different for a good reason. 

It may well be, though, that having where clauses come in a different place with respect to the `=` for functions as opposed to other items confuses a lot of people. I'm not sure, but I think there's a good chance it won't. Think again of the muscle memory. To implement a function today, you:

1. Copy the `fn foo<T>();` from the trait definition.
2. Replace the `;` with the body of the function `{ ... }`

You would now do the same, but instead of replacing `;`, you would insert `= value`.

calebcartwright:
> @calebcartwright I might be missing something here but I don't think this change introduces a lot of complexity to rustfmt. 

Thanks @nikomatsakis 

It's more so about process as opposed to the specifics of the formatting/difficulty of implementation in rustfmt. My ask is simply that for now rustfmt not be modified to reformat aliases with the updated syntax (that can be accomplished by returning `None` on the relevant code paths) and that we have a chance for the formatting to first be discussed in the [fmt-rfcs repo](https://github.com/rust-dev-tools/fmt-rfcs) and then added to the Style Guide. 

As for my rationale:

Regardless of how simple and straightforward it may seem at first blush, there's (somewhat) edgier cases we have to account for which I don't think anyone but rustfmt necessarily cares about.

Examples of some cases with the old position where I'd have some questions around how they should be formatted in the new position:

```rust
impl<T> Iterable for Vec<T> {
    // with new syntax lhs and rhs wouldn't fit on the same line
    type SomeRatherAbsurdlyLongIdentBecauseReasonsAndFormattingFun<'a>
    where
        Self: 'a,
    = <&'a [T] as IntoIterator>::IntoIter;
}

impl ... {
    // what will we want to do with a type that can't fit on one line?
    type TRef<'a>
    where
        <North as Foo>::ABCDEFGH: 'a,
        <South as Foo>::ABCDEFGH: 'a,
        <East as Foo>::ABCDEFGH: 'a,
        <West as Foo>::ABCDEFGH: 'a,
        <NorthEast as Foo>::ABCDEFGH: 'a,
    = Any<
        &'a North::ABCDEFGH,
        &'a South::ABCDEFGH,
        &'a East::ABCDEFGH,
        &'a West::ABCDEFGH,
        &'a NorthEast::ABCDEFGH,
    >;
}

```

Those are solely illustrative, and admittedly heavily contrived, but hopefully paint a picture of completely plausible scenarios where due to idents, indentation levels, comments, whatever, rustfmt couldn't have a single line with the lhs and rhs preceding the clause in the new position. Additionally with new syntax we also have to consider interactions with config options.

This why I typically prefer that opinionated reformatting behavior not be incorporated with new syntax/syntax changes, and the approach I'm requesting here is the same tact that was taken for the new let-else statements. Once rustfmt starts reformatting constructs, any additional updates/changes have the potential to introduce breaking formatting updates which is something we strive to avoid.

More generally beyond some of the aforementioned TBDs, I want to make sure we don't put ourselves in a position where the Style Guide explicitly says clauses and aliases must be formatted like X, while in some cases rustfmt violates that and formats them like Y due to an as-of-yet unwritten set rules.

I also want to ensure everyone has a chance to weigh in on the formatting since we always communicate to users, especially those frustrated with rustfmt output, that formatting decisions in the Style Guide undergo a similar type RFC process in the fmt-rfcs repo.

petrochenkov:
I still think doing this is a mistake.

For the sake of a minor stylistic improvement this proposals starts multiplying entities and creating new issues and questions.
```rust
type Alias
    where Long: WhereClause
        = TypeValuse;
```
Already works and looks fine when formatted even if the where clause is long, no extra action is required and people are free to occupy themselves with something more useful.

This part
>the value of a type alias or associated type, is conceptually part of the "signature"

is highly unobvious and unintuitive.
With inconsistent rules like this nobody is going to remember where the `where` is written in each specific case.

The existing tuple struct case `struct S<T, U>(T, U) where T: Trait;` is unfortunate, but it was introduced due to a syntactic ambiguity caused by the parentheses, not due to the "end of signature" reasoning.
It wasn't widely discussed when introduced, because it's so old that almost no changes were widely discussed back then.
Even if we cannot remove inconsistencies like this, we should at least not introduce new ones.

joshtriplett:
@rfcbot concern further-thought

I'd like to give @petrochenkov's argument further thought, because it gives me serious pause, and I don't feel comfortable going forward with this right now.

nikomatsakis:
I discussed this some with @jackh726 today. He said he was feeling on the fence. I will say that I don't feel it's an open-and-shut case, though I still lean towards changing the syntax. Nonetheless, I had a few examples come up today which pushed in (slightly) opposite directions.

First, take a look at [this commit](https://github.com/vandenheuvel/relp/commit/aa471b3272cb53f8f497217585f89599cc9343c3). I was struck by two things:

1. I continue to feel that the `: Bounds` and `= Type` play analogous roles in my brain, and I was surprised by the way that the `where Self: 'a` appeared at the end of the `type` items in trait definitions but at the beginning of the `type` items in `impl` items.
2. On the other hand, I admit that simple cases like `type Cost<'a> where Self: 'a = MP::Cost<'a>;` seem natural enough.

Moving on, [this playground] included a snippet with `=` sign formatted on the same starting point as `where`:

```rust
    type Output<'a>
    where
        Self: 'a,
    = <<T as Deref>::Target as GetOutput>::Output<'a>;
```

Overall my reaction was "ugly but clear enough".

No conclusions, just observations. I do think the best thing at this stage would be for folks to try writing code that uses GATs and reformatting it to the distinct syles.

JakobDegen:
Sorry if I'm not caught up here (I tried to do as much reading as I could before responding, but I may have missed the answer to this question). What is the justification for allowing where clauses to be written *at all* when the trait is being implemented (and the type "assigned" or whatever you wanna call it)? Based on my understanding of GATs, implementers of the trait may not strengthen or weaken the where clauses listed in the trait definition. If I am not mistaken, this would then be unique to Rust in that it is the only place in which where clauses are allowed to be written, but are effectively not allowed to have any meaning.

dtolnay:
@JakobDegen ["Why do you even need where clauses in the impl anyway?"](https://github.com/rust-lang/generic-associated-types-initiative/blob/86716c4c6eda59f53b89415e2653c777b6d39a8a/design-discussions/where-the-where-1.md#why-do-you-even-need-where-clauses-in-the-impl-anyway)

JakobDegen:
Thank you! To be honest, I'm not terribly convinced by that argument, but this being unnecessarily irreversible is a good point and I am happy to see that this is being acknowledged as a direction in which we might head in the future.

nikomatsakis:
@joshtriplett and I had a conversation a few days back and he indicated that he was happy so long as we have good diagnostics for both variants, in short, with an actionable rustfix. I also feel (and he did not disagree) that there ought to be one "officially preferred" location. I further feel (and he did not disagree) that the preferred location ought to be *after* the `=`, for the reasons I've previously stated.

That said, we were debating between two variants:

* accept both variants and concatenate them, but issue a lint warning (with rustfix) to rewrite where clauses before the `=` into where clauses after the `=`
* produce a hard error for lint warnings before the `=` with actionable rustfix

To be honest, I somewhat lean towards the first. I don't like "more than one way to do it for no particularly good reason", but the lint warning addresses that concern. Furthermore, parsing both variants means whichever people more naturally do it, the code will compile. The fact that both are legal Rust also means we can have rustfmt silently normalize the non-deprecated variant. If there *were* some *particular* use case (e.g. with macros) where it's more convenient for the where clause to come before the `=`, that is still possible. It's hard for me to see the downside in practical terms.

Therefore, I'm curious to hear what folks thing about that idea:

**Accept both and concatenate, but have "before the `=`" be deprecated.**

jackh726:
I think that's a pretty decent plan.

There's two benefits to this approach:
1) Any existing code with GATs will continue to compile. Yes, it's an unstable feature, but it would be nice to not make it a chore to update from a current nightly. 
2) It would give us real data for which variant people prefer (with a bias for after the equals, of course). This could be used to later inform hard-comitting to one or the other at an edition.

calebcartwright:
> The fact that both are legal Rust also means we can have rustfmt silently normalize the non-deprecated variant.

If this is indeed the variant/route selected, would folks be open to doing this in a phased approach wherein for some time rustfmt would continue to format clauses in the existing position as dictated by the style guide, and at some future date transition to performing the transformation to the new position once we've had time to codify all the rules (including the fun edge cases)?

nikomatsakis:
@calebcartwright I would be fine with that. We would issue a rustfix regardless I think.

jackh726:
@joshtriplett Is there any further concern for the FCP? Or can that be resolved?

joshtriplett:
No further concern, assuming we parse both and either lint or hard error on one with a rustfix towards the other.

@rfcbot resolve further-thought

apiraino:
removing `T-compiler` label which I have probably mistakenly added (by reading [this comment](https://github.com/rust-lang/rust/pull/90076#issuecomment-962195046))
@rustbot label -T-compiler

bstrie:
Well, if you're thinking of allowing flexible `where` positions in every context, I've always thought it might look nice to put `where` *before* everything else...

```rust
where K: PartialOrd
type StringMap<K> = BTreeMap<K, String>;
```

Less indentation, and it solves the inconsistency with where to declare the type params on `impl` blocks. :)

```rust
where 'a, T
impl Foo for Bar<'a, T> {
```

Finally, I can grep for `impl Foo`!

jackh726:
I'm going through now to look at what diagnostics might degrade. I'll add tests if I find any, rebase, then this should be good to go after review.

jackh726:
This is ready for review.

I wasn't able to find any current tests where a diagnostic might be degraded with both where clauses. This is probably likely due to incomplete coverage of relevant diagnostics with GATs. Imo, it's not worth the effort to try to come up with test cases here for that.

compiler/rustc_ast/src/ast.rs
```diff
@@ -2663,10 +2663,23 @@ pub struct Trait {
     pub items: Vec<P<AssocItem>>,
 }
 
+/// The location of a where clause on a `TyAlias` (`Span`) and whether there was
+/// a `where` keyword (`bool`). This is split out from `WhereClause`, since there
+/// are two locations for where clause on type aliases, but their predicates
+/// are concatenated together.
```
nikomatsakis, line 2669: Rust examples might help here

compiler/rustc_ast_passes/src/ast_validation.rs
```diff
@@ -122,6 +122,40 @@ impl<'a> AstValidator<'a> {
...
+            let mut first = true;
+            for p in before_predicates.iter() {
+                if !first {
+                    state.word_space(",");
+                }
+                first = false;
+                state.print_where_predicate(p);
+            }
+            let suggestion = state.s.eof();
+            sess.struct_span_err(where_clauses.0.1, "where clause not allowed here")
```
nikomatsakis: hard error?

nikomatsakis: I expected a lint

jackh726: This gets changed to a lint in a later commit.

src/test/ui/generic-associated-types/issue-84931.stderr
```diff
@@ -2,9 +2,9 @@ error[E0309]: the parameter type `T` may not live long enough
   --> $DIR/issue-84931.rs:15:21
    |
 LL |     type Item<'a> = &'a mut T;
-   |                  -  ^^^^^^^^^ ...so that the reference type `&'a mut T` does not outlive the data it points at
-   |                  |
-   |                  help: consider adding a where clause: `where T: 'a`
+   |                     ^^^^^^^^^- help: consider adding a where clause: `where T: 'a`
```
nikomatsakis, line 5: nice

src/test/ui/parser/type-alias-where.stderr
```diff
@@ -1,40 +1,34 @@
...
-  --> $DIR/type-alias-where.rs:24:38
+warning: where clause not allowed here
+  --> $DIR/type-alias-where.rs:22:16
    |
-LL |     type Assoc2 where u32: Copy = () where i32: Copy;
-   |                                -     ^^^^^^^^^^^^^^^
-   |                                |
-   |                                help: move it here: `, i32: Copy`
+LL |     type Assoc where u32: Copy = ();
+   |                ^^^^^^^^^^^^^^^     - help: move it here: `where u32: Copy`
```
nikomatsakis, line 17: can we add a `.fixed` test?

nikomatsakis: I think maybe you have to add `// run-rustfix` or something like this, you may alsonewed to factor this case out into its own test

nikomatsakis (changes requested):
Looking good, some more nits and suggestions.

compiler/rustc_ast/src/ast.rs
```diff
@@ -2662,10 +2662,35 @@ pub struct Trait {
...
+///   type Assoc<'a, 'b> where Self: 'a, Self: 'b;
+/// }
+/// impl Foo for () {
+///   type Assoc<'a, 'b> where Self: 'a = () where Self: 'b;
+///   //                 ^^^^^^^^^^^^^^ first where clause
+///   //                                     ^^^^^^^^^^^^^^ second where clause
+/// }
+/// ```
+#[derive(Copy, Clone, Encodable, Decodable, Debug, Default)]
+pub struct TyAliasWhereClause(pub bool, pub Span);
```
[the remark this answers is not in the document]
nikomatsakis: Can you extend the comment to include that?

nikomatsakis (changes requested):
r=me with extended comment

compiler/rustc_ast/src/ast.rs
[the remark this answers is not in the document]
jackh726: Added "If there is no where clause, then this is `false` with `DUMMY_SP`."

jackh726:
@bors rollup=never

just in case

jackh726:
Let's try this again.

@bors r=nikomatsakis

Issue #87852, opened 2021-08-07
Tracking Issue for const_num_from_num


usbalbin:
Feature gate: `#![feature(const_num_from_num)]`

This is a tracking issue for constifying the `impl From<U> for T` and `impl TryFrom<U> for T` for conversions between the basic built in numeric types(both `T` and `U` are one of `f32`, `f64`, `iX`, `uX`). It also constifies the corresponding `From` conversions for `NonZeroUX` and `NonZeroIX` and from `NonZeroUX`/`NonZeroIX` to `uX`/`iX` of same type and signedness. Additionally `AtomicUX` from `uX` and `AtomicIX` from `iX`, this is however only for types of same size. 

*Note that all of this does *not* add any new implementations, it only constifies the existing ones.*

In short, this will enable things like below without having to use `as` casts

```rust
const FOO_U32: u32 = u32::from(bar_returing_u16()); // u16 -> u32
const BAZ_F32: f32 = f32::from(bar_returing_u8());  // u8 -> f32

const FROM_NONZERO_U8: u8 = u8::from(bar_returing_nonzero_u8()); // NonZeroU8 -> u8
const NONZERO_FROM_SMALLER: NonZeroU32 = NonZeroU32::from(bar_returing_nonzero_u8()); // NonZeroU8 -> NonZeroU32

const ATOMIC_FROM_U8: AtomicU8 = AtomicU8::from(bar_returing_u8()); // u8 -> AtomicU8
```

### Public API

```rust
# TODO
```

### Steps / History

- [x] Implementation of conversions of primitive numbers: #86840
- [x] Implementation of conversions of atomic and non-zero integers: #90077
- [ ] Final comment period (FCP)
- [ ] Stabilization PR

### Unresolved Questions

- None yet.

woppopo:
`From` implementations of atomic and non-zero integers is now const.

chrisbouchard:
Is this still feature a thing? Rust doesn't recognize the feature name when I enable it on the latest nightly.

```
error[E0635]: unknown feature `const_num_from_num`
 --> src/lib.rs:1:12
  |
1 | #![feature(const_num_from_num)]
  |            ^^^^^^^^^^^^^^^^^^
```

But also numeric `From` and `TryFrom`s are still non-const.

[Playground Link](https://play.rust-lang.org/?version=nightly&mode=debug&edition=2021&gist=3ee231f232402498797c21f2e6edeb3b)

oli-obk:
[...]

PR #90077, opened 2021-10-20
Make `From` impls of NonZero integer const.

lilasta:
I also changed the feature gate added to `From` impls of Atomic integer to `const_num_from_num` from `const_convert`.

Tracking issue: #87852


oli-obk:
we are putting adding more const trait impls on hold until T-libs had a chance to discuss it. for more details see https://github.com/rust-lang/rust/pull/89876

PR #90079, opened 2021-10-20
enable `i8mm` target feature on aarch64 and arm

SparrowLii:
As in https://github.com/rust-lang/stdarch/issues/1233, `i8mm` needs to be turned on to support the implementation of `vmmla` and `vusmmla`neon instructions in stdarch.
r? @Amanieu


SparrowLii:
It seems that `i8mm` is still unavailable. In this case, we may not be able to support commands such as `vmmla` in stdarch for the time being.

hkratz:
> It seems that `i8mm` is still unavailable. In this case, we may not be able to support commands such as vmmla in stdarch for the time being.

It seems that LLVM 11 has support for i8mm, so once #90062 has landed this can go forward.

Issue #86302, opened 2021-06-14
Tracking Issue for const_ptr_write


usbalbin:
Feature gate: `#![feature(const_ptr_write)]`

This is a tracking issue for making the functions `ptr::write` and `ptr::write_unaligned`, and the same methods on `*mut T`, `const fn`.

### Public API

```rust
mod ptr {
    pub const unsafe fn write<T>(dst: *mut T, src: T);
    pub const unsafe fn write_unaligned<T>(dst: *mut T, src: T);
    pub const unsafe fn write_bytes<T>(dst: *mut T, val: u8, count: usize);
}

impl<T> *mut T {
    pub const unsafe fn write(self, val: T);
    pub const unsafe fn write_unaligned(self, val: T);
    pub const unsafe fn write_bytes(self, val: u8, count: usize);
}

impl <T> NonNull<T> {
    pub const unsafe fn write(self, val: T);
    pub const unsafe fn write_unaligned(self, val: T);
    pub const unsafe fn write_bytes(self, val: u8, count: usize);
}
```

### Steps / History

- [x] Implementation: #81167
- [x] Reverted: #86003
- [x] Re-constified: #86295
- [x] `core::intrinsics::write_bytes` aka `ptr::write_bytes` - #90081 
- [x] `(*mut T)::write_bytes` - #91824
- [x] blocked on https://github.com/rust-lang/rust/issues/57349
- [x] `NonNull` methods stabilized, `const`ness moved to this gate https://github.com/rust-lang/rust/pull/124498
- [ ] Final commenting period (FCP)
- [ ] Stabilization PR

#### Related
* const_ptr_read #80377

### Unresolved Questions

- None yet.

woppopo:
`core::intrinsics::write_bytes` and `(*mut T)::write_bytes` have been constified: #90081 #91824

SUPERCILEX:
How does one initiate the FCP? Or are there any blockers?

c410-f3r:
AFAICT, there are no blockers so this feature can probably be stabilized.

RalfJung:
How does stabilizing this help? You can't create mutable pointers on stable anyway...

RalfJung:
IOW, this is blocked on https://github.com/rust-lang/rust/issues/57349 getting stabilized first.

wyfo:
> How does stabilizing this help? You can't create mutable pointers in const on stable anyway...

[...]

PR #90081, opened 2021-10-20
Make `intrinsics::write_bytes` const

lilasta:
This is required to constify `MaybeUninit::zeroed` and `(*mut T)::write_bytes`.

Tracking issue: #86302

### DIFF
--- a/library/core/src/intrinsics.rs
+++ b/library/core/src/intrinsics.rs
@@ -2242,13 +2242,29 @@ pub const unsafe fn copy<T>(src: *const T, dst: *mut T, count: usize) {
 /// assert_eq!(*v, 42);
 /// ```
 #[stable(feature = "rust1", since = "1.0.0")]
+#[rustc_const_unstable(feature = "const_ptr_write", issue = "86302")]
 #[inline]
-pub unsafe fn write_bytes<T>(dst: *mut T, val: u8, count: usize) {
+pub const unsafe fn write_bytes<T>(dst: *mut T, val: u8, count: usize) {
     extern "rust-intrinsic" {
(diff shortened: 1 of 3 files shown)


woppopo:
Rebased.

library/core/src/intrinsics.rs
```diff
@@ -2242,13 +2242,33 @@ pub const unsafe fn copy<T>(src: *const T, dst: *mut T, count: usize) {
...
     extern "rust-intrinsic" {
+        #[rustc_const_unstable(feature = "const_ptr_write", issue = "86302")]
         fn write_bytes<T>(dst: *mut T, val: u8, count: usize);
     }
 
-    debug_assert!(is_aligned_and_not_null(dst), "attempt to write to unaligned or null pointer");
+    #[cfg(not(bootstrap))]
+    const fn compiletime_check<T>(_ptr: *mut T) {
+        // The CTFE-engine checks if the pointer is correctly aligned and not null.
+    }
```
oli-obk, line 2256: similarly to the other PR, please just call this function `nop` and mention on the const_eval_select call that it's just a debug assertion and so it's not a problem if CTFE does not check it, even if it checks some of the rules

ecstatic-morse:
@woppopo `MaybeUninit::zeroed` was one of the functions that the Rust-for-Linux devs mentioned specifically in today's CTCFT as being useful in a const context. Any chance you'd be able to incorporate Oli's feedback in the near future? It would be nice to get this onto nightly.

woppopo:
Thank you for telling me. I added `#[cfg(debug_assertions)]` to the functions needed for checking, and unified with the same style as 90041.

camelid:
It seems this branch was pushed to after the PR was approved, so it needs a new review.

PR #90084, opened 2021-10-20
Make printed message match the code comment

noncombatant:
I think this code is getting L0, not L1 cache size, if I'm reading the Intel manual right. (I might not be.) Either way, the code comment and the printed message should match, whichever way is right. :)


steveklabnik (approved):

steveklabnik:
Thanks! @bors: r+ rollup

PR #90086, opened 2021-10-20
Fix more minor typos

noncombatant:
Also, the Unsupported Register table should mention that LLVM reserves `rbx` (as noted in a previous PR I filed tonight.)

### DIFF
--- a/src/doc/unstable-book/src/library-features/asm.md
+++ b/src/doc/unstable-book/src/library-features/asm.md
@@ -340,7 +340,7 @@ unsafe {
 assert_eq!(x, 0xabab);
 ```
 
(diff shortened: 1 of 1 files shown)


JohnTitor:
Could you move the changes in #90082 and #90084 to this PR? It doesn't make much sense to separate those and reviewing at once would be easier.

Amanieu:
@noncombatant Are you still working on this PR? I am currently preparing the `asm!` documentation for inclusion in the reference and in rust by example, based on the current contents of the unstable book page.

noncombatant:
I don't have time, unfortunately.

Amanieu:
I've incorporated the fixes from this PR in https://github.com/rust-lang/reference/pull/1105 and https://github.com/rust-lang/rust-by-example/pull/1483. Thanks @noncombatant

PR #90087, opened 2021-10-20
Sync rustfmt subtree (1.4.38)

calebcartwright:
There's a large number of small fixes and new features, but nothing too big. Detailed changelog for those interested can be found in https://github.com/rust-lang/rustfmt/blob/master/CHANGELOG.md#1438-2021-10-20


calebcartwright:
@Mark-Simulacrum I'm going to go ahead and approve as things look good and I don't think there'd be any scheduling concerns since the next beta has already been cut. Let me know if that's not the case though and we need to hold off on merging for now.

@bors r+

PR #90089, opened 2021-10-20
Improve display of enum variants

### LINKED ISSUE #90061: Fix display of enums with subfields
https://docs.rs/xmlparser/0.13.4/xmlparser/enum.Token.html

image

(I've expanded one of the "Show fields" toggles in the above screenshot)

Problems:

  - The font weight of "Fields of Comment" is _higher_ than the font weight of the "Comment" heading above it. "Comment" should be larger and heavier weight - similar to the weight of methods. And "Fields of _" should be lightweight.
  - "Fields of Comment" is an `<h3>`, but "Comment" is not a heading at all. This should be reversed.

EDIT(camelid): The toggles seen in the image above have already been removed. See https://github.com/rust-lang/rust/issues/90061#issuecomment-947146254 for an updated screenshot.

jsha:
Use h3 and h4 for the variant name and the "Fields" subheading.
Remove the "of T" part of the "Fields" subheading.
Remove border-bottom from "Fields" subheading.
Move docblock below "Fields" listing.

Fixes #90061

Demo:

https://jacob.hoffman-andrews.com/rust/xmlparser-updated/xmlparser/enum.Token.html#variants
https://jacob.hoffman-andrews.com/rust/fix-enum-variants/std/io/enum.ErrorKind.html#variants
https://jacob.hoffman-andrews.com/rust/fix-enum-variants/std/result/enum.Result.html#variants

r? @camelid

### DIFF
--- a/src/librustdoc/html/render/print_item.rs
+++ b/src/librustdoc/html/render/print_item.rs
@@ -1080,7 +1080,7 @@ fn item_enum(w: &mut Buffer, cx: &Context<'_>, it: &clean::Item, e: &clean::Enum
                 cx.derive_id(format!("{}.{}", ItemType::Variant, variant.name.as_ref().unwrap()));
             write!(
                 w,
-                "<div id=\"{id}\" class=\"variant small-section-header\">\
+                "<h3 id=\"{id}\" class=\"variant small-section-header\">\
                     <a href=\"#{id}\" class=\"anchor field\"></a>\
                     <code>{name}",
                 id = id,
@@ -1093,9 +1093,7 @@ fn item_enum(w: &mut Buffer, cx: &Context<'_>, it: &clean::Item, e: &clean::Enum
             }
             w.write_str("</code>");
             render_stability_since(w, variant, it, cx.tcx());
-            w.write_str("</div>");
-            document(w, cx, variant, Some(it), HeadingOffset::H3);
-            document_non_exhaustive(w, variant);
+            w.write_str("</h3>");
 
             use crate::clean::Variant;
             if let Some((extra, fields)) = match *variant.kind {
@@ -1109,12 +1107,8 @@ fn item_enum(w: &mut Buffer, cx: &Context<'_>, it: &clean::Item, e: &clean::Enum
                     variant.name.as_ref().unwrap()
                 ));
                 write!(w, "<div class=\"sub-variant\" id=\"{id}\">", id = variant_id);
-                write!(
-                    w,
-                    "<h3>{extra}Fields of <b>{name}</b></h3><div>",
-                    extra = extra,
-                    name = variant.name.as_ref().unwrap(),
-                );
+                write!(w, "<h4>{extra}Fields</h4>", extra = extra,);
+                document_non_exhaustive(w, variant);
                 for field in fields {
                     match *field.kind {
                         clean::StrippedItem(box clean::StructFieldItem(_)) => {}
@@ -1126,7 +1120,8 @@ fn item_enum(w: &mut Buffer, cx: &Context<'_>, it: &clean::Item, e: &clean::Enum
                             ));
                             write!(
                                 w,
-                                "<span id=\"{id}\" class=\"variant small-section-header\">\
+                                "<div class=\"sub-variant-field\">\
+                                 <span id=\"{id}\" class=\"variant small-section-header\">\
                                     <a href=\"#{id}\" class=\"anchor field\"></a>\
                                     <code>{f}:&nbsp;{t}</code>\
                                 </span>",
@@ -1134,13 +1129,16 @@ fn item_enum(w: &mut Buffer, cx: &Context<'_>, it: &clean::Item, e: &clean::Enum
                                 f = field.name.as_ref().unwrap(),
                                 t = ty.print(cx)
                             );
-                            document(w, cx, field, Some(variant), HeadingOffset::H4);
+                            document(w, cx, field, Some(variant), HeadingOffset::H5);
+                            write!(w, "</div>");
                         }
                         _ => unreachable!(),
                     }
                 }
-                w.write_str("</div></div>");
+                w.write_str("</div>");
             }
+
+            document(w, cx, variant, Some(it), HeadingOffset::H4);
         }
     }
     let def_id = it.def_id.expect_def_id();
--- a/src/librustdoc/html/static/css/rustdoc.css
+++ b/src/librustdoc/html/static/css/rustdoc.css
@@ -1101,25 +1101,28 @@ a.test-arrow:hover{
 	margin-right: 5px;
 }
 
-.sub-variant, .sub-variant > h3 {
-	margin-top: 0px !important;
-	padding-top: 1px;
-}
-
-#main .sub-variant > h3 {
-	font-size: 15px;
-	margin-left: 25px;
-	margin-bottom: 5px;
-}
-
-.sub-variant > div {
-	margin-left: 20px;
+h3.variant {
+	font-weight: 600;
+	font-size: 1.1em;
 	margin-bottom: 10px;
+	border-bottom: none;
 }
 
-.sub-variant > div > span {
-	display: block;
-	position: relative;
+.sub-variant h4 {
+	font-size: 1em;
+	font-weight: 400;
+	border-bottom: none;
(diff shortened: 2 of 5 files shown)


Urgau:
Just a quick message to say that I think that the layout is (with your changes) even more confusing. I don't know where I should put my eyes. The variant name is so big that it's disturbing, I feel that I should stop at every single of them. The all "Variant" layout feels out of place with the rest of the documentation.

Maybe instead of putting the variant name in h{3,4} you could make it bolder, this would would accentuate the difference but not be as disturbing as your changes.

Urgau:
Nice, way better.

One thing, it think you can completely remove the "Fields" title (we can clearly tell that they are fields). This would look like this:
Without fields title

This would simplify the layout to only let the important information.

camelid:
> One thing, it think you can completely remove the "Fields" title (we can clearly tell that they are fields).

I was going to suggest this, but one problem is that for tuple fields, the docs might be confusing. The fields aren't actually named `0`, `1`, etc., but there's no indication that they belong to a tuple variant.

GuillaumeGomez:
> > One thing, it think you can completely remove the "Fields" title (we can clearly tell that they are fields).
> 
> I was going to suggest this, but one problem is that for tuple fields, the docs might be confusing. The fields aren't actually named `0`, `1`, etc., but if you remove that heading, there's no indication that they belong to a tuple variant.

Currently they are, so I don't see this as an issue.

@jsha: I find it harder to read the documentation because it's actually not that easy to find where the field starts and where its documentation starts.

jsha:
I've updated the CSS like so:

 - More vertical separation between variant name and the "Fields" heading.
 - More vertical separation between the bottom of the variant's fields and the beginning of the variant's documentation.
 - Leftward indentation is bigger and more consistent with what we do elsewhere: 24px.
 
I think this should improve visual distinctiveness between the bottom of the fields and the beginning of the doccomment. I don't want to remove the "Fields" heading right now, since I think it is important for explaining the section it comes in front of.

Also I'd like to point out this is intended to fix some significant problems with the display of enum field variants (horizontal line inverts the hierarchy; fields show up below doccomment; fields heading is bold, inverting the hierarchy). I always want to try and get the details right, of course, but I also don't want to let a few pixels of whitespace take up too much time here - there are other big display issues I'd like to move on to fixing, also.

camelid:
> > I was going to suggest this, but one problem is that for tuple fields, the docs might be confusing. The fields aren't actually named `0`, `1`, etc., but if you remove that heading, there's no indication that they belong to a tuple variant.
> 
> Currently they are, so I don't see this as an issue.

What do you mean by "currently they are"? The issue I was describing is that in the docs, the fields are displayed as `0:` etc., which makes it look like the variant was defined as `Variant { 0: i32 }`, which is invalid syntax. That's why it seems potentially useful to keep the "Tuple Fields" header, to make it clear that it's actually a tuple variant.

jsha:
> Putting the fields before the variant's doc-comment seems like it could be confusing. The variant's docs are the summary, while the fields are the details; and in general, one would not put the summary after the details. Can you explain your rationale for this change?

In other places, the doccomment is always the last part of a section, and fields / other generated stuff is at the top. Consider the top of the page, for instance - it has struct contents, or enum contents, and then the doccomment underneath. However, I don't feel strongly about this, so I've reverted to the previous behavior of fields on the bottom.

I've also updated the font sizes as you suggest, and pushed the demos.

camelid (commented):
This is a good improvement, thanks for doing this! I'll review the code soon; for now I have one last UI suggestion.

camelid (commented):
I'm still learning how to review rustdoc's CSS, so I'm starting with the HTML :)

I'll review the CSS soon.

jsha (commented):
I'm pretty sure I sync'ed the `std` demo after my latest change, but I've just sync'ed again to be sure.

src/librustdoc/html/static/css/rustdoc.css
```diff
@@ -1094,25 +1094,33 @@ a.test-arrow:hover{
...
+	margin-top: 0;
+	margin-bottom: 0;
 }
 
-.sub-variant > div {
-	margin-left: 20px;
-	margin-bottom: 10px;
+.sub-variant {
+	margin-left: 24px;
+	margin-bottom: 20px;
```
camelid, line 1114: I feel like there's not quite enough visual separation between the fields' docs and the variant's docs:



Adding more space between them (e.g., `margin-bottom: 40px` here) could help, but I wonder if there's a better way to separate them. Do you have any ideas?

src/librustdoc/html/static/css/rustdoc.css
```diff
@@ -1094,25 +1094,33 @@ a.test-arrow:hover{
...
 }
 
-.sub-variant > div > span {
-	display: block;
-	position: relative;
+.sub-variant {
+	margin-top: 0px !important;
```
camelid, line 1118: This line doesn't seem to do anything (nothing happens when I toggle it in devtools). It seems it's pre-existing; do you want to remove it?

src/librustdoc/html/static/css/rustdoc.css
```diff
@@ -1094,25 +1094,33 @@ a.test-arrow:hover{
...
+	padding-top: 1px;
```
camelid, line 1119: This is pre-existing, but maybe we should change this to 0px or 5px? It seems like an odd size.

Or control it from the `h3.variant` margin-bottom instead. It's a bit confusing to have both.

camelid:
Also, can you rebase over master now that #90186 has merged? (You can remove the draft status then too.)

src/librustdoc/html/static/css/rustdoc.css
jsha: I went with the `margin-bottom: 40px`.

jsha:
Rebased and made requested changes. I've updated https://rustdoc.crud.net/jsha/enum-headings/foo/enum.Token.html but not the other demos.

camelid:
(I'll try to review this soon.)

src/librustdoc/html/render/print_item.rs
```diff
@@ -1126,21 +1120,25 @@ fn item_enum(w: &mut Buffer, cx: &Context<'_>, it: &clean::Item, e: &clean::Enum
...
+                            write!(w, "</div>");
                         }
                         _ => unreachable!(),
                     }
                 }
-                w.write_str("</div></div>");
+                w.write_str("</div>");
             }
+
+            document(w, cx, variant, Some(it), HeadingOffset::H4);
```
camelid, line 1141: Hmm, should this be H3?

camelid (commented):
The Rust changes mostly look fine to me, but I'd like @GuillaumeGomez to review the CSS/HTML tests.

src/librustdoc/html/render/print_item.rs
GuillaumeGomez: I *think* (iirc) that it means the biggest heading will be `h4` (please someone correct me if I'm wrong).

GuillaumeGomez (approved):

GuillaumeGomez:
Looks good to me as well. Once we have confirmation for @camelid's last message, we can approve it. :)

src/librustdoc/html/render/print_item.rs
camelid: But I think the heading immediately above is h3?

camelid:
Once https://github.com/rust-lang/rust/pull/90089#discussion_r751487785 is addressed, `r=camelid,GuillaumeGomez` :)

src/librustdoc/html/render/print_item.rs
jsha: Yes, this is correct. In the demo: https://rustdoc.crud.net/jsha/enum-headings/foo/enum.Token.html#variant.Declaration

"Declaration" is h3, and then "[Variant-First](https://rustdoc.crud.net/jsha/enum-headings/foo/enum.Token.html#variant-first)" is h4, as it should be.

camelid: Ok, I guess I'm still confused by `HeadingOffset`'s numbering.

camelid:
Thanks again for making this great UI improvement!

PR #90100, opened 2021-10-20
Skip documentation for tier 2 targets on dist-x86_64-apple-darwin

Mark-Simulacrum:
I don't have an easy way to test this locally, but I believe it should work. Based on one log result should shave ~14 minutes off the dist-x86_64-apple builder (doesn't help with aarch64 dist or x86_64 test builder, so not actually decreasing total CI time most likely).

r? @pietroalbini

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -287,7 +287,7 @@ jobs:
             os: ubuntu-latest-xl
           - name: dist-x86_64-apple
             env:
-              SCRIPT: "./x.py dist"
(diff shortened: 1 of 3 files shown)


pietroalbini:
Well the r=me was mostly about the code change itself :sweat_smile:

Mark-Simulacrum:
OK, I have a local setup for monitoring CI times now that should hopefully work well enough that we can detect changes caused by this PR (given the expected multi-minute impact).

I'm going to go ahead and approve (per earlier r=me) and will check in on the auto branch builds in a few days to see what the effect of this PR is. My guess is it'll help, but we'll see.

@bors r=pietroalbini

For the record, currently seeing (rolling 32 commit median time):

image

PR #90104, opened 2021-10-20
Implement coherence checks for negative trait impls

spastorino:
The main purpose of this PR is to be able to [move Error trait to core](https://github.com/rust-lang/project-error-handling/issues/3).

This feature is necessary to handle the following from impl on box.

```rust
impl From<&str> for Box<dyn Error> { ... }
```

Without having negative traits affect coherence moving the error trait into `core` and moving that `From` impl to `alloc` will cause the from impl to no longer compiler because of a potential future incompatibility. The compiler indicates that `&str` _could_ introduce an `Error` impl in the future, and thus prevents the `From` impl in `alloc` that would cause overlap with `From<E: Error> for Box<dyn Error>`. Adding `impl !Error for &str {}` with the negative trait coherence feature will disable this error by encoding a stability guarantee that `&str` will never implement `Error`, making the `From` impl compile.

We would have this in `alloc`:

```rust
impl From<&str> for Box<dyn Error> {} // A 
impl<E> From<E> for Box<dyn Error> where E: Error {} // B
```

and this in `core`:

```rust
trait Error {}
impl !Error for &str {}
```

r? @nikomatsakis 

This PR was built on top of @yaahc PR #85764.

Language team proposal: to https://github.com/rust-lang/lang-team/issues/96


nikomatsakis (changes requested):
Left a few minor things but I think this is basically ready to land.

nikomatsakis (changes requested):
Left two nits -- one to open and issue and one to tweak the code.

r=me after that.

PR #90126, opened 2021-10-21
Update Clippy

flip1995:
r? @Manishearth 

This adds a new dep to Clippy: [indoc](https://crates.io/crates/indoc)

And transitively: [unindent](https://crates.io/crates/unindent)


jplatte:
Probably not important enough. I've silenced the lint where it triggered wrongly.

PR #90128, opened 2021-10-21
Stabilize -Z symbol-mangling-version=v0 as -C symbol-mangling-version=v0

joshtriplett:
This allows selecting `v0` symbol-mangling without an unstable option. Selecting `legacy` still requires -Z unstable-options.

This does not change the default symbol-mangling-version. See https://github.com/rust-lang/rust/pull/89917 for a pull request changing the default. Rationale, from #89917:

Rust's current mangling scheme depends on compiler internals; loses information about generic parameters (and other things) which makes for a worse experience when using external tools that need to interact with Rust symbol names; is inconsistent; and can contain . characters which aren't universally supported. Therefore, Rust has defined its own symbol mangling scheme which is defined in terms of the Rust language, not the compiler implementation; encodes information about generic parameters in a reversible way; has a consistent definition; and generates symbols that only use the characters A-Z, a-z, 0-9, and _.

Support for the new Rust symbol mangling scheme has been added to upstream tools that will need to interact with Rust symbols (e.g. debuggers).

This pull request allows enabling the new v0 symbol-mangling-version.

See #89917 for references to the implementation of v0, and for references to the tool changes to decode Rust symbols.


Mark-Simulacrum:
I think unless the goal here is to only stabilize *just* the existence, not the specifics, even not changing the default still feels like it deserves a writeup similar to that given as part of the other PR seeking to change the default. There were several concerns raised on that PR that should get addressed in such a writeup, I believe. @michaelwoerister was going to work on that next week.

joshtriplett:
Exactly. I was proposing to leave legacy unstable for now. We could change that when changing the default, or we could leave it permanently unstable and eventually remove it.

nikomatsakis:
@rfcbot reviewed

wesleywiser (approved):

camelid:
@joshtriplett I think if you rebase this it's good to go :)

joshtriplett:
@camelid Rebased.

camelid:
Great! @wesleywiser could you re-review/approve this? Thanks!

@rustbot ready

wesleywiser (approved):

wesleywiser:
Thanks for the reminder @camelid!

@bors r+

ehuss:
@joshtriplett Would you be willing to add docs for this at https://github.com/rust-lang/rust/blob/master/src/doc/rustc/src/codegen-options/index.md?

PR #90131, opened 2021-10-21
Fix a format_args span to be expansion

camsteffen:
I found this while exploring solutions for rust-lang/rust-clippy#7843.

r? @m-ou-se

### DIFF
--- a/src/test/ui/try-block/try-block-maybe-bad-lifetime.stderr
+++ b/src/test/ui/try-block/try-block-maybe-bad-lifetime.stderr
@@ -21,6 +21,8 @@ LL |             ::std::mem::drop(x);
 LL |         };
 LL |         println!("{}", x);
    |                        ^ value borrowed here after move
+   |
+   = note: this error originates in the macro `$crate::format_args_nl` (in Nightly builds, run with -Z macro-backtrace for more info)
 
 error[E0506]: cannot assign to `i` because it is borrowed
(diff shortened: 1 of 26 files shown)


src/test/ui/try-block/try-block-maybe-bad-lifetime.stderr
```diff
@@ -21,6 +21,8 @@ LL |             ::std::mem::drop(x);
 LL |         };
 LL |         println!("{}", x);
    |                        ^ value borrowed here after move
+   |
+   = note: this error originates in the macro `$crate::format_args_nl` (in Nightly builds, run with -Z macro-backtrace for more info)
```
camelid, line 25: This new output seems non-ideal, but there's probably not a good way to hide it. I guess it could potentially be useful if/when `-Z macro-backtrace` is stabilized?

camsteffen: Not great but there are other instances like this.

camsteffen:
@m-ou-se can I bother you for a review? The wrong span context is causing surprising behavior in Clippy.

m-ou-se:
Oh, sorry for the delay.

I don't think I'm the right person to review this though.

Maybe someone from the compiler team?

r? rust-lang/compiler

cjgillot:
Since the `AddrOf` node is created by the macro, the macro context indeed seems like the right one.
@bors r+

Issue #86177, opened 2021-06-09
-Zinstrument-coverage ignores file (or module) 


taiki-e:
`-Z instrument-coverage` seems to ignore files or modules where all items are dead code.

*Originally found in https://github.com/taiki-e/cargo-llvm-cov/pull/26*.

## How to reproduce

```sh
mkdir repro
cd repro
mkdir src

echo 'mod module;
pub use module::*;
#[test]
fn f() {}' >src/lib.rs

echo 'pub fn func(x: u32) {
    match x {
        0 => {}
        1 => {}
        _ => {}
    }
}' >src/module.rs

echo '[package]
name = "repro"
version = "0.0.0"' >Cargo.toml

RUSTFLAGS="-Z instrument-coverage" \
    LLVM_PROFILE_FILE="repro-%m.profraw" \
    cargo test --tests

cargo profdata -- merge \
    -sparse repro-*.profraw -o repro.profdata

cargo cov -- report \
    $( \
      for file in \
        $( \
          RUSTFLAGS="-Z instrument-coverage" \
            cargo test --tests --no-run --message-format=json \
              | jq -r "select(.profile.test == true) | .filenames[]" \
              | grep -v dSYM - \
        ); \
      do \
        printf "%s %s " -object $file; \
      done \
    ) \
  --instr-profile=repro.profdata --summary-only
```

Result (`src/module.rs` is ignored):

```text
Filename                      Regions    Missed Regions     Cover   Functions  Missed Functions  Executed       Lines      Missed Lines     Cover    Branches   Missed Branches     Cover
-----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
src/lib.rs                          3                 0   100.00%           3                 0   100.00%           3                 0   100.00%           0                 0         -
-----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
TOTAL                               3                 0   100.00%           3                 0   100.00%           3                 0   100.00%           0                 0         -
```

### Workaround

The known workaround is using `-C link-dead-code`, but it seems not the correct workaround and is likely to cause another problem. See also https://github.com/taiki-e/cargo-llvm-cov/pull/26#issuecomment-858035691

### Meta

`rustc --version --verbose`:
```
rustc 1.54.0-nightly (ed597e7e1 2021-06-08)
binary: rustc
commit-hash: ed597e7e19d0fe716d9f81b1e840a5abbfd7c28d
commit-date: 2021-06-08
[...]

Issue #79121, opened 2020-11-17
Tracking Issue for LLVM InstrProf code coverage (current flag: `-Zinstrument-coverage`)


richkadel:
This is a tracking issue for the MCP "Implement LLVM-compatible source-based code coverage" (rust-lang/compiler-team#278).

Original feature request: Issue #34701.

### About tracking issues

Tracking issues are used to record the overall progress of implementation.
They are also used as hubs connecting to other relevant issues, e.g., bugs or open design questions.
A tracking issue is however *not* meant for large scale discussion, questions, or bug reports about a feature.
Instead, open a dedicated issue for the specific matter and add the relevant feature gate label.

### Steps

- [x] Implement the MCP
- [x] Adjust documentation ([see instructions on rustc-dev-guide][doc-guide])
  * The `Rustc Dev Guide` includes detailed documentation on the implementation in a new section, [`LLVM Source-Based Code Coverage `](https://rustc-dev-guide.rust-lang.org/llvm-coverage-instrumentation.html#llvm-and-covmap-versions)
  * Usage is documented in the `Rust Unstable Book` at https://doc.rust-lang.org/nightly/unstable-book/compiler-flags/source-based-code-coverage.html
  * Also see the `Inside Rust Blog` article at https://blog.rust-lang.org/inside-rust/2020/11/12/source-based-code-coverage.html
- [x] Stabilization PR ([see instructions on rustc-dev-guide][stabilization-guide])

[stabilization-guide]: https://rustc-dev-guide.rust-lang.org/stabilization_guide.html#stabilization-pr
[doc-guide]: https://rustc-dev-guide.rust-lang.org/stabilization_guide.html#documentation-prs

### Unresolved Questions

There are no unresolved questions from the MCP.

Some implementation options were debated and, if deferred, were logged as `FIXME` comments in the compiler source code or tests. `FIXME` comments related to stabilization have been added as `rust-lang/rust` issues.

* If/when stabilized, will the compiler flag be updated to `-C instrument-coverage`? (If so, the `-Z` variant could also be supported for some time, to ease migrations for existing users and scripts.)

* The Rust coverage implementation depends on (and automatically turns on) `-Z symbol-mangling-version=v0`. Will stabilizing this feature depend on stabilizing `v0` symbol-mangling first? If so, what is the current status and timeline?

[...]

Issue #85461, opened 2021-05-19
Unable to link regex crate with "-Z instrument-coverage" on x86_64-pc-windows-msvc


ririsoft:
Hello,

Linking [regex](https://docs.rs/regex/1.5.4/regex/) crate on `x86_64-pc-windows-msvc` fails when using `-Z instrument-coverage`.
The link succeeds when using `-Z instrument-coverage=except-unused-functions`

```toml
# Cargo.toml
[package]
name = "foobar"
version = "0.1.0"
authors = ["foobar"]
edition = "2018"

[dependencies]
regex = "1.5.4"
```

```rust
// main.rs
fn main() {
    use regex::Regex;
    let re = Regex::new(r"^\d{4}-\d{2}-\d{2}$").unwrap();
    assert!(re.is_match("2014-01-01"));
}
```

I expected to see this happen: linking successful.

Instead, this happened:

```sh
error: linking with `link.exe` failed: exit code: 1227

note: libmemchr-40eebd4a74e50761.rlib(memchr-40eebd4a74e50761.memchr.2b5ledse-cgu.14.rcgu.o) : fatal error LNK1227: conflicting weak extern definition for '_RINvNtNtNtCs838Isxeq56n_6memchr6memmem9prefilter11genericsimd14find_in_chunk3pEB8_'.  New default '.weak._RINvNtNtNtCs838Isxeq56n_6memchr6memmem9prefilter11genericsimd14find_in_chunk3pEB8_.default._RINvMNtCs93I3MPx9COt_4core5sliceSh13get_uncheckedINtNtNtB5_3ops5range7RangeTojEECs838Isxeq56n_6memchr' conflicts with old default '.weak._RINvNtNtNtCs838Isxeq56n_6memchr6memmem9prefilter11genericsimd14find_in_chunk3pEB8_.default._RINvMNtNtCs93I3MPx9COt_4core3ptr9const_ptrPh4castuECs838Isxeq56n_6memchr' in libmemchr-40eebd4a74e50761.rlib(memchr-40eebd4a74e50761.memchr.2b5ledse-cgu.3.rcgu.o)
```

### Meta

`rustc --version --verbose`:
```
rustc +nightly --version --verbose       
rustc 1.54.0-nightly (3e99439f4 2021-05-17)
binary: rustc
commit-hash: 3e99439f4dacc8ba0d2ca48d221694362d587927
commit-date: 2021-05-17
host: x86_64-pc-windows-msvc
[...]

PR #90132, opened 2021-10-21
Stabilize `-Z instrument-coverage` as `-C instrument-coverage`

joshtriplett:
(Tracking issue for `instrument-coverage`: https://github.com/rust-lang/rust/issues/79121)

This PR stabilizes support for instrumentation-based code coverage, previously provided via the `-Z instrument-coverage` option. (Continue supporting `-Z instrument-coverage` for compatibility for now, but show a deprecation warning for it.)

Many, many people have tested this support, and there are numerous reports of it working as expected.

Move the documentation from the unstable book to stable rustc documentation. Update uses and documentation to use the `-C` option.

Addressing questions raised in the tracking issue:

> If/when stabilized, will the compiler flag be updated to -C instrument-coverage? (If so, the -Z variant could also be supported for some time, to ease migrations for existing users and scripts.)

This stabilization PR updates the option to `-C` and keeps the `-Z` variant to ease migration.

> The Rust coverage implementation depends on (and automatically turns on) -Z symbol-mangling-version=v0. Will stabilizing this feature depend on stabilizing v0 symbol-mangling first? If so, what is the current status and timeline?

This stabilization PR depends on https://github.com/rust-lang/rust/pull/90128 , which stabilizes `-C symbol-mangling-version=v0` (but does not change the default symbol-mangling-version).

> The Rust coverage implementation implements the latest version of LLVM's Coverage Mapping Format (version 4), which forces a dependency on LLVM 11 or later. A compiler error is generated if attempting to compile with coverage, and using an older version of LLVM.

Given that LLVM 13 has now been released, requiring LLVM 11 for coverage support seems like a reasonable requirement. If people don't have at least LLVM 11, nothing else breaks; they just can't use coverage support. Given that coverage support currently requires a nightly compiler and LLVM 11 or newer, allowing it on a stable compiler built with LLVM 11 or newer seems like an improvement.

The [tracking issue](https://github.com/rust-lang/rust/issues/79121) and the [issue label A-code-coverage](https://github.com/rust-lang/rust/labels/A-code-coverage) link to a few open issues related to `instrument-coverage`, but none of them seem like showstoppers. All of them seem like improvements and refinements we can make after stabilization.

The original `-Z instrument-coverage` support went through a compiler-team MCP at https://github.com/rust-lang/compiler-team/issues/278 . Based on that, @pnkfelix suggested that this needed a stabilization PR and a compiler-team FCP.


davidtwco (approved):
Changes in this PR look good to me.

antoyo (approved):
The gcc part looks good to me.

richkadel:
You may also need to stablize the `#[no_coverage]` attribute as well. See #84605. It's a minimal solution at present, but I think sufficient for the time being. Stablizing will allow you to move the references to that attribute to the stable docs, resolving the broken link here.

Mark-Simulacrum:
The documentation I think currently points users towards installing/using https://crates.io/crates/cargo-cov, which looks both somewhat old and is not a rust-lang crate. I think we generally try to stay away from third party tools being directly referenced by our documentation, and especially it worries me when it says "may eat your laundry". We also point towards https://crates.io/crates/cargo-binutils and json5format, the latter of which seems entirely unnecessary.

I also note that the recommended path is making use of the llvm-tools-preview component, which I am somewhat uncomfortable with: as it notes, it's intentionally a preview component and while we have no firm policy, we don't treat it as really stable. Obviously, users can find llvm-cov in distros and such too, but that feels a little bit outside the beaten path. This component is not *currently* nightly-only, but only because it's introduction predates the ability to declare components as nightly only, I think.

In general, I think usage of LLVM tools (e.g., profdata, cov) should be OK -- we may want to consider uplifting them to a separate component and we should check with upstream that they're expected to stick around. (For example, the split-dwarf effort discovered that the LLVM tool it wanted to use is not well maintained, IIRC). But I would like us to drop references to third-party crates or consider whether they should be blessed from the documentation before we stabilize.

I think regardless, we need a description in the documentation for this flag as to what is considered stable (and if LLVM upstream or whatever changes, we need to workaround and patch to keep same behavior) and what isn't. This will be useful for future PR review making changes to this feature, such as https://github.com/rust-lang/rust/pull/90047.

richkadel:
@Mark-Simulacrum - These are all good concerns/recommendations.

Since I wrote most of the initial docs, its reasonable someone might want me to make at least some of the required changes, but...

I'll be out of office for about 3 weeks, without a laptop, starting tonight. (I'll probably check email via my phone, periodically, in case there's a question I can answer.)

Please keep @tmandry in the loop. He may be able to help with some parts of this. And if I can help when I return, I will.

Thanks!

cuviper:
> requiring LLVM 11 for coverage support seems like a reasonable requirement.

We're also bumping the overall minimum to LLVM 11 in #90062.

BlackHoleFox:
This feature [doesn't currently work](https://github.com/rust-lang/rust/issues/85461) with MSVC targets without extra `RUSTFLAGS`. I think stabilization should be postponed, making sure that it works correctly on all the tier 1 platforms. first.

davidhewitt:
> Rust to offer it's own CLI for interacting with coverage that didn't require me to drop down to the LLVM tools

Depending on exactly what you're doing, I can thoroughly recommend the excellent [`cargo-llvm-cov`](https://github.com/taiki-e/cargo-llvm-cov) subcommand by @taiki-e.

I've been using it on PyO3 to merge coverage data from multiple packages and feature sets into a single report for codecov with much joy.

joshtriplett:
Rebased.

joshtriplett:
I can add documentation regarding the current issues with instrumenting on Windows (suggesting either `instrument-coverage=except-unused-functions` or `-Ccodegen-units=1`). Would that help?

BlackHoleFox:
I think that'd be good to avoid confusion. I haven't tried `instrument-coverage=except-unused-functions` myself, though.

joshtriplett:
Done.

nagisa:
One thought that I just had is that maybe we should have `-Cinstrument` cover both the needs of coverage and various other instrumentations that we do (`-Zsanitize`, `-Ccontrol-flow-guard` and many others) as they all are conceptually just the compiler instrumenting the code in some way or another.

apiraino:
Discussed during T-compiler meeting [on Zulip](https://rust-lang.zulipchat.com/#narrow/stream/238009-t-compiler.2Fmeetings/topic/.5Bweekly.5D.202021-12-02.20.2354818/near/263458442), outcome is @wesleywiser to draft an FCP

@rustbot label -I-compiler-nominated

wesleywiser:
@rfcbot fcp merge

joshtriplett:
Rebased to apply to current HEAD.

jhpratt:
I think we should take a good look at #86177 before stabilizing this. It can be _massively_ wrong if an entire module is uncovered, giving users a false sense of security.

wesleywiser:
I think I know what's going on with that issue and I'll take a stab at fixing it.

wesleywiser:
I posted #92142 to fix #86177. 

@joshtriplett I would like to recommend that we only stabilize `-Cinstrument-coverage` and `-Cinstrument-coverage=off` and do not stabilize the `except-unused-generics` and `except-unused-functions` arguments. These options were added to help users work around issues linking crates and generating coverage data. We've resolved all the known issues in this class and it's not clear that these options are useful for any other reason.

QiangHeisenberg:
I'm sorry, I'm confused. Are the results of the test coverage really correct? Or is my way wrong?https://github.com/rust-lang/rust/issues/91661

joshtriplett:
@wesleywiser Sounds reasonable; I can work on that.

joshtriplett:
@wesleywiser Done; `-C instrument-coverage=except-*` options now require `-Zunstable-options`. (Assuming CI passes.)

nagisa:
> Given that LLVM 13 has now been released, requiring LLVM 11 for coverage support seems like a reasonable requirement. If people don't have at least LLVM 11, nothing else breaks

Minimum required LLVM version is 12 since October 24th.

nagisa:
I have read through the discussion here and it isn't entirely clear to me what exactly the FCP covers.

For example, the text in documentation describes the output and how to interpret it using the `llvm-*` tools, effectively stabilizing the output format too. My impression was that we are looking to somehow not stabilize the output format, but there is no mention to that effect anywhere in the docs…

richkadel:
Note that #91207 added support for the latest coverage mapping formats, either version 5 or 6... 6 if using LLVM 13.

joshtriplett:
@nagisa The intention is that the exact output format is not stabilized, just the fact that it can be interpreted using an appropriate version of the LLVM tools. The format could change completely, as long as the new format were interpretable by an appropriate version of the LLVM tools.

Quoting text this PR adds to `src/doc/rustc/src/codegen-options/index.md`:
> Note that while the `-C instrument-coverage` option is stable, the profile data
> format produced by the resulting instrumentation may change, and may not work
> with coverage tools other than those built and shipped with the compiler.

nikomatsakis:
@rfcbot reviewed

eddyb:
> @nagisa The intention is that the exact output format is not stabilized, just the fact that it can be interpreted using an appropriate version of the LLVM tools. The format could change completely, as long as the new format were interpretable by an appropriate version of the LLVM tools.

I suppose that's not very different from something like `--emit=llvm-bc`, or the ability to do cross-language LTO, in which the backend controls the format, and matching versions (of LLVM in this case) have to be used.
(That is, I think there's precedent here, and going forward with stabilizing this seems fine)

wesleywiser (approved):

richkadel:
Thank you @joshtriplett, @wesleywiser, and the Rust compiler team!

Exciting to see this feature graduate to stable!

cc: @tmandry

matthiaskrgr:
@bors r- 
I'm getting a test failure locally (file might have bene added since last PR rebase), probably the test just needs to be updated.

````
iiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiF..
failures:

---- [ui] ui/issues/issue-85461.rs stdout ----
normalized stderr:
warning: `-Z instrument-coverage` is deprecated; use `-C instrument-coverage`




The actual stderr differed from the expected stderr.
Actual stderr saved to /home/matthias/vcs/github/rust/build/x86_64-unknown-linux-gnu/test/ui/issues/issue-85461/issue-85461.stderr
To update references, rerun the tests and pass the `--bless` flag
To only update this specific test, also pass `--test-args issues/issue-85461.rs`

error: 1 errors occurred comparing output.
status: exit status: 0
command: "/home/matthias/vcs/github/rust/build/x86_64-unknown-linux-gnu/stage1/bin/rustc" "/home/matthias/vcs/github/rust/src/test/ui/issues/issue-85461.rs" "-Zthreads=1" "--target=x86_64-unknown-linux-gnu" "--error-format" "json" "--json" "future-incompat" "-Ccodegen-units=1" "-Zui-testing" "-Zdeduplicate-diagnostics=no" "-C" "prefer-dynamic" "--out-dir" "/home/matthias/vcs/github/rust/build/x86_64-unknown-linux-gnu/test/ui/issues/issue-85461" "-A" "unused" "-Crpath" "-Cdebuginfo=0" "-Lnative=/home/matthias/vcs/github/rust/build/x86_64-unknown-linux-gnu/native/rust-test-helpers" "-Zinstrument-coverage" "-Ccodegen-units=4" "--crate-type" "dylib" "-Copt-level=0" "-L" "/home/matthias/vcs/github/rust/build/x86_64-unknown-linux-gnu/test/ui/issues/issue-85461/auxiliary"
stdout:

stderr:
warning: `-Z instrument-coverage` is deprecated; use `-C instrument-coverage`


failures:
    [ui] ui/issues/issue-85461.rs

test result: FAILED. 4 passed; 1 failed; 12583 ignored; 0 measured; 0 filtered out; finished in 0.47s

Some tests failed in compiletest suite=ui mode=ui host=x86_64-unknown-linux-gnu target=x86_64-unknown-linux-gnu
Build completed unsuccessfully in 0:00:16
````

richkadel:
@joshtriplett - Looks like it's in a test @wesleywiser added, merge Dec 10 I think. I see you converted older tests using `-Z instrument-coverage` to use `-C` so doing the same for this test (unless others slipped in as well) should resolve it.

wesleywiser:
@joshtriplett I took the liberty of pushing a commit to your branch that changes the three occurrences of `-Zinstrument-coverage` in tests which have been added since this PR was opened to `-Cinstrument-coverage`. 

@bors r+

wesleywiser:
Looks like `i686-msvc-1` timed out while running stage 2 tests. 

@bors retry

PR #90144, opened 2021-10-21
Use a hardcoded constant instead of calling OpenProcessToken.

smmalis37:
GetCurrentProcessToken is defined in processthreadsapi.h as:
```c++
FORCEINLINE
HANDLE
GetCurrentProcessToken (
    VOID
    )
{
    return (HANDLE)(LONG_PTR) -4;
}
```

Since it's very unlikely that this constant will ever change, let's just use it instead of making api calls to get the same information.

This also (I think) removes the need for rust binaries to directly link against advapi32. Let me know if some changes are needed in those build-related bits, that I'm much less sure on.

### DIFF
--- a/library/std/src/sys/windows/c.rs
+++ b/library/std/src/sys/windows/c.rs
@@ -676,17 +676,8 @@ if #[cfg(not(target_vendor = "uwp"))] {
 
     pub const HANDLE_FLAG_INHERIT: DWORD = 0x00000001;
 
-    pub const TOKEN_READ: DWORD = 0x20008;
-
-    #[link(name = "advapi32")]
-    extern "system" {
-        // Allowed but unused by UWP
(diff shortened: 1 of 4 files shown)


joshtriplett:
cc @ChrisDenton 

Is this a reasonable assumption to make?

smmalis37:
Unfortunately I don't know the answer to that. Will CI test Windows 7? Or does someone have a machine/vm to test on?

smmalis37:
Dang, oh well.

ChrisDenton:
It's undocumented but if it's only used as a fallback then it could be worth it.

On the other hand, `GetUserProfileDirectoryW` is only used by `env::home_dir` which is now deprecated. So the import should not appear in most Rust programs and can be avoided by using a third party implementation if it's needed.

PR #90146, opened 2021-10-21
Reduce use of LocalDefId <-> HirId maps

cjgillot:
This is an attempt to reduce the perf effect of https://github.com/rust-lang/rust/pull/89278.
r? @ghost

### DIFF
--- a/compiler/rustc_middle/src/hir/map/mod.rs
+++ b/compiler/rustc_middle/src/hir/map/mod.rs
@@ -361,8 +361,7 @@ impl<'hir> Map<'hir> {
         id.as_local().and_then(|id| self.find(self.local_def_id_to_hir_id(id)))
     }
 
-    pub fn get_generics(&self, id: DefId) -> Option<&'hir Generics<'hir>> {
-        let id = id.as_local()?;
+    pub fn get_generics(&self, id: LocalDefId) -> Option<&'hir Generics<'hir>> {
         let node = self.tcx.hir_owner(id)?;
         match node.node {
             OwnerNode::ImplItem(impl_item) => Some(&impl_item.generics),
--- a/compiler/rustc_middle/src/hir/map/mod.rs
+++ b/compiler/rustc_middle/src/hir/map/mod.rs
@@ -117,13 +117,13 @@ pub struct ParentOwnerIterator<'hir> {
 }
 
 impl<'hir> Iterator for ParentOwnerIterator<'hir> {
-    type Item = (HirId, OwnerNode<'hir>);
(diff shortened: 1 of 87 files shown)


cjgillot:
r? rust-lang/compiler

davidtwco (approved):
This looks reasonable to me, but probably warrants another opinion.

davidtwco:
r? rust-lang/compiler

nagisa:
Will not be able to take a look at this before 27th. Please reassign if this needs a quicker review.

compiler/rustc_middle/src/hir/map/mod.rs
```diff
@@ -340,11 +340,23 @@ impl<'hir> Map<'hir> {
...
+    }
+
     /// Retrieves the `Node` corresponding to `id`, panicking if it cannot be found.
     pub fn get(&self, id: HirId) -> Node<'hir> {
         self.find(id).unwrap_or_else(|| bug!("couldn't find hir id {} in the HIR map", id))
     }
 
+    /// Retrieves the `Node` corresponding to `id`, panicking if it cannot be found.
+    #[inline]
+    pub fn get_def(&self, id: LocalDefId) -> Node<'hir> {
```
nagisa, line 356: This function name suggests to me “get a def(id)” while what it really means is “get keyed by defid”. Would it be too onerous to make the name more accurate?

compiler/rustc_resolve/src/late/lifetimes.rs
```diff
@@ -2030,32 +2027,31 @@ impl<'a, 'tcx> LifetimeContext<'a, 'tcx> {
...
-                    debug!("hir id first={:?}", hir_id);
-                    if let Some((id, span, name)) = match self.tcx.hir().get(hir_id) {
-                        Node::Lifetime(hir_lifetime) => Some((
-                            hir_lifetime.hir_id,
-                            hir_lifetime.span,
-                            hir_lifetime.name.ident(),
-                        )),
-                        Node::GenericParam(param) => {
-                            Some((param.hir_id, param.span, param.name.ident()))
+                    debug!("hir id first={:?}", def_id);
```
nagisa, line 2030: this is no longer printing the hir_id, is it? That said, `debug!(?def_id)` should be more future proof ^^

nagisa (approved):
Yeah, the idea seems sound. Couple nits inline, up to you if you think its worth to address them or not.

nagisa:
I'm happy to r=me this. Please rebase and squash the history a little before merging though.

PR #90174, opened 2021-10-22
Add error::Report type 

seanchen1991:
This implements `std::error::Report`, a wrapper around an error that exposes the entire error chain, along with options to format the output of the report as either a single line or as multiple lines. 

Tracking issue: #90172.

### DIFF
--- a/library/std/src/error.rs
+++ b/library/std/src/error.rs
@@ -811,6 +811,8 @@ impl dyn Error + Send + Sync {
 /// An error reporter that exposes the entire error chain for printing.
 /// It also exposes options for formatting the error chain, either entirely on a single line,
 /// or in multi-line format with each cause in the error chain on a new line.
+/// `Report` only requires that the wrapped error implements `Error`. It doesn't require that the
+/// wrapped error be `Send`, `Sync`, or `'static`.
 ///
 /// # Examples
 ///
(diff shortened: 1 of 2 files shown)


m-ou-se (changes requested):

library/std/src/error.rs
```diff
@@ -807,3 +807,246 @@ impl dyn Error + Send + Sync {
...
+{
+    /// Create a new `Report` from an input error.
+    #[unstable(feature = "error_reporter", issue = "90172")]
+    pub fn new(source: E) -> Report<E> {
+        Report { source, show_backtrace: false, pretty: false }
+    }
+
+    /// Enable pretty-printing the report.
+    #[unstable(feature = "error_reporter", issue = "90172")]
+    pub fn pretty(mut self) -> Self {
```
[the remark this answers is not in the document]
yaahc: Yea, I think its particularly important that `fallible_fn().map_err(Report::new).unwrap()` be usable, so forcing a `.build()` call at the end feels a little awkward to me.

library/std/src/error/tests.rs
```diff
@@ -35,3 +35,294 @@ fn downcasting() {
...
+    let expected = String::from("SuperErrorSideKick is here!");
+
+    assert_eq!(expected, actual);
+}
+
+#[test]
+fn error_with_backtrace_outputs_correctly() {
+    use backtrace::Backtrace;
+
+    env::remove_var("RUST_BACKTRACE");
```
m-ou-se, line 114: `remove_var` should normally not be used in multi-threaded contexts such as in tests. This will affect the entire test process, and might cause trouble depending on what other tests are doing in parallel.

Is there a better way to test what you need to test, without having to change this variable in the process?

library/std/src/error/tests.rs
```diff
@@ -35,3 +35,294 @@ fn downcasting() {
...
+    let error = GenericError::new_with_source(MyMessage, error);
+    let error = GenericError::new_with_source(MyMessage, error);
+    let report = Report::new(error).pretty(true);
+    let expected = r#"
+The message
+
+
+Caused by:
+   0: The message
+   1: The message"#;
```
m-ou-se, line 301: Why are the newlines removed here?

m-ou-se (changes requested):
Looks like there's a few more bugs:

Empty lines are dropped in sources:

```
line 1

line 3

Caused by:
   0: line 1
      line 3
   1: line 1
      line 3
```

And with a single source, multi-line indentation breaks:

```
line 1
line 2

Caused by:
    line 1
line 2
```

yaahc:
Almost done resolving the recent round of comments but I have a couple of style questions that came up during the implementation that I could use some input on.

First, I want to make sure we have the right level of indentation setup for our multiline errors so they match up nicely with the output of `Backtrace`, and I'm wondering what you think would be best @m-ou-se.

Here's the current version with the changes I made and one source error:

```
Error: The source of the error

Caused by:
    Error with backtrace

Stack backtrace:
   0: std::error::tests::error_with_backtrace_outputs_correctly
   1: test::__rust_begin_short_backtrace
   2: <alloc::boxed::Box<F,A> as core::ops::function::FnOnce<Args>>::call_once
   3: std::panicking::try
   4: test::run_test_in_process
   5: std::sys_common::backtrace::__rust_begin_short_backtrace
   6: std::panicking::try
   7: core::ops::function::FnOnce::call_once{{vtable.shim}}
   8: <alloc::boxed::Box<F,A> as core::ops::function::FnOnce<Args>>::call_once
   9: std::sys::unix::thread::Thread::new::thread_start
  10: start_thread
             at /build/glibc-eX1tMB/glibc-2.31/nptl/pthread_create.c:477:8
  11: clone
```

And with 2 or more sources

```
Error: Error with two sources

Caused by:
    0: The source of the error
    1: Error with backtrace

Stack backtrace:
   0: std::error::tests::error_with_backtrace_outputs_correctly_with_two_sources
   1: test::__rust_begin_short_backtrace
   2: <alloc::boxed::Box<F,A> as core::ops::function::FnOnce<Args>>::call_once
   3: std::panicking::try
   4: test::run_test_in_process
   5: std::sys_common::backtrace::__rust_begin_short_backtrace
   6: std::panicking::try
   7: core::ops::function::FnOnce::call_once{{vtable.shim}}
   8: <alloc::boxed::Box<F,A> as core::ops::function::FnOnce<Args>>::call_once
   9: std::sys::unix::thread::Thread::new::thread_start
  10: start_thread
             at /build/glibc-eX1tMB/glibc-2.31/nptl/pthread_create.c:477:8
```

This is identical to the formatting rules for `anyhow`, as far as I can tell. Right now it seems to want to line up the error messages so that each message starts on the same column, but my instinct is to line it up so that the causes start on the same column as the symbols in the backtrace. The case where the errors aren't numbered already doesn't line up with the outermost source error message, so either way I'd like it to be changed for consistency, whether or not that ends up being based on the first error message or the backtrace.

yaahc:
Merged as https://github.com/rust-lang/rust/pull/91938, closing this PR now

PR #90181, opened 2021-10-22
fix(rustc_typeck): report function argument errors on matching type

notriddle:
Fixes #90101


notriddle:
CC @nagisa

Not sure how to assign to both of you?

PR #90196, opened 2021-10-23
Fix and extent ControlFlow `traverse_inorder` example

### LINKED ISSUE #90063: `traverse_inorder` example in `ControlFlow` compiles, but can't be called
Found by Inky-developer in https://users.rust-lang.org/t/how-to-use-documentation-example-for-std-controlflow/66115?u=scottmcm

The example compiles, but can't actually be monomorphized as it adds another `&mut` at each level of the recursion.

It needs a fix like https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=0b534dc88883f0ebd67774f56a857da2

(And it'd probably be good to show a call in the example anyway.)

yanok:
Fix and extent ControlFlow `traverse_inorder` example

1. The existing example compiles on its own, but any usage fails to be monomorphised and so doesn't compile. Fix that by using Fn trait instead of FnMut.
2. Added an example usage of `traverse_inorder` showing how we can terminate the traversal early.

Fixes #90063

### DIFF
--- a/library/core/src/ops/control_flow.rs
+++ b/library/core/src/ops/control_flow.rs
@@ -24,7 +24,7 @@ use crate::{convert, ops};
 /// ```
 ///
 /// A basic tree traversal:
-/// ```no_run
+/// ```
 /// use std::ops::ControlFlow;
 ///
 /// pub struct TreeNode<T> {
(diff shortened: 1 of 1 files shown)


scottmcm (changes requested):
Thanks for making the PR, @yanok!  I've left some other comments with requested changes.

Once you've made changes and would like me to look at it again, please do `@rustbot label: +S-waiting-on-review -S-waiting-on-author`

scottmcm (approved):

PR #90197, opened 2021-10-23
Suggest adding a `#[cfg(test)]` to a test module

### LINKED ISSUE #88138: Unused import warning in a test could suggest #[cfg(test)]
Given the following code: [Playground link](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=8ab3d6b8fab93ac0c34cf822a91664cb)

```rust
pub struct Example();

mod test {
    use super::Example;
    #[test]
    fn test_f() {
        Example();
    }
}
```

The current output is:

```
warning: unused import: `super::Example`
 --> src/lib.rs:4:9
  |
4 |     use super::Example;
  |         ^^^^^^^^^^^^^^
  |
```

However, the import _is_ actually required for the test.  It would be helpful if the compiler suggested adding `#[cfg(test)]` to `mod test` to fix the issue.  (Thanks to pie_flavor on Discord for pointing out that that's the issue :))

TaKO8Ki:
closes #88138

### DIFF
--- a/compiler/rustc_resolve/src/check_unused.rs
+++ b/compiler/rustc_resolve/src/check_unused.rs
@@ -320,7 +320,16 @@ impl Resolver<'_> {
                 visitor.r.local_def_id(unused.use_tree_id).to_def_id(),
             );
             let test_module_span = match module_to_string(parent_module) {
-                Some(module) if module.contains("test") => Some(parent_module.span),
+                Some(module)
+                    if module == "test"
+                        || module == "tests"
+                        || module.starts_with("test_")
+                        || module.starts_with("tests_")
+                        || module.ends_with("_test")
+                        || module.ends_with("_tests") =>
+                {
+                    Some(parent_module.span)
+                }
                 _ => None,
             };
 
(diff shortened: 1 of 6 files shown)


TaKO8Ki:
Could you review this PR? @wesleywiser @estebank

wesleywiser (approved):

wesleywiser:
Looks good to me. Thanks @TaKO8Ki!

@bors r+

TaKO8Ki:
@wesleywiser Thank you!

compiler/rustc_resolve/src/check_unused.rs
```diff
@@ -315,12 +316,20 @@ impl Resolver<'_> {
                 "remove the unused import"
             };
 
+            let parent_module = visitor.r.get_nearest_non_block_module(
+                visitor.r.local_def_id(unused.use_tree_id).to_def_id(),
+            );
+            let test_module_span = match module_to_string(parent_module) {
+                Some(module) if module.contains("test") => Some(parent_module.span),
```
rhysd, line 323: @TaKO8Ki @wesleywiser

If my understanding is correct, This would detect modules like the following as 'test' module.

```rust
mod fastest_search {
  // ...
}
```

Is it intentional? I'm afraid that this loose check causes many false positives. I'd recommend to check more strictly like `module == "test" || module == "tests"`.

TaKO8Ki: @wesleywiser I fixed this false positive. Could you review again?

rhysd: As of c597dc8, it looks good to me :+1:

TaKO8Ki:
Could you review this PR again? @wesleywiser @estebank

PR #90202, opened 2021-10-23
Improve and test cross-crate hygiene

matthewjasper:
- Decode the parent expansion for traits and enums in `rustc_resolve`, this was already being used for resolution in typeck
- Avoid suggesting importing names with def-site hygiene, since it's often not useful
- Add more tests

r? @petrochenkov


petrochenkov:
Thanks!
@bors r+

PR #90204, opened 2021-10-23
Make lowering pull-based

cjgillot:
~Based on https://github.com/rust-lang/rust/pull/90451~
Part of https://github.com/rust-lang/rust/pull/88186

The current lowering code visits all the item-likes in the AST in order, and lowers them one by one.
This PR changes it to index the AST and then proceed to lowering on-demand. This is closer to the logic of query-based lowering.

### DIFF
--- a/compiler/rustc_hir/src/definitions.rs
+++ b/compiler/rustc_hir/src/definitions.rs
@@ -99,6 +99,7 @@ impl DefPathTable {
 #[derive(Clone, Debug)]
 pub struct Definitions {
     table: DefPathTable,
+    next_disambiguator: FxHashMap<(LocalDefId, DefPathData), u32>,
 
     /// Item with a given `LocalDefId` was defined during macro expansion with ID `ExpnId`.
     expansions_that_defined: FxHashMap<LocalDefId, ExpnId>,
@@ -340,6 +341,7 @@ impl Definitions {
 
         Definitions {
             table,
+            next_disambiguator: Default::default(),
             expansions_that_defined: Default::default(),
             def_id_to_span,
             stable_crate_id,
@@ -357,7 +359,6 @@ impl Definitions {
         parent: LocalDefId,
         data: DefPathData,
         expn_id: ExpnId,
-        mut next_disambiguator: impl FnMut(LocalDefId, DefPathData) -> u32,
         span: Span,
     ) -> LocalDefId {
         debug!("create_def(parent={:?}, data={:?}, expn_id={:?})", parent, data, expn_id);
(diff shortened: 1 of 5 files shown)


michaelwoerister:
Marking this as blocked on https://github.com/rust-lang/rust/pull/90451 for now.

michaelwoerister:
@cjgillot, is this ready for reviewing or is it still blocked on something (other PRs or performance)?

cjgillot:
@michaelwoerister this is reviewable as is.
I did not investigate much the performance results since last measurement. I expect some kind of regression, since we are doing strictly more work by indexing the AST first.

michaelwoerister:
OK, I'll schedule some time for taking a look at this next week :+1:

compiler/rustc_hir/src/definitions.rs
```diff
@@ -28,6 +28,7 @@ pub struct DefPathTable {
     index_to_key: IndexVec<DefIndex, DefKey>,
     def_path_hashes: IndexVec<DefIndex, DefPathHash>,
     def_path_hash_to_index: DefPathHashMap,
+    next_disambiguator: FxHashMap<(LocalDefId, DefPathData), u32>,
```
michaelwoerister, line 31: I would prefer this to live in `Definitions`, since that is where it's used.

michaelwoerister (approved):
Thanks for the PR, @cjgillot! 

This looks good to me. I only left one little nit.

I imagine that the performance regression comes from creating multiple lowering contexts instead of a single one. It seems to be in an acceptable range for the real world benchmarks and will hopefully be made up for when lowering becomes incremental.

cjgillot:
@spastorino do you mind if we merge the `node_id_to_hir_id` refactor here? You'd continue the perf investigations on top of it.

@michaelwoerister the perf effect probably comes from the `node_id_to_hir_id` refactor and the indexing the AST (which is strictly more work). However, I'm not sure that creating distinct `LoweringContext`s has an effect.

spastorino:
@cjgillot seems good to me.

spastorino:
@cjgillot now that #93438 is in, you can get rid of the first commit, rebase and then maybe re-run perf?.

cjgillot:
PR https://github.com/rust-lang/rust/pull/93438 has been merged.
If you don't have outstanding concerns, I think we mostly need to decide if the performance hit is acceptable. The perf run is mostly red <1%, with the exception of unused-warnings in the 1-2% regression.
I'll rebase when I get the occasion.

michaelwoerister:
Do you think that some of the performance will be recovered once the whole refactoring is done?
Overall, the [performance hit](https://perf.rust-lang.org/compare.html?start=03c8ffaacb040a8753ef8e1accea701bc9f5be85&end=647bffbdd0ae2c27f52c85431f5c26fce2b46de3) of this PR doesn't look very big. IMO, it's not a blocking issue.

cjgillot:
> Do you think that some of the performance will be recovered once the whole refactoring is done?

I don't know, though I hope so. For now, https://github.com/rust-lang/rust/pull/88186 is a perf regression (up to 4% on primary benchmarks, up to 11% with stress benchmarks). If we want to skip parts of lowering altogether, we will need in addition:
- to hash the AST and part of ResolverOutputs -> the perf tradeoff is not obvious;
- avoid many accesses to HIR (for instance, #95004 may help to get us closer).

michaelwoerister:
OK, as stated before, I think the performance hit is acceptable as it should not be noticeable in any of the real-world benchmarks. You can r=me after rebasing. Thanks for all the hard work you are putting into this!

rylev:
Marking this as triaged since a justification was already given [here](https://github.com/rust-lang/rust/pull/90204#pullrequestreview-878629128).

PR #90205, opened 2021-10-23
Repace use of `static_nobundle` with `native_link_modifiers` within Rust codebase

mati865:
This fixes warnings when building Rust and running tests:
```
warning: library kind `static-nobundle` has been superseded by specifying `-bundle` on library kind `static`. Try `static:-bundle`
warning: `rustc_llvm` (lib) generated 2 warnings (1 duplicate)
```


petrochenkov:
>Let me know if I should drop tests changes

Nah, `static-nobundle` will (hopefully) soon removed anyway, so the tests will need to be updated too.
@bors r+

PR #90233, opened 2021-10-24
Tooltip overflow

### LINKED ISSUE #89185: rustdoc: Scrollbar shown instead of warning (or error) icon
It looks like it's inheriting `overflow-x: auto` for some reason.

### nightly (broken)

image

### beta (good)

image

GuillaumeGomez:
Fixes #89185.

r? @camelid

### DIFF
--- a/src/librustdoc/html/static/css/rustdoc.css
+++ b/src/librustdoc/html/static/css/rustdoc.css
@@ -532,7 +532,7 @@ nav.sub {
 	position: relative;
 }
 
-.docblock > * {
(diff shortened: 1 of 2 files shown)


camelid (commented):
r=me after fixing the selector. It might not be necessary, but it seems that removing the `*` is preferred by MDN and Firefox devtools. When I tried to enter `.docblock > *:not(.information)` into the devtools before, it changed it to `.docblock > :not(.information)`.

Issue #59853, opened 2019-04-10
Hide unstable items by default in stable rustdoc


lambda:
Right now, the docs produced on the stable and beta channels include unstable items. The docs are already fairly complex and contain a lot of information to sort through, and finding documentation for items that are not usable for the majority of developers working on stable Rust does not help with this. In addition, unstable APIs have very attention-grabbing warnings associated with them, making them much more prominent than the APIs that are actually usable.

<img width="1031" alt="Screen shot demonstrating eye-catching unstable warning" src="https://user-images.githubusercontent.com/37398/55966332-3220fe80-5c46-11e9-9efe-6c181e7351e6.png">

This has been broken out of #59829 to provide smaller, actionable items that can be independently discussed and worked on. A review of rustdoc for accessibility, especially for dyslexia and attention disorders, finds that there are a number of visually distracting elements in rustdoc which can draw attention away from the main content. This should be considered along with the other issues mentioned in #59829 about the overall effect of a number of distracting elements on the page.

This has been previously discussed and closed in #25863 and #31156. However, when it was first discussed was very shortly after the Rust 1.0 release, when there was a large amount of API surface which had not been stabilized for the 1.0 release, and what could be done with stable Rust was fairly limited (for example, no stable proc macros so Serde was only usable on nightly Rust).

At this point, most people can use stable Rust for most tasks. There is a lot less reason for the average Rust developer to need to switch to nightly to use unstable standard library features.

There is also the open issue #31204 that the stable docs of unstable items may be out of date; because these items are unstable, they can change in nightly releases, so not only are they a distraction, but they may be misleading for people who actually try to use them.

Between the extra information that most people don't need, distracting UI, potential for being out of date, and time since this was last discussed, I think that it is worth reconsidering this decision and hiding unstable items in the stable docs by default.

This would require either providing an opt-in setting for seeing them, or make finding the nightly docs easier.

An alternative discussed in the comments is to just make them less visually prominent; unstable items are already rendered at lower contrast and at the bottom on module index pages, but within struct or trait documentation they have big bright warning boxes calling attention to them. Making them less visually prominent, such as by reducing their contrast and making the warning just plain unstyled text, perhaps in the same place that stability information goes (though see #59851 for discussion on that), would help out a lot.

Centril:
> This would require either providing an opt-in setting for seeing them

I would prefer this option. One benefit of showing the unstable APIs is that it invites people to test them, so having an opt-in for the curious would be good.

[...]

Issue #59829, opened 2019-04-09
Rustdoc formatting presents accessibility issues


lambda:
In a [thread on Reddit](https://www.reddit.com/r/rust/comments/bazy8u/how_do_experienced_rust_developers_read_the/ekg4ipg/) about finding Rust's documentation hard to read, one user pointed out that it's especially bad for people with dyslexia.

Not having dyslexia myself, I did a little bit of research on UX design for dyslexia. One resource I found pretty good was [this article](https://uxplanet.org/designing-for-dyslexia-6d12e8c41cd7), and in particular the presentations at the end.

One of the points about design for dyslexia, which can even help for those who are not dyslexic, is reducing the amount of visual noise and clutter. A quick pass shows that there are a lot of noisy elements. Most of them exist for a reason, but I'm not sure all of those reasons fully justify the amount of noise. I have highlighted a lot of the things that catch my attention which are not content.

![Screenshot of std::net::Ipv4Addr docs with visually noisy elements pointed out](https://i.imgur.com/TDTcrEv.png)

There are a lot of different colors used on the page. Blue is used as the standard link color; but links to other items in declarations also use a custom color scheme for different types of items, like primitives vs. structs vs. enums vs. type aliases. I don't know about other experienced Rustaceans, but as an experienced Rustacean myself, I certainly couldn't tell you what color corresponded to what kind of type it is; but rather, as an experienced Rustacean, for most of those types I either know what it is, due to familiarity with the type or familiarity with conventions, or I don't really need to know what it is and can click through to find out if I want to.

Then the example code is syntax highlighted using a different scheme than the declarations. Again, syntax highlighting has some value for readability, though I find some aspects of syntax highlighting to be more helpful for *writing* code than reading it. I would be interested in hearing about usability studies on readability of code based on various syntax highlighting schemes; my guess would bee that some aspects of highlighting, like longer spans of text which may be commented or quoted, would be more helpful for reading than others that are already locally obvious enough in the bare syntax. Anyhow, syntax highlighting may or may not be unnecessary noise, but in conjunction with the other highlighting scheme for declarations, there's definitely some room for improvement and simplification.

[...]

PR #90242, opened 2021-10-24
Add an item-info for non-public visibility, and remove redundant `pub`

jsha:
For most documentation, the `pub` in front of each method is redundant: you know the method is public because you are reading the documentation.

However, when running with `--document-private`, it's important to see which items are public and which are private. This adds an item-info tag for private items showing their visibility, and removes the `pub` from methods and fields.

This doesn't currently touch the summary code block at the top of the page, which will still have `pub` or `pub(crate)` or whatever is appropriate.

This refactors some code in Visibility to allow better reuse.

Part of #59829. Note: [one of the pieces of feedback](https://github.com/rust-lang/rust/issues/59853) there was that the item-infos (experimental, deprecated, platform-specific, etc) are too prominent. That may be the case, but for this change, I think the important thing is matching the other item-infos, and we can change the style of item-infos later if we want to.

r? @GuillaumeGomez
/cc @camelid 

Demo:

https://jacob.hoffman-andrews.com/rust/document-private/foo/outer_mod/inner_mod/inmost_mod/struct.Boo.html
https://jacob.hoffman-andrews.com/rust/document-private/foo/outmost_mod/outer_mod/inner_mod/inmost_mod/struct.Foo.html
https://jacob.hoffman-andrews.com/rust/document-private/foo/outmost_mod/outer_mod/inner_mod/inmost_mod/enum.Bar.html

image

### DIFF
--- a/src/librustdoc/html/format.rs
+++ b/src/librustdoc/html/format.rs
@@ -1221,50 +1221,80 @@ impl clean::FnDecl {
     }
 }
 
+pub(crate) enum VisibilityScope {
+    Inherited,
+    Crate,
+    Super,
+    Path(Vec<String>),
+}
+
+impl VisibilityScope {
+    crate fn new<'tcx>(vis_did: DefId, item_did: DefId, tcx: TyCtxt<'tcx>) -> VisibilityScope {
+        // FIXME(camelid): This may not work correctly if `item_did` is a module.
+        //                 However, rustdoc currently never displays a module's
+        //                 visibility, so it shouldn't matter.
+        let parent_module = find_nearest_parent_module(tcx, item_did);
+
+        if vis_did.index == CRATE_DEF_INDEX {
+            VisibilityScope::Crate
+        } else if parent_module == Some(vis_did) {
+            // `pub(in foo)` where `foo` is the parent module
+            // is the same as no visibility modifier
+            VisibilityScope::Inherited
+        } else if parent_module.map(|parent| find_nearest_parent_module(tcx, parent)).flatten()
+            == Some(vis_did)
+        {
+            VisibilityScope::Super
+        } else {
+            let path = tcx.def_path(vis_did);
+            debug!("path={:?}", path);
+            let mut components: Vec<String> = vec![];
+            for seg in &path.data {
+                components.push(seg.data.get_opt_name().unwrap().to_string());
+            }
+            VisibilityScope::Path(components)
+        }
+    }
+}
+
+impl fmt::Display for VisibilityScope {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        match self {
+            VisibilityScope::Inherited => f.write_str(""),
+            VisibilityScope::Crate => f.write_str("crate"),
+            VisibilityScope::Super => f.write_str("super"),
+            VisibilityScope::Path(components) => {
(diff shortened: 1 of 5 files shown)


camelid:
> This doesn't make the visibility display dependent on the --document-private-items flag. It removes pub from struct methods in all cases. The changes to --document-private-items are a followup from your comment a few months ago: if we omit pub from certain places, we need a way to distinguish things that are private.

Sorry, I wasn't clear. I was referring to Mark's suggestion from above:

> Perhaps we can instead only emit visibility annotations if --document-private-items is passed, but not modify their style? These feel like orthogonal changes to me.

-----

> `@Mark-Simulacrum` said:
> 
> > I'd personally rather avoid having rustdoc invent a new way of describing privacy -- and the lack of a visibility label does not imply public in most Rust code, so it seems surprising to take that approach here.
> 
> This is not really a new way to describe privacy: It's how privacy is described at https://doc.rust-lang.org/reference/visibility-and-privacy.html:
> 
> > By default, everything in Rust is _private_
> 
> The difference is that privacy is expressed one way in the code, and one way in the reference documentation. I think it makes sense to align rustdoc with the reference documentation rather than the code. The code expression is geared towards one use case: privacy should be the default. The documentation expression should be geared towards another use case: public visibility is the default.

What do you mean by the way privacy is expressed in the Reference documentation? The Rust Reference should be describing how Rust source code works. Where does the Reference say that public visibility is the default? I want to make sure we're not miscommunicating :)

-----

> Would this idea improve things? If a struct is private, we would mark that struct private at the top of its page, but we would not separately annotate its fields and methods as private, since that is implied by the struct being private.

It's hard for me to know without experiencing it first-hand. I wonder if we could consider changing the UI behind a feature flag so we could get practical experience with it before landing it for all users?

> Also, I'd be really interested to hear more details about your workflow - what crates do you use `--document-private` on, and what tasks are you looking to accomplish when you do? Are they mostly crates that will eventually be published? Or are they crates where everything is perpetually private?

Mostly, I'm referring to the rustc and rustdoc API docs (/nightly/nightly-rustc), which I use frequently while working on Rust. I also sometimes use `--document-private-items` with a private side project (that's an executable, not a library), but not very frequently. I also sometimes document rust-lang/triagebot with private items when working on it.

For the most part, I use `--document-private-items` to be able to easily browse large and/or unfamiliar codebases and as a quick way to search for items I see in the code that I want to learn more about.

-----

Overall, part of the issue is that I and others, like Mark, are used to the docs the way they are. Thus, it's hard for me to tell whether changing the format would be an improvement or regression, and either way, I'd have to get used to the new UI. Like I mentioned above, maybe we could figure out a way to land this behind a feature flag (or, rather, an unstable CLI flag)? Otherwise, it feels like we have to commit to the new UI before even trying it out "in the wild".

JohnCSimon:
ping from triage:
@jsha  
thanks.

Dylan-DPC:
@jsha any updates on this?

camelid:
I still prefer the current approach of trying to match source code syntax. And, if we were to switch to a less-close-to-syntax approach, I quite dislike using "item-infos" since they're very visually noisy.

jsha:
Yeah, I agree item-info is not the way to go here. I hadn't really grokked how much people use rustdoc to navigate private aspects of a crate. Closing for now; I'll try again with a different approach later.

PR #90253, opened 2021-10-25
Change several HashMaps to IndexMap to improve incremental hashing performance

Kobzol:
Stable hashing hash maps in incremental mode takes a lot of time, especially for some benchmarks like `clap`. As noted by @Mark-Simulacrum [here](https://github.com/rust-lang/rust/pull/89404#issuecomment-950043892), this cost could be reduced by replacing some hash maps by indexmaps.

I gathered some statistics and found several hash maps that took a lot of time to hash and replaced them by indexmaps. However, in order for this to work, we need to make sure that these indexmaps have deterministic insertion order. These three are used only in visitors as far as I can see, which seems deterministic. Can we enforce this somehow? Or should some explaining comment be included for these maps?

### DIFF
--- a/compiler/rustc_middle/src/middle/region.rs
+++ b/compiler/rustc_middle/src/middle/region.rs
@@ -7,7 +7,7 @@
 //! [rustc dev guide]: https://rustc-dev-guide.rust-lang.org/borrow_check.html
 
 use crate::ty::TyCtxt;
-use rustc_data_structures::fx::FxHashMap;
+use rustc_data_structures::fx::{FxHashMap, FxIndexMap};
 use rustc_data_structures::stable_hasher::{HashStable, StableHasher};
 use rustc_hir as hir;
 use rustc_hir::Node;
@@ -215,14 +215,14 @@ pub struct ScopeTree {
     /// conditional expression or repeating block. (Note that the
     /// enclosing scope ID for the block associated with a closure is
     /// the closure itself.)
-    pub parent_map: FxHashMap<Scope, (Scope, ScopeDepth)>,
+    pub parent_map: FxIndexMap<Scope, (Scope, ScopeDepth)>,
 
     /// Maps from a variable or binding ID to the block in which that
     /// variable is declared.
-    var_map: FxHashMap<hir::ItemLocalId, Scope>,
+    var_map: FxIndexMap<hir::ItemLocalId, Scope>,
 
     /// Maps from a `NodeId` to the associated destruction scope (if any).
-    destruction_scopes: FxHashMap<hir::ItemLocalId, Scope>,
+    destruction_scopes: FxIndexMap<hir::ItemLocalId, Scope>,
 
     /// `rvalue_scopes` includes entries for those expressions whose
     /// cleanup scope is larger than the default. The map goes from the


davidtwco (approved):
LGTM, don't think we need to be too concerned about deterministic insertion order for these maps, so unless you want to add a comment, r=me.

JohnCSimon:
Returning to author to address comments.

Kobzol:
I'm not sure if there's anything to address here, it's more like we are waiting for changes in accessing DefId, etc. @cjgillot ?

Kobzol:
@cjgillot The sort was removed from stable hashing hash maps in https://github.com/rust-lang/rust/pull/91837. I wonder if we could use the same trick with multiplying the individual hashes for IndexMap, which would make sure that the hash stays the same  even if the insertion order changes? Then we could switch these hash maps to index maps.

cjgillot:
IIUC, #91837 removed the sort in `impl HashStable for HashMap`. Should we re-run perf on this PR to see if we still have the perf improvements?

Kobzol:
Yup, that would be nice! Thanks.

cjgillot:
Could you resolve conflicts? I'll start perf afterwards.

Kobzol:
Oh, I didn't notice it, sorry. Rebased.

Kobzol:
@cjgillot Btw, since we're no longer sorting by the stable key, do we even need `ToStableHashKey` anymore? Before I tried to use `HashStable` directly for the keys and it didn't work because some of their `HashStable` impls panicked (I think it was for HIR bodies). But can't we just take the `ToStableHashKey` impl, use it for `HashStable` impl, and remove `ToStableHashKey`?

Kobzol:
It seems that it still helps `clap` to use an `IndexMap`. The main difference is that when hashing an `IndexMap`, the keys don't have to go through `ToStableHashKey` and there are no intermediate `StableHasher`s being created.

michaelwoerister:
I think the changes here are questionable from a correctness point of view.

Kobzol:
I agree that it's a bit worrying. However, what about current usages of `FxIndexMap`? Isn't basically any usage of it in stable hashing dangerous?

michaelwoerister:
> But can't we just take the ToStableHashKey impl, use it for HashStable impl, and remove ToStableHashKey?

Yes, `ToStableHashKey` was introduced just for sorting values in a stable way. We should be able to remove it now.

apiraino:
By reading a bit the recent comment thread I _think_ this is not `s-blocked` anymore, can anyone confirm (cc @cjgillot )? And I think it can be switched to waiting for the author to address the latest comments.

But in case please feel free to post an update of the current status. Thanks!

@rustbot  author

Kobzol:
To be honest, I don't know how much has the situation changed. I saw that there is some lint that guards against iteration of structures without deterministic insertion order and there have been a few changes to incr. comp. in the meantime, but other than that, I'm not sure. Let's see what @cjgillot says :)

michaelwoerister:
I think we have better knowledge now about what is safe and what isn't around fingerprinting of hash maps:

- IndexMap and IndexSet is always safe because fingerprinting takes the order into account. Fingerprint verification can catch errors if they occur.
- One still has to take care to construct such collections in a deterministic, stable order. Otherwise there'll be random cache invalidations (which poses a performance problem).
- This kind of determinism is only necessary for things that implement `HashStable`, i.e. things that can be query results or query keys.
- An alternative to using order-preserving collections is to use collections that don't expose the internal ordering of elements. @hameerabbasi is looking into that in https://github.com/rust-lang/rust/pull/94188.

For ScopeTree in particular it looks to me like we could probably use collections that don't expose the order at all. As it stands now though, switching to IndexMap actually seems safer to me than sticking with the regular HashMap.

Kobzol:
Ok, so this PR could be merged as-is? We should probably do another perf. run to evaluate the effect, there have been some changes to incremental since the latest perf. run.

michaelwoerister:
> Ok, so this PR could be merged as-is?

Yes, I'd say so. But I want to give @cjgillot a chance to give an opinion.

@bors try @rust-timer queue

Kobzol:
Perf. looks ok (in the benchmark suite, this really only helps `clap`).

Kobzol:
Hmm, not really sure what has happened.

rylev:
Given that the performance seen here largely mirrors what was seen when running perf previously (large improvements to clap-rs and otherwise an overall performance wash), I'll mark this as triaged. Let me know if you disagree with this assessment.

@rustbot triage label: perf-regression-triaged

Issue #86941, opened 2021-07-07
Stabilise std::is_aarch64_feature_detected


adamgemmell:
This issue constitutes the stabilisation report as per the [rustc dev guide](https://rustc-dev-guide.rust-lang.org/stabilization_guide.html#write-a-stabilization-report)

### Summary

I'm hoping to stabilise [std::is_aarch64_feature_detected](https://doc.rust-lang.org/std/macro.is_aarch64_feature_detected.html) which allows for runtime feature detection on aarch64 platforms. It was recently updated to include all features supported by both linux and LLVM 12, though the detection for some features isn't on non-linux platforms yet.

### Documentation

The documentation for it is in this [file](https://github.com/rust-lang/stdarch/blob/master/crates/std_detect/src/detect/arch/aarch64.rs). There's doc comments for individual features in the macro but I don't see these exposed anywhere, just on a hidden enum. Perhaps this could be improved. I've also got a small patch to mention it in the Rust Reference [here](https://github.com/rust-lang/reference/pull/1061).

### Tests

There are tests for it in [stdarch](https://github.com/rust-lang/stdarch/blob/master/crates/std_detect/tests/cpu-detection.rs) as well as [rust](https://github.com/rust-lang/rust/blob/master/library/std/tests/run-time-detect.rs). These are currently all up to date.

### Unresolved questions

The only backwards-compatible question left to resolve is whether to split `pauth` into `paca` (address authentication) and `pacg` (generic authentication) as linux does. The current behaviour (together) matches LLVM (and the ARMv8-A feature FEAT_PAuth).

The x86 macro was stabilised under the `simd_x86` feature - does this mean I should create a new `simd_aarch64` feature to stabilise this and other aarch64 stdsimd work under?

See also:
https://github.com/rust-lang/rust/issues/48556 for the tracking issue for the stdsimd feature.
https://github.com/rust-lang/rust/issues/90620 for stabilising the `target_feature` attribute on aarch64 platforms.

adamgemmell:
After some internal discussion my opinion on splitting `PAuth` has changed a little - apparently there have been attempts to enable partial userspace support of the feature in the past and it's reasonably possible that it could happen in the future, in which case we would want separate `paca` and `pacg` options. These could of course be added in the future, though we'll end up with those in addition to `pauth`, slightly confusingly.

Would it make sense to split them preemptively for runtime feature detection? Or would the above scenario be ok?

Amanieu:
[...]

PR #90271, opened 2021-10-25
Stabilise `is_aarch64_feature_detected!` under `simd_aarch64` feature 

adamgemmell:
Initial implementation, looking for feedback on the approach here. https://github.com/rust-lang/rust/issues/86941

One point I noticed was that I haven't seen different "since" versions for the same feature - does this mean that other features can't be added to to the `simd_aarch64` feature once this is in stable? If so it might need a more specific name.

r? @Amanieu

### DIFF
--- a/library/std/tests/run-time-detect.rs
+++ b/library/std/tests/run-time-detect.rs
@@ -3,10 +3,9 @@
 #![cfg_attr(
     any(
         all(target_arch = "arm", any(target_os = "linux", target_os = "android")),
-        all(target_arch = "aarch64", any(target_os = "linux", target_os = "android")),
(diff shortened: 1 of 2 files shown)


library/std/tests/run-time-detect.rs
```diff
@@ -6,7 +6,6 @@
         all(target_arch = "aarch64", any(target_os = "linux", target_os = "android")),
         all(target_arch = "powerpc", target_os = "linux"),
         all(target_arch = "powerpc64", target_os = "linux"),
-        any(target_arch = "x86", target_arch = "x86_64"),
```
Amanieu, line 9: Should this be removed for aarch64 as well? You might need to put this behind `cfg(bootstrap)` so the stage0 compiler still accepts it.

PR #90273, opened 2021-10-25
Clean up special function const checks

nbdd0121:
Mark them as const and `#[rustc_do_not_const_check]` instead of hard-coding them in const-eval checks.

r? @oli-obk 
@rustbot label A-const-eval T-compiler


fee1-dead (approved):
LGTM! Thanks!

PR #90274, opened 2021-10-25
Cleanup: Eliminate ConstnessAnd

oli-obk:
This is almost a behaviour-free change and purely a refactoring. "almost" because we appear to be using the wrong ParamEnv somewhere already, and this is now exposed by failing a test using the unstable `~const` feature. I am looking into that, but not sure if I'll get to it this week.

We most definitely need to review all `without_const` and at some point should probably get rid of many of them by using TraitPredicate instead of TraitRef.

This pull request is best reviewed commit by commit, as it works on each commit (modulo some ui changes that will go away once I squash 764aac2 into 8e84aa9

r? @fee1-dead 

cc @spastorino @ecstatic-morse


fee1-dead (commented):
Some queries do not care about if we are in a const context. Maybe we could always change the ParamEnv to be NonConst before caching?

I also think that we need a test for reproducing the error caused by the bad caching system.

fee1-dead:
@oli-obk: do you mind if I take this PR? You seem quite busy with other stuff and I really want this to land. I tried rebasing the PR a while back and I couldn't push to your repository.

oli-obk:
Yea, feel free to take over. I recommend splitting it into parts though, as the non-param-env parts should be perf and behaviour neutral

spastorino:
@fee1-dead ohh great. I've promised @oli-obk that I was going to take this over but had some stuff on my plate and couldn't yet. If you're going ahead please do, otherwise I'd take it over probably this week.

oli-obk:
Thanks @fee1-dead for carrying this over the finish line

PR #90277, opened 2021-10-25
Improve terminology around "after typeck"

### LINKED ISSUE #70258: Terminology: prefer "outside inference" to "after typeck" (or worse "during codegen").
Example offenders:
https://github.com/rust-lang/rust/blob/a03921701cdfe0b2c5422240f3ae370ab21069f1/src/librustc/ty/normalize_erasing_regions.rs#L18-L20

https://github.com/rust-lang/rust/blob/c6b172f7885cb83842fa15e4b0e2ff1647aebb40/src/librustc_trait_selection/traits/codegen/mod.rs#L18-L19

None of this is particularly true anymore (not that it was great when it used to be true), especially since we started using `miri` for constant evaluation, which uses all the same tools codegen does, and is interleaved/mutually recursive with type-checking.

The only real thing to keep in mind is that if you have inference variables, you're responsible for using `InferCtxt` correctly.

cc @rust-lang/compiler

pierwill:
Closes #70258.

### DIFF
--- a/compiler/rustc_infer/src/infer/mod.rs
+++ b/compiler/rustc_infer/src/infer/mod.rs
@@ -200,7 +200,7 @@ pub struct InferCtxtInner<'tcx> {
     // Opaque types found in explicit return types and their
     // associated fresh inference variable. Writeback resolves these
     // variables to get the concrete type, which can be used to
-    // 'de-opaque' OpaqueTypeDecl, after typeck is done with all functions.
+    // 'de-opaque' OpaqueTypeDecl outside of type inference.
     pub opaque_types: OpaqueTypeMap<'tcx>,
 
     /// A map from inference variables created from opaque
--- a/compiler/rustc_lint/src/context.rs
+++ b/compiler/rustc_lint/src/context.rs
@@ -523,7 +523,7 @@ impl LintStore {
     }
 }
 
-/// Context for lint checking after type checking.
+/// Context for lint checking outside of type inference.
 pub struct LateContext<'tcx> {
     /// Type context we're checking in.
     pub tcx: TyCtxt<'tcx>,
--- a/compiler/rustc_middle/src/mir/mod.rs
+++ b/compiler/rustc_middle/src/mir/mod.rs
@@ -864,7 +864,7 @@ pub struct LocalDecl<'tcx> {
     /// across a suspension point against the type components of the generator
     /// which type checking knows are live across a suspension point. We need to
     /// flag drop flags to avoid triggering this check as they are introduced
-    /// after typeck.
+    /// outside of type inference.
     ///
     /// This should be sound because the drop flags are fully algebraic, and
     /// therefore don't affect the auto-trait or outlives properties of the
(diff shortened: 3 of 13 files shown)


pierwill:
I can't really evaluate uses of the phrase "during codegen".

Here are the uses in `compiler/`:

<details>
<summary>search results</summary>

```
$ rg 'during code'
rustc_codegen_gcc/src/coverageinfo.rs
59:64:/// `function_coverage_map` (keyed by function `Instance`) during codegen.

rustc_codegen_llvm/src/context.rs
65:49:/// when we ptrcast, and we have to ptrcast during codegen

rustc_codegen_llvm/src/coverageinfo/mod.rs
85:64:/// `function_coverage_map` (keyed by function `Instance`) during codegen.

rustc_codegen_ssa/src/back/write.rs
333:48:// Handler to use for diagnostics produced during codegen.
1824:29:bug!("panic during codegen/LLVM phase");

rustc_codegen_ssa/src/base.rs
715:51:// Since the main thread is sometimes blocked during codegen, we keep track

rustc_codegen_ssa/src/coverageinfo/map.rs
102:62:/// from `u32::MAX`. Theses operands are translated only during code generation, after all

rustc_codegen_ssa/src/traits/mod.rs
8:81://! actual codegen, while the builder stores the information about the function during codegen and

rustc_middle/src/middle/codegen_fn_attrs.rs
90:13:/// during codegen.

rustc_middle/src/mir/coverage.rs
48:9:/// during codegen. LLVM expects zero-based indexes.

rustc_middle/src/mir/mod.rs
1210:9:/// during codegen where distinct kinds of basic blocks may be

rustc_middle/src/traits/mod.rs
666:43:/// is `Obligation`, as one might expect. During codegen, however, this

rustc_middle/src/ty/instance.rs
347:46:/// in a monomorphic context (i.e., like during codegen), then it is guaranteed to return

rustc_middle/src/ty/mod.rs
1215:54:/// Typically, this is `Reveal::UserFacing`, but during codegen we
1314:30:/// the desired behavior during codegen and certain other special

rustc_middle/src/ty/sty.rs
1409:60:/// Erased region, used by trait selection, in MIR and during codegen.

rustc_middle/src/ty/util.rs
191:9:/// during codegen.
267:9:/// during codegen.

rustc_mir_transform/src/coverage/mod.rs
47:85:/// counters, via intrinsic `llvm.instrprof.increment`, and/or inject metadata used during codegen

rustc_mir_transform/src/coverage/query.rs
74:70:// allocate arrays when generating the coverage map (during codegen), so choose

rustc_mir_transform/src/lower_intrinsics.rs
93:28:// during codegen. Issue #35310.

rustc_mir_transform/src/remove_storage_markers.rs
1:64://! This pass removes storage markers if they won't be emitted during codegen.

rustc_trait_selection/src/opaque_types.rs
229:20:// during codegen.

rustc_trait_selection/src/traits/codegen.rs
57:65:"encountered ambiguity selecting `{:?}` during codegen, presuming due to \
70:77:"Encountered error `Unimplemented` selecting `{:?}` during codegen",
77:65:bug!("Encountered error `{:?}` selecting `{:?}` during codegen", e, trait_ref)
```

</details>

So this might not fully close the issue.

r? @eddyb

JohnCSimon:
Ping from triage: assigning to author

pierwill:
Hi @JohnCSimon! 👋  Just wondering, how come you added `S-waiting-on-author`? I think this is ready for review.

PR #90291, opened 2021-10-26
Loosen the bound on the Debug implementation of Weak.

geeklint:
Both `rc::Weak<T>` and `sync::Weak<T>` currently require `T: Debug` in their own `Debug` implementations, but they don't currently use it;  they only ever print a fixed string.

A general implementation of Debug for Weak that actually attempts to upgrade and rely on the contents is unlikely in the future because it may have unbounded recursion in the presence of reference cycles, which Weak is commonly used in.  (This was the justification for why the current implementation [was implemented the way it is](https://github.com/rust-lang/rust/pull/19388/commits/f0976e2cf3f6b0027f118b791e0888b29fbb41a7)).

When I brought it up [on the forum](https://internals.rust-lang.org/t/could-the-bound-on-weak-debug-be-relaxed/15504), it was suggested that, even if an implementation is specialized in the future that relies on the data stored within the Weak, it would likely rely on specialization anyway, and could therefore easily specialize on the Debug bound as well.

### DIFF
--- a/library/alloc/src/rc.rs
+++ b/library/alloc/src/rc.rs
@@ -2393,7 +2393,7 @@ impl<T: ?Sized> Clone for Weak<T> {
 }
 
 #[stable(feature = "rc_weak", since = "1.4.0")]
-impl<T: ?Sized + fmt::Debug> fmt::Debug for Weak<T> {
+impl<T: ?Sized> fmt::Debug for Weak<T> {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
         write!(f, "(Weak)")
     }
--- a/library/alloc/src/sync.rs
+++ b/library/alloc/src/sync.rs
@@ -303,7 +303,7 @@ impl<T: ?Sized + Unsize<U>, U: ?Sized> CoerceUnsized<Weak<U>> for Weak<T> {}
 impl<T: ?Sized + Unsize<U>, U: ?Sized> DispatchFromDyn<Weak<U>> for Weak<T> {}
 
 #[stable(feature = "arc_weak", since = "1.4.0")]
-impl<T: ?Sized + fmt::Debug> fmt::Debug for Weak<T> {
+impl<T: ?Sized> fmt::Debug for Weak<T> {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
         write!(f, "(Weak)")
     }


joshtriplett:
This seems reasonable to me for exactly the stated reason: the Debug impl for Weak should never try to upgrade to get access to the thing it contains in order to print it.

@rfcbot merge

BurntSushi:
The reasoning here all makes sense to me. But what I'm missing or not understanding is a reason to do this in the first place. What concrete benefit do we gain by relaxing this impl?

geeklint:
I specifically ran into this while attempting to add a `#[derive(Debug)]` to a custom type that contained a Weak.   I had the existing understanding that Weak always formatted as just "(Weak)" so I was mildly surprised when the derive didn't work.

I questioned why the bound was there in the first place and most people I spoke to seemed to share the understanding that it was a remnant of the original implementation that was no longer required, and that therefore this PR would be worthwhile.

BurntSushi:
@geeklint interesting. Would it be possible to show a short code example? I think the derive would only not work if the type wrapped by your Weak didn't impl Debug either? That seems a little odd to me, but I could be misunderstanding. I think code will help make things more concrete for me.

> I questioned why the bound was there in the first place and most people I spoke to seemed to share the understanding that it was a remnant of the original implementation that was no longer required, and that therefore this PR would be worthwhile.

Yeah that makes sense. I just want to make sure we have all our bases covered here, since this isn't a change we can walk back. (Folks have mentioned specialization, but specialization is in a pretty weird state.) If this extra restriction isn't actually getting in the way of anything, then I'm skeptical of removing it.

geeklint:
@BurntSushi Yes you're correct, the type I was wrapping did not implement Debug - it probably could have but I chose not to.  This is a rough paraphrase of the code I had:

```rust
struct Shared {
    ptr: AtomicPtr<_>,
    index: AtomicUsize,
}

// the recv side has the owned copy, and would have a custom Debug implementation
pub struct RecvSide {
    shared: Arc<Shared>,
}

// the send side I was hoping could be simplified by using #[derive]
#[derive(Debug)]
pub struct SendSide {
    shared: Weak<Shared>,
}
```

I intentionally didn't want to `#[derive(Debug)]` on the Shared item because the contents are implementation detail in my library, and I'm not personally a fan of the Debug implementations for Atomics (that opinion is not related to this PR).  I suppose I could have manually implemented Debug for Shared so that it could have been derived for the public-facing types - I guess didn't think of that at the time - but I still think if there's no conceivable plan to rely on the `T: Debug` in `Weak<T>`, it doesn't need to retain the bound.

BurntSushi:
Thanks for sharing! It makes me a little uneasy to do this without something more compelling to be honest. If you're already doing a custom debug impl for `Arc<Shared>`, it doesn't seem that bad to do one more for `Weak<Shared>`. I grant that everyone here is right in that `Weak` probably doesn't need the `fmt::Debug` restriction, but I'd rather take the conservative side of things here: if the restriction isn't costing us much and putting it back would be difficult should we turn out to be wrong, then I'd rather just keep it.

@rfcbot concern better-use-case

yaahc:
Hey @geeklint, are you still interested in pursuing this change?

geeklint:
If the consensus doesn't find the use case compelling enough, I'm content to close this issue.

If I may conclude with a more theoretical postulation, I would suggest the idea that the Weak impl is overly restrictive is dependent on the existence of a type which intentionally doesn't impl Debug. If the widely held understanding is that all types should have a Debug implementation, then I support that the case I ran into (where `T` does not impl Debug) is almost certainly a "bug" with `T`, rather than with the bound on Weak.

If someone can provide an example of a type which shouldn't implement Debug, then I think this change is worth reconsidering, but as I do not personally have such an example, I no longer have a convincing argument for this change.

joshtriplett:
I personally do think this is a sufficient use case already: `derive(Debug)` should work on types containing `Weak`. I don't think it's a given that you always need to manually implement `Debug` yourself for `Arc<T>` to want to derive `Debug` for a type containing a `Weak<T>`.

m-ou-se:
Some related context: The `Debug` implementation of `Mutex` tries to lock the mutex and either prints just 'locked' if it was locked, or it Debug prints the `T` within the `Mutex<T>` if it wasn't locked.

yaahc:
Nominating this for discussion in next weeks libs-api team meeting since it seems to be blocked currently.

m-ou-se:
@BurntSushi any updates on your position after the comments since your last comment?

BurntSushi:
I think I would personally still like to see a more compelling example. @joshtriplett would it be possible to show something where you aren't doing a debug impl by hand on `Arc` and want a `derive` on `Weak`?

To me, I don't have any argument _against_ doing this other than the standard conservative take: we can't undo this once done, so we should be sure we're on firm ground before doing so. I don't feel super strongly here, so I'm not necessarily looking for something super compelling.

geeklint:
I thought of a potentially valuable example where the `T` in Arc<T>/Rc<T>/Weak<T> would not implement Debug: trait objects.  While there's a manual impl of Debug for `dyn Any`, other traits don't get the same.  For instance, `Weak<dyn Fn()>` doesn't implement Debug (and user code can't implement Debug for it manually).

carbotaniuman:
I've come across a use-case here for an interpreter, where the only reason I can't derive `Debug` is because  `Weak` isn't `Debug`. The object basically stores an `Arc` to one object, and a `Weak` to two others in a sort of linked-list for reference.

thomcc:
This seems to have gotten the needed checkboxes, but the FCP never began. Does anybody know the invocation to trigger that?

BurntSushi:
@thomcc It's because I have a concern registered, which blocks FCP. I hadn't seen the latest replies here (or forgot about them), but I think it's good enough for me.

@rfcbot resolve better-use-case

dtolnay (approved):

Issue #88716, opened 2021-09-07
linux: std::process, insecure dropping of ancillary groups


bnoordhuis:
Commit 22ddcd1a13082b7be0fc99b720677efd2b733816 made libstd drops ancillary groups when uid == 0:
https://github.com/rust-lang/rust/blob/385f8e2078b00282a7a7ffcd58aad17b405f57bf/library/std/src/sys/unix/process/process_unix.rs#L312-L314
Before that it unconditionally dropped group membership.

The new logic is wrong on Linux: it doesn't account for processes whose uid != 0 but have the CAP_SETGID capability.

Such processes can and should drop ancillary groups, otherwise child processes inherit permissions they otherwise wouldn't have.

Suggested change:
```rust
if self.get_groups().is_none() {
    let _ = libc::setgroups(0, ptr::null()); // or return unless EPERM
}
```

joshtriplett:
This seems like a potentially reasonable approach. It'd be a behavior change, but technically the previous change was a behavior change in the case of CAP_SETGID.

(It seems worth noting that CAP_SETGID is generally root-equivalent, so this doesn't actually seem like a security issue of any kind.)

cuviper:
Whatever we do, we need to reconcile this with #90292 as it becomes documented.
I agree the suggested change makes sense though.

bnoordhuis:
> It seems worth noting that CAP_SETGID is generally root-equivalent, so this doesn't actually seem like a security issue of any kind.

IMO, it is. Scenario: non-root, sudo-like program that:

1. has CAP_SETGID + CAP_SETUID in order to switch users when spawning child processes, and
2. is a member of one or more supplementary groups in order to read or write sensitive files

Not dropping group membership is a catastrophic security breach in that setup.

joshtriplett:
Wouldn't it need to drop that capability though, before the exec?

In any case, this does still need fixing.

bnoordhuis:
> Wouldn't it need to drop that capability though, before the exec?

Not necessarily, it depends on the capability set (threads have several) the capability is part of. Some persist across execve() calls, some don't.

(I wanted to write up a short explanation of the algorithm but I can't hope to do it justice in a few words, it's complicated.)

I'm pessimistic checking AT_SECURE is sufficient. `man 7 capabilities` claims the following:

> If  ambient  capabilities cause  a  process's  permitted  and  effective  capabilities  to  increase  during an execve(2), this does not trigger the secure-execution mode described in ld.so(8).

[...]

PR #90292, opened 2021-10-26
Document `setgroups` call caused by `std::os::unix::process::CommandExt.uid`

### LINKED ISSUE #39186: Document that std::os::unix::process::CommandExt.uid() triggers setgroups(0,0)
While digging around for something else I have just discovered that setting the UID for a subprocess with `CommandExt.uid()` also causes a call to `setgroups(0,0)` in the child.  This should be documented.

(Related to #38527)

Elliot-Roberts:
Fixes  #39186
It's a small change but this is my first time contributing to open source so any feedback is very appreciated.

I also have some questions:
I documented that the `setgroup` call only happens when the parent is root. But this is potentially not the desired behavior, as described in #88716.
What more needs to happen decision-wise for that? Does #88716 remaining open mean that it is an approved and wanted change? I could fix that issue in this PR too if that would be ok.

### DIFF
--- a/library/std/src/os/unix/process.rs
+++ b/library/std/src/os/unix/process.rs
@@ -21,6 +21,14 @@ pub trait CommandExt: Sealed {
     /// Sets the child process's user ID. This translates to a
     /// `setuid` call in the child process. Failure in the `setuid`
     /// call will cause the spawn to fail.
+    ///
+    /// # Notes
+    ///
+    /// This will also trigger a call to `setgroups(0, NULL)` in the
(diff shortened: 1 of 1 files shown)


Amanieu:
I agree that would should document this, however we should first resolve #88716 to decide the actual behavior we want to stabilize.

JohnCSimon:
triage: 
looks like #88716 is still waiting on resolution

Elliot-Roberts:
Closing in favor of #95982

PR #90297, opened 2021-10-26
Append .0 to unsuffixed float if it would otherwise become int token

### LINKED ISSUE #1085: Negative LitFloat seems to parse incorrectly in procedural macro
I have come across a weird issue in my procedural macro:

I parse a float, like so:
```
let float = syn::Lit::Float(syn::LitFloat::new("-10.0", Span::call_site()));
```

I would expect `float` to be: `Float(LitFloat { token: - 10.0 })`. However, it is `Float(LitFloat { token: - 10 })` instead. This causes the code generated by my macro to miscompile, since `-10` is not interpreted as a float.

Weirdly, this only happens when the parsing is done inside a procedural macro. If I do the parsing in a regular binary it works fine.

I have a reproduction case [here](https://github.com/sagacity/syn-float-token-issue). Run it with `cargo run --example example` and it will panic with the parsed float showing as: `Float(LitFloat { token: - 10 })`. I would have expected `Float(LitFloat { token: - 10.0 })` instead.

I've tried this both on Rust 1.55.0 and the currently nightly, on MacOS.

dtolnay:
Previously the unsuffixed f32/f64 constructors of `proc_macro::Literal` would create literal tokens that are definitely not a float:

```rust
Literal::f32_unsuffixed(10.0)  // 10
Literal::f32_suffixed(10.0)    // 10f32
Literal::f64_unsuffixed(10.0)  // 10
Literal::f64_suffixed(10.0)    // 10f64
```

Notice that the `10` are actually integer tokens if you were to reparse them, not float tokens.

This diff updates `Literal::f32_unsuffixed` and `Literal::f64_unsuffixed` to produce tokens that unambiguously parse as a float. This matches longstanding behavior of the proc-macro2 crate's implementation of these APIs dating back at least 3.5 years, so it's likely an unobjectionable behavior.

```rust
Literal::f32_unsuffixed(10.0)  // 10.0
Literal::f32_suffixed(10.0)    // 10f32
Literal::f64_unsuffixed(10.0)  // 10.0
Literal::f64_suffixed(10.0)    // 10f64
```

Fixes https://github.com/dtolnay/syn/issues/1085.

### DIFF
--- a/library/proc_macro/src/lib.rs
+++ b/library/proc_macro/src/lib.rs
@@ -1074,7 +1074,11 @@ impl Literal {
         if !n.is_finite() {
             panic!("Invalid float literal {}", n);
         }
-        Literal(bridge::client::Literal::float(&n.to_string()))
+        let mut repr = n.to_string();
+        if !repr.contains('.') {
+            repr.push_str(".0");
+        }
+        Literal(bridge::client::Literal::float(&repr))
     }
 
     /// Creates a new suffixed floating-point literal.
@@ -1115,7 +1119,11 @@ impl Literal {
         if !n.is_finite() {
             panic!("Invalid float literal {}", n);
         }
-        Literal(bridge::client::Literal::float(&n.to_string()))
+        let mut repr = n.to_string();
+        if !repr.contains('.') {
+            repr.push_str(".0");
+        }
+        Literal(bridge::client::Literal::float(&repr))
     }
 
     /// Creates a new suffixed floating-point literal.
--- a/src/test/ui/proc-macro/auxiliary/api/parse.rs
+++ b/src/test/ui/proc-macro/auxiliary/api/parse.rs
@@ -1,3 +1,5 @@
+// ignore-tidy-linelength
+
 use proc_macro::Literal;
 
 pub fn test() {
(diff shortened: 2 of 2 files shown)


jhpratt (approved):

library/proc_macro/src/lib.rs
```diff
@@ -1074,7 +1074,11 @@ impl Literal {
         if !n.is_finite() {
             panic!("Invalid float literal {}", n);
         }
-        Literal(bridge::client::Literal::float(&n.to_string()))
+        let mut repr = n.to_string();
+        if !repr.contains('.') {
+            repr.push_str(".0");
```
petrochenkov, line 1079: From reading the standard library code it looks like `to_string` should never produce an exponential form, so this shouldn't result in something like `1e100.0`.

The `Debug` impl for floats does exactly what we need (`10.0`) for small floats, but it does produce exponential forms for large floats.

Too bad `fn float_to_decimal_common_shortest` in `library\core\src\fmt\float.rs` is private, it can do exactly what we need here (although in theory it can be made public, but unstable and hidden, for `library/proc_macro`).

It's actually unfortunate that there's no public way to "print float as a float" in the standard library.
I wonder why the `Display` impl doesn't use the minimal precision 1 by default.

petrochenkov:
@dtolnay 
r=me after reading (and optionally answering) the comment https://github.com/rust-lang/rust/pull/90297#discussion_r743803242.

library/proc_macro/src/lib.rs
dtolnay: I added a test involving `1e100` so that at least something breaks if `to_string` starts turning it into `1e100.0`, though I don't expect that change to ever get made.

PR #90328, opened 2021-10-26
Move the Error trait into core (without `fn backtrace`)

yaahc:
current status: it works!


JohnCSimon:
Returning to author to address broken tests

compiler/rustc_typeck/src/check/method/probe.rs
```diff
@@ -612,8 +612,55 @@ impl<'a, 'tcx> ProbeContext<'a, 'tcx> {
...
+                    }
+                    (true, true) => {
+                        let lang_def_id = lang_items.errorsendsync_impl();
+                        self.assemble_inherent_impl_for_primitive(lang_def_id);
+                        let lang_def_id = lang_items.errorsendsync_alloc_impl();
+                        self.assemble_inherent_impl_for_primitive(lang_def_id);
+                    }
+                    (false, true) => panic!("unexpected combination of marker traits"),
+                }
+            }
```
yaahc, line 663: This bit is currently incorrect, I yolo'ed it to see what would happen and it was not good. Trying to get a better understanding of what `assemble_probe` actually does now before proceeding.

library/core/src/str/mod.rs
```diff
@@ -2549,3 +2549,7 @@ impl_fn_for_zst! {
         unsafe { from_utf8_unchecked(bytes) }
     };
 }
+
+#[stable(feature = "rust1", since = "1.0.0")]
```
yaahc, line 2553: presumably this stability attribute needs to get changed

library/std/src/error.rs
```diff
@@ -16,23 +16,45 @@
...
 use crate::str;
+#[cfg(bootstrap)]
 use crate::string;
+#[cfg(bootstrap)]
 use crate::sync::Arc;
+#[cfg(bootstrap)]
 use crate::time;
 
+#[cfg(not(bootstrap))]
+#[stable(feature = "rust1", since = "1.0.0")]
```
yaahc, line 55: This stability attribute I'm much less sure about...

yaahc:
There are a couple of ui tests that are broken because of the new lang item for `Error` which I'm just going to leave broken for right now.

compiler/rustc_typeck/src/coherence/inherent_impls.rs
```diff
@@ -49,13 +49,61 @@ impl ItemLikeVisitor<'v> for InherentCollect<'tcx> {
 
         let self_ty = self.tcx.type_of(item.def_id);
         let lang_items = self.tcx.lang_items();
+        let error_def_id = self.tcx.require_lang_item(hir::LangItem::Error, None);
+        let send_def_id = self.tcx.require_lang_item(hir::LangItem::Send, None);
+        let sync_def_id = self.tcx.require_lang_item(hir::LangItem::Sync, None);
```
bjorn3, line 54: Can you move these into the `Dynamic` case or make them optional? That would avoid having to change `#![no_core]` tests.

yaahc: oh, absolutely, thank you for pointing that out!

yaahc:
@nrc I went ahead and tested out integrating the error in core work with the provider api to implement generic member access in https://github.com/rust-lang/rust/pull/90328/commits/e171b46c317cb7b4eea7cf9557b8b802fecc1b00

Notes:
* I ran into issues calling `request_by_type_tag` on a `&dyn Error` since it required `&dyn Provider` which I couldn't get without enabling the incomplete `trait_upcasting` feature
  * I worked around this by changing the type for the provider in that function to be generic instead of a trait object
* The original plan outlined in the provider RFC for using a blanket impl and extension trait make it so we have to implement generic member access and move to `core` all at the same time, otherwise the `impl<T> Provider for T where T: Error` impl would be an orphan impl

We can probably resolve both of these issues by making the impl for `Error` more in line with the bits it stole from `Any`, where we would reimplement the methods on the `Provider` trait rather than using it as an extension. I almost just did this initially with `Error` still in `std` but needed to make `ConcreteRequisition` pub at which point I backtracked to the original plan.

Otherwise everything seemed to work as expected! I didn't do the complete implementation here, such as the additional copies of `request_ref` for `Send` and `Sync` errors, or the value / tag based getter equivalents. I also didn't split this up into its own feature yet, since I'm not really sure how that will interact with `Requisition` being part of the API for one of the methods on `Error`.

yaahc:
Perhaps even more ambitiously I've added https://github.com/rust-lang/rust/pull/90328/commits/c6dab38fda41ac2843db516c329f7e788c5f9ecf which changes the default implementation of `source` to use generic member access. I like this because I think it really demonstrates how this new API relates to the existing APIs on `Error`.

src/test/ui/error-trait/error-in-core-no-feature.rs
```diff
@@ -0,0 +1,18 @@
...
+#[derive(Debug)]
+struct MyCoreError(());
+
+impl fmt::Display for MyCoreError {
+    fn fmt(&self, fmt: &mut fmt::Formatter) -> fmt::Result {
+        unimplemented!()
+    }
+}
+
+impl core::error::Error for MyCoreError {} //~ ERROR use of unstable library feature 'error_in_core'
```
yaahc, line 16: Turns out that the `unstable` attributes don't actually work the way I expected them to. I had expected this to be an error the same way the it is in the above test but it actually compiles just fine...

Dylan-DPC:
@yaahc any updates?

yaahc:
This is superceded by https://github.com/rust-lang/rust/pull/98072 and a subsequent PR to move `Error` into core once https://github.com/rust-lang/rust/pull/95956 has landed

Issue #88971, opened 2021-09-15
regression: int_roundings conflicts with existing APIs


Mark-Simulacrum:
Crater run for 1.56 detected a considerable amount of breakage due to int_roundings on nightly. We have some support in the compiler to avoid stable breakage due to unstable features, but it does not work for this case (@cuviper theorized this is due to "the self inherent method is resolved before the &self trait method is considered at all."

There are:

* 236 regressed crates
* 580 regressed github repositories
* Roughly ~10 roots, but some may be masked by other errors.

This is a level of breakage significant enough that it seems prudent to back this out or rename the functions, to avoid breaking a good portion of the ecosystem.

cc:

* #88581
* https://github.com/rust-lang/rust/issues/88878
* https://rust-lang.zulipchat.com/#narrow/stream/241545-t-release/topic/possible.20semi-breaking.20change.20in.20next.20release

cuviper:
This may be a nitpick on what is defined as a root cause, but `num-integer` itself shouldn't be failing to compile as a dependency. The direct log errors you show are only in its tests, fixed in rust-num/num-integer#44 but not yet published. Other crate errors come from use of its `Integer` trait, so it is a root cause in that sense, but I can't fix that in `num-integer`.

cuviper:
I know it's not a blame game or anything, but I think the distinction matters when considering how many places have to be fixed to deal with this change.

Mark-Simulacrum:
Yeah, unfortunately, with this scope of breakage the number of root causes may not be terribly important -- many of those need somewhat manual intervention for the .lock bumps, in practice, which can take time.

joshtriplett:
These are unstable methods, so can we come up with reasonable names to rename them to, knowing that we don't have to commit to those names long-term if we come up with better ones before stabilization?

joshtriplett:
Conclusion from the libs-api meeting: let's temporarily rename these to something non-conflicting that we know we don't want to use long-term; for instance, let's prefix them with `unstable_`.

m-ou-se:
Discussed in the libs-api meeting. We'll rename them for now and beta-backport that.

joshtriplett:
Long-term fix:

- We should fix the unstable-vs-stable resolution here: if the compiler finds an inherent method that's unstable and not opted into, and we're going to error out anyway, keep going and see if we find a stable method we can use.
- We could potentially do autoderef here, since the type is `Copy`.

joshtriplett:
[...]

PR #90329, opened 2021-10-26
Try all stable method candidates first before trying unstable ones

nbdd0121:
Currently we try methods in this order in each step:
* Stable by value
* Unstable by value
* Stable autoref
* Unstable autoref
* ...

This PR changes it to first try pick methods without any unstable candidates, and if none is found, try again to pick unstable ones.

Fix #90320
CC #88971, hopefully would allow us to rename the "unstable_*" methods for integer impls back.

@rustbot label T-compiler T-libs-api

### DIFF
--- a/compiler/rustc_typeck/src/check/method/probe.rs
+++ b/compiler/rustc_typeck/src/check/method/probe.rs
@@ -92,7 +92,7 @@ impl<'a, 'tcx> Deref for ProbeContext<'a, 'tcx> {
     }
 }
 
-#[derive(Debug)]
+#[derive(Debug, Clone)]
 struct Candidate<'tcx> {
     // Candidates are (I'm not quite sure, but they are mostly) basically
     // some metadata on top of a `ty::AssocItem` (without substs).
@@ -132,7 +132,7 @@ struct Candidate<'tcx> {
     import_ids: SmallVec<[LocalDefId; 1]>,
 }
 
-#[derive(Debug)]
+#[derive(Debug, Clone)]
 enum CandidateKind<'tcx> {
     InherentImplCandidate(
         SubstsRef<'tcx>,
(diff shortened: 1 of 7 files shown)


joshtriplett:
Seems reasonable to me! I look forward to a PR un-renaming the integer rounding methods.

compiler/rustc_typeck/src/check/method/probe.rs
```diff
@@ -1102,8 +1103,75 @@ impl<'a, 'tcx> ProbeContext<'a, 'tcx> {
...
+            .next();
+
+        // Fast path if there are no unstable candidates.
+        if unstable_candidates.is_empty() {
+            return pick;
+        }
+
+        if let Some(pick) = pick {
+            if !self.is_suggestion.0 {
+                if let Ok(p) = &pick {
```
nagisa, line 1163: Nit: This likely would be slightly easier to follow as a metch:

```rust
match pick {
    // Emit a lint if there are unstable candidates alongside the stable ones.
    // We suppress warning if we're picking the method only because it is a
    // suggestion.
    Some(Ok(p)) if !self.is_suggestion.0 => {
        self.emit_unstable_name_collision_hint(p, &unstable_candidates);
        return pick;
    }
    Some(_) => return pick,
    _  => ()
}
```

compiler/rustc_typeck/src/check/method/probe.rs
```diff
@@ -1102,8 +1103,75 @@ impl<'a, 'tcx> ProbeContext<'a, 'tcx> {
...
+                                Some(&mut unstable_candidates),
+                            )
+                        })
+                        .or_else(|| {
+                            self.pick_const_ptr_method(
+                                step,
+                                self_ty,
+                                Some(&mut unstable_candidates),
+                            )
+                        })
```
nagisa, line 1150: Would be ideal if we could deduplicate this section of code between code picking a stable method and one picking an unstable one. Otherwise it seems like we'd be prone to changing one but not the other and introducing an inconsistency in the future.

nagisa (commented):
I don't see an inherent issue with the idea – the unstable methods are… unstable, so I don't see a huge problem in de-emphasizing them, even if it could change the behaviour of the code that targets nightly rustc.

With that in mind, though, I would like to see a couple other changes to the PR:

First is a `-Z` debug flag to revert to the previous order of pick priority. Such a flag would help significantly with a potential situation where we receive a bug about a behaviour change to somebody's nightly code.

And then I would love to see a test that verifies this new pick order between stable and unstable items expanded significantly. The existing test seems pretty anemic to me in that it only verifies a stable autorefd method is preferred to an unstable by-value one. But then there are cases like mutable vs immutable autoref (e.g. do we still pick an unstable method autoref if a mutable autoref necessary for a stable method is not valid?), const ptr methods, etc.

nbdd0121:
> I don't see an inherent issue with the idea – the unstable methods are… unstable, so I don't see a huge problem in de-emphasizing them, even if it could change the behaviour of the code that targets nightly rustc.

No; this wouldn't change the behaviour. A unstable method with corresponding feature enabled is not considered an "unstable_candidate"; only unstable methods with feature gates not enabled are considered unstable candidates. So code that already compiles shouldn't be affected.

> With that in mind, though, I would like to see a couple other changes to the PR:
> 
> First is a `-Z` debug flag to revert to the previous order of pick priority. Such a flag would help significantly with a potential situation where we receive a bug about a behaviour change to somebody's nightly code.

Added.

> And then I would love to see a test that verifies this new pick order between stable and unstable items expanded significantly. The existing test seems pretty anemic to me in that it only verifies a stable autorefd method is preferred to an unstable by-value one. But then there are cases like mutable vs immutable autoref (e.g. do we still pick an unstable method autoref if a mutable autoref necessary for a stable method is not valid?), const ptr methods, etc.

I expanded the test a bit. For unstable autoref vs a stable mutable autoref on a non-mutable variable scenario, the code will now pick the stable mutable autoref instead of the unstable autoref (method picking doesn't have this information; actually whether Deref or DerefMut is used depends on which method gets picked). So the error will change in this scenario. But as explained above, this wouldn't affect code that currently compiles.

nagisa:
Nice, thanks!

@bors r+ rollup=never

nbdd0121:
> Is my understanding correct that we may want to consider this for a backport to beta?

We have #90329 which is deemed acceptable breakage by libs team, and #88971 which is reverted in #89184, so I don't think this needs backport.

PR #90330, opened 2021-10-26
Update PlaceTy to use ty::Variant

zhamlin:
Part 1 of the plan defined [here](https://github.com/rust-lang/rust/pull/89745#issuecomment-941169095)

### DIFF
--- a/compiler/rustc_borrowck/src/places_conflict.rs
+++ b/compiler/rustc_borrowck/src/places_conflict.rs
@@ -243,10 +243,10 @@ fn place_components_conflict<'tcx>(
                     return false;
                 }
 
-                (ProjectionElem::Field { .. }, ty::Adt(def, _), AccessDepth::Drop) => {
+                (ProjectionElem::Field { .. }, _, AccessDepth::Drop) => {
                     // Drop can read/write arbitrary projections, so places
                     // conflict regardless of further projections.
-                    if def.has_dtor(tcx) {
+                    if base_ty.has_dtor(tcx) {
                         return true;
                     }
                 }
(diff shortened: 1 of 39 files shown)


compiler/rustc_borrowck/src/places_conflict.rs
```diff
@@ -211,7 +211,7 @@ fn place_components_conflict<'tcx>(
             let proj_base = &borrow_place.projection[..access_place.projection.len() + i];
             let base_ty = Place::ty_from(borrow_local, proj_base, body, tcx).ty;
 
-            match (elem, &base_ty.kind(), access) {
+            match (elem, &base_ty.strip_variant_type().kind(), access) {
```
oli-obk, line 214: We should document each such use to explain why it doesn't matter. E.g. this case is because we only care about whether the type has a drop impl itself. Though we could just add a method to Ty that returns whether it has a drop impl and make that function work for both variant and adt.

compiler/rustc_trait_selection/src/traits/select/mod.rs
```diff
@@ -1930,7 +1934,7 @@ impl<'cx, 'tcx> SelectionContext<'cx, 'tcx> {
         &self,
         t: ty::Binder<'tcx, Ty<'tcx>>,
     ) -> ty::Binder<'tcx, Vec<Ty<'tcx>>> {
-        match *t.skip_binder().kind() {
+        match *t.skip_binder().strip_variant_type().kind() {
```
oli-obk, line 1937: I think we should not strip here, but instead recurse

oli-obk (changes requested):
I didn't go through the unimplemented! calls yet

compiler/rustc_borrowck/src/places_conflict.rs
zhamlin: I went with the adding a method to Ty approach, seemed better than using strip_variant_type here.

zhamlin:
For the unimplemented!, I can go ahead and update them to use recursion where possible.

PR #90333, opened 2021-10-27
rustdoc: remove flicker during page load

### LINKED ISSUE #90246: rustdoc: search box flickers on page load
Steps to reproduce:

1. Visit, e.g. https://doc.rust-lang.org/std/string/trait.ToString.html
2. Open Chrome Developer Tools
3. Click the Sources tab, then main1.56.0.js.
4. Click the first line of main1.56.0.js to set a breakpoint.
5. Reload.
6. Notice that the search box is grey, but it will later become white.

flicker-on-load

The search box has for a long time been disabled on page load, until the JS runs. It being grey is a side effect of https://github.com/rust-lang/rust/pull/85367. Before that, our `search-focus:disabled` style wasn't getting applied. After that, it is. As a quick fix I think we should remove the grey styling on the disabled search bar. I know this means that no-JS users won't have as good a visual indication that the search bar is disabled. Perhaps we should address this by having styles inside a `<noscript>` tag that hide the search bar entirely.

It's good to minimize the number of changes that happen during page load - this reduces visual distractions and help page loads feel fast.

jsha:
The search bar has a `:disabled` style that makes it grey, which creates a distracting flicker from grey to white when the page finishes loading. The search bar should stay the same color throughout page load.

A blank white search bar might create an incorrect impression for users with JS turned off. Since they can't use the search functionality, I've hidden the search bar in noscript.css.

Fixes #90246
r? @GuillaumeGomez 

Demo: https://rustdoc.crud.net/jsha/flashy-searchbar/std/string/struct.String.html

### DIFF
--- a/src/librustdoc/html/static/js/main.js
+++ b/src/librustdoc/html/static/js/main.js
@@ -263,7 +263,9 @@ function hideThemeButtonState() {
                 search_input.placeholder = searchState.input.origPlaceholder;
             });
 
-            search_input.removeAttribute('disabled');
+            if (search_input.value != '') {
+                loadSearch();
+            }
 
             // `crates{version}.js` should always be loaded before this script, so we can use it
             // safely.
--- a/src/test/rustdoc-gui/javascript-disabled.goml
+++ b/src/test/rustdoc-gui/javascript-disabled.goml
@@ -1,0 +1,6 @@
+// When JavaScript is disabled, we hide the search bar, because it
+// can't be used without JS.
+javascript: false
+
+goto: file://|DOC_PATH|/test_docs/struct.Foo.html
+assert-css: (".sub", {"display": "none"})
--- a/src/librustdoc/html/static/css/noscript.css
+++ b/src/librustdoc/html/static/css/noscript.css
@@ -13,3 +13,12 @@ rules.
 	/* It requires JS to work so no need to display it in this case. */
 	display: none;
 }
+
+.sub {
+	/* The search bar and related controls don't work without JS */
+	display: none;
+}
+
+#theme-picker {
+	display: none;
+}
--- a/src/librustdoc/html/static/css/themes/ayu.css
+++ b/src/librustdoc/html/static/css/themes/ayu.css
@@ -255,10 +255,6 @@ details.undocumented > summary::before {
 	box-shadow: 0 0 0 1px #148099,0 0 0 2px transparent;
 }
 
-.search-input:disabled {
-	background-color: #3e3e3e;
-}
-
 .module-item .stab,
 .import-item .stab {
 	color: #000;
--- a/src/librustdoc/html/static/css/themes/dark.css
+++ b/src/librustdoc/html/static/css/themes/dark.css
@@ -219,10 +219,6 @@ details.undocumented > summary::before {
 	border-color: #008dfd;
 }
 
-.search-input:disabled {
-	background-color: #c5c4c4;
-}
-
 #crate-search + .search-input:focus {
 	box-shadow: 0 0 8px 4px #078dd8;
 }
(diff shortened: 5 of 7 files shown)


jsha:
Updated to remove the color in noscript.

GuillaumeGomez:
So now, we can enter things in the input, it's just that nothing will happen... Is there a CSS way to handle such case?

jyn514:
@GuillaumeGomez I don't understand why you asked to remove the grey coloring. It looks fine in all three different themes, and AFAICT you didn't have an alternative in mind.

GuillaumeGomez:
I didn't asked the background colour to be removed. Also, I'm currently complaining about the behaviour when JS is disabled.

jsha:
> So now, we can enter things in the input, it's just that nothing will happen... Is there a CSS way to handle such case?

Unfortunately, no. The way to prevent entering things in the input is with the HTML `disabled` attributes, and you can't set HTML attributes from CSS. I guess we could say `search-input:focused { background-color: red }` or something similar in `<noscript>`, but that seems awkward..

GuillaumeGomez:
Can't we keep the `disabled` part and simply remove it at the beginning of the JS script?

Or maybe we should go with what you suggested: simply not displaying the buttons and the search input at all. That seems like the best and simplest solution. Sorry for making you go back to it...

If you go with this approach, please add a GUI test for it (you can disable js).

jsha:
> Can't we keep the disabled part and simply remove it at the beginning of the JS script?

@jyn514 made a good point that someone facing a slow loading page may want to start typing in the search before the JS is loaded. It's fairly easy to support that, so I added that in this PR. Going back to using the disabled attribute would prevent that use case.

> Or maybe we should go with what you suggested: simply not displaying the buttons and the search input at all. That seems like the best and simplest solution.

Great, I also think this is the nicest solution. :-)

src/test/rustdoc-gui/javascript-disabled.goml
```diff
@@ -0,0 +1,5 @@
+javascript: false
```
GuillaumeGomez, line 1: Please add a little explanation comment at the top.

src/test/rustdoc-gui/javascript-disabled.goml
```diff
@@ -0,0 +1,5 @@
...
+
+goto: file://|DOC_PATH|/test_docs/struct.Foo.html
+assert-css: (".sub", {"display": "none"})
+
```
GuillaumeGomez, line 5: Please remove this empty line.

src/librustdoc/html/static/js/main.js
```diff
@@ -263,7 +263,9 @@ function hideThemeButtonState() {
                 search_input.placeholder = searchState.input.origPlaceholder;
             });
 
-            search_input.removeAttribute('disabled');
+            if (search_input.value != '') {
+                loadSearch();
```
GuillaumeGomez, line 267: Can you check how many times the `search` function is called when you load a page with `?search=something`? I found a few cases before where the function was called more than once. Just to be sure it won't be triggered by this.

jsha: I just checked, and found that master currently calls `search` twice when you load with `?search=something`, and this branch also calls it twice. So there is a problem, and we should fix it, but it's not introduced or made worse by this PR.

GuillaumeGomez: Ah so the bug is bug. Well if it's not from this PR then it's fine!

GuillaumeGomez (approved):

GuillaumeGomez:
r=me once CI pass. I'll open an issue for the double call on the search.

jsha:
Just opened an issue :-D https://github.com/rust-lang/rust/issues/90454

PR #90341, opened 2021-10-27
[perf-check] Revert "Rollup merge of #88860 - nbdd0121:panic, r=m-ou-se"

JohnTitor:
This reverts commit f7024998c7eae5b582de680effaca105198aa95c, reversing
changes made to 84fe598f0094c8d6a1dddcdc38ac0fa28df32f44.

r? @ghost cc https://github.com/rust-lang/rust/pull/90067#issuecomment-952080100

### DIFF
--- a/compiler/rustc_const_eval/src/const_eval/machine.rs
+++ b/compiler/rustc_const_eval/src/const_eval/machine.rs
@@ -72,7 +72,9 @@ impl<'mir, 'tcx> InterpCx<'mir, 'tcx, CompileTimeInterpreter<'mir, 'tcx>> {
             let span = self.find_closest_untracked_caller_location();
             let (file, line, col) = self.location_triple_for_span(span);
             return Err(ConstEvalErrKind::Panic { msg, file, line, col }.into());
-        } else if Some(def_id) == self.tcx.lang_items().panic_fmt() {
+        } else if Some(def_id) == self.tcx.lang_items().panic_fmt()
+            || Some(def_id) == self.tcx.lang_items().begin_panic_fmt()
+        {
             // For panic_fmt, call const_panic_fmt instead.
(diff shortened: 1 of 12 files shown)


jackh726:
I think this is more a T-libs PR than T-compiler.

JohnCSimon:
@JohnTitor Can you please address the merge conflicts?

JohnTitor:
I'd like to see @m-ou-se's (or someone from t-libs) thoughts if this is worth merging before resolving the conflicts.

m-ou-se:
Do we know why and how this change affects performance? The perf results don't show a lot of difference.

JohnTitor:
> Do we know why and how this change affects performance? The perf results don't show a lot of difference.

The regression we found is https://github.com/rust-lang/rust/pull/90067#issuecomment-947755499, according to the bot: `Moderate regression in instruction counts (up to 1.1% on incr-patched: b9b3e592dd cherry picked builds of style-servo)`
And the perf result here is: `Small improvement in instruction counts (up to -1.0% on incr-patched: b9b3e592dd cherry picked builds of style-servo)`
So we could assume the regression came from that PR, I think.

JohnTitor:
> > So we could assume the regression came from that PR, I think.
> 
> Sure, but do we have any idea how it caused it? The difference is small enough that it could be something unrelated.

Not really, because this PR was created on 27th Oct 2021 and I don't remember the time I investigated. I think you're more familiar with the changes as you're the reviewer of it. Looking at the benchmark code (https://github.com/rust-lang/rustc-perf/tree/master/collector/benchmarks/style-servo) would be helpful. And the point here is if it's worth reverting, I guess.

Dylan-DPC:
@JohnTitor do we still need this pr?

m-ou-se:
Since the impact is minimal and we don't even know how that PR caused the small regression, I think we should just close this.

Issue #65225, opened 2019-10-08
Tracking issue for `#![feature(entry_insert)]`


passcod:
This is a tracking issue for the `Entry::insert_entry` method on HashMap and BTreeMap introduced in https://github.com/rust-lang/rust/pull/60142#issuecomment-487416553.

- [x] Implementation for HashMap: #64656
	- [x] ~~Stabilised in 1.59: #90345~~ Re-stabilised in 1.83: #130290
	  - ~~De-stabilized in #94105~~
  ```rust
   impl<'a, K, V> Entry<'a, K, V> {
      pub fn insert_entry(self, value: V) -> OccupiedEntry<'a, K, V> {…}
  }
  ```
- [x] Implementation for BTreeMap: #133042 
    - [ ] Stabilization

TethysSvensson:
This does not appear to be compatible with #44286, as the following code will panic.

```rust
#![feature(map_entry_replace, entry_insert)]
use std::collections::HashMap;

fn main() {
    HashMap::new().entry(0).insert(0).replace_key();
}
```

It seems to me that the two APIs are fundamentally incompatible, or at least they are currently incompatible without panicking. 

The API in #44286 assumes that an `OccupiedEntry` will always refer to two keys, the existing one and the one used for the lookup, while this feature will break that assumption. In fact, the `key` field in `OccupiedEntry` and `RustcOccupiedEntry` in hashbrown could be a `K` instead of an `Option<K>` if this API was not supported.

TethysSvensson:
It seems to me that it is a bit silly to use an `Option` to track at run-time what we should already know at compile time. Wouldn't it make more sense to have a different type from `OccupiedEntry`, which did not contain a key?

I could imagine this would also be useful as a stepping stone towards implementing something like rust-lang/rfcs#1769.

passcod:
> if this API was not supported.


I'm a bit rusty with this PR, but am I correct in saying that the entry_insert API is impossible to replicate from user code, while the map_entry_replace one is, as per the original PR? (Albeit not with the same efficiency)


Of course I'd much rather a solution that supports both uses is found

TethysSvensson:
@passcod I believe you mostly are correct on both parts of your statement.

I do not think the `Entry::insert` function can be implemented in safe rust, as it allows the creation of an `OccupiedEntry` without an associated key. I cannot see how it would be possible to do that in safe rust.

[...]

PR #64656, opened 2019-09-21
Implement (HashMap) Entry::insert as per #60142

passcod:
Implementation of `Entry::insert` as per @SimonSapin's comment on #60142. This requires a patch to hashbrown:

```diff
diff --git a/src/rustc_entry.rs b/src/rustc_entry.rs
index fefa5c3..7de8300 100644
--- a/src/rustc_entry.rs
+++ b/src/rustc_entry.rs
@@ -546,6 +546,32 @@ impl<'a, K, V> RustcVacantEntry<'a, K, V> {
         let bucket = self.table.insert_no_grow(self.hash, (self.key, value));
         unsafe { &mut bucket.as_mut().1 }
     }
+
+    /// Sets the value of the entry with the RustcVacantEntry's key,
+    /// and returns a RustcOccupiedEntry.
+    ///
+    /// # Examples
+    ///
+    /// ```
+    /// use hashbrown::HashMap;
+    /// use hashbrown::hash_map::RustcEntry;
+    ///
+    /// let mut map: HashMap<&str, u32> = HashMap::new();
+    ///
+    /// if let RustcEntry::Vacant(v) = map.rustc_entry("poneyland") {
+    ///     let o = v.insert_and_return(37);
+    ///     assert_eq!(o.get(), &37);
+    /// }
+    /// ```
+     #[inline]
+    pub fn insert_and_return(self, value: V) -> RustcOccupiedEntry<'a, K, V> {
+        let bucket = self.table.insert_no_grow(self.hash, (self.key, value));
+        RustcOccupiedEntry {
+            key: None,
+            elem: bucket,
+            table: self.table
+        }
+    }
 }
 
 impl<K, V> IterMut<'_, K, V> {
```

This is also only an implementation for HashMap. I tried implementing for BTreeMap, but I don't really understand BTreeMap's internals and require more guidance on implementing the equivalent `VacantEntry::insert_and_return` such that it returns an `OccupiedEntry`. Notably, following the original PR's modifications I end up needing a `Handle<NodeRef<marker::Mut<'_>, _, _, marker::LeafOrInternal>, _>` while I only have a `Handle<NodeRef<marker::Mut<'_>, _, _, marker::Internal>, _>` and don't know how to proceed.

(To be clear, I'm not asking for guidance right now; I'd be happy getting only the HashMap implementation — the subject of this PR — reviewed and ready, and leave the BTreeMap implementation for a latter PR.)

PR #90345, opened 2021-10-27
Stabilise entry_insert

passcod:
This stabilises `HashMap:Entry::insert_entry` etc. Tracking issue #65225. It will need an FCP.

This was implemented in #64656 two years ago.

This PR includes the rename and change discussed in https://github.com/rust-lang/rust/issues/65225#issuecomment-910652430, happy to split if needed.

### DIFF
--- a/library/std/src/collections/hash/map.rs
+++ b/library/std/src/collections/hash/map.rs
@@ -2460,7 +2460,7 @@ impl<'a, K, V> Entry<'a, K, V> {
     /// ```
     #[inline]
     #[unstable(feature = "entry_insert", issue = "65225")]
-    pub fn insert(self, value: V) -> OccupiedEntry<'a, K, V> {
+    pub fn insert_entry(self, value: V) -> OccupiedEntry<'a, K, V> {
         match self {
             Occupied(mut entry) => {
                 entry.insert(value);
--- a/library/std/src/collections/hash/map.rs
+++ b/library/std/src/collections/hash/map.rs
@@ -2450,7 +2450,6 @@ impl<'a, K, V> Entry<'a, K, V> {
     /// # Examples
     ///
     /// ```
-    /// #![feature(entry_insert)]
     /// use std::collections::HashMap;
     ///
     /// let mut map: HashMap<&str, String> = HashMap::new();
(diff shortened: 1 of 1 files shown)


joshtriplett:
@rfcbot merge

dtolnay (approved):

qm3ster:
I object, because this locks `OccupiedEntry.key` to be an `Option<K>`.
This is unnecessary, as until this feature all of the methods unwrapping it are assumed to be infallible, and will now start panicking, which is technically somewhat a breaking change.
This is also incompatible with #44286.
`OccupiedEntry.key` should become just `K`, and instead of putting an `OccupiedEntry` into a dysfunctional state, a new type should be introduced.
The new type, which is *only* a view into the map and doesn't hold an owned key, can then be used as an implementation detail of `OccupiedEntry`.
That type might also happen to be `RawOccupiedEntryMut` 🤔.

Amanieu:
Unfortunately I think that it's too late to revert the stabilization at this point: the stable release is coming out next week.

eaglgenes101:
I'll second not stabilizing this method, as the documentation makes no mention either on `insert_entry`or the key-getting methods that the key getting methods might panic if `insert_entry` is used. This is an important detail that users should know before using `insert_entry`, and the possibility of panic is usually prominently indicated in other `std` APIs. 

I do have other objections, but this is the most immediate and pertinent one.

Amanieu:
cc @rust-lang/libs-api There seem to be several last-minute objections to stabilizing this feature (which is due to be released in stable next week). Thoughts?

cuviper:
It's not too late to revert yet, but that decision should be made ASAP.

qm3ster:
Since unlike `hashbrown`, `std` doesn't have `Entry::insert(v) -> OccupiedEntry`, this can all be resolved very happily in a backward-compatible way.
At no point am I saying we should have https://github.com/rust-lang/rust/issues/44286 *instead* of this.
We can ergonomically have all of the semantics of both.

However, not reverting this now will permanently import this accidental complexity from hashbrown. We will never get rid of `key: Option`. We will never have statically-distinguished keyful and keyless entries.

The current situation is a consequence of not enough communication between the features, not a fundamental/hard problem.

m-ou-se:
These look like valid concerns to me. Reverting stabilization right now is a bit late, but not *too* late. If possible, I'd prefer delaying/reverting stabilizaiton for now, so we have time to discuss these concerns.

Mark-Simulacrum:
If we would like to revert stabilization, please file a PR targeting master and beta nominate+accept it; it can get picked up in the last round we typically do.

cuviper:
Just as a reminder, we'll need to remove it from the release notes and blog post as well.

yaahc:
Has anyone started working on the revert?

5225225:
i'll do it. currently AFK but I can get a PR up in 30ish minutes.

that would just be a revert of this PR, plus removing it from any release notes that may be stored in this repo? (not sure where they are).

PR #90348, opened 2021-10-27
Add features gates for experimental asm features

Amanieu:
This PR splits off parts of `asm!` into separate features because they are not ready for stabilization.

Specifically this adds:
- `asm_const` for `const` operands.
- `asm_sym` for `sym` operands.
- `asm_experimental_arch` for architectures other than x86, x86_64, arm, aarch64 and riscv.

r? @nagisa


joshtriplett:
Looks reasonable to me. r=me once it passes CI.

bnaecker:
I know this has already been merged, but I'm a bit concerned about this change. For context, I maintain the [`usdt` crate](https://github.com/oxidecomputer/usdt), which allows developers to insert DTrace probes into userland Rust code. For a variety of reasons, that crate works by generating declarative macros (macros-by-example) for each probe, whose expanded bodies contain inline assembly. Because the expansion site contains `asm!`, which requires a feature flag, developers using the `usdt` crate have previously been required to include a `#![feature(asm)]` in their own code that defines and uses probes. As of this change, there are now _two_ features required for downstream users of the `usdt` crate, `asm` and (at least on macOS platforms), `asm_sym`.

More generally, this manner in which this new flag was included seems fraught. As I understand it, any code with a call to `asm!` that uses a symbol operand will not compile across this threshold. _Existing_ code has been effectively hidden behind a _new_ feature flag.

I'm not sure how to resolve this. Since there is no support for compile-time detection of the Rust toolchain (that I'm aware of!), it doesn't seem possible to write a library that uses symbol operands, and that can be compiled across this change. Including the `asm_sym` flag will generate compiler errors on an older toolchain, since that feature flag is unknown in that case. But omitting the `asm_sym` flag will generate compiler errors on this toolchain. How can one write the set of feature flags required to build such a library for a range of compiler versions?

cc @ahl

PR #90349, opened 2021-10-27
Fix rare ICE during typeck in rustdoc scrape_examples

willcrichton:
While testing the `--scrape-examples` extension on the [wasmtime](https://github.com/bytecodealliance/wasmtime) repository, I found some additional edge cases. Specifically, when asking to typecheck a body containing a function call, I would sometimes get an ICE if:
* The body doesn't exist
* The function's HIR node didn't have a type

This adds checks for both of those conditions.

(Also this updates a test to check that the sources of a reverse-dependency are correctly generated and linked.)

r? @jyn514

### DIFF
--- a/src/librustdoc/scrape_examples.rs
+++ b/src/librustdoc/scrape_examples.rs
@@ -132,12 +132,28 @@ where
     fn visit_expr(&mut self, ex: &'tcx hir::Expr<'tcx>) {
         intravisit::walk_expr(self, ex);
 
-        // Get type of function if expression is a function call
         let tcx = self.tcx;
+
+        // If we visit an item that contains an expression outside a function body,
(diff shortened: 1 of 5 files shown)


jyn514 (changes requested):

jyn514 (changes requested):

willcrichton:
@jyn514 does that resolve your concern?

jyn514:
@willcrichton I think so - I'm still slightly confused how this can happen when visitor passes OnlyBodies, but not enough to block on it. Can you update the comment to say "the caller needs to ensure this is a function body" rather having a red herring about overriding typeck?

willcrichton:
Done!

PR #90352, opened 2021-10-27
Simplify `for` loop desugar

camsteffen:
Basically two intermediate bindings are inlined. I could have left one intermediate binding in place as this would simplify some diagnostic logic, but I think the difference in that regard would be negligible, so it is better to have a minimal HIR.

For checking that the pattern is irrefutable, I added a special case when the `match` is found to be non-exhaustive.

The reordering of the arms is purely stylistic. I don't *think* there are any perf implications.

```diff
  match IntoIterator::into_iter($head) {
      mut iter => {
          $label: loop {
-             let mut __next;
              match Iterator::next(&mut iter) {
-                 Some(val) => __next = val,
                  None => break,
+                 Some($pat) => $block,
              }
-             let $pat = __next;
-             $block
          }
      }
  }
```


compiler/rustc_mir_build/src/thir/pattern/check_match.rs
```diff
@@ -197,7 +194,10 @@ impl<'p, 'tcx> MatchVisitor<'_, 'p, 'tcx> {
 
         match source {
             hir::MatchSource::ForLoopDesugar | hir::MatchSource::Normal => {
-                report_arm_reachability(&cx, &report)
+                // skip the first desugared match in a for loop
+                if !(source == hir::MatchSource::ForLoopDesugar && arms.len() == 1) {
```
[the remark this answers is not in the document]
camsteffen: addressed this with a rebase

src/tools/clippy/tests/ui/pattern_type_mismatch/syntax.stderr
```diff
@@ -23,14 +23,6 @@ LL |     while let Some(_) = ref_value {
    |
    = help: use `*` to dereference the match expression or explicitly match against a `&_` pattern and adjust the enclosed variable bindings
 
-error: type of pattern does not match the expression type
```
oli-obk, line 26: Any clue why this one got lost? If not, @rust-lang/clippy should we track this in an issue to get it reinstated?

oli-obk:
The regressions seem to all be in `-opt` builds and solely part of the time spent in LLVM, so I'm hoping it's that more optimizations apply now (and worst case some optimizations require more work but don't result in better code).

@bors r+ rollup=never

camsteffen:
I think this is an improvement in the coverage report. Here the loop is entered 6 times and iterated 18 times.

```diff
    41|      6|fn display<T: Display>(xs: &[T]) {
-   42|     18|    for x in xs {
+   42|     24|    for x in xs {
+                      ^18
    43|     18|        print!("{}", x);
    44|     18|    }
    45|      6|    println!();
```

src/tools/clippy/tests/ui/pattern_type_mismatch/syntax.stderr
camsteffen: I went ahead and fixed this.

oli-obk:
Looks like more coverage tests:

```
Error: 1
  left: `2`,
 right: `3`: the argument was wrong', ../coverage/assert.rs:6:5
Error: 1
Error: 1
warning: function is never used: `do_not_add_coverage_not_called`
  --> ../coverage/no_cov_crate.rs:15:4
```

oli-obk:
> The regressions seem to all be in `-opt` builds and solely part of the time spent in LLVM, so I'm hoping it's that more optimizations apply now (and worst case some optimizations require more work but don't result in better code).

PR #90354, opened 2021-10-27
Document clippy on nightly-rustc

xFrednet:
Adding Clippy's docs to nightly-rustc, based on commit 01cf0bde. This PR only adds `clippy_utils` to the documentation. I've decided to only document one crate for now, as `clippy_lints` etc. contain very specific and undocumented functions which aren't really reusable. I'm guessing that they would mostly clutter up the search results with little benefit.

`./x.py --stage 1 doc src/tools/clippy` if working fine now after the help that `@jyn514` and `@ehuss` have provided. A big THANK YOU to them!

### DIFF
--- a/src/bootstrap/builder.rs
+++ b/src/bootstrap/builder.rs
@@ -482,6 +482,7 @@ impl<'a> Builder<'a> {
                 doc::RustByExample,
                 doc::RustcBook,
                 doc::CargoBook,
+                doc::Clippy,
                 doc::EmbeddedBook,
                 doc::EditionGuide,
             ),
--- a/src/bootstrap/doc.rs
+++ b/src/bootstrap/doc.rs
@@ -755,6 +755,7 @@ tool_doc!(
     "src/tools/rustfmt",
     ["rustfmt-nightly", "rustfmt-config_proc_macro"],
 );
+tool_doc!(Clippy, "clippy", "src/tools/clippy", ["clippy_utils"]);
 
 #[derive(Ord, PartialOrd, Debug, Copy, Clone, Hash, PartialEq, Eq)]
 pub struct ErrorIndex {


Mark-Simulacrum:
Hm, so the PR discussion seems to indicate that the was_explicitly_invoked call should be removed, but I do not see that in the diff. Did it perhaps forget to get pushed?

FWIW, I don't think that's quite the right fix. The exact way we do defaults here is a bit iffy, as @jyn514 mentions, and we don't have a great story around it, but just dropping the was-explicitly-invoked will break compiler-docs=false builds. I think the appropriate fix is a little more expansive:

* Gate tool doc builds on compiler-docs = true by default (i.e., add a `default_condition(run.builder.config.compiler_docs)` to the should_run method generated by the macro). They are all generating what is essentially compiler-docs anyway, so this makes sense IMO.
* Modify the default_condition on doc::Rustc to use compiler_docs, not regular docs.
* Remove the gate in doc::Rustc which checks for compiler-docs being enabled as well as the was_explicitly_invoked call, it's no longer needed, as we shouldn't reach that code unless (a) we've enabled compiler-docs or (b) some tool or rustc were explicitly invoked.

xFrednet:
> Hm, so the PR discussion seems to indicate that the was_explicitly_invoked call should be removed, but I do not see that in the diff. Did it perhaps forget to get pushed?

My intention with this PR was only to add Clippy to the `nightly-rustc` docs. I was at first unable to test the doc building locally due to some missing knowledge, which lead to the PR discussion.

I haven't really looked at the inner workings of the bootstrapping and docs building process. Most of the things you mentioned is new territory for me. I can try my hand on it with some guidance if you'd like. @Mark-Simulacrum

Mark-Simulacrum:
So this PR works as-is today?

xFrednet:
Yes! I can locally build the documentation, and it includes the `clippy_utils` crate.

I only needed help to figure out the comment and changes to the `config.toml` to do so. The complexity of this process was the starter of the discussion, how this could be improved and what causes it :)

Mark-Simulacrum:
@bors r+ rollup

Ok, seems good enough for now then.

PR #90356, opened 2021-10-27
Deny `#[rustc_const_stable]` on `#[unstable]` functions

### LINKED ISSUE #79551: Forbid `#[rustc_const_stable]` without `#[stable]`
Currently, it is possible for a function in std to be marked stably `const` without the function being stable. In #79548, it was noted that rustdoc assumes that stably const implies stable, and that seems to be a reasonable requirement. There should be an error added to ensure the consistency of const stability.

jhpratt:
This adds a lint against both function declarations and extern function declarations that are `#[unstable]` but declared `#[rustc_const_stable]`, as this is almost certainly a mistake.

Resolves #79551

@rustbot label A-const-fn A-lint A-stability C-enhancement T-libs

cc @oli-obk

### DIFF
--- a/compiler/rustc_lint/src/internal.rs
+++ b/compiler/rustc_lint/src/internal.rs
@@ -401,4 +401,29 @@ impl LateLintPass<'_> for IncompatibleStability {
             lint.build("functions cannot be const-stable if they are unstable").emit();
         });
     }
+
+    fn check_foreign_item(&mut self, cx: &LateContext<'_>, item: &rustc_hir::ForeignItem<'_>) {
+        use rustc_attr::{ConstStability, Stability, StabilityLevel};
+
+        if !matches!(item.kind, rustc_hir::ForeignItemKind::Fn(..)) {
+            return;
+        }
+
+        if !matches!(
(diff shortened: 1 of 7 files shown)


oli-obk:
> Also, was I correct in that I needed to gate the `#[allow]` behind bootstrapping? It makes sense intuitively.

Yes



> libcore doesn't have any instances of the lint firing other than intrinsics (silenced for well-known reasons) and the two erroneous places. Once stdarch is updated then CI should be green.

Cool! The PR lgtm

> You'd know the details of that attribute better than myself

My brain is a sieve XD I just thought we needed that, but looks like we either don't or we don't anymore.

jhpratt:
Turns out your understanding was correct — even private methods need to be `#[rustc_const_stable]`. While I don't necessarily agree with this, let's assume this isn't going to change. What should be done with this PR?

jhpratt:
Running into another use case for this as I tackle a bug I introduced a little bit ago re. const panic. As this doesn't seem to apply to _only_ intrinsics as suspected, I'm going to go ahead and close this. My recommendation is to close #79551 as well.

Issue #89554, opened 2021-10-05
Tracking Issue for `#![feature(non_exhaustive_omitted_patterns_lint)]`


Nadrieril:
This is a tracking issue for the `non_exhaustive_omitted_patterns` lint.
The feature gate for the issue is `#![feature(non_exhaustive_omitted_patterns_lint)]`.

### About tracking issues

Tracking issues are used to record the overall progress of implementation.
They are also used as hubs connecting to other relevant issues, e.g., bugs or open design questions.
A tracking issue is however *not* meant for large scale discussion, questions, or bug reports about a feature.
Instead, open a dedicated issue for the specific matter and add the relevant feature gate label.

### Steps

- [ ] Implemented in https://github.com/rust-lang/rust/pull/86809
- [ ] Set a feature gate https://github.com/rust-lang/rust/pull/89428
- [ ] Adjust documentation ([see instructions on rustc-dev-guide][doc-guide])
- [ ] Stabilization PR ([see instructions on rustc-dev-guide][stabilization-guide])

[stabilization-guide]: https://rustc-dev-guide.rust-lang.org/stabilization_guide.html#stabilization-pr
[doc-guide]: https://rustc-dev-guide.rust-lang.org/stabilization_guide.html#documentation-prs

### Unresolved Questions

- [ ] Lint name is potentially unclear https://github.com/rust-lang/rust/issues/89042#issuecomment-929464660

### Implementation history

- [ ] Implemented in https://github.com/rust-lang/rust/pull/86809
- [ ] ICE was discovered in https://github.com/rust-lang/rust/issues/89382 and fixed https://github.com/rust-lang/rust/pull/89423
- [ ] A bug was discovered in https://github.com/rust-lang/rust/issues/89042, about to be fixed with https://github.com/rust-lang/rust/pull/89105
- [ ] Feature gate PR https://github.com/rust-lang/rust/pull/89428

DevinR528:
We'll need one more PR/check box: Add filtering of unstable and hidden fields to structs.

Kampfkarren:
I would like to move forward on this as we are currently using this in [selene](https://github.com/Kampfkarren/selene), but gated by nightly-only CI.

The "Lint message needs to be improved" is doable. 

Unsure how the lint name could be improved. CC @camelid since you brought up the concern

Nemo157:
I sort of want a more expanded version of this lint that isn't limited to just patterns. I have some code parsing a non-exhaustive enum:

```rust
[...]

PR #90358, opened 2021-10-28
Fix exposing fields marked unstable or doc hidden

### LINKED ISSUE #89837: `non_exhaustive_omitted_patterns` lint exposes unstable and hidden fields
_Issue created from a thread starting at https://github.com/rust-lang/rust/issues/89042#issuecomment-934650963._

cc @Nadrieril @DevinR528

DevinR528:
Closes https://github.com/rust-lang/rust/issues/89837

Work towards https://github.com/rust-lang/rust/issues/89554

Filter fields that are marked `doc(hidden)` or are unstable with that feature turned off. This brings structs and enums into alignment behavior-wise when emitting warning/errors about pattern exhaustiveness/reachability.

cc @Nadrieril


compiler/rustc_typeck/src/check/pat.rs
```diff
@@ -1308,6 +1308,11 @@ impl<'a, 'tcx> FnCtxt<'a, 'tcx> {
                 .copied()
                 .filter(|(field, _)| {
                     field.vis.is_accessible_from(tcx.parent_module(pat.hir_id).to_def_id(), tcx)
+                        && !matches!(
+                            tcx.eval_stability(field.did, None, DUMMY_SP, None),
+                            EvalResult::Deny { .. }
+                        )
+                        && !tcx.is_doc_hidden(field.did)
```
camelid, line 1315: Hmm, I don't think we should be hiding `doc(hidden)` fields in the same crate as their definition.

src/test/ui/pattern/usefulness/doc-hidden-fields.stderr
```diff
@@ -0,0 +1,44 @@
+error: pattern requires `..` due to inaccessible fields
+  --> $DIR/doc-hidden-fields.rs:8:9
+   |
+LL |     let HiddenStruct { one, two, } = HiddenStruct::default();
+   |         ^^^^^^^^^^^^^^^^^^^^^^^^^^
+   |
+help: ignore the inaccessible and unused fields
+   |
+LL |     let HiddenStruct { one, two, .., } = HiddenStruct::default();
+   |                                ++++
```
camelid, line 10: This error is incorrect (even if it were outside the crate of definition). `doc(hidden)` doesn't cause the field to be inaccessible; it's only meant to discourage public use by hiding the field from the generated documentation.

I feel mixed about `non_exhaustive_omitted_patterns` also relying on `doc(hidden)`; but at least in that case, it just suppresses the lint in some cases, rather than explicitly saying that the field is inaccessible.

DevinR528: Do you mean the wording of messages with `doc(hidden)` fields, so it's  `pattern requires `..` due to `doc(hidden)` fields` or something like that?

I agree (I think you're kind of hinting at this) with you that making doc(hidden) behave in this way is odd/off. We have changed the meaning from (at least for error messages) this thing doesn't show its docs to, now, this is more like a `non_exhaustive` thing shrug.

Anyways I can conditionally alter the error messages for structs with doc hidden fields.

camelid: > I agree (I think you're kind of hinting at this) with you that making doc(hidden) behave in this way is odd/off. We have changed the meaning from (at least for error messages) this thing doesn't show its docs to, now, this is more like a `non_exhaustive` thing shrug.

Yeah, that's what I mean. I think the compiler should really just ignore `doc(hidden)`; it's meant for rustdoc.

> Do you mean the wording of messages with `doc(hidden)` fields, so it's `` pattern requires `..` due to `doc(hidden)` fields`` or something like that?

That'd still be incorrect since the pattern doesn't actually _require_ `..`. `doc(hidden)` is just a nudge that some part of a library is an implementation detail. I don't think rustc should be changing its error messages based on `doc(hidden)`; it's surprising and not what `doc(hidden)` means.

DevinR528: Would pinging this person https://github.com/rust-lang/rust/issues/89042 for some input be a good idea.

DevinR528: Friendly ping, is there any place I should take this issue for more feedback I tried [zulip #t-lang/major changes](https://rust-lang.zulipchat.com/#narrow/stream/243200-t-lang.2Fmajor-changes/topic/Non.20exhaustive.20reachable.20patterns.20lint.20lang-team.23112/near/262076253) to no avail.

camelid: You could try `#t-compiler` or `#t-lang` (rather than `#t-lang/major changes`). Sometimes though it's just hard to get feedback unfortunately.

> Would pinging this person #89042 for some input be a good idea.

You could try pinging m-ou-se, but she gets a lot of pings for library review, so not sure if she'd be able to respond.

You could also make this PR focus on just unstable fields (with the feature turned off) for now.

DevinR528: Thanks for the quick response! I'll try the first few ideas and if it's still isn't moving I'll just focus on unstable fields.

DevinR528:
Now that final exams are over I can finish this up!

@m-ou-se I collected some more feedback on converting `doc(hidden)` field/variants and it was mixed ([zulip](https://rust-lang.zulipchat.com/#narrow/stream/131828-t-compiler/topic/Should.20error.20msgs.20list.20doc.28hidden.29.20fields) which was started from the above review [here](https://github.com/rust-lang/rust/pull/90358#discussion_r747763197)). So my plan is to remove all the changes made in this PR about `doc(hidden)` and if needed open a new PR reverting back to `doc(hidden)`s previous/default behavior. (Sorry I didn't want to add to the number of pings I imagine you get but I wanted to let you know)

JohnCSimon:
Ping from triage: 
@DevinR528 what is the status of this PR? 

`@rustbot ready` to switch to `S-waiting-on-review`

thanks.

DevinR528:
@JohnCSimon I am waiting for a slightly more official go-ahead to remove the special treatment of `doc(hidden)` from @m-ou-se mostly. Or more discussion needs to be had?. There was a stream on zulip I started but I can't for the life of me find it.

Nadrieril:
Sounds like you're waiting for a review; let's update the tags
@rustbot ready

camelid:
> There was a stream on zulip I started but I can't for the life of me find it.

@DevinR528 is this what you were looking for? https://rust-lang.zulipchat.com/#narrow/stream/131828-t-compiler/topic/Should.20error.20msgs.20list.20doc.28hidden.29.20fields

DevinR528:
That's the one thanks @camelid !

m-ou-se:
I still think we should not expose doc(hidden) fields and variants through suggestions and diagnostics. This attribute is currently the only way users of stable Rust have to mark something as 'not part of the public API' for things that need to be `pub`, e.g. for macro reasons. In the standard library we have `#[unstable]`, but that attribute is unstable and won't be usable by other crates on stable any time soon. We shouldn't take away this tool from library authors.

apiraino:
I think it can now go back to:

@rustbot author

(but please update the labels if it's the case - thanks)

DevinR528:
Ok, I'm fine with doing whatever so then the only review was from @camelid about removing the special casing of `doc(hidden)` so I think this PR might be ready for another review if we are leaving `doc(hidden)` the way that it is since my change here https://github.com/rust-lang/rust/pull/89105/commits/2a042d61055263e404d9588d39046ed51e96a560.

I fixed filtering `doc(hidden)` when crate local changed/added tests for that case.

wesleywiser:
r? @camelid 

@camelid would you mind taking ownership of the review here? Otherwise, feel free to re-roll for a new reviewer or you can assign to me.

Thanks!

camelid:
Sorry, I don't have enough time to take ownership of reviewing. I can try to weigh in on discussions, but that's about it. r? @wesleywiser

src/test/ui/pattern/usefulness/doc-hidden-fields.stderr
```diff
@@ -0,0 +1,59 @@
+error: pattern requires `..` due to inaccessible fields
+  --> $DIR/doc-hidden-fields.rs:15:9
+   |
+LL |     let HiddenStruct { one, two, } = HiddenStruct::default();
+   |         ^^^^^^^^^^^^^^^^^^^^^^^^^^
+   |
+help: ignore the inaccessible and unused fields
+   |
+LL |     let HiddenStruct { one, two, .., } = HiddenStruct::default();
```
jackh726, line 9: Why is there a comma after this `..` but not after the one on line 20?

src/test/ui/pattern/usefulness/doc-hidden-fields.stderr
```diff
@@ -0,0 +1,59 @@
...
+
+error[E0027]: pattern does not mention field `two`
+  --> $DIR/doc-hidden-fields.rs:21:9
+   |
+LL |     let HiddenStruct { one, hide } = HiddenStruct::default();
+   |         ^^^^^^^^^^^^^^^^^^^^^^^^^^ missing field `two`
+   |
+help: include the missing field in the pattern
+   |
+LL |     let HiddenStruct { one, hide, two } = HiddenStruct::default();
```
jackh726, line 35: I think this can be left as followup, but ideally we should also suggest hiding in one go.

jackh726 (approved):
A nit, but r=me

jackh726:
Needs rebase

r? @jackh726

src/test/ui/pattern/usefulness/doc-hidden-fields.stderr
DevinR528: It seems that if the source text has the trailing comma i.e. `let Struct { a, b, } = thing;` then the last item, whatever it is (field name or rest operator), has the trailing comma. I fixed it by removing it from the example.

src/test/ui/pattern/usefulness/doc-hidden-fields.stderr
DevinR528: Even if the `doc(hidden)` field is being explicitly matched on?

jackh726: Hmm. Good point. Not sure.

PR #90361, opened 2021-10-28
Enable verification for 1/32th of queries loaded from disk

Mark-Simulacrum:
This is a limited enabling of incremental verification for query results loaded from disk, which previously did not run without -Zincremental-verify-ich. If enabled for all queries, we see a probably unacceptable hit of ~50% in the worst case, so this pairs back the verification to a more limited set based on the hash key.

Per collected [perf results](https://github.com/rust-lang/rust/pull/84227#issuecomment-953350582), this is a regression of at most 7% on coercions opt incr-unchanged, and typically less than 0.5% on other benchmarks (largely limited to incr-unchanged). I believe this is acceptable performance to land, and we can either ratchet it up or down fairly easily.

We have no real sense of whether this will lead to a large amount of assertions in the wild, but since those assertions may lead to miscompilations today, it seems potentially warranted. We have a good bit of lead time until the next stable release, though the holiday season will also start soon; we may wish to discuss the timing of enabling this and weigh the desire to prevent (possible) miscompilations against assertions.

cc @rust-lang/wg-incr-comp

### DIFF
--- a/compiler/rustc_query_system/src/query/plumbing.rs
+++ b/compiler/rustc_query_system/src/query/plumbing.rs
@@ -518,9 +518,22 @@ where
         prof_timer.finish_with_query_invocation_id(dep_node_index.into());
 
         if let Some(result) = result {
+            let prev_fingerprint = tcx
+                .dep_context()
+                .dep_graph()
+                .prev_fingerprint_of(dep_node)
+                .unwrap_or(Fingerprint::ZERO);
             // If `-Zincremental-verify-ich` is specified, re-hash results from
             // the cache and make sure that they have the expected fingerprint.
-            if unlikely!(tcx.dep_context().sess().opts.debugging_opts.incremental_verify_ich) {
+            //
+            // If not, we still seek to verify a subset of fingerprints loaded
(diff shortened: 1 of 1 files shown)


michaelwoerister (commented):
Thanks, @Mark-Simulacrum! I'm in favor of doing this but I wonder if we should discuss it with the wider compiler team.

michaelwoerister:
Since we didn't get to discuss this in the triage meeting two times in a row now, I'm planning to just r+ this (if the performance hit is indeed acceptable). We are still some time away from the next release and this can be quickly reverted if necessary. We'll just have to keep an eye out for bug reports.

Mark-Simulacrum:
Results do not significantly differ from the previous ones (see PR description for the link), and I think they're OK. We can reduce the impact with a smaller fraction being processed (for example, 1/64), but I think that there's a certain level of diminishing returns -- most affected benchmarks are incr-unchanged, and the few println or other patchsets are only slightly affected. I think the potential for catching bugs is higher the more we check, so I'd rather not reduce without strong motivation.

PR #90365, opened 2021-10-28
[EXPERIMENT] rustdoc: Cleanup `external_traits`

camelid:
r? @ghost
cc @jyn514

### DIFF
--- a/src/librustdoc/clean/inline.rs
+++ b/src/librustdoc/clean/inline.rs
@@ -674,16 +674,11 @@ crate fn record_extern_trait(cx: &mut DocContext<'_>, did: DefId) {
         return;
     }
 
-    {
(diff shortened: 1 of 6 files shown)


jyn514 (changes requested):

camelid:
I'm going to close this for now since it's not the right approach.

jyn514:
@camelid did you get a chance to try out https://github.com/rust-lang/rust/pull/90365#discussion_r738825148 ? I think with that change this would be the right approach :)

camelid:
Yeah, I did. The issue is that it causes multiple mutable borrows of `self` to be live simultaneously in the folder/visitor: one reference for the `DocContext` (because it's stored as a field of the folder) and one reference for the visiting of the external traits.

I don't think there's a good way to solve that problem without coming up with a different approach.

PR #90372, opened 2021-10-28
move out repeating macro normalization from cycle

klensy:


davidtwco (approved):
Seems like this should be a fairly straightforward performance improvement, perf queue isn't long at the moment so I'll queue a perf run before approving.

davidtwco:
Closing, feel free to re-open if you have other things to try.

PR #90339, opened 2021-10-27
Add missing documentation for doc alias

GuillaumeGomez:

PR #90374, opened 2021-10-28
Unify titles in rustdoc book doc attributes chapter

GuillaumeGomez:
As discussed in https://github.com/rust-lang/rust/pull/90339.

I wasn't able to find out where the link to the titles was used so let's see if the CI fails. :)

r? @camelid

### DIFF
--- a/compiler/rustc_passes/src/check_attr.rs
+++ b/compiler/rustc_passes/src/check_attr.rs
@@ -768,7 +768,7 @@ impl CheckAttrVisitor<'tcx> {
                             "not a `use` item",
                         );
                     }
-                    err.note("read https://doc.rust-lang.org/nightly/rustdoc/the-doc-attribute.html#docno_inlinedocinline for more information")
+                    err.note("read https://doc.rust-lang.org/nightly/rustdoc/the-doc-attribute.html#inline-and-no_inline for more information")
                         .emit();
                 },
             );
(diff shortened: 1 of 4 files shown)


camelid (commented):
r=me with the following comments addressed and the error message I mentioned above updated (you'll have to bless tests as well)

Please make sure to use `r=camelid` rather than `r+` :)

GuillaumeGomez:
Updated! I'll let you handle the r+ this time. ;)

compiler/rustc_passes/src/check_attr.rs
```diff
@@ -768,7 +768,7 @@ impl CheckAttrVisitor<'tcx> {
                             "not a `use` item",
                         );
                     }
-                    err.note("read https://doc.rust-lang.org/nightly/rustdoc/the-doc-attribute.html#docno_inlinedocinline for more information")
+                    err.note("read https://doc.rust-lang.org/nightly/rustdoc/the-doc-attribute.html#inline_and_no_inline for more information")
```
camelid, line 771: I think this link is not quite correct. It should be:

```suggestion
                    err.note("read https://doc.rust-lang.org/nightly/rustdoc/the-doc-attribute.html#inline-and-no_inline for more information")
```

GuillaumeGomez: Arf, copied the link I created and forgot to regen the book... Great catch!

camelid:
Thanks!

@bors r+ rollup

PR #90383, opened 2021-10-28
Extend check for UnsafeCell in consts to cover unions

tmiasko:
A validity companion to changes from #90373.

@rust-lang/wg-const-eval


RalfJung:
Sorry, this fell through the cracks because GH does not always send email notifications for force-pushes. This change makes sense to me.

@bors r+

tmiasko:
Rebased and updated the test output.

Issue #90187, opened 2021-10-22
rustdoc: Replace all remaining uses of `Type::def_id_no_primitives()` with `Type::def_id()`


camelid:
### Steps

1. Pass `&Cache` where it will be needed for `def_id()`.
1. Replace all uses of `def_id_no_primitives()` with `def_id()`.
2. Remove `def_id_no_primitives()` and simplify `Type::inner_def_id()`.

PR #90385, opened 2021-10-28
rustdoc: use Type::def_id() instead of Type::def_id_no_primitives()

mfrw:
For: #90187

r? @jyn514

### DIFF
--- a/src/librustdoc/formats/cache.rs
+++ b/src/librustdoc/formats/cache.rs
@@ -303,7 +303,7 @@ impl<'a, 'tcx> DocFolder for CacheBuilder<'a, 'tcx> {
                             desc,
                             parent,
                             parent_idx: None,
-                            search_type: get_index_search_type(&item, self.tcx),
+                            search_type: get_index_search_type(&item, self.tcx, self.cache),
                             aliases: item.attrs.get_doc_aliases(),
                         });
                     }
--- a/src/librustdoc/html/render/cache.rs
+++ b/src/librustdoc/html/render/cache.rs
@@ -257,7 +257,7 @@ crate fn get_real_types<'tcx>(
         tcx: TyCtxt<'_>,
         ty: Type,
         mut generics: Vec<TypeWithKind>,
-        cache: &Cache,
+        _cache: &Cache,
     ) {
         let is_full_generic = ty.is_full_generic();
 
@@ -309,7 +309,7 @@ crate fn get_real_types<'tcx>(
             // We remove the name of the full generic because we have no use for it.
             index_ty.name = Some(String::new());
             res.push(TypeWithKind::from((index_ty, ItemType::Generic)));
-        } else if let Some(kind) = ty.def_id(cache).map(|did| tcx.def_kind(did).into()) {
+        } else if let Some(kind) = ty.def_id_no_primitives().map(|did| tcx.def_kind(did).into()) {
             res.push(TypeWithKind::from((index_ty, kind)));
         } else if ty.is_primitive() {
             // This is a primitive, let's store it as such.
@@ -324,7 +324,9 @@ crate fn get_real_types<'tcx>(
 
     if let Type::Generic(arg_s) = *arg {
         if let Some(where_pred) = generics.where_predicates.iter().find(|g| match g {
-            WherePredicate::BoundPredicate { ty, .. } => ty.def_id(cache) == arg.def_id(cache),
+            WherePredicate::BoundPredicate { ty, .. } => {
+                ty.def_id_no_primitives() == arg.def_id_no_primitives()
+            }
             _ => false,
         }) {
             let mut ty_generics = Vec::new();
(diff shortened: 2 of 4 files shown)


camelid:
Hmm, somehow this makes generics show up in search results (that's why the test is failing AFAICT). I don't think that's supposed to happen; otherwise searching for `T` would return a *lot* of results. cc @GuillaumeGomez

GuillaumeGomez:
Indeed, for the search index, we don't want to keep generics as is. It can be fixed by turning the `cache` argument into an `Option` and passing `None` when using it for the search index.

camelid:
> Indeed, for the search index, we don't want to keep generics as is. It can be fixed by turning the `cache` argument into an `Option` and passing `None` when using it for the search index.

But why does switching to `def_id` from `def_id_no_primitives` affect whether generics show up in the search results? They're not primitives...

GuillaumeGomez:
I'll test locally tomorrow to see what's going on because from what I can see, the code is perfectly valid. You have allowed to uncover a bug in rustdoc. (yeay! :partying_face: :tada: )

mfrw:
> I'll test locally tomorrow to see what's going on because from what I can see, the code is perfectly valid. You have allowed to uncover a bug in rustdoc. (yeay! 🥳 🎉 )

Yayay .. I am noob and I only could stare at the code and now was reading more about rust if i was missing something trivial. 
ALso while we are here, what is the guidance on removing this when we are implementing a trait like [this](https://github.com/rust-lang/rust/blob/master/src/librustdoc/passes/stripper.rs#L130). It seems it does not have access to the `Cache` directly.

camelid:
>  ALso while we are here, what is the guidance on removing this when we are implementing a trait like [this](https://github.com/rust-lang/rust/blob/master/src/librustdoc/passes/stripper.rs#L130). It seems it does not have access to the `Cache` directly.

You should be able to add a `cache: &'a Cache` field to the struct that the trait is being implemented on. I'm in the process of adding Cache fields to all those places already, so you could also just skip that usage for now.

GuillaumeGomez:
Sorry, I was focused on other things and completely forgot about this. Scheduling it for tomorrow.

mfrw:
> You should be able to add a `cache: &'a Cache` field to the struct that the trait is being implemented on. I'm in the process of adding Cache fields to all those places already, so you could also just skip that usage for now.

Since you adding Cache fields, should I wait and then update the PR. @camelid

camelid:
> > You should be able to add a `cache: &'a Cache` field to the struct that the trait is being implemented on. I'm in the process of adding Cache fields to all those places already, so you could also just skip that usage for now.
> 
> Since you adding Cache fields, should I wait and then update the PR. @camelid

That work is blocked for now, so feel free to add the fields yourself if you can.

GuillaumeGomez:
I found where the issue was coming from: https://github.com/rust-lang/rust/pull/90726. I'm just very surprised we didn't find it out sooner. Well, I'm quite happy you uncovered it because thanks to you, it allowed to fix this issue and helped me uncover another one: https://github.com/rust-lang/rust/pull/90727. So thanks! :)

mfrw:
> I found where the issue was coming from: #90726. I'm just very surprised we didn't find it out sooner. Well, I'm quite happy you uncovered it because thanks to you, it allowed to fix this issue and helped me uncover another one: #90727. So thanks! :)

Happy to be of any help, even though I am a complete noob. Thank you all for being welcoming!

GuillaumeGomez:
We all started in your position, so we know it's nice to have help. So anyway, once #90726 is merged, just rebase on it and that should work just fine.

GuillaumeGomez:
@mfrw: The fix in #90726 is invalid. So to fix the issue, you'll need to check that the returned type isn't a `Primitive::Reference` in `render/cache.rs`.

Unless you have a better idea @jyn514 @camelid ?

camelid:
> @mfrw: The fix in #90726 is invalid. So to fix the issue, you'll need to check that the returned type isn't a `Primitive::Reference` in `render/cache.rs`.
> 
> Unless you have a better idea @jyn514 @camelid ?

Another option is to not change the problem `def_id_no_primitive` call for now, to avoid blocking the rest of this PR. Then a followup PR could focus on just that use.

GuillaumeGomez:
@camelid This is a good suggestion! If so, please someone open an issue as always so we don't forget. :)

mfrw:
> > @mfrw: The fix in #90726 is invalid. So to fix the issue, you'll need to check that the returned type isn't a `Primitive::Reference` in `render/cache.rs`.
> > Unless you have a better idea @jyn514 @camelid ?
> 
> Another option is to not change the problem `def_id_no_primitive` call for now, to avoid blocking the rest of this PR. Then a followup PR could focus on just that use.

I followed this approach. Should I create an issue for the cleanup of the remaining `def_id_no_primitive` in `render/cache.rs` ?

jyn514:
I don't think this needs a new issue, just leave https://github.com/rust-lang/rust/issues/90187 open :)

GuillaumeGomez (approved):

GuillaumeGomez:
Thanks!

@bors: r+

GuillaumeGomez:
Spurious error.

@bors: retry

GuillaumeGomez:
I think CI is broken. cc @rust-lang/infra

PR #90388, opened 2021-10-28
Add unstable feature 'split_rinclusive', adding a right-inclusive version of `str::split_inclusive`

logannc:
Apologies, no issue associated with this one. If that is necessary for a change this small, please let me know and I'll create an issue or pFCP as needed.

### Motivation

`str::split_inclusive` allows for easily splitting a string and including the `separator` in the substring on the "left".

This change proposes `str::split_rinclusive` which allows to easily split a string but include the `separator` on the "right".

This does not change result order like `rsplit`, so it is not, for example,  a hypothetical `rsplit_inclusive`. Actually, they might be somewhat inverses of each other, but telling people to use `rsplit_inclusive(...).rev()` is less ergonomic and harder to reason about. Either way, neither currently exists.

### Implementation

For the most part, just copying the existing `SplitInternal` impl and changing the matcher indices used. 

We also introduce another param: `allow_leading_empty`, an opposite to the existing `allow_trailing_empty`. This is to mirror the functionality of `split_inclusive` when the final character matches (we do not return a trailing empty substring). Likewise, for this implementation, we do not return a leading empty string if the first character matches.

### DIFF
--- a/library/alloc/tests/lib.rs
+++ b/library/alloc/tests/lib.rs
@@ -25,6 +25,7 @@
 #![feature(const_btree_new)]
 #![feature(const_default_impls)]
 #![feature(const_trait_impl)]
+#![feature(split_rinclusive)]
 
 use std::collections::hash_map::DefaultHasher;
 use std::hash::{Hash, Hasher};
--- a/library/alloc/tests/str.rs
+++ b/library/alloc/tests/str.rs
@@ -1334,13 +1334,8 @@ fn test_split_char_iterator_rinclusive() {
     assert_eq!(split, ["\nMäry häd ä little lämb", "\nLittle lämb", "\n"]);
 
     let uppercase_separated = "SheepSharkTurtleCat";
-    let mut first_char = true;
     let split: Vec<&str> = uppercase_separated
-        .split_rinclusive(|c: char| {
-            let split = !first_char && c.is_uppercase();
-            first_char = split;
-            split
-        })
+        .split_rinclusive(char::is_uppercase)
         .collect();
     assert_eq!(split, ["Sheep", "Shark", "Turtle", "Cat"]);
 }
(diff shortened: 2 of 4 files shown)


joshtriplett:
Can you give some examples of use cases for this method? (I can imagine some myself, but I'd like to know what you had in mind.)

logannc:
To be perfectly honest, I'm having trouble thinking of examples besides something like CamelCaseSplitting. I'm sure there are some formats which use sentinel values to indicate the start of a new section, but I'm not widely versed enough to name any.

An argument could be made for symmetry between `split_inclusive` and `split_rinclusive`, but more examples would be better.

A sillier but still valid argument would be that I think there should be an easy way to split on caps without using Regex! Even if we include the iterator methods, its hard to implement. We can do it with `fold` or `scan`, but its more effort than I'd expect. 

Of course, we could also just implement such a thing imperatively, so it's not like we're stuck without it. :)

Alternatively, I could see omitting this in favor of a dedicated function for it - it isn't an uncommon operation.

But maybe this is just my own itch.

joshtriplett:
CamelCase splitting and similar "records start with something that looks like this" formats seem like a reasonable use case. My main concern is the number of potential variations here: `split_rinclusive`, `split_inclusive`, `rsplit_inclusive`, `rsplit_rinclusive`...

JohnCSimon:
Ping from triage: 
@logannc Can you please address the build failures?

JohnCSimon:
ping from triage:
@logannc 
Returning to you to address comments from reviewer


FYI: when a PR is ready for review, post a message containing 
`@rustbot ready` to switch the PR to `S-waiting-on-review` so the PR appears in the reviewer's backlog.
Thanks

JohnCSimon:
@logannc 
Thanks for your contribution.

@rustbot label: +S-inactive

PR #90391, opened 2021-10-29
rustdoc: Compute some fields of `clean::Crate` on-demand to reduce size

camelid:
`clean::Crate` is frequently moved by-value -- for example, in `DocFolder`
implementations -- so reducing its size should improve performance.

This PR reduces the size of `clean::Crate` from 168 bytes to 104 bytes.

r? @jyn514

### DIFF
--- a/src/librustdoc/clean/types.rs
+++ b/src/librustdoc/clean/types.rs
@@ -117,7 +117,6 @@ impl From<DefId> for ItemId {
 #[derive(Clone, Debug)]
 crate struct Crate {
     crate name: Symbol,
-    crate src: FileName,
     crate module: Item,
     crate externs: Vec<ExternalCrate>,
     crate primitives: ThinVec<(DefId, PrimitiveType)>,
@@ -128,7 +127,13 @@ crate struct Crate {
 
 // `Crate` is frequently moved by-value. Make sure it doesn't unintentionally get bigger.
 #[cfg(all(target_arch = "x86_64", target_pointer_width = "64"))]
-rustc_data_structures::static_assert_size!(Crate, 168);
+rustc_data_structures::static_assert_size!(Crate, 104);
+
+impl Crate {
+    crate fn src(&self, tcx: TyCtxt<'_>) -> FileName {
+        ExternalCrate::LOCAL.src(tcx)
+    }
+}
 
 /// This struct is used to wrap additional information added by rustdoc on a `trait` item.
 #[derive(Clone, Debug)]
--- a/src/librustdoc/clean/types.rs
+++ b/src/librustdoc/clean/types.rs
@@ -116,7 +116,6 @@ impl From<DefId> for ItemId {
 /// The crate currently being documented.
 #[derive(Clone, Debug)]
 crate struct Crate {
-    crate name: Symbol,
     crate module: Item,
     crate externs: Vec<ExternalCrate>,
     crate primitives: ThinVec<(DefId, PrimitiveType)>,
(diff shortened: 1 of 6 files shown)


jyn514 (commented):
This is great, thanks! Just a couple nits.

jyn514:
... apparently that made the RSS regression go away? how strange ...

camelid:
> ... apparently that made the RSS regression go away? how strange ...

In my experience, `deep-vector` often has spurious results, so it's not super surprising. For example, on the most recent perf run above, `deep-vector opt` had a max-rss regression of 4.85%, which is certainly spurious since this PR did not touch rustc.

camelid:
It *is* quite strange how this change doesn't seem to have had much of a performance impact. I think I'll try removing the `num` and `externs` field and see if that improves performance. I might also try re-adding `name` since it's used in several places.

jyn514:
@camelid I would rather not add back `name()`, it's cheap to compute and rustdoc should be doing things on-demand.

camelid:
> @camelid I would rather not add back `name`, it's cheap to compute and rustdoc should be doing things on-demand.

Ok, sounds good. I'll try removing some of the other fields too.

src/librustdoc/clean/types.rs
```diff
@@ -115,8 +115,7 @@ impl From<DefId> for ItemId {
 
 #[derive(Clone, Debug)]
 crate struct Crate {
-    crate name: Symbol,
-    crate src: FileName,
+    crate num: CrateNum,
     crate module: Item,
     crate externs: Vec<ExternalCrate>,
```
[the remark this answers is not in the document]
camelid: Update: In order to remove this field, `Cache::populate` will need to take a DocContext, which I think is possible but will require some changes. I'm also working on that as part of #90365, so I'll save removing this field for a future PR.

camelid:
(Currently squashing.)

camelid:
Ok, should be good for a new review!

src/librustdoc/clean/types.rs
```diff
@@ -138,6 +151,8 @@ crate struct ExternalCrate {
 }
 
 impl ExternalCrate {
+    crate const LOCAL: Self = Self { crate_num: LOCAL_CRATE };
```
jyn514, line 154: this can be private I think, it's only used from this file
```suggestion
    const LOCAL: Self = Self { crate_num: LOCAL_CRATE };
```

src/librustdoc/clean/types.rs
```diff
@@ -125,6 +124,20 @@ crate struct Crate {
...
+// `Crate` is frequently moved by-value. Make sure it doesn't unintentionally get bigger.
+#[cfg(all(target_arch = "x86_64", target_pointer_width = "64"))]
+rustc_data_structures::static_assert_size!(Crate, 104);
+
+impl Crate {
+    crate fn name(&self, tcx: TyCtxt<'_>) -> Symbol {
+        ExternalCrate::LOCAL.name(tcx)
+    }
+
+    crate fn src_file_name(&self, tcx: TyCtxt<'_>) -> FileName {
```
jyn514, line 136: nit: I would rather name this `src` to be consistent with `ExternalCrate`

jyn514 (changes requested):
r=me with those nits fixed :)

src/librustdoc/clean/types.rs
camelid: Yeah, consistency is probably best here. `src` is just a bit of a confusing name ;)

I renamed it back.

src/librustdoc/clean/types.rs
```diff
@@ -125,6 +124,20 @@ crate struct Crate {
     crate collapsed: bool,
 }
 
+// `Crate` is frequently moved by-value. Make sure it doesn't unintentionally get bigger.
+#[cfg(all(target_arch = "x86_64", target_pointer_width = "64"))]
+rustc_data_structures::static_assert_size!(Crate, 104);
```
GuillaumeGomez, line 129: Nice idea to add this check! 👍

camelid: Thanks :)

PR #90406, opened 2021-10-29
Collect `panic/panic_bounds_check` during monomorphization

### LINKED ISSUE #10019: nightly-2021-10-28 and later broke the `panic_immediate_abort` feature of -Zbuild-std
*EDIT*: Given that the cargo commit hashes are the same in `nightly-2021-10-28` and `nightly-2021-10-27`, it might be rustc itself which is broken, somehow between hashes `e269e6bf47f40c9046cd44ab787881d700099252` and `4e0d3973fafdfb1c51011bc74e44257b5e3863f1` e.g.  [in this set of commits](https://github.com/rust-lang/rust/compare/e269e6bf47f40c9046cd44ab787881d700099252...4e0d3973fafdfb1c51011bc74e44257b5e3863f1).

*EDIT 2*: Might be coming from this commit: https://github.com/rust-lang/rust/commit/cc4345a1c5fce8dec8a44b7fd80be8867ac6e649

*EDIT 3*: `-Zbuild-std` is broken *only* with `build-std-features = ["panic_immediate_abort"]`. If I comment out that line in `.cargo/config.toml`, the linkage succeeds. Which increases the suspicion that the commit from *EDIT 2* is the offending one, especially the changes to `panicking.rs`.

### Problem

Since `nightly-2021-10-28`, `-Zbuild-std` reports linkage errors where `core::panicking::panic::*` cannot be found.
The latest nightly that works is `nightly-2021-10-27` 

### Steps

1. `rustup install nightly` (or `rustup install nightly-2021-10-28`)
2. `rustup default nightly` (or `rustup default nightly-2021-10-28`)
3. Download and unpack the attached demo crate
4. `cargo build --target x86_64-apple-darwin` (target does not matter; it is broken everywhere)
5. Expected result: successful linkage
6. Observed result:
```
  = note: Undefined symbols for architecture x86_64:
            "core::panicking::panic::h019220a5efa21cf3", referenced from:
                panic_unwind::real_imp::find_eh_action::h77f57cf800eb60b2 in libpanic_unwind-2e68708bd30aab08.rlib(panic_unwind-2e68708bd30aab08.panic_unwind.cae1f51d-cgu.0.rcgu.o)
                panic_unwind::dwarf::eh::find_eh_action::h5331216282f6ca73 in libpanic_unwind-2e68708bd30aab08.rlib(panic_unwind-2e68708bd30aab08.panic_unwind.cae1f51d-cgu.7.rcgu.o)
                panic_unwind::dwarf::eh::round_up::h50516c03cc2ea37e in libpanic_unwind-2e68708bd30aab08.rlib(panic_unwind-2e68708bd30aab08.panic_unwind.cae1f51d-cgu.7.rcgu.o)
                panic_unwind::dwarf::eh::read_encoded_pointer::ha8d12fdd91646c7b in libpanic_unwind-2e68708bd30aab08.rlib(panic_unwind-2e68708bd30aab08.panic_unwind.cae1f51d-cgu.7.rcgu.o)
                panic_unwind::dwarf::DwarfReader::read_uleb128::hf52d44d8f44bcf32 in libpanic_unwind-2e68708bd30aab08.rlib(panic_unwind-2e68708bd30aab08.panic_unwind.cae1f51d-cgu.5.rcgu.o)
                panic_unwind::dwarf::DwarfReader::read_sleb128::he29ae0fcb92e7992 in libpanic_unwind-2e68708bd30aab08.rlib(panic_unwind-2e68708bd30aab08.panic_unwind.cae1f51d-cgu.5.rcgu.o)
          ld: symbol(s) not found for architecture x86_64
          clang: error: linker command failed with exit code 1 (use -v to see invocation)
```

### Possible Solution(s)

No solution, just some initial research, as per the *EDIT* notes above:
* Given that the cargo commit hashes are the same in `nightly-2021-10-28` and `nightly-2021-10-27`, it might be rustc itself which is broken, somehow between hashes `e269e6bf47f40c9046cd44ab787881d700099252` and `4e0d3973fafdfb1c51011bc74e44257b5e3863f1` e.g.  [in this set of commits](https://github.com/rust-lang/rust/compare/e269e6bf47f40c9046cd44ab787881d700099252...4e0d3973fafdfb1c51011bc74e44257b5e3863f1).
* Might be coming from this commit: https://github.com/rust-lang/rust/commit/cc4345a1c5fce8dec8a44b7fd80be8867ac6e649

### Notes

_No response_

### Version

```text
cargo 1.58.0-nightly (6c1bc24b8 2021-10-24)
release: 1.58.0
commit-hash: 6c1bc24b8b49d4bc965f67d7037906dc199c72b7
commit-date: 2021-10-24
host: x86_64-apple-darwin
libgit2: 1.3.0 (sys:0.13.23 vendored)
libcurl: 7.64.1 (sys:0.4.49+curl-7.79.1 system ssl:(SecureTransport) LibreSSL/2.8.3)
os: Mac OS 11.5.2 [64-bit]
```

### LINKED ISSUE #90405: Implicit panic calls not collected during monomorphization
Implicit panic calls can be generated from `Assert` MIR terminator, but they are not collected during monomorphization. This means that link error will occur if the panic functions are not codegenned with `core` crate, e.g. when it is marked `inline`.

This causes breakage in rust-lang/cargo#10019 when #90273 changes the function to be inline if `panic_immediate_abort` is set and exposes this issue.

nbdd0121:
This would prevent link time errors if these functions are `#[inline]` (e.g. when `panic_immediate_abort` is used).

Fix #90405
Fix rust-lang/cargo#10019

@rustbot label: T-compiler A-codegen

### DIFF
--- a/src/test/codegen-units/item-collection/implicit-panic-call.rs
+++ b/src/test/codegen-units/item-collection/implicit-panic-call.rs
@@ -1,0 +1,58 @@
+// compile-flags:-Zprint-mono-items=lazy
+
+// rust-lang/rust#90405
+// Ensure implicit panic calls are collected
+
+#![feature(lang_items)]
+#![feature(no_core)]
+#![crate_type = "lib"]
+#![no_core]
+#![no_std]
+
+#[lang = "panic_location"]
+struct Location<'a> {
+    _file: &'a str,
+    _line: u32,
+    _col: u32,
+}
+
+#[lang = "panic"]
+#[inline]
+#[track_caller]
+fn panic(_: &'static str) -> ! {
+    loop {}
+}
+
+#[lang = "sized"]
+trait Sized {}
+
+#[lang = "copy"]
+trait Copy {}
+
+#[lang = "freeze"]
+trait Freeze {}
+
+impl Copy for i32 {}
+
+#[lang = "div"]
+trait Div<Rhs = Self> {
+    type Output;
+    fn div(self, rhs: Rhs) -> Self::Output;
+}
+
+impl Div for i32 {
+    type Output = i32;
+    fn div(self, rhs: i32) -> i32 {
+        self / rhs
+    }
+}
+
+#[allow(unconditional_panic)]
+pub fn foo() {
+    // This implicitly generates a panic call.
+    let _ = 1 / 0;
+}
+
+//~ MONO_ITEM fn foo
+//~ MONO_ITEM fn <i32 as Div>::div
+//~ MONO_ITEM fn panic
--- a/compiler/rustc_hir/src/lang_items.rs
+++ b/compiler/rustc_hir/src/lang_items.rs
@@ -281,12 +281,12 @@ language_item_table! {
     // in the sense that a crate is not required to have it defined to use it, but a final product
     // is required to define it somewhere. Additionally, there are restrictions on crates that use
     // a weak lang item, but do not have it defined.
-    Panic,                   sym::panic,               panic_fn,                   Target::Fn,             GenericRequirement::None;
+    Panic,                   sym::panic,               panic_fn,                   Target::Fn,             GenericRequirement::Exact(0);
     PanicFmt,                sym::panic_fmt,           panic_fmt,                  Target::Fn,             GenericRequirement::None;
     PanicDisplay,            sym::panic_display,       panic_display,              Target::Fn,             GenericRequirement::None;
     PanicStr,                sym::panic_str,           panic_str,                  Target::Fn,             GenericRequirement::None;
     ConstPanicFmt,           sym::const_panic_fmt,     const_panic_fmt,            Target::Fn,             GenericRequirement::None;
-    PanicBoundsCheck,        sym::panic_bounds_check,  panic_bounds_check_fn,      Target::Fn,             GenericRequirement::None;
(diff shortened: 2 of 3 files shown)


src/test/codegen-units/item-collection/implicit-panic-call.rs
```diff
@@ -0,0 +1,58 @@
+// compile-flags:-Zprint-mono-items=eager
```
tmiasko, line 1: Wouldn't the eager mode consider `panic` function to be one of roots and hide the bug?

nbdd0121: Fixed. Good catch!

cjgillot:
Thanks @nbdd0121! LGTM.

`Instance::mono` will panic if the function has generics.
What happens if the input program (erroneously) declares the `panic` lang item function to have generic parameters?
```
#[lang = "panic"]
fn panic<T>(_: T) -> ! { loop {} }
```

Is this error correctly caught by lang item check or does `Instance::mono` ICE.

nbdd0121:
> Is this error correctly caught by lang item check or does `Instance::mono` ICE.

It'll ICE. But it will ICE regardless this change since `rustc_codegen_ssa` uses `Instance::mono` to create an instance of `panic` as well.

nbdd0121:
(with that said this is a very simple check so added it anyway)

PR #90408, opened 2021-10-29
Remove `PartialOrd`, `Ord` from `LocalDefId`

pierwill:
Part of work on https://github.com/rust-lang/rust/issues/90317.

### DIFF
--- a/compiler/rustc_data_structures/src/graph/vec_graph/mod.rs
+++ b/compiler/rustc_data_structures/src/graph/vec_graph/mod.rs
@@ -1,3 +1,5 @@
+use std::cmp::Ord;
+
 use crate::graph::{DirectedGraph, GraphSuccessors, WithNumEdges, WithNumNodes, WithSuccessors};
 use rustc_index::vec::{Idx, IndexVec};
 
@@ -17,10 +19,10 @@ pub struct VecGraph<N: Idx> {
     edge_targets: Vec<N>,
 }
 
-impl<N: Idx> VecGraph<N> {
+impl<N: Idx + Ord> VecGraph<N> {
     pub fn new(num_nodes: usize, mut edge_pairs: Vec<(N, N)>) -> Self {
         // Sort the edges by the source -- this is important.
-        edge_pairs.sort_by_key(|&edge_pairs| (edge_pairs.0.index(), edge_pairs.1.index()));
+        edge_pairs.sort();
 
         let num_edges = edge_pairs.len();
 
@@ -100,7 +102,7 @@ impl<'graph, N: Idx> GraphSuccessors<'graph> for VecGraph<N> {
     type Iter = std::iter::Cloned<std::slice::Iter<'graph, N>>;
 }
 
-impl<N: Idx> WithSuccessors for VecGraph<N> {
+impl<N: Idx + Ord> WithSuccessors for VecGraph<N> {
     fn successors(&self, node: N) -> <Self as GraphSuccessors<'_>>::Iter {
         self.successors(node).iter().cloned()
     }
--- a/compiler/rustc_data_structures/src/graph/vec_graph/mod.rs
+++ b/compiler/rustc_data_structures/src/graph/vec_graph/mod.rs
@@ -20,7 +20,7 @@ pub struct VecGraph<N: Idx> {
 impl<N: Idx> VecGraph<N> {
     pub fn new(num_nodes: usize, mut edge_pairs: Vec<(N, N)>) -> Self {
         // Sort the edges by the source -- this is important.
-        edge_pairs.sort();
+        edge_pairs.sort_by_key(|&edge_pairs| (edge_pairs.0.index(), edge_pairs.1.index()));
 
         let num_edges = edge_pairs.len();
 
--- a/compiler/rustc_hir/src/hir_id.rs
+++ b/compiler/rustc_hir/src/hir_id.rs
@@ -11,7 +11,7 @@ use std::fmt;
 /// the `local_id` part of the `HirId` changing, which is a very useful property in
 /// incremental compilation where we have to persist things through changes to
 /// the code base.
-#[derive(Copy, Clone, PartialEq, Eq, Hash, Debug, PartialOrd, Ord)]
+#[derive(Copy, Clone, PartialEq, Eq, Hash, Debug)]
 #[derive(Encodable, Decodable)]
 pub struct HirId {
     pub owner: LocalDefId,
@@ -32,11 +32,27 @@ impl HirId {
     pub fn make_owner(owner: LocalDefId) -> Self {
         Self { owner, local_id: ItemLocalId::from_u32(0) }
     }
+
+    pub fn index(self) -> (usize, usize) {
+        (rustc_index::vec::Idx::index(self.owner), rustc_index::vec::Idx::index(self.local_id))
+    }
 }
 
 impl fmt::Display for HirId {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
         write!(f, "{:?}", self)
+    }
+}
+
+impl Ord for HirId {
+    fn cmp(&self, other: &Self) -> std::cmp::Ordering {
+        (self.index()).cmp(&(other.index()))
+    }
+}
+
+impl PartialOrd for HirId {
+    fn partial_cmp(&self, other: &Self) -> Option<std::cmp::Ordering> {
+        Some(self.cmp(&other))
     }
 }
 
--- a/compiler/rustc_index/src/bit_set.rs
+++ b/compiler/rustc_index/src/bit_set.rs
@@ -715,6 +715,10 @@ impl<T: Idx> SparseBitSet<T> {
         self.elems.iter()
     }
 
+    bit_relations_inherent_impls! {}
+}
+
+impl<T: Idx + Ord> SparseBitSet<T> {
     fn last_set_in(&self, range: impl RangeBounds<T>) -> Option<T> {
         let mut last_leq = None;
         for e in self.iter() {
@@ -724,8 +728,6 @@ impl<T: Idx> SparseBitSet<T> {
         }
         last_leq
     }
-
-    bit_relations_inherent_impls! {}
 }
 
 /// A fixed-size bitset type with a hybrid representation: sparse when there
@@ -802,7 +804,10 @@ impl<T: Idx> HybridBitSet<T> {
     /// Returns the previous element present in the bitset from `elem`,
     /// inclusively of elem. That is, will return `Some(elem)` if elem is in the
     /// bitset.
-    pub fn last_set_in(&self, range: impl RangeBounds<T>) -> Option<T> {
+    pub fn last_set_in(&self, range: impl RangeBounds<T>) -> Option<T>
+    where
+        T: Ord,
+    {
         match self {
             HybridBitSet::Sparse(sparse) => sparse.last_set_in(range),
             HybridBitSet::Dense(dense) => dense.last_set_in(range),
--- a/compiler/rustc_index/src/bit_set.rs
+++ b/compiler/rustc_index/src/bit_set.rs
@@ -675,7 +675,7 @@ impl<T: Idx> SparseBitSet<T> {
 
     fn insert(&mut self, elem: T) -> bool {
         assert!(elem.index() < self.domain_size);
-        let changed = if let Some(i) = self.elems.iter().position(|&e| e >= elem) {
+        let changed = if let Some(i) = self.elems.iter().position(|&e| e.index() >= elem.index()) {
             if self.elems[i] == elem {
                 // `elem` is already in the set.
                 false
--- a/compiler/rustc_index/src/vec.rs
+++ b/compiler/rustc_index/src/vec.rs
@@ -12,7 +12,7 @@ use std::vec;
 /// Represents some newtyped `usize` wrapper.
 ///
 /// Purpose: avoid mixing indexes for different bitvector domains.
-pub trait Idx: Copy + 'static + Ord + Debug + Hash {
+pub trait Idx: Copy + 'static + Eq + PartialEq + Debug + Hash {
     fn new(idx: usize) -> Self;
 
     fn index(self) -> usize;
--- a/compiler/rustc_metadata/src/rmeta/encoder.rs
+++ b/compiler/rustc_metadata/src/rmeta/encoder.rs
@@ -1306,7 +1306,7 @@ impl<'a, 'tcx> EncodeContext<'a, 'tcx> {
             })
             .collect::<Vec<_>>();
         // Sort everything to ensure a stable order for diagnotics.
-        keys_and_jobs.sort_by_key(|&(def_id, _, _)| def_id);
+        keys_and_jobs.sort_by_key(|&(def_id, _, _)| def_id.index());
         for (def_id, encode_const, encode_opt) in keys_and_jobs.into_iter() {
             debug_assert!(encode_const || encode_opt);
 
--- a/compiler/rustc_middle/src/mir/mono.rs
+++ b/compiler/rustc_middle/src/mir/mono.rs
@@ -6,7 +6,7 @@ use rustc_data_structures::fingerprint::Fingerprint;
 use rustc_data_structures::fx::FxHashMap;
 use rustc_data_structures::stable_hasher::{HashStable, StableHasher};
 use rustc_hir::def_id::{CrateNum, DefId, LOCAL_CRATE};
-use rustc_hir::{HirId, ItemId};
+use rustc_hir::ItemId;
 use rustc_query_system::ich::{NodeIdHashingMode, StableHashingContext};
 use rustc_session::config::OptLevel;
 use rustc_span::source_map::Span;
@@ -355,7 +355,7 @@ impl<'tcx> CodegenUnit<'tcx> {
         // The codegen tests rely on items being process in the same order as
         // they appear in the file, so for local items, we sort by node_id first
         #[derive(PartialEq, Eq, PartialOrd, Ord)]
-        pub struct ItemSortKey<'tcx>(Option<HirId>, SymbolName<'tcx>);
+        pub struct ItemSortKey<'tcx>(Option<usize>, SymbolName<'tcx>);
 
         fn item_sort_key<'tcx>(tcx: TyCtxt<'tcx>, item: MonoItem<'tcx>) -> ItemSortKey<'tcx> {
             ItemSortKey(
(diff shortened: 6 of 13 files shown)


cjgillot (commented):
Thanks for the PR @pierwill! I left a few comments.
Those trait bounds on `LocalDefId` are used a lot, it may be easier to handle `Idx` and `PartialOrd/Ord` separately.
Changing data structures means we will have to be careful about perf implications.

pierwill:
> Those trait bounds on LocalDefId are used a lot, it may be easier to handle Idx and PartialOrd/Ord separately.

Thanks for reviewing, @cjgillot! I'll work on just the ordering traits for now. I also might see if `DefId` is easier to start with (https://github.com/rust-lang/rust/issues/90317#issuecomment-955132184).

compiler/rustc_span/src/span_encoding.rs
```diff
@@ -60,7 +60,7 @@ use rustc_data_structures::fx::FxIndexSet;
 /// the dependency to the parent definition's span. This is performed
 /// using the callback `SPAN_TRACK` to access the query engine.
 ///
-#[derive(Clone, Copy, Eq, PartialEq, Hash)]
+#[derive(Clone, Copy, Eq, PartialEq, Hash, PartialOrd, Ord)]
```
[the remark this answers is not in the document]
pierwill: I had a feeling this change was more than we wanted. 👍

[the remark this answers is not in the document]
pierwill: > The former PartialOrd implementation using SpanData attempted to sort spans by position.

This is in historical code?

> Instead, we should keep an implementation in terms of SpanData, but stop comparing SpanData::parent

So will `Span` end up using manual implementations for `Ord` and `PartialOrd`?

pierwill:
> Those trait bounds on LocalDefId are used a lot, it may be easier to handle Idx and PartialOrd/Ord separately.

Looks like `Idx` requires `Ord`...

compiler/rustc_span/src/lib.rs
```diff
@@ -440,6 +440,18 @@ pub struct SpanData {
...
 
+impl Ord for SpanData {
+    fn cmp(&self, other: &Self) -> Ordering {
+        (self.lo, self.hi, self.ctxt).cmp(&(other.lo, other.hi, other.ctxt))
+    }
+}
+
+impl PartialOrd for SpanData {
+    fn partial_cmp(&self, other: &Self) -> Option<Ordering> {
+        Some((self.lo, self.hi, self.ctxt).cmp(&(other.lo, other.hi, other.ctxt)))
```
cjgillot, line 451:
```suggestion
        Some(self.cmp(other))
```

compiler/rustc_data_structures/src/graph/vec_graph/mod.rs
```diff
@@ -20,7 +20,7 @@ pub struct VecGraph<N: Idx> {
 impl<N: Idx> VecGraph<N> {
     pub fn new(num_nodes: usize, mut edge_pairs: Vec<(N, N)>) -> Self {
         // Sort the edges by the source -- this is important.
-        edge_pairs.sort();
+        edge_pairs.sort_by_cached_key(|&edge_pairs| (edge_pairs.0.index(), edge_pairs.1.index()));
```
cjgillot, line 23: Why use the `cached_` version?

compiler/rustc_span/src/lib.rs
```diff
@@ -440,6 +440,18 @@ pub struct SpanData {
     pub parent: Option<LocalDefId>,
 }
 
+impl Ord for SpanData {
+    fn cmp(&self, other: &Self) -> Ordering {
+        (self.lo, self.hi, self.ctxt).cmp(&(other.lo, other.hi, other.ctxt))
```
cjgillot, line 445: Could you explicit destructure so we get a compile error when a field is added to `SpanData`? Can you add a comment on why we ignore `parent`?
`let SpanData { lo: s_lo, hi: s_hi, ctxt: s_ctxt, parent: _ } = self;` and likewise for `other`.

compiler/rustc_hir/src/hir_id.rs
```diff
@@ -11,13 +12,29 @@ use std::fmt;
...
 /// the code base.
-#[derive(Copy, Clone, PartialEq, Eq, Hash, Debug, PartialOrd, Ord)]
+#[derive(Copy, Clone, PartialEq, Eq, Hash, Debug)]
 #[derive(Encodable, Decodable)]
 pub struct HirId {
     pub owner: LocalDefId,
     pub local_id: ItemLocalId,
 }
 
+impl Ord for HirId {
```
cjgillot, line 22: Can you remove the impls for PartialOrd/Ord for HirId too? It compares LocalDefIds, which we want to avoid.

cjgillot (commented):
Thanks @pierwill. I added a few nits. The main remark is that we'd like to remove the PartialOrd/Ord impl for HirId, since it has exactly the same issue as the impl for LocalDefId.

compiler/rustc_data_structures/src/graph/vec_graph/mod.rs
pierwill: No particular reason. I'd seen this method used over `sort_by_key` elsewhere in the compiler.

compiler/rustc_hir/src/hir_id.rs
pierwill: On it. 👍

compiler/rustc_index/src/vec.rs
```diff
@@ -12,7 +12,7 @@ use std::vec;
 /// Represents some newtyped `usize` wrapper.
 ///
 /// Purpose: avoid mixing indexes for different bitvector domains.
-pub trait Idx: Copy + 'static + Ord + Debug + Hash {
+pub trait Idx: Copy + 'static + Eq + PartialEq + Debug + Hash {
```
pierwill, line 15: I don't remember why I had to do this... 🤔

pierwill: Ah, okay, it was needed in `rustc_data_structures::graph`.

pierwill:
Going to need to locate the source of an infinite loop...

compiler/rustc_middle/src/mir/mono.rs
```diff
@@ -366,7 +366,13 @@ impl<'tcx> CodegenUnit<'tcx> {
                             // instances into account. The others don't matter for
                             // the codegen tests and can even make item order
                             // unstable.
-                            InstanceDef::Item(def) => Some(def.did.index.as_usize()),
+                            InstanceDef::Item(def) => Some(
+                                def.did
+                                    .as_local()
+                                    .map(|def_id| tcx.hir().local_def_id_to_hir_id(def_id))
+                                    .unwrap()
```
pierwill, line 373: This will probably panic at some point...

pierwill:
Huh. I did not expect this to affect any UI tests... https://github.com/rust-lang/rust/runs/4088478139?check_suite_focus=true#step:25:7148

compiler/rustc_middle/src/mir/query.rs
```diff
@@ -358,13 +358,57 @@ pub enum ConstraintCategory {
...
+            _ => None,
+        };
+        let o = match other {
+            Self::ClosureUpvar(hir_id) => Some(hir_id.index()),
+            _ => None,
+        };
+
+        s.cmp(&o)
+    }
+}
```
pierwill, line 375: This doesn't feel right to me...

compiler/rustc_middle/src/mir/query.rs
```diff
@@ -358,13 +358,57 @@ pub enum ConstraintCategory {
...
+            &Self::ClosureUpvar(h) => Some(h.index()),
+        };
+        let o = match other {
+            &Self::Normal => None,
+            &Self::ClosureUpvar(h) => Some(h.index()),
+        };
+
+        s.cmp(&o)
+    }
+}
```
pierwill, line 404: Same here.

Could these `Ord` implementations, if they're incorrect, be causing the changes in the UI tests?

compiler/rustc_data_structures/src/sorted_map/index_map.rs
```diff
@@ -97,7 +97,7 @@ impl<I: Idx, K: Eq, V: Eq> Eq for SortedIndexMultiMap<I, K, V> {}
 impl<I: Idx, K: PartialEq, V: PartialEq> PartialEq for SortedIndexMultiMap<I, K, V> {
     fn eq(&self, other: &Self) -> bool {
         // No need to compare the sorted index. If the items are the same, the index will be too.
-        self.items == other.items
+        self == other
```
pierwill, line 100: Is this correct?

compiler/rustc_index/src/bit_set.rs
```diff
@@ -590,13 +590,13 @@ impl<T: Idx> SparseBitSet<T> {
 
     fn contains(&self, elem: T) -> bool {
         assert!(elem.index() < self.domain_size);
-        self.elems.contains(&elem)
+        self.elems.iter().any(|e| elem.index() == e.index())
```
bjorn3, line 593: Are these changes necessary? `Eq` is still implemented, right?

pierwill: Ah! Let me check. I think there were some changes need as I was working to keep the build working, but not after completing certain changes. (This must be a common experience...)

compiler/rustc_data_structures/src/graph/vec_graph/mod.rs
```diff
@@ -20,7 +20,7 @@ pub struct VecGraph<N: Idx> {
 impl<N: Idx> VecGraph<N> {
     pub fn new(num_nodes: usize, mut edge_pairs: Vec<(N, N)>) -> Self {
         // Sort the edges by the source -- this is important.
-        edge_pairs.sort();
+        edge_pairs.sort_by_key(|&edge_pairs| (edge_pairs.0.index(), edge_pairs.1.index()));
```
pierwill, line 23: Looks like this change was necessary.

compiler/rustc_span/src/lib.rs
```diff
@@ -435,6 +435,33 @@ pub struct SpanData {
...
 }
 
+impl Ord for SpanData {
+    fn cmp(&self, other: &Self) -> Ordering {
+        let SpanData {
+            lo: s_lo,
+            hi: s_hi,
+            ctxt: s_ctxt,
+            // `HirId` does not implement `Ord`.
+            parent: _,
```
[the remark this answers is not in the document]
pierwill: Now that I think of it, if we're [not removing](https://github.com/rust-lang/rust/pull/90408#issuecomment-963621886) `Ord` from `HirId` in this PR, this change isn't necessary.

compiler/rustc_span/src/lib.rs
```diff
@@ -435,6 +435,33 @@ pub struct SpanData {
     pub parent: Option<LocalDefId>,
 }
 
+impl Ord for SpanData {
```
cjgillot, line 438: This should probably add a comment: "Order spans by position in the file."

compiler/rustc_span/src/lib.rs
```diff
@@ -435,6 +435,33 @@ pub struct SpanData {
...
+    fn cmp(&self, other: &Self) -> Ordering {
+        let SpanData {
+            lo: s_lo,
+            hi: s_hi,
+            ctxt: s_ctxt,
+            // `HirId` does not implement `Ord`.
```
cjgillot, line 444:
```suggestion
            // `LocalDefId` does not implement `Ord`. Other fields are enough to determine in-file order.
```

compiler/rustc_span/src/lib.rs
```diff
@@ -435,6 +435,33 @@ pub struct SpanData {
...
+            parent: _,
+        } = self;
+        let SpanData {
+            lo: o_lo,
+            hi: o_hi,
+            ctxt: o_ctxt,
+            // `HirId` does not implement `Ord`.
```
cjgillot, line 451:
```suggestion
            // `LocalDefId` does not implement `Ord`. Other fields are enough to determine in-file order.
```

compiler/rustc_metadata/src/rmeta/encoder.rs
```diff
@@ -1300,7 +1300,7 @@ impl EncodeContext<'a, 'tcx> {
             })
             .collect::<Vec<_>>();
         // Sort everything to ensure a stable order for diagnotics.
-        keys_and_jobs.sort_by_key(|&(def_id, _, _)| def_id);
+        keys_and_jobs.sort_by_key(|&(def_id, _, _)| def_id.index());
```
cjgillot, line 1303: Should `mir_keys` eventually be an `IndexSet` to avoid the need for sorting?

cjgillot (approved):
Thanks @pierwill, LGTM. A few remaining nits and this will be ready for approval.

compiler/rustc_metadata/src/rmeta/encoder.rs
pierwill: Sounds good to me.

pierwill:
Thanks, @cjgillot! Addressed your comments and rebased. 👍

cjgillot:
Just in case, to avoid any surprise
@bors try @rust-timer queue

pierwill:
> **Summary**: This change led to large relevant mixed results 🤷 in compiler performance.
> 
> * Large improvement in instruction counts (up to -2.9% on `full` builds of `deeply-nested-async`)
> * Large regression in instruction counts (up to 3.1% on `incr-unchanged` builds of `deeply-nested-async`)

So the effects are mixed, but limited to async heavy code. Nice that there's no regression in `incr-patched` builds, which seem like the most common cases for those developing async heavy code.

compiler/rustc_metadata/src/rmeta/encoder.rs
pierwill: This is a very simple. Maybe it's worth doing here. Could this have positive performance benefits?

pierwill: @cjgillot For consideration: https://github.com/pierwill/rust/pull/1.

pierwill: Opened https://github.com/rust-lang/rust/pull/90842.

pierwill:
Looks like a rebase might fix this...

cjgillot:
The failure seems legit. Probably just needs blessing the output.

pierwill:
> 

Testing this now 👍

pierwill:
> The failure seems legit. Probably just needs blessing the output.

Testing now 👍

pierwill:
@cjgillot Maybe this will work? https://github.com/rust-lang/rust/pull/90408/commits/ab9dc25e22150a0a3a557a3f4da5d4356487b6fe

pierwill:
New error has crept in. I'm having trouble making sense of this:

```
error[E0277]: can't compare `T` with `_`
   --> compiler/rustc_index/src/bit_set.rs:721:31
    |
721 |             if range.contains(e) {
    |                      -------- ^ no implementation for `T < _` and `T > _`
    |                      |
    |                      required by a bound introduced by this call
    |
note: required by a bound in `contains`
   --> /Users/wpierce/repos/rust/library/core/src/ops/range.rs:812:12
    |
812 |         T: PartialOrd<U>,
    |            ^^^^^^^^^^^^^ required by this bound in `contains`
```

pierwill:
Having trouble understanding this error:

```
error[E0277]: can't compare `T` with `_`
   --> compiler/rustc_index/src/bit_set.rs:721:31
    |
721 |             if range.contains(e) {
    |                      -------- ^ no implementation for `T < _` and `T > _`
    |                      |
    |                      required by a bound introduced by this call
    |
note: required by a bound in `contains`
   --> /Users/u/repos/rust/library/core/src/ops/range.rs:812:12
    |
812 |         T: PartialOrd<U>,
    |            ^^^^^^^^^^^^^ required by this bound in `contains`
```

pierwill:
This code was added in https://github.com/rust-lang/rust/commit/03afb61b53e27019166b8a2637a5eb7094017300. @Mark-Simulacrum: could offer some advice on troubleshooting this? 😇 

https://github.com/rust-lang/rust/runs/4433279740?check_suite_focus=true#step:25:509

Mark-Simulacrum:
I'm not sure what aid I can offer, but it likely that the relevant method will want a where clause bound on having Ord, or something like that.

compiler/rustc_data_structures/src/graph/vec_graph/mod.rs
Mark-Simulacrum, line 23: This seems a little concerning - I would rather expect that this data structure/method would require that the index type implements Ord. (Some types which are comparable, e.g. mir block indices, should continue to implement Ord, even as we remove it from DefId). Otherwise we're just hiding the actual Ord dependence which doesn't really help, imo.

pierwill:
> Having trouble understanding this error:
> 
> ```
> error[E0277]: can't compare `T` with `_`
>    --> compiler/rustc_index/src/bit_set.rs:721:31
> [...]
> ```

What does it mean to compare with `_`? 😕

compiler/rustc_data_structures/src/graph/vec_graph/mod.rs
pierwill: How does this look? https://github.com/rust-lang/rust/pull/90408/commits/44e023a64901a47348284b67b2c32c3817c1beb3

pierwill:
Still working on some improvements in the spirit of this comment. @rustbot label -S-waiting-on-review

pierwill:
@cjgillot I squashed one small commit just now, FYI. Not sure what caused that error...

bjorn3:
It says that it was canceled. Probably spurious.

@bors retry

pierwill:
@cjgillot Looks like bors isn't picking up the approval:

Screen Shot 2021-12-09 at 5 08 28 PM
 
My rebase probably invalidated it.

pierwill:
This change has been failing the same Windows backtrace test as https://github.com/rust-lang/rust/pull/89836. We're [assessing](https://github.com/rust-lang/rust/pull/89836#issuecomment-992646395) whether to bless the test or continue digging.

pierwill:
Once https://github.com/rust-lang/rust/pull/89836 is merged (🤞🤞), this should be ready to go. :)

matthiaskrgr:
Just for the record, the failure here https://github.com/rust-lang/rust/pull/90408#issuecomment-991879747 is probably unrelated to this PR since it happened in other ci-runs as well :( :  https://github.com/rust-lang/rust/pull/91995#issuecomment-995710617

Made a tracking ticket: https://github.com/rust-lang/rust/issues/92000

pierwill:
This is ready for another bors attempt! @cjgillot

pierwill:
See previous comment here: https://github.com/rust-lang/rust/pull/90408#issuecomment-967448819.

PR #90411, opened 2021-10-29
[EXPERIMENT] rustdoc: Mark `Context::tcx()` as `#[inline(always)]`

camelid:
I'm wondering if `tcx()` not being inlined could be affecting the perf in #90391.

r? @ghost


camelid:
It could also make sense to move `tcx` out of `shared`, since it's behind an `Rc`.

For now, let's see what effect this has.

@bors try @rust-timer queue

camelid:
Hmm, mixed or slightly negative results.

PR #90412, opened 2021-10-30
Remove unnecessary `macro_use`s in rustdoc

jyn514:


src/librustdoc/html/render/context.rs
```diff
@@ -22,7 +22,7 @@ use super::{
     BASIC_KEYWORDS,
 };
 
-use crate::clean;
+use crate::{clean, try_err};
```
camelid, line 25: nit: It feels like these should be on separate lines. Maybe also merge together `use crate::clean` and `use crate::clean::ExternalCrate`?

camelid (approved):
Thanks! r=me with CI passing and, optionally, the nit below addressed

Issue #56604, opened 2018-12-07
Coercing &mut to *const should not create a shared reference


RalfJung:
It has long been a rule in Rust that you must not mutate through a shared reference, or a raw pointer obtained from a shared reference.

Unfortunately, that rule currently forbids the following code:
```rust
fn direct_mut_to_const_raw() {
    let x = &mut 0;
    let y: *const i32 = x;
    unsafe { *(y as *mut i32) = 1; }
    assert_eq!(*x, 1);
}
```
The reason for this is that coercing `&mut T` to `*const T` implicitly first creates a shared reference and then coerces that to `*const T`, meaning `y` in the example above is technically a raw pointer obtained from a shared reference.

We should fix our coercion logic to no longer create this intermediate shared reference.

Cc @eddyb @nikomatsakis

RalfJung:
I believe the relevant code might be

https://github.com/rust-lang/rust/blob/f504d3f05985f158c4845816c3d5f7b7a8afeb13/src/librustc_typeck/check/coercion.rs#L734-L752

but I am not sure because this is type *checking*. I have no idea where the code that decides about lowering of coercions to reborrows/casts lives.

eddyb:
> I have no idea where the code that decides about lowering of coercions to reborrows/casts lives.

What do you mean? That code is generating a deref and a borrow, each of those gets turned into the equivalent MIR later, and the result is as if the user wrote `&*` explicitly.

eddyb:
@nikomatsakis I believe the comment above `if is_ref {` has been outdated for quite a while (and makes no sense for NLL), can you confirm?

RalfJung:
@eddyb that code generates "adjustments". I have no idea what that means. I found no use of `AutoBorrow::RawPtr` that would turn this into casts or reborrows or so.

eddyb:
Oh I got confused as to what this is doing, I missed the `AutoBorrow::RawPtr` bit, looks like "borrow" and "cast reference to raw pointer" are fused into one adjustment. The lowering is here, btw:
https://github.com/rust-lang/rust/blob/1c3236afc3560b5781acf942a4746aeb11e5374e/src/librustc_mir/hair/cx/expr.rs#L152-L180

So looking again at the testcase, `y as *mut i32` still goes through the reference->raw pointer coercion logic, but it does so for coercing `&mut T` to `*mut T` which does a mutable reborrow instead of an immutable reborrow (and then there's a separate `*mut T` to `*const T` coercion).

[...]

PR #90413, opened 2021-10-30
`addr_of!` grants mutable access, maybe?

tmiasko:
The exact set of permissions granted when forming a raw reference is
currently undecided https://github.com/rust-lang/rust/issues/56604.

To avoid presupposing any particular outcome, adjust the const
qualification to be compatible with decision where raw reference
constructed from `addr_of!` grants mutable access.

Additionally, to avoid keeping `MaybeMutBorrowedLocals` in sync with
const qualification, remove it. It's no longer used.

@rust-lang/wg-const-eval


tmiasko:
@oli-obk @ecstatic-morse to address concerns from #90214 about duplication with `MaybeMutBorrowedLocals`, I decided to remove it, since we no longer used it anywhere else, but let me know if you prefer to retain it.

oli-obk:
Ah, if that was the only user, then this is perfect!

compiler/rustc_const_eval/src/transform/check_consts/resolver.rs
```diff
@@ -265,6 +273,17 @@ pub(super) struct State {
     pub borrow: BitSet<Local>,
 }
 
+impl Clone for State {
```
oli-obk, line 276: Please put the comment from the commit on this impl, so future readers don't have to look at blame

oli-obk:
Impl lgtm

r? @RalfJung

compiler/rustc_const_eval/src/transform/check_consts/resolver.rs
```diff
@@ -265,6 +273,19 @@ pub(super) struct State {
...
+    fn clone(&self) -> Self {
+        State { qualif: self.qualif.clone(), borrow: self.borrow.clone() }
+    }
+
+    // Data flow engine when possible uses `clone_from` for domain values. Implementing it should
+    // avoid some intermediate memory allocations.
+    fn clone_from(&mut self, other: &Self) {
```
RalfJung, line 283: So this is not the default implementation generated by `derive`?

src/test/ui/mir-dataflow/indirect-mutation-offset.rs
```diff
@@ -1,48 +0,0 @@
...
-}
-
-#[rustc_mir(rustc_peek_indirectly_mutable,stop_after_dataflow)]
-const BOO: i32 = {
-    let x = PartialInteriorMut {
-        zst: [],
-        cell: UnsafeCell::new(0),
-    };
-
-    let p_zst: *const _ = &x.zst ; // Doesn't cause `x` to get marked as indirectly mutable.
```
RalfJung, line 28: The ZST is `Freeze`, so this test should behave as before, shouldn't it? Why is it being removed?

src/test/ui/mir-dataflow/indirect-mutation-offset.rs
```diff
@@ -1,48 +0,0 @@
...
-
-    let rmut_cell = unsafe {
-        // Take advantage of the fact that `zst` and `cell` are at the same location in memory.
-        // This trick would work with any size type if miri implemented `ptr::offset`.
```
RalfJung, line 32: FWIW, Miri (https://github.com/rust-lang/miri/) implements it, but the CTFE interpreter does not. Or rather, it didn't use to -- these days, `offset` is unstably `const fn`.

compiler/rustc_const_eval/src/transform/check_consts/resolver.rs
RalfJung: Ah, that seems like an improvement worth making... but it has nothing to do with this PR, of course.

src/test/ui/mir-dataflow/indirect-mutation-offset.rs
tmiasko: The test requires `MaybeMutBorrowedLocals` data flow which no longer exists.

compiler/rustc_const_eval/src/transform/check_consts/resolver.rs
tmiasko: Moved to a separate pull request.

src/test/ui/mir-dataflow/indirect-mutation-offset.rs
RalfJung: Ah okay. I don't know what this mir_peek stuff is doing anyway so I will just ignore this then. ;)

RalfJung:
The analysis makes sense to me. The MaybeMutBorrowedLocals removal is code I have never seen but Oli reviewed it, so let's go ahead.

@bors r=RalfJung,oli-obk

PR #90417, opened 2021-10-30
stabilize `relaxed_struct_unsize`

### LINKED ISSUE #81793: Tracking Issue for relaxed struct unsizing rules
The feature gate for the issue is `#![feature(relaxed_struct_unsize)]`.

This features changes the when a struct implements `Unsize` which
is required for unsize coercion.

A generic struct previously implemented `Unsize` if and only if:
- the tail field depends on at least one type or const parameter
- **all type and const parameters used in the last field must only be used in the last field**
- the target struct can be created from the source by replacing only the parameters found in the last struct field
- the tail field implements `Unsize` from source to target

With this feature gate the requirements are changed to:
- the tail field depends on at least one type or const parameter **not used in any other field**
- the target struct can be created from the source by replacing only the parameters **only** found in the last struct field
- the tail field implements `Unsize` from source to target

### About tracking issues

Tracking issues are used to record the overall progress of implementation.
They are also used as hubs connecting to other relevant issues, e.g., bugs or open design questions.
A tracking issue is however *not* meant for large scale discussion, questions, or bug reports about a feature.
Instead, open a dedicated issue for the specific matter and add the relevant feature gate label.

### Steps

- [ ] Adjust documentation ([see instructions on rustc-dev-guide][doc-guide])
- [ ] Stabilization PR ([see instructions on rustc-dev-guide][stabilization-guide])

[stabilization-guide]: https://rustc-dev-guide.rust-lang.org/stabilization_guide.html#stabilization-pr
[doc-guide]: https://rustc-dev-guide.rust-lang.org/stabilization_guide.html#documentation-prs

### Implementation history

This feature has been implemented in #80726

lcnr:
closes #81793

the fcp is already complete.


ehuss:
Does this affect the documentation at https://doc.rust-lang.org/nightly/core/marker/trait.Unsize.html and https://doc.rust-lang.org/nightly/reference/type-coercions.html#unsized-coercions?

lcnr:
yes, the documentation was previously incorrect and is now isn't :sweat_smile: I therefore don't think that the documentation has to change with this stabilization.

src/test/ui/feature-gates/feature-gate-relaxed_struct_unsize.rs
```diff
@@ -1,10 +0,0 @@
-// Test that we allow unsizing even if there is an unchanged param in the
-// field getting unsized.
-struct A<T, U: ?Sized + 'static>(T, B<T, U>);
-struct B<T, U: ?Sized>(T, U);
-
-fn main() {
-    let x: A<[u32; 1], [u32; 1]> = A([0; 1], B([0; 1], [0; 1]));
-    let y: &A<[u32; 1], [u32]> = &x; //~ ERROR mismatched types
```
wesleywiser, line 8: I assume this was deleted because the error no longer occurs? Should this be turned into a check-pass test then?

lcnr: we already have this as a check-pass test, this was only to check that the feature gate is actually needed

wesleywiser: Ah, excellent. Thanks!

Issue #90418, opened 2021-10-30
rustdoc: Tracking issue for rustdoc internals


GuillaumeGomez:
This is a tracking issue for the internal feature *rustdoc internals*.
The feature gate for this issue is `rustdoc_internals`.

It hosts the following non-exhaustive list of subfeatures (not guaranteed to be up to date):

* `#[doc(keyword = "…")]`
  * Initial impl: https://github.com/rust-lang/rust/pull/51140
* `#[doc(attribute = "…")]`
  * Initial impl: https://github.com/rust-lang/rust/pull/142472

> [!NOTE]
>
> Tracking issues are used to record the overall progress of implementation.
> They are also used as hubs connecting to other relevant issues, e.g., bugs or open design questions.
>
> A tracking issue is however *not* meant for large scale discussion, questions, or bug reports.
> Instead, open a dedicated issue for the specific matter or post in the [#t-rustdoc](https://rust-lang.zulipchat.com/#narrow/channel/266220-t-rustdoc) Zulip channel.
>
> Discussion comments will get marked as off-topic or deleted.
> Repeated discussions on the tracking issue may lead to the tracking issue getting locked.

jhpratt:
I thought we didn't need tracking issues for features not intended to be stabilized?

GuillaumeGomez:
Closing as we don't plan to ever stabilize these flags.

PR #90420, opened 2021-10-30
Create rustdoc_internals feature gate

GuillaumeGomez:
As suggested by @camelid [here](https://github.com/rust-lang/rust/pull/90398#issuecomment-955093851), since `doc_keyword` and `doc_primitive` aren't meant to be stabilized, we could put them behind a same feature flag.

This is pretty much what it would look like (needs to update the tests too).

The tracking issue is https://github.com/rust-lang/rust/issues/90418.


compiler/rustc_feature/src/active.rs
```diff
@@ -688,6 +690,10 @@ declare_features! (
     /// not changed from prior instances of the same struct (RFC #2528)
     (incomplete, type_changing_struct_update, "1.58.0", Some(86555), None),
 
+    /// Allows using internal rustdoc features like doc(primitive) or doc(keyword)
+    /// without a future-incompat warning.
```
camelid, line 694: `doc(keyword)` on stable is a hard error, only `doc(primitive)` is a warning. Also, please use inline code ;)

compiler/rustc_ast_passes/src/feature_gate.rs
```diff
@@ -322,7 +322,8 @@ impl<'a> Visitor<'a> for PostExpansionVisitor<'a> {
                     cfg_hide => doc_cfg_hide
                     masked => doc_masked
                     notable_trait => doc_notable_trait
-                    keyword => doc_keyword
+                    keyword => rustdoc_internals
+                    primitive => rustdoc_internals
```
camelid, line 326: I'm pretty sure this will make `doc(primitive)` a hard error. Where is feature-gating of `doc(primitive)` currently handled?

jyn514: It's in `check_attr.rs`.
```
[nix-shell:~/rustc3]$ rg doc_primitive compiler/
compiler/rustc_feature/src/active.rs
664:    (active, doc_primitive, "1.56.0", Some(88070), None),

compiler/rustc_passes/src/check_attr.rs
965:                            if !self.tcx.features().doc_primitive {
```

jyn514:
This seems ... kind of unnecessary to me? We can document on the tracking issues that these aren't meant to be stabilized, I don't think it also needs to be in the source code of everything using them. And as you've noticed the logic for `doc_primitive` is pretty different from that of `doc_keyword`, so I don't think it's particularly useful to combine them.

camelid:
> This seems ... kind of unnecessary to me? We can document on the tracking issues that these aren't meant to be stabilized, I don't think it also needs to be in the source code of everything using them.

The standard library often uses the naming convention of `X_internals` to indicate that something is perma-unstable. Whereas `#![feature(doc_keyword)]` doesn't give any indication that it's only meant for internal use. People don't necessarily look at tracking issues, so IMO it's best to indicate in the feature name that something is internal-only. Another example is `#[rustc_layout(debug)]` which is under the feature-gate `rustc_attrs`, not something like `debug_layout`.

If merging the feature gates would be too complicated, I'm okay to leave them separate, but I would prefer to merge them and make it clear in the name that they are internal-only.

> I don't think it also needs to be in the source code of everything using them.

They're only used in the standard library and rustdoc IIUC, and this would reduce the number of feature gates necessary?

jyn514:
Hmm, ok, seems reasonable.

src/test/ui/feature-gates/feature-gate-rustdoc_internals.stderr
```diff
@@ -0,0 +1,12 @@
+error[E0658]: `#[doc(keyword)]` is experimental
```
camelid, line 1: Can you also change this message to something like the following?

```suggestion
error[E0658]: `#[doc(keyword)]` is meant for internal use only
```

GuillaumeGomez:
Updated the message and set the PR as ready to review.

compiler/rustc_feature/src/active.rs
```diff
@@ -417,6 +417,7 @@ declare_features! (
     /// Allows `'a: { break 'a; }`.
     (active, label_break_value, "1.28.0", Some(48594), None),
 
+    // FIXME: to be removed once rustdoc_internals is merged.
```
camelid, line 420: (This and the below FIXME still need to be fixed.)

GuillaumeGomez: We need this PR to be merged first to do so (and wait for a new nightly). :)

camelid: Why do we need to wait?

GuillaumeGomez: Because the stage 0 is compiled with an old rustc version which doesn't have that feature yet. So we need to keep the old ones around until the nightly used for stage 0 is updated.

camelid: I don't think that's necessary. You can just do this:

```rust
#![cfg_attr(bootstrap, feature(doc_keyword)]
#![cfg_attr(not(bootstrap), feature(rustdoc_internals)]
```

(You may have to swap the `bootstrap` and `not(bootstrap)`; I always get them mixed up.)

GuillaumeGomez: I tried and it failed. Also, the problem is mainly in `libcore` and `libstd` which (unless I'm wrong) needs an older rustc in any case. At least I tried with `bootstrap` and failed. But maybe I simply inverted the bootstrap?

GuillaumeGomez:
Updated!

GuillaumeGomez:
Updated!

GuillaumeGomez:
@camelid I think my comment wasn't added so: you were right, I inverted the `cfg_attr` condition for `bootstrap`. Works much better now. :)

library/core/src/lib.rs
```diff
@@ -161,7 +161,10 @@
 #![feature(decl_macro)]
 #![feature(doc_cfg)]
 #![feature(doc_notable_trait)]
-#![feature(doc_primitive)]
+// FIXME: to be removed once rustdoc_internals is merged.
```
camelid, line 164: Please remove these FIXMEs. Whoever updates the bootstrap compiler will update all the uses of `cfg(bootstrap)` :)

src/test/ui/feature-gates/feature-gate-rustdoc_internals.rs
```diff
@@ -0,0 +1,5 @@
+#[doc(keyword = "match")] //~ ERROR: `#[doc(keyword)]` is meant for internal use only
```
camelid, line 1: Why did you add this test? It looks identical to `feature-gate-doc_keyword.rs`. I think it'd be better to just rename the `doc_keyword` test to `rustdoc_internals`.

src/test/ui/feature-gates/feature-gate-rustdoc_internals.stderr
```diff
@@ -0,0 +1,12 @@
+error[E0658]: `#[doc(keyword)]` is meant for internal use only
```
camelid, line 1: Can you add a test to make sure the "feature gate renamed error" is shown for `#![feature(doc_keyword)]` and `doc_primitive`?

compiler/rustc_ast_passes/src/feature_gate.rs
```diff
@@ -325,8 +325,12 @@ impl<'a> Visitor<'a> for PostExpansionVisitor<'a> {
                     cfg_hide => doc_cfg_hide
                     masked => doc_masked
                     notable_trait => doc_notable_trait
-                    keyword => doc_keyword
                 );
+
+                if nested_meta.has_name(sym::keyword) {
+                    let msg = concat!("`#[doc(keyword)]` is meant for internal use only");
```
camelid, line 331: Hmm, maybe just allow a custom message to be chosen with the macro? I worry this will fall out of sync with the macro.

src/test/ui/feature-gates/feature-gate-rustdoc_internals.rs
```diff
@@ -0,0 +1,5 @@
...
+/// wonderful
+mod foo{}
```
camelid, line 3: nit (feel free to ignore):

```suggestion
mod foo {}
```

src/test/ui/rustdoc/renamed_features.rs
```diff
@@ -0,0 +1,5 @@
+#![feature(doc_keyword)] //~ ERROR
```
camelid, line 1: Could you rename this test to something a bit more descriptive, that includes rustdoc_internals in it?

GuillaumeGomez:
> Could you rename this test to something a bit more descriptive, that includes rustdoc_internals in it?

Any suggestion? I'm very bad at naming...

camelid:
> > Could you rename this test to something a bit more descriptive, that includes rustdoc_internals in it?
> 
> Any suggestion? I'm very bad at naming...

I think this should be good enough: `renamed-features-rustdoc_internals.rs`

GuillaumeGomez:
Ok!

GuillaumeGomez:
Updated the name.

compiler/rustc_feature/src/active.rs
```diff
@@ -161,6 +161,9 @@ declare_features! (
     (active, profiler_runtime, "1.18.0", None, None),
     /// Allows using `rustc_*` attributes (RFC 572).
     (active, rustc_attrs, "1.0.0", None, None),
+    /// Allows using internal rustdoc features like `doc(primitive)` or `doc(keyword)`
+    /// without a future-incompat warning.
```
camelid, line 165: Only `doc(primitive)` is a future-incompat warning. To avoid this getting out of date, let's just do this:

```suggestion
    /// Allows using internal rustdoc features like `doc(primitive)` or `doc(keyword)`.
```

GuillaumeGomez:
Done!

compiler/rustc_feature/src/active.rs
```diff
@@ -161,6 +161,8 @@ declare_features! (
     (active, profiler_runtime, "1.18.0", None, None),
     /// Allows using `rustc_*` attributes (RFC 572).
     (active, rustc_attrs, "1.0.0", None, None),
+    /// Allows using internal rustdoc features like `doc(primitive)` or `doc(keyword)`.
+    (active, rustdoc_internals, "1.58.0", Some(90418), None),
```
camelid, line 165: This is in the wrong section; it should be next to doc_notable_trait.

camelid: Also, can you create an actual tracking issue? We shouldn't just link to this PR.

compiler/rustc_feature/src/removed.rs
```diff
@@ -76,6 +76,12 @@ declare_features! (
     /// Allows the use of `#[derive(Anything)]` as sugar for `#[derive_Anything]`.
     (removed, custom_derive, "1.32.0", Some(29644), None,
      Some("subsumed by `#[proc_macro_derive]`")),
+    /// Allows using `#[doc(keyword = "...")]`.
+    (removed, doc_keyword, "1.28.0", Some(51315), None,
+     Some("merged into `#![feature(rustdoc_internals)]`")),
+    /// Allows using doc(primitive) without a future-incompat warning
```
camelid, line 82: nit:

```suggestion
    /// Allows using `doc(primitive)` without a future-incompat warning
```

camelid (approved):
Left a few comments. r=me after fixing those.

GuillaumeGomez:
> Also, can you create an actual tracking issue? We shouldn't just link to this PR.

It does link to the tracking issue. Unless I missed something?

Otherwise, I updated your comments.

camelid:
Sorry, maybe GitHub glitched and showed me the wrong page. Tracking issue looks good to me.

compiler/rustc_feature/src/active.rs
```diff
@@ -206,6 +206,8 @@ declare_features! (
     (active, rustc_allow_const_fn_unstable, "1.49.0", Some(69399), None),
     /// Allows using compiler's own crates.
     (active, rustc_private, "1.0.0", Some(27812), None),
+        /// Allows using internal rustdoc features like `doc(primitive)` or `doc(keyword)`.
```
camelid, line 209:
```suggestion
    /// Allows using internal rustdoc features like `doc(primitive)` or `doc(keyword)`.
```

compiler/rustc_feature/src/removed.rs
```diff
@@ -76,6 +76,12 @@ declare_features! (
     /// Allows the use of `#[derive(Anything)]` as sugar for `#[derive_Anything]`.
     (removed, custom_derive, "1.32.0", Some(29644), None,
      Some("subsumed by `#[proc_macro_derive]`")),
+    /// Allows using `#[doc(keyword = "...")]`.
+    (removed, doc_keyword, "1.28.0", Some(51315), None,
+     Some("merged into `#![feature(rustdoc_internals)]`")),
+    /// Allows using `doc(primitive)` without a future-incompat warning
```
camelid, line 82:
```suggestion
    /// Allows using `doc(primitive)` without a future-incompat warning.
```

camelid (approved):
r=me with these two comments fixed

compiler/rustc_feature/src/removed.rs
camelid: (Totally optional to fix this by the way.)

GuillaumeGomez:
@camelid Your comments made perfect sense (and I have no clue how I messed up the indent...). Updated!

camelid:
Thanks!

@bors r+ rollup

PR #89899, opened 2021-10-15
Add #[must_use] to remaining alloc functions

jkugelman:
I've run out of compelling reasons to group functions together across crates so I'm just going to go module-by-module. This is everything remaining from the `alloc` crate.

I ignored these because they might be used to purposefully leak memory... or other allocator shenanigans? I dunno. I'll add them if y'all tell me to.

```rust
alloc::alloc          unsafe fn alloc(layout: Layout) -> *mut u8;
alloc::alloc          unsafe fn alloc_zeroed(layout: Layout) -> *mut u8;
alloc::sync::Arc<T>   fn into_raw(this: Self) -> *const T;
```

I don't know why clippy ignored these. I added them myself:

```rust
alloc::collections::btree_map::BTreeMap<K, V>   fn range<T: ?Sized, R>(&self, range: R) -> Range<'_, K, V>;
alloc::collections::btree_set::BTreeSet<T>      fn range<K: ?Sized, R>(&self, range: R) -> Range<'_, T>;
```

I added these non-mutating `mut` functions:

```rust
alloc::collections::btree_map::BTreeMap<K, V>     fn range_mut<T: ?Sized, R>(&mut self, range: R) -> RangeMut<'_, K, V>;
alloc::collections::btree_map::BTreeMap<K, V>     fn iter_mut(&mut self) -> IterMut<'_, K, V>;
alloc::collections::btree_map::BTreeMap<K, V>     fn values_mut(&mut self) -> ValuesMut<'_, K, V>;
alloc::collections::linked_list::LinkedList<T>    fn iter_mut(&mut self) -> IterMut<'_, T>;
alloc::collections::linked_list::LinkedList<T>    fn cursor_front_mut(&mut self) -> CursorMut<'_, T>;
alloc::collections::linked_list::LinkedList<T>    fn cursor_back_mut(&mut self) -> CursorMut<'_, T>;
alloc::collections::linked_list::LinkedList<T>    fn front_mut(&mut self) -> Option<&mut T>;
alloc::collections::linked_list::LinkedList<T>    fn back_mut(&mut self) -> Option<&mut T>;
alloc::collections::linked_list::CursorMut<'a, T> fn current(&mut self) -> Option<&mut T>;
alloc::collections::linked_list::CursorMut<'a, T> fn peek_next(&mut self) -> Option<&mut T>;
alloc::collections::linked_list::CursorMut<'a, T> fn peek_prev(&mut self) -> Option<&mut T>;
[...]

PR #90427, opened 2021-10-31
Add #[must_use] to alloc functions that would leak memory

jkugelman:
As [requested](https://github.com/rust-lang/rust/pull/89899#issuecomment-955600779) by @joshtriplett.

> Please do go ahead and add the ones whose only legitimate use for ignoring the return value is leaking memory. (In a separate PR please.) I think it's sufficiently error-prone to call something like alloc and ignore the result that it's legitimate to require `let _ =` for that.

I added `realloc` myself. Clippy ignored it because of its `mut` argument.

```rust
alloc/src/alloc.rs:123:1   alloc   unsafe fn realloc(ptr: *mut u8, layout: Layout, new_size: usize) -> *mut u8;
```

Parent issue: #89692

r? @joshtriplett

### DIFF
--- a/library/alloc/src/sync.rs
+++ b/library/alloc/src/sync.rs
@@ -804,6 +804,7 @@ impl<T: ?Sized> Arc<T> {
     /// let x_ptr = Arc::into_raw(x);
     /// assert_eq!(unsafe { &*x_ptr }, "hello");
     /// ```
+    #[must_use = "losing the pointer will leak memory"]
     #[stable(feature = "rc_raw", since = "1.17.0")]
     pub fn into_raw(this: Self) -> *const T {
         let ptr = Self::as_ptr(&this);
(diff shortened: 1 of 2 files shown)


library/alloc/src/sync.rs
```diff
@@ -804,6 +804,7 @@ impl<T: ?Sized> Arc<T> {
     /// let x_ptr = Arc::into_raw(x);
     /// assert_eq!(unsafe { &*x_ptr }, "hello");
     /// ```
+    #[must_use = "losing the pointer will leak memory"]
```
matklad, line 807: This one feels like it could lead to false positives. This is Arc, so you *can* get the data back via another instance.

not sure if it happens in practice though. I *think* I used that once to simulate 
increment_strong_count? Not sure.

PR #90439, opened 2021-10-31
Add JoinHandle::is_running.

m-ou-se:
This adds:
```rust
impl<T> JoinHandle<T> {
    /// Checks if the the associated thread is still running its main function.
    ///
    /// This might return `false` for a brief moment after the thread's main
    /// function has returned, but before the thread itself has stopped running.
    pub fn is_running(&self) -> bool;
}
```
The usual way to check if a background thread is still running is to set some atomic flag at the end of its main function. We already do that, in the form of dropping an Arc which will reduce the reference counter. So we might as well expose that information.

This is useful in applications with a main loop (e.g. a game, gui, control system, ..) where you spawn some background task, and check every frame/iteration whether the background task is finished to .join() it in that frame/iteration while keeping the program responsive.

### DIFF
--- a/library/std/src/thread/tests.rs
+++ b/library/std/src/thread/tests.rs
@@ -2,9 +2,13 @@ use super::Builder;
 use crate::any::Any;
 use crate::mem;
 use crate::result;
-use crate::sync::mpsc::{channel, Sender};
+use crate::sync::{
+    mpsc::{channel, Sender},
+    Arc, Barrier,
+};
 use crate::thread::{self, ThreadId};
 use crate::time::Duration;
+use crate::time::Instant;
 
 // !!! These tests are dangerous. If something is buggy, they will hang, !!!
 // !!! instead of exiting cleanly. This might wedge the buildbots.       !!!
@@ -44,6 +48,36 @@ fn test_run_basic() {
         tx.send(()).unwrap();
     });
     rx.recv().unwrap();
+}
+
+#[test]
+fn test_is_running() {
+    let b = Arc::new(Barrier::new(2));
+    let t = thread::spawn({
+        let b = b.clone();
+        move || {
+            b.wait();
+            1234
+        }
+    });
+
+    // Thread is definitely running here, since it's still waiting for the barrier.
+    assert_eq!(t.is_running(), true);
+
+    // Unblock the barrier.
+    b.wait();
+
+    // Now check that t.is_running() becomes false within a reasonable time.
+    let start = Instant::now();
+    while t.is_running() {
+        assert!(start.elapsed() < Duration::from_secs(2));
+        thread::sleep(Duration::from_millis(15));
+    }
+
+    // Joining the thread should not block for a significant time now.
+    let join_time = Instant::now();
+    assert_eq!(t.join().unwrap(), 1234);
+    assert!(join_time.elapsed() < Duration::from_secs(2));
 }
 
 #[test]
--- a/library/std/src/thread/mod.rs
+++ b/library/std/src/thread/mod.rs
@@ -1407,7 +1407,7 @@ impl<T> JoinHandle<T> {
     ///
     /// This might return `false` for a brief moment after the thread's main
     /// function has returned, but before the thread itself has stopped running.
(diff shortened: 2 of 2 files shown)


Mark-Simulacrum:
> This is useful in applications with a main loop (e.g. a game, gui, control system, ..) where you spawn some background task, and check every frame/iteration whether the background task is finished to .join() it in that frame/iteration while keeping the program responsive.

It feels like instead of a is_running() if, it might be better to add a try_join, like we have https://doc.rust-lang.org/nightly/std/process/struct.Child.html#method.try_wait? That would avoid both essentially checking twice and potentially "missing" the result in some sense since it combines the check with actually getting the results. The described use case at least sounds like it would benefit from try_join rather than is_running, though it's likely to be a small difference in practice.

m-ou-se:
> That would avoid both essentially checking twice and potentially "missing" the result in some sense since it combines the check with actually getting the results

.join() doesn't do any checks that .is_running() does (it doesn't even check the refcount), so there's no duplicate check that can be avoided. And there's no way to "miss" the result, since the only way to get the result is to .join() which consumes the (only, non-clonable) JoinHandle.

I could add a try_join() that just does `if !is_running() { join() }` for convenience, but that hides the fact that it might still block for a brief moment while the thread is still running the implementation's cleanup code. Not sure how important that is.

It can still be useful to have `.is_running()` too, to be able to check if the works is done without having to use the result directly.

m-ou-se:
The signature of try_join() would also be an interesting question, since it has to consume Self when it joins, bit leave it alone when it doesn't join.

the8472:
> I could add a try_join() that just does `if !is_running() { join() }` for convenience, but that hides the fact that it might still block for a brief moment while the thread is still running the implementation's cleanup code. Not sure how important that is.

Wouldn't a `Future`-returning join be more useful then which can be polled for the Result (instead of polling `is_running`) without blocking on the thread cleanup?

m-ou-se:
> Wouldn't a Future-returning join be more useful then which can be polled for the Result (instead of polling is_running)

Then as a user I'd need to provide a waker/context to be able to poll the thread.

> without blocking on the thread cleanup?

You're referring to the code that runs on the thread after its main function returns? There's no portable way to check if a thread is really 'done'*. E.g. pthread doesn't provide anything for that. The only way would be to detach the thread and never join it at all.

(*For various definitions of 'done'.)

Mark-Simulacrum:
Hm. OK. I think we could do something like a try_take with basically https://doc.rust-lang.org/nightly/std/sync/struct.Arc.html#method.get_mut, which would let us try_get_result or something. I think this implementation is reasonable enough for now, and we can consider alternatives as part of stabilization; it's not clear to me that there's anything too much better given the limitations of the underlying interfaces. It's definitely the case that we expect a *very* short wait after is_running is false in general -- it should just be the system cleanup for the most part. We could in theory pull out the result value of the thread within the is_running if strong_count = 1, but the user (or we) would still need to call the inner join() in order to not leak the thread resources, so it's not trivially done I suspect. Maybe something to consider in the future.

r=me with the tracking issue filled in

the8472:
> We could in theory pull out the result value of the thread within the is_running if strong_count = 1, but the user (or we) would still need to call the inner join() in order to not leak the thread resources

Wouldn't detaching the thread after pulling out the result work on platforms that don't have a non-blocking join?

library/std/src/thread/tests.rs
```diff
@@ -46,6 +50,36 @@ fn test_run_basic() {
...
+    let start = Instant::now();
+    while t.is_running() {
+        assert!(start.elapsed() < Duration::from_secs(2));
+        thread::sleep(Duration::from_millis(15));
+    }
+
+    // Joining the thread should not block for a significant time now.
+    let join_time = Instant::now();
+    assert_eq!(t.join().unwrap(), 1234);
+    assert!(join_time.elapsed() < Duration::from_secs(2));
```
lygstate, line 80: Does this relaible? Maybe Duration::from_secs(2 + 1.0) better?

m-ou-se: Why would three seconds be more reliable than two?

PR #90440, opened 2021-10-31
[Experiment] Test rustc perf if compiled for x86-64-v3

hkratz:
This can of course not be committed but it might be interesting too see how much perf would be gained (or lost) from AVX2 auto-vectorization and other [x86-64-v3 micro-architecture](https://en.wikipedia.org/wiki/X86-64#Microarchitecture_levels) optimizations.

r? @ghost


the8472:
looks like a network error

@bors retry

hkratz:
> looks like a network error

It's a cert verification failure due to another updated cert whose ca is no recognized by the ancient Debian image.

the8472:
Oh.

@bors cancel

PR #90442, opened 2021-10-31
Windows thread-local keyless drop

ChrisDenton:
`#[thread_local]` allows us to maintain a per-thread list of destructors. This also avoids the need to synchronize global data (which is particularly tricky within the TLS callback function).

r? @alexcrichton

### DIFF
--- a/library/std/src/sys/windows/thread_local_dtor.rs
+++ b/library/std/src/sys/windows/thread_local_dtor.rs
@@ -1,4 +1,39 @@
+//! Implements thread-local destructors that are not associated with any
+//! particular data.
+
 #![unstable(feature = "thread_local_internals", issue = "none")]
 #![cfg(target_thread_local)]
+use super::c;
 
-pub use crate::sys_common::thread_local_dtor::register_dtor_fallback as register_dtor;
+// Using a per-thread list avoids the problems in synchronizing global state.
+#[thread_local]
+static mut DESTRUCTORS: Vec<(*mut u8, unsafe extern "C" fn(*mut u8))> = Vec::new();
+
+pub unsafe fn register_dtor(t: *mut u8, dtor: unsafe extern "C" fn(*mut u8)) {
+    DESTRUCTORS.push((t, dtor));
+}
+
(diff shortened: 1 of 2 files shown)


library/std/src/sys/windows/thread_local_dtor.rs
```diff
@@ -1,4 +1,39 @@
...
+
+pub unsafe fn register_dtor(t: *mut u8, dtor: unsafe extern "C" fn(*mut u8)) {
+    DESTRUCTORS.push((t, dtor));
+}
+
+// See windows/thread_local_keys.rs for an explanation of this callback function.
+// The short version is that all the function pointers in the `.CRT$XL*` array
+// will be called whenever a thread or process starts or ends.
+
+#[link_section = ".CRT$XLD"]
```
alexcrichton, line 20: This all seems pretty reasonable to me! Could the previously-registered "at thread exit" function get a new hook to call into this module though? I think it 'd probably be best to avoid these `#[link_section]` annotations if we can and have them consolidated into one place if possible.

library/std/src/sys/windows/thread_local_dtor.rs
```diff
@@ -1,4 +1,39 @@
...
+        //
+        // Note: While this is potentially an infinite loop, it *should* be
+        // the case that this loop always terminates because we provide the
+        // guarantee that a TLS key cannot be set after it is flagged for
+        // destruction.
+        while let Some((ptr, dtor)) = DESTRUCTORS.pop() {
+            (dtor)(ptr);
+        }
+        // We're done so free the memory.
+        DESTRUCTORS.shrink_to_fit();
```
alexcrichton, line 37: I think this might be a bit clearer as `DESTRUCTORS = Vec::new()` perhaps?

library/std/src/sys/windows/thread_local_dtor.rs
ChrisDenton: Ok, I've used a normal Rust function for running the new dtors and simply made the exit function call it. Is this what you had in mind?

alexcrichton:
@bors: r+

Looks great to me!

PR #90443, opened 2021-10-31
Merge `DocContext.{ty,lt,ct}_substs` into one map

camelid:
It should be impossible to have more than one entry with a particular
key across the three maps, so they should be one map. In addition to
making it impossible for multiple entries to exist, this should improve
memory usage since now only one map is allocated on the stack and heap.

r? @GuillaumeGomez

### DIFF
--- a/src/librustdoc/clean/mod.rs
+++ b/src/librustdoc/clean/mod.rs
@@ -1148,7 +1148,6 @@ impl Clean<Item> for ty::AssocItem {
 }
 
 fn clean_qpath(hir_ty: &hir::Ty<'_>, cx: &mut DocContext<'_>) -> Type {
-    use rustc_hir::GenericParamCount;
     let hir::Ty { hir_id: _, span, ref kind } = *hir_ty;
     let qpath = match kind {
         hir::TyKind::Path(qpath) => qpath,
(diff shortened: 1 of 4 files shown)


camelid:
It's probably best to review this commit-by-commit, and potentially with whitespace hidden.

src/librustdoc/clean/mod.rs
```diff
@@ -222,7 +222,7 @@ impl Clean<Lifetime> for hir::Lifetime {
             | rl::Region::Free(_, node_id),
         ) = def
         {
-            if let Some(lt) = cx.lt_substs.get(&node_id).cloned() {
+            if let Some(lt) = cx.substs.get(&node_id).and_then(|p| p.as_lt()).cloned() {
```
camelid, line 225: I could've made this `unwrap` the Option returned by `as_lt` since it's a rustdoc bug for it to be None here, but it seemed simpler to just use `and_then`. Let me know if you want me to use `unwrap` as a sort of assertion. (Likewise with `as_ty` below.)

camelid:
Hmm, somehow I left trailing whitespace...

src/librustdoc/clean/mod.rs
GuillaumeGomez: I guess it's fine... Maybe add a FIXME on it so we can come back to it later on?

GuillaumeGomez:
There doesn't seem to have any memory usage to be displayed in the performance comparison... Kinda hard to judge if your PR has an impact on it haha.

GuillaumeGomez:
Otherwise the code looks good to me, but I'd be really interested about the RSS usage change.

camelid:
@GuillaumeGomez There's a max-rss comparison on the page rust-timer links to: https://perf.rust-lang.org/compare.html?start=68b554e6af18726fe6fa8de2134c59c441e0b019&end=57ef206d81386eab59b1c32b9e291ae1010c7f8e&stat=max-rss. There are some regressions for whatever reason, but it says they are all insignificant.

src/librustdoc/clean/mod.rs
camelid: Let's either leave it as-is or switch it to `unwrap`. I don't think it's worth a FIXME either way, and it's not actionable.

GuillaumeGomez:
Oh right, I forgot it was "hidden". So that still seems surprising that the memory usage grows in here...

camelid:
> Oh right, I forgot it was "hidden". So that still seems surprising that the memory usage grows in here...

I wonder if it's because an enum is being stored, so there's an extra tag? We should probably find a way to get rid of these tables at some point too.

GuillaumeGomez (approved):

GuillaumeGomez:
Looks good to me, thanks! We can look for a way to remove the tables in another PR.

@bors: r+

camelid:
The instruction count regressions that were reported are spurious—this change doesn't touch the compiler.

Good news: no significant max-rss changes, so the regression reported previously was probably spurious.

camelid:
A bunch of very small instruction count improvements for doc builds though!

image

rylev:
I'm adding back the perf-regression as its currently our only way of tracking interesting perf related cases. Don't worry about actually looking into this. The perf team will want to investigate why this change to rustdoc led to fairly significant changes in non-noisy test cases. We should not be spurious changes that are this large. 

@rustbot label: +perf-regression

cc @Mark-Simulacrum

PR #90446, opened 2021-10-31
Lint elided lifetimes in path during lifetime resolution.

### LINKED ISSUE #55768: 2018 idioms: duplicate suggestion for anonymous lifetime
First [reported here](https://github.com/rust-lang/cargo/issues/6273#issuecomment-436684533) this code:

```rust
#![warn(rust_2018_idioms)]

macro_rules! foo {
    ($name:ident) => {
        pub struct $name;
        impl ::std::fmt::Debug for $name {
            fn fmt(&self, _fmt: &mut ::std::fmt::Formatter) -> Result<(), ::std::fmt::Error> {
                Ok(())
            }
        }
    }
}

foo!(One);
foo!(Two);

fn main() {}
```

[when compiled](https://play.rust-lang.org/?version=beta&mode=debug&edition=2018&gist=307ad8068feba5ab5a4591fdf0de889f) yields two warnings to insert anonymous lifetimes. While we as humans can figure this out pretty well, Cargo as a merciless program can't figure this out and `cargo fix` attempts to apply the fix twice, creating invalid `<'_><'_>` syntax.

It's not clear whether it's best to fix this in `cargo fix` or fix this in the compiler itself. I figure it's easiest to track all idiom-lint related issues in this repository though, so I've opted to move it over here.

### LINKED ISSUE #60199: Multiple anonymous lifetimes bounds upset compiler
TL;DR Rust is unable to compile its own hint with anonymous lifetimes
([Playground](https://play.rust-lang.org/?version=nightly&mode=debug&edition=2018&gist=a683984600a6e012b7d0e507f1b4489a))

```rust
#![warn(rust_2018_idioms)]

use clap::{App, SubCommand};

fn app() -> App {
    SubCommand::with_name("app")
}

fn main () {
    let _matches = app().get_matches();
}
```

Rust doesn't like it as `App` is defined with two lifetimes - `App<'a, 'b>`
and indeed the hint is dead on:
```
warning: hidden lifetime parameters in types are deprecated
 --> src/lib.rs:5:13
  |
5 | fn app() -> App {
  |             ^^^- help: indicate the anonymous lifetimes: `<'_, '_>`
  |
note: lint level defined here
 --> src/lib.rs:1:9
  |
1 | #![warn(rust_2018_idioms)]
  |         ^^^^^^^^^^^^^^^^
  = note: #[warn(elided_lifetimes_in_paths)] implied by #[warn(rust_2018_idioms)]
```

So we do as suggested and add **two** anonymous lifetimes to `App<'_, '_>`
```
#![warn(rust_2018_idioms)]

use clap::{App, SubCommand};

fn app() -> App<'_, '_> {
    SubCommand::with_name("app")
}

fn main () {
    let _matches = app().get_matches();
}
```

This time it is even worth as now compiler coudn't tell these two lifetimes apart

```
error[E0106]: missing lifetime specifiers
 --> src/lib.rs:5:17
  |
5 | fn app() -> App<'_, '_> {
  |                 ^^ expected 2 lifetime parameters

```

So it is either anonymous liftime logic should be fixed to work similar to the type hints, or the help message should avoid suggesting incorrect syntax.

### LINKED ISSUE #63110: Redundant E0106 missing lifetime error and rust_2018_idioms missing lifetime lint
While hacking in the compiler I got the following error:

```
error: hidden lifetime parameters in types are deprecated
   --> src/libsyntax_ext/format.rs:113:43
    |
113 |     arg_with_formatting_spans: Vec<(Span, parse::FormatSpec)>,
    |                                           ^^^^^^^^^^^^^^^^^- help: indicate the anonymous lifetime: `<'_>`
    |
note: lint level defined here
   --> src/libsyntax_ext/lib.rs:5:9
    |
5   | #![deny(rust_2018_idioms)]
    |         ^^^^^^^^^^^^^^^^
    = note: #[deny(elided_lifetimes_in_paths)] implied by #[deny(rust_2018_idioms)]

error[E0106]: missing lifetime specifier
   --> src/libsyntax_ext/format.rs:113:43
    |
113 |     arg_with_formatting_spans: Vec<(Span, parse::FormatSpec)>,
    |                                           ^^^^^^^^^^^^^^^^^ expected lifetime parameter
```

The errors are correct (and the lint is even helpful!) but they are 1) redundant and 2) E0106 is not as helpful as the `rust_2018_idioms` lint. We should get them both to parity and hide one if the other has triggered.

### LINKED ISSUE #71957: Can't allow `elided_lifetimes_in_paths` inside a module that denies `rust_2018_idioms`, other lints work
## What happened

This code fails to compile:

```rust
#![deny(rust_2018_idioms)]

mod blah {
    #![allow(rust_2018_idioms)]

    struct Thing<'a>(&'a i32);

    struct Bar<T>(T);

    fn foo(b: Bar<Thing>) {}
}
```

with the error:

```
error: hidden lifetime parameters in types are deprecated
  --> src/lib.rs:10:19
   |
10 |     fn foo(b: Bar<Thing>) {}
   |                   ^^^^^- help: indicate the anonymous lifetime: `<'_>`
   |
  --> src/lib.rs:1:9
   |
1  | #![deny(rust_2018_idioms)]
   |         ^^^^^^^^^^^^^^^^
   = note: `#[deny(elided_lifetimes_in_paths)]` implied by `#[deny(rust_2018_idioms)]`
```

and I expected the inner `allow` to override the outer `deny`.

## Variations

What's weird is that this code, with `unused_variables` instead of `rust_2018_idioms`, compiles (with other warnings) as I expect:

```rust
#![deny(unused_variables)]

mod blah {
    #![allow(unused_variables)]

    struct Thing<'a>(&'a i32);

    struct Bar<T>(T);

    fn foo(b: Bar<Thing>) {}
}
```

(and removing the `allow` causes it to not compile as I would expect)

AND this code compiles (with other warnings) as I expect:

```rust
#![deny(rust_2018_idioms)]

mod blah {
    #![allow(rust_2018_idioms)]
    extern crate rand;
    use rand::Rng;
}
```

so it seems to be something to do with `elided_lifetimes_in_paths` in particular...?

I also tried this, which I would expect to compile but does not:

```rust
#![deny(rust_2018_idioms)]

mod blah {
    #![allow(elided_lifetimes_in_paths)]

    struct Thing<'a>(&'a i32);

    struct Bar<T>(T);

    fn foo(b: Bar<Thing>) {}
}
```

This I would also expect to compile but does not:

```rust
#![deny(elided_lifetimes_in_paths)]

mod blah {
    #![allow(elided_lifetimes_in_paths)]

    struct Thing<'a>(&'a i32);

    struct Bar<T>(T);

    fn foo(b: Bar<Thing>) {}
}
```

I started thinking maybe it had something to do with allow-by-default lints, but this compiles as I would expect:

```rust
#![deny(missing_docs)]
//! hi

/// Hi
pub mod blah {
    #![allow(missing_docs)]

    pub struct Foo {
        pub field: i32,
    }
}
```

Sooo I don't have any real guesses without compiler internals knowledge 🤷‍♀️ 

## Related

It sounds related to this issue: https://github.com/rust-lang/rust/issues/70819
but I decided to file a new issue because that one appears to be about the *same* "scope level" and I think I'm doing *different* scope levels? Please close if this is a dupe!

## Versions

I'm using stable 1.43; I also confirmed that 1.42 and 1.41 have the same behavior so this is not a recent regression. 

I also confirmed that beta 1.44.0-beta.2 and nightly 2020-05-05 f8d394e5184fe3af761e have the same behavior.

## Why this matters

I'm trying to `include!` code generated by a build script from another crate that generates 2015 edition Rust, from within a crate where I'd like to enforce 2018 idioms.

cjgillot:
The lifetime elision lint is known to be brittle and can be redundant with later lifetime resolution errors. This PR aims to remove the redundancy by performing the lint after lifetime resolution.

This PR proposes to carry the information that an elision should be linted against by using a special `LifetimeName`. I am not certain this is the best solution, but it is certainly the easiest.

Fixes https://github.com/rust-lang/rust/issues/60199
Fixes https://github.com/rust-lang/rust/issues/55768
Fixes https://github.com/rust-lang/rust/issues/63110
Fixes https://github.com/rust-lang/rust/issues/71957

### DIFF
--- a/compiler/rustc_ast_lowering/src/lib.rs
+++ b/compiler/rustc_ast_lowering/src/lib.rs
@@ -1928,6 +1928,7 @@ impl<'a, 'hir> LoweringContext<'a, 'hir> {
                 let param_name = match lt.name {
                     hir::LifetimeName::Param(param_name) => param_name,
                     hir::LifetimeName::Implicit
+                    | hir::LifetimeName::ImplicitMissing
                     | hir::LifetimeName::Underscore
                     | hir::LifetimeName::Static => hir::ParamName::Plain(lt.name.ident()),
                     hir::LifetimeName::ImplicitObjectLifetimeDefault => {
@@ -2322,11 +2323,12 @@ impl<'a, 'hir> LoweringContext<'a, 'hir> {
         &'s mut self,
         span: Span,
         count: usize,
+        param_mode: ParamMode,
     ) -> impl Iterator<Item = hir::Lifetime> + Captures<'a> + Captures<'s> + Captures<'hir> {
-        (0..count).map(move |_| self.elided_path_lifetime(span))
+        (0..count).map(move |_| self.elided_path_lifetime(span, param_mode))
     }
 
-    fn elided_path_lifetime(&mut self, span: Span) -> hir::Lifetime {
+    fn elided_path_lifetime(&mut self, span: Span, param_mode: ParamMode) -> hir::Lifetime {
         match self.anonymous_lifetime_mode {
             AnonymousLifetimeMode::CreateParameter => {
                 // We should have emitted E0726 when processing this path above
@@ -2342,7 +2344,12 @@ impl<'a, 'hir> LoweringContext<'a, 'hir> {
             // lifetime. Instead, we simply create an implicit lifetime, which will be checked
             // later, at which point a suitable error will be emitted.
             AnonymousLifetimeMode::PassThrough | AnonymousLifetimeMode::ReportError => {
-                self.new_implicit_lifetime(span)
+                if param_mode == ParamMode::Explicit {
+                    let id = self.resolver.next_node_id();
+                    self.new_named_lifetime(id, span, hir::LifetimeName::ImplicitMissing)
+                } else {
+                    self.new_implicit_lifetime(span)
+                }
             }
         }
     }
@@ -2536,7 +2543,9 @@ fn lifetimes_from_impl_trait_bounds(
 
         fn visit_lifetime(&mut self, lifetime: &'v hir::Lifetime) {
             let name = match lifetime.name {
-                hir::LifetimeName::Implicit | hir::LifetimeName::Underscore => {
+                hir::LifetimeName::Implicit
+                | hir::LifetimeName::ImplicitMissing
+                | hir::LifetimeName::Underscore => {
                     if self.collect_elided_lifetimes {
                         // Use `'_` for both implicit and underscore lifetimes in
                         // `type Foo<'_> = impl SomeTrait<'_>;`.
--- a/compiler/rustc_ast_lowering/src/lib.rs
+++ b/compiler/rustc_ast_lowering/src/lib.rs
@@ -1786,7 +1786,7 @@ impl<'a, 'hir> LoweringContext<'a, 'hir> {
             GenericArg::Lifetime(hir::Lifetime {
                 hir_id: self.next_id(),
                 span: self.lower_span(span),
-                name: hir::LifetimeName::Implicit,
+                name: hir::LifetimeName::Implicit(false),
             })));
         let generic_args = self.arena.alloc_from_iter(generic_args);
 
@@ -1927,8 +1927,7 @@ impl<'a, 'hir> LoweringContext<'a, 'hir> {
                     });
                 let param_name = match lt.name {
                     hir::LifetimeName::Param(param_name) => param_name,
-                    hir::LifetimeName::Implicit
-                    | hir::LifetimeName::ImplicitMissing
+                    hir::LifetimeName::Implicit(_)
                     | hir::LifetimeName::Underscore
                     | hir::LifetimeName::Static => hir::ParamName::Plain(lt.name.ident()),
                     hir::LifetimeName::ImplicitObjectLifetimeDefault => {
@@ -2291,7 +2290,7 @@ impl<'a, 'hir> LoweringContext<'a, 'hir> {
 
             AnonymousLifetimeMode::ReportError => self.new_error_lifetime(None, span),
 
-            AnonymousLifetimeMode::PassThrough => self.new_implicit_lifetime(span),
+            AnonymousLifetimeMode::PassThrough => self.new_implicit_lifetime(span, false),
         }
     }
 
@@ -2344,12 +2343,7 @@ impl<'a, 'hir> LoweringContext<'a, 'hir> {
             // lifetime. Instead, we simply create an implicit lifetime, which will be checked
             // later, at which point a suitable error will be emitted.
             AnonymousLifetimeMode::PassThrough | AnonymousLifetimeMode::ReportError => {
-                if param_mode == ParamMode::Explicit {
-                    let id = self.resolver.next_node_id();
-                    self.new_named_lifetime(id, span, hir::LifetimeName::ImplicitMissing)
-                } else {
-                    self.new_implicit_lifetime(span)
-                }
+                self.new_implicit_lifetime(span, param_mode == ParamMode::Explicit)
             }
         }
     }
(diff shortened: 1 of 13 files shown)


compiler/rustc_hir/src/hir.rs
```diff
@@ -94,6 +94,9 @@ pub enum LifetimeName {
     /// User wrote nothing (e.g., the lifetime in `&u32`).
     Implicit,
 
+    /// User wrote nothing and should have.
+    ImplicitMissing,
```
jackh726, line 98: I feel like this would make more sense as an `Implicit { missing: bool }`

compiler/rustc_ast_lowering/src/lib.rs
```diff
@@ -2351,7 +2353,12 @@ impl<'a, 'hir> LoweringContext<'a, 'hir> {
             // lifetime. Instead, we simply create an implicit lifetime, which will be checked
             // later, at which point a suitable error will be emitted.
             AnonymousLifetimeMode::PassThrough | AnonymousLifetimeMode::ReportError => {
-                self.new_implicit_lifetime(span)
+                if param_mode == ParamMode::Explicit {
+                    let id = self.resolver.next_node_id();
+                    self.new_named_lifetime(id, span, hir::LifetimeName::ImplicitMissing)
+                } else {
+                    self.new_implicit_lifetime(span)
+                }
```
jackh726, line 2361: Probably a nicer way to do this would be to make `new_implicit_lifetime` just take a `missing: bool` arg

compiler/rustc_resolve/src/late/diagnostics.rs
```diff
@@ -1877,6 +1877,116 @@ impl<'tcx> LifetimeContext<'_, 'tcx> {
         err.emit();
     }
 
+    crate fn report_elision_failure(
+        &mut self,
+        db: &mut DiagnosticBuilder<'_>,
+        params: &[ElisionFailureInfo],
+    ) -> bool /* add `'static` lifetime to lifetime list */ {
```
jackh726, line 1884: What is this comment?

jackh726 (changes requested):
Overall looks good. Didn't give this a super close read yet, but had some thoughts.

compiler/rustc_hir/src/hir.rs
cjgillot: Having `missing: bool` increases the storage size of `LifetimeName`.

jackh726: Really? `LifetimeName::Param` already has `ParamName`, so I wouldn't imagine that an `Implicit { missing: bool }` variant would be larger than that.

And if it does...does it matter?

jackh726: @cjgillot can you change this for a perf run?

cjgillot:
@jackh726 the variant with `Implicit(bool)` is here: https://github.com/rust-lang/rust/pull/91271
> Finished benchmarking commit ([3a16e8d](https://github.com/rust-lang/rust/commit/3a16e8d1317082cc536140a5e6f2b312a196a417)): [comparison url](https://perf.rust-lang.org/compare.html?start=6d246f0c8d3063fea86abbb65a824362709541ba&end=3a16e8d1317082cc536140a5e6f2b312a196a417).
> 
> **Summary**: This change led to large relevant mixed results shrug in compiler performance.
> 
>     * Large improvement in instruction counts (up to -3.9% on `incr-unchanged` builds of `deep-vector`)
> 
>     * Moderate regression in instruction counts (up to 1.4% on `incr-unchanged` builds of `wg-grammar`)
>

jackh726:
I'm going to go ahead and say that I prefer #91271. Aside from wg-grammar opt, perf is pretty similar. I would attribute that to as maybe just noise. However, that PR looks much cleaner imo. That said, it does look like the "Probably a nicer way to do this would be to make new_implicit_lifetime just take a missing: bool arg" comment got missed in this PR but is present there.

r=me on the variant PR once fmt done + CI green. Or r=me on this PR with above comment addressed.

PR #90447, opened 2021-10-31
rustdoc: Remove `Crate.primitives`

camelid:
This involves no longer filtering out impls on primitive types that
aren't defined with `#[doc(primitive)]` somewhere in the dependency
graph. But, I don't think that filtering is necessary or really has any
effect.

### DIFF
--- a/src/librustdoc/passes/collect_trait_impls.rs
+++ b/src/librustdoc/passes/collect_trait_impls.rs
@@ -20,8 +20,6 @@ crate fn collect_trait_impls(krate: Crate, cx: &mut DocContext<'_>) -> Crate {
         (synth.fold_crate(krate), synth.impls)
     });
 
-    let prims: FxHashSet<PrimitiveType> = krate.primitives.iter().map(|p| p.1).collect();
-
     let crate_items = {
         let mut coll = ItemCollector::new();
         krate = cx.sess().time("collect_items_for_trait_impls", || coll.fold_crate(krate));
@@ -51,7 +49,7 @@ crate fn collect_trait_impls(krate: Crate, cx: &mut DocContext<'_>) -> Crate {
         }
     }
 
-    let mut cleaner = BadImplStripper { prims, items: crate_items };
+    let mut cleaner = BadImplStripper { items: crate_items };
     let mut type_did_to_deref_target: FxHashMap<DefId, &Type> = FxHashMap::default();
 
     // Follow all `Deref` targets of included items and recursively add them as valid
@@ -62,9 +60,7 @@ crate fn collect_trait_impls(krate: Crate, cx: &mut DocContext<'_>) -> Crate {
     ) {
         if let Some(target) = map.get(&type_did) {
             debug!("add_deref_target: type {:?}, target {:?}", type_did, target);
-            if let Some(target_prim) = target.primitive_type() {
-                cleaner.prims.insert(target_prim);
-            } else if let Some(target_did) = target.def_id_no_primitives() {
+            if let Some(target_did) = target.def_id_no_primitives() {
                 // `impl Deref<Target = S> for S`
                 if target_did == type_did {
                     // Avoid infinite cycles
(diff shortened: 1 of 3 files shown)


jyn514:
> This involves no longer filtering out impls on primitive types that
aren't defined with #[doc(primitive)] somewhere in the dependency
graph

Wait, what does this mean? doc(primitive) can only go on modules, I don't know what it has to do with impls ...

src/librustdoc/passes/collect_trait_impls.rs
```diff
@@ -62,9 +60,7 @@ crate fn collect_trait_impls(krate: Crate, cx: &mut DocContext<'_>) -> Crate {
     ) {
         if let Some(target) = map.get(&type_did) {
             debug!("add_deref_target: type {:?}, target {:?}", type_did, target);
-            if let Some(target_prim) = target.primitive_type() {
-                cleaner.prims.insert(target_prim);
-            } else if let Some(target_did) = target.def_id_no_primitives() {
+            if let Some(target_did) = target.def_id_no_primitives() {
                 // `impl Deref<Target = S> for S`
```
jyn514, line 64: This in particular makes me worried we'll no longer show Deref methods if an item derefs to a primitive type. Can you add a test for that case?

camelid:
> > This involves no longer filtering out impls on primitive types that
> > aren't defined with #[doc(primitive)] somewhere in the dependency
> > graph
> 
> Wait, what does this mean? doc(primitive) can only go on modules, I don't know what it has to do with impls ...

Sorry, perhaps I wasn't clear. Let me explain my understanding of the old and new bheavior with an example. Take this impl:

```rust
impl SomeTrait for u8 { ... }
```

**Before** this PR, that impl would only be kept if `#[doc(primitive = "u8")] mod some_mod {}` existed somewhere in the dependency graph.

**After** this PR, that impl would be unconditionally kept (although the current code might not work correctly; I'll add some tests).

camelid:
Come to think of it, I wonder if `BadImplStripper` even needs to exist. Probably, but I'm checking just to be sure.

camelid:
> Come to think of it, I wonder if `BadImplStripper` even needs to exist. Probably, but I'm checking just to be sure.

It turns out that removing it fixes some latent bugs in rustdoc 😆 

Specifically, the Notable traits tooltip appears in more places, e.g. the definition (not the std re-export) of [`Box::pin_in`](https://doc.rust-lang.org/nightly/alloc/boxed/struct.Box.html#method.pin_in).

| | Before change | After change |
|-|---|---|
| alloc | ❌ | ✅ |
| std | ✅  | ✅  |

There might be regressions mixed into the results too, but this is promising!

camelid:
> There might be regressions mixed into the results too, but this is promising!

So far, the only changes I've found make the `alloc` docs consistent with the `std` re-exported docs!

camelid:
@bors rollup=never

(This may affect perf.)

GuillaumeGomez:
> Come to think of it, I wonder if `BadImplStripper` even needs to exist. Probably, but I'm checking just to be sure.

Running rustdoc on `proc_macro` panicked for me when I removed it. Also, it allows to have less implementations to keep around. I wonder if perf won't be impacted (negatively) by such a change...

camelid:
> > Come to think of it, I wonder if `BadImplStripper` even needs to exist. Probably, but I'm checking just to be sure.
> 
> Running rustdoc on `proc_macro` panicked for me when I removed it. Also, it allows to have less implementations to keep around. I wonder if perf won't be impacted (negatively) by such a change...

I'll try it on `proc_macro`, that's unfortunate. Re perf: it's better for rustdoc to be consistent than to erroneously strip out impls. Also, perf may be somewhat neutral because there will be more impls, but rustdoc won't be spending time filtering them in the hot `collect_trait_impls` loop.

GuillaumeGomez:
In theory, we don't care about most of the foreign impls. The `BadImplStripper` is there to remove what we don't need.

camelid:
> In theory, we don't care about most of the foreign impls. The `BadImplStripper` is there to remove what we don't need.

But in practice, the `BadImplStripper` only seems to make `std` and `alloc`'s docs inconsistent, like by excluding traits from the "Notable traits" dialog that should actually be there.

GuillaumeGomez:
Well, good luck in removing it then. :wink: 

Im really interested in the perf change if any.

camelid:
> Well, good luck in removing it then. 😉
> 
> Im really interested in the perf change if any.

The hard part of removing it is making sure I don't break anything, but if anything, this will be including *more* information in the docs, so it shouldn't be possible to have a serious bug (modulo panics).

jyn514:
Marking this as blocked on https://github.com/rust-lang/rust/issues/83761 due to the panic.

(@camelid mentioned to me it may be possible to fix this specific panic by fixing the MCVE in #90489, but I'm not confident that's true.)

JohnCSimon:
@camelid 
Thanks for your contribution.

@rustbot label: +S-inactive

PR #90451, opened 2021-10-31
Decouple rustc_resolve and rustc_ast_lowering

cjgillot:
~Based on https://github.com/rust-lang/rust/pull/89090 and https://github.com/rust-lang/rust/pull/90446.~
Part of https://github.com/rust-lang/rust/pull/88186.

This PR removes the `ResolverAstLowering` trait which existed to pass information from resolution to lowering. The required information is passed through the `ResolverOutputs` struct.

r? @michaelwoerister


compiler/rustc_hir/src/hir.rs
```diff
@@ -94,6 +94,9 @@ pub enum LifetimeName {
     /// User wrote nothing (e.g., the lifetime in `&u32`).
     Implicit,
 
+    /// User wrote nothing and should have.
```
pierwill, line 97:
```suggestion
    /// User wrote nothing, but should have provided something.
```

compiler/rustc_middle/src/ty/mod.rs
```diff
@@ -138,6 +137,23 @@ pub struct ResolverOutputs {
     /// Mapping from ident span to path span for paths that don't exist as written, but that
     /// exist under `std`. For example, wrote `str::from_utf8` instead of `std::str::from_utf8`.
     pub confused_type_with_std_module: FxHashMap<Span, Span>,
+    pub item_generics_num_lifetimes: FxHashMap<LocalDefId, usize>,
+
+    pub legacy_const_generic_args: FxHashMap<DefId, Option<Vec<usize>>>,
+
+    /// Resolutions for nodes that have a single resolution.
+    pub partial_res_map: NodeMap<hir::def::PartialRes>,
+    /// Resolutions for import nodes, which have multiple resolutions in different namespaces.
```
pierwill, line 146: Do they *necessarily* have resolutions in multiple namespaces?

```suggestion
    /// Resolutions for import nodes, which can have multiple resolutions in different namespaces.
```

michaelwoerister:
I think @petrochenkov might be more qualified to review this.

compiler/rustc_typeck/src/astconv/mod.rs
```diff
@@ -2289,13 +2291,19 @@ impl<'o, 'tcx> dyn AstConv<'tcx> + 'o {
...
+    /// internal notion of a type.  This is meant to be used within a path.
+    pub fn ast_ty_to_ty_in_path(&self, ast_ty: &hir::Ty<'_>) -> Ty<'tcx> {
+        self.ast_ty_to_ty_inner(ast_ty, false, true)
     }
 
     /// Turns a `hir::Ty` into a `Ty`. For diagnostics' purposes we keep track of whether trait
     /// objects are borrowed like `&dyn Trait` to avoid emitting redundant errors.
     #[tracing::instrument(level = "debug", skip(self))]
-    fn ast_ty_to_ty_inner(&self, ast_ty: &hir::Ty<'_>, borrowed: bool) -> Ty<'tcx> {
+    fn ast_ty_to_ty_inner(&self, ast_ty: &hir::Ty<'_>, borrowed: bool, in_path: bool) -> Ty<'tcx> {
```
b-naber, line 2306: Can we have different versions of `ast_ty_to_ty_inner` instead of using multiple boolean parameters?

petrochenkov:
Blocked on #89090 and #90446.

petrochenkov:
r=me on https://github.com/rust-lang/rust/pull/90451/commits/5c4f8281f1686a22e61ab9a250edbdc283f9e723, https://github.com/rust-lang/rust/pull/90451/commits/f19a7f1fd5c227619e459538c5a02fe5e618fcb6 and https://github.com/rust-lang/rust/pull/90451/commits/baef1d2e72a88245039452a83cbd13e07e1ad112.

I need more context for other changes, why are they useful?
As for https://github.com/rust-lang/rust/pull/90451/commits/0a78293c26354df096bc798a476166ae4f3223bd, definitions and crate store are resolver outputs at least conceptually, what are the reasons to split them up?
For https://github.com/rust-lang/rust/pull/90451/commits/63e9a5ab190cca6dfdb4a09fef1ffe58ca7ba012 I just need some more detailed design to understand the motivation.

cjgillot:
The context is #88186: the objective is to make lowering a query. For this to happen, we need:
- a way to create new `LocalDefId`s inside the query system, so `Definitions` needs special treatment;
- `CrateStore` is used in a similar way to `Definitions`, so it just made sense to treat them so;
- keep immutable resolver information while `TyCtxt` lives instead of directly using the `Resolver` (63e9a5a).

I agree these two commits do not bring any obvious value right now. I can delay them to another PR if you prefer.

compiler/rustc_ast_lowering/src/lib.rs
```diff
@@ -85,13 +87,17 @@ struct LoweringContext<'a, 'hir: 'a> {
...
+    definitions: &'a mut Definitions,
+    cstore: &'a CrateStoreDyn,
+    resolver: &'a mut ResolverOutputs,
 
     /// HACK(Centril): there is a cyclic dependency between the parser and lowering
     /// if we don't have this function pointer. To avoid that dependency so that
     /// `rustc_middle` is independent of the parser, we use dynamic dispatch here.
     nt_to_tokenstream: NtToTokenstream,
 
+    item_generics_num_lifetimes: fn(&Session, &CrateStoreDyn, DefId) -> usize,
```
petrochenkov, line 99: Maybe it's better to make this a method on the `CrateStore` trait?

compiler/rustc_ast_lowering/src/lib.rs
```diff
@@ -160,48 +166,79 @@ struct LoweringContext<'a, 'hir: 'a> {
...
-    fn def_key(&self, id: DefId) -> DefKey;
-
-    fn def_span(&self, id: LocalDefId) -> Span;
-
-    fn item_generics_num_lifetimes(&self, def: DefId) -> usize;
-
-    fn legacy_const_generic_args(&mut self, expr: &Expr) -> Option<Vec<usize>>;
-
-    /// Obtains resolution for a `NodeId` with a single resolution.
+trait ResolverAstLoweringExt {
```
petrochenkov, line 169: Why does this need to be a trait rather than a set of inherent methods?

petrochenkov:
@cjgillot 
I'd still like to know more details about https://github.com/rust-lang/rust/pull/90451#issuecomment-1001680330 (without reading the whole diff of https://github.com/rust-lang/rust/pull/88186).

The main question is how mutable are `ResolverOutputs`, `Definitions` and `CrateStore`, and how long they live.

I expected `ResolverOutputs` to be immutable, but they now include things like `next_node_id` that *are* mutable.
I thought `Definitions` are separated from `ResolverOutputs` due to their mutability (they are mutable because new `LocalDefIds` need to be created), but that's not true given the previous sentence, `ResolverOutputs` are mutable too.
It's still not clear how mutable `CrateStore` is and why it's separated.

It also doesn't feel right that a bunch of tables in `ResolverOutputs` are only used during HIR lowering, but live forever now (as long as `tcx`).

Perhaps we need more separate structures created using the next criteria?
"Needed only during lowering **vs** Needed in tcx after lowering" x "Immutable **vs** Mutable only during lowering **vs** Mutable in tcx after lowering"

cjgillot:
The basic idea is to make lowering incremental, so we will have to keep some resolver outputs longer than we currently have to.

The issue you have with mutable vs immutable outputs probably comes from a poor choice of mine when splitting up #88186. If you don't mind, I will close this PR and try to make the splitting more logical.

`Definitions` is considered as an interner, so will be mutable, behind a private `RwLock`.
`ResolverOutputs` and `CrateStore` will be immutable. `next_node_id` will be the highest `NodeId` given by the resolver, and lowering will manage subsequent `NodeId`s. `trait_map` will not be modified, HIR will just keep a reference to it.

PR #90460, opened 2021-11-01
[stable] Fix CVE-2021-42574 and prepare Rust 1.56.1

emilyalbini:
This PR implements new lints to mitigate the impact of [CVE-2021-42574], caused by the presence of bidirectional-override Unicode codepoints in the compiled source code. [See the advisory][advisory] for more information about the vulnerability.

The changes in this PR will be released later today as part of Rust 1.56.1.

[CVE-2021-42574]: https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2021-42574
[advisory]: https://blog.rust-lang.org/2021/11/01/cve-2021-42574.html


pietroalbini:
The fix was developed by @estebank and reviewed by @nikomatsakis out of band.

@bors r=nikomatsakis p=1000 rollup=never

pietroalbini:
Disabled the broken test, as it doesn't affect stable (it's for an unstable feature, no_core).

@bors r=nikomatsakis,pietroalbini p=1000 rollup=never

compiler/rustc_lint/src/hidden_unicode_codepoints.rs
```diff
@@ -0,0 +1,161 @@
+use crate::{EarlyContext, EarlyLintPass, LintContext};
```
bjorn3, line 1: Github says this file contains bidir control characters. This will break once bootstrap gets this lint, right?

bjorn3: Would `#[doc = "...\u{...}..."] work to avoid directly writing the bidir control characters?

pietroalbini: Uh, that's a *really* good point, cc @estebank. Good catch!

For the stable and beta patches I'd keep it this way, mostly because this is the patch we shared under embargo with Linux distributions ahead of time and I don't want us to apply a different patch. I'll update the master PR as soon as possible.

pietroalbini: Ok this is a bit harder than I thought since the lint docs generator tool doesn't understand `#[doc = ""]`. I'll merge the PR to master without that change as I want nightly to be fixed ASAP, we'll take care of this in a future PR.

PR #90462, opened 2021-11-01
[master] Fix CVE-2021-42574

emilyalbini:
This PR implements new lints to mitigate the impact of [CVE-2021-42574], caused by the presence of bidirectional-override Unicode codepoints in the compiled source code. [See the advisory][advisory] for more information about the vulnerability.

The changes in this PR will be released in tomorrow's nightly release.

[CVE-2021-42574]: https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2021-42574
[advisory]: https://blog.rust-lang.org/2021/11/01/cve-2021-42574.html

### DIFF
--- a/compiler/rustc_lint/src/hidden_unicode_codepoints.rs
+++ b/compiler/rustc_lint/src/hidden_unicode_codepoints.rs
@@ -1,0 +1,161 @@
+use crate::{EarlyContext, EarlyLintPass, LintContext};
+use rustc_ast as ast;
+use rustc_errors::{Applicability, SuggestionStyle};
+use rustc_span::{BytePos, Span, Symbol};
+
+declare_lint! {
+    /// The `text_direction_codepoint_in_literal` lint detects Unicode codepoints that change the
+    /// visual representation of text on screen in a way that does not correspond to their on
+    /// memory representation.
+    ///
+    /// ### Explanation
+    ///
+    /// The unicode characters `\u{202A}`, `\u{202B}`, `\u{202D}`, `\u{202E}`, `\u{2066}`,
+    /// `\u{2067}`, `\u{2068}`, `\u{202C}` and `\u{2069}` make the flow of text on screen change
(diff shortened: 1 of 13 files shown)


pietroalbini:
The fix was developed by @estebank and reviewed by @nikomatsakis out of band.

@bors r=nikomatsakis p=250 rollup=never

pietroalbini:
Fixed formatting.

@bors r=nikomatsakis,pietroalbini p=250 rollup=never

compiler/rustc_lint/src/hidden_unicode_codepoints.rs
```diff
@@ -0,0 +1,161 @@
...
+    /// crafted literals that make the surrounding code seem like it's performing one action, when
+    /// in reality it is performing another. Because of this, we proactively lint against their
+    /// presence to avoid surprises.
+    ///
+    /// ### Example
+    ///
+    /// ```rust,compile_fail
+    /// #![deny(text_direction_codepoint_in_literal)]
+    /// fn main() {
+    ///     println!("{:?}", '‮');
```
est31, line 26: How is this able to compile without an `#[allow]`? I've thought the lint also forbids usage in doc comments? Or are examples exempt?

cuviper: Interesting... The test itself is a `compile_fail`, but you're saying it should fail at the library level?

estebank: @est31 is saying that the lint should trigger when compiling `hidden_unicode_codepoints.rs` because it appears in a doc-comment. Now that it is pointed out, clearly there's something amiss as well here.

est31: It's been pointed out on the main PR as well (should have checked that PR before making the comment): https://github.com/rust-lang/rust/pull/90460#pullrequestreview-793969130

PR #90473, opened 2021-11-01
stabilize format args capture

joshtriplett:
Works as expected, and there are widespread reports of success with it, as well as interest in it.

RFC: rust-lang/rfcs#2795
Tracking issue: https://github.com/rust-lang/rust/issues/67984

Addressing items from the tracking issue:

- We don't support capturing arguments from a non-literal format string like `format_args!(concat!(...))`. We could add that in a future enhancement, or we can decide that it isn't supported (as suggested in https://github.com/rust-lang/rust/issues/67984#issuecomment-801394736 ).
- I've updated the documentation.
- `panic!` now supports capture as well.
- There are potentially opportunities to further improve diagnostics for invalid usage, such as if it looks like the user tried to use an expression rather than a variable. However, such cases are all already caught and provide reasonable syntax errors now, and we can always provided even friendlier diagnostics in the future.


joshtriplett:
@rust-lang/lang, shall we stabilize `println!("{local_variable}")` and family, now that we've shipped Rust 2021 with working `panic!("{local_variable}")` support?

@rfcbot merge

m-ou-se:
> panic! now supports capture as well.

Note that many macros in the ecosytem do *not* support this right now. E.g. `anyhow!("{x}")` will just produce an error with the message `"{x}"`.

joshtriplett:
@m-ou-se True, but I don't think that should be a blocker to stabilization. I *do* think that the release notes for the release stabilizing this should encourage propagating this behavior, though.

m-ou-se:
It might be a good idea to inform the maintainers of popular crates with such a macro now. Once this stabilizes their users will probably expect it to work with their macros too.

cc @dtolnay for `anyhow`.
cc @yaahc for `eyre`.

If I find the time for it, I'll try to hack together some regex to find other crates with a similar macro.

joshtriplett:
cc @hawkw for `tracing`.

nikomatsakis:
Is there a PR adding this to the reference?

hawkw:
> cc @hawkw for `tracing`.

I *think* this ought to work fine with `tracing` out of the box without any work on our end &mdash; when a user adds a `format_args!` message to a tracing macro, we just capture all the tokens as a repeated set of token trees, and pass them directly to a `format_args!` invocation. So, as far as I know, any format args accepted by `format_args!` should be accepted by `tracing` as well without any changes on our part!

nikomatsakis:
@rfcbot reviewed

Josh pointed out in the lang team meeting that this doesn't need to appear in the rust-lang reference, as it's (conceptually, at least) a library feature, and those docs are updated. Huzzah!

scottmcm:
I was going to be scared if this affected tokenization, but because it doesn't take general expressions, 👍

m-ou-se:
> I _think_ this ought to work fine with `tracing` out of the box without any work on our end

Yup, `log` and `tracing` and many others always use `format_args`, so don't require any changes:

```
error: there is no argument named `x`
 --> src/main.rs:2:21
  |
2 |     tracing::info!("{x}");
  |                     ^^^
  |
  = help: if you intended to capture `x` from the surrounding scope, add `#![feature(format_args_capture)]` to the crate attributes
```

The only problem is macros that special-case the single argument case to skip `format_args!()`. E.g. Rust 2015/2018's `panic!("{}")` and Anyhow's `anyhow!("{}")` (etc.) (https://github.com/dtolnay/anyhow/issues/55).

joshtriplett:
Rebased.

joshtriplett:
Rebased.

davidhewitt:
Just wanted to say thank-you to @joshtriplett & @m-ou-se (and everyone else who has input into design and implementation since I originally drafted the RFC). 

I'm really excited to see this landing - I stumbled on this idea when I was trying to find a way to explain Rust's string formatting macros to beginners, and I think this syntax will be a nice way to simplify how we can introduce users to their first `println!` invocation.

```
fn main() {
    let target = "World";
    println!("Hello, {target}");
}
```

... just seems a little easier to beginners to have intuition about than `println!("Hello, {}", target)`.

 On top of that, I'm definitely planning to use this a lot in my own code 😂

joshtriplett:
Rebased. I expect this to continue conflicting regularly, anytime another feature gate is accepted. Now that the FCP has ended, could I get an r+ please?

PR #90475, opened 2021-11-01
rustdoc: Add `DocVisitor` and use it where possible

camelid:
`DocFolder` allows transforming the docs, accomplished by making its methods take and return types by-value. However, several of the rustdoc `DocFolder` impls only *visit* the docs; they don't change anything. Passing around types by-value is thus unnecessary, confusing, and potentially inefficient for those impls.

`DocVisitor` is very similar to `DocFolder`, except that its methods take shared references and return nothing (i.e., the unit type). This should both be more efficient and make the code clearer.

There is an additional reason to add `DocVisitor`, too. As part of my cleanup of `external_traits`, I'm planning to add a `fn cache(&mut self) -> &mut Cache` method to `DocFolder` so that `external_traits` can be retrieved explicitly from the `Cache`, rather than implicitly via `Crate.external_traits` (which is an `Rc<RefCell<...>>`). However, some of the `DocFolder` impls that could be turned into `DocVisitor` impls only have a shared reference to the `Cache`, because they are used during rendering. (They have to access the `Cache` via `html::render::Context.shared.cache`, which involves an `Rc`.)

Since `DocVisitor` does not mutate any of the types it's visiting, its equivalent `cache()` method will only need a shared reference to the `Cache`, avoiding the problem described above.

r? @GuillaumeGomez
cc @jyn514

### DIFF
--- a/src/librustdoc/fold.rs
+++ b/src/librustdoc/fold.rs
@@ -84,14 +84,12 @@ crate trait DocFolder: Sized {
     fn fold_crate(&mut self, mut c: Crate) -> Crate {
         c.module = self.fold_item(c.module).unwrap();
 
-        {
-            let external_traits = { std::mem::take(&mut *c.external_traits.borrow_mut()) };
-            for (k, mut v) in external_traits {
-                v.trait_.items =
-                    v.trait_.items.into_iter().filter_map(|i| self.fold_item(i)).collect();
-                c.external_traits.borrow_mut().insert(k, v);
-            }
+        let external_traits = { std::mem::take(&mut *c.external_traits.borrow_mut()) };
+        for (k, mut v) in external_traits {
+            v.trait_.items = v.trait_.items.into_iter().filter_map(|i| self.fold_item(i)).collect();
+            c.external_traits.borrow_mut().insert(k, v);
         }
+
         c
     }
 }
--- a/src/librustdoc/fold.rs
+++ b/src/librustdoc/fold.rs
@@ -60,8 +60,26 @@ crate trait DocFolder: Sized {
                 }
                 Variant::CLike => VariantItem(Variant::CLike),
             },
-            // FIXME: list all cases explicitly
-            x => x,
+            ExternCrateItem { src: _ }
+            | ImportItem(_)
+            | FunctionItem(_)
+            | TypedefItem(_, _)
+            | OpaqueTyItem(_)
+            | StaticItem(_)
+            | ConstantItem(_)
+            | TraitAliasItem(_)
+            | TyMethodItem(_)
+            | MethodItem(_, _)
+            | StructFieldItem(_)
+            | ForeignFunctionItem(_)
+            | ForeignStaticItem(_)
+            | ForeignTypeItem
+            | MacroItem(_)
+            | ProcMacroItem(_)
+            | PrimitiveItem(_)
+            | AssocConstItem(_, _)
+            | AssocTypeItem(_, _)
+            | KeywordItem(_) => kind,
         }
     }
 
--- a/src/librustdoc/fold.rs
+++ b/src/librustdoc/fold.rs
@@ -59,7 +59,7 @@ crate trait DocFolder: Sized {
                         let fields = fields.into_iter().filter_map(|x| self.fold_item(x)).collect();
                         VariantItem(Variant::Tuple(fields))
                     }
-                    i => VariantItem(i),
+                    Variant::CLike => VariantItem(Variant::CLike),
                 }
             }
             x => x,
--- a/src/librustdoc/fold.rs
+++ b/src/librustdoc/fold.rs
@@ -46,22 +46,21 @@ crate trait DocFolder: Sized {
                 i.items = i.items.into_iter().filter_map(|x| self.fold_item(x)).collect();
                 ImplItem(i)
             }
-            VariantItem(i) => {
-                match i {
-                    Variant::Struct(mut j) => {
-                        let num_fields = j.fields.len();
-                        j.fields = j.fields.into_iter().filter_map(|x| self.fold_item(x)).collect();
-                        j.fields_stripped |= num_fields != j.fields.len()
-                            || j.fields.iter().any(|f| f.is_stripped());
-                        VariantItem(Variant::Struct(j))
-                    }
-                    Variant::Tuple(fields) => {
-                        let fields = fields.into_iter().filter_map(|x| self.fold_item(x)).collect();
-                        VariantItem(Variant::Tuple(fields))
-                    }
-                    Variant::CLike => VariantItem(Variant::CLike),
+            VariantItem(i) => match i {
+                Variant::Struct(mut j) => {
+                    let num_fields = j.fields.len();
+                    j.fields = j.fields.into_iter().filter_map(|x| self.fold_item(x)).collect();
+                    j.fields_stripped |=
+                        num_fields != j.fields.len() || j.fields.iter().any(|f| f.is_stripped());
+                    VariantItem(Variant::Struct(j))
                 }
-            }
+                Variant::Tuple(fields) => {
+                    let fields = fields.into_iter().filter_map(|x| self.fold_item(x)).collect();
+                    VariantItem(Variant::Tuple(fields))
+                }
+                Variant::CLike => VariantItem(Variant::CLike),
+            },
+            // FIXME: list all cases explicitly
             x => x,
         }
     }
--- a/src/librustdoc/fold.rs
+++ b/src/librustdoc/fold.rs
@@ -47,7 +47,6 @@ crate trait DocFolder: Sized {
                 ImplItem(i)
             }
             VariantItem(i) => {
-                let i2 = i.clone(); // this clone is small
                 match i {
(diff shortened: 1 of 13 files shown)


src/librustdoc/visit.rs
```diff
@@ -0,0 +1,53 @@
...
+            },
+            // FIXME: list all cases explicitly
+            _ => return,
+        }
+    }
+
+    /// don't override!
+    fn visit_item_recur(&mut self, item: &Item) {
+        match &*item.kind {
+            StrippedItem(i) => self.visit_inner_recur(i),
```
camelid, line 34: This should probably be moved into `visit_inner_recur` instead of having the `unreachable!`. (This is also the case in `DocFolder`.)

camelid:
Hmm, not sure why there's a regression, but the improvements outweigh the regression, and this is a good cleanup regardless. Perhaps I can look into if there's a way to avoid the regression; it's a bit surprising, but it could just be the optimizer behaving differently.

src/librustdoc/passes/collect_trait_impls.rs
```diff
@@ -1,7 +1,6 @@
 use super::Pass;
-use crate::clean::*;
 use crate::core::DocContext;
-use crate::fold::DocFolder;
+use crate::{clean::*, visit::DocVisitor};
```
camelid, line 3: These imports should probably be unmerged (and others like this).

src/librustdoc/visit.rs
```diff
@@ -0,0 +1,53 @@
...
+            UnionItem(i) => i.fields.iter().for_each(|x| self.visit_item(x)),
+            EnumItem(i) => i.variants.iter().for_each(|x| self.visit_item(x)),
+            TraitItem(i) => i.items.iter().for_each(|x| self.visit_item(x)),
+            ImplItem(i) => i.items.iter().for_each(|x| self.visit_item(x)),
+            VariantItem(i) => match i {
+                Variant::Struct(j) => j.fields.iter().for_each(|x| self.visit_item(x)),
+                Variant::Tuple(fields) => fields.iter().for_each(|x| self.visit_item(x)),
+                Variant::CLike => {}
+            },
+            // FIXME: list all cases explicitly
```
GuillaumeGomez, line 26: Since you're adding this code, maybe do it now instead of postponing it. :)

camelid: Yeah, I was planning to. It was a self-todo :)

camelid: Done.

camelid:
@notriddle Guillaume is busy right now so can't do an in-depth review. Would you be interested in reviewing this? You don't have to do a full review, but I'd appreciate some feedback if you have time :)

camelid:
It looks like the largest regression only regressed by 0.01 seconds of wall-clock time.

src/librustdoc/visit.rs
fbstj: is there a reason for `visit_inner_recur` to not be merged into `visit_item_recur` ?

camelid: I don't know of one, but it could be some of the code relies on them being separate.

src/librustdoc/html/sources.rs
```diff
@@ -98,16 +99,20 @@ struct SourceCollector<'a, 'tcx> {
     emitted_local_sources: FxHashSet<PathBuf>,
 }
 
-impl DocFolder for SourceCollector<'_, '_> {
-    fn fold_item(&mut self, item: clean::Item) -> Option<clean::Item> {
+impl DocVisitor for SourceCollector<'_, '_> {
+    fn visit_item(&mut self, item: &clean::Item) {
+        if !self.cx.include_sources {
+            return;
```
notriddle, line 105: This line of code cries out for a more explicit way to do control flow than this. This conditional seems to stop it from trying to recurse into child nodes, but it can't stop the parent from trying to visit all of the siblings (and that seems wasteful), and I really only figured that out by carefully reading the implementation.

The API used by [`TypeVisitor`](https://doc.rust-lang.org/nightly/nightly-rustc/rustc_middle/ty/fold/trait.TypeVisitor.html) seems like a good source for plagiarizing, since `ControlFlow::Break` provides a way to "just bail out" like we want here.

src/librustdoc/visit.rs
```diff
@@ -0,0 +1,71 @@
+use crate::clean::*;
+
+crate trait DocVisitor: Sized {
+    fn visit_item(&mut self, item: &Item) {
+        self.visit_item_recur(item)
+    }
+
+    /// don't override!
```
notriddle, line 8: Like how TypeFoldable provides this method for types, could an extension trait be attached to ItemKind to take care of this?

src/librustdoc/html/sources.rs
```diff
@@ -16,23 +16,25 @@ use std::ffi::OsStr;
 use std::fs;
 use std::path::{Component, Path, PathBuf};
 
-crate fn render(cx: &mut Context<'_>, krate: clean::Crate) -> Result<clean::Crate, Error> {
```
notriddle, line 19: :+1: That's way cleaner. I like this change; the other comments really are just nitpicking, and the PR could easily be merged as-is.

src/librustdoc/fold.rs
```diff
@@ -46,24 +46,40 @@ crate trait DocFolder: Sized {
...
-                        let fields = fields.into_iter().filter_map(|x| self.fold_item(x)).collect();
-                        VariantItem(Variant::Tuple(fields))
-                    }
-                    _ => VariantItem(i2),
+            VariantItem(i) => match i {
+                Variant::Struct(mut j) => {
+                    let num_fields = j.fields.len();
+                    j.fields = j.fields.into_iter().filter_map(|x| self.fold_item(x)).collect();
+                    j.fields_stripped |=
+                        num_fields != j.fields.len() || j.fields.iter().any(|f| f.is_stripped());
```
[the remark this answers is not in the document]
camelid: Probably the reason it doesn't use a conditional is that no one thought of it ;)

It's a good idea; do you want to open a followup PR for it? Honestly, I think we should just remove the `fields_stripped` fields altogether and compute this information on-demand, but that'll take some time.

src/librustdoc/html/sources.rs
camelid: Yeah, `ControlFlow` makes everything better 🙃 

I'd be happy to try out switching `DocVisitor` to use `ControlFlow`, but that'd make this change even bigger, so I think let's save it for a future PR.

src/librustdoc/visit.rs
camelid: Hmm, could you elaborate? I'm not sure why that would be better. I'm generally not a fan of too many extension traits since they add complexity.

camelid:
> 👍 That's way cleaner. I like this change; the other comments really are just nitpicking, and the PR could easily be merged as-is.

@notriddle does that mean it's okay to `r=notriddle`? Or would you like Guillaume to review this too?

And thanks for the review! ❤️

src/librustdoc/fold.rs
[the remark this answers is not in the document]
notriddle: Sure, I'll do that.

src/librustdoc/visit.rs
notriddle: If you don't like the idea, then it's fine, let's not do that. I brought it up because it would enforce the "don't override" rule instead of just making it a comment.

notriddle:
> does that mean it's okay to r=notriddle? Or would you like Guillaume to review this too?

Yeah, that's what it means.

camelid:
Thanks again! @bors r=notriddle rollup=never

src/librustdoc/visit.rs
camelid: I think I see what you mean. We could do this without an extension trait since `ItemKind` is local. But I'm not sure if the extra complexity of not having all the visitor code in one place is worth it. Also, the current way `DocVisitor` (and `DocFolder`) works is consistent with similar visitors in rustc (like `mir::Visitor`). But I'd still be open to us trying it.

src/librustdoc/fold.rs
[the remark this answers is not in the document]
camelid: You can assign me on it if you like.

camelid:
Marking the perf regression as triaged:

- The improvements outweigh the regressions.
- It's possible the regression was spurious or out of our control due to different machine code layout.
- This is a good cleanup; the old code was more complex and harder to understand, while the new version makes it clear that no transformations are being performed.
- This cleanup will allow future performance improvements.

rylev:
Going to mark this as triaged as the regressions are way smaller than the improvements and confined to the `deeply-nested-async` benchmark which has been acting up lately. This is really an improvement and not a mixed result. 🎉

camelid:
> Going to mark this as triaged as the regressions are way smaller than the improvements and confined to the `deeply-nested-async` benchmark which has been acting up lately. This is really an improvement and not a mixed result. 🎉

I agree that this is really an improvement, but AFAICT `deeply-nested-async` didn't have any changes here; only `webrender` and `derive` regressed.

PR #90478, opened 2021-11-01
Use apple-a14 as target CPU for aarch64-apple-darwin

hkratz:
After updating the minimum required LLVM version to 12 (#90175) we can use `apple-a14` as target CPU, because that CPU is similar in features to the Apple M1 (see [LLVM 13 source](https://github.com/llvm/llvm-project/blob/b8016b626ec7095c7d57ebfffb2135dc5c3077b8/llvm/lib/Target/AArch64/AArch64.td#L1127)). Once the minimum required LLVM version is updated to 13 we can use `apple-m1` here.


wesleywiser (approved):
Looks reasonable to me since A14 and M1 use the same ARM revision and the M1 was the first generation of `aarch64-apple-darwin` devices.

PR #90485, opened 2021-11-01
Don't destructure args tuple in format_args!

camsteffen:
This allows Clippy to parse the HIR more simply since `arg0` is changed to `_args.0`. (cc rust-lang/rust-clippy#7843). From rustc's perspective, I think this is something between a lateral move and a tiny improvement since there are fewer bindings.

r? @m-ou-se

### DIFF
--- a/compiler/rustc_builtin_macros/src/format.rs
+++ b/compiler/rustc_builtin_macros/src/format.rs
@@ -760,9 +760,9 @@ impl<'a, 'b> Context<'a, 'b> {
     /// Actually builds the expression which the format_args! block will be
     /// expanded to.
     fn into_expr(self) -> P<ast::Expr> {
-        let mut locals =
-            Vec::with_capacity((0..self.args.len()).map(|i| self.arg_unique_types[i].len()).sum());
-        let mut counts = Vec::with_capacity(self.count_args.len());
+        let mut args = Vec::with_capacity(
+            self.arg_unique_types.iter().map(|v| v.len()).sum::<usize>() + self.count_args.len(),
+        );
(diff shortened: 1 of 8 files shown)


camsteffen:
@m-ou-se I picked you as a reviewer for this and #90131 since you wrote the Clippy issue, but feel free to reassign.

This will conflict with rust-lang/rust-clippy#7906. We probably want to merge and sync that PR first so that I can handle the conflict here.

m-ou-se:
Looks good to me!

One question: Why `_args` and not  just `args`?

camsteffen:
`_args` is an unused binding in `format_args!("just a string")`. We could replace `match () { _args => [] }` with `[]` in that case but that's a different change.

camsteffen:
@flip1995 can I get your approval on that last commit for Clippy?

flip1995 (approved):
Clippy changes LGTM :+1:

rylev:
@camsteffen @m-ou-se this seems to have caused some compiler performance issues. The impacted query seems to be `expand_crate` which I imagine could be impacted by this change. The change to the html5ever crate is fairly large at 21.8x the significance threshold. Given this change is for a tool that not everyone may use (clippy) but has the potential to impact all compiler workloads, we may want to investigate this further. 

That being said, none of the code looked suspicious to me so we may have to run something like cachegrind to really find out. Any ideas before we investigate further?

camsteffen:
No ideas here. The complexity savings on Clippy's end is rather small so it might not justify the regression.

I've noticed a couple times now that the final automated perf run shows significantly worse results than the manual run. Is there a particular reason for that ?

Issue #90309, opened 2021-10-26
Add a rustdoc user guide to https://doc.rust-lang.org/rustdoc


jyn514:
Right now, https://doc.rust-lang.org/rustdoc is mostly developer-oriented, it has lots of info about how to use rustdoc the tool but not how to use the generated docs. We should add a user-focused page to it that mentions options people don't usually know about, e.g.
- `?` for search tips
- the "In Parameters" and "In Return Types" tab of search
- the settings page maybe?

NoraCodes:
Throwing my hat in the ring on this one. Let me know if you'd like me to add some docs for the settings page!

jyn514:
This was added in #90487; thanks @NoraCodes !

PR #90487, opened 2021-11-01
Add a chapter on reading Rustdoc output

NoraCodes:
Includes documentation for:
 - general page structure
 - navigation
 - searching
 - themes
 - deep-linking

Doesn't include docs on the settings page.

Per https://github.com/rust-lang/rust/issues/90309

### DIFF
--- a/src/doc/rustdoc/src/how-to-read-rustdoc.md
+++ b/src/doc/rustdoc/src/how-to-read-rustdoc.md
@@ -1,0 +1,107 @@
+# How to read rustdoc output
+
+Rustdoc's HTML output includes a friendly and useful navigation interface which
+makes it easier for users to navigate and understand your code.
+This chapter covers the major features of that interface,
+and is a great starting point for documentation authors and users alike.
+
+## Structure
+
+The `rustdoc` output is divided into three sections.
+Along the left side of each page is a quick navigation bar,
+which shows contextual information about the current entry.
+The rest of the page is taken up by the search interface at the top
+and the documentation for the current item below that.
+
+## The Item Documentation
+
+The majority of the screen is taken up with the documentation text for the item
+currently being viewed.
+At the top is some at-a-glance info and controls:
+
+- the type and name of the item,
+  such as "Struct `std::time::Duration`",
+- a button to copy the item's path to the clipboard,
+  which is a clipboard item
+- a button to collapse or expand the top-level documentation for that item
+  (`[+]` or `[-]`),
+- a link to the source code (`[src]`),
+  if [configured](the-doc-attribute.html#html_no_source),
+  and present (the source may not be available if
+  the documentation was created with `cargo doc --no-deps`),
+- and the version in which the item became stable,
+  if it's a stable item in the standard library.
+
+Below this is the main documentation for the item,
+including a definition or function signature if appropriate,
+followed by a list of fields or variants for Rust types.
+Finally, the page lists associated functions and trait implementations,
+including automatic and blanket implementations that `rustdoc` knows about.
+
+### Navigation
+
+Subheadings, variants, fields, and many other things in this documentation
+are anchors and can be clicked on and deep-linked to,
+which is a great way to communicate exactly what you're talking about.
+The typograpical character "§" appears next to lines with anchors on them
+when hovered or given keyboard focus.
+
+## The Navigation Bar
+
+For example, when looking at documentation for the crate root,
+it shows all the crates documented in the documentation bundle,
+and quick links to the modules, structs, traits, functions, and macros available
+from the current crate.
+At the top, it displays a [configurable logo](the-doc-attribute.html#html_logo_url)
+alongside the current crate's name and version,
+or the current item whose documentation is being displayed.
+
+## The Theme Picker and Search Interface
+
+When viewing `rustdoc`'s output in a browser with JavaScript enabled,
+a dynamic interface appears at the top of the page.
+To the left is the theme picker, denoted with a paint-brush icon,
+and the search interface, help screen, and options appear to the right of that.
+
+### The Theme Picker
+
+Clicking on the theme picker provides a list of themes -
+by default `ayu`, `light`, and `dark` -
+which are available for viewing.
+
+### The Search Interface
+
+Typing in the search bar instantly searches the available documentation for
+the string entered with a fuzzy matching algorithm that is tolerant of minor
+typos.
+
+By default, the search results give are "In Names",
+meaning that the fuzzy match is made against the names of items.
+Matching names are shown on the left, and the first few words of their
+descriptions are given on the right.
+By clicking an item, you will navigate to its particular documentation.
+
+There are two other sets of results, shown as tabs in the search results pane.
+"In Parameters" shows matches for the string in the types of parameters to
+functions, and "In Return Types" shows matches in the return types of functions.
+Both are very useful when looking for a function whose name you can't quite
+bring to mind when you know the type you have or want.
+
+When typing in the search bar, you can prefix your search term with a type
+followed by a colon (such as `mod:`) to restrict the results to just that
+kind of item. (The available items are listed in the help popup.)
+
+### Shortcuts
+
+Pressing `S` while focused elsewhere on the page will move focus to the
+search bar, and pressing `?` shows the help screen,
+which includes all these shortcuts and more.
+Pressing `T` focuses the theme picker.
+
+When the search results are focused,
+the left and right arrows move between tabs and the up and down arrows move
+among the results.
+Pressing the enter or return key opens the highlighted result.
+
+When looking at the documentation for an item, the plus and minus keys expand
+and collapse all sections in the document.
--- a/src/doc/rustdoc/src/SUMMARY.md
+++ b/src/doc/rustdoc/src/SUMMARY.md
@@ -1,6 +1,7 @@
 # The Rustdoc Book
 
 - [What is rustdoc?](what-is-rustdoc.md)
+- [How to read rustdoc output](how-to-read-rustdoc.md)
 - [How to write documentation](how-to-write-documentation.md)
 - [What to include (and exclude)](what-to-include.md)
 - [Command-line arguments](command-line-arguments.md)


src/doc/rustdoc/src/how-to-read-rustdoc.md
```diff
@@ -0,0 +1,101 @@
...
+to make it easier for users to navigate and understand your code.
+This chapter covers the major features of that interface,
+and is a great starting point for documentation authors and users alike.
+
+## Structure
+
+The `rustdoc` output is divided into three sections.
+Along the left side of each page is a quick navigation bar,
+which shows contextual information about the current entry.
+The rest of the is taken up by the search interface at the top
```
GuillaumeGomez, line 13: "of the **page**" I guess?

src/doc/rustdoc/src/how-to-read-rustdoc.md
```diff
@@ -0,0 +1,101 @@
...
+and the documentation for the current item below that.
+
+## The Item Documentation
+
+The majority of the screen is taken up with the main event:
```
GuillaumeGomez, line 18: "the main event"?

src/doc/rustdoc/src/how-to-read-rustdoc.md
NoraCodes: LOL, yes, thanks!

src/doc/rustdoc/src/how-to-read-rustdoc.md
```diff
@@ -0,0 +1,101 @@
...
+the documentation text for the item currently being viewed.
+The main title is the type and name of the item,
+such as "Struct `std::time::Duration`",
+along with a button to collapse or expand the top-level documentation for that
+item (`[+]` or `[-]`),
```
GuillaumeGomez, line 23: @jsha If we merge this PR with this doc, we'll need to not forget to update it in case we change the collapse/expand buttons.

src/doc/rustdoc/src/how-to-read-rustdoc.md
NoraCodes: English (American?) idiom meaning "the point" - think what data is to metadata. If it confused you, though, it'll confuse others, so I'll change it.

src/doc/rustdoc/src/how-to-read-rustdoc.md
```diff
@@ -0,0 +1,101 @@
...
+a link to the source code (`[src]`),
+if [configured](the-doc-attribute.html#html_no_source),
+and the version in which the item became stable,
+if it's a stable item in the standard library.
+
+Below this is the main documentation for the item,
+including a definition or function signature if appropriate,
+followed by a list of fields or variants for structs, unions, and enums.
```
GuillaumeGomez, line 31: Instead of listing the different types, wouldn't it be better to simply say "for rust types"?

NoraCodes: Seems reasonable!

src/doc/rustdoc/src/how-to-read-rustdoc.md
```diff
@@ -0,0 +1,101 @@
...
+
+Clicking on the theme picker provides a list of themes -
+by default `ayu`, `light`, and `dark` - 
+which are available for viewing.
+
+### The Search Interface
+
+Typing in the search bar instantly searches the available documentation for
+the string entered, with a fuzzy matching algorithm that is tolerant of omitted
+letters but not extra ones.
```
GuillaumeGomez, line 70: If I look for "stringa" in the std docs, it returns `String`. So simply saying "is tolerant to a given amount of typos" or something along the line is fine I guess.

src/doc/rustdoc/src/how-to-read-rustdoc.md
```diff
@@ -0,0 +1,101 @@
...
+
+By default, the search results give are "In Names", meaning that the fuzzy
+match is made against the names of items.
+Matching names are shown on the left, and the first few words of their
+descriptions are given on the right.
+By clicking an item, you will navigate to its particular documentation.
+
+There are two other modes, shown as tabs in the search results pane.
+"In Parameters" searches for the string in the types of parameters to functions,
```
GuillaumeGomez, line 79: It doesn't search when clicking on another tab, it's simply how they're classified.

src/doc/rustdoc/src/how-to-read-rustdoc.md
```diff
@@ -0,0 +1,101 @@
...
+and "In Return Types" searches in the return types of functions.
```
GuillaumeGomez, line 80: Same here.

src/doc/rustdoc/src/how-to-read-rustdoc.md
```diff
@@ -0,0 +1,101 @@
...
+When typing in the search bar, you can prefix your search term with a type
+followed by a colon (such as `mod:`) to restrict the results to just that
+kind of item - one of `fn`, `mod`, `struct`, `enum`, `trait`, `type`, `macro`,
+or `const`.
+
+### Shortcuts
+
+Pressing `S` while focused elsewhere on the page will move focus to the
+search bar, and pressing `?` shows a help screen with many shortcuts.
+Pressing `T` focuses the theme picker.
```
GuillaumeGomez, line 93: Maybe talk about the help popup to see the full list (the one appearing when clicking on the `?` button or pressing the `?` key).

src/doc/rustdoc/src/how-to-read-rustdoc.md
NoraCodes: Ah, important distinction to make. Thanks!

src/doc/rustdoc/src/how-to-read-rustdoc.md
NoraCodes: Should I not include the full list here?

GuillaumeGomez:
So overall, it's really nice. I wonder: could we add screenshots to make it easier for readers to follow? Or would it be too difficult to keep it up-to-date? (cc @rust-lang/rustdoc)

src/doc/rustdoc/src/how-to-read-rustdoc.md
GuillaumeGomez: It's fine to keep them. Maybe add a note that they are listed directly in the help popup too.

NoraCodes:
I considered that, but first of all it would make this page inaccessible, or at least it would be very verbose if we wanted to keep it accessible to people who can't see images for whatever reason - and, yeah, I suspect it would be quite annoying to keep up to date.

src/doc/rustdoc/src/how-to-read-rustdoc.md
NoraCodes: Great! Reduced verbosity.

src/doc/rustdoc/src/how-to-read-rustdoc.md
```diff
@@ -0,0 +1,101 @@
...
+There are two other modes, shown as tabs in the search results pane.
+"In Parameters" searches for the string in the types of parameters to functions,
+and "In Return Types" searches in the return types of functions.
+Both are very useful when looking for a function whose name you can't quite
+bring to mind when you know the type you have or want.
+
+When typing in the search bar, you can prefix your search term with a type
+followed by a colon (such as `mod:`) to restrict the results to just that
+kind of item - one of `fn`, `mod`, `struct`, `enum`, `trait`, `type`, `macro`,
+or `const`.
```
GuillaumeGomez, line 87: There are actually other things that the search input supports. I'm working on fixing the search input, so better to keep this and not go into deeper explanations and instead mention that more possibilities are listed in the help popup.

NoraCodes: Gotchya!

src/doc/rustdoc/src/how-to-read-rustdoc.md
camelid: It sounds a bit funny to me too. I'd suggest something like "The majority of the screen is taken up by the main focus:" or, even simpler, "The majority of the screen is taken up by the documentation text\[...\]".

src/doc/rustdoc/src/how-to-read-rustdoc.md
```diff
@@ -0,0 +1,101 @@
...
+The majority of the screen is taken up with the main event:
+the documentation text for the item currently being viewed.
+The main title is the type and name of the item,
+such as "Struct `std::time::Duration`",
+along with a button to collapse or expand the top-level documentation for that
+item (`[+]` or `[-]`),
+a link to the source code (`[src]`),
+if [configured](the-doc-attribute.html#html_no_source),
+and the version in which the item became stable,
+if it's a stable item in the standard library.
```
camelid, line 27: This sentence is long; maybe use bullet points instead?

camelid (commented):
The one concern I have with this guide is that it will fall out of date as we improve the UI. Hopefully we can try to keep it up to date. Let's just make sure that the guide doesn't have scope creep, which would make it harder to keep current :)

camelid:
> I wonder: could we add screenshots to make it easier for readers to follow? Or would it be too difficult to keep it up-to-date?

Yeah, I think it'd be too difficult to keep up-to-date. People can compare side-by-side with a docs page if they want. Also, like NoraCodes said, it'd not be as accessible.

camelid:
> The one concern I have with this guide is that it will fall out of date as we improve the UI.

A related thing we should keep in mind is that docs generated with older versions of rustdoc (e.g., many crates on docs.rs) won't match up to the description in the guide. (Side note: Would it be feasible for docs.rs to regenerate docs, say once a year, for crates over a certain download rate even if they haven't been updated?)

src/doc/rustdoc/src/how-to-read-rustdoc.md
jsha: Yep, makes sense! We would also need to update https://doc.rust-lang.org/std/#how-to-read-this-documentation.

jyn514:
>. (Side note: @jyn514 would it be feasible for docs.rs to regenerate docs, say once a year, for crates over a certain download rate even if they haven't been updated?)

https://github.com/rust-lang/docs.rs/issues/464

src/doc/rustdoc/src/how-to-read-rustdoc.md
NoraCodes: Good call, done!

jyn514:
cc @pierwill

src/doc/rustdoc/src/how-to-read-rustdoc.md
```diff
@@ -0,0 +1,105 @@
...
+At the top is some at-a-glance info and controls:
+
+- the type and name of the item,
+  such as "Struct `std::time::Duration`",
+- a button to copy the item's path to the clipboard,
+  which is a clipboard item
+- a button to collapse or expand the top-level documentation for that item
+  (`[+]` or `[-]`),
+- a link to the source code (`[src]`),
+  if [configured](the-doc-attribute.html#html_no_source),
```
jyn514, line 29:
```suggestion
  if [configured](the-doc-attribute.html#html_no_source) and present. The source may not be available if the documentation was created with `cargo doc --no-deps`.
```

src/doc/rustdoc/src/how-to-read-rustdoc.md
```diff
@@ -0,0 +1,105 @@
...
+When viewing `rustdoc`'s output in a browser with JavaScript enabled,
+a dynamic interface appears at the top of the page.
+To the left is the theme picker, denoted with a paint-brush icon,
+and the search interface, help screen, and options appear to the right of that.
+
+### The Theme Picker
+
+Clicking on the theme picker provides a list of themes -
+by default `ayu`, `light`, and `dark` -
+which are available for viewing.
```
jyn514, line 68:
```suggestion
which are available for viewing. You can add custom themes with the [`--theme`](command-line-arguments.md#--theme-add-a-theme-to-the-documentation-output) flag.
```

src/doc/rustdoc/src/how-to-read-rustdoc.md
```diff
@@ -0,0 +1,105 @@
...
+meaning that the fuzzy match is made against the names of items.
+Matching names are shown on the left, and the first few words of their
+descriptions are given on the right.
+By clicking an item, you will navigate to its particular documentation.
+
+There are two other sets of results, shown as tabs in the search results pane.
+"In Parameters" shows matches for the string in the types of parameters to
+functions, and "In Return Types" shows matches in the return types of functions.
+Both are very useful when looking for a function whose name you can't quite
+bring to mind when you know the type you have or want.
```
jyn514, line 86: It's also useful if you don't know whether the function exists, but you think it should (e.g. `Option<T> -> T` or `String -> PathBuf`).

jyn514 (commented):
This is great! Thank you for working on this :)

src/doc/rustdoc/src/how-to-read-rustdoc.md
camelid: This is a guide for rustdoc *readers*, so I don't think we should mention `--theme`. This page shouldn't be comprehensive IMO, or it'll be overwhelming and people won't read it.

src/doc/rustdoc/src/how-to-read-rustdoc.md
camelid: Just noting that neither of those examples work as search queries currently. I assume you didn't mean using those as queries verbatim?

jyn514: Sure it does: 
image
 https://doc.rust-lang.org/stable/std/path/struct.PathBuf.html#method.from-5

camelid: The results include functions that don't have `String` in their parameters (like `Path::join`). `String -> PathBuf` implies that it will show only functions that takes `String`s and return `PathBuf`s.

jyn514: :man_shrugging: I don't think there exist any useful type-based search queries for which rustdoc doesn't have bugs. I don't think we should block documenting how to use the feature on fixing the bugs.

camelid: I just don't think we should give examples like `String -> PathBuf` since they won't behave as expected. We could suggest `-> PathBuf` or `* -> PathBuf` instead though (not sure which one of those is the more "correct" way).

jyn514: Those aren't *useful* searches though. We should give an example of why you would want to use the feature.

camelid: What do you mean? `String -> PathBuf` isn't useful either, because it returns the same results. Personally, I think we should avoid documenting type-based search here until the dust settles a bit (since Guillaume is working on an overhaul of it right now). After Guillaume finishes his overhaul, IIUC `String -> PathBuf` should work correctly.

> We should give an example of why you would want to use the feature.

Giving a broken example of why you'd want to use the feature seems worse than giving no example at all.

NoraCodes: Yeah, I think we should wait on this until it's updated, if that's already in progress.

NoraCodes:
Okay, I believe I've addressed all MR feedback here.

GuillaumeGomez:
Thanks! I'll review it tomorrow.

src/doc/rustdoc/src/how-to-read-rustdoc.md
GuillaumeGomez: (I opened the PR [here](https://github.com/rust-lang/rust/pull/90630)).

GuillaumeGomez (approved):
Looks good to me.

src/doc/rustdoc/src/how-to-read-rustdoc.md
jyn514: Hi @NoraCodes - did you see this comment? Was it intentional to not add this sentence?

jyn514 (commented):
Looks good to me other than one nit :) thanks for working on this!

src/doc/rustdoc/src/how-to-read-rustdoc.md
NoraCodes: Yeah, sorry! I thought there was a discussion about this but now I can't find it - did I hallucinate that? lol

NoraCodes: Okay, yeah, still can't find it. I don't think we should mention this here because it's intended to be a guide for docs readers - we document `--no-deps` elsewhere.

jyn514: Hmm, so the reason I thought to mention it was because it's pretty common for people to be confused by reading docs and having missing links. But I don't feel super strongly about it.

jyn514:
@bors r+ rollup

Thanks again!

src/doc/rustdoc/src/how-to-read-rustdoc.md
camelid: > Yeah, sorry! I thought there was a discussion about this but now I can't find it - did I hallucinate that? lol

@NoraCodes I think you're thinking of https://github.com/rust-lang/rust/pull/90487#discussion_r741328790, which was a discussion about another command-line flag.

PR #90488, opened 2021-11-01
More powerful const panic

nbdd0121:
Add a `const_panic_extra` feature gate that enables const panic to panic with any format arguments.

It should make the example given in the rust-lang/rfcs#2345 possible:
```rust
const fn parse_bool(s: &str) -> bool {
    match s {
        "true" => true,
        "false" => false,
        other => panic!("`{}` is not a valid bool", other),
    }
}
parse_bool("true");
parse_bool("false");
parse_bool("foo");
```

The mechanism implemented in this PR essentially extracts arguments from const-eval `Arguments` and delegate the formatting to formatting machinary that's compiled with rustc. Currently only these types are formatted:
* Non-reference primitives
* Reference to types under the blanket impl (e.g. `impl<T: Display> for &Display`)
* `String` (the only const `String` is empty string)
* `Arguments`
* `dyn Trait` if `Trait` is one of the formatting traits.

Other types will be formatted to `<failed to format {ty}>`.

I *think* it's possible to require that only types that can be correctly formatted are accepted by having an additional marker type, but that'll require const trait impl and specialization like this:
```rust
trait ConstDebug: Debug {}
impl<T: Debug> ConstDebug for T {}
impl const ConstDebug for &str {}
```

IMO it's okay to accept types that couldn't be formatted, given that const panic messages are considered diagnostics. It'll also make constification of `assert_eq!` easier (which uses `&dyn Debug` internally) since we current have no way to represent `&dyn ~const ConstDebug`.

I implement the feature in this way:
* This is reasonable simple;
* We won't see the whole formatting machinary be accepted by const-eval in near feature (previously suggested idea like using unleash miri doesn't work because the machinary's MIR isn't all available);
* People are removing info from panic messages to make it const, which is not desirable;
* There is a real need to make things like `Result::unwrap` const;

I plan to file/work on a few follow up PRs if this one is accepted:
* Constification of `Result::unwrap`
* Const panic fmt of additional types from std
* Const panic fmt of ADTs if the `Debug` implementation is derived

This is a bulky PR, but I've separated it to individually meaningful commits so I'd suggest to review by commits. Some changes are made to `ArgumentV1::new` to allow the `const_panic_extra` being used without `const_fn_fn_ptr_basics` and `const_mut_refs`. Given this is hot code I'd suggest a perf run.

r? @oli-obk 
@rustbot label: T-compiler A-const-eval

### DIFF
--- a/compiler/rustc_const_eval/src/const_eval/panic.rs
+++ b/compiler/rustc_const_eval/src/const_eval/panic.rs
@@ -1,0 +1,239 @@
+use rustc_hir::def_id::DefId;
+use rustc_middle::ty::{self, layout::LayoutOf, subst::Subst};
+use std::cell::Cell;
+use std::fmt::{self, Debug, Formatter};
+
+use crate::interpret::{FnVal, InterpCx, InterpErrorInfo, InterpResult, OpTy};
+
+use super::CompileTimeInterpreter;
+
+struct Arg<'mir, 'tcx, 'err> {
+    cx: &'err InterpCx<'mir, 'tcx, CompileTimeInterpreter<'mir, 'tcx>>,
+    arg: OpTy<'tcx>,
+    fmt_trait: DefId,
+    err: &'err Cell<Option<InterpErrorInfo<'tcx>>>,
+}
+
+impl Debug for Arg<'_, '_, '_> {
+    fn fmt(&self, f: &mut Formatter<'_>) -> fmt::Result {
+        match self.cx.fmt_arg(self.arg, self.fmt_trait, f) {
+            Ok(_) => Ok(()),
+            Err(e) => {
+                self.err.set(Some(e));
+                Err(fmt::Error)
+            }
+        }
+    }
+}
+
+impl<'mir, 'tcx> InterpCx<'mir, 'tcx, CompileTimeInterpreter<'mir, 'tcx>> {
+    fn fmt_arg(
+        &self,
+        arg: OpTy<'tcx>,
+        fmt_trait: DefId,
+        f: &mut Formatter<'_>,
+    ) -> InterpResult<'tcx> {
+        let fmt_trait_sym = self.tcx.item_name(fmt_trait);
+        let fmt_trait_name = fmt_trait_sym.as_str();
+
+        macro_rules! dispatch_fmt {
+            ($e: expr, $($t: ident)|*) => {
+                let _ = match &*fmt_trait_name {
+                    $(stringify!($t) => fmt::$t::fmt($e, f),)*
+                    _ => Debug::fmt($e, f),
+                };
+            }
+        }
+
+        match arg.layout.ty.kind() {
+            ty::Bool => {
+                let v = self.read_scalar(&arg)?.to_bool()?;
+                dispatch_fmt!(&v, Display);
+            }
+            ty::Char => {
+                let v = self.read_scalar(&arg)?.to_char()?;
+                dispatch_fmt!(&v, Display);
+            }
+            ty::Int(int_ty) => {
+                let v = self.read_scalar(&arg)?.check_init()?;
+                let v = match int_ty {
+                    ty::IntTy::I8 => v.to_i8()?.into(),
+                    ty::IntTy::I16 => v.to_i16()?.into(),
+                    ty::IntTy::I32 => v.to_i32()?.into(),
+                    ty::IntTy::I64 => v.to_i64()?.into(),
+                    ty::IntTy::I128 => v.to_i128()?,
+                    ty::IntTy::Isize => v.to_machine_isize(self)?.into(),
+                };
+                dispatch_fmt!(
+                    &v,
+                    Display | Binary | Octal | LowerHex | UpperHex | LowerExp | UpperExp
+                );
+            }
+            ty::Uint(int_ty) => {
+                let v = self.read_scalar(&arg)?.check_init()?;
+                let v = match int_ty {
+                    ty::UintTy::U8 => v.to_u8()?.into(),
+                    ty::UintTy::U16 => v.to_u16()?.into(),
+                    ty::UintTy::U32 => v.to_u32()?.into(),
+                    ty::UintTy::U64 => v.to_u64()?.into(),
+                    ty::UintTy::U128 => v.to_u128()?,
+                    ty::UintTy::Usize => v.to_machine_usize(self)?.into(),
+                };
+                dispatch_fmt!(
+                    &v,
+                    Display | Binary | Octal | LowerHex | UpperHex | LowerExp | UpperExp
+                );
+            }
+            ty::Float(ty::FloatTy::F32) => {
+                let v = f32::from_bits(self.read_scalar(&arg)?.to_u32()?);
+                dispatch_fmt!(&v, Display);
+            }
+            ty::Float(ty::FloatTy::F64) => {
+                let v = f64::from_bits(self.read_scalar(&arg)?.to_u64()?);
+                dispatch_fmt!(&v, Display);
+            }
+            ty::Str => {
+                let Ok(place) = arg.try_as_mplace() else {
+                    bug!("str is not in MemPlace");
+                };
+                let v = self.read_str(&place)?;
+                dispatch_fmt!(v, Display);
+            }
+            ty::Array(..) | ty::Slice(..) => {
+                let Ok(place) = arg.try_as_mplace() else {
+                    bug!("array/slice is not in MemPlace");
+                };
+                let err = Cell::new(None);
+                let mut debug_list = f.debug_list();
+                for field in self.mplace_array_fields(&place)? {
+                    debug_list.entry(&Arg { cx: self, arg: field?.into(), fmt_trait, err: &err });
+                }
+                let _ = debug_list.finish();
+                if let Some(e) = err.into_inner() {
+                    return Err(e);
+                }
+            }
+            ty::RawPtr(..) | ty::FnPtr(..) => {
+                // This isn't precisely how Pointer is implemented, but it's best we can do.
+                let ptr = self.read_pointer(&arg)?;
+                let _ = write!(f, "{:?}", ptr);
+            }
+            ty::Tuple(substs) => {
+                let err = Cell::new(None);
+                let mut debug_tuple = f.debug_tuple("");
+                for i in 0..substs.len() {
+                    debug_tuple.field(&Arg {
+                        cx: self,
+                        arg: self.operand_field(&arg, i)?,
+                        fmt_trait,
+                        err: &err,
+                    });
+                }
+                let _ = debug_tuple.finish();
+                if let Some(e) = err.into_inner() {
+                    return Err(e);
+                }
+            }
+
+            // FIXME(nbdd0121): extend to allow fmt trait as super trait
+            ty::Dynamic(list, _) if list.principal_def_id() == Some(fmt_trait) => {
+                let Ok(place) = arg.try_as_mplace() else {
+                    bug!("dyn is not in MemPlace");
+                };
+                let place = self.unpack_dyn_trait(&place)?.1;
+                return self.fmt_arg(place.into(), fmt_trait, f);
+            }
+
+            ty::Ref(..) if fmt_trait_name == "Pointer" => {
+                let ptr = self.read_pointer(&arg)?;
+                let _ = write!(f, "{:?}", ptr);
+            }
+            ty::Ref(..) => {
+                // FIXME(nbdd0121): User can implement trait on &UserType, so this isn't always correct.
+                let place = self.deref_operand(&arg)?;
+                return self.fmt_arg(place.into(), fmt_trait, f);
+            }
+
+            // FIXME(nbdd0121): ty::Adt(..) => (),
+            _ => {
+                let _ = write!(f, "<failed to format {}>", arg.layout.ty);
+            }
+        }
+        Ok(())
+    }
+
+    fn fmt_arguments(&self, arguments: OpTy<'tcx>, f: &mut Formatter<'_>) -> InterpResult<'tcx> {
+        // Check we are dealing with the simple form
+        let fmt_variant_idx = self.read_discriminant(&self.operand_field(&arguments, 1)?)?.1;
+        if fmt_variant_idx.as_usize() != 0 {
+            // FIXME(nbdd0121): implement complex format
+            let _ = write!(f, "<cannot evaluate complex format>");
+            return Ok(());
+        }
+
+        // `pieces: &[&str]`
+        let pieces_place = self.deref_operand(&self.operand_field(&arguments, 0)?)?;
+        let mut pieces = Vec::new();
+        for piece in self.mplace_array_fields(&pieces_place)? {
+            let piece: OpTy<'tcx> = piece?.into();
+            pieces.push(self.read_str(&self.deref_operand(&piece)?)?);
+        }
+
+        // `args: &[ArgumentV1]`
+        let args_place = self.deref_operand(&self.operand_field(&arguments, 2)?)?;
+        let mut args = Vec::new();
+        let err = Cell::new(None);
+        for arg in self.mplace_array_fields(&args_place)? {
+            let arg: OpTy<'tcx> = arg?.into();
+
+            let fmt_fn = self.memory.get_fn(self.read_pointer(&self.operand_field(&arg, 1)?)?)?;
+            let fmt_fn = match fmt_fn {
+                FnVal::Instance(instance) => instance,
+                FnVal::Other(o) => match o {},
+            };
+
+            // The formatter must an instance of fmt method of a fmt trait.
+            let Some(fmt_impl) = self.tcx.impl_of_method(fmt_fn.def_id()) else {
+                throw_unsup_format!("fmt function is not from trait impl")
+            };
+            let Some(fmt_trait) = self.tcx.impl_trait_ref(fmt_impl) else {
+                throw_unsup_format!("fmt function is not from trait impl")
+            };
+
+            // Retrieve the trait ref with concrete self ty.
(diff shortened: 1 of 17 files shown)


camelid:
(FYI, you may want to wait to fix tests until after the bors try run finishes; IIRC, sometimes pushing will cancel a try.)

library/core/src/fmt/mod.rs
```diff
@@ -323,6 +335,16 @@ impl<'a> ArgumentV1<'a> {
...
 
+    arg_new!(new_display, Display);
+    arg_new!(new_debug, Debug);
+    arg_new!(new_octal, Octal);
+    arg_new!(new_lower_hex, LowerHex);
+    arg_new!(new_upper_hex, UpperHex);
+    arg_new!(new_pointer, Pointer);
+    arg_new!(new_binary, Binary);
+    arg_new!(new_lower_exp, LowerExp);
+    arg_new!(new_upper_exp, UpperExp);
```
bjorn3, line 346: Can these be inlined into the format_args!() expansion?

nbdd0121: These are generic functions so I think yes.

nbdd0121:
(Pushed to the wrong branch)

compiler/rustc_const_eval/src/const_eval/panic.rs
```diff
@@ -0,0 +1,262 @@
...
+        f: &mut Formatter<'_>,
+    ) -> InterpResult<'tcx> {
+        let fmt_trait_sym = self.tcx.item_name(fmt_trait);
+        let fmt_trait_name = fmt_trait_sym.as_str();
+
+        macro_rules! dispatch_fmt {
+            ($e: expr, $($t: ident)|*) => {
+                let _ = match &*fmt_trait_name {
+                    $(stringify!($t) => fmt::$t::fmt($e, f),)*
+                    _ => Debug::fmt($e, f),
```
RalfJung, line 44: Could you add a comment explaining what this macros does? I cannot make sense of it.

Also please not explicitly that we are dispatching to the standard library formatting machinery here, so we are guaranteed that this stays in sync with the logic in the standard library -- is that correct?

compiler/rustc_const_eval/src/const_eval/panic.rs
```diff
@@ -0,0 +1,262 @@
...
+                    ty::IntTy::I16 => v.to_i16()?.into(),
+                    ty::IntTy::I32 => v.to_i32()?.into(),
+                    ty::IntTy::I64 => v.to_i64()?.into(),
+                    ty::IntTy::I128 => v.to_i128()?,
+                    ty::IntTy::Isize => v.to_machine_isize(self)?.into(),
+                };
+                dispatch_fmt!(
+                    &v,
+                    Display | Binary | Octal | LowerHex | UpperHex | LowerExp | UpperExp
+                );
```
RalfJung, line 71: This makes some assumptions about how integers are formatted, namely that using a larger integer type will not make a difference. This assumption should at least be explicitly documented in a comment here.

Also, do these `LowerHex` etc things her mean that hex output is supported? I don't see the tests covering that possibility.

RalfJung:
Wow, I am impressed how well this works!

However, I have some concerns of maintainability here. This requires us to commit to maintaining a rather tricky bit of interpreter code and keep it in sync with the type hierarchy. To be fair this is done very well and kept outside the core interpreter engine, good job! But we should still at least be explicit about this, and also be wary of going too far down the road of implementing things "on the meta level" of the interpreter.

nbdd0121:
> However, I have some concerns of maintainability here. This requires us to commit to maintaining a rather tricky bit of interpreter code and keep it in sync with the type hierarchy. To be fair this is done very well and kept outside the core interpreter engine, good job! But we should still at least be explicit about this, and also be wary of going too far down the road of implementing things "on the meta level" of the interpreter.

Luckily we don't have a lot const-constructible types in std with custom `Debug` impls. I plan to add a few common ones from std and probably later add a fallback mechanism that prints out the fields for types that couldn't be formatted, so the diagnostics would still be useful to users for troubleshooting but wouldn't complicate the const panic logic.

compiler/rustc_const_eval/src/const_eval/panic.rs
nbdd0121: This just generates match arms like `"Display" => fmt::Display::fmt(e, f),`.

Yes, it's correct that this would mean the formatting result to be in sync with std. Though I don't think we want to guarantee  that to the user.

camelid: It seems unfortunate to do string comparisons on the trait name. I wonder if it would be better to make Display etc. lang items instead?

nbdd0121: They certainly shouldn't be lang items but could be made into diagnostic items. With that said, `format_args!` expansion in `rustc_builtin_macros` is using strings to name these as well, so I don't think it's a big deal.

RalfJung:
Users can define their own such types though. I guess the idea is that const checking will reject code that formats those?

compiler/rustc_const_eval/src/const_eval/panic.rs
RalfJung: > Yes, it's correct that this would mean the formatting result to be in sync with std. Though I don't think we want to guarantee that to the user.

I think we have to make this guarantee. All this machinery should be entirely invisible, it just works around limitations in the main CTFE machinery.

RalfJung: Also, instead of having a `Debug` fallback case, shouldn't this `bug!` in case of an unknown trait name, to ensure we do dispatch to the right code?

compiler/rustc_const_eval/src/transform/check_consts/check.rs
```diff
@@ -893,7 +893,7 @@ impl Visitor<'tcx> for Checker<'mir, 'tcx> {
                 if Some(callee) == tcx.lang_items().begin_panic_fn() {
```
RalfJung, line 893: The comment on the previous line seems outdated now? (Same below.)

RalfJung:
Oh I see, you mention this in the OP:

> Other types will be formatted to <failed to format {ty}>.

Hm... I guess when it's just panics this might be okay? Still seems a bit sketchy though. And certainly this needs a test.

bjorn3:
> I plan to add a few common ones from std and probably later add a fallback mechanism that prints out the fields for types that couldn't be formatted

You could copy the code from priroda: https://github.com/oli-obk/priroda/blob/14c8c48cdf729955d283f85b36549de9a690ae1b/src/render/locals.rs#L148

oli-obk:
While I would not have recommended putting in the work to do this PR, and the code will go away once (far in the future) we interpret the debug impls themselves, I think we should indeed go forward with this PR. We can always fine tune the way messages are rendered, but it's imo of no consequence if the messages at compile time are lossy or badly formatted. We have a path to making it better and it is already better than nothing.

I'll review in more detail later today, as I have a few forward compat concerns that I need to be sure are not a problem

library/core/src/macros/mod.rs
```diff
@@ -137,7 +137,7 @@ macro_rules! assert_ne {
 /// assert_matches!(c, Ok(x) | Err(x) if x.len() < 100);
 /// ```
 #[unstable(feature = "assert_matches", issue = "82775")]
-#[allow_internal_unstable(core_panic)]
+#[allow_internal_unstable(core_panic, const_format_args)]
 #[rustc_macro_transparency = "semitransparent"]
 pub macro assert_matches {
```
oli-obk, line 142: So... my worry about forward compatibility is that I believe one could now call `assert_matches` on a user defined type that does not implement `const Display`. If the match succeeds, there will be no error, even if we may want to error in the future because there is no `const Display` impl.

For the assert macros that's not so much a problem, as they already require `const PartialEq` essentially.

I'm also not sure how this affects the `panic!` macro. Can we now do `if cond { panic!("{}", some_user_defined_type) }`? If so, I think that will bite us in the future.

nbdd0121: > I'm also not sure how this affects the `panic!` macro. Can we now do `if cond { panic!("{}", some_user_defined_type) }`? If so, I think that will bite us in the future.

Yes. But I don't think this will bite us in the future. We can just display "<cannot be formatted>" for types that didn't have `const Display` implementation (could be achieved either by specialization on whether a const Display impl exists, or an intrinsic/lang_item).

IMO panicking with some info is better than not allowing the user to panic; user would just remove information from panic messages to those accepted by const eval, and then the info that could have been displayed in runtime would be lost.

oli-obk: Well... the same argument goes the other way. Users may want to know when we are ignoring the code they wrote because they missed a `const` somewhere.

Like... it feels to me like allowing `panic!("{}", non_display_type)` at runtime and just printing "sorry can't do" when panicking.

It would also mean we would have to keep the fallback logic forever, which seems unfortunate when we could just force users to use const Display impls at some point.

nbdd0121: > Well... the same argument goes the other way. Users may want to know when we are ignoring the code they wrote because they missed a `const` somewhere.
> 
> Like... it feels to me like allowing `panic!("{}", non_display_type)` at runtime and just printing "sorry can't do" when panicking.

Well, this all happens at compile time, so (1) it's easy to reproduce for the user (no need to run the program and wait for the panic to hit), so less info is acceptable than if it happens in the runtime and (2) the compiler has a lot of info to produce useful diagnostics. The compiler could produce additional diagnostics saying `<ty> cannot be formatted because its Display impl is not const`.

> 
> It would also mean we would have to keep the fallback logic forever, which seems unfortunate when we could just force users to use const Display impls at some point.

It wouldn't be a lot or a complex logic though. And honestly I don't think we are anywhere near of getting to a point that we can have a working (and ergnomical)  `const Display` any time soon. Like whether `#[derive(Debug)]` would implement `const Debug` or just `Debug`? With the current approach I can just eval all derived Debug, and if any field isn't const formattable it would only need a diagnostics.

And then there is an issue that people want more powerful panic right now, e.g. with `Result::unwrap`. If we want to reserve the possibily of removing the fallback logic, what would the bound look like? Solution would most likely involve an unstable `ConstDebug` bound and then it could probably never be stablised.

RalfJung: > Well, this all happens at compile time, so (1) it's easy to reproduce for the user (no need to run the program and wait for the panic to hit), so less info is acceptable than if it happens in the runtime 

Not really: the panic might be in a `const fn` in a library, so while it runs at compile time it does not necessarily run at compile time *of the current crate*.

oli-obk: Opened a zulip thread in https://rust-lang.zulipchat.com/#narrow/stream/219381-t-libs/topic/const.20panic!.20for.20all.20types.20.2390488 so we can discuss this with the libs API team, as I am unsure of how to proceed here.

steffahn:
Is there precedent for `const fn`s behaving differently between compile-time and run-time?

oli-obk:
> Is there precedent for `const fn`s behaving differently between compile-time and run-time?

I think that is not the issue here, as we are only changing the output in case of a compilation failure. If we're being pedantic, panicking at compile-time already is very different from the same panic at runtime as there are spans and backtraces that are completely different.

steffahn:
> I think that is not the issue here

Okay. I don’t ask because I necessarily thing that would be an issue. I’m asking because a general mechanism for _allowing_ different behavior at compile-time vs. runtime, e.g. a macro
```rs
compiler_vs_run_time! {
    // if at compiler time:
    {
        // some code here
    }
    // if at run time:
    {
        // some code here
    }
}
```
that can be used in a `const fn` and allows arbitrary code (without the restrictions of `const` evaluation) in it’s “at run time” branch. Then, for anyone that needs it _now_, custom `const_panic` macros or `<Result as ResultExt>::const_unwrap` methods could be offered in an external crate, without any need for the standard library’s `panic` or `Result::unwrap` to _indefinitely_ support questionable fall-back logic with under-constrained trait bounds.

Admitted, with such a crate then _every_ case of compile-time evaluation of those panics would result in _not_ printing the values involved, but it’d still enable people to keep their run-time diagnostics good.

<hr>

If there’s concerns that a general mechanism for different run-time vs. compile-time behavior is unwanted, there’s also an option of a `compiler_vs_run_time_panic` macro that’s the same as explained above, but requires that both branches eventually panic (e.g. by requiring type `!` and disallowing any use of `return`/`continue`/`break`).

fee1-dead:
FWIW, you do not need `const_eval_select` to make `Result::unwrap` work. You just need `const_trait_impl` and impose a `~const Drop` bound on the error type.

steffahn:
Ah, I see, I didn’t know about `const_eval_select`, hence my question for

> precedent for `const fn`s behaving differently between compile-time and run-time

Looking at its documentation, it seems like it doesn’t necessarily need to be `unsafe` in case you only allow callbacks returning `!`, right? Unless… what exactly does “The Rust compiler assumes that it is sound to replace a call to a `const fn` with the result produced by evaluating it at compile-time.” entail in case the const fn call results in a panic?

Of course it uses `const` traits for the callbacks, in particular `const FnOnce`, so it’s not really possible to stabilize an API like it anytime soon; a macro seems more reasonable.

fee1-dead:
The problem for a macro is that we would have to lower it all the way to MIR so that CTFE would interpret it differently to what code generation would produce. It's too much work, and a function is easier to implement.

library/core/src/fmt/mod.rs
bjorn3: I actually meant can we inline their definition into the format_args!() expansion and omit those function definitions themself.

Dylan-DPC:
@nbdd0121 this has conflicts to be resolved, once done we can get this reviewed quickly after that

oli-obk:
Summary of [the zulip thread](https://rust-lang.zulipchat.com/#narrow/stream/219381-t-libs/topic/const.20panic!.20for.20all.20types.20.2390488):

There are three ways to continue:

* do nothing, keep status quo. Not bad, not great either
* do this PR, allowing *all* types that implement `Display`, but bail out by printing some form of `<can't print>`. This means we can just print all types conveniently, but it also means we need to keep this complex logic around forever and possibly extend it with more and more special handling. It may also be surprising, similar to how it would be suprising if `{:?}` would accept types that don't implement `Debug`, but just print some default message
* restrict this PR to types that we know can have some CTFE logic to print them in the future, so that we can revert the PR some day without a breaking change and instead.

Ralf and I consider option 2 to be problematic and would prefer to go with option 3. We can always go for option 2 later, but if we start with it we can't revert to option 3.

That said, we need a feature gate for this anyway, so we could just merge it modulo some additional tests making sure the feature gate works and split out option 3 under a separate gate later.

c410-f3r:
In order to make https://github.com/rust-lang/rust/pull/96496 work in constant environments (or even a future stabilization), this PR or functionality is a must have.

nbdd0121:
This PR has been stagnated due to lack of strong motivating use case -- I am glad that it now has one!

oli-obk:
Can you elaborate on what functionality of this PR is needed? Afaict option 3 is sufficient for https://github.com/rust-lang/rust/pull/96496

nbdd0121:
If we have `a, b` of type `T: Debug` but not a restricted const printable type, then `assert!(a == b)` currently works in const context, but not with #96496.

nbdd0121:
> That sounds like a bug in the implementation of #96496.
> 
> Inside a `const fn`, it can only debug-print things if they are `const Debug`, of course.

That'll make a const fn and a non-const fn behave differently even if the non-const fn is not called in const context.

oli-obk:
> That'll make a const fn and a non-const fn behave differently even if the non-const fn is not called in const context.

you cannot call non-const fn from const fn at all. Did you flip `non-const` and `const` in the second part of your sentence?

oli-obk:
As long as https://github.com/rust-lang/rust/pull/96496 only requires integers and booleans to be formattable, then it will work out of the box. If it requires arbitrary types to be formattable (e.g. `Mutex` or something else that does things that are impossible in const eval), then it will not be possible in const eval

c410-f3r:
> As long as #96496 only requires integers and booleans to be formattable, then it will work out of the box. If it requires arbitrary types to be formattable (e.g. `Mutex` or something else that does things that are impossible in const eval), then it will not be possible in const eval

Yes, types are arbitrary but all must implement `Copy` and `Debug` to be eligible for displaying. Will it suffice the required restrictions of `const` printing?

c410-f3r:
> > Yes, types are arbitrary but all must implement `Copy` and `Debug` to be eligible for displaying. Will these bounds satisfy the required restrictions of `const` printing?
> 
> ah perfect. If you are going via trait bounds instead of just syntax, then it will reject all types that do not implement `const Debug`, but only within const contexts. Even if your current impl does not manage this yet, we can always address that later. I feared it was a syntactical thing, where `assert!(non_debug_type == foo)` would fail to compile in the future.

Thanks! Looking forward to see 

```rust
const _: () = {
    let a = 1i32;
    panic!("Something {}", a);
};
```

Work someday

oli-obk:
> That'll make a const fn and a non-const fn behave differently even if the const fn is not called in const context.

This is what will happen for all other traits, too, not just Debug/Display.

On the PR itself:

I would like to move this PR along. But neither Ralf nor I will approve the current version. I know you feel strongly for allowing arbitrary types to be printed in const eval, even if they are unprintable (e.g. pointers). While I personally don't ever want this, and think you should not propose it, I think the right way to get this approved in your version would be an RFC/lang-team-MCP. Your proposed scheme just diverges too far from what we have right now, it can't just be a simple feature and stabilization.

Please consider restricting your implementation to builtin types that are guaranteed to get const evaluable Debug and Display impls

Dylan-DPC:
Closing this based on [comment](https://github.com/rust-lang/rust/pull/90488#issuecomment-1125276519).

PR #90489, opened 2021-11-01
rustdoc: Go back to loading all external crates unconditionally

### LINKED ISSUE #84738: Index out of bounds when running `cargo doc` in `rustc_metadata/src/creader.rs`

jyn514:
This *continues* to cause regressions. This code will be unnecessary
once access to the resolver happens fully before creating the tyctxt
(#83761), so load all crates unconditionally for now. To minimize churn, this leaves in the code for loading crates selectively.

"Fixes" https://github.com/rust-lang/rust/issues/84738. Previously: https://github.com/rust-lang/rust/pull/83738, https://github.com/rust-lang/rust/pull/85749, https://github.com/rust-lang/rust/pull/88215

r? @petrochenkov cc @camelid (this should fix the "index out of bounds" error you had while looking up `crate_name`).

### DIFF
--- a/src/librustdoc/core.rs
+++ b/src/librustdoc/core.rs
@@ -16,13 +16,15 @@ use rustc_middle::hir::map::Map;
 use rustc_middle::middle::privacy::AccessLevels;
 use rustc_middle::ty::{ParamEnv, Ty, TyCtxt};
 use rustc_resolve as resolve;
+use rustc_resolve::Namespace::TypeNS;
 use rustc_session::config::{self, CrateType, ErrorOutputType};
 use rustc_session::lint;
 use rustc_session::DiagnosticOutput;
 use rustc_session::Session;
+use rustc_span::def_id::CRATE_DEF_INDEX;
 use rustc_span::source_map;
 use rustc_span::symbol::sym;
-use rustc_span::Span;
+use rustc_span::{Span, DUMMY_SP};
 
 use std::cell::RefCell;
 use std::lazy::SyncLazy;
@@ -299,13 +301,43 @@ crate fn create_config(
 }
 
 crate fn create_resolver<'a>(
+    externs: config::Externs,
     queries: &Queries<'a>,
     sess: &Session,
 ) -> Rc<RefCell<interface::BoxedResolver>> {
     let (krate, resolver, _) = &*abort_on_err(queries.expansion(), sess).peek();
     let resolver = resolver.clone();
 
-    crate::passes::collect_intra_doc_links::load_intra_link_crates(resolver, krate)
+    let resolver = crate::passes::collect_intra_doc_links::load_intra_link_crates(resolver, krate);
+
+    // FIXME: somehow rustdoc is still missing crates even though we loaded all
+    // the known necessary crates. Load them all unconditionally until we find a way to fix this.
+    // DO NOT REMOVE THIS without first testing on the reproducer in
+    // https://github.com/jyn514/objr/commit/edcee7b8124abf0e4c63873e8422ff81beb11ebb
+    let extern_names: Vec<String> = externs
+        .iter()
+        .filter(|(_, entry)| entry.add_prelude)
+        .map(|(name, _)| name)
+        .cloned()
+        .collect();
+    resolver.borrow_mut().access(|resolver| {
+        sess.time("load_extern_crates", || {
+            for extern_name in &extern_names {
+                debug!("loading extern crate {}", extern_name);
+                if let Err(()) = resolver
+                    .resolve_str_path_error(
+                        DUMMY_SP,
+                        extern_name,
+                        TypeNS,
+                        LocalDefId { local_def_index: CRATE_DEF_INDEX }.to_def_id(),
+                  ) {
+                    warn!("unable to resolve external crate {} (do you have an unused `--extern` crate?)", extern_name)
+                  }
+            }
+        });
+    });
+
+    resolver
 }
 
 crate fn run_global_ctxt(
--- a/src/librustdoc/lib.rs
+++ b/src/librustdoc/lib.rs
@@ -755,6 +755,7 @@ fn main_options(options: config::Options) -> MainResult {
     let default_passes = options.default_passes;
     let output_format = options.output_format;
     // FIXME: fix this clone (especially render_options)
+    let externs = options.externs.clone();
     let manual_passes = options.manual_passes.clone();
     let render_options = options.render_options.clone();
     let scrape_examples_options = options.scrape_examples_options.clone();
@@ -773,7 +774,7 @@ fn main_options(options: config::Options) -> MainResult {
             // We need to hold on to the complete resolver, so we cause everything to be
             // cloned for the analysis passes to use. Suboptimal, but necessary in the
             // current architecture.
-            let resolver = core::create_resolver(queries, sess);
+            let resolver = core::create_resolver(externs, queries, sess);
 
             if sess.has_errors() {
                 sess.fatal("Compilation failed, aborting rustdoc");
--- a/src/librustdoc/passes/collect_intra_doc_links/early.rs
+++ b/src/librustdoc/passes/collect_intra_doc_links/early.rs
@@ -1,3 +1,4 @@
+use ast::visit;
 use rustc_ast as ast;
 use rustc_hir::def::Namespace::TypeNS;
 use rustc_hir::def_id::{LocalDefId, CRATE_DEF_ID};
@@ -16,7 +17,7 @@ crate fn load_intra_link_crates(resolver: Resolver, krate: &ast::Crate) -> Resol
     let mut loader = IntraLinkCrateLoader { current_mod: CRATE_DEF_ID, resolver };
     // `walk_crate` doesn't visit the crate itself for some reason.
     loader.load_links_in_attrs(&krate.attrs, krate.span);
-    ast::visit::walk_crate(&mut loader, krate);
+    visit::walk_crate(&mut loader, krate);
     loader.resolver
 }
 
@@ -54,7 +55,12 @@ impl IntraLinkCrateLoader {
     }
 }
 
-impl ast::visit::Visitor<'_> for IntraLinkCrateLoader {
+impl visit::Visitor<'_> for IntraLinkCrateLoader {
+    fn visit_foreign_item(&mut self, item: &ast::ForeignItem) {
+        self.load_links_in_attrs(&item.attrs, item.span);
+        visit::walk_foreign_item(self, item)
+    }
+
     fn visit_item(&mut self, item: &ast::Item) {
         use rustc_ast_lowering::ResolverAstLowering;
 
@@ -64,12 +70,29 @@ impl ast::visit::Visitor<'_> for IntraLinkCrateLoader {
             let old_mod = mem::replace(&mut self.current_mod, new_mod);
 
             self.load_links_in_attrs(&item.attrs, item.span);
-            ast::visit::walk_item(self, item);
+            visit::walk_item(self, item);
 
             self.current_mod = old_mod;
         } else {
             self.load_links_in_attrs(&item.attrs, item.span);
-            ast::visit::walk_item(self, item);
+            visit::walk_item(self, item);
         }
+    }
+
+    // NOTE: if doc-comments are ever allowed on function parameters, this will have to implement `visit_param` too.
+
+    fn visit_assoc_item(&mut self, item: &ast::AssocItem, ctxt: visit::AssocCtxt) {
+        self.load_links_in_attrs(&item.attrs, item.span);
+        visit::walk_assoc_item(self, item, ctxt)
+    }
+
+    fn visit_field_def(&mut self, field: &ast::FieldDef) {
+        self.load_links_in_attrs(&field.attrs, field.span);
+        visit::walk_field_def(self, field)
+    }
+
+    fn visit_variant(&mut self, v: &ast::Variant) {
+        self.load_links_in_attrs(&v.attrs, v.span);
+        visit::walk_variant(self, v)
     }
 }
--- a/src/test/rustdoc-ui/intra-doc/auxiliary/dep1.rs
+++ b/src/test/rustdoc-ui/intra-doc/auxiliary/dep1.rs
@@ -1,0 +1,1 @@
+// intentionally empty
--- a/src/test/rustdoc-ui/intra-doc/auxiliary/dep2.rs
+++ b/src/test/rustdoc-ui/intra-doc/auxiliary/dep2.rs
@@ -1,0 +1,1 @@
+// intentionally empty
--- a/src/test/rustdoc-ui/intra-doc/auxiliary/dep3.rs
+++ b/src/test/rustdoc-ui/intra-doc/auxiliary/dep3.rs
@@ -1,0 +1,1 @@
+// intentionally empty
--- a/src/test/rustdoc-ui/intra-doc/auxiliary/dep4.rs
+++ b/src/test/rustdoc-ui/intra-doc/auxiliary/dep4.rs
@@ -1,0 +1,1 @@
+// intentionally empty
--- a/src/test/rustdoc-ui/intra-doc/extern-crate-load.rs
+++ b/src/test/rustdoc-ui/intra-doc/extern-crate-load.rs
@@ -1,0 +1,26 @@
+// check-pass
+// aux-crate:dep1=dep1.rs
+// aux-crate:dep2=dep2.rs
+// aux-crate:dep3=dep3.rs
+// aux-crate:dep4=dep4.rs
+#![deny(rustdoc::broken_intra_doc_links)]
+
+pub trait Trait {
+    /// [dep1]
+    type Item;
+}
+
+pub struct S {
+    /// [dep2]
+    pub x: usize,
+}
+
+extern "C" {
+    /// [dep3]
+    pub fn printf();
+}
+
+pub enum E {
+    /// [dep4]
+    A
+}


camelid:
> (this should fix the "index out of bounds" error you had while looking up `crate_name`).

Just a followup so we don't get confused later on: This PR doesn't fix that ICE; they seem to be different bugs.

petrochenkov:
@jyn514 
The minimized reproduction in #84738 looks like https://github.com/rust-lang/rust/pull/88679#discussion_r703056971.
Could you check whether visiting associated items in `IntraLinkCrateLoader` fixes #84738?

jyn514:
@petrochenkov given how many regressions this has caused, I don't particularly want to keep playing wack-a-mole with fixing the bugs. I'm fine with fixing the bugs in parallel, but I don't think we should be testing things "in-prod" as it were, we don't need to enable this until rustdoc's fundamental problems with crate loading are resolved (https://github.com/rust-lang/rust/issues/83761).

petrochenkov:
>given how many regressions this has caused

Which regressions?
I only see #84738 right now, which is caused by the missing visiting of associated items in `IntraLinkCrateLoader` with 99% certainty.
I'd also add visiting for at least fields and enum variants in the same PR, because the doc link resolver scans them as well. I'm not sure which other nodes need to be visited, I've seen some relevant `enum` somewhere in rustdoc, but don't remember how it's called.

jyn514:
> Which regressions?

https://github.com/rust-lang/rust/issues/84738 is the only one I know of, but lots of people have been hitting it, and it was hit the last time we tried to land this patch too.

> I'd also add visiting for at least fields and enum variants in the same PR, because the doc link resolver scans them as well. I'm not sure which other nodes need to be visited, I've seen some relevant enum somewhere in rustdoc, but don't remember how it's called.

Right, yes, this is what I mean, there's no guarantee that fixing one issue will fix the others, and it happens rarely enough that usually the bug doesn't get caught until the patch hits stable. And I'm not sure what *benefit* it brings, no one is using https://github.com/rust-lang/rust/issues/68427 in practice.

jyn514:
I just spent about half an hour to come up with the following impls for the visitor:
```rust
    fn visit_foreign_item(&mut self, item: &ast::ForeignItem) {
        self.load_links_in_attrs(&item.attrs, item.span);
        visit::walk_foreign_item(self, item)
    }

    // NOTE: if doc-comments are ever allowed on function parameters, this will have to implement `visit_param` too.

    fn visit_assoc_item(&mut self, item: &ast::AssocItem, ctxt: visit::AssocCtxt) {
        self.load_links_in_attrs(&item.attrs, item.span);
        visit::walk_assoc_item(self, item, ctxt)
    }

    fn visit_field_def(&mut self, field: &ast::FieldDef) {
        self.load_links_in_attrs(&field.attrs, field.span);
        visit::walk_field_def(self, field)
    }

    fn visit_variant(&mut self, v: &ast::Variant) {
        self.load_links_in_attrs(&v.attrs, v.span);
        visit::walk_variant(self, v)
    }
}
```
finding https://github.com/rust-lang/rust/issues/90610 in the process, and I'm still not sure I got them all. If you really think this is a good idea, I'd suggest adding a `visit_attr_list` function to `Visitor` so I don't have to rewrite all of these and be worried I'm missing one. But again, I think this is really high risk for the little benefit we get from it.

petrochenkov:
>I've seen some relevant `enum` somewhere in rustdoc

It's `ItemKind`, and `Item`s are constructed from AST by the `impl Clean<Item> for ...` function.
So the visitor methods above should cover all of the required AST nodes.
These methods are necessary for #83761 anyway, so let's land them now.
(Also could you add test cases for all these new nodes loading crates through doc links?)

The only remaining *potential* source of missing extern crates that I'm aware of is reexport inlining.
If you can build a test case with a link on an import attempting to access a crate not previously loaded by `IntraLinkCrateLoader`, then we can land the "load all external crates" hack as well, because I don't want you to implement all that reexport walking now (it should be implemented in #88679, but it may cause perf issues there).

jyn514:
@petrochenkov the cases I added didn't actually fix the issue, it still panics. I'll add them to this PR anyway.

I'm not sure how to come up with an MCVE for the ICEs this fixes. I can put some more time into it but it will probably be a few days before I have time.

petrochenkov:
I see, so it's probably reexport inlining after all.
We have 3.5 weeks until the next release, so there's some time to come up with an MCVE.

I'm also still interested in tests for fields, variants, and associated and foreign items.
A doc comment on such nodes needs to load a crate that is not loaded in any other way (so we either need at least 4 dependency crates and one test, or 4 separate tests and one dependency crate).

jyn514:
Ok, I added a test for all 4 new methods (and ran the test on master to ensure it failed before this PR).

petrochenkov:
>We have 3.5 weeks until the next release, so there's some time to come up with an MCVE.

@rustbot label +S-waiting-on-author -S-waiting-on-review

jyn514:
@petrochenkov https://github.com/jyn514/objr/commit/cf1c87598efdfee5608964206c2b038636b0e116

It requires at least 4 crates, a re-export from a module, and a blanket impl.
```
$ (cd corevideor/ && cargo tree)
corevideor v0.1.0 (/home/joshua/test-rustdoc/rustc-84738/corevideor)
├── coregraphicsr v0.1.0 (/home/joshua/test-rustdoc/rustc-84738/coregraphicsr)
│   └── objr v0.2.0 (/home/joshua/test-rustdoc/rustc-84738/objr-upstream)
└── objr v0.1.0 (/home/joshua/test-rustdoc/rustc-84738/objr)
$ fd .rs | while read f; do echo $f; bat $f; done
coregraphicsr/src/lib.rs
   1 extern crate objr;
corevideor/src/lib.rs
   1 extern crate coregraphicsr;
objr/src/lib.rs
objr-upstream/src/lib.rs
   1 extern crate self as objr;
   2 
   3 mod m {
   4     pub trait Trait {
   5         /// [objr]
   6         fn f() {}
   7     }
   8     impl <T> Trait for T {
   9         fn f() {}
  10     }
  11 }
  12 
  13 pub use m::Trait;
```

jyn514:
The reason it's not being caught by the early pass is because rustdoc is explicitly visit traits from other crates for some reason :/ https://github.com/rust-lang/rust/blob/43192ca3d72ed0ca42ba913137585219ba0a691f/src/librustdoc/visit.rs#L67

jyn514:
The traits are being added here: https://github.com/rust-lang/rust/blob/43192ca3d72ed0ca42ba913137585219ba0a691f/src/librustdoc/clean/utils.rs#L442-L444

Removing that block ... doesn't actually cause any tests to fail???
@GuillaumeGomez do you know what's going on here? Why is rustdoc storing traits (and only traits) from other crates?

jyn514:
Oh, there were multiple places `record_extern_trait` is called. The second one does cause a whole bunch of tests to fail; but it also fixes the ICE.

<details>

```
failures:
    [rustdoc] rustdoc/auto-impl-primitive.rs
    [rustdoc] rustdoc/blanket-reexport-item.rs
    [rustdoc] rustdoc/const-generics/add-impl.rs
    [rustdoc] rustdoc/const-generics/const-impl.rs
    [rustdoc] rustdoc/const-generics/lazy_normalization_consts/const-equate-pred.rs
    [rustdoc] rustdoc/deref-mut-methods.rs
    [rustdoc] rustdoc/deref-recursive-pathbuf.rs
    [rustdoc] rustdoc/deref-recursive.rs
    [rustdoc] rustdoc/deref-typedef.rs
    [rustdoc] rustdoc/doc-cfg-traits.rs
    [rustdoc] rustdoc/empty-impls.rs
    [rustdoc] rustdoc/escape-deref-methods.rs
    [rustdoc] rustdoc/extern-default-method.rs
    [rustdoc] rustdoc/generic-impl.rs
    [rustdoc] rustdoc/inline_cross/impl-inline-without-trait.rs
    [rustdoc] rustdoc/inline_cross/issue-31948-2.rs
    [rustdoc] rustdoc/inline_cross/issue-32881.rs
    [rustdoc] rustdoc/inline_cross/trait-vis.rs
    [rustdoc] rustdoc/intra-doc/cross-crate/traits.rs
    [rustdoc] rustdoc/issue-15169.rs
    [rustdoc] rustdoc/issue-17476.rs
    [rustdoc] rustdoc/issue-19190-2.rs
    [rustdoc] rustdoc/issue-19190-3.rs
    [rustdoc] rustdoc/issue-19190.rs
    [rustdoc] rustdoc/issue-35169-2.rs
    [rustdoc] rustdoc/issue-35169.rs
    [rustdoc] rustdoc/issue-50159.rs
    [rustdoc] rustdoc/issue-51236.rs
    [rustdoc] rustdoc/issue-54705.rs
    [rustdoc] rustdoc/issue-55321.rs
    [rustdoc] rustdoc/issue-56822.rs
    [rustdoc] rustdoc/issue-60726.rs
    [rustdoc] rustdoc/issue-72340.rs
    [rustdoc] rustdoc/issue-80233-normalize-auto-trait.rs
    [rustdoc] rustdoc/issue-82465-asref-for-and-of-local.rs
    [rustdoc] rustdoc/issue-86620.rs
    [rustdoc] rustdoc/negative-impl-sidebar.rs
    [rustdoc] rustdoc/negative-impl.rs
    [rustdoc] rustdoc/primitive-generic-impl.rs
    [rustdoc] rustdoc/recursive-deref-sidebar.rs
    [rustdoc] rustdoc/recursive-deref.rs
    [rustdoc] rustdoc/spotlight-from-dependency.rs
    [rustdoc] rustdoc/src-links-auto-impls.rs
    [rustdoc] rustdoc/synthetic_auto/basic.rs
    [rustdoc] rustdoc/synthetic_auto/complex.rs
    [rustdoc] rustdoc/synthetic_auto/crate-local.rs
    [rustdoc] rustdoc/synthetic_auto/lifetimes.rs
    [rustdoc] rustdoc/synthetic_auto/manual.rs
    [rustdoc] rustdoc/synthetic_auto/negative.rs
    [rustdoc] rustdoc/synthetic_auto/nested.rs
    [rustdoc] rustdoc/synthetic_auto/no-redundancy.rs
    [rustdoc] rustdoc/synthetic_auto/overflow.rs
    [rustdoc] rustdoc/synthetic_auto/project.rs
    [rustdoc] rustdoc/synthetic_auto/self-referential.rs
    [rustdoc] rustdoc/synthetic_auto/static-region.rs
    [rustdoc] rustdoc/traits-in-bodies.rs

test result: FAILED. 423 passed; 56 failed; 3 ignored; 0 measured; 0 filtered out; finished in 37.61s
```

</details>

Still confused why this is necessary in the first place though :/

GuillaumeGomez:
My wild guess is that we need external traits for blanket impls. But that seems weird because that should be handled when we look for impl blocks. I'm really not sure why this is needed at this point... Maybe something we did before that we don't anymore?

jyn514:
Hmm, this looks incorrect:
```
DEBUG rustdoc::passes::collect_intra_doc_links resolving empty in DefId(20:0 ~ objr[b427])
INFO rustc_metadata::creader resolving crate `empty`
INFO rustc_metadata::creader falling back to a load
INFO rustc_metadata::locator rmeta reading metadata from: /home/joshua/test-rustdoc/rustc-84738/corevideor/target/debug/deps/libempty-c427b7c2497d53d1.rmeta
```
@petrochenkov shouldn't the crate not be loaded, because `empty` wasn't originally available in `objr`? This is cargo tree for context:
```
corevideor v0.1.0 (/home/joshua/test-rustdoc/rustc-84738/corevideor)
├── empty v0.1.0 (/home/joshua/test-rustdoc/rustc-84738/empty)
└── objr v0.2.0 (/home/joshua/test-rustdoc/rustc-84738/objr)
```

camelid:
@jyn514 FWIW, `RUSTDOC_LOG=rustc_metadata::creader=trace` may be enlightening since it will show what crates are already loaded, and why the already loaded crates can't be used.

petrochenkov:
Update: The logic in `register_res` was added in https://github.com/rust-lang/rust/pull/14346 back in 2014, removing it doesn't break any tests.

But the main question is: are `DocContext::external_traits` (and then `Cache::traits`) only used for documenting corresponding impls, or they are also used for documenting the traits themselves and maybe something else? I'll try to check, but only tomorrow.

camelid:
> Update: The logic in `register_res` was added in #14346 back in 2014, removing it doesn't break any tests.

FWIW, rustdoc's test coverage is not so great right now, so it's very possible it still breaks something. But if we _can_ safely remove that logic, it'd be great :)

petrochenkov:
>But the main question is: are `DocContext::external_traits` (and then `Cache::traits`) only used for documenting corresponding impls, or they are also used for documenting the traits themselves and maybe something else?

From what I can see, these tables are only used for traits from impls, so the tables should be removable in favor of inlining the `TraitWithExtraInfo` stuff directly into the `clean::Impl`.

apiraino:
backport suspended, waiting on #90489 ([Zulip discussion](https://rust-lang.zulipchat.com/#narrow/stream/238009-t-compiler.2Fmeetings/topic/.5Bweekly.5D.202021-11-04.20.2354818/near/260278752))

jyn514:
@apiraino I don't understand your comment - #90489 is this PR. My understanding from the meeting is they were just waiting for this PR to stop changing before approving a backport (which should be the case now; this isn't the sort of change I would expect to pass PR CI but fail bors).

apiraino:
Beta backport approved as per compiler team [on Zulip](https://rust-lang.zulipchat.com/#narrow/stream/238009-t-compiler.2Fmeetings/topic/.5Bweekly.5D.202021-11-11.20.2354818/near/261148043)

@rustbot label +beta-accepted

apiraino:
stable backport declined as per compiler team [on Zulip](https://rust-lang.zulipchat.com/#narrow/stream/238009-t-compiler.2Fmeetings/topic/.5Bweekly.5D.202021-11-18.20.2354818/near/261945258) (and previous comments)

@rustbot label -stable-nominated

Issue #31157, opened 2016-01-24
Massive compiler perf regression


Gankro:
[First reported on Reddit here](https://www.reddit.com/r/rust/comments/42cfly/huge_compilation_time_regression_from_jan_15_to/). See that post for more details and examples.

[This crate](https://github.com/White-Oak/rust-compilation-time-ram-regression) has gone from a build time of 17 seconds with the compiler consuming 168MB of RAM at peak to 320 secs and 1670MB. Apparently this regression occurred somewhere between jan 15-18, and is in the current beta. The author hypothesizes that it may have to do with complex usage of traits.

dirk:
Adding a comment from Reddit (see [here](https://www.reddit.com/r/rust/comments/42cfly/huge_compilation_time_regression_from_jan_15_to/cz9gagg)) so it doesn't get lost:

> Paging /u/arielby and /u/nikomatsakis, they always know what's going on in the type checker
>
> EDIT: Does the Jan 18 nightly contain the obligation forest (#30533), by any chance?

aturon:
cc @nikomatsakis @arielb1

White-Oak:
+1 on this. When talking about usage of complex traits (example, once again, [here](https://github.com/White-Oak/rust-compilation-time-ram-regression/blob/short-version/long_names.md)), it was always been slow, but on an acceptible rate. The author of library even goes about that in a [README.md](https://github.com/DanSimon/peruse/blob/master/README.md#other-notes):
> Be aware, due to an ongoing issue with rustc, the compile time of your code will exponentially increase with the complexity of your parsers. In practice I've found things get bad after about 10 combinations or so. You can get around this by boxing a parser...

The issue is https://github.com/rust-lang/rust/issues/22204. This (current) may be considered  duplicate of that, though, again, the prblem was not of the current magnitude.

White-Oak:
Ah, as well, this may be considered as a regression from stable to beta:
Code compiles rather quickly on stable 1.6, but slowly on beta 1.7. I just went further to find exact nightlies, which differ in compilation times.

gereeter:
Note that the obligation forest PR was merged on January 16, so it did get merged in the window in which the regression happened.

nikomatsakis:
Seems pretty likely to have to do with the obligation forest and also with the change in caching behavior that accompanied it. I can investigate some.

petevine:
[...]

PR #90490, opened 2021-11-01
Update cargo

ehuss:
3 commits in 6c1bc24b8b49d4bc965f67d7037906dc199c72b7..94ca096afbf25f670e76e07dca754fcfe27134be
2021-10-24 17:51:41 +0000 to 2021-10-29 14:45:06 +0000
- Chore: prefer `HashMap::from` rather than collecting `Vec` of tuples (rust-lang/cargo#10018)
- Change --scrape-examples flag to -Z rustdoc-scrape-examples (rust-lang/cargo#10017)
- Scrape code examples from examples/ directory for Rustdoc (rust-lang/cargo#9525)


pnkfelix:
Visiting for weekly performance triage

Only `regression-31157 opt` regressed, by 1.6-2.0%.

I don't think its worth investing effort trying to figure out the root cause of this limited regression.

(I will note that regression #31157 was itself put in to catch a performance regression with respect to time and RAM, but that regression was more like 1,882% in time and 10x in space.)

Issue #90445, opened 2021-10-31
Literally initializing `[&str; N]` (using proc macro) doesn't scale, `rustc` consumes huge amounts of time and RAM (stable and nightly)


twelho:
Hi Rust folks,

I'm using a procedural macro to read and preprocess some strings to be stored in an application without hitting the disk at any intermediate point. The macro outputs a `TokenStream` for a literal fixed-size array of string slices. When scaling up the amount of strings to a moderately high count (100k), I noticed something strange: compiling the application suddenly takes orders of magnitude more time and makes the compiler consumes almost 40 GiB of RAM in the process. Here's a MWE, for both the proc macro `lib` and the consumer binary `bin`:

`lib/Cargo.toml`:
```toml
[package]
name = "lib"
version = "0.1.0"
edition = "2021"

# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html

[lib]
proc-macro = true

[dependencies]
```

`lib/src/lib.rs`:
```rust
use proc_macro::TokenStream;

fn parse_count(stream: TokenStream) -> u32 {
    stream.to_string().parse().unwrap()
}

#[proc_macro]
pub fn generate_data(input: TokenStream) -> TokenStream {
    let mut s = String::from("[");
    for i in 0..parse_count(input) {
        if i > 0 {
            s.push_str(", ");
        }
        s.push_str(&format!("\"{}\"", i));
    }
    s.push_str("]");
    s.parse().unwrap()
}
```

`bin/Cargo.toml`:
```toml
[package]
name = "bin"
version = "0.1.0"
edition = "2021"

# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html

[dependencies]
lib = {path = "../lib"}
```

`bin/src/main.rs`:
```rust
use lib::generate_data;

fn main() {
    let data = generate_data!(100000);
    println!("{}", data.len());
}
```

Compiling the above takes many minutes, compared to just some seconds when the issue does not occur (see cases below). The memory consumption also peaks very high (truncated from the start, there's a slow increase over many minutes):

![image](https://user-images.githubusercontent.com/38858901/139585270-aba408a6-d4a1-4d2a-9b60-172b9966109b.png)

[...]

PR #90491, opened 2021-11-02
Optimize live point computation

Mark-Simulacrum:
This refactors the live-point computation to lower per-MIR-instruction costs by operating on a largely per-block level. This doesn't fundamentally change the number of operations necessary, but it greatly improves the practical performance by aggregating bit manipulation into ranges rather than single-bit; this scales much better with larger blocks.

On the benchmark provided in #90445, with 100,000 array elements, walltime for a check build is improved from 143 seconds to 15.

I consider the tiny losses here acceptable given the many small wins on real world benchmarks and large wins on stress tests. The new code scales much better, but on some subset of inputs the slightly higher constant overheads decrease performance somewhat. Overall though, this is expected to be a big win for pathological cases (as illustrated by the test case motivating this work) and largely not material for non-pathological cases. I consider the new code somewhat easier to follow, too.

### DIFF
--- a/compiler/rustc_borrowck/src/region_infer/values.rs
+++ b/compiler/rustc_borrowck/src/region_infer/values.rs
@@ -60,6 +60,11 @@ impl RegionValueElements {
         PointIndex::new(start_index)
     }
 
+    /// Return the PointIndex for the block start of this index.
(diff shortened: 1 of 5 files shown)


Mark-Simulacrum:
r? rust-lang/compiler

PR #90498, opened 2021-11-02
Clarifications in the target tier policy

joshtriplett:
We've added several targets since the introduction of the target tier policy. Based on experiences of those adding such targets, and discussions around such additions, clarify the target tier policy to make it easier to follow and work with.

None of these changes substantively change the requirements on targets. (In some cases the changes do direct target submitters to follow specific process requirements for the addition of a target, such as how to respond to requirements, where to put target-specific documentation, or what should appear in that documentation. Those changes are procedural in nature and document the procedures we already direct people to follow.)

- Clarify how to quote and respond to the target tier policy requirements. Several times, people have seemed unclear on how to respond to some of the policy requirements, particularly those that just state things the target developers must *not* do (e.g. not posting to PRs that break the target). Add a note that such requirements just need acknowledgement, nothing more.
- Clarify dependency requirements in the face of cross-compilation. I previously phrased this confusingly in terms of "host tools", since that is the case where an exception applies (allowing proprietary target libraries commonly used by binaries for the target). Rephrase it to apply equally to cross-compilation. This doesn't change the net effect of the requirements, since other requirements already cover the dependencies of the Rust toolchain.
- Clarify documentation about running binaries. The requirement for target documentation talks about "running tests", but tier 3 targets often don't support running the full testsuite, and in practice the documentation for how to run an individual binary may be more useful. Change "running tests" to "running binaries, or running tests".
- Explain where to place target-specific documentation (a subdirectory of platform-support, with a link from the platform-support entry for the target).
- Add a template for target-specific documentation.

### DIFF
--- a/src/doc/rustc/src/SUMMARY.md
+++ b/src/doc/rustc/src/SUMMARY.md
@@ -13,6 +13,7 @@
 - [JSON Output](json.md)
 - [Tests](tests/index.md)
 - [Platform Support](platform-support.md)
+    - [Template for target-specific documentation](platform-support/TEMPLATE.md)
     - [aarch64-apple-ios-sim](platform-support/aarch64-apple-ios-sim.md)
     - [\*-kmc-solid_\*](platform-support/kmc-solid.md)
 - [Target Tier Policy](target-tier-policy.md)
--- a/src/doc/rustc/src/platform-support/TEMPLATE.md
+++ b/src/doc/rustc/src/platform-support/TEMPLATE.md
@@ -1,0 +1,52 @@
+# `target-name-here`
+
+**Tier: 3**
+
+One-sentence description of the target (e.g. CPU, OS)
+
+## Target maintainers
+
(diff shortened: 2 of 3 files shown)


joshtriplett:
r? @pietroalbini

I don't think this needs a full three-team FCP like the original policy. It's documenting process that we're already following for the addition of targets, and clarifying corner cases and unclear points without substantively changing the requirements for targets.

All of this grew out of discussions around the additions of targets thus far, including unclear processes (where does documentation go), and corner cases (what about cross-compilation).

pietroalbini:
Can someone else review this? Won't have time myself.

Mark-Simulacrum:
I agree the changes in this PR do not imply/suggest any significant change or need full three-team FCP. I'm going to go ahead and r+, but also intend to announce this PR in a T-compiler meeting to bring attention to the new template in particular.

(I imagine there may be edits and/or suggestions to the text, but it seems clear that the text can be iterated on in the future).

@bors r+ rollup

Issue #90497, opened 2021-11-02
rustdoc `doc_auto_cfg` incorrectly picks up cfg's


Mark-Simulacrum:
*[View all comments](https://triagebot.infra.rust-lang.org/gh-comments/rust-lang/rust/issues/90497)*

**NOTE(camelid):** This bug is now feature-gated behind the `doc_auto_cfg` flag and, as such, is no longer a regression. The original issue text follows.

-----

For example, https://doc.rust-lang.org/nightly/std/primitive.usize.html#method.trailing_zeros has a tag indicating it is only supported on 64-bit, but in practice the method is defined on 32-bit as well; it's just defined separately from that definition.

This seems like it'll probably hurt particularly new users who may be annoyed at the lack of cross-target compatibility in such a method -- if we can't design a fix, I'd prefer that we disable the autosetting of cfgs for now, since it seems like this is likely to be widespread.

Mark-Simulacrum:
cc @rust-lang/rustdoc

jyn514:
@Mark-Simulacrum this doesn't need rustdoc changes, you can just remove feature(doc_cfg) from the standard library (and replace it with the manual cfgs from before). I don't think we should disable it for the rest of the ecosystem just because the standard library has some weird edge cases.

GuillaumeGomez:
It's not a regression from stable to beta but from nightly to nightly.

@jyn514 It's actually because of rustdoc here:

```rust
#[cfg(target_pointer_width = "32")]
#[lang = "usize"]
impl usize {
    widening_impl! { usize, u64, 32 }
    uint_impl! { usize, u32, isize, 32, 4294967295, 8, "0x10000b3", "0xb301", "0x12345678",
    "0x78563412", "0x1e6a2c48", "[0x78, 0x56, 0x34, 0x12]", "[0x12, 0x34, 0x56, 0x78]",
    usize_isize_to_xe_bytes_doc!(), usize_isize_from_xe_bytes_doc!() }
}

#[cfg(target_pointer_width = "64")]
#[lang = "usize"]
impl usize {
    widening_impl! { usize, u128, 64 }
    uint_impl! { usize, u64, isize, 64, 18446744073709551615, 12, "0xaa00000000006e1", "0x6e10aa",
    "0x1234567890123456", "0x5634129078563412", "0x6a2c48091e6a2c48",
    "[0x56, 0x34, 0x12, 0x90, 0x78, 0x56, 0x34, 0x12]",
[...]

PR #90502, opened 2021-11-02
Split doc_cfg and doc_auto_cfg features

GuillaumeGomez:
Part of #90497.

With this feature, `doc_cfg` won't pick up items automatically anymore.

cc @Mark-Simulacrum 
r? @jyn514

### DIFF
--- a/compiler/rustc_feature/src/active.rs
+++ b/compiler/rustc_feature/src/active.rs
@@ -688,6 +688,9 @@ declare_features! (
     /// not changed from prior instances of the same struct (RFC #2528)
     (incomplete, type_changing_struct_update, "1.58.0", Some(86555), None),
 
+    /// Tells rustdoc to automatically generate `#[doc(cfg(...))]`.
+    (active, doc_auto_cfg, "1.58.0", Some(43781), None),
+
     // -------------------------------------------------------------------------
     // feature-group-end: actual feature gates
     // -------------------------------------------------------------------------
--- a/src/librustdoc/clean/types.rs
+++ b/src/librustdoc/clean/types.rs
@@ -789,6 +789,7 @@ impl AttributesExt for [ast::Attribute] {
     fn cfg(&self, tcx: TyCtxt<'_>, hidden_cfg: &FxHashSet<Cfg>) -> Option<Arc<Cfg>> {
         let sess = tcx.sess;
         let doc_cfg_active = tcx.features().doc_cfg;
+        let doc_auto_cfg_active = tcx.features().doc_auto_cfg;
 
         fn single<T: IntoIterator>(it: T) -> Option<T::Item> {
             let mut iter = it.into_iter();
@@ -799,24 +800,26 @@ impl AttributesExt for [ast::Attribute] {
             Some(item)
         }
 
-        let mut cfg = if doc_cfg_active {
+        let mut cfg = if doc_cfg_active || doc_auto_cfg_active {
             let mut doc_cfg = self
                 .iter()
                 .filter(|attr| attr.has_name(sym::doc))
                 .flat_map(|attr| attr.meta_item_list().unwrap_or_else(Vec::new))
                 .filter(|attr| attr.has_name(sym::cfg))
                 .peekable();
-            if doc_cfg.peek().is_some() {
+            if doc_cfg.peek().is_some() && doc_cfg_active {
                 doc_cfg
                     .filter_map(|attr| Cfg::parse(attr.meta_item()?).ok())
                     .fold(Cfg::True, |cfg, new_cfg| cfg & new_cfg)
-            } else {
+            } else if doc_auto_cfg_active {
                 self.iter()
                     .filter(|attr| attr.has_name(sym::cfg))
                     .filter_map(|attr| single(attr.meta_item_list()?))
                     .filter_map(|attr| Cfg::parse(attr.meta_item()?).ok())
                     .filter(|cfg| !hidden_cfg.contains(cfg))
                     .fold(Cfg::True, |cfg, new_cfg| cfg & new_cfg)
+            } else {
+                Cfg::True
             }
         } else {
             Cfg::True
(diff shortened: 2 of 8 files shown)


compiler/rustc_feature/src/active.rs
```diff
@@ -688,6 +688,9 @@ declare_features! (
     /// not changed from prior instances of the same struct (RFC #2528)
     (incomplete, type_changing_struct_update, "1.58.0", Some(86555), None),
 
+    /// Allows rustdoc to automatically generate `#[doc(cfg(...))]`.
```
jyn514, line 691: "allows" sounds like you have to do some other step in addition to enabling the feature.
```suggestion
    /// Tells rustdoc to automatically generate `#[doc(cfg(...))]`.
```

src/librustdoc/clean/types.rs
```diff
@@ -810,13 +811,15 @@ impl AttributesExt for [ast::Attribute] {
                 doc_cfg
                     .filter_map(|attr| Cfg::parse(attr.meta_item()?).ok())
                     .fold(Cfg::True, |cfg, new_cfg| cfg & new_cfg)
-            } else {
+            } else if doc_auto_cfg_active {
```
jyn514, line 814: This only has an effect if `doc_cfg` is also active. I think it makes sense to enable them independently, but if you don't want to work on that right now, please make it a hard error to enable doc_auto_cfg without also enabling `doc_cfg`.

jyn514 (changes requested):

src/librustdoc/clean/types.rs
GuillaumeGomez: Good point. In addition to that, it's pretty easy to only enable `doc_auto_cfg` so it'd be a shame not to do it!

src/librustdoc/clean/types.rs
```diff
@@ -799,24 +800,26 @@ impl AttributesExt for [ast::Attribute] {
             Some(item)
         }
 
-        let mut cfg = if doc_cfg_active {
+        let mut cfg = if doc_cfg_active || doc_auto_cfg_active {
             let mut doc_cfg = self
```
jyn514, line 804: Do you mind moving this into the first `if` block to enforce that it only has an effect when the feature is enabled?

src/test/rustdoc-ui/feature-gate-doc_cfg_hide.rs
```diff
@@ -1,3 +1,4 @@
+// gate-test-doc_auto_cfg
```
jyn514, line 1: Please add a separate test showing that doc(cfg) isn't added unless the feature is enabled (preferably a revision test to make the difference easy to see).

src/librustdoc/clean/types.rs
jyn514: Please also add a test for that (adding doc(cfg) and doc_auto_cfg, but not doc_cfg).

src/test/rustdoc-ui/feature-gate-doc_cfg_hide.rs
GuillaumeGomez: I can but I added this here for the feature gate, not for testing `doc_auto_cfg`. ;)

src/librustdoc/clean/types.rs
GuillaumeGomez: Sure.

GuillaumeGomez: Actually, moving this into the `if` proved to be rather tricky because I still need an `Option` in the end to ensure that at least one `cfg` was found. So I tried to replace `fold` with `reduce` but then realized it would require the first item to be `Cfg::True`. So I'll simply keep the code as is.

jyn514: 👍 Seems fine as long as there's a test :)

GuillaumeGomez:
Ok, let's see if `tidy` is happy now. :)

src/test/rustdoc-ui/feature-gate-doc_auto_cfg.rs
```diff
@@ -0,0 +1,2 @@
+// check-pass
+pub fn foo() {}
```
jyn514, line 2: This doesn't actually do anything. Can you remove this test and rename `src/test/rustdoc/doc-not-auto-cfg.rs` to be the feature gate?

GuillaumeGomez: I thought that it needed to be a `ui` test. Makes things simpler if not!

jyn514:
@GuillaumeGomez do you mind updating tidy to also look in `src/test/rustdoc`? It already special cases rustdoc-ui, I don't think adding the new folder should be hard.

GuillaumeGomez:
Ah so that's what I thought. Adding a check in tidy then.

GuillaumeGomez:
Ah perfect, seems like CI passed this time! \o/

camelid:
This PR is beta-nominated. Does it make more sense to just disable `doc_cfg` on beta? cc @rust-lang/rustdoc for backport discussion

jyn514:
I don't think we should disable doc_cfg, no, the standard library uses it extensively. I don't think this is a terribly large PR to backport, and I'm nervous about applying a different patch to beta than to nightly.

apiraino:
Beta backport accepted as per compiler team [on Zulip](https://rust-lang.zulipchat.com/#narrow/stream/238009-t-compiler.2Fmeetings/topic/.5Bweekly.5D.202021-11-04.20.2354818/near/260279115)

@rustbot label +beta-accepted

compiler/rustc_feature/src/active.rs
```diff
@@ -688,6 +688,9 @@ declare_features! (
     /// not changed from prior instances of the same struct (RFC #2528)
     (incomplete, type_changing_struct_update, "1.58.0", Some(86555), None),
 
+    /// Tells rustdoc to automatically generate `#[doc(cfg(...))]`.
+    (active, doc_auto_cfg, "1.58.0", Some(43781), None),
```
cuviper, line 692: Since this was `beta-accepted`, I took the liberty of adjusting it to "1.57.0" in #90938. Would someone care to fix that on master as well?

PR #90505, opened 2021-11-02
Change `array_from_fn` signatures

c410-f3r:
Superseded by #91286

### Implementation

These functions are using the same inner code of other array methods and such feature already ensures a reliable execution.

### Signature

Originally, indices returned by the callback were introduced as auxiliary parameters since arrays are hereditary indexed and known at compile-time.

* https://github.com/PyO3/pyo3/blob/fc90092ef9090e034ff05323348da21d486735fa/src/conversions/array.rs#L63
* https://github.com/c410-f3r/ndsparse/blob/7c6803109aa1a8c68dcdb95c498c60feca011ca7/ndsparse/src/csl/csl_utils.rs#L84
* https://github.com/PistonDevelopers/hematite/blob/5e387561d7412b28f7b35977d35fa74b3cb1a677/src/minecraft/region.rs#L19
* https://github.com/LukasKalbertodt/lina/blob/3de7d87826528e17a03eb24ea690789e2535ac7a/src/vec/point.rs#L132

But due to https://github.com/rust-lang/rust/pull/75644#discussion_r689066495 and https://github.com/rust-lang/rust/pull/75644#issuecomment-939539568, no index will be returned to avoid further concerns or discussions.

### DIFF
--- a/library/core/src/array/mod.rs
+++ b/library/core/src/array/mod.rs
@@ -11,7 +11,7 @@ use crate::fmt;
 use crate::hash::{self, Hash};
 use crate::iter::TrustedLen;
 use crate::mem::{self, MaybeUninit};
-use crate::ops::{Index, IndexMut};
+use crate::ops::{ControlFlow, Index, IndexMut, Try};
 use crate::slice::{Iter, IterMut};
 
 mod equality;
@@ -24,37 +24,37 @@ pub use iter::IntoIter;
 ///
 /// # Arguments
 ///
-/// * `cb`: Callback where the passed argument is the current array index.
+/// * `cb`: Callback that must provide the array element
 ///
 /// # Example
 ///
 /// ```rust
 /// #![feature(array_from_fn)]
 ///
-/// let array = core::array::from_fn(|i| i);
+/// let mut idx: usize = 0;
+/// let array = core::array::from_fn(|| {
+///   let rslt = idx;
+///   idx += 1;
+///   rslt
+/// });
 /// assert_eq!(array, [0, 1, 2, 3, 4]);
 /// ```
 #[inline]
 #[unstable(feature = "array_from_fn", issue = "89379")]
 pub fn from_fn<F, T, const N: usize>(mut cb: F) -> [T; N]
 where
-    F: FnMut(usize) -> T,
+    F: FnMut() -> T,
 {
-    let mut idx = 0;
-    [(); N].map(|_| {
-        let res = cb(idx);
-        idx += 1;
-        res
-    })
+    [(); N].map(|_| cb())
 }
 
 /// Creates an array `[T; N]` where each fallible array element `T` is returned by the `cb` call.
-/// Unlike `core::array::from_fn`, where the element creation can't fail, this version will return an error
+/// Unlike [core::array::from_fn], where the element creation can't fail, this version will return an error
 /// if any element creation was unsuccessful.
 ///
 /// # Arguments
 ///
-/// * `cb`: Callback where the passed argument is the current array index.
+/// * `cb`: Callback that must provide the array element
 ///
 /// # Example
 ///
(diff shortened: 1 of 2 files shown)


c410-f3r:
cc @rust-lang/libs
r? @yaahc

library/core/src/array/mod.rs
```diff
@@ -29,23 +29,21 @@ pub use iter::IntoIter;
 /// # Example
 ///
 /// ```rust
-/// #![feature(array_from_fn)]
-///
-/// let array = core::array::from_fn(|i| i);
+/// let mut idx: usize = 0;
+/// let array = core::array::from_fn(|| {
+///   let rslt = idx;
+///   idx = idx.wrapping_add(1);
```
dtolnay, line 35: I find this wrapping_add distracting. Array sizes in Rust are `usize`, so any array has at most `usize::MAX` elements, so `from_fn` will run the closure at most `usize::MAX` times, which means incrementing `idx` from 0 at most `usize::MAX` times by 1, which does not overflow. Recommend `+=` instead.

dtolnay (changes requested):

CryZe:
I've seen that nowadays `try_` functions return `R: core::ops::Try<Output = T>` instead. I think `try_from_fn` should probably follow that. E.g. [`Iterator::try_for_each`](https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.try_for_each)

library/core/src/array/mod.rs
c410-f3r: Done

c410-f3r:
After a week without comments, I am **assuming** that using the `Try` trait is indeed the way to go.

dtolnay (changes requested):
I am not prepared to stabilize this instantly with the signature change. This needs to go into nightly with the new signature so that we can get a picture of whether this gives rise to inconvenient inference issues in practical use, or anything else that may come up.

c410-f3r:
> I am not prepared to stabilize this instantly with the signature change. This needs to go into nightly with the new signature so that we can get a picture of whether this gives rise to inconvenient inference issues in practical use, or anything else that may come up.

Sigh...

Done

scottmcm:
> whether this gives rise to inconvenient inference issues in practical use

The usual problem is that an inferred return type is not compatible with `?`.

I copied over the function signature here into the playground <https://play.rust-lang.org/?version=nightly&mode=debug&edition=2021&gist=0cdced05401eea1621827f86537ccd18>, and sure enough, inference failures:
```plain
error[E0284]: type annotations needed
  --> src/lib.rs:17:9
   |
17 |         try_from_fn(|| -> Result<i32, TryFromIntError> { i32::try_from(i) })?,
   |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ cannot infer type
   |
   = note: cannot satisfy `<_ as Try>::Residual == _`
```

And, importantly, that's happening despite the explicitly-specified closure type and a fully available return type (no try blocks or IIFEs or similar to confuse it).

That `?` usage, however, compiles fine with the current nightly `try_from_fn`: <https://play.rust-lang.org/?version=nightly&mode=debug&edition=2021&gist=b42ef33453354dce78f3299a1a40809a>

It also compiles fine with the approach in #91286: <https://play.rust-lang.org/?version=nightly&mode=debug&edition=2021&gist=3bc991d03ff52c972fc4c3b6dd02c46e>

(Sorry for not seeing this PR sooner.)

c410-f3r:
The suddenly and surprising depreciation of this PR over https://github.com/rust-lang/rust/pull/91286 was unexpected but I don't think anyone could do better than the very person who elaborated and implemented the new try infrastructure

Superseded by https://github.com/rust-lang/rust/pull/91286

scottmcm:
Thanks for continuing to push these forward, @c410-f3r -- they're critical building blocks and I'd love to see them in stable.

PR #90507, opened 2021-11-02
Suggest `extern crate alloc` when using undeclared module `alloc`

### LINKED ISSUE #90136: Suggest `extern crate alloc;` when using undeclared module `alloc`
Given the following code: [link](https://play.rust-lang.org/?version=nightly&mode=debug&edition=2018&gist=ae9e67981e817e2a887ea8a532b2868a)

```rust
use alloc::rc::Rc;
```

The current output is:

```
error[E0433]: failed to resolve: use of undeclared crate or module `alloc`
 --> src/lib.rs:2:5
  |
2 | use alloc::rc::Rc;
  |     ^^^^^ use of undeclared crate or module `alloc`

```

```
error[E0433]: failed to resolve: use of undeclared crate or module `alloc`
 --> src/lib.rs:2:5
  |
2 | use alloc::rc::Rc;
  |     ^^^^^ use of undeclared crate or module `alloc`
help: add `extern crate alloc` to use the buildin `alloc` module
```

TaKO8Ki:
closes #90136

### DIFF
--- a/compiler/rustc_resolve/src/lib.rs
+++ b/compiler/rustc_resolve/src/lib.rs
@@ -2527,7 +2527,7 @@ impl<'a> Resolver<'a> {
                                     Some((
                                         vec![],
                                         String::from(
-                                            "add `extern crate alloc` to use the builtin `alloc` module",
+                                            "add `extern crate alloc` to use the `alloc` crate",
                                         ),
                                         Applicability::MaybeIncorrect,
                                     ))
--- a/compiler/rustc_resolve/src/lib.rs
+++ b/compiler/rustc_resolve/src/lib.rs
@@ -2523,19 +2523,29 @@ impl<'a> Resolver<'a> {
                         } else {
                             (
                                 format!("use of undeclared crate or module `{}`", ident),
-                                self.find_similarly_named_module_or_crate(
-                                    ident.name,
-                                    &parent_scope.module,
-                                )
-                                .map(|sugg| {
-                                    (
-                                        vec![(ident.span, sugg.to_string())],
+                                if ident.name == sym::alloc {
+                                    Some((
+                                        vec![],
                                         String::from(
-                                            "there is a crate or module with a similar name",
+                                            "add `extern crate alloc` to use the builtin `alloc` module",
                                         ),
                                         Applicability::MaybeIncorrect,
+                                    ))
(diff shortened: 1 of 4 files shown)


compiler/rustc_resolve/src/lib.rs
```diff
@@ -2523,19 +2523,29 @@ impl<'a> Resolver<'a> {
...
-                                )
-                                .map(|sugg| {
-                                    (
-                                        vec![(ident.span, sugg.to_string())],
+                                if ident.name == sym::alloc {
+                                    Some((
+                                        vec![],
                                         String::from(
-                                            "there is a crate or module with a similar name",
+                                            "add `extern crate alloc` to use the builtin `alloc` module",
```
PatchMixolydic, line 2530:
```suggestion
                                            "add `extern crate alloc` to use the `alloc` crate",
```
I'm not sure if "builtin" is accurate since `alloc` is separate from the compiler (as far as I'm aware, at least). Alternative phrasings (aside from the one suggested here) might include "the standard `alloc` crate" or "the `alloc` crate in the sysroot".

TaKO8Ki: Thank you. I applied your suggestion.

joshtriplett:
This seems reasonable and helpful, as an improvement to the error message. :+1: for making this change.

But also, I wonder if we could arrange to always make `alloc` available, the way that `std` and `core` are always available, so that this error doesn't arise in the first place?

jackh726:
@bors r+ rollup

I think making alloc available like std or core would be nice, but should be a followup.

TaKO8Ki:
@jackh726 
Is there any particular reason why alloc is not available like std or core?

If not, should you add a new issue about making core available?

jackh726:
@TaKO8Ki I'm not sure :)

PatchMixolydic:
Pulling in `alloc` in a `no_std` binary seems like it requires a global memory allocator to be present ([example](https://play.rust-lang.org/?version=nightly&mode=debug&edition=2021&gist=513df6270e3ed3465498e57836c4ac1f)), which might be impossible in memory-constrained environments.

jackh726:
In no_std, yes. But when std is available, alloc should be, probably.

PR #90518, opened 2021-11-03
update rustc_ast crate descriptions in documentation

calebcartwright:
I noticed this the other day and figured I'd suggest a refresh. It seems like a relic from the days of `libsyntax` that got missed as things were split out into separate crates, since the current documentation text references elements that were moved into their own respective crates (e.g. `rustc_parse`)

### DIFF
--- a/compiler/rustc_ast/README.md
+++ b/compiler/rustc_ast/README.md
@@ -1,6 +1,5 @@
 The `rustc_ast` crate contains those things concerned purely with syntax
-– that is, the AST ("abstract syntax tree"), parser, pretty-printer,
-lexer, macro expander, and utilities for traversing ASTs.
+– that is, the AST ("abstract syntax tree"), along with some definitions for tokens and token streams, data structures/traits for mutating ASTs, and shared definitions for other AST-related parts of the compiler (like the lexer and macro-expansion).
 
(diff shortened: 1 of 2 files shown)


compiler/rustc_ast/README.md
```diff
@@ -1,6 +1,5 @@
 The `rustc_ast` crate contains those things concerned purely with syntax
-– that is, the AST ("abstract syntax tree"), parser, pretty-printer,
-lexer, macro expander, and utilities for traversing ASTs.
+– that is, the AST ("abstract syntax tree"), along with some definitions for tokens and token streams, data structures/traits for mutating ASTs, and shared definitions for other AST-related parts of the compiler (like the lexer and macro-expansion).
```
camelid, line 2: Hmm, I'm not even sure why this file exists. I thought all the `compiler/` READMEs were moved to the rustc-dev-guide.

calebcartwright: I'd assume that's because this gets published to crates.io as part of the auto publish process

camelid: Ah, I see 👍

wesleywiser (approved):

wesleywiser:
Thanks!

@bors r+ rollup

PR #90519, opened 2021-11-03
Keep spans for generics in `#[derive(_)]` desugaring

### LINKED ISSUE #84003: Misleading diagnostics with missing bounds when deriving a trait
Given the following code: https://play.rust-lang.org/?edition=2018&gist=5f8c9b8b72174c48746e236c827ab881

```rust
use std::fmt;

trait Trait {}

struct X<T>(T);

impl<T: fmt::Debug + Trait> fmt::Debug for X<T> {
    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
        f.debug_tuple("X").field(&self.0).finish()
    }
}

#[derive(Debug)]
struct Y<T>(X<T>);
```

The current output is:

```
error[E0277]: the trait bound `T: Trait` is not satisfied
  --> src/lib.rs:14:13
   |
14 | struct Y<T>(X<T>);
   |             ^^^^ the trait `Trait` is not implemented for `T`
   |
   = note: required because of the requirements on the impl of `Debug` for `X<T>`
   = note: 1 redundant requirements hidden
   = note: required because of the requirements on the impl of `Debug` for `&X<T>`
   = note: required for the cast to the object type `dyn Debug`
   = note: this error originates in a derive macro (in Nightly builds, run with -Z macro-backtrace for more info)
help: consider further restricting this bound
   |
13 | #[derive(Debug + Trait)]
   |                ^^^^^^^
```

The proposed solution is obviously wrong, the compiler should suggest a manual impl instead.

estebank:
Keep the spans for generics coming from a `derive`d Item, so that errors
and suggestions have better detail.

Fix #84003.

### DIFF
--- a/compiler/rustc_builtin_macros/src/deriving/generic/mod.rs
+++ b/compiler/rustc_builtin_macros/src/deriving/generic/mod.rs
@@ -609,28 +609,13 @@ impl<'a> TraitDef<'a> {
 
         // and similarly for where clauses
         where_clause.predicates.extend(generics.where_clause.predicates.iter().map(|clause| {
-            match *clause {
-                ast::WherePredicate::BoundPredicate(ref wb) => {
-                    ast::WherePredicate::BoundPredicate(ast::WhereBoundPredicate {
-                        span: wb.span,
-                        bound_generic_params: wb.bound_generic_params.clone(),
-                        bounded_ty: wb.bounded_ty.clone(),
-                        bounds: wb.bounds.to_vec(),
-                    })
-                }
-                ast::WherePredicate::RegionPredicate(ref rb) => {
-                    ast::WherePredicate::RegionPredicate(ast::WhereRegionPredicate {
-                        span: rb.span,
-                        lifetime: rb.lifetime,
-                        bounds: rb.bounds.to_vec(),
-                    })
-                }
-                ast::WherePredicate::EqPredicate(ref we) => {
+            match clause {
+                ast::WherePredicate::BoundPredicate(_)
+                | ast::WherePredicate::RegionPredicate(_) => clause.clone(),
+                ast::WherePredicate::EqPredicate(we) => {
                     ast::WherePredicate::EqPredicate(ast::WhereEqPredicate {
                         id: ast::DUMMY_NODE_ID,
-                        span: we.span,
-                        lhs_ty: we.lhs_ty.clone(),
-                        rhs_ty: we.rhs_ty.clone(),
+                        ..we.clone()
                     })
                 }
             }
--- a/compiler/rustc_builtin_macros/src/deriving/generic/mod.rs
+++ b/compiler/rustc_builtin_macros/src/deriving/generic/mod.rs
@@ -567,8 +567,10 @@ impl<'a> TraitDef<'a> {
             })
         });
 
-        let Generics { mut params, mut where_clause, span } =
+        let Generics { mut params, mut where_clause, .. } =
             self.generics.to_generics(cx, self.span, type_ident, generics);
+        where_clause.span = generics.where_clause.span;
+        let span = generics.span;
 
         // Create the generic parameters
         params.extend(generics.params.iter().map(|param| match &param.kind {
@@ -589,7 +591,7 @@ impl<'a> TraitDef<'a> {
                         param.bounds.iter().cloned()
                     ).collect();
 
-                cx.typaram(self.span, param.ident, vec![], bounds, None)
+                cx.typaram(param.ident.span, param.ident, vec![], bounds, None)
             }
             GenericParamKind::Const { ty, kw_span, .. } => {
                 let const_nodefault_kind = GenericParamKind::Const {
@@ -610,7 +612,7 @@ impl<'a> TraitDef<'a> {
             match *clause {
                 ast::WherePredicate::BoundPredicate(ref wb) => {
                     ast::WherePredicate::BoundPredicate(ast::WhereBoundPredicate {
-                        span: self.span,
+                        span: wb.span,
                         bound_generic_params: wb.bound_generic_params.clone(),
                         bounded_ty: wb.bounded_ty.clone(),
                         bounds: wb.bounds.to_vec(),
@@ -618,7 +620,7 @@ impl<'a> TraitDef<'a> {
                 }
                 ast::WherePredicate::RegionPredicate(ref rb) => {
                     ast::WherePredicate::RegionPredicate(ast::WhereRegionPredicate {
-                        span: self.span,
+                        span: rb.span,
                         lifetime: rb.lifetime,
                         bounds: rb.bounds.to_vec(),
                     })
@@ -626,7 +628,7 @@ impl<'a> TraitDef<'a> {
                 ast::WherePredicate::EqPredicate(ref we) => {
                     ast::WherePredicate::EqPredicate(ast::WhereEqPredicate {
                         id: ast::DUMMY_NODE_ID,
-                        span: self.span,
+                        span: we.span,
                         lhs_ty: we.lhs_ty.clone(),
                         rhs_ty: we.rhs_ty.clone(),
                     })
@@ -691,13 +693,13 @@ impl<'a> TraitDef<'a> {
             .iter()
             .map(|param| match param.kind {
                 GenericParamKind::Lifetime { .. } => {
-                    GenericArg::Lifetime(cx.lifetime(self.span, param.ident))
+                    GenericArg::Lifetime(cx.lifetime(param.ident.span, param.ident))
                 }
                 GenericParamKind::Type { .. } => {
-                    GenericArg::Type(cx.ty_ident(self.span, param.ident))
+                    GenericArg::Type(cx.ty_ident(param.ident.span, param.ident))
                 }
                 GenericParamKind::Const { .. } => {
-                    GenericArg::Const(cx.const_ident(self.span, param.ident))
+                    GenericArg::Const(cx.const_ident(param.ident.span, param.ident))
                 }
             })
             .collect();
@@ -1556,11 +1558,9 @@ impl<'a> TraitDef<'a> {
 
         let is_tuple = matches!(struct_def, ast::VariantData::Tuple(..));
         match (just_spans.is_empty(), named_idents.is_empty()) {
-            (false, false) => cx.span_bug(
-                self.span,
-                "a struct with named and unnamed \
-                                          fields in generic `derive`",
-            ),
+            (false, false) => {
+                cx.span_bug(self.span, "a struct with named and unnamed fields in generic `derive`")
+            }
             // named fields
             (_, false) => Named(named_idents),
             // unnamed fields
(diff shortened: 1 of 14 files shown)


compiler/rustc_builtin_macros/src/deriving/generic/ty.rs
```diff
@@ -245,7 +237,15 @@ impl Bounds {
...
 
-        mk_generics(generic_params, span)
+        Generics {
+            params,
+            where_clause: ast::WhereClause {
+                has_where_token: false,
+                predicates: Vec::new(),
+                span: self_generics.where_clause.span,
+            },
+            span: self_generics.span,
```
petrochenkov, line 247: This function is used for 1) building generics for generated impls and 2) building generics for methods inside these generated impls.
`self_generics` are generics on the derive's target ADT (struct or enum).
In the second case the spans have no relation to the ADT's generics spans, so these spans should either be passed from the outside of `to_generics`, or set outside of `to_generics`, like `params` and `predicates`.

src/test/ui/suggestions/derive-macro-missing-bounds.rs
```diff
@@ -0,0 +1,89 @@
...
+    struct Inner<T>(T);
+
+    impl Debug for Inner<()> {
+        fn fmt(&self, f: &mut Formatter<'_>) -> Result {
+            todo!()
+        }
+    }
+
+    #[derive(Debug)]
+    struct Outer<T>(Inner<T>); //~ ERROR E0277
```
petrochenkov, line 12:
```suggestion
    struct Outer<T>(Inner<T>); //~ ERROR `a::Inner<T>` doesn't implement `Debug`
```
Could you use a human readable annotations here and below?

compiler/rustc_hir/src/hir.rs
```diff
@@ -526,9 +525,17 @@ pub struct GenericParam<'hir> {
 impl GenericParam<'hir> {
     pub fn bounds_span(&self) -> Option<Span> {
         self.bounds.iter().fold(None, |span, bound| {
-            let span = span.map(|s| s.to(bound.span())).unwrap_or_else(|| bound.span());
-
-            Some(span)
+            if let ExpnKind::Macro(MacroKind::Derive, _) =
+                bound.span().ctxt().outer_expn_data().kind
+            {
+                // We ignore bounds that come from exclusively from a `#[derive(_)]`, because we
```
petrochenkov, line 531: `derive` is irrelevant, the problem is hard to concatenate spans from different syntactic contexts.

If you want this logic, please remove this change from this PR, and add the logic for concatenating a set of potentially incompatible spans to `rustc_span` in a separate PR.
The "concatenatable" subset should come from the same context, but which subset to use if there are multiple such subsets is a question. Using the root one (`SyntaxContext::empty()`) if it exists, would probably be a good start approximation.

estebank: This change is to avoid the following in the new test:

```
help: consider further restricting this bound
   |
LL |     #[derive(Debug + c::Trait)]
   |                    ++++++++++
```

I understand that we would want to check that all the spans belong to the same context and do something smarter than this, but we know that *this* will happen. `bounds_span` is used in three suggestions (add a type param when `_` used in a field or arg, suggest restricting an existing type param in middle and trait_selection). Because of that, I thought it appropriate to avoid emitting a completely wrong suggestion. I can also modify the caller and avoid calling `bounds_span`, but that seemed worse to me than having this "hidden logic" here. Thoughts?

compiler/rustc_builtin_macros/src/deriving/generic/ty.rs
estebank: Changed.

src/test/ui/suggestions/derive-macro-missing-bounds.rs
estebank: Changed.

compiler/rustc_hir/src/hir.rs
petrochenkov: Two questions:
- Is it possible to remove `fn bounds_span` entirely and use a multi-span or something else instead?
It looks like the full combined span isn't even used at the `bounds_span`'s points of use, it's always `shrink_to_hi`'ed before being reported.
- If `bounds_span` is kept, could you move the span-combining logic to a separate span-combining function in `rustc_span`. The `MacroKind::Derive` logic can be kept inside it, but it will need a big FIXME saying that the derive is actually irrelevant and the function needs to be rewritten to be more general. IIRC, `bounds_span` isn't the only place where such span combining is used.

compiler/rustc_builtin_macros/src/deriving/generic/mod.rs
```diff
@@ -607,28 +609,13 @@ impl<'a> TraitDef<'a> {
...
-                        bounds: rb.bounds.to_vec(),
-                    })
-                }
-                ast::WherePredicate::EqPredicate(ref we) => {
+            match clause {
+                ast::WherePredicate::BoundPredicate(_)
+                | ast::WherePredicate::RegionPredicate(_) => clause.clone(),
+                ast::WherePredicate::EqPredicate(we) => {
                     ast::WherePredicate::EqPredicate(ast::WhereEqPredicate {
                         id: ast::DUMMY_NODE_ID,
```
petrochenkov, line 617: I'm pretty sure that the `clause`'s ID is also dummy, so the predicates can be extended with just `generics.where_clause.predicates.iter().cloned()`.

petrochenkov: There's one more thing though - the person who previously wrote this clearly didn't know about hygiene.

If you generate code from a derive macro, you cannot reuse the input spans directly.
You can keep the locations, but you have to add one more macro layer to the hygiene parts of the spans.
This is done through the construction `inherited_span.with_ctxt(self.span)` or `self.span.with_span_pos(inherited_span)`, see e.g. `fn typaram`.
The existing code uses the correct hygiene (but with suboptimal position) by using just `self.span`, but only for the top level, all the nested spans are still cloned unmodified.
This PR makes things worse by cloning everything in where clauses unmodified, including the top level.
What is needed here is a small mutable visitor deeply updating all the spans in the original where clauses.

(Other cloned pieces of AST like generic parameters also need to be double-checked for this issue.)

estebank: I'll annotate all the spans I find with a new context (as appropriate).

compiler/rustc_hir/src/hir.rs
estebank: - `bounds_span` could return a `MultiSpan` instead, that would make more sense.
 - sounds reasonable.

estebank: Added `Span::can_be_used_for_suggestions` which checks that it's either on the Root context *or* it is a `derive` macro and the position is different to its parent span. This is only an approximation (that might be brittle), but when we use `self.span` in the compiler's `derive`s the parent span will be the same position. The problem happens is if the `ctxt` depth is higher than one (some macro producing another macro, or whatever a user supplied proc-macro might do), but it *seems* safe enough to me for merging. What do you think?

compiler/rustc_builtin_macros/src/deriving/generic/mod.rs
estebank: Changed. I left the cases where `trait_.span` is used under the assumption that those are already correctly pointing at the `derive` name `Symbol` with the appropriate context.

compiler/rustc_hir/src/hir.rs
```diff
@@ -524,12 +523,21 @@ pub struct GenericParam<'hir> {
...
-            Some(span)
-        })
+    pub fn bounds_span_for_suggestions(&self) -> Option<Span> {
+        self.bounds
+            .iter()
+            .fold(None, |span: Option<Span>, bound| {
+                // We include bounds that come from a `#[derive(_)]` but point at the user's code,
+                // as we use this method to get a span appropriate for suggestions.
+                // FIXME: a more "principled" approach should be taken here.
+                if !bound.span().can_be_used_for_suggestions() {
```
petrochenkov, line 533:
```
let bound_span = bound.span();
bound_span.can_be_used_for_suggestions().then(|| span.map(|s| s.to(bound_span)).unwrap_or(bound_span))
```

petrochenkov: Also the comment with FIXME is no longer relevant.

compiler/rustc_builtin_macros/src/deriving/generic/mod.rs
petrochenkov: The spans are still not updated deeply in the cloned where clauses, for example `wb.bounded_ty`, `we.(l,r)hs_ty` will keep the original spans.

>What is needed here is a small mutable visitor deeply updating all the spans in the original where clauses.

estebank: Got it. Are there other places where we should use the `MutVisitor` on? It almost feels like once I add this it would be better to yank out most of the span `with_ctxt` markings and replace with it a top level `visit_item` or similar call. Seems less error prone.

petrochenkov: Usually built-in derives generate entirely new code, not clone existing AST (except for tiny things like identifiers that don't need recursive visiting), so the `where` clause cloning case seems unique enough.

I'm not sure how to do all of this less error prone in centralized way, probably needs some experimentation.
Maybe generating an item first and then visiting it and adjusting all spans is the way to go, it only can cause issues if some parts of the output *don't* need to be adjusted, but with built-in derives it seems unlikely.

estebank: Could you take a look at the following draft https://github.com/rust-lang/rust/pull/91441?

I'm clearly doing something wrong, as I'm getting failures in the following tests:

```
failures:
    [ui] ui/derives/derive-hygiene.rs
    [ui] ui/hygiene/nested_macro_privacy.rs
    [ui] ui/proc-macro/issue-42708.rs
    [ui] ui/repr/repr-no-niche.rs
    [ui] ui/symbol-names/const-generics-structural-demangling.rs
```

related to proper hygiene when multiple levels of nesting occur. I've tried multiple things but all failed one way or another, so I wanted to check in with you as you might be able to spot some obvious thing I'm missing. I've also pushed the last WIP commit to this PR just in case you wanted to see what else I was doing, but the new draft has *only* the visitor change.

petrochenkov:
Blocked on https://github.com/rust-lang/rust/pull/91441.

petrochenkov:
https://github.com/rust-lang/rust/pull/91441#issuecomment-985181683, unblocking.

compiler/rustc_hir/src/hir.rs
estebank: removed

estebank:
Updated the PR without the MutVisitor.

Issue #77218, opened 2020-09-26
ICE on `if let Some() = ...` expression in loop with variable shadowing and missing let keyword


zdenek-crha:
### Code

```Rust
use std::collections::BTreeMap;

// Type aliases to replace complex data types used in the original code.
pub type Data = usize;
pub type Key = String;
pub type DataRef<'c> = &'c Data;

pub struct Storage(BTreeMap<Key, Data>);

impl Storage {
    pub fn list<'a>(&'a self, key: &Key) -> Vec<&Data> {
        unimplemented!()
    }
}

pub struct Cache<'c> {
    storage: Storage,
    cached_data_refs: BTreeMap<Key, DataRef<'c>>,
}

impl<'c> Cache<'c> {

    pub fn data_ref_list<'a>(&'a mut self, key: &Key) {
        for reference in self.storage.list(key) {
            self.cached_data_refs.insert(key.clone(), reference);

            // Following `if` statement is the reason for compiler crash. To trigger the crash it
            // is required to:
            //
            // * have `let` keyword missing from `if let Some(...)` expression
            // * have binding name in the `Some(...)` portion of the `if let Some()` shadow the
            //   name from the outer loop
            //
            if /*let*/ Some(reference) = self.cached_data_refs.get(key) {
                unimplemented!()
            }
        }
    }
}

fn main() {}
```

### Meta

`rustc --version --verbose`:
```
rustc 1.48.0-nightly (043f6d747 2020-09-25)
binary: rustc
commit-hash: 043f6d747c15068f0053a0542e9b0f17ae7f4de4
commit-date: 2020-09-25
host: x86_64-unknown-linux-gnu
release: 1.48.0-nightly
LLVM version: 11.0
```

### Error output

```
$ rustc bugtest.rs
error: internal compiler error: compiler/rustc_typeck/src/check/fn_ctxt.rs:459:25: while adjusting Expr { hir_id: HirI
d { owner: DefId(0:20 ~ bugtest[317d]::{{impl}}[1]::data_ref_list[0]), local_id: 70 }, kind: Path(Resolved(None, Path
{ span: bugtest.rs:34:42: 34:46 (#0), res: Local(HirId { owner: DefId(0:20 ~ bugtest[317d]::{{impl}}[1]::data_ref_list
[0]), local_id: 1 }), segments: [PathSegment { ident: self#0, hir_id: Some(HirId { owner: DefId(0:20 ~ bugtest[317d]::
[...]

PR #90521, opened 2021-11-03
Stabilize `destructuring_assignment`

### LINKED ISSUE #71126: Tracking Issue for `destructuring_assignment`

jhpratt:
Closes #71126

- [Stabilization report](https://github.com/rust-lang/rust/issues/71126#issuecomment-941148058)
- [Completed FCP](https://github.com/rust-lang/rust/issues/71126#issuecomment-954914819)

@rustbot label +F-destructuring-assignment +T-lang
Also needs +relnotes but I don't have permission to add that tag.


michaelwoerister:
r? rust-lang/lang

michaelwoerister:
r? rust-lang/lang-team

joshtriplett:
This looks correct to me.

matklad:
Do we have tests and overall plan regarding the interaction between destructing assignment and let else? With both features on track to stabilization, I feel we should figure out their interaction at least on some level *before* stabilization, least we realize we need some future-proofing somewhere *after* stabilization.

jhpratt:
I can't speak as to the status of tests and whatnot, but I would not think there's any interaction at all. After all, let-else inherently requires `let`, which is not present for destructing assignment.

matklad:
> After all, let-else inherently requires let, which is not present for destructing assignment.

I don’t think it inherently requires let. I can imagine the following working:

```rust
let mut x = …;
…
Some(x) = foo() else { return };
```

Anecdotally, couple of time I wanted to use let-else, I needed exactly this form, to update an existing mutable binding.

jhpratt:
I'll try to see if I can come up with some regex to find a relevant test. If not I'll just add a test myself.

Aloso:
The stabilization report doesn't mention destructing references. Is the plan to stabilize that later?

> ```rust
> let mut x = …;
> …
> Some(x) = foo() else { return };
> ```

@matklad This code is not accepted by the parser at the moment, but I don't see any reason why that couldn't be added later.

jhpratt:
After a quick check, confirming that the parser doesn't accept this. I don't see any potential for compatibility concerns; it seems reasonable that this would be accepted (both syntactically and semantically) in the future.

matklad:
I think I see at least some syntactic ambiguities of dangling-else variety:

```
let Some(x) = Some(y) = foo() else { return };
Some(x) = Some(y) = foo() else { return };
```

I don't think it makes sense to decide what are we going to do with them right now, but I do think that we should have an unresolved question somewhere, which at least outlines design space and good and poor interactions.

To clarify, the above is not an exhaustive list of potential problems, it's rather a counterexample to "it'll obviously just work together".

veber-alex:
I wanted to check how `(a,b) = (b,a)` compares to `mem::swap`, it looks like `mem::swap` applies optimizations when `T` is 32 bytes or bigger but destructuring_assignment doesn't.

https://godbolt.org/z/czfTv3nzW

jhpratt:
Let's keep this PR related to the stabilization and potential compatibility concerns. For optimization issues with `destructuring_assignment` it would probably be best to open a new, explicit issue.

jhpratt:
Rebased with CI passing.

jhpratt:
Rebased. Hopefully with a recent PR landing merge conflicts will be less likely from now on.

jackh726:
> How do docs look currently? Are the reference and book updated?

This is still an open question from me. If all docs are updated (or PRs ready for them), then I can review this.

jhpratt:
I have not personally looked at documentation. Not sure if anyone else has.

jackh726:
Okay, in that case, I'm going to defer to @rust-lang/lang whether we should block this on requiring PRs to docs (or maybe just issues opened).

varkor:
For what it's worth, I don't think there is that much that needs to be written to update the documentation, especially using the [RFC](https://github.com/rust-lang/rfcs/blob/master/text/2909-destructuring-assignment.md) for reference. (Unfortunately, I have very little time at the moment; otherwise I would do it myself.)

compiler/rustc_feature/src/accepted.rs
```diff
@@ -112,6 +112,8 @@ declare_features! (
     (accepted, default_type_params, "1.0.0", None, None),
     /// Allows `#[deprecated]` attribute.
     (accepted, deprecated, "1.9.0", Some(29935), None),
+    /// Allows the use of destructuring assignments.
+    (accepted, destructuring_assignment, "1.49.0", Some(71126), None),
```
jhpratt, line 116: Note to self: the version isn't correct here.

jhpratt:
This is still waiting on docs to my knowledge. I will resolve the trivial conflict once that is done.

varkor:
I've opened up a documentation pull request for the Reference (https://github.com/rust-lang/reference/pull/1116). But stabilisation is a blocker for the Reference pull request rather than the other way around. I think this is just waiting for @pnkfelix's review.

jhpratt:
Ah, alright. I will get to fixing the conflict soon. Realistically the conflict doesn't affect a review in any way, as it's just a change from the most recent bootstrap.

jhpratt:
Rebased.

varkor:
Personally, I wouldn't worry about continually rebasing: `feature_gate.rs` in particular is liable to frequent conflicts. As you say, it doesn't affect the review, and you can just rebase after approval.

jhpratt:
Yeah, I was just doing that so it _could_ be merged. Feature gates conflict much less now than they used to, as they're alphabetized instead of sorted by version. But I definitely intend to hold off until review is complete at this point.

src/test/ui/issues/issue-77218.rs
```diff
@@ -0,0 +1,9 @@
+fn main() {
```
jackh726, line 1: What is this test for?

Should probably be moved to a more specific directory than `issues`

jackh726 (approved):
One question/comment. But given that there has already been a completed FCP and doc PRs are ready, r=me when comment addressed

src/test/ui/issues/issue-77218.rs
jhpratt: It's a regression test for a previous ICE in #77218. The code is taken verbatim from [this comment](https://github.com/rust-lang/rust/issues/77218#issuecomment-702107741). Where should it be placed? The issues directory is typically used for ICE regression tests, no?

jackh726: Wait, I'm confused...this test [already exists](https://github.com/rust-lang/rust/blob/master/src/test/ui/issues/issue-77218/issue-77218.rs) and the associated issue is already closed? Why is this getting added?

jhpratt: Whoops. Just looked at the full diff and see that this test already exists. I probably just messed something up with tab autocomplete on the command line and never noticed.

I'll get this fixed tomorrow most likely and approve it on your review at that point (I think I'm able to do that, not positive).

jackh726:
@bors delegate+

Just r=jackh726 once you've removed the duplicate test :)

pnkfelix:
LGTM. Thanks @jhpratt ! (And sorry for how much time I've taken to look at this.)

jhpratt:
Thanks @pnkfelix. No worries about the delay; I was under the impression this was waiting on docs anyways.

matthiaskrgr:
@bors r-
I think a test needs an update?
https://github.com/rust-lang/rust/pull/91897#issuecomment-993269873

jhpratt:
Surprised that running `x t --bless` wasn't sufficient. Either way, I've done it on the clippy tests specifically. Let's give this another shot.

@bors r=jackh726,pnkfelix

PR #90529, opened 2021-11-03
Skip reborrows in AbstractConstBuilder

### LINKED ISSUE #90455: Combining generic_const_exprs with `adt_const_params` talks about derefs that don't exist in the code
Combining this with `adt_const_params` can give weird errors:
```rust
#![feature(generic_const_exprs, adt_const_params)]

struct FieldElement<const N: &'static str> {
    n: [u64; num_limbs(N)],
}
const fn num_limbs(_: &str) -> usize {
    0
}
```
Error:
```rust
error: overly complex generic constant
 --> src/lib.rs:5:14
  |
5 |     n: [u64; num_limbs(N)],
  |              ^^^^^^^^^^-^
  |                        |
  |                        dereferencing is not supported in generic constants
  |
  = help: consider moving this anonymous constant into a `const` function
  = note: this operation may be supported in the future
  ```
  https://play.rust-lang.org/?gist=a6a2aecfba83485968a81b8599b86cc5
  
  (I hoped I can try and play with these to see if it would be possible to implement typed algebraic fields)

_Originally posted by @elichai in https://github.com/rust-lang/rust/issues/76560#issuecomment-955777960_

### PR DESCRIPTION
**Pull request by b-naber**
Fixes https://github.com/rust-lang/rust/issues/90455

Temporary fix to prevent confusing diagnostics that refer to implicit borrows and derefs until we allow borrows and derefs on constant expressions.

r? @oli-obk

### DIFF
--- a/compiler/rustc_trait_selection/src/traits/const_evaluatable.rs
+++ b/compiler/rustc_trait_selection/src/traits/const_evaluatable.rs
@@ -399,13 +399,25 @@ impl<'a, 'tcx> AbstractConstBuilder<'a, 'tcx> {
                 let arg = self.recurse_build(source)?;
                 self.nodes.push(Node::Cast(abstract_const::CastKind::As, arg, node.ty))
             }
+            ExprKind::Borrow{ arg, ..} => {
+                let arg_node = &self.body.exprs[*arg];
 
+                // Skip reborrows for now until we allow Deref/Borrow/AddressOf
+                // expressions.
+                // FIXME(generic_const_exprs): Verify/explain why this is sound
+                if let ExprKind::Deref {arg} = arg_node.kind {
+                    self.recurse_build(arg)?
+                } else {
+                    self.maybe_supported_error(
+                        node.span,
+                        "borrowing is not supported in generic constants",
+                    )?
+                }
+            }
             // FIXME(generic_const_exprs): We may want to support these.
-            ExprKind::AddressOf { .. }
-            | ExprKind::Borrow { .. }
-            | ExprKind::Deref { .. } => self.maybe_supported_error(
+            ExprKind::AddressOf { .. } | ExprKind::Deref {..}=> self.maybe_supported_error(
                 node.span,
-                "dereferencing is not supported in generic constants",
+                "dereferencing or taking the address is not supported in generic constants",
             )?,
             ExprKind::Repeat { .. } | ExprKind::Array { .. } =>  self.maybe_supported_error(
                 node.span,
--- a/src/test/ui/const-generics/generic_const_exprs/closures.stderr
+++ b/src/test/ui/const-generics/generic_const_exprs/closures.stderr
@@ -4,7 +4,7 @@ error: overly complex generic constant
 LL | fn test<const N: usize>() -> [u8; N + (|| 42)()] {}
    |                                   ^^^^-------^^
    |                                       |
-   |                                       dereferencing is not supported in generic constants
+   |                                       borrowing is not supported in generic constants
    |
    = help: consider moving this anonymous constant into a `const` function
    = note: this operation may be supported in the future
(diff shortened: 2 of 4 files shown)

### CODE REVIEW

**Comment by oli-obk** [comment]
cc @BoxyUwU @lcnr 

This is just about reborrows, right? I think I'd prefer to treat Rvalue::Ref(Place::Deref(_x)) as the same as a use of _x, but there may be implications I'm missing

**Comment by oli-obk** [comment]
Unless we never want these adjustments in general as they may just be noise

@bors try @rust-timer queue

File: compiler/rustc_mir_build/src/thir/cx/expr.rs [hunk 1]
```diff
@@ -93,6 +94,20 @@ impl<'tcx> Cx<'tcx> {
     ) -> Expr<'tcx> {
         let Expr { temp_lifetime, .. } = expr;
 
+        // FIXME(generic_const_exprs): Once expressions of kind `AdressOf`,
+        // `Borrow` and `Deref` are supported, we can allow adjustments here
+        if let ExprKind::Literal {
```
lcnr, line 99: I would assume that `&*expr` is used for borrowck reasons :thinking: 

While I am not too familiar with this code, this impl seems worrying to me because:
- in places where we only have one of the adjustments, i.e. either `&CONST` or `*CONST` this is incorrect. Don't know if something like that is possible rn but it seems likely to me.
- this probably causes some subtle changes to borrowck, though considering that const params can currently only be `&'static Ty` this is probably not too much of a problem :shrug:

b-naber: I completely misunderstood the issue, I initially thought that we would reject these borrows and derefs anyway, and this seems to be the case inside anonymous const blocks when `generic_const_exprs` is  used, where we always throw an error when we encounter them during the construction of the `AbstractConst` ([here](https://github.com/rust-lang/rust/blob/18fa4342fc215fe9e714307db694eaa8f5dc4a0d/compiler/rustc_trait_selection/src/traits/const_evaluatable.rs#L403-L409)), but treating all Borrow and Deref exprs as nops on all constants in general is not what we want obviously. 

If we do want to allow reborrows, then the current implementation of raising the bug during the construction of the `AbstractConst` seems to actually be what we want here right? And we should just keep this until we allow Borrows and Derefs inside anonymous constant.

b-naber: Maybe we should try to give a better error message here that makes the implicit adjustments explicit to the user?

b-naber:
@oli-obk and @lcnr 

I changed this PR, we now don't apply reborrow adjustments whenever we're inside an anonconst context. I think this is what we want to do here, since we would reject these adjustments during the const evaluatable checks anyway.

I'm not sure whether solving this issue warrants the changes I had to make, introducing an expression check context solely for this might be considered unnecessary. Also I'm not sure whether this could result in performance problems, we would definitely need another timer run here.

Also I would clean this up a little, just wanted to see whether the idea in general would be viable.

oli-obk:
> This is just about reborrows, right? I think I'd prefer to treat Rvalue::Ref(Place::Deref(_x)) as the same as a use of _x, but there may be implications I'm missing 

This is still true. For a temporary fix this PR adds way too much logic that needs to go away again. Can't we "just" make the abstract-const builder handle this case instead?

b-naber:
@oli-obk Implemented skipping re-borrows during AbstractConst construction. This does work, although I'm not completely sure whether this is correct here, we would unify reborrowed expressions with expressions that don't contain a reborrow. I think this should be fine, but maybe I'm missing something. Could this be a problem, @lcnr?

lcnr:
I am fine with skipping reborrows, maybe add a `FIXME(generic_const_exprs): Verify/explain why this is sound` to the abstract const builder, than r=me on that change by itself. This PR still contains the previous changes, does it? Can these now be removed?

PR #90530, opened 2021-11-03
Simplify js tester a bit

GuillaumeGomez:
r? @notriddle

### DIFF
--- a/src/tools/rustdoc-js/tester.js
+++ b/src/tools/rustdoc-js/tester.js
@@ -414,7 +414,7 @@ function parseOptions(args) {
         "test_folder": "",
         "test_file": "",
     };
-    var correspondances = {
+    var correspondences = {
         "--resource-suffix": "resource_suffix",
         "--doc-folder": "doc_folder",
         "--test-folder": "test_folder",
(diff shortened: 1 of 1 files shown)


src/tools/rustdoc-js/tester.js
```diff
@@ -423,11 +423,7 @@ function parseOptions(args) {
     };
 
     for (var i = 0; i < args.length; ++i) {
-        if (args[i] === "--resource-suffix"
-            || args[i] === "--doc-folder"
-            || args[i] === "--test-folder"
-            || args[i] === "--test-file"
-            || args[i] === "--crate-name") {
+        if (correspondances[args[i]] !== undefined) {
```
notriddle, line 426:
```suggestion
        if (Object.hasOwn(correspondances, args[i])) {
```

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/hasOwn

camelid: Also, a nit: "correspondances" should be spelled "correspondences". While you're at it, do you want to fix the spelling?

GuillaumeGomez: Sure!

GuillaumeGomez:
Updated. TIL thanks to both of you. :)

GuillaumeGomez:
@notriddle I didn't check the requirements but they're quite high in fact! I'll revert to `hasOwnProperty`.

GuillaumeGomez:
So apparently, `Object.hasOwnProperty` isn't a thing, but `obj.hasOwnProperty` is. Oh well, JS.

notriddle (approved):

PR #90533, opened 2021-11-03
Add note about x86 instruction prefixes in asm! to unstable book

### LINKED ISSUE #82314: x86 `asm!` accepts "lock" as a standalone instruction
```rust
#![feature(asm)]

#[no_mangle]
pub fn aaaaaaaaaaaaaaaaaa(x: &mut u32) {
    unsafe { asm!("lock"); }
    *x += 1;
}
```

This is not rejected, and, according to objdump, results in the following assembly (when optimizations are on):

```
0000000000000000 <aaaaaaaaaaaaaaaaaa>:
   0:	48 83 ec 08          	sub    $0x8,%rsp
   4:	f0 83 07 01          	lock addl $0x1,(%rdi)
   8:	58                   	pop    %rax
   9:	c3                   	ret    
```

I'd expect this to result in an error, since `lock` by itself isn't a valid instruction (it doesn't really seem desirable to let `asm!()` affect how rustc-produced instructions are decoded).

syvb:
Since rustc doesn't do the assembly parsing itself, it is unable to detect when inline assembly ends with an instruction prefix, which doesn't make sense since it would apply to instructions from the compiler. This fixes #82314 by mentioning that x86 instruction prefixes must not be used in inline assembly. AFAICT x86 is the only instruction set with instruction prefixes.


src/doc/unstable-book/src/library-features/asm.md
```diff
@@ -885,5 +885,7 @@ The compiler performs some additional checks on options:
     - You are responsible for switching any target-specific state (e.g. thread-local storage, stack bounds).
     - The set of memory locations that you may access is the intersection of those allowed by the `asm!` blocks you entered and exited.
 - You cannot assume that an `asm!` block will appear exactly once in the output binary. The compiler is allowed to instantiate multiple copies of the `asm!` block, for example when the function containing it is inlined in multiple places.
+- On x86, inline assembly must not end with an instruction prefix (such as `LOCK`) that would apply to instructions generated by the compiler.
+  - The compiler is currently unable to detect this due to the way inline assembly is compiled.
```
joshtriplett, line 889:
```suggestion
  - The compiler is currently unable to detect this due to the way inline assembly is compiled, but may catch and reject this in the future.
```

joshtriplett:
Made one suggestion to further clarify this.

Also, thank you for qualifying this as "that would apply to instructions generated by the compiler". There are valid reasons to have a prefix separated from an instruction for other purposes, such as for patching instructions.

Separately from this PR, would you please submit a report to upstream LLVM, requesting that its assembler catch trailing instruction prefixes like this?

Smittyvb:
LLVM bug: https://bugs.llvm.org/show_bug.cgi?id=52405

PR #90544, opened 2021-11-03
Demote metadata load warning to "info".

### LINKED ISSUE #90525: "rustc_metadata::locator no metadata found: no `.rustc` section" warnings during compilation
When compiling our project, many warnings regarding "no .rustc section" are printed.

To reproduce, please check out https://github.com/farmaazon/bug-repro , go to `src/rust` directory and run `cargo build`.
You should see some warning similar to 
```
WARN rustc_metadata::locator no metadata found: no `.rustc` section in '<path-to-repo>/src/rust/target/debug/deps/libenso_shortcuts_examples.so'
```

As far as I checked, the warnings are not displayed when building crates separately. 

### Meta

`rustc --version --verbose`:
```
rustc 1.58.0-nightly (c390d69a6 2021-10-28)
binary: rustc
commit-hash: c390d69a615f095208ac94841f3310268521b2ee
commit-date: 2021-10-28
release: 1.58.0-nightly
LLVM version: 13.0.0
```
The repository contains rust-toolchain file.

ehuss:
There is a warn log message for whenever the crate loader fails to load metadata from a candidate file. I think this warning is too aggressive, as there are several situations where metadata information might not be found in a candidate file, which is normal. Also, this warning is somewhat confusing, and non-actionable in most cases for a user (most users will not know what it means). 

If the crate loader ultimately does not find a valid crate, then an error will be reported (and hopefully #88368 will improve that error message). 

If a rustc developer wants to debug a loader problem, they can still use `RUSTC_LOG=rustc_metadata=debug` and get the details.

There is more discussion of this particular warning at https://github.com/rust-lang/rust/issues/89795#issuecomment-940798190.

Fixes #90525

### DIFF
--- a/compiler/rustc_metadata/src/locator.rs
+++ b/compiler/rustc_metadata/src/locator.rs
@@ -236,7 +236,7 @@ use std::fmt::Write as _;
 use std::io::{Read, Result as IoResult, Write};
 use std::path::{Path, PathBuf};
 use std::{cmp, fmt, fs};
-use tracing::{debug, info, warn};
+use tracing::{debug, info};
 
 #[derive(Clone)]
 crate struct CrateLocator<'a> {
@@ -549,7 +549,7 @@ impl<'a> CrateLocator<'a> {
                         }
                     }
                     Err(err) => {
-                        warn!("no metadata found: {}", err);
+                        info!("no metadata found: {}", err);
                         continue;
                     }
                 };


petrochenkov:
bjorn3 wanted these warnings to be available without rebuilding rustc, but apparently we don't have a mechanism that allows the warnings to be disabled by default, but enabled by e.g. an environment variable.
I think it's more important to not show these warnings by default, so let's choose requiring a rebuild for rustc developers.
@bors r+ rollup

ehuss:
Oh, to be clear, `info` level messages are available in the default `rustc` build (no need to rebuild). It just requires setting `RUSTC_LOG`, whereas `warn` is shown without the environment variable.

It is `debug` and `trace` level messages that require rebuilding.

PR #90545, opened 2021-11-03
Implement a faster stable sort algorithm

wpwoodjr:
Hi everyone, this is my first PR.

This PR replaces slice's [TimSort](https://doc.rust-lang.org/std/primitive.slice.html#method.sort) with a stable two stage merge sort having pre-sorted prefix optimization.  The total running time is O(n * log(n)) worst-case.

**Highlights:**

- Improved performance, especially on random and pre-sorted inputs.
- Improved performance on other patterns such as shuffle and stagger.
- Performs fewer comparisons on random and most other patterns.

**Benchmark:**
The benchmark was run on six OS's / CPUs:

- Ubuntu 20.04.3 LTS / Intel(R) Core(TM) i7-8559U CPU @ 2.70GHz
- WIN 10 KVM on Ubuntu / Intel(R) Core(TM) i7-8559U CPU @ 2.70GHz
- macOS Big Sur Version 11.6 / MacBook Pro (16-inch, 2019) 2.4 GHz 8-Core Intel Core i9
- Debian Crostini Linux / Google PixelBook Intel(R) 7th Gen Core(TM) i7-7Y75
- Debian Crostini Linux / Asus Chromebook Flip CX5 Intel(R) 11th Gen Core(TM) i5-1130G7
- Debian Crostini Linux / HP Chromebook X2 11 Qualcomm(R) Snapdragon(TM) 7c

For each OS/CPU, the benchmark was run on vecs of `i16`, `i32`, `i64`, `i128`, and `String`. The vecs were of random lengths from 0 to 10,000,000.

The proposed two stage merge sort indicates a speedup of **13.1%** for random unsorted data, with **95%** of sort runs being faster. For random data, including unsorted and forward / reverse sorted variants, the results indicate a speedup of **20.2%**, with **92%** of sort runs being faster. Other data patterns are faster too, except for the sawtooth pattern, which is about the same. The number of comparisons performed is **-2.92%** less over all tests.

**Note:** The above benchmark figures have been corrected since my original post.  The original post did not include sort results for `vecs` of length 0-100.

Full results and benchmark code are at [wpwoodjr/rust-merge-sort](https://github.com/wpwoodjr/rust-merge-sort)

**The new sort**
While not really new, the proposed sort benefits from:

- Top-down recursive depth-first merge, which helps data locality
- Small slices sort using a fast insertion sort, then merge
- Pre-sorted prefix optimization for forward and reverse sorted sub-slices

The new sort builds on existing code wherever possible.

### DIFF
--- a/library/alloc/src/slice.rs
+++ b/library/alloc/src/slice.rs
@@ -889,10 +889,9 @@ fn insert_end<T, F>(v: &mut [T], is_less: &mut F)
 where
     F: FnMut(&T, &T) -> bool,
 {
-    debug_assert!(v.len() >= 2);
-    let end = v.len() - 1;
-    unsafe {
-        if is_less(v.get_unchecked(end), v.get_unchecked(end - 1)) {
+    let end = v.len().saturating_sub(1);
+    if end > 0 && is_less(&v[end], &v[end - 1]) {
+        unsafe {
             // There are three ways to implement insertion here:
             //
             // 1. Swap adjacent elements until the last one gets to its final destination.
@@ -957,8 +956,9 @@ where
 ///
 /// The two slices must be non-empty and `mid` must be in bounds. Buffer `buf` must be long enough
 /// to hold a copy of the shorter slice. Also, `T` must not be a zero-sized type.
+#[allow(unused_unsafe)]
 #[cfg(not(no_global_oom_handling))]
-fn merge<T, F>(v: &mut [T], mid: usize, buf: *mut T, is_less: &mut F)
+unsafe fn merge<T, F>(v: &mut [T], mid: usize, buf: *mut T, is_less: &mut F)
 where
     F: FnMut(&T, &T) -> bool,
 {
@@ -1038,14 +1038,14 @@ where
     // it will now be copied into the hole in `v`.
 
     #[inline(always)]
-    fn get_and_increment<T>(ptr: &mut *mut T) -> *mut T {
+    unsafe fn get_and_increment<T>(ptr: &mut *mut T) -> *mut T {
         let old = *ptr;
         *ptr = unsafe { ptr.offset(1) };
         old
     }
 
     #[inline(always)]
-    fn decrement_and_get<T>(ptr: &mut *mut T) -> *mut T {
+    unsafe fn decrement_and_get<T>(ptr: &mut *mut T) -> *mut T {
         *ptr = unsafe { ptr.offset(-1) };
         *ptr
     }
@@ -1186,7 +1186,9 @@ where
                     }
                     mid = SMALL_SLICE_LEN;
                 }
-                merge(v, mid, buf_ptr, is_less);
+                unsafe {
+                    merge(v, mid, buf_ptr, is_less);
+                }
             }
         }
     }
@@ -1196,9 +1198,11 @@ where
     /// # Safety
     ///
     /// `buf_ptr` must point to a slice of `buf` which is long enough to hold `v[mid..]`
-    /// v.len() >= mid because `swap_slices` is only called when: sorted < len && mid == sorted.max(len / 2)
+    ///   `v[mid..].len() <= buf.len()` because buf is sized as `(v.len() + 1) / 2' and `mid == sorted.max(v.len() / 2)`
+    ///
+    /// `mid` must be <= `v.len()`
+    ///   `mid <= v.len()` because `swap_slices` is only called when: `sorted < len && mid == sorted.max(len / 2)`
     fn swap_slices<T>(v: &mut [T], mid: usize, buf_ptr: *mut T) {
-        debug_assert!(v.len() >= mid);
         let rlen = v.len() - mid;
         let v_ptr = v.as_mut_ptr();
         unsafe {
--- a/library/alloc/src/slice.rs
+++ b/library/alloc/src/slice.rs
@@ -249,10 +249,10 @@ impl<T> [T] {
     ///
     /// # Current implementation
     ///
-    /// The current algorithm is an adaptive, iterative merge sort inspired by
-    /// [timsort](https://en.wikipedia.org/wiki/Timsort).
-    /// It is designed to be very fast in cases where the slice is nearly sorted, or consists of
-    /// two or more sorted sequences concatenated one after another.
+    /// This is a stable two stage merge sort with pre-sorted prefix optimization.
+    /// It is quite fast in cases where the slice is nearly sorted, or consists of
+    /// two or more sorted sequences concatenated one after another, while remaining very fast for
+    /// randomly sorted sequences.
     ///
     /// Also, it allocates temporary storage half the size of `self`, but for short slices a
     /// non-allocating insertion sort is used instead.
@@ -302,10 +302,10 @@ impl<T> [T] {
     ///
     /// # Current implementation
     ///
-    /// The current algorithm is an adaptive, iterative merge sort inspired by
-    /// [timsort](https://en.wikipedia.org/wiki/Timsort).
-    /// It is designed to be very fast in cases where the slice is nearly sorted, or consists of
-    /// two or more sorted sequences concatenated one after another.
+    /// This is a stable two stage merge sort with pre-sorted prefix optimization.
+    /// It is quite fast in cases where the slice is nearly sorted, or consists of
+    /// two or more sorted sequences concatenated one after another, while remaining very fast for
+    /// randomly sorted sequences.
     ///
     /// Also, it allocates temporary storage half the size of `self`, but for short slices a
     /// non-allocating insertion sort is used instead.
@@ -879,34 +879,38 @@ impl<T: Clone> ToOwned for [T] {
 // Sorting
 ////////////////////////////////////////////////////////////////////////////////
 
-/// Inserts `v[0]` into pre-sorted sequence `v[1..]` so that whole `v[..]` becomes sorted.
+/// Inserts `v[v.len() - 1]` into pre-sorted sequence `v[..v.len() - 1]` so that whole `v[..]` becomes sorted.
 ///
 /// This is the integral subroutine of insertion sort.
 #[cfg(not(no_global_oom_handling))]
-fn insert_head<T, F>(v: &mut [T], is_less: &mut F)
+// benchmarking indicated that inlining makes a substantial improvement, yet only requires a couple of hundred bytes
+#[inline(always)]
+fn insert_end<T, F>(v: &mut [T], is_less: &mut F)
 where
     F: FnMut(&T, &T) -> bool,
 {
-    if v.len() >= 2 && is_less(&v[1], &v[0]) {
-        unsafe {
+    debug_assert!(v.len() >= 2);
+    let end = v.len() - 1;
+    unsafe {
+        if is_less(v.get_unchecked(end), v.get_unchecked(end - 1)) {
             // There are three ways to implement insertion here:
             //
-            // 1. Swap adjacent elements until the first one gets to its final destination.
+            // 1. Swap adjacent elements until the last one gets to its final destination.
             //    However, this way we copy data around more than is necessary. If elements are big
             //    structures (costly to copy), this method will be slow.
             //
-            // 2. Iterate until the right place for the first element is found. Then shift the
-            //    elements succeeding it to make room for it and finally place it into the
+            // 2. Iterate until the right place for the last element is found. Then shift the
+            //    elements preceeding it to make room for it and finally place it into the
             //    remaining hole. This is a good method.
             //
-            // 3. Copy the first element into a temporary variable. Iterate until the right place
+            // 3. Copy the last element into a temporary variable. Iterate until the right place
             //    for it is found. As we go along, copy every traversed element into the slot
-            //    preceding it. Finally, copy data from the temporary variable into the remaining
+            //    succeeding it. Finally, copy data from the temporary variable into the remaining
             //    hole. This method is very good. Benchmarks demonstrated slightly better
             //    performance than with the 2nd method.
             //
             // All methods were benchmarked, and the 3rd showed best results. So we chose that one.
-            let tmp = mem::ManuallyDrop::new(ptr::read(&v[0]));
+            let tmp = mem::ManuallyDrop::new(ptr::read(v.get_unchecked(end)));
 
             // Intermediate state of the insertion process is always tracked by `hole`, which
             // serves two purposes:
@@ -918,15 +922,14 @@ where
             // If `is_less` panics at any point during the process, `hole` will get dropped and
             // fill the hole in `v` with `tmp`, thus ensuring that `v` still holds every object it
             // initially held exactly once.
-            let mut hole = InsertionHole { src: &*tmp, dest: &mut v[1] };
-            ptr::copy_nonoverlapping(&v[1], &mut v[0], 1);
+            let mut hole = InsertionHole { src: &*tmp, dest: v.get_unchecked_mut(end - 1) };
+            ptr::copy_nonoverlapping(hole.dest, v.get_unchecked_mut(end), 1);
 
-            for i in 2..v.len() {
-                if !is_less(&v[i], &*tmp) {
-                    break;
-                }
-                ptr::copy_nonoverlapping(&v[i], &mut v[i - 1], 1);
-                hole.dest = &mut v[i];
+            let mut i = end - 1;
+            while i > 0 && is_less(&*tmp, v.get_unchecked(i - 1)) {
+                hole.dest = v.get_unchecked_mut(i - 1);
+                ptr::copy_nonoverlapping(hole.dest, v.get_unchecked_mut(i), 1);
+                i -= 1;
             }
             // `hole` gets dropped and thus copies `tmp` into the remaining hole in `v`.
         }
@@ -955,7 +958,7 @@ where
 /// The two slices must be non-empty and `mid` must be in bounds. Buffer `buf` must be long enough
 /// to hold a copy of the shorter slice. Also, `T` must not be a zero-sized type.
 #[cfg(not(no_global_oom_handling))]
-unsafe fn merge<T, F>(v: &mut [T], mid: usize, buf: *mut T, is_less: &mut F)
+fn merge<T, F>(v: &mut [T], mid: usize, buf: *mut T, is_less: &mut F)
 where
     F: FnMut(&T, &T) -> bool,
 {
@@ -1034,13 +1037,15 @@ where
     // Finally, `hole` gets dropped. If the shorter run was not fully consumed, whatever remains of
     // it will now be copied into the hole in `v`.
 
-    unsafe fn get_and_increment<T>(ptr: &mut *mut T) -> *mut T {
+    #[inline(always)]
+    fn get_and_increment<T>(ptr: &mut *mut T) -> *mut T {
         let old = *ptr;
         *ptr = unsafe { ptr.offset(1) };
         old
     }
 
-    unsafe fn decrement_and_get<T>(ptr: &mut *mut T) -> *mut T {
+    #[inline(always)]
+    fn decrement_and_get<T>(ptr: &mut *mut T) -> *mut T {
         *ptr = unsafe { ptr.offset(-1) };
         *ptr
     }
@@ -1063,27 +1068,38 @@ where
     }
 }
 
-/// This merge sort borrows some (but not all) ideas from TimSort, which is described in detail
-/// [here](https://github.com/python/cpython/blob/main/Objects/listsort.txt).
+/// This is a stable two stage merge sort with pre-sorted prefix optimization. The two stages are:
 ///
-/// The algorithm identifies strictly descending and non-descending subsequences, which are called
-/// natural runs. There is a stack of pending runs yet to be merged. Each newly found run is pushed
-/// onto the stack, and then some pairs of adjacent runs are merged until these two invariants are
-/// satisfied:
+/// 1) Top-down recursive depth-first merge, which helps data locality
+/// 2) Small slices sort using a fast insertion sort, then merge
 ///
-/// 1. for every `i` in `1..runs.len()`: `runs[i - 1].len > runs[i].len`
-/// 2. for every `i` in `2..runs.len()`: `runs[i - 2].len > runs[i - 1].len + runs[i].len`
-///
-/// The invariants ensure that the total running time is *O*(*n* \* log(*n*)) worst-case.
+/// The total running time is *O*(*n* \* log(*n*)) worst-case.
 #[cfg(not(no_global_oom_handling))]
 fn merge_sort<T, F>(v: &mut [T], mut is_less: F)
 where
     F: FnMut(&T, &T) -> bool,
 {
+    // `gt!` macro centralizes and clarifies the logic.
+    // Unchecked array access gives an approximate 20% performance improvement.
+    //
+    // # Safety
+    //
+    // `$left` and `$right` must be < `$v.len()`
+    macro_rules! gt {
+        ($v: ident, $left: expr, $right: expr, $is_less: ident) => {
+            // $is_less(&$v[$right], &$v[$left])
+            {
+                debug_assert!($left < $v.len() && $right < $v.len());
+                $is_less(unsafe { &$v.get_unchecked($right) }, unsafe { &$v.get_unchecked($left) })
+            }
+        };
+    }
+
+    // Benchmarking determined these are the best sizes.
+    // Recursive merge switches to insertion sort / merge when slice length is <= SMALL_SLICE_LEN*2.
+    const SMALL_SLICE_LEN: usize = 10;
     // Slices of up to this length get sorted using insertion sort.
     const MAX_INSERTION: usize = 20;
-    // Very short runs are extended using insertion sort to span at least this many elements.
-    const MIN_RUN: usize = 10;
 
     // Sorting has no meaningful behavior on zero-sized types.
     if size_of::<T>() == 0 {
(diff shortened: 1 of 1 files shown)


library/alloc/src/slice.rs
```diff
@@ -861,34 +861,37 @@ impl<T: Clone> ToOwned for [T] {
...
 ////////////////////////////////////////////////////////////////////////////////
 
-/// Inserts `v[0]` into pre-sorted sequence `v[1..]` so that whole `v[..]` becomes sorted.
+/// Inserts `v[v.len() - 1]` into pre-sorted sequence `v[..v.len() - 1]` so that whole `v[..]` becomes sorted.
 ///
 /// This is the integral subroutine of insertion sort.
 #[cfg(not(no_global_oom_handling))]
-fn insert_head<T, F>(v: &mut [T], is_less: &mut F)
+// benchmarking indicated that inlining makes a substantial improvement, yet only requires a couple of hundred bytes
+#[inline(always)]
```
sanmai-NL, line 869: Why not just `#[inline]`?

library/alloc/src/slice.rs
```diff
@@ -861,34 +861,37 @@ impl<T: Clone> ToOwned for [T] {
...
+fn insert_end<T, F>(v: &mut [T], is_less: &mut F)
 where
     F: FnMut(&T, &T) -> bool,
 {
-    if v.len() >= 2 && is_less(&v[1], &v[0]) {
+    let end = v.len() - 1;
```
sanmai-NL, line 874: `saturating_sub` would prevent underflow.

library/alloc/src/slice.rs
```diff
@@ -900,23 +903,22 @@ where
...
-            for i in 2..v.len() {
-                if !is_less(&v[i], &*tmp) {
-                    break;
-                }
-                ptr::copy_nonoverlapping(&v[i], &mut v[i - 1], 1);
-                hole.dest = &mut v[i];
+            let mut hole = InsertionHole { src: &*tmp, dest: v.get_unchecked_mut(end - 1) };
+            ptr::copy_nonoverlapping(hole.dest, v.get_unchecked_mut(end), 1);
+
+            let mut i = end - 1;
```
sanmai-NL, line 909: `NonZeroUsize` + `while let` would express the semantics better.

library/alloc/src/slice.rs
```diff
@@ -1036,149 +1040,150 @@ where
 
     impl<T> Drop for MergeHole<T> {
         fn drop(&mut self) {
-            // `T` is not a zero-sized type, so it's okay to divide by its size.
-            let len = (self.end as usize - self.start as usize) / mem::size_of::<T>();
             unsafe {
+                // use slice method for length calculation
+                let len = self.end.offset_from(self.start) as usize;
```
sanmai-NL, line 1045: Does this line need to move inside the `unsafe` block?

library/alloc/src/slice.rs
```diff
@@ -1036,149 +1040,150 @@ where
...
-        let mut start = end - 1;
-        if start > 0 {
-            start -= 1;
-            unsafe {
-                if is_less(v.get_unchecked(start + 1), v.get_unchecked(start)) {
-                    while start > 0 && is_less(v.get_unchecked(start), v.get_unchecked(start - 1)) {
-                        start -= 1;
+    // `is_less` panics. When merging two slices, this buffer holds a copy of the right-hand slice,
+    // which will always have length at most `(len + 1) / 2`.
+    let mut buf = Vec::with_capacity((len + 1) / 2);
```
sanmai-NL, line 1103: Can this become an array rather than a `Vec`?

library/alloc/src/slice.rs
wpwoodjr: Yes, otherwise:
image

library/alloc/src/slice.rs
wpwoodjr: No, because it needs to be allocated with a length based on the length of the vec to be sorted.

library/alloc/src/slice.rs
sanmai-NL: Yes, but was it strictly needed to change to `offset_from`?

library/alloc/src/slice.rs
sanmai-NL: Using const generics?

hkratz: But the length of the slice to be sorted is not const.

library/alloc/src/slice.rs
wpwoodjr: Not strictly needed, but I think it is better to use a slice method instead of "rolling your own" as it was before:
```
            // `T` is not a zero-sized type, so it's okay to divide by its size.
            let len = (self.end as usize - self.start as usize) / mem::size_of::<T>();
```

library/alloc/src/slice.rs
wpwoodjr: Good catch.  v.len() must be >= 2 for insert_end() to be called.  I will add a debug_assert:
```
    debug_assert!(v.len() >= 2);
```

library/alloc/src/slice.rs
wpwoodjr: I don't understand.  `i` can be `0`

library/alloc/src/slice.rs
wpwoodjr: I don't think we can guarantee the performance improvements with just `#[inline]`, as the compiler may decide not to inline it.

library/alloc/src/slice.rs
sanmai-NL: I was thinking of doing the capacity computation as a constant expression, but I see [this isn't supported](https://rust-lang.github.io/rfcs/2000-const-generics.html#when-a-const-variable-can-be-used) (yet).

sanmai-NL: If you call `merge_sort` on with a slice of 18446744073709551615 elements, this will overflow.  

```suggestion
    let mut buf = Vec::with_capacity((len / 2) + 1);
```
should fix that.

sanmai-NL: If you call `merge_sort` on with a slice of 18446744073709551615 elements,`len` will overflow.  

```suggestion
    let mut buf = Vec::with_capacity((len / 2) + 1);
```
should fix that.

wpwoodjr: The suggestion would change the semantics... `(10 + 1) / 2` != `(10 / 2) + 1`.  Performance is sensitive to how this is calculated.

> If you call merge_sort on with a slice of 18446744073709551615 elements,len will overflow.

But it is not possible to create a slice that big.

library/alloc/src/slice.rs
```diff
@@ -1016,13 +1021,15 @@ where
...
+    #[inline(always)]
+    fn get_and_increment<T>(ptr: &mut *mut T) -> *mut T {
         let old = *ptr;
         *ptr = unsafe { ptr.offset(1) };
         old
     }
 
-    unsafe fn decrement_and_get<T>(ptr: &mut *mut T) -> *mut T {
+    #[inline(always)]
+    fn decrement_and_get<T>(ptr: &mut *mut T) -> *mut T {
```
fbstj, line 1030: I believe all these were unsafe for a reason; without that you can call them with arguments that can be unsound without any warning that you've done so.

library/alloc/src/slice.rs
sanmai-NL: Do you think `saturating_sub` would degrade performance? Debug assertions are off generally, do I misunderstand how your fix prevents the underflow?

library/alloc/src/slice.rs
sanmai-NL: If it's zero, the loop will not be entered as usual (it would be `None`).

library/alloc/src/slice.rs
wpwoodjr: I was getting this warning so I removed it:
image

library/alloc/src/slice.rs
wpwoodjr: If it did saturate that would be a bug.  Underflow might be better because it definitely would panic when indexing `v[end]`.  `insert_end` is never called unless `v.len() >= 2`, the debug_assert is there to indicate that, it is not possible for insert_end() to be called when `v.len() < 2`.

library/alloc/src/slice.rs
wpwoodjr: I see...  I think stylistically it is very clear as written, the NonZeroUsize adds additional cognitive overhead?

I'm not clear on the performance implications and also it would require a nightly feature:
image

library/alloc/src/slice.rs
sanmai-NL: This could show up as possible integer overflow using analysis tools though, and why not prevent that.

#88581 hasn't landed yet.
Can you use [`saturating_add`](https://doc.rust-lang.org/std/primitive.usize.html#method.saturating_add) instead?

(Don't know if there is some API to e.g., represent memory a mapped file as an extremely large slice or so, or some extreme supercomputer in the future, but the practicality of such a slice wasn't my point.)

library/alloc/src/slice.rs
```diff
@@ -1036,149 +1040,150 @@ where
...
-            if n >= 3 && runs[n - 3].len < runs[n - 1].len { Some(n - 3) } else { Some(n - 2) }
-        } else {
-            None
+    /// swap contents of left-hand and right-hand slices divided at `mid`
+    ///
+    /// # Safety
+    ///
+    /// `buf_ptr` must point to a slice of `buf` which is long enough to hold `v[mid..]`
+    fn swap_slices<T>(v: &mut [T], mid: usize, buf_ptr: *mut T) {
+        let rlen = v.len() - mid;
```
sanmai-NL, line 1182: Overflow risk, see previous comments.

library/alloc/src/slice.rs
```diff
@@ -1036,149 +1040,150 @@ where
...
-        {
-            if n >= 3 && runs[n - 3].len < runs[n - 1].len { Some(n - 3) } else { Some(n - 2) }
-        } else {
-            None
+    /// swap contents of left-hand and right-hand slices divided at `mid`
+    ///
+    /// # Safety
+    ///
+    /// `buf_ptr` must point to a slice of `buf` which is long enough to hold `v[mid..]`
+    fn swap_slices<T>(v: &mut [T], mid: usize, buf_ptr: *mut T) {
```
sanmai-NL, line 1181: Can't you use safe code for this? 
I haven't studied it thoroughly, but would [`swap_with_slice`](https://doc.rust-lang.org/std/primitive.slice.html#method.swap_with_slice) work?

library/alloc/src/slice.rs
sanmai-NL: I'm not so sure about that, actually. This is the Rust standard library. The source code doesn't have to be pretty if it can't be, but safety very important. Every unsafe statement adds to the risk of a defect creeping in.

sanmai-NL:
Before even considering performance improvements, which I personally find very important, I do hope the Rust maintaincers get convincing fuzz test results, and possibly applying AddressSanitizer and MemorySanitizer.

Furthermore, can the insertion sort be factored out or reused? Maybe precomputed sorting networks would be even better.

library/alloc/src/slice.rs
wpwoodjr: I believe the purpose of unsafe functions is to warn the potential user of the api provided by the function.  Since these functions are not meant as APIs but are utility functions local to the sort, I had decided to remove the `unsafe` **on** the function rather that remove the `unsafe` **in** the function.  Leaving it in the function makes it cleared exactly what is unsafe.  Am I wrong to do that?

library/alloc/src/slice.rs
wpwoodjr: `saturating_add` would create a slice that is too short by `1`.  However it is not possible to create a slice that is the size of memory, and then allocate an additional slice that is 1/2 that size.  If an extreme supercomputer existed that could represent this, then the max value of `usize` would also have to increase.

the8472:
The git history contains merge commits, you'll have to rebase that.

https://rustc-dev-guide.rust-lang.org/contributing.html#opening-a-pr

library/alloc/src/slice.rs
wpwoodjr: `swap_with_slice` only works on slices of equal size.

library/alloc/src/slice.rs
wpwoodjr: Can you explain please?  `mid` is always < `v.len()`.

wpwoodjr:
> Before even considering performance improvements, which I personally find very important, I do hope the Rust maintaincers get convincing fuzz test results, and possibly applying AddressSanitizer and MemorySanitizer.


The benchmarking code includes a fuzz test of sorts.  Pass `eq` and the number of runs:
```
$ target/release/newsort eq --nruns 40
Std: Std, Newsort: Newsort
Range 10 to 10000
String array size = 1095696
Running sort equality test with 40 runs...
20 equality test runs completed
40 equality test runs completed; new to standard comparisons ratio: 0.9848
```
Each run sorts over 900 randomly sized arrays between 0 and 100,000 in length.

> Furthermore, can the insertion sort be factored out or reused? Maybe precomputed sorting networks would be even better.
I started out with that idea too, but insertion sort was the fastes.

library/alloc/src/slice.rs
fbstj: it depends, it may be more important to indicate the unsafety where these helpers are used rather than what parts of the helpers are unsafe.

library/alloc/src/slice.rs
sanmai-NL: It depends on what you call the function with, doesn't it? `v` can be empty, `mid` can be incorrect. Similar to your `debug_assert!()` elsewhere, you may at minimum want one assertion here. But I personally simply wouldn't use unchecked arithmetic.

library/alloc/src/slice.rs
sanmai-NL: To be clear, the change I proposed as fallback should calculate an equal capacity unless overflow occurs:

```suggestion
    let mut buf = Vec::with_capacity(len.saturating_add(1) / 2);
```

If you mean this change would prevent an overflow but leave `buf` possibly one element to small, then you may be right (the wording is that 'at most' this scratch space is needed, if we can prove this isn't so in this corner case, the argument would not hold). 

You would avoid underallocating with my alternate formula, that overallocates one element for even sizes of `self`. Your scratch buffer procedure cannot work at `len == 18446744073709551615` anyway (under 64 bits word size, but mind lower word sizes and memory mapped files), since as you write, the rest of the algorithm requires more than half of that size in space. 

Alternatively, you may simply store one overflow scratch element in a variable, and keep the scratch space at a neat `self.len() / 2` length.

I do not know exactly the guaranteed properties of slices other than being 'contiguous'. But in my mind, a sparse bit vector could be a slice with 18446744073709551615 elements but take far fewer bytes in storage. So the range of `usize` would not have to increase in that case. I just hope to encourage you to think about this, since this is fundamental standard library function for which we cannot make assumptions.

By the way, the (current and original) wording in the API docs that it allocates 'half' the size of `self` in scratch space is subtly incorrect, then.

library/alloc/src/slice.rs
sanmai-NL: Looking at it, it seems `mid` should really be the middle element. 
So am I right to believe that both subslices would be of equal sizes of off-by-one?
You can then swap the single overflow element manually, no?

library/alloc/src/slice.rs
sanmai-NL: You are right about the nightliness of addition, I forgot about that.
Checked addition is slower and I wouldn't advise that here, but I thought saturating addition isn't necessarily, by the way.

Yeah, I see how some programmers will find `NonZeroUsize` odd as it stands out. On the other hand, some will like the correctness by construction aspect and informativeness.

library/alloc/src/slice.rs
sanmai-NL: Never called with those arguments, but that's not a precondition checked in this function (outside debug mode).
It was originally, on this line. If you perform fewer checks then that may improve performance, but people can recognize (part of) the performance gain is due to shortcuts taken rather than actual improvements. I'm not claiming you must not do that, but in general, I would err on the side of caution, especially around unsafe source code in the standard library. (There are other sorting libraries available.)

wpwoodjr:
> The git history contains merge commits, you'll have to rebase that.
> 
> https://rustc-dev-guide.rust-lang.org/contributing.html#opening-a-pr

Thanks.  I'm not sure how to do that.  I have my fork locally.  I think I made a mistake by fetching upstream changes between my commits.  How do I fix it?

the8472:
Rebasing automatically removes merge commits.

```
git remote add upstream git@github.com:rust-lang/rust.git # only if you haven't added it already
git fetch upstream
git rebase upstream/master # this will prompt you to resolve conflicts if needed
git push -f
```

Add `-i` to reorder, reword or squash individual commits

```
git rebase -i
```

library/alloc/src/slice.rs
fee1-dead: Just use any nightly features if you want. The Standard Library is always built by the latest nightly compiler.

library/alloc/src/slice.rs
wpwoodjr: I've inserted a debug_assert! and comments explaining why mid is always <= v.len().

library/alloc/src/slice.rs
wpwoodjr: No, mid can be much greater than the middle element if the slice is pre-sorted past the middle element.

library/alloc/src/slice.rs
lambda-fairy: > But it is not possible to create a slice that big.

With zero-sized types it is.

```rust
fn main() {
    [(); usize::MAX];  // works
}
```

the8472: That's already prevented

```rust
    // Sorting has no meaningful behavior on zero-sized types.
    if size_of::<T>() == 0 {
        return;
    }
```

library/alloc/src/slice.rs
wpwoodjr: Thanks.  I believe it is best to leave the `while i > 0` loop as is.  `NonZeroUsize` would require more code and wouldn't necessarily be clearer.

library/alloc/src/slice.rs
wpwoodjr: I prefer to use the standard slice method.  The original code relies on knowledge of the underlying layout/representation.

library/alloc/src/slice.rs
wpwoodjr: I've removed the debug_assert and provided detailed comments on why the call to swap_slices is safe in this context.

library/alloc/src/slice.rs
wpwoodjr: I've put the `unsafe`s back.

library/alloc/src/slice.rs
wpwoodjr: I've added saturating_sub as requested, with no performance penalty.

wpwoodjr:
Hi @kennytm, what are the next steps for this PR please?  I think I've addressed the comments.  Thanks!

sanmai-NL:
Do you have test results to support your claim this source code is safe? You didn't address that part, in my opinion.  A randomized test within modest size ranges is not a fuzz test. I advise you to specifically search for panics, and use said sanitizers.

In terms of performance, the results are mixed. See https://github.com/rust-lang/rust/pull/90545#issuecomment-961255348.
Suppose you do obtain some overall improvement with a few further tweaks/the new results, which I hope for. You expanded unsafe source code in this branch, and whether any such improvements are because of that or a slightly better algorithm, isn't clear to me.

wpwoodjr:
@kennytm how can I re-run the performance regression tests?  I tried adding the label `+S-waiting-on-perf` but that didn't seem to work.

the8472:
Someone with permission has to start them.

@bors try @rust-timer queue

Dylan-DPC:
@wpwoodjr if you can resolve the conflict we can push this forward

wpwoodjr:
> @wpwoodjr if you can resolve the conflict we can push this forward

Thanks, I am on vacation but will look into it when I return March 10th.  Which conflict in particular?

library/alloc/src/slice.rs
wpwoodjr: > Is this then more a [`rotate`](https://doc.rust-lang.org/nightly/std/primitive.slice.html#method.rotate_left) than a `swap`? A name change might clear things up. I think this is one of the `with_buffer` rotation algorithms, per page 186 of http://elementsofprogramming.com/eop.pdf

Good point, it can be looked at as a large rotate.  Generally, unless the slice is pre-sorted beyond the middle of the slice, `mid` will be in the middle of the slice, so I think of it as a swap of left and right sides of the slice, rather than a large rotate.  Does that make sense?

Dylan-DPC:
`library/alloc/src/slice.rs` is currently conflicting after some changes were merged into master (a rebase should do the trick). No worries, enjoy your vacation and feel free to take your time to get back at it :)

JohnCSimon:
Ping again from triage:
@wpwoodjr if you can resolve the conflict we can push this forward

wpwoodjr:
> Ping again from triage: @wpwoodjr if you can resolve the conflict we can push this forward

OK will take a look thanks

tstj:
The following points could be interesting to readers of the source code, if you know the answer to some of them, it would be nice to add them as comments to the source code:

- the maximim recursion depth of `slice_merge_sort` (if it's `log(n)` do you know the base?)
- does it still recognize already sorted runs?
   if so,
     - does it still reverse reversed runs?
     - is it still [`k`-aware](https://arxiv.org/pdf/1809.08411.pdf) and if so, for which `k`?
- the maximum length difference of two slices that get merged
- the name of this sorting method (or is it a new one?)
- the base of the logarithm of the worst case (not always easy to say)
- do you know a worst case input that with the new method is much slower (like two times slower or more) than with the old method?

Thanks!

wpwoodjr:
Thanks for the questions.  I thought I'd respond here before adding any comments to the source code.
- the maximum recursion depth of slice_merge_sort (if it's log(n) do you know the base?)
  - Its a merge sort, so basically log(n) base 2.  However it resorts to insertion sort for n <= 20, so max recursion depth is log(n) - 4.
- does it still recognize already sorted runs?
  - Yes, it detects sorted prefixes
- does it still reverse reversed runs?
  - Yes, it detects reverse sorted prefixes and reverses them
- is it still [k-aware](https://arxiv.org/pdf/1809.08411.pdf) and if so, for which k?
  - This is a merge sort, it doesn't maintain a stack of sorted runs (from the reference, _A deterministic
sorting algorithm is said to be (k, ℓ)-aware (or simply k-aware if k = ℓ) if it sorts arrays of data by
manipulating a stack of (sorted) runs_)
- the maximum length difference of two slices that get merged
  - The mid-point for merging is determined by `mid = sorted.max(len / 2)`, so in the extreme case mid could be `len - 1` if the slice is sorted in the first `len - 1` elements.
- the name of this sorting method (or is it a new one?)
  - It's a merge sort with pre-sorted prefix optimization.  I'm guessing that prior art exists :)
- the base of the logarithm of the worst case runtime (not always easy to say)
  - At its heart it's a merge sort so O(n log n), base 2
- do you know a best/worst case input that with the new method is much faster/slower (like two times faster/slower or more) than with the old method?
  - It is never more than a few percent slower. In a few cases it can be >100% faster.  See [results](https://github.com/wpwoodjr/rust-merge-sort#results).

wpwoodjr:
@Dylan-DPC @JohnCSimon I've resolved the conflicts.  Interestingly the resolved changes were similar to changes I had made to remove unneeded `mut`s and improve pointer arithmetic.

What are the next steps?

Dylan-DPC:
r? rust-lang/t-libs

Dylan-DPC:
r? rust-lang/libs

thomcc:
Hm, this looks fine to me, and I've done it out on paper a few times to see how it works, but my unfamiliarity with the questions and answers https://github.com/rust-lang/rust/pull/90545#issuecomment-1126945161 betray that I'm probably not the best reviewer for this, I think someone more familiar with the theory behind sorting algorithms might be better.

r? rust-lang/libs

tstj:
@wpwoodjr Thank you for the detailed answer! I just saw it now because I didn't get notifications. That sounds quite good so far. There is one thing that came to my mind recently:
Currently, the TimSort-based implementation doesn't implement Galopping in the merging operation. There is a new paper that shows that implementing Galopping won't have a negative effect in general and will improve the performance in many cases. I have not implemented it so far, but compared the non-galopping current stable sort to stable in place symmerge (which uses binary searching and probably could use galopping, too). In many cases symmerge will outperform the non-galopping current stable sort by factor 2 to 3. I assume, that after the implementation of galopping for stable sort, symmerge will not be faster anymore and stable sort will be more than 3 times faster than the current implementation for inputs, where runs are merged that have different lengths (the bigger distance between the lengths, the better). Since you are using the same current merge function, it would be interesting to know, if the galopping search could benefit in your approach, too. And if not, if your approach is still better than the current TimSort. I'm not too deep into the topic, so I can't say it for sure :(

orlp:
> who also had some improvements to the stable sort code that he was interested in PRing

@thomcc @tstj 

Well, yesn't. I am implementing a complete new stable algorithm from scratch (glidesort), which goes way beyond "some improvements". I have a working version without panic safety, and am working on a version with full panic safety with less unsafe code (which is getting delayed due to personal circumstances). I recently held a talk about glidesort here: https://www.youtube.com/watch?v=2y3IK1l6PI4

The results are thus also much more impactful. On my machine (2021 Apple M1 Pro) the speedups are such that glidesort is *faster* than `sort_unstable` for random data, and 3.5x faster than `sort`:

image

When dealing with a low cardinality dataset (under projection of the comparison operator), e.g. sorting by age stored in integer years, which will likely only have ~120 unique values, glidesort becomes faster still (benchmark is 256 unique values by masking integers with `0xFF` before comparing):

image

tstj:
@orlp Congratulations to this algorithm, this sounds very promising!
Maybe I missed some points in your presentation, but I think you have not said which N\*Log(N)\*Log(N) fallback you are using. Is it Symmerge? If so, you could probably still improve it, by helping out the rotation algorithm by the (at this point probably unused) extra buffer. I have implemented a similar (in the sense of the fallback only) hybrid method (but based on current stable sort and Symmerge) which could benefit from this when sizeof(T) is appropriate (otherwise default rotate was faster). In my version, the "append 1%" is much faster (~2-3 times) than the Rust version, because I let Symmerge solve the problem, once the length difference is large (that probably only works better because of missing galloping in current Rust version).

In your presentation you say, that GlideSort will assume cheap comparisons. Will it become slower than another algorithm, when the comprisons are not cheap (for example image comparisons)?
You also say that cheap moves (I assume small sizeof(T) is meant) are assumed.
In the cases where sizeof(T) > ~384 Symmerge with Buffer will slow down and will become slower than GrailSort because of it's N\*Log(N)\*Log(N) worst case on moves. It would be interesting if your algorithm experiences the same problems and will be N\*Log(N)\*Log(N) if sizeof(T) is larger and a constant buffer (e.g. 1024) is used. That was the problem with my code...

It would be interesting to see tests when sizeof(T) >= 256 or even sizeof(T) >= 384, in my opinion, a sizeof(T) < usize is not so common and thus not so important for a sorting algorithm but of course a very good feature.

Also, during my research I found your `shift_head` and `shift_tail`, from pdqsort, I think they do linear searches. Could they do galloping, too?

orlp:
> Maybe I missed some points in your presentation, but I think you have not said which N*Log(N)*Log(N) fallback you are using. Is it Symmerge? 

I did briefly talk about it on the "creating more parallelism slide": https://youtu.be/2y3IK1l6PI4?t=727

I tried symmerge but it was slower and more complicated. I use a simplified scheme based on symmerge that uses block swaps rather than rotations. This is slightly less optimal in terms of comparisons but the smaller, simpler code and better codegen for blockswaps compared to rotations more than made up for it. It also made implied block swaps simpler.

The logic I use is very similar to symmerge. I find the "crossover point" using binary search:

```rust
// Given sorted left, right of equal size n it finds smallest i such that
// for all l in left[i..] and r in right[..n-i] we have l > r. By vacuous truth
// n is always a solution (but not necessarily the smallest).
//
// Panics if left, right aren't equal in size or if is_less does.
pub fn crossover_point<T, F: Cmp<T>>(left: &[T], right: &[T], is_less: &mut F) -> usize { ... }
```

Which then partitions an array as such:

```rust
    let minlen = left.len().min(right.len());
    let left_skip = left.len() - minlen;
    let i = crossover_point(
        &left.as_mut_slice()[left_skip..],
        &right.as_mut_slice()[..minlen],
        is_less,
    );
    let (left0, left1) = left.split_at(left_skip + i).unwrap_abort();
    let (right0, right1) = right.split_at(minlen - i).unwrap_abort();
    (left0, left1, right0, right1)
```

Then merging `left0, right0` followed by `left1, right1` is equivalent to stably merging `left, right`. Note that we don't always have to actually swap `left1` and `right0` - if we have the scratch space available all we do is reassign destination/source buffers. I call this an implied swap.

> In your presentation you say, that GlideSort will assume cheap comparisons. Will it become slower than another algorithm, when the comprisons are not cheap (for example image comparisons)?

If comparisons dominate runtime entirely glidesort pretty much uses the same number of comparisons as `sort` currently does for random data. For patterns it's going to depend on exactly the pattern, overall I'd say better.

Note that if your comparison operator is exceptionally expensive like image comparison, you are much better suited with an algorithm that gets much closer to optimality than mergesort, like [merge-insertion sort](https://en.wikipedia.org/wiki/Merge-insertion_sort).

> In the cases where sizeof(T) > ~384 Symmerge with Buffer will slow down and will become slower than GrailSort (only on random data) because of it's N*Log(N)*Log(N) worst case on moves.

I honestly don't really care about benchmarking the performance for sorting such large structs/arrays, because if anyone actually cared for the performance of sorting those they would be using indirection. Constantly shuffling such large structures around during sorting will never be performant. But yes, with a fixed-size buffer glidesort is O(n log(n)²).

> You also say that cheap moves (I assume small sizeof(T) is meant) are assumed.

I am generally focusing on objects that fit in a couple machine words. E.g. strings, integers, an `(x, y, z)` `f64` triple, a struct of two integers and a string, etc. Larger objects reduce to this case through indirection. Your cache performance will likely suffer with indirection, but then again, you're really not getting that much cache benefits with large objects anyway.

> It would be interesting to see tests when sizeof(T) >= 256 or even sizeof(T) >= 384, in my opinion, a sizeof(T) < sizeof(usize) is not so common and thus not so important for a sorting algorithm but of course a very good feature.

I hold the exact opposite opinion, that `sizeof(T) <= sizeof(usize) * 8` is by far the most important domain to benchmark for a generic system sort.

> Also, during my research I found your shift_head and shift_tail, from pdqsort, I think they do linear searches. Could they do galloping, too?

Neither of these functions are found in pdqsort. And traditional 'galloping' is mostly a waste of time for a system sort in my opinion. It is relatively complex, mostly improves very academic notions of pre-sortedness that have little to no real-world justification (note how the [new paper](https://arxiv.org/abs/2012.03996) you cited does not contain any real-world benchmarks, or even synthetic benchmarks at all), and slows down the hottest core loops. There are some tricks similar to galloping (which I'd rather call 'optimistic binary search') I still might fit in at some points out of the core loops, but I really doubt they would benefit much.

Most of the points you bring up very much come from the old school of thinking about sorting that glidesort (and pdqsort) precisely tries to puncture. Real-world system sorting isn't about minimizing comparisons at all. It's not about moves either, for the most part. It's about minimizing time. And most of the time is gained or lost in (instruction-level) parallelism, branch mispredictions and cache misses. That's why glidesort is 3.5x faster than the current `slice::sort` for random u32s (same for u64s) on my machine, instead of just a modest couple %, or only on a specific distribution.

tstj:
Thank you for the detailed answer, sounds really good to me!

> I hold the exact opposite opinion, that `sizeof(T) <= sizeof(usize) * 8` is by far the most important domain to benchmark for a generic system sort.

I would agree to this, on my machine `sizeof(usize) * 8` is 512 (I meant bits not bytes), that's why I mentioned the sizes 256 and 384 ;)

orlp:
> I would agree to this, on my machine `sizeof(usize) * 8` is 512 (I meant bits not bytes), that's why I mentioned the sizes 256 and 384 ;)

Ah alright, I thought you meant bytes.

yjhn:
If we are discussing possible sort improvements, maybe [vectorizing](https://opensource.googleblog.com/2022/06/Vectorized%20and%20performance%20portable%20Quicksort.html) would be of interest? This might be possible to implement using portable SIMD (probably not, though, as in the linked blog post they use SIMD runtime dispatch, which would probably [not work](https://github.com/rust-lang/portable-simd/issues/178) with Rust's portable SIMD).

orlp:
> If we are discussing possible sort improvements, maybe [vectorizing](https://opensource.googleblog.com/2022/06/Vectorized%20and%20performance%20portable%20Quicksort.html) would be of interest? This might be possible to implement using portable SIMD (probably not, though, as in the linked blog post they use SIMD runtime dispatch, which would probably [not work](https://github.com/rust-lang/portable-simd/issues/178) with Rust's portable SIMD).

Vectorized quicksort / sorting networks have been around for quite a while, but they're pretty much limited to sorting small integers without custom comparison functions. Not very applicable to a generic sort for arbitrary types with arbitrary comparison functions. Maybe in the future we can specialize for it, but I wouldn't consider it high priority.

Logarithmus:
@wpwoodjr https://spin.atomicobject.com/2020/05/05/git-configurations-default/
**TLDR;** `git config --global pull.rebase true`

Voultapher:
Running [my test suite](https://github.com/Voultapher/sort-research-rs/blob/7e2fc1e544973955a7a6b7758878bd0189684414/tests/main.rs#L288) miri complains about UB:

```
cargo miri test random

Seed: 17511317657358896758
error: Undefined Behavior: attempting a read access using <389996> at alloc165698[0x0], but that tag does not exist in the borrow stack for this location
   --> /projects/rust/sort-research-rs/src/wpwoodjr_stable_sort/mod.rs:144:13
    |
144 |             ptr::copy_nonoverlapping(hole.dest, v.get_unchecked_mut(end), 1);
    |             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    |             |
    |             attempting a read access using <389996> at alloc165698[0x0], but that tag does not exist in the borrow stack for this location
    |             this error occurs as part of an access at alloc165698[0x0..0x4]
    |
    = help: this indicates a potential bug in the program: it performed an invalid operation, but the Stacked Borrows rules it violated are still experimental
    = help: see https://github.com/rust-lang/unsafe-code-guidelines/blob/master/wip/stacked-borrows.md for further information
help: <389996> was created by a retag at offsets [0x0..0x4]
   --> /projects/rust/sort-research-rs/src/wpwoodjr_stable_sort/mod.rs:142:23
    |
142 |                 dest: v.get_unchecked_mut(end - 1),
    |                       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
help: <389996> was later invalidated at offsets [0x0..0x8]
   --> /projects/rust/sort-research-rs/src/wpwoodjr_stable_sort/mod.rs:144:49
    |
144 |             ptr::copy_nonoverlapping(hole.dest, v.get_unchecked_mut(end), 1);
    |                                                 ^^^^^^^^^^^^^^^^^^^^^^^^
    = note: backtrace:
    = note: inside `sort_comp::wpwoodjr_stable_sort::insert_end::<i32, [closure@sort_comp::wpwoodjr_stable_sort::sort_by<i32, [closure@tests/main.rs:56:52: 56:58]>::{closure#0}]>` at /projects/rust/sort-research-rs/src/wpwoodjr_stable_sort/mod.rs:144:13
note: inside `sort_comp::wpwoodjr_stable_sort::merge_sort::<i32, [closure@sort_comp::wpwoodjr_stable_sort::sort_by<i32, [closure@tests/main.rs:56:52: 56:58]>::{closure#0}]>` at /projects/rust/sort-research-rs/src/wpwoodjr_stable_sort/mod.rs:339:13
   --> /projects/rust/sort-research-rs/src/wpwoodjr_stable_sort/mod.rs:339:13
    |
339 |             insert_end(&mut v[..=i], &mut is_less);
    |             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
note: inside `sort_comp::wpwoodjr_stable_sort::sort_by::<i32, [closure@tests/main.rs:56:52: 56:58]>` at /projects/rust/sort-research-rs/src/wpwoodjr_stable_sort/mod.rs:95:5
   --> /projects/rust/sort-research-rs/src/wpwoodjr_stable_sort/mod.rs:95:5
    |
95  |     merge_sort(arr, |a, b| compare(a, b) == Ordering::Less);
```

Voultapher:
As noted in the other PR I ran my benchmark suite, but only zen3, with this implementation and got:

Screenshot 2022-08-24 at 18 22 18

A statistically significant speedup in most cases, with mostly ascending_saw and descending_saw seeing regressions. Levelling out at around 6% for random u64 inputs hot.

Screenshot 2022-08-24 at 18 22 44

Strings see smaller improvements, but overall a slight improvement. You can find the [full results here](https://voultapher.github.io/sort-research-rs/results/wpwoodjr/analysis/zen3/).

KittyBorgX:
@rustbot author
@wpwoodjr Hey! It looks like there are some merge conflicts, can you sort them out?
Also, any update on the status of the PR? 

Once you're done doing everything, type in `@rustbot review` here to change the tag to `S-waiting-on-review`

wpwoodjr:
> @rustbot author @wpwoodjr Hey! It looks like there are some merge conflicts, can you sort them out? Also, any update on the status of the PR?
> 
> Once you're done doing everything, type in `@rustbot review` here to change the tag to `S-waiting-on-review`

OK, will try and get to it in the next week or so.

wpwoodjr:
> > * It does not track and merge "runs".  However, on each call to `slice_merge_sort`, the beginning of the slice is checked for the number of already sorted elements.  These already sorted elements are never re-checked or redundantly sorted, saving considerable time.
> 
> Does it also do this for the end of the slice? One of the things that's nice about the current sort is that you can `extend` a sorted collection with another once, call `sort`, and know it's going to do nearly as well as calling a dedicated merge function -- which `std` doesn't yet have. (Or, in general, `chain` a bunch of them and have the sort adaptively take advantage of the work that was already done.)

Yes, it will detect the sorted prefix collection, then merge in the second collection.

wpwoodjr:
> > @rustbot author @wpwoodjr Hey! It looks like there are some merge conflicts, can you sort them out? Also, any update on the status of the PR?
> > Once you're done doing everything, type in `@rustbot review` here to change the tag to `S-waiting-on-review`
> 
> OK, will try and get to it in the next week or so.

This will take a bit longer than I anticipated, as the sort routines have been moved to the core library, and I am traveling next week.  I will get back to it when I return.

JohnCSimon:
@wpwoodjr 

ping from triage - can you post your status on this PR? There hasn't been an update in a few months. Thanks!


JohnCSimon:
@wpwoodjr 
Note: if you are going to continue please open the PR **BEFORE** you push to it, else you won't be able to reopen - this is a quirk of github. 
Thanks for your contribution.

@rustbot label: +S-inactive

Issue #83751, opened 2021-04-01
Windows - std::fs bug - specific file name causes thread to hang or false positive on file open.


nothingbutnetcode:
When trying to open a file (which in this case did not exist) that starts with con. - eg: con.txt, con.toml - fs::read_to_string will hang the thread.  File::open will incorrectly succeed, and subsequently panic when methods on the file handle are called.

I tried this code:

```rust
use std::fs::File;
use std::{fs};

fn main() {
    // 1 - this will hang the app
    let _ = fs::read_to_string("con.cxv").unwrap();

    // 2 - this returns a file (but it does not exist)
    match File::open("con.txt") {
        Ok(file) => {
            println!("length {}", file.metadata().unwrap().len()) // so this will panic
        }
        Err(err) => {
            println!("{:?}", err);
        }
    }
}
```

I expected to see this happen:  File not found error returned.

Instead, this happened: Thread hang or incorrect result as per description above.

### Meta

Bug exists in nightly as well as stable.

`rustc --version --verbose`:
```
rustc 1.51.0 (2fd73fabe 2021-03-23)
binary: rustc
commit-hash: 2fd73fabe469357a12c2c974c140f67e7cdd76d0
commit-date: 2021-03-23
host: x86_64-pc-windows-msvc
release: 1.51.0
LLVM version: 11.0.1
```

<details><summary>Backtrace</summary>
<p>

```
<backtrace>
```

</p>
</details>

jonas-schievink:
That's because any file named `con` (ignoring the extension) refers to the CONsole device: https://superuser.com/questions/86999/why-cant-i-name-a-folder-or-file-con-in-windows

Reading from it presumably will block until the console is closed, so the hang is expected.

the8472:
If you use a verbatim prefix (`\\?\`) then the resolution to legacy device names shouldn't happen. But then some programs then may have issues dealing with a directory containing `con.txt`.

nothingbutnetcode:
The main issue here is, in the case of read_to_string which hangs, you can check if it exists first with Path::new("con.txt").exists(), this gives the create answer of false, but according to the read_to_string API doc - "This function will return an error if path does not already exist."  So one would expect that error to return, rather than DOS your app.

the8472:
The `exists()` result is not the canonical definition of what exists from the perspective of the filesystem. Rust just tries to probe existence in one particular way (fetching `metadata()`) without opening the file. Opening a path may give a different result.
[...]

PR #90547, opened 2021-11-03
Make `path::Display` output user-style paths instead of verbatim

### LINKED ISSUE #31789: Windows paths should not be prefixed with \\?\ when displayed
That just turns this already confusing error into an insult:
```
kernel\std\std.rs:17:1: 17:46 error: found possibly newer version of crate `core` which `core_collections` depends on [E0460]
kernel\std\std.rs:17 extern crate collections as core_collections;
                     ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
kernel\std\std.rs:17:1: 17:46 note: perhaps this crate needs to be recompiled?
kernel\std\std.rs:17 extern crate collections as core_collections;
                     ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
kernel\std\std.rs:17:1: 17:46 note: crate `core` path #1: \\?\B:\Programmering\SandboxOS\AveryRust\build\sysroot\lib\rustlib\x86_64-avery-kernel\lib\libcore.rlib
kernel\std\std.rs:17:1: 17:46 note: crate `core_collections` path #1: \\?\B:\Programmering\SandboxOS\AveryRust\build\sysroot\lib\rustlib\x86_64-avery-kernel\lib\libcollections.rlib
```

ChrisDenton:
Fixes #31789

Windows verbatim paths are essentially an API hack to allow smuggling NT kernel paths through the Win32 API. This allows bypassing certain historic limits of the Win32 APIs. Most users do not normally encounter them and can find it confusing if they do. However some APIs, most notably the one used by `canonicalize`, will return verbatim paths.

This PR changes `Path::Display` so that it will prefer to display user-style paths instead of verbatim. Examples of the transformations that are applied are:

|Verbatim|User|
|--|--|
|`\\?\C:\directory\file.txt`|`C:\directory\file.txt`|
|`\\?\UNC\server\share\file.txt`|`\\server\share\file.txt`|

This will not affect verbatim device paths. I think, from a user perspective, `\\?\PIPE\name` and `\\.\PIPE\name` are equally strange and rarely encountered.

### DIFF
--- a/library/std/src/path.rs
+++ b/library/std/src/path.rs
@@ -2841,6 +2841,11 @@ impl fmt::Debug for Path {
 /// println!("{}", path.display());
 /// ```
 ///
+/// # Windows
+///
+/// Verbatim paths may be converted to their more familiar form.
+/// For example, `\\?\C:\Program Files\Rust` may display as `C:\Program Files\Rust`.
+///
 /// [`Display`]: fmt::Display
 /// [`format!`]: crate::format
 #[stable(feature = "rust1", since = "1.0.0")]
@@ -2858,7 +2863,26 @@ impl fmt::Debug for Display<'_> {
 #[stable(feature = "rust1", since = "1.0.0")]
 impl fmt::Display for Display<'_> {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
-        self.path.inner.display(f)
+        let path = if cfg!(windows) && self.path.as_u8_slice().starts_with(br"\\?\") {
+            // Convert Windows drive verbatim and UNC verbatim paths to their
+            // user path equivalents.
+            // SAFETY: Paths are only ever split on ASCII boundaries.
+            match self.path.as_u8_slice()[4..] {
+                // \\?\C:\, \\?\D:\, etc to C:\, D:\, etc
+                ref path @ [drive @ _, b':', b'\\', ..] if drive.is_ascii_alphabetic() => unsafe {
+                    Path::from_u8_slice(path)
+                },
+                // \\?\UNC\ to \\
+                [b'U', b'N', b'C', b'\\', ref path @ ..] => {
+                    f.write_str(r"\\")?;
+                    unsafe { Path::from_u8_slice(path) }
+                }
+                _ => self.path,
+            }
+        } else {
+            self.path
+        };
+        path.inner.display(f)
     }
 }
 
--- a/library/std/src/path.rs
+++ b/library/std/src/path.rs
@@ -2863,22 +2863,17 @@ impl fmt::Debug for Display<'_> {
 #[stable(feature = "rust1", since = "1.0.0")]
 impl fmt::Display for Display<'_> {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
-        let path = if cfg!(windows) && self.path.as_u8_slice().starts_with(br"\\?\") {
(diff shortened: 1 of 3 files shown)


the8472:
What about paths exceeding MAX_PATH?

ChrisDenton:
As this is only for display, I don't think that should matter? It's not intended to be passed directly to the Windows API.

the8472:
I guess the user might copy-paste it when working on the CLI

ChrisDenton:
Sure. On the other hand `Display` is already (potentially) lossy. And if the program that it's handed to can't handle long paths then it's likely to run into difficulty in any case.

joshtriplett:
I do think it's reasonable to expect to be able to copy-paste a path and use it. If this is *only* an issue of long paths, I agree that the target program could take care of that. Are there other cases where a verbatim path is necessary for round-tripping, where the program that receives a pasted path *won't* be able to correct for it?

ChrisDenton:
Hmm, having thought about more in terms of copy/pasting, I do have some concerns.

* If a path has `.` or `..` components then these are interpreted literally for verbatim paths but are parsed lexically for other paths. However, `.` and `..` are special so in any case they cannot be used as literal file names on a filesystem.
* If a drive path (e.g. `C:\` paths) ends with a legacy dos device (e.g. `NUL`) then the Windows API will convert it to `\\.\NUL` (aka the `NUL` device). But verbatim paths will treat the file name literally. Such file names will likely ruin someone's day however they're interpreted. Creating a literal file or directory named `NUL` on the filesystem confuses programs (including the Windows file explorer). But sending (via copy/paste) other programs a file path that magically turns into a device can also cause problems.
* With non-verbatim paths, trailing `.` and ` ` (ASCII dot and space) are stripped from the the final component. But with verbatim paths they're kept.

In the last two cases, it's technically possible for applications to recover the intended meaning. After all, it's a fair assumption that a user [intended](https://github.com/rust-lang/rust/issues/83751) `C:\directory\NUL` to be interpreted as they wrote it and not as `\\.\NUL`. Even if the Windows API disagrees for legacy reasons. On the other hand, you cannot at all rely on other applications to "correctly" interpret intent. They might or might not.

Therefore, it might be best to only do the conversion on unambiguous paths. In which case I'll need to rethink my approach here.

ChrisDenton:
I've update this PR so that it only converts from verbatim if doing so wouldn't change the meaning of the path according to Win32 parsing rules. This essentially means scanning the path for anything that would be stripped during normal path handling, or for DOS device that cause the path to completely change. My only concern now is that the code is nowhere near as trivial as where I started from.

ChrisDenton:
Ok, I think I'll close this now. My hope was to make it easy for Rust applications to display regular looking paths by default but I feel like there are too many pitfalls and adding all this on to the `Display` struct is doing too much behind the scenes magic.

I do think it'd be nice to have something similar but perhaps in a way that's more explicit. Perhaps in a similar way that you can choose the format of some other types. Or perhaps something else. I'm not sure.

the8472:
> I do think it'd be nice to have something similar but perhaps in a way that's more explicit. 

So, [dunce](https://crates.io/crates/dunce)?

ChrisDenton:
> So, [dunce](https://crates.io/crates/dunce)?

Sure! But may main motivation here was that, for example, Mac or Linux devs wouldn't have to go out of their way to do the more natural thing for Windows users. But maybe that goal is unattainable in this case.

PR #90550, opened 2021-11-03
Update certificates in some Ubuntu 16 images.

ehuss:
These images use crosstool-ng, which needs to download various things off the internet. The certificate for `www.kernel.org` no longer works with the ca-certificates in Ubuntu 16. This resolves the issue by grabbing from a newer image a certificate bundle from https://curl.se/ca/cacert.pem, which is usually somewhat up to date.

### DIFF
--- a/src/ci/docker/host-x86_64/dist-aarch64-linux/Dockerfile
+++ b/src/ci/docker/host-x86_64/dist-aarch64-linux/Dockerfile
@@ -1,4 +1,17 @@
+FROM ubuntu:20.04
+RUN apt-get update && \
+    apt-get install -y --no-install-recommends \
+        curl \
+        ca-certificates
+WORKDIR /tmp
+RUN curl -f https://curl.se/ca/cacert.pem -o cacert.pem
+
 FROM ubuntu:16.04
+
+# The ca-certificates in ubuntu-16 is too old, so update the certificates
+# with something more recent.
+COPY --from=0 /tmp/cacert.pem /tmp/cacert.pem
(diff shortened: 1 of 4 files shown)


ehuss:
There are some alternate approaches this could take:
* Try to download everything that's needed outside of Ubuntu 16 (like what was done in #90457). However, I don't know how to hook those cached downloads into the crosstools script.
* Download the `ca-certificates.deb` file from Ubuntu 20 and copy that over and install it manually. I think that is relatively simple, so I'm happy to try that solution if desired.
* Update to a newer image. I tried briefly to update to Ubuntu 18, but the crosstools config needs updating for that to work, and I don't want to dig into that. However, I think this in general is a good idea, as these images continue to get older.

I only updated the images that are using crosstools. I did not check if other Ubuntu 16 images need to download things, but a quick skim I didn't see any.

BTW, is there a way to reuse parts of a Dockerfile? I looked around, and didn't see any simple solutions.

JohnCSimon:
Triage: looks like this still needs work

ehuss:
@JohnCSimon This is actually waiting on review.

@kennytm Do you think this is something you can review?  Or perhaps @Mark-Simulacrum would prefer to review it?

Mark-Simulacrum:
I think this is okay.

I'm not too happy with the impl here, but in general I'm not sure we have great alternatives, and the duplication seems OK. I suspect there's no really great way to change this that is still nice in general. The general pattern we seem to have is that our Dockerfiles get increasingly mangled with workarounds over time, unfortunately, and I'm not sure we can do much to help there.

@bors r+ rollup=iffy

PR #90551, opened 2021-11-04
Add lossless integer conversion by reference

vandenheuvel:
This pull request adds conversion of integer types by reference, e.g. `i16::from(&1_u8)`.

This is useful for writing code that is generic over number types. Just like we have number operations by reference (e.g. `std::ops::Add` impls allow `let _: u16 = 1_u16 + &1_u16`), it can be necessary to convert number types by reference. 

The alternative is cloning before converting, which is not always an option. Take for example the code below. When the right-hand side `MaybeLarge2` is large type, such as an arbitrary precision type, the intermediate `clone` is too expensive, as it could require an allocation.

```rust
fn compute<'a, MaybeLarge1, MaybeLarge2>(left: &mut MaybeLarge1, right: &'a MaybeLarge2)
where
    MaybeLarge1: From<'a MaybeLarge2>,
    ... // A few std::ops and number traits
{
    if left.is_negative() {
        *left = From::from(right);
        // the alternative would be `*left = From::from(right.clone());`
    } else {
        left += right;
    }
    
    ...
}
```

Adding these conversion implementations makes such generic code possible.

The implementation could be constified with the rest of #87852.

### DIFF
--- a/library/core/src/convert/num.rs
+++ b/library/core/src/convert/num.rs
@@ -430,6 +430,8 @@ macro_rules! nzint_impl_from {
                 }
             }
         }
+
+        impl_from_ref! { $Small, $Large }
     };
     ($Small: ty, $Large: ty, #[$attr:meta]) => {
         nzint_impl_from!($Small,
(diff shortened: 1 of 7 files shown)


Mark-Simulacrum:
What is the motivation for adding these? It seems like most code can fairly easily dereference as needed to call the existing impls -- and if e.g. diagnostics are not so good today, then I would rather improve them then add this particular special case, personally.

vandenheuvel:
@Mark-Simulacrum I wrote a [PR description](https://github.com/rust-lang/rust/pull/90551#issue-1044236359). Once https://github.com/rust-lang/rust/pull/90560 is merged (perhaps you could take a look? I haven't gotten any response there...), I will rebase this.

Mark-Simulacrum:
Ok. I'm not sure I'm personally convinced - I think a pointer to an actual set of code (rather than theoretical) would help me.

r? @dtolnay for T-libs-api FCP since this would be insta stable

vandenheuvel:
This arose in a linear programming framework I'm developing. [This trait](https://github.com/vandenheuvel/relp/blob/master/src/algorithm/two_phase/tableau/inverse_maintenance/mod.rs#L30) is implemented by using an internal number type that might be arbitrary precision depending on the specific problem being solved. That problem is represented by an implementor of [this trait](https://github.com/vandenheuvel/relp/blob/master/src/algorithm/two_phase/matrix_provider/mod.rs#L37), which in turn can have any number type to represent some problem. Some problems might need arbitrary precision, while others might be able to get by with small rational numbers or even integers or bools.

These two traits are essentially connected through traits like [this one](https://github.com/vandenheuvel/relp/blob/master/src/algorithm/two_phase/tableau/inverse_maintenance/ops.rs#L72). Part of that alias is the [From bound](https://github.com/vandenheuvel/relp/blob/master/src/algorithm/two_phase/tableau/inverse_maintenance/ops.rs#L78), which currently prohibits using this system with integers. The newtype pattern is of course incredibly cumbersome to use in this situation.

dtolnay (commented):
I agree with Mark in being a little skeptical about adding this.

<br>

> ``` // the alternative would be `*left = From::from(right.clone());` ```

As a counterproposal, I wonder how you feel about the downstream crate doing its trait bounds like this:

```rust
pub trait CheapCopy {
    type Type;
    fn cheap_copy(self) -> Self::Type;
}

impl CheapCopy for &i32 {
    type Type = i32;
    fn cheap_copy(self) -> Self::Type { *self }
}

struct BigInteger;
impl<'a> CheapCopy for &'a BigInteger {
    type Type = &'a BigInteger;
    fn cheap_copy(self) -> Self::Type { self }
}

pub type DoCheapCopy<T> = <T as CheapCopy>::Type;

fn compute<'a, MaybeLarge1, MaybeLarge2>(left: &mut MaybeLarge1, right: &'a MaybeLarge2)
where
    &'a MaybeLarge2: CheapCopy,
    MaybeLarge1: From<DoCheapCopy<&'a MaybeLarge2>>,
{
    if left.is_negative() {
        *left = From::from(right.cheap_copy());
    } else {
        // ...
    }
}
```

I figure you are going to want this **anyway**, so that you can switch to `right: MaybeLarge2` and avoid requiring callers to do the silly thing of passing primitives by reference. This is an even cleaner result than the trait bounds you'd get from this PR, as you won't need explicit lifetime annotations anymore.

```rust
// The above impls, plus impl CheapCopy for i32 { type Type = i32 }.

fn compute<MaybeLarge1, MaybeLarge2>(left: &mut MaybeLarge1, right: MaybeLarge2)
where
    MaybeLarge2: CheapCopy,
    MaybeLarge1: From<DoCheapCopy<MaybeLarge2>>;
```

dtolnay:
Gonna close but we can come back to this if someone decides they're still justified after my [suggestion](https://github.com/rust-lang/rust/pull/90551#pullrequestreview-803400003).

dtolnay:
Thanks anyway!

vandenheuvel:
@dtolnay I'm still looking into your suggestion, thanks.

Issue #85532, opened 2021-05-21
Tracking Issue for bigint helper methods


clarfonthey:
Feature gate: `#![feature(bigint_helper_Methods)]` and `#![feature(const_bigint_helper_methods)]`

This is a tracking issue for the following methods on integers:

* `carrying_add`
* `borrowing_sub`
* `carrying_mul`
* `widening_mul`

These methods are intended to help centralise the effort required for creating efficient big integer implementations, by offering a few methods which would otherwise require special compiler intrinsics or custom assembly code in order to do efficiently. They do not alone constitute big integer implementations themselves, but are necessary building blocks for a larger implementation.

### Public API

```rust
/// `self + rhs + carry` (full adder)
fn carrying_add(self, rhs: Self, carry: bool) -> (Self, bool);

/// `self - rhs - carry` (full "subtractor")
fn borrowing_sub(self, rhs: Self, carry: bool) -> (Self, bool);

/// `self * rhs + carry` (multiply-accumulate)
fn carrying_mul(self, rhs: Self, carry: Self) -> (Self, Self);

/// `self * rhs` (wide multiplication)
fn widening_mul(self, rhs: Self) -> (Self, Self);
```

### Steps / History

- [x] `widening_mul` RFC: rust-lang/rfcs#2417
- [x] Initial implementation of `carrying_add`, `borrowing_sub`, `carrying_mul`, and `widening_mul` #85017
- [ ] Final commenting period (FCP)
- [ ] Stabilization PR

### Unresolved Questions

- Should these be implemented using compiler intrinsics? (probably yes)
- Should an alternative API be provided for `widening_mul` that simply returns the next-larger type? What would be do for `u128`/`i128`?
- What should the behaviour be for signed integers? Should there be implementations for signed integers at all?
- Is the "borrowing" terminology worth it for subtraction, or should we simply call that "carrying" as well for consistency?
- Are there other methods that should be added in addition to the existing ones?

leonardo-m:
> Are there other methods that should be added in addition to the existing ones?

I'd like a mul_mod, as shown in #85017, because I think you can't implement it efficiently without asm and it's a basic block for power_mod and other things.

clarfonthey:
Another set of methods that could be useful that I'll probably offer implementations for at some point:

```rust
/// `(self << rhs) | carry`
fn carrying_shl(self, rhs: u32, carry: Self) -> (Self, Self); 

/// `(self >> rhs) | carry`
[...]

Issue #90493, opened 2021-11-02
The carry in signed carrying_mul on integers should be unsigned


dzimmanck:
I tried this code:

```rust
// use signed carry_mul to chain two multiplies (x*y) + (x*y) >> 64
let carry: i32 = 0;
let x: i32 = 200000;
let y: i32 = 100000;
let (carry, prod) = x.carrying_mul(y, carry);
let (carry, prod) = x.carrying_mul(y, carry);

// due the calculation with larger integers
let large = i64::from(x) * i64::from(y);
let carry = large - ((large>>32)<<32);
let expected = ((large + carry)>>32) as i32;

assert_eq!(expected, prod);
```

I expected to see this happen: I expected the "prod" to be 5.

Instead, this happened: The product is 4 due to the carry in the first calculation being negative (overflowing) causing it to accumulate incorrectly.

### Meta

`rustc --version --verbose`:
```
rustc 1.58.0-nightly (db062de72 2021-11-01)
binary: rustc
commit-hash: db062de72b0a064f45b6f86894cbdc7c0ec68844
commit-date: 2021-11-01
host: x86_64-pc-windows-msvc
release: 1.58.0-nightly
LLVM version: 13.0.0
```

</p>
</details>

matthewjasper:
cc #85532

dzimmanck:
I'm not in a position to get setup to do a well articulated pull request at the moment, but IMO, the change is simple.  These should always be UNSIGNED.

![image](https://user-images.githubusercontent.com/16160230/139881907-2ff190e9-7bbc-483a-8100-2aa224248d03.png)

scottmcm:
Hmm, I think it might help for the comment on these methods to clarify exactly what the interpretation of the output values are.  For unsigned it's obvious enough (as it's `self * rhs + carry = out.0 + out.1 * (MAX + 1)`), but the asymmetrical range of signed numbers makes that less clear.

PR #90556, opened 2021-11-04
Add more text and examples to `carrying_{add|mul}`

scottmcm:
`feature(bigint_helper_methods)` tracking issue https://github.com/rust-lang/rust/issues/85532

cc @clarfonthey

### DIFF
--- a/library/core/src/num/mod.rs
+++ b/library/core/src/num/mod.rs
@@ -94,7 +94,13 @@ depending on the target pointer size.
 }
 
 macro_rules! widening_impl {
-    ($SelfT:ty, $WideT:ty, $BITS:literal) => {
+    ($SelfT:ty, $WideT:ty, $BITS:literal, unsigned) => {
+        widening_impl!($SelfT, $WideT, $BITS, "");
+    };
+    ($SelfT:ty, $WideT:ty, $BITS:literal, signed) => {
+        widening_impl!($SelfT, $WideT, $BITS, "# //");
+    };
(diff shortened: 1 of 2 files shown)


library/core/src/num/mod.rs
```diff
@@ -148,6 +154,33 @@ macro_rules! widening_impl {
         /// assert_eq!(5u32.carrying_mul(2, 10), (20, 0));
         /// assert_eq!(1_000_000_000u32.carrying_mul(10, 0), (1410065408, 2));
         /// assert_eq!(1_000_000_000u32.carrying_mul(10, 10), (1410065418, 2));
+        #[doc = concat!($AdaptiveTestPrefix, "assert_eq!(",
+            stringify!($SelfT), "::MAX.carrying_mul(", stringify!($SelfT), "::MAX, ", stringify!($SelfT), "::MAX), ",
+            "(0, ", stringify!($SelfT), "::MAX));"
+        )]
```
scottmcm, line 160: `$AdaptiveTestPrefix` is a bit of a hack because this currently doesn't pass on signed types.  It's unclear to me whether it should -- there's open questions about this method on signed types in things like https://github.com/rust-lang/rust/issues/90493 -- so for now it's just skipped and hidden.

joshtriplett:
Er, oops. I assigned this to you because you posted what looked like a review comment, and I didn't catch that you were also the PR author. :)

r? @joshtriplett

joshtriplett:
This looks reasonable to me.

The issue with the signed methods definitely needs resolving before we consider stabilizing `bigint_helper_methods`.

Issue #90514, opened 2021-11-02
coercions perf benchmark regressed by 3% to 14% after bidi lint addition


pnkfelix:
Spawned off of https://github.com/rust-lang/rust/pull/90462#issuecomment-956335341

A bunch of benchmarks regressed. Its understandable to see overhead from a new on-by-default lint, especially one that operates at the lexing phase (and thus will affect a very broad range of benchmarks).

But coercions seems like it was hit particularly hard in some cases. I have to imagine we could do something to address that 14% regression to coercions check. (Note though that the biggest regressions are only on incremental builds of coercions. Full builds of coercions regressed by "only" 6.7%.)

hkratz:
#90559 fixes most of the performance regression so I think this issue can be closed.

alex:
a) It sounds like this can be closed
b) I believe it should be I-comptime, not I-slow

PR #90559, opened 2021-11-04
Optimize bidi character detection.

hkratz:
Should fix most of the performance regression of the bidi character detection (#90514), to be confirmed with a perf run.

### DIFF
--- a/compiler/rustc_parse/src/lexer/mod.rs
+++ b/compiler/rustc_parse/src/lexer/mod.rs
@@ -153,14 +153,15 @@ impl<'a> StringReader<'a> {
             match core::slice::memchr::memchr(0xE2, &bytes) {
                 Some(idx) => {
                     // bytes are valid UTF-8 -> E2 must be followed by two bytes
-                    match bytes[idx + 1] {
+                    let ch = &bytes[idx..idx + 3];
+                    match ch[1] {
                         0x80 => {
-                            if (0xAA..=0xAE).contains(&bytes[idx + 2]) {
+                            if (0xAA..=0xAE).contains(&ch[2]) {
                                 break true;
                             }
                         }
                         0x81 => {
-                            if (0xA6..=0xA9).contains(&bytes[idx + 2]) {
+                            if (0xA6..=0xA9).contains(&ch[2]) {
                                 break true;
                             }
                         }
--- a/compiler/rustc_parse/src/lexer/mod.rs
+++ b/compiler/rustc_parse/src/lexer/mod.rs
@@ -137,12 +137,45 @@ impl<'a> StringReader<'a> {
         // Opening delimiter of the length 2 is not included into the comment text.
         let content_start = start + BytePos(2);
         let content = self.str_from(content_start);
-        let span = self.mk_sp(start, self.pos);
-        const UNICODE_TEXT_FLOW_CHARS: &[char] = &[
-            '\u{202A}', '\u{202B}', '\u{202D}', '\u{202E}', '\u{2066}', '\u{2067}', '\u{2068}',
-            '\u{202C}', '\u{2069}',
-        ];
-        if content.contains(UNICODE_TEXT_FLOW_CHARS) {
+
+        // Char   - UTF-8
+        // U+202A - E2 80 AA
+        // U+202B - E2 80 AB
+        // U+202C - E2 80 AC
+        // U+202D - E2 80 AD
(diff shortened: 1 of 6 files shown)


compiler/rustc_parse/src/lexer/mod.rs
```diff
@@ -137,12 +137,45 @@ impl<'a> StringReader<'a> {
...
+        // U+2068 - E2 81 A8
+        // U+2069 - E2 81 A9
+        let mut bytes = content.as_bytes();
+        let contains_flow_control_chars = loop {
+            match core::slice::memchr::memchr(0xE2, &bytes) {
+                Some(idx) => {
+                    // bytes are valid UTF-8 -> E2 must be followed by two bytes
+                    match bytes[idx + 1] {
+                        0x80 => {
+                            if (0xAA..=0xAE).contains(&bytes[idx + 2]) {
```
joshtriplett, line 158: Given that indexing includes a bounds check, I wonder if the compiler will be able to optimize these into a single test, rather than two separate tests. It doesn't matter much, since the appearance of 0xE2 will already make this the uncommon case, but it may become an issue if people start writing other Unicode that includes 0xE2.

To give the compiler the best chance of optimizing this, how about starting the `Some` case above with `let ch = bytes[idx..=idx+2];` (right after the comment explaining why that's correct), and then indexing into `ch`? I think that'll make the code a little clearer regardless.

compiler/rustc_parse/src/lexer/mod.rs
```diff
@@ -137,12 +137,45 @@ impl<'a> StringReader<'a> {
...
+        // U+202E - E2 80 AE
+        // U+2066 - E2 81 A6
+        // U+2067 - E2 81 A7
+        // U+2068 - E2 81 A8
+        // U+2069 - E2 81 A9
+        let mut bytes = content.as_bytes();
+        let contains_flow_control_chars = loop {
+            match core::slice::memchr::memchr(0xE2, &bytes) {
+                Some(idx) => {
+                    // bytes are valid UTF-8 -> E2 must be followed by two bytes
```
joshtriplett, line 155:
```suggestion
                    // bytes are valid UTF-8 -> E2 must be followed by two bytes
                    let ch = bytes[idx..=idx+2];
```

compiler/rustc_parse/src/lexer/mod.rs
hkratz: `let ch = &bytes[idx..=idx+2];` compiles with optimization to _three_ comparisons to differentiate between those conditions: `idx+2 != usize::max` (for panicking with "attempted to index slice up to maximum usize"), `idx<=idx+2` (not optimized away because the addtion could overflow) and `idx + 2 < len`. We could get rid of the first one using `let ch = &bytes[idx..idx+3];` but that would still mean one more comparison for the common case that second byte is neither 0x80 nor 0x81. ([Godbolt](https://godbolt.org/z/cdK4ejjq7))

hkratz: The code size is quite a bit smaller though, so it might be worth it anyway.

hkratz:
Hmm, it seems the [coercions regression](https://github.com/rust-lang/rust/pull/90462#issuecomment-956335341) is not related to the lexer changes. I will have a look at it later.

Nevertheless I would like to see if the small modification suggested by @joshtriplett changed anything. Can I have another perf run?

joshtriplett:
@borst try @rust-timer queue

hkratz:
I missed doc comments and literals being linted at a later stage. The new commits should reduce the performance impact as much as possible for the `coercions` benchmark (which is mostly string literals) as well.

compiler/rustc_parse/src/lexer/mod.rs
the8472: Does evaluating the bytes separate improve performance compared to

```rust
match ch {
   [_, 0x80, 0xAA..=0xAE] | [_, 0x81, 0xA6..=0xA9] => break true,
   _ => {}
}
```

hkratz: Evaluating the bytes separately made a difference before without the subslice because one overflow check could be delayed. Now all checks are done at subslice creation so the assembly is the same (modulo reordering). I will change it.

hkratz:
Can we run the benchmarks once more?

davidtwco (approved):

PR #90561, opened 2021-11-04
[beta] Backport Clippy lint renaming / regrouping

flip1995:
This backports PR rust-lang/rust-clippy#7810 to beta, where a lint added in the 1.57 (current beta) release cycle was renamed shortly after the beta cutoff. We want to backport this, so we don't have unnecessary `lint_store.registered_renamed` calls in the Clippy codebase forever.

(The formatting changes were necessary to pass this backport in the Clippy CI, removing it in a subtree sync would be a bit painful...)

cc @camsteffen


Mark-Simulacrum:
@bors r+ rollup=never

This seems OK. I'd probably prefer to delete the newly introduced lint, reverting the patch introducing it, (rather than backporting the rename), but the changes are small enough that it doesn't seem that bad -- and the deletion would probably be closer to a revert but still likely to be larger, so not necessarily better.

flip1995:
This seems like a general certificate issue with the beta branch. I'm not sure what to do here.

Mark-Simulacrum:
We need to backport one of the fix PRs, most likely. I'll add it to my list.

matthiaskrgr:
let's try this again now that the beta backports with the python cert stuff was merged
@bors r=Mark-Simulacrum  rollup=never

PR #90569, opened 2021-11-04
Fix tests using `only-i686` to use the correct `only-x86` directive

wesleywiser:
We translate `i686` to `x86` which means tests marked as `only-i686`
never ran. Update those tests to use `only-x86`.

We parse the `only-` architecture directive here

https://github.com/rust-lang/rust/blob/27143a9094b55a00d5f440b05b0cb4233b300d33/src/tools/compiletest/src/util.rs#L160-L168

and we translate `i686` to `x86` here

https://github.com/rust-lang/rust/blob/27143a9094b55a00d5f440b05b0cb4233b300d33/src/tools/compiletest/src/util.rs#L56

### DIFF
--- a/src/test/ui/extern/extern-methods.rs
+++ b/src/test/ui/extern/extern-methods.rs
@@ -1,5 +1,5 @@
 // run-pass
-// only-i686
+// only-x86
 
 trait A {
     extern "fastcall" fn test1(i: i32);
--- a/src/test/ui/extern/extern-thiscall.rs
+++ b/src/test/ui/extern/extern-thiscall.rs
@@ -1,5 +1,5 @@
 // run-pass
-// only-i686
+// only-x86
 
 #![feature(abi_thiscall)]
 
--- a/src/test/ui/extern/extern-vectorcall.rs
+++ b/src/test/ui/extern/extern-vectorcall.rs
@@ -1,7 +1,7 @@
 // run-pass
 // revisions: x64 x32
 // [x64]only-x86_64
-// [x32]only-i686
+// [x32]only-x86
 
 #![feature(abi_vectorcall)]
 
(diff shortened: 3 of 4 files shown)


Mark-Simulacrum:
Hm, how do we ensure new tests don't get added? I would have expected the panic! in that code to get triggered... maybe we can fix that as well?

wesleywiser:
The panic isn't triggered because we do find the arch but in the calling code, it doesn't match the `only-` directive:

https://github.com/rust-lang/rust/blob/27143a9094b55a00d5f440b05b0cb4233b300d33/src/tools/compiletest/src/header.rs#L689

Mark-Simulacrum:
Right, I guess, what I'm trying to get at is that it seems like we *should* error on such a test that can never run. I'm not sure I'm fully understanding, but perhaps we should add an assert whenever `name` is present on the "left hand" of the mapping table?

Alternatively, we could *also* call the equivalent of get_arch on the `name` that maps it, but I'd prefer we don't have both only-i686 and only-x86 tests.

wesleywiser:
Right. I agree, ideally we would be able to detect an invalid `only-` flag which will cause the test to never run but compiletest isn't really set up to do that right now and it would be a much larger change than just fixing this. I'm also a little hesitant to introduce validation just for architectures as I think developers would naturally expect the other kinds of flags to be validated as well. 

We could certainly file an issue to improve compiletest though. 

>  I'm not sure I'm fully understanding, but perhaps we should add a panic whenever name is present on the "left hand" of the mapping table?

We also can't do that because some names are present as both keys and values in the mapping table (`arm`, `aarch64`, `avx`, etc).

Mark-Simulacrum:
Alright. r=me with an issue filed.

Did you spot check the other mappings in the table to check we don't have extra tests like this (for non-i686)?

wesleywiser:
Yes, I didn't see any other issues.

@bors r=Mark-Simulacrum rollup

PR #90571, opened 2021-11-04
Fix missing bottom border for headings in sidebar

### LINKED ISSUE #90568: Titles in the sidebar aren't underlined anymore
Current beta:

Screenshot from 2021-11-04 15-30-47

Current nightly:

Screenshot from 2021-11-04 15-30-51

It was caused by https://github.com/rust-lang/rust/pull/90156.

GuillaumeGomez:
Fixes #90568.

r? @jsha

### DIFF
--- a/src/librustdoc/html/static/css/rustdoc.css
+++ b/src/librustdoc/html/static/css/rustdoc.css
@@ -156,7 +156,8 @@ h1.fqn > .in-band > a:hover {
 	section hierarchies. */
 h2,
 .top-doc h3,
-.top-doc h4 {
+.top-doc h4,
+.sidebar .others h3 {
 	border-bottom: 1px solid;
 }
 h3.code-header {
--- a/src/librustdoc/html/static/js/main.js
+++ b/src/librustdoc/html/static/js/main.js
@@ -566,6 +566,7 @@ function hideThemeButtonState() {
     // delayed sidebar rendering.
     window.initSidebarItems = function(items) {
         var sidebar = document.getElementsByClassName("sidebar-elems")[0];
+        var others;
         var current = window.sidebarCurrent;
 
         function addSidebarCrates(crates) {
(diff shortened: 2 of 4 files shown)


src/librustdoc/html/static/css/rustdoc.css
```diff
@@ -156,7 +156,9 @@ h1.fqn > .in-band > a:hover {
 	section hierarchies. */
 h2,
 .top-doc h3,
-.top-doc h4 {
+.top-doc h4,
+.sidebar .sidebar-elems h3:not(.sidebar-title),
```
jsha, line 160: I find the `not` selectors are pretty hard to maintain and reason about. Can we do this with a positive selector instead? For instance, it looks like the headings we need underlines on here are nested under `.block`, so maybe we could do `.sidebar .block h3`?

Also it's not clear to me why both `.sidebar` and `.sidebar-elems` are needed in this selector.

jsha (changes requested):
Thanks for catching and fixing this!

src/librustdoc/html/static/css/rustdoc.css
GuillaumeGomez: Except that the `.sidebar-title` are in `.block` as well. :3

It was more complicated than I expected to actually add this rule without breaking something.

jsha: What about selecting the specific block types that get the underline treatment? Like:

```
.sidebar .block.struct h3, .sidebar .block.trait h3, .sidebar .block.type, ... {
```

Alternately, we could change the HTML structure slightly so everything after "Other items in ..." is nested under a specific class, like `.other-items`. Then we could make this selector `.sidebar .other-items h3`.

GuillaumeGomez: I think I'd prefer to have a CSS class in the generated HTML, seems much simpler.

GuillaumeGomez:
@jsha It now looks much better, thanks for the suggestion!

jsha:
Looks great. r=me once CI passes

jsha (approved):

PR #90575, opened 2021-11-04
Improve suggestions for compatible variants on type mismatch.

### LINKED ISSUE #90553: Special case suggestion for missing tail expression when dealing with `Ok(())` and `Some(())`
The suggestion in the following code isn't great:

```
  --> src/main.rs:72:5
   |
72 | /     while let Some(i) = x.try_next().await? {
73 | |         println!("{:?}", i);
74 | |     }
   | |_____^ expected enum `Option`, found `()`
   |
   = note:   expected enum `Option<()>`
           found unit type `()`
help: try using a variant of the expected enum
   |
72 ~     Some(while let Some(i) = x.try_next().await? {
73 +         println!("{:?}", i);
74 +     })
   |
```

It makes sense why it happens: the `while let` expression resolves to type `()`, it's the tail expression of an `async fn` that returns `Option<()>`, so wrapping the expression in `Some(())` would make the type checker happy, but suggesting that seems almost always wrong? I think I would prefer it if we special cased this and didn't gave a suggestion for `Result<(), _>` or `Option<()>`.

m-ou-se:
Fixes #90553.

Before: 
image

After: 
image

r? @estebank

### DIFF
--- a/compiler/rustc_span/src/source_map.rs
+++ b/compiler/rustc_span/src/source_map.rs
@@ -593,14 +593,19 @@ impl SourceMap {
     }
 
     pub fn span_to_margin(&self, sp: Span) -> Option<usize> {
-        match self.span_to_prev_source(sp) {
-            Err(_) => None,
-            Ok(source) => {
-                let last_line = source.rsplit_once('\n').unwrap_or(("", &source)).1;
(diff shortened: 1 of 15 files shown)


estebank:
r=me after rebasing

estebank (approved):

PR #90577, opened 2021-11-04
clippy::perf fixes

matthiaskrgr:


library/test/src/formatters/junit.rs
```diff
@@ -56,7 +56,7 @@ impl<T: Write> OutputFormatter for JunitFormatter<T> {
         _state: &ConsoleTestState,
     ) -> io::Result<()> {
         // Because the testsuit node holds some of the information as attributes, we can't write it
```
calebcartwright, line 58: Given the typo correction below, I think these other two may be worth fixing at the same time?

```suggestion
        // Because the testsuite node holds some of the information as attributes, we can't write it
```

library/test/src/formatters/junit.rs
```diff
@@ -56,7 +56,7 @@ impl<T: Write> OutputFormatter for JunitFormatter<T> {
...
-        // until all of the tests has ran. Instead of writting every result as they come in, we add
+        // until all of the tests has ran. Instead of writing every result as they come in, we add
```
calebcartwright, line 59:
```suggestion
        // until all of the tests have ran. Instead of writing every result as they come in, we add
```

petrochenkov:
Nothing here is going to actually affect performance in practice, so I'm going to just @bors r+ without an additional perf run.

PR #79291, opened 2020-11-22
Add error message for private fn

JulianKnodt:
Attempts to add a more detailed error when a `const_evaluatable` fn from another scope is used inside of a scope which cannot access it. 

r? @lcnr

PR #90586, opened 2021-11-04
Relax priv-in-pub lint on generic bounds and where clauses of trait impls.

jswrenn:
The priv-in-pub lint is a legacy mechanism of the compiler, supplanted by a reachability-based [type privacy](https://github.com/rust-lang/rfcs/blob/master/text/2145-type-privacy.md) analysis. This PR does **not** relax type privacy; it only relaxes the lint (as proposed by the type privacy RFC) in the case of trait impls.

## Current Behavior
On public trait impls, it's currently an **error** to have a `where` bound constraining a private type with a trait:
```rust
pub trait Trait {}
pub struct Type {}

struct Priv {}
impl Trait for Priv {}

impl Trait for Type
where
    Priv: Trait // ERROR
{}
```

...and it's a **warning** to have have a public type constrained by a private trait:
```rust
pub trait Trait {}
pub struct Type {}

pub struct Pub {}
trait Priv {}
impl Priv for Pub {}

impl Trait for Type
where
    Pub: Priv // WARNING
{}
```

This lint applies to `where` clauses in other contexts, too; e.g. on free functions:
```rust
struct Priv<T>(T);
pub trait Pub {}
impl<T: Pub> Pub for Priv<T> {}

pub fn function<T>()
where
    Priv<T>: Pub // WARNING
{}
```

**These constraints could be relaxed without issue.**

## New Behavior
This lint is relaxed for `where` clauses on trait impls, such that it's okay to have a `where` bound constraining a private type with a trait:
```rust
pub trait Trait {}
pub struct Type {}

struct Priv {}
impl Trait for Priv {}

impl Trait for Type
where
    Priv: Trait // OK
{}
```

...and it's okay to have a public type constrained by a private trait:
```rust
pub trait Trait {}
pub struct Type {}

pub struct Pub {}
trait Priv {}
impl Priv for Pub {}

impl Trait for Type
where
    Pub: Priv // OK
{}
```

## Rationale
While the priv-in-pub lint is not essential for soundness, it *can* help programmers avoid pitfalls that would make their libraries difficult to use by others. For instance, such a lint *is* useful for free functions; e.g. if a downstream crate tries to call the `function` in the previous snippet in a generic context:
```rust
fn callsite<T>()
where
    Priv<T>: Pub // ERROR: omitting this bound is a compile error, but including it is too
{
    function::<T>()
}
```
...it cannot do so without repeating `function`'s `where` bound, which we cannot do because `Priv` is out-of-scope. A lint for this case is arguably helpful.

However, this same reasoning **doesn't** hold for trait impls. To call an unconstrained method on a public trait impl with private bounds, you don't need to forward those private bounds, you can forward the public trait:
```rust
mod upstream {
    pub trait Trait {
        fn method(&self) {}
    }
    pub struct Type<T>(T);
    
    pub struct Pub<T>(T);
    trait Priv {}
    impl<T: Priv> Priv for Pub<T> {}
    
    impl<T> Trait for Type<T>
    where
        Pub<T>: Priv // WARNING
    {}
}

mod downstream {
    use super::upstream::*;
    
    fn function<T>(value: Type<T>)
    where
        Type<T>: Trait // <- no private deets!
    {
        value.method();
    }
}
```

**This PR only eliminates the lint on trait impls.** It leaves it intact for all other contexts, including trait definitions, inherent impls, and function definitions. It doesn't need to exist in those cases either, but I figured I'd first target a case where it's mostly pointless.

## Other Notes
- See discussion [on zulip](https://rust-lang.zulipchat.com/#narrow/stream/213817-t-lang/topic/relax.20priv-in-pub.20lint.20for.20trait.20impl.20.60where.60.20bounds/near/222458397).
- This PR effectively reverts #79291.

### DIFF
--- a/compiler/rustc_privacy/src/lib.rs
+++ b/compiler/rustc_privacy/src/lib.rs
@@ -2069,7 +2069,11 @@ impl<'tcx> Visitor<'tcx> for PrivateItemsInPublicInterfacesVisitor<'tcx> {
             // Subitems of trait impls have inherited publicity.
             hir::ItemKind::Impl(ref impl_) => {
                 let impl_vis = ty::Visibility::of_impl(item.def_id, tcx, &Default::default());
-                self.check(item.def_id, impl_vis).generics().predicates();
+                // check that private components do not appear in the generics or predicates of inherent impls
+                // this check is intentionally NOT performed for impls of traits, per #90586
+                if impl_.of_trait.is_none() {
+                    self.check(item.def_id, impl_vis).generics().predicates();
+                }
                 for impl_item_ref in impl_.items {
                     let impl_item_vis = if impl_.of_trait.is_none() {
                         min(tcx.visibility(impl_item_ref.id.def_id), impl_vis, tcx)
--- a/src/test/ui/const-generics/generic_const_exprs/eval-privacy.rs
+++ b/src/test/ui/const-generics/generic_const_exprs/eval-privacy.rs
@@ -9,12 +9,7 @@ pub trait Trait {
     fn assoc_fn() -> Self::AssocTy;
 }
 
-impl<const U: u8> Trait for Const<U>
-//~^ WARN private type
-//~| WARN this was previously
-//~| WARN private type
-//~| WARN this was previously
-
+impl<const U: u8> Trait for Const<U> // OK, trait impl predicates
 where
     Const<{ my_const_fn(U) }>: ,
 {
--- a/src/test/ui/privacy/private-in-public-warn.rs
+++ b/src/test/ui/privacy/private-in-public-warn.rs
@@ -63,8 +63,7 @@ mod traits {
     }
     impl<T: PrivTr> Pub<T> {} //~ ERROR private trait `traits::PrivTr` in public interface
         //~^ WARNING hard error
-    impl<T: PrivTr> PubTr for Pub<T> {} //~ ERROR private trait `traits::PrivTr` in public interface
-        //~^ WARNING hard error
+    impl<T: PrivTr> PubTr for Pub<T> {} // OK, trait impl predicates
 }
 
 mod traits_where {
@@ -87,9 +86,7 @@ mod traits_where {
     impl<T> Pub<T> where T: PrivTr {}
         //~^ ERROR private trait `traits_where::PrivTr` in public interface
         //~| WARNING hard error
-    impl<T> PubTr for Pub<T> where T: PrivTr {}
-        //~^ ERROR private trait `traits_where::PrivTr` in public interface
-        //~| WARNING hard error
+    impl<T> PubTr for Pub<T> where T: PrivTr {} // OK, trait impl predicates
 }
 
 mod generics {
--- a/src/test/ui/privacy/where-priv-type.rs
+++ b/src/test/ui/privacy/where-priv-type.rs
@@ -1,0 +1,90 @@
+// priv-in-pub lint tests where the private type appears in the
+// `where` clause of a public item
+
+#![crate_type = "lib"]
+#![feature(generic_const_exprs)]
+#![allow(incomplete_features)]
+
+
+struct PrivTy;
+trait PrivTr {}
+pub struct PubTy;
+pub struct PubTyGeneric<T>(T);
+pub trait PubTr {}
+impl PubTr for PrivTy {}
+pub trait PubTrWithAssocTy { type AssocTy; }
+impl PubTrWithAssocTy for PrivTy { type AssocTy = PrivTy; }
+
+
+pub struct S
+//~^ WARNING private type `PrivTy` in public interface
+//~| WARNING hard error
+where
+    PrivTy:
+{}
+
+
+pub enum E
+//~^ WARNING private type `PrivTy` in public interface
+//~| WARNING hard error
+where
+    PrivTy:
+{}
+
+
+pub fn f()
+//~^ WARNING private type `PrivTy` in public interface
+//~| WARNING hard error
+where
+    PrivTy:
+{}
+
+
+impl S
+//~^ ERROR private type `PrivTy` in public interface
+where
+    PrivTy:
+{
+    pub fn f()
+    //~^ WARNING private type `PrivTy` in public interface
+    //~| WARNING hard error
+    where
+        PrivTy:
+    {}
+}
+
+
+impl PubTr for PubTy
+where
+    PrivTy:
+{}
+
+
+impl<T> PubTr for PubTyGeneric<T>
+where
+    T: PubTrWithAssocTy<AssocTy=PrivTy>
+{}
+
+
+pub struct Const<const U: u8>;
+
+pub trait Trait {
+    type AssocTy;
+    fn assoc_fn() -> Self::AssocTy;
+}
+
+impl<const U: u8> Trait for Const<U>
+where
+    Const<{ my_const_fn(U) }>: ,
+{
+    type AssocTy = Const<{ my_const_fn(U) }>;
+    //~^ ERROR private type
+    fn assoc_fn() -> Self::AssocTy {
+        Const
+    }
+}
+
+const fn my_const_fn(val: u8) -> u8 {
+    // body of this function doesn't matter
+    val
+}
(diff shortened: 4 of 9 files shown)


petrochenkov:
I'm going to send this to the lang team for sanity checking.

This PR removes the private-in-public warnings from generic bounds and where clauses on `impl` items specifically.
- The behavior introduced by this PR matches the "Type privacy" RFC precisely if the `private_bounds` lint (https://github.com/rust-lang/rfcs/blob/master/text/2145-type-privacy.md#lint-2-private-traitstypes-in-secondary-interface-of-effectively-public-items) ends up being allow-by-default (which seems somewhat reasonable), or will similarly not be reported for `impl` items specifically.
- The behavior introduced by this PR will differ from the "Type privacy" RFC, if the `private_bounds` lint ends up being warn-by-default, in that case code allowed by this PR will start reporting warnings.

nikomatsakis:
@petrochenkov I cannot quite tell from your comment what your take is here -- do you feel this is in the spirit of the original RFC? In what way does it differ? It seems surprising to me that you would say it conforms or not to the RFC based on whether the lint in the RFC is allow by default.

petrochenkov:
>do you feel this is in the spirit of the original RFC?

It's in the spirit of the RFC.
The RFC is just not entirely precise

>This lint replaces part of private-in-public errors and can be reported [**as warn-by-default or allow-by-default**](https://github.com/rust-lang/rfcs/blob/master/text/2145-type-privacy.md#lint-2-private-traitstypes-in-secondary-interface-of-effectively-public-items).

Back then I wanted to start with warn-by-default to get some experience, but maybe be we already have enough experience to just allow it now.

jswrenn:
This PR also draws a finer distinction than [the proposed Lint #2 makes](https://github.com/rust-lang/rfcs/blob/master/text/2145-type-privacy.md#lint-2-private-traitstypes-in-secondary-interface-of-effectively-public-items): this PR distinguishes between private items in the bounds of trait impls, and private items in other bounds. I describe the rationale for this distinction in the description of this PR, and this PR only impacts private items in the `where` bounds of trait impls.

---

@scottmcm [suggested](https://rust-lang.zulipchat.com/#narrow/stream/213817-t-lang/topic/relax.20priv-in-pub.20lint.20for.20trait.20impl.20.60where.60.20bounds/near/222483732) that all uses of private items in bounds should be allowed, but this PR doesn't go that far. I'm inclined to agree, though. The only danger of private items in bounds is that end-users can't 'forward' such bounds when they're writing generic code. But:
- not all bounds *need* to be forwarded
- you can already write un-forwardable bounds, just by adding a tad of indirection
- sometimes unforwardable bounds are desirable

So, the lint proposed proposed by the type privacy RFC would just a heuristic that _sometimes_ helps avoid an API design decision that _sometimes_ is a pitfall. There's enough "sometimes"s in there, that I'd argue such a lint would probably be more at home in clippy than rustc.

scottmcm:
FWIW, I like the middle-ground proposed here.  The OP did a good job convincing me that the "you can forward the public trait" is an interesting distinction.  We can always remove more cases later.

scottmcm:
We didn't have time to go into this in detail in today's lang team meeting, so let's try to make progress async:

@rfcbot fcp merge

The OP here has a good description of what this allows.  People seem to want to do this enough that the workarounds are well-known; I think we should just allow it directly.

nikomatsakis:
@rfcbot reviewed

src/test/ui/privacy/where-priv-type.rs
```diff
@@ -0,0 +1,83 @@
+#![crate_type = "lib"]
```
petrochenkov, line 1: Could you add a comment explaining what this test is testing in a couple of sentences?

petrochenkov: (And to the second new test too.)

src/test/ui/privacy/private-in-public-warn.rs
```diff
@@ -85,11 +83,7 @@ mod traits_where {
         //~| WARNING hard error
     }
     impl<T> Pub<T> where T: PrivTr {}
-        //~^ ERROR private trait `traits_where::PrivTr` in public interface
-        //~| WARNING hard error
     impl<T> PubTr for Pub<T> where T: PrivTr {}
```
petrochenkov, line 86:
```suggestion
    impl<T> PubTr for Pub<T> where T: PrivTr {} // OK, trait impl predicates
```

src/test/ui/privacy/private-in-public-warn.rs
```diff
@@ -61,10 +61,8 @@ mod traits {
         fn f<T: PrivTr>(arg: T) {} //~ ERROR private trait `traits::PrivTr` in public interface
         //~^ WARNING hard error
     }
-    impl<T: PrivTr> Pub<T> {} //~ ERROR private trait `traits::PrivTr` in public interface
-        //~^ WARNING hard error
-    impl<T: PrivTr> PubTr for Pub<T> {} //~ ERROR private trait `traits::PrivTr` in public interface
-        //~^ WARNING hard error
+    impl<T: PrivTr> Pub<T> {}
+    impl<T: PrivTr> PubTr for Pub<T> {}
```
petrochenkov, line 65:
```suggestion
    impl<T: PrivTr> PubTr for Pub<T> {} // OK, trait impl predicates
```

src/test/ui/const-generics/generic_const_exprs/eval-privacy.rs
```diff
@@ -10,11 +10,6 @@ pub trait Trait {
 }
 
 impl<const U: u8> Trait for Const<U>
```
petrochenkov, line 12:
```suggestion
impl<const U: u8> Trait for Const<U> // OK, trait impl predicates
```

compiler/rustc_privacy/src/lib.rs
```diff
@@ -2069,7 +2069,6 @@ impl<'tcx> Visitor<'tcx> for PrivateItemsInPublicInterfacesVisitor<'tcx> {
             // Subitems of trait impls have inherited publicity.
             hir::ItemKind::Impl(ref impl_) => {
                 let impl_vis = ty::Visibility::of_impl(item.def_id, tcx, &Default::default());
-                self.check(item.def_id, impl_vis).generics().predicates();
```
petrochenkov, line 2072: This change affects inherent impls too, which contradicts the PR description.

petrochenkov: Also, could you add a comment saying that generic parameters and where clauses are specifically NOT checked, and it's not an accidental omission, and link to this PR.

jswrenn: Fixed with https://github.com/rust-lang/rust/pull/90586/commits/6a5c0d6039e6095edfb302b4902997b76c46ce35!

src/test/ui/privacy/where-priv-type.rs
jswrenn: Resolved by https://github.com/rust-lang/rust/pull/90586/commits/932a34ce2005ccfae225aab2bf424e7233d54e48.

src/test/ui/privacy/private-in-public-warn.rs
```diff
@@ -85,11 +84,9 @@ mod traits_where {
         //~| WARNING hard error
     }
     impl<T> Pub<T> where T: PrivTr {}
-        //~^ ERROR private trait `traits_where::PrivTr` in public interface
-        //~| WARNING hard error
-    impl<T> PubTr for Pub<T> where T: PrivTr {}
-        //~^ ERROR private trait `traits_where::PrivTr` in public interface
+        //~^ ERROR private trait `traits_where::PrivTr` in public interfac
```
petrochenkov, line 87:
```suggestion
        //~^ ERROR private trait `traits_where::PrivTr` in public interface
```

src/test/ui/privacy/private-in-public.rs
```diff
@@ -31,7 +31,7 @@ mod traits {
     pub enum E<T: PrivTr> { V(T) } //~ ERROR private trait `traits::PrivTr` in public interface
     pub fn f<T: PrivTr>(arg: T) {} //~ ERROR private trait `traits::PrivTr` in public interface
     pub struct S1<T: PrivTr>(T); //~ ERROR private trait `traits::PrivTr` in public interface
-    impl<T: PrivTr> Pub<T> { //~ ERROR private trait `traits::PrivTr` in public interface
+    impl<T: PrivTr> Pub<T> {  //~ ERROR private trait `traits::PrivTr` in public interface
```
petrochenkov, line 34: Nit: unnecessary whitespace change.

petrochenkov:
r=me after squashing commits and removing accidental changes from the resulting commit.

petrochenkov (approved):

ehuss:
@bors retry

network error

PR #90596, opened 2021-11-05
Optimize Eq and Hash for Path/PathBuf

the8472:
```
# new

test path::tests::bench_hash_path_long                            ... bench:          86 ns/iter (+/- 1)
test path::tests::bench_hash_path_short                           ... bench:          13 ns/iter (+/- 1)
test path::tests::bench_path_hashset                              ... bench:         197 ns/iter (+/- 6)
test path::tests::bench_path_hashset_miss                         ... bench:          94 ns/iter (+/- 4)

# old

test path::tests::bench_hash_path_long                            ... bench:         192 ns/iter (+/- 2)
test path::tests::bench_hash_path_short                           ... bench:          33 ns/iter (+/- 1)
test path::tests::bench_path_hashset                              ... bench:       1,121 ns/iter (+/- 24)
test path::tests::bench_path_hashset_miss                         ... bench:         273 ns/iter (+/- 6)
```

### DIFF
--- a/library/std/src/path.rs
+++ b/library/std/src/path.rs
@@ -1033,12 +1033,11 @@ fn compare_components(mut left: Components<'_>, mut right: Components<'_>) -> cm
     // the middle of one
     if left.prefix.is_none() && right.prefix.is_none() && left.front == right.front {
         // possible future improvement: a [u8]::first_mismatch simd implementation
-        let first_difference =
-            match left.path.iter().zip(right.path.iter()).position(|(&a, &b)| a != b) {
-                None if left.path.len() == right.path.len() => return cmp::Ordering::Equal,
-                None => left.path.len().min(right.path.len()),
-                Some(diff) => diff,
-            };
+        let first_difference = match left.path.iter().zip(right.path).position(|(&a, &b)| a != b) {
+            None if left.path.len() == right.path.len() => return cmp::Ordering::Equal,
+            None => left.path.len().min(right.path.len()),
+            Some(diff) => diff,
+        };
 
         if let Some(previous_sep) =
             left.path[..first_difference].iter().rposition(|&b| left.is_sep_byte(b))
--- a/library/std/src/path.rs
+++ b/library/std/src/path.rs
@@ -2873,9 +2873,35 @@ impl cmp::PartialEq for Path {
 #[stable(feature = "rust1", since = "1.0.0")]
 impl Hash for Path {
     fn hash<H: Hasher>(&self, h: &mut H) {
-        for component in self.components() {
-            component.hash(h);
+        let bytes = self.as_u8_slice();
+
+        let mut component_start = 0;
+        let mut bytes_hashed = 0;
+
+        for i in 0..bytes.len() {
+            if is_sep_byte(bytes[i]) {
+                if i > component_start {
+                    let to_hash = &bytes[component_start..i];
+                    h.write(to_hash);
+                    bytes_hashed += to_hash.len();
+                }
+
+                // skip over separator and optionally a following CurDir item
+                // since components() would normalize these away
+                component_start = i + match bytes[i..] {
+                    [_, b'.', b'/', ..] | [_, b'.'] => 2,
+                    _ => 1,
+                };
+            }
         }
+
+        if component_start < bytes.len() {
+            let to_hash = &bytes[component_start..];
+            h.write(to_hash);
+            bytes_hashed += to_hash.len();
+        }
+
+        h.write_usize(bytes_hashed);
     }
 }
 
--- a/library/std/src/path.rs
+++ b/library/std/src/path.rs
@@ -2892,6 +2892,14 @@ impl cmp::PartialEq for Path {
 impl Hash for Path {
     fn hash<H: Hasher>(&self, h: &mut H) {
         let bytes = self.as_u8_slice();
+        let prefix_len = match parse_prefix(&self.inner) {
+            Some(prefix) => {
+                prefix.hash(h);
+                prefix.len()
+            }
+            None => 0,
+        };
+        let bytes = &bytes[prefix_len..];
 
         let mut component_start = 0;
         let mut bytes_hashed = 0;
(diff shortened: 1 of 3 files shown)


library/std/src/path.rs
```diff
@@ -2873,9 +2873,35 @@ impl cmp::PartialEq for Path {
...
+                if i > component_start {
+                    let to_hash = &bytes[component_start..i];
+                    h.write(to_hash);
+                    bytes_hashed += to_hash.len();
+                }
+
+                // skip over the separator and one immediately following '.' (if present)
+                // since components() would normalize that away
+                component_start = i + match bytes[i..] {
+                    [_, b'.', ..] => 2,
```
LingMan, line 2905: Not every `.` directly after a separator is necessarily a `CurDir` component. E.g. `/home/user/.ssh`

the8472: Yes, but this is for `Hash`, it should be OK to calculate the same hash for `user/.ssh` and `user/ssh` as long as it doesn't cause too many collisions. This will only allow 1 bit worth of collisions per path separator.

LingMan: OK, from the commit message I initially thought your intent was to match the original splitting behavior exactly. Nvm then.

the8472: Well, it's possible to make it more precise but that costs a few cycles. I'm not sure if it's worth it.

the8472: After thinking about it some more I improved the normalization so this case can't be exploited to cause hash collisions.

the8472:
Almost entirely improvements except for `webrender-wrench` which spends more time in llvm. I guess that it makes use of a lot of paths and has to spend more time to optimize it, likely trading compile time for runtime in that case.

joshtriplett:
This seems like a reasonable set of performance numbers to me; it's paying compile time to get more runtime performance.

library/std/src/path.rs
```diff
@@ -979,6 +979,19 @@ impl FusedIterator for Components<'_> {}
 impl<'a> cmp::PartialEq for Components<'a> {
     #[inline]
     fn eq(&self, other: &Components<'a>) -> bool {
+        // fast path for exact matches, e.g. for hashmap lookups
+        if self.path.len() == other.path.len()
+            && self.front == other.front
+            && self.back == State::Body
+            && other.back == State::Body
+        {
+            // this could bail out earlier if there were a reverse memcmp/bcmp comparing back to front
```
Mark-Simulacrum, line 988: Can we phrase this as a "future improvement" or something? Currently, it's not bailing out early, right? This comment makes that a little unclear.

library/std/src/path.rs
```diff
@@ -979,6 +979,19 @@ impl FusedIterator for Components<'_> {}
...
+            if self.path == other.path {
```
Mark-Simulacrum, line 989: I'm a little bit wary of this since it doesn't check all the fields -- I suppose we can ignore them because we're already in State::Body (in particular, has_physical_root and prefix)? I think it'd be good to write up a comment explaining why we can do so and potentially adding an explicit de-structuring matching on every field so that we need to edit this code whenever Components changes.

library/std/src/path.rs
```diff
@@ -2873,9 +2886,35 @@ impl cmp::PartialEq for Path {
...
+            }
         }
+
+        if component_start < bytes.len() {
+            let to_hash = &bytes[component_start..];
+            h.write(to_hash);
+            bytes_hashed += to_hash.len();
+        }
+
+        h.write_usize(bytes_hashed);
```
Mark-Simulacrum, line 2917: This seems OK -- I'm not immediately seeing anything wrong -- but trying to map this code as being equivalent to the component iterator equality seems quite difficult. Is there something we can do to build confidence it's correct, particularly as we potentially change the implementation of components()?

Perhaps we could instead improve the performance of next() on Components for similar effect?

the8472: >  Is there something we can do to build confidence it's correct, particularly as we potentially change the implementation of components()?

Tests make sure that Ord and Eq are now consistent with hashing.

> Perhaps we could instead improve the performance of next() on Components for similar effect?

I'm not sure if that's possible without regressing other things. E.g. slapping `#[inline]` on some of the functions may regress compile times on things not related to hashing.

`components` also have to operate a more complex state machine to produce the enums that it does, we're not interested in the finer distinctions here.

Some of the costs also came from hashing each component separately which is treated as a slice which the hashes the slice length which doesn't happen in the new code.

library/std/src/path.rs
the8472: > I suppose we can ignore them because we're already in State::Body

We can ignore them because `Eq` only compares the `Component` items, not the prefix. If we're in `front == State::Body` then `prefix` is also covered by `path`. If `Components` has already been partially advanced then the `path` slice doesn't cover it anymore and thus wouldn't be part of the comparison.

But there might be some edge-cases on windows how `CurDir` items are interpreted on verbatim vs. non-verbatim paths so even when the suffix is the same it may have different meaning.

So your concern is warranted. I'll add some extra condition and more comments to check those cases too.

the8472: I added comments, exra checks and the destructuring. But that destructuring is ugly and overly verbose imo, I would prefer not to have that.

Mark-Simulacrum: Oh, sorry, I think you may have misinterpreted / I was less clear than desirable -- I meant a sort no-op destructuring at the top of the method `let Foo { field1: _, field2: _ } = foo` so that we edit the method if fields are added/removed, but there's no need to make it actually part of the real code.

Mark-Simulacrum:
OK, this looks good to me. I think the commit history can be cleaned up a little -- 8 commits seems a little too many :)

But otherwise r=me.

the8472:
Apparently the drive prefixes are the *only* thing where we do case-insensitive comparison on paths. 😮‍💨
Not sure yet how to solve that efficiently.

ChrisDenton:
> Apparently the drive prefixes are the _only_ thing where we do case-insensitive comparison on paths.
They are the only place that are guaranteed to be ASCII case-insensitive. Other parts of the path could be case-sensitive.

the8472:
@Mark-Simulacrum ready for another review, the latest commit should fix the windows failure.

PR #90597, opened 2021-11-05
Warn for variables that are no longer captured

### LINKED ISSUE #90465: Missing warning for migration of `let _ = x` between Rust 2018 and 2021 edition
The following example  does not get a migration warning in Rust 2018 (or Rust 2021) -- ([playground](https://play.rust-lang.org/?version=nightly&mode=debug&edition=2018&gist=8841e3a3e754a39fa51f448edb136af1)):

```rust

fn main() {
    struct Foo(u32);
    impl Drop for Foo {
        fn drop(&mut self) {
            println!("dropped {}", self.0);
        }
    }
    
    let f0 = Foo(0);
    let f1 = Foo(1);
    
    let c0 = move || {
        let _ = f0;
    };
    
    let c1 = move || {
        let _ = &f1;
    };
    
    println!("dropping 0");
    drop(c0);
    println!("dropping 1");
    drop(c1);
    println!("dropped all");
}
```

The behavior of `let _ = x` changed in the new edition, but there ought to have been a warning.

<details>
<summary>Original issue</summary>

See the below code

```rust
fn main() {
    struct Foo(u32);
    impl Drop for Foo {
        fn drop(&mut self) {
            println!("dropped {}", self.0);
        }
    }
    
    let f0 = Foo(0);
    let f1 = Foo(1);
    
    let c0 = move || {
        let _ = f0;
    };
    
    let c1 = move || {
        let _ = &f1;
    };
    
    println!("dropping 0");
    drop(c0);
    println!("dropping 1");
    drop(c1);
    println!("dropped all");
}
```

With 2018 edition ([playground link](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=e8b0d8c2e4d9709b40639bf4efa05a88)) this prints

```
dropping 0
dropped 0
dropping 1
dropped 1
dropped all
```

With 2021 edition ([playground link](https://play.rust-lang.org/?version=nightly&mode=debug&edition=2021&gist=489fea42e35157e720ffef1d3fe6bffc)) this prints

```
dropping 0
dropping 1
dropped 1
dropped all
dropped 0
```

That is, with the 2021 edition the first closure is apparently not capturing `f0` at all (and drops it at the end of `main`) while with the 2018 edition it did.

This might be intentional and expected or not, which is why I'm creating this issue. It might be worth mentioning it as a potential porting trap in the edition guide. Note that `cargo fix --edition` did not catch this one (but a couple of others where the behaviour difference was irrelevant).

This subtle behaviour change caused [a test in gtk-rs](https://github.com/gtk-rs/gtk-rs-core/pull/321) to timeout and fail.

</details>

nikomatsakis:
r? @wesleywiser 

cc @rust-lang/wg-rfc-2229 

Fixes #90465

### DIFF
--- a/compiler/rustc_typeck/src/check/upvar.rs
+++ b/compiler/rustc_typeck/src/check/upvar.rs
@@ -88,19 +88,53 @@ impl<'a, 'tcx> FnCtxt<'a, 'tcx> {
 /// name (i.e: a.b.c)
 #[derive(Clone, Debug, PartialEq, Eq, PartialOrd, Ord, Hash)]
 enum CapturesInfo {
+    /// We previously captured all of `x`, but now we capture some sub-path.
     CapturingLess { source_expr: Option<hir::HirId>, var_name: String },
+    //CapturingNothing {
+    //    // where the variable appears in the closure (but is not captured)
+    //    use_span: Span,
+    //},
 }
 
-/// Intermediate format to store information needed to generate migration lint. The tuple
-/// contains the hir_id pointing to the use that resulted in the
-/// corresponding place being captured, a String which contains the captured value's
-/// name (i.e: a.b.c) and a String which contains the reason why migration is needed for that
-/// capture
-type MigrationNeededForCapture = (Option<hir::HirId>, String, String);
+/// Reasons that we might issue a migration warning.
+#[derive(Clone, Debug, Default, PartialEq, Eq, PartialOrd, Ord, Hash)]
+struct MigrationWarningReason {
+    /// When we used to capture `x` in its entirety, we implemented the auto-trait(s)
+    /// in this vec, but now we don't.
+    auto_traits: Vec<&'static str>,
+
+    /// When we used to capture `x` in its entirety, we would execute some destructors
+    /// at a different time.
+    drop_order: bool,
+}
+
+impl MigrationWarningReason {
+    fn migration_message(&self) -> String {
+        let base = "changes to closure capture in Rust 2021 will affect";
+        if !self.auto_traits.is_empty() && self.drop_order {
+            format!("{} drop order and which traits the closure implements", base)
+        } else if self.drop_order {
+            format!("{} drop order", base)
+        } else {
+            format!("{} which traits the closure implements", base)
+        }
+    }
+}
+
+/// Intermediate format to store information needed to generate a note in the migration lint.
+struct MigrationLintNote {
+    captures_info: CapturesInfo,
+
+    /// reasons why migration is needed for this capture
+    reason: MigrationWarningReason,
+}
 
 /// Intermediate format to store the hir id of the root variable and a HashSet containing
 /// information on why the root variable should be fully captured
-type MigrationDiagnosticInfo = (hir::HirId, Vec<MigrationNeededForCapture>);
+struct NeededMigration {
+    var_hir_id: hir::HirId,
+    diagnostics_info: Vec<MigrationLintNote>,
+}
 
 struct InferBorrowKindVisitor<'a, 'tcx> {
     fcx: &'a FnCtxt<'a, 'tcx>,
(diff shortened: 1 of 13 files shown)


nikomatsakis:
Nominating for backport: this bug fixes incorrect migration that can cause subtle changes in execution order (see #90465)

compiler/rustc_typeck/src/check/upvar.rs
```diff
@@ -997,21 +1046,37 @@ impl<'a, 'tcx> FnCtxt<'a, 'tcx> {
...
+            // let x = move || {
+            //     let _ = y;
+            // });
+            // ```
+            debug!("no path starting from it is used");
+
 
             match closure_clause {
                 // Only migrate if closure is a move closure
-                hir::CaptureBy::Value => return Some(FxHashSet::default()),
```
arora-aman, line 1009: the error was this is empty? Little confusing what fixed it along side all the refactoring.

compiler/rustc_typeck/src/check/upvar.rs
```diff
@@ -86,18 +86,55 @@ impl<'a, 'tcx> FnCtxt<'a, 'tcx> {
 /// Intermediate format to store the hir_id pointing to the use that resulted in the
 /// corresponding place being captured and a String which contains the captured value's
 /// name (i.e: a.b.c)
-type CapturesInfo = (Option<hir::HirId>, String);
+#[derive(Clone, Debug, PartialEq, Eq, PartialOrd, Ord, Hash)]
+enum CapturesInfo {
```
arora-aman, line 90: Is this only used for migration? If so I'd name is UpvarMigrationInfo for two reasons:
1. Upvar indicates we are taling about root variable, and migration helps with inteded purpose.
2. CaptureInfo very similarly named struct stores information about the information that we collect when run the analysis algorithm.

arora-aman (commented):
Thanks for looking into this. I only have nits.

compiler/rustc_typeck/src/check/upvar.rs
```diff
@@ -707,47 +744,64 @@ impl<'a, 'tcx> FnCtxt<'a, 'tcx> {
...
-                                    missing_trait,
-                                    captured_name,
-                                    missing_trait,
-                                ));
+                            for &missing_trait in &lint_note.reason.auto_traits {
+                                // not capturing something anymore cannot cause a trait to fail to be implemented:
+                                match &lint_note.captures_info {
+                                    CapturesInfo::CapturingLess { var_name: captured_name, .. } => {
+                                        diagnostics_builder.span_label(closure_head_span, format!("in Rust 2018, this closure implements {missing_trait} as `{x}` implements {missing_trait}, but in Rust 2021, this closure will no longer implement {missing_trait} as `{p}` does not implement {missing_trait}",
+                                            missing_trait = missing_trait,
```
mbartlett21, line 796:
```suggestion
```

compiler/rustc_typeck/src/check/upvar.rs
nikomatsakis: Yes, that was the error. The refactoring was just me figuring out how to actually emit a good error at the end. =)

compiler/rustc_typeck/src/check/upvar.rs
nikomatsakis: Yes. Good thought.

compiler/rustc_typeck/src/check/upvar.rs
```diff
@@ -86,18 +86,55 @@ impl<'a, 'tcx> FnCtxt<'a, 'tcx> {
...
+
+    /// When we used to capture `x` in its entirety, we would execute some destructors
+    /// at a different time.
+    drop_order: bool,
+}
+
+impl MigrationWarningReason {
+    fn migration_message(&self) -> String {
+        let base = "changes to closure capture in Rust 2021 will affect";
+        if !self.auto_traits.is_empty() && self.drop_order {
```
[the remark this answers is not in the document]
nikomatsakis: We do print them out, but not in the summary message. I didn't think it added much in the summary, but I don't have a strong opinion.

compiler/rustc_typeck/src/check/upvar.rs
nikomatsakis: Do you not need that anymore? Or not in rustc, anyway?

Side note that rustfmt was being very unhelpful for this code and I don't understand why.

mbartlett21: This crate has `#![feature(format_args_capture)]` in lib.rs, so it should work without the line.

wesleywiser (approved):
One small suggestion, feel free to take it or leave it 🙂 

r=me with `CapturesInfo` renamed per @arora-aman's suggestion

compiler/rustc_typeck/src/check/upvar.rs
nikomatsakis: Cool.

apiraino:
Beta backport approved as per compiler team [on Zulip](https://rust-lang.zulipchat.com/#narrow/stream/238009-t-compiler.2Fmeetings/topic/.5Bweekly.5D.202021-11-11.20.2354818/near/261145746)

@rustbot label +beta-accepted

Issue #85907, opened 2021-06-01
ICE: const_panic panics in some cases


m-ou-se:
```rust
#![feature(const_panic)]

const fn hey() -> usize {
    panic!(123);
}

fn main() {
    let _: [u8; hey()] = todo!();
}
```

Playground: https://play.rust-lang.org/?version=nightly&mode=debug&edition=2018&gist=31e3a36a34e378389943330936b81dbf

<details><summary>Output</summary>

```
   Compiling playground v0.0.1 (/playground)
error: argument to `panic!()` in a const context must have type `&str`
 --> src/main.rs:4:5
  |
4 |     panic!(123);
  |     ^^^^^^^^^^^^
  |
  = note: this error originates in the macro `$crate::panic::panic_2015` (in Nightly builds, run with -Z macro-backtrace for more info)

thread 'rustc' panicked at '`ref_to_mplace` called on non-ptr type', compiler/rustc_mir/src/interpret/place.rs:307:47
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace

error: internal compiler error: unexpected panic

note: the compiler unexpectedly panicked. this is a bug.

note: we would appreciate a bug report: https://github.com/rust-lang/rust/issues/new?labels=C-bug%2C+I-ICE%2C+T-compiler&template=ice.md

note: rustc 1.54.0-nightly (657bc0188 2021-05-31) running on x86_64-unknown-linux-gnu

note: compiler flags: -C embed-bitcode=no -C codegen-units=1 -C debuginfo=2 --crate-type bin

note: some of the compiler flags provided by cargo are hidden

query stack during panic:
#0 [eval_to_allocation_raw] const-evaluating + checking `main::{constant#0}`
#1 [eval_to_const_value_raw] simplifying constant for the type system `main::{constant#0}`
end of query stack
error: aborting due to previous error

error: could not compile `playground`

To learn more, run the command again with --verbose.
```
</details>

cc @RalfJung

RalfJung:
Ah, yes, array lengths... this is like https://github.com/rust-lang/rust/issues/79047

RalfJung:
Cc @rust-lang/wg-const-eval

RalfJung:
This ICEs on stable:
```rust
const fn hey() -> usize {
    panic!(123);
}

fn main() {
    let _: [u8; hey()] = todo!();
}
```
[...]

Issue #79047, opened 2020-11-14
Constant gets evaluated even when it contains invalid `transmute`


RalfJung:
Consider the following code:
```rust
const DST: &[u8] = unsafe { std::mem::transmute(1usize) };

fn main() {
    match &b""[..] {
        DST => {}
    }
}
```
This code hits a peculiar code path in the interpreter:

https://github.com/rust-lang/rust/blob/30e49a9ead550551e879af64ba91a0316da1c422/compiler/rustc_mir/src/interpret/place.rs#L916-L927

This code path *should* be unreachable: just like the interpreter can assume that code code it runs on is well-typed, it should be able to assume that the code it runs on has its transmutes checked. But something seems to be different about `transmute` when compared with "normal" type-checking.

`TransmuteSizeDiff` is a hack; we should instead arrange things in a way that failing the transmute check inhibits const-evaluation the same way that `true + 4` inhibitis const-evaluation.

RalfJung:
The same issue also arises with `SizeOf` and its requirement that the type be sized.
Cc https://github.com/rust-lang/rust/pull/81185#issuecomment-765217308

osa1:
> The same issue also arises with SizeOf and its requirement that the type be sized.

Is it really the same issue? My understanding is the `SizeOf` issue is a type error (as the argument's `Sized` bound is not satisfied), but the transmute error is caught not by the type checker, but rather it has its own check in

https://github.com/rust-lang/rust/blob/85e355ea9bd86ac6580a5d422a65dbf689845808/compiler/rustc_passes/src/intrinsicck.rs#L71-L124

So I think it's possible that we have two different bugs here.

RalfJung:
Well, maybe not exactly the same issue, but something very similar at least.

PR #90602, opened 2021-11-05
Unstably constify `impl<I: Iterator> IntoIterator for I`

mbartlett21:
This constifies the default `IntoIterator` implementation under the `const_intoiterator_identity` feature.

Tracking Issue: #90603

### DIFF
--- a/src/test/ui/consts/const-fn-error.stderr
+++ b/src/test/ui/consts/const-fn-error.stderr
@@ -22,8 +22,8 @@ LL |     for i in 0..x {
 note: impl defined here, but it is not `const`
   --> $SRC_DIR/core/src/iter/traits/collect.rs:LL:COL
    |
-LL | impl<I: Iterator> IntoIterator for I {
-   | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+LL | impl<I: ~const Iterator> const IntoIterator for I {
+   | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    = note: calls in constant functions are limited to constant functions, tuple structs and tuple variants
 
 error[E0658]: mutable references are not allowed in constant functions
--- a/library/core/src/iter/traits/collect.rs
+++ b/library/core/src/iter/traits/collect.rs
@@ -267,8 +267,9 @@ pub trait IntoIterator {
     fn into_iter(self) -> Self::IntoIter;
 }
 
+#[rustc_const_unstable(feature = "const_intoiterator_identity", issue = "90603")]
 #[stable(feature = "rust1", since = "1.0.0")]
-impl<I: Iterator> IntoIterator for I {
+impl<I: ~const Iterator> const IntoIterator for I {
     type Item = I::Item;
     type IntoIter = I;
 
--- a/src/test/ui/consts/const-for.stderr
+++ b/src/test/ui/consts/const-for.stderr
@@ -7,8 +7,8 @@ LL |     for _ in 0..5 {}
 note: impl defined here, but it is not `const`
   --> $SRC_DIR/core/src/iter/traits/collect.rs:LL:COL
    |
-LL | impl<I: Iterator> IntoIterator for I {
-   | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+LL | impl<I: ~const Iterator> const IntoIterator for I {
+   | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    = note: calls in constants are limited to constant functions, tuple structs and tuple variants
 
 error[E0015]: cannot call non-const fn `<std::ops::Range<i32> as Iterator>::next` in constants
--- a/src/test/ui/never_type/issue-52443.stderr
+++ b/src/test/ui/never_type/issue-52443.stderr
@@ -47,8 +47,8 @@ LL |     [(); { for _ in 0usize.. {}; 0}];
 note: impl defined here, but it is not `const`
   --> $SRC_DIR/core/src/iter/traits/collect.rs:LL:COL
    |
-LL | impl<I: Iterator> IntoIterator for I {
-   | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+LL | impl<I: ~const Iterator> const IntoIterator for I {
+   | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    = note: calls in constants are limited to constant functions, tuple structs and tuple variants
 
 error[E0658]: mutable references are not allowed in constants


mbartlett21:
Do I need to do a test to make sure it is unstable?
Something like:
```rust
const fn check() {
    let a = Some(5).into_iter();
}
```

src/test/ui/consts/const-fn-error.stderr
```diff
@@ -40,7 +40,7 @@ error[E0080]: evaluation of constant value failed
 LL |     for i in 0..x {
    |              ^^^^
    |              |
-   |              calling non-const function `<std::ops::Range<usize> as IntoIterator>::into_iter`
+   |              calling non-const function `iter::range::<impl Iterator for std::ops::Range<usize>>::next`
```
mbartlett21, line 43: It seems that if a non-const function is called, unstably const functions don't get a mention. Is that how it is meant to be?

fee1-dead: I don't know how that test works. Somehow despite all the errors we get from checking the MIR for feature gates and non-const functions, it still gets fed into the CTFE, which emits this error.

Perhaps someone else from @rust-lang/wg-const-eval can tell what's going on?

I wouldn't call this allowing calling unstablly const fns in stable code, since there are already errors.


Once we allow `Iterator` to have `const` `impl`s, we would add a test to see if for loops are gated, right now I don't think any actions will be necessary.

RalfJung: > Somehow despite all the errors we get from checking the MIR for feature gates and non-const functions, it still gets fed into the CTFE, which emits this error.

This might be something similar to https://github.com/rust-lang/rust/issues/79047, https://github.com/rust-lang/rust/issues/85907.
Some errors don't abort compilation (so that we can show more than 1 error) and we don't have the proper logic in place to skip CTFE of things that are already 'tainted' by errors.

fee1-dead:
We need a decision for `IntoIterator`: Should we make `IntoIter: ~const Iterator`?

If we do, then this impl would have stricter bounds on const impls. This should be decided before stabilizing `const_trait_impl`.

mbartlett21:
> We need a decision for `IntoIterator`: Should we make `IntoIter: ~const Iterator`?

Assuming you are meaning on the trait definition:
That seems like an improvement, since it would shorten bounds where `impl IntoIterator` is taken:
`where T: ~const IntoIterator<IntoIter: ~const Iterator, Item = u32>`
versus
`where T: ~const IntoIterator<Item = u32>`

scottmcm:
It sounds like there are questions about what the bounds should be here, and I'm not really familiar with impl constification (though the change appears fine to me), so I'm going to try sending this to someone from api for comment.

r? rust-lang/libs-api

JohnCSimon:
Triage: still looks like this is waiting on someone from libs-api to comment.

yaahc:
> We need a decision for `IntoIterator`: Should we make `IntoIter: ~const Iterator`?
> 
> If we do, then this impl would have stricter bounds on const impls. This should be decided before stabilizing `const_trait_impl`.

I'm not familiar enough with the effects of the conditional const bounds on trait impls as suggested. Can you elaborate some more on what change you're considering and how the bounds it introduces would be stricter?

Also, as a side note as of https://github.com/rust-lang/rust/pull/92433#pullrequestreview-853627952 we're "tapping the brakes a little on filling the standard library with ~const and allow rustdoc to catch up a little" so if the second suggested change to `IntoIterator` would be visible in rustdoc we'd likely want to wait for the same changes that 92433 is waiting on. Though I'm assuming the changes in the diff currently don't affect the generated rustdoc output, in which case we should be able to merge those separately (assuming that even makes sense here).

mbartlett21:
@yaahc 

Here are the two different approaches:

# Approach 1 - Bounds on the use:

```rust
// in core::iter

impl<I: Iterator> const IntoIterator for I { ... }
```

Use in a `const fn`:
```rust
const fn convert_and_iterate<I: ~const IntoIterator<IntoIter = J, Item = i32>, J: ~const Iterator<Item = i32>>(iter: I) -> i32 {
    iter.into_iter().next()
}
```

RFC 2289 would shorten it a bit:
```rust
#![feature(associated_type_bounds)]

const fn convert_and_iterate<I: ~const IntoIterator<IntoIter: ~const Iterator, Item = i32>>(iter: I) -> i32 {
    iter.into_iter().next()
}
```

# Approach 2 - Bounds on the impl:

```rust
// in core::iter

impl<I: ~const Iterator> const IntoIterator for I { ... }
```

Use in a `const fn`:
```rust
const fn convert_and_iterate<I: ~const IntoIterator<Item = i32>>(iter: I) -> i32 {
    iter.into_iter().next()
}
```

# Summary

The first approach means that `const IntoIterator` will apply to more types, but unfortunately also means more verbose type bounds. The second approach shortens the bounds, but means that it doesn't apply to as many types.

I would prefer the second approach, because of the usability aspect of it.

mbartlett21:
I might wait for this until #93412 is done.

yaahc:
I think this comes down to two questions.

1. What use cases are there for types like `NonConstIter` implementing `const IntoIterator`?
2. Would we be able to loosen the `~const Iterator` bound on `IntoIterator` in the future if we started with approach 2?

I think the answer to 2 is no because it would break code that previously relied on `~const Iterator` implying `~const IntoIterator`. Though I think this may imply that the opposite is true, that if we start with approach 1 we could always go back later and add the `~const Iterator` bound in the future to `IntoIterator` if we find that there are no valid usecases from the first question.

My instinct is that we'd want to pick option 1, and I wonder if there might be some opportunity to introduce language sugar here to simplify expressing repetative `~const` bounds. I expect either way resolving this question will likely be setting a precident for a lot of future const library APIs, so I think we're going to want to discuss this in more depth with the rest of the libs-api team and probably the lang team as well. cc @rust-lang/lang @rust-lang/libs-api

fee1-dead:
If `~const IntoIterator` and `I: ~const Iterator` were to be made orthogonal, it would require some changes in rustdoc to temporarily work around https://github.com/rust-lang/rust/pull/92433#pullrequestreview-865702419. And once we stabilize `~const`, we would have to display it in rustdoc which means the temporary workaround would have to be removed and these elaborated bounds will have to be displayed.

I do think that someone could have the need to convert something to an iterator but not to use it. But if CTFE just becomes more powerful each day I don't think there will be many cases where it cannot be iterated over in compile time, and those special cases could exist as inherent methods instead.

mbartlett21:
@yaahc @fee1-dead 
I've updated my comment above to clarify it a bit better (with the trait definition, not the identity impl):
https://github.com/rust-lang/rust/pull/90602#issuecomment-1020912885

The current PR doesn't change the trait definition. It only changes the identity impl using `~const` (which is compatible with both definitions for the trait, and can be relaxed later):
```diff
- impl<I: Iterator> IntoIterator for I {
+ impl<I: ~const Iterator> const IntoIterator for I {
```

yaahc:
Okay so, let me see if I understand this correctly. This PR only changes IntoIters identity impl to make it possible to use into iter in const for iterators which are usable in const, but it doesn't include the additional change to the trait definition required to shorten the usage in bounds. Adding this change by itself could be loosened to make all Iterator types const IntoIterator regardless of the iterators constness, but that would become breaking if we also added the const Iterator bound to the trait definition itself because then people could begin relying on one implying the other, which is also responsible for the shortened bounds.

yaahc:
> If `~const IntoIterator` and `I: ~const Iterator` were to be made orthogonal, it would require some changes in rustdoc to temporarily work around [#92433 (review)](https://github.com/rust-lang/rust/pull/92433#pullrequestreview-865702419). And once we stabilize `~const`, we would have to display it in rustdoc which means the temporary workaround would have to be removed and these elaborated bounds will have to be displayed.

Do we currently plan on stabilizing ~const without removing the workaround from 92433?

fee1-dead:
If we stabilize `~const`, then it will be part of the language. Rustdoc should document those bounds accurately.

yaahc:
cc @oli-obk

oli-obk:
Status update:

1. rustdoc hides the `~const` part of  `~const Trait` bounds and only displays them as `Trait`, so there is no problem anymore with this polluting stable docs
2. the original version of the PR was more verbose, but the current change is very simple. It changes the blanket `impl<I: Iterator> IntoIterator for I {` to a const impl while also binding `I` to `~const Iterator`.
    * the effects of this are that you can only use `into_iter` on iterators if the iterator has a `const` impl.
    * this is overly conservative, so we can relax it in the future, even if we end up deciding to stabilize it
3. The current @rust-lang/wg-const-eval stance is to only add trivial uses of `~const Trait`
    * so no new `where` bounds that weren't there before
    * no `~const Trait` bounds on associated type projections
    * this PR satisfies the "trival" criteria


So, imo we should merge this now, and in the future push for the trivial-use-stance more strongly in order to not put more burden on libs-api. We will revisit this stance once the const eval story is better fleshed out.

mbartlett21:
Has @bors forgotten this?

faptc:
Closing and reopening may help bors update this PR's state.

Issue #77174, opened 2020-09-25
Tracking Issue for inherent_ascii_escape


clarfonthey:
Feature gate: `#![feature(inherent_ascii_escape)]`

### Public API

```rust
impl u8 {
    pub fn escape_ascii(self) -> core::ascii::EscapeDefault;
}

impl [u8] {
    pub fn escape_ascii(&self) -> EscapeAscii<'_>;
}

// core::slice
pub struct EscapeAscii<'a>;
```

### Steps / History

- [x] Implementation: ~~#73111~~ #83130
- [x] Final commenting period (FCP)
- [ ] Stabilization PR

### Unresolved Questions

- [ ] Naming. `escape_ascii`, `escape_default`, `ascii_escape_default`, ... ?
- [ ] Deprecate `std::ascii::escape_default`? (That is, all of `std::ascii`.)

lzutao:
typo: `inherent_ascii_escpe`

tanriol:
What's the status of this? Is it ready for stabilization or waiting for something?

clarfonthey:
One thing that wasn't mentioned in FCP: should stabilising this also lead to deprecating the `std::ascii` module? Since the naming seems to be fine re: the first unresolved question, but there's the one about deprecation as well.

QuineDot:
Happy to see this land.

The documentation should
* link to [`escape_default`](https://doc.rust-lang.org/stable/std/ascii/fn.escape_default.html) or include the list of escapes inline
* mention that the iterator is over `u8`
* mention that it implements `Display` for those wishing to display escaped bytes without allocating (who aren't aware of the connection with `ToString`)

clarfonthey:
Adding another comment to this, since there hasn't been a stabilisation PR: should stabilising this also deprecate the `std::ascii` module, since this is basically the last thing from that module that hasn't been moved over to an inherent method? I personally vote yes, but I understand if others disagree. Would rather not close this issue before an official decision is made, but that's also not my decision, so :p

yaahc:
Nominated so we can make sure this gets discussed during our next libs-api team meeting on Wednesday.

clarfonthey:
Also adding more details on what the deprecation would look like, for full clarity. When I say `std` here, that includes `core` as well.

1. Deprecate `std::ascii` module, `std::ascii::EscapeDefault` struct, and `std::ascii::escape_default` function.
[...]

PR #90604, opened 2021-11-05
Re-export some iterators from `core` in `std`

mbartlett21:
These iterators seem to have been forgotten to be re-exported from `std` (through `alloc`)

These are stable:

`core::slice::{SplitInclusive, SplitInclusiveMut}`

This one is still unstable:

`core::slice::EscapeAscii` (cc #77174)

### DIFF
--- a/library/alloc/src/str.rs
+++ b/library/alloc/src/str.rs
@@ -46,7 +46,7 @@ pub use core::str::pattern;
 pub use core::str::EncodeUtf16;
 #[stable(feature = "split_ascii_whitespace", since = "1.34.0")]
 pub use core::str::SplitAsciiWhitespace;
-#[stable(feature = "split_inclusive", since = "1.53.0")]
+#[stable(feature = "split_inclusive", since = "1.51.0")]
 pub use core::str::SplitInclusive;
 #[stable(feature = "rust1", since = "1.0.0")]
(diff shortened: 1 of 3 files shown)


kennytm:
```
error[E0658]: use of unstable library feature 'inherent_ascii_escape'
   --> library/alloc/src/slice.rs:112:9
    |
112 | pub use core::slice::EscapeAscii;
    |         ^^^^^^^^^^^^^^^^^^^^^^^^
    |
    = note: see issue #77174 <https://github.com/rust-lang/rust/issues/77174> for more information
    = help: add `#![feature(inherent_ascii_escape)]` to the crate attributes to enable
```

library/alloc/src/str.rs
```diff
@@ -46,7 +46,7 @@ pub use core::str::pattern;
 pub use core::str::EncodeUtf16;
 #[stable(feature = "split_ascii_whitespace", since = "1.34.0")]
 pub use core::str::SplitAsciiWhitespace;
-#[stable(feature = "split_inclusive", since = "1.53.0")]
+#[stable(feature = "split_inclusive", since = "1.51.0")]
```
mbartlett21, line 49: This was actually incorrect. It hadn't been picked up, since it was across crates.

PR #90607, opened 2021-11-05
Make slice->str conversion and related functions `const`

WaffleLapkin:
This PR marks the following APIs as `const`:
```rust
// core::str
pub const fn from_utf8(v: &[u8]) -> Result<&str, Utf8Error>;
pub const fn from_utf8_mut(v: &mut [u8]) -> Result<&mut str, Utf8Error>;
pub const unsafe fn from_utf8_unchecked_mut(v: &mut [u8]) -> &mut str;

impl Utf8Error {
    pub const fn valid_up_to(&self) -> usize;
    pub const fn error_len(&self) -> Option<usize>;
}
```

Everything but `from_utf8_unchecked_mut` uses `const_str_from_utf8` feature gate, `from_utf8_unchecked_mut` uses `const_str_from_utf8_unchecked_mut` feature gate.

---

I'm not sure why `from_utf8_unchecked_mut` was left out being  non-`const`, considering that `from_utf8_unchecked` is not only `const`, but **`const` stable**.

---

r? @oli-obk (performance-only `const_eval_select` use)

### DIFF
--- a/library/core/src/str/converts.rs
+++ b/library/core/src/str/converts.rs
@@ -82,10 +82,16 @@ use super::Utf8Error;
 /// assert_eq!("💖", sparkle_heart);
 /// ```
 #[stable(feature = "rust1", since = "1.0.0")]
-pub fn from_utf8(v: &[u8]) -> Result<&str, Utf8Error> {
-    run_utf8_validation(v)?;
-    // SAFETY: Just ran validation.
-    Ok(unsafe { from_utf8_unchecked(v) })
+#[rustc_const_unstable(feature = "const_str_from_utf8", issue = "none")]
+pub const fn from_utf8(v: &[u8]) -> Result<&str, Utf8Error> {
+    // This should use `?` again, once it's `const`
+    match run_utf8_validation(v) {
+        Ok(_) => {
+            // SAFETY: validation succeeded.
+            Ok(unsafe { from_utf8_unchecked(v) })
+        }
+        Err(err) => Err(err),
+    }
 }
 
 /// Converts a mutable slice of bytes to a mutable string slice.
@@ -119,10 +125,16 @@ pub fn from_utf8(v: &[u8]) -> Result<&str, Utf8Error> {
 /// See the docs for [`Utf8Error`] for more details on the kinds of
 /// errors that can be returned.
 #[stable(feature = "str_mut_extras", since = "1.20.0")]
-pub fn from_utf8_mut(v: &mut [u8]) -> Result<&mut str, Utf8Error> {
-    run_utf8_validation(v)?;
-    // SAFETY: Just ran validation.
-    Ok(unsafe { from_utf8_unchecked_mut(v) })
+#[rustc_const_unstable(feature = "const_str_from_utf8", issue = "none")]
+pub const fn from_utf8_mut(v: &mut [u8]) -> Result<&mut str, Utf8Error> {
+    // This should use `?` again, once it's `const`
+    match run_utf8_validation(v) {
+        Ok(_) => {
+            // SAFETY: validation succeeded.
+            Ok(unsafe { from_utf8_unchecked_mut(v) })
+        }
+        Err(err) => Err(err),
+    }
 }
 
 /// Converts a slice of bytes to a string slice without checking
@@ -184,7 +196,8 @@ pub const unsafe fn from_utf8_unchecked(v: &[u8]) -> &str {
 #[inline]
 #[must_use]
 #[stable(feature = "str_mut_extras", since = "1.20.0")]
-pub unsafe fn from_utf8_unchecked_mut(v: &mut [u8]) -> &mut str {
+#[rustc_const_unstable(feature = "const_str_from_utf8_unchecked_mut", issue = "none")]
+pub const unsafe fn from_utf8_unchecked_mut(v: &mut [u8]) -> &mut str {
     // SAFETY: the caller must guarantee that the bytes `v`
     // are valid UTF-8, thus the cast to `*mut str` is safe.
     // Also, the pointer dereference is safe because that pointer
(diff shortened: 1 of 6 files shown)


library/core/src/str/validations.rs
```diff
@@ -105,21 +106,36 @@ const NONASCII_MASK: usize = 0x80808080_80808080u64 as usize;
...
+    const fn ctfe_align(_: &[u8], _: usize) -> usize {
+        usize::MAX
+    }
+
+    // SAFETY:
+    // It is permisseble for `align_offset` to always return `usize::MAX`,
+    // algorithm correctness can not depend on `align_offset` returning non-max values.
+    //
+    // As such the behaviour can't change after replacing `align_offset` with `usize::MAX`, only performance can.
+    let align = unsafe { const_eval_select((v, usize_bytes), ctfe_align, rt_align) };
```
Mark-Simulacrum, line 138: It seems better to put this inside align_offset itself, since the justification should work for all calls to it, right?

WaffleLapkin: While I think it would be great to put this inside `align_offset` itself, wouldn't this allow for user code to behave differently in `const` and non-`const` context?
```rust
const fn is_likely_called_in_ctfe() -> bool {
    let arr = [0u8; 2];
    (&arr[0] as *const u8).align_offset(2) == usize::MAX 
        && (&arr[1] as *const u8).align_offset(2) == usize::MAX
}
```
([playground](https://play.rust-lang.org/?version=nightly&mode=debug&edition=2021&gist=1b4b921b3efb38003e3d740accb8d48a))

WaffleLapkin: Actually even `(&0u8 as *const u8).align_offset(1) == usize::MAX` will do. Basically while `const_eval_select` in the caller of  `align_offset` shouldn't change behaviour, `const_eval_select` in `align_offset` itself clearly changes it's return value depending on whatever it's called at runtime or compile time.

Sadly, it seems that we don't want to allow this, right? :(

library/core/src/str/converts.rs
```diff
@@ -82,10 +82,15 @@ use super::Utf8Error;
 /// assert_eq!("💖", sparkle_heart);
 /// ```
 #[stable(feature = "rust1", since = "1.0.0")]
-pub fn from_utf8(v: &[u8]) -> Result<&str, Utf8Error> {
-    run_utf8_validation(v)?;
-    // SAFETY: Just ran validation.
-    Ok(unsafe { from_utf8_unchecked(v) })
+#[rustc_const_unstable(feature = "const_str_from_utf8", issue = "none")]
+pub const fn from_utf8(v: &[u8]) -> Result<&str, Utf8Error> {
+    match run_utf8_validation(v) {
```
oli-obk, line 87: leave a comment that we should move back to `?` once that is const

library/core/src/str/error.rs
```diff
@@ -94,10 +95,14 @@ impl Utf8Error {
     ///
     /// [U+FFFD]: ../../std/char/constant.REPLACEMENT_CHARACTER.html
     #[stable(feature = "utf8_error_error_len", since = "1.20.0")]
+    #[rustc_const_unstable(feature = "const_str_from_utf8", issue = "none")]
     #[must_use]
     #[inline]
-    pub fn error_len(&self) -> Option<usize> {
-        self.error_len.map(|len| len as usize)
+    pub const fn error_len(&self) -> Option<usize> {
+        match self.error_len {
```
oli-obk, line 102: leave a comment that this should become `map` again when that is const evaluable

library/core/src/str/validations.rs
```diff
@@ -229,31 +245,31 @@ pub(super) fn run_utf8_validation(v: &[u8]) -> Result<(), Utf8Error> {
     Ok(())
 }
 
-// https://tools.ietf.org/html/rfc3629
```
oli-obk, line 232: this change seems unnecessary, please undo it.

library/core/src/str/validations.rs
oli-obk: > `const_eval_select` in `align_offset` itself clearly changes it's return value depending on whatever it's called at runtime or compile time.
> 
> Sadly, it seems that we don't want to allow this, right? :(

Actually, [the docs](https://doc.rust-lang.org/std/primitive.pointer.html#method.align_offset) specifically state

> If it is not possible to align the pointer, the implementation returns usize::MAX. It is permissible for the implementation to always return usize::MAX. Only your algorithm’s performance can depend on getting a usable offset here, not its correctness.
> There are no guarantees whatsoever that offsetting the pointer will not overflow or go beyond the allocation that the pointer points into. It is up to the caller to ensure that the returned offset is correct in all terms other than alignment.

The only reason `align_offset` even exists is because I wanted to enable miri to do such trickery back when it wasn't able to do ptr2int conversions.

So, we can definitely make `align_offset` behave this way in const fn, as long as we have a tracking issue for the const stabilization that specifically raises this concern before we bake it into a stabilized function.

WaffleLapkin: I can make a follow-up PR then

oli-obk: I have a slight preference for doing a PR for just `align_offset` first, since then this PR does less work, but if you prefer to first merge this one I'm ok with that.

library/core/src/str/validations.rs
WaffleLapkin: I can undo moving this item into the function and move it back here, but `static`->`const` change is required since `const fn` can't refer to statics...

library/core/src/str/validations.rs
WaffleLapkin: I can do either way if you prefer `align_offset` first, then so be it :)

WaffleLapkin:
@oli-obk I've rebased to use #90958 & applied suggestions from your review

WaffleLapkin:
wait, shouldn't we create a tracking issue first?

oli-obk:
Oops thanks!

@bors r+ rollup

PR #90583, opened 2021-11-04
Fix ICE when rustdoc is scraping examples inside of a proc macro

willcrichton:
This PR provides a clearer semantics for how --scrape-examples interacts with macros. If an expression's span AND it's enclosing item's span both are not `from_expansion`, then the example will be scraped. The added test case `rustdoc-scrape-examples-macros` shows a variety of situations.

* A macro-rules macro that takes a function call as input: good
* A macro-rules macro that generates a function call as output: bad
* A proc-macro that generates a function call as output: bad
* An attribute macro that generates a function call as output: bad
* An attribute macro that takes a function call as input: good, if the proc macro is designed to propagate the input spans

I ran this updated rustdoc on pyo3 and confirmed that it successfully scrapes examples from inside a proc macro, eg

Screen Shot 2021-11-04 at 1 11 28 PM

(cc @mejrs) 

Additionally, this PR fixes an ordering bug in the highlighting logic.

Fixes https://github.com/rust-lang/rust/issues/90567.

r? @jyn514

PR #90611, opened 2021-11-05
Fix another ICE in rustdoc scrape_examples

fee1-dead:
This has occurred to me when documenting a crate with the arguments. Not sure what could have caused it.

r? @willcrichton


jyn514:
@fee1-dead can you please add a test for this?

willcrichton:
Thanks @fee1-dead !!

One request. Can you add a `trace!` call for these two early returns? Just like I'm introducing in this other PR: https://github.com/rust-lang/rust/pull/90583/files#diff-77f9ef3a3f7e3637167a2856739a2fbb33946a33c1bccb328119dfaaa801a02dR187

fee1-dead:
I found the culprit, which is the `percent-encoding` crate. I don't know the root cause of the ICE, but I'm just going to use the file as a ui test.

The file is available to download from [here](https://raw.githubusercontent.com/servo/rust-url/master/percent_encoding/src/lib.rs)

camelid:
Please try to make the test case an MCVE too. I don't think we should have a 500 line test case if we can avoid it.

jyn514:
@fee1-dead if you like, I know that https://github.com/chengniansun/perses is a tool for trying to reduce test cases, it may or may not be easier than trying to reduce this manually.

fee1-dead:
Thanks @jyn514! Perses is a very powerful tool and it reduced it to `use core;` which shocked me.

jyn514:
Hmm, that doesn't actually give me a lot of confidence that we know how this crash happens ... but I guess it's not super important as long as it's fixed and we have a test it won't regress.

@bors r=jyn514,willcrichton

petrochenkov:
@fee1-dead 
The added test writes the `t.calls` file right to the root of the repository (at least if you run `x.py` from it).
```
$ git status
On branch master
Your branch is up to date with 'origin/master'.

Untracked files:
  (use "git add <file>..." to include in what will be committed)
        t.calls
```
All test outputs should be limited to the build directory.

src/test/rustdoc-ui/scrape-examples-ice.rs
```diff
@@ -0,0 +1,4 @@
+// compile-flags: -Z unstable-options --scrape-examples-output-path t.calls --scrape-examples-target-crate foobar
```
jyn514, line 1: I think you can just remove `output-path` here.

fee1-dead: Don't think so, it will error if either of the two options are unset. Perhaps it should point to build/tmp/?

jyn514: > it will error if either of the two options are unset.

well that's odd ... @willcrichton is there a reason you made that a hard error instead of giving it a default value?

anyway, in the meantime, try `{{build-base}}/t.calls`: https://rustc-dev-guide.rust-lang.org/tests/adding.html#input-normalization

jyn514:
> All test outputs should be limited to the build directory.

Maybe it would be useful to write a tidy check for this?

src/test/rustdoc-ui/scrape-examples-ice.rs
willcrichton: There isn't an obvious default value IMO. Eg doing `{crate_name}.calls` isn't desirable due to the many conflicting crate names in a workspace.

jyn514: We could do `{crate_name}-{crate_type}` maybe. I don't think it has to be *amazing*, just some reasonable default to match `--output-format` json.

jyn514: Any tool that cares about conflicts should be using `output-path` like you suggested.

PR #90613, opened 2021-11-05
Allow to run a specific rustdoc-js* test

GuillaumeGomez:
r? @Mark-Simulacrum

### DIFF
--- a/src/bootstrap/test.rs
+++ b/src/bootstrap/test.rs
@@ -763,7 +763,7 @@ impl Step for RustdocJSStd {
     const ONLY_HOSTS: bool = true;
 
     fn should_run(run: ShouldRun<'_>) -> ShouldRun<'_> {
-        run.path("src/test/rustdoc-js-std")
+        run.suite_path("src/test/rustdoc-js-std")
     }
 
     fn make_run(run: RunConfig<'_>) {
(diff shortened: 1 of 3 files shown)


src/bootstrap/test.rs
```diff
@@ -783,6 +783,13 @@ impl Step for RustdocJSStd {
                 .arg(builder.doc_out(self.target))
                 .arg("--test-folder")
                 .arg(builder.src.join("src/test/rustdoc-js-std"));
+            for path in &builder.paths {
+                if let Some(name) = path.file_name().and_then(|f| f.to_str()) {
+                    if name.ends_with(".js") {
```
Mark-Simulacrum, line 788: This is probably not quite right -- the typical filter can be found in this file, if you search for "Get test-args by striping suite path" -- the logic is somewhat more complicated. Extracting that into a helper function seems like a good idea to me.

GuillaumeGomez: Indeed! I'll extract the function then.

GuillaumeGomez: I can't re-use the same code because I need the full path. However, I think I 
can do something similar.

GuillaumeGomez: I can't re-use the same code because I need the full path. However, I think I can do something similar.

GuillaumeGomez:
@Mark-Simulacrum Thanks for the suggestion! Thanks to it, I was able to clean another test-suite argument as well.

src/bootstrap/test.rs
```diff
@@ -938,8 +947,11 @@ impl Step for RustdocGUI {
             .arg("--tests-folder")
             .arg(builder.build.src.join("src/test/rustdoc-gui"));
         for path in &builder.paths {
-            if let Some(name) = path.file_name().and_then(|f| f.to_str()) {
-                if name.ends_with(".goml") {
+            if let Some(p) = util::is_valid_test_suite_arg(path, "src/test/rustdoc-gui", builder) {
+                if !p.ends_with(".goml") {
+                    continue;
```
Mark-Simulacrum, line 952: Why this filter?

I think we should give an error message if someone passes a non-goml path inside src/test/rustdoc-gui. (And similar for the other cases where you added a similar filter).

GuillaumeGomez: Fine by me!

GuillaumeGomez:
Added a panic in case the file extension wasn't the expected one.

Mark-Simulacrum:
r=me with commits squashed

Thanks!

Issue #90620, opened 2021-11-05
Stabilise `aarch64_target_feature`


adamgemmell:
This issue constitutes the stabilisation report as per the [rustc dev guide](https://rustc-dev-guide.rust-lang.org/stabilization_guide.html#write-a-stabilization-report). Tracking issue: https://github.com/rust-lang/rust/issues/44839

This came about as a result of the discussion on stabilising the aarch64 runtime feature detection macro: https://github.com/rust-lang/rust/issues/86941

Stabilisation PR: https://github.com/rust-lang/rust/pull/90621

### Summary

Along with [`std::is_aarch64_feature_detected!`](https://github.com/rust-lang/rust/issues/86941), we'd like to stabilise `aarch64_target_feature`. This allows using the `target_feature` attribute on aarch64 platforms like such:

```
#[target_feature(enable = "neon")]
unsafe fn asimd_code() {
    ...
```

This allows for unconditional code generation - the above function will be compiled as if assumed to only run on hosts that support the `neon` feature.

Note that this differs from compile-time feature detection (via `cfg!`) and run-time feature detection (via the `std::is_aarch64_feature_detected!` macro).

### Test cases

There are a variety of x86_64 tests in place testing the behaviour of the `target_feature` attribute. Since this proposal does not change the behaviour of `target_feature`, only stabilises some of its inputs, I do not believe there's any pressing need to update these tests.

### Documentation

I've updated the Rust reference: https://github.com/rust-lang/reference/pull/1102

### Unresolved questions

It's possible in the future that `PAuth` will have reason to be split into two separate features (`paca` and `pacg` in Linux's `AT_HWCAP`. If this happens the plan would be to add two new `target_feature` options and, if necessary, keep PAuth as the combination of the two. It doesn't make sense to pre-emptively separate them before they're separate in LLVM, nor keep `PAuth` unstable for an undetermined amount of time.

PR #90621, opened 2021-11-05
Stabilise `aarch64_target_feature`

adamgemmell:
This PR stabilises `aarch64_target_feature` - see https://github.com/rust-lang/rust/issues/90620

### DIFF
--- a/compiler/rustc_codegen_ssa/src/target_features.rs
+++ b/compiler/rustc_codegen_ssa/src/target_features.rs
@@ -44,105 +44,108 @@ const ARM_ALLOWED_FEATURES: &[(&str, Option<Symbol>)] = &[
 
 const AARCH64_ALLOWED_FEATURES: &[(&str, Option<Symbol>)] = &[
     // FEAT_AdvSimd
-    ("neon", Some(sym::aarch64_target_feature)),
+    ("neon", None),
     // FEAT_FP
-    ("fp", Some(sym::aarch64_target_feature)),
+    ("fp", None),
     // FEAT_FP16
-    ("fp16", Some(sym::aarch64_target_feature)),
+    ("fp16", None),
     // FEAT_SVE
-    ("sve", Some(sym::aarch64_target_feature)),
+    ("sve", None),
     // FEAT_CRC
-    ("crc", Some(sym::aarch64_target_feature)),
+    ("crc", None),
     // FEAT_RAS
-    ("ras", Some(sym::aarch64_target_feature)),
+    ("ras", None),
     // FEAT_LSE
-    ("lse", Some(sym::aarch64_target_feature)),
+    ("lse", None),
     // FEAT_RDM
-    ("rdm", Some(sym::aarch64_target_feature)),
+    ("rdm", None),
     // FEAT_RCPC
-    ("rcpc", Some(sym::aarch64_target_feature)),
+    ("rcpc", None),
     // FEAT_RCPC2
-    ("rcpc2", Some(sym::aarch64_target_feature)),
+    ("rcpc2", None),
     // FEAT_DotProd
-    ("dotprod", Some(sym::aarch64_target_feature)),
+    ("dotprod", None),
     // FEAT_TME
-    ("tme", Some(sym::aarch64_target_feature)),
+    ("tme", None),
     // FEAT_FHM
-    ("fhm", Some(sym::aarch64_target_feature)),
+    ("fhm", None),
     // FEAT_DIT
-    ("dit", Some(sym::aarch64_target_feature)),
+    ("dit", None),
     // FEAT_FLAGM
-    ("flagm", Some(sym::aarch64_target_feature)),
+    ("flagm", None),
     // FEAT_SSBS
-    ("ssbs", Some(sym::aarch64_target_feature)),
+    ("ssbs", None),
     // FEAT_SB
-    ("sb", Some(sym::aarch64_target_feature)),
+    ("sb", None),
     // FEAT_PAUTH (address authentication)
-    ("paca", Some(sym::aarch64_target_feature)),
+    ("paca", None),
     // FEAT_PAUTH (generic authentication)
-    ("pacg", Some(sym::aarch64_target_feature)),
+    ("pacg", None),
     // FEAT_DPB
-    ("dpb", Some(sym::aarch64_target_feature)),
+    ("dpb", None),
     // FEAT_DPB2
-    ("dpb2", Some(sym::aarch64_target_feature)),
+    ("dpb2", None),
     // FEAT_SVE2
-    ("sve2", Some(sym::aarch64_target_feature)),
+    ("sve2", None),
     // FEAT_SVE2_AES
-    ("sve2-aes", Some(sym::aarch64_target_feature)),
+    ("sve2-aes", None),
     // FEAT_SVE2_SM4
-    ("sve2-sm4", Some(sym::aarch64_target_feature)),
+    ("sve2-sm4", None),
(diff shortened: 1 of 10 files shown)


michaelwoerister:
Thanks for the PR, @adamgemmell!

I'm assigning this to a compiler team lead since I don't really know if we have anyone owning this feature and I don't feel qualified to review myself. 

r? @wesleywiser 

cc @rust-lang/compiler-contributors in case someone has a special interest in moving this forward.

nagisa:
I would like a confirmation that the feature names have been double-checked to be the names we want to stabilize. Beyond `neon`, `fp*`, `crc` and some others the names are quite cryptic, so an affirmation that these names are consistent with e.g. Arm's documentation would be good.

adamgemmell:
More eyes are always welcome. My philosophy was to keep names close to the ARM `FEAT_*` names, with small deviations allowed if the LLVM names are close. There's [overrides](https://github.com/rust-lang/rust/blob/e4c23daeb461ac02413eb36c8cefcc5530638a05/compiler/rustc_codegen_llvm/src/llvm_util.rs#L153) if the two are different.

The [ARM ARM](https://developer.arm.com/documentation/ddi0487/latest) is the authority of most of the features with [this](https://developer.arm.com/architectures/cpu-architecture/a-profile/exploration-tools/feature-names-for-a-profile) page being a handy reference (though it currently mistakenly lists FEAT_DotProd2). [This file](https://github.com/rust-lang/llvm-project/blob/rustc/13.0-2021-09-30/llvm/lib/Target/AArch64/AArch64.td) lists the LLVM feature names.

adamgemmell:
@nagisa to clarify, I believe these names have been checked by at least 3 pairs of eyes.

r? @Amanieu

Amanieu:
I'm having second thoughts about the `v8.X` features, it's not clear what they represent and they don't work with the feature detection macros anyways. Could we leave those as unstable? Perhaps in the future we can remove them entirely.

Amanieu:
This and the associated documentation in the reference is probably going to miss the beta cutoff tomorrow, so you should update the version number to 1.61.0.

compiler/rustc_typeck/src/collect.rs
```diff
@@ -1,4 +1,3 @@
-// ignore-tidy-filelength
 //! "Collection" is the process of determining the type and other external
 //! details of each item in Rust. Collection is specifically concerned
 //! with *inter-procedural* things -- for example, for a function
```
matthiaskrgr, line 3: this causes CI to fail
https://github.com/rust-lang/rust/pull/94161#issuecomment-1046053573

adamgemmell: Something confusing happened https://github.com/rust-lang/rust/commit/7ca1c48bbb6d300bda665a59b7ad226bc71e7be9#diff-ac1e6e554bac799e1e2d3886bea8fdc0d0bf04b20d33df59f5a8e92c41987ac6

I've just rebased this so it should pass CI now - though the version number still needs updating if this isn't going into 1.60 beta as Amanieu said.

compiler/rustc_feature/src/accepted.rs
```diff
@@ -46,6 +46,8 @@ declare_features! (
     // feature-group-start: accepted features
     // -------------------------------------------------------------------------
 
+    /// Allows `#[target_feature(...)]` on aarch64 platforms
+    (accepted, aarch64_target_feature, "1.60.0", Some(44839), None),
```
marmeladema, line 50: Should it be `1.61.01` instead?

Amanieu:
Ah nice, catch.

@bors r-

compiler/rustc_codegen_ssa/src/target_features.rs
```diff
@@ -41,95 +41,95 @@ const ARM_ALLOWED_FEATURES: &[(&str, Option<Symbol>)] = &[
...
     // FEAT_VHE
-    ("vh", Some(sym::aarch64_target_feature)),
+    ("vh", None),
     // FEAT_PMUv3
-    ("pmuv3", Some(sym::aarch64_target_feature)),
+    ("pmuv3", None),
     // FEAT_SPE
-    ("spe", Some(sym::aarch64_target_feature)),
+    ("spe", None),
     ("v8.1a", Some(sym::aarch64_target_feature)),
```
nagisa, line 133: These probably should use a different target feature now that the `aarch64_target_feature` has been marked as “accepted”.

nagisa: More generally, please add a test which ensure these target features are still being gated.

adamgemmell:
@workingjubilee @Amanieu I brought the suggested changes from https://github.com/rust-lang/rust/pull/91608 into this PR as resolving it blocks stabilisation anyway. Any more thoughts on how this is done?

See also the changes in the documentation here: https://github.com/rust-lang/reference/pull/1102

hkratz:
This PR causes a compile-time error on every aarch64 function which just has the neon feature explicitly enabled  via `#[target_feature(enable = "neon")`, even if fp is enabled implicitly. This will cause a good amount of breakage across the ecosystem - [Github Code Search](https://cs.github.com/?q=%23%5Btarget_feature%28enable+%3D+%22neon%22%29%5D). After this change even running `TARGET=aarch64-apple-darwin cargo test` on `stdarch` fails with lots of compile-time errors.

(I found out because of a sudden CI failure for simdutf8 [here](https://github.com/rusticstuff/simdutf8/runs/5564403668?check_suite_focus=true).)

hkratz:
This PR causes a compile-time error on every aarch64 function which just has the neon feature explicitly enabled  via `#[target_feature(enable = "neon")`, even if `fp` is enabled implicitly. This will cause a good amount of breakage across the ecosystem - [Github Code Search](https://cs.github.com/?q=%23%5Btarget_feature%28enable+%3D+%22neon%22%29%5D). After this change even running `TARGET=aarch64-apple-darwin cargo test` on `stdarch` fails with lots of compile-time errors.

Apparently the stdarch build problems cause any aarch64 usage to fail in current nightly. Trying to compile this:
```rust
fn main() {
    unsafe {
        let _ = core::arch::aarch64::vmovq_n_u8(0);
    }
}
```
results in:
```
$ cargo +nightly build
   Compiling xx v0.1.0 (/Users/hans/dev/xx)
error: the target features fp, neon must all be either enabled or disabled together
    --> /Users/hans/.rustup/toolchains/nightly-aarch64-apple-darwin/lib/rustlib/src/rust/library/core/src/../../stdarch/crates/core_arch/src/arm_shared/neon/mod.rs:5191:1
     |
5191 | #[target_feature(enable = "neon")]
     | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
     |
     = help: add the missing features in a `target_feature` attribute

error: the target features fp, neon must all be either enabled or disabled together
    --> /Users/hans/.rustup/toolchains/nightly-aarch64-apple-darwin/lib/rustlib/src/rust/library/core/src/../../stdarch/crates/core_arch/src/arm_shared/neon/mod.rs:5721:1
     |
5721 | #[target_feature(enable = "neon")]
     | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
     |
     = help: add the missing features in a `target_feature` attribute

error: could not compile `xx` due to 2 previous errors
```


(I found out because of a sudden CI failure for simdutf8 [here](https://github.com/rusticstuff/simdutf8/runs/5564403668?check_suite_focus=true).)

workingjubilee:
My apologies, I didn't catch that the implementation of this did not canonicalize +neon to also mean +fp.

PR #90623, opened 2021-11-05
Remove more checks for LLVM < 12

cuviper:
We already updated the minimum to 12 in #90175, but we missed a few `get_version()` checks.

### DIFF
--- a/compiler/rustc_codegen_llvm/src/abi.rs
+++ b/compiler/rustc_codegen_llvm/src/abi.rs
@@ -1,7 +1,6 @@
 use crate::builder::Builder;
 use crate::context::CodegenCx;
 use crate::llvm::{self, AttributePlace};
-use crate::llvm_util;
 use crate::type_::Type;
 use crate::type_of::LayoutLlvmExt;
 use crate::value::Value;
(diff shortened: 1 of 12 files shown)


compiler/rustc_codegen_llvm/src/abi.rs
```diff
@@ -53,15 +52,10 @@ pub trait ArgAttributesExt {
...
-    cx.tcx
-        .sess
-        .opts
-        .debugging_opts
-        .mutable_noalias
-        .unwrap_or_else(|| llvm_util::get_version() >= (12, 0, 0))
+    // LLVM prior to version 12 had known miscompiles in the presence of
+    // noalias attributes (see #54878), but we don't support earlier
+    // versions at all anymore. We now enable mutable noalias by default.
+    cx.tcx.sess.opts.debugging_opts.mutable_noalias.unwrap_or(true)
```
nikic, line 58: You'll want to update the option description in compiler/rustc_session/src/options.rs as well.

cuviper: Done!

cuviper:
There are a few more comments around `NoAliasMutRef` -- let me get those too...

PR #90625, opened 2021-11-05
Add `UnwindSafe` to `Once`

### LINKED ISSUE #43469: std::sync::Once: Should Once be UnwindSafe or RefUnwindSafe?
Currently, the following program [will not compile](https://play.rust-lang.org/?gist=b2be5c6728418ded9d6a559c780779b7&version=stable) because `Once` isn't `RefUnwindSafe`. Is this the desired behavior, or should we consider making it `RefUnwindSafe` or `UnwindSafe`?

```rust
use std::sync::Once;
use std::panic::catch_unwind;

fn main() {
    let o = Once::new();
    catch_unwind(|| o.call_once(|| panic!("foo")));
}
```

Milo123459:
Fixes #43469

### DIFF
--- a/library/std/src/sync/once.rs
+++ b/library/std/src/sync/once.rs
@@ -90,6 +90,7 @@ mod tests;
 use crate::cell::Cell;
 use crate::fmt;
 use crate::marker;
+use crate::panic::UnwindSafe;
 use crate::sync::atomic::{AtomicBool, AtomicUsize, Ordering};
 use crate::thread::{self, Thread};
 
@@ -122,6 +123,9 @@ pub struct Once {
 unsafe impl Sync for Once {}
 #[stable(feature = "rust1", since = "1.0.0")]
 unsafe impl Send for Once {}
+
+#[stable(feature = "sync_once_ref_unwind_safe", since = "1.59.0")]
+impl UnwindSafe for Once {}
 
 /// State yielded to [`Once::call_once_force()`]’s closure parameter. The state
 /// can be used to query the poison status of the [`Once`].
--- a/library/std/src/sync/once.rs
+++ b/library/std/src/sync/once.rs
@@ -90,7 +90,7 @@ mod tests;
 use crate::cell::Cell;
 use crate::fmt;
 use crate::marker;
-use crate::panic::UnwindSafe;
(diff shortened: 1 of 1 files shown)


camelid:
r? rust-lang/libs

camelid:
r? rust-lang/libs-api

dtolnay (changes requested):
This needs a manual `RefUnwindSafe` impl as well.

dtolnay:
@rust-lang/libs-api:
@rfcbot fcp merge

All of the `Once` methods behave atomically and there is no possibility of the `Once` being left in a corrupted state following an unwind. `UnwindSafe` and `RefUnwindSafe` auto trait impls only didn't kick in because of the presence of a `Cell` inside `Waiter`.

library/std/src/sync/once.rs
```diff
@@ -123,6 +124,9 @@ unsafe impl Sync for Once {}
 #[stable(feature = "rust1", since = "1.0.0")]
 unsafe impl Send for Once {}
 
+#[stable(feature = "ref_unwind_safe", since = "1.56.1")]
```
camelid, line 127: (Version will need to be updated.)

camelid: Also, should this use a new feature—like `sync_once_ref_unwind_safe`—to avoid confusion?

library/std/src/sync/once.rs
```diff
@@ -124,7 +124,7 @@ unsafe impl Sync for Once {}
 #[stable(feature = "rust1", since = "1.0.0")]
 unsafe impl Send for Once {}
 
-#[stable(feature = "ref_unwind_safe", since = "1.56.1")]
+#[stable(feature = "sync_once_ref_unwind_safe", since = "1.57.0")]
```
camelid, line 127: This actually needs to be at least 1.59.0, since that is the current nightly version. It may be better to wait until this is ready to be merged to update the version though.

```suggestion
#[stable(feature = "sync_once_ref_unwind_safe", since = "1.59.0")]
```

dtolnay:
- Squashed, and added `RefUnwindSafe` as requested above (and implied by the branch name)

dtolnay (approved):

PR #90628, opened 2021-11-05
Clarify error messages caused by re-exporting `pub(crate)` visibility to outside

ken-matsui:
This PR clarifies error messages and suggestions caused by re-exporting pub(crate) visibility outside the crate.

Here is a small example ([Rust Playground](https://play.rust-lang.org/?version=nightly&mode=debug&edition=2018&gist=e2cd0bd4422d4f20e6522dcbad167d3b)):

```rust
mod m {
    pub(crate) enum E {}
}
pub use m::E;

fn main() {}
```

This code is compiled to:

```
error[E0365]: `E` is private, and cannot be re-exported
 --> prog.rs:4:9
  |
4 | pub use m::E;
  |         ^^^^ re-export of private `E`
  |
  = note: consider declaring type or module `E` with `pub`

For more information about this error, try `rustc --explain E0365`.
```

However, enum `E` is actually public to the crate, not private totally—nevertheless, rustc treats `pub(crate)` and private visibility as the same on the error messages. They are not clear and should be segmented distinctly.

By applying changes in this PR, the error message below will be the following message that would be clearer:

```
error[E0365]: `E` is only public within the crate, and cannot be re-exported outside
 --> prog.rs:4:9
  |
4 | pub use m::E;
  |         ^^^^ re-export of crate public `E`
  |
  = note: consider declaring type or module `E` with `pub`

For more information about this error, try `rustc --explain E0365`.
```

### DIFF
--- a/compiler/rustc_resolve/src/imports.rs
+++ b/compiler/rustc_resolve/src/imports.rs
@@ -1180,11 +1180,17 @@ impl<'a, 'b> ImportResolver<'a, 'b> {
 
         let mut reexport_error = None;
         let mut any_successful_reexport = false;
+        let mut crate_private_reexport = false;
         self.r.per_ns(|this, ns| {
             if let Ok(binding) = source_bindings[ns].get() {
                 let vis = import.vis.get();
                 if !binding.vis.is_at_least(vis, &*this) {
                     reexport_error = Some((ns, binding));
+                    if let ty::Visibility::Restricted(binding_def_id) = binding.vis {
+                        if binding_def_id.is_top_level_module() {
+                            crate_private_reexport = true;
+                        }
+                    }
                 } else {
                     any_successful_reexport = true;
                 }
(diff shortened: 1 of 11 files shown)


compiler/rustc_resolve/src/imports.rs
```diff
@@ -1207,24 +1213,34 @@ impl<'a, 'b> ImportResolver<'a, 'b> {
...
             } else {
-                let msg = format!("`{}` is private, and cannot be re-exported", ident);
-                let note_msg =
-                    format!("consider marking `{}` as `pub` in the imported module", ident,);
-                struct_span_err!(self.r.session, import.span, E0364, "{}", &msg)
-                    .span_note(import.span, &note_msg)
-                    .emit();
+                let error_msg = if crate_private_reexport {
+                    format!(
+                        "`{}` is only public to inside of the crate, and cannot be re-exported outside",
```
oli-obk, line 1219: just a wording nit: "is only public within the crate", then this lgtm

ken-matsui: @oli-obk 
Thank you for your review! I've updated it :)

PR #90635, opened 2021-11-05
rustdoc: clippy::complexity fixes

matthiaskrgr:


GuillaumeGomez:
Nice thanks! I don't think it'll impact perf but just in case...

@bors: r+ rollup=iffy

Issue #85071, opened 2021-05-08
Unused variable warning for used variable.


koehlma:
When trying to compile the following code I get an unused variable warning despite the variable is being used:

https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=66167b96f0e14d53616abbd0e6b3d12c

The current output is:
```
warning: unused variable: `typ`
  --> src/lib.rs:62:17
   |
62 |             let typ = T::infer_type(self);
   |                 ^^^ help: if this is intentional, prefix it with an underscore: `_typ`
   |
   = note: `#[warn(unused_variables)]` on by default
```

Ideally the output should not produce such a warning.

`cargo fix` produces the following output:

```
warning: failed to automatically apply fixes suggested by rustc to crate `rust_playground`

after fixes were automatically applied the compiler reported errors within these files:

  * src\lib.rs

This likely indicates a bug in either rustc or cargo itself,
and we would appreciate a bug report! You're likely to see
a number of compiler warnings after this message which cargo
attempted to fix but failed. If you could open an issue at
https://github.com/rust-lang/rust/issues
quoting the full output of this command we'd be very appreciative!
Note that you may be able to make some more progress in the near-term
fixing code with the `--broken-code` flag

The following errors were reported:
error[E0425]: cannot find value `typ` in this scope
  --> src\lib.rs:64:30
   |
64 |             info.typ.replace(typ);
   |                              ^^^ help: a local variable with a similar name exists: `_typ`

error: aborting due to previous error

For more information about this error, try `rustc --explain E0425`.
Original diagnostics will follow.

warning: unused variable: `typ`
  --> src\lib.rs:62:17
   |
62 |             let typ = T::infer_type(self);
   |                 ^^^ help: if this is intentional, prefix it with an underscore: `_typ`
   |
   = note: `#[warn(unused_variables)]` on by default

warning: 1 warning emitted

warning: unused variable: `typ`
  --> src\lib.rs:62:17
   |
62 |             let typ = T::infer_type(self);
   |                 ^^^ help: if this is intentional, prefix it with an underscore: `_typ`
   |
   = note: `#[warn(unused_variables)]` on by default

warning: 1 warning emitted

    Finished dev [unoptimized + debuginfo] target(s) in 1.04s
```

ehuss:
[...]

PR #90636, opened 2021-11-05
Fix false positives in `unreachable_code` lint

### LINKED ISSUE #89779: Unreachable code warning even when no code follows
#85556 introduced an unreachable code warning after uninhabited expressions; unfortunately that warning fires even when no code follows the expression.

This makes it impossible to use patterns such as:
```rust
fn foo() -> Result<std::convert::Infallible, u64> {
    Err(5)
}

fn main() { 
    match foo().expect("should") {}
}
```
where the `match` is being used in the manner suggested by @nikomatsakis in a [blog post](http://smallcultfollowing.com/babysteps/blog/2018/08/13/never-patterns-exhaustive-matching-and-uninhabited-types-oh-my/).

FabianWolff:
Fixes #89779 (both the false positives as well as the empty `match`) and fixes #88393.

r? @Mark-Simulacrum

### DIFF
--- a/compiler/rustc_passes/src/liveness.rs
+++ b/compiler/rustc_passes/src/liveness.rs
@@ -1300,25 +1300,37 @@ impl<'a, 'tcx> Liveness<'a, 'tcx> {
             // that we do not emit the same warning twice if the uninhabited type
             // is indeed `!`.
 
-            self.ir.tcx.struct_span_lint_hir(
-                lint::builtin::UNREACHABLE_CODE,
-                expr_id,
-                expr_span,
-                |lint| {
-                    let msg = format!("unreachable {}", descr);
-                    lint.build(&msg)
-                        .span_label(expr_span, &msg)
-                        .span_label(orig_span, "any code following this expression is unreachable")
-                        .span_note(
-                            orig_span,
-                            &format!(
-                                "this expression has type `{}`, which is uninhabited",
-                                orig_ty
-                            ),
-                        )
-                        .emit();
-                },
-            );
+            if let Some(hir::Node::Expr(hir::Expr {
+                kind: hir::ExprKind::Match(_, [], _), ..
+            })) = self.ir.tcx.hir().find(expr_id)
+            {
+                // The expression we are about to lint on is an empty match, which
+                // can be used as a kind of "uninhabitedness assertion" (issue
+                // #89779). We do not want to lint in this case.
+            } else {
+                self.ir.tcx.struct_span_lint_hir(
+                    lint::builtin::UNREACHABLE_CODE,
+                    expr_id,
+                    expr_span,
+                    |lint| {
+                        let msg = format!("unreachable {}", descr);
+                        lint.build(&msg)
+                            .span_label(expr_span, &msg)
+                            .span_label(
+                                orig_span,
+                                "any code following this expression is unreachable",
+                            )
+                            .span_note(
+                                orig_span,
+                                &format!(
+                                    "this expression has type `{}`, which is uninhabited",
+                                    orig_ty
+                                ),
+                            )
+                            .emit();
+                    },
+                );
+            }
         }
     }
 }
--- a/compiler/rustc_passes/src/liveness.rs
+++ b/compiler/rustc_passes/src/liveness.rs
@@ -515,6 +515,33 @@ struct Liveness<'a, 'tcx> {
     cont_ln: HirIdMap<LiveNode>,
 }
 
+// As we go through the code, we warn about unreachable code after expressions
+// with uninhabited types. We do not want to warn if the last expression of an
+// if or match arm has an uninhabited type, though, because the code after the
+// if/match could still be reachable via another arm -- unless all arms
+// diverge. The following two enums are used for tracking this information.
+
+#[derive(Clone, Copy, PartialEq, Eq)]
+enum WarnUnreachable {
+    Yes,
+    No,
+}
+
+#[derive(Clone, Copy, PartialEq, Eq)]
+enum Diverges {
+    Yes,
+    No,
+}
+impl std::ops::BitOr for Diverges {
+    type Output = Self;
+    fn bitor(self, rhs: Self) -> Self {
+        match (self, rhs) {
+            (Diverges::No, Diverges::No) => Diverges::No,
+            _ => Diverges::Yes,
+        }
+    }
+}
+
 impl<'a, 'tcx> Liveness<'a, 'tcx> {
     fn new(ir: &'a mut IrMaps<'tcx>, body_owner: LocalDefId) -> Liveness<'a, 'tcx> {
         let typeck_results = ir.tcx.typeck(body_owner);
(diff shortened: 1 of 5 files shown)


compiler/rustc_passes/src/liveness.rs
```diff
@@ -803,38 +827,67 @@ impl<'a, 'tcx> Liveness<'a, 'tcx> {
...
         }
     }
 
-    fn propagate_through_exprs(&mut self, exprs: &[Expr<'_>], succ: LiveNode) -> LiveNode {
-        exprs.iter().rev().fold(succ, |succ, expr| self.propagate_through_expr(&expr, succ))
+    fn propagate_through_exprs(
+        &mut self,
+        exprs: &[Expr<'_>],
+        succ: LiveNode,
+        warn_about_unreachable: bool,
```
Mark-Simulacrum, line 845: Can we factor out a `enum WarnUnreachable { Yes, No }` for this? That way it's clear at the call site what is getting passed around.

Mark-Simulacrum (commented):
Can you give a high level description of the changes in the second commit? I can probably figure it out based on the diff, but it would be good to document that in the commit message and PR description to make it easier to identify what's going on.

compiler/rustc_passes/src/liveness.rs
FabianWolff: Good idea; I've added two new enums `WarnUnreachable` and `Diverges`.

FabianWolff:
> Can you give a high level description of the changes in the second commit? I can probably figure it out based on the diff, but it would be good to document that in the commit message and PR description to make it easier to identify what's going on.

I've added a comment in the code explaining what is happening.

Honestly, at this point, I think it would be better to have a separate pass that does unreachable code warnings. The warnings in the type checker are incomplete (#85071), and adding them to the liveness analysis was a bad idea in hindsight because it just isn't a good fit (and avoiding duplicate warnings if the warning can be issued in two different places is also cumbersome).

This PR should solve #89779 for now, but we should look elsewhere for a good long-term solution.

tabokie:
This will also fix #90636.

FabianWolff:
> This will also fix #88393.

Right, thanks for pointing this out! I have added a test case for it.

@rustbot ready

compiler/rustc_passes/src/liveness.rs
```diff
@@ -515,6 +515,33 @@ struct Liveness<'a, 'tcx> {
     cont_ln: HirIdMap<LiveNode>,
 }
 
+// As we go through the code, we warn about unreachable code after expressions
+// with uninhabited types. We do not want to warn if the last expression of an
+// if or match arm has an uninhabited type, though, because the code after the
+// if/match could still be reachable via another arm -- unless all arms
```
Mark-Simulacrum, line 521: Hm, can you say more about the specific case being addressed here? It's not quite clear to me why diverges would be yes and warn unreachable would be no, yet we shouldn't lint on that expression?

FabianWolff:
```rust
enum Void {}
fn void() -> Void {todo!()}

fn bar(b: bool) {
    if b {
        void(); // A
    } else {
        // B
    }
    // C
}
```
In this example, `A` diverges because it has an uninhabited type. We can't issue a warning that "any code [in particular, `C`] following this expression is unreachable", though, because `C` might be reachable via the `else` branch. Hence `A` diverges and we don't want to warn about unreachable successors (`C`). If `B` also turns out to be diverging, we will warn about the entire `if` expression diverging and making `C` unreachable.

Mark-Simulacrum:
I left one question -- the code at a brief glance seems OK, but would like to review it more thoroughly with a better understanding of the intent behind the changes here (per the comment I left).

I tend to agree with you (https://github.com/rust-lang/rust/pull/90636#issuecomment-967695669) that we should be exploring other options. In particular, "dead code paths" seem like a pass that is well-suited for MIR, particularly since it may let us run const-prop and similar to detect more cases (though it's not clear whether that's actually desirable -- sometimes worse quality lints are actually better). On the other hand, MIR may be too low level to effectively map back. I think all the linting we do in the liveness pass should be possible to do on MIR, though -- we already have similar passes there, I think. If we feel that such a rewrite is feasible, I'd probably suggest pursuing that over patchwork fixes.

compiler/rustc_passes/src/liveness.rs
Mark-Simulacrum: Hm, so it seems like when we're exiting the if in the CFG, we should only consider C diverging if A *and* B have diverged -- so at point C we should have diverges = no, and as such not emit any warnings?

FabianWolff: Exactly.

Mark-Simulacrum: My comment was meant to suggest that I (still) don't understand the purpose/necessity for WarnUnreachable -- it seems to me that we should already not be warning when seeing C just based on Diverges = no at that point, so WarnUnreachable isn't actually necessary?

FabianWolff: Ah, sorry. Consider another example then:
```rust
if ... {
    diverges(); // X
    alsoDiverges(); // Y
}
// Z
```
Y has the successor Z and diverges, so Diverges=Yes, but WarnUnreachable=No because Z is also reachable via the (implicit) else branch. By contrast, X has the successor Y and diverges, so Diverges=Yes, and WarnUnreachable=Yes because X is Y's only predecessor. If there was a non-implicit else branch, the Diverges information would be used to check whether the entire `if` is diverging (in which case we would want to warn about Z being unreachable).

So you are right in that WarnUnreachable doesn't make a difference if Diverges=No, but it does for Diverges=Yes.

Mark-Simulacrum: I would expect the implicit else branch to have Diverges::No as its "result", making point Z have the same diverges as point A (presumably No, in your example):

```
// A
if ... {
   diverges(); // X
   diverges(); // Y
}
// Z
```

Y's successor is Z, but Z has two predecessors in the CFG: (roughly) A or Y, depending on if the if was taken. (Really, the predecessor of Z is the if condition and Y, I guess).

Getting this right is likely to be relatively hard, which is why I'm not sure we should be trying to roll this on our own -- a MIR rewrite would make it much easier.

FabianWolff: > Getting this right is likely to be relatively hard, which is why I'm not sure we should be trying to roll this on our own -- a MIR rewrite would make it much easier.

I totally agree, but I don't have the MIR knowledge and time to do this myself at the moment.

JohnCSimon:
Ping from triage: Can you address the comment from the reviewer and fix the build? 
@FabianWolff 

FYI: when a PR is ready for review, post a message containing 
`@rustbot ready` to switch to `S-waiting-on-review` so the PR appears in the reviewer's backlog.

JohnCSimon:
@FabianWolff 
Thanks for your contribution.

@rustbot label: +S-inactive

PR #90639, opened 2021-11-05
Add a query for resolving an impl item from the trait item

matthewjasper:
This makes finding the item in an impl that implements a given trait item a query. This is for a few reasons:

- To slightly improve performance
- To avoid having to do name resolution during monomorphisation
- To make it easier to implement potential future features that create anonymous associated items

### DIFF
--- a/compiler/rustc_typeck/src/check/check.rs
+++ b/compiler/rustc_typeck/src/check/check.rs
@@ -931,8 +931,9 @@ fn check_impl_items_against_trait<'tcx>(
             tcx.sess.delay_span_bug(impl_item.span, "missing associated item in trait");
             continue;
         };
-        match impl_item.kind {
-            hir::AssocItemKind::Const => {
+        let impl_item_full = tcx.hir().impl_item(impl_item.id);
+        match impl_item_full.kind {
+            hir::ImplItemKind::Const(..) => {
                 // Find associated const definition.
                 compare_const_impl(
                     tcx,
@@ -942,7 +943,7 @@ fn check_impl_items_against_trait<'tcx>(
                     impl_trait_ref,
                 );
             }
-            hir::AssocItemKind::Fn { .. } => {
+            hir::ImplItemKind::Fn(..) => {
                 let opt_trait_span = tcx.hir().span_if_local(ty_trait_item.def_id);
                 compare_impl_method(
                     tcx,
@@ -953,12 +954,12 @@ fn check_impl_items_against_trait<'tcx>(
                     opt_trait_span,
                 );
             }
-            hir::AssocItemKind::Type => {
+            hir::ImplItemKind::TyAlias(impl_ty) => {
                 let opt_trait_span = tcx.hir().span_if_local(ty_trait_item.def_id);
                 compare_ty_impl(
                     tcx,
                     &ty_impl_item,
-                    impl_item.span,
+                    impl_ty.span,
                     &ty_trait_item,
                     impl_trait_ref,
                     opt_trait_span,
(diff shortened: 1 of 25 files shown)


compiler/rustc_typeck/src/check/mod.rs
```diff
@@ -248,6 +248,11 @@ pub fn provide(providers: &mut Providers) {
         check_item_well_formed,
         check_trait_item_well_formed,
         check_impl_item_well_formed,
+        impl_item_base_ids,
+        impl_item_implementor_ids: |tcx, def_id| {
+            let local_def_id = def_id.expect_local();
+            tcx.impl_item_base_ids(local_def_id).iter().map(|(i, t)| (*t, *i)).collect()
```
cjgillot, line 254: What is the gain of having two queries instead of one?

cjgillot (commented):
This seems like there is an unfortunate recomputation of `typeck`. Do you have an idea where it comes from?
The three first commits look good. Can they be merged independently?
I have the impression this PR introduces queries for two separate things:
- first a query to translate impl item <-> trait item;
- second a query for associated_item_lookup.

Should their effect be assessed independently?

matthewjasper:
The recomputation probably comes from the `Span` of the `Ident` in `NamedAssocItem` being included in the stable hash. That said I agree that this PR should be split up to make it easier to experiment on each change individually.

matthewjasper:
I can't reproduce any significant perf changes locally. Trying again in case the results were spurious
@bors try @rust-timer queue

cjgillot:
r=me once rebased.

camelid:
Ok, I rebased this PR! @cjgillot this should be ready for a final review :)

@rustbot ready

camelid:
Uh oh... looks like a span is wrong somewhere.

camelid:
I think I accidentally reverted part of 5e1972eba7e95e68f6fc05b07dec15fb1e553e1b, which was landed after this PR's branch was cut from the master branch.

cjgillot:
Just in case:
@bors try @rust-timer queue

camelid:
Perf looks good.

camelid:
Hmm, not sure why the summary included the doc improvements before but not now. The doc improvements are still there though.

compiler/rustc_typeck/src/check/check.rs
```diff
@@ -926,174 +920,82 @@ pub(super) fn check_impl_items_against_trait<'tcx>(
...
         // Check for missing items from trait
         let mut missing_items = Vec::new();
-        for trait_item in tcx.associated_items(impl_trait_ref.def_id).in_definition_order() {
+        for &trait_item_id in tcx.associated_item_def_ids(impl_trait_ref.def_id) {
             let is_implemented = ancestors
-                .leaf_def(tcx, trait_item.ident, trait_item.kind)
-                .map(|node_item| !node_item.defining_node.is_from_trait())
-                .unwrap_or(false);
+                .leaf_def(tcx, trait_item_id)
+                .map_or(false, |node_item| node_item.item.defaultness.has_value());
```
WaffleLapkin, line 985: I think this changed values of `is_implemented`. Previously it was `true` only if it's implemented in this implementation (and was `false` when there were a default impl in the trait). Is this intended? (this broke #92164 and not I'm figuring out how to fix it)




(the first screenshot from e012a191d768adeda1ee36a99ef8b92d51920154 commit, just before these changes, the second is after these changes; I've added a `debug!` in this loop)

WaffleLapkin: And if this was intended, how can I specifically check that this item is implemented manually in a given impl (ie it doesn't use default implementation from the trait)?

PR #90642, opened 2021-11-06
use matches!() macro in more places

matthiaskrgr:


compiler/rustc_mir_build/src/thir/pattern/check_match.rs
```diff
@@ -514,10 +514,7 @@ fn non_exhaustive_match<'p, 'tcx>(
         err.span_label(sp, pattern_not_covered_label(&witnesses, &joined_patterns));
     };
 
-    let is_variant_list_non_exhaustive = match scrut_ty.kind() {
-        ty::Adt(def, _) if def.is_variant_list_non_exhaustive() && !def.did.is_local() => true,
-        _ => false,
-    };
+    let is_variant_list_non_exhaustive = matches!(scrut_ty.kind(), ty::Adt(def, _) if def.is_variant_list_non_exhaustive() && !def.did.is_local());
```
cjgillot, line 517:
```suggestion
    let is_variant_list_non_exhaustive = match scrut_ty.kind() {
        ty::Adt(def, _) => def.is_variant_list_non_exhaustive() && !def.did.is_local(),
        _ => false,
    };
```

compiler/rustc_trait_selection/src/traits/select/mod.rs
```diff
@@ -320,10 +320,7 @@ impl<'cx, 'tcx> SelectionContext<'cx, 'tcx> {
 
     /// Returns `true` if the trait predicate is considerd `const` to this selection context.
     pub fn is_trait_predicate_const(&self, pred: ty::TraitPredicate<'_>) -> bool {
-        match pred.constness {
-            ty::BoundConstness::ConstIfConst if self.is_in_const_context => true,
-            _ => false,
-        }
+        matches!(pred.constness, ty::BoundConstness::ConstIfConst if self.is_in_const_context)
```
cjgillot, line 323:
```suggestion
        matches!(pred.constness, ty::BoundConstness::ConstIfConst) && self.is_in_const_context
```

compiler/rustc_typeck/src/structured_errors/wrong_number_of_generic_args.rs
```diff
@@ -296,12 +296,11 @@ impl<'a, 'tcx> WrongNumberOfGenericArgs<'a, 'tcx> {
...
-                    hir::TyKind::Path(hir::QPath::Resolved(
-                        None,
-                        hir::Path { res: hir::def::Res::Def(_, id), .. },
-                    )) if *id == def_id => true,
-                    _ => false,
+                fn_sig.decl.inputs.iter().any(|ty| {
+                    matches!(ty.kind, hir::TyKind::Path(hir::QPath::Resolved(
+                 None,
+                 hir::Path { res: hir::def::Res::Def(_, id), .. },
+                  )) if *id == def_id)
```
cjgillot, line 303: In general, should we avoid using `matches!` with arm guards? Can't we just simplify the original match by moving the guard into the actual arm?

compiler/rustc_const_eval/src/transform/check_consts/qualifs.rs
```diff
@@ -170,11 +170,11 @@ impl Qualif for NeedsNonConstDrop {
...
             selcx.select(&obligation)
         });
-        match implsrc {
+        !matches!(
+            implsrc,
             Ok(Some(ImplSource::ConstDrop(_)))
-            | Ok(Some(ImplSource::Param(_, ty::BoundConstness::ConstIfConst))) => false,
-            _ => true,
-        }
+                | Ok(Some(ImplSource::Param(_, ty::BoundConstness::ConstIfConst)))
```
fbstj, line 176: is this a candidate for "interior patterns": `Ok(Some(ImplSource::ConstDrop(_) | ImplSource::Param(...)))`

compiler/rustc_mir_build/src/thir/pattern/check_match.rs
matthiaskrgr: reverted

compiler/rustc_typeck/src/structured_errors/wrong_number_of_generic_args.rs
matthiaskrgr: LIke this, right?

````rust
        let is_used_in_input = |def_id| {
            fn_sig.map_or(false, |fn_sig| {
                fn_sig.decl.inputs.iter().any(|ty| match ty.kind {
                    hir::TyKind::Path(hir::QPath::Resolved(
                        None,
                        hir::Path { res: hir::def::Res::Def(_, id), .. },
                    )) => *id == def_id,
                    _ => false,
                })
            })
        };
````

Issue #83163, opened 2021-03-15
Tracking Issue for const_swap


usbalbin:
Feature gate: `#![feature(const_swap)]`

This is a tracking issue for making the functions `mem::swap` and `ptr::swap[_nonoverlapping]` and some other swap related functions `const fn`.

### Public API

```rust
mod ptr {
    pub const unsafe fn swap<T>(x: *mut T, y: *mut T);
}

mod mem {
    pub const fn swap<T>(x: &mut T, y: &mut T);
}

impl<T> [T] {
    pub const fn swap(&mut self, a: usize, b: usize);
}

impl<T: ?Sized> *mut T {
    pub const unsafe fn swap(self, with: *mut T);
}

impl <T> NonNull<T> {
    pub const unsafe fn swap(self, with: NonNull<T>);
}
```

### Steps / History

- [x] Implementation (`mem` and `ptr`): #83091
- [x]  Implementation (`[T]` and `*mut T )` #90644
- [x] `NonNull` methods stabilized, `const`ness moved to this gate https://github.com/rust-lang/rust/pull/124498
- [ ] Final commenting period (FCP)
- [ ] Stabilization PR

### Unresolved Questions

- What do we do with the [bytewise pointer swap situation](https://github.com/rust-lang/rust/issues/83163#issuecomment-1145917372)?

RalfJung:
This got reverted again by https://github.com/rust-lang/rust/pull/86003, but hopefully we can re-land it soon.

RalfJung:
@pnkfelix I don't quite understand why you un-constified `swap`... it seems `write` is under underlying source of the problem. Does it not work to add the const stability attribute [here](https://github.com/rust-lang/rust/pull/86003/files#diff-c175d4e27676febf62c061d31cf9756d256b46e2e44cc6b3177d4ff75e932567R876)?

usbalbin:
I have reverted the constness reverts in #86295

est31:
I've filed #90644 to extend the `const_swap` feature to three more functions. Some of them can panic, but `const_panic` is stable as of #89508.

jhpratt:
This should be blocked on const mut refs, I believe.

RalfJung:
The following code used to work with `swap_nonoverlapping`, but doesn't any more:
```rust
#![feature(const_swap)]
#![feature(const_mut_refs)]
use std::{
    mem::{self, MaybeUninit},
    ptr,
};

const X: () = {
[...]

PR #90644, opened 2021-11-06
Extend the const swap feature

est31:
Adds the `const_swap` feature gate to three more swap functions. cc tracking issue #83163

```Rust
impl<T> [T] {
    pub const fn swap(&mut self, a: usize, b: usize);
    pub const unsafe fn swap_unchecked(&mut self, a: usize, b: usize);
}
impl<T: ?Sized> *mut T {
    pub const unsafe fn swap(self, with: *mut T);
}


Mark-Simulacrum:
Can we squash commits? r=me once that's done

est31:
@Mark-Simulacrum I've squashed them. so re-r? @Mark-Simulacrum

PR #90650, opened 2021-11-06
Document mpsc Sender/Receiver are only Send if T is Send

SebastiaanYN:
Explain that `mpsc::Sender` and `mpsc::Receiver` are only `Send` if `T` is `Send`. This is currently not immediately clear by just looking at the documentation, as the impl does not have a trait bound for `Send`
```rs
impl<T> Sender<T> {
	pub fn send(&self, t: T) -> Result<(), SendError<T>>
}
```
and it can be easy to overlook the `Send` impl
```rs
impl<T: Send> Send for Sender<T>
```

### DIFF
--- a/library/std/src/sync/mpsc/mod.rs
+++ b/library/std/src/sync/mpsc/mod.rs
@@ -315,6 +315,8 @@ mod cache_aligned;
 /// The receiving half of Rust's [`channel`] (or [`sync_channel`]) type.
 /// This half can only be owned by one thread.
 ///
+/// Receiver can only be sent between threads if `T` is [`Send`][`core::marker::Send`]
+///
 /// Messages sent to the channel can be retrieved using [`recv`].
 ///
 /// [`recv`]: Receiver::recv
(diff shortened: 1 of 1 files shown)


library/std/src/sync/mpsc/mod.rs
```diff
@@ -315,6 +315,8 @@ mod cache_aligned;
 /// The receiving half of Rust's [`channel`] (or [`sync_channel`]) type.
 /// This half can only be owned by one thread.
 ///
+/// Receiver can only be sent between threads if `T` is [`Send`][`core::marker::Send`]
```
m-ou-se, line 318: You don't need the link target if the link text already resolves to the right thing:
```suggestion
/// A receiver can only be sent between threads if `T` is [`Send`].
```
(Same below.)

JohnCSimon:
ping from triage:
@SebastiaanYN  
Returning to you to address comments.

`@rustbot ready` to switch back to `S-waiting-on-review`

Dylan-DPC:
@SebastiaanYN closing this due to inactivity and since it's a small PR. Feel free to reopen it when you have the time. Thanks for contributing.

PR #90652, opened 2021-11-06
use filter(|x| matches!(..)) instead of filter_map(|x| match x ... => Some(xy))

matthiaskrgr:


compiler/rustc_resolve/src/late/lifetimes.rs
```diff
@@ -1370,9 +1367,8 @@ impl<'a, 'tcx> Visitor<'tcx> for LifetimeContext<'a, 'tcx> {
...
                             bound_generic_params
                                 .iter()
-                                .filter_map(|param| match param.kind {
-                                    GenericParamKind::Lifetime { .. } => Some(param),
-                                    _ => None,
+                                .filter(|param| {
+                                    matches!(param.kind, GenericParamKind::Lifetime { .. })
                                 })
                                 .enumerate()
                                 .map(|(late_bound_idx, param)| {
```
matthiaskrgr, line 1374: hrm, apparently rustfmt really prefers
````rust
.filter(|param| {
     matches!(param.kind, GenericParamKind::Lifetime { .. })
})
````
over
````rust
 .filter(|param| matches!(param.kind, GenericParamKind::Lifetime { .. }))
````
:/

File: compiler/rustc_typeck/src/coherence/builtin.rs [hunk 2]
```diff
@@ -204,11 +204,11 @@ fn visit_implementation_of_dispatch_from_dyn(tcx: TyCtxt<'_>, impl_did: LocalDef
                                 ))
                                 .emit();
 
-                                return None;
+                                return false;
                             }
                         }
 
-                        Some(field)
+                        return true;
```
spastorino, line 211:
```suggestion
                        true
```

PR #90657, opened 2021-11-06
Fix bug with `#[doc]` string single-character last lines

### LINKED ISSUE #90618: rustdoc cuts off the last line in a `#[doc]` attribute if it only has a single character
When using `#[doc = include_str!("...")]` to insert another file as a docstring, it seems that the last line gets cut off if it contains only one character.

`src/inc1`:
```
inc1
ab
```

`src/inc2`:
```
inc2
x
```

(Both `inc` files have a trailing new line.)

`src/lib.rs`:
```rust
#[doc = include_str!("inc1")]  // OK: "inc1 ab"
pub const A: usize = 1;

#[doc = include_str!("inc2")]  // Bad: "inc2"
pub const B: usize = 2;
```

Running `cargo doc --open` on the above produces output like this:

```
A inc1 ab 
B inc2
```

That is, the `inc1` file gets included correctly, but the `inc2` file loses its second line. This bug only seems to happen if the last line is a single character.

### Meta

Reproduced this bug on both stable and nightly.

`rustc --version --verbose`:
```
rustc 1.56.1 (59eed8a2a 2021-11-01)
binary: rustc
commit-hash: 59eed8a2aac0230a8b53e89d4e99d55912ba6b35
commit-date: 2021-11-01
release: 1.56.1
LLVM version: 13.0.0
```
```
rustc 1.58.0-nightly (4961b107f 2021-11-04)
binary: rustc
commit-hash: 4961b107f204e15b26961eab0685df6be3ab03c6
commit-date: 2021-11-04
release: 1.58.0-nightly
LLVM version: 13.0.0
```

GuillaumeGomez:
Fixes #90618.

This is because `.iter().all(|c| c == '*')` returns `true` if there is no character checked. And in case the last line has only one character, it simply returns `true`, making the last line behind removed.


compiler/rustc_ast/src/util/comments.rs
```diff
@@ -38,7 +38,7 @@ pub fn beautify_doc_string(data: Symbol) -> Symbol {
             i += 1;
         }
         // like the first, a last line of all stars should be omitted
-        if j > i && lines[j - 1].chars().skip(1).all(|c| c == '*') {
+        if j > i && lines[j - 1].is_empty() && lines[j - 1].chars().all(|c| c == '*') {
```
klensy, line 41: if `j-1` index `is_empty()`, there no sense in next check, i think there is some error in logic ;-)

GuillaumeGomez: Woups, forgot to commit local changes... Thanks!

PR #90663, opened 2021-11-07
Support early stopping too old pre-installed `tidy` command for macOS in the HTML checker

ken-matsui:
This PR brings early stopping the HTML checker before errors, which leave some macOS users confused, and suggesting installing a newer `tidy` command.

The pre-installed `tidy` command on macOS is too old, released on 31 October 2006. Additionally, I can see the same date at [StackOverflow](https://stackoverflow.com/questions/22283382/overwrite-osx-tidy
) seven years ago. The `tidy` does not support two indispensable options: `--mute-id` and `--mute`. So, the `./x.py test` command fails with a bunch of errors due not to muting them.

I could confirm the `./x.py test` command before installing a newer `tidy` failed and its command after the installation succeeded.

### DIFF
--- a/src/tools/html-checker/main.rs
+++ b/src/tools/html-checker/main.rs
@@ -79,11 +79,34 @@ fn find_all_html_files(dir: &Path) -> (usize, usize) {
     (files_read, errors)
 }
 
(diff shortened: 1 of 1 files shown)


Mark-Simulacrum:
Seems OK. @bors r+ rollup=iffy

PR #72443, opened 2020-05-21
Add `Arc::new_with()` function for constructing self-referential data structures

Diggsey:
It should be noted that this exposes a new behaviour: a `Weak` can now become "live" after having previously observed it as being "dead".

This does not impact the safety of the `Arc` because we ensure the weak reference count does not reach zero during this time.

~In the process this PR changes the `ArcInner::data` field to use an `UnsafeCell` - my understanding of the unsafe code guidelines is that this change will become necessary even without the new API addition. However, the new function undeniably makes the data field internally mutable, whereas before there seemed to be some debate.~

To me, all of the prior uses of `ptr.as_mut()` seemed suspect as they could result in aliasing mutable references. Using an `UnsafeCell` allows those usages to be replaced with a raw pointer which has no aliasing requirements.

edit:
This UB was fixed in a separate PR.

Perhaps @RalfJung can provide some clarity on whether these additional changes are actually required.

PR #90666, opened 2021-11-07
Stabilize arc_new_cyclic

### LINKED ISSUE #75861: Tracking Issue for arc_new_cyclic
The feature gate for the issue is `#![feature(arc_new_cyclic)]`.

Tracking the following API:

```rust
impl<T> Arc<T> {
    pub fn new_cyclic(data_fn: impl FnOnce(&Weak<T>) -> T) -> Arc<T>;
}

impl<T> Rc<T> {
    pub fn new_cyclic(data_fn: impl FnOnce(&Weak<T>) -> T) -> Rc<T>;
}
```

### Steps

- #72443
- [x] Implement `Arc::new_cyclic` (#75505)
- [x] Implement `Rc::new_cyclic` (#75994)
- [x] FCP: https://github.com/rust-lang/rust/pull/90666#issuecomment-1019288209
- [x] Stabilization PR: https://github.com/rust-lang/rust/pull/90666

### Unresolved Questions

-

bdbai:
This stabilizes feature `arc_new_cyclic` as the implementation has been merged for one year and there is no unresolved questions. The FCP is not started yet.

Closes #75861 .

@rustbot label +T-libs-api

### DIFF
--- a/library/alloc/src/rc.rs
+++ b/library/alloc/src/rc.rs
@@ -374,32 +374,47 @@ impl<T> Rc<T> {
         }
     }
 
-    /// Constructs a new `Rc<T>` using a weak reference to itself. Attempting
-    /// to upgrade the weak reference before this function returns will result
-    /// in a `None` value. However, the weak reference may be cloned freely and
-    /// stored for use at a later time.
+    /// Constructs a new `Rc<T>` using a closure `data_fn` that has access to a
+    /// weak reference to the constructing `Rc<T>`.
+    ///
+    /// Generally, a structure circularly referencing itself, either directly or
+    /// indirectly, should not hold a strong reference to prevent a memory leak.
+    /// In `data_fn`, initialization of `T` can make use of the weak reference
+    /// by cloning and storing it inside `T` for use at a later time.
+    ///
+    /// Since the new `Rc<T>` is not fully-constructed until `Rc<T>::new_cyclic`
+    /// returns, calling [`upgrade`] on the weak reference inside `data_fn` will
+    /// fail and result in a `None` value.
+    ///
+    /// # Panics
+    /// If `data_fn` panics, the panic is propagated to the caller, and the
+    /// temporary [`Weak<T>`] is dropped normally.
     ///
     /// # Examples
     ///
     /// ```
-    /// #![feature(arc_new_cyclic)]
     /// #![allow(dead_code)]
     /// use std::rc::{Rc, Weak};
     ///
     /// struct Gadget {
-    ///     self_weak: Weak<Self>,
-    ///     // ... more fields
+    ///     me: Weak<Gadget>,
     /// }
+    ///
     /// impl Gadget {
-    ///     pub fn new() -> Rc<Self> {
-    ///         Rc::new_cyclic(|self_weak| {
-    ///             Gadget { self_weak: self_weak.clone(), /* ... */ }
-    ///         })
+    ///     /// Construct a reference counted Gadget.
+    ///     fn new() -> Rc<Self> {
+    ///         Rc::new_cyclic(|me| Gadget { me: me.clone() })
+    ///     }
+    ///
+    ///     /// Return a reference counted pointer to Self.
+    ///     fn me(&self) -> Rc<Self> {
+    ///         self.me.upgrade().unwrap()
     ///     }
     /// }
     /// ```
+    /// [`upgrade`]: Weak::upgrade
     #[cfg(not(no_global_oom_handling))]
-    #[unstable(feature = "arc_new_cyclic", issue = "75861")]
+    #[stable(feature = "arc_new_cyclic", since = "1.59.0")]
     pub fn new_cyclic(data_fn: impl FnOnce(&Weak<T>) -> T) -> Rc<T> {
         // Construct the inner in the "uninitialized" state with a single
         // weak reference.
--- a/library/alloc/src/rc.rs
+++ b/library/alloc/src/rc.rs
@@ -415,7 +415,10 @@ impl<T> Rc<T> {
     /// [`upgrade`]: Weak::upgrade
     #[cfg(not(no_global_oom_handling))]
     #[stable(feature = "arc_new_cyclic", since = "1.59.0")]
-    pub fn new_cyclic(data_fn: impl FnOnce(&Weak<T>) -> T) -> Rc<T> {
+    pub fn new_cyclic<F>(data_fn: F) -> Rc<T>
+    where
+        F: FnOnce(&Weak<T>) -> T,
+    {
         // Construct the inner in the "uninitialized" state with a single
         // weak reference.
         let uninit_ptr: NonNull<_> = Box::leak(box RcBox {
--- a/library/alloc/src/rc.rs
+++ b/library/alloc/src/rc.rs
@@ -414,7 +414,7 @@ impl<T> Rc<T> {
     /// ```
     /// [`upgrade`]: Weak::upgrade
     #[cfg(not(no_global_oom_handling))]
-    #[stable(feature = "arc_new_cyclic", since = "1.59.0")]
+    #[stable(feature = "arc_new_cyclic", since = "1.60.0")]
     pub fn new_cyclic<F>(data_fn: F) -> Rc<T>
     where
         F: FnOnce(&Weak<T>) -> T,
--- a/library/alloc/src/sync.rs
+++ b/library/alloc/src/sync.rs
@@ -391,7 +391,7 @@ impl<T> Arc<T> {
     /// [`upgrade`]: Weak::upgrade
     #[cfg(not(no_global_oom_handling))]
     #[inline]
-    #[stable(feature = "arc_new_cyclic", since = "1.59.0")]
+    #[stable(feature = "arc_new_cyclic", since = "1.60.0")]
     pub fn new_cyclic<F>(data_fn: F) -> Arc<T>
     where
         F: FnOnce(&Weak<T>) -> T,
--- a/library/alloc/src/sync.rs
+++ b/library/alloc/src/sync.rs
@@ -351,29 +351,47 @@ impl<T> Arc<T> {
         unsafe { Self::from_inner(Box::leak(x).into()) }
     }
 
-    /// Constructs a new `Arc<T>` using a weak reference to itself. Attempting
-    /// to upgrade the weak reference before this function returns will result
-    /// in a `None` value. However, the weak reference may be cloned freely and
-    /// stored for use at a later time.
+    /// Constructs a new `Arc<T>` using a closure `data_fn` that has access to
+    /// a weak reference to the constructing `Arc<T>`.
     ///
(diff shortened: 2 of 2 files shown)


dtolnay:
I like these!

@rust-lang/libs-api 
@rfcbot fcp merge

```rust
impl<T> Rc<T> {
    pub fn new_cyclic(f: impl FnOnce(&Weak<T>) -> T) -> Rc<T>;
}

impl<T> Arc<T> {
    pub fn new_cyclic(f: impl FnOnce(&Weak<T>) -> T) -> Arc<T>;
}
```

BurntSushi:
I think the docs for this API leave a lot to be desired. I can't figure out what this API is really supposed to do or, more importantly, why or when I should (or shouldn't) use it. Here are some points of confusion that I have after looking at this for a few minutes:

* The name of the API includes the word "cyclic," but nothing in the docs mention it.
* I'm grateful for the existence of the example, but the example doesn't really help me understand what this routine does or why I would want to use it. It kind of seems like a "trivial" example to me, which is fine in some cases, but this API feels like it deserves something better.
* The docs mention a `None` value, but I don't quite grok this. I don't see any `Option<T>` anywhere, so it's not clear to me where this `None` value comes from.
* What happens if `data_fn` panics? Does a leak occur? Something worse? Should we specify the behavior of this routine if `data_fn` panics?

My suspicion is that this is a fine API to add, but I'd like to see the docs improved here a bit before signing off.

yaahc:
> * The docs mention a `None` value, but I don't quite grok this. I don't see any `Option<T>` anywhere, so it's not clear to me where this `None` value comes from.

It looks like the None is referring to calling `upgrade` on the `Weak` argument inside of the provided closure. Seems easy to clarify by turning the `upgrade` reference into a link to the relevant function on `Weak`.

yaahc:
> * I'm grateful for the existence of the example, but the example doesn't really help me understand what this routine does or why I would want to use it. It kind of seems like a "trivial" example to me, which is fine in some cases, but this API feels like it deserves something better.

I did some digging in the history of this API to see if I couldn't find any better explanations and found [this comment](https://github.com/rust-lang/rust/pull/72443#issuecomment-651046259) which explicitly mentions being able to give out reference counted pointers to ones self. Perhaps the example just needs to additionally include a simple method on `Foo`, naming notwithstanding.

```rust
impl Foo {
    fn this_ptr(&self) -> Arc<Self> {
        self.me.upgrade()
    }
}
```

yaahc:
@rfcbot reviewed
@rfcbot concern clarify doc example

ogoffart:
That's similar to the current docs for Rc::new_cyclic:  https://doc.rust-lang.org/std/rc/struct.Rc.html#method.new_cyclic

yaahc:
Fair point, we should definitely also make sure all the docs on `Rc::new_cyclic` and `Arc::new_cyclic` are consistent.

The necessary changes as I see them:

- [ ] Update `Arc::new_cyclic` docs to add the methods mentioned above
- [ ] Rename the `Foo` type in the `Arc` example to `Gadget`
- [ ] Add the second method from the methods mentioned above to the `Rc` example
- [ ] Update the description for both `Arc` and `Rc` so that the mention of `upgrade` becomes a link to `Weak::upgrade`

bdbai:
Regarding the doc example, I would like to propose the following alternative:

```rust
use std::sync::{Arc, Weak};

struct Root {
    left: Leaf,
    right: Leaf,
}

struct Leaf {
    root: Weak<Root>,
}

let tree = Arc::new_cyclic(|me| {
    assert!(root.upgrade().is_none());
    Root {
        left: Leaf { root: me.clone() },
        right: Leaf { root: me.clone() },
    }
});

assert!(tree.left.root.upgrade().is_some());
```

I noticed there was a section [Breaking cycles with `Weak`](https://doc.rust-lang.org/std/sync/struct.Arc.html#breaking-cycles-with-weak) in the documentation of `Arc<T>`, where a tree-like data structure was introduced as an example. I guess the doc example here is the right place to reflect that.

bdbai:
The description was amended as follows:

>    /// Constructs a new `Arc<T>` using a closure `data_fn` that has access to
>    /// a weak reference to the constructing `Arc<T>`.
>    ///
>    /// Generally, a structure circularly referencing itself, either direct or
>    /// indirect, should not hold a strong reference to prevent a memory leak.
>    /// In `data_fn`, initialization of `T` can make use of the weak reference
>    /// by cloning and storing it inside `T` for use at a later time.
>    ///
>    /// Since the new `Arc<T>` is not fully-constructed until
>    /// `Arc<T>::new_cyclic` returns, calling [`upgrade`] on the weak
>    /// reference inside `data_fn` will fail and result in a `None` value.
>    ///
>    /// # Panics
>    /// If `data_fn` panics, the panic is propagated to the caller, and the
>    /// temporary [`Weak<T>`] is dropped normally.
>    ///
>    /// # Example
>    /// ...
>    /// [`upgrade`]: Weak::upgrade

Since I am not a native English speaker, I would appreciate it if any suggestions could be made in order to make readers feel more idiomatic. 😊

yaahc:
@bdbai I edited a couple of the verb tenses in your updated doc comment but otherwise everything in your example looked correct grammatically. Am I correct in understanding that you've not made the mentioned changes in the PR itself?

As for the example, I have a slight preference for the example that I suggested earlier since it seems closer to real world examples while still being relatively brief, where as I feel like the tree example you linked is very simplified from what a real world tree structure would look like and is still a much longer example. That said, I'm fine with either example, so I'll leave it up to other libs team members to voice a stronger preference.

I would still like to see the Rc and Arc docs updated to match regardless of which example we go with.

bdbai:
@yaahc 

> I edited a couple of the verb tenses in your updated doc comment but otherwise everything in your example looked correct grammatically.

Thanks!

> Am I correct in understanding that you've not made the mentioned changes in the PR itself?

I failed to mention that the changes of the text description would be committed into the PR once it got corrected and approved. Sorry about that.

> I would still like to see the Rc and Arc docs updated to match regardless of which example we go with.

Sure. I will update both while changing the codebase. For the sake of brevity, I would like to mention only one of them during discussion since the both does not differ in terms of this API.

bdbai:
@yaahc The documentation in `Arc::new_cyclic` and `Rc::new_cyclic` has been updated.

btw why did rustbot add the `T-compiler` tag?

yaahc:
@bdbai not sure what's up with the T-compiler label but the updates look great!

@rfcbot resolve clarify doc example

veber-alex:
Can we change the signatures of those methods to use "normal generics" instead of `impl Trait` ?
It seems to be more consistent with the rest of the standard library.

bdbai:
@veber-alex fixed.

wx-csy:
Can't wait for the stabilization to build data structures with circular references!

Currently, this function takes a closure to initialize the arc, but it operates badly with async programming. I think we can offer an API that returns a pair of `(Initializer, Weak<T>)`, where the non-clonable `Initializer` can be used to initialize the arc and after that, the weak pointer can be promoted to a strong pointer. In this case, I can pass the initializer to an async block then do the initialization asynchronously.

bdbai:
@wx-csy good point!
I suggest this feature may be generalized as `ArcBuilder` with the following API interface:

```rust
struct ArcBuilder<T> { ... }
impl<T> ArcBuilder<T> {
    fn new() -> Self;
    fn weak(&self) -> &Weak<T>;
    fn build(self) -> Arc<T>;
}
```

Since the pair of `new_cyclic` functions have been in the standard library for a while, reworking this feature may introduce breaking changes. @yaahc what do you think?

m-ou-se:
That sounds like a discussion to be had on the tracking issue, before stabilizing this: https://github.com/rust-lang/rust/issues/75861

wx-csy:
> @wx-csy good point! I suggest this feature may be generalized as `ArcBuilder` (or `TempArc`?) with the following API interface:
> 
> ```rust
> struct ArcBuilder<T> { ... }
> impl<T> ArcBuilder<T> {
>     fn new() -> Self;
>     fn get_weak(&self) -> &Weak<T>;
>     fn with_value(self, value: T) -> Arc<T>;
> }
> ```
> 
> Once implemented, it will also address the need of fallible cyclic `Arc` creation mentioned [here](https://github.com/rust-lang/rust/issues/75861#issuecomment-817809279). Furthermore, interoperability with `MaybeUninit` can be added to it.
> 
> Since the pair of `new_cyclic` functions have been in the standard library for a while, reworking this feature may introduce breaking changes. @yaahc what do you think?

Well, I think it's fine to move forward stabilizing `new-cyclic` and just add `ArcBuilder` later. After that, we may mark `new-cyclic` as a shorthand for `ArcBuilder`.

Btw, we may additionally return an `ArcBuilder` when unwrapping the last strong pointer (e.g., in `Arc::try_unwrap`), so that we may reinitialize the arc within exactly the same allocation, though I've not yet found a real-world use case of this :)

library/alloc/src/rc.rs
```diff
@@ -374,33 +374,51 @@ impl<T> Rc<T> {
...
+    ///     fn me(&self) -> Rc<Self> {
+    ///         self.me.upgrade().unwrap()
     ///     }
     /// }
     /// ```
+    /// [`upgrade`]: Weak::upgrade
     #[cfg(not(no_global_oom_handling))]
-    #[unstable(feature = "arc_new_cyclic", issue = "75861")]
-    pub fn new_cyclic(data_fn: impl FnOnce(&Weak<T>) -> T) -> Rc<T> {
+    #[stable(feature = "arc_new_cyclic", since = "1.59.0")]
```
m-ou-se, line 417:
```suggestion
    #[stable(feature = "arc_new_cyclic", since = "1.60.0")]
```

library/alloc/src/sync.rs
```diff
@@ -351,30 +351,51 @@ impl<T> Arc<T> {
...
+    ///         self.me.upgrade().unwrap()
+    ///     }
+    /// }
     /// ```
+    /// [`upgrade`]: Weak::upgrade
     #[cfg(not(no_global_oom_handling))]
     #[inline]
-    #[unstable(feature = "arc_new_cyclic", issue = "75861")]
-    pub fn new_cyclic(data_fn: impl FnOnce(&Weak<T>) -> T) -> Arc<T> {
+    #[stable(feature = "arc_new_cyclic", since = "1.59.0")]
```
m-ou-se, line 394:
```suggestion
    #[stable(feature = "arc_new_cyclic", since = "1.60.0")]
```

m-ou-se:
> Well, I think it's fine to move forward stabilizing `new-cyclic` and just add `ArcBuilder` later. After that, we may mark `new-cyclic` as a shorthand for `ArcBuilder`.

Sounds good.

Note that this `ArcBuilder` idea has some overlap with the `ArcMut` idea mentioned here: https://github.com/rust-lang/rust/issues/63291#issuecomment-565378780

m-ou-se:
> This will be merged soon.

@bors r+

PR #90670, opened 2021-11-07
Reformat doc-comments in `core::alloc`

mbartlett21:


jyn514:
@mbartlett21 what is the goal of this change? Why is the new formatting better?

scottmcm:
I think I'm with jyn514 on this one -- if this isn't fixing a tidy suppression, I'm not convinced this is worth doing.  It doesn't seem obviously better, and making `git blame` that little bit more cluttered has a cost, as does potentially causing merge conflicts for people.

What was the impetus to make this change?  I'm open to being convince it's worth doing.

mbartlett21:
Ok. I'll leave this then.

PR #90675, opened 2021-11-07
rustdoc: Cleanup `clean::Impl` and other parts of `clean`

camelid:
This PR cleans up and reduces the size of `clean::Impl`, makes some other small performance improvements, and removes some Clean impls that are either unnecessary or potentially confusing.

r? @jyn514

### DIFF
--- a/src/librustdoc/clean/mod.rs
+++ b/src/librustdoc/clean/mod.rs
@@ -80,12 +80,6 @@ impl<T: Clean<U>, U> Clean<U> for Rc<T> {
     }
 }
 
-impl<T: Clean<U>, U> Clean<Option<U>> for Option<T> {
-    fn clean(&self, cx: &mut DocContext<'_>) -> Option<U> {
-        self.as_ref().map(|v| v.clean(cx))
-    }
-}
-
 impl Clean<Item> for doctree::Module<'_> {
     fn clean(&self, cx: &mut DocContext<'_>) -> Item {
         let mut items: Vec<Item> = vec![];
@@ -460,7 +454,7 @@ impl Clean<GenericParamDef> for hir::GenericParam<'_> {
                 GenericParamDefKind::Type {
                     did: cx.tcx.hir().local_def_id(self.hir_id).to_def_id(),
                     bounds: self.bounds.clean(cx),
-                    default: default.clean(cx).map(Box::new),
+                    default: default.map(|t| t.clean(cx)).map(Box::new),
                     synthetic,
                 },
             ),
@@ -935,7 +929,7 @@ impl Clean<Item> for hir::TraitItem<'_> {
                     TyMethodItem(t)
                 }
                 hir::TraitItemKind::Type(bounds, ref default) => {
-                    AssocTypeItem(bounds.clean(cx), default.clean(cx))
+                    AssocTypeItem(bounds.clean(cx), default.map(|t| t.clean(cx)))
                 }
             };
             let what_rustc_thinks =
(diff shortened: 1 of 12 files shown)


src/librustdoc/clean/mod.rs
```diff
@@ -80,12 +80,6 @@ impl<T: Clean<U>, U> Clean<U> for Rc<T> {
     }
 }
 
-impl<T: Clean<U>, U> Clean<Option<U>> for Option<T> {
-    fn clean(&self, cx: &mut DocContext<'_>) -> Option<U> {
-        self.as_ref().map(|v| v.clean(cx))
-    }
-}
```
jyn514, line 87: Why this change?

jyn514 (commented):
Why did you remove the Clean impls that was used? That seems like a lot of churn for not a ton of benefit ... the rest of the commits are great though :)

src/librustdoc/clean/types.rs
```diff
@@ -2178,11 +2178,22 @@ crate struct Impl {
...
+        self.kind.is_auto()
+    }
+
+    crate fn is_blanket_impl(&self) -> bool {
+        self.kind.is_blanket()
+    }
+
+    crate fn blanket_impl_ty(&self) -> Option<&Type> {
+        self.kind.as_blanket_ty()
+    }
```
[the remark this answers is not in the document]
camelid: The main reason I have `blanket_impl_ty()` is to make future changes to the representation easier. I'm mixed about it though, so if you prefer, I'm happy to remove the top-level wrapper functions like this one.

src/librustdoc/clean/mod.rs
camelid: I'd rather be more explicit that the code is dealing with an `Option`. I find the whole `Clean` trait quite confusing, and this way it's a bit simpler IMO. The impl was only used in a few places too.

If you want me to put it back, I can, but I think it's clearer to use `map` at the call site.

src/librustdoc/clean/types.rs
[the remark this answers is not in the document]
jyn514: Yeah, I think the wrapper functions for `kind` enough to make future changes easier :)

src/librustdoc/clean/mod.rs
jyn514: I think as long as `Clean` is a trait it makes sense to keep the impls on wrapper types (the vec diff especially was super big and didn't seem particularly illuminating). If you want to make a PR at some point removing clean, I would be ok with removing these then :)

camelid:
> Why did you remove the Clean impls that was used? That seems like a lot of churn for not a ton of benefit ...

Overall, the reasons are as follows:

1. I'd rather the code be more explicit about what is happening (e.g., that a Vec is being mapped over).
2. These removals are small steps in the direction of getting rid of Clean altogether, or at least significantly reducing its use.
3. Unused trait impls are not detected by the compiler and can thus accumulate. Removing them proactively prevents this from happening ;)

I also have specific reasons for each removed impl (except for the unused `Rc<T>` impl, and the `&T` impl that could be removed just by skipping an `&` in one spot):

- Option: https://github.com/rust-lang/rust/pull/90675#discussion_r744361265
- IndexVec: It was only used in one place, so it seems better to just write the code there.
- Vec: Like with Option, I'd prefer the code be more explicit about the mapping and what impl is being dispatched to. Also, this removal uncovered some places that were collecting into a Vec and *then* running `clean` on that Vec, which causes unnecessary allocations and re-traversals. Forcing a manual `map` helps prevents situations like this from happening in the first place.

src/librustdoc/clean/mod.rs
camelid: > If you want to make a PR at some point removing clean, I would be ok with removing these then :)

The problem there is that removing Clean all at once would be a huge PR. This is a small step that would make a future removal more feasible.

> (the vec diff especially was super big and didn't seem particularly illuminating)

Also, see my comments about hidden performance issues caused by the Vec impl below :)

jyn514:
r=me with https://github.com/rust-lang/rust/pull/90675#discussion_r744347462 done :)

@bors rollup=never

camelid:
Done!

@bors r=jyn514 rollup=never

\[you can never say rollup=never too many times ;)\]

Issue #35428, opened 2016-08-06
Add `is_empty` function to `ExactSizeIterator`


frewsxcv:
Tracking issue for functionality introduced in https://github.com/rust-lang/rust/pull/34357

frewsxcv:
Added reference to this issue in https://github.com/rust-lang/rust/pull/35429

SimonSapin:
I wanted to use this today until I realized that it’s unstable. (I could use `.len() == 0` in the mean time.) Seems straightforward enough. FCP to stabilize?

alexcrichton:
@rfcbot fcp merge

Seems good to have consistency!

bluss:
@SimonSapin I'm curious, when do you use ExactSizeIterator and have the opportunity to use this?

SimonSapin:
When using a slice or vec iterator as input for a parser, `.is_empty()` is a way to tell "have I reach the end of the input yet?" https://github.com/servo/html5ever/blob/b5c4552fab/macros/match_token.rs#L175

bluss:
Aha, that explains it for me: using a specific iterator type, not using ESI generically. This method is good, reinforces existing conventions.

bluss:
used in PR #37943

bluss:
Here's a thought: Some iterators can implement a good is_empty() even if they can't do ESI or .len(). One example is `.chars()`. Does it matter?

aturon:
@bluss Sorry I missed your comment when prepping the stabilization PR.

My feeling is that these iterators can/should provide an inherent `is_empty` method if we think that's useful. The addition of the method to `ExactSizeIterator` is more about convenience than generic programming, IMO.

FWIW, it's also a longstanding desire on the lang side to have an API evolution path for splitting a trait into supertraits without breaking existing code. It's not guaranteed to happen, but that would allow us to split out `is_empty` into its own trait if we wanted to program generically with it in `std` in the future. (Of course, we can always add a separate trait with a blanket impl, worst case.)

aturon:
I'm removing `is_empty` from the current stabilization PR, to give more time for discussion here.

bluss:
- I think emptiness should be a trait, so that it can pass through adapters
- Iterators as lazy sequences have a different relationship to emptiness than collections, and the common `chars()` can already tell emptiness but not length; it's a common case in for example parsers (maybe not the most common, since byte-level parsing seems to be the most popular).

aturon:
@bluss 

[...]

PR #90676, opened 2021-11-07
Add `is_empty()` to Iterator

fee1-dead:
This is a draft PR as an alternative to #35428.

The original feature was blocked because we felt that it shouldn't be specific to `ExactSizeIterator`, as some iterators can know whether they are empty or not but do not have an exact size e.g. `Filter`.

I didn't see any proposal looking exactly like this. @scottmcm created [an experiment](https://github.com/rust-lang/rust/compare/master...scottmcm:is_empty-alternative?expand=1) that separated it to its own trait, but the implementation seemed too restrictive to me, and adding another trait might not be worth it for a single method: functions using `Iterator`s should not require the iterator to know whether it is empty or not, it should already be able to tell using the size hint.

cc @SimonSapin

*EDIT:* Let's discuss this change before me fixing errors. I could also just remove `ExactSizeIterator::is_empty()` in this PR.

### DIFF
--- a/library/core/src/iter/traits/iterator.rs
+++ b/library/core/src/iter/traits/iterator.rs
@@ -161,6 +161,33 @@ pub trait Iterator {
         (0, None)
     }
 
+    /// Returns whether the iterator has no items left.
+    /// 
+    /// The default implementation returns `true` when [`size_hint()`] returns <code>(0, [Some](0))</code>,
+    /// and `false` otherwise.
+    ///
+    /// # Examples
+    ///
+    /// Basic usage:
+    ///
+    /// ```
+    /// let a = [1, 2, 3];
+    /// let mut iter = a.iter();
+    ///
+    /// assert!(!iter.is_empty());
+    /// 
+    /// for _ in iter {
+    ///     // consume the iterator, leaving no items left
+    /// }
+    /// 
+    /// assert!(iter.is_empty());
+    /// ```
+    #[inline]
+    #[unstable(feaure = "iter_is_empty", reason = "recently added", issue = "0")]
+    fn is_empty(&self) -> bool {
+        (0, Some(0)) == self.size_hint()
+    }
+
     /// Consumes the iterator, counting the number of iterations and returning it.
     ///
     /// This method will call [`next`] repeatedly until [`None`] is encountered,


library/core/src/iter/traits/iterator.rs
```diff
@@ -161,6 +161,33 @@ pub trait Iterator {
...
+    /// for _ in iter {
+    ///     // consume the iterator, leaving no items left
+    /// }
+    /// 
+    /// assert!(iter.is_empty());
+    /// ```
+    #[inline]
+    #[unstable(feaure = "iter_is_empty", reason = "recently added", issue = "0")]
+    fn is_empty(&self) -> bool {
+        (0, Some(0)) == self.size_hint()
```
scottmcm, line 188: There's no requirement that the `size_hint` be this precise (unless it's `ExactSizeIterator`), so I don't think this implementation is ok.  For example, this won't work with `from_fn`: <https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=0fce510cb4a02171b289d6d3bd0e8808>

This would either need to be `-> Option<bool>` (so it can return `None` if the size_hint is inconclusive), or be on a different trait (as you mention).

fee1-dead: `from_fn` obviously doesn't work, so I don't see why this is bad. This method is not always correct when it returns `false` (return value of `false` does not mean the iterator has items), but it should be correct when it returns `true`.

scottmcm: That's not what I'd expect it to do, at least.  I'd expect it, since Rust has all the type system machinery it needs to do this, to only compile when it works -- same as `.len()` isn't "well, here's what it might be".

Someone who wants unreliable checks can always just do `.size_hint().0 > 0` or `.size_hint().1 == Some(0)`, depending on which direction they want; I don't think convenience methods for that are appropriate.

fee1-dead: ESI is a safe trait, and to me `len()` is also "here's what it might be" since we can't say not reporting accurate size is UB. It doesn't have to reside in another trait, but we could add a trait for users to require iterators to produce accurate `is_empty` results, like `FusedIterator`.

the8472:
> it should already be able to tell using the size hint.

The size hint is not sufficient since returning `(0, None)` is a valid and in fact the default implementation for it. If we get generator-based iterators in the future than computing a better size hint than that may be equivalent to solving the halting problem.
So for a vast class of iterators this method would always return false even when they are empty.

library/core/src/iter/traits/iterator.rs
the8472: > ESI is a safe trait, and to me len() is also "here's what it might be" since we can't say not reporting accurate size is UB.

UB is not the only kind of forbidden behavior. Safe traits can also make promises where breaking these promises is a bug that can cascade into further downstream misbehavior such as infinite loops, panics, leaks and other API contracts being broken which easily can result in data loss or miscomputations. Sure, it's not a buffer overflow or double-free but that's little consolation when your blob storage containing petabytes of critical data suddenly become inaccessible, your account balance gets miscomputed or an airliner has an intersection event with a mountain.

So from the perspective of *safe code* `ExactSizeIterator::len` is guaranteed to be correct and it is free to operate on that assumption for all `T`. B is correct if A is correct.

`is_empty` here does not meet this level of conditional correctness and would have to be weakened to the level of uselessness because it would have to explicitly mention something like

> For some iterator types this function may always return `false` even when they are empty.

Which renders it useless as a *trait method*, because then people would have to reason about concrete types rather than the trait.

